Issue #5286, opened 2023-04-28
Migrate `SQL Runner sidebar` from Blueprint to Mantine


IrakliJani:
### Description

Migrate SQL runner sidebar to Mantine

<img width="396" alt="CleanShot 2023-04-28 at 16 28 51@2x" src="https://user-images.githubusercontent.com/962095/235147550-00e7dd38-44d0-4343-b0ec-75398ef53885.png">

PR #5296, opened 2023-05-01
feat: mantine sql runner sidebar

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/5286

### Description:

old
CleanShot 2023-05-01 at 11 48 36@2x

vs new
CleanShot 2023-04-30 at 23 46 02@2x


packages/frontend/src/components/common/SqlRunner/CatalogTree.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
joaoviana, line 1: I don't think we need to import React here 🤔

IrakliJani: using React.FC below

joaoviana: Yup, you shouldn't need to import it because our React version knows what to do. 
If you're always referencing `React.X` then no need to import. 
If using named imports, then yes 🤔

IrakliJani: updated

IrakliJani: we need `FC` because I use `children` we don't need whole React import because of JSX not requiring it anymore if that's what you mean

joaoviana: Yes, it was more on the usage of imports, rather than what you were importing, which I agree 100%(with using `FC` and `children`)

Issue #4923, opened 2023-03-27
migrate `forgot password` page from Blueprint to Mantine


IrakliJani:
### Description

migrate

<img width="658" alt="CleanShot 2023-03-27 at 15 29 33@2x" src="https://user-images.githubusercontent.com/962095/227929267-6339a920-cbf0-4d07-8cbf-0a1c321ebe9e.png">

PriPatel:
@joaoviana In this process if its a quick one, please can you remove the monkey emoji?😅
It's soo not appropriate any more :D

joaoviana:
> @joaoviana In this process if its a quick one, please can you remove the monkey emoji?😅 It's soo not appropriate any more :D

sure thing @PriPatel , will remove on the PR 👍

PR #5306, opened 2023-05-02
feat: migrate forgot password form to mantine

joaoviana:
Closes: #4923 

### Description:

Before:
image

After: 
image

Before:
Screenshot 2023-05-02 at 10 56 35

After:
image

### DIFF
--- a/packages/frontend/src/pages/PasswordRecovery.tsx
+++ b/packages/frontend/src/pages/PasswordRecovery.tsx
@@ -1,15 +1,10 @@
-import React, { FC } from 'react';
+import { Card, Image, Stack } from '@mantine/core';
+import { FC } from 'react';
 import { Redirect } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/pages/PasswordRecovery.tsx
```diff
@@ -25,14 +20,18 @@ const PasswordRecovery: FC = () => {
 
     return (
         <Page isFullHeight>
-            <FormWrapper>
-                <LogoWrapper>
-                    <Logo src={LightdashLogo} alt="lightdash logo" />
-                </LogoWrapper>
-                <CardWrapper elevation={2}>
+            <Stack w={400} mt="4xl">
```
yasmin-e, line 23: can you add a FIXME comment here so we update the width to a non-hardcoded value later?

PR #5331, opened 2023-05-02
refactor: create common EmptyState component

joaoviana:
### Description:

Extract `EmptyState` component so it can be used as an alternative to `NonIdealState` when migrating to mantine

(cc @yasmin-e )

Thanks @IrakliJani for pointing this out!


IrakliJani (approved):
🚀

Issue #5076, opened 2023-04-13
User is being shown incorrect stats on their Usage Analytics dashboard


PriPatel:
<img width="980" alt="Screenshot 2023-04-13 at 08 23 18" src="https://user-images.githubusercontent.com/12776522/231683774-816eb08d-3b63-4a61-bdcb-23913f6b2fa7.png">

[The user](https://lightdash-community.slack.com/archives/C03MG2VTR08/p1681370560918039) knows for a fact that they have users in Lightdash not running queries and they definitely know that more than 3 chart updates happen in a week. Needs investigating.

TuringLovesDeathMetal:
Another reporting of a usage analytics data issue!

In one of the instances, the table of `users who have not run a query in the last 90 days` is empty. Even though you can see from this table: https://analytics.lightdash.cloud/share/p-31wprcxlgfQXyOX3Xaq, that there are users who were created > 90 days ago who have **_never_** run a query.

rephus:
SOmething might be wrong on this SQL https://github.com/lightdash/lightdash/blob/main/packages/backend/src/models/AnalyticsModelSql.ts#L90

PR #5346, opened 2023-05-03
fix: no queries analytics

rephus:
Closes: #5076

### Description:

We now show users who have never run a query 

image


TuringLovesDeathMetal (approved):
weoooo! perfect. Hopefully this fixes it for most people 🤞

owlas (approved):

PR #5352, opened 2023-05-03
fix: remove unused PasswordInput

joaoviana:
### Description:

remove unused PasswordInput. 

now we only have on PasswordInput component!


owlas:
approved if tests pass ✅

Issue #5069, opened 2023-04-12
If new series are added to a chart, they don't respect my `smooth` chart settings


rephus:
### Description

If I have a chart setting like this 
![image](https://user-images.githubusercontent.com/1983672/231495553-09c05ee2-9930-44b5-95be-383921fb9502.png)

And later a new series is added to the chart, then this new serie is added with the default settings (smooth off) 

![image](https://user-images.githubusercontent.com/1983672/231495808-28bb50c7-0fe1-4209-93b3-dc31946712cf.png)

Not sure if this is affecting `grouped  series` only. 

My guess is that this property is added directly to the series when edited, but the settings is generic, it should be possible to re-apply this setting on real time

PR #5353, opened 2023-05-03
fix: fix new smooth grouped series

rephus:
Closes:  #5069
### Description:

Note: We already check the `first of series` smooth/showSymbol value to set the default. I'm applying the same logic to all series. 

Screenshot from 2023-05-03 15-36-22

Before

[Screencast from 03-05-23 15:34:26.webm](https://user-images.githubusercontent.com/1983672/235933200-7f9824e2-36db-4a67-94ca-7684dfcf3bae.webm)

After

[Screencast from 03-05-23 15:35:15.webm](https://user-images.githubusercontent.com/1983672/235933217-08ac5046-bed3-447a-a337-52e999bd876c.webm)


TuringLovesDeathMetal (approved):
Nice one! Looks great :) loving these few-line fixes today

owlas (approved):
Just to confirm, checking the first series is enough because we only let the user configure "smooth" for all series but not for the individual series. Is that right?

Is it ever possible that the new series added are the first series?

So in this case:

1. On save the chart config has `[{ smooth: true }]`
2. On load there is a new series (from pivoting probably)
3. The new series appears first `[{}, {smooth: true}]`
4. Now we apply smooth false to everything: `[{ smooth: false }, { smooth: false }]`

owlas:
^ sorry this was meant to be "comment" not "approve"

rephus:
> Just to confirm, checking the first series is enough because we only let the user configure "smooth" for all series but not for the individual series. Is that right?

checking the first series is enough because that's the logic we already apply when setting the "default" smooth for all the series in the chat config panel. 

You can still set a different smooth for individual series and override it (if overrided, we keep its value) 

Because I do: 

```
smooth: firstSerie?.smooth, 
...series, 
```

if a serie has a smooth value set already, it shoudl override the firstSerie?.smooth value, right ?

rephus:
> the UI is a bit confusing

I agree, I thought we had this `smooth` setting as a global setting for all series, but right now we rely on the first of the series, we should fix change this at some point (this is related to other open issues we have about colors and other chart config).

But yeah this PR should fix the `new series` issue without causing many side effects.

ZeRego (approved):

owlas (approved):

ZeRego (changes requested):
I think the fix should be in `getExpectedSeriesMap` packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
where we generate the new series. In the same way we set the default `defaultCartesianType` 

That way the config panel will be more consistent ( even though is not reliable ) The overall configs should all be a dropdown so they can display "mixed" and not a checkbox. But that change can be in a different PR.

rephus:
> After looking more into the code I think the fix should be in getExpectedSeriesMap

Done :+1:

ZeRego (approved):

Issue #5342, opened 2023-05-03
Add endpoint for listing spaces in project


owlas:
### Description

Alternative endpoint to list spaces without all the content and views in it.

Usage in frontend for `useSpaces`

- `SavedChartsHeader/index.tsx`
- `BrowseMenu.tsx`
- `AddTilesToDashboard.tsx`

PR #5354, opened 2023-05-03
feat: add faster spaces endpoint

owlas:
Closes: #5342

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -50,6 +50,7 @@ import {
     RequestMethod,
     ResultRow,
     SessionUser,
+    SpaceSummary,
     SummaryExplore,
     TablesConfiguration,
     TableSelectionType,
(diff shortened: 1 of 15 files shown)


rephus:
It seems some e2e tests are failing :thinking:

packages/frontend/src/hooks/useSpaces.ts
```diff
@@ -26,6 +32,25 @@ export const useSpaces = (
...
+        body: undefined,
+    });
+};
+
+export const useSpaceSummaries = (
+    projectUuid: string,
+    queryOptions?: UseQueryOptions<SpaceSummary[], ApiError>,
+) => {
+    return useQuery<SpaceSummary[], ApiError>(
+        ['projects', projectUuid, 'spaces'],
```
rephus, line 48: do we need to refetch spaces if queryOptions changes ?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1750,4 +1751,28 @@ export class ProjectService {
...
+            user.ability.cannot(
+                'view',
+                subject('Project', { organizationUuid, projectUuid }),
+            )
+        ) {
+            throw new ForbiddenError();
+        }
+
+        const spaces = await this.spaceModel.find({ projectUuid });
+        const allowedSpaces = spaces.filter(
```
rephus, line 1770: Should we also check CASL `view space` ability ?

rephus (approved):
LGTM, a couple of questions . 

Once e2e tests are fixed, you can merge

packages/frontend/src/hooks/useSpaces.ts
owlas: Ah yes we should!

owlas: I'm not sure actually, the only time it is used is to change the `onSuccess` - so potentially it could be useful. Are you saying we should manually trigger a refetch if it happens? I think right now it won't because the data is already cached

packages/backend/src/services/ProjectService/ProjectService.ts
owlas: This is the logic that _should_ be in the CASL ability but it isn't now, my plan is to refactor and centralise all the logic back in CASL.

Issue #5329, opened 2023-05-02
Migrate personal access token list to use mantine components


joaoviana:
See parent issue in #5316

PR #5358, opened 2023-05-03
feat: migrate access tokens list to mantine table

joaoviana:
Closes: #5329 

### Description:

Change PATs List view to a `<Table>`
Create styles hook `useTableStyles` to be shared across the app. This hook also allows overrides with CSS selectors+styling

NOTE: this work does not include `Delete token` modal layout

before: 
image
after:
image

Resource List component still looks as it did before:
image

### DIFF
--- a/packages/frontend/src/components/UserSettings/AccessTokensPanel/AccessTokens.styles.ts
+++ b/packages/frontend/src/components/UserSettings/AccessTokensPanel/AccessTokens.styles.ts
@@ -1,6 +1,0 @@
-import { H5 } from '@blueprintjs/core';
-import styled from 'styled-components';
-
-export const PanelTitle = styled(H5)`
-    margin: 0;
-`;
--- a/packages/frontend/src/components/UserSettings/AccessTokensPanel/AccessTokens.styles.ts
+++ b/packages/frontend/src/components/UserSettings/AccessTokensPanel/AccessTokens.styles.ts
@@ -1,42 +1,5 @@
-import { Card, H5, Tag } from '@blueprintjs/core';
+import { H5 } from '@blueprintjs/core';
 import styled from 'styled-components';
-
-export const HeaderActions = styled.div`
-    display: flex;
-    justify-content: space-between;
-    align-items: center;
-    margin-bottom: 20px;
-`;
-
-export const AccessTokenWrapper = styled(Card)`
-    display: flex;
-    flex-direction: column;
-    margin-bottom: 1.25em;
-    width: 100%;
-`;
-
-export const ItemContent = styled.div`
-    display: flex;
-    align-items: center;
-    justify-content: space-between;
-`;
-
-export const AccessTokenInfo = styled.div`
-    margin: 0;
-    flex: 1;
-    display: flex;
-    gap: 3px;
-    flex-direction: column;
-`;
-
-export const AccessTokenLabel = styled.b`
-    margin: 0;
-    margin-right: 0.625em;
-`;
-
-export const ExpireAtLabel = styled(Tag)`
-    width: fit-content;
-`;
 
 export const PanelTitle = styled(H5)`
     margin: 0;
--- a/packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
+++ b/packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
@@ -1,0 +1,124 @@
+import { Button, Classes, Dialog } from '@blueprintjs/core';
+import { ApiPersonalAccessTokenResponse, formatDate } from '@lightdash/common';
+import {
+    Button as MantineButton,
+    createStyles,
+    Paper,
+    Table,
+} from '@mantine/core';
+import { FC, useState } from 'react';
+import {
+    useAccessToken,
+    useDeleteAccessToken,
+} from '../../../hooks/useAccessToken';
+
+const useTableStyles = createStyles((theme) => ({
+    root: {
+        '& thead tr': {
+            backgroundColor: theme.colors.gray[0],
+        },
+
+        '& thead tr th': {
+            color: theme.colors.gray[6],
+            fontWeight: 600,
+        },
+
+        '& thead tr th, & tbody tr td': {
+            padding: '12px 20px',
+        },
+
+        '& tr td:last-child': {
+            textAlign: 'right',
+        },
+
+        '&[data-hover] tbody tr': theme.fn.hover({
+            cursor: 'pointer',
+            backgroundColor: theme.fn.rgba(theme.colors.gray[0], 0.5),
+        }),
+    },
+}));
+
+const TokenItem: FC<{
+    token: ApiPersonalAccessTokenResponse;
+}> = ({ token }) => {
+    const { description, expiresAt, uuid } = token;
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    const { mutate, isLoading: isDeleting } = useDeleteAccessToken();
+
+    return (
+        <>
+            <tr>
+                <td>{description}</td>
+                <td>
+                    {expiresAt ? formatDate(expiresAt) : 'No expiration date'}
+                </td>
+                <td>
+                    <MantineButton
+                        variant="outline"
+                        size="xs"
+                        color="red"
+                        onClick={() => setIsDeleteDialogOpen(true)}
+                    >
+                        Delete
+                    </MantineButton>
+                </td>
+            </tr>
+            <Dialog
+                isOpen={isDeleteDialogOpen}
+                icon="delete"
+                onClose={() => !isDeleting && setIsDeleteDialogOpen(false)}
+                title={`Delete token ${description}`}
+                lazy
+            >
+                <div className={Classes.DIALOG_BODY}>
+                    <p>
+                        Are you sure ? This will permanently delete the
+                        <b> {description} </b> token.
+                    </p>
+                </div>
+                <div className={Classes.DIALOG_FOOTER}>
+                    <div className={Classes.DIALOG_FOOTER_ACTIONS}>
+                        <Button
+                            disabled={isDeleting}
+                            onClick={() => setIsDeleteDialogOpen(false)}
+                        >
+                            Cancel
+                        </Button>
+                        <Button
+                            disabled={isDeleting}
+                            intent="danger"
+                            onClick={() => {
+                                mutate(uuid || '');
+                            }}
+                        >
+                            Delete
+                        </Button>
+                    </div>
+                </div>
+            </Dialog>
+        </>
+    );
+};
+
+export const TokensTable = () => {
+    const { data } = useAccessToken();
+    const { classes } = useTableStyles();
+    return (
+        <Paper withBorder>
+            <Table className={classes.root} highlightOnHover>
+                <thead>
+                    <tr>
+                        <th>Name</th>
+                        <th>Expiration date</th>
+                        <th></th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {data?.map((token) => (
+                        <TokenItem key={token.uuid} token={token} />
+                    ))}
+                </tbody>
+            </Table>
+        </Paper>
+    );
+};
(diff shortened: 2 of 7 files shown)


packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+        },
+
+        '& tr td:last-child': {
+            textAlign: 'right',
+        },
+
+        '&[data-hover] tbody tr': theme.fn.hover({
+            cursor: 'pointer',
+            backgroundColor: theme.fn.rgba(theme.colors.gray[0], 0.5),
+        }),
```
joaoviana, line 37: this is the same done in `packages/frontend/src/components/common/ResourceView/ResourceViewList/index.tsx`

with the exception of font-size not changing, and a text align right to the last td of a tr.

joaoviana:
- [ ] fix border issue

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+                        variant="outline"
+                        size="xs"
+                        color="red"
+                        onClick={() => setIsDeleteDialogOpen(true)}
+                    >
+                        Delete
+                    </MantineButton>
+                </td>
+            </tr>
+            <Dialog
```
IrakliJani, line 36: is it possible to remove modal rendering for every single item or can it be a 1 modal for all access tokens but open with a proper data inside a modal?

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+                        <th>Name</th>
+                        <th>Expiration date</th>
+                        <th></th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {data?.map((token) => (
+                        <TokenItem key={token.uuid} token={token} />
+                    ))}
+                </tbody>
```
IrakliJani, line 94: I guess this needs an EmptyState if there are no tokens

packages/frontend/src/components/UserSettings/AccessTokensPanel/index.tsx
```diff
@@ -145,12 +34,25 @@ const AccessTokensPanel: FC = () => {
...
+    return (
+        <>
+            <Flex justify="space-between" align="center" mb="lg">
+                <Title order={5}>Personal access tokens</Title>
+                <Button onClick={() => setCreateInvitesPanel(true)}>
+                    Generate new token
+                </Button>
+            </Flex>
+
+            <TokensTable />
```
IrakliJani, line 54: I think modal should be rendered only once outside of the table (coming up in the next PR I guess)

IrakliJani (commented):
looks great. left some comments. mostly questions

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -0,0 +1,98 @@
+import { Button, Classes, Dialog } from '@blueprintjs/core';
```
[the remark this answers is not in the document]
joaoviana: exactly @IrakliJani, dialog/modal on delete will be refactored next!

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
joaoviana: Thanks, just did it! 👨‍🍳

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
joaoviana: Going to resolve this !

packages/frontend/src/components/UserSettings/AccessTokensPanel/index.tsx
```diff
@@ -109,30 +19,9 @@ const AccessTokensPanel: FC = () => {
...
-            {hasAvailableTokens ? (
-                <div>
-                    {data?.map((token) => (
-                        <TokenListItem key={token.uuid} token={token} />
-                    ))}
-                </div>
-            ) : (
+    if (!hasAvailableTokens) {
+        return (
+            <Flex h="100%" direction="column" justify="center" align="center">
```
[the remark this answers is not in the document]
joaoviana: Thanks for reminding me about this, I have removed the Flex and looks great! 

image

packages/frontend/src/components/UserSettings/AccessTokensPanel/index.tsx
joaoviana: Great point, I put it inside `TokensTable` but on the same level as `Paper`. what do you think? Or should I move the modal to this view instead? I would prefer inside `TokensTable`

packages/frontend/src/hooks/styles/useTableStyles.tsx
```diff
@@ -0,0 +1,39 @@
...
+                if (key in defaultStyles) {
+                    defaultStyles[key] = {
+                        ...defaultStyles[key],
+                        ...overrideStyles[key],
+                    };
+                } else {
+                    defaultStyles[key] = overrideStyles[key];
+                }
+            });
+        }
```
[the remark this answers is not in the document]
joaoviana: That was my first approach actually, but if I spread `overrideStyles` inside `defaultStyles` it will not keep the content of a key that already exists. For example, if I follow that approach, for a `'& thead tr th'` and I want to override `fontWeight` and keep `color` intact, then that spreading will not extend the object and override the property I'm sending, it will override the whole key value. Hope this makes sense. 

Maybe there's a more elegant way to do this, but not sure tbh.

packages/frontend/src/components/UserSettings/AccessTokensPanel/AccessTokens.styles.ts
```diff
@@ -1,43 +1,6 @@
...
-export const AccessTokenLabel = styled.b`
-    margin: 0;
-    margin-right: 0.625em;
-`;
-
-export const ExpireAtLabel = styled(Tag)`
-    width: fit-content;
-`;
-
 export const PanelTitle = styled(H5)`
```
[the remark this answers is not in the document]
joaoviana: PanelTitle is being used somewhere else... Let me check!

packages/frontend/src/components/UserSettings/ProjectManagementPanel/ProjectManagementPanel.styles.ts
```diff
@@ -44,3 +44,7 @@ export const ProjectTag = styled(Tag)`
     width: fit-content;
     margin-top: 0.3em;
 `;
+
+export const PanelTitle = styled(H5)`
+    margin: 0;
+`;
```
joaoviana, line 50: Figured that moving this wouldn't hurt. Kept it as it is so that I avoid checking that ProjectManagementPanel looks ok if I replaced it with mantine's `Title` instead @IrakliJani

packages/frontend/src/components/UserSettings/AccessTokensPanel/index.tsx
IrakliJani: sounds good

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    return (
+        <>
+            <tr>
+                <td>{description}</td>
+                <td>
+                    {expiresAt ? formatDate(expiresAt) : 'No expiration date'}
+                </td>
+                <td>
+                    <MantineButton
```
IrakliJani, line 27: can it just be Button? I don't think you also export blueprint button in this file

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+export const TokensTable = () => {
+    const { data } = useAccessToken();
+
+    const { classes } = useTableStyles({
+        '& tr td:last-child': {
+            textAlign: 'right',
+        },
+    })();
```
IrakliJani, line 48: ok I was thinking and have something in mind. in createStyles you can have as many classes as you want. you can create one specifically for `textAlign: 'right'` and you can call it `alignRight`

IrakliJani: and you can use that on whatever td you like in code

packages/frontend/src/components/UserSettings/AccessTokensPanel/index.tsx
```diff
@@ -109,48 +19,38 @@ const AccessTokensPanel: FC = () => {
...
+                <Button onClick={() => setCreateInvitesPanel(true)}>
+                    Generate token
+                </Button>
+            </EmptyState>
+        );
+    }
+
+    return (
+        <>
+            <Flex justify="space-between" align="center" mb="lg">
```
IrakliJani, line 45: you can use `Group position="apart"` and you don't need align anymore.

also, not a big fan of applying margins to one element to have a gap between two separate sections.

I would wrap the whole content inside a <Stack spacing='lg'>

packages/frontend/src/components/common/ResourceView/ResourceViewList/index.tsx
```diff
@@ -95,7 +72,11 @@ const ResourceViewList: FC<ResourceViewListProps> = ({
     defaultSort,
     onAction,
 }) => {
-    const { classes } = useTableStyles();
+    const { classes } = useTableStyles({
+        '& thead tr th': {
+            fontSize: '12px',
+        },
+    })();
```
IrakliJani, line 79: you can have a class name for that. smallHeaderText or something

packages/frontend/src/hooks/styles/useTableStyles.tsx
```diff
@@ -0,0 +1,39 @@
...
+                padding: '12px 20px',
+            },
+
+            '&[data-hover] tbody tr': theme.fn.hover({
+                cursor: 'pointer',
+                backgroundColor: theme.fn.rgba(theme.colors.gray[0], 0.5),
+            }),
+        };
+
+        if (overrideStyles) {
```
IrakliJani, line 25: if you keep adding classes you don't need to override styles anymore. either adding classes or props (prop example in Page component)

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
joaoviana: Not yet, this PR: https://github.com/lightdash/lightdash/pull/5371 will remove the blueprint reference 😉

packages/frontend/src/components/UserSettings/AccessTokensPanel/index.tsx
joaoviana: that's great, thank you! Wrapped these 2 with a Stack, and applied mb to the Stack!

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
joaoviana: That's a great alternative, going to try that right now 👍

packages/frontend/src/hooks/styles/useTableStyles.tsx
joaoviana: I went with the approach of adding the classes to the hook, let me know what you think :)

packages/frontend/src/components/common/ResourceView/ResourceViewList/index.tsx
```diff
@@ -352,7 +329,10 @@ const ResourceViewList: FC<ResourceViewListProps> = ({
     }, [items, columnSorts, columns]);
 
     return (
-        <Table className={classes.root} highlightOnHover>
+        <Table
+            className={`${classes.root} ${classes.smallHeaderText}`}
```
IrakliJani, line 333: `useTableStyles` also exports `cx` utility that you can use

`className={cx(classes.root, classes.smallHeaderText)}`

IrakliJani (approved):
good stuff! one minor suggestion. good to ship 🚢

PR #5362, opened 2023-05-04
fix: broken big number chart

rephus:
Closes: #5357

### Description:
 

Fix big number chart when `resultsData?.rows?.[0]?.[selectedField]` is undefined. 

Before: we were getting a blank screen

Now: 

image


IrakliJani (approved):
good one.
one day we should enforce array lookups to be undefined | result

joaoviana (approved):

Issue #5319, opened 2023-05-02
Migrate `Settings > Org Settings > Appearance` from Blueprint to Mantine


IrakliJani:
### Description

- [ ] use new Mantine color input https://mantine.dev/core/color-input/

<img width="1437" alt="CleanShot 2023-05-02 at 17 37 09@2x" src="https://user-images.githubusercontent.com/962095/235683054-8f926983-aef8-418f-89b8-1524525ae488.png">

PR #5365, opened 2023-05-04
feat: migrate appearance view with mantine

joaoviana:
Closes: #5319 

### Description:

Migrate appearance view with Mantine
* Added ColorInput with swatches
* Added a form + form validation
* Changed from 3 rows to 4 rows 

image

Added swatches that are the first 8 `echarts`' default colors
image

If hex code is not in correct format (#XXX or #XXXXXX) we display error message
image


packages/frontend/src/components/UserSettings/AppearancePanel/index.tsx
```diff
@@ -1,29 +1,58 @@
...
-    let [colors, setColors] = useState<string[]>(
-        data?.chartColors || ECHARTS_DEFAULT_COLORS.slice(0, 8),
-    );
+    const IS_HEX_CODE_COLOR_REGEX = /^#([a-fA-F0-9]{3}|[a-fA-F0-9]{6})$/;
+
+    const getColorFields = (colors: string[]) =>
+        colors.reduce(
+            (acc, color, index) => ({ ...acc, [`color${index}`]: color }),
+            {},
+        );
```
joaoviana, line 28: generates `color0` to `color7` fields, unless we want to increment the `index` by 1 to match what the user sees. what do we think? 🤔

yasmin-e: match what the user sees is better imo

yasmin-e (commented):
Looks great! I would just change a few tiny things:

- The CTA when I save changes sounds a bit odd, I would change it to something more descriptive like "Success! Your default chart colours were updated." or "Success! The default chart colours of your organization were updated." (to make sure they know this doesn't apply to them only)
Screenshot 2023-05-05 at 10 05 19

- This one is very optional, do as you please, but when I enter an invalid hex color I get the validation only when I click "save changes", I would have the error state apply as soon as I finish writing

Screenshot 2023-05-05 at 10 06 27
Screenshot 2023-05-05 at 10 06 33

packages/frontend/src/components/UserSettings/AppearancePanel/index.tsx
IrakliJani: im not entirely sure why we need this

packages/frontend/src/components/UserSettings/AppearancePanel/index.tsx
```diff
@@ -1,29 +1,58 @@
...
-} from './AppearancePanel.styles';
 
 const AppearancePanel: FC = () => {
     const ability = useAbilityContext();
     const { isLoading: isOrgLoading, data } = useOrganization();
     const updateMutation = useOrganizationUpdateMutation();
-    let [colors, setColors] = useState<string[]>(
-        data?.chartColors || ECHARTS_DEFAULT_COLORS.slice(0, 8),
-    );
+    const IS_HEX_CODE_COLOR_REGEX = /^#([a-fA-F0-9]{3}|[a-fA-F0-9]{6})$/;
```
IrakliJani, line 22: this can be out of this component function

packages/frontend/src/components/UserSettings/AppearancePanel/index.tsx
IrakliJani: (also can be outside of the component function)

packages/frontend/src/components/UserSettings/AppearancePanel/index.tsx
```diff
@@ -32,62 +61,60 @@ const AppearancePanel: FC = () => {
...
-                    text="Save changes"
-                    loading={updateMutation.isLoading}
-                    onClick={update}
-                />
-            </Can>
-        </AppearancePanelWrapper>
+        <Box>
+            <Title order={5} mb="md">
+                Default chart colors
+            </Title>
```
IrakliJani, line 77: we can simplify this
- Box → `<Stack spacing="md"`
- Title mb can be removed

IrakliJani:
@joaoviana 

can you also add inside `packages/frontend/src/mantineTheme.ts`

under `components`

```
        ColorInput: {
            styles: (theme, _params) => ({
                label: {
                    // FIXME: this is a hack to fix label position. remove after Blueprint migration is complete
                    marginBottom: theme.spacing.xxs,
                },
            }),
        },
```

we still have blueprint global css leftovers which are messing with labels. looks like we gotta add it one by one for each input type

joaoviana:
(Github was acting up and wouldn't let me quote-reply to https://github.com/lightdash/lightdash/pull/5365#pullrequestreview-1414348763 @yasmin-e )

Thanks for suggesting these 2 improvements! 

1. I feel like this should be out of scope for this PR as this is used somewhere else and it would probably imply extending the hook mutation to accept a different message. If it's ok with you I would like this to be covered when toaster x mantine work comes. 

2. Added onblur validation, good shout! 🎉

packages/frontend/src/components/UserSettings/AppearancePanel/index.tsx
joaoviana: done @yasmin-e , thanks!

> im not entirely sure why we need this

@IrakliJani, I want to set the name of the fields for the form, so it looks like this: 
```
{
color1: colors[0]
color2:  colors[1]
... 
color8: colors[7]
}
```

how would you do this? I thought about just writing it (no loops or wtv) but not sure what to do here or if I'm just missing something 😞

Issue #2852, opened 2022-08-03
Enable sourcemap files


ZeRego:
### Description

We don't generate source maps for the frontend package because Blueprint typing throws an error. 

We are disabling the source map generation:
```
        "start": "GENERATE_SOURCEMAP=false react-scripts start",
        "build": "GENERATE_SOURCEMAP=false react-scripts build",
```

[Blueprint open issue](https://github.com/palantir/blueprint/issues/5274)

### What problem does this solve?

Help debug errors in production.

ZeRego:
We are now using Vite so we can enable the source maps. We just need to optimise the chunks that are generated so it doesn't run out of memory. 
We can also reduce the bundle size by improving how we import libraries `import {Button} from 'example'` vs `import Button from 'example/Button'` but that depends on how the libraries are structured.

PR #5373, opened 2023-05-05
feat: add source maps

rephus:
Closes: #2852

### Description:

Now we have optimized/removed some of blueprint and we use vite, we can generate sourcemaps again

image

### DIFF
--- a/packages/frontend/vite.config.ts
+++ b/packages/frontend/vite.config.ts
@@ -35,6 +35,7 @@ export default defineConfig({
         // sourcemap: true,
         target: 'es2015',
         minify: true,
+        sourcemap: true, 
         rollupOptions: {
             output: {
                 manualChunks: mapManualChunks({


IrakliJani (commented):
build was failing 2 weeks ago. i dont think it will go through. there is an unresolved memory overflow problem with vite

joaoviana:
> bulls was failing 2 weeks ago. i dont think it will go through. there is an unresolved memory overflow problem with vite

Yea, tried it as well and couldn't find anything that would resolve it :/

rephus:
> bulls was failing 2 weeks ago. i dont think it will go through. there is an unresolved memory overflow problem with vite


You are right, it failed :( is it worth increasing the memory on github actions to build this ?  @owlas

owlas:
We can signup to a beta for larger runners if we need it. Apparently that's available to paid GitHub orgs

owlas:
@rephus do you know how much memory this consumed for you running it locally?

IrakliJani:
@owlas @rephus  I was testing it using this flag `export NODE_OPTIONS="--max-old-space-size=1024"` I think it needs around 6gb-s

rephus:
Adding memory as ENV didn't fix the issue

.github/workflows/main.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       - name: Build and push
         id: docker_build
         uses: docker/build-push-action@v2
+        env: 
+          NODE_OPTIONS: --max-old-space-size=8192
```
IrakliJani, line 43: this only fixes the github test/build
shouldn't we add the same node options to docker build?

IrakliJani:
@rephus  before this I guess?
https://github.com/lightdash/lightdash/blob/338ebbcf499cf0170ed754f23051c51723d5cf62/dockerfile#L77

rephus:
> @rephus nice! seems like gh/render/docker all succeeded the build. I was looking to the diff, is yarn.lock change a necessity?

I think it is working now with some minimal changes

IrakliJani (approved):
✨✨✨

packages/frontend/vite.config.ts
```diff
@@ -35,6 +35,7 @@ export default defineConfig({
         // sourcemap: true,
```
joaoviana, line 35: nit-picking here, could you remove this comment pls? 😅

joaoviana (approved):
🚀

joaoviana (approved):
🚀

Issue #5350, opened 2023-05-03
Performance issues when parsing all warehouse results values


ZeRego:
### Description

Some warehouse clients have a `parseCell` function to normalize some of the data values. 
This function is looping over all the rows and cells causing the server to crash when handling large amounts of data.

We could:
- run the parser in a worker
- only parse the necessary values based on the column type

### App Version

0.520.0

owlas:
We could also stream the results!

PR #5377, opened 2023-05-05
feat: stream bigquery results

rephus:
Partially closes: https://github.com/lightdash/lightdash/issues/5350

### Description:

image

image


owlas (changes requested):
I found how to get the response schema without making two queries (although it does involve multiple API calls but that's ok because it happens anyway under the hood!)

- The results schema is only available in the full API Response that comes from bigquery
- The only way to get the full API Response is to get results with `autoPaginate: false`. So if you do `getQueryResults({autoPaginate: true})` or `getQueryResultsStream()` then the client will automatically hide the full API response. That's because under the hood it's actually making lots of API calls paginating through the rows.
- So we need to get the full response and then get the rows

```typescript
// start the job
const [job] = await client.createQueryJob({ ... });

// Get the full api response but we can request zero rows
const [rows, _, apiResponse] = await job.getQueryResults({
  autoPaginate: false, // v. important, without this we wouldn't get the apiResponse object
  maxApiCalls: 1, // only allow one api call - not sure how essential this is
  maxResults: 0, // don't fetch any results
});

console.log(apiResponse.schema) // let's party

// now get the full rows
const rows = await job.getQueryResults() // or getQueryResultsStream
```

I'm about 90% sure that worst case this will request the first page (however big that is) of results twice. Best case it doesn't fetch any results twice, just the results metadata. But I don't know exactly what raw api calls are happening under the hood with this library

rephus:
Updated bigquery to use jobs

owlas (approved):
merge once tests pass

Issue #5121, opened 2023-04-18
Bad rendering of tables on dashboards


rephus:
### Description

See video here https://lightdash-community.slack.com/archives/C03MG2VTR08/p1681769386075439?thread_ts=1681769380.931989&cid=C03MG2VTR08

ZeRego:
### Context:

We use the library [@tanstack/react-virtual](https://tanstack.com/virtual/v3 ) to virtualize the table rows and make the page performant. 

This means that the browser is only rendering the visible rows instead of all of them. This makes a great impact when there are thousands of results.

The library is constantly calculating what rows should be visible. That takes into account the visible area, the scroll position, and the size of the rows. 

### Problem:

Our table headers don't have a fixed width. Their width adapts depending on the cell's values and that impacts row height as well.
Since we only render the visible rows the cell's values are always different causing the column's width to also change.

This might cause an edge case where: 
1. virtualize library tries to render a new row at the bottom since there is space
2. table columns change width
3. table rows change height
4. virtualize library notices that the last row is not visible anymore and removes it 
5. table columns change width again
6. table rows change height again
7. ...back to point 1



https://user-images.githubusercontent.com/9117144/233420327-6e8b7aa9-ba27-4dac-b3da-260e078b8b94.mp4

hamzahc1:
Can we do something like set a max row height (and scroll if text goes beyond that rather than wrap?)

ZeRego:
> Can we do something like set a max row height (and scroll if text goes beyond that rather than wrap?)

That could reduce slightly the occurrence but wouldn't solve the problem. 

Fixed column widths seem to be the only solution to remove this edge case. 
We could let editors change the column width when editing a table visualization and save the width values.

Example: https://react-table-v7.tanstack.com/docs/examples/column-resizing

ZeRego:
Tried to reduce the number of rows that we preload from 25 to 1 but we still have the same problem. Also thought about increasing the preload to 100 but wouldn't work either. There only needs to be 1 row with big text to cause the issue so it doesn't matter the number of preload rows. We won't be able to stabilize the column width.  

https://user-images.githubusercontent.com/9117144/233689683-fc28c667-cbc4-4273-906a-62c6df3ccd77.mp4

Tested columns with fixed width but also got weird behavior from the library while scrolling. 

https://user-images.githubusercontent.com/9117144/233690131-699a98b6-8d50-4595-96c8-af437f49142a.mp4

[...]

Issue #2396, opened 2022-06-10
Make numbers monospaced in results table + table viz


owlas:
### Description

Numbers are much harder to read in a column if they aren't monospaced

ZeRego:
I’m in favour of changing all the results to a monospaced font, not just numbers.

owlas:
Any benefit to strings being monospaced though? Readability would (I think?) be worse. But maybe the table would look very inconsistent using two fonts

JohnRomanski:
> Numbers are much harder to read in a column if they aren't monospaced

This seems subjective to me. Are you referring to the result tables? What about the table visualization?

Could you allow this to be user-selectable? themeable? (We have people that spend a lot of time in Excel to get our branded fonts right for presentations.)

TuringLovesDeathMetal:
@PriPatel any preference for a monospaced font here?

PriPatel:
@ZeRego 
My proposal from the UI redesign work from the last couple of months is to use the font `Inter`.
There is some info here about a tabular version:
https://rsms.me/inter/#features/tnum

What are your thoughts? Would this work?

ZeRego:
@PriPatel would that be enabled for the entire website or just in specific places?

PriPatel:
@ZeRego That would be enabled for all results/viz table including letters if needed. (and perhaps any other areas that require this).

PriPatel:
Closing as we've made decisions on the font and [implemented this on pivot tables.](https://github.com/lightdash/lightdash/pull/5960#event-9568739560)  We will be using Inter ( it has a tabular version). We can open extra tickets to roll out this plus other frontend pivot table changes to other table types in the app.

PR #5388, opened 2023-05-05
feat: truncate large text in table cell

ZeRego:
Closes: #5121 

### Description:

To avoid issues with virtualization, where the rows have different heights, we now truncate large texts.
We consider large text if there are more than 35 chars. ( an arbitrary number that is usually close to 300px )
We set those cells to have 300px width and expand them on hover or when the cell is selected.

Preview:

May-05-2023 16-49-17

Preview in PR: 
https://lightdash-pr-5388.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/saved/1db7e2c3-8f73-489d-a145-d30ff0043b1c/view
Note that our demo data doesn't have large text so I had to use a table calculation (table calculation values are always aligned to the right).

### Cons:

- Words longer that 300px won't be fully visible.

Captura de ecrã 2023-05-05, às 15 40 22

- We force the cell to be 300px when the text is longer than 35 chars, even if the text isn't truncated. This problem exists because [we don't use a monospaces font in the tables](https://github.com/lightdash/lightdash/issues/2396).

Captura de ecrã 2023-05-05, às 15 58 03

- Didn't add support for such cases in pivot tables with metrics as rows. Should we support it ?

Captura de ecrã 2023-05-05, às 16 19 59

### DIFF
--- a/packages/frontend/src/components/Explorer/ResultsCard/CellContextMenu.tsx
+++ b/packages/frontend/src/components/Explorer/ResultsCard/CellContextMenu.tsx
@@ -42,7 +42,7 @@ const CellContextMenu: FC<
     const fieldValues = mapValues(cell.row.original, (v) => v?.value) || {};
 
     return (
-        <Menu>
+        <Menu style={{ maxWidth: 500 }}>
             {!!value.raw && isField(item) && (
                 <UrlMenuItems
                     urls={item.urls}
--- a/packages/frontend/src/components/common/Table/BodyCell.tsx
+++ b/packages/frontend/src/components/common/Table/BodyCell.tsx
@@ -20,6 +20,7 @@ interface CommonBodyCellProps {
     fontColor?: string;
     copying?: boolean;
     selected?: boolean;
+    isLargeText?: boolean;
     tooltipContent?: string;
     minimal?: boolean;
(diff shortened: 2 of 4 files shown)


packages/frontend/src/components/Explorer/ResultsCard/CellContextMenu.tsx
```diff
@@ -42,7 +42,7 @@ const CellContextMenu: FC<
     const fieldValues = mapValues(cell.row.original, (v) => v?.value) || {};
 
     return (
-        <Menu>
+        <Menu style={{ maxWidth: 500 }}>
```
ZeRego, line 45: Make sure the cell context menu isn't massive with large texts.

IrakliJani:
> Didn't add support for such cases in pivot tables with metrics as rows. Should we support it ?

@ZeRego we can do it separately in another PR. there are a lot of stuff missing from the new pivot table. it's not mature enough.. no virtualization and stuff like that

IrakliJani (approved):
yeah I think this change is good to go. we can think more how we want to improve tables v2 (i.e. fixed, resizeable columns and stuff like that) including using monospaced or not

Issue #5201, opened 2023-04-26
Copying a list of IDs into filters makes it lowercase 


rephus:
### Description

In some cases, they are trying to copy and paste a large number of ids for items into a filter (e.g. G123, C234, H345, I7658). However, when they actually paste this into Lightdash, this appears lowercase within the filter textbox (e.g. g123, c234, h345, i7658 ). Therefore, if they're using the is equal to filter then this would return No Results.

PR #5392, opened 2023-05-06
fix: Copying a list of IDs into filters makes it lowercase

muhsinkamil:
Closes:  https://github.com/lightdash/lightdash/issues/5201

### Description:

- Remove lowercase conversion while merge and do lowercase conversion only for unique checking

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/autoCompleteUtils.ts
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/autoCompleteUtils.ts
@@ -31,10 +31,7 @@ export const mergeUniqueValues = (
     values: string[],
     newValues: string[],
 ): string[] => {
-    const normalizedValues = [
-        ...values.map(normalize),
-        ...newValues.map(normalize),
-    ];
+    const allValues = [...values, ...newValues];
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/autoCompleteUtils.ts
```diff
@@ -31,10 +31,7 @@ export const mergeUniqueValues = (
     values: string[],
     newValues: string[],
 ): string[] => {
-    const normalizedValues = [
```
ZeRego, line 34: @IrakliJani do you remember why we were lowercasing all the values? Since our filters are case-sensitive it makes sense for us to not lower case the values.

IrakliJani: @ZeRego I am not sure, now that I think of I think it's totally unnecessary.
@muhsinkamil can you remove all code that is responsible of lowercasing?

`const normalize = (item: string) => item.toLocaleLowerCase();` you can remove this and follow the ts compiler. thanks!

owlas:
@muhsinkamil closing this for now due to inactivity. If you have time to pick up the PR comments we can re-open it!

Issue #5401, opened 2023-05-09
Can't navigate to Space when browsing All Spaces


joaoviana:
### Description

Users can't navigate to space when browsing All Spaces

### Reason

Issue with anchor tag link ![image](https://user-images.githubusercontent.com/7611706/237055323-7aeeaaff-c604-457a-adc4-d8a690b5b1bb.png)

PR #5410, opened 2023-05-09
fix: get correct link for a space in all spaces grid

joaoviana:
Closes: #5401 

### Description:
When viewing All Spaces, each Link has an empty string for `projectUuid`. 
This only happens on spaces. 

Before
(see issue)

After
image

### DIFF
--- a/packages/frontend/src/components/common/ResourceView/resourceUtils.ts
+++ b/packages/frontend/src/components/common/ResourceView/resourceUtils.ts
@@ -50,7 +50,9 @@ export const getResourceUrl = (projectUuid: string, item: ResourceViewItem) => {
         case ResourceViewItemType.CHART:
             return `/projects/${projectUuid}/saved/${item.data.uuid}`;
         case ResourceViewItemType.SPACE:
-            return `/projects/${projectUuid}/spaces/${item.data.uuid}`;
+            return `/projects/${projectUuid || item.data.projectUuid}/spaces/${
+                item.data.uuid
+            }`;
         default:
             return assertUnreachable(item, `Can't get URL for ${itemType}`);
     }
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/ResourceView/resourceUtils.ts
```diff
@@ -50,7 +50,9 @@ export const getResourceUrl = (projectUuid: string, item: ResourceViewItem) => {
         case ResourceViewItemType.CHART:
             return `/projects/${projectUuid}/saved/${item.data.uuid}`;
         case ResourceViewItemType.SPACE:
-            return `/projects/${projectUuid}/spaces/${item.data.uuid}`;
+            return `/projects/${projectUuid || item.data.projectUuid}/spaces/${
+                item.data.uuid
+            }`;
```
joaoviana, line 55: in `ResourceViewGrid` we set some props to empty strings by default:
```
  pinnedItemsProps = { projectUuid: '', pinnedListUuid: '' },
```

There could be another way of solving this bug but this is a quick fix for now. Essentially we should accommodate no empty strings for space items? What do we think?

IrakliJani: @yasmin-e was this a breaking change from the backend changes? maybe it should be fixed in the backend?

ZeRego: Looks like a FE issue where the projectUuid property is not being passed to this component. Usually, the project uuid is part of the url.

joaoviana: Absolutely @ZeRego, as you've mentioned it was a props issue. 

Have done this instead, what do we think? https://github.com/lightdash/lightdash/pull/5410/commits/dd26f9fe00a07d41b24100d5b87b52bfc04fbb2a

(cc @IrakliJani , @yasmin-e )

PR #5412, opened 2023-05-09
fix: undefined item on chart click

ZeRego:
Closes: #5356 

### Description:

Haven't been able to reproduce ( it might be an edge case where a column was removed from the query or yml ) but the error report points to this line of code. The code assumes the `selectedFields` is not empty and selectedField can't be undefined, which is not true.

### DIFF
--- a/packages/frontend/src/components/MetricQueryData/MetricQueryDataProvider.tsx
+++ b/packages/frontend/src/components/MetricQueryData/MetricQueryDataProvider.tsx
@@ -69,11 +69,15 @@ export const getDataFromChartClick = (
         (item) => !isDimension(item),
     );
 
-    const selectedField =
-        selectedMetricsAndTableCalculations.length > 0
-            ? selectedMetricsAndTableCalculations[0]
-            : selectedFields[0];
-    const selectedValue = e.data[getItemId(selectedField)];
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/MetricQueryData/MetricQueryDataProvider.tsx
```diff
@@ -69,11 +69,15 @@ export const getDataFromChartClick = (
...
+    if (selectedMetricsAndTableCalculations.length > 0) {
+        selectedField = selectedMetricsAndTableCalculations[0];
+    } else if (selectedFields.length > 0) {
+        selectedField = selectedFields[0];
+    }
+    const selectedValue = selectedField
+        ? e.data[getItemId(selectedField)]
+        : undefined;
     const fieldValues: Record<string, ResultValue> = Object.entries(
         e.data as Record<string, any>,
```
IrakliJani, line 82: the check here <string, any> is making the whole thing unreliable including the return value. im wondering if returning `value: { raw: selectedValue, formatted: selectedValue },` with undefined is going to break something elsewhere. it's probably worth it to replace  `Record<string, any> `with  `Record<string, ResultValue>`
and make sure returned type satisfies places where `getDataFromChartClick` is being called atm

owlas:
Rockstar!

Issue #5411, opened 2023-05-09
dimensions/metrics pane now seem to truncate names if greater than or equal to 50 characters


IrakliJani:
### Description

dimensions/metrics pane now seem to truncate if greater than or equal to 50 characters

PR #5413, opened 2023-05-09
fix: overflow with mantine tooltips in explore panel

IrakliJani:
Closes: #5411 

### Description:

CleanShot 2023-05-09 at 16 58 50@2x

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
@@ -83,7 +83,7 @@ const TreeSingleNode: FC<Props> = ({ node }) => {
                         multiline
                         sx={{ whiteSpace: 'normal' }}
                         disabled={!item.description}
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
```diff
@@ -110,6 +115,7 @@ const TreeSingleNode: FC<Props> = ({ node }) => {
                     {item.hidden ? (
                         <Tooltip
                             withArrow
+                            withinPortal
```
IrakliJani, line 118: I feel like those two should be a default props. thoughts?

joaoviana (approved):

packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
joaoviana: perhaps the withArrow @IrakliJani ! Don't know about withinPortal at first glance!

PR #5419, opened 2023-05-09
fix: run databrick query async

ZeRego:
Closes: 

### Description:

Set option to run the query asynchronously.

[Databricks docs](https://docs.databricks.com/dev-tools/nodejs-sql-driver.html#id15) don't provide much more information about this option


owlas (approved):

rephus:
Can we test this before we release ?

Issue #5415, opened 2023-05-09
Migrate `Settings > Social logins` to use Mantine


IrakliJani:
### Description

![image](https://github.com/lightdash/lightdash/assets/962095/2a628d39-6c4e-49c2-9839-f04782f24ff4)

Issue #4028, opened 2022-12-21
Add login with Google screen is confusing


yasmin-e:
### Description
current:
![image](https://user-images.githubusercontent.com/67699259/208912163-0d7fc731-68c7-493f-903d-beca9276e291.png)

expected:
![image](https://user-images.githubusercontent.com/67699259/208910901-51eabad4-9168-4082-b206-d862d712a0f2.png)

owlas:
This looks much better!! Great job @yasmin-e

yasmin-e:
all credit goes to @PriPatel for the beautiful and non-confusing design

PR #5421, opened 2023-05-09
feat: migrate social logins to mantine

joaoviana:
Closes: #4028
Closes: #5415

### Description:

Abstract the OpenIdentity social login buttons and reduce repeated code. 

Change the data structure the endpoint for `/identities` returns so that the UI doesn't have to do the grouping and stacking of current issuer types (e.g. grouping by `Okta`, `google`, etc)

Create enum for `OpenIdIdentityIssuerType` to improve consistency and avoid repeating the set of strings in the future.

Added `type-fest` package to use the `Entries` type so that `Object.entries()` returns typed

Screenshots

image

login form still works and looks as before:
image

same for register:
image

same for invite link: 
image


packages/backend/src/models/OpenIdIdentitiesModel.ts
```diff
@@ -84,7 +86,14 @@ export class OpenIdIdentityModel {
...
                 email: id.email,
                 createdAt: id.createdAt,
-            }));
+            }))
+            .reduce(
+                (acc, curr) => ({
+                    ...acc,
+                    [curr.issuerType]: [...(acc[curr.issuerType] || []), curr],
+                }),
+                { okta: [], google: [], oneLogin: [] },
```
joaoviana, line 95: I was also thinking of creating a issuerType enum to be used as a value for these. What do we think? 🤔

packages/frontend/src/components/common/ThirdPartySignInButton/index.tsx
```diff
@@ -42,7 +42,7 @@ const ThirdPartySignInButtonBase: FC<
                     : ''
             }`}
             leftIcon={
-                <Image width={20} src={logo} alt={`${providerName} logo}`} />
+                <Image width={16} src={logo} alt={`${providerName} logo}`} />
```
joaoviana, line 45: @IrakliJani any other way I could give a value to an image `width`? 'sm'/'md' don't work :/

IrakliJani: can you use mantine's Image here?

joaoviana: This is mantine's `Image`...

joaoviana: Or you mean that I probably shouldn't use it?

IrakliJani: oops sorry, thought it was a styled.image. can you try `w` instead of width?

packages/frontend/src/components/UserSettings/SocialLoginsPanel/index.tsx
```diff
@@ -1,87 +1,100 @@
...
-            </GoogleButtonWrapper>
-            <GoogleButtonWrapper>
-                <OneLoginLoginButton />
-            </GoogleButtonWrapper>
-        </div>
+        <Stack spacing="md">
+            {health &&
+                (
+                    Object.entries(userSocialLogins ?? {}) as Entries<
+                        typeof userSocialLogins
```
joaoviana, line 45: We could probably do the following in a `entries.d.ts`

```ts
declare global {   
  interface ObjectConstructor {    
     entries<T extends object>(o: T): Entries<T>  
    }
 }
```

but don't know how others  feel with overriding this stuff 😅

packages/backend/src/models/OpenIdIdentitiesModel.ts
```diff
@@ -72,19 +73,37 @@ export class OpenIdIdentityModel {
...
         const identities = await this.database('openid_identities').where(
             'user_id',
             userId,
         );
+
+        const defaultIdentities = Object.values(
+            OpenIdIdentityIssuerType,
+        ).reduce(
+            (acc, curr) => ({ ...acc, [curr]: [] }),
+            {} as Record<OpenIdIdentityIssuerType, OpenIdIdentitySummary[]>,
```
ZeRego, line 86: I think you can remove all the "forced" typing if you type the reduce `.reduce<Record<OpenIdIdentityIssuerType, OpenIdIdentitySummary[]>(...)`

joaoviana: 😮  thanks @ZeRego !

packages/frontend/package.json
```diff
@@ -102,6 +102,7 @@
         "eslint-plugin-react-hooks": "^4.6.0",
         "postcss-normalize": "10.0.1",
         "ts-unused-exports": "^9.0.4",
+        "type-fest": "^3.10.0",
```
ZeRego, line 105: Interesting package :o  didn't know about it!

packages/backend/src/models/OpenIdIdentitiesModel.ts
joaoviana: ok, was able to remove one type casting, couldn't fix the default object one :/

joaoviana: I would prefer returning all (even if some/all are empty) so we can check if they're available on the frontend - unless we want to move that check to the backend? 🤔

ZeRego (approved):
Tested with OneLogin, all working well :) good job! ⭐

packages/frontend/src/components/common/ThirdPartySignInButton/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+} from '@lightdash/common';
+import { Button, ButtonProps, Image } from '@mantine/core';
+import { FC } from 'react';
+import useToaster from '../../../hooks/toaster/useToaster';
+import { useFlashMessages } from '../../../hooks/useFlashMessages';
+import { useApp } from '../../../providers/AppProvider';
+
+type ThirdPartySignInButtonProps = {
+    inviteCode?: string;
+    intent?: 'signin' | 'add';
```
joaoviana, line 13: add `signup` intent

packages/frontend/src/components/common/ThirdPartySignInButton/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+    providerName,
+    ...props
+}) => {
+    const { health } = useApp();
+
+    switch (providerName) {
+        case OpenIdIdentityIssuerType.GOOGLE:
+            return health.data?.auth.google.oauth2ClientId ? (
+                <ThirdPartySignInButtonBase
+                    loginPath={health.data.auth.google.loginPath}
```
joaoviana, line 70: pass `inviteCode` prop

packages/frontend/src/components/UserSettings/SocialLoginsPanel/index.tsx
```diff
@@ -1,87 +1,100 @@
...
-            </GoogleButtonWrapper>
-            <GoogleButtonWrapper>
-                <OktaLoginButton />
-            </GoogleButtonWrapper>
-            <GoogleButtonWrapper>
-                <OneLoginLoginButton />
-            </GoogleButtonWrapper>
-        </div>
+        <Stack spacing="md">
+            {health &&
```
joaoviana, line 42: if `health` not defined, return null above

packages/frontend/src/pages/Invite.tsx
```diff
@@ -143,17 +140,14 @@ const Invite: FC = () => {
...
-                <OktaLoginButton inviteCode={inviteCode} />
-            )}
-            {health.data?.auth.oneLogin.enabled && (
-                <OneLoginLoginButton inviteCode={inviteCode} />
-            )}
-        </>
+        <Stack>
+            {Object.values(OpenIdIdentityIssuerType).map((providerName) => (
+                <ThirdPartySignInButton
+                    key={providerName}
```
joaoviana, line 146: add `inviteCode`

PR #5423, opened 2023-05-10
chore: mantine tooltip arrows

IrakliJani:
sets mantine's default tooltip props to have `withArrow` prop


packages/frontend/src/mantineTheme.ts
```diff
@@ -85,6 +85,9 @@ export const getMantineThemeOverride = (overrides?: {
         },
 
         Tooltip: {
+            defaultProps: {
+                withArrow: true,
+            },
```
joaoviana, line 90: 🙌

joaoviana (approved):

Issue #2045, opened 2022-05-10
I want more context on big number tiles and want to see how the stat compares to last week/month.


PriPatel:
<img width="787" alt="Screenshot 2023-03-21 at 17 15 34" src="https://user-images.githubusercontent.com/12776522/226688323-2f15e733-4dc5-42bf-b6cb-cdc049d77e8c.png">

### MVP requirements
**Big Number Chart config panel**:

- [ ] If the user has chosen a numerical field (metric, dimension, table calculation), show an additional option in the chart config panel : `Compare to previous row`. This can be toggled on and off. 
- [ ] The user should be able to choose between `raw value` and `percentage`.

**Big Number tile**:

- [ ] Show the comparison below the label. We can play around with sizing, placement, responsiveness etc.
- [ ] Icons : [arrow-up-right](https://tabler-icons.io/i/arrow-up-right) and [arrow-down-right](https://tabler-icons.io/i/arrow-down-right)
- [ ] Content:

  | Increase | Decrease | No change | Not applicable
-- | -- | -- | -- | --
% | +5% | -5% | +0% | n/a
Raw | +5 | -5 | +0 | n/a

TuringLovesDeathMetal:
> In chart config, allow the user to choose what to compare with ( is it always a timeframe?)

I think you should just always compare the one row below it (doesn't really matter if it's a timeframe or not - that's up to you). 

^An example of where I'd want this without a timeframe: comparing two experimental groups. E.g. "Group A had 100 conversions. That's 20% more than group B" (where group A would be the first row, and group B would be the second)

There should also be a way of changing the label for the comparison

JohnRomanski:
It may be easier to have the visualization always display data from two columns in the same row. You could use table calculations to determine the formatting of the second field, a net difference or a change in percentage.

TuringLovesDeathMetal:
@JohnRomanski I don't think I'd use two columns as often as two rows. 

Typically, you'd be using comparison chart types for comparing over some time series. (e.g. last week's sales vs. this week's sales). 

Your date dimension will normally the dimension in your results table, and the metric you're comparing will be increasing/decreasing from one row to the next. 

Otherwise, you'll need to create a table calculation (like you said) to create another column which looks at the next row (or the _x_th next row) to compare it to the current value. 

[...]

Issue #5055, opened 2023-04-11
Update styling of Big Number Comparison


PriPatel:
<img width="1264" alt="Screenshot 2023-04-11 at 18 20 52" src="https://user-images.githubusercontent.com/12776522/231239613-6800c1bc-5c27-4f27-9213-fa4084bacc2c.png">
Once we're happy with the functionality of the other tickets in the milestone, we can think about updating the layout and styling the tiles. 

Here are some options that Mantine provides:
https://ui.mantine.dev/category/stats

PriPatel:
Opened frontend tickets and given feedback and tweaks as the feature has been developed. There is an open discussion on using the Mantine component:
https://www.notion.so/lightdash/Using-Mantine-components-for-Big-Number-charts-e2ac1c1fde6d4b6a99b7ce3fd7c306f1
..but we will revisit it once we have more feedback and the frontend peeps are back from their holiday

PR #5429, opened 2023-05-10
feat: big number comparison

yasmin-e:
Closes: #2045 

### MVP requirements
**Big Number Chart config panel**:
- [x] The user can hide the label
- [x] If the user has chosen a numerical field (metric, dimension, table calculation), show an additional option in the chart config panel : `Compare to previous row`. This can be toggled on and off. 
- [x] The user should be able to choose between `raw value` and `percentage`.

**Big Number tile**:

- [x] Show the comparison below the label. We can play around with sizing, placement, responsiveness etc.
- [x] Icons : [arrow-up-right](https://tabler-icons.io/i/arrow-up-right) and [arrow-down-right](https://tabler-icons.io/i/arrow-down-right)
- [x] If there is no previous row to compare to, the comparison should show `n/a`
- [x] If the user has selected a table calculation that is a string, the comparison should show `n/a`

- Screenshot 2023-05-11 at 11 25 13
- Screenshot 2023-05-11 at 11 25 19
- Screenshot 2023-05-11 at 11 25 53
- Screenshot 2023-05-11 at 11 28 27
- Screenshot 2023-05-11 at 11 32 47

### DIFF
--- a/packages/frontend/src/components/BigNumberConfig/index.tsx
+++ b/packages/frontend/src/components/BigNumberConfig/index.tsx
@@ -42,7 +42,7 @@ const BigNumberConfigPanel: React.FC = () => {
             setShowComparison,
             comparisonFormat,
             setComparisonFormat,
-            formatValues,
+            comparisonFormatTypes,
         },
     } = useVisualizationContext();
     const [isOpen, setIsOpen] = useState(false);
@@ -134,17 +134,18 @@ const BigNumberConfigPanel: React.FC = () => {
                     {showComparison && (
                         <RadioGroup
                             onChange={(e) => {
+                                // @ts-ignore
                                 setComparisonFormat(e.currentTarget.value);
                             }}
                             selectedValue={comparisonFormat}
                         >
                             <Radio
                                 label="By raw value"
-                                value={formatValues.RAW}
+                                value={comparisonFormatTypes.RAW}
                             />
                             <Radio
                                 label="By percentage"
-                                value={formatValues.PERCENTAGE}
+                                value={comparisonFormatTypes.PERCENTAGE}
                             />
                         </RadioGroup>
                     )}
--- a/packages/frontend/src/components/BigNumberConfig/index.tsx
+++ b/packages/frontend/src/components/BigNumberConfig/index.tsx
@@ -134,8 +134,11 @@ const BigNumberConfigPanel: React.FC = () => {
                     {showComparison && (
                         <RadioGroup
                             onChange={(e) => {
-                                // @ts-ignore
-                                setComparisonFormat(e.currentTarget.value);
+                                setComparisonFormat(
+                                    e.currentTarget.value === 'raw'
+                                        ? comparisonFormatTypes.RAW
+                                        : comparisonFormatTypes.PERCENTAGE,
+                                );
                             }}
                             selectedValue={comparisonFormat}
                         >
--- a/packages/frontend/src/components/BigNumberConfig/index.tsx
+++ b/packages/frontend/src/components/BigNumberConfig/index.tsx
@@ -8,7 +8,12 @@ import {
     Switch,
 } from '@blueprintjs/core';
 import { Popover2 } from '@blueprintjs/popover2';
-import { CompactConfigMap, CompactOrAlias, getItemId } from '@lightdash/common';
+import {
+    CompactConfigMap,
+    CompactOrAlias,
+    ComparisonFormatTypes,
+    getItemId,
+} from '@lightdash/common';
 import { IconEye, IconEyeOff } from '@tabler/icons-react';
 import React, { useState } from 'react';
 import FieldAutoComplete from '../common/Filters/FieldAutoComplete';
@@ -42,7 +47,6 @@ const BigNumberConfigPanel: React.FC = () => {
             setShowComparison,
             comparisonFormat,
             setComparisonFormat,
-            comparisonFormatTypes,
         },
     } = useVisualizationContext();
     const [isOpen, setIsOpen] = useState(false);
@@ -136,19 +140,19 @@ const BigNumberConfigPanel: React.FC = () => {
                             onChange={(e) => {
                                 setComparisonFormat(
                                     e.currentTarget.value === 'raw'
-                                        ? comparisonFormatTypes.RAW
-                                        : comparisonFormatTypes.PERCENTAGE,
+                                        ? ComparisonFormatTypes.RAW
+                                        : ComparisonFormatTypes.PERCENTAGE,
                                 );
                             }}
                             selectedValue={comparisonFormat}
                         >
                             <Radio
                                 label="By raw value"
-                                value={comparisonFormatTypes.RAW}
+                                value={ComparisonFormatTypes.RAW}
                             />
                             <Radio
                                 label="By percentage"
-                                value={comparisonFormatTypes.PERCENTAGE}
+                                value={ComparisonFormatTypes.PERCENTAGE}
                             />
                         </RadioGroup>
                     )}
--- a/packages/frontend/src/components/BigNumberConfig/index.tsx
+++ b/packages/frontend/src/components/BigNumberConfig/index.tsx
@@ -1,6 +1,15 @@
-import { Button, FormGroup, HTMLSelect, InputGroup } from '@blueprintjs/core';
+import {
+    Button,
+    FormGroup,
+    HTMLSelect,
+    InputGroup,
+    Radio,
+    RadioGroup,
+    Switch,
+} from '@blueprintjs/core';
 import { Popover2 } from '@blueprintjs/popover2';
 import { CompactConfigMap, CompactOrAlias, getItemId } from '@lightdash/common';
+import { IconEye, IconEyeOff } from '@tabler/icons-react';
 import React, { useState } from 'react';
 import FieldAutoComplete from '../common/Filters/FieldAutoComplete';
 import { useVisualizationContext } from '../LightdashVisualization/VisualizationProvider';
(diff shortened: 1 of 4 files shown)


TuringLovesDeathMetal (changes requested):
looks good! the only thing is the comparison is the wrong way. for exampple, this should say "increase by 827" 
image

TuringLovesDeathMetal (approved):
Nice! looks good to me :)

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -148,7 +158,42 @@ const useBigNumberConfig = (
...
+    const [showComparison, setShowComparison] = useState<boolean>(false);
+    const [comparisonFormat, setComparisonFormat] = useState<string>(
+        formatValues.RAW,
+    );
+
+    const calculateComparisonValue = (a: number, b: number, format: string) => {
+        const rawValue = Math.round(a - b);
+        switch (format) {
+            case formatValues.PERCENTAGE:
+                return Math.round(((a - b) / b) * 100);
```
ZeRego, line 174: We can call `formatValue(((a - b) / b) * 100, { format: 'percent', round: 0})` to get the formatted value.

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -148,7 +158,42 @@ const useBigNumberConfig = (
...
+            case formatValues.RAW:
+                return rawValue === 0 ? 'No change' : rawValue;
+            default:
+                return rawValue;
+        }
+    };
+
+    const comparisonValue =
+        firstRowValueRaw &&
+        secondRowValueRaw &&
```
ZeRego, line 185: Be careful when checking number values like this since `0` is considered false. Checking for `isFirstRowValueNumber` and `isSecondRowValueNumber` should be enough to confirm we can compare the values.

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -148,7 +158,42 @@ const useBigNumberConfig = (
...
+    const formatValues = {
+        RAW: 'raw',
+        PERCENTAGE: 'percentage',
+    };
+    const [showComparison, setShowComparison] = useState<boolean>(false);
+    const [comparisonFormat, setComparisonFormat] = useState<string>(
+        formatValues.RAW,
+    );
+
+    const calculateComparisonValue = (a: number, b: number, format: string) => {
```
ZeRego, line 170: We can move this function outside the hook since it doesn't depend on any external values, only on it's arguments.

packages/frontend/src/components/SimpleStatistic/index.tsx
```diff
@@ -114,11 +134,36 @@ const SimpleStatistic: FC<SimpleStatisticsProps> = ({
...
+                        comparisonFormat === formatValues.PERCENTAGE
+                            ? `${comparisonValue}%`
+                            : comparisonValue}
+                        {typeof comparisonValue === 'number' &&
+                        comparisonValue > 0 ? (
+                            <IconArrowUpRight style={{ marginLeft: 5 }} />
+                        ) : typeof comparisonValue === 'number' &&
+                          comparisonValue < 0 ? (
+                            <IconArrowDownRight style={{ marginLeft: 5 }} />
+                        ) : null}
```
ZeRego, line 163: Based on the requested changes, we shouldn't rely on the `comparisonValue` type since we are going to format the value there to a string. It would also simplify this code since the business logic is all in the hook.

I propose something like this:
The hook only returns the properties:
```
comparisonValue: string; // eg: '$200'
comparisonDiffType: 'positive' | 'negative' | 'none'; // we can create a new enum for this
```
We render
```
<BigNumber ... >
   {comparisonValue}
   { comparisonDiff === 'positive' && <IconArrowUpRight ... /> }
   { comparisonDiff === 'negative' && <IconArrowDownRight ... /> }
</BigNumber>
```

ZeRego (changes requested):
This is looking good :D Got some change requests for you

ZeRego (changes requested):
This is looking good :D Got some change requests for you

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -148,7 +158,42 @@ const useBigNumberConfig = (
...
+        firstRowValueRaw &&
+        secondRowValueRaw &&
+        isSecondRowValueNumber &&
+        isFirstRowValueNumber
+            ? calculateComparisonValue(
+                  Number(firstRowValueRaw),
+                  Number(secondRowValueRaw),
+                  comparisonFormat,
+              )
+            : 'Comparison not applicable';
```
ZeRego, line 193: @PriPatel @TuringLovesDeathMetal 
In the ticket, it mentions displaying the text `n/a` when we can't make a comparison for whatever reason. Are we ok with `Comparison not applicable`?

Also, when the difference is 0 do we want to still show  `0%` or the text `No change`. I would have expected the earliest.

PriPatel: If there is no change, for **%**, show` 0%` with no icon, and for **raw value** show `No change` with no icon

TuringLovesDeathMetal: Would kind of expect `0` to appear instead of `no change` even for raw values

Let's just display N/A instead of `comparison not applicable` in the text (it feels quite wordy).

packages/frontend/src/components/BigNumberConfig/index.tsx
```diff
@@ -1,6 +1,15 @@
-import { Button, FormGroup, HTMLSelect, InputGroup } from '@blueprintjs/core';
+import {
+    Button,
+    FormGroup,
+    HTMLSelect,
+    InputGroup,
+    Radio,
+    RadioGroup,
+    Switch,
+} from '@blueprintjs/core';
```
joaoviana, line 9: What do you think will be the plan for this to be then migrated to mantine? after V1/MVP?

packages/frontend/src/hooks/useBigNumberConfig.ts
PriPatel: `No change` in all cases!

packages/frontend/src/components/BigNumberConfig/index.tsx
ZeRego: Is the last ticket in the milestone. Still being "designed" 
https://github.com/lightdash/lightdash/issues/5055

PriPatel:
I'm not able to see the config change reflected in the dashboard tiles , should I be able to yet? I'm looking at the Jaffle Shop project currently!

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -122,23 +123,32 @@ const useBigNumberConfig = (
     useEffect(() => {
         if (bigNumberConfigData?.selectedField !== undefined)
             setSelectedField(bigNumberConfigData.selectedField);
+        if (!showLabel) setBigNumberLabel('');
```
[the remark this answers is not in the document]
yasmin-e: none, I thought I'd removed it 😅

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -122,23 +123,32 @@ const useBigNumberConfig = (
...
+
+    const isFirstRowValueNumber =
         isNumericItem(item) &&
-        !(bigNumberRaw instanceof Date) &&
-        !valueIsNaN(bigNumberRaw);
+        !(firstRowValueRaw instanceof Date) &&
+        !valueIsNaN(firstRowValueRaw);
+
+    const isSecondRowValueNumber =
+        !(secondRowValueRaw instanceof Date) && !valueIsNaN(secondRowValueRaw);
```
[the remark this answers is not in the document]
yasmin-e: that always refers to the big number not the 2nd row value, and in `comparisonValue` I'm checking `isFirstRowValueNumber` so idk why I would need it here

packages/frontend/src/components/SimpleStatistic/index.tsx
yasmin-e: sleek, I was thinking of a refactor along those lines! thx 👍🏼

packages/frontend/src/hooks/useBigNumberConfig.ts
TuringLovesDeathMetal: From Slack:

- we're going to show `N/A` instead of `comparison not applicable`
- we're going to show `+0%` or `+0` for comparisons where there is no change

TuringLovesDeathMetal:
@PriPatel you can see the changes in the deployment environment here: https://lightdash-pr-5429.onrender.com/

You just click on `show environments` here, then `show deployment`

image

PriPatel:
@TuringLovesDeathMetal yep thats what I meant, sorry!🙈
The Jaffle shop project in that environment is not showing me updated dashboards there when I change the config on my big number chart.

yasmin-e:
> @TuringLovesDeathMetal yep thats what I meant, sorry!🙈 The Jaffle shop project in that environment is not showing me updated dashboards there when I change the config on my big number chart.

@PriPatel this has not been handled yet!

Updated format:
Screenshot 2023-05-11 at 11 25 13
Screenshot 2023-05-11 at 11 25 19
Screenshot 2023-05-11 at 11 25 53
Screenshot 2023-05-11 at 11 28 27
Screenshot 2023-05-11 at 11 32 47

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -171,6 +276,16 @@ const useBigNumberConfig = (
         selectedField,
         setSelectedField,
         getField,
+        comparisonValue,
+        showLabel,
+        setShowLabel,
+        showComparison,
+        setShowComparison,
+        comparisonFormatTypes: ComparisonFormatTypes,
```
ZeRego, line 284: These enums can be imported into any file as they are right now. We can remove them from here.

PR #5434, opened 2023-05-10
feat: enable new pivot table

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/4932

## TODO
- [x] column totals (case when you have not enabled metrics as rows)
- [x] replace old pivot table with the new one
- [x] remove old pivot table related code
- [x] fix pivots with only indexes or only headers
- [x] fix existing specs
- [x] add specs for "everything on columns"

https://github.com/lightdash/lightdash/assets/962095/aebe976a-ffc6-4a9d-9a4d-81d479d362a8

### header only dimensions (**without** metrics as rows):
image

### index only dimensions (**with** metrics as rows)
image

### error state
CleanShot 2023-05-11 at 23 02 26@2x


TuringLovesDeathMetal (changes requested):
Looking way better than the old pivot tables 😍 

Two changes:

1. Don't remove column totals when you only have one row. Agreed that it’s useless, but I’d actually keep this in there. The reason being: If I clicked on it and only had one row, I’d be confused why it wasn’t appearing. I think we should keep it in and if someone doesn’t like the look of it, then they can just remove the totals option from their config.
2. Should we sum 0s in totals? Yes, let’s add this - it looks weird that they’re not summed.

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -165,90 +158,78 @@ const useTableConfig = (
...
-                resultsData,
-                isColumnVisible,
-                showTableNames,
-                getFieldLabelOverride,
-                isColumnFrozen,
-            });
+            return {
+                rows: [],
+                columns: [],
+            };
```
ZeRego, line 166: Not sure why we return empty rows in this case.

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -165,90 +158,78 @@ const useTableConfig = (
...
-
-                return { data: data, error: undefined };
-            } catch (e) {
-                return { data: undefined, error: e.message };
-            }
-        } else {
-            return { data: undefined, error: undefined };
+            return { data: data, error: undefined };
+        } catch (e) {
+            console.error(e);
```
ZeRego, line 226: Reminder: remove console.log

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -347,6 +376,73 @@ const PivotTable: FC<PivotTableProps> = ({
...
+                                    total,
+                                    totalColIndex,
+                                );
+
+                                return total ? (
+                                    <TotalCell
+                                        key={`column-total-${totalRowIndex}-${totalColIndex}`}
+                                        value={value}
+                                    >
+                                        {value.formatted}
```
ZeRego, line 430: The total calculation doesn't have the same rules as before. We only show totals if:
- metric type is: number, count, count distinct, sum
- formatting is not 'percent'

This pivot table shows totals for any numeric column.

ZeRego:
Fixes:
- only show totals for correct columns/rows
- show total `0` 
- show the total when there is only 1 row ( this was already working )

Captura de ecrã 2023-05-16, às 10 43 44
Captura de ecrã 2023-05-16, às 10 41 05
Captura de ecrã 2023-05-16, às 10 40 27

TuringLovesDeathMetal (approved):
looks good!

Issue #5439, opened 2023-05-11
Remove unneeded uses of `spaces-and-content`


owlas:
This API endpoint is extremely expensive. It's currently called twice when I load a dashboard (e.g. for the menu items in the toolbar).

This should only really be needed on content pages (and eventually we even want to migrate those and delete `spaces-and-content`).

I found many usages of `spaces-and-content` that can either be deleted entirely or replaced with:

- The new `/spaces` endpoint, which just lists space metadata not the content inside
- Getting a single chart or listing charts

PR #5448, opened 2023-05-11
chore: remove unecessary uses of `spaces-and-content`

yasmin-e:
Closes: #5439 

### Description:

before
Screenshot 2023-05-11 at 15 55 01
after
Screenshot 2023-05-11 at 15 54 34


owlas (commented):
There might also be some usages of `useSpaces` / `useSpaceSummaries` that aren't actually needed. I recall at least one place where they were used just to find the `spaceName` from a `spaceUuid` when getting a chart. But the chart endpoint now includes the `spaceUuid` and `spaceName`

yasmin-e:
> There might also be some usages of `useSpaces` / `useSpaceSummaries` that aren't actually needed. I recall at least one place where they were used just to find the `spaceName` from a `spaceUuid` when getting a chart. But the chart endpoint now includes the `spaceUuid` and `spaceName`

 In order to do that we would need the `dashboard` endpoint to do the same, cause it's either places with mixed resources (both `charts` and `dashboards`) or places where we need to show `all spaces` on `project`
so for now, we can't get rid of any :/

PR #5446, opened 2023-05-11
feat: validation endpoint

rephus:
Closes: #5225

### Description:

Screenshot from 2023-05-11 13-35-17
image

image

### Acceptance criteria:
- [x] the validator is run at a project-level (i.e. it validates the content in a single project ID)
- [ ] only admins + developers can run the validator.
- [ ] The validator checks for:
  - tables - these are errors caused by fields no longer existing.
      - joins using fields that no longer exist
      - metric + dimension definitions (e.g. custom SQL) using fields that no longer exist
      - metric filters using fields that no longer exist
  - saved charts - these are errors caused by fields no longer existing _**or**_ fields being broken because of one of the errors listed above ^
      - [x] fields selected
      - [x] custom metrics 
      - [x] sorting 
      - [x] filters 
      - [ ] table calcs 
      - [ ] order of columns in the results table ( I feel this s a bit redundant , because we already filter the fields on the metricquery)
  - dashboards - these are errors caused by fields no longer existing _**or**_ charts being broken because of one of the errors listed above ^
      - [x] filters 
      - [x] charts that are broken
- [x] A single chart, dashboard, or table can have multiple errors associated to it. 
- [ ] For each error, we track information about: 
  - [x] When the validation was run
  - [x] Chart/dashboard/table name (with link to chart/dashboard/table)
  - [ ] Where it is (space)
  - [ ] who created
  - [ ] when created
  - [x] last updated by
  - [ ] last updated time
  - [ ] error information (this is potentially a long error message!)
  - [ ] share link to the error (so someone else can see the error in the table with this link)

PR #5456, opened 2023-05-11
feat: add get validation endpoint

joaoviana:
Relates to PR: #5446


packages/common/src/types/validation.ts
```diff
@@ -6,7 +6,7 @@ export type ValidationResponse = {
     projectUuid: string;
     summary: string;
     error: string;
-    lastUpdatedBy: string;
+    lastUpdatedBy?: string;
```
joaoviana, line 9: safe to make it optional if there's no chart or dashboard _attached_ to this?

packages/backend/src/models/ValidationModel/ValidationModel.ts
```diff
@@ -34,4 +48,110 @@ export class ValidationModel {
...
+                .whereNull('dashboard_uuid');
+
+        const tableValidationErrors = await Promise.all(
+            tableValidationErrorsRows.map(async (validationError) => ({
+                createdAt: validationError.created_at,
+                projectUuid: validationError.project_uuid,
+                summary: validationError.summary,
+                error: validationError.error,
+                // TODO: Check if this is correct
+                name: 'Table',
```
joaoviana, line 147: what could the name be for these 🤔

Issue #5454, opened 2023-05-11
Update styling of Comparison when the result is n/a


PriPatel:
Update the styling of the Big Number comparison when it shows `n/a` (in the case where there is no previous row to compare to or the table calculation is a string).

- [ ] If result is `n/a`, change the colour to match that of the label 

<img width="366" alt="Screenshot 2023-05-11 at 17 10 01" src="https://github.com/lightdash/lightdash/assets/12776522/a9ec3598-884d-4b0f-86d9-69df3f6c81dc">

PR #5463, opened 2023-05-12
feat: updated styling when comparison is n/a

yasmin-e:
Closes: #5454 

### Description:
Update the styling of the Big Number comparison when it shows `n/a` (in the case where there is no previous row to compare to or the table calculation is a string).

- [x] If result is `n/a`, change the colour to match that of the label 

Screenshot 2023-05-12 at 10 40 02

### DIFF
--- a/packages/frontend/src/components/SimpleStatistic/index.tsx
+++ b/packages/frontend/src/components/SimpleStatistic/index.tsx
@@ -1,3 +1,4 @@
+import { Colors } from '@blueprintjs/core';
 import { ComparisonDiffTypes } from '@lightdash/common';
 import { IconArrowDownRight, IconArrowUpRight } from '@tabler/icons-react';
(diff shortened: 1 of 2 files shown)


PriPatel (approved):

packages/frontend/src/components/SimpleStatistic/index.tsx
```diff
@@ -102,6 +103,15 @@ const SimpleStatistic: FC<SimpleStatisticsProps> = ({
         };
     }, [observerElementSize]);
 
+    const comparisonValueColor = useMemo(() => {
```
owlas, line 106: Is this necessary if both cases return the same colour?

yasmin-e: it's extended in the next PR 🙈

Issue #4884, opened 2023-03-21
I should be able to 'switch' which colour is used to represent an increase in Big Number comparisons


PriPatel:
For cases where an increase in a KPI metric is not a good thing, the user should be able to switch what 'green' represents. By default, red = bad and green = good. 
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/12776522/226694205-b089fdb7-3f65-4d07-b455-4d4ea9e7f88f.gif)

- [ ] Add colour to the comparison number and arrow.  By default, `arrow-down = red6`  and `arrow-up = green6`
- [ ] Add a toggle in the config panel: ~~`Negative colour on increase`~~ `Flip colors`
- [ ] If the user turns the toggle on, `arrow-down = green6`  and `arrow-up = red6`
- [ ] If there is no change, the label colour remains unaffected.

hamzahc1:
Would there also be a neutral colour for no change?

PriPatel:
Yes, it will remain the same (neutral colour). I'll add a point to this ticket for clarity !

PR #5464, opened 2023-05-12
feat: big number comp color switch

yasmin-e:
Closes: #4884 

### Description:

For cases where an increase in a KPI metric is not a good thing, the user should be able to switch what 'green' represents. By default, red = bad and green = good. 
- [x] Add colour to the comparison number and arrow.  By default, `arrow-down = red6`  and `arrow-up = green6`
- [x] Add a toggle in the config panel: `Flip colors`
- [x] If the user turns the toggle on, `arrow-down = green6`  and `arrow-up = red6`
- [x] If there is no change, the label colour remains unaffected.

https://github.com/lightdash/lightdash/assets/67699259/e9991323-ceac-4fd0-a939-33accb74e899

### DIFF
--- a/packages/frontend/src/components/SimpleStatistic/index.tsx
+++ b/packages/frontend/src/components/SimpleStatistic/index.tsx
@@ -65,6 +65,7 @@ const SimpleStatistic: FC<SimpleStatisticsProps> = ({
             showComparison,
             showLabel,
             comparisonDiff,
+            flipColors,
         },
         isSqlRunner,
     } = useVisualizationContext();
@@ -109,6 +110,12 @@ const SimpleStatistic: FC<SimpleStatisticsProps> = ({
                 return Colors.GRAY3;
             case ComparisonDiffTypes.UNDEFINED:
                 return Colors.GRAY3;
+            case ComparisonDiffTypes.POSITIVE:
+                return Colors.GREEN3;
+            case ComparisonDiffTypes.NEGATIVE:
+                return Colors.RED3;
+            case ComparisonDiffTypes.NONE:
+                return 'inherit';
         }
     }, [comparisonDiff]);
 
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/SimpleStatistic/index.tsx
```diff
@@ -161,7 +168,9 @@ const SimpleStatistic: FC<SimpleStatisticsProps> = ({
                             marginTop: 10,
                             display: 'flex',
                             alignItems: 'center',
-                            color: comparisonValueColor,
+                            color: `${
+                                flipColors ? comparisonValueColor : 'inherit'
```
ZeRego, line 172: should we check the flip logic in `comparisonValueColor` ? 
```
            case ComparisonDiffTypes.POSITIVE:
                return flipColors ? Colors.RED3 : Colors.GREEN3;
            case ComparisonDiffTypes.NEGATIVE:
                return flipColors ? Colors.GREEN3 : Colors.RED3;
```

yasmin-e: I completely misunderstood the requirements 🥲 will re-work this

ZeRego: eg: in this case the comparison value should be red

Captura de ecrã 2023-05-12, às 10 39 58

yasmin-e: got it!

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -203,6 +206,7 @@ const useBigNumberConfig = (
         setComparisonFormat(
             bigNumberConfigData?.comparisonFormat ?? ComparisonFormatTypes.RAW,
         );
+        setFlipColors(bigNumberConfigData?.flipColors ?? true);
```
ZeRego, line 209: Let's have `false` as default value. 

false = positive green / negative red
true = prositive red / negative green

ZeRego (approved):
🎉

Issue #5433, opened 2023-05-10
ResizeObserver loop limit exceeded


IrakliJani:
### Context

sentry error: https://sentry.io/organizations/lightdash/alerts/rules/lightdash/9333608/details/
slack thread: https://lightdash.slack.com/archives/C0234T34WEQ/p1683110496155169

PR #5472, opened 2023-05-12
fix: return if no entries on resizeObserver

joaoviana:
Closes: #5433 

### Description:

Return if no `entries` to fix `resizeObserver` loop error.

Unfortunately, I can't replicate this locally :/

Reference: https://github.com/adazzle/react-data-grid/pull/2933/files


joaoviana:
tested the following on Render: set up a scheduled delivery and checked that the image wasn't tiny, so it still works as expected.

ZeRego (approved):

joaoviana:
@IrakliJani  after our chat it feels like there's more to be done here, but in the meantime I'll just merge and assess if the return actually did something. I will keep investigating this error in Sentry and if it persists, will see other approaches 👍

Issue #4885, opened 2023-03-21
I should be able to add a label to my Big Number Comparison


PriPatel:
- [ ] Move the 'Comparison' fields to a new tab
- [ ] Add an option for the user to add a label to the comparison ( it is optional, so can be left blank)
- [ ] Show the label in the Big Number tile
<img width="848" alt="Screenshot 2023-05-10 at 11 02 19" src="https://github.com/lightdash/lightdash/assets/12776522/e173867e-9b52-446f-b68a-3392d0a0751d">

PR #5474, opened 2023-05-12
feat: big number comp label

yasmin-e:
Closes: #4885 

### Description:
- [ ] Move the 'Comparison' fields to a new tab
- [ ] Add an option for the user to add a label to the comparison ( it is optional, so can be left blank)
- [ ] Show the label in the Big Number tile

Screenshot 2023-05-12 at 14 23 32

https://github.com/lightdash/lightdash/assets/67699259/a9cffe27-cfa3-4e32-b3fa-910cfe76056a


packages/common/src/types/savedCharts.ts
```diff
@@ -35,13 +35,14 @@ export enum ComparisonDiffTypes {
 }
 
 export type BigNumber = {
-    label?: string;
```
ZeRego, line 38: changing these properties will break existing big number charts. Let's keep them as they are

yasmin-e: done!

PriPatel (commented):
Screenshot 2023-05-12 at 15 02 21
Screenshot 2023-05-12 at 15 03 30

If the `compare to previous row` toggle is off, you shouldn't see the options below it

Issue #5475, opened 2023-05-12
"refresh dbt" button sometimes fails


owlas:
### Description

The "refresh dbt" button is sometimes not doing anything (there is no toast showing the status of the compiling). But in the logs it is compiling.

Discovered it's because we poll only when the job status is `RUNNING` - but now we have a new state `STARTED` which should also be polled.

PR #5476, opened 2023-05-12
fix: refetch job data for status STARTED

owlas:
Closes: #5475 

After change: first api response can show `STARTED` and still polls

CleanShot 2023-05-12 at 14 34 02@2x

### DIFF
--- a/packages/frontend/src/hooks/useRefreshServer.tsx
+++ b/packages/frontend/src/hooks/useRefreshServer.tsx
@@ -103,7 +103,13 @@ export const useJob = (
         queryKey: ['job', jobId],
         queryFn: () => getJob(jobId || ''),
         enabled: !!jobId,
-        refetchInterval: (data) => data?.jobStatus === 'RUNNING' && 500,
+        refetchInterval: (data) =>
+            [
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useRefreshServer.tsx
```diff
@@ -103,7 +103,13 @@ export const useJob = (
         queryKey: ['job', jobId],
         queryFn: () => getJob(jobId || ''),
         enabled: !!jobId,
-        refetchInterval: (data) => data?.jobStatus === 'RUNNING' && 500,
+        refetchInterval: (data) =>
+            [
+                JobStatusType.DONE.valueOf(),
+                JobStatusType.ERROR.valueOf(),
```
ZeRego, line 109: I don't think we need `.valueOf()`

owlas: thanks! the logic was also a bit wrong. Before it would set to false if `data` was undefined but in my case it was going to `500`

ZeRego (approved):

Issue #5367, opened 2023-05-04
Remove view counts from `savedChartsModel.get`


owlas:
### Description

Counting views is expensive and we do it every time we fetch a chart with this model. We should remove `views` as a requirement in the `SavedCharts` object. 

Views are shown in two places:

1. When listing (e.g. recently updated, charts in a space, all dashboards etc.). These components all use the `spaces-and-content` endpoint, which has it's own methods in the `SpaceModel` to get a `SavedChart` - so they are unaffected by removing views from `savedChartsModel.get`
2. When viewing a single chart / dashboard, there is a view count for that single chart/dashboard in the corner. We can make a separate API call `dashboard/{id}/views` or `saved/{id}/views` rather than getting it from the GET endpoint.

---

This would make dashboard load times much faster (and many operations in the backend)

PR #5477, opened 2023-05-12
fix: get chart view stats in a separate endpoint

ZeRego:
Closes: #5367 

### Description:

Changes:
- remove view stats from the chart type
- in the chart page, get the chart view stats from a separate endpoint

Captura de ecrã 2023-05-12, às 14 43 55

### DIFF
--- a/packages/backend/src/models/AnalyticsModel.ts
+++ b/packages/backend/src/models/AnalyticsModel.ts
@@ -2,15 +2,14 @@ import {
     OrganizationMemberRole,
     UserActivity,
     UserWithCount,
+    ViewStatistics,
 } from '@lightdash/common';
+import * as Sentry from '@sentry/node';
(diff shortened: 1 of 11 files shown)


packages/backend/src/models/AnalyticsModel.ts
```diff
@@ -38,12 +37,36 @@ export class AnalyticsModel {
...
 
-        return Number(count);
+        try {
+            const stats = await this.database(AnalyticsChartViewsTableName)
+                .count({ views: '*' })
+                .min({
+                    first_viewed_at: 'timestamp',
+                })
+                .select('chart_uuid')
+                .groupBy('chart_uuid')
```
owlas, line 56: You don't need the `.select()` or `.groupBy` here. You can just have the two aggregations (count, min)

owlas (approved):

Issue #5226, opened 2023-04-27
Create a page where you can see the latest validation information


TuringLovesDeathMetal:
### Description

There's a table somewhere in the UI which presents the latest validation run. The purpose of this table is to help users get an overview of what is failing validation, why it's failing, and easily get access to the failing thing.

### Acceptance criteria:

Miro sketch: https://miro.com/app/board/uXjVMNQbbSs=/?moveToWidget=3458764553256657510&cot=14
Example error messages: https://docs.google.com/spreadsheets/d/1Vmo6NN2-nVdc7h7NAOx8w73-Wm7WM6iu7jeUtvOxanY/edit#gid=34406004

https://www.figma.com/proto/gtLzWNBwSX1ejmwz7MEm4W/Untitled?node-id=90-13593&scaling=min-zoom&page-id=77%3A24113&starting-point-node-id=79%3A9450

- [ ] There is a table in the app which lets you see the results of the latest content validation in the UI. 
- [ ] Only admins and developers get access to this table
- [ ] The table only contains information about the latest validation (not any history of validations)
- [ ] The table has details about:
  - When the validation was run
  - Chart/dashboard/table name (with link to chart/dashboard/table)
  - Where it is
  - last updated by
  - error information (this is potentially a long error message!)
  - share link to the error (so someone else can see the error int he table with this link)
- [ ] The order of the table goes:
  - order by content type: tables first, then charts, then dashboards
  - order by content name: alphabetical A-Z within each content type.
- [ ] If there is any content that developers don't have access to, then they see some indication that the entry is `private content` and they don't see any details except: when the validation was run + share link to the error (so they can share this with an admin).
- [ ] I'm able to manually trigger the validation from the UI.
- [ ] If a validation is in progress from anywhere (e.g. from a GitHub action) then the validator button appears "in progress"
- [ ] The table can scroll horizontally (if we need it to...I don't mind)
- [ ] #5504
- [ ] There is a way to dismiss a validation entry
   - [ ] dismissing an entry applies to everyone else's table as well (e.g. if I dismiss error 1, then Priyanka's table will also have error 1 dismissed)
  - [ ] If the validator is rerun and the same content is broken again, then the validation entry will reappear.

PR #5487, opened 2023-05-15
feat: add validations view

joaoviana:
Closes: #5226 

Screenshots:

image

if no errors: 

image

left todo: 
* private content verification (label that says: 'private content')
* scrolling (vertically and horizontally)

### Description:
- [x] There is a table in the app which lets you see the results of the latest content validation in the UI. 
- [x] Only admins and developers get access to this table
- [x] The table only contains information about the latest validation (not any history of validations)
- [x] The table has details about:
  - When the validation was run
  - Chart/dashboard/table name (with link to chart/dashboard/table)
  - last updated by
- [x] The order of the table goes:
  - order by content type: tables first, then charts, then dashboards
  - order by content name: alphabetical A-Z within each content type.
- [ ] If there is any content that developers don't have access to, then they see some indication that the entry is `private content` and they don't see any details except: when the validation was run + share link to the error (so they can share this with an admin).
- [x] I'm able to manually trigger the validation from the UI.
- [ ] The table can scroll horizontally (if we need it to...I don't mind)
- [ ] The table scrolls vertically if there are more than a certain number of items (let's play around and see how many this should be).
~- [ ] There is a way to dismiss a validation entry (not V1)~


TuringLovesDeathMetal (changes requested):
Weooooo! this is looking really nice 😍 

Some small things from me:

1. What do you thin about renaming the tab `content validator` instead of `validation`? There's something about `validation` that feels a bit vague to me and less like a product, more like an action (if that makes sense).
2. THe `last updated` column makes me think that it's the last time the validation error was updated for that entry. Maybe we rename it `last edited`?
3. Also, I might have missed it, but are we adding the option to `dismiss`? Or is that coming in a later version?

joaoviana:
> Weooooo! this is looking really nice 😍
> 
> Some small things from me:
> 
> 1. What do you thin about renaming the tab `content validator` instead of `validation`? There's something about `validation` that feels a bit vague to me and less like a product, more like an action (if that makes sense).
> 2. THe `last updated` column makes me think that it's the last time the validation error was updated for that entry. Maybe we rename it `last edited`?
> 3. Also, I might have missed it, but are we adding the option to `dismiss`? Or is that coming in a later version?

1. Ooooh I like it!
2. `last edited` sounds good! 
3. Option to dismiss will be added after V1 👍

packages/frontend/src/components/SettingsValidator/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+        validationError: ValidationResponse;
+    }) => {
+        if (validationError.chartUuid)
+            return <IconBox icon={IconChartBar} color="blue.8" />;
+        if (validationError.dashboardUuid)
+            return <IconBox icon={IconLayoutDashboard} color="green.8" />;
+        return <IconBox icon={IconTable} color="indigo.6" />;
+    };
+
+    // TODO: add error state
```
rephus, line 69: TODO ?

packages/frontend/src/components/SettingsValidator/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+                        position="apart"
+                        p="md"
+                        sx={{
+                            borderBottomWidth: 1,
+                            borderBottomStyle: 'solid',
+                            borderBottomColor: theme.colors.gray[3],
+                        }}
+                    >
+                        <Text fw={500} fz="xs" c="gray.6">
+                            {/* TODO: show created at timestamp on hover of formatted date instead? */}
```
rephus, line 85: TODO ?

packages/frontend/src/hooks/validation/useValidation.ts
```diff
@@ -0,0 +1,58 @@
...
+    return useMutation<ValidationResponse[], ApiError>({
+        mutationKey: ['validation', projectUuid],
+        mutationFn: () => updateValidation(projectUuid),
+        onSuccess: () => {
+            queryClient.invalidateQueries({ queryKey: ['validation'] });
+            showToastSuccess({
+                title: 'Success! Validation complete.',
+            });
+        },
+        // TODO: investigate error message
```
rephus, line 46: TODO

rephus (commented):
Code looks good. I still need to check if this validation could potentially block the API. 

I can quickly check, and if it is blocking, we can temporarily comment this tab and merge , and enable it once the blocking issue gets fixed.

packages/frontend/src/components/SettingsValidator/index.tsx
joaoviana: removing!

packages/frontend/src/hooks/validation/useValidation.ts
joaoviana: removing!

packages/frontend/src/components/SettingsValidator/index.tsx
joaoviana: need to check with @TuringLovesDeathMetal , but it's currently matching the design. Shouldn't be non-blocking for merge IMO

TuringLovesDeathMetal: I think it's fine without it , so no need for the to-do :) (sorry, my designs are like 80% instead of 100% so there's a bit of room for creativity haha)

joaoviana: thanks @TuringLovesDeathMetal ! 
good to know haha creativity: through the roof

rephus:
> > Code looks good. I still need to check if this validation could potentially block the API.
> > I can quickly check, and if it is blocking, we can temporarily comment this tab and merge , and enable it once the blocking issue gets fixed.
> 
> Cool! sounds like a plan. Let me know what I can do - will wait then for the investigation and then move on +1

The POST /validate request is blocking the API (tested on analytics) 

image

we need to move this method to the scheduler before we enable this feature

Let's comment the tab and merge

joaoviana:
> > > Code looks good. I still need to check if this validation could potentially block the API.
> > > I can quickly check, and if it is blocking, we can temporarily comment this tab and merge , and enable it once the blocking issue gets fixed.
> > 
> > 
> > Cool! sounds like a plan. Let me know what I can do - will wait then for the investigation and then move on +1
> 
> The POST /validate request is blocking the API (tested on analytics)
> 
> image
> 
> we need to move this method to the scheduler before we enable this feature
> 
> Let's comment the tab and merge

thanks for checking this @rephus , will do that very soon then 🖖

joaoviana:
done @rephus  
image
no validator tab enabled

rephus (approved):

TuringLovesDeathMetal (approved):
Nice! looks good :)

PR #5490, opened 2023-05-15
fix: use space summaries where possible in frontend

owlas:
Just noticed there are still `/spaces-and-content` queries happening on the dashboard page.

Went and cleared up more usages.


owlas:
still trying to get e2e tests to pass but there's an error with the graphile migration

PR #5492, opened 2023-05-15
fix: only pass property if first name

joaoviana:
### Description:

only pass property `lastUpdatedBy` if `first_name` is truthy

### DIFF
--- a/packages/backend/src/models/ValidationModel/ValidationModel.ts
+++ b/packages/backend/src/models/ValidationModel/ValidationModel.ts
@@ -104,7 +104,9 @@ export class ValidationModel {
                 projectUuid: validationError.project_uuid,
                 error: validationError.error,
                 name: validationError.name,
-                lastUpdatedBy: `${validationError.first_name} ${validationError.last_name}`,
+                ...(!!validationError.first_name && {
+                    lastUpdatedBy: `${validationError.first_name} ${validationError.last_name}`,
+                }),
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/ValidationModel/ValidationModel.ts
```diff
@@ -104,7 +104,9 @@ export class ValidationModel {
                 projectUuid: validationError.project_uuid,
                 error: validationError.error,
                 name: validationError.name,
-                lastUpdatedBy: `${validationError.first_name} ${validationError.last_name}`,
+                ...(!!validationError.first_name && {
```
rephus, line 107: Not sure I prefer this syntax, or 
```
lastUpdatedBy: validationError.first_name && `${validationError.first_name} ${validationError.last_name}` 
```

or even 

```
lastUpdatedBy: validationError.first_name  ? `${validationError.first_name} ${validationError.last_name}` : undefined
```

We don't need to completely remove the property, just set it to undefined.

rephus (approved):
Just 1 comment, but it look good. The property is already optional, so we don't need to make any change to the type

packages/backend/src/models/ValidationModel/ValidationModel.ts
joaoviana: sounds good, open to both, will change !

Issue #5499, opened 2023-05-15
Migrate `Organization settings > General` to use Mantine


joaoviana:
### Description

![image](https://github.com/lightdash/lightdash/assets/7611706/2e294ea1-4cb0-4a3b-9a21-ae7315d23ed6)

PR #5500, opened 2023-05-15
feat: migrate organization general settings to mantine

joaoviana:
Closes: #5499


PriPatel:
@joaoviana you made the select component work!!!  nicee

joaoviana:
Need to test it in Render as well !

joaoviana:
> Need to test it in Render as well !

tested and works fine, screen recording coming soon

packages/frontend/src/components/UserSettings/OrganizationPanel/index.tsx
```diff
@@ -1,59 +1,55 @@
...
-    const { setValue } = methods;
 
     useEffect(() => {
         if (data) {
-            setValue('name', data?.name);
+            form.setFieldValue('organizationName', data?.name);
         }
-    }, [data, setValue]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [data, data?.name]);
```
ZeRego, line 25: Why do we need to disable this rule?

packages/frontend/src/components/UserSettings/AllowedDomainsPanel/index.tsx
```diff
@@ -1,161 +1,186 @@
...
+                {!!form.values.emailDomains.length && (
                     <>
-                        <Select2
+                        <Select
                             label="Default role"
                             name="role"
                             placeholder="Organization viewer"
                             disabled={isLoading}
-                            items={roleOptions}
+                            data={roleOptions}
```
ZeRego, line 140: Something is wrong with this component. Has duplicated options. I was able to select the same project multiple times. 
Would be nice to see what projects are selected or not in the dropdown.


Captura de ecrã 2023-05-17, às 16 25 48

ZeRego (changes requested):

packages/frontend/src/components/UserSettings/AllowedDomainsPanel/index.tsx
```diff
@@ -1,161 +1,186 @@
...
+                    data={form.values.emailDomains.map((emailDomain) => ({
+                        value: emailDomain,
+                        label: emailDomain,
+                    }))}
+                    searchable
+                    creatable
+                    getCreateLabel={(query: string) => `+ Add ${query} domain`}
+                    defaultValue={form.values.emailDomains}
+                    onCreate={(value) => {
+                        if (!isValidEmailDomain(value)) {
```
ZeRego, line 107: Should we check if there is at least one `.` in the string? 

wrong value
Captura de ecrã 2023-05-17, às 16 28 18

packages/frontend/src/components/UserSettings/AllowedDomainsPanel/index.tsx
joaoviana: I think there's an issue with the Render-deployed env, look at the project list dropdown on the nav bar, also has duplicates 🤔  
image

packages/frontend/src/components/UserSettings/AllowedDomainsPanel/index.tsx
joaoviana: https://github.com/lightdash/lightdash/assets/7611706/293c2744-22e3-4b00-b4ed-53dd1034a5c3

here it is!

Issue #5502, opened 2023-05-16
Fix space permissions checks


owlas:
Some endpoints in the backend are not checking for space permissions, allowing users within a project to access private spaces.

To make the dev experience easier for enforcing permissions we should move all these checks into the `projectAbility` using CASL.

### Space permissions

Update the space model and types (`Space` and `SpaceSummary`) to always include an `access` list of uuids (this will avoid us needing multiple trips to the database and we already expose the access list for spaces to viewers).

CASL checks

```
// On project view ability
// If the space is in their project AND is public
can('view', 'Space', {
  projectUuid: member.projectUuid,
  isPrivate: false,
}
// If the space is in their project AND they are in the access list
can('view', 'Space', {
  projectUuid: member.projectUuid,
  access: user.userUuid,
}

// repeat for organization view ability
// repeat for write/delete/update permissions

// Usage in a service
if (user.ability.cannot('view', subject('Space', space)) {...}
```

### Charts and dashboards

There are two approaches:

#### Option 1: attach space access to charts and dashboards and implement CASL checks for charts

Charts/dashboards types already have a `spaceUuid` and `spaceName` property. We could also add `spaceIsPrivate` and `spaceAccess`, that way we could use the same checks for charts and dashboards as we do above.

```
// usage in service
if (user.ability.cannot('view', 'Chart', chart) {...}
```

#### Option 2: only use checks at the space level

In this case we can remove any CASL checks for charts/dashboards (since they don't have their own permissions). And do the following:
1. Get chart + space
2. Check user has space access
3. Return chart

In some cases it'll be faster to:
1. Fetch all spaces that user has access to
2. Search charts/dashboards inside those spaces 

## todo

- savedchartservice.togglepinning
- savedchartservice.updatemultiple
- savedchartservice.create
- dashboardservice.checkupdateaccess (not sure exactly how this is being used)
- projectservice.runviewchartquery

PR #5503, opened 2023-05-16
feat: fix space permission + e2e tests

rephus:
Closes: #5502

### Description:

Add permissions to these endpoints 

- [x]    savedchartservice.togglepinning
- [x]    savedchartservice.updatemultiple
- [x]    savedchartservice.create
- [x]    dashboardservice.checkupdateaccess (used on scheduler methods )
- [x]    projectservice.runviewchartquery

image


owlas (approved):
Looks good but we need to follow up with a refactor. In this code using `getFullSpace` we are counting analytics views on all charts/dashboards in the space just to check access.

On `updateMultiple` we are doing that for the whole project.

PR #5506, opened 2023-05-16
feat: add scroll if > 6 validation errors

joaoviana:
Closes: #5504

### Description:
Add vertical scrolling to table if > 6 validation errors 

#### Screen recordings

6 rows + adjusting height
https://github.com/lightdash/lightdash/assets/7611706/f907bf10-9da6-4004-952d-497682f0f78e

more than 6 rows
https://github.com/lightdash/lightdash/assets/7611706/cb6ed29c-d7d5-438c-8a2c-d65d8ebfca79

See comment below for latest screen recording: https://github.com/lightdash/lightdash/pull/5506#issuecomment-1549901980

### DIFF
--- a/packages/frontend/src/components/SettingsValidator/index.tsx
+++ b/packages/frontend/src/components/SettingsValidator/index.tsx
@@ -1,6 +1,7 @@
 import { ValidationResponse } from '@lightdash/common';
 import {
     Alert,
+    Box,
     Button,
     clsx,
     Flex,
@@ -10,6 +11,7 @@ import {
     Text,
     useMantineTheme,
 } from '@mantine/core';
+import { useElementSize } from '@mantine/hooks';
 import {
     IconAlertCircle,
     IconChartBar,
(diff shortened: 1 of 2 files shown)


ZeRego:
I expected the table header to stay visible during scroll 🤔

joaoviana:
> I expected the table header to stay visible during scroll 🤔

Sure! let me change that wrapper now! Good point @ZeRego !

packages/frontend/src/components/SettingsValidator/index.tsx
```diff
@@ -65,10 +65,29 @@ const SettingsValidator: FC<{ projectUuid: string }> = ({ projectUuid }) => {
         return <IconBox icon={IconTable} color="indigo.6" />;
     };
 
+    const [maxTableHeight, setMaxTableHeight] = useState(0);
```
ZeRego, line 68: What is the advantage of calculating the max height based on the first row height instead of a hardcoded max height? eg: 500px

joaoviana: Dynamic height for a specific number of rows I guess? 😅  That way we can change the min of 6 to whatever number and it will just work. 
Heights of rows can differ between devices so calculating that is more adaptable IMO

joaoviana: now with the `500px` you've suggested, thanks @ZeRego !

image

ZeRego (approved):

Issue #5515, opened 2023-05-17
Fix timezone issue on query filters


rephus:
### Description

> we've identified the cases where there's a month filter - the date is changing from the first day of the chosen month before saving to the last day of the month prior after saving the chart.

before:
![Captura de ecrã 2023-05-17, às 10 29 18](https://github.com/lightdash/lightdash/assets/9117144/c241a9f3-6c87-4040-88fd-d60c128b4cd7)
after saving:
![Captura de ecrã 2023-05-17, às 10 29 27](https://github.com/lightdash/lightdash/assets/9117144/38c7c471-5e21-41bf-8754-350df2841cd2)

This is happening on SAST (UTC +2)

See thread https://lightdash.slack.com/archives/C03HJ10UYQK/p1684251449643099?thread_ts=1684242056.342979&cid=C03HJ10UYQK

PR #5519, opened 2023-05-17
fix: filter timezone when saving chart

ZeRego:
Closes: #5515 

### Description:

Remove timezone from filter dates, right before making the API request.

Related to fix in this recent PR https://github.com/lightdash/lightdash/commit/cf1c4765f1fe128979bb2067148afa3dc1572fb1


rephus (approved):
:raised_hands:

Issue #5324, opened 2023-05-02
Create a notification when there are any errors caught in the validation


TuringLovesDeathMetal:
### Description

If there's a new validation that's been run and there is an error, I'm notified.

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVMNQbbSs=/?moveToWidget=3458764553256657782&cot=14

- [ ] only developers and admins see this notification
- [ ] The notification is included in the `notifications` tab from the nav bar
- [ ] The Headway updates are moved to a new icon called `what's new`
- [ ] You only get a validation notification if there are errors in your validation
- [ ] When your project is validated, you get sent a new notification if there are any errors
- [ ] There’s a link to the validation summary page from the notification
- [ ] There's the number of errors that were caught included int he notification
- [ ] The time since the validation was run is included in the notification (i've sketched it as being shown as relative time, but if that's too hard, we can show a static timestamp to the nearest minute instead)

PR #5530, opened 2023-05-18
feat: add validation notification component

joaoviana:
Closes: #5324 

### Description:

Added `use-local-storage-state` because unfortunately `mantine`'s and `react-use` weren't working properly when multiple components are watching a local storage key...  (similar here https://github.com/streamich/react-use/issues/785)

Besides that: 
`LargeMenuItem` allows `iconProps` so we can turn the icon `red` or any colour ( see demo)
Only admins and developers see validation error notifications
See `indicator`/notification badge when there's a new validation timestamp that the user hasn't read yet. 
Once user clicks on the bell, it will be marked as unread.  
If someone within the org triggered a new validation, the next time a user logs in or if their actions cause a refetch of validation results, they will see a notification badge. 

Demo: 

https://github.com/lightdash/lightdash/assets/7611706/3121e72e-9c50-48a2-b22e-d3f3271cdb50


joaoviana:
fixed disabled+loading states:

https://github.com/lightdash/lightdash/assets/7611706/59bb36bd-cc06-455b-a281-caba3a06a0de

The user won't still be able to click while validation is going on, and with nicer styling now

packages/frontend/src/hooks/validation/useValidation.ts
```diff
@@ -73,3 +102,39 @@ export const useValidationMutation = (
...
+            'manage',
+            subject('Validation', {
+                organizationUuid: project.organizationUuid,
+                projectUuid,
+            }),
+        );
+    return canUserSeeValidationErrorsNotifications;
+};
+
+export const useValidationNotificationChecker = (): [boolean, () => void] => {
```
ZeRego, line 122: The local storage is only updated when `useValidation` is triggered. That hook is only used in the validation page. We need to get the latest validation date in this hook and update the local storage.

joaoviana: Yes, understand your point, but the hook is also used in `ValidationErrorNotification.tsx` so I guess it would trigger the notification check on `NotificationsMenu/index.tsx`, no? 🤔

ZeRego: `ValidationErrorNotification ` is only rendered if `useValidationNotificationChecker` returns true.

ZeRego: my bad, was getting confused about what hook was being used to conditionally render the  `NotificationsMenu` 👍

rephus:
Should we hide the icon if there are no notifications ?

ZeRego (approved):

Issue #3245, opened 2022-09-14
Row Totals


ZeRego:
### Description

Be able to see row totals. Specifically on pivoted tables.

### User story

I have a table that has the revenue per product, monthly which is pivoted on `product`. I want to know the total revenue that we have per month (so I need the row total for each month).
<img width="1474" alt="image" src="https://user-images.githubusercontent.com/31848148/206405206-76d28245-d59b-4409-a17c-d9147f6bd9dc.png">

### Acceptance Criteria

- [ ] There is an tick-box option for `show row totals` in the Table Visualization type config: 
<img width="232" alt="image" src="https://user-images.githubusercontent.com/31848148/212281133-f3eeac95-d977-4e63-aeaa-644d0771158c.png">

- [ ] This only shows up on the table visualization chart type
- [ ] I can only `show row totals` if I have at least one group selected (i.e. the table _**must**_ be pivoted). If it's not pivoted, then the option to `show row totals` is greyed out with help text on hover (saying something like `you need to add at least one group to use row totals`)
- [ ] Row totals calculate the sum of the values that are pivoted (i.e. the values from the metrics). 
- [ ] There is one row total that appears per metric
 For example, in this table, only `count page views` would be summed. So you'd have a `Total count page views` column that would be the sum of hte count page views.
<img width="1348" alt="image" src="https://user-images.githubusercontent.com/31848148/221844129-76074eb4-41e9-464c-a8b6-d36cd33fb64d.png">
([source](https://analytics.lightdash.cloud/share/2IbtngalySkCnBI-sGyJ3))

If I added another metric for `anonymous device count`, then I would get another column called `Total anonymous device count` in my row totals
<img width="1349" alt="image" src="https://user-images.githubusercontent.com/31848148/221844866-918d5c5c-4639-4376-8cdb-00e7d7618a36.png">
([source](https://analytics.lightdash.cloud/share/2OXXs6DsiC2-kxToUxjXF))

- [ ] If I've selected `row totals` then I select `metrics as rows`, my `row totals` become `column totals` and the label in the chart config changes to reflect this (i.e. the `show row totals` becomes `show column totals`)

tallensick:
+1 for this feature please!

rephus:
I think this has been already implemented @TuringLovesDeathMetal @ZeRego are we missing anything ? 
![image](https://user-images.githubusercontent.com/1983672/208373543-fed096ab-9d46-40e2-bf66-097adbe0a75b.png)

[...]

PR #5532, opened 2023-05-18
feat: row totals

ZeRego:
Closes: #3245 

### Description:

Changes:
- add row totals option ( column totals when metrics as rows )
- take into account hidden metrics and metrics that can't have a total ( eg: average )

Row totals
Captura de ecrã 2023-05-18, às 12 24 43

Column totals ( metric as rows )
Captura de ecrã 2023-05-18, às 12 25 02

The option is available when there is a group ( the same way we have for "metric as rows" ):
Captura de ecrã 2023-05-18, às 12 28 26
Captura de ecrã 2023-05-18, às 12 28 39

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -60,6 +60,23 @@ const PivotTable: FC<PivotTableProps> = ({
     );
 
     const getRowTotalValueFromAxis = useCallback(
+        (total: unknown, colIndex: number): ResultValue | null => {
+            const value = last(data.rowTotalFields)?.[colIndex];
+
+            if (!value || !value.fieldId) throw new Error('Invalid pivot data');
+            const item = getField(value.fieldId);
+
+            const formattedValue = formatItemValue(item, total);
+
+            return {
+                raw: total,
+                formatted: formattedValue,
+            };
+        },
+        [data.rowTotalFields, getField],
+    );
+
+    const getMetricAsRowTotalValueFromAxis = useCallback(
         (total: unknown, rowIndex: number): ResultValue | null => {
             const value = last(data.indexValues[rowIndex]);
             if (!value || !value.fieldId) throw new Error('Invalid pivot data');
@@ -123,8 +140,7 @@ const PivotTable: FC<PivotTableProps> = ({
     const hasColumnTotals =
         !data.pivotConfig.metricsAsRows && data.pivotConfig.columnTotals;
 
-    const hasRowTotals =
-        data.pivotConfig.metricsAsRows && data.pivotConfig.rowTotals;
+    const hasRowTotals = data.pivotConfig.rowTotals;
 
     return (
         <Table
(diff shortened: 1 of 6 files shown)


TuringLovesDeathMetal:
Yahooo! this is looking great 😍 

Just one thing that I find tricky is I wish the totals were differentiated more clearly from other columns. Here are some ideas for potential options:

1. We could "lock" the totals on the right of the table in view (for row totals) and on the bottom of the table in view (for column totals). THis would be similar to the locked columns feature that we have
2. We could add some sort of thicker line between teh "end of the table" and the "beginning of totals"
image


What do you think?

ZeRego:
> Yahooo! this is looking great 😍
> 
> Just one thing that I find tricky is I wish the totals were differentiated more clearly from other columns. Here are some ideas for potential options:
> 
> 1. We could "lock" the totals on the right of the table in view (for row totals) and on the bottom of the table in view (for column totals). THis would be similar to the locked columns feature that we have
> 2. We could add some sort of thicker line between teh "end of the table" and the "beginning of totals"
> 
> image
> What do you think?

I agree that we should lock the total rows/columns and make them more distinguishable but we should do it in a separate ticket. That is going to take some time to implement and don't want to block this feature.

TuringLovesDeathMetal (approved):
Looks good! I added the suggested changes to the styling in the table styling issue that's later in the milestone :)

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -60,6 +60,23 @@ const PivotTable: FC<PivotTableProps> = ({
     );
 
     const getRowTotalValueFromAxis = useCallback(
+        (total: unknown, colIndex: number): ResultValue | null => {
+            const value = last(data.rowTotalFields)?.[colIndex];
+
+            if (!value || !value.fieldId) throw new Error('Invalid pivot data');
+            const item = getField(value.fieldId);
```
joaoviana, line 67: I think the return type of `ResultValue | null` doesn't match what's happening in the code. What do you think? Should it return something before the `last(...)` call like it's done on `getMetricAsRowColumnTotalValueFromAxis` ? 🤔

packages/frontend/src/hooks/pivotTable/pivotQueryResults.ts
```diff
@@ -309,39 +327,76 @@ export const pivotQueryResults = ({
...
             const N_TOTAL_COLS = indexValueTypes.length;
 
             columnTotalFields = create2DArray(N_TOTAL_ROWS, N_TOTAL_COLS);
             columnTotals = create2DArray(N_TOTAL_ROWS, N_DATA_COLUMNS);
 
-            metrics.forEach((metric, metricIndex) => {
+            summableMetricFieldIds.forEach((fieldId, metricIndex) => {
                 columnTotalFields![metricIndex][N_TOTAL_COLS - 1] = {
-                    fieldId: metric.fieldId,
+                    fieldId: fieldId,
```
joaoviana, line 378:
```suggestion
                    fieldId
```
if you want 😅

joaoviana (changes requested):
one Q!

rephus:
It looks great on metrics as rows 

image

packages/frontend/src/components/common/PivotTable/index.tsx
ZeRego: Changing the new functions type and code to not return null.

joaoviana (approved):
:shipit:

PR #5554, opened 2023-05-19
feat: move headway badge to top right

joaoviana:
Relates to: #5324 

### Description:

- Move Headway icon to top right
- Make sure both indicators/badges are matching in size and color (notifications vs headway )
- Remove some broken animations from Headway's widget badge: `bounce`, `shake`, and `wobble`

Recording: 

https://github.com/lightdash/lightdash/assets/7611706/15f6e85c-20a7-4ea0-aebc-eaca7ae6cab4

### DIFF
--- a/packages/frontend/src/components/NavBar/HeadwayMenuItem.tsx
+++ b/packages/frontend/src/components/NavBar/HeadwayMenuItem.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Tooltip } from '@mantine/core';
+import { Box, Button, Indicator, Tooltip } from '@mantine/core';
 import { IconSparkles } from '@tabler/icons-react';
 import { FC, useEffect } from 'react';
 import useHeadway from '../../hooks/thirdPartyServices/useHeadway';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/NavBar/HeadwayMenuItem.tsx
```diff
@@ -64,7 +65,7 @@ const HeadwayMenuItem: FC<Props> = ({ projectUuid }) => {
     if (!isHeadwayloaded || !projectUuid) return null;
 
     return (
-        <Tooltip color="dark" label="What's new?">
+        <Tooltip color="dark" label="What's new?" withinPortal>
```
rephus, line 68: what's this withinPortal ?

rephus (approved):
It looks good on render

packages/frontend/src/components/NavBar/HeadwayMenuItem.tsx
joaoviana: good spot, this allows the `Button.Group` wrapper to render it properly. 
See here: https://mantine.dev/core/button/#buttongroup : "If you want to use [Tooltip](https://mantine.dev/core/tooltip/), [Popover](https://mantine.dev/core/popover/) or other overlay component, it is required to set withinPortal prop:"

So this way the `Tooltip` is rendered within a Portal and we ensure the styles applied by the `Button.Group` on the parent are added correctly. Without this `prop` the button Group wouldn't style these 4 icons correctly with nice dividers and border radius on first and last button of the group: 
image

Issue #5559, opened 2023-05-21
Partitioning doesn't work with timestamp data type when using BigQuery.


UKanamo:
### Description

I am partitioning using the timestamp data type in BigQuery, but the partitions do not work when filtering. Currently, even with the timestamp data type, BigQuery's DATETIME_TRUNC is being used, but in this case, the partitions are not effective (resulting in high query costs).

<img width="1044" alt="スクリーンショット 2023-05-21 20 46 47" src="https://github.com/lightdash/lightdash/assets/52604268/0a28ad9b-0342-43ca-95a0-ec7ed36d52b2">

When using TIMESTAMP_TRUNC, the partitions work effectively (reducing query costs), so I would like to use TIMESTAMP_TRUNC when filtering with the timestamp data type.

<img width="1036" alt="スクリーンショット 2023-05-21 20 47 01" src="https://github.com/lightdash/lightdash/assets/52604268/296a1012-896a-4459-ae6c-180800190468">

### What problem does this solve?

Reducing query costs in BigQuery.

UKanamo:
👋 Hey @TuringLovesDeathMetal

Thank you for promptly addressing the issue and making the necessary modifications!
After testing on Lightdash Cloud following the release, it appears that DATETIME_TRUNC is still being used without any changes from before.
Could you please verify if the fix has been applied correctly?

UKanamo:
Hi @owlas

I am using both of them. I confirmed that self-hosting is working with version v0.568.4. Also, I verified the functionality of Lightdash Cloud about 5 hours ago. However, I just checked the version of Lightdash Cloud, and it seems to be v0.568.1 now.

UKanamo:
@owlas 
I have confirmed that it continues to use DATETIME_TRUNC without changes in LightDash Cloud v0.568.4.🤔

PR #5576, opened 2023-05-22
fix: replace truncate function for Bigquery timestamps

ZeRego:
Closes: #5559


joaoviana (approved):

owlas:
> Using TIMESTAMP_TRUNC should reduce warehouse costs.

If the user has partitioned on a timestamp!

owlas (approved):
Looks good! My main worry was that the timestamp type would change but it seems like both `TIMESTAMP_TRUNC` and `DATETIME_TRUNC` output the same type as their function argument.

For example here I pass `CURRENT_TIMESTAMP` and the output is the same no matter how I truncate it. 

CleanShot 2023-05-23 at 15 42 35@2x

PR #5577, opened 2023-05-22
feat: add validation status to spaceModel

joaoviana:
Related to: https://github.com/lightdash/lightdash/issues/5325 and a consequence of: https://github.com/lightdash/lightdash/pull/5570#issuecomment-1557352109

### Description:

* Add `validationError` property (if found) to charts and dashboards queried in the SpaceModel. 

There will be more work done after this PR on: 
* Saved Charts model
* Dashboards Model 
And wherever we might need.

PR #5582, opened 2023-05-23
feat: add validation status to dashboards

joaoviana:
Related to: #5577 

### Description:

Add validations status to dashboards.

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -48,6 +48,10 @@ import {
 } from '../../database/entities/savedCharts';
 import { getSpaceId, SpaceTableName } from '../../database/entities/spaces';
 import { UserTable, UserTableName } from '../../database/entities/users';
+import {
+    ValidationSummaryQuery,
+    ValidationTableName,
+} from '../../database/entities/validation';
 import Transaction = Knex.Transaction;
 
 export type GetDashboardQuery = Pick<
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -336,6 +347,8 @@ export class DashboardModel {
                 order,
                 views,
                 first_viewed_at,
+                validation_error,
```
rephus, line 350: This order by will give you only the last error for this dashboard_uuid, right ?

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -394,11 +413,17 @@ export class DashboardModel {
...
+                `${DashboardsTableName}.dashboard_uuid`,
+            )
             .select<
-                (GetDashboardQuery & {
-                    space_uuid: string;
-                    spaceName: string;
-                })[]
+                (GetDashboardQuery &
+                    ValidationSummaryQuery & {
+                        space_uuid: string;
```
rephus, line 424: wait, why are we mixing cases here ?

joaoviana: this is the type of ValidationSummaryQuery. Should I rename it to something clearer? 
image

```ts
export type ValidationSummaryQuery = {
    validation_error: DbValidationTable['error'];
    validation_created_at: DbValidationTable['created_at'];
};
```

joaoviana: 😱 didn't even notice, good spot, will also fix that!

PR #5593, opened 2023-05-23
feat: add validationErrors to space queries in model

joaoviana:
Re-opens: #5577 but added now all validationErrors to queries that belong in a space

### Description:

Query all validationErrors that belong to a Space Query or Dashboard.

image

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -45,8 +45,8 @@ import {
     SpaceTableName,
 } from '../database/entities/spaces';
 import { UserTableName } from '../database/entities/users';
+import { DbValidationTable } from '../database/entities/validation';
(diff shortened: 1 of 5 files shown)


packages/backend/src/models/SpaceModel.ts
```diff
@@ -448,8 +425,15 @@ export class SpaceModel {
...
-                `${ValidationTableName}.created_at as validation_created_at`,
+                this.database.raw(`
+                    COALESCE(
+                        (
+                            SELECT json_agg(validations.*) 
+                            FROM validations 
+                            WHERE validations.saved_chart_uuid = saved_queries.saved_query_uuid
+                        ), '[]'
+                    ) as validation_errors
+                `),
```
joaoviana, line 436: What do we think now? @ZeRego & @rephus 

looking good on the FE, but will make changes to FE after this PR
image

rephus (approved):

PR #5598, opened 2023-05-24
fix: Improve copy project types and remove undefined db fields

rephus:
### Description:

Error when copying content to preview project:
```cannot insert into column "space_id"```

Changes:
- remove undefined properties so it's compatible with older Postgres versions

### CODE REVIEW

File: packages/backend/src/models/ProjectModel/ProjectModel.ts [hunk 1]
```diff
@@ -929,13 +950,8 @@ export class ProjectModel {
...
                 .leftJoin('spaces', 'dashboards.space_id', 'spaces.space_id')
-                .leftJoin(
-                    'projects',
-                    'projects.project_id',
-                    'spaces.project_id',
-                )
-                .where('projects.project_uuid', projectUuid)
-                .select('dashboards.*');
+                .where('spaces.project_id', projectId)
+                .select<DbDashboard[]>('dashboards.*');
```
rephus, line 954: improve response type

ZeRego (approved):

PR #5586, opened 2023-05-23
fix: get all errors per chart and dashboard

joaoviana:
Closes: #5587

### Description:

Get all errors per chart and dashboard by removing the `distinctOn`.

PR #5604, opened 2023-05-24
fix: distinct on validation error

joaoviana:
### Description:

After merging: https://github.com/lightdash/lightdash/pull/5586, another issue came up with dashboards and charts and tables having duplicate errors.

### DIFF
--- a/packages/backend/src/models/ValidationModel/ValidationModel.ts
+++ b/packages/backend/src/models/ValidationModel/ValidationModel.ts
@@ -93,6 +93,10 @@ export class ValidationModel {
             ])
             .orderBy([
                 {
+                    column: `${ValidationTableName}.error`,
+                    order: 'asc',
+                },
+                {
                     column: `${SavedChartsTableName}.name`,
                     order: 'asc',
                 },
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/ValidationModel/ValidationModel.ts
```diff
@@ -158,6 +163,10 @@ export class ValidationModel {
                 `${SpaceTableName}.space_uuid`,
             ])
             .orderBy([
+                {
+                    column: `${ValidationTableName}.error`,
+                    order: 'asc',
+                },
```
ZeRego, line 169: This is going to ruin the order by dashboard name. And having a distinct just for the error column would remove the same error across multiple dashboards.

Changes:
- distinct on name + dashboard_id + error
- order by name, dashboard_id, error

joaoviana: got it, thanks!

joaoviana: before your requested changes ('another' dashboard didn't show up bc it shared the same error as the other one)

Screenshot 2023-05-24 at 14 59 47

now the errors for "another" appears:
Screenshot 2023-05-24 at 15 16 58

PR #5606, opened 2023-05-24
fix: duplicated tile uuids

rephus:
Closes: 

### Description:

The tile_uuids behave a bit different to other uuids, i tis important we keep the same uuid between dashboard versions, as they need to match version + tile uuid 

image


rephus:
Hopefully this is the last fix on the copy method

Issue #4879, opened 2023-03-21
Update the table visualization config to include `rows`, `columns` and `metrics`


TuringLovesDeathMetal:
### Description

Right now, the pivoting functionality isn't very cohesive. You have `pivot` for dimensions, but also `set metrics as rows`. But we should be using the same vocabulary between the two options and we can make the experience a bit nicer by adding more drag-and-drop type elements. 

We've taken a lot of inspiration from how Google Sheets does pivot tables (see here: https://docs.google.com/spreadsheets/d/1h8M8WsE2XHOFkn3_lA7LycTBtgSRV7oxxIPMUxe9RRA/edit#gid=409058531)

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVPg2a-WI=/?moveToWidget=3458764549466798507&cot=14

- [x] There are three sections in the table config: `rows`, `columns`, and `metrics`
- [x] By default, dimensions are put in the `rows`
- [x] dimensions can be dragged individually between `rows` and `columns`
- [x] moving a dimension to a `column` has the same behaviour as adding it as a `pivot` (based on the current UX for pivots in Lightdash)
- [x] you cannot drag dimensions _**within**_ rows and columns to re-order them
- [x] you can only drag dimensions individually (not as a group)
- [x] all metrics + table calculations are included in the `metrics` section
- [x] `metrics` have the option for `metrics as`: rows or columns
- [x] by default, metrics are set as `columns` 
- [x] setting metrics as `rows` is the same as doing this: #3253 
- [x] all metrics must be changed together as `rows` or `columns` (they can't be individually set)
- [x] you can't drag metrics 
- [x]  it's clear that I can drag dimensions individually between rows/columns (e.g. there's a clear draggable-looking part of the dimension label in the config)
- [x] it's clear where I can drag things to when I start dragging them (e.g. when I start dragging a dimension, the places I can put it become grey or something to indicate where I can drag it to)
- [x] there is an empty state when there's nothing in rows or columns (it should be clear that I can move dimensions into them) - maybe we can have some copy or help text if you hover over columns like: "Move dimensions to columns to pivot them in your table"

PR #5611, opened 2023-05-24
feat: table configuration with drag and drop 

ZeRego:
Closes: #4879 

### Description:

Note: Only introduced a couple of simple mantine components to avoid a big refactor and make it really different from the rest of the chart config.

Without pivot:
Captura de ecrã 2023-05-25, às 17 41 29

With pivot:
Captura de ecrã 2023-05-25, às 17 58 57

### DIFF
--- a/packages/frontend/src/components/TableConfigPanel/ColumnConfiguration.tsx
+++ b/packages/frontend/src/components/TableConfigPanel/ColumnConfiguration.tsx
@@ -9,13 +9,11 @@ import {
 } from '@tabler/icons-react';
 import React from 'react';
 import { useVisualizationContext } from '../LightdashVisualization/VisualizationProvider';
-import { ColumnConfigurationWrapper } from './ColumnConfiguration.styles';
 
-const ColumnConfiguration: React.FC = () => {
+const ColumnConfiguration: React.FC<{ fieldId: string }> = ({ fieldId }) => {
     const {
         pivotDimensions,
         tableConfig: {
-            selectedItemIds,
             updateColumnProperty,
             getFieldLabelOverride,
             getFieldLabelDefault,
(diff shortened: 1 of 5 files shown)


TuringLovesDeathMetal:
Okay, just played around with this a bit and was really confused why I couldn't find metrics as rows. 

Can we have this option appear, but greyed out/not able to interact with it? And on-hover it says `to use metrics as rows, you need to pivot your table`

Then once you have at least one dimension pivoted, it appears as usable

image

TuringLovesDeathMetal:
@ZeRego sorry, one tiny change to copy. 

Can it say:

`To use metrics as rows, you need to add a pivot.`

TuringLovesDeathMetal:
And can we remove the row totals being greyed out? 

image


I don't think we need to show them until they're available

joaoviana:
https://github.com/lightdash/lightdash/assets/7611706/83989e82-80d9-4fba-937d-0e25ef6dd60b

not sure if part of the PR scope, but I can't reorder `rows`. @ZeRego is that ok?

packages/frontend/src/components/TableConfigPanel/ColumnConfiguration.tsx
```diff
@@ -9,13 +9,11 @@ import {
 } from '@tabler/icons-react';
 import React from 'react';
 import { useVisualizationContext } from '../LightdashVisualization/VisualizationProvider';
-import { ColumnConfigurationWrapper } from './ColumnConfiguration.styles';
 
-const ColumnConfiguration: React.FC = () => {
+const ColumnConfiguration: React.FC<{ fieldId: string }> = ({ fieldId }) => {
```
joaoviana, line 13: nice

ZeRego:
> not sure if part of the PR scope, but I can't reorder rows. @ZeRego is that ok?
@joaoviana Yes, it is out of scope.

joaoviana (approved):

TuringLovesDeathMetal (approved):
Looks good to me!

Issue #5572, opened 2023-05-22
Content validator is flagging custom metrics as errors


TuringLovesDeathMetal:
### Description

The content validator is defining custom metrics as errors. In particular, it seems ok with the field itself but not with it being used in sorting

Issue #5567, opened 2023-05-22
False positive validation error when sorting using table calculation


rephus:
### Description

we get this error in analytics.lightdash, but the field does exist as a table calculation

![image](https://github.com/lightdash/lightdash/assets/1983672/a3afdef5-4d2c-49ef-b6b6-4a814e6607c6)

TuringLovesDeathMetal:
related to #5572 !

PR #5613, opened 2023-05-25
fix: fix validation sorting false positives

rephus:
Closes: #5567
Closes: #5572

### Description:

Before: 
Screenshot from 2023-05-25 10-41-07

AFter:

Screenshot from 2023-05-25 10-41-48

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.ts
@@ -187,6 +187,11 @@ export class ValidationService {
                     [],
                 );
 
+            const filterTableCalculations = (fieldId: string) =>
+                !chart.metricQuery.tableCalculations
+                    ?.map((tableCalculation) => tableCalculation.name)
+                    ?.includes(fieldId);
+
             const filterErrors = getFilterRules(
                 chart.metricQuery.filters,
             ).reduce<CreateValidation[]>(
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -200,25 +205,26 @@ export class ValidationService {
...
+                            acc,
+                            existingFieldIds,
+                            field.fieldId,
+                            `Sorting error: the field '${field.fieldId}' no longer exists`,
+                        ),
+                    [],
+                );
 
             /*
             // I think these are redundant, as we already check dimension/metrics
```
joaoviana, line 226: should we just remove all the commented code below?

joaoviana (approved):

packages/backend/src/services/ValidationService/ValidationService.ts
rephus: not for now... I feel we might want to add it anyway, and it took me a while to generate those, specially the one about table calculations. For example, the method to filter table calculations was commented out and I have just recovered it.

rephus: if you prefer, I can remove it and add a comment pointing to github, so we can easily track the code

joaoviana: Cool, keep it @rephus, as long as we clean it up closer to the milestone completion then we're good! No action then, 👍

Issue #2650, opened 2022-07-07
Support ssh tunnelling for redshift connections


owlas:
### Description

If I'm using redshift, I may have my database only accessible in a VPC and to get access I need to Lightdash to use an SSH tunnel into my VPC before trying to connect.

### In scope ? 
Lightdash *should generate the SSH keypair* and store the private key, while returning the public key for the user to copy (see comments on how to generate an openssh compatible keypair). This might add some complexity to the onboarding process / API because we'd need to generate the keypair _first_ before the user runs saves their credentials and show them the public key.

Should this be in scope for the first PR or should we ship v1 without it? 

## Example

**Update - ssh tunnel config example**

I got this working with an SSH proxy into our render instance to access a private database. We'd need to add this to the postgres warehouse client in `@lightdash/warehouses` package. We might need to move the pool creation to `runQuery` instead of the constructor, so we can `await` the tunnel creation:

```typescript
import { PoolConfig } from 'pg';

import SSH2Promise from 'ssh2-promise';
import SSHConfig from 'ssh2-promise/lib/sshConfig';

        const remoteMachineConfig: SSHConfig = {
            host: 'ssh.oregon.render.com', // host of remote machine
            port: 22, // ssh port on remote machine
            username: '...', // username for remote machine
            privateKey:'-----BEGIN OPENSSH PRIVATE KEY-----', // openssh formatted private ssh key
            passphrase: '...', // optional passphrase for an encrypted privateKey
        } as SSHConfig;

        const sshCon = new SSH2Promise(remoteMachineConfig);

        const tunnel = await sshCon.addTunnel({
            remoteAddr: 'dpg-...', // host for the database (as seen by the remote machine)
            remotePort: 5432, // port for the database (as seen by the remote machine)
        });

        const credentials: PoolConfig = {
            port: tunnel.localPort, // port on lightdash server forwarded from ssh (this will now be the db)
            host: 'localhost', // host on lightdash server with port forwarded from ssh
            // standard postgres config...
            user: '...',
            password: '...',
            database: '...',
        };
```

To support this we'll need the following configuration for postgres/redshift warehouses:
- ssh host
- ssh port
- ssh username
- ssh password (optional)
- private key (optional)
- private key passphrase (optional)

Also to update:
- [ ] Frontend forms for Postgres / redshift
- [ ] Ensure these fields are added to the`sensitiveCredentials` list so they aren't returned by the API
- [ ] 

```

owlas:
[...]

PR #5616, opened 2023-05-25
feat: add ssh keypairs api (stacked 2)

owlas:
Stacked PR2 for #2650 (for details see https://github.com/lightdash/lightdash/pull/5612)

Adds an API to generate SSH public-private keys-pairs. The public key is shown to the user (for them to install into their own server) and the private key is encrypted and cached by lightdash.


packages/backend/package.json
```diff
@@ -78,6 +79,7 @@
         "puppeteer": "^18.0.5",
         "redoc-express": "^1.0.0",
         "simple-git": "^3.16.0",
+        "sshpk": "^1.17.0",
```
ZeRego, line 82: are we missing the `yarn.lock` updates ?

ZeRego (approved):
lgtm :) is just missing the yarn.lock file

rephus (approved):

PR #5617, opened 2023-05-25
feat: add ssh tunnel (stacked 3)

owlas:
Stacked PR3 for #2650 (for details see https://github.com/lightdash/lightdash/pull/5612)

Adds `SshTunnel` and uses throughout the project service.

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -11,6 +11,7 @@ import {
     CreateJob,
     CreateProject,
     CreateProjectMember,
+    CreateWarehouseCredentials,
     DashboardAvailableFilters,
     DbtProjectType,
     deepEqual,
@@ -59,7 +60,9 @@ import {
     UpdateProject,
     UpdateProjectMember,
     WarehouseClient,
+    WarehouseTypes,
 } from '@lightdash/common';
+import { SshTunnel } from '@lightdash/warehouses';
 import * as Sentry from '@sentry/node';
 import { URL } from 'url';
 import { v4 as uuidv4 } from 'uuid';
@@ -76,6 +79,7 @@ import { OnboardingModel } from '../../models/OnboardingModel/OnboardingModel';
 import { ProjectModel } from '../../models/ProjectModel/ProjectModel';
 import { SavedChartModel } from '../../models/SavedChartModel';
 import { SpaceModel } from '../../models/SpaceModel';
+import { SshKeyPairModel } from '../../models/SshKeyPairModel';
 import { projectAdapterFromConfig } from '../../projectAdapters/projectAdapter';
 import { buildQuery } from '../../queryBuilder';
 import { compileMetricQuery } from '../../queryCompiler';
@@ -90,6 +94,7 @@ type ProjectServiceDependencies = {
     jobModel: JobModel;
     emailClient: EmailClient;
     spaceModel: SpaceModel;
+    sshKeyPairModel: SshKeyPairModel;
 };
 
 export class ProjectService {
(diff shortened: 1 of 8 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -122,32 +130,60 @@ export class ProjectService {
...
+        sshTunnel: SshTunnel<CreateWarehouseCredentials>;
+    }> {
         // Always load the latest credentials from the database
         const credentials =
             await this.projectModel.getWarehouseCredentialsForProject(
                 projectUuid,
             );
-        // Check cache for existing client
+        // Setup SSH tunnel for client (user needs to close this)
+        const sshTunnel = new SshTunnel(credentials);
```
rephus, line 167: I feel like this tunnel should be optional, so people without tunnes will not run any of this, like 

```
if (sshCredentials) {
const sshTunnel = new SshTunnel(credentials);
sshTunnel.connect();
} else {
client.connect(); 
}

File: packages/backend/src/services/ProjectService/ProjectService.ts [hunk 2]
```diff
@@ -122,32 +130,60 @@ export class ProjectService {
...
+                warehouseConnection: {
+                    ...args.warehouseConnection,
+                    sshTunnelPrivateKey: privateKey,
+                },
+            };
+        }
+        return args;
+    }
+
+    private async _getWarehouseClient(projectUuid: string): Promise<{
```
rephus, line 157: perhaps this function should become a wrapper, so the tunnel closes once the method finishes, or return a callback to close the client

packages/warehouses/src/ssh/sshTunnel.ts
```diff
@@ -0,0 +1,82 @@
...
+                break;
+            default:
+                assertUnreachable(type, new Error('Unknown warehouse type'));
+        }
+        return this.overrideCredentials;
+    };
+
+    disconnect = async (): Promise<void> => {
+        if (this.sshConnection) {
+            console.info('Closing SSH tunnel');
```
ZeRego, line 78: suggestion: we should use our `logger.debug` instead of `console.info` to match the rest of the logs.

packages/warehouses/src/ssh/sshTunnel.ts
```diff
@@ -0,0 +1,82 @@
...
+                            remotePort: this.originalCredentials.port,
+                        });
+                        console.info(
+                            `SSH tunnel ready at ${sshTunnel.localPort}`,
+                        );
+                        this.localPort = sshTunnel.localPort;
+                        this.overrideCredentials = {
+                            ...this.originalCredentials,
+                            host: '127.0.0.1',
+                            port: sshTunnel.localPort,
```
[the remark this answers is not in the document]
owlas: Every query might run on a different port, we let the OS allocate ports for us. Typically the largest port is like 64000 so we have a lot of room for allocation as long as we clean up. 

Practically I don't know what the limit is for number of concurrent open SSH Tunnel

packages/warehouses/src/ssh/sshTunnel.ts
owlas: I started with this but I don't think we make the logger available in the warehouses package? Is there a way to do that so it integrates well with the logger in the main server?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -122,32 +130,60 @@ export class ProjectService {
...
         const credentials =
             await this.projectModel.getWarehouseCredentialsForProject(
                 projectUuid,
             );
-        // Check cache for existing client
+        // Setup SSH tunnel for client (user needs to close this)
+        const sshTunnel = new SshTunnel(credentials);
+        const warehouseSshCredentials = await sshTunnel.connect();
+
+        // Check cache for existing client (always false if ssh tunnel was connected)
```
ZeRego, line 170: How is it always false if ssh tunnel connected ?

PR #5618, opened 2023-05-25
feat: add ssh tunnel to project config form

owlas:
Final stacked PR for #2650 (for details see https://github.com/lightdash/lightdash/pull/5612)

Closes: #2650 

Showing new fields - error handling on generating a key - copying a key

CleanShot 2023-06-01 at 16 35 38

### DIFF
--- a/packages/frontend/src/components/ProjectConnection/WarehouseForms/PostgresForm.tsx
+++ b/packages/frontend/src/components/ProjectConnection/WarehouseForms/PostgresForm.tsx
@@ -1,8 +1,12 @@
+import { Button } from '@blueprintjs/core';
 import { WarehouseTypes } from '@lightdash/common';
-import { Anchor } from '@mantine/core';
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/ProjectConnection/WarehouseForms/PostgresForm.tsx
```diff
@@ -181,6 +182,31 @@ const PostgresForm: FC<{
                 />
                 <Input name="warehouse.role" label="Role" disabled={disabled} />
                 <StartOfWeekSelect disabled={disabled} />
+                <BooleanSwitch
+                    name="warehouse.useSshTunnel"
+                    label="Use SSH tunnel"
+                    disabled={disabled}
+                />
+                <Input
```
rephus, line 190: we could show these inputs only if ssh tunnel is enabled

PR #5622, opened 2023-05-25
fix: get fieldId for dashboard error filter

joaoviana:
### Description:

Similar to the way we get the `fieldId` for error filters in charts, we should do the same for dashboards


packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -312,7 +312,7 @@ export class ValidationService {
                         acc,
                         existingFieldIds,
                         filter.target.fieldId,
-                        `Filter error: the field '${filter}' no longer exists`,
+                        `Filter error: the field '${filter.target.fieldId}' no longer exists`,
```
joaoviana, line 315: before
image


after:

image

Issue #5541, opened 2023-05-19
Hide custom metrics section if there are no custom metrics defined


owlas:
### Description

Only show custom metrics when at least has been defined on the model

![CleanShot 2023-05-17 at 14 02 15@2x](https://github.com/lightdash/lightdash/assets/11660098/7de78d12-5f35-4b5f-a8c2-0f6de5f50f1d)

PR #5627, opened 2023-05-25
fix: show custom metrics when at least 1 is defined

yasmin-e:
Closes: #5541 

### Description:

https://github.com/lightdash/lightdash/assets/67699259/3fcf5881-eaa6-49e7-8641-619f35617c58

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/TableTreeSections.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/TableTreeSections.tsx
@@ -34,11 +34,13 @@ const TableTreeSections: FC<Props> = ({
     }, [table.metrics]);
 
     const customMetrics = useMemo(() => {
-        return additionalMetrics.reduce<Record<string, AdditionalMetric>>(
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Explorer/ExploreTree/TableTree/TableTreeSections.tsx
```diff
@@ -34,11 +34,13 @@ const TableTreeSections: FC<Props> = ({
     }, [table.metrics]);
 
     const customMetrics = useMemo(() => {
-        return additionalMetrics.reduce<Record<string, AdditionalMetric>>(
-            (acc, item) => ({ ...acc, [getItemId(item)]: item }),
-            {},
-        );
-    }, [additionalMetrics]);
+        return additionalMetrics
+            .filter((metric) => metric.table === table.name)
```
yasmin-e, line 38: moved the filtering here because it was causing a bug where: `hasCustomMetrics` comes back `false` even when we have some defined.
Since the filtering was happening on a map of tables, 2/3 were necessarily false and it was triggering a re-render.

joaoviana (approved):

Issue #5366, opened 2023-05-04
There is no overflow or wrapping in the project dropdown


PriPatel:
<img width="225" alt="Screenshot 2023-05-04 at 14 28 29" src="https://user-images.githubusercontent.com/12776522/236217998-d2947258-8e00-4e47-98ca-9e5ec88830f9.png">

IrakliJani:
wdyt should there be a wrap or truncated text?

PriPatel:
Some of these preview names are really long so lets do truncated for now! Wrapping might look super weird.
 I suppose the dropdown could also be wider? (doesn't have to match the width of the input, right?) 

 I imagine later we want to make preview ones look different somehow etc so this can just be a quick fix! wdyt?

IrakliJani:
@PriPatel right now it has a vertical scrolling which is not ideal, I'll try to increase dropdown width and truncation as well
https://github.com/lightdash/lightdash/assets/962095/cfa16095-65b5-488b-b18a-13d9b7609413

PR #5629, opened 2023-05-25
fix: project dropdown styling

yasmin-e:
Closes: #5366 

### Description:
Increased the width to match the global search width `200px` > `250px`
Screenshot 2023-05-25 at 17 00 28


joaoviana:
do you have a before and after for this pls?

joaoviana (approved):
I think it's also worth seeing if we can remove the horizontal scrolling on longer project names, but can be done in another PR!

yasmin-e:
> I think it's also worth seeing if we can remove the horizontal scrolling on longer project names, but can be done in another PR!

@PriPatel what do you think?

Issue #5229, opened 2023-04-27
Create a command in the CLI for `lightdash validate`


TuringLovesDeathMetal:
### Description

This validates the changes in yaml in the dev branch against the content in the preview project (which has been copied from prod). 

You can't fetch the latest changes from prod in your dev branch - be sure to make this clear in the docs!

### Acceptance criteria

- [ ] running `lightdash validate` on the CLI triggers the validation endpoint 
    - [ ] I imagine we'll need to build a preview environment to run `lightdash validate` against...so we'll possible need to run `lightdash preview` before we trigger the validator here! 
- [ ] There is a loading state while the CLI is running the validation
  - something like: `[12s] validating your project` where `[12s]` is a timer that shows you how long your project has been validating for
  - Ideally there would be "stages" that are validated and this is shown on the CLI. E.g. 

```
    [12s] validating your project
    
    - validating tables [COMPLETED]
    - validating charts [COMPLETED]
    - validating dashboards [2s]
```

- [ ] there is some "completion" state which shows you the errors. 
    - [ ] It tells you the overall number of items that were validated
    - [ ] It tells you how long the validation took
    - [ ] It breaks down the validation across tables, charts and dashboards
    - [ ] It gives you a table of the errors ordered by type: (tables, then charts, then dashboards). Then alphabetically by the item name within each type (A-Z)
```
Successfully validated 150 project items in 123s, SUCCESS=130 ERRORS=20

- Tables: SUCCESS = 30, ERRORS = 0
- Charts: SUCCESS = 55, ERRORS = 12
- Dashboards: SUCCESS = 20, ERRORS = 8

Item name   |   Link   |  Error  |
-----------------------------
Katie's Chart. | app.lightdash.cloud/chart-id | blah bah something's broken error
KPI dashboard | app.lightdash.cloud/dashboard-id | blah balsdndskladjksla

Done 🕶
```

TuringLovesDeathMetal:
Idea: Imagine if there was a way to "open this validation in Lightdash" 🤯 

So it would open the "validation" overview page, but in your preview project. I just feel like interacting with these errors would be much nicer in the UI compared to on the CLI

TuringLovesDeathMetal:
Two example use cases:

1. I want to add `lightdash validate` to my CI/CD so that PRs are blocked from being merged if they fail validation. 
2. I want to be able to run `lightdash validate` while I'm developing my project so I can check to see if my changes will break anything in prod.

[...]

PR #5642, opened 2023-05-26
feat: cli validate command

rephus:
Closes: #5229

### Description:

image

Screenshot from 2023-05-26 10-49-55

New version: 

image


packages/cli/src/handlers/dbt/validate.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (validation.length === 0) {
+        spinner?.succeed(`  Validation finished without errors`);
+    } else {
+        spinner?.fail(`  Validation finished with errors\n`);
+
+        validation.forEach((v) => {
+            console.error(
+                `- ${styles.error(v.name)}: ${styles.warning(v.error)}`,
+            );
+        });
```
ZeRego, line 107: We could use a library like this one to create a table/columns https://www.npmjs.com/package/columnify

packages/backend/src/controllers/validationController.ts
```diff
@@ -34,12 +35,17 @@ export class ValidationController extends Controller {
     async post(
         @Path() projectUuid: string,
         @Request() req: express.Request,
+        @Query() context?: ValidateProjectPayload['context'],
```
ZeRego, line 38: We can use the existing header information to detect if a request comes from the CLI, APP, other. This header is available in all requests.

Example in the projectRouter
```getRequestMethod(req.header(LightdashRequestMethodHeader)),```

With the controller, it should be possible to have something like:
```
@Header(LightdashRequestMethodHeader) context?: string;
...

jobId: await validationService.validate(
                    req.user!,
                    projectUuid,
                    context: getRequestMethod(context),
                ),
```

rephus: good idea ! I'll do it this way

ZeRego (approved):

rephus:
@TuringLovesDeathMetal chart/dashboard urls don't look great on the CLI output 
image

TuringLovesDeathMetal:
oof, ya, let's keep the links out of there

TuringLovesDeathMetal:
@rephus related to this comment:

> we don't return success on validation, only errors, I can still filter by chart/dashboard/Table and show you a summary before the table if you want

Yes! That would be really helpful :) Once we implement that, I'm happy with it!

Issue #5640, opened 2023-05-26
Saved charts/queries in validator table have the icon that belongs to their type


joaoviana:
### Description

Assign the correct icon to each chart by type. 

### What problem does this solve?

Currently users see the bar chart icon for any type of chart error, which might confuse them when going through the erro

![image](https://github.com/lightdash/lightdash/assets/7611706/4781d4c4-7f8e-4807-960e-8a8cdd1a113a)

PR #5644, opened 2023-05-26
feat: add correct icon for charts in validator table

joaoviana:
Closes: #5640 

### Description:

Screenshots:

https://github.com/lightdash/lightdash/assets/7611706/c7d8b70d-caeb-4d7f-9318-4abb369fb18c

image

image

### DIFF
--- a/packages/common/src/types/validation.ts
+++ b/packages/common/src/types/validation.ts
@@ -1,8 +1,11 @@
+import type { ChartKind } from './savedCharts';
+
 export type ValidationResponse = {
     validationId: number;
     createdAt: Date;
     name: string;
     chartUuid?: string;
+    chartType?: ChartKind;
     dashboardUuid?: string;
     projectUuid: string;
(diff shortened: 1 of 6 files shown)


packages/common/src/types/validation.ts
```diff
@@ -1,8 +1,11 @@
+import type { ChartKind } from './savedCharts';
```
joaoviana, line 1: this resolves a circular dependency between this file and `packages/common/src/types/savedCharts.ts` (which imports `ValidationSummary` from `validation.ts`) so we just import the type definition. 
What do you think?
If not, what should I do here? I still want to reference `ChartKind` enum here...

yasmin-e: I think it's fine like this
@ZeRego  ??

yasmin-e (approved):

ZeRego (approved):

PR #5646, opened 2023-05-26
chore: clean projects in e2e tests

ZeRego:
### Description:

Problem:
Each time the e2e tests run we create more projects. Sometimes ending up with dozens of duplicated projects in PR environments.

We still want to keep the latest created projects so reviewers can manually test projects connected to different warehouses.

Changes:
- deletes older projects before creating new ones via UI
- delete CLI project at the end of the test

Example of 1 project being deleted before creating a new one

Captura de ecrã 2023-05-26, às 16 22 07

### DIFF
--- a/packages/cli/src/handlers/deploy.ts
+++ b/packages/cli/src/handlers/deploy.ts
@@ -21,6 +21,7 @@ import { checkLightdashVersion, lightdashApi } from './dbt/apiClient';
 import { DbtCompileOptions } from './dbt/compile';
 
 type DeployHandlerOptions = DbtCompileOptions & {
+    name?: string;
     projectDir: string;
     profilesDir: string;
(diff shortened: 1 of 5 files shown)


joaoviana (approved):

rephus (commented):
IMO I think it is useful to keep 1 project per warehouse, for manual testing. So I'd keep the delete previous project on before, and I'd remove the delete on end.

ZeRego:
> IMO I think it is useful to keep 1 project per warehouse, for manual testing. So I'd keep the delete previous project on before, and I'd remove the delete on end.

@rephus I agree we should keep one project per warehouse. The only project we don't keep is the project we create via the CLI. Are you suggesting keeping that one as well ?

Issue #4593, opened 2023-02-24
When using a `is not equal to` filter expression, we should include NULL values


TuringLovesDeathMetal:
### Description

Currently, if I have a filter: `string_filed is not equal to 'blah'`, the SQL generated is like: `string_field NOT IN('blah')`. 

This means that values that are equal to `NULL` are removed from the results. 

For example:
- if I have the `string_field`: ['blah', 'bloop', 'bark', NULL] 
- and I filter `string_field is not equal to 'blah'`
- This will return: ['bloop', 'bark']
- It will not include `NULL`

We should include `NULL` in the filter `string_field is not equal to 'blah'`, so the generated SQL for `is not equal to` for filters (numeric, string, date, etc.) should be: `NOT IN('blah') OR IS NULL`

### What problem does this solve?

Intuitively, `null` is not equal to `blah`, so it feels like it should be included (and it's not obvious that it won't be).

PR #5658, opened 2023-05-29
fix: is not equal return null values

rephus:
Closes: #4593

### Description:

My results: (I had to set to NULL some values on db for testing) 

Screenshot from 2023-05-29 16-32-27

Before the fix 

Screenshot from 2023-05-29 16-35-57

After the fix 

Screenshot from 2023-05-29 16-36-27

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.ts
+++ b/packages/common/src/compiler/filtersCompiler.ts
@@ -91,7 +91,7 @@ export const renderNumberFilterSql = (
                 ? 'true'
                 : `((${dimensionSql}) NOT IN (${filter.values.join(
                       ',',
-                  )})  OR (${dimensionSql}) IS NULL)`;
+                  )}) OR (${dimensionSql}) IS NULL)`;
         case FilterOperator.NULL:
             return `(${dimensionSql}) IS NULL`;
         case FilterOperator.NOT_NULL:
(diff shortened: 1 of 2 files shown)


yasmin-e (approved):

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -46,9 +46,9 @@ export const renderStringFilterSql = (
         case FilterOperator.NOT_EQUALS:
             return !escapedFilterValues || escapedFilterValues.length === 0
                 ? 'true'
-                : `(${dimensionSql}) NOT IN (${escapedFilterValues
+                : `((${dimensionSql}) NOT IN (${escapedFilterValues
                       .map((v) => `${stringQuoteChar}${v}${stringQuoteChar}`)
-                      .join(',')})`;
+                      .join(',')} ) OR (${dimensionSql}) IS NULL)`;
```
ZeRego, line 51: Do we need the same fix for numbers `renderNumberFilterSql` ?

rephus: Added dates and numbers 

Screenshot from 2023-05-30 12-43-14

ZeRego (approved):
lgtm 👍 

we just need to fix unit tests

Issue #5460, opened 2023-05-12
Update global search icons


IrakliJani:
### Description

global search icons are outdated and does not have a color-coding that we use on the homepage
different chart types should also have different icons like on the homepage

![CleanShot 2023-05-12 at 10 28 13@2x](https://github.com/lightdash/lightdash/assets/962095/7f7829e6-3560-41e3-b141-c797d74fce26)
![CleanShot 2023-05-12 at 10 28 39@2x](https://github.com/lightdash/lightdash/assets/962095/befe1658-b1c7-4c1d-8e1f-6e5743e7d671)

<img width="556" alt="CleanShot 2023-05-12 at 10 29 21@2x" src="https://github.com/lightdash/lightdash/assets/962095/2138bbe2-9fee-4fa7-a658-3f1bdc7cf435">

PR #5663, opened 2023-05-30
feat: update icons in global search

joaoviana:
Closes: #5460 

### Description:

Migrate old icons to new icons on global search

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/3452e71b-1caa-4eb8-865b-051b12ee8cab


packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
```diff
@@ -1,37 +1,44 @@
...
     IconAbc,
     IconAppWindow,
-    IconChartAreaLine,
     IconFolder,
     IconLayoutDashboard,
     IconTable,
 } from '@tabler/icons-react';
 import { FC } from 'react';
-import MantineIcon, { MantineIconProps } from '../../common/MantineIcon';
+import { getChartIcon, IconBox } from '../../common/ResourceView/ResourceIcon';
```
IrakliJani, line 11: I think it's time we move out ResourceIcon out from ResourceView to common/ResourceIcon and extend it with additional types field, table, page. otherwise it's getting tangled

joaoviana: sounds like a plan!

IrakliJani: @joaoviana  ideally ResourceIcon should contain all of that, since it seems like we are starting to use ResourceIcon in other places too - it will be good for consistency.
^ this can be done in the separate PR. no need to implement it here 👌

joaoviana: agree 100% @IrakliJani , I will open a PR as soon as this gets done 👍

IrakliJani (approved):
looks great 🙌

Issue #5563, opened 2023-05-22
When I create a chart or dashboard from within a space, it should default to adding it to the space I'm in


TuringLovesDeathMetal:
### Description

By default, when I create a new chart or dashboard from within a space, it should save it to the space I'm in. Instead, it saves it to the first space that was created in the project.

https://www.loom.com/share/27a17a2a176c4c40b1f8dbc483a05934

PR #5669, opened 2023-05-30
fix: show correct default space when creating resource

yasmin-e:
Closes: #5563 

### Description:

https://github.com/lightdash/lightdash/assets/67699259/51383ffa-c827-4051-9c55-b38f9b6a564f

https://github.com/lightdash/lightdash/assets/67699259/d7b63f3c-0545-40bf-85f6-895865ea54ff

### DIFF
--- a/packages/frontend/src/components/common/modal/ChartCreateModal.tsx
+++ b/packages/frontend/src/components/common/modal/ChartCreateModal.tsx
@@ -23,23 +23,22 @@ import {
 interface ChartCreateModalProps extends DialogProps {
     savedData: CreateSavedChartVersion;
     onClose?: () => void;
-    spaceUuid?: string;
+    defaultSpaceUuid?: string;
     onConfirm: (savedData: CreateSavedChartVersion) => void;
 }
 
 const ChartCreateModal: FC<ChartCreateModalProps> = ({
     savedData,
     onClose,
-    spaceUuid,
+    defaultSpaceUuid,
     ...modalProps
 }) => {
     const { projectUuid } = useParams<{ projectUuid: string }>();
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/common/modal/ChartCreateModal.tsx
```diff
@@ -23,34 +23,45 @@ import {
...
 
 const ChartCreateModal: FC<ChartCreateModalProps> = ({
     savedData,
     onClose,
+    spaceUuid,
     ...modalProps
 }) => {
     const { projectUuid } = useParams<{ projectUuid: string }>();
-    const { data: spaces } = useSpaceSummaries(projectUuid);
+    // const { data: spaces } = useSpaceSummaries(projectUuid);
```
ZeRego, line 37: forgotten comment

packages/frontend/src/components/common/modal/ChartCreateModal.tsx
```diff
@@ -23,34 +23,45 @@ import {
 interface ChartCreateModalProps extends DialogProps {
     savedData: CreateSavedChartVersion;
     onClose?: () => void;
+    spaceUuid?: string;
```
ZeRego, line 26: suggestion: I would have called this `defaultSpaceUuid` and kept the `spaceUuid` state. 

When reading the code, I know that `defaultSpaceUuid` is not a safe value to use. While `spaceUuid` sounds like a set in stone value.

yasmin-e: applied the same rule to `DashboardCreateModal.tsx` for consistency

Issue #4617, opened 2023-02-27
Charts should default to sorting by date, even if a query was already run with a metric before 


hamzahc1:
### Description

If you run a query with a metric and a date dimension selected, it works as expected and the chart is sorted by date by default. 

However if you: 
- Select a metric, then run query
- Add a date dimension and then hit run query

---> the chart doesn't sort by the date dimension anymore which can be confusing for users. It seems like maybe it does sort bar charts sometimes, but all other chart types look quite strange!

PR #5670, opened 2023-05-30
fix: default sort date chart

rephus:
Closes: https://github.com/lightdash/lightdash/issues/4617

### Description:

Only sort results for chart, if the date is selected as xAxis dimension and date is not sorted by default. 

Screenshot from 2023-05-30 13-08-17

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEcharts.ts
+++ b/packages/frontend/src/hooks/echarts/useEcharts.ts
@@ -40,7 +40,6 @@ import { useMemo } from 'react';
 import { defaultGrid } from '../../components/ChartConfigPanel/Grid';
 import { useVisualizationContext } from '../../components/LightdashVisualization/VisualizationProvider';
 import { EMPTY_X_AXIS } from '../cartesianChartConfig/useCartesianChartConfig';
-import { sortDimensions } from '../cartesianChartConfig/utils';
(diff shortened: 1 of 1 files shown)


rephus:
@ZeRego could you re-review when you have a moment ?

packages/frontend/src/hooks/echarts/useEcharts.ts
```diff
@@ -1269,6 +1270,61 @@ const useEcharts = () => {
...
+        try {
+            if (!explore) return results;
+            const dimensions = getDimensions(explore);
+
+            const xFieldId = validCartesianConfig?.layout?.xField;
+            if (xFieldId === undefined) return results;
+
+            const alreadySorted = resultsData?.metricQuery.sorts?.find(
+                (sort) => sort.fieldId === xFieldId,
+            );
```
ZeRego, line 1290: We need to check if the first sort matches the x field id.
```suggestion
            const alreadySorted = resultsData?.metricQuery.sorts?.[0].fieldId ==- xFieldId;
```


Example of data that is sorted first by name and date second:
```
jose, 2020-10-03
jose, 2020-10-04
javi, 2020-10-01 
javi, 2020-10-02 
```

ZeRego (approved):

Issue #5653, opened 2023-05-26
UI improvements - Validator table


PriPatel:
**Problems**:
<img width="678" alt="Screenshot 2023-05-26 at 19 42 04" src="https://github.com/lightdash/lightdash/assets/12776522/eadb3cff-36fa-4e8e-b792-ba7e2c451371">
- The 'last updated by' column gets pretty squished
- The error itself is quite hard to digest
- The table in general is too small to fit everything we would like- https://github.com/lightdash/lightdash/issues/5652

**Proposal**
- [ ] Move 'last updated by' under item to give more space to display full names
- [ ] Add number of views 
- [ ] Split up the Error by category 'eg. `Metric error`' `Dimension error` `Missing chart`

<img width="868" alt="Screenshot 2023-05-26 at 19 54 44" src="https://github.com/lightdash/lightdash/assets/12776522/2cfad6eb-ae3a-4b1a-a22f-3c773cf837af">

PriPatel:
Dimension error | field no longer exists
-- | --
Metric error | `field` no longer exists
Filter error | `field` no longer exists
Sorting error | `field` no longer exists
Broken chart | `chart name` is broken on this dashboard

joaoviana:
Same (proposal 1 🎉 )! I can start working on this, but need to sync with @rephus on what best way to get the Error UI look like that

joaoviana:
Implement better messaging for Table/Model errors as well

Users should be able to see what is the name of the Table that is throwing the error. 
Table errors that throw: 
* 'Model has no table relation' error
* Field at xxx errors (AjvError)

PR #5682, opened 2023-05-30
feat: update validation model and table

joaoviana:
Relates to: #5653 

### Description:

Add the following new columns: 
* 'error_type' (see enum)
* 'chart_name (for dashboard errors)
* 'field_name' (for any type of error except tables)
* 'model_name' (for table/explore errors)
* 'dimension_name' (for table/explore errors)

Keep the `error` message for now

Improve types and add type guards

Will then make changes to the UI - no current properties have changes, so it will not break the UI)

### DIFF
--- a/packages/backend/src/database/migrations/20230530130533_validations_update_table.ts
+++ b/packages/backend/src/database/migrations/20230530130533_validations_update_table.ts
@@ -4,11 +4,11 @@ const ValidationTableName = 'validations';
 
 export async function up(knex: Knex): Promise<void> {
     await knex.schema.alterTable(ValidationTableName, (tableBuilder) => {
-        tableBuilder.string('error_type').nullable().defaultTo(null);
-        tableBuilder.string('chart_name').nullable().defaultTo(null);
-        tableBuilder.string('field_name').nullable().defaultTo(null);
-        tableBuilder.string('model_name').nullable().defaultTo(null);
-        tableBuilder.string('dimension_name').nullable().defaultTo(null);
+        tableBuilder.string('error_type').nullable();
+        tableBuilder.string('chart_name').nullable();
+        tableBuilder.string('field_name').nullable();
+        tableBuilder.string('model_name').nullable();
+        tableBuilder.string('dimension_name').nullable();
     });
 }
 
--- a/packages/backend/src/database/migrations/20230530130533_validations_update_table.ts
+++ b/packages/backend/src/database/migrations/20230530130533_validations_update_table.ts
@@ -1,0 +1,25 @@
+import { Knex } from 'knex';
+
+const ValidationTableName = 'validations';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(ValidationTableName, (tableBuilder) => {
+        tableBuilder.string('error_type').nullable().defaultTo(null);
+        tableBuilder.string('chart_name').nullable().defaultTo(null);
+        tableBuilder.string('field_name').nullable().defaultTo(null);
+        tableBuilder.string('model_name').nullable().defaultTo(null);
+        tableBuilder.string('dimension_name').nullable().defaultTo(null);
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(ValidationTableName, (tableBuilder) => {
+        tableBuilder.dropColumns(
+            'error_type',
+            'chart_name',
+            'field_name',
+            'model_name',
+            'dimension_name',
+        );
+    });
+}
--- a/packages/backend/src/models/ValidationModel/ValidationModel.ts
+++ b/packages/backend/src/models/ValidationModel/ValidationModel.ts
@@ -226,8 +226,8 @@ export class ValidationModel {
                 validationId: validationError.validation_id,
                 spaceUuid: validationError.space_uuid,
                 errorType: validationError.error_type,
-                fieldName: validationError.field_name,
-                chartName: validationError.chart_name,
+                fieldName: validationError.field_name ?? null,
+                chartName: validationError.chart_name ?? null,
             }));
 
         const tableValidationErrorsRows: DbValidationTable[] =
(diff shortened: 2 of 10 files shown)


rephus:
Need to fix tests 

src/types/validation.ts(96,15): error TS2339: Property 'chartUuid' does not exist on type 'never'.
src/types/validation.ts(97,15): error TS2339: Property 'dashboardUuid' does not exist on type 'never'.

packages/backend/src/database/migrations/20230530130533_validations_update_table.ts
```diff
@@ -0,0 +1,25 @@
+import { Knex } from 'knex';
+
+const ValidationTableName = 'validations';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(ValidationTableName, (tableBuilder) => {
+        tableBuilder.string('error_type').nullable().defaultTo(null);
```
rephus, line 7: I don't think this defaultTo(null) is necessary, check if we do it on other migrations

packages/backend/src/models/ValidationModel/ValidationModel.ts
```diff
@@ -38,10 +42,28 @@ export class ValidationModel {
         await this.database.transaction(async (trx) => {
             const insertPromises = validations.map((validation) =>
                 trx(ValidationTableName).insert({
-                    saved_chart_uuid: validation.chartUuid,
-                    dashboard_uuid: validation.dashboardUuid,
+                    saved_chart_uuid: isChartValidationError(validation)
```
rephus, line 45: why do you need to do this check ? isChartValidationError

packages/backend/src/models/ValidationModel/ValidationModel.ts
```diff
@@ -38,10 +42,28 @@ export class ValidationModel {
...
+                        ? validation.chartUuid
+                        : null,
+                    dashboard_uuid: isDashboardValidationError(validation)
+                        ? validation.dashboardUuid
+                        : null,
                     project_uuid: validation.projectUuid,
                     error: validation.error,
+                    error_type: validation.errorType,
+                    model_name:
+                        'modelName' in validation ? validation.modelName : null,
```
rephus, line 55: can't you do `validation.modelName || null` ?

packages/common/src/types/validation.ts
```diff
@@ -1,24 +1,73 @@
...
-    chartUuid?: string;
-    chartType?: ChartKind;
-    dashboardUuid?: string;
+    error: string;
+    errorType: ValidationErrorType | null;
     projectUuid: string;
     spaceUuid?: string;
-    error: string;
+
+    // TODO: refactor these
```
rephus, line 12: todo ?

packages/backend/src/models/ValidationModel/ValidationModel.ts
rephus: oh I see, you created new types with different fields, in that case, how about creating a differnet `insertChart / insertDashboard` methods with the right CreateChatValidation as argument, so you don't need to do all these checks

rephus (commented):
Nice factoring, it is a bit messy now, but I can't think of a better way (without abusing undefines)

packages/backend/src/models/ValidationModel/ValidationModel.ts
joaoviana: Yea, the new types make it clearer on what gets inserted, so having those type checks is necessary. I've take your suggestions here and here: https://github.com/lightdash/lightdash/pull/5682#discussion_r1211372953, and refactored this to make it cleaner. LMK what you think 😄 https://github.com/lightdash/lightdash/pull/5682/commits/6718f6bcd45bf37bb6ecadc6a139d5840f35de57

packages/common/src/types/validation.ts
joaoviana: thanks!! doooone!

joaoviana:
> Need to fix tests
> 
> src/types/validation.ts(96,15): error TS2339: Property 'chartUuid' does not exist on type 'never'. src/types/validation.ts(97,15): error TS2339: Property 'dashboardUuid' does not exist on type 'never'.

fixed!

packages/backend/src/database/migrations/20230530130533_validations_update_table.ts
joaoviana: exactly, we don't! thank you

packages/common/src/types/validation.ts
```diff
@@ -1,24 +1,72 @@
...
-export type ValidationResponse = {
+type ValidationResponseBase = {
     validationId: number;
     createdAt: Date;
     name: string;
-    chartUuid?: string;
-    chartType?: ChartKind;
-    dashboardUuid?: string;
+    error: string;
+    errorType: ValidationErrorType | null;
```
rephus, line 8: on validation response we can do `errorType?: validationErrorType` or `errorType: validationErrorType | undefined` 

afaik null is only used on db models

packages/common/src/types/validation.ts
```diff
@@ -1,24 +1,72 @@
...
     lastUpdatedBy?: string;
     lastUpdatedAt?: Date;
 };
 
-export type CreateValidation = Pick<
-    ValidationResponse,
-    'name' | 'chartUuid' | 'dashboardUuid' | 'projectUuid' | 'error'
+export type ValidationErrorDashboardResponse = ValidationResponseBase & {
+    dashboardUuid: string | undefined; // NOTE: can be undefined if private content
+    chartName?: string | null;
```
rephus, line 23: no need to use null here

rephus (commented):
can we remove `null`s on the non-db models ? 

We used to store enums in db as well, but I don't think we need it here

packages/common/src/types/validation.ts
joaoviana: amazing, thanks! yes yes

packages/common/src/types/validation.ts
joaoviana: 🤦 yes!

rephus (approved):
LGTM

Issue #5673, opened 2023-05-30
Lightdash validate should default to running on my current project ID


TuringLovesDeathMetal:
### Description

Ideally, my flow would be something like this:
1. If I create a preview environment, the validator should run and appear in my project validation notifications (just like it would on my regular project). So I see a bell icon with any errors when I’m in there already
2. If I don’t want to create a preview environment, then I can run lightdash validate and it will validate against my project I’m “logged into”.

For (2) - I’d expect this to act the same way whether or not I created a preview environment before…basically, I’d want it to check the most up-to-date version of my project because if I go to try to open a PR for it, it will check against the most up-to-date version anyways. So if it’s changed between me running lightdash preview, I’d want to know that my changes aren’t breaking any new stuff either

### Acceptance criteria

- [ ] By default, if I run `lightdash validate`, it should run against the project I've selected on the CLI. 
- [ ] If you haven't selected a project, then we show you an error saying you need to specify the project UUID (same as the error we have right now)

PR #5685, opened 2023-05-31
feat: default project on validate

rephus:
Closes:  #5673

### Description:

image

### DIFF
--- a/packages/cli/src/handlers/validate.ts
+++ b/packages/cli/src/handlers/validate.ts
@@ -86,10 +86,14 @@ export const validateHandler = async (options: ValidateHandlerOptions) => {
         console.error(
             `Validating preview project ${styles.bold(
                 config.context?.previewName,
-            )} with projectUuid ${projectUuid}\n`,
+            )}\n`,
         );
     } else {
-        console.error(`Validating project with projectUuid ${projectUuid}\n`);
+        console.error(
+            `Validating project ${styles.bold(
+                config.context?.projectName || projectUuid,
+            )}\n`,
+        );
     }
 
     const timeStart = new Date();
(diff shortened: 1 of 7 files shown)


packages/cli/src/handlers/validate.ts
```diff
@@ -6,10 +6,10 @@ import {
     SchedulerJobStatus,
 } from '@lightdash/common';
 import columnify from 'columnify';
-import { getConfig } from '../../config';
-import GlobalState from '../../globalState';
-import * as styles from '../../styles';
-import { checkLightdashVersion, lightdashApi } from './apiClient';
+import { getConfig } from '../config';
```
rephus, line 9: I moved this validate out of the dbt folder

packages/cli/src/handlers/validate.ts
```diff
@@ -63,26 +63,33 @@ export const validateHandler = async (options: ValidateHandlerOptions) => {
...
 
     if (options.project) {
         console.error(`Validating project ${projectUuid}\n`);
-    } else {
+    } else if (config.context?.previewProject) {
         console.error(
             `Validating preview project ${styles.bold(
                 config.context?.previewName,
-            )} with projectUuid ${config.context?.previewProject}\n`,
+            )} with projectUuid ${projectUuid}\n`,
```
TuringLovesDeathMetal, line 89: Can we change this to the project name? 

So like: `Validating preview project **Humble Bumblebee**` (where the project name is bold)

Or `Validating project **Lightdash Analytics**`


^I know that projects can have the same name, but it's really unlikely for them to be named the same and I feel like the name is more intuitive/helpful than the UUID

TuringLovesDeathMetal (changes requested):
This looks really great! Just one small change for replacing teh UUID to the project name in one fo the messages :)

packages/cli/src/handlers/validate.ts
rephus: at the moment we don't store the project name, but I can add it , let me change this

TuringLovesDeathMetal (approved):
nice!

Issue #5480, opened 2023-05-12
It looks like you can collapse Settings Sidebar groups (when you can't)


PriPatel:
<img width="410" alt="Screenshot 2023-05-12 at 17 57 05" src="https://github.com/lightdash/lightdash/assets/12776522/17e71431-8edc-4fbb-9ebd-15639db09934">

The toggle makes it seem like I can collapse the group. I should not be able to collapse it.

PR #5687, opened 2023-05-31
fix: settings restructure collapsible navbar

IrakliJani:
Closes: #5480 

### Description:

before
CleanShot 2023-05-31 at 12 45 15@2x

after
CleanShot 2023-05-31 at 12 44 21@2x


yasmin-e (approved):
looks great

Issue #5666, opened 2023-05-30
Update openapi.json to include space/project permissions


owlas:
The existing openapi spec doesn't include any documentation on how to use the space and project permissions API.

PR #5691, opened 2023-05-31
chore: replace api docs with tsoa generated docs

owlas:
Removes the old `.json` files and uses the new autogenerated spec from the tsoa controllers. The new documentation is not as comprehensive yet, so some routes lose documentation. we'll address this in a separate PR such as #5666 

### Docs site reference

docs

### Backend raw json

backend-raw

### Backend redocly

backend-redocly


ZeRego:
@owlas what is the new command to generate the swagger file ?
We should update `"generate-api": "tsoa routes",` command. And make it part of our release github action to make sure on every merge is up to date.

ZeRego (approved):

PR #5693, opened 2023-05-31
fix: move pivot data calculations to web worker

ZeRego:
Closes:

### DIFF
--- a/packages/frontend/src/components/SimpleTable/index.tsx
+++ b/packages/frontend/src/components/SimpleTable/index.tsx
@@ -45,7 +45,7 @@ const SimpleTable: FC<SimpleTableProps> = ({
         explore,
     } = useVisualizationContext();
     const scrollableContainerRef = useRef<HTMLDivElement>(null);
-    if (isLoading) return <LoadingChart />;
+    if (isLoading || pivotTableData.loading) return <LoadingChart />;
 
     if (error) {
         return (
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -20,6 +20,7 @@
         "@mantine/spotlight": "^6.0.9",
         "@sentry/react": "^7.37.2",
         "@sentry/tracing": "^7.37.2",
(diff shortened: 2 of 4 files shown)


packages/frontend/src/components/SimpleTable/index.tsx
```diff
@@ -57,28 +57,34 @@ const SimpleTable: FC<SimpleTableProps> = ({
...
             >
-                <PivotTable
-                    containerRef={scrollableContainerRef}
-                    data={pivotTableData.data}
-                    conditionalFormattings={conditionalFormattings}
-                    getFieldLabel={getFieldLabel}
-                    getField={getField}
-                    hideRowNumbers={hideRowNumbers}
-                />
+                {pivotTableData.data ? (
```
ZeRego, line 76: If `scrollableContainerRef` is null when <PivotTable> mounts, the virtualization won't work as expected.
This was the only way I found to guarantee that `scrollableContainerRef` is not null when we load the pivot table.

joaoviana: cool, we could also wrap a check in a `useEffect` to track when it mounts if you want to be more specific, like so

```tsx
useEffect(() => {
    if (scrollableContainerRef.current) {
      setIsScrollableContainerRefMounted(true)
    }
  }, [scrollableContainerRef.current]); 
```

But this works so that's cool!

joaoviana (approved):
🎉

packages/frontend/src/components/SimpleTable/index.tsx
joaoviana: edit: I think you've tried that but it didn't work perhaps 😅

ZeRego: mutating the reference won't trigger a render, therefore, won't trigger the useEffect.

joaoviana: 🤦  of course!

PR #5699, opened 2023-05-31
fix: virtual table with empty ref

ZeRego:
### Description:

Fixes edge case where the virtual table doesn't recognize the latest `ref` value and doesn't render more rows.

More details on this issue from the library github https://github.com/TanStack/virtual/issues/263#issuecomment-1040073538

Updating to the latest version solves the issue since it now has a callback to fetch the latest `ref` value.

https://github.com/lightdash/lightdash/assets/9117144/ff55961f-0428-4bc4-aae0-0a18fed87970

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -12,7 +12,7 @@ import { Table, TableProps } from '@mantine/core';
 import last from 'lodash-es/last';
 import React, { FC, useCallback } from 'react';
 
-import { useVirtual } from '@tanstack/react-virtual';
+import { useVirtualizer } from '@tanstack/react-virtual';
 import { isSummable } from '../../../hooks/useColumnTotals';
 import HeaderCell from './HeaderCell';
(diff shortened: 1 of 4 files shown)


IrakliJani (approved):

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -181,17 +181,19 @@ const PivotTable: FC<PivotTableProps> = ({
...
 
-    const rowVirtualizer = useVirtual({
-        parentRef: containerRef,
-        size: data.dataValues.length,
-        overscan: 10,
+    const rowVirtualizer = useVirtualizer({
+        getScrollElement: () => containerRef.current,
+        count: data.dataValues.length,
+        estimateSize: () => 34,
+        overscan: 25,
```
joaoviana, line 188: non-blocking, could you explain why 34?

ZeRego: is the current row height in pixels

joaoviana (approved):
🚢 it!

Issue #5695, opened 2023-05-31
Deleting a project should follow the same UX as deleting an organization


TuringLovesDeathMetal:
### Description

Right now, it's really easy to delete a project. 

Instead, it should follow the same flow as deleting an organization.

![image](https://github.com/lightdash/lightdash/assets/31848148/ac0f8b4d-eaf9-4e55-90fd-f00b8a998013)

### Acceptance criteria

- [ ] You need to enter the name of the project to delete it
- [ ] The modal header says `delete project`
- [ ] The modal body says `Type the name of this project Project Name to confirm you want to delete this project. This action is not reversible.`

PR #5709, opened 2023-06-01
feat: delete project dialog

rephus:
Closes:  #5695
### Description:

[Screencast from 01-06-23 11:51:09.webm](https://github.com/lightdash/lightdash/assets/1983672/2fb7094f-1902-48ab-b449-579b8a78a74a)


packages/frontend/src/components/UserSettings/ProjectManagementPanel/index.tsx
```diff
@@ -38,9 +30,7 @@ const ProjectListItem: FC<{
 }> = ({ isCurrentProject, project: { projectUuid, name, type } }) => {
     const { user } = useApp();
     const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
-    const { mutate: deleteProjectMutation, isLoading: isDeleting } =
-        useDeleteProjectMutation();
-    const { mutate: deleteLastProjectMutation } =
```
IrakliJani, line 43: I can't find any reference to this in the new modal. we need to call this in order to remove it from the localStorage if it's the active project that user is deleting

packages/frontend/src/components/UserSettings/DeleteProjectPanel/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return (
+        <Modal
+            size="md"
+            centered
+            opened={opened}
+            title={
+                <Group spacing="xs">
+                    <MantineIcon size="lg" icon={IconAlertCircle} color="red" />
+                    <Title order={4}>Delete Project</Title>
+                </Group>
```
IrakliJani, line 50: strange that mantine modal does not have an icon prop :/

packages/frontend/src/components/UserSettings/DeleteProjectPanel/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+            }
+            onClose={handleOnClose}
+        >
+            <Stack>
+                <Text>
+                    Type the name of this Project <b>{project.name}</b> to
```
IrakliJani, line 56: I find <b> to be very heavy. instead you can use <Text inline fw={600}>{project.name}</Text>

packages/frontend/src/components/UserSettings/ProjectManagementPanel/index.tsx
rephus: It is done here, in this file, like before, in a callback  instead of the modal . Let me know if you want me to change this. https://github.com/lightdash/lightdash/pull/5709/files#diff-718ea4de3540be10337e33184af77ff0d804d58f7be84651dc81d0aba63eab7eR86

packages/frontend/src/components/UserSettings/DeleteProjectPanel/DeleteProjectModal.tsx
rephus: I copied the existing delete org modal , so I didn't change this logic, if we decide to change this style I'd suggest we change both, maybe on a separate PR

packages/frontend/src/components/UserSettings/ProjectManagementPanel/index.tsx
IrakliJani: there are two mutations that needs to be run 

this is old code and it's very confusing. perhaps deleteProjectMutation should call that `deleteLastProjectMutation` mutation. one deletes the project and another deletes localStorage entry if it's an active project
```
                            onClick={() => {
                                deleteProjectMutation(projectUuid);
                                if (isCurrentProject) {
                                    deleteLastProjectMutation();
                                }
                            }}
```

packages/frontend/src/components/UserSettings/DeleteProjectPanel/DeleteProjectModal.tsx
IrakliJani: it's all good! I am just surprised and confused that mantine does not support icon prop

Issue #5546, opened 2023-05-19
Use AWS SSL cert bundles with postgres connections (for RDS users)


owlas:
### Description

Slack community thread: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1684265111924659

Users cannot connect to Postgres on AWS RDS because we don't have the SSL certs and they require ssl connections. 

- Cert bundle: https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
- Docs: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html

PR #5718, opened 2023-06-01
feat: add aws rds certs to postgres

owlas:
Closes: #5546 

Injects the aws ssl certs into the default CA bundle used by node TLS.

To do this I've had to replicate the behaviour for converting a connection string `sslmode` into an `ssl` object for node TLS.

Need to test this with an amazon RDS instance and local postgres instance.

### Testing

Created a project here with `sslmode=verify-full` which should require an matching SSL certificate. Queries execute and project compiles with dbt! So both are reachable with full SSL enabled.

https://lightdash-pr-5718.onrender.com/generalSettings/projectManagement/982d8ded-bcc7-4088-b3e1-258ee8978008/settings


owlas:
Marking as draft since we need to ensure this works with "refresh dbt"

packages/backend/src/dbt/profiles.ts
```diff
@@ -6,6 +6,12 @@ import {
 import * as yaml from 'js-yaml';
 import path from 'path';
 
+console.log(
```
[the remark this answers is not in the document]
owlas: ty

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -252,16 +261,36 @@ export class PostgresClient<
...
+    switch (mode) {
+        case 'disable':
+            return false;
+        case 'prefer':
+        case 'require':
+        case 'allow':
+        case 'verify-ca':
+        case 'verify-full':
+            return {
+                ca: POSTGRES_CA_BUNDLES,
```
ZeRego, line 275: Do we need to check the host value like we do in the profiles.ts ? 
```credentials.host.endsWith('.rds.amazonaws.com')```
Will this have an impact on all the postgres connections that aren't hosted in aws ?

owlas: So node doesn't use postgres drivers at all to communicate with postgres and uses the node TLS package to communicate. So it's quite different to how it works with `profiles.ts` 

With node we can actually specific a whole array of certificates. All I've done is append the RDS one to the list of built in certs (which is the mozilla CA cert bundle).

So this should include _all_ certs we already had PLUS the RDS ones. With the postres driver (used by dbt) we have to point to a single file of certificates

BUT you're right, this code will touch all postgres deployments (I don't think there are many of those in production at all). If we added the if statement, we would reduce the amount of users affected by this change. But also might be more confusing for devs, since it's not really necessary.

ZeRego (approved):

Issue #4334, opened 2023-01-31
Metric filter does not work for NULL values


TuringLovesDeathMetal:
### Description

I want to add `is equal to null` to my metric filters: https://docs.lightdash.com/references/metrics#available-filter-types

So, I try adding it like this:

```yml
missed_hours_unknown_reason:
  label: Hours - Without Parent Note
  type: sum
  filters:
    - explanation_source: null
```

But, the filter doesn't work. 

I want to be able to use metric filters for `NULL` and `!NULL`

rephus:
I can add `field: null` but I can't add `field: !null` to the grammar because that is not a valid yml syntax, so dbt doesn't like it:

![image](https://github.com/lightdash/lightdash/assets/1983672/e7fd2f04-3825-45eb-9421-ef6da35dd2fe)

I could do `field: "!null" ` or any other valid yml syntax 

@TuringLovesDeathMetal @owlas

owlas:
You can wrap it in quotes, those won't be there after the yaml has been parsed and passed to the grammar.

PR #5729, opened 2023-06-02
feat: add is_null operator grammar

rephus:
Closes: #4334

### DIFF
--- a/packages/common/src/types/filterGrammar.test.ts
+++ b/packages/common/src/types/filterGrammar.test.ts
@@ -2,7 +2,7 @@ import peg from 'pegjs';
 import { FilterOperator, FilterRule } from './filter';
 import filterGrammar, { parseFilters } from './filterGrammar';
 
-describe('attachTypesToModels', () => {
+describe('Parse grammar', () => {
     const parser = peg.generate(filterGrammar);
 
     it('Simple peg grammar test', async () => {
@@ -89,6 +89,20 @@ describe('attachTypesToModels', () => {
         expect(parser.parse('>= 32')).toEqual(expected);
         expect(parser.parse(' >=32')).toEqual(expected);
         expect(parser.parse(' >= 32')).toEqual(expected);
+    });
+
+    it('Is null', async () => {
+        expect(parser.parse('NULL')).toEqual({
+            is: true,
+            type: 'null',
+        });
+    });
+
+    it('Is not null', async () => {
+        expect(parser.parse('!null')).toEqual({
+            is: false,
+            type: 'null',
+        });
     });
 });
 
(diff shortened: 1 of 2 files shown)


packages/common/src/types/filterGrammar.test.ts
```diff
@@ -90,6 +90,20 @@ describe('attachTypesToModels', () => {
...
+
+    it('Is null', async () => {
+        expect(parser.parse('NULL')).toEqual({
+            is: true,
+            type: 'null',
+        });
+    });
+
+    it('Is not null', async () => {
+        expect(parser.parse('!null')).toEqual({
```
owlas, line 102: Should it work for both lower and uppercase?

packages/common/src/types/filterGrammar.test.ts
```diff
@@ -90,6 +90,20 @@ describe('attachTypesToModels', () => {
         expect(parser.parse(' >=32')).toEqual(expected);
         expect(parser.parse(' >= 32')).toEqual(expected);
     });
+
+    it('Is null', async () => {
```
owlas, line 94: can you escape the null value to match the exact string "null"? I don't think it's a requirement but what's the current behaviour?

owlas (commented):
see inline

packages/common/src/types/filterGrammar.test.ts
rephus: yes, because of this lowercase https://github.com/lightdash/lightdash/pull/5729/files#diff-546a539fae0c7fc3cf99abec54c5cb175292ec43d5f4525a15bdb84080ba0be5R181

ZeRego (commented):
We should update the docs with the new option 🙌

packages/common/src/types/filterGrammar.test.ts
rephus: Not sure what you mean, but this parser method takes only a  string as an argument, to process `null`s I don't use `peg` https://github.com/lightdash/lightdash/pull/5729/files#diff-546a539fae0c7fc3cf99abec54c5cb175292ec43d5f4525a15bdb84080ba0be5R218

IrakliJani (approved):
feels little misleading since **operation** is **equals** or **not equals** and **null** is a **value**. otherwise everything seems fine by me

TuringLovesDeathMetal (approved):
nice! looks great 👍

docs/docs/references/metrics.mdx
```diff
@@ -587,7 +587,8 @@ These filters do not appear in the `Filters` tab in the Explore view, instead, t
 | is greater than or equal to | Number of orders is greater than or equal to 4 | `num_orders: '>= 4'`  |
 | is less than                | Number of orders is less than 4                | `num_orders: '< 4'`   |
 | is less than or equal to    | Number of orders is less than or equal to 4    | `num_orders: '<= 4'`  |
-
+| is null                     | Status is null                                 | `status: null         |
+| is not null                 | Status is not null                             | `status: '!null'`     |
```
TuringLovesDeathMetal, line 591:
```suggestion
| is not null                 | Status is not `NULL`                             | `status: '!null'`     |
```

packages/common/src/types/filterGrammar.test.ts
owlas: I mean would be possible to create a filter that's like `myvalue === "null"` rather than `myvalue === null`

Issue #5369, opened 2023-05-04
Scheduled Deliveries dashboard UI update


PriPatel:
### Pre-requisite: #5678 

**Design / Prototype**
We should use the same table component that we use on the homepage:
https://www.figma.com/proto/gtLzWNBwSX1ejmwz7MEm4W/Untitled?node-id=84-21853&scaling=min-zoom&page-id=77%3A24113&starting-point-node-id=89%3A12838

**Content for each column and each tab**
Some more info about each column, different cases and what to expect on hover:
https://www.notion.so/lightdash/Scheduled-Deliveries-Dashboard-a1484d74b5ce4ff1ba389324f401fee8#d1f8fa76fbc74fe1854948f929ad1aed

**Icons and colours**
Icon / Content | Tabler icon | Colour (mantine)
-- | -- | --
Schedule type : csv | `csv` | indigo6
Schedule type : Image | `photo` | indigo6
Status : Error | `alert-triangle-filled` | red6
Status : In progress | `progress` | yellow6
Status : Scheduled | `clock-filled` | blue6
Status: Completed | `circle-check-filled` | green6

**Run history query changes:**

Only show logs where delivery scheduled date is between 7 days ago and current time.

PR #5731, opened 2023-06-02
feat: schedulers view

yasmin-e:
Closes: #5369


TuringLovesDeathMetal:
niiiice! this looks really good 😍 

One small change: for the `refresh` icon, what does it actually do? 

image

Can we add some text on hover (e.g. `refresh the status of the scheduled deliveries`).

Can we also add a success toast once the refresh is complete? Something like `Success! Scheduled deliveries status update completed.`

TuringLovesDeathMetal:
I saw that you re-requested a review - I'm still not seeing anything on the refresh button 🤔 but not sure if you haven't changed that yet!

TuringLovesDeathMetal:
I just ran a test scheduled delivery. 

When I opened the `all jobs` toggle, it showed the same job, but repeated 3 times. I'd expect it to only be shown once (i.e. `send email notification`, `handle scheduled delivery`). 

image

Also, in the design, there was an icon showing success/failure for each of the jobs. (then one for the "overall" job). 

Have we scrapped that? 
image

yasmin-e:
> I just ran a test scheduled delivery.
> 
> When I opened the `all jobs` toggle, it showed the same job, but repeated 3 times. I'd expect it to only be shown once (i.e. `send email notification`, `handle scheduled delivery`).
> 
> image
> Also, in the design, there was an icon showing success/failure for each of the jobs. (then one for the "overall" job).
> 
> Have we scrapped that? image

all fixed! 💯

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                                    logs,
+                                                )[0],
+                                                theme,
+                                            )}
+                                        </Stack>
+                                    </Collapse>
+                                </Stack>
+                            ) : (
+                                <Text fz={12} color="gray.6">
+                                    -
```
IrakliJani, line 295: do we really want to show this dash?

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                    );
+                },
+            },
+            {
+                id: 'status',
+                label: 'Status',
+                cell: (item) => {
+                    return (
+                        <Stack align="center" justify="center">
+                            {currentLogs(item, logs).length > 0 ? (
```
IrakliJani, line 268: better if you save this above as a variable and reuse below when you pass it to the function

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                        </Text>
+                    );
+                },
+            },
+            {
+                id: 'status',
+                label: 'Status',
+                cell: (item) => {
+                    return (
+                        <Stack align="center" justify="center">
```
IrakliJani, line 267: I think `<Center>` component is more appropriate here

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                cell: (item) => {
+                    return currentLogs(item, logs).length > 0 ? (
+                        <Box>
+                            <Group spacing="xxs">
+                                <Text fz={12} fw={500} color="gray.6">
+                                    {formatTime(
+                                        currentLogs(item, logs)[0].createdAt,
+                                    )}
+                                </Text>
+                            </Group>
```
IrakliJani, line 237: do we need Grid wrapper with spacing if we have only 1 text element inside?

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                            <Collapse in={opened}>
+                                <Text fz={12} fw={500} pt="md" color="gray.6">
+                                    {formatTime(
+                                        handleScheduledDeliveryLogs(
+                                            item,
+                                            logs,
+                                        )[0].createdAt,
+                                    )}
+                                </Text>
+                                <Text fz={12} fw={500} pt="md" color="gray.6">
```
IrakliJani, line 247: both of those are wrapped in a same text component. you can have one component (for example Collapse) which will apply font and color properties and two <Text elements without props and they will inherit those font styles from the parent

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                    />
+                                </ActionIcon>
+                            </Group>
+                            <Collapse in={opened}>
+                                <Text fz={12} fw={500} pt="md" color="gray.6">
+                                    {handleScheduledDeliveryLogs(item, logs)[0]
+                                        .task.replace(/([A-Z])/g, ' $1')
+                                        .toLowerCase()}
+                                </Text>
+                                <Text fz={12} fw={500} pt="md" color="gray.6">
```
IrakliJani, line 173: same wrappers, can be one Stack or Text component with this set of props and two nested Text components with no props.

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                    {sendNotificationLogs(item, logs)[0]
+                                        .task.replace(/([A-Z])/g, ' $1')
+                                        .toLowerCase()}
```
IrakliJani, line 176: this is repeated, maybe extract it as a separate function

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                              (chart) =>
+                                  chart.savedChartUuid === item.savedChartUuid,
+                          )
+                        : dashboards.find(
+                              (dashboard) =>
+                                  dashboard.dashboardUuid ===
+                                  item.dashboardUuid,
+                          );
+                    return (
+                        <Anchor
```
IrakliJani, line 87: can you check if `unstyled` prop works here? sx={} might be unnecessary

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                            <Stack spacing="two">
+                                                <Text fz={12} color="gray.5">
+                                                    Schedule type:{' '}
+                                                    <Text color="white" span>
+                                                        {item.format === 'csv'
+                                                            ? 'CSV'
+                                                            : 'Image'}
+                                                    </Text>
+                                                </Text>
+                                                <Text fz={12} color="gray.5">
```
IrakliJani, line 112: same comment about the container with identical text styles

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                    return (
+                        <Anchor
+                            sx={{
+                                color: 'unset',
+                                ':hover': {
+                                    color: 'unset',
+                                    textDecoration: 'none',
+                                },
+                            }}
+                            href={getSchedulerLink(item, projectUuid)}
```
IrakliJani, line 95: if the link is internal we should try using react router navigation instead of using href.

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                            </Stack>
+                                        }
+                                    >
+                                        <Text
+                                            fw={600}
+                                            lineClamp={1}
+                                            sx={{
+                                                overflowWrap: 'anywhere',
+                                                '&:hover': {
+                                                    textDecoration: 'underline',
```
IrakliJani, line 128: it feels like you are trying to add an anchor to the whole cell and only apply the hover link UX to the name. should we consider making this component an anchor and removing it from the whole cell?

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                        <Box>
+                            <Group spacing="two">
+                                <Text fz={12} fw={500}>
+                                    All jobs
+                                </Text>
+                                <ActionIcon onClick={toggle} size="sm">
+                                    <MantineIcon
+                                        icon={IconChevronDown}
+                                        color="black"
+                                        size={13}
```
IrakliJani, line 163: does `sm` look too bad? I feel like we are still using a lot of custom sizes and trying to understand the reason

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                },
+            },
+            {
+                id: 'deliveryScheduled',
+                label: 'Delivery scheduled',
+                cell: (item) => {
+                    return currentLogs(item, logs).length > 0 ? (
+                        <Box>
+                            <Group spacing="xxs">
+                                <Text fz={12} fw={500} color="gray.6">
```
IrakliJani, line 194: you should be able to use xs, sm and stuff like that instead of 12

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                                            logs,
+                                        )[0].scheduledTime,
+                                    )}
+                                </Text>
+                                <Text fz={12} fw={500} pt="md" color="gray.6">
+                                    {formatTime(
+                                        sendNotificationLogs(item, logs)[0]
+                                            .scheduledTime,
+                                    )}
+                                </Text>
```
IrakliJani, line 215: if both of the elements have PT I believe the wrapping Box can become a Stack with sizing

packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                <tr>
+                    {columns.map((column) => {
+                        return (
+                            <Box
+                                component="th"
+                                key={column.id}
+                                style={column?.meta?.style}
+                            >
+                                {column?.label}
+                            </Box>
```
IrakliJani, line 320: you can directly use <th component if there's nothing special and mantine-y going on with the element

packages/frontend/src/components/SchedulersView/SchedulersView.tsx
```diff
@@ -0,0 +1,378 @@
...
+};
+
+export const getLogStatusIcon = (log: Log, theme: MantineTheme) => {
+    switch (log.status) {
+        case SchedulerJobStatus.SCHEDULED:
+            return (
+                <Tooltip label={SchedulerJobStatus.SCHEDULED}>
+                    <MantineIcon
+                        icon={IconClockFilled}
+                        color="blue.3"
```
IrakliJani, line 91: I feel like those functions should be extracted as separate utils, because they are being used in both views

packages/frontend/src/components/SchedulersView/SchedulersView.tsx
```diff
@@ -0,0 +1,378 @@
...
+        : `/projects/${projectUuid}/dashboards/${item.dashboardUuid}/view`;
+};
+
+export const formatTime = (date: Date) => {
+    return new Date(date).toLocaleString('en-US', {
+        timeZone: 'UTC',
+        dateStyle: 'short',
+        timeStyle: 'short',
+    });
+};
```
IrakliJani, line 146: these ones too

packages/frontend/src/components/SchedulersView/LogsView.tsx
yasmin-e: unfortunately `sm` is too small and `md` too big :/

packages/frontend/src/components/SchedulersView/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                <Tabs.Panel value="scheduled-deliveries">
+                    {data && data.schedulers.length > 0 ? (
+                        <Schedulers {...data} projectUuid={projectUuid} />
+                    ) : (
+                        emptyState
+                    )}
+                </Tabs.Panel>
+                <Tabs.Panel value="run-history">
+                    {data &&
+                    data.schedulers.length > 0 &&
```
IrakliJani, line 107: shouldn't this still render if schedulers are 0?

you also check `data && data.schedulers.length > 0 ?` below which is confusing in terms of how logic branches

packages/frontend/src/components/SchedulersView/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                            ml="auto"
+                            onClick={() =>
+                                queryClient
+                                    .invalidateQueries('schedulerLogs')
+                                    .then(() =>
+                                        showToastSuccess({
+                                            title: 'Scheduled deliveries refreshed successfully',
+                                        }),
+                                    )
+                            }
```
IrakliJani, line 87: I would extract this above in the component level as a handler function `handleRefresh`

packages/frontend/src/components/SchedulersView/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+            title="No scheduled deliveries on this project"
+            description="Go to a chart or dashboard to set up your first scheduled delivery"
+        />
+    );
+    const logsEmptyState = (
+        <ResourceEmptyState
+            title="Scheduled deliveries have not run any jobs as of now"
+            description="Check in later of hit the refresh button to see if any jobs have run"
+        />
+    );
```
IrakliJani, line 30: you can extract these out of this function to avoid calling it each time component rerenders

packages/frontend/src/components/SchedulersView/SchedulersViewUtils.tsx
```diff
@@ -0,0 +1,124 @@
...
+                />
+            );
+        case 'image':
+            return (
+                <IconBox
+                    icon={IconPhoto}
+                    color="indigo.6"
+                    style={{ color: theme.colors.indigo[6] }}
+                />
+            );
```
IrakliJani, line 55: can you add default case with assertUnreachable check? it's always nice to have exhaustive switch blocks

packages/frontend/src/components/SchedulersView/SchedulersViewUtils.tsx
```diff
@@ -0,0 +1,124 @@
...
+        case SchedulerJobStatus.ERROR:
+            return (
+                <Tooltip label={log?.details?.error}>
+                    <MantineIcon
+                        icon={IconAlertTriangleFilled}
+                        color="red.6"
+                        style={{ color: theme.colors.red[6] }}
+                    />
+                </Tooltip>
+            );
```
IrakliJani, line 100: same here

packages/frontend/src/components/SchedulersView/SchedulersViewActionMenu.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </Menu>
+            <SchedulerDeleteModal
+                lazy={true}
+                isOpen={isDeleting}
+                schedulerUuid={item.schedulerUuid}
+                onConfirm={() => {
+                    setIsDeleting(false);
+                    queryClient.invalidateQueries('schedulerLogs');
+                }}
+                onClose={() => queryClient.invalidateQueries('schedulerLogs')}
```
IrakliJani, line 107: will be nice toextract this to handleDelete function

packages/frontend/src/components/SchedulersView/SchedulersViewActionMenu.tsx
```diff
@@ -0,0 +1,113 @@
...
+                        }
+                    >
+                        Edit schedule
+                    </Menu.Item>
+                    <Menu.Item
+                        component="button"
+                        role="menuitem"
+                        icon={<IconSquarePlus size={18} />}
+                        onClick={() =>
+                            history.push(getItemLink(item, projectUuid))
```
IrakliJani, line 81: history push is not convenient since it's basically removing a good old link UX. 
try doing this
```
                <Menu.Item
                    component={Link}
                    to={url}
```
where `Link` is from `react-router-dom`

packages/frontend/src/components/SchedulersView/SchedulersViewActionMenu.tsx
```diff
@@ -0,0 +1,113 @@
...
+                            sx={(theme) => ({
+                                ':hover': {
+                                    backgroundColor: theme.colors.gray[1],
+                                },
+                            })}
+                        >
+                            <IconDots size={16} />
+                        </ActionIcon>
+                    </Box>
+                </Menu.Target>
```
IrakliJani, line 63: does it work if you remove Box and all that controllable menu state? (remove `opened={isOpen}` and remove <Box completely

packages/frontend/src/components/SchedulersView/SchedulersView.tsx
```diff
@@ -0,0 +1,252 @@
...
+        ],
+        [users, charts, dashboards, projectUuid, logs, theme],
+    );
+
+    return (
+        <Table className={classes.root} highlightOnHover>
+            <thead>
+                <tr>
+                    {columns.map((column) => {
+                        return (
```
IrakliJani, line 226: you can use inline function without `return (` to remove extra indentation since this is a very simple rendering function without extra logic

IrakliJani (commented):
added some more comments

packages/frontend/src/components/SchedulersView/index.tsx
yasmin-e: if there are no schedulers, there can't be logs so no

yasmin-e: I will fix the statement though

IrakliJani (approved):
🎉 🎉 🎉

PR #5738, opened 2023-06-02
feat: reset current validations and improve types

joaoviana:
Relates to: #5653 

### Description:

* add migration file that deletes entries where `error_type` is `null`
* change `error_type` and `errorType` to required
* remove _stale_ validation results message. 
* remove `dimension_name` - not used anymore

### DIFF
--- a/packages/backend/src/database/migrations/20230602131116_validation-reset-current.ts
+++ b/packages/backend/src/database/migrations/20230602131116_validation-reset-current.ts
@@ -1,0 +1,46 @@
+import { Knex } from 'knex';
+
+const ValidationTableName = 'validations';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(ValidationTableName))) {
+        await knex.schema.createTable(ValidationTableName, (table) => {
+            table.specificType(
+                'validation_id',
+                'INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY',
+            );
+            table
+                .timestamp('created_at', { useTz: false })
+                .notNullable()
+                .defaultTo(knex.fn.now());
+            table
+                .uuid('saved_chart_uuid')
+                .references('saved_query_uuid')
+                .inTable('saved_queries')
+                .onDelete('CASCADE');
+            table
+                .uuid('dashboard_uuid')
+                .references('dashboard_uuid')
+                .inTable('dashboards')
+                .onDelete('CASCADE');
+            table.string('error').notNullable();
+            table
+                .uuid('project_uuid')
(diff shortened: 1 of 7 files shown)


packages/backend/src/database/migrations/20230602131116_validation-reset-current.ts
```diff
@@ -0,0 +1,51 @@
...
+        await knex.schema.alterTable(ValidationTableName, (table) => {
+            table.dropColumn('dimension_name');
+        });
+    }
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(ValidationTableName, (table) => {
+        table.string('dimension_name').nullable();
+    });
```
joaoviana, line 50: I don't think I should add this column back if it's not being used... What do you think @ZeRego ?

ZeRego: The code before was using it so we should add it back.

packages/backend/src/database/migrations/20230602131116_validation-reset-current.ts
```diff
@@ -0,0 +1,51 @@
+import { Knex } from 'knex';
+
+const ValidationTableName = 'validations';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(ValidationTableName))) {
+        await knex.schema.createTable(ValidationTableName, (table) => {
```
ZeRego, line 7: Only handle the case where we update the table. If the table isn't there something really bad happened.

packages/backend/src/database/migrations/20230602131116_validation-reset-current.ts
```diff
@@ -0,0 +1,51 @@
...
+            table.string('chart_name').nullable();
+            table.string('field_name').nullable();
+            table.string('model_name').nullable();
+
+            table.index(['project_uuid']);
+        });
+    } else {
+        await knex(ValidationTableName).whereNull('error_type').del();
+        await knex.schema.alterTable(ValidationTableName, (table) => {
+            table.dropColumn('dimension_name');
```
ZeRego, line 42: we should also change the `error_type` column to not be nullable.

packages/backend/src/database/migrations/20230602131116_validation-reset-current.ts
ZeRego: and we should make the error_type nullable again

joaoviana: thank you! added to down 👍

rephus (approved):

Issue #5725, opened 2023-06-02
Users can go and explore the table that has a validation error


joaoviana:
### Description

Currently users can click on a chart or dashboard validation error and they get redirected to the respective chart/dashboard. 
However, they can't do so for Table/Model validation errors. 

### What problem does this solve?

Links the validation error to the tables view (if the model is broken, they can't view the model), allowing for better continuity in the app.

PR #5742, opened 2023-06-02
feat: let users go to tables when model error

joaoviana:
Closes: #5725

### Description:

When user clicks on model error they get to `/tables` view.

I also made sure the 2 column rows were of equal width


IrakliJani (approved):
✅

Issue #5690, opened 2023-05-31
Clicking on the validator notifications in a preview environment causes the project to change


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/1f693ae435914d289872652704250f82

PR #5749, opened 2023-06-05
fix: invalidate validations query on change of project

joaoviana:
Closes: #5690

### Description:

Invalidate `validations` query when changing active project.


rephus (approved):
nice catch

Issue #5639, opened 2023-05-26
There is a visual indication that a chart/dashboard/table is broken when you're searching for something in the search bar.


joaoviana:
Similar to: https://github.com/lightdash/lightdash/issues/5325

There's a visual indication that a chart/dashboard/table is broken when the user is searching through the project: 

![image](https://github.com/lightdash/lightdash/assets/7611706/906ffde7-5876-471a-a504-77d4beee3527)

joaoviana:
there's one more PR coming for model errors

PR #5757, opened 2023-06-05
feat: add visual indicator to search results

joaoviana:
Closes: #5639


packages/backend/src/models/SearchModel/index.ts
```diff
@@ -72,6 +73,17 @@ export class SearchModel {
                 `${DashboardsTableName}.name`,
                 `${DashboardsTableName}.description`,
                 { spaceUuid: 'space_uuid' },
+                {
+                    validationErrors: this.database.raw(`
```
owlas, line 77: I don't really know how the performance of this is going to be and whether postgres will:
1. filter everything first and then run this subquery on only a small number of rows
2. first compute this field (loads of lookups on another table) and then filter afterwards

To be safe, could we move this to another query in the same transaction so it would go:
1. Do search to find the relevant charts
2. Do seperate query to get validation errors for returned chart ids from step 1
3. Return them together

packages/backend/src/models/SearchModel/index.ts
```diff
@@ -115,6 +127,17 @@ export class SearchModel {
                 { spaceUuid: 'space_uuid' },
                 { chartType: 'saved_queries_versions.chart_type' },
                 { chartConfig: 'saved_queries_versions.chart_config' },
+                {
```
owlas, line 130: same as above

packages/common/src/types/search.ts
```diff
@@ -3,17 +3,31 @@ import { Table } from './explore';
...
 
 export type SpaceSearchResult = Pick<Space, 'uuid' | 'name' | 'uuid'>;
 export type DashboardSearchResult = Pick<
     Dashboard,
     'uuid' | 'name' | 'description' | 'spaceUuid'
->;
+> & {
+    validationErrors: Pick<
+        ValidationErrorDashboardResponse,
+        'error' | 'createdAt' | 'validationId'
```
owlas, line 18: `createdAt` isn't used in the rendering, will it be?

owlas (changes requested):
Subquery in the postgres query could be a big performance problem later. Can we run this as two separate queries instead? It'll be easier to measure their individual performance too

owlas (changes requested):
Subquery in the postgres query could be a big performance problem later. Can we run this as two separate queries instead? It'll be easier to measure their individual performance too

joaoviana:
> Subquery in the postgres query could be a big performance problem later. Can we run this as two separate queries instead? It'll be easier to measure their individual performance too

Thanks @owlas , will do that instead! Great point 👍

packages/common/src/types/search.ts
joaoviana: Yea, good point, I'm only keeping `validationId`s for now then 👍

owlas (approved):

packages/frontend/src/components/NavBar/GlobalSearch/index.tsx
```diff
@@ -227,6 +253,7 @@ const GlobalSearch: FC<GlobalSearchProps> = ({ projectUuid }) => {
 
             <MantineProvider inherit theme={{ colorScheme: 'light' }}>
                 <SpotlightProvider
+                    zIndex={200}
```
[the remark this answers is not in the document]
IrakliJani: you can leave a FIXME: mark to get rid of it after we remove blueprint

packages/frontend/src/components/NavBar/GlobalSearch/index.tsx
```diff
@@ -165,38 +168,61 @@ const GlobalSearch: FC<GlobalSearchProps> = ({ projectUuid }) => {
...
-                    properties: {
-                        action: 'result_click',
-                    },
-                });
+        return items.map<SpotlightAction>((item) => {
+            const isSearchItemWithValidationError =
+                item.item &&
+                ['dashboard', 'saved_chart'].includes(item.type) &&
+                'validationErrors' in item.item &&
+                item.item.validationErrors?.length > 0;
```
IrakliJani, line 176: this can be a `itemHasValidationError(item: ItemType)` and can be extracted outside of the component completely

packages/frontend/src/components/NavBar/GlobalSearch/index.tsx
```diff
@@ -132,6 +133,8 @@ const GlobalSearch: FC<GlobalSearchProps> = ({ projectUuid }) => {
     const { track } = useTracking();
     const project = useProject(projectUuid);
 
+    const canUserManageValidation = useValidationUserAbility(projectUuid);
```
IrakliJani, line 136: I like that! its just not clear from the name `useValidationUserAbility` that its about `Manage` permission

packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
```diff
@@ -42,3 +49,56 @@ export const SearchIcon: FC<SearchIconProps> = ({ searchItem }) => {
             return <IconBox icon={IconAppWindow} color="gray.7" />;
     }
 };
+
+export const SearchIconWithIndicator: FC<{
+    searchResult: SearchItem;
+    projectUuid: string;
+    canUserManageValidation: boolean;
+}> = ({ searchResult, projectUuid, canUserManageValidation }) =>
+    searchResult.item && 'validationErrors' in searchResult.item ? (
```
IrakliJani, line 58: nit (can ignore): I personally find it hard to read ternary operator like this.
what I usually do is have an explicit "guard"

```
if (searchResult.item && 'validationErrors' in searchResult.item) return null

return component stuff

IrakliJani (commented):
looks great! left few comments

packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
joaoviana: Cool, yeah, I can see why but since this is a short condition I'll keep it as is 👍

packages/frontend/src/components/NavBar/GlobalSearch/index.tsx
joaoviana: hmmm I see! I would like to keep it because it matches the current usage of `user.ability`

packages/frontend/src/components/NavBar/GlobalSearch/index.tsx
joaoviana: that's great, will change that!

IrakliJani (approved):

Issue #5547, opened 2023-05-19
Hide column totals for table calculations


owlas:
### Description

In this issue we present a better way to handle table calculations column totals: https://github.com/lightdash/lightdash/issues/4067

But while we work on this it might be better to remove the column totals completely for table calculations, or give users the option of doing this since it can be confusing for users.

Community slack thread: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1684239528791709?thread_ts=1684230133.042719&cid=C03MG2VTR08

CVamsi27:
Hello, I would like to work in this.

Issue #5540, opened 2023-05-19
Remove column and row totals for table calculations and metrics of certain types


rephus:
### Description

currently it seems the totals are in all cases SUM of values in the table. However, if the values in a metric are a count distinct, then having the total display a SUM instead of a count distinct can lead to confusion for end users.

E.g. my total users by month could be 10, 10, 10, 10, 10 for the months of Jan to May - in total that could be the same 10 users but currently in Lightdash the total would show 50 which would be misleading. I just got off a call where this was the kind of confusion a user was facing so wanted to report. 

### Acceptance criteria

- [ ] only apply `totals` to metrics that have: sum, count or number types
- [ ] don't show totals for metrics with average, count_distinct, average, median, percentile
- [ ] don't show totals for table calculations https://github.com/lightdash/lightdash/issues/5547

Eng Note: update the logic in `isSummable()`

erikgregorywebb:
This is becoming a big source of confusion for our end-users. Please prioritize!

PR #5764, opened 2023-06-06
fix: total exclusions for certain types

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/5540
Closes: https://github.com/lightdash/lightdash/issues/5547

### Description:

- [x] only apply totals to metrics that have: sum, count or number types
- [x] don't show totals for metrics with average, count_distinct, average, median, percentile
- [x] don't show totals for table calculations

### DIFF
--- a/packages/frontend/src/hooks/useColumnTotals.ts
+++ b/packages/frontend/src/hooks/useColumnTotals.ts
@@ -4,10 +4,7 @@ import {
     Field,
     FieldId,
     getItemId,
-    isAdditionalMetric,
     isDimension,
-    isField,
-    isMetric,
     isTableCalculation,
     Item,
     MetricType,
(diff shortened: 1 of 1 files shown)


TuringLovesDeathMetal (approved):
Nice! Looks good - tested it out a bit and it matched all of the types I would expect :)

packages/frontend/src/hooks/useColumnTotals.ts
```diff
@@ -18,20 +21,24 @@ type Args = {
...
 };
 
-export const isSummable = (item: Field | TableCalculation) => {
-    if (isField(item) || isAdditionalMetric(item)) {
+export const isSummable = (item: Item) => {
+    if (isTableCalculation(item)) {
+        return false;
+    }
+
+    if (isField(item) || isMetric(item) || isAdditionalMetric(item)) {
```
owlas, line 29: What's the logic for `isField` vs `isMetric` ? I think if `isField` is true then `isMetric` is always true since it's a subset of fields.

packages/frontend/src/hooks/useColumnTotals.ts
```diff
@@ -18,20 +21,24 @@ type Args = {
     itemsMap: Record<FieldId, Field | TableCalculation>;
 };
 
-export const isSummable = (item: Field | TableCalculation) => {
-    if (isField(item) || isAdditionalMetric(item)) {
+export const isSummable = (item: Item) => {
+    if (isTableCalculation(item)) {
```
owlas, line 25: Do we need this because `isTableCalculation` also triggers `isField` ?

packages/frontend/src/hooks/useColumnTotals.ts
```diff
@@ -18,20 +21,24 @@ type Args = {
...
-            MetricType.COUNT_DISTINCT,
             MetricType.SUM,
         ];
         const isPercent = item.format === 'percent';
         const isDatePart = isDimension(item) && item.timeInterval;
         return numericTypes.includes(item.type) && !isPercent && !isDatePart;
     }
-    return true;
+
+    return false;
```
owlas, line 41: what cases are caught by this?

owlas (commented):
questions about the logic

packages/frontend/src/hooks/useColumnTotals.ts
IrakliJani: none, just makes the return type explicit

owlas: shall we leave it and not change it in that case?

packages/frontend/src/hooks/useColumnTotals.ts
owlas: You can replace two if statements and all the conditions inside them with:

```
if (tableCalculation) {
  return false 
}
// otherwise it's a field/metric/dimension/additionalMetric...
// do other checks
```

Issue #5762, opened 2023-06-06
Running lightdash on codespaces doesn't work 


rephus:
### Description

![image](https://github.com/lightdash/lightdash/assets/1983672/6575e727-0c52-4797-9bae-c4e92536f6ec)

rephus:
From codespaces:


The Compose file '/var/lib/docker/codespacemount/workspace/lightdash/docker/docker-compose.dev.yml' is invalid because:
services.db-dev.ports contains an invalid type, it should be a number, or an object
services.lightdash-dev.ports contains an invalid type, it should be a number, or an object

PR #5765, opened 2023-06-06
fix: codespaces config

rephus:
Closes: #5762

### Description:

How to test

- go to this branch
- Click code
- Run codespaces on this branch 

image


rephus:
I tried adding our enf_file to docker-compose or devcontainer, but I can't make it work, in any case I think this is very specific so I don't think is good to reuse the same file (eg: if we want to add a new service, we'll have to modify this file anyway)

owlas (approved):
Thanks for fixing, looks awesome

Issue #5748, opened 2023-06-05
Remove `show error logs` button from the navbar


TuringLovesDeathMetal:
### Description

Right now, when you trigger an error in Lightdash, you're able to see the logs for errors using the `show error logs` button in the nav bar. 

![image](https://github.com/lightdash/lightdash/assets/31848148/0e14762a-b7b7-4ac9-9ca5-a52b22821a0f)

This button only appears if you trigger an error. I don't think anyone actually uses it though, and it's cluttering the nav bar...so we should remove it. 

### Acceptance criteria

- [ ] remove the `show error logs` button from the nav bar. 
- [ ] If a user triggers an error, they're shown the header with the error information and the option to dismiss the error. But, if they close that header, they are no longer shown the `show error logs` button.

PR #5775, opened 2023-06-06
fix: remove show error logs button from navbar

yasmin-e:
Closes: #5748 

### Description:
https://github.com/lightdash/lightdash/assets/67699259/ebe97f68-67e6-4652-a552-ed6a7a851a62

### DIFF
--- a/packages/frontend/src/components/ErrorLogsDrawer.tsx
+++ b/packages/frontend/src/components/ErrorLogsDrawer.tsx
@@ -58,7 +58,13 @@ const ErrorCard: React.FC<ErrorLogEntry & { onDismiss: () => void }> = ({
 );
 
 export const ErrorLogsDrawer = () => {
-    const errorLogs = useErrorLogs();
+    const {
+        errorLogs,
+        setAllLogsRead,
+        setErrorLogsVisible,
+        errorLogsVisible,
+        deleteErrorLogEntry,
+    } = useErrorLogs();
     return (
         <Drawer
             autoFocus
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/ErrorLogsDrawer.tsx
owlas, line 1: Why did this file change? It's not needed right?

packages/frontend/src/components/NavBar/index.tsx
```diff
@@ -114,11 +113,12 @@ const NavBar = memo(() => {
                 <Box sx={{ flexGrow: 1 }} />
 
                 <Group sx={{ flexShrink: 0 }}>
+                    {errorLogs.length > 0
```
owlas, line 116: This logic should be in the hook not in the rendering function.

Here it should just be:

```
errorLogs.length > 0 ? <render> : null
```

any logic related to changing the logs themselves (i.e. setting visible) should be in the hook

owlas (changes requested):
move logic to the hook

packages/frontend/src/components/ErrorLogsDrawer.tsx
yasmin-e: just some code clean up

packages/frontend/src/components/NavBar/index.tsx
yasmin-e: Setting visible is the trigger to open the drawer and show the errors in the UI, it doesn't modify the logs... previously there was a button that triggered the opening of the drawer to show the errors, now we're doing it based on having unread logs or not.
I tried to store this "logic" in the hook which is a provider actually 😅  and it gave me a "Too many re-renders. React limits the number of renders to prevent an infinite loop." error.
I think it's fine if it stays here.

TuringLovesDeathMetal:
Looks great to me! 😚 👌

IrakliJani (approved):
code looks good! can you reopen that drawer after you close it? I was thinking if notification menu (where we show validation errors) is the proper place to move

owlas (approved):

yasmin-e:
> code looks good! can you reopen that drawer after you close it? I was thinking if notification menu (where we show validation errors) is the proper place to move

@TuringLovesDeathMetal what do you think?

Issue #5777, opened 2023-06-06
Fix z-index issues with validation error indicator under overlays


joaoviana:
### Description

When users see a modal they're on a page with Resource view icons (for example, on the Home Page), and some items have validation errors, then the alert icons _go over_ the overlay

### To Reproduce

Screenshot:
![image](https://github.com/lightdash/lightdash/assets/7611706/1e5400ea-6069-4bd4-b6bb-2b894ccf3eec)

Steps to reproduce: 
1. Go to Home page
2. Have items with validation errors
3. Trigger a modal render
4. See Alert icons on top of the overlay

Thanks @rephus for reporting this!

PR #5779, opened 2023-06-06
feat: add z-index 20 to Indicator in theme

joaoviana:
Closes: #5777

### Description:

Add `z-index` `20` to root `Indicator`, similar to the `Tooltip` root styling approach. 

Example with it working: 
Screenshot 2023-06-06 at 15 17 28

Previously, without the overriding, the Indicator had a `z-index` set to `100`, which we don't need. 
Once we fully migrate to using Mantine UI then we can set our own `z-index` scale, if necessary.


IrakliJani (approved):
👍

Issue #3379, opened 2022-10-03
Measure Lightdash usage using query tags


owlas:
### Description

I want to know which users/dashboards/projects are running the most queries / most expensive queries so that I can answer:

 - which dashboards are costing me the most money?
 - which users are using lightdash the most?
 - which users aren't using lightdash?

We could do this using tagging when we execute SQL on the data warehouse. I don't think this is a complete replacement for full event-driven analytics for Lightdash BUT I think it would be the best way to associate Lightdash queries with warehouse costs.

We have a few options: 

 - **generic warehouse**: add a sql comment at the top of the query
 - **bigquery**: use job labels: https://cloud.google.com/bigquery/docs/adding-labels#job-label
 - **snowflake**: use snowflake query_tag: https://docs.snowflake.com/en/sql-reference/parameters.html#query-tag

I think we could store an object as a json with the following structure:

```json
{
  "lightdash_version": "0.279.1",
  "user_id": "6ccdac34-633b-4671-96cb-367653084652",
  "organisation_id": "6ccdac34-633b-4671-96cb-367653084652",
  "project_id": "6ccdac34-633b-4671-96cb-367653084652",
  "content_trigger": "dashboard", # can be 'saved_chart', 'explore_view' or 'sql_runner',
  "content_id": "dashboard_id_123", # this is the UUID for the saved chart or dashboard. This is empty for Explore view and SQL runner queries
  "query_id": "6ccdac34-633b-4671-96cb-367653084652",
}
```

PR #5784, opened 2023-06-06
feat: add query tags

owlas:
Closes: #3379

### Bigquery

Example of labels persisting on biquery job (fetched using `bq` cli)

CleanShot 2023-06-06 at 16 51 27@2x

### Snowflake

CleanShot 2023-06-09 at 22 12 22@2x

### Redshift / databricks / trino / postgres

Appends to sql

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -85,7 +85,18 @@ import { buildQuery } from '../../queryBuilder';
 import { compileMetricQuery } from '../../queryCompiler';
 import { ProjectAdapter } from '../../types';
 import { runWorkerThread, wrapSentryTransaction } from '../../utils';
+import { VERSION } from '../../version';
 import { hasSpaceAccess } from '../SpaceService/SpaceService';
+
+type RunQueryTags = {
+    projectUuid?: string;
+    userUuid?: string;
+    organizationUuid?: string;
+    chartUuid?: string;
+};
+const DEFAULT_QUERY_TAGS = {
+    lightdashVersion: VERSION.replaceAll('.', '-'),
+};
 
 type ProjectServiceDependencies = {
     projectModel: ProjectModel;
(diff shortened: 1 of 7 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -85,8 +85,19 @@ import { buildQuery } from '../../queryBuilder';
...
 import { ProjectAdapter } from '../../types';
 import { runWorkerThread, wrapSentryTransaction } from '../../utils';
+import { VERSION } from '../../version';
 import { hasSpaceAccess } from '../SpaceService/SpaceService';
 
+type RunQueryTags = {
+    projectUuid?: string;
+    userUuid?: string;
+    organizationUuid?: string;
+    chartUuid?: string;
```
joaoviana, line 95: From reading the issue, what about dashboards? I guess that dashboards contain charts, so that's ok. 
Just wondering because on the issue it mentioned the `content_trigger` 🤔

joaoviana (approved):
:shipit:

packages/backend/src/services/ProjectService/ProjectService.ts
owlas: Yes good point, I should have proposed an update to the spec. To answer the questions:

- Dashboards: when we run a query the API doesn't actually know the dashboard context when the query runs! We could pass this from the frontend but I didn't think it was worth the dependency until somebody asks for it explicitly
- content_trigger: we could definitely introduce this but I didn't want to commit to the trigger names yet (without dashboards etc.).

Actually since the issue we should also add scheduled delivery id possible 🤔 

But the idea is to ship this and see what users find useful and we can extend it!

joaoviana: Sure @owlas, it makes sense to wait until someone asks for it, and this is a great base for what might come next, until new requirements arise! 
Go for it ✌️

Issue #5291, opened 2023-04-29
Docker compose ends up with error


patrikbraborec:
### Description

I followed this [tutorial](https://docs.lightdash.com/self-host/self-host-lightdash-docker-compose/) in the docs. Unfortunately, it ends up with error `error: Error starting standalone scheduler worker getaddrinfo ENOTFOUND db-dev`. I think that issue is in `.env` file: https://github.com/lightdash/lightdash/blob/main/.env#L7.

### To Reproduce

Just follow the docs: https://docs.lightdash.com/self-host/self-host-lightdash-docker-compose/.

### App Version

v0.525.0

ZeRego:
@all-contributors add @patrikbraborec for bug

ZeRego:
Change suggestions https://github.com/lightdash/lightdash/pull/5292#issuecomment-1531116172

> At the moment .env shouldn't be used for production, only for development.
> The [self-hosting docs](https://docs.lightdash.com/self-host/self-host-lightdash-docker-compose/) don't mention using the .env file. @patrikbraborec Was it picked by default or you assumed you should use it?
> 
> Either way, the .env file should be set with production like variables.
> 
> Proposed changes:
> 
> rename .env to .env.development
> update package.json and contribution docs to reference new file name and .env.development.local
> create .env with main env vars with empty values ( check .env.fast-install ) and all the rest of possible env vars commented out
> update self-hosting docs to configure and use the .env file

PR #5787, opened 2023-06-07
fix: update env files

rephus:
Closes: #5291


.env
```diff
@@ -1,38 +1,44 @@
...
-RUDDERSTACK_DATA_PLANE_URL=https://analytics.lightdash.com
-
-DBT_DEMO_DIR=/usr/app/examples/full-jaffle-shop-demo
+LIGHTDASH_LOG_LEVEL=debug
+LIGHTDASH_QUERY_MAX_LIMIT=5000
 
-SENTRY_DSN=''
 
-NODE_ENV=development
+NODE_ENV=production
```
IrakliJani, line 20: do we need to change this?

IrakliJani (approved):
looks good to me!

we all have to rename `.env.local` to `.env.development.local` right?

rephus:
> we all have to rename .env.local to .env.development.local right?

Only if you run docker and follow the contributing guide, 
if you don't use docker, this will not affect you 
if you use docker compose, and point to the right file with --env-file then you don't need to change anything

.env
rephus: Jose said this .env should represent production, so I think having NODE_ENV=production on .env makes sense to me

PR #5788, opened 2023-06-07
docs: updates reference docs titles

TuringLovesDeathMetal:
updates reference docs titles and moves things around in the sidebar


owlas (changes requested):
can we override labels in sidebar instead or renaming the docs

Issue #5793, opened 2023-06-07
pivot table totals should not include dimension fields


IrakliJani:
- numeric dimensions are being included in the summarisation
- it does not sum up field values properly

- [ ] exclude field values from summarisation in pivot table totals

![image](https://github.com/lightdash/lightdash/assets/962095/7e99e8b1-24d8-47d4-90f8-29d221402cc6)

PR #5794, opened 2023-06-07
fix: exclude dimensions from summable metric field ids

IrakliJani:
Closes: #5793 

before:
CleanShot 2023-06-07 at 16 55 14@2x

after:
CleanShot 2023-06-07 at 16 54 35@2x

### DIFF
--- a/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
+++ b/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
@@ -235,10 +235,18 @@ const useTableConfig = (
         const summableMetricFieldIds = selectedItemIds?.filter((fieldId) => {
             const field = getField(fieldId);
 
-            return (
-                isSummable(field) &&
-                !(hiddenMetricFieldIds || []).includes(fieldId)
-            );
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -235,10 +235,18 @@ const useTableConfig = (
...
+            }
+
+            if (
+                hiddenMetricFieldIds &&
+                hiddenMetricFieldIds.includes(fieldId)
+            ) {
+                return false;
+            }
+
+            return isSummable(field);
```
yasmin-e, line 249: nitpicking but what about an inline expression like this ?
```
return !isDimension(field) 
&& (!hiddenMetricFieldIds || !hiddenMetricFieldIds.includes(fieldId)) 
&& isSummable(field)
```

yasmin-e (approved):

TuringLovesDeathMetal (approved):
Nice! Looks good to me :) I'm not actually sure if this worked before, so let's see if people ask for it back 😅

PR #5818, opened 2023-06-08
docs: adjust instructions to use new .env.development file

elton-okawa:
### Description:

Update instructions of using new `.env.development.local` based on `with docker-compose` from #5787 to `without docker instructions`


rephus (approved):
Thanks !

Issue #5816, opened 2023-06-08
validation errors should not disappear if the content is deleted


TuringLovesDeathMetal:
### Description

Right now, if there is an error with a chart or dashboard, and you delete that chart or dashboard, then the error disappears from the validator. 

This feels inconsistent/a bit confusing compared to how the rest of the errors behave. 

Instead, it would be great if the error persisted, even if you deleted the chart or dashboard. Then to make it disappear, you'd either need to dismiss the item, or re-run the validator.

### Acceptance criteria

- [ ] errors in the validator persist, even if the content is deleted
- [ ] the errors are removed if a user dismisses them
- [ ] the errors are removed if the content is deleted, then the validator is re-run.
- [ ] if the user clicks on the chart/dashboard link for the error after they have deleted the content, then they are linked to the chart/dashboard but see an error "chart no longer exists" or "dashboard no longer exists" (the same as now when you open a link to a deleted chart/dashboard)

PR #5839, opened 2023-06-09
feat: keep validation if content deleted

rephus:
Closes: #5816

### Description:

image

### Acceptance criteria

- [x] errors in the validator persist, even if the content is deleted
- [x] the errors are removed if a user dismisses them
- [x] the errors are removed if the content is deleted, then the validator is re-run.
- [ ] if the user clicks on the chart/dashboard link for the error after they have deleted the content, then they are linked to the chart/dashboard but see an error "chart no longer exists" or "dashboard no longer exists" (the same as now when you open a link to a deleted chart/dashboard) 
    - we can't link to the chart because we no longer have the uuid, instead we don't link to anything

### DIFF
--- a/packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
+++ b/packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
@@ -47,17 +47,10 @@ const getLinkToResource = (
     return;
 };
 
-const isDeleted = (validationError: ValidationResponse) => {
-    if (isChartValidationError(validationError) && !validationError.chartUuid)
-        return true;
-    if (
-        isDashboardValidationError(validationError) &&
-        !validationError.dashboardUuid
-    )
-        return true;
-
-    return false;
-};
+const isDeleted = (validationError: ValidationResponse) =>
+    (isChartValidationError(validationError) && !validationError.chartUuid) ||
+    (isDashboardValidationError(validationError) &&
+        !validationError.dashboardUuid);
 
 const Icon = ({ validationError }: { validationError: ValidationResponse }) => {
     if (isChartValidationError(validationError))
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
```diff
@@ -35,15 +35,30 @@ const getLinkToResource = (
...
+        return true;
+    if (
+        isDashboardValidationError(validationError) &&
+        !validationError.dashboardUuid
+    )
+        return true;
+
+    return false;
+};
+
```
joaoviana, line 61: could do this!
```suggestion
const isDeleted = (validationError: ValidationResponse) => 
    (isChartValidationError(validationError) && !validationError.chartUuid) ||
    (isDashboardValidationError(validationError) && !validationError.dashboardUuid);

```

packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
```diff
@@ -119,23 +134,26 @@ const TableValidationItem = forwardRef<
...
-                                            Last edited by{' '}
-                                            <Text span fw={500}>
-                                                {validationError.lastUpdatedBy}
-                                            </Text>
-                                        </>
-                                    ) : null}
-                                </Text>
-                            )}
+                                isDashboardValidationError(validationError)) &&
+                                !isDeleted(validationError) && (
```
joaoviana, line 138: I wonder if here we can just have `!isDeleted(validationError)` 🤔

joaoviana (approved):
Looks good! just 2 non-blocking changes requested 👍

packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
rephus: mm,m I don't think so, because you want to know if it is a chart or a dashboard, and not deleted. These views don't work with tables

joaoviana: got it, that's what I thought!

TuringLovesDeathMetal (approved):
nice! looks good :)

Issue #5470, opened 2023-05-12
create a ContextProvider for `Pinned Items`


yasmin-e:
### Description
Currently, the business logic for pinned items is scattered throughout 3 layers:

- `pages > Home.tsx`

fetching data (pinned items)

- `components > PinnedItemsPanel > index.tsx`

checking permissions, rendering data, empty states 

- `components > ResourceViewGrid > index.tsx`

re-ordering items (calling useReorder hook), handling drag and drop

This leads to bugs like #5424 and will make it harder to scale this feature in the future.
We think pinned items could benefit from a contextprovider so we can pass the variables needed to parent components without prop drilling or exposing unecessary context (ex. `pinnedItemsProps` and `useReorder` inside `ResourceViewGrid`).
### What problem does this solve?
DevX, feature scalability

Issue #5847, opened 2023-06-11
Split `ResourceView` into more components


elton-okawa:
### Description

Current `ResourceView` component can have some variations and each one requires unrelated set of properties because they are not used at the same time:
- it can have `tabs`, `header` or `none`
- it can be `empty`, `list` or a `grid`

```
const ResourceView = ({
    view = ResourceViewType.LIST,
    items,
    maxItems,
    tabs,
    gridProps = {},
    listProps = {},
    headerProps = {},
    emptyStateProps = {},
    hasReorder = false,
}) => {}
```

I think that we could split it in composable components:
- `ResourceViewContainer`: container style and provides context - this context can provide actions to manipulate state (e.g sorting)
- `ResourceViewHeader`: header
- `ResourceViewTabs`: tabs
- `ResourceViewEmpty`: already exists
- `ResourceViewList`: already exists
- `ResourceViewGrid`: already exists

The usage would be like:
```
<ResourceViewContainer>
   <ResourceViewHeader />
   <ResourceViewGrid />
</ResourceViewContainer>
```

```
<ResourceViewContainer>
   {data ? <ResourceViewEmpty /> : <ResourceViewList />}
</ResourceViewContainer>
```

It's just a draft proposal 📝 , totally open to feedback! 😄 

### What problem does this solve?

devx and maintainability

PR #5848, opened 2023-06-11
refactor: add pinned items context

elton-okawa:
Closes: #5470

### Description:

I'm not sure if I implemented the way the issue envisioned.
One downside I noticed is that every time a `ResourceView` type `Grid` is used, it must be wrapped by `PinnedItemsProvider` even if reordering is not enabled. (it happens on `Spaces` page)

Maybe #5847 could solve the prop drilling problem and bring more bonus. I'm opening this PR as draft to discuss it

### DIFF
--- a/packages/frontend/src/components/PinnedItemsPanel/index.tsx
+++ b/packages/frontend/src/components/PinnedItemsPanel/index.tsx
@@ -1,4 +1,3 @@
-import { subject } from '@casl/ability';
 import {
     DashboardBasicDetails,
     PinnedItems,
@@ -7,42 +6,32 @@ import {
 } from '@lightdash/common';
 import { Card, Group, Text } from '@mantine/core';
 import { IconPin } from '@tabler/icons-react';
-import React, { FC } from 'react';
-import { useApp } from '../../providers/AppProvider';
+import { FC } from 'react';
+import { usePinnedItemsContext } from '../../providers/PinnedItemsProvider';
 import MantineIcon from '../common/MantineIcon';
 import MantineLinkButton from '../common/MantineLinkButton';
 import ResourceView, { ResourceViewType } from '../common/ResourceView';
 
 interface Props {
-    data: PinnedItems;
-    projectUuid: string;
-    pinnedListUuid: string;
-    organizationUuid: string;
+    pinnedItems: PinnedItems;
     dashboards: DashboardBasicDetails[];
     savedCharts: SpaceQuery[];
 }
 
 const PinnedItemsPanel: FC<Props> = ({
-    data,
-    projectUuid,
-    pinnedListUuid,
-    organizationUuid,
+    pinnedItems,
     dashboards,
     savedCharts,
 }) => {
-    const { user } = useApp();
-    const userCanManagePinnedItems = user.data?.ability.can(
-        'manage',
-        subject('PinnedItems', { organizationUuid, projectUuid }),
-    );
+    const { userCanManage } = usePinnedItemsContext();
 
     const enablePinnedPanel = dashboards.length + savedCharts.length > 0;
 
-    return data && data.length > 0 ? (
+    return pinnedItems && pinnedItems.length > 0 ? (
         <ResourceView
-            items={data}
+            items={pinnedItems}
             view={ResourceViewType.GRID}
-            hasReorder={userCanManagePinnedItems}
+            hasReorder={userCanManage}
             gridProps={{
                 groups: [
                     [ResourceViewItemType.SPACE],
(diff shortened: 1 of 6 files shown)


packages/frontend/src/pages/Spaces.tsx
```diff
@@ -79,27 +84,32 @@ const Spaces: FC = () => {
...
-                                </Button>
-                            ) : undefined,
-                    }}
-                    pinnedItemsProps={{ projectUuid, pinnedListUuid: '' }}
-                />
+                <PinnedItemsProvider
+                    projectUuid={projectUuid}
+                    organizationUuid={user.data?.organizationUuid ?? ''}
+                    pinnedListUuid={project.data?.pinnedListUuid ?? ''}
+                >
```
IrakliJani, line 91: this is not necessary since Spaces page does not have a reorder enabled and we also get if space is pinned or not from the data that is inside `spaces`

IrakliJani: inside `ResourceViewGrid` you use `usePinnedItemsContext` which throws an error if you don't have a `PinnedItemsProvider`. we can `usePinnedItemsContext` return undefined and not throw an error instead

packages/frontend/src/components/PinnedItemsPanel/index.tsx
```diff
@@ -7,42 +6,32 @@ import {
...
+    pinnedItems,
     dashboards,
     savedCharts,
 }) => {
-    const { user } = useApp();
-    const userCanManagePinnedItems = user.data?.ability.can(
-        'manage',
-        subject('PinnedItems', { organizationUuid, projectUuid }),
-    );
+    const { userCanManage } = usePinnedItemsContext();
```
IrakliJani, line 26: can you name this `userCanManagePinnedItems` to make it more explicit?

packages/frontend/src/providers/PinnedItemsProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return <Context.Provider value={value}>{children}</Context.Provider>;
+};
+
+export const usePinnedItemsContext = (): PinnedItemsContext => {
+    const context = useContext(Context);
+    if (!context) {
+        throw new Error(
+            'usePinnedItemsContext must be used within a PinnedItemsContext',
+        );
+    }
```
IrakliJani, line 54: this should not throw an error when we don't have a reordering enabled in some pages (aka: not have a `PinnedItemsProvider` wrapper - in Spaces page for example)

IrakliJani (changes requested):
great work, left some comments. let me know if you need any help and thanks for contributing to lightdash ⚡ 🙌

yasmin-e:
Hey @elton-okawa, great work cleaning up Pinned Items!! 🙌🏼  

About #5847, we discussed it internally and a lot of upcoming changes will be happening to `ResourceView`. 

The more we think about it, the more we realise it's not fit to be a reusable component as it solves a very specific use case that's too different from our other table features. We share its styling in `useTableStyles()` but it doesn't make sense for us to further abstract this component.

I'm going to go ahead and close this ticket for now, thank you for taking the time to come up with a solution for this... I guess you were missing crucial context 😅 

Let me know if you have any more questions about our comments on this PR or if you need help!
Thanks again 🎉 ⚡

elton-okawa:
Hi @yasmin-e thanks for letting me know your thoughts about the other issue! 😄

packages/frontend/src/providers/PinnedItemsProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const Context = createContext<PinnedItemsContext>({
+    userCanManagePinnedItems: false,
+    reorderItems: () => {
+        throw new Error(
+            'Default value, did you wrap the component with PinnedItemsProvider?',
+        );
+    },
+});
```
elton-okawa, line 25: Hi @IrakliJani thanks for the suggestions! I've implemented all of them with a single observation:

Originally, not throwing an error when the component is not wrapped by `PinnedItemsProvider` made `usePinnedItemsContext` return `PinnedItemsContext | null` which affects usage on destructuring:

```
const { reorderItems } = usePinnedItemsContext();
// cannot destructure because return might be null
```

I thought it would be better to add a default value for the context and throw an error if we inadvertently call its methods without wrapping with provider instead of silently not calling it like `context?.reorderItems(...)`.
Doing this, we can destructure values and be explicit when the provider is missing.

What do you think? 🤔

IrakliJani (approved):

IrakliJani:
@elton-okawa all tests are passing, works great. thanks!

Issue #4523, opened 2023-02-20
Define custom metrics with filters in the UI


owlas:
### Description

Sometimes I want my users to have the ability to create custom metrics in the UI with filters. The current custom metrics only allow me to have simple metrics.

This saves the data team some time defining additional metrics as code.

### Acceptance criteria

Miro board with sketch: https://miro.com/app/board/uXjVPlE0Ow8=/?moveToWidget=3458764546742412818&cot=14

- [ ] When a user creates a custom metric, they are shown a modal. In this modal, you can:
    - [ ] You can update the name of a custom metric
    - [ ] You can add filters to a custom metric
- [ ] The default name for the custom metric is the same as we have now
- [ ] By default, the filters toggle is toggled closed
- [ ] The filtering options use the same filter UX/UI as the explore filter. 
- [ ] There is a toast confirmation once the user hits the 'create' CTA (in the bottom right, like when someone creates/saves a dashboard successfully)
- [ ] #5972

### Documentation

- [ ] add docs here: https://docs.lightdash.com/guides/how-to-create-metrics#adding-custom-metrics-in-the-explore-view
- [ ] add a section called `adding filters to your custom metrics` after `creating a custom metric from a dimension` and before `Custom metrics are different from the metrics defined in your YAML` 
- [ ] Let me know if you need help with creating a screenshot that shows the flow going from `create metric` --> `add filter` (I use Miro!)

### Analytics events

- [ ] add parameter to `query_executed` event which is `custom_metric_filter_count`. This counts the total number of custom metrics which have a filter included.

PriPatel:
Option 1 makes most sense for discoverability! I would add a toast confirmation once the user hits the 'create' CTA. That's about it!

tnightengale:
Hi, would love to this feature! Could unblock using Lightdash with a major client I working with!

PR #5865, opened 2023-06-12
feat: allow definition of custom metrics with filters

joaoviana:
Closes: #4523 

### Description:

Adds support for filters when creating Custom Metrics. 

Users can now Create and Edit a custom metric label when adding it through the UI. 
Users can also add filters (only Dimensions) to a Custom Metric. 
Users can also use that Custom Metric column to view Underlying Data. 

Screen recording:

https://github.com/lightdash/lightdash/assets/7611706/570fd5f2-165b-49e7-acf0-3eae800e5d37

Changes to note: 
* Before we were sending a field reference (`tablename. fieldname`) but calling it `fieldId` when calling `parseFilters`. A `fieldId` is of the following format: `tablename_fieldname`, so the chosen property name was incorrect.
* `AdditionalMetric` type has type `filters: MetricFilterRule` -> similar to `FilterRule` but instead of `target.fieldId` we now have `target.fieldRef`, to accommodate the compilation of queries in the exploreCompiler. When getting the output of `getParseReference`, we pass the `fieldRef` (of format `tablename.fieldname`) so that it does the splitting correctly. 
* Added migration file that adds `uuid`, `filters`, and `base_dimension_name` columns; also ensures `uuid` column is filled for other rows that are `null`. 
 
What does an AdditionalMetric with filters look like:
* added `uuid` -> easy to track what custom metric to edit/identify
* added `filters` -> `MetricFilterRule[]` that gets JSON stringified when sent to DB 
* added `base_dimension_name` -> keep track of base dimension to ensure subsequent edits to custom metric keep the dimension used as a base. 

Notes:
* Edit additional metric action has to know previous vs. new name of the custom metric to ensure it replaces whatever is in `state.unsavedChartVersion.metricQuery`
* Created hook `useFieldsWithSuggestions` that is used when providing data to the `FiltersProvider`. 
* Removed `as` - renaming import - of enum `ConditionalOperator` because it was breaking tsoa routes generation.

TODO (on separate PR)
* [x] - documentation
* [ ] - analytics events - can't work out where actual change needs to be made from reading GH issue!
* [x] - show toast when creating Custom Metric through modal


packages/common/src/types/filter.ts
```diff
@@ -1,9 +1,4 @@
-import {
-    ConditionalOperator as FilterOperator,
```
joaoviana, line 2: tsoa doesn't like using enums+`as` keyword

packages/common/src/types/filter.ts
```diff
@@ -1,9 +1,4 @@
...
-    ConditionalRule,
-} from './conditionalRule';
```
joaoviana, line 4: Fixes tsoa routes not being happy with an `enum` alias. export with `as` happens at the end of the file now.

TuringLovesDeathMetal:
@joaoviana for the analytics event, it's just a matter of adding a parameter: `custom_metric_filter_count` to the existing `lightdash_server_query_executed` event :)

TuringLovesDeathMetal:
Woooo! Amazing job 😍 

Just a few things I noticed trying it out:

- Are we going to add the ability to edit the custom metric from the table? (it was in the acceptance criteria, but I'm not sure if we're going to build it in the next version or something!)
image

- there's a bug when I try to create a custom metric with the same name. https://www.loom.com/share/1b3056b7eb7543e69641f32be2d334ce

^Instead of having it get removed from the table, if someone tries to create a metric with the same name, can we show them an error message saying `A metric with the name "my metric name" already exists in your results.`

I imagine it would appear as a toast in the bottom right when I try to hit `create` here. The modal would stay open, and I'd just have to change the metric name: 

image

TuringLovesDeathMetal:
oop, just noticed you already built this!

image

I just don't think it's working properly 😅

joaoviana:
> * Are we going to add the ability to edit the custom metric from the table? (it was in the acceptance criteria, but I'm not sure if we're going to build it in the next version or something!)

oops! I took the `OR` too literally... in the issue: "Once a user has created a custom metric, they can edit it from the three-dot-menu in the results table, or in the sidebar."

should be easy to enable!

Regarding duplicate labels + overriding - will check that out, thanks @TuringLovesDeathMetal !

packages/common/src/types/field.ts
```diff
@@ -123,6 +123,14 @@ export type FieldId = string;
 export const fieldId = (field: Pick<Field, 'table' | 'name'>): FieldId =>
     `${field.table}_${field.name}`;
 
+export const convertFieldIdToFieldRef = (field: Field) =>
+    `${field.table}.${field.name}`;
+
+export const convertFieldRefToFieldId = (fieldRef: string) => {
+    const [table, name] = fieldRef.split('.');
+    return `${table}_${name}`;
+};
```
ZeRego, line 132: We should handle the case where the reference doesn't have a table. In those cases we assume it's the same table from the metric. 
Copy the logic from `resolveQueryFieldReference` to here and delete that function.

packages/common/src/types/field.ts
```diff
@@ -123,6 +123,14 @@ export type FieldId = string;
 export const fieldId = (field: Pick<Field, 'table' | 'name'>): FieldId =>
     `${field.table}_${field.name}`;
 
+export const convertFieldIdToFieldRef = (field: Field) =>
```
ZeRego, line 126: This is a duplicate of `getFieldRef` util function. We can delete this one :)

joaoviana: ahhh, thanks! removed

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+                        ...customMetricFiltersWithIds,
+                        {
+                            ...newFilterRule,
+                            target: {
+                                ...newFilterRule.target,
+                                fieldRef: convertFieldIdToFieldRef(field),
+                            },
+                        },
+                    ]);
+                }
```
ZeRego: Since we only support filters targeting dimensions I would remove all the other conditions/logic and keep this block with `if(isDimension(field)) {...}`

docs/docs/snippets/how-to-create-metrics.mdx
```diff
@@ -150,23 +154,35 @@ But, if there's something missing from this list of metrics, you can use `custom
...
 
 If you want to delete the custom metric, you can just click on the three-dot `options` menu and hit `Remove custom metric`
 
-<img src={DeleteCustomMetric} width="527" height="137" style={{display: "block", margin: "0 auto 20px auto"}}/>
+<img src={CustomMetricDeleting} width="300" height="137" style={{display: "block", margin: "0 auto 20px auto"}}/>
+
+### Setting filters to your custom metric
+
+These filters are based on the dimensions set in your project, allowing for more precise metrics.
+The ability to add filters is now available when you're creating a new custom metric or editing an existing one.
```
TuringLovesDeathMetal, line 178:
```suggestion
You can add filters when you're creating a new custom metric or editing an existing one.
```

joaoviana: thanks!

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
joaoviana: great point, thanks! i've simplified it here: https://github.com/lightdash/lightdash/pull/5865/commits/280660114ac27d93fbf0ca16c9a5e8b825ebee63

joaoviana:
> @joaoviana just noticed another thing
> 
> When you go to `edit` the custom metric, the filter input is looking weird (the field bit)
> 
> image

I think the same happens at the moment for the whole metric query filters... I will need to investigate this 👍

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
joaoviana: can hide the `filters`? 🤷

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+        >
+            <form onSubmit={handleOnSubmit}>
+                <Stack>
+                    <TextInput
+                        label="Label"
+                        required
+                        placeholder="Enter custom metric label"
+                        {...form.getInputProps('customMetricLabel')}
+                    />
+                    {defaultFilterRuleFieldId && (
```
ZeRego, line 206: We should allow filters to be added to old custom metrics. Or is there any reason why they shouldn't work ?

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+        (context) => context.state.isEditMode,
+    );
+    const { fieldsMap } = useFiltersContext();
+
+    const dimensions = Object.values(fieldsMap).filter(isDimension);
+
+    const addFieldRule = useCallback(() => {
+        const field = fieldsMap[defaultFilterRuleFieldId];
+
+        const newFilterRule = createFilterRuleFromField(field);
```
ZeRego, line 43: We should handle the case where field is undefined. eg: the base dimension was removed or is missing.

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
joaoviana: 🤦 Exactly, they should work regardless! 
Reverting this change.

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
joaoviana: Yes yes, going to do that now!

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+}) => {
+    const isEditMode = useExplorerContext(
+        (context) => context.state.isEditMode,
+    );
+    const { fieldsMap } = useFiltersContext();
+
+    const dimensions = Object.values(fieldsMap).filter(isDimension);
+
+    const addFieldRule = useCallback(() => {
+        const field = fieldsMap[defaultFilterRuleFieldId ?? 0];
```
ZeRego, line 41: I don't think using an index would work in a map. 

Captura de ecrã 2023-06-20, às 10 49 13

Suggestion
```
const fallbackField = Object.values(fieldsMap)[0];
const defaultField = defaultFilterRuleFieldId ? fieldsMap[defaultFilterRuleFieldId] : undefined;
const field = defaultField || fallbackField;

if(!field) {
   return;
}

```

joaoviana: oops, got mixed up with the values on line 38 and that. Changing! thanks 😄

ZeRego (approved):

Issue #5822, opened 2023-06-09
Scheduled deliveries dashboard: show all runs


yasmin-e:
### Description
In the run history tab, whenever there’s a new run, the entry is replaced for the old run. There should be a log of all of the runs in the last 30 days.

PR #5873, opened 2023-06-13
fix: Run history tab should show all runs

yasmin-e:
Closes: #5822 

### Description:
before
Screenshot 2023-06-13 at 14 07 46

after
Screenshot 2023-06-13 at 14 07 05


yasmin-e:
> weooo! nice!
> 
> Is it just me, or is it weird that the first row happened earlier than the second row in your example screenshot 🤔
> 
> Also, I can't seem to open a preview environment for this to test it out - any ideas what might be happening??

@TuringLovesDeathMetal I edited the schedule mid-runs, I think that's why!
I will debug Render so you can test it proeprly

Issue #3454, opened 2022-10-11
I want to be able to format my table calculations as a %


TuringLovesDeathMetal:
### Description

When I create a table calculation, I want the option to format the number as a % value. 

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVMM2H93o=/?moveToWidget=3458764554140396307&cot=14

- [ ] There are two tabs in the `table calculation` modal: `SQL` and `format`
- [ ] The `format` tab has a drop-down for `type` with the options `default` and `percent`
- [ ] By default, the `default` option is selected - this gives the same formatting as we use right now
- [ ] If a user selects `percent`, there are two more options:
    - [ ] `round` - this is an integer value. It can be negative or positive. E.g. `-2` applied to the value 2450` would give you `2500`.
        - [ ] The text when empty says `Number of decimal places`  
        - [ ] If the user leaves this blank, then we don't apply any rounding 
    - [ ] `separator style` - this is a drop-down list. It gives different separator style options (as listed in the miro sketch) that would be applied to your numbers. The options are: `100,000.00`, `100 000.00`, `100.000.00`, `100000.00`
        - [ ] By default, the separator style is set to `100,000.00`
- [ ] there is an example of what the output values would look like shown in the modal (see `looks like` on the miro sketch)

PR #5883, opened 2023-06-13
feat: format table calculation

rephus:
Closes: #3454

### Description:

image

[Screencast from 14-06-23 12:31:55.webm](https://github.com/lightdash/lightdash/assets/1983672/348ff757-430c-434f-a02c-e6e788190dfa)

### Description

When I create a table calculation, I want the option to format the number as a % value. 

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVMM2H93o=/?moveToWidget=3458764554140396307&cot=14

- [x] There are two tabs in the `table calculation` modal: `SQL` and `format`
- [x] The `format` tab has a drop-down for `type` with the options `default` and `percent`
- [x] By default, the `default` option is selected - this gives the same formatting as we use right now
- [x] If a user selects `percent`, there are two more options:
    - [x] `round` - this is an integer value. It can be negative or positive. E.g. `-2` applied to the value 2450` would give you `2500`.
        - [x] The text when empty says `Number of decimal places`  
        - [x] If the user leaves this blank, then we don't apply any rounding 
    - [x] `separator style` - this is a drop-down list. It gives different separator style options (as listed in the miro sketch) that would be applied to your numbers. The options are: `100,000.00`, `100 000.00`, `100.000.00`, `100000.00`
        - [x] By default, the separator style is set to `100,000.00`
- [x] there is an example of what the output values would look like shown in the modal (see `looks like` on the miro sketch)


TuringLovesDeathMetal:
ahhhhh! so exciting! This is super cool 😍 and I think it will be really exciting for users as well! 

Just some comments:

1. I realize this was in the acceptance criteria, so my bad! But, if a value is formatted as a percent, plotting that value should make it appear like a % as well. So, for example, in this chart, I'd expect the value labels to be 35.35%, etc. and the axis to be from 0 - 100 (not from 0 - 1). 
image

image

2. The bottom-right option should say `separator style` not `type`
3. Can we remove the `see our templates` bit from the `format` tab? (i.e. the bottom grey block)
4. Can we have the separator style be filled in with the style applied, by default? I.e. when I go to the format page and change it to percentage, I'd see this: 
image

5. The spacing between the top and bottom boxes looks a bit squished - I think they need to be more spread out.
image

6. I'm not sure if the bug was introduced here, or some other time, but you can't actually do anything when you try to expand the table calculation input to full screen: https://www.loom.com/share/605b367a997245ad90afd9e39f2f4b4a

PriPatel:
Looks good Javi! 
I have some suggestions here as well. @TuringLovesDeathMetal Let me know if you disagree with any of these :)


- If the `**Name**` field could be updated to a Mantine input, it would make everything here look consistent.

-  Could the `Round` input be https://mantine.dev/core/number-input/ this component? I was playing around with the round numbers and thought that it would be better for usability if I could adjust the numbers using buttons.

- Maybe the 'inspiration alert' at the bottom could be hidden at this point? Its quite overbearing and makes a lot more sense when I'm on the SQL tab.
Screenshot 2023-06-15 at 13 00 30

rephus:
@PriPatel 

> Could the Round input be https://mantine.dev/core/number-input/ this component? I was playing around with the round numbers and thought that it would be better for usability if I could adjust the numbers using buttons.

That would be nice, but we also need this field to be able to accept `undefined` , I don't think numberINput support this. I think the same thing happened on reference lines in chart config

rephus:
> Also, this has happened to me like 6 times in a row now: every time I use that input heavily, the app crashes and I get a blank screen! 

Uhhh, no idea, but I should investigate... could you open the JS console and see if there is an error ?

PriPatel:
@rephus Looking good! 
Screenshot 2023-06-15 at 15 49 03

Final spacing tweak -  I think there was spacing between the tabs and the bottom of the input and its now disappeared probably because the input has been replaced. It's just looking a bit tight . Other than that - looks great :)

rephus:
Fixed @PriPatel , good catch 
image

packages/frontend/src/components/TableCalculationModels/TableCalculationModal.tsx
```diff
@@ -131,71 +254,59 @@ const TableCalculationModal: FC<Props> = ({
                 }}
             >
                 <DialogBody className={Classes.DIALOG_BODY}>
-                    <Input
+                    <TextInput
+                        mb="sm"
                         label="Name"
                         name="name"
                         disabled={isDisabled}
-                        rules={{
```
rephus, line 138: we used to block table calculations with the same name, but now we have a function that allows having the same name, but assigns a different autoincremental id to each TC, so luckily we don't need this validation rule anymore

packages/frontend/src/components/TableCalculationModels/TableCalculationModal.tsx
```diff
@@ -85,9 +197,19 @@ const TableCalculationModal: FC<Props> = ({
         defaultValues: {
             name: tableCalculation?.displayName,
             sql: tableCalculation?.sql,
+            format: {
```
rephus, line 200: I started migrating react form to mantine use-form, but  because one of the components is an ace-editor I think is not that simple, so didn't refactor much, and new componets are all mantine.

TuringLovesDeathMetal:
Nice! Looking really good 😍 Just a few last things from me:

1. Adding value labels causes Lightdash to crash: https://www.loom.com/share/3527c344f2644f00ba115b8150ee11c8
2. The % values are still being plotted on a scale from 0 - 1 (I think they should be plotted on a scale from 0 - 100).

rephus:
@TuringLovesDeathMetal Fixed both !

image

TuringLovesDeathMetal:
Amazingggg!!!

Okay, one last thing: I've opened a separate issue to split numeric an percentage series across axes by default :https://github.com/lightdash/lightdash/issues/5935

The problem right now is that if I flip from a percentage axis to adding a numeric count, it doesn't make sense: https://www.loom.com/share/bb03ea4ac470445c99b4802e939e146a

Can we remove the percentage formatting for the axis labels for now? (then we can just fix it in the issue listed above)?

TuringLovesDeathMetal:
^Actually, @rephus I take it back! 

Let's leave that behaviour as-is. Then we can improve it with the other issue. 

Otherwise, people can't get % to appear at all on their axis labels (which is worse than having to fiddle around with the series to get it working properly)

packages/common/src/index.ts
```diff
@@ -675,11 +676,18 @@ export const getAxisName = ({
 export function getFieldMap(
     explore: Explore,
     additionalMetrics: AdditionalMetric[] = [],
+    tableCalculations: TableCalculation[] = [],
```
ZeRego, line 679: Not sure how typescript allowed these changes since this method should only return field ( dimensions/metrics ).

We also have `getItemMap` that does what you implemented here.

packages/common/src/utils/formatting.ts
```diff
@@ -271,13 +275,82 @@ export function formatFieldValue(
...
     convertToUTC?: boolean,
 ): string {
     if (value === null) return '∅';
     if (value === undefined) return '-';
+    if (
+        item !== undefined &&
+        !isAdditionalMetric(item) &&
+        isTableCalculation(item)
+    )
+        return formatTableCalculationValue(item, value);
```
ZeRego, line 353: Can we use brackets `{}` in this if block? was hard to tell the difference between the two returns

packages/frontend/src/hooks/echarts/useEcharts.ts
```diff
@@ -633,12 +650,21 @@ const getSimpleSeries = ({
...
                 formats[yFieldHash] && {
-                    formatter: (value: any) =>
-                        formatValue(value?.value?.[yFieldHash], {
-                            format: formats[yFieldHash].format,
-                            round: formats[yFieldHash].round,
-                            compact: formats[yFieldHash].compact,
-                        }),
+                    formatter: (value: any) => {
+                        const field = formats[yFieldHash];
+                        if ('round' in field) {
```
ZeRego, line 655: `round` is an optional property in `Field` type. This conditional doesn't feel safe has might have a lot of fields ending up in the `else` block. What do you think ?

ZeRego (changes requested):
Looking good :) 
Just a few improvements/fixes left.
We should apply the formatting in the big number chart.
Captura de ecrã 2023-06-16, às 15 57 42

rephus:
Done ! 
image

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -231,15 +232,7 @@ const useBigNumberConfig = (
...
           resultsData?.rows?.[0]?.[selectedField]?.value.formatted
-        : formatValue(firstRowValueRaw, {
-              format: isField(item) ? item.format : undefined,
-              round: bigNumberStyle
-                  ? 2
-                  : isField(item)
-                  ? item.round
-                  : undefined,
-              compact: bigNumberStyle,
-          });
```
ZeRego, line 242: I I suppose the big number configuration won't work if we remove these overrides.

ZeRego (changes requested):

packages/frontend/src/hooks/useBigNumberConfig.ts
rephus: You were right, I guess I didn't test it enough. Fixed

ZeRego (approved):

Issue #5838, opened 2023-06-09
Update errors to use error toast instead of popover


TuringLovesDeathMetal:
### Description

When you try to exit the error message using the "overall" `x`, it just automatically reappears. 

You need to hit `dismiss`, then `x` to have it disappear. 

WE should just replace the popover with an error toast.

https://www.loom.com/share/12af94b8380d425683db445128f7571b

### Acceptance criteria

- [ ] replace the popover toast with an error toast
- [ ] the toast should not disappear automatically - it should only disappear when dismissed.

### Questions

- [ ] Do we ever have situations where there are multiple errors listed at once? 

If yes, they should appear in the toast like this:

```
Error

- error 1
- error 2
- error 3
```

PR #5893, opened 2023-06-14
fix: update errors to use toasts

yasmin-e:
Closes: #5838 

### Description:
https://github.com/lightdash/lightdash/assets/67699259/e83fb03f-cf6a-4ca8-a6c4-40ab02c9d6f7


TuringLovesDeathMetal:
I'm getting a bunch of old errors appear when I trigger a new error. 

https://www.loom.com/share/30c41609d8054ab2a68b0f820a2892db

TuringLovesDeathMetal:
Here's how I got there:

- created a table calculation using `status / 0` (this won't work becuase I'm trying to divide a string by an int)
- created a table calculation using `avg amount / 0` (I wanted to see what would happen if I had multiple errors at once)
- got rid of `status / 0` table calculation and reran the query
- The old errors for `string division by int` popped up, even though that was no longer an error

TuringLovesDeathMetal:
^FYI I think what's happening is in the old error popover. If an error hadn't been dismissed, it would re-appear in the popover if a new error occurred. 

So the behaviour now in the toast looks the same as it used to be in the old popover. 

Instead, I think what we want to do now is assume that an error is automatically "dismissed" once it's been and disappears (after a certain period of time) _**or.**_ if it's been manually dismissed. 

Also, I just realized that this was in the acceptance criteria:

> the toast should not disappear automatically - it should only disappear when dismissed.


But I don't think that's true. I think the toast should disappear automatically after some time (e.g. 30 seconds). _**or**_ if a new error appears, the old error toasts should disappear automatically.

yasmin-e:
> Here's how I got there:
> 
> * created a table calculation using `status / 0` (this won't work becuase I'm trying to divide a string by an int)
> * created a table calculation using `avg amount / 0` (I wanted to see what would happen if I had multiple errors at once)
> * got rid of `status / 0` table calculation and reran the query
> * The old errors for `string division by int` popped up, even though that was no longer an error



> ^FYI I think what's happening is in the old error popover. If an error hadn't been dismissed, it would re-appear in the popover if a new error occurred.
> 
> So the behaviour now in the toast looks the same as it used to be in the old popover.
> 
> Instead, I think what we want to do now is assume that an error is automatically "dismissed" once it disappears after a certain period of time _**or.**_ if it's been manually dismissed.
> 
> Also, I just realized that this was in the acceptance criteria:
> 
> > the toast should not disappear automatically - it should only disappear when dismissed.
> 
> But I don't think that's true. I think the toast should disappear automatically after some time (e.g. 30 seconds). _**or**_ if a new error appears, the old error toasts should disappear automatically.

I just fixed the issue of errors re-appearing, I just forgot to delete them once we show them to the user.
I can implement toast fade-out but I thought dismissing manually would be safer?

TuringLovesDeathMetal:
Nice! Let me play around with it again - you might have already implemented the thing I'm talking about above 😅 and if not, I'll try to explain a bit better what I mean

TuringLovesDeathMetal:
Ya, it's still causing both errors to pop up. 

https://www.loom.com/share/f9dc54368100410aba0bceb28e1a5898?from_recorder=1&focus_title=1

I think that the error toasts are actually disappearing automatically, but they're not actually being dismissed.

https://www.loom.com/share/24e2be97ac29475b9d9446edfa6173a1

Can we:

- keep the automatic disappearing (I think it's great and looks good with teh timer it's on right now)
- make it so that if an error is dismissed manually _**or**_ disappears automatically, it's considered "dismissed" so it doesn't reappear (unless the error is triggered again)

yasmin-e:
> Ya, it's still causing both errors to pop up.
> 
> https://www.loom.com/share/f9dc54368100410aba0bceb28e1a5898?from_recorder=1&focus_title=1
> 
> I think that the error toasts are disappearing automatically, but they're not actually being dismissed.
> 
> https://www.loom.com/share/24e2be97ac29475b9d9446edfa6173a1
> 
> Can we:
> 
> * keep the automatic disappearing (I think it's great and looks good with teh timer it's on right now)
> * make it so that if an error is dismissed manually _**or**_ disappears automatically, it's considered "dismissed" so it doesn't reappear (unless the error is triggered again)

just tested on PR env

https://github.com/lightdash/lightdash/assets/67699259/e716c8a7-0d0a-480c-8031-861d8dfbbe19



> ^FYI I think what's happening is in the old error popover. If an error hadn't been dismissed, it would re-appear in the popover if a new error occurred.
> 
> So the behaviour now in the toast looks the same as it used to be in the old popover.
> 
> Instead, I think what we want to do now is assume that an error is automatically "dismissed" once it disappears after a certain period of time _**or.**_ if it's been manually dismissed.
> 
> Also, I just realized that this was in the acceptance criteria:
> 
> > the toast should not disappear automatically - it should only disappear when dismissed.
> 
> But I don't think that's true. I think the toast should disappear automatically after some time (e.g. 30 seconds). _**or**_ if a new error appears, the old error toasts should disappear automatically.
> 
> So, to summarize, the error should disappear if:
> 
> * it's been manually dismissed
> * 30 seconds have passed
> * a new error has appeared
> 
> The last one might be a bit tricky with multiple errors...

all good
https://github.com/lightdash/lightdash/assets/67699259/2ae24b33-7b25-40a7-9269-f059552cfc16

TuringLovesDeathMetal (approved):
Niiiiice! looks good to me :)

packages/frontend/src/components/Explorer/ExploreSideBar/index.tsx
```diff
@@ -63,13 +60,7 @@ const BasePanel = () => {
...
                 icon="error"
                 title="Could not load explores"
-                description="Check error logs for more details"
-                action={
-                    <ShowErrorsButton
-                        errorLogs={errorLogs.errorLogs}
-                        setErrorLogsVisible={errorLogs.setErrorLogsVisible}
-                    />
-                }
+                description="Check error toasts for more details"
```
IrakliJani, line 63: this is an odd message. do we have a screen both this showing up and toasts?

packages/frontend/src/hooks/useQueryError.ts
```diff
@@ -23,11 +23,11 @@ const useQueryError = (): Dispatch<SetStateAction<ApiError | undefined>> => {
                     // drawer
```
IrakliJani, line 23: we don't have a drawer anymore, right? will be good if we search for that keyword and remove any unnecessary mentions of it

IrakliJani (commented):
code looks good and clean (and happy to see that drawer removed!), just few concerns about the overall UX

IrakliJani (approved):

Issue #5081, opened 2023-04-13
The totals on my chart should adjust when I toggle fields as visible/not visible from view


TuringLovesDeathMetal:
### Description

When I toggle fields on/off, the bar total labels should update. 

![image](https://user-images.githubusercontent.com/31848148/231697852-fc6f188d-d465-4275-8304-b5507fd02cdb.png)
![image](https://user-images.githubusercontent.com/31848148/231697885-fa7df3b9-4605-4b31-8c97-f4ed1d21b994.png)

PriPatel:
Requested by someone in the community
https://lightdash-community.slack.com/archives/C03MG2VTR08/p1683139712560319

PR #5894, opened 2023-06-14
fix: adjust chart totals based on legends selected

CVamsi27:
Closes: #5081 

### Description:
Adjust chart totals based on legends selected

### DIFF
--- a/packages/frontend/src/components/SimpleChart/index.tsx
+++ b/packages/frontend/src/components/SimpleChart/index.tsx
@@ -2,7 +2,7 @@ import { NonIdealState, Spinner } from '@blueprintjs/core';
 import { PivotReference } from '@lightdash/common';
 import EChartsReact from 'echarts-for-react';
 import { EChartsReactProps, Opts } from 'echarts-for-react/lib/types';
-import { FC, memo, useCallback, useEffect, useMemo } from 'react';
+import { FC, memo, useCallback, useEffect, useMemo, useState } from 'react';
 import useEcharts from '../../hooks/echarts/useEcharts';
 import { useVisualizationContext } from '../LightdashVisualization/VisualizationProvider';
 
@@ -44,6 +44,12 @@ export type EchartSeriesClickEvent = EchartBaseClickEvent & {
 
 type EchartClickEvent = EchartSeriesClickEvent | EchartBaseClickEvent;
 
+type LegendClickEvent = {
+    selected: {
+        [name: string]: boolean;
+    };
+};
+
 export const EmptyChart = () => (
     <div style={{ height: '100%', width: '100%', padding: '50px 0' }}>
         <NonIdealState
@@ -73,7 +79,10 @@ const SimpleChart: FC<SimpleChartProps> = memo((props) => {
     const { chartRef, isLoading, onSeriesContextMenu } =
         useVisualizationContext();
 
-    const eChartsOptions = useEcharts();
+    const [selectedLegends, setSelectedLegends] = useState({});
+
+    const eChartsOptions = useEcharts(selectedLegends);
+
     useEffect(() => {
         const listener = () => {
             const eCharts = chartRef.current?.getEchartsInstance();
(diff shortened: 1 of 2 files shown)


CVamsi27:
Hello all, I am trying to fix the issue, but facing some difficulty. I would like to get some help!!
Here I was able to get the list of legends and their selection status - true/false
I was able to successfully make the totals - 0 when selected, but, on click of legend, the previous functionality that visually removes the bar is lost and also found some errors in the console.

Error:
```
Uncaught TypeError: Cannot read properties of null (reading 'trigger')
    at ECharts2.doDispatchAction (echarts.js:1606:23)
    at ECharts2.flushPendingActions (echarts.js:1627:26)
    at ECharts2.dispatchAction (echarts.js:1085:25)
    at Group5.dispatchSelectAction (LegendView.js:521:7)
    at Group5.<anonymous> (util.js:310:21)
    at Eventful2.trigger (Eventful.js:103:33)
    at Handler2.dispatchToElement (Handler.js:138:16)
    at Handler.<computed> [as click] (Handler.js:220:14)
    at Eventful2.trigger (Eventful.js:103:33)
    at localDOMHandlers.<computed> (HandlerProxy.js:162:14) 
```
ScreenShots:
image

packages/frontend/src/components/SimpleChart/index.tsx
```diff
@@ -105,13 +114,22 @@ const SimpleChart: FC<SimpleChartProps> = memo((props) => {
...
+            setSelectedLegends(params.selected);
+        },
+        [],
+    );
+
     const onEvents = useMemo(
         () => ({
             contextmenu: onChartContextMenu,
             click: onChartContextMenu,
+            legendselectchanged: onLegendChange,
```
rephus, line 128: I do'nt see any reference to this `legendselectchanged`...

rephus:
I don't think that state is going to persist if you save , perhaps we should store this list of enabled items in `validCartesianConfig` using a provider

CVamsi27:
Hello @rephus, thanks for the inputs, I will try to fix the issue and revert back with updates

packages/frontend/src/components/SimpleChart/index.tsx
CVamsi27: legendselectchanged gives a list of all legends that are currently available and their selection status - true/false

CVamsi27:
> I don't think that state is going to persist if you save a chart , perhaps we should store this list of enabled items in `validCartesianConfig` using a provider

Hello @rephus, true, I observed this, will use validCartesianConfig to store the state.

CVamsi27:
Hello @rephus, I successfully implemented Provider for legends selection, but still didn't work.
When I log `validCartesianConfigLegend` on clicking credit_card legend, its going back to original state when I use it useMemo - stackedSeries, eChartsOptions:
log:
image
Chart:
image

when removed from useMemo, log looks good:
image

What may be the issue?
Thanks

packages/frontend/src/components/SimpleChart/index.tsx
```diff
@@ -105,42 +121,50 @@ const SimpleChart: FC<SimpleChartProps> = memo((props) => {
         [onSeriesContextMenu, eChartsOptions],
     );
 
+    const onLegendChange = useCallback((params: LegendClickEvent) => {
+        setSelectedLegends(params.selected);
+    }, []);
+
     const onEvents = useMemo(
```
rephus, line 128: perhaps this doesn't need to be an useMemo, you can add these properties directly on `<EChartsReact`

rephus:
> when removed from useMemo, log looks good:

what's the memo that you say it is causing this issue ? 

This usually happens when you don't include all needed dependencies in `useMemo`

packages/frontend/src/components/SimpleChart/index.tsx
CVamsi27: Sure, looking into it!

owlas:
Hey @CVamsi27 javi isn't available this week, I'm going to close this while it's still a work in progress. Could you re-open with changes next week? Thanks!

CVamsi27:
> Hey @CVamsi27 javi isn't available this week, I'm going to close this while it's still a work in progress. Could you re-open with changes next week? Thanks!

Sure @owlas, will push required changes to the associated branch and re-open it next week

Issue #5797, opened 2023-06-07
There should be a gap between the label and the comparison number 


PriPatel:
<img width="562" alt="Screenshot 2023-06-07 at 16 09 58" src="https://github.com/lightdash/lightdash/assets/12776522/9da5da38-81bf-4533-958b-969deabd1701">

This is in cases where no arrow icon is shown

TuringLovesDeathMetal:
^btw, this only happens for a value of 0

PR #5895, opened 2023-06-14
fix: add gap between label and comparison

yasmin-e:
Closes: #5797 

### Description:
before
Screenshot 2023-06-14 at 15 35 57

after
Screenshot 2023-06-14 at 15 34 41


PriPatel (approved):
Yep - looks correct now!

PR #5899, opened 2023-06-14
fix: duplicated chart query

rephus:
Closes: #5890

### Description:

Issue: we were returning 1 chart per version !, I think this was causing the search filter as well. I haven't been able to replicate the search issue after fixing this.

Before:

After:

Screenshot from 2023-06-14 16-56-43


yasmin-e (approved):
nice catch 👍🏼

ZeRego (approved):

Issue #5892, opened 2023-06-14
You should be able to clear an explore using the keyboard shortcut [command + option + k]


PriPatel:
This is a quick way to implement https://github.com/lightdash/lightdash/issues/1235 without touching the UI
This should only affect explores, rather than saved charts. We will tackle those separately with redo/undo buttons.

@elton-okawa

PriPatel:
fyi I will be working on getting this action into this UI this week. ( so you can see the keyboard shortcut on hover as well!)

PR #5904, opened 2023-06-15
feat: add clear query shortcut

elton-okawa:
Closes: #5892

### Description:

If someone with a MacOS could confirm for me that `mod + alt + k` is equivalent as `command + option + k` in MacOS I'd appreciate!
- `Edit`: confirmed, thanks @ZeRego!

I think it would also be interesting to communicate for the user in some way that this shortcut exists 🤔 

`Edit`: Updated video showing that selected table is persisted

https://github.com/lightdash/lightdash/assets/24387035/a21f4130-9a1c-4a1c-85e5-f618ef7bd5ec

### DIFF
--- a/packages/e2e/cypress/e2e/explore.cy.ts
+++ b/packages/e2e/cypress/e2e/explore.cy.ts
@@ -320,7 +320,10 @@ describe('Explore', () => {
         // clear query hotkeys
         cy.get('body').type('{ctrl}{alt}{k}');
 
-        // verify empty selections
-        cy.findByText('Select a table').should('exist');
+        // verify empty query keeping selected table
+        cy.findByText('Tables', { selector: 'a' })
+            .parent()
+            .should('have.text', 'Tables/Orders');
+        cy.findByText('Pick a metric & select its dimensions').should('exist');
     });
 });
--- a/packages/e2e/cypress/e2e/explore.cy.ts
+++ b/packages/e2e/cypress/e2e/explore.cy.ts
@@ -304,4 +304,23 @@ describe('Explore', () => {
             });
         });
     });
+
+    it('Should clear query using hotkeys', () => {
+        cy.visit(`/projects/${SEED_PROJECT.project_uuid}/tables`);
+
+        cy.findByText('Orders').click();
+        cy.findByText('Is completed').click();
+
+        // run query
+        cy.get('button').contains('Run query').click();
+
+        // wait for query to finish
+        cy.findByText('Loading results').should('not.exist');
+
+        // clear query hotkeys
+        cy.get('body').type('{ctrl}{alt}{k}');
+
+        // verify empty selections
+        cy.findByText('Select a table').should('exist');
+    });
 });
--- a/packages/frontend/src/hooks/useExplorerRoute.ts
+++ b/packages/frontend/src/hooks/useExplorerRoute.ts
@@ -47,7 +47,9 @@ export const useExplorerRoute = () => {
     const queryResultsData = useExplorerContext(
         (context) => context.queryResults.data,
     );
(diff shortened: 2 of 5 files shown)


packages/frontend/src/hooks/useExplorerRoute.ts
```diff
@@ -54,17 +54,14 @@ export const useExplorerRoute = () => {
...
-            history.replace(
-                getExplorerUrlFromCreateSavedChartVersion(
-                    pathParams.projectUuid,
-                    {
-                        ...unsavedChartVersion,
-                        metricQuery: queryResultsData.metricQuery,
-                    },
-                ),
-            );
-        }
```
elton-okawa, line 67: Reseting query means reseting `explorer state` and then `query results`, these state changes causes one render for each one.
1º render: `unsavedChartVersion` as `default state` and `queryResults` with values
2º render: `unsavedChartVersion` as `default state` and `queryResults` as `undefined`

This verification skipped the url update on the (2), so even though there is nothing selected, `metricQuery` on url still had previous values.
As I thought this verification a bit strange and I didn't find any difference or bug after removing it, I chose this solution, if it breaks something I'd change the approach without any problem

packages/e2e/cypress/e2e/explore.cy.ts
```diff
@@ -304,4 +304,23 @@ describe('Explore', () => {
...
+        cy.get('button').contains('Run query').click();
+
+        // wait for query to finish
+        cy.findByText('Loading results').should('not.exist');
+
+        // clear query hotkeys
+        cy.get('body').type('{ctrl}{alt}{k}');
+
+        // verify empty selections
+        cy.findByText('Select a table').should('exist');
```
ZeRego, line 324: We want to clear the explore page but not leave the selected table. The current `clear` action has a different purpose and should be renamed and we can add a new `clear` action for this use case.

ZeRego (changes requested):

packages/frontend/src/hooks/useExplorerRoute.ts
ZeRego: The only problem I seen is that the url params aren't cleared when the user navigates back to the explores page, to pick a new table.

Perhaps we need to keep a condition `if(!unsavedChartVersion.tableName)`

elton-okawa: I noticed that when the user navigates back, the **url params is present but with the default** `unsavedChartVersion` state, so it's being cleared 🤔 

This happens because without the condition, the url params is filled with the default `unsavedChartVersion` state as soon as the user enters `explorer page`

Do you think that the url params being present with default `unsavedChartVersion` is a problem? I saw that `run query` and `save chart` button are still disabled in this case 🤔

packages/frontend/src/providers/ExplorerProvider.tsx
```diff
@@ -1105,14 +1106,28 @@ export const ExplorerProvider: FC<{
...
+            payload: {
+                ...defaultState,
+                unsavedChartVersion: {
+                    ...defaultState.unsavedChartVersion,
+                    tableName: unsavedChartVersion.tableName,
+                },
+            },
+        });
+        resetQueryResults();
+    }, [resetQueryResults, unsavedChartVersion.tableName]);
```
elton-okawa, line 1129: Hi @ZeRego, thanks for the code review!

You suggested renaming original `clear` because its use case was different, but looking at its usage I thought that `clear` is a good name 🤔 

I didn't find a more suitable name 😅, so in the end I renamed the original one to `clearExplore` and create the suggested one as `clearQuery`. Let me know what you think about!

ZeRego (approved):
lgtm :) 

We will soon have the button in the UI and display the shortcut there.

Issue #5910, opened 2023-06-15
Add the API to get a project


yu-iskw:
### Description
We would like to manage the API to get a project with the schema.

### What problem does this solve?
We can manage the API with the schema.

PR #5911, opened 2023-06-15
feat: Add the API to get a project

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/5910

### Description:
We would like to manage the API to get a project with `swagger.json`.


rephus:
Sorry but a test is failing, I'm trying to figure out why..

$ yarn workspace backend build
$ tsc --build tsconfig.json
src/generated/routes.ts(890,5): error TS1117: An object literal cannot have multiple properties with the same name.

rephus:
Ok, I know why, it seems routes.ts was incorrectly merged.

Could you generate-api again and push that change ? 

image

yu-iskw:
@rephus I have re-generated `routes.ts` and `swagger.json`. 

By the way, we may want to have something to automatically generate the two files with husky or GitHub Actions so that we can keep them fresh.

rephus:
@yu-iskw 

> By the way, we may want to have something to automatically generate the two files with husky or GitHub Actions so that we can keep them fresh.

Yes, I agree, we recently tried with husky, but it was creating conflicts all the time , even when routes were not changing. We need to fix this https://github.com/lightdash/lightdash/issues/5930 before we can fully automate it

rephus (approved):
Manually tested on render, looks good !

Issue #5912, opened 2023-06-15
When a space has  a chart but no dashboards, it shows an empty state


yasmin-e:
### Description
https://www.loom.com/share/500bef58be1d4e4a8817fcb17fbd9cc0?sid=5cd4955a-eb4b-4c47-b709-38ab81929afe

### To Reproduce

- Create a space
- Add a chart to it
- Go to space page, you will see an empty state even though you have 1 chart saved in the space

### App Version
Version: v0.613.2

PR #5913, opened 2023-06-15
fix: when space has a chart but no dashboards, it shouldn't be empty

yasmin-e:
Closes: #5912 

### Description:
before

https://www.loom.com/share/500bef58be1d4e4a8817fcb17fbd9cc0?sid=e321f62f-cbfc-48c6-993a-bc778643bee5

after

https://www.loom.com/share/e19f3857b2e044b58ab6108fe8493462?sid=7356e2a5-8578-4885-abbd-6bf311a849a5

### DIFF
--- a/packages/frontend/src/components/common/ResourceView/index.tsx
+++ b/packages/frontend/src/components/common/ResourceView/index.tsx
@@ -84,7 +84,7 @@ const ResourceView: React.FC<ResourceViewProps> = ({
         type: ResourceViewItemAction.CLOSE,
     });
 
-    const [activeTabId, setActiveTabId] = useState<string | null>(null);
+    const [activeTabId, setActiveTabId] = useState(tabs?.[0].id);
 
     const handleAction = useCallback(
         (newAction: ResourceViewItemActionState) => {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/ResourceView/index.tsx
```diff
@@ -84,7 +84,7 @@ const ResourceView: React.FC<ResourceViewProps> = ({
         type: ResourceViewItemAction.CLOSE,
     });
 
-    const [activeTabId, setActiveTabId] = useState(tabs?.[0]?.id);
+    const [activeTabId, setActiveTabId] = useState(tabs?.[3]?.id);
```
IrakliJani, line 87: im very confused. how does this solve the problem? I mean, why it is a problem on the first place

IrakliJani: what happens if you refresh (hard refresh) the page. does it then show up the chart item? if so, then this is a caching issue

IrakliJani: ok, we should probably fix the root cause because if we add/remove tabs this can slip out yet as another bug.

yasmin-e: you mean the logic of `slicedSortedItems`

yasmin-e: done 💯

PR #5915, opened 2023-06-15
feat: migrate limit popover to mantine

IrakliJani:
closes: https://github.com/lightdash/lightdash/issues/5917

CleanShot 2023-06-15 at 14 09 14@2x


PriPatel (changes requested):
Looks so much better! I like the way it goes up in increments of 100!

Only thing that's missing for me is the error messages we currently show when the user has entered something invalid. Having said that, the sensible defaults made sense. (eg. I entered `-1` and it applied `1`) or I entered `100sss` and it applied `100`. Maybe its fine for those cases, but I think it should be clear why when I add `10000`, it defaults to `5000` (as thats the max)

IrakliJani:
@PriPatel great points. I think defaulting to a number (10ssss → 10) and (-1 to 1) is a default behaviour of a normal number input and we used to have a text box before. I think we can still add reasonable error messages like it used to.

IrakliJani:
@PriPatel 
https://github.com/lightdash/lightdash/assets/962095/6c912184-00de-45d3-bfd4-198550b43de3

IrakliJani:
@yasmin-e @PriPatel I decided to keep the large length since error messages are taking a lot of vertical space. 
@PriPatel have no idea how to structure this differently with large error emssages. lmk if you have thoughts about it and I'll do a separate PR for it!

PriPatel:
@IrakliJani 
Good point! How about some shorter error messages?
Over 500: `Maximum value : 500`
A Negative number: `Minimum value: 1`
String : `Invalid value`
Am I missing anything else here?

I don't think we need to go into more detail than that, it just needs to be clear that there is an error. wdyt?

IrakliJani:
@PriPatel great! I created an issue

PR #5916, opened 2023-06-15
feat: migrate sort popover to mantine

IrakliJani:
closes: https://github.com/lightdash/lightdash/issues/5918

CleanShot 2023-06-15 at 14 11 44@2x


PriPatel (changes requested):
Wahoooo - this looks so much nicer!!

Only comment is on the drag handles - they look 'horizontal' when they should be 'vertical' (like the current way attached?) 

Screenshot 2023-06-15 at 12 41 10

IrakliJani:
@PriPatel updated

CleanShot 2023-06-15 at 15 58 23@2x

packages/frontend/src/components/LimitButton/index.tsx
```diff
@@ -28,7 +28,7 @@ const LimitButton: FC<Props> = memo(
             <Popover
                 disabled={disabled}
                 opened={opened}
-                position="bottom"
+                position="top"
```
IrakliJani, line 31: why top? good question. popover does not overlap results table, you can still see headers and data, find it a better UX especially for sorting popover

joaoviana: agree!

yasmin-e (approved):

Issue #5817, opened 2023-06-08
Include filters in the dashboard URLS 


PriPatel:
In Lightdash, if I'm on a dashboard and change the filters, neither the URL nor the link button update when filters have been changed. This means there is no way to filter a dashboard and then share that filtered view with someone. 

- This would enable sharing and bookmarking
- It means I could create one dashboard, and create 5 'versions' of it by using different filters and saving the links. (eg. One URL has a `UK` filter applied, another URL has a `US` filter applied. These links would be generated off a single dashboard) 

Note : This could be a crossover of: 
https://github.com/lightdash/lightdash/issues/4994

TuringLovesDeathMetal:
Note: I think this only happens when youre updating an existing filter, not creating a new one

yasmin-e:
> Note: I think this only happens when youre updating an existing filter, not creating a new one

^ That's right! When adding temporary filters (not in edit mode) to a dashboard, the URL updates with the applied filters in real time.
https://www.loom.com/share/99ac2cd4fa3845ad9dbc0c6b68ea640a?sid=d9e7e913-2f8e-4072-8877-cebd5d5a315f

However, when editing an existing filter on a dashboard, the URL does not update.
@PriPatel @TuringLovesDeathMetal we want it to update in this case too, right?

PriPatel:
@yasmin-e  Lets leave it as all filters in view mode for now! 

That would help the users experiecing the problem for now. We might need to think a bit more about what it means to update the URL in edit mode - we already don't show the Link button in edit mode. Could this get confusing etc? 
Thankssss

TuringLovesDeathMetal:
Here's an example of the bug: https://www.loom.com/share/311dfde5be65447780aad5921e4cea57

yasmin-e:
Hi @stellar-geoff, thank you for the feedback ! 🙌🏼 
I do have a follow up question:

Can you confirm that despite the URL not being verbose about the changes, the filters are being updated adequately?

I investigated and this is what I get:
- I have a saved filter on the dashboard, the URL has filters params 
- If I copy my URL and open it in a new or incognito tab, my filters appear
- If I change the value of my filters, the URL seems to stay the same
- If I copy this URL after updating my filters and open it in a new tab, my filters are updated

So, essentially, the behaviour is the expected one, I just don't see the values written out explicitly in the URL.

yasmin-e:
Hey @stellar-geoff, worked on a fix this afternoon 👍🏼 
[...]

PR #5922, opened 2023-06-15
fix: dashboard URL should have filters (temporary and saved)

yasmin-e:
Closes: #5817 

### Description:
```
In the params,
- temporary filters are 'tempFilters'
- saved filters are 'filters'
```

https://www.loom.com/share/8eaf9a23e2924c9cb7821f72254c3872?sid=2fd2ac4e-7436-42f8-8e2c-a850b7835ce8


rephus (approved):
Looks good. Nice demo.

In theory we don't need `filters` on the URL if the filter match the saved filter from the dashboard. Right ? But it doesn't hurt (for now)

rephus:
I've added a new ticket to the backlog to properly minimize URLs from dashboards https://github.com/lightdash/lightdash/issues/5929

Issue #5891, opened 2023-06-14
When there are no dashboards in a space, the active content tab should be on charts by default.


PriPatel:
<img width="903" alt="Screenshot 2023-06-14 at 11 22 24" src="https://github.com/lightdash/lightdash/assets/12776522/4e131df8-b081-493d-9ec2-b21b9a3f5b89">
This is not useful to me - I have to click on 'charts' everytime. It's better to default to the charts tab when there are no dashboards saved.

PR #5931, opened 2023-06-16
fix: land on charts if dashboards are empty

CVamsi27:
Closes: #5891 

### Description:
If no dashboards, the default tab will be Charts

### ScreenShots: 
Default tab is Charts,
image
image

### DIFF
--- a/packages/frontend/src/components/common/ResourceView/index.tsx
+++ b/packages/frontend/src/components/common/ResourceView/index.tsx
@@ -11,6 +11,7 @@ import {
     useMantineTheme,
 } from '@mantine/core';
 import { IconInfoCircle } from '@tabler/icons-react';
+import identity from 'lodash-es/identity';
 import React, { useCallback, useMemo, useState } from 'react';
 import { useTableTabStyles } from '../../../hooks/styles/useTableTabStyles';
 import ResourceActionHandlers, {
@@ -64,7 +65,7 @@ interface ResourceViewProps extends ResourceViewCommonProps {
 
 const ResourceView: React.FC<ResourceViewProps> = ({
     view = ResourceViewType.LIST,
-    items,
+    items: allItems,
     maxItems,
     tabs,
     gridProps = {},
@@ -81,10 +82,6 @@ const ResourceView: React.FC<ResourceViewProps> = ({
         type: ResourceViewItemAction.CLOSE,
     });
 
-    const [activeTabId, setActiveTabId] = useState(
-        defaultActiveTab || tabs?.[0]?.id,
-    );
-
     const handleAction = useCallback(
         (newAction: ResourceViewItemActionState) => {
             setAction(newAction);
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/ResourceView/index.tsx
```diff
@@ -84,7 +90,11 @@ const ResourceView: React.FC<ResourceViewProps> = ({
         type: ResourceViewItemAction.CLOSE,
     });
 
-    const [activeTabId, setActiveTabId] = useState(tabs?.[0]?.id);
+    const [activeTabId, setActiveTabId] = useState(
+        dashboards?.length === undefined || dashboards?.length === 0
+            ? tabs?.[1]?.id
+            : tabs?.[0]?.id,
+    );
```
yasmin-e, line 97: Hi @CVamsi27, unfortunately we can't fix this bug like this because ResourceView is used in many other cases where the tabs are not 'charts' 'dashboard 'all-items', and we need to be careful not to introduce bugs to other use cases.
Also, the tabs could change in the future and make this obsolete 😅 
It's not an easy bug to fix, we have a draft PR open for it and @IrakliJani is currently investigating it > https://github.com/lightdash/lightdash/pull/5913

yasmin-e (changes requested):

packages/frontend/src/components/common/ResourceView/index.tsx
ZeRego: Possible solution: 
- add new property to set the default active tab

```<ResourceView defaultActiveTab={!dashboards || dashboards.length === 0 ? 'charts' | 'dashboards' }>```

Comment (CVamsi27) [reply packages/frontend/src/components/common/ResourceView/index.tsx:97]: Hello @yasmin-e, thanks for that, I thought it's a simple fix as it was tagged with Good First Issue! I was not able to track already implemented PR as the issue was not tagged in the PR.

Hello @ZeRego, should I implement your suggestions in this PR of wait for #5913 
or should I close this PR

Comment (yasmin-e) [reply packages/frontend/src/components/common/ResourceView/index.tsx:97]: cc @CVamsi27, @ZeRego 
@IrakliJani just merged a fix for it!
https://github.com/lightdash/lightdash/pull/5938

Comment (yasmin-e) [reply packages/frontend/src/components/common/ResourceView/index.tsx:97]: actually @CVamsi27, the PR doesn't fix this bug 😅  I got confused.
You can go ahead and implement the fix Jose suggested on this PR or I can pick it up!
Let me know, and sorry for the confusion.

Comment (CVamsi27) [reply packages/frontend/src/components/common/ResourceView/index.tsx:97]: @yasmin-e No worries, I will implement the changes here

Comment (CVamsi27) [reply packages/frontend/src/components/common/ResourceView/index.tsx:91]: @yasmin-e Implemented the fix suggested by @ZeRego

**Review by yasmin-e — APPROVED** [review]
looks great 💯

File: packages/frontend/src/pages/Space.tsx [hunk 2]
```diff
@@ -317,6 +317,11 @@ const Space: FC = () => {
                         icon: <IconLayoutDashboard size={30} />,
                         title: 'No items added yet',
                     }}
+                    defaultActiveTab={
+                        !dashboards || dashboards.length === 0
+                            ? 'charts'
+                            : 'dashboards'
+                    }
```
IrakliJani, line 324: we can actually shift the logic and move it inside resource view to select a tab based on the items passed to it. we'll no longer need to have `defaultActiveTab` manually supplied to the component @yasmin-e thoughts?

CVamsi27: Hello @IrakliJani, will implement and update here. Please let me know if something different is required

IrakliJani: @CVamsi27 
yep, I think it will be best to iterate over tabs and also check for items.

for example:

check for charts: 0 items (skip)
check for dashboards: 1 item (set)

make sure if both charts and dashboards are 0 then the empty state shows up

yasmin-e: cc @IrakliJani @CVamsi27 
🤔  considering ResourceView will get more complex soon, having the prop can be good but I think we should still check inside the comp as default behaviour like @ZeRego suggests.

IrakliJani: @ZeRego @yasmin-e overriding sounds good. since `ResourceView` is used in a lot of places I would love to have some kind of consistency and a sense that every page which uses `ResourceView` is behaving the same way. in this PR and referenced issue we are only addressing problem on the Spaces page. but I am guessing same problem might also exist on the home page (where we have `ResourceView` with tabs).

I think @ZeRego suggestion is the best of both ideas:

- setting default tab that has first found resource (length > 0)
- having an override `defaultActiveTab` which will take a priority ^

cc: @CVamsi27

CVamsi27: Hello @IrakliJani, should I change something in this PR or its good to go?

IrakliJani: hi @CVamsi27 it will be good if you can incorporate some of the changes that we discussed here. we agreed to have  `defaultActiveTab` but it will also be good if ResourceView detects which tab to set as a default if `defaultActiveTab` is not provided to the component. happy to take it from what you've already done or wait for your changes

CVamsi27: Hello @IrakliJani, I am little confused here and a little busy this weekend. Please go ahead and make the changes. Sorry for the inconvenience.
Thanks, Should I close this PR?

Issue #5668, opened 2023-05-30
Upgrade dbt on Lightdash server to 1.5


owlas:
### Description

Upgrade dbt on the lightdash server to 1.5

This will break any existing lightdash projects that have a `packages.yml` listing a dependency that doesn't support dbt 1.5. This happens commonly because users pin their packages in `packages.yml`.

### (optional) - make project version configurable

We could install multiple versions of dbt on the server and allow the user to specify a version.

This could be achieved by:
- installing multiple dbt versions with python virtual environments
- adding an option to the UI to select the project version (although this only matters when hitting "refresh dbt" not when doing lightdash deploy).
- when executing dbt commands, target the correct dbt version

ZeRego:
Details on how we can support multiple dbt versions in https://github.com/lightdash/lightdash/issues/5109

Issue #5109, opened 2023-04-17
Support multiple dbt versions


ZeRego:
### Description

Support multiple dbt versions in the same Lightdash instance. 
At the moment user projects need to be compatible with the only dbt version supported by Lightdash. 

The user should be able to specify what dbt version should be used for each project. They should also be able to update the dbt version at any time. 

### What problems does this solve?

When Lightdash updates the version, users might get incompatibility errors.
Users using the most recent version aren't blocked by Lightdash using an older version.

### Possible implementation

This work can be split into multiple PRs.

1. Add tools to support multiple dbt versions

- remove dbt install from dockerfile and instead install https://github.com/brooklyn-data/dbtenv
- change `dbtCliClient.ts` to install the necessary dbt version if not yet installed ( note that the version should be hardcoded for now v1.4.0 )
- change `dbtCliClient.ts` commands from `dbt` to `dbtenv` with a specific version

2. Allow users to select dbt version for their project

- migrate DB so the projects table has a `dbt version` string column
- update types, models, services, endpoints
- update `dbtCliClient.ts` to use the project dbt version instead of the hardcoded version
- add a dropdown in the project from where the user can select the dbt major+minor version eg: 1.3, 1.4.

Possible improvements:
- pre-install some versions on server start

#### How to handle the different manifest versions

We rely heavily on dbt manifest file that is generated with `dbt compile`. At the moment we have 1 function that parses this json and collects the necessary information. 
Since we are going to support multiple versions of this manifest we need one function/class per version.

TuringLovesDeathMetal:
For the first version, let's not make it possible for users to select the dbt version they're using. Instead let's just run 1.5 by default, and if it fails, we run 1.4 instead

rephus:
I've been sharing this ticket with some customer https://github.com/lightdash/lightdash/issues/5668, let's merge that one too  once this ticket is done

PR #5936, opened 2023-06-16
feat: multiple dbt version on Dockerfile

rephus:
Closes: #5109
Closes: #5668

### Description:

Screenshot from 2023-06-16 11-41-54
Screenshot from 2023-06-16 11-41-58

Forcing an error to test fallback: 

Screenshot from 2023-06-16 13-53-54

Another error (fails on 1.4, but not on 1.5) 

Screenshot from 2023-06-16 14-00-48

How to test without docker
```
# assuming dbt (1.4) is already installed locally
python3 -m venv dbt/dbt1.5
dbt/dbt1.5/bin/pip install "dbt-postgres==1.5.0"
sudo ln -s ${PWD}/dbt/dbt1.5/bin/dbt /usr/local/bin/dbt1.5
```

WIth an uncompatible schema and 1.5

Screenshot from 2023-06-16 15-50-41


docker/docker-compose.dev.yml
```diff
@@ -16,15 +16,14 @@ services:
...
+      - RUDDERSTACK_WRITE_KEY=1vikeGadtB0Y0oRDFNL2Prdhkbp
+      - RUDDERSTACK_DATA_PLANE_URL=https://analytics.lightdash.com
       - SECURE_COOKIES=false
       - LIGHTDASH_SECRET="not very secret"
       - LIGHTDASH_LOG_LEVEL=debug
-      - SENTRY_DSN=""
-      - NODE_ENV="development"
-      - LIGHTDASH_CONFIG_FILE="/usr/app/lightdash.yml"
-      - DBT_DEMO_DIR="/usr/app/examples/full-jaffle-shop-demo"
+      - NODE_ENV=development
```
rephus, line 24: I've updated these env variables because strings go wtihout quotes. Otherwise it causes different unexpected errors like:

Screenshot from 2023-06-16 13-47-30

Screenshot from 2023-06-16 13-47-18

packages/cli/src/dbt/manifest.ts
```diff
@@ -8,10 +8,7 @@ type LoadManifestArgs = {
     targetDir: string;
 };
 
-export enum DbtManifestVersion {
```
rephus, line 11: moved to common

packages/backend/src/dbt/dbtCliClient.ts
```diff
@@ -207,14 +252,27 @@ export class DbtCliClient implements DbtClient {
...
-            filename,
-        );
-        return DbtCliClient.loadDbtFile(fullPath);
+    private async loadDbtTargetArtifact(
+        version: DbtManifestVersion,
+        filename: string,
+    ): Promise<any> {
+        switch (version) {
+            case DbtManifestVersion.V9:
+                return DbtCliClient.loadDbtFile('./target/manifest.json');
```
ZeRego, line 261: can't the target dir be configured in dbt 1.5 ? Or do we need to change the default value in `_getTargetDirectory` depending on the version ?

ZeRego:
This looks promising 🎉  but we need to find a scalable and cleaner way to add support for next versions without changing so many files and install dbt version on the fly.

packages/backend/src/dbt/dbtCliClient.ts
rephus: yes, you can change the target on 1.5 too, I'll move the logic to the _getTargetDirectory method then

rephus: Done

rephus:
> This looks promising tada but we need to find a scalable and cleaner way to add support for next versions without changing so many files and install dbt version on the fly.

Thanks , yeah, we could 1 or 2 dbt versions preinstalled , and once we allow users to select their own version, we could run this command before running dbt commands. 

```
RUN python3 -m venv /usr/local/dbt1.x
RUN /usr/local/dbt1.x/bin/pip install  "dbt-...~=1.x" 
RUN ln -s /usr/local/dbt1.x/bin/dbt /usr/local/bin/dbt1.x
```

However, I think we'll need to add manifests manually for new versions, sometimes we need to change the attributes when dbt renames things (like in v9 )

let me know if there is anything else you thing we can refactor/ make it better.

dockerfile
```diff
@@ -18,18 +18,30 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     wget \
     && apt-get clean
 
-# dbt
-RUN python3 -m venv /usr/local/venv
-RUN /usr/local/venv/bin/pip install \
+# Installing multiple versions of dbt
```
ZeRego, line 21: Do we need to update dockerfile-prs ? at the moment render envs are only installing dbt 1.4

packages/backend/src/dbt/dbtCliClient.ts
```diff
@@ -118,7 +120,44 @@ export class DbtCliClient implements DbtClient {
...
+                    version:
+                        dbtExec === 'dbt1.5'
+                            ? DbtManifestVersion.V9
+                            : DbtManifestVersion.V7,
+                    logs: await this._runDbtCommand(dbtExec, ...command),
+                };
+            } catch (e) {
+                Sentry.captureException(e, { extra: { dbtExec } });
+                Logger.warn(
+                    `Error running ${dbtExec} command "${command.join(' ')}"`,
```
ZeRego, line 147: Should we add the error message to this warning ?

ZeRego (approved):

PR #5938, opened 2023-06-16
fix: resource view tabs not visible when there is one item

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/5912

there was a mistake in the logic where it was hiding the tab bar when there was only 1 item.


yasmin-e (approved):
great catch!!! 💯

Issue #5933, opened 2023-06-16
Lightdash CLI doesn't recognise dbt metrics when using dbt 1.5


ZeRego:
### Description

Note: when using dbt 1.4 everything works well.

### To Reproduce

Steps 
1. install dbt 1.5
2. add dbt metrics
3. run `lightdash preview` 
4. See project was created without dbt metrics

Tech note: dbt 1.5 changed the `refs` data structure so the CLI logic doesn't recognise the table that metric should belong to. 

### App Version

0.616.0

PR #5940, opened 2023-06-16
fix: dbt 1.5 metrics in CLI commands

ZeRego:
Closes: #5933 

### Description:

dbt 1.5 changed the metrics refs data structure. We use the references to know in which table the metrics should be displayed. 
Updated logic to handle both dbt 1.4 and 1.5 ref data structures.


rephus (approved):
How can we test this ?

Issue #4798, opened 2023-03-14
Can't set custom value in filter when there are partial matches


ZeRego:
### To Reproduce

When there are no matches:
![Captura de ecrã 2023-03-14, às 16 50 37](https://user-images.githubusercontent.com/9117144/225078944-d1921543-0c11-43b5-adc5-cf70f35e4b3c.png)

When there are partial matches:
![Captura de ecrã 2023-03-14, às 16 51 13](https://user-images.githubusercontent.com/9117144/225079043-28baffd5-9413-49a2-831d-43d874fe92dc.png)

Even when there are partial matches we should have the option to add the custom value.

### App Version

0.462.0

yasmin-e:
Resolved ✅ 
The only time you can not add a custom value is when you have an exact match
<img width="353" alt="Screenshot 2023-06-16 at 14 37 10" src="https://github.com/lightdash/lightdash/assets/67699259/e9884463-364b-4f7e-989c-4930fc739471">
<img width="334" alt="Screenshot 2023-06-16 at 14 39 49" src="https://github.com/lightdash/lightdash/assets/67699259/68367798-95e1-4a06-bf23-0c64314046d2">
<img width="333" alt="Screenshot 2023-06-16 at 14 40 33" src="https://github.com/lightdash/lightdash/assets/67699259/6a5fbbd2-c707-458f-98bb-28a1a8b0b8c2">

PR #5941, opened 2023-06-16
fix: escape underscore for string filters in the UI

yasmin-e:
Closes: #4798 

### Description:
before
Screenshot 2023-06-16 at 14 23 55

after
Screenshot 2023-06-16 at 15 19 09

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.ts
+++ b/packages/common/src/compiler/filtersCompiler.ts
@@ -29,10 +29,12 @@ export const renderStringFilterSql = (
     const filterType = filter.operator;
     const escapedFilterValues = filter.values?.map((v) =>
         typeof v === 'string'
-            ? v.replaceAll(
-                  stringQuoteChar,
-                  `${escapeStringQuoteChar}${stringQuoteChar}`,
-              )
+            ? v
+                  .replaceAll(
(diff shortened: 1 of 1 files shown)


packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -29,10 +29,12 @@ export const renderStringFilterSql = (
...
-            ? v.replaceAll(
-                  stringQuoteChar,
-                  `${escapeStringQuoteChar}${stringQuoteChar}`,
-              )
+            ? v
+                  .replaceAll(
+                      stringQuoteChar,
+                      `${escapeStringQuoteChar}${stringQuoteChar}`,
+                  )
+                  .replaceAll('', '\\_')
```
ZeRego, line 37:
```suggestion
                  .replaceAll('_', '\\_')
```

Would be good to also escape the rest of the wildcards. 
The `escapedFilterValues` is used by all filters, not just `LIKE` filters. Would they still work as intended ? 
eg: `name equals jo_se`

yasmin-e: How do I know that a character is a wildcard?
And yes, it breaks them 😭  could I isolate this to `not_include` filter?

PR #5946, opened 2023-06-16
chore: update docusaurus port to 3002

joaoviana:
### Description:

When writing documentation while delivering a feature in the same PR, it's quicker to get some screenshots on local development. 
`docusaurus`'s default port is `3000` which conflicts when running dev environment.


rephus (commented):
3001 is also taken by headless browser, can we give it something different ?

rephus (approved):

Issue #5900, opened 2023-06-14
Prevent browsers from populating warehouse credentials inputs


ZeRego:
### Description

There are HTML attributes that indicate to the browser to not populate specific inputs, but browsers aren't required to comply.
More details:
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-password

We should use these attributes even if it doesn't cover 100% of the browsers. We hope all browsers will eventually comply.

### To Reproduce

Steps:
1. have your lightdash user/password saved in Chrome password manager
2. Go to project settings
4. See user and password inputs populated

### App Version

0.611.0

owlas:
I believe this behaviour is very useful depending on the browser implementation. This was raised because a user found that the form was autocompleted with their Lightdash username + password but this is not always the case. But in some cases the users actually do have their warehouse credentials in their password manager.

wasifkareem:
Hi @ZeRego , Can I work on this issue??

TuringLovesDeathMetal:
I'm closing this issue as not planned! THis hasn't been requested by other users since opened and is stale

PR #5948, opened 2023-06-17
Fix/prevent autofill on warehouse credentials

elton-okawa:
Closes: #5900

### Description:

It seems that modern browsers doesn't support `autocomplete='off'` on purpose:

```
...
in-browser password management is generally seen as a net gain for security...
For this reason, many modern browsers do not support autocomplete="off" for login fields
```

An alternative approach would be using `autocomplete='new-password'` which makes sense for the issue case.

[MDN docs Reference](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields)

### Browser compatibility

Browser compatibility table for `autocomplete='new-password'` says that only Firefox support this value, but I tested on Chrome and it worked as expected 🤔 

If someone could test it out on Safari, it'd be a great help!

[MDN docs autocomplete browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#browser_compatibility)

### Compatibility when migrating `Input` from `blueprint` to `mantine`

Both `blueprint` and `mantine` spreads received props on `input` `DOM` element, so theoretically it should work without any problems.

References:
- [blueprint - InputGroup](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/forms/inputGroup.tsx#L232)
- [mantine - Input](https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Input/Input.tsx#L132-L133)
- [mantine - Box](https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Box/Box.tsx#L16)

### Videos

<details>
  <summary>Chrome</summary>

   https://github.com/lightdash/lightdash/assets/24387035/c8ab617f-51e5-4fc9-ae61-617999d9cdb6

</details>

<details>
  <summary>Firefox</summary>

  https://github.com/lightdash/lightdash/assets/24387035/93a72ea0-d4ab-4d4f-a2bf-bba9c682658f

</details>


IrakliJani (changes requested):
@elton-okawa I think the simplest solution here would be to rename fields with unpredictable names for password managers and forget about cross-platform compatibility.

let me know what you think and if you need help on anything

owlas:
The aim of https://github.com/lightdash/lightdash/issues/5900 is to simplify the experience of completing the form by removing password managers, however this change would make password managers make an even worse suggestion (of filling in a new password for the form).

https://github.com/lightdash/lightdash/issues/5900

PR #5954, opened 2023-06-19
fix: declare `OrganizationMemberProfileUpdate` without `Partial<>`

yu-iskw:
### Description:
We may want to declare `OrganizationMemberProfileUpdate` without `Partial<>`. That make the generated swagger.json more complicated. As a result, some libraries for Open API can't part the complicated recursive declarations. 

https://github.com/lightdash/lightdash/blob/main/packages/common/src/types/organizationMemberProfile.ts#L40-L42
```
export type OrganizationMemberProfileUpdate = Partial<
    Pick<OrganizationMemberProfile, 'role'>
>;
```

### DIFF
--- a/packages/common/src/types/organizationMemberProfile.ts
+++ b/packages/common/src/types/organizationMemberProfile.ts
@@ -37,8 +37,9 @@ export type OrganizationMemberProfile = {
     isInviteExpired?: boolean;
 };
 
-export type OrganizationMemberProfileUpdate = Partial<
-    Pick<OrganizationMemberProfile, 'role'>
+export type OrganizationMemberProfileUpdate = Pick<
+    OrganizationMemberProfile,
+    'role'
 >;
 
 export type ApiOrganizationMemberProfiles = {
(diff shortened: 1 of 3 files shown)


packages/common/src/types/organizationMemberProfile.ts
```diff
@@ -37,9 +37,9 @@ export type OrganizationMemberProfile = {
     isInviteExpired?: boolean;
 };
 
-export type OrganizationMemberProfileUpdate = Partial<
-    Pick<OrganizationMemberProfile, 'role'>
->;
+export type OrganizationMemberProfileUpdate = {
+    role: OrganizationMemberRole;
+};
```
yu-iskw, line 42: We declare the type without `Partial` and `Pick`.

rephus:
Mmm, your change makes sense, but I don't think we can change that type without affecting other parts of the app... Let me run tests

yu-iskw:
@rephus I totally understand it would be better to take advantage of `Pick` for reusability. We can also improve tsoa to generate simpler declarations with `Pick<>`. I will consult with the tsoa community later.

yu-iskw:
FYI: Aside from that, tsoa has another issue to generate a schema. 

https://github.com/lukeautry/tsoa/issues/1191

rephus:
I can't find anything wrong with the change of types. Let's merge !

Issue #5964, opened 2023-06-19
On date filters, the max date is set to `6 months from now` by default


yasmin-e:
### Description

It's the default behaviour of Blueprint's DateInput component as shown in their docs:
<img width="535" alt="Screenshot 2023-06-19 at 14 56 14" src="https://github.com/lightdash/lightdash/assets/67699259/1503e7c2-004b-4638-8a02-7bd601c55ce1">

### To Reproduce
- Add any date filter
- Try to select a date later than 6 months from now
- The date is greyed out
<img width="837" alt="Screenshot 2023-06-19 at 14 57 12" src="https://github.com/lightdash/lightdash/assets/67699259/f052fe22-2b2b-458e-9a29-c483e98c0812">

PR #5967, opened 2023-06-19
fix: date inputs max date

yasmin-e:
Closes: #5964


ZeRego (changes requested):
Don't harcode a date. 
Let have `current date + 5 years` or similar

ZeRego (approved):

Issue #5972, opened 2023-06-19
Once a user has created a custom metric, they can edit it from the three-dot-menu in the results table


joaoviana:
Parent issue: https://github.com/lightdash/lightdash/issues/4523

Similar as being done in parent but allow for `Edit custom metric` in 3-dot menu in Results table.

PR #5973, opened 2023-06-20
feat: can edit filters in table results

joaoviana:
Closes: #5972

### Description:

Users can edit and add filters to a custom metric when viewing it on the table results of an explore. 

screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/388c22cf-093c-4245-8051-ffd4495e6581

### DIFF
--- a/packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
+++ b/packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
@@ -1,9 +1,7 @@
 import { Divider, Menu, Position } from '@blueprintjs/core';
 import { MenuItem2, Popover2 } from '@blueprintjs/popover2';
 import {
-    AdditionalMetric,
     fieldId,
-    isAdditionalMetric,
     isField,
     isFilterableField,
     TableCalculation,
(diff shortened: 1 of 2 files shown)


rephus (approved):
It looks great

rephus (dismissed):
Oh wait, this query breaks if I add a filter to a date custom metric...

[Screencast from 21-06-23 12:24:42.webm](https://github.com/lightdash/lightdash/assets/1983672/8688a2e5-89c6-4cbc-ac60-6e26b864a12e)

TuringLovesDeathMetal:
nice! We;re also still getting that weird overlap of the field type when you re-open the filters in edit mode 🤔 

This might be out of scope (so feel free to open another issue if it is) but can we add a number beside the filters toggle if there are filters applied? (showing how many filters have been applied)

TuringLovesDeathMetal:
image

joaoviana:
> nice! We;re also still getting that weird overlap of the field type when you re-open the filters in edit mode 🤔
> 
> This might be out of scope (so feel free to open another issue if it is) but can we add a number beside the filters toggle if there are filters applied? (showing how many filters have been applied)

Yes @TuringLovesDeathMetal I was trying to work out why it's doing that. I think we need to just replace that component with the Mantine one 😫 

Regarding the number of filters, sure, adding now in a separate PR!

joaoviana:
> Oh wait, this query breaks if I add a filter to a date custom metric...
> 
>  Screencast.from.21-06-23.12.24.42.webm

thanks for raising this @rephus , fixed on #5997 !

packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
```diff
@@ -41,6 +41,25 @@ const ContextMenu: FC<ContextMenuProps> = ({
...
+    const additionalMetrics = useExplorerContext(
+        (context) =>
+            context.state.unsavedChartVersion.metricQuery.additionalMetrics,
+    );
+
+    const additionalMetric = useMemo(
+        () =>
+            !!additionalMetrics &&
+            !!item &&
+            additionalMetrics.find((am) => am.name === item.name),
```
ZeRego, line 53: We should check field IDs since names aren't unique across tables.
```suggestion
            additionalMetrics.find((am) => getItemId(am) === getItemId(item)),
```

joaoviana: 🤦 thanks for spotting this!

ZeRego (approved):

Issue #5942, opened 2023-06-16
Only tables included in the `tables configuration` should be shown in teh validator


TuringLovesDeathMetal:
### Description

Right now, tables that are not actually shown in Lightdash are included in the validator (i.e. tables that have been filtered out using the `tables configuration` settings for a project). 

Ideally, only tables that are actually shown and used in Lightdash should be included in the validator

### Acceptance criteria

- [ ] Tables filtered from Lightdash in teh `tables configuration` should also be filtered from the validator. 
- [ ] Only tables that appear as Explores in Lightdash should have errors shown in the validator

### Questions

- [ ] Can we also apply this to running `lightdash validate` on the CLI?

PR #5981, opened 2023-06-20
fix: only show selected tables on validation

rephus:
Closes: #5942

### How to replicate when selecting tables

I introduce an error on payments table that was appearing on validation.

When  selected customers/orders (not payments): 

Screenshot from 2023-06-20 16-53-17

Before:

Screenshot from 2023-06-20 16-53-12

AFter:

Screenshot from 2023-06-20 16-53-34

Logs: 
Screenshot from 2023-06-20 16-53-46

### How to replicate with tags

I added a tag to orders

Screenshot from 2023-06-20 17-00-46

I selected that tag

No "payment errors":

Screenshot from 2023-06-20 17-00-40

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.mock.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.mock.ts
@@ -16,6 +16,8 @@ import {
     SessionUser,
     ShareUrl,
     SupportedDbtAdapter,
+    TablesConfiguration,
+    TableSelectionType,
 } from '@lightdash/common';
 import { LightdashConfig } from '../../config/parseConfig';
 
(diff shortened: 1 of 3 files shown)


ZeRego:
The problem is that those tables can be joined. So showing the errors would make sense. 
The table selection configuration is just to determine what tables/explores are available in the tables page.

TuringLovesDeathMetal (approved):
niceee! thank you!

ZeRego (changes requested):
(see previous comment)

rephus:
> The table selection configuration is just to determine what tables/explores are available in the tables page.

Yeah, he is right, this mainly affects visibility in the explore, but we still compile and generate errors (that's why they appear on validation) 

Should we then close this issue without merging ? @TuringLovesDeathMetal

rephus:
By default it was not showing "orders" if orders was disabled, but payments was selected (with a join). So I made some changes in the code to show joined tables (but only 1 level, without recursion) 

Screenshot from 2023-06-21 13-15-30

Screenshot from 2023-06-21 13-33-52

ZeRego (approved):
lgtm 👍 
would be good to have a unit test

Issue #5945, opened 2023-06-16
Assert unreachable of Scheduled Delivery with potential `null` format property


joaoviana:
### Description

We currently assume that `format` in `scheduler` table is `notNullable`. 

However, users have reported that scheduled deliveries previously set to send as an `image` get sent as `csv`. This shouldn't be the case as per this migration file https://github.com/lightdash/lightdash/blob/main/packages/backend/src/database/migrations/20230216141637_add_scheduler_format.ts#L8, where we set `format` as a `notNullable` column. 

### Proposed implementation: 

on: `/packages/backend/src/scheduler/SchedulerTask.ts` we should assert unreachable when `format` isn't neither `SchedulerFormat.IMAGE` nor `SchedulerFormat.CSV` in `getNotificationPageData`, providing more logs and preventing the sending of a CSV Scheduled delivery when it was intended to be sent as `image`.

PR #5984, opened 2023-06-20
fix: assert unreachable scheduled delivery format

joaoviana:
Closes: #5945 

### Description:

Read ticket for description ^ 

I believe this is where it needs to be done 👍

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -1,4 +1,5 @@
 import {
+    assertUnreachable,
     CompileProjectPayload,
     DownloadCsvPayload,
     EmailNotificationPayload,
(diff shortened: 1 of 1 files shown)


packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -118,7 +119,7 @@ export const getNotificationPageData = async (
         if (imageUrl === undefined) {
             throw new Error('Unable to unfurl image');
         }
-    } else {
+    } else if (format === SchedulerFormat.CSV) {
```
rephus, line 122: you could also do a `switch` here with `default` being `assertUnreachable`.

Did you look at what's stored in db ?  

Image was first, then we introduced CSV, perhaps IMage should be the default. Perhaps those customers created the schedule before we added CSV, we could check created_at on db

rephus (approved):

packages/backend/src/scheduler/SchedulerTask.ts
joaoviana: Image is actually set by default as per this migration file: https://github.com/lightdash/lightdash/blob/main/packages/backend/src/database/migrations/20230216141637_add_scheduler_format.ts#L8

So I really wonder what happened for users to experience this... 
I will change this to a `switch`, thanks for that!

Issue #5987, opened 2023-06-20
Allow API key authentication on `UpdateOrganizationMember`


yu-iskw:
### Description
`UpdateOrganizationMember` isn't available with the API key authentication due to the lack of `allowApiKeyAuthentication`.
https://lightdash-community.slack.com/archives/C03MG2VTR08/p1687237887940019

```
api_key="xxx"
userUuid="xxx"
url="[https://app.lightdash.cloud/api/v1/org/users/${userUuid}](https://app.lightdash.cloud/api/v1/org/users/$%7BuserUuid%7D)"
curl -X PATCH \
  -H "Content-Type: application/json" \
  -H "Authorization: ApiKey ${api_key}" \
  -d '{"role": "editor"}' \
  "$url"

{"status":"error","error":{"statusCode":401,"name":"AuthorizationError","message":"Failed to authorize user","data":{}}}
```

### What problem does this solve?
If the issue is resolved, we can update roles at organization level with the API.

PR #5988, opened 2023-06-20
feat: allow API key authentication on `UpdateOrganizationMember`

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/5987

### Description:
UpdateOrganizationMember isn't available with the API key authentication due to the lack of allowApiKeyAuthentication.


rephus (approved):

yu-iskw:
Thank you for merging!

yu-iskw:
@rephus I tried to call the API to update an user’s role on the free tier which show the version v0.623.2 . But, I still get the same error `Failed to authorize user`. According to the comment in the PR, the change was released at 0.623.0 .  Do we need something else to enable the API?

PR #5997, opened 2023-06-21
fix: get correct key for date filters

joaoviana:
Relates to: https://github.com/lightdash/lightdash/pull/5973#pullrequestreview-1490373217

Caused by: #5998 

### Description:

Get correct key for date filters. 

Currently, the cached explores return the time format uppercased (e.g. `order_date_DAY`) so the code was breaking for date filters applied to custom metrics due to ref names not matching (`order_date_day` vs `order_date_DAY`). 

Since we can assume that dimensions in an explore `table` are all unique, we can also find the one that matches the lowercased ref.

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.ts
+++ b/packages/common/src/compiler/exploreCompiler.ts
@@ -290,7 +290,7 @@ export class ExploreCompiler {
 
                 const table = tables[refTable];
 
-                // NOTE: date dimensions from explores have their time format uppercased (.e.g. order_date_DAY)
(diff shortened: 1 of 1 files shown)


ZeRego (approved):

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -287,7 +287,18 @@ export class ExploreCompiler {
                     fieldRef,
                     metric.table,
                 );
-                const dimensionField = tables[refTable]?.dimensions[refName];
+
+                const table = tables[refTable];
```
joaoviana, line 291: need to check here if table exists. if not, throw compileError again?

Issue #5507, opened 2023-05-16
I want to be able to automatically assign interactive viewer to users who create accounts using SSO 


rephus:
### Description

Add interactive viewer as an option here: 

![image](https://github.com/lightdash/lightdash/assets/1983672/e3acab4f-f694-432b-86c8-09c534504de6)

Also add also editors/developers/admins

PR #6018, opened 2023-06-22
feat: add editor and interactive viewer to default role options

ZeRego:
Closes: #5507 

### Description:

Note: not adding developer and admin as that options since they have high level of permissions and access. Setting these roles as default role by mistake could cause security issues for that org. We can reconsider this in the future.

Captura de ecrã 2023-06-22, às 11 49 22

### DIFF
--- a/packages/backend/src/models/OrganizationAllowedEmailDomainsModel.ts
+++ b/packages/backend/src/models/OrganizationAllowedEmailDomainsModel.ts
@@ -1,4 +1,8 @@
-import { AllowedEmailDomains } from '@lightdash/common';
+import {
+    AllowedEmailDomains,
+    isAllowedEmailDomainsRole,
+    OrganizationMemberRole,
+} from '@lightdash/common';
 import { Knex } from 'knex';
 import {
     DbOrganizationAllowedEmailDomains,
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/UserSettings/AllowedDomainsPanel/index.tsx
```diff
@@ -14,6 +14,16 @@ import { useProjects } from '../../../hooks/useProjects';
 import { isValidEmailDomain } from '../../../utils/fieldValidators';
 
 const roleOptions = [
+    {
```
owlas, line 17: so the api accepts any role as the default role? Do we want to change that?

ZeRego: Added validation in the API

packages/backend/src/models/OrganizationAllowedEmailDomainsModel.ts
```diff
@@ -23,7 +27,11 @@ export class OrganizationAllowedEmailDomainsModel {
             organizationUuid:
                 dbOrganizationAllowedEmailDomains.organization_uuid,
             emailDomains: dbOrganizationAllowedEmailDomains.email_domains,
-            role: dbOrganizationAllowedEmailDomains.role,
+            role: isAllowedEmailDomainsRole(
```
owlas, line 30: Do we need this check? I assumed the controller should do the validation for us

ZeRego: The column allows any org type. I could have added a migration to create a new table and change this column reference but felt like overkill.

owlas (approved):

Issue #5450, opened 2023-05-11
Add the option to format a table calculation as a currency


TuringLovesDeathMetal:
### Description

When I'm creating a table calculation, I want to be able to format it as a currency.

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVMM2H93o=/?moveToWidget=3458764554142184604&cot=14

- [ ] In the `format` tab, there is another `type` option called `currency`
- [ ] There is a field called `currency` where you select the currency type
    - [ ] let's add the 30 currencies from here to start out with: https://fastspring.com/blog/how-to-format-30-currencies-from-countries-all-over-the-world/
    - [ ] in the drop-down list, you're shown a preview of what the currency format looks like as well as the standard three-letter name
    - [ ] **Can we add some way to search through these currencies from the drop-down menu?**
- [ ] applying a currency format will automatically apply separator styles and prefix/suffixes that are standard for that currency
e.g. for brl (Brazilian Real), an amount in reais is written as R$. R$ 6,00 (with a comma) is six reais and R$ 6.000 (with a period) is six thousand reais. - i.e. we apply the `100.000,00` separator style
- [ ] these values are still treated as numbers when using them/manipulating them in a results table
- [ ] We include `separator style` and `decimal places` like in the percent format type
- [ ] We also add `compact` which is a drop-down list (just like in the big number charts) which will add a suffix based on the compact option selected (e.g. $100,000.00 with the compact option `thousands` would appear like: `$100k`)

### Analytics events

- add `table_calculation_currency_format_count` to `query_executed` event. This tells us how many table calculations had the currency format applied

### Docs

- update table calculation formatting docs to include currency formatting options

PR #6019, opened 2023-06-22
feat: table calculation currency format

rephus:
Closes: #5450

image

### Description

When I'm creating a table calculation, I want to be able to format it as a currency.

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVMM2H93o=/?moveToWidget=3458764554142184604&cot=14

- [x] In the `format` tab, there is another `type` option called `currency`
- [x] There is a field called `currency` where you select the currency type
    - [x] let's add the 30 currencies from here to start out with: https://fastspring.com/blog/how-to-format-30-currencies-from-countries-all-over-the-world/
    - [x] in the drop-down list, you're shown a preview of what the currency format looks like as well as the standard three-letter name
    - [x] **Can we add some way to search through these currencies from the drop-down menu?**
- [x] applying a currency format will automatically apply separator styles and prefix/suffixes that are standard for that currency
e.g. for brl (Brazilian Real), an amount in reais is written as R$. R$ 6,00 (with a comma) is six reais and R$ 6.000 (with a period) is six thousand reais. - i.e. we apply the `100.000,00` separator style
- [x] these values are still treated as numbers when using them/manipulating them in a results table
- [x] We include `separator style` and `decimal places` like in the percent format type
- [x] We also add `compact` which is a drop-down list (just like in the big number charts) which will add a suffix based on the compact option selected (e.g. $100,000.00 with the compact option `thousands` would appear like: `$100k`)

### Analytics events

- [x] add `table_calculation_currency_format_count` to `query_executed` event. This tells us how many table calculations had the currency format applied

### Docs

- [x] update table calculation formatting docs to include currency formatting options
- [ ] 
image


rephus:
Not sure what this error is, I'll investigate


image

packages/backend/src/generated/swagger.json
```diff
@@ -715,7 +715,6 @@
                     "scatter",
                     "area",
                     "mixed",
-                    "pie",
```
ZeRego, line 718: fyi: I think we are rolling back changes from main.

ZeRego (changes requested):
~Currency format not showing in the big number.~

ps: ignore, I was using the metric instead of the table calc in the big number

ZeRego (approved):
lgtm

ps: ignore my previous review, I was using the metric instead of the table calc in the big number

packages/backend/src/generated/swagger.json
rephus: let me rebase from main again

rephus:
I don't think these e2e errors are related with my changes, also they happened after I rebased from main. 

I'm going to merge and take a look at those tests on main

Issue #6022, opened 2023-06-22
View underlying data - broken for metrics with filters that target `fieldId`s instead of `fieldRef`s


joaoviana:
### Description

App breaks when viewing underlying data for metrics that contain filters (and those filters are `target.fieldId` instead of `target.fieldRef`)

### To Reproduce

Create query
Add metric (that includes a filter)
Click to view `Underlying data` 
App breaks because the modal expects `filter.target.fieldRef` to be there, when it can be undefined and `fieldId` is defined.

PR #6023, opened 2023-06-22
fix: get correct fieldId when viewing underlying data

joaoviana:
Closes: #6022

### Description:

Loop through explores from cache in `getExploreFromCache` and add `fieldRef` to metric.filters. 
Convert those refs back to `fieldId`s when viewing underlying data.
Those same `fieldRef`s might, or might not have a table name prefix. 
1. _standard_ filters' targets have only the field name (e.g.: `is_completed`)
2. filters from joins have `tableName.fieldName` (e.g.: `orders.is_completed`)

Because of 1. we add an optional `fallbackTableName` property to `convertFieldRefToFieldId` for `fieldRef`s that have the table name missing (when viewing underlying data of metric filters, where a `target.fieldRef` has the table missing)


owlas (changes requested):
Isn't this a change we should make on the backend when loading saved charts with old style custom metrics, and fix it further upstream (in case there are other FE dependencies that haven't been updated)

joaoviana:
> Isn't this a change we should make on the backend when loading saved charts with old style custom metrics, and fix it further upstream (in case there are other FE dependencies that haven't been updated)

but these are metrics with filters applied 🤔  but agree with changing on the BE, looking into it!

packages/backend/src/models/ProjectModel/ProjectModel.test.ts
```diff
@@ -77,4 +79,62 @@ describe('ProjectModel', () => {
...
+
+    describe('should update outdated metric filters in explores', () => {
+        const fieldRefsToTest: {
+            fieldRef: string;
+            tableName: string;
+            isOutdated: boolean;
+        }[] = [];
+        const convertedExplore =
+            ProjectModel.convertMetricFiltersFieldIdsToFieldRef(
+                mockExploresFromCache[0],
```
joaoviana, line 91: can remove the other explores and just keep the first one in `ProjectModel.mock`. What do we think?

ZeRego (approved):

PR #6025, opened 2023-06-22
fix: dashboard URL should update with filter values

yasmin-e:
Closes: #5817 

### Description:
- found a syntax bug, setting values to the wrong state
- we were overriding the filters on dashboard load with dash saved filters

### DIFF
--- a/packages/frontend/src/providers/DashboardProvider.tsx
+++ b/packages/frontend/src/providers/DashboardProvider.tsx
@@ -133,11 +133,11 @@ export const DashboardProvider: React.FC = ({ children }) => {
         useState<boolean>(false);
 
     useEffect(() => {
-        if (dashboard) {
+        if (dashboard && dashboardFilters === emptyFilters) {
             setDashboardFilters(dashboard.filters);
             setHaveFiltersChanged(false);
         }
-    }, [dashboard]);
+    }, [dashboardFilters, dashboard]);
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -273,7 +272,9 @@ export const DashboardProvider: React.FC = ({ children }) => {
         ) {
             newParams.delete('filters');
         } else {
-            newParams.set('filters', JSON.stringify(dashboardTemporaryFilters));
+            newParams.set('filters', JSON.stringify(dashboardFilters));
+            const searchParams = new URLSearchParams(search);
+            console.log(searchParams.get('filters'));
```
rephus, line 277: leftover console log

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -273,7 +272,9 @@ export const DashboardProvider: React.FC = ({ children }) => {
         ) {
             newParams.delete('filters');
         } else {
-            newParams.set('filters', JSON.stringify(dashboardTemporaryFilters));
+            newParams.set('filters', JSON.stringify(dashboardFilters));
+            const searchParams = new URLSearchParams(search);
```
rephus, line 276: You can remove this as well

Issue #5869, opened 2023-06-13
I'm able to set a default project for new users in my organization settings. 


TuringLovesDeathMetal:
### Description

New users that are added to an organization don't always land in the right project. We should have an option for setting a default project in the organization settings so users land in this project by default. 

### Acceptance criteria:

Miro sketch: https://miro.com/app/board/uXjVM_KRilE=/?share_link_id=236055190221

- [ ] There is a section to set a default project in the organization settings page - it is a project picker.
- [ ] When a user first joins the org, they open Lightdash in the default project
- [ ] The deafult project is only used when the user first logs in, afterwards, we use the same project logic as we have now (i.e. the last project you were looking at is opened again)
- [ ] If a user doesn't have access to the default project, then they open Lightdash to the project that was created _**first**_ out of the projects they have access to
- [ ] By default, the default project is not selected and the empty state says "No project selected"

### Docs

- [ ] Update the roles and permissions doc here to include information about the `default project`

PR #6030, opened 2023-06-23
feat: organization default project

ZeRego:
Closes: #5869 

### Acceptance criteria:

- [x] There is a section to set a default project in the organization settings page - it is a project picker.
- [x] When a user first joins the org, they open Lightdash in the default project
- [x] The default project is used when the user first logs in, afterwards, we use the same project logic as we have now (i.e. the last project you were looking at is opened again)
- [x] If a user doesn't have access to the default project, then they open Lightdash to the project that was created _**first**_ out of the projects they have access to
- [x] By default, the default project is not selected and the empty state says "No project selected"

### Docs

- [x] Update the roles and permissions doc here to include information about the `default project`

Default state:

Captura de ecrã 2023-06-23, às 10 14 59

Setting default project + new users joining:

https://github.com/lightdash/lightdash/assets/9117144/21f16486-0e96-4042-aedd-797a8f9436fd

### DIFF
--- a/packages/backend/src/database/migrations/20230623082425_add-prg-default-project.ts
+++ b/packages/backend/src/database/migrations/20230623082425_add-prg-default-project.ts
@@ -1,0 +1,19 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('organizations', (table) => {
+        table
+            .uuid('default_project_uuid')
+            .references('project_uuid')
+            .inTable('projects')
+            .nullable()
+            .unique()
+            .onDelete('SET NULL');
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('organizations', (table) => {
+        table.dropColumn('default_project_uuid');
+    });
+}
--- a/packages/backend/src/database/migrations/20230623082425_add-prg-default-project.ts
+++ b/packages/backend/src/database/migrations/20230623082425_add-prg-default-project.ts
@@ -7,7 +7,6 @@ export async function up(knex: Knex): Promise<void> {
             .references('project_uuid')
             .inTable('projects')
             .nullable()
-            .unique()
             .onDelete('SET NULL');
     });
 }
--- a/packages/backend/src/models/OrganizationModel.ts
+++ b/packages/backend/src/models/OrganizationModel.ts
@@ -25,6 +25,9 @@ export class OrganizationModel {
             organizationUuid: data.organization_uuid,
             name: data.organization_name,
             chartColors: data.chart_colors,
+            defaultProjectUuid: data.default_project_uuid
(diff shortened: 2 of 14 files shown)


packages/backend/src/database/migrations/20230623082425_add-prg-default-project.ts
```diff
@@ -0,0 +1,19 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('organizations', (table) => {
+        table
+            .uuid('default_project_uuid')
+            .references('project_uuid')
+            .inTable('projects')
+            .nullable()
+            .unique()
```
owlas, line 10: why unique?

packages/backend/src/models/OrganizationModel.ts
```diff
@@ -25,6 +25,9 @@ export class OrganizationModel {
             organizationUuid: data.organization_uuid,
             name: data.organization_name,
             chartColors: data.chart_colors,
+            defaultProjectUuid: data.default_project_uuid
```
owlas, line 28: do we have this ternary because default project uuid can have `null` instead of `undefined`? I'm never sure exactly how `node-pg` handles nulls in postgres

owlas (commented):
couple questions - just doing Qa

owlas (changes requested):
This changes my current project when I update the settings.

CleanShot 2023-06-23 at 11 31 18

packages/backend/src/models/OrganizationModel.ts
ZeRego: yes, the query returns `null` values.

packages/backend/src/database/migrations/20230623082425_add-prg-default-project.ts
ZeRego: removed

ZeRego:
@owlas nice catch :) that happens when the user doesn't have any project uuid saved in localstorage. We only update the local storage in a few situations ( eg: when user switch project in nav bar, when going to project settings ) 
I changed the code so it updates the local storage as soon as possible. When the page loads, it updates local storage if is empty.

packages/frontend/src/hooks/useActiveProject.ts
```diff
@@ -63,22 +65,44 @@ export const useActiveProjectUuid = () => {
...
+
+    const lastProject = projects?.find(
+        (project) => project.projectUuid === lastProjectUuid,
+    );
+
+    useEffect(() => {
+        const newValue =
+            paramProject?.projectUuid || defaultProject?.projectUuid;
+        if (!isLoading && !lastProject && newValue) {
+            mutate(newValue);
```
ZeRego, line 85: Update local storage as soon as possible if there it is empty.

owlas (approved):
working

PR #6034, opened 2023-06-23
fix: only set url params with pristine state

ZeRego:
Closes: 

### Description:

- only update explore url params where there is a pristine state ( aka: when a use runs a query )
- clear the url params when user hits `command + option + k`


rephus (approved):
If e2e tests work (or at least the ones that were failing) then it is good to go !

joaoviana (approved):

ZeRego:
Ignoring e2e test for dashboard. We know what the problem is and are fixing it in a separate PR

Issue #5767, opened 2023-06-06
I wish I could see the chart type in the `add charts` modal in a dashboard


TuringLovesDeathMetal:
### Description

When I'm adding tiles to a dashboard, if things are named similarly it can be hard to figure out which chart is which. One thing that would help with this is actually having the chart type icon in the modal (the same way we do in search).

<img width="516" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/e1bfb4e3-b502-4fb6-b812-8eb67b482ed7"> 

### Acceptance criteria

- [ ] When I add charts to my dashboard, I can see the chart type in the list of results.

yasmin-e:
⚠️⚠️  We need to add `ChartConfig` property to `useChartSummaries`, this way we can apply `getChartType()` method on the charts and show the corresponding icon.

joaoviana:
This is done now with after: https://github.com/lightdash/lightdash/pull/10758
With the `v2` endpoint for `/content` we can get the `chartKind`!

PR #6036, opened 2023-06-23
chore: amend dashboard test

ZeRego:
Closes: 

### Description:

Problem:
Captura de ecrã 2023-06-23, às 16 51 50

e2e test was failing because the `Add` button was behind the dropdown. 
Added command to close it.

PR #6037, opened 2023-06-23
feat: add chart type icon when adding to dashboard

joaoviana:
Closes: #5767 

### Description:

Make modal bigger;
Get chart `config` from chart summaries so that we get the chart kind to render the icon
Grey out icon if chart is already in dashboard. 

https://github.com/lightdash/lightdash/assets/7611706/5a663aab-e6e7-4351-ac3b-5d29067d82c9

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
@@ -107,7 +107,7 @@ const AddChartTilesModal: FC<Props> = ({ onAddTiles, onClose }) => {
                     group: spaceName,
                     disabled: alreadyAddedChart !== undefined,
                     description: alreadyAddedChart
-                        ? 'This chart has been already added to this dashboard'
+                        ? 'This chart has already been added to this dashboard'
                         : undefined,
                     ...(chartConfig &&
                         chartType && {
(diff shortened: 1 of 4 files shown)


IrakliJani:
@joaoviana is there a separate issue to fix e2e test? I believe error inside `Should create dashboard with tiles` is caused because of the modal changes

joaoviana:
> @joaoviana is there a separate issue to fix e2e test? I believe error inside `Should create dashboard with tiles` is caused because of the modal changes

I think this: https://github.com/lightdash/lightdash/pull/6036 should fix it 🤔 what do you think @IrakliJani ?

packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
```diff
@@ -65,24 +92,30 @@ const AddChartTilesModal: FC<Props> = ({ onAddTiles, onClose }) => {
...
-                    : undefined,
-            };
-        });
+                return {
+                    value: uuid,
+                    label: name,
+                    group: spaceName,
+                    disabled: alreadyAddedChart !== undefined,
+                    description: alreadyAddedChart
+                        ? 'This chart has been already added to this dashboard'
```
IrakliJani, line 110: grammar correction
```suggestion
                        ? 'This chart has already been added to this dashboard'
```

IrakliJani (approved):
looks awesome! added comment about grammar and comment about e2e

packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
joaoviana: oh 🤦 thank youuu

Issue #6021, opened 2023-06-22
User Management page : UI bug when a dropdown is open and you scroll down the page


PriPatel:
<img width="707" alt="Screenshot 2023-06-22 at 12 11 52" src="https://github.com/lightdash/lightdash/assets/12776522/31c72a8d-0f9d-4cdc-ac0e-385263c22073">

z-index issue. When the dropdown is open and I scroll down the page, it remains in front of the navbar. It makes things look broken.

anjali29singh:
can you assign this issue to me ?

jayanand05:
Got the issue!

jayanand05:
it's working fine now

https://github.com/lightdash/lightdash/assets/113882904/88415b6e-58a4-4ca8-8045-b85b04a276d3

PR #6040, opened 2023-06-25
fix: increased the  zIndex value

jayanand05:
Closes: #6021 

### Description:
z-index issue. When the dropdown was open and I scroll down the page, it remained in front of the navbar, increase the zIndex value to fix it.

https://github.com/lightdash/lightdash/assets/113882904/daf6e0dd-769a-47d6-a098-4b20abab2667


packages/frontend/src/components/NavBar/index.tsx
```diff
@@ -72,7 +72,7 @@ const NavBar = memo(() => {
                 display="flex"
                 px="md"
                 // FIXME: adjust after removing Blueprint
-                zIndex={10}
+                zIndex={999}
```
joaoviana, line 75: I've just checked the `zIndex` applied to the `Select` dropdown and it's `300`. Could we go for `500` here? What do you think? 😃

joaoviana (changes requested):
Thanks for contributing @jayanand05 🙌 

Could you remove the changes to the `yarn.lock` file? 👍

packages/frontend/src/components/NavBar/index.tsx
jayanand05: I checked it's working fine with 300 as well we can go for 500 but 300 is also that we can stick with.

joaoviana: let's go with `300` then pls @jayanand05 👍

jayanand05: 👍👍

joaoviana:
One final thing @jayanand05, could you re-add the `yarn.lock` please? 
Ensure you are on node 16, and then run `yarn`. The file should be added again. 
Or you can remove the commits that made that change? Thanks!

packages/frontend/src/components/NavBar/index.tsx
jayanand05: I Updated necessary changes request you to look into it?

joaoviana:
@all-contributors add <github-handle> for code

> can you tell me where to re-add yarn.lock file ?

sure! 
so on the file changes for this PR, you can see that `yarn.lock` was deleted on the root folder: 
image

You can bring it back in 2 ways: 
* delete commits (with rebase) that changed that file 
or 
* re-run `yarn` in root: `/`

jayanand05:
Hi @joaoviana I ran the yarn in root how to check re-adding was successful?

joaoviana:
> Hi @joaoviana I ran the yarn in root how to check re-adding was successful?

try to commit that `yarn.lock` file and push to your branch. Check this PR - the files changed - and you should see only 1 change: the increase of the `zIndex`!

joaoviana:
@jayanand05 `zIndex` is still `999`, could you change to `300`? 
also, there's an empty line in `yarn.lock`, maybe remove it? and push it. Then you should end up with just one file change 👍

owlas:
@all-contributors add @jayanand05 for code

owlas:
duplicated and fixed on: https://github.com/lightdash/lightdash/pull/6049

Issue #6005, opened 2023-06-21
Add echarts visualization for pie charts in a visualization card


IrakliJani:
- [ ] add a echarts pie chart visualization based on the configuration mentioned here https://github.com/lightdash/lightdash/issues/6004

Issue #6007, opened 2023-06-21
Add configuration options for pie charts to show value labels as `value`, `% of total` or both


IrakliJani:
- [ ] If you turn on the value labels, there is the option to have the value labels show as value or % of total or both
- [ ] you only see these options if you have selected to show value labels
- [ ] If you select **both** then the two values appear like `xx % - yy` where `yy` is the value and `xx` is the `%` of total

PR #6041, opened 2023-06-26
feat: pie charts echarts visualization

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6005

### Description:

CleanShot 2023-06-26 at 12 32 30@2x

https://github.com/lightdash/lightdash/assets/962095/bc6d5b89-f620-4673-b4f1-2f6d53f3ff55

### DIFF
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -11,7 +11,6 @@ import {
     getDimensions,
     getMetrics,
     isNumericItem,
-    isTableCalculation,
     Metric,
     TableCalculation,
 } from '@lightdash/common';
@@ -128,8 +127,6 @@ const VisualizationProvider: FC<Props> = ({
         );
     }, [explore, resultsData?.metricQuery.dimensions]);
 
-    const dimensionIds = useMemo(() => dimensions.map(fieldId), [dimensions]);
-
     const metrics = useMemo(() => {
         if (!explore) return [];
         return getMetrics(explore).filter((field) =>
@@ -167,12 +164,6 @@ const VisualizationProvider: FC<Props> = ({
         () => allMetrics.filter((m) => isNumericItem(m)),
         [allMetrics],
     );
-
-    const allNumericMetricIds = useMemo(() => {
-        return allNumericMetrics.map((m) =>
-            isTableCalculation(m) ? m.name : fieldId(m),
-        );
-    }, [allNumericMetrics]);
 
     const bigNumberConfig = useBigNumberConfig(
         initialChartConfig?.type === ChartType.BIG_NUMBER
@@ -241,8 +232,8 @@ const VisualizationProvider: FC<Props> = ({
         initialChartConfig?.type === ChartType.PIE
             ? initialChartConfig.config
             : undefined,
-        dimensionIds,
-        allNumericMetricIds,
+        dimensions,
+        allNumericMetrics,
     );
 
     const { validPieChartConfig } = pieChartConfig;
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -239,10 +239,11 @@ const VisualizationProvider: FC<Props> = ({
     const { validCartesianConfig } = cartesianConfig;
 
     const pieChartConfig = usePieChartConfig(
+        explore,
+        resultsData,
         initialChartConfig?.type === ChartType.PIE
             ? initialChartConfig.config
             : undefined,
-        explore,
         dimensionIds,
         allMetricIds,
     );
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -10,6 +10,8 @@ import {
     fieldId,
     getDimensions,
     getMetrics,
+    isNumericItem,
+    isTableCalculation,
     Metric,
     TableCalculation,
 } from '@lightdash/common';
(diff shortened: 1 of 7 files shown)


owlas (changes requested):
This crashes if the metrics / dimensions are changed and the query is re-executed: 

```
Uncaught TypeError: Cannot read properties of undefined (reading 'value')
    at index.tsx:47:48
    at Array.map (<anonymous>)
    at index.tsx:47:18
    at Array.map (<anonymous>)
    at index.tsx:45:34
    at Object.uE [as useMemo] (react-dom.production.min.js:168:377)
    at ue.useMemo (react.production.min.js:23:113)
    at index.tsx:42:18
    at Kp (react-dom.production.min.js:157:137)
    at Mf (react-dom.production.min.js:180:154)
```

CleanShot 2023-06-28 at 16 22 40

IrakliJani:
@owlas fixed the bug that you caught

owlas (commented):
Is it expected to render like this if I only choose a metric? 

CleanShot 2023-06-29 at 11 25 03@2x

IrakliJani:
@owlas good question. I expected since there is one result it should render something. but does not make sense. especially without label

IrakliJani:
@owlas initial spec suggest that we don't show a chart in that case. removed!

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -51,6 +51,7 @@ type VisualizationContext = {
     isSqlRunner: boolean;
     dimensions: Dimension[];
     metrics: Metric[];
+    allMetrics: (Metric | TableCalculation)[];
```
owlas, line 54: what's the difference between this and `metrics` and `tableCalculations` ? 

Is it different to `allMetrics = [...metrics, ...tableCalculations]` ? If yes it should be named differently

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -124,13 +125,17 @@ const VisualizationProvider: FC<Props> = ({
         );
     }, [explore, resultsData?.metricQuery.dimensions]);
 
+    const dimensionIds = useMemo(() => dimensions.map(fieldId), [dimensions]);
```
owlas, line 128: how come so much is using `useMemo` this won't be that slow

packages/frontend/src/components/PieChartConfig/PieChartLayoutConfig.tsx
```diff
@@ -61,10 +59,10 @@ const PieLayoutConfig: React.FC = () => {
     } = useVisualizationContext();
 
     const selectedMetric = useMemo(() => {
-        return [...metrics, ...customMetrics, ...tableCalculations].find((m) =>
```
owlas, line 64: perhaps this should live on the viz provider

packages/frontend/src/hooks/usePieChartConfig.ts
```diff
@@ -24,120 +15,143 @@ type PieChartConfig = {
...
-    fields: {
-        dimensions: Dimension[];
-        metrics: Metric[];
-        customMetrics: AdditionalMetric[];
-        tableCalculations: TableCalculation[];
-    },
+    explore: Explore | undefined,
+    dimensionIds: string[],
+    allMetricIds: string[],
+    // resultsData: ApiQueryResults | undefined,
```
owlas, line 21: Do we still need this?

owlas (commented):
just some last questions on the code!

owlas (dismissed):
There's an error on save, I lose my configuration after saving the chart.

CleanShot 2023-06-29 at 14 18 22

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
IrakliJani: im not sure what to name this tbh. we keep using name metrics for metrics, additional metrics and table calculations...

IrakliJani:
no fields selected
CleanShot 2023-06-30 at 21 50 45@2x

one dimension selected
CleanShot 2023-06-30 at 21 53 07@2x

one metric selected
CleanShot 2023-06-30 at 21 55 44@2x

one dimension and one metric selected
CleanShot 2023-06-30 at 21 56 30@2x

two dimensions selected and one metric but only one group:
CleanShot 2023-06-30 at 21 56 54@2x

two dimensions selected and one metric but both groups:
CleanShot 2023-06-30 at 21 57 51@2x

packages/frontend/src/components/PieChartConfig/PieChartLayoutConfig.tsx
IrakliJani: allMetrics live inside `pieChartConfig` and in an updated config `allMetrics` comes from visualization context

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
IrakliJani: it's better to have variables as memos inside a context to avoid rerenderings of components that are subscribed to context. besides, this particular operation does not cost much

IrakliJani:
using custom metric:
CleanShot 2023-07-03 at 13 23 03@2x

using table calculation:
CleanShot 2023-07-03 at 13 22 49@2x

IrakliJani:
2 e2e tests that failed are not related to code changes.

TuringLovesDeathMetal:
Niiiice! All the pictures look great :) 

I'm not able to see the pie chart option on the chart type panel 🤔 but I'm not sure if it's just not added in there yet? 
image

IrakliJani:
@TuringLovesDeathMetal it's behind a "feature flag" so I can keep shipping. you can enable it by running `localStorage.setItem('enablePieCharts', true)` in the browser console and refreshing the page

ZeRego:
We should only have numeric metrics available in the pie config:
Screenshot 2023-07-03 at 14 27 38

Do we need a special message/state when the values are 0 ? or is this a bug where we check for undefined values but don't handle properly `0` ?
Screenshot 2023-07-03 at 14 28 13

IrakliJani:
@ZeRego 👍 
- We should only have numeric metrics available in the pie config < will fix it here
- Do we need a special message/state when the values are 0 ? or is this a bug where we check for undefined values but don't handle properly 0 ? < will check and fix here
- to keep this PR small I will handle formatting here https://github.com/lightdash/lightdash/issues/6007

IrakliJani:
@ZeRego
- filtered out non-numeric items:
CleanShot 2023-07-03 at 18 05 26@2x
- fixed 0 values. but it looks bit confusing since all values are 0
CleanShot 2023-07-03 at 18 17 47@2x

Issue #6046, opened 2023-06-26
Dashboards with long URLs break when size limit is exceeded


joaoviana:
### Description

Users have been reporting `431` errors when refreshing Dashboards with loooong URLs. These URLs now include `filters` which increase the length significantly. 

### To Reproduce

1. Go to a Dashboard
2. Apply a lot of filters 
3. Save
4. Refresh
5. see 431 error

Address the size of the URL by finding a way to decrease it - possibly through `tileTargets` and unnecessary bloat added to the search params.

joaoviana:
Ideally we should also work on: https://github.com/lightdash/lightdash/issues/5929
once this issue has been mitigated

PR #6047, opened 2023-06-26
Fix/address tile targets data on dashboard url params

joaoviana:
Closes: #6046

### DIFF
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -99,6 +99,15 @@ export type DashboardFilters = {
     metrics: DashboardFilterRule[];
 };
 
+export type DashboardFiltersFromSearchParams = {
+    dimensions: (Omit<DashboardFilterRule, 'tileTargets'> & {
+        tileTargets?: string[];
+    })[];
+    metrics: (Omit<DashboardFilterRule, 'tileTargets'> & {
+        tileTargets?: string[];
+    })[];
+};
+
 /* Utils */
 
 export const isOrFilterGroup = (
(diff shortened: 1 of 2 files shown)


packages/common/src/types/filter.ts
```diff
@@ -171,4 +180,71 @@ export const convertDashboardFiltersToFilters = (
...
+): DashboardFiltersFromSearchParam =>
+    Object.entries(dashboardFilters).reduce(
+        (result, [key, value]) => ({
+            ...result,
+            [key]: value.map((f) => ({
+                ...f,
+                ...(f.tileTargets && {
+                    tileTargets: Object.entries(f.tileTargets).map(
+                        ([tileTargetKey, tileTargetValue]) =>
+                            tileTargetValue.fieldId === f.target.fieldId
```
owlas, line 240: missign a check for `tileTargetValue.tableName === f.target.tableName`

owlas (changes requested):
Works beautifully and great function names given that it's a bit of an obscure transformation. We could even call it `compressDashboardFiltersToParam` so it's clear we're just trying to compress information for the url.

Although the `fieldId` is usually unique, we're missing a check that the `tableName` in one of the conversion functions.

owlas:
Oh one limitation I've noticed is that is you remove all filters, they're re-added on load. But this is also true in prod already so let's leave it

packages/common/src/types/filter.ts
joaoviana: sure, thanks! Adding that now and changing the name of the function!

joaoviana:
> Oh one limitation I've noticed is that is you remove all filters, they're re-added on load. But this is also true in prod already so let's leave it

ok sure, should add an issue for this right 🤔

owlas (approved):

Issue #6052, opened 2023-06-27
Enable to pass the created project name to `lightdash deploy`


yu-iskw:
### Description
We would like to pass the created project name to the `lightdash deploy` command.

### What problem does this solve?
We can interactively name a new project. If we can pass a new project name to the CLI options, we can deploy it with CI and so on.

PR #6053, opened 2023-06-27
feat: allow user to pass project name on cli deploy

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/6052

### Description:
We would like to pass the created project name to a `lightdash deploy --create` command.


packages/cli/src/index.ts
```diff
@@ -369,6 +369,11 @@ program
 program
     .command('deploy')
     .description('Compile and deploy Lightdash project')
+    .option(
+        '--name [project name]',
+        'Name for the created project. If a project with this name already exists, the deployment is failed',
```
owlas, line 374: Why do you want this to fail if the project name already exists? We don't enforce this anywhere else

yu-iskw: @owlas Thank you for the comment. I understand. How about updating the existing project whose name is the same as the given project name? As far as I know, the start-preview command takes the approach.

yu-iskw: I have tentatively implemented   https://github.com/lightdash/lightdash/pull/6053/commits/e7a8ef0a76d1aed124b59e9d44521009cd4681a1 so that we can update a project with the given name which already exists.

owlas: @yu-iskw did you only want this for creating new projects or did you want to update an existing project?

yu-iskw: @owlas At first, I merely wanted to start a new project. However, if we can create and update a project with a single CI job, just like with GitHub Actions, even with the '--name' option. That is, if a project with the passed name by `--name` doesn't exist yet, the API enables us to create a new one. Meanwhile, if the project already exists, we can update it. What do you think?

owlas: My suggestion, which is a smaller change to the current CLI api:

### Currently

We have the `--create` flag which will interactively prompt for a new project name. Then it sets the active project to the created project. So future lightdash deploys will change update the new project

```
lightdash deploy --create
lightdash deploy
```

### Proposed change

Add an optional name argument to the `--create` flag. So a new project will always be created but if the name isn't given then the user will be prompted for a name interactively:

```
# non-interactive create
lightdash deploy --create my-new-project

# interactive create
lightdash deploy --create

# deploy currently active project
lightdash deploy
```

yu-iskw: That sounds good. And I made sure the commander module enables us to implement such a behavior. 

```typescript
program
  .option('--create [project name]', 'If set, create a new project', undefined);
console.log(program.opts());
```

```basn
node index.js --create test
{ create: 'test' }

node index.js --create
{ create: true }

node index.js
{ create: undefined }
```

Let me clarify one thing. Do you want to allow to create a new project name whose name is used for another existing one? Personally, the same project names potentially causes confusing. So, I prefer disallowing creating a new project whose name is already used. What do you think?

owlas: We don't have this restriction anywhere else in the app when creating / renaming projects, so it would be hard to enforce even if we added the restriction here. 

So I wouldn't add the restriction, and let the user decide

yu-iskw: I got it. I will modify the implementation not to check duplication. Thank you!

yu-iskw: @owlas Can you make sure the changed code behaves as you intended? Thanks!

owlas (approved):
works in all 3 cases:

### interactive create
CleanShot 2023-07-04 at 10 46 55@2x

### name provided on cli

CleanShot 2023-07-04 at 10 47 46@2x

### non-interactive default

CleanShot 2023-07-04 at 10 51 11@2x

yu-iskw:
@owlas thank you for your follow-up!

Issue #6024, opened 2023-06-22
Migrate hotkeys from blueprint to mantine


ZeRego:
### Description

Migrate all uses of blueprint hotkeys provider and hooks ( [docs](https://blueprintjs.com/docs/#core/hooks/use-hotkeys) )  with mantine hotkeys hook ( [docs](https://mantine.dev/hooks/use-hotkeys/) )

ankur0904:
@IrakliJani Can I work on this issue?

IrakliJani:
@ankur0904 go for it!

PR #6063, opened 2023-06-27
feat: Migrate hotkeys hooks to mantine

muhsinkamil:
Closes:  https://github.com/lightdash/lightdash/issues/6024

### Description:

- Migrate `useHotkeys` hook to mantine

**Note to reviewers: 
               - I haven't checked on non-mac system, can take a look if it fails on windows**

### DIFF
--- a/packages/frontend/src/components/RefreshButton.tsx
+++ b/packages/frontend/src/components/RefreshButton.tsx
@@ -1,6 +1,7 @@
-import { Classes, HotkeyConfig, KeyCombo, useHotkeys } from '@blueprintjs/core';
+import { Classes } from '@blueprintjs/core';
 import { Tooltip2 } from '@blueprintjs/popover2';
-import { memo, useCallback, useMemo } from 'react';
+import { useHotkeys } from '@mantine/hooks';
+import { memo, useCallback } from 'react';
 import { useExplorerContext } from '../providers/ExplorerProvider';
 import { useTracking } from '../providers/TrackingProvider';
 import { EventName } from '../types/Events';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/RefreshButton.tsx
```diff
@@ -20,54 +21,37 @@ export const RefreshButton = memo(() => {
...
-                ) : (
-                    <KeyCombo combo="mod+enter" />
-                )
+                !hasUnfetchedChanges
+                    ? 'You need to make some changes before running a query'
+                    : ''
             }
             position="bottom"
-            disabled={isLoading || !isValidQuery}
+            disabled={canRunQuery}
```
ZeRego, line 45: We want to show the tooltip if `hasUnfetchedChanges` is false.

packages/frontend/src/components/RefreshButton.tsx
```diff
@@ -20,54 +21,37 @@ export const RefreshButton = memo(() => {
...
         <Tooltip2
             content={
-                !hasUnfetchedChanges ? (
-                    'You need to make some changes before running a query'
-                ) : (
-                    <KeyCombo combo="mod+enter" />
-                )
+                !hasUnfetchedChanges
+                    ? 'You need to make some changes before running a query'
+                    : ''
```
ZeRego, line 42: @PriPatel Mantine doesn't have a component to show key combos like blueprint had. Are we happy to remove them from our app our should we build our own component ?


Captura de ecrã 2023-06-28, às 10 37 59

ZeRego (changes requested):
Thanks for contributing again @muhsinkamil :) 

There are a few more places that we need to replace the hook 

Captura de ecrã 2023-06-28, às 10 40 59

And we should also remove the provider

Captura de ecrã 2023-06-28, às 10 41 52

packages/frontend/src/components/RefreshButton.tsx
PriPatel: @ZeRego  In that case, we should build our own as this is useful information to show (and for future shortcuts as well). Shall I open a ticket for this? I guess it will need design work first.

ZeRego: @PriPatel found this mantine component https://mantine.dev/core/kbd/ that we can use. 

@muhsinkamil we could create a `HotKey` component that parses the string with `parseHotkey` and then render different symbols/keys depending on the user OS. 

```import { parseHotkey } from '@mantine/hooks/lib/use-hotkeys/parse-hotkey';```

We use similar logic as the blueprint component, how it maps the keys to icons/strings https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/hotkeys/keyComboTag.tsx

ZeRego: @muhsinkamil Maybe we should leave the current blueprint component for now and open a new ticket to migrate those components in a separate PR. What do you think ?

muhsinkamil: > Maybe we should leave the current blueprint component for now and open a new ticket to migrate those components in a separate PR.

@ZeRego You mean a separate PR for parsing logic ? If so, I agree, I can try to make it in separate PR

ZeRego: A separate PR to replace the HotKeys component. So we should leave the current one.

muhsinkamil:
@ZeRego Thanks for the review !

Resolved comments in latest commits: d1381c37a13c13387ca773b547f2c948752a3ea3 & 474a43da26b6893b20646e29495c666c284b6e24

Leaving KeyCombo component parse logic for now, Will try to take a look separately

ZeRego (approved):

PR #6045, opened 2023-06-26
feat: migrate visualization card actions to Mantine

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6043

### Description:

bonus: upgrades mantine to fix this issue: `[@mantine/core] Popover: Fix incorrect closeOnClickOutside logic (https://github.com/mantinedev/mantine/pull/4148)`

https://github.com/lightdash/lightdash/assets/962095/7a442121-2233-4ba0-845b-49828e605dde

PR #6064, opened 2023-06-27
fix: correct color picker z index issue

joaoviana:
Closes: N/A

### Description:

Fix z-index issue with chart config panels where any component that renders a popover or a dropdown gets displayed behind the panel. 

Disable `closeOnClickOutside` to avoid closing the config panel when selecting an item. 
NOTE: this means that when users click on other `tabs` like `Export as` the current tab doesn't close automatically. But that can be done separately because it doesn't block functionality.

### DIFF
--- a/packages/frontend/src/components/BigNumberConfig/index.tsx
+++ b/packages/frontend/src/components/BigNumberConfig/index.tsx
@@ -14,7 +14,12 @@ const BigNumberConfigPanel: React.FC = () => {
     const disabled = !selectedField;
 
     return (
-        <Popover {...COLLAPSABLE_CARD_POPOVER_PROPS} disabled={disabled}>
+        <Popover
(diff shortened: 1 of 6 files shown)


hamzahc1:
@joaoviana - the same issue is affecting all the dropdown menus in the config panel 😓 


In the Layout section when selecting x-axis, y-axis and group values:
CleanShot 2023-06-27 at 23 36 23@2x


In the Display tab when searching for fields to apply reference lines: 
CleanShot 2023-06-27 at 23 37 07@2x


Let me know if this should be a new issue or if you can fix here. seems urgent!

joaoviana:
closing as the work done on [#[migrate card config](https://github.com/lightdash/lightdash/pull/6045)](https://github.com/lightdash/lightdash/pull/6045) has been reverted

PR #6067, opened 2023-06-28
fix: revert of mantine config panels

IrakliJani:
- reverts reverted mantine migration changes
- reverts back to popover2 (from mantine to blueprint popover) for table config and chart config

skip revert and only review this commit: https://github.com/lightdash/lightdash/pull/6067/commits/26e91a45c8adac8660a0573100c404b7c8dbdacb


joaoviana (approved):
looks good! 
Tested
- chart config panel
- table config panel
- export as -> uses mantine and all good!

Issue #5977, opened 2023-06-20
When I hover over the table name in a grouped dropdown, I should see its description


PriPatel:
https://github.com/lightdash/lightdash/issues/5154

<img width="419" alt="Screenshot 2023-06-30 at 10 39 36" src="https://github.com/lightdash/lightdash/assets/7611706/9994a6a9-1983-44bf-b79d-25ed446ff82e">

Once we have implemented the grouped dropdowns, the user should be able to see the table description when they hover over the table name ( just incase you are seeing the table concept for the first time as a business user). 

- [ ] In this case, I should hover over the word 'Orders' and be able to see the table description
- [ ] This should affect dropdowns when Im adding filters in the explore and the dashboard

joaoviana:
`item.description` is too verbose.

Revert: https://github.com/lightdash/lightdash/commit/61dfbee24ba8cffa97055702184e5d8fc04992e1

Add `table.description` `Tooltip` on hover of the **table** name - as described on ticket ^

PR #6077, opened 2023-06-28
feat: display item description popover on renderFilterItem

joaoviana:
Closes: #5977

### Description:

Add `Tooltip` of description on hover of fields in `FieldAutocomplete`
ForwardRef to `MenuItem2`: https://mantine.dev/core/tooltip/#custom-components-with-tooltip

Screenrecording: 

https://github.com/lightdash/lightdash/assets/7611706/35c29ef8-043a-449f-8c29-79bccd301b3b

### DIFF
--- a/packages/frontend/src/components/common/Filters/renderFilterItem.tsx
+++ b/packages/frontend/src/components/common/Filters/renderFilterItem.tsx
@@ -4,10 +4,41 @@ import {
     Field,
     getItemId,
     getItemLabelWithoutTableName,
+    isField,
     TableCalculation,
 } from '@lightdash/common';
+import { Tooltip } from '@mantine/core';
+import { forwardRef, ReactNode } from 'react';
 import FieldIcon from './FieldIcon';
 import FieldLabel from './FieldLabel';
+
+const FilterItem = forwardRef<
+    HTMLDivElement,
+    {
+        item: Field | TableCalculation;
+        active: boolean;
+        disabled: boolean;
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/Filters/renderFilterItem.tsx
```diff
@@ -17,17 +48,20 @@ export const renderFilterItem: ItemRenderer<Field | TableCalculation> = (
...
+        <Tooltip
+            withinPortal
+            label={isField(item) ? item.description : ''}
+            disabled={!isField(item) || (isField(item) && !item.description)}
+        >
+            <FilterItem
+                item={item}
+                active={modifiers.active}
+                disabled={modifiers.disabled}
+                text={<FieldLabel item={item} />}
```
owlas, line 60: Why is the `text` set here but the `icon` is set inside `FilterItem` ? I haven't worked with these components so maybe it's obvious to frontend devs but I was surprised

owlas (approved):
The description appears in popovers 👍

packages/frontend/src/components/common/Filters/renderFilterItem.tsx
joaoviana: oh, good q, I set the `text` here because it's different between `renderFilterItem` and `renderFilterItemWithoutTableName`, but the `icon` is the same, so let  `FilterItem` set it there. 
This is not reused anywhere else so did it this way 😅

PR #6090, opened 2023-06-29
fix: dont override custom metric when creating a chart from another c…

ZeRego:
…hart

### Description:

Bug where a custom metric is removed when creating a chart from an existing chart

Replication steps:
- create chart with custom metric
- open saved chart 
- click "explore from here"
- save a new chart
- open initial chart
- see that custom metric is not there anymore

https://github.com/lightdash/lightdash/assets/9117144/2c72fb3f-fd1e-40ce-9f4a-dccd8c8cd8c1

Fix:

https://github.com/lightdash/lightdash/assets/9117144/c4178736-7bc4-4c30-8a0e-46da7bee1f35

### DIFF
--- a/packages/backend/src/models/SavedChartModel.ts
+++ b/packages/backend/src/models/SavedChartModel.ts
@@ -25,7 +25,6 @@ import {
     CreateDbSavedChartVersionField,
     CreateDbSavedChartVersionSort,
     DBFilteredAdditionalMetrics,
-    DbSavedChartAdditionalMetric,
     DbSavedChartAdditionalMetricInsert,
     DbSavedChartTableCalculationInsert,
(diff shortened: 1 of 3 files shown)


packages/backend/src/models/SavedChartModel.ts
```diff
@@ -95,8 +94,6 @@ const createSavedChartVersionAdditionalMetrics = async (
 ) => {
     const results = await trx(SavedChartAdditionalMetricTableName)
         .insert(data)
-        .onConflict('uuid')
-        .merge()
```
joaoviana, line 99: oops 😓

joaoviana (approved):

Issue #6098, opened 2023-06-30
Support the `--vars` option of dbt in the CLI


yu-iskw:
### Description
We would like to add the `vars` option to the Lightdash CLI commands as `lightdash generate`.

### What problem does this solve?
We use dbt variables so that we can dynamically change models based on the environment as the development and the production environments. To do so, we have to pass the `--vars` option to a dbt command.

```
$ dbt build --help
...
  --vars YAML                     Supply variables to the project. This
                                  argument overrides variables defined in your
                                  dbt_project.yml file. This argument should
                                  be a YAML string, eg. '{my_variable:
                                  my_value}'
```

PR #6099, opened 2023-06-30
feat: support `--vars` option in the CLI

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/6098

### Description:
The `--vars` option is supported in some sub commands of Lightdash CLI. However, `CompileHandlerOptions` doesn't have the `vars` field.


rephus:
Looks good, testing on render

rephus:
There is an error on the e2e test : 

image

rephus (changes requested):
See error

yu-iskw:
@rephus Sure. I will look into it.

packages/cli/src/handlers/dbt/run.ts
```diff
@@ -3,17 +3,13 @@ import execa from 'execa';
...
+import { DbtCompileOptions } from './compile';
 
-type DbtRunHandlerOptions = {
-    profilesDir: string;
-    projectDir: string;
-    target: string | undefined;
-    profile: string | undefined;
-    select: string[] | undefined;
-    models: string[] | undefined;
+type DbtRunHandlerOptions = DbtCompileOptions & {
```
yu-iskw, line 8: Some type aliases like `GenerateHandlerOptions` take advantage of `CompileHandlerOptions`. But, I feel `DbtCompileOptions` would look more natural in this case, as `DbtRunHandlerOptions` isn't exposed as command line options.

rephus (approved):

PR #6109, opened 2023-06-30
feat: add description tooltip to table names when adding filters

joaoviana:
Closes: #5977

### Description:

- Removes show tooltip on hover of field (in pr #6077) - so changes to `renderFilterItem` are what's reverted
- Instead, adds tooltip on hover of table name (if there's a `description`)
- Changed a util name and added another one to get the `name` and `tableLabel` for a `field`.

screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/d6bcd774-428a-45c6-8603-09be8029097a

### DIFF
--- a/packages/common/src/utils/item.ts
+++ b/packages/common/src/utils/item.ts
@@ -48,6 +48,9 @@ export const getItemLabel = (item: Field | TableCalculation) =>
     isField(item) ? `${item.tableLabel} ${item.label}` : item.displayName;
 
 export const getItemTableName = (item: Field | TableCalculation) =>
+    isField(item) ? `${item.table}` : item.displayName;
+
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/item.ts
```diff
@@ -48,6 +48,9 @@ export const getItemLabel = (item: Field | TableCalculation) =>
     isField(item) ? `${item.tableLabel} ${item.label}` : item.displayName;
 
 export const getItemTableName = (item: Field | TableCalculation) =>
+    isField(item) ? `${item.table}` : item.displayName;
```
joaoviana, line 51: what should be here if the `item` is not `Field` ? 🤔

PriPatel (approved):

ZeRego (approved):

PR #6120, opened 2023-07-03
chore: run cypress tests in parallel

ZeRego:
### Description:

Objective: make e2e tests faster

Changes:
- run e2e tests in parallel
- move timezone tests to the same github action
- run timezone tests in parallel
- delete old action to trigger e2e tests manually

Before:
- e2e tests: ~20 min
- timezone tests: ~13 min

After:
- e2e tests: ~10 min
- timezone tests: ~6 min

Note that we now have a check per parallel job which makes the checklist longer than we would want.
There is an open discussion to group/hide jobs. https://github.com/orgs/community/discussions/26246


rephus:
Globalsearch test is failing because it is typing jaffle twice... perhaps we don't clear the search input before adding new stuff 

image

.github/workflows/e2e-tests.yml
```diff
@@ -17,11 +16,21 @@ jobs:
...
-          echo "::set-output name=url::$DEPLOYMENT_URL"
+        id: deployment-url
+        run: echo "url=https://lightdash-pr-${{steps.regex.outputs.first_match}}.onrender.com" >> "$GITHUB_OUTPUT"
+
+  app-tests:
+    runs-on: ubuntu-latest
+    needs: prepare-preview
+    container:
+      image: cypress/browsers:node18.12.0-chrome107
+      options: --user 1001
```
rephus, line 27: I have no idea what any of this means :upside_down_face:

rephus (approved):

.github/workflows/e2e-tests.yml
```diff
@@ -67,7 +78,57 @@ jobs:
...
         if: failure()
         with:
-          name: cypress-screenshots-and-videos
+          name: cypress-screenshots-and-videos-{{ strategy.job-index }}
+          path: |
+            packages/e2e/cypress/screenshots
+            packages/e2e/cypress/videos
+          if-no-files-found: ignore
+
+  timezone-tests:
```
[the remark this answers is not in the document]
ZeRego: yes, jobs run in parallel and we can reuse the same `prepare-preview` logic.

.github/workflows/e2e-tests.yml
ZeRego: recommended container for cypress and option is needed to avoid some permission errors. We already use this in the current github actions.

Issue #6104, opened 2023-06-30
Allow for relative `./target` path in getManifestPath for dbt 1.5


joaoviana:
### Description

Some users are encountering an error while trying to access the target file, particularly while running `lightdash deploy` from a directory that is not the dbt project root. The error message is: ENOENT: no such file or directory, open 'target/manifest.json'.

This seems to be linked to the path `./target` for the manifest file which is not correctly locating the manifest.json file when the command is invoked outside the dbt project directory.

Reference file: https://github.com/lightdash/lightdash/blame/main/packages/cli/src/dbt/manifest.ts#L19-L21

### To Reproduce

1. Navigate to a directory outside the dbt 1.5 project root.
2. Run `lightdash deploy`.

### Slack

https://lightdash-community.slack.com/archives/C03MG2VTR08/p1688075039229679

```
https://github.com/lightdash/lightdash/blame/main/packages/cli/src/dbt/manifest.ts#L19-L21 I'm getting an error accessing the target file: ENOENT: no such file or directory, open 'target/manifest.json', and I believe it is because the path is wrong. the working directory where I'm invoking lightdash deploy from is not my dbt project directory, so ./target won't work.
```

rephus:
see https://github.com/dbt-labs/dbt-core/issues/7465 - it looks like the behavior you are describing was a regression that was fixed and target is relative to the project-directory as before now

vietsqsp:
@rephus Is there a way to get the old behaviour described in https://github.com/dbt-labs/dbt-core/issues/7465? I prefer it to the `target` folder is relative to the dbt project.

Right now, I can specify a custom target path:
```
dbt run --target-path "$(pwd)/target/" --full-refresh --models +$model+
```
However, I'll have to clear the files in that "local" `target` folder every time I run `dbt run`. Otherwise, I'll get this error:
```
dbt found two models with the name "MY_MODEL_NAME"
Since these resources have the same name, dbt will be unable to find the correct resource
  when looking for ref("MY_MODEL_NAME").
  
  To fix this, change the name of one of these resources:
  - model.* (models/marts/<PATH>/target/run/<PATH>/models/<PATH>/MY_MODEL_NAME.sql)
  - model.* (models/<PATH>/MY_MODEL_NAME.sql)
```

PR #6132, opened 2023-07-04
fix: target dbt 1.5.2

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/6104

   There was a bug between dbt>=1.5.0 and <1.5.2 where the manifest was not being generated in the dir when using project-dir 
   https://github.com/dbt-labs/dbt-core/releases/tag/v1.5.2
   https://github.com/dbt-labs/dbt-core/issues/7819

This is affecting both the CLi and backend, where a user can use 1.5.0 and get the bug, or use 1.5.2 where the bug is no more

On cloud, we install on docker ~1.5.0 which always gets the last minor version available (1.5.2 right now) 

image

### Description:

Screenshot from 2023-07-04 09-33-45

image

Backend with 1.5.0

Screenshot from 2023-07-04 09-38-36

Backend with 1.5.2

image

### DIFF
--- a/.github/workflows/cli-test-1.5.yml
+++ b/.github/workflows/cli-test-1.5.yml
@@ -34,7 +34,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.7.x'
-      - run: pip install dbt-postgres==1.5
+      - run: pip install dbt-postgres~=1.5.0
 
       ## DBT run
       - name: Run DBT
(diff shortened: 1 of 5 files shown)


ZeRego (approved):

ZeRego (approved):

ZeRego:
@rephus We need to change the cli test action to install dbt 1.5.* instead of dbt 1.5

ZeRego (approved):

PR #6136, opened 2023-07-04
chore: add mantine import counter

IrakliJani:
### Description:
adds total number of imports from mantine


ZeRego (approved):
The github action is not triggered in this PR since it doesn't have changes in the frontend package. 

Ran the script locally:
Screenshot 2023-07-04 at 13 26 46

Issue #5994, opened 2023-06-21
Ui bug - The roles column in the User Management tab looks broken


PriPatel:
<img width="924" alt="Screenshot 2023-06-21 at 10 07 47" src="https://github.com/lightdash/lightdash/assets/12776522/94538a4b-bb92-45d4-a8ab-08c434742fef">

This is only in the demo - cant reproduce in our own instance

https://demo.lightdash.com/generalSettings/userManagement

PriPatel:
This could be to do with permissions? Rather than hiding just the info in the cell, we are hiding the entire cell.

PR #6147, opened 2023-07-05
fix: fix user management permission table

rephus:
Closes: #5994

### Description:

Before: 
image

As editor: 

Screenshot from 2023-07-05 08-57-21

As admin: 
Screenshot from 2023-07-05 08-57-36


packages/frontend/src/components/UserSettings/UserManagementPanel/index.tsx
```diff
@@ -242,6 +243,10 @@ const UserManagementPanel: FC = () => {
     const [showInviteModal, setShowInviteModal] = useState(false);
     const { data: organizationUsers, isLoading } = useOrganizationUsers();
 
+    if (user.data?.ability.cannot('view', 'OrganizationMemberProfile')) {
```
rephus, line 246: I added this, this is the same condition we have to "show" the button on the sidebar. However, I think this is always true (for now) , even for viewers.

ZeRego (approved):

PR #6157, opened 2023-07-05
chore: dont update user email in e2e tests

ZeRego:
Changing the email had an impact on other e2e tests that run in parallel.

### DIFF
--- a/packages/e2e/cypress/e2e/settings/profile.cy.ts
+++ b/packages/e2e/cypress/e2e/settings/profile.cy.ts
@@ -18,13 +18,12 @@ describe('Settings - Profile', () => {
         });
     });
 
-    it('Should change name and email', () => {
+    it('should update user names', () => {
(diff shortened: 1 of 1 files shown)


ZeRego:
> looks good, we can also rename and revert to original? 🤔
> 
> just curious: does changing an email require re-verification?

@IrakliJani we revert to the original values in the `after` block

```
    after(() => {
        // Reset to default values
        cy.request({
            url: 'api/v1/user/me',
            method: 'PATCH',
            body: {
                firstName: SEED_ORG_1_ADMIN.first_name,
                lastName: SEED_ORG_1_ADMIN.last_name,
                email: SEED_ORG_1_ADMIN_EMAIL.email,
            },
        });
    });
```

When a user changes their email we will show a button for them to verify the email. This doesn't happen if they don't have an email server configured.

Issue #6038, opened 2023-06-24
Admins can't see private spaces in the "All spaces" page


hamzahc1:
### Description

Admins have access to private spaces. At the moment they can only see the space by accessing via the url.

Admins should be able to see all the private spaces in the spaces page.

### App Version

0.640.0

PriPatel:
<img width="810" alt="Screenshot 2023-06-26 at 12 07 43" src="https://github.com/lightdash/lightdash/assets/12776522/d74d9143-04d7-46a3-93d3-a3ab3a45e52a">

With the addition of all these new spaces, the Admin's `All Spaces` pages will be crowded and hard to manage. My suggestion to get the page into a 'list' view and have some new management options. (manage between 'all spaces' that I can see in admin view vs spaces that have been specifically shared with me)

PR #6161, opened 2023-07-05
feat: allow admins to list all private spaces

ZeRego:
Closes: https://github.com/lightdash/lightdash/issues/6038

### Description:

Changes:
- stop using `spaces-and-content` endpoint since it queries all dashboards and charts
- update chart summary endpoint to have all the necessary information
- allow users that can manage projects to see all private spaces

https://github.com/lightdash/lightdash/assets/9117144/4b659fb0-7fa1-4767-acfa-9be915989885

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -1641,7 +1641,7 @@ const models: TsoaRoute.Models = {
         },
     },
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
-    'Pick_Space.organizationUuid-or-projectUuid-or-uuid-or-name-or-isPrivate_':
+    'Pick_Space.organizationUuid-or-projectUuid-or-uuid-or-name-or-isPrivate-or-pinnedListUuid-or-pinnedListOrder_':
         {
             dataType: 'refAlias',
             type: {
@@ -1651,6 +1651,22 @@ const models: TsoaRoute.Models = {
                     organizationUuid: { dataType: 'string', required: true },
                     uuid: { dataType: 'string', required: true },
                     projectUuid: { dataType: 'string', required: true },
+                    pinnedListUuid: {
+                        dataType: 'union',
+                        subSchemas: [
+                            { dataType: 'string' },
+                            { dataType: 'enum', enums: [null] },
+                        ],
+                        required: true,
+                    },
+                    pinnedListOrder: {
+                        dataType: 'union',
+                        subSchemas: [
+                            { dataType: 'double' },
+                            { dataType: 'enum', enums: [null] },
+                        ],
+                        required: true,
+                    },
                     isPrivate: { dataType: 'boolean', required: true },
                 },
                 validators: {},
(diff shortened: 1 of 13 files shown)


ZeRego:
> I can't check the PR preview yet but just checking this only appears for admins right?
> 
> > allow users that can manage projects to see all private spaces
> > This comment ^^ wasn't clear that's the case

yes, for organization admins and project admins.

hamzahc1:
You get the "loading spaces" spinner again when you click on "Include all private spaces" is that expected?

ZeRego:
> You get the "loading spaces" spinner again when you click on "Include all private spaces" is that expected?

Is not ideal but out of scope since most pages behave like this and the results component doesn't have a loading state.
This issue will be addressed in this [milestone](https://github.com/lightdash/lightdash/milestone/96) where the results component would allow to search/filter/sort and only avoid an entire page spinner.

rephus:
I thought for this issue we were only going to update this line https://github.com/lightdash/lightdash/blob/main/packages/backend/src/services/SpaceService/SpaceService.ts#L73 and then on a separate ticket, make the distinction about private spaces. but it seems you did it already, what a pro

packages/frontend/src/hooks/useSpaces.ts
```diff
@@ -32,21 +32,25 @@ export const useSpaces = (
...
 
 export const useSpaceSummaries = (
     projectUuid: string,
+    includePrivateSpaces: boolean = false,
     queryOptions?: UseQueryOptions<SpaceSummary[], ApiError>,
 ) => {
     return useQuery<SpaceSummary[], ApiError>(
-        ['projects', projectUuid, 'spaces'],
-        () => getSpaceSummaries(projectUuid),
+        ['projects', projectUuid, 'spaces', includePrivateSpaces],
```
rephus, line 52: This means that everytime you toggle you will make a new request to the backend 
[Screencast from 06-07-23 08:14:47.webm](https://github.com/lightdash/lightdash/assets/1983672/27335c6e-a5b0-413b-a5fc-c1389169533b)

Can't we change this so a flag is returned in the response and filtered in the frontend ?

packages/frontend/src/pages/Spaces.tsx
```diff
@@ -97,6 +102,24 @@ const Spaces: FC = () => {
                         )}
                         headerProps={{
                             title: 'Spaces',
+                            action: (
+                                <Can
```
rephus, line 106: nice permission  check

rephus (commented):
This works well, and the code looks good, but I'm wondering if we could make return a `isPrivate` flag on the space to do the filtering on the frontend instead of making a backend request every time you toggle .

On the other hand, this makes the initial/default non-private request more efficient, as we don't return all the data ... :thinking: 

perhaps we create a different hook so we can cache both request in the frontend ? 

I guess it is not a big deal, I'm happy to merge this as it is

ZeRego:
@rephus @hamzahc1 @PriPatel 
Changed code to avoid loading spinners and use tabs. The price of this change is possible performance issues since we load all the spaces on the page load even if not necessary. We have the same issue in the other pages where we load all dashboards and charts depict only having one tab visible. 

Spaces page for admins:

https://github.com/lightdash/lightdash/assets/9117144/2e8f920f-8cc8-4d5e-95da-060e4e460d22

Spaces page for other roles:
Screenshot 2023-07-06 at 09 58 05

Dropdowns that list spaces are not affected for admins:
Screenshot 2023-07-06 at 09 56 50

packages/backend/src/generated/routes.ts
```diff
@@ -1932,6 +1932,7 @@ const models: TsoaRoute.Models = {
             'greaterThan',
             'greaterThanOrEqual',
             'inThePast',
+            'notInThePast',
```
rephus, line 1935: this is from another PR

rephus (approved):

Issue #6070, opened 2023-06-28
For each row in the`all jobs` in scheduled deliveries, there should be only one row per  job with the most recent status


TuringLovesDeathMetal:
### Description

This is currently what the `all jobs` looks like for a single row in the `run history` tab of the scheduled deliveries dashboard. 

<img width="937" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/fe5ebe51-50b1-4c80-bec6-18328ab99f96">

Instead of having one row per job + status, it should be one row per job, with the most recent status. 

so, in the example above, only items with the green ticks would show. 

Also, it looks like the times and the job names are misaligned!

### Questions:

- why are there 2x `send slack notification` jobs? Is it because there are 2 slack channels we're delivering the notification to?

hamzahc1:
If we want this to be a timeline, let's just make it a horizontal timeline (like if you're tracking a parcel from dispatched --> delivered).

PR #6167, opened 2023-07-06
feat: remove duplicated job scheduler

rephus:
Closes: #6070

### Description:

Before:

Screenshot from 2023-07-06 11-33-02

After:

Screenshot from 2023-07-06 11-33-19

About the missaligned issue: 

This is happening because we use collapsible groups inside a cell , this means we can't align the job name with the timestamps. 

However, I was not able to replicate this locally, (see screenshots) 

A solution would be to incrase the job name size to avoid text on  2 lines , or use ellipsis... but we can do this on a separate PR.

### DIFF
--- a/packages/frontend/src/components/SchedulersView/LogsView.tsx
+++ b/packages/frontend/src/components/SchedulersView/LogsView.tsx
@@ -1,4 +1,4 @@
-import { SchedulerWithLogs } from '@lightdash/common';
+import { SchedulerLog, SchedulerWithLogs } from '@lightdash/common';
 import {
     ActionIcon,
     Anchor,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/SchedulersView/LogsView.tsx
```diff
@@ -70,11 +70,21 @@ const Logs: FC<LogsProps> = ({
...
-        [logs],
-    );
-
+    const groupedLogs = useMemo(() => {
+        // Get only last log per job
+        // Logs should already by sorted by status and time
+        const uniqueLogs = logs.reduce<SchedulerLog[]>((acc, log) => {
+            if (
+                acc.some(
+                    (l) => l.jobGroup === log.jobGroup && l.task === log.task,
```
ZeRego, line 79: should this be filtered in the SQL query? So we only get the latest task state

rephus: I thought about that first, but I was afraid this was used somewhere else... but it seems `logs` are only used here, so I think we can do it

rephus: mmm, I'm not doing it in the sql, but I'm doing it in the model, after parsing and sorting, what do you think ?

ZeRego (approved):

PR #6168, opened 2023-07-06
chore: is render available

rephus:
Closes: https://github.com/lightdash/lightdash-internal-work/issues/778

### Description:

When render is not available: 

Screenshot from 2023-07-06 13-51-55

Screenshot from 2023-07-06 13-52-09

When render is available
image


rephus:
## To fix 
This code kind of does what I want, however there is a bit of conflict with other render deployments...

- [ ] If another render build is done (dpeloyment event) , then this test is flagged as skipped, we don't keep the previous result.
I want to keep the previous result if `!contains(github.event.deployment.environment, '- lightdash PR ')`

- [ ] If `lightdash PR` does not exist, this will not fail. 
this was one of the cases I wanted to cover, this happens when the branch is not a branch from main. 

One way to fix it could be to store the result of the previous "failed" run and return the same result if environment  != lightdash PR

rephus:
Ok, this works now as expected, it is always failing until Lightdas PR is success, assuming that's the deployment_status triggered: 
Screenshot from 2023-07-07 12-22-56



Screenshot from 2023-07-07 12-22-05

.github/workflows/e2e-tests.yml
```diff
@@ -3,11 +3,16 @@ on: [ deployment_status ]
...
   prepare-preview:
     # Only trigger for correct environment and status
-    if: ${{ github.event.deployment_status.state == 'success' && (contains(github.event.deployment.environment, '- lightdash PR ') || contains(github.event.deployment.environment, '- lightdash-')) }}
     runs-on: ubuntu-latest
     outputs:
       url: ${{ steps.deployment-url.outputs.url }}
     steps:
+      - name: is render ready
+        # This works assuming "Lightdash PR" is always the last deployment_status to be "success"
+        if: ${{ github.event.deployment_status.state != 'success' || (!contains(github.event.deployment.environment, '- lightdash PR ') || !contains(github.event.deployment.environment, '- lightdash-'))}}
```
ZeRego, line 12: there is a problem with this logic, has the action is failing with this message:
```Render environment chore-is-render-available-check - lightdash PR #6168 is not ready: success```

the environment and state seem correct

ZeRego (changes requested):

.github/workflows/e2e-tests.yml
rephus: Yep, I think the last OR (||) was not correctly wrapped

ZeRego (approved):

rephus:
Fixed, 
image

Issue #6105, opened 2023-06-30
Viewer user shouldn't see buttons that edit data


ZeRego:
### To Reproduce

Viewer can see the edit and options buttons. They should be hidden.

![Captura de ecrã 2023-06-30, às 10 34 41](https://github.com/lightdash/lightdash/assets/9117144/05f16ecf-5f68-4bc1-a27f-aa5bd84251cc)
![Captura de ecrã 2023-06-30, às 10 34 51](https://github.com/lightdash/lightdash/assets/9117144/02ce7152-d2a3-4f93-846f-f5b55a624f75)
![Captura de ecrã 2023-06-30, às 10 35 01](https://github.com/lightdash/lightdash/assets/9117144/a1d9ddb8-3ee8-4d90-ac45-5ed8eebf179e)

![Captura de ecrã 2023-06-30, às 10 36 06](https://github.com/lightdash/lightdash/assets/9117144/1c59ed42-f6fd-4401-bfdf-51cdcdc4a8cb)

![Captura de ecrã 2023-06-30, às 10 31 11](https://github.com/lightdash/lightdash/assets/9117144/2acf7943-be2d-4a48-a989-7474626f87bc)
### App Version

0.640.0

rephus:
LT;DR : An `editor` on a project can also get edit access on a project he's only supposed to `view` . The problem is likely that we don’t pass the project uuid to the check

This should only be affecting some buttons on the frontend. 

## How to replicate: 

- as an admin, create a Project A and a project B 
- Invite a new user, with `view` permissions on the org
- Give this user `edit` permissions to project A
- :fire: Now this user will also have `edit` buttons on project B

## How we should fix this

 Instead of 
 
 ```
  <Can
                        I={'update'}
                        this={subject('Organization', {
                            organizationUuid: data?.organizationUuid,
                        })}
                    >
                    ```
                    
                    
                    We need to do add the `                    projectUuid` to the subject like this: 
                    
                     ```
  <Can
                        I={'update'}
                        this={subject('Organization', {
                            organizationUuid: data?.organizationUuid,
                            projectUuid
                        })}
                    >
                    ```

PR #6174, opened 2023-07-06
fix: hide edit buttons from users without edit

magnew:
Closes: #6105 

### Description:

Users with edit abilities on one project were seeing the edit buttons on every project (including ones where they didn't have access). This adds the project and organization uuids to the access checks so the edit interactions show/hide is scoped to the project.

I think I got them all, but if someone knows of one I missed, let me know. 

Question for reviewers: is there a reason to **ever** not include `organizationUuid` and `projectUuid` in these checks?

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -436,7 +436,7 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
                     user.data?.ability?.can(
                         'manage',
                         subject('Explore', {
-                            organizationUuid: user.data?.organizationUuid,
+                            organizationUuid: savedChartUuid,
                             projectUuid,
                         }),
                     ) && (
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -433,7 +433,13 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
                 isLoading={isLoading || isLoadingExplore}
                 extraMenuItems={
                     savedChartUuid !== null &&
-                    user.data?.ability?.can('manage', 'Explore') && (
+                    user.data?.ability?.can(
+                        'manage',
+                        subject('Explore', {
+                            organizationUuid: user.data?.organizationUuid,
+                            projectUuid,
+                        }),
+                    ) && (
                         <>
                             {user.data?.ability?.can(
                                 'manage',
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -433,13 +433,7 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
                 isLoading={isLoading || isLoadingExplore}
                 extraMenuItems={
                     savedChartUuid !== null &&
-                    user.data?.ability?.can(
-                        'manage',
-                        subject('Explore', {
-                            organizationUuid: savedChartUuid,
-                            projectUuid,
-                        }),
-                    ) && (
+                    user.data?.ability?.can('manage', 'Explore') && (
                         <>
                             {user.data?.ability?.can(
                                 'manage',
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -236,7 +236,13 @@ const SavedChartsHeader: FC = () => {
                         </>
                     )}
                 </PageTitleAndDetailsContainer>
-                {user.data?.ability?.can('manage', 'SavedChart') && (
+                {user.data?.ability?.can(
+                    'manage',
+                    subject('SavedChart', {
+                        organizationUuid: user.data?.organizationUuid,
+                        projectUuid,
+                    }),
+                ) && (
                     <PageActionsContainer>
                         {!isEditMode ? (
                             <>
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -239,7 +239,7 @@ const SavedChartsHeader: FC = () => {
                 {user.data?.ability?.can(
                     'manage',
                     subject('SavedChart', {
-                        organizationUuid: user.data?.organizationUuid,
+                        organizationUuid: savedChart?.organizationUuid,
                         projectUuid,
                     }),
                 ) && (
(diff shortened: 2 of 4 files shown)


rephus:
> is there a reason to ever not include organizationUuid and projectUuid in these checks?

No, we should always include this info. 

Not checking the org, means that if a user is an admin in **any** org, it could modify something in another org. :fire: same with projects (like in this case)

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -433,7 +433,13 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
                 isLoading={isLoading || isLoadingExplore}
                 extraMenuItems={
                     savedChartUuid !== null &&
-                    user.data?.ability?.can('manage', 'Explore') && (
+                    user.data?.ability?.can(
+                        'manage',
+                        subject('Explore', {
+                            organizationUuid: user.data?.organizationUuid,
```
rephus, line 439: mm, ok, so this check organizationUuid is basically checking nothing

This is checking that user.orgUuid == user.orgUuid . 
What we should check here is if `organizationUuid: savedChart.orgUuid `

I can elaborate more if you don't see the difference

if you can't get the info or it is not available, I'd rather remove this from the subject

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
```diff
@@ -236,7 +236,13 @@ const SavedChartsHeader: FC = () => {
                         </>
                     )}
                 </PageTitleAndDetailsContainer>
-                {user.data?.ability?.can('manage', 'SavedChart') && (
+                {user.data?.ability?.can(
+                    'manage',
+                    subject('SavedChart', {
+                        organizationUuid: user.data?.organizationUuid,
```
rephus, line 242: same

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
```diff
@@ -97,7 +98,10 @@ const DashboardHeader = ({
     const { user } = useApp();
     const userCanManageDashboard = user.data?.ability.can(
         'manage',
-        'Dashboard',
+        subject('Dashboard', {
+            organizationUuid: user.data?.organizationUuid,
```
rephus, line 102: same

packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
```diff
@@ -57,17 +57,41 @@ const ResourceViewActionMenu: FC<ResourceViewActionMenuProps> = ({
 
     switch (item.type) {
         case ResourceViewItemType.CHART:
-            if (user.data?.ability?.cannot('manage', 'SavedChart')) {
+            if (
+                user.data?.ability?.cannot(
+                    'manage',
+                    subject('SavedChart', {
+                        organizationUuid: organizationUuid,
```
rephus, line 64: same

rephus (changes requested):
See comments

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
rephus: this is the ability definition https://github.com/lightdash/lightdash/blob/main/packages/common/src/authorization/organizationMemberAbility.ts#L62

rephus: This is a good "frontend" example about how to do this https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/hooks/validation/useValidation.ts#L121

rephus (approved):
ok, let's merge your PR as it is, and let's open a new ticket to also fix the existing org subjects

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
magnew: Done

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
magnew: done

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
magnew: done

packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
magnew: On this one the Chart and the dashboard objects dont have the orgUuid, so I'm just going to leave them out. We could add the orgUuid to those if we want to check it too. I can do that in this or the next one if you want.

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -433,7 +433,13 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
                 isLoading={isLoading || isLoadingExplore}
                 extraMenuItems={
                     savedChartUuid !== null &&
-                    user.data?.ability?.can('manage', 'Explore') && (
+                    user.data?.ability?.can(
+                        'manage',
+                        subject('Explore', {
+                            organizationUuid: savedChartUuid,
```
rephus, line 439: this is wrong , is comparing organizationUuid and  savedChartUuid

rephus (changes requested):

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
```diff
@@ -97,7 +98,10 @@ const DashboardHeader = ({
     const { user } = useApp();
     const userCanManageDashboard = user.data?.ability.can(
         'manage',
-        'Dashboard',
+        subject('Dashboard', {
+            organizationUuid: dashboardUuid,
```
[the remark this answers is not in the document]
rephus: image
this is causing that the header is empty ,even for admins

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
magnew: oops. My bad.

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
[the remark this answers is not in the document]
magnew: I see. I thought I checked that but I missed it this time around will fix.

magnew:
Unfortunately we found some more issues with the FE permissions while working on this. So the fix here makes things better, but doesn't fix all of the issues. There are at least 3 issues with front end permissions (maybe more)

1. We aren’t checking subject permissions in some places, so some users will see edit buttons on elements they don’t have permissions to change. For example, in the `ResourceActionMenu`we check permissions on charts and dashboards with just `if (user.data?.ability?.cannot('manage', 'SavedChart'))`which has no `subject`call. To find these issues look for ability checks with no subject. They are all suspect, although there might be an acceptable case I don’t know about. 
2. Some objects don’t have `organizationUuid`. For example, `ResourceViewItem`for charts and dashboards. Without that, we can’t add the `subject`calls from issue # 1 or org admins who don’t have explicit permissions on projects won’t see the edit controls. To fix this I suggested adding `organizationUuid` to all `ResourceViewItem` and also looking for other ability calls that don’t include it
3. In a number of places we are making the wrong checks using `can`. For example, we have a lot of checks that send the userUuid like this: `organizationUuid: user.organizationUuid,` which is essentially a no-op. The id sent here should be the objects in question’s organizationUuid and it gets checked against the user’s.

rephus (approved):

PR #6187, opened 2023-07-06
feat: pie charts value formatting options

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6007

### Description:

⚠️ ⚠️ ⚠️  TESTING VIDEO WITH UI QUESTION: https://cln.sh/r1zJRgK4

- [x] If you turn on the value labels, there is the option to have the value labels show as value or % of total or both
- [x] you only see these options if you have selected to show value labels
- [x] If you select **both** then the two values appear like `xx % - yy` where `yy` is the value and `xx` is the `%` of total

items left:
- [x] sync tooltips with value label config
- [x] fix echarts line bug (when switching between display option between outside<>inside)
- [x] format values based on metric formatting options

CleanShot 2023-07-07 at 20 08 13@2x

CleanShot 2023-07-07 at 20 07 08@2x

### DIFF
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -67,9 +67,11 @@ export type PieChart = {
     metricId?: string;
     isDonut?: boolean;
     valueLabel?: PieChartValueLabel;
-    showLegend?: boolean;
+    showValue?: boolean;
+    showPercentage?: boolean;
     groupLabelOverrides?: Record<string, string>;
     groupColorOverrides?: Record<string, string>;
+    showLegend?: boolean;
 };
 
 export type PieChartConfig = {
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -62,7 +62,7 @@ export const PieChartValueLabels = {
 
 export type PieChartValueLabel = keyof typeof PieChartValueLabels;
 
-export type PieChartLabelOptions = {
+export type PieChartValueOptions = {
     valueLabel: PieChartValueLabel;
     showValue: boolean;
     showPercentage: boolean;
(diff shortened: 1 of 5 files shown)


TuringLovesDeathMetal:
Niiiice! looks great 😍 

For the question about showing default vs. overrides - you mentioned in charts how we say "mixed" - this feels like a better UX than the toggles right now...

The reason being, I can "override" the "override" with the drop-down menu in the chart config (e.g. if I overrode it in one chart, but then I Was like, actually, I want _**everything**_ to be hidden, then I can just select `hidden` as the option)

How weird would it be for it to be a toggle, _**until**_ you override something, then it turns into a dropdown? That would be weird right...

magnew (approved):
LGTM, although it's possible for now someone else should look too since there might be things to look for I don't know about. But the code change makes sense. And it all looks good, nice work! 

I did see one minor issue which may not be considered part of this review, but if you place the labels inside and hover over series, the label lines still show:
Screenshot 2023-07-10 at 14 15 22

That's fairly minor and not a blocker. I can file it separately if you want.

rephus (approved):
Code looks good

IrakliJani:
> For the question about showing default vs. overrides - you mentioned in charts how we say "mixed" - this feels like a better UX than the toggles right now...
> 
> The reason being, I can "override" the "override" with the drop-down menu in the chart config (e.g. if I overrode it in one chart, but then I Was like, actually, I want _**everything**_ to be hidden, then I can just select `hidden` as the option)
> 
> How weird would it be for it to be a toggle, _**until**_ you override something, then it turns into a dropdown? That would be weird right...

@TuringLovesDeathMetal 

what if I change from "Toggle" to a checkbox with indeterminate state? this is already an adopted UX. and for the select dropdown we show "Mixed" like we have in other chart configs?
CleanShot 2023-07-10 at 17 46 43@2x

TuringLovesDeathMetal:
Yes? Honestly, I can't picture it because I've never seen an indeterminate state...how hard is it to try it out and see what it looks like? 

Or would it be easier to ask Priyanka for thoughts on this...(just don't want to ping her if it's not necessary)

TuringLovesDeathMetal:
@IrakliJani let's do the indeterminate check boxes - and then let's see if it's confusing

IrakliJani:
@TuringLovesDeathMetal 

https://github.com/lightdash/lightdash/assets/962095/8f102d76-c562-46a5-aa3d-7120a4f604a1

TuringLovesDeathMetal:
@IrakliJani Nice! I think it looks great :) good call with the indeterminate checkboxes!

Issue #6150, opened 2023-07-05
Update `edit tile content` action to use new grouped dropdown component


TuringLovesDeathMetal:
### Description

In edit mode, on a dashboard, you have the option to `edit tile content` which lets you replace the chart in a tile with another chart. 

This is still using the old dropdown component. 

Can we update this to look like the same component as when you're selecting charts to add to your dashboard?

![image](https://github.com/lightdash/lightdash/assets/31848148/c3d81a9b-4b8f-4604-97e7-02b48633368c)
![image](https://github.com/lightdash/lightdash/assets/31848148/89857527-03bb-4098-9969-9f712045671e)

### Acceptance criteria

- [ ] `edit tile content` dropdown uses the same component as the `add charts to dashboard` grouped dropdown component.

jayanand05:
Working on it.

PR #6190, opened 2023-07-06
feat: Update edit tile content action to use new grouped dropdown component

jayanand05:
Closes: #6150

### Description:
Update edit tile content action to use new grouped dropdown component

Before:
215D32C1-2BF2-47AB-AA45-99178FF5F4A4_1_201_a
After:
061B3FEB-0A7B-4E7D-A794-DE9B15BEAFA4_1_201_a


rephus:
Creating render environment for testing

rephus:
Is this the expected behaviour @PriPatel @IrakliJani  ? 

When I add an item, it is removed from the list (not disabled, like the others) 

[Screencast from 07-07-23 10:14:33.webm](https://github.com/lightdash/lightdash/assets/1983672/816e28f9-d788-41cb-bfdb-543b1de8bb00)

PriPatel:
@rephus It is expected behaviour yes. (not ideally, but we can update them all later!) but yeah when you're in the process of adding, I believe they should disappear. Next time you open up the modal, it will appear to be disabled tick☑️

rephus (approved):
Cool, the new modal works as expected:

rephus (commented):
@PriPatel sorry 1 more question. This is how it looks now: 

image

Shoudl we update the rest of the popup so it looks like the the add chart ?

PriPatel:
@rephus Yes that would be ideal, yes!!

rephus (changes requested):
hi @jayanand05 could you apply change this modal to use the same components we use when adding a chart ?

jayanand05:
hi @rephus, I think it's already using the same dropdown component as Add saved chart

jayanand05:
hi @rephus, can you explain what component is it dropdown component or something else?

rephus:
@jayanand05 we should use the Mantine Modal component like in here https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx#L112

jayanand05:
@rephus, i tried to update it with mantine model but the file TileUpdateModal is giving error since both are connected i tried to shift all the code to one file(chartTileForm.tsx) but then what can we do about TileUpdateModal file ?

yasmin-e:
Hi @jayanand05, sorry this task turned out to be more complex than expected. 
As you saw, `TileUpdateModal` is not reusable right now so I'm working on making it so and I will update this afterwards :) 
This task is part of a Milestone that I'm currently working on and needs quite a bit of context.
I suggest picking up task with the tag `Good first issue` to make sure this doesn't happen!
Let me know if you need help with anything!

PR #6199, opened 2023-07-07
chore: cache yarn in github actions

ZeRego:
Changes:
- cache yarn dependencies
- use `--frozen-lockfile` option 

This should reduce the time of most of github actions since we install all our dependencies multiple times.

Note: 
Still getting a few `cache not found` and I'm not sure why. Would still merge it and keep checking it during the next week.


IrakliJani (approved):
looks promising! 👏

rephus (approved):

Issue #6203, opened 2023-07-07
Resource Info icon shouldn't wrap when names are too long


magnew:
### Description

When resource names are long, the new info icon wraps to a new line. It shouldn't

### To Reproduce

1. Change a chart name to be too long for the resource list
2. Mouse over
3. Info icon is on its own line

### App Version
v0.658.0

PR #6204, opened 2023-07-07
fix: stop info icon from wrapping on resource list

magnew:
Closes: #6203 

### Description:

Stops the info icon from wrapping when resource names are too long


ZeRego (approved):
lgtm :) 


https://github.com/lightdash/lightdash/assets/9117144/b29f64fd-0634-4636-a31f-129865ea5b19

PR #6208, opened 2023-07-07
chore: refactor "create project" e2e tests

ZeRego:
Changes:
- refactor part of the test that was testing results
  - replaced UI tests with API tests
- improve selectors

Results when creating postgres project locally:

Before - 1min
Screenshot 2023-07-07 at 13 58 34

After - 30sec
Screenshot 2023-07-07 at 15 13 42

Results when creating projects in github action:

Before - 5min 28sec
Screenshot 2023-07-07 at 13 57 59

After - 3min 48sec
Screenshot 2023-07-07 at 15 52 14

### DIFF
--- a/packages/e2e/cypress/e2e/createProjects.cy.ts
+++ b/packages/e2e/cypress/e2e/createProjects.cy.ts
@@ -1,3 +1,5 @@
+import { ResultRow } from '@lightdash/common';
+
 const warehouseConfig = {
     postgresSQL: {
         name: 'Jaffle PostgreSQL test',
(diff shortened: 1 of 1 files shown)


rephus (approved):
Looks good, we could also consider creating these projects on the CLI , perhaps this way is faster than using the UI. 

testQuery and testRunQuery already have API tests, right ?

packages/e2e/cypress/e2e/createProjects.cy.ts
```diff
@@ -55,85 +57,99 @@ const warehouseConfig = {
...
-    cy.get('[name="dbt.type"]').select('dbt local server');
-    cy.get('[name="dbt.target"]').type('test');
-    cy.get('[name="warehouse.schema"]').type(config.schema);
+    cy.get('select[name="dbt.type"]').select('dbt local server');
+    cy.get('input[name="dbt.target"]').type('test');
+    cy.get('input[name="warehouse.schema"]').type(config.schema);
 };
 
-const testCompile = () => {
+const testCompile = (): Cypress.Chainable<string> => {
```
rephus, line 152: What's the advantage of using this chainable ?

ZeRego:
> Looks good, we could also consider creating these projects on the CLI , perhaps this way is faster than using the UI.
> testQuery and testRunQuery already have API tests, right ?

We want to test that part in the UI since it is a core functionality. 
I removed some parts and refactored other parts to be via the API since we already have tests for the explore page.

packages/e2e/cypress/e2e/createProjects.cy.ts
ZeRego: So we can chain other actions and pass the project uuid 

```
        testCompile().then((projectUuid) => {
            testFilterStringEscaping(projectUuid);
            testTimeIntervalsResults(projectUuid);
            testPercentile(projectUuid);
        });
```

Note that cypress actions are not Promises so I couldn't simply return the string.

Issue #6082, opened 2023-06-28
The position of Mantine modals on the page should match the position of the BP modals


PriPatel:
<img width="899" alt="Screenshot 2023-06-28 at 16 47 53" src="https://github.com/lightdash/lightdash/assets/12776522/3cb7ac95-7884-47c2-8f34-45fea1f03ce2">

The Mantine modals feel a bit off being right in the middle of the screen ,the position-top should match that of the BP modals. This is so things don't feel jumpy around the page

PR #6211, opened 2023-07-07
style: move mantine modals up to match blueprint's

magnew:
Closes: #6082 

### Description:

Defaults the Mantine modals' y-position to match the Blueprint ones. Removes centering from the Mantine modals.


rephus (approved):
LGTM, I don't know if there is a better way to do this, perhaps @IrakliJani know ?

IrakliJani:
looks good @magnew @rephus 🙌

PR #6212, opened 2023-07-07
chore: skip e2e action on main branch

ZeRego:
Avoid check error in `main` commits
Screenshot 2023-07-07 at 16 24 41


rephus (approved):
thanks for fixing !

PR #6213, opened 2023-07-07
docs: add branch prefix explanation

magnew:
### Description:

Adds more information to the contributing doc about PR tags. I could have used this info 😄 . 

Note that I did not open a ticket for this since it is very small and docs-only. But I can if we think that's better.


rephus (approved):
Yup, that's correct

IrakliJani (approved):

Issue #6188, opened 2023-07-06
When I click on a dashboard link from the chart information icon, the dashboard should open in a new tab


PriPatel:
<img width="423" alt="Screenshot 2023-07-06 at 17 48 41" src="https://github.com/lightdash/lightdash/assets/12776522/5e1d2693-389a-43de-b2f8-069615a1cbe2">

- [ ] When I click on the dashboard name, it should open in a new tab, rather than on the same one

PR #6214, opened 2023-07-07
fix: make dashboard link to open in new tab for saved charts

muhsinkamil:
Closes: #6188 

### Description:

- Make link to open in new tab

### DIFF
--- a/packages/frontend/src/components/common/ResourceInfoPopup/DashboardList.tsx
+++ b/packages/frontend/src/components/common/ResourceInfoPopup/DashboardList.tsx
@@ -1,6 +1,5 @@
 import { Anchor, List, Text } from '@mantine/core';
 import { FC } from 'react';
-import { Link } from 'react-router-dom';
 import { useDashboardsContainingChart } from '../../../hooks/dashboard/useDashboards';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


rephus:
I'll duplicate this branch on Lightdash to test this on render, @magnew could you review and test the code ?

packages/frontend/src/components/common/ResourceInfoPopup/DashboardList.tsx
```diff
@@ -25,8 +24,8 @@ export const DashboardList: FC<Props> = ({ resourceItemId, projectUuid }) => {
                     {relatedDashboards.map(({ uuid, name }) => (
                         <List.Item key={uuid}>
                             <Anchor
-                                component={Link}
-                                to={`/projects/${projectUuid}/dashboards/${uuid}/view`}
+                                href={`${window.location.origin}/projects/${projectUuid}/dashboards/${uuid}/view/`}
```
IrakliJani, line 27: changing from Link (react-router-dom) removes SPA navigation functionality (refreshes the page) my understanding is that `<Anchor component={Link}` supports opening to a new tab. maybe this bug has something to do something with `e.preventDefault()`?

muhsinkamil: Yes, it boots up / refreshes the app again for a new tab. Its not above `e.preventDefault()`, I tested by removing just that and it doesn't open in new tab.,
And as far as I know, `Link` from react-router is only for internal routing, here we are trying to open in a new tab, hence going out. 
Correct me if I'm wrong please

IrakliJani: ok! current changes look good then!

IrakliJani (approved):

packages/frontend/src/components/common/ResourceInfoPopup/DashboardList.tsx
yasmin-e: Thank you for your contribution @muhsinkamil! ⚡ 🙌🏼

Issue #2708, opened 2022-07-14
Rename dashboard tile titles


owlas:
### Description

Related to https://github.com/lightdash/lightdash/issues/1848 - I want to be able to rename a chart tile title. For example I might give it a long unique name for searching "my bar chart that is very important and should be looked at every day" but in the context of a dashboard I might have it say "All customers".

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVM5_vIl0=/?share_link_id=339879375259

- [ ] I'm able to rename a chart title using the three-dot-menu on a chart when I'm in `edit` mode on a dashboard. There is an option which is `Edit title`
- [ ] Rename the `edit tile content` option to say `replace chart in tile` and update the icon to use a "replace" type icon
- [ ] If I rename a chart in a dashboard, when I search for the chart in search, it will appear in search using the original chart title.

PriPatel:
Really relevant. Now that we allow users to add multiple copies of the same chart on their dashboards, it is getting hard for them to manage (especially in the tiles tab of filters) 
<img width="521" alt="Screenshot 2023-06-28 at 12 29 09" src="https://github.com/lightdash/lightdash/assets/12776522/9893aa61-e8c1-4d15-8b62-bba8cc299849">
Suggestion:

 When you're 'editing tile content', add a field here to change the title. Default should be the chart name.

PR #6220, opened 2023-07-10
remove centered

rephus:
Closes:  #2708

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVM5_vIl0=/?share_link_id=339879375259

- [ ] I'm able to rename a chart title using the three-dot-menu on a chart when I'm in `edit` mode on a dashboard. There is an option which is `Edit title`
- [ ] Rename the `edit tile content` option to say `replace chart in tile` and update the icon to use a "replace" type icon
- [ ] If I rename a chart in a dashboard, when I search for the chart in search, it will appear in search using the original chart title.

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileForms/TileUpdateChartTitle.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileForms/TileUpdateChartTitle.tsx
@@ -1,0 +1,52 @@
+import { Button, Modal, Stack, TextInput } from '@mantine/core';
+import { useForm } from '@mantine/form';
+
+interface TileUpdateModalProps {
+    title: string;
+    isOpen: boolean;
+    onClose: () => void;
+    onConfirm: (title: string) => void;
+}
+
+const TileUpdateModal = ({
+    title,
+    isOpen,
+    onClose,
+    onConfirm,
+}: TileUpdateModalProps) => {
+    const form = useForm({
+        initialValues: {
+            title,
+        },
+    });
+
+    const handleOnSubmit = form.onSubmit(({ title: newTitle }) => {
+        onConfirm(newTitle);
+        onClose();
+    });
+    return (
+        <Modal
+            size="lg"
+            opened={isOpen}
+            className="non-draggable"
+            onClose={onClose}
+            title="Edit chart title"
+        >
+            <form onSubmit={handleOnSubmit}>
+                <Stack spacing="md">
+                    <TextInput
+                        label="Chart title"
+                        placeholder={title}
+                        required
+                        {...form.getInputProps('title')}
+                    />
+                    <Button type="submit" ml="auto">
+                        Update title
+                    </Button>
+                </Stack>
+            </form>
+        </Modal>
+    );
+};
+
+export default TileUpdateModal;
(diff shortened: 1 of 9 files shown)


PriPatel (commented):
Screenshot 2023-07-10 at 14 51 50
This works well! I was able to add multiple versions of the same chart onto the dashboard, rename one of them, open my dashboard filters tab, and add filters to the right one. (which solves the original problem!)


Screenshot 2023-07-10 at 14 41 38

- One issue (which you are probably already aware of ) is the modal position (because of the recent changes made in another branch perhaps)

packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
```diff
@@ -95,7 +95,7 @@ const AddChartTilesModal: FC<Props> = ({ onAddTiles, onClose }) => {
                 return {
                     uuid: uuid4(),
                     properties: {
-                        title: savedChart?.name || null,
+                        title: savedChart?.name,
```
ZeRego, line 98: Do we want to save the chart title as the tile tile when the tile is created? 
If I change the chart title it won't be reflated in the dashboard. 

I thought we were going to have that input optional as show the chart name as a placeholder. so they now it defaults to the chart name. @PriPatel

PriPatel: @ZeRego 
If I don't edit the tile title:
- The default tile title should be the saved chart's name
- If I edit the saved chart's name , it should update the tile's name name on the dashboard.

If I edit the tile title:
- If I edit the tile title, any changes to the saved chart name should not affect the tile.

packages/frontend/src/components/DashboardTiles/TileForms/TileUpdateChartTitle.tsx
```diff
@@ -0,0 +1,52 @@
...
+            size="lg"
+            opened={isOpen}
+            className="non-draggable"
+            onClose={onClose}
+            title="Edit chart title"
+        >
+            <form onSubmit={handleOnSubmit}>
+                <Stack spacing="md">
+                    <TextInput
+                        label="Chart title"
```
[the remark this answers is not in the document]
PriPatel: @ZeRego YES good spot, I missed this. It should definitely be 'tile'

rephus:
Updated 

[Screencast from 11-07-23 08:50:26.webm](https://github.com/lightdash/lightdash/assets/1983672/f9905df6-350e-487d-940b-1064d1941f60)

packages/frontend/src/components/DashboardTiles/TileForms/TileUpdateChartTitle.tsx
```diff
@@ -0,0 +1,52 @@
...
+        onConfirm(newTitle);
+        onClose();
+    });
+    return (
+        <Modal
+            size="lg"
+            opened={isOpen}
+            className="non-draggable"
+            onClose={onClose}
+            title="Edit chart title"
```
yasmin-e, line 33: you can add `centered` prop to Mantime modals to center them vertically

rephus: This is not needed anymore   https://github.com/lightdash/lightdash/commit/5eab5b10827666304a432f5475e9b4d373becd08

ZeRego (approved):

Issue #6209, opened 2023-07-07
Formatting a currency using an invalid currency format throws an error 


rephus:
### Description
![image](https://github.com/lightdash/lightdash/assets/1983672/be6b8985-3989-4f09-8f94-ca059f7def01)

this happens when the dimension/metric has an incorrect format value in their yml 

```
format: number
```

See thread: https://lightdash.slack.com/archives/C04T5L766EM/p1688735262491309

PR #6224, opened 2023-07-10
refactor: move manifest validation to common package

ZeRego:
Related to: #6209

### Description:

Changes:
- move all logic on dbt manifest validation to common package
- remove duplicate code and files from CLI and Backend packages
 
Kept the scope of this refactor small even though there is more code that should be moved to common.


packages/common/src/dbt/validation.ts
```diff
@@ -0,0 +1,158 @@
...
+                    `Missing dbt manifest version "${manifestVersion}" in validation.`,
+                ),
+            );
+    }
+
+    const ajv = new Ajv({
+        schemas: [
+            getDbtManifestWithLightdashMetadata(dbtManifestSchema.$id),
+            lightdashMetadataSchema,
+            dbtManifestSchema,
```
ZeRego, line 96: Our validation now depends on 3 schemas:
- **dbtManifestSchema**: is the original schema from dbt. The schema file depends on what version of dbt it was used to generate the manifest.
- **lightdashMetadataSchema**: contains the definitions for the Lightdash meta properties
- **getDbtManifestWithLightdashMetadata()**: combines the original dbt schema with Lightdash definitions

packages/common/src/dbt/validation.ts
```diff
@@ -0,0 +1,158 @@
...
+import { ParseError, UnexpectedServerError } from '../types/errors';
+import assertUnreachable from '../utils/assertUnreachable';
+import lightdashMetadataSchema from './schemas/lightdashMetadata.json';
+import dbtManifestSchemaV8 from './schemas/manifestV8.json';
+import dbtManifestSchemaV9 from './schemas/manifestV9.json';
+
+const LightdashMetadataSchemaRef =
+    'https://schemas.lightdash.com/dbt/manifest.json';
+
+const getDbtManifestWithLightdashMetadata = (dbtSchemaRef: string) => ({
```
rephus, line 19: this is ok for  v8 and v9 have the same definitions, but adding support (again) for v7 , or a potentially a different future version, might not look great. What if we create a separate "static" version instead ?

rephus (commented):
Good refactoring, this is a lot better. I just have a question about adding v7 again

packages/common/src/dbt/validation.ts
ZeRego: yes, that makes sense 👍 
Will change it.

Issue #5418, opened 2023-05-09
Date filters don't work with Trino warehouse


ZeRego:
### Description

Related threads:
https://lightdash-community.slack.com/archives/C03MG2VTR08/p1678734136416819
https://lightdash-community.slack.com/archives/C03MG2VTR08/p1683642670434689

```
Hi Lightdash community. Has anyone had problems using Lightdash on Trino relating to filtering timestamps? I don't believe Trino automatically converts between varchar and timestamps so filtering on the Lightdash UI results in
Cannot apply operator: timestamp(3) = varchar(10)
Here is a sample of the SQL it generates
SELECT
  DATE_TRUNC('DAY', "report_agg".datetime) AS "report_agg_datetime_day"
FROM hive.default.report_agg AS "report_agg"

WHERE ((
  ("report_agg".datetime) = ('2023-05-09')
))
GROUP BY 1
```

```
The error is caused by a clause such as this:
...
WHERE ((
  (DATE_TRUNC('DAY', "table".date_field)) = ('2023-03-13')
))
where "table".date_field is a date field (which is correctly truncated), but compared directly to a varchar / string value, which doesn't work in trino. The solution would be to wrap the value in a simple date() call, like:
... = date('2023-03-13')
I suppose when using other engines that this value gets casted automatically, so that's why it's not a problem currently.
```

Possible solution:

```
For the original issue 
[@Rodolfo Ferreira](https://lightdash-community.slack.com/team/U04MHCRTJE5)
 we just need change the rendered sql from this:

WHERE ((
  (DATE_TRUNC('DAY', "table".date_field)) = ('2023-03-13')
))
To this:

WHERE ((
  (DATE_TRUNC('DAY', "table".date_field)) = DATE('2023-03-13')
))
Is that right? If so, I’m pretty sure this is ANSI SQL compliant and will work on Snowflake, postgres, redshift, bigquery, databricks AND trino. So we can make a one liner change to fix this.
```

### App Version

0.538.0

ZeRego:
`DATE('2023-03-13')` could work for all warehouses but what would we use for timestamps ?

owlas:
@ZeRego for timestamps I don't think there is a standard that works across all. `TO_TIMESTAMP(str, format)` is supported by trino, postgres, snowflake but not bigquery. For bigquery it's `FORMAT_TIMESTAMP`. So we'd need to conditionally render the SQL based on warehouse

PR #6233, opened 2023-07-10
fix: Update renderDateFilterSql to solve the filters of the timestamp on trino adapter

vapolonio:
Closes: #5418

### Description:
This PR update the sql generated by the renderDateFilterSql function to produce a valid sql for trino adapter, to do it becomes necessary propagate the adapter type through renderFilterRuleSql function so with it the renderDateFilterSql function can be generate conditional sql based on the adaptor type.
In the filtersCompiler.test.ts has included tests with the sql produced by the new approch.

Screenshot 2023-07-10 at 19 16 19


packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -121,39 +122,50 @@ export const renderNumberFilterSql = (
 export const renderDateFilterSql = (
     dimensionSql: string,
     filter: DateFilterRule,
+    adapterType: SupportedDbtAdapter,
     dateFormatter: (date: Date) => string = formatDate,
     startOfWeek: WeekDay | null | undefined = undefined,
 ): string => {
     const filterType = filter.operator;
+    const castValue = (value: string): string => {
```
ZeRego, line 130: @vapolonio pushed a commit to simplify the changes in `renderDateFilterSql`. 
By having this `castValue` function we don't have to duplicate all the filters sql generation.
All unit tests are passing so it's generating the same SQL :)

ZeRego (approved):

Issue #6117, opened 2023-07-03
Save chart version metadata in chart table


ZeRego:
Before working on this we should tackle some tech debt:
- In the chart table, we should add the columns: 
   - `last_version_chart_kind`: ChartKind | null
   - `last_version_updated_at`: Date
   - `last_version_updated_by_user_uuid`: reference to user table | null
- Update those columns on chart update
- Remove `saved_queries_versions` joins and use new columns.

_Originally posted by @ZeRego in https://github.com/lightdash/lightdash/issues/5767#issuecomment-1617728745_

PR #6247, opened 2023-07-11
Feat save chart metadata

rephus:
Closes:  #6117

### Description:
- [x] In the chart table, we should add the columns: 
>    - `last_version_chart_kind`: ChartKind | null
>    - `last_version_updated_at`: Date
>    - `last_version_updated_by_user_uuid`: reference to user table | null
- [x] Update those columns on chart update
- [ ] Remove `saved_queries_versions` joins and use new columns.

We are not going to remove the joins at the moment, because the fields are not populated. 

We'll be doing this on a later review (follow all TODOs) 

Right now 

Screenshot from 2023-07-11 12-42-13

With empty state

Screenshot from 2023-07-11 12-36-50

After updating 1 chart: 

Screenshot from 2023-07-11 12-37-14

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -425,7 +425,7 @@ export class SpaceModel {
                 'saved_queries_versions.updated_by_user_uuid',
                 'users.user_uuid',
             )
-            .leftJoin(
+            /* .leftJoin(
                 PinnedChartTableName,
                 `${PinnedChartTableName}.saved_chart_uuid`,
                 `${SavedChartsTableName}.saved_query_uuid`,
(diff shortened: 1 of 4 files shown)


packages/backend/src/models/SpaceModel.ts
```diff
@@ -419,12 +420,17 @@ export class SpaceModel {
                 'saved_queries_versions',
                 `saved_queries.saved_query_id`,
                 `saved_queries_versions.saved_query_id`,
-            )
+            ) // TODO remove
```
ZeRego, line 423: Most models that use `saved_queries_versions` need to be updated as well:
- ResourceViewItemModel
- SpaceModel
- SavedChartModel

Let's create a new ticket for these changes and link to this ticket/PR.

ZeRego (approved):
lgtm 🙌

Issue #5963, opened 2023-06-19
add `context` to `query_executed` event


TuringLovesDeathMetal:
### Description

add a parameter called `context` which is a string indicating where the query was executed.

- `dashboard_view` - if the query was executed because the user loaded a dashboard. 
- `explore_view` - if the query was executed in the explore view
- `chart_view` - if the query was executed in the view chart mode
- `view_underlying_data` - do we actually run a query when someone does view underlying data? or do we just show the results table? 

This way we can identify where users are querying their data from

PR #6276, opened 2023-07-12
chore: add context to query executed

rephus:
Closes: #5963

### Description

add a parameter called `context` which is a string indicating where the query was executed.

- [x] `dashboard_view` - if the query was executed because the user loaded a dashboard. 
- [x] `explore_view` - if the query was executed in the explore view
- [x] `chart_view` - if the query was executed in the view chart mode
- [x] `view_underlying_data` - do we actually run a query when someone does view underlying data? or do we just show the results table? 

Bonus

- [x] Add `csv_download` 

This way we can identify where users are querying their data from 

Screenshot from 2023-07-12 10-11-58
Screenshot from 2023-07-12 10-11-01
Screenshot from 2023-07-12 10-10-33
Screenshot from 2023-07-12 10-09-31
Screenshot from 2023-07-12 10-09-00


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -873,6 +876,10 @@ export class ProjectService {
             projectUuid,
             savedChart.tableName,
             undefined,
+            filters
```
rephus, line 879: this works even when filters are empty

ZeRego (approved):

PR #6195, opened 2023-07-07
feat: echarts show tooltip only on hovered lines/bars

rephus:
Closes: #5395
### Description:

[Screencast from 07-07-23 09:15:44.webm](https://github.com/lightdash/lightdash/assets/1983672/adaf86eb-24fa-4ffc-a833-57fe769c581e)

PR #6280, opened 2023-07-12
revert: echarts show tooltip only on hovered lines/bars (#6195)

rephus:
This reverts commit 6cf9580d47a8ce8b8d41d764b88d9957fa7db920.

Closes:


rephus:
This PR is not running e2e tests, shoudl not be affecting those tests anyway

PR #6281, opened 2023-07-12
feat: update edit tile content modal

yasmin-e:
Closes: #6150 
Closes: #6287

### Description:
a bug was created for the title problem: https://github.com/lightdash/lightdash/issues/6285

https://github.com/lightdash/lightdash/assets/67699259/14565970-a416-4b4d-8270-1c383e8a6519


packages/frontend/src/components/DashboardTiles/TileForms/TileUpdateChartTitle.tsx
```diff
@@ -28,15 +31,20 @@ const TileUpdateModal = ({
...
+                <Flex align="center" gap="xs">
+                    <MantineIcon icon={IconPencil} size="lg" color="blue.8" />
+                    <Title order={4}>Edit tile title</Title>
+                </Flex>
+            }
         >
             <form onSubmit={handleOnSubmit}>
                 <Stack spacing="md">
                     <TextInput
+                        required
```
ZeRego, line 47: Title should not be required since we want it to default to the chart title.

PriPatel (changes requested):
@yasmin-e its all looking good

- [ ] I think the action should read `Replace chart` rather than add the 'in tile' bit? We have too many 'tiles' and 'titles' going on in that dropdown. I assume this is a quick change?


Screenshot 2023-07-12 at 15 30 16

- [ ] The CTA should be 'Replace' and not 'Add'

ZeRego (approved):

Issue #6179, opened 2023-07-06
Test & compile error with wrong dbt version


ZeRego:
### Description

At the moment Lightdash tries to compile the projects with dbt 1.4 and if there is an error it tries with dbt 1.5.
But the UI only shows the error for the fallback ( dbt 1.5 )

### To Reproduce

Requisites: 
- you need a dbt project for dbt 1.4.*

Steps:
- change you yml to cause a compile error
- go to project settings page
- click Test & Compile
- see error about dbt 1.5 dependencies

We should show the compile error for dbt 1.4

### App Version

0.654.0

rephus:
I think this can happen because we install dependencies and then compile in different methods, potentially using different versions. 

For example: 

- have  a schema that works with 1.5
- have a dependency that works with 1.4

- We install dependencies using 1.4 (that works) 
- then we compile with 1.4 : it fails 
- then we compile with 1.5 . That should work, but because dependencies is using 1.4 version, that might fail.  


We should run `dependencies` and `compile` as a block, and use the same version for both.

ZeRego:
Updated ticket title/description after debugging the problem. The issue was not directly related to the project packages versions.

PR #6286, opened 2023-07-12
fix: explore dbt errors for all versions

ZeRego:
Closes: #6179 

### Description:

Screenshot 2023-07-12 at 14 20 18

### DIFF
--- a/packages/backend/src/dbt/dbtCliClient.ts
+++ b/packages/backend/src/dbt/dbtCliClient.ts
@@ -67,6 +67,15 @@ type DbtCliArgs = {
     target?: string;
 };
 
+enum DbtCommands {
+    DBT_1_4 = 'dbt',
+    DBT_1_5 = 'dbt1.5',
+}
+const DbtCommandVersion = {
+    [DbtCommands.DBT_1_4]: '1.4.*',
+    [DbtCommands.DBT_1_5]: '1.5.*',
+};
+
 export class DbtCliClient implements DbtClient {
     dbtProjectDirectory: string;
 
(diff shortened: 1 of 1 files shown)


packages/backend/src/dbt/dbtCliClient.ts
```diff
@@ -140,15 +150,49 @@ export class DbtCliClient implements DbtClient {
...
                 );
+
+                if (!(e instanceof DbtError)) throw e;
+                errorLogs[dbtExec] = e.logs || [];
                 if (dbtExecs[dbtExecs.length - 1] === dbtExec) {
-                    throw e; // Throw last error
+                    throw new DbtError(
+                        `We failed to run "dbt ${command.join(
+                            ' ',
+                        )}" successfully across all version of dbt that Lightdash currently supports. You can see specific errors below for each dbt version we tried to use.`,
```
hamzahc1, line 167: Typo - should read: "...across all version**s** of..."

hamzahc1, line 167: Replace "dbt version we tried to use." with "dbt version we tried to compile with."

hamzahc1 (changes requested):
One small typo and a copy change

packages/backend/src/dbt/dbtCliClient.ts
ZeRego: Is not always during the compile command. This error message is returned independently of the dbt command.

hamzahc1: ok then leave as is!

rephus (approved):
That look great :+1:

packages/backend/src/dbt/dbtCliClient.ts
ZeRego, line 167:
```suggestion
                        )}" successfully across all versions of dbt that Lightdash currently supports. You can see specific errors below for each dbt version we tried to use.`,
```

Issue #5914, opened 2023-06-15
Add an API to get a user by email


yu-iskw:
### Description
We would like to find an user by email. In reality, we can find a user from all users in an organization using the `ListOrganizationMembers` method. But, personally, it isn't so smart.

### What problem does this solve?
If we deal with something about an user like granting a role at project level, we need the UUID of the user. To do so, we may want to find the UUID by email.

PR #6245, opened 2023-07-11
feat: Add an API to get an organization member

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/5914

**Update @owlas**

You can now test with:

```
### Login
POST http://localhost:8080/api/v1/login
Content-Type: application/json

{
    "email": "demo@lightdash.com",
    "password": "demo_password!"
}

### Get all members
GET http://localhost:8080/api/v1/org/users

### Get a member by uuid
GET http://localhost:8080/api/v1/org/users/b264d83a-9000-426a-85ec-3f9c20f368ce

### Get a member by email address
GET http://localhost:8080/api/v1/org/users/emails/demo@lightdash.com

### Get a member that doesn't exist
GET http://localhost:8080/api/v1/org/users/emails/notexists

### Get a member that doesn't exist
GET http://localhost:8080/api/v1/org/users/notexists

### Get a member from another org
GET http://localhost:8080/api/v1/org/users/57cd4548-cbe3-42b3-aa13-97821713e307
```

### Description:
I would like to add an API to get a member's profile. Although we already have the API to get all members of an organization, it wouldn't be efficient if we want to get just one.

In my opinion, there are two use cases to get an user. First, we want to get an user by the UUID of an user. Second, we want to get an user by email.

We consider if we implement a terraform provider for Lightdash. We may want to get a user by the UUID, if we create a terraform data source. Meanwhile, we may want to get an user by the email, if we create a terraform resource. That's why the API I implemented supports either of UUID and email.

```terraform
data "lightdash_organization_member" "sample_member" {
  uuid = "xxxxx-xxxx-xxxx"
}

resource "lightdash_organization_member" "sample_member_02" {
  email = "sample-member-02@test.com"
  role = ...
}
```

PR #6288, opened 2023-07-12
feat: api to get organization members

owlas:
Closes: #5914 
Mirrors: #6245

### DIFF
--- a/packages/backend/src/controllers/organizationController.ts
+++ b/packages/backend/src/controllers/organizationController.ts
@@ -176,28 +176,6 @@ export class OrganizationController extends Controller {
     }
 
     /**
-     * Get the member profile for a user in the current user's organization by email
-     * @param req express request
-     * @param primary email of the user
-     */
-    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
-    @Get('/users/emails/{email}')
-    @OperationId('GetOrganizationMemberByEmail')
-    async getOrganizationMemberByEmail(
-        @Request() req: express.Request,
-        @Path() email: string,
-    ): Promise<ApiOrganizationMemberProfile> {
-        this.setStatus(200);
-        return {
-            status: 'ok',
-            results: await organizationService.getMemberByPrimaryEmail(
-                req.user!,
-                email,
-            ),
-        };
-    }
-
-    /**
      * Get the member profile for a user in the current user's organization by uuid
      * @param req express request
      * @param userUuid the uuid of the user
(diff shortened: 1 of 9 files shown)


packages/common/src/types/api/uuid.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Stringified UUIDv4.
+ * See [RFC 4112](https://tools.ietf.org/html/rfc4122)
+ * @pattern [0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}
+ * @format uuid
+ */
+export type UUID = string;
```
yu-iskw, line 7: I like it.

packages/backend/src/controllers/organizationController.ts
```diff
@@ -174,6 +175,50 @@ export class OrganizationController extends Controller {
...
+     * @param req express request
+     * @param primary email of the user
+     */
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @Get('/users/emails/{email}')
+    @OperationId('GetOrganizationMemberByEmail')
+    async getOrganizationMemberByEmail(
+        @Request() req: express.Request,
+        @Path() email: string,
+    ): Promise<ApiOrganizationMemberProfile> {
```
yu-iskw, line 189: I wondered if we should make it a GET method, since the path parameter of email appears in access logs and so on. Email addresses can be sensitive. It doesn't look great to make it a POST method. But, that enables us to hide email addresses in access logs and so on. What do you think?

magnew: Good question. A formal API would usually require this to be tokenized. But that's kind of what we are doing with the UUID versions of this. Do we need the the human readable version?

If so, this probably should be a POST. @owlas I don't know what we like to do here? I've used POST for this reason in the past, but we should be consistent.

magnew (approved):
LGTM pending the get/post question. I tested it a fair bit and it seems to work as advertized.

owlas:
I hadn't thought about the email, that's a great suggestion. Let's move it to a post request. I have two possible implementations:

### Search endpoint for organization members

From an API perspective this makes the most sense and is reusable:

```
POST /org/users/search
{
  email: "email@search.com"
}

[
  {
     ...user
   }
]
```

We could use this to allow searching users by different criteria. However it would always return a list. But when you search by primary email, we **know** that there can only be zero or one users

### Specific email endpoint

This is a weird syntax for the API but fits @yu-iskw 's use case better for terraform I think. If they want to uniquely 

```
POST /org/users/findByEmail
{
  "email": "my@email.com"
}

404 
OR
200
{
  ...user
}
```

magnew (approved):
👍

Issue #6285, opened 2023-07-12
If I replace a chart on my dashboard, the chart title doesn't change


rephus:
### Description

This only happens when title is undefined, perhaps the title is not updated correctly because react thins it hasn't changed. 

[Screencast from 12-07-23 14:59:43.webm](https://github.com/lightdash/lightdash/assets/1983672/a5f4dddb-afbd-4bf5-ae50-30e314be5a4a)

PR #6291, opened 2023-07-12
fix:  title change tile

rephus:
Closes:  #6285

### Description:

[Screencast from 12-07-23 16:35:24.webm](https://github.com/lightdash/lightdash/assets/1983672/7e124026-3e5b-49bf-b328-d4db2ffc01fe)

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
@@ -210,7 +210,13 @@ const TileBase = <T extends Dashboard['tiles'][number]>({
                             tile={tile}
                             onClose={() => setIsReplacingChart(false)}
                             onConfirm={(newChartTile) => {
-                                onEdit(newChartTile as T);
+                                onEdit({
+                                    ...newChartTile,
+                                    properties: {
+                                        ...newChartTile.properties,
+                                        title: undefined,
+                                    },
+                                } as T);
                                 setIsReplacingChart(false);
                             }}
                         />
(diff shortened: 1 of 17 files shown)


packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
```diff
@@ -199,7 +199,13 @@ const TileBase = <T extends Dashboard['tiles'][number]>({
                         tile={tile}
                         onClose={() => setIsReplacingChart(false)}
                         onConfirm={(data) => {
-                            onEdit(data);
+                            onEdit({
+                                ...data,
+                                properties: {
+                                    ...data.properties,
+                                    title: undefined,
```
ZeRego, line 206: `TileUpdateModal` is used for all tile types. This would remove the title from any change in the loom/markdown tile

ZeRego (changes requested):

packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
ZeRego: This fix might be easier after the changes from this PR https://github.com/lightdash/lightdash/pull/6281

yasmin-e (commented):
Looks great 👍🏼 
I just noticed something and I'm not sure this is the intended behaviour 🤔 (it's not crucial though)
Here are the steps:
1. Replace chart 
2. Edit title
3. The title placeholder is the original chart's title (not the replaced one)
Screenshot 2023-07-13 at 9 22 55

rephus:
Mmm, I can't replicate this... 

[Screencast from 13-07-23 10:20:30.webm](https://github.com/lightdash/lightdash/assets/1983672/36b1c238-427d-42b5-9092-819e694a2907)

ZeRego (approved):

Issue #6301, opened 2023-07-13
New chart on dashboard have empty title on filters 


rephus:
### Description

![image](https://github.com/lightdash/lightdash/assets/1983672/db6b49f9-d844-4439-a092-e3becf54d161)

### To Reproduce

1. go to dashboard
2. go into edit mode
3. add a chart
4. do not save yet
5. add a filter
6. select tiles
7. :fire: last added chart appears with empty title

PR #6303, opened 2023-07-13
fix: filter empty chart title

rephus:
Closes: #6301
### Description:

I think this is the best way to fix it without introducing extra attributes to the chart ChartTile, but le tme know if you prefer a different approach.

Before:

image

After:

image


yasmin-e (approved):
fine by me 👍🏼

Issue #5441, opened 2023-05-11
You can create attributes and assign users to them.


owlas:
### Description

You can add attributes and assign users to them. Only org admins can do this. These user attributes can be used to add metadata about users and can later be used to apply row/column based security based on their attributes.

Leads to many customisations in the UI/UX based on the current user

### UI

See Miro sketch: https://miro.com/app/board/uXjVM34fyak=/

- [ ] there is a new page in the `organization settings` called `user attributes`. 
- [ ] This page is only accessible to org. admins
- [ ] admins can create a new attribute by cliking on `add new attribute` button.
- [ ] When you create an attribute, you assign it a name (this must only be using [a-z_0-9] characters)
- [ ] Attributes can optionally have a description.
- [ ] If a user tries to name an attribute with anything but [a-z_0-9], they see an error message, and they are not able to save the attribute. (backend restriction with optional frontend validation)
- [ ] you're able to assign users to attributes and give them values for the attribute. 
- [ ] The attributes page lists all of the attributes and tells you:
    - [ ] the attribute description when you hover over the "info" icon
    - [ ] the number of users who have the attribute assigned to them in text below the entry
    - [ ] the option to edit + delete the attribute. 

### API

Each member has many attributes. Each attribute has:

- `name` - must be unique per member (so can't have two attributes with the same name). They can only use [a-z_]
- `value` - the value of the attribute. can be any string

Lets use UUID

```
### Show all attributes in the org (and users inheriting it?)
GET /org/attributes

### Create a specific attribute on the org (name or description)
POST /org/attributes

### Update the whole attribute payload
PUT /org/attributes/{uuid}
```

### Database

Add a new table for `user_attributes`

- primary key is a UUID
- another index on (orgUuid, name)
- has a description

Add a new table for `organization_member_user_attributes`

- foreign key to `organization_memberships`(orgId, userId)
- foreign key to `user_attributes` (orgId, attributeId)
- a value

### Out of scope

- Non-string attributes (number, date etc.) - in future we could support any dimension type
- attributes are user viewable / editable - in future admins could configure if a user can edit a particular attribute on their profile
- no groups, attributes must be directly assigned to a single user

### Analytics events

- [ ] user_attributes_page_viewed - event that's triggered when the user attribute page is viewed
    - organization_id
    - user_id
    - timestamp
- [ ] user_attribute_created - event that's triggered when a user attribute it created
    - user_id - user who created the attribute
    - organization_id
[...]

PR #6305, opened 2023-07-13
feat: add user attributes

rephus:
Closes:  #5441
### Description

You can add attributes and assign users to them. Only org admins can do this. These user attributes can be used to add metadata about users and can later be used to apply row/column based security based on their attributes.

Leads to many customisations in the UI/UX based on the current user

This is behind a feature flag: 

    localStorage.setItem('USER_ATTRIBUTES', true)

[Screencast from 17-07-23 10:44:10.webm](https://github.com/lightdash/lightdash/assets/1983672/841745ee-14eb-4ca8-8884-f98c9d3f9b4c)

### UI

See Miro sketch: https://miro.com/app/board/uXjVM34fyak=/

- [x] there is a new page in the `organization settings` called `user attributes`. 
- [x] This page is only accessible to org. admins
- [x] admins can create a new attribute by cliking on `add new attribute` button.
- [x] When you create an attribute, you assign it a name (this must only be using [a-z_0-9] characters)
- [x] Attributes can optionally have a description.
- [x] If a user tries to name an attribute with anything but [a-z_0-9], they see an error message, and they are not able to save the attribute. (backend restriction with optional frontend validation)
- [x] you're able to assign users to attributes and give them values for the attribute. 
- [x] The attributes page lists all of the attributes and tells you:
    - [x] the attribute description when you hover over the "info" icon
    - [x] the number of users who have the attribute assigned to them in text below the entry
    - [x] the option to edit + delete the attribute. 

### API

Each member has many attributes. Each attribute has:

- `name` - must be unique per member (so can't have two attributes with the same name). They can only use [a-z_]
- `value` - the value of the attribute. can be any string

Lets use UUID

```
### Show all attributes in the org (and users inheriting it?)
GET /org/attributes

### Create a specific attribute on the org (name or description)
POST /org/attributes

### Update the whole attribute payload
PUT /org/attributes/{uuid}
```

### Database

Add a new table for `user_attributes`

- primary key is a UUID
- another index on (orgUuid, name)
- has a description

Add a new table for `organization_member_user_attributes`

- foreign key to `organization_memberships`(orgId, userId)
- foreign key to `user_attributes` (orgId, attributeId)
- a value

### Out of scope

- Non-string attributes (number, date etc.) - in future we could support any dimension type
- attributes are user viewable / editable - in future admins could configure if a user can edit a particular attribute on their profile
- no groups, attributes must be directly assigned to a single user

### Analytics events
These will be done in a separate ticket


packages/common/src/types/userAttributes.ts
```diff
@@ -0,0 +1,29 @@
+export type OrgAttribute = {
```
rephus, line 1: I'm not too sure about this naming, if you have suggestion for this model and the one below, let me know.

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+    orgUserAttribute: OrgAttribute;
+    allUserAttributes: OrgAttribute[];
+}> = ({ orgUserAttribute, allUserAttributes }) => {
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    const [isEditDialogOpen, setIsEditDialogOpen] = useState(false);
+
+    const { mutate: deleteUserAttribute } = useUserAttributesDeleteMutation();
+
+    return (
+        <>
```
IrakliJani, line 44: this `<>` seems to be extra

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+            <tr>
+                <td width={800}>
```
IrakliJani, line 46: (note for self - check this in preview env)

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                                    >
+                                        <MantineIcon
+                                            icon={IconInfoCircle}
+                                            color="gray.6"
+                                        />
+                                    </Tooltip>
+                                )}
+                            </Group>
+                            <Text fz="xs" color="gray.6">
+                                {' '}
```
IrakliJani, line 66: don't think we need this space here

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+import { SettingsCard } from '../../common/Settings/SettingsCard';
+import ForbiddenPanel from '../../ForbiddenPanel';
+import UserAttributeModal from './UserAttributeModal';
+
+const UserListItem: FC<{
+    orgUserAttribute: OrgAttribute;
+    allUserAttributes: OrgAttribute[];
+}> = ({ orgUserAttribute, allUserAttributes }) => {
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    const [isEditDialogOpen, setIsEditDialogOpen] = useState(false);
```
IrakliJani, line 39: there's this hook that can be handful here https://mantine.dev/hooks/use-disclosure/

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                </td>
+                <td>
+                    <Flex justify="space-between" align="right">
+                        <Button
+                            variant="subtle"
+                            onClick={() => setIsEditDialogOpen(true)}
+                        >
+                            <MantineIcon
+                                icon={IconEdit}
+                                size="xlg"
```
IrakliJani, line 81: (note for self - check this in preview env)

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+    useUserAttributesDeleteMutation,
+} from '../../../hooks/useUserAttributes';
+import { useApp } from '../../../providers/AppProvider';
+import LoadingState from '../../common/LoadingState';
+import MantineIcon from '../../common/MantineIcon';
+import { SettingsCard } from '../../common/Settings/SettingsCard';
+import ForbiddenPanel from '../../ForbiddenPanel';
+import UserAttributeModal from './UserAttributeModal';
+
+const UserListItem: FC<{
```
IrakliJani, line 34: I believe this should be UserAttributesItem

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                        )}
+                        <Button
+                            leftIcon={<MantineIcon icon={IconCircleX} />}
+                            variant="outline"
+                            onClick={() => setIsDeleteDialogOpen(true)}
+                            color="red"
+                        >
+                            Delete
+                        </Button>
+                        <Modal
```
IrakliJani, line 101: ideally we can render model once and keep in state uuid of user attribute that is open

IrakliJani: also, this will help the readability to define this modal as a separate component or even in separate file

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                                >
+                                    {' '}
+                                    docs
+                                </a>
+                                .
+                            </div>
+                        }
+                    >
+                        {/* TODO add link to docs */}
+                        {/* TODO keep tooltip open on hover */}
```
IrakliJani, line 194: will this be addressed in the upcoming PRs?

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+                        <MantineIcon icon={IconInfoCircle} color="gray.6" />
+                    </Tooltip>
+                </Group>
+                <>
+                    <Button onClick={() => setShowAddAttributeModal(true)}>
+                        Add new attributes
+                    </Button>
+                    <UserAttributeModal
```
IrakliJani, line 203: 👍

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+    Group,
+    Modal,
+    Select,
+    Stack,
+    Text,
+    Textarea,
+    TextInput,
+    Title,
+} from '@mantine/core';
+import { useForm } from '@mantine/form';
```
IrakliJani, line 13: 👍

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                                    (_, i) => i !== index,
+                                                ),
+                                            );
+                                        }}
+                                    />
+                                </Group>
+                            );
+                        })}
+                        <Button
+                            w={200}
```
IrakliJani, line 169: (note for self - check this in preview env)

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                    ...(form.values.users || []),
+                                    { userUuid: '', email: '', value: '' },
+                                ]);
+                            }}
+                        >
+                            Add user
+                        </Button>
+                    </Stack>
+
+                    <Group spacing="xs" position="right" mt="md">
```
IrakliJani, line 182: if you have a parent stack of this set to spacing='lg' you don't really need to use `mt` anymore

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                </Text>
+                                <Text fw={500}>Value</Text>
+                            </Group>
+                        )}
+
+                        {form.values.users?.map((user, index) => {
+                            return (
+                                <Group key={index}>
+                                    <Select
+                                        w={200}
```
IrakliJani, line 125: (note for self - check this in preview env)

packages/common/src/types/userAttributes.ts
IrakliJani: seems fine to me

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                        {isEditDialogOpen && (
+                            <UserAttributeModal
+                                opened={isEditDialogOpen}
+                                userAttribute={orgUserAttribute}
+                                onClose={() => setIsEditDialogOpen(false)}
+                                allUserAttributes={allUserAttributes}
+                            />
+                        )}
+                        <Button
+                            leftIcon={<MantineIcon icon={IconCircleX} />}
```
IrakliJani, line 94: can you use the trash icon? I believe that's what we use in other pages

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
IrakliJani: 'xlg' is not necesary, you can make this variant="outline" like the second button
CleanShot 2023-07-17 at 19 44 42@2x

also, change `<Flex justify="space-between" align="right">`
to just `<Group>` to get a space between two icons

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
IrakliJani: instead of having one TD and placing both text and action items in one cell you can create a second cell with `width="1%"` << this is a "hack" to set cell with minimal width.
that way you can get rid of `width={800}` and next line of `<Flex justify="space-between" align="center">`

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
IrakliJani: CleanShot 2023-07-17 at 19 51 32@2x

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                    <TextInput
+                                        name={`users.${index}.value`}
+                                        placeholder="E.g. US"
+                                        required
+                                        {...form.getInputProps(
+                                            `users.${index}.value`,
+                                        )}
+                                    />
+                                    <Button
+                                        pr={5}
```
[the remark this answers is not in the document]
IrakliJani: I don't think this is going to be necessary after addressing the previous comment

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
IrakliJani: you can ignore aligning it perfectly to first input and just make its weight "auto" (by removing the prop)

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
```diff
@@ -0,0 +1,61 @@
...
+                .primary()
+                .defaultTo(knex.raw('uuid_generate_v4()'));
+            table.string('name').notNullable();
+            table.string('description').nullable();
+
+            table
+                .timestamp('created_at', { useTz: false })
+                .notNullable()
+                .defaultTo(knex.fn.now());
+            table
```
owlas, line 20: If we use `organization_id` here we can enforce that members are only assigned to attributes in their orgs. See below

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
```diff
@@ -0,0 +1,61 @@
...
+            organizationMemberUserAttributes,
+            (table) => {
+                table.integer('user_id').notNullable();
+                table.integer('organization_id').notNullable();
+                table
+                    .foreign(['user_id', 'organization_id'])
+                    .references(['user_id', 'organization_id'])
+                    .inTable('organization_memberships')
+                    .onDelete('CASCADE');
+                table
```
owlas, line 40: Before we discussed having a `[user_attribute_uuid, organization_id]` foreign key so we can enforce matching organization ids between users and attributes

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
```diff
@@ -0,0 +1,61 @@
...
+                    .uuid('user_attribute_uuid')
+                    .notNullable()
+                    .references('user_attribute_uuid')
+                    .inTable(userAttributesTableName)
+                    .onDelete('CASCADE');
+                table.string('value').notNullable();
+                table.primary([
```
owlas, line 47: Why is the primary key like this? Couldn't it just be user_id, user_attribute_uuid

packages/backend/src/services/UserAttributesService/UserAttributesService.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export class UserAttributesService {
+    private readonly userAttributesModel: UserAttributesModel;
+
+    constructor(dependencies: Dependencies) {
+        this.userAttributesModel = dependencies.userAttributesModel;
+    }
+
+    async getAll(user: SessionUser): Promise<OrgAttribute[]> {
```
owlas, line 21: Missing ability check

owlas (changes requested):
See inline

packages/backend/src/services/UserAttributesService/UserAttributesService.ts
```diff
@@ -0,0 +1,82 @@
...
+        orgAttributeUuid: string,
+        orgAttribute: CreateOrgAttribute,
+    ): Promise<OrgAttribute> {
+        const savedAttribute = await this.userAttributesModel.get(
+            orgAttributeUuid,
+        );
+
+        if (
+            user.ability.cannot(
+                'manage',
```
owlas, line 51: Optional, but after recent refactoring of permissions, I'm wondering if we add new subjects for things we might later want to permission differently.

For example this could be
```typescript
cannot('update', subject('UserAttribute', {organizationUuid, attribute.organizationUuid}))
```

That way we have explicit permissions throughout the app, and we only have to change them by updating the ability definitions files

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
rephus: very cool

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
rephus: mmm, group does not keep it in the same line, perhaps because of the width=1% hack ? 

image

rephus: Button variant outline (instead of subtle) makes it different :thinking: 

image

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
rephus: This property is not set to allign all the inputs, but to alling the input with the headers, otherwise  `value` is off

image

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
rephus: mm, using flexgrow does not make any difference `                                <Group key={index} sx={{flexGrow:1}}>`

image

rephus: how do I fix the orange extra space ? should I wrap both groups in a stack ? or playing with margins

packages/common/src/types/userAttributes.ts
```diff
@@ -0,0 +1,29 @@
...
+    uuid: string;
+    createdAt: Date;
+    name: string;
+    organizationUuid: string;
+    description?: string;
+    users: UserAttribute[];
```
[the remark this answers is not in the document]
rephus: I think I prefer `users` :thinking:  , you ?

packages/backend/src/services/UserAttributesService/UserAttributesService.ts
rephus: I think this is better to implement when we need it, otherwise, permissions can grow very quickly and become  very hard to maintain.

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
rephus: Updated

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                            onClick={() => setIsEditDialogOpen(true)}
+                        >
+                            <MantineIcon
+                                icon={IconEdit}
+                                size="xlg"
+                                color="gray.7"
+                            />
+                        </Button>
+                        {isEditDialogOpen && (
+                            <UserAttributeModal
```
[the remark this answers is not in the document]
rephus: If I remove this condition, then there is a weird bug where the form does not update correctly with new data... do you know how to fix this ? 



[Uploading Screencast from 18-07-23 11:06:12.webm…]()

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
IrakliJani: you should apply `sx={{flexGrow:1}}` to individual input elements

IrakliJani: I'll check further on on that orange extra space

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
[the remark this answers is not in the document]
IrakliJani: you can use https://mantine.dev/core/action-icon/

packages/frontend/src/components/UserSettings/UserAttributesPanel/index.tsx
IrakliJani: what about `<Group noWrap>`

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
IrakliJani: I think keeping User email and value labels as a separate line is not going to be consistent with the layout that im suggesting.
here's an idea - what if you set mantine input label prop only for the first row (where index===0)?
in that case label will have a proper spacing and will follow the inputs properly.

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
rephus: Thanks ! updated 

image

And the orange extra space has been automaitcally fixed by using labels !

IrakliJani:
@rephus UI changes inside modal looks great!
there's just this error when I go to user attributes in settings and does not show a list anymore.

image

IrakliJani (approved):
@rephus yep its fixed now. approving

there's one minor nit



this icon is huge, way bigger than the delete one. I was thinking maybe make a regular button with regular sized icon with "Edit" text in it and make it blue?

other than that, approving!

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
```diff
@@ -0,0 +1,58 @@
...
+                .notNullable()
+                .defaultTo(knex.fn.now());
+            table
+                .integer('organization_id')
+                .notNullable()
+                .references('organization_id')
+                .inTable('organizations')
+                .onDelete('CASCADE');
+            table.unique(['name', 'organization_id']);
+            table.unique(['user_attribute_uuid', 'organization_id']); // needed for foreign key
```
owlas, line 27: Is this just so it builds an index? You can use `.index` to make it obvious why it's there

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
owlas: awesome

packages/backend/src/services/UserAttributesService/UserAttributesService.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export class UserAttributesService {
+    private readonly userAttributesModel: UserAttributesModel;
+
+    constructor(dependencies: Dependencies) {
+        this.userAttributesModel = dependencies.userAttributesModel;
+    }
+
+    async getAll(user: SessionUser): Promise<UserAttribute[]> {
+        if (user.ability.cannot('manage', 'Organization')) {
```
owlas, line 22: Shouldn't we check the organization? 

```typescript
user.ability.cannot('manage', subject('Organization', { organizationUuid }))
```

I know each user now has one organization but if we assume that then it couples this function to that business logic but it doesn't need to be. Makes it easier when we allow many orgs per user

packages/backend/src/services/UserAttributesService/UserAttributesService.ts
```diff
@@ -0,0 +1,86 @@
...
+        return this.userAttributesModel.find({
+            organizationUuid: user.organizationUuid!,
+        });
+    }
+
+    async create(
+        user: SessionUser,
+        orgAttribute: CreateOrgAttribute,
+    ): Promise<UserAttribute> {
+        if (user.ability.cannot('manage', 'Organization')) {
```
owlas, line 35: same here

owlas (changes requested):
One last change in the ability checks

examples/api/user-attributes.http
```diff
@@ -0,0 +1,30 @@
...
+### Create user attributes
+POST http://localhost:8080/api/v1/org/attributes
+Content-Type: application/json
+
+{
+    "name":"country",
+    "description":"description",
+    "users":[{
+        "userUuid":"b264d83a-9000-426a-85ec-3f9c20f368ce",
+        "email":"",
```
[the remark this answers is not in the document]
rephus: Fixed

packages/backend/src/database/migrations/20230714124708_user_attributes.ts
rephus: No, we require an unique constraint in order to use ` table .foreign(['user_attribute_uuid', 'organization_id'])` on `organizationMemberUserAttributes` 

An index is not enough

> alter table "organization_member_user_attributes" add constraint "organization_member_user_attributes_user_attribute_uuid_organization_id_foreign" foreign key ("user_attribute_uuid", "organization_id") references "user_attributes" ("user_attribute_uuid", "organization_id") on delete CASCADE - there is no unique constraint matching given keys for referenced table "user_attributes"
error: alter table "organization_member_user_attributes" add constraint "organization_member_user_attributes_user_attribute_uuid_organization_id_foreign" foreign key ("user_attribute_uuid", "organization_id") references "user_attributes" ("user_attribute_uuid", "organization_id") on delete CASCADE - there is no unique constraint matching given keys for referenced table "user_attributes"

owlas (approved):

Issue #4097, opened 2023-01-06
Invalid alias name for nested column in BigQuery


pilis:
### Description

I've tried adding a dimension to an explore that is a nested column in BigQuery. In dbt properties file nested columns are registered using "dot" syntax in the format: `group.column` (dot in-between). When Lightdash compiles the query it's using exact dimension name to make an alias in the format: `table_dimension` (underscore in between). For nested columns in BigQuery it gives: `table_group.column` (first underscore then dot). That's invalid column alias in BigQuery according to error message: `Invalid field name "table_group.column". Fields must contain only letters, numbers, and underscores, start with a letter or underscore, and be at most 300 characters long.`

### To Reproduce

1. Create a table in BigQuery with a nested column
2. Create respective Ligthdash model in dbt
3. Try to create an explore with nested column dimension
4. You should receive an error: `Invalid field name "table_group.column". Fields must contain only letters, numbers, and underscores, start with a letter or underscore, and be at most 300 characters long`

### App Version

0.361.0

pilis:
I have traced the error down to https://github.com/lightdash/lightdash/blob/main/packages/backend/src/queryBuilder.ts#L72 where I could simply substitute a dot in `alias` variable, but I don't know what implications it will give on frontend yet since there is some variables binding implemented and I lack front-end skills

TuringLovesDeathMetal:
Hey @pilis thanks for opening this issue and for all of the great details in here!

We'll take a look at it and see what we can do 👀

ZeRego:
This means we need to change how we create the field ids. Not sure what side effects this may cause 🤔.

_packages/common/src/types/field.ts_
```
export const fieldId = (field: Pick<Field, 'table' | 'name'>): FieldId =>
    `${field.table}_${field.name}`;
```

TuringLovesDeathMetal:
@pilis for now, you'll need to flatten the table in dbt to use nested fields in Lightdash. We have a related issue here: #4102 where we talk about a potential solution, but were not going to focus on it for this quarter (we definitely want to improve how we handle nested fields in the future though!)

So, for your example column, you'd need to make it into a field in your dbt model, like:

```sql
SELECT
field.name AS field_name

FROM table
```

Then you could access the field in Lightdash in your .yml

pilis:
@TuringLovesDeathMetal Fair enough! 

This workaround is valid, but I had a hope that it would be easy enough to support nested fields and not to flatten wide tables (100s of columns). 

[...]

PR #6320, opened 2023-07-13
fix: handle dotted field names when generating field id

ZeRego:
Closes: #4097 

### Description:

**Changes:**
When generating a field ID, replace all the `.` in the field name to `__` ( double underscore )

I'm using double underscore to avoid conflicts

eg: the fields `dimension.column` and `dimension_column` in the same table would both end up as `table_dimension_column` 
 
**Note 1:** Even with a double underscore is possible to cause this conflict but less likely. 

**Note 2:** This PR doesn't handle the necessary SQL to access json fields in each warehouse. The developer will have to define it in their yml.  

YML:

Screenshot 2023-07-13 at 18 09 01

Manifest:

Screenshot 2023-07-13 at 18 15 06

Query:

Screenshot 2023-07-13 at 18 08 40

### DIFF
--- a/packages/common/src/types/field.test.ts
+++ b/packages/common/src/types/field.test.ts
@@ -1,18 +1,47 @@
-import { convertFieldRefToFieldId } from './field';
+import { convertFieldRefToFieldId, fieldId } from './field';
 
 describe('field util functions', () => {
-    it('should convert field references to field ids', async () => {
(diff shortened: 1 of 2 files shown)


yu-iskw:
@ZeRego Thank you very much for addressing the issue!

rephus (approved):
Nice

PR #6322, opened 2023-07-13
refactor: move vis config panels to shared folder

magnew:
### Description:
Moves the various vis config panels to a shared folder. This logical grouping will help us cleanly share components between them as we convert to mantine. A couple things to note:

- There are a lot of file changes here, but it's almost all moving files. The only real change is to move the FieldSelectItem to it's own file. 
- There is more to be done here as far as sharing components, but this is the smallest useful change I could make.


IrakliJani (approved):
lgtm!

Issue #6250, opened 2023-07-11
Migrate "Big number config" to mantine


IrakliJani:
### Description

configuration for "Big number charts" is quite straightforward, so it doesn't appear necessary to break it down into smaller tasks

<img width="319" alt="CleanShot 2023-07-11 at 17 05 30@2x" src="https://github.com/lightdash/lightdash/assets/962095/01486849-07b0-47d8-ac9f-1e2664ac017d">

**acceptance criteria:**

- [ ] change the popover
- [ ] change tabs
- [ ] change all the inputs
  - [ ] use field selector from pie charts (also move it to shared components)

<img width="207" alt="CleanShot 2023-07-11 at 17 05 35@2x" src="https://github.com/lightdash/lightdash/assets/962095/724875c9-cea5-46de-8401-51fa6b6293ed">

PR #6324, opened 2023-07-13
feat: big number config panel using Mantine

magnew:
Closes: #6250 

### Description:

Moves the Big Number config to use Mantine.


magnew:
I made a mistake with this. Will post again.

PR #6325, opened 2023-07-13
feat: big number config panel using mantine

magnew:
Closes: #6250 

### Description:

Migrates the big number config panel to Mantine

### DIFF
--- a/packages/frontend/src/hooks/useBigNumberConfig.ts
+++ b/packages/frontend/src/hooks/useBigNumberConfig.ts
@@ -137,6 +137,7 @@ const useBigNumberConfig = (
     const [selectedField, setSelectedField] = useState<string | undefined>();
     const getField = useCallback(
         (fieldNameOrId: string | undefined) => {
+            if (fieldNameOrId === undefined) return;
             return availableFields.find((f) => {
                 return isField(f)
                     ? fieldId(f) === fieldNameOrId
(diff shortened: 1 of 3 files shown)


IrakliJani:
render is failing for some reason, rebuilding

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -136,7 +136,7 @@ const useBigNumberConfig = (
 
     const [selectedField, setSelectedField] = useState<string | undefined>();
     const getField = useCallback(
-        (fieldNameOrId: string) => {
+        (fieldNameOrId: string | undefined) => {
```
IrakliJani, line 139: highly unlikely but wonder if this change might cause a bug.
just to be safe we add this guard in the function `if (fieldNameOrId === undefined) return`

IrakliJani (approved):
code looks great! tested on render, UI looks awesome as well! added one minor suggestion and will be good to address that before merging.

packages/frontend/src/hooks/useBigNumberConfig.ts
magnew: I'm happy to do that. But I think this will return undefined when passed undefined anyway, no?

IrakliJani: `isField` will handle properly but not sure about the second condition `.name === fieldNameOrId`

magnew: Ok, I added the guard. Honestly I like to use unit tests for things like this. I might add some.

ZeRego:
@all-contributors add @magnew for code

Issue #6133, opened 2023-07-04
There's a way to create a new chart from within a dashboard


TuringLovesDeathMetal:
### Description

There's a way to create a new chart from _**within**_ a dashboard. Adding a chart from within a dashboard will mean that it doesn't appear in the list of content items in a space and you're not able to add it to any other dashboards.

### Acceptance criteria:

- [ ]  There is a way to create a new chart from within a dashboard when you're in `edit` mode on a dashboard. **NOTE: we don't actually ship this in this ticket though! we need to ship the other features in this milestone before we expose the button. We put this button behind a feature flag, and build the rest, then expose the functionality once it's fully functional after.**
- [ ] These charts can only exist if there is a tile with the chart in it (i.e. if you delete the tile with the chart, then the chart is deleted).
- [ ] There is a clear distinction between creating a new chart vs. adding a new tile which is using an existing chart. Basically, there's an explore view that exists "only inside the dashboard" when you're creating a new chart from within a dashboard.
     - [ ]  the explore view has no header, only an `x` in the top-right, then there is some banner saying `you are creating this chart from within [dashboard name]`
     - [ ] The context of all of the changes to the dashboard you are coming from are retained when you're in this explore view (so you don't need to save the dashboard before you try to add a new chart)
     - [ ] When you click `save` it doesn't actually save anything...it just passes this context back to teh dashboard view. Only when you save the dashboard will the chart actually be saved. 
     - [ ] If I add a new chart, then click `cancel` to the dashboard changes before I've hit `save`, the new chart I created would be deleted (it would never actually have been created/saved to the database).
     - [ ] E.g. I have a dashboard, I add two tiles from existing charts (so these are not actually tiles in the dashboard). then, I create a new chart from within the dashboard. When I hit save, this chart is not actually created (because this needs a tile to exist - charts created in dashboards can only exist if it exists in a tile), instead, the context is passed back to teh dashboard. Only when I hit save are all three tiles created, and that new chart is created. 
- [ ]  When you save the chart, there is information about the dashboard that you're saving it to
- [ ] When you first create the saved chart, it exits the Explore view and opens the dashboard, in edit mode, anchored to the new tile you've added.
- [ ] for now, these charts appear like any other chart in the app in terms of navigation/discoverability (we'll hide them from content lists, etc. in the next ticket)

- [ ] The only way to create a chart that lives in a dashboard is to create it from _**within**_ the dashboard (we'll have flows for adding it directly to dashboards from the `table` view later...)

[...]

PR #6331, opened 2023-07-14
feat: create chart at tile creation

yasmin-e:
Closes: **1/? of** [#6133](https://github.com/lightdash/lightdash/issues/6133) 

### Description:
Changes:
- allow a chart to be created when creating/updating a dashboard
- add unit tests
- update API docs

Screenshot 2023-07-17 at 17 55 47

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -136,40 +136,22 @@ export class DashboardModel {
                 .returning('*');
 
             switch (tile.type) {
-                case DashboardTileTypes.SAVED_CHART:
+                case DashboardTileTypes.SAVED_CHART: {
+                    let chartUuid = tile.properties.savedChartUuid;
                     if (tile.properties.newChartData) {
-                        await createSavedChart(
+                        chartUuid = await createSavedChart(
                             trx,
                             projectUuid,
                             userUuid,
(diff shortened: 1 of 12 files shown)


packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -125,6 +87,23 @@ export class DashboardService {
         };
     }
 
+    async hasDashboardSpaceAccess(
```
ZeRego, line 90: any reason to move `hasDashboardSpaceAccess` and `checkUpdateAccess` ?

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -134,6 +137,14 @@ export class DashboardModel {
 
             switch (tile.type) {
                 case DashboardTileTypes.SAVED_CHART:
+                    if (tile.properties.newChartData) {
+                        await createSavedChart(
```
ZeRego, line 141: we are creating the chart but not creating the tile with the new chart uuid

packages/backend/src/services/DashboardService/DashboardService.ts
yasmin-e: I only added the projectUuids here, the re-formatting is probably ESLint or prettier

ZeRego (approved):

Issue #6329, opened 2023-07-14
Add an API to get a project member


yu-iskw:
### Description
We already have the API to get a list of members in a project by `GetProjectAccessList`. If we have another API to get just a member in a project, that would be useful to implement something like a terraform provider.

https://docs.lightdash.com/api/v1/#tag/Projects/operation/GetProjectAccessList

```terraform
data "lightdash_project_member" "sample" {
  project_uuid = "..."
  user_uuid = "..."
}

resource "lightdash_project_member" "sample" {
  project_uuid = "..."
  user_uuid = "..."
  role = '"..."
}
```

### What problem does this solve?
The implementation for a terraform provider can get much simpler.

PR #6338, opened 2023-07-14
feat: Add an API to get a project member

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/6329

### Description:
Although we already have the API to get a list of members for a project, we would like another one to get just a member for a project.


rephus (commented):
The code looks great, well done. 

This endpoint is not going to be used in the UI, eventually we will start wondering why we need its, and we might remove it by accident. 

So I'd suggest you add a comment explaining this is used to extend our `public API` capabilities, and maybe add a description about what this is trying to achieve. 

We also need for you to add a new api e2e test that checks the response here: https://github.com/lightdash/lightdash/blob/main/packages/e2e/cypress/e2e/api/api.cy.ts

We normally add e2e permission tests but just having 1 api test should be enough.

yu-iskw:
@rephus We develop a terraform provider so that we can manage permissions of Lightdash in terraform.

- https://registry.terraform.io/providers/ubie-oss/lightdash/latest/docs/resources/project_role_member

The terraform provider offers the terraform resource of `lightdash_project_role_member` which enables us to manage the project role of an user. As we don't have an API to get an user in a project by user's UUID, we have to find one from all members of a project using the [`GetProjectAccessList`](https://docs.lightdash.com/api/v1/#tag/Projects/operation/GetProjectAccessList) API. I think it isn't efficient to get just a member. That's why I would like to implement the API in the pull request. 

- https://github.com/ubie-oss/terraform-provider-lightdash/blob/4a223a4b88aeb1646eff62ac81a785a607d1e6de/internal/lightdash/api/get_project_memberr_v1.go#L41-L49

rephus:
> @rephus We develop a terraform provider so that we can manage permissions of Lightdash in terraform.

Nice, yeah that makes sense, could you add a brief anonymous comment saying in the endpoint comment block so we know about this in the future ?

packages/backend/src/controllers/projectController.ts
```diff
@@ -119,6 +120,35 @@ export class ProjectController extends Controller {
         };
     }
 
+    /**
+     * Get a project member's access for a project.
+     *
+     * NOTE:
+     * We don't use the API on the frontend. Instead, we can call the API
+     * so that we make sure of the user's access to the project.
```
yu-iskw, line 128: @rephus I have added the comment.

rephus:
Quickly running tests

Issue #3936, opened 2022-12-07
I want to be able to include colour ranges in my table conditional formatting rules


TuringLovesDeathMetal:
### User story

I have a cohort retention table. It looks like this (but imagine it without the colours 🌚 )

![image](https://user-images.githubusercontent.com/31848148/206213892-81b1870d-9b70-4122-b994-d2de44bc8e76.png)

I want to add colours so it's easy to see at what stage users are retaining the most vs. churning the most. 

I could do this by adding a bunch of rules to my conditional formatting. But, it would be really nice if Lightdash would just "make these rules" for me automatically (basically, automatically building buckets/a scale for my colours). 

### Acceptance Criteria 

here's an example sketch: https://miro.com/app/board/uXjVPE1t3As=/?moveToWidget=3458764540678815107&cot=14

- [ ] when I create a rule, I can pick if it's a `single color` rule or a `color range` rule
- [ ] For a range of colours rule, you need to apply the following settings:
    - [ ] Select a colour that you want to have for the minimum values in your colour range
    - [ ] Select a colour that you want to have for the maximum values in your colour range.
    - [ ] Select the minimum value that you want the range to start at.
    - [ ] Select the maximum value for the range. 
- [ ] Ranges are set to 5 colours (separate issue for setting a custom number of values in the colour range here: https://github.com/lightdash/lightdash/issues/3764).
- [ ] for now, you can only set the min and max colour.

rohan-thakur:
For this feature, it would be great if the `min` and `max` values didn't have to be fixed and by default just applied to the minimum and maximum values in the result set. I see this in the Miro sketch but wanted to say that that would probably make it easiest for the end-user as a default.

IrakliJani:
> For this feature, it would be great if the `min` and `max` values didn't have to be fixed and by default just applied to the minimum and maximum values in the result set. I see this in the Miro sketch but wanted to say that that would probably make it easiest for the end-user as a default.

@rohan-thakur specced another ticket for auto min/max values 🙌  https://github.com/lightdash/lightdash/issues/6406

Issue #3764, opened 2022-11-11
I want to set a custom number of colours to be included in my colour range (table conditional formatting)


TuringLovesDeathMetal:
### Description

I want to set a custom number of colours to be included in the range of colours applied to the values in my conditional formatting. 

We apply 5 values to a range, by default. But I should be able to pick the number of values in the range (e.g. 10 would create 10 different colours in the colour range)

PR #6360, opened 2023-07-17
feat: conditional formatting color range

IrakliJani:
Closes: #3936 

### Description:

acceptance criteria:
- [x] when I create a rule, I can pick if it's a `single color` rule or a `color range` rule
- [x] For a range of colours rule, you need to apply the following settings:
    - [x] Select a colour that you want to have for the minimum values in your colour range
    - [x] Select a colour that you want to have for the maximum values in your colour range.
    - [x] Select the minimum value that you want the range to start at.
    - [x] Select the maximum value for the range. 
- [x] Ranges are set to 5 colours (separate issue for setting a custom number of values in the colour range here: https://github.com/lightdash/lightdash/issues/3764).
- [x] for now, you can only set the min and max colour.

persisted configuration:

```js
conditionalFormattings: [
  // for single color range and conditional operator
  {
    "color": "#fac858",
    "rules": [
      {
        "id": "90f8b88a-b623-4853-9a9b-9ba460e7de4b",
        "values": [
          30
        ],
        "operator": "greaterThan"
      }
    ],
    "target": {
      "fieldId": "orders_total_order_amount"
    }
  },
  // for multi color range and min/max values
  {
    "rule": {
      "max": 40,
      "min": 30
    },
    "color": {
      "end": "#73c0de",
      "start": "#b39e6f",
      "steps": 5 // <--------- hardcoding `5` for now (generates 5 colors between `start` and `end`
    },
    "target": {
      "fieldId": "orders_customer_id"
    }
  }
]

```

video demo:

https://github.com/lightdash/lightdash/assets/962095/e83a4502-1e1e-469f-9711-bc412ea4a55b

in results table:
CleanShot 2023-07-17 at 16 17 00@2x

in pivot tables:
CleanShot 2023-07-17 at 16 13 48@2x

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormatting.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormatting.tsx
@@ -41,7 +41,6 @@ import FieldIcon from '../../common/Filters/FieldIcon';
 import { fieldLabelText } from '../../common/Filters/FieldLabel';
 import { FiltersProvider } from '../../common/Filters/FiltersProvider';
 import MantineIcon from '../../common/MantineIcon';
-import ColorSelector from '../ColorSelector';
 import FieldSelectItem from '../FieldSelectItem';
 import ConditionalFormattingRule from './ConditionalFormattingRule';
 
@@ -296,13 +295,15 @@ const ConditionalFormatting: FC<ConditionalFormattingProps> = ({
                             config,
                         ) ? (
                             <>
-                                <Group spacing="xs">
-                                    <Text fw={500}>Select color</Text>
-                                </Group>
-
-                                <ColorSelector
+                                <ColorInput
+                                    withinPortal={false}
+                                    withEyeDropper={false}
+                                    format="hex"
+                                    swatches={defaultColors}
+                                    swatchesPerRow={defaultColors.length}
+                                    label="Select color"
                                     color={config.color}
-                                    onColorChange={handleChangeSingleColor}
+                                    onChange={handleChangeSingleColor}
                                 />
 
                                 {config.rules.map((rule, ruleIndex) => (
--- a/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormatting.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormatting.tsx
@@ -106,11 +106,15 @@ const ConditionalFormatting: FC<ConditionalFormattingProps> = ({
         switch (newConfigType) {
             case ConditionalFormattingConfigType.Single:
                 return handleChange(
-                    createConditionalFormattingConfigWithSingleColor(),
+                    createConditionalFormattingConfigWithSingleColor(
+                        config.target,
+                    ),
                 );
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormatting.tsx
```diff
@@ -49,82 +66,181 @@ const ConditionalFormatting: FC<ConditionalFormattingProps> = ({
...
+
+    const handleChangeRule = useCallback(
+        (
+            index: number,
+            newRule: ConditionalFormattingWithConditionalOperator,
+        ) => {
+            if (isConditionalFormattingConfigWithSingleColor(config)) {
+                handleChange(
+                    produce(config, (draft) => {
+                        // FIXME: check if we can fix this problem in number input
```
rephus, line 182: do we still need to fix this ?

rephus (commented):
Code looks good, let me test this on render

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormatting.tsx
IrakliJani: yep, unfortunately this comes from blueprint input which is in filter components (not part of this change)

TuringLovesDeathMetal:
weooooo!!! Looks awesome 😍 Super excited for this!

The functionality is great! It works as expected which is awesome :) Just a few things I noticed (but if these require a lot of work, let me know and I'll rethink if we actually need them):

- Can we add the separator text/subheading `select type` (this should look the same as `select field` and `select colour`) above the `single colour` vs. `colour range`? 
image

- Can we have colours applied by default to the range? So, white for min value, then the first colour in their default palette for the max value? 
image

- Can we have the max value set by default to the max value in the column they're applying the conditional formatting to? (if this would take more than 30 mins, then let's not do it) 
image
 
- okay, this one might be complicated...but on hover, you see the rule, and it says `greater than (min value) and less than (max value)`. I'd expect it to say what colour it is based on the "rules"/intervals we've set in the colour range. E.g. if white = values between [0, 5], I'd expect to see that when hovering over white. Not `values between [0, 100]`.
image

- I don't think this is something that was introduced with the ranges...but one thing I found weird was how my conditional formatting disappeared if I made my table into a pivot table 🤔 I'm not moving the column that has the conditional formatting rule...so it feels weird that the conditional formatting disappears once I've added a pivot. https://www.loom.com/share/f8d539c0044a4c11b4b58bc7ef844b8f

IrakliJani:
@TuringLovesDeathMetal thanks for a review!
- [x] adjusted defaults
- [x] added label "Select type" to a input

re min/max: it's not going to be a simple addition so I'm gonna spec a new ticket for it!

IrakliJani:
@TuringLovesDeathMetal 

re pivot tables: we don't apply conditional formatting to columns that are either in header or indexes. only for values. this is a "bug" in a sense that we allow conditional formatting to be added to that columns, or this can be a separete "feature request" to support cond-formatting to header/indexes in pivot tables.

p.s. specced ticket for min/max auto setting https://github.com/lightdash/lightdash/issues/6406

TuringLovesDeathMetal (approved):
Looks great!

PR #6361, opened 2023-07-17
refactor: make pie chart color picker shared

magnew:
### Description:

Move the pie chart color picker to it's own component so it is sharable.


IrakliJani (approved):
tested on render, works as expected. let's wait for tests to pass before merging

nits: maybe call it `ColorSwatchPicker`? since there already is `ColorPicker` and `ColorInput`?

magnew:
> tested on render, works as expected. let's wait for tests to pass before merging
> 
> nits: maybe call it `ColorSwatchPicker`? since there already is `ColorPicker` and `ColorInput`?

Done. Tests passing. I'm merging this.

Issue #6182, opened 2023-07-06
I can't see the number of charts in a Pinned Space when I hover over it


PriPatel:
When the `Space` is displayed on the homepage:
<img width="286" alt="Screenshot 2023-07-06 at 17 04 29" src="https://github.com/lightdash/lightdash/assets/12776522/f2ca4f52-4853-47f6-b3b6-9dbc93bb308a">

When the `Space` is displayed on the `All Spaces` page:
<img width="290" alt="Screenshot 2023-07-06 at 17 06 35" src="https://github.com/lightdash/lightdash/assets/12776522/43bcf27c-a1d0-4ad5-a2b8-caa49a56eeb0">

- [ ] The number of charts inside a Pinned Space should be displayed when I hover over it

PR #6368, opened 2023-07-17
fix: I can't see the number of charts in a Pinned Space when I hover over it 

muhsinkamil:
Closes: #6182

### Description:

- Change reference to chart_count field in query result


ZeRego (approved):
lgtm :) 

we can merge if tests pass

PR #6373, opened 2023-07-18
fix: reference chart_count in all_spaces query

ZeRego:
duplicate of `fix: I can't see the number of charts in a Pinned Space when I hover over it #6368` #6368


joaoviana:
tests passed, Render env looked fine, all good!

Issue #6349, opened 2023-07-17
Add `sql_filter` to models


owlas:
### Description

More context: #5442 

Users can declare a filter in their `yaml`, which will always be applied to any query against that model.

```yaml
models:
  - name: users
    meta:
      sql_where: ${my_dimension} = 12 OR ${TABLE}.hello = 'bye'
```

- Can reference dimensions with `${my_dimension}`
- Can reference the base table with `${TABLE}`
- `sql_where` is only applied if the model is being queried directly. It doesn't apply if another table is queried and joins `users`
- Cannot reference metrics with `${metric}` 

The generated SQL for any query should always include this

```
SELECT
  dimensions,
  metrics
GROUP BY dimensions
WHERE true
  AND ${ sql_where }
  AND (.... remaining filter logic ....)
```

The UI doesn't indicate that this is applied but it's visible in the compiled SQL.

Open questions:
- Should you be able to reference joined tables? `${my_joined_table.dimension}` - we'd need to scan for these references when we decide what tables to join.

### Analytics events

- [ ]  we should add two columns to the `project_compile` event:
  - [ ] `models_with_sql_filters_count` - the number of models that have at least one sql filter on them. (Note: models can only have a single sql_filter on them)

TuringLovesDeathMetal:
@owlas what's the reason that we went with `sql_where` over specifying that this is a `user attribute` filter? 

(i.e. the questions in the issue that you linked)

owlas:
@TuringLovesDeathMetal since #5442 included a few (not-mutually exclusive) filter options, I've started to break them out into individual issues.

I'll open a new ticket for the other syntax

Issue #2049, opened 2022-05-10
Reference metrics/dimensions from all joined models


TuringLovesDeathMetal:
### Description: What is it?

Imagine a scenario where I have three models: `organizations`, `users`, and `user_details_pii`. 

`users` is joined to `organizations`
`user_details_pii` is joined to `users`

I have a metric in `users` that requires a field from `organizations`. 

you should be able to click on that metric in `user_details_pii` and just see the results. 

(here's the example YAML file for these three models)

```yaml
version: 2
models:
- name: organizations
  columns:
  - name: organization_id
  - name: organization_type
```

```yaml
version: 2
models:
- name: users
  meta:
    joins:
    - join: organizations
      sql_on: ${users.organization_id} = ${organizations.organization_id}
  columns:
  - name: user_id
    meta:
      metrics:
        count_cloud_users:
          type: count_distinct
          sql: IF(${organizations.organization_type} = 'cloud', ${users.user_id}, NULL)
```

```yaml
version: 2
models:
- name: user_details_pii
  meta:
    joins:
    - join: users
      sql_on: ${users.user_id} = ${user_details_pii.user_id}
  columns:
  - name: user_id
```

### Problem: What problem does this solve?

Right now, if I have the setup described above, then `user_details_pii` just throws an error and doesn't appear in my models. 

<img width="1081" alt="image" src="https://user-images.githubusercontent.com/31848148/167634591-064e5968-fecd-4c8c-9743-4a378e8a66cf.png">

Sometimes, there are metrics that I want to use which require me to reference another data model. These should "just work" when I pull them across to other models via joins

### What: Roughly, what does this look like in the product?

- [ ] In the example above, if I went to `user_details_pii` and clicked on `count_cloud_users` from the joined `users` table, it would work. 
- [ ] I wouldn't have to join `organizations` to `user_details_pii`.

TuringLovesDeathMetal:
Closing this issue - not interaction in 1.5 years, so will reopen if it becomes an issue!

PR #6375, opened 2023-07-18
feat: add sql_filter to models

rephus:
Closes:  #6349

### Description:

- [x] Added validation 

image

image

- [x] Sql where on table
- [x] Compiled sql on explore compile

```
  - name: customers
    meta:
      sql_where: '(${first_name} IS NOT NULL AND ${last_name} IS NOT NULL)'
...
  - name: payments
    meta:
      sql_where: '${orders.amount} > 50'
```

image

image

## Analytics events

- [x]     whatever approach we decide to take for adding the filters, we should add this to the project_compile event. E.g. row_access_filters_count (number of row access filters....i've added teh row prefix because we'll have column_access_filters_count in another issue)

image


rephus:
I'm still missing analytic events

packages/backend/src/queryBuilder.ts
```diff
@@ -245,11 +245,23 @@ export const buildQuery = ({
...
+    );
+
     const nestedFilterSql = getNestedFilterSQLFromGroup(filters.dimensions);
+    const nestedFilterWhere = nestedFilterSql ? [nestedFilterSql] : [];
+    const allSqlFilters = [...tablesSqlWhere, ...nestedFilterWhere];
     const sqlWhere =
-        filters.dimensions !== undefined && nestedFilterSql
-            ? `WHERE ${nestedFilterSql}`
-            : '';
+        allSqlFilters.length > 0 ? `WHERE ${allSqlFilters.join(' AND')}` : '';
```
rephus, line 264: I tried to avoid doing `WHERE true` because I'm not sure if that will be compatible with all warehouses

owlas: It works! 
- https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types
- https://docs.databricks.com/sql/language-manual/data-types/boolean-type.html
- https://docs.snowflake.com/en/sql-reference/data-types-logical
- https://www.postgresql.org/docs/current/datatype-boolean.html
- https://trino.io/docs/current/language/types.html#boolean

packages/backend/src/queryBuilder.ts
```diff
@@ -245,11 +245,23 @@ export const buildQuery = ({
         return undefined;
     };
 
+    const tablesSqlWhere = [...joinedTables].reduce<string[]>(
```
owlas, line 248: let's only apply the `sqlWhere` from the main table

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -213,8 +215,18 @@ export class ExploreCompiler {
             }),
             {},
         );
+        const compiledSqlWhere = table.sqlWhere
```
owlas, line 218: looks good!

packages/backend/src/queryBuilder.ts
rephus: do you still prefer using `WHERE true`

packages/backend/src/queryBuilder.ts
rephus: Are you sure about this ? don't we want to also filter joined tables if yoou add a sqlWhere filter to it ? specially once we introduce user attribute row level access...

packages/backend/src/queryBuilder.ts
owlas: Good point, it's done now, in future we could change this if we're adding more and more filters

packages/backend/src/queryBuilder.ts
owlas: I _think_ so. It's easier to start including `sql_filter` from joins later if we change our mind than doing the opposite. So I think we should just have.

In any case, I think it would get confusing because joins are only conditionally included in the query (depending on the filters selected) and those joined tables could have their own `sql_filter` referencing other join

rephus: I imagine you can have  a schema like this 

```
models: 
  - name: users
    meta: 
      sql_filter: "${ld.user.user_id} = ${ld.user_id}"
  - companies: 
    meta: 
        join_on: 'users.company_id = companies.company_id'
```

This is configured so users can only see their own users (or any other attribute that will give them access) . However if you select from `company` while joining users, that filter will no longer apply, so any user can see all other users in the joined result.

rephus:
TODO: add sql_filter alias to dbt schema

packages/backend/src/queryBuilder.ts
rephus: Like this ? 

image

My `customers` filter is ignored
```
  - name: customers
    meta:
      sql_where: '(${first_name} IS NOT NULL AND ${last_name} IS NOT NULL)'
``

**Comment by rephus** [comment]
Render is hanging :thinking: 

```
Jul 19 04:26:04 PM  #14 108.2   Downloading SQLAlchemy-1.3.24-cp39-cp39-manylinux2010_x86_64.whl (1.3 MB)
Jul 19 04:45:37 PM  #14 1281.1 INFO: pip is looking at multiple versions of sqlalchemy to determine which version is compatible with other requirements. This could take a while.
```

packages/backend/src/queryBuilder.ts
owlas: I think right now it's confusing because each model serves two purposes:
1. as a table that can be joined to other tables (which ignores things like `join` or `hidden` or `group_label`) but includes it's dimensions metrics
2. as a base table, which joins other tables (including all the extra metadata)

So I'm thinking of `sql_filter` as something that is only applicable to when you use it as a base table.

owlas: Actually @rephus 

If I join model A to model B. and model C to model B. 

Can model A reach model C?

rephus:
> could we update this to always include joins to tables if they're referenced in the sql_where ?

This might be hard, because we'll have to extract the dimensions and fields from the sql string, does this currently work if you use a custom sql on a dimension ?

packages/backend/src/queryBuilder.ts
rephus: you're right, this #2049 can be a problem when joining multiple tables. 

In fact, even the compile doesn't like it. 

eg: I added a new model ( a copy of events)  only joined with payments 

image

I got this compile error

> - ERROR> joined : Model "payments" has a dimension reference: ${orders.amount} which matches no dimension

owlas:
> This might be hard/unstable, because we'll have to extract the dimensions and fields from the sql string, does this currently work if you use a custom sql on a dimension ?

Ok no problem, let's leave this for a separate issue

packages/backend/src/queryBuilder.ts
owlas: Yep let's only do the base table.

Instead of #2049 - we might also make the yaml clearer so it's obvious when you're configuring the underlying table or an "explore" which is a bunch of joined tables with extra config (like `sql_where` and access)

owlas:
Just testing in PR env

owlas (changes requested):
CleanShot 2023-07-20 at 11 26 54@2x

What am I doing wrong? I can't get this to work

rephus:
> CleanShot 2023-07-20 at 11 26 54@2x
> 
> What am I doing wrong? I can't get this to work

I think this is happening because you are running a CLI version that doesn't have my changes, so the compile that happens locally is not adding the right sql_filter attribute to the explore.

Issue #6227, opened 2023-07-10
Charts created from within a dashboard do not appear in any chart lists in the app.


TuringLovesDeathMetal:
### Description

- [ ] Charts that are created from within a dashboard **do not** appear any chart lists in the app. They DO NOT appear in:
    - [ ] The space `all content` list
    - [ ] The space `charts` list 
    - [ ] The `all charts` list from the nav bar
    - [ ] The `most popular` list on the homepage
    - [ ] The `recently updated` list on the homepage
    - [ ] The list of charts when you're `adding a new tile` to a dashboard
    - [ ] Global search

PR #6382, opened 2023-07-18
feat: save chart in dashboard

ZeRego:
Relates to: #6133 
Closes: #6227 

### Description:

Changes:
- amend knex table typing
- update chart table to have a dashboard uuid column and space id column be nullable

Table before:
```
create table saved_queries
(
    saved_query_id                    integer generated always as identity
        primary key,
    saved_query_uuid                  uuid         default uuid_generate_v4() not null
        constraint saved_queries_saved_query_uuid_unique
            unique,
    created_at                        timestamp    default CURRENT_TIMESTAMP  not null,
    name                              text                                    not null,
    space_id                          integer                                 not null
        constraint saved_queries_space_id_foreign
            references spaces
            on delete cascade,
    description                       text,
    last_version_chart_kind           varchar(255) default 'vertical_bar'::character varying,
    last_version_updated_at           timestamp    default CURRENT_TIMESTAMP,
    last_version_updated_by_user_uuid uuid
        constraint saved_queries_last_version_updated_by_user_uuid_foreign
            references users (user_uuid)
            on delete set null
);
```

Table now:

```
create table saved_queries
(
    saved_query_id                    integer generated always as identity
        primary key,
    saved_query_uuid                  uuid         default uuid_generate_v4() not null
        constraint saved_queries_saved_query_uuid_unique
            unique,
    created_at                        timestamp    default CURRENT_TIMESTAMP  not null,
    name                              text                                    not null,
    space_id                          integer
        constraint saved_queries_space_id_foreign
            references spaces
            on delete cascade,
    description                       text,
    last_version_chart_kind           varchar(255) default 'vertical_bar'::character varying,
    last_version_updated_at           timestamp    default CURRENT_TIMESTAMP,
    last_version_updated_by_user_uuid uuid
        constraint saved_queries_last_version_updated_by_user_uuid_foreign
            references users (user_uuid)
            on delete set null,
    dashboard_uuid                    uuid
        constraint saved_queries_dashboard_uuid_foreign
            references dashboards (dashboard_uuid)
            on delete cascade
);

```

E2e test passing:

Screenshot 2023-07-18 at 15 43 46

Preview:

https://github.com/lightdash/lightdash/assets/9117144/7f4329e9-619c-4147-89de-3fa603b728ef

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -116,7 +116,7 @@
         "migrate-production": "knex migrate:latest --knexfile dist/database/knexfile.js",
         "seed": "knex seed:run --knexfile src/database/knexfile.ts",
         "seed-production": "knex seed:run --knexfile dist/database/knexfile.js",
-        "rollback-last": "knex migrate:rollback --knexfile src/database/knexfile.ts",
+        "rollback-last": "knex migrate:down --knexfile src/database/knexfile.ts",
         "rollback-all": "knex migrate:rollback --all --knexfile src/database/knexfile.ts",
         "rollback-all-production": "knex migrate:rollback --all --knexfile dist/database/knexfile.js"
     }
--- a/packages/backend/src/database/migrations/20230718093901_add_dashbord_uuid_to_chart_table.ts
+++ b/packages/backend/src/database/migrations/20230718093901_add_dashbord_uuid_to_chart_table.ts
@@ -1,0 +1,21 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('saved_queries', (t) => {
+        t.uuid('dashboard_uuid')
+            .references('dashboard_uuid')
+            .inTable('dashboards')
+            .nullable()
+            .onDelete('CASCADE');
+        // make space id nullable
+        t.integer('space_id').nullable().alter({ alterType: false });
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('saved_queries', (t) => {
+        t.dropColumns('dashboard_uuid');
+        // make space id non nullable
+        t.integer('space_id').notNullable().alter({ alterType: false });
+    });
+}
--- a/packages/backend/src/database/migrations/20230718093901_add_dashbord_uuid_to_chart_table.ts
+++ b/packages/backend/src/database/migrations/20230718093901_add_dashbord_uuid_to_chart_table.ts
@@ -13,6 +13,7 @@ export async function up(knex: Knex): Promise<void> {
 }
 
 export async function down(knex: Knex): Promise<void> {
+    await knex('saved_queries').delete().where('space_id', null);
     await knex.schema.alterTable('saved_queries', (t) => {
         t.dropColumns('dashboard_uuid');
         // make space id non nullable
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -98,6 +98,7 @@ export class DashboardModel {
 
     private static async createVersion(
         trx: Transaction,
+        dashboardUuid: string,
         dashboardId: number,
         version: DashboardVersionedFields,
         projectUuid: string,
@@ -143,6 +144,7 @@ export class DashboardModel {
                             trx,
                             projectUuid,
                             userUuid,
+                            dashboardUuid,
                             tile.properties.newChartData,
                         );
                     }
@@ -673,6 +675,7 @@ export class DashboardModel {
 
             await DashboardModel.createVersion(
                 trx,
+                newDashboard.dashboard_uuid,
                 newDashboard.dashboard_id,
                 {
                     ...dashboard,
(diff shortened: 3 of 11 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -875,11 +875,17 @@ export class ProjectModel {
...
+                                      throw new Error(
+                                          `Chart ${d.saved_query_id} has no space_id`,
+                                      );
+                                  }
                                   const createChart = {
                                       ...d,
                                       saved_query_id: undefined,
                                       saved_query_uuid: undefined,
                                       space_id: getNewSpace(d.space_id),
+                                      dashboard_uuid: null,
```
ZeRego, line 888: We are going to support preview content in a following PR

packages/backend/src/database/migrations/20230718093901_add_dashbord_uuid_to_chart_table.ts
```diff
@@ -0,0 +1,21 @@
...
+        // make space id nullable
+        t.integer('space_id').nullable().alter({ alterType: false });
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('saved_queries', (t) => {
+        t.dropColumns('dashboard_uuid');
+        // make space id non nullable
+        t.integer('space_id').notNullable().alter({ alterType: false });
```
rephus, line 19: we probably need to remove all the saved queries with space_id null to apply this

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -98,6 +98,7 @@ export class DashboardModel {
 
     private static async createVersion(
         trx: Transaction,
+        dashboardUuid: string,
```
rephus, line 101: is this not optional ? `dashboardUuid: string | undefined`

rephus:
> This branch has not been deployed

Can we get this in render ?

packages/backend/src/models/DashboardModel/DashboardModel.ts
ZeRego: Even though it might not be used, we don't know at this point, so it is mandatory. Note this is a method to create a dashboard version, not a chart version.

packages/backend/src/database/migrations/20230718093901_add_dashbord_uuid_to_chart_table.ts
ZeRego: good catch 👍

packages/backend/package.json
```diff
@@ -116,7 +116,7 @@
         "migrate-production": "knex migrate:latest --knexfile dist/database/knexfile.js",
         "seed": "knex seed:run --knexfile src/database/knexfile.ts",
         "seed-production": "knex seed:run --knexfile dist/database/knexfile.js",
-        "rollback-last": "knex migrate:rollback --knexfile src/database/knexfile.ts",
```
ZeRego, line 119: `knex migrate:rollback` is to rollback the last batch of migrations. eg: If you migrated everything in 1 go, it would rollback everything. 

We just want to rollback the last migration script.

rephus:
Do we have this API dashboard twice because we run e2e tests twice ? 

image

perhaps we want to delete it "at the beggining" of a new e2e test

rephus:
I can't edit the title or the content of a chart in dashboard

rephus:
I haven't tested scheduled delivieres for these charts, but I can see they can be previewed on /minimal, so I assume they work

rephus:
Interesting behaviour on these chart in dashboard options: 

image

- save chart as: creates a new chat that doens't belong to a dashboard
- add to dashboard: you can add this chart to another dashboard, but I guess this will be deleted as well if the original dashboard gets deleted. 
- move to space: errors (tries to update the chart, see previous comment)

rephus (changes requested):
I think the only thing we need to fix is the ability to update chart in dashboards. Everything else seems ok, or not a blocker.

ZeRego:
@rephus 
> Do we have this API dashboard twice because we run e2e tests twice ?
> perhaps we want to delete it "at the beggining" of a new e2e test

We already delete these dashboards and charts at the beginning of the test. This may happen when the test retries.

> I can't edit the title or the content of a chart in dashboard

Fixed chart summary query and added test.

Screenshot 2023-07-21 at 10 22 18


> Interesting behaviour on these chart in dashboard options:

We have tickets in the milestone to disable some of these options. Shouldn't be an issue while the feature is under a feature flag.

rephus (approved):
Everything works now. I'm happy about your comments.

Just 1 minor bug, we can fix it on a separate PR, or even disable the `move to space`  option int he UI branch 

If I move a chart in dashboard into another space (which doesn't make sense anyway) , the chart becomes "visible" 

image

ZeRego:
> If I move a chart in dashboard into another space (which doesn't make sense anyway) , the chart becomes "visible"

We will have the option to move charts from dashboards to the space. The BE will set the dashboard to null in that case. In your case we set the space uuid but didn't set the dashboarduuid to null causing that problem.

Issue #6273, opened 2023-07-11
Use same z-indexes for floating components in mantine and blueprint


IrakliJani:
### Description

we have a lot of bugs because of mantine's and blueprint's floating elements use different z-indexes

spend timeboxed amount to find out if it's easy to make them use same zIndex space

IrakliJani:
mantine z-index:
https://github.com/mantinedev/mantine/blob/d03cb692eedd7e525275bdedcc9b077b141f0b78/src/mantine-styles/src/theme/utils/get-default-z-index/get-default-z-index.ts#L9

IrakliJani:
blueprint z-index:
https://github.com/palantir/blueprint/blob/bfbf04ee89fdc60989bd6eac5d2204d8325e4850/packages/core/src/common/_variables.scss#L79-L83

PR #6385, opened 2023-07-18
chore: mantine and blueprint z indices

IrakliJani:
Closes: #6273

### Description:

- alters blueprint-core.css with new z-indexes that is aligned with mantine
- clones blueprint-popover2.css
- alters blueprint-popover2.css with new z-indexes
- cleans up mantine z-index patches

tested:
- homepage / content pages / action modals and menus
- settings + project settings
- dashboards with modals and menus + dashboard filters
- explore:
  - all chart configs
  - custom filters
  - filters
  - sorting
  - limits
  - export csv
  - table / viz context menus and underlying data + etc..

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
+++ b/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
@@ -184,7 +184,6 @@ export const CustomMetricModal = () => {
     return item ? (
         <Modal
             size="xl"
-            zIndex={15}
             onClick={(e) => e.stopPropagation()}
             opened={isOpen}
             onClose={() => toggleModal(undefined)}
--- a/packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
+++ b/packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
@@ -67,7 +67,6 @@ export const SearchIconWithIndicator: FC<{
                 multiline: true,
                 offset: -2,
                 position: 'bottom',
-                zIndex: 201,
             }}
             tooltipLabel={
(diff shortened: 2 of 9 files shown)


magnew (approved):
Nice work. I checked this out and tested it by putting mantine components in blueprint modals, then by putting blueprint components in mantine modals. I couldn't find any issues, but that doesn't mean there aren't any. 

The blueprint css update set up seems a little fragile to me if we were going to keep this, but if we do it right, we will be removing blueprint. 

Did you consider doing it the other way, setting zIndexes on the Mantine components to match blueprint? It might not work better, but it might help us keep this a little more contained and understandable. We might be able to get away with not much more than something like this in the mantine theme:

```   
zIndex: {
          dropdown: 20,
          sticky: 20,
          fixed: 20,
          backdrop: 20,
          modal: 20,
          popover: 20,
          tooltip: 40,
},
```

Maybe different numbers, but you get the idea.

That said, this seems to work and I'm fine with checking it in. As long as we kill it soon 😄

IrakliJani:
@magnew great point on doing the other way around.

here was my reasoning behind the decision:
- mantine uses css-in-js thus all zIndex-ing logic lives in their source code
  - it is hard to find all occurences in the source code and I don't want to assume that they respect zInices based on the configuration (I've seen few of their components that do not respect the theme) it's rather on not trusting than avoiding a use a bad practice (in this case altering blueprint's raw css)
- blueprint utilizes sass (or less) to build one giant blob of css so there is a reliable way to predict which z-index is used where.
  - we already have a modified blueprint css (which is altered btw to make it compatible with blueprint, and we don't plan to bump blueprint versions anymore)

Issue #6389, opened 2023-07-18
Using multiple filters generates ambiguous sql


owlas:
### Description

If you select a filter on a metric but don't have it selected, Lightdash adds it automatically to the sql query (nice). But if you reference the same filter multiple times, it's added multiple times.

Example from demo.lightdash.com:

![CleanShot 2023-07-18 at 16 27 44@2x](https://github.com/lightdash/lightdash/assets/11660098/7eb87def-6ec9-4960-a229-cf98fd82ed05)

This doesn't happen if the metric in the filter is selected in the query (it appears in the `metricSelects` not the `filterMetricSelects` in `queryBuilder.ts`)

Metric references should be de-duplicated here: 

https://github.com/lightdash/lightdash/blob/e06a4543d9253e400420dbd3be66867aaf9e4a53/packages/backend/src/queryBuilder.ts#L163-L183

PR #6392, opened 2023-07-18
fix: deduplicate metric sql from filters

owlas:
Closes: #6389

<pending PR preview for screenshot>

### DIFF
--- a/packages/backend/src/queryBuilder.ts
+++ b/packages/backend/src/queryBuilder.ts
@@ -175,10 +175,8 @@ export const buildQuery = ({
             explore,
             compiledMetricQuery,
         );
-        return [
-            ...acc,
-            `  ${metric.compiledSql} AS ${fieldQuoteChar}${alias}${fieldQuoteChar}`,
-        ];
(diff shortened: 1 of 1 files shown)


rephus (approved):

packages/backend/src/queryBuilder.ts
```diff
@@ -175,10 +175,8 @@ export const buildQuery = ({
             explore,
             compiledMetricQuery,
         );
-        return [
-            ...acc,
-            `  ${metric.compiledSql} AS ${fieldQuoteChar}${alias}${fieldQuoteChar}`,
-        ];
+        const renderedSql = `  ${metric.compiledSql} AS ${fieldQuoteChar}${alias}${fieldQuoteChar}`;
+        return acc.includes(renderedSql) ? acc : [...acc, renderedSql];
```
IrakliJani, line 179: @owlas  is there a chance that the same sql chunk will be repeated in some other part of the generated sql?

IrakliJani: I guess this is localised to a one where group

PR #6395, opened 2023-07-18
chore: pin dbt-trino to 1.5.0

joaoviana:
Closes: N/A

### Description:

After this release: https://github.com/starburstdata/dbt-trino/blob/master/CHANGELOG.md

our Docker build started failing. First instance: https://github.com/lightdash/lightdash/actions/runs/5579941834/jobs/10196171946

Pinning `dbt-trino` to its previous version works, but this is a temporary solution - we need to figure out which deps are clashing and how to handle it: on the code and/or by reporting a bug

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -39,7 +39,7 @@ RUN /usr/local/dbt1.5/bin/pip install \
     "dbt-snowflake~=1.5.0" \
     "dbt-bigquery~=1.5.0" \
     "dbt-databricks~=1.5.0" \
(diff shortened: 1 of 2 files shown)


IrakliJani (approved):
I guess this will not pass on render until we merge it, so approving ✅ 

```
docker succesfully pulled trino 1.5.0
#12 2.332 Collecting dbt-trino==1.5.0
#12 2.346   Downloading dbt_trino-1.5.0-py3-none-any.whl (32 kB)
```

Issue #6258, opened 2023-07-11
Migrate "Table config" general elements to mantine 


IrakliJani:
### Description

Migrate remaining items to mantine

acceptance criteria:
- [x] use mantine popover
- [x] use mantine tabs for general / conditional formatting (pie charts config as a reference)
- [x] use mantine form elements for "options"
  - <img width="211" alt="CleanShot 2023-07-11 at 18 44 06@2x" src="https://github.com/lightdash/lightdash/assets/962095/60f2da7b-1823-4ec2-8937-d2da027cf333">

<img width="287" alt="CleanShot 2023-07-11 at 18 44 53@2x" src="https://github.com/lightdash/lightdash/assets/962095/df09712a-e7da-48a0-ad0d-3e2cece07da9">

PR #6407, opened 2023-07-19
chore: table config popup in Mantine

magnew:
Closes: #6258 

### Description:

Moves the table config popup, tabs and checkboxes to Mantine. 

### Criteria
- [ ] use mantine popover
- [ ] use mantine tabs for general / conditional formatting (pie charts config as a reference)
- [ ] use mantine form elements for "options"

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
@@ -1,5 +1,5 @@
-import { Checkbox, FormGroup } from '@blueprintjs/core';
-import { Box, Title, Tooltip } from '@mantine/core';
+import { FormGroup } from '@blueprintjs/core';
+import { Box, Checkbox, Stack, Title, Tooltip } from '@mantine/core';
 import React, { FC, useCallback, useMemo, useState } from 'react';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
```diff
@@ -1,5 +1,5 @@
-import { Checkbox, FormGroup } from '@blueprintjs/core';
-import { Box, Title, Tooltip } from '@mantine/core';
+import { FormGroup } from '@blueprintjs/core';
```
IrakliJani, line 1: do we use FormGroup anywhere?

magnew: The drag and drop parts are still blueprint. Now that you did the z-index thing I think I can convert them too.

magnew: I looked a little more closely. It's still wrapping the column config thing that is in Blueprint. I'm planning to port that in the next patch.

IrakliJani: perfect

Issue #6358, opened 2023-07-17
When I remove the only chart that's using a filter on a dashboard, the filter stays


yasmin-e:
### Description
When I remove the only chart that's using a filter on a dashboard, the filter stays with an error message.

https://github.com/lightdash/lightdash/assets/67699259/5152e406-5456-49a2-9c42-84ef1a634971

### To Reproduce
- Go to a dashboard with at least 2+ charts
- Apply a filter to a single chart
- Remove that chart from the dashboard
- An error message appears in the place of the filter

### App Version
Version: v0.667.0

TuringLovesDeathMetal:
Suggested solution: If there are no charts that the filter applies to, the filter is removed from the filters list automatically.

PR #6411, opened 2023-07-19
fix: address ux after only chart with filter applied to is removed from dashboard

joaoviana:
Closes: #6358 

### Description:

Remove active filter if not targeting any field id

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/bc086d5d-4ffe-4f80-8939-10a32616bb51

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
@@ -25,34 +25,38 @@ const ActiveFilters: FC<ActiveFiltersProps> = ({ isEditMode }) => {
 
     return (
         <ActiveDashboardFiltersWrapper>
-            {dashboardFilters.dimensions.map((item, index) => (
-                <ActiveFilter
-                    key={item.id}
-                    isEditMode={isEditMode}
-                    fieldId={item.target.fieldId}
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
```diff
@@ -25,34 +25,38 @@ const ActiveFilters: FC<ActiveFiltersProps> = ({ isEditMode }) => {
...
-                    field={fieldMap[item.target.fieldId]}
-                    filterRule={item}
-                    onRemove={() => removeDimensionDashboardFilter(index, true)}
-                    onUpdate={(value) =>
-                        updateDimensionDashboardFilter(value, index, true)
-                    }
-                />
-            ))}
+            {dashboardFilters.dimensions
+                .filter((item) => !!fieldMap[item.target.fieldId])
```
magnew, line 29: Since you are just doing a filter here, it doesn't get removed in the DB, right? So does the filter come back if you add back a chart that *was* using it?

Will that be surprising to people? Or maybe that is actually what we want.

joaoviana: Cool! I'm glad the video explained it 😅

Issue #6193, opened 2023-07-07
We can't copy SQL code on Chrome 


rephus:
### Description

It seems the shortcut CTRL+C is not working on chrome on explore

It seems this is affecting anythign I try to copy on explore , except results or filters. 
Like SQL, or static text in the page like "Refresh dbt" or "Table name" (eg: Tables / Orders) 

Also we can't copy the name of the chart from a dashboard

Copying code from SQL runner seems ok 

This is not happening on Firefox 

Reported on windows. Chrome. Version 114.0.5735.199 (Official Build) (64-bit)
Tested on Linux Version 112.0.5615.121 (Official Build) (64-bit)

PR #6416, opened 2023-07-19
fix: unlock clipboard copy when visualising tables

joaoviana:
Closes: #6193 

### Description:

When visualising tables, the `ValueCell` component that currently uses `useHotKeys` would allow the copy of the `Cell` value if user pressed `mod + c`. However, this is blocking the copy-to-clipboard functionality everywhere else that is not a `ValueCell`. 
I believe the usage of the `useHotKeys` is affecting the other elements due to the way we're implementing the `ValueCellMenu`+`ValueCell`. 

Instead of adding the `onKeyDown` events to the `td` elements, we listen to: 
* is isMenuOpen truthy
* if user presses the copy keyboard shortcut (we now use `getHotkeyHandler`  https://mantine.dev/hooks/use-hotkeys/#targeting-elements )

NOTE: `td` elements aren't focusable by nature, so this addresses that

Don't have a screen-recording bc Quicktime player + keyboard shortcuts aren't working as expected for me. PLease use Render to review UX

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/ValueCell.tsx
+++ b/packages/frontend/src/components/common/PivotTable/ValueCell.tsx
@@ -8,7 +8,7 @@ import {
     ResultValue,
     TableCalculation,
 } from '@lightdash/common';
-import { useClipboard, useHotkeys } from '@mantine/hooks';
+import { useClipboard } from '@mantine/hooks';
 import { FC, useCallback, useMemo, useState } from 'react';
 
 import { getColorFromRange, readableColor } from '../../../utils/colorUtils';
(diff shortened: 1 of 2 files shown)


IrakliJani (commented):
@joaoviana 
makes sense and code looks great. but we also have another place where we should be having the same problem. check `packages/frontend/src/components/common/Table/TableProvider.tsx`

joaoviana:
> @joaoviana makes sense and code looks great. but we also have another place where we should be having the same problem. check `packages/frontend/src/components/common/Table/TableProvider.tsx`

we don't have that problem from what I've tested! But will double-check 👍

joaoviana:
> @joaoviana makes sense and code looks great. but we also have another place where we should be having the same problem. check `packages/frontend/src/components/common/Table/TableProvider.tsx`

thanks for spotting this @IrakliJani , it was working absolutely fine in dev mode locally, but not in Render 🤯 

Removed that usage and replaced with the other handler instead, and all working fine 👍

IrakliJani (approved):
looks good!

Issue #6351, opened 2023-07-17
Disallow models called `lightdash` `ld` or containing `.`


owlas:
### Description 

For our references to work in sql `${ model.column }` or `${ lightdash.user.email }` we cannot allow models called `lightdash` or `ld` or contain `.` 

If during compiling a lightdash project, we find these, we should throw an error.

rephus:
@owlas by model you mean we should throw an error if a `table` , `dimension` or `metric` is called `ld` or `lightdash` ?

rephus:
Should we do this on the `schema` validation ? 

![image](https://github.com/lightdash/lightdash/assets/1983672/b6c0da8c-192a-4a8e-86d5-3a7b80db66f4)

![image](https://github.com/lightdash/lightdash/assets/1983672/6d648339-dcbe-42fa-a48d-42723ebc074d)

owlas:
1. `${ dimension }` reference a field sql
2. `${ TABLE }` reference current table sql
3. `${ table_a.dimension_a }` reference a field sql from a joined table `table_a`
4. `${ lightdash.something }` access something from the lightdash namespace

Only `3` and `4` clash when the table name is `lightdash` or `ld` - so not sure we need to check fields

owlas:
Another decision is how default user metadata is handled vs. attributes 

For example, all users have a `userUuid` `email` `role` etc. So we could expose that in future. But if we did:

- `lightdash.user.email`
- `lightdash.user.uuid`
- `lightdash.user.attr_1`

^ then `email` and `uuid` would clash with attributes named `email` and `uuid`.

Should we scope attributes to a namespace?

- `lightdash.user.email`
- `lightdash.user.uuid`
- `lightdash.user.attributes.attr_1`

^ but then the references are starting to look quite long, so it'll be frustrating to type out.

rephus:
Edit: this schema property will check that we don't allow models called lightdash or ld or contain .

![image](https://github.com/lightdash/lightdash/assets/1983672/eaae531d-459b-40fd-9547-c2b94cf66698)

owlas:
Nice!

owlas:
You could also make it `"^[^.]+$"` to exclude empty strings

PR #6425, opened 2023-07-20
feat:  Disallow models called lightdash ld or containing .

rephus:
Closes: #6351

## Dimension == TABLE
Screenshot from 2023-07-20 11-52-18

image

Screenshot from 2023-07-20 11-56-13

## Model == lightdash
image

image

image

## Model contains '.'

image

image

image


owlas (commented):
The validation is useful but do we not want some business logic in the compiler that throws this error specifically?

rephus:
> The validation is useful but do we not want some business logic in the compiler that throws this error specifically?

I think we need to choose between validation or doing it `in code` , because afaik the validation happens before and it blocks the compilation.

I think I prefer the validation, because I think this is where this logic belongs, and it is easier to understand and update. The only thing I don't like about the validation is that the error is not very meaningful (specially the pattern)

owlas (approved):

PR #6429, opened 2023-07-20
fix: add cursor: pointer to checkboxes and radios

magnew:
### Description:

For some reason Mantine doesn't apply cursor: pointer to checkboxes and radios by default. This sets the setting to do so. 

More here:
https://mantine.dev/theming/theme-object/#cursortype


rephus (approved):
Tested, it look good !

Issue #6334, opened 2023-07-14
When I edit a chart's title from explore, it is not reflected on the dashboard


yasmin-e:
### Description

https://github.com/lightdash/lightdash/assets/67699259/e2e517f2-973d-49b6-80c5-3f8a01d6cde5

### To Reproduce
1. From a dashboard, click on edit chart
2. In explore, edit the chart's title
3. Go back to the dashboard
4. The title is not up to date

### App Version
Version: v0.667.4

PR #6432, opened 2023-07-20
fix: chart tile title undefined

rephus:
Closes:  #6334

### Description:

[Screencast from 20-07-23 16:16:23.webm](https://github.com/lightdash/lightdash/assets/1983672/2bdea71a-97fa-4c44-a63d-3d1a2664c764)


packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -495,7 +495,6 @@ export class DashboardModel {
                 this.database.raw(
                     `COALESCE(
                         ${DashboardTileChartTableName}.title,
-                        ${SavedChartsTableName}.name,
```
rephus, line 498: This is the fix, if we keep title undefined, everything works as expected. And in the UI we already do `title || chart.name` in most places.

rephus:
tests faailed :eyes:

Issue #4143, opened 2023-01-11
Bug on format when using multiple series on left and right Y axes


rephus:
### Description

![image](https://user-images.githubusercontent.com/1983672/211773193-25a29980-f90f-4b53-912b-213a3be95cdc.png)

Left Y axis should be of percent format, and not $ 

![image](https://user-images.githubusercontent.com/1983672/211773271-6d857c2e-3a69-46ba-b558-6e8ef706609c.png)

I think there is a bug when getting leftAxisYId and `rightAxisYId` because I get the same on both

![image](https://user-images.githubusercontent.com/1983672/211774604-d1fe7f40-3d1a-48e7-b709-618cacecaea7.png)

PR #6435, opened 2023-07-20
fix: add default yAxisIndex value to avoid breaking series behaviour

joaoviana:
Closes: #4143 

### Description:

When creating a chart for the first time, the `yAxisIndex` for each serie is not defined. 
What the bug reported describes is when the user sets 2 metrics on the Y Axis, and then chooses to move the first one to the right axis, the remaining one will then break the visualisation, as its `yAxisIndex` is not defined and our current code doesn't catch that. 

Instead, we can set a default value for `yAxisIndex` when users create a chart. 
However, we need to address the case when users have a _broken-yAxisIndex_-type of chart already saved. The FE should do proper mapping of the `series` and add the default value when it's not defined. There's a test case for this. 

Also, added a React hook test for `useCartesianChartConfig`. 
For this to work, I've done the following: 
- downgraded react testing library packages to the most recent version that supports React <18
- added `react-test-renderer` - required for running React hook tests

Note: @ZeRego thanks for pointing out that I should have a look at `xAxisIndex` behaviour as well. I didn't see any odd behaviour, but pls let me know if you encounter something else 👍


packages/frontend/src/hooks/echarts/useEcharts.ts
```diff
@@ -858,7 +858,7 @@ const getEchartAxis = ({
     const leftAxisYId = validCartesianConfig.layout.flipAxes
         ? validCartesianConfig.layout?.xField
         : validCartesianConfig.eChartsConfig.series?.find(
-              (serie) => serie.yAxisIndex === 0,
+              (serie) => serie.yAxisIndex === 0 || !serie.yAxisIndex, // NOTE: yAxisIndex is undefined on first time adding a serie
```
joaoviana: when getting the `leftAxisYId`, check if there's an element in the series that has a `yAxisIndex` 0 or undefined (if they're on the right axis, the value is always set on change of the html select)

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -614,7 +614,13 @@ const useCartesianChartConfig = ({
             isCompleteEchartsConfig(dirtyEchartsConfig)
                 ? {
                       layout: dirtyLayout,
-                      eChartsConfig: dirtyEchartsConfig,
+                      eChartsConfig: {
+                          ...dirtyEchartsConfig,
+                          series: dirtyEchartsConfig.series.map((serie) => ({
+                              ...serie,
+                              ...(!serie.yAxisIndex && { yAxisIndex: 0 }),
+                          })),
```
joaoviana, line 622: default any serie that does not have a `yAxisIndex` to `0`... 🤔

packages/common/src/types/savedCharts.ts
```diff
@@ -163,7 +163,7 @@ export type Series = {
     };
     name?: string;
     color?: string;
-    yAxisIndex?: number;
```
ZeRego, line 166: The problem with making this type of change ( optional to required ) is that we aren't migrating the existing chart configs.

joaoviana: I seeee, so we should change this back to optional and then address the possible scenarios on the FE only... thanks!
Will create a chart (without the default value) then bring back this code and see how it behaves...

ZeRego (approved):

Issue #6354, opened 2023-07-17
Improve tooltips in pie charts


IrakliJani:
### Description

- [ ] when hovering over to a slice it should show group label and value
  - <img width="366" alt="CleanShot 2023-07-17 at 13 26 15@2x" src="https://github.com/lightdash/lightdash/assets/962095/9b2be5ab-3c81-43d9-bbf0-389a9c95cdae">

- [ ] if there are no value/percentage selected it should just show the group label instead of nothing
  - <img width="180" alt="CleanShot 2023-07-17 at 13 24 44@2x" src="https://github.com/lightdash/lightdash/assets/962095/b847b099-fff1-4b66-878d-7e9eed906777">

PR #6436, opened 2023-07-20
fix: pie charts tooltip and label configuration

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6354

### Description:

- [x] when hovering over to a slice it should show group label and value
- [x] if there are no value/percentage selected it should just show the group label instead of nothing

⚠️ it seems like echarts or echarts react wrapper has a bug rendering "label lines" when changing configuration and happens when changing from "value label" between "Inside" and "Outside"

CleanShot 2023-07-20 at 21 56 31@2x

CleanShot 2023-07-20 at 22 00 13@2x


yasmin-e (approved):
tested and it works as expected 👍🏼

Issue #6420, opened 2023-07-20
Create a `conditional_formatting_rule_saved` analytics event


TuringLovesDeathMetal:
- create an analytics event `conditional formatting rule saved` 
- This event is triggered when a user creates a saved chart with a conditional formatting rule applied to it. 
- There is one row per conditional formatting rule applied to the chart (let me know if this isn't possible/is super complicated and I can rethink the event structure...)
- The parameters are:
  - user_id
  - organization_id
  - project_id
  - chart_id - I think this might be saved_query_id?
  - timestamp
  - type - either `single color` or `color range`
  - num_conditions - the total number of conditions applied to the rule

PR #6437, opened 2023-07-20
feat: conditional formatting analytics events

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6420

### Description:

events are sent per conditional formatting rule:
```js
[
{
  event: 'conditional_formatting_rule.saved',
  userId: 'b264d83a-9000-426a-85ec-3f9c20f368ce',
  properties: {
    projectId: '3675b69e-8324-4110-bdca-059031aa8da3',
    organizationId: '172a2270-000f-42be-9c68-c4752c23ae51',
    savedQueryId: '2a229b98-5db8-441c-858d-46ad08ebc229',
    type: 'single color',
    numConditions: 2
  }
},
{
  event: 'conditional_formatting_rule.saved',
  userId: 'b264d83a-9000-426a-85ec-3f9c20f368ce',
  properties: {
    projectId: '3675b69e-8324-4110-bdca-059031aa8da3',
    organizationId: '172a2270-000f-42be-9c68-c4752c23ae51',
    savedQueryId: '2a229b98-5db8-441c-858d-46ad08ebc229',
    type: 'color range',
    numConditions: 1
  }
}
]
```

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -269,6 +269,18 @@ export type DuplicatedChartCreatedEvent = BaseTrack & {
     };
 };
 
+export type ConditionalFormattingRuleSavedEvent = BaseTrack & {
+    event: 'conditional_formatting_rule.saved';
+    userId: string;
+    properties: {
+        projectId: string;
+        organizationId: string;
+        savedQueryId: string;
+        type: 'single color' | 'color range';
+        num_conditions: number;
+    };
+};
+
 type ProjectEvent = BaseTrack & {
     event: 'project.updated' | 'project.created';
     userId: string;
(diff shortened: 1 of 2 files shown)


IrakliJani:
@TuringLovesDeathMetal

I am triggering this event only on chart creation, not when a chart is updated to a new version. Do we need a separate event for chart updates?

If we also want to capture chart updates, we should consider the following: The system will send the event every time the user clicks 'Save', regardless of whether there were changes to the conditional formatting or not. 

This could happen unless we compare what is being saved and only send the event if the 'Save' includes new conditional formatting rules, but this could get complicated. So, I wanted to ask if it's worth implementing.

yasmin-e (approved):

TuringLovesDeathMetal:
@IrakliJani can we trigger this same event when a chart is updated? 

It's totally fine if there are duplicates - I can just filter to the latest timestamp + chart_id. 

Does that make sense? 

So every time a chart is created, or updated (i.e. saved), we trigger this conditional formatting event.

IrakliJani:
@TuringLovesDeathMetal yep, that sounds good

Issue #6350, opened 2023-07-17
`sql_filter` can reference user attributes


owlas:
### Description

The `sql_filter` field on a model #6349 should include `${ lightdash.attribute.attr_1 }` where `attr_1` is a user attribute belonging to that user.
- If `attr_1` doesn't exist, the query builder should throw a ForbiddenError
- Should interpret the value at run time
- Can use `ld` as an alias for `lightdash`

PR #6445, opened 2023-07-21
feat: sql filter user attributes

rephus:
Closes:  #6350
### Description:

image

Screenshot from 2023-07-24 11-07-33

Screenshot from 2023-07-24 11-09-12

Screenshot from 2023-07-24 11-09-01

With strings 

Screenshot from 2023-07-24 11-32-13

On join (ignore) , we only use base table

image


packages/backend/src/services/UserAttributesService/UserAttributesService.ts
```diff
@@ -93,4 +93,41 @@ export class UserAttributesService {
...
+                );
+            }
+            if (userAttribute.users.length !== 1) {
+                throw new ForbiddenError(
+                    `Invalid or missing user attribute "${attribute}" on sqlFilter "${sqlFilter}"`,
+                );
+            }
+            return acc.replace(
+                sqlAttribute,
+                `${stringQuoteChar}${userAttribute.users[0].value}${stringQuoteChar}`,
```
rephus, line 127: I'm wrapping the value into "string quotes" , this is not needed for numbers, but it is needed for strings or other values. I tested it works for both on postgres . Another option ist o tell customers to wrap their own data in quotes like this: 

`'${ld.attr.string}' == 'US'`

packages/backend/src/services/UserAttributesService/UserAttributesService.ts
```diff
@@ -93,4 +93,41 @@ export class UserAttributesService {
         }
         await this.userAttributesModel.delete(orgAttributeUuid);
     }
+
+    static replaceUserAttributes(
```
owlas, line 97: I think this code should live closer to the sql generation than with the user attribute service code. Although it's business logic, it's function is much more closely related to the `queryBulider` than it is to event tracking, loading data, etc.

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -16,9 +16,12 @@ import {
     Metric,
 } from '../types/field';
 import { WarehouseClient } from '../types/warehouse';
+
 import { renderFilterRuleSql } from './filtersCompiler';
 
-export const lightdashVariablePattern = /\$\{([a-zA-Z0-9_.]+)\}/g;
+// exclude lightdash prefix from variable pattern
+export const lightdashVariablePattern =
+    /\$\{((?!lightdash\b)[a-zA-Z0-9_.]+)\}/g;
```
owlas, line 24: should this also include `ld` ?

packages/common/src/compiler/exploreCompiler.test.ts
```diff
@@ -207,3 +207,24 @@ describe('Compile metrics with filters', () => {
         );
     });
 });
+
+describe('Parse dimension reference', () => {
```
owlas, line 211: good tests! 

should also include `ld` ?

owlas (changes requested):
Request change to co-locate the sql generation code (including new user attributes stuff).

Also I noticed we haven't changed any of the `buildQuery` usages, so this won't take affect in the UI right?

packages/backend/src/queryBuilder.ts
```diff
@@ -52,16 +54,55 @@ const getOperatorSql = (filterGroup: FilterGroup | undefined) => {
...
 };
 
+export const replaceUserAttributes = (
+    sqlFilter: string,
+    userAttributes: UserAttribute[],
+    stringQuoteChar: string = "'",
+): string => {
+    const userAttributeRegex =
+        /\$\{(?:lightdash|ld)\.(?:attribute|attr)\.(\w+)\}/g;
+    const sqlAttributes = sqlFilter.match(userAttributeRegex);
```
owlas, line 64: We're checking for known attributes but the user might reference an attribute that doesn't exist. 

We could use `liquidjs` like we do in the CLI to parse this reference without having to regex ourselves

owlas (changes requested):
The pattern matching only works if user references a known attribute, if I never defined it then my template stays in the sql creating a syntax error.

Suggest we use liquidjs (already a dependency for dimension urls) to handle replacing. We can only use this after we've resolved all our ${TABLE}/${dimension} type references.

https://liquidjs.com/index.html

```yaml
 version: 2
 models:
   - name: events
     meta:
       sql_where: ${TABLE}.date >= '2020-08-01'
       sql_filter: ${date} >= '2020-07-01' AND ${lightdash.attributes.foo} = 'foo'
```

CleanShot 2023-07-25 at 15 54 04@2x

packages/backend/src/queryBuilder.ts
owlas: Sorry my bad!!! I used `attributes` instead of `attribute`

rephus: now we support also `attributeS`

owlas (approved):

rephus:
we've manually changed the schema on this render, so it is expected for e2e tests to fail. They were working before the testing and since then the changes have been small.

Issue #6449, opened 2023-07-21
Migrate chart config floating elements to Mantine


magnew:
### Description

To migrate the chart config to Mantine, we need to move all of the Blueprint floating elements to mantine first. We've found that mixing Blueprint floating elements with Mantine floating elements causes problems. 

#### Acceptance criteria
- [x] Move 'Layout' tab selects to Mantine
- [x] Move the series color picker to Mantine
- [x] Move the reference line select to Mantine
- [x] Use Mantine popover instead of Blueprint

PR #6461, opened 2023-07-24
refactor: chart config floating elements in mantine

magnew:
Closes:  #6449 

### Description:

Moves the chart configuration floating elements (Selects, color picker, the main popups) to Mantine. This is necessary first so that we don't have conflicting functionality between the containing popup and the blueprint/mantine floating elements. Note that many of the inputs are still blueprint-based but they are elements that should be fine in the mantine popup. 

**For reviewers**: I tested this a lot, but please give it a spin and see if you find any issues. I could have missed something. 

#### Acceptance criteria
- [ ] Move 'Layout' tab selects to Mantine
- [ ] Move the series color picker to Mantine
- [ ] Move the reference line select to Mantine
- [ ] Use Mantine popover instead of Blueprint


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SeriesColorPicker.tsx
IrakliJani, line 1: 🎉

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+import FieldSelectItem from '../FieldSelectItem';
+
+interface Props {
+    selectedField?: Field | TableCalculation;
+    fieldOptions: (Field | TableCalculation)[];
+    placeholder?: string;
+    onChange: (newValue: string | null) => void;
+}
+
+const FieldSelect: FC<Props> = ({
```
IrakliJani, line 15: (not in the scope of this PR) feels like we are recreating this for every single config we have and we've only been reusing `FieldSelectItem`. I believe there's one one in pie charts/series and table/condit-formatting

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldLayoutOptions.tsx
```diff
@@ -165,17 +165,16 @@ const FieldLayoutOptions: FC<Props> = ({ items }) => {
...
-                                        : availableYFields
-                                }
-                                activeField={activeField}
-                                onChange={(item) => {
-                                    updateYField(index, getItemId(item));
+                            <FieldSelect
+                                selectedField={activeField}
+                                fieldOptions={yFieldsOptions}
+                                onChange={(newValue) => {
+                                    updateYField(index, newValue ?? '');
```
IrakliJani, line 177: should we be allowing setting to '' or we should do the same as `setXField`

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/index.tsx
```diff
@@ -1,30 +1,38 @@
...
 import ChartConfigTabs from './ChartConfigTabs';
 
 const ChartConfigPanel: React.FC = () => {
     const eChartsOptions = useEcharts();
     const disabled = !eChartsOptions;
-
     return (
-        <Popover2
+        <Popover
+            withinPortal={true}
```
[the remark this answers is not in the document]
magnew: Yea, this was a late addition :). It's because when you click 'Flip axes' on the bar chart, it becomes a horizontal bar chart; that causes the buttons to change size, and the popup jumps around a bit. Putting it in the portal fixes this. The other configs don't need this cause they don't have buttons that change the page layout while the panel is open. 

FWIW, this is kind of what the portal is for -- to take an element out of page layout and make it stay put. We have this panel as a dropdown at the moment, but it's really more like a persistent modal.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SeriesColorPicker.tsx
magnew: 🎉

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldSelect.tsx
magnew: I refactored it here, but I have a ticket to use it everywhere else. I'hope to get to that soon.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldLayoutOptions.tsx
magnew: I'm not sure it matters functionally, but you're right that they should be consistent. I think this is kinda more 'correct' than setting it to undefined. We aren't controlling these components here (which is a little odd), but for controlled components react doesn't like undefined. So this might be a better default answer than undefined. Although null might be even better. Let me see if Mantine has an opinion.

IrakliJani: ok sounds good!

IrakliJani (approved):
code looks good + tested in render and everything seems to be working fine. great work!

Issue #6344, opened 2023-07-14
`time_intervals: OFF` fails Lightdash validation


owlas:
### Description

If you use `time_intervals: OFF` (docs: https://docs.lightdash.com/references/dimensions#to-turn-off-time-intervals-for-a-dimension-set-time_intervals-off)

Then doing `lightdash compile` throws an error that it's invalid syntax and `time_intervals` must be an array. 

Blocking a user: https://lightdash.slack.com/archives/C048FE7206Q/p1689179823303329

PR #6463, opened 2023-07-24
fix: time interval off boolean

rephus:
Closes: #6344

### Description:

 In this case, "OFF" is a valid YAML boolean value, and without quotes, it will be interpreted as a boolean rather than a string.

So I've added support for `boolean` 

Screenshot from 2023-07-24 15-44-08

Screenshot from 2023-07-24 15-43-57

Screenshot from 2023-07-24 15-43-43

Screenshot from 2023-07-24 15-43-29


packages/common/src/compiler/translator.mock.ts
```diff
@@ -898,6 +910,12 @@ export const MODEL_WITH_OFF_TIME_INTERVAL_DIMENSIONS: DbtModelNode & {
     columns: COLUMN_WITH_OFF_TIME_INTERVALS,
 };
 
+export const MODEL_WITH_OFF_BOOLEAN_TIME_INTERVAL_DIMENSIONS: DbtModelNode & {
+    relation_name: string;
+} = {
+    ...model,
+    columns: COLUMN_WITH_OFF_TIME_INTERVALS,
```
ZeRego, line 917: Should it be ?
```suggestion
    columns: COLUMN_WITH_OFF_BOOLEAN_TIME_INTERVALS,
```

ZeRego (changes requested):

ZeRego (approved):

Issue #6262, opened 2023-07-11
Migrate "Chart Config - Layout Tab" to Mantine


IrakliJani:
### Description

acceptance criteria:

only migrate Layout tab content:
- [ ] Y and X axis field selections
- [ ] Fliping axes
- [ ] Adding groups (pivoting)
- [ ] Stacking

ignore:
- popover
- tab switcher

![CleanShot 2023-07-11 at 20 19 53@2x](https://github.com/lightdash/lightdash/assets/962095/dbed3ede-6c37-4a5e-b236-8d75b45288ad)

PR #6465, opened 2023-07-24
feat: 6262 migrate chart config layout tab to mantine

magnew:
Closes: #6262 

### Description:

Migrate chart 'Layout' config tab to mantine. Also removes a lot of custom styles.

#### Criteria
- [x] Y and X axis field selections
- [x] Fliping axes
- [x] Adding groups (pivoting)
- [x] Stacking


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldLayoutOptions.tsx
```diff
@@ -117,59 +112,63 @@ const FieldLayoutOptions: FC<Props> = ({ items }) => {
...
                     </Button>
                 ) : (
-                    <AxisFieldDropdown>
+                    <Group spacing="xs">
                         <FieldSelect
                             selectedField={xAxisField}
                             fieldOptions={items}
                             onChange={(newValue) => {
-                                setXField(newValue ?? undefined);
+                                setXField(newValue ?? '');
```
IrakliJani, line 147: im still not sure about this. can we have this instead `setXField(newValue ?? undefined)` and zselectedField={xAxisField ?? ''}`

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldLayoutOptions.tsx
```diff
@@ -178,119 +177,118 @@ const FieldLayoutOptions: FC<Props> = ({ items }) => {
...
+                            alignSelf: 'start',
+                        }}
+                        onClick={() =>
+                            setPivotDimensions([
+                                ...(pivotDimensions || []),
+                                getItemId(availableGroupByDimensions[0]),
+                            ])
+                        }
+                    >
+                        + Add
```
IrakliJani, line 274: can we use + icon instead?

IrakliJani (changes requested):
UI looks good, however group feature is not working the same way as it did before

here are before/after screenshots

before
CleanShot 2023-07-24 at 20 44 42@2x

after:
CleanShot 2023-07-24 at 20 47 50@2x

magnew:
> UI looks good, however group feature is not working the same way as it did before
> 
> here are before/after screenshots
> 
> before CleanShot 2023-07-24 at 20 44 42@2x
> 
> after: CleanShot 2023-07-24 at 20 47 50@2x

Ah, ok, I did this on purpose but I didn't know you could get it in a state where that input was empty and still present. It feels a little broken to have a tooltip on that whole bottom section when there is no input, but I will change it back.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldLayoutOptions.tsx
magnew: Sure!

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/FieldLayoutOptions.tsx
magnew: Sure. I imagine you know something about our internals that I don't 😄 . React and Mantine are both pretty opinionated about using the empty string as the empty value for inputs, so we probably should do this translation everywhere or investigate whether we are comfortable using empty string. But I'm happy to do this for now.

magnew: selectedField looks like it can't be undefined, so I will just change this callback.

IrakliJani (approved):
looks great! rerunning e2e

Issue #6417, opened 2023-07-19
Chart config is disabled when all series are hidden


owlas:
### Description

If you disable all the series in a grouped chart, the configuration is disabled

Example in demo: 

![CleanShot 2023-07-19 at 17 26 47](https://github.com/lightdash/lightdash/assets/11660098/99e87566-9aef-4f70-9907-2b9721a71382)

PR #6468, opened 2023-07-24
fix: check if configure is disabled with visualisation context

joaoviana:
Closes: #6417 

### Description:

When viewing a chart, and when there's grouped series and we hide them all, the `Configure` button gets disabled. 

Reason for this is that the usage of `useEcharts` does not give enough granularity for this scenario: when `series` is a `[]` but `rows`/results data has length > 0.

Instead, use `useVisualisationProvider` like on other config panels to disable the button correctly. 

Screen recording:

https://github.com/lightdash/lightdash/assets/7611706/93ba69ac-b8b1-4b24-b5bf-2d03facdcb46


yasmin-e (approved):
works as expected 👍🏼

Issue #6421, opened 2023-07-20
the year is cut off on the date picker in filters


TuringLovesDeathMetal:
### Description

<img width="533" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/08b331c2-a483-4d22-a603-a007621e496c">

joaoviana:
I can't replicate the cut-off effect @TuringLovesDeathMetal ... 

<img width="530" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/b9b4bc01-336c-458c-bfb0-c89d65e45dba">

hamzahc1:
@joaoviana what browser are you in? 

I was able to recreate on Mac in Chrome in the explorer
![CleanShot 2023-07-24 at 19 40 39](https://github.com/lightdash/lightdash/assets/14341285/1089efcc-5c0e-4fe4-a38e-c3ba022680e5)

joaoviana:
> @joaoviana what browser are you in?
> 
> I was able to recreate on Mac and in Chrome in the explorer ![CleanShot 2023-07-24 at 19 40 39](https://user-images.githubusercontent.com/14341285/255683719-1089efcc-5c0e-4fe4-a38e-c3ba022680e5.png)

same browser and setup... 
Was able to replicate it on a chart @TuringLovesDeathMetal sent me, will see if I can replicate the same locally, thanks @hamzahc1 !

PR #6470, opened 2023-07-25
style: add padding to datepicker caption select

joaoviana:
Closes: #6421 

### Description:

Padding isn't applied to date picker caption on production bundles: 
image

locally: 
image

Add style with higher specificity to override the padding not being set.


yasmin-e (changes requested):
Looks good! Just one thing while you're at it:
The picker for months is really close to the text compared to years, can we make them match?
Screenshot 2023-07-25 at 12 16 43

yasmin-e (approved):
looks great 🎉

PR #6473, opened 2023-07-25
chore: add `concurrently` to improve dev step and introduce `npm-run-all2`

joaoviana:
Closes: N/A

### Description:

Currently we use `npm-run-all` that hasn't been updated since 2018. 
Replaced it with `concurrently` and set colors and names for processes to easily see to which process the log belongs to. 

Since we have some scripts using `run-s` (https://github.com/mysticatea/npm-run-all/blob/master/docs/run-s.md), which basically runs a set of scripts sequentially, I added `npm-run-all2` to keep the current behaviour on those scripts (`concurrently` does not provide a command to run a set of scripts sequentially).

We could do the `dev` command with `npm-run-all2` as well but `concurrently` allows for better customisation/labelling of the processes that are being run so figured it would add more value to our devx. But happy to remove and just have 1 package for all instead!

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/acd82a68-8e2c-44c6-ab5a-b8a122390d2f

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
         ]
     },
     "dependencies": {
-        "concurrently": "^8.2.0"
+        "concurrently": "^8.2.0",
+        "npm-run-all2": "^6.0.6"
     },
     "devDependencies": {
         "@types/jest": "^29.0.0",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
         "fix-lint": "run-s common-fix-lint backend-fix-lint frontend-fix-lint e2e-fix-lint cli-fix-lint",
         "fix-format": "run-s common-fix-format backend-fix-format frontend-fix-format e2e-fix-format cli-fix-format",
         "test": "run-s common-test warehouses-test cli-test backend-test frontend-test",
-        "dev": "npm-run-all --parallel common-dev warehouses-dev backend-dev frontend-dev",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,7 +92,7 @@
         "fix-lint": "run-s common-fix-lint backend-fix-lint frontend-fix-lint e2e-fix-lint cli-fix-lint",
         "fix-format": "run-s common-fix-format backend-fix-format frontend-fix-format e2e-fix-format cli-fix-format",
         "test": "run-s common-test warehouses-test cli-test backend-test frontend-test",
-        "dev": "npm-run-all --parallel common-dev warehouses-dev backend-dev frontend-dev",
+        "dev": "concurrently -n \"common,warehouses,backend,frontend\" --prefix-colors \"bgWhite.bold,bgGreenBright.bold,bgMagenta.bold,bgBlue.bold\" \"yarn:common-dev\" \"yarn:warehouses-dev\" \"yarn:backend-dev\" \"yarn:frontend-dev\"",
```
ZeRego, line 95: The colors are really nice. 🎉 
Screenshot 2023-07-25 at 16 16 39

ZeRego (approved):
lgtm 🎉

joaoviana:
was able to login in Render and view a chart!

PR #6475, opened 2023-07-25
fix: disable stacking if x axis field isnt numeric

joaoviana:
Closes:


magnew (approved):
Tested, works well. Code looks good!

Issue #6266, opened 2023-07-11
Migrate "Chart config - Reference lines" to Mantine


IrakliJani:
### Description

acceptance criteria:
- [x] use pie charts series/conditional formatting config styles as a reference (they kinda look the same)
- [x] migrate rule container with collapsible element
- [x] migrate all the buttons
- [x] migrate field selector input
- [x] migrate label input
- [x] use new mantine color picker (reuse color picker used in pie chart series and conditional formatting)
- [x] ⚠️ ignore value input (that one is reused from "filters" which we'll tackle as a separate milestone)
  - ![CleanShot 2023-07-11 at 20 43 41@2x](https://github.com/lightdash/lightdash/assets/962095/6ccd8abf-fbd3-4ccd-8bc1-cfd79204ead9)
---

<img width="326" alt="CleanShot 2023-07-11 at 20 42 54@2x" src="https://github.com/lightdash/lightdash/assets/962095/817abc19-f886-4116-9545-b622b4915a15">

PR #6479, opened 2023-07-25
feat: migrate reference line config to mantine

magnew:
Closes: #6266 

### Description:

Migrates reference line config to mantine.


IrakliJani (approved):
@magnew code looks great! approving with one UI suggestion.

In terms of UI and since we have a lot of space here - IMO will be better to use https://mantine.dev/core/color-input/


CleanShot 2023-07-26 at 10 01 56@2x

cc: @PriPatel

magnew:
> @magnew code looks great! approving with one UI suggestion.
> 
> In terms of UI and since we have a lot of space here - IMO will be better to use https://mantine.dev/core/color-input/
> 
> CleanShot 2023-07-26 at 10 01 56@2x
> cc: @PriPatel

Sure! I actually thought about it but opted for doing something more similar to what we already had. I'm happy to do this though.

IrakliJani (approved):
tested in preview, looks great. let's wait for e2e before merging ✅

Issue #6257, opened 2023-07-11
Remove saved_queries_versions joins and use new columns.


rephus:
Remove saved_queries_versions joins and use new columns.

Most models that use `saved_queries_versions` need to be updated as well:

- ResourceViewItemModel
- SpaceModel
- SavedChartModel

Let's create a new ticket for these changes and link to this ticket/PR.

_Originally posted by @ZeRego in https://github.com/lightdash/lightdash/pull/6247#discussion_r1259795822_

PR #6486, opened 2023-07-26
feat: enable new columns chart_kind on saved_chart and remove joins

rephus:
Closes: #6257

### Description:

I'm also adding chart_kind metadata on saved chart creation (I forgot last time :gull_screaming:)

Before:

Screenshot from 2023-07-26 09-59-15

After: 
Screenshot from 2023-07-26 09-59-57

Screenshot from 2023-07-26 10-21-47

This is how a chart looks if the saved chart does not have that metadata

Screenshot from 2023-07-26 10-23-04

Screenshot from 2023-07-26 15-35-25

Validation: 

Screenshot from 2023-07-27 11-06-11

Search:

image


packages/backend/src/models/SpaceModel.ts
```diff
@@ -491,17 +476,14 @@ export class SpaceModel {
             ])
             .orderBy([
                 {
-                    column: `saved_queries_versions.saved_query_id`, // TODO remove
-                    // column: `saved_queries.saved_query_uuid`, // TODO add
+                    column: `saved_queries.saved_query_uuid`,
```
ZeRego, line 479: Not sure if this is a good replacement since this is now ordered alphabetically.
I don't think we need this anymore. Ordering by updated_at should be enough 🤔 what do you think ?

ZeRego (changes requested):
We also need the same fix in `packages/backend/src/models/SpaceModel.ts` and `packages/backend/src/models/ResourceViewItemModel.ts`

ZeRego (approved):

Issue #6154, opened 2023-07-05
Ability to interact with metric values in pie chart


IrakliJani:
### Description

- [x] add ability to copy value
- [x] add ability to see underlying data in pie chart.
- [ ] #6573

![image](https://github.com/lightdash/lightdash/assets/962095/b783f4d9-d575-4d37-ba09-a71c24a60bf1)

PR #6488, opened 2023-07-26
feat: pie charts underlying data

IrakliJani:
related to: https://github.com/lightdash/lightdash/issues/6154

### Description:

adds underlying data to pie charts

CleanShot 2023-07-26 at 13 14 13@2x
CleanShot 2023-07-26 at 13 14 24@2x

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -7,7 +7,6 @@ import {
 } from '@blueprintjs/popover2';
 import { subject } from '@casl/ability';
 import {
-    AdditionalMetric,
     ChartType,
     DashboardChartTile as IDashboardChartTile,
     DashboardFilterRule,
@@ -22,7 +21,6 @@ import {
     getVisibleFields,
     isFilterableField,
     isTableChartConfig,
-    Metric,
     PivotReference,
     ResultValue,
     SavedChart,
@@ -302,7 +300,7 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
         DashboardFilterRule[]
     >([]);
     const [viewUnderlyingDataOptions, setViewUnderlyingDataOptions] = useState<{
-        item: Field | Metric | AdditionalMetric | TableCalculation | undefined;
+        item: Field | TableCalculation | undefined;
         value: ResultValue;
         fieldValues: Record<string, ResultValue>;
         dimensions: string[];
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -7,6 +7,7 @@ import {
 } from '@blueprintjs/popover2';
 import { subject } from '@casl/ability';
 import {
+    AdditionalMetric,
     ChartType,
     DashboardChartTile as IDashboardChartTile,
     DashboardFilterRule,
@@ -21,6 +22,7 @@ import {
     getVisibleFields,
     isFilterableField,
     isTableChartConfig,
+    Metric,
     PivotReference,
     ResultValue,
     SavedChart,
@@ -300,7 +302,7 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
         DashboardFilterRule[]
     >([]);
     const [viewUnderlyingDataOptions, setViewUnderlyingDataOptions] = useState<{
-        item: Field | TableCalculation | undefined;
+        item: Field | Metric | AdditionalMetric | TableCalculation | undefined;
         value: ResultValue;
         fieldValues: Record<string, ResultValue>;
         dimensions: string[];
--- a/packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
+++ b/packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
@@ -12,6 +12,7 @@ import { FC } from 'react';
 import { useParams } from 'react-router-dom';
 import useDashboardFiltersForExplore from '../../hooks/dashboard/useDashboardFiltersForExplore';
 import useToaster from '../../hooks/toaster/useToaster';
+import { useProject } from '../../hooks/useProject';
 import { useApp } from '../../providers/AppProvider';
 import { useTracking } from '../../providers/TrackingProvider';
 import { EventName } from '../../types/Events';
@@ -42,6 +43,7 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
 }) => {
     const { projectUuid } = useParams<{ projectUuid: string }>();
     const { user } = useApp();
+    const { data: project } = useProject(projectUuid);
     const { explore, pieChartConfig } = useVisualizationContext();
     const dashboardFiltersThatApplyToChart = useDashboardFiltersForExplore(
         tileUuid,
@@ -53,7 +55,7 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
     const tracking = useTracking(true);
     const metricQueryData = useMetricQueryDataContext(true);
 
-    if (!value || !tracking || !metricQueryData) {
+    if (!value || !tracking || !metricQueryData || !project) {
         return null;
     }
 
@@ -64,8 +66,8 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
     const canViewUnderlyingData = user.data?.ability?.can(
         'view',
         subject('UnderlyingData', {
-            organizationUuid: user.data?.organizationUuid,
-            projectUuid: projectUuid,
+            organizationUuid: project?.organizationUuid,
+            projectUuid: project?.projectUuid,
         }),
     );
 
@@ -75,8 +77,8 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
         user.data?.ability?.can(
             'manage',
             subject('Explore', {
-                organizationUuid: user.data?.organizationUuid,
-                projectUuid: projectUuid,
+                organizationUuid: project?.organizationUuid,
+                projectUuid: project?.projectUuid,
             }),
         );
 
--- a/packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
+++ b/packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
@@ -5,7 +5,7 @@ import {
     ResultRow,
     ResultValue,
 } from '@lightdash/common';
-import { Box, Menu, MenuProps, Portal, Text } from '@mantine/core';
+import { Box, Menu, MenuProps, Portal } from '@mantine/core';
 import { useClipboard } from '@mantine/hooks';
 import { IconArrowBarToDown, IconCopy, IconStack } from '@tabler/icons-react';
 import { FC } from 'react';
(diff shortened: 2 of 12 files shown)


packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
```diff
@@ -16,40 +24,42 @@ export type PieChartContextMenuProps = {
...
-            'view',
-            subject('UnderlyingData', {
-                organizationUuid: user.data?.organizationUuid,
-                projectUuid: projectUuid,
-            }),
-        );
+    const canViewUnderlyingData = user.data?.ability?.can(
+        'view',
+        subject('UnderlyingData', {
+            organizationUuid: user.data?.organizationUuid,
```
magnew, line 59: I think this should be more like 

`organizationUuid: metricQueryData.organiztionUuid` 

I don't know if that metricQueryData object has uuid, but I think you need to pass the orgUuid of the object in question and it gets compared to the users orgUuid. So passing the user's own uuid is a no-op.

I'm admittedly new to this thing though, so if I am not understanding something feel free to disagree 😄

owlas (changes requested):
This doesn't apply dashboard filters on drill down

CleanShot 2023-07-26 at 10 54 53

IrakliJani:
hey @owlas I chunked up PR's as it was too big to review on one go. I have drilldown PR and Dashboard filters as a follow up. but need to ship this first

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
IrakliJani: @magnew in this case we need user's organization uuid to check if user has an ability to use explore features. metricQueryData does not have organizationUuid

owlas:
I mean that the "view underlying" behaviour isn't correct. It doesn't include the applied dashboard filters. 

# view underlying on bar chart

view underlying on "shipped" gives 4 results

CleanShot 2023-07-26 at 11 09 25@2x

CleanShot 2023-07-26 at 11 09 55@2x

# view underlying on pie chart

view underlying on "shipped" gives 17 results but should be 4

CleanShot 2023-07-26 at 11 08 50@2x

CleanShot 2023-07-26 at 11 09 36@2x

IrakliJani:
@owlas im aware of the limitation, I am addressing dashboard filters in a follow up PR.

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
magnew: I find the CASL pattern a bit confusing, but in a lot of places we have something like this:

```
const hasAdminAccess = user.ability.can(
        'manage',
        subject('Project', {
            organizationUuid: space.organizationUuid,
            projectUuid: space.projectUuid,
        }),
    );
```
Passing the uuid of the object in question. My understanding is that that uuid gets compared to the user uuid to see if they have the access to it. 

A fair amount of our frontend code does what you are doing (passed the users uuid to `subject`), but almost all of the backend code passes the uuid of the object in question. I assume we don't want totally different patterns for this on FE and BE, so one must be righter 😄.

I apologize if I am just reading something wrong.

owlas: That's right it should ideally be something on the object that we're testing. For underlying data, it's consistent across the project. It is correct to pass in the projectUuid and organizationUuid of the _project_ and not of the user:

```typescript
{ organizationUuid, projectUuid } = useProject(...);
```

^ these are the correct values to pass into the check rather than using the user.

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
```diff
@@ -1,55 +1,73 @@
...
+    groupDimensions,
+    rows,
     opened,
     onOpen,
     onClose,
 }) => {
+    const { projectUuid } = useParams<{ projectUuid: string }>();
     const { user } = useApp();
+    const { explore, pieChartConfig } = useVisualizationContext();
+    const dashboardFiltersThatApplyToChart = useDashboardFiltersForExplore(
```
owlas, line 46: looks tidy!

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
```diff
@@ -72,20 +90,41 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
...
+        if (!pieChartConfig.selectedMetric) return;
+
+        openUnderlyingDataModal({
+            item: pieChartConfig.selectedMetric,
+            value,
+            fieldValues: {},
+            dashboardFilters: dashboardFiltersThatApplyToChart,
+            pivotReference: {
+                field: getItemId(pieChartConfig.selectedMetric),
+                pivotValues: groupDimensions
```
owlas, line 102: Is the `pivotReference` here just the dimension that splits the pie chart up? I'm not sure I completely follow the logic here. Is this something that we can just get from `pieChartConfig` similarly to how you do `pieChartConfig.selectedMetric` ?

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -300,7 +302,7 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
         DashboardFilterRule[]
     >([]);
     const [viewUnderlyingDataOptions, setViewUnderlyingDataOptions] = useState<{
-        item: Field | TableCalculation | undefined;
+        item: Field | Metric | AdditionalMetric | TableCalculation | undefined;
```
owlas, line 305: Changing this `item` interface in many places makes me nervous about the impact of this PR. Why do we need to add this? What's different about pie charts that it cannot use the existing types?

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
```diff
@@ -6,21 +6,23 @@ import {
...
 
-const useDashboardFiltersForExplore = (
-    tileUuid: string,
+function useDashboardFiltersForExplore(
+    tileUuid: string | undefined,
     explore: Explore | undefined,
-): DashboardFilters => {
-    const { dashboardFilters, dashboardTemporaryFilters } =
-        useDashboardContext();
+    failSilently: boolean = false,
```
owlas, line 12: This seems like a red flag? What's the purpose of this? In what scenario do you expect this to fail?

Putting this here makes me think we have used the wrong abstraction somewhere upstream of this function

owlas (changes requested):
see inline

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
```diff
@@ -6,21 +6,23 @@ import {
 import { useCallback, useMemo } from 'react';
 import { useDashboardContext } from '../../providers/DashboardProvider';
 
-const useDashboardFiltersForExplore = (
-    tileUuid: string,
+function useDashboardFiltersForExplore(
```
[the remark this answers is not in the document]
IrakliJani: useless change, will revert

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
IrakliJani: @owlas good question, right now when we use a context that is not direct child of a context provider is throwing an error. we also have `/minimal/` pages for dashboards and charts that do not use those providers like `TrackingProvider` or `DashboardProvider`. that does not mean that chart should not render or throw an error.

I don't want to "prop drill" all the way down to the menu component and calling a react hook conditionally is a bad practice and causes bugs.

this seems like more explicit solution to it

so for `minimal` page this will just disable the menu rather than throw an error and it is not affecting any other places in the codebase

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -344,12 +344,16 @@ export const DashboardProvider: React.FC = ({ children }) => {
     return <Context.Provider value={value}>{children}</Context.Provider>;
 };
 
-export const useDashboardContext = (): DashboardContext => {
+export function useDashboardContext<S extends boolean = false>(
+    failSilently?: S,
```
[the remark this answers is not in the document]
IrakliJani: correct. see more detailed answer above https://github.com/lightdash/lightdash/pull/6488#discussion_r1275158054

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
IrakliJani: open to suggestions if you have something better in mind that will be much safer alternative. Ideally this should be handled in the hook itself rather than outside of it. but we have a convention to make every hook "throw" an error if provider is not in the parent tree. that's why I decided to have an option to avoid throwing. maybe a naming change can improve this or another hook that does not throw an error?

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
IrakliJani: we are viewing underlying data of a metric and metric can be AdditionalMetric.

They have almost the same type interface but the key difference is that Metric extends Field but AdditionalMetric does not.

most likely somewhere we have a missing type of AdditionalMetric but in this case here I believe changing types makes sense

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
IrakliJani: updated, thanks both!

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
owlas: The way we've previously done this is with `convertAdditionalMetric` 

for example in the bignumberconfig: https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/hooks/useBigNumberConfig.ts#L95-L95

Might not be the best way but it's consistent, and if it's consistent it's easier to refactor later

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
owlas: There's a couple of ways, the common pattern we've used is using a different context menu when on a dashboard: https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/components/SimpleTable/index.tsx#L111-L120

However it could also be done by using component composition, providing context options to the component or optional dashboard options to the component, and these can be passed in further up the call stack

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
IrakliJani: updated!

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
IrakliJani: updated! (created separate component to handle that)

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
```diff
@@ -72,20 +86,42 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
...
-        //     },
-        // });
+        if (!pieChartConfig.selectedMetric) return;
+
+        openUnderlyingDataModal({
+            item: pieChartConfig.selectedMetric,
+            value,
+            fieldValues: {},
+            dashboardFilters,
+            pivotReference: {
```
owlas, line 96: Try using `fieldValues` instead of `pivotReference` I don't think it's needed here

owlas (changes requested):

packages/frontend/src/components/SimplePieChart/PieChartDashboardContextMenu.tsx
```diff
@@ -0,0 +1,31 @@
+import { FC } from 'react';
```
owlas, line 1: tidy!

owlas (approved):

Issue #6413, opened 2023-07-19
`sql_on` can reference user attributes


owlas:
### Description

I want to be able to reference user attributes in `sql_on`:

```yaml
models:
  - name: users
    meta:
      joins:
        - join: events
           sql_on: ${users.user_id} = ${events.user_id} AND ${users.email} = ${lightdash.user.email}
```

PR #6498, opened 2023-07-26
feat: sql on reference user attribute

rephus:
Closes: #6413

### Description:

Screenshot from 2023-07-26 15-55-38

If attribute does not exist 
Screenshot from 2023-07-26 15-55-43

If attribute exists: 

Screenshot from 2023-07-26 16-29-04


owlas:
testing the preview now

owlas:
Hey @rephus one thing I just bumped into we spoke about on the original spec, if I put this in a sql_on:

```
   - name: orders
     description: >-
       This table has basic information about orders, as well as some derived
       facts based on payments
     meta:
       joins:
         - join: customers
           sql_on: ${customers.customer_id} = ${orders.customer_id} AND ${customer_id} = ${lightdash.attributes.jaffle_cx_id}
```

So here I wanted to restrict orders based on the customer id. But if I query orders, it doesn't limit unless i choose a field from the customers table. that's because we conditionally use the join logic depending on if we reference the table.

I think this is ok and we should probably have `required` field on the join as a separate ticket:

```
joins:
  - join: table
    sql_on: ...
    required: true
```

and we'll always join the table.

What do you think?

rephus:
> But if I query orders, it doesn't limit unless i choose a field from the customers table. 

Is that not the point of the join?, if you want to always filter, then you should be using the `sql_filter` we already implemented.

owlas:
Agreed!

owlas (approved):
Tested
-  missing attribute
- attribute exists but not assigned to user
- attribute exists and does filtering

Issue #6497, opened 2023-07-26
Allow dashboard filters to be disabled (backend only)


owlas:
### Description

Backend implementation before we update the UI.

Suggest we add a `disabled` attribute to all filters, which would effectively render `true` in the sql:

Current filters: https://github.com/lightdash/lightdash/blob/ce17a361c639cf25e9c9e29350755838501d6d16/packages/common/src/types/filter.ts#L57-L65

Suggested change is to have:

```
type DashboardFilterRule = {
   ...
   disabled?: boolaen
}
```

When we render the sql:

```
WHERE `${filter.disabled ? 'true' : renderFilter(filter)}`
```

PR #6500, opened 2023-07-26
feat: allow filters to be disabled

joaoviana:
Closes: #6497 

### Description:

Allow for `FilterRule`s to be `disabled` by adding an optional prop `disabled`
This will allow users to interact with Dashboard Filters in this way: 
* They add a filter to the dashboard
* They don't know what value they want to filter by, _yet_

As proposed on the ticket, a `disabled`+clause that does no filtering (`1=1`) is being introduced here. However, not sure if there could be any issues with this approach for the current filters approach or if it should be done differently.


packages/common/src/types/filter.ts
```diff
@@ -62,6 +62,7 @@ export type DashboardFilterRule<
 > = FilterRule<O, T, V, S> & {
     tileTargets?: Record<string, DashboardFieldTarget>;
     label: undefined | string;
+    disabled?: boolean;
```
owlas, line 65: let's move to `filterRule`

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -293,14 +294,18 @@ const renderBooleanFilterSql = (
...
+    filterRule: FilterRule<FilterOperator, unknown> | DashboardFilterRule,
     field: CompiledField,
     fieldQuoteChar: string,
     stringQuoteChar: string,
     escapeStringQuoteChar: string,
     startOfWeek: WeekDay | null | undefined,
     adapterType: SupportedDbtAdapter,
 ): string => {
+    if ('disabled' in filterRule && filterRule.disabled) {
+        return `WHERE 1=1`;
```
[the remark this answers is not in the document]
joaoviana: Got it! removed the `WHERE` and added a comment 👍

packages/common/src/types/filter.ts
joaoviana: done 👍

packages/common/src/compiler/filtersCompiler.ts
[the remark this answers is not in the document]
IrakliJani: should we be doing this or just returning undefined and then filtering undefined values when generating a final WHERE statement?

the question is: do we want disabled filters to show as 1=1 in the final query?

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -301,6 +301,10 @@ export const renderFilterRuleSql = (
     startOfWeek: WeekDay | null | undefined,
     adapterType: SupportedDbtAdapter,
 ): string => {
+    if ('disabled' in filterRule && filterRule.disabled) {
```
owlas, line 304:
```suggestion
    if (!!filterRule.disabled) {
```

joaoviana: thank you 🙏 
Removed `!!` as ESLint told me it's redundant!

rephus (approved):
This looks good to me . I don't think this should break in any warehouse, right ? Recently @owlas tested something similar using `True` in all warehouses, not sure which one is better.

ZeRego:
I think this might cause unexpected results because the selected fields and joined tables are still being affected by these disabled filters.
Check all the logic where we use `filters` in `buildQuery()` in `queryBuilder.ts`

ZeRego (changes requested):
(details in comment)

joaoviana:
putting in draft to address review comments + test this further

packages/backend/src/queryBuilder.test.ts
```diff
@@ -133,6 +149,16 @@ describe('Query builder', () => {
...
+    test('Should build query with metric filter (where filter is disabled) and metric references a dimension from a joined table', () => {
+        expect(
+            buildQuery({
+                explore: EXPLORE,
+                compiledMetricQuery: METRIC_QUERY_WITH_METRIC_DISABLED_FILTER,
+                warehouseClient: warehouseClientMock,
+            }).query,
+        ).toStrictEqual(METRIC_QUERY_WITH_METRIC_FILTER_DISABLED_SQL);
+    });
+
```
joaoviana, line 161: @ZeRego could this match the scenario you were describing with metric filters + joined tables?

ZeRego (approved):
lgtm ⭐  thanks for the unit tests 😍

Issue #6422, opened 2023-07-20
Removing a column from the results table with a table chart type causes the column to jump to the far-right of the table


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/676031f0caaf4d13bb3e4df7845912b4

Krishna2323:
hey @TuringLovesDeathMetal, can I work on this?

joaoviana:
hey @Krishna2323, go for it!

joaoviana:
hey again @Krishna2323 , just started working on it, hope it's ok!

Krishna2323:
@joaoviana, ok sure, will pick up any other issue in few days, I'm busy with some freelance work.

PR #6507, opened 2023-07-27
fix: do not add column to columns if not part of columnOrder

joaoviana:
Closes: #6422

### Description:

Ensure we do not add column X if it's not referenced in `columnOrder`

Screen recording:

https://github.com/lightdash/lightdash/assets/7611706/b21c5c9f-f4d2-4912-ba8f-a8e3d347f4b5


magnew (approved):
lgtm! Do we need to think about this with other visualization types as well?

joaoviana:
> lgtm! Do we need to think about this with other visualization types as well?

nope! this hook is only used in Table visualisations (as well as the util) 👍

Issue #6265, opened 2023-07-11
Migrate "Chart Config - Display Tab" to Mantine


IrakliJani:
### Description

~~🚫 blocked by: #6264~~

acceptance criteria:
- [x] migrate show legend input
- [x] scroll input
- [x] position inputs
- [x] orientation input

ignore:
- reference lines

![CleanShot 2023-07-11 at 20 37 43@2x](https://github.com/lightdash/lightdash/assets/962095/8a9a7132-7547-43d3-847e-e7a37d2899d0)

PR #6513, opened 2023-07-27
feat: migrate legend config to Mantine

magnew:
Closes: #6265 

### Description:

Migrates the chart legend config to Mantine. This is the remaining work to migrate the display tab.

**Note** that this also removes use of `ReactHookForm/Form` from this component. It didn't look like we were getting much out of the Form and I think the code is quite a lot simpler now. But if there is some reason to keep the form around that I am not seeing, I am open to a different approach.


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/index.tsx
```diff
@@ -42,88 +42,87 @@ const LegendPanel: FC<Props> = ({ items }) => {
...
                 {[
                     [Positions.Left, Positions.Right],
                     [Positions.Top, Positions.Bottom],
                 ].map((positionGroup) => (
-                    <SectionRow key={positionGroup.join(',')}>
+                    <SimpleGrid
+                        cols={2}
+                        mt="xs"
+                        ml="xs"
+                        mb="md"
```
IrakliJani, line 90: I feel like at least `ml` is not necessary here
CleanShot 2023-07-27 at 18 09 47@2x

IrakliJani (changes requested):
left one comment related to alignment and there's this - two legend switchers
CleanShot 2023-07-27 at 18 11 38@2x

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/index.tsx
IrakliJani: sorry, I thought it was scrolling legend.

IrakliJani: so at least labels are aligned to scan text easily

magnew: 1. Good point about the Show legend/Scroll legend switches. It is a little weird that one is a switch and one is a check box. Should I just make them both one or the other? I changed the name of "Scroll" to "Scroll legend" because I though it was clearer, but if you think it would make more sense I could change it back. Or the scroll one could get moved. They have a little different scope -- one decides if you want a legend at all and the other says what kind -- giving them the same visual weight right next to each other is a little odd. But maybe I'm overthinking it 😄.

2. I put the left margin there on purpose cause I think it makes it more clear that all of those boxes are sub-elements of Position. It looked a little funny without it. I still think it looks clearer, but if you don't agree, no problem.

PriPatel: I like the spacing to make the sub-elements more obvious and I agree that switches are the best way :)

Issue #5741, opened 2023-06-02
Control access to fields based on user attributes (aka column-level-security)


owlas:
### Description

In order to hide sensitive data from certain users, we want to be able to control access to certain fields in the model. We can do this based on the attributes that the user has:

```
columns:
  - name: secret_salary
     meta:
       dimension:
          required_attributes:
            can_view_financial_data: "true"
            team: "product"
```

### Restricting access

If the user doesn't have all the required_attributes then they won't see the field in the sidebar when exploring and any query involving that field should throw a forbidden error. 

Note the user could get access if they have sql runner access or we introduce custom dimensions. 

### Analytics events

-  add this to the `project_compile` event. E.g. `column_access_filters_count`

### Documentation

- add column-based security access to the dimension reference docs: https://docs.lightdash.com/references/dimensions
- add `column access filters` to the "where can attributes be used?" in the user attributes docs: https://github.com/lightdash/lightdash/issues/6246

PR #6516, opened 2023-07-27
feat: column level security

rephus:
Closes: #5741

### Description:

Screenshot from 2023-07-27 16-06-29

Screenshot from 2023-07-27 16-06-21

Filter `customer_lifetime_value` on explore 

Screenshot from 2023-07-31 11-52-50

image

### ACceptance criteria

- [x] If the user doesn't have all the required_attributes then they won't see the field in the sidebar when exploring and any query involving that field should throw a forbidden error.

- [x] Analytics events:   add this to the project_compile event. E.g. column_access_filters_count

- [x]  Documentation

    add column-based security access to the dimension reference docs: https://docs.lightdash.com/references/dimensions
    add column access filters to the "where can attributes be used?" in the user attributes docs: 

- [x] Exclude from global search


packages/backend/src/queryBuilder.ts
```diff
@@ -118,6 +130,17 @@ export const buildQuery = ({
     const dimensionSelects = dimensions.map((field) => {
         const alias = field;
         const dimension = getDimensionFromId(field, explore);
+        // Throw error if user does not have the right requiredAttribute for this dimension
+        if (dimension.requiredAttributes)
```
ZeRego, line 134: Users can bypass this by using the dimension in metric or table calculations.

ZeRego (changes requested):

packages/backend/src/queryBuilder.ts
rephus: You can't use TC without having dimensions on the query 
Screenshot from 2023-07-31 10-32-05


You can use custom metrics, but that is already included 
image

Good comment though, I had to test this to make sure it works.

rephus: I guess this is not a problem for viewers, 
developers and admins can already have access through sql runner. 
So this might only affect editors

@owlas should we limit those too ?

rephus: You can't get a TC without referencing the dimension first 



But it is a problem with custom metrics :eyes:

rephus: Fixed

Screenshot from 2023-07-31 11-12-17


Screenshot from 2023-07-31 11-12-54

owlas: @rephus and @ZeRego could we add the auth check to the `getDimensionById` function which is always used to convert a reference into a dimension. We already have a "dimension does not exists" error. We can add to that an auth error depending on attributes. I think this would catch most (probably all) use cases

rephus:
Excluding also from global search

With user attribute (false)
Screenshot from 2023-08-01 10-19-50


With user attribute (true) 

Screenshot from 2023-08-01 10-20-11

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1423,6 +1427,18 @@ export class ProjectService {
                         },
                         0,
                     ),
+                    columnAccessFiltersCount: explores.reduce<number>(
+                        (acc, explore) => {
+                            if (!isExploreError(explore)) {
+                                return getDimensions(explore).filter(
+                                    ({ requiredAttributes }) =>
+                                        requiredAttributes !== undefined,
+                                ).length;
```
ZeRego, line 1436: Should it be `acc + (...).length`?

packages/backend/src/queryBuilder.ts
```diff
@@ -130,6 +159,18 @@ export const buildQuery = ({
         return `  ${metric.compiledSql} AS ${fieldQuoteChar}${alias}${fieldQuoteChar}`;
     });
 
+    if (additionalMetrics)
+        additionalMetrics.forEach((metric) => {
+            if (!metrics.includes(metric.name)) return;
+
+            const dimensionId = getCustomMetricDimensionId(metric);
+            const dimension = getDimensionFromId(dimensionId, explore);
+            assertValidDimensionRequiredAttribute(
```
ZeRego, line 168: Would be good to add unit tests for these errors.

packages/backend/src/queryBuilder.ts
```diff
@@ -118,6 +141,12 @@ export const buildQuery = ({
     const dimensionSelects = dimensions.map((field) => {
         const alias = field;
         const dimension = getDimensionFromId(field, explore);
+
+        assertValidDimensionRequiredAttribute(
```
ZeRego, line 145: Do we need to check filters as well? If I know the dimension, I can get valuable information using filters.  
If we did `filterDimensionsFromExplore(explore, userAttributes)` at the top, we would cover all direct references ( filters, select, order). It still won't cover custom sql in metrics and table calculations.

rephus:
You can't sort by a field that is not selected: 

image

rephus:
I got this when trying to make a request with a private dimension as filter

image

rephus:
> I got this when trying to make a request with a private dimension as filter

Ahh, this is because buildQuery is using the explore from `getExplore` which already filters the private dimensions.... so I think we are safe, in fact the `Forbidden` error seems unreachable now...

rephus:
Restarting e2e tests after rebasing master

ZeRego (approved):

PR #6523, opened 2023-07-28
chore: create scheduler queue size endpoint

rephus:
Closes:  #6527

### Description:

Screenshot from 2023-07-28 12-02-16

### DIFF
--- a/packages/backend/src/controllers/schedulerController.ts
+++ b/packages/backend/src/controllers/schedulerController.ts
@@ -4,7 +4,6 @@ import {
     ApiScheduledJobsResponse,
     ApiSchedulerAndTargetsResponse,
     ApiSchedulerLogsResponse,
-    SchedulerAndTargets,
 } from '@lightdash/common';
 import { Delete } from '@tsoa/runtime';
 import express from 'express';
@@ -22,7 +21,6 @@ import {
     SuccessResponse,
     Tags,
 } from 'tsoa';
-import { SchedulerService } from '../services/SchedulerService/SchedulerService';
 import { schedulerService } from '../services/services';
 import {
     allowApiKeyAuthentication,
@@ -34,30 +32,6 @@ import {
 @Response<ApiErrorPayload>('default', 'Error')
 @Tags('Schedulers')
 export class SchedulerController extends Controller {
-    /**
-     * Get all scheduler jobs
-     * This can be used to monitor the status of the scheduled jobs
-     */
-    @Middlewares([])
-    @SuccessResponse('200', 'Success')
-    @Get('/')
-    @OperationId('getAllScheduledJobs')
-    async getAllScheduledJobs(): Promise<{
-        status: 'ok';
-        results: {
-            errorSize: number;
-            queueSize: number;
-            errors: any[];
-            queued: any[];
-        };
-    }> {
-        this.setStatus(200);
-        return {
-            status: 'ok',
-            results: await SchedulerService.getAllJobs(),
-        };
-    }
-
     /**
      * Get scheduled logs
      * @param req express request
--- a/packages/backend/src/controllers/schedulerController.ts
+++ b/packages/backend/src/controllers/schedulerController.ts
@@ -4,6 +4,7 @@ import {
     ApiScheduledJobsResponse,
     ApiSchedulerAndTargetsResponse,
     ApiSchedulerLogsResponse,
+    SchedulerAndTargets,
 } from '@lightdash/common';
 import { Delete } from '@tsoa/runtime';
 import express from 'express';
@@ -21,6 +22,7 @@ import {
     SuccessResponse,
     Tags,
 } from 'tsoa';
+import { SchedulerService } from '../services/SchedulerService/SchedulerService';
 import { schedulerService } from '../services/services';
 import {
     allowApiKeyAuthentication,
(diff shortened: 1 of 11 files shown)


packages/backend/src/controllers/schedulerController.ts
```diff
@@ -32,6 +34,30 @@ import {
...
+    @Middlewares([])
+    @SuccessResponse('200', 'Success')
+    @Get('/')
+    @OperationId('getAllScheduledJobs')
+    async getAllScheduledJobs(): Promise<{
+        status: 'ok';
+        results: {
+            errorSize: number;
+            queueSize: number;
+            errors: any[];
```
owlas, line 50: This would expose raw error logs to any unauthenticated client using the API

owlas (changes requested):
Looks like this exposes sensitive information on a public endpoint

packages/backend/src/scheduler.ts
```diff
@@ -1,3 +1,6 @@
+// organize-imports-ignore
+import './otel'; // must be imported first
```
owlas, line 2: I'm concerned that importing this is going to interfere with the SIGINT handler that graphile provides. Usually it will gracefully shutdown queues, but otel.ts has it's own handler

packages/backend/src/schedulerMetrics.ts
```diff
@@ -0,0 +1,25 @@
+import opentelemetry, { ValueType } from '@opentelemetry/api';
```
owlas, line 1: see docs for adding metrics https://opentelemetry.io/docs/instrumentation/js/manual/#metrics

packages/backend/src/schedulerMetrics.ts
```diff
@@ -0,0 +1,25 @@
...
+import { schedulerClient } from './clients/clients';
+import { SchedulerWorker } from './scheduler/SchedulerWorker';
+import { VERSION } from './version';
+
+const meter = opentelemetry.metrics.getMeter('lightdash-worker', VERSION);
+const queueSizeCounter = meter.createObservableUpDownCounter<{
```
owlas, line 7: Creates a metric `queue.jobs.count` with 4 possible series (from 2 boolean labels). Use UpDownCounter, which is a metric that increments / decrements in whole numbers and can be added (i.e. we can sum all customers queues and we have total of all jobs in queues).

packages/backend/src/index.ts
```diff
@@ -274,9 +275,44 @@ export const slackService = new SlackService({
...
+let worker: SchedulerWorker | undefined;
 if (lightdashConfig.scheduler?.enabled) {
-    const worker = new SchedulerWorker({ lightdashConfig });
+    worker = new SchedulerWorker({ lightdashConfig });
     worker.run().catch((e) => {
         Logger.error('Error starting scheduler worker', e);
     });
 }
+
+const onExit = () => {
```
owlas, line 286: Need a single exit handler to gracefull shutdown **both**
- opentelemetry sdk (pushes all metrics in buffer)
- worker (finishes all executing tasks)

packages/backend/src/index.ts
```diff
@@ -274,9 +275,44 @@ export const slackService = new SlackService({
...
+    asyncExit()
+        .catch((e) => {
+            Logger.error('Error stopping server', e);
+        })
+        .finally(() => {
+            process.exit();
+        });
+};
+
+process.on('SIGUSR2', onExit);
```
owlas, line 314: These are all the signals that graphile worker was responding to with a graceful shutdown, so we copy that behaviour

packages/backend/src/otel.ts
```diff
@@ -111,7 +111,8 @@ const sdk = new NodeSDK({
     // Resource detection automatically sets resource attributes.
     autoDetectResources: true,
     resource: new Resource({
-        [SemanticResourceAttributes.SERVICE_NAME]: 'lightdash',
+        [SemanticResourceAttributes.SERVICE_NAME]:
+            process.env.OTEL_SERVICE_NAME || 'lightdash',
```
owlas, line 115: allow the service name to be set from env

packages/backend/src/otel.ts
```diff
@@ -123,11 +124,5 @@ try {
 } catch (e) {
     diag.error('Error starting OpenTelemetry SDK. No telemetry exporting', e);
 }
-process.on('SIGTERM', () => {
```
owlas, line 126: moved to server / scheduler

packages/backend/src/scheduler.ts
```diff
@@ -58,7 +70,7 @@ async function onHealthCheck() {
 }
 
 createTerminus(server, {
-    signal: 'SIGINT',
+    signals: ['SIGUSR2', 'SIGTERM', 'SIGINT', 'SIGHUP', 'SIGABRT'],
```
owlas, line 73: graphile was listening to all these signals 

(in fact ignoring SIGTERM before meant that the worker was never gracefully shutting down in kubernetes, e.g. when we upgrade lightdash cloud, the existing containers were getting SIGTERM and we cancelling existing tasks half way through their jobs)

packages/backend/src/scheduler/SchedulerWorker.ts
```diff
@@ -71,7 +151,7 @@ export class SchedulerWorker {
             connectionString: this.lightdashConfig.database.connectionUri,
             logger: workerLogger,
             concurrency: this.lightdashConfig.scheduler?.concurrency,
-            noHandleSignals: false,
+            noHandleSignals: true,
```
owlas, line 154: server / schedulers must handle shutdown manually

packages/backend/src/scheduler/SchedulerWorker.ts
```diff
@@ -28,6 +32,82 @@ type SchedulerWorkerDependencies = {
...
+const meter = opentelemetry.metrics.getMeter('lightdash-worker', VERSION);
+const tracer = opentelemetry.trace.getTracer('lightdash-worker', VERSION);
+const taskDurationHistogram = meter.createHistogram<{
+    task_name: string;
+    error: boolean;
+}>('worker.task.duration_ms', {
+    description: 'Duration of worker tasks in milliseconds',
+});
+
+const traceTask = (taskName: string, task: Task): Task => {
```
owlas, line 44: adds metrics + traces to tasks

rephus (commented):
Reviewed and tested . It works pretty well (at least from the logs) Well done :raised_hands:  this should solve all our visibility problems. 

Screenshot from 2023-07-31 10-17-41
Screenshot from 2023-07-31 10-18-25
Screenshot from 2023-07-31 10-23-12
Screenshot from 2023-07-31 10-23-42

owlas (approved):

Issue #6336, opened 2023-07-14
Tooltip in the dashboard header persists even after I stop hovering over the button


PriPatel:
<img width="264" alt="Screenshot 2023-07-14 at 13 03 53" src="https://github.com/lightdash/lightdash/assets/12776522/e6600149-3cd9-4d58-9eb2-c71b0b7c4216">

1. Go to a dashboard
2. Open in edit mode
3. (Don't make any changes to the dashboard) Hover over the disabled `Save` button
4. The tooltip appears. 
5. Stop hovering over the button
6. The tooltip persists

PR #6529, opened 2023-07-28
fix: correct hover tooltip ux when no changes to save in dashboard button

joaoviana:
Closes: #6336

### Description:

`Tooltip`s from `blueprint` + `disabled` buttons don't work as expected. 

Instead, wrap the button with `mantine`'s `Tooltip` and wrap the button with a `Box` component so the `ref` is forwarded. 

Screen recording:

https://github.com/lightdash/lightdash/assets/7611706/8a5ff7d7-9df3-4a29-9125-32e856f2adf6


magnew (approved):
Nice!

Issue #3700, opened 2022-11-07
I want to be able to nest and/or filters in the same filter block


TuringLovesDeathMetal:
### Description

Right now, I can use `any` and `all` to switch between joining all of my filters by `AND` or `OR`. But, I can't mix and match and/or within a filter block. 

Here are some examples of filters I can't do right now with our filter structure:

I can do this:
```
Users Count unique user IDs: is greater than 10 
AND 
Users Date created: is before January 1st, 2020
```

But I can't do this:
```
Users Count unique user IDs: is greater than 10 OR is less than 5. 
AND 
Users Date created: is before January 1st, 2020
```

And I can't do this:
```
Users Count unique user IDs: is greater than 0 AND Users Date created: is before January 1st, 2020
OR 
Users Count unique user IDs: is less than 5 AND Users Date created: is before January 1st, 2020
```

Being able to nest and/or filters in the same block is important for analyses where I'm creating "groups" of users using filters. E.g. "I want to compare weekly active users who are under 20 with weekly active users who are over 50". To do this, I need nested filters:

```
is_7d_active: is TRUE 
AND age: is < 20 OR age: is > 50
```

### Acceptance criteria
Here's a sketch: 
<img width="915" alt="Screenshot 2022-11-24 at 10 48 22" src="https://user-images.githubusercontent.com/12776522/203774311-b1611383-e274-4d23-8060-d52a8bcbdf21.png">

- [ ] I'm able to add `groups` and `new rules` to my filters. 
- [ ] All of the groups of conditions are joined with `AND`. 
- [ ] Conditions within groups can be joined with `AND` or `OR` (depending on if the group is using `ALL` or `ANY`)
- [ ] Adding a group adds a nested set of filters with `any` or `all` rules applied. For example, if I add a group here, I would get a nested group within the `all` filters 
<img width="1071" alt="image" src="https://user-images.githubusercontent.com/31848148/220148941-e8a44e73-5ad6-4758-82ed-14fc17947bcb.png">
- [ ] adding a rule to a group means that rule is added to the `any` or `all` set of rules for that group. For example, if I add a rule to the outer `all` group, it would be added like this:
<img width="1191" alt="image" src="https://user-images.githubusercontent.com/31848148/220149767-229c547f-ff8c-4d9b-a69b-b83c169929bd.png">
Adding a group the the inner `any` group would look like this:
[...]

PR #6530, opened 2023-07-28
feat: nested filters

ZeRego:
Closes: #3700 

### Description:

Functionality was already there, we just had to add the option in the UI.

Changes:
- allow for users to convert filter rules into a group ( only 1 level deep )
- refactor button components to Mantine
- small UI changes: lines, button position

note: not worth cleaning up too much the UI/code since we are going to do a complete revamp of the filters. 

Screenshot 2023-07-28 at 15 36 35

Screenshot 2023-07-28 at 15 39 13

https://github.com/lightdash/lightdash/assets/9117144/554d0310-f951-4bc6-8f97-8009f3884491

### DIFF
--- a/packages/backend/src/queryBuilder.mock.ts
+++ b/packages/backend/src/queryBuilder.mock.ts
@@ -590,6 +590,52 @@ export const METRIC_QUERY_WITH_METRIC_DISABLED_FILTER_THAT_REFERENCES_JOINED_TAB
         compiledAdditionalMetrics: [],
     };
 
+export const METRIC_QUERY_WITH_NESTED_METRIC_FILTERS: CompiledMetricQuery = {
+    dimensions: ['table1_dim1'],
+    metrics: ['table1_metric1'],
+    filters: {
+        metrics: {
+            id: 'root',
+            and: [
+                {
+                    id: '1',
+                    target: {
+                        fieldId: 'table1_metric1',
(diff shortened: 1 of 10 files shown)


owlas (commented):
Is this an existing bug? It doesn't look like we're nesting the filters in the metrics section: 

CleanShot 2023-07-31 at 11 36 45@2x

CleanShot 2023-07-31 at 11 37 13@2x

owlas:
✅ works well even when nesting goes beyond 2 levels deep:

CleanShot 2023-07-31 at 11 42 02@2x

ZeRego:
@owlas Added support for nested metric filters. ( unit test included )

Screenshot 2023-07-31 at 15 14 46

owlas (changes requested):
If you "explore from here" with no dashboard filters, it renders a little off. Functionally still works.

Tested everything else and worked perfectly, so last change! 

CleanShot 2023-08-01 at 12 13 25@2x

ZeRego:
@owlas fixed. We now only show the dimension/metric section if there are rules in it.

Explore from here a chart in a dashboard that has dimension filters
Screenshot 2023-08-01 at 13 38 09

Explore from here a chart in a dashboard that has metrics filters
Screenshot 2023-08-01 at 13 38 41

Issue #5359, opened 2023-05-03
I should not be able to create an unnamed dashboard


PriPatel:
Currently, Im not forced to name my dashboard, so I could end up being confused whilst searching through a list of items:

<img width="926" alt="Screenshot 2023-05-03 at 15 38 52" src="https://user-images.githubusercontent.com/12776522/235955049-d4f79776-9bf0-4f39-9797-0a05ce2bf23a.png">

- [ ] The name field should be mandatory, like in the Space Creation modal. 

<img width="522" alt="Screenshot 2023-05-03 at 15 53 09" src="https://user-images.githubusercontent.com/12776522/235954706-8b8eb4d9-267e-44eb-abfd-e253e220e157.png">

PriPatel:
Yes

PR #6533, opened 2023-07-28
fix: no unnamed dashboards or spaces

yasmin-e:
Closes: #5359 

### Description:

https://github.com/lightdash/lightdash/assets/67699259/2f95cb1f-2fdb-4ea3-a845-09a1712109d0

### DIFF
--- a/packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
+++ b/packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
@@ -4,20 +4,19 @@ import {
     DialogBody,
     DialogFooter,
     DialogProps,
-    FormGroup,
-    HTMLSelect,
-    InputGroup,
-    Intent,
 } from '@blueprintjs/core';
 import { Dashboard, Space } from '@lightdash/common';
 import { FC, useCallback, useState } from 'react';
+import { useForm } from 'react-hook-form';
 import { useCreateMutation } from '../../../hooks/dashboard/useDashboard';
 import {
     useCreateMutation as useSpaceCreateMutation,
     useSpaceSummaries,
 } from '../../../hooks/useSpaces';
 import { useApp } from '../../../providers/AppProvider';
-import {} from '../ShareSpaceModal/ShareSpaceModal.style';
+import Form from '../../ReactHookForm/Form';
+import Input from '../../ReactHookForm/Input';
+import Select from '../../ReactHookForm/Select';
 
 interface DashboardCreateModalProps extends DialogProps {
     projectUuid: string;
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
```diff
@@ -4,20 +4,19 @@ import {
...
     DialogFooter,
     DialogProps,
-    FormGroup,
-    HTMLSelect,
-    InputGroup,
-    Intent,
 } from '@blueprintjs/core';
 import { Dashboard, Space } from '@lightdash/common';
 import { FC, useCallback, useState } from 'react';
+import { useForm } from 'react-hook-form';
```
joaoviana, line 10: Wondering the reason for using `react-hook-form` rather than using mantine's own?

packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
```diff
@@ -56,55 +57,46 @@ const DashboardCreateModal: FC<DashboardCreateModalProps> = ({
...
 
-        if (isCreatingNewSpace) {
-            newSpace = await createSpace({
-                name: newSpaceName,
-                isPrivate: false,
-                access: [],
-            });
-        }
+            if (isCreatingNewSpace) {
+                console.log(data.newSpaceName);
```
joaoviana, line 81: lil log here 👁️

joaoviana (commented):
2 qs!

packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
```diff
@@ -118,111 +110,81 @@ const DashboardCreateModal: FC<DashboardCreateModalProps> = ({
...
-                            }
-                        />
-                    </>
-                }
-            />
+                            <Button
+                                minimal
+                                icon="arrow-left"
+                                intent="primary"
+                                onClick={() => setIsCreatingNewSpace(false)}
```
joaoviana, line 163: I think this needs to be type `submit`...

joaoviana (changes requested):
some Qs on properties!

packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
yasmin-e: I just felt like it was out of scope to migrate the modal to Mantine but now I realise I could've totally done it, too late 🤷🏼

packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
```diff
@@ -118,111 +110,81 @@ const DashboardCreateModal: FC<DashboardCreateModalProps> = ({
...
+                title="Create Dashboard"
+                methods={form}
+                onSubmit={handleConfirm}
+            >
+                <DialogBody>
+                    <Input
+                        label="Enter a memorable name for your dashboard"
+                        name="dashboardName"
+                        placeholder="eg. KPI Dashboards"
+                        disabled={isCreatingDashboard}
```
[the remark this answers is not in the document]
yasmin-e: it's not a prop that's available on this comp, required is passing through rules

packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
yasmin-e: it doesn't submit the form though 🤔 it's just to go back to the select instead of input

joaoviana: ignore! 🤦

PriPatel (approved):

packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
```diff
@@ -118,111 +109,81 @@ const DashboardCreateModal: FC<DashboardCreateModalProps> = ({
...
```
joaoviana, line 122: here we should also take whatever `useForm` gives us and

```
 {...form.register('dashboardName')}
```
And remove the `name` prop
What do you think?

joaoviana (approved):
one non-blocking comment, great stuff 🎉

Issue #5652, opened 2023-05-26
Increase max-width of content inside Project Settings


PriPatel:
<img width="1440" alt="Screenshot 2023-05-26 at 19 32 27" src="https://github.com/lightdash/lightdash/assets/12776522/679e5139-af10-4189-bcc8-371d4f768249">

- There is a lot to fit into the scheduled deliveries and validator table. On laptops, the tables in particular look ok, but any larger screen sizes than that, there's a lot of wasted space considering there is too much too fit into the tables in the first place.
- The sidebar is too big in width. It also doesnt need to be resizeable. The minimum size is ok and leaves room for a bigger content area in the middle.

PriPatel:
@IrakliJani this has the `needs spec` label still. Happy to chat about it sometime whilst you're working on the Settings pages. Maybe we can pair and try and get a POC working  otherwise we can remove it from the backlog for this week.

PriPatel:
Reopening to remind myself to make a decision on this!

joaoviana:
Summary of user feedback on resizing dashboards

why some users like to reduce the window size of a dashboard: when multitasking, so one half of the screen is on LD dashboard, the other for something esle

Individual opinions on resizing:
1. Dashboard automatically resizes as the window resizes. The advantage is that it offers flexibility to the end user, allowing them to determine their preferred size. However, a significant downside is that elements may become too small to read when the window is significantly reduced in size.
2. Dashboard crops instead of resizing when the window changes. The benefit here is maintaining the size quality of the dashboard, ensuring that elements do not become too small and unreadable. The drawback is a lack of flexibility, as users cannot adjust the size of the dashboard elements.
3. It employs smart resizing and organizing. This approach likely combines aspects of resizing and reorganizing elements in a user-friendly way to maintain readability and usability regardless of the window size.

jpetey75:
The settings pages have been redesigned since this was filed, most recently with the layout harmonization in [#26104](https://github.com/lightdash/lightdash/pull/26104), and table-heavy pages now get full-width treatment. If a settings page still feels cramped, let us know and we'll reopen.

PR #6534, opened 2023-07-28
fix: settings panel max width

yasmin-e:
Closes: #5652

### DIFF
--- a/packages/frontend/src/components/common/Settings/SettingsCard.tsx
+++ b/packages/frontend/src/components/common/Settings/SettingsCard.tsx
@@ -3,7 +3,7 @@ import { FC } from 'react';
 
 const SettingsCard: FC<PaperProps> = ({ children, ...rest }) => {
     return (
-        <Paper shadow="sm" withBorder p="md" {...rest}>
+        <Paper shadow="sm" withBorder p="md" maw={1200} {...rest}>
             {children}
         </Paper>
     );


packages/frontend/src/components/common/Settings/SettingsCard.tsx
```diff
@@ -3,7 +3,7 @@ import { FC } from 'react';
 
 const SettingsCard: FC<PaperProps> = ({ children, ...rest }) => {
     return (
-        <Paper shadow="sm" withBorder p="md" {...rest}>
+        <Paper shadow="sm" withBorder p="md" maw={1200} {...rest}>
```
ZeRego, line 6: This won't fix it since this is controlled by our generic Page component in `packages/frontend/src/components/common/Page/Page.tsx`.

Our pages can have:
- a fixed width ( using `withFixedContent` ) 
- minimum width possible ( using `withFitContent` )
- use all the width ( no prop )


When applying a fixed with we default to 900px:
```
export const PAGE_CONTENT_WIDTH = 900;
```

Changing this value would impact all the pages with `withFixedContent`, which might be a bit percentage of the app.

@IrakliJani might know more about the impact of changing this default value.

yasmin-e: yep, I talked to him and he said he'd have to make this change himself :) 
I'm going to close this ticket...

Issue #3174, opened 2022-09-06
Update spacing around titles on dashboard markdown tiles. 


PriPatel:
<img width="512" alt="Screenshot 2022-09-06 at 16 20 42" src="https://user-images.githubusercontent.com/12776522/188673321-1bfa57f7-d62f-457c-b43e-ed71b3aca406.png">

There's too much space above! This affects markdown tiles where the user has left the `title` blank.

PriPatel:
<img width="313" alt="Screenshot 2022-09-06 at 17 34 26" src="https://user-images.githubusercontent.com/12776522/188688710-36aacf34-88b7-4960-afa7-2b2b83ec0769.png">

Woops - its not a title, its a heading type

rephus:
@PriPatel that empty space is happening because you didn't set a title. So the title is empty , leaving a white space at the top. 

![image](https://user-images.githubusercontent.com/1983672/206165641-d48cd0f9-a53e-4232-a95e-4d53f02d6159.png)
![image](https://user-images.githubusercontent.com/1983672/206165777-d5aef9f4-626f-48d4-909f-4c4476595f7b.png)

---

![image](https://user-images.githubusercontent.com/1983672/206165870-96a97d59-9bee-43db-aca9-47b986bf2434.png)
![image](https://user-images.githubusercontent.com/1983672/206165895-0ad5c2ce-a8a1-43d1-8414-39edaca11b79.png)



What do you suggest we do in this case ?

PriPatel:
Are we easily able to remove the 'title' (white space)  bit **if** they have left it blank?

rephus:
> If the buttons only appeared on hover, overlaying the title,

Yes !

yasmin-e:
Investigated: Action menu is not overlayed on tile, so removing height when no title is given on `TileBase > HeaderContainer` makes the action menu unclickable

Issue #5773, opened 2023-06-06
The actions/filters button overlays need to be decoupled from the dashboard title


PriPatel:
**Investigated**
- Both title and action menu live inside `HeaderContainer` in `TileBase`
- In the case where a user created a `Markdown tile` with **no title**,  we would like to be able to  remove the height from `HeaderContainer` to avoid the empty space, but by doing so we make the `action menu` unclickable

<img width="610" alt="Screenshot 2023-06-06 at 11 00 37" src="https://github.com/lightdash/lightdash/assets/12776522/ff1e09e9-b22c-4d99-9d40-cabc11ca899e">

This is causing issue : https://github.com/lightdash/lightdash/issues/3174

joaoviana:
done here: https://github.com/lightdash/lightdash/pull/6535

PR #6535, opened 2023-07-28
fix: check if markdown tile title is empty and reduce height

joaoviana:
Closes: #3174 #5773

### Description:

If Dashboard title `MARKDOWN` has an empty `title`(`''`), then remove the height reserved for the header, but still make sure that the 3-dot-vertical icon actions are styled and working correctly. 

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/1c0a53c3-e4ff-4ad7-b4a5-82ee47352edd

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileBase/TileBase.styles.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileBase/TileBase.styles.tsx
@@ -4,6 +4,7 @@ import styled, { createGlobalStyle, css } from 'styled-components';
 interface HeaderContainerProps {
     $isEditMode: boolean;
     $isHovering?: boolean;
+    $isEmpty?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


yasmin-e (approved):
WOW! Finally 🎉 
image

packages/frontend/src/components/DashboardTiles/TileBase/TileBase.styles.tsx
```diff
@@ -55,6 +64,7 @@ export const GlobalTileStyles = createGlobalStyle`
 
 interface TileTitleProps {
     $hovered?: boolean;
+    $test?: boolean;
```
joaoviana, line 67: oops, removing this once all tests pass and then just merge 👍

joaoviana: done

Issue #6538, opened 2023-07-28
`Edit tile content` form fields should be populated


yasmin-e:
### Description
A chunk of #6431 
- [ ] The current chart should always be shown in the chart selection dropdown
- [ ]  If I have not updated the title, if I swap the chart, the placeholder text in the title field should update
- [ ]  If I update the title, the next time I open the modal, I should see the title instead of the placeholder

<img width="529" alt="Screenshot 2023-07-28 at 18 24 37" src="https://github.com/lightdash/lightdash/assets/67699259/58a90aab-278a-4db3-ad0d-6c338ce22ffd">

Issue #6431, opened 2023-07-20
Updates to the 'edit tile content' modal for charts 


PriPatel:
⛔️ Needs a kick-off session ⛔️
### Walkthrough:
https://www.figma.com/proto/btY2rdBe5I2W4xLR3upmQG/Untitled?type=design&node-id=2029-189139&t=RUrZG8yVbeMFKBVX-0&scaling=min-zoom&page-id=2029%3A189138&starting-point-node-id=2029%3A189139
- [ ] Swap the fields around (so chart selection should show first, tile renaming should come second)
- [ ] Add helper text for each field
- [ ] The current chart should always be shown in the chart selection dropdown
- [ ] CTA should change to `Save Changes`
- [ ] CTA should be disabled until the user actually makes changes
- [ ] If I have not updated the title, if I swap the chart, the placeholder text in the title field should update
- [ ] If I update the title, the next time I open the modal, I shouldn't see placeholder text in that field

### When the original chart has been deleted:
https://www.figma.com/proto/btY2rdBe5I2W4xLR3upmQG/Untitled?type=design&node-id=2029-189320&t=RUrZG8yVbeMFKBVX-0&scaling=min-zoom&page-id=2029%3A189138&starting-point-node-id=2029%3A189139
- [ ] The 'chart' field show say that there is `No chart selected`

### When the chart has been saved from the dashboard (after this milestone has been completed)
- [ ] The user shouldn't be able to swap the chart, only rename the title

https://www.figma.com/proto/btY2rdBe5I2W4xLR3upmQG/Untitled?type=design&node-id=2029-189389&t=RUrZG8yVbeMFKBVX-0&scaling=min-zoom&page-id=2029%3A189138&starting-point-node-id=2029%3A189139

PR #6539, opened 2023-07-28
fix: edit tile content fields are empty

yasmin-e:
Closes: #6538 
A chunk of #6431 

### Description:
- [ ] The current chart should always be shown in the chart selection dropdown
- [ ]  If I have not updated the title, if I swap the chart, the placeholder text in the title field should update
- [ ]  If I update the title, the next time I open the modal, I should see the title instead of the placeholder

before
Screenshot 2023-07-28 at 18 24 37

after

https://github.com/lightdash/lightdash/assets/67699259/e08bcd5f-3c10-454b-bbe0-4125a45e4bdf

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileForms/ChartUpdateModal.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileForms/ChartUpdateModal.tsx
@@ -1,3 +1,4 @@
+import { DashboardChartTile } from '@lightdash/common';
 import {
     Button,
     Flex,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/DashboardTiles/TileForms/ChartUpdateModal.tsx
```diff
@@ -17,34 +18,32 @@ import { useChartSummaries } from '../../../hooks/useChartSummaries';
...
     const { data: savedCharts, isLoading } = useChartSummaries(projectUuid);
 
-    const handleConfirm = form.onSubmit(
-        ({ title: newTitle, uuid: newChartUuid }) => {
-            onConfirm?.(newTitle, newChartUuid);
-        },
+    const [chartUuid, setChartUuid] = React.useState<string>(
+        tile.properties.savedChartUuid ?? '',
+    );
+    const [chartTitle, setChartTitle] = React.useState<string>(
```
ZeRego, line 39: Why can't we track these input values in the useForm?

yasmin-e: we need the input fields to be controlled so we can populate them

ZeRego (approved):

PR #6574, opened 2023-08-01
feat: dashboard filters default value

IrakliJani:
### Description:

**changes in popover:**
CleanShot 2023-08-01 at 16 25 10@2x

---

**default is disabled:**
CleanShot 2023-08-01 at 16 31 44@2x

---

**changes in active dashboard filters:**
CleanShot 2023-08-01 at 16 25 20@2x

---

**changes in dashboard chart tile:**
CleanShot 2023-08-01 at 16 25 38@2x


IrakliJani:
bug: explore from here should ignore rules with disabled values

CleanShot 2023-08-01 at 17 25 16@2x

joaoviana (approved):

Issue #5979, opened 2023-06-20
Supporting multiple versions of DBT 


rephus:
### Description

This PR will support multiple versions of DBT in the server https://github.com/lightdash/lightdash/pull/5936

This is how it currently works: 

* we install multiple dbt versions using venv on Dockerfile
* dbt (default) is set to 1.4
* If compile/deps fails on one version, we will try 1.5 (using linked binary `dbt1.5` created on Dockerfile)
* If that fails, we throw an error

On some parts of the code where we need to validate the yaml, we need to load a different manifest for different versions. 
dbt 1.4 uses manifest v7 , and dbt 1.5 uses manifest v9. 

### New proposal

we need more dbt versions (maybe because we allow the user to select his own specific version) , 

we will have to download the manifest beforehand (and run some regex to identify the version and its manifest). Specially if the attributes in dbt have changed. 

Install ~=1.4.0 (dbt) , ~=1.5.0 (dbt1.5) on Dockerfile. 

Add a new parameter on project settings `dbtVersion` (string | null ) 

<img width="792" alt="image" src="https://github.com/lightdash/lightdash/assets/1983672/768aed38-cf28-4050-a5bf-55648c4ceb1b">

If undefined, we'll use the default dbt version (~1.4.0 currently). Otherwise we will use only the version specified by the customer. 

## Acceptance criteria 

- [ ] Add new `dbtVersion: string | null` attribute to projects db (defaults to null) 
- [ ] Update project settings page to allow customers select their dbt version
- [ ] Use that version during compile 
- [ ] Remove fallback behaviour

## Questions

- Do we want to use the  current dbt version behaviour (we try 1.4.6 or 1.5.2 if that fails)  if dbtVersion==null ? or we only want to use 1 version at a time. 
- Should we default to 1.4.6 or 1.5.2 ?

## out of scope

- We will only support the latest major dbt version, we don't support minor versions
- If we eventually want to support minor versions, we could install them on the fly
- This will easily allow support for  1.6 in the future, but that will be done in a separate PR
- This should not affect the CLI

rephus:
>   Do we want to use the current dbt version behaviour (we try 1.4.6 or 1.5.2 if that fails) if dbtVersion==null ? or we only want to use 1 version at a time.

I think we should remove this "fallback" behaviour completely, that would throw some errors for customers using 1.5.2 already, but the fix is easy, they just need to pin their version

PR #6584, opened 2023-08-02
feat: support multiple dbt

rephus:
Closes: #5979

### Description:

[Screencast from 02-08-23 17:43:40.webm](https://github.com/lightdash/lightdash/assets/1983672/a31722f9-5e3c-416d-a9d0-37945515be0c)

image

Using default or 1.4

image

Using 1.5 (not compatible) 

image

Screenshot from 2023-08-02 10-15-41

Run tests in loop 

image

On CLI 

image


packages/frontend/src/components/ProjectConnection/index.tsx
```diff
@@ -226,6 +224,7 @@ export const UpdateProjectConnection: FC<{
                 name,
                 dbtConnection,
                 warehouseConnection,
+                dbtVersion: dbtVersion ? dbtVersion : undefined,
```
ZeRego, line 227: isn't this the same as`dbtVersion,` ?

packages/frontend/src/components/ProjectConnection/index.tsx
```diff
@@ -208,6 +202,9 @@ export const UpdateProjectConnection: FC<{
                 name: data.name,
                 dbt: data.dbtConnection,
                 warehouse: data.warehouseConnection,
+                dbtVersion: data?.dbtVersion
+                    ? (data?.dbtVersion as SupportedDbtVersions)
+                    : undefined,
```
ZeRego, line 207: isn't this the same as `dbtVersion: data?.dbtVersion` ?

packages/frontend/src/components/ProjectConnection/index.tsx
```diff
@@ -199,6 +190,9 @@ export const UpdateProjectConnection: FC<{
             warehouse: {
                 ...data?.warehouseConnection,
             },
+            dbtVersion: data?.dbtVersion
+                ? (data?.dbtVersion as SupportedDbtVersions)
```
ZeRego, line 194: why do we need to force the type here? isn't it recognizing the SupportedDbtVersions from the project type ?

packages/backend/src/types.ts
```diff
@@ -21,14 +22,17 @@ export interface ProjectAdapter {
...
+    getDbtManifest(
+        dbtVersion: SupportedDbtVersions,
+    ): Promise<DbtRpcGetManifestResults>;
+    getDbtCatalog(
+        dbtVersion: SupportedDbtVersions,
+    ): Promise<DbtRpcDocsGenerateResults>;
+    getDbtPackages?(
+        dbtVersion: SupportedDbtVersions,
+    ): Promise<DbtPackages | undefined>;
+    test(dbtVersion: SupportedDbtVersions): Promise<void>;
```
ZeRego, line 35: Since this is a class, perhaps the `dbtVersion` should be passed in the constructor. That way we only need to pass it once and is guarantied that all methods of the same instance use the same version.

ZeRego (changes requested):

packages/frontend/src/components/ProjectConnection/index.tsx
rephus: almost, if dbtVersion is "", it will flag it as undefined

packages/frontend/src/components/ProjectConnection/index.tsx
rephus: data?.dbtVersion could be "", from the select (we can't have `undefined` on selects)

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -299,6 +303,7 @@ export class ProjectModel {
                   warehouse_type: null;
                   organization_uuid: string;
                   pinned_list_uuid?: string;
+                  dbt_version: string;
```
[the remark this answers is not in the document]
rephus: dbt_version in dbt is notNullable() now, we default to 1.4

ZeRego (approved):
lgtm 🎉 

Would be good to have this information in the project.created and project.updated events. We can do it in a separate PR ;)

Issue #6589, opened 2023-08-02
Tables prefixed with lightdash aren't being referenced correctly in SQL templates


magnew:
### Description

Tables prefixed with `lightdash` aren't being referenced correctly in SQL templates.

PR #6590, opened 2023-08-02
fix: fix lightdash variable prefix

rephus:
Closes:  #6589

### Description:

Setup 

image
Screenshot from 2023-08-02 15-59-43

Before: 
Screenshot from 2023-08-02 16-00-46

After:

Screenshot from 2023-08-02 16-01-50

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.ts
+++ b/packages/common/src/compiler/exploreCompiler.ts
@@ -21,7 +21,7 @@ import { renderFilterRuleSql } from './filtersCompiler';
 
 // exclude lightdash prefix from variable pattern
 export const lightdashVariablePattern =
-    /\$\{((?!lightdash|ld\b)[a-zA-Z0-9_.]+)\}/g;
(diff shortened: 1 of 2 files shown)


ZeRego (approved):

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -21,7 +21,7 @@ import { renderFilterRuleSql } from './filtersCompiler';
 
 // exclude lightdash prefix from variable pattern
 export const lightdashVariablePattern =
-    /\$\{((?!lightdash|ld\b)[a-zA-Z0-9_.]+)\}/g;
+    /\$\{((?!(lightdash|ld)\.)[a-zA-Z0-9_.]+)\}/g;
```
ZeRego, line 24: Does it work with `table_lightdash.something` ?

rephus: yes, that works, I've added a test for it

Issue #6383, opened 2023-07-18
Update placeholder text for all empty filter fields


PriPatel:
Now that we allow 'empty' filters to be added to dashboards, we need nicer empty states for each type of string, number and date filter. (right now, they tend to be completely empty, with no placeholder text, so it's not really obvious what the user needs to) 
<img width="342" alt="Screenshot 2023-07-18 at 19 03 28" src="https://github.com/lightdash/lightdash/assets/12776522/070f5e55-eb8b-4280-8733-d6d25bd3972b">

Filter type |   | Empty value placeholder text
-- | -- | --
Number | is null | -
  | is not null | -
  | is equal to | Enter value(s)
  | is not equal to | Enter value(s)
  | is less than | Enter value
  | is greater than | Enter value
  |   |  
String | is null |  
  | is not null |  
  | is equal to | Start typing to filter results
  | is not equal to | Start typing to filter results
  | starts with | Enter value(s)
  | ends with | Enter value(s)
  | includes | Enter value(s)
  | does not include | Enter value(s)
Date/Time | is null | -
  | is not null | -
  | is equal to | Select a date
  | is not equal to | Select a date
  | in the last | #
  |   | Select a time period
  | not in the last | #
  |   | Select a time period
  | in the next | #
  |   | Select a time period
  | in the current |  Select a time period
  | is before | Select a date
  | is on or after | Select a date
  | is after | Select a date
  | is on or after | Select a date
  | is between | Start date End date
Boolean | is null | -
  | is not null | -
  | is equal to | True / False

https://www.figma.com/file/btY2rdBe5I2W4xLR3upmQG/Untitled?type=design&node-id=1922-191155&mode=design&t=Ks26oJQK60dE21EA-0

joaoviana:
closed in: https://github.com/lightdash/lightdash/pull/6581

PR #6596, opened 2023-08-02
fix: ability to have empty values in date inputs

IrakliJani:
related to #6383


joaoviana:
On view mode: 
* have empty-value filter
* add value 
* apply
* remove value
* apply button is still enabled, when it should be disabled

image

PR #6603, opened 2023-08-03
Feat/multiple ux improvements chart in dashboard

ZeRego:
Closes: 

### Description:

Changes:

- [x] update dashboard banner button from cross icon to Cancel. On hover, the tooltip should read "Cancel chart creation and return to dashboard"
- [x] update copy in save chart modal to have the dashboard name eg: "Save chart to KPI dashboard"
- [x] There should a be confirmation toast (success) when you land in the dashboard after creating a chart. 'Success! [Chart name] was added to [Dashboard name]'
- [x] remove "dashboard" header when user cancel edit the chart that belongs to the dashboard
- [x] remove double divider in tile menu


magnew (approved):
Looks good!

Issue #6272, opened 2023-07-11
Migrate "Chart Config" popover and tabs


IrakliJani:
### Description

⚠️ this should be done only after completing other tickets in the milestone

acceptance criteria:
- [ ] migrate chart config popover to mantine
- [ ] migrate tabs inside the popover to mantine

![CleanShot 2023-07-11 at 21 23 46@2x](https://github.com/lightdash/lightdash/assets/962095/338db110-daf0-4bb0-8806-9d720fe5f0a8)

PR #6607, opened 2023-08-03
feat: chart config tabs in mantine

magnew:
Closes: #6272 

### Description:

Move the chart config tabs to mantine. Note that this required making the chart config a little wider than the others to keep the tabs from wrapping. I think it still looks fine. It covers more of the chart, but not by much. And we are planning on making that experience better anyway.

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ChartConfigTabs.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ChartConfigTabs.tsx
@@ -1,4 +1,3 @@
-import { Tab, Tabs } from '@blueprintjs/core';
 import {
     convertAdditionalMetric,
     fieldId,
@@ -8,8 +7,8 @@ import {
     Metric,
     TableCalculation,
 } from '@lightdash/common';
-import { Box } from '@mantine/core';
-import { FC, useState } from 'react';
+import { Tabs } from '@mantine/core';
+import { FC } from 'react';
 import { useVisualizationContext } from '../../LightdashVisualization/VisualizationProvider';
 import AxesOptions from './AxesOptions';
 import FieldLayoutOptions from './FieldLayoutOptions';
@@ -19,7 +18,6 @@ import SeriesTab from './Series';
 
 const ChartConfigTabs: FC = () => {
     const { explore, resultsData } = useVisualizationContext();
-    const [tab, setTab] = useState<string | number>('layout');
 
     const dimensionsInMetricQuery = explore
         ? getDimensions(explore).filter((field) =>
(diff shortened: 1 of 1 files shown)


IrakliJani:
code looks good, but something is wrong with render env and we need to run e2e because I think we might need to update e2e tests

magnew:
> code looks good, but something is wrong with render env and we need to run e2e because I think we might need to update e2e tests

Ok, let me see if I can figure that out

IrakliJani:
looks like render did not succeed, creating a missing service in render and retrying now

magnew:
> looks like render did not succeed, creating a missing service in render and retrying now

Thank you! I didn't know how to do this.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ChartConfigTabs.tsx
```diff
@@ -58,35 +56,40 @@ const ChartConfigTabs: FC = () => {
...
-                />
-                <Tab
-                    id="legend"
-                    title="Display"
-                    panel={<LegendPanel items={items} />}
-                />
-                <Tab id="grid" title="Margins" panel={<GridPanel />} />
-            </Tabs>
-        </Box>
+        <Tabs w={335} defaultValue="layout">
```
IrakliJani, line 59: will w="100%" work here?

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ChartConfigTabs.tsx
```diff
@@ -58,35 +56,40 @@ const ChartConfigTabs: FC = () => {
...
+            <Tabs.List mb="sm">
+                <Tabs.Tab px="sm" value="layout">
```
IrakliJani, line 61: why do we need these ?

IrakliJani (commented):
@magnew left few comments, UI looks great, let's also wait for e2e

IrakliJani:
> > looks like render did not succeed, creating a missing service in render and retrying now
> 
> Thank you! I didn't know how to do this.

- whenever render fails to deploy you can go to https://dashboard.render.com/blueprints
- find your blueprint associated to your PR
- click "Manual Sync" button
- it will confirm which service needs to be created
- monitor logs when it starts creating service to find more about an error associated with the failing build

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ChartConfigTabs.tsx
magnew: The padding? It moves the tabs just a little closer together so that the dialog doesn't have to be even wider to accommodate them. It's a bit of a balance :)

IrakliJani: oh I see, so default is MD I guess?

magnew: I think so. I think that's typically the case.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ChartConfigTabs.tsx
magnew: No, this one is slightly wider than all of the others. It has 5 tabs. In my draggable prototype I set these up to all share the same popover, but for now they all make their own and set up the width themselves.

IrakliJani (approved):
all tests passing, UI looks great! merging, thanks @magnew !

IrakliJani:
oops, actually, one test is still failing, rerunning it just in case

Issue #6246, opened 2023-07-11
docs for attribute based access control


TuringLovesDeathMetal:
**Link to existing docs:** 

**Changes?**

PR #6616, opened 2023-08-04
docs: add user attributes reference

owlas:
Closes: #6246

### DIFF
--- a/docs/docs/references/user-attributes.mdx
+++ b/docs/docs/references/user-attributes.mdx
@@ -1,0 +1,83 @@
+---
+tags:
+ - role:admin
+---
+
+# User Attributes
+
+User attributes provide a way to customise Lightdash charts, dashboards, and other behaviour depending on the user 
+that is logged in. For example, you could only show your sales team members data that are relevant to their region.
+
+User attributes are defined for your whole Organization and can only be a text value (not a date or number). Some 
+examples of user attributes are:
+
+- Sales region
+- Department
+- Can view PII
+- Can view financial data
+
+To start user attributes you need to follow 2 steps:
+
+1. Define the user attribute, users can only have user attributes that are explicitly created by admins
+2. Set the user attribute value per user.
+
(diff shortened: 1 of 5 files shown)


docs/docs/references/user-attributes.mdx
```diff
@@ -0,0 +1,85 @@
...
+      sql_filter: ${TABLE}.sales_region = {{ lightdash.attributes.sales_region }}
+```
+
+### 2. Filtering joins with `sql_on`
+
+If you're joining a table, you can also customise the rows that are returned 
+
+You can use user attributes to filter the rows returned by a join. This is useful if you want to restrict the data 
+returned from the joined table. To reference a user attribute in your sql, use the special lightdash reference 
+`{{ lightdash.attributes.<attribute_name> }}`. For example, if you have a user attribute called `sales_region` you can
```
rephus, line 67: Is this the right reference, it should be ${} not {{}}

docs/docs/references/user-attributes.mdx
```diff
@@ -0,0 +1,85 @@
...
+User attributes can be assigned to users by navigating to **Organization Settings** > **User Attributes** and 
+clicking on the user attribute you'd like to assign. Select a user by email address and set their value.
+
+## Using user attributes in Lightdash
+
+There are several places in Lightdash where you can customise behaviour based on user attributes.
+
+### 1. Row filtering with `sql_filter`
+
+You can use user attributes to filter the rows returned by a query. This is useful if you want to restrict the data
```
rephus, line 49: should we mention different alliases ? like sql_where or `ld` `attr`

rephus (commented):
Really nice docs, and amazing videos. well done

Just a few minor comments

docs/docs/references/user-attributes.mdx
owlas: Thanks that's an error!

Issue #6605, opened 2023-08-03
Show dashboard unsaved changes alert when creating chart


ZeRego:
### Description

When creating a chart in a dashboard, when there are unsaved changes from the dashboard, make sure that if you close tab, you get a warning about saving the dashboard.

PR #6623, opened 2023-08-04
fix: show dashboard unsaved changes alert in chartindash flow

yasmin-e:
Closes: #6605 

### Description:
Implemented but creates a side-effect bug.
If I click `Cancel` on the alert, I'm taken out of the chart-in-dashboard flow

https://github.com/lightdash/lightdash/assets/67699259/0e076a34-ba89-42a0-9adb-3c0e5668ad84

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
+++ b/packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
@@ -35,8 +35,13 @@ const AddTileButton: FC<Props> = ({ onAddTiles, intent, popoverPosition }) => {
     const { projectUuid } = useParams<{
         projectUuid: string;
     }>();
-    const { dashboard, dashboardTiles, dashboardFilters } =
-        useDashboardContext();
+    const {
+        dashboard,
+        dashboardTiles,
+        dashboardFilters,
+        haveTilesChanged,
+        haveFiltersChanged,
+    } = useDashboardContext();
     const history = useHistory();
 
     return (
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
```diff
@@ -93,6 +98,13 @@ const AddTileButton: FC<Props> = ({ onAddTiles, intent, popoverPosition }) => {
                                                 ),
                                             );
                                         }
+                                        sessionStorage.setItem(
+                                            'hasDashboardChanges',
+                                            JSON.stringify(
```
ZeRego, line 103: do we need to stringify boolean values?

packages/frontend/src/components/common/modal/ChartCreateModal.tsx
```diff
@@ -152,7 +152,7 @@ const ChartCreateModal: FC<ChartCreateModalProps> = ({
             `/projects/${projectUuid}/dashboards/${dashboardUuid}/edit`,
         );
         showToastSuccess({
-            title: `'Success! ${name} was added to ${fromDashboard}`,
```
ZeRego, line 155: nice catch

packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
yasmin-e: it wasn't catching it when it says 'false' as a string

Issue #6502, opened 2023-07-26
format `NULL` date values as `,,` instead of `,Invalid date,` in exported CSV


IrakliJani:
### Description

`NULL` date values that are then exported to CSV (either Raw or Formatted) are converted to an `Invalid date` string

these should be configured to pass `NULL` values as `,,` to the csv

PR #6627, opened 2023-08-04
fix: write null dates to csvs

owlas:
Closes: #6502

Fixes a bug in `convertRowToCsv` that tried to convert `null` values with `moment()` leading to `Invalid Date` appearing in csvs.

Note this is a breaking change:

 - Existing `null` strings returned `'∅'`
 - Existing `undefined` strings returned `'-'`
 - Now `null` and `undefined` both return `null` and `undefined` respectively

Nulls and undefined are handled by `csv-stringify` and print `,,` 

CleanShot 2023-08-04 at 17 58 51@2x

### DIFF
--- a/packages/backend/src/services/CsvService/CsvService.test.ts
+++ b/packages/backend/src/services/CsvService/CsvService.test.ts
@@ -126,4 +126,36 @@ $4.00,value_4,2020-03-16
 
         expect(csv).toEqual([1, 'value_1', '2020-03-16']);
     });
+
+    it('Should convert with row with null date', async () => {
+        const row = {
+            column_number: 1,
+            column_string: `value_1`,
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/CsvService/CsvService.test.ts
```diff
@@ -126,4 +126,36 @@ $4.00,value_4,2020-03-16
 
         expect(csv).toEqual([1, 'value_1', '2020-03-16']);
     });
+
+    it('Should convert with row with null date', async () => {
```
joaoviana, line 130: I don't think we need `async` keyword here and on line `146` 🤔

joaoviana (approved):
💪 

one non-blocking comment!

Issue #6478, opened 2023-07-25
Simpler dashboard filters in "view mode"


owlas:
### Description

(keep in blueprint)

When the user is in view mode for a dashboard, clicking on a permanent filter has fewer option than edit mode (note that adding temporary filters shows the normal full configuration)

- [ ] Removed "tiles" configuration tab
- [ ] Removed tabs from top
- [ ] Removed filter label
- [ ] If default values were entered by editor, show them in the widget
- [ ] If default values were not entered, show Placeholder text in the widget (but `any value` in the pill)
- [ ] Reset action (see below)

Viewers should also be able to reset the filter to it's original state (to get back to the "any value" / "filter disabled" state or the original default value selected by the editor).

Reset the filter to it's original state
<img width="455" alt="Screenshot 2023-07-18 at 18 07 29" src="https://github.com/lightdash/lightdash/assets/12776522/199353ad-afb1-4e10-b59b-62458eab2268">

### after

<img width="244" alt="CleanShot 2023-07-25 at 15 42 56@2x" src="https://github.com/lightdash/lightdash/assets/11660098/2e835986-97e6-4c05-b2ac-bfd57ea30bc7">

PR #6631, opened 2023-08-04
feat: dashboard filters simplified view mode

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6478

### Description:

- [x] Removed "tiles" configuration tab
- [x] Removed tabs from top
- [x] Removed filter label
- [x] If default values were entered by editor, show them in the widget
- [x] If default values were not entered, show Placeholder text in the widget (but `any value` in the pill)
- [x] Reset action

CleanShot 2023-08-04 at 20 57 47@2x

reset to original value:
CleanShot 2023-08-07 at 16 40 03@2x

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
@@ -47,7 +47,7 @@ interface Props {
     filterRule?: DashboardFilterRule;
     popoverProps?: Popover2Props;
     selectedTabId?: string;
-    isActiveFilter?: boolean;
+    isCreatingNewFilter?: boolean;
     isEditMode: boolean;
     onTabChange: (tabId: FilterTabs) => void;
     onSave: (value: DashboardFilterRule) => void;
@@ -55,7 +55,7 @@ interface Props {
 }
 
 const FilterConfiguration: FC<Props> = ({
-    isActiveFilter = false,
+    isCreatingNewFilter = false,
     isEditMode,
     selectedTabId = DEFAULT_TAB,
     tiles,
@@ -83,9 +83,9 @@ const FilterConfiguration: FC<Props> = ({
         );
 
     const isFilterModified = useMemo(() => {
-        if (!originalFilterRule || !originalFilterRule) return false;
+        if (!originalFilterRule) return false;
 
-        // fix serialization of date fields
+        // fixes serialization of date values
         const serializedInternalFilterRule = produce(
             internalFilterRule,
             (draft) => {
@@ -175,7 +175,7 @@ const FilterConfiguration: FC<Props> = ({
                 value={selectedTabId}
                 onTabChange={(tabId: FilterTabs) => onTabChange(tabId)}
             >
-                {isEditMode || !isActiveFilter ? (
+                {isCreatingNewFilter || isEditMode ? (
                     <Tabs.List mb="md">
                         <Tooltip
                             label="Select the value you want to filter your dimension by"
@@ -226,7 +226,7 @@ const FilterConfiguration: FC<Props> = ({
                 <Box sx={{ flexGrow: 1 }} />
 
                 {isFilterModified && (
-                    <Tooltip label="Reset to original value">
+                    <Tooltip label="Reset to original value" position="left">
                         <Button
                             size="xs"
                             variant="default"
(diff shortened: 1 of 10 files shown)


IrakliJani:
p.s. @PriPatel I don't know why but I picked this button.
we also have these options:

CleanShot 2023-08-07 at 16 54 21@2x

wdyt?

PriPatel:
@IrakliJani thanks for this!
the icon you have chosen already works best :)

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -138,81 +164,95 @@ const FilterConfiguration: FC<Props> = ({
...
+                value={selectedTabId}
+                onTabChange={(tabId: FilterTabs) => onTabChange(tabId)}
             >
-                <Tab
-                    id="settings"
-                    title={
-                        <Tooltip2
-                            content="Select the value you want to filter your dimension by"
-                            position="bottom"
+                {isEditMode || !isActiveFilter ? (
```
IrakliJani, line 176: remove isActiveFilter

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -81,6 +82,31 @@ const FilterConfiguration: FC<Props> = ({
                   ),
         );
 
+    const isFilterModified = useMemo(() => {
+        if (!originalFilterRule || !originalFilterRule) return false;
```
IrakliJani, line 86: remove one of them

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -81,6 +82,31 @@ const FilterConfiguration: FC<Props> = ({
...
+
+        // fix serialization of date fields
+        const serializedInternalFilterRule = produce(
+            internalFilterRule,
+            (draft) => {
+                if (draft.values && draft.values.length > 0) {
+                    draft.values = draft.values.map((v) => v.toString());
```
IrakliJani, line 93: check date type first and then convert

packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
```diff
@@ -83,6 +83,12 @@ const DateFilterInputs = <T extends ConditionalRule = DateFilterRule>(
                             <MonthAndYearInput
                                 disabled={disabled}
                                 placeholder={placeholder}
+                                popoverProps={{
+                                    onOpen: () =>
+                                        popoverProps?.onOpened?.(null as any),
```
IrakliJani, line 88: add comment why we are doing this

packages/frontend/src/components/common/MonthAndYearInput.tsx
```diff
@@ -1,35 +1,44 @@
...
             size="xs"
+            onClick={toggle}
+            {...props}
             popoverProps={{
                 withArrow: true,
                 withinPortal: false,
                 shadow: 'md',
+                ...props.popoverProps,
+                opened: isPopoverOpen,
+                onOpen: () => {
```
IrakliJani, line 28: link issue related to mantine here as a comment

joaoviana:
Create dashboard filter with empty value; 
Save
Click on filter in view mode
`Reset` button should be disabled: 
image

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -138,81 +164,95 @@ const FilterConfiguration: FC<Props> = ({
...
+                            variant="default"
+                            color="gray"
+                            onClick={() => {
+                                if (!originalFilterRule) return;
+                                handleChangeFilterRule(originalFilterRule);
+                            }}
+                        >
+                            <MantineIcon icon={IconRotate2} />
+                        </Button>
+                    </Tooltip>
```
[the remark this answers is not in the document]
IrakliJani: btw, it also resets tile settings to default (no matter if we show button there or not.
so the question is, should we be reverting tile config changes with this button or not? it kinda seems useful but it's controversial 

https://github.com/lightdash/lightdash/assets/962095/bde6f007-0c5b-45b6-babc-4d9f5d4d8ede

joaoviana (approved):

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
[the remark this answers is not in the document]
PriPatel: @IrakliJani The button should only be for the values. The tile config is interesting but maybe we shouldn't go there yet! 
But also... the viewer shouldn't be seeing the tiles tab anyway right? So they would never be in a position to 'revert tile config?'

Issue #6620, opened 2023-08-04
Add password requirements 


owlas:
Add 3 password rules:
- minimum length
- must contain a letter
- must contain a number or a symbol

The UI should show whether these rules have been satisfied `onChange` in the password input field (i.e. revalidate on every keystroke)

ZeRego:
@iampranavdhar If you are comfortable with making changes to the FE and BE.

Pages that need that fix: register, accept an invite, user settings page 
Both the form and the endpoint should have the same validation so perhaps we need code the `common` package which can be used by both.

iampranavdhar:
@ZeRego, @owlas  Can you change the assignee to me?

iampranavdhar:
Made a pull request, consider reviewing the code in your free time :)

PR #6632, opened 2023-08-05
feat: add password requirements

iampranavdhar:
This commit validates password input in all the places

Closes: #6620 

### Description:
Adds an extra validation for password inputs in frontend and backend across invite page, register page and user settings page.

#### On without Validation in Frontend and Sending Data to Backend these are the responses:

image
image
image

#### On validating in frontend the action buttons won't get activated:

image
image
image


iampranavdhar:
Only added text for now as I am not aware of which icons we are using. It looks much more good with the check mark icons.

hamzahc1:
@iampranavdhar are you able to update the commit title (you're missing a space between `feat:` and `add`? Thanks for the contribution btw :D

iampranavdhar:
> @iampranavdhar are you able to update the commit title (you're missing a space between `feat:` and `add` to enable tests to pass? Thanks for the contribution btw :D

Thanks @hamzahc1 rebased the commit message.

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -52,6 +53,10 @@ apiV1Router.get('/flash', (req, res) => {
 
 apiV1Router.post('/register', unauthorisedInDemo, async (req, res, next) => {
     try {
+        if (req.body.password && !validatePassword(req.body.password)) {
```
rephus, line 56: perhaps you meant `!validatePassword(req.body.password).isPasswordValid `?

packages/common/src/index.ts
```diff
@@ -154,6 +155,21 @@ export const validateEmail = (email: string): boolean => {
     return re.test(String(email).toLowerCase());
 };
 
+export function validatePassword(password: string): PasswordValidationResult {
```
rephus, line 158: It would be great to have unit tests for this method

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -52,6 +53,10 @@ apiV1Router.get('/flash', (req, res) => {
...
+            next(new Error('Password does not meet requirements'));
```
rephus, line 57: Better to use `ParameterError` than Error ? `ParameterError` returns a 400 code, while Error I think is 500 .

packages/frontend/src/components/UserSettings/PasswordPanel/index.tsx
```diff
@@ -19,6 +29,10 @@ const PasswordPanel: FC = () => {
     const { isLoading, mutate: updateUserPassword } =
         useUserUpdatePasswordMutation();
 
+    const handleChange = (password: string) => {
```
rephus, line 32: I think this may require an useCallback wrapper, but I think you can get rid of this method by doing `setValidationResult(validatePassword(password));` inside the onChange

packages/frontend/src/components/UserSettings/PasswordPanel/index.tsx
```diff
@@ -1,12 +1,22 @@
+import { validatePassword } from '@lightdash/common';
+import { PasswordValidationResult } from '@lightdash/common/src/types/passwordValidationResult';
```
rephus, line 2: This import is wrong, you are importing this type from `src` (not from dist). It should be like `validatePassword`

packages/common/src/types/passwordValidationResult.ts
```diff
@@ -0,0 +1,6 @@
+export type PasswordValidationResult = {
```
rephus, line 1: If you create a new type file, you also need to export it (like in here https://github.com/lightdash/lightdash/blob/main/packages/common/src/index.ts#L133) so we can use it without pointing to the `src`

rephus: tbh those types are more for models used in DB and APIs, I think you can move this to /index.ts

packages/frontend/src/components/RegisterForms/CreateUserForm.tsx
```diff
@@ -25,6 +39,10 @@ const CreateUserForm: FC<Props> = ({ isLoading, readOnlyEmail, onSubmit }) => {
         },
     });
 
+    const handleChange = (password: string) => {
+        setValidationResult(validatePassword(password));
```
rephus, line 43: move this `setValidationResult` to `onChange` and remove handleChange method

packages/common/src/index.ts
```diff
@@ -154,6 +155,21 @@ export const validateEmail = (email: string): boolean => {
...
+    const minLength = 8;
+    const hasLetter = /[a-zA-Z]/.test(password);
+    const hasNumberOrSymbol = /[\d\W]/.test(password);
```
rephus, line 161: This regex should capture all symbols, however it does not work with `_` :thinking: do we consider that a symbol ? 
image

packages/frontend/src/components/common/PasswordValidationMessages.tsx
```diff
@@ -0,0 +1,29 @@
+import { PasswordValidationResult } from '@lightdash/common/src/types/passwordValidationResult';
```
rephus, line 1: wrong import

rephus (changes requested):
I made some comments, the one about the import is the most important. 

I'm deploying this to render so we can test it there

packages/common/src/index.ts
iampranavdhar: Should we include it? Not sure whether we include this as a symbol or not

packages/common/src/index.ts
iampranavdhar: Done.

iampranavdhar:
@rephus Except that special character everything is done.

packages/frontend/src/components/RegisterForms/CreateUserForm.tsx
rephus: this hasn't been resolved

packages/common/src/index.ts
rephus: Yes, I think we should include `_` it should be ` /[\d\W_]/` but let's create an unit test to make sure

packages/frontend/src/components/RegisterForms/CreateUserForm.tsx
```diff
@@ -1,7 +1,13 @@
-import { CreateUserArgs, validateEmail } from '@lightdash/common';
+import {
+    CreateUserArgs,
+    validateEmail,
+    validatePassword,
+} from '@lightdash/common';
+import { PasswordValidationResult } from '@lightdash/common/src/types/passwordValidationResult';
```
rephus, line 6: this is still using the wrong import

rephus (changes requested):

iampranavdhar:
@rephus Resolved.

rephus (dismissed):
I've tested the functionality, it all seems to be working fine. There is still a UI thing we need to change. The errors should not show by default on any of the pages, they should only appear when a password is typed, same  behaviour as here https://login.mailchimp.com/signup/ 

image

Sorry for all these changes, this should have been part of the ticket spec.

iampranavdhar:
> I've tested the functionality, it all seems to be working fine. There is still a UI thing we need to change. The errors should not show by default on any of the pages, they should only appear when a password is typed, same behaviour as here https://login.mailchimp.com/signup/
> 
> image
> 
> Sorry for all these changes, this should have been part of the ticket spec.

No issue will implement that tonight :)

packages/frontend/src/components/common/PasswordValidationMessages.tsx
```diff
@@ -0,0 +1,29 @@
...
+            </Text>
+            <Text color={getColor(hasNumberOrSymbol)}>
+                Must contain at least one symbol or number
+            </Text>
+            <Text color={getColor(hasLetter)}>
+                Must contain at least one letter
+            </Text>
+        </Stack>
+    );
+};
```
IrakliJani, line 27: would be nice to show validation errors on actual input rather than on a separate view

here are a lot of examples including how to integrate with zod

https://mantine.dev/form/validation/

packages/common/src/index.ts
```diff
@@ -154,6 +154,28 @@ export const validateEmail = (email: string): boolean => {
...
+    const isLengthValid = password.length >= minLength;
+    const isPasswordValid = isLengthValid && hasLetter && hasNumberOrSymbol;
+
+    return {
+        isLengthValid,
+        hasLetter,
+        hasNumberOrSymbol,
+        isPasswordValid,
+    };
+}
```
IrakliJani, line 177: if we rewrite this using ZOD we can simplify a lot of frontend code. mantine integrates with zod nicely and we already use that here for example `packages/frontend/src/components/LimitButton/LimitForm.tsx`

it will simplify this chunk of code as well:

```
const passwordSchema = z.object({
  password: z
    .string()
    .min(8) 
    .regex(/[a-zA-Z]/, { message: 'must contain a letter' })
    .regex(/[\\d\\W_]/, { message: 'must contain a number or symbol' })
});
```

iampranavdhar: Will look into that.

iampranavdhar: Hey @IrakliJani , It's good only but it shows error one by one right ? first it check for size and until size is okay it's shows same error and then it shows next error and then next error, it's like three warnings one by one. If the user enters first eight chars and then he sees that he need to add one number then he thinks and then he adds one number or symbol then there is other error which says atleast one char is needed. Won't it be like halting the user three times and increasing the number of clicks as those form validation get's checked after submission. Instead we can use `PasswordInput` but that won't decrease the code lines instead add some more. 

https://mantine.dev/core/password-input/#strength-meter-example

But with that previous code we can just implement show the PasswordMessages only when input value length. So two options:

1. Adding condition with the previous code.
2. Keeping this schema validation and then also that strength so that user while typing can see the constraints and can choose password accordingly

So which one are we going with? I am done with the schema validation part along with unit test's just, can go with anything as per the requirement.

https://github.com/lightdash/lightdash/assets/73348574/e5992fab-0f6a-4f96-b468-c8a7f8e818e9

IrakliJani: I see... that banner with password strength looks pretty nice.
we should probably ignore UI related changes in the PR and go with what's been implemented so far

tagging @PriPatel since this is part of user's onboarding experience

PriPatel: @iampranavdhar You are right - we should not show the user errors 1 by 1. 
If there are 3 errors, all errors must be shown at the same time. 

Thank you so much!

IrakliJani: @iampranavdhar, what you already have is great. let's skip banner as well, we'll create a separate ticket for it and do that as a separate PR. thanks @iampranavdhar 🙌

iampranavdhar: Hello @IrakliJani  and @PriPatel Tried that strength along with schema validations. This is how it looks. I can push if this looks good.

Uploading Screen Recording 2023-08-08 at 11.51.49 PM.mov…

IrakliJani: looks great @iampranavdhar 🙌

iampranavdhar: Pushed 🚀 @IrakliJani

PriPatel: Yeah @iampranavdhar looks awesome. Thanks so much for this!

IrakliJani: testing this locally now

IrakliJani: hey @iampranavdhar banner items and progress bar is not in sync, can you have a look?

https://github.com/lightdash/lightdash/assets/962095/5266cbfd-887e-4332-aa1c-061a955b80a3

also, let me know if you need help, happy to take this PR from here

iampranavdhar: @IrakliJani and @owlas Can you check it now?

https://github.com/lightdash/lightdash/assets/73348574/52e5a529-27f4-400f-97cd-e5d2c8dbcf44

IrakliJani: found few issues with the view. Im gonna push a fix now

IrakliJani:
need to fix e2e before merging

IrakliJani:
@rephus after backend changes e2e fails because seeded user's password `SEED_ORG_1_ADMIN_PASSWORD = 'demo_password!'` what do you think is the best idea here? I'm hesitant to change seed user's password but e2e is heavily tied to that user.

IrakliJani:
@iampranavdhar @rephus fixed e2e tests and cleaned up UI related stuff.

@rephus feel free to merge
@iampranavdhar thanks for your work on this!

IrakliJani (approved):

IrakliJani:
@all-contributors add @iampranavdhar for code

Issue #6578, opened 2023-08-01
Better JSON-Blob/Complex type support in tables


magnew:
### Description

Complex types in Lightdash results tables can be hard to read and don't always display well. 

https://www.loom.com/share/c89dad0f002547a1baeee3a37ddb5986?sid=73262a3f-a430-4880-9327-cf9c47442b1a

hamzahc1:
This looks great! Only UX improvement I'd suggest would put the "copy to clipboard" option at the bottom of the modal rather than on each line

magnew:
Let me look into whether I can move the clipboard copy. It's coming from the library I'm using, but maybe there is an option. I assume that wouldn't be a blocker?

Yea, I can refactor this a bit and it could theoretically go anywhere.

PR #6640, opened 2023-08-07
feature: json viewer for complex types

magnew:
Closes: #6578 

### Description:

Adds a JSON viewer to the results table. Cells that parse as JSON have an expand option added to their context menu which opens a JSON viewer dialog. 

Note that this is only added to the results table for now. We could bring it to more tables if it seems like a good idea, but this was the first thing asked for and the smallest useful change. 

For reviewers: you can test this by putting JSON in a table calculation.

### DIFF
--- a/packages/frontend/src/components/Explorer/ResultsCard/CellContextMenu.tsx
+++ b/packages/frontend/src/components/Explorer/ResultsCard/CellContextMenu.tsx
@@ -27,8 +27,9 @@ import UrlMenuItems from './UrlMenuItems';
 const CellContextMenu: FC<
     Pick<CellContextMenuProps, 'cell' | 'isEditMode'> & {
         itemsMap: Record<string, Field | TableCalculation>;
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/Explorer/ResultsCard/CellContextMenu.tsx
```diff
@@ -41,6 +42,19 @@ const CellContextMenu: FC<
     const value: ResultValue = cell.getValue()?.value || {};
     const fieldValues = mapValues(cell.row.original, (v) => v?.value) || {};
 
+    let parseResult: null | object = null;
+    if (
+        !!value.raw &&
+        typeof value.raw === 'string' &&
+        (value.raw.startsWith('{') || value.raw.startsWith('['))
```
owlas, line 49: nice catch to reduce computational load

joaoviana (approved):

Issue #6505, opened 2023-07-27
Create an e2e test for the Chart in Dashboard flow


yasmin-e:
### Description
Test should cover this flow
1. Go to dashboard
2. Click "Create new chart"
3. Go to the Dashboard explorer
4. Create the chart
5. Save the chart
6. Save the dashboard
7. Go to All saved charts and make sure the chart is not there

PR #6641, opened 2023-08-07
feat: allow charts to be created in dashboards

ZeRego:
Closes: [#6505](https://github.com/lightdash/lightdash/issues/6505)

### Description:

Changes:
- remove feature flag, so everyone can create charts in dashboards
- update e2e test to create chart in dashboard

Aug-07-2023 19-14-31
Screenshot 2023-08-07 at 19 14 56

### DIFF
--- a/packages/e2e/cypress/e2e/dashboard.cy.ts
+++ b/packages/e2e/cypress/e2e/dashboard.cy.ts
@@ -54,7 +54,7 @@ describe('Dashboard', () => {
         cy.contains('bank_transfer').should('have.length', 0);
     });
 
-    it.only('Should create dashboard with tiles', () => {
+    it('Should create dashboard with tiles', () => {
         cy.visit(`/projects/${SEED_PROJECT.project_uuid}/dashboards`);
 
         cy.contains('Create dashboard').click();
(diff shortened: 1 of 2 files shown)


packages/e2e/cypress/e2e/dashboard.cy.ts
```diff
@@ -69,6 +69,21 @@ describe('Dashboard', () => {
         cy.findByRole('dialog').get('.mantine-MultiSelect-input').click(); // Close dropdown
         cy.findByText('Add').click();
 
+        cy.findAllByText('Add tile').click({ multiple: true });
```
rephus, line 72: should we make this a different test ? If it fails, I don't know if it is because the "chat in dashboard" feature is wrong, or is adding an existing title wrong.

packages/e2e/cypress/e2e/dashboard.cy.ts
```diff
@@ -69,6 +69,21 @@ describe('Dashboard', () => {
...
+        );
+        cy.findByText('Orders').click();
+        cy.findByText('Status').click();
+        cy.findByText('Average order size').click();
+        cy.findByText('Save chart').click();
+        cy.get('input#chart-name').type('Average order size per status');
+        cy.findByText('Save').click();
+        cy.findByText(
+            'Success! Average order size per status was added to Title',
+        ).should('exist');
```
rephus, line 85: We should check the chart appears on the dashboard, and its content matches

rephus (approved):
Some minor comments, non blocking, feel free to merge it. 

We could also add a test to check if the `chat in dashboard`is not appearing on any chart list. This can be done on a separate PR

rephus:
Retriggering render

rephus:
:sheep: it !

PR #6658, opened 2023-08-08
refactor: move field selector to common

magnew:
### Description:

Moves the FieldSelect and FieldItem components from chart config to common. We are going to use the select in filters as well. Also updates its props.

### DIFF
--- a/packages/frontend/src/components/common/FieldSelect/FieldSelect.tsx
+++ b/packages/frontend/src/components/common/FieldSelect/FieldSelect.tsx
@@ -9,9 +9,6 @@ type Props = Omit<SelectProps, 'data'> & {
     label?: string;
     selectedField?: Field | TableCalculation;
     fieldOptions: (Field | TableCalculation)[];
-    placeholder?: string;
-    disabled?: boolean;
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/common/FieldSelect/FieldSelect.tsx
```diff
@@ -1,18 +1,18 @@
...
+import { fieldLabelText } from '../Filters/FieldLabel';
+import FieldSelectItem from './FieldSelectItem';
 
-interface Props {
+type Props = Omit<SelectProps, 'data'> & {
     label?: string;
     selectedField?: Field | TableCalculation;
     fieldOptions: (Field | TableCalculation)[];
     placeholder?: string;
     disabled?: boolean;
```
joaoviana, line 13: can remove these if we're getting `SelectProps`

joaoviana: maybe same for `onChange`?

magnew: Good call

Issue #6396, opened 2023-07-18
view underlying data doesn't apply dashboard filters


owlas:

owlas:
big number tile

PR #6659, opened 2023-08-08
fix: big number dashboard filters + underlying data

IrakliJani:
Closes: #6396 

:warning: if you just wanna review dashboard filter logic and skip mantine menu - then have a look to this commit only
[fix: big number + underlying values + dashboard filters](https://github.com/lightdash/lightdash/pull/6659/commits/48b711137e1a5f10d7dee8e1e423a3976c36711e)

### Description:

https://github.com/lightdash/lightdash/assets/962095/83a11417-3988-4584-94e2-92b2805c1224

### DIFF
--- a/packages/frontend/src/components/LightdashVisualization/index.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/index.tsx
@@ -29,8 +29,9 @@ const LightdashVisualization: FC<LightdashVisualizationProps> = memo(
                 return (
                     <SimpleStatistic
                         minimal={minimal}
-                        isDashboard={isDashboard}
                         isTitleHidden={isTitleHidden}
+                        isDashboard={isDashboard}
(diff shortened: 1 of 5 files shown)


joaoviana:
On render, the following is breaking for me: 
- Go to Jaffle dashboard
- Edit
- Add filter payment method `is equal` to `coupon, credit card`
- Go to big number tile
- 3-dot icon
- explore from here
- Get a lightdash error + error toast

Could you confirm you get the same?

IrakliJani:
thanks for the review @joaoviana, looks like there's other error (server crashing) which should not be because of this changes. rebased and waiting now for render to test again

IrakliJani:
rebased, created separated dashboard for testing @joaoviana, could not reproduce that bug in different PR. existing dashboard has "explore from here" even without that filter

Issue #6372, opened 2023-07-18
Update the operator dropdown in dashboard filters


PriPatel:
<img width="408" alt="Screenshot 2023-07-18 at 09 43 10" src="https://github.com/lightdash/lightdash/assets/12776522/9cf8a94e-0bbb-415e-8a78-a96404a8e4f1">

Right now, we use the default browser dropdown. As this is something a dashboard viewer will have to interact with, it should look and feel nicer.

### Appearance
- [x] Update the operator dropdowns to Mantine components. This includes view mode and edit mode. 

### Copy
Updating the 'main' operator to be shorter will save on real estate and make things look simpler. Update this copy across all operator dropdowns:
- [x] `is equal to` -> `is`
- [x] `is not equal to` -> `is not`

PR #6660, opened 2023-08-08
feat: update filter operator dropdown

joaoviana:
Closes: #6372

### Description:

Replace component with mantine's `Select`

Change Filter label copy: 
* `is equal to` -> `is`
* `is not equal to` -> `is not`

Screen recording

https://github.com/lightdash/lightdash/assets/7611706/81cd7d19-70df-408c-be1a-8399abf53b0a


magnew (approved):
Looks good!

PR #6661, opened 2023-08-08
chore: delete unused images in docs

joaoviana:
Closes: N/A

### Description:

Remove unused images in `docs/`

Stopped at `references/` - to do in next PR

PR #6664, opened 2023-08-09
chore: remove unused images

joaoviana:
Closes: N/A

### Description:

Related to #6661 

Compiler didn't throw errors -> tested with an actual used image and it threw an error, so no false alarm


magnew (approved):
Confirmed there are no compile errors!

PR #6680, opened 2023-08-09
chore: refactor google oauth library

rephus:
Closes: 

### Description:

I had to refactor the library we use to connect to Google because it was not returning the refresh_token

Screenshot from 2023-08-09 16-00-06

[Screencast from 09-08-23 16:12:10.webm](https://github.com/lightdash/lightdash/assets/1983672/dbe4c33b-83df-43e4-94cc-462d0d651edf)

When used this refresh token on the Google client (WIP) 

image

Scopes and google client spec might change in a future PR 

This can be merged without feature flag


packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -0,0 +1,55 @@
+import { lightdashConfig } from '../../config/lightdashConfig';
+import { OpenIdIdentityModel } from '../../models/OpenIdIdentitiesModel';
+
+const { google } = require('googleapis');
+// TODO WIP
+export class GoogleDriveClient {
```
rephus, line 6: Sample google drive client for testing, this will change in a future PR

packages/backend/src/@types/passport-google-oidc.d.ts
```diff
@@ -1,47 +0,0 @@
-declare module 'passport-google-oidc' {
```
rephus, line 1: Removed custom Strategy, used library instead

packages/backend/package.json
```diff
@@ -79,6 +81,7 @@
         "nodemailer": "^6.7.2",
         "nodemailer-express-handlebars": "^5.0.0",
         "passport": "^0.6.0",
+        "passport-google-oauth20": "^2.0.0",
```
rephus, line 84: import a new library that contains the right spec and types for passport  google auth

packages/backend/src/controllers/authentication.ts
```diff
@@ -164,18 +168,27 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
               ).href,
               passReqToCallback: true,
           },
-          async (req, issuer, profile, done) => {
+          async (
```
rephus, line 171: This was the main issue with the previous implementation, the callback method was returning an incomplete response , without refreshToken.

packages/backend/src/controllers/authentication.ts
```diff
@@ -187,18 +200,22 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
                           issuerType: OpenIdIdentityIssuerType.GOOGLE,
                       },
                   };
+
                   const user = await userService.loginWithOpenId(
                       openIdUser,
                       req.user,
                       inviteCode,
+                      refreshToken,
```
rephus, line 208: save refreshtoken on the db

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -146,6 +146,21 @@ apiV1Router.get(
         scope: ['profile', 'email'],
     }),
 );
+apiV1Router.get(
+    '/login/gdrive',
```
rephus, line 150: new endpoint for authentication on gdrive with the right scope and refreshtoken ("offline" accesstype)

packages/backend/src/database/migrations/20230809132300_add_google_oauth_token.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('openid_identities', (t) => {
+        t.string('refresh_token').nullable();
```
owlas, line 5: if you want this to have no limit it should be `.text` 

`.string` is equal to `VARCHAR(255)` so it'll error if there are more than 255 characters

packages/backend/src/controllers/authentication.ts
```diff
@@ -164,18 +168,27 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
...
+          async (
+              req: Express.Request,
+              accessToken: string,
+              refreshToken: string,
+              params: GoogleCallbackParameters,
+              profile: Profile,
+              done: VerifyCallback,
+          ) => {
               try {
+                  const issuer = 'https://accounts.google.com';
```
owlas, line 180: ✅ Verified this is also hardcoded in the `passport-google-oidc` package, so this works

packages/backend/src/controllers/authentication.ts
```diff
@@ -164,18 +168,27 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
...
                   const { inviteCode } = req.session.oauth || {};
                   req.session.oauth = {};
-                  const [{ value: email }] = profile.emails;
+                  const [{ value: email }] = profile.emails || [];
```
owlas, line 183: is the typing correct here? `email` could be undefined but not sure if typscript catches that

packages/backend/src/controllers/authentication.ts
```diff
@@ -164,18 +168,27 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
...
                   const { id: subject } = profile;
                   if (!(email && subject)) {
-                      return done(null, false, {
+                      return done(null, undefined, {
```
owlas, line 186: what does this change?

packages/backend/src/controllers/authentication.ts
```diff
@@ -164,18 +168,27 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
...
                           message: 'Could not parse authentication token',
                       });
                   }
-                  // TODO why we ned this
+
+                  const { scope } = params;
```
owlas, line 191: Should we store this as is? Apparently it should be a space-separated list of scopes, so we could parse it with `.split(' ')` - do we actually use the scopes?

packages/backend/src/controllers/authentication.ts
owlas: We could still store the scope it should just be `openid` or maybe `openid profile`

packages/backend/src/database/entities/openIdIdentities.ts
```diff
@@ -8,6 +8,8 @@ export type DbOpenIdIdentity = {
     user_id: number;
     created_at: Date;
     email: string;
+    refresh_token?: string;
+    scope?: string;
```
owlas, line 12: Was wondering whether to store the raw or parsed version. So should this be a list of `scopes` ? The benefit of storing raw is we don't need to bother parsing it upfront and can deal with it later. The downside is that we might have some weird `scope` strings in the database

I don't mind either way

packages/backend/src/database/migrations/20230809132300_add_google_oauth_token.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('openid_identities', (t) => {
+        t.text('refresh_token').nullable();
+        t.string('scope').nullable();
```
owlas, line 6:
```suggestion
        t.text('scope').nullable();
```

packages/backend/src/models/OpenIdIdentitiesModel.ts
```diff
@@ -144,4 +148,15 @@ export class OpenIdIdentityModel {
                 .delete();
         });
     }
+
+    async getRefreshToken(userId: number) {
```
owlas, line 152: We almost always use `userUuid` in the model layers

owlas: I guess this can return `null` although it's not in the return type. 

Should this be `findRefreshToken`? Usually we throw an error in methods named `get..` instead of returning undefined.

packages/backend/src/models/OpenIdIdentitiesModel.ts
```diff
@@ -144,4 +148,15 @@ export class OpenIdIdentityModel {
...
+        const [row] = await this.database(OpenIdIdentitiesTableName)
+            .where('user_id', userId)
+            .select('refresh_token');
+
+        if (!row) {
+            throw new NotFoundError('No user found');
```
owlas, line 158: catches missing user but not `null` refresh token. maybe intended?

packages/backend/src/routers/apiV1Router.ts
owlas: this is awesome!

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -146,6 +146,21 @@ apiV1Router.get(
...
+    passport.authenticate('google', {
+        scope: [
+            'profile',
+            'email',
+            'https://www.googleapis.com/auth/drive.metadata.readonly',
+        ],
+        accessType: 'offline',
+        prompt: 'consent',
+        session: false,
+        includeGrantedScopes: true,
```
owlas, line 161: more importantly, should we also have `includeGrantedScopes` on the original auth endpoint? 

In the case that the user authorizes drive scope and then later logs in again on the login endpoint (which doesn't include the drive scope). We don't want to lose it.

owlas (changes requested):
Great work! Feels awesome to unblock this. 

I asked a few questions inline

packages/backend/src/routers/apiV1Router.ts
rephus: Let's add this `includeGrantedScopes: true,` to the other google endpoint

rephus:
It is failing userattributes e2e tests becuase the initial migration didn't have the right dbt project. We can ignore

Issue #6655, opened 2023-08-08
`lightdash generate -s tag:one,tag:two` doesn't handle the comma operator in the same way that dbt does


IrakliJani:
### Description

`lightdash generate -s tag:one,tag:two` doesn't handle the comma operator in the same way that dbt does

note that ` ` (space) separator works but `,` does not.

screenshots:

![CleanShot 2023-08-08 at 18 16 45@2x](https://github.com/lightdash/lightdash/assets/962095/465bf18b-0313-402d-aa54-3d018c3f72ab)
![CleanShot 2023-08-08 at 18 17 27@2x](https://github.com/lightdash/lightdash/assets/962095/540102db-5d06-4460-aab1-808c1fa0ceed)

you can test how dbt selectors work here using `dbt ls -s` command

![CleanShot 2023-08-08 at 18 19 59@2x](https://github.com/lightdash/lightdash/assets/962095/7ea8df1f-7271-4522-b83e-09ac4ade1d16)

💭 can delegating to `dbt ls -s tag:test1,tag:test2 --output=json` be useful rather than implementing parser on our end?

IrakliJani:
related: https://github.com/lightdash/lightdash/issues/5140

PR #6684, opened 2023-08-09
feat: generate with dbt compatible --select and --exclude

IrakliJani:
🔍 exploring 🔍 

Closes: https://github.com/lightdash/lightdash/issues/6655

### Description:

delegating --select and --explore to dbt instead of figuring out model dependencies ourselves

### DIFF
--- a/packages/cli/src/dbt/models.ts
+++ b/packages/cli/src/dbt/models.ts
@@ -316,14 +316,20 @@ export const getCompiledModelsFromManifest = async ({
             args.push('--exclude', exclude.join(' '));
         }
 
+        const spinner = GlobalState.startSpinner(`Selecting models`);
         try {
             const { stdout } = await execa('dbt', [
                 'ls',
                 ...args,
+                '--resource-type=model',
                 '--output=json',
             ]);
 
-            const filteredModelIds = JSON.parse(stdout)
+            const filteredModelIds = stdout
+                .split('\n')
+                .map((line) => line.trim())
+                .filter((line) => line.length > 0)
+                .map((line) => JSON.parse(line))
                 .filter((model: any) => model.resource_type === 'model')
                 .map((model: any) => model.unique_id);
 
@@ -332,6 +338,8 @@ export const getCompiledModelsFromManifest = async ({
             );
         } catch (e) {
             console.error(styles.error(`Failed to filter models: ${e}`));
+        } finally {
+            spinner.stop();
         }
     }
 
--- a/packages/cli/src/dbt/models.ts
+++ b/packages/cli/src/dbt/models.ts
@@ -317,12 +317,15 @@ export const getCompiledModelsFromManifest = async ({
         }
 
         try {
-            const { stdout } = await execa('dbt', ['ls', ...args]);
+            const { stdout } = await execa('dbt', [
+                'ls',
+                ...args,
+                '--output=json',
+            ]);
 
-            const filteredModelIds = stdout
-                .split('\n')
-                .filter((line) => line !== '')
-                .map((line) => `model.${line}`);
+            const filteredModelIds = JSON.parse(stdout)
+                .filter((model: any) => model.resource_type === 'model')
+                .map((model: any) => model.unique_id);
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/dbt/models.ts
```diff
@@ -312,131 +292,57 @@ export const getModelsFromManifest = (
...
+            args.push('--exclude', exclude.join(' '));
+        }
+
+        try {
+            const { stdout } = await execa('dbt', ['ls', ...args]);
+
+            const filteredModelIds = stdout
+                .split('\n')
+                .filter((line) => line !== '')
+                .map((line) => `model.${line}`);
```
rephus, line 325: will this work if we also use `tags` instead of just model names ?

rephus: we need to handle the output better, I got all sorts of responses when running `dbt ls` eg: 

image


image

image

Some dbt versions might use `stderr` instead of `stdout` :fire:

rephus (commented):
This looks promissing, the logic is a lot better than before, my only concern is with the different `selection` types.  

I think we need to implement at least some unit tests for this `getCompiledModelsFromManifest` method

packages/cli/src/dbt/models.ts
IrakliJani: @rephus, yes! that works with tags.
I will try the json output which should be way better to deal with

packages/e2e/cypress/cli/cli.cy.ts
```diff
@@ -48,9 +49,134 @@ describe('CLI', () => {
...
+            .should('contain', 'payments')
+            .should('contain', 'stg_customers')
+            .should('contain', 'stg_orders')
+            .should('contain', 'stg_payments')
+            .should('contain', 'Done 🕶');
+    });
+
+    it('Should lightdash generate with --select and --exclude', () => {
+        cy.exec(
+            `${cliCommand} generate -y --project-dir ${projectDir} --profiles-dir ${profilesDir} --select +orders --exclude stg_orders`,
```
rephus, line 154: can we exclude multiple models ? similar to the test you wrote with multiple `--select`

packages/e2e/cypress/cli/cli.cy.ts
```diff
@@ -48,9 +49,134 @@ describe('CLI', () => {
...
+            .should('not.contain', 'stg_customers')
+            .should('not.contain', 'stg_orders')
+            .should('contain', 'stg_payments')
+            .should('contain', 'Done 🕶');
+    });
+
+    it('Should lightdash generate all model', () => {
         cy.exec(
-            `${cliCommand} generate  -y --project-dir ${projectDir} --profiles-dir ${profilesDir}`,
+            `${cliCommand} generate -y --project-dir ${projectDir} --profiles-dir ${profilesDir}`,
```
rephus, line 179: :kissing: :ok_hand:

packages/e2e/cypress/cli/cli.cy.ts
```diff
@@ -48,9 +49,134 @@ describe('CLI', () => {
...
+                },
+            },
+        )
+            .its('stderr')
+            .should('contain', 'Filtering models')
+            .should('not.contain', 'customers')
+            .should('contain', 'orders')
+            .should('not.contain', 'events')
+            .should('not.contain', 'users')
+            // .should('not.contain', 'payments')
```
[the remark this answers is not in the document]
IrakliJani: it does not, but it returns stg_payments and output has weird (coloring escape) characters in it which I don't wanna include in tests. any ideas?

packages/e2e/cypress/cli/cli.cy.ts
IrakliJani: yep, I can add that to the test

IrakliJani:
> I guess you can't test with `tags` because we don't use it on our jaffle shop. We could create a separate ticket for adding tags and test it here.

yep, exactly. here's a ticket https://github.com/lightdash/lightdash/issues/6769

PR #6690, opened 2023-08-10
fix: set to default tab filter tab when saving

joaoviana:
Closes: N/A

### Description:

Steps to reproduce bug: 
* Go to dashboard
* enable edit mode
* click on saved filter
* go to tiles tab
* click 'apply'
* save dashboard
* (now on view mode)
* click on same filter
* can only view tiles tab (not the correct view mode tab) 

When applying/saving a filter, reset tab to be the default - `FilterTabs.SETTINGS`. This fixes the issue. 

current bug

https://github.com/lightdash/lightdash/assets/7611706/de253354-1b59-4a44-b72b-2186255d7aa4

fixed with this pr:

https://github.com/lightdash/lightdash/assets/7611706/3ce3f477-85eb-4f90-b24b-a019e1bc3a02


magnew (approved):
Looks good

Issue #6645, opened 2023-08-08
Create new export type in Scheduled Deliveries for Google Sheets


hamzahc1:
### Description

Acceptance Criteria: 

- [ ] When user selects Google Sheets as a scheduled delivery option, use [Google Picker](https://developers.google.com/drive/picker/guides/overview) to authenticate user and request permission to Google Drive to select which sheet they want to share to. 
- [ ] IF we reuse the existing Google OAuth token but we don't want people to necessarily authenticate via Google to use this integration (e.g they have an SSO provider) we need to add an extra environment variable to disable Google login. 
- [ ] Data should be sent to a new page/tab in selected Google Sheet 
- [ ] Each scheduled delivery should update data in the selected sheet

owlas:
We'll need users to sign-in with google again and update the scopes to include viewing google drive files. Then store this oauth2 token in the backend (just like we do for SSO). Since we'll need these credentials to open the google picker. 

I'm not sure if we should use the existing table in the database that contains user identities or not. I think it makes sense to, we just have to make sure it has the correct scopes.

rephus:
@hamzahc1 2 UI questions: 

- Do we call the destination `Google drive` ? or `Google sheets` ? 
- how do we reorganize the items on the scheduler modal ? 

![image](https://github.com/lightdash/lightdash/assets/1983672/9ac4a868-ec61-4182-88a6-bd84f1e8aef5)

hamzahc1:
It should be `Google Sheets` not Google Drive. In terms of UI organization, perhaps we should make this a drop-down menu so it's a bit easier to view? That seems like the smallest change we can make. 

Long term, I think we would rejig the UI so you pick the destination first, then everything else. For now - dropdown! would that work?

rephus:
@hamzahc1 I can get all these details from the google picker. do you want to display/store sometihng in particular ? we could use the URl to point to the file in a _blank page 

![image](https://github.com/lightdash/lightdash/assets/1983672/71b64bd4-886b-433d-8c81-d42dc5ae6b60)

This is what I currently have

[Screencast from 10-08-23 11:51:03.webm](https://github.com/lightdash/lightdash/assets/1983672/8f018600-9484-400e-a767-92dc5b88098e)

hamzahc1:
I think this looks right although the upload tab isn't useful here since you can't upload anything here?

Issue #6673, opened 2023-08-09
Create scheduled delivery option to Google Sheets for dashboards


hamzahc1:
### Description

Acceptance Criteria: 

- [ ] UX should be similar to scheduling a chart, BUT you can select which charts from the dashboard (or select all) you want to export to Google Sheets 
- [ ] Each chart would be a new tab in the Google Sheet

rephus:
Should we start with `all` dashboards ? like when sending a csv via email. this way we don't have to change the UI

PR #6696, opened 2023-08-10
feat: export google drive

rephus:
Closes: #6645

[Screencast from 11-08-23 15:38:20.webm](https://github.com/lightdash/lightdash/assets/1983672/b8a4f4b8-a409-4b72-83f5-b5b2e6bc1471)

Polling Auth flow: 

[Screencast from 10-08-23 16:30:38.webm](https://github.com/lightdash/lightdash/assets/1983672/13c309e4-cd59-4178-8d37-5c9a09332a2d)

## Acceptance Criteria: 

- [x] When user selects Google Sheets as a scheduled delivery option, use [Google Picker](https://developers.google.com/drive/picker/guides/overview) to authenticate user and request permission to Google Drive to select which sheet they want to share to. 
- [x] IF we reuse the existing Google OAuth token but we don't want people to necessarily authenticate via Google to use this integration (e.g they have an SSO provider) we need to add an extra environment variable to disable Google login. 
- [x] Data should be sent to a new page/tab in selected Google Sheet 
- [x] Each scheduled delivery should update data in the selected sheet 

## Out of scope
- [x] better UI to prevent selecting image type with gsheets
- [ ] gsheets scheduled delivery on dashboards
- [ ] We store the google sheet name on db when selected, we don't update it if the sheet name changes.
- [ ] Docs about how to enable this 
- [ ] New ENV variables on cloud 
- [ ] Reauthenticate on google flow


packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -52,4 +48,47 @@ export class GoogleDriveClient {
...
+                            },
+                        },
+                    },
+                ],
+            },
+        });
+
+        await sheets.spreadsheets.values.append({
+            spreadsheetId: fileId,
+            range: `'${tabTitle}'!A:B`,
```
rephus, line 85: if we don't want to upload this data on a new tab we just need to update this range and remove the previous `batchUpdate`

packages/backend/package.json
```diff
@@ -18,6 +18,7 @@
         "@types/passport": "^1.0.12",
         "@types/passport-google-oauth20": "^2.0.11",
         "@types/passport-local": "^1.0.34",
+        "@types/passport-oauth2-refresh": "^1.1.1",
```
rephus, line 21: This is for generating an accessToken from a refreshToken

packages/backend/src/config/parseConfig.ts
```diff
@@ -305,6 +307,8 @@ const mergeWithEnvironment = (config: LightdashConfigIn): LightdashConfig => {
                     process.env.AUTH_GOOGLE_OAUTH2_CLIENT_SECRET,
                 loginPath: '/login/google',
                 callbackPath: '/oauth/redirect/google',
+                googleDriveApiKey: process.env.GOOGLE_DRIVE_API_KEY,
+                enabled: process.env.AUTH_GOOGLE_ENABLED !== 'false',
```
rephus, line 311: We will use the same AUTH_GOOGLE_OAUTH2_CLIENT_SECRET for giving access to gdrive, but some people might not wan to use it as a login method. So we introduced a new var to disable this

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -746,6 +758,180 @@ export const sendEmailNotification = async (
...
+            const refreshToken = await userService.getRefreshToken(
+                scheduler.createdBy,
+            );
+            await googleDriveClient.appendToSheet(
+                refreshToken,
+                gdriveId,
+                csvContent,
+            );
+        } else if (dashboardUuid) {
+            throw new Error('Not implemented');
```
rephus, line 835: This will be implemented next in https://github.com/lightdash/lightdash/issues/6673

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -746,6 +758,180 @@ export const sendEmailNotification = async (
...
+            targetType: 'gdrive',
+            status: SchedulerJobStatus.ERROR,
+            details: { error: e.message },
+        });
+
+        throw e; // Cascade error to it can be retried by graphile
+    }
+};
+
+const logScheduledTarget = async (
```
rephus, line 891: refactor this logic, now we have 1 more target we can't do `isSlackTarget(target)) ? slack : email `

packages/backend/src/services/CsvService/CsvService.ts
```diff
@@ -377,13 +381,23 @@ export class CsvService {
             });
             const s3Url = await this.s3Service.uploadCsv(csvContent, fileId);
 
-            await fsPromise.unlink(`/tmp/${fileId}`);
-
-            return { filename: `${chart.name}`, path: s3Url };
+            // Delete local file in 10 minutes, we could still read from the local file to upload to google sheets
+            setTimeout(async () => {
+                await fsPromise.unlink(`/tmp/${fileId}`);
```
rephus, line 386: mm, yeah... this method is called on a separate scheduler job(handle) , so it is not easy to know if a target is going to be using this file to read. (on gdrive we read from file and upload the csvcontent) . 

I don't want to return the `csvResults` on this method becuase that could cause OOM errors

packages/backend/src/services/UserService.ts
```diff
@@ -848,4 +849,46 @@ export class UserService {
...
+            );
+        });
+    }
+
+    /**
+     * This method is used on the gdrive API to get the accessToken for listing files on the user's drive
+     * @param user
+     * @returns accessToken
+     */
+    async getAccessToken(user: SessionUser): Promise<string> {
```
rephus, line 876: I wanted to do this on `authentication` file, but we call useService in there, so this was going to cause circular dependencies

packages/frontend/src/components/SchedulerModals/SchedulerModalBase/SchedulerForm.tsx
```diff
@@ -221,9 +225,51 @@ const SchedulerForm: FC<{
...
+
+    const handleOpenPicker = useCallback(
+        (callback) => {
+            if (
+                !health.data?.auth.google.oauth2ClientId ||
+                !health.data.auth.google.googleDriveApiKey
+            )
+                return;
+
+            if (gdriveAuth === undefined) {
```
rephus, line 244: auth flow in the frontend. If not authenticated, it will open a popup and `poll` to get the new `accessToken` once it is authenticated.

packages/frontend/src/components/SchedulerModals/SchedulerModalBase/SchedulerForm.tsx
```diff
@@ -430,6 +474,30 @@ const SchedulerForm: FC<{
...
+                                                    setShowDestinationLabel(
+                                                        true,
+                                                    );
+                                                }}
+                                                disabled={disabled}
+                                            />
+                                        </TargetRow>
+                                    );
+                                } else if ('gdriveId' in target) {
+                                    //TODO replace with isGdrive
```
rephus, line 491: I tried, but typescript didn't like this typing

packages/frontend/src/components/UserSettings/SocialLoginsPanel/index.tsx
```diff
@@ -22,7 +22,7 @@ const isIssuerTypeAvailable = (
 ) => {
     switch (issuerType) {
         case OpenIdIdentityIssuerType.GOOGLE:
-            return !!health.auth.google.oauth2ClientId;
+            return health.auth.google.enabled;
```
rephus, line 25: oauth2ClientId can be used for gdrive, so instead we disable goggle auth on new variable (and more consistent with the rest of SSO)

rephus:
I figured out why this was happening on my PRs 
image

It should not happen anymore

packages/frontend/src/components/SchedulerModals/SchedulerModalBase/SchedulerForm.tsx
```diff
@@ -430,6 +474,30 @@ const SchedulerForm: FC<{
                                                     />
                                                 </Box>
                                             </Tooltip>
+                                            <Button
```
IrakliJani, line 477: I think it's fine to use blueprint in this PR as everything in scheduler is blueprint and it is not adding a lot of migration debt. we'll do a migration all together later

IrakliJani:
layout looks fine to me 👍 

CleanShot 2023-08-14 at 18 25 41@2x

packages/backend/src/config/parseConfig.ts
owlas: @rephus this is a breaking change for hosting lightdash right? Cloud + self-hosters must now activate this env var to have their integration continue to work?

packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -52,4 +49,44 @@ export class GoogleDriveClient {
             console.debug(`${file.name} (${file.id})`);
         });
     }
+
+    async appendToSheet(
+        refreshToken: string,
+        fileId: string,
+        csvContent: string,
```
rephus, line 56: Instead of passing the file as content, we could do stream reading

packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -52,4 +49,44 @@ export class GoogleDriveClient {
...
+    ) {
+        if (!this.isEnabled) {
+            throw new Error('Google Drive is not enabled');
+        }
+        const authClient = await GoogleDriveClient.getCredentials(refreshToken);
+
+        const sheets = google.sheets({ version: 'v4', auth: authClient });
+        const sheetRows = csvContent.split('\n').map((row) => row.split(','));
```
owlas, line 64: I don't think this can be used in place of a csv-reader library. For example this will split quoted strings `","` in a single cell.

packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -52,4 +49,44 @@ export class GoogleDriveClient {
...
+
+        // Clear first sheet before writting
```
owlas, line 66: we could ask the user to specify a specific sheet (this would prevent the problem where a user reorders the sheet in the UI and then another sheet gets overwritten)

owlas (commented):
some architecture questions

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -746,6 +758,180 @@ export const sendEmailNotification = async (
...
+        const schedulerUrl = `${url}?scheduler_uuid=${schedulerUuid}`;
+
+        if (format === SchedulerFormat.IMAGE) {
+            throw new Error('Not implemented');
+        } else if (savedChartUuid) {
+            if (csvUrl === undefined) {
+                throw new Error('Missing CSV URL');
+            }
+
+            const csvContent = await fsPromise.readFile(csvUrl.localPath, {
```
owlas, line 822: from our experience this will certainly cause OOM without streaming

packages/backend/src/config/parseConfig.ts
rephus: no,  if "oauth" is enabled and this variable is undefined, the integration should still be available. 

Only if you add oauth but you don't want to have it on `login`, you add `AUTH_GOOGLE_ENABLED='false'`

packages/backend/src/clients/Google/GoogleDriveClient.ts
rephus: I'm using `csvtojson` to convert the file into a `Record<string,string>[]` , that library even supports streaming. I don't see the advantage of writting a json format at the moment, and that requires a bit of additional refactoring . We can consider this on a separate ticket

packages/backend/src/scheduler/SchedulerTask.ts
rephus: Yes, that is going to happen, but I'd rather implement this on a separate PR

packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -52,4 +49,44 @@ export class GoogleDriveClient {
...
+        // The method "SheetId: 0" only works if the first default sheet tab still exists (it's not deleted by the user)
+        // So instead we select all the cells in the first tab by its name
+        try {
+            const spreadsheet = await sheets.spreadsheets.get({
+                spreadsheetId: fileId,
+            });
+            const firstSheetName = spreadsheet.data.sheets[0].properties.title;
+            Logger.debug(`Clearing sheet name ${firstSheetName}}`);
+            await sheets.spreadsheets.values.clear({
```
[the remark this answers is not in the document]
rephus: Good idea about protected ranges , I forgot about them. 

I'm not sure if we can do `all cells` in a named-range. The problem with deletion is that you want to remove all the content before writting, and you can't rely on the number of rows/columns for the content (on the new request you could have more or less rows/columns)

packages/backend/src/config/parseConfig.ts
rephus: let's do `AUTH_GOOGLE_ENABLED === 'true' ` and change the infra

packages/backend/src/clients/Google/GoogleDriveClient.ts
rephus: let's create a ticket

packages/backend/src/clients/Google/GoogleDriveClient.ts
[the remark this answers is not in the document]
rephus: cool, let's do it on a new ticket
the way it works : upload data -> create name-range -> on new data -> delete previous name range and update

packages/backend/src/config/parseConfig.ts
rephus: Done. we need to change the infra before we merge this PR

packages/frontend/src/components/SchedulerModals/SchedulerModalBase/SchedulerForm.tsx
```diff
@@ -289,6 +340,10 @@ const SchedulerForm: FC<{
                                     disabled: isImageDisabled,
                                 },
                                 { value: SchedulerFormat.CSV, label: 'CSV' },
+                                {
+                                    value: SchedulerFormat.GSHEETS,
```
rephus, line 344: New type here

packages/backend/src/scheduler/SchedulerTask.ts
rephus: Created a new ticket in milestone

packages/backend/src/clients/Google/GoogleDriveClient.ts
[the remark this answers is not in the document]
rephus: Created a new ticket in milestone

packages/backend/src/clients/Google/GoogleDriveClient.ts
rephus: Created a new ticket in milestone

packages/backend/src/config/parseConfig.ts
```diff
@@ -305,6 +307,8 @@ const mergeWithEnvironment = (config: LightdashConfigIn): LightdashConfig => {
                     process.env.AUTH_GOOGLE_OAUTH2_CLIENT_SECRET,
                 loginPath: '/login/google',
                 callbackPath: '/oauth/redirect/google',
+                googleDriveApiKey: process.env.GOOGLE_DRIVE_API_KEY,
+                enabled: process.env.AUTH_GOOGLE_ENABLED === 'true',
```
owlas, line 311: @owlas open an issue on lightdash-cloud to update this

packages/backend/src/database/entities/scheduler.ts
```diff
@@ -81,3 +89,24 @@ export type SchedulerLogTable = Knex.CompositeTableType<
...
+            type: 'slack',
+        };
+    }
+    if (isEmailTarget(target)) {
+        return {
+            schedulerTargetId: target.schedulerEmailTargetUuid,
+            type: 'email',
+        };
+    }
+    return assertUnreachable(target, 'Uknown target type');
```
owlas, line 111: typo

packages/backend/src/services/UserService.ts
owlas: `Pick<SessionUser, 'userUuid'>` then you can reuse?

rephus:
Render is messed up because we have removed a migration, so we are going to merge it without running tests :see_no_evil:

Issue #6652, opened 2023-08-08
Fix inconsistent graphile worker migration error


rephus:
### Description

It happens on render from time to time 

Error migrating graphile worker

> Aug 8 11:07:25 AM 2023-08-08 09:07:25 [Lightdash] error: Error migrating graphile worker column "revision" of relation "jobs" already exists

PR #6700, opened 2023-08-11
fix: graphile migration

rephus:
Closes: #6652

### Description:

Running reset-db

image

image

image


packages/backend/src/scheduler/SchedulerClient.ts
```diff
@@ -66,11 +66,6 @@ export class SchedulerClient {
         this.graphileUtils = makeWorkerUtils({
             connectionString: lightdashConfig.database.connectionUri,
         })
-            .then((utils) => utils)
```
rephus, line 69: This was not doing the migrate anyway, previously it was `utisl.migrate().catch`

owlas: For some reason I thought this did actually do a migration when doing `makeWorkerUtils` but checking the source code I can't see it anymore. 

In this discussion the maintainer mentions it runs migrations? Or maybe they're talking about `.run`

https://github.com/graphile/worker/issues/101

How did you decide that this doesn't run migrations?

rephus: According to the docs, the way to run migration is with `utils.migrate()` https://github.com/graphile/worker#library-usage-queueing-jobs

But I'm replacing this method (that was failing randomly) with the migration on `migrate`

owlas: Found it! It _does_ migrate: https://github.com/graphile/worker/blob/5fb57d42f53929cda1e64838fff65da1ed3cc442/src/lib.ts#L222-L224

If you follow `makeWorkerUtils` -> `getUtilsAndReleasersFromOptions` you can see it does `.migrate` in there. I have no idea why. I can't tell for sure if it executes or not from the source, I think it does inside of the `withReleasers`

rephus: Mm, confusing, that probably means we can remove the new migration on `migrate`and even the previous `.then((utils) => utils)` I'm still not sure how to prevent these random errors

owlas: here's what I figured out previously doing this:

- there is no migration lock (i.e. the migration can run multiple times simultaneously) unlike knex migrations
- if the migration runs multiple times simultaneously than it will most likely error
- **but** if the migration has already happened successfully, then running the migration again even if many migrations run at the same time, we won't get the error.


So we need to reliably run the migration _once_ and then afterwards it doesn't matter that we `makeWorkerUtils` many times. 

I like the fix of trying to run migrations in `yarn migrate` - it could still cause problems with render though because they seem to create the container multiple times (i.e. runs migrations multiple times)

rephus: Sooo, do I need to make any changes ?

owlas: Leave the `catch` here so that it crashes the app (i.e. restarts the pod in k8s) if the utils can't be created. 


If you don't then the app will keep running but `utils` won't exist at all and all scheduler operations will fail

owlas (changes requested):
Remove changes to `SchedulerClient` and merge the rest

rephus:
e2e seems to be failing, so I don't trust this change... I'll close it for now, we can reopen this later 

> graphile_worker.add_job(text, json, text, timestamp with time zone, integer, text, integer, text[]) does not exist\u001b[39m\n\u001b[33m

Issue #6536, opened 2023-07-28
Error when copying content on preview


rephus:
### Description

This request failed:

>  "message": "Unable to copy content on preview error: insert into \"saved_queries_version_additional_metrics\" (\"base_dimension_name\", \"compact\", \"description\", \"filters\", \"format\", \"hidden\", \"label\", \"name\", \"round\", \"saved_queries_version_id\", \"sql\", \"table\", \"type\") values ($1... $2275) returning * - invalid input syntax for type json",

https://lightdash.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22collectors%5C%22%7D%20%7C%3D%20%60duplic%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D

tnightengale:
Same exact error right now...

PR #6703, opened 2023-08-11
fix: stringify additional metrics filters on copy

owlas:
Closes: #6536

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -1051,6 +1051,7 @@ export class ProjectModel {
             const copyChartVersionContent = async (
                 table: string,
                 excludedFields: string[],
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1093,6 +1099,7 @@ export class ProjectModel {
             await copyChartVersionContent(
                 'saved_queries_version_additional_metrics',
                 ['saved_queries_version_additional_metric_id', 'uuid'],
+                { filters: (value: any) => JSON.stringify(value) },
```
IrakliJani, line 1102: if this can be unknown that will be great! same thing in the fieldPreprocess arg. we've been victims of `any` too many times now 😆

IrakliJani: btw, just curious, any ideas why the json parser was failing?

IrakliJani (approved):

PR #6710, opened 2023-08-11
feat: lighten dashboard-header's shadow

joaoviana:
Closes: N/A

### Description:

Lighten `pageheader`'s shadow.

before: 
image

after:
image

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -188,7 +188,7 @@ const SavedChartsHeader: FC = () => {
                 </p>
             </Alert>
 
-            <PageHeader>
+            <PageHeader withShadow>
                 <PageTitleAndDetailsContainer>
                     {savedChart && (
                         <>
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -188,7 +188,7 @@ const SavedChartsHeader: FC = () => {
                 </p>
             </Alert>
 
-            <PageHeader withShadow>
+            <PageHeader>
                 <PageTitleAndDetailsContainer>
                     {savedChart && (
                         <>
--- a/packages/frontend/src/components/common/Page/PageHeader.tsx
+++ b/packages/frontend/src/components/common/Page/PageHeader.tsx
@@ -13,6 +13,10 @@ const PageHeader: FC = ({ children }) => (
         px="lg"
         py="md"
         bg="white"
+        /**
+         * FIXME: This shadow should be sourced from Mantine's theme config;
+         * Once migration from Blueprint is complete, address default shadow
+         */
         shadow="0 0 0 1px #bec1c426"
         radius="unset"
         sx={{ zIndex: 1 }}
--- a/packages/frontend/src/components/common/Page/PageHeader.tsx
+++ b/packages/frontend/src/components/common/Page/PageHeader.tsx
@@ -1,11 +1,9 @@
 import { Card, Flex } from '@mantine/core';
-import { FC, ReactNode } from 'react';
+import { FC } from 'react';
 
 export const PAGE_HEADER_HEIGHT = 80;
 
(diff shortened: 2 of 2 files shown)


packages/frontend/src/components/common/Page/PageHeader.tsx
```diff
@@ -1,26 +1,26 @@
...
-            radius="unset"
-            sx={{ zIndex: 1 }}
-        >
-            {children}
-        </Card>
-    );
-};
+type Props = { withShadow?: boolean; children: ReactNode };
+
+const PageHeader: FC<Props> = ({ withShadow, children }) => (
```
joaoviana, line 8: what do you think? 🤔 
@IrakliJani & @magnew

packages/frontend/src/components/common/Page/PageHeader.tsx
```diff
@@ -1,26 +1,26 @@
...
+    <Card
+        component={Flex}
+        justify="flex-end"
+        align="center"
+        pos="relative"
+        h={PAGE_HEADER_HEIGHT}
+        px="lg"
+        py="md"
+        bg="white"
+        shadow={withShadow ? 'xs' : 'none'}
```
IrakliJani, line 18: does none work? it must be doing a lookup to the theme config dictionary. it should probably be undefined

packages/frontend/src/components/common/Page/PageHeader.tsx
IrakliJani: yeah, why not

IrakliJani: you don't need children type, FC comes with it

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
```diff
@@ -188,7 +188,7 @@ const SavedChartsHeader: FC = () => {
                 </p>
             </Alert>
 
-            <PageHeader>
+            <PageHeader withShadow>
```
IrakliJani, line 191: so, only charts have headers but not dashboards right? is there any place other than that two where we use this?

IrakliJani:
maybe it will be better with a light border, same border we have for view mode tiles? 🤔

PriPatel:
There needs to be something here under the header , otherwise it blends into the tiles. 
Screenshot 2023-08-11 at 14 07 44

packages/frontend/src/components/common/Page/PageHeader.tsx
joaoviana: it did bc it got ignored 😆

joaoviana:
@IrakliJani & @PriPatel , I've made the changes to: https://github.com/lightdash/lightdash/pull/6710/commits/a341505dc51dafc8f60826c2c7534adaf3a33542

so i've reverted the `withShadow` prop, then just changed the `shadow` value. It looks good on both the chart and dashboard headers!

packages/frontend/src/components/common/Page/PageHeader.tsx
```diff
@@ -3,24 +3,22 @@ import { FC } from 'react';
...
+    <Card
+        component={Flex}
+        justify="flex-end"
+        align="center"
+        pos="relative"
+        h={PAGE_HEADER_HEIGHT}
+        px="lg"
+        py="md"
+        bg="white"
+        shadow="0 0 0 1px #bec1c426"
```
IrakliJani, line 16: This is a highly customized shadow. Can you leave a FIXME note to indicate that this shadow should be sourced from the theme config? As of now, we haven't modified the shadow configuration in the Mantine theme and are using the default Mantine settings

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
joaoviana: This page header is only used in charts + dashboards!

packages/frontend/src/components/common/Page/PageHeader.tsx
joaoviana: 100%, you're right... I'll add a FIXME yes 👍

IrakliJani (approved):

Issue #5402, opened 2023-05-09
Migrate `project settings` forms to use Mantine


IrakliJani:
### Description

migration should cover all of the 3 cards under project settings

<img width="1243" alt="CleanShot 2023-05-09 at 13 30 18@2x" src="https://user-images.githubusercontent.com/962095/237055397-9c979b7b-9411-4e98-b51e-f62d9b5f7549.png">

use Card component from Mantine as a card wrapper.
use mantine-form for managing form state.
use either Collapse or accordion for "advanced configuration options"

- [x] general settings
- [x] warehouse connection
- [ ] dbt connection

jayanand05:
working..

PR #6712, opened 2023-08-11
feat: Migrate general settings forms to use Mantine

jayanand05:
### Description:
This PR closes one task (i.e general settings) defined in the issue #5402 

925F7233-DC27-4487-87B8-1C4441E638BB_1_201_a

### DIFF
--- a/packages/frontend/src/components/ProjectConnection/index.tsx
+++ b/packages/frontend/src/components/ProjectConnection/index.tsx
@@ -1,4 +1,4 @@
-import { Callout, H5, Intent } from '@blueprintjs/core';
+import { Callout, Intent } from '@blueprintjs/core';
 import { subject } from '@casl/ability';
 import {
     CreateWarehouseCredentials,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/ProjectConnection/index.tsx
```diff
@@ -72,17 +79,16 @@ const ProjectForm: FC<Props> = ({
             {showGeneralSettings && (
                 <SettingsGridCard>
                     <div>
-                        <H5>General settings</H5>
+                        <Title order={5}>General settings</Title>
                     </div>
 
                     <div>
-                        <Input
+                        <TextInput
```
IrakliJani, line 86: I believe this whole form is wrapped with react-form-hook library and previous import of `ReactHookForm/Input` is react-form-hook enabled component. So 99% sure this should not work. to test this, change project name, save, reload

IrakliJani (changes requested):

packages/frontend/src/components/ProjectConnection/index.tsx
jayanand05: hii @IrakliJani , so should i use mantine form hooks in place of react-form-hook ?

IrakliJani: since this is a part of bigger change - we can use react hook form and it's hooks to get it working. it can be switched to mantine-form-hook after finishing whole project settings migration. 

this should be enough:
```
const { register } = useFormContext();
<MantineTextInput [props that you have] {...register("field name that was used before on Input field")} />
```

docs: https://www.react-hook-form.com/api/useformcontext/

IrakliJani: @jayanand05 let me know if that works or if you need help!

jayanand05: @IrakliJani, I updated the code request you to look once?

joaoviana (approved):
Looks really good, thanks for your contribution @jayanand05 ! 
Tested the following:
- checked that e2es passed with duplicate PR
- can't leave empty
- project name changes if filled

PR #6713, opened 2023-08-11
feat: Migrate warehouse connections to use Mantine

jayanand05:
Closes: #5402

### Description:
This PR closes second task(i.e warehouse connection) defined in the issue #5402 

https://github.com/lightdash/lightdash/assets/113882904/da598bd8-4a60-4146-bc76-1d1aea622f58


IrakliJani:
hey @jayanand05, this PR looks great but we don't have a capacity to review massive migrations like that. can we open up separate PR's for each warehouse and close this one? we should also avoid refactoring/migrating common  components to mantine. as it's used across the codebase. I'm talking about form components that are in common package. mantine form hooks can be used instead of react form hooks to avoid using form components. let me know if you have any questions!

jayanand05:
Sure @IrakliJani , I will create separate PR's for each warehouse and will take care of the rest :)

IrakliJani:
thanks @jayanand05, appreciated 🙌

Issue #6716, opened 2023-08-11
Closing the dashboard filter editor without applying it keeps the changes


magnew:
### Description

If you make changes and close the dashboard filter editor without applying them, it keep the changes state. It can be very confusing. 

### To Reproduce

1. Make a dashboard filter
2. Edit the filter
3. Make a change
4. Click outside of the filter editor to close it without applying changes
5. Open the editor again. You will see the changes you made, but they are not applied.

PR #6717, opened 2023-08-11
fix: dashboard filters keep state when not applied

magnew:
Closes: #6716

### Description:

We weren't clearing the dashboard editor state when closing it without applying changes. This sets the state back when the editor is closed. 

Note that this is a fairly targeted fix for this bug. Ideally we should refactor the code a bit and combine ActiveFilter with DashboardFilter. This improves the situation though.


PriPatel:
awesome spot!

joaoviana (approved):

Issue #6702, opened 2023-08-11
Apply button is inactive when updating tile settings in dashboard filters


IrakliJani:
### Description

Unable to apply after updating tile settings in dashboard filter.
"Apply" button remains inactive.

Current workaround is to delete and recreate the dashboard filter.

![CleanShot 2023-08-11 at 12 24 53@2x](https://github.com/lightdash/lightdash/assets/962095/de79baba-8097-4144-b695-6287960877bf)

PR #6720, opened 2023-08-11
fix: add a tooltip when apply button is disabled

magnew:
Closes: #6702

### Description:

Adds a tooltip explaining why the apply button is disabled. 

Note that the original bug here is mostly fixed elsewhere -- this is the final part, making it clear why you are in this state.

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
@@ -283,22 +283,31 @@ const FilterConfiguration: FC<Props> = ({
                         </Tooltip>
                     )}
 
-                <Button
-                    size="xs"
-                    variant="filled"
-                    className={Classes.POPOVER2_DISMISS}
-                    disabled={
-                        !isFilterConfigurationApplyButtonEnabled(
-                            draftFilterRule,
-                        )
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -283,22 +283,31 @@ const FilterConfiguration: FC<Props> = ({
...
-                    }}
+                <Tooltip
+                    label="Filter field and value required"
+                    disabled={isFilterConfigurationApplyButtonEnabled(
+                        draftFilterRule,
+                    )}
                 >
-                    Apply
-                </Button>
+                    <Box>
```
joaoviana, line 292: we need to wrap in `Box` in case the button is `disabled` - that would mean the tooltip events wouldn't work right?

joaoviana (commented):
one q!

joaoviana (approved):

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
magnew: Yea, exactly. I don't know why they still have it that way, but this seems to work.

Issue #6072, opened 2023-06-28
Updates to the Tiles tab in the Dashboard filters modal


PriPatel:
Includes blueprint->mantine

<img width="464" alt="Screenshot 2023-07-18 at 13 41 47" src="https://github.com/lightdash/lightdash/assets/12776522/deeb9c87-b3d4-4a52-8a8d-4044aeded3bb">

- [x] When a chart has not been selected, hide the field picker. 
- [ ] Show the chart icon type next to the chart name (https://github.com/lightdash/lightdash/issues/5768)

Add a checkbox above all charts . Users can interact with this to select/deselect all the charts. 
- [x] Checkbox is checked : All charts are selected 
- [x] Checkbox is unchecked : No charts are selected
- [x] Checkbox is indeterminate : Some tiles are selected  (show number of charts eg. 2 charts selected)

mustajab-ikram:
Can I work on this @PriPatel ?

joaoviana:
> Can I work on this @PriPatel ?

@mustajab-ikram 👋  hi! 
Unfortunately I have to work on this since it's already part of a milestone that we're currently working on. Please feel free to pick other tickets 👍

PR #6732, opened 2023-08-14
refactor: create ChartIcon component

joaoviana:
Closes: N/A

### Description:

When starting work on https://github.com/lightdash/lightdash/issues/6072, more specifically on adding the icon next to the checkbox, I realised that the `getChartIcon` util would apply default styles. 

With this in mind:
* `getChartIcon` returns only the Icon for each type
* `ChartIcon` component is a shared, styled `IconBox` with the correct Icon from `getChartIcon`, utilised in many places.

### DIFF
--- a/packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
+++ b/packages/frontend/src/components/NavBar/GlobalSearch/SearchIcon.tsx
@@ -12,7 +12,7 @@ import {
 import { FC } from 'react';
 import { Link } from 'react-router-dom';
 import {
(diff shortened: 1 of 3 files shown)


magnew (approved):
Looks good! We might want to add a couple other props (I'm wondering about col9r), but we can do that when we need it.

joaoviana:
> Looks good! We might want to add a couple other props (I'm wondering about col9r), but we can do that when we need it.

100%! Maybe for the next PR on https://github.com/lightdash/lightdash/issues/6072, will see! 👍

PR #6733, opened 2023-08-14
fix: pivot table dashboard filters

IrakliJani:
Closes: 

### Description:

- before CleanShot 2023-08-14 at 14 47 40@2x
- after CleanShot 2023-08-14 at 14 48 13@2x


joaoviana (approved):
Works great! 
tested the drill into and the dashboard filters were applied to the chart!

PR #6738, opened 2023-08-14
feat: google drive dashboards

rephus:
Closes: #6673

### Description:

We upload all charts to the selected file
Each chart is a different tab in the google sheet 
Tab title is the name of the chart
if the tab already exists, it is overwritten, we don't create a new tab

Screenshot from 2023-08-14 15-43-32


hamzahc1 (commented):
- Do we need to create a new ticket to have the pop up for google auth automatically close after logging in? It looks a bit odd right now 

- Right now you can't select which charts to include on the Google Sheets export (this is on the acceptance criteria for the original issue), but I think we can also make this a separate ticket as you could view it as a configuration option, not a blocker.

rephus:
> Do you think we should add a separate sheet with this data

Yes, that is a good solution

Issue #6671, opened 2023-08-09
Google Sheet should show when data was loaded in


hamzahc1:
### Description

- [ ] Google sheet should show when data was last received/updated - maybe by updating a cell above the info?

rephus:
how about renaming the "tab" with the timestamp ? 
![image](https://github.com/lightdash/lightdash/assets/1983672/004844b3-8847-48c3-9a11-c5474a293b7b)

hamzahc1:
Ok - let's leave it as is for now, if we hear different. feedback we can change this. My sense is that people will want to rename the tab name to something that makes sense for them 🤔 but happy to leave it for now!

PR #6740, opened 2023-08-14
feat: update tab name with update timestamp

rephus:
Closes: #6671

### Description:

image

[Screencast from 14-08-23 16:14:34.webm](https://github.com/lightdash/lightdash/assets/1983672/feba2638-2581-4d6c-9968-d4e7caa37aea)


hamzahc1 (approved):
This looks good to me! (I reviewed via the video)

packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -50,6 +50,34 @@ export class GoogleDriveClient {
         });
     }
 
+    private static async changeTabTitle(
+        sheets: any,
```
joaoviana, line 54: Maybe this was answered in previous PRs, but was wondering why we're not referencing Google's types from `googleapis` - so this one would be:
```ts
sheets: sheets_v4.Sheets
```

I had to update the import to `import { google, sheets_v4 } from 'googleapis';` so I could get typescript to show me the errors btw 👍

rephus: Yes of course, great suggestion, I'll fix the types , but perhaps not in this PR but on the main PR or maybe another PR.

joaoviana: sounds good! thanks 👍

Issue #6709, opened 2023-08-11
When I have picked 'is null' or 'is not null' as an operator, I should not see the 'any value' field


PriPatel:
<img width="341" alt="Screenshot 2023-08-11 at 11 21 31" src="https://github.com/lightdash/lightdash/assets/12776522/3484fafe-66a5-4125-a6c1-19a3a20292e6">

- [ ] When:
 - The user has picked `is null` or is `not null` as the operator; 
-  **_and_** the default value toggle is off;
the user should not be able to see any value fields.

PR #6743, opened 2023-08-14
feat: hide any value input when null/not null

joaoviana:
Closes: #6709

### Description:

hide disabled input when:
* filter is `any value` 
* filter operator is of type `null` and `not null`

screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/a6e564cf-94d5-4db0-90fb-57e7443b03d6


magnew (approved):
👍

Issue #6719, opened 2023-08-11
Update the colour of the 'close' icon on dashboard filters


PriPatel:
<img width="367" alt="Screenshot 2023-08-11 at 15 44 26" src="https://github.com/lightdash/lightdash/assets/12776522/92b33961-3641-4510-a1be-aa61e4b22013">

- [ ] Update from black to gray-6
The 'close' button should not look to be at the same hierarchy as the value colour, so it will look more appropriate in grey.

PR #6750, opened 2023-08-15
fix: filter close icon color

magnew:
Closes: #6719 

### Description:

Change the filter close button color. I did this by just using the built-in mantine CloseButton

Before:
Screenshot 2023-08-15 at 15 07 14

After:
Screenshot 2023-08-15 at 15 07 26


joaoviana (approved):
lets gooo

Issue #6725, opened 2023-08-14
Closing the filter popover without clicking 'Apply' doesn't clear popover state correctly


joaoviana:
### Description

Closing the filter popover without clicking 'Apply' doesn't clear popover state correctly. We changed the component to stay mounted between interactions to avoid an extra fetch in it's tree, but now it doesn't clear state correctly when closed. Ideally the fix for this is:
1. Add a filter controller between the buttons and the form to unify state.
2. Stop fetching the subcomponent every mount and unmount the component when close. If this doesn't work, the controller  needs to carefully clear state.

PR #6754, opened 2023-08-15
fix: clear dashboard filter state when closed

magnew:
Closes: #6725 

### Description:

We were keeping the dashboard popovers mounted because they were fetching data every time they loaded, causing perf problems. This PR makes it so that they unmount (clearing state) when closed. Most of the work here is to make it so that they don't fetch data on mount. 

@IrakliJani and I talked through this and there are some longer-term things we could do to standardize this, but for now, for the relevant fetches, I added props for consumers to control the fetching behavior of the underlying useQueries.


IrakliJani (commented):
yeah, I think this all makes sense. let's wait for e2e to pass

magnew:
> yeah, I think this all makes sense. let's wait for e2e to pass

No luck. Looks like it might be a real failure. I'll take a look in the morning.

IrakliJani (approved):

Issue #6752, opened 2023-08-15
It's unclear what `Cancel` does when editing a chart from a dashboard that I just finished editing


joaoviana:
### Description

Users reported that the action-group-button section when editing a chart (that was triggered when viewing a dashboard) is confusing. 

**Screenshot reference**
See `Save Changes` and `Cancel` button
<img width="1602" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/ee00124a-db66-4c4f-86a9-8356561c55df">

### To Reproduce

* open dashboard
* click `edit chart`
* (now on chart edit mode)
* make change
* save
* see `saved changes` button now `disabled`, however `Cancel` is enabled. It's unclear what that actually will do: Cancel the changes the user made? Everything OK + go back to dashboard? 

Proposed solutions: 
1. Change copy of the button to either `Close`
2. Disable `Cancel` when there are no changes to be saved
3. Add `go back` button that will always be _on_

PR #6755, opened 2023-08-15
feat: improve chart from dashboard edit ux

joaoviana:
Closes: #6752

### Description:

When editing a chart from a dashboard: 
* See `Save changes`
* See `Cancel changes` **only** when there are unsaved changes
* Add go-back (Icon arrow back) button that goes back only, displays a prompt if the user has unsaved changes, before going back to the og dashboard. 
* Display tooltip on hover of new go-back button

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/5c30fc41-29c5-4438-b6c2-e4d7f57cd20d

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -8,7 +8,9 @@ import {
 } from '@blueprintjs/core';
 import { MenuItem2, Popover2 } from '@blueprintjs/popover2';
 import { subject } from '@casl/ability';
+import { Box, Tooltip } from '@mantine/core';
 import {
+    IconArrowBack,
     IconCheck,
     IconCirclePlus,
     IconCopy,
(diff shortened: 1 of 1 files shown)


hamzahc1:
@joaoviana - this works well but the copy `Go Back` doesn't seem right 🤔. Can we use a symbol for `Return` instead? I wouldn't use the same icon we use for "cancel changes" or undo like from dashboard filters but maybe just a backwards arrow with a tooltip on hover that says "Return to dashboard"?

joaoviana:
> @joaoviana - this works well but the copy `Go Back` doesn't seem right 🤔. Can we use a symbol for `Return` instead? I wouldn't use the same icon we use for "cancel changes" or undo like from dashboard filters but maybe just a backwards arrow with a tooltip on hover that says "Return to dashboard"? 

That sounds great @hamzahc1, the icon + tooltip approach shall be done ✔️ 
Yea thinking about it now, "go back" is quite vague and not that appropriate for this context

joaoviana:
@hamzahc1 added new icon + tooltip: 

https://github.com/lightdash/lightdash/assets/7611706/a8963ad0-37f0-4e1c-a68a-f016672f7f3f

I think it looks ok on the left of the 3-dot button, but could also be the first button, like this: 
image

hamzahc1 (approved):
Works great!

magnew (approved):
Looks good!

Issue #6747, opened 2023-08-15
Upload  data from rows directly into google sheets instead of CSV


rephus:
### Description

This will prevent from getting OOM errors

PR #6763, opened 2023-08-16
fix: upload gsheet content from rows instead of csv

rephus:
Closes: #6747

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -133,6 +133,7 @@ export enum QueryExecutionContext {
     CHART = 'chartView',
     VIEW_UNDERLYING_DATA = 'viewUnderlyingData',
     CSV = 'csvDownload',
+    GSHEETS = 'gsheets',
 }
 
 type QueryExecutionEvent = BaseTrack & {
--- a/packages/backend/src/scheduler/SchedulerClient.ts
+++ b/packages/backend/src/scheduler/SchedulerClient.ts
@@ -138,7 +138,7 @@ export class SchedulerClient {
         return { jobId: id, date };
     }
 
-    private async addGsheetsNotificationJob(
+    private async addGsheetsUploadJob(
         date: Date,
         jobGroup: string,
         scheduler: Scheduler,
@@ -151,14 +151,10 @@ export class SchedulerClient {
             scheduledTime: date,
         };
 
-        const { id } = await graphileClient.addJob(
-            'sendGsheetsNotification',
-            payload,
-            {
-                runAt: date,
-                maxAttempts: 1,
-            },
-        );
+        const { id } = await graphileClient.addJob('uploadGsheets', payload, {
+            runAt: date,
+            maxAttempts: 1,
+        });
         analytics.track({
             event: 'scheduler_notification_job.created',
             anonymousId: LightdashAnalytics.anonymousId,
@@ -293,7 +289,7 @@ export class SchedulerClient {
                 Logger.info(
                     `Creating gsheet notification jobs for scheduler ${scheduler.schedulerUuid}`,
                 );
-                const job = await this.addGsheetsNotificationJob(
+                const job = await this.addGsheetsUploadJob(
                     scheduledTime,
                     parentJobId,
                     scheduler,
--- a/packages/backend/src/scheduler/SchedulerClient.ts
+++ b/packages/backend/src/scheduler/SchedulerClient.ts
@@ -2,6 +2,7 @@ import {
     CompileProjectPayload,
     DownloadCsvPayload,
     EmailNotificationPayload,
+    GsheetsNotificationPayload,
     isSlackTarget,
     NotificationPayloadBase,
     ScheduledDeliveryPayload,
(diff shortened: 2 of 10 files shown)


packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -799,53 +804,74 @@ export const sendGsheetsNotification = async (
...
+                return acc;
+            }, []);
 
             const refreshToken = await userService.getRefreshToken(
                 scheduler.createdBy,
             );
-            const googleUploadPromises = csvUrls.map(async (cu) => {
-                const csvContent = await csv().fromFile(cu.localPath);
 
+            // We want to process all charts in sequence, so we don't load all chart results in memory
```
rephus, line 845: :kissing: :ok_hand:   Screenshot from 2023-08-16 10-16-23

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -133,6 +133,7 @@ export enum QueryExecutionContext {
     CHART = 'chartView',
     VIEW_UNDERLYING_DATA = 'viewUnderlyingData',
     CSV = 'csvDownload',
+    GSHEETS = 'gsheets',
```
owlas, line 136: good catch!

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -799,53 +804,74 @@ export const sendGsheetsNotification = async (
...
             // TODO use csv().fromStream to do stream reading/writting on gsheets
             const refreshToken = await userService.getRefreshToken(
                 scheduler.createdBy,
             );
-            await googleDriveClient.appendToSheet(
-                refreshToken,
-                gdriveId,
-                csvContent,
-            );
+            await googleDriveClient.appendToSheet(refreshToken, gdriveId, rows);
```
owlas, line 825: The arguments and types for `googleDriveClient` don't look right. It's expecting csvContent but now it gets rows. Is it writing the header correctly?

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -799,53 +804,74 @@ export const sendGsheetsNotification = async (
...
                     refreshToken,
                     gdriveId,
-                    csvContent,
+                    rows,
                     tabName,
                 );
-            });
-
-            Promise.all(googleUploadPromises);
+            }, Promise.resolve());
```
owlas, line 874: why do we manually resolve?

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -799,53 +804,74 @@ export const sendGsheetsNotification = async (
             targetType: 'gsheets',
             status: SchedulerJobStatus.STARTED,
```
owlas, line 805: this is missing an `await` is this intended?

owlas: actually in my editor I see a lot of warnings about missing awaits in this whole file

packages/backend/src/scheduler/SchedulerClient.ts
```diff
@@ -137,35 +138,69 @@ export class SchedulerClient {
         return { jobId: id, date };
     }
 
+    private async addGsheetsNotificationJob(
```
owlas, line 141: the naming is a bit confusing since it isn't a notification

packages/backend/src/scheduler/SchedulerClient.ts
```diff
@@ -137,35 +138,69 @@ export class SchedulerClient {
...
+    ) {
+        const graphileClient = await this.graphileUtils;
+
+        const payload: GsheetsNotificationPayload = {
+            schedulerUuid: scheduler.schedulerUuid,
+            jobGroup,
+            scheduledTime: date,
+        };
+
+        const { id } = await graphileClient.addJob(
```
owlas, line 154: Nice and clean!! Payload makes sense. Naming is just a bit confusing if I was looking at the job name in analytics or in the UI

owlas (changes requested):
great to see we don't need to read the csv anymore.

3 bits of feedback:

1. Missing awaits in `SchedulerTask`
2. Update `package.json` to remove the new csv reader package that was added in the previous PR
3. I'm starting to get confused by the architecture and intended order for the jobs happening on the backend. Could you write some simple docs to explain this in a readme by the scheduler entrypoint.

packages/backend/src/scheduler/SchedulerTask.ts
rephus: csvContent was Record<string,string>[] which is the same type as rows. Afaik the first row contains the headers. Note. this is before we `format rows`

packages/backend/src/scheduler/SchedulerTask.ts
rephus: I think this is how you implement a reduce of async with promises, the `acc` value is a resolved Promise type

packages/backend/src/scheduler/SchedulerClient.ts
```diff
@@ -137,35 +138,69 @@ export class SchedulerClient {
...
     ) {
+        if (!target) {
+            throw new Error('Missing target for slack or email notification');
+        }
+        if (!page) {
+            throw new Error(
+                'Missing page data for slack or email notification',
+            );
+        }
+        if (scheduler.format === SchedulerFormat.GSHEETS) {
```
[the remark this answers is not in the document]
rephus: The handler runs for all `scheduler` types and targets. but for gsheets we don't get the Page data (that downloads csvs or images (needed for slack or email "multiple" targets) ) , for gsheets we simply schedule the `gsheets` task (and do the log + analytics) , that will perform the `db query` and upload on the same task.

This way, once we add this tasks to the `scheduler dashboard` we should see
- for google upload: `handle` -> `gsheetsUpload`
- for slack/email : `handle` -> `slackNotification` + `emailNotification`

I'll write the docs anyway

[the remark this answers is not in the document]
rephus: I've added README on scheduler

packages/backend/src/scheduler/SchedulerClient.ts
rephus: I updated gsheet task name

packages/backend/src/scheduler/SchedulerTask.ts
rephus: Updated and added await for all log methods in this file . Good catch. I didn't get any warnings, I'm wondering if you have an extension I don't...

packages/backend/src/scheduler/SchedulerTask.ts
owlas: rows are `Record<string,any>` so it's not erroring, really rows should be `Record<string,unknown>` then you'd have seen a compile error here

owlas (approved):

rephus:
Unrelated e2e test errors. Merging

PR #6787, opened 2023-08-16
refactor: dashboard filter state

magnew:
Closes: #6725 

### Description:

This is a little refactor of the dashboard filters with 2 goals in mind:
1. Clean up partially duplicated code between DashboardFilter and ActiveFilter. ActiveFilter has been removed and a new Filter has been added. The logic from Active filter and the duplicated logic from Dashboard Filter have been moved to Filter. 
2. Centralize state handling among these components. All draft state is not in the popup, which gets unmounted when closed, simplifying clearing state. 

This isn't perfect -- we could abstract these things a little better, but this does a lot to combine diverging code paths and make state easier to work with in this code.


IrakliJani (approved):
tested thoroughly and looks great. code change simplifies a lot of things. thanks @magnew 
although I found two bugs which is not related to the PR (tested against analytics)

https://github.com/lightdash/lightdash/issues/6789
https://github.com/lightdash/lightdash/issues/6788

Issue #6667, opened 2023-08-09
Google Sheets deliveries should be included in my scheduled deliveries overview


hamzahc1:
### Description

Acceptance Criteria: 

- [ ] All scheduled deliveries to a Google Sheet are available in scheduled delivery manager 
- [ ] Links to the connected Google sheet are available

rephus:
> Links to the connected Google sheet are available 

Where are we adding the link for this ? 

What UI changes do we need to make ?

hamzahc1:
![Untitled](https://github.com/lightdash/lightdash/assets/14341285/d0a04538-72a5-4f01-aa19-43c274b9ab7c)

hamzahc1:
These are the changes for the overview board: 
- Icon should be Gsheets
- Destination = google sheets
- On hover - show name of Google Sheet 
- On click - navigate to Google Sheet (opens in new tab)

PR #6793, opened 2023-08-17
feat: gsheet on scheduler dashboard

rephus:
Closes: #6667

### Description:

Screenshot from 2023-08-17 09-46-37

### DIFF
--- a/packages/frontend/src/components/SchedulersView/SchedulersView.tsx
+++ b/packages/frontend/src/components/SchedulersView/SchedulersView.tsx
@@ -1,6 +1,8 @@
 import {
     getHumanReadableCronExpression,
+    isSchedulerGsheetsOptions,
     isSlackTarget,
+    SchedulerFormat,
     SchedulerWithLogs,
 } from '@lightdash/common';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/SchedulersView/SchedulersView.tsx
```diff
@@ -93,9 +105,7 @@ const Schedulers: FC<SchedulersProps> = ({
                                                 <Text color="gray.5">
                                                     Schedule type:{' '}
                                                     <Text color="white" span>
-                                                        {item.format === 'csv'
-                                                            ? 'CSV'
-                                                            : 'Image'}
+                                                        {format()}
```
owlas, line 108: how come it's a function?

owlas (approved):

packages/frontend/src/components/SchedulersView/SchedulersView.tsx
rephus: previously it was a `format===csv?csv:image` now it needs to be csv, image or google sheets, I find a function with return more readable.

PR #6795, opened 2023-08-17
feat: add organization_id attribute to schedule worker task

joaoviana:
Closes: N/A

### Description:

When running tasks on the scheduler, if there is an `organizationUuid`, pass it and set it as a span attribute.

This increases our traceability when looking at grafana logs

NOTE: upgraded `TypeScript` because narrowing down of types on an `unknown` object wasn't working on the type checker. This was fixed after 4.9  ref: https://www.reddit.com/r/typescript/comments/yv9tnk/narrowing_unknown_in_type_guard/iwew6ru/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
         "ts-jest": "^29.0.0",
         "ts-node-dev": "^2.0.0",
         "tslib": "^2.6.0",
-        "typescript": "^5.1.6"
+        "typescript": "^4.9.5"
     },
     "scripts": {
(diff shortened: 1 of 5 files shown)


owlas (approved):
if it builds and tests, merge!

joaoviana:
> if it builds and tests, merge!

just to be sure, will do a quick test on render bc of the typescript upgrade 👍

PR #6805, opened 2023-08-17
fix: compile correct project settings on project update

owlas:


joaoviana (approved):
Tested the following scenario:

* Run Jaffle shop locally
* Click test & compile 
* Successful
* Rename `Jaffle` **schema** field value to `jaffle` 
* Scheduler doesn't run, user sees success toast within seconds

With these changes it compiles on both times

rephus (approved):

Issue #6686, opened 2023-08-10
When I've populated an empty filter in 'view-mode', I can't properly reset it to be empty again


PriPatel:
### **Current behaviour:**
1. Click on a permanent filter in view-mode where the value is `any value`
2. Populate the field with a value 
3. Hit `Apply`
4. Value has been applied
5. Click on the permanent filter again
6. Hit  the `Reset` icon
7. The form resets but the change is not actually applied
8. I can't do anything now because the `Apply` button is disabled. There is no way to clear the filter.

<img width="748" alt="Screenshot 2023-08-10 at 12 07 58" src="https://github.com/lightdash/lightdash/assets/12776522/5a832d99-55a1-47d5-9ef9-5d1501fb4fa5">
I've hit reset. The form has reset, but as you can see the value is still equal to `hamzahc1`. I can't do anything now because the 'Apply' button is disabled.

### Expected behaviour:
1. Click on a permanent filter in view-mode where the value is `any value`
2. Populate the field with a value 
3. Hit `Apply`
4. Value has been applied
5. Click on the permanent filter again
6. Hit  the `Reset` icon
7. The form resets and the change is applied.
8. End

- [ ] When I hit reset, the form should reset, but the change should be applied to the end filter.

joaoviana:
This should be higher priority than the other ones I assume @PriPatel ?

PriPatel:
@joaoviana yup, its at the top of `to-do `👍

PR #6806, opened 2023-08-17
fix: any-value filters could not be reset

magnew:
Closes: #6686 

### Description:

We had a problem where filters with 'any' value could not be reset. This patch does a few different things, all necessary to get the filters resetting. Here's a little breakdown of the main things this is doing and where:
- Add 'disabled' to the revertable properties (`/utils/index.ts`)
- Make it possible to initialize a rule with no default value (`filters.ts`). 
- Remove the useEffect that was overwriting the filterRule passes into it and switch to calculating whether to show the input based on props (`FilterSettings.tsx`)
- Fix usage of useState setter so it properly bases the new state on the old one (`FilterConfiguration/index.tsx`)
- Change inputs to show the right message when there is no value.

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/FilterSettings.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/FilterSettings.tsx
@@ -39,17 +39,6 @@ const FilterSettings: FC<FilterSettingsProps> = ({
         [filterType],
     );
 
-    useEffect(() => {
-        if (!isEditMode && isFilterDisabled) {
-            onChangeFilterRule({
-                ...filterRule,
-                disabled: false,
-                values: undefined,
-                settings: undefined,
-            });
-        }
-    }, [isEditMode, isFilterDisabled, onChangeFilterRule, filterRule]);
-
     // Set default label when using revert (undo) button
     useEffect(() => {
         if (filterLabel !== '') {
@@ -66,14 +55,27 @@ const FilterSettings: FC<FilterSettingsProps> = ({
         );
     };
 
+    const showValueInput = useMemo(() => {
+        // Always show the input in view mode
+        if (!isEditMode) {
+            return true;
+        }
+        // In edit mode, only don't show input when disabled
+        if (isFilterDisabled) {
+            return false;
+        }
+        return true;
+    }, [isFilterDisabled, isEditMode]);
+
     const showAnyValueDisabledInput = useMemo(
         () =>
             isFilterDisabled &&
+            isEditMode &&
             ![FilterOperator.NULL, FilterOperator.NOT_NULL].includes(
                 filterRule.operator,
             ),
 
-        [filterRule.operator, isFilterDisabled],
+        [filterRule.operator, isFilterDisabled, isEditMode],
     );
 
     return (
@@ -157,7 +159,7 @@ const FilterSettings: FC<FilterSettingsProps> = ({
                         })}
                     />
                 )}
-                {!isFilterDisabled && (
+                {showValueInput && (
                     <filterConfig.inputs
                         popoverProps={popoverProps}
                         filterType={filterType}
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
@@ -125,6 +125,7 @@ const FilterConfiguration: FC<Props> = ({
                 createDashboardFilterRuleFromField(
                     newField,
                     availableTileFilters,
+                    !isEditMode,
                 ),
             );
             setSelectedField(newField);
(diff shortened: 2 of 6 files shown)


packages/frontend/src/components/DashboardFilter/FilterConfiguration/FilterSettings.tsx
```diff
@@ -39,17 +39,6 @@ const FilterSettings: FC<FilterSettingsProps> = ({
...
-        if (!isEditMode && isFilterDisabled) {
-            onChangeFilterRule({
-                ...filterRule,
-                disabled: false,
-                values: undefined,
-                settings: undefined,
-            });
-        }
-    }, [isEditMode, isFilterDisabled, onChangeFilterRule, filterRule]);
-
```
joaoviana, line 52: nice 😅

joaoviana:
When on view mode:
* add filter (e.g. created month)
* save
* see `is any value` on filter pill
* click filter pill
* see populated value

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/3c986842-8fa8-40c2-8203-c229289c422d

magnew:
> When on view mode:
> 
> * add filter (e.g. created month)
> * save
> * see `is any value` on filter pill
> * click filter pill
> * see populated value
> 
> Screen recording:
> 
>  view-mode-default-value.mov

Good catch. I *think* its fixed. There is a lot of surface area here.

packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx
```diff
@@ -40,7 +40,9 @@ const DefaultFilterInputs = <T extends ConditionalRule>({
     const placeholder = getPlaceholderByFilterTypeAndOperator({
         type: filterType,
         operator: rule.operator,
-        disabled: isFilterRule(rule) ? rule.disabled : undefined,
+        disabled: isFilterRule(rule)
+            ? rule.disabled && !rule.values
```
joaoviana, line 44: will `values` always be falsy in this case? what about `values` being set to `[]` 🤔

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -248,6 +261,12 @@ const FilterConfiguration: FC<Props> = ({
         ],
     );
 
+    const showApplyButton = !isFilterConfigurationApplyButtonEnabled(
```
joaoviana, line 264: This variable name should be different. 
I would go for : 
```suggestion
    const isApplyButtonDisabled = !isFilterConfigurationApplyButtonEnabled(
```

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -125,6 +125,8 @@ const FilterConfiguration: FC<Props> = ({
                 createDashboardFilterRuleFromField(
                     newField,
                     availableTileFilters,
+                    false,
+                    isCreatingNew && !isEditMode,
```
joaoviana, line 129: I would even have a variable for this: 
```suggestion
                 const isCreatingNewInViewMode =  isCreatingNew && !isEditMode,
```

packages/frontend/src/components/DashboardFilter/Filter.tsx
```diff
@@ -198,7 +198,7 @@ const Filter: FC<Props> = ({
                                 </Text>
                             </Tooltip>
                             <Text fw={400} span>
-                                {filterRule?.disabled ? (
+                                {!filterRule?.values?.length ? (
```
joaoviana, line 201: why can't we use `disabled` here? 🤔

joaoviana (changes requested):
looking great! just a few Qs on this. 

Also, need to test date filters that are of type: `in the last...` to check the values on: 
`values` and `settings`

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
magnew: Oops, I don't know how I got that backwards 😄

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
magnew: 👍

packages/frontend/src/components/DashboardFilter/Filter.tsx
magnew: Good question. I think the (somewhat unsatisfying) answer is that I did this first before fixing it so that disabled and length === 0 are always the same. I think I can put this back now. Nice catch.

packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx
magnew: Interestingly ![] is falsey. I can make this more explicit if you want.

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
magnew: Done

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
magnew: Done

packages/frontend/src/components/DashboardFilter/Filter.tsx
magnew: Done

joaoviana (approved):

Issue #6798, opened 2023-08-17
Investigate Dashboard performance issues


joaoviana:
### Description

Users are reporting slow loading times for dashboards with numerous filters.

Loading charts blocks UI thread; 
Loading of the page (assets, queries) - could be faster
Adding a filter, by selecting it on the dropdown, could be faster - `renderFilterList`?
The more the tiles, the slower the dashboard

### What problem does this solve?

Better experience for our users

joaoviana:
Users reported the experience to be slow on: 

- Initial load of a dashboard
- Editing a report (adding dimensions, metrics or table calculations)
- Adding a filter to a report (just loading the list of fields to filter on can be slow)
- Editing dashboard

PR #6808, opened 2023-08-18
fix: improve saving state on dashboards

joaoviana:
Closes: N/A

### Description:

Found this while working on #6798;

Disabled `add tile` button when dashboard is saving
Add loading state when dashboard is saving 

https://github.com/lightdash/lightdash/assets/7611706/5deb0d0e-abab-4f8a-b520-0d99e72ff01f


magnew (approved):
Nice improvement! I think we might want to 'disable' the whole dashboard when awaiting the save, but that can be future work.

PR #6809, opened 2023-08-18
chore: upgrade typescript-eslint/x to support ts 4.9

joaoviana:
Closes: N/A

### Description:

After upgrading TypeScript to 4.9 in #6795, some packages also need to be upgraded to support 4.9 as their threshold is at 4.8.

Upgraded to the version number that matches this released support: https://github.com/typescript-eslint/typescript-eslint/releases/tag/v5.45.0

This is what the FE process is logging when running `dev` 

image


owlas (approved):

rephus (approved):
Tested locally, running and linting works 

image

PR #6811, opened 2023-08-18
feat: improve render filter list performance

joaoviana:
Relates to #6798 

### Description:

Users have reported that when they click to find a field to filter a dashboard by, it takes more than 1s to get the list of grouped fields. 

This is because of an expensive method to group items by their table to then provide a better experience to our users. 

This PR covers the following: 
* converting tables.find((t) => t.name === item.table to a JS `Map` which ensures lookup is O(n), instead of O(n^2)
* memoization of `groupedItems` with React's own `useMemo` and `useCallback`
* Creation of `useGroupedItems` hook - only specific to this component, so it stays on same file

Screen-recording: 

https://github.com/lightdash/lightdash/assets/7611706/b9a634a6-40d0-4cfd-90bb-fc7e75e35423

### DIFF
--- a/packages/frontend/src/components/common/Filters/renderFilterList.tsx
+++ b/packages/frontend/src/components/common/Filters/renderFilterList.tsx
@@ -8,9 +8,16 @@ import {
 } from '@lightdash/common';
 import { Group, Text, Tooltip } from '@mantine/core';
 import { IconTable } from '@tabler/icons-react';
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useCallback, useMemo } from 'react';
 import styled from 'styled-components';
 import MantineIcon from '../MantineIcon';
+
+type GroupTable = Pick<SummaryExplore, 'name' | 'description'>;
+
+type GroupedItems<T> = {
+    group: GroupTable;
+    items: T[];
+};
 
 type MenuDividerProps = {
     $isFirst: boolean;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/Filters/renderFilterList.tsx
```diff
@@ -61,40 +68,57 @@ const StickyMenuDivider = forwardRef<HTMLLIElement, StickyMenuDividerProps>(
...
+    tables: Pick<SummaryExplore, 'name' | 'description'>[];
+};
+
+const RenderFilterList = <T extends Field | TableCalculation>({
+    items,
+    itemsParentRef,
+    renderItem,
+    tables,
+}: RenderFilterListProps<T>) => {
+    const groupedItems = useGroupedItems<T>(items, tables);
```
magnew, line 117: What does making it a hook get you? And easy place to put useMemo?

magnew (approved):
Nice. This looks like good stuff that will help. I think we're going to need more. Have you done anything to measure how much impact this has?

packages/frontend/src/components/common/Filters/renderFilterList.tsx
joaoviana: Great Q, I extracted it into its own hook just to separate the rendering logic from the formatting logic. 
happy to revert and keep it all inside the component 😄

magnew: No, don't need a change, just wondering. I don't think theres a relevant perf impact.

PR #6823, opened 2023-08-21
chore: remove deprecated github action

ZeRego:
### Description:

Github action `AsasInnab/regex-action@v1` is no longer maintained and will soon stop working. 

Screenshot 2023-08-21 at 17 33 41

The deployment url is already available in the github event so we don't need a regex.


owlas (approved):
merge if e2e tests pass

Issue #6726, opened 2023-08-14
dashboard filters not applied in drill into/underlying data for pivot tables


IrakliJani:
### Description

When using the "drill into" or "underlying data" functionality on pivot table tiles in the dashboard, the top-level dashboard filters are not being carried over to the explore tab.

dashboard
![CleanShot 2023-08-14 at 13 05 48@2x](https://github.com/lightdash/lightdash/assets/962095/84e09443-3a00-429f-9504-94c7afcea4ef)

filters after underlying data, drill-into
<img width="886" alt="CleanShot 2023-08-14 at 13 05 30@2x" src="https://github.com/lightdash/lightdash/assets/962095/e9782a36-2c97-4f52-89cb-7ff344381bbe">

joaoviana:
The work on this was reverted; Find solution that is more performant

PR #6832, opened 2023-08-22
fix: apply dashboard filters pivot tables' cells (drill-down | underlying)

joaoviana:
Closes: #6726

### Description:

* reworks fix on: https://github.com/lightdash/lightdash/pull/6733, but requests data on a higher level by adding Dashboard filters to `metricQuery` with `useSavedQueryWithDashboardFilters` 
* replaces `useSavedQuery` with `useSavedQueryWithDashboardFilters` to get necessary data from a single hook
* add `isError` data to return of `useSavedQueryWithDashboardFilters` hook

Screen recording of usage:

https://github.com/lightdash/lightdash/assets/7611706/6c1731d9-97f1-4997-937f-7e696da73ce8

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/ValueCell.tsx
+++ b/packages/frontend/src/components/common/PivotTable/ValueCell.tsx
@@ -11,7 +11,6 @@ import {
 import { getHotkeyHandler, useClipboard } from '@mantine/hooks';
 import { FC, useCallback, useEffect, useMemo, useState } from 'react';
 
-import useDashboardFiltersForExplore from '../../../hooks/dashboard/useDashboardFiltersForExplore';
 import { getColorFromRange, readableColor } from '../../../utils/colorUtils';
 import { getConditionalRuleLabel } from '../Filters/configs';
 import Cell, { CellProps } from './Cell';
@@ -28,7 +27,6 @@ interface ValueCellProps extends CellProps {
         colIndex: number,
         rowIndex: number,
     ) => Record<string, ResultValue>;
-    appliedDashboardFilters?: ReturnType<typeof useDashboardFiltersForExplore>;
 }
 
 const SMALL_TEXT_LENGTH = 30;
@@ -40,7 +38,6 @@ const ValueCell: FC<ValueCellProps> = ({
     colIndex,
     conditionalFormattings,
     getUnderlyingFieldValues,
-    appliedDashboardFilters,
     ...rest
 }) => {
     const conditionalFormatting = useMemo(() => {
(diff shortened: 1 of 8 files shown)


packages/frontend/src/components/common/PivotTable/ValueCell.tsx
```diff
@@ -110,6 +112,7 @@ const ValueCell: FC<ValueCellProps> = ({
             onCopy={handleCopy}
             onOpen={() => setIsMenuOpen(true)}
             onClose={() => setIsMenuOpen(false)}
+            appliedDashboardFilters={appliedDashboardFilters}
         >
             <Cell
```
magnew, line 117: It's definitely possible that if we need more perf in this table in the future, we'll want to stop creating one of these menus for each cell. That's a lot of dom to add for something that is duplicated everywhere.

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
```diff
@@ -7,9 +7,9 @@ import { useCallback, useMemo } from 'react';
 import { useDashboardContext } from '../../providers/DashboardProvider';
 
 const useDashboardFiltersForExplore = (
-    tileUuid: string,
+    tileUuid: string | undefined,
     explore: Explore | undefined,
-): DashboardFilters => {
+): DashboardFilters | undefined => {
```
magnew, line 12: Everything else that relies on this works with undefined? I guess so, but I thought I'd ask. That's the only real point of concern here 😄

magnew (approved):
Looks good! Looking at the surrounding code, there is definitely more we could do to make this table perform better, but this review does what I think we need to do now.

packages/frontend/src/components/common/PivotTable/ValueCell.tsx
joaoviana: 100%! such a great idea @magnew

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
joaoviana: only one was not being caught, done here: https://github.com/lightdash/lightdash/pull/6832/commits/08d1a4ea5707e43a8af81e17196ac71250edec3b

joaoviana:
need to fix this issue: 
image
when users create a pivot table outside the dashboard context 😢

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -112,6 +113,7 @@ const GridTile: FC<
                 <MetricQueryDataProvider
                     metricQuery={savedQuery?.metricQuery}
                     tableName={savedQuery?.tableName || ''}
+                    appliedDashboardFilters={dashboardFilters}
```
ZeRego, line 116: since we changed to `useSavedQueryWithDashboardFilters` I don't think we need `appliedDashboardFilters` has the filters should already be port of the `metricQuery` param.  Might be the reason why you get the duplicate filter.

joaoviana: done! updated the description with a new screen recording 👍

ZeRego (approved):

Issue #6681, opened 2023-08-09
Add UI loading state to filters


joaoviana:
### Description

When loading a dashboard the first time, users aren't given a visual indication that the filters they've applied to a dashboard are being loaded. 

### What problem does this solve?

Notifies the user that the app is working on pulling the data to render the filter pills + information. 

- [ ] Add loading state to filter container 
- [ ] Add loading state to `add filter` popover when filter data is being pulled

PR #6837, opened 2023-08-22
fix: loading state for dashboard filters

magnew:
Closes: #6681

### Description:

Add loading state to the following: 
* `Add filter` button - should be in `loading` state if dashboard filters are loading
* Add `Skeleton`s to Dashboard filter pills list whilst dashboard filters are loading

Other improvements:
* Do not show anything related to Dashboard Filters if there are no chart tiles in dashboard (added `hasChartTiles` to DashboardContext)

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/49453016-72c9-4419-9b38-f683669f4be0

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/Filter.tsx
+++ b/packages/frontend/src/components/DashboardFilter/Filter.tsx
@@ -1,7 +1,6 @@
 import {
     applyDefaultTileTargets,
     DashboardFilterRule,
-    DashboardTileTypes,
     FilterableField,
 } from '@lightdash/common';
 import { Button, CloseButton, Popover, Text, Tooltip } from '@mantine/core';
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/DashboardFilter/Filter.tsx
```diff
@@ -219,25 +212,27 @@ const Filter: FC<Props> = ({
...
-                    onSave={handelSaveChanges}
-                    // FIXME: remove this once we migrate off of Blueprint
-                    popoverProps={{
-                        onOpened: () => openSubPopover(),
-                        onOpening: () => openSubPopover(),
-                        onClose: () => closeSubPopover(),
-                        onClosing: () => closeSubPopover(),
-                    }}
-                />
+                {filterableFieldsByTileUuid && (
```
magnew, line 215: This should probably be !isPopoverDisabled. It should work the same, but it'll be less confusing to keep it symmetric like that.

magnew (commented):
I left one comment, but otherwise looks good!

I apparently can't approve cause I started this one. But I approve!

joaoviana (approved):
approving on behalf of @magnew - handed-over work

PR #6847, opened 2023-08-22
feat: remove footer from Dashboard page

joaoviana:
Closes: N/A

### Description:

Remove footer from Dashboard page.

`Home` is the only page with Footer.


hamzahc1 (approved):
Looks great!

Issue #6101, opened 2023-06-30
Booleans appear as True/False in table calculations instead of Yes/No


TuringLovesDeathMetal:
### Description

booleans appear as Yes/No in Lightdash dimensions, but as True/False in table calcs and filters. 

### Acceptance criteria

- [ ] Filters and table calculations have True/False appear as Yes/No (to match dimension formatting)

PR #6855, opened 2023-08-23
fix: format boolean values as true/false

magnew:
Closes: #6101

### Description:

Users were confused that boolean values in results were formatted as Yes/No. This changes booleans in tables and filters to be 'True' or 'False'. 

Before:
Screenshot 2023-08-23 at 14 51 27

After:
Screenshot 2023-08-23 at 14 50 32

### DIFF
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -58,7 +58,9 @@ export const currencies = [
 ];
 
 export const formatBoolean = <T>(v: T) =>
-    ['True', 'true', 'yes', 'Yes', '1', 'T'].includes(`${v}`) ? 'Yes' : 'No';
+    ['True', 'true', 'yes', 'Yes', '1', 'T'].includes(`${v}`)
+        ? 'True'
+        : 'False';
 
 export const getDateFormat = (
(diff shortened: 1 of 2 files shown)


packages/common/src/utils/formatting.ts
```diff
@@ -58,7 +58,9 @@ export const currencies = [
 ];
 
 export const formatBoolean = <T>(v: T) =>
-    ['True', 'true', 'yes', 'Yes', '1', 'T'].includes(`${v}`) ? 'Yes' : 'No';
+    ['True', 'true', 'yes', 'Yes', '1', 'T'].includes(`${v}`)
```
joaoviana, line 61: I believe you need to update tests in: `/packages/common/src/utils/formatting.test.ts`

Not sure if we need to update docs but didn't really see a reference to `yes` on our codebase!

joaoviana (changes requested):
one check to make!

packages/common/src/utils/formatting.ts
magnew: Thanks, good call!

joaoviana (approved):
Nice! Once tests (and e2e) pass, let's merge :)

Issue #6841, opened 2023-08-22
Disable action button whilst Google Sheet is exporting to not have multiple exports


hamzahc1:
### Description

Currently you cant hit the export Google Sheets button multiple times and it will trigger loads of Google Sheets exports

PR #6856, opened 2023-08-23
fix: add loading state when exporting to google sheets

joaoviana:
Closes: #6841

### Description:

Add loading state whilst google sheets is completing its scheduler job.
Every 2 seconds, refetch the csv endpoint to check if there's a successful response: data.url being defined. 
Loading state is enabled when the user clicks on the button (starts scheduler job) until getting a correct response (exporting successful)

screen recording:
https://github.com/lightdash/lightdash/assets/7611706/4eada6b7-fe6f-4efa-b894-7cab9eb1a9f1

Future PRs will include: 
* Moving Google exports hook to its own file;
* When user is exporting it will disable `csv` functionality; 
* When user is exporting from the `chart` context, we will disable the popover on results, and not allow to close the current popover (and the same for `results` context)

### DIFF
--- a/packages/frontend/src/components/ChartDownload.tsx
+++ b/packages/frontend/src/components/ChartDownload.tsx
@@ -261,7 +261,6 @@ export const ChartDownloadMenu: React.FC<ChartDownloadMenuProps> = memo(
 
                     <Popover.Dropdown>
                         <ExportSelector
-                            context="chart"
                             rows={resultsData?.rows}
                             getCsvLink={async (
                                 limit: number | null,
--- a/packages/frontend/src/components/ChartDownload.tsx
+++ b/packages/frontend/src/components/ChartDownload.tsx
@@ -12,12 +12,13 @@ import {
     ChartType,
     getCustomLabelsFromColumnProperties,
 } from '@lightdash/common';
+import { Button, Popover } from '@mantine/core';
+import { IconShare2 } from '@tabler/icons-react';
 import EChartsReact from 'echarts-for-react';
 import JsPDF from 'jspdf';
 import React, { memo, RefObject, useCallback, useState } from 'react';
+import { useIsMutating } from 'react-query';
 
-import { Button, Popover } from '@mantine/core';
-import { IconShare2 } from '@tabler/icons-react';
 import useEcharts from '../hooks/echarts/useEcharts';
 import { useApp } from '../providers/AppProvider';
 import { Can } from './common/Authorization';
@@ -218,6 +219,12 @@ interface ChartDownloadMenuProps {
 
 export const ChartDownloadMenu: React.FC<ChartDownloadMenuProps> = memo(
     ({ getCsvLink, getGsheetLink, projectUuid }) => {
+        const isGoogleSheetsExportingInResults = useIsMutating([
+            'google-sheets-results',
+        ]);
+        const isGoogleSheetsExportingInChart = useIsMutating([
+            'google-sheets-chart',
+        ]);
         const {
             chartRef,
             chartType,
@@ -231,7 +238,8 @@ export const ChartDownloadMenu: React.FC<ChartDownloadMenuProps> = memo(
                 resultsData.rows.length <= 0) ||
             !resultsData?.metricQuery ||
             chartType === ChartType.BIG_NUMBER ||
-            (chartType === ChartType.CARTESIAN && !eChartsOptions);
+            (chartType === ChartType.CARTESIAN && !eChartsOptions) ||
+            !!isGoogleSheetsExportingInResults;
 
         const { user } = useApp();
 
@@ -246,6 +254,7 @@ export const ChartDownloadMenu: React.FC<ChartDownloadMenuProps> = memo(
                 <Popover
                     {...COLLAPSABLE_CARD_POPOVER_PROPS}
                     disabled={disabled}
+                    closeOnClickOutside={!isGoogleSheetsExportingInChart}
                     position="bottom-end"
                 >
                     <Popover.Target>
@@ -261,6 +270,7 @@ export const ChartDownloadMenu: React.FC<ChartDownloadMenuProps> = memo(
 
                     <Popover.Dropdown>
                         <ExportSelector
+                            context="chart"
                             rows={resultsData?.rows}
                             getCsvLink={async (
                                 limit: number | null,
--- a/packages/frontend/src/components/ChartDownload.tsx
+++ b/packages/frontend/src/components/ChartDownload.tsx
@@ -12,12 +12,12 @@ import {
     ChartType,
     getCustomLabelsFromColumnProperties,
 } from '@lightdash/common';
-import { Button, Popover } from '@mantine/core';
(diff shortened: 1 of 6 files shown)


joaoviana:
need to fix the following: 
while the user is waiting on google sheets, we should disable the csv export option, so we don't impact the scheduler.

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
```diff
@@ -25,10 +25,28 @@ const ExportGsheets: FC<ExportGsheetProps> = memo(
...
+                return;
+            } catch (error) {
+                AppToaster.dismiss('exporting-gsheets');
+
+                showToastError({
+                    title: `Unable to upload Google Sheets`,
+                    subtitle: error?.error?.message,
+                });
+                throw error;
+            }
```
joaoviana, line 43: waits for all of these to run so that we don't get a false `onSuccess`.

If there's an error, we have to throw it so that `react-query` can call the `onError` callback

packages/frontend/src/components/ExportSelector/index.tsx
```diff
@@ -33,8 +36,13 @@ const ExportSelector: FC<
         );
     } else if (hasGoogleDrive && getGsheetLink) {
         return (
-            <Stack>
-                <Button variant="subtle" onClick={() => setExportType('csv')}>
+            <Stack spacing="xs">
+                <Button
+                    size="xs"
+                    variant="default"
+                    disabled={!!isGoogleSheetsExporting}
```
joaoviana, line 43: button should be disabled if `react-query` can detect there's a `google-sheets` query still running

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
```diff
@@ -26,49 +30,67 @@ const ExportGsheets: FC<ExportGsheetProps> = memo(
...
-                        title: `Unable to upload to Google Sheets`,
-                        subtitle: error?.error?.message,
-                    });
-                },
+        const {
+            mutateAsync: startGoogleSheetExport,
+            data: startGoogleSheetExportData,
+        } = useMutation(['google-sheets-start'], getGsheetLink, {
+            onMutate: () => {
+                showToast({
```
magnew, line 40: Does this one close automatically?

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
```diff
@@ -26,49 +30,67 @@ const ExportGsheets: FC<ExportGsheetProps> = memo(
...
 
-                    showToastError({
-                        title: `Unable to upload to Google Sheets`,
-                        subtitle: error?.error?.message,
-                    });
-                },
+        const {
+            mutateAsync: startGoogleSheetExport,
+            data: startGoogleSheetExportData,
+        } = useMutation(['google-sheets-start'], getGsheetLink, {
```
magnew, line 38: 1/2: So this one just initiates the export, shows the spinner, then we ignore it? THEN...

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
```diff
@@ -26,49 +30,67 @@ const ExportGsheets: FC<ExportGsheetProps> = memo(
...
+                AppToaster.dismiss('exporting-gsheets');
+                showToastError({
+                    title: `Unable to upload to Google Sheets`,
+                    subtitle: error?.error?.message,
+                });
+            },
+        });
+
+        const { data: exportGoogleSheetData } = useQuery<
+            ApiDownloadCsv | undefined,
```
magnew, line 63: 2/2 THEN: this one just polls for data in the right location until it sees it?

Is there a failure on timeout set up somewhere that I don't see?

magnew (commented):
Some non-blocking questions. Nice work!

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
joaoviana: Added a retry condition!

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
joaoviana: what do you mean? the toast? it opens and stays there until I tell it to be dismissed...

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
```diff
@@ -26,49 +30,67 @@ const ExportGsheets: FC<ExportGsheetProps> = memo(
...
+            },
+            refetchOnWindowFocus: false,
+            refetchOnMount: false,
+            enabled: !!startGoogleSheetExportData,
+        });
+
+        const buttonIsLoading =
+            !!startGoogleSheetExportData && !exportGoogleSheetData?.url;
+
+        // TODO: move to own hook
```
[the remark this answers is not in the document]
joaoviana: removed!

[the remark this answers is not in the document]
joaoviana: yes!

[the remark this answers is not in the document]
joaoviana: reorganised the comments, thanks for spotting it!

packages/frontend/src/components/ChartDownload.tsx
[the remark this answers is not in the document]
joaoviana: reverted! thank you p🙏

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
```diff
@@ -26,49 +30,78 @@ const ExportGsheets: FC<ExportGsheetProps> = memo(
...
+                startGoogleSheetExportData
+                    ? getCsvFileUrl(startGoogleSheetExportData)
+                    : Promise.reject({
+                          error: new Error(
+                              "Couldn't create scheduler job for google sheets export",
+                          ),
+                      }),
+            retry: (failureCount) => {
+                if (failureCount === 5) {
+                    resetStartGoogleSheetExport();
```
joaoviana, line 77: after 5 failures, reset the data on first action so this doesn't refetch anymore

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
joaoviana: exactly!

packages/frontend/src/components/Explorer/ExportGsheets/index.tsx
joaoviana: https://github.com/lightdash/lightdash/assets/7611706/275dfc8a-2f5f-441e-97d8-65e0c3fb09d2

magnew: Cool! Is 5 probably enough?

magnew (approved):
Looks good. Very cool!

Issue #6773, opened 2023-08-16
Custom Percentile Metric is missing option to set the percentile.


gordonkjlee:
### Description
it's not possible to specify the percentile value, e.g. 10th percentile, etc. when defining a custom metric with `percentile` type.

We should add `percentile` as a field where they can input an integer value between 0 - 100. THis should sit above the `filters` toggle.

### To Reproduce
1. Click on the 3-dot menu on a numeric dimension
2. Select "Add Custom Metric">"Percentile"
3. There isn't an option to define the percentile

![image](https://github.com/lightdash/lightdash/assets/7281432/e1f20a56-c1ea-466b-a3aa-4cca668270eb)

### App Version
0.744.1

TuringLovesDeathMetal:
👋 Hey @gordonkjlee 

I'm not too sure I understand what the bug is here 🤔 this sounds more like a feature request than a bug. 

Where did you hear that percentile custom metrics were an option? Or were you just expecting that they would be a type that was available?

TuringLovesDeathMetal:
ah, I see - what bit of it isn't configurable? 

We have a "standard" option for all of the custom metrics where you're able to add filters to the custom metrics, but that's the only configuration option available. 

Is there another configuration you'd want to see for percentile metrics?

TuringLovesDeathMetal:
great point 👀 

I haven't used the percentile custom metric yet, (only pre-defined in the .yml). 

Here's teh description for defining it in the .yml: https://docs.lightdash.com/references/metrics#percentile

So yep, we're missing a bit which asks what percentile you want

TuringLovesDeathMetal:
I've updated the description and added it to the backlog! Thanks for flagging this @gordonkjlee 🙏

PR #6860, opened 2023-08-23
fix: add an input for percentile in metrics

magnew:
Closes: #6773 

### Description:

Custom metric for percentile didn't have anywhere to choose the percentile. This adds an input box to choose what you want the percentile to be and plumbs that through. 

I hadn't interacted with these things much -- if anyone sees something I didn't think of or should have done differently, I'm very open to changing this.

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
+++ b/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
@@ -151,8 +151,6 @@ export const CustomMetricModal = () => {
                 percentile,
             });
 
-            console.log(data);
-
             if (isEditing && isAdditionalMetric(item)) {
                 editAdditionalMetric(
                     {
@@ -218,6 +216,8 @@ export const CustomMetricModal = () => {
                     {customMetricType === MetricType.PERCENTILE && (
                         <NumberInput
                             w={100}
+                            max={100}
+                            min={0}
                             required
                             label="Percentile"
                             {...form.getInputProps('percentile')}
--- a/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
+++ b/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
@@ -3,11 +3,13 @@ import {
     friendlyName,
     isAdditionalMetric,
     isDimension,
+    MetricType,
 } from '@lightdash/common';
 import {
     Accordion,
     Button,
     Modal,
+    NumberInput,
     Stack,
     Text,
     TextInput,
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
```diff
@@ -127,45 +136,50 @@ export const CustomMetricModal = () => {
...
+                percentile,
             });
-            showToastSuccess({
-                title: 'Custom metric added successfully',
-            });
-        }
-        toggleModal();
-    });
+
+            console.log(data);
```
joaoviana, line 154: you can remove this 👍

joaoviana:
@all-contributors add @magnew bug

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
```diff
@@ -201,6 +215,14 @@ export const CustomMetricModal = () => {
                         placeholder="Enter custom metric label"
                         {...form.getInputProps('customMetricLabel')}
                     />
+                    {customMetricType === MetricType.PERCENTILE && (
+                        <NumberInput
+                            w={100}
+                            required
```
joaoviana, line 221: I would also add: 
```
   max={100}
   min={0}
```

packages/frontend/src/components/Explorer/CustomMetricModal/utils/index.ts
```diff
@@ -95,6 +97,11 @@ export const prepareCustomMetricData = ({
             ? { round: dimension.round }
             : defaultRound;
 
+    const percentile =
+        shouldCopyFormatting && dimension.percentile
+            ? dimension.percentile
+            : metricPercentile;
```
joaoviana, line 103: wouldn't this always send `metricPercentile` `50` if it was non-percentile?

joaoviana (changes requested):
Tested the following: 
* Creating/editing chart: added custom metric; edited; added another percentile type & non-percentile custom metric;
* Added chart to dashboard; Explored, all good

Left some changes/suggestions!

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
magnew: Oops. Will do.

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
magnew: Good call.

packages/frontend/src/components/Explorer/CustomMetricModal/utils/index.ts
magnew: Hm, yea. It should be benign, but good catch. Will change.

joaoviana (approved):

Issue #6796, opened 2023-08-17
I want to refresh my dashboard data


joaoviana:
### Description

Currently users have to refresh their browser (or wait for the data to be considered stale by react-query) when viewing a dashboard 

### What problem does this solve?

Incorporating this action on the UI would bring a similar UX to the `Run Query` in charts, but in dashboards. Dashboards allow users to edit charts, tiles, add filters, etc that might affect the results in a dashboard.

A similar improvement is #3693

PR #6898, opened 2023-08-25
feat: add refresh button to dashboard

joaoviana:
Closes: #6796 

### Description:

Add Refresh button to trigger the invalidation of the following queries: 

```
{ queryKey: ['queryResults'] },
{ queryKey: ['saved_query'] },
{ queryKey: ['saved_dashboard_query', dashboardUuid] },
{ queryKey: ['dashboards', 'availableFilters'] },
```

A Chart tile is also considered to be in loading state when data is being fetched;
Invalidate filters as well so this depending data refreshes as well;

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/b6f8ba6d-806f-4253-8d09-11ee1c4ffbbe

### DIFF
--- a/packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
+++ b/packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
@@ -4,6 +4,7 @@ export const useDashboardRefresh = (dashboardUuid: string) => {
     const queryClient = useQueryClient();
 
     const queriesToRefresh = [
+        { queryKey: ['queryResults'] },
         { queryKey: ['saved_query'] },
         { queryKey: ['saved_dashboard_query', dashboardUuid] },
         { queryKey: ['dashboards', 'availableFilters'] },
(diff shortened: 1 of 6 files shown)


magnew (approved):

packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
```diff
@@ -0,0 +1,23 @@
+import { useQueries, useQueryClient } from 'react-query';
+
+export const useDashboardRefresh = (dashboardUuid: string) => {
+    const queryClient = useQueryClient();
+
+    const queriesToRefresh = [
+        { queryKey: ['saved_query'] },
```
ZeRego, line 7: I think we are missing `queryResults` in this list, otherwise is fetching the new chart but not new results.

ZeRego (changes requested):

packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
joaoviana: thanks! Just updated :)

ZeRego (approved):

PR #6899, opened 2023-08-25
chore: remove small dashboard header icon

joaoviana:
Closes: N/A

### Description:

add a `Pencil icon` on the left of the dashboard popover last-updated section that doesn't make sense 

this is without:
image


hamzahc1 (approved):
Much better! Looks great

Issue #6866, opened 2023-08-24
Update API documentation to include user register endpoints


TuringLovesDeathMetal:
### Description

See this Slack thread for more details: https://lightdash.slack.com/archives/C03C78UDJAH/p1691070452935399

Endpoints:
- Create a user without org
- Add user to org

PR #6901, opened 2023-08-25
feat: combine user registration endpoints

ZeRego:
Closes: #https://github.com/lightdash/lightdash/issues/6866 

### Description:

Main changes:

- moved **api/v1/user [Get]** to UserController
- moved and combined **api/v1/user [Post]** and **api/v1/register [Post]** to UserController
  - in the app this would correspond to the register action from `/register` and to accept an invite `/invite`

Smaller changes:

- Added email type so the API can validate it
- Amended some typing around user creation / activation ( haven't renamed all types as that would increase a lot the PR scope )
- Update e2e tests and FE with new types and endpoint


joaoviana (approved):
Looks great!

packages/common/src/types/api/email.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Email
+ * @pattern ^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$
+ */
+export type Email = string;
```
joaoviana, line 5: only thing I don't understand is how this pattern is consumed 🤔

ZeRego: Is consumed by `tsoa` lib when we run `yarn workspace backend generate-api` which generates files that validate API requests and documentation. 

Docs on tsoa @pattern annotation https://tsoa-community.github.io/docs/annotations.html

joaoviana: tyyy!!! makes sense

Issue #6853, opened 2023-08-23
Editing an in-dashboard chart while the dashboard is in edit mode loses the chart


magnew:
### Description

If you create a chart in a dashboard while the dashboard is in edit-mode, then edit the chart, the chart tile disappears. 

### To Reproduce

1. Create a dashboard
2. While in edit mode, add a tile with a chart. Save it. 
3. Verify that the chart is in the dashboard
4. Edit the chart
5. Click the 'Return to Dashboard' button
6. The chart is gone.

There are a lot of variations of moving between editing the in-dashboard chart and the dashboard that result in a similar issue (edit the chart, then click cancel, for example)

### App Version

Version: v0.755.0

magnew:
The customer also reported this related scenario to check with this fix:

1. Dashboard - 1 filter + 1 table chart. The table chart was created within the dashboard.
2. Without editing the dashboard, edit the chart 
3. Saved the chart, and I return to the dashboard 
4. No DATA at all, like the query is failing.
5. I removed the filter. Data again.
6. I added the exact same filter, data filtered again.

PR #6903, opened 2023-08-25
fix: editing an in-dashboard chart loses the chart

magnew:
Closes: #6853 

### Description:

NOTE to reviewers: There are a number of issues with charts in dashboards and I am starting with a targeted fix for this one. Things won't work perfectly with just this review but they will be better. This introduces a hook for managing these temp charts which will have more functionality with in next few reviews. 

Editing a chart in a dashboard was causing the chart to be lost. This was happening when the dashboard hadn't been saved yet. The draft state of the dashboard wasn't getting passed back to the edit experience. 

In this review I created a hook for storing the dashboard between pages and reused that hook where necessary. In the next few reviews I plan to move more of the interaction with session storage into the hook. 

Also note if you want to test it out, there is a feature flag. In the console:
`localStorage.setItem('enable-create-chart-in-dashboard', true)`

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -29,13 +29,14 @@ import {
 } from '@lightdash/common';
 import React, { FC, useCallback, useEffect, useMemo, useState } from 'react';
 import CopyToClipboard from 'react-copy-to-clipboard';
-import { useParams } from 'react-router-dom';
+import { useHistory, useParams } from 'react-router-dom';
 import { v4 as uuidv4 } from 'uuid';
 
 import { Text } from '@mantine/core';
 import { IconFolders } from '@tabler/icons-react';
 import { downloadCsv } from '../../api/csv';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -493,10 +506,22 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
                                 'manage',
                                 'SavedChart',
                             ) && (
-                                <LinkMenuItem
+                                <MenuItem2
```
ZeRego, line 509: Would it work if we kept the `LinkMenuItem` with `href` + `onClick`? So people could still right click the menu item to edit charts (especially charts that are not specific to the dashboard ) in other tabs.

ZeRego (approved):

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
magnew: I'll give it a shot

magnew: Yup, looks good. Done.

Issue #6902, opened 2023-08-25
Adding a dashboard filter on July 2023 causes the date filter to actually filter on June 30th, 2023 causing no results


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/00a2564c9fbe4747ba321a7de8f92778

### App Version

I think this might have something to do with timezones? (maybe) 

Both me and the user that was also running into this issue are in GMT + 2 (i.e. Paris/Berlin time)

IrakliJani:
same but with year picker when I select 2020 as a year
<img width="505" alt="CleanShot 2023-08-28 at 18 31 44@2x" src="https://github.com/lightdash/lightdash/assets/962095/f6070f71-0580-40a6-af84-691642324ccf">

PR #6907, opened 2023-08-28
fix: filter input year and year month timezone issue

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6902

I was able to reproduce this with selecting timezone that is ahead of local timezone

my locale: gmt+4:00
overriden: gmt+5:30

### Description:

CleanShot 2023-08-28 at 19 11 54@2x
CleanShot 2023-08-28 at 19 11 17@2x

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
@@ -92,17 +92,23 @@ const DateFilterInputs = <T extends ConditionalRule = DateFilterRule>(
                                     onClose: () =>
                                         popoverProps?.onClose?.(null as any),
                                 }}
-                                value={rule.values ? rule.values[0] : null}
+                                value={
+                                    rule.values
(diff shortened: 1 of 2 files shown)


magnew (commented):
So do we store everything in the database as UTC then? I think that's a good answer if its consistent. Are there other places where we *don't* strip off the timezone that could have similar issues?

magnew:
> @magnew I understand your concern, this solution might be what I'd call a 'duct-tape fix.
> 
> Filters are saved as a JSON blob, and I believe they are serialized in the same manner as a string using [toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString).
> 
> While there may be other places where the local date format is taken into account, in this case, we're only using parts of the date: for the year, it's 'YYYY,' and for the month, it's 'YYYY-MM.' This is evident from the SQL we generate:
> 
> CleanShot 2023-08-28 at 19 49 09@2x

Ok, seems reasonable to ship a fix that makes it better and then go after more of a complete solution next.

Issue #6905, opened 2023-08-28
Intermittent dashboard crash 


magnew:
### Description

We are seeing an intermittent dashboard crash. It's not 100% consistent, but it seems to happen most when a dashboard is reloading right after being edited. There are some console logs that are not especially helpful but look related.  

<img width="1461" alt="Screenshot 2023-08-28 at 15 00 16" src="https://github.com/lightdash/lightdash/assets/1864179/b61db29e-ef7b-4085-880d-55402b447acb">

### To Reproduce

This isn't always re-producible, but it's fairly consistent with the following steps.

1. Create a dashboard with a couple charts.
2. Edit the dashboard. 
3. Edit a chart. 
4. Change the chart name. 
5. Click 'Return to dashboard' 
6. Before the dashboard is fully reloaded, it often crashes

### App Version

Version: v0.758.0

TuringLovesDeathMetal:
Could this be related to the dashboard refresh work we shipped? (since we're only seeing it today 👀 )

magnew:
> Could this be related to the dashboard refresh work we shipped? (since we're only seeing it today 👀 )

Yes, very much.

PR #6908, opened 2023-08-28
fix: dashboard crash on refresh

magnew:
Closes: #6905 

### Description:

There was an issue with the refresh button setup where it was calling `useQueries` on the default client with existing query strings and no query functions, essentially creating a new cache for those functions. There was a race condition that caused the page to break with an unknown error when one of those was called.

This changes the pattern to just invalidate those queries in the existing client. 

This also changes the `queryResults` key to `savedChartResults`. `queryResults` doesn't exist. 

One caveat here is that the loading state on the button will now show whenever something is getting fetched on the page, not just from the refresh button. We could potentially pare that down, but I didn't figure out how to scope `isFetching` to just these queries. It will take a little more work. And it wasn't clear to me that that would be better anyway--it's possible the button should be disabled whenever a fetch is happening.

### DIFF
--- a/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
+++ b/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
@@ -96,8 +96,8 @@ const DashboardHeader = ({
         projectUuid: string;
         dashboardUuid: string;
     }>();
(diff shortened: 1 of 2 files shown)


IrakliJani (approved):
looks right 👍  - as per our chat, we just need to find a way to check isFetching scoped for that particular invalidated queries (separate issue/PR)

Issue #6904, opened 2023-08-25
Custom Percentile metric not saving correctly and defaults back to 50%


gordonkjlee:
### Description
Custom percentile metrics do not appear to be setting/saving correctly. I can create custom percentile metrics and create nice charts, but when these are revisited (e.g. refresh the page, or add it to a dashboard and see it there), all the percentiles show the median (i.e. 50th percentile).

### To Reproduce
1. Open new query
2. Create custom percentile metric at 10%
3. Save the chart (even without any metrics selected or graphed)
4. Refresh the page and inspect the custom metric again (this will now show as 50%)

### App Version
0.757.1

TuringLovesDeathMetal:
@gordonkjlee thanks for flagging this! We're fixing this now - should be out today or tomorrow 👀

PR #6914, opened 2023-08-29
fix: custom metrics percentile

IrakliJani:
Closes: #6904 

### Description:

https://github.com/lightdash/lightdash/assets/962095/ca0334f0-68a5-4cc4-bc42-0a119d03cf0a

### DIFF
--- a/packages/backend/src/database/migrations/20230829123133_add_additional_metrics_percentile_column.ts
+++ b/packages/backend/src/database/migrations/20230829123133_add_additional_metrics_percentile_column.ts
@@ -1,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.table(
+        'saved_queries_version_additional_metrics',
+        (table) => {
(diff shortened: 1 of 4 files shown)


packages/backend/src/database/migrations/20230829123133_add_additional_metrics_percentile_column.ts
```diff
@@ -0,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.table(
+        'saved_queries_version_additional_metrics',
+        (table) => {
+            table.integer('percentile');
```
joaoviana, line 7: would it leave `null` by default?

joaoviana (approved):
LGTM!

packages/backend/src/database/migrations/20230829123133_add_additional_metrics_percentile_column.ts
IrakliJani: yep

Issue #6909, opened 2023-08-28
Dashboard refresh button shows loading state for any fetch on the page


magnew:
### Description

Dashboard refresh button shows loading state for any fetch on the page, not just refresh state. Ideally it would only show loading when it is clicked. The current implementation relies on useQuery's isFetching and needs to be scoped down to just the fetches from the refresh button.

PR #6918, opened 2023-08-29
fix: dashboard refresh loader

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6909

### Description:

only listen to certain keys when observing to the loader

### DIFF
--- a/packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
+++ b/packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
@@ -1,27 +1,39 @@
-import { useIsFetching, useQueryClient } from 'react-query';
+import { useCallback, useMemo } from 'react';
+import { Query, useIsFetching, useQueryClient } from 'react-query';
+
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/dashboard/useDashboardRefresh.tsx
```diff
@@ -1,27 +1,39 @@
...
+    'saved_dashboard_query',
+    'dashboards',
+];
+
+const queryPredicate = (query: Query) => {
+    return QUERIES_TO_REFRESH.some((key) => {
+        const firstQueryKey =
+            typeof query.queryKey === 'string'
+                ? query.queryKey
+                : query.queryKey[0];
```
joaoviana, line 16: this is because there's no partial query key matching when adding a `predicate` right?

joaoviana (commented):
testing it in render soon, looks great!

joaoviana (approved):
Tested the following :
* on load of dashboard
* on edit of dashboard
* Refresh before and after

Issue #6849, opened 2023-08-22
The Run history tab shows times in 24h, everywhere else it's 12h


hamzahc1:
### Description

They incorrectly show a 24h time but also have an am/pm too. 
<img width="854" alt="CleanShot 2023-08-22 at 19 24 33@2x" src="https://github.com/lightdash/lightdash/assets/14341285/8b22064b-6e8d-416d-a3aa-c4cdf7b3b454">

Issue #6129, opened 2023-07-03
chore: replace moment with dayjs


magnew:
### Description

We're depending on moment for time calculation and manipulation, but moment is no longer maintained. We should consider switching to dayjs, which has a similar api but is more modern, faster, safer (due to active maintenance) and works better with React (because it uses immutable objects). Dayjs docs here: https://day.js.org/

If we are planning on doing a fair amount of timezone-related work in the near future, we should consider doing this first. 

### What problem does this solve?

Relying on an unmaintained library is a bit dangerous. Moment is also a common source of time-related bugs because it uses mutable objects that don't play nice with React. This will also make it easier to do some of the proposed timezone work correctly.

owlas:
I was going to close this issue since we haven't had any problems with moment. dayjs is supposed to be faster but we can't attribute any performance problems to moment currently.

However it's a good point that for security we might want to move to the maintained version. Are there known vulnerabilities in moment @magnew ?

owlas:
Closing to track with existing vulnerability scanners

PR #6923, opened 2023-08-29
fix: change scheduler time format to 12H and use dayjs

muhsinkamil:
Closes: https://github.com/lightdash/lightdash/issues/6849

### Description:

- Change scheduler time format to 12 H
- Using dayjs as https://github.com/lightdash/lightdash/issues/6129 is already open


joaoviana (approved):
Can confirm that both `moment` and `dayjs` output the same string for the same config 👍 
Screenshot 2023-08-30 at 10 30 47
Screenshot 2023-08-30 at 10 30 52

Thanks for contributing! Will do a local test and merge once done 👍

Issue #6893, opened 2023-08-24
Unable to click on link in tooltip while adding email/slack in a scheduler on the dashboard


ujwalkumar1995:
### Description

While trying to add a scheduler on a dashboard, there are options to add slack or email. There is a tooltip on both these buttons which have a hyperlink to the docs on how to do it.  As a developer I can go to the code and look at the exact hyperlink in the code but as a user there does not seem a way to click on that hyperlink. 

### To Reproduce

- Got to an existing dashboard.
- Try to add a scheduler.
- In the scheduler dialog, hover on add email or add slack option. Try to click on the hyperlinks.

Added video for reference.

https://github.com/lightdash/lightdash/assets/20976813/520cc421-8dcd-4d7d-99f1-54f1e9d9f73a

### App Version

I found this issue from the latest code on my local machine.

PR #6924, opened 2023-08-29
fix: tooltip hover effect on add email and add slack

ujwalkumar1995:
Closes: #6893 

### Description:
In add schedule delivery modal, it was not possible to click on the hyperlinks in the tooltip. Have added hoverDelay of 500ms so the tooltip stays there for some time and we are able to click on the hyperlinks. Have also made it so the links open in a new tab which felt a better UX to me. Please let me know in case of feedback. Have attached video for reference.

https://github.com/lightdash/lightdash/assets/20976813/ba7141ef-e7bf-44de-ac7b-e5d76d671646


joaoviana (approved):
looks great, thanks for contributing @ujwalkumar1995 !! 👍 
and nice addition to open them on a new tab 😄

Issue #6878, opened 2023-08-24
Charts created in dashboards don't have titles in the filter tiles tab


magnew:
### Description

Charts created in dashboards don't have titles in the filter tiles tab

PR #6937, opened 2023-08-30
fix: tile name not showing up for charts in dashes

magnew:
Closes: #6878 

### Description:

Chart names were not showing up in the tile selector for filters. Two things caused this. First,  we were fetching the project charts to get the names, which don't include charts in dashboards. Second, we were not creating the tiles with a name. This fixes both of those issues.


ZeRego:
This fixed the issue for charts that belong to a dashboard but broke the case where charts don't belong to a dashboard

Screenshot 2023-08-31 at 10 42 45

ZeRego (changes requested):
(check previous message)

magnew:
> (check previous message)

Woah! My bad. I had checked it on charts already in the dashbaord but not newly added ones.

Issue #6843, opened 2023-08-22
Date formatting isn't applied to custom metrics


TuringLovesDeathMetal:
### Description

I have a date that I've selected as a dimension, but when I try to create a custom metric from it (e.g. `MAX(date)`) the formatting results in it showing up like a timestamp in my results table. 

Here's an example of a date dimension, then the max of that dimension as a metric side-by-side:

<img width="1472" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/c9a830d9-b4b1-4032-882f-55aa3c08f0fb">

I'd expect that the metric for my date would have the same date formatting as the dimension

PR #6940, opened 2023-08-31
fix: additional metric date formatting

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6843

### Description:

respect date formatting of the base dimension in additional metrics.

CleanShot 2023-08-31 at 11 42 00@2x

### DIFF
--- a/packages/common/src/types/dbt.ts
+++ b/packages/common/src/types/dbt.ts
@@ -14,8 +14,7 @@ import {
     Source,
 } from './field';
 import { parseFilters } from './filterGrammar';
-import { AdditionalMetric } from './metricQuery';
-import { OrderFieldsByStrategy, TableBase } from './table';
+import { OrderFieldsByStrategy } from './table';
 import { TimeFrames } from './timeFrames';
 
 export enum SupportedDbtAdapter {
--- a/packages/common/src/types/dbt.ts
+++ b/packages/common/src/types/dbt.ts
@@ -411,29 +411,6 @@ export const convertColumnMetric = ({
         source,
         tableLabel,
     });
-type ConvertAdditionalMetricArgs = {
-    additionalMetric: AdditionalMetric;
-    table: TableBase;
-};
-export const convertAdditionalMetric = ({
-    additionalMetric,
-    table,
-}: ConvertAdditionalMetricArgs): Metric => {
-    const metric = convertColumnMetric({
-        modelName: table.name,
-        dimensionSql: additionalMetric.sql,
-        name: additionalMetric.name,
-        metric: { ...additionalMetric, filters: undefined },
-        tableLabel: table.label,
-    });
-
-    return {
-        ...metric,
-        ...(additionalMetric.filters && {
-            filters: additionalMetric.filters,
-        }),
-    };
-};
 
 export enum DbtManifestVersion {
     V7 = 'v7',
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -305,6 +305,8 @@ export const parseMetricType = (metricType: string): MetricType => {
             return MetricType.STRING;
         case 'date':
             return MetricType.DATE;
+        case 'timestamp':
+            return MetricType.TIMESTAMP;
         case 'boolean':
             return MetricType.BOOLEAN;
         default:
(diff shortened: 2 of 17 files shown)


packages/common/src/types/dbt.ts
```diff
@@ -411,29 +411,6 @@ export const convertColumnMetric = ({
...
-        tableLabel: table.label,
-    });
-
-    return {
-        ...metric,
-        ...(additionalMetric.filters && {
-            filters: additionalMetric.filters,
-        }),
-    };
-};
```
IrakliJani, line 436: had to extract this to avoid circular dependency cycle

packages/common/src/types/metricQuery.ts
```diff
@@ -26,6 +27,7 @@ export interface AdditionalMetric {
     baseDimensionName?: string;
     uuid?: string | null;
     percentile?: number;
+    timeInterval?: TimeFrames;
```
IrakliJani, line 30: additional metrics does not extend field so it's duplicated

packages/common/src/types/timeFrames.ts
```diff
@@ -1,15 +1,17 @@
...
-    DAY = 'DAY',
-    WEEK = 'WEEK',
-    MONTH = 'MONTH',
-    QUARTER = 'QUARTER',
+
     YEAR = 'YEAR',
+    QUARTER = 'QUARTER',
+    MONTH = 'MONTH',
+    WEEK = 'WEEK',
+    DAY = 'DAY',
```
IrakliJani, line 9: just sorting this for better readability from year → to milliseconds

packages/common/src/utils/additionalMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+    table,
+}: ConvertAdditionalMetricArgs): Metric => {
+    // dimensions with date ends with caps: DAY, MONTH, YEAR
+    // for example `order_date_DAY`, `order_date_MONTH`
+    const baseDimension = additionalMetric.baseDimensionName
+        ? Object.entries(table.dimensions).find(
+              ([name]) =>
+                  name.toLowerCase() === additionalMetric.baseDimensionName,
+          )?.[1]
+        : undefined;
```
IrakliJani, line 22: getting base dimension from an additional metric

packages/common/src/utils/additionalMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    return {
+        ...metric,
+        ...(additionalMetric.filters && {
+            filters: additionalMetric.filters,
+        }),
+        ...(baseDimension && {
+            timeInterval: baseDimension.timeInterval,
+        }),
```
IrakliJani, line 39: extracting timeInterval from base dimension

packages/common/src/utils/formatting.ts
```diff
@@ -282,28 +277,29 @@ export function formatFieldValue(
...
+                        TimeFrames.QUARTER,
+                        TimeFrames.MONTH,
+                        TimeFrames.WEEK,
+                        TimeFrames.DAY,
+                    ].includes(field.timeInterval)
+                ) {
+                    return formatDate(value, field.timeInterval, convertToUTC);
+                }
+
+                return formatTimestamp(value, field.timeInterval, convertToUTC);
```
IrakliJani, line 302: maybe it will be better to create a new formatting function which will cover both time and date?

packages/common/src/utils/additionalMetrics.ts
owlas: why do we need to lower case?

packages/common/src/types/field.ts
```diff
@@ -264,6 +264,7 @@ export enum MetricType {
     MEDIAN = 'median',
     STRING = 'string',
     DATE = 'date',
+    TIMESTAMP = 'timestamp',
```
ZeRego, line 267: We need to make sure Timestamp is handled everywhere in the codebase. Search the repo by `MetricType.DATE`.

eg: we need timestamp in this list
```
const NonAggregateMetricTypes = [
    MetricType.STRING,
    MetricType.NUMBER,
    MetricType.DATE,
    MetricType.BOOLEAN,
];
```

ZeRego (approved):
lgtm 👍

PR #6958, opened 2023-08-31
chore: fix small copy typo

joaoviana:
Closes: N/A

### Description:
small copy typo when deleting scheduled deliveries


magnew (approved):
Nice. 

Maybe a question for the future, but should we remove the period at the end too? You typically don't use em in titles.

joaoviana:
> Nice.
> 
> Maybe a question for the future, but should we remove the period at the end too? You typically don't use em in titles.

good shout! thanks @magnew , removed!

Issue #6865, opened 2023-08-24
Hitting the back button from creating a chart in a dashboard breaks the header


magnew:
### Description

The app header breaks if you push the back button from the creating a chart in dashboard experience.

PR #6963, opened 2023-08-31
fix: header breaks returning from chart edit

magnew:
Closes: #6865 

### Description:

Going back to a dashboard from editing a chart in that dashboard was breaking the header. The special header for editing charts would stay around forever. This change moves more of the interaction with sessionStorage into a hook, then uses that hook to read and clear storage items used for the header. The clear now happens in a useEffect on the dashboard page instead of before the re-direct. 

There is more to be done as far as moving things into the hook, but this fixes the header problem.

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -25,6 +25,7 @@ import React, { FC, useEffect, useState } from 'react';
 import { useHistory, useLocation, useParams } from 'react-router-dom';
 import { useToggle } from 'react-use';
 import { useChartViewStats } from '../../../hooks/chart/useChartViewStats';
+import useDashboardStorage from '../../../hooks/dashboard/useDashboardStorage';
 import {
     useDuplicateChartMutation,
     useMoveChartMutation,
@@ -81,6 +82,9 @@ const SavedChartsHeader: FC = () => {
         (context) => context.state.savedChart,
     );
     const reset = useExplorerContext((context) => context.actions.reset);
+
+    const { clearIsEditingDashboardChart } = useDashboardStorage();
+
     const [blockedNavigationLocation, setBlockedNavigationLocation] =
         useState<string>();
     const [isSaveWarningModalOpen, setIsSaveWarningModalOpen] =
@@ -527,6 +531,7 @@ const SavedChartsHeader: FC = () => {
                     opened={isMovingChart}
                     onClose={() => setIsMovingChart(false)}
                     onConfirm={() => {
+                        clearIsEditingDashboardChart();
                         history.push(
                             `/projects/${projectUuid}/saved/${savedChart.uuid}/edit`,
                         );
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -182,8 +182,6 @@ const SavedChartsHeader: FC = () => {
         history.push({
             pathname: `/projects/${savedChart?.projectUuid}/dashboards/${dashboardUuid}`,
         });
-        sessionStorage.removeItem('fromDashboard');
-        sessionStorage.removeItem('dashboardUuid');
     };
 
     const handleCancelClick = () => {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -186,7 +189,14 @@ const Dashboard: FC = () => {
...
+        // there are saved tiles, then checking for unsaved tiles,
+        // then replacing the saved tiles with the unsaved ones if they exist.
         if (savedTiles) {
+            // TODO: maybe this should move in the future, but it makes
+            // some sense here since this useEffect is essentially handling
+            // sessions storage
+            clearIsEditingDashboardChart();
             const unsavedDashboardTilesRaw = sessionStorage.getItem(
                 'unsavedDashboardTiles',
             );
```
ZeRego, line 202: Doesn't line 199 delete the storage data that is then requested in the next line ?

magnew: No. The pattern is a little confusing, but I'm working with what was already in place. There are essentially 2 pieces of information in storage 1) Whether or not we are editing a chart in a dashboard and which dashboard, used by the header and the explores page, 2) The data about the unsaved charts that are being edited. When we land on the dashboard page we need to clear #1 since we are no longer editing a chart in a dashboard. Then we pick up number #2 to add the unsaved charts to the dashboard. There is a separate clear for clearing the local storage. So we have:

- `clearIsEditingDashboardChart` -> to indicate we are no longer editing a chart in a dashboard
- `clearDashboardStorage` -> clears out all of the dashboard storage

I'm open to suggestions to make this more obvious. I did it this way partly to make it work with what's already there. I can keep my eye on cleaning up this pattern a bit. As I say in the comment, I think this useEffect is a bit wrong anyway -- it's possible the hook should just provide some way to get the stored charts and clear state as one operation. Thoughts?

ZeRego: thanks for explaining. I was mixing up the 2 clear functions.

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
```diff
@@ -529,8 +527,6 @@ const SavedChartsHeader: FC = () => {
                     opened={isMovingChart}
                     onClose={() => setIsMovingChart(false)}
                     onConfirm={() => {
-                        sessionStorage.removeItem('fromDashboard');
-                        sessionStorage.removeItem('dashboardUuid');
```
ZeRego, line 533: I haven't tested this change ( waiting for PR env to be ready ) but I suspect this is broken now. Where is the storage being cleared for this scenario?

magnew: Thats the one that gets cleared when arriving on the dashboard. 

One thing this change does is makes it so the dashboard clears the editing indicators when it's loaded. Before this change, every place that redirected back to the dashboard was clearing that indicator before the redirect (like here). But there are a lot of places you can go back to the dashboard: the save button, this button, a similar button in the toast, the browser back button. And some of them weren't doing it right so the header was broken. 

Instead of trying to keep all of those in sync, the dashboard will now clear that indicator on mount. That way it gets done correctly and you don't have to duplicate that operation everywhere. AND you never want those indicators there on the dashboard anyway.

ZeRego: But this action doesn't redirect to the dashboard. This modal/action is used when the user wants to move the chart, that belongs to a dashboard, to a space. So the header should disappear. 

Another problem comes to mind: What happens if I'm editing a dashboard -> edit chart -> move chart to space -> ?? ( I lose the dashboard unsaved changes )

magnew: Ah, I see, I thought this was the other one in this file. I will check on this one and make sure this PR doesn't break it. 

To your second point, there are a lot of these kinds of issues with unsaved changes. I can show you other ones like that 😄 . I'm trying to fix one issue at a time. This PR was for the header issue specifically, but losing unsaved changes is the next thing to tackle.

magnew: Updated to fix this.

ZeRego (approved):

Issue #6942, opened 2023-08-31
.yml metric filters don't work for non-aggregate metrics


TuringLovesDeathMetal:
### Description

If I have a non-aggregate metric, like this:

```
metrics:
            links_clicked_pct:
              type: number
              format: percent
              label: Percentage of links clicked
              description: >-
                The percentage of single- and multi-product links that have been
                clicked at least once.
              group_label: Select link metrics
              sql: ${distinct_links_clicked} / ${links_ordered}
              filters:
                - is_gmv_qualifying_order_item_status: TRUE
        ```

Then the SQL that's generated in Lightdash looks like this:

```
 CASE WHEN (("core_order_items".is_gmv_qualifying_order_item_status) IN ('true')) THEN ((COUNT(DISTINCT CASE WHEN (("core_order_items".is_gmv_qualifying_order_item_status) IN ('true') AND ("core_order_items".is_clicked) = true) THEN ("core_order_items".is_clicked) ELSE NULL END)) / (COUNT(DISTINCT CASE WHEN (("core_order_items".is_gmv_qualifying_order_item_status) IN ('true')) THEN ("core_order_items".select_link_id) ELSE NULL END))) ELSE NULL END AS "core_order_items_links_clicked_pct"
```

This causes this error

![image](https://github.com/lightdash/lightdash/assets/31848148/73f85656-f6ff-4c10-ab52-f4780b862a6b)

The problem is that the filter attached to the metric is not present as a column in the query, because it's an aggregate metric, so this error occurs. This can be worked around by building the filters into the fields that appear in the sql (i.e. in the example above it would mean adding the filter to distinct_links_clicked  and links_ordered ).

### To Reproduce

1. Add a non-aggregate metric with a filter
2. run a query with that metric

**Comment by TuringLovesDeathMetal** [comment]
Okay, we've decided that we're not going to implement filters for non-aggregate metrics because it requires adding a new CTE to the SQL and it's a way more complicated feature. Instead, we're going to:

1. Update the docs to specify that you can't use filters with non-aggregate metrics + explain a work-around (using filters in the aggregate metrics you're using in the non-aggregate metric).
2. Trigger an error message (ideally on validation/compile!) explaining that you can't use filters with non-aggregate metrics. 

Copy for the error message should be something like:

```
Error: You cannot use metric filters with non-aggregate metrics. Learn more about this in the docs [here](link to docs).
```

 cc @rephus

PR #6978, opened 2023-09-01
fix: throw error when filtering on non aggregated metric

rephus:
Closes: #6942
### Description:

Screenshot from 2023-09-01 15-41-29


TuringLovesDeathMetal (approved):
noiiiice! Looks good to me :)

Issue #6971, opened 2023-09-01
Time dimensions don't show in the results table


ZeRego:
When make a query with the dimension `"order_id__ordered_at"` we get back the column `"order_id__ordered_at__day"`. 

Since the column name doesn't match the fields we don't show that column.

![Image](https://user-images.githubusercontent.com/9117144/264996215-d688d4ba-4544-42cb-83d4-1cc9418058cb.png)

PR #6980, opened 2023-09-01
fix: add default suffix to metricflow time dimensions

ZeRego:
Closes: #6971 

### Description:

**Problem:**
If we group by a time dimension without specifying a grain we expected the results columns to match but that's not the case.

eg: for the following query
```
mutation {
  createQuery(
    environmentId: "204496"
    metrics: [{ name: "revenue"}]
    groupBy: [{ name: "order_id__ordered_at"}]
  ) {
    queryId
  }
}
```
I expect the columns:  ‘revenue’ and ‘order_id__ordered_at’  but we get:  ‘revenue’ and ‘order_id__ordered_at__day’ ( note the grain suffix )
This causes problems because there is a mismatch between the dimension name (aka ID ) the user selected and what was returned.

We had to implement a workaround by renaming all the time dimensions we receive from dbt.
`isTimeDimension ? `${name}__day` : name`
This works well because the API accepts the name with the grain suffix.
```
mutation {
  createQuery(
    environmentId: "204496"
    metrics: [{ name: "revenue"}]
    groupBy: [{ name: "order_id__ordered_at__day"}]
  ) {
    queryId
  }
}
```
But this is not in the docs so I’m unsure if it is intentional and if they are going to remove support for it.

### DIFF
--- a/packages/frontend/src/features/metricFlow/utils/convertMetricFlowFieldsToExplore.ts
+++ b/packages/frontend/src/features/metricFlow/utils/convertMetricFlowFieldsToExplore.ts
@@ -20,13 +20,14 @@ export default function convertMetricFlowFieldsToExplore(
 ) {
     const dimensionsMap = metricFlowFields.dimensions.reduce(
         (acc, { name, description, type }) => {
+            const isTimeDimension = type === MetricFlowDimensionType.TIME;
             const dimension: CompiledDimension = {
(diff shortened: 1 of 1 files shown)


joaoviana (approved):
LGTM! A bit unexpected from their side, but can always revisit once they address this 😆

Issue #6962, opened 2023-08-31
Editing a chart in a dashboard causes other unsaved charts to be removed


magnew:
### Description

Editing a chart in a dashboard causes other unsaved charts to be removed. You can essentially only edit one chart in dashboard at a time. We should consider just turning off the edit button in this case for now.

PR #6983, opened 2023-09-01
fix: dashbaord charts disappearing during edit

magnew:
Closes: #6962 

### Description:

Fixes a range of issues that could arise from editing charts *while* editing the dashboard they are in. The main thing this does is disables the ability to edit a chart while the dashboard is being edited. The options are disabled and there is an explanatory tooltip:

Screenshot 2023-09-01 at 16 54 42

This also disables the ability to delete an in-dashboard chart from the explores page. You have to go to the dashboard to do that. I kept the option in the menu and disabled it with an explanation:

Screenshot 2023-09-01 at 16 56 07

It could also be removed, but I thought that was a little confusing.

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -88,7 +88,8 @@ const SavedChartsHeader: FC = () => {
     );
     const reset = useExplorerContext((context) => context.actions.reset);
 
-    const { clearIsEditingDashboardChart } = useDashboardStorage();
+    const { clearIsEditingDashboardChart, getIsEditingDashboardChart } =
+        useDashboardStorage();
 
     const [blockedNavigationLocation, setBlockedNavigationLocation] =
         useState<string>();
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
```diff
@@ -486,14 +487,23 @@ const SavedChartsHeader: FC = () => {
...
-                                        text="Delete"
-                                        intent="danger"
-                                        onClick={() =>
-                                            setIsDeleteDialogOpen(true)
-                                        }
-                                    />
+                                    <Tooltip
+                                        disabled={!getIsEditingDashboardChart()}
+                                        position="bottom"
+                                        label="This chart can be deleted from its dashboard"
```
joaoviana, line 493: "This chart can only be deleted from the dashboard it was created from" ? IDK, what do you think?

joaoviana (commented):
I think this looks great! 

Will test locally ;)

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
magnew: I think it needs to be shorter 😄. But I don't have a much better suggestion. I will see what Priyanka thinks when she has time.

joaoviana: Cool, this LGTM, I will let @ZeRego review the rest if that's ok 👍

Issue #6906, opened 2023-08-28
Non-default filters are ignored when exporting a dashboard


christian-endres:
### Description

When exporting a dashboard as an image, the dashboard is always rendered with the default filter values.

This seems to be caused by the `UnfurlService` not appending the filter query params to the "minimal" URL in 
https://github.com/lightdash/lightdash/blob/d59cae8c23763ce2535f27b94a060b63f21cf30b/packages/backend/src/services/UnfurlService/UnfurlService.ts#L228

### To Reproduce

1. Go to a dashboard and change a filter value
2. Go to "Export dashboard"
![image](https://github.com/lightdash/lightdash/assets/126677954/5738da17-f392-4ae4-ae3b-57d9a52684ac)
3. The screenshot is rendered with default filter values

Note: The screenshot is only opened correctly if https://github.com/lightdash/lightdash/issues/6546 is fixed/worked around.

### App Version

`v0.705.1`

TuringLovesDeathMetal:
👋 Hi @christian-endres!

Thanks for opening this issue - I've added this to the backlog (as well as the other issue you've linked) and someone will look into it soon 👀

tlfbrito:
👍

Issue #6546, opened 2023-07-31
API doesn't return exported (rendered) dashboard image


christian-endres:
### Description

When exporting a dashboard as an image, the user is directed to `https://<LIGHTDASH_URL>/api/v1/slack/image/<IMAGE>`. This API route only returns images if their filename starts with`slack-image`. The `UnfurlService` does not include this prefix in the filename. This results in the API returning the following error instead of the image:
`{"status":"error","error":{"statusCode":404,"name":"NotFoundError","message":"Slack image not found <IMAGE>","data":{}}}`

These are the relevant places in code:
https://github.com/lightdash/lightdash/blob/7a73747f171dfc63790faa1be47d61ad4ee9353b/packages/backend/src/services/UnfurlService/UnfurlService.ts#L242

https://github.com/lightdash/lightdash/blob/7a73747f171dfc63790faa1be47d61ad4ee9353b/packages/backend/src/routers/slackRouter.ts#L63

### To Reproduce

1. On a dashboard, go to "Export dashboard"
![image](https://github.com/lightdash/lightdash/assets/126677954/5738da17-f392-4ae4-ae3b-57d9a52684ac)

2. Wait for a new tab to open which displays the error instead of the image.

### App Version

`v0.691.0`

PR #6984, opened 2023-09-01
fix: export dashboard with temporary filters

rephus:
Closes: #6906
Closes: #6546

### Description:

Before

[Screencast from 01-09-23 17:51:24.webm](https://github.com/lightdash/lightdash/assets/1983672/6fd3c2de-c215-41de-9ccb-a5541ac1326d)

After:

[Screencast from 01-09-23 17:52:18.webm](https://github.com/lightdash/lightdash/assets/1983672/5df9e020-7867-48d2-a279-53e2284cb1e2)

### DIFF
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -16,7 +16,7 @@ import React, {
     useState,
 } from 'react';
 import { Layout, Responsive, WidthProvider } from 'react-grid-layout';
-import { useHistory, useParams } from 'react-router-dom';
+import { useHistory, useLocation, useParams } from 'react-router-dom';
 
 import DashboardHeader from '../components/common/Dashboard/DashboardHeader';
 import ErrorState from '../components/common/ErrorState';
@@ -175,6 +175,7 @@ const Dashboard: FC = () => {
         showRedirectButton: true,
     });
     const { mutate: exportDashboard } = useExportDashboard();
+    const location = useLocation();
 
     const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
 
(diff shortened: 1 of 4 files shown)


packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -377,7 +378,7 @@ const Dashboard: FC = () => {
 
     const handleExportDashboard = () => {
         if (!dashboard) return;
-        exportDashboard(dashboard);
+        exportDashboard({ dashboard, queryFilters: location.search });
```
joaoviana, line 381: Can't we use the `allFilters` that get returned from `useDashboardContext()` here? They're of type `DashboardFilters`

Also, we remove some properties in dashboard filters' URL search params to reduce the URL size. But we then re-format these with `convertDashboardFiltersParamToDashboardFilters` (see `DashboardProvider` l:253-281

So I think that we can just use `allFilters` here instead of `location.search` 🤔

joaoviana (changes requested):
one Q on data sent to the export dashboard 👍

packages/frontend/src/pages/Dashboard.tsx
rephus: I agree the location feels a bit hacky. I want the /minimal page to behave exactly as this dashboard at this very moment. What I really want to copy is the result of this effect , https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/providers/DashboardProvider.tsx#L315.

I think using the location will be more stable than replicating any logic using allFilters, what do you think ?

joaoviana: ahh, I see, ok! 
Let's do this then, sounds good 👍 
That means that the `/minimal` page then would just do its thing with the search params and convert accordingly, I believe.

joaoviana (approved):

PR #6985, opened 2023-09-01
feat: disable csv while google sheets are exporting

joaoviana:
Closes: N/A

### Description:

`ApiDownloadCsv` also returns `status` 

We can use the above `status` to check when to disable the `csv` button while Google Sheets is exporting

https://github.com/lightdash/lightdash/assets/7611706/e1354f1e-0a1a-4f85-b1c1-ee1e06f74ca4


packages/frontend/src/components/ExportSelector/index.tsx
```diff
@@ -19,6 +25,13 @@ const ExportSelector: FC<
 
     const [exportType, setExportType] = useState<string | undefined>();
 
+    const { data } = useQuery<ApiDownloadCsv | undefined, ApiError>({
+        queryKey: [`google-sheets`],
+        enabled: false,
+    });
```
joaoviana, line 31: This is the best way I could do to subscribe to `google-sheets` changes.

`useIsFetching` is not sufficient because we want to check:
* if query is `fetching`
* if query is `refetching`
* if query has `status` `started` and when that same satus changes to something that is not started. 

`enabled` is set to `false` so the query isn't run.

rephus (approved):
LGTM

PR #6986, opened 2023-09-01
fix: remove feature flag for charts in dashboards

magnew:
### Description:

Re-enables charts in dashboards. It's in better shape now.


joaoviana (approved):
looks great!

Issue #6662, opened 2023-08-08
Get Project Ids from CLI OR pass Project Name to `--project`


tnightengale:
### Description
I would like to validate preview projects as part of CI/CD.

As I understand it, the only way to validate a specific project is using `lightdash validate --project <project_id>`. However, passing this ID in a script, non-interactively, is a headache.  

It would be nice to output the current project names and ids. Perhaps something like:
```
$ lightdash config list-projects
| Name | Project ID | Preview |
| Prod | asdfasdfaa | False |
| branchA | 123213232 | True |

etc...
```
Then at least the output could be parsed with `grep`.

Other options would be to add the `--name` flag for previews to `lightdash validate --name <preview_name>`.

### What problem does this solve?

Validate preview projects in the CI/CD without having to extract the project id from `lightdash start-preview`.

Issue #6828, opened 2023-08-22
Enforce unique project naming


joaoviana:
### Description

Users have reported that they can create a second project with the same name; Although both will have different IDs, it would be difficult for users to distinguish which one is which. 

Example: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1692634245547059

### To Reproduce

1. create project "ABC"
2. create new project also called "ABC"
3. see no errors

ujwalkumar1995:
Hey @joaoviana 
I would like to work on this issue. Could you please assign?
Also wanted to confirm one small thing from my initial analysis, I noticed that the create project api had a retry count of 3 in case of failures. Currently it retries irrespective of the kind of error thrown from backend which does not seem right.
Any suggestions if we should remove the retry count or should there be some logic on when to retry?

PR #6987, opened 2023-09-01
feat: cli set project non-interactively

tnightengale:
Closes: #6662

### Description:
Sometimes, in a CI/CD setting, it would be nice to be able to switch between projects without interaction.

This PR adds a two new options to run the Lightdash CLI `set-project` command non-interactively:
1. `--name <project_name>`
2. `--uuid <project_uuid>` 

The options are mutually exclusive. Per #6828, unique project names are not currently enforced. This feature makes no assumptions about name uniqueness, and will select the first project name that matches, in the list provided by the `GET`. 

Uuid should be used if a project has concerns about uniqueness, until #6828 is resolved.

### Use Case
Our use case, is that any step of a CI/CD job centered around a _preview_, can easily switch to that preview when performing `validate`. This is because refreshing an existing preview does not cause it to be the selected project.

Thus adding this explicit functionality seemed prudent to navigate this, and other CI/CD behaviors that may be desired in the future.

### Loom
https://www.loom.com/share/a847e3f810a244eb8835a283b02c57c2?sid=2048342c-8fcb-4c05-9345-d40f1403aaf2


packages/cli/src/index.ts
```diff
@@ -162,10 +165,15 @@ configProgram
     .description(
         'Interactively choose project.\nSee https://docs.lightdash.com/guides/cli/cli-authentication#set-active-project for more help and examples',
     )
-    .description('Interactively choose project')
     .option('--verbose', undefined, false)
-
     .action(setProjectInteractivelyHandler);
+configProgram
+    .command('set-project-by-name <projectName>')
```
IrakliJani, line 171: would be nice to have something like `set-project --uuid=uuid`

packages/cli/src/handlers/setProject.ts
```diff
@@ -49,6 +49,39 @@ export const setProjectInteractively = async () => {
...
+    });
+
+    GlobalState.debug(
+        `> Set project returned response: ${JSON.stringify(projects)}`,
+    );
+
+    if (projects.length === 0) throw new Error('No projects found.');
+
+    const selectedProject = projects.find(
+        (project) => project.name === projectName,
```
IrakliJani, line 66: project names are not unique so there can be multiple projects with the same name. should we change it to uuid?

we have a ticket about project name uniqueness: https://github.com/lightdash/lightdash/issues/6828

IrakliJani (commented):
hi @tnightengale, thanks for contributing to Lightdash ⚡ !
I added few comments related to project name uniqueness. Otherwise it looks good!

packages/cli/src/handlers/setProject.ts
tnightengale: I think #6828 is on the right track: project names should be unique. If that is put in, then I think name makes more sense. Imo, using uuid in the interface is cumbersome, because there's currently no way to fetch the uuid's. 

Whereas, projects can often be created using a unique name. In our flow for example, which inspired this feature, we are doing `lightdash start-preview --name $CI_COMMIT_REF` to create/or refresh a uniquely named preview.

packages/cli/src/index.ts
tnightengale: Yes... maybe best would be to pass _either_ `--uuid` or `--name` as an option. I initially choose this approach of adding a new subcommand because I wanted to extend, not modify. However, I like the two options as alternatives. Giving the most flexibility.

packages/cli/src/handlers/setProject.ts
IrakliJani: hi @tnightengale, I see, yeah, that makes sense.

I was thinking that we might need to make slight changes in the PR so it's consistent with what we have now. for example:

we already have `lightdash config set-project` which launches interactive project selector.

adding a `--name` there will be the best option which will not run interactive project selector and picks project by its name

`lightdash config set-project --name [project name]`

what do you think?

tnightengale: @IrakliJani Yes, I agree ✅ I'll make the change.

IrakliJani: sounds great, thanks @tnightengale 🙌

IrakliJani:
@tnightengale tested locally and works great! waiting for e2e to pass before merging

IrakliJani:
@all-contributors add @tnightengale for code

IrakliJani:
@all-contributors add @tnightengale for code

IrakliJani (approved):

IrakliJani:
@all-contributors please add @tnightengale for code

IrakliJani:
@all-contributors add @tnightengale for code

PR #6988, opened 2023-09-03
feat: Migrate project settings forms to use Mantine

jayanand05:
### Description:
This PR closes one task from Warehouse connection (i.e PostgreSQL) defined in the issue #5402

https://github.com/lightdash/lightdash/assets/113882904/35cc890c-8001-47e7-8394-6e986c8bf494


packages/frontend/src/components/ProjectConnection/WarehouseForms/PostgresForm.tsx
```diff
@@ -68,228 +72,279 @@ const PostgresForm: FC<{
...
-                                rel="noreferrer"
-                            >
-                                postgresqlco documentation
-                            </Anchor>
-                            .
-                        </p>
+                <TextInput
+                    label="User"
+                    description="This is the database user name."
+                    required
```
ZeRego, line 94: User and password are not required when editing.
```required: requireSecrets ? 'Required field' : undefined,```

File: packages/frontend/src/components/ReactHookForm/BooleanSwitch.tsx [hunk 2]
```diff
@@ -11,10 +11,8 @@ const BooleanSwitch: FC<Props> = ({ switchProps, ...rest }) => (
         {...rest}
         render={(props, { field }) => (
             <Switch
-                inline
-                large
-                innerLabelChecked="Yes"
-                innerLabel="No"
+                onLabel="Yes"
```
ZeRego, line 14: I think we need to increase the Switch size as I can't read the labels.
Now:
Screenshot 2023-09-08 at 14 10 30
Before
Screenshot 2023-09-08 at 14 12 27

packages/frontend/src/components/ProjectConnection/WarehouseForms/PostgresForm.tsx
```diff
@@ -68,228 +72,279 @@ const PostgresForm: FC<{
...
                 </FormSection>
-            </FormSection>
-            <AdvancedButtonWrapper>
-                <AdvancedButton
-                    icon={isOpen ? 'chevron-up' : 'chevron-down'}
-                    text={`Advanced configuration options`}
-                    onClick={toggleOpen}
-                />
-            </AdvancedButtonWrapper>
+                <Accordion
```
ZeRego, line 326: I think we should keep using a simple button as we don't take advantage of the accordion functionality. A button with a variant "subtle" and a left icon would simplify the code.

packages/frontend/src/components/ProjectConnection/WarehouseSettingsForm.tsx
```diff
@@ -42,7 +42,7 @@ const WarehouseSettingsForm: FC<WarehouseSettingsFormProps> = ({
 }) => {
     const warehouseType: WarehouseTypes = useWatch({
         name: 'warehouse.type',
-        defaultValue: WarehouseTypes.BIGQUERY,
+        defaultValue: WarehouseTypes.POSTGRES,
```
ZeRego, line 45: We should keep Bigquery as the default since it's the majority of the users use it.

packages/frontend/src/components/ProjectConnection/WarehouseForms/Inputs/StartOfWeekSelect.tsx
```diff
@@ -1,7 +1,8 @@
+import { Select } from '@mantine/core';
 import React, { FC } from 'react';
-import Select2 from '../../../ReactHookForm/Select2';
+import { Controller } from 'react-hook-form';
 
-const defaultOption = { value: null, label: 'Auto' };
+const defaultOption = { value: ' ', label: 'Auto' };
```
ZeRego, line 5: An empty string will cause issues later on once we enforce API validation.

Let's use `clearable` property instead 👍 
```
                <Select
                    clearable
                    placeholder="Auto"
                    label="Start of week"
                    description="Will be taken into account when using 'WEEK' time interval"
                    data={daysOfWeekOptions}
                    value={field.value}
                    onChange={field.onChange}
                    disabled={disabled}
                    dropdownPosition="top"
                />
```

ZeRego (changes requested):

jayanand05:
Hi @ZeRego, i made the necessary changes request you to look at once.

packages/frontend/src/components/ProjectConnection/WarehouseForms/Inputs/StartOfWeekSelect.tsx
ZeRego: We can remove `const defaultOption = { value: ' ', label: 'Auto' };` now.

examples/full-jaffle-shop-demo/dbt
ZeRego, line 1: I think this is rolling back the submodule. Perhaps is better to rebase this branch or rollback this change.

packages/frontend/src/components/ProjectConnection/WarehouseSettingsForm.tsx
ZeRego: I think you forgot to update this value.

ZeRego (changes requested):

ZeRego (approved):
Thanks for contributing @jayanand05 🙌 Just had to make some small adjustments.

Issue #6981, opened 2023-09-01
When I click on an item in the validator, it opens in two tabs, not one.


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/824bb50c3f3e4bce9443e48ff50b6da1

ujwalkumar1995:
Hey @TuringLovesDeathMetal 
I would like to work on this issue. Could you please assign?

PR #6992, opened 2023-09-04
fix: validator item opens two tabs

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6981

### Description:

removes unnecessary anchor inside first td

### DIFF
--- a/packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
+++ b/packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
@@ -7,6 +7,7 @@ import {
     ValidationResponse,
 } from '@lightdash/common';
 import {
+    Anchor,
     Box,
     Flex,
     Stack,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
```diff
@@ -107,63 +106,47 @@ const TableValidationItem = forwardRef<
...
-                <Anchor
-                    sx={{
-                        color: 'unset',
-                        ':hover': {
-                            color: 'unset',
-                            textDecoration: 'none',
-                        },
-                    }}
-                    href={getLinkToResource(validationError, projectUuid)}
-                    target="_blank"
```
joaoviana, line 119: I see the benefit in keeping this `Anchor`, so users can copy the row link.  Could we instead do something with the click event in `handleOnValidationErrorClick`? 

Or remove the click event and keep the `Anchor`?

joaoviana (changes requested):
one q on UX

packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
IrakliJani: @joaoviana updated PR and preserved old behavior. its not possible to semantically wrap table with anchor without doing some hacks so this should be good enough 👍

packages/frontend/src/components/SettingsValidator/ValidatorTable/index.tsx
```diff
@@ -108,15 +108,18 @@ const TableValidationItem = forwardRef<
...
                         ':hover': {
                             color: 'unset',
                             textDecoration: 'none',
                         },
                     }}
-                    href={getLinkToResource(validationError, projectUuid)}
-                    target="_blank"
+                    onClick={(e) => {
+                        e.stopPropagation();
+                    }}
```
joaoviana, line 122: that's the one!

joaoviana (approved):
thank you, merge!

Issue #6876, opened 2023-08-24
Admins should see all spaces when moving items


TuringLovesDeathMetal:
### Description

When I'm moving an item into a private space that I haven't been explicitly invited to, I can't see that space int he list of possible spaces. 

For example, here is the list I'm shown for the options of spaces I can move this dashboard to. 

<img width="1326" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/245a8797-6a67-492c-8679-d646d68f23b3">

But the space `HiKT` is not an option (even though I have access to it by default, because I'm an admin):

<img width="888" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/2a1025b7-7118-4ed2-bb21-5f24eff0c445">

This is related to this issue:
https://github.com/lightdash/lightdash/issues/6391

PR #6993, opened 2023-09-04
fix: admins should see all spaces when moving items

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6876

### Description:

CleanShot 2023-09-04 at 13 30 42@2x
CleanShot 2023-09-04 at 13 30 54@2x
CleanShot 2023-09-04 at 13 30 56@2x

### DIFF
--- a/packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
+++ b/packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
@@ -51,7 +51,7 @@ const ResourceViewActionMenu: FC<ResourceViewActionMenuProps> = ({
     const location = useLocation();
     const { projectUuid } = useParams<{ projectUuid: string }>();
     const organizationUuid = user.data?.organizationUuid;
-    const { data: spaces = [] } = useSpaceSummaries(projectUuid);
+    const { data: spaces = [] } = useSpaceSummaries(projectUuid, true);
     const isPinned = !!item.data.pinnedListUuid;
     const isDashboardPage = location.pathname.includes('/dashboards');
 


IrakliJani:
@TuringLovesDeathMetal should we distinguish "admin content spaces" somehow in the menu?
im worried about the clutter admins might get 🤔

packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
```diff
@@ -51,7 +51,7 @@ const ResourceViewActionMenu: FC<ResourceViewActionMenuProps> = ({
     const location = useLocation();
     const { projectUuid } = useParams<{ projectUuid: string }>();
     const organizationUuid = user.data?.organizationUuid;
-    const { data: spaces = [] } = useSpaceSummaries(projectUuid);
+    const { data: spaces = [] } = useSpaceSummaries(projectUuid, true);
```
joaoviana, line 54: shouldn't `true` be instead `user.data.ability.can(something-that-indicates-the-user-is-admin)`?

joaoviana (changes requested):

packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
IrakliJani: @joaoviana that part is handled on the backend

joaoviana (approved):

PriPatel:
@IrakliJani 

> should we distinguish "admin content spaces" somehow in the menu?
> im worried about the clutter admins might get 🤔

Just came here to comment on this! I think using subheadings that match the tabs in the Spaces page would help! (grouped headers, one saying 'Shared with me' and another saying 'admin content view')  We can update copy later but for now it should match the tabs in the Spaces page.

Issue #6588, opened 2023-08-02
Hiding a series in a stacked group shows the wrong total when there are other series


magnew:
### Description

Hiding a series in a stacked group shows the wrong total when there are other series. The total stays the same for the whole stack, but it should not include the hidden amount. 

### To Reproduce

1. Create a chart with two grouped series. 
2. Stack them
3. Show totals
4. Hide one of the series in a stack
5. Notice the totals dont change. 

Here's a video

https://github.com/lightdash/lightdash/assets/1864179/0572061b-5a4a-4134-bc2f-43d6ebdf1653

### App Version

Version: v0.715.0

mustajab-ikram:
hey @magnew, 
Would like to work on this, can you assign this to me?

TuringLovesDeathMetal:
Users are still reporting this as an issue

rephus:
Mmm, I can't find a way to replicate this issue, am I missing something @magnew  ? 

[Screencast from 05-10-23 10:30:49.webm](https://github.com/lightdash/lightdash/assets/1983672/ccbc8999-89ca-4350-8959-5c1fbe896321)

rephus:
Right, thanks @hamzahc1 I can replicate if the first serie is a line chart. 

![image](https://github.com/lightdash/lightdash/assets/1983672/26da73b0-5b2d-4659-b581-e5e3beefcfa5)

PR #6994, opened 2023-09-04
fix: fix total legend with multiple series

rephus:
Closes: #6588

### Description:

[Screencast from 04-09-23 11:44:39.webm](https://github.com/lightdash/lightdash/assets/1983672/cd0dfd2f-4a63-4763-b631-e37511e1cdc1)

What was happening: 

Before we were simply checking the `group/pivot` of the series (eg: shipped, completed) but when you have multiple sries, the pivot appears multiple times. so that didn't work. 

Now I'm checking the  full serie display name against the selected legend

Screenshot from 2023-09-04 11-44-33


magnew (approved):
Nice! I thought this would be more involved.

Issue #6735, opened 2023-08-14
There is a lag between hovering over a chart's information icon and seeing the accurate number of dashboards it belongs in


PriPatel:
https://www.loom.com/share/4899dc8737e24ea5abb803efa40d3137

When I hover over the information icon, I can see it says `Used in 0 dashboards` briefly before it updates to show the accurate number. This could be misleading if I'm scanning through this quickly and I might delete a chart that I believe isn't being used anywhere

ujwalkumar1995:
Hey @PriPatel 
I have analysed the issue. The data in the information icon is being fetched from an api call, that is why it is showing up as 0 until the response is received. The issue is more prominent on a very slow network. I had to throttle the api in my local setup to replicate it. 
One quick fix could be to add a loader till the api returns. Have attached video for reference. Please let me know if this works, would be happy to submit a PR.

https://github.com/lightdash/lightdash/assets/20976813/a8a7c0cc-fee6-4ac4-93a0-d931cc4f3fec

PriPatel:
Thanks @ujwalkumar1995

PR #7006, opened 2023-09-04
fix: add loader on hover of information icon on chart list item

ujwalkumar1995:
Closes: #6735 

### Description:

There is a slight lag when we hover on the information icon in dashboard list item (specially if we have a very slow network). Have added a loader on this so the user don't get confused.

Attaching video for reference.

https://github.com/lightdash/lightdash/assets/20976813/77341cb0-6893-40f9-b506-9a21fc414dc6


ZeRego (approved):
lgtm 👍 running github checks

Issue #6910, opened 2023-08-29
I want to create a default value for a user attribute


TuringLovesDeathMetal:
### Description

Right now, to assign everyone in your company to a user attribute, you'd need to manually assign each individual. Instead, I'd love to be able to set a default value for users, then I can adjust the default value after that. 

This means I only need to assign specific values to a handful of users (instead of trying to set values for every user in my instance). 

### Acceptance criteria

- [ ] Above `assign to users` there is a section called `default value` which has a toggle that's default off
<img width="656" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/e2adbf17-643c-4ec7-9b2c-d6e0aee206fd">
- [ ] If you toggle it on, then there's a text box that appears where you can enter a value
- [ ] There is an info icon beside `default value`. On hover, the text says `Use this to assign a default value for this attribute to all users in your organization. Adding specific user values below will override this default (it will not add to the default value).`

rephus:
+1 to this feature. I thought about this when we built this some time ago

PR #7008, opened 2023-09-05
feat: default user attribute value

rephus:
Closes:  #6910

### Description:

Screenshot from 2023-09-05 09-54-56
Screenshot from 2023-09-05 09-57-34
Screenshot from 2023-09-05 10-02-13


TuringLovesDeathMetal:
@rephus can we make the check-box into a toggle please! 🙏 

I asked priyanka about this once (when to use checkbox vs. toggle) and this was her response: 

> checkboxes = you can select more than 1
> toggles = you can only have one

So it would be great for us to keep this consistent :)

TuringLovesDeathMetal:
Also, the spacing here looks a bit wonky 👀 

I think the second box is closer to `description` than the first box?

image

rephus:
Like this ? @TuringLovesDeathMetal 
[Screencast from 05-09-23 12:46:48.webm](https://github.com/lightdash/lightdash/assets/1983672/fd7a248f-bb7a-4ed5-9b6a-67041a7ab04d)

TuringLovesDeathMetal:
@rephus yep, perfect!

TuringLovesDeathMetal:
The spacing is still weird. 

I think that the spacing between `Attribute name` and the text box works the best. 

The other two below this have different spacing between the text box and the header (i.e. `description` and `default value` need more space between the words and the text box below them to match `attribute name`)

image

TuringLovesDeathMetal (commented):
Sorry, one more thing - can we add docs + analytics events for this please 🙏 

If you want to take a stab at it without me speccing it, that would be great - but let me know fi you want my help. we should also do a headway update!

TuringLovesDeathMetal:
^Actually, I'll open a separate issue for docs + analytics 👀

packages/backend/src/database/migrations/20230905070644_add_default_user_attribute.ts
```diff
@@ -0,0 +1,13 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('user_attributes', (t) => {
+        t.text('default').nullable();
```
owlas, line 5: might be safer to rename this since `default` is a reserved word in sql. Maybe `attribute_default` or `default_attribute_value`

packages/backend/src/queryBuilder.ts
```diff
@@ -69,19 +69,25 @@ export const replaceUserAttributes = (
...
             (ua) => ua.name === attribute,
         );
+
         if (userAttribute === undefined) {
             throw new ForbiddenError(
                 `Missing user attribute "${attribute}" on ${filter}: "${sqlFilter}"`,
             );
         }
-        if (userAttribute.users.length !== 1) {
+        if (userAttribute.users.length !== 1 && !userAttribute.default) {
```
owlas, line 78: Should we check explicitly for null, in case we later want to support falsy values for the default?

packages/backend/src/queryBuilder.ts
```diff
@@ -69,19 +69,25 @@ export const replaceUserAttributes = (
...
             throw new ForbiddenError(
                 `Invalid or missing user attribute "${attribute}" on ${filter}: "${sqlFilter}"`,
             );
         }
+
+        const userValue =
```
owlas, line 84: nice! makes sense for the business logic to live here

packages/backend/src/models/UserAttributesModel.ts
```diff
@@ -75,7 +75,9 @@ export class UserAttributesModel {
             );
         }
         if (filters.userUuid) {
-            query.where(`users.user_uuid`, filters.userUuid);
+            query
```
owlas, line 78: this is a cool trick, I feel like it encodes a bit of business logic though. I can't think of a better approach

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -1,20 +1,18 @@
...
     userAttributes: UserAttribute[],
     attributeName: string,
     value: string,
 ) =>
-    userAttributes.some(
-        (ua) =>
-            ua.name === attributeName &&
-            ua.users.some((u) => u.value === value),
-    );
+    userAttributes.some((ua) => {
```
owlas, line 8: Is this for column level security? So we're checking if any user attribute has a matching key *and* value to the provided `attributeName` and `value` ?

Is it meant to be valid for any user or the specific user? This looks like it returns true if any user has been assigned  the argument`value`. This also duplicates the logic a bit of ```userValue || defaultValue```

owlas (changes requested):
- [ ] Remove `default` column name
- [ ] Explain `hasUserAttribute` logic

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
rephus: This is used on column level security, on query build and global search

I added a new collection of unit tests on `UserAttributeUtil` about how it works :

I think this logic is the most interesting one: 

> should be false if attribute value does not match even if default is present

packages/backend/src/database/entities/userAttributes.ts
```diff
@@ -6,7 +6,7 @@ export type DbUserAttribute = {
     name: string;
     description?: string;
     organization_id: number;
-    default: string | null;
+    attributeDefault: string | null;
```
owlas, line 9: this should be `attribute_default` right?

rephus: Updated and tested 

Screenshot from 2023-09-07 16-02-57
Screenshot from 2023-09-07 16-03-06

owlas (approved):
approved but lets improve the interface of that `hasUserAttributes` function right away

Issue #7010, opened 2023-09-05
Handle `ntz` timestamps and timestamps with a non-UTC timezone correctly


joaoviana:
### Description

When users select a `ntz` (no timezone) field, they see the values for this field formatted with a Timezone in Lightdash. 

The intention behind selecting `ntz` fields is to display raw data and there should not be any timezone associated with it. 

Example:
![image](https://github.com/lightdash/lightdash/assets/7611706/0b49a7d0-0115-4b82-a1f3-aa63d9f0403a)

joaoviana:
Reference as to why this is not that straightforward: https://github.com/lightdash/lightdash/pull/7017#issuecomment-1706957586

as suggested, we could: 
* add `ntz` to dimension configuration.
* add `timezone` to dimension configuration

ceik:
@joaoviana you mean adding a format option for ntz for dimensions like https://docs.lightdash.com/references/dimensions/#format? 

Makes sense and would solve at least my original issue.

ceik:
Just to illustrate the problem, this is how our Europe/Berlin timestamps (Snowflake NTZ timestamps) currently look in Lightdash:

![image](https://github.com/lightdash/lightdash/assets/7529722/411d69a4-b48e-4f7c-b5c3-3a6c992acce1)

sergioluka:
We are experiencing a similar issue here.
Is it possible to not apply any conversions to the timestamp fields that could be seen here:

```sql
TO_TIMESTAMP_NTZ(CONVERT_TIMEZONE('UTC', "my_table".updated_at))
```
?

jpetey75:
This closed duplicate issue had useful implementation details about a potential solution: https://github.com/lightdash/lightdash/issues/9737

jpetey75:
Closing as shipped as part of the timezones project 🎉 Two mechanisms now cover this:

A connection-level data timezone setting that declares what timezone your naive (ntz) timestamps are stored in — [#21533](https://github.com/lightdash/lightdash/pull/21533), [#25819](https://github.com/lightdash/lightdash/pull/25819)
A per-dimension `convert_timezone: false` override to display raw values with no conversion at all — [#22762](https://github.com/lightdash/lightdash/pull/22762)

More details in the [Timezones that just work](https://changelog.lightdash.com/timezones-that-just-work-341970) changelog post.

Note: per-column timezone annotation (different zones for different naive columns) isn't supported yet — if you need that, please open a new issue.

PR #7017, opened 2023-09-05
feat: do not include timezone on raw timeframes

joaoviana:
Closes: #7010

### Description:

Remove Timezone from `RAW` timestamps formatting. At least visually Users don't get confused with an irrelevant timezone on `ntz`s

Screenshot 2023-09-05 at 17 13 46

I think all makes sense 👍

### DIFF
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -115,7 +115,9 @@ const getTimeFormat = (
             timeFormat = 'HH:mm:ss:SSS';
             break;
     }
-    return `YYYY-MM-DD, ${timeFormat} (Z)`;
+    return `YYYY-MM-DD, ${timeFormat} ${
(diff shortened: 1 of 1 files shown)


joaoviana:
> `TimeFrames.RAW` doesn't mean `NTZ`
> 
> All date/time columns ( with or without a timezone ) are mapped to dimension type date/timestamp and any of them can have the `TimeFrames.RAW`.
> 
> Lightdash converts timestamps to UTC as we only support that timezone at the moment. https://github.com/lightdash/lightdash/blob/main/packages/common/src/compiler/translator.ts#L42
> 
> So if we are going to remove the timezone from the formatted value, we would have to remove it for all of them or for none. 🤔 We could support dimension timezone ( + ntz ) configuration but that requires significant work.

Understood, I'm closing this PR as the approach isn't correct. 
Moving this discussion to GH issue then :)

Issue #6414, opened 2023-07-19
Allow the user to customise the sql join type


owlas:
### Description

User wants to use `inner` join instead of the `left` join we use by default: https://lightdash.slack.com/archives/C053ESBT36V/p1689776286272479

```yaml
models:
  - name: users
    meta:
      joins: 
        - join: companies
          type: inner
          sql_on: ${users.user_id} = ${companies.user_id}
```

### Acceptance criteria

- [ ] There is a way to specify the join type in the join meta (probably `type`?)
- [ ] there are these options for join type: 
    - [ ] `inner`: corresponds to `INNER JOIN` SQL logic
    - [ ]  `full`: corresponds to `FULL OUTER JOIN` SQL logic
    - [ ]  `left`: corresponds to `LEFT OUTER JOIN`  
    - [ ] `right`: corresponds to `RIGHT OUTER JOIN`
- [ ] by default, if no `type` is defined, we use `LEFT OUTER JOIN`

### Docs

Update the `tables` reference doc to include:

- [ ] a `joins configuration` section like in metrics here: https://docs.lightdash.com/references/metrics/#metric-configuration
- [ ] This includes an example join with all of the configurations applied
- [ ] It also includes a table with all the configurations described in it
- [ ] These configurations in the table are linked to the sections already in the doc (which explain the configuration options)
- [ ] add a new configuration row + section in the docs for "Specify your join type using `type`"
- [ ] Remove the info at the top of the `joins` doc which says "all joins are LEFT OUTER joins"
- [ ] Add an info block in the `specify your join type` section which says "By default, if no `type` is specified, all joins are `LEFT OUTER` joins"

jw-aleksandr-zolotukhin:
probably makes sense to allow not only `inner`, but also all other types of joins along.

PR #7019, opened 2023-09-05
docs: join table type docs

IrakliJani:
docs for: #6414 

### Description:

- [ ] a `joins configuration` section like in metrics here: https://docs.lightdash.com/references/metrics/#metric-configuration
- [ ] This includes an example join with all of the configurations applied
- [ ] It also includes a table with all the configurations described in it
- [ ] These configurations in the table are linked to the sections already in the doc (which explain the configuration options)
- [ ] add a new configuration row + section in the docs for "Specify your join type using `type`"
- [ ] Remove the info at the top of the `joins` doc which says "all joins are LEFT OUTER joins"
- [ ] Add an info block in the `specify your join type` section which says "By default, if no `type` is specified, all joins are `LEFT OUTER` joins"

### DIFF
--- a/docs/docs/references/joins.md
+++ b/docs/docs/references/joins.md
@@ -12,13 +12,8 @@ Joins let you connect different models to each other so that you can explore mor
 
 Joins are defined at the same level as your model parameters in your YAML file.
 
-:::info
-
-All joins are defined as `LEFT OUTER` joins.
-
-:::
-
-```version: 2
+```yaml
+version: 2
 models:
   - name: users
     meta:
@@ -27,6 +22,7 @@ models:
           sql_on: ${web_sessions.user_id} = ${users.user_id}
         - join: subscriptions
           sql_on: ${subscriptions.user_id} = ${users.user_id} AND ${subscriptions.is_active}
+          type: inner
 ```
 
 When you open Lightdash, your joined models' dimensions and metrics will appear below the ones in your selected model.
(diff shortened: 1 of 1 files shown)


docs/docs/references/joins.md
```diff
@@ -78,6 +74,92 @@ Note the following important differences when aliasing models in joins:
...
+          sql_on: ${messages.sent_by} = ${users.user_id}
+```
+
+:::info
+
+By default, if no `type` is specified, all joins are `LEFT OUTER` joins.
+
+:::
+
+## Join types
```
TuringLovesDeathMetal, line 99:
```suggestion
```

TuringLovesDeathMetal, line 99:
```suggestion
```

docs/docs/references/joins.md
```diff
@@ -78,6 +74,92 @@ Note the following important differences when aliasing models in joins:
...
+    meta:
+      joins:
+        - join: subscriptions
+          sql_on: ${users.user_id} = ${subscriptions.user_id}
+          type: inner
+```
+
+### left
+
+A left join returns all rows from the left table, and the matching rows from the right table. Non-matching rows will have `NULL` for right table's columns. For example, if you have a `users` table and a `subscriptions` table, a left join would return all users, and the subscription information for users who have a subscription.
```
TuringLovesDeathMetal, line 129:
```suggestion
A left join returns all rows from the left table (i.e. the model where you're adding the `join` to), and the matching rows from the right table (i.e. the model you've specified in `- join:`). Non-matching rows will have `NULL` for right table's columns. For example, if you have a `users` table and a `subscriptions` table, a left join would return all users, and the subscription information for users who have a subscription.
```

TuringLovesDeathMetal (commented):
Just a few tiny changes, once you change those, feel free to merge!

Looks great 😍

IrakliJani:
> Just a few tiny changes, once you change those, feel free to merge!
> 
> Looks great 😍

thanks @TuringLovesDeathMetal !

Issue #4758, opened 2023-03-13
Rename .csv exports to me more helpful


TuringLovesDeathMetal:
### Description

When I export a .csv, I'd love for it to have a more helpful name than `csv-someuniqueid`

### Acceptance criteria

- [ ] If I export a .csv from a saved chart, the .csv should have the name like `name-of-chart_export-timestamp.csv`
- [ ] If I export a .csv from an unsaved chart, the .csv should have the name like `name-of-table_eport-timestamp.csv`

etadelta222:
+ 1 on this one

PR #7025, opened 2023-09-06
feat: rename CSV files

rephus:
Closes: #4758

### Description:

- [x] If I export a file from a saved chart, the file should have the name like `csv-NameOfChart_export-timestamp.csv`

Exporting `Which customers have not recently ordered an item?` chart : 

> csv-which_customers_have_not_recently_ordered_an_item_-2023-09-06-11-36-24-6970.csv

- [x] If I export a .file from an unsaved chart, the .file should have the name like `csv-NameOfTable_export-timestamp.csv`

Exporting `payments` unsaved chart or results:

> csv-payments-2023-09-06-11-36-24-6970.csv

- [x] This behaviour should also apply to exports included in scheduled deliveries

### DIFF
--- a/packages/backend/src/services/CsvService/CsvService.ts
+++ b/packages/backend/src/services/CsvService/CsvService.ts
@@ -181,6 +181,19 @@ export class CsvService {
         });
     }
 
+    static generateFileId(
+        fileName: string,
+        time: moment.Moment = moment(),
+    ): string {
+        const timestamp = time.format('YYYY-MM-DD-HH-mm-ss-SSSS');
+        const sanitizedFileName = fileName
+            .toLowerCase()
+            .replace(/[^a-z0-9]/gi, '_') // Replace non-alphanumeric characters with underscores
+            .replace(/_{2,}/g, '_'); // Replace multiple underscores with a single one
+        const fileId = `csv-${sanitizedFileName}-${timestamp}.csv`;
+        return fileId;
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/CsvService/CsvService.ts
```diff
@@ -200,7 +201,12 @@ export class CsvService {
             `writeRowsToFile with ${rows.length} rows and ${selectedFieldIds.length} columns`,
         );
 
-        const fileId = `csv-${nanoid()}.csv`;
+        const timestamp = moment().format('YYYY-MM-DD-HH-mm-ss-SSSS');
+        const sanitizedFileName = fileName
+            .toLowerCase()
+            .replace(/[^a-z0-9]/gi, '_') // Replace non-alphanumeric characters with underscores
+            .replace(/_{2,}/g, '_'); // Replace multiple underscores with a single one
+        const fileId = `csv-${sanitizedFileName}-${timestamp}.csv`;
```
joaoviana, line 209: we could have this as part of a util in `CsvService` 😉 and some tests maybe? 

but non-blocking

joaoviana (approved):

packages/backend/src/services/CsvService/CsvService.ts
rephus: Good idea. Done

Issue #7022, opened 2023-09-06
On click of `Add x` filter value for strings, the event causes the popover and all to close and no filter is applied


joaoviana:
### Description

When users click on `Add X ` filter value for strings when on dashboards, and if _under_ that button there is a clickable element (for example, a chart) then on click the popover will close and no value for the filter will be set. 

This should be an event propagation issue.

PR #7026, opened 2023-09-06
fix: only pass onOpening and onClosing props to MultiAutoComplete

joaoviana:
Closes: #7022

### Description:

Currently, users: 
* add filter to a dashboard of type string
* add operator that will trigger the MultiAutoComplete (for example `includes`)
* add value
* remove that value
* add another one. If that click happened outside of the popover, it would close the whole popover.

This was because of the `popoverProps` being spread.

Now, we only pass the necessary ones: 
`onClosing` and `onOpening`


magnew (approved):
Tested, looks good! Nice job.

PR #7029, opened 2023-09-06
chore: improve dev backend workflow with tsx

joaoviana:
Closes: N/A

### Description:

Replaces `ts-node-dev` with `esbuild`'s `tsx`: https://github.com/esbuild-kit/tsx
Spinning up dev env should be faster now


ZeRego (approved):
lgtm 👍 

Reduced  startup time, when using docker, from 3 minutes down to 1 minute 🎉

Issue #7030, opened 2023-09-06
Upgrade Node version 


joaoviana:
Node 16 is at EOL on 11/09/2023: https://endoflife.date/nodejs

Move to latest Node 20; 

This should include: 
- [x] docs updates
- [x] addition of an `.nvmrc` file

update to node version 20 on the following files:
- [x] packages/cli/src/index.ts (OPTIMIZED_NODE_VERSION)
- [x] dockerfile (L4 and L116)
- [x] dockerfile-prs (L4 and L111)

PR #7031, opened 2023-09-06
chore: add nvmrc and recommend its node version

joaoviana:
Relates to: #7030 

### Description:

Adds `.nvmrc` file with node 20 
Updates docs to reference version number

Will add more PRs later


.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
ZeRego, line 1: Is the intention to only use Node 20 for development or also for building + production? 
If so, we need to update dockerfiles as they are using node16.

joaoviana: Yes, all! https://github.com/lightdash/lightdash/issues/7030 contains the list we should follow to achieve the closure of this ticket

ZeRego:
Eng team agreed that is better to change everything in 1 PR than having multiple node versions ( 1 for development and 1 for production ).

Issue #6198, opened 2023-07-07
Existing dashboard filters should apply to tiles added to a dashboard 


hamzahc1:
### Description

If you apply a dashboard filter, then add a chart to that dashboard - the existing filters are not applied. This should happen automatically. 

For example, if I have a filter `users - date created = '2023-01-01'`, then I add a new chart which is built from an Explore which has the field `users - date created`, then by default, this new chart should have the `users - date created = ' 2023-01-01'` filter applied to it. 

### Acceptance criteria

- [x] Charts added to dashboards should have existing filters applied (where possible)
- [x] Filters are only applied if the chart is built from an explore which has the field used to define the dashboard filter 
- [x] The charts appear in the `tiles` tab of the dashboard filter with the filter applied as soon as the chart appears on the dashboard.

IrakliJani:
user comment on this:

> When creating a new dashboard, I added a table and set up several filters. All good. Later, I decided to add another table using the same model. It took me some time to realize that the filters I created earlier weren't automatically applied to this new table. I had to manually add the new table into the filter scope. I can see why it behaves this way, but it might be beneficial to either apply these filters to the new table by default or provide a warning/notification about it.

PR #7032, opened 2023-09-06
feat: dashboard filters apply to new charts

magnew:
Closes: #6198 

### Description:

This makes it so that charts added to a dashboard will have filters that affect them applied automatically. Charts using a different set of tables will not have the filters automatically applied. 

For reviewers:
- The change here is to make it so that filters automatically apply to all of the relevant charts on the dashboard by default
- This requires also making it so there is a way to disable them. The tileConfig object is already used to override the filter defaults, and this PR adds a disabled state for it. When a tile is unchecked in the list, it gets the disabled state. There is room for improvement here if we want to do it -- I left TODO about the disabled state in the code. 

What's changed in these files?
- **TileFilterConfiguration**: logic about when to apply the filter in the list of checkboxes has changed. Now it applies the filter if there is a matching field and the filter hasn't been disabled OR if there is an overriding tileConfig
- **FilterConfiguration**: now sets the tile config to the disabled config when boxes checked off. 
- **useDashboardFiltersForExplore**: now applies default filters when there is no tile config

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/TileFilterConfiguration.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/TileFilterConfiguration.tsx
@@ -15,6 +15,7 @@ import {
     Flex,
     Stack,
     Text,
+    Tooltip,
     useMantineTheme,
 } from '@mantine/core';
 import { FC, useCallback, useMemo } from 'react';
@@ -95,19 +96,21 @@ const TileFilterConfiguration: FC<Props> = ({
             const tile = tiles.find((t) => t.uuid === tileUuid);
 
             // tileConfig overrides the default filter state for a tile
-            // if it is there, we set the selected field to its
-            // stored value. If not, we use the field default
+            // if it is a field, we use that field for the filter.
+            // If it is the empty string, the filter is disabled.
             const tileConfig = filterRule.tileTargets?.[tileUuid];
-            const fieldId = tileConfig?.fieldId || getFieldId(field);
-            const selectedField = filters?.find((f) => {
-                return getFieldId(f) === fieldId;
-            });
+            const selectedField =
+                tileConfig?.fieldId !== ''
+                    ? filters?.find((f) => {
+                          return (
+                              tileConfig?.fieldId === getFieldId(f) ||
+                              matchFieldExact(f)(field)
+                          );
+                      })
+                    : undefined;
 
             const isFilterAvailable =
                 filters?.some(matchFieldByType(field)) ?? false;
-            const filterApplies =
-                (isFilterAvailable && tileConfig?.fieldId !== '') ||
-                !!tileConfig?.fieldId;
 
             const sortedFilters = filters
                 ?.filter(matchFieldByType(field))
@@ -132,7 +135,8 @@ const TileFilterConfiguration: FC<Props> = ({
             return {
                 key: tileUuid + index,
                 label: tileLabel,
-                checked: filterApplies,
+                checked: !!selectedField,
+                disabled: !isFilterAvailable,
                 tileUuid,
                 ...(tile &&
                     isDashboardChartTileType(tile) && {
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/DashboardFilter/FilterConfiguration/TileFilterConfiguration.tsx
```diff
@@ -93,10 +93,21 @@ const TileFilterConfiguration: FC<Props> = ({
...
+            // if it is there, we set the selected field to its
+            // stored value. If not, we use the field default
             const tileConfig = filterRule.tileTargets?.[tileUuid];
+            const fieldId = tileConfig?.fieldId || getFieldId(field);
+            const selectedField = filters?.find((f) => {
+                return getFieldId(f) === fieldId;
+            });
 
-            const isFilterAvailable =
-                filters?.some(matchFieldByType(field)) ?? false;
```
IrakliJani, line 99: hey @magnew , I am not sure why this was removed and logic seems little bit off.

here is how the logic worked (/supposed to work) before:

- if you pick `order.date` field and tile explore has order model then it should automatically select `order.date` field in the UI and automatically apply that filter to the tile
- if you pick `order.date` field and tile **does not** have order model but has `customer.created_date` then UI should still show input as available and customer can set an overriding field - altho it should not override automatically to another field with the same type
- if you pick `order.date` field and tile **does not** have anything in their explore with date type then the input should be disabled and override should not exist/should not be possible - filter should not apply to the tile

in the `filterApplies` logic I don't see where we check that

```
    (!!selectedField && tileConfig?.fieldId !== '') ||
                !!tileConfig?.fieldId;
```

to help with the logic I created `matchFieldExact` `matchFieldByTypeAndName` and `matchFieldByType` filters which does not seem to be utilized anymore

magnew: Oh, I remember. It looks unnecessary since the array we are looping through (sortedTileWithFilters) is already filtered using those functions. FWIW, it does seem to work how you described. Did you find a case where it doesn't? 

I will triple check.

magnew: Ok, @IrakliJani, thank you! I was wrong. I dug in again and I see what you're saying. I think I fixed it.

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -52,6 +52,16 @@ export enum FilterActions {
...
 
+// TODO: add a 'disabled' field to tileConfig.
+// For now we use this empty tile config to indicate that
+// a chart should not get the default filters applied to it.
+// A clearer solution would be to add a 'disabled' field we could
+// toggle to mark the filter as disabled for that tile.
+const DISABLE_FILTER_TILE_CONFIG = {
+    fieldId: '',
+    tableName: '',
+};
```
IrakliJani, line 63: is it going to be a lot of work if we add disabled field to the tileConfig now? im worried that if we change later we should still have a legacy code to support backward compatibility of the config

IrakliJani (approved):
approved, with a comment about backward compatibility
nice work 👍

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
magnew: That's a good question. I guess then I wonder: will it be more work later than now? What if we want to add it now? We'd have to add a migration? How hard is that to do?

IrakliJani: since it's saved as a json I don't think we'll ever write a migration for that. what we can do here is to introduce a type change. in this case is
```tsx
type DashboardFieldTarget = {
    fieldId: string;
    tableName: string;
};
```

we can change this to

```tsx
type DashboardFieldTarget = {
  fieldId: string;
  tableName: string;
} | {
  disabled: true;
}

I think this should work and is a fairly straightforward change

rephus:
Let's ship it ! :sheep:

magnew:
> Let's ship it ! 🐑

🐑 I think it's ready now.

Issue #6678, opened 2023-08-09
Ability to attach PDF of Dashboard in scheduled deliveries emails


IrakliJani:
### Description

When large dashboards are scheduled for delivery via email, the embedded image's text can be difficult to read without expanding the email. 

### What problem does this solve?

To enhance the readability and accessibility for users, it would be beneficial to provide an option to include the same dashboard image as a PDF attachment within the same email. This would offer users an alternative way to view the dashboard with better clarity.

### Acceptance criteria:

- [ ] If a user selects `image` as the export option, there is a toggle below the format which allows you to `Also include image as PDF attachment in email`
<img width="182" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/233474a1-7dbd-4901-b201-54a749f29871">

- [ ] by default, this toggle is `off` 
- [ ] This is only for email to start with (we can open a separate issue for Slack)

rephus:
Backend merged, frontend will be done on a separate PR

PR #7048, opened 2023-09-07
feat: UI for pdf image delivery option

magnew:
Closes: #6678

### Description:

UI for choosing to send a PDF in a scheduled delivery. Backend is already checked in. 

I followed the existing pattern in this form, but I don't really like the way we are integrating with useForm. We could make this better.

### DIFF
--- a/packages/frontend/src/components/SchedulerModals/SchedulerModalBase/SchedulerForm.tsx
+++ b/packages/frontend/src/components/SchedulerModals/SchedulerModalBase/SchedulerForm.tsx
@@ -12,7 +12,7 @@ import {
     CreateSchedulerAndTargetsWithoutIds,
     SchedulerFormat,
 } from '@lightdash/common';
-import { Anchor, Box, Tooltip } from '@mantine/core';
+import { Anchor, Box, Switch, Tooltip } from '@mantine/core';
 import React, { FC, useEffect, useMemo, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import useHealth from '../../../hooks/health/useHealth';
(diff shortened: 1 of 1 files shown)


hamzahc1:
I'm a bit confused by this UI: 
CleanShot 2023-09-07 at 21 26 10@2x

Is the option now for the user: 
- Deliver an image 
- Deliver an image and a PDF? 

Why would I want the image and the PDF? wouldn't it be one or the other ?

magnew:
> I'm a bit confused by this UI: CleanShot 2023-09-07 at 21 26 10@2x
> 
> Is the option now for the user:
> 
> * Deliver an image
> * Deliver an image and a PDF?
> 
> Why would I want the image and the PDF? wouldn't it be one or the other ?

It does indeed deliver the image as a PDF *and* an image in the email. The main request for this was from users that were having issues with images being stripped out of emails by their email servers. But it's a good point, it's a bit redundant. @TuringLovesDeathMetal, I think you were talking to the customer. This was the intent, right?

TuringLovesDeathMetal:
Ya, this was the intent...but I do understand what hamzah means...

What about we make the toggle `Deliver image as PDF attachment` so it's either embedded, or a PDF attachment, but not both. 

Sorry, to change this, @magnew

PriPatel:
To keep the interface less cluttered, there could instead be 3 options in the dropdown for format, rather than an additional toggle.
- csv
- Image (pdf)
- image (jpeg)

TuringLovesDeathMetal:
^Priyanka to the rescue. Ya, I'm down for that suggestion!

rephus:
^ (having a new format / choose between image OR pdf ) introduces some extra complexity on the scheduler backend logic. 

But if you're happy with this approach, we can change it

Remember we have another ticket to `add images/csv as an attachment` instead of using links, not sure if this new design is compatible .

rephus (approved):
Code looks good to me, fix testes and then we can merge (we will revisit this UI in the future if we need to )

magnew:
> Code looks good to me, fix testes and then we can merge (we will revisit this UI in the future if we need to )

@hamzahc1, @PriPatel, It sounds like we talked about this in person last week and we'd rather check this in an have some solution working for people than wait to do the extra work. I'm going to merge this now and we'll revisit with the upcoming change Javi mentioned.

Issue #7016, opened 2023-09-05
Docs + analytics events for default attribute


TuringLovesDeathMetal:
### Analytics events

- [ ] add `default_value` to `user_attribute_created` + `updated` events. 
- [ ] `default_value` should be NULL if there is no default value
- [ ] `default_value` should contain the default value if one has been enabled. 

### Docs

- [ ] Add a screenshot to this section of the docs showing how to get to the user attributes in the UI: https://docs.lightdash.com/references/user-attributes#managing-user-attributes (you can see an example in Miro [here](https://miro.com/app/board/uXjVPdFXdWo=/?share_link_id=181226454123))
- [ ] remove the numbering `1.` `2.` from the sub-sections in the `managing user attributes` section
- [ ] add a sub-section called `setting a default value for your user attribute` which shows an image of how to do this + describes what it does

PR #7052, opened 2023-09-08
docs: analytics and docs for user attribute default value

rephus:
Closes: #7016

### Description:

Screenshot from 2023-09-08 13-15-14
Screenshot from 2023-09-08 13-15-56

image


TuringLovesDeathMetal (commented):
beautiful! 

Just to double check, the analytics event was also applied to `attribute_updated` correct? (not just `created`)

rephus:
> `attribute_updated`



> beautiful!
> 
> Just to double check, the analytics event was also applied to `attribute_updated` correct? (not just `created`)

Yep, that's correct

TuringLovesDeathMetal (approved):
looks great! thanks for doing this :)

Issue #7056, opened 2023-09-08
True/False default filter value shows `true/false` but is actually set to `false`


TuringLovesDeathMetal:
### Description

By default, when you add a filter to a boolean dimension, it adds `true/false` as the value in the filters tab:

<img width="1051" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/02c42ce5-5aae-4967-845e-712811f0f527">

But, if you head to the SQL, the filter is actually filtering on `is_paying` = `false` by default (not `true` or `false`)

<img width="709" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/dc19d1ea-81f5-4f97-9ae7-788e9d1c265b">

By default, when you add a filter to a boolean dimension, it should show `false` then (if we're adding `false` by default) instead of `true/false`.

PR #7059, opened 2023-09-09
fix: change default display value for boolean filter to False

ujwalkumar1995:
Closes: #7056 

### Description:

The default display value for boolean filters is True/False. The default value applied on the query is False. Change the default display value to be consistent with the backend query.
Attached video for reference.

https://github.com/lightdash/lightdash/assets/20976813/f6d7c0a7-6788-48bc-9a40-1869db37ee58


rephus:
Duplicate PR for testing

rephus (approved):
Tested, code also looks good.

PR #7062, opened 2023-09-11
feat: Show the dbt profile and target when compiling with the CLI

yu-iskw:
### Description:
I have an issue to compile a dbt project with the Lightdash CLI. I am still investigating the root cause. But, I would like to make sure the CLI uses the passed dbt profile and target by showing the logs.


rephus (approved):
Looks good

Issue #7070, opened 2023-09-13
Increase expiration time of scheduled deliveries to 7 days


TuringLovesDeathMetal:
### Description

Right now, scheduled deliveries expire after 3 days, but we want to increase this expiry date to 7 days. 

The reason being: I send reports to external folks using scheduled deliveries (they don't have access to Lightdahs, only to the deliveries I send them). 

if someone from a partner is on holiday / off sick, they can’t access their report.

7 days gives a much better lee-way for this (and if they're away beyond this time, I can resend the report).

ujwalkumar1995:
Hey @TuringLovesDeathMetal 
I have submitted a PR for the above issue. Looking at the code I thought that this expiration time is configurable from env variables, so not sure if this change is even needed. Please feel free to provide any feedback if my understanding was wrong.

rephus:
We've decided internally that we should keep the default as it is, and change the expiry via ENV variable when needed

PR #7074, opened 2023-09-13
feat: increase scheduled delivery expiration time to 7 days

ujwalkumar1995:
Closes: #7070 

### Description:

Increased the expiration time for scheduled deliveries to 7 days as described in the mentioned issue.


rephus (commented):
Code looks good, I just need to confirm if we need to introduce any changes on cloud before deploying this

rephus:
We've decided internally that we should keep the default as it is, and change the expiry via ENV variable when needed . Thanks  for your contribution anyway and sorry for the confusion.

Issue #7071, opened 2023-09-13
Dashboards don't warn users when navigating away with unsaved changes


magnew:
### Description

Dashboards don't warn users when navigating away with unsaved changes

### To Reproduce

1. Make a new dashboard 
2. Add a chart
3. Navigate away
4. There should be a warning, but there isn't 
5. Changes to the dashboard are lost

### App Version
Version: v0.773.4

nimit9:
Hey, I'd like to work on this issue

PR #7075, opened 2023-09-13
fix: warning users before navigating away from dashboard when there a…

nimit9:
Closes: #7071 

Modified the already present code to show a wanring dialog box while navigaing away from dashboard when there are unsaved changes
 
[Screencast from 2023-09-14 02-02-33.webm](https://github.com/lightdash/lightdash/assets/37402791/e14d9afb-49b0-4316-99ce-d2c1953490cd)

### DIFF
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -400,28 +400,37 @@ const Dashboard: FC = () => {
     }, [haveTilesChanged, haveFiltersChanged, isEditMode]);
 
     useEffect(() => {
-        const createChartInDashboardFlow = sessionStorage.getItem(
-            'unsavedDashboardTiles',
-        );
-        history.block((prompt) => {
-            if (
-                isEditMode &&
-                (haveTilesChanged || haveFiltersChanged) &&
-                !prompt.pathname.includes(
-                    `/projects/${projectUuid}/dashboards/${dashboardUuid}`,
-                ) &&
-                createChartInDashboardFlow
-            ) {
-                setBlockedNavigationLocation(prompt.pathname);
-                setIsSaveWarningModalOpen(true);
-                return false; // blocks history
-            }
-            return undefined; // allow history
-        });
+        // Check if in edit mode and changes have been made
+        if (isEditMode && (haveTilesChanged || haveFiltersChanged)) {
+            // Define the navigation block function
+            const navigationBlockFunction = (prompt: { pathname: string }) => {
+                // Check if the user is navigating away from the current dashboard
+                if (
+                    !prompt.pathname.includes(
+                        `/projects/${projectUuid}/dashboards/${dashboardUuid}`,
+                    ) &&
+                    // Allow user to add a new table
+                    !sessionStorage.getItem('unsavedDashboardTiles')
+                ) {
+                    // Set the blocked navigation location to navigate on confirming from user
+                    setBlockedNavigationLocation(prompt.pathname);
+                    // Open a warning modal before blocking navigation
+                    setIsSaveWarningModalOpen(true);
+                    // Return false to block history navigation
+                    return false;
+                }
+                // Allow history navigation
+                return undefined;
+            };
 
-        return () => {
-            history.block(() => {});
-        };
+            // Set up navigation blocking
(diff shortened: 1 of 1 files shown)


rephus:
Duplicate PR for testing

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -387,41 +387,35 @@ const Dashboard: FC = () => {
...
-            'unsavedDashboardTiles',
-        );
-        history.block((prompt) => {
-            if (
-                isEditMode &&
-                (haveTilesChanged || haveFiltersChanged) &&
-                !prompt.pathname.includes(
-                    `/projects/${projectUuid}/dashboards/${dashboardUuid}`,
-                ) &&
-                createChartInDashboardFlow
```
rephus, line 413: @ZeRego do you remember if we are supposed to block navigation if someone is trying to add a chart in dashboard ? I think this is what the condition was for 

[Screencast from 18-09-23 09:25:39.webm](https://github.com/lightdash/lightdash/assets/1983672/f136973c-afdb-4925-8e68-720f3576d6d9)

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -387,41 +387,35 @@ const Dashboard: FC = () => {
         useState<string>();
 
     useEffect(() => {
-        const checkReload = (event: BeforeUnloadEvent) => {
```
rephus, line 390: IIRC this useEffect was supposed to block page refresh (F5)  I don't think this is happening now

magnew: Yea, I think we need `beforeunload` for this to work correctly with reload. Back button seems to be handled in in Chrome at least.

packages/frontend/src/pages/Dashboard.tsx
magnew: This wasn't really handled well before. Fixing it might be what caused the reload warning to break to begin with. With the current design we should *not* show this warning when the user adds a chart to a dashboard.

magnew (changes requested):
@nimit9 thanks for helping with this! A couple behavior changes need to be made:

- We should warn before reloading the page.
- We should *not* warn when a user tries to create a new chart in the dashboard. 

I apologize for not having these details in the spec. This flow has been changing a lot.

nimit9:
> @nimit9 thanks for helping with this! A couple behavior changes need to be made:
> 
> - We should warn before reloading the page.
> - We should *not* warn when a user tries to create a new chart in the dashboard. 
> 
> I apologize for not having these details in the spec. This flow has been changing a lot. 

Got it @magnew. Will make the necessary changes

packages/frontend/src/pages/Dashboard.tsx
nimit9: Made the change to allow user to add a chart to the dashboard without a warning

packages/frontend/src/pages/Dashboard.tsx
nimit9: Made the change which shows a warning when user tries to reload

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -400,28 +400,37 @@ const Dashboard: FC = () => {
...
+        if (isEditMode && (haveTilesChanged || haveFiltersChanged)) {
+            // Define the navigation block function
+            const navigationBlockFunction = (prompt: { pathname: string }) => {
+                // Check if the user is navigating away from the current dashboard
+                if (
+                    !prompt.pathname.includes(
+                        `/projects/${projectUuid}/dashboards/${dashboardUuid}`,
+                    ) &&
+                    // Allow user to add a new table
+                    !sessionStorage.getItem('unsavedDashboardTiles')
```
nimit9, line 413: Checks if user is adding a new chart -> then dont show a warning

magnew: Cool. Just so you know (if you are looking in the future) I'm probably going to change where these go in the near future, but this is fine for now.

magnew (approved):
Tested, works great. Nice job.

Issue #7064, opened 2023-09-12
When applying a string filter, ignore when filter value is empty 


joaoviana:
### Current behaviour

Currently, when a user sets a filter of type string with an empty value, then it compiles as `WHERE (false)` which then yields no results. 

Example: 
<img width="1499" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/be5e4c98-00dc-4b4b-a866-d3de7ac1acea">

compiled query: 
```sql
SELECT
  `dbt_orders`.shipping_city AS `dbt_orders_shipping_city`,
  `dbt_orders`.partner_name AS `dbt_orders_partner_name`,
  COUNT(DISTINCT `dbt_orders`.order_id) AS `dbt_orders_count_distinct_order_id`
FROM `lightdash-analytics`.`lightdash_demo_gardening`.`dbt_orders` AS `dbt_orders`

WHERE ((
  false
))
GROUP BY 1,2
ORDER BY `dbt_orders_count_distinct_order_id` DESC
LIMIT 500
```

see `WHERE ((
  false
))`

### Expected behaviour

Nothing should affect the query if the filter is empty.

ujwalkumar1995:
Hey @joaoviana 
I have raised a PR for the above issue. Could you please review?
I encountered couple of similar bugs. Raised #7083 and #7082 which may be related to this as well.

PR #7081, opened 2023-09-16
fix: query generation with empty value on string and number filters

ujwalkumar1995:
Closes: #7064 

### Description:

When using string filters and not having any value in the criteria, the query generated used to have a where clause (false) as mentioned in the referenced issue. Have changed it to have a where clause as (true) so it does not impact rest of the query. Have made similar change for number type filters as the issue was also persistent in that case as well (not mentioned on the original issue). Please let me know in case this was not needed.

Attached video for reference.

https://github.com/lightdash/lightdash/assets/20976813/497ef34e-4275-416e-8f1d-1d54263a29f5


ZeRego (approved):

ZeRego:
Thanks for the contribution @ujwalkumar1995 🙌

Issue #7095, opened 2023-09-18
Minify CSS with Lightning CSS


joaoviana:
### Description

Use: https://main.vitejs.dev/guide/features.html#lightning-css

And hopefully will see: 5% decrease of the CSS files when bundled

PR #7102, opened 2023-09-18
chore: use lightning css instead of esbuild to reduce asset size

joaoviana:
Closes: #7095 

### Description:

Use `lightningcss` to transform CSS. 

This allows the decrease of CSS assets by 5%. Please request internal notion doc for stats


joaoviana:
Tested on Render and checked the logs, nothing related to CSS not being processed correctly; Assets load fine. Merging

Issue #7107, opened 2023-09-18
Fix get project member access API


ZeRego:
### To Reproduce

https://docs.lightdash.com/api/v1/#tag/Roles-and-Permissions/operation/GetProjectMemberAccess

Typo `$` in `/api/v1/projects/{projectUuid}/user/${userUuid}` + SQL error.

### App Version

0.776.1

PR #7109, opened 2023-09-18
fix(api): get project member access

ZeRego:
Closes: https://github.com/lightdash/lightdash/issues/7107

### Description:

Changes:
- fix path
- fix sql filter
- add API tests


rephus (approved):
Good fix, extra points for tests :+1:

PR #7122, opened 2023-09-19
fix: warning users before navigating away from dashboard when there a…

rephus:
…re unsaved changes (#7075)

Closes:


magnew:
Duplicated ticket is merged

Issue #7106, opened 2023-09-18
Add sentry logging around calls to data warehouse to get results


magnew:
### Description

Add logging so that we can see in Sentry (or Grafana) how long calls to the data warehouses are taking. It should be possible to log when calls are sent and received.

At the moment Sentry shows when the FE is waiting to the backend -- getting a better understanding of how long we are waiting for our own code/database vs the warehouse would be very useful.

PR #7127, opened 2023-09-19
chore: add sentry logging for getting chart results and warehouse query

joaoviana:
Closes: #7106 

### Description:

Adds 2 Sentry Transactions: 

`getResultsForChartWithWarehouseQuery`: The total time for a saved chart result

`runWarehouseQuery`: The total time for a warehouse query

This would benefit observability, specially when encapsulating multiple warehouse requests at a time into one Sentry transaction so it's easier to calculate how much time between Lightdash DB and Warehouse client operations

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1013,17 +1013,26 @@ export class ProjectService {
         user: SessionUser,
         chartUuid: string,
     ): Promise<Record<string, any>[]> {
-        const chart = await this.savedChartModel.get(chartUuid);
-        const { metricQuery } = chart;
-        const exploreId = chart.tableName;
+        const rows = await wrapSentryTransaction(
+            'getResultsForChartWithWarehouseQuery',
+            {
+                user,
+                chartUuid,
+            },
+            async () => {
+                const chart = await this.savedChartModel.get(chartUuid);
+                const { metricQuery } = chart;
+                const exploreId = chart.tableName;
 
-        const rows = await this.runMetricQuery(
-            user,
-            metricQuery,
-            chart.projectUuid,
-            exploreId,
-            undefined,
-            QueryExecutionContext.GSHEETS,
+                return this.runMetricQuery(
+                    user,
+                    metricQuery,
+                    chart.projectUuid,
+                    exploreId,
+                    undefined,
+                    QueryExecutionContext.GSHEETS,
+                );
+            },
         );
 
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1013,17 +1013,26 @@ export class ProjectService {
...
-            user,
-            metricQuery,
-            chart.projectUuid,
-            exploreId,
-            undefined,
-            QueryExecutionContext.GSHEETS,
+        const rows = await wrapSentryTransaction(
+            'getResultsForChartWithWarehouseQuery',
+            {
+                user,
```
rephus, line 1019: should we only use userUuid , and/or name instead of the full user object ? I'm not sure if sentry can even serialize that

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1166,9 +1175,14 @@ export class ProjectService {
                         'warehouse.type',
                         warehouseClient.credentials.type,
                     );
-                    const { rows } = await warehouseClient.runQuery(
-                        query,
-                        queryTags,
+                    const { rows } = await wrapSentryTransaction(
+                        'runWarehouseQuery',
+                        {
+                            query,
```
rephus, line 1181: query is the sqlQuery right ? is this more useful than metricQuery ?

packages/backend/src/services/ProjectService/ProjectService.ts
joaoviana: `CustomSamplingContext` accepts a `data` property that is an object of key `string`s and value `any`, so I would like to keep this to see what it does 👍 
image

packages/backend/src/services/ProjectService/ProjectService.ts
joaoviana: Thanks for spotting this, I want to see: 
- if the sqlQuery could be monitored used as part of tracing so we can check something in Sentry (I know it's already available in Grafana, but trying to maximise the usage with Sentry)
- Added stringified `metricQuery` (don't know?)
- Added `context` to see where this query is being run

packages/backend/src/services/ProjectService/ProjectService.ts
joaoviana: > should we only use userUuid

yes, took your suggestion, thank you!

rephus (approved):

Issue #7058, opened 2023-09-08
Investigate how to improve dashboard page time to interactivity


ZeRego:
### Description

Goal:

- [ ] investigate main problems
- [ ] find solutions
- [ ] create actionable tickets

In this case, it could 30 seconds for a dashboard to be interactive.

![Screenshot 2023-09-08 at 19 00 00](https://github.com/lightdash/lightdash/assets/9117144/c578da08-c8e3-4100-938d-53bbc927f5e6)

Ideas:
- can we reduce the amount of requests? get all charts in 1 request with available filters
- can we make the JS computation non blocking? eg: pivot table computation in web worker. 
- only load chart and results for tiles that are visible 

### What problem does this solve?

Do not have to wait so long to interact with dashboards.

joaoviana:
Investigate lightdash vs warehouse: https://github.com/lightdash/lightdash/issues/7106
This also should improve TTI: https://github.com/lightdash/lightdash/issues/7099

Keeping open because ideas in issue description are very relevant

joaoviana:
Another idea: 
* `/availableFilters` could be optimised: https://github.com/lightdash/lightdash/blob/main/packages/backend/src/services/ProjectService/ProjectService.ts#L1846C65-L1865C65 - we don't need to get the whole chart to get the available filters

joaoviana:
Closing since all tickets for this have been filed

PR #7131, opened 2023-09-19
chore: use promiseAll for some queries in saved chart get

joaoviana:
Related to: #7058 

### Description:

Executes the following queries in parallel when getting a Saved Chart:
* Fields
* Sorts
* Table Calculations
* Additional Metrics


rephus (approved):
Looks good, not sure if this will make any difference in terms of performance, but perhaps we could also wrap this on the same db transaction

IrakliJani (approved):

joaoviana:
> Looks good, not sure if this will make any difference in terms of performance, but perhaps we could also wrap this on the same db transaction (Requires investigation, can be done on a separate PR)

Thanks for this idea @rephus, need to chat with you about this! 💪

Issue #7124, opened 2023-09-19
Back button shouldn't go back to editing a dashboad


magnew:
### Description

The browser back button shouldn't go back to editing. 

### To Reproduce

1. Make a dashboard
2. Add a chart
3. Save
4. Hit the browser back button
5. It goes back to edit state (🛑 wrong)

It should skip the edit state in history.

### App Version

Version: v0.778.1

nimit9:
Hey, I'd like to work on this issue

magnew:
> Hey @magnew , I'd like to work on this issue

Go for it. I suspect it basically needs a history.replace() instead of a push() somewhere, but see what you find.

nimit9:
Hey @magnew, you were right, it needed history.replace() instead of push(). It was also going back to edit mode, if clicking on cancel button while editing, and then hitting the browser back button. Resolved that with history.replace() as well. 

Should I raise a new PR with a new branch? or should I push it to PR [#7075](https://github.com/lightdash/lightdash/pull/7075)

magnew:
> Hey @magnew, you were right, it needed history.replace() instead of push(). It was also going back to edit mode, if clicking on cancel button while editing, and then hitting the browser back button. Resolved that with history.replace() as well.
> 
> Should I raise a new PR with a new branch? or should I push it to PR [#7075](https://github.com/lightdash/lightdash/pull/7075)

Hey @nimit9, a new PR would be great. I merged the other one, but also we generally try to keep PRs as small and targeted as possible so that we can have independent fixes. For example, in this case if something wasn't right with 7075, we could back it out and still keep the fix for this new one. 

Thank you!

nimit9:
Right, got it. Will raise a new PR with the fix

PR #7134, opened 2023-09-19
fix: should not go to editing a dashboard after saving or cancelling

nimit9:
Closes: #7124 
### Description:
Changed history.push() to history.replace() to avoid going back to edit mode

[Screencast from 2023-09-19 23-16-40.webm](https://github.com/lightdash/lightdash/assets/37402791/8ed16591-25ca-453c-aec0-4812834b3e71)


magnew:
@nimit9 Sorry for the delay. Tested locally and this looks good! Nice work, I will merge it. 

I found a couple other issues with the back button while testing this, but I will file them separately.

magnew (approved):
Tested locally, works well.

PR #7141, opened 2023-09-20
docs: update metrics.mdx filters

TuringLovesDeathMetal:
Updates filters docs to include double quotes around the examples (otherwise, some of hte examples actually fail on compile e.g. `user_name: %katie%` needs double quotes)


joaoviana (approved):
great 💪

Issue #4380, opened 2023-02-03
Be able to manually trigger the scheduled delivery


ZeRego:
### Description

Add "send now" button. 

https://miro.com/app/board/uXjVP1D_L4w=/?moveToWidget=3458764564476305389&cot=14

### What problem does this solve?

Be able to quickly get the report and test the scheduler. 

### Acceptance criteria

- [ ] there is an option to `send now` in the buttons at the bottom of hte scheduler creation
- [ ] `send now` sends the delivery to whatever destinations have been selected in step 4
- [ ] These deliveries appear in the scheduled deliveries dashboard

### Documentation

- [ ] add a section above this: https://docs.lightdash.com/guides/how-to-create-scheduled-deliveries#editing-and-deleting-your-scheduled-deliveries
- [ ] Section called: "'Send now' option for a delivery"
- [ ] Add info like: "You can test that your scheduled deliveries are working as expected or send content to users immediately using the "send now" button in your scheduled delivery setup. This will send the content you've selected to the destinations you've added immediately."

PriPatel:
Regarding the use case we spoke about in stand up., I think seeing something like this in the modal itself when Im creating the report would be more useful as I wouldn't necessarily think to look at the 3-dots to do this _after_ I've set it up.
<img width="261" alt="Screenshot 2023-02-16 at 15 30 49" src="https://user-images.githubusercontent.com/12776522/219411447-8e95ac4a-a012-4be3-afd0-9d38ff9bd595.png">

ZeRego:
Questions:
- would that button send it to all destinations or can the user choose?
- if this is for test purposes, should this button be in the scheduler creation/update form?

PR #7146, opened 2023-09-20
feat: send now scheduler

rephus:
Closes: #4380

### Description:

The hard part of this task is to modify scheduler to accept a `schedulerUuid` or a createSchedulerAndTargets, so there is some refactoring. 

Screenshot from 2023-09-21 11-36-52

Slack (see footer) 

https://lightdash.slack.com/archives/C04P2S2UZ32/p1695302162068819

image

image

Email is harder to update (Requires updating the template) so I'm keeping it the same 

Scheduler now logs 

image

image


packages/backend/src/services/SchedulerService/SchedulerService.ts
```diff
@@ -242,4 +247,36 @@ export class SchedulerService {
...
+                'You must specify at least 1 destination before sending a scheduled delivery',
+            );
+        }
+
+        const slackChannels = scheduler.targets
+            .filter(isCreateSchedulerSlackTarget)
+            .map((target) => target.channel);
+        await slackClient.joinChannels(user.organizationUuid, slackChannels);
+
+        schedulerClient.addScheduledDeliveryJob(
```
ZeRego, line 274: We need to validate if the user has access to the chart/dashboard.

rephus: I also tested this on private spaces 

image

rephus: However, this forbiddenError says it is not captured => causes backend to crash.

how do we capture errors on controllers ? is this happening on other places ?

ZeRego (approved):

Issue #7040, opened 2023-09-07
Lightdash Preview is not copying dashboard filters 


TuringLovesDeathMetal:
### Description

When I open a preview environment and open a dashbaord which has filters in prod, the dashboard in my preview environment has no filters. 

<img width="1463" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/92183bcf-0ef9-4ed1-b596-f0c033195339">
<img width="1524" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/12cfeddc-6d80-4d95-bea3-0f201aa5b597">

### To Reproduce

1. Find a dashboard in `prod` that has filters
2. Open a preview environment
3. look at the dashboard from `prod` that has filters, but in your preview environment
4. notice that it doesn't have filters

PR #7174, opened 2023-09-21
fix: copy dashboard filters to preview

ZeRego:
Closes: #7040 

### Description:

Main project
Screenshot 2023-09-21 at 16 25 55
Preview
Screenshot 2023-09-21 at 16 26 30

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -30,7 +30,10 @@ import {
 import { Knex } from 'knex';
 import { DatabaseError } from 'pg';
 import { LightdashConfig } from '../../config/parseConfig';
-import { DbDashboard } from '../../database/entities/dashboards';
+import {
+    DashboardViewsTableName,
+    DbDashboard,
+} from '../../database/entities/dashboards';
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1208,6 +1211,27 @@ export class ProjectModel {
...
+            );
+
+            if (dashboardViews.length > 0) {
+                await trx(DashboardViewsTableName).insert(
+                    dashboardViews.map((d) => ({
+                        ...d,
+                        dashboard_view_uuid: undefined,
+                        dashboard_version_id: dashboardVersionsMapping.find(
+                            (m) => m.id === d.dashboard_version_id,
+                        )?.newId!,
```
joaoviana, line 1230: so by passing this `newId` it gets the filters applied? 🤔

ZeRego: When we duplicate an asset we keep a map of the old and new ids. `dashboardVersionsMapping` 

When we copy the filters row ( table `dashboard_views` ) we need to replace the old version id ( which is from the original dashboard version ) with the new one ( the newly created dashboard version ).

Database structure.

joaoviana (approved):

Issue #7161, opened 2023-09-21
`run query` button should never be disabled


TuringLovesDeathMetal:
### Description

We currently grey it out if the user hasn’t changed metrics / dimensions but commonly the underlying data can change (maybe because you’re fixing the underlying model while experimenting in lightdash) and in that case you really want to run the query again. 

The `run query` button shouldn't get disabled.

PR #7176, opened 2023-09-21
feat: Keep run query enabled in explore even when underlying metrics/dimensions have not been modified

ujwalkumar1995:
Closes:  #7161 
### Description:

As per the issue description, we have to keep the run query button enabled in explore view even when underlying metrics/dimensions have not been modified. Have made the suggested change.

### DIFF
--- a/packages/frontend/src/components/RefreshButton.tsx
+++ b/packages/frontend/src/components/RefreshButton.tsx
@@ -17,11 +17,8 @@ export const RefreshButton = memo(() => {
     const fetchResults = useExplorerContext(
         (context) => context.actions.fetchResults,
     );
-    const hasUnfetchedChanges = useExplorerContext(
-        (context) => context.hasUnfetchedChanges,
-    );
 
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/RefreshButton.tsx
```diff
@@ -17,11 +17,8 @@ export const RefreshButton = memo(() => {
     const fetchResults = useExplorerContext(
         (context) => context.actions.fetchResults,
     );
-    const hasUnfetchedChanges = useExplorerContext(
-        (context) => context.hasUnfetchedChanges,
-    );
 
-    const canRunQuery = !isLoading && isValidQuery && hasUnfetchedChanges;
+    const canRunQuery = !isLoading && isValidQuery;
```
rephus, line 21: I don't think we need this variable here anymore,

we should also remove it from line # 47

rephus (changes requested):

rephus (approved):

Issue #7097, opened 2023-09-18
Ensure we're using text compression/serving gzipped assets


joaoviana:
### Description

Currently, we don't serve gzipped assets on deployments. 
From a Lighthouse report, we can see that `echarts-x` JS asset is 2.5Mb: 
<img width="715" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/cef0f143-8dc3-4e63-b830-7ad773fc679a">

When in fact, when building the prod assets, we can see that the gzipped version is way smaller: 
<img width="978" alt="current-build" src="https://github.com/lightdash/lightdash/assets/7611706/c6591f18-3d85-47bc-b76a-81a6e9776c4f">

For reference, as well, on the network tab:
<img width="880" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/5e76662c-f412-4ee0-9379-0b9044f95955">

The assets are cached, so this only happens on first-time-load, but it could definitely be improved, since every time we release, we generate a new hash suffix, which will then invalidate the cache, and cause a re-fetch of the assets. 

### What problem does this solve?

FCP times. 

### Performance insights for echarts asset: 
<img width="261" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/cfb05b53-bde1-48ee-af3a-18ee17781494">
<img width="1203" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/de55d3b9-ef25-4d4f-b369-6329e04e568b">

joaoviana:
tldr on gzip:
Render envs **compresses** the assets
Vitejs does **not** compress the assets
Current GCP config does **not** compress the assets because of the `Via` header and some other config we're missing.

joaoviana:
Closed related PR as it didn't work. 
Alternative: do it on the express server side.

joaoviana:
Sentry: https://lightdash.sentry.io/issues/4338486790/?project=5959292&query=&referrer=performance-related-issues-issue-stream&statsPeriod=24h

PR #7188, opened 2023-09-22
chore: compress vite assets and serve them if they exist

joaoviana:
Closes: #7097 

### Description:

Gzips `js` and `css` assets and serves them if they exist.

Utilises `vite-plugin-compression2` a package that is currently being maintained: https://github.com/nonzzz/vite-plugin-compression

Uses `express-static-gzip` to serve pre-compressed files

When this gets merged, we should test if it works on our instance first. It works locally (by running the BE + a FE build) and also in Render, but had to clear the cache on Render for it to work.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -66,6 +66,7 @@
         "execa": "^5.0.0",
         "express": "^4.17.3",
         "express-session": "^1.17.2",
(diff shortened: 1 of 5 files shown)


joaoviana:
merging to `alpha` to do further testing

PR #7204, opened 2023-09-25
fix: disable move-to-space option for viewers

magnew:
### Description:

Viewers and Interactive viewers were seeing the 'Move to space' button on charts in dashboards. It errored out when they tried, but this makes it so they won't see the button.


ZeRego (approved):
lgtm 👍 

Interactive viewer:

Screenshot 2023-09-26 at 10 26 59

PR #7132, opened 2023-09-19
chore: reduce query to get available filters

joaoviana:
Relates to: #7058 

### Description:

`availableFilters` was getting the whole chart from the chart model when all info that is necessary is:

 ```
uuid: string;
name: string;
spaceUuid: string;
tableName: string;
projectUuid: string;
```

Sentry `/availableFilters` duration:  https://lightdash.sentry.io/performance/summary/spans/projectService.getAvailableFiltersForSavedQuery:ca6069760127bd19/?project=5959292&statsPeriod=30d&transaction=POST+%2Fapi%2Fv1%2Fdashboards%2FavailableFilters

PR #7226, opened 2023-09-26
fix: remove limit and get all available filters per chart

joaoviana:
Closes: N/A

### Description:

There was a `limit` applied to this query so it was only getting the `availableFilters` for one chart in a dashboard. 
This now fixes it and subqueries the latest version per chart 

Fixes broken change introduced in: https://github.com/lightdash/lightdash/pull/7132/files#r1337073050


magnew (commented):
Tested, seems to fix the issue.

Issue #7227, opened 2023-09-26
default value for user attributes takes first value from list and applies to all


TuringLovesDeathMetal:
### Description

https://lightdash-community.slack.com/archives/C03MG2VTR08/p1695407149865919

When adding a default value, it seems like all users are assigned a value that is not the default value. This applies to users that have a value set and to users that don't have any value set.
Removing the default value fixed the filtering issue.

The value that users are assigned is the first value from the list of assigned values

PR #7232, opened 2023-09-26
fix: user attribute default value

ZeRego:
Closes: #7227 

## Context on what was wrong:

**Wrong sql filters**
The following code intended to fetch values for specific user uuids and for attributes with default values.
```javascript
        // packages/backend/src/models/UserAttributesModel.ts
        if (filters.userUuid) {
            query
                .where(`users.user_uuid`, filters.userUuid)
                .orWhereNotNull(`attribute_default`); 
        }
```
The problem is that it is a `OR` condition so it was returning values for all users where the attribute had a default value.

**Assumptions** 

The logic that builds the query assumed the user attributes argument only had information for 1 user.
```
        const userValue =
            userAttribute.users.length > 0
                ? userAttribute.users[0].value
                : userAttribute.attributeDefault;
```
The problem is that argument had more users values ( see first problem above ).

## Fixes:

- New method to fetch attribute values for a specific user
  - this method abstract the default logic and returns a `key/value` object
- Remove assumptions and default logic from the rest of the app

## Tests:

Table yml
Screenshot 2023-09-26 at 16 51 03

### Case where there is **no** default value:

Screenshot 2023-09-26 at 16 54 21

Admin query

Screenshot 2023-09-26 at 16 54 32

Viewer query

Screenshot 2023-09-26 at 16 54 50

### Case where there is a default value:

Screenshot 2023-09-26 at 16 50 31

Admin query

Screenshot 2023-09-26 at 16 50 41

Viewer query

Screenshot 2023-09-26 at 16 50 48

### DIFF
--- a/packages/backend/src/models/UserAttributesModel.ts
+++ b/packages/backend/src/models/UserAttributesModel.ts
@@ -1,5 +1,5 @@
 import { CreateUserAttribute, UserAttribute } from '@lightdash/common';
-import { Knex } from 'knex';
+import knex, { Knex } from 'knex';
 import { OrganizationTableName } from '../database/entities/organizations';
 import {
     DbOrganizationMemberUserAttribute,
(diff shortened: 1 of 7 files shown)


joaoviana (approved):
Explanation made it clear; 
Great stuff! Just a few non-blocking comments

ZeRego:
@joaoviana addressing those points in follow up PR

Issue #7068, opened 2023-09-13
Scheduled CSV deliveries on dashboard don't respect dashboard filters


rephus:
### Description

Dashboard (not temporary) filters should be applied to CSVs on scheduled deliveries

The results on the CSV we send should match the results on the dashboard. 

Slack thread: https://lightdash.slack.com/archives/C04KD95TF0U/p1694546524211369

PR #7233, opened 2023-09-26
fix: csv export dashboard filters

IrakliJani:
Closes: #7068 

### Description:

OG chart:
CleanShot 2023-09-26 at 20 23 35@2x

dashboard with chart + dashboard filter:
CleanShot 2023-09-26 at 20 23 17@2x

old buggy csv:
CleanShot 2023-09-26 at 20 22 59@2x

new fixed csv:
CleanShot 2023-09-26 at 20 22 51@2x

### DIFF
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -380,48 +380,55 @@ export const getFilterRulesByFieldType = (
         },
     );
 
-// TODO: reuse this function in useDashboardFiltersForExplore
-const getDashboardFiltersForTile = (
-    explore: Explore,
+export const getDashboardFilterRulesForTile = (
     tileUuid: string,
+    tables: string[],
+    rules: DashboardFilterRule[],
+): DashboardFilterRule[] =>
+    rules
+        .filter((rule) => !rule.disabled)
+        .map((filter) => {
+            const tileConfig = filter.tileTargets?.[tileUuid];
+
+            // If the config is false, we remove this filter
+            if (tileConfig === false) {
+                return null;
+            }
+
+            // If the tile isn't in the tileTarget overrides,
+            // we return the filter and don't treat this tile
+            // differently.
+            if (tileConfig === undefined) {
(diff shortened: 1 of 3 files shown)


packages/common/src/utils/filters.ts
```diff
@@ -376,3 +379,103 @@ export const getFilterRulesByFieldType = (
...
+        dashboardFiltersForTile && dashboardFiltersForTile.metrics.length > 0
+            ? {
+                  id: 'no',
+                  and: [
+                      ...(metricQuery.filters.metrics
+                          ? [metricQuery.filters.metrics]
+                          : []),
+                      ...dashboardFiltersForTile.metrics,
+                  ],
+              }
```
IrakliJani, line 469: hey @ZeRego is this the right way to combine filters? in this case dashboard filters to the existing metric filters? if yes, does it matter what the id's will be? I copy pasted this from here: https://github.com/jayanand05/lightdash/blob/360ebbcaf1b74b49ff982193d036f1b4e149bff0/packages/frontend/src/hooks/dashboard/useSavedQueryWithDashboardFilters.ts#L62

I was also thinking to extract this as a function, and name it something like, "combineFilters"

ZeRego: > is this the right way to combine filters?

Yes that is correct. 
We also have a static method `ProjectService.combineFilters` that does almost the same.

> does it matter what the id's will be?

not at the moment

magnew (approved):
This looks good to me. I tested the FE dashboard filter side of this and everything works right.

packages/common/src/utils/filters.ts
IrakliJani: @ZeRego great, as this isn't going to hit the DB and the feature's working - I'm gonna merge it. [Created a ticket](https://github.com/lightdash/lightdash/issues/7246) to tackle duplicated code tomorrow. which will combine `combineFilters` function for both frontend and backend 👌

Issue #7116, opened 2023-09-18
Charts sometimes bring their (dashboard) filters with them 


magnew:
### Description

Sometimes importing a chart to a new dashboard imports filters that were applied to it in another dashboard. 

But not always - needs investigation. It has something to do with the URL.

### To Reproduce

1. Add a saved chart to a dashboard 
2. Add a filter for that chart
3. Create a new dashboard
4. Import that same chart
5. Filters come with it 😞

PR #7243, opened 2023-09-27
fix: clear filters when creating new dashboard

magnew:
Closes: #7116 

### Description:

The problem was:
- When creating a new dashboard from an existing dashboard, the page never remounted, leaving all of the state from the  dashboard context.

This fix:
- Forces a re-mount of the dashboard by keying the page to the dashboard uuid. This clears the filters, but is also probably the right thing to do for other state.

### DIFF
--- a/packages/e2e/cypress/e2e/app/dashboard.cy.ts
+++ b/packages/e2e/cypress/e2e/app/dashboard.cy.ts
@@ -31,7 +31,7 @@ describe('Dashboard', () => {
         });
     });
 
-    it.only('Should use dashboard filters, should clear them for new dashboards', () => {
+    it('Should use dashboard filters, should clear them for new dashboards', () => {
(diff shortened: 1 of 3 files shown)


packages/e2e/cypress/e2e/app/dashboard.cy.ts
```diff
@@ -31,7 +31,7 @@ describe('Dashboard', () => {
         });
     });
 
-    it('Should use dashboard filters', () => {
+    it.only('Should use dashboard filters, should clear them for new dashboards', () => {
```
ZeRego, line 34:
```suggestion
    it('Should use dashboard filters, should clear them for new dashboards', () => {
```

magnew: Oops :( Thank you!

ZeRego (approved):

PR #7244, opened 2023-09-27
chore: improve typing, test coverage, performance of org member attributes values

ZeRego:
Follow up from https://github.com/lightdash/lightdash/pull/7232 review.

Changes:
- improve typing
- improve logic
- add unit test

### DIFF
--- a/packages/backend/src/models/UserAttributesModel.test.ts
+++ b/packages/backend/src/models/UserAttributesModel.test.ts
@@ -1,0 +1,43 @@
+import knex from 'knex';
+import { getTracker, MockClient, Tracker } from 'knex-mock-client';
+import {
+    OrganizationMemberUserAttributesTable,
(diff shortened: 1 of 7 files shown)


packages/backend/src/models/UserAttributesModel.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    let tracker: Tracker;
+    beforeAll(() => {
+        tracker = getTracker();
+    });
+    afterEach(() => {
+        tracker.reset();
+    });
+
+    test('should get dashboard by uuid', async () => {
```
joaoviana, line 28: I think this text needs to change, no?

ZeRego:
```suggestion
    test('should combine attributes default values and org member values', async () => {
```

Issue #6564, opened 2023-07-31
dashboard links should more generically include saved filters (and this should be dynamic)


TuringLovesDeathMetal:
### Description

There are two problems with dashboard links (and saved chart links) right now:

1. The links are really long, so they can get cut off in Slack or in Google slides ([source](https://lightdash-community.slack.com/archives/C03MG2VTR08/p1690221959697409))
2. If I share a link to a dashboard, and the saved filters update, the old link is no longer accurate ([source](https://lightdash.slack.com/archives/C04FQQU4RL7/p1689870805164969))

### Suggested approach

Dashboards/saved charts should have links like: `http://lightdash.com/dashboard_123`

That way, if the default filters changed, you’d still see the dashboard with the correct default filters in any old shared links.

Then, if you add new filters (i.e. temporary filters) on top of the “default” filters (i.e. the ones that are saved to the dashboard) they would append to the end like `http://lightdash.com/dashboard_123?filters=temp_filters_only`

So the saved filters are captured inside the `dashboard_123` bit. But the temporary filters are then appended to that.

### Acceptance criteria

- [x] dashboards/saved chart links should show the correct, most up-to-date saved filters (even if these have changed since the link was shared)
- [x] dashboard links should be short enough that they can be shared in Slack, etc. without breaking.

PR #7261, opened 2023-09-28
feat: 6564 dashboard links should more generically include saved filters and this should be dynamic

magnew:
Closes: #6564 

### Description:

Don't put saved filters in the dashboard URL. Temp filters still get stored there. 

I'm going to follow this up with a bit of a cleanup of the context file.


IrakliJani (approved):
This all looks good to me. I was trying to think of a reason why we were keeping saved filters in there, but I couldn't come up with any ✅

Issue #7246, opened 2023-09-27
Refactor duplicated code with `combineFilters` utility


IrakliJani:
### Description

Refactor duplicated code with `combineFilters` utility
We've got some duplicated code around combining filters in both frontend and backend.

PR #7262, opened 2023-09-28
chore: reuse combine filters

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/7246

### Description:

extracts filter combinators to common/filter/utils and reuses them in fe/be

### DIFF
--- a/packages/backend/src/services/CsvService/CsvService.ts
+++ b/packages/backend/src/services/CsvService/CsvService.ts
@@ -11,6 +11,7 @@ import {
     formatItemValue,
     friendlyName,
     getCustomLabelsFromTableConfig,
+    getDashboardFiltersForTile,
     getItemLabel,
     getItemLabelWithoutTableName,
     getItemMap,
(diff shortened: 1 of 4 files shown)


magnew (changes requested):
I don't obviously see what's causing this in the code, but I think this change keeps the all dashboard charts from loading. I checked render and locally and I see this:

Kapture 2023-09-28 at 19 19 05

Looks like that's why the tests are failing too, so yay tests! 😄

IrakliJani:
> I don't obviously see what's causing this in the code, but I think this change keeps the all dashboard charts from loading. I checked render and locally and I see this: 
> Looks like that's why the tests are failing too, so yay tests! 😄

looks like my change was missing a useMemo so it was re-rendering + refetching infinitely 😂

magnew:
> > I don't obviously see what's causing this in the code, but I think this change keeps the all dashboard charts from loading. I checked render and locally and I see this:
> > Looks like that's why the tests are failing too, so yay tests! 😄
> 
> looks like my change was missing a useMemo so it was re-rendering + refetching infinitely 😂

That's how to keep the data up to date :D

magnew (approved):
Looks good!

Issue #7270, opened 2023-09-29
Error toast message should be applied correct styling


joaoviana:
### Description

Apply correct styling to message in Error toast message

### To Reproduce

Produce an _incorrect_ query on the SQL runner
Run Query

see: 

<img width="577" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/2169d59e-ffd5-480c-ba8f-fabcdb09f480">

PR #7272, opened 2023-09-29
style: fix toaster styling

joaoviana:
Closes: #7270 

### Description:

image


IrakliJani (approved):

IrakliJani:
looks great @joaoviana ! I think there's one more place where we might need to update the styles. component is called `ProjectStatusCallout` feel free to do it in the PR or as a separate one if necessary

Issue #7264, opened 2023-09-28
Temp filters got lost when creating and in-dashboard chart


magnew:
### Description

Temp filters got lost when creating and in-dashboard chart

### To Reproduce

1. Create a dashboard
2. Add a temporary filter
3. Add a new chart (not saved)
4. When you are re-directed to the dashboard the temporary filter is lost

Suggested fix: we are just not saving the tempt filters. We could.

TuringLovesDeathMetal:
Closing this as not planned!

PR #7273, opened 2023-09-29
fix: save temp filters when adding dashboard chart

magnew:
Closes: #7264 

### Description:

Saves dashboard filters when adding a new temp chart. 

This is a draft for now. I think it works, but it needs a bit more test.


magnew:
Closing draft. Will bring it back when I fix an issue.

PR #7279, opened 2023-10-02
chore: clean up dashboard context

magnew:
### Description:

This is just some code cleanup to make the dashboard provider easier to read and make future reviews easier to read. It's mostly moving code around and shouldn't have any functionality change. It:

- Groups useStates together
- Groups useEffects together
- Adds a couple comments describing the useEffects
- Changes a useEffect to a useMemo for allFilterableFields


packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -133,22 +202,44 @@ export const DashboardProvider: React.FC = ({ children }) => {
...
-            setHaveFiltersChanged(false);
+    const hasChartTiles = useMemo(
+        () =>
+            dashboardTiles.filter(
+                (tile) => tile.type === DashboardTileTypes.SAVED_CHART,
+            ).length >= 1,
+        [dashboardTiles],
+    );
+
+    useMemo(() => {
```
ZeRego, line 226: I don't understand the change from useEffect to useMemo. useEffect seems to be the correct since the hook doesn't return a value.

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -97,6 +109,63 @@ export const DashboardProvider: React.FC = ({ children }) => {
...
+            );
+        }
+
+        history.replace({
+            pathname,
+            search: newParams.toString(),
+        });
+    }, [dashboardTemporaryFilters, history, pathname, search]);
+
+    // Gets filters from storage after redirect
```
ZeRego, line 143:
```suggestion
    // Gets filters from URL and storage after redirect
```

packages/frontend/src/providers/DashboardProvider.tsx
magnew: It shouldn't make much difference, I tend to follow the react docs best practices and use useEffect for side effects like fetches and the like. And useMemo for caching computed values, like we're doing here. 

Too many useEffects on a page tends to make it really hard to follow the logic, but I don't have a strong opinion. I'm happy to change this back if you prefer. Or if I am missing something about how this would break 😄

packages/frontend/src/providers/DashboardProvider.tsx
magnew: 👍

packages/frontend/src/providers/DashboardProvider.tsx
ZeRego: > And useMemo for caching computed values, like we're doing here.

But we are not caching/memoizing the computed value.

When I'm scanning the code and see a useMemo I know all the logic for that variable is encapsulated in there. In this case the value is stored in a different state `const [fieldsWithSuggestions, setFieldsWithSuggestions] = useState<FieldsWithSuggestions>({});` that is also being updatedin other places.

> I tend to follow the react docs best practices and use useEffect for side effects like fetches and the like.

Can you find examples where useMemo doesn't return a value?
Because the golden rule I've followed: 
- if it returns a value -> useMemo
- if it returns a function -> useCallback
- if doesn't return anything -> useEffect

magnew: That's fine with me. This can be a useEffect.

ZeRego (approved):

Issue #7285, opened 2023-10-03
Fix tooltip on hover of a line with symbol hidden


joaoviana:
### To Reproduce

Create Line chart with no symbol 
Hover on the line
Can't see isolated view of the tooltip (trigger: `item`)

What the tooltip looks like when the symbol is hidden: 

https://github.com/lightdash/lightdash/assets/7611706/d151a269-35b3-4752-b6ee-afbfbf90e164

What it looks like when there's a symbol:
<img width="128" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/45f2a087-79ce-4d9c-8c73-925aeadd2702">

PR #7290, opened 2023-10-03
fix: keep tooltip trigger axis if line series with no symbol

joaoviana:
Closes: #7285 

### Description:

in: https://github.com/apache/echarts/issues/14563 we can see that there's issues when triggering the tooltip with `trigger: 'item'` + `showSymbol: false`

This PRs checks if the hovered item has that property set to `false` and then will not set the option to change the trigger to `item` if that matches.

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/1574f73e-38cc-4b0b-a84a-87a6d1b293d9


rephus (approved):
Reviewed and tested :+1:

Issue #7288, opened 2023-10-03
Move scheduled deliveries to feature-based folder structure


magnew:
### Description

Move the existing scheduled deliveries code to the new feature based folder structure: https://www.notion.so/lightdash/FE-Folder-structure-255af532a11745d28118e25d3a24882e

- [x] Move components to /features/scheduler/components
- [x] Move hooks and utilities 
- [x] Set up barrel-file imports
- [ ] -Add @features and `@ui` aliases to shorten imports-
- [x] Remove old schedulers location

PR #7291, opened 2023-10-03
chore: move scheduler to features

magnew:
Closes: #7288 

### Description:

Moves the existing scheduled deliveries UI to the new feature-based file structure. 

For reviewers: 
- There are a lot of files here, but there is no logic change. This only changes locations and imports. 
- Specifically
  - Moves all scheduled delivery modal components to `src/features/scheduler/components`
  - Move all scheduler hooks to `src/features/scheduler/hooks`
  - Move some imports out of component files and into `src/features/scheduler/utils`
  - Move GSheets icons to common
  - Re-link all imports
- Note: I tried to use absolute imports but we need to change the build a bit. This keeps the relative imports for now.


IrakliJani (approved):
looks good 👍 let's just wait for "test / build" step to pass

magnew:
> looks good 👍 let's just wait for "test / build" step to pass

Yea, I'll test the deployment as well.

Issue #7289, opened 2023-10-03
Add feature flag for new Scheduled deliveries form


magnew:
### Description

Add a feature flag for the new scheduled deliveries form for development

PR #7294, opened 2023-10-03
chore: feature flag scheduled deliveries ui

magnew:
Closes: #7289 

### Description:

Adds a feature flag to hide the new scheduled deliveries work in progress. I combined the two dialog roots into a single file so there is only one code path. 

To turn the flag on and see the work in progress (not much yet 😄), paste this in the console: 
`localStorage.setItem('scheduler-modal-2', true)`


rephus (approved):
This is also unifying the chart + dashboard modal into a single file, right ? I assume there is no logic change

magnew:
> This is also unifying the chart + dashboard modal into a single file, right ? I assume there is no logic change

Yup, correct. No logic change in the old modal. I combined those two files so the flag only happens in one place.

Issue #5405, opened 2023-05-09
Migrate `Project settings > Project access list` to use Mantine


IrakliJani:
### Description

<img width="709" alt="CleanShot 2023-05-09 at 13 37 06@2x" src="https://user-images.githubusercontent.com/962095/237057119-f490fc33-c466-40aa-a559-d874d827e56b.png">

- [ ] use the same table layout as we used it in tokens page

jayanand05:
working on it.

jayanand05:
hi @IrakliJani , should i make this similar to User Management page because it looks very similar to that page?

ho-dor:
@IrakliJani I see you have removed this from the milestone, is it available to work on ?

hrutik7:
this issue not exist any more i think

IrakliJani:
@ho-dor @hrutik7 this is still relevant, there has been slight changes with the layout but the project access list still needs to be migrated to mantine. let me know if you want to work on this ticket. thanks 🙌 

![CleanShot 2023-09-28 at 12 30 19@2x](https://github.com/lightdash/lightdash/assets/962095/e53fb49a-2681-4edf-811a-10f8f448e189)

hrutik7:
yes ! I love to work on this

IrakliJani:
great, assigned to you!

PR #7298, opened 2023-10-03
fix: migrate UI from blueprint to mantine

Ankush263:
Closes: #5405

### Description:

I have transitioned the UI from using blueprints to Mantine.

 
Screenshot 2023-12-01 at 13 20 47


rephus:
Duplicating PR for review

packages/frontend/src/components/ProjectAccess/ProjectAccess.styles.ts
```diff
@@ -26,19 +25,42 @@ export const UserName = styled.b`
...
+    width: fit-content;
+    border-radius: 2px;
+    background: #8f99a826;
+    color: #1c2127;
+    margin-top: 0.3em;
+    min-height: 20px;
+    padding: 2px 6px;
+    font-size: 12px;
+    line-height: 16px;
+    box-sizing: border-box;
```
IrakliJani, line 65: Hi @Ankush263, thank you for your contribution! As we're transitioning from Blueprint to Mantine, our aim is to eliminate the use of styled component files. I noticed that this file introduces a significant amount of custom CSS, including hardcoded colors and values. While your efforts are appreciated, we're looking to avoid such customizations moving forward. Would you be able to revise this?

packages/frontend/src/components/ProjectAccess/ProjectAccess.tsx
```diff
@@ -57,50 +65,78 @@ const UserListItem: FC<{
...
                     {relevantOrgRole && (
-                        <Tooltip2
-                            content={`This user inherits the organization role: ${relevantOrgRole}`}
+                        <Tooltip
+                            position="left"
+                            color="dark"
+                            withArrow
+                            arrowSize={10}
+                            label={`This user inherits the organization role: ${relevantOrgRole}`}
+                            sx={{ padding: '10px', backgroundColor: '#57606a' }}
```
IrakliJani, line 82: We're aiming to avoid overrides of the default component colors. Is there a specific reason for this override?

Ankush263: Hello @IrakliJani,
by default the tooltip from mantile looks like this


Screenshot (99)


and in Lightdash it looks like this


Screenshot (100)

If you are OK with the by-default looks of mantile, I can change that accordingly

packages/frontend/src/components/ProjectAccess/ProjectAccess.styles.ts
Ankush263: I can fix these but some of the colour components and styles are going to change as Mantile components have different styles than Blueprintjs. So, some of the components look different from the previous. 
Also, if we want to eliminate the style component files, we have to use `sx` property in Mantine as we might need to change some default styling in Mantine in some cases.
If you are okay with that, I can refactor this branch😀

Ankush263: Got it! I will work on restructuring this PR.🚀

Ankush263:
Hi @IrakliJani 👋, I've made some changes to this PR, waiting for your review...

owlas:
Hey @Ankush263, thanks for the contribution! This PR requires changes but the team doesn't currently have time to provide in-depth feedback. I'm going to close this PR while we close out our existing work in progress and then we'll be able to pick this up again

packages/frontend/src/components/ProjectAccess/ProjectAccess.tsx
```diff
@@ -35,140 +14,8 @@ import {
...
-import {
-    ItemContent,
-    RoleSelectButton,
-    SectionWrapper,
-    UserEmail,
-    UserInfo,
-    UserName,
-} from './ProjectAccess.styles';
-
-const UserListItem: FC<{
```
ZeRego, line 49: moved to separate file

Issue #7299, opened 2023-10-03
Improve the `access` parameter of CreateSpace API


yu-iskw:
### Description

According to the API documentation at Lightdash API (0.794.0), the `CreateSpaceInProject` API requires much information about users to share space accesses. Instead, it would be good to pass only an array of user UUIDs.

![image](https://github.com/lightdash/lightdash/assets/1523515/f5c6f4a7-3d57-410c-a302-6de416a4e6c5)

### What problem does this solve?
Lightdash users who manage share access with APIs don't have to collect information about users like their names to create a space with accesses.

https://registry.terraform.io/providers/ubie-oss/lightdash/latest

PR #7300, opened 2023-10-03
fix: Improve the `access` parameter of `CreateSpaceInProject`

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/7299

### Description:
We pass an array of user UUIDs. We drop the parameters about project roles, first names and last names of users.

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -3861,6 +3861,17 @@ const models: TsoaRoute.Models = {
         },
     },
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
+    'Pick_SpaceShare.userUuid_': {
+        dataType: 'refAlias',
+        type: {
+            dataType: 'nestedObjectLiteral',
+            nestedProperties: {
+                userUuid: { dataType: 'string', required: true },
+            },
+            validators: {},
+        },
+    },
+    // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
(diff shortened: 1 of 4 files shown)


rephus:
Duplicating PR for testing yu-iskw:improve-create-space-api

rephus (changes requested):
The change makes sense, however, this API change is not compatible with our UI, which is expecting a full SpaceShare object; it is throwing a `validation error: excesss property` 

So we'll have to update the frontend request to only provide the user uuids in the `body`

image

Let me know if you're up for making this change, otherwise we can do it ourselves

yu-iskw:
@rephus Thank you for catching up with it. I will look into the frontend side too.

yu-iskw:
@rephus I am trying to improve the frontend too. Please tell me, If there is anything else we have to do. Though I have walked through the code of the frontend, I'm totally new to that part. I haven't completely understand what we should do for this purpose.

https://github.com/lightdash/lightdash/pull/7300/commits/a25779ec5e9346aeaae0c886e591aa54e5949adf

rephus (approved):
Great job, I tested this on render and everything is working fine, we can merge .

rephus:
One of the e2e tests failed : `Should get success response (200) from GET savedChartRouter endpoints` I'm trying to understand why

rephus:
I am running that test locally and it seems fine, I guess this is failing because I've made some changes on the render environment to test this feature manually.

yu-iskw:
@rephus Thank you for the support!

yu-iskw:
@rephus How should we handle the issue reported by sentry?

joaoviana:
> @rephus How should we handle the issue reported by sentry?

Hey @yu-iskw, I've fixed it here already https://github.com/lightdash/lightdash/pull/7381 👍

Issue #7076, opened 2023-09-14
Labels indicating the total value should be positioned correctly above the bars in a stacked bar chart


joaoviana:
### Description

Fix the behaviour when displaying total labels on each bar in a stacked bar chart. 
Some might be rendered like this: 
<img width="19" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/0394754f-f820-47be-843f-af7e94d92e73">

When the number should be placed correct just above the upper most stack.

joaoviana:
I'm not able to replicate it @TuringLovesDeathMetal, do you have any ideas?

IrakliJani:
can reproduce this when having a stacked chart + Quarterly dates as X axis

![Image](https://github.com/lightdash/lightdash/assets/962095/7b5ef52b-a66a-4240-8eb8-a7f38416a603)

PR #7304, opened 2023-10-04
fix: fix total label on stacked chart

rephus:
Closes: #7076

### What was the issue: 

It took me a while to understand this issue, but it was really easy to replicate: 

- If you select a chart and select a date as X axis
- But then you sort by "another" dimension
- Then in the code we check if fthe `field` matches the sort : `alreadySorted` 
- If not, then we run a  `results.sort` 
- This method was messing with the results, the series was not matching the results
- This was causing that all results were displayed in the wrong place. 

### Description:

Before:

X: axis Month
Sort by: quarter

Screenshot from 2023-10-04 09-15-25

After:

Screenshot from 2023-10-04 09-22-43


joaoviana:
just restarting render deployment to test, code looks good! 👍

IrakliJani:
restarted tests. looking good!

IrakliJani (approved):

Issue #7287, opened 2023-10-03
First item in list of available Slack channels looks `selected` in scheduled deliveries form


TuringLovesDeathMetal:
### Description

The first channel in the Slack option in scheduled deliveries always looks selected (even though it isn't). 

<img width="523" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/7cdab5d3-9b4d-4001-a1bd-a82dcf50c164">

PR #7305, opened 2023-10-04
fix: set activeItem null if undefined

joaoviana:
Closes: #7287

### Description:

When `activeItem` is `undefined` on the Autocomplete, we should return `null` to indicate no item is active: https://blueprintjs.com/docs/#select/suggest

screen recording:

https://github.com/lightdash/lightdash/assets/7611706/5660de79-1911-49d5-924a-6616fd7e9a2a


magnew (approved):
Nice

Issue #7323, opened 2023-10-04
Migrate SQL section in Explore to use Mantine's


joaoviana:
### Description

Similar to how we render the SQL block in MetricFlow, let's do the same in the explorer

PR #7325, opened 2023-10-04
feat: use prism to render sql code

joaoviana:
Closes: #7323

### Description:

Similar to how SQL code is rendered in MetricFlow, let's also use `Prism` to render the generated SQL in explore: 

Screenshot 2023-10-04 at 14 18 22


ZeRego (approved):

rephus (approved):
Looks nice

Issue #7229, opened 2023-09-26
Metrics are stacked in the SQL runner (even with no group selected)


TuringLovesDeathMetal:
### Description

Right now, if you have a dimension with two metrics in the SQL runner, the metrics are stacked on top of each other (even though there's no group selected). I

<img width="1123" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/6e3b9890-a69e-47d4-acf4-d97900ca5ea6">

<img width="1118" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/cfb2f77e-ed41-4ac0-8a4c-9cc9ec243209">

<img width="1114" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/5f4ab120-af55-426c-a0f0-07f7e84483c9">

https://analytics.lightdash.cloud/share/QG6cIJI50CwTsCMjpg17c

f you build the same sort of query in the Explore view, you see that metrics aren't stacked:

<img width="1126" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/8fc55ae4-66a6-4f23-bc4b-e864fb5e1cec">

https://analytics.lightdash.cloud/share/aow0k_lNIQdU-0XUunxjZ 

Unless you have a `group` selected in the SQL runner, metrics shouldn't be stacked.

ZeRego:
Possible duplicate of https://github.com/lightdash/lightdash/issues/5131

TuringLovesDeathMetal:
^ooh, I think you're totally right!

This only happens if you go:

- bar chart
- then area chart (which we stack by default)
- then back to bar chart

joaoviana:
covered here: https://github.com/lightdash/lightdash/pull/7353

Issue #5131, opened 2023-04-18
bar chart stacking setting persists even if user removes groups


IrakliJani:
### Description

if a user removes groups, the bar chart stacking setting will still be in effect

### To Reproduce

![CleanShot 2023-04-18 at 14 44 42](https://user-images.githubusercontent.com/962095/232754120-cbc5845b-13e2-4bcd-8317-d97b83324743.gif)

TuringLovesDeathMetal:
@IrakliJani I feel like some users are using this as a feature actually? But I can't remember if it's this issue?

IrakliJani:
still relevant. as mentioned we need to support stacking without groups too

ZeRego:
Closing this since we want to [support stacking without grouping](https://github.com/lightdash/lightdash/issues/8701) and fixing this bug would undo the work for that feature.

PR #7353, opened 2023-10-05
fix: turn off stacking if no pivot dimensions

joaoviana:
Closes: #5131 , Closes: #7229

### Description:

Scenario 1:
Turn off stacking if deleted last element in `Groups`

https://github.com/lightdash/lightdash/assets/7611706/cf97386f-c745-4093-9089-f74c478b5f20

Scenario 2: 
Turn off stacking if moving from chart type AREA to something else, with no pivot dimensions.
Example:
* Create bar chart
* Change to Area
* Change back to bar

Screen recording with fix:

https://github.com/lightdash/lightdash/assets/7611706/a0f9d3a2-89cd-4824-b851-e6921a8be296

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
@@ -32,6 +32,7 @@ const VisualizationCardOptions: FC = memo(() => {
         resultsData,
         cartesianConfig,
         setPivotDimensions,
+        pivotDimensions,
         cartesianConfig: { setStacking },
     } = useVisualizationContext();
     const disabled = isLoading || !resultsData || resultsData.rows.length <= 0;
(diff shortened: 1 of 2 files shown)


IrakliJani:
hey @joaoviana does this allow stacking without groups too? I remember there was a customer who was using this "bug" as a "feature"

joaoviana:
> hey @joaoviana does this allow stacking without groups too? I remember there was a customer who was using this "bug" as a "feature"

Thanks for mentioning this @IrakliJani , I've chatted with @TuringLovesDeathMetal , and the bit that we can solve now is: https://github.com/lightdash/lightdash/issues/7229 - https://github.com/lightdash/lightdash/pull/7353/commits/0ea5b1c5329b2a4517e4fa3aba3984e31c3dc5b2

Regarding: https://github.com/lightdash/lightdash/issues/5131, we should allow: 
* Stacking when there's no groups, but we stack the metrics
* Stacking when there are groups

But ☝️ this will require some thinking so will park this one for now

Issue #7330, opened 2023-10-04
Replace MetricFlow branding with dbt Semantic Layer


hamzahc1:
### Description

From a branding perspective, we’re only referring to MetricFlow as a way to power the dbt Semantic Layer. in the app we should refer to dbt Semantic Layer. 

Acceptance Criteria: 

- [ ] URL updated from /metricflow to /dbtsemanticlayer
- [ ] In the UI - replace Metricflow in title to dbt Semantic Layer
- [ ] Replace on-hover description: `The dbt Semantic Layer integration is in beta...`

![Image](https://github.com/lightdash/lightdash/assets/14341285/5f5779c0-1b9e-4cb6-a7eb-c24514bbaff0)

![Image](https://github.com/lightdash/lightdash/assets/14341285/3c19af01-1cd0-421c-b639-465b759f08b3)

PR #7354, opened 2023-10-05
style: dbt semantic layer copy

rephus:
Closes: [#7330](https://github.com/lightdash/lightdash/issues/7330)

### Description:

I replace the frontend URL as well as the Backend API URL, just in case someone wants to use it . 

Screenshot from 2023-10-05 16-07-33


rephus:
@owlas any idea about how to test this locally ?

Issue #7283, opened 2023-10-02
When a tile's title is hidden, the entire header should be removed


gordonkjlee:
### Description
When a tile's title is hidden, the entire header should be removed. Otherwise, the whitespace is unnecessary and may cause the contents to be cut.

### What problem does this solve?
1. The title is hidden, but
2. The whitespace remains and cuts into the tile's content
<img width="655" alt="image" src="https://github.com/lightdash/lightdash/assets/7281432/e54a311e-12ab-4eac-871c-e88a60107a2c">

or even...
<img width="166" alt="image" src="https://github.com/lightdash/lightdash/assets/7281432/320e9a38-c801-4eef-9fb5-51566e0faac2">

PR #7360, opened 2023-10-05
style: center and hide header if hidden title

joaoviana:
Closes: #7283

### Description:

Remove height of a Tile's header block if this is set as hidden -> similar approach to the way we hide the header block if the Markdown Tile's Header is empty: https://github.com/lightdash/lightdash/pull/6535

Example of a tile with hidden title that doesn't have cropped-data issues anymore:
Screenshot 2023-10-05 at 17 55 53


rephus (approved):
Reviewed and tested

PR #7369, opened 2023-10-06
fix: turn off stacking if no pivot dimensions

joaoviana:
Closes: #7229

### Description:

Screen recording: 

Turn off stacking if moving from chart type AREA to something else, with no pivot dimensions.
Example:
* Create bar chart
* Change to Area
* Change back to bar

Screen recording with fix:

https://github.com/lightdash/lightdash/assets/7611706/a0f9d3a2-89cd-4824-b851-e6921a8be296


rephus (approved):
Makes sense

TuringLovesDeathMetal (approved):
Looks great! Thanks for digging into this ❤️

PR #7370, opened 2023-10-06
fix: limit dashboard query for last 10 recently updated items

joaoviana:
Closes: N/A

### Description:

Reduce client-side processing of all dashboards to then get recently updated items by limiting by 10 dashboards. 

Why 10? 

We display the 10 recently updated items -> these can be dashboards **or** charts. 

By getting the last 10 we guarantee that we don't miss any data when aggregating these. 

Future PRs should include: 
-  [ ] limit charts (`useRecentlyUpdatedCharts`)
-  [ ] merge these 2 hooks into one: `useRecentlyUpdatedItems` or similar.


joaoviana:
back to draft because it needs to get pinned items separately

rephus:
I created some dashboards that are not appearing on recently updated (eg:test8) . Any idea why ? 

Screenshot from 2023-10-06 15-34-25

Screenshot from 2023-10-06 15-34-33

joaoviana:
exactly @rephus, thanks for raising this! Need to make more changes to: 
- Most popular
- Most recent 
Verify everything works as expected !

rephus (changes requested):
I don't think the recently updated list is working as expected. I created a new dashboard and it doesn't appear there. 

[Screencast from 12-10-23 11:21:55.webm](https://github.com/lightdash/lightdash/assets/1983672/cdfe57a5-84a6-44f7-964a-1e0da1d1bd13)

joaoviana:
> I don't think the recently updated list is working as expected. I created a new dashboard and it doesn't appear there.
> 
>  Screencast.from.12-10-23.11.21.55.webm

thanks for spotting this @rephus , I wasn't doing it correctly because of the `distinctOn` and the constraint when performing an `orderBy` (first the `dashboard_id`, then the `created_at`) 
the solution for this was to create a sub query so that I could just order by `views` or `created_at` - now it works for both dashboards and charts. Waiting for Render env to be ready 👍

packages/backend/src/models/SpaceModel.ts
```diff
@@ -269,17 +279,37 @@ export class SpaceModel {
...
+            .as('subQuery');
+
+        let dashboardsQuery = this.database.select('*').from(subQuery);
+
+        if (filters?.recentlyUpdated || filters?.mostPopular) {
+            const sortByColumn = filters.mostPopular
+                ? 'views'
+                : 'dashboard_version_created_at';
+
+            dashboardsQuery = dashboardsQuery
```
rephus, line 297: this makes sense

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2188,6 +2193,89 @@ export class ProjectService {
...
+
+        const mostPopular = await this.getMostPopular(allowedSpaces);
+        const recentlyUpdated = await this.getRecentlyUpdated(allowedSpaces);
+
+        return {
+            mostPopular: mostPopular
+                .sort((a, b) => b.views - a.views)
+                .slice(
+                    0,
+                    this.spaceModel.MOST_POPULAR_OR_RECENTLY_UPDATED_LIMIT,
```
rephus, line 2227: is this slice limit necessary ? we're already limiting the results on the query

rephus (approved):
Overall looks good to me

packages/backend/src/services/ProjectService/ProjectService.ts
joaoviana: Oh yes, this combines the most popular dashboards and most popular charts, so they might be 20. But we resort by views, and then slice by the max: `10`

PR #7372, opened 2023-10-06
chore: new chart type samples

rephus:
Closes: 

### Description:

I coded 2 new sample charts for pivot tables and pie charts. 

Screenshot from 2023-10-06 16-03-51
Screenshot from 2023-10-06 16-05-32


rephus:
A test that seems unrelated is failling, both saved chart uuids are pointing to the same url, I think perhaps this is getting the uuid from a different test run 

image

joaoviana (approved):
Looks good! Tested on Render!

PR #7376, opened 2023-10-07
feat: add posthog

owlas:
Features:
- don't identify opt-out users
- events are anonymous before signin/register
- anonymous events before sign-in are attributed to user after sign-in
- new session begins with new anonymous user after sign-out
- sensitive components masked
- users are assigned to groups
- use react library for posthog provider
- use react library for feature flags

Also see issue: https://github.com/lightdash/lightdash-internal-work/issues/961

### DIFF
--- a/packages/frontend/src/pages/SqlRunner.tsx
+++ b/packages/frontend/src/pages/SqlRunner.tsx
@@ -46,7 +46,7 @@ import { SectionName } from '../types/Events';
 
 const generateBasicSqlQuery = (table: string) =>
     `SELECT *
-     FROM ${table} LIMIT 25`;
+   FROM ${table} LIMIT 25`;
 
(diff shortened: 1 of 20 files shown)


packages/frontend/src/pages/SqlRunner.tsx
```diff
@@ -46,7 +46,7 @@ import { SectionName } from '../types/Events';
 
 const generateBasicSqlQuery = (table: string) =>
     `SELECT *
-   FROM ${table} LIMIT 25`;
+     FROM ${table} LIMIT 25`;
```
rephus, line 49: unrelated change

rephus (approved):
Code looks good, I haven't tested it though

owlas:
rebasing

Issue #7371, opened 2023-10-06
When trying to connect a GitLab token, we throw an error because we're checking for github personal access token patterns


TuringLovesDeathMetal:
i have an error when trying to sync lightdash with my dbt project hosted in Gitlab. It seems like a bug in the form, because even though I select 'Gitlab' as my provider, I get the github error message.

![image](https://github.com/lightdash/lightdash/assets/31848148/46e636ac-ac53-4c2b-8dfb-a0cd2b8f558e)

PR #7378, opened 2023-10-09
fix: fix gitlab token form

rephus:
Closes: #7371

### Description:

Tested on create project and update project

[Screencast from 09-10-23 10:37:00.webm](https://github.com/lightdash/lightdash/assets/1983672/2d7bdba2-03f3-4e05-a67c-0203e94e8caf)

### DIFF
--- a/packages/backend/src/projectAdapters/dbtGitlabProjectAdapter.ts
+++ b/packages/backend/src/projectAdapters/dbtGitlabProjectAdapter.ts
@@ -37,7 +37,7 @@ export class DbtGitlabProjectAdapter extends DbtGitProjectAdapter {
         cachedWarehouse,
         dbtVersion,
     }: DbtGitlabProjectAdapterArgs) {
-        const remoteRepositoryUrl = `https://:${gitlabPersonalAccessToken}@${
+        const remoteRepositoryUrl = `https://lightdash:${gitlabPersonalAccessToken}@${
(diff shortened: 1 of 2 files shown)


packages/backend/src/projectAdapters/dbtGitlabProjectAdapter.ts
```diff
@@ -37,7 +37,7 @@ export class DbtGitlabProjectAdapter extends DbtGitProjectAdapter {
         cachedWarehouse,
         dbtVersion,
     }: DbtGitlabProjectAdapterArgs) {
-        const remoteRepositoryUrl = `https://:${gitlabPersonalAccessToken}@${
+        const remoteRepositoryUrl = `https://lightdash:${gitlabPersonalAccessToken}@${
```
rephus, line 40: This was another issue, username can't be empty 

https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#clone-repository-using-personal-access-token

joaoviana: sorry, don't get this part. Does it have to be `lightdash`?

rephus: according to the docs, it can be anything, as long as it is not empty, I tested this with lightdash and `oauth2` , both work. I think we use lightdash on github already

PR #7381, opened 2023-10-09
fix: make access possibly undefined

joaoviana:
Closes: N/A

### Description:

Found Sentry error: https://lightdash.sentry.io/issues/4535709279/events/d4a562e0e27b464c96e8d6aaf8fc8936/?project=5959292&referrer=replay-errors

This improves the type of `access` and adds optional chaining where needed

### DIFF
--- a/packages/frontend/src/components/common/SpaceActionModal/index.tsx
+++ b/packages/frontend/src/components/common/SpaceActionModal/index.tsx
@@ -229,7 +229,7 @@ const SpaceActionModal: FC<Omit<ActionModalProps, 'data' | 'isDisabled'>> = ({
             const result = await createMutation({
                 name: state!.name,
                 isPrivate: state!.isPrivate,
-                access: state!.access.map((access) => ({
+                access: state!.access?.map((access) => ({
                     userUuid: access.userUuid,
                 })),
             });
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/common/SpaceActionModal/index.tsx
```diff
@@ -229,7 +229,7 @@ const SpaceActionModal: FC<Omit<ActionModalProps, 'data' | 'isDisabled'>> = ({
             const result = await createMutation({
                 name: state!.name,
                 isPrivate: state!.isPrivate,
-                access: state!.access.map((access) => ({
+                access: state!.access?.map((access) => ({
```
owlas, line 232: is it ok to mutate with undefined? OR would you prefer to default to an empty list?

owlas (approved):

packages/frontend/src/components/common/SpaceActionModal/index.tsx
joaoviana: Would prefer to default to empty list! Sounds good @owlas

joaoviana: it is ok to mutate with undefined bc of: 
https://github.com/lightdash/lightdash/blob/main/packages/backend/src/services/SpaceService/SpaceService.ts#L121

PR #734, opened 2021-11-04
Add devcontainer

owlas:
* Adds `devcontainer.json` to work with Codespaces
* Makes it possible to run code mounted into docker in another location (requires configuring `LIGHTDASH_CONFIG_FILE` and `DBT_DEMO_DIR`
* Removes outdated entries in `faq.md`
* No longer migrates when running dev docker image (previously `entrypoint-dev.sh`)
* Adds scripts for common actions in dev container `docker/scripts/...`
* Refactor `dockerfile`

PR #7383, opened 2023-10-09
chore: create devcontainer that references docker compose yml 

joaoviana:
Closes: N/A

### Description:

Create `devcontainer` that references Docker compose file so users can try out codespaces with our tailored config

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,5 +2,12 @@
     "name": "Lightdash Dev Environment",
     "dockerComposeFile": ["../docker/docker-compose.dev.yml"],
     "service": "lightdash-dev",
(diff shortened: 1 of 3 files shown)


joaoviana:
@lightdash/engineering I will probably need to merge this to `main` so I can test this. 

it works fine locally with the `Dev Container CLI`: https://github.com/devcontainers/cli

ZeRego:
We should set host requiremener to 8gb or memory since I got a warning that the memory was at 100% when installing dependencies. 
We also need to fix port forwarding.

First PR on codespaces that might help with missing configuration:
https://github.com/lightdash/lightdash/pull/734/files
All PRs related to codespaces
https://github.com/lightdash/lightdash/pulls?q=is%3Apr+is%3Aclosed+codespace+ (edited)

Issue #7329, opened 2023-10-04
Organize Dimensions by entity


hamzahc1:
### Description

Dimensions are easier consumable if you show the entity it comes from for clarity (e.g., customer__customer_name could be nested under Customer)

Future idea: Show the dimension name as is, and only show the entity if there needs disambiguation. One downside of this is that there’s no organization

PR #7384, opened 2023-10-09
style: render semantic layer table label from dimension label

ZeRego:
Closes: #7329

### Description:

The initial idea was to group them by "table" and have a tree. 
This will involve quite a bit of work and needs design. 

I opted to make the labels cleaner for now.

Before:
Screenshot 2023-10-09 at 17 49 06

After:

 
Screenshot 2023-10-09 at 17 48 36


rephus (approved):
Restarting render, but the code looks good, feel free to merge. You'll have to create a new commit to fix the prefix though.

image

Issue #7295, opened 2023-10-03
Deliveries setup tab


magnew:
### Description

Implement the new scheduled deliveries setup tab. Designs here: https://www.figma.com/file/btY2rdBe5I2W4xLR3upmQG/Untitled?type=design&node-id=2981-3667&mode=design

PR #7385, opened 2023-10-09
chore: scheduled deliveries setup in mantine

magnew:
Closes: #7295 

### Description:

> **NOTE**: to test these changes, turn the feature flag on by pasting and running this in the console: `localStorage.setItem('scheduler-modal-2', true)`

This implements the deliveries setup form in Mantine and the new design. This is the bulk of this work, but the review is getting large and it's probably better to check in and add the remaining functionality in smaller reviews. It's feature-flagged so the missing functionality shouldn't affect anyone. 

There is a lot here because I made the UI changes and also moved to the mantine form handler. I would be happy to go over it with someone if that would be helpful. I'm bound to have missed something 😄 

Not yet Implemented functionality coming in (hopefully much smaller) follow ups:

- Send now. At the moment the button is a no-op. 
- Edit mode. This allows creation, but not editing. 
- Form validation
- Advanced tab - mostly the markdown message
- Cron inputs 
- e2e test

Here's what it looks like:
Screenshot 2023-10-09 at 20 31 58


IrakliJani (approved):
I see a lot of hardcoded values and left few comments, wonder if we can get rid of them as it's not easy to maintain them. but the PR looks good!

magnew:
Irakli and I talked about this one in slack. Merging as is and there will be a styles pass incoming.

Issue #7332, opened 2023-10-04
Introduce granularity modifications for any time dimensions


hamzahc1:
### Description

Initially even with event time (metric time) will be fine

PR #7391, opened 2023-10-09
feat: semantic layer time granularity

ZeRego:
Closes: #7332 

### Description:

Screenshot 2023-10-09 at 20 17 53


packages/frontend/src/features/metricFlow/utils/convertMetricFlowFieldsToExplore.ts
```diff
@@ -18,30 +18,53 @@ export default function convertMetricFlowFieldsToExplore(
...
+            hidden: false,
+        };
+        acc[fieldId(dimension)] = dimension;
 
-            return { ...acc, [fieldId(dimension)]: dimension };
-        },
-        {},
-    );
+        if (isTimeDimension && queryableGranularities) {
+            queryableGranularities.forEach((timeGranularity) => {
```
ZeRego, line 43: Generate 1 dimension per time granularity. Is the same approach we have atm with our dimensions + time intervals.

packages/frontend/src/features/metricFlow/utils/convertMetricFlowFieldsToExplore.ts
```diff
@@ -18,30 +18,53 @@ export default function convertMetricFlowFieldsToExplore(
...
+                const timeDimension: CompiledDimension = {
+                    fieldType: FieldType.DIMENSION,
+                    type: DimensionType.TIMESTAMP,
+                    // Note: time columns in results are suffixed with the time granularity eg:'__day'
+                    name: `${name}__${timeGranularity
+                        .toString()
+                        .toLowerCase()}`,
+                    description,
+                    label: `${friendlyName(name)} (${friendlyName(
```
ZeRego, line 52: Wait for the other PR to be merged https://github.com/lightdash/lightdash/pull/7384
So we can use the correct labels.

rephus (commented):
Do you prefer this UI instead of the collapsible dimension we use on explore ? Looks cleaner..

rephus:
Updating render env to test it there

ZeRego:
> Do you prefer this UI instead of the collapsible dimension we use on explore ? Looks cleaner..

Similar to approach on the previous PR, went with the fastest solution. We would have to drop the `navlink` component and build a tree component and manage much more state if we want to have something similar to the explore. Don't wan't to invest that much time at the moment when we know this will get a full revamp later.

rephus (approved):
It works well when 1 dimension is selected, I got an error when I selected 2 date dimensions (with same or different granurality). I don't think this error is related with this change. so feel free to merge. 

image

ZeRego:
> It works well when 1 dimension is selected, I got an error when I selected 2 date dimensions (with same or different granurality). I don't think this error is related with this change. so feel free to merge.
> 
> image

yes, that looks like an internal error from dbt. I reported the error to them 👍

Issue #7388, opened 2023-10-09
Form validation in new scheduled deliveries form


magnew:
### Description

Implement validation in the new scheduled deliveries form. It's set up as a mantine form and we should add validators to the form handler.

- [ ] Validate form fields
- [ ] Make required fields required

PR #7394, opened 2023-10-10
chore: scheduled deliveries form validation

magnew:
Closes: #7388 

### Description:

Add validation to scheduled deliveries form. There might be more we can do here, but these are what is clearly needed. 

- Require name
- Require custom limit when limit is chosen
- Enforce custom limit is an integer
- Enforce emails are emails

### DIFF
--- a/packages/common/src/types/scheduler.ts
+++ b/packages/common/src/types/scheduler.ts
@@ -6,17 +6,6 @@ export type SchedulerCsvOptions = {
     limit: 'table' | 'all' | number;
 };
 
-export enum Limit {
-    TABLE = 'table',
-    ALL = 'all',
-    CUSTOM = 'custom',
-}
-
-export enum Values {
-    FORMATTED = 'formatted',
-    RAW = 'raw',
-}
-
 export type SchedulerImageOptions = {
     withPdf?: boolean;
 };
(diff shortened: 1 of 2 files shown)


IrakliJani (approved):
This looks good, and I think it should get merged without further modifications.

Some general thoughts on frontend form validations:

Next level here can be to add ZOD schemas for validation for the entire scheduler form. by utilizing ZOD:
- We can eliminate the need for custom implementations like `validateEmail` and integer validation.
- seamless integration with mantine UI to display errors.

here's alink where we've used zod validations with mantine: https://github.com/lightdash/lightdash/blob/2710019810636c58d223be0c299bb4c6176eb4e7/packages/frontend/src/components/LimitButton/LimitForm.tsx#L15

To reiterate, this isn't a suggestion for the current task - it's definitely out of scope; it's just a thought for future.

magnew:
> ext level here can be to add ZOD schemas for validation for the entire scheduler form. by utilizing ZOD:

Cool, yea, I'm definitely on board with that. I didn't realize we already had zod -- I was just following what was here. I'll check this in but maybe the future call to action would be to bring zod into more places and remove these custom validators that we have.

Issue #7352, opened 2023-10-05
Exporting a JPG from a chart throws an error: Cannot read properties of null (reading 'length')


rephus:
### Description

Not sure how to replicate exactly, perhaps when the are no results and chart type is pie chart ? 

Sentry error https://lightdash.sentry.io/issues/4471311124/?referrer=issue_alert-slack&notification_uuid=7a26a849-258a-4d73-96a9-de3a5226c62a&alert_rule_id=9333608&alert_type=issue

Replays: https://lightdash.sentry.io/issues/4471311124/replays/?referrer=slack

ydv-ankit:
i want to work on this issue

ydv-ankit:
Okay no problem

PR #7413, opened 2023-10-10
fix: try-catch error when downloading jpeg image from pie chart

rephus:
Closes: #7352

### Description:

This PR will not fix the issue when downloading (that requires a bit more of investigation) , this is just preventing the user from getting a white page when trying to download an image 

image

I believe this error might be happening on echarts library (I couldn't find an issue about this). But only happens when the label is `inside`, not when it is `outside`. This error doesn't happen even if we select outside first, and then outside. I'll create a separate PR 

[Screencast from 10-10-23 16:59:38.webm](https://github.com/lightdash/lightdash/assets/1983672/f87695cf-39c5-416b-8e40-99f7266ad4e0)

### How to test

Try to download an image from the new pie chat on jaffle shop, you should see the error on console, previously that was throwing a white page

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/ChartDownload.tsx
+++ b/packages/frontend/src/components/ChartDownload.tsx
@@ -129,35 +129,44 @@ const ChartDownloadOptions: React.FC<DownloadOptions> = ({
             throw new Error('Chart instance not reachable');
         }
 
-        const svgBase64 = echartsInstance.getDataURL();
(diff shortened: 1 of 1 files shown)


rephus:
I'll create a separate issue to investigate the real issue with pie charts

ZeRego (approved):

Issue #6434, opened 2023-07-20
Pivot table drag and drop config is confusing


magnew:
### Description

When creating a pivot table, the drag and drop experience is a bit confusing. When a field is dragged all of the other fields disappear, losing some context. 

Proposal: when a field is being dragged, other fields are disabled/grayed out but stay in the stay put.

Ankush263:
Hello @magnew 🙋‍♂️, I can work on this feature

magnew:
> Hello @magnew 🙋‍♂️, I can work on this feature

Hi @Ankush263, sure, this would be helpful. Is the proposed problem and fix clear enough to you? When dragging items in this UI, currently the other items disappear, but they should not. 

<img width="349" alt="Screenshot 2023-10-11 at 10 54 38" src="https://github.com/lightdash/lightdash/assets/1864179/20033279-7838-457e-a0df-d82d8edd0f54">

I think we want something more similar to the behavior in the chart series config, where the elements not being dragged stay where they are.

magnew:
> > Hello @magnew 🙋‍♂️, I can work on this feature
> 
> Hi @Ankush263, sure, this would be helpful. Is the proposed problem and fix clear enough to you? When dragging items in this UI, currently the other items disappear, but they should not.
> 
> <img alt="Screenshot 2023-10-11 at 10 54 38" width="349" src="https://user-images.githubusercontent.com/1864179/274206310-20033279-7838-457e-a0df-d82d8edd0f54.png">
> I think we want something more similar to the behavior in the chart series config, where the elements not being dragged stay where they are.

Ok, obviously I missed that there is already a review for this. Conversation can continue there.

ZeRego:
Relevant discussion here with problem/solutions
https://github.com/lightdash/lightdash/pull/7415#discussion_r1354469438

PR #7415, opened 2023-10-10
fix: Pivot table drag and drop

Ankush263:
Closes: #6434

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/DroppableItemsList.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/DroppableItemsList.tsx
@@ -34,7 +34,6 @@ const DroppableItemsList: FC<DroppableItemsListProps> = ({
     droppableId,
     itemIds,
     isDragging,
-    disableReorder,
     placeholder,
 }) => {
     return (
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/DroppableItemsList.tsx
```diff
@@ -75,15 +74,6 @@ const DroppableItemsList: FC<DroppableItemsListProps> = ({
                                         mb="xs"
                                         ref={innerRef}
                                         {...draggableProps}
-                                        style={{
```
ZeRego, line 78: The reason the items were hidden was to hide the problem where the items can't be reordered in the `rows` section.

https://github.com/lightdash/lightdash/assets/9117144/eb18a520-0f94-4fcb-9bbb-6c6ed75b77de

Ideas:
- allow users to reorder the rows ? this might be require a bit of work since we don't support that atm
- only allow the item to be dropped in the correct position
- keep it like this. maybe have a info box explaining why it happens?

cc @magnew

magnew: It would be great to add row-reordering, but I think we should treat that as a future improvement. How about this:

- Keep both dropzones the way they are, but freeze the elements that aren't being dragged. When an item gets dropped it doesn't change the order, just moves it to the other category. Does that sound like it'll work?

ZeRego: That makes sense. 
@Ankush263 I'm going to add all these comments to the main ticket and close this PR but feel free to reopen if you find a solution. Looks like this thread is about the same problem and there are potential solutions there.
https://github.com/atlassian/react-beautiful-dnd/issues/374

Issue #7423, opened 2023-10-11
Add more details and new charts to Usage analytics dashboard


rephus:
### Description

List of improvements: 

- [ ] Remove top 10 limit on `Which users have run the most queries in the last 7 days? (top 10)`
- [ ] Add a new chart to show `dashboards + views` (at least top 20) 
- [ ] Add a new chart to show `charts + views`

Thread: https://lightdash.slack.com/archives/C053ESBT36V/p1696952800523729?thread_ts=1696937066.677159&cid=C053ESBT36V

12ian34:
This is for JustWatch

PR #7432, opened 2023-10-11
feat: add new user activity charts

rephus:
Closes: #7423

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/AnalyticsModel.ts
+++ b/packages/backend/src/models/AnalyticsModel.ts
@@ -11,8 +11,10 @@ import {
     AnalyticsDashboardViewsTableName,
 } from '../database/entities/analytics';
 import {
+    chartViewsSql,
     chartWeeklyAverageQueriesSql,
     chartWeeklyQueryingUsersSql,
+    dashboardViewsSql,
     numberWeeklyQueryingUsersSql,
     tableMostCreatedChartsSql,
     tableMostQueriesSql,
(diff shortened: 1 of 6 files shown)


packages/backend/src/models/AnalyticsModel.ts
```diff
@@ -136,6 +138,10 @@
             chartWeeklyAverageQueriesSql(userUuids, projectUuid),
         );
 
+        const dashboardViews = await this.database.raw(
+            dashboardViewsSql(projectUuid),
```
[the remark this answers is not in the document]
rephus: This is how we are doing requests for analytics, if we need to refactor this I'd refactor all the things, instead of making changes only to these new methods

rephus:
If you are testing this, you should not see the 2 new charts, that's because they are behind a feature flag. You can check the response from the api

packages/frontend/src/pages/UserActivity.tsx
```diff
@@ -299,6 +317,36 @@ const UserActivity: FC = () => {
...
+                            </thead>
+                            {showTableViews(
+                                'dashboard-views',
+                                data.dashboardViews,
+                            )}
+                        </Table>
+                    </ActivityCard>
+                    <ActivityCard grid="table-chart-views">
+                        <Description>Chart views (top 20)</Description>
+                        <Table bordered $showFooter={false}>
```
ZeRego, line 339: If we are going to have more than 10 rows, we need to fix the table scroll.

(ignore data since I hacked FE to print the same results multiple times)
https://github.com/lightdash/lightdash/assets/9117144/ee89936d-5f96-49bc-8269-551ec76e3e42

rephus: good catch

rephus: Replaced react table with mantine 

[Screencast from 12-10-23 13:09:03.webm](https://github.com/lightdash/lightdash/assets/1983672/80861aec-e2f1-431b-b612-22373a29e359)

Issue #6698, opened 2023-08-10
It should show an error and fail when you `stop-preview` an incorrect project


IrakliJani:
### Description

It should show an error and fail when you `stop-preview` an incorrect project

- this should not say done: <img width="434" alt="CleanShot 2023-08-10 at 20 24 11@2x" src="https://github.com/lightdash/lightdash/assets/962095/3340812c-f9fe-4039-a140-e0eaaba53c53">

- this is an example how it should display the error: <img width="410" alt="CleanShot 2023-08-10 at 20 24 23@2x" src="https://github.com/lightdash/lightdash/assets/962095/19552bf0-8ca7-445c-a9e0-5c360e672502">

jayanand05:
working...

jayanand05:
Working on it .

jayanand05:
working..

PR #7439, opened 2023-10-12
fix: It should show an error and fail when you stop-preview an incorrect project

jayanand05:
Closes: #6698

### Description:
It shows an error and fail when you stop-preview an incorrect project

### DIFF
--- a/packages/cli/src/handlers/preview.ts
+++ b/packages/cli/src/handlers/preview.ts
@@ -338,7 +338,7 @@ export const stopPreviewHandler = async (
         });
 
         console.error(
-            `Could not find preview project with name ${projectName}`,
+            styles.error(`You don't have access to this resource or action`),
         );
     }
 };


packages/cli/src/handlers/preview.ts
```diff
@@ -338,7 +338,7 @@ export const stopPreviewHandler = async (
         });
 
         console.error(
-            `Could not find preview project with name ${projectName}`,
+            styles.error(`You don't have access to this resource or action`),
```
ZeRego, line 341: The ticket was referring to how the error message was displayed ( eg: red colour ) and not about the message itself.

ZeRego (changes requested):

packages/cli/src/handlers/preview.ts
jayanand05: Thanks @ZeRego, let me go through it again.

ZeRego:
@jayanand05 feel free to reopen once is ready.

Issue #7317, opened 2023-10-04
Migrate TagInput to Mantine


IrakliJani:
### Description

Migrate TagInput to Mantine

![CleanShot 2023-10-04 at 13 34 33@2x](https://github.com/lightdash/lightdash/assets/962095/40214350-aa0c-432a-b0d2-b0bed311e689)

PR #7452, opened 2023-10-12
feat: filter inputs tag input to mantine

IrakliJani:
Closes: #7317 

### Description:

This is a TagInput component that I am introducing to mantine v6 since we are stuck on v6.

Why are we stuck on mantine v6:
- mantine v7 depends on react 18
- upgrading to react 18 breaks blueprint
- blueprint is the problem
- but we should also migrate v6 mantine to v7 because of the breaking changes

---

TagInput Code comes from three sources but **it must be removed with Mantine v7!!!**
- PR that was going to introduce TagInput to Mantine v6.1 https://github.com/mantinedev/mantine/pull/3674
- TagInput from v7 https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/components/TagsInput/TagsInput.tsx#L93
- me

new tag input:
CleanShot 2023-10-12 at 18 15 46@2x

overflow:
CleanShot 2023-10-12 at 18 15 27@2x

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/TagInput/TagInput.tsx
+++ b/packages/frontend/src/components/common/TagInput/TagInput.tsx
@@ -99,6 +99,9 @@ export interface TagInputProps
 
     /** Called when search changes */
     onSearchChange?: (value: string) => void;
+
+    /** Add on blur */
+    addOnBlur?: boolean;
 }
 
 function splitTags(splitChars: string[] | undefined, value: string) {
@@ -204,6 +207,7 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
             allowDuplicates,
             inputFieldPosition,
             clearButtonProps,
+            addOnBlur = true,
             ...others
         } = useComponentDefaultProps('TagInput', defaultProps, props);
 
@@ -263,16 +267,6 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
             }
         };
 
-        const handleInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
-            if (typeof onBlur === 'function') {
-                onBlur(event);
-            }
-
-            if (clearInputOnBlur) {
-                setSearchValue('');
-            }
-        };
-
         const handleAddTags = (newTags: string[]): boolean => {
             let tags = newTags;
             if (readOnly) {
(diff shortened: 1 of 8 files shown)


IrakliJani:
check if I can enter 0 values...
check if I can on-blur add values

ZeRego:
> * check if I can enter 0 values...
> * check if I can on-blur add values

@IrakliJani  
✅  0 value works
🟥  doesn't support negative values 
🟥  doesn't support decimal values
🟥  doesn't "add" values on blur

IrakliJani:
updated regex
CleanShot 2023-10-13 at 14 28 12@2x

IrakliJani:
> > * check if I can enter 0 values...
> > * check if I can on-blur add values
> 
> @IrakliJani
✅ 0 value works
✅ paste multiple numbers with ',' separator
✅ doesn't support negative values
✅ doesn't support decimal values
🟥 doesn't "add" values on blur

IrakliJani:
✅ 0 value works
✅ paste multiple numbers with ',' separator
✅ doesn't support negative values
✅ doesn't support decimal values
✅ doesn't "add" values on blur

https://github.com/lightdash/lightdash/assets/962095/cf131a8a-c968-4603-965c-9983994cb558

I also tested pasting: 1,2,3,4,0.123,-3.210,0,,,,,8

CleanShot 2023-10-13 at 19 03 07@2x

packages/frontend/src/components/common/TagInput/TagInput.tsx
```diff
@@ -313,6 +307,21 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
             return false;
         };
 
+        const handleInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+            if (addOnBlur && _searchValue !== '') {
+                handleAddTags([_searchValue]);
```
ZeRego, line 312: We need to handle multiple values on blur eg: 1, 10, 100
Check how it's handled in `handleInputKeydown` 
```
if (splitChars!.includes(event.key) && length > 0) {
                const tags = getSplittedTags({
                    splitChars,
                    allowDuplicates,
                    maxTags,
                    value: _searchValue,
                    currentTags: _value,
                });
```

ZeRego (approved):
lgtm 👍

packages/frontend/src/components/common/TagInput/TagInput.tsx
IrakliJani: it is not possible to have `1, and something else in the input after comma` because once you do 1, it will be automatically added

ZeRego: makes sense 👍

IrakliJani: but let me test this more

Issue #7444, opened 2023-10-12
Add error handling to catch ssh tunnel closing


magnew:
### Description

When using an SSH tunnel to connect to a warehouse, the Lightdash server doesn't handle errors thrown when the tunnel closes. We should add a try/catch and keep the server from crashing in the case of these unhandled exceptions. 

Here is a sample log when this crash happens:

```
level”:“error”,“message”:“Uncaught Exception thrown This socket has been ended by the other party”,“stack”:“Error: This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (node:net:569:14)
    at Channel.ondata (node:internal/streams/readable:777:22)
    at Channel.emit (node:events:517:28)
    at Channel.emit (node:domain:552:15)
    at addChunk (node:internal/streams/readable:335:12)
    at readableAddChunk (node:internal/streams/readable:308:9)
    at Readable.push (node:internal/streams/readable:245:10)
    at CHANNEL_DATA (/usr/app/node_modules/ssh2/lib/client.js:587:23)
    at 94 (/usr/app/node_modules/ssh2/lib/protocol/handlers.misc.js:930:16)
    at Protocol.onPayload (/usr/app/node_modules/ssh2/lib/protocol/Protocol.js:2052:10)
    at AESGCMDecipherNative.decrypt (/usr/app/node_modules/ssh2/lib/protocol/crypto.js:987:26)
    at Protocol.parsePacket [as _parse] (/usr/app/node_modules/ssh2/lib/protocol/Protocol.js:2021:25)
    at Protocol.parse (/usr/app/node_modules/ssh2/lib/protocol/Protocol.js:306:16)
    at Socket.<anonymous> (/usr/app/node_modules/ssh2/lib/client.js:775:21)
    at Socket.emit
```

rephus:
Logs on gCP https://console.cloud.google.com/logs/query;query=SEARCH%2528%22This%20socket%20has%20been%20ended%22%2529;cursorTimestamp=2023-10-12T10:49:32.933168249Z;startTime=2023-10-12T04:51:56.698Z;endTime=2023-10-12T11:51:56.698793Z?referrer=search&authuser=1&project=lightdash-cloud-beta

rephus:
Investigation thread: https://lightdash.slack.com/archives/C02GQKJK84Q/p1697121193135549?thread_ts=1697114449.868889&cid=C02GQKJK84Q

rephus:
[...]

PR #7496, opened 2023-10-16
fix: ssh test connection before run query

rephus:
Closes: #7444

### Description:

This will test connection before running a query, makes the server responsive in case one or more requests fail because of the tunnel. 

Screenshot from 2023-10-16 16-35-42

The tunnel object: 

With active connection

Screenshot from 2023-10-16 16-28-26

Without active connection

Screenshot from 2023-10-16 16-28-43

I was hoping this ssh library was using the event emitter interface, so I was expecting to do a `tunnel.on('close')`  to keep track of its state, but I couldn't find anything in the docs 

### Reviewer actions


IrakliJani (approved):
I won't be able to test it but code looks good to me

Issue #5198, opened 2023-04-25
[Custom dimension] Binning with fixed number


ZeRego:
### Description

Allow users to define a dimension that is going to return a fixed amount of bins. 

Design mockups:
https://www.figma.com/file/JLRWyriHFvNnGu6F3klL8o/Untitled?node-id=0-1

- [ ] only numeric dimensions can be 'binned'
- [ ] if you have a numeric dimension, you can open the three-dot menu and see the option to `create custom dimension` (within this, there is the option to select `bin`)
- [ ] There is a modal that pops up to create a binned custom dimension
- [ ] You must name your custom dimension.
  - [ ] By default, this is just `[name of dimension] bins` (e.g. creating a bin on a dimension `age` would be named `age bins` by default)
- [ ] There are no `Bin Types` section (yet) - the only option is fixed number of bins for this ticket.
- [ ] There is a number picker (like in filters) where you select the number of bins that you want. 
- [ ] This number must be an integer.
- [ ] There is copy below the number of bins which tells you what the bin ranges will be for your data 
- [ ] There is copy that warns the user that these bins might change if their underlying data changes. 
- [ ] Once you've saved the custom dimension, it appears on the bottom of the sidebar under `custom dimensions`, above the `metrics` section, but below the `dimensions` section 
- [ ] The custom dimension is selected by default and added to your table. It behaves like a custom metric (for deletion and persistence)

Technical details:
1. add custom dimension to `MetricQuery` type. We should basically follow what we did for the custom metric ( aka additional metric ) 
2. if a binning custom dimension is selected, the query builder should:
  - add the "original" dimension in the main query
  - add the necessary sql in the second query to create the bins
3. if we are sorting by the binning custom dimension, the query builder should:
  - add a custom order by 
```
// example
ORDER BY CASE `age`
         WHEN '< 0' THEN 1
         WHEN '0 to 100' THEN 2
         WHEN '100+' THEN 3
         ELSE 4
         END,
         `age`  
```
4. in the chart visualization, if the binning custom dimension is used, we need to add custom sorting to the series. Otherwise, it will be sorted alphabetically. 

<img width="900" alt="Captura de ecrã 2023-04-25, às 16 21 42" src="https://user-images.githubusercontent.com/9117144/234325278-e4bdc94b-1e1c-431c-a4e5-5481aabd4375.png">

### What problem does this solve?

As a user, I want to be able to create histograms so that I can see how my data is distributed across a specific number of groups.

PR #7515, opened 2023-10-17
feat: [Custom dimension] Binning with fixed number

rephus:
Closes: #5198

### Description

Allow users to define a dimension that is going to return a fixed amount of bins. 

Design mockups:
https://www.figma.com/file/JLRWyriHFvNnGu6F3klL8o/Untitled?node-id=0-1

- [x] only numeric dimensions can be 'binned'
- [x] if you have a numeric dimension, you can open the three-dot menu and see the option to `create custom dimension` (within this, there is the option to select `bin`)
- [x] There is a modal that pops up to create a binned custom dimension
- [x] You must name your custom dimension.
  - [x] By default, this is just `[name of dimension] bins` (e.g. creating a bin on a dimension `age` would be named `age bins` by default)
- [x] There are no `Bin Types` section (yet) - the only option is fixed number of bins for this ticket.
- [x] There is a number picker (like in filters) where you select the number of bins that you want. 
- [x] This number must be an integer.
- [ ] There is copy below the number of bins which tells you what the bin ranges will be for your data 
- [ ] There is copy that warns the user that these bins might change if their underlying data changes. 
- [ ] Once you've saved the custom dimension, it appears on the bottom of the sidebar under `custom dimensions`, above the `metrics` section, but below the `dimensions` section 
- [ ] The custom dimension is selected by default and added to your table. It behaves like a custom metric (for deletion and persistence)

Extra acceptance criteria: 

- [x] Make sure it works for charts
- [x] Sort bins
- [x] Make sure you can select this custom dimension on a chart
- [x] Chart tooltips 
- [x] Pie charts
- [x] Grouped charts
- [x] Pivot tables https://github.com/lightdash/lightdash/issues/7563
- [x] table charts
- [x] You can delete a custom dimension from explore 
- [x] It works with filters  https://github.com/lightdash/lightdash/issues/7564
- [x] It works with sorting
- [x] Build unit tests for bulidquery
- [x] Build e2e tests for custom dimension https://github.com/lightdash/lightdash/issues/7580
- [x] Disable `non-number` inputs in bins
- [ ] Work on `preview results` on the FE
- [ ] Rework `toggleActiveField`
- [x] Test underlying dasta - created a ticket
- [x] Wrap this in feature flag
- [x] Replace getCustomDimensionId with unique label toa void conflicts with other custom dimensions
- [x] Test multiple custom dimensions from the same dimension
- [x] Reference lines - created a ticket
- [x] CTE compatible with table calculations
- [x] persistence (migration + get + save) 
- [x] 1 bucket only edge case
- [x] multiple joins 
- [x] Test on dashboards 

Alternative SQL implementation https://github.com/lightdash/lightdash/pull/7545

Technical details:
1. add custom dimension to `MetricQuery` type. We should basically follow what we did for the custom metric ( aka additional metric ) 
2. if a binning custom dimension is selected, the query builder should:
  - add the "original" dimension in the main query
  - add the necessary sql in the second query to create the bins
  
 https://stackoverflow.com/questions/15464574/use-google-bigquery-to-build-histogram-graph
 

3. if we are sorting by the binning custom dimension, the query builder should:
  - add a custom order by 
```
// example
ORDER BY CASE `age`
         WHEN '< 0' THEN 1
         WHEN '0 to 100' THEN 2
         WHEN '100+' THEN 3
         ELSE 4
         END,
         `age`  
```
4. in the chart visualization, if the binning custom dimension is used, we need to add custom sorting to the series. Otherwise, it will be sorted alphabetically. 

Captura de ecrã 2023-04-25, às 16 21 42

As a user, I want to be able to create histograms so that I can see how my data is distributed across a specific number of groups.
### Reviewer actions

### DIFF
--- a/packages/backend/src/@types/knex-tables.d.ts
+++ b/packages/backend/src/@types/knex-tables.d.ts
@@ -173,7 +173,7 @@ declare module 'knex/types/tables' {
         [SavedChartVersionsTableName]: SavedChartVersionsTable;
         [SavedChartVersionFieldsTableName]: SavedChartVersionFieldsTable;
         [SavedChartVersionSortsTableName]: SavedChartVersionSortsTable;
-        [SavedChartTableCalculationTableName]: SavedQueryTableCalculationTable;
+        [SavedChartTableCalculationTableName]: SavedChartTableCalculationTable;
         [SavedChartAdditionalMetricTableName]: SavedChartAdditionalMetricTable;
         [SpaceTableName]: SpaceTable;
         [DashboardsTableName]: DashboardTable;
--- a/packages/backend/src/@types/knex-tables.d.ts
+++ b/packages/backend/src/@types/knex-tables.d.ts
@@ -83,16 +83,18 @@ import {
 import {
     SavedChartAdditionalMetricTable,
     SavedChartAdditionalMetricTableName,
+    SavedChartCustomDimensionsTable,
+    SavedChartCustomDimensionsTableName,
     SavedChartsTableName,
     SavedChartTable,
+    SavedChartTableCalculationTable,
+    SavedChartTableCalculationTableName,
     SavedChartVersionFieldsTable,
     SavedChartVersionFieldsTableName,
     SavedChartVersionSortsTable,
     SavedChartVersionSortsTableName,
     SavedChartVersionsTable,
     SavedChartVersionsTableName,
-    SavedQueryTableCalculationTable,
-    SavedQueryTableCalculationTableName,
 } from '../database/entities/savedCharts';
 import { SessionTable, SessionTableName } from '../database/entities/sessions';
 import { ShareTable, ShareTableName } from '../database/entities/share';
@@ -171,7 +173,7 @@ declare module 'knex/types/tables' {
         [SavedChartVersionsTableName]: SavedChartVersionsTable;
         [SavedChartVersionFieldsTableName]: SavedChartVersionFieldsTable;
         [SavedChartVersionSortsTableName]: SavedChartVersionSortsTable;
-        [SavedQueryTableCalculationTableName]: SavedQueryTableCalculationTable;
+        [SavedChartTableCalculationTableName]: SavedQueryTableCalculationTable;
         [SavedChartAdditionalMetricTableName]: SavedChartAdditionalMetricTable;
         [SpaceTableName]: SpaceTable;
         [DashboardsTableName]: DashboardTable;
@@ -216,5 +218,6 @@ declare module 'knex/types/tables' {
         [SshKeyPairTableName]: SshKeyPairTable;
         [UserAttributesTable]: DbUserAttribute;
         [OrganizationMemberUserAttributesTable]: DbOrganizationMemberUserAttribute;
+        [SavedChartCustomDimensionsTableName]: SavedChartCustomDimensionsTable;
     }
 }
--- a/packages/backend/src/database/migrations/20231020102654_custom_dimensions.ts
+++ b/packages/backend/src/database/migrations/20231020102654_custom_dimensions.ts
@@ -1,0 +1,38 @@
+import { Knex } from 'knex';
+
+const CUSTOM_DIMENSIONS_TABLE_NAME = 'saved_queries_version_custom_dimensions';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(CUSTOM_DIMENSIONS_TABLE_NAME))) {
+        await knex.schema.createTable(
+            CUSTOM_DIMENSIONS_TABLE_NAME,
+            (tableBuilder) => {
+                tableBuilder.specificType(
+                    'saved_queries_version_custom_dimension_id',
+                    'integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY',
+                );
+                tableBuilder.text('id').notNullable();
+                tableBuilder.text('name').notNullable();
+                tableBuilder.text('dimension_id').notNullable();
+                tableBuilder.text('table').notNullable();
+                tableBuilder.integer('order').notNullable();
+                tableBuilder.text('bin_type').notNullable();
+                tableBuilder.text('bin_number').nullable();
+                tableBuilder
+                    .integer('saved_queries_version_id')
+                    .notNullable()
+                    .references('saved_queries_version_id')
+                    .inTable('saved_queries_versions')
+                    .onDelete('CASCADE');
+
+                tableBuilder.index([
+                    'saved_queries_version_custom_dimension_id',
+                ]);
+            },
+        );
+    }
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.dropTableIfExists(CUSTOM_DIMENSIONS_TABLE_NAME);
+}
--- a/packages/backend/src/database/migrations/20231020102654_custom_dimensions.ts
+++ b/packages/backend/src/database/migrations/20231020102654_custom_dimensions.ts
@@ -17,7 +17,7 @@ export async function up(knex: Knex): Promise<void> {
                 tableBuilder.text('table').notNullable();
                 tableBuilder.integer('order').notNullable();
                 tableBuilder.text('bin_type').notNullable();
-                tableBuilder.text('bin_number').nullable();
+                tableBuilder.integer('bin_number').nullable();
                 tableBuilder
                     .integer('saved_queries_version_id')
                     .notNullable()
--- a/packages/backend/src/queryBuilder.test.ts
+++ b/packages/backend/src/queryBuilder.test.ts
@@ -1,4 +1,4 @@
-import { ForbiddenError } from '@lightdash/common';
+import { BinType, ForbiddenError } from '@lightdash/common';
 import {
     assertValidDimensionRequiredAttribute,
     buildQuery,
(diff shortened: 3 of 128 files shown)


packages/common/src/index.ts
```diff
@@ -694,7 +696,7 @@ export const getAxisName = ({
     axisIndex: number;
     axisName?: string;
     series?: Series[];
-    items: Array<Field | TableCalculation>;
+    items: Array<Field | TableCalculation | CustomDimension>;
```
rephus, line 699: Potential things to refactor: 

We should use unique types (from common) and avoid joining types, this way we don't need to update methods like this: 

eg: AxisItems = Field | TableCalculation | CustomDimension

packages/common/src/types/field.ts
```diff
@@ -103,7 +119,12 @@ export type TableCalculation = {
 };
 
 export const isTableCalculation = (item: Item): item is TableCalculation =>
-    item ? !!item.sql && !('type' in item) && !('tableName' in item) : false;
+    item
+        ? !('binType' in item) &&
```
rephus, line 123: Potential things to refactor: 

This type checking is not very reliable. I had to add `!('binType' in item)` after I added CustomDimension to Item becuase it was getting confused

rephus:
You can have multiple custom dimensions (aka multiple CTEs) 
image

rephus:
Good news about the SQL builder, even though I'm introducing many changes and new CTEs, I kept the SQL withotu custom dimensions the same , so old tests are still passing 

image

packages/backend/src/queryBuilder.ts
```diff
@@ -128,6 +131,105 @@ const getJoinType = (type: DbtModelJoinType = 'left') => {
     }
 };
 
+export const getCustomDimensionSql = ({
```
rephus, line 134: Potential things to refactor:

I extracted all the logic for custom dimensions on a method, this way is easier to generate the sql, we should probably do the same for table calculations, etc.

packages/backend/src/queryBuilder.ts
```diff
@@ -393,7 +509,11 @@ export const buildQuery = ({
             sqlGroupBy,
         ].join('\n');
         const cteName = 'metrics';
-        const cte = `WITH ${cteName} AS (\n${cteSql}\n)`;
+        const ctes = [
```
rephus, line 512: Potential things to refactor:

This is a special edgecase that forces us to wrap the whole query into a CTE so we can use table calculations.

This makes me think we should probably have a better way to join CTEs in the future.

rephus:
Restarted db on render to apply migration

rephus:
Tested with grouping + save + dashboards 

image

packages/backend/src/@types/knex-tables.d.ts
```diff
@@ -171,7 +173,7 @@ declare module 'knex/types/tables' {
         [SavedChartVersionsTableName]: SavedChartVersionsTable;
         [SavedChartVersionFieldsTableName]: SavedChartVersionFieldsTable;
         [SavedChartVersionSortsTableName]: SavedChartVersionSortsTable;
-        [SavedQueryTableCalculationTableName]: SavedQueryTableCalculationTable;
+        [SavedChartTableCalculationTableName]: SavedQueryTableCalculationTable;
```
rephus, line 176: This was a typo on table calculations, spotted here and fixed.

packages/backend/src/@types/knex-tables.d.ts
```diff
@@ -171,7 +173,7 @@ declare module 'knex/types/tables' {
         [SavedChartVersionsTableName]: SavedChartVersionsTable;
         [SavedChartVersionFieldsTableName]: SavedChartVersionFieldsTable;
         [SavedChartVersionSortsTableName]: SavedChartVersionSortsTable;
-        [SavedQueryTableCalculationTableName]: SavedQueryTableCalculationTable;
+        [SavedChartTableCalculationTableName]: SavedChartTableCalculationTable;
```
rephus, line 176: This was a typo on table calculations, spotted here and fixed

joaoviana:
> Looks really good - a few things noted:
> 
> 1. I shouldn't see this "edit calculation button" that then opens the table calc modal:

on `1`, I have a PR that will be ready for review soon: https://github.com/lightdash/lightdash/pull/7590 screen-recording incoming

hamzahc1:
One more thing @rephus - pie charts aren't working for me. Whenever I select Pie chart, the chart section goes blank.

rephus:
> 2\. Using "to" makes sense in the modal when creating a dimension, but on the chart and in results table etc we should use a hyphen (e.g. 0-10, 10-20 not 0 to 10, 10 to 20)

The `to` was part of a screenshot on the spec. I don't mind switching, it should be a really simple change. 

> I couldn't test this in the PR - but do the bin widths automatically update if the min/max values of the dimensions change?

yes becuase the range is calculated in realtime in  the SQL, we don't store the current values on the custom dimension

hamzahc1:
> The `to` was part of a screenshot on the spec. I don't mind switching, it should be a really simple change.
> 
> >

Yep let's change it for hyphens! 

> mmm, it seems ok to me . What can I do to replicate your issue ?

Just tried it again, seems to work! May have been a one-off problem I had in the testing env 🤷

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+    const { setFieldValue } = form;
+
+    useEffect(() => {
+        if (isEditing && isCustomDimension(item)) {
+            setFieldValue('customDimensionLabel', item.name);
+            setFieldValue('binType', item.binType);
+            setFieldValue(
+                'binConfig.fixedNumber.binNumber',
+                item.binNumber ? +item.binNumber : 0,
```
rephus, line 78: should the default be 2 ?

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+                            </Tooltip>
+                        </Group>
+                    </Radio.Group>
+
+                    {form.values.binType === BinType.FIXED_NUMBER && (
+                        <NumberInput
+                            w={100}
+                            label="Bin number"
+                            required
+                            min={2}
```
rephus, line 187: should we set a max ?

packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
```diff
@@ -51,10 +57,19 @@ const TreeSingleNode: FC<Props> = ({ node }) => {
...
     const isFiltered = isField(item) && isFilteredField(item);
 
-    const label = timeIntervalLabel || item.label || item.name;
-
+    const label = isField(item) ? timeIntervalLabel || item.label : item.name;
+    const description = isField(item) ? item.description : undefined;
     const bgColor = getItemBgColor(item);
 
+    const getFieldIconColor = (field: Item | AdditionalMetric) => {
+        if (isCustomDimension(field) || isDimension(field)) return 'blue.9';
```
rephus, line 65: Potential things to refactor, 

Instead of doing all these `if is type` methods, we could have a function (getFieldType) that return sa string, that we can use on a `switch`

packages/frontend/src/components/VisualizationConfigs/PieChartConfig/PieChartLayoutConfig.tsx
```diff
@@ -23,23 +24,25 @@ const PieChartLayoutConfig: React.FC = () => {
         },
     } = useVisualizationContext();
 
+    const allDimensions = [...dimensions, ...customDimensions];
```
rephus, line 27: @joaoviana we already do this on pie charts :eyes:

rephus (commented):
@joaoviana your changes look good to me :+1:

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
rephus: image
 :see_no_evil: it doesn't error

rephus, line 187: I thought you said we should allow `1` , (the backend supports it now)

joaoviana (approved):

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
joaoviana: let's go with `1` since the backend allows that now?

packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
```diff
@@ -85,8 +86,7 @@ const getDataAndColumns = ({
                                 <TableHeaderBoldLabel>
                                     {item === undefined
                                         ? 'Undefined'
-                                        : item.displayName ||
-                                          friendlyName(item.name)}
+                                        : item.name || friendlyName(item.name)}
```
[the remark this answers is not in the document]
rephus: Yes, good catch

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
joaoviana: Let's leave it for now and then we can find through user feedback 😄  https://github.com/lightdash/lightdash/issues/7600

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -364,17 +365,28 @@ const useCartesianChartConfig = ({
...
         return [
-            [...sortedDimensions, ...metrics, ...tableCalculations],
-            sortedDimensions,
+            [
+                ...sortedDimensions,
+                ...metrics,
+                ...tableCalculations,
+                ...customDimensions,
+            ],
+            [...sortedDimensions, ...customDimensions],
```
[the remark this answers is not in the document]
rephus: Yes, this is used to return `customdimensions`on the list of `availableDimensions` which is going to be used to  `Set fallout layout values`

packages/backend/src/database/migrations/20231020102654_custom_dimensions.ts
```diff
@@ -0,0 +1,38 @@
...
+                    'saved_queries_version_custom_dimension_id',
+                    'integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY',
+                );
+                tableBuilder.text('id').notNullable();
+                tableBuilder.text('name').notNullable();
+                tableBuilder.text('dimension_id').notNullable();
+                tableBuilder.text('table').notNullable();
+                tableBuilder.integer('order').notNullable();
+                tableBuilder.text('bin_type').notNullable();
+                tableBuilder.text('bin_number').nullable();
```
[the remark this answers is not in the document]
rephus: yes

packages/frontend/src/components/VisualizationConfigs/PieChartConfig/PieChartLayoutConfig.tsx
joaoviana: So this would help on the toggling... correct? 🤔

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
joaoviana: Amazing that it supports it now! changing to 2!

packages/backend/src/queryBuilder.test.ts
```diff
@@ -390,3 +392,113 @@ describe('assertValidDimensionRequiredAttribute', () => {
         expect(result).toBeUndefined();
     });
 });
+
+describe('with custom dimensions', () => {
+    it('getCustomDimensionSql with empty custom dimension', async () => {
```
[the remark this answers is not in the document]
rephus: you are right, I copied from the tests above

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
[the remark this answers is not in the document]
joaoviana: Thanks for explaining!

PR #7516, opened 2023-10-17
fix: revert adapter destroy on finally

rephus:
Closes: 

### Reviewer actions

test-cli test-frontend test-backend

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -633,9 +633,9 @@ export class ProjectService {
             await adapter.test();
         } catch (e) {
             Logger.error(`Error testing project adapter: ${e}`);
+            await adapter.destroy();
             throw e;
         } finally {
-            await adapter.destroy();
             await sshTunnel.disconnect();
         }
         return { adapter, sshTunnel };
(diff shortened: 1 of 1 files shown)


magnew (approved):

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -633,9 +633,9 @@ export class ProjectService {
             await adapter.test();
         } catch (e) {
             Logger.error(`Error testing project adapter: ${e}`);
+            await adapter.destroy();
             throw e;
         } finally {
-            await adapter.destroy();
             await sshTunnel.disconnect();
```
ZeRego, line 639: don't we need to move this line as well to the catch?
otherwise, the tunnel is going to close before we run the other warehouse queries.

ZeRego (changes requested):

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: mmm, does that mean the tunnel might remain open forever ?

ZeRego: in theory, the "parent" method should disconnect the tunnel the same way it destroys the adaptor.

ZeRego (approved):

PR #7522, opened 2023-10-17
feat: add a validation function to TagInput

magnew:
### Description:

Add a validationFunction to TagInput. It already has a way to pass a regex, but it's a common pattern to pass a function for validation. For example, we use validateEmail and others in a number of places. 

- If neither validationRegex nor validationFunction is set, any value is allowed
- If only one of them is set, rejections are based on that one
- If both are set, new values must pass both tests

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/TagInput/TagInput.tsx
+++ b/packages/frontend/src/components/common/TagInput/TagInput.tsx
@@ -76,8 +76,11 @@ export interface TagInputProps
 
     splitChars?: string[];
 
-    /** Called for validation when add tags */
+    /** Used for validation when adding tags */
     validationRegex?: RegExp;
+
+    /** Called for validation when adding tags */
+    validationFunction?: (data: string) => boolean;
 
     /** Called when validationRegex reject tags */
     onValidationReject?: (data: string[]) => void;
@@ -200,6 +203,7 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
             maxTags,
             splitChars,
             validationRegex,
+            validationFunction,
             searchValue,
             defaultSearchValue,
             onValidationReject,
@@ -279,9 +283,16 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
                 );
             }
 
-            const rejectedTags = tags.filter(
-                (tag) => !validationRegex!.test(tag),
-            );
+            const rejectedTags = tags.filter((tag) => {
+                if (validationFunction) {
+                    return (
+                        !validationRegex!.test(tag) ||
+                        (validationFunction ? !validationFunction(tag) : true)
+                    );
+                }
+
+                return !validationRegex!.test(tag);
+            });
 
             if (maxTags && maxTags >= 0) {
                 const remainingLimit = Math.max(maxTags - _value.length, 0);
@@ -447,6 +458,7 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
                 disabled={disabled}
                 // @ts-ignore
                 autoComplete="off"
+                data-1p-ignore
                 {...rest}
             />
         );


packages/frontend/src/components/common/TagInput/TagInput.tsx
```diff
@@ -447,6 +458,7 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
                 disabled={disabled}
                 // @ts-ignore
                 autoComplete="off"
+                data-1p-ignore
```
magnew, line 461: This is unrelated. Without it one password shows up in this input. It seems like a one-off solution to a problem that might come up with other plugins. But it also seems harmless. 

If we'd rather not do this, I'm happy to remove it.

rephus:
## Blueprint refactor progress

<details>
<summary>Show report</summary>

```
@blueprintjs/core - Total: 340
  Button: 35
  Colors: 32
  NonIdealState: 27
  Spinner: 26
  Classes: 20
  Intent: 18
  Dialog: 17
  FormGroup: 15
  Icon: 15
  Menu: 13
  DialogBody: 10
  DialogFooter: 10
  DialogProps: 10
  MenuDivider: 7
  Divider: 6
  HTMLSelect: 6
  AnchorButton: 6
  Callout: 6
  Card: 5
  InputGroup: 4
  Radio: 4
  PopoverPosition: 4
  Tag: 4
  Position: 3
  H5: 3
  H4: 3
  NumericInput: 3
  RadioGroup: 3
  Collapse: 3
  IconName: 2
  ButtonGroup: 2
  Portal: 2
  Alert: 2
  H3: 2
  Toaster: 1
  AnchorButtonProps: 1
  mergeRefs: 1
  HTMLTable: 1
  ButtonProps: 1
  Drawer: 1
  Text: 1
  Checkbox: 1
  Overlay: 1
  ToastProps: 1
  SpinnerSize: 1
  FocusStyleManager: 1

@blueprintjs/popover2 - Total: 52
  MenuItem2: 25
  Tooltip2: 11
  Popover2Props: 6
  Popover2: 6
  Popover2TargetProps: 2
  MenuItem2Props: 1
  Classes: 1

@blueprintjs/select - Total: 24
  ItemRenderer: 8
  Select2: 5
  MultiSelect2: 4
  ItemPredicate: 2
  Suggest2: 2
  MultiSelect2Props: 1
  ItemListRenderer: 1
  ItemListRendererProps: 1

@blueprintjs/datetime - Total: 8
  DateInput2: 3
  TimePicker: 3
  DateRangeInput2: 1
  DateInput2Props: 1

@blueprintjs/datetime2 - Total: 5
  DateInput2: 3
  DateRangeInput2: 1
  DateInput2Props: 1

-------------------------------------------------------
Total imports from blueprint libraries: 429/993 (43.2%)
Total imports from mantine libraries: 1249
-------------------------------------------------------
```
</details>

packages/frontend/src/components/common/TagInput/TagInput.tsx
```diff
@@ -279,9 +283,16 @@ export const TagInput = forwardRef<HTMLInputElement, TagInputProps>(
...
             }
 
-            const rejectedTags = tags.filter(
-                (tag) => !validationRegex!.test(tag),
-            );
+            const rejectedTags = tags.filter((tag) => {
+                if (validationFunction) {
+                    return (
+                        !validationRegex!.test(tag) ||
+                        (validationFunction ? !validationFunction(tag) : true)
```
IrakliJani, line 290: I was hesitant about using validation regex in the first place because v7 version does not provide any of that https://mantine.dev/core/tags-input/?t=props
I think the reason behind that is react-hook-form and form validation. we can ship this feature but eventually we have to remove both of the props and provide alternative solution to the problem

packages/frontend/src/components/common/TagInput/TagInput.tsx
IrakliJani: oh, did not know about this one!

packages/frontend/src/components/common/TagInput/TagInput.tsx
magnew: Agreed. I looked at the v7 api and decided not to try to make it work here. 

I suggest if we upgrade to v7 we use the new input anyway. In any case I think this function is closer to what's there than passing the regex. We could consider removing the regex.

IrakliJani: > I suggest if we upgrade to v7 we use the new input anyway
💯 absolutely. this is temporary

IrakliJani (approved):

Issue #7434, opened 2023-10-11
If I try to join a model, but don't include the field I'm joining on in the `fields` list, then it doesn't compile


TuringLovesDeathMetal:
### Description

In this example, I'm joining `dbt_orders` to `dbt_support_requests` on `order_id`. I don't want to include `order_id` in the joined model (dbt_orders), because then I have a duplicate field name which is confusing and clutters the sidebar.

<img width="691" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/ac23f76b-553b-4078-98fb-f8f9b9dada1f">

But, if I try to compile this, I get an error because Lightdash doesn't include the joined field.

<img width="916" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/e20cf0aa-3ee6-4570-93aa-994b82566a7d">

We should make it possible to hide the joined field, but still have the model compile.

PR #7525, opened 2023-10-17
feat: joined table fields can now omit fields used in the SQL

ZeRego:
Closes: #7434

### Description:

example yml
```
  - name: orders
    meta:
      joins:
        - join: customers
          sql_on: ${customers.customer_id} = ${orders.customer_id}
          fields: [first_name, last_name] # we don't need to include 'customer_id' anymore
```

Query using join table, customer id is used in the join but not available in the sidebar.

Screenshot 2023-10-17 at 16 40 39

### Reviewer actions

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.test.ts
+++ b/packages/common/src/compiler/exploreCompiler.test.ts
@@ -6,6 +6,7 @@ import {
     compiledJoinedExploreOverridingJoinAlias,
     compiledJoinedExploreOverridingJoinLabel,
     compiledJoinedExploreWithSubsetOfFields,
+    compiledJoinedExploreWithSubsetOfFieldsThatDontIncludeSqlFields,
     compiledSimpleJoinedExplore,
     exploreCircularDimensionReference,
     exploreCircularDimensionShortReference,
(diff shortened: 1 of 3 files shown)


packages/common/src/compiler/exploreCompiler.test.ts
```diff
@@ -139,12 +140,14 @@ describe('Explores with a base table and joined table', () => {
             compiler.compileExplore(joinedExploreWithSubsetOfFields),
         ).toStrictEqual(compiledJoinedExploreWithSubsetOfFields);
     });
-    test('should throw error if referenced field is removed from join', () => {
-        expect(() =>
+    test('should compile with a subset of fields selected on join what dont include the fields in the join SQL', () => {
```
ZeRego, line 143: We already had a test for this scenario but we were forcing the user to include all referenced fields in the fields array. Updated mock+test to reflect the new behaviour.

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -120,29 +129,34 @@ export class ExploreCompiler {
...
+                                requiredDimensionsForJoin.includes(
+                                    dimensionKey,
+                                );
+                            const isVisible =
+                                join.fields === undefined ||
+                                join.fields.includes(dimensionKey) ||
+                                (tableDimensions[dimensionKey].group !==
+                                    undefined &&
+                                    join.fields.includes(
+                                        tableDimensions[dimensionKey].group!,
```
IrakliJani, line 147: `!` should not be necessary. was the type system complaining here?

IrakliJani (approved):
reviewed code, one question about type narrowing - otherwise looks good

Issue #7529, opened 2023-10-18
Values in autocomplete are not sorted correctly


ZeRego:
### To Reproduce

First bug: wrong sorting after multiple searches

https://github.com/lightdash/lightdash/assets/9117144/ec583d45-7808-451d-9f77-d87dba3e4bd7

Second bug: doesn't show results on the second search

https://github.com/lightdash/lightdash/assets/9117144/818146f3-e78c-4513-9a8a-7246a9047f64

### App Version

0.813.1

PR #7533, opened 2023-10-18
fix: filter value aucomplete order and multiple searches

ZeRego:
Closes: #7529 

### Description:

https://github.com/lightdash/lightdash/assets/9117144/63498686-7a02-4dec-b809-9c5269a98fa0

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useFieldValues.ts
+++ b/packages/frontend/src/hooks/useFieldValues.ts
@@ -78,12 +78,15 @@ export const useFieldValues = (
             });
 
             setResults((oldSet) => {
-                return new Set([...oldSet, ...data.results]);
+                return new Set(
+                    [...oldSet, ...data.results].sort((a, b) =>
+                        a.localeCompare(b),
+                    ),
+                );
             });
         },
         [filters, initialData],
     );
-
     const query = useQuery<FieldValueSearchResult, ApiError>(
         ['project', projectId, tableName, fieldName, 'search', debouncedSearch],
         () =>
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useFieldValues.ts
```diff
@@ -96,7 +99,6 @@ export const useFieldValues = (
             ),
         {
             // make sure we don't cache for too long
-            staleTime: 60 * 1000, // 1 minute
```
ZeRego, line 99: Not entirely sure why this was not working. 
If I searched, went back to the tables page, go again the to table and made the same search, useQuery would not trigger the `onSuccess` with the cached data.

joaoviana: apparently react-query doesn't refetch a stale query on every render (ref: https://stackoverflow.com/a/72992274)

joaoviana (approved):

packages/frontend/src/hooks/useFieldValues.ts
ZeRego: Yup, but I expected the `onSuccess` to be triggered with the cached results.
But I suppose what happens is they return the cached data as in `const {data} = useQuery(...)` without triggering the `onSuccess`. So this means we should move all the logic from onSuccess to `useEffect(..., [data]);`

joaoviana: yup, exactly! Guess it aligns with them deprecating `onSuccess` and other callbacks in most recent versions 👍

ZeRego: Decided that this would involve a bit of a refactor and we would need to retest everything. 
Going to merge the current fix without state time.

Issue #7551, opened 2023-10-19
ability to increase pivot table column limit using an ENV variable


IrakliJani:
### Description

At the moment, we cap the number of columns in a pivot table at 60. If a pivoted result exceeds this limit, an error is displayed. The customer is requesting the ability to have more than 60 columns.

### Acceptance criteria:

- [ ] introduce `LIGHTDASH_PIVOT_TABLE_MAX_COLUMN_LIMIT`
- [ ] default it to 60

PR #7554, opened 2023-10-19
feat: pivot table max column env variable

IrakliJani:
Closes: #7551 

### Description:

we've introduced pivot table column limit. this allows individual instances to be configured with custom limit

CleanShot 2023-10-19 at 18 37 33@2x

CleanShot 2023-10-19 at 18 37 29@2x

### Reviewer actions


packages/frontend/src/hooks/pivotTable/pivotQueryResults.ts
```diff
@@ -3,21 +3,23 @@ import {
...
 type PivotQueryResultsArgs = {
     pivotConfig: PivotConfig;
     metricQuery: Pick<
         MetricQuery,
         'dimensions' | 'metrics' | 'tableCalculations' | 'additionalMetrics'
     >;
     rows: ResultRow[];
+    options?: {
+        maxColumns: number;
+    };
```
ZeRego, line 22: imo, having this as optional and defaulting to a constant is going to cause bugs in the future. 
eg: tomorrow someone implements CSV download with pivot data, uses this function but doesn't set the limit. 

I would change the code to require a limit and delete the constant from the codebase so the only place that defines the limit is in the lightdash config.

IrakliJani: yeah, makes sense. I'll update

ZeRego (approved):
lgtm 👍  we just need to delete `packages/common/src/utils/pivot.ts` and fix unit tests before merging.

PR #7568, opened 2023-10-20
chore: improve lightdash team marker

magnew:
### Description:

Lightsmash day!

Makes the internal dev banner look nicer. I'm not sure if we want to keep this thing, but if we do this at least makes it look better.

Before:
Screenshot 2023-10-20 at 15 13 54

After:
Screenshot 2023-10-20 at 15 13 37

If we don't do this, we should probably remove this thing instead 😄 

### Reviewer actions


joaoviana (approved):

IrakliJani (approved):
⚡

Issue #7531, opened 2023-10-18
page layout issues on different pages


IrakliJani:
### Description

![CleanShot 2023-10-17 at 18 26 20@2x](https://github.com/lightdash/lightdash/assets/962095/bd18c64a-bee0-4ccd-be63-785ddc406fd6)

there seems to be a conflicting configuration option to `Page` component especially on pages with banners

joaoviana:
Is it specifically on this page or on others? 
I've merged: https://github.com/lightdash/lightdash/pull/7455
And I tested: chart within dashboard, but no `New chart within dashboard` 😢

IrakliJani:
@joaoviana when creating chart within dashboard but the logic is off and Page component needs more thinking and changes. I have a work in progress branch. that PR also caused some other issues see thread here: https://lightdash.slack.com/archives/C02GQKJK84Q/p1697925877802889

PriPatel:
Will this close issue https://github.com/lightdash/lightdash/issues/8364 ?

joaoviana:
> Will this close issue https://github.com/lightdash/lightdash/issues/8364 ?

Yup @PriPatel

PR #7569, opened 2023-10-20
chore: sidebar for chart config

magnew:
### Description:

Lightsmash! Experiment. This adds a sidebar as for the chart config. Solving at least 3 annoying problems:
- Config panels covering the charts
- Having to go back and forth between two dropdowns to choose chart type, then config
- Chart config layout issues where the config panels get too long

This is behind a feature flag. I want to get @PriPatel's input before we turn it on for everyone, but I think it's ready to check in and have people try it. There are a couple minor issues that could make it better:
- The first time it opens it is slow. I don't know why. I think it's acceptable as is, but it could be snappier. 
- charts in dashboards header is currently broken and this doesn't look great there. That is a separate issue that needs to be addressed: #7531, then this feature should be updated. 
- Opening it causes a chart re-render. Probably something with the context. I'll fix this, but it doesn't seem like a blocker.

Here's a demo:
 
Kapture 2023-10-20 at 17 43 44

### Reviewer actions


hamzahc1:
This looks and feels so much better ❤️ 

Noticed a couple of things though: 
- The sidebar covers the title of the chart too, was this intentional? It'd be cleaner if it just covered the same real estate as the sidebar 
- It doesn't totally cover the sidebar if the sidebar is expanded (which means you can also click on the dimensions/metrics behind the configuration panel
CleanShot 2023-10-21 at 21 11 47@2x

magnew:
> This looks and feels so much better ❤️. We should release this as soon as the code is approved!
> 
> Noticed a couple of things though:
> 
> * The sidebar covers the title of the chart too, was this intentional? It'd be cleaner if it just covered the same real estate as the sidebar
> * It doesn't totally cover the sidebar if the sidebar is expanded (which means you can also click on the dimensions/metrics behind the configuration panel
>   CleanShot 2023-10-21 at 21 11 47@2x

Good points! Yea, this is the one-day version of this. We can definitely improve it, but I already like it better. 

- Good point about the title. I think eventually we might want to put the editable title in the sidebar, but for now I will make it at least not over the title. 
- The other sidebar is a bit trickier. I thought about this and didn't have a good, quick answer. Changing or interacting with the rest of the page layout (the field picker) will be more complex. And I'm not entirely sure what the behaviour should be. I will see if there is something straightforward to do. You think this is a blocker?

hamzahc1:
> The other sidebar is a bit trickier. I thought about this and didn't have a good, quick answer. Changing or interacting with the rest of the page layout (the field picker) will be more complex. And I'm not entirely sure what the behaviour should be. I will see if there is something straightforward to do. You think this is a blocker?

No definitely not a blocker! Can be opened as a separate issue for the future.

ZeRego (approved):

Issue #7571, opened 2023-10-21
Registering via Google sends you back to register page 


hamzahc1:
### Description

If you register for a new Lightdash Cloud account via the Google login, it redirects you back to the register page

PR #7572, opened 2023-10-21
fix: redirect on register

owlas:
Closes: #7571 

Things to tidy up:
- Refactor redirect code in Login, Register, Invite components - so that all methods use the same redirectURL
- Session data is cleared on `.authenticate` in passport: https://medium.com/passportjs/fixing-session-fixation-b2b68619c51d - are we accounting for this?
- Apply redirect logic consistently across all OIDC providers

Cases to test (common across password and all SSO providers):
- After logging in, I'm redirected to the homepage or `from` if it exists
- After registering via `/register` I'm redirected to the homepage or `from` if it exists
- After registering via `/login` with SSO I'm redirected to the homepage or `from` if it exists
- After registering via an invite link I'm redirected to the hompage or `from` if it exists

### DIFF
--- a/packages/backend/src/controllers/authentication.ts
+++ b/packages/backend/src/controllers/authentication.ts
@@ -135,7 +135,7 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
                 redirectUrl.host === originUrl.host ||
                 process.env.NODE_ENV === 'development'
             ) {
-                req.session.oauth.returnTo = redirect;
+                req.session.oauth.returnTo = redirectUrl.href;
             }
         } catch (e) {
             next(); // fail silently if we can't parse url
@@ -143,7 +143,17 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
     }
     next();
 };
-export const redirectOIDCSuccess: RequestHandler = (req, res) => {
+
+export const getSuccessURLWithReturnTo = (req: Request): string => {
+    const url = new URL('/api/v1/oauth/success', lightdashConfig.siteUrl);
+    if (req.session.oauth?.returnTo) {
+        url.searchParams.set('returnTo', req.session.oauth.returnTo);
+    }
+    return url.href;
+};
+
+export const redirectOIDC: RequestHandler = (req, res) => {
+    // Workaround for https://github.com/jaredhanson/passport/pull/941
     const queryReturn =
         typeof req.query.returnTo === 'string'
             ? decodeURIComponent(req.query.returnTo.toString())
@@ -158,10 +168,6 @@ export const redirectOIDCSuccess: RequestHandler = (req, res) => {
     }
 };
 
-export const redirectOIDCFailure: RequestHandler = (req, res) => {
-    const { returnTo } = req.session.oauth || {};
-    res.redirect(returnTo || '/');
-};
 export const googlePassportStrategy: GoogleStrategy | undefined = !(
     lightdashConfig.auth.google.oauth2ClientId &&
     lightdashConfig.auth.google.oauth2ClientSecret
--- a/packages/backend/src/controllers/authentication.ts
+++ b/packages/backend/src/controllers/authentication.ts
@@ -129,7 +129,7 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
     }
     if (typeof redirect === 'string') {
         try {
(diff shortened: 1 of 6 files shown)


packages/frontend/src/pages/Register.tsx
```diff
@@ -69,6 +70,7 @@ const Register: FC = () => {
                     key={providerName}
                     providerName={providerName}
                     intent="signup"
+                    redirect={redirectUrl}
```
owlas, line 73: Apply redirect also to SSO login

packages/frontend/src/pages/Register.tsx
```diff
@@ -33,6 +33,9 @@ const Register: FC = () => {
     const allowPasswordAuthentication =
         !health.data?.auth.disablePasswordAuthentication;
     const { identify } = useTracking();
+    const redirectUrl = location.state?.from
```
owlas, line 36: pull out redirect logic across SSO and password

packages/frontend/src/components/common/ThirdPartySignInButton/index.tsx
```diff
@@ -71,21 +72,17 @@ export const ThirdPartySignInButton: FC<ThirdPartySignInButtonProps> = ({
...
 }) => {
     const { health } = useApp();
-    const location = useLocation<{ from?: Location } | undefined>();
 
     switch (providerName) {
         case OpenIdIdentityIssuerType.GOOGLE:
             return health.data?.auth.google.enabled ? (
                 <ThirdPartySignInButtonBase
                     loginPath={health.data.auth.google.loginPath}
-                    redirect={
```
owlas, line 84: remove google-specific default

packages/frontend/src/components/common/ThirdPartySignInButton/index.tsx
```diff
@@ -71,21 +72,17 @@ export const ThirdPartySignInButton: FC<ThirdPartySignInButtonProps> = ({
     inviteCode,
     intent = 'signin',
     providerName,
+    redirect,
```
owlas, line 75: redirect applies to all providers

packages/backend/src/controllers/authentication.ts
```diff
@@ -149,15 +149,15 @@ export const redirectOIDCSuccess: RequestHandler = (req, res) => {
             ? decodeURIComponent(req.query.returnTo.toString())
             : undefined;
 
-    if (queryReturn && queryReturn.startsWith(lightdashConfig.siteUrl)) {
+    if (
+        queryReturn &&
+        new URL(queryReturn, lightdashConfig.siteUrl).host ===
```
owlas, line 154: allows `queryReturn` to be path only e.g. `/` or `/path/to/chart` without the host

packages/backend/src/controllers/authentication.ts
```diff
@@ -149,15 +149,15 @@ export const redirectOIDCSuccess: RequestHandler = (req, res) => {
...
+            new URL(lightdashConfig.siteUrl).host
+    ) {
         res.redirect(queryReturn);
-        return;
     }
-
-    const { returnTo } = req.session.oauth || {};
```
owlas, line 157: don't use session redirect anymore (this is redundant now)

packages/backend/src/controllers/authentication.ts
```diff
@@ -129,7 +129,7 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
     }
     if (typeof redirect === 'string') {
         try {
-            const redirectUrl = new URL(redirect);
+            const redirectUrl = new URL(redirect, lightdashConfig.siteUrl);
```
owlas, line 132: allows queryReturn to be path only e.g. / or /path/to/chart without the host

rephus (commented):
Sorry, this was probably introduced by https://github.com/lightdash/lightdash/pull/7495 ( Not sure what is wrong, I tested this locally :/ ) Feel free to revert my code if you prefer

rephus (commented):
You have done a great job applying the change to multiple SSO , and extracting the logic into a separate method. I'm not sure I understand what's the real fix here , is the host prefix ? Were you able to reproduce the redirect issue ?

rephus:
I tested locally and it works fine

rephus (approved):

Issue #7596, opened 2023-10-23
Optimize/remove code that converts warehouse values to date objects


ZeRego:
### Description

Bigquery and Snowflake don't return date objects. 

We have a function that loops over all rows and cells to format those values.

We should ideally remove this code:
- problems I found if the values aren't date objects
  - table calculation and Metric Min/Max won't be formatted correctly when the value is a date/timestamp. 

There might be some configuration in their clients to return date objects instead of number/string.

Otherwise, we can at least optimize the code:
- if there are no date/timestamp columns, skip function
- loop over the date/timestamp columns instead of all of them

PR #7624, opened 2023-10-24
fix: only parse necessary columns

ZeRego:
Closes: #7596

### Description:

Changes:
- only parse rows if there are columns that need to be parsed
- only parse necessary columns
- stop formatting arrays/objects for Bigquery. It was the only warehouse doing it so I amended the tests.

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.test.ts
+++ b/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.test.ts
@@ -1,6 +1,5 @@
 import { DimensionType } from '@lightdash/common';
 import { createConnection } from 'snowflake-sdk';
-import { Readable } from 'stream';
 import {
     mapFieldType,
     SnowflakeWarehouseClient,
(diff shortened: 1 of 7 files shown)


packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.test.ts
```diff
@@ -11,19 +10,11 @@ import {
     expectedFields,
     expectedRow,
     expectedWarehouseSchema,
+    mockStreamRows,
     queryColumnsMock,
 } from './SnowflakeWarehouseClient.mock';
 import { config } from './WarehouseClient.mock';
 
-const mockStreamRows = () =>
```
ZeRego, line 18: moved to mock file

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -276,7 +288,7 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
                             }),
                             {},
                         );
-                        resolve({ fields, rows: parseRows(data) });
+                        resolve({ fields, rows: data });
```
ZeRego, line 291: we don't need to parse these results as it is only used to fetch the catalog and such.

magnew (approved):

PR #7626, opened 2023-10-24
feat: account for preview banner in chart sidebar

magnew:
### Description:

Account for preview banner in the chart config sidebar. Without this the sidebar doesn't layout right in preview.

This hardcoded solution is temporary until we have more wholistic page layout. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationSidebar.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationSidebar.tsx
@@ -16,76 +16,80 @@ import VisualizationCardOptions from '../VisualizationCardOptions';
 const VisualizationSidebar: FC<{
     chartType: ChartType;
     savedChart?: SavedChart;
-    isEditingDashboardChart: boolean;
-}> = memo(({ chartType, savedChart, isEditingDashboardChart }) => {
-    const [isOpen, setIsOpen] = useState(false);
+    isProjectPreview?: boolean;
+    isEditingDashboardChart?: boolean;
+}> = memo(
+    ({ chartType, savedChart, isProjectPreview, isEditingDashboardChart }) => {
+        const [isOpen, setIsOpen] = useState(false);
 
-    const sidebarVerticalOffset =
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/Explorer/VisualizationCard/VisualizationSidebar.tsx
```diff
@@ -16,76 +16,80 @@ import VisualizationCardOptions from '../VisualizationCardOptions';
...
+    isEditingDashboardChart?: boolean;
+}> = memo(
+    ({ chartType, savedChart, isProjectPreview, isEditingDashboardChart }) => {
+        const [isOpen, setIsOpen] = useState(false);
 
-    const sidebarVerticalOffset =
-        (isEditingDashboardChart ? 35 : 50) + // Normal header or dashboardChart header
-        (savedChart === undefined ? 0 : 80); // Include the saved chart header or not
+        const sidebarVerticalOffset =
+            (isProjectPreview && !isEditingDashboardChart ? 35 : 0) + // Preview header
```
magnew, line 26: This is the main change.

packages/frontend/src/components/Explorer/index.tsx
```diff
@@ -24,6 +26,13 @@ const Explorer: FC<{ hideHeader?: boolean }> = memo(
         );
         const { projectUuid } = useParams<{ projectUuid: string }>();
 
+        const { data: projects } = useProjects({ refetchOnMount: false });
+        const isProjectPreview = !!projects?.find(
+            (project) =>
```
magnew, line 31: This is the way we currently determine if the project is a preview

IrakliJani:
@magnew there's a failing e2e. don't think it's gonna be related. just something to be aware of

magnew:
> @magnew there's a failing e2e. don't think it's gonna be related. just something to be aware of - don't wanna block the PR because of that

It's a feature flagging issue we haven't come up with a good answer for. Cypress is looking for the old ui. For now I will turn the flag off, but I'll migrate the tests when we flip the flag permanently. 

We're gonna have these issues if we don't set cypress up to have it's own flags though.

Issue #7633, opened 2023-10-24
Do not allow custom dimensions starting with numbers


rephus:
### Description

Or it throws a SQL error 

Thread: https://lightdash.slack.com/archives/C02DGV9RWKV/p1698159508304209

rephus:
Done

PR #7639, opened 2023-10-25
fix: error custom dimension start number

rephus:
Closes:  #7633

### Description:

image

Screenshot from 2023-10-25 09-20-07

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
+++ b/packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
@@ -96,6 +96,14 @@ export const CustomDimensionModal = () => {
                 .replace(/_{2,}/g, '_') // Replace multiple underscores with a single one
                 .replace(/^_|_$/g, ''); // Remove leading and trailing underscores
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
```diff
@@ -96,6 +96,14 @@ export const CustomDimensionModal = () => {
                 .replace(/_{2,}/g, '_') // Replace multiple underscores with a single one
                 .replace(/^_|_$/g, ''); // Remove leading and trailing underscores
 
+            if (/^[0-9]/.test(sanitizedId)) {
+                form.setFieldError(
+                    'customDimensionLabel',
+                    'Custom dimension label must start with a letter',
+                );
+                return;
+            }
```
joaoviana, line 105: I would move this logic to the `validate` in `useForm` so that users can know about this rule before submitting. what do you think?

joaoviana (changes requested):
great! I would move this to the `form`'s validation step though

joaoviana (approved):

Issue #7628, opened 2023-10-24
increase error alert timeout to 1 minute


IrakliJani:
Based on user feedback, it has been identified that the current duration for displaying error messages is too short. This makes it difficult for users to read and understand the error before it disappears. To improve user experience, the display time for error messages should be extended only for errors

PR #7637, opened 2023-10-25
feat: Increased the timeout for toast errors from the default 5 sec to 1 minute.

iMac7:
Closes: #7628 

### Reviewer actions

- [x] I have manually tested the changes in the preview environment
- [x] I have reviewed the code
- [x] I understand that "request changes" will block this PR from merging

PR #7647, opened 2023-10-25
fix: increase error alert timeout to 1 minute

ZeRego:
duplicate: https://github.com/lightdash/lightdash/pull/7637

Closes: https://github.com/lightdash/lightdash/issues/7628

Thanks @iMac7 for the contribution ❤️  just made 2 small changes where we were overriding the error timeout.


ZeRego:
> LGTM! I wonder if we have a max toast number; what would happen if I click multiple times on a _bad_ sql query for example, would I see like 8 on the screen? I haven't checked this out btw!

@joaoviana we have a max of 3 

https://github.com/lightdash/lightdash/blob/64c3e61f73f6fa489b77d06864e18ed9f8bcfb1d/packages/frontend/src/components/AppToaster.tsx#L8

PR #7649, opened 2023-10-25
feat: add otel traces to metric query functions

owlas:


ZeRego (approved):
The code looks good 👍 

Would love it if we could track both sentry and otel in the same wrapper.

Issue #7617, opened 2023-10-24
Improve SSH tunnel connection settings


rephus:
### Description

Test with other ssh configuration to capture when the tunnel is disconnected

thread: https://lightdash.slack.com/archives/C02GQKJK84Q/p1698072544781529?thread_ts=1697114449.868889&cid=C02GQKJK84Q

PR #7653, opened 2023-10-25
fix: ssh reconnect

rephus:
Closes: #7617

### Description:

TL;DR it seems trying to run a postgres query with an active ssh connection while this is trying to reconnect, it is failing. 

I also upgraded pg libs

Original error: 

Screenshot from 2023-10-25 13-57-18

I fixed this by removing `reconnect` property from ssh 

Now we get the following (not server -killing error) : 

Screenshot from 2023-10-25 15-32-58

I think I also fixed one issue about PG :  capturing `connection terminated unexpectedly` from pg. This seems harder to replicate,so I'm not 100% I fixed this.

Screenshot from 2023-10-25 15-28-17

Related issues: https://github.com/brianc/node-postgres/issues/2283

### How I tested this 

I was killing my ssh periodically 

> while true; do sleep 3; sudo pkill ssh ; sudo service sshd restart ; sudo service ssh restart;  done

I was running queries periodically from terminal

> while true; do sleep 1 ; curl 'http://localhost:3000/api/v1/saved/3ae94507-60df-4585-b8ac-838e31590821/results ... ; done 

[Screencast from 25-10-23 16:27:12.webm](https://github.com/lightdash/lightdash/assets/1983672/e18e8690-14de-4048-87b5-21f1acee8fc8)

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1167,7 +1167,6 @@ export class ProjectService {
                         warehouseClient.credentials.type,
                     );
 
-                    // add 5 seconds delay with promise
                     const { rows } = await wrapSentryTransaction(
                         'runWarehouseQuery',
                         {
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -103,6 +103,14 @@ import {
     filterDimensionsFromExplore,
 } from '../UserAttributesService/UserAttributeUtils';
 
+process.on('uncaughtException', function (err) {
+    console.error((new Date).toUTCString() + ' uncaughtException:', err.message);
+    console.error(err.stack);
+    // Send the error log to your email
+    process.exit(1);
+  })
+
+  
 type RunQueryTags = {
     project_uuid?: string;
     user_uuid?: string;
@@ -1028,6 +1036,7 @@ export class ProjectService {
         queryTags?: RunQueryTags,
     ): Promise<Record<string, any>[]> {
         const tracer = opentelemetry.trace.getTracer('default');
+        const randomId = parseInt(`${Math.random() * 1000}`)
         return tracer.startActiveSpan(
             'ProjectService.runMetricQuery',
             async (span) => {
@@ -1058,9 +1067,16 @@ export class ProjectService {
                             metricQuery,
                             csvLimit,
                         );
+                        console.log('BEFORE sshTunnel',randomId)
+                        await new Promise((resolve) => setTimeout(resolve, Math.random() * 1000));
 
                     const { warehouseClient, sshTunnel } =
                         await this._getWarehouseClient(projectUuid);
+                        console.log('AFTER getExplore',randomId)
+
+                        await new Promise((resolve) => setTimeout(resolve, Math.random() * 1000));
+                        console.log('BEFORE getExplore',randomId)
+
                     const explore = await this.getExplore(
                         user,
                         projectUuid,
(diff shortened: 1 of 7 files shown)


packages/warehouses/src/ssh/sshTunnel.ts
```diff
@@ -39,7 +39,9 @@ export class SshTunnel<T extends CreateWarehouseCredentials> {
                             privateKey:
                                 this.originalCredentials.sshTunnelPrivateKey ||
                                 '',
+                            reconnect: false,
```
rephus, line 42: This is the main change

owlas:
@rephus based on you finding that `reconnect` was the problem, I was checking the `ssh2-promises` library and that reconnect feature is one that was added on top of `ssh2`. I checked the code and I think they're missing a catch in this promise chain: https://github.com/sanketbajoria/ssh2-promise/blob/b05272def4b4ac9c4575fec649d0ede078fbd2f3/src/sshConnection.ts#L315-L335

The author made the same error somewhere else too: https://github.com/sanketbajoria/ssh2-promise/issues/73

I'm wondering if we should drop `ssh2-promise` and use the base `ssh2` library and write our own promise wrappers.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1164,11 +1166,9 @@ export class ProjectService {
                             metricQuery: JSON.stringify(metricQuery),
                             type: warehouseClient.credentials.type,
                         },
-                        async () => {
-                            await sshTunnel.testConnection();
```
owlas, line 1168: what's the reason for removing the ssh test? Is it needed to open the connection to the tunnel ?

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -170,12 +170,23 @@ export class PostgresClient<
...
+                    done();
+                    return;
+                }
                 if (err) {
                     reject(err);
                     done();
                     return;
                 }
+
+                client.on('error', (e) => {
```
owlas, line 186: what does this fix?

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -170,12 +170,23 @@ export class PostgresClient<
             rows: Record<string, any>[];
         }>((resolve, reject) => {
             pool = new pg.Pool(this.config);
+
             pool.connect((err, client, done) => {
+                if (!client) {
```
owlas, line 175: when is the client undefined?

owlas (commented):
some questions

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: I added rhis test last week to mitigate the issue . It is not needed anymore.

Afaik, if reconnect is true, connection is open (reconnecting) but tunnel might be closed. If reconmect is false, connection and tunnels will be closed

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
rephus: I removed it, ignore.

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
rephus: I upgrade the library and its types. It seems connect can return undentified clients, so I'm capturing this as an error

rephus:
While I was testing, I made some extra changes, like the pg lib  upgrade. If you are not sure about thoes changes, I can revert those and only push the `reconnect: false`

owlas:
The pg.lib upgrade is good I want to understand what the new event handlers are for

Issue #7547, opened 2023-10-19
custom dashboard filter is not included in the shared url


IrakliJani:
### Description

custom dashboard filter is not included in the shared url

### To Reproduce
- navigate to any dashboard that has filters
- in view mode update custom filter
- click share button

https://github.com/lightdash/lightdash/assets/962095/47a5564a-1bcc-4abc-b5fb-ced21190b5e0

magnew:
When we fix this, we should make sure we are NOT storing the entire state of the permanent filters in the URL. Only ones that are modified. There is a related ticket about making it possible to control permanent filters from the URL and this should probably be done with the same design. 

Related ticket here: https://github.com/lightdash/lightdash/issues/7494

joaoviana:
> NOT storing the entire state of the permanent filters in the URL

Makes sense @magnew, what would you expect to be stored? The filter id (or similar) + the value?

magnew:
> > NOT storing the entire state of the permanent filters in the URL
> 
> Makes sense @magnew, what would you expect to be stored? The filter id (or similar) + the value?

Yup. We need to decide what that I'd is, but that's the idea. We could use the UUID but that would make this less usable for other scenarios.

PR #7666, opened 2023-10-26
chore: create util to construct dashboard saved filter override subparam

joaoviana:
Relates to: #7547

### Description:

When users view a dashboard, and alter its **saved** filters, they want to share a URL for that dashboard with the results they've customised. This PR covers the encoding of a difference between a saved filter and the override the user set; This string will then be added to the URL as part of an `overrides` search param. This util will then be used when a user updates a dimension/metric **saved** filter - however, we should agree on the format of these overrides. 

These overrides should: 
* include the `fieldId`
* include the `operator` (original or updated, if changed)
* include the `values` (if there's a change)
* include the `settings` (if need/changed)

**NOTE**: no logic for duplicate filters has been added yet. This PR isolates the scenario where the user updates a **saved* * filter, not loading a URL - that shall be done on next PRs

### Reviewer actions

### DIFF
--- a/packages/frontend/src/utils/dashboardSavedFiltersOverride.test.ts
+++ b/packages/frontend/src/utils/dashboardSavedFiltersOverride.test.ts
@@ -47,7 +47,7 @@ describe('dashboardSavedFiltersOverride', () => {
                     values: [false],
                 },
             ),
-        ).toBe('orders_is_completed.notNull:false');
+        ).toBe('orders_is_completed.notNull');
 
         // If no change to original filter, return null
         expect(
--- a/packages/frontend/src/utils/dashboardSavedFiltersOverride.test.ts
+++ b/packages/frontend/src/utils/dashboardSavedFiltersOverride.test.ts
@@ -1,0 +1,340 @@
+import { ConditionalOperator, DashboardFilterRule } from '@lightdash/common';
+import { createOverrideDashboardSavedFiltersUrlSubParam } from './dashboardSavedFiltersOverride';
+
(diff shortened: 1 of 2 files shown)


packages/frontend/src/utils/dashboardSavedFiltersOverride.test.ts
```diff
@@ -0,0 +1,340 @@
...
+        expect(
+            createOverrideDashboardSavedFiltersUrlSubParam(
+                originalBooleanFilter,
+                {
+                    ...originalBooleanFilter,
+                    operator: ConditionalOperator.NOT_NULL,
+                    values: [false],
+                },
+            ),
+        ).toBe('orders_is_completed.notNull:false');
```
magnew, line 50: Maybe we should make sure the boolean ones are symetrical? Like, you have 
`orders_is_completed.isNull` with no value, but `orders_is_completed.notNull:false`. 

Would it make more sense if these were either both valueless like `orders_is_completed.isNull` and `orders_is_completed.notNull`

OR

`orders_is_completed.isNull:true` and `orders_is_completed.notNull:true` ?

magnew (approved):
Nice! This looks good to me. One comment for thought, but not blocking. 

Good tests -- they make it super easy to understand how this is supposed to work.

packages/frontend/src/utils/dashboardSavedFiltersOverride.test.ts
joaoviana: > orders_is_completed.isNull

Agree with this approach! updating the code :)

Issue #7645, opened 2023-10-25
Update deliveries list to use Mantine


magnew:
### Description

Update deliveries list to use Mantine

PR #7668, opened 2023-10-26
feat: scheduled deliveries list in mantine

magnew:
Closes: #7645 

### Description:

Port scheduled deliveries list to mantine.

Before:
Screenshot 2023-10-26 at 11 18 21

After:
Screenshot 2023-10-26 at 11 18 42

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulersListItem.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulersListItem.tsx
@@ -1,23 +1,19 @@
-import { Classes } from '@blueprintjs/core';
 import {
     getHumanReadableCronExpression,
     SchedulerAndTargets,
 } from '@lightdash/common';
-import { Divider, Menu } from '@mantine/core';
-import { IconDots, IconPencil, IconTrash } from '@tabler/icons-react';
+import {
+    ActionIcon,
+    Box,
+    Group,
+    Paper,
+    Stack,
+    Text,
+    Tooltip,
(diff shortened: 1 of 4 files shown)


packages/frontend/src/features/scheduler/components/SchedulersListItem.tsx
```diff
@@ -31,47 +27,53 @@ const SchedulersListItem: FC<SchedulersListItemProps> = ({
...
+                            sx={(theme) => ({
+                                color: theme.colors.gray[4],
+                                marginTop: '-6px',
+                            })}
                         >
-                            Edit
-                        </Menu.Item>
-                        <Divider />
+                            <IconCircleFilled style={{ width: 5, height: 5 }} />
+                        </Box>
```
ZeRego, line 52: Should we use CSS to create the circle instead of an icon?
eg
```
<Box
    sx={(theme) => ({
        width: 5,
        height: 5,
        background-color: theme.colors.gray[4],
        border-radius: '50%',
    })}
/>
```

ZeRego (approved):

packages/frontend/src/features/scheduler/components/SchedulersListItem.tsx
magnew: That could work here, but the icon is probably the right idea for re-usability. I'll see if I can fix that wrapper class -- we're going to want that elsewhere.

PR #7674, opened 2023-10-26
fix: hidden columns in table appearing in downloaded csv

### LINKED ISSUE #7660: Hidden columns show up in CSV exports from table chart pane
### Description

If I have a chart formatted as a table, and one or more of the results columns are hidden in the table configuration, I would not expect those columns to appear in a CSV export triggered from the chart pane. However, CSV exports always include all columns, even if they're hidden in the UI.

### To Reproduce

1. Create a query with multiple columns
2. In the chart pane, select the "table" option
3. Hide one or more columns in the table configuration
4. Export a CSV *from the chart pane, not the results pane*
5. Oh look it's those columns you just hid

### App Version

v0.823.0

iMac7:
Fixes #7660 

The columns hidden in the UI were not used before to check for the columns to send to the api, which create the CSV. I have added the checks in the VisualizationCard.tsx itself and not in its state provider and I am not sure if this is the correct approach.

I have added the checks in the arguments provided to the downloadCSV() function to successfully download a CSV. 

Having a sort field that is not supposed to be present in the CSV breaks the downloading process, which makes me think there might be more edge cases involved. Please let me know if I missed some.

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationCard.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationCard.tsx
@@ -114,14 +114,52 @@ const VisualizationCard: FC<{
         customLabels?: Record<string, string>,
     ) => {
         if (explore?.name && unsavedChartVersion?.metricQuery && projectUuid) {
+            let filteredColumns: string[] | undefined;
+            let filteredColumnOrder: string[] | undefined;
+
+            if (
+                unsavedChartVersion.chartConfig.type === 'table' &&
+                unsavedChartVersion.chartConfig.config?.columns
+            ) {
+                const columnsConfig =
+                    unsavedChartVersion.chartConfig.config?.columns;
+                filteredColumns =
+                    unsavedChartVersion?.metricQuery.dimensions.filter(
+                        (column) =>
+                            columnsConfig[column]
+                                ? columnsConfig[column].visible !== false
+                                : true,
+                    );
+                filteredColumnOrder = columnOrder.filter(
+                    (column) => filteredColumns?.indexOf(column) !== -1,
+                );
+            }
+
             const csvResponse = await downloadCsv({
                 projectUuid,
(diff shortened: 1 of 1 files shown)


iMac7:
file:///home/mash/Videos/Screencasts/Screencast%20from%2026-10-23%2016:17:45.webm

iMac7:
[Screencast from 26-10-23 16:17:45.webm](https://github.com/lightdash/lightdash/assets/76876702/45078195-ed41-49f6-b81e-757682e2ca0c)

rephus:
Nice work. your solution works for downloading CSVs from charts

However , this is not working when downloading a CSV from a dashboard (and by extension, not from scheduler) 

I'm going to merge this and create another ticket for dashboards. 

See video: 

[Screencast from 27-10-23 15:48:22.webm](https://github.com/lightdash/lightdash/assets/1983672/48dd9ae4-2d54-4e78-8f55-a3c9e66ab4a0)

rephus (approved):
Fixes the original issue, I'll create another ticket to fix the csvs on dashboards

rephus (changes requested):
Sorry, 1 more thing, your only comment doesn't follow our `prefixes` , could you make another commit (it can be empty) so we can squash them.

iMac7:
> Sorry, 1 more thing, your only comment doesn't follow our `prefixes` , could you make another commit (it can be empty) so we can squash them.

Sorry I didn't catch that, which part of the comment doesn't follow the prefixes? Was it the PR title, or the format of explaining what the PR achieved?

rephus (approved):

rephus:
> Sorry, 1 more thing, your only comment doesn't follow our `prefixes` , could you make another commit (it can be empty) so we can squash them.

It was on the message of your first and only commit [fixed hidden columns appearing in csv](https://github.com/lightdash/lightdash/pull/7674/commits/acb6b40952dc6e177539acbda02694b636c3f9e8)

Issue #7620, opened 2023-10-24
Cache identical metric queries


owlas:
### Description

The simplest cache strategy is one that returns a previously computed set of results for the exact `MetricQuery` object that would have been sent to the warehouse.

By starting simple we can increase the complexity of the cache technology and strategies later.

### v1 cache

**ProjectService** (or any service executing queries from a MetricQuery)
- Build the `CompiledMetricQuery`
- Query the cache: `cacheService.get(compliedMetricQuery, cacheExpiry)`
- If the cache hits, return the results
- If the cache misses, loop back to the warehouse - then write to cache

**CacheService**
- Build a key from the `compiledMetricQuery`
- Lookup key in list of keys (if no key then miss)
- If hit, check the age of the cache and compare with `cacheExpiry`
- If cache is expired then miss, delete cache
- If cache is a hit then return hit

**Storage**

- Use parquet files via S3 for storage

**Future improvements**

- If the cache is queried with a `MetricQuery` that can be computed from a subset of results from another cached MetricQuery (i.e. not the exact query but with sufficient results), then we reuse the other cache. This complexity would be hidden in the cache
- Control caching for all charts on a dashboard

PR #7697, opened 2023-10-27
feat: add results cache

magnew:
Closes: #7620 

### Description:

Initial results cache work. Still in an experimental phase.

### Reviewer actions

### DIFF
--- a/packages/backend/src/clients/Aws/s3.ts
+++ b/packages/backend/src/clients/Aws/s3.ts
@@ -131,7 +131,6 @@ export class S3Service {
                 Body: results,
                 ContentType: 'application/json',
                 Metadata: metadata,
-                ACL: 'private',
             });
             const response = await this.s3.send(command);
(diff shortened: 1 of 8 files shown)


packages/backend/src/clients/Aws/s3.ts
```diff
@@ -105,6 +109,114 @@ export class S3Service {
         return this.uploadFile(csvName, csv, 'text/csv');
     }
 
+    async uploadResults(
```
owlas, line 112: Is this function actually results specific? It just seems like a generic upload

ZeRego: I would say yes because of the configuration used in those methods (cache enabled + cache bucket) 🤔 
If we make it more abstract and pass the bucket configuration via argument we would have to consider doing the same for the other methods in this class.

ZeRego (approved):

Issue #7560, opened 2023-10-20
Remove underlying data for custom dimensions


rephus:
### Description

![image](https://github.com/lightdash/lightdash/assets/1983672/dd162f19-36aa-40a7-8fd0-f177b71faef7)

rephus:
Ok, that makes sense, but should we show the `custom dimension` column if we show underlying data for a metric ?

PR #7677, opened 2023-10-26
feat: pivot table with custom dimensions

rephus:
Closes: #7563
Closes: #7641

### Description:

Screenshot from 2023-10-26 16-09-04
Screenshot from 2023-10-26 16-12-47

### Reviewer actions

- [x] I have manually tested the changes in the preview environment
- [x] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #7699, opened 2023-10-27
fix: disable underlying data custom dimension

rephus:
Closes:  #7560

### Description:

Screenshot from 2023-10-27 15-37-58
Screenshot from 2023-10-31 08-37-08

Screenshot from 2023-10-31 08-37-17
Screenshot from 2023-10-31 08-38-11
Screenshot from 2023-10-31 08-43-18
Screenshot from 2023-10-31 08-43-58

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
+++ b/packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
@@ -1,5 +1,10 @@
 import { subject } from '@casl/ability';
-import { DashboardFilters, ResultRow, ResultValue } from '@lightdash/common';
+import {
+    DashboardFilters,
+    hasCustomDimension,
+    ResultRow,
+    ResultValue,
+} from '@lightdash/common';
 import { Box, Menu, MenuProps, Portal } from '@mantine/core';
 import { useClipboard } from '@mantine/hooks';
 import { IconArrowBarToDown, IconCopy, IconStack } from '@tabler/icons-react';
@@ -47,8 +52,7 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
         return null;
     }
 
-    const { openUnderlyingDataModal /*, openDrillDownModal */ } =
-        metricQueryData;
+    const { openUnderlyingDataModal, metricQuery } = metricQueryData;
     const { track } = tracking;
 
     const canViewUnderlyingData = user.data?.ability?.can(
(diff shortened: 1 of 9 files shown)


ZeRego (approved):

ZeRego (changes requested):
We also need to remove this option from other chart types.

rephus:
> We also need to remove this option from other chart types.

I've been looking at the code, and I don't think we limit `view underlying data` based on the type (we don't have a single item/column like in the table (or table chart) , I think still makes sense to `view` based on the metric

> Could we rebase this branch once we merge https://github.com/lightdash/lightdash/pull/7677? does "view underlying data" work when results are pivoted by a bin?

It doesn't break, but it doesn't return the right results  (like you mention later) , mainly because we don't use the right filter / custom dimension on the `runUnderlyingDataQuery` method.

> Should we disable underlying data when a bin dimension is in the query? because when we see the underlying data of a metric we should add a filter "where the dimension used to bin has a value between X and Y" but that isn't happening.

Yep, you're correct, I guess this is the easiest solution for now.

packages/frontend/src/components/SimpleTable/CellContextMenu.tsx
```diff
@@ -46,36 +53,40 @@ const CellContextMenu: FC<Pick<CellContextMenuProps, 'cell'>> = ({ cell }) => {
...
-                            });
-                        }}
-                    />
-                </Can>
-            )}
+            {item &&
+                !isDimension(item) &&
+                !isCustomDimension(item) &&
+                !hasCustomDimension(metricQuery) &&
+                !hasCustomDimension(metricQuery) && (
```
ZeRego, line 60: duplicate

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
```diff
@@ -156,10 +160,11 @@ const PieChartContextMenu: FC<PieChartContextMenuProps> = ({
                     Copy
                 </Menu.Item>
 
-                {canViewUnderlyingData ? (
+                {canViewUnderlyingData && !hasCustomDimension(metricQuery) ? (
                     <Menu.Item
                         icon={<MantineIcon icon={IconStack} />}
                         onClick={handleOpenUnderlyingDataModal}
+                        title="Underlying data is not available with custom dimensions."
```
ZeRego, line 167: we don't need the title since is not visible

ZeRego (approved):
lgtm 👍  
Left a few suggestions.

packages/frontend/src/components/SimplePieChart/PieChartContextMenu.tsx
rephus: Yeah, sorry, a leftover from testing

rephus:
Unrelated failling e2e tests

Issue #7696, opened 2023-10-27
If I change the value of a saved filter in view mode, then try to export that dashboard, the filter does not apply. 


TuringLovesDeathMetal:
### Description

- Try adjusting a saved filter on a dashboard in view mode
- Export the dashboard
- see that the data is not filtered to the temporary filter value

I'd expect that the exported data would be filtered to whatever temporary filter value I've applied to my saved filter int he dashboard. 

https://www.loom.com/share/739217e0742d48349e0b2694174ee537

hamzahc1:
I've updated this to high priority as it's urgent for a customer!

PR #7704, opened 2023-10-27
fix: readd saved filter overrides

joaoviana:
Closes: #7696

### Description:

Add JSON.stringified overrides of saved filters to the dashboard url as `filters` search param; 

Reintroduces changes reverted in #7261 (which were effectively fixing an issue with URLs being too long and unprocessable by the browser), but also **only** adds these overrides if they're actually overriding the **saved** filter. 

This ensures that when users share dashboard urls + if there are overrides to **saved** filters, the URL will include these; 
This also ensures that when users override **saved** dashboard filters, these will also be reflected on `Export Dashboard` functionality. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useSavedDashboardFiltersOverrides.tsx
+++ b/packages/frontend/src/hooks/useSavedDashboardFiltersOverrides.tsx
@@ -2,7 +2,9 @@ import { DashboardFilterRule, DashboardFilters } from '@lightdash/common';
 import { useReducer } from 'react';
 import { useLocation } from 'react-router-dom';
 
-export const hasNonEmptyOverrides = (overrides: DashboardFilters | undefined) =>
+export const hasSavedFiltersOverrides = (
+    overrides: DashboardFilters | undefined,
+) =>
     !!(
         overrides &&
         (overrides.dimensions?.length > 0 || overrides.metrics?.length > 0)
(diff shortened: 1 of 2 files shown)


joaoviana:
need to test in render etc

joaoviana:
not working for `/minimal` - almost done

joaoviana:
Tested in Render: 
* just one saved filter override
* reverting saved filter overrides
* adding temp filters + saved filter override 
* if no overrides at all, URL should not have the search param
* exported dashboard in each scenario

magnew (approved):
This code makes sense and in testing it, I didn't see any problems. @hamzahc1's issues seem concerning, but I really doubt they have to do with this fix. Maybe we should try to repro them without this?

joaoviana:
> @joaoviana - this worked as expected in testing but one thing I noticed, the first time I exported, it missed some charts (see first photo): CleanShot 2023-10-30 at 10 54 43
> 
> versus the second time I exported a minute later (see below) charts rendered correctly. I don't know that this error is related to your changes, but should we open a separate issue for this behaviour? CleanShot 2023-10-30 at 10 55 30

Ah, thanks for sharing this @hamzahc1, I believe this is separate because it seems that the missing charts were moved to the bottom - on the first screenshot the last row tile: `Welcome to Lightdash!` should have been next to the tutorial (on first row second column). 

I've opened another ticket: https://github.com/lightdash/lightdash/issues/7724

joaoviana:
> This code makes sense and in testing it, I didn't see any problems. @hamzahc1's issues seem concerning, but I really doubt they have to do with this fix. Maybe we should try to repro them without this?

pretty sure this happened before and we've made some changes to improve this, but I'm having a hard time getting the correct keywords on GH search...

Issue #7717, opened 2023-10-30
Show cache information in the UI


ZeRego:
### Description

Show cache info in the UI.

For now we will only show a message in the dashboards saying some data is cached and giving a time.

### What problem does this solve?

Users know why they can't see the latest data because the chart is using cached results from X time ago.

PR #7738, opened 2023-10-30
chore: add cache info to chart results

magnew:
Relates to: #7717 

### Description:

Adds cache information to chart results. When we hit the cache we return the time the cache was updated. This will be used to inform users. 

I tested this a fair amount, but a lot of the system passes through these functions. Is there anything I should be sure to do before merging this? 

### Reviewer actions


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1024,7 +1026,7 @@ export class ProjectService {
     async getResultsForChart(
         user: SessionUser,
         chartUuid: string,
-    ): Promise<Record<string, any>[]> {
+    ): Promise<{ rows: Record<string, any>[]; cacheMetadata?: CacheMetadata }> {
```
ZeRego, line 1029: any reason for cacheMetadata to be optional ?

magnew: Nope. Will fix.

magnew:
Unit tests are (unsurprisingly) failing. Will fix.

ZeRego (approved):

magnew:
Discussed with @ZeRego -- we're going to merge this with a couple failing tests that are failing everywhere.

Issue #7736, opened 2023-10-30
Making dbt refresh button in sync with job status


joaoviana:
### Description

When users click `Refresh dbt` it triggers a job to compile the dbt model. 
This job usually takes some time, so while it's running in the background, the button changes to its loading state; 

However, users can click the button multiple times! We should display UI loading spinner as soon as the user clicks the button; 
If the user only clicks once, the job completes fine;  If multiple, then it doesn't job the loading state, but still runs in the background. However, visually, it doesn't indicate that, so users tend to trigger the `Refresh dbt` over and over again.

This should fix the UI, loading state, and current job status synchronisation. 

### To Reproduce

See scenario above ^

PR #7740, opened 2023-10-30
fix: improve refresh dbt button logic and sync UI with state

joaoviana:
Closes: #7736

### Description:

Set the loading state of `Refresh dbt button` once the user clicks it, ensuring more than 1 job gets triggered in very short time span; 
Syncs the loading state with: 
* User clicks on the button
* Job gets triggered ('started') and running ('running')
Loading state stops once job errors ('error') or is done('done')

> [!IMPORTANT]
> Also, we were invalidating the `tables/TABLE_ID` queries (so the /explores/:exploreUuid) on every success of `useJob`; Since we're refetching on a `500`ms interval, we should only invalidate `onSucess` + job status is `DONE`. This reduces calls to the network (even if cached)

Enabled state:
https://github.com/lightdash/lightdash/assets/7611706/a304c176-f4e2-41f1-a580-c83e8f4f31d0

Disabled state:
https://github.com/lightdash/lightdash/assets/7611706/5a74e301-871e-4f63-a241-f558cc949c93

Preview env state:
https://github.com/lightdash/lightdash/assets/7611706/e6e95d54-260d-43ac-b1fa-e05ceac87198

And with no permissions:
image

### Reviewer actions


packages/frontend/src/components/RefreshDbtButton/index.tsx
```diff
@@ -1,48 +1,84 @@
...
-            <Tooltip2
-                hoverCloseDelay={500}
-                interactionKind="hover"
-                content={
-                    <p>
+            <Popover withinPortal withArrow width={300}>
+                <Popover.Target>
+                    <Box
+                        sx={{
+                            cursor: 'pointer',
```
joaoviana, line 67: so the user knows they can click the container but the button is disabled.

rephus: I don't see the pointer cursor when the button is disabled :thinking:

packages/frontend/src/hooks/useRefreshServer.tsx
```diff
@@ -109,7 +109,9 @@ export const useJob = (
                 ? false
                 : 500,
```
rephus, line 110: We could reduce this frequency

rephus (approved):
Tested, works as expected :ok_hand:

packages/frontend/src/components/RefreshDbtButton/index.tsx
joaoviana: oh! Checking!

joaoviana: https://github.com/lightdash/lightdash/assets/7611706/4fec0583-141c-42b9-8980-9813cfa8811b

interesting ^ I still see it...

PR #7741, opened 2023-10-30
feat: show dashboard cache data

magnew:
Closes: #7717 

### Description:

Show a message about caching on dashboards with cached charts. 

image

### Reviewer actions


packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -109,6 +113,7 @@ export const DashboardProvider: React.FC = ({ children }) => {
         useState<DashboardFilters>(emptyFilters);
     const [haveFiltersChanged, setHaveFiltersChanged] =
         useState<boolean>(false);
+    const [resultsCacheTimes, setResultsCacheTimes] = useState<any[]>([]);
```
ZeRego, line 116:
```suggestion
    const [resultsCacheTimes, setResultsCacheTimes] = useState<CacheMetadata[]>([]);
```

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -429,6 +434,19 @@ export const DashboardProvider: React.FC = ({ children }) => {
...
+            setResultsCacheTimes((old) => [
+                ...old,
+                cacheMetadata.cacheUpdatedTime,
+            ]);
+        }
+    }, []);
+
+    const getOldestCacheTime = useCallback(() => {
+        return min(resultsCacheTimes);
+    }, [resultsCacheTimes]);
```
ZeRego, line 448: What is the advantage of returning a callback instead of the oldest cache time value?

ZeRego (approved):

packages/frontend/src/providers/DashboardProvider.tsx
magnew: Good catch

packages/frontend/src/providers/DashboardProvider.tsx
magnew: None really as is. I often like getters as a pattern cause they are flexible to add to, but it's not doing anything here. I can make it a value if you prefer.

Issue #6730, opened 2023-08-14
Migrate String-type Filter inputs to Mantine


joaoviana:
### Description

In: `packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx`

`MultiAutoComplete`: 

When: 
```
 case FilterOperator.STARTS_WITH:
        case FilterOperator.ENDS_WITH:
        case FilterOperator.INCLUDE:
        case FilterOperator.NOT_INCLUDE:
        case FilterOperator.EQUALS:
        case FilterOperator.NOT_EQUALS: {
       
```
Migrate `NumericInput` Filter input to Mantine ahead of the work to be done on: https://github.com/lightdash/lightdash/issues/6490

screenshot reference: 

<img width="371" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/3e4220dd-070e-470d-b2a2-fd647c99b971">

### What problem does this solve?

Supports the complete migration of the Filter UI

PR #7742, opened 2023-10-30
feat: filter inputs - string autocomplete to mantine

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/6730

### Description:

CleanShot 2023-10-31 at 01 26 23@2x

CleanShot 2023-10-31 at 01 26 38@2x

CleanShot 2023-10-31 at 01 27 18@2x

CleanShot 2023-10-31 at 01 27 31@2x

### Reviewer actions


IrakliJani:
@magnew I think remaining tests that fail are related to the chart config sidebar change. not sure why it's enabled here 🤔

magnew:
> @magnew I think remaining tests that fail are related to the chart config sidebar change. not sure why it's enabled here 🤔

My bad. I turned it on to test something. Will disable and re-run.

packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/FilterMultiAutoComplete.tsx
```diff
@@ -0,0 +1,151 @@
...
+        search,
+        initialSuggestionData,
+        projectUuid,
+        field,
+        autocompleteFilterGroup,
+        true,
+        { refetchOnMount: 'always' },
+    );
+
+    const results = useMemo(() => [...resultsSet], [resultsSet]);
```
magnew, line 57: No big deal, but does this useMemo do anything? It seems like it just keeps an updated copy of the results, but you could just use them directly, right? Maybe I am missing something and this is critical :D

packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/FilterMultiAutoComplete.tsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+    const handleOnClose = useCallback(() => {
+        setSearch('');
+    }, [setSearch]);
+
+    const data = useMemo(() => {
+        // Mantine does not show value tag if value is not found in data
+        // so we need to add it manually here
+        // also we are merging status indicator as a first item
```
magnew, line 81: Isn't this what onCreate is for? This probably works too, but i think we usually add them to the data in onCreate. Which is maybe a little clearer and doesnt need an extra memo.

magnew (approved):
Looks good! Left a couple non-blocking questions.

packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/FilterMultiAutoComplete.tsx
IrakliJani: oh it returns a set not an array. I think it was like that before let me check

packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/FilterMultiAutoComplete.tsx
IrakliJani: so, `data` prop is basically `options` - what options show up in the dropdown. and `value` is for an array of selected options from data. if we have 

data of = [1,2,3]
and values of = [2,3,4]
then it only displays 2,3 as tags and omits 4

hope this explanation makes sense

packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/FilterMultiAutoComplete.tsx
magnew: No problem, it seems ok, just a little strange

packages/frontend/src/components/common/Filters/FilterInputs/AutoComplete/FilterMultiAutoComplete.tsx
magnew: Ok, yea, I see. Thank you.

magnew (approved):

IrakliJani:
@magnew I was doing one more round of testing and it turns out I forgot onPaste function which I just added. I also renamed component and did some cleanups

Issue #7744, opened 2023-10-31
Custom Metrics filter options kept hidden


Ankush263:
### Description

https://github.com/lightdash/lightdash/assets/86042508/8e717c0e-afdb-456b-bee1-6ef1858c65db

IrakliJani:
duplicate of https://github.com/lightdash/lightdash/issues/7337

PR #7745, opened 2023-10-31
fix: Custom Metrics filter options kept hidden

Ankush263:
Closes: #7744

### Reviewer actions


joaoviana:
HI @Ankush263, thanks for contributing to this PR but there's a few more things missing: 
* On any dropdown-like component, we should address the z-index or render these in a React portal with the `withinPortal` . I'm currently looking into this so will have to close the PR. Thanks for understanding 😃

Issue #7613, opened 2023-10-23
custom dimension is wrongly selecting dimensions


rephus:
### Description

If a custom dimension matches the name of the dimension, the explorer sidebar thinks the dimension is selected

![image](https://github.com/lightdash/lightdash/assets/1983672/25d303ba-cfd0-46bc-ab12-c6342bc78482)

adarshlilha:
Hi @rephus 
I came across Lightdash and trying to dig into this issue, can I take this up if available?

A help on how to create a custom dimension would be super useful, TIA

joaoviana:
Hi @adarshlilha I'm actually looking into this! Thanks for the initiative, but this is a milestone that @rephus and I are collaborating on 👍

PR #7755, opened 2023-10-31
fix: create consistent id/names for custom dimensions

joaoviana:
Closes: #7613

### Description:

When creating a custom dimension with the same name as another dimension, this could result in confusing UI behaviour. For example: 
* User has a dimension "payment id" from table "payments"
* User created custom dimension "payments - payment id" 
* User sees both the custom dimension and the base dimension selected

On the explore tree, items are selected if their `id` is in the selected items array. So, in order to avoid adding extra checks to where this happens - in `TreeSingleNode` - let's always prefix the `id` of a custom dimension with the base custom dimension + the new label 

so for this custom dimension:
image
its `id:"order_id_payments_payment_id"` which matches what is also done for custom metrics on here for example: https://github.com/lightdash/lightdash/blob/fix/custom-dimensions-incorrectly-selected-state/packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx#L79

### Reviewer actions


rephus:
YOu'll have to fix e2e tests

joaoviana:
> YOu'll have to fix e2e tests

oops, thanks! Just updated 👍  waiting for them to pass

rephus (approved):
Cool, yeah this should fix the issue. I think we should try to apply some `good practices` here anyway, I don't think customers should name `their bins` with the same `dimension id`, instead they should use `... bins` or `... range`  

Note: this should not break existing custom dimensions, this will only affect custom dimensions created

Issue #4286, opened 2023-01-26
arm64 builds in dockerhub


ckdake:
### Description

It would be helpful to have arm64 builds of lightdash available in dockerhub

### What problem does this solve?

When using docker on M* macs to run lightdash, having arm64 builds available would save doing the local tradeoff of building locally or taking the performance hit of running the amd64 container through emulation.

arm64 would also facilitate on running on ARM compute in AWS which seems to be faster and more cost performant these days. Thanks!

ckdake:
Still a thing!

ckdake:
Yes, still relevant.

wasd171:
I would be happy to work on it

wasd171:
@owlas I see that arm64 builds were reverted in https://github.com/lightdash/lightdash/pull/7844, could I improve something to enable them again?

mmospanenko:
yeah, it is important to have arm64, we have whole infra on arm(

CincyBC:
I was looking forward to giving Lightdash a try but we have a strict policy to use Graviton/ARM processors. Is there any reason this was reverted and requests for it to be added back go ignored?

torimwhaley:
Reopening this. It was auto-closed when #7766 merged in November 2023, but #7844 reverted the arm builds the same afternoon to stabilize the release pipeline, so no arm64 image is published today and the "completed" state on this issue was misleading.

There is renewed demand. A self-hosted enterprise customer is migrating to arm64-based Kubernetes clusters and needs a native arm64 image to keep deploying Lightdash, since their arm nodes cannot run the amd64 image. For them this is a hard blocker on their infrastructure migration rather than a nice-to-have.

Worth noting the 2023 rollback was a CI stability call at the time, not an arm-specific technical blocker, and the build setup has changed substantially since then. One suggestion raised internally is building both architectures under the same release tag via Depot, at the cost of longer build times. Suggested approach only, implementer's call.

PR #7766, opened 2023-11-01
feat: Implement multi-arch Docker builds in the CI

wasd171:
Closes: https://github.com/lightdash/lightdash/issues/4286

### Description:
This PR adds arm64 build to the Github Actions

### Reviewer actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Deploy to docker hub
+name: Deploy to Docker Hub
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -45,3 +50,4 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          platforms: 'linux/amd64,linux/arm64'
```
HectorCastelli, line 53: This approach runs each build sequentially and can be a little slow as you start adding new platforms.

Would you be comfortable making a change to follow [this approach](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners) instead? It allows for builds to run in parallel and will keep the time to publish pretty much unchanged, regardless of how many platforms you choose to build.

wasd171: Sure, let me update this PR

owlas: We want to build docker images ourselves in github and then push to docker hub, because builds in docker hub are painfully slow. We'll hold off this change until then

owlas (approved):

Issue #7274, opened 2023-09-29
cached_explores: Add foreign key to project UUID


joaoviana:
### Description

- [ ] Add foreign key to `project_uuid`
- [ ] Ensure we never create a new row without this foreign key

### What problem does this solve?

Facilitates the migration to Separate Explores #7162

amanx123:
Hey , I would like work on this. Should i go ahead with the pr or you want to assign first ?

PR #7770, opened 2023-11-01
fix: add missing foreign key to cached explores table

ZeRego:
Closes: #7274 

### Description:

- delete all cached explores for projects that no longer exist ( eg: preview projects ) 
- add foreign key to project table so the rows are deleted when a project is deleted

Force tests:

test-frontend
test-cli

### Reviewer actions

### DIFF
--- a/packages/backend/src/database/migrations/20231101164443_add_cache_explores_foreign_key.ts
+++ b/packages/backend/src/database/migrations/20231101164443_add_cache_explores_foreign_key.ts
@@ -1,0 +1,34 @@
+import { Knex } from 'knex';
+
+const CACHED_EXPLORES_TABLE_NAME = 'cached_explores';
+
+export async function up(knex: Knex): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/backend/src/database/migrations/20231101164443_add_cache_explores_foreign_key.ts
```diff
@@ -0,0 +1,34 @@
+import { Knex } from 'knex';
+
+const CACHED_EXPLORES_TABLE_NAME = 'cached_explores';
+
+export async function up(knex: Knex): Promise<void> {
```
owlas, line 5: everything in this block runs as a transaction right?

owlas (approved):

packages/backend/src/database/migrations/20231101164443_add_cache_explores_foreign_key.ts
ZeRego: yes

Issue #7718, opened 2023-10-30
User can clear cached results


ZeRego:
### Description

Add option to force fetch results from warehouse
- How are going to show this?

eg:
1. User click button to "clear cache
2. FE fetches the chart results again with URL param ?cache=false
3. BE fetches results from warehouse and updates cache

### What problem does this solve?

When the chart is using cached results but the user wants the latest results.

PR #7772, opened 2023-11-01
feat: refresh button clears results cache

magnew:
Closes: #7718 

### Description:

This is a proof of concept for clearing the results cache with the refresh button. This doesn't have:

- a good way to turn caching back on

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/savedChartController.ts
+++ b/packages/backend/src/controllers/savedChartController.ts
@@ -48,13 +48,14 @@ export class SavedChartController extends Controller {
         this.setStatus(200);
         return {
             status: 'ok',
-            results: await projectService.runViewChartQuery(
-                req.user!,
+            results: await projectService.runViewChartQuery({
+                user: req.user!,
                 chartUuid,
-                body.filters,
-                undefined,
-                req.query?.invalidateCache?.toString() === 'true',
-            ),
+                filters: body.filters,
+                versionUuid: undefined,
+                invalidateCache:
+                    req.query?.invalidateCache?.toString() === 'true',
+            }),
         };
     }
 
@@ -122,12 +123,12 @@ export class SavedChartController extends Controller {
         this.setStatus(200);
         return {
             status: 'ok',
-            results: await projectService.runViewChartQuery(
-                req.user!,
+            results: await projectService.runViewChartQuery({
+                user: req.user!,
                 chartUuid,
-                undefined,
+                filters: undefined,
                 versionUuid,
-            ),
+            }),
         };
     }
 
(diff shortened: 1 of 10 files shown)


packages/backend/src/controllers/savedChartController.ts
```diff
@@ -52,6 +52,8 @@ export class SavedChartController extends Controller {
                 req.user!,
                 chartUuid,
                 body.filters,
+                undefined,
+                req.query?.invalidateCache?.toString() === 'true',
```
magnew, line 56: Is there a better way than this to get query parameters from the url?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -900,6 +901,7 @@ export class ProjectService {
                 : QueryExecutionContext.CHART,
 
             queryTags,
+            invalidateCache,
```
magnew, line 904: These should all probably become named parameters

owlas:
Could we make this a parameter of when we `mutate()` to get new results? i.e. on `runQuery()` we would do `runQuery(useCache?)`

packages/frontend/src/hooks/useQueryResults.tsx
```diff
@@ -163,8 +167,17 @@ export const useUnderlyingDataResults = (
...
+export const useChartResults = (
+    chartUuid: string,
+    filters?: Filters,
+    invalidateCache?: boolean,
+) => {
+    const queryKey = [
+        'savedChartResults',
+        chartUuid,
+        JSON.stringify(filters),
+        invalidateCache,
```
owlas, line 179: @magnew do we actually need to track the cache in the state? I don't think we need the charts to rerun when this status changes. We just need the charts to re-run when the user hits refresh

magnew: Right, this is the part where we are fighting with react query a bit. When you hit refresh, we need to invalidate the queries. One way to do that is to put some cache indicator in the query key. That's why I'm keeping it here, but I would like to isolate it from the view code.

packages/backend/src/controllers/savedChartController.ts
```diff
@@ -48,11 +48,14 @@ export class SavedChartController extends Controller {
...
-                req.user!,
+            results: await projectService.runViewChartQuery({
+                user: req.user!,
                 chartUuid,
-                body.filters,
-            ),
+                filters: body.filters,
+                versionUuid: undefined,
+                invalidateCache:
+                    req.query?.invalidateCache?.toString() === 'true',
```
magnew, line 57: @ZeRego is there a better way to get this query arg? A decorator or something?

magnew: Done

ZeRego (approved):

Issue #7776, opened 2023-11-02
If I login on google using a different account for gdrive, I don't get the right authentication back 


rephus:
### Description

And the oauth does not end well. 

Instead, it will create 2 openids , one with refresh_token
![image](https://github.com/lightdash/lightdash/assets/1983672/65abe13f-297a-48bd-893e-7715137445ab)

We should simply get the one with user_id and refresh_token

PR #7779, opened 2023-11-02
fix: multiple oauth accounts

rephus:
Closes:  #7776
### Reviewer actions

### DIFF
--- a/packages/backend/src/models/UserModel.ts
+++ b/packages/backend/src/models/UserModel.ts
@@ -696,6 +696,7 @@ export class UserModel {
                 'openid_identities.user_id',
             )
             .where('user_uuid', userUuid)
+            .whereNotNull('refresh_token')
             .select('refresh_token');
 
         if (!row) {


ZeRego (approved):

owlas (changes requested):
If I authenticate with Google-1 and then authenticate with Google-2, it might still use my previous Google-1 credentials. That would be confusing. We need to do *one* of the following:

- unset all 'refresh_tokens' other than the one that is set
- add a flag for 'active_refresh_token'
- add a new table for refresh_tokens with a fk to identities

rephus:
> If I authenticate with Google-1 and then authenticate with Google-2, it might still use my previous Google-1 credentials

If you authenticate with google-1 on `login` (oauth/google) : you will have `refresh_token` null, because that authentication path doesn't ask for refresh_token or extra scopes. 
If you authenticate with google-2 on `google sync` (oauth/gdrive) you will  ask for extra scopes and your refresH-token will be stored in db. 
This means you have 2 openids
This PR is fixing this scenario, by returning the `refresh_token is not null`

image

@owlas is that what you mean ?

owlas:
@rephus I mean actually mean if I authenticate with google drive twice:

- Step 1 - I click to export google sheets
- Step 2 - I'm asked to link a Google account. I've got 2 accounts Google-1 and Google-2. I want to connect Google-2
- Step 3 - I **accidentally** click on Google-1 !
- Step 4 - I need to select Google-2 !
- Step 5 - I re-authenticate with Google-2 for google drive
- Step 6 - Lightdash will continue to use Google-1 as the authentication source

I just realised from your comment though that we haven't implemented step 5. So it's impossible. In fact it's a bigger problem that if I accidentally choose the wrong account I can't even fix it.

> We should have a way to logout or select a different account on gdrive (this can remove the openid with the refresh_token), I think there is a ticket for this.

owlas (approved):
fixes the issue 👍

Issue #7737, opened 2023-10-30
Validator is not catching errors with joined tables


TuringLovesDeathMetal:
### Description

- delete a join which breaks a chart
- rerun the validator
- see that there's no validation error shown

https://www.loom.com/share/9d281871e00f4f75b816d912b263df12

PR #7786, opened 2023-11-02
fix: validation on joined fields

rephus:
Closes: #7737

### Description:

Screenshot from 2023-11-02 13-19-12
Screenshot from 2023-11-02 13-19-24
Screenshot from 2023-11-02 13-23-02

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.ts
@@ -194,6 +194,7 @@ export class ValidationService {
     private async validateCharts(
         projectUuid: string,
         explorerFields: CompiledField[],
+        explores: Explore[],
     ): Promise<CreateChartValidation[]> {
         const explorerDimensionIds = explorerFields
             .filter(isDimension)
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -244,7 +245,20 @@ export class ValidationService {
                 CreateChartValidation,
                 'error' | 'errorType' | 'fieldName'
             >) => {
-                if (!fieldIds?.includes(fieldId)) {
+                const { tableName } = chart;
+                const joinedTables =
+                    explores
+                        .find((e) => e.name === tableName)
+                        ?.joinedTables.map((jt) => jt.table) || [];
```
ZeRego, line 252: I think there will be issues with join alias.  Could we add unit tests for this?

I was looking at the code and expected the `fieldIds` argument to only have valid values. 
And having all the `explores` as an argument defeats the purpose of the `explorerFields: CompiledField[],` argument. 

Should we precalculate all the valid field ids per explore and have the argument `exploreFields: Record<string, { dimensionIds: string[]; metricIds: string[]}>` ?
This way we only do it once per explore while now we do it multiple times per chart and field.

rephus: Added tests and refactored code

ZeRego (approved):

Issue #7784, opened 2023-11-02
Allow minimal dashboards to have context of scheduler


joaoviana:
### Description

TBD

PR #7787, opened 2023-11-02
chore: pass schedulerUuid param to minimal dashboard if it exists

joaoviana:
Closes: #7784 

### Description:

Pass `schedulerUuid` to minimal dashboard params on scheduled deliveries if it exists;
We then can use the scheduler context on Minimal dashboards like so:
```tsx
const { search } = useLocation();
    const searchParams = new URLSearchParams(search);
    const schedulerUuid = searchParams.get('schedulerUuid');
    const { data: scheduler, isLoading } = useScheduler(schedulerUuid ?? '');
```

### Reviewer actions

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -94,7 +94,7 @@ const getChartOrDashboard = async (
             minimalUrl: `${lightdashConfig.siteUrl}/minimal/projects/${
                 dashboard.projectUuid
             }/dashboards/${dashboardUuid}${
-                schedulerUuid ? `?schedulerId=${schedulerUuid}` : ''
+                schedulerUuid ? `?schedulerUuid=${schedulerUuid}` : ''
             }`,
             details: {
                 name: dashboard.name,
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -65,6 +65,7 @@ import {
 const getChartOrDashboard = async (
     chartUuid: string | null,
     dashboardUuid: string | null,
+    schedulerUuid: string | undefined,
 ) => {
     if (chartUuid) {
         const chart = await schedulerService.savedChartModel.getSummary(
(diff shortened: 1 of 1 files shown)


packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -119,14 +125,21 @@ export const getNotificationPageData = async (
     let csvUrl;
     let csvUrls;
     let pdfFile;
+
+    const schedulerUuid =
+        'schedulerUuid' in scheduler &&
+        typeof scheduler.schedulerUuid === 'string'
+            ? scheduler.schedulerUuid
+            : undefined;
```
joaoviana, line 133: added all these checks so it conforms to the backwards compatibility mentioned in: https://github.com/lightdash/lightdash/blob/chore/allow-for-minimal-to-have-access-to-scheduler-uuid/packages/backend/src/scheduler/SchedulerTask.ts#L295
But the `schedulerUuid` value is always set on new and recent scheduled deliveries

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -87,9 +88,14 @@ const getChartOrDashboard = async (
...
             dashboardUuid,
         );
+
         return {
             url: `${lightdashConfig.siteUrl}/projects/${dashboard.projectUuid}/dashboards/${dashboardUuid}/view`,
-            minimalUrl: `${lightdashConfig.siteUrl}/minimal/projects/${dashboard.projectUuid}/dashboards/${dashboardUuid}`,
+            minimalUrl: `${lightdashConfig.siteUrl}/minimal/projects/${
+                dashboard.projectUuid
+            }/dashboards/${dashboardUuid}${
+                schedulerUuid ? `?schedulerId=${schedulerUuid}` : ''
```
IrakliJani, line 97: should param also be `schedulerUuid` as well for consistency?

IrakliJani (approved):
one comment, otherwise looks good!

packages/backend/src/scheduler/SchedulerTask.ts
joaoviana: oops! good catch, renaming!

Issue #7162, opened 2023-09-21
Separate cached explores


joaoviana:
### Description

Currently, we store all the explores in one JSON blob. 
We could separate them and index them to optimise DB queries so that `explores/:exploreName` API endpoints run faster

### What problem does this solve?

Faster processing of saved charts results and queries that depend on getting these explores.

joaoviana:
See Sentry for `availableFilters` that includes a DB query to the `cached_explores` https://lightdash.sentry.io/issues/4461022655/?project=5959292&query=is%3Aunresolved&referrer=issue-stream&stream_index=2

PR #7788, opened 2023-11-02
feat: cache explores individually

ZeRego:
Closes: #7162 

### Description:

- keep existing table that caches all explores
- save explores individually in a new table
- if doesn't find explore by name in new table, fallback to old table
- add otel tracking

### Reviewer actions

### DIFF
--- a/packages/backend/src/database/migrations/20231102094206_add_cached_explore_table.ts
+++ b/packages/backend/src/database/migrations/20231102094206_add_cached_explore_table.ts
@@ -1,0 +1,29 @@
+import { Knex } from 'knex';
+
+const CACHED_EXPLORE_TABLE_NAME = 'cached_explore';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(CACHED_EXPLORE_TABLE_NAME))) {
+        await knex.schema.createTable(CACHED_EXPLORE_TABLE_NAME, (table) => {
+            table
+                .uuid('cached_explore_uuid')
+                .primary()
+                .notNullable()
+                .defaultTo(knex.raw('uuid_generate_v4()'));
+            table
+                .uuid('project_uuid')
+                .references('project_uuid')
+                .inTable('projects')
+                .notNullable()
+                .onDelete('CASCADE');
+            table.string('name').notNullable();
+            table.specificType('table_names', 'TEXT[]').notNullable();
+            table.jsonb('explore').notNullable();
+            table.unique(['name', 'project_uuid']); // can't have duplicate explore names in the same project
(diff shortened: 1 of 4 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -569,57 +570,121 @@ export class ProjectModel {
...
-            .onConflict('project_uuid')
-            .merge()
-            .returning('*');
-        return cachedExplores;
+        return wrapOtelSpan(
+            'ProjectModel.saveExploresToCache',
+            {},
+            async () => {
+                // cache explores individually
+                await this.database(CachedExploreTableName)
```
owlas, line 662: Doesn't this change the existing behaviour.

If I previously pushed 3 explores. Then afterwards I push just 2 explores. Don't I still end up with all 3 explores? 

You should delete all the explores before inserting the new ones. So if the user pushes a new project with fewer explores, they'll be removed from the database.

Maybe I've missed something.

packages/backend/src/database/migrations/20231102094206_add_cached_explore_table.ts
```diff
@@ -0,0 +1,29 @@
...
+                .primary()
+                .notNullable()
+                .defaultTo(knex.raw('uuid_generate_v4()'));
+            table
+                .uuid('project_uuid')
+                .references('project_uuid')
+                .inTable('projects')
+                .notNullable()
+                .onDelete('CASCADE');
+            table.string('name').notNullable();
```
rephus, line 19: should we add an index on `name` ?

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -569,57 +570,125 @@ export class ProjectModel {
...
+                    .where('project_uuid', projectUuid)
+                    .delete();
+                // cache explores individually
+                await this.database(CachedExploreTableName)
+                    .insert(
+                        explores.map((explore) => ({
+                            project_uuid: projectUuid,
+                            name: explore.name,
+                            table_names: Object.keys(explore.tables || {}),
```
rephus, line 671:
```suggestion
                            table_names: Object.keys(explore.tables || []),
```
Database is storing  `Text[]` should this be [] instead of {} ?

rephus (approved):
Looks good, I just added a couple of questions

packages/backend/src/models/ProjectModel/ProjectModel.ts
ZeRego: this value is inside `Object.keys()`

Issue #4425, opened 2023-02-09
Custom filters on scheduled delivery dashboards


owlas:
### Description

When I create a scheduled delivery for a dashboard, there is an option to update the existing dashboard filters in the scheduled delivery

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVNUPGKr0=/?share_link_id=792869617470

- [x] There is a third tab in the scheduled delivery modal called `filters`
- [x] The filters tab includes a list of all of the existing dashboard filters
- [x] You cannot add more filters or delete filters from the dashboard, only adjust the values. 
- [x] When you click on a filter, you see the same "adjustment" modal as appears when you're in `view` mode on a dashboard
- [ ] The filters are listed using the `label` value (not the filter field value). 
- [ ] The `label` value appears when you go to edit the filter (this should also exist in view mode in a dashboard when we do this issue: #7746, so it will match `view mode` filter interaction in a dashboard once that's changed - or we can fix the component in dashboard view mode and then just reuse it here!)
- [x] Adjusting the filter has the same behaviour as adjusting saved filters in a dashboard. E.g.:
  - [x] if the filter is applied to different fields on different tiles, tehn this is still the case for the scheduled delivery
  - [x] all the filter conditions are joined using `AND` 
  - [x] You can override the operator
  - [x] you can override the value
  - [x] you CANNOT override the field that's being filtered
- [x] temporary filters that you add in `view` mode are _**not**_ included _**or**_ able to be adjusted in the scheduled delivery  (they _**have**_ to be saved filters)
- [x] If you remove a saved filter from a dashboard, then it would also get removed fromt eh scheduled delivery (even if you overrode it or something)

TuringLovesDeathMetal:
Nice! I've added a spec for this here: https://miro.com/app/board/uXjVP1D_L4w=/?moveToWidget=3458764542491754845&cot=14

@owlas I have one question about how we would redirect people to these charts/dashboards with added filters though: https://miro.com/app/board/uXjVP1D_L4w=/?moveToWidget=3458764546051005231&cot=14

joaoviana:
My next focus: passing the `filters` to the Scheduled Deliveries form and sending that info to the API endpoint

joaoviana:
Done and rolled out to all users

PR #7794, opened 2023-11-02
feat: add filters to schedulers

joaoviana:
Relates to: #4425

### Description:

Adds migration file to add `filters` column to a scheduler
Updates scheduler DB with `filters` type and adds type guard `isDashboardScheduler`

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SchedulerModel/index.ts
+++ b/packages/backend/src/models/SchedulerModel/index.ts
@@ -3,6 +3,7 @@ import {
     CreateSchedulerLog,
     isChartScheduler,
     isCreateSchedulerSlackTarget,
+    isDashboardScheduler,
     isSlackTarget,
     isUpdateSchedulerEmailTarget,
     isUpdateSchedulerSlackTarget,
@@ -64,6 +65,7 @@ export class SchedulerModel {
             dashboardUuid: scheduler.dashboard_uuid,
             format: scheduler.format,
             options: scheduler.options,
+            filters: scheduler.filters,
         } as Scheduler;
     }
 
@@ -211,6 +213,11 @@ export class SchedulerModel {
                     dashboard_uuid: newScheduler.dashboardUuid,
                     updated_at: new Date(),
                     options: newScheduler.options,
+                    filters:
+                        isDashboardScheduler(newScheduler) &&
+                        newScheduler.filters
+                            ? JSON.stringify(newScheduler.filters)
+                            : null,
                 })
                 .returning('*');
             const targetPromises = newScheduler.targets.map(async (target) => {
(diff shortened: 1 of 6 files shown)


packages/backend/src/models/SchedulerModel/index.ts
```diff
@@ -211,6 +213,11 @@ export class SchedulerModel {
                     dashboard_uuid: newScheduler.dashboardUuid,
                     updated_at: new Date(),
                     options: newScheduler.options,
+                    filters:
+                        isDashboardScheduler(newScheduler) &&
+                        newScheduler.filters
+                            ? JSON.stringify(newScheduler.filters)
+                            : null,
```
ZeRego, line 220: What is the reason for the conditions to be different between create and update?

packages/common/src/types/scheduler.ts
```diff
@@ -76,9 +77,15 @@ export type ChartScheduler = SchedulerBase & {
...
 };
+
+export const isDashboardScheduler = (
+    scheduler: Scheduler | CreateSchedulerAndTargets,
+): scheduler is DashboardScheduler => scheduler.dashboardUuid !== undefined;
+
 export type DashboardScheduler = SchedulerBase & {
     savedChartUuid: null;
     dashboardUuid: string;
+    filters?: DashboardFilterRule[];
```
ZeRego, line 88: Do we need to update the API schema?

joaoviana: thanks for reminding me, pushed 👍

packages/backend/src/models/SchedulerModel/index.ts
joaoviana: When creating, the type includes `dashboardUuid`: https://github.com/lightdash/lightdash/blob/feat/add-filters-to-schedulers/packages/common/src/types/scheduler.ts#L140
When updating, the type does not include `dashboardUuid` so I can't use the type guard :/  see line 151-153 in `scheduler.ts` below on the PR ⬇️

packages/common/src/types/scheduler.ts
joaoviana: hmm it didn't make changes to the routes 🤔

ZeRego (approved):

packages/backend/src/models/SchedulerModel/index.ts
```diff
@@ -211,6 +213,11 @@ export class SchedulerModel {
                     dashboard_uuid: newScheduler.dashboardUuid,
                     updated_at: new Date(),
                     options: newScheduler.options,
+                    filters:
+                        isDashboardScheduler(newScheduler) &&
+                        newScheduler.filters
+                            ? JSON.stringify(newScheduler.filters)
```
IrakliJani, line 219: do we need to manually stringify?

IrakliJani (commented):
@joaoviana. looks good! I know this is merged but just had a look before we start integrating FE/BE. added one question regarding jsonb requiring to manually stringify

PR #7796, opened 2023-11-02
feat: scheduled deliveries filter state

IrakliJani:
Related to: https://github.com/lightdash/lightdash/issues/4425

### Description:

adds state management to scheduled delivery filters

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
@@ -1,4 +1,5 @@
 import {
+    ConditionalOperator,
     Dashboard,
     DashboardFilterRule,
     FilterType,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -24,19 +26,29 @@ import {
 import { useProject } from '../../../hooks/useProject';
 import { useDashboardContext } from '../../../providers/DashboardProvider';
 
+type SchedulerFilterRule = DashboardFilterRule & {
+    tileTargets: undefined;
```
joaoviana, line 30: do we want it `undefined` or optional? 
maybe... `optional`? so we don't need to add this line. What do you think?

joaoviana (approved):
one Q on types 👍

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
IrakliJani: @joaoviana good question! ok so I was thinking about this a lot. whenever you update filter it returns new `DashboardFilterRule` and it includes `tileTargets`.

typescript does not complain whenever you have additional properties in the object so it is really easy to forget removing that property and thought to make it explicit - `SchedulerFilterRule` should  contain `tileTargets` equal to undefined.

let me know if this is useless and we should take care of it on the BE side?

joaoviana: ahhhhh! I see!! Amazing, let's keep your change! 👍

Issue #7662, opened 2023-10-25
Re-renders in the chart config reference line component cause focus to be lost


magnew:
### Description

When creating a reference line, incorrect re-renders cause the inputs to lose focus. 

Note that this happens both in the sidebar and config drop-downs. 

### To Reproduce

- Add a reference line to a chart
- Spend some time trying to edit and change the labels
- The inputs will sometimes lose focus 😞

PR #7798, opened 2023-11-03
fix: update reference line label on blur

rephus:
Closes: https://github.com/lightdash/lightdash/issues/7662

### Description:

We previously used a `debounce` to mitigate this issue, because re-render makes it quite slow. That was a bit of a hack, instead I moved the render logic to `onblur` for label. 

[Screencast from 03-11-23 09:07:05.webm](https://github.com/lightdash/lightdash/assets/1983672/34d84b7b-61a2-4640-9bb5-4b246e9bcd5c)

### Reviewer actions


magnew (approved):
Nice.

Issue #7795, opened 2023-11-02
Change the 'Edit chart' button to be a Mantine one


PriPatel:
<img width="201" alt="Screenshot 2023-11-02 at 19 58 29" src="https://github.com/lightdash/lightdash/assets/12776522/8055bed4-6f9b-4b6d-8438-f1f20c8f7ea4">

The three dot menu has been updated to Mantine, so now the buttons are different heights.

- [ ] Update button
- [ ] Update pen icon to be a Tabler one

Ankush263:
hey @PriPatel 👋, can I take this task...???

PriPatel:
@Ankush263 yes, go for it!

PR #7811, opened 2023-11-03
feat: update chart related buttons

joaoviana:
Closes: #7795

### Description:

Update saved chart headers in explore etc

https://github.com/lightdash/lightdash/assets/7611706/4467e213-3343-44f9-8450-0aa7560a3738

### Reviewer actions


joaoviana:
I'm getting a not-so-good button style rendering when editing a chart from a dashboard:
* the button `Save changes` should be disabled by default when editing a chart from a dashboard on initial load; it goes green, then disabled.
Will check on Render if that's the case or just my local set up that has issues 😅

ZeRego:
@joaoviana
Screenshot 2023-11-03 at 17 40 36

joaoviana:
> @joaoviana isn't this the wrong green ? Screenshot 2023-11-03 at 17 40 36

should be darker? Need to find out the correct green.{some number}

joaoviana:
> > @joaoviana isn't this the wrong green ? Screenshot 2023-11-03 at 17 40 36
> 
> should be darker? Need to find out the correct green.{some number}

changed to this: `green.7`

image

IrakliJani:
@joaoviana should we just change it to blue? I don't know if we are using green buttons elsewhere

joaoviana:
> @joaoviana should we just change it to blue? I don't know if we are using green buttons elsewhere

mantine buttons? No, I don't think so. Cool, can change it!

IrakliJani (approved):
looks great!

Issue #7689, opened 2023-10-27
When changing from date filter from month to year, the value updates on the UI but not on the generated SQL


joaoviana:
### To Reproduce

* Create chart with filter `Month` and set to `is` non-january month - e.g. `Feb 2023`
* Modify just the field from `Month` to a year - UI updates the value from `Feb 2023` to just `2023`
* However, the generated SQL does not - User has to manually click on the `2023` input so that the query runs as expected

Sai-Karthik-Tumu:
Hey @joaoviana , can I work on this issue

Sai-Karthik-Tumu:
Hey @joaoviana ,
I know that you haven't approved my request. As I already have the fix ready thought it would help you save some time. So raised the PR.

Hope this helps you.

Thanks

jpetey75:
this doesn't happen anymore

PR #7817, opened 2023-11-04
fix: When changing from date filter from month to year, the value updates on the UI but not on the generated SQL #7689

Sai-Karthik-Tumu:
Closes: https://github.com/lightdash/lightdash/issues/7689

### Description:

1. When the **FilterYearPicker** or **FilterMonthAndYearPicker** components are rendered for the first time, the update on the Filter rule is not done because it is handled in the onChange function. 
2. If the onChange function is not triggered there is no way that Filter rule has the updated date value. 
3. So I added this logic before the component is rendered and component will have the updated Filter Rule even when the component is rendered for the first time

https://github.com/lightdash/lightdash/assets/86364607/20135228-97b5-4bbc-967c-fd18aad06b32

### Reviewer actions


owlas:
Hey @Sai-Karthik-Tumu we were just talking about this and we've decided this is a fix we need to make on the backend. Well re-word the issue.

Issue #7753, opened 2023-10-31
The ordering of my columns in my csv + Google Sheets export should match the ordering of the columns in my results table


TuringLovesDeathMetal:
### Description

The ordering of the columns in my results table gets swapped around in my google sheets or CSV export (It applies the ordering of the very initial results table, before I move any columns around). If I move any columns around, they are not exported in the same order in the Google Sheet/csv

Some users are reporting that this is only affectig table calculations, and this is a recent regression (table calculations are on the far right in the table, but when exported appear on the far left in the .csv)

https://www.loom.com/share/dc94e8cb1a054a6eb52abf9032d3f864

Shivanshu-09:
Is this still an issue? I reproduced it but it seems to be working. @TuringLovesDeathMetal

Issue #7702, opened 2023-10-27
Hidden collumns on tables should not appear on CSV export on dashboards


rephus:
This works when donwloading a csv from a chart

However, this is not working when downloading a CSV from a dashboard (and by extension, not from scheduler) 

See video: 

[Screencast from 27-10-23 15:48:22.webm](https://github.com/lightdash/lightdash/assets/1983672/48dd9ae4-2d54-4e78-8f55-a3c9e66ab4a0)

_Originally posted by @rephus in https://github.com/lightdash/lightdash/issues/7674#issuecomment-1782958938_

rephus:
We can't do the filtering by removing dimensions/metrics on metricquery on the frontend when calling /downloadCSV. Becuase a table calculation, or ordering might need it. 

INstead, we should refactor the code to send the `tableconfig` info to /downloadCSv,  generate the results, and then remove the `not visible` columns

PR #7842, opened 2023-11-07
fix: download csv column order

rephus:
Closes: https://github.com/lightdash/lightdash/issues/7753

### Description:

Screenshot from 2023-11-07 15-34-47
Screenshot from 2023-11-07 15-24-21

### Reviewer actions


packages/frontend/src/components/Explorer/VisualizationCard/VisualizationCard.tsx
```diff
@@ -108,52 +109,18 @@ const VisualizationCard: FC<{
...
-            let filteredColumnOrder: string[] | undefined;
-
-            if (
-                unsavedChartVersion.chartConfig.type === 'table' &&
-                unsavedChartVersion.chartConfig.config?.columns
-            ) {
-                const columnsConfig =
-                    unsavedChartVersion.chartConfig.config?.columns;
-                filteredColumns =
-                    unsavedChartVersion?.metricQuery.dimensions.filter(
```
rephus, line 121: the old logic was wrong, it was not considering `visible` metrics

packages/frontend/src/utils/csvUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+        const filterFieldNotVisible = (fieldId: string) =>
+            !filteredColumnIds?.includes(fieldId);
+
+        return {
+            metricQuery: {
+                ...metricQuery,
+                metrics: metricQuery.metrics.filter(filterFieldNotVisible),
+                dimensions: metricQuery.dimensions.filter(
+                    filterFieldNotVisible,
+                ),
```
ZeRego, line 29: I don't think we should manipulate the metricQuery values.
What happens if a table calculation depends on the dimension/metric?

rephus: yeah, great point, instead we should pass the table config and hide the columns on the csv generation, right ? 


I can simply remove the filtering to fix https://github.com/lightdash/lightdash/issues/7753 but not https://github.com/lightdash/lightdash/issues/7702

rephus:
Reverted "filter" code , removed old filtering to fix sorting

image

Fixing visibility will be done in a separate PR

ZeRego (approved):

Issue #7480, opened 2023-10-14
Ability to filter on Table Calculation


gordonkjlee:
### Description
I would like to be able to filter on Table Calculations.

Snowflake allows this to be achieved using the QUALIFY statement, although for compatibility across databases this should probably be implemented by wrapping the SQL and then applying a WHERE clause after.

### What problem does this solve?
More complex analysis that requires filtering on table calculations.

Issue #1270, opened 2022-02-02
Filters general: Add filters for table calculations


TuringLovesDeathMetal:
### Description: What is it?

You should be able to filter your results table based on table calculations

### Problem: What problem does this solve?

You can't filter results based on a table calculation. For example, if I had a table calculation for % week on week change and I wanted to know which weeks had changes > 10%, I wouldn't be able to do this right now.

### Success: How will we know if we’ve solved this problem?

- you're able to filter on table calculations in explores
- You're able to filter on table calculations in dashboards.

### Audience: Who are we building this for?

Data builders/people building the charts

### What: Roughly, what does this look like in the product?

Exactly the same as the metrics + dimensions filters, just for table calculations

pelletstove:
This would be a useful feature for filtering in topN after performing a rank().

PR #7846, opened 2023-11-07
feat: Add filters for table calculations

Shifan-Gu:
Closes: [7480](https://github.com/lightdash/lightdash/issues/7480) and [1270](https://github.com/lightdash/lightdash/issues/1270)

### Description:
allow add filter based on table calculations.

jaffle shop example:

step 1: create a ranking table calculation using total order amount
ranking_table_calculation

step 2:
select_table_calculation

step 3:
create filter to find top 5 most expensive order
top_5_order

### Reviewer actions


TuringLovesDeathMetal:
one thing worth checking here for whoever is reviewing: what happens when the table calculation returns a string or boolean data type

TuringLovesDeathMetal:
Also, @Shifan-Gu the engineering team is at an offsite this week, so we won't review it this week, but we'll take a look at it once the engineers are back! 👀

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -551,6 +559,9 @@ export class DashboardModel {
                 dashboard.dashboard_version_id,
             );
 
+        const tableCalculationFilters = view?.filters?.tableCalculations;
+        view.filters.tableCalculations = tableCalculationFilters || [];
```
rephus, line 563: Is this necessary ? It feels a bit redundant. I don't see it happening for other filters like `dimensions` or `metrics`

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -299,6 +301,24 @@ const renderBooleanFilterSql = (
...
+export const renderTableCalculationFilterRuleSql = (
+    filterRule: FilterRule<FilterOperator, unknown>,
+    field: CompiledTableCalculation | undefined,
+): string => {
+    if (!field) return '1=1';
+    switch (field.format?.type) {
+        case TableCalculationFormatType.DEFAULT:
+        case TableCalculationFormatType.PERCENT:
+        case TableCalculationFormatType.CURRENCY:
+        case TableCalculationFormatType.NUMBER: {
```
rephus, line 313: does that mean we can only filter numeric results from table calculations ?

packages/backend/src/queryBuilder.ts
```diff
@@ -720,9 +735,12 @@ export const buildQuery = ({
             ? `WHERE ${whereMetricFilters}`
             : '';
         const secondQuery = [finalSelect, finalFrom, finalSqlWhere].join('\n');
+        const finalQuery = tableCalculationFilters
+            ? `SELECT * FROM (${secondQuery}) query_result WHERE ${tableCalculationFilters}`
```
rephus, line 739: note to self: test this

packages/backend/src/queryBuilder.ts
```diff
@@ -604,6 +605,14 @@ export const buildQuery = ({
     const sqlOrderBy =
         fieldOrders.length > 0 ? `ORDER BY ${fieldOrders.join(', ')}` : '';
     const sqlFilterRule = (filter: FilterRule, fieldType: FieldType) => {
+        if (fieldType === FieldType.TABLE_CALCULATION) {
+            const field = compiledMetricQuery.compiledTableCalculations?.find(
+                (tc) =>
+                    `table_calculation_${tc.name}` === filter.target.fieldId,
```
rephus, line 611: note to self: test this

packages/common/src/types/field.ts
```diff
@@ -123,6 +123,15 @@ export type TableCalculation = {
     format?: TableCalculationFormat;
 };
 
+export interface TableCalculationFilter extends Field {
```
rephus, line 126: I don't think TableCalculationFilter should to extend Field, they are different types, a filter should not be a field.

rephus: WHy do we need this type ? FilterableItem already supports TableCalculation

packages/common/src/utils/filters.ts
```diff
@@ -67,8 +70,8 @@ export const hasNestedGroups = (filters: Filters): boolean => {
         return items.some(isFilterGroup);
     };
     return (
-        (!!filters.dimensions && hasGroups(filters.dimensions)) ||
-        (!!filters.metrics && hasGroups(filters.metrics))
+        (!!filters.metrics && hasGroups(filters.metrics)) ||
```
rephus, line 73: why have we removed the `(!!filters.dimensions && hasGroups(filters.dimensions))` result ?

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
```diff
@@ -34,6 +34,14 @@ const useDashboardFiltersForExplore = (
                 ...dashboardFilters.metrics,
                 ...(dashboardTemporaryFilters?.metrics ?? []),
             ]),
+            tableCalculations: getDashboardFilterRulesForTile(
```
rephus, line 37: This method does not exist, perhaps you mean `getDashboardFilterRulesForTileAndTables` ?

rephus:
2 compilation errors: 

```
18.23 src/hooks/dashboard/useDashboardFiltersForTile.ts(14,5): error TS2741: Property 'tableCalculations' is missing in type '{ dimensions: DashboardFilterRule<ConditionalOperator, DashboardFieldTarget, any, any>[]; metrics: DashboardFilterRule<ConditionalOperator, DashboardFieldTarget, any, any>[]; }' but required in type 'DashboardFilters'.
18.23 src/utils/dateFilter.ts(52,5): error TS2741: Property 'tableCalculations' is missing in type '{ dimensions: DashboardFilterRule<ConditionalOperator, DashboardFieldTarget, any, any>[]; metrics: DashboardFilterRule<ConditionalOperator, DashboardFieldTarget, any, any>[]; }' but required in type 'DashboardFilters'.
```

rephus:
There are some APi e2e errors

>     1) Should get success response (200) from POST chart results with filters

rephus:
Using a more complex TC expression (just adding brackets) does not work 

Screenshot from 2023-11-20 10-39-15

Screenshot from 2023-11-20 10-39-08

Screenshot from 2023-11-20 10-38-59

rephus:
Returning a string (concat) throws an error 
image

packages/backend/src/models/DashboardModel/DashboardModel.ts
Shifan-Gu: Hi Rephus, reason doing so is because dashbaord view is stored as json in the database, so i'm not sure what's the recommended way to migrate the json object to include default table calculation filters list, so instead i set it to empty list here if its not exist in the database.

packages/common/src/types/field.ts
Shifan-Gu: it conflicts some where in the code i remember if i only use FilterableItem in the code, let me figure this out and come back to you.

packages/common/src/utils/filters.ts
Shifan-Gu: good catch, that's a mistake when resolving the conflict.

packages/frontend/src/hooks/dashboard/useDashboardFiltersForExplore.ts
Shifan-Gu: same as above, i will correct this

Shifan-Gu:
Hi @rephus, 

thanks for the feedback! i will work on the comments.

rephus:
Also, some minor things (these can be addressed in separate issues) 

- We're missing the `filter by` options we have in results table 
Screenshot from 2023-11-20 10-37-08
Screenshot from 2023-11-20 10-37-12

- these filters don't work on dashboards 

image

rephus:
> Using a more complex TC expression (just adding brackets) does not work . It is silently not filtering anything.

Could we throw an exception (eg: not supported) every time a Table calculation filter is not supported ? rather than silently failing with a `1==1` condition .

Shifan-Gu:
Hi @rephus 

I was changing DEFAULT type table calculation filter to be used as string filter, but i noticed that for dimension and metrics they use <FilterStringAutoComplete/> which fetches the data from backend to pre-populate the select dropdown list.

Table calculation column is derived column so it can't have the same behaviour. i was thinking using the TagsInput instead of MultiSelect input but it requires the Maintine version 7.2.2.

Do you guys have any plan to upgrade Maintine to the latest version or do you have any suggestion on what component i can use for this filter?

Shifan-Gu:
> Returning a string (concat) throws an error image

Checked this issue and i think its an valid error, because the table calculation result is cast to string but then its trying to filter as numeric value. 

You will see the same type of error when you creating the table calculation with string return type but set the format to number

Shifan-Gu:
> Using a more complex TC expression (just adding brackets) does not work . It is silently not filtering anything.
> 
> Screenshot from 2023-11-20 10-39-15
> 
> Screenshot from 2023-11-20 10-39-08
> 
> Screenshot from 2023-11-20 10-38-59

I checked this issue as well, its because when you edit the table calculation in the modal if you don't give it a new name, it will call a method to make it unique, so it's an expected behaviour.

So if you delete that filter and create it again it should work

Shifan-Gu:
Hi team, new code patch is ready for review

updates: 
1. resolved conflicts and fixed build issues
2. fixed broken e2e api tests
3. changed table calculation with Default format to use string filter component
4. throws the exception when a table calculation format type is not supported to render a sql

rephus:
Something weird is happening on the filter, the "field" thinks my table calculation is a number (see icon) but the filters I get are for `strings` (I should get `greater than`, etc) 

image

rephus:
Every time I update the table calculation, I loose the filter, it seems the table calculation id is changing. This can be fixed in a separate PR , https://github.com/lightdash/lightdash/issues/7996



[Screencast from 22-11-23 09:17:55.webm](https://github.com/lightdash/lightdash/assets/1983672/c2e4799c-33c1-4f4d-921d-47691663e246)

rephus:
Edge case, if I try to filter on a `date` field, I get this error, this can be fixed in a separate PR 

image

rephus (approved):
The filter functionality works as expected, great job 

There are a few minor issues with types on table calculations but they can be fixed separately.

PR #7865, opened 2023-11-08
fix: remove duplicate requests on page load

ZeRego:
### Description:

This change will make sure there are no duplicate requests within 30 seconds. This will make all pages faster to load.
Hooks that use `onSuccess` need to have staleTime: 0 

Before:
Notice duplicate requests for /health /org /user /projects
Screenshot 2023-11-08 at 12 45 01

After:
Screenshot 2023-11-08 at 12 42 40

### Reviewer actions


joaoviana (approved):
🎉

Issue #7875, opened 2023-11-08
Cannot add second reference line


12ian34:
Trying to add a second reference line (whether on the same series or not) is not working.
This was initially reported by a user in the community:

> Hi all regarding References Lines.
> I am trying to display 2 reference lines but it seems to be failing and only showing a single reference line can someone help me out?
> This is through the chart development not through the sql runner.
Link: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1699451234334669?thread_ts=1699451234.334669&cid=C03MG2VTR08

I can reproduce this error on our instance - and in my case, adding a second reference line is not working.

An additional issue is that when typing in the value for the reference line, the text input box defocuses after typing the first character! 

Screen recording: 
https://share.cleanshot.com/kGN7wWCC

This was on the latest version as of 2023-11-08

PR #7884, opened 2023-11-09
fix: multiple reference lines

rephus:
Closes:  #7875

### Description:

What happened: I recently moved from using `name` to `value` to set a label instead of an uuid (but kept the name for backwards compatibility) , however, the update method was not updated to use the new id field, this was causing that all the lines were updated at the same time because it was sending name (reference line).

Screenshot from 2023-11-09 10-04-09

### Reviewer actions


magnew (approved):
👍

Issue #7835, opened 2023-11-07
The filter lists shouldn't reset to the top when I select a value from the list


TuringLovesDeathMetal:
### Description

Right now, the filter list resets to the top each time I select a value. Ideally, it should stay in the same "place" as I'm currently in when:

1. I've scrolled to somewhere in the list
2. I've typed some letters to filter the list

Here's an example of both of those options and how they're currently resetting the filter list position

https://www.loom.com/share/c07ee1203855400895850d6379030bb9

iMac7:
Please assign me this. I have linked a PR for it above with comments

ZeRego:
Something to consider before applying a fix. 
Comment from #7917 
> If we test this problem with the vanilla Mantine component we can see that it doesn't scroll when picking a value.
This makes me wonder if our code is not efficient and is rerendering when it shouldn't causing the scroll to reset.
Would be good to investigate and confirm what is causing this behaviour before pushing a fix where we control the scroll.

PR #7917, opened 2023-11-12
feat: scroll restoration on filter lists

iMac7:
Closes: #7835 

### Description:
This restores the scroll position after selecting an item from the filter dropdown for requirement 1. UseLayouteffect is preferred to useEffect, because useeffect has sudden noticeable shift when scrolling back after the component re-renders.

The second requirement  is unclear to me, and is not covered in this PR yet. 
When searching using a search term, the results change, which also changes the size of a list. I am not sure if keeping the scroll for different size lists is the intended behavior. Also when using a search term, the letters are cleared when clicking on a result in the list, and the list resets. 

I'd be happy to continue work on this provided the information, or otherwise  take over the PR and make changes :)

[Screencast from 12-11-23 10:15:05.webm](https://github.com/lightdash/lightdash/assets/76876702/602515ef-173e-4da0-9625-48ec6e7707e6)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.tsx
@@ -9,7 +9,15 @@ import {
 } from '@mantine/core';
 import { IconPlus } from '@tabler/icons-react';
 import uniq from 'lodash-es/uniq';
-import { FC, ReactNode, useCallback, useMemo, useState } from 'react';
+import {
+    FC,
+    ReactNode,
+    useCallback,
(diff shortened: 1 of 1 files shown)


ZeRego:
If we test this problem with the vanilla Mantine component we can see that it doesn't scroll when picking a value. 
This makes me wonder if our code is not efficient and is rerendering when it shouldn't causing the scroll to reset. 
Would be good to investigate and confirm what is causing this behaviour before pushing a fix where we control the scroll.

ZeRego:
I'm going to close this PR based on my last comment to keep our PR list clean.
Feel free to open a new one if you find the root cause of the issue.

PR #7928, opened 2023-11-13
fix: dashboard filters performance (Part 1)

IrakliJani:
### Description:

I was analyzing performance related to 'filter pills,' and it appears that we were recalculating all the functions I memoized in the PR after every minor action, such as drag-and-drop, opening menus, and hovering over buttons. This is likely due to the heavy Dashboard provider state and not 'slicing' useDashboardContext anywhere.

Not memoizing the variables in the filter pill component leads to significant rerendering of child components. This happens because all recalculated variables are references (not value types), and React only performs shallow diffing.


magnew (approved):
Looks like a good idea and pretty safe.

PR #7933, opened 2023-11-13
fix: reuse fields with suggestions

joaoviana:
Closes: N/A

### Description:

Reuse `fieldsWithSuggestions` from `DashboardProvider` which is basically the same as what was being done with `fieldMap`

```tsx
const fieldMap = allFilterableFields?.reduce<
        Record<FieldId, FilterableField>
    >((acc, field) => ({ ...acc, [fieldId(field)]: field }), {});
```

In `DashboardProvider` we do similar already: https://github.com/lightdash/lightdash/blob/fix/reuse-fields-with-suggestions/packages/frontend/src/providers/DashboardProvider.tsx#L308

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
@@ -1,4 +1,3 @@
-import { FieldId, fieldId, FilterableField } from '@lightdash/common';
 import { Group, Skeleton } from '@mantine/core';
 import { FC } from 'react';
 import { useDashboardContext } from '../../../providers/DashboardProvider';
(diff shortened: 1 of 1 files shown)


IrakliJani (approved):
yeah I think this should work just fine!

its set using useEffect, which I don't like a little bit but we can optimize that later. underneath it's the same `allFilterableFields` `reduce` function

```
    useEffect(() => {
        if (allFilterableFields && allFilterableFields.length > 0) {
            setFieldsWithSuggestions((prev) => {
                return allFilterableFields.reduce<FieldsWithSuggestions>(
                    (sum, field) => ({
                        ...sum,
                        [fieldId(field)]: {
                            ...field,
                            suggestions:
                                prev[fieldId(field)]?.suggestions || [],
                        },
                    }),
                    {},
                );
            });
        }
    }, [allFilterableFields]);
```

PR #7942, opened 2023-11-14
fix: improve available filters and metric query suggestions

joaoviana:
Closes: #7946

### Description:

**Improving `/availableFilters` payload by reducing its size:**
Currently, we store **all** available filterable fields per tile in a massive object, like so: 
```
{
[chartUuid]: [{...field1}, {...field2}]
}
```
This can be extremely big if the table the chart belongs to contains many fields + joined tables

With this change, we are: 
* caching explores in memory - reducing the call of `getExplore` if charts in a dashboard share the same `tableName`
* getting all filterable fields
* and an indexed map per tile
* Sending extra info to the body of the request, by sending an array of objects that contains both the tileUuid and the savedChartUuid
like so (from render)

image

**Note^**

Now the Uuids are the tile Uuids and not the saved query Uuids - these don't need to be computed on the FE anymore

This payload then gets computed on the FE - where each tile has these fields populated by the index mapped between the tile and the filterable fields map. 
Besides this, I've removed the `allFilterableFields` FE computation and reused the map sent already by the BE `/availableFilters`; also removed `filterableFieldsBySavedQueryUuid` from the provider because it's not being used

>  [!IMPORTANT]
> We've removed the `suggestions` logic for each chart, so the above comments on **Suggestions per metricQuery** are no longer relevant. Instead, we are not getting any pre-fetched suggestions and we let the filter components do their thing. This is so we avoid looping through all rows of a chart on the FE

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -60,6 +60,7 @@ import {
     ProjectType,
     RequestMethod,
     ResultRow,
+    SavedChartsInfoForDashboardAvailableFilters,
     SessionUser,
     SpaceQuery,
     SpaceSummary,
@@ -2113,7 +2114,7 @@ export class ProjectService {
 
     async getAvailableFiltersForSavedQueries(
         user: SessionUser,
-        savedQueryUuids: string[],
+        savedChartUuidsAndTileUuids: SavedChartsInfoForDashboardAvailableFilters,
     ): Promise<DashboardAvailableFilters> {
         const transaction = Sentry.getCurrentHub()
             ?.getScope()
@@ -2127,6 +2128,10 @@ export class ProjectService {
             uuid: string;
             filters: CompiledDimension[];
         }[] = [];
+
+        const savedQueryUuids = savedChartUuidsAndTileUuids.map(
+            ({ savedChartUuid }) => savedChartUuid,
+        );
 
         try {
             const savedCharts =
@@ -2206,11 +2211,12 @@ export class ProjectService {
             });
         });
 
-        const savedQueryFilters = savedQueryUuids.reduce<
+        const savedQueryFilters = savedChartUuidsAndTileUuids.reduce<
             DashboardAvailableFilters['savedQueryFilters']
-        >((acc, savedQueryUuid) => {
+        >((acc, savedChartUuidAndTileUuid) => {
             const filterResult = allFilters.find(
-                (result) => result.uuid === savedQueryUuid,
+                (result) =>
+                    result.uuid === savedChartUuidAndTileUuid.savedChartUuid,
             );
             if (!filterResult || !filterResult.filters.length) return acc;
 
@@ -2219,7 +2225,7 @@ export class ProjectService {
             );
             return {
                 ...acc,
-                [savedQueryUuid]: filterIndexes,
+                [savedChartUuidAndTileUuid.tileUuid]: filterIndexes,
             };
         }, {});
 
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -938,6 +938,21 @@ export class ProjectService {
             explore,
         });
 
+        const suggestionsToAddToFields = dashboardFilters
+            ? metricQuery.dimensions.reduce((sum, dimensionId) => {
+                  //! this is bad, but necessary for UX - optimise so it's not extremely slow when there are many rows
+                  const newSuggestions: string[] =
+                      rows.reduce<string[]>((acc, row) => {
+                          const value = row[dimensionId]?.value.raw;
+                          if (typeof value === 'string') {
+                              return [...acc, value];
+                          }
+                          return acc;
+                      }, []) || [];
+                  return { ...sum, [dimensionId]: newSuggestions };
+              }, {})
+            : {};
+
         return {
             chart: savedChart,
             explore,
@@ -945,6 +960,7 @@ export class ProjectService {
             cacheMetadata,
             rows,
             appliedDashboardFilters,
+            suggestionsToAddToFields,
         };
     }
 
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2138,27 +2138,23 @@ export class ProjectService {
                 await this.savedChartModel.getInfoForAvailableFilters(
                     savedQueryUuids,
                 );
+            const exploreCacheKeys: Record<string, boolean> = {};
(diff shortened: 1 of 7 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -938,13 +938,29 @@ export class ProjectService {
...
         });
 
+        const suggestionsToAddToFields = dashboardFilters
+            ? metricQuery.dimensions.reduce((sum, dimensionId) => {
+                  //! this is bad, but necessary for UX - optimise so it's not extremely slow when there are many rows
+                  const newSuggestions: string[] =
+                      rows.reduce<string[]>((acc, row) => {
+                          const value = row[dimensionId]?.value.raw;
+                          if (typeof value === 'string') {
+                              return [...acc, value];
```
magnew, line 948: I'm not sure how big these get, but if we are concerned about perf I think concat will be a little faster than the spread.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2133,6 +2149,33 @@ export class ProjectService {
...
+                const key = `${chart.projectUuid}_${chart.tableName}`;
+                if (!uniqueExplores.has(key)) {
+                    uniqueExplores.add(key);
+                }
+            });
+
+            const explorePromises = Array.from(uniqueExplores).map(
+                async (key) => {
+                    const [projectUuid, tableName] = key.split('_');
+                    const explore = await this.getExplore(
```
magnew, line 2165: This await means you are running them sequentially, right? I think this will block before sending the next one. But since it returns promises, you should be able to run them in parallel, then await all of them on line 2174 with Promise.all. No?

I haven't tested this, so do let me know if I am missing something -- I could be wrong.

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -420,17 +424,29 @@ export const DashboardProvider: React.FC = ({ children }) => {
...
+
+                        const currentSuggestions = field?.suggestions ?? [];
+
+                        const newSuggestions =
+                            newSuggestionsMap[key] || new Set();
                         const suggestions = Array.from(
-                            new Set([...currentSuggestions, ...newSuggestions]),
+                            new Set([
+                                ...currentSuggestions,
+                                ...Array.from(newSuggestions),
```
magnew, line 439: How many of these do we think there are? Like how many things in this loop? If its just a few this might be ok, but if there are thousands I think this is going to be slow. We create a new set on line 435, then allocate memory for a new array on 436, then a new set on 437, then a new array on 439. I think we could tighten this up an have fewer allocations. I'm happy to look closer if you want. 

I'm not trying to be too nit-picky! We are doing perf work or I wouldn't notice so much 😄

magnew (commented):
A couple questions/suggestions. I'm happy to go over this took.

packages/frontend/src/providers/DashboardProvider.tsx
joaoviana: Removing suggestions https://github.com/lightdash/lightdash/pull/7942/commits/ae13563c187da046e8f97a56ff124f25315a2c5f ! Testing this PR + testing on `main` and will chat with @IrakliJani on the consequences of this
But will address this spreading and other operations that would be heavy if we consider re-adding these `fieldsWithSuggestions`, thanks 👍

packages/backend/src/services/ProjectService/ProjectService.ts
joaoviana: You mean the `explorePromises` or this + the other `filterPromises`?

packages/backend/src/services/ProjectService/ProjectService.ts
joaoviana: 💯 , I've removed this code to see if we _actually_ need it. But your suggestion makes a lot of sense 👍

packages/backend/src/services/ProjectService/ProjectService.ts
magnew: I mean for explorePromises, you are awaiting each one, here, then you are awaiting all of them on line 2174, no? I don't think you need this await and I think it is causing them to run in series instead of in parallel.

joaoviana: Ah I see your concern, makes sense! However, since the map method doesn't wait for the promises to resolve, all calls to getExplore are initiated almost simultaneously, those are then awaited in the `promise.all`

magnew: Interesting. Not sure I follow what's going with needing the await, but that's fine with me. I do see what you are saying about map not waiting, so my concern is probably not valid.

owlas: @joaoviana I see what you mean that the promise is actually the one from the anonymous function so all the anonymous functions fire in parallel.

but I think inlining this makes the behaviour a bit confusing. could you pull out the anonymous function like:

```typescript
const f = async (key) => {... await getExplore(..) ... }
explorePromises = Array.from(uniqueExplores).map(f)
```

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2133,6 +2133,33 @@ export class ProjectService {
                 await this.savedChartModel.getInfoForAvailableFilters(
                     savedQueryUuids,
                 );
+            const exploreCache: Record<string, Explore> = {};
+
+            const uniqueExplores = new Set<string>();
+            savedCharts.forEach((chart) => {
+                const key = `${chart.projectUuid}_${chart.tableName}`;
```
owlas, line 2140: You don't need the project uuid for this key to be unique

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2133,6 +2133,33 @@ export class ProjectService {
                 await this.savedChartModel.getInfoForAvailableFilters(
                     savedQueryUuids,
                 );
+            const exploreCache: Record<string, Explore> = {};
+
+            const uniqueExplores = new Set<string>();
+            savedCharts.forEach((chart) => {
```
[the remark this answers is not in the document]
joaoviana: But if these are also run in parallel, then we might not hit the cache ever... no? 🤔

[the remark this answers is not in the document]
joaoviana: ignore the above ^ was looking at the wrong thing

joaoviana:
only e2e fail on `createProjects`

owlas (approved):
- Works great in pr preview ✅ 
- Code looks good ✅ 

🚢

PR #7952, opened 2023-11-14
fix: upgrade zlib and merge dependencies

rephus:
Closes: 

### Reviewer actions

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -24,7 +24,7 @@
         "copyfiles": "^2.4.1",
         "jest-fetch-mock": "^3.0.3",
         "knex-mock-client": "^1.11.0",
-        "tsoa": "^4.1.3"
+        "@tsoa/cli": "^4.1.3"
     },
     "dependencies": {
         "@aws-sdk/client-s3": "^3.272.0",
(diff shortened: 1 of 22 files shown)


packages/backend/package.json
```diff
@@ -55,7 +56,7 @@
         "@types/sshpk": "^1.17.1",
         "ajv": "^8.3.0",
         "ajv-formats": "^2.1.0",
-        "bcrypt": "^5.0.1",
+        "bcrypt": "^5.1.1",
```
rephus, line 59: bcrypt is depending on minizlib which depends on zlib

packages/backend/package.json
```diff
@@ -101,7 +102,7 @@
         "slackify-markdown": "^4.4.0",
         "sshpk": "^1.17.0",
         "tempy": "^1.0.1",
-        "tsoa": "^4.1.3",
+        "@tsoa/runtime": "^4.1.3",
```
rephus, line 105: moevd tsoa/cli to devdependencies, which is using `merge` and added here tsoa/runtime 
image

packages/backend/package.json
rephus: I'm not 100% sure this will fix the issue, but 5.1.0 did upgrade its depedencies

owlas (approved):

Issue #6315, opened 2023-07-13
You should be able to apply temporary sorting to table charts in dashboards


PriPatel:
Miro sketch: https://miro.com/app/board/uXjVNeDsQHw=/?share_link_id=24821374308

### Acceptance critieria: 

- [ ] You should be able to apply temporary sorting to columns in table charts in dashboards. This is for exploration whilst in view-mode.
- [ ] This only applies to table chart types
- [ ] The sorting isn't saved once the user refreshes the page, or exits the dashboard (just like dashboard filters in view mode)
- [ ] By default the table is sorted according to the sort of the underlying table. 
- [ ] Clicking any of the headers then completely overrides the sort configuration on the underlying table and the UI should indicate the sort order. 
- [ ] Clicking on the same column header:
  - [ ] once = sorting descending
  - [ ] twice = sorting ascending
  - [ ] three times = sorting turned off 
- [ ] If I've added a sort on one column, then I click on another column, the sort from the first column is removed and only the second column sort is applied. 
- [ ] This sorting behaviour works in `view mode` of a chart, in `view mode` on a dashboard, and in the table in `view underlying data`
- [ ] Users can only apply a single sort 
- [ ] It should apply sorting in the underlying sql query

This allows the user to override the `sort` configuration on a `MetricQuery` like we have for dashboard filters. It should apply sorting in the underlying sql query

owlas:
what about multisort?

owlas:
For now let's only support a single sort type. 

By default the table is sorted according to the sort of the underlying table. Clicking any of the headers then completely overrides the sort configuration on the underlying table and the UI should indicate the sort order. The user can sort a single column at a time, no multi-sort.

hamzahc1:
This was requested by a potential customer today.

hamzahc1:
This should also apply to View underlying values

mhawkins-runa:
Is this essentially the same issue as [this](https://github.com/lightdash/lightdash/issues/3591)? I was searching for this issue as I'm interested in the same feature, and came across both tickets.

hamzahc1:
> Is this essentially the same issue as [this](https://github.com/lightdash/lightdash/issues/3591)? I was searching for this issue as I'm interested in the same feature, and came across both tickets.

Good catch! I've closed the other issue in favour of this one

rephus:
Let's also show the default sorting applied at the chart level (even without adding temporary sorting)

[...]

PR #7957, opened 2023-11-14
feat: sort dashboard table

rephus:
Closes: #6315

### Description:

[Screencast from 15-11-23 10:06:04.webm](https://github.com/lightdash/lightdash/assets/1983672/ed9627c4-329f-4adf-955c-4d0db73020b6)

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/savedChartController.ts
+++ b/packages/backend/src/controllers/savedChartController.ts
@@ -4,6 +4,7 @@ import {
     ApiGetChartVersionResponse,
     ApiSuccessEmpty,
     DashboardFilters,
+    SortField,
 } from '@lightdash/common';
 import { Body, Get, Post } from '@tsoa/runtime';
 import express from 'express';
@@ -47,6 +48,7 @@ export class SavedChartController extends Controller {
         body: {
             dashboardFilters?: any; // DashboardFilters; temp disable validation
             invalidateCache?: boolean;
+            dashboardSorts?: SortField[];
         },
         @Path() chartUuid: string,
         @Request() req: express.Request,
@@ -60,6 +62,7 @@ export class SavedChartController extends Controller {
                 dashboardFilters: body.dashboardFilters,
                 versionUuid: undefined,
                 invalidateCache: body.invalidateCache,
+                dashboardSorts: body.dashboardSorts,
             }),
         };
     }
--- a/packages/frontend/src/components/SimpleTable/DashboardHeaderContextMenu.tsx
+++ b/packages/frontend/src/components/SimpleTable/DashboardHeaderContextMenu.tsx
@@ -6,8 +6,8 @@ import {
 } from '@lightdash/common';
 import { ActionIcon, Flex, Menu } from '@mantine/core';
 import {
-    IconCaretDownFilled,
-    IconCaretUpFilled,
+    IconArrowDown,
+    IconArrowUp,
     IconCheck,
     IconChevronDown,
 } from '@tabler/icons-react';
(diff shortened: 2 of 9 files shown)


packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -115,6 +118,8 @@ export const DashboardProvider: React.FC = ({ children }) => {
 
     const [invalidateCache, setInvalidateCache] = useState<boolean>(false);
 
+    const [chartSort, setChartSort] = useState<Record<string, SortField[]>>({});
```
owlas, line 121: Why do we store the state on the dashboard provider instead of something at the tile level like `useDashboardChart`? @IrakliJani or @joaoviana could confirm, but if it's here it could trigger re-render of all charts when sorting changes (since all charts depend on the chart sort object).

owlas (commented):
Backend is elegant and simple! Just a suggestion for naming.

Frontend might be a risky change. I'm asking @IrakliJani or @joaoviana to make a recommendation on storing the sort state. Since it's only local to the chart it's really important that it only makes that single tile re-render and not the whole dashboard.

packages/frontend/src/providers/DashboardProvider.tsx
joaoviana: once this is rebased, I can have a look on Render environment and check with the Performance profiler, sounds good @owlas  👍

rephus: we should also be able to `share` this later https://github.com/lightdash/lightdash/issues/7948 , in case this is a blocker for moving the logic out of the dashboardProvider

rephus: This logic seems to be affecting other charts in the dashboard, at least they are not re-rendered. But let's confirm this  
[Screencast from 15-11-23 12:13:10.webm](https://github.com/lightdash/lightdash/assets/1983672/e026eb06-a6f5-47f2-ae21-828884bb3ccf)

packages/frontend/src/components/SimpleTable/DashboardHeaderContextMenu.tsx
```diff
@@ -0,0 +1,148 @@
...
+} from '@tabler/icons-react';
+import { FC, useMemo } from 'react';
+import MantineIcon from '../common/MantineIcon';
+import { HeaderProps, TableColumn } from '../common/Table/types';
+
+import { useDashboardContext } from '../../providers/DashboardProvider';
+import { getSortDirectionOrder, getSortLabel } from '../../utils/sortUtils';
+import { BolderLabel } from '../Explorer/ResultsCard/ColumnHeaderContextMenu.styles';
+
+export enum SortDirection {
```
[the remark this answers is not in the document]
rephus: This enum is used in the method `getSortDirectionOrder` to generate a list of orders (and labels) I'm  reusing some of the logic that we use on `results` table. In the end, this is converted into a `boolean` in SortField .

packages/backend/src/controllers/savedChartController.ts
```diff
@@ -47,6 +48,7 @@ export class SavedChartController extends Controller {
         body: {
             dashboardFilters?: any; // DashboardFilters; temp disable validation
             invalidateCache?: boolean;
+            dashboardSorts?: SortField[];
```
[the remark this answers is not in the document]
rephus: in the context of this `API request` these filters and sorts are applied to a single chart. 

I didn't want to say `override` , and be more generic, in case we change the logic and decide to `append` like we do in filters, but that is just a personal preference,  I can change if you prefer.

packages/frontend/src/providers/DashboardProvider.tsx
joaoviana: I can confirm that the paint flashes only target the table you're sorting, so it looks good! 💪 

https://github.com/lightdash/lightdash/assets/7611706/2c3d99b4-7c60-4ca3-a863-47732745d892

TuringLovesDeathMetal (approved):
Nice! Looks good to me :) 

Like we talked about at standup - let's not deal with pivot tables for now!

owlas (approved):

Issue #7930, opened 2023-11-13
Dashboard filter modals don't auto-hide when clicking away IF you focus the filter value input box


12ian34:
### Description

Dashboard filter modals don't auto-hide when clicking away IF you focus the text box

This leads to views like this - if you focus a filter value input box but don't type anything
<img width="871" alt="image" src="https://github.com/lightdash/lightdash/assets/9979944/dfb5c2bf-ab88-4c84-8a17-96506f7cbdf1">

### To Reproduce

1. Open a dashboard with filters. 
2. Focus the filter value text box
3. Attempt to click away
4. You're free to open another filter modal and the old one won't disappear even if you didn't type anything.

### App Version

latest as of 2023-11-13

PR #7963, opened 2023-11-14
fix: hide dashboard filter modals when clicking away

iMac7:
Closes: #7930 

### Description:

closeonclickoutside defaults to true, so it always closes even with the sub-popover focused

[Screencast from 14-11-23 22:31:15.webm](https://github.com/lightdash/lightdash/assets/76876702/2bca2d9d-6d4d-476e-a4a7-c4ac615b5d96)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/Filter.tsx
+++ b/packages/frontend/src/components/DashboardFilter/Filter.tsx
@@ -113,7 +113,6 @@ const Filter: FC<Props> = ({
             trapFocus
             opened={isPopoverOpen}
             closeOnEscape={!isSubPopoverOpen}
-            closeOnClickOutside={!isSubPopoverOpen}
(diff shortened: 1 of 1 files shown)


IrakliJani:
hi @iMac7, thanks for contributing 🙌 

 while this fixes a bug with the linked issue it introduces another issue:

- if you open filter modal
- then you open filter input that has a dropdown (i.e. date filter)
- you click outside
- you loose all state because it closes both modals. that's what that line was protecting from.

customers sometimes click outside by accident and then they loose all filter state

Perhaps we should consider coordinating the filter modals that are open and have a centralized state for all filters. This way, opening one modal would close the others. 

However, the current change isn't producing the desired outcome. I'm going to close this PR, but please feel free to open a new one that addresses both issues.

Issue #7937, opened 2023-11-13
Changing a chart's type requires a dashboard refresh to show up


magnew:
### Description

If I edit a saved chart from a dashboard, then go back to the dashboard, the new chart type doesn't show up without a refresh

### To Reproduce

1. Make a dashboard
2. Add a saved chart to it. 
3. Save
4. Edit the chart from the tile
5. Change the chart type
6. Save
7. Click back to the dashboard
8. You'll see the old chart
9. Refresh

iMac7:
@magnew I can do this, please assign me

PR #7984, opened 2023-11-15
fix: changing chart data requires dashboard refresh

iMac7:
Closes: #7937 

### Description:

I tried invalidating manually with the react-query devtools but only resetting that query works. Chart config was also not changing in the cache but it is now.

[Screencast from 15-11-23 19:11:41.webm](https://github.com/lightdash/lightdash/assets/76876702/84ba927b-d529-4c24-8550-20bc57317f5a)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useSavedQuery.ts
+++ b/packages/frontend/src/hooks/useSavedQuery.ts
@@ -457,7 +457,7 @@ export const useAddVersionMutation = () => {
             );
 
             queryClient.setQueryData(['saved_query', data.uuid], data);
-            queryClient.resetQueries(['savedChartResults', data.uuid]);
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useSavedQuery.ts
```diff
@@ -457,6 +457,7 @@ export const useAddVersionMutation = () => {
             );
 
             queryClient.setQueryData(['saved_query', data.uuid], data);
+            queryClient.resetQueries(['savedChartResults', data.uuid]);
```
ZeRego, line 460:
```suggestion
            await queryClient.resetQueries(['savedChartResults', data.uuid]);
```

ZeRego (approved):
lgtm 👍 tested locally

PR #7985, opened 2023-11-15
feat: dashboard filters with image format

joaoviana:
Relates to: #4425

### Description:

Adds the ability to save `filters` into a scheduler
Overrides dashboard filters on Minimal page aka scheduler page for it to take a screenshot of. 

- [x] There is a third tab in the scheduled delivery modal called `filters`
- [x] The filters tab includes a list of all of the existing dashboard filters
- [x] You cannot add more filters or delete filters from the dashboard, only adjust the values. 
- [x] Adjusting the filter has the same behaviour as adjusting saved filters in a dashboard. E.g.:
  - [x] if the filter is applied to different fields on different tiles, then this is still the case for the scheduled delivery
  - [x] all the filter conditions are joined using `AND` 
  - [x] You can override the operator
  - [x] you can override the value
  - [x] you CANNOT override the field that's being filtered
- [x] temporary filters that you add in `view` mode are _**not**_ included _**or**_ able to be adjusted in the scheduled delivery  (they _**have**_ to be saved filters)
- [x] If you remove a saved filter from a dashboard, then it would also get removed fromt eh scheduled delivery (even if you overrode it or something)

**note**: This doesn't work with `Send now` feature yet

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
@@ -89,6 +89,7 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
                             tileTargets: undefined,
                         });
                     }}
+                    popoverProps={{ withinPortal: true }}
                 />
             </Flex>
         </Stack>
--- a/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
@@ -111,7 +111,7 @@ const updateFilters = (
     originalFilter: DashboardFilterRule,
     schedulerFilters: SchedulerFilterRule[] | undefined,
 ): SchedulerFilterRule[] | undefined => {
-    if (isFilterReverted(originalFilter, schedulerFilter) && schedulerFilters) {
+    if (schedulerFilters && isFilterReverted(originalFilter, schedulerFilter)) {
         return schedulerFilters.filter((f) => f.id !== schedulerFilter.id);
     }
 
@@ -125,7 +125,7 @@ const updateFilters = (
             : originalFilter;
 
     if (hasFilterChanged(filterToCompareAgainst, schedulerFilter)) {
-        if (isExistingFilter && schedulerFilters) {
+        if (schedulerFilters && isExistingFilter) {
             return schedulerFilters.map((f) =>
                 f.id === schedulerFilter.id ? schedulerFilter : f,
             );
--- a/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
@@ -29,7 +29,7 @@ import {
 import { useProject } from '../../../hooks/useProject';
 import { useDashboardContext } from '../../../providers/DashboardProvider';
 
-type SchedulerFilterRule = DashboardFilterRule & {
+type SchedulerFilterRule = Omit<DashboardFilterRule, 'tileTargets'> & {
     tileTargets: undefined;
 };
 
@@ -100,24 +100,26 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
 type SchedulerFiltersProps = {
     dashboard?: Dashboard;
     onChange: (schedulerFilters: SchedulerFilterRule[]) => void;
+    schedulerFilters: DashboardFilterRule[] | undefined;
 };
 
 const SchedulerFilters: FC<SchedulerFiltersProps> = ({
     dashboard,
-    // onChange,
+    schedulerFilters,
+    onChange,
 }) => {
     const { data: project, isLoading } = useProject(dashboard?.projectUuid);
     // TODO: should read initial state from the BE
-    const [schedulerFilters, setSchedulerFilters] = useState<
-        DashboardFilterRule[]
-    >([]);
+    const [filters, setFilters] = useState<DashboardFilterRule[]>(
+        schedulerFilters ?? [],
+    );
 
     const handleUpdateSchedulerFilter = useCallback(
         (schedulerFilter: SchedulerFilterRule) => {
             // TODO: this should diff if the filter is actually
             // different from the dashboard filter
 
-            const newState = produce(schedulerFilters, (draft) => {
+            const newState = produce(filters, (draft) => {
                 const filterIndex = draft.findIndex(
                     (f) =>
                         f.target.fieldId === schedulerFilter.target.fieldId &&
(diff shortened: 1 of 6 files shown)


packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -29,7 +29,7 @@ import {
 import { useProject } from '../../../hooks/useProject';
 import { useDashboardContext } from '../../../providers/DashboardProvider';
 
-type SchedulerFilterRule = DashboardFilterRule & {
+type SchedulerFilterRule = Omit<DashboardFilterRule, 'tileTargets'> & {
```
IrakliJani, line 32: 👍

joaoviana:
I'm aware of some `z-index` issues with select popovers. Looking into it!

packages/frontend/src/pages/MinimalDashboard.tsx
```diff
@@ -19,12 +24,18 @@ const ResponsiveGridLayout = WidthProvider(Responsive);
...
         data: dashboard,
         isError,
         error,
     } = useDashboardQuery(dashboardUuid);
 
+    const { data: scheduler } = useScheduler(schedulerUuid!, {
+        enabled: !!schedulerUuid,
+    });
+
     if (isError) {
```
ZeRego, line 39: Non blocking issue: We need to update this condition based on error from fetching the scheduler. We could do it in a separate PR.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -166,20 +201,27 @@ const SchedulerFilters: FC<SchedulerFiltersProps> = ({
...
+                    {dashboard?.filters?.dimensions.map((filter) => {
+                        const schedulerFilter =
+                            schedulerFiltersData && schedulerFiltersData.length
+                                ? schedulerFiltersData.find(
+                                      (f) =>
+                                          f.target.fieldId ===
+                                              filter.target.fieldId &&
+                                          f.target.tableName ===
+                                              filter.target.tableName,
+                                  )
```
ZeRego, line 214: In `applyDimensionOverrides` and in line 130 in this file, we use the filter id to find a match. Here we use the field ID  and table name. 

This will be a problem if:

- create dashboard with filter A ( with ID X + field ID 1 and table name W)
- create scheduler with filter B (with ID X + field ID 1 and table name W )  --> scheduler filter is valid
- update dashboard with no filters  --> scheduler filter is invalid
- update dashboard with filter C (with ID Y + field ID 1 and table name W) --> scheduler filter is shown since field id and table name match, but onChange returns undefined since the filter ids don't match

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -100,51 +100,86 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
...
+            if (hasChanged) {
+                if (isExistingFilter) {
+                    updatedFilters[filterIndex] = schedulerFilter;
+                } else {
+                    updatedFilters.push(schedulerFilter);
+                }
+            }
+            if (isReverted) {
+                updatedFilters.splice(filterIndex, 1);
+            }
```
ZeRego, line 171: Note: found it hard to review this logic because we are creating variables and making calculations ahead of time. I had to go back and forward in the code to recreate all the different scenarios. I still don't know if there is an edge case where both `hasChanged` and `isReverted` can be `true`


eg: if we know they reverted the filter value, we don't need the `filterToCompareAgainst`, `hasChanged` etc

I wonder if we could have a `if else` where all the logic for `hasChanged` would be inside the else block

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -100,51 +101,86 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
...
-            setSchedulerFilters(newState);
+            if (!originalFilter) return;
 
-            // TODO: sync with upper component
-            // call onChange - this can be debounced
-        },
-        [schedulerFilters],
-    );
+            let updatedFilters = schedulerFiltersData
+                ? [...schedulerFiltersData]
```
ZeRego, line 137: Is there a place where we remove the invalid scheduler filters and then save? 
Looking at the code it looks like we don't show the invalid filters but they will keep poluting the database.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: Ah, I see. Makes total sense to not use this `fieldId` and `tableName` - and to avoid dashboard filters updates + changes to fields with diff operators, values, that could cause unexpected results. Changing it to search by `id` so we ensure uniqueness!

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: great point, going to replicate that and find a solution. I think this is not handled yet!

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -79,6 +79,7 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
                             tileTargets: undefined,
                         });
                     }}
+                    withinPortal
```
ZeRego, line 82: We need the same fix for `<FilterInputComponent`

Screenshot 2023-11-20 at 15 44 14

joaoviana: exactly! thanks 😅

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: Yea, agree. This is def not that straightforward. I've separated this logic into 2 functions: one that handles the `onChange` and a middle one for updating the fields and getting the correct types + updating/removing the changes. 
Also, the check if `isReverted` has been moved further up and it returns so that we avoid extra calculations when that's the case. 

I'm also not sure if there is such a case: reverted **and** changed. I think it's not possible atm

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: done :https://github.com/lightdash/lightdash/pull/7985/commits/e6314449726d23b5e5775c6679d88fb1c5bc1a3d

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: done: https://github.com/lightdash/lightdash/pull/7985/commits/7088c21f0c0364d08a963fdcadef9fd34b98b866

there's also the possibility that the user doesn't interact with the UI, but the scheduler keeps going. _old_/invalid filters should then be removed on the BE through some type of logic. But I think that can be handled on a different PR

joaoviana: I've also reordered some conditional checks to make it more readable but still allow `schedulerFilters` to be `undefined` if no change etc etc [chore: reorder conditional checks on scheduler Filters](https://github.com/lightdash/lightdash/pull/7985/commits/58b4811894e910a1d5e194888185ecaae5524132)

joaoviana:
Tested: 
* dashboard filter -> override on scheduler -> done
* dashboard filter -> remove filter that was being overridden -> done

ZeRego (approved):

Issue #7988, opened 2023-11-16
Charts in Dashboards are getting saved to the wrong place


magnew:
### Description

Charts in Dashboards are getting saved to the wrong place. We have multiple reports of this.

owlas:
`getDefaultSpace` ?

PR #8000, opened 2023-11-16
fix: charts dashboards saved to the wrong place

magnew:
Closes: #7988 

### Description:

Charts were getting created in spaces instead of in dashboards because we weren't sending a dashboard Id. I fixed that and added better types. I guess we need to be very careful with mantine form types. 

I'll think about how else we could have caught this. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/modal/ChartCreateModal.tsx
+++ b/packages/frontend/src/components/common/modal/ChartCreateModal.tsx
@@ -37,6 +37,13 @@ interface ChartCreateModalProps {
     onConfirm: (savedData: CreateSavedChartVersion) => void;
 }
 
+interface ChartCreateModalFormValues {
+    name: string;
+    spaceUuid: string;
+    description: string;
+    newSpaceName: string | null;
+}
+
 const ChartCreateModal: FC<ChartCreateModalProps> = ({
     savedData,
     isOpen,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/modal/ChartCreateModal.tsx
```diff
@@ -118,14 +126,14 @@ const ChartCreateModal: FC<ChartCreateModalProps> = ({
...
-        async (values) => {
+        async (values: ChartCreateModalFormValues) => {
             if (!fromDashboard || !unsavedDashboardTiles || !dashboardUuid)
                 return;
             const newChartInDashboard: CreateChartInDashboard = {
                 ...savedData,
                 name: values.name,
                 description: values.description,
-                dashboardUuid: values.dashboardUuid,
+                dashboardUuid: dashboardUuid,
```
joaoviana, line 136: good spot, that's so easy to dismiss. Guess rule of thumb is to always pass a generic to the `useForm` from now on 👍

joaoviana (approved):

packages/frontend/src/components/common/modal/ChartCreateModal.tsx
magnew: Yea, I thought it would do better with the types. 

I want to add a test too.

Issue #7991, opened 2023-11-16
Custom metric sorting cannot be removed if custom metric name is changed, leads to unfixable error


mhawkins-runa:
### Description

When sorting by a custom metric, if you then change the name of the custom metric, the old metric name persists in the compiled SQL, and cannot be removed. In the screenshot below, it says I am sorting by 3 fields, but I only have 2 selected. This happened after I changed the name of a custom metric that was previously included in the sorting. Now, I cannot run my query without getting an error.

![image](https://github.com/lightdash/lightdash/assets/135957659/79746be5-2444-449f-8c9a-bd0755406acd)

### To Reproduce

Steps & screenshot/gif/loom to reproduce the behavior
1. Create a custom metric in the explorer
2. Sort by the custom metric
3. Change the name of the custom metric
4. Run query and see error

### App Version

v0.863.12

iMac7:
@mhawkins-runa please assign me this

PR #8002, opened 2023-11-17
fix: change custom metric sorting with custom metric name

iMac7:
Closes: #7991 

### Description:

editing a custom metric now changes affected columns. I removed the `TOGGLE-METRIC` dispatch because it resets the custom metric's sort field which I don't think is necessary for an edit.

[Screencast from 17-11-23 10:18:37.webm](https://github.com/lightdash/lightdash/assets/76876702/8fc0aa58-40ec-4996-991d-de43df5d7f43)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/providers/ExplorerProvider.tsx
+++ b/packages/frontend/src/providers/ExplorerProvider.tsx
@@ -804,6 +804,9 @@ function reducer(
         }
 
         case ActionType.EDIT_ADDITIONAL_METRIC: {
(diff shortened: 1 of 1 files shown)


rephus:
Code look good, does this fix the issue for filters ? ( I'll test it in a second) If not, we could fix it if it is not too much trouble 

image

rephus (approved):
This fixes the sorting , I'll create a separate issue for the filter

Issue #8006, opened 2023-11-17
Charts in dashboards sometimes dont save


magnew:
### Description

Charts in dashboards sometimes dont save

PR #8007, opened 2023-11-17
fix: charts sometimes dont save to dashboards

magnew:
Closes: #8006 

### Description:

There was a race condition when returning from creating a chart in a dashboard that cause the new chart to sometimes not be saved. The useEffect that adds the live charts to the dashboard was setting the charts in the dashboard unsafely. It was setting the set of charts when a new chart was added, but sometimes then setting the charts back to the saved charts from the server. This makes sure we only set to the saved charts in the server when there are no other charts. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -204,24 +204,22 @@ const Dashboard: FC = () => {
 
     const { tiles: savedTiles } = dashboard || {};
     useEffect(() => {
-        // TODO: The logic in this useEffect isn't right. It's checking if
-        // there are saved tiles, then checking for unsaved tiles,
-        // then replacing the saved tiles with the unsaved ones if they exist.
         if (savedTiles) {
-            // TODO: maybe this should move in the future, but it makes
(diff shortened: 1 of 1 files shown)


packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -235,14 +233,18 @@ const Dashboard: FC = () => {
...
                 }
+            } else {
+                // If there are no dashboard tiles, set them to the saved ones
+                // This is the first time the dashboard is being loaded.
+                if (!dashboardTiles) {
+                    setDashboardTiles(savedTiles);
+                }
             }
-
-            setDashboardTiles(unsavedDashboardTiles || savedTiles);
```
magnew, line 240: This was the main problem. Sometimes this useEffect would run a second time after setting unsavedDashboardTiles, so it was undefined here and it set them back to the savedOnes.

owlas: Great catch!! Multiple effects interacting, definitely not ideal

owlas (approved):

Issue #7965, opened 2023-11-15
Filter autocomplete doesn't show all available options


hamzahc1:
### Description

A Cloud user reported that filters on dashboards weren't working as expected for the last week, and on certain dashboards, the search wasn't returning results that they knew existed. They were able to manually type out the whole filter value and get the filter to work as expected. 

See link for more details:

TuringLovesDeathMetal:
This query should have `FROM organizations` not `FROM stripe`

<img width="496" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/acdcb9a6-f7a6-4855-ad7f-dec1c7fa1c66">

12ian34:
❤️

PR #8031, opened 2023-11-21
fix: explore search by fieldId to prioritise base table match over joined tables

owlas:
Closes: #7965 

Field search behaviour was different depending on which explore cache we hit. This makes the behaviour the same.

note that this function still has the same limitations of the previous version (finding a field_id is not always reliable depending on explore and alias names)

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -622,7 +622,6 @@ export class ProjectModel {
                     })
                     .select()
                     .whereRaw('? = ANY(table_names)', tableName)
-                    .orWhereRaw("explore->>'base_table' = ?", tableName)
                     .andWhere('project_uuid', projectUuid)
                     .orderBy('baseMatch', 'desc')
                     .first();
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -613,9 +613,18 @@ export class ProjectModel {
             async (span) => {
                 // check individually cached explore
                 let exploreCache = await this.database(CachedExploreTableName)
-                    .select('explore')
+                    .columns({
+                        explore: 'explore',
+                        baseMatch: this.database.raw(
+                            "? = explore->>'base_table'",
+                            [tableName],
```
ZeRego, line 620: We can use the `name` column for this instead of using the jsonb.

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -613,9 +613,18 @@ export class ProjectModel {
...
+                        ),
+                    })
+                    .select()
                     .whereRaw('? = ANY(table_names)', tableName)
+                    .orWhereRaw("explore->>'base_table' = ?", tableName)
```
ZeRego, line 625: The base table is part of the table names so we don't need this.

ZeRego (approved):

Issue #8025, opened 2023-11-21
Can `Send Now` with custom filters on Scheduled deliveries


joaoviana:
### Description

Users can hit `Send now` and these will include the changes made to the scheduled delivery's filters

### What problem does this solve?

Users can test their changes without having to wait (and create) for a scheduled delivery

PR #8032, opened 2023-11-21
feat: add send now scheduler filters to URL

joaoviana:
Closes: #8025 

### Description:

Note: this was already working with CSV exports

With the `Send Now` feature for scheduled deliveries, we want to be able to include the filters to the minimal URL (like we do with the exporting of a dashboard (with saved filter overrides) 

I was looking at different ways to avoid using the URL search param: 
* reuse the `schedulerUuid` search param, but the `send now` feature does not create one 
* set a cookie - didn't work as expected. Even when set properly, the screenshot wouldn't work properly
* set localstorage - this didn't work at all

So, instead, we set a searchParam `sendNowSchedulerFilters` that has the same format as `SchedulerFilterRule[]` and get that or the other (from the scheduler with the `schedulerUuid`), depending on what's present. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -27,6 +27,7 @@ import {
     NotificationPayloadBase,
     ScheduledDeliveryPayload,
     SchedulerAndTargets,
+    SchedulerFilterRule,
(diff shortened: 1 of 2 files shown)


rephus (approved):
This is a good approach.

Issue #7845, opened 2023-11-07
On Redshift `concat` throws an error when using custom dimensions 


rephus:
### Description

Thread: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1699370727543019

Docs : https://docs.aws.amazon.com/redshift/latest/dg/r_CONCAT.html

PR #8046, opened 2023-11-22
fix: concat redshift table calculation

rephus:
Closes:  #7845

### Description:

Before: 

After:

Screenshot from 2023-11-22 10-46-28

Screenshot from 2023-11-22 10-55-28

### Reviewer actions

### DIFF
--- a/packages/backend/src/queryBuilder.test.ts
+++ b/packages/backend/src/queryBuilder.test.ts
@@ -397,9 +397,9 @@ describe('with custom dimensions', () => {
     it('getCustomDimensionSql with empty custom dimension', () => {
         expect(
             getCustomDimensionSql({
+                warehouseClient: bigqueryClientMock,
                 explore: EXPLORE,
                 compiledMetricQuery: METRIC_QUERY,
-                fieldQuoteChar: '"',
                 userAttributes: {},
                 sorts: [],
             }),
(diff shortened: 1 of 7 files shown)


packages/backend/src/queryBuilder.test.ts
```diff
@@ -397,9 +397,9 @@ describe('with custom dimensions', () => {
     it('getCustomDimensionSql with empty custom dimension', () => {
         expect(
             getCustomDimensionSql({
+                warehouseClient: bigqueryClientMock,
```
rephus, line 400: by default I'm testing with bigquery, which is going to be using the default CONCAT method, however, the `quotes` are different in bigquery, so I had to update most of the SQL statements.

packages/backend/src/queryBuilder.test.ts
```diff
@@ -409,9 +409,10 @@ describe('with custom dimensions', () => {
     it('getCustomDimensionSql with custom dimension', () => {
         expect(
             getCustomDimensionSql({
+                warehouseClient: bigqueryClientMock,
+
                 explore: EXPLORE,
                 compiledMetricQuery: METRIC_QUERY_WITH_CUSTOM_DIMENSION,
-                fieldQuoteChar: '"',
```
joaoviana, line 414: Could you explain what this was used for?

joaoviana (approved):
LGTM

packages/backend/src/queryBuilder.test.ts
rephus: This is the character the warehouse uses to wrap their fields , eg `"table.column"`

Issue #8050, opened 2023-11-22
Compute totals on BE instead of FE


joaoviana:
description tbd

PR #8058, opened 2023-11-22
fix: remove calculate total from frontend on chart edit and view

rephus:
Closes:  #8050
### Description:

I didn't remove the `useColumnTotals` hook because it is still used in underlying data and sql runner

Screenshot from 2023-11-23 13-07-54

### Reviewer actions


packages/frontend/src/hooks/useCalculateTotal.ts
```diff
@@ -119,23 +117,12 @@ export const getCalculationColumnFields = (
...
         ?.map((item) => {
             return itemsMap[item];
         })
-        .filter(
-            (item) =>
-                isField(item) &&
-                isMetric(item) &&
-                !numericTypes.includes(item.type.toString()),
-        );
+        .filter((item) => isField(item));
```
ZeRego, line 125: We removed `isMetric()` on purpose? 
Now it makes a request even if there are only dimensions. 
We can also move this logic inside the reduce and have a single loop and simplify the code. 

Screenshot 2023-11-24 at 09 18 08

rephus: I did it in purpose, but  perhaps it was not a good idea, I'll put it back

ZeRego (approved):

Issue #8049, opened 2023-11-22
Scheduled delivery filters should work with overridden disabled saved filters


joaoviana:
### Description

Scheduled Delivery filters do not work as expected when a overriding a saved filter that is set as disabled (aka any value)

To Reproduce:
* Have Scheduled delivery filters feature enabled
* Create a Saved Filter with `any value` (disabled)
* Override value in Scheduled Delivery filters section
* No changes applied to the output of the delivery

Expected: 
* Set values to disabled filter
* See the changes on the outputted results (image/csv)

PR #8059, opened 2023-11-22
fix: handle disabled scheduled filters overrides

joaoviana:
Closes: #8049

### Description:

When overriding a saved filter that is `disabled` (accepts any value), we should turn off that flag if there have been changes. 

Also, when checking if a filter is reverted back to original, we should not compare the `disable` values.

### Reviewer actions


owlas (approved):

owlas:
CleanShot 2023-11-24 at 14 32 53@2x

Should this show the `any value` state? It would be more consistent with the dashboard.

joaoviana:
> CleanShot 2023-11-24 at 14 32 53@2x
> 
> Should this show the `any value` state? It would be more consistent with the dashboard.

great point! Let me do something for this, sharing soon!

Issue #7990, opened 2023-11-16
Update docs for column totals. 


TuringLovesDeathMetal:
Maybe we add a page in `troublsehooting` called `Why don't my totals match the values in my table?` (see Looker doc for inspo: https://cloud.google.com/looker/docs/best-practices/how-to-troubleshoot-looker-totals)

- [ ] explain how you can get different totals than if you're just adding things up in the table:
     - [ ] explain how metric filters aren't included 
     - [ ] explain how the limit in the table isn't included

joaoviana:
Add note in docs that we don't have this on pivot tables

PR #8060, opened 2023-11-22
docs: adds totals to chart type docs and updates sidebar

TuringLovesDeathMetal:
Closes: #7990 

### Description:
Adds docs for table totals.


docs/sidebars.js
```diff
@@ -75,20 +75,6 @@ module.exports = {
             type: 'category',
             label: 'Guides',
             items: [
-                {
```
owlas, line 78: what's this change for? Does it change the URL?

TuringLovesDeathMetal: It doesn't change the URL. And I've been meaning to do it for a while now - it just feels weird inside of the `guides` submenu (it feels a bit too hidden). 

I can revert it if you don't like it

owlas: makes sense! I was checking if this was intended because it wasn't related to the PR

PR #8072, opened 2023-11-23
fix: upgrade tsoa to 6.0.0-rc.5

joaoviana:
Closes: N/A

### Description:

Fix the following error when generating `spec-and-routes` with `tsoa`: 
image

### Reviewer actions

Trigger all e2e tests:
test-cli
test-backend
test-frontend

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -4,6 +4,7 @@
     "main": "dist/index",
     "license": "MIT",
     "devDependencies": {
+        "@tsoa/cli": "6.0.0-rc.5",
         "@types/bcrypt": "^5.0.0",
         "@types/connect-flash": "^0.0.37",
         "@types/cookie-parser": "^1.4.2",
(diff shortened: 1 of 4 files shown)


rephus (approved):
I tested on render and everything seems to be working fine, however, it seems the routes has changed its type, not sure if we should also migrate @tsoa/runtime to match this new version or if this is backwards compatible.

Screenshot from 2023-11-24 08-48-22
Screenshot from 2023-11-24 08-48-42

The only breaking change I found was on 5.0.0 : 

> Breaking changes: Dropped support for Node 12

rephus:
Some user at tributes are failing, I don't think it is related with this PR 

image

joaoviana:
> I tested on render and everything seems to be working fine, and the routes file also seem very similar.
> 
> I'm not sure if we should also migrate @tsoa/runtime to match this new version or leave it like this.
> 
> Screenshot from 2023-11-24 08-48-22 Screenshot from 2023-11-24 08-48-42
> 
> The only breaking change I found was on 5.0.0 :
> 
> > Breaking changes: Dropped support for Node 12

oooh, forgot to check that one! let me see if it supports the same version 👍

Issue #8036, opened 2023-11-21
Add `date zoom` to dashboards in view mode


TuringLovesDeathMetal:
### Problem

My end users want to easily switch between different date granularities on my dashboard. For example, seeing sales by day instead of week, or week instead of month, etc. 

Right now, to do this, I need to create multiple dashboards with different date granularities for each chart to accomplish this. 

This is really annoying because basically every chart I make that’s important and is grouped by dates, I’ll need to make multiple copies of the chart (where the only thing that changes is the date granularity).

### Solution

There should be a widget/button on a dashboard where I can change the date granularity of the charts on my dashboard as a person viewing the dashboard. 

![CleanShot 2023-11-21 at 16 47 42@2x](https://github.com/lightdash/lightdash/assets/11660098/ce0cdc6f-a408-4d8a-addf-0cfd11b481e1)

### Acceptance Criteria

- [x] There is a button at the top of the dashboard which I can interact with in view mode to change the date granularity. 
  - [x] The button shows up in `edit` mode,  but I can't interact with it.
  - [x] This is only available in dashboards, not in charts
- [x] There are four options in the date granularity: `default`, `day`, `week`, `month`, `quarter`, `year`
  - [x] For now, we won't support any granularity less than a day
  - [x] By default, date zoom is set to `default` - this makes no changes to the underlying date dimension data.
- [x] date zoom is _**always**_ available in view mode (for now, there's no way to turn it on/off...that will come later)
- [ ]  If there is a date dimension in my results table, then the granularity of that dimension is changed.
    - e.g. in big value comparisons, I’d expect the date granularity to change and the comparison to update
    - in a chart with a date axis, I’d expect the date granularity to change
- [ ]  It doesn’t matter what my available dimension granularities are, we just override it.
    - e.g. if I have `date` and I have *********not********* included `date week` as a dimension, users can still pick `week` in the dashboard and it will work on that date field.
- [ ] If there are multiple date fields in a results table, we change the first date field results table(this is the left-most field in the results table).
- [x] You cannot save changes in date zoom. 
  - E.g. if I set date zoom to `week`, then I close and reopen the dashboard, the dashboard would have `date zoom` back to `default`.
- [x] date zoom is not applied to scheduled deliveries. E.g. if I have date zoom set, then I try to schedule a delivery of a dashboard, it will have the date zoom set to default. 
[...]

PR #8078, opened 2023-11-24
feat: date zoom

rephus:
Closes: #8036

### Description:
[Screencast from 27-11-23 09:15:40.webm](https://github.com/lightdash/lightdash/assets/1983672/6358c942-15f5-4e4c-8699-aa83e54da1e3)

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/savedChartController.ts
+++ b/packages/backend/src/controllers/savedChartController.ts
@@ -54,7 +54,7 @@ export class SavedChartController extends Controller {
             dashboardFilters?: any; // DashboardFilters; temp disable validation
             invalidateCache?: boolean;
             dashboardSorts?: SortField[];
-            granularity?: string;
+            granularity?: DateGranularity;
         },
         @Path() chartUuid: string,
         @Request() req: express.Request,
@@ -69,9 +69,7 @@ export class SavedChartController extends Controller {
                 versionUuid: undefined,
                 invalidateCache: body.invalidateCache,
                 dashboardSorts: body.dashboardSorts,
-                granularity: body.granularity
-                    ? (body.granularity as DateGranularity)
-                    : undefined,
+                granularity: body.granularity,
             }),
         };
     }
--- a/packages/backend/src/controllers/savedChartController.ts
+++ b/packages/backend/src/controllers/savedChartController.ts
@@ -5,6 +5,7 @@ import {
     ApiGetChartVersionResponse,
     ApiSuccessEmpty,
     DashboardFilters,
+    DateGranularity,
     SortField,
     TimeFrames,
 } from '@lightdash/common';
@@ -68,7 +69,9 @@ export class SavedChartController extends Controller {
                 versionUuid: undefined,
                 invalidateCache: body.invalidateCache,
                 dashboardSorts: body.dashboardSorts,
-                granularity: body.granularity as TimeFrames,
+                granularity: body.granularity
+                    ? (body.granularity as DateGranularity)
+                    : undefined,
             }),
         };
     }
--- a/packages/backend/src/controllers/savedChartController.ts
+++ b/packages/backend/src/controllers/savedChartController.ts
@@ -48,13 +48,10 @@ export class SavedChartController extends Controller {
     @SuccessResponse('200', 'Success')
     @Post('/results')
     @OperationId('postChartResults')
-    async postDashboardTile(
+    async postChartResults(
         @Body()
         body: {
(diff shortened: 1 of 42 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -956,12 +1021,23 @@ export class ProjectService {
             explore,
         });
 
+        // TODO quick hack to get the old results on the chart with new granularity
+        // We should investigate if we can make the changes in the frontend
```
joaoviana, line 1025: If you want to, we can pass `oldData` to the hook on the FE, so that the BE does minimal changes on this `runViewChartQuery`. So it could only do: 
- updateMetricQueryGranularity(
            metricWithOverrideSorting,
            explore,
            granularity,
        );
        and 
-  row mapping

what do you think?

rephus: I think we should do 2 (modify chart query in frontend) https://lightdash.slack.com/archives/C067071JKKN/p1700823674033019 and remove this fake result mapping. to ensure charts are consistent with the granularity, but we could do it on a separate PR 

We can't make changes to the metricquery in the frontend for security reasons.

packages/backend/src/queryBuilder.ts
```diff
@@ -26,18 +28,37 @@ import {
...
+
+    if (dimension === undefined) {
+        const { baseDimensionId, newTimeFrame } = getDateDimension(dimId);
+
+        if (baseDimensionId) {
+            const baseField = getDimensionFromId(baseDimensionId, explore);
+            if (baseField && newTimeFrame)
+                return {
+                    ...baseField,
+                    compiledSql: `DATE_TRUNC('${newTimeFrame}', ${baseField.compiledSql})`,
```
rephus, line 53: Creates a missing time frame on the fly

packages/backend/src/controllers/savedChartController.ts
```diff
@@ -52,6 +54,7 @@ export class SavedChartController extends Controller {
             dashboardFilters?: any; // DashboardFilters; temp disable validation
             invalidateCache?: boolean;
             dashboardSorts?: SortField[];
+            granularity?: string;
```
joaoviana, line 57:
```suggestion
            granularity?: DateGranularity;
```
can't we do this?

packages/frontend/src/hooks/dashboard/useDashboardChart.ts
```diff
@@ -7,11 +7,13 @@ const useDashboardChart = (tileUuid: string, savedChartUuid: string | null) => {
     const dashboardFilters = useDashboardFiltersForTile(tileUuid);
     const chartSort = useDashboardContext((c) => c.chartSort);
     const tileSort = chartSort[tileUuid] || [];
+    const granularity = useDashboardContext((c) => c.dateGranularity);
```
joaoviana, line 10:
```suggestion
    const granularity = useDashboardContext((c) => c.dateZoomGranularity);
```

packages/common/src/utils/timeFrames.test.ts
```diff
@@ -61,4 +61,40 @@ describe('TimeFrames', () => {
             );
         });
     });
+
+    describe('getDateDimension', () => {
+        test('should parse dates', async () => {
```
joaoviana, line 66: 🙌

packages/backend/src/queryBuilder.ts
joaoviana: Would there be a case where the code inside the if statement (L46-55) could be run without the intent to Date Zoom? 
For example, a date dimension that can't be found, but we return a new one with the `missing time frame on the fly` but the user action is not to `date zoom` but instead compile a project or something?

joaoviana (commented):
left some comments ahead of review 👍

rephus:
> Would there be a case where the code inside the if statement (L46-55) could be run without the intent to Date Zoom?

I don't think so :thinking:  afaik this buildQuery only runs when we run queries or when we show the SQL on explore. . In those cases we always provide valid dimensions. 

And we are only doing this for missing time frames, which is something we might use in the future for other use cases too.

joaoviana (approved):
Code looks good!

On https://lightdash-pr-8078.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/dashboards/b3df331b-ed0f-47b7-becf-d95b18ffaf51/view

when I choose `Date Zoom: month` the data visualisation looks good expected for the axis, which I think it's related to the chartConfig changes that will be done later, correct?

Issue #8067, opened 2023-11-23
The `which users have run the most queries in the last 7 days` chart looks like it's undercounting the number of queries executed


TuringLovesDeathMetal:
### Description

Here's is what I'm seeing in our internal data (using this query: https://analytics.lightdash.cloud/share/VaANS390xwaveTreeYn5x)
<img width="1445" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/ea9c17b3-5342-4e3b-8141-3e747438d04a">

Here is what I'm seeing on our usage analytics dashboard:

<img width="666" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/4b79c971-8881-48df-a5b1-314f24ecb54c">

PR #8079, opened 2023-11-24
refactor: split chart results endpoint

ZeRego:
Closes: 

### Description:

The `/results` endpoint was being used for the dashboard page and the chart page in view mode.
This was causing a few issues:
- increasing logic and params to handle both scenarios 
- since we stopped fetching the chart separately on the dashboard page, we stopped counting chart views (#8067)
  - this will be fixed in following PR 
 
Now we have
`/results` used by chart page in view mode
`/chart-and-results` used by dashboard page

Even though most of the service code is similar, it is now simpler and easier to keep diverging logic between them. eg: count for chart view in `/chart-and-results` 

### Reviewer actions

### DIFF
--- a/packages/backend/src/generated/swagger.json
+++ b/packages/backend/src/generated/swagger.json
@@ -2590,6 +2590,7 @@
                     },
                     "filters": {
                         "properties": {
+                            "tableCalculations": {},
                             "metrics": {},
                             "dimensions": {}
                         },
--- a/packages/backend/src/generated/swagger.json
+++ b/packages/backend/src/generated/swagger.json
@@ -2590,7 +2590,6 @@
                     },
                     "filters": {
                         "properties": {
-                            "tableCalculations": {},
                             "metrics": {},
                             "dimensions": {}
                         },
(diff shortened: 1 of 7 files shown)


joaoviana (commented):
Looks great! Will also test in Render `Explore from here` once conflicts are resolved 👍

packages/backend/src/generated/swagger.json
```diff
@@ -2590,7 +2590,6 @@
                     },
                     "filters": {
                         "properties": {
-                            "tableCalculations": {},
```
joaoviana, line 2593: 🤔 why... it shouldn't remove them, unless there's something weird going on with tsoa?

ZeRego: Ah!, because `generate-api` doesn't build the types from common. I'm going to fix the command to always build common before running tsoa.

ZeRego: PR to fix that https://github.com/lightdash/lightdash/pull/8089

joaoviana (approved):

PR #8080, opened 2023-11-24
chore: enable posthog on backend

joaoviana:
Closes: N/A

### Description:

Adds `node-posthog` (followed this: https://posthog.com/docs/libraries/node)

Instantiates new PostHog client on the server; 
Identifies the user and group (same way we do in: https://github.com/lightdash/lightdash/blob/chore/enable-posthog-usage-on-backend/packages/frontend/src/providers/ThirdPartyServicesProvider.tsx#L12)
Shuts down `onExit`

Example: 
```ts
  const isFeatureEnabled = await postHogClient.isFeatureEnabled(
            'date-zoom',
            user.userId.toString(),
        );
```
image

### Reviewer actions

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -41,7 +41,7 @@ import { SchedulerWorker } from './scheduler/SchedulerWorker';
 import { VERSION } from './version';
 import { registerNodeMetrics } from './nodeMetrics';
 import { wrapOtelSpan } from './utils';
-import { postHogClient } from './posthog';
+import { postHogClient } from './postHog';
 
 // @ts-ignore
 // eslint-disable-next-line no-extend-native, func-names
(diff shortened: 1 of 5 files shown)


packages/backend/src/index.ts
```diff
@@ -280,6 +281,30 @@ passport.deserializeUser(async (id: string, done) => {
...
+
+        postHogClient.groupIdentify({
+            groupType: 'organization',
+            groupKey: 'organization',
+            properties: {
+                uuid: user.organizationUuid,
+                name: user.organizationName,
+            },
+        });
+
```
joaoviana, line 307: I'm not sure if I should do it here, or somewhere else. What do we think?

ZeRego (approved):

owlas (changes requested):
There is an existing function for identifying, the logic should live there I believe: https://github.com/lightdash/lightdash/blob/30f3a84d6979ec3f50af99353d18f432bb57519b/packages/backend/src/analytics/client.ts#L17

joaoviana:
> There is an existing function for identifying, the logic should live there I believe:
> 
> https://github.com/lightdash/lightdash/blob/30f3a84d6979ec3f50af99353d18f432bb57519b/packages/backend/src/analytics/client.ts#L17

ahhh, I see! Thanks, will move it then 👍

PR #8083, opened 2023-11-24
Feat/invalidate charts with dates when date zooming

joaoviana:
Relates to: #8036

### Reviewer actions


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1031,6 +1032,21 @@ export class ProjectService {
                       [oldDimension]: row[newDimension],
                   }))
                 : rows;
+        const exploreDimensions = getDimensions(explore);
```
joaoviana, line 1035: I know we also `getDimensions` when running `updateMetricQueryGranularity` in https://github.com/lightdash/lightdash/pull/8078/files - maybe we can pass the this value to `updateMetricQueryGranularity` and we run this piece of code L1035-1049 before that? what do you think @rephus ?

rephus: yeah, feel free to change this to make it more efficient

packages/frontend/src/hooks/useQueryResults.tsx
```diff
@@ -187,21 +187,57 @@ export const useChartResults = (
...
+    >(queryKey);
+
     const timezoneFixFilters =
         dashboardFilters && convertDateDashboardFilters(dashboardFilters);
 
+    useEffect(() => {
+        if (!isDateZoomFeatureEnabled) return;
+
+        // Check if dateZoomGranularity has changed and if date dimensions are present
+        const hasDateZoomChanged =
```
rephus, line 220: this is already present in the queryKey, why do you need to check again here ?

packages/common/src/types/metricQuery.ts
```diff
@@ -52,6 +52,9 @@ export type MetricQuery = {
     tableCalculations: TableCalculation[]; // calculations to append to results
     additionalMetrics?: AdditionalMetric[]; // existing metric type
     customDimensions?: CustomDimension[];
+    metadata?: {
```
rephus, line 55: mm, instead of adding a new metadata property on the request, I think we can just send the metric query and recalculate the `hasDateDimension` from the metricQuery in the backend. But this is personal opinion.

joaoviana: In order to recalculate the `hasDateDimension` we would have to getExplore(...), which can be heavy and reduce interactivity. Doing it on the BE allows us to use the existing explore (because we query it already on `getChartAndResults` )

packages/frontend/src/hooks/useQueryResults.tsx
joaoviana: we don't! you're right. Thanks for spotting this

packages/frontend/src/hooks/useQueryResults.tsx
```diff
@@ -202,21 +202,48 @@ export const useChartAndResults = (
...
         dashboardFilters && convertDateDashboardFilters(dashboardFilters);
 
+    useEffect(() => {
+        if (!isDateZoomFeatureEnabled) return;
+
+        const hasDateDimensions =
+            apiChartAndResults?.metricQuery?.metadata?.hasADateDimension;
+
+        if (hasDateDimensions) {
+            queryClient.invalidateQueries(queryKey, { refetchInactive: true });
```
rephus, line 237: perhaps we could also use the queryKey for this, 

for example: 

```
queryKey = [hasDateDimensions? granularity: undefined] 
```

would it work ?

packages/frontend/src/hooks/useQueryResults.tsx
```diff
@@ -202,35 +202,72 @@ export const useChartAndResults = (
...
-        sortKey,
-        granularity,
-    ];
+    const queryKey = useMemo(
+        () => [
+            'savedChartResults',
+            chartUuid,
+            dashboardFilters,
+            invalidateCache,
+            sortKey,
```
joaoviana, line 219: @rephus removed the `granularity` from the `queryKey` because after setting a Date Zoom value, all the queries would be requested.
Then, thought about `invalidatingQueries` (without the granularity in the query key) and that didn't work because React Query would fetch the invalidated query with the same payload.

With the useEffect below, we refetch the queries that have a date dimension. 
This also allows us to refetch the affected queries if, for example, the date zoom is reset (back to `undefined`)

rephus: > This also allows us to refetch the affected queries if, for example, the date zoom is reset

Why do we want to do this ? If a query is cached by the frontend, then we don't need to make a request to fetch the data again, afaik this is not like the `invalidate cache` argument. 

For example, I don't think the second `month` request should happen here: 

[Screencast from 28-11-23 08:51:33.webm](https://github.com/lightdash/lightdash/assets/1983672/dbecd846-4bc0-41b1-bfd2-4bdf64b646d5)

joaoviana: > Why do we want to do this ? If a query is cached by the frontend, then we don't need to make a request to fetch the data again, afaik this is not like the invalidate cache argument. 

I got it! Great point on the  `For example, I don't think the second month request should happen here:` 

I've added the `granuarity` to the `queryKey` conditionally. Thanks for spotting this @rephus, pushing soon

joaoviana: and here it is! https://github.com/lightdash/lightdash/pull/8083/commits/0cd6022420452e22ba91615603b005a1900cd40f

PR #8084, opened 2023-11-24
feat: display filter summary if not overridden

joaoviana:
Closes: N/A

### Description:

As suggested in: #8059, add copy next to the field label when: 
* filter is `disabled`/accepts any value 
and
* there's been no changes 

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/0126b09f-e85f-41f2-84e6-996967dc216f

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
@@ -23,6 +23,7 @@ import FieldIcon from '../../../components/common/Filters/FieldIcon';
 import FieldLabel from '../../../components/common/Filters/FieldLabel';
 import {
     FilterInputComponent,
+    getConditionalRuleLabel,
     getFilterOperatorOptions,
 } from '../../../components/common/Filters/FilterInputs';
 import {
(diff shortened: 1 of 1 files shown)


ZeRego:
Not a blocker, but I found it confusing to see `is any value` above the empty inputs. 
eg idea:
UI when there is no override: 
```Filter label + value summary + [pencil icon] ```
UI when there is no override: 
```Filter label + inputs + [rollback icon]```

**Comment by joaoviana** [comment]
> Not a blocker, but I found it confusing to see `is any value` above the empty inputs. eg idea: UI when there is no override: `Filter label + value summary + [pencil icon] ` UI when there is no override: `Filter label + inputs + [rollback icon]`
> 
> It will make the UI smaller since we won't display so many inputs that are not being used. And it's clear what filters are being overridden or not.

nice, understand this concern @ZeRego, I've made the following changes with your suggestions

https://github.com/lightdash/lightdash/assets/7611706/445af7b9-51e7-4599-8448-62b75a32ad50

I kept the revert icon on the left just for position reasons. But happy to change 👍

**Comment by ZeRego** [comment]
Edit not working as expected.

https://github.com/lightdash/lightdash/assets/9117144/a3a2c941-34e8-4e4b-81b7-687fba6f7fa4

**Comment by joaoviana** [comment]
> Edit not working as expected.
> 
>  jose.-.Lightdash.-.27.November.2023.mp4 
> Btw, my suggestion was for all untouched filters but feel free to ignore. eg: if filter is original ? show summary : show inputs

oops, checking that! 
I see what you mean, having a look

File: packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx [hunk 1]
```diff
@@ -34,6 +35,33 @@ import { isFilterConfigRevertButtonEnabled as hasSavedFilterValueChanged } from
...
+            <Text span color="gray.7">
+                {filterSummary?.operator}{' '}
+            </Text>
+            <Text fw={700} span>
+                {filterSummary?.value}
+            </Text>
+        </Text>
+    );
+};
+
```
joaoviana, line 64: now we display the filter summary (like on the dashboards) if not overriden

Issue #7397, opened 2023-10-10
Analytics events for scheduled deliveries filters


TuringLovesDeathMetal:
### Description

Add parameter saying if a scheduled delivery that was created has a filter adjusted. 

This should be in the `scheduler_created` and `scheduler_updated` events.

`filters_updated` (True/False maybe? maybe `filters_updated_num`)

PR #8112, opened 2023-11-27
feat: add filter data to scheduler events

joaoviana:
Closes: #7397

### Description:
Add `filtersUpdatedNum` to: 
* `scheduler.created`
* `scheduler.updated`

Also, type event data objects with the already created types.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/SavedChartsService/SavedChartService.ts
+++ b/packages/backend/src/services/SavedChartsService/SavedChartService.ts
@@ -29,6 +29,7 @@ import { analytics } from '../../analytics/client';
 import {
     ConditionalFormattingRuleSavedEvent,
     CreateSavedChartVersionEvent,
+    SchedulerUpsertEvent,
 } from '../../analytics/LightdashAnalytics';
 import { schedulerClient, slackClient } from '../../clients/clients';
 import { getSchedulerTargetType } from '../../database/entities/scheduler';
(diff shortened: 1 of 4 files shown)


packages/backend/src/services/SavedChartsService/SavedChartService.ts
```diff
@@ -664,7 +665,8 @@ export class SavedChartService {
                         ? []
                         : scheduler.targets.map(getSchedulerTargetType),
```
rephus, line 666: why is this not failing if `filtersUpdatedNum` is not optional ?

joaoviana: here it uses `SchedulerUpsertEvent` which doesn't include `filtersUpdatedNum` 
```ts
export type SchedulerUpsertEvent = BaseTrack & {
    event: 'scheduler.created' | 'scheduler.updated';
    userId: string;
    properties: {
        projectId: string;
        organizationId: string;
        schedulerId: string;
        resourceType: 'dashboard' | 'chart';
        cronExpression: string;
        cronString: string;
        resourceId: string;
        format: SchedulerFormat;
        targets: Array<{
            schedulerTargetId: string;
            type: 'slack' | 'email';
        }>;
    };
};
```

In `DashboardService` it uses `SchedulerDashboardUpsertEvent` type, which includes that property `filtersUpdatedNum` as required 👍

rephus (approved):

Issue #8113, opened 2023-11-27
In the SQL runner, fields with dots in the name error out


magnew:
### Description

In the SQL runner, when there is a `.` in a field name, we use a fieldId to reference the column, which removes the dot. But the data still has the original field ID, so when the table tries to reference it, it fails. 

### To Reproduce

In the SQL runner, run any query with a dot in the field name eg 

`SELECT name as "thing.thing" FROM "postgres"."jaffle"."generated_a"  LIMIT 25`

TuringLovesDeathMetal:
closing this as not planned - I believe that we changed teh SQl runner, so I'm assuming this isn't a problem anymore. can always reopen

PR #8114, opened 2023-11-27
fix: dont crash when column name contains a dot

magnew:
### Description:

When a column name in the SQL runner contains a dot, the front end crashes. This fixes the crash by catching the error and logging the error. We could do more to handle this case, but I wasn't sure the best approach to fixing the underlying problem and fixing the crash is necessary anyway. With this change a field that has an error will show the string 'Error' in the table and log in the console. I hadn't put 'Error' in the results at first, but almost anything else could be misleading -- I wan't to make sure they know they aren't seeing data. 

The real problem here is that when there is a `.` in a field name, we use a fieldId to reference the column, which removes the dot. But the data still has the original field ID, so when the table tries to reference it, it fails. I'm not sure the best fix and since it's a bit of an edge-case with good workarounds, I left the rest of the behavior as is. I filed a ticket: https://github.com/lightdash/lightdash/issues/8113

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useSqlRunnerColumns.ts
+++ b/packages/frontend/src/hooks/useSqlRunnerColumns.ts
@@ -32,9 +32,16 @@ const useSqlRunnerColumns = ({
                             ? columnHeader(dimension)
                             : dimension.label,
                     cell: (info) => {
-                        const {
-                            value: { raw },
-                        } = info.getValue();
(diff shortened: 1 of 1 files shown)


rephus (approved):
The fix makes sense. Is there a way to replicate this on render without updating the schema ?

Issue #6826, opened 2023-08-22
Add dashboard_uuid to query tags


ian-whitestone:
### Description

A bunch of useful metadata already gets added to the Snowflake query tags from the [usage analytics feature](https://docs.lightdash.com/references/usage-analytics/#query-tags). Here's an example:

`{"organization_uuid":"a2eca2c6-6bdb-470a-aef9-bf50f02e1b86","project_uuid":"18206b89-d721-4fe5-8edf-2ac395159961","user_uuid":"5090ebbb-102d-4f02-b892-35ec09838cc4","chart_uuid":"411a978d-b13d-42ab-8627-28aa695870c3"}`

I would love if `dashboard_uuid` was also populated. 

### What problem does this solve?

* Understand the Snowflake compute costs associated with each dashboard
* Understand the total time required to run all queries for a given dashboard 
* Understand which dashboards are being used the most frequently
* Understand which charts are used across the most dashboard

PR #8129, opened 2023-11-28
feat: add dashboard uuid to query tags

ZeRego:
Closes: #6826

### Description:

- require dashboard uuid
- update other params to be required

Screenshot 2023-11-28 at 14 06 02

### Reviewer actions


packages/frontend/src/hooks/dashboard/useDashboardChart.ts
```diff
@@ -11,6 +12,7 @@ const useDashboardChart = (tileUuid: string, savedChartUuid: string | null) => {
 
     return useChartAndResults(
         savedChartUuid,
+        dashboardUuid ?? null,
```
joaoviana, line 15: I guess it would be really unlikely for this to run without a `dashboardUuid`, right? 
should we update types in `postDashboardTile` to be `string | null`?

joaoviana: Ignore, it's only `enabled` if it exists

joaoviana (approved):

PR #8148, opened 2023-11-29
fix: filter autocomplete scroll

ZeRego:
Closes: #7835

### Description:

changes:
- memo dropdown component
- don't clear search after selecting one value

https://github.com/lightdash/lightdash/assets/9117144/a254d38e-f194-4a86-a411-e84d3a26e7f3

### Reviewer actions


joaoviana (approved):
Works great! 

Not sure if worth it, but when selecting with `Enter` it clears the search because of the `handleKeyDown` 👍

ZeRego:
> Not sure if worth it, but when selecting with Enter it clears the search because of the handleKeyDown 👍

Good find, fixing that will then cause the search not to be cleared when adding a new value. The handleKeyDown is actually working on a happy coincidence 🙃 something to worry about later since there is no other solution.

PR #8203, opened 2023-12-04
fix: props missing from gsheets api

magnew:
Closes: 

### Description:

A couple of the newer props, `tableCalculations` and `metadata`, were missing from the types used by the gsheets export endpoint, causing a validation error. 

I tested this and it fixes the error. But I don't know a lot about these props and I'm happy to change this if another approach is needed. 

### Reviewer actions


rephus (approved):
Good catch, we should build e2e tests for the gsheet api to make sure this doesn't happen again, let's create a ticket for it.

PR #8206, opened 2023-12-05
fix: binning width type casting

IrakliJani:
### Description:

bin width should be an integer where it can't divide a number to an int we should convert float to int

### Reviewer actions


rephus (approved):
Looks good, merge if tests are passing

rephus (approved):

Issue #8153, opened 2023-11-29
Refactor VisualizationProvider to expose an itemMap instead of the explore


ZeRego:
The visualization provider shouldn't be directly dependent on the explore object/data.
Instead, it should base itself on an `itemsMap: Record<string, Item>` which contains all the necessary data for the visualization and configuration.

 
Update: 
- add `itemMap` to context ( the value should come from the explore + resultsData?.metricQuery )
- remove `explore`, `dimensions`, `metrics`, `customDimensions`, `customMetrics`, `allMetrics`, `allNumericMetrics` from the context and use the `itemMap` instead

This will:
- reduce duplicated code
- reduce hooks loops
- centralize data+logic

This is preparation work for [Refactor metric query results types#8145](https://github.com/lightdash/lightdash/issues/8145)

PR #8214, opened 2023-12-05
refactor: add explorer name to MetricQuery

ZeRego:
Relates to: #8153 

### Description:

The MetricQuery should have the explore name since it depends on it. At the moment we pass the explorer name as a separate argument to functions that also consume the MetricQuery.

This PR is just updating types and making sure all code works. No business logic was changed. 
Added some logic for URL params be backwards compatible.

### Reviewer actions


joaoviana:
Tested: 
explore from here on chart, dashboard
sql runner

joaoviana (changes requested):
Would there be a way to allow `exploreName` to be optional? 

Users might refer to some `explore from here`s made in the past and would be annoying if they couldn't run these queries again - and the only way would be to re-create them 😞

this is what they would see if `exploreName` wasn't on the URL, right? 
image

Rest of the scenarios look great 👍

joaoviana (approved):
tested: 
* backwards compatibility without the `exploreName` - viz remains intact 👍

Issue #8105, opened 2023-11-27
Date zoom should be updating the chart config to show the right time frame on axes and labels


rephus:
On https://lightdash-pr-8078.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/dashboards/b3df331b-ed0f-47b7-becf-d95b18ffaf51/view

when I choose `Date Zoom: month` the data visualisation looks good expected for the axis, which I think it's related to the chartConfig changes that will be done later, correct?

_Originally posted by @joaoviana in https://github.com/lightdash/lightdash/pull/8078#pullrequestreview-1750650733_

rephus:
We should do first: [Refactor VisualizationProvider to expose an itemMap instead of the explore#8153](https://github.com/lightdash/lightdash/issues/8153)

PR #8225, opened 2023-12-06
refactor: stop using all explore fields in viz provider

ZeRego:
Closes: 

### Description:

This unblocks #8105 as the BE should return the correct label.

Updated the BE to convert additional metrics to metrics. 

### Reviewer actions

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -17,7 +17,7 @@ import {
     FilterableField,
     friendlyName,
     isDimension,
-    Item,
+    ItemsMap,
     Metric,
     TableCalculation,
 } from './types/field';
@@ -323,7 +323,7 @@ export type ApiQueryResults = {
     metricQuery: MetricQuery;
     cacheMetadata: CacheMetadata;
     rows: ResultRow[];
-    fields: Record<string, Item | AdditionalMetric>;
+    fields: ItemsMap;
 };
 
 export type ApiChartAndResults = {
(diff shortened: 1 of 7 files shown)


packages/common/src/index.ts
```diff
@@ -756,11 +756,6 @@ export function getFieldMap(
     );
 }
 
-export type ItemsMap = Record<
```
ZeRego, line 759: moved to `packages/common/src/types/field.ts` to avoid cycle dependencies

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -154,14 +153,8 @@ const VisualizationProvider: FC<Props> = ({
...
     const itemsMap = useMemo(() => {
-        if (explore && resultsData)
-            return getItemMap(
-                explore,
-                resultsData.metricQuery.additionalMetrics,
-                resultsData.metricQuery.tableCalculations,
-                resultsData.metricQuery.customDimensions,
-            );
-    }, [explore, resultsData]);
+        return resultsData?.fields;
```
ZeRego, line 156: This is where we stop using all the fields from the explore and only use the fields provided by the BE.

packages/common/src/utils/fields.ts
```diff
@@ -44,15 +41,19 @@ export const getFieldsFromMetricQuery = (
...
-            [getItemId(additionalMetric)]: additionalMetric,
-        }),
-        {},
-    );
+    const additionalMetrics = (metricQuery.additionalMetrics || []).reduce<
+        Record<string, Metric>
+    >((acc, additionalMetric) => {
+        const table = explore.tables[additionalMetric.table];
+        if (table) {
+            const metric = convertAdditionalMetric({
```
rephus, line 49: why do we want to convert ? is this what the explore does ?

ZeRego: yes, the VizProvider converts it atm.

rephus (approved):
Tested and reviewed

Issue #8223, opened 2023-12-06
Deleting a chart still shows the chart on a cached dashboard  


rephus:
### Description
We should invalidate caches for that dashboard / chats 

[Screencast from 30-11-23 11:46:32.webm](https://github.com/lightdash/lightdash/assets/1983672/c3da6607-4ed2-4576-81b3-fb88c4f40ef1)

iMac7:
@rephus is it just the cache? I added a chart to my dashboard then deleted it, and it leaves an empty tile with no data. maybe the dashboard tile is not being deleted when the chart is deleted?

![image](https://github.com/lightdash/lightdash/assets/76876702/c71ba843-9b91-43b0-8782-0c92cf5a3e8a)

PR #8231, opened 2023-12-07
fix: remove deleted chart frontend cache on dashboards

rephus:
Closes: #8223

### Description:

Before:
[Screencast from 07-12-23 08:53:22.webm](https://github.com/lightdash/lightdash/assets/1983672/d380d11b-cb2c-4c1d-aa06-c7440d8519cd)

After:
[Screencast from 07-12-23 09:38:13.webm](https://github.com/lightdash/lightdash/assets/1983672/ad1a125b-e7a2-4a8c-bd6f-b0bc5d667e48)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useSavedQuery.ts
+++ b/packages/frontend/src/hooks/useSavedQuery.ts
@@ -182,27 +182,34 @@ export const useChartVersionRollbackMutation = (
 export const useSavedQueryDeleteMutation = () => {
     const queryClient = useQueryClient();
     const { showToastSuccess, showToastError } = useToaster();
-    return useMutation<undefined, ApiError, string>(deleteSavedQuery, {
-        mutationKey: ['saved_query_create'],
-        onSuccess: async () => {
-            await queryClient.invalidateQueries('spaces');
-            await queryClient.invalidateQueries('space');
-            await queryClient.invalidateQueries('pinned_items');
-            await queryClient.invalidateQueries(
-                'most-popular-and-recently-updated',
-            );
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useSavedQuery.ts
```diff
@@ -182,27 +182,34 @@ export const useChartVersionRollbackMutation = (
...
-        onSuccess: async () => {
-            await queryClient.invalidateQueries('spaces');
-            await queryClient.invalidateQueries('space');
-            await queryClient.invalidateQueries('pinned_items');
-            await queryClient.invalidateQueries(
-                'most-popular-and-recently-updated',
-            );
+    return useMutation<undefined, ApiError, string>(
+        async (data) => {
+            await queryClient.removeQueries(['savedChartResults', data]);
```
IrakliJani, line 187: `invalidateQueries` did not work here? also isn't it better to do in the onSuccess ?

rephus: I've tried with invalidateQueries and it didn't work, see my thread : https://lightdash.slack.com/archives/C02GQKJK84Q/p1701937791955869

I'm doing it here instead of `onSuccess` because I don't have access to the `chartUuid` (response from delete is empty) . 

Let me know if you have a better way of doing this.

IrakliJani: ah that makes sense. so data in that case is UUID

IrakliJani (approved):

Issue #8227, opened 2023-12-06
Version history is not showing old versions of charts


TuringLovesDeathMetal:
### Description

- some charts are showing version history, but only from relatively recent versions. Other charts just aren't showing version history at all (even though there's been changes made). 

Example of no version history being shown: https://www.loom.com/share/adee6718023e4bb8b8496e28b36d9503

Here's an example of a chart that has version history: 
<img width="1535" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/bee717d0-3202-40b9-a7eb-6ee4ce0c75ca">

https://analytics.lightdash.cloud/projects/8a223608-a1e4-426e-adcd-0598ea72b9f8/saved/0440b6ae-ad75-4a5a-a4ec-ba1fc7cd9b15/history

PR #8233, opened 2023-12-07
fix: chat history returns at least 1 old version

rephus:
Closes: #8227

### Description:

After:

With Old version
Screenshot from 2023-12-07 12-13-19

New versions override old version 

Screenshot from 2023-12-07 12-13-48

Screenshot from 2023-12-07 12-14-11

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SavedChartModel.ts
+++ b/packages/backend/src/models/SavedChartModel.ts
@@ -360,6 +360,19 @@ export class SavedChartModel {
         };
     }
 
+    private getLastVersionUuidQuery(chartUuid: string) {
+        return this.database(SavedChartVersionsTableName)
+            .leftJoin(
+                SavedChartsTableName,
+                `${SavedChartVersionsTableName}.saved_query_id`,
+                `${SavedChartsTableName}.saved_query_id`,
+            )
+            .select('saved_queries_version_uuid')
+            .where(`${SavedChartsTableName}.saved_query_uuid`, chartUuid)
+            .limit(1)
+            .orderBy(`${SavedChartVersionsTableName}.created_at`, 'desc');
+    }
+
     private getVersionSummaryQuery() {
         return this.database(SavedChartVersionsTableName)
             .leftJoin(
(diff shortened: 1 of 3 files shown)


packages/backend/src/models/SavedChartModel.ts
```diff
@@ -416,24 +403,31 @@ export class SavedChartModel {
...
-                // get all versions from the last X days + the current version ( in case is older than X days )
-                this.whereRaw(
-                    `${SavedChartVersionsTableName}.created_at >= DATE(current_timestamp - interval '?? days')`,
-                    [daysLimit],
-                ).orWhere(
-                    `${SavedChartVersionsTableName}.saved_queries_version_uuid`,
-                    getLastVersionUuidSubQuery,
-                );
-            })
             .orderBy(`${SavedChartVersionsTableName}.created_at`, 'asc');
```
rephus, line 409: results should be sorted by created_at

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -416,24 +403,31 @@ export class SavedChartModel {
...
 
-        return chartVersions.map(SavedChartModel.convertVersionSummary);
+        return chartVersions
+            .reduce<VersionSummaryRow[]>((acc, version) => {
```
ZeRego, line 412: I'm concerned about performance as we will fetch all the versions from the warehouse. 
What if we keep the existing query and add `if(rows.length < 2) { ... run second query to fetch the last version }`

rephus: I though performance was not going to be an issue because we purge versions periodically (do we ?) 

But yeah, we can do another query

ZeRego: We don't purge versions yet.

rephus: Updated to make  a separate query, also updated tests

ZeRego (approved):

PR #8238, opened 2023-12-07
feat: apply date zoom without time intervals dimensions

ZeRego:
Closes: #8105 

### Description:

Before:
We would find the dimension with the correct time interval in the explore and use it. 
This meant updating the metric query, chart config and results with the new column id.
Now:
We replace the label and SQL of the dimensions used, keeping the existing column id. 

When date dimension has `time_intervals: OFF` 

https://github.com/lightdash/lightdash/assets/9117144/bf2ce3ab-555e-458f-852e-11a21b1a7ef6

When chart has multiple date columns, including the date column for the same granularity.

https://github.com/lightdash/lightdash/assets/9117144/219b705a-0efd-47c7-ae3c-da08eb67500a

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -14,6 +14,7 @@ import {
     countCustomDimensionsInMetricQuery,
     countTotalFilterRules,
     CreateDbtCloudIntegration,
+    createDimensionWithGranularity,
     CreateJob,
     CreateProject,
     CreateProjectMember,
@@ -22,13 +23,11 @@ import {
     DashboardBasicDetails,
     DashboardFilters,
     DateGranularity,
-    dateGranularityToTimeFrameMap,
     DbtProjectType,
     deepEqual,
     DefaultSupportedDbtVersion,
     DimensionType,
     Explore,
-    ExploreCompiler,
     ExploreError,
     fieldId as getFieldId,
     FilterableField,
@@ -42,7 +41,6 @@ import {
     getDimensions,
     getFields,
     getItemId,
-    getItemMap,
     getMetrics,
     hasIntersection,
     isDateItem,
@@ -66,6 +64,7 @@ import {
     ProjectMemberProfile,
     ProjectMemberRole,
     ProjectType,
+    replaceDimensionInExplore,
     RequestMethod,
     ResultRow,
     SavedChartsInfoForDashboardAvailableFilters,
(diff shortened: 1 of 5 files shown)


packages/common/src/compiler/translator.ts
```diff
@@ -300,16 +300,17 @@ export const convertTable = (
...
-                            index,
-                            adapterType,
-                            model,
-                            tableLabel,
-                            column,
-                            undefined,
-                            interval,
-                            startOfWeek,
-                        ),
+                        [`${column.name}_${interval.toLowerCase()}`]:
```
ZeRego, line 303: Unnoticed bug where the key of time interval dimensions wouldn't match the `name` property.

`order_date_DAY` vs `order_date_day`

Screenshot 2023-12-07 at 13 26 08

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1341,12 +1268,12 @@ export class ProjectService {
...
                                                   workerData: {
                                                       rows,
-                                                      itemMap,
+                                                      fields,
                                                   },
                                               },
                                           ),
                                       )
-                                    : formatRows(rows, itemMap);
+                                    : formatRows(rows, fields);
```
ZeRego, line 1276: Use fields from `runMetricQuery()` so formatting is correct.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -733,19 +734,79 @@ export class ProjectService {
...
+
     private static async _compileQuery(
         metricQuery: MetricQuery,
         explore: Explore,
         warehouseClient: WarehouseClient,
         userAttributes: UserAttributeValueMap,
+        granularity?: DateGranularity,
     ): Promise<CompiledQuery> {
-        const compiledMetricQuery = compileMetricQuery({
+        const exploreWithOverride = ProjectService.updateExploreWithGranularity(
```
ZeRego, line 797: Before compiling the SQL we update the explore to have the dimension with the new granularity

packages/common/src/compiler/translator.ts
rephus: will this affect existing explores ? 
So if someone is trying to filter use date zoom on a explore that hasn't been updated, it will not work, right ?

rephus (approved):
Code looks good, nice job. I also tested this on render, tests are also happy.  
I only have a question about backwards compatibility, but merge if you think this is safe.

Issue #6240, opened 2023-07-11
I want to be able to specify which snowflake warehouse my Lightdash models are being executed in


TuringLovesDeathMetal:
### Scope
Snowflake DWH connection only

### Description
It would be really great if we could (optionally) specify what snowflake warehouse Lightdash should use when fetching the data from the model.

**Why**?
In our default project we have several datasets for various things. Raw tracking data, clean product session/page views, AB testing focus, and etc.
Volumes of those are VERY different. From tens of billions rows to several millions. Query usage level is not the same as well, we can't control really how often what source is being used. It's pretty unpredictable.
Right now all those models will be queried by using the warehouse configured in connection.
It's not ideal as if we use big sized warehouse (let's say XL) to ensure querying big datasets is fast enough then we're burning the money on querying smaller datasets. As smaller ones would be probably totally cool to query with M sized warehouse.
If we set it smaller, then users of bigger datasets are waiting and cluttering the warehouse (there's a limit how many queries you can run on one warehouse... and we're not on enterprise snowflake edition so auto horizontal scaling is not working for us ).

### Potential solution
There's already configuration for this in dbt, we're just not reading it in Lightdash. We should be reading this config in Lightdash: https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-virtual-warehouses

<img width="598" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/5ed5b696-bb40-4b75-88ef-63b71cd3253a">

jw-aleksandr-zolotukhin:
what a great idea :-)

owlas:
it looks like we don't need to add this to the `meta` since it's already a valid dbt configuration option

TuringLovesDeathMetal:
@jw-aleksandr-zolotukhin just following up from Oliver's comment. 

There's this configuration in dbt here: https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-virtual-warehouses

Are you using this already? Is it just that Lightdash isn't "reading" this?

jw-aleksandr-zolotukhin:
@TuringLovesDeathMetal 
yes. reading that native dbt setting would be ideal.

PR #8255, opened 2023-12-08
feat: snowflake warehouse connection

rephus:
Closes: #6240

### Description:

https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-virtual-warehouses

### How to use snowflake warehouse per model

- add it to the model (via config or yml) 

Screenshot from 2023-12-08 16-12-48

- compile and deploy explore

Screenshot from 2023-12-08 15-46-55

- Run query with warehouse 
Screenshot from 2023-12-08 16-21-29

- If warehouse doesn't exist we get this error 

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -212,7 +212,7 @@ export class ProjectService {
 
     private async _getWarehouseClient(
         projectUuid: string,
-        warehouse?: string,
+        snowflakeVirtualWarehouse?: string,
     ): Promise<{
         warehouseClient: WarehouseClient;
         sshTunnel: SshTunnel<CreateWarehouseCredentials>;
@@ -226,7 +226,7 @@ export class ProjectService {
         const sshTunnel = new SshTunnel(credentials);
         const warehouseSshCredentials = await sshTunnel.connect();
 
-        const cacheKey = `${projectUuid}${warehouse || ''}`;
+        const cacheKey = `${projectUuid}${snowflakeVirtualWarehouse || ''}`;
         // Check cache for existing client (always false if ssh tunnel was connected)
         const existingClient = this.warehouseClients[cacheKey] as
             | typeof this.warehouseClients[string]
@@ -243,7 +243,8 @@ export class ProjectService {
             credentials.type === WarehouseTypes.SNOWFLAKE
                 ? {
                       ...warehouseSshCredentials,
-                      warehouse: warehouse || credentials.warehouse,
+                      warehouse:
+                          snowflakeVirtualWarehouse || credentials.warehouse,
                   }
                 : credentials;
         const client = this.projectModel.getWarehouseClientFromCredentials(
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -210,7 +210,10 @@ export class ProjectService {
         return args;
     }
 
-    private async _getWarehouseClient(projectUuid: string): Promise<{
+    private async _getWarehouseClient(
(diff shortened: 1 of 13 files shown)


rephus:
Question, because I added warehouse to the explore, it is returned on the api, should I `omit` this field in the response ? 

Screenshot from 2023-12-08 15-52-38

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1323,6 +1323,7 @@ export class ProjectService {
         metricQuery,
         queryTags,
         invalidateCache,
+        warehouse,
```
owlas, line 1326: this naming is really confusing because the function is called `getResultsFromCacheOrWarehouse` but `warehouse` has nothing to do with the `warehouse` in the name.

packages/common/src/types/warehouse.ts
```diff
@@ -27,6 +27,7 @@ export interface WarehouseClient {
     runQuery(
         sql: string,
         tags?: Record<string, string>,
+        warehouse?: string,
```
owlas, line 30: This is only relevant to the snowflake warehouse so it's not a very accurate typing on the generic `runQuery` operation. 

Since we create a WarehouseClient on each query, can't we just override the credentials passed into the constructor when we `new WarehouseClient({credentials})`

owlas (commented):
This makes the typing more error-prone and not as obvious to me

packages/common/src/types/warehouse.ts
rephus: Updated :+1:

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -223,8 +226,9 @@ export class ProjectService {
         const sshTunnel = new SshTunnel(credentials);
         const warehouseSshCredentials = await sshTunnel.connect();
 
+        const cacheKey = `${projectUuid}_${warehouse || ''}`;
```
owlas: nice

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -210,7 +210,10 @@ export class ProjectService {
         return args;
     }
 
-    private async _getWarehouseClient(projectUuid: string): Promise<{
+    private async _getWarehouseClient(
+        projectUuid: string,
+        warehouse?: string,
```
owlas, line 215: the name is still confusing!

rephus: What name do you suggest ?

owlas: `snowflakeVirtualWarehouse`

owlas (approved):
approved after name change

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -235,10 +239,18 @@ export class ProjectService {
...
         }
         // otherwise create a new client and cache for future use
+        const credentialsWithWarehouse =
+            credentials.type === WarehouseTypes.SNOWFLAKE
+                ? {
+                      ...warehouseSshCredentials,
+                      warehouse:
+                          snowflakeVirtualWarehouse || credentials.warehouse,
+                  }
+                : credentials;
```
owlas, line 249: This is a bug, it should be `warehouseSshCredentials` which contains the credentials including the proxied host name and port for any active SSH connections.

owlas: Before, if it wasn't snowflake it was returning `credentials` (i.e. the raw credentials) not `warehouseSshCredentials` (the proxied credentials)

PR #8277, opened 2023-12-12
refactor: table config form to mantine

ZeRego:
Closes: 

### Description:

Before:
Screenshot 2023-12-12 at 10 42 50

After:
Screenshot 2023-12-12 at 12 18 34
Screenshot 2023-12-12 at 12 18 55
Screenshot 2023-12-12 at 12 18 45

### Reviewer actions


packages/frontend/src/components/ProjectTablesConfiguration/ProjectTablesConfiguration.tsx
```diff
@@ -156,119 +158,124 @@ const ProjectTablesConfiguration: FC<{
...
+                                            Show models with any of these tags{' '}
+                                            <DocumentationHelpButton href="https://docs.getdbt.com/reference/resource-configs/tags#examples" />
+                                        </>
+                                    }
+                                    description="Write a list of tags you want to include, separated
+                                    by commas."
+                                />
+                                {form.values.type ===
+                                    TableSelectionType.WITH_TAGS && (
+                                    <MultiSelect
```
magnew, line 223: At the moment if there aren't any tags and you select this, it's marked as required, but you can't input anything there. Maybe if there aren't any it should be disabled? The old behavior is about the same, but its still bad.

Or do we expect people to be able to add tags here that aren't already defined? In which case maybe we need the tag input.

magnew (commented):
Tested, looks good. I made a UX suggestion, but it doesn't have to block.

packages/frontend/src/components/ProjectTablesConfiguration/ProjectTablesConfiguration.tsx
ZeRego: Mantine tooltip doesn't work when wrapping a disabled input 😢 Used the error message instead, even though it doesn't stop them from saving but that seems minor. 
Screenshot 2023-12-12 at 17 11 33

IrakliJani: you can put tag input inside a `div` (tooltip > div > taginput)
btw, it's the same for all elements in react. if input is disabled it does not trigger any event from the pointer

IrakliJani:
I've re-seeded the db and re-running e2e tests now

IrakliJani (commented):
code looks good, testing on render now

IrakliJani (approved):
looks great! 🚢

Issue #8267, opened 2023-12-11
Scheduled deliveries with filters are showing incorrect results 


TuringLovesDeathMetal:
### Description

We have about 17 scheduled deliveries with different filters for the same dashboard. THey are all sent at the same time. But, some of the scheduled deliveries are showing the wrong results in the delivered reports. 

What could be going wrong? 

Link to slack thread: https://lightdash.slack.com/archives/C04KD95TF0U/p1702313137951689?thread_ts=1702312906.461489&cid=C04KD95TF0U

PR #8281, opened 2023-12-12
fix: recheck if scheduler filter should be disabled

joaoviana:
Closes: #8267

### Description:

To reproduce: 
1. Add **saved filter**, type string for example: payment method is bank transfer
2. On scheduler
3. Override payment method filter, but first: remove `bank transfer`, and then add `card` -> the `disabled: true` attribute was added when removing the initial value, and then wouldn't be inverted when readding values. 

This PR now checks if that should be actually `disabled`. 

Also, renames some utils to make them more generic so it's easier to share across different parts of the app 👍 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
@@ -210,8 +210,6 @@ const updateFilters = (
     originalFilter: DashboardFilterRule,
     schedulerFilters: SchedulerFilterRule[] | undefined,
 ): SchedulerFilterRule[] | undefined => {
-    console.log('updateFilters');
-
     if (
         schedulerFilters &&
         // Check if filters are the same, regardless of disabled state (accepts any value)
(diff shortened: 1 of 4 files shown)


packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -230,7 +233,12 @@ const updateFilters = (
     if (hasFilterChanged(filterToCompareAgainst, schedulerFilter)) {
         if (schedulerFilters && isExistingFilter) {
             return schedulerFilters.map((f) =>
-                f.id === schedulerFilter.id ? schedulerFilter : f,
+                f.id === schedulerFilter.id
+                    ? {
+                          ...schedulerFilter,
+                          disabled: !isFilterEnabled(schedulerFilter),
+                      }
+                    : f,
```
joaoviana, line 241: this is bit!

magnew (approved):
Code looks good. I tested this on render and it seems to work. Note that I could only try email there, but from the code I expect if one path works they all do. Nice quick fix!

Issue #8334, opened 2023-12-15
User completion modal  not closing


joaoviana:
### Description

PR ref: https://github.com/lightdash/lightdash/pull/8290/files

### To Reproduce

error: 
```
data
: 
{}
message
: 
"You don't have access to this resource or action"
name
: 
"ForbiddenError"
statusCode
: 
403
```

PR #8335, opened 2023-12-15
fix: only send org name if it didn't exist

joaoviana:
Closes: #8334 

### Description:

Only send `organizationName` if it didn't exist in the first place. 
User has to set it if there's no org set up 

### Reviewer actions


rephus (approved):
LGTM

PR #8351, opened 2023-12-18
feat: space modals in mantine

IrakliJani:
Closes: 

### Reviewer actions


packages/frontend/src/components/ReactHookForm/RadioGroup.tsx
magnew, line 1: 🎉

packages/frontend/src/components/common/SpaceActionModal/CreateSpaceModalContent.tsx
```diff
@@ -38,26 +28,35 @@ const renderUser = (user: {
...
-}) => {
-    return (
-        <FlexWrapper key={user.userUuid}>
-            <UserCircle>
-                {user.firstName.substr(0, 1) + user.lastName.substr(0, 1)}
-            </UserCircle>
+}) => (
+    <Group spacing="sm" position="apart" noWrap>
+        <Group>
+            <Avatar radius="xl" tt="uppercase" color="blue">
```
magnew, line 34: Are these looking blue at all to you? I see them as grey even though they are set to blue here.

magnew (approved):
Looks great! Tested and everything seems to work. 

One minor issue I found in test, not really a bug but a little confusing: when are creating a space and don't set a name, the create button on the permissions section is active, but does nothing. We could either disable 'Continue' if there is no name, or disable create.

I'm fine with doing this in a follow up if you want. 
 
Here's what I mean

Kapture 2023-12-19 at 15 16 02

packages/frontend/src/components/common/SpaceActionModal/CreateSpaceModalContent.tsx
IrakliJani: can you take a screenshot? there are few places where I had to update colors

magnew: It seems like everywhere I see these avatars they are gray. Whatever their access.

Screenshot 2023-12-19 at 16 02 02

IrakliJani: must be render not picking up the changes. triggering manual sync!

PR #8356, opened 2023-12-19
chore: remove blueprint from user activity page

magnew:
### Description:

Removes blueprint from user activities page. ~~This page uses a lot of styled components, which we could also update, but for now I focussed on blueprint.~~

Also removes styled components.  

### Reviewer actions


packages/frontend/src/pages/UserActivity.styles.ts
```diff
@@ -20,7 +19,11 @@ export const Container = styled.div`
         'table-dashboard-views table-dashboard-views table-chart-views table-chart-views';
 `;
 
-export const ActivityCard = styled(Card)<{ grid: string }>`
+export const ActivityCard = styled.div<{ grid: string }>`
+    background-color: #fff;
+    border-radius: 2px;
+    padding: 20px;
+    border: 1px solid #dee2e6;
```
magnew, line 26: Styles that used to come from the card

packages/frontend/src/pages/UserActivity.styles.ts
```diff
@@ -44,12 +47,12 @@ export const BigNumber = styled.div`
     font-size: 4em;
     line-height: 1.196em;
     font-weight: 500;
-    color: ${Colors.DARK_GRAY4};
+    color: #383e47;
```
magnew, line 50: From blueprint colors. I figured it was ok to hardcode these for now.

packages/frontend/src/pages/UserActivity.styles.ts
IrakliJani: we can use mantine's Card component here. but if its too much work then ignore

packages/frontend/src/pages/UserActivity.styles.ts
IrakliJani: can we use mantine equivalent of it?

packages/frontend/src/pages/UserActivity.styles.ts
```diff
@@ -44,12 +47,12 @@ export const BigNumber = styled.div`
...
 `;
 
 export const BigNumberLabel = styled.h2`
     text-align: center;
-    color: ${Colors.GRAY3};
```
IrakliJani, line 52: same here

IrakliJani (approved):
looks good to me !

there is one minor one but I think this issue already existed in prod.
CleanShot 2023-12-20 at 16 34 56@2x

magnew:
> looks good to me !
> 
> there is one minor one but I think this issue already existed in prod. CleanShot 2023-12-20 at 16 34 56@2x

Yea, that could be improved. I didn't change it though.

PR #8372, opened 2023-12-20
feat: table without blueprint

IrakliJani:
Closes: 

### Description:

CleanShot 2023-12-20 at 19 51 30@2x
CleanShot 2023-12-20 at 19 53 24
CleanShot 2023-12-20 at 20 08 45@2x
CleanShot 2023-12-20 at 20 08 51@2x

### Reviewer actions


packages/frontend/src/components/common/Table/Table.styles.ts
```diff
@@ -1,9 +1,10 @@
-import { Colors, HTMLTable } from '@blueprintjs/core';
+import { DEFAULT_THEME } from '@mantine/core';
```
magnew, line 1: Hm, I guess I could have done this instead of hard coding them like I did ¯\_(ツ)_/¯

packages/frontend/src/components/common/Table/Table.styles.ts
```diff
@@ -39,69 +40,117 @@ export const TableContainer = styled.div<TableContainerProps>`
...
+
+    tbody tr:hover td {
+        background-color: rgba(143, 153, 168, 0.3);
+        cursor: pointer;
+    }
+    tbody tr:active td {
+        background-color: rgba(143, 153, 168, 0.35);
+    }
+
+    /* FIXME: everything above this line is copied from blueprint's table css */
```
magnew, line 101: Do you know what this fixme will require? Like, will we move to a new table or just keep this css forever?

magnew (approved):
Nice work! I tried a lot of things and don't see any issues.

packages/frontend/src/components/common/Table/Table.styles.ts
IrakliJani: yeah, you can ignore. we should chase down all `#` and `rgb` in the codebase and migrate to mantine at some point

packages/frontend/src/components/common/Table/Table.styles.ts
IrakliJani: @magnew 
I think we should migrate to `LightTable` (component that Pivot tables use)
should be a fun lightsmash friday task

Issue #8373, opened 2023-12-20
Improve csrf protection


ZeRego:
### Description

More of this: https://snyk.io/blog/how-to-protect-node-js-apps-from-csrf-attacks/
TLDR:
Cross-site request forgery, or CSRF/XSRF, is an attack that relies on the user's privileges by hijacking their session. This strategy allows an attacker to circumvent our security by essentially deceiving the user into submitting a malicious request on behalf of the attacker. 

Update cookies to have `samesite="Strict"`

**Lax** allows the cookie to be sent on some cross-site requests, whereas **Strict** never allows the cookie to be sent on a cross-site request.

ZeRego:
Reverting existing solution. New approach needs to be compatible with azure login. More details in this bug: https://github.com/lightdash/lightdash/issues/8476

PR #8374, opened 2023-12-20
feat: strict same site cookies

ZeRego:
Closes: #8373 

### Description:

Tested:
✅  scheduled delivery

### Reviewer actions

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -129,7 +129,7 @@ app.use(
             maxAge: (lightdashConfig.cookiesMaxAgeHours || 24) * 60 * 60 * 1000, // in ms
             secure: lightdashConfig.secureCookies,
             httpOnly: true,
-            sameSite: 'lax',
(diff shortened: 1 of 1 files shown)


packages/backend/src/index.ts
```diff
@@ -129,7 +129,7 @@ app.use(
             maxAge: (lightdashConfig.cookiesMaxAgeHours || 24) * 60 * 60 * 1000, // in ms
             secure: lightdashConfig.secureCookies,
             httpOnly: true,
-            sameSite: 'lax',
+            sameSite: 'strict',
```
rephus, line 132: This change means now all embeds will be unauthenticated even if they are logged in on lightdash, right ?

rephus (approved):
Tested scheduler too, LGTM

packages/backend/src/index.ts
ZeRego: Since they have the same domain, cookies will still be sent.

Issue #8066, opened 2023-11-23
dbt model names are not always unique


owlas:
### Description

Lightdash assumes that dbt model names are unique. This was true until this PR was merged: https://github.com/dbt-labs/dbt-core/pull/7374 - with this change, Project A can have model `users` and **also** have a dependency on Project B with a model called `users`

The resulting manifest will have two models with the name `users` but they'll have different unique ids `model.project_a.users` and `model.project_b.users`

TuringLovesDeathMetal:
There is also versioning models in dbt which causes this issue! https://docs.getdbt.com/docs/collaborate/govern/model-versions#how-to-create-a-new-version-of-a-model

TuringLovesDeathMetal:
We should potentially make the unique ID a different thing than the name (before, teh name used to be unique, but now it's not anymore - we should update the model's unique ID to be something other than the name. E.g. project.model_name)

ZeRego:
In this [PR](https://github.com/lightdash/lightdash/pull/8390) I'm avoiding this error by discarding/ignoring explores with duplicate names. It will only use the first explore for each name.

This fix is focused on not throwing an error to unblock users that deploy their project via the UI which deploys all their models. We can open a new ticket to support multiple explores with the same name.

PR #8390, opened 2023-12-22
fix: discard explores with duplicate name

ZeRego:
Closes: #8066 

### Description:

In this PR I'm avoiding this error by discarding/ignoring models with duplicate names. It will only use the first model for each name.

This fix is focused on not throwing an error to unblock users that deploy their project via the UI which deploys all their models. We can open a new ticket to support multiple models with the same name but that hasn't been requested by users.

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.test.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.test.ts
@@ -1,12 +1,18 @@
 import knex from 'knex';
 import { getTracker, MockClient, RawQuery, Tracker } from 'knex-mock-client';
 import { FunctionQueryMatcher } from 'knex-mock-client/types/mock-client';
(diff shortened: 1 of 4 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.test.ts
```diff
@@ -24,7 +30,7 @@ function queryMatcher(
         sql.includes(tableName) &&
         params.length === bindings.length &&
         params.reduce(
-            (valid, arg, index) => valid && bindings[index] === arg,
+            (valid, arg, index) => valid && isEqual(bindings[index], arg),
```
ZeRego, line 33: use `isEqual` so it can match arrays and objects.

magnew (approved):
LGTM. This also seems like a reasonable constraint to me. Not sure if we need any other buy-in.

Issue #8195, opened 2023-12-04
Inputs appear misaligned when there is an error on any field in the 'Project Invite' modal


PriPatel:
To reproduce:
Enter an invalid email address in either the organization access or project access modals.

<img width="632" alt="Screenshot 2023-12-04 at 14 46 06" src="https://github.com/lightdash/lightdash/assets/12776522/5b378a64-4972-44b8-88b4-8e89083acbef">

- [ ] The fields should be aligned when there is a default state or an error state

alfredlouisw8:
hi can i work on this issue?

raheelam:
@PriPatel hello can i work on this issue?

PriPatel:
@alfredlouisw8 , go for it !

PR #8409, opened 2023-12-26
fix: sort alignment on email error message

jogeshwar01:
Closes: #8195

### Description:

The issue seemed to be because of the flex-end here.
https://github.com/lightdash/lightdash/blob/78633f90cf541f84b81480c65913f2eff328295d/packages/frontend/src/components/UserSettings/InvitesModal/index.tsx#L76

When we have an email error message, it increases the height of `TextInput`. Subsequently, the height of the Group also increases and it aligns itself to the end of it vertically. So a conditional bottom padding to the Group fixes this.

_Note: This may cause an extra padding at the bottom if we don't have all the inputs in the same line (but this shouldn't happen ideally)._ 

We don't need this fix in the `ProjectAccess` section as we are using a `Select` for the email there.

Before:
image

After:
Without Error -
image

With Error - 
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/InvitesModal/index.tsx
+++ b/packages/frontend/src/components/UserSettings/InvitesModal/index.tsx
@@ -73,7 +73,10 @@ const InvitesModal: FC<{
                         handleSubmit(values),
                     )}
                 >
-                    <Group align="flex-end" spacing="xs">
+                    <Group
+                        spacing="xs"
+                        align={form.errors.email ? 'center' : 'end'}
+                    >
                         <TextInput
                             name="email"
                             label="Enter user email address"
@@ -83,7 +86,10 @@ const InvitesModal: FC<{
                             w="43%"
                             {...form.getInputProps('email')}
                         />
-                        <Group spacing="xs">
+                        <Group
+                            spacing="xs"
+                            align={form.errors.email ? 'center' : 'end'}
+                        >
                             {user.data?.ability?.can(
                                 'manage',
                                 'Organization',


packages/frontend/src/components/UserSettings/InvitesModal/index.tsx
```diff
@@ -83,7 +83,12 @@ const InvitesModal: FC<{
                             w="43%"
                             {...form.getInputProps('email')}
                         />
-                        <Group spacing="xs">
+                        <Group
+                            spacing="xs"
+                            style={{
```
joaoviana, line 88:
```suggestion
                            pb={form.errors.email ? 'sm' : 0}
```
how does it look @jogeshwar01 ?

jogeshwar01: @joaoviana Thank you! This is definitely a cleaner syntax. But I don't think we can use the default spacing options of mantine for the value.
These have pixel values of 12px(0.75rem), 16px (1rem), 20px(1.25rem) but we need them to be 18px.

image

jogeshwar01: We can use calc and var from Mantine if we want to use these options. Basically multiplying the sm by 1.5 to get 18px.
Like this - `pb={form.errors.email ? "calc(var(--mantine-spacing-sm)*1.5)" : 0}`

joaoviana: Thanks for the report @jogeshwar01! I've also had a closer look on this, and we actually don't need any specific numbers to address this issue. 

If there's an error, center the containers. If there isn't, align them to the end.

Here is the diff that you can implement on this PR, which will then avoid the `padding` approach and solely rely on flex alignment 👍 

image

joaoviana (approved):
nice @jogeshwar01 ! Will merge once build passes 👍 thanks for contributing!

PR #8415, opened 2023-12-27
chore: delete the apt-get lists after installing dependencies

UlisesGascon:
### Description:

This change ensures the cleanup of the apt cache and removal of /var/lib/apt/lists to reduce the image size. Since we already execute apt-get update when installing new dependencies, the cache remains consistently fresh.

For additional reference, [see](https://docs.docker.com/develop/develop-images/instructions/#apt-get)

### Reviewer actions

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -16,16 +16,21 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     software-properties-common \
     unzip \
     wget \
-    && apt-get clean
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


rephus:
I tested this locally, the difference between the 2 images are minimal (~20MB) 
image

But I think we should do it anyway, would you mind resolving the conflict and applying the change to `dockerfile-prs` ?

UlisesGascon:
Thanks @rephus for consider this PR change even if is a small space safe 😉.

rephus (approved):

PR #8417, opened 2023-12-27
feat: add healthcheck to dockerfile

UlisesGascon:
### Description:

This change will introduce an additional statement in the Dockerfile to implement a health check, ensuring the proper functioning of the backend service. I wasn't entirely certain if this was something we wanted to include, and I'm open to discussing the details.

I'd love to hear your thoughts and suggestions on this! 😄

**Current Proposal**

The idea is to use curl to check the response from `http://localhost:8080/api/v1/health`. If the request is unsuccessful, we `EXIT` with a `1` error code, prompting Docker to consider the container as unhealthy.

We can also consider additional options, such as `interval`, `timeout`, and `retries`. With this configuration, we maintain performance by limiting requests to every 30 seconds and ensuring that the service is marked as broken after 5 retries.

### Reviewer actions

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -164,5 +164,10 @@ ENV LIGHTDASH_CONFIG_FILE /usr/app/lightdash.yml
 COPY ./docker/prod-entrypoint.sh /usr/bin/prod-entrypoint.sh
 
 EXPOSE 8080
+
+# Healthcheck for backend service
+HEALTHCHECK --interval=30s --timeout=3s --retries=5 \
+  CMD curl -f http://localhost/:8080 || exit 1
+
 ENTRYPOINT ["/usr/bin/prod-entrypoint.sh"]
 CMD ["yarn", "workspace", "backend", "start"]
--- a/dockerfile
+++ b/dockerfile
@@ -167,7 +167,7 @@ EXPOSE 8080
 
 # Healthcheck for backend service
 HEALTHCHECK --interval=30s --timeout=3s --retries=5 \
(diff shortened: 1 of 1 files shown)


dockerfile
```diff
@@ -164,5 +164,10 @@ ENV LIGHTDASH_CONFIG_FILE /usr/app/lightdash.yml
 COPY ./docker/prod-entrypoint.sh /usr/bin/prod-entrypoint.sh
 
 EXPOSE 8080
+
+# Healthcheck for backend service
+HEALTHCHECK --interval=30s --timeout=3s --retries=5 \
```
rephus, line 169: Great suggestion, for reference, we use the following settings for livenessProbe on k8s , should we match them ? :thinking:  https://github.com/lightdash/helm-charts/blob/main/charts/lightdash/values.yaml#L90

```
  initialDelaySeconds: 30
  timeoutSeconds: 60
  periodSeconds: 30
  ```




At least we should use `--start-period` to make sure we don't get a false positive from /health when for example a migration is running. 

Since Kubernetes 1.8, the Docker HEALTHCHECK has been [disabled explicitly](https://github.com/kubernetes/kubernetes/pull/50796) in Kubernetes. so we don't need to make any changes on cloud. This will only affect local development.

joaoviana:
Closing due to inactivity, will re-open once question: https://github.com/lightdash/lightdash/pull/8417#discussion_r1441518231 is addressed 👍

Issue #8412, opened 2023-12-27
Groups can be created


magnew:
### Description

UI for group creation

PR #8422, opened 2023-12-27
chore:  create group dialog

magnew:
Closes: #8412 

### Description:

Adds the ability to create groups from the UI. For now they can only be named, next will be the ability to add users to them.  

Screenshot 2023-12-27 at 17 37 43

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/CreateGroupModal/index.tsx
+++ b/packages/frontend/src/components/UserSettings/CreateGroupModal/index.tsx
@@ -1,4 +1,12 @@
-import { Button, Group, Modal, Stack, TextInput, Title } from '@mantine/core';
+import {
+    Button,
+    Group,
+    Modal,
+    ModalProps,
+    Stack,
+    TextInput,
+    Title,
+} from '@mantine/core';
 import { useForm } from '@mantine/form';
 import { IconUsersGroup } from '@tabler/icons-react';
 import React, { FC } from 'react';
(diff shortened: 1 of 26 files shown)


packages/frontend/src/components/UserSettings/CreateGroupModal/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useForm } from '@mantine/form';
+import { IconUsersGroup } from '@tabler/icons-react';
+import React, { FC } from 'react';
+import { useGroupCreateMutation } from '../../../hooks/useOrganizationGroups';
+import { useApp } from '../../../providers/AppProvider';
+import MantineIcon from '../../common/MantineIcon';
+
+const CreateGroupModal: FC<{
+    opened: boolean;
+    onClose: () => void;
```
joaoviana, line 11: I would pick from `ModalProps` 👍 you can just do: `FC<ModalProps>`

packages/frontend/src/components/UserSettings/CreateGroupModal/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+}> = ({ opened, onClose }) => {
+    const form = useForm<{ name: string }>({
+        initialValues: {
+            name: '',
+        },
+        validate: {
+            name: (value: string) =>
+                value.length ? null : 'Group name is required',
```
joaoviana, line 19: could we also `.trim()` here?

packages/frontend/src/hooks/useOrganizationGroups.ts
```diff
@@ -18,3 +19,35 @@ export const useOrganizationGroups = () => {
         onError: (result) => setErrorResponse(result),
     });
 };
+
+const createGroupQuery = async (data: { name: string }) =>
```
joaoviana, line 23: not a blocker, but the type `{name: string}` could be added to the `ApiResults` in common so it can reused here and also on `useMutation<Group, ApiError, { name: string }>(`

joaoviana (approved):
Looks great!

packages/frontend/src/hooks/useOrganizationGroups.ts
magnew: Yea, i should have said, but I think we will redo these types a bit so I didn't make them very fleshed out. If it's cool with you, I am going to skip this for now and add shared types in a follow up.

packages/frontend/src/components/UserSettings/CreateGroupModal/index.tsx
magnew: Good call.

packages/frontend/src/components/UserSettings/CreateGroupModal/index.tsx
magnew: 👍

PR #8426, opened 2023-12-28
refactor: cleanup users and groups

magnew:
### Description:

Cleans up the user and groups files and fixes types. There is no new logic here, just moving things around. Specifically:

- Rename UserManagementPanel -> UsersAndGroupsPanel (for clarity)
- Move InvitesModal/index.tsx -> UsersAndGroupsPanel/InvitesModal.tsx (only used here)
- Move CreateGroupModal/index.tsx -> UsersAndGroupsPanel/CreateGroupModal.tsx (only used here)
- Update types to use `CreateGroup` Type instead of `{name: string}`
- Update ability checks to check `Group` instead of `Organization`
- Split `UsersAndGroups` tabs into `UsersView` and `GroupsView` (these are both about to get more complex)

### Reviewer actions


packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
magnew, line 1: The diff on this file got a little messed up. This is a new view, not the rename. Should be clearish though.

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
```diff
@@ -266,184 +259,88 @@ const UserListItem: FC<{
...
     const [search, setSearch] = useState('');
+
     const { data: organizationUsers, isLoading: isLoadingUsers } =
         useOrganizationUsers(search);
-    const { data: groups, isLoading: isLoadingGroups } =
-        useOrganizationGroups();
 
-    if (user.data?.ability.cannot('view', 'OrganizationMemberProfile')) {
-        return <ForbiddenPanel />;
+    if (isLoadingUsers) {
```
joaoviana, line 272: with the removal of this check: `user.data?.ability.cannot('view', 'OrganizationMemberProfile')` do we need to add it somewhere else?

joaoviana (commented):
looks good, just one q!

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
magnew: This is a silly artifact of the bad diff. This now happens in the parent file.

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const UsersAndGroupsPanel: FC = () => {
+    // TODO: this is a feature flag while we are building groups.
+    // Remove this when groups are ready to be released.
+    const groupManagementEnabled = useFeatureFlagEnabled('group-management');
+    const { user } = useApp();
+
+    if (user.data?.ability.cannot('view', 'OrganizationMemberProfile')) {
+        return <ForbiddenPanel />;
+    }
```
magnew, line 21: @joaoviana it's here now.

joaoviana: oh, I see!

joaoviana (approved):

Issue #8276, opened 2023-12-12
Scheduled delivery dashboard is being formatted incorrectly between two scheduled deliveries


TuringLovesDeathMetal:
### Description

Getting some inconsistent layouts in scheduled emails. With no changes, different emails show different layouts. Do you know what might be causing this?

Here's a link to the slack thread with some screenshots: https://lightdash.slack.com/archives/C04DLEN7BJS/p1702310726983879

joaoviana:
Relevant issue: https://github.com/react-grid-layout/react-grid-layout/issues/1787

joaoviana:
https://github.com/lightdash/lightdash/pull/8478
https://github.com/lightdash/lightdash/pull/8477
https://github.com/lightdash/lightdash/pull/8428
https://github.com/lightdash/lightdash/pull/8389

All the above were to introduce `previews` and custom width for Dashboard exports and scheduled deliveries _only_ for dashboards

PR #8428, opened 2023-12-28
feat: add preview choice to dashboard scheduled deliveries

joaoviana:
Relates to: #8276

### Description:

Add `<SchedulerPreview />` component
Add new column to `scheduler` DB table: `custom_viewport_width`
Allow users to preview their scheduled deliveries for dashboard in image format in 3 sizes
Allow users to set their selection (if `Medium`, then that's the default value) -> Subsequent deliveries will respect the width selection
Create shared component `PreviewAndCustomizeScreenshot` that is shared by the Dashboard Export feature and the Scheduler Preview feature

> [!NOTE]
> This doesn't work yet with `Send Now` feature. 

Demo: 

https://github.com/lightdash/lightdash/assets/7611706/e0aa440b-14aa-4462-b762-e473964d4848

Scheduled delivery with Large setting: 
image

Scheduled delivery with Small setting (see the 2nd markdown tile from above ^ being moved randomly to another row further down)
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
@@ -50,6 +50,7 @@ import SlackSvg from '../../../svgs/slack.svg?react';
 import { isInvalidCronExpression } from '../../../utils/fieldValidators';
 import SchedulerFilters from './SchedulerFilters';
 import SchedulersModalFooter from './SchedulerModalFooter';
+import { SchedulerPreview } from './SchedulerPreview';
 
 export enum Limit {
     TABLE = 'table',
@@ -83,6 +84,7 @@ const DEFAULT_VALUES = {
     emailTargets: [] as string[],
     slackTargets: [] as string[],
     filters: undefined,
+    customViewportWidth: undefined,
 };
 
 const getFormValuesFromScheduler = (schedulerData: SchedulerAndTargets) => {
@@ -128,6 +130,7 @@ const getFormValuesFromScheduler = (schedulerData: SchedulerAndTargets) => {
         slackTargets: slackTargets,
         ...(isDashboardScheduler(schedulerData) && {
             filters: schedulerData.filters,
+            customViewportWidth: schedulerData.customViewportWidth,
         }),
     };
 };
@@ -257,6 +260,7 @@ const SchedulerForm: FC<Props> = ({
                 targets,
                 ...(resource?.type === 'dashboard' && {
                     filters: values.filters,
+                    customViewportWidth: values.customViewportWidth,
                 }),
             };
         },
@@ -343,6 +347,7 @@ const SchedulerForm: FC<Props> = ({
                         <Tabs.Tab value="filters">Filters</Tabs.Tab>
                     ) : null}
                     <Tabs.Tab value="customization">Customization</Tabs.Tab>
+                    <Tabs.Tab value="preview">Preview</Tabs.Tab>
                 </Tabs.List>
 
                 <Tabs.Panel value="setup" mt="md">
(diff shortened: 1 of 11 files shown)


packages/frontend/src/features/scheduler/components/SchedulerPreview.tsx
```diff
@@ -0,0 +1,87 @@
...
+        getSchedulerFilterOverridesQueryString,
+    ]);
+
+    useEffect(() => {
+        onChange(
+            previewChoice === CUSTOM_WIDTH_OPTIONS[1].value
+                ? undefined
+                : previewChoice,
+        );
+    }, [onChange, previewChoice]);
```
magnew, line 75: Do you need to use this useEffect to call an onchange? Can you not pass an onChange to the child that it just calls?

I guess this works, but it's a little confusing.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -343,6 +358,11 @@ const SchedulerForm: FC<Props> = ({
                         <Tabs.Tab value="filters">Filters</Tabs.Tab>
                     ) : null}
                     <Tabs.Tab value="customization">Customization</Tabs.Tab>
+                    {form.values.format === SchedulerFormat.IMAGE && (
+                        <Tabs.Tab value="preview">
+                            Preview and change layout
```
magnew, line 363: This tab title looks a bit long in next to the others. How about just 'Preview'? Or 'Preview and size'?

We should probably chat the Customization tab name too, but that a little out of scope.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -343,6 +358,11 @@ const SchedulerForm: FC<Props> = ({
                         <Tabs.Tab value="filters">Filters</Tabs.Tab>
                     ) : null}
                     <Tabs.Tab value="customization">Customization</Tabs.Tab>
+                    {form.values.format === SchedulerFormat.IMAGE && (
```
magnew, line 361: Maybe we should disable this tab in this case instead of hiding it? It feels a little odd when it pops in.

magnew (approved):
This seems great. I saw a warehouse issue when testing it that I doubt is related, but who knows. 

One UX thing that isn't critical, but is real: on my screen the "generate preview" button is hidden until you scroll down. I wonder if it shouldnt be higher, like right under the radios. 

A couple non-blocking comments. 

Nice work!

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
joaoviana: Done 👍

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
joaoviana: Agree with `Preview and Size` 👍 Yea, was thinking that as well, should they even be on the same Tab? 🤔

packages/frontend/src/features/scheduler/components/SchedulerPreview.tsx
joaoviana: Well... good point. 
let me see if I can avoid this usage 👍

joaoviana:
> This seems great. I saw a warehouse issue when testing it that I doubt is related, but who knows.
> 
> One UX thing that isn't critical, but is real: on my screen the "generate preview" button is hidden until you scroll down. I wonder if it shouldnt be higher, like right under the radios.
> 
> A couple non-blocking comments.
> 
> Nice work!

And on this one I've decreased the Image size 👍

Issue #8430, opened 2023-12-28
Groups can be deleted


magnew:
### Description

Enable group deletion from the UI

PR #8431, opened 2023-12-28
chore: enable deleting groups

magnew:
Closes: #8430 

### Description:

Enable deleting groups from the UI. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
+++ b/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
@@ -1,13 +1,99 @@
-import { Button, Stack, Table } from '@mantine/core';
-import { IconPlus } from '@tabler/icons-react';
-import { FC, useState } from 'react';
+import { Group as UserGroup } from '@lightdash/common';
+import {
+    Button,
+    Group,
+    Modal,
+    ModalProps,
+    Stack,
+    Table,
+    Text,
+    Title,
+} from '@mantine/core';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
```diff
@@ -1,23 +1,123 @@
...
+                </Group>
+            }
+        >
+            <Text pb="md">Are you sure you want to delete this group?</Text>
+            <Group spacing="xs" position="right">
+                <Button
+                    disabled={disableControls}
+                    onClick={onClose}
+                    variant="outline"
+                    color="dark"
```
joaoviana, line 82: Can we remove this? We don't do `color="dark"` for most `Cancel` buttons in Modals across the app. Unless I'm wrong!

joaoviana (approved):
Works great!

I would expect to see an empty state message when there's no groups - can be done later of course 👍 

image

And left a non-blocking comment

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
magnew: Keeping it this way for now as discussed in slack

magnew:
> Works great!
> 
> I would expect to see an empty state message when there's no groups - can be done later of course 👍
> 
> image
> And left a non-blocking comment

Coming soon!

PR #8439, opened 2023-12-29
chore: improve unit tests for explore joins

### LINKED ISSUE #8305: If I use an alias in my model, then I try to join a model, but I don't include the join key in the `fields` list, I get an error.
### Description

We solved an issue here: https://github.com/lightdash/lightdash/issues/7434

Where basically, if I try joining a table, but I only select some fields in the `fields:` list, if I don't include the field I'm joining on, there was an error. We fixed this. _**but**_ a user was reporting this error when they were using a join alias. 

I'm wondering if we missed including some information about aliases in the fix done in #7434 ? 

Here's the info: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1702484509641839

jogeshwar01:
Related to issue #8305 

### Description:
~Here we have a few typos in the values assigned to a few keys for dim2 in `exploreReferenceInJoin` and `exploreReferenceInJoinCompiled`.~

~I have fixed issue #8305 but it sort of considers this change.~
~So will be waiting for the response on this before proceeding.~
~If this is fine, will continue with the existing code else will update my tests accordingly before opening the PR.~

Check comments for context

### Reviewer actions


ZeRego:
Unsure why we need to change these mocks.
These mocks are valid scenarios, even if unlikely, because users can write custom SQL.

Eg where one dimension has a reference to a joined table. The opposite could also happened, where the table B has a reference to table A
```
models:
  - name: a
    meta:       
      joins:
        - join: org
          sql_on: ${b.user_id} = ${a.user_id}
    columns:
      - name: user_id
      - name: org_id
      - name: user_org_hash
        meta:
          dimension:
            type: string
            sql: CONCAT(b.org_id, a.user_id) # here we have a reference to the joined table.
- name: b
  columns:
    - name: org_id
```

ZeRego (changes requested):
(more info in previous comment)
I would accept the PR if it changed the mocks to have SQL with fields from both table A and B. This could avoid confusion to future developers.

jogeshwar01:
Ahh, got it! Thanks for pointing out that case.
As per what I understand now, this case although rare, will indeed help us cover more edge cases in our tests.
So I think we can keep this unchanged. 

I can add another dimension `dim3` to `exploreReferenceInJoin` that would reference the table b. This way we can have SQL fields from both tables (avoiding any confusion)
@ZeRego Do you think this approach fits our needs?

ZeRego:
> Ahh, got it! Thanks for pointing out that case. As per what I understand now, this case although rare, will indeed help us cover more edge cases in our tests. So I think we can keep this unchanged.
> 
> I can add another dimension `dim3` to `exploreReferenceInJoin` that would reference the table b. This way we can have SQL fields from both tables (avoiding any confusion) @ZeRego Do you think this approach fits our needs?

yes, that also helps. 👍

jogeshwar01:
@ZeRego Updated the exploreReferenceInJoin to include dim3.

ZeRego (approved):
lgtm :)

Issue #8138, opened 2023-11-29
Allow having single selection filters in dashboard filters


IrakliJani:
### Description

currently there's no way to configure a filter to accept only single selection. Even if it's `is` user can choose several elements from drop down.

Operators affected: 
- IS
- IS NOT
- STARTS WITH
- ENDS WITH
- INCLUDES
- NOT INCLUDES

Types affected
- STRING
- NUMBER

AC (for saved filters/created in edit mode)
- [ ] The dashboard owner can add a saved filter with the above operators **and** disallow multiple values (a checkbox next to the operator)
- [ ] By default, that checkbox is unchecked, so we always allow multiple values
- [ ] If I save the filter, and someone else overrides it in view mode, it's not possible to select multiple if the saved filter disallows multiple values

AC (for draft filters in view mode)
- [ ] Keep current behaviour as is

### What problem does this solve?

Probably adding new option "is only" would be most elegant in current UX.

pierre-zigzag:
Hi there, I just thought about the same feature, and I'd like to add that it would help UX if we could choose to have Radio buttons as a way to apply those unique value filters.

I was thinking having the ability to display all radio buttons directly in the filter bar for fields with low cardinality; or, for fields with high cardinality, a drop-down menu on the filter pill, like we have today. 

Having all options available directly in the filter bar could help reduce friction for non-technical users on some dashboards (but again, this would require low cardinality in the field, and short strings, otherwise it would clutter the UI very quickly).

hamzahc1:
Requested here: https://lightdash.slack.com/archives/C08DQB442MC/p1741630341160659

PR #8446, opened 2024-01-01
feat: add single selection filter input type and operator

jogeshwar01:
Closes: #8138 

### Description:
Added an `EQUALS_ONLY` FilterOperator

There are basically 2 main types for which we would need this – `STRING` and `NUMBER`.
For these, I have kept the backend same ie. using the IN clause in sql, just that now we would have a single element array when EQUALS_ONLY is selected.

For the frontend,
the numbers can directly use our `FilterNumberInput` which accepts a simple `NumberInput`.

In strings, for cases (like EQUALS, NOT_EQUALS etc), we use the components `FilterStringAutoComplete` and `FilterMultiStringInput` depending on whether it `isTableCalculationField`.
Modifying both of these components was also an alternative here (but that might have made the logic more complex)

So, I have added a new component `FilterSingleStringAutoComplete` as this felt a lot cleaner to me.
This handles both cases - Normal as well as TableCalculationFields using a flag. 
Here we would have a Select rather than a Multi-Select

I have also added some logic in useEffect to handles option changes from multiple to single inputs.
This is useful when a user has selected lets say 3 inputs (in `is` option) and then switches to `is_only` option. Then we update the value with the first element of the input. I do think there will be an optimised way to do this. Would love to discuss this.

Here are some screenshots of the output,

Strings -
image

image

Numbers -
image

image

Combination - 
image

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx
@@ -13,6 +13,7 @@ import { useFiltersContext } from '../FiltersProvider';
 import { getPlaceholderByFilterTypeAndOperator } from '../utils/getPlaceholderByFilterTypeAndOperator';
 import FilterMultiStringInput from './FilterMultiStringInput';
 import FilterNumberInput from './FilterNumberInput';
+import FilterSingleStringAutoComplete from './FilterSingleStringAutoComplete';
 import FilterStringAutoComplete from './FilterStringAutoComplete';
 
 const DefaultFilterInputs = <T extends ConditionalRule>({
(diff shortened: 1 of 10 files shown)


packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx
```diff
@@ -119,20 +121,55 @@ const DefaultFilterInputs = <T extends ConditionalRule>({
...
-                        onChange({
-                            ...rule,
-                            values: newValue ? [newValue] : [],
-                        });
-                    }}
-                />
-            );
+        case FilterOperator.IN_BETWEEN: {
+            switch (filterType) {
+                case FilterType.STRING:
```
joaoviana, line 126: perhaps this (and the case in 115) can be moved to its own `case`? I think it's a bit confusing 🤔

joaoviana: Thank you @jogeshwar01 for the prompt push. I will review this locally and with a duplicate PR to test e2es :)

joaoviana (changes requested):
Looks great @jogeshwar01 !!

I've tested so far: 
* Charts + filters
* Dashboard + saved filters + temporary filters + overrides + exports + scheduled deliveries (send now and scheduled)

Could you add this change?

In: packages/frontend/src/components/DashboardFilter/FilterConfiguration/utils/index.ts

image

joaoviana:
One for maybe @TuringLovesDeathMetal 
I add filter with value `x` with the `is only` operator
I try to add temporary filter with value `y` with the `is only` operator as well
It doesn't let me choose the other values. 

image

Expected? I think so, makes sense

jogeshwar01:
> Could you add this change?
> 
> In: packages/frontend/src/components/DashboardFilter/FilterConfiguration/utils/index.ts

@joaoviana 
Ah, I shouldn't have missed that. Thanks for pointing it out.
I have added this change.

TuringLovesDeathMetal:
Yep, that behaviour makes sense to me!

joaoviana:
Hi @jogeshwar01 , thanks so much for the contribution on this issue. 
We've discussed this internally, and although an operator like `is only` would allow for single selection filtering, we believe that the current `equals/is` filter type input should be extended/reworked to allow only single selection. This requires a bit of thinking with Product, Design, and other customers/users. 
Unfortunately I have to close this PR.

jogeshwar01:
@joaoviana Appreciate the review. I'm on board with adjusting the current equals/is filter input for single selection. I'll be happy to contribute on this once the proposal is finalized.

Issue #8450, opened 2024-01-02
On the first onboarding screen, the 'Organization Name' input disappears when I try to fill it out


PriPatel:
When I try to fill out my Org Name on the first onboarding screen, the input disappears. This means we could potentially not be capturing Org Names from new users. 

https://www.loom.com/share/e8ed8803a8754aeeacd5bd03b50a3050

lancerael:
I think I have a fix for this: https://github.com/lightdash/lightdash/pull/8458

PriPatel:
Thanks @lancerael !

PR #8458, opened 2024-01-02
fix: Remedy for disappearing Organisation input issue 8450

lancerael:
Closes: [On the first onboarding screen, the 'Organization Name' input disappears when I try to fill it out](https://github.com/lightdash/lightdash/issues/8450)

### Description:
Fixed the issue where the Organisation input disappears. It also fixes an unreported issue where the checkbox to allow users with a valid domain disappears in the same circumstances.

_BEFORE:_
image

_AFTER:_
image

_NOTES:_
I believe the intention was to check the existence of an existing organisation name and use it to show/hide these form elements. But the check was accidentally made against the form state.  Updated to check against the value loaded from the context instead.

### Reviewer actions


joaoviana (approved):

joaoviana:
Thanks again for contributing @lancerael ! This brings back the correct implementation 🙌

Issue #8454, opened 2024-01-02
Group membership can be edited


magnew:
### Description

Make it possible to add and remove users from groups.

PR #8471, opened 2024-01-04
chore: add ability to update existing group membership

magnew:
Relates to: #8454 

### Description:

API for updating group membership. Adds a `members` field to the group update endpoint which takes a list of the group's members. Members not yet in the group are added. Members in the group but not in the update are removed. And empty list clears group membership. 

Group name can be updated as before. 

> **Question**: Should we also have separate add and delete users endpoints? This requires the full list of membership, which can be desirable, but it can be problematic for large memberships. 

Example payload--set name and set membership to one user
`{
    "name": "new name",
    "members": [{ "userUuid":"some-id"}, { "userUuid":"some-id2"}]
}`

### Reviewer actions


packages/backend/src/models/GroupsModel.ts
```diff
@@ -187,38 +189,83 @@ export class GroupsModel {
...
+                        .andWhere('groups.group_uuid', groupUuid);
+
+                    await trx('group_memberships')
+                        .insert(newMembers)
+                        .onConflict()
+                        .ignore()
+                        .returning('*');
+                }
+                if (membersToRemove.length > 0) {
+                    await trx('group_memberships')
```
magnew, line 252: Question for someone who knows Knex better than I do: could this delete call be done using `this.removeGroupMember` and still be part of this transaction?

packages/backend/src/models/GroupsModel.ts
```diff
@@ -187,38 +189,83 @@ export class GroupsModel {
...
+                    .update({ name: update.name })
+                    .where('group_uuid', groupUuid)
+                    .returning('*');
+            }
+            if (update.members !== undefined) {
+                const membersToAdd = differenceBy(
+                    update.members,
+                    existingGroup.members,
+                    'userUuid',
+                );
```
ZeRego, line 215: We should make sure the values are unique. API users might send duplicate users.

```
test('check duplicate members ',() => {
    const membersToAdd = differenceBy(
        [{userUuid: 'a'}, {userUuid: 'a'}, {userUuid: 'b'}, {userUuid: 'b'}],
        [{userUuid: 'a'}, {userUuid: 'c'}],
        'userUuid',
    );
    expect(membersToAdd).toStrictEqual([{userUuid: 'b'}]);
})
```

packages/backend/src/models/GroupsModel.ts
```diff
@@ -187,38 +189,83 @@ export class GroupsModel {
...
+                            'users.user_uuid',
+                            membersToAdd.map((m) => m.userUuid),
+                        )
+                        .andWhere('groups.group_uuid', groupUuid);
+
+                    await trx('group_memberships')
+                        .insert(newMembers)
+                        .onConflict()
+                        .ignore()
+                        .returning('*');
```
ZeRego, line 249: Interesting, we haven't used ignore before, what was the thought process?
We don't need the `.returning('*')` btw

packages/backend/src/models/GroupsModel.ts
ZeRego: It wouldn't. 
We would need to change that method arguments to receive the transaction or default to `this.database`. eg:
`async removeGroupMember(member: GroupMembership, trx: Knex = this.database )`

packages/backend/src/models/GroupsModel.ts
```diff
@@ -187,38 +189,83 @@ export class GroupsModel {
...
+                            'users.user_uuid',
+                            membersToRemove.map((m) => m.userUuid),
+                        )
+                        .del()
+                        .returning('*');
+                }
+            }
+        });
+        // TODO: fix include member count
+        return this.getGroupWithMembers(groupUuid, 10000);
```
ZeRego, line 269: Is this TODO for a different PR?

packages/backend/src/models/GroupsModel.ts
```diff
@@ -187,38 +189,83 @@ export class GroupsModel {
...
+                            'group_memberships.user_id',
+                            'users.user_id',
+                        )
+                        .where('group_uuid', groupUuid)
+                        .whereIn(
+                            'users.user_uuid',
+                            membersToRemove.map((m) => m.userUuid),
+                        )
+                        .del()
+                        .returning('*');
```
ZeRego, line 264: We can remove `.returning('*')`

packages/backend/src/models/GroupsModel.ts
magnew: Ok. I will leave this as is for now and think about setting that up if we have more interplay between these.

packages/backend/src/models/GroupsModel.ts
```diff
@@ -187,38 +189,83 @@ export class GroupsModel {
...
-        };
+
+        await this.database.transaction(async (trx) => {
+            let group;
+            if (update.name) {
+                group = await trx('groups')
+                    .update({ name: update.name })
+                    .where('group_uuid', groupUuid)
+                    .returning('*');
+            }
```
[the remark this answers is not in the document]
magnew: I guess not at the moment. Will remove.

packages/backend/src/models/GroupsModel.ts
magnew: Right, that was why to do the .ignore(). Do you think it's better to thow if they are not unique? I don't have a good reason to go one way or the other so ignoring conflicts seems ok.

packages/backend/src/models/GroupsModel.ts
magnew: The thought process was to copy the behavior of the existing raw query for adding one group member which contains a `ON CONFLICT DO NOTHING`. So we'd be ignoring attempts to add the same member twice instead of throwing. Which seems fine to me, but if there is a reason to throw instead we can to that.

packages/backend/src/models/GroupsModel.ts
magnew: 👍

packages/backend/src/models/GroupsModel.ts
magnew: Yes. I think regardless of this current PR we should change getGroupWithMembers so that it can return all of the members. We're going to need that for okta integration instead. It seems reasonable to have it return all of them if you don't pass a number here, but we could also require an explicit param if anyone has an opinion

packages/backend/src/models/GroupsModel.ts
ZeRego: makes sense 👍

PR #8475, opened 2024-01-04
docs: adds embedding reference doc

TuringLovesDeathMetal:
### Description:
Creates a `reference` doc for embedding where there are clear details on what each component in the embed page does.


docs/docs/references/embedding.mdx
```diff
@@ -0,0 +1,99 @@
...
+For embedded content, you can either set user attributes to constant values, or you can adjust the value of the user attribute based on a variable coming from your app. 
+
+For example, if I want users in my app to only see data filtered to their company, I can use a variable user attribute. 
+
+To do this, I would create a user attribute in Lightdash called `company_name` and set some default (e.g. `null`) then, I would pass the Lightdash an external value for company name in the embed code snippet, like:
+
+```node
+import jwt from 'jsonwebtoken';
+const LIGHTDASH_EMBED_SECRET = 'secret'; // replace with your secret
+const projectUuid = '21eef0b9-5bae-40f3-851e-9554588e71a6';
```
ZeRego, line 89:
```suggestion
const projectUuid = '21eef0b9-5bae-40f3-851e-9554588e71a6';
const userCompanyName = await getUserCompanyNameFromDatabase();
```

docs/docs/references/embedding.mdx
```diff
@@ -0,0 +1,99 @@
...
+const data = {
+    content: {
+        type: 'dashboard',
+        dashboardUuid: 'a392ed6d-4c53-4102-89b7-1c57a87df391',
+    },
+    userAttributes: {"company_name":${my_app_company_name_value},
```
ZeRego, line 95:
```suggestion
    userAttributes: {"company_name": userCompanyName},
```

rephus (approved):
The docs are correct, but I find everything a bit confusing (specially around user attributes)...   it might be more useful to have a  separate guide with more detailed use cases like. 

### I want to show a different embed  for different users on my app 
In here you explain how to use user attributes
### I want to make my embed links more secure and let them expire
In here you explain the code snippets and the expiration time 
### I want to generate a publicly available dashboard that doesn't expire
In here you explain the risks about doing this and how to achieve it. 

But this is good as it is, we can do this on a separate PR if you think is a good idea

TuringLovesDeathMetal:
^yesss - this is a really great suggestion

I feel like I can move a bunch of this stuff out to the `how to embed content` guide and then update teh reference doc to be a bit more minima;

PR #8389, opened 2023-12-22
feat: customise width and preview dashboard exports

joaoviana:
Consequence of investigation: #8276 

### Description:

Feature: 
* Allow dashboard `export now` - default behaviour
* Provide customisation section. Users can: change the width + preview what their dashboard would look like

**Demo (most recent):**

https://github.com/lightdash/lightdash/assets/7611706/8cf98155-f5b1-4ae9-8dd1-52fb612ae725

**Demo (outdated):**

https://github.com/lightdash/lightdash/assets/7611706/b1470203-2d27-4511-9e53-478603281009

Added a loading spinner to the notification for `exporting in progress` 

Notes:
* upgrade `puppeteer`
* convert some long-args-methods to one single object
* use `page.screenshot` instead - worked great on `dashboards` and `charts` exports

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #8477, opened 2024-01-04
fix: get screenshot from element instead

joaoviana:
Closes: N/A

### Description:

Some dashboard screenshots are being cropped because we were targeting the page.screenshot when we should continue targetting the `element`. 

PR that introduced this change was: https://github.com/lightdash/lightdash/pull/8389/files#diff-9b86624a0467c9253f755c66928390e96818008cdd1df400c3509984ce4d1071L473-L475

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -493,10 +493,8 @@ export class UnfurlService {
                         });
                     }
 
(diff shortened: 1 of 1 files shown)


rephus (approved):
was there any advantage on using `page.screenshot` ?

joaoviana:
> was there any advantage on using `page.screenshot` ?

nope! just thought it would be more appropriate for this case... But apparently it wasn't 😓

PR #8482, opened 2024-01-04
chore: group membership can be edited

magnew:
Closes: #8454 

### Description:

Adds an edit button and edit dialog to groups that lets the user change the name and membership of a group. 

Note: this is disabled in render for the tests to pass. Check out locally to test?

Here's a little demo:
Kapture 2024-01-04 at 16 55 46

### Reviewer actions


packages/frontend/src/hooks/useOrganizationGroups.ts
```diff
@@ -68,6 +69,41 @@ export const useGroupCreateMutation = () => {
...
+
+export const useGroupUpdateMutation = () => {
+    const queryClient = useQueryClient();
+    const { showToastSuccess, showToastError } = useToaster();
+    return useMutation<Group, ApiError, UpdateGroupWithMembers>(
+        (data) => updateGroupQuery(data),
+        {
+            mutationKey: ['update_group'],
+            onSuccess: async (_, updateGroup) => {
+                await queryClient.invalidateQueries(['update_group']);
```
ZeRego, line 90: I think we can remove this line since there is no query with that key.

ZeRego (approved):

packages/frontend/src/hooks/useOrganizationGroups.ts
magnew: Good catch

PR #8489, opened 2024-01-05
fix: error when using groups with members

magnew:
### Description:

Some group fetches include members, and some don't. We weren't using the member count as part of the query key, so it wasn't being refetched. That caused an error. This fixes that by including member count in the query key. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useOrganizationGroups.ts
+++ b/packages/frontend/src/hooks/useOrganizationGroups.ts
@@ -29,7 +29,7 @@ export const useOrganizationGroups = (
 ) => {
     const setErrorResponse = useQueryError();
     return useQuery<GroupWithMembers[], ApiError>({
-        queryKey: ['organization_groups', includeMembers, queryOptions],
+        queryKey: ['organization_groups', includeMembers],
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/useOrganizationGroups.ts
```diff
@@ -29,7 +29,7 @@ export const useOrganizationGroups = (
 ) => {
     const setErrorResponse = useQueryError();
     return useQuery<GroupWithMembers[], ApiError>({
-        queryKey: ['organization_groups'],
+        queryKey: ['organization_groups', includeMembers, queryOptions],
```
ZeRego, line 32: Nice catch 👍 Query key should always have all the variables that are part of the request. 
I think we could exclude the queryOptions though as they should have no impact on the results.

ZeRego (approved):

packages/frontend/src/hooks/useOrganizationGroups.ts
magnew: Agreed.

PR #8503, opened 2024-01-08
chore: add table for group attribute mappings

magnew:
### Description:

Add table mapping groups to user attributes. 

Note that the groups table references the same user attributes table, `user_attributes` as the user attributes.

We could potentially clean up some of the names here.

### Reviewer actions

### DIFF
--- a/packages/backend/src/database/entities/userAttributes.ts
+++ b/packages/backend/src/database/entities/userAttributes.ts
@@ -16,6 +16,15 @@ export type DbOrganizationMemberUserAttribute = {
     value: string;
 };
 
+export type DbGroupUserAttribute = {
+    group_id: number;
+    organization_id: number;
+    user_attribute_uuid: string;
+    value: string;
+};
+
 export const UserAttributesTable = 'user_attributes';
 export const OrganizationMemberUserAttributesTable =
     'organization_member_user_attributes';
+
+export const GroupUserAttributesTable = 'group_user_attributes';
--- a/packages/backend/src/database/entities/userAttributes.ts
+++ b/packages/backend/src/database/entities/userAttributes.ts
@@ -18,7 +18,7 @@ export type DbOrganizationMemberUserAttribute = {
 
 export type DbGroupUserAttribute = {
     group_id: number;
-    organization_id: number;
+    group_uuid: string;
     user_attribute_uuid: string;
     value: string;
 };
--- a/packages/backend/src/database/entities/userAttributes.ts
+++ b/packages/backend/src/database/entities/userAttributes.ts
@@ -17,7 +17,6 @@ export type DbOrganizationMemberUserAttribute = {
 };
 
 export type DbGroupUserAttribute = {
-    group_id: number;
     group_uuid: string;
     user_attribute_uuid: string;
     value: string;
--- a/packages/backend/src/database/migrations/20240108152115_group_attributes.ts
+++ b/packages/backend/src/database/migrations/20240108152115_group_attributes.ts
@@ -1,0 +1,31 @@
+import { Knex } from 'knex';
+
+const userAttributesTableName = 'user_attributes';
+const groupUserAttributes = 'group_user_attributes';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(groupUserAttributes))) {
+        await knex.schema.createTable(groupUserAttributes, (table) => {
+            table.uuid('group_uuid').notNullable();
+            table.integer('organization_id').notNullable();
+
+            table
+                .foreign(['group_uuid', 'organization_id'])
+                .references(['group_uuid', 'organization_id'])
+                .inTable('groups')
+                .onDelete('CASCADE');
+            table.uuid('user_attribute_uuid').notNullable();
+            table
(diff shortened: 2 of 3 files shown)


packages/backend/src/database/migrations/20240108152115_group_attributes.ts
```diff
@@ -0,0 +1,31 @@
...
+const groupUserAttributes = 'group_user_attributes';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(groupUserAttributes))) {
+        await knex.schema.createTable(groupUserAttributes, (table) => {
+            table.uuid('group_uuid').notNullable();
+            table.integer('organization_id').notNullable();
+
+            table
+                .foreign(['group_uuid', 'organization_id'])
```
owlas, line 13: we can do this without the extra org_id

magnew: Ok. Same for line 14 or 19? Do we need those extra constraints?

packages/backend/src/database/migrations/20240108152115_group_attributes.ts
```diff
@@ -0,0 +1,30 @@
+import { Knex } from 'knex';
+
+const userAttributesTableName = 'user_attributes';
+const groupUserAttributes = 'group_user_attributes';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(groupUserAttributes))) {
+        await knex.schema.createTable(groupUserAttributes, (table) => {
+            table.uuid('group_uuid').notNullable();
+            table.integer('organization_id').notNullable();
```
ZeRego, line 10: I think we need this to be a foreign key and delete on cascade. 

Example code from : packages/backend/src/database/migrations/20230714124708_user_attributes.ts

```
table
                .integer('organization_id')
                .notNullable()
                .references('organization_id')
                .inTable('organizations')
                .onDelete('CASCADE');
```

packages/backend/src/database/migrations/20240108152115_group_attributes.ts
```diff
@@ -0,0 +1,30 @@
+import { Knex } from 'knex';
+
+const userAttributesTableName = 'user_attributes';
+const groupUserAttributes = 'group_user_attributes';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasTable(groupUserAttributes))) {
+        await knex.schema.createTable(groupUserAttributes, (table) => {
+            table.uuid('group_uuid').notNullable();
```
ZeRego, line 9: We could combine this line with line 11 so all the instructions for the same column are chained. 
And do the same for the lines for `user_attribute_uuid`.

packages/backend/src/database/migrations/20240108152115_group_attributes.ts
magnew: I just changed this a little after looking around and learning a bit more. What do you think of how it is now? No more org id. Just the mapping between group and attribute + a value

packages/backend/src/database/entities/userAttributes.ts
```diff
@@ -16,6 +16,15 @@ export type DbOrganizationMemberUserAttribute = {
     value: string;
 };
 
+export type DbGroupUserAttribute = {
+    group_id: number;
```
ZeRego, line 20: The migration doesn't mention a column `group_id`

magnew: Updated.

packages/backend/src/database/migrations/20240108152115_group_attributes.ts
ZeRego: I think that makes sense since the `user_attributes` table already has that relation.

packages/backend/src/database/migrations/20240108152115_group_attributes.ts
magnew: Knex doesn't seem to like them chained when its a foreign key? 

This give an error, for example

`table
    .uuid('group_uuid')
    .foreign('group_uuid')
    .references('group_uuid')
    .inTable('groups')
     .onDelete('CASCADE');`

owlas: If you put it on the column you don't need `.foreign`

```
table.uuid('group_uuid')
.notNullable()
.references('group_uuid')
.inTable('groups')
.onDelete('CASCADE')
```

packages/backend/src/database/migrations/20240108152115_group_attributes.ts
```diff
@@ -0,0 +1,29 @@
...
+                .inTable('groups')
+                .onDelete('CASCADE');
+            table.uuid('user_attribute_uuid').notNullable();
+            table
+                .foreign('user_attribute_uuid')
+                .references('user_attribute_uuid')
+                .inTable(userAttributesTableName)
+                .onDelete('CASCADE');
+            table.string('value').notNullable();
+            table.primary(['group_uuid', 'user_attribute_uuid']);
```
owlas, line 22: nice

owlas (approved):

Issue #8498, opened 2024-01-08
Add Okta + groups integration


joaoviana:
### Description

* Work on Okta login
* Get the list of groups the user is part of
* Create/update the Lightdash group(s) accordingly. 

### What problem does this solve?

UX onboarding

PR #8508, opened 2024-01-08
feat: update groups when users sign up and in (okta)

joaoviana:
Closes: #8498

### Description:

Change Okta Sign up/in flow by using `node-openid-client`: https://github.com/panva/node-openid-client, so that we can get `groups` when requesting `user info`. 
With the previous implementation (using `passport-openidconnect`), the package wouldn't send back `groups` even if they were available: https://github.com/jaredhanson/passport-openidconnect/blob/master/lib/profile.js and relevant issue here: https://github.com/jaredhanson/passport-openidconnect/issues/100

Now, we get the `groups`, create the openId user, and then we add the user to the groups where the Okta group `name` === the `name` of the group in Lightdash. 

Most code on the `OpenIDClientOktaStrategy.authenticate` (which is only used for Okta atm), comes from the readme here: https://github.com/panva/node-openid-client?tab=readme-ov-file#quick-start
^ this includes changes to the current passport strategy, where we utilise a custom strategy. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/authentication.ts
+++ b/packages/backend/src/controllers/authentication.ts
@@ -1,6 +1,7 @@
 /// <reference path="../@types/passport-openidconnect.d.ts" />
 /// <reference path="../@types/express-session.d.ts" />
 import {
+    ArgumentsOf,
     AuthorizationError,
     isOpenIdUser,
     isSessionUser,
@@ -11,6 +12,8 @@ import {
     SessionUser,
 } from '@lightdash/common';
 import { Request, RequestHandler } from 'express';
+import { generators, Issuer, UserinfoResponse } from 'openid-client';
+import type { Profile as PassportProfile } from 'passport';
 import passport from 'passport';
 import {
     GoogleCallbackParameters,
@@ -24,6 +27,7 @@ import {
     Strategy as OpenIDConnectStrategy,
     VerifyFunctionWithRequest,
 } from 'passport-openidconnect';
+import { Strategy } from 'passport-strategy';
 import path from 'path';
 import { URL } from 'url';
 import { lightdashConfig } from '../config/lightdashConfig';
@@ -71,6 +75,195 @@ export const getOpenIdUserFromRequest = ({
         return user;
     }
     return undefined;
+};
+
+const createOpenIdUserFromUserInfo = (
+    userInfo: UserinfoResponse,
+    issuer: string,
+    issuerType: OpenIdIdentityIssuerType,
+    fail: OpenIDClientOktaStrategy['fail'],
+) => {
+    if (!userInfo.email || !userInfo.sub) {
+        return fail(
+            {
+                message: 'Could not parse authentication token',
+            },
+            401,
+        );
+    }
+
+    const displayName = userInfo.name || '';
+    const [fallbackFirstName, fallbackLastName] = displayName.split(' ');
+    const firstName = userInfo.given_name || fallbackFirstName;
+    const lastName = userInfo.family_name || fallbackLastName;
+
+    const openIdUser: OpenIdUser = {
+        openId: {
+            issuer: issuer || '',
+            email: userInfo.email,
+            subject: userInfo.sub,
+            firstName,
+            lastName,
+            issuerType: issuerType || '',
+        },
+    };
+
+    if (userInfo?.groups && Array.isArray(userInfo.groups)) {
+        openIdUser.openId.groups = userInfo.groups;
+    }
+
+    return openIdUser;
+};
+
+const createOpenIdUserFromProfile = (
+    profile: PassportProfile,
+    issuer: string,
+    issuerType: OpenIdIdentityIssuerType,
+    done: ArgumentsOf<VerifyFunctionWithRequest>['3'],
+) => {
+    const email = profile.emails?.[0]?.value;
+    const subject = profile.id;
+    if (!(email && subject)) {
+        return done(null, false, {
+            message: 'Could not parse authentication token',
+        });
+    }
+
+    const displayName = profile.displayName || '';
+    const [fallbackFirstName, fallbackLastName] = displayName.split(' ');
+    const firstName = profile.name?.givenName || fallbackFirstName;
+    const lastName = profile.name?.familyName || fallbackLastName;
+
+    const openIdUser: OpenIdUser = {
+        openId: {
+            issuer: issuer || '',
+            email,
+            subject,
+            firstName,
+            lastName,
+            issuerType: issuerType || '',
+        },
+    };
+
+    return openIdUser;
+};
+
+const setupClient = async () => {
+    const oktaIssuer = await Issuer.discover(
+        `https://${lightdashConfig.auth.okta.oktaDomain}/oauth2/default`,
+    );
+
+    const redirectUri = new URL(
+        `/api/v1${lightdashConfig.auth.okta.callbackPath}`,
+        lightdashConfig.siteUrl,
+    ).href;
+
+    const client = new oktaIssuer.Client({
+        client_id: lightdashConfig.auth.okta.oauth2ClientId ?? '',
+        client_secret: lightdashConfig.auth.okta.oauth2ClientSecret,
+        redirect_uris: [redirectUri],
+        response_types: ['code'],
+    });
+
+    return client;
+};
+
+export class OpenIDClientOktaStrategy extends Strategy {
+    async authenticate(req: Request) {
+        try {
+            const client = await setupClient();
+
+            const redirectUri = new URL(
+                `/api/v1${lightdashConfig.auth.okta.callbackPath}`,
+                lightdashConfig.siteUrl,
+            ).href;
+
+            const params = client.callbackParams(req);
+            const tokenSet = await client.callback(redirectUri, params, {
+                code_verifier: req.session.oauth?.codeVerifier,
+                state: req.session.oauth?.state,
+            });
+
+            const userInfo = tokenSet.access_token
+                ? await client.userinfo(tokenSet.access_token)
+                : undefined;
+
+            if (!userInfo) return this.fail(401);
+
+            try {
+                const { inviteCode } = req.session.oauth || {};
+                req.session.oauth = {};
+
+                const openIdUser = createOpenIdUserFromUserInfo(
+                    userInfo,
+                    tokenSet.claims().iss,
+                    OpenIdIdentityIssuerType.OKTA,
+                    this.fail,
+                );
+
+                if (openIdUser) {
+                    const user = await userService.loginWithOpenId(
+                        openIdUser,
+                        req.user,
+                        inviteCode,
+                    );
+                    return this.success(user);
+                }
+
+                return this.fail(
+                    { message: 'Unexpected error processing user information' },
+                    401,
+                );
+            } catch (e) {
+                if (e instanceof LightdashError) {
+                    return this.fail({ message: e.message }, 401);
+                }
+                Logger.warn(`Unexpected error while authorizing user: ${e}`);
+                return this.error(e);
+            }
+        } catch (err) {
+            return this.error(err);
+        }
+        return this.fail(401);
+    }
+}
+
+export const initiateOktaOpenIdLogin: RequestHandler = async (
+    req,
+    res,
+    next,
+) => {
+    try {
+        const client = await setupClient();
+
+        const redirectUri = new URL(
+            `/api/v1${lightdashConfig.auth.okta.callbackPath}`,
+            lightdashConfig.siteUrl,
+        ).href;
+
+        const state = generators.state();
+        const codeVerifier = generators.codeVerifier();
+        const codeChallenge = generators.codeChallenge(codeVerifier);
+
+        const authorizationUrl = client.authorizationUrl({
+            redirect_uri: redirectUri,
+            response_type: 'code',
+            scope: 'openid profile email',
+            code_challenge: codeChallenge,
+            code_challenge_method: 'S256',
+            state,
+        });
+
+        req.session.oauth = {
+            ...(req.session.oauth ?? {}),
+            codeVerifier,
+            state,
+        };
+
+        return res.redirect(authorizationUrl);
+    } catch (e) {
+        return next(e);
+    }
 };
 
 export const generateOktaUrl = (apiUrlPath: string): string => {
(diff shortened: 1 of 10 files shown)


packages/backend/src/controllers/authentication.ts
```diff
@@ -73,6 +75,104 @@ export const getOpenIdUserFromRequest = ({
...
+                        lastName: userInfo.family_name || fallbackLastName,
+                        issuerType: OpenIdIdentityIssuerType.OKTA,
+                        groups: userInfo.groups as string[] | undefined,
+                    },
+                };
+                const user = await userService.loginWithOpenId(
+                    openIdUser,
+                    req.user,
+                    inviteCode,
+                );
```
joaoviana, line 160: create resuable function - merge this with `genericOidcHandler`

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -63,9 +65,35 @@ apiV1Router.post('/login', passport.authenticate('local'), (req, res, next) => {
 apiV1Router.get(
     lightdashConfig.auth.okta.loginPath,
     storeOIDCRedirect,
-    passport.authenticate('okta', {
-        scope: ['openid', 'profile', 'email'],
-    }),
+    async (req, res) => {
```
owlas, line 68: let's move this handler to `authentication.ts` it could be called `initiateOktaOpenIDLogin`

packages/backend/src/controllers/authentication.ts
owlas: let's refactor this later, we can move the other services off passport generic handler

packages/backend/src/controllers/authentication.ts
```diff
@@ -73,6 +75,104 @@ export const getOpenIdUserFromRequest = ({
     return undefined;
 };
 
+export async function setupClient(): Promise<Client> {
```
owlas, line 78: shouldn't need to export if we move the handler from apiV1.ts

packages/backend/src/controllers/authentication.ts
```diff
@@ -73,6 +75,104 @@ export const getOpenIdUserFromRequest = ({
...
+        client_id: lightdashConfig.auth.okta.oauth2ClientId ?? '',
+        client_secret: lightdashConfig.auth.okta.oauth2ClientSecret,
+        redirect_uris: [redirectUri],
+        response_types: ['code'],
+    });
+
+    return client;
+}
+
+export class OpenIDClientStrategy extends Strategy {
```
owlas, line 98: Rename to be Okta specific for now

packages/backend/src/routers/apiV1Router.ts
joaoviana: done!

packages/backend/src/controllers/authentication.ts
joaoviana: done!

packages/backend/src/controllers/authentication.ts
joaoviana: and done!

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -63,9 +64,7 @@
 apiV1Router.get(
     lightdashConfig.auth.okta.loginPath,
     storeOIDCRedirect,
-    passport.authenticate('okta', {
-        scope: ['openid', 'profile', 'email'],
-    }),
+    initiateOktaOpenIdLogin,
```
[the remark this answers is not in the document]
owlas: this is handled by load balancers

packages/backend/src/controllers/authentication.ts
```diff
@@ -73,6 +76,150 @@ export const getOpenIdUserFromRequest = ({
     return undefined;
 };
 
+const createOpenIdUser = (
```
owlas, line 79: this makes more sense as two separate functions. One for openid userinfo objects and one for passport profiles. Currently this function does 2 distinct things

packages/backend/src/controllers/authentication.ts
```diff
@@ -73,6 +76,150 @@ export const getOpenIdUserFromRequest = ({
...
+        `https://${lightdashConfig.auth.okta.oktaDomain}/oauth2/default`,
+    );
+
+    const redirectUri = new URL(
+        `/api/v1${lightdashConfig.auth.okta.callbackPath}`,
+        lightdashConfig.siteUrl,
+    ).href;
+
+    const client = new oktaIssuer.Client({
+        client_id: lightdashConfig.auth.okta.oauth2ClientId ?? '',
```
owlas, line 126: Did you test this when okta isn't setup? I'm expecting that the passport middleware is simply not mounted based on the code further down in `index.ts`

packages/backend/src/controllers/authentication.ts
joaoviana: Agree, checking how to separate them

packages/backend/src/controllers/authentication.ts
joaoviana: Looking into this!

packages/backend/src/index.ts
```diff
@@ -260,8 +261,8 @@ if (googlePassportStrategy) {
     passport.use(googlePassportStrategy);
     refresh.use(googlePassportStrategy);
 }
-if (oktaPassportStrategy) {
-    passport.use('okta', oktaPassportStrategy);
+if (isOktaPassportStrategyAvailableToUse) {
```
joaoviana, line 264: here @owlas

packages/backend/src/controllers/authentication.ts
joaoviana: oh, it won't even work because of the line in: https://github.com/lightdash/lightdash/pull/8508/files#r1445921312

packages/backend/src/controllers/authentication.ts
```diff
@@ -73,6 +76,150 @@ export const getOpenIdUserFromRequest = ({
...
+    const authorizationUrl = client.authorizationUrl({
+        redirect_uri: redirectUri,
+        response_type: 'code',
+        scope: 'openid profile email',
+        code_challenge: codeChallenge,
+        code_challenge_method: 'S256',
+        state,
+    });
+
+    req.session.oauth = {
```
owlas, line 214: does the redirect work correctly btw? Could you test accessing a dashboard link directly when logged out then complete the okta sign in and you should end up on the dashboard

packages/backend/src/controllers/authentication.ts
```diff
@@ -242,28 +389,9 @@ const genericOidcHandler =
...
-                    subject,
-                    firstName: profile.name?.givenName || fallbackFirstName,
-                    lastName: profile.name?.familyName || fallbackLastName,
-                    issuerType,
-                },
-            };
+
             const user = await userService.loginWithOpenId(
-                openIdUser,
+                createOpenIdUser(undefined, profile, issuer, issuerType),
```
owlas, line 394: yeah this call looks confusing (what is `undefined`?) - it should be `openIdUserFromPassportProfile`

packages/backend/src/models/GroupsModel.ts
```diff
@@ -221,4 +222,17 @@ export class GroupsModel {
             organizationUuid: updatedGroup.organization_uuid,
         };
     }
+
+    async addUserToGroupIfItExists(userUuid: string, groups: string[]) {
```
owlas, line 226:
```suggestion
    async addUserToGroupsIfExist(userUuid: string, groups: string[]) {
```

packages/backend/src/models/GroupsModel.ts
```diff
@@ -221,4 +222,17 @@ export class GroupsModel {
...
+        const existingGroupsUuids = await this.database('groups')
+            .whereIn('name', groups)
+            .select('group_uuid', 'name', 'organization_id');
+
+        existingGroupsUuids.forEach(async (group) => {
```
owlas, line 231: You could do this with a single query instead of reusing an existing model method, which is going to execute queries in a loop. If somebody had a lot of groups this could get expensive

packages/backend/src/services/UserService.ts
```diff
@@ -320,6 +326,17 @@ export class UserService {
             openIdUser.openId.subject,
         );
 
+        if (
```
owlas, line 329: This won't trigger for a newly registered user. This should execute at the bottom of this function with the userUuid from all three cases:
- User is logged in with existing openid
- User is linked to a new openid 
- User is registered with a new openid

packages/backend/src/models/GroupsModel.ts
joaoviana: got it!

packages/backend/src/controllers/authentication.ts
joaoviana: Checking this now!

joaoviana: It works! Copied dashboard link, logged out, input the dashboard link, sign in, automatic redirect to the dashboard

packages/backend/src/models/GroupsModel.ts
joaoviana: sounds good to me! looking into it now 👍

packages/backend/src/models/GroupsModel.ts
joaoviana: done:  https://github.com/lightdash/lightdash/pull/8508/commits/b5291f0b699f43b57c4aafc273276bae890c44ee

packages/backend/src/models/GroupsModel.ts
joaoviana: done here: https://github.com/lightdash/lightdash/pull/8508/commits/25d2f63b6f3cd71cc3b75714d97a982b4b0b9990

packages/backend/src/services/UserService.ts
joaoviana: Thanks for spotting this! I've added this check + insert on those 3 cases. LMK if you would like me to go further in refactoring: https://github.com/lightdash/lightdash/pull/8508/commits/b0e1e46858816e2c2019af91faf3cce0f1b57e2f

packages/backend/src/models/GroupsModel.ts
```diff
@@ -228,13 +228,23 @@ export class GroupsModel {
...
+        const organization = await this.database('organizations')
+            .where('organization_uuid', organizationUuid)
+            .first('organization_id');
+        if (!organization) {
+            throw new NotExistsError('Cannot find organization');
+        }
+
         const existingGroups = await this.database('groups')
             .whereIn('name', groups)
+            .andWhere('organization_id', organization.organization_id)
```
joaoviana, line 247: here it is @owlas

owlas:
just testing this

packages/backend/src/models/GroupsModel.ts
```diff
@@ -221,4 +223,54 @@ export class GroupsModel {
...
+                    .first('user_id')
+            )?.user_id;
+        if (!userIdToInsert) {
+            throw new NotExistsError('Cannot find user');
+        }
+
+        const insertData = existingGroups.map((group) => ({
+            group_uuid: group.group_uuid,
+            user_id: userIdToInsert,
+            organization_id: group.organization_id,
```
owlas, line 268: awesome. to be explicit it would be better to also use the `organization_id` from the organization not those attached to the groups returned from the select statement before

joaoviana: sure! Sounds good :+1:

joaoviana: https://github.com/lightdash/lightdash/pull/8508/commits/fa50b43f1df2e1686ef1c130d74154c573f37f74 and here it is!

owlas (changes requested):
blocking until we improve error handling in okta handlers

owlas (approved):

packages/backend/src/models/GroupsModel.ts
```diff
@@ -221,4 +223,54 @@ export class GroupsModel {
...
+        userUuid,
+        userId,
+        groups,
+        organizationUuid,
+    }: {
+        userUuid?: string;
+        userId?: number;
+        groups: string[];
+        organizationUuid: string;
+    }) {
```
ZeRego, line 237: Had to read the logic to understand why we supported the user uuid and id. 

Suggestion:
```{
        userId: number;
        groups: string[];
        organizationUuid: string;
    } || {
        userUuid: string;
        groups: string[];
        organizationUuid: string;
    }
```

packages/backend/src/models/GroupsModel.ts
```diff
@@ -221,4 +223,54 @@ export class GroupsModel {
...
+        const organization = await this.database('organizations')
+            .where('organization_uuid', organizationUuid)
+            .first('organization_id');
+        if (!organization) {
+            throw new NotExistsError('Cannot find organization');
+        }
+
+        const existingGroups = await this.database('groups')
+            .whereIn('name', groups)
```
ZeRego, line 246: Do we hare requirements for case sensitivity?

packages/backend/src/models/GroupsModel.ts
joaoviana: makes sense, thanks! adding now

packages/backend/src/models/GroupsModel.ts
joaoviana: @owlas and I agreed on that - case sensitivity

packages/backend/src/services/UserService.ts
```diff
@@ -356,6 +374,18 @@ export class UserService {
...
                 },
             });
+
+            if (
+                Array.isArray(openIdUser.openId.groups) &&
+                openIdUser.openId.groups.length &&
+                sessionUser.organizationUuid
+            )
+                await this.groupsModel.addUserToGroupsIfExist({
+                    userId: sessionUser.userId,
```
owlas, line 384: This can be `sessionUser.userUuid` and the `addUserToGroupsIfExist` method can be cleaner

owlas (changes requested):

packages/backend/src/services/UserService.ts
joaoviana: done in: https://github.com/lightdash/lightdash/pull/8508/commits/5ed74ae867c0119c2bf5c9b874aeb204bdc8b252

owlas (approved):

ZeRego (approved):

PR #8516, opened 2024-01-09
docs: update embedding docs

TuringLovesDeathMetal:
### Description:

- updates guide on embedding (how to use user attributes, for example)
- Adds interactivity section to reference on embedding
- adds user metadata section to reference on embedding.


TuringLovesDeathMetal:
@rephus can you please double check the wording in the `user metadata` section? Specifically the last line where I'm like "if you don't add an embed ID, we'll create one using blah blah"

docs/docs/references/embedding.mdx
```diff
@@ -69,32 +66,36 @@ Clicking `generate & copy url` will generate an embed URL based on the configura
...
 To make this easier to integrate, we included some code snippets you can copy and use in your app to generate a valid embed URL.
 
-### Passing external variables to user attributes
+<img src={EmbedDeveloper} width="668" height="466" style={{display: "block", margin: "0 auto 20px auto"}}/>
 
-For embedded content, you can either set user attributes to constant values, or you can adjust the value of the user attribute based on a variable coming from your app. 
+### User metadata
 
-For example, if I want users in my app to only see data filtered to their company, I can use a variable user attribute. 
+You can pass user metadata (specifically, an external user ID) from your application so that anytime someone views your embedded dashboard and runs queries in Lightdash, these query logs are enriched with this user metadata. Specifcally, this user metadata gets added to the [query tags](/references/usage-analytics.mdx#query-tags) for queries run in Lightdash.
```
rephus, line 79: just so you know, on bigquery, only lowercase alphanumeric number (and _) are allowed on externalIds... not sure if this is something we should add to the docs, or just keep in mind. https://lightdash.slack.com/archives/C069QF4UTL5/p1704816492099229

rephus (approved):
LGTM

docs/docs/references/embedding.mdx
```diff
@@ -69,32 +66,35 @@ Clicking `generate & copy url` will generate an embed URL based on the configura
...
     },
-    userAttributes: {"company_name": userCompanyName},
 };
 const token = jwt.sign(data, LIGHTDASH_EMBED_SECRET, { expiresIn: '1 hour' });
 const url = `https://analytics.lightdash.cloud/embed/${projectUuid}/#${token}`;
-```
\ No newline at end of file
+```
+
+If you don't assign an external ID in your embed code, then we will automatically generate and an assign an external ID based on the embed token.
```
rephus, line 100: yes, this is correct, more specifically, by default we use the `iat` (issued at) from the token, and it will be unique for a generated token.

TuringLovesDeathMetal: great, thanks!

docs/docs/references/embedding.mdx
TuringLovesDeathMetal: Is this just in BigQuery, or in all data warehouses?

Issue #8500, opened 2024-01-08
use group attributes in column and row access


TuringLovesDeathMetal:
**********************Permission checks**********************

- Compare OrganizationMember role AND all groups role. If any pass the permission check then grant access.
- ******************************Column security****************************** - if attribute matches user or any group attribute (performance problem)
- **************************Row security************************** - User attributes take precedence over group attributes. Group attribute behaviour not defined, if multiple groups reference the same value

PR #8517, opened 2024-01-09
feat: filter rows and columns based on group attributes

ZeRego:
Closes: #8500 

### Description:

- [x] Column security - valid if the attribute matches a user or any group attribute
- [x] Row security - validated against one value. Value at user level takes precedence over values at group level and if neither is defined, uses the default value. In the case of multiple groups with values for the same attribute, only one will be used. There is no specific group order.

This PR does not improve query performance for user attributes. To be done in separate PR.

### Tested

Needs to be tested locally since we can't edit group attributes in the UI( needs to be changed directly in DB ).

Row filtering eg
```
models:
  - name: events
    meta:
      sql_filter: ${TABLE}.event = ${lightdash.attributes.event}
```

Column filtering eg

```
      - name: event
        meta:
          dimension:
            required_attributes:
              is_admin: "true"
```

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/UserAttributesModel.mock.ts
+++ b/packages/backend/src/models/UserAttributesModel.mock.ts
@@ -3,10 +3,27 @@ export const MOCK_GET_ATTR_VALUES_FOR_ORG_MEMBER_FILTERS = {
     userUuid: 'fake_user_uuid',
 };
 export const MOCK_USER_ATTRIBUTES_NAME_DEFAULT_VALUE = [
-    { name: 'fruit', attribute_default: 'default_fruit' },
-    { name: 'vegetable', attribute_default: 'default_vegetable' },
-    { name: 'section', attribute_default: null },
-    { name: 'category', attribute_default: null },
+    {
+        user_attribute_uuid: 'fruit',
+        name: 'fruit',
+        attribute_default: 'default_fruit',
+    },
+    {
+        user_attribute_uuid: 'vegetable',
+        name: 'vegetable',
+        attribute_default: 'default_vegetable',
+    },
+    {
+        user_attribute_uuid: 'section',
+        name: 'section',
+        attribute_default: null,
+    },
+    {
+        user_attribute_uuid: 'category',
+        name: 'category',
+        attribute_default: null,
+    },
+    { user_attribute_uuid: 'area', name: 'area', attribute_default: null },
 ];
 
 export const MOCK_ORG_MEMBER_ATTRIBUTES_VALUE = [
@@ -14,9 +31,15 @@ export const MOCK_ORG_MEMBER_ATTRIBUTES_VALUE = [
     { name: 'section', value: 'user_section' },
 ];
 
+export const MOCK_GROUP_USER_ATTRIBUTES_VALUE = [
+    { name: 'fruit', value: 'group_fruit' },
+    { name: 'area', value: 'group_area' },
+];
+
 export const EXPECTED_ORG_MEMBER_ATTRIBUTE_VALUES = {
-    fruit: 'user_fruit',
-    vegetable: 'default_vegetable',
-    section: 'user_section',
-    category: null,
+    fruit: ['user_fruit', 'group_fruit'], // expect both user and group values
+    vegetable: ['default_vegetable'], // expect default value
+    section: ['user_section'], // expect user value
+    category: [], // expect no value
+    area: ['group_area'], // expect group value
 };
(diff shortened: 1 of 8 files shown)


packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -4,7 +4,9 @@ export const hasUserAttribute = (
     userAttributes: UserAttributeValueMap,
     attributeName: string,
     value: string,
-) => userAttributes[attributeName] === value;
+): boolean =>
+    !!userAttributes[attributeName] &&
+    userAttributes[attributeName].includes(value);
```
ZeRego, line 9: Updated column permission logic to handle array.

packages/backend/src/queryBuilder.ts
```diff
@@ -108,7 +108,8 @@ export const replaceUserAttributes = (
 
     return sqlAttributes.reduce<string>((acc, sqlAttribute) => {
         const attribute = sqlAttribute.replace(userAttributeRegex, '$1');
-        const userValue: string | null | undefined = userAttributes[attribute];
+        const userValue: string | null | undefined =
+            userAttributes[attribute]?.[0]; // Pick first value, where user value takes precedence over group values
```
ZeRego, line 112: Update row permission logic to use the first value available.

packages/backend/src/models/UserAttributesModel.mock.ts
```diff
@@ -3,20 +3,43 @@ export const MOCK_GET_ATTR_VALUES_FOR_ORG_MEMBER_FILTERS = {
...
 export const EXPECTED_ORG_MEMBER_ATTRIBUTE_VALUES = {
-    fruit: 'user_fruit',
-    vegetable: 'default_vegetable',
-    section: 'user_section',
-    category: null,
+    fruit: ['user_fruit', 'group_fruit'], // expect both user and group values
+    vegetable: ['default_vegetable'], // expect default value
+    section: ['user_section'], // expect user value
+    category: [], // expect no value
+    area: ['group_area'], // expect group value
```
ZeRego, line 44: Should cover all combinations between: default value, user value, group value

packages/backend/src/queryBuilder.ts
owlas: @ZeRego this feels like a fragile api, as a dev how do I know that [0] is the highest precedence attribute?

Somewhere there should be very clear business logic that we have user attributes + group attributes and how we pick the correct one. This approach masks the underlying logic

ZeRego: > Somewhere there should be very clear business logic that we have user attributes + group attributes and how we pick the correct one. This approach masks the underlying logic

In that case, the model would expose the default value, user value, groups values and we need to move the logic to utils functions since we will need it in multiple places. 

```
export type UserAttributeValueMap = Record<string, {
    default: string | null;
    user: string | null;
    groups: string[];
}>;
```

Util functions

```
 export const getAllAttributeValues = (userAttribute: UserAttributeValueMap[number]): string[] => {
     let finalValues: string[];
     if (userAttribute.user) {
         finalValues = [userAttribute.user, ...userAttribute.groups];
     } else if (userAttribute.groups.length > 0) {
         finalValues = userAttribute.groups;
     } else if (userAttribute.default) {
         finalValues = [userAttribute.default];
     } else {
         finalValues = [];
     }
     return finalValues;
 }

export const getAttributeValue = (userAttribute: UserAttributeValueMap[number]): string | null => {
    if (userAttribute.user) {
        return userAttribute.user;
    } else if (userAttribute.groups.length > 0) {
        return userAttribute.groups[0];
    } else if (userAttribute.default) {
        return userAttribute.default;
    } else {
        return null;
    }
}
```

For row permission: 

```
const userValue: string | null = userAttributes[attribute] ? getAttributeValue(userAttributes[attribute]) : null;
```

for column permissions:

```
export const hasUserAttribute = (
    userAttributes: UserAttributeValueMap,
    attributeName: string,
    value: string,
): boolean =>
    !!userAttributes[attributeName] &&
    getAllAttributeValues(userAttributes[attributeName]).includes(value);
```

owlas (approved):

packages/backend/src/queryBuilder.ts
ZeRego: this work will be done in this ticket [Add reference to all attribute values in row level filtering#8536](https://github.com/lightdash/lightdash/issues/8536)

Issue #8061, opened 2023-11-22
Renaming a custom metric breaks a filter


rephus:
If you rename a custom metric that is present on a filter, then the filter breaks and you can't run queries anymore.

![image](https://github.com/lightdash/lightdash/assets/1983672/4feb5dc2-6dac-4078-821f-e919af80b245)

_Originally posted by @rephus in https://github.com/lightdash/lightdash/issues/8002#issuecomment-1822551399_

Sai-Karthik-Tumu:
Hey @rephus , 
Can I work on this issue?

PR #8521, opened 2024-01-09
fix: rename a custom metric with filter

jogeshwar01:
Closes: #8061 

### Description:

Added the logic for updating filters when custom metrics are edited.
As our filter structure can be nested as well, so I have updated the custom metric names recursively.

Initial filters:

image

On renaming:

image

### Reviewer actions

### DIFF
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -167,6 +167,38 @@ export const getFilterRules = (filters: Filters): FilterRule[] => {
     return rules;
 };
 
+export const updateFieldIdInFilterGroupItem = (
+    filterGroupItem: FilterGroupItem,
+    previousName: string,
+    newName: string,
+): void => {
+    if (isFilterGroup(filterGroupItem)) {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        updateFieldIdInFilters(filterGroupItem, previousName, newName);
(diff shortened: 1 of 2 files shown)


rephus:
Renaming a metric on a filter works fine, but if I add a sort and then I rename, I get an blank screen and a JS error 

image

[Screencast from 15-01-24 10:11:07.webm](https://github.com/lightdash/lightdash/assets/1983672/8de8c840-c7e3-4986-9c57-68efa968ce8b)

jogeshwar01:
@rephus Thanks for the review. I have made the corrections. The blank screen error should be fixed now.

I've noticed a couple more issues during testing:

1. When a table calculation involves a custom metric, renaming the metric causes it to break.
2. Similarly, if we apply a filter and then **remove** the custom metric, it results in a similar issue to the one we resolved here related to renaming.

If you come across anything else, please let me know. 
 
We can handle these fixes in the current PR or create separate issues for them. I'd appreciate your thoughts on this.

rephus:
> I've noticed a couple more issues during testing:

I tested this again , thanks for fixing the issue with sorting. 

I can confirm the issues you mentioned, however, they are a bit of an edge case, and more importantly, they don't break the whole app (blank page) like it happened with sorting. 

So I think we can ship this as it is, and create separate issues for the rest.

rephus (approved):

PR #8539, opened 2024-01-10
feat: improve attributes dialog UI

magnew:
### Description:

Improve attribute dialog:
- Sticky footer so buttons don't scroll away
- Header underline
- Higher on the page with more space by default (matches other large dialogs)

Before:
Screenshot 2024-01-10 at 19 44 44

After:
Screenshot 2024-01-10 at 19 41 54

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
+++ b/packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
@@ -136,7 +136,12 @@ const UserAttributeModal: FC<{
                     {userAttribute ? 'Update' : 'Add'} user attribute
                 </Title>
             }
+            yOffset={65}
             size="lg"
+            styles={(theme) => ({
+                header: { borderBottom: `1px solid ${theme.colors.gray[4]}` },
+                body: { padding: 0 },
+            })}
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
```diff
@@ -136,15 +136,20 @@ const UserAttributeModal: FC<{
                     {userAttribute ? 'Update' : 'Add'} user attribute
                 </Title>
             }
+            yOffset={65}
```
IrakliJani, line 139: is the offset from top? we should probably abstract this at some point

IrakliJani (approved):
nice!

packages/frontend/src/components/UserSettings/UserAttributesPanel/UserAttributeModal.tsx
magnew: Yea, it's from the top. For dialogs that can grow it's best not to start them in the middle of the page. Agreed we should abstract it. I think a lot of this dialog style should be centralized. I'm gonna merge this, but I do want to address shared dialog styles soon.

Issue #8511, opened 2024-01-09
Can't filter between two times in the same day


ZeRego:
### To Reproduce

1. Add filter for date with interval that is less than a data. eg: hour, minute, raw
2. select first date
3. select second date in the same day but different hour

Notice it is not possible to select the same day as the first value.

<img width="815" alt="Screenshot 2024-01-09 at 09 50 05" src="https://github.com/lightdash/lightdash/assets/9117144/8ff2b858-e855-49fe-a6d3-3a2741271778">

### App Version

0.925.1

lancerael:
I have a fix for this - would you like to assign me please @ZeRego ?
Or just approve my PR 😸 https://github.com/lightdash/lightdash/pull/8581

PR #8546, opened 2024-01-10
fix: filter between two times in the same day

jogeshwar01:
Closes: #8511 

### Description:
Changed **day** to **second** in `maxDate` and `minDate` in `FilterDateTimeRangePicker` to allow same day range filter.
I have also added the logic to prevent the user from selecting an invalid date.

_There is an alternative to handling this. We can also allow the user to select any date and as per the selection, can modify the other date._

_For example - If a user selects a Start Date that is after the End Date, we can change the End Date to (Start Date + 1 hour) and vice versa._

Here is the output:

image

We can also add a red error state on the input if the time range is invalid.

### Reviewer actions


ZeRego (approved):
lgtm 🏅

PR #8554, opened 2024-01-11
fix: use pathname on sso redirection

rephus:
Closes:  #1304

### Description:

[Screencast from 11-01-24 10:58:46.webm](https://github.com/lightdash/lightdash/assets/1983672/8e43fa7c-be2f-4812-9356-ff70f2905a8a)

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/authentication.ts
+++ b/packages/backend/src/controllers/authentication.ts
@@ -340,24 +340,26 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
 export const getSuccessURLWithReturnTo = (req: Request): string => {
     const url = new URL('/api/v1/oauth/success', lightdashConfig.siteUrl);
     if (req.session.oauth?.returnTo) {
-        url.searchParams.set('returnTo', req.session.oauth.returnTo);
+        url.searchParams.set(
+            'returnTo',
+            decodeURIComponent(req.session.oauth.returnTo),
+        );
     }
     return url.href;
 };
 
 export const redirectOIDC: RequestHandler = (req, res) => {
     // Workaround for https://github.com/jaredhanson/passport/pull/941
-    const queryReturn =
-        typeof req.query.returnTo === 'string'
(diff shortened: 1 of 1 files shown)


packages/backend/src/controllers/authentication.ts
```diff
@@ -340,24 +340,26 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
 export const getSuccessURLWithReturnTo = (req: Request): string => {
     const url = new URL('/api/v1/oauth/success', lightdashConfig.siteUrl);
     if (req.session.oauth?.returnTo) {
-        url.searchParams.set('returnTo', req.session.oauth.returnTo);
+        url.searchParams.set(
+            'returnTo',
+            decodeURIComponent(req.session.oauth.returnTo),
```
owlas, line 345: Why did this move?

owlas: Oh it makes sense, parse and validate when stored, then the function loading this data doesn't need to know to decode the string

packages/backend/src/controllers/authentication.ts
```diff
@@ -340,24 +340,26 @@ export const storeOIDCRedirect: RequestHandler = (req, res, next) => {
...
     // Workaround for https://github.com/jaredhanson/passport/pull/941
-    const queryReturn =
-        typeof req.query.returnTo === 'string'
-            ? decodeURIComponent(req.query.returnTo.toString())
-            : undefined;
-
+    const queryReturn = req.query.returnTo;
     if (
         queryReturn &&
+        typeof queryReturn === 'string' &&
```
owlas, line 356: let's parse this just once, so that we operate on the validated value instead of reconstructing the url:
```typescript
...
let url
try {
  url = new URL(queryReturn, ...)
  // do stuff with url
}
catch (e) {
  if (e.name == 'ERR_INVALID_URL') {
    res.redirect('/')
  }
}

owlas (approved):

PR #8563, opened 2024-01-11
docs: adds docs about groups and user precedence in user attributes

TuringLovesDeathMetal:
- adds information about how user + group attribute values are handled

### DIFF
--- a/docs/docs/references/user-attributes.mdx
+++ b/docs/docs/references/user-attributes.mdx
@@ -187,7 +187,7 @@ In this example, the `tropical_fruits_column` will be visible to them because `c
 
 ### Row filtering
 
-If the user has group and user attribute values assigned, the user attribute values take precedence over group attribute values. 
+If the user has group and user attribute values assigned, the user attribute values take precedence over group attribute values. If there are multiple group attribute values, there is no prioritization over which value will be used. 
 
 For example, if a user is part of a group with the attribute value `kiwi`, another group with the attribute value `orange`, and they've also been assigned as a user to the attribute value `coconut`.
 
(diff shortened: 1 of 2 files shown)


docs/docs/references/user-attributes.mdx
```diff
@@ -167,8 +158,56 @@ models:
...
+          fruits: "coconut"
+```
+
+In this example, the `tropical_fruits_column` will be visible to them because `coconut` is listed in their attribute values `['kiwi','orange', 'coconut']`.
+
+### Row filtering
+
+If the user has group and user attribute values assigned, the user attribute values take precedence over group attribute values. If there are multiple group attribute values, there is no prioritization over which value will be used. 
+
+For example, if a user is part of a group with the attribute value `kiwi`, another group with the attribute value `orange`, and they've also been assigned as a user to the attribute value `coconut`.
```
ZeRego, line 192: This text felt unnecessary for me since we have the scenarios bellow.

docs/docs/references/user-attributes.mdx
```diff
@@ -167,8 +158,56 @@ models:
...
+    meta:
+      sql_filter: ${TABLE}.fruit = ${lightdash.attributes.fruit}
+```
+
+**Scenario 1**:
+- The default attribute value is `banana`.
+- The user's been assigned the value `coconut`
+- They also have group values assigned of `kiwi` and `orange` 
+
+The `${lightdash.attributes.fruit}` values above will equal `coconut`. User attribute values take precedence over group attribute values.
```
ZeRego, line 206: Should we remove the default value from the scenarios as it is not used?

I should replace `values` with `reference` 

```suggestion
The `${lightdash.attributes.fruit}` reference will equal `coconut`. User attribute values take precedence over group attribute values.
```

docs/docs/references/user-attributes.mdx
TuringLovesDeathMetal, line 192:
```suggestion
Let's walk through an example:
```

ZeRego (approved):

docs/docs/references/user-attributes.mdx
```diff
@@ -167,8 +158,56 @@ models:
...
+- They also have group values assigned of `kiwi` and `orange` 
+
+The `${lightdash.attributes.fruit}` reference will equal `coconut`. User attribute values take precedence over group attribute values.
+
+**Scenario 2**:
+- The default attribute value is `banana`.
+- The user's been assigned no value.
+- They also have group values assigned of `kiwi` and `orange` 
+
+The `${lightdash.attributes.fruit}` values above will equal either `kiwi` or `orange. The order of multiple group attribute values is not defined.
```
ZeRego, line 213:
```suggestion
The `${lightdash.attributes.fruit}` reference will equal either `kiwi` or `orange. The order of multiple group attribute values is not defined. 
```

Issue #3798, opened 2022-11-16
Restrict access to tables based on user attributes


owlas:
### Description

Some tables should not be accessible (queryable or searchable) by users depending on their attributes such as group/permissions/tags

The user wouldn't see these in the "Explore -> Tables" view, they would also not be able to view charts or dashboard tiles that require queries to be executed against those tables. 

So if I can't see the Users table then a chart showing "users over time" would show as unavailable.

I imagine this would look the same as user attributes applied to columns, but on a table level. E.g. 

```yaml
models:
  name: users
  meta: 
    ...
    required_attributes: 
      colour: [blue, red]
      age: 18
```

TuringLovesDeathMetal:
**Note:** when I talk about "tables", I mean a model in dbt + any joined models using the `meta` tag in the .yml.

Here's how I'd expect joins to behave. Imagine that I have three .yml files that look like this:

**Table 1: Users**
```yaml
version: 2
models:
  - name: users
    meta:
      joins:
        - join: organizations
          sql_on: ${users.organization_id} = ${organizations.organization_id}
    required_attributes: 
      has_pii_access: 'true'
```

**Table 2: organizations**
```yaml
version: 2
models:
  - name: organizations
```

**Table 3: Tracks**
```yaml
version: 2
models:
  - name: tracks
    meta:
      joins:
        - join: tracks
          sql_on: ${users.user_id} = ${tracks.user_id}
```

As a user with `has_pii_access = false`, this is what I'll see in the UI:

**Table 1: Users**
- This table will not appear in the tables sidebar. 
- If someone else built a chart from this table, but only uses fields from the joined table `organizations`, (so, something like this: https://analytics.lightdash.cloud/share/FnG6tHpkmNXTfhVJv2XQE), I **would not have access to that chart**
- Any chart built from this table or any tables joined to it will not be available for me to explore in Lightdash (they will appear as `no access` if I open them)

**Table 2: organizations**
- This table is not affected by the access restriction. It still shows up in teh sidebar, charts built from this table are still accessible to me. 

**Table 3: Tracks**
- This table appears in the tables sidebar. 
- If I open this table in the Explore view, then the `users` model is hidden from the sidebar
- If someone else built a chart from this table, and it includes a metric or dimension from teh `users` model, then I would see a "no access" error
- If someone else built a chart from this table and it _**did not**_ include a metric or dimensions from the `users` model, I would be able to access the chart just fine.

PR #8567, opened 2024-01-11
feat: restrict access to models with required attributes

ZeRego:
Closes: #3798 

### Description:

Allow developers to restrict access to specific tables using user attributes.

Note: docs in separate PR https://github.com/lightdash/lightdash/pull/8615

eg: 
```
version: 2
models:
  - name: plan
    meta:
      required_attributes:
        is_admin: 'true'
    columns:
      ...
```

Requirements:

- [x] table should not show in /tables page
- [x] table should not show in explore when joined
- [x] dimension or metric that depends on the restricted table should not be available
- [x] neither table, their fields or fields that depend on that table should show in the search results

Demo explores

https://github.com/lightdash/lightdash/assets/9117144/bfae3b5c-b1a6-4a6a-b688-57f19f4a19ca

Demo search

https://github.com/lightdash/lightdash/assets/9117144/e3309cf6-5fd4-4064-b843-6c82a47d1328

### Reviewer actions


packages/common/src/dbt/schemas/lightdashMetadata.json
```diff
@@ -1,5 +1,24 @@
 {
     "definitions": {
+        "RequiredAttributes": {
```
ZeRego, line 3: Define reusable type + improve YML validation to make sure the values are strings or array of strings.

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -28,38 +32,90 @@ export const hasUserAttributes = (
...
-        Object.values(table.dimensions).some(
-            (dimension) => dimension.requiredAttributes !== undefined,
-        ),
+    Object.values(explore.tables).some(
+        (table) =>
+            table.requiredAttributes !== undefined ||
+            Object.values(table.dimensions).some(
+                (dimension) => dimension.requiredAttributes !== undefined,
+            ),
+    );
```
ZeRego, line 41: Update logic to check if either tables or fields have required attributes

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -28,38 +32,90 @@ export const hasUserAttributes = (
...
+    userAttributes: UserAttributeValueMap,
+) =>
+    explore.tables[explore.baseTable].requiredAttributes === undefined ||
+    hasUserAttributes(
+        explore.tables[explore.baseTable].requiredAttributes,
+        userAttributes,
     );
-export const filterDimensionsFromExplore = (
+
+export const getFilteredExplore = (
```
ZeRego, line 53: Update logic to:
- throw error if the user doesn't have access to explore.
- remove tables the user doesn't have access to
- remove joins 
- remove fields that depend on tables the user doesn't have access to

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2109,36 +2110,59 @@ export class ProjectService {
...
-                name: explore.name,
-                label: explore.label,
-                tags: explore.tags,
-                groupLabel: explore.groupLabel,
-                databaseName: explore.tables[explore.baseTable].database,
-                schemaName: explore.tables[explore.baseTable].schema,
-                description: explore.tables[explore.baseTable].description,
-            };
-        });
+        const allExploreSummaries = explores.reduce<SummaryExplore[]>(
```
ZeRego, line 2119: convert map into a reduce so we can also filter out the explores the user doesn't have access to in the same loop

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2109,36 +2110,59 @@ export class ProjectService {
...
+                                explore.tables?.[explore.baseTable]?.schema,
+                            description:
+                                explore.baseTable &&
+                                explore.tables?.[explore.baseTable]
+                                    ?.description,
+                        },
+                    ];
+                }
+                if (
+                    doesExploreMatchRequiredAttributes(explore, userAttributes)
```
owlas, line 2144: why don't we do this with casl? It helps us to centralise access logic

```
ability.can('view', subject('Explore', payload))
```

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -28,38 +32,90 @@ export const hasUserAttributes = (
...
-                            return { ...acc, [dimensionName]: dimension };
-                        return acc;
-                    },
-                    [],
-                ),
-            },
-        };
-    }, {}),
-});
+    return {
```
owlas, line 71: these searches look so expensive

not for this PR but wondering how we can pre-cache some of this work. For example, if there are 10 dimensions and 5 of them required `is_admin` - then there are only two outcomes: either you're an admin and get 10 or you're not and you get 5.

owlas (commented):
some questions

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: Yes, we could have a `Explore` and `Field` subject. 
This means every request, with authentication, would need to fetch all the explores and user attributes to generate the permission rules.
This would slow down the app significantly unless we could cache the authenticated user instead of recreating it on every request.

owlas: It wouldn't need to do that. The explores are the subject of the authentication request (not the definition of an ability) so we don't pre-load any explores in order to build the abilities. But yes we would need the user attributes to build the abilities.

I'm creating a separate PR to clean up all permissions: https://github.com/lightdash/lightdash/issues/8601

packages/backend/src/models/SearchModel/index.ts
```diff
@@ -278,6 +279,17 @@ export class SearchModel {
                                     requiredAttributes: isDimension(field)
                                         ? field.requiredAttributes
                                         : undefined,
+                                    tablesRequiredAttributes: (
```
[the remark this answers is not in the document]
owlas: Did you mean this PR or another PR?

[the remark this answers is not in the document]
ZeRego: done

owlas (approved):

PR #8568, opened 2024-01-11
fix: only append authorizationServerId if exists

joaoviana:
Closes: N/A

Related to: #8498 

### Description:

users were getting: 
```
{"status":"error","error":{"statusCode":500,"name":"UnexpectedServerError","message":"OPError: expected 200 OK, got: 401 Unauthorized","data":{}}}
```

when signing in with okta because it the `authorizationUrl` had `/default` by default. Some Okta apps might not have that Authorization Id (even though it's added by default and can be later disabled - ref: https://developer.okta.com/docs/concepts/auth-servers/#default-custom-authorization-server).

This ensures we append that only when it's available from the configuration. 

Also, remove unused `generateOktaUrl` function - not needed anymore because the package generates the other URLs for us.

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/authentication.ts
+++ b/packages/backend/src/controllers/authentication.ts
@@ -149,13 +149,16 @@ const createOpenIdUserFromProfile = (
 };
 
 const setupClient = async () => {
-    const issuer = lightdashConfig.auth.okta.oauth2Issuer;
-    const authorizationServerPath = lightdashConfig.auth.okta
-        .authorizationServerId
-        ? `/oauth2/${lightdashConfig.auth.okta.authorizationServerId}`
-        : '';
-    const issuerUrl = new URL(authorizationServerPath, issuer);
-    const oktaIssuer = await Issuer.discover(issuerUrl.href);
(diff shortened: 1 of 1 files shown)


packages/backend/src/controllers/authentication.ts
```diff
@@ -150,7 +150,11 @@ const createOpenIdUserFromProfile = (
 
 const setupClient = async () => {
     const oktaIssuer = await Issuer.discover(
-        `https://${lightdashConfig.auth.okta.oktaDomain}/oauth2/default`,
+        `${lightdashConfig.auth.okta.oauth2Issuer}${
+            lightdashConfig.auth.okta.authorizationServerId
+                ? `/oauth2${lightdashConfig.auth.okta.authorizationServerId}`
+                : ''
+        }`,
```
joaoviana, line 157: used `oauth2Issuer` because it must contain the `https://` and only append `/oauth2` when there's an `authorizationServerId`

Tested this with: 
* `default` authorizationServerId
* no `authorizationServerId`

owlas (approved):

Issue #8065, opened 2023-11-23
I want to be able to format my comparison differently to my big value


TuringLovesDeathMetal:
### Description

Right now, if I have a big value chart and I add a format (e.g. M for millions), then teh comparison is shown in the same format. 

Sometimes, the increases are only a few thousands a week, so actually having the format for the comparison in Millions is not very useful. I wish that I could format my comparison differently to my big value. 

### Acceptance criteria

- [ ] There is a `format` option in the `comparison` tab as well as the `layout` tab
- [ ] When i set the format in the `layout` tab, it automatically sets the same format in the `comparison` tab
- [ ] I can override the format in the `comparison` tab
- [ ] Anytime I change teh format in the `layout` tab, it automatically changes the format int he `comparison` tab to be the same.

PR #8572, opened 2024-01-11
feat: format comparison differently in big value charts

jogeshwar01:
Closes: #8065 

### Description:

Added separate style for comparisons in big value charts.

- [x]  There is a format option in the comparison tab as well as the layout tab
- [x]  When we set the format in the layout tab, it automatically sets the same format in the comparison tab
- [x]  We can override the format in the comparison tab

Output -

Layout Tab -

image

Comparison Tab -

image

Example with different formats for big number and comparison -

image

### Reviewer actions


rephus:
image
we should hide the format config on `percentage`

jogeshwar01:
@rephus Thanks for the review.
I have added the condition for comparisonFormat.

rephus (approved):

Issue #8545, opened 2024-01-10
Show inherited roles on groups in project access settings list


magnew:
### Description

Show inherited roles on groups

it should be displayed the same way we display the role inheritance from the Org

![Image](https://github.com/lightdash/lightdash/assets/962095/cde1e76d-ab4e-4898-8b23-52930f582a2a)

PR #8573, opened 2024-01-12
feat: group api with member uuids

IrakliJani:
related to: https://github.com/lightdash/lightdash/issues/8545

### Description:

this is for two reasons:
- for UI to count total users in a group
- show inherited user role from groups in the UI

> why do we have `includeMembers=5`?

to display avatar groups like this: https://mantine.dev/core/avatar/#avatargroup
CleanShot 2024-01-12 at 12 01 42@2x

CleanShot 2024-01-12 at 11 56 03@2x

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/GroupsModel.ts
+++ b/packages/backend/src/models/GroupsModel.ts
@@ -5,24 +5,21 @@ import {
     GroupWithMembers,
     NotExistsError,
     NotFoundError,
-    OpenIdUser,
     ProjectGroupAccess,
-    ProjectMemberRole,
     UnexpectedDatabaseError,
-    UpdateGroup,
     UpdateGroupWithMembers,
-    UpdateProjectGroupAccess,
 } from '@lightdash/common';
 import { Knex } from 'knex';
 import differenceBy from 'lodash/differenceBy';
-import { GroupTableName } from '../database/entities/groups';
+import { EmailTableName } from '../database/entities/emails';
+import { GroupMembershipTableName } from '../database/entities/group_memberships';
 import { OrganizationTableName } from '../database/entities/organizations';
 import {
     DBProjectGroupAccess,
     ProjectGroupAccessTableName,
     UpdateDBProjectGroupAccess,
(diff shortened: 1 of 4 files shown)


packages/backend/src/models/GroupsModel.ts
```diff
@@ -106,58 +103,63 @@ export class GroupsModel {
...
+            name: group.name,
+            createdAt: group.createdAt,
+            organizationUuid: group.organizationUuid,
+            members: memberProfiles.map((row) => ({
                 userUuid: row.user_uuid,
                 email: row.email,
                 firstName: row.first_name,
                 lastName: row.last_name,
             })),
+            memberUuids: memberUuids?.[0]?.member_uuids ?? [],
```
ZeRego, line 162: What would be the difference between this and the FE using the members array? `group.members.map(member => member.userUuid)`

IrakliJani: @ZeRego,
`members` primarily displays the user name and avatar information and has a limit. It includes a query argument, for example, includeMembers=5, which selects only 5 members.

The "memberUuids" serves two purposes:
1. Displaying the total number of users within the group in the UI.
2. Deriving group-inherited permissions, as shown here for orgs: CleanShot 2024-01-15 at 22 00 41@2x

ZeRego: Got it 👍 wasn't understanding the impact of  `includeMembers=5` on this. 
As a developer who consumes the `GroupWithMembers` type, I have no clue that `members` might be incomplete. Perhaps we need to rename that prop to `previewMembers` or have a different a new type for this scenario.
This is something we need to fix in a separate PR.

ZeRego (approved):

packages/backend/src/models/GroupsModel.ts
IrakliJani: that's a good point. we'll have to work further on the API once we do more UI work around groups. I'll have a separate PR for it

PR #8594, opened 2024-01-15
fix: explore from here long url

rephus:
Closes: 

### Description:

https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
Short answer - de facto limit of 2000 characters

I thought about keeping the existing href functionality for links < 2000 chars, but links right now are  unlikely to be 2000 chars even for simple bignumber charts. 

[Screencast from 15-01-24 12:11:06.webm](https://github.com/lightdash/lightdash/assets/1983672/2b1a4dda-dd8f-4699-a7a0-5e2fca34af5f)

side effects: 
* Clicking `back` on explore does`t take you back to the dashboard (you need to click it twice) , this is because we do a redirection, we should fix it on a separate pr an do a replacement on history rather than a push. 
* You can't copy the link from the LinkButton anymore, this is because the shared link is generated on click. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -46,7 +46,7 @@ import {
     IconTelescope,
 } from '@tabler/icons-react';
 import React, { FC, useCallback, useEffect, useMemo, useState } from 'react';
-import { useParams } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


TuringLovesDeathMetal:
I think that's fine! People can still click on the title to "open link in new tab". 

If people complain about it, then we can look to improve it 👀

TuringLovesDeathMetal (approved):

Issue #8533, opened 2024-01-10
Add `how to authenticate` docs link to Docs API page


joaoviana:
Add to this page: https://docs.lightdash.com/api/v1/ the link to authenticate: https://docs.lightdash.com/references/personal_tokens#how-to-authenticate

PR #8598, opened 2024-01-15
chore: add how to authenticate section to tsoa and convert to yml

joaoviana:
Closes: #8533

### Description:

Add tag `Authorization` section and link to docs in `spec.description`

Converted `tsoa.json` to `tsoa.yml`

image

To be done separately: add `authentication` to reference `api key` necessary auth to interact with API.

### Reviewer actions


packages/backend/tsoa.json
```diff
@@ -12,6 +14,10 @@
         },
         "license": "MIT",
         "tags": [
+            {
```
ZeRego, line 17: I know is not as clean but we should have this in the `info.description` instead of `tags`.

We can use markdown to create sections 
```
"description": "Open API documentation for all public Lightdash API endpoints. \n # Authentication\nBefore you get started, you might need to create a Personal Access Token to authenticate via the API. You can create a token by following this guide: https://docs.lightdash.com/references/personal_tokens",
        
```
Screenshot 2024-01-15 at 17 24 31

ZeRego: We could also convert `tsoa.json` to `tsoa.yml` and update the command to `tsoa spec-and-routes --configuration tsoa.yml` 
This way we can have the description over multiple lines.

joaoviana: thanks for both suggestions, will apply them soon!

ZeRego (approved):

Issue #8604, opened 2024-01-16
Generate exposures throws error TypeError: Cannot read properties of null (reading 'replace')


ZeRego:
### To Reproduce

Some users reporting an error when generating exposures
```
✖   Failed to generate exposures file'
TypeError: Cannot read properties of null (reading 'replace')
```

### App Version

0.939.0

PR #8609, opened 2024-01-16
fix(cli): generate exposures for dashboards

ZeRego:
Closes: #8604 

### Description:

Changes:
- Amend SQL to ignore any possible null values and handle when dashboard doesn't have any charts
- Amend SQL to only return one row per dashboard instead of 1 row per dashboard version

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -809,7 +809,7 @@ export class DashboardModel {
         Array<
             Pick<Dashboard, 'uuid' | 'name' | 'description'> &
                 Pick<LightdashUser, 'firstName' | 'lastName'> & {
-                    chartUuids: string[];
+                    chartUuids: string[] | null;
                 }
         >
     > {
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -873,6 +873,7 @@ export class DashboardModel {
                 `${DashboardVersionsTableName}.dashboard_id`,
                 `${DashboardVersionsTableName}.created_at`,
             )
+            .distinctOn(`${DashboardVersionsTableName}.dashboard_id`)
```
rephus, line 876: why is this distinct needed if you're doing .table(DashboardsTableName) and innerJoin is with projects ?

rephus (approved):
Code look good, just 1 non blocking question

packages/backend/src/models/DashboardModel/DashboardModel.ts
ZeRego: because is joining the dashboard version table

PR #8617, opened 2024-01-16
feat: dbt select

ZeRego:
Closes: 

### Description:

Filter by model names or by tag

Screenshot 2024-01-16 at 19 41 27

Screenshot 2024-01-16 at 19 18 06

### Reviewer actions

### DIFF
--- a/packages/backend/src/projectAdapters/dbtBaseProjectAdapter.ts
+++ b/packages/backend/src/projectAdapters/dbtBaseProjectAdapter.ts
@@ -86,7 +86,9 @@ export class DbtBaseProjectAdapter implements ProjectAdapter {
 
         // Validate models in the manifest - models with invalid metadata will compile to failed Explores
         const models = Object.values(manifest.nodes).filter(
-            (node: any) => node.resource_type === 'model',
+            (node: any) =>
+                node.resource_type === 'model' &&
+                (node as DbtRawModelNode).compiled,
         ) as DbtRawModelNode[];
         const manifestVersion = GetDbtManifestVersion(this.dbtVersion);
         Logger.debug(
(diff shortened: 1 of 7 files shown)


packages/backend/src/projectAdapters/dbtBaseProjectAdapter.ts
```diff
@@ -86,7 +86,9 @@ export class DbtBaseProjectAdapter implements ProjectAdapter {
 
         // Validate models in the manifest - models with invalid metadata will compile to failed Explores
         const models = Object.values(manifest.nodes).filter(
-            (node: any) => node.resource_type === 'model',
+            (node: any) =>
+                node.resource_type === 'model' &&
+                (node as DbtRawModelNode).compiled,
```
owlas, line 91: @ZeRego why do we filter for compiled?

owlas: Ah it's because this PR still uses `dbt compile` and `.compiled` is a good way to determine what was selected.

If we rebase this on `dbt ls` we can use the output of `dbt ls` to determine the models

ZeRego: correct 👍 this branch needs to be rebased and adapted for dbt list ( similar to how we handle it in the CLI ) 

https://github.com/lightdash/lightdash/blob/e516966ab7b5b3ad4eadc8e432d6f1ad4fcd6fe4/packages/cli/src/handlers/compile.ts#L111
```
let compiledModelIds: string[] | undefined;
    if (options.useDbtList) {
        compiledModelIds = await dbtList(options);
    } else if (!options.skipDbtCompile) {
        await dbtCompile(options);
    }
    
 ...

 const models = getModelsFromManifest(manifest).filter((model) => {
        if (compiledModelIds) {
            return compiledModelIds.includes(model.unique_id);
        }
        // in case they skipped the compile step, we check if the models are compiled
        return model.compiled;
    });
    ```

Issue #8629, opened 2024-01-17
if we apply a filter and then **remove** the custom metric, it results in a similar issue to the one we resolved here related to renaming.


rephus:
if we apply a filter and then **remove** the custom metric, it results in a similar issue to the one we resolved here related to renaming.

_Originally posted by @jogeshwar01 in https://github.com/lightdash/lightdash/issues/8521#issuecomment-1894363261_

PR #8638, opened 2024-01-17
fix: remove a custom metric with filter

jogeshwar01:
Closes: #8629 

### Description:
Added logic for updating filters when custom metrics are removed.

Example -

Before:

image
image

On removing:

image
image

### Reviewer actions

### DIFF
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -199,6 +199,60 @@ export const updateFieldIdInFilters = (
     }
 };
 
+export const isMetricToDelete = (
+    item: FilterGroupItem,
+    metricName: string,
+): boolean => !isFilterGroup(item) && item.target.fieldId === metricName;
+
+export const removeMetricFromFilterGroupItem = (
+    item: FilterGroupItem,
+    metricName: string,
+): void => {
+    if (isFilterGroup(item)) {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        removeMetricFromFilters(item, metricName);
+    }
+};
+
+export const removeMetricFromFilters = (
+    filterGroup: FilterGroup | undefined,
+    metricName: string,
+): void => {
+    if (!filterGroup) return;
+
+    const processGroupItems = (items: FilterGroupItem[]): FilterGroupItem[] =>
+        items
+            .filter((item) => !isMetricToDelete(item, metricName))
+            .map((item) => {
+                removeMetricFromFilterGroupItem(item, metricName);
+                return item;
+            });
+
+    const isNotEmptyGroup = (item: FilterGroupItem): boolean => {
+        if (isOrFilterGroup(item)) return item.or.length !== 0;
+        if (isAndFilterGroup(item)) return item.and.length !== 0;
+        return true;
+    };
+
+    /* eslint-disable no-param-reassign */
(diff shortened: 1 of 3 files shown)


owlas:
let's review by tomorrow if possible

packages/common/src/types/filter.ts
```diff
@@ -199,6 +199,60 @@ export const updateFieldIdInFilters = (
...
+    };
+
+    /* eslint-disable no-param-reassign */
+    if (isOrFilterGroup(filterGroup)) {
+        filterGroup.or = processGroupItems(filterGroup.or);
+        if (filterGroup.or.length === 0) filterGroup = undefined;
+        else
+            filterGroup.or = filterGroup.or.filter((item) =>
+                isNotEmptyGroup(item),
+            );
```
ZeRego, line 244: Since this is a recursive function, do we need this `else`? wouldn't the previous line `if (filterGroup.or.length === 0) filterGroup = undefined;` remove all empty groups?

packages/common/src/types/filter.ts
```diff
@@ -199,6 +199,60 @@ export const updateFieldIdInFilters = (
...
+    } else if (isAndFilterGroup(filterGroup)) {
+        filterGroup.and = processGroupItems(filterGroup.and);
+        if (filterGroup.and.length === 0) filterGroup = undefined;
+        else
```
ZeRego, line 248: We should add brakets to all conditions for better readability.

packages/common/src/types/filter.ts
```diff
@@ -199,6 +199,60 @@ export const updateFieldIdInFilters = (
...
+    item: FilterGroupItem,
+    metricName: string,
+): void => {
+    if (isFilterGroup(item)) {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        removeMetricFromFilters(item, metricName);
+    }
+};
+
+export const removeMetricFromFilters = (
```
ZeRego, line 217: Can we add some unit tests? 

should return remaining filters ( simple case + nested groups )
should return undefined when all filter rules where removed ( simple case + nested groups )

ZeRego (changes requested):

packages/common/src/types/filter.ts
jogeshwar01: @ZeRego 
This case was mostly to filter out the cases when the `and`, `or` are not of length 0 but rather contain a empty nested `and`, `or` inside. I realized this might fail while writing test cases due to the way the parameters are updated locally inside a function. There would be problems setting it to undefined as it is passed as a reference. So I have moved the code for undefined separately.
It works fine now. Just wanted to know, do you think it shouldn't be this complicated?

jogeshwar01:
@ZeRego 
- Added unit tests for both the above mentioned cases.
- Added brackets to all conditions.
- Updated code to correctly return undefined in case of all filters being removed.

ZeRego (approved):

packages/common/src/types/filter.ts
ZeRego: > Just wanted to know, do you think it shouldn't be this complicated?

Thanks for the changes and especially the tests ❤️  I still ended up refactoring the function as I felt it could be simplified even more. 
It got easier to read and test once I removed the mutability nature. Then I was able to consolidate all the functions into 1 and only check once for empty groups.

Issue #7819, opened 2023-11-06
The description box should expand with text


TuringLovesDeathMetal:
### Description

Chart descriptions/dashboard descriptions can be longer than the text block that is provided int he popover. We should either: 
- Increase the size of the description box as the text increases past one line (i.e. word wrap and expand the box as this happens)
- Increase teh size of the description box by default and word wrap. Then, if the text exceeds the size of the box, we have a scroller to scroll through the description 

<img width="628" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/0f50e223-2b66-4181-96fa-2e23045e83d4">

PR #8639, opened 2024-01-17
feat: expand description box in charts and dashboards

jogeshwar01:
Closes: #7819 

### Description:
Changed TextInput to Textarea to support multiple rows and scroll in our chart and dashboard descriptions.
I have added maxRows of 3 so the box will expand until 3 rows and will scroll from there on.

Example - Update Dashboard
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/SavedDashboards/AddTilesToDashboardModal.tsx
+++ b/packages/frontend/src/components/SavedDashboards/AddTilesToDashboardModal.tsx
@@ -10,6 +10,7 @@ import {
     Modal,
     Select,
     Stack,
+    Textarea,
     TextInput,
     Title,
 } from '@mantine/core';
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/SavedDashboards/AddTilesToDashboardModal.tsx
```diff
@@ -247,10 +248,13 @@ const AddTilesToDashboardModal: FC<AddTilesToDashboardModalProps> = ({
...
                                 {...form.getInputProps('dashboardName')}
                             />
-                            <TextInput
+                            <Textarea
                                 id="dashboard-description"
                                 label="Dashboard description"
                                 placeholder="A few words to give your team some context"
+                                autosize
+                                maxRows={3}
+                                style={{ overflowY: 'auto' }}
```
joaoviana, line 257: do we need this @jogeshwar01 ?

jogeshwar01: @joaoviana Its more of a UI/UX thing.
Basically If we skip these, the only thing that's an issue is a default 2 liner textarea even if there is single line of description. Like this :
image

The `autosize` and `maxRows` helps us counter this. We can skip the overflowY:'auto' here.

I would love your opinion on which option we should choose here. Will update the PR accordingly.

joaoviana: Can you remove the `style={{ overflowY: 'auto' }}` everywhere then @jogeshwar01 ?

joaoviana (changes requested):
Besides the removal of the `overflow` style, one more change is required (since the description can be looooong)

`Dashboard header`'s `Popover.Dropdown` should have `maw={500}`

joaoviana:
just done it! thanks for contributing @jogeshwar01 🙌

joaoviana (approved):

PR #8647, opened 2024-01-18
feat: use lightdash generate for dbt deps packages

owlas:
User requested to generate the metrics in a package they important into their dbt project (i.e. they live in `packages.yml` and `dbt_packages`). 

`lightdash generate` could find the model but was outputting them into the main project instead of `dbt_packages` 

### Reviewer actions

### DIFF
--- a/packages/cli/src/dbt/models.ts
+++ b/packages/cli/src/dbt/models.ts
@@ -137,6 +137,7 @@ type FindAndUpdateModelYamlArgs = {
     docs: Record<string, DbtDoc>;
     includeMeta: boolean;
     projectDir: string;
+    projectName: string;
 };
 export const findAndUpdateModelYaml = async ({
     model,
@@ -144,6 +145,7 @@ export const findAndUpdateModelYaml = async ({
     docs,
     includeMeta,
     projectDir,
+    projectName,
 }: FindAndUpdateModelYamlArgs): Promise<{
     updatedYml: YamlSchema;
     outputFilePath: string;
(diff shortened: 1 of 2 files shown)


packages/cli/src/dbt/models.ts
```diff
@@ -156,8 +158,17 @@ export const findAndUpdateModelYaml = async ({
     const filenames = [];
     const { patchPath } = model;
     if (patchPath) {
-        const { path: expectedYamlSubPath } = patchPathParts(patchPath);
-        const expectedYamlPath = path.join(projectDir, expectedYamlSubPath);
+        const { project: expectedYamlProject, path: expectedYamlSubPath } =
+            patchPathParts(patchPath);
+        const projectSubpath =
+            expectedYamlProject !== projectName
+                ? path.join('dbt_packages', expectedYamlProject)
```
ZeRego, line 165: What is the purpose of updating the `dbt_packages` folder? I thought this feature was for users with [local packages](https://docs.getdbt.com/docs/build/packages#local-packages).

owlas: `dbt_packages` contains any packages listed in the `packages.yml` including those downloaded from a remote server.

If the user wants to generate the yml for one of those external packages then the we put the output in `dbt_packages.yml` and they can manually copy that in to the repo.

It sounds a bit long-winded but testing this with the user, it worked as best as expected! 

I suppose if the `packages.yml` link _is_ a URI to somewhere we can write (like the local filesystem) then it would be 100x better to write it directly to that path. But I'm not sure how commonly that's the case. The current flow works for any package, remote or local

ZeRego: > and they can manually copy that in to the repo.

I suppose these are breaking changes as we currently generate the package's models in the main project. Would current users be upset they now need to manually copy these? 

I assume `dbt_packages` folder is likely git ignored similar to `node_modules`. How do users know what changed after running the generate command?

That's why I thought this was mainly for local packages which have a separate git repo.

ZeRego (approved):

Issue #6327, opened 2023-07-13
Be able to show all labels in the x axis (even if they might overlap)


gordonkjlee:
### Description
X-labels are hidden
<img width="594" alt="image" src="https://github.com/lightdash/lightdash/assets/7281432/1d87452e-3948-4f2c-a5a5-ef4209d9b555">

1. Allow us to "show all labels" even if they overlap (in the example above, they could actually all be displayed without the text overlapping. I think when one label exceeds the bounds, the automatic behaviour is to show/hide every other label)

This is related to: https://github.com/lightdash/lightdash/issues/2272 and https://github.com/lightdash/lightdash/issues/2028

magnew:
There are some options we could add here. Including angled labels which a few people have asked for

erikgregorywebb:
Hi team, bumping this feature request per this ping.
```
Hi team, Happy New Year!
Is there a setting available to rotate x-axis labels? In the chart below, only a handful of values appear.
```

jamiemlee:
+1. Setting rotation, font size or line breaks would useful!

PR #8657, opened 2024-01-18
feat: chart x-axis label rotation

magnew:
Closes: #6327 

### Description:

Enables label rotation for the x-axis (non-flipped). We can add this for the other axes if necessary. I did some of the setup, but wanted to see if we found this the right approach. And if we add it fro the other labels we probably should think about the UI differently; here it is just a single number input. 

A couple notes:
- This moves the axis name in a way that *usually* looks good, but can end up pushing the name offscreen if the fields are long. We could either add field truncation (I played with this and have some ideas) or allow them to place the name. 
- When there are *a lot* of labels, some still get hidden. Theres an echarts bug about this. 

The code in the config file is a little tangled -- I think I made reasonable tradeoffs 😅, but I'm happy to make changes. I added a couple comments in this review to try to clarify.  

Example:
Screenshot 2024-01-18 at 19 24 38

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
@@ -177,7 +177,7 @@ const AxesOptions: FC<Props> = ({ itemsMap }) => {
                             dirtyEchartsConfig?.xAxis?.[0].rotate || 0
                         }
                         min={0}
-                        max={180}
+                        max={90}
                         step={15}
                         size="xs"
                         w={80}
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
@@ -8,6 +8,7 @@ import {
 import {
     Checkbox,
     Group,
+    NumberInput,
     SegmentedControl,
     Stack,
     Switch,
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
```diff
@@ -165,6 +167,26 @@ const AxesOptions: FC<Props> = ({ itemsMap }) => {
                     }}
                 />
             </Group>
+            {!dirtyLayout?.flipAxes && (
```
magnew, line 170: Doesn't show when axes are flipped

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -749,7 +749,25 @@ const calculateWidthText = (text: string | undefined): number => {
     return width;
 };
 
-const getEchartAxis = ({
+const getLongestLabel = ({
```
magnew, line 752: Moved here from inlines below

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1276,12 +1331,12 @@ const useEchartsCartesianConfig = (
         );
     }, [validCartesianConfig, resultsData, itemsMap, pivotDimensions]);
 
-    const axis = useMemo(() => {
+    const axes = useMemo(() => {
```
magnew, line 1334: Just a rename for clarity

rephus (approved):
Tested and reviewed, it works great :+1:

Issue #8628, opened 2024-01-17
When a table calculation involves a custom metric, renaming the metric causes it to break.


rephus:
I've noticed a couple more issues during testing:

When a table calculation involves a custom metric, renaming the metric causes it to break.
 

_Originally posted by @jogeshwar01 in https://github.com/lightdash/lightdash/issues/8521#issuecomment-1894363261_

jogeshwar01:
This also breaks in case of removing a custom metric. So we would have to handle that as well.

PR #8673, opened 2024-01-21
fix: rename table calculation involving custom metric

jogeshwar01:
Closes: #8628 

### Description:

Added code for updating table calculation sql when a custom metric is edited/renamed.

Before:

image
image

On renaming:

image
image

For the case of removing a custom metric, we can solve that in another PR as it is a bit different to this.
There we would have to decide whether we want to delete the table calculation when the custom metric is removed or maybe prompt the user in some way.

### Reviewer actions


almeidabbm (approved):
Thank you for your contribution!

Issue #8674, opened 2024-01-22
Improvements to week number time interval


TuringLovesDeathMetal:
### Description

- [ ] Week number should appear as an integer, not a float on the chart axis ticks + value labels
<img width="423" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/132b0c8a-69c8-4e56-a7a5-ad8e05cf00c1">

- [ ] When week number is added to the results table first, it should be sorted by this value first. Right now, it acts differently in th esorting logic to other dimensions. 
- [ ] 
https://www.loom.com/share/a20123a345734b8a81677867189f14ce

PR #8702, opened 2024-01-23
fix: week num chart float

rephus:
Closes: #8674

### Description:

I'm not sure if there is a better fix for it, I spent an hour on this and I couldn't find a better solution. getAxisLabelFormatter on timeframes will not work here, we could add a new function, but that seems a bit too much for this edge case. 

I think the underlying issue is that echarts by default formats some numbers into floats (perhaps this depends on the `axis type` (time/value/category) ) 

If this is true, I guess it is affecting more types , but this PR is only fixing week_num to avoid introducing some new bugs. I am going to also look a some date bugs in tooltips, so I might reuse some of this code. 

Before:
Screenshot from 2024-01-23 10-15-11

After:

Screenshot from 2024-01-23 10-15-02

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -27,6 +27,7 @@ import {
     Series,
     TableCalculation,
     timeFrameConfigs,
+    TimeFrames,
 } from '@lightdash/common';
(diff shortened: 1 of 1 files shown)


joaoviana (approved):
🚢 it!

Tested this scenario +

- week (date) - works still ok
- other strings - works still ok

PR #8708, opened 2024-01-23
test: fix cypress errors in dashboard and projects

magnew:
Closes: 

### Description:

Fixes a couple persistent cypress errors after react/cypress upgrade. I don't have satisfying explanations about why these broke, but these fixes unblock some of the tests. 

1. Find a specific textarea in the markdown dialog. The test was failing because this was returning multiple elements. I don't know why. 
2. Hide the job-in-progress toast between steps. It seems we were rendering each step in a new dtoast instead of updating the old one. There is more to be done here to make this 'correct' -- we should refactor the showToast hook to use `notifications` and update toasts. But this should work for now. 

### Reviewer actions

### DIFF
--- a/packages/e2e/cypress/e2e/app/dashboard.cy.ts
+++ b/packages/e2e/cypress/e2e/app/dashboard.cy.ts
@@ -237,7 +237,7 @@ describe('Dashboard', () => {
         cy.findAllByText('Add tile').click();
         cy.findByText('Markdown').click();
         cy.findByLabelText('Title').type('Title');
-        cy.get('textarea').type('Content');
+        cy.get('.mantine-Modal-body').find('textarea').type('Content');
         cy.findByText('Add').click();
 
         cy.findByText('Save').click();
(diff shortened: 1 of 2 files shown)


ZeRego (approved):

packages/e2e/cypress/e2e/app/createProjects.cy.ts
```diff
@@ -218,6 +218,7 @@ const defaultRowValues = [
     '11',
     '224',
     '2020-08-10',
+    '33',
```
ZeRego, line 221: If we results have more columns, we may need to fix `bigqueryRowValues`, `databricksRowValues`, `snowflakeRowValues`, `trinoRowValues` as well

magnew: You are right. Trino and Databricks are skipped ATM, I dont know why. I'm not going to update them because I think it will be misleading. 

I've updated BQ and Snowflake.

Issue #2943, opened 2022-08-15
Customise range of x-axis


owlas:
### Description

On a cartesian chart, I can't customise the x-axis range. Although I can customise the y-axis range. 

### Acceptance criteria
Miro sketch: https://miro.com/app/board/uXjVNFeuOjE=/?share_link_id=690462604539

- [ ] There is a new toggle added to the `axes` part of the chart configuration called `auto x-axis range`
- [ ] By default, this is toggled on and hte behaviour of the x-axis is like it is today
- [ ] if you toggle it off, then you can set the min and max values of the x-axis
- [ ] This toggle _only_ appears for x-axes with the type `date/timestamp` or `number`. 
- [ ] If the x-axis is a date or timestamp, then clicking on the input field opens the date picker
- [ ] If the x-axis is a number, then you can click ont he input field and just manually enter a number (just like how it is with the y-axis picker today).

TuringLovesDeathMetal:
What if you could limit the rows selected from your results table in your chart? 
Or, add a filter to the values on the x-axis? 

Also, I guess if we customized the axis range, this would only work for numeric and date types? (not categorical)

ZeRego:
Here is an example where this feature is useful.

> Josiah McDonald
  [19 hours ago](https://lightdash-community.slack.com/archives/C03MG2VTR08/p1675716307216639)
I'm working on visualizing a metric with a 7 day lag -- is there a way to hide part of a visualization without filtering it? Right now the first 7 days of my graph are missing because I'm using a lag function.
<img width="505" alt="image" src="https://user-images.githubusercontent.com/9117144/217287662-b3538ff4-0737-4107-84c4-44d37872bfaa.png">

rephus:
This has been requested again

ceik:
Also ran into this issue today. Filtering (as kinda implied above) did not work. I was able to work around it by turning the dimension into a date, but that only works in this special case.

![image](https://github.com/lightdash/lightdash/assets/7529722/52dec9a9-2fcc-4e23-bca3-8daee9e2ea65)

rephus:
This would also allow customize the range so the weird wide barchart doesn't go out of the axis https://lightdash.slack.com/archives/C0415AUACGY/p1701994842624749

Issue #8262, opened 2023-12-11
When axes are flipped, the yAxis becomes `type:category` 


rephus:
When a chart is flipped (and the xAxis becomes the yAxis) the min/max range doesn't work, 

This is because when the axis changes, it goes from `type: value` to `type: category`

This should be fixed in a separate PR 

![Screenshot from 2023-12-11 12-08-48](https://github.com/lightdash/lightdash/assets/1983672/508e58e7-abfb-4ede-ba12-4f9ca1c679da)
![Screenshot from 2023-12-11 12-09-12](https://github.com/lightdash/lightdash/assets/1983672/61a64f7f-f8c2-4493-ae5e-22da29f342e0)

I'm not sure if this is intentional, but it is affecting min/max ranges 

_Originally posted by @rephus in https://github.com/lightdash/lightdash/issues/8261#issuecomment-1849860341_

jpetey75:
The category type on flipped bar charts is intentional for ECharts, and the min/max breakage this caused was fixed in [#9941](https://github.com/lightdash/lightdash/pull/9941), which applies ranges only to value axes and maps your axis config onto the value axis when axes are flipped. If this is still an issue, let us know and we'll reopen.

PR #8715, opened 2024-01-24
feat: limit axis range 2

rephus:
Closes: #2943

### Description:

This PR is restoring an old commit that was reverted because the bars looked bad outside of the chart, this PR is also addressing that issue. 

Before:

Screenshot from 2024-01-24 13-49-15
Screenshot from 2024-01-24 14-04-02

After:.

Screenshot from 2024-01-24 14-04-42

Screenshot from 2024-01-24 14-04-52
Screenshot from 2024-01-24 14-05-13
Screenshot from 2024-01-24 14-07-16

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
@@ -1,9 +1,12 @@
 import {
+    DimensionType,
     getAxisName,
     getDateGroupLabel,
     getItemLabelWithoutTableName,
+    isField,
     isNumericItem,
     ItemsMap,
+    MetricType,
 } from '@lightdash/common';
 import {
     Checkbox,
@@ -95,6 +98,8 @@ const AxesOptions: FC<Props> = ({ itemsMap }) => {
         setYAxisName,
         setYMinValue,
         setYMaxValue,
+        setXMinValue,
+        setXMaxValue,
         setShowGridX,
         setShowGridY,
         setInverseX,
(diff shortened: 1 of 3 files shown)


TuringLovesDeathMetal (approved):
Yay! this looks great! 🤩 well done Javi!

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
```diff
@@ -142,6 +158,19 @@ const AxesOptions: FC<Props> = ({ itemsMap }) => {
                 }
                 onBlur={(e) => setXAxisName(e.currentTarget.value)}
             />
+            {isXFieldDateOrNumber() && (
```
ZeRego, line 161: Since we run this function on each render, we could use a `useMemo` instead.

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1433,13 +1426,26 @@ const useEchartsCartesianConfig = (
                 : getResultValueArray(rows, true);
         try {
             if (!itemsMap) return results;
-
             const xFieldId = validCartesianConfig?.layout?.xField;
             if (xFieldId === undefined) return results;
+            const { min, max } = axes.xAxis[0];
```
ZeRego, line 1431: Not sure if related to this line, but the range was not applied when I flipped the axis

Screenshot 2024-01-25 at 13 33 48

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/AxesOptions.tsx
rephus: yeah, good idea, however, this function depends on a few variables that are created after 
>     if (!isCartesianVisualizationConfig(visualizationConfig)) return null;
and we can't `error  React Hook "useMemo" is called conditionally.`

Which means we'll have to move everything before `if (!isCartesianVisualizationConfig(visualizationConfig)) return null;` which may be even less performant

What do you think ?

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
ZeRego: ticket created https://github.com/lightdash/lightdash/issues/8262

PR #8704, opened 2024-01-23
feat: enable groups env variable

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/8703

### Description:

- adds env variable `GROUPS_ENABLED`
- removes feature flag `group-management`
- documents env variable in docs

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #8727, opened 2024-01-25
docs: adds info about groups availability

TuringLovesDeathMetal:
adds info about how. groups is only available in pro or enterprise or for self-hosted users.

### DIFF
--- a/docs/docs/references/groups.mdx
+++ b/docs/docs/references/groups.mdx
@@ -4,6 +4,12 @@ import GroupsManagementPage from './assets/groups-management-page.jpg';
 
 Groups are a way to manage access to multiple users at the same time. A user can belong to multiple groups. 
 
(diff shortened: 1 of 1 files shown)


IrakliJani (commented):
@TuringLovesDeathMetal have a similar change here: https://github.com/lightdash/lightdash/pull/8704/files#diff-8c986bf44b56bf769dfb249404639f25b987c6d32d469803168aa6ff713b3068R13

which one should we ship?

TuringLovesDeathMetal:
@IrakliJani ah sorry! I didn't see that!

We can ship yours - just maybe move it to the top of the doc? (so where I put mine?) I'll close this PR :)

IrakliJani:
@TuringLovesDeathMetal ok cool I just updated mine

Issue #8728, opened 2024-01-25
Add "!true" and "!false" metric filter types


GtheSheep:
### Description

I want to be able to filter my boolean values like `!false` and `!true` in metric filters.

If you tyr to do this today, then you get an error:

"No function implemented to render sql for filter type notEquals  on dimension of boolean type"

This has been problematic when using JOINs to table that do not have the same level of completeness as the base table and trying to apply filters

### Acceptance criteria

- [ ] `!true` and `!false` work as metric filters using boolean dimensions
- [ ] we've updated the filters docs: https://docs.lightdash.com/references/metrics#available-filter-types

TuringLovesDeathMetal:
👋 Hey @GtheSheep 

I've made this into a feature request instead of a bug - I think that we just don't support that metric type filter for now! 

You can see all of the metric filter types we support here: https://docs.lightdash.com/references/metrics#available-filter-types

(It looks like `true`/`false` are not treated like strings - so we'll need to add a specific metric filter type for it, like we have for `null` values)

GtheSheep:
@TuringLovesDeathMetal - opened a PR for it if anyone could comment on whether or not it makes sense as a way to go?

PR #8729, opened 2024-01-25
feat: Render not equals SQL for boolean filter

GtheSheep:
Closes: #8728

### Description:
This may be intentionally designed as such, so apologies if so!!
But we were attempting to filter by "!true" and came across this, in a column that was brought in from a join onto our base table in such a way that true, false, null were all available, and we explicitly want [false, null].
Filtering as a list of values also fails as it casts them to string if that's preferable to fix?

Screenshot 2024-01-25 at 17 30 52

### Reviewer actions


ZeRego (approved):
lgtm :) 
Running tests before merging

Issue #8505, opened 2024-01-08
Date on tooltip showing wrong date


TuringLovesDeathMetal:
### Description

This was "fixed" here: https://github.com/lightdash/lightdash/issues/2657

But it's popped up again for some users. 

See slack thread here: https://lightdash.slack.com/archives/C03EC7PCW8H/p1704724704742839

rephus:
I know why this is happening, and I can replicate, but I'm not sure how to fix it without doing a major refactoring 

![Screenshot from 2024-01-23 06-23-57](https://github.com/lightdash/lightdash/assets/1983672/6e4a0128-f512-472d-827e-d1e4ca1f1bb3)

My backend is on UTC+2 and my frontend is on UTC-9 

![Screenshot from 2024-01-23 06-24-05](https://github.com/lightdash/lightdash/assets/1983672/57014875-6c24-46db-b898-98a4f8e81e11)

My backend is returning the `2020-06-30 22:00` as `2020-07` in raw. 

![Screenshot from 2024-01-23 06-23-50](https://github.com/lightdash/lightdash/assets/1983672/4c16dd48-034c-4352-b5c7-0a8bb8b28f15)

We use the timestamp to construct the chart, echarts converts this date into a timestamp `1593554400000` , even if we try to convert that back to UTC, we can get `2020-06-30 22:00`

![Screenshot from 2024-01-23 06-24-36](https://github.com/lightdash/lightdash/assets/1983672/ed070638-1692-42a6-bf8a-f8cc96bc86cf)

Possible solutions. 

- Major refactoring to return results data or without timezone 
- could we add the formatted month as metadata in the chart ?

rephus:
I think the fix should happen on the backend, where we strip the timezone information from the dates. 

The fix should happen here https://github.com/lightdash/lightdash/blob/main/packages/common/src/index.ts#L852

Example: 

My backend gets the data from the database `Moment<2020-07-01T00:00:00+02:00>` (it has timezone info) , and this is converted to a `month` 

![image](https://github.com/lightdash/lightdash/assets/1983672/0835288a-a060-4e5b-a4ad-da629e9e2d17)

However, the data we return to the frontend is

![image](https://github.com/lightdash/lightdash/assets/1983672/a7c5573f-7a5f-4df7-b80b-62e473c42e83)
WHich is wrong. 

We use the `formatted` for most of the charts / results , however the `raw` is a wrong data. 
[...]

PR #8733, opened 2024-01-26
fix: timezone raw results

rephus:
Closes: https://github.com/lightdash/lightdash/issues/8505

### How to test

Change your own timezone, refresh the browser, and make sure the tooltip matches the data. It is easier to test if you limit the results to just 1 row
### Description:

This should fix the timezone tooltip issue by making sure everything is on UTC, including the results we return from the backend (in raw) 

Screenshot from 2024-01-26 01-09-08
Screenshot from 2024-01-26 01-09-17
Screenshot from 2024-01-26 01-12-37

## Raw timestamp

Screenshot from 2024-01-26 01-12-37

### Reviewer actions

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import { z } from 'zod';
 import { UserActivity, ViewStatistics } from './types/analytics';
 import {
@@ -92,11 +93,12 @@ import { SlackChannel } from './types/slack';
 import { Space } from './types/space';
 import { ApiSshKeyPairResponse } from './types/SshKeyPair';
 import { TableBase } from './types/table';
+import { TimeFrames } from './types/timeFrames';
 import { LightdashUser, UserAllowedOrganization } from './types/user';
 import { ValidationResponse } from './types/validation';
 import { convertAdditionalMetric } from './utils/additionalMetrics';
 import { getFields } from './utils/fields';
-import { formatItemValue } from './utils/formatting';
+import { formatItemValue, formatTimestamp } from './utils/formatting';
 import { getItemId, getItemLabelWithoutTableName } from './utils/item';
 
 export * from './authorization/index';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1570,22 +1570,6 @@ const useEchartsCartesianConfig = (
 
                         return `${tooltipHeader}<br/><table>${tooltipRows}</table>`;
                     }
-                    if (
```
rephus, line 1573: This code was buggy and redundant, we just need to make sure axislabel is correct (this is what we use in tooltips by default) 

Screenshot from 2024-01-26 01-12-37

packages/common/src/index.ts
```diff
@@ -1,3 +1,4 @@
+import moment from 'moment';
```
magnew, line 1: Can we use Dayjs instead of moment? We have been trying to move that way. But I dont know if there is a constraint here.

magnew (approved):
Tested, looks good! Makes sense. If we can use dayjs instead of moment, that would be cool.

packages/common/src/index.ts
rephus: sure, let me try that

rephus: dayjs requires a ¿plugin? and the implementation might be slightly different, so I'm afraid of replacing this with dayjs... 

```
var utc = require('dayjs/plugin/utc')
dayjs.extend(utc)
```

https://day.js.org/docs/en/plugin/utc

rephus: ok, I made the change, could you quickly review and test again ?

magnew: Interesting, I didn't know it needed a plugin for this. It looks like it's because of package size. It's *supposed* to have basically the same API, so I would expect this to work. But if it doesn't seem to, we can keep moment here. I'm not trying to add more work to this, just following the moment docs that suggest using dayjs:

https://momentjs.com/docs/#/-project-status/recommendations/

We do already have moment anyway, so its not super important to avoid it here

packages/common/src/index.ts
```diff
@@ -849,6 +853,22 @@ export function itemsInMetricQuery(
...
+    value: any,
+) {
+    const isTimestamp =
+        isField(field) &&
+        (field.type === DimensionType.DATE ||
+            field.type === DimensionType.TIMESTAMP);
+
+    if (isTimestamp) {
+        // We want to return the datetime in UTC to avoid timezone issues in the frontend like in chart tooltips
+        return dayjs(value).utc(true).format();
```
rephus, line 867: previous moment implementation 

> return moment(value).utc(true).format();

magnew: Looks the same, no? It seems to work? I'm happy to test again after standup.

almeidabbm:
Tested this one, works for me as well 🎉

magnew (approved):
I tested again with the dayjs version and it still seems to work. Thanks for making the change!

@almeidabbm it might be wise to try whatever test you did again.

rephus:
this e2e error seem unrelated

image

rephus:
all timezone related tests are passing (including timezone e2e tests)

Issue #8736, opened 2024-01-26
Inconsistent spacing between label and field 


PriPatel:
<img width="227" alt="Screenshot 2024-01-26 at 10 00 56" src="https://github.com/lightdash/lightdash/assets/12776522/58c36c79-e179-4f32-b724-c9ee35d0eec9">

The spacing between the Delivery name and the field below it is no consistent with the other fields in the same form

- [ ] Delivery name spacing should match the other fields in the form

jogeshwar01:
@PriPatel Can I take this up?

PriPatel:
@jogeshwar01 yep go for it!

PR #8740, opened 2024-01-26
fix: inconsistent spacing in scheduler form

jogeshwar01:
Closes: #8736 

### Description:

To add spacing b/w label and field in TextInput, we cannot use mt or mb as these props are typically used for setting margin on the component as a whole, not for internal elements like labels.

So I have added a marginBottom of 0.25rem to the label. Considering other inputs have a margin of xxs which is equal to rem(4) in our mantine styles. Hence in rem it is, 4/16 = 0.25rem.

Reference - https://mantine.dev/styles/rem/

Before:

image

After:

image

### Reviewer actions


joaoviana (approved):
nice! Thanks again @jogeshwar01 🙌

Issue #8749, opened 2024-01-29
Development docker setup not picking up environment variables


almeidabbm:
### Description

When running Lightdash locally using docker-compose, environment variables are not being picked up from `.env.development.local`

### To Reproduce

1. Change an environment variable in `.env.development.local`
2. Run docker compose: `docker compose -p lightdash-app -f docker/docker-compose.dev.yml --env-file .env.development.local up --detach --remove-orphans`
3. Go into the container: `docker exec -it lightdash-app-lightdash-dev-1 bash`
4. Run `env` in the container

### App Version

v0.965.3

PR #8750, opened 2024-01-29
chore: reference environment variables in docker-compose.dev.yml

almeidabbm:
Closes: #8749
### Description:
When running Lightdash locally using docker-compose, environment variables are not being picked up from `.env.development.local`

Steps to reproduce:
1. Change an environment variable in `.env.development.local`
2. Run docker compose: `docker compose -p lightdash-app -f docker/docker-compose.dev.yml --env-file .env.development.local up --detach --remove-orphans`
3. Go into the container: `docker exec -it lightdash-app-lightdash-dev-1 bash`
4. Run `env` in the container

### Reviewer actions

- [ ] I have manually tested the changes locally


.env.development
```diff
@@ -57,3 +57,8 @@ OTEL_SDK_DISABLED=false
 #EMAIL_SMTP_ALLOW_INVALID_CERT=
 #EMAIL_SMTP_SENDER_NAME=
 #EMAIL_SMTP_SENDER_EMAIL=
+
+HEADLESS_BROWSER_HOST=headless-browser
```
rephus, line 61: You still need to run headless browser in a separate docker container..

rephus (approved):

.env.development
rephus: that's ok, it could be confusing for some people because the error is less clear if you are not running headless_browser. But we can merge as it is

Issue #8754, opened 2024-01-29
Error when showing datetime filters 


rephus:
### Description

I created a chart with an `in between` datetime filter 

On both edit and view mode, I try to show the filters panel, but I get a blank page instead

![image](https://github.com/lightdash/lightdash/assets/1983672/be9c2986-9a7d-430c-bbcc-4585b317d65f)

![image](https://github.com/lightdash/lightdash/assets/1983672/b79ada08-70e3-4d26-b5ab-5fae7a0f2a3a)

```
Uncaught RangeError: invalid date
    formatValue HiddenDatesInput.tsx:17
    HiddenDatesInput HiddenDatesInput.tsx:24
    React 14
    toggleExpandedSection ExplorerProvider.tsx:1212
    onToggle FiltersCard.tsx:117
    handleToggle CollapsableCard.tsx:57
    onClick CollapsableCard.tsx:89
    React 25
    <anonymous> index.tsx:9
[HiddenDatesInput.tsx:17:17](http://localhost:3000/node_modules/@mantine/dates/src/components/HiddenDatesInput/HiddenDatesInput.tsx)
```

This might be related wtih the MS issue we fixed recently

rephus:
This prevents adding `milliseconds` to new `in between` filters, which causes the `render`  to break. 

If a user already has a chart with miliseconds stored and he is trying to visualize, it will fail anyway. That might require a different fix.

PR #8755, opened 2024-01-29
fix: in between millisecond

rephus:
Closes:  #8754

### Description:

The component is not able to `render` milliseconds ,  so we need to format it to avoid those from being saved. 
Before:

image

:fire: blank page on panel toggle :fire: 

After:

image

image

### Reviewer actions


IrakliJani (approved):
maybe we should leave a `FIXME:` mark here to revisit this after mantine v7 migration?

Issue #8688, opened 2024-01-22
I can apply formatting to my numeric Custom Metric


joaoviana:
### Description

- [x] When you create a metric that is a numeric type, you have the option to format the custom metric while you're creating it.
- [x]  Add another accordion to apply the formatting (below filters) if numeric type - hide that if not
- [x] This option is only available for numeric-type metrics
- [x] The formatting options are the same as table calculations
- [x] Add new column to the additional_metrics table, which should mimic the same structure as the table_calculations one
- [x] I can enable and disable the formatting

PR #8756, opened 2024-01-29
feat: save custom metric percent format

almeidabbm:
Relates to: #8688 

### Description:

- Adds prefix, suffix, separator and currency columns to `'saved_queries_version_additional_metrics'` table.
- Saves and displays custom metric percent format.
- Improvements to typings.
- `convertAdditionalMetric` function returns new saved properties for custom metrics ( prefix, suffix and separator )
- renames types on formatting (for both Table Calculations and Custom metrics) so these are shared

Formatting settings that currently work:
- Round
- Compact
- Percent

Missing:
- Prefix
- Suffix
- Separator
- Currency

### Reviewer actions


packages/common/src/types/field.ts
```diff
@@ -395,7 +395,10 @@ export const isMetric = (field: Field | undefined): field is Metric =>
 export const isNonAggregateMetric = (field: Field): boolean =>
     isMetric(field) && NonAggregateMetricTypes.includes(field.type);
 
-export interface Metric extends Field {
+export interface Metric
+    extends Field,
+        Pick<CustomFormat, 'prefix' | 'suffix' | 'separator'>,
+        Pick<AdditionalMetric, 'baseDimensionName'> {
```
joaoviana, line 401: we should remove this bit ^

packages/common/src/utils/formatting.ts
```diff
@@ -270,7 +284,7 @@ export function formatFieldValue(
         case MetricType.COUNT:
         case MetricType.COUNT_DISTINCT:
         case MetricType.SUM:
-            return formatValue(value, { format, round, compact });
+            return formatValue(value, formatOpts);
```
joaoviana, line 287: let's bring the previous change back: 

```
return formatValue(value, { format, round, compact });
```

and remove the `isAdditionalMetric` check 👍

joaoviana (changes requested):
2 changes requested 👍

packages/common/src/types/field.ts
almeidabbm: Done!

packages/common/src/utils/formatting.ts
almeidabbm: Done as well 👌

joaoviana (approved):
Looks great!

Issue #8536, opened 2024-01-10
Add reference to all attribute values in row level filtering


ZeRego:
Related to #8500

## Description

A user can now have multiple values for the same attribute but row filtering only supports 1 value.

Current business logic:
Value at user level takes precedence over values at group level and if neither is defined, uses the default value. In the case of multiple groups with values for the same attribute, only one will be used. There is no specific group order.

## Proposed changes

https://github.com/lightdash/lightdash/pull/8517#discussion_r1447464839

~After that refactor we would need to find a new reference for all the values that is backwards compatible.~

There is no use case to access just 1 value when that value is unreliable. They should always treat the SQL as if there are multiple values. 
This would match the same behaviour for column and table filtering where we match against all values independently if they are set at the user/group/default level.

## Acceptance criteria

- [ ] access all values in SQL template (i.e. `attribute_value IN ('attribute_1', 'attribute_2', 'attribute_3')`)
- [ ] update docs https://docs.lightdash.com/references/user-attributes#1-row-filtering-with-sql_filter

PR #8769, opened 2024-01-30
feat: access all attribute values when filtering rows

ZeRego:
Closes: #8536 

### Description:

Developers should have access to all attribute values, independently if they are set at the user level, group level, or default value. This matches the behaviour we have for Table and Column filtering. 

- [x] backwards compatible ( if not using group attributes )
- [x] access all values in SQL template
- [x] compatible with all warehouses: [bigquery](https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#in_operators) [postgres](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-in/) [snowflake](https://docs.snowflake.com/en/sql-reference/functions/in) [databricks](https://docs.databricks.com/en/sql/language-manual/functions/in.html) [trino](https://trino.io/docs/current/functions/comparison.html#row-comparison-in)

- [x] update documentation
  - https://deploy-preview-8769--peaceful-bassi-cbf284.netlify.app/references/user-attributes#row-filtering-with-sql_filter
  - https://deploy-preview-8769--peaceful-bassi-cbf284.netlify.app/references/user-attributes#row-filtering

Screenshot 2024-01-30 at 15 06 54

---

### Example 1: User has an attribute  with 1 value

**Current behaviour**

```sql_filter: ${TABLE}.sales_region = ${lightdash.attributes.sales_region}```

That would convert to

```sales.sales_region = 'USA'```

**New behaviour**

```sql_filter: ${TABLE}.sales_region = ${lightdash.attributes.sales_region}```
That would convert to
```sales.sales_region = 'USA'```

Note that is it backwards compatible.

--- 

### Example 2: User has an attribute with 2 values 

**Current behaviour**

```sql_filter: ${TABLE}.sales_region = ${lightdash.attributes.sales_region}```

That would convert to

```sales.sales_region = 'USA'```

Note that the user can only access 1 value. This is the problem we are solving in this PR.

**New behaviour** 

```sql_filter: ${TABLE}.sales_region = ${lightdash.attributes.sales_region}```

That would convert to

```sales.sales_region = 'USA', 'EU'```

Note that this would throw an error since it is not valid SQL. This is good because users won't have access to something they shouldn't have while admins can fix the SQL/YML. 

Our docs will now recommend using `IN` operator when handling user attributes to make sure it can handle multiple values.

So they would have the following SQL

```sql_filter: ${TABLE}.sales_region IN (${lightdash.attributes.sales_region})```

That would convert to

```sales.sales_region IN ('USA','EU')```

### Reviewer actions

### DIFF
--- a/docs/docs/references/user-attributes.mdx
+++ b/docs/docs/references/user-attributes.mdx
@@ -63,14 +63,17 @@ When referencing user attributes in SQL you can use the following [SQL variables
 
 You can use user attributes to filter the rows returned by a query. This is useful if you want to restrict the data 
 based on the user's attributes. To reference a user attribute in your sql, use the special lightdash reference 
-`${lightdash.attributes.<attribute_name> }`. For example, if you have a user attribute called `sales_region` you can
+`${lightdash.attributes.<attribute_name> }`.
+We recommend using the `IN` operator because the attribute might have multiple values.
+
+For example, if you have a user attribute called `sales_region` you can
 use it in your sql like this:
 
 ```yaml
 models:
   - name: my_model
     meta:
-      sql_filter: ${TABLE}.sales_region = ${lightdash.attributes.sales_region}
+      sql_filter: ${TABLE}.sales_region IN(${lightdash.attributes.sales_region})
 ```
 
 ### Column filtering with `required_attributes`
@@ -205,7 +208,7 @@ In this example, the `tropical_fruits_column` will be visible to them because `c
 
 ### Row filtering
 
-If the user has group and user attribute values assigned, the user attribute values take precedence over group attribute values. If there are multiple group attribute values, there is no prioritization over which value will be used. 
+The template reference will be replaced by an array of the user's group or user attribute values.
 
 Let's walk through an example:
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/docs/references/user-attributes.mdx [hunk 1]
```diff
@@ -63,14 +63,17 @@ When referencing user attributes in SQL you can use the following [SQL variables
 
 You can use user attributes to filter the rows returned by a query. This is useful if you want to restrict the data 
 based on the user's attributes. To reference a user attribute in your sql, use the special lightdash reference 
-`${lightdash.attributes.<attribute_name> }`. For example, if you have a user attribute called `sales_region` you can
+`${lightdash.attributes.<attribute_name> }`.
+We recommend using the `IN` operator because the attribute might have multiple values.
```
owlas, line 67: Instead of recommend, say that this is the way to do it. The docs will be easier to read if it's opinionated!

docs/docs/references/user-attributes.mdx
```diff
@@ -63,14 +63,17 @@ When referencing user attributes in SQL you can use the following [SQL variables
...
+
+For example, if you have a user attribute called `sales_region` you can
 use it in your sql like this:
 
 ```yaml
 models:
   - name: my_model
     meta:
-      sql_filter: ${TABLE}.sales_region = ${lightdash.attributes.sales_region}
+      sql_filter: ${TABLE}.sales_region IN(${lightdash.attributes.sales_region})
```
owlas, line 76: typo missing space!

docs/docs/references/user-attributes.mdx
```diff
@@ -205,27 +208,16 @@ In this example, the `tropical_fruits_column` will be visible to them because `c
...
-
-The `${lightdash.attributes.fruit}` reference will equal `coconut`. User attribute values take precedence over group attribute values.
-
-**Scenario 2**:
-- The default attribute value is `banana`.
-- The user's been assigned no value.
-- They also have group values assigned of `kiwi` and `orange` 
-
-The `${lightdash.attributes.fruit}` reference will equal either `kiwi` or `orange. The order of multiple group attribute values is not defined. 
+In this example, the `${lightdash.attributes.fruit}` will be replaced with `'kiwi','orange','coconut'`.
```
owlas, line 222: great!

packages/backend/src/queryBuilder.ts
```diff
@@ -108,23 +108,27 @@ export const replaceUserAttributes = (
...
 
         return acc.replace(
             sqlAttribute,
-            `${stringQuoteChar}${userValue}${stringQuoteChar}`,
+            attributeValues
+                .map(
+                    (attributeValue) =>
+                        `${stringQuoteChar}${attributeValue}${stringQuoteChar}`,
+                )
+                .join(','),
```
owlas, line 131: would it look cleaner to have a space after the commas?

owlas: The attribute can only be set by admins so not a big concern now but we should be sql escaping this value, since it's user provided

owlas (approved):

PR #8770, opened 2024-01-30
fix: TableHeader empty tooltip on results

filp:
### Description:

Fixes the occasional empty tooltip, particularly if using table calculations as part of a result, when looking at result table header fields. This behavior applies to anything that isn't a field.

Before:

image

Now: 
image

(No tooltip is displayed)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
+++ b/packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
@@ -86,7 +86,7 @@ const TableHeader: FC<TableHeaderProps> = ({ minimal = false }) => {
                                                         label={tooltipLabel}
                                                         position="top"
                                                         disabled={
-                                                            tooltipLabel ==
+                                                            !tooltipLabel ||
                                                                 null ||
                                                             minimal ||
                                                             snapshot.isDropAnimating ||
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
```diff
@@ -78,15 +83,11 @@ const TableHeader: FC<TableHeaderProps> = ({ minimal = false }) => {
...
-                                                            meta?.item &&
-                                                            isField(meta?.item)
-                                                                ? meta.item
-                                                                      .description
-                                                                : undefined
-                                                        }
+                                                        label={tooltipLabel}
                                                         position="top"
                                                         disabled={
+                                                            tooltipLabel ==
```
joaoviana, line 89: perhaps this instead?
```suggestion
                                                            !tooltipLabel ||
```

in case it's `undefined`

joaoviana (changes requested):
one q!

packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
filp: The soft-comparison will cover `undefined` here, but in retrospect the `== null` pattern isn't super common in the rest of the codebase, so this is a good suggestion regardless - will include it!

filp:
Oops, auto-commit on the suggestion introduced an issue just now, will fix it.

joaoviana (approved):
🥳

PR #8771, opened 2024-01-30
feat: Support soft-wrapping in SqlEditor for table calculations

filp:
### Description:
Adds support for soft-wrapping while editing table calculation expressions.

Kapture 2024-01-30 at 14 33 30

Simply exposes the underlying ace editor feature, and should be straight-forward to include this behavior in an SqlRunner component follow-up.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
@@ -2,6 +2,7 @@ import {
     ActionIcon,
     Alert,
     Anchor,
+    Box,
     ScrollArea,
     Text,
     Tooltip,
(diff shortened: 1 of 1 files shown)


joaoviana (approved):
looks great!

packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
```diff
@@ -41,8 +48,42 @@ const SqlEditor = styled(AceEditor)<
...
+                isSoftWrapEnabled
+                    ? 'Disable editor soft-wrapping'
+                    : 'Enable editor soft-wrapping'
+            }
+            withArrow
+            position="left"
+        >
+            <ActionIcon onClick={onToggleSoftWrap}>
+                {isSoftWrapEnabled ? (
+                    <IconTextWrapDisabled size="1rem" />
```
joaoviana, line 75: non -blocking, but we usually use `MantineIcon` component for `icons`, like so: 
```tsx
<MantineIcon icon={IconTextWrapDisabled} />
```

packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
```diff
@@ -41,8 +48,42 @@ const SqlEditor = styled(AceEditor)<
...
+    onToggleSoftWrap: () => void;
+}> = ({ isSoftWrapEnabled, onToggleSoftWrap }) => (
+    <div
+        style={{
+            position: 'absolute',
+            bottom: 0,
+            right: 0,
+            // Avoids potential collision with ScrollArea scrollbar:
+            marginRight: '5px',
+        }}
```
joaoviana, line 62: another non-blocking, but you can use Mantine's `Box` and pass these styles as props

joaoviana (changes requested):
moving to requested so we keep using `MantineIcon` and enable global changes to the icon component in case we want to change the default size etc 👍

filp:
Thanks @joaoviana! Made a quick pass based on your feedback.

joaoviana (approved):
🚢 it!

PR #8775, opened 2024-01-30
fix: experiment with 2 approaches when taking screenshot of dashboard

joaoviana:
Closes: N/A

### Description:

Add 2 approaches here with Feature Flags for `element.screenshot`
1. `puppeteer-scroll-element-into-view`  - scroll into view of `element`. https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/api/ElementHandle.ts#L1239 
There was a _hidden_ feature on `element.screenshot()` - we can pass a `scrollIntoView` of element which will attempt a second `getBoundingBox` call if the element isn't _ready_. This is related to the `node has 0 height` error.

2. `puppeteer-set-viewport-dinamically` - https://github.com/puppeteer/puppeteer/issues/2423#issuecomment-590738707 Now this is another where screenshots don't work as expected. Could be a case where the element is not visible yet. This approach will: 
- Get the viewport size of the full page after charts are loading
- Wait for selector `.react-grid-layout` instead of `body` for dashboards. 

> [!IMPORTANT]  
> These 2 feature flags will only perform on `dashboards`. If they are successful, we can apply this to charts later. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -351,9 +351,9 @@ export class UnfurlService {
         const startTime = Date.now();
         let hasError = false;
 
-        const isPuppeteerSetViewportDinamicallyEnabled =
+        const isPuppeteerSetViewportDynamicallyEnabled =
             (await postHogClient?.isFeatureEnabled(
-                'puppeteer-set-viewport-dinamically',
+                'puppeteer-set-viewport-dynamically',
                 userUuid,
                 organizationUuid !== undefined
                     ? {
@@ -495,7 +495,7 @@ export class UnfurlService {
                             ? `[data-testid="visualization"]`
                             : 'body';
                     if (
-                        isPuppeteerSetViewportDinamicallyEnabled &&
+                        isPuppeteerSetViewportDynamicallyEnabled &&
                         lightdashPage === LightdashPage.DASHBOARD
                     ) {
                         selector = '.react-grid-layout';
@@ -506,7 +506,7 @@ export class UnfurlService {
                     });
 
                     if (
-                        isPuppeteerSetViewportDinamicallyEnabled &&
+                        isPuppeteerSetViewportDynamicallyEnabled &&
                         lightdashPage === LightdashPage.DASHBOARD
                     ) {
                         const fullPage = await page.$('.react-grid-layout');
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -562,10 +562,6 @@ export class UnfurlService {
                         });
                     }
 
-                    if (isPuppeteerScrollElementIntoViewEnabled) {
-                        await element.scrollIntoView();
-                    }
-
                     const imageBuffer = await element.screenshot({
                         path,
                         ...(isPuppeteerScrollElementIntoViewEnabled &&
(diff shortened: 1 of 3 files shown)


joaoviana:
testing in render too

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -465,10 +465,14 @@ export class UnfurlService {
                             ? `[data-testid="visualization"]`
                             : 'body';
 
-                    const element = await page.waitForSelector(selector, {
+                    await page.waitForNetworkIdle({
```
rephus, line 468: we already do this here... :thinking:  https://github.com/lightdash/lightdash/pull/8775/files#diff-9b86624a0467c9253f755c66928390e96818008cdd1df400c3509984ce4d1071L441 is this not redundant ?

joaoviana: changed approach ^ resolving this.

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -512,8 +562,18 @@ export class UnfurlService {
                         });
                     }
 
+                    if (isPuppeteerScrollElementIntoViewEnabled) {
+                        await element.scrollIntoView();
```
rephus, line 566: :+1:

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -460,15 +490,35 @@ export class UnfurlService {
...
+                        timeout: 60000,
                     });
 
+                    if (
+                        isPuppeteerSetViewportDynamicallyEnabled &&
+                        lightdashPage === LightdashPage.DASHBOARD
+                    ) {
+                        const fullPage = await page.$('.react-grid-layout');
+                        const fullPageSize = await fullPage?.boundingBox();
+                        await page.setViewport({
```
rephus, line 514: :+1:

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -512,8 +562,18 @@ export class UnfurlService {
...
+                    if (isPuppeteerScrollElementIntoViewEnabled) {
+                        await element.scrollIntoView();
+                    }
+
                     const imageBuffer = await element.screenshot({
                         path,
+                        ...(isPuppeteerScrollElementIntoViewEnabled &&
+                        lightdashPage === LightdashPage.DASHBOARD
+                            ? {
+                                  scrollIntoView: true,
```
rephus, line 574: why is scrollIntoView done twice ?

rephus (approved):
Looks good, good idea about adding these feature flags. 

Have you considered adding a bunch of verbose logs when this `node` error happens ? so we can try to understand why it is happening and replicate.

rephus (commented):
I can't `export dashboard`  on this preview, any idea why ? 

image

rephus (changes requested):
let's make sure this still works without feature flags before merging

packages/backend/src/services/UnfurlService/UnfurlService.ts
joaoviana: true, removing the first call

joaoviana:
https://us.posthog.com/project/39172/feature_flags/29553
https://us.posthog.com/project/39172/feature_flags/29559

Just tried one turned on, both turned on, and both turned off. 

on all scenarios the screenshot worked 👍

rephus (approved):

joaoviana:
adding logs later as suggested 👍

Issue #8724, opened 2024-01-25
Add ability to save custom visualizations


magnew:
### Description

Enable saving of custom visualizations

- [x] Vises can be saved with custom config
- [x] Saved vises can be duplicated and edited
- [x] Saved vises work in dashboards

PR #8776, opened 2024-01-30
feat: save custom visualizations

magnew:
Closes: #8724 

### Description:

Makes it possible to save custom visualisations. There is more to do here but now you can 
- Save visualizations
- Add them to dashboards
- Copy them and change them

The main things this review does are
- Remove the custom viz context and tie it into the main visualization context
- Remove restrictions on saving custom vizes
- rename VisualizationCustomConfigProps -> VisualizationCustomConfig
- Adds the custom viz type to the chart types table

Things that still need work (there are other features we could add but these would help complete this):
- Better error handling when things go wrong
- Some kind of validation
- Chart resizing could be better

I will file tickets.

Screenshot 2024-01-30 at 17 55 53

### Reviewer actions


packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -37,13 +37,13 @@ import VisualizationTableConfig, {
...
-    VisualizationConfigCustom,
-} from './VisualizationCustomConfigProps';
+    VisualizationCustomConfigType,
+} from './VisualizationCustomConfig';
 
 export type VisualizationConfig =
     | VisualizationConfigBigNumber
     | VisualizationConfigCartesian
-    | VisualizationConfigCustom
+    | VisualizationCustomConfigType
```
IrakliJani, line 46: `Type` suffix 👍

packages/frontend/src/hooks/useCustomVisualizationConfig.ts
```diff
@@ -0,0 +1,51 @@
...
+                key,
+                rowValue.value.raw,
+            ]),
+        );
+    });
+};
+
+export interface CustomVisualizationProps {
+    validConfig: CustomVis;
+    setVisSpec: (spec: string) => void;
```
IrakliJani, line 17: if this is a vega spec we can use type from vega? I think?

IrakliJani (commented):
some Qs. I'm gonna test this on render now

IrakliJani:
few notes from testing the render env:

- when switching to custom viz it starts with an errored state CleanShot 2024-01-31 at 17 20 30@2x

- maybe we should default to `{}` but then it's an empty screen CleanShot 2024-01-31 at 17 21 48@2x

- ooor maybe we should default to a bar chart like we do with pie bar chart. simple x/y bar chart should not be hard to generate CleanShot 2024-01-31 at 17 33 35@2x

- I should not be able to save a chart with an error 
CleanShot 2024-01-31 at 17 31 48@2x

packages/frontend/src/components/CustomVisualization/index.tsx
```diff
@@ -1,107 +1,60 @@
...
-type CustomVisualizationProps = {
+type Props = {
     className?: string;
     'data-testid'?: string;
 };
 
-const CustomVisualization: FC<CustomVisualizationProps> = (props) => {
-    const { chartConfig, rows } = useCustomVisualizationContext();
+const CustomVisualization: FC<Props> = (props) => {
+    // TODO: isSqlRunner
```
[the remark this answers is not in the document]
magnew: I don't know. That todo could be //TODO: does this need to work in the sqlRunner? :)

packages/frontend/src/hooks/useCustomVisualizationConfig.ts
magnew: It's not exactly. It's the vega spec but it's missing a couple things. Most notably 'data', which is required in vega. We inject that prop before passing this to Vega, but it makes sense not to store that prop at least. 

Good point about types though. Let me see if it's reasonably straight forward to use the vega type with a couple Omits or something.

magnew:
> few notes from testing the render env:
> 
> * when switching to custom viz it starts with an errored state CleanShot 2024-01-31 at 17 20 30@2x
> 
> * maybe we should default to `{}` but then it's an empty screen CleanShot 2024-01-31 at 17 21 48@2x
> 
> * ooor maybe we should default to a bar chart like we do with pie bar chart. simple x/y bar chart should not be hard to generate CleanShot 2024-01-31 at 17 33 35@2x
> 
> * I should not be able to save a chart with an error
>   CleanShot 2024-01-31 at 17 31 48@2x

All good points. I actually filed a ticket to make this better: #8791. How do you feel about doing some of this in a follow up? We might want Priyanka's input on the UI details. 

That said, I think you're right that starting in an error state is no good and saving the error state is no good. I might make it start with {} and be empty for now. And block saving on error. Sound good?

IrakliJani:
> That said, I think you're right that starting in an error state is no good and saving the error state is no good. I might make it start with {} and be empty for now. And block saving on error. Sound good?

yep, sounds good!

packages/frontend/src/components/CustomVisualization/index.tsx
```diff
@@ -1,107 +1,42 @@
...
+        !spec
+    ) {
+        return null;
     }
 
-    const data = { table: rows };
+    // TODO: 'chartConfig' is more props than config. It has data and
+    // configuration for the chart. We should consider renaming it generally.
+    const visProps =
+        visualizationConfig.chartConfig as CustomVisualizationConfigAndData;
```
IrakliJani, line 37: you don't need `as CustomVisualizationConfigAndData` because `if` condition above with `isCustomVisualizationConfig(visualizationConfig)` already narrows down the type for you

packages/frontend/src/components/LightdashVisualization/VisualizationCustomConfig.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const isCustomVisualizationConfig = (
+    visualizationConfig: VisualizationConfig | undefined,
+): visualizationConfig is VisualizationCustomConfigType => {
+    return visualizationConfig?.chartType === ChartType.CUSTOM;
+};
+
+type VisualizationCustomConfigProps =
+    VisualizationConfigCommon<VisualizationCustomConfigType> & {
+        // TODO: shared prop once all visualizations are converted
```
IrakliJani, line 22: do we use this in custom viz?

packages/frontend/src/hooks/useCustomVisualizationConfig.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Update object when spec changes
+    useEffect(() => {
+        try {
+            if (visSpec) {
+                setVisSpecObject(JSON.parse(visSpec));
+            }
+        } catch (e) {
+            //TODO: handle error
+        }
+    }, [visSpec]);
```
IrakliJani, line 53: I think there's a bug somewhere where this happens. happened when I was testing locally:

https://github.com/lightdash/lightdash/assets/962095/b2e44bc5-8e37-4105-abc8-acf139c553a3

IrakliJani (approved):
I found a minor bug and left a few Qs and suggestions.

but, not a blocker

packages/frontend/src/components/LightdashVisualization/VisualizationCustomConfig.tsx
magnew: No, but I thought it was a good idea to keep it around cause that comment implies we want to use it everywhere.

packages/frontend/src/hooks/useCustomVisualizationConfig.ts
magnew: Good find. I will fix this in a follow up if you dont mind. In some ways its a feature -- this spec cannot start with any character except '{' anyway. But I will fix.

Issue #8738, opened 2024-01-26
The 'Save' button on Settings forms should be inactive until a change has actually been made


PriPatel:
<img width="908" alt="Screenshot 2024-01-26 at 10 07 51" src="https://github.com/lightdash/lightdash/assets/12776522/a63cce60-ad7e-43f8-a58f-ca42331b9549">

The `Save` button here is active even though I've made no changes.

**Problems**
- When I do a make a change, the state of the button doesn't change, so I'm not sure if my changes have saved(!)
- When I want to undo my changes, I feel like I need to refresh to cancel them
- On a page with many forms, there are many active primary CTAs - it's information overload. 

**Proposal**
- [ ] The `Save` Button should be disabled until a change has been made
- [ ] Once a change has been made, there should be an option to cancel that change. When I click on this, the changes revert and the `Save` button goes back to being disabled

The **Appearance** tab in the Settings Page already does this correctly. Follow this pattern!

**Areas of impact in the form settings**

- [ ] User settings - **Profile form**
- [ ] User settings -  **Password form** 
- [ ] Project - **Tables Configuration form** 
- [ ] Organizations - **General form**
- [ ] Organizations - **Embedding configuration form**

jogeshwar01:
@PriPatel Can I work on this?

Also, I just wanted to check on the ` Organizations - Embedding configuration form`. It doesn't seem to be present in the code.
I do see it in the docs - https://docs.lightdash.com/references/embedding/. Is the code for this closed-source for now?

PR #8779, opened 2024-01-30
feat: button inactive until change

jogeshwar01:
Closes: #8738 

### Description:

Added code to keep save/update buttons inactive until a change is made. Also includes cancel button to revert back to original state.

- [x] The Save Button should be disabled until a change has been made
- [x]  Once a change has been made, there should be an option to cancel that change. When I click on this, the changes revert and the Save button goes back to being disabled

Areas of impact in the form settings
- [x] User settings - Profile form
- [x] User settings - Password form
- [x] Project - Tables Configuration form
- [x] Organizations - General form
- [ ]  Organizations - Embedding configuration form

Just wanted to check on the `Organizations - Embedding configuration form`.
I do see it in the docs - https://docs.lightdash.com/references/embedding/. Is the code for this closed-source for now?

Output:

Profile Form -
Before:

image

On Change:

image

Table Configuration Form-
Before:

image

On Change:

image

General Form-
Before:

image

On Change:

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/ProjectTablesConfiguration/ProjectTablesConfiguration.tsx
+++ b/packages/frontend/src/components/ProjectTablesConfiguration/ProjectTablesConfiguration.tsx
@@ -144,6 +144,45 @@ const ProjectTablesConfiguration: FC<{
         }
     }, [isSuccess, onSuccess]);
 
+    const setFormValuesFromData = useCallback(() => {
+        if (data) {
+            form.setValues({
+                type: data.tableSelection.type,
+                tags:
+                    data.tableSelection.type === TableSelectionType.WITH_TAGS &&
+                    data.tableSelection.value
+                        ? data.tableSelection.value
+                        : [],
+                names:
+                    data.tableSelection.type ===
+                        TableSelectionType.WITH_NAMES &&
+                    data.tableSelection.value
+                        ? data.tableSelection.value
+                        : [],
+            });
+
+            form.resetDirty({
+                type: data.tableSelection.type,
+                tags:
+                    data.tableSelection.type === TableSelectionType.WITH_TAGS &&
+                    data.tableSelection.value
+                        ? data.tableSelection.value
+                        : [],
+                names:
+                    data.tableSelection.type ===
+                        TableSelectionType.WITH_NAMES &&
+                    data.tableSelection.value
+                        ? data.tableSelection.value
+                        : [],
+            });
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [data]);
+
+    useEffect(() => {
+        setFormValuesFromData();
+    }, [setFormValuesFromData]);
+
     const handleSubmit = form.onSubmit(async (formData: FormData) => {
         track({
             name: EventName.UPDATE_PROJECT_TABLES_CONFIGURATION_BUTTON_CLICKED,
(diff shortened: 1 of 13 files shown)


PriPatel:
@jogeshwar01 
Screenshot 2024-01-30 at 18 51 01
Thanks for this! For consistency and usability, would it be possible to have the buttons like the example above where they are next to eachother ? (instead of at opposite ends?) 

Thanks again!!

jogeshwar01:
@PriPatel Sure, will update the alignment.

jogeshwar01:
@PriPatel Fixed the alignment and updated the output screenshots.

PriPatel (commented):
Screenshots looks good, we're just going to test the behaviour in a mirror PR!

packages/frontend/src/components/UserSettings/PasswordPanel/index.tsx
```diff
@@ -53,6 +53,22 @@ const PasswordPanel: FC = () => {
...
 
+    const setFormValuesFromData = useCallback(() => {
+        form.setValues({
+            currentPassword: '',
+            newPassword: '',
+        });
+        form.resetDirty({
+            currentPassword: '',
+            newPassword: '',
+        });
```
IrakliJani, line 64: hey @jogeshwar01 why do we need to set default value to the form and then resetDirty? can you explain how this works?

IrakliJani: this seems to me that it's serving two purposes, one is initializing the data and another one is resetting `dirty` state. I think this should be called `handleFormReset` and should be called from cancel button only. LMK!

packages/frontend/src/components/ProjectTablesConfiguration/ProjectTablesConfiguration.tsx
```diff
@@ -136,6 +137,45 @@ const ProjectTablesConfiguration: FC<{
...
+                    data.tableSelection.value
+                        ? data.tableSelection.value
+                        : [],
+                names:
+                    data.tableSelection.type ===
+                        TableSelectionType.WITH_NAMES &&
+                    data.tableSelection.value
+                        ? data.tableSelection.value
+                        : [],
+            });
```
IrakliJani, line 170: this seems repetitive. can we have a one object and use that to call both of the functions? more questions about those two functions below

IrakliJani (changes requested):
thanks for the contribution @jogeshwar01 I added few questions in the PR. please, have a look and let me know! thanks 🙌

packages/frontend/src/components/ProjectTablesConfiguration/ProjectTablesConfiguration.tsx
jogeshwar01: @IrakliJani  Thanks for the review, will update this to remove the repetition.

packages/frontend/src/components/UserSettings/PasswordPanel/index.tsx
jogeshwar01: @IrakliJani The `form.setValues` is used to track the current values in the form while the `form.resetDirty` controls our current state that also helps us to display/hide the Cancel button on change

For instance, once a user clicks on `Save changes/Update` , we would want to set both our form values as well as the dirty state to the current changed state (to prevent Cancel button from showing after state has been updated)

So basically, we need this to be called in 2 cases - either user Saves the changes (useEffect) or cancels them.

jogeshwar01:
@IrakliJani Fixed the redundancy in the code and changed the function name to `handleFormReset` for `PasswordPanel`.

packages/frontend/src/components/UserSettings/PasswordPanel/index.tsx
jogeshwar01: In case of Password Panel, as we redirect to /login when we update the password, we call it from the Cancel button only.
There are no dependencies for the `handleFormReset useCallback` and so it would never go inside the useEffect except  for the first time.

IrakliJani:
hi @jogeshwar01

I've made few additional changes to this PR:
- upgraded to mantine form v7 which introduces `setInitialValues` and simplifies the logic
- use ZOD for validation in all of the forms

waiting for e2e tests before I merge this. thanks for your work 🙌

IrakliJani (approved):

PR #8780, opened 2024-01-30
feat: Add additional attribute for # of table calculations in `ProjectService.runMetricQuery` trace

filp:
### Description:

- Includes support for a new `new-table-calculations-engine` feature flag
- When enabled, includes a new span attribute for `ProjectService.runMetricQuery`:
  - `tableCalculationsNum`: number of table calculations for this query

The goal is to measure the impact of upcoming table calculation changes on metrics queries. The span attribute will be moved outside the feature flag condition in a follow-up PR, to allow comparison testing (with/without new table calculations changes).

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1558,6 +1558,19 @@ export class ProjectService {
                         );
                     }
 
+                    const useNewTableCalculationsEngine =
+                        (await postHogClient?.isFeatureEnabled(
+                            'new-table-calculations-engine',
+                            user.userUuid,
+                            user.organizationUuid !== undefined
+                                ? {
+                                      groups: {
+                                          organization: user.organizationUuid,
+                                      },
+                                  }
+                                : {},
+                        )) ?? false;
+
                     const { organizationUuid } =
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1680,6 +1693,17 @@ export class ProjectService {
 
                     Logger.debug(`Fetch query results from cache or warehouse`);
                     span.setAttribute('generatedSql', query);
+
+                    /**
+                     * If enabled, we include additional attributes for this span allowing us to measure
```
owlas, line 1698: do you also want a boolean to indicate the flag explicitly? Rather than having "if tableCalculationsNum is there then we are using new table calculations"?

owlas (approved):

packages/backend/src/services/ProjectService/ProjectService.ts
filp: Definitely, was going to add that in a follow-up but there's really no reason not to include it here - `tableCalculationsNum` will stop being an indicator of the feature flag once we're measuring the impact of any changes.

PR #8795, opened 2024-01-31
feat: Add `isFeatureFlagEnabled` utility

filp:
### Description:

- Adds a new `isFeatureFlagEnabled` helper on the backend
- Adds a new `FeatureFlags` enum to `common`
- Updates feature flag checks in `ProjectService` to use this new utility

Example usage:

```ts
const useDbtLs = await isFeatureFlagEnabled(
    FeatureFlags.UseDbtLs,
    user,
);
```

`FeatureFlags` will be extended in a follow-up, and an equivalent utility included in frontend.

### Reviewer actions


packages/backend/src/services/FeatureFlagService/FeatureFlags.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Before using a FeatureFlag, include it as a member in this enum,
+ * to ensure you can use it with FeatureFlagService. Consider adding
+ * a short description of the feature flag and how it will be used.
+ *
+ * If the feature flag is no longer in use, remove it from this enum.
+ */
+export enum FeatureFlags {
```
magnew, line 8: Should we expose this enum in common so it can be used in the front end? We'd need to build some equivalent context or something equivalent to this service on the FE to make good use of it. 

We sometimes have flags that are used in both places. If you think it's better to keep the BE and FE enums separate for any reason, I can get on board with that too. But if one of the goals is to keep track of which of these exist better, sharing this enum as a source of truth for that could be a good idea.

filp: Definitely! I left frontend as an exercise for the future since I'm not yet familiar with house we're using feature flags there, but I fully agree on principle, and there's no good reason to not include the enum in `common` from the start, even if there's some lag on actually using it in frontend - I'll make that change!

filp:
@owlas I don't have strong opinions on exposing this as a service vs a utility method, so happy to make that change!

One counter argument to consider, which may not be relevant at all, this service becomes a convenient place to introduce flag-specific logic for more complex checks, for example:

```ts
class FeatureFlagService {
  // ...
  export async function isUsingSomeAdvancedFeature() {
    const flagA = ...
    const flagB = ...

    return flagA && !flagB;
  }
}
```

Which is a pattern I've had a use before, and has some nice benefits for readability/type safety.

Would love to know what you think, and if the above seems useful and/or changes anything about your previous opinion. Like I said, no super strong feelings on my end, so definitely happy to go with the simpler solution here!

packages/backend/src/services/FeatureFlagService/FeatureFlags.ts
magnew: Perfect, yea we dont need to use it there for now, but making the enum shared would be good. 👍

owlas:
I can't think of a use case for it currently!

filp:
@magnew Updated the description and included `FeatureFlags` in `common`

magnew (approved):
👍

owlas (approved):

packages/backend/src/postHog.ts
```diff
@@ -6,3 +7,38 @@ export const postHogClient = lightdashConfig.posthog.projectApiKey
...
+    flag: FeatureFlags,
+    user: Pick<SessionUser, 'userUuid' | 'organizationUuid'>,
+    {
+        postHog = postHogClient,
+    }: {
+        postHog?: PostHog;
+    } = {},
+): Promise<boolean> {
+    /** If we don't have a PostHog client instance, we return false for all checks */
+    if (!postHog) {
```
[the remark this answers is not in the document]
filp: The global instance is assigned as the default - makes it easier to mock later on if necessary.

PR #8802, opened 2024-01-31
chore: fix dom nesting warning in create dashboard

magnew:
### Description:

Opening the Create Dashboard modal causes a console error about nesting that can be annoying. Wrapping the modal title in a Box seems to fix the error. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
+++ b/packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
@@ -1,6 +1,7 @@
 import { Dashboard, Space } from '@lightdash/common';
 import {
     ActionIcon,
+    Box,
     Button,
     Group,
     MantineProvider,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/modal/DashboardCreateModal.tsx
```diff
@@ -133,7 +134,11 @@ const DashboardCreateModal: FC<DashboardCreateModalProps> = ({
     return (
         <MantineProvider inherit theme={{ colorScheme: 'light' }}>
             <Modal
-                title={<Title order={5}>Create dashboard</Title>}
+                title={
+                    <Box>
```
ZeRego, line 138: If I was editing this component and saw this I would remove the `Box` component as I don't know why it's there.
I assume this is a potential problem on all modals. Can we change the Modal component to have a `div` instead of a `h2` for the title?

magnew: Yea, that's a fair point -- this is pretty unintuitive. It's basically a bug in mantine that this is working around. The docs say this title prop takes a react node, but it doesn't replace the h2 in the title, it makes what gets passed here a child of that h2. 

I checked if there is a way to override the title type and there is not. The whole header can be overriden, but then you lose the close button. 

A solution to this that solves these concerns is to implement our own modal sub-components like ModalHeader, ModalBody, ModalFooter. I think we should do something like that anyway for consistency, but it's a bigger thing. 

My inclination would be to fix these as 1-offs until we do that. The good news is that if someone does remove this Box or forget it, the result is a console error but things still work. If you'd prefer to keep the console errors and not introduce this kind of workaround, we can close this.

ZeRego (approved):

Issue #8767, opened 2024-01-30
Standardise formatting logic to be used across all types of fields


joaoviana:
### Description

Refactor `formatting.ts` in `common` to utilise the same formatting logic for the following field types:
* Metrics
* Dimensions
* Custom Metrics
* Table Calculations

Currently, Table Calculations uses a different util to display formatted values: `formatTableCalculationValue` while the other fields use `formatFieldValue`. 

Since we are allowing more formatting options to Custom Metrics + these align with current Table Calculations formatting options + Custom metrics are converted to metrics, we should
* Remove table-calculation-specific formatting utils
* Refactor current formatting functions to be used by all field types
* Still ensure that the `yml` configuration and metadata isn't broken
* Less field-type-specific types and reuse them wherever possible. Example: `Format` type should be used for the types: `CustomMetricFormat` and `TableCalculationFormat` (this one I'm not too sure, TBD after discovery)

### What problem does this solve?

DevX

PR #8807, opened 2024-01-31
feat: unify legacy Format with CustomFormat functions

almeidabbm:
Closes: #8767 

### Description:

- Unifies the legacy formatting functions used for Metrics and Dimensions with the TableCalculation ones adhering to CustomFormat type
- Adds a convert function from legacy Format to CustomFormat
- Standardises rounding so that when round is undefined, then up to 3 decimal places are kept
- Adds support for negative rounding to Metrics and Dimensions. Previously it was just available for TableCalculations
- Fixes the tests

### Reviewer actions

### DIFF
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -120,6 +120,7 @@ export enum CustomFormatType {
     PERCENT = 'percent',
     CURRENCY = 'currency',
     NUMBER = 'number',
+    ID = 'id',
 }
 
 export type TableCalculation = {
@@ -139,7 +140,7 @@ export interface TableCalculationField extends Field {
     sql: string;
 }
 
-export const isTableCalculation = (item: Item): item is TableCalculation =>
+export const isTableCalculation = (item: any): item is TableCalculation =>
     item
         ? !('binType' in item) &&
           !!item.sql &&
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -140,7 +140,9 @@ export interface TableCalculationField extends Field {
     sql: string;
 }
 
-export const isTableCalculation = (item: any): item is TableCalculation =>
+export const isTableCalculation = (
+    item: Item | AdditionalMetric,
+): item is TableCalculation =>
     item
         ? !('binType' in item) &&
           !!item.sql &&
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -8,405 +8,812 @@ import {
     NumberSeparator,
 } from '../types/field';
 import {
+    applyCustomFormat,
     currencies,
-    formatFieldValue,
     formatItemValue,
-    formatTableCalculationNumber,
-    formatTableCalculationValue,
-    formatValue,
+    formatNumberValue,
+    getCustomFormatFromLegacy,
 } from './formatting';
 import { dimension, metric, tableCalculation } from './formatting.mock';
 
 describe('Formatting', () => {
-    describe('format and round', () => {
-        test('formatValue should return the right format', async () => {
+    describe('convert legacy Format type', () => {
+        [Format.KM, Format.MI].forEach((format) => {
+            describe(`when it is ${format.toUpperCase()}`, () => {
+                test('getCustomFormatFromLegacy should return the correct CustomFormat options', () => {
+                    expect(getCustomFormatFromLegacy({ format })).toEqual({
+                        type: CustomFormatType.NUMBER,
+                        suffix: ` ${format}`,
+                        compact: undefined,
+                        round: undefined,
+                    });
+                });
+            });
+        });
+
+        [Format.EUR, Format.GBP, Format.USD].forEach((format) => {
+            describe(`when it is ${format.toUpperCase()}`, () => {
+                test('getCustomFormatFromLegacy should return the correct CustomFormat options', () => {
+                    expect(getCustomFormatFromLegacy({ format })).toEqual({
+                        type: CustomFormatType.CURRENCY,
+                        currency: format.toUpperCase(),
+                        compact: undefined,
+                        round: undefined,
+                    });
+                });
+            });
+        });
+
+        describe(`when it is ${Format.PERCENT.toUpperCase()}`, () => {
+            test('getCustomFormatFromLegacy should return the correct CustomFormat options', () => {
+                expect(
+                    getCustomFormatFromLegacy({ format: Format.PERCENT }),
+                ).toEqual({
+                    type: CustomFormatType.PERCENT,
+                    compact: undefined,
+                    round: undefined,
+                });
+            });
+        });
+
+        describe(`when it is ${Format.ID.toUpperCase()}`, () => {
+            test('getCustomFormatFromLegacy should return the correct CustomFormat options', () => {
+                expect(
+                    getCustomFormatFromLegacy({ format: Format.ID }),
+                ).toEqual({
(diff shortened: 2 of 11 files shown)


packages/e2e/cypress/e2e/app/formats.cy.ts
```diff
@@ -98,7 +98,7 @@ describe('Explore', () => {
             '£1,999,000',
             '1,999,000 km',
             '1,999,000 mi',
-            '199900000%',
+            '199,900,000%',
```
almeidabbm, line 101: Since now we're using the same formatting function for everything, this included the separators for percent metrics. I believe the goal was to unify everything so the test was changed to match the new format. Do let me know if this should be different 😁

packages/common/src/types/field.ts
```diff
@@ -139,7 +140,7 @@ export interface TableCalculationField extends Field {
     sql: string;
 }
 
-export const isTableCalculation = (item: Item): item is TableCalculation =>
+export const isTableCalculation = (item: any): item is TableCalculation =>
```
joaoviana, line 143: can't we keep `Item`? forgot 😓

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -54,39 +55,42 @@ const formatComparisonValue = (
     }
     switch (format) {
         case ComparisonFormatTypes.PERCENTAGE:
-            return `${prefix}${formatValue(value, {
-                format: Format.PERCENT,
+            return `${prefix}${applyCustomFormat(value, {
                 round: 0,
+                type: CustomFormatType.PERCENT,
```
joaoviana, line 60: is there a way we can pass the prefix to the `applyCustomFormat`?

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -54,39 +55,42 @@ const formatComparisonValue = (
...
             })}`;
         case ComparisonFormatTypes.RAW:
             if (item !== undefined && isTableCalculation(item)) {
-                return `${prefix}${formatTableCalculationValue(
-                    item.format,
-                    value,
-                )}`;
+                return `${prefix}${formatItemValue(item, value)}`;
```
joaoviana, line 64: same here ^

packages/frontend/src/hooks/useBigNumberConfig.ts
```diff
@@ -54,39 +55,42 @@ const formatComparisonValue = (
...
-            return `${prefix}${formatValue(value, {
-                format: isField(item) ? item.format : undefined,
-                round: bigNumberComparisonStyle
-                    ? 2
-                    : isField(item)
-                    ? item.round
-                    : undefined,
-                compact: bigNumberComparisonStyle,
-            })}`;
+            return `${prefix}${applyCustomFormat(
```
joaoviana, line 66: and here

packages/common/src/utils/formatting.test.ts
```diff
@@ -1,412 +1,883 @@
...
+                    type: CustomFormatType.ID,
+                });
+            });
+        });
+    });
+
+    describe('applying CustomFormat to value', () => {
+        describe('when using legacy format', () => {
+            describe('if Format is distance unit', () => {
+                test('it should return the right format', () => {
```
joaoviana, line 72: nit-pick: can just have one `test` and remove the extra `describe` in `71` 
```
test('returns formatted value if distance-formatting') 
```
or something similar 😅

packages/common/src/utils/formatting.test.ts
```diff
@@ -1,412 +1,883 @@
...
+                        applyCustomFormat(
+                            5.9999999,
+                            getCustomFormatFromLegacy(roundTenOptions),
+                        ),
+                    ).toEqual('5.9999999000');
+                });
+            });
+
+            describe('when round is negative number', () => {
+                test('it should return the right round', () => {
```
joaoviana, line 325: 🙌

packages/common/src/utils/formatting.test.ts
```diff
@@ -1,412 +1,883 @@
...
+                            }),
+                        ),
+                    ).toEqual('$0.00K');
+
+                    expect(
+                        applyCustomFormat(
+                            4,
+                            getCustomFormatFromLegacy({
+                                compact: K,
+                                round: 3,
```
joaoviana, line 690: suggestion: 
test this
```ts
 getCustomFormatFromLegacy({
                                compact: K,
                               
                                format: Format.USD,
                            }),
```
basically, no `round` set

packages/common/src/utils/formatting.test.ts
```diff
@@ -1,412 +1,883 @@
...
+                    applyCustomFormat(
+                        1000000000,
+                        getCustomFormatFromLegacy({ compact: 'billion' }),
+                    ),
+                ).toEqual('1B');
+
+                expect(
+                    applyCustomFormat(
+                        1000000000000,
+                        getCustomFormatFromLegacy({ compact: 'T' }),
```
joaoviana, line 836: to check if this test exists:
* compact 
* round < 0

packages/common/src/utils/formatting.ts
```diff
@@ -372,11 +202,97 @@ export function formatTableCalculationNumber(
     }
 }
 
-export function formatTableCalculationValue(
-    format: CustomFormat | undefined,
+function defaultFormat(value: unknown) {
```
joaoviana, line 205:
```suggestion
function applyDefaultFormat(value: unknown) {
```

packages/common/src/utils/formatting.ts
```diff
@@ -146,211 +141,46 @@ export function isNumber(value: unknown): value is number {
...
-        : { maximumFractionDigits: round, minimumFractionDigits: round };
-
-    if (isValidCurrencyFormat) {
-        return new Intl.NumberFormat('en-US', {
-            style: 'currency',
-            currency: format?.toUpperCase(),
-            ...validFractionDigits,
-        }).format(Number(value));
-    }
+    const getFormatOptions = () => {
```
joaoviana, line 148: to consider
* could we move this out of this function?

joaoviana (commented):
💪 
Will test on Render now

packages/frontend/src/hooks/useBigNumberConfig.ts
almeidabbm: We only support adding `prefix` via format to `CustomFormatType.NUMBER` types. This one is taking in a `CustomFormatType.PERCENT` so it is not possible.

packages/frontend/src/hooks/useBigNumberConfig.ts
almeidabbm: Same reason, but here we're passing `item.format` (legacy) which isn't necessarily percent.

packages/frontend/src/hooks/useBigNumberConfig.ts
almeidabbm: Same

joaoviana (approved):
🎉

packages/common/src/utils/formatting.ts
```diff
@@ -131,226 +121,66 @@ export function formatTimestamp(
...
-): string {
-    if (value === null) return '∅';
-    if (value === undefined) return '-';
-    if (!field) {
-        return `${value}`;
-    }
-    const { type, round, format, compact } = field;
-
-    switch (type) {
-        case DimensionType.STRING:
```
joaoviana, line 266: here

PR #8856, opened 2024-02-07
feat: add shared chart color support under feature flag

filp:
### Description:
Taken from the method comment:

```ts
    /**
     * Given the org's color palette, and a dimension identifier, return the color
     * in the palette assigned to that identifier.
     *
     * This works by hashing the identifier into an integer value, and then projecting
     * that value into the org's color space - effectivelly getting a number from 0 to
     * <number of colors in palette>.
     *
     * This is a straight-forward way to handle hashing dimensions into colors, with
     * two major caveats:
     *
     * - We're no longer cycling over colors in the palette, potentially not following
     *   an intentionally-designed best-neighbor-color approach.
     * - We have no guarantee different dimensions won't be assigned the same color due
     *   to the narrow color space - we can fix this by shifting colors aside when
     *   compiling the chart config.
     */
```

tl;dr: Hooks into `useCartesianChartConfig` to add initial support for mapping colors in a predictable manner, based on dimensions. This is likely to require additional adjustments and support for different chart configurations, but this PR allows testing against real data via a feature flag.

### Reviewer actions

### DIFF
--- a/packages/common/src/types/featureFlags.ts
+++ b/packages/common/src/types/featureFlags.ts
@@ -10,4 +10,11 @@ export enum FeatureFlags {
      * See  ProjectService for example usage.
      */
     UseDbtLs = 'use-dbt-ls',
+
+    /**
+     * Use shared color assignments for cartesian-type charts, when possible. This
+     * essentially means we try to have the same dimension translate into the same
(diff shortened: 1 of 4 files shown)


magnew (approved):
👍

Issue #8855, opened 2024-02-07
Table calculation round with no decimals is not working 


ZeRego:
### Description

When we set round to `0` it doesn't work.

### To Reproduce

<img width="796" alt="Screenshot 2024-02-07 at 08 58 00" src="https://github.com/lightdash/lightdash/assets/9117144/2e116d68-aa9f-4adf-a4a2-1cbe5e36795e">
<img width="799" alt="Screenshot 2024-02-07 at 08 57 54" src="https://github.com/lightdash/lightdash/assets/9117144/e93c661d-7435-4cc8-87ab-a67cb76ffbb7">
<img width="1004" alt="Screenshot 2024-02-07 at 08 57 48" src="https://github.com/lightdash/lightdash/assets/9117144/5389cb08-4429-40dd-ad14-7a2071e9a78e">

PR #8862, opened 2024-02-07
fix: round with no decimals in table calculations

jogeshwar01:
Closes: #8855 

### Description:

The issue was due to the `maximumSignificantDigits` that we had set for case `round <= 0.`
When `round = 0`, `maximumSignificantDigits` was set to 1. This worked fine for numbers greater than 1.
But for cases like 0.001, the only significant digit ie. 1 is being taken into account and we get 0.001 as output even if round was set to 0.

Fixed by removing the = from the if condition. Added a test for the same.

image

image

image

### Reviewer actions


almeidabbm (approved):
LGTM 🎉

Issue #3195, opened 2022-09-08
I want to be able to access TV mode from the dashboard or chart header


PriPatel:
#3195 - users should be able to access the TV mode for a chart/dashboard from the header. 

As a business user, there are elements of the app that Im not really interested in seeing when browsing dashboards. Seeing the dashboard in full-screen without the build tools / actions / navbar would be ideal. I guess this would help in presentations as well.

**MVP**

Add a 'View in TV mode' option under the context menu. Use a 'screen icon'
When user hits the TV mode, Lightdash navigation disappears (for now)
A message reminding users to hit 'esc' to leave TV mode appears for a few seconds.
<img width="324" alt="Screenshot 2023-02-23 at 14 05 09" src="https://user-images.githubusercontent.com/12776522/220934184-e39f532a-fb85-4864-b418-31d64951049d.png">

IrakliJani:
zen mode

PriPatel:
🧘‍♀️, exactly. No distractions pls.

hamzahc1:
I like this - I think hiding the app navbar, the edit dashboard buttons and displaying the new dashboard filters would be a vibe ☮️ (You could even move the Title so that in TV mode the charts start higher up the page)

12ian34:
A prospect has requested that the TV mode includes a configurable refresh schedule

PriPatel:
we could have something like this in the header
<img width="339" alt="Screenshot 2023-02-08 at 14 27 09" src="https://user-images.githubusercontent.com/12776522/217503027-46a05b4d-a893-4bd1-b4e2-340b25794065.png">

ZeRego:
Could we add the auto-refresh UI and functionality in a separate PR ? We have an open ticket for that btw https://github.com/lightdash/lightdash/issues/3693 ?

12ian34:
This isn't stale! We'll definitely be working on this at some point!

PR #8874, opened 2024-02-08
feat: TV/Fullscreen mode for dashboards

filp:
### Description:
- closes: #3195 
- Adds a TV/Fullscreen mode for dashboards:

image

- Hides navbar when in fullscreen mode:
image

Additionally, correctly handles browser-native fullscreen controls (e.g if the user exits fullscreen manually, we also disable TV mode in Lightdash).

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
+++ b/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
@@ -14,6 +14,8 @@ import {
     Tooltip,
 } from '@mantine/core';
 import {
+    IconArrowsMaximize,
+    IconArrowsMinimize,
     IconCheck,
     IconChevronRight,
     IconCopy,
@@ -66,6 +68,7 @@ type DashboardHeaderProps = {
     hasDashboardChanged: boolean;
     isEditMode: boolean;
     isSaving: boolean;
+    isFullscreen: boolean;
     oldestCacheTime?: Date;
     onAddTiles: (tiles: Dashboard['tiles'][number][]) => void;
     onCancel: () => void;
@@ -74,6 +77,7 @@ type DashboardHeaderProps = {
     onDuplicate: () => void;
     onMoveToSpace: (spaceUuid: string) => void;
     onExport: () => void;
+    onToggleFullscreen: () => void;
 };
 
 const DashboardHeader = ({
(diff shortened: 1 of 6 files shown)


joaoviana:
Relates to #3195

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
```diff
@@ -267,7 +273,32 @@ const DashboardHeader = ({
                 <PageActionsContainer>
                     {userCanExportData && <DashboardRefreshButton />}
 
-                    {!!userCanManageDashboard && (
+                    {!isEditMode && document.fullscreenEnabled && (
```
joaoviana, line 276: can we use 
```suggestion
                    {!isEditMode && isFullScreen && (
``` 
?

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -273,6 +275,40 @@ const Dashboard: FC = () => {
...
+        };
+
+        document.addEventListener('fullscreenchange', onFullscreenChange);
+
+        return () =>
+            document.removeEventListener(
+                'fullscreenchange',
+                onFullscreenChange,
+            );
+    });
```
joaoviana, line 310: this is great, but we can also use this v useful hook 😉 https://mantine.dev/hooks/use-fullscreen/

joaoviana (approved):
Works great! 

Another thing to consider (can be on a separate PR):
1. Go to dashboard
2. Enter full screen mode
3. Click `Explore from here` on a chart tile
4. User gets redirected to a chart view 
5. Full screen remains

Not too sure on what we expect the behaviour to be like but I would probably exit Full screen mode on route change 👍

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
filp: The check on `document.fullscreenEnabled` is used to hide the button if fullscreen isn't available in this user agent, despite the super confusing name 😵‍💫

packages/frontend/src/pages/Dashboard.tsx
filp: Awesome, thanks for the heads up -- I keep getting surprised by features built into mantine 😄 I'll refactor with this hook.

filp:
Merging as-is and following up with small improvements in another PR - using `useFullscreen` from mantine is actually not super straight-forward since it introduces some wonky behavior (tl;dr: the extra indirection compared to this implementation can cause render-loop-related issues when figuring out if/when to go/exit fullscreen).

Issue #8834, opened 2024-02-02
Compile error when exploring from a dashboard chart


ZeRego:
### To Reproduce

go to dashboard with filters
on a chart tile, click "explore from here" 

<img width="1404" alt="Screenshot 2024-02-02 at 11 51 07" src="https://github.com/lightdash/lightdash/assets/9117144/cf5d5e79-9a62-4029-8efb-e669da928911">

Problem is an API validation error:
The MetricQuery type doesn't support the dashboard filter properties.

PR #8883, opened 2024-02-09
fix: allow additional properties tsoa

rephus:
Closes: https://github.com/lightdash/lightdash/issues/8834

### Test

- Go to jaffle dashboard 
- explore from here on a chart
- show sql code 
- it should not fail now 

### Reviewer actions

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -575,7 +575,7 @@ const models: TsoaRoute.Models = {
             settings: { dataType: 'any' },
             disabled: { dataType: 'boolean' },
         },
-        additionalProperties: false,
+        additionalProperties: true,
     },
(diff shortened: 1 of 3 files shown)


rephus:
Another option, instead of making this global change. We could enable additional properties in the body of some endpoints, like this: 

```
import { Schema } from '@tsoa/runtime';

export interface MetricQuery {
    // Your existing properties here
    [key: string]: any; // This allows additional properties
}

export const MetricQuerySchema: Schema = {
    properties: {
        // Define your existing properties here
    },
    additionalProperties: true, // Enable additional properties
    required: [], // You can specify required properties if needed
};

```

ZeRego (approved):

rephus:
Tested : 
open sql on `explore from here` chart, and it fixed the issue

image

Issue #8773, opened 2024-01-30
change --use-dbt-list to be the default behaviour in the CLI


ZeRego:
We have seen a big speed improvement when using the option --use-dbt-list and haven't heard about issues. 

We should apply this option by default.

If possible, we should allow `--use-dbt-list=false` in case people have issues and need to use `dbt compile`

PR #8889, opened 2024-02-11
feat: set --use-dbt-list as default in cli

jogeshwar01:
Closes: #8773 

### Description:

- Set -`-use-dbt-list` option true as default in cli
- Allow the user to use `--use-dbt-list=false` in case they need to use dbt compile
- Setting it to any value other than `true` or `false` would set it back to true

Example - 

The following commands will consider --use-dbt-list to true.
```
node <index.js path> deploy --create --profiles-dir .
node <index.js path> deploy --create --profiles-dir . --use-dbt-list
node <index.js path> deploy --create --profiles-dir . --use-dbt-list=true
node <index.js path> deploy --create --profiles-dir . --use-dbt-list true
```

Output - 

image

For setting it to false, we can use -
```
node <index.js path> deploy --create --profiles-dir . --use-dbt-list=false
node <index.js path> deploy --create --profiles-dir . --use-dbt-list false
```

Output -

image

### Reviewer actions

### DIFF
--- a/.github/workflows/cli-tests.yml
+++ b/.github/workflows/cli-tests.yml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dbt: ['1.3', '1.4', '1.5', '1.6', '1.7']
+        dbt: ['1.4', '1.5', '1.6', '1.7']
     name: dbt ${{ matrix.dbt }}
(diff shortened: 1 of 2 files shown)


ZeRego (approved):
lgtm :) 
Running tests before merging

PR #8907, opened 2024-02-12
chore: add more span attributes to screenshots

joaoviana:
Closes: N/A

### Description:

Add more span attributes to screenshot logger

### Reviewer actions


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -402,14 +420,14 @@ export class UnfurlService {
                             width: gridWidth ?? viewport.width,
                         });
                     }
-                    await page.on('requestfailed', (request) => {
+                    page.on('requestfailed', (request) => {
```
rephus, line 423: why did you remove the await ?  in case of error, will it throw an `uncaught error` that can break the server ?

rephus (approved):

packages/backend/src/services/UnfurlService/UnfurlService.ts
joaoviana: Because the `await` had no effect on this expression, 

image

The error would be caught anyway by the `try/catch` wrap I believe

Issue #8906, opened 2024-02-12
I would like to export charts as `png` by default


joaoviana:
### Description

Users like the background transparency + `png` exports usually render in better quality. 

1. Move `png` as the default export
2. Allow the toggling of `background` transparency - set opaque by default

### What problem does this solve?

Better exports

PR #8908, opened 2024-02-12
feat: export png by default

jogeshwar01:
Closes: #8906 

### Description:

- Set `png` as the default export option in charts
- Allow the toggling of background transparency in `png` - set transparent by default
- I have added select options `Transparent` and `Opaque` for png exports. We can change the values to whatever feels suitable. Can also create an enum for the transparency options similar to how its done for download type if need be.

Output - 

`bg-transparency option in png` - 

image

Transparent PNG -

image

Opaque PNG - 

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/ChartDownload.tsx
+++ b/packages/frontend/src/components/ChartDownload.tsx
@@ -5,7 +5,14 @@ import {
     ChartType,
     getCustomLabelsFromColumnProperties,
 } from '@lightdash/common';
-import { Button, Popover, Select, Stack, Text } from '@mantine/core';
+import {
+    Button,
+    Popover,
+    SegmentedControl,
+    Select,
+    Stack,
+    Text,
+} from '@mantine/core';
 import { IconDownload, IconShare2 } from '@tabler/icons-react';
 import EChartsReact from 'echarts-for-react';
 import JsPDF from 'jspdf';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/ChartDownload.tsx
```diff
@@ -180,6 +191,21 @@ const ChartDownloadOptions: React.FC<DownloadOptions> = ({
...
+                    onChange={(value) =>
+                        setBackgroundTransparent(value === 'Transparent')
+                    }
+                    data={[
+                        { value: 'Transparent', label: 'Transparent' },
+                        { value: 'Opaque', label: 'Opaque' },
+                    ]}
+                />
+            )}
+
```
joaoviana, line 208: https://v6.mantine.dev/core/segmented-control/

Could we use a `Segmented Control` where `opaque` is the default? 

Just made the change to the original ticket 👍

packages/frontend/src/components/ChartDownload.tsx
```diff
@@ -50,7 +51,10 @@ async function base64SvgToBase64Image(
             canvas.height = width / ratio;
             const ctx = canvas.getContext('2d');
             if (ctx) {
-                if (type === 'jpeg') {
+                if (
+                    type === 'jpeg' ||
+                    (type === 'png' && !backgroundTransparent)
```
joaoviana, line 56:
```suggestion
                    (type === 'png' && !isBackgroundTransparent)
```

joaoviana (changes requested):
Thanks for the speedy contribution @jogeshwar01!
Minor change in requirements: 
* `opaque` is set by default
* use `SegmentedControl` instead of a `Select`

Thanks!

jogeshwar01:
@joaoviana Pushed the suggested changes.

Here is the `Segmented Control` - 

image

joaoviana (approved):

PR #8922, opened 2024-02-14
feat: enable soft-wrapping by default, store user choice locally

filp:
### Description:

Follow-up to: #8771 

- Enables soft-wrapping by default
- Saves the user's selection locally using local storage

Note: this implementation tracks sql runner and the table calculations sql form separately; there's likely a good case for reusing the same state between both, but given the different dimensions & context for either, there's potentially good reasoning for keeping them separate, too.

### Reviewer actions


almeidabbm (approved):
LGTM 🎉 
Failing tests are not related

Issue #3961, opened 2022-12-09
Threshold alerting (slack)


12ian34:
### Description

I would like to be alerted in some way when certain metrics reach thresholds that I can define myself on a per metric basis.

### What problem does this solve?

This is a common feature request from our customers that some other tools provide. This solves a problem of needing to repeatedly check charts or dashboards for metrics. For example, if your Sales revenue was falling below a certain threshold, it would be great to just be alerted about that when it happens, rather than to keep checking the Sales dashboard every day.

### MVP criteria
**How do I access the feature?**
- [ ] You should be able to create one or more alerts for any saved chart using the three-dot menu

**How do I set up the threshold?**
- [ ] You should be able to name your threshold alert
- [ ] You should be able to pick any numeric field that is used in the selected chart (metric / custom metric / dimension) to set the threshold on. 
- [ ] You should be able to pick from two conditions - `is greater than (or equal to) xxx` or `is less than (or equal to) xxx` 
- [ ] You should be able to pick the value from a number input field

**How do I pick the destination?**
- [ ] You should be able to pick one or more Slack channels to send the alert to. This should work similarly to the way you pick slack channels in the `scheduled deliveries` form

**What happens next?**
- [ ] Lightdash will run a daily query to check if the threshold has been crossed (when exactly?) 
- [ ] If the threshold has been crossed, the relevant Slack channels will be sent a message 

**What if I want to update the delivery?**
- [ ] You can see a complete list of alerts per chart when you click on 'alerts' from the 3-dot menu
- [ ] You can edit the alert by clicking on it
- [ ] You can delete the alert

12ian34:
This has been requested by a few customers!

yu-iskw:
I would like the feature too. I don't think Lightdash supports something like anomaly detection. But, it would be useful to send notifications if a monitored value meets defined conditions.

tlfbrito:
We would also like to see this implemented 👍

tlfbrito:
We want to create a set of rules that once they reach to a certain threshold, we want to send alerts/notifications.

PR #8924, opened 2024-02-14
feat: threshold alerting 2

rephus:
Closes:  #3961

### Reviewer actions


packages/frontend/src/components/LightdashVisualization/VisualizationConfigPie.tsx
```diff
@@ -47,15 +47,7 @@ const VisualizationPieConfig: FC<VisualizationConfigPieProps> = ({
...
-        const tableCalculations = Object.entries(itemsMap ?? {}).reduce<
-            Record<string, TableCalculation>
-        >((acc, [key, value]) => {
-            if (isTableCalculation(value)) {
-                return { ...acc, [key]: value };
-            }
-            return acc;
-        }, {});
-
+        const tableCalculations = getTableCalculationsFromItemsMap(itemsMap);
```
magnew, line 50: I can split this out if anyone wants. We should clean up how everything is getting these

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -84,8 +92,28 @@ const DEFAULT_VALUES = {
...
+        {
+            fieldId: '',
+            operator: ThresoldOperator.GREATER_THAN,
+            value: 0,
+        },
+    ],
+};
+
+const thresholdOperatorOptions = [
+    { label: 'Greater than', value: ThresoldOperator.GREATER_THAN },
```
rephus, line 111: you could make this dynamic by iterating on ThresholdOperator enum,and using friendlyName for the label. This works too though.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -182,6 +211,8 @@ type Props = {
     onBack?: () => void;
     loading?: boolean;
     confirmText?: string;
+    isAlert?: boolean;
```
rephus, line 214: isAlert might lead to confusion, I prefer `isThreshold` or even `isThresholdAlert`

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -368,174 +416,250 @@ const SchedulerForm: FC<Props> = ({
...
+                                            `thresholds.0.value`,
+                                        )}
+                                    />
+                                </Group>
+                            </Stack>
+                        )}
+                        {isAlert ? (
+                            <Input.Wrapper label="Frequency">
+                                <Text color="gray" mt={8}>
+                                    Alerts will be processed at 10am daily
```
rephus, line 480: perhaps add `UTC` ?

rephus (commented):
LGTM, just a few minor comments

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
magnew: I thought about it, but since those are the values that get stored in the DB, I don't love to tie them so directly to the UI. If we want to change what the display text is, it would break existing saved things OR you'd have to do this anyway. 

It already came up, actually 😄. We changed this to 'is greater than'. If it had been using friendly name I would have had to change the type values. 

I'll leave this for now and if you disagree or theres something I'm not considering, let me know and I wont argue further.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
magnew: Ok, fair enough, I will change it to isThresholdAlert to be the most explicit.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
magnew: 😅 Ok. Is that right? When people pick a time is it in UTC?

magnew: I put it in there. If it's wrong, we are going to change this soon anyway.

magnew (approved):

PR #8925, opened 2024-02-14
fix: wrap sql_filter with ()

joaoviana:
Closes: https://github.com/lightdash/lightdash-commercial/issues/73

### Description:

(See ticket from commercial on how to replicate ^)

When users set a `sql_filter` that is similar to this expression: 
```
${TABLE}.status = ${lightdash.attributes.order_status} or ${lightdash.attributes.order_status} = 'ALL'
```

And if the following statement is true: `${TABLE}.status = ${lightdash.attributes.order_status}`, then if the user sets extra filters on a chart, these will be ignored because there's a `( )` wrap missing. 

This will then affect charts and their generated SQL + where they're being used (dashboards, embedded dashboards)

This will then wrap the `replacedUserAttributesSQL` with a `()`

Question: should this be done somewhere else? I'm not sure... 

### Reviewer actions

### DIFF
--- a/packages/backend/src/queryBuilder.mock.ts
+++ b/packages/backend/src/queryBuilder.mock.ts
@@ -1259,7 +1259,7 @@ export const METRIC_QUERY_WITH_SQL_FILTER = `SELECT
   MAX("table1".number_column) AS "table1_metric1"
 FROM "db"."schema"."table1" AS "table1"
 
-WHERE 'EU' = 'US'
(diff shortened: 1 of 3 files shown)


rephus (approved):
Looks good to me

joaoviana:
merging - only e2e test issue is with `dashboard.cy.ts` and it's a uuid mismatch

PR #8943, opened 2024-02-15
fix: use dbt model package_name for lightdash generate

owlas:
### Problem

some dbt configurations are not listing out models in packages for example:

```
# this does not work
dbt ls -s other_package.model_in_other

# this does work
dbt ls -s model_in_other
```

In the first case `dbt` fails. In the second case `lightdash fails`. This fixes the second case, by checking for the `package_name` on the model metadata instead of relying on the `patch_path` to determine it (it seems sometimes the `patch_path` is missing)

Slack thread: https://lightdash.slack.com/archives/C0699FEJCR0/p1708003529824679?thread_ts=1705415250.660129&cid=C0699FEJCR0

### DIFF
--- a/packages/cli/src/dbt/models.ts
+++ b/packages/cli/src/dbt/models.ts
@@ -23,6 +23,7 @@ type CompiledModel = {
     database: string;
     originalFilePath: string;
     patchPath: string | null | undefined;
+    packageName: string;
     alias?: string;
 };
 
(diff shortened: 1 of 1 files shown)


ZeRego (approved):

packages/cli/src/dbt/models.ts
```diff
@@ -172,7 +173,7 @@ export const findAndUpdateModelYaml = async ({
         filenames.push(expectedYamlPath);
     }
     const defaultYmlPath = path.join(
-        path.dirname(path.join(projectDir, model.originalFilePath)),
```
ZeRego, line 175: are we making assumption the package name equals the folder name?

owlas: I'm assuming they are the same, I don't know if it's possible for them to differ

Issue #8891, opened 2024-02-12
adding `is null` filter or `is empty string` filters from a chart in a dashboard are incorrect. 


TuringLovesDeathMetal:
### Description

Right now, when I click on a chart value that `is null` and try to add it as a filter to my dashboard, it filters it as a string (i.e. `string is 'null'`) instead of filtering it as a null (i.e. `string is NULL`).

Here's what's happening today:
![image](https://github.com/lightdash/lightdash/assets/31848148/71c10370-0f21-4b5b-9c50-3f70f95fcf7e)

Here's what should be happening: 
![image](https://github.com/lightdash/lightdash/assets/31848148/1352e623-9647-46f0-8567-fed2e1f3228b)

PR #8953, opened 2024-02-15
fix: "filter dashboard to" for null value

joaoviana:
Closes: #8891

### Description:

On a dashboard, when the value on a cell or chart point is `null`, and when users clicked on `filter dashboard to` the following happened: 
* `filter dashboard to` with empty value
* when user clicked on that (to filter the dashboard by the `null` value), it wouldn't create a temp filter with `is null` operator, it would instead create a filter with `is/EQUAL` with empty value. 

This PR adds:
- seed data to the events table that includes `null`/empty `event_id`s
- utilises the util `createDashboardFilterRuleFromField` more across the codebase 
- creates a shared component `FilterDashboardTo` that is used on the char tiles and cell context menu

Screen recording:

https://github.com/lightdash/lightdash/assets/7611706/ec9f2b59-00f2-4be0-a22f-2757166de9cb

### Reviewer actions


packages/frontend/src/components/SimpleTable/DashboardCellContextMenu.tsx
```diff
@@ -59,34 +58,36 @@ const DashboardCellContextMenu: FC<
...
+
+        acc.push(
+            createDashboardFilterRuleFromField({
+                field: pivotField,
+                availableTileFilters: {},
+                isTemporary: true,
+                value: pivot.value,
+            }),
+        );
+        return acc;
```
IrakliJani, line 89: nit: this can be `return [...acc, createDashboarblabla]` to avoid using push

packages/frontend/src/components/CustomSqlPanel/CustomMetricsTable.tsx
```diff
@@ -67,11 +67,6 @@ const CustomMetricsTable: FC<Props> = ({
         [checkedUuids],
     );
     const toggleAllCheckedUuids = useCallback(() => {
-        console.log(
-            'toggleAllCheckedUuids',
-            checkedUuids.size,
-            customMetrics.length,
-        );
```
IrakliJani, line 74: logs

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -750,40 +745,10 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
...
-                                                            filter.values[0],
-                                                        )}
-                                                </Text>
-                                            </Menu.Item>
-                                        ),
-                                    )}
-                                </>
+                                <FilterDashboardTo
+                                    filters={dashboardTileFilterOptions}
+                                    addFilter={handleAddFilter}
```
IrakliJani, line 750: nit: `onAddFilter={handleAddFilter}`

packages/frontend/src/components/SimpleTable/DashboardCellContextMenu.tsx
```diff
@@ -59,34 +58,36 @@ const DashboardCellContextMenu: FC<
...
-            operator: FilterOperator.EQUALS,
-            values: [pivot.value],
-            label: undefined,
-        };
-    });
+        if (
+            !pivotField ||
+            !isField(pivotField) ||
+            !isFilterableField(pivotField)
+        )
```
IrakliJani, line 78: can you explain what is the reason behind this change?

packages/common/src/utils/filters.ts
```diff
@@ -323,7 +329,7 @@ export const createDashboardFilterRuleFromField = (
             disabled: !isTemporary,
             label: undefined,
         },
-        includeDefaultValue ? [] : null,
+        value !== undefined ? [value] : [],
```
IrakliJani, line 332: so if we are creating a "is null" operator do we still have to provide it as a value?

operator: NULL
value: [null] << ?

packages/common/src/utils/filters.ts
```diff
@@ -304,17 +304,23 @@ export const applyDefaultTileTargets = (
...
+    isTemporary: boolean;
+    value?: any;
+}): DashboardFilterRule =>
     getFilterRuleWithDefaultValue(
         field,
         {
             id: uuidv4(),
-            operator: FilterOperator.EQUALS,
+            operator:
+                value === null ? FilterOperator.NULL : FilterOperator.EQUALS,
```
IrakliJani, line 323: is this function being used somewhere where user might not create a filter like this one:
field: `name` (string)
operator: `INCLUDE`
values [null, 'irakli'] << not sure if this is possible today. just asking

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -393,9 +392,8 @@ const DashboardChartTileMain: FC<DashboardChartTileMainProps> = (props) => {
         [],
     );
 
-    const [dashboardTileFilterOptions, setDashboardFilterOptions] = useState<
-        DashboardFilterRule[]
-    >([]);
+    const [dashboardTileFilterOptions, setDashboardTileFilterOptions] =
```
IrakliJani, line 395: 👍

IrakliJani (commented):
looks good, nice tidy-up! have some Qs and nits

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
joaoviana: nice!

packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+    if (filter.values)
+        return (
+            <Text span fw={500}>
+                {String(filter.values[0])}
+            </Text>
+        );
+
+    return null;
+};
```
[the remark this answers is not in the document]
joaoviana: sure, will move it down!

packages/frontend/src/components/SimpleTable/DashboardCellContextMenu.tsx
joaoviana: oh because the `pivotField` has to be filterable (of this type Field | Dimension | Metric | TableCalculationField) 

When creating a Dashboard Filter Rule it has to be `FilterableField`, otherwise we don't allow it. This will work as before, just more type robust

packages/frontend/src/components/SimpleTable/DashboardCellContextMenu.tsx
joaoviana: sure!

packages/common/src/utils/filters.ts
joaoviana: Good spot, I mean `value === null`, thank you!

packages/common/src/utils/filters.ts
joaoviana: thanks for spotting this! It makes sense to rework this a bit more to avoid the bug you've just mentioned.

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -109,12 +109,11 @@ const FilterConfiguration: FC<Props> = ({
 
         if (newField && isField(newField) && isFilterableField(newField)) {
             setDraftFilterRule(
-                createDashboardFilterRuleFromField(
-                    newField,
+                createDashboardFilterRuleFromField({
+                    field: newField,
```
joaoviana, line 113: this is where it could be _wrong_ but it's not! 
At the moment, if we don't pass a `value` to this functions' object, we set the operator by default to `EQUALS` - this was done before 
ONly if we pass a `value` `null` to it we change its operator to `NULL`

packages/common/src/utils/filters.ts
```diff
@@ -329,7 +329,7 @@ export const createDashboardFilterRuleFromField = ({
             disabled: !isTemporary,
             label: undefined,
         },
-        value === null ? undefined : [value],
+        value ? [value] : null, // When `null`, don't set default value if no value is provided
```
joaoviana, line 332: If value is `falsy`, don't set the default value that is handled by `getFilterRuleWithDefaultValue`

This works OK when: 
`filtering dashboard to...` and when `creating a filter`

packages/common/src/utils/filters.ts
IrakliJani: thanks for the explanation. that makes sense

packages/common/src/utils/filters.ts
```diff
@@ -304,17 +304,23 @@ export const applyDefaultTileTargets = (
...
+export const createDashboardFilterRuleFromField = ({
+    field,
+    availableTileFilters,
+    isTemporary,
+    value,
+}: {
+    field: FilterableField;
+    availableTileFilters: Record<string, FilterableField[] | undefined>;
+    isTemporary: boolean;
+    value?: any;
```
IrakliJani, line 316: I suggest using `unknown` which is "more safe"
```suggestion
    value?: unknown;
```

packages/frontend/src/components/SimpleTable/DashboardCellContextMenu.tsx
IrakliJani: thanks!

packages/common/src/utils/filters.ts
IrakliJani: so in this case if value is `null` then we don't need to include in the values array. code is a bit confusing but comment makes it clear. thanks

IrakliJani (commented):
code looks good 🎉 
I'm going to test on render now before approving

IrakliJani (changes requested):
@joaoviana I've noticed a weird behavior when working with "any value" filters

here's a video on how to reproduce it: https://github.com/lightdash/lightdash/assets/962095/8ace6cd5-e5b8-4557-aeaa-cc8882249618

packages/common/src/utils/filters.ts
joaoviana: done!

joaoviana:
> @joaoviana I've noticed a weird behavior when working with "any value" filters
> 
> here's a video on how to reproduce it: https://github.com/lightdash/lightdash/assets/962095/8ace6cd5-e5b8-4557-aeaa-cc8882249618

Replace the `any` to `unknown`

I've raised the bug you've reported here: https://github.com/lightdash/lightdash/issues/9029

The util that was refactored doesn't get called when changing the operator, so this is unrelated and is in `main` already 😢

IrakliJani (approved):

Issue #7150, opened 2023-09-20
I want to be able to specify email destinations for my threshold alerts 


PriPatel:
Blocked by  #3961  and [internal design ticket ](https://github.com/lightdash/lightdash-internal-work/issues/1011)

**The 'threshold alert' form**
- [ ] I should be able to set one or more email addresses as destinations. This should work similarly to 'scheduled deliveries'

**The email**
- [ ] - The email subject should show the alert message
- [ ] - The body message should reference the change and the original threshold
- [ ] - A pic of the latest chart should be attached.
- [ ] - A link to the chart should be shown 

**Open questions:**
- Does 'expiration' still affect these emails?

rephus:
>     Does 'expiration' still affect these emails?

Yes, because we store the image in the same s3 repo

PR #8959, opened 2024-02-16
feat: send email on threshold alerts

rephus:
Closes:   #7150

### Description:

Screenshot from 2024-02-16 11-11-05

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
@@ -80,7 +80,7 @@ enum SlackStates {
 const DEFAULT_VALUES = {
     name: '',
     message: '',
-    format: SchedulerFormat.CSV,
+    format: SchedulerFormat.IMAGE,
     cron: '0 9 * * 1',
     options: {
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -80,7 +80,7 @@ enum SlackStates {
 const DEFAULT_VALUES = {
     name: '',
     message: '',
-    format: SchedulerFormat.CSV,
+    format: SchedulerFormat.IMAGE,
```
magnew, line 83: We want to change the default? That seems fine with me. Should we also switch the order of these? So the default is on the left still

Screenshot 2024-02-16 at 17 25 24

magnew (approved):
LGTM. Left a question.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
rephus: ohh, good question. no, I was only supposed to change the default for ALERT,

I push a change. 

good catch.

Issue #8978, opened 2024-02-19
Do not assign shared colors on non-grouped charts


filp:
When assigning shared colors in charts, exclude non-grouped chart values from this logic, and retain the previous behavior, where colors are assigned in order.

PR #8988, opened 2024-02-19
feat: Do not apply shared colors to non-group series

filp:
Closes: #8978 

### Description:

Skips the shared color assignment logic for non-group series, instead using the default behavior that applies color from the palette in order.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/AppearanceSettingsPanel/index.tsx
+++ b/packages/frontend/src/components/UserSettings/AppearanceSettingsPanel/index.tsx
@@ -1,4 +1,5 @@
 import { subject } from '@casl/ability';
+import { ECHARTS_DEFAULT_COLORS } from '@lightdash/common';
 import {
     ActionIcon,
     Button,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/UserSettings/AppearanceSettingsPanel/index.tsx
```diff
@@ -47,15 +48,8 @@ const defaultColorPalettes: ColorPalette[] = [
...
-            '#ff8933',
-            '#ffbd33',
-            '#ffe133',
-            '#e8ff33',
-            '#b4ff33',
-            '#80ff33',
-            '#4cff33',
-            '#33ff49',
+            // Use the initial 9 colors directly from ECHARTS to keep them in sync:
+            ...ECHARTS_DEFAULT_COLORS,
```
filp, line 52: Not really related to this task, but makes the provenance of those initial 9 colors clearer, and avoids things going out of sync in the future.

packages/frontend/src/hooks/useChartColorConfig.ts
```diff
@@ -5,6 +5,13 @@ import { EChartSeries } from './echarts/useEchartsCartesianConfig';
 
 export type SeriesLike = EChartSeries | Series;
 
+export const isGroupedSeries = (series: SeriesLike) => {
```
filp, line 8: Has a bit of duplication with the logic below, but opted to keep it like this for clarity.

filp:
@magnew Basically color sharing will happen only in cases where there are distinct groups (so pie charts, or other charts with a grouping dimension). This means most charts will follow predictably the order of colors as they are defined in the palette.

If there's a grouping dimension, shared color assignment kicks in and attempts to give each group a deterministic color value.

magnew:
> @magnew Basically color sharing will happen only in cases where there are distinct groups (so pie charts, or other charts with a grouping dimension). This means most charts will follow predictably the order of colors as they are defined in the palette.
> 
> If there's a grouping dimension, shared color assignment kicks in and attempts to give each group a deterministic color value.

Cool, thanks for explaining. I liked when the colors got shared across dashboards, but I guess people dont like change 😄

Issue #5631, opened 2023-05-25
Use full text search to improve search bar


owlas:
## Problem

Our current search is based on using 

```sql
LOWER(name) LIKE LOWER('%query%')
OR
LOWER(description) LIKE LOWER('%query%')
```

This has the following problems:
- Punctuation has to be matched exactly
- Word order has to be matched exactly
- names and descriptions are equally important
- It's SLOW, because we don't index `lower()` we have to do a full table scan and `lower()` everything
- no ranking of results

## Solution

We can use Postgres' included full text search, this converts the the string into a vector of tokens that:
- removes punctuation
- removes "stop words" like "a" "an" "the" ...
- stems English words (I think we can only use one language at a time?) => ordering, orders, order *all match* "order"
- can be queried without order being important => active users *can match* "users are active"
- can weight the name more than the description!
- can rank results by how accurately they match

### migrations

Each resource (chart, dashboard, ...) that needs to be searched will need a new field added, the search field. This will include the tokenised version of chart that will be indexed. The index is _always_ up to date so we don't need to maintain the search index.

The following example for saved charts combines both the name and description field, but weights names as more relevant than descriptions:

```typescript
export async function up(knex: Knex): Promise<void> {
    if (await knex.schema.hasTable('saved_queries')) {
        await knex.schema.alterTable('saved_queries', (table) => {
            table.specificType(
                'search',
                `tsvector GENERATED ALWAYS AS (setweight(to_tsvector('english', name), 'A') || ' ' || setweight(to_tsvector('english', description), 'B') :: tsvector) STORED`,
            );
            table.index(['search'], 'saved_queries_search_index', 'GIN');
        });
    }
}

export async function down(knex: Knex): Promise<void> {
    if (await knex.schema.hasTable('saved_queries')) {
        await knex.schema.alterTable('saved_queries', (table) => {
            table.dropColumn('search');
            table.dropIndex('saved_queries_search_index');
        });
    }
}
```

### Querying

When querying we can use `websearch_to_tsquery` which makes the order of query terms not important but the user can use some symbols to control search.

On saved charts, this would and a where clause to filter the results and then rank them from most relevant to least relevant:

```typescript
            .andWhereRaw(`search @@ websearch_to_tsquery('english', ?)`, [
                query,
            ])
            .orderByRaw(
                `ts_rank(search, websearch_to_tsquery('english', ?)) DESC`,
                [query],
            );
```

to query this the user can type:

[...]

PR #8992, opened 2024-02-19
feat: add full text search to global search

almeidabbm:
Closes: [#5631](https://github.com/lightdash/lightdash/issues/5631)

### Description:

- Adds full text search to `charts`, `dashboards` and `spaces`
- Uses a custom english dictionary with no stop-words

### Reviewer actions


packages/backend/src/database/seeds/development/05_generate_fake_data.ts
```diff
@@ -0,0 +1,151 @@
+import { faker } from '@faker-js/faker';
```
owlas, line 1: do we always want this seed for devs?

packages/backend/package.json
```diff
@@ -32,6 +32,7 @@
         "@aws-sdk/lib-storage": "^3.306.0",
         "@aws-sdk/s3-request-presigner": "^3.272.0",
         "@casl/ability": "^5.4.3",
+        "@faker-js/faker": "^8.4.0",
```
owlas, line 35: this should be in dev dependencies, we don't want to bundle into prod

packages/backend/src/database/seeds/development/05_generate_fake_data.ts
almeidabbm: This seed only runs if `FAKE_DATA_LENGTH >= 1`, so unless we explicitly declare it this won't add anything

packages/backend/src/database/migrations/20240207133121_add-english-dictionary-with-custom-stop-words.ts
```diff
@@ -0,0 +1,35 @@
+import { Knex } from 'knex';
```
owlas, line 1: you could merge into a single migration, why separate?

almeidabbm: good point, this was merged over multiple PRs which resulted in separate migrations.
also, merging them into 1 single file won't impact readability too much, doing it 💪

IrakliJani (approved):
🚀

PR #8995, opened 2024-02-19
docs: update faq to intercom articles

TuringLovesDeathMetal:
Removes our existing FAQs page and updates it to use intercom articles


owlas (changes requested):
@TuringLovesDeathMetal could you update the redirects so the previous links point to intercom? Maybe to a more generic intercom link? In case these are indexed by google

https://github.com/lightdash/lightdash/blob/290a773f88af0e7276080552f15bff2e24fb36f1/docs/docusaurus.config.js#L24

TuringLovesDeathMetal:
@owlas how do I do that? 

The two articles/docs that I moved were an FAQ that I created a month ago (about totals) and the contact us page, so do we think they'll be super important to redirect? (if it's super quick, I can do it - if this will take more than 30 mins, then I probably won't bother)

docs/docusaurus.config.js
```diff
@@ -59,6 +59,14 @@ module.exports = {
                         to: '/self-host/self-host-lightdash',
                         from: ['/self-host'],
                     },
+                    {
+                        to: 'https://intercom.help/lightdash-faqs/en/collections/8439498-faqs',
+                        from: ['/help-and-contact/faqs/faqs']
```
ZeRego, line 64: Looks like the version we use doesn't allow redirects to external link

> 9:00:43 AM: [ERROR] ValidationError: "redirects[8].to" is not a valid pathname. Pathname should start with slash and not contain any domain or query string.

They added this feature in v3
https://github.com/facebook/docusaurus/pull/9171

TuringLovesDeathMetal: ah, fair enough. I'll do that then!

owlas: sorry all!

Issue #7182, opened 2023-09-22
I should be able to temporarily pause scheduled deliveries & threshold alerts (threshold alerting)


PriPatel:
- I might only want to get alerted during certain periods of the month/year (eg. Christmas sales). 
- I might want to temporarily pause scheduled deliveries
- If for some reason, the values consistently remain above the threshold, and I'm being alerted daily/hourly, there should be a way for me to turn the alerts off temporarily, without having to set it back up again later on. 

- [ ] There should be a toggle in the saved list of alerts and saved list of scheduled deliveries
- [ ] The toggle is 'on' by default
- [ ] If the user switches the toggle off, they will stop receiving notifications. 

<img width="427" alt="Screenshot 2023-09-22 at 10 37 36" src="https://github.com/lightdash/lightdash/assets/12776522/d83bf766-bc4d-4f93-8599-fa9120f57ef6">

PR #8996, opened 2024-02-19
feat: scheduler active flag

rephus:
Closes:  #7182

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/schedulerController.ts
+++ b/packages/backend/src/controllers/schedulerController.ts
@@ -112,6 +112,36 @@ export class SchedulerController extends Controller {
     }
 
     /**
+     * Set scheduler active/inactive
+     * @param schedulerUuid The uuid of the scheduler to update
+     * @param req express request
+     * @param body the active flag
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('201', 'Updated')
+    @Patch('{schedulerUuid}')
+    @OperationId('updateSchedulerActive')
+    async patchActive(
+        @Path() schedulerUuid: string,
+        @Request() req: express.Request,
+        @Body() body: { active: boolean },
+    ): Promise<ApiSchedulerAndTargetsResponse> {
+        this.setStatus(200);
+        return {
+            status: 'ok',
+            results: await schedulerService.setSchedulerActive(
+                req.user!,
+                schedulerUuid,
+                body.active,
+            ),
+        };
+    }
+
+    /**
      * Delete a scheduler
      * @param schedulerUuid The uuid of the scheduler to delete
      * @param req express request
--- a/packages/backend/src/controllers/schedulerController.ts
+++ b/packages/backend/src/controllers/schedulerController.ts
@@ -123,7 +123,7 @@ export class SchedulerController extends Controller {
         unauthorisedInDemo,
     ])
     @SuccessResponse('201', 'Updated')
-    @Patch('{schedulerUuid}')
+    @Patch('{schedulerUuid}/enabled')
     @OperationId('updateSchedulerEnabled')
     async patchEnabled(
         @Path() schedulerUuid: string,
--- a/packages/backend/src/controllers/schedulerController.ts
+++ b/packages/backend/src/controllers/schedulerController.ts
@@ -112,10 +112,10 @@ export class SchedulerController extends Controller {
     }
 
     /**
-     * Set scheduler active/inactive
+     * Set scheduler enabled
      * @param schedulerUuid The uuid of the scheduler to update
      * @param req express request
-     * @param body the active flag
+     * @param body the enabled flag
      */
     @Middlewares([
         allowApiKeyAuthentication,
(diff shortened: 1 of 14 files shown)


packages/backend/src/controllers/schedulerController.ts
```diff
@@ -111,6 +111,36 @@ export class SchedulerController extends Controller {
...
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('201', 'Updated')
+    @Patch('{schedulerUuid}')
+    @OperationId('updateSchedulerActive')
+    async patchActive(
```
rephus, line 128: alternative implementation: 

UPdate existing udpate endpoint to accept `active` flag

Const: , for disabling a scheduler, we will require to pass the full object

packages/backend/src/database/migrations/20240219150917_scheduler_active_flag.ts
```diff
@@ -0,0 +1,13 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable('scheduler', (t) => {
+        t.boolean('active').defaultTo(true);
```
rephus, line 5: alternative implementation: 

use a string or an emum instead of a boolean that we can use for multiple conditions (disabled, send_once....) 

I think is a bit harder to use, I personally prefer to have 2 columns, 1 with the boolean and another with the condition.

packages/backend/src/models/SchedulerModel/index.ts
```diff
@@ -252,6 +254,20 @@ export class SchedulerModel {
...
     }
 
+    async setSchedulerEnabled(
+        schedulerUuid: string,
+        enabled: boolean,
+    ): Promise<SchedulerAndTargets> {
+        await this.database(SchedulerTableName)
+            .update({
+                enabled,
+                updated_at: new Date(),
```
magnew, line 264: Woah, I didn't know you had to set updated_at manually. Do you have to do this every time you update a database entry?

packages/backend/src/services/SchedulerService/SchedulerService.ts
```diff
@@ -219,7 +219,36 @@ export class SchedulerService {
             SchedulerModel.getSlackChannels(scheduler.targets),
         );
 
-        await schedulerClient.generateDailyJobsForScheduler(scheduler);
+        // We only generate jobs if the scheduler is enabled
+        if (scheduler.enabled)
```
magnew, line 223: Does checking for 'enabled' vs using a disabled flag mean that exisitng schedulers won't run unless they get enabled=true added to the database?

I imagine you have a good reason for doing it this way, but I would have assumed the new flag would be 'disabled' because the default state of schedulers seems like it would enabled.

packages/backend/src/controllers/schedulerController.ts
```diff
@@ -111,6 +111,36 @@ export class SchedulerController extends Controller {
...
+     * @param req express request
+     * @param body the enabled flag
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('201', 'Updated')
+    @Patch('{schedulerUuid}/enabled')
```
magnew, line 126: You kind of alluded to this in one of your comments, but would it make sense to make the least the name of this endpoint more generic? `'{schedulerUuid}/control'` or something. So that we send other control signals (pause once maybe) to the same endpoint. Or should we just use the generic scheduler patch endpoint? 

Those suggestions have their downsides too -- if you did it this way for a specific reason, that's cool with me.

magnew (approved):
I put in a couple comments/questions about the pattern here, but they don't have to be blocking. The code looks good, I'm just wondering about the call pattern. There are probably reasons you did it this way that I haven't thought through. Nice work.

packages/backend/src/models/SchedulerModel/index.ts
rephus: in theory, it is not needed, because there is a trigger in db. However, this is kind of a type limitation in the model.

packages/backend/src/controllers/schedulerController.ts
rephus: >  Or should we just use the generic scheduler patch endpoint?

I already made a comment about this here https://github.com/lightdash/lightdash/pull/8996#discussion_r1494785458

`/control` could work, but I don't think we are going to add more control flags in the future (even `notify me once` I think it will be part of the `generic patch` endpoint ) 

So I called it /enabled so we can be explicit about what this does... 

But I have no strong opinion, so let me know what you think.

magnew: Right, I read your comment, then forgot to look back at it. 

Personally I'd expect it to be part of the generic update endpoint. But I see your point about it being annoying to pass the whole object to change this. The tradeoff is having this extra endpoint to maintain. 

I think I'd lean towards using /update, but I don't think I have a strong enough feeling about it to ask you to change it 😄

packages/backend/src/services/SchedulerService/SchedulerService.ts
magnew: I think that does make sense, but what I mean is more like, why use a flag that must be defined in order for them to run instead of a flag that must be defined in order to turn them off -- does that make sense? If we add 'disabled' instead, all existing schedulers won't have to change. Their behavior will only change when someone disables them. With 'enabled' we'll have to add the entry to the database for each existing scheduler, right?  

Maybe I'm missing some magic you know about 😄

rephus: > 'enabled' we'll have to add the entry to the database for each existing scheduler, right?

We already do that, on the migration, and it defaultsTo (true) . SO we don't need to manually change anything about existing schedulers, all existing schedulers will be enabled by default. And they can only be disabled by clicking the toggle

PR #9005, opened 2024-02-19
fix: change name to 'Alerts' in chart context menus

maxdiplogit:
Closes: #8980

### Description:
This PR changes name from `Threshold Alerts` to `Alerts` in Chart Context Menus and also for the Threshold Alert Modal as shown in the following screenshots.

Please let me know if there are any other changes to be made, I will be happy to make them as desired :)

Screenshot from 2024-02-20 05-06-13
Screenshot from 2024-02-20 05-08-43

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #9013, opened 2024-02-20
fix: change name to Alerts in chart context menus [mirror]

IrakliJani:
mirror of: https://github.com/lightdash/lightdash/pull/9005


PriPatel:
Looks correct to me

PR #9015, opened 2024-02-20
fix: generate valid urls based on site url

ZeRego:
If the SITE_URL variable ends with a `/` we generate invalid URLs. Example:

> Cannot POST //api/v1/headless-browser/login/ff2a9f1d-08bd-49df-8ce5-540e8a06197f

Change updates all the cases where we were concatenating the SITE_URL to use [URL constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).

This should also fix any security issues related to path traversing. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CsvService/CsvService.ts
+++ b/packages/backend/src/services/CsvService/CsvService.ts
@@ -477,7 +477,10 @@ export class CsvService {
             DownloadFileType.CSV,
         );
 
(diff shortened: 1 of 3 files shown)


almeidabbm (approved):
Tested locally, with SITE_URL ending in `/`.
Fix looks good 🎉

PR #9019, opened 2024-02-20
chore: change name of thresholdAlert to alert in code

magnew:
### Description:

We decided to make the name of threshold alerts more generic and just call them alerts. This changes the code (not the display strings) to reflect that. 

@rephus I put this together, but let me know if you don't think it's worth doing. In general it seems like a good idea to have the names match, but since this contains a migration maybe it's overkill? 

### Reviewer actions

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -711,7 +711,7 @@ export type SchedulerJobEvent = BaseTrack & {
         jobId: string;
         schedulerId: string | undefined;
         sendNow?: boolean;
-        isThresholdAlert?: boolean;
+        isAlert?: boolean;
     };
 };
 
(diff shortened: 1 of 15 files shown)


rephus:
I don't mind the name being different in the models/functions and the UI , this is something natural (we are not going to change the backend code everytime we rename a feature (this happened on google sheets) ) , and it is not very different anyway `threshold alert` -> `alert`. 

I agree the migration itself is a bit overkill, so if you really want to go ahead, my suggestion would be to keep the old name on the database and just change the rest ?

rephus:
and the conflicts :facepalm:

magnew:
> I don't mind the name being different in the models/functions and the UI , this is something natural (we are not going to change the backend code everytime we rename a feature (this happened on google sheets, or even savedCharts, or additionalMetrics -> customMetrics) ) , and it is not very different anyway `threshold alert` -> `alert`. In fact, in the backend we might want to be a bit more verbose, and avoid using a generic `alert`
> 
> I agree the migration itself is a bit overkill, so if you really want to go ahead, my suggestion would be to keep the old name on the database and just change the rest ?

Eh. This is the kind of thing that is fine to do if it's simple, but the amount of changes make me agree with you -- we don't need this.

Issue #8981, opened 2024-02-19
When a user is deleted, the `user_id` in share URL should be set to `null` 


TuringLovesDeathMetal:
### Description

Right now, when a user is deleted from an organization, we don't do anything to the `share URL` `created_by_user` values. This means that the share URLs are broken. 

Instead, we should set the `created_by_user_uuid` column to `null` as we do for charts and dashboards. This way the URLs still work even after teh user is deleted.

PR #9020, opened 2024-02-20
feat: set user id to null in share URL when user deleted

jogeshwar01:
Closes: #8981 

### Description:

- When a user is deleted, the `current_user_id` in share URLs is set to null. Currently we have the `ON DELETE CASCADE` for the foreign key `current_user_id` in the `shares` table. Due to this, when a user is deleted, any share URL created by him are broken. 

- Added a migration file to update this to `ON DELETE SET NULL` instead (also removed the `NOT NULL` constraint). Verified the changes in postgres database as well. (also checked the rollback command)

Share URL initially - 

image

When a user is deleted :

Before: 

image

After:

image

### Reviewer actions


ZeRego (approved):
lgtm 🎉 
waiting for e2e tests to pass

PR #9054, opened 2024-02-21
feat: Track color mappings via shared context

filp:
Closes: #9046 

### Description:

Uses a shared context to track color mappings globally, in order to better handle assignment-shifting/color collisions for common identifiers.

tl;dr: Colors are currently assigned on a per-chart basis, but in specific scenarios, particularly with high group distribution, color assignments can drift across two charts in the same dashboard. This change tracks assignments per identifier once, and then reuses it across the dashboard, and across different dashboards too.

### Reviewer actions


packages/frontend/src/hooks/useChartColorConfig.tsx
```diff
@@ -1,10 +1,53 @@
...
+
+    if (ctx == null) {
+        console.warn(
+            'No color mapping context available, using empty color map.',
+        );
+
+        return { colorMappings: new Map<string, string>() };
+    }
+
+    return ctx;
```
IrakliJani, line 48: we can "safely" throw an error since this is wrapping the routes and providers above that should not use the context

packages/frontend/src/hooks/useChartColorConfig.tsx
```diff
@@ -1,10 +1,53 @@
...
+    createContext<ChartColorMappingContextProps | null>(null);
+
+/**
+ * Exposes a map of identifier->color values, which can be shared across
+ * a context, for shared color assignment.
+ */
+export const ChartColorMappingContextProvider: FC<
+    React.PropsWithChildren<{}>
+> = ({ children }) => {
+    const [colorMappings] = useState(new Map<string, string>());
```
IrakliJani, line 22: Using a mutable map (or any mutable data structure) directly as state in React goes against the core principles of React's state management.

Whenever you call `colorMappings.set` (from `useChartColorConfig`), React's re-rendering will bail on that because the reference remains the same. So, it acts like a `useRef`.

You can either change it to something like this:

```jsx
const [colorMappings, setColorMappings] = useState(new Map<string, string>());

const addColor = (identifier, color) => {
  setColorMappings(new Map(colorMappings.set(identifier, color)));
}
```

and expose `addColor` from the provider...

...or just use an object, which might make more sense here unless you really want to use a Map:

```jsx
const [colorMappings, setColorMappings] = useState<Record<string, string>>({});

const addColor = (identifier, color) => {
  setColorMappings({ ...colorMappings, [identifier]: color });
}
```

packages/frontend/src/hooks/useChartColorConfig.tsx
filp: Good catch, I was being a bit safe here because at one point I had the provider further down the tree - will fix it.

packages/frontend/src/hooks/useChartColorConfig.tsx
IrakliJani: thanks for explaining 💟 and now it makes sense, although it's not clear when reading the code. I think either using `useRef` will make it explicit, or adding a comment to explain the reasoning behind it.

IrakliJani (approved):
nice work @filp 🎉 

let's make sure that e2e passes before merging. you have to rebase on main first because the build fails puppeteer download. that was recently fixed and merged on main

Issue #9056, opened 2024-02-21
Add a link to docs in the SDs and Alerts Modals


PriPatel:
<img width="942" alt="Screenshot 2024-02-21 at 16 04 07" src="https://github.com/lightdash/lightdash/assets/12776522/b582b2fa-d3e7-48a7-abaf-1b19cc877948">

- [ ] Add a `minimal` docs icon at the bottom-left of the modals. 
- [ ] Add this to the SDs modal linking to https://docs.lightdash.com/guides/how-to-create-scheduled-deliveries
- [ ] Add this to the Alerts modal linking to the new docs

PriPatel:
greattt

PR #9022, opened 2024-02-21
docs: alert docs

rephus:
Closes: #8958

### Description:

https://deploy-preview-9022--peaceful-bassi-cbf284.netlify.app/guides/how-to-create-alerts/

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #9057, opened 2024-02-21
fix: add docs panel

rephus:
Closes:  #9056

### Description:

Screenshot from 2024-02-21 17-19-40
### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/scheduler/components/SchedulerModal.tsx
+++ b/packages/frontend/src/features/scheduler/components/SchedulerModal.tsx
@@ -3,6 +3,7 @@ import { Group, Modal, Text } from '@mantine/core';
 import { IconBell, IconSend } from '@tabler/icons-react';
 import React, { FC } from 'react';
 import MantineIcon from '../../../components/common/MantineIcon';
+import DocumentationHelpButton from '../../../components/DocumentationHelpButton';
 import SchedulerModalContent from './SchedulerModalContent';
 
 const SchedulersModal: FC<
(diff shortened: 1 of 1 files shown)


packages/frontend/src/features/scheduler/components/SchedulerModal.tsx
```diff
@@ -34,11 +35,21 @@ const SchedulersModal: FC<
                     <Group spacing="xs">
                         <MantineIcon icon={IconBell} size="lg" color="gray.7" />
                         <Text fw={600}>Alerts</Text>
+                        <DocumentationHelpButton
+                            href="https://docs.lightdash.com/guides/how-to-create-alerts"
```
rephus, line 39: depends on https://github.com/lightdash/lightdash/pull/9022

packages/frontend/src/features/scheduler/components/SchedulerModal.tsx
```diff
@@ -34,11 +35,21 @@ const SchedulersModal: FC<
...
+                            pos="relative"
+                            top="2px"
+                        />
                     </Group>
                 ) : (
                     <Group spacing="xs">
                         <MantineIcon icon={IconSend} size="lg" color="gray.7" />
                         <Text fw={600}>Scheduled deliveries</Text>
+                        <DocumentationHelpButton
+                            href="https://docs.lightdash.com/guides/how-to-create-scheduled-deliveries"
```
magnew, line 49: Do we think these should open in another window? These only show up in part of the app (this modal) where there is an incomplete form and no saved draft. So if they did already do something in this form, then click that link, they will always loose state. 

@PriPatel any thoughts?

PriPatel: Oh woops! yes these should definitely open in a new tab @rephus @magnew

rephus: this component already opens links in a new tab, we use this component in a few other places . Does it not open in a new tab for you ?

magnew: Re-tested and this does what I expect. I must have been losing it yesterday. Sorry about the confusion!

magnew (approved):

Issue #8999, opened 2024-02-19
I can mention/tag someone in a comment


joaoviana:
### Description

I also want to mention someone in my comment so they can later on acknowledge it. 

Role: `interactive viewer` and above

**Acceptance Criteria**
- [ ] I can clearly see how to mention someone in a comment via a placeholder
- [ ] I can initiate a mention by typing the @ symbol, followed by the letters of the person in my project that I want to mention (add debounced search of at least 200ms)
- [ ] I get a list of users that match my search input
- [ ] I can select a user from the dropdown
- [ ] I can see a list that displays at least 5 users at a time and support scrolling for additional matches.
- [ ] I can see a `no users found` message when my search retrieves no results
- [ ] I can mention multiple users in a comment
- [ ] I receive a notification if a user mentions me and vice-versa (and the notification links to the dashboard)
- [ ] Notifications must be real-time or with minimal delay.
- [ ] Mentioned usernames in comments are visually distinct (e.g., highlighted or a different color) from the rest of the text.

joaoviana:
Assigning myself to work on UI component to render mentions' suggestions - using `react-mentions` https://www.npmjs.com/package/react-mentions

PR #9071, opened 2024-02-22
feat: comment mentions

rephus:
Closes:  #8999

### Description:

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/commentsController.ts
+++ b/packages/backend/src/controllers/commentsController.ts
@@ -21,7 +21,7 @@ import {
     Tags,
 } from '@tsoa/runtime';
 import express from 'express';
-import { commentService, dashboardService } from '../services/services';
+import { commentService } from '../services/services';
 import { allowApiKeyAuthentication, isAuthenticated } from './authentication';
 
 @Route('/api/v1/comments')
--- a/packages/backend/src/models/CommentModel/CommentModel.ts
+++ b/packages/backend/src/models/CommentModel/CommentModel.ts
@@ -1,11 +1,10 @@
 import { Comment, LightdashUser, NotFoundError } from '@lightdash/common';
 import { Knex } from 'knex';
+import { DashboardTileCommentsTableName } from '../../database/entities/comments';
 import {
     DashboardsTableName,
-    DashboardTileCommentsTableName,
     DashboardTilesTableName,
     DashboardVersionsTableName,
-    DbDashboardTileComments,
 } from '../../database/entities/dashboards';
 import { UserTableName } from '../../database/entities/users';
 
@@ -85,7 +84,7 @@ export class CommentModel {
                 dashboard_tile_uuid: dashboardTileUuid,
                 reply_to: replyTo ?? null,
                 user_uuid: user.userUuid,
-                saved_chart_uuid: savedChartUuid,
+                saved_chart_uuid: savedChartUuid ?? null,
             })
             .returning('comment_id');
 
--- a/packages/backend/src/database/entities/comments.ts
+++ b/packages/backend/src/database/entities/comments.ts
@@ -10,11 +10,16 @@ export type DbDashboardTileComments = {
     dashboard_tile_uuid: string;
     user_uuid: string;
     resolved: boolean;
+    saved_chart_uuid: string | null;
 };
 
 type DbDashboardTileCommentsInsert = Pick<
     DbDashboardTileComments,
-    'text' | 'dashboard_tile_uuid' | 'reply_to' | 'user_uuid'
+    | 'text'
+    | 'dashboard_tile_uuid'
+    | 'reply_to'
+    | 'user_uuid'
+    | 'saved_chart_uuid'
 >;
 
 type DbDashboardTileCommentsUpdate = Pick<DbDashboardTileComments, 'resolved'>;


rephus:
Looking for UI feedback, I think we can do better in the UI. 

[Screencast from 22-02-24 17:08:06.webm](https://github.com/lightdash/lightdash/assets/1983672/e952c148-b313-4c7f-90ab-ec7acdbe1e11)

Ideally the `autocomplete` input should be `inline` 

Some ideas: 

- use ComboBox  BUT is not available on v6.  :( 
- a bit hacky, but we could hide the `autocomplete` input behind/hidden. and use the `textarea` value to chnage the value of the autocomplete. Do you know what I mean ? 
- Show the autocomplete input on top of the `message` . To make it look nice, we should calculate the position of the cursor int he existing textarea, I tried adding it in the middle and it doesn't look good. 
- any other component that allows us to use this list ?

packages/frontend/src/features/comments/components/CommentForm.tsx
```diff
@@ -0,0 +1,139 @@
...
+        validate: {
+            text: (value) => {
+                if (value.trim() === '') {
+                    return 'Comment cannot be empty';
+                }
+                return null;
+            },
+        },
+    });
+    const { data: listUsers } = useOrganizationUsers();
```
joaoviana, line 48: we might want to get the project users, no? 🤔

joaoviana:
> Ideally the `autocomplete` input should be `inline`
> 
> Some ideas:
> 
> * use ComboBox  BUT is not available on v6.  :(
> * a bit hacky, but we could hide the `autocomplete` input behind/hidden. and use the `textarea` value to chnage the value of the autocomplete. Do you know what I mean ?

Ooooh this sounds interesting 😆 should be easy tbh 🤔 

> * Show the autocomplete input on top of the `message` . To make it look nice, we should calculate the position of the cursor int he existing textarea, I tried adding it in the middle and it doesn't look good.

I tried to do this with a `Popover` (we can make it more accessible etc) - but haven't played around with the actual positioning! 

https://github.com/lightdash/lightdash/assets/7611706/410c3dc2-f425-4c65-aa38-fb5ccc73cbe9


> * any other component that allows us to use this list?

We've looked at: https://github.com/signavio/react-mentions perhaps we can try this one out?

packages/frontend/src/features/comments/components/CommentForm.tsx
rephus: mmm, I didn't think about that. This is what we already use on user invite (I don't think we have an enpdoint for project users) 
Project users is a subset of organization users. in fact. Project users = org users - members (without project role)

So if you send a message to a user that doesn't have access to a project:
- this user will see a notification about a mention
- if he  clicks on the notification he will get an error saying he doesn't have access

The question is, is that a good thing ? or not. The user can ask the admin for permission if he finds out he doesn'tn have access

On the other hand, if we only show project users, and I want to send a message to a person I know is on the org, but doesn't have access to a project, I'll be wondering why he doesn't appear on the list...

packages/backend/src/controllers/commentsController.ts
```diff
@@ -21,7 +21,7 @@ import {
     Tags,
 } from '@tsoa/runtime';
 import express from 'express';
-import { dashboardService } from '../services/services';
+import { commentService, dashboardService } from '../services/services';
```
joaoviana, line 24: and I think `dashboardService` can be removed

packages/backend/src/models/CommentModel/CommentModel.ts
```diff
@@ -0,0 +1,229 @@
...
+        const commentsWithUsers = await this.database(
+            DashboardTileCommentsTableName,
+        )
+            .leftJoin(
+                UserTableName,
+                `${DashboardTileCommentsTableName}.user_uuid`,
+                '=',
+                `${UserTableName}.user_uuid`,
+            )
+            .select(
```
joaoviana, line 128: .select<
                (DbDashboardTileComments &
                    Pick<DbUser, 'first_name' | 'last_name'>)[]
            >(
```suggestion
            .select(
```

packages/backend/src/models/CommentModel/CommentModel.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    async resolveComment(commentId: string): Promise<void> {
+        await this.database(DashboardTileCommentsTableName)
+            .update({ resolved: true })
+            .where('comment_id', commentId);
+    }
+
+    async getCommentOwner(commentId: string): Promise<string | null> {
+        const result = await this.database(DashboardTileCommentsTableName)
+            .select<Pick<DbDashboardTileComments, 'user_uuid'>>('user_uuid')
```
joaoviana, line 216:
```suggestion
            .select('user_uuid')
```

packages/backend/src/models/CommentModel/CommentModel.ts
```diff
@@ -0,0 +1,229 @@
...
+            .select(
+                `${DashboardTilesTableName}.dashboard_tile_uuid`,
+                `saved_queries.saved_query_uuid`,
+            )
+            .first();
+
+        if (!dashboardTile) {
+            throw new NotFoundError('Dashboard tile not found for dashboard');
+        }
+        return { savedChartUuid: dashboardTile.saved_query_uuid };
```
joaoviana, line 64: this will also work for `dashboarTile`s that have no saved_query_uuid right? I think 🤔 So if they are Looms or Markdowns

rephus: yes, I tested that, they return null, so we insert nothing on `saved_chart_uuid`

image

joaoviana (approved):

Issue #8916, opened 2024-02-13
I want to display filtering options in the search UI


IrakliJani:
### Description

related ticket: https://github.com/lightdash/lightdash/issues/5266

![CleanShot 2024-02-13 at 19 53 09@2x](https://github.com/lightdash/lightdash/assets/962095/da6e30b5-c0bb-4c71-8853-415d76f89170)

### What problem does this solve?

sometimes most of the results gets overcrowded by fields. users want to have an ability to choose a type and only search within that type

IrakliJani:
see if it is possible to use just the mantine's spotlight component. if not - create a new component which is compatible with the mantine one

IrakliJani:
v7 of mantine actually provides more flexibility to construct the search component - but this is blocked since we are on v6

- v6: https://v6.mantine.dev/others/spotlight/
- v7: https://mantine.dev/others/spotlight/

PR #9072, opened 2024-02-22
feat: search v2 type filters

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/8916

### Reviewer actions


packages/frontend/src/features/omnibar/api/search.ts
```diff
@@ -1,15 +1,24 @@
...
     projectUuid: string;
     query: string;
-}) =>
-    lightdashApi<SearchResults>({
-        url: `/projects/${projectUuid}/search/${encodeURIComponent(query)}`,
+    filters?: SearchFilters;
+}) => {
+    const searchParams = filters
+        ? '?' + new URLSearchParams(filters).toString()
+        : '';
```
almeidabbm, line 15:
```suggestion
    const searchParamsString = new URLSearchParams(filters).toString();
    const searchParams = searchParamsString && `?${searchParamsString}`;
```

The current condition will break if `filters` is an empty object. If that ever happens, then searchparams will be `?`

After trying it out, `new URLSearchParams(filters).toString()` will always return `''` if either `filters` is empty object or undefined, so it is safe to call without check and do the check just for the `searchParams` string on the next line

almeidabbm (changes requested):

almeidabbm:
I believe this is a little bug:


https://github.com/lightdash/lightdash/assets/22939015/4da75c8b-a4b9-4523-b4d7-d31a69569d90

almeidabbm:
> I believe this is a little bug:
> 
>  Screen.Recording.2024-02-23.at.09.34.25.mov

fixed here: https://github.com/lightdash/lightdash/pull/9076

packages/frontend/src/features/omnibar/api/search.ts
IrakliJani: fixed

almeidabbm (approved):

Issue #9062, opened 2024-02-22
Dashboard filter label is not applied inside the modal


ZeRego:
### To Reproduce

Apply filter label

<img width="445" alt="Screenshot 2024-02-22 at 11 07 44" src="https://github.com/lightdash/lightdash/assets/9117144/97939e65-7283-46e8-9c62-f8c2ef1f8915">

In view mode we show the custom label in the "pill" but not inside the modal.

<img width="454" alt="Screenshot 2024-02-22 at 11 07 35" src="https://github.com/lightdash/lightdash/assets/9117144/11a0bfb0-0644-4652-acbb-a478dc143820">

### App Version

0.1001.6

jogeshwar01:
@ZeRego Can I take this up?
Just to confirm, the expected title n the modal was `Customers Month` instead of `Customers Created month`, right?

ZeRego:
@jogeshwar01 Yes you can work on it. Expected is the custom label, is this case `Month` as it shows in the filter "pill".

PR #9074, opened 2024-02-22
fix: dashboard filter label inside modal

jogeshwar01:
Closes: #9062 

### Description:

Moved `filterLabel` state from `FilterSettings` to `FilterConfiguration` to use it in `FieldLabel`.

Example - 
Set field label of `Order Date Year` to `Year of order date`

Before:

image

After:

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
@@ -298,7 +298,14 @@ const FilterConfiguration: FC<Props> = ({
                             selectedField && (
                                 <Group spacing="xs">
                                     <FieldIcon item={selectedField} />
-                                    <FieldLabel item={selectedField} />
+                                    {originalFilterRule?.label &&
+                                    !isEditMode ? (
+                                        <Text span fw={500}>
+                                            {originalFilterRule.label}
+                                        </Text>
+                                    ) : (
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -298,7 +300,10 @@ const FilterConfiguration: FC<Props> = ({
                             selectedField && (
                                 <Group spacing="xs">
                                     <FieldIcon item={selectedField} />
-                                    <FieldLabel item={selectedField} />
+                                    <FieldLabel
```
ZeRego, line 303: This component is not supposed to know about filters. 
We also want to keep the field name in edit mode so editors don't lose track of what field they are filtering.
I think we could rollback all the changes and have:
```
                                    {originalFilterRule?.label && !isEditMode ? (
                                        <Text span fw={500}>originalFilterRule.label</Text>
                                    ) : (
                                        <FieldLabel item={selectedField}/>
                                    )}
```

ZeRego (changes requested):

jogeshwar01:
@ZeRego Thanks for the review.
I have updated the code.

View mode - 

image

Edit mode - 

image

ZeRego (approved):
lgtm 🎉 
running tests before merging

Issue #9026, opened 2024-02-21
There should be an option to be 'notified only once' once my threshold has reached a certain value


PriPatel:
If I've set up hourly checks, I don't want emails every hour. In the setup form, there should be a small toggle - 'notify me only once' and this should be turned on by default.

PR #9089, opened 2024-02-23
feat: send once option for threshold alerts

magnew:
Closes: #9026 

### Description:

Note that this is a draft at the moment. It seems to work, but I'd like to test it more rigorously. 

Adds a checkbox to send scheduled alerts just once. This does this by disabling the alert once it's been sent. I added some helper text when the box is turned off. Should we add some text when it is on as well.

Send once state:
Screenshot 2024-02-23 at 17 57 43

Send always state:
Screenshot 2024-02-23 at 17 57 59

### Reviewer actions

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -25,6 +25,7 @@ import {
     isSchedulerImageOptions,
     isTableChartConfig,
     LightdashPage,
+    NotificationFrequency,
     NotificationPayloadBase,
     operatorAction,
     ScheduledDeliveryPayload,
@@ -1049,7 +1050,7 @@ export const sendEmailNotification = async (
     }
 };
 
-const isPositiveThesholdAlert = (
+const isPositiveThresholdAlert = (
     thresholds: ThresholdOptions[],
     results: Record<string, any>[],
 ): boolean => {
@@ -1441,6 +1442,7 @@ export const handleScheduledDelivery = async (
             savedChartUuid,
             dashboardUuid,
             thresholds,
+            notificationFrequency,
         } = scheduler;
         if (thresholds !== undefined && thresholds.length > 0) {
             // TODO add multiple AND conditions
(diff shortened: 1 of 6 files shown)


magnew:
> * I think I prefer to see this with the Alert conditions section? So perhaps at the bottom of it and spaced so it feels like part of that section. Do you also think it makes sense there?

Sure, let's give that a shot

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -1452,7 +1454,22 @@ export const handleScheduledDelivery = async (
                     savedChartUuid,
                 );
 
-                if (isPositiveThesholdAlert(thresholds, rows)) {
+                if (isPositiveThresholdAlert(thresholds, rows)) {
+                    // If the delivery frequency is once, we disable the scheduler.
+                    // It will get sent once this time.
+                    if (
+                        notificationFrequency === NotificationFrequency.ONCE &&
```
rephus, line 1461: Nice

rephus: this means some scheduler events can happen even when `scheduler` is disabled, like send email/slack notifications. We do the disabled check here https://github.com/lightdash/lightdash/pull/9089/files#diff-76827588cb3d2be92b5d1fec14b04f04caf22bbcd510e49a98517da3192d3fcbR1416 so it should not cause any issues. Just something to keep in mind.

rephus (commented):
Code looks good, waiting for render to be fixed to test it

magnew:
> Could you fix the migration issue in render ? this happens if the `migration` file was not in the first commit...

Sure, I kicked up a database reset/migration. That should do it right?

packages/backend/src/scheduler/SchedulerTask.ts
magnew: Good point. I agree I *think* it should be safe. I could make this just set a local variable like disableAfterSending and disable it at the end of this try block. Would that be better or more confusing for DevX?

rephus (approved):
Yeah, it works as expected. 

It is true `enabling` the toggle on render is very slow :thinking: we should fix it somehow... 

This was not introduced by this PR and it not a blocker anyway.

PR #9103, opened 2024-02-26
feat: add mentions UI component

joaoviana:
Relates to: #8999 

### Description:

Adds mentions input made with `mantine`'s Rich Text Editor + tiptap 

Followed this: https://tiptap.dev/docs/editor/api/nodes/mention
and this as well to get the correct TS types: https://github.com/sjdemartini/mui-tiptap/blob/4ff56f8c77e565186656d7cc97adfd1a7960a572/src/demo/mentionSuggestionOptions.ts

### Reviewer actions


packages/frontend/src/features/comments/components/CommentForm.tsx
```diff
@@ -19,27 +22,38 @@ export const CommentForm: FC<Props> = ({
...
+    const handleSubmit = commentForm.onSubmit(async () => {
+        if (commentText === '') return;
 
-        commentForm.reset();
+        // get comment text - commentText
+        // get comment html - commentHtml
+        // get mentions - mentions
+        // submit comment with these 3 values: text, html, mentions
+        // await onSubmit(commentText);
+        setShouldClearEditor(true);
```
joaoviana, line 56: we should submit the form with 3 things: 
1. plain text (in case we need to refer to it in the future/utilise it for comment previews)
2. html (to render the comment with nice formatting, accepting multiline)
3. mentions (an array of strings that will be useful when displaying notification data)

Some people recommend storing the JSON as well - which would be handy for edit functionality. 

https://github.com/ueberdosis/tiptap/discussions/148#discussioncomment-7319

what do you think?

packages/frontend/src/features/comments/components/CommentWithMentions/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+            Placeholder.configure({
+                placeholder: 'Add comment (type @ to mention someone)',
+            }),
+        ],
+        content,
+        onUpdate(e) {
+            setCommentText?.(e.editor.getText());
+            setCommentHtml?.(e.editor.getHTML());
+            setMentions?.(parseMentions(e.editor.getJSON()));
+        },
```
joaoviana, line 66: these should not be called on `onUpdate` but only when the user hits submit

packages/frontend/src/features/comments/components/CommentForm.tsx
rephus: tl;dr let's store plain_text and html :muscle: 

yeah, storing html makes sense, because that contains information that can't be generated from the plain text.
I don't think we need json... I'm thinking on emails, html and plain text is all you need. 

If we need to pick one, I'd choose storing html over plain text, because you can generate plain text from html by removing the html tags, but not the other way around.

packages/frontend/src/features/comments/components/CommentWithMentions/index.tsx
rephus: yeah, that makes sense, is there any problem on doing that ?

packages/frontend/package.json
```diff
@@ -47,6 +48,13 @@
         "@tanstack/react-query-devtools": "^4.36.1",
         "@tanstack/react-table": "^8.5.11",
         "@tanstack/react-virtual": "^3.0.0-beta.54",
+        "@tiptap/extension-link": "^2.2.4",
+        "@tiptap/extension-mention": "^2.2.4",
+        "@tiptap/extension-placeholder": "^2.2.4",
+        "@tiptap/pm": "^2.2.4",
+        "@tiptap/react": "^2.2.4",
+        "@tiptap/starter-kit": "^2.2.4",
+        "@tiptap/suggestion": "^2.2.4",
```
rephus, line 57: do we need all these dependencies ?

packages/frontend/src/features/comments/components/CommentWithMentions/index.tsx
rephus: I made a commit for this, feel free to revert if you don't like it https://github.com/lightdash/lightdash/pull/9103/commits/174572cb8e3d2f379190c46031f7f49a15cf6aa9

joaoviana: nice, that's great, thanks for doing it 🙏

packages/frontend/src/features/comments/components/CommentForm.tsx
joaoviana: >If we need to pick one, I'd choose storing html over plain text, because you can generate plain text from html by removing the html tags, but not the other way around.

great point, didn't think of that. 
Also, `tiptap` provides utils to parse this content on all types: `text`, `html`, and `json`

packages/frontend/package.json
joaoviana: I think we do! 
I'm going to check if we need `@tiptap/pm` one but all the others are required

joaoviana: all required

Issue #9109, opened 2024-02-26
Limit search results to 10 per type


almeidabbm:
### Description

Currently we limit the results of dashboards, spaces and charts to 10 per type. Fields and tables should also have the same limits.

PR #9123, opened 2024-02-27
feat: limit results 10 per type and improve query matching for tables and fields

almeidabbm:
Closes: [#9109](https://github.com/lightdash/lightdash/issues/9109)

### Description:

- Limits the results in fields and tables to 10
- Uses regex to match the properties to the query
- Sorts tables and fields by amount of matches in both fields

Note: `fields` & `tables` are part of explores, therefore there's no way to limit them through the query. In this PR we're first collecting all the items that match a `regex` built from the user query, sorting those results by most amount of matches ( we're not counting duplicates ) and then slicing the top 10 results.

### Reviewer actions


packages/backend/src/models/SearchModel/index.ts
```diff
@@ -326,22 +329,38 @@ export class SearchModel {
...
-        const lowerCaseQuery = query.toLowerCase();
-        return explores
+        // Building regex to match any of the words in the query and then using it to match against the label and description
+        // results are sorted by the number of matches - we create a set out of the matches to remove duplicates
+        const sanitizedQuery = escapeRegExp(query);
+        const splitQuery = compact(
+            Array.from(new Set(sanitizedQuery.split(' '))),
+        );
+
+        const splitQueryRegex = new RegExp(splitQuery.join('|'), 'ig');
```
IrakliJani, line 339: Since we are trying to find the most relevant matches here, it might be beneficial to use Fuse.js, which we already utilize in the frontend and does not incur a performance overhead. It also offers a way to return relevance scores and features a full-text-like matching algorithm, which is even better because it works great with substrings

almeidabbm: After some testing with fuse, we can see an impact on the time it takes to search the fields and tables. This is mostly because we first need to fetch all the results in the explore so that then we can use `fuse` to perform the search.

It can will also get more expensive as the number of `fields` and `tables` grow.
This isn't the case with the `regex` approach because we stop collecting results as soon as the limit is reached - ultimately can produce worse results

Here's the file with the collected times. I added a model with 5000 columns and tried it out with both `fuse` and previous `regex` implementation.

After collecting these, I also tried to add a model with 10000 columns, and could see that `fuse` was even slower, but `regex` remained around the same time.

[times.md](https://github.com/lightdash/lightdash/files/14423050/times.md)

almeidabbm: We ended up deciding going the `regex` route, first collecting all the `fields` and `tables` that matched it and then getting the top 10 results - to have more complete results. This still ended up being faster than fuse (tested with model having 40000 columns):

- `{ sortedTablesLen: 1, sortedFieldsLen: 40000 }`
- `searchTablesAndFields: 20-30ms`

packages/backend/src/models/SearchModel/utils/search.ts
```diff
@@ -0,0 +1,43 @@
...
+export function getFullTextSearchRankCalcSql(
+    database: Knex,
+    tableName: string,
+    searchVectorColumnName: string,
+    query: string,
+) {
+    return database.raw(
+        `ROUND(
+            ts_rank_cd(
+                :searchVectorColumn:,
```
IrakliJani, line 14: 👍

packages/backend/src/models/SearchModel/utils/search.ts
```diff
@@ -0,0 +1,43 @@
...
+            query,
+        },
+    );
+}
+
+export function getRegexFromUserQuery(query: string) {
+    const sanitizedQuery = escapeRegExp(query);
+    const splitQuery = compact(Array.from(new Set(sanitizedQuery.split(' '))));
+
+    return new RegExp(splitQuery.join('|'), 'ig');
```
IrakliJani, line 31: is the order important?

for example:

if field name is:

dodge
and you search for dodge
will `oddge` have the same result?

IrakliJani (approved):
looks good. added one question

packages/backend/src/models/SearchModel/utils/search.ts
almeidabbm: This regex is matching full words, so switching letters around won't give out the same results

PR #9132, opened 2024-02-27
fix: disable sidebar options when fetching new options

ZeRego:
Closes: 

### Description:

Context: when a user selects dimensions we ask dbt semantic layer to return all valid metrics for those dimensions. And the opposite, when a user selects metrics we ask dbt semantic layer to return all valid dimensions for those metrics.

Problem: the user could click on multiple metrics/dimensions before we refreshed the page with the valid ones. This could cause the user to select incompatible fields and throwing an error. 

To fix this, I broke down the call into 2. One to fetch metrics and one to fetch dimensions and disable the sidebar options when we are fetching.

Before:

https://github.com/lightdash/lightdash/assets/9117144/bb1982cb-f87d-429c-b2fa-94ac8ae177f1

After:

https://github.com/lightdash/lightdash/assets/9117144/ed7b8ff2-9af9-499e-99ae-66e2a68bfa72

### Reviewer actions


magnew (approved):
Code looks good. I tested this and it works well. 

My only minor question would be whether we want to use all these names with 'semanticLayer' -- do we think these are always going to be tightly coupled with semantic layer? Or maybe this is actually better than calling it metric flow cause its more general? Obviously not a blocker

ZeRego:
> My only minor question would be whether we want to use all these names with 'semanticLayer' -- do we think these are always going to be tightly coupled with semantic layer? Or maybe this is actually better than calling it metric flow cause its more general? Obviously not a blocker

@magnew "metric flow" was the terminology they had before "semantic layer". 
We should slowly rename everything to semantic layer.

PR #9139, opened 2024-02-27
fix: semantic layer expired query

ZeRego:
Closes: 

### Description:

Context: we need to first create a query and then use the query ID to fetch the results. 
Problem: `@tanstack/react-query` useQuery was caching the query ID so we were trying to fetch the results from an expired query ID.

Before:

https://github.com/lightdash/lightdash/assets/9117144/8925fa5e-b16e-42f7-965f-7dcb0fda4068

After:

https://github.com/lightdash/lightdash/assets/9117144/94bee1cf-8458-496b-840b-35c43d5ba21b

### Reviewer actions


rephus (approved):
I reviewed your video, code makes sense.

Issue #9141, opened 2024-02-28
I can easily track updates to new comments on Dashboard Tiles


joaoviana:
### Description

When users land on a dashboard, they should be able to understand:
1. Which tiles have no comments
2. Which tiles have comments
3. Which tiles have comments where I've been mentioned

**Acceptance Criteria**

**Scenario 1: Tile with No Comments**

- [ ] When a Dashboard Tile has no comments, no comment indicator is visible by default.
- [ ] Upon hovering over the Tile, a comment icon appears indicating the current (neutral) state, but disappears when not hovering.

**Scenario 2: Tile with Existing Comments**

- [ ] When a Dashboard Tile has existing comments (read or unread), the comment icon is permanently visible.
- [ ] Display the total number of comments (including replies and normal comments) on a grey indicator alongside the comment icon.

**Scenario 3: Tile with New Comments Including Mentions**

- [ ] For Tiles with new comments where the user is mentioned, display the comment icon with a total number of comments on a red indicator.
- [ ] Upon opening the comment thread, the indicator turns grey to signify that the new comments have been acknowledged.
- [ ] This also marks the notification related to this mention as read
- [ ] Once read, the dashboard provider is capable of tracking these state changes (from new/mentioned to acknowledged).

PR #9152, opened 2024-02-28
feat: display indicator on dashboard tile comments icon

rephus:
Closes:  #9141

### Description:

Scenario 1: Tile with No Comments

- Show comment button/icon only on hover

Scenario 2: Tile with Existing Comments

- Show comment button/icon always with an indicator of how many comments there are

Scenario 3: Tile with New Comments Including Mentions

- Show comment button/icon always with a red indicator
- When user clicks, all notifications related to that tile that aren't read, will be marked as read
- Indicator turns from red to gray.

### Reviewer actions


rephus:
> When user clicks, all notifications related to that tile that aren't read, will be marked as read

There is no way to know which ones are read, right ? 
Similar to what slack does (new) 

Perhaps for v2 

image

rephus:
I've been testing this, everything works great, I only find a minor bug, I don't think it is worth investigating... 

I left a comment with a mention, the icon then updated  , however, it was not red, but gray, a few seconds later, it turned red. 

This is the first screenshots,

Screenshot from 2024-03-04 09-40-37

A few seconds later, it turned red, you can see the mention (left browser is my user, right is David's) 

Screenshot from 2024-03-04 09-40-42

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -159,6 +160,15 @@ export const DashboardProvider: React.FC<
             !!dashboardCommentsCheck.isDashboardTileCommentsFeatureEnabled &&
             !!dashboardCommentsCheck.userCanViewDashboardComments,
     );
+    const hasTileComments = useCallback(
```
rephus, line 163: nice

packages/frontend/src/features/comments/components/DashboardTileComments.tsx
```diff
@@ -31,6 +41,60 @@ export const DashboardTileComments: FC<
...
+                    !n.viewed &&
+                    n.metadata?.dashboardTileUuid === dashboardTileUuid &&
+                    n.metadata?.dashboardUuid === dashboardUuid,
+            ),
+        [notifications, dashboardTileUuid, dashboardUuid],
+    );
+
+    const showIndicator = comments && comments.length > 0;
+    const indicatorColor = useMemo(() => {
+        if (unreadCommentsForThisTile && unreadCommentsForThisTile.length > 0)
```
rephus, line 64: naming `unreadCommentsForThisTile` is confusing, because it relates to `mentions` , not just comments, right ?

joaoviana:
> > When user clicks, all notifications related to that tile that aren't read, will be marked as read
> 
> There is no way to know which ones are read, right ? Similar to what slack does (new)
> 
> Perhaps for v2
> 
> image

oooh that would be cool! 

I guess we could: 
- check if comment is visible to the user 
- add a way to display that that comment mentions me (like a circle?) so when you view it it disappears and the notification related to that disappears as well

Agree for V2 but should be quick as well

joaoviana:
> I've been testing this, everything works great, I only find a minor bug, I don't think it is worth investigating...
> 
> I left a comment with a mention, the icon then updated , however, it was not red, but gray, a few seconds later, it turned red.
> 
> This is the first screenshot, when the button appeared Screenshot from 2024-03-04 09-40-37
> 
> A few seconds later, it turned red, you can see the mention (left browser is my user, right is David's)
> 
> Screenshot from 2024-03-04 09-40-42

Could it be that when I create a comment, the query for `notifications` isn't validated? Perhaps that should be added 🤔

rephus (commented):
Looks good, I can't approve my own PR

packages/frontend/src/features/comments/components/DashboardTileComments.tsx
joaoviana: oops, yes 😆  will rename

rephus:
After looking at all this code, one thing we could have done is to create a new `CommentProvider` to wrap all the comment logic .

I guess we can do this refactoring in a future PR  if we need it (specially if we add it to charts)

joaoviana:
> After looking at all this code, one thing we could have done is to create a new `CommentProvider` to wrap all the comment logic .
> 
> I guess we can do this refactoring in a future PR if we need it (specially if we add it to charts)

100%, was just thinking that as well 👍

packages/frontend/src/features/comments/components/DashboardTileComments.tsx
joaoviana: I renamed to `unreadNotificationsForTile` because `unreadMentionNotificationsForTile` was too long IMO, lmk what you think

joaoviana (approved):

Issue #9088, opened 2024-02-23
Validate preview project not working


ZeRego:
### Description

This might be the reason why customers see "ghost" validation errors. 

There are 2 problems:
- There is a bug in the CLI that shows the message "Validating preview project ..." but it is not validating the newly created preview project.
- The example in our docs is incorrect. https://docs.lightdash.com/guides/cli/how-to-use-lightdash-validate#configure-github-actions

We should update example to have the command:
`run: lightdash validate --project ${{ steps.start-preview.outputs.project_uuid }}`

And we need to update the message logic to only show the preview message if `config.context?.previewProject === projectUuid`

Replicate with 2 commands:

```shell
lightdash start-preview
lightdash validate
```

Then check what is the preview project uuid and the project uuid that was validated.

### App Version

0.1009.0

PR #9159, opened 2024-02-28
fix: validate preview project

jogeshwar01:
Closes: #9088 

### Description:

Added the condition to only show preview message if `config.context?.previewProject === projectUuid` and updated the command in docs.

The bug was occuring because the `start-preview` command does not affect the current project selected in the CLI in the config which in turn is used by `lightdash validate`. 
Our condition only checks for the presence of `previewProject` and not if its currently selected.

https://github.com/lightdash/lightdash/blob/feae273a5e985223a7eb9487ee6c7df0ff2012bc/packages/cli/src/handlers/validate.ts#L94

If we want to validate preview projects after `start-preview` without the `--project` flag, we can call `setProject` method to update selected project in CLI whenever `start-preview` is run.

### Reviewer actions

### DIFF
--- a/docs/docs/guides/cli/how-to-use-lightdash-validate.mdx
+++ b/docs/docs/guides/cli/how-to-use-lightdash-validate.mdx
@@ -46,7 +46,7 @@ You can use this output to block a deploy on Github actions like this
 - name: Start preview 
   run: lightdash start-preview 
 - name: Validate 
-  run: lightdash validate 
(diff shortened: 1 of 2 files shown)


rephus:
Sorry, but I'm going to close this PR because we were already working on this ticket . To avoid confusion in the future, make sure you ask this ticket to be assigned to you first. https://github.com/lightdash/lightdash/blob/main/.github/CONTRIBUTING.md#how-to-contribute-code-to-lightdash

Issue #3120, opened 2022-08-31
When I have many dimensions, I want to be able to collapse my groups (subtotals)


TuringLovesDeathMetal:
### Description

If I have many pivots, I want to be able to collapse the groups in my pivots. Doing this should re-aggregate the metrics shown in my cells based on the new grouping. 

### User acceptance criteria:

Miro sketch of the whole flow: https://miro.com/app/board/uXjVN6GxSOI=/?moveToWidget=3458764579463536132&cot=14

- [ ] Subtotals are calculated for all dimensions other than the rightmost dimension. 
- [ ] when I enable subtotals, toggles appear for all of my "groupable" dimensions. They are "open" by default. 
- [ ] I'm able to collapse/expand values in a "groupable" dimension column 
- [ ] Only dimension column types have grouping applied (not for metrics or table calculations) 
- [ ] To change the dimensions that are subtotalled, the user needs to reposition the dimensions in the table
Here's a video explaining the three criteria above: https://www.loom.com/share/ab8997f6d0d24df49e8d76a9ed62442f

See miro sketch for the criteria below here: https://miro.com/app/board/uXjVN6GxSOI=/?moveToWidget=3458764579459713948&cot=14
- [ ] This feature is only available for table chart types that are NOT pivoted (we'll deal with this in pivoted charts another time). 
  - [ ] When the table is pivoted, the subtotals option is disabled with a helpful message on hover. 

Miro sketch of these two criteria below here: https://miro.com/app/board/uXjVN6GxSOI=/?moveToWidget=3458764579465742749&cot=14
- [ ] Subtotals only appear for `count` and `sum` metric types.
- [ ] For other metric types and table calculations, the columns appear empty when toggled "shut"  

.
- [ ] Collapsing a group results in only showing the subtotal for that group.  
- [ ] These expanded/collapsed configurations are saved when I save my table or share my table with someone else
- [ ] Subtotals that sum to `0` should show as `0` (not as `-`)
- [ ] NULL values should be treated like `0`s in subtotals
- [ ] Grouping a dimension shouldn't aggregate its values, it should just show the group value. (e.g. for numeric dimensions, I shouldn't sum the values - I should just have the option to group unique values)
- [ ] If there is only one dimension in the table, then turning on `show subtotals` shouldn't affect the table (you need at least 2 or more dimensions for some sort of `collapsing`/`groups` to appear)

TuringLovesDeathMetal:
^This was pulled out from #3114

[...]

PR #9160, opened 2024-02-28
feat: table grouping and subtotals

agha4to:
Closes: https://github.com/lightdash/lightdash/issues/3120

### Description:

Adds a "Show subtotals" checkbox to chart configuration which groups rows based on dimension values, and provides subtotals for sum and count metrics.

image

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/SimpleTable/index.tsx
+++ b/packages/frontend/src/components/SimpleTable/index.tsx
@@ -49,6 +49,7 @@ const SimpleTable: FC<SimpleTableProps> = ({
         getFieldLabel,
         getField,
         showResultsTotal,
+        showSubtotals,
     } = visualizationConfig.chartConfig;
 
     if (isLoading) return <LoadingChart />;
@@ -116,6 +117,7 @@ const SimpleTable: FC<SimpleTableProps> = ({
                 columnOrder={columnOrder}
                 hideRowNumbers={hideRowNumbers}
                 showColumnCalculation={showColumnCalculation}
+                showSubtotals={showSubtotals}
                 conditionalFormattings={conditionalFormattings}
                 footer={{
                     show: showColumnCalculation,
--- a/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
@@ -150,10 +150,12 @@ const GeneralSettings: FC = () => {
         setShowColumnCalculation,
         setShowResultsTotal,
         setShowRowCalculation,
+        setShowSubtotals,
         setShowTableNames,
         showColumnCalculation,
         showResultsTotal,
         showRowCalculation,
+        showSubtotals,
         showTableNames,
     } = chartConfig;
 
@@ -248,6 +250,25 @@ const GeneralSettings: FC = () => {
                         setShowResultsTotal(!showResultsTotal);
                     }}
                 />
+                <Tooltip
+                    disabled={!canUsePivotTable}
+                    label="Subtotals can only be used on tables that aren't pivoted"
+                    w={300}
+                    multiline
+                    withinPortal
+                    position="top-start"
+                >
+                    <Box>
+                        <Checkbox
+                            label="Show subtotals"
+                            checked={showSubtotals}
+                            onChange={() => {
+                                setShowSubtotals(!showSubtotals);
+                            }}
+                            disabled={!!canUsePivotTable}
+                        />
+                    </Box>
+                </Tooltip>
             </Stack>
         </Stack>
     );
--- a/packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
+++ b/packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
@@ -5,11 +5,13 @@ import {
     isNumericItem,
     ResultRow,
 } from '@lightdash/common';
+import { IconChevronDown, IconChevronRight } from '@tabler/icons-react';
 import { flexRender, Row } from '@tanstack/react-table';
 import { useVirtualizer } from '@tanstack/react-virtual';
 import React, { FC } from 'react';
 import { getColorFromRange, readableColor } from '../../../../utils/colorUtils';
 import { getConditionalRuleLabel } from '../../Filters/FilterInputs';
+import MantineIcon from '../../MantineIcon';
 import { ROW_HEIGHT_PX, Tr } from '../Table.styles';
 import { TableContext, useTableContext } from '../TableProvider';
 import BodyCell from './BodyCell';
@@ -61,14 +63,14 @@ const TableRow: FC<TableRowProps> = ({
                 const conditionalFormattingConfig =
                     getConditionalFormattingConfig(
                         field,
-                        cellValue?.value.raw,
+                        cellValue?.value?.raw,
                         conditionalFormattings,
                     );
 
                 const conditionalFormattingColor =
                     getConditionalFormattingColor(
                         field,
-                        cellValue?.value.raw,
+                        cellValue?.value?.raw,
                         conditionalFormattingConfig,
                         getColorFromRange,
                     );
@@ -79,19 +81,20 @@ const TableRow: FC<TableRowProps> = ({
                     getConditionalRuleLabel,
                 );
 
+                const toggleExpander = row.getToggleExpandedHandler();
+                const fontColor =
+                    conditionalFormattingColor &&
+                    readableColor(conditionalFormattingColor) === 'white'
+                        ? 'white'
+                        : undefined;
+
                 return (
                     <BodyCell
                         minimal={minimal}
                         key={cell.id}
                         style={meta?.style}
                         backgroundColor={conditionalFormattingColor}
-                        fontColor={
-                            conditionalFormattingColor &&
-                            readableColor(conditionalFormattingColor) ===
-                                'white'
-                                ? 'white'
-                                : undefined
-                        }
+                        fontColor={fontColor}
                         className={meta?.className}
                         index={index}
                         cell={cell}
@@ -99,14 +102,68 @@ const TableRow: FC<TableRowProps> = ({
                         hasData={!!meta?.item}
                         cellContextMenu={cellContextMenu}
                         isLargeText={
-                            (cellValue?.value.formatted || '').length >
+                            (cellValue?.value?.formatted || '').length >
                             SMALL_TEXT_LENGTH
                         }
                         tooltipContent={tooltipContent}
                     >
-                        {flexRender(
-                            cell.column.columnDef.cell,
-                            cell.getContext(),
+                        {cell.getIsGrouped() ? (
+                            <>
+                                <button
+                                    {...{
+                                        onClick: (e) => {
+                                            e.stopPropagation();
+                                            e.preventDefault();
+                                            toggleExpander();
+                                        },
+                                        style: {
+                                            cursor: row.getCanExpand()
+                                                ? 'pointer'
+                                                : 'normal',
+                                            border: 'none',
+                                            background: 'none',
+                                            float: 'left',
+                                        },
+                                    }}
+                                >
+                                    <MantineIcon
+                                        icon={
+                                            row.getIsExpanded()
+                                                ? IconChevronDown
+                                                : IconChevronRight
+                                        }
+                                        color={fontColor}
+                                        size="xl"
+                                        style={{
+                                            float: 'left',
+                                            margin: -4,
+                                        }}
+                                    />
+                                    <span
+                                        style={{
+                                            color: fontColor,
+                                            marginLeft: 4,
+                                        }}
+                                    >
+                                        ({row.subRows.length})
+                                    </span>
+                                </button>
+                                {flexRender(
+                                    cell.column.columnDef.cell,
+                                    cell.getContext(),
+                                )}
+                            </>
+                        ) : cell.getIsAggregated() ? (
+                            flexRender(
+                                cell.column.columnDef.aggregatedCell ??
+                                    cell.column.columnDef.cell,
+                                cell.getContext(),
+                            )
+                        ) : cell.getIsPlaceholder() ? null : (
+                            flexRender(
+                                cell.column.columnDef.cell,
+                                cell.getContext(),
+                            )
                         )}
                     </BodyCell>
                 );
--- a/packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
+++ b/packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
@@ -1,8 +1,9 @@
 import { Draggable } from '@hello-pangea/dnd';
-import { isField } from '@lightdash/common';
+import { FieldType, isField } from '@lightdash/common';
 import { Tooltip } from '@mantine/core';
 import { flexRender } from '@tanstack/react-table';
-import React, { FC } from 'react';
+import isEqual from 'lodash/isEqual';
+import React, { FC, useEffect } from 'react';
 import {
     TABLE_HEADER_BG,
     Th,
@@ -15,12 +16,42 @@ import { HeaderDndContext, HeaderDroppable } from './HeaderDnD';
 
 interface TableHeaderProps {
     minimal?: boolean;
+    showSubtotals?: boolean;
 }
 
-const TableHeader: FC<TableHeaderProps> = ({ minimal = false }) => {
+const TableHeader: FC<TableHeaderProps> = ({
+    minimal = false,
+    showSubtotals = true,
+}) => {
     const { table, headerContextMenu, columns } = useTableContext();
     const HeaderContextMenu = headerContextMenu;
     const currentColOrder = React.useRef<Array<string>>([]);
+
+    useEffect(() => {
+        if (showSubtotals) {
+            const groupedColumns: any = columns
+                .filter(
+                    (col: any) =>
+                        col.meta?.item?.fieldType === FieldType.DIMENSION,
+                )
+                .map((col: any) => col.id);
+            const sortedColumns: any = table
+                .getState()
+                .columnOrder.reduce((acc: any[], sortedId) => {
+                    if (groupedColumns.includes(sortedId)) acc.push(sortedId);
+                    return acc;
+                }, []);
+            sortedColumns.pop();
+
+            if (!isEqual(sortedColumns, table.getState().grouping))
+                table.setGrouping(sortedColumns);
+        }
+    }, [showSubtotals, columns, headerContextMenu, table]);
+
+    useEffect(() => {
+        if (!showSubtotals) table.resetGrouping();
+    }, [showSubtotals, table]);
+
     if (columns.length <= 0) {
         return null;
     }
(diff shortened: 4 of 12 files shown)


TuringLovesDeathMetal (changes requested):
This is amazing! Well done 🚀 

I've been playing around with it in this PR environment here: 

Looks, and feels really awesome and super excited to share this with people. 

Just three changes from me (sorry, I realize I missed some of these on the acceptance criteria, so this is mostly up to me missing details in the spec!):

- Interactivity on subtotalled cells should be turned off (`copy value`, `show underlying data`, etc. all seem to "pull" from the first value in the subtotal, so I think let's just turn it off for this first version, then we can always add it in later when people ask for it!). The `copy` seems to be working for the `toggled`/`grouped` cell, so we can leave that in if it's easy 👀 

image

- Can we make the number shown in the toggle represent the total number of underlying rows (instead of the number of rows in the _**next**_ grouping level? E.g. the `date month` here, I would expect this to say `19` instead of `4`. 

image

- Can we make it so that `show subtotals` only appears when there are 2 or more dimensions in the results table? (e.g. here, the `show subtotals` option shouldn't appear)

image

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
```diff
@@ -248,6 +250,25 @@ const GeneralSettings: FC = () => {
                         setShowResultsTotal(!showResultsTotal);
                     }}
                 />
+                <Tooltip
+                    disabled={!canUsePivotTable}
```
IrakliJani, line 254: I think the variable name is misleading. We should either introduce another variable that better describes whether the pivot table is enabled or just rename this one to `isPivotTableEnabled` or something similar.

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
```diff
@@ -79,34 +81,89 @@ const TableRow: FC<TableRowProps> = ({
...
                             SMALL_TEXT_LENGTH
                         }
                         tooltipContent={tooltipContent}
                     >
-                        {flexRender(
-                            cell.column.columnDef.cell,
-                            cell.getContext(),
+                        {cell.getIsGrouped() ? (
+                            <>
+                                <button
```
IrakliJani, line 112: this can be a mantine's `ActionIcon` component with minimal styling (https://v6.mantine.dev/core/action-icon/)

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
```diff
@@ -79,34 +81,89 @@ const TableRow: FC<TableRowProps> = ({
...
+                                        },
+                                        style: {
+                                            cursor: row.getCanExpand()
+                                                ? 'pointer'
+                                                : 'normal',
+                                            border: 'none',
+                                            background: 'none',
+                                            float: 'left',
+                                        },
+                                    }}
```
IrakliJani, line 127: I think an extra spread here is not a requirement

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
```diff
@@ -79,34 +81,89 @@ const TableRow: FC<TableRowProps> = ({
...
+                            flexRender(
+                                cell.column.columnDef.aggregatedCell ??
+                                    cell.column.columnDef.cell,
+                                cell.getContext(),
+                            )
+                        ) : cell.getIsPlaceholder() ? null : (
+                            flexRender(
+                                cell.column.columnDef.cell,
+                                cell.getContext(),
+                            )
```
IrakliJani, line 166: Looking at this file, I feel like there's a possibility to clean up some of the styling code and leverage Mantine components more. But we can do that at the end of the stage.

packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
```diff
@@ -15,12 +16,42 @@ import { HeaderDndContext, HeaderDroppable } from './HeaderDnD';
...
+        if (showSubtotals) {
+            const groupedColumns: any = columns
+                .filter(
+                    (col: any) =>
+                        col.meta?.item?.fieldType === FieldType.DIMENSION,
+                )
+                .map((col: any) => col.id);
+            const sortedColumns: any = table
+                .getState()
+                .columnOrder.reduce((acc: any[], sortedId) => {
```
IrakliJani, line 40: This code makes sense, however, it would be nice if we could avoid having `ANY` as it often introduces unnecessary runtime bugs. I'm happy to chime in here and update types.

packages/frontend/src/components/common/Table/TableProvider.tsx
```diff
@@ -75,10 +80,12 @@ const calculateColumnVisibility = (columns: Props['columns']) =>
 export const TableProvider: FC<React.PropsWithChildren<Props>> = ({
     hideRowNumbers,
     showColumnCalculation,
+    showSubtotals,
     children,
     ...rest
 }) => {
     const { data, columns, columnOrder, pagination } = rest;
+    const [grouping, setGrouping] = React.useState<GroupingState>([]);
```
IrakliJani, line 88:
```suggestion
    const [grouping, setGrouping] = useState<GroupingState>([]);
```

packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
```diff
@@ -55,6 +66,32 @@ const getDataAndColumns = ({
             }
             const headerOverride = getFieldLabelOverride(itemId);
 
+            const anyItem = item as any;
```
IrakliJani, line 69: same about `any` type in this function

IrakliJani:
Hey @agha4to! This looks really good! I did the first round of code review and added comments, mostly about minor details. Let me know if you have any questions or need any help. I'll do a second round of review after you make the changes that were mentioned above 🙌

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
agha4to: Tried ActionIcon, but was not rendering nicely with both icon and count label being clickable. Tried UnstyledButton, but settled on Button with variant 'subtle`, but please let me know if we should change it.

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
agha4to: Nice catch!

packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
agha4to: I got rid of the `ANY`'s but not sure my code to replace them is what you have in mind. Please let me know.

TuringLovesDeathMetal:
Nice! looks great :) 

There's just one tiny thing - it looks like we removed the items from the interactivity menu when you click on a subtotal, but there's still a little box that appears. Is there any way we can remove that? 

image

agha4to:
Committed changes for the issues you found:
0) Any non-number in a column now stops subtotal loop and displays blank cell instead of subtotal value
1) Cell dropdown menu artifact removed for empty group cells and subtotal cells
2) Subtotals checkbox shows as unchecked when disabled

TuringLovesDeathMetal (approved):
wahoooo! looks great :)

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
```diff
@@ -0,0 +1,225 @@
...
+// Per TanStack's recommendation:
+// https://tanstack.com/table/v8/docs/guide/row-models#customizefork-row-models
+
+function groupBy<TData extends RowData>(rows: Row<TData>[], columnId: string) {
+    const groupMap = new Map<any, Row<TData>[]>();
+
+    return rows.reduce((map, row) => {
+        // This line changed from original since getGroupingValue does not exist here as it should.
+        // const resKey = `${row.getGroupingValue(columnId)}`
+        const resKey = (row.original as any)[columnId].value.raw;
```
IrakliJani, line 21: can these `any` be `unknown` (without a lot of changes)?

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
```diff
@@ -0,0 +1,225 @@
...
+                const existingGrouping = grouping.filter((columnId) =>
+                    table.getColumn(columnId),
+                );
+
+                const groupedFlatRows: Row<TData>[] = [];
+                const groupedRowsById: Record<string, Row<TData>> = {};
+                // const onlyGroupedFlatRows: Row[] = [];
+                // const onlyGroupedRowsById: Record<RowId, Row> = {};
+                // const nonGroupedFlatRows: Row[] = [];
+                // const nonGroupedRowsById: Record<RowId, Row> = {};
```
IrakliJani, line 54: can we remove commented out code from this file? or maybe we could add a comment for the context why it's commented out?

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -158,13 +162,26 @@ const useTableConfig = (
...
+
+        return columnOrder.filter((fieldId) => {
+            const item = itemsMap[fieldId];
+            return item && isField(item)
+                ? item.fieldType === FieldType.DIMENSION
+                : false;
+        });
+    }, [columnOrder, itemsMap]);
+
+    const canUseSubtotals = dimensions.length > 1;
```
IrakliJani, line 183: can we use `resultsData && resultsData.metricQuery.dimensions.length > 0` instead? if so we could avoid `useMemo` above

packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
```diff
@@ -100,6 +165,24 @@ const getDataAndColumns = ({
...
+                    // Some features work in the TanStack Table demos but not here, for unknown reasons.
+                    // For example, setting grouping value here does not work. The workaround is to use
+                    // a custom getGroupedRowModel.
+                    // getGroupingValue: (row) => { // Never gets called.
+                    //     const value = row[itemId]?.value.raw;
+                    //     return value === null || value === undefined ? 'null' : value;
+                    // },
+                    // aggregationFn: 'sum', // Not working.
+                    // aggregationFn: 'max', // At least results in a cell value, although it's incorrect.
+
```
IrakliJani, line 178: thanks, this is super helpful!

IrakliJani (commented):
Hey @agha4to, I did a thorough review of the code. I feel like we've addressed all aspects of the feature and the bugs as well. This is looking great; I just added a couple of questions/minor suggestions.

Let me know if you have any questions!

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
agha4to: This file is a copy of https://github.com/TanStack/table/blob/main/packages/table-core/src/utils/getGroupedRowModel.ts which has these lines commented out. We can remove the lines though. The only reason to keep them is to make comparison to the source file more straightforward in the future.

packages/frontend/src/components/SimpleTable/index.tsx
```diff
@@ -139,6 +141,13 @@ const SimpleTable: FC<SimpleTableProps> = ({
                                 itemsMap={itemsMap}
                             />
                         );
+                    const cell = props.cell;
+                    if (
+                        cell.getValue() === undefined ||
```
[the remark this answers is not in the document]
agha4to: That's a really nice catch. It's especially good because the changes in this file are no longer needed, since the changes to TableBody.tsx were recently added. Removes redundant code.

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
```diff
@@ -79,34 +92,88 @@ const TableRow: FC<TableRowProps> = ({
...
+                const fontColor =
+                    conditionalFormattingColor &&
+                    readableColor(conditionalFormattingColor) === 'white'
+                        ? 'white'
+                        : undefined;
+
+                const suppressContextMenu =
+                    cell.getValue() === undefined ||
+                    cell.getIsPlaceholder() ||
+                    cell.getIsAggregated();
```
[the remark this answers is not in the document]
agha4to: `getValue()` was needed when this code ran (above) in `SimpleTable/index.tsx` but not here, perhaps due to the recent change where any non-Number value in a column results in subtotal short-circuiting.

packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
```diff
@@ -15,12 +16,44 @@ import { HeaderDndContext, HeaderDroppable } from './HeaderDnD';
...
+                        : false;
+                })
+                .map((col) => col.id);
+            const sortedColumns = table
+                .getState()
+                .columnOrder.reduce((acc: string[], sortedId) => {
+                    if (groupedColumns.includes(sortedId)) acc.push(sortedId);
+                    return acc;
+                }, []);
+            sortedColumns.pop();
```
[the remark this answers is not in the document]
agha4to: The last dimension column essentially groups rows for each unique value in that column. Without the `pop()` each row would have an extra row group containing 1 item.

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
agha4to: `resultsData.metricQuery.dimensions` does not update when dimensions are removed, so if all but one dimension is removed and "Configure" is chosen then the "Show subtotals" checkbox will still be enabled because `resultsData.metricQuery.dimensions` still has the original number of dimensions while the `dimensions` constant calculated here reflects the change to one dimension.

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
agha4to: Yes, looks OK.

agha4to:
@TuringLovesDeathMetal, should the "Show subtotals" checkbox also appear unchecked when there are fewer than 2 dimensions?
image

TuringLovesDeathMetal:
@agha4to - oooh, ya, that would be nice! 

It's not a blocking change from me though - so if you feel like doing it you can :)

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
IrakliJani: fair enough, let's keep them! thanks Tony!

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
IrakliJani: perfect, thanks!

packages/frontend/src/components/common/Table/ScrollableTable/TableHeader.tsx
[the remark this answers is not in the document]
IrakliJani: gotcha, thanks!

packages/frontend/src/components/SimpleTable/index.tsx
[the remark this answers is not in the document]
IrakliJani: perfect!

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
IrakliJani: that makes sense! `useTableConfig` hook is so confusing!

IrakliJani:
@agha4to, thank you for your comment replies and for clarifying my questions. It was super helpful.

I've pushed a few minor updates to your branch:
- Slight style updates to the group collapse button.
- Ran the API generation script for the backend.

---

I noticed rather odd behavior in the minimal pages' table when collapsing/expanding the groups.

"explore page" and "dashboard" seem fine. The bug is only present on "minimal explore page" (which is not supposed to be interactive). However, This raised a valid concern about the nature of the bug and whether it could affect anything else:


https://github.com/lightdash/lightdash/assets/962095/95a07d7d-916c-464b-a198-dc9eaa4c5521

---


I propose that we deploy this behind a feature flag and we do a thorough testing in Workday / Lightdash organizations before announcing it to the public.

cc: @TuringLovesDeathMetal

TuringLovesDeathMetal:
@IrakliJani I'm not sure what you mean 🤔 are you saying that in the `view` mode of a chart, you shouldn't be able to expand/collapse the groups?

IrakliJani:
@TuringLovesDeathMetal 

Apologies if the video was confusing...
Watch closely starting at 0:30 when I collapse/open subgroups - you'll notice the rows get duplicated.

I've been testing tables in various places, and one of the pages I examined is the "minimal explore view" page (the one where we take a screenshot for scheduled deliveries). For what it's worth, if this bug only exists on the minimal page table, then it's not a major issue because that page's interactivity is not crucial (it's intended for screenshots only).

IrakliJani:
just pushed one more commit which fixes the build 🙌

packages/frontend/src/components/common/Table/getGroupedRowModelLightdash.ts
agha4to: Actually, I think it is no big deal to remove them. It does not make the comparison to the original especially difficult.

agha4to:
@IrakliJani, I did not know how to access `minimal` mode until now 🐤. Fixed this issue by using a unique row index for this type of table.

I searched to see if `getRowModel()` might be an issue anywhere else and found when "Show subtotals" and "Show number of results" are both checked, and there is only one page of data, the result total was incorrect. Now fixed.
image

@TuringLovesDeathMetal, changed "Show subtotals" checkbox behavior per above discussion.

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
IrakliJani: this part is very clear. thanks @agha4to !

IrakliJani:
@agha4to thanks for fixing that bug!

I just added few more commits and finished reviewing:
- improved types in random places
- fixed aggregation function bug

before:
CleanShot 2024-03-29 at 17 16 41@2x
after:
CleanShot 2024-03-29 at 17 16 10@2x

IrakliJani:
let me know if my changes make sense to you 🙌 
im going to approve the PR once the e2e pass

agha4to:
Lgtm. The before/after labels are switched above btw.
I introduced that subtotal bug when I switched to `some`, which I almost never use 😞 Nice find! I'm always learning to be more careful.

IrakliJani:
@agha4to cool!

Regarding "non-number sum": I believe it is acceptable to ignore any non-integer values and simply sum-up the remaining ones, as we are aggregating metric values that are either `COUNT` or `SUM`, which should always be numeric (almost always).

Regarding end-to-end testing: There was an issue with the rendering environment. I just reset the database, and now the end-to-end tests are all passing!

IrakliJani (approved):
amazing work @agha4to 🙌

IrakliJani:
@all-contributors add @agha4to for code

Issue #9053, opened 2024-02-21
Sync & scheduled deliveries doesn't show syncs from charts in dashboards


ZeRego:
### To Reproduce

1. create chart in dashboard & save 
2. create google sync for that chart
3. go to Sync & scheduled deliveries

Actual:
I don't see the google sync

Expected:
I should see the google sync

### App Version

0.1000.0

PR #9164, opened 2024-02-29
fix: dashboard chart schedulers not appearing

almeidabbm:
Closes: #9053 

### Description:

`charts` created in `dashboards` were not showing up because they don't belong to a `space` and therefore have `space_id` column set to `NULL`

Since those `charts` belong to a `dashboard` the fix is to join that table in the query to get access to the `space_id` column

### Reviewer actions


magnew (approved):
Good one!

Icons update is just incidental? Do you wanna drop that change from this PR? It doesnt hurt anything I guess, but it's unrelated, right?

almeidabbm:
> ons update is just

yep, 100% unrelated
I just ran `yarn install` after rebasing main and there was this change, removing it

Issue #9148, opened 2024-02-28
Add extended markdown descriptions to models


filp:
Extended the behavior of #1174 to models as well.

Issue #1174, opened 2022-01-19
Viewing longer markdown dbt descriptions


owlas:
### Description: What is it?

Be able to view longer documentation strings in Lightdash and render markdown documentation

### Problem: What problem does this solve?

For complex columns in dbt, some users have very long description fields, even including markdown tables. In Lightdash today the descriptions only show in a small tooltip, which doesn't render markdown and cuts off long descriptions.

### Acceptance criteria

- [ ] I'm able to click `read more` or `see full description` on description boxes. 
- [ ] doing this opens a modal which shows the description with the markdown formatting. 
- [ ] Maybe descriptions which are longer than say 3 lines are cropped and we show this `read more`/`see full description option`? Instead of trying to show the whole description on hover.
- [ ] \

IrakliJani:
still relevant

jpjsonic:
I think the expansion window is a good approach. The one thing I'd keep in mind with the expansion window is that it's pretty common with markdown to use a header followed by plain text.

For example we use the pattern below frequently for our models at Empora:
```
### Purpose
This table rolls up deals to an order level. The order is an umbrella of the deal. You can have situations where a single real estate transaction can have multiple `deals` rows in our system. 

### Filters
None

### Example Use Cases 
1. Revenue reporting is at the file_number grain, so this is the table to use for revenue-order analysis. 
1. Time series of order cancellations over time.
1. Adding order level dimensions to a dashboard. 

### Maintainers
* Jared
{% enddocs %}
```
When people use the headers it could lead to confusion if only the first line is shown. So maybe what you could do is show lines up until and including the first plain text line. So in this case that would look something like:
```
### Purpose
This table rolls up deals to an order level. The order is an umbrella ...
<expand call to action>
```

PR #9167, opened 2024-02-29
feat: Add support for markdown descriptions for models, in the Explore sidebar models list

filp:
Follow-up to #1174 
Ticket: #9148 

### Description:

Extends markdown/longer description support to the Explore sidebar's models list

Kapture 2024-02-29 at 10 42 55
Kapture 2024-02-29 at 13 44 54

- Removes tooltip from breadcrumb for simplicity for now, will re-introduce it with markdown support on a follow-up if deemed necessary, but requires more work to ensure it works well across different breadcrumb uses.
- Adds support for markdown descriptions on models
- Reshuffles some code around to remove duplication around model descriptions + minor cleanup

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreSideBar/index.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreSideBar/index.tsx
@@ -18,6 +18,7 @@ import MantineIcon from '../../common/MantineIcon';
 import PageBreadcrumbs from '../../common/PageBreadcrumbs';
 import SuboptimalState from '../../common/SuboptimalState/SuboptimalState';
 import ExplorePanel from '../ExplorePanel';
+import { ItemDetailProvider } from '../ExploreTree/TableTree/ItemDetailContext';
 import ExploreGroup from './ExploreGroup';
 import ExploreNavLink from './ExploreNavLink';
 
@@ -180,15 +181,17 @@ const ExploreSideBar = memo(() => {
     }, [clearExplore, history, projectUuid]);
 
     return (
-        <TrackSection name={SectionName.SIDEBAR}>
-            <Stack h="100%" sx={{ flexGrow: 1 }}>
-                {!tableName ? (
-                    <BasePanel />
-                ) : (
-                    <ExplorePanel onBack={handleBack} />
-                )}
-            </Stack>
-        </TrackSection>
+        <ItemDetailProvider>
+            <TrackSection name={SectionName.SIDEBAR}>
+                <Stack h="100%" sx={{ flexGrow: 1 }}>
+                    {!tableName ? (
+                        <BasePanel />
+                    ) : (
+                        <ExplorePanel onBack={handleBack} />
+                    )}
+                </Stack>
+            </TrackSection>
+        </ItemDetailProvider>
     );
 });
 
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/ItemDetailPreview.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/ItemDetailPreview.tsx
@@ -1,7 +1,19 @@
-import { Anchor, Box, Flex, Title, useMantineTheme } from '@mantine/core';
+import {
+    Anchor,
+    Box,
+    Flex,
+    Group,
+    Popover,
+    Text,
+    Title,
+    useMantineTheme,
+} from '@mantine/core';
+import { IconTable } from '@tabler/icons-react';
 import ReactMarkdownPreview from '@uiw/react-markdown-preview';
-import { FC } from 'react';
+import { FC, PropsWithChildren } from 'react';
 import { rehypeRemoveHeaderLinks } from '../../../../utils/markdownUtils';
+import MantineIcon from '../../../common/MantineIcon';
+import { useItemDetail } from './ItemDetailContext';
 
 /**
  * Renders markdown for an item's description, with additional constraints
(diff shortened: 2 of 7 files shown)


hamzahc1:
@filp do we want to open a separate issue for the related explores so we can track it?

packages/frontend/src/components/Explorer/ExploreSideBar/index.tsx
```diff
@@ -180,15 +181,17 @@ const ExploreSideBar = memo(() => {
...
-        <TrackSection name={SectionName.SIDEBAR}>
-            <Stack h="100%" sx={{ flexGrow: 1 }}>
-                {!tableName ? (
-                    <BasePanel />
-                ) : (
-                    <ExplorePanel onBack={handleBack} />
-                )}
-            </Stack>
-        </TrackSection>
+        <ItemDetailProvider>
```
filp, line 184: Pulled the provider higher up in the tree to also support model descriptions.

packages/frontend/src/components/Explorer/ExploreTree/TableTree/ItemDetailPreview.tsx
```diff
@@ -0,0 +1,89 @@
+import { Anchor, Box, Flex, Title, useMantineTheme } from '@mantine/core';
```
filp, line 1: Extracted these components out into their own file.

filp:
@hamzahc1 No longer necessary, added the missing piece directly as part of this PR, just required a bit of shuffling code around.

packages/frontend/src/components/Explorer/ExploreTree/TableTree/ItemDetailPreview.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    ta={'center'}
+                    /**
+                     * Forces the 'Read more' option to slightly overlap with the content, and show
+                     * a slight fade effect.
+                     */
+                    mt={-30}
+                    pt={20}
+                    style={{
+                        background:
+                            'linear-gradient(rgba(255,255,255,0) 0%, rgba(255,255,255,1) 60%, rgba(255,255,255,1) 100%)',
```
IrakliJani, line 74: there are some mantine helpers functions to create gradients from mantine colors. although im ok with this css since its just a white color / transparent color. btw I think you can use `transparent` instead of `rgba(255,255,255,1)`.

packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
```diff
@@ -11,121 +11,30 @@ import {
...
-            </Box>
-            {isTruncated && (
-                <Box
-                    ta={'center'}
-                    /**
-                     * Forces the 'Read more' option to slightly overlap with the content, and show
-                     * a slight fade effect.
-                     */
-                    mt={-30}
-                    pt={20}
```
IrakliJani, line 109: do we need this?

packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
```diff
@@ -11,121 +11,30 @@ import {
...
-                <NodeDetailMarkdown source={description} />
-            </Box>
-            {isTruncated && (
-                <Box
-                    ta={'center'}
-                    /**
-                     * Forces the 'Read more' option to slightly overlap with the content, and show
-                     * a slight fade effect.
-                     */
-                    mt={-30}
```
IrakliJani, line 108: To avoid hardcoded values and negative margin-top, we could change this div to absolute positioning and make the popover relative. Then, you can simply use bottom={0}

IrakliJani (commented):
this looks great!

something I noticed:

hovercard seems detached from the "trigger." Maybe adding an arrow will help maintain the connection.
https://github.com/lightdash/lightdash/assets/962095/29c400b4-2c6e-43a0-aef4-3b2ec6a4abc8

filp:
Thanks @IrakliJani - made some tweaks with your suggestions in consideration:

Kapture 2024-03-01 at 01 44 29

Basically:

- Added arrows to the popovers
- Changed the preview fade out effect to not require any sort of absolute positioning or hard-coded negative margins. The effect is instead achieved with a CSS mask image, which has pretty good browser support (and the fallback behavior is absolutely fine).
- Tweaked offsets to prevent the popover from overlapping with the model/table chevron icon.

packages/frontend/src/components/Explorer/ExploreTree/TableTree/ItemDetailPreview.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <Box>
+            <Box
+                mah={120}
+                style={{
+                    overflow: 'hidden',
+
+                    // If we're over the truncation limit, use a mask to fade out the bottom of the container.
+                    maskImage: isTruncated
+                        ? 'linear-gradient(180deg, white 0%, white 80%, transparent 100%)'
+                        : undefined,
```
IrakliJani, line 73: oh nice! I've never used these

filp: Yeah same! I had the idea in the back of my head it was a viable solution, but had never actually used it.

Forgot to mention re: using a mantine helper for gradients: ended up not using them because they are not very flexible in terms of how the gradient steps are spaced out!

IrakliJani: Agreed! It can be useful only if you're using Mantine colors and you don't want to hardcode a HEX. Otherwise, this approach is more readable.

ZeRego:
@all-contributors add @filp for code.

Issue #9122, opened 2024-02-27
I am notified when I've been mentioned in a comment


joaoviana:
### Description

Now that I can comment on a dashboard tile, I also want to be notified if someone mentions me

**Acceptance Criteria**
- [ ] I am notified when someone tags/mentions me in a comment
- [ ] I can see what new notifications I have, and when these were pushed
- [ ] When I click on the notification, it should direct me to the dashboard that the tile belongs to
- [ ] After I click on the notification, and go back to my notifications list, I can visually understand which ones are new and which ones I still need to have a look at
- [ ] The notification should include who commented, when, what dashboard it belongs to and the name of tile where the new comment is in

**Nice-to-have Acceptance Criteria**
- [ ] #9047
- [x] #9048

joaoviana:
Also include `refetching` strategy of comments and notifications here

Issue #9135, opened 2024-02-27
UI improvements to Dashboard comments 


joaoviana:
Internal Slack thread: https://lightdash.slack.com/archives/C04E1CCKN64/p1709049672165599

### Description

1. Manage text inputs visibility. 
Only display reply-type text input when user clicks `reply`

~2. Replies could be expanded by default.~

3. Reduce Text input padding

4. Convert Suggestions' List to a Menu/Menu.Item 

5. Reduce font size of createdAt tooltip

6. Add a top border/shadow to the main comment text input

7. Once a new comment has been added, scroll down to the new comment automatically, so no new information is obscured

8. Review `Scrollable` section, avoiding visible cuts when scrolling through the content

9. Remove background color of Action Icons: Delete and Reply buttons

PR #9168, opened 2024-02-29
feat: display notification if user mentioned in comment

joaoviana:
Closes: #9122 

### Description:

Adds UI for notifications for dashboard comments

Has `viewed`/`read` state

On click, it will direct the user to the dashboard 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
+++ b/packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
@@ -1,28 +1,53 @@
+import { NotificationResourceType } from '@lightdash/common';
 import { Button, Indicator, Menu } from '@mantine/core';
 import { IconBell } from '@tabler/icons-react';
 import { FC } from 'react';
+import { useDashboardCommentsCheck } from '../../../features/comments';
+import {
+    DashboardCommentsNotifications,
+    useGetNotifications,
+} from '../../../features/notifications';
 import {
     useValidation,
     useValidationNotificationChecker,
     useValidationUserAbility,
 } from '../../../hooks/validation/useValidation';
+import { useApp } from '../../../providers/AppProvider';
 import MantineIcon from '../../common/MantineIcon';
 import { ValidationErrorNotification } from './ValidationErrorNotification';
 
 export const NotificationsMenu: FC<{ projectUuid: string }> = ({
     projectUuid,
 }) => {
+    const { user } = useApp();
+
+    // Validator notifications
     const { data: validationData } = useValidation(projectUuid, false);
     const canUserManageValidations = useValidationUserAbility(projectUuid);
     const [hasReadValidationNotification, setHasReadValidationNotification] =
         useValidationNotificationChecker();
-
     const hasValidationErrors = validationData && validationData?.length > 0;
 
+    // Dashboard comments notifications
+    const dashboardCommentsCheck = useDashboardCommentsCheck(user?.data);
+    const userCanViewDashboardComments =
+        !!dashboardCommentsCheck.isDashboardTileCommentsFeatureEnabled &&
+        !!dashboardCommentsCheck.userCanViewDashboardComments;
+    const { data: dashboardCommentsNotifications } = useGetNotifications(
+        NotificationResourceType.DashboardComments,
+        userCanViewDashboardComments,
+    );
     const disableBadge =
-        !canUserManageValidations ||
-        !hasValidationErrors ||
-        hasReadValidationNotification;
+        (!canUserManageValidations ||
+            !hasValidationErrors ||
+            hasReadValidationNotification) &&
+        !dashboardCommentsNotifications?.filter((n) => !n.viewed)?.length;
+
+    const hasValidationNotifications =
+        canUserManageValidations && hasValidationErrors;
+
+    const hasDashboardCommentsNotifications =
+        !!dashboardCommentsNotifications?.length;
 
     return validationData ? (
         <Menu
@@ -55,15 +80,25 @@ export const NotificationsMenu: FC<{ projectUuid: string }> = ({
                 </Button>
             </Menu.Target>
 
-            <Menu.Dropdown>
-                {canUserManageValidations && hasValidationErrors ? (
+            <Menu.Dropdown maw="400px">
+                {hasValidationNotifications && (
                     <ValidationErrorNotification
                         projectUuid={projectUuid}
                         validationData={validationData}
                     />
-                ) : (
-                    <Menu.Item>No notifications</Menu.Item>
                 )}
+
+                {hasDashboardCommentsNotifications && (
+                    <DashboardCommentsNotifications
+                        notifications={dashboardCommentsNotifications}
+                        projectUuid={projectUuid}
+                    />
+                )}
+
+                {!hasValidationNotifications &&
+                    !hasDashboardCommentsNotifications && (
+                        <Menu.Item>No notifications</Menu.Item>
+                    )}
             </Menu.Dropdown>
         </Menu>
     ) : null;
--- a/packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
+++ b/packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
@@ -26,30 +26,45 @@ export const NotificationsMenu: FC<{ projectUuid: string }> = ({
     const canUserManageValidations = useValidationUserAbility(projectUuid);
     const [hasReadValidationNotification, setHasReadValidationNotification] =
         useValidationNotificationChecker();
-    const hasValidationErrors = validationData && validationData?.length > 0;
+    const hasValidationNotifications =
+        validationData && validationData.length > 0;
 
     // Dashboard comments notifications
-    const dashboardCommentsCheck = useDashboardCommentsCheck(user?.data);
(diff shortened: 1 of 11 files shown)


packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
```diff
@@ -1,28 +1,53 @@
...
     const canUserManageValidations = useValidationUserAbility(projectUuid);
     const [hasReadValidationNotification, setHasReadValidationNotification] =
         useValidationNotificationChecker();
-
     const hasValidationErrors = validationData && validationData?.length > 0;
 
+    // Dashboard comments notifications
+    const dashboardCommentsCheck = useDashboardCommentsCheck(user?.data);
+    const userCanViewDashboardComments =
+        !!dashboardCommentsCheck.isDashboardTileCommentsFeatureEnabled &&
```
filp, line 34: Consider enforcing these to be booleans (with a default to `false`) in `useDashboardCommentsCheck` to avoid the `!!` boolean cast.

packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
```diff
@@ -1,28 +1,53 @@
...
+        !!dashboardCommentsCheck.userCanViewDashboardComments;
+    const { data: dashboardCommentsNotifications } = useGetNotifications(
+        NotificationResourceType.DashboardComments,
+        userCanViewDashboardComments,
+    );
     const disableBadge =
-        !canUserManageValidations ||
-        !hasValidationErrors ||
-        hasReadValidationNotification;
+        (!canUserManageValidations ||
```
filp, line 41: Nit: Consider adding a comment on how this condition (broadly) works, to aid readability.

packages/frontend/src/features/notifications/components/DashboardCommentsNotifications.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Menu, useMantineTheme } from '@mantine/core';
+import { IconCircleFilled } from '@tabler/icons-react';
+import { FC, useCallback } from 'react';
+import { useHistory } from 'react-router-dom';
+import MantineIcon from '../../../components/common/MantineIcon';
+import { useUpdateNotification } from '../hooks/useNotifications';
+
+type Props = {
+    projectUuid: string;
+    notifications: Notification[] | undefined;
```
filp, line 11: Can you remove `undefined` from the union here, and forcing the callers to handle an empty array default instead? That would help simplify some of the code in the component itself, and make it less ambiguous.

packages/frontend/src/features/notifications/components/DashboardCommentsNotifications.tsx
```diff
@@ -0,0 +1,65 @@
...
+            });
+
+            history.push(`/projects/${projectUuid}${notification.url}`);
+        },
+        [history, projectUuid, updateNotification],
+    );
+
+    return (
+        <>
+            {!!notifications?.length
```
filp, line 39: Ditto on above: consider not allowing `notifications` to be `undefined` here.

hamzahc1 (changes requested):
CleanShot 2024-02-29 at 12 49 58

- [x] can we replace `mention` with `tag`? It's shorter so the notifications will be shorter

- [ ] There's no info on when the notification is from (In acceptance criteria for ticket it says I should see when I was tagged. 


- When I'm not an admin - I don't see the notification icon in the navbar, so I don't get any notifications:
CleanShot 2024-02-29 at 12 52 57

filp:
@joaoviana unrelated to this, but noticed the reply container can get hidden away out of site. There's a scrollbar, but should consider either forcing the scrollbar to be visible when necessary, allowing the container to grow a bit more vertically, or force the scrolling area to scroll to the bottom to ensure the actions are visible while writing a reply.

image

filp:
@joaoviana Bit of an edge case, don't think it's necessarily a blocker but might also be a quick fix:

If you get notified about a new comment on a dashboard which you already have open, clicking notification unsurprisingly doesn't take you anywhere (since you're already in the right dashboard), but the new comment is not visible until the dashboard is refreshed.

filp (approved):
Looks good from my side, aside from the aspects @hamzahc1 pointed out. Approving so I'm not blocking this, but feel free to re-request review as needed.

joaoviana:
> * [ ]  can we replace `mention` with `tag`? It's shorter so the notifications will be shorter

- this is done 👍 


> * [ ]  There's no info on when the notification is from (In acceptance criteria for ticket it says I should see when I was tagged.

- will add now!

> * When I'm not an admin - I don't see the notification icon in the navbar, so I don't get any notifications:

- this was mishap on my end. done as well

joaoviana:
> @joaoviana unrelated to this, but noticed the reply container can get hidden away out of site. There's a scrollbar, but should consider either forcing the scrollbar to be visible when necessary, allowing the container to grow a bit more vertically, or force the scrolling area to scroll to the bottom to ensure the actions are visible while writing a reply.
> 
> image

100%. This is on the #9135 thanks for reporting!

joaoviana:
> @joaoviana Bit of an edge case, don't think it's necessarily a blocker but might also be a quick fix:
> 
> If you get notified about a new comment on a dashboard which you already have open, clicking notification unsurprisingly doesn't take you anywhere (since you're already in the right dashboard), but the new comment is not visible until the dashboard is refreshed.

oh good spot: added a TODO here https://github.com/lightdash/lightdash/pull/9168/commits/d964d028df8aef5a34ce03b5fdb8adeced42a80d

joaoviana:
@filp I believe I've covered all your comments/suggestions. 

What's left for me to do is to add the timestamp as @hamzahc1 pointed out. I'm just discussing internally where this should placed.

joaoviana:
And timestamp is done
Also small tweak on the notification message copy

hamzahc1:
@joaoviana - I'm not sure if these two bugs should be in this PR or separate issues: 
1. As a viewer, I can see the notification icon, but I can't see the comments 
CleanShot 2024-02-29 at 19 18 25

2. Probably a separate issue but the comment box keeps expanding if you type more 😬 
CleanShot 2024-02-29 at 19 15 50

hamzahc1:
I can confirm though that I did get notifications as expected when I was tagged ✅

joaoviana:
> @joaoviana - I'm not sure if these two bugs should be in this PR or separate issues:
> 
> 1. As a viewer, I can see the notification icon, but I can't see the comments

Thanks for pointing this out. I didn't notice it because I tested a lot with markdowns and looms. But didn't check for chart tiles recently. This is fixed now here: [fix: get correct check for viewers and fix empty comment message for …](https://github.com/lightdash/lightdash/pull/9168/commits/e786b11a9a03c1463103556678f471dedfd35b5b)

> 2. Probably a separate issue but the comment box keeps expanding if you type more 😬

Yes :/ I believe it's just a prop we need to pass, so that's good 💪

packages/frontend/src/features/comments/hooks/useDashboardCommentsCheck.ts
```diff
@@ -5,16 +5,16 @@ import { UserWithAbility } from '../../../hooks/user/useUser';
 export const useDashboardCommentsCheck = (
     user: UserWithAbility | undefined,
 ) => {
-    const isDashboardTileCommentsFeatureEnabled = useFeatureFlagEnabled(
+    const isDashboardTileCommentsFeatureEnabled = !!useFeatureFlagEnabled(
```
filp, line 8: This should be updated to use Lightdash's own `useFeatureFlagEnabled` hook (under `src/hooks`), which is correctly typed with the `FeatureFlags` enum and always returns a `boolean` (and not `boolean | undefined`).

packages/frontend/src/features/notifications/components/DashboardCommentsNotifications.tsx
```diff
@@ -1,14 +1,35 @@
...
-    notifications: Notification[] | undefined;
+    notifications: Notification[];
+};
+
+const NotificationTime: FC<{ createdAt: Date }> = ({ createdAt }) => {
+    const date = useTimeAgo(createdAt);
+    return (
+        <Tooltip
+            position="top-end"
+            offset={-2}
```
filp, line 21: Consider adding a note on why this offset is used

packages/frontend/src/features/notifications/components/DashboardCommentsNotifications.tsx
```diff
@@ -36,30 +57,30 @@ export const DashboardCommentsNotifications: FC<Props> = ({
...
-                          onClick={() =>
-                              handleOnNotificationClick(notification)
-                          }
-                          fz="xs"
-                      >
-                          {notification.message}
-                      </Menu.Item>
-                  ))
-                : null}
+            {notifications.map((notification) => (
```
filp, line 60: 🚀 💅

filp (approved):
A couple of minor comments; thanks for addressing my previous feedback!

packages/frontend/src/features/comments/hooks/useDashboardCommentsCheck.ts
joaoviana: done!

hamzahc1 (approved):

Issue #9155, opened 2024-02-28
Use dependency injection for "analytics"


ZeRego:
### Description

Do not import directly `analytics` to services and others. It should be passed as an argument when instantiating a class.

Example of what not to do:
```import { analytics } from '../../analytics/client';```

PR #9176, opened 2024-02-29
fix: analytics dependency injection

ZeRego:
Closes: [#9155](https://github.com/lightdash/lightdash/issues/9155)

### Description:

Changes:
- refactor all services to get analytics via constructor argument
- create analytics instances where needed ( we have separate tickets to address those files )
- amend tests/mocks

Outstanding work:
All the places that create an analytics instance will soon be removed. We have separate tickets to address those files refactors.
Screenshot 2024-02-29 at 14 55 38

### Reviewer actions


packages/backend/setupJest.ts
```diff
@@ -1,22 +1,8 @@
...
-        writeKey: 'notrack',
-        dataPlaneUrl:
-            'notrack',
-        options: {
-            enable: false,
-        }
-    }),
-    identifyUser: jest.fn(),
-}));
+}));
```
filp, line 8: Should this have an empty final newline?

filp (approved):
Not necessarily light reading, but looks good! Did some sanity testing, but putting a good amount of trust on the automated testing in this case.

packages/backend/setupJest.ts
ZeRego: We never had a preference or lint rule for it. We can adjust prettier/eslint to enforce one going forward if you want.

Issue #9165, opened 2024-02-29
Re-introduce removed feature flag logic after Posthog outage


filp:
Follow-up to #9154 

## Tasks:

- Validate and potentially increase `350ms` timeout on feature flag checks
- Re-introduce removed feature flag checks, via `isFeatureFlagEnabled` helper

PR #9186, opened 2024-03-01
chore: Revert posthog-outage related changes, migrate feature flags to `isFeatureFlagEnabled` helper

filp:
Closes: #9165 
Follow-up to: #9154 

### Description:

- Slightly increases the timeout value for the Posthog feature flag check, and adds an escape hatch via an env var to control the timeout value.
- Migrates `PuppeteerScrollElementIntoView` and `PuppeteerSetViewportDynamically` to the `FeatureFlags` enum, and re-adds the checks for these flags using the `isFeatureFlag` enabled helper
- Re-enables the feature flag check for `UseInMemoryTableCalculations`

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -4,6 +4,7 @@ import {
     AuthorizationError,
     ChartType,
     DownloadFileType,
+    FeatureFlags,
     ForbiddenError,
     LightdashPage,
     SessionUser,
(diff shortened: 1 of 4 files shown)


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -375,8 +376,16 @@ export class UnfurlService {
         const startTime = Date.now();
         let hasError = false;
 
-        const isPuppeteerSetViewportDynamicallyEnabled = false;
-        const isPuppeteerScrollElementIntoViewEnabled = false;
+        const isPuppeteerSetViewportDynamicallyEnabled =
+            await isFeatureFlagEnabled(
+                FeatureFlags.PuppeteerSetViewportDynamically,
+                { userUuid },
```
ZeRego, line 382: Before we were passing the organization uuid. Is that not needed? 

```
        const isPuppeteerSetViewportDynamicallyEnabled =
            (await postHogClient?.isFeatureEnabled(
                'puppeteer-set-viewport-dynamically',
                userUuid,
                organizationUuid !== undefined
                    ? {
                          groups: {
                              organization: organizationUuid,
                          },
                      }
                    : {},
            )) ?? false;
```

ZeRego (approved):

packages/backend/src/services/UnfurlService/UnfurlService.ts
filp: Good catch - the helper supports it but it's optional; I should be passing it explicitly here.

Issue #9173, opened 2024-02-29
refactor scheduler/SchedulerTask.ts to class to allow DI


ZeRego:
Dependencies (lightdash config, analytics, services) should be injected instead of imported directly.

PR #9195, opened 2024-03-01
fix: scheduler dependency injection

ZeRego:
Closes: #9173 

### Description:

No business logic was changed.

Changes:
- created a SchedulerTask class
- moved all functions inside that class
- SchedulerWorker class now extends SchedulerTask class
- fixed functions to use class properties instead of importing services/clients

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/services.ts
+++ b/packages/backend/src/services/services.ts
@@ -61,7 +61,7 @@ import { UserAttributesService } from './UserAttributesService/UserAttributesSer
 import { UserService } from './UserService';
 import { ValidationService } from './ValidationService/ValidationService';
 
-export const analytics = new LightdashAnalytics({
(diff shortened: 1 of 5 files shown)


packages/backend/src/services/services.ts
```diff
@@ -61,7 +61,7 @@ import { UserAttributesService } from './UserAttributesService/UserAttributesSer
 import { UserService } from './UserService';
 import { ValidationService } from './ValidationService/ValidationService';
 
-export const analytics = new LightdashAnalytics({
+const analytics = new LightdashAnalytics({
```
ZeRego, line 64: This was a typo in a previous PR. This instance should not be exposed.

rephus (approved):

Issue #9192, opened 2024-03-01
As a viewer, I should be able to view comments, but not create/reply


joaoviana:
### Description

**As a viewer**

I should be able to:
- View comments
- Get notifications if I've been tagged

I should not be able to: 
- Delete comments
- Create comments
- Reply to comments

PR #9196, opened 2024-03-01
fix: viewers cant reply to dashboard comments

joaoviana:
Closes: #9192 

### Description:

Change ability checker in packages/frontend/src/features/comments/hooks/useDashboardCommentsCheck.ts:
Instead of: "checking if user can `manage` dashboard comments", we now check if the user can `create` dashboard comments.

We only need the `manage` ability to check if the user can remove **any** comment - and that is already covered in the `canRemove` property sent back from the BE.

When a user can only `view` comments, they basically can only view comments. So essentially they can't `create`, `reply`, or `delete` comments

### Reviewer actions


rephus (approved):
Reviewed and tested

Issue #9191, opened 2024-03-01
When searching for a member user, the delete button is disabled.


magnew:
### Description

When searching for a 'member' user, the delete button is disabled. It is not disabled when scrolling to the user. Editors and admins don't seem to have this problem.

PR #9201, opened 2024-03-01
fix: dont disable delete on user search

magnew:
Closes: #9191 
### Description:

A user reported they couldnt delete users they had searched for. This is a tiny fix. I don't know why the code was like this, but I everything seems to work well now. 

You can test by searching for a user. If there was only one search result, the delete button was being disabled. Now they should be deletable. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
+++ b/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
@@ -398,7 +398,7 @@ const UsersView: FC = () => {
                                     disabled={
                                         user.data?.userUuid ===
                                             orgUser.userUuid ||
-                                        organizationUsers.length <= 1
+                                        organizationUsers.length < 1
                                     }
                                 />
                             ))


ZeRego:
@magnew context: this check was to disable the button when there was just 1 member in the org. Since we introduced search, we might need the check to use the original list with all org members instead of the filtered list.

joaoviana:
On it! 
Will think of a suggestions after @ZeRego's comment

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
```diff
@@ -398,7 +398,7 @@ const UsersView: FC = () => {
                                     disabled={
                                         user.data?.userUuid ===
                                             orgUser.userUuid ||
-                                        organizationUsers.length <= 1
+                                        organizationUsers.length < 1
```
joaoviana, line 401: Perhaps we can pass a new param to `useOrganizationUsers` to then set the total number of users on `select`, regardless of the search input. So we can

```ts
  disabled={
                                        user.data?.userUuid ===
                                            orgUser.userUuid || totalNumberOfUsers === 1
```

magnew: @joaoviana and @ZeRego, the admin user can never delete themself from the UI, so there will always be at least one. Isn't that sufficient? If you would like to be more careful here, we can, but I think this case is handled. 

You also can't change the last admin to not be an admin anymore

Screenshot 2024-03-04 at 15 22 01

joaoviana: Oh I see @magnew

I just tested this: 
- Couldnt' delete myself
- Searched for me and couldn't delete myself
- Searched for others and could delete them
- Deleted everyone except me 
- Couldn't delete myself, and couldn't delete myself while searching

joaoviana (approved):

Issue #9179, opened 2024-02-29
Field names not formatted correctly on "filter to"


owlas:
Relates to https://github.com/lightdash/lightdash/issues/9051

In `FilterDashboardTo` we directly format the fieldId like `friendlyName(fieldId)` but it should be `${friendlyName(tableLabel)} ${friendlyName(fieldLabel)}` or use the labels directly.

Using friendly name fails on `table_MY_COLUMN` giving => `Table M Y C O L U M N`

PR #9202, opened 2024-03-01
fix: use table and field names for FilterDashboardTo component

almeidabbm:
Closes: #9179 
### Description:

- Uses table name and field name in `FilterDashboardTo` component

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
@@ -1,5 +1,5 @@
 import {
-    DashboardFilterRule,
+    FilterDashboardToRule,
     FilterOperator,
     friendlyName,
 } from '@lightdash/common';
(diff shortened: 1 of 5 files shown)


almeidabbm:
Failing test is related to: https://github.com/lightdash/lightdash/issues/9207

packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
```diff
@@ -24,7 +24,8 @@ export const FilterDashboardTo: FC<Props> = ({ filters, onAddFilter }) => (
                 icon={<MantineIcon icon={IconFilter} />}
                 onClick={() => onAddFilter(filter, true)}
             >
-                {friendlyName(filter.target.fieldId)} is{' '}
+                {friendlyName(filter.target.tableName)} -{' '}
+                {friendlyName(filter.target.fieldName)} is{' '}
```
owlas, line 28: will this work for existing filters that are stored in the database but don't have the `target.fieldName` property?

almeidabbm: Good catch! I believe this is only working for filters without `fieldName` because of the `any` in `lightdash/packages/backend/src/controllers/savedChartController.ts:77` and the fact that we don't use it to actually get the results in the backend.

I can try and instead of adding `fieldName` to the filters type, extract it from the `fieldId`. What do you think?

almeidabbm: @owlas & @filp I changed this type a bit and made `fieldName` a nullable property, this way there's no need for any change in the backend

also, since the filters of `FilterDashboardTo` component come from `createDashboardFilterRuleFromField` frontend function which depends on field, `fieldName` will always exist

PR #9205, opened 2024-03-01
chore: apply ui improvements to dashboard comments

joaoviana:
Closes: #9135 

### Description:

- Smaller comment timestamp tooltip
- Use `Menu` component to list user mentions
- Address width issues - fixed and popover doesn't increase width if the comment text is looooong
- Make icon to reply more subtle, but also make it less easy to delete a comment (under another menu)
- Add a top border to the main comment input
- Increase height of the comment list
- Bolder comment author names
- Scroll to the bottom every time there's a new comment added by the user - this does not apply to replies
- Only show the reply text input if I hit reply
- Less padding on the text input (mimics current text inputs)
- Suggestions list (users that I want to tag) can be scrollable with a maxheight as well

https://github.com/lightdash/lightdash/assets/7611706/84b201e2-57fb-4623-a5a1-38dc9f4ff2f1

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/comments/components/DashboardTileComments.tsx
+++ b/packages/frontend/src/features/comments/components/DashboardTileComments.tsx
@@ -7,8 +7,9 @@ import {
     Stack,
     Text,
 } from '@mantine/core';
+import { useScrollIntoView } from '@mantine/hooks';
 import { IconMessage } from '@tabler/icons-react';
-import { FC } from 'react';
+import { FC, useCallback } from 'react';
 import MantineIcon from '../../../components/common/MantineIcon';
 import { useApp } from '../../../providers/AppProvider';
 import { useDashboardContext } from '../../../providers/DashboardProvider';
@@ -21,6 +22,8 @@ import { DashboardCommentAndReplies } from './DashboardCommentAndReplies';
 type Props = {
     dashboardTileUuid: string;
 };
+
+const COMMENTS_LIST_MAX_HEIGHT = 500;
 
 export const DashboardTileComments: FC<
     Props & Pick<PopoverProps, 'opened' | 'onClose' | 'onOpen'>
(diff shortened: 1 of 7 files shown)


rephus:
I can see myself (david) 3 times in the list of mentions, is this  abug ? 
image

rephus:
You can see how at the bottom, there is a record for a user I invited, but he hasn't registered (so it has no names) , we should filter out user without names or pending ones

image

rephus:
This PR fixed the weird scroll bar :+1:

joaoviana:
> I can see myself (david) 3 times in the list of mentions, is this abug ? FOr some reason, only the first one works image

100% a change that I didn't revert (to test the long list of users) Will revert now!

joaoviana:
> You can see how at the bottom, there is a record for a user I invited, but he hasn't registered (so it has no names) , we should filter out user without names or pending ones
> 
> image

Oh yes, that's true. Will check this out :)

rephus:
Is this `...` menu for the delete icon on a thread necessary ? :thinking: 

image

packages/frontend/src/features/comments/components/DashboardTileComments.tsx
```diff
@@ -29,12 +40,47 @@ export const DashboardTileComments: FC<
         (c) => c.dashboardComments && c.dashboardComments[dashboardTileUuid],
     );
 
+    // Scroll to the last comment when a new comment is added
+    const { scrollIntoView, targetRef, scrollableRef } =
```
rephus, line 44: nice

rephus (approved):
LGTM

joaoviana:
> Is this `...` menu for the delete icon on a thread intended/necessary ? 🤔
> 
> image

This is what I could think of from looking at other examples + just wanted to make it less easy to delete a comment. 
I guess we could refine this later but this was what came to my mind 😆

Issue #9169, opened 2024-02-29
Do not instantiate services within controllers


filp:
Investigate using tsoa's DI facilities for service/dependency injection.

Issue #9151, opened 2024-02-28
Use dependency injection for services in controllers/routes


ZeRego:
### Description

Stop instantiating service classes in services/index.ts and then importing them in the controllers/routes.

Instead, these dependencies should be injected in the controller/route classes

filp:
Unintentionally closed via a related PR - the routes part of this ticket is still missing.

PR #9232, opened 2024-03-04
chore: Add service container support, new BaseController class

filp:
Part of #9169, #9151

### Description:

Adds foundational changes to allow instantiating services via a service container. This is done via:

- A new `BaseController` class that extends `tsoa`'s own `Controller`
- A new `tsoaServiceContainer` module that `tsoa` calls to instantiate controllers.

This PR doesn't do much on its own, besides introduce the above. Follow-up work will be part of separate PRs; but the two main things to do are:

- Update controllers to use `BaseController`
- Update those controllers to use `this.services` to access services
- Update `ServiceRepository` to be able to instantiate all services, and provide a `ServiceRepository` singleton (for now) as part of the controller instantiation

### Reviewer actions

### DIFF
--- a/packages/backend/tsoa.yml
+++ b/packages/backend/tsoa.yml
@@ -1,41 +1,42 @@
 entryFile: src/index.ts
 controllerPathGlobs:
-  - src/**/*Controller.ts
+    - src/**/*Controller.ts
 spec:
-  name: Lightdash API
-  description: >
-    Open API documentation for all public Lightdash API endpoints. 
-    # Authentication
-    Before you get started, you might need to create a Personal Access Token to authenticate via the API. You can create a token by following this guide: https://docs.lightdash.com/references/personal_tokens
-  contact:
-    name: Lightdash Support
-    email: support@lightdash.com
-    url: "https://docs.lightdash.com/help-and-contact/contact/contact_info/"
-  license: MIT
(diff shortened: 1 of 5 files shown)


packages/backend/tsoa.yml
filp, line 1: Auto-formatted with prettier, using the default prettier settings for `backend`.

packages/backend/tsoa.yml
```diff
@@ -1,41 +1,42 @@
...
+            name: connect.sid
+        api_key:
+            type: apiKey
+            in: header
+            name: Authorization
+            description: Value should be 'ApiKey <your key>'
 routes:
-  routesDir: ./src/generated/
+    routesDir: ./src/generated/
+    iocModule: ./src/services/tsoaServiceContainer
```
filp, line 42: This is the relevant part for this file - simply tell `tsoa` to use our service container.

ZeRego (approved):

ZeRego:
@filp Do we need to regenerate the tsoa files after these changes?

filp:
@ZeRego I don't think so, but I don't trust that I fully understand `tsoa` yet, so I'll rebase with `main` and run the api generation here too (and do the same for the follow-up PR that also touches `tsoa`)

PR #9233, opened 2024-03-04
chore: update controllers to use new `BaseController`

filp:
Follow-up to #9232 

_Will rebase before merging, once the above is merged._

### Description:

Updates all controllers to use `BaseController`. This essentially does not change anything in the controllers' behavior, and depends on work that'll be part of a follow-up PR. See the related PR above for more context.

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #9234, opened 2024-03-04
chore: update controllers to use injected `ServiceRepository` for service access

filp:
**See #9232 for the full context.**

This PR is a follow-up to #9233, and will be rebased once that (and its parent) are merged.
Closes: #9169
Implements part of #9151 

### Description:

Updates all controllers to use the injected `ServiceRepository` to access services, instead of relying on the singleton export from `./services`.

This is functionally the same as before, since we're ultimately using our singleton services as part of the `ServiceRepository` setup, but this allows us to later move forward to proper dependency injection.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ServiceRepository.ts
+++ b/packages/backend/src/services/ServiceRepository.ts
@@ -1,1 +1,204 @@
-export class ServiceRepository {}
+import type { AnalyticsService } from './AnalyticsService/AnalyticsService';
+import type { CommentService } from './CommentService/CommentService';
+import type { CsvService } from './CsvService/CsvService';
+import type { DashboardService } from './DashboardService/DashboardService';
+import type { DownloadFileService } from './DownloadFileService/DownloadFileService';
+import type { EncryptionService } from './EncryptionService/EncryptionService';
+import type { GdriveService } from './GdriveService/GdriveService';
+import type { GithubAppService } from './GithubAppService/GithubAppService';
+import type { GitIntegrationService } from './GitIntegrationService/GitIntegrationService';
+import type { GroupsService } from './GroupService';
+import type { HealthService } from './HealthService/HealthService';
+import type { NotificationsService } from './NotificationsService/NotificationsService';
+import type { OrganizationService } from './OrganizationService/OrganizationService';
+import type { PersonalAccessTokenService } from './PersonalAccessTokenService';
+import type { PinningService } from './PinningService/PinningService';
+import type { ProjectService } from './ProjectService/ProjectService';
+import type { SavedChartService } from './SavedChartsService/SavedChartService';
+import type { SchedulerService } from './SchedulerService/SchedulerService';
+import type { SearchService } from './SearchService/SearchService';
+import type { ShareService } from './ShareService/ShareService';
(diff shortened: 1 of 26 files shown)


packages/backend/src/services/ServiceRepository.ts
```diff
@@ -1 +1,204 @@
...
+    unfurlService: UnfurlService;
+    userAttributesService: UserAttributesService;
+    userService: UserService;
+    validationService: ValidationService;
+}
+
+/**
+ * Enforces the presence of getter methods for all services declared in the manifest.
+ */
+type ServiceFactoryMethod<T extends ServiceManifest> = {
```
filp, line 65: This is used to ensure you get a type error if `ServiceRepository` is not up to date with the service manifest above, and also to enforce the specific format of the method name (`get<serviceName` camelcased).

Screenshot 2024-03-04 at 20 37 40

packages/backend/src/services/ServiceRepository.ts
```diff
@@ -1 +1,204 @@
...
+ * via dependency injection.
+ *
+ * NOTE: For now, this repository simply exposes services instantiated in `./services.ts`,
+ *       and provided to this repository directly. At a later stage, this repository will
+ *       handle instantiating all services internally, including cross-service dependencies.
+ */ export class ServiceRepository
+    extends ServiceRepositoryBase
+    implements ServiceFactoryMethod<ServiceManifest>
+{
+    public getAnalyticsService(): AnalyticsService {
```
filp, line 101: This whole section for now is pretty pointless, until `ServiceRepository` is actually instantiating services. The individual methods allow us to have fine control over how each service is instantiated, and for e.g test mocks or overrides to extend `ServiceRepository` and override service getters.

packages/backend/src/services/services.ts
```diff
@@ -282,3 +283,37 @@ export const commentService = new CommentService({
...
+        schedulerService,
+        searchService,
+        shareService,
+        spaceService,
+        sshKeyPairService,
+        unfurlService,
+        userAttributesService,
+        userService,
+        validationService,
+        notificationService: notificationsService,
```
filp, line 316: Opted to stay consistent with the service class name.

packages/backend/src/services/tsoaServiceContainer.ts
```diff
@@ -28,10 +29,10 @@ const isTsoaControllerCtor = (
  */
 export const iocContainer: IocContainerFactory = () => ({
     async get<T extends RouteControllerKlass>(Ctor: T) {
-        if (isTsoaControllerCtor(Ctor)) {
-            return new Ctor();
+        if (isBaseControllerCtor(Ctor)) {
```
filp, line 32: Inverted check to ensure we're doing the right thing here, now that this is important (`BaseController` extends `tsoa`'s `Controller`, so the order is important).

packages/backend/src/services/ServiceRepository.ts
ZeRego: Nice typing 🏅

ZeRego (approved):
Nice atomic PR's ❤️

PR #9236, opened 2024-03-05
fix: validate preview cli

rephus:
Closes: #9088

### Description:

Without --preview 
Screenshot from 2024-03-05 11-05-34

with --preview 

image

### Reviewer actions

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -525,6 +525,7 @@ program
     .option('--vars <vars>')
     .option('--threads <number>')
     .option('--no-version-check')
+    .option('--preview', 'Validate the last preview if available', false)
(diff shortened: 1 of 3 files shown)


packages/cli/src/index.ts
```diff
@@ -525,6 +525,7 @@ program
     .option('--vars <vars>')
     .option('--threads <number>')
     .option('--no-version-check')
+    .option('--preview', 'Validate the last preview if available', false)
```
rephus, line 528: for now I'm only doing this on `validate` and not on `deploy` because I?m not sure if that will break some existing deploys for users . we can do it on a separate PR

ZeRego (approved):

Issue #8765, opened 2024-01-30
if my dimension is filtered via user attribute, then derived metrics should not show


rephus:
### Description

> for the example given in the docs, filtering out salary dimension, but if you're making that into a metric in the yaml also for sum of salaries by teams etc, then that not being filtered out means everyone would have access to it

thread: https://lightdash.slack.com/archives/C0415AUACGY/p1706568471066729?thread_ts=1706491418.004739&cid=C0415AUACGY

These should be filtered  on compile time in the explore.

TuringLovesDeathMetal:
@rephus I think this is kind of related to this? 

https://github.com/lightdash/lightdash/issues/8472

I feel like implementing #8472 would solve both of these issues? I'm wondering if people would always expect metrics to be filtered (or if there's actually a situation where they would want someone to have access to the metric, but not the dimension?)

TuringLovesDeathMetal:
^Also, what about metrics that use custom SQL? (which doesn't even reference the dimension). Or metrics that reference the dimension, but are defined at the model level?

PR #9237, opened 2024-03-05
fix: filter metrics from hidden dimension

rephus:
Closes: #8765

### Description:

Before: I filter a dimension, this dimension has a derived metric, this metric is not filtered

Screenshot from 2024-03-05 09-21-14

Screenshot from 2024-03-05 09-21-42

AFter: 

Metrics and dimensions are both filtered
Screenshot from 2024-03-05 09-47-16

Once I enable user attribute, both appear again

Screenshot from 2024-03-05 09-47-40

### Reviewer actions


packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -82,14 +96,24 @@ export const getFilteredExplore = (
...
-                                    (tableReference) =>
-                                        filteredTableNames.includes(
-                                            tableReference,
-                                        ),
-                                ),
+                            ([metricName, metric]) => {
+                                const parentDimensionIsHidden =
+                                    hiddenDimensions.some(
+                                        (dimension) =>
+                                            dimension.sql === metric.sql,
```
rephus, line 103: afaik this is the most reliable  way to know if a metric comes from a dimension, right ? 
Screenshot from 2024-03-05 09-42-25

examples/full-jaffle-shop-demo/dbt/models/customers.sql
```diff
@@ -59,6 +59,7 @@ final as (
         customers.customer_id,
         customers.first_name,
         customers.last_name,
+        30 as age, -- fixed age is filtered using required_attributes on schema.yml
```
rephus, line 62: added for testing, but we can keep it there, I've added a new "fixed" field to not break anything else, let me know if you want me to remove it or  replace it

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
ZeRego: Yes, but won't work for metrics with custom SQL. To support those, the metrics should have a `dimensionReferences: Array<{table: string; name: string;}>;` similar to the `tablesTeferences`.

rephus: Yeah, I've been trying to figure out a way to link dimensions in metrics, but at least on the exploreCompiler, metrics and dimension are not linked, we might need to look before that happens, perhaps at `convertExplores` on `translator` ... I'll keep looking :eyes:

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -82,14 +96,25 @@ export const getFilteredExplore = (
...
-                                    (tableReference) =>
-                                        filteredTableNames.includes(
-                                            tableReference,
-                                        ),
-                                ),
+                            ([metricName, metric]) => {
+                                const parentDimensionIsHidden =
+                                    hiddenDimensions.some(
+                                        (dimension) =>
+                                            dimension.name ===
```
rephus, line 103: not it checks against the metric.dimensionReference

rephus: Screenshot from 2024-03-06 10-34-30

ZeRego: This is still going to cause us problems as I can reference the dimension A in metrics defined under dimension B.

I propose update`exploreCompiled.ts` (pseudo code):

compileDimensionReference() - now returns the referenced dimension
```
        return {
            ...
            dimension: referencedDimension,
        };
```

compileMetricSql() - aggregates all the dimension referenced and then returns 
```
let dimensionsReferences =  new Set<{id: string, table: string, name: string}>();

...

                if(compiledReference.dimension) {
                    dimensionsReferences = ([
                        ...dimensionsReferences,
                        { id: getItemId(compiledReference.dimension), table: compiledReference.dimension.table, name: compiledReference.dimension.name}
                    ]);
                }
                
...
 
                 return { sql: compiledSql, tablesReferences, dimensionsReferences };
```

Using `{id: string, table: string, name: string}` so we have all the information possible to relate fields/tables.

ZeRego (approved):

Issue #9219, opened 2024-03-04
I should not be able to tag someone that doesn't have access to the dashboard 


joaoviana:
### Description

Disable the selection of users that don't have access to the dashboard

**Acceptance Criteria**
- [ ] display a useful message as to why they can't select them

PR #9250, opened 2024-03-05
fix: only tag users with dashboard access

joaoviana:
Closes: #9219 

### Description:

Display tooltip and make button disabled when the user I'm trying to tag  doesn't have access to the dashboard's space.

There's also a performance issue that is reported in a TODO in `userNames`

https://github.com/lightdash/lightdash/assets/7611706/38c3bf38-4fd3-4e18-aadd-d623471c9b5f

### Reviewer actions


rephus (approved):
it looks good. I like the way `slack` does this, you can mention anyone, and if they are not in the channel, you can invite them (or send a private message). We could give access to the space automatically if thoes users are selected, perhaps for version 2

joaoviana:
> it looks good. I like the way `slack` does this, you can mention anyone, and if they are not in the channel, you can invite them (or send a private message). We could give access to the space automatically if thoes users are selected, perhaps for version 2

That's so cool! Adding it to the Notion doc, thanks @rephus, another feature unlocked 😆

Issue #5129, opened 2023-04-18
When I'm saving a chart or dashboard, there should be a loading spinner before the "saved" toast appears


TuringLovesDeathMetal:
### Description

Right now, when I try to save a chart or a dashboard, I get this weird limbo state where I don't really know what's going on. 

I clicked `save`, but I don't know if it's actually saving? Or if something's gone wrong? I end up clicking save a few times because I think I've not clicked hte button properly or something. 

https://www.loom.com/share/a1ba0f25cb174fd899d8832cb086f68d

Instead of having this "limbo" state where there is no apparent activity and the user is just waiting for the `success` toast, we should have a loading toast that turns into a `success` toast (i.e. the toast should say `saving chart` or `saving dashboard` with a spinner, then should become the green toast we have now once it's completed)

muhsinkamil:
Hi @TuringLovesDeathMetal Is this necessary ? When I tried locally, the process is super fast to me, 
Its maybe, because I don't have many charts / dashboards. 
( if it is necessary, I can implement this )

PR #9253, opened 2024-03-05
fix: add loading state to edit chart save button

la55u:
Closes: https://github.com/lightdash/lightdash/issues/5129

### Description:
The same button mentioned in the related issue for saving a dashboard already has a loading state during mutation. I figured the same technique could be applied to the chart save button too, which would IMO be an easy solution for the mentioned "limbo" state.

### Reviewer actions


IrakliJani (approved):

IrakliJani:
tested locally and works as expected. thanks @la55u for contributing 🙌

Issue #7996, opened 2023-11-16
Adjusting the format of my table calculation shouldn't reset my entire chart config


TuringLovesDeathMetal:
### Description

Rght now, if I'm using a table calculation as one of the fields in my chart, then I change the table calculation, all of the chart config I had set gets reset. This is really frustrating because I could have made a load of adjustments to the chart config before trying to adjust my table calculation. 

Ideally, when I adjust my table calculation, it would keep the chart config the same adn just update the values in my chart. 

https://www.loom.com/share/848d1bc09a534376b67a80500192fecf?from_recorder=1&focus_title=1

iMac7:
hey @TuringLovesDeathMetal please assign me this

gavin-ob:
This is closed but doesn't seem to fixed in v0.937.2 (latest version at time of writing), was this intentional? agree with OP that this is frustrating.

PR #9269, opened 2024-03-06
fix: update x and y field in chartConfig when tc updates

almeidabbm:
Relates to: #7996 

### Description:

- Updates `chartConfig` layout table calculation names on `UPDATE_TABLE_CALCULATION`
- changes `packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts` to update `dirtyLayout` and `dirtyEchartsConfig` only when external changed and is different from the dirty state in the hook - this makes sure we don't go into a cycle when trying to update the internal state
- changes `packages/frontend/src/components/LightdashVisualization/VisualizationConfigCartesian.tsx` to update external state only when config changed from the hook and the external one wasn't the source of that change, this makes sure we don't go into a state update cycle

Some notes on the second point:

1. When the Table Calculation updates, the name changes to a different (unique) one.
2. The previous implementation of `useCartesianChartConfig` received the new name in the `x` or `y` fields, but because the validity checks for those fields checked if the new name existed in the results we still didn't have the new table calculation there so it would be seen as invalid and therefore would fallback. 
The solution to this problem was to update the validity checks to also take into account `columnOrder` which didn't depend on the request results and had the new TC name added to it.
3. Even after the update of the validity checks to use `columnOrder` there was still some mismatches between the hook state and the `explore` context state which resulted again in the fallback, so the solution was to keep track of when config changed externally and update the internal state if it's different

https://github.com/lightdash/lightdash/assets/22939015/55dd9cfd-340a-4425-a354-124d423011cb

### Reviewer actions


packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -120,21 +122,40 @@ const useCartesianChartConfig = ({
...
+              }
+            : EMPTY_CARTESIAN_CHART_CONFIG;
+    }, [dirtyLayout, dirtyEchartsConfig]);
+
+    const hasChartConfigChangedExternally = useMemo(() => {
+        return (
+            !isEqual(initialChartConfig, validConfig) &&
+            !isEqual(initialChartConfig, prevInitialChartConfig)
+        );
+    }, [initialChartConfig, prevInitialChartConfig, validConfig]);
```
almeidabbm, line 150: The only reason this is duplicated in `packages/frontend/src/components/LightdashVisualization/VisualizationConfigCartesian.tsx` is so that we keep the call to update the state in the explore there and don't have to move it here

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -432,46 +458,50 @@ const useCartesianChartConfig = ({
...
     useEffect(() => {
-        if (availableFields.length > 0) {
+        if (availableFields.length > 0 || columnOrder.length > 0) {
             setDirtyLayout((prev) => {
                 const isCurrentXFieldValid: boolean =
                     prev?.xField === EMPTY_X_AXIS ||
-                    (!!prev?.xField && availableFields.includes(prev.xField));
+                    (!!prev?.xField &&
+                        (availableFields.includes(prev.xField) ||
+                            columnOrder.includes(prev.xField)));
```
joaoviana, line 498: could we add a comment here explaining why it could be done on `availableFields` or in `columnOrder`?

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -120,21 +122,40 @@ const useCartesianChartConfig = ({
...
+        return (
+            !isEqual(initialChartConfig, validConfig) &&
+            !isEqual(initialChartConfig, prevInitialChartConfig)
+        );
+    }, [initialChartConfig, prevInitialChartConfig, validConfig]);
+
+    useEffect(() => {
+        if (!hasChartConfigChangedExternally) return;
+
+        // Update dirty chart config ONLY when chart config changes externally and this hook has not changed it
```
joaoviana, line 155: what we mean by `externally` is  in the `exploreProvider`, right?

joaoviana (commented):
Testing in Render a bit more

joaoviana (approved):
Looks great! 

I've tested many scenarios: 
- chart has table calculation already 
renamed, ran query, removed, re-added, no errors
- chart doesn't have table calculations yet
added, saved, edited, saved, etc

For both scenarios, I've: 
- checked the configure panel, to confirm that the field changed
- was able to add it as a filter 
- tried to break table calculation and errors are as expected (checked against current behaviour in analytics)


2 things to cover in future PRs

**Proposal 1**
This should also work for other types of visualizations: 
- Pie chart
- Big Number

**Proposal 2**
1. Add table calculation to chart
2. Add a reference line
3. Reference table calculation
4. Change the name of the table calculation
5. Reference line select field goes blank (should it change to the new name? )


Anyway, only concern is with the deep equality check in `hasChartConfigChangedExternally` but I didn't see much of an issue with many charts in Render - it _feels_ like it works as well as before 👍 

Great work on debugging this 🙌

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
almeidabbm: yes, by externally I meant the chart config that comes from the props, will adjust the comment

almeidabbm:
> Looks great!
> 
> I've tested many scenarios:
> 
> * chart has table calculation already
>   renamed, ran query, removed, re-added, no errors
> * chart doesn't have table calculations yet
>   added, saved, edited, saved, etc
> 
> For both scenarios, I've:
> 
> * checked the configure panel, to confirm that the field changed
> * was able to add it as a filter
> * tried to break table calculation and errors are as expected (checked against current behaviour in analytics)
> 
> 2 things to cover in future PRs
> 
> **Proposal 1** This should also work for other types of visualizations:
> 
> * Pie chart
> * Big Number
> 
> **Proposal 2**
> 
> 1. Add table calculation to chart
> 2. Add a reference line
> 3. Reference table calculation
> 4. Change the name of the table calculation
> 5. Reference line select field goes blank (should it change to the new name? )
> 
> Anyway, only concern is with the deep equality check in `hasChartConfigChangedExternally` but I didn't see much of an issue with many charts in Render - it _feels_ like it works as well as before 👍
> 
> Great work on debugging this 🙌

100% I'll create another PR to apply similar fix for the other chart types!

packages/frontend/src/providers/ExplorerProvider.tsx
```diff
@@ -441,6 +441,33 @@ const calcColumnOrder = (
...
 };
 
+const updateChartConfigWithTableCalc = (
+    prevChartConfig: ChartConfig,
+    oldTableCalculationName: string,
+    newTableCalculationName: string,
+) => {
+    const newConfig = cloneDeep(prevChartConfig);
+
+    if (newConfig.type !== ChartType.CARTESIAN || !newConfig.config) {
```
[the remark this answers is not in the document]
almeidabbm: Yep yep! Nice catch, will create a separate PR for that asap :D

PR #9270, opened 2024-03-06
fix: allow interactive viewers setting filter operator to isNull and not null

magnew:
Closes: #9029 

### Description:

Allow viewers to set filters that were 'any value' to null or not null. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
@@ -6,6 +6,7 @@ import {
     Field,
     fieldId,
     FilterableField,
+    FilterOperator,
     isField,
     isFilterableField,
     matchFieldByType,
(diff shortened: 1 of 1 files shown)


rephus:
can you fix render ?

packages/frontend/src/components/DashboardFilter/FilterConfiguration/index.tsx
```diff
@@ -138,14 +139,22 @@ const FilterConfiguration: FC<Props> = ({
...
-                return oldFilterRule &&
-                    !oldFilterRule?.values?.length &&
-                    oldFilterRule?.disabled &&
-                    newFilterRule.values?.length
-                    ? { ...newFilterRule, disabled: false }
-                    : newFilterRule;
+                // value but now has a value, enable it. Also enable it if
+                // The operator requires no value (null/not null)
+                // This is a way of keeping disabled and 'no value' in sync.
+                let isNewFilterDisabled = newFilterRule.disabled;
```
rephus, line 145: this logic doesn't have permissions into account, so this can also be done by viewers, right ?

magnew: Good find. I think there is an existing bug where viewers can change this. They probably shouldn't be able to. This fix doesn't change the permissions though. 

I'm inclined to check this in as is and fix that separately if that's cool with you.

rephus:
Mmmm, I'm not sure how this works, I'm an admin and I can't set it to `null`

[Screencast from 07-03-24 15:46:58.webm](https://github.com/lightdash/lightdash/assets/1983672/74d46943-98db-460f-bcbb-8a5ae36de2dc)

rephus (approved):
If I set is_competed is null I get a few errors, I don't think it is related with this change, just that we don't have any payments with `is_completed is null `
image

Issue #9248, opened 2024-03-05
Support some form of service overrides within ServiceRepository


filp:
A couple of options are:

- Support sub-classing `ServiceRepository` and providing this sub class to `App`
- Support specifying per-service overrides as part of the `ServiceRepository` confirguration

PR #9282, opened 2024-03-07
chore: Add service override support, OperationContext, service memoization

filp:
Closes: #9248

### Description:

Expands `ServiceRepository` to support:

- Memoizing services
- Accepting service providers to override service instantiation

Nothing actually changes as far as how things work with this change, it just allows us to complete the last remaining steps of the milestone.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ServiceRepository.ts
+++ b/packages/backend/src/services/ServiceRepository.ts
@@ -67,6 +67,36 @@ type ServiceFactoryMethod<T extends ServiceManifest> = {
 };
 
 /**
+ * Structure for describing service providers:
+ *
+ *   <serviceName> -> providerMethod
+ */
+type ServiceProviderMap<T extends ServiceManifest> = Partial<{
+    [K in keyof T]: (providerArgs: {
+        repository: ServiceRepository;
+        context: OperationContext;
+    }) => T[keyof T];
+}>;
+
+/**
+ * Placeholder ServiceRepository context.
+ *
+ * The OperationContext object is created alongside the ServiceRepository, and shares
+ * its lifetime with a unit of work/single operation - for example, a single
+ * request thread.
+ *
+ * OperationContext can be subclassed to provide additional functionality for specific
+ * types of operations, if necessary.
+ */
+export class OperationContext {
+    constructor(
+        protected readonly properties: {
+            operationId: string;
+        },
+    ) {}
+}
+
+/**
  * Intermediate abstract class used to enforce service factory methods via the `ServiceFactoryMethod`
  * type. We need this extra thin layer to ensure we are statically aware of all members.
  */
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/ServiceRepository.ts
```diff
@@ -94,111 +159,222 @@ abstract class ServiceRepositoryBase {
...
- */ export class ServiceRepository
+ */
+export class ServiceRepository
     extends ServiceRepositoryBase
     implements ServiceFactoryMethod<ServiceManifest>
 {
+    /**
+     * Holds memoized instances of services after their initial instantiation:
+     */
+    protected _serviceCache: Partial<ServiceManifest> = {};
```
filp, line 170: If this looks like it does the same thing as `_services`: it does! One of them will be removed once we're no longer sharing services outside of the repository.

ZeRego: That means once we remove `services` the methods would be like this?

```
    public getAnalyticsService(): AnalyticsService {
        return this.getService(
            'analyticsService',
            () => new AnalyticsService(...),
        );
    }
```

packages/backend/src/services/ServiceRepository.ts
```diff
@@ -94,111 +159,222 @@ abstract class ServiceRepositoryBase {
...
+    }
+
+    /**
+     * Handles initializing a service, and taking into account service
+     * providers + memoization.
+     *
+     * If a factory is not provided, and a service provider is not defined,
+     * this method throws an error. This should not happen in normal operation.
+     */
+    public getService<
```
ZeRego, line 355: Minor: should this be private or is there a use case for this method to be used outside?

ZeRego: I would even consider moving this and the `_serviceCache` to `ServiceRepositoryBase` 🤔 
This way this class is only for the service methods.

ZeRego (approved):

packages/backend/src/services/ServiceRepository.ts
filp: Good catch, shouldn't be public!

packages/backend/src/services/ServiceRepository.ts
filp: Correct!

packages/backend/src/services/ServiceRepository.ts
filp: Re: moving it to `ServiceRepositoryBase`, I agree on principle - but the method needs to be aware of the actual service definitions, since we want to pass `ServiceRepository` down to the providers once it's "finalized" (so that providers can also depend on services from the repository).

Leaving the method as part of `ServiceRepository` is the most straight-forward way to do this.

PR #9286, opened 2024-03-07
chore: update eslint configuration (unused imports, vars, type modifier)

filp:
**NOTE:** This PR is part of a bundle of PRs; due to the amount of automated changes, I'm splitting them up and merging them into this single PR after testing/review.

### Description:

Updates default `eslint` rules across all packages:

- Unused vars are disallowed, unless as part of a rest assignment or args prefixed with `_`
- Unused imports raise a warning. The goal is to later convert this to an error, once it's verified to be safe
- Type-only imports must be prefixed with the `type` modifier, with preference for the inline modifier: (`import { type X } from 'foo'`)

I intentionally kept each project's configuration separate so that we can adjust these shared rules per-project as needed, for now.

### Reviewer actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
         "eslint-plugin-prettier": "^4.2.1",
         "eslint-plugin-react-hooks": "^4.6.0",
         "eslint-plugin-testing-library": "^5.6.3",
+        "eslint-plugin-unused-imports": "^3.1.0",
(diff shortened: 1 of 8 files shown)


filp:
Closing this, will move each package's `eslint` changes into the invidual fix PRs.

PR #9287, opened 2024-03-07
chore: fix backend lint errors

filp:
See #9286 for context.

### Description:

Fixes eslint errors for `packages/backend`. Mostly via `lint --fix`, with some manual fixes in the mix.

### Reviewer actions


packages/backend/src/clients/EmailClient/EmailClient.ts
```diff
@@ -165,7 +165,8 @@ export default class EmailClient {
                 roleAction = 'manage';
                 break;
             default:
-                const nope: never = projectMember.role;
+                // NOOP - Intentionally left blank
```
IrakliJani, line 168: Hey @filp, This call checks whether the switch/case is "exhaustive" or not. Was this change esaential?

Fyi, We can use the `assertUnreachable` utility function from common.

packages/backend/src/clients/EmailClient/EmailClient.test.ts
```diff
@@ -41,6 +43,8 @@ describe('EmailClient', () => {
             );
         });
         test('should create transported with secure connection when using port 465', async () => {
+            // Relies on side-effects from instancing EmailClient
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
             const client = new EmailClient({
```
IrakliJani, line 48: maybe we can use `client_` or `_client` don't remember which one

packages/backend/src/services/OrganizationService/OrganizationService.ts
```diff
@@ -110,7 +110,7 @@ export class OrganizationService {
     }
 
     async updateOrg(
-        { organizationUuid, organizationName, userUuid, ability }: SessionUser,
```
IrakliJani, line 113: unused?

packages/backend/src/projectAdapters/projectAdapter.ts
```diff
@@ -118,7 +118,6 @@ export const projectAdapterFromConfig = async (
                 useDbtLs,
             });
         default:
-            const never: never = config;
```
IrakliJani, line 121: same here

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -186,7 +178,6 @@ export class DashboardModel {
                     });
                     break;
                 default: {
-                    const never: never = tile;
```
IrakliJani, line 189: one more

packages/backend/src/logging/winston.ts
```diff
@@ -1,5 +1,5 @@
-import { SessionUser } from '@lightdash/common';
-import * as express from 'express';
+import { type SessionUser } from '@lightdash/common';
+import type * as express from 'express';
```
IrakliJani, line 2: how does this work?

packages/backend/src/clients/EmailClient/EmailClient.ts
filp: Was this supposed to raise an `eslint` error? I agree with using `assertUnreachable`, but didn't know it was available under `common` and didn't want to introduce new exceptions as part of this changeset.

Agreed overall though, I'll make the adjustments!

packages/backend/src/clients/EmailClient/EmailClient.test.ts
filp: Good point, will fix!

packages/backend/src/services/OrganizationService/OrganizationService.ts
filp: Correct, though in retrospect I'll re-introduce this and tag it as unused instead, to keep with my own rule of not modifying implementation too much.

packages/backend/src/clients/EmailClient/EmailClient.ts
IrakliJani: good point! if we don't want to raise then maybe we could do `_nope` or `nope_`?
but then again, looks like a code block that should never ever happen and its better to throw in that case

filp:
Closing until I have more bandwidth to tackle this + include other proposed changes.

Issue #3790, opened 2022-11-15
Replace type 'any' with 'unknown'


ZeRego:
The use of the type `any` has prevented us from catching bugs while developing. 

We should:
- stop using the type `any` and use `unknown` where can't have better typing
- enable tsconfig rule `useUnknownInCatchVariables: true`
- enable "noImplicitAny": true,
- enable "@typescript-eslint/no-explicit-any": "error"

This work should be broken down into multiple PRs so it's easier to review.

Ankush263:
hello @ZeRego, can I work on this Issue?
my initial approach will be first to change all the front-end code and make a PR
then go for the backend

ZeRego:
@Ankush263 yes 👍

Inder782:
Hi @ZeRego I tried doing to do it , can you provide some feedback so that I can proceed further 

#9161

ZeRego:
Hi @Inder782 👋 
Your PR is going in the right direction :) 
FYI: I updated this ticket to have a few more ts/eslint rules.

owlas:
Came up today in a post-mortem and making it a priority!

IrakliJani:
Before starting work on this, please see the comment here: https://lightdash.slack.com/archives/C02GQKJK84Q/p1737368131551269?thread_ts=1737130494.998139&cid=C02GQKJK84Q

PR #9292, opened 2024-03-07
refactor: any => unknown in CLI (Part II.)

la55u:
Related: https://github.com/lightdash/lightdash/issues/3790

### Description:
Part II of converting `any` to `unknown` in the CLI. With this PR the CLI package is now free from all `any` except one in `handlers/dbt/run.ts` which I couldn't find a proper type for. Also, the `@typescript-eslint/no-explicit-any` rule is configured to throw an error from now on.

### Reviewer actions

### DIFF
--- a/packages/cli/src/handlers/dbt/run.ts
+++ b/packages/cli/src/handlers/dbt/run.ts
@@ -12,6 +12,7 @@ type DbtRunHandlerOptions = DbtCompileOptions & {
 
 export const dbtRunHandler = async (
     options: DbtRunHandlerOptions,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     command: any,
 ) => {
     GlobalState.setVerbose(options.verbose);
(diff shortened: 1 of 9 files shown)


packages/cli/src/handlers/dbt/run.ts
```diff
@@ -12,6 +12,7 @@ type DbtRunHandlerOptions = DbtCompileOptions & {
 
 export const dbtRunHandler = async (
     options: DbtRunHandlerOptions,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     command: any,
```
ZeRego, line 16: We should use the type `Command` from  `import {Command} from "commander";`

ZeRego (changes requested):

packages/cli/src/handlers/dbt/run.ts
ZeRego: fixed it

ZeRego (approved):

Issue #8942, opened 2024-02-15
datepicker calendar value displayed in different tz instead of local value


Shifan-Gu:
### Description

relates to: https://github.com/lightdash/lightdash/pull/8705

Hi team, the applied date picker timezone fix is not fully working, when i open the date picker in a non-UTC zone, date picker placeholder value is not consistent with the calendar value.

![Screenshot 2024-02-15 at 03 52 42](https://github.com/lightdash/lightdash/assets/101873365/c70cc1cb-8e18-4ab8-aed8-d1a256e3964e)

### App Version
0.979.3

PR #9303, opened 2024-03-08
fix: ensure date time picker inputs are all in sync

joaoviana:
Closes: #8942

### Description:

This ensures that `value` depends on `valueFormat` 
Also changed the `moment` usage with `dayjs` 

When `valueFormat` is processed, then `value` is assigned to its value but parsed 

Tested when locally I'm in San Francisco - this is related to the issue reported, where the value of the input wasn't matching the day selected on the date calendar input + the time input below
Screenshot 2024-03-08 at 15 52 29

Then in London
Screenshot 2024-03-08 at 15 52 55

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -216,8 +216,14 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
                             : moment();
 
                     const dateValue = valueIsDate
-                        ? formatDate(value, undefined, true)
+                        ? formatDate(
+                              // Treat the date as UTC, then remove its timezone information before formatting
+                              moment.utc(value).format('YYYY-MM-DD'),
+                              undefined,
+                              false,
+                          )
                         : formatDate(defaultDate, undefined, false);
+
                     filterRuleDefaults.values = [dateValue];
                 }
                 break;
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -108,12 +108,7 @@ export function formatDate(
     timeInterval: TimeFrames = TimeFrames.DAY,
     convertToUTC: boolean = false,
 ): string {
-    const momentDate = moment(date);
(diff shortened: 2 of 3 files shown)


joaoviana:
To test, you can try adding a filter for `timestamp ntz` - raw
image

rephus (approved):
It works great, thanks !

packages/common/src/utils/filters.ts
```diff
@@ -216,8 +216,14 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
                             : moment();
 
                     const dateValue = valueIsDate
-                        ? formatDate(value, undefined, true)
+                        ? formatDate(
+                              // Treat the date as UTC, then remove its timezone information before formatting
+                              moment.utc(value).format('YYYY-MM-DD'),
+                              undefined,
+                              false,
```
joaoviana, line 223: brought back the `false` because of the bug this value was fixing in: https://github.com/lightdash/lightdash/pull/3763

joaoviana:
Tested 3 scenarios here for timezones: San Francisco, and Madrid (as well as London)

1. Filter dashboard to: have a chart (with base dimension `event ltz` day) - the first bar should be `2020-07-01` and when you click on it, you should see `filter dashboard to 2020-07-01` on any timezone
2. Add a timestamp filter to a chart - the default value of the input should also match the ones on the datepicker: calendar should have today's day; time below the calendar should match the one on the input. Essentially, everything is consistent
3. Add a date day filter to a chart - when setting a `day` filter, change its operator. See comment in: https://github.com/lightdash/lightdash/pull/9303#discussion_r1519931060 - this basically means that by not forcing a `utc` conversion + then formatting we can ensure the timezone info is removed before using the date. What we do instead is: by using this method, we get a date string that's neutral regarding timezones and DST, ensuring consistency across different locales without directly mentioning or handling specific timezones

rephus (approved):

PR #9304, opened 2024-03-08
chore: instantiate services within `ServiceRepository`

filp:
Closes: #9246 

### Description:

Migrates services instantiated under `services.ts` to `ServiceRepository` instantiation methods, and updates code referencing those service instance singletons directly.

Note: this is still an ongoing change, and for now all services continue to be singletons due to the shared `serviceRepository` instance (which simply acts as a container).

### Reviewer actions


packages/backend/src/controllers/authentication.ts
```diff
@@ -297,7 +307,7 @@ export const apiKeyPassportStrategy = new HeaderAPIKeyStrategy(
     true,
     async (token, done) => {
         try {
-            const user = await userService.loginWithPersonalAccessToken(token);
+            const user = userService.loginWithPersonalAccessToken(token);
```
ZeRego, line 310: why remove the `await` here?

ZeRego (approved):

packages/backend/src/controllers/authentication.ts
filp: Good catch, unintentional.

PR #9305, opened 2024-03-08
chore: share `ServiceRepository` instance from `App`/`SchedulerApp`

filp:
Closes: #9247 
Based on #9304, will rebase after merge.

### Description:

Moves further along the process of introducing `ServiceRepository` + removing direct references to service singleton instances.

- Updates `App` to instantiate `ServiceRepository`
- Updates `SchedulerApp` to instantiate `ServiceRepository`
- Uses the App's own instance across `express`, `passport`, and `tsoa`
  - `passport` and `tsoa` access the shared instance via the express request (`req.services`) or via factory functions
- Deprecates the `serviceRepository` singleton, which is currently used only from within a knex seed file.

Keyword to force frontend e2e tests:
test-frontend

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -44,8 +44,10 @@ import { registerNodeMetrics } from './nodeMetrics';
 import { postHogClient } from './postHog';
 import { apiV1Router } from './routers/apiV1Router';
 import { SchedulerWorker } from './scheduler/SchedulerWorker';
-import type { ServiceRepository } from './services/ServiceRepository';
-import { serviceRepository } from './services/services';
+import {
+    OperationContext,
+    ServiceRepository,
+} from './services/ServiceRepository';
 import { wrapOtelSpan } from './utils';
 import { VERSION } from './version';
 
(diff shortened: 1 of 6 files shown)


packages/backend/src/App.ts
```diff
@@ -102,6 +106,13 @@ export default class App {
                     this.lightdashConfig.rudder.dataPlaneUrl,
             },
         });
+        this.serviceRepository = new ServiceRepository({
```
filp, line 109: Eventually this will be done per-request, but this approach allows us to encapsulate App configuration already (via the `OperationContext` object + service overrides in `ServiceRepository`), while keeping the semantics consistent downstream for later.

packages/backend/src/App.ts
```diff
@@ -185,7 +196,7 @@ export default class App {
          * request context - for now we simply proxy the existing service repository singleton.
          */
         expressApp.use((req, res, next) => {
-            req.services = serviceRepository;
+            req.services = this.serviceRepository;
```
filp, line 199: There's probably a good case to instead/also pass the App object itself along the request thread.

packages/backend/src/App.ts
```diff
@@ -400,13 +417,14 @@ export default class App {
             analytics: this.analytics,
             // TODO: Do not use serviceRepository singleton:
```
filp, line 418: Left the comment here, since we continue to use what's functionally a singleton here, and ultimately we don't want to.

ZeRego (approved):

Issue #9200, opened 2024-03-01
Error deploying project with Databricks warehouse


ZeRego:
### Description

Deploying project fails with error. Project uses databrick warehouse.

### To Reproduce

slack thread: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1709230149346069?thread_ts=1709224173.353359&cid=C03MG2VTR08

```
2024-02-29 17:55:51 [Lightdash] debug: Failed task 5271 (testAndCompileProject) with error The background threadpool cannot accept new task for execution, please retry the operation (166337.93ms):
  WarehouseQueryError: The background threadpool cannot accept new task for execution, please retry the operation
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Perhaps we are making too many queries to the warehouse when fetching the table's schema.

### App Version

0.1021.9

PR #9306, opened 2024-03-08
fix: databricks catalog queries to run in batches

ZeRego:
Closes: #9200

### Description:

Changes:

- await query to close
- batch catalog queries in groups of 100
- throw original error instead of `close` error

To replicate the original error you need to override the requests to be 500 or more and change the DATABRICKS_QUERIES_BATCH_SIZE to be 500 ( basically, it creates a single batch of the promises ). 

```
requests = Array.from({ length: 500 }, () => ({
    database: 'hive_metastore',
    schema: 'default',
    table: 'payments',
}));
```

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
@@ -121,6 +121,27 @@ const normaliseDatabricksType = (type: string): DatabricksTypes => {
     return match[0] as DatabricksTypes;
 };
 
+const DATABRICKS_QUERIES_BATCH_SIZE = 100;
+
+async function processPromisesInBatches<T, R>(
+    items: Array<T>,
+    batchSize: number,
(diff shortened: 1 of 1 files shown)


packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
```diff
@@ -257,27 +278,34 @@ export class DatabricksWarehouseClient extends WarehouseBaseClient<CreateDatabri
...
+                },
+            );
+        } catch (e: any) {
+            throw new WarehouseQueryError(e.message);
         } finally {
-            await close();
+            try {
+                await close();
+            } catch (e: any) {
+                // Only console error. Don't allow close errors to override the original error
```
magnew, line 306: Nice

magnew (commented):
Reproed, tested. Looks good. Nice work! 

Do we think 100 is just a pretty good guess?

magnew (approved):

Issue #9278, opened 2024-03-07
Filter value stays there, even though an option has been selected


TuringLovesDeathMetal:
if you choose an option in a filter after you’ve started writing on it the option will be chosen but the word will stay there and you have to manually delete it

![image](https://github.com/lightdash/lightdash/assets/31848148/570bec87-5e69-4215-821b-34e012b8d1c3)

IrakliJani:
Note: This is a controversial UX decision. Clearing the string means it also "resets" results. The best UX practice could be to clear the string but keep the previous results. This might be significant work if Mantine makes it hard to implement.

PR #9309, opened 2024-03-09
fix: clear search on option select

la55u:
Closes: https://github.com/lightdash/lightdash/issues/9278

### Description:
This is an easy fix as it's a built in Mantine prop. Also, I added a gap between options because it looked ugly if two or more options were selected under each other.

Before:
image

After: 
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.tsx
@@ -171,6 +171,7 @@ const FilterStringAutoComplete: FC<Props> = ({
                 </Group>
             )}
             styles={{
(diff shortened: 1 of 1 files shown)


piyush-at-sangti (approved):

piyushchauhan (approved):

IrakliJani (approved):
Hi @la55u, I just tested this locally and decided to only merge `clearSearchOnChange` as introducing a gap between the items would cause disparity among other select components. Feel free to open up a ticket so we can consider changing that globally. Thanks again!

IrakliJani:
Note: This is a controversial UX decision. Clearing the string means it also "resets" results. The best UX practice could be to clear the string but keep the previous results. This might be significant work if Mantine makes it hard to implement.

Let's wait for the feedback from the customers

PR #9324, opened 2024-03-11
feat: add ServiceRepository.getEmbedService stand-in

filp:
### Description:

Adds a stand-in for `EmbedService` in `ServiceRepository`, which must be configured explicitly during repository initialization.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ServiceRepository.ts
+++ b/packages/backend/src/services/ServiceRepository.ts
@@ -581,9 +581,7 @@ export class ServiceRepository
     }
 
     public getEmbedService<EmbedServiceImplT>(): EmbedServiceImplT {
-        return this.getService('embedService', () => {
-            throw new Error('EmbedService is not implemented');
-        });
+        return this.getService('embedService');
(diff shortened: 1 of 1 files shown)


ZeRego (approved):

packages/backend/src/services/ServiceRepository.ts
```diff
@@ -577,6 +580,12 @@ export class ServiceRepository
         );
     }
 
+    public getEmbedService<EmbedServiceImplT>(): EmbedServiceImplT {
+        return this.getService('embedService', () => {
+            throw new Error('EmbedService is not implemented');
```
ZeRego, line 585: Should we throw the error here or omit the function and let the `getService()` throw `Unable to initialize service '${serviceName}' - no factory or provider.` ?

filp: Fair point, I'll let it fall to `getService` to avoid potential confusion later down the road.

PR #9338, opened 2024-03-12
fix: user space access query

ZeRego:
Closes: 

### Description:

We were fetching all the org users when checking the user space access. The more users an org has the slower the query was. Since we checked the space access in most endpoints, it was slowing down the entire app.

Changes:
- filter query by user uuid
- move user table join to be first to optimize query filtering + fix sequencial scans in following joins
- removed logic that removes user access on get
- remove space access call in scheduler

The only places where we still fetch all the org members access are for:
- get space endpoint
- deprecated endpoint `/spaces-and-content` 

Force frontend e2e test:
test-frontend

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -537,7 +537,7 @@ export class SpaceModel {
 
     private async _getSpaceAccess(
         spaceUuid: string,
-        filters: { userUuid?: string } | undefined,
+        filters?: { userUuid?: string },
     ): Promise<SpaceShare[]> {
         const access = await this.database
             .table(SpaceTableName)
@@ -711,12 +711,8 @@ export class SpaceModel {
                     ...inheritedGroupRoles,
                 ]);
 
-                // exclude all users that were converted to organization members and have no space access
+                // exclude users with no space role
                 if (!highestRole) {
-                    // revoke direct access from the space
-                    if (user_with_direct_access) {
-                        this.removeSpaceAccess(spaceUuid, user_uuid);
-                    }
                     return acc;
                 }
                 return [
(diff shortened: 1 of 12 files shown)


packages/backend/src/models/SpaceModel.ts
```diff
@@ -705,7 +713,10 @@ export class SpaceModel {
 
                 // exclude all users that were converted to organization members and have no space access
                 if (!highestRole) {
-                    this.removeSpaceAccess(spaceUuid, user_uuid); // remove access from the space if it exists
+                    // revoke direct access from the space
+                    if (user_with_direct_access) {
+                        this.removeSpaceAccess(spaceUuid, user_uuid);
```
joaoviana, line 718: Related to this right? `amend logic to only trigger delete query when needed`

joaoviana:
And there's unit test errors on some files

in `DashboardService.test.ts` L56
```
spaceModel: {
        getFullSpace: jest.fn(async () => publicSpace),
        getSpaceSummary: jest.fn(async () => publicSpace),
        getFirstAccessibleSpace: jest.fn(async () => space),
        getSpaceAccess: jest.fn(async () => []),
    },
```

should change to 
```
spaceModel: {
        getFullSpace: jest.fn(async () => publicSpace),
        getSpaceSummary: jest.fn(async () => publicSpace),
        getFirstAccessibleSpace: jest.fn(async () => space),
        getUserSpaceAccess: jest.fn(async () => []),
    },
```

packages/backend/src/models/SpaceModel.ts
```diff
@@ -604,6 +607,11 @@ export class SpaceModel {
             )
             .where(`${EmailTableName}.is_primary`, true)
             .where(`${SpaceTableName}.space_uuid`, spaceUuid)
+            .modify((query) => {
```
owlas, line 610: this is cool!

joaoviana: +1!!

packages/backend/src/models/SpaceModel.ts
ZeRego: This is trying to keep the db clean when users no longer have direct access to a space.
Example:
 - user A is an org editor
 - user A is given access to a private space in project X
 - user A is changed to org member and has no access to the project X

ZeRego: Removed logic.

joaoviana (approved):

Issue #9315, opened 2024-03-11
Split login/register/sign-up form


owlas:
In split login, the user provides an email which is used to customise the login experience

### frontend

This must be tested rigorously with password managers to make sure it's still auto filled correctly. Is it accessible if we dynamically change the form? Could the elements simply be hidden?

1. Email capture, ask for email and check for available login methods. (do we add a button or make it debounced check?)
2. Potentially forced redirect the user, including the email as a login hint (should be available for most oidc integrations) 
3. Show login/register/invite options for the user (possibly password or google or sso button (see logic below)

### backend

API required

```
// 
// GET /login_options?email=hello@email.com
//
200
{
	"show_options": ["email", "sso", "google"],
	"force_redirect": true,
  "redirect_uri": "/auth/sso?connection_uuid=uuid&email=hello%40mail.com"
}
```

Force redirect happens in the case the user has an existing SSO identity. It should redirect to the correct provider. It should also use `login_hint` to prefill the provider with the email address.

For example the redirect for google would be
```
//
// GET /auth/google?email=hello%40mail.com
//
302
https://googleauth...
```

Full precheck logic:

1. Existing SSO sign in -> redirect to sso
2. Existing Google sign in -> redirect to google
3. Existing password -> show password field + sign in with google (we might later do account linking) 
4. No existing account -> show password field + sign in with google 
5. Passwords disabled and no existing SSO/Google, then show SSO/Google options. 

### infra

rate limited required for unauthenticated endpoint

PR #9340, opened 2024-03-12
feat: split login

rephus:
Closes: #9315

### Description:

With existing google open id  javier@lightdash.com

Screenshot from 2024-03-12 16-43-27

with existing (fake) okta open id 

Screenshot from 2024-03-12 16-43-39

without open id 
Screenshot from 2024-03-12 16-43-47

### Reviewer actions


rephus:
Something weird is happening when I have a sso account and I type my email (even before clicking `continue` ) I don't see an options request

[Screencast from 15-03-24 08:43:00.webm](https://github.com/lightdash/lightdash/assets/1983672/a25bb362-b98e-47a2-b646-5b2c92f9316d)

rephus:
One more thing, I disabled my password locally my screen flashed, but no error showed up, I should get an error. perhaps errors are not handled correctly on the login query.

I'm testing this here https://github.com/lightdash/lightdash/pull/9351 

[Screencast from 15-03-24 09:09:05.webm](https://github.com/lightdash/lightdash/assets/1983672/106c1c1d-7f3f-4250-8867-4d79835dedad)

rephus:
Ok, this is weird, not sure we can do about it . Because of the feature flag, first we get the first form,a nd then we get the other. I know this is a temporary issue, but may be we can just not show anything until the flag is loaded... 

[Screencast from 15-03-24 09:24:09.webm](https://github.com/lightdash/lightdash/assets/1983672/69c376ba-3831-4a61-b23b-4307cc66c819)

magnew:
> Something weird is happening when I have a sso account and I type my email , redirection doesn't work on the first `continue`, then typing my email reloads the page.
> 
>  Screencast.from.15-03-24.08.46.11.webm

I'll see what I missed here

magnew:
> r I think this could be a redir

I'll look at this too.

magnew:
> Ok, this is weird, not sure we can do about it . Because of the feature flag, first we get the first form,a nd then we get the other (sometimes it takes a few seconds !) . I know this is a temporary issue, but may be we can just not show anything until the flag is loaded...
> 
>  Screencast.from.15-03-24.09.24.09.webm

I saw this and figured maybe it wasn't worth the effort to fix since its a temporary flagging issue. But let me see if there is something obvious to do.

packages/frontend/src/features/users/components/LoginLanding.tsx
```diff
@@ -0,0 +1,261 @@
...
+        loginOptionsSuccess && loginOptions?.showOptions ? 'login' : 'precheck';
+
+    const isEmailLoginAvailable =
+        loginOptions?.showOptions &&
+        loginOptions?.showOptions.includes(LocalIssuerTypes.EMAIL);
+
+    const disableControls =
+        (loginOptionsLoading && loginOptionsFetched) || isLoading || isSuccess;
+
+    const googleAuthAvailable = health.data?.auth.google.enabled;
```
magnew, line 129: Is there a better way to do this?

rephus: no, that's the best way

magnew (approved):
Tested

Issue #9297, opened 2024-03-08
Chart colors may not be consistently assigned from chart editor to dashboard


filp:
### Description

User reports indicate chart colors may not be predictable going from the chart editor to the dashboard in some situations.

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C05F0N3UPSP/p1710235792414099?thread_ts=1710235792.414099&cid=C05F0N3UPSP

owlas:
This issue was mentioned by a user in slack: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1709820382461129?thread_ts=1709820382.461129&cid=C03MG2VTR08

PR #9342, opened 2024-03-12
feat: consistent color mapping via sequential assignment

filp:
Closes: #9297 

### Description:

Changes consistent color assignment to provide higher quality mapping, via grouped+sequential assignment.

In short: colors are assigned to series respecting the order of the color palette, instead of through a hash of the identifier string.

image
image

### Implementation details:

- Null, empty strings, etc are always assigned a gray color.
- Colors are no longer assigned based on a hash of the group or identifier
- Colors are assigned following the order defined in the color palette
- If the same color has already been assigned to a group or identifier, we reuse that assignment across different charts in a dashboard
- Assignments are tracked through sub-groups, based on their underlying table/model or other grouping mechanism. This allows multiple unrelated charts to cycle through available colors in parallel.
- Assignments are tracked via the numeric index of the color in the palette, instead of the resolved color hexcode.

Example structure of the `colorMappings` map after rendering a dashboard:

image

## How is this better?

- Colors _feel_ more predictable, and more closely match the color palette - not just in absolute colors used, but also in how they are displayed in order.
- Greatly reduces the chance of color collisions for neighboring series.

## How is this worse?

- Color assignments are more susceptible to load times: a dashboard containing multiple charts across the same dimensions, but with different filters, will still have colors consistently assigned, but those colors may become less neatly ordered.
  - E.g if the first chart to load for `customer_first_name` has one filter, and the second chart has a different filter on the same dimension, the values of the first loaded chart will have first pick of the color assignments, and could influence the order of colors for all subsequent charts on the same dimension.

Example of two charts on the same dashboard, with the pie chart of the left side having a filter on `First Name`. Colors are still applied consistently, but the pie chart has a segment that does not follow the expected color order, since 'Aaron' was first "seen" before the pie chart was rendered.

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -75,7 +75,7 @@ type VisualizationContext = {
     setPivotDimensions: (value: string[] | undefined) => void;
 
     getSeriesColor: (seriesLike: SeriesLike, seriesIndex: number) => string;
-    getGroupColor: (groupNameOrId: string) => string;
+    getGroupColor: (groupPrefix: string, groupName: string) => string;
 
     colorPalette: string[];
 };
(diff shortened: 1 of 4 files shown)


magnew (approved):
LGTM. I played with this a bunch and the behavior seems good. And the code makes sense. Nice work. It's hard to know exactly what people will want, but we're already there.

PR #9347, opened 2024-03-13
feat: Rate limiter

rephus:
Closes: 

### Description:

This only applies rate limiting to the new login_options endpoint

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/rateLimiter.ts
+++ b/packages/backend/src/controllers/rateLimiter.ts
@@ -3,7 +3,7 @@ import { rateLimit } from 'express-rate-limit';
 
 export const limiter = rateLimit({
     windowMs: 60 * 1000, // 1 minute
-    limit: 100, // Limit each IP to 100 requests per `window` (here, per minute).
+    limit: 100, // Limit each IP to 100 requests per `window` (per minute).
     standardHeaders: 'draft-7', // draft-6: `RateLimit-*` headers; draft-7: combined `RateLimit` header
     legacyHeaders: false, // Disable the `X-RateLimit-*` headers.
 });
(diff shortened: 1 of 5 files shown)


packages/backend/src/services/UserService.ts
```diff
@@ -1180,9 +1180,16 @@ export class UserService {
             return {
                 showOptions: [openIdIssuer],
                 forceRedirect: true,
-                redirectUri: getRedirectUri(openIdIssuer),
+                redirectUri: new URL(
+                    `/api/v1${getRedirectUri(openIdIssuer)}`,
```
rephus, line 1184: full url, not related with rate limit. I pushed here by mistake, we should merge it anyway

packages/backend/src/controllers/rateLimiter.ts
```diff
@@ -0,0 +1,12 @@
+import { RequestHandler } from 'express';
+import { rateLimit } from 'express-rate-limit';
+
+export const limiter = rateLimit({
+    windowMs: 60 * 1000, // 1 minute
+    limit: 100, // Limit each IP to 100 requests per `window` (per minute).
```
rephus, line 6: is this a good limit ? it seems too much for this single endpoint, but at the same time, we should be able to reuse this limit for other endpoints

owlas:
@rephus we should handle this here: https://github.com/lightdash/lightdash-cloud/blob/f9cd7f3bd8def49332958c4b8d89aa31daf8cf87/platform-automation/src/main.ts#L94-L94

magnew (approved):
Nice. I know this is waiting on the other PR.

Issue #9317, opened 2024-03-11
UserService login to organization


owlas:
### Description

On login the organization logic is inside the authentication handlers. Instead create an explicit user service method to login to an org.

1. The user is first authenticated without an org
2. Then we check organization membership

```
loginToOrganization(userUuid, loginMethod)
```

1. Check for existing memberships
2. Throw error if user is a member of multiple
3. Handle case where user has no organization to login
4. Throw error if user has access to one organization but `loginMethod` is not allowed (currently usused)
5. Returns organizationUuid for the organization if allowed

### Session management

The `organizationUuid` should be stored on the session so that when we use `deserializeUser` we always build the user with the `userUuid` and `organizationUuid` that were determined on first session creation. 

For example: if the user's membership is revoked, then this will fail because there's no entry joining the user and organization. However if the organization / instance changed it's login policy, the session would still persist.

PR #9364, opened 2024-03-14
feat: login to organization

rephus:
Closes: #9317

### Breaking-ish change: 

User serssion serialization has changed, instead of the `user id` , now we have a JSON with `id + organization`

image

The desialization method supports both, for backwards compatibility

### Reviewer actions


packages/backend/src/services/UserService.ts
```diff
@@ -682,11 +682,20 @@ export class UserService {
...
             const user = await this.userModel.getUserByPrimaryEmailAndPassword(
                 email,
                 password,
             );
-            this.identifyUser(user);
+            const userOrganization = this.loginToOrganization(
+                user.userUuid,
+                'password',
+            );
+            const userWithOrganization = {
```
rephus, line 694: I believe we want to do this, add the organization to the user from the `loginToOrganization` instead of from the model.
If so, we should delete the code that already fetches the org for the user `userDetailsQueryBuilder`

packages/backend/src/services/UserService.ts
```diff
@@ -1041,6 +1050,28 @@ export class UserService {
...
+    > {
+        const organizations = await this.userModel.getOrganizationsForUser(
+            userUuid,
+        );
+        if (organizations.length === 0) {
+            throw new NotExistsError('User not part of any organization');
+        } else if (organizations.length > 1) {
+            throw new ForbiddenError('User is part of multiple organizations');
+        }
+        // TODO check valid login methods allowed in org
```
rephus, line 1070: afaik we don't have this

owlas: Yep this is fine as a placeholder. 

There are valid cases where it's ok not to have an organisation. For example on first registering. So this error needs to be easy to catch or the function should return undefined

packages/backend/src/services/UserService.ts
owlas: yes! good point! try removing this and see what the impact is

owlas:
how does this affect when we deserialise a user from the session for each request? we should store the organisation id on the session to be explicit about which org the user is logged in to

rephus:
> how does this affect when we deserialise a user from the session for each request? we should store the organisation id on the session to be explicit about which org the user is logged in to


I can add the organization uuid to the serialized session and support just the user for backwards compatibility 
image

packages/backend/src/models/UserModel.ts
```diff
@@ -570,6 +614,40 @@ export class UserModel {
         };
     }
 
+    async findSessionUserAndOrgByUuid(
+        userUuid: string,
+        organizationUuid: string,
+    ): Promise<SessionUser> {
+        const [user] = await userDetailsQueryBuilder(this.database)
+            .where('user_uuid', userUuid)
+            .andWhere('organizations.organization_uuid', organizationUuid) // We filter organizationUuid here
```
rephus, line 623: same as the now deprecated `findSessionUserByUUID` , but with the org filter

magnew (approved):
LGTM. Tested with:

I tested login with:

PW login and google enabled, log in with google
PW login and google enabled, log in with PW
PW login enabled, google disabled, log in with PW
PW login disabled, google enabled, log in with google

owlas:
Does it actually need to be backwards compatible? I would just drop the sessions table to logout any active sessions when the migration runs, that way the table will be empty and you can ensure that the `organizationUuid` is always present on the session.

packages/backend/src/models/UserModel.ts
```diff
@@ -88,6 +92,7 @@ const userDetailsQueryBuilder = (
         .joinRaw(
             'LEFT JOIN emails ON users.user_id = emails.user_id AND emails.is_primary',
         )
+        // TODO remove this org join, we should do this in the service
```
owlas, line 95: I think the join is fine as long as we're filtering on a specific organization

packages/backend/src/models/UserModel.ts
```diff
@@ -541,6 +579,12 @@ export class UserModel {
         return this.getUserDetailsByUuid(user.user_uuid);
     }
 
+    /**
+     * Returns the user with the default organization
+     * Used in old deserialization method, for backwards compatibility
+     * You should use findSessionUserAndOrgByUuid instead
+     * @deprecated
```
owlas, line 586: We could remove this instead of deprecating it by doing a `delete from sessions` migration

rephus:
> Does it actually need to be backwards compatible? I would just drop the sessions table to logout any active sessions when the migration runs, that way the table will be empty and you can ensure that the `organizationUuid` is always present on the session.

that would make the code simpler... ok. I'll do that.

packages/backend/src/models/UserModel.ts
rephus: mm, this is also used in other methods like csv downloads or the new github oauth authentication to get the organizationUuid... 

image

we should do this on a separate PR, to remove all dependencies to `user.organization`

rephus: I made the migration and updated the deserializer method, I kept the deprectated method and updated its description

Issue #8364, opened 2023-12-20
The sidebar in the explore view is not full height when I try to create a chart from within a dashboard


PriPatel:
<img width="206" alt="Screenshot 2023-12-20 at 11 01 29" src="https://github.com/lightdash/lightdash/assets/12776522/8957b6d2-d6eb-4497-97b3-6ec24aef2b1e">

There's a gap at the bottom

amanx123:
Can i solve this issue ?

PR #9396, opened 2024-03-15
fix: navbar layout issues

joaoviana:
Closes: #7531 #8364

### Description:

When users were editing charts from dashboards, they would see an issue with the Page layout where the Explorer sidebar and the whole `Page` seemed cropped. 
The reason why this is happening currently if because: 
- the main nav bar behaves one way, has specific height, styling, etc
- the blue banner that we render when `editing charts from dashboards` is a different component, with less of the styling we apply to the main nav bar

Instead of replacing the `Navbar` completely when that's the case, we can just render different content inside the same component.

What was done: 
- Created separate components for: `main nav bar content`, `dashboard explorer banner`, and the `preview banner`
- On the `index.tsx` of `NavBar`, we have now a Nav bar chooser hook, that controls the content & styles of the Header. 
- Consequently, we now don't need the `hasBanner` prop for the `Page` component, but we **still** let the Page component apply styles when the project is PREVIEW.

Here is a screen recording:
https://github.com/lightdash/lightdash/assets/7611706/8c041466-ea8a-4e3a-8ca7-2cec775f9812

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
+++ b/packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
@@ -1,0 +1,72 @@
+import { Button, Text, Tooltip } from '@mantine/core';
+import { IconInfoCircle } from '@tabler/icons-react';
+import { useEffect, useMemo, type FC } from 'react';
+import { useHistory, useParams } from 'react-router-dom';
+import useDashboardStorage from '../../hooks/dashboard/useDashboardStorage';
+import MantineIcon from '../common/MantineIcon';
+
+type Props = {
+    projectUuid: string;
+};
+
+export const DashboardExplorerBanner: FC<Props> = ({ projectUuid }) => {
+    const history = useHistory();
+    const { savedQueryUuid, mode } = useParams<{
+        savedQueryUuid: string;
+        mode?: string;
+    }>();
+
+    const { getEditingDashboardInfo, clearDashboardStorage } =
+        useDashboardStorage();
+    const { name: dashboardName, dashboardUuid } = getEditingDashboardInfo();
+
+    useEffect(() => {
+        window.addEventListener('unload', clearDashboardStorage);
+        return () =>
+            window.removeEventListener('unload', clearDashboardStorage);
+    }, [clearDashboardStorage]);
+
+    const action = useMemo(() => {
+        if (!savedQueryUuid) {
+            return 'creating';
+        }
+        switch (mode) {
+            case 'edit':
+                return 'editing';
+            case 'view':
+            default:
+                return 'viewing';
+        }
+    }, [savedQueryUuid, mode]);
+
+    return (
+        <>
+            <MantineIcon icon={IconInfoCircle} color="white" size="sm" />
+            <Text color="white" fw={500} fz="xs" mx="xxs">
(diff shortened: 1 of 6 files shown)


joaoviana:
https://github.com/lightdash/lightdash/assets/7611706/8ae01e82-6b21-490a-8feb-b41ceaa28203

IrakliJani:
I am not sure what is wrong with the render environment. I re-launched the headless browser, but the tests won't start. You can close the PR and reopen it.

IrakliJani:
looks great in all of the edge cases that I know of...

even this one 😱 

CleanShot 2024-03-19 at 00 30 51@2x

joaoviana:
E2E tests passed @IrakliJani , thanks for mentioning this issue :)

packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    const { getEditingDashboardInfo, clearDashboardStorage } =
+        useDashboardStorage();
+    const { name: dashboardName, dashboardUuid } = getEditingDashboardInfo();
+
+    useEffect(() => {
+        window.addEventListener('unload', clearDashboardStorage);
+        return () =>
+            window.removeEventListener('unload', clearDashboardStorage);
+    }, [clearDashboardStorage]);
```
IrakliJani, line 27: ok im little bit lost here

packages/frontend/src/components/NavBar/index.tsx
```diff
@@ -1,213 +1,111 @@
...
+    );
     const headerContainerHeight =
         NAVBAR_HEIGHT + (isCurrentProjectPreview ? BANNER_HEIGHT : 0);
 
+    const renderNavBarContent = () => {
+        switch (navBarMode) {
+            case NavBarMode.EDITING_DASHBOARD_CHART:
+                return <DashboardExplorerBanner projectUuid={projectUuid} />;
+            case NavBarMode.DEFAULT:
+            default:
```
IrakliJani, line 84: navbarMode can never be undefined so you can move the content from `default` block to NavbarMode.DEFAULT and call `assertUnreachable` inside `default` block

IrakliJani (approved):
this is great. one question. otherwise 👍

packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
joaoviana: Same 😆 
This was here before, so it clears the dashboardStorage when the banner unloads. This usage is not recommended, but will address on a separate PR

packages/frontend/src/components/NavBar/index.tsx
joaoviana: Yes 🤦 
Thanks! Will add this 👍

packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
IrakliJani: oh cool 😆

packages/frontend/src/components/NavBar/index.tsx
IrakliJani: perfect 👍

PR #9422, opened 2024-03-18
refactor: slack integration service

ZeRego:
Closes: 

### Description:

Following the work from https://github.com/lightdash/lightdash/pull/9420 this PR is meant to remove the `models/models.ts` import from the slack router.

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -27,7 +27,7 @@ import { ClientManifest } from './clients/clients';
 import DbtCloudGraphqlClient from './clients/dbtCloud/DbtCloudGraphqlClient';
 import EmailClient from './clients/EmailClient/EmailClient';
 import { GoogleDriveClient } from './clients/Google/GoogleDriveClient';
-import { SlackService } from './clients/Slack/Slackbot';
+import { SlackBot } from './clients/Slack/Slackbot';
 import { SlackClient } from './clients/Slack/SlackClient';
 import { LightdashConfig } from './config/parseConfig';
 import {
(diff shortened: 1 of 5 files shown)


packages/backend/src/clients/Slack/Slackbot.ts
```diff
@@ -37,14 +37,14 @@ const notifySlackError = async (
...
-type SlackServiceArguments = {
+type SlackBotArguments = {
     slackAuthenticationModel: SlackAuthenticationModel;
     lightdashConfig: LightdashConfig;
     analytics: LightdashAnalytics;
     unfurlService: UnfurlService;
 };
 
-export class SlackService {
+export class SlackBot {
```
filp, line 47: Good job tidying this potential source of confusion 🧹

packages/backend/src/App.ts
```diff
@@ -387,7 +387,7 @@ export default class App {
     }
 
     private initSlack() {
-        const slackService = new SlackService({
+        const slackBot = new SlackBot({
```
filp, line 390: Not a blocker, but would it make sense to move this to a method on `SlackIntegrationService`, or another similar service? e.g `getIntegrationService().createSlackBot(...)`

filp (approved):

packages/backend/src/App.ts
ZeRego: The slack bot starts an "ExpressReceiver" in port 4352 so it should be initialized in the App class. But I agree that this is not obvious and can be improved.

PR #9432, opened 2024-03-18
refactor: model repository

ZeRego:
Closes: 

### Description:

Add model repository where all the model classes are generated and can be overridden. 

Changes:
- add ModelRepository class ( identical to ServiceRepository without `context` )
- update ServiceRepository to receive a ModelRepository and get the models acordingly
- update App and SchedulerApp to create a ModelRepository

### Reviewer actions


packages/backend/src/models/ModelRepository.ts
```diff
@@ -0,0 +1,465 @@
...
+        );
+    }
+
+    public getGithubAppInstallationsModel(): GithubAppInstallationsModel {
+        return this.getModel(
+            'githubAppInstallationsModel',
+            () =>
+                new GithubAppInstallationsModel({
+                    database: this.database,
+                    encryptionService: new EncryptionService({
```
filp, line 204: Ideally should use the service repository to access services in this phase. We might get a circular dependency though.

filp (approved):

packages/backend/src/models/ModelRepository.ts
ZeRego: `EncryptionService` needs to be renamed `EncryptionUtils` as it is not a real service

Issue #9427, opened 2024-03-18
CLI profiles.yml should also accept `schema` instead of `dataset` for bigquery


almeidabbm:
### Description

From the DBT [docs](https://docs.getdbt.com/docs/core/connect-data-platform/bigquery-setup#oauth-via-gcloud), a `bigquery` profile should be allowed to have `schema` defined instead of `dataset`.

Currently we are enforcing the `dataset` field.

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C047XUL7QAZ/p1710763679226799?thread_ts=1710763679.226799&cid=C047XUL7QAZ

PR #9437, opened 2024-03-18
fix: allow schema in profiles.yml for bigquery

almeidabbm:
Closes: #9427

### Description:

- Allows `bigquery` profiles to have `schema` instead of `dataset` for CLI by passing it as `dataset` in `convertBigquerySchema`

### Reviewer actions

### DIFF
--- a/packages/cli/src/dbt/targets/Bigquery/index.ts
+++ b/packages/cli/src/dbt/targets/Bigquery/index.ts
@@ -15,6 +15,7 @@ import {
 type BigqueryTarget = {
     project: string;
     dataset: string;
+    schema: string;
     priority?: 'interactive' | 'batch';
     retries?: number;
     location?: string;
@@ -24,11 +25,15 @@ type BigqueryTarget = {
 
 export const bigqueryTargetJsonSchema: JSONSchemaType<BigqueryTarget> = {
     type: 'object',
+    required: ['project'],
     properties: {
         project: {
             type: 'string',
         },
         dataset: {
+            type: 'string',
+        },
+        schema: {
             type: 'string',
         },
         priority: {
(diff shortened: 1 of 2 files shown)


packages/cli/src/dbt/targets/Bigquery/index.ts
```diff
@@ -24,13 +25,17 @@ type BigqueryTarget = {
 
 export const bigqueryTargetJsonSchema: JSONSchemaType<BigqueryTarget> = {
     type: 'object',
+    required: ['project'],
```
IrakliJani, line 28: is this essential for the PR?

packages/cli/src/dbt/targets/Bigquery/index.ts
```diff
@@ -93,7 +105,13 @@ export const convertBigquerySchema = async (
         };
     }
     const lineErrorMessages = (validate.errors || [])
-        .map((err) => `Field at ${err.instancePath} ${err.message}`)
+        .map((err) => {
+            if (err.keyword === 'oneOf') {
+                return 'Profile should contain either `dataset` or `schema`';
+            }
+
+            return `Field at ${err.instancePath} ${err.message}`;
```
IrakliJani, line 113: Sometimes a validation error occurs without an `instancePath`. In such cases, should we simply log the error message and completely ignore the `instancePath`? I believe a similar issue arose with a protobuf breaking change, making the error confusing to interpret.

packages/cli/src/dbt/targets/Bigquery/index.ts
almeidabbm: This already existed but I removed `dataset` from there, since it will now either `schema` or `dataset` are required in `oneOf`. I must've moved this up in the schema somehow, pushed a commit to put it back in place 👌

packages/cli/src/dbt/targets/Bigquery/index.ts
almeidabbm: I've pushed a commit to use `better-ajv-errors` that other targets where using instead of doing our own parsing.

Note: In that commit I added also to snowflake target since that one and bigquery were the only targets not using `better-ajv-errors`

IrakliJani: sweet. thank you!

IrakliJani (approved):
looks good. pre-approving assuming tests pass 👌

PR #9451, opened 2024-03-19
chore: fix race condition in dashboard test

ZeRego:
Closes: 

### Description:

These tests sometimes fail because we save all the dashboard tiles in parallel so we shouldn't rely on the order of the tiles array. 

Updated the test to find the correct tile instead.

### Reviewer actions

### DIFF
--- a/packages/e2e/cypress/e2e/api/dashboard.cy.ts
+++ b/packages/e2e/cypress/e2e/api/dashboard.cy.ts
@@ -21,6 +21,7 @@ const chartMock: CreateSavedChart = {
     name: 'chart in dashboard',
     tableName: 'orders',
     metricQuery: {
+        exploreName: 'orders',
         dimensions: ['orders_customer_id'],
         metrics: [],
         filters: {},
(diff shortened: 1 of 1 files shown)


packages/e2e/cypress/e2e/api/dashboard.cy.ts
```diff
@@ -163,24 +164,22 @@ describe('Lightdash dashboard', () => {
...
+                        ({ properties }: DashboardChartTile) =>
+                            properties.savedChartUuid === newChart.uuid &&
+                            properties.belongsToDashboard,
+                    );
+                    const secondTile = updatedDashboard2.tiles.find(
+                        ({ properties }: DashboardChartTile) =>
+                            properties.savedChartUuid === newChart2.uuid &&
+                            properties.belongsToDashboard,
+                    );
+                    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
ZeRego, line 177: Tried to fix linting as instructed [here](https://github.com/cypress-io/eslint-plugin-cypress?tab=readme-ov-file#chai-and-no-unused-expressions) without success. Looks like one of the plugins are not maintain anymore.

packages/e2e/cypress/e2e/api/dashboard.cy.ts
```diff
@@ -163,24 +164,22 @@ describe('Lightdash dashboard', () => {
...
-                    // assert second tile is correct
-                    expect(
-                        secondTile.properties.savedChartUuid,
-                        'Check if second tile is correct',
-                    ).to.eq(newChart2.uuid);
-                    expect(
-                        secondTile.properties.belongsToDashboard,
-                        'Check if second tile belongs to a dashboard',
-                    ).to.eq(true);
+                    const firstTile = updatedDashboard2.tiles.find(
```
rephus, line 167: so the orders of the tiles are not consistent ?

rephus (approved):

packages/e2e/cypress/e2e/api/dashboard.cy.ts
ZeRego: correct, we insert the tiles in parallel into the database so one can be faster than the other. The order in the array is not relevant for the UI as the library figures the position based on the x/y h/w properties.

Issue #9349, opened 2024-03-13
Users can create "quick calculations" from metrics


TuringLovesDeathMetal:
### Description

Right now, if I don't know SQL as a user, then it's really hard for me to use table calculations. We have table calculation SQL templates in our docs, but actually, these should just be in the app directly. 

To make writing these calculations completely "automatic", we just **need to know the field that the user wants to reference in their calculation.**

There are two ways we could do this:

1. Adding some `quick calculation` thing with a `field picker` to the table calculation modal 
2. Adding `quick calculations` as an action _**from**_ a metric

WE're going to do (2) for now - but whoever picks this up, if you feel like tackling (1)/have any good ideas, then please add them!

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVNjTC2kE=/
Notion doc: https://www.notion.so/lightdash/67d3fd912f764d9eaf0f8e741d76fa00?v=805962b15e6a49958b88d1ec7dce15c4&p=cb50d8856e614e1e9a4ca31f0dd56d34&pm=s

- [ ] When I click on the `metrics` submenu, there is a section called `quick calculations`
- [ ] Some quick calculations are available for some metric types and not others (this information is included below...)
- [ ] I've tested these SQL queries with: Postgres, BigQuery, Snowflake. But we still need to test this with Redshift and Databricks! (also, what do we do if someone is using Trino?...this feels like pretty standard SQL across all flavours, so I'm willing to risk that it "just works" for people using Trino).

#### Percent change from previous

- [ ] This calculates the % change from a value vs. the previous row. 
- [ ] The created table calculation has % formatting applied with `decimal points` = 2
- [ ] Metrics that are number types are the only metrics that have this `quick calculation` option shown

Example where there is ordering in the table:

```sql
SELECT
(
  ${metric} / LAG(${metric}) OVER(
    ORDER BY 
      ${order_column_1}, ${order_column_2}, etc -- this should match the ordering of any dimensions/metrics in the table
  )
) - 1
```

Example where there is no ordering in the table:

```sql
SELECT
(
  ${metric} / LAG(${metric}) OVER()
) - 1
```

#### Percent of previous value

- [ ] This calculates the % of a value compared to the value in the row above it.
- [ ] The created table calculation has % formatting applied with `decimal points` = 2
- [ ] Metrics that are number types are the only metrics that have this `quick calculation` option shown

Example where there is ordering in the table:

```sql
SELECT
[...]

PR #9452, opened 2024-03-19
feat: quick calculations

rephus:
Closes:   #9349

### Description:

Screenshot from 2024-03-19 16-41-50

### Reviewer actions


TuringLovesDeathMetal (changes requested):
woweeee! so exciting!

This looks really great :) I think the only thing that isn't working as expected for me is the ordering in the table calculation. 

I think the ordering in the `ORDER BY` bit of the table calculation should match whatever's in my results table ordering. 

e.g. here:

image


I'd expect the `order by` to be `payments_method, order_date_month DESC`

This is what it actually is: 

image

So, two things:

1. The `desc` is applied when I've ordered a column descending (e.g. date month here)
2. There is only ordering applied to the columns that have ordering in the table (e.g. here, there is no ordering applied to `total revenue`, so it doesn't get ordering in the table calculation).

TuringLovesDeathMetal:
sorry, one more thing: have we built analytics events for this? (I think that a frontend event might be the best here 👀, but not sure if it's possible to make it a backend event ).

aaaand docs - not sure if you just wanted a review on functionality first! Then were going to do them separately :) but let me know if you need help!

rephus:
Hey @TuringLovesDeathMetal thanks for the review

I fixed the sorting, one thing to notice, if you change the sorting , the table calculation will not get automatically updated ... 
image

> have we built analytics events for this
There is already a `create_quick_table_calculation_button.click`  there is no backend for this

> aaaand docs
Also done https://github.com/lightdash/lightdash/pull/9452/files#diff-aefcc74f448e37d7af568b590923f62fd8f8ca9e441b33a16cdbcc42aab8c1ffR40

docs/docs/guides/table-calculations/adding-table-calculations.mdx
```diff
@@ -32,11 +33,26 @@ Watch out: table calculations can be easier to create than regular metrics/dimen
...
 Table calculations can only be built using the dimensions + metrics that you've included in your results table. So, to create a table calculation, first, you need to add the dimensions and/or metrics for your table calculation to your results table.
 
-#### 2. To add a table calculation just click on `+ Table calculation` in the `Results` tab
+### Using Quick Calculations
+
+Quick calculations are shortcuts to the most common table calculations, these are only available to metrics in the contextual menu in the results table.
+
+<img
+  src={QuickTableCalculation}
+  width="880"
```
TuringLovesDeathMetal, line 46:
```suggestion
  width="352"
```

docs/docs/guides/table-calculations/adding-table-calculations.mdx
```diff
@@ -32,11 +33,26 @@ Watch out: table calculations can be easier to create than regular metrics/dimen
...
+  height="557"
```
TuringLovesDeathMetal, line 47:
```suggestion
  height="223"
```

docs/docs/guides/table-calculations/adding-table-calculations.mdx
```diff
@@ -32,11 +33,26 @@ Watch out: table calculations can be easier to create than regular metrics/dimen
...
+  style={{ display: 'block', margin: '0 auto 20px auto' }}
+/>
+
```
TuringLovesDeathMetal, line 50:
```suggestion
To learn more about what these calculations are doing, [check out the docs here](https://docs.lightdash.com/guides/table-calculations/sql-templates).
```

TuringLovesDeathMetal (changes requested):
Niiiice! 

Ya, I think the image is definitely a bit big, I've added some suggested size for it (normally, I just use 40% of the image size and it's pretty good?). 

I've also linked out to the table calculation templates which explain each calculation. 

And yep! That makes sense for the sorting - playing with it definitely makes me feel like this ist he first step to sql-less table calculations 👀

TuringLovesDeathMetal:
@rephus sorry, my markdown was obviously wrong, I don't think the link rendered properly

image

rephus:
weird, I copied the same url we have in the doc and still it doesn't work 

image

rephus:
ok, this worked. 
image

TuringLovesDeathMetal (approved):
amazing! looks great 🚀

Issue #9298, opened 2024-03-08
When I set a label for my field in my dbt .yml, it doesn't apply when I add the field on the x-axis


TuringLovesDeathMetal:
### Description

Here's what my .yml looks like:

```yaml
      - name: organization_created_ts
        meta:
          dimension:
            label: organization created
            type: timestamp
```

So, in Lightdash, it appears like this: 
<img width="376" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/2db6a6c3-1b34-45f6-9299-3c7506690273">

And in the results table, it appears like this:

<img width="522" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/2d3bbbbd-b52d-46bc-a161-b8f4d394b6ef">

But then, in my chart, the label appears as the raw field name (not the label name):
<img width="825" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/1b004e87-ad98-4b4d-9826-13c1f4c4f993">

Instead of saying `organization created ts`, it should say `organization created week`

joaoviana:
We've introduced a change here: https://github.com/lightdash/lightdash/issues/8094, to simplify the date x-axis to include the `friendlyName` of the dimension + ignoring the date granularity. This was helpful for the `date zoom` functionality (where at one point we couldn't update the date granularity of the x axis label I believe), but now we can. 

With this change, we will bring back the DATE LABEL + GRANULARITY. 

I've just tested date zooming a chart with the new changes and it changed the granularity correctly. 
<img width="866" alt="Screenshot 2024-03-18 at 12 29 06" src="https://github.com/lightdash/lightdash/assets/7611706/e53225dc-49e7-4058-8626-58a74334460e">
<img width="429" alt="Screenshot 2024-03-18 at 12 28 58" src="https://github.com/lightdash/lightdash/assets/7611706/99b0679b-3dad-4d98-8a36-999e2a97a1da">

sounds good @TuringLovesDeathMetal ?

joaoviana:
proposal 2 it is

Issue #8094, opened 2023-11-27
Improved default axis labels for date fields


owlas:
### Description

When we generate a label for a date dimension it'll say something like "Order date day" or "Order date week" - it should just be "Order date". This is cleaner and makes date-zooming clearer

This should only be applied to chart axis labels (not to the results table).

rephus:
Perhaps this is cleaner in the chart, but what about results table / table viz ? YOu might want to add different columns with different time frames, and you don't want to show `Order date` on all of them @owlas

rephus:
If we do this, do we even need to do https://github.com/lightdash/lightdash/issues/8105 ?

PR #9453, opened 2024-03-19
fix: display label and if so remove time interval

joaoviana:
Closes: #9298 

### Description:

After this: https://github.com/lightdash/lightdash/issues/8094, we've removed the `date granularity` from the date group labels on the x axis, but we don't respect the `label` set on the yml.

This PR uses proposal 2, which originally was: 
`Use date label only (e.g. Order created) - and if no label specified, then just generate the friendly name` in https://github.com/lightdash/lightdash/issues/9298#issuecomment-2003813784. 

What this PR does is: 
- if it's in a `group`, then check its `timeInterval`, and display the `label` without the `time interval`
- if the `time interval` - for any reason - is invalid, then use the friendly name

I'm open to suggestions on doing this differently... 

This is an example `axisItem`:

```json
{
    "sql": "DATE_TRUNC('DAY', ${TABLE}.timestamp_tz)",
    "name": "timestamp_tz_day",
    "type": "date",
    "group": "timestamp_tz",
    "index": 0,
    "label": "Timestamp (with timezone) day",
    "table": "events",
    "hidden": false,
    "fieldType": "dimension",
    "tableLabel": "Events",
    "compiledSql": "DATE_TRUNC('DAY', \"events\".timestamp_tz)",
    "description": "",
    "timeInterval": "DAY",
    "tablesReferences": [
        "events"
    ]
}
```

### Reviewer actions

test-cli test-frontend test-backend

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -113,6 +113,7 @@ import {
 import { type UserWarehouseCredentials } from './types/userWarehouseCredentials';
 import { type ValidationResponse } from './types/validation';
 
+import { TimeFrames } from './types/timeFrames';
 import { convertAdditionalMetric } from './utils/additionalMetrics';
 import { getFields } from './utils/fields';
 import { formatItemValue } from './utils/formatting';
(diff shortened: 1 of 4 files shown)


packages/common/src/index.ts
```diff
@@ -770,9 +771,23 @@ export const getDateGroupLabel = (axisItem: ItemsMap[string]) => {
...
+        axisItem.label &&
+        axisItem.timeInterval
+    ) {
+        const timeFrame =
+            TimeFrames[axisItem.timeInterval]?.toLowerCase() || '';
+
+        if (timeFrame && axisItem.label.endsWith(` ${timeFrame}`)) {
+            // Remove the time frame from the end of the label - e.g. from 'Order created day' to 'Order created'.
+            return getItemLabelWithoutTableName(axisItem).replace(
+                new RegExp(`\\s+${timeFrame}$`),
```
rephus, line 784: why the escape character in `\\s`?

rephus (approved):
Reviewed and tested

packages/common/src/index.ts
joaoviana: so it interprets \s as a whitespace character match rather than an escape sequence 👍

joaoviana:
Just forcing E2Es for App and API to run @rephus

Issue #9368, opened 2024-03-14
Reference line of 1 sets the chart y axis minimum to 1. 


12ian34:
There's a strange bug when adding reference lines to charts. I've set a reference line value of 1 on the Y axis and then the minimum Y value then defaults to 1. However, when setting to anything above one it works as expected.

[See this](https://lightdash-community.slack.com/archives/C03MG2VTR08/p1710410143121219) for screenshots.

owlas:
This issue was mentioned by a user in slack: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1710410143121219?thread_ts=1710410143.121219&cid=C03MG2VTR08

Issue #9379, opened 2024-03-14
Auto y-axis range of line charts does not adjust properly


almeidabbm:
### Description

Auto y-axis range is clipping line chart lines

### To Reproduce

Haven't been able to reproduce

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C05F0N3UPSP/p1710345039666669?thread_ts=1710345039.666669&cid=C05F0N3UPSP

PR #9458, opened 2024-03-19
fix: reference lines breaking axis ranges

magnew:
Closes: #9368 #9379 

### Description:

There were a few bugs around reference lines when multiple series were involved. If you had a reference line on the **first** series in your chart, and that line was outside of the range of the data in that series, we were setting the entire chart's range to that reference line (either min or max). Any other series' values above or below the range wouldn't show. 

This fix changes `getMinAndMaxReferenceLines` to take an array of series to find the min and max values across all of the series, not just the first one. There's a note in the review about the core problem. The rest of the code here is just supporting the notion of checking the range of all of the series. 

I'll list the repro steps here, but the main thing to test is that charts aren't broken. Try multi series, multi axes, reference lines. 

To repro the original issue in jaffle shop:
- Open 'How many orders we have over time?'
- Add a reference line to 'unique order count' with a value of 6 (above the max for that series)
- Move 'unique order count' to be the top series
- The max range is now 6

A similar thing can happen with lines below the ranges. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -316,63 +316,61 @@ const maxDate = (a: number | string, b: string) => {
 };
 
 export const getMinAndMaxValues = (
-    axis: string | undefined,
+    series: string[] | undefined,
     rows: ResultRow[],
 ): (string | number)[] => {
-    if (!axis) return [];
+    if (!series || series.length === 0) return [];
 
-    return rows
-        .map((row) => row[axis]?.value.raw)
-        .reduce<(string | number)[]>(
-            (acc, value) => {
-                if (
-                    typeof value === 'string' &&
-                    dayjs(value, 'YYYY-MM-DD', false).isValid()
-                ) {
-                    // is date
-                    const min = minDate(acc[0], value);
-                    const max = maxDate(acc[1], value);
+    let rawValues = [];
+    for (const s of series) {
+        for (const row of rows) {
+            rawValues.push(row[s]?.value.raw);
+        }
+    }
+
+    return rawValues.reduce<(string | number)[]>(
+        (acc, value) => {
+            if (
+                typeof value === 'string' &&
+                dayjs(value, 'YYYY-MM-DD', false).isValid()
+            ) {
+                // is date
+                const min = minDate(acc[0], value);
+                const max = maxDate(acc[1], value);
+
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -988,16 +991,23 @@ const getEchartAxes = ({
         axisId: bottomAxisXId,
     });
 
-    const leftAxisYId = validCartesianConfig.layout.flipAxes
+    const leftAxisYFieldIds = validCartesianConfig.layout.flipAxes
         ? validCartesianConfig.layout?.xField
-        : validCartesianConfig.eChartsConfig.series?.find(
-              (serie) => serie.yAxisIndex === 0,
```
magnew, line 994: The main issue was here. We only used one series to calculate the range

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -316,63 +316,61 @@ const maxDate = (a: number | string, b: string) => {
...
-
-                        return [min, max];
-                    }
                 }
-                return acc;
-            },
-            [0, 0],
-        );
+            } else {
+                // TODO: this case comes up more than it should given that
```
rephus, line 360: should we add a log here instead of silently fail ?

rephus (approved):
Nice fix, I tested and everything looks fine, I found some issues with reference lines and `horizontal bar charts` (aka flipped) but I tested those on my branch and they keep happening, so they are not new, let's just create an issue 

For example: max axis doesn't work on grouped flipped chats

[Screencast from 21-03-24 08:46:36.webm](https://github.com/lightdash/lightdash/assets/1983672/39d13779-a8ab-4afb-843c-a8a206c28cda)

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
magnew: Would you log to the console? I wasn't entirely sure what it would say. You think it should log to the console? I put this here cause I figure it's helpful for devs, but I didn't throw or anything because there is nothing a user can do about it.

magnew: I'm going to merge without this. But I might look into this TODO. If you want, I'm happy to add a log here.

PR #9460, opened 2024-03-19
Starrocks client

hhhonzik:
### Description:

adds starrocks support to Lightdash, still a lot of WIP and most likely will never be merged (due to conditions of dbt-starrocks connector), but I just wanted to see if anything like this is possible. 

### Action points

- [x] connection via SQL
- [ ] cli support - catalog query needs to be rewritten


owlas:
Hey @hhhonzik this is awesome and looks like a great start. The team is really limited on availability for reviews, so at this time I'm going to close the PR as it's not a top priority for us currently. However, the branch will remain and we hopefully have time to revisit this.

Prioritisation is almost entirely determined by user interest, so if you have potential Lightdash users that are requesting this, then sharing that information would help us to get to this sooner!

Issue #9386, opened 2024-03-15
Preview projects should copy content the user has access to


almeidabbm:
### Description

Currently `admin`, `developer`, `editor` and `interactive viewer` role users can create preview projects.
For security restrictions only `admin` user created preview projects get the content copied from the base project.
The other user role's preview projects do not get any of the content copied over.

As a `developer`, `editor` and `interactive viewer` I would like my preview projects to have the same content I have access to in the source project.

### What problem does this solve?

Roles other than `admin` do not have access to any content in preview projects.

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C06B2PP2478/p1710277691872089?thread_ts=1710277691.872089&cid=C06B2PP2478

PR #9461, opened 2024-03-20
feat: clone preview without private spaces

rephus:
Closes: #9386

### Description:

Screenshot from 2024-03-20 11-07-46
Screenshot from 2024-03-20 11-08-32

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -410,7 +410,7 @@ export class ProjectService {
                 // We only allow copying from projects if the user is an admin until we remove the `createProjectAccess` call above
                 if (
                     user.ability.cannot(
-                        'manage',
+                        'create',
                         subject('Project', {
                             organizationUuid,
                             projectUuid: data.copiedFromProjectUuid,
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -410,7 +410,7 @@ export class ProjectService {
                 // We only allow copying from projects if the user is an admin until we remove the `createProjectAccess` call above
                 if (
                     user.ability.cannot(
-                        'manage',
+                        'create',
```
rephus, line 413: interactive_viewer and above

ZeRego (approved):
lgtm 👍 

We need to update docs as well
Screenshot 2024-03-20 at 15 25 09

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: I would have set this to `view` as this check is to check if the user has access to this project. The check for creating a project was before the action to create a project.

rephus: perhaps we can remove the ability  check here, you're right, the `create` was done already.

Issue #9419, opened 2024-03-18
Add value label to Reference Lines based on 'average'


12ian34:
From a user

> When you add a reference line, and use value = ‘average’ it creates an average line and labels it in the chart ‘average’ - as per my image above. It would be great if that label was the actual value -“average=7.8%”

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C049LQZGVD3/p1710756234195909?thread_ts=1710756234.195909&cid=C049LQZGVD3

hamzahc1:
@magnew - another one that could do with a headway update when it merges 😄

PR #9470, opened 2024-03-20
feat: show value of average reference lines

magnew:
Closes: #9419

### Description:

Displays the value on average reference lines. This ended up being a little more than I expected. We had been using the 'value' prop on reference lines to store the UUID, but to get echarts to do display this value correctly, we need to pass that field as undefined. So I added a uuid to reference lines and it seems to work. BUT it makes this a little riskier. 

The basic approach is:
- Someone types 'average' in the value box, we will show an average line (apparently this is already a feature)
- We have to set the `type` to average and the right formatter. And clear the axis value.
- I used the reference line name to store the label.

@rephus, if you see a reason I am missing that this is a bad idea, I can try it a different way. One answer would be to make specific UI controls for this, but the 'average' hack is already out there. I did some work to test backwards compat, but I might do more. 

Kapture 2024-03-20 at 18 20 56

Edit: I added a checkbox as a UI as well as a placement interface
Screenshot 2024-03-25 at 19 03 49

### Reviewer actions


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLine.tsx
```diff
@@ -202,11 +202,11 @@ export const ReferenceLine: FC<Props> = ({
     }, [referenceLine]);
 
     const [value, setValue] = useState<string | undefined>(
-        referenceLine.data[markLineKey],
+        referenceLine.data[markLineKey] || referenceLine.data.type,
     );
 
     const [label, setLabel] = useState<string | undefined>(
-        referenceLine.data.label?.formatter,
+        referenceLine.data.name,
```
magnew, line 209: Might need a back-compat check here

rephus: yeah, this is risky, last time I changed something like this  I broke the ability to have multiple reference lines...

rephus (commented):
Typing "average" ont he value seems like a bad UI , perhaps we can replace this with a `toggle` boolean that `disables` the `value` input , or you can ask for feedback on #design

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLines.tsx
```diff
@@ -70,21 +70,35 @@ export const ReferenceLines: FC<Props> = ({ items, projectUuid }) => {
...
                         lineStyle: { color: updateColor },
-                        label: updateLabel ? { formatter: updateLabel } : {},
-                        xAxis: undefined,
-                        yAxis: undefined,
-                        [dirtyLayout?.xField === fieldId ? 'xAxis' : 'yAxis']:
-                            updateValue,
+                        label: updateLabel
+                            ? { formatter: updateLabel }
+                            : useAverage
+                            ? { formatter: 'ave: {c}' }
```
rephus, line 83: avg instead of ave?

rephus:
Another UI thing, 
if you use average on big values, most of the number is hidden, perhaps we need to add some offset, rotate the number, or add some tooltip

image

packages/common/src/types/savedCharts.ts
```diff
@@ -134,8 +134,10 @@ export const isPivotReferenceWithValues = (
 export type MarkLineData = {
     yAxis?: string;
     xAxis?: string;
-    name: string;
-    value: string;
+    name?: string;
+    value?: string;
+    type?: string;
+    uuid: string;
```
rephus, line 140: iirc the reason for using `name / value` instead of uuid is because we had to rely on the properties of `markLine` to be able to use the chartconfig to store the values. This seem to work though, so I'm not sure if that is a requirement any more. 

https://echarts.apache.org/en/option.html#series-line.markLine

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLine.tsx
rephus: I tested locally and it seems to be backwards compatible :+1:

packages/common/src/types/savedCharts.ts
magnew: Yea, I wondered about this, but we don't seem to be using their types, so I figured I'd just try it. I don't see any errors...

magnew:
> Another UI thing, if you use average on big values, most of the number is hidden, perhaps we need to add some offset, rotate the number, or add some tooltip
> 
> image

Yea, this is a real problem. It's even worse if I use 'average' instead of 'ave'. It's one of those things that could take a long time to get right automatically. I wonder if providing a placement option for the label would be too much?

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLines.tsx
magnew: Yea, that's the right idea. I was saving space, but that's a losing battle :D

magnew:
> Typing "average" ont he value seems like a bad UI , perhaps we can replace this with a `toggle` boolean that `disables` the `value` input , or you can ask for feedback on #design

Yea, this is a bad UI 😄. I came into this trying to fix a bug in the existing 'feature', but you're right that it should probably really be an explicit feature. It's probably wouldn't be hard to just make a toggle, but should I break the existing behavior to do it? Maybe; I don't think a lot of people are using this.

magnew:
> Edit: I added a checkbox as a UI as well as a placement interface
> Screenshot 2024-03-25 at 19 03 49

As we discussed, you are correct. I added a checkbox and a way to place these.
Screenshot 2024-03-25 at 19 03 49

magnew:
> > Typing "average" ont he value seems like a bad UI , perhaps we can replace this with a `toggle` boolean that `disables` the `value` input , or you can ask for feedback on #design
> 
> Yea, this is a bad UI 😄. I came into this trying to fix a bug in the existing 'feature', but you're right that it should probably really be an explicit feature. It's probably wouldn't be hard to just make a toggle, but should I break the existing behavior to do it? Maybe; I don't think a lot of people are using this.

@rephus As we discussed, you are correct. I added a checkbox and a way to place these.
Screenshot 2024-03-25 at 19 03 49

rephus (changes requested):
The new UI looks great 
I found a bug though, when selecting a `date` field on the chart `How many orders we have over time ?` on a reference line, I get an white screen

image

magnew:
> > The new UI looks great I found a bug though, when selecting a `date` field on the chart `How many orders we have over time ?` on a reference line, I get an white screen
> > image
> > Perhaps we just need to disable this option for anythign that is not a number
> 
> Good catch. I did test dates, but I guess I missed a thing here. I think you're right though, maybe I'll just disable for dates. I doesn't make a lot of sense.

@rephus I fixed this crash. It was about the xAxis lines not being correct, not about the dates. I talked to Priyanka and we might need to change this based on some of the other chart config work. We can work it out when you are back.

rephus (approved):
Sorry for the delay, it looks good to me

magnew:
> Sorry for the delay, it looks good to me

No problem. I'm sorry it took me so long to get back to it!

PR #9475, opened 2024-03-20
fix: improve behavior when ordering non-grouped series, cleanup temporary color assignment code 

filp:
### Description:

- Cleans up temporary code introduced during the color mapping changes
- Makes color fallback behavior more reliable, and fixes some edge cases around sorting, removing and re-adding series.

Note that this change is not directly related to color mapping (though it uses some of the same code).

- Ensures re-ordering series preserves the "original" color (as well as manually set colors as before)

Kapture 2024-03-20 at 18 34 25

- Ensures removing and re-adding a series works, and picks up the correct fallback color (this fixes an issue where colors were sticky, and removing a series and re-adding it would re-add it with the old color, which may no longer be in the correct position)

Kapture 2024-03-20 at 18 36 03

- Works across different chart types

Kapture 2024-03-20 at 18 38 08

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -194,12 +194,17 @@ const VisualizationProvider: FC<React.PropsWithChildren<Props>> = ({
         }
     }, [resultsData?.metricQuery, columnOrder]);
 
+    /**
+     * Build a local set of fallback colors, used when dealing with ungrouped series.
+     *
+     * Colors are pre-calculated per-series, and re-calculated when series change.
+     */
     const fallbackColors = useMemo<Record<string, string>>(() => {
         if (!chartConfig?.config || chartConfig.type !== ChartType.CARTESIAN) {
             return {};
         }
 
-        const entries = Object.fromEntries(
+        return Object.fromEntries(
             (chartConfig.config.eChartsConfig.series ?? []).map((series, i) => {
                 return [
                     calculateSeriesLikeIdentifier(series).join('|'),
@@ -207,10 +212,6 @@ const VisualizationProvider: FC<React.PropsWithChildren<Props>> = ({
                 ];
             }),
         );
-
-        console.log(entries);
-
-        return entries;
     }, [chartConfig, colorPalette]);
 
     const handleChartConfigChange = useCallback(
@@ -256,6 +257,7 @@ const VisualizationProvider: FC<React.PropsWithChildren<Props>> = ({
             return isGroupedSeries(seriesLike)
                 ? calculateSeriesColorAssignment(seriesLike)
                 : fallbackColors[
+                      // Note: we don't use getSeriesId since we may not be dealing with a Series type here
                       calculateSeriesLikeIdentifier(seriesLike).join('|')
                   ];
         },
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -2,7 +2,6 @@ import {
     assertUnreachable,
     ChartType,
     getCustomDimensionId,
-    getDefaultSeriesColor,
     type ApiQueryResults,
     type ChartConfig,
     type DashboardFilters,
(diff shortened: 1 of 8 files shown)


packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -236,10 +256,12 @@ const VisualizationProvider: FC<React.PropsWithChildren<Props>> = ({
              */
             return isGroupedSeries(seriesLike)
                 ? calculateSeriesColorAssignment(seriesLike)
-                : colorPalette[seriesIndex % colorPalette.length] ??
-                      getDefaultSeriesColor(seriesIndex);
+                : fallbackColors[
```
filp, line 259: This is basically what was done in different places already, this change centralizes it in `VisualizationProvider` and exposes `getSeriesColor` as the single source of truth (ish, there's still some more cleanup to be done)

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/BasicSeriesConfiguration.tsx
```diff
@@ -20,19 +20,12 @@ type BasicSeriesConfigurationProps = {
...
-     * if shared colors are not enabled.
-     */
-    seriesIndex: number;
 };
 
 const BasicSeriesConfiguration: FC<BasicSeriesConfigurationProps> = ({
     isSingle,
     layout,
     series,
-    seriesIndex,
```
filp, line 35: Since series fallback colors are back to being pre-calculated (like they were before the color assignment changes), we no longer need the series index at all within child components.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/GroupedSeriesConfiguration.tsx
```diff
@@ -134,11 +132,7 @@ const GroupedSeriesConfiguration: FC<GroupedSeriesConfigurationProps> = ({
     const onDragEnd = useCallback(
         (result: DropResult) => {
             const allSerieIds = series.map(getSeriesId);
-            const seriesWithColor: Series[] = series.map((s, i) => ({
-                ...s,
-                color: getSeriesColor(s, i),
-            }));
-            const serie = seriesWithColor.find(
+            const serie = series.find(
```
filp, line 135: Grouped series always pick up a consistent color mapping, so this step is no longer doing anything.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/index.tsx
```diff
@@ -78,9 +78,9 @@ const SeriesTab: FC<Props> = ({ items }) => {
             const reorderedSeries = reorderedSeriesGroups.reduce<Series[]>(
                 (acc, seriesGroup) => [
                     ...acc,
-                    ...seriesGroup.value.map((s, i) => ({
+                    ...seriesGroup.value.map((s) => ({
                         ...s,
-                        color: getSeriesColor(s, i),
+                        color: getSeriesColor(s),
```
filp, line 83: We no longer need the series index to get a series color, since we're precalculating `seriesId->color` in the parent context.

packages/frontend/src/hooks/useChartColorConfig.tsx
```diff
@@ -91,13 +91,35 @@ export const isGroupedSeries = (series: SeriesLike) => {
     );
 };
 
+export const calculateSeriesLikeIdentifier = (series: SeriesLike) => {
```
filp, line 94: Extracted from further down below, mostly unchanged (now returns all identifier components as a tuple, instead of just the part we assume to be useful).

magnew: This makes sense. Some issues I saw def need this.

magnew (approved):
This makes sense and it solves the issue we talked about 🎉.

I found another edge case when testing it, but I don't think it's related to this patch. I'd be inclined to check this in and fix it separately: horizontal charts don't seem to pick up the colors right

filp:
Nice catch @magnew! Not 100% sure it's related, but definitely not-not related 😄 I'll try to fix it still as part of this if it's a quick adjustment, otherwise I'll add a follow-up.

magnew (approved):
Retested -- flipped axis issue is fixed. Ship it

Issue #7268, opened 2023-09-29
dashboard-level filters should override chart-level filters in the case that the chart is already filtering on the same field.


TuringLovesDeathMetal:
### Description

Right now, dashboard filters are "additive" to the chart filters in their tiles. For example, if I have a chart that's filtered on `date = 2023-01-01`, then a dashboard filter for `date = 2023-01-04`, the chart on that dashboard will be filtered on `date = 2023-01-01 AND date = 2023-01-04`. 

Instead, what I'm proposing is that if I have a chart that's filtered on a field, and it's in a dashboard where there's a filter applied to that _**same**_ field in the chart, then the dashboard filter should override the chart filter. 

Here's a really clear explanation of what the behaviour should be like: https://www.googlecloudcommunity.com/gc/Exploring-Curating-Data/Overriding-Look-Filters-in-the-Explore-Section-With-Dashboard/td-p/582846

### Acceptance criteria

- [ ] If I have a chart that is filtered on a field, then I apply a dashboard filter on teh same field, the chart filter should be overridden by the dashboard filter. 
- [ ] The operators _**do not**_ need to match, but the fields need to match.

### Why do we think this is a good idea? 

Right now, I have to create saved charts with no date filters (e.g. num users weekly) because otherwise, the chart could have data that's fltered in a different timeframe to everything else on the dashboard. If we added the `override`, then I could add a 14 week filter to my "num users weekly" chart so that when people open the saved chart, they're not running some massive query. But when it's added to a dashboard also being filtered on date, then the date filter could be overriden to 16 weeks (for example).

ZeRego:
Replacing filter values without context is dangerous.

Example 1
Big number chart "Total revenue" has the filter `order_status=completed`.
If the Dashboard has a filter `order_status=pending` it will override the chart and the data won't match the title.

Example 2
Bar chart "Last week total revenue" has the filter `date in last 1 week`
If the dashboard has a filter `date=2023-01-04` it will override the chart and the data won't match the title.


What if the user can mark what chart filters can be overridden? 
In Example 1, the editor would not allow the `order_status=completed` filter to be overridden as the chart highly depends on it.

In Example 2, the editor would rename the chart to "Total revenue" and allow the date filter to be overridden. This way the chart filter would act as a default filter in case the dashboard doesn't have a different value.

In cases where the dashboard has `date=any` should the chart still apply its filter or remove the filter?

TuringLovesDeathMetal:
[...]

PR #9490, opened 2024-03-21
feat: let dashboard filter overrides chart filter when its not required

Shifan-Gu:
Closes: 
https://github.com/lightdash/lightdash/issues/7268
### Description:

let dashboard filter override chart filter when 
1. chart filter is not marked as required
2. chart filter and dashboard filter has same target and operator value

Test:
Screenshot 2024-03-21 at 15 26 10
Screenshot 2024-03-21 at 15 26 36
Screenshot 2024-03-21 at 15 25 48

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -20,7 +20,9 @@ import {
     FilterType,
     isAndFilterGroup,
     isFilterGroup,
+    isFilterRule,
     UnitOfTime,
+    type AndFilterGroup,
     type DashboardFieldTarget,
     type DashboardFilterRule,
     type DashboardFilters,
@@ -30,6 +32,7 @@ import {
     type FilterGroupItem,
     type FilterRule,
     type Filters,
+    type OrFilterGroup,
 } from '../types/filter';
 import { type MetricQuery } from '../types/metricQuery';
 import { TimeFrames } from '../types/timeFrames';
@@ -179,6 +182,7 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
                     filterRuleDefaults.settings = {
                         unitOfTime: UnitOfTime.days,
                         completed: false,
+                        required: true,
                     } as DateFilterRule['settings'];
                 } else if (isTimestamp) {
                     const valueIsDate =
@@ -241,7 +245,7 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
     return {
         ...filterRule,
         values: values !== undefined && values !== null ? values : [],
-        settings: undefined,
+        settings: { required: true },
         ...filterRuleDefaults,
     };
 };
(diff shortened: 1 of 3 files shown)


TuringLovesDeathMetal:
seeing it in the product makes me realize that it’s a kind of horrible experience to have to think about this each time I add a filter... Which is super helpful! So thanks for making it possible to see that.
I’m definitely leaning more towards just having filters be overridden by dashboard filters, always. (it would break some things, but I think it would be overall a better experience).

^If that’s the case, then I just need to think a bit more about how we’d communicate this to customers and how we’d help mitigate the changes.

Just wanted to share an update to say that “I’ve seen this, but I’m just trying to figure out what to do next now”

TuringLovesDeathMetal:
@Shifan-Gu was playing around with this today (trying to make a demo explaining what will change) and I saw this behaviour: https://www.loom.com/share/18139ef4e690467bac2fe6383a49c824

Any chance we could remove this adding new filters bit? (I have a fear that it's going to be complicated because that's how all of the dashboard filters work...but let me know 🤞 )

Shifan-Gu:
> @Shifan-Gu was playing around with this today (trying to make a demo explaining what will change) and I saw this behaviour: https://www.loom.com/share/18139ef4e690467bac2fe6383a49c824
> 
> Any chance we could remove this adding new filters bit? (I have a fear that it's going to be complicated because that's how all of the dashboard filters work...but let me know 🤞 )

Hi Katie, nice catch 👍 i can fix this.

packages/common/src/utils/filters.ts
```diff
@@ -520,12 +524,53 @@ export const addFiltersToMetricQuery = (
     },
 });
 
+const findAndOverrideChartFilter = (
+    item: FilterGroupItem,
+    filterRulesList: FilterRule[],
+): FilterGroupItem => {
+    const overridableDashboardFilter = isFilterRule(item)
```
ZeRego, line 531: Note that there might be multiple dashboard filters for the same field. 

We should add unit tests for the following cases:

- if a dashboard has multiple filters for the same field, we replace them as a nested filter group.

```
// chart filters 
status=updated OR is_complete=true

// dashboard filters 
status=updated AND status not null

// result
(status=updated AND status not null) OR is_complete=true
```

- if the chart has multiple filters for the same field, we replace all of them

```
// chart filters 
status=updated OR status is null

// dashboard filters 
status=updated

// result
status=updated OR status=updated
```

packages/common/src/utils/filters.ts
```diff
@@ -520,12 +524,53 @@ export const addFiltersToMetricQuery = (
...
+        ? filterRulesList.find(
+              (filterRule) =>
+                  filterRule.target.fieldId === item.target.fieldId &&
+                  filterRule.operator === item.operator &&
```
ZeRego, line 535: Is there a reason why we need to match the filter operators?

ZeRego:
We already messaged you privately but wanna make it public so everyone is aware of the state of this feature.
We decided to go back to the original idea where dashboard filters should override the chart filters so we won't need the `required` option for each chart filter. 
You can reopen the PR once it is updated with the new title/description/code. 🙏

Issue #9423, opened 2024-03-18
Switching a reference line to the wrong type crashes the explore page


magnew:
### To Reproduce

- Choose a chart with a date and numerical axis
- Set up a reference line on the numerical axis
- Change the line to reference the date field
- Page crash

magnew:
I can't repro this an hour later 😕. Closing for now. Will reopen if I see it again.

PR #9491, opened 2024-03-21
fix: page crash when setting date to invalid value

magnew:
Closes: #9423

### Description:

Checks that a date is valid before passing it to the date inputs in a reference line.

To repro the original bug:
- Add a reference line to a chart with a date axis (eg "How many orders we have over time ?" in jaffle shop)
- Set the value of the reference line to something like 'a' on a field that allows it (date input doesnt)
- Switch to the date field 
- 😢 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLine.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLine.tsx
@@ -76,12 +76,17 @@ const ReferenceLineValue: FC<ReferenceLineValueProps> = ({
     if (isCustomDimension(field)) return <></>;
     if (isDateItem(field)) {
         if (isDimension(field) && field.timeInterval) {
+            // Uses the current date if the provided value is invalid
+            const parsedDate = dayjs(value).isValid()
+                ? dayjs(value).toDate()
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLine.tsx
```diff
@@ -76,12 +76,17 @@ const ReferenceLineValue: FC<ReferenceLineValueProps> = ({
     if (isCustomDimension(field)) return <></>;
     if (isDateItem(field)) {
         if (isDimension(field) && field.timeInterval) {
+            // Uses the current date if the provided value is invalid
+            const parsedDate = dayjs(value).isValid()
+                ? dayjs(value).toDate()
+                : new Date();
```
joaoviana, line 82: could we do 
```suggestion
                : dayjs().toDate()
```
here?

joaoviana (approved):

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/ReferenceLine.tsx
magnew: Sure

Issue #9264, opened 2024-03-06
I want to be able to run a dashboard only when a filter is selected


TuringLovesDeathMetal:
### Description

Right now, when you open a dashboard, everything automatically loads. 

This is fine, if your data isn't too big. But sometimes, I build a dashboard which is _**meant**_ to be filtered (e.g. in Lightdash, our Cloud Activity Tracker). 

If I have a really big dataset, then I actually don't want my dashboard to load until I've filtered my dashboard. 

There should be a way to set the dashboard to only run once a filter value has been selected. 

E.g. here, I would want to be able to set the dashboard to `load on filter` or something. Then pick the `date week` filter as "needing to have a value"

<img width="390" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/c33ffe39-540d-430f-a562-8d3c42973549">

owlas:
@TuringLovesDeathMetal and @joaoviana is this something we should set on a filter-level in the filter config panel.

i.e. "this filter must have a value"

joaoviana:
> @TuringLovesDeathMetal and @joaoviana is this something we should set on a filter-level in the filter config panel.
> 
> i.e. "this filter must have a value"

I was thinking this could be on a _global_ level, like on this example:
<img width="1003" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/46a757f3-c936-4363-abee-38b16086a4db">

Where you can select which filters would _block_ the dashboard if no filter is applied. 

I can, instead, follow your approach!

hamzahc1:
When this merges - it should have a headway update @joaoviana  😄

joaoviana:
> When this merges - it should have a headway update @joaoviana  😄 

And will also update docs @hamzahc1 !

PR #9505, opened 2024-03-22
feat: require some filters to have values before running dashboard 

joaoviana:
Closes: #9264

### Description:

Adds new property `required` to filters
Displays TileBase when the dashboard is locked
Can set filter as required when creating it

If the dashboard is locked, show helpful message to end user stating that they should interact with the required filters first

Feedback I'm looking for:
- Styling of Locked state dashboard, more precisely: Modal copy, and UI state of required filters that haven't been set yet.
- Checkbox for Filter settings `require value for dashboard to run`: UI and copy

Demo

https://github.com/lightdash/lightdash/assets/7611706/e348d136-4803-4536-a0f7-45967af66a60

### Reviewer actions


packages/common/src/types/filter.ts
```diff
@@ -46,6 +46,7 @@ export interface FilterRule<
     target: T;
     settings?: S;
     disabled?: boolean;
+    required?: boolean;
```
joaoviana, line 49: Added this extra property, LMK if I should rename to something more specific like: 
`requiredToHaveValue` or similar

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -624,42 +638,74 @@ const Dashboard: FC = () => {
...
+                                To reveal this dashboard's insights, simply
+                                unlock it by setting values to filters that are
+                                required.
+                            </Text>
+                            <Text>
+                                Your organization has chosen these specific
+                                filters to ensure you get precisely the data you
+                                need.
+                            </Text>
+                        </Stack>
```
joaoviana, line 699: copy to review here ^

packages/common/src/types/filter.ts
joaoviana: From this PR: https://github.com/lightdash/lightdash/pull/9490
Seems like there will be a new property `required` that will live in `settings` 

Got suggestions on naming @owlas ?

PriPatel (approved):
ui changes look good to me

packages/common/src/types/filter.ts
owlas: Are these filter rules not also used for saved charts too? Or is this property only being attached to dashboard filters?

It would be confusing to apply this to saved charts if it has no effect

packages/frontend/src/providers/DashboardProvider.tsx
```diff
@@ -570,6 +575,38 @@ export const DashboardProvider: React.FC<
         [resultsCacheTimes],
     );
 
+    // Filters that are required to have a value set
+    const requiredDashboardFilters = useMemo(
+        () =>
+            dashboardFilters.dimensions
+                .filter((f) => f.required && f.disabled)
```
owlas, line 582: what is the difference between required and disabled?

packages/frontend/src/components/DashboardFilter/Filter.tsx
```diff
@@ -169,68 +177,107 @@ const Filter: FC<Props> = ({
...
-                        size="xs"
-                        variant={isTemporary ? 'outline' : 'default'}
-                        bg="white"
-                        rightIcon={
-                            (isEditMode || isTemporary) && (
-                                <CloseButton size="sm" onClick={onRemove} />
+                    <Indicator
+                        size={24}
+                        position="top-end"
+                        disabled={
```
owlas, line 183: There's a logical error somewhere here. If I set the date type but not the actual value then the overlay disappears but the dashboard doesn't run.

Can we centralise this logic? They should be in sync.

CleanShot 2024-03-27 at 10 13 33@2x

packages/common/src/types/filter.ts
owlas: Just found a practical example of this. If I "explore from here" I can see that the URL for the saved chart contains a filter with `required: true` and this will get saved onto the chart. What's the expected behaviour here? Could the filter somehow interfere with other dashboard filters later?

We need to type the dashboard filters separately if they have different properties and have an explicit filter conversion when we explore from here.

owlas (commented):
After reviewing, it's actually not clear to me from the UI that only the blue filter needs to be set. The copy "setting the filter values above" sounds like I need to set all of them.

Added more comments inline.

packages/common/src/types/filter.ts
joaoviana: Thanks for pointing this out @owlas, makes no sense to save this onto the chart! I've created a PR to handle the `required` + this filtering before this piece of work here: https://github.com/lightdash/lightdash/pull/9559

packages/frontend/src/components/DashboardFilter/Filter.tsx
joaoviana: Thanks for sharing this! 
Logic is centralised and fixed here: https://github.com/lightdash/lightdash/pull/9505/commits/c65f39cf8d788fe2681ed2db780bdf59510fdd97

packages/frontend/src/providers/DashboardProvider.tsx
joaoviana: disabled: no default value set / is any
required: requires a value to be set

joaoviana: https://github.com/lightdash/lightdash/pull/9505/commits/7fbe5bf65a3ca1514f87ceb22320ac48bf9957fe

joaoviana:
Putting this PR on draft until we get https://github.com/lightdash/lightdash/pull/9559 reviewed/merged

packages/common/src/types/filter.ts
joaoviana: This has been addressed and we can use `required` in dashboard filters in this PR, now that the above PR has been merged

joaoviana:
image

Added `required` indicator, which will hopefully help our users

joaoviana:
Fixed the bug you've raised on this commit @almeidabbm: https://github.com/lightdash/lightdash/pull/9505/commits/8c2f561ad405133b5e880bb53139b958bc6c2eba

almeidabbm (approved):
LGTM 🚀

Issue #9501, opened 2024-03-22
New login stops spinning before redirect


magnew:
### Description

The login spinner shouldnt stop spinning before redirect

PR #9507, opened 2024-03-22
fix: missing spinner when automatically logged in

magnew:
Closes: #9501 

### Description:

In the case where the user is know and they are automatically re-directed, we were not showing loading spinners while waiting for the redirect. Also, the button text was changing and it shouldn't

### Reviewer actions


magnew:
Demo
Kapture 2024-03-22 at 16 33 42

owlas (approved):

PR #9514, opened 2024-03-25
fix: update table calculation name in chart configs

almeidabbm:
Closes: [#7996](https://github.com/lightdash/lightdash/issues/7996)

### Description:

- Doesn't reset chart config when updating table calculation
- Adds tests

https://github.com/lightdash/lightdash/assets/22939015/105267b4-ca58-462d-ab45-145964e49924

### Reviewer actions


packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
```diff
@@ -177,7 +178,13 @@ const TableCalculationModal: FC<Props> = ({
                             <Button variant="outline" onClick={onClose}>
                                 Cancel
                             </Button>
-                            <Button type="submit"> Save </Button>
+                            <Button
+                                type="submit"
+                                data-testid="table-calculation-save-button"
+                            >
+                                {' '}
+                                Save{' '}
```
joaoviana, line 186: can remove these extra spaces

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -482,16 +486,79 @@ const useCartesianChartConfig = ({
...
+            return {
+                ...prev,
+                yField: newYFields,
+            };
+        });
+    }, [dirtyLayout?.yField, tableCalculationsMetadata]);
+
     // Set fallout layout values
     // https://www.notion.so/lightdash/Default-chart-configurations-5d3001af990d4b6fa990dba4564540f6
     useEffect(() => {
```
joaoviana, line 548: Wondering if we could handle the table calculation metadata logic all in this `useEffect` so that we don't call `setDirtyLayout` _more than once_ - what do you think @almeidabbm ? 
Also, would recommend adding a comment to make it clear why we're doing it 👍

packages/frontend/src/hooks/usePieChartConfig.ts
```diff
@@ -192,8 +195,20 @@ const usePieChartConfig: PieChartConfigFn = (
         if (isLoading || allNumericMetricIds.length === 0) return;
         if (metricId && allNumericMetricIds.includes(metricId)) return;
 
+        // If the metric is a table calculation, we need to find the new name
```
joaoviana, line 198: Like this comment here - I like it but I would go further and explain why we have to find the new name - I think this should be added to every use...Config hook we changed - what do you think?

packages/frontend/src/providers/ExplorerProvider.tsx
```diff
@@ -1125,6 +1194,12 @@ function reducer(
                 );
             return {
                 ...state,
+                metadata: {
+                    tableCalculations:
+                        state.metadata?.tableCalculations?.filter(
+                            (tc) => tc.name !== action.payload,
+                        ),
+                },
```
joaoviana, line 1202: nice - is there somewhere we can _clear_ this `metadata` property to keep it clean? 
Like on `discard` and/or on `save` 🤔

joaoviana (commented):
Left some Qs 👍

packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
almeidabbm: For sure! These must've been from auto formatting? 🤔

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
almeidabbm: Yeah I will try and do it, I had it out because I didn't want to pollute that hook too much as it is too big, but will find a way to make it readable 💪

packages/frontend/src/providers/ExplorerProvider.tsx
almeidabbm: I believe setting it on `update` and removing on `tc remove` should be enough as it will always be the source of truth after every TC update on the explore.
I tried clearing it after the request to get the results but the problem is that after the 3rd table calculation update it will go back to a default value since it finds another mismatch in the name and doesn't have the metadata to update it in the hook :(

packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts
```diff
@@ -452,55 +450,126 @@ const useCartesianChartConfig = ({
...
         if (availableFields.length > 0) {
             setDirtyLayout((prev) => {
+                /**
+                 * Get the fields with the current table calculation names when they are a table calculation with the old name
+                 * otherwise keep the fields as they are
+                 */
+                const xField = getXField(prev?.xField);
+                const yFields = getYFields(prev?.yField);
```
joaoviana, line 549: so much clearer! 👏

packages/frontend/src/providers/ExplorerProvider.tsx
```diff
@@ -216,6 +217,9 @@ type Action =
 export interface ExplorerReduceState {
     shouldFetchResults: boolean;
     expandedSections: ExplorerSection[];
+    metadata?: {
+        tableCalculations?: TableCalculationMetadata[];
```
joaoviana, line 221:
```suggestion
// Temporary state that tracks changes to `table calculations` - keeps track of new name and previous name to ensure these get updated correctly when making changes to the layout & config of a chart 
        tableCalculations?: TableCalculationMetadata[];
```

joaoviana (approved):

packages/frontend/src/providers/ExplorerProvider.tsx
joaoviana: I just seen that you've added the filtering on removal of a TC - thanks!

Issue #9128, opened 2024-02-27
Use smaller variants for the inputs in the chart config panel


PriPatel:
https://www.notion.so/lightdash/Explore-View-Changes-5d71ec3783f14c5187bfa1e2d29085ec?pvs=4#880ce8040b15486b92f30b32119221ed

PR #9528, opened 2024-03-26
feat: cartesian chart config UI/UX changes

joaoviana:
Closes: #9128

### Description:

This PR focusses on changing the UI/UX of cartesian chart config. 

## Demo

https://github.com/lightdash/lightdash/assets/7611706/034dab25-1939-44b6-a01d-d7f0aa2c030f

Many things are affected here. Find below the run down of changes

#### Refactor folder structure
- To keep things a tad more consistent, renamed files + added to folders to match the names of the Tabs: 
  - Layout
  - Series
  - Legend
  - Grid

#### `Config` component:
Example: 
image

- X-axis is the label of a Config group
- A Config Group can have an action button like: `+Add` or `Flip axes`
- Each child of the `Config.Group` should be smaller, have lighter font colour, higher font weight, etc (see theme override description below ⬇️ )
- This will be used in all components for each type of Config: Cartesian, Big Number, Table, Pie, etc.

#### Theme override:
in `ConfigTabs/mantineTheme` we override the theme only for the Tabs and their panels:
- Most components should be `xs`
- Most components' labels should have a custom font weight & colour
- Some `label`s should have less padding
- This way we avoid explicitly setting `size="xs"` to each component

##### `AddButton`
- This is used a lot so created a common component

##### `EditableText`
Example: 
image
- To avoid having a Label and an input, I've created this `EditableText` component that is basically a styled input that just looks like plain text but on hover users know they can edit it inline.

##### `GrabIcon`
image
- Reusable grab icon box so that we don't have to edit the same component on every file that uses it. I can see this being reused in more places, but for now, it's on `ChartConfigPanel` folder. 

## Tab-by-tab review

The following comments assume that things are smaller and other styles are applied

### Layout
- Flip axes is now an action icon

### Series
- Can edit labels 
- `ChartTypeSelect` doesn't have text anymore (but kept color as transparent for screen-readers)
-  Inputs in Grouped series are now full width:
image

### Axes
- When x-axis range is custom, configuration is inline
- Sort segmented control is now icons

### Display
- Small changes to show/hide of Legend 
- Clearer copy on Segmented controls
- New component: `UnitInputsGrid` that allows us to remove the input labels, now that we display the inputs on the position they belong to. Only show this if user sets a Custom position
image
- Reference Lines are `Accordions` now, with a clearer `remove` button, and more _compact_ layout
image

### Margins
- Uses the component `UnitInputsGrid` with different configuration (values for position Margin instead) 
image

### Reviewer actions


joaoviana:
Remaining: E2E tests

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+            }
+            return acc;
+        },
+        [false, false],
+    );
+
+    return (
+        <Stack>
+            <Config>
+                <Config.Group>
```
magnew, line 79: Niiice

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+                                />
+                            </Group>
+                        )}
+                    </Group>
+                </Config.Group>
+            </Config>
+
+            <Config>
+                <Config.Group>
+                    <Config.Label>{`${
```
magnew, line 163: These are like, group labels, right? 

Oh I see, you have Label and SubLabel. I think I'd prefer GroupLabel for the groups and just Label for the sub things. I think it might be clearer how to use them. 

Bu this is *very minor*, feel free to ignore.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+                                    defaultValue={
+                                        dirtyEchartsConfig?.xAxis?.[0].rotate ||
+                                        0
+                                    }
+                                    min={0}
+                                    max={90}
+                                    step={15}
+                                    maw={54}
+                                    rightSection="°"
+                                    onChange={(value) => {
```
magnew, line 151: Some of these inputs are onChange and some are onBlur. It can be a little confusing to use. 

But now that I play around with it, this is true in main as well. I guess maybe that can be food for future thought.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/ConfigTabs/mantineTheme.ts
magnew, line 1: Its pretty cool that Mantine enables this

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SingleSeriesConfiguration.tsx
```diff
@@ -52,109 +54,92 @@ const SingleSeriesConfiguration: FC<Props> = ({
...
     return (
-        <Group noWrap={!wrapExtraComponents} spacing={0}>
-            <Group
-                noWrap
-                spacing="xs"
-                sx={
-                    // TODO: this is here to position the color
-                    // picker correctly in the grouped vs ungrouped cases.
-                    // This isn't a great answer and we probably could
-                    // clean up the layout in this file after the other parts
```
magnew, line 71: 🎉

magnew (approved):
I did a lot of testing this and it look super good. Nice work! 

Left a few minor comments but nothing blocking. This code looks pretty clean too. Mantine FTW.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
joaoviana: yes they are! I like that change yes. 
Lemme change everywhere!
> I think it might be clearer how to use them.

Makes sense!

joaoviana: Ok, one question before I make the change: 
There's already a `LabelGroup`, which is a Group for the Label and it can have another child element like the `AddButton`. 
If I have `GroupLabel`, then there's two v similar components. What would you suggest doing in this case?

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
joaoviana: That's true. I'm going to have a look at every single section where we do an `onBlur` and sections where I've changed from `onBlur` to `onChange` and figure out what's best 👍

joaoviana: And here it is! https://github.com/lightdash/lightdash/pull/9528/commits/cc6befd3fae4268f9678a909e327b91ded160fe3

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
joaoviana: > Are you saying there's already a Config.LabelGroup?

yup, there's one already! 

So, with this in mind: 
- `Config.Group`
- `Config.Heading` - was `Label` before
- `Config.GroupHeading` - was `GroupLabel` before
- `Config.Label` - was `SubLabel` before

joaoviana: What do you think? ^

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/common/UnitInputsGrid.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    size="xs"
+                    w={80}
+                    name="top"
+                    units={units}
+                    value={config.top || defaultConfig.top}
+                    defaultValue={defaultConfig.top}
+                    onChange={(value) => handleUpdate(Positions.Top, value)}
+                />
+            </Flex>
+            <EmptySpace />
```
IrakliJani, line 80: There's this if it's helpful   https://v6.mantine.dev/core/space/ 🤔

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/index.tsx
```diff
@@ -1,23 +1,24 @@
...
-    SimpleGrid,
     Stack,
     Switch,
-    Text,
 } from '@mantine/core';
-import startCase from 'lodash/startCase';
 import { type FC } from 'react';
 import { useParams } from 'react-router-dom';
-import UnitInput from '../../../common/UnitInput';
+import { useToggle } from 'react-use';
```
IrakliJani, line 17: you probably wanted to import from mantine?

PR #9543, opened 2024-03-27
fix: TS project configuration errors

filp:
### Description:

Fixes a couple of annoying (editor/TS) errors related to the workspace configuration, particularly around TS project references.

Example in VSCode:

image

- Removes `noEmit` and sets `emitDeclarationsOnly` for `frontend` (referenced projects cannot have `noEmit: false)`
- Adds a `declarationsDir` for `frontend` under `build/`, to avoid accidentally generating `.d.ts` files alongside source files
- Adds `composite: true` to the `e2e` package, since it's referenced via the root TS project (referenced projects must have `composite: true`).

Should otherwise have no noticeable impact.

### Reviewer actions


packages/frontend/tsconfig.json
```diff
@@ -16,7 +16,8 @@
         "moduleResolution": "node",
         "resolveJsonModule": true,
         "isolatedModules": true,
-        "noEmit": true,
+        "emitDeclarationOnly": true,
+        "declarationDir": "./build/d",
```
ZeRego, line 20:
```suggestion
        "declarationDir": "./build/types",
```

ZeRego (approved):

packages/frontend/tsconfig.json
filp: Great point 🤔

PR #9559, opened 2024-03-27
chore: add required field & improve metric query with dashboard filters

joaoviana:
Relates to: #9264 & PR:https://github.com/lightdash/lightdash/pull/9505

### Description:
After the comment here: https://github.com/lightdash/lightdash/pull/9505#discussion_r1540851368, I've added the following changes
- Added `required` optional property to `DashboardFilterRule`
- Before returning the `metricQuery` in `charts-and-results`, ensure that the converted filters do not include any dashboard-only properties

~After the comment here: https://github.com/lightdash/lightdash/pull/9505#discussion_r1540851368, I've added the following changes:~
~- Added `required` optional property to `DashboardFilterRule`~
~- Before sending the `dashboardFilters` to the `chat-and-results`, we should discard unnecessary/unrelated-to-charts properties: `tileTargets`, `required`, and `label` - the saved chart doesn't know what to do with these so it's best to remove them before handling the dashboard filters applied to a chart tile.~ 
~- With the above in mind, this PR defines 2 new types: `DashboardFiltersForTile` & `DashboardFilterRuleForTile`, which omit the 3 properties we don't want the chart to handle~
~- In `getDashboardFilterRulesForTile` we remove these 3 properties~

~NOTE:~
~On `useTableConfig` we use `dashboardFilters` that come from `useDashboardFiltersForTile` which has the new structure now. The latter hook (`useDashboardFiltersForTile`) is also used when getting chart results on dashboards (`chart-and-results`) meaning both sides are now converted to the expected format.~

### Reviewer actions

### DIFF
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -127,9 +127,6 @@ export type DashboardFilters = {
     tableCalculations: DashboardFilterRule[];
 };
 
-/**
- * See DashboardFilterRuleForTile for details on the structure of individual filter rules.
- */
 export type DashboardFiltersForTile = {
     dimensions: DashboardFilterRuleForTile[];
     metrics: DashboardFilterRuleForTile[];
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -70,13 +70,13 @@ export type DashboardFilterRule<
 };
 
 /**
- * DashboardFilterRuleForTile is a DashboardFilterRule without the `tileTargets`, `label`, and `required` fields.
+ * DashboardFilterRuleForTile is a DashboardFilterRule without properties that are unknown to FilterRule: e.g. `tileTargets`, `label`, and `required` fields
  * This is used to represent the filter rules that are applied to a specific tile.
  * This is also to ensure we don't polute the filters on a chart tile with unnecessary fields that can't be used when viewing a chart in isolation.
  */
-export type DashboardFilterRuleForTile = Omit<
+export type DashboardFilterRuleForTile = Pick<
     DashboardFilterRule,
-    'tileTargets' | 'required' | 'label'
+    'id' | 'target' | 'operator' | 'values' | 'settings' | 'disabled'
 >;
 
 export type FilterDashboardToRule = DashboardFilterRule & {
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -66,7 +66,18 @@ export type DashboardFilterRule<
 > = FilterRule<O, T, V, S> & {
     tileTargets?: Record<string, DashboardTileTarget>;
     label: undefined | string;
+    required?: boolean;
 };
+
+/**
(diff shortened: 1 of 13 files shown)


packages/common/src/types/filter.ts
```diff
@@ -116,6 +127,15 @@ export type DashboardFilters = {
     tableCalculations: DashboardFilterRule[];
 };
 
+/**
```
owlas, line 130: this comment is redundant from the code below it

packages/common/src/utils/filters.ts
```diff
@@ -425,45 +427,57 @@ export const getFilterRulesByFieldType = (
...
             if (tileConfig === false) {
-                return null;
+                return acc;
             }
 
+            const dashboardFilterRuleForTile = filter;
+
+            // Delete properties that are not needed for filters on tiles
+            if ('tileTargets' in dashboardFilterRuleForTile)
+                delete dashboardFilterRuleForTile.tileTargets;
```
owlas, line 445: should avoid mutation, it could change the filters passed in as arguments, did you mean to copy the filter?

`const dashboardFilterRuleFortile = {...filter}` ?

packages/common/src/types/filter.ts
```diff
@@ -66,8 +66,19 @@ export type DashboardFilterRule<
...
     label: undefined | string;
+    required?: boolean;
 };
 
+/**
+ * DashboardFilterRuleForTile is a DashboardFilterRule without properties that are unknown to FilterRule: e.g. `tileTargets`, `label`, and `required` fields
+ * This is used to represent the filter rules that are applied to a specific tile.
+ * This is also to ensure we don't polute the filters on a chart tile with unnecessary fields that can't be used when viewing a chart in isolation.
+ */
+export type DashboardFilterRuleForTile = Pick<
```
owlas, line 77: Isn't this type identical to `FilterRule` ?

- `MetricQuery` has `filters` which is type `Filters` contains `FilterRule`s
- `Dasbboard` has `dashboardFilters` has type `DashboardFilters` contains `DashboardFilterRule`s

The problem with the last code was that explore from here, was including dashboard filter rule properties as regular filter rules. We need to be explicit when we're casting from `DashboardFilterRule` to a simple `FilterRule`

owlas (commented):
I think the new type is identical to the existing `FilterRule`! The conversion from DashboardFilterRule makes sense but it should convert to a simple FilterRule instead of DashboardFilterRule.

packages/common/src/utils/filters.ts
joaoviana: oops! Missed that on a refactor - thank you

packages/common/src/types/filter.ts
joaoviana: Almost identical! 
DashboardFilterRule has DashboardFieldTarget: 
```
export type DashboardFieldTarget = {
    fieldId: string;
    tableName: string;
};

```

whereas FilterRule has FieldTarget: 

```
export type FieldTarget = {
    fieldId: string;
};
```

Checking if making the complete conversion to FilterRule works, thanks!

owlas (approved):

Issue #9555, opened 2024-03-27
Error "there must be 1 user in this space" when removing last user with direct access.


ZeRego:
### Description

<img width="352" alt="Screenshot 2024-03-27 at 16 15 58" src="https://github.com/lightdash/lightdash/assets/9117144/58acf08b-b01e-4c02-b244-d4dfc202875d">

We should remove this check as there is always an org admin with access to spaces.

PR #9561, opened 2024-03-27
fix: can remove all users access from space

ZeRego:
Closes: #9555 

### Description:

Before we show an error when we try to remove the last user with direct access from a space. Even though there are org/project admins with inherited access.

Screenshot 2024-03-27 at 16 51 16

Now we can remove them. We don't need this check as there will always be an org admin with access.

Screenshot 2024-03-27 at 16 49 35

### Reviewer actions


joaoviana (approved):
👍

Issue #9551, opened 2024-03-27
I have to scroll horizontally/manually expand the panel width to access the 3-dot menu in the metrics/dimension sidebar


PriPatel:
I could only reproduce in analytics, but not the demo. 
<img width="426" alt="Screenshot 2024-03-27 at 15 31 57" src="https://github.com/lightdash/lightdash/assets/12776522/ce24ef25-6d0e-4ac2-a745-0ea7c7ff84f7">

- [ ] I expect to see the three dot menu on the field I've hovered over currently
- [ ] I do not expect to ever see a horizontal scroll bar in the sidebar

PR #9562, opened 2024-03-27
fix: apply viewport child block display

joaoviana:
Closes: #9551

### Description:

Fixed by overriding the `viewport`'s child element `display` value from `table` to `block`. Couldn't find a selector with higher specificity than the inline styling that radix x Mantine apply to the `viewport`'s child

https://github.com/lightdash/lightdash/assets/7611706/05a144cd-f58c-49fb-84cd-799c7e451877

### Reviewer actions

### DIFF
--- a/packages/frontend/src/mantineTheme.ts
+++ b/packages/frontend/src/mantineTheme.ts
@@ -114,7 +114,7 @@ export const getMantineThemeOverride = (overrides?: {
                             },
                     },
                     viewport: {
-                        '& > div': {
+                        '.only-vertical & > div': {
                             display: 'block !important', // Only way to override the display value (from `table`) of the Viewport's child element
                         },
                     },
--- a/packages/frontend/src/mantineTheme.ts
+++ b/packages/frontend/src/mantineTheme.ts
@@ -113,6 +113,11 @@ export const getMantineThemeOverride = (overrides?: {
                                 ),
                             },
                     },
+                    viewport: {
+                        '& > div': {
+                            display: 'block !important', // Only way to override the display value (from `table`) of the Viewport's child element
+                        },
+                    },
                 }),
             },
         },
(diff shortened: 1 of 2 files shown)


packages/frontend/src/mantineTheme.ts
```diff
@@ -113,6 +113,11 @@ export const getMantineThemeOverride = (overrides?: {
                                 ),
                             },
                     },
+                    viewport: {
+                        '& > div': {
+                            display: 'block !important', // Only way to override the display value (from `table`) of the Viewport's child element
+                        },
+                    },
```
ZeRego, line 120: This change would remove horizontal scroll in all ScrollAreas.
We can apply this behaviour via a classname:

```
            <ScrollArea variant="primary" offsetScrollbars scrollbarSize={8} className={'only-vertical'}>
```

```
                    viewport: {
                        '.only-vertical & > div': {
                            display: 'block !important', // Only way to override the display value (from `table`) of the Viewport's child element
                        },
                    },
```

packages/frontend/src/mantineTheme.ts
```diff
@@ -113,6 +113,11 @@ export const getMantineThemeOverride = (overrides?: {
                                 ),
                             },
                     },
+                    viewport: {
+                        '.only-vertical & > div': {
+                            display: 'block !important', // Only way to override the display value (from `table`) of the Viewport's child element
```
IrakliJani, line 118: I don't understand why would they use `display: table`... any explanations?

IrakliJani: reason why im asking is that mantine use ScrollArea in their own components like dropdowns and I've seen few bugs because of that...

IrakliJani (approved):
tested.. looks good

packages/frontend/src/mantineTheme.ts
joaoviana: Here is why! https://github.com/radix-ui/primitives/blob/main/packages/react/scroll-area/src/ScrollArea.tsx#L180

```
 {/**
           * `display: table` ensures our content div will match the size of its children in both
           * horizontal and vertical axis so we can determine if scroll width/height changed and
           * recalculate thumb sizes. This doesn't account for children with *percentage*
           * widths that change. We'll wait to see what use-cases consumers come up with there
           * before trying to resolve it.
           */}
```

Issue #4137, opened 2023-01-10
431 error when url is too long


IrakliJani:
### Description

![image](https://user-images.githubusercontent.com/962095/211615558-35cb0b8c-4cda-4cc4-a125-a8ae1d82a545.png)

### To Reproduce
There are multiple waits to replicate this:
- In a dashboard with a lot of filters, click "explore from here" in a chart tile
- Or by copy-pasting the explore URL
1. Click `Run query from tables`
2. Click Payments table - from seed data
3. Add the following fields and filters:
![image](https://github.com/lightdash/lightdash/assets/22939015/a616e2cd-349a-4c22-b89a-053e8c5572ea)
4. Click `Run query`
5. Copy the URL and paste on a separate tab

rephus:
Fixed by https://github.com/lightdash/lightdash/pull/8594

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C05Q96U8FHN/p1710494236558959?thread_ts=1710494236.558959&cid=C05Q96U8FHN

TuringLovesDeathMetal:
I'm going to close this issue as not planned - it hasn't come up from another customer in at least 3 months, so I'm assuming something has changed...can reopen if needed

PR #9581, opened 2024-03-29
fix: Avoid long URLs in SQL Runner

filp:
Part of #4137, does not close it since this change only affects the SQL Runner for now.

The vast majority of this code is reusable for other similar cases, so it's a matter of migrating that over for other cases, once tested & validated.

### Description:

- Introduces `useStateCache` utility
- Introduces `update` action for `ShareUrl` controller, model
- Introduces new `useSqlRunnerRoute` hook to replace existing sql runner route management code. 
  - Loads data from 'legacy' parameters, if present
  - User operations are saved to local storage, under a cache id appended as a query parameter
  - When a user runs a query, the local cache is flushed to a ShareURL, and the url updated accordingly
  - The same share URL is reused across a session
  - Cache writes are debounced
- Introduces `shareNanoId` prop to `ShareShortLinkButton` to reuse an existing share ID when using the permalink/share button on the SQL Runner

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/shareController.ts
+++ b/packages/backend/src/controllers/shareController.ts
@@ -2,12 +2,14 @@ import {
     ApiErrorPayload,
     ApiShareResponse,
     CreateShareUrl,
+    UpdateShareUrl,
 } from '@lightdash/common';
 import {
     Body,
     Get,
     Middlewares,
     OperationId,
+    Patch,
     Path,
     Post,
     Request,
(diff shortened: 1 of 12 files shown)


filp:
Closing - deprioritised for now.

filp:
This PR gets us maybe 99.9% to target, but it's flawed in annoying ways that could prove more problematic than it's worth at this point:

- Chart config in the sql runner is indirectly tied to the query result, in a way that makes it tricky(ish) to manage draft from saved state
- There's a ton of intermediate state in the runner, and I had to introduce more to handle things like flushing data to cache without losing responsiveness; there's almost certainly a much better way to do this
- This PR introduces some unintended rerenders that are visibly annoying
- This PR requires blocking the SQL runner until share data is done loading, and then requires some awkward dancing around to once again distinguish draft from saved state (e.g adding and removing search params to keep track of state precendence)

Also not directly related but kinda:

- `searchParams` handling is awkward, we should update to latest react-router version
- `useLocalStorage` hook in Mantine <v7 handles initial state in a lazy way, which requires going to `localStorage` directly to get initial data to use with `useLocalStorage`

PR #9583, opened 2024-03-29
chore: remove some styled components

magnew:
Removes some of the remaining styled components. There are some more, but this is progress. These were a little hard to verify, but I think I tested all cases.

### DIFF
--- a/packages/frontend/src/components/ForbiddenPanel.tsx
+++ b/packages/frontend/src/components/ForbiddenPanel.tsx
@@ -1,4 +1,4 @@
-import { Anchor } from '@mantine/core';
+import { Anchor, Box } from '@mantine/core';
 import { IconLock } from '@tabler/icons-react';
 import { type FC } from 'react';
 import { Link } from 'react-router-dom';
@@ -7,7 +7,7 @@ import SuboptimalState from './common/SuboptimalState/SuboptimalState';
 
 const ForbiddenPanel: FC<{ subject?: string }> = ({ subject }) => {
     return (
-        <div style={{ marginTop: '30vh' }}>
+        <Box mt="30vh">
             <SuboptimalState
                 title={`You don't have access${
                     subject ? ` to this ${subject}` : ''
@@ -34,7 +34,7 @@ const ForbiddenPanel: FC<{ subject?: string }> = ({ subject }) => {
                 }
                 icon={IconLock}
             />
-        </div>
+        </Box>
     );
 };
 
--- a/packages/frontend/src/components/ForbiddenPanel.tsx
+++ b/packages/frontend/src/components/ForbiddenPanel.tsx
@@ -2,17 +2,12 @@ import { Anchor } from '@mantine/core';
 import { IconLock } from '@tabler/icons-react';
 import { type FC } from 'react';
 import { Link } from 'react-router-dom';
-import styled from 'styled-components';
 import { Can } from './common/Authorization';
 import SuboptimalState from './common/SuboptimalState/SuboptimalState';
 
-const ForbiddenPanelWrapper = styled.div`
-    margin-top: 30vh;
-`;
-
 const ForbiddenPanel: FC<{ subject?: string }> = ({ subject }) => {
     return (
-        <ForbiddenPanelWrapper>
+        <div style={{ marginTop: '30vh' }}>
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/ForbiddenPanel.tsx
```diff
@@ -2,17 +2,12 @@ import { Anchor } from '@mantine/core';
...
 import SuboptimalState from './common/SuboptimalState/SuboptimalState';
 
-const ForbiddenPanelWrapper = styled.div`
-    margin-top: 30vh;
-`;
-
 const ForbiddenPanel: FC<{ subject?: string }> = ({ subject }) => {
     return (
-        <ForbiddenPanelWrapper>
+        <div style={{ marginTop: '30vh' }}>
```
joaoviana, line 10: could we do
```suggestion
        <Box mt='30vh'>
```
?

packages/frontend/src/pages/MinimalSavedExplorer.tsx
```diff
@@ -56,11 +52,13 @@ const MinimalExplorer: FC = () => {
             colorPalette={savedChart.colorPalette}
         >
             <MantineProvider inherit theme={themeOverride}>
-                <StyledLightdashVisualization
-                    // get rid of the classNames once you remove analytics providers
-                    className="sentry-block ph-no-capture"
-                    data-testid="visualization"
-                />
+                <Box mih="inherit" h="100%">
+                    <LightdashVisualization
```
joaoviana, line 56: shouldn't this be styled instead of adding another?

joaoviana (approved):

packages/frontend/src/components/common/ErrorState/index.tsx
```diff
@@ -56,7 +49,12 @@ const ErrorState: FC<{
         }
     }, [error]);
 
-    return <StyledSuboptimalState $hasMarginTop={hasMarginTop} {...props} />;
+    return (
+        <SuboptimalState
+            style={{ marginTop: hasMarginTop ? '20px' : undefined }}
```
IrakliJani, line 54: underneath `SuboptimalState` is Mantine's Stack but it does not forward props correctly.

ideally SuboptimalState's state should be 

`interface Props extends StackProps { ... }`

and rest + sx spread


```
            {...rest}
            sx={{
              ...rest?.sx
            }}

```

packages/frontend/src/components/common/ErrorState/index.tsx
```diff
@@ -56,7 +49,12 @@ const ErrorState: FC<{
...
+            {...props}
+        />
```
IrakliJani, line 56: underneath `SuboptimalState` is Mantine's Stack but it does not forward props correctly.

ideally SuboptimalState's prop type should be 

`interface Props extends StackProps { ... }`

and rest + sx spread


```
{...rest}
sx={{
  ...rest?.sx
}}
```

magnew: Updated. I was trying not to change this thing cause it's used everywhere and its very hard to verify. But yolo. It's better to fix it, good call.

packages/frontend/src/components/ForbiddenPanel.tsx
magnew: Good call. done.

packages/frontend/src/pages/MinimalSavedExplorer.tsx
magnew: Good callout. I waffled a lot about this one. My hesitation is that `LightdashVisualization` actually returns any one of the visualization types and they all handle style props a little differently. So changing those seems a little risky. Especially since style-components was relying on className and we'd be switching to passing a style. 

So I did this since it is a very specific case: minimal explore view. I didn't want to break visualizaitons in some edge-casey way just to fix this specific view. Am I being too cautious? 😄 If you guys would like, I can dig into this more.

magnew: I'm gonna merge the less risky version of this. This one the the Box. I tested all of the chart types in minimal and it looks good. If either of you would like, I can follow up and try to clean up the style handling of the various vizes

Issue #7759, opened 2023-10-31
Metric seems to be broken based on the length of the identifier


Kobold:
### Description

I created a few metrics and they seemed to conk out even though the underlying SQL is correct. After messing around, I discovered that it seems to fail once an identifier is too long. It would be ideal if Lightdash did either or both of:

- Not fail for reasonable length identifiers.
- Warn that metric won't work, explain why, and how to fix it. (in CLI, web UI, or both).

### To Reproduce

Here's the start of the model yaml with metrics declarations. The two metrics are the same, just with different names.

```yaml
version: 2
models:
  - name: lightdash__historical_analysis_date_range
    meta:
      label: Historical Analysis for Advisors Explore
      metrics:
        ...
        count_advisor_history_user_ids:
          type: count_distinct
          sql: ${lightdash__historical_advisor.history_user_id}
        cahu:
          type: count_distinct
          sql: ${lightdash__historical_advisor.history_user_id}
```

Here's what shows up in the web UI:

<img width="1427" alt="image" src="https://github.com/lightdash/lightdash/assets/36694/866d8c6c-1d13-4e76-aaf8-0c325de6d357">

The actual underlying SQL query runs just fine:

_Edit: this was accidentally the same screenshot as above. That query above when run in the SQL runner produces the same value (2,751) for both columns._

### App Version

Web UI: 0.832.2
CLI: 0.837.0

owlas:
This issue was mentioned by a user in slack: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1710719110166299?thread_ts=1710719110.166299&cid=C03MG2VTR08

owlas:
This issue was mentioned by a user in slack: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1706184281214839?thread_ts=1706184281.214839&cid=C03MG2VTR08

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C05UPNC0LGP/p1711484380885419?thread_ts=1711484380.885419&cid=C05UPNC0LGP

rephus:
This issue seems to affect postgres mainly , other warehouses have bigger length,

postgres: 64 chars
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
bigquery: 300 chars 
https://cloud.google.com/bigquery/quotas#dataset_limits
snowflake: 255 chars

Tested in postgres: 
[...]

PR #9592, opened 2024-04-01
fix: custom metric long name

rephus:
Closes: https://github.com/lightdash/lightdash/issues/7759

### Description:

This PR fixes the issue with custom metrics, where they are too long , 
this fix should be backwards compatible and straightforward, we are just making a shorter name for the custom metric. 

it doesn't fix the max column name on dbt, that will require a bigger fix

I'm using 64 as the max limit because that's the lowest limit (from postgres) , other database have higher limits.  

Before and after
Screenshot from 2024-04-01 17-56-48

### Reviewer actions


almeidabbm (approved):

rephus:
Unrelated e2e error

Issue #9482, opened 2024-03-21
There should be an option to export results from a dashboard as .csv (not just image)


TuringLovesDeathMetal:
### Description

Right now, there's no way to export the results from a dashboard as .csv files. Instead, I need to set up a scheduled delivery for this and `send now`. 

This should be an option when I hit `export dashboard`

When I open `export dashboard` there should be an option to switch betwen `csv` or `image` and the modal should update with the correct configurations (just like in the scheduled deliveries) 

<img width="375" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/363cbc04-895d-43f6-95ba-f7a10b6e636e">

### Analytics event

we should update the export dashboard event to include the file type of the export.

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C053ESBT36V/p1711101285332779

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C053ESBT36V/p1711101285332779?thread_ts=1711101285.332779&cid=C053ESBT36V

rephus:
@TuringLovesDeathMetal  What does exporting a CSV mean on a dashboard ? is it just a `zip` with the results of all charts ?

TuringLovesDeathMetal:
^Sorry, I really didn't make that clear. Yes, that's exactly it. 

Let's make this even more limited as well and say that there's only the option to `export the results in table` to .csv files. 

So you'd click `csv` and then there would be some copy saying `export results in table for all charts` (maybe it would be on hover over an `export` button?)

<img width="166" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/40d79962-6e39-4875-81e2-f2d1a2e68713">


There's no option to set the custom limit/all results/etc. at the start!

PR #9604, opened 2024-04-02
feat: export dashboard csv

rephus:
Closes: #9482

### Description:

[Screencast from 03-04-24 11:39:59.webm](https://github.com/lightdash/lightdash/assets/1983672/6a31d901-1305-41b0-be2c-ceba0d513fbc)

### Reviewer actions


packages/frontend/src/components/common/modal/DashboardExportModal.tsx
```diff
@@ -1,28 +1,70 @@
...
 
-export const DashboardExportModal: FC<Props & ModalProps> = ({
-    opened,
+type CsvExportProps = {
+    dashboard: Dashboard;
+};
+
+const CsvExport: FC<CsvExportProps & ModalProps> = ({ dashboard, onClose }) => {
+    const exportCsvDashboardMutation = useExportCsvDashboard();
+    const dashboardFilters: SchedulerFilterRule[] | undefined = [];
```
rephus, line 35: I still need to get the `current` dashboard filters to send it to the backend, in the image we use the location, but we can't use a plain string here, any ideas @joaoviana  ?

packages/backend/src/services/CsvService/CsvService.ts
```diff
@@ -816,4 +816,53 @@ export class CsvService {
             throw e;
         }
     }
+
+    async exportCsvDashboard(
+        user: SessionUser,
+        dashboardUuid: string,
+        queryFilters: SchedulerFilterRule[] | undefined,
```
rephus, line 823: I copied this from the `getCsvsForDashboard` ,b ut we can change it if needed

rephus:
This method runs in the API , not in scheduler, this should be ok, as long as we only get results from the `table` and not `all`, if we change this option we'll need to move it to the scheduler to avoid blocking the API

TuringLovesDeathMetal (approved):
Nice! looks great :) 

Just double checking: have we updated the analytics event for dashbaord downloads to include `csv` as a type? Also, can we update the docs please 🙏

joaoviana (changes requested):
Looking great! 
Few requested changes 👍

packages/frontend/src/components/common/modal/DashboardExportModal.tsx
```diff
@@ -84,43 +172,39 @@ export const DashboardExportModal: FC<Props & ModalProps> = ({
...
+                    w="50%"
+                    mb="xs"
+                    defaultValue="image"
+                    onChange={setExportType}
+                />
+                {exportType === 'csv' && (
+                    <CsvExport
+                        dashboard={dashboard}
+                        onClose={onClose}
+                        opened={true}
```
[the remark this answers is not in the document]
rephus: mmm, we do if we take `& ModalProps` in the component

rephus:
@joaoviana thanks, that worked, 
Screenshot from 2024-04-04 08-56-37

Before:

Screenshot from 2024-04-04 08-56-08

After:

Screenshot from 2024-04-04 09-04-08

joaoviana (approved):
🎉

PR #9607, opened 2024-04-02
fix: reference line accordion item index

joaoviana:
Closes: N/A

### Description:

Use `index` instead of referenceLine's `data.value` to fix this issue: https://lightdash.sentry.io/issues/5131175770/?project=5959292&referrer=slack

`Accordion.Item`s will break if their `value` attribute value is not correct.

### Reviewer actions


magnew:
> Shi it

I like this typo

Issue #9545, opened 2024-03-27
Allow private key auth for OIDC to Azure


owlas:
### Description

OIDC can use client secret auth for OAuth2 authentication or it can use a private/public key pair. The latter is an improvement because private secrets don't need to be exchanged between the lightdash server and the lightdash admin.

Resources:
- More info on OAuth2 client authentication options: https://oauth.net/2/client-authentication/
- Detailed explanation of auth options: https://cloudentity.com/developers/basics/oauth-client-authentication/client-authentication-overview/
- Credentials for Azure AD Single sign on: https://learn.microsoft.com/en-us/entra/identity-platform/security-best-practices-for-app-registration
- Azure AD specific docs for `private_key_jwt` OIDC authentication, specifies we should use `rs256` for the signing algorithm (the most widely adopted, but slow): https://learn.microsoft.com/en-us/entra/identity-platform/certificate-credentials

Behaviour for the node OIDC client can be configured with `token_endpoint_auth_method` which is default "client secret". We should use `private_key_jwt: https://github.com/panva/node-openid-client/blob/main/docs/README.md#client-authentication-methods.

It would be useful to have a script to generate a set of credentials (this would eventually live in Lightdash) - Azure accepts certificates with .cer, .pem, and .cert formats, usually .pem is used for a self signed public key. Script should generate an assymetric RSA keypair (RSA required for `alg: RS256`) and the public and private key in `.pem` format 

Currently we use passport for azure but I'm not sure if it's supports either of these methods, so may need to refactor to use the OIDC client like Okta (which is actually preferable, since it's more transparent on how we're doing PKCE / state checks).

PR #9612, opened 2024-04-02
feat: AzureAd OIDC `private_key_jwt` token authentication

filp:
Closes: #9545

### Description:

- Adds support for OIDC `private_key_jwt`, currently for AzureAd only (but easily migrated to others and/or generic OIDC)
- Adds configuration options for the above

See inline comments for additional details.

### Reviewer actions


packages/backend/src/App.ts
```diff
@@ -183,15 +184,15 @@ export default class App {
             App.initNodeProcessMonitor();
         }
 
-        const expressApp = this.initExpress();
+        const expressApp = await this.initExpress();
```
filp, line 187: Issuer discovery requires an async call at setup time, if configured to use `private_key_jwt` for Azure.

packages/backend/src/controllers/authentication.ts
```diff
@@ -95,7 +103,15 @@ const createOpenIdUserFromUserInfo = (
     issuerType: OpenIdIdentityIssuerType,
     fail: OpenIDClientOktaStrategy['fail'],
 ) => {
-    if (!userInfo.email || !userInfo.sub) {
+    /**
```
filp, line 106: One of many quirks to account for when using `openid-client` + its built-in strategy + azuread

packages/backend/src/controllers/authentication.ts
```diff
@@ -127,14 +143,27 @@ const createOpenIdUserFromUserInfo = (
     return openIdUser;
 };
 
+/**
+ * Not all possible values are reflected in the underlying PassportProfile type,
+ * at least for AzureAD + openid-client
+ */
+type PassportProfileLike = PassportProfile &
+    Partial<{
+        email: string;
```
filp, line 152: I'm 99% sure this is openid-client's strategy misbehaving, but opted to not chase it further for now.

packages/backend/src/controllers/authentication.ts
```diff
@@ -190,7 +221,7 @@ const setupClient = async () => {
 export class OpenIDClientOktaStrategy extends Strategy {
     async authenticate(req: Request) {
         try {
-            const client = await setupClient();
+            const client = await setupOktaIssuerClient();
```
filp, line 224: Renamed method to be less generic.

packages/backend/src/controllers/authentication.ts
```diff
@@ -467,7 +499,7 @@ const genericOidcHandler =
 
             const openIdUser = createOpenIdUserFromProfile(
                 profile,
-                issuer,
+                issuerOverride || issuer,
```
filp, line 502: `openid-client` returns a `TokenSet` as the issuer (instead of a string), which breaks some of the checks downstream (e.g `UserService` checks if something is an openId user, including by checking if issuer is a string.

packages/backend/src/controllers/authentication.ts
```diff
@@ -499,28 +531,193 @@ export const isOktaPassportStrategyAvailableToUse = !!(
...
+            usePKCE: true,
+            passReqToCallback: true,
+            params: {
+                redirect_uri: new URL(
+                    `/api/v1${azuread.callbackPath}`,
+                    lightdashConfig.siteUrl,
+                ).href,
+            },
+            extras: {
+                clientAssertionPayload: {
```
filp, line 649: More Azure quirks

packages/backend/src/controllers/authentication.ts
```diff
@@ -499,28 +531,193 @@ export const isOktaPassportStrategyAvailableToUse = !!(
...
+            OpenIdIdentityIssuerType.AZUREAD,
+            new URL(azuread.openIdConnectMetadataEndpoint!).origin,
+        ) as unknown as StrategyVerifyCallback<unknown>,
+    );
+};
+
+/**
+ * Creates the appropriate AzureAd passport strategy based on the available configuration.
+ */
+export const createAzureAdPassportStrategy = () => {
```
filp, line 676: To avoid unintended issues, I kept both strategies around, and only use the new strategy if using `private_key_jwt`. In theory they should behave the same, and we can migrate other issuers to also use `openid-client`.

packages/backend/src/models/SshKeyPairModel.ts
```diff
@@ -1,45 +1,13 @@
-import { SshKeyPair } from '@lightdash/common';
-import { generateKeyPair } from 'crypto';
+import { type SshKeyPair } from '@lightdash/common';
 import { Knex } from 'knex';
-import { parseKey } from 'sshpk';
-import { EncryptionService } from '../services/EncryptionService/EncryptionService';
+import { type EncryptionService } from '../services/EncryptionService/EncryptionService';
```
filp, line 3: Minor unrelated tweaks, was digging around this code early on.

packages/backend/src/controllers/authentication.ts
ZeRego: But this method is only used in the okta authentication.

packages/backend/src/controllers/authentication.ts
ZeRego: Do we need to update the type in `packages/backend/src/@types/passport-openidconnect.d.ts`?

packages/backend/src/controllers/authentication.ts
```diff
@@ -499,28 +532,174 @@ export const isOktaPassportStrategyAvailableToUse = !!(
...
+const loadJWKFromConfig = (jwkConfigValue: string): Record<string, unknown> => {
+    try {
+        return JSON.parse(jwkConfigValue);
+    } catch (err) {
+        throw new Error(
+            `Failed to parse JWK from configuration - must be valid JSON`,
+            err,
+        );
+    }
+};
```
ZeRego, line 564: These functions are not used atm.

packages/backend/src/controllers/authentication.ts
```diff
@@ -499,28 +532,174 @@ export const isOktaPassportStrategyAvailableToUse = !!(
...
+
+/**
+ * Azure-specific (but also mostly generic) implementation of private_key_jks
+ * token authentication, using openid-client.
+ *
+ * Requires its own strategy since this method is not supported by OpenIdConnect.
+ */
+export const azureAdPrivateKeyJksStrategy = async (): Promise<
```
ZeRego, line 572:
```suggestion
const azureAdPrivateKeyJksStrategy = async (): Promise<
```

packages/backend/src/controllers/authentication.ts
filp: Good catch, I refactored the configuration process late at night - will clean this up.

packages/backend/src/config/parseConfig.ts
```diff
@@ -359,6 +388,13 @@ const mergeWithEnvironment = (config: LightdashConfigIn): LightdashConfig => {
                 oauth2TenantId: process.env.AUTH_AZURE_AD_OAUTH_TENANT_ID,
                 callbackPath: '/oauth/redirect/azuread',
                 loginPath: '/login/azuread',
+                x509PublicKeyCertPath: process.env.AUTH_AZURE_AD_X509_CERT_PATH,
```
ZeRego, line 391: We should update our azure docs with the new env vars
- https://docs.lightdash.com/self-host/customize-deployment/use-sso-login-for-self-hosted-lightdash#azure-active-directory
- https://docs.lightdash.com/self-host/customize-deployment/environment-variables

docs/docs/references/roles.mdx
```diff
@@ -68,16 +68,17 @@ Users with `Full Access` to a space can restrict and increase a user’s inherit
...
 - an interactive viewer who is given `Can edit` space permissions can save content in that space but not in any other space (unless given `edit` access to another space)
 - an editor who is given `Can view` space permissions cannot edit the content in that space, but they can create/edit content in other parts of the project (unless they're given `Can view` access).
 
-| Action                   | Full Access  | Can Edit | Can View |
-| :----------------------- | :----------: | :------: | :------: |
-| View space content       |      ✅      |     ✅    |    ✅    |
-| Manage space content     |      ✅      |     ✅    |    ❌    |
-| Manage space access      |      ✅      |     ❌    |    ❌    |
-| Manage space details     |      ✅      |     ❌    |    ❌    |
+| Action               | Full Access | Can Edit | Can View |
```
filp, line 76: Automated formatting while updating docs.

packages/backend/src/config/parseConfig.ts
filp: Done - there's some additional work here as far as outlining the `private_key_jwt` flow setup, but will hold off on that until we've had more time to test and gather any points to incorporate into the docs.

packages/backend/src/controllers/authentication.ts
filp: Yep - that wasn't the case for a while, will revert these changes to avoid confusion.

packages/backend/src/controllers/authentication.ts
filp: Didn't realize we managed our own type implementation -- I'll move these changes there!

PR #9630, opened 2024-04-03
feat: table chart config - conditional formatting (3/3)

joaoviana:
Relates to: #9568 

### Description:

Table config changes to `Conditional Formatting` tab

Single colour
image

Range
image

- Inputs are smaller
- Color selects are different depending on mode
- It's an accordion now
- Different UI when adding more and more conditions

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/common/hooks/useControlledAccordion.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/common/hooks/useControlledAccordion.tsx
@@ -1,0 +1,21 @@
+import { useCallback, useState } from 'react';
+
+export const useControlledAccordion = (defaultOpenItems = []) => {
+    const [openItems, setOpenItems] = useState<string[]>(defaultOpenItems);
+
+    const handleAccordionChange = useCallback((itemValues: string[]) => {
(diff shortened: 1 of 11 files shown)


joaoviana:
image
looks better now: https://github.com/lightdash/lightdash/pull/9630/commits/c0a3291d3bb8c1f2a25298f406a18c3b30733a88

packages/frontend/src/components/VisualizationConfigs/common/hooks/useControlledAccordion.tsx
```diff
@@ -0,0 +1,21 @@
+import { useCallback, useState } from 'react';
+
+export const useControlledAccordion = (defaultOpenItems = []) => {
```
magnew, line 3: This is an interesting approach. To have a hook instead of a wrapper or something. But I think I like it.

magnew (approved):
Looks great and it's a much improved UX. 

One thing I noticed that isn't great is that there is a really long delay when clicking this button the first time. Over a second. 

Screenshot 2024-04-08 at 11 57 04

I didn't debug it deeply, but I wonder if there is something we can do about it

Issue #9615, opened 2024-04-03
dashboard timestamp filters show timezones in the filter, but don't use timezones in the query.


TuringLovesDeathMetal:
**Note:** I _think_ this might be an issue where we're treating a timestamp as a `timestamptz` in the filter, but as a `timestamp` in the query 👀 

**Today**:

- If I add a timestamp filter to a dashboard, then it shows up with my timezone on it (e.g. I’m in Paris timezone, so if I filter on 2PM, Feb 29th, it will show up as `2024-02-29, 14:00:00 (+2:00)` on my dashboard filter).
- This makes it seem like the timestamp filter is using my timezone.
- But, if I open the SQL for the filter, it’s actually using UTC time in the filter, not my timezone. E.g. from above, I’d expect the filter to be `2024-02-29, 12:00:00 UTC`, since it’s `+2:00 UTC`. But, instead, the SQL just shows `2024-02-29, 14:00:00`. 

![image](https://github.com/lightdash/lightdash/assets/31848148/3c99073e-0016-4393-acc8-071ec80698f8)
![image](https://github.com/lightdash/lightdash/assets/31848148/8bd5de23-f878-49ac-a2de-55392ea38775)

**What we’d expect:**

- In dashboard filters, if I add a timezone filter, it should match the time being filtered in the SQL (this is a bug!). E.g. from above, I’m in Paris timezone, so if I filter on 2PM, Feb 29th, it should show up as 2024-02-29, 14:00:00 on my dashboard filter
- Let’s keep it as UTC everywhere for now. At least this is consistent. There’s a bigger piece of work around setting local timezones here: https://github.com/lightdash/lightdash/milestone/97

PR #9662, opened 2024-04-05
fix: time filter using the wrong timezone

magnew:
Closes: #9615

### Description:

Fix inconsistencies in time filters related to timezones. I am posting this as a draft. There are still a couple UX questions and I need to figure out a test plan. I will write out more of an explanation when its ready. 

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -1,3 +1,4 @@
+import dayjs from 'dayjs';
 import moment from 'moment';
 import { v4 as uuidv4 } from 'uuid';
 import {
@@ -187,9 +188,10 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
                     const valueIsDate =
                         value !== undefined && typeof value !== 'number';
 
+                    // NOTE: Using .format() makes this a standard ISO string
                     const timestampValue = valueIsDate
-                        ? moment(value).format('YYYY-MM-DDTHH:mm:ssZ')
-                        : moment().utc(true).format('YYYY-MM-DDTHH:mm:ssZ');
+                        ? dayjs(value).format()
+                        : dayjs().format();
 
                     filterRuleDefaults.values = [timestampValue];
                 } else {
--- a/packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
@@ -170,24 +170,13 @@ const DateFilterInputs = <T extends ConditionalRule = DateFilterRule>(
             }
 
             if (isTimestamp) {
-                const parsedTimestampToUtc = rule.values
-                    ? dayjs(
-                          parseTimestamp(
-                              formatTimestamp(
-                                  rule.values[0],
-                                  TimeFrames.MILLISECOND,
-                              ),
-                              TimeFrames.MILLISECOND,
-                          ),
-                      ).utc()
-                    : null;
+                // For display only
+                const displayFormat = 'YYYY-MM-DD HH:mm:ss (Z)';
 
-                const valueFormat =
-                    parsedTimestampToUtc?.format('YYYY-MM-DD HH:mm:ss') ??
-                    undefined;
-
-                // NOTE: This ensures that valueFormat and value are always in sync with each other and that there's no mismatch between the input value and the calendar + time input values set
-                const value = dayjs(valueFormat).toDate() ?? null;
+                let value =
+                    rule.values && rule.values[0]
+                        ? dayjs(rule?.values?.[0]).toDate()
+                        : dayjs().toDate(); // Create
 
                 return (
                     <FilterDateTimePicker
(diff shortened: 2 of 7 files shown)


packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
```diff
@@ -170,24 +170,13 @@ const DateFilterInputs = <T extends ConditionalRule = DateFilterRule>(
...
-                              formatTimestamp(
-                                  rule.values[0],
-                                  TimeFrames.MILLISECOND,
-                              ),
-                              TimeFrames.MILLISECOND,
-                          ),
-                      ).utc()
-                    : null;
+                // For display only
+                const displayFormat = 'YYYY-MM-DD HH:mm:ss (Z)';
```
joaoviana, line 174: just checking that this change hasn't affected this one: https://github.com/lightdash/lightdash/issues/8942 with the PR that fixed it here: https://github.com/lightdash/lightdash/pull/9303

joaoviana: After chatting with @magnew the approach done here: https://github.com/lightdash/lightdash/pull/9303 is not recommended and didn't fully solve the issue. Thanks for demo'ing the problem!

magnew, line 174: Note that this type of (non-standard) format is dangerous. If you try to parse a string with this format with moment of dayjs, you will not get a correct time because the timezone will be dropped. 

I kept it anyway for compatibility, but it is only for view-time.

packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
```diff
@@ -170,24 +170,13 @@ const DateFilterInputs = <T extends ConditionalRule = DateFilterRule>(
...
             if (isTimestamp) {
-                const parsedTimestampToUtc = rule.values
-                    ? dayjs(
-                          parseTimestamp(
-                              formatTimestamp(
-                                  rule.values[0],
-                                  TimeFrames.MILLISECOND,
-                              ),
-                              TimeFrames.MILLISECOND,
-                          ),
```
magnew, line 181: This parse timestamp was returning a non-standard string, then the .utc() call on the next line was causing the TZ to be dropped and this time was not correct.

packages/frontend/src/components/common/Filters/FilterInputs/DateFilterInputs.tsx
```diff
@@ -206,15 +195,8 @@ const DateFilterInputs = <T extends ConditionalRule = DateFilterRule>(
...
-                                    v === null
-                                        ? []
-                                        : [
-                                              formatTimestamp(
-                                                  v,
-                                                  TimeFrames.SECOND,
-                                              ),
-                                          ],
+                                // format as an ISO string, not for display
+                                values: v === null ? [] : [dayjs(v).format()],
```
magnew, line 199: We should *always* pass time around either as a Date object or the ISO string you get from format().

rephus (commented):
most of the cypress tests for dates `dates.cy.ts` are disabled (not really sure why, we should fix/re-enable at some point) , perhaps we could run them (locally) to see if these changes are breaking something ?

magnew:
> most of the cypress tests for dates `dates.cy.ts` are disabled (not really sure why, we should fix/re-enable at some point) , perhaps we could run them (locally) to see if these changes are breaking something ?

Good call. I didn't notice that. I can run them locally, but also work on fixing them.

rephus:
Inconsistent format (`-` and `/` ) 
image

image

rephus (approved):
I tested everything I can think of, in different timezones, I couldn't find anything wrong , everything works as expected,a nd uses the right timezone !

packages/common/src/utils/filters.ts
```diff
@@ -187,9 +188,10 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
                     const valueIsDate =
                         value !== undefined && typeof value !== 'number';
 
+                    // NOTE: Using .format() makes this a standard ISO string
                     const timestampValue = valueIsDate
-                        ? moment(value).format('YYYY-MM-DDTHH:mm:ssZ')
-                        : moment().utc(true).format('YYYY-MM-DDTHH:mm:ssZ');
```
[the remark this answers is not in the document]
magnew: This change makes it so that this input is initialized with a local time value instead of a UTC value. This is a change we are making, but it's intentional.

magnew:
> Can't we update the times without using the timepicker ? I'd like to edit as text . I guess this is a mantine thing

I agree with you, but I think this is a limitation of the mantine input :/

magnew:
> most of the cypress tests for dates `dates.cy.ts` are disabled (not really sure why, we should fix/re-enable at some point) , perhaps we could run them (locally) to see if these changes are breaking something ?

A lot of them use blueprint. But they are a good source of what should be right. I will work on updating them, but I'll put it up separately.

Issue #9468, opened 2024-03-20
`/charts` endpoint doesn't return dashboard name


ZeRego:
### Description

`/charts` has dashboardName but it returns empty for every chart

owlas:
This issue was mentioned by a user in slack: https://lightdash.slack.com/archives/C04KD95TF0U/p1709757760828279?thread_ts=1709757760.828279&cid=C04KD95TF0U

ZeRego:
Note that this is only available for charts that belong to a dashboard. (aka when saved_queries.dashboard_uuid is populated)

magnew:
It looks like the previous PRs here were closed because there were setting the singular dashboard name, but that is only for charts in dashboards. Is there any reason we couldn't solve the scenario by adding a list of dashboards a chart is used in? It could make this endpoint a little slower, but it's not especially heavily used. Or including all of the dashboards could be an option.

PR #9674, opened 2024-04-08
fix: dashboardName missing for dashboard charts in charts endpoint

almeidabbm:
Closes: [#9468](https://github.com/lightdash/lightdash/issues/9468)

### Description:

- Returns dashboard name for dashboard charts on `/charts` endpoint

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -165,10 +165,21 @@ export class SpaceModel {
         const space = await this.getFirstAccessibleSpace(projectUuid, userUuid);
         const savedQueries = await this.database('saved_queries')
             .leftJoin(
-                SpaceTableName,
-                `saved_queries.space_id`,
-                `${SpaceTableName}.space_id`,
+                DashboardsTableName,
+                `${DashboardsTableName}.dashboard_uuid`,
+                `${SavedChartsTableName}.dashboard_uuid`,
             )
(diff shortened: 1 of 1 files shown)


rephus:
We had a discussion the other day with @filp and @ZeRego about this, the /charts is only used in `spaces` and this `Dashboard name` should only be populated with `charts within dashboards` so it makes sense it is undefined. Not sure what we're trying to fix here exactly

packages/backend/src/models/SpaceModel.ts
```diff
@@ -165,10 +165,21 @@ export class SpaceModel {
...
         const savedQueries = await this.database('saved_queries')
             .leftJoin(
-                SpaceTableName,
-                `saved_queries.space_id`,
-                `${SpaceTableName}.space_id`,
+                DashboardsTableName,
+                `${DashboardsTableName}.dashboard_uuid`,
+                `${SavedChartsTableName}.dashboard_uuid`,
             )
+            .leftJoin(SpaceTableName, function spaceJoin() {
```
rephus, line 172: what happens if the same chat is used in multiple dashboards ?

almeidabbm:
Closing as what @rephus said makes sense, this endpoint is only used in `spaces` and therefore it shouldn't return `charts` within `dashboards`. `dashboardName` is not present for `charts` saved in `spaces`

Issue #9660, opened 2024-04-05
When a new series in a group is generated, the colour resets


rephus:
### Description

I have a chart that generates different series every day. I updated the color, but every day the color resets. 

Proposal: Store a  metadata mapping somewhere that contains the `serie_name:color` . When we generate a serie, we check this metadata and use the right color `based on the name of the serie` 

UI should not change. This means when the user manually select a color for a serie, this color is applied on the serie and stored in this metadata at the same time.

Make this metadata reusable: this solution should also help with other property issues we have like `line styles`  , not just `color`, etc. So we fix colors in this issue, but we need to build a metadata structure that will allow storing different values. 

Thread: https://app.usepylon.com/issues?conversationID=967d95dd-ca43-4e30-882f-649b70df319a

Related issue: https://github.com/lightdash/lightdash/issues/4166

### To Reproduce

The easiest way to replicate is to add a filter that removes the series and add it back. 

[Screencast from 08-04-24 10:03:14.webm](https://github.com/lightdash/lightdash/assets/1983672/c3d8855e-9baa-4f2a-a3fe-db1b156677a3)

PR #9681, opened 2024-04-08
fix: series color change

rephus:
Closes: #9660

### Description:
The goal, store the mapping of colors in a `metadata` in chart config, so we can persist this info and load the right color

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
+++ b/packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
@@ -6,6 +6,7 @@ import {
     type ChartConfig,
     type DashboardFilters,
     type ItemsMap,
+    type SeriesMetadata,
     type TableCalculationMetadata,
 } from '@lightdash/common';
 import type EChartsReact from 'echarts-for-react';
@@ -77,7 +78,8 @@ type VisualizationContext = {
 
     getSeriesColor: (seriesLike: SeriesLike) => string;
     getGroupColor: (groupPrefix: string, groupName: string) => string;
-
+    setMetadata: (metadata: Record<string, SeriesMetadata>) => void;
+    metadata?: Record<string, SeriesMetadata>;
     colorPalette: string[];
 };
 
@@ -152,6 +154,7 @@ const VisualizationProvider: FC<React.PropsWithChildren<Props>> = ({
 
     const chartRef = useRef<EChartsReact>(null);
 
+    const [metadata, setMetadata] = useState<Record<string, SeriesMetadata>>();
     const [lastValidResultsData, setLastValidResultsData] =
         useState<ApiQueryResults>();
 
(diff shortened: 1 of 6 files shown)


filp:
I think this approach will be a bit of a headache to deal with. I'd probably consider either...

- Saving the metadata mapping as part of the visualization provider, instead of hooking into the color selector directly
- Or hooking into `useChartColorConfig` and allow flushing the mappings stored by it - we'd also need a way to load these mappings back into the color mapping instance, or include it as a condition for `getSeriesColor`, e.g:

```ts
onChartConfigChange(() => {
  const mappedColors = /* build a map of series colors mappings using getSeriesColor */
  const newMetadata = /* merge existing metadata with the series colors */

  // ...

  return newConfig
));
```

And then factoring it in during color assignment:

```ts
getSeriesColor = () => {
  // ...
  if (metadata[seriesId]) return metadata[seriesId];
  if (series.color) return series.color;

  // ... existing behavior
}
```

This of course being super pseudo-code.

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SingleSeriesConfiguration.tsx
```diff
@@ -93,6 +97,24 @@ const SingleSeriesConfiguration: FC<Props> = ({
...
                                     color,
                                 });
+                                const serieId =
+                                    calculateSeriesLikeIdentifier(series).join(
+                                        '.',
+                                    );
+
+                                if (
+                                    visualizationConfig.chartType ===
+                                    ChartType.CARTESIAN
```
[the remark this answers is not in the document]
rephus: I tested with pie charts, and they already work 
[Screencast from 17-04-24 10:04:49.webm](https://github.com/lightdash/lightdash/assets/1983672/98460a34-dd1e-49b0-bd31-aac88789d94a)

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -75,9 +76,11 @@ type VisualizationContext = {
     setChartType: (value: ChartType) => void;
     setPivotDimensions: (value: string[] | undefined) => void;
 
-    getSeriesColor: (seriesLike: SeriesLike) => string;
+    getSeriesColor: (
+        seriesLike: SeriesLike,
+        metadata?: Record<string, SeriesMetadata>,
+    ) => string;
```
almeidabbm, line 82:
```suggestion
    getSeriesColor: (
        seriesLike: SeriesLike,
    ) => string;
```

I haven't seen this argument getting used/passed anywhere - also the function on line 255 doesn't seem to take it into account

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
```diff
@@ -6,6 +6,7 @@ import {
     type ChartConfig,
     type DashboardFilters,
     type ItemsMap,
+    type SeriesMetadata,
```
almeidabbm, line 9: If we remove the `getSeriesColor` argument, this won't be needed
```suggestion
```

almeidabbm (approved):
From my tests locally it worked super well!! 🚀 
Just a small nitpick, ignore if you don't think it should be applied

packages/frontend/src/components/LightdashVisualization/VisualizationProvider.tsx
rephus: yes, you're right, good catch. This was probably a leftover for my tests

almeidabbm (approved):

Issue #9685, opened 2024-04-09
2 step login form breaks if I'm using sync with a different google account


rephus:
### Description

![image](https://github.com/lightdash/lightdash/assets/1983672/6b942479-5fc3-458c-ba04-9389320f8beb)
or redirect automatically to google, but I still want to use user/password

Related https://github.com/lightdash/lightdash/issues/9643 ?

Thread: https://lightdash.slack.com/archives/C03MCQ08UKS/p1712649021480559?thread_ts=1712586360.979469&cid=C03MCQ08UKS

### To Reproduce
- Have a lightdash account
- Use `syncs` -> login with google : This will create an openid
- Logout and try to login with your password
- Now you're force to login with google

PR #9686, opened 2024-04-09
fix: fix used email error when using 2 page login and sync

rephus:
Closes: #9685

### Description:

Before:
Screenshot from 2024-04-09 09-51-51

After:

Screenshot from 2024-04-09 09-56-36

### Reviewer actions


rephus:
We know about this bug, but we are not sure if this is the right way to fix it, or we should simply allow using the sync to also authenticate. Closing this for now until we get more feedback.

PR #9687, opened 2024-04-09
fix: only use enabled openid issuers

rephus:
Closes: 

### Description:

Question: Is it possible to have 2 existing open_id types linked to the same primary email ? if so we should do the filter in the getOPenIdIssuer model method, or return all (instead of first) and filter them later

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/UserModel.ts
+++ b/packages/backend/src/models/UserModel.ts
@@ -864,20 +864,17 @@ export class UserModel {
         return row.refresh_token;
     }
 
-    async getOpenIdIssuer(
-        email: string,
-    ): Promise<OpenIdIdentityIssuerType | undefined | null> {
-        const row = await this.database('emails')
+    async getOpenIdIssuers(email: string): Promise<OpenIdIdentityIssuerType[]> {
+        const rows = await this.database('emails')
             .leftJoin(
                 'openid_identities',
                 'emails.user_id',
                 'openid_identities.user_id',
             )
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/UserService.ts
```diff
@@ -1318,8 +1318,12 @@ export class UserService extends BaseService {
         ].filter(Boolean) as OpenIdIdentityIssuerType[];
```
ZeRego, line 1318: Just noticed that `GENERIC_OIDC` is not accounted for in this array.

ZeRego (approved):

packages/backend/src/services/UserService.ts
rephus: good catch, I added it here

packages/backend/src/models/UserModel.ts
```diff
@@ -864,20 +864,17 @@ export class UserModel {
         return row.refresh_token;
     }
 
-    async getOpenIdIssuer(
-        email: string,
-    ): Promise<OpenIdIdentityIssuerType | undefined | null> {
-        const row = await this.database('emails')
+    async getOpenIdIssuers(email: string): Promise<OpenIdIdentityIssuerType[]> {
```
rephus, line 867: now we return a list of issuers, and we check which ones are disabled on the service

owlas (approved):

Issue #9667, opened 2024-04-08
If I have a table calculation filter in a group, then I delete the table calculation, it doesn't delete the filter (causing a validation error that I can't fix)


TuringLovesDeathMetal:
### Description

If I have a table calculation filter with any other dimension filters in Lightdash, then it creates a separate group in the filters. 

If I have a table calculation filter in a separate group, then I delete the table calculation, the filter doesn't appear greyed out/broken, it just disappears (compare this to the second half of teh video, where there is no group wtih the table calculation filter and i'm givent he option to delete it). 

The problem with this is that it triggers validation errors (because I'm filtering on a field that no longer exists), but there's no way of fixing these errors (because the filter doesn't appear greyed out/able to be deleted). 

https://www.loom.com/share/bdc209ff780b43819407e3c4cf2d3f39

PR #9696, opened 2024-04-09
fix: get invalid filters and allow deletion even in groups

joaoviana:
Closes: #9667

### Description:

The issue: 

- When _invalid_ filters (no longer available) are in `Filters` card, we display a message `tried to reference...` 
- This doesn't apply when there's a Filter group - the component doesn't handle it at the moment (see `FilterGroupForm`)

The solution for this is to: 
- Get valid and invalid filter rules in the parent: `FiltersForm` (see Filters/index.ts)
- Improve: `getFilterRulesByFieldType` to return also invalid filters
- Use the invalid filters and render the same component in `FiltersForm` - the parent of  `FilterRuleForm`

Demo: 
https://github.com/lightdash/lightdash/assets/7611706/e815e51f-11d9-4df4-89c9-1bf116ead35a

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
+++ b/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
@@ -119,6 +119,7 @@ const FiltersCard: FC = memo(() => {
                 <>
                     {totalActiveFilters > 0 && !filterIsOpen ? (
                         <Tooltip
+                            variant="xs"
                             arrowOffset={12}
                             label={
                                 <div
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
```diff
@@ -119,6 +119,7 @@ const FiltersCard: FC = memo(() => {
                 <>
                     {totalActiveFilters > 0 && !filterIsOpen ? (
                         <Tooltip
+                            variant="xs"
```
joaoviana, line 122: just looks nicer: 
Screenshot 2024-04-09 at 14 36 48

Issue #9655, opened 2024-04-05
Improve dashboard responsiveness on mobile view


rephus:
### Description

On mobile view, grid elements may be positioned awkwardly within the constraints of the responsive dashboard.

![Image](https://github.com/lightdash/lightdash/assets/382538/e363638e-a867-48d4-b8f6-579b2d7461be)

Grid elements should be more aggressively positioned on a simple vertical stack.

hamzahc1:
Let me know if you make headway with this @filp

PR #9699, opened 2024-04-09
fix: force chart tiles into a vertical stack in mobile minimal view

filp:
Closes: #9655

### Description:

- Introduces a new feature flag (`DashboardMobileVerticalStack`)
- Refactors `getResponsiveGridLayoutProps` to accept options based on the feature flag above
- If the feature flag is enabled, stacks tiles vertically (single column) on the smallest (mobile) breakpoint.

Before:
image

After (same dashboard):

Kapture 2024-04-09 at 16 49 32

### Reviewer actions

### DIFF
--- a/packages/common/src/types/featureFlags.ts
+++ b/packages/common/src/types/featureFlags.ts
@@ -36,4 +36,10 @@ export enum FeatureFlags {
 
     /* Show user groups */
     UserGroupsEnabled = 'user-groups-enabled',
+
+    /** Collapse dashboard into a simple vertical stack on mobile */
+    DashboardMobileVerticalStack = 'dashboard-mobile-vertical-stack',
+
+    /** */
+    LazyLoadDashboardTiles = 'lazy-load-dashboard-tiles',
 }
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -1,6 +1,7 @@
 import {
     assertUnreachable,
     DashboardTileTypes,
+    FeatureFlags,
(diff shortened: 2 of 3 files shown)


packages/common/src/types/featureFlags.ts
```diff
@@ -36,4 +36,10 @@ export enum FeatureFlags {
 
     /* Show user groups */
     UserGroupsEnabled = 'user-groups-enabled',
+
+    /** Collapse dashboard into a simple vertical stack on mobile */
+    DashboardMobileVerticalStack = 'dashboard-mobile-vertical-stack',
+
+    /** */
+    LazyLoadDashboardTiles = 'lazy-load-dashboard-tiles',
```
filp, line 44: Unrelated, just migrating a feature flag in the same file.

ZeRego (approved):
lgtm 👍

hamzahc1:
@filp I'm not sure if this is related to your change but see this video for context. when I switched to ipad and refreshed in chrome's emulator - I lost the nav bar and one chart randomly rendered a bit off 

https://github.com/lightdash/lightdash/assets/14341285/fc4705d2-7193-4a5a-ad52-31fcbc11952f

filp:
@hamzahc1 this PR only accounts for the smallest breakpoint, which would be most phones - any modern iPad (I believe) falls outside that range, so it for now continues to have the previous behavior.

I'll tackle this in a follow-up PR, but the approach for larger displays is a bit more subtle (in terms of making a decision of where the line stands for stacking vertically vs accepting some weird arranging in exchange for more user control).

Issue #4431, opened 2023-02-09
x-axis labels overlaps on small  monthly / weekly charts


rephus:
### Description

![image](https://user-images.githubusercontent.com/1983672/217866608-8de7ce46-3d05-459a-8a04-dd23c22c95aa.png)

on a small chart on a dashboard looks like this :

![image](https://user-images.githubusercontent.com/1983672/217866753-a736f1c7-7d54-4d20-97db-63524d26b74c.png)

magnew:
Not stale, we still see issues with this

rephus:
I don't think this is an issue anymore, I can't replicate. This probably got fixed in a nother ticket. 

For week and month, we avoid overlapping by skipping months if needed (automatically handled by echarts) 
![image](https://github.com/lightdash/lightdash/assets/1983672/0e64bed3-a956-463b-b2ac-465881dbe4ad)

![image](https://github.com/lightdash/lightdash/assets/1983672/ef81f350-93ad-44d3-8e28-b40f8b55da27)

TuringLovesDeathMetal:
This bug is still happening in some parts of the app. I can't find a pattern to reproduce it. 

Here are examples where it breaks:
[...]

PR #9700, opened 2024-04-09
fix: chart labels overlapping on small chart tiles

filp:
Closes: #4431 

### Description:

- Force-hides overlapping chart labels (x and y, regardless of axis type).

Before:

image

After: 
image

### Reviewer actions


hamzahc1:
@filp - there's something strange that happens (you can see in your screenshots too), where the x-axis ticks don't really line up with the bars on the chart. Or it seems like they do initially, and then stop - is that related at all to the fix you're making or should that be a separate issue?

hamzahc1:
Also in relation to this PR - while I see what you've done here, one of the original requests for this was actually that we figure out a way to display all the chart labels without overlapping (e.g. auto rotate labels etc). It's a hard one, but thought I'd mention in case we get complaints based on this proposed change.

filp:
Thanks for the extra context @hamzahc1  - I'll close this PR for now and reopen it with an adjusted approach.

PR #9701, opened 2024-04-09
fix: improve time filter layout

magnew:
### Description:
Improve time filter layout. Before the field select took up too much space, now the space is space split is more even

Before
image

With this change:
Screenshot 2024-04-09 at 19 01 08

### Reviewer actions


rephus (approved):
Great, it look better now. Although I don't understand why your new code is changing the size of the filter input..

magnew:
> Great, it looks better now. Although I don't understand why your new code is changing the size of the filter input..

Mostly the right hand side (time inputs) werent set to 100% so they didn't flex larger even when they could and they other one was set to grow. The effect of this is basically to give both sides equal flex weight. If that makes sense...

Issue #9703, opened 2024-04-09
Missing quote char wrapping ORDERBY's in Quick Table Calculations


joaoviana:
### Description

On the BE we use `fieldQuoteChar` in joins: https://github.com/lightdash/lightdash/blob/main/packages/backend/src/queryBuilder.ts#L728

Let's do the same in: https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx#L99

### What problem does this solve?

Quick calculations erroring with snowflake users

PR #9704, opened 2024-04-09
fix: quick calculation order bys quote char

joaoviana:
Closes: #9703

### Description:

Follow the same convention for sorted field IDs on the BE by wrapping the `fieldId` in the quick calculation `sorts` with the fieldquotechar - dependant on warehouse

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
+++ b/packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
@@ -2,6 +2,7 @@ import {
     assertUnreachable,
     CustomFormatType,
     MetricType,
+    WarehouseTypes,
     type CustomFormat,
     type Metric,
     type SortField,
@@ -11,7 +12,9 @@ import { Menu } from '@mantine/core';
 import { type FC } from 'react';
 import { useExplorerContext } from '../../../providers/ExplorerProvider';
 
+import { useParams } from 'react-router-dom';
 import { getUniqueTableCalculationName } from '../../../features/tableCalculation/utils';
+import { useProject } from '../../../hooks/useProject';
 import { useTracking } from '../../../providers/TrackingProvider';
 import { EventName } from '../../../types/Events';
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
```diff
@@ -90,13 +93,36 @@ const getSqlForQuickCalculation = (
     quickCalculation: QuickCalculation,
     fieldReference: string,
     sorts: SortField[],
+    warehouseType: WarehouseTypes | undefined,
 ) => {
+    let fieldQuoteChar = '"';
```
owlas, line 98: this logic should already be available in common, do we need to redefine the mapping between warehouses and quote chars?

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
```diff
@@ -90,13 +93,36 @@ const getSqlForQuickCalculation = (
...
+        switch (warehouseType) {
+            case WarehouseTypes.BIGQUERY:
+            case WarehouseTypes.DATABRICKS:
+                fieldQuoteChar = '`';
+                break;
+            case WarehouseTypes.SNOWFLAKE:
+            case WarehouseTypes.REDSHIFT:
+            case WarehouseTypes.POSTGRES:
+            case WarehouseTypes.TRINO:
+                break;
```
rephus, line 109: personal opinion: We should return `"` even if this already set on the `let` variable, for readability.

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
```diff
@@ -90,13 +93,36 @@ const getSqlForQuickCalculation = (
...
     fieldReference: string,
     sorts: SortField[],
+    warehouseType: WarehouseTypes | undefined,
 ) => {
+    let fieldQuoteChar = '"';
+    if (warehouseType) {
+        switch (warehouseType) {
+            case WarehouseTypes.BIGQUERY:
+            case WarehouseTypes.DATABRICKS:
+                fieldQuoteChar = '`';
```
rephus, line 103: personal opinion: when dealing with this type of logic, I prefer a function-const + switch + return, rather than a `let variable` + switch + break.

rephus (commented):
It is a shame we need to duplicate logic, I might consider moving the `getFieldQuoteChar` from warehouses to common, but that will require a bigger refactoring, perhaps we should create a tech debt ticket ? what do you think ? Apart from that, the fix looks good.

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
joaoviana: sounds good, will change 👍

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
joaoviana: sure!

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
joaoviana: Unfortunately, this is only really available in `packages/warehouses` and then used mostly on the BE.
I would have to get the warehouse client through a method like: https://github.com/lightdash/lightdash/blob/main/packages/warehouses/src/warehouseClientFromCredentials.ts
But that isn't available yet for consumption on the `frontend`...

Like @rephus mentioned here: https://github.com/lightdash/lightdash/pull/9704#pullrequestreview-1990904779 it would require a bigger refactoring to get this working as expected: a centralised class/util that would do this job for us  in both `backend` and `frontend` packages.

We then should follow the suggestion to create a tech debt ticket ^ and I can also add a TODO here. LMK what you think...

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
joaoviana: done! https://github.com/lightdash/lightdash/pull/9704/commits/96e540daad1a3e3e633ae5c96684b171e5d6a85d

packages/frontend/src/components/Explorer/ResultsCard/QuickCalculations.tsx
joaoviana: done: https://github.com/lightdash/lightdash/pull/9704/commits/96e540daad1a3e3e633ae5c96684b171e5d6a85d

owlas (approved):

Issue #9710, opened 2024-04-10
Group color assignment may be confusing when only a single group is available


filp:
### Description

When a single group is available with multiple metrics within it, we default to showing the top level group, which can lead to a single color being used

### Steps to Reproduce the Bug or Issue

N/A

### Cloud or self-hosting

None

PR #9716, opened 2024-04-10
fix: improve color assignment for charts with multiple groups

filp:
Closes: #9710

### Description:

- Color mapping groups are now suffixed with its number of groups if >1 (e.g `order_status_n2.shipped`) to improve serial color assignment when switching chart groups, by preventing different combinations of groups from taking colors from a single color pool.
- All pivot values are considered when building color mapping identifiers, so all permutations of groups are assigned a unique color (e.g `[shipped - <some other metric>]` is no longer the same color as just `[shipped]`)

### Reviewer actions


almeidabbm (approved):
Looks super! 🚀 

image

flipped axis

image

PR #9719, opened 2024-04-10
chore: add env var to enable oidc linking

ZeRego:
Closes: 

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UserService.test.ts
+++ b/packages/backend/src/services/UserService.test.ts
@@ -286,11 +286,6 @@ describe('UserService', () => {
             );
         });
         test('should create user', async () => {
-            // Mock that no identity is found for that email
-            (
-                openIdIdentityModel.findIdentitiesByEmail as jest.Mock
-            ).mockImplementationOnce(() => []);
-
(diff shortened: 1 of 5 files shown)


packages/backend/src/services/UserService.test.ts
```diff
@@ -353,7 +343,14 @@ describe('UserService', () => {
             );
         });
         test('should link openid to an existing user that has another OIDC with the same email', async () => {
-            await userService.loginWithOpenId(openIdUser, undefined, undefined);
+            const service = createUserService({
+                ...lightdashConfigMock,
+                auth: {
+                    ...lightdashConfigMock.auth,
+                    enableOidcLinking: true,
```
rephus, line 350: do we need another test to check what happens if this is false ?

owlas (approved):

rephus (approved):
LGTM, 1 small suggestion

packages/backend/src/services/UserService.test.ts
ZeRego: `should create user` covers that case as both use `await service.loginWithOpenId(openIdUser, undefined, undefined);` but this once has the feature flag enabled.

PR #9725, opened 2024-04-10
chore: re-enable e2e datetime tests

rephus:
Closes: 

### Reviewer actions

### DIFF
--- a/packages/e2e/cypress/e2e/app/dates.cy.ts
+++ b/packages/e2e/cypress/e2e/app/dates.cy.ts
@@ -126,144 +126,146 @@ describe('Date tests', () => {
         });
     });
 
-    it.skip('Should use UTC dates', () => {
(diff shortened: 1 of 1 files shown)


rephus:
Of course they don't work on render... 
image

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -13,6 +12,21 @@ function getLocalISOString(date: Date) {
...
+
+    // Check if we found a match
+    if (match) {
+        return match[1]; // eg: +0200
+    }
+    return null; // Return null if no match was found
+}
+
+function getLocalTime(datetime: string) {
+    return dayjs(`${datetime} UTC`).format('YYYY-MM-DD HH:mm:ss');
```
rephus, line 28: This method returns the current date we use in filters (localtime)

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -126,144 +159,169 @@ describe('Date tests', () => {
...
         cy.visit(
-            `/projects/${SEED_PROJECT.project_uuid}/tables/orders${exploreStateUrlParams}`,
+            `/projects/${SEED_PROJECT.project_uuid}/tables/events${exploreStateUrlParams}`,
```
rephus, line 165: Update to use events , which contains more datetime fields , and orders has changed.

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -126,144 +159,169 @@ describe('Date tests', () => {
...
+        cy.contains('SQL');
+        cy.findAllByText('Loading chart').should('have.length', 0);
 
+        cy.contains('1 active filter');
         cy.findByTestId('Filters-card-expand').click();
-        cy.findAllByText('Loading chart').should('have.length', 0);
-        cy.findByDisplayValue('2022-07-11, 14:23:11:000');
-        cy.get('svg g text').contains('2017-02-11, 03:00:00:000 (+00:00)');
-        cy.get('tbody td').contains('2017-02-11, 03:00:00:000 (+00:00)');
+        cy.contains('11 Aug 2020 22:58:00'); // Filter in UTC
```
rephus, line 172: This is the same for all timezones, as expected

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -126,144 +159,169 @@ describe('Date tests', () => {
...
+
+        cy.contains(`2020-08-11, 22:58:00:000 (+00:00)`); // Data in results, this comes from the server, so depends on the server timezone
```
rephus, line 174: This is very interesting, the +00:00 can change depending on the server timezone, so if I'm running `backend` on UTC+2, this will show +02:00, it doens't matter what's my warehouse

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -126,144 +159,169 @@ describe('Date tests', () => {
...
+
+        // Time sensitive fields in localtime
+        const timezone = Cypress.env('TZ');
+        switch (timezone) {
+            case 'UTC':
+                cy.contains('2020-08-11 22:58:00'); // Filter in localtime //Timezone sensitive
```
rephus, line 180: Instead of using a dynamic method, I prefer to chose between the different timezones we use and type it manually

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -126,144 +159,169 @@ describe('Date tests', () => {
...
+        }
+
         cy.findByTestId('SQL-card-expand').click();
-        cy.get('code')
-            .invoke('text')
-            .should(
-                'include',
-                '("customers".created) < (\'2022-07-11 14:23:11\')',
-            );
+        cy.contains(`("events".timestamp_tz) = ('2020-08-11 22:58:00')`); // SQL
```
rephus, line 198: This is also the same across all timezones

packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -126,144 +159,169 @@ describe('Date tests', () => {
...
+
+        // TODO tooltip in charts
+        // cy.get('svg g').trigger('mouseover')
+        // cy.contains('2020-08-11 22:58:00');
     });
 
-    it.skip('Should filter by date on results table', () => {
+    it('Should filter by date on results table', () => {
+        // This test should not be timezone sensitive
```
rephus, line 206: when dealing with `dates` , all timezones return the same thing.

packages/e2e/cypress/e2e/app/dates.cy.ts
magnew: Weird name though. If it's called timestamp_tz it should probably have tz information

rephus: timestamp_tz does have timezone information, but I don't think we use it in the sql, (it is probably stripped)

magnew (approved):
Excellent

PR #9731, opened 2024-04-11
chore: make bug report description field a textarea

filp:
### Description:

On the issue report form, makes the description field a textarea (like remaining fields) instead of a single line input. The single line input is awkward to write in (particularly since most useful descriptions will be a couple of lines, or at least a longer single line), and doesn't currently show the full placeholder text.

Field in question as it is now (and the second field is what it'll look like instead):

image

### Reviewer actions

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -2,42 +2,43 @@ name: Bug report
 description: Create a report to help us improve
 labels: 🐛 bug
 body:
-  - type: input
-    id: description
-    attributes:
-      label: Description
-      description: What happened and what did you expect?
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -2,42 +2,43 @@ name: Bug report
...
-    id: cloud
-    attributes:
-      label: Cloud or self-hosting
-      description: Did you get this bug on Lightdash cloud or self-hosting
-      options:
-        - cloud
-        - self-hosting
-    validations:
-      required: false
+    - type: textarea
```
filp, line 5: This is the relevant change, everything else is YAML auto-formatting with prettier.

magnew (approved):
Nice

Issue #9387, opened 2024-03-15
Add timezone support for scheduled deliveries


IrakliJani:
### Description

add timezone support for scheduled report deliveries. 

![image](https://github.com/lightdash/lightdash/assets/962095/955ba4ca-54eb-400c-af25-9aee5dd22053)

### What problem does this solve?

This feature will enable users to set and adjust timezones for their report schedules, ensuring consistent delivery times regardless of Daylight Saving Time changes.

It also solves the problem of users having to manually convert times between their current timezone and UTC.

owlas:
This issue was mentioned by a user in slack: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1710175646725509?thread_ts=1710175646.725509&cid=C03MG2VTR08

ayush117:
Hi @IrakliJani, I've addressed the concerns in issue #9738 . Could you please take a moment to review my work and provide feedback? Thank you!

IrakliJani:
Appreciate it @ayush117 🙌  we'll have a look to the PR shortly

jpetey75:
This issue is a duplicate of https://github.com/lightdash/lightdash/issues/4958 so I will close it in favor of that one.

PR #9738, opened 2024-04-11
feat: Add timezone support

ayush117:
Closes: #9387 

### Description:
This feature enables users to set and adjust timezones for their report schedules.

[Screencast from 13-04-24 05:09:44 PM IST.webm](https://github.com/lightdash/lightdash/assets/25580209/9969b14b-2544-45f2-b60e-5509e319d942)

### Reviewer actions


rephus:
Hi @ayush117 . I've been thinking about this approach, instead of passing the local timezone to the backend, and storing it on the db. Why don't we keep your UI but simply convert the datetime into UTC so we keep storing the datetime in UTC ? This is more consistent with the rest of our codebase and easier to implement. What do you think  ?

rephus:
Hi @ayush117 , I'm going to close this PR for now, feel free to reopen the PR and implement the changes I suggested to simplify this feature.

Issue #9720, opened 2024-04-10
We should consider timezones in date filters for dates that have time information.


TuringLovesDeathMetal:
### Description

We don’t consider timezones in date filters. Even if date filters are based off of timestamp columns. (e.g. 11PM GMT, April 9th is 12AM CET, April 10th - we don’t take this into account). We should change this.

Specifically:
- date filters built off of columns which have time information should consider time (or timezones) when generating SQL for the filters
- date filters built off of columns which have no time information (e.g. order date in the orders table in Jaffle shop) should not consider time (or timezones) in the filters.

### Acceptance criteria

Here's an example user story and how I'd expect it to work at each step:

1. I am in PST.
2. I build a query in Lightdash to answer the question: "How many queries were executed on April 9th?"
3. When I say "April 9th", I mean "between 00:00:00 (-7:00) and 24:00:00 (-7:00) - i.e. April 9th PST"
4. I use the `timestamp day` value to filter my chart (i.e. the field in my data warehouse has time information, since it's a timestamp, but in Lightdash, I'm using the `day` time interval)
<img width="386" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/ba1a7e61-9e76-4995-aa6c-64ae2f8f26cc">
5. The filter shows up in my `filters` tab **and has some indication of the timezone**
<img width="1083" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/869f9a14-e53d-435a-b8a0-3e42d7bf54a7">

6. This filter would result in SQL that queries any `query executed` events that happened between `2024-04-09 07:00:00 UTC (+00:00)` AND `2024-04-10 07:00:00 UTC (+00:00)`

7. The results in my table should have `timestamp day` values that are `2024-04-09` AND `2024-04-10` (assuming that there are events that occur between `2024-04-10 00:00:00 UTC (+00:00)` and `2024-04-10 07:00:00 UTC (+00:00)`. Basically, the values in my results table should show me values in UTC.

PR #9741, opened 2024-04-12
feat: add timezone on warehouse session

rephus:
Closes: #9720

### Description:

Screenshot from 2024-04-12 01-44-57

### Things not done in this PR 

- [ ] Sharing URLs : If I (Person A) share a URL to a chart that isn’t saved with someone (Person B) in a different timezone, the chart should show in Person A’s timezone.
- [ ] explore from here? does this take into consideration the timezone from the chart?: It uses the timezone from the chart, not your local timezone.
- [ ] Bigquery timezone conversion

### Reviewer actions


rephus:
Bigquery doesn't support tiemzone with sessions, a date conversion is required, it will be done on a separate PR

packages/common/src/types/metricQuery.ts
```diff
@@ -143,4 +144,5 @@ export type MetricQueryRequest = {
     customDimensions?: CustomDimension[];
     granularity?: DateGranularity;
     metadata?: MetricQuery['metadata'];
+    timezone?: string;
```
magnew, line 147: I'm planning to make this an Enum with the picker checkin. Any concerns?

magnew (commented):
Looks good!

magnew (approved):

packages/frontend/src/hooks/useQueryResults.tsx
```diff
@@ -82,6 +83,7 @@ const getQueryResults = async ({
     const timezoneFixQuery = query && {
```
magnew, line 83: There are some other places where we do something similar to this. I'm just noting that we should update those too (ex useCompliedSQL, useCalculateTotals)

packages/warehouses/.eslintrc.js
```diff
@@ -17,5 +17,6 @@ module.exports = {
         'no-case-declarations': 'off',
         'import/prefer-default-export': 'off',
         'class-methods-use-this': 'off',
+        'no-console': 'off',
```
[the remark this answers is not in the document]
rephus: no, we decided to remove console log warnings on warehouses

packages/common/src/types/metricQuery.ts
rephus: yeah, that makes sense, I didn't know what values to give, so I used a generic string

Issue #9574, opened 2024-03-28
Remove the separation between metrics and dimensions when applying filters to the explore view


PriPatel:
- [ ] There should only be one section -> `Any/All of the following conditions match`.

### Examples of how to convert existing configurations
Example 1: 
<img width="889" alt="Screenshot 2024-03-28 at 13 55 17" src="https://github.com/lightdash/lightdash/assets/12776522/360fbb34-517a-4448-9d31-0c51373ed3e9">

...turns into something like this: 

<img width="875" alt="Screenshot 2024-03-28 at 14 09 07" src="https://github.com/lightdash/lightdash/assets/12776522/2778342c-1ac6-4663-9b2f-a7304f33c239">

Example 2: 

![image](https://github.com/lightdash/lightdash/assets/12776522/b4148574-ded5-43be-9a07-d0817e0fc331)
...turns into something like this: 
![image](https://github.com/lightdash/lightdash/assets/12776522/c84a634f-a1a5-4e50-80f9-4bcebbf3cdd3)

Example 3:

<img width="926" alt="Screenshot 2024-03-28 at 14 28 28" src="https://github.com/lightdash/lightdash/assets/12776522/233fd9d7-f996-4f5c-be80-93a7a4e0b7ac">
...turns into something like this: 
<img width="927" alt="Screenshot 2024-03-28 at 14 27 57" src="https://github.com/lightdash/lightdash/assets/12776522/3ac9232b-c84d-4f9f-9e55-7f9f3e2abb3a">

### Adding restrictions to certain field combinations

- [ ] If the user has used the 'Any' condition for a dimension field, they cannot add a metric field to this list. Disable the metric fields in the dropdown and show a tooltip saying `You cannot combine metrics and dimensions by using the ‘or’ function' Create a nested group instead' 
- [ ] If the user has used the 'Any' condition for a metric field, they cannot add a dimension field to this list. 
- [ ] If the user has used the 'All' condition and included a metric field + a dimension field, they cannot change this to an Any condition. Disable the dropdown with the message above.

PR #9743, opened 2024-04-12
chore: remove type separation in filters and then convert it back

almeidabbm:
Closes: #9574

### Reviewer actions


packages/common/src/utils/filters.ts
```diff
@@ -421,6 +422,133 @@ export const getFilterRulesByFieldType = (
...
+                        filters.tableCalculations,
+                    ],
+                } as FilterGroup;
+            }
+        }
+
+        return accumulator;
+    }, {} as Filters);
+};
+
```
almeidabbm, line 551: Will need to make this function a bit more readable as here is where most of the logic is, basically it converts the filter groups into filters of type `Filters` with the types associated with them so they can be saved in the database

packages/frontend/src/components/common/Filters/FilterGroupForm.tsx
```diff
@@ -40,14 +46,51 @@ const FilterGroupForm: FC<Props> = ({
...
+            setConditionLabel('metric');
+            return metrics;
+        }
+
+        if (filters.tableCalculations) {
+            setConditionLabel('table calculation');
+            return tableCalculations;
+        }
+
+        return [];
```
almeidabbm, line 92: Here as well, will need to check if I find a smarter way to go about this

packages/common/src/utils/filters.ts
almeidabbm: nvm, spoke with @owlas we can keep this as is since and refactor in case there's any performance hits because of it

packages/common/src/utils/filters.ts
```diff
@@ -421,6 +421,125 @@ export const getFilterRulesByFieldType = (
         },
     );
 
+const flattenFilterGroup = (filterGroup: FilterGroup): FilterGroup => {
```
filp, line 424: A quick comment explaining what the method does up top would be a good extra touch 💅

packages/common/src/utils/filters.ts
```diff
@@ -421,6 +421,125 @@ export const getFilterRulesByFieldType = (
...
+): Filters => {
+    const flatFilterGroup = flattenFilterGroup(filterGroup);
+    const items = getItemsFromFilterGroup(flatFilterGroup);
+    return items.reduce<Filters>((accumulator, item) => {
+        if (isFilterRule(item)) {
+            const fieldInRule = fields.find(
+                (field) => fieldId(field) === item.target.fieldId,
+            );
+
+            if (fieldInRule) {
```
filp, line 464: The code is fairly straight-forward, but some short comments explaining the business reasoning behind each of these branches would be great too 💅

almeidabbm: I've added the comments to both functions 👌 
Also fixed a slight bug regarding the filter group type https://github.com/lightdash/lightdash/pull/9743/commits/9530d022a018b6eab87410e4581ad153707282ce - see commit description

rephus:
UI question, why is the `all/any` selectable only when there are 3+ filters ? why not 2 ? 

image

image

almeidabbm:
> Is this expected ?
> 
>  Screencast.from.22-04-24.13.20.40.webm

This is 100% not expected, nice catch! it's probably due to only flattening the and filters 🤔 will check it out

almeidabbm:
> UI question, why is the `all/any` selectable only when there are 3+ filters ? why not 2 ?
> 
> image
> 
> image

Spoke with @PriPatel and we agreed on displaying the select for `any/or` if the filter count is >= 2. This has been added in the most recent commit. The bug you found should also be fixed now 👌

rephus:
I can't add filters for dates. 

This was probably not introduced here, so feel free to merge if that's the case, we can fix on another PR if this bug is already in master
[Screencast from 23-04-24 09:43:31.webm](https://github.com/lightdash/lightdash/assets/1983672/71c130c3-0dc0-4002-8743-63d21ffbf8de)

rephus (approved):

almeidabbm:
> I can't add filters for dates.
> 
> This was probably not introduced here, so feel free to merge if that's the case, we can fix on another PR if this bug is already in master
> 
>  Screencast.from.23-04-24.09.43.31.webm

I am not sure I can reproduce that, before I merge, could you just confirm on your side if that field `days_since_last_order` has `hidden: true` in your  `schema.yml`?
For me to be able to show that field I had to remove that `hidden` but then I could add the filter, only got the error when setting `hidden: true` again:


https://github.com/lightdash/lightdash/assets/22939015/915a8882-c8dd-428e-ae6f-6ae5b5b21e51

rephus (approved):

Issue #9745, opened 2024-04-12
Add a timezone picker to chart filters


magnew:
### Description

Add a timezone picker to chart filters. It will choose the session time the SQL is run in. It has two special values 1) UTC, 2) Local time (dynamic), and options to select other times.

PR #9753, opened 2024-04-14
feat: add timezone picker to the explores page

magnew:
Closes: #9745 

### Description:

Adds a timezone picker to the explores page that sets the timezone the explore will run in. It's behind a feature flag for now. This review should set us up to have something usable, but there are a few things still to be done:
- This needs **design work**. It's pretty basic at the moment (screenshot below)
- What kind of strings do we want to show here? This adds a subset of the IANA names, but should we also show offsets?
- How do we want to handle selecting a local value? Should it be a special value called 'Local' in the list? At the moment if any of them match your local TZ, it will say so in the list, but this is 1) Probably too subtle, 2) not great cause we only have a subset of the strings, so it will not always show up.
- Bring this to dashboards or wherever else it's needed
- More testing

Screenshot 2024-04-14 at 15 07 07

### Reviewer actions


packages/frontend/src/hooks/useQueryResults.tsx
```diff
@@ -83,7 +83,7 @@ const getQueryResults = async ({
     const timezoneFixQuery = query && {
         ...query,
         filters: convertDateFilters(query.filters),
-        timezone: dayjs.tz.guess(),
+        timezone: query.timezone || TimeZone.UTC,
```
rephus, line 86: I think if not specified, this should be undefined, and let the backend handle it properly (without forcing to UTC)

rephus: Not sure if we want to include a `default` timezone in the picker that makes this undefined. 


I'm saying this because I'm not sure if `SET timezone UTC` is the same as `not setting the timezone` in the warehouse.

rephus: Note this will not be captured by the feature flag, so it will affect users as soon as we release this.

packages/frontend/src/components/Explorer/ExplorerHeader/index.tsx
```diff
@@ -82,6 +83,8 @@ const ExplorerHeader: FC = memo(() => {
                     </Tooltip>
                 )}
 
+                <TimeZonePicker />
```
rephus, line 86: should this be behind a feature flag ?

magnew: I put the flag in the component so that anywhere we put it it will be hidden. It returns null if the flag is off

TuringLovesDeathMetal:
> How do we want to handle selecting a local value?

^I'm not sure what this means. I think there should _**always**_ be a timezone selected. I think by default, it should use your local timezone when building a query. 

In the list, there should be some indication that it's your local timezone (optional, it wouldn't be the end of the world if it wasn't there, you should know your local timezone). 

**One thing that's missing:** can we add `timezone` as copy in the timezone picker (otherwise, it's not clear what the `Paris/Madrid` actually means).

rephus (approved):
LGTM

packages/frontend/src/hooks/useQueryResults.tsx
magnew: Updated

magnew:
> > How do we want to handle selecting a local value?
> 
> ^I'm not sure what this means. I think there should _**always**_ be a timezone selected. I think by default, it should use your local timezone when building a query.
> 
> In the list, there should be some indication that it's your local timezone (optional, it wouldn't be the end of the world if it wasn't there, you should know your local timezone).
> 
> **One thing that's missing:** can we add `timezone` as copy in the timezone picker (otherwise, it's not clear what the `Paris/Madrid` actually means).

Ok, at the moment its doesnt define a timezone by default because that will mean we aren't changing existing behavior before we ship this. Once we are ready to turn the feature flag on and let people use it, we could make it so the default it local.

At the moment I've implemented it as you said: it tells you which one is the local timezone in the list. My question was kind of around a wrinkle with that though -- since we don't have every TZ in the list, your local TZ won't necessarily match one of them. So its maybe harder than it should be to know your local timezone. If I am in Spain I have to know to pick Paris. There are a couple things we could do to make this better: 1) we could include every timezone. Probably not great cause there are 300 of them, 2) We could highlight the one from the list that has the same offset as you do 3) We could show this same list but always include your local. Maybe this is the best answer.

TuringLovesDeathMetal:
I'd vote for (2)!

Issue #9757, opened 2024-04-15
Add `timezone` to `query_executed` event


TuringLovesDeathMetal:
### Description

Add `timezone` information to the `query executed` event. 

Specifically, the `timezone` should tell us the timezone that the query was using to be executed.

PR #9760, opened 2024-04-15
chore: add timezone to query executed analytic event

rephus:
Closes: https://github.com/lightdash/lightdash/issues/9757

### Description:

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/api/csv.ts
+++ b/packages/frontend/src/api/csv.ts
@@ -47,6 +47,7 @@ export const downloadCsv = async ({
             columnOrder,
             hiddenFields,
             chartName,
+            timezone: query.timezone ?? undefined,
         }),
     });
 };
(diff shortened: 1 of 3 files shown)


TuringLovesDeathMetal:
@rephus Can we add the time offset to the timezone field? (if it's not too hard 🤞 ) e.g. `Paris/Madrid (+01:00)`

If it's complicated to add, then just let me know and I'll approve in the current state :)

packages/frontend/src/api/csv.ts
```diff
@@ -47,6 +47,7 @@ export const downloadCsv = async ({
             columnOrder,
             hiddenFields,
             chartName,
+            timezone: query.timezone ?? undefined,
```
rephus, line 50: fix from Bruno

rephus:
> @rephus Can we add the time offset to the timezone field? (if it's not too hard 🤞 ) e.g. `Paris/Madrid (+01:00)`
> 
> If it's complicated to add, then just let me know and I'll approve in the current state :)

Not impossible, but I'll have to generate it myself from the timezone, I don't have the data at the moment. 

Let me try it quickly, I can probably use some dayjs util function.

rephus:
@TuringLovesDeathMetal 
image

PR #9773, opened 2024-04-16
chore(docs): Update connect-project.mdx

TuringLovesDeathMetal:
Adds requirement for project connection for GitLab based on user feedback

### DIFF
--- a/docs/docs/get-started/setup-lightdash/connect-project.mdx
+++ b/docs/docs/get-started/setup-lightdash/connect-project.mdx
@@ -752,7 +752,7 @@ So, in Lightdash, I'd add a new environment variable and put `key` as `DBT_USER`
 This is used to access your repo.
 See the [instructions for creating a personal access token here](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html).
 
-Select `read_repository` scope when you're creating the token.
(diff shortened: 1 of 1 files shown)


docs/docs/get-started/setup-lightdash/connect-project.mdx
```diff
@@ -752,7 +752,7 @@ So, in Lightdash, I'd add a new environment variable and put `key` as `DBT_USER`
 This is used to access your repo.
 See the [instructions for creating a personal access token here](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html).
 
-Select `read_repository` scope when you're creating the token.
+Select `read_repository` scope when you're creating the token. The token also needs to have a `Reporter` role.
```
almeidabbm, line 755:
```suggestion
Select `read_repository` scope when you're creating the token. The token, if using a **project access token**, or the user, when using a **personal access token**, needs to have permission to download the code. Normally this would be the `Reporter` role.
```

Was checking gitlab's documentation on the roles of tokens, and it seems like it only asks for token role when it is a `Project access token` which is what the user was using. When it is a personal access token it will have the role of the user. This role can be a custom one, so I think it is better to say it needs to have permissions to download code and point them to the example `Reporter` role

image

almeidabbm (approved):

Issue #9797, opened 2024-04-17
Pasting from the clipboard into a filter, it adds it immediately as a filter value


almeidabbm:
### Description

When I paste something from the clipboard into a dashboard, it immediately adds it as a filter value.

The expected behavior here would be to paste the value and use it to narrow down the dropdown options.

### Steps to Reproduce the Bug or Issue

1. Go to a dashboard/chart
2. Click to add a filter to a string item
3. Copy something into the clipboard and then paste into the filter value input

https://github.com/lightdash/lightdash/assets/22939015/f8b805cf-835e-415d-96dd-8820123a0ffa

### version

v0.1070.3

### Cloud or self-hosting

None

PR #9816, opened 2024-04-18
fix: filter adding value after paste instead of searching

almeidabbm:
Closes: [#9797](https://github.com/lightdash/lightdash/issues/9797)
### Description:

Before:

https://github.com/lightdash/lightdash/assets/22939015/d7867ed8-beef-4706-a70f-af52b5380ab3

After:

https://github.com/lightdash/lightdash/assets/22939015/35ec84a5-cbb0-4f67-ad13-92a9c4592dbe

### Reviewer actions


owlas (changes requested):
This removes the ability to paste many comma-seperated values, which was a request from the community at one point.

I think the change should be that if you paste only a single value (without commas) then it doesn't add the value

almeidabbm:
> This removes the ability to paste many comma-seperated values, which was a request from the community at one point.
> 
> I think the change should be that if you paste only a single value (without commas) then it doesn't add the value

Gotcha! That does make sense, I'll adjust the code to match your suggestion and add a comment on the comma separated value handling so we know why that's there 👌 thank you!

IrakliJani (approved):

owlas (approved):
CleanShot 2024-04-22 at 10 01 50@2x

On pasting unfinished value ✅ 

CleanShot 2024-04-22 at 10 02 33@2x

On pasting multiple values ✅ 

CleanShot 2024-04-22 at 10 03 04@2x

On pasting multiple inc. unknown item

PR #9833, opened 2024-04-19
chore: support slack bot dependency injection

ZeRego:
Closes: 

### Description:

Allow us to extend the slack bot class.

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -77,6 +77,19 @@ declare global {
     }
 }
 
+const slackBotFactory = (context: {
+    lightdashConfig: LightdashConfig;
+    analytics: LightdashAnalytics;
+    serviceRepository: ServiceRepository;
+    models: ModelRepository;
+}) =>
+    new SlackBot({
+        lightdashConfig: context.lightdashConfig,
+        analytics: context.analytics,
+        slackAuthenticationModel: context.models.getSlackAuthenticationModel(),
+        unfurlService: context.serviceRepository.getUnfurlService(),
+    });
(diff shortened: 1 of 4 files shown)


packages/backend/src/clients/Slack/Slackbot.ts
```diff
@@ -63,8 +63,6 @@ export class SlackBot {
         this.analytics = analytics;
         this.slackAuthenticationModel = slackAuthenticationModel;
         this.unfurlService = unfurlService;
-
-        void this.start();
```
ZeRego, line 67: Bad practice to run an async function in the constructor. We now call the `start` in App.ts.
This also allows this class methods to be overridden in classes that extend this one.

packages/backend/src/App.ts
```diff
@@ -181,7 +198,7 @@ export default class App {
 
         const expressApp = await this.initExpress();
         this.initSentry(expressApp);
-        this.initSlack();
+        await this.initSlack();
```
IrakliJani, line 201: do we need to await for this?

IrakliJani (approved):
one Q

packages/backend/src/App.ts
ZeRego: There are 3 options:
`void` - the promise doesn't block the thread but can fail silently and trigger an "unhandled promise error"
`await` - the promise blocks the thread but the error is handled
`.catch(...)` - the promise doesn't block the thread and the error is handled.

I tend to await all promises and find `void` to be the worst option of all.

PR #9418, opened 2024-03-18
fix: extends SpaceQuery from ChartSummary type. Add missing fields

almeidabbm:
Closes: #9413 

### Description:

- Changes `SpaceQuery` type to contain same fields as `ChartSummary`
- Adds missing fields to query results

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #9838, opened 2024-04-19
fix: wrong chart icon in pinned list

almeidabbm:
### Description:

This started happening after https://github.com/lightdash/lightdash/pull/9418/files#diff-97f739a09ecb82dabeb72908a35fe8e622e4d81c213616bc90df3a38d065e23f got merged. we stopped returning `chartType` as `chartKind` in the queries + handling of the frontend icons now used `chartKind` which this endpoint didn't return.

Before:

image

After:

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ResourceViewItemModel.ts
+++ b/packages/backend/src/models/ResourceViewItemModel.ts
@@ -95,7 +95,7 @@ const getCharts = async (
             updatedAt: row.updated_at,
             views: row.views,
             firstViewedAt: row.first_viewed_at,
-            chartType: row.chart_kind,
+            chartKind: row.chart_kind,
             updatedByUser: row.updated_by_user_uuid && {
                 userUuid: row.updated_by_user_uuid,
                 firstName: row.updated_by_user_first_name,


packages/backend/src/models/ResourceViewItemModel.ts
```diff
@@ -32,6 +36,7 @@ const getCharts = async (
                 'saved_queries.last_version_updated_by_user_uuid',
             order: 'pinned_chart.order',
             chart_kind: `saved_queries.last_version_chart_kind`,
```
rephus, line 38: instead of returning chart_type, should we return a valid chart_kind ? is it because it is not stored correctly in db ? this way we can avoid making a unnecessary join, which will slow down the call.

almeidabbm: `chart_type` is actually not needed here, only `chart_kind`
The problem was that we were returning `chartType: row.chart_kind` instead of `chartKind: row.chart_kind`, so I can remove the join 👌

rephus (approved):
Tested, works as expected

Issue #9803, opened 2024-04-18
Setting `x-axis` range in horizontal bar charts applies the range to `y-axis`


almeidabbm:
### Description

When I have a horizontal bar chart and disable the auto `x-axis` range by adding a value, the chart will be blank.
This is happening because although we set a value for `x-axis` range, it is applying in the `y-axis`

### Steps to Reproduce the Bug or Issue

1. Create an horizontal bar chart
2. Open chart configuration
3. Click on the axes tab and disable `auto x-axis range`
4. Add a value to `x-axis` range - all charts will disappear
5. Add a value of the `y-axis` to the `x-axis` range - chart will be cut by the `y-axis`

https://github.com/lightdash/lightdash/assets/22939015/f81b7a60-c26e-4978-bfa0-a4de8b9a4a26

### version

v0.1072.3

### Cloud or self-hosting

None

rephus:
Related ticket: https://github.com/lightdash/lightdash/issues/8262

PR #9844, opened 2024-04-22
fix: min max range on horizontal bar chart

almeidabbm:
Closes: #9803 

### Description:

Before:

image

After:

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -1451,7 +1451,10 @@ const useEchartsCartesianConfig = (
                 : getResultValueArray(rows, true);
         try {
             if (!itemsMap) return results;
-            const xFieldId = validCartesianConfig?.layout?.xField;
+            const xFieldId = validCartesianConfig?.layout.flipAxes
+                ? validCartesianConfig?.layout?.yField?.[0]
+                : validCartesianConfig?.layout?.xField;
+
             if (xFieldId === undefined) return results;
             const { min, max } = axes.xAxis[0];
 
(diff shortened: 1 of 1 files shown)


rephus:
I'm still getting the issue described in https://github.com/lightdash/lightdash/issues/9803

[Screencast from 22-04-24 13:43:42.webm](https://github.com/lightdash/lightdash/assets/1983672/fd26a0d3-cabc-4548-ad0f-1f005f871ce4)

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1535,11 +1538,13 @@ const useEchartsCartesianConfig = (
         }
     }, [
         validCartesianConfig?.layout?.xField,
+        validCartesianConfig?.layout.flipAxes,
+        validCartesianConfig?.layout?.yField,
         validCartesianConfig?.eChartsConfig?.series,
         rows,
         itemsMap,
+        axes,
```
rephus, line 1546: unnecesary field ?

almeidabbm:
> I'm still getting the issue described in #9803
> 
>  Screencast.from.22-04-24.13.43.42.webm

This should be fixed now, the range filter was hiding values that were bigger than the max: [fix commit](https://github.com/lightdash/lightdash/pull/9844/commits/744349f00d8242c29088ca04e261189d3645f84f)

rephus:
> > I'm still getting the issue described in #9803
> > Screencast.from.22-04-24.13.43.42.webm
> 
> This should be fixed now, the range filter was hiding values that were bigger than the max: [fix commit](https://github.com/lightdash/lightdash/pull/9844/commits/744349f00d8242c29088ca04e261189d3645f84f)
> 
> But I now have a question, do we want to keep values greater than max hidden? If so then I believe the prev implementation was correct in removing them from the values

I think

TuringLovesDeathMetal:
I feel like we should keep the bar charts showing? 👀 Even if they exceed the axis limit. 

Ideally, there's some indication that they've exceeded the limit (or an option to include one), but let's not add that for now until someone complains. 

I'd expect the same behaviour for area charts btw!

rephus (approved):

Issue #9832, opened 2024-04-19
Delete last tab should not delete tiles


ZeRego:
### Description

There are a few problems that might be related:

- When I remove the first tab (aka default) it doesn't show me the correct amount of tiles in the modal.
   - Fix: When I create the default tab, the existing tiles should be updated with the default tab uuid. 
- When we delete the last tab we should not show the delete modal and we keep the tiles

### Steps to Reproduce the Bug or Issue
Issue 1: 
1. go to dashboard with X tiles
2. add tab
3. delete default tab
4. see modal says 0 tiles will be deleted

Expected:
modal says X tiles will be deleted

---
Issue 2:
1. go to dashboard with tabs
2. delete all tabs
3. no tiles remain

Expected:
the tiles from the last tab remain and the tab component disappears. [Walkthrough here](https://www.figma.com/proto/rlqMIJ4pm3t8aAsiMd3CZJ/Lightdash-Product-New?type=design&node-id=1180-2177&t=5V4JUEEQxER1ZAI7-0&scaling=min-zoom&page-id=1163%3A1580&starting-point-node-id=1180%3A2177&show-proto-sidebar=1)

### Cloud or self-hosting

self-hosting

PR #9848, opened 2024-04-22
fix: Delete last tab should not delete tiles

Shifan-Gu:
Closes: 
[9832](https://github.com/lightdash/lightdash/issues/9832)

### Description:

fix: 
- update all tiles with default tile uuid when first time creating tabs.
- rename "Remove Tab" to "Remove Tabs Component" when only one tab exists.
- delete tab directly instead of opening a modal when deleting last tab.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -325,8 +325,8 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                                                                 )
                                                             }
                                                         >
-                                                            Remove Tab
-                                                            Components
+                                                            Remove Tabs
+                                                            Component
                                                         </Menu.Item>
                                                     ) : (
                                                         <Menu.Item
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -437,7 +437,7 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                         tab={activeTab}
                         dashboardTiles={dashboardTiles}
                         onClose={() => setDeletingTab(false)}
-                        opened={isDeletingTab && dashboardTabs.length > 1}
+                        opened={isDeletingTab && dashboardTabs?.length > 1}
                         onDeleteTab={(uuid) => {
                             handleDeleteTab(uuid);
                         }}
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -419,7 +437,7 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                         tab={activeTab}
                         dashboardTiles={dashboardTiles}
                         onClose={() => setDeletingTab(false)}
-                        opened={isDeletingTab}
+                        opened={isDeletingTab && dashboardTabs.length > 1}
```
rephus, line 440: `dashboardTabs` seems not optional, but I can see this in many places `dashboardTabs?.[0];` , should we check if  dashboardTabs is optional here (and other places) ? like `dashboardTabs?.length`

packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -192,6 +192,9 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                     order: 0,
                 };
                 newTabs.push(firstTab);
+                dashboardTiles?.forEach((tile) => {
+                    tile.tabUuid = firstTab.uuid; // move all tiles to default tab
```
rephus, line 196: I tried deleting a tab with charts, but they were not copied to the first tab.  See recording

[Screencast from 23-04-24 11:12:11.webm](https://github.com/lightdash/lightdash/assets/1983672/164c6c70-64c8-4833-b0b6-998bc41b4d43)

packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -314,13 +317,28 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
...
-                                                        onClick={() =>
-                                                            setDeletingTab(true)
-                                                        }
-                                                    >
-                                                        Remove Tab
-                                                    </Menu.Item>
+                                                    {sortedTabs.length === 1 ? (
+                                                        <Menu.Item
+                                                            onClick={() =>
+                                                                handleDeleteTab(
```
rephus, line 323: this works :+1:

packages/frontend/src/components/DashboardTabs/index.tsx
Shifan-Gu: sorry this comment might be misleading... i mean when you create the tab for the first time, this line will update all tiles' tabUuid to use default tab uuid, like they are 'moved' to default tab.

the video is the expected behaviours, when you delete the tab, all tiles in this tab will be removed as well.
unless the tab is the last one, then it will keep all the tiles.

rephus (approved):

PR #9860, opened 2024-04-23
fix: date ranger picker value not rendered to UTC in sql

Shifan-Gu:
Closes: 

issue:
Screenshot 2024-04-23 at 03 18 15

fix: 
Screenshot 2024-04-23 at 03 15 47

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -742,7 +742,7 @@ describe('Formatting', () => {
                     },
                     new Date('2021-03-10T00:00:00.000Z'),
                 ),
-            ).toEqual('2021-03-10, 00:00:00:000 +00:00');
+            ).toEqual('2021-03-10, 00:00:00:000 (+00:00)');
             expect(
                 formatItemValue(
                     {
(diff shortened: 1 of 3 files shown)


magnew:
@Shifan-Gu thanks for this fix -- I thought I had gotten to it but I must have missed it somehow. 

I updated this to avoid using the custom formatters at all; that's the direction I think we should go. I think keeping your change is fine as well as it makes the formatter safer to use, but I hope we can remove them soon. 

If you don't mind, could you verify the update solves the issue on your end? If so we will merge this. And I am going to follow up with a test that would have caught this.

Shifan-Gu:
> @Shifan-Gu thanks for this fix -- I thought I had gotten to it but I must have missed it somehow.
> 
> I updated this to avoid using the custom formatters at all; that's the direction I think we should go. I think keeping your change is fine as well as it makes the formatter safer to use, but I hope we can remove them soon.
> 
> If you don't mind, could you verify the update solves the issue on your end? If so we will merge this. And I am going to follow up with a test that would have caught this.

Hi @magnew, i patched this fix on our side and our QA has verified this, so far looks all good.

magnew:
> > @Shifan-Gu thanks for this fix -- I thought I had gotten to it but I must have missed it somehow.
> > I updated this to avoid using the custom formatters at all; that's the direction I think we should go. I think keeping your change is fine as well as it makes the formatter safer to use, but I hope we can remove them soon.
> > If you don't mind, could you verify the update solves the issue on your end? If so we will merge this. And I am going to follow up with a test that would have caught this.
> 
> Hi @magnew, i patched this fix on our side and our QA has verified this, so far looks all good.

@Shifan-Gu there was a failing test that made me rethink keeping the first version of this. It was changing the table formatting, which I don't want to do as a side effect. I took out that part but I think this should still work for you.

magnew (approved):

Issue #9850, opened 2024-04-22
Table calculations that are date/timestamp type do not get plotted on the chart


TuringLovesDeathMetal:
### Description

When you create a date/timestamp table calculation, then try to use it on the x-axis, no values appear. 

### Steps to Reproduce the Bug or Issue

https://lightdash-demo-thyme.onrender.com/share/Y9ge_g5SX8wp3KMLqXFe2

### Cloud or self-hosting

None

PR #9862, opened 2024-04-23
feat: select table calculation type

rephus:
Closes: https://github.com/lightdash/lightdash/issues/9850
Closes: https://github.com/lightdash/lightdash/issues/9850

### Description:

I tested everything I can think of, including chart filters, charst axis

Hopefully everything should be backwards compatible, and if no type is defined, default to our current behaviour

Before: 

Screenshot from 2024-04-23 13-16-02

After: 

image

image

### List of changes 

- Now date table calculations work 
Screenshot from 2024-04-23 13-18-08

- You can specify the type of a tc and still apply formatting
image

Screenshot from 2024-04-23 12-50-49

- axis icons fixed 
Screenshot from 2024-04-23 13-58-06

- table calculation filters icons and sql fixed
Screenshot from 2024-04-23 13-50-59

Screenshot from 2024-04-23 13-44-59
Screenshot from 2024-04-23 13-44-35

### Out of scope

- select table calculations on Filters on dashboards
- select table calculations on pivots 

### Reviewer actions

### DIFF
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -160,6 +160,7 @@ export const isTableCalculation = (
         ? !('binType' in item) &&
           !!item.sql &&
           !('description' in item) &&
+          !('tableLabel' in item) &&
           !('tableName' in item)
         : false;
 
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -146,7 +146,7 @@ export type TableCalculationMetadata = {
 
 export interface TableCalculationField extends Field {
     fieldType: FieldType.TABLE_CALCULATION;
-    type: CustomFormatType;
+    type: TableCalculationType;
     index?: number;
     name: string;
     displayName: string;
@@ -159,7 +159,7 @@ export const isTableCalculation = (
     item
         ? !('binType' in item) &&
           !!item.sql &&
-          !('type' in item) &&
+          !('description' in item) &&
           !('tableName' in item)
         : false;
 
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -123,12 +123,20 @@ export enum CustomFormatType {
     ID = 'id',
 }
 
+export enum TableCalculationType {
+    NUMBER = 'number',
+    STRING = 'string',
+    DATE = 'date',
+    TIMESTAMP = 'timestamp',
+    BOOLEAN = 'boolean',
+}
 export type TableCalculation = {
     index?: number;
     name: string;
     displayName: string;
     sql: string;
     format?: CustomFormat;
+    type?: TableCalculationType;
 };
 
 export type TableCalculationMetadata = {
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -154,13 +154,12 @@ export interface TableCalculationField extends Field {
 }
 
 export const isTableCalculation = (
-    item: Item | AdditionalMetric,
+    item: Item | AdditionalMetric | TableCalculationField,
 ): item is TableCalculation =>
     item
         ? !('binType' in item) &&
           !!item.sql &&
           !('description' in item) &&
-          !('tableLabel' in item) &&
           !('tableName' in item)
         : false;
 
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -10,6 +10,7 @@ import {
     isTableCalculation,
     isTableCalculationField,
     MetricType,
+    TableCalculationType,
     type Dimension,
     type Field,
     type FilterableDimension,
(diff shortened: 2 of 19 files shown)


packages/common/src/utils/filters.ts
```diff
@@ -104,15 +104,21 @@ export const filterableDimensionsOnly = (
...
         return FilterType.NUMBER;
     }
 
     const { type } = item;
 
+    if (type === undefined) {
+        // Type check for TableCalculationType
+        return FilterType.NUMBER;
+    }
+
```
almeidabbm, line 117: [nitpick]

wouldn't this be the same as just doing:

```suggestion
    const { type } = item;

    if (type === undefined) {
        // Type check for TableCalculationType
        return FilterType.NUMBER;
    }
```

packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
```diff
@@ -161,7 +168,36 @@ const TableCalculationModal: FC<Props> = ({
...
+                            'Manually select the type of the result of this SQL table calculation, this will help us to treat this field correctly in filters or results.'
+                        }
+                    >
+                        <Select
+                            label={'Result type'}
+                            id="download-type"
+                            defaultValue={'default'}
+                            value={
+                                form.getInputProps('type').value ?? 'default'
+                            }
```
almeidabbm, line 187: Default will always be treated as `number` in the filters, I believe this can be confusing if I try to create a custom dimension that returns string and I leave the `type` as default as the filter then doesn't allow entering the `string` value:

https://github.com/lightdash/lightdash/assets/22939015/33964443-d7bc-4d0c-9513-a044e15cce40

Suggestion could be to have the `defaultValue` of the type select be number, what do you think?

rephus: > Default will always be treated as number in the filters
This is the current behaviour

> Suggestion could be to have the defaultValue of the type select be number, what do you think?
Not a bad idea :thinking: @TuringLovesDeathMetal what do you think ?

TuringLovesDeathMetal: I don't think that's true? I think it's treated as a "value" type when it's selected as a filter, no?

rephus: it is treated as a `value` on the chart, but in LIghtdash, we default to `numbers` for example, on filters, so it makes sense to not have a default, and use number instead. (undefined becomes number)

rephus: Unless we want to use `default` in the future for something like `auto detecting the field based on the SQL` or something

TuringLovesDeathMetal: ^Just to write down what we talked about on the call:

Test that if I have an existing table calculation filter that's using a string operator (e.g. `name` is equal to 'blah'), and we switch table calculations to be assumed as numbers by default, then the existing filter doesn't break.

rephus: Right, so the query keeps working, there was only one bug I had to fix, but now it looks like this. I added  a warning message if you're trying to use a string operation in a numberf filter. 
> Unexpected operator ${type} for number filter type. If you are using a table calculation, please update its result type to string.

I didn't want to add more `backward compatibility` code, so I think it is ok like this for now. 

Screenshot from 2024-04-24 13-00-00

rephus: I also removed the `default` in the UI

packages/common/src/types/field.ts
```diff
@@ -151,7 +159,8 @@ export const isTableCalculation = (
     item
         ? !('binType' in item) &&
           !!item.sql &&
-          !('type' in item) &&
+          !('description' in item) &&
+          !('tableLabel' in item) &&
```
almeidabbm, line 163: Table calculation icons in filters always displays the `numbers` icon
I believe it is because of this which is causing `isTableCalculation` to return `false` in `getFieldIcon` function `packages/frontend/src/components/common/Filters/FieldIcon.tsx`


https://github.com/lightdash/lightdash/assets/22939015/a2dbd301-a6aa-4835-a5e4-5640abe7cb1d

rephus: yes, I can't get this combination right, we have a problem identifying types based on their fields...

rephus: Fixed: 

Screenshot from 2024-04-24 15-09-36

almeidabbm (approved):
LGTM 🚀

Issue #9845, opened 2024-04-22
Null values should be treated as their own value when creating bins 


TuringLovesDeathMetal:
### Description

Right now, I can create bins for my custom dimensions. But, NULL values are  put into the highest bin.

Instead, `NULL` values should be treated as a separate value when creating bins.

PR #9866, opened 2024-04-23
fix: add NULL as an extra bin

almeidabbm:
Closes: [#9845](https://github.com/lightdash/lightdash/issues/9845)

### Description:

- Custom range:
image

- Fixed bins:
image

- Fixed width - was already working without any change:
image

### Reviewer actions

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -775,12 +775,12 @@ export type UpdateProject = Omit<
 
 export const getResultValueArray = (
     rows: ResultRow[],
-    onlyRaw: boolean = false,
+    preferRaw: boolean = false,
 ): Record<string, unknown>[] =>
     rows.map((row) =>
         Object.keys(row).reduce<Record<string, unknown>>((acc, key) => {
-            const value = onlyRaw
-                ? row[key]?.value.raw
+            const value = preferRaw
+                ? row[key]?.value.raw ?? row[key]?.value.formatted
                 : row[key]?.value.formatted || row[key]?.value.raw;
 
             return { ...acc, [key]: value };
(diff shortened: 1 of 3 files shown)


rephus (approved):
LGTM 

We are not showing tooltips or axisLabels on null dimensions. But I guess this is not something we've introduced in this PR, so I'm happy to fix it on a separate PR . 

image

almeidabbm:
> LGTM
> 
> We are not showing tooltips or axisLabels on null dimensions. But I guess this is not something we've introduced in this PR, so I'm happy to fix it on a separate PR .
> 
> image

I have made a fix for this as it seemed like a quick thing in this [commit](https://github.com/lightdash/lightdash/pull/9866/commits/dc518c63edcc29b6e7f86ee1f477b3c8d4a665bb)

Not sure if there could be any side effects tho 🤔 but don't think so as we're only running a fallback to formatted if raw doesn't exist - by using a `nullish coalesce`

packages/common/src/index.ts
```diff
@@ -775,12 +775,12 @@ export type UpdateProject = Omit<
...
     rows: ResultRow[],
-    onlyRaw: boolean = false,
+    preferRaw: boolean = false,
 ): Record<string, unknown>[] =>
     rows.map((row) =>
         Object.keys(row).reduce<Record<string, unknown>>((acc, key) => {
-            const value = onlyRaw
-                ? row[key]?.value.raw
+            const value = preferRaw
+                ? row[key]?.value.raw ?? row[key]?.value.formatted
```
rephus, line 783: I thought linter was not happy about this nested conditions

almeidabbm: Same, I have changed this so either are assigned to a variable and easier to read 👌

TuringLovesDeathMetal:
I feel like extra group looks right 👀 nice one!

PR #9867, opened 2024-04-23
chore: update transformer types and add descriptions

IrakliJani:
### Description:
- adds descriptions to zod type definitions
- updates table column type def

### DIFF
--- a/packages/common/src/transformers/VizConfigTransformers/Table/TableConfigTransformer.ts
+++ b/packages/common/src/transformers/VizConfigTransformers/Table/TableConfigTransformer.ts
@@ -5,21 +5,50 @@ import {
     type VizConfigTransformerArguments,
 } from '../AbstractVizConfigTransformer';
 
-const columnSchema = z.object({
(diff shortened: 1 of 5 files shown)


packages/common/src/transformers/VizConfigTransformers/Table/TableConfigTransformer.ts
```diff
@@ -55,19 +84,19 @@ export class TableConfigTransformer<
...
+                (fieldId) =>
+                    currentColumns?.find(
+                        (column) => column.fieldId === fieldId,
+                    ) ?? {
+                        fieldId,
+                        visible: true,
+                        frozen: false,
+                        name: fieldId,
+                    },
+            );
```
IrakliJani, line 100: I changed the type because "tool calling" struggles with json schema with "additionalProperties".

[fieldId]: object --became--> array of {...object, fieldId}

magnew (approved):
👍

Issue #9537, opened 2024-03-27
'Exploring from here' on a chart from a dashboard adds an unnecessary layer of nesting in the filters


PriPatel:
**Current behaviour:**

![image](https://github.com/lightdash/lightdash/assets/12776522/a3050b06-5587-4045-8e2d-b948088da2c1)

**Expected behaviour:**
![image](https://github.com/lightdash/lightdash/assets/12776522/66040a85-f3cd-40a5-bee2-aa3fbf72fdb5)

joaoviana:
I believe this is no longer reproducible! 
(cc @TuringLovesDeathMetal )

joaoviana:
Can't reproduce this when `exploring from here`. 

One thing to note though:

- When exploring from here on a chart when there's dashboard filters applied
Unnecessary nesting doesn't apply: 
<img width="1475" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/b3256089-85ed-42c5-8f14-f138742a6371">

- On same chart, same dashboard & filters, when exploring from here on `view underlying data`: 
<img width="1478" alt="image" src="https://github.com/lightdash/lightdash/assets/7611706/fa9fd14e-6841-4b5e-94d3-ddf37de76c2e">

PriPatel:
Reopening after the conversation in standup:

Scope:

- [ ] Fix The unncessary nesting takes place here:
dashboard > apply filter to chart > view underlying data > explore from here

- [ ] Simplify the nesting even further, "When exploring from here on a chart when there's dashboard filters applied"
-In the example below, it still makes sense to say 'all' of the following conditions apply and list the 3 filters.

- [ ] Lets test even more cases where its really not necessary to group anything.  

![Image](https://github.com/lightdash/lightdash/assets/12776522/08f54dd8-4e71-4fb7-9054-38c8ee4b9b5a)

almeidabbm:
Depends on [#9574](https://github.com/lightdash/lightdash/issues/9574) being merged

bhavna-shukla:
Still seeing above issue if Dashboard operator (eg "in between) is different from chart filter ( eg is equal to)
created #9965

PR #9874, opened 2024-04-24
fix: flatten root filter group

almeidabbm:
Closes: [#9537](https://github.com/lightdash/lightdash/issues/9537)

### Description:

- Flattens root filter group

**Before**

https://github.com/lightdash/lightdash/assets/22939015/c81ab972-068b-4fe3-821b-1fb74efff690

**After**

https://github.com/lightdash/lightdash/assets/22939015/79443612-e7f9-4cba-94fc-0504fa51c6cc

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/MetricQueryData/UnderlyingDataModalContent.tsx
+++ b/packages/frontend/src/components/MetricQueryData/UnderlyingDataModalContent.tsx
@@ -6,6 +6,7 @@ import {
     FilterOperator,
     getDimensions,
     getFields,
+    getFiltersFromGroup,
     isDimension,
     isField,
     isMetric,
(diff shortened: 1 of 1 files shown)


IrakliJani (commented):
@almeidabbm 
I think it would be best if we fixed this at the point where we build the underlying metric query, rather than after we've created the metric query. It seems like the problem is originating from there.

almeidabbm:
> @almeidabbm I think it would be best if we fixed this at the point where we build the underlying metric query, rather than after we've created the metric query. It seems like the problem is originating from there.

I changed the logic to where we determine the filters of the underlying metric query. Basically instead of flattening at the component level, we're now creating the filters object from the group instead of forcing an `AND` group in the dimensions filters by calling `getFiltersFromGroup` which also flattens the groups and removes any unnecessary nesting

IrakliJani (approved):
approving - assuming tests pass! well done 🚀

Issue #9853, opened 2024-04-23
Add a non-unique slug for charts + dashboards


owlas:
Each chart or dashboard should have a slug generated from it's location + name in lightdash:

always lower case, use hyphens instead of spaces

```
# for charts
charts/space_name/chart_name

# for dashboards
dashboards/space_name/dashboard_name
```

It doesn't need to be unique for now

rephus:
Just to clarify, What happens if the name changes , do we update the slug ? 
I've seen this pattern before, and normally the slug doesn't change once it is created

owlas:
- Index ✅ 
- Make them fixed? Yes this makes sense if we plan to use these in the urls. We can always switch them to be updated later

Finally, would it be possible to have the slug be unique but nullable?

PR #9875, opened 2024-04-24
feat: new resource-slugs

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/9853
### Description:

After migration: 

Screenshot from 2024-04-24 15-46-05
Screenshot from 2024-04-24 15-46-14
Screenshot from 2024-04-24 15-46-26

Adding new charts and spaces:
Screenshot from 2024-04-25 09-03-23

image

### Reviewer actions

test-frontend test-backend

### DIFF
--- a/packages/backend/src/database/entities/savedCharts.ts
+++ b/packages/backend/src/database/entities/savedCharts.ts
@@ -54,6 +54,7 @@ export type SavedChartTable = Knex.CompositeTableType<
             | 'last_version_updated_at'
             | 'last_version_updated_by_user_uuid'
             | 'dashboard_uuid'
+            | 'slug'
         >
     >
 >;
@@ -70,7 +71,7 @@ export type DbSavedChart = {
     last_version_updated_at: Date;
     last_version_updated_by_user_uuid: string | undefined;
     search_vector: string;
-    slug: string;
+    slug: string | undefined; // can be undefined if the chart is in a dashboard
 };
 
 export type DbSavedChartVersion = {
--- a/packages/backend/src/database/entities/savedCharts.ts
+++ b/packages/backend/src/database/entities/savedCharts.ts
@@ -71,7 +71,7 @@ export type DbSavedChart = {
     last_version_updated_at: Date;
     last_version_updated_by_user_uuid: string | undefined;
     search_vector: string;
-    slug?: string; // can be undefined if the chart is in a dashboard
+    slug: string;
 };
 
 export type DbSavedChartVersion = {
--- a/packages/backend/src/database/entities/savedCharts.ts
+++ b/packages/backend/src/database/entities/savedCharts.ts
@@ -22,6 +22,7 @@ type InsertChartInSpace = Pick<
     | 'description'
     | 'last_version_chart_kind'
     | 'last_version_updated_by_user_uuid'
+    | 'slug'
 > & {
     space_id: number;
     dashboard_uuid: null;
@@ -69,6 +70,7 @@ export type DbSavedChart = {
     last_version_updated_at: Date;
     last_version_updated_by_user_uuid: string | undefined;
     search_vector: string;
+    slug: string;
 };
 
 export type DbSavedChartVersion = {
--- a/packages/backend/src/database/entities/savedCharts.ts
+++ b/packages/backend/src/database/entities/savedCharts.ts
@@ -71,7 +71,7 @@ export type DbSavedChart = {
     last_version_updated_at: Date;
     last_version_updated_by_user_uuid: string | undefined;
     search_vector: string;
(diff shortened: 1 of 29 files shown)


packages/backend/src/database/migrations/20240424131220_populate_slugs.ts
```diff
@@ -0,0 +1,45 @@
+import { Knex } from 'knex';
```
rephus, line 1: Running `populate` data in a separate migration, because I remember last time we did something similar, we had problems with the migration because the trasnaction was not committed

packages/backend/src/database/migrations/20240424131220_populate_slugs.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+    const sanitizeSlug = (slug: string) =>
+        `REGEXP_REPLACE(
```
rephus, line 5: unlike the regex in utils, this regex will not remove the trailing `-`  or combine multiple `--` , but to achieve that we will have to add nested regex... I think that's ok.

packages/backend/src/database/entities/spaces.ts
```diff
@@ -10,11 +10,12 @@ export type DbSpace = {
     organization_uuid: string;
     created_by_user_id?: number;
     search_vector: string;
+    slug: string;
```
rephus, line 13: I made changes to all db models and added slug on creation, I didn't add it to the `Space/Chart/Dashboard` model because we don't expose this on the API or anywhere else yet.

packages/backend/src/database/migrations/20240424131220_populate_slugs.ts
```diff
@@ -0,0 +1,45 @@
...
+            LOWER(
+                ${slug}
+            ),
+            '[^a-z0-9]+', '-', 'g'
+        )`;
+    // This will populate slugs only for charts in spaces, not charts in dashboards
```
rephus, line 11: Charts in dashboards will have not have slugs. I think that makes sense, as we can `promote` the dashboard, and all its charts.

packages/backend/src/services/SavedChartsService/SavedChartService.ts
```diff
@@ -378,9 +384,18 @@ export class SavedChartService extends BaseService {
             );
         }
 
+        // If the chart is being moved from a dashboard to a space, we need to create a new slug.
```
rephus, line 387: This

rephus:
I'm getting this error on e2e tests (expected if we're saving the same charts) 
image

rephus:
> This seems like a good reason to generate a slug for every chart, I don't see why not. the slugs just need to be unique within a project and be more friendly than a uuid to look at

Because we said chat slugs are built doing: `charts/spaceName/chartUuid` , if the chart doesn't have a space, we can't generate a slug 

UNLESS you use the dashboard's space for the slug :eyes

packages/backend/src/services/SavedChartsService/SavedChartService.ts
rephus: done

packages/backend/src/database/migrations/20240424131157_add_slugs.ts
```diff
@@ -0,0 +1,36 @@
+import { Knex } from 'knex';
+
+const chartsTable = 'saved_queries';
+const dashboards = 'dashboards';
+const spaces = 'spaces';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(chartsTable, (table) => {
+        table.string('slug').nullable().index();
```
rephus, line 9: this is still nullable because the slugs will be populated on a later migration, and I don't want to give it a permanent defaultTo. 

We could set the field back to `notNullabe` after the populate migration happens, what do you think ?

packages/backend/src/database/entities/savedCharts.ts
```diff
@@ -70,6 +72,7 @@ export type DbSavedChart = {
     last_version_updated_at: Date;
     last_version_updated_by_user_uuid: string | undefined;
     search_vector: string;
+    slug: string;
```
owlas, line 75: suggests the underlying table is `notNullable()`

packages/backend/src/database/migrations/20240424131157_add_slugs.ts
rephus: done

owlas: should be `text` instead of `varchar(255)`

owlas (approved):
looks good! suggest using `text` not `varchar(255)`

PR #9890, opened 2024-04-25
chore: remove unused langchain package

IrakliJani:
### Description:

these packages are not being used anywhere

### Reviewer actions


magnew (approved):

packages/backend/package.json
```diff
@@ -33,8 +33,6 @@
         "@aws-sdk/s3-request-presigner": "^3.272.0",
         "@casl/ability": "^5.4.3",
         "@godaddy/terminus": "^4.12.0",
-        "@langchain/core": "^0.1.51",
```
rephus, line 36: are not they used in commercial ?

rephus: also, there is a deprecation notice when you use `langhain` this is the right way to import langchain

IrakliJani: yes, but do we need them in here? we have different versions on commercial

IrakliJani: and we end up having a merge conflict every time we merge upstream

Issue #9840, opened 2024-04-22
When I apply a dashboard filter and I open a chart from here, the filter in the chart is appllying a wrong filter 


rephus:
### Description

Thread: https://lightdash.slack.com/archives/C05F0N3UPSP/p1713771484020209

### Steps to Reproduce the Bug or Issue

- create a chart with a date filter 
- go to dashboard
- add a date filter on the dashboard to override the chart filter
- explore from here
- the filter result is wrong.

### Cloud or self-hosting

cloud

PR #9928, opened 2024-04-30
fix: keep identical filter settings if they exist when overriding

almeidabbm:
Closes: #9840 

### Description:

- Keeps filter settings when overriding chart filter

Date filter value was being overridden but the settings weren't. Because of this the chart filter's `unitOfTime` was kept the same and wasn't overridden with the dashboard filter one

**Before**

https://github.com/lightdash/lightdash/assets/22939015/c09105e0-3a96-4fb9-922a-9d157c2fe61a

**After**

https://github.com/lightdash/lightdash/assets/22939015/51d47f10-93eb-4acd-b41a-1489c6780964

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -665,6 +665,11 @@ const findAndOverrideChartFilter = (
               ...item,
               id: identicalDashboardFilter.id,
               values: identicalDashboardFilter.values,
+              ...(identicalDashboardFilter.settings
+                  ? {
+                        settings: identicalDashboardFilter.settings,
+                    }
+                  : {}),
           }
         : item;
 };


almeidabbm:
> With above fix seeing over ride filters are not working: steps:
> 
> 1. create chart with filter 1
> 2. create db using chart (created step 1)
> 3. create filters in db (different from step 1)
> 
> Actual result: DB and chart filters are sent in sql resulting in no or incorrect data. Expected Result : DB filters shall override chart filter
> 
> Ex: WHERE ((( (DATE_TRUNC('DAY', cast(parse_datetime(datetime, 'yyyy-MM-dd') as timestamp))) = CAST('2024-04-01' AS timestamp) )) AND ( (DATE_TRUNC('DAY', cast(parse_datetime(datetime, 'yyyy-MM-dd') as timestamp))) = CAST('2024-02-01' AS timestamp) ))

Hello @bhavna-shukla 👋 

The feature of overriding chart filters with dashboard filters is currently behind a feature flag hence why you're probably not seeing the overriding behavior.
This feature flag should be removed this week so that dashboard filter overriding chart filters becomes the default behavior.

PR #9943, opened 2024-05-01
docs: promote charts

rephus:
Closes: 

### Description:

https://deploy-preview-9943--peaceful-bassi-cbf284.netlify.app/guides/how-to-promote-content

### Reviewer actions


docs/docs/guides/how-to-promote-content.md
```diff
@@ -0,0 +1,66 @@
+import SelectProject from './assets/promote-select-project.png';
+import PromoteViewChart from './assets/promote-view-chart.png';
+import PromoteListChart from './assets/promote-list-chart.png';
+import SucessMessage from './assets/promote-success.png';
+
+# How to promote content
+
+Promoting content enables you to copy content from one project to another, this can also be used to replicate changes from one `development project` into a `production project`. This helps keep everything consistent and makes it easier to manage changes.
```
TuringLovesDeathMetal, line 8:
```suggestion
Promoting content enables you to copy content from one project to another, this can also be used to replicate changes from one `preview project` into a `production project`. This helps keep everything consistent and makes it easier to manage changes.
```

I don't know if we should use preview project here? Development project is more like "intuitive", but it's likely that the dev project is actually a preview project.

TuringLovesDeathMetal (approved):
Looks great!

One tiny suggestion/question is around the term `'development project" vs. "preview project". I'm not sure if they're the same thing? Or if actually a development project is different? If they're the same, then it would be great if we could keep the naming consistent (so, calling it a `preview project` so it matches the rest of the product/docs)

^This isn't blocking if you don't think it's worth doing (which is why I've approved!) but let me know if you want to. make changes

owlas (approved):

Issue #9918, opened 2024-04-29
Upload sourcemaps to Sentry


joaoviana:
### Description

Use Sentry x Vite's plugin to upload sourcemaps: https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/uploading/vite/

AC: 
- generate sourcemaps on vite config file
- upload sourcemaps on GH action 
- add necessary secrets to CI config
- upload only on release 
- ensure that releases from package json match the uploaded sourcemaps' release done by the plugin

### What problem does this solve?

We're able to see where the error happened exactly on the raw source code, rather than seeing it on compiled code.

joaoviana:
Reopening because GH action still needs tweaking

PR #9944, opened 2024-05-01
fix: use gh cli and build common before frontend

joaoviana:
Relates to: #9918 

### Description:

Uses `gh` cli to help with posting the workflow dispatch. Tested this locally and was able to trigger this run: https://github.com/lightdash/lightdash/actions/runs/8907197505/job/24460632498 - with this error I found that we should also build `common` before building `frontend`.

Use this as reference: https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event
and this: https://josh-ops.com/posts/gh-auth-login-in-actions/

### Reviewer actions

### DIFF
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -56,13 +56,9 @@ jobs:
       - name: Trigger Sentry release
         if: steps.semantic.outputs.new_release_published == 'true'
         run: |
-          curl -X POST \
+          gh api --method POST \
             -H "Accept: application/vnd.github+json" \
-            -H "Authorization: token ${{ secrets.CI_GITHUB_TOKEN }}" \
-            -d '{
-                  "ref": "main",
-                  "inputs": {
-                    "version": "${{ steps.semantic.outputs.new_release_version }}",
-                  }
-                }' \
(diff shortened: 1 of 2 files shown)


.github/workflows/release-on-merge.yml
```diff
@@ -55,14 +55,12 @@ jobs:
...
-            -H "Authorization: token ${{ secrets.CI_GITHUB_TOKEN }}" \
-            -d '{
-                  "ref": "main",
-                  "inputs": {
-                    "version": "${{ steps.semantic.outputs.new_release_version }}",
-                  }
-                }' \
-            https://api.github.com/repos/lightdash/lightdash/actions/workflows/create-sentry-release.yml/dispatches
\ No newline at end of file
+            -H "X-GitHub-Api-Version: 2022-11-28" \
```
joaoviana, line 63: This API version is probably not needed, but just need to check if all works well first! This was taken from this example: https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event

ZeRego (approved):

PR #9945, opened 2024-05-01
chore: analytic events for promote

rephus:
Closes: 

### Description:

image

image

### Reviewer actions


TuringLovesDeathMetal:
@rephus can we add a `from_project_id` and then `to_project_id`? (it looks like there's already only one project ID in htere, but I'm not too sure if it's the `from` or `to` project - if we could make it really clear, that would be great :) )

rephus:
> @rephus can we add a `from_project_id` and then `to_project_id`? (it looks like there's already only one project ID in htere, but I'm not too sure if it's the `from` or `to` project - if we could make it really clear, that would be great :) )

great suggestion, will do

Issue #3621, opened 2022-10-31
The xtick labels should be aligned with the values being plotted 


TuringLovesDeathMetal:
### Description

Below, I have a chart which is plotting bars on a weekly cadence. But, you can see that the ticks are not in the center of the bars, so just by looking at the chart, I'm not too sure which week each of the bars actually represents. 
<img width="1000" alt="(source)" src="https://user-images.githubusercontent.com/31848148/198968814-9198cb93-3452-40fc-918e-b62c25ed6959.png">

For example, the bar for the week commencing on the 2nd of October has just an `October` label slightly to the left of the bar. Ideally, I'd want the bar to have `Oct 2` as the tick label, in the center of the bar. 
<img width="620" alt="Screenshot 2022-10-31 at 09 16 22" src="https://user-images.githubusercontent.com/31848148/198968821-4fcdcf9a-9f0d-4ca1-b655-6e43ba845deb.png">

**Note:* It looks like this only happens when you have the month name shown on the chart axis (becuase it forces teh month name to appear ont he first day of the month on the axis, then it adds 7 days after that tick). Maybe there's a way of always showing the month name on each tick? Then maybe it will work? 

### What problem does this solve?

You can't read the chart without needing to hover over each of the bars to figure out which dates they're representing. 

### User acceptance criteria:
- By default, the x-tick labels match the values being plotted (e.g if I'm plotting 2nd October, I want the tick under my bar to be `Oct 2`). 
- If there are too many x-tick labels to fit, then we do the same thing as we do now which is remove some of the labels from the axis. But, the remaining labels should still be centered on the bars/values.

PR #9957, opened 2024-05-01
feat: align x-axis ticks with data

magnew:
Closes: #3621 

### Description:

The goal of this is to make data line up with the axis labels and ticks better when looking at WEEK timeframe data. 

To do this, I made the week-timeframe use a categorical axis, which gives much more control over axis styles and such. The downside of the categorical axis is it's not continuous -- it doesn't fill holes where there is no data. To handle this, I found the min and max weeks and made the x-axis always a continuous range between them. 

This does a good job of lining up the bars with the ticks, but more could still be done about overlapping labels. 

This is what a week-based chart (with gaps in the data) looked like before:
Screenshot 2024-05-02 at 15 56 46

And here's what it looks like with this change
Screenshot 2024-05-02 at 15 57 24

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -101,6 +101,12 @@ const getAxisTypeFromField = (item?: ItemsMap[string]): string => {
             case MetricType.DATE:
             case TableCalculationType.DATE:
             case TableCalculationType.TIMESTAMP:
+                if (
+                    'timeInterval' in item &&
+                    item.timeInterval === TimeFrames.WEEK
+                ) {
+                    return 'category';
+                }
                 return 'time';
             default: {
                 return 'category';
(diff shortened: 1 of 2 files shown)


packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1115,6 +1149,10 @@ const getEchartAxes = ({
                         : showGridX,
                 },
                 inverse: !!xAxisConfiguration?.[0].inverse,
+                data: useWeekAxisConfiguration ? customXRange : undefined,
```
ZeRego, line 1152: We need to apply this to all necessary axis.
example when I flip axis

Screenshot 2024-05-03 at 10 05 27

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1115,6 +1149,10 @@ const getEchartAxes = ({
...
+                axisTick: useWeekAxisConfiguration
+                    ? { alignWithLabel: true, interval: 0 }
+                    : undefined,
```
ZeRego, line 1155: Curious about this change as ticks seem to be used as delimiters for categorical axis. So should we change this for all categorical axis or for none or just for weeks 🤔 
Example 

Screenshot 2024-05-03 at 10 07 49

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
magnew: Good call. I meant to do that.

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
magnew: I think just for weeks. We want weeks to behave more like a time axis, but we have to use category axis to get there. In the time axes, the ticks line up with the data and labels.

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
magnew: Done

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -931,6 +939,7 @@ const getEchartAxes = ({
             // This is to ensure the value is correctly formatted on some types
             switch (axisItem.timeInterval) {
                 case TimeFrames.WEEK_NUM:
+                case TimeFrames.WEEK:
                     axisConfig.axisLabel = {
                         formatter: (value: any) => {
                             return formatItemValue(axisItem, value, false);
```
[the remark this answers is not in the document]
magnew: Good question. No, these are already in UTC, converting them would result in moving the times TO utc, which isn't correct. We definitely have some unclarity in the code around this at the moment. I think it's getting better, but its not great that this isn't very clear.

ZeRego (approved):

Issue #8480, opened 2024-01-04
Snowflake error: JWT token is invalid


ZeRego:
Haven't reproduced but I believe the reproduction steps are:
1. Create a project via the CLI using snowflake JWT credentials
2. (wait certain amount of time so JWT expires)
3. Go to project settings in UI and set a user and password
4. try to compile
5. throws error "Snowflake error: JWT token is invalid"

![Image](https://github.com/lightdash/lightdash/assets/9117144/15c9c460-9f3a-4d97-983d-2492e4ea4455)

Snowflake warehouse client should prioritize user and password authentication over JWT token.

PR #9973, opened 2024-05-03
fix: use snowflake user and password if provided

joaoviana:
Closes: #8480

### Description:

If user provides `password` then the `authenticator` method is `SNOWFLAKE` - the default
If the user provides a `decodedPrivateKey`, then it should use `SNOWFLAKE_JWT`

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
@@ -143,11 +143,19 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
                 credentials.quotedIdentifiersIgnoreCase;
         }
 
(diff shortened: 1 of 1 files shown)


packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -143,12 +143,24 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
                 credentials.quotedIdentifiersIgnoreCase;
         }
 
+        let authenticationOptions: Partial<ConnectionOptions> = {};
+
+        if (credentials.password) {
```
joaoviana, line 148: followed these: https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#authentication-methods

owlas (approved):

PR #9977, opened 2024-05-03
feat: pylon chat

ZeRego:
Closes: https://github.com/lightdash/lightdash-internal-work/issues/1651 

### Description:

Changes:
- add new env vars
- pass app id and verification hash to frontend
- init pylon if needed
- show pylon chat instead of intercom if initialized

Note:
- We still show intercom chat if pylon chat is not available
- Keep intercom in login and verification pages as Pylon doesn't support chat when user is not authenticated.

How to test:

1. pull branch locally
2. update `HelpMenu.tsx` to have `{isCloudCustomer || true && (`
3. set env vars

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/HealthService/HealthService.mock.ts
+++ b/packages/backend/src/services/HealthService/HealthService.mock.ts
@@ -1,100 +1,80 @@
-import { LightdashMode, SessionUser } from '@lightdash/common';
-import { LightdashConfig } from '../../config/parseConfig';
+import { HealthState, LightdashMode, SessionUser } from '@lightdash/common';
 
-export const BaseResponse = {
+export const BaseResponse: HealthState = {
     healthy: true,
     version: '0.1.0',
     mode: LightdashMode.DEFAULT,
     isAuthenticated: false,
     requiresOrgRegistration: false,
     localDbtEnabled: true,
-    auth: {
-        disablePasswordAuthentication: false,
-        google: {
-            loginPath: '',
-            oauth2ClientId: '',
-        },
-        okta: {
-            enabled: false,
-            loginPath: '',
-        },
-        oneLogin: {
-            enabled: false,
-            loginPath: '',
-        },
-        azuread: {
-            enabled: false,
-            loginPath: '',
-        },
-        oidc: {
-            enabled: false,
-            loginPath: '',
-        },
-    },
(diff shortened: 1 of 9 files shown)


packages/frontend/src/providers/ThirdPartyServicesProvider.tsx
```diff
@@ -1,10 +1,50 @@
...
+import { Helmet } from 'react-helmet';
 import { IntercomProvider } from 'react-use-intercom';
 import { Intercom } from '../components/Intercom';
 import useSentry from '../hooks/thirdPartyServices/useSentry';
 import { useApp } from './AppProvider';
 
+const Pylon = () => {
+    const { user, health } = useApp();
+    useEffect(() => {
+        if (health.data?.pylon?.appId && user.data) {
```
joaoviana, line 12: what do you think of adding this `health.data.pylon.verificationHash` to the condition?

```suggestion
        if (health.data?.pylon?.appId && health.data.pylon.verificationHash && user.data) {
```

joaoviana (approved):
Tested with the env vars
- Displayed Pylon chat
- Could send message and then with my support user I could reply to the conversation

Tested without the env vars
- intercom chat appeared

packages/frontend/src/providers/ThirdPartyServicesProvider.tsx
ZeRego: Pylon's identity verification is optional. So it can work without the `verificationHash`

packages/backend/src/services/HealthService/HealthService.mock.ts
```diff
@@ -31,70 +42,39 @@ export const BaseResponse = {
...
+        environment: '',
         frontend: {
-            dsn: 'frontend-dsn.sentry.io',
+            dsn: '',
         },
-        environment: 'environment',
-        release: '1234',
+        release: '',
     },
-} as LightdashConfig; // TODO: Refactor this to be a mock of the actual configuration, rather than a partial that might contain incorrect properties.
```
ZeRego, line 97: Removed this mock and used the one from `import { lightdashConfigMock } from '../../config/lightdashConfig.mock';`. 

Update the `BaseResponse` mock to have proper typing and valid values.

Issue #9976, opened 2024-05-03
Add support for some styles in markdown tiles


ZeRego:
### Description

We use sanitize-html library to sanitise the markdown input. They have an option to allow specific styles and values which we can set up. We can start by allowing text alignment and other styles that have a specific set of options.  We should not allow users to enter free text for style values. 
https://github.com/apostrophecms/sanitize-html?tab=readme-ov-file#allowed-css-styles

### What problem does this solve?

Be able to align text to the center.

PR #9979, opened 2024-05-04
feat: allow common properties for text styling in markdown tile html tags

filp:
Closes: #9976

### Description:

Adds support for common text styling/formatting CSS properties, with basic value validations, as part of the `sanitize-html` configuration. This is applied only for sanitization with the markdown tile rule-set.

Updates the existing `sanitizeHtml` utility test suite to match!

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/sanitizeHtml.test.ts
+++ b/packages/common/src/utils/sanitizeHtml.test.ts
@@ -32,62 +32,147 @@ describe('sanitizeHtml', () => {
         ).toEqual('<a href="https://www.lightdash.com/">Lightdash</a>');
     });
 
-    test('markdown tile rule set', () => {
-        expect(
-            sanitizeHtml(
-                '<iframe src="https://google.com" width=400 height="300"></iframe>',
-                HTML_SANITIZE_MARKDOWN_TILE_RULES,
-            ),
-        ).toEqual(
-            '<iframe src="https://google.com" width="400" height="300"></iframe>',
-        );
-
-        expect(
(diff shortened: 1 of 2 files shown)


packages/common/src/utils/sanitizeHtml.ts
```diff
@@ -1,5 +1,47 @@
...
+];
+
+/**
+ * Defines a list of CSS properties and value RegExps, which will be allowed as part of the
+ * style attribute in the above tags.
+ */
+const allowedTextStylingProperties: NonNullable<
+    sanitize.IOptions['allowedStyles']
+>[string] = {
+    'font-size': [/^\d+(?:px|em|rem|%)$/], // 12px, 1em, 20%
```
filp, line 33: There's probably a good case for just making these all `/.*/`, but basic validation is easy enough regardless (with a couple of exceptions at the end)

packages/common/src/utils/sanitizeHtml.test.ts
```diff
@@ -32,62 +32,147 @@ describe('sanitizeHtml', () => {
         ).toEqual('<a href="https://www.lightdash.com/">Lightdash</a>');
     });
 
-    test('markdown tile rule set', () => {
```
filp, line 35: Moved these into a sub-section of the tests, for the markdown tile ruleset.

ZeRego (approved):
lgtm 👍 added a strict validation for color values
waiting for tests to pass before merging

PR #9996, opened 2024-05-07
fix: init sentry before express handlers

joaoviana:
Closes: N/A

### Description:

Initialise Sentry before any other express handlers/middleware

```
The Sentry request handler middleware must be added before any other handlers
app.use(Sentry.Handlers.requestHandler());
```
From the guide here: https://docs.sentry.io/platforms/javascript/guides/express/

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -225,7 +225,10 @@ export default class App {
             App.initNodeProcessMonitor();
         }
 
-        await this.initExpress();
+        const expressApp = express();
+        // NOTE: Sentry must be initialized before any other handlers or middleware so that it can capture requests accordingly
+        await this.initSentry(expressApp);
+        await this.initExpress(expressApp);
 
         this.initSlack().catch((e) => {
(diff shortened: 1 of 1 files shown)


packages/backend/src/App.ts
```diff
@@ -225,8 +225,8 @@ export default class App {
             App.initNodeProcessMonitor();
         }
 
-        const expressApp = await this.initExpress();
-        this.initSentry(expressApp);
+        await this.initExpress();
```
joaoviana, line 228: happy to have this instead: 
```ts
const expressApp = express();
await this.initSentry(expressApp);
await this.initExpress(expressApp);
```
And then make the change to `initExpress` to accept a param

owlas: I like this ^^^ the ordering of the routes is more explicit - also explain inline why it must be in this order

joaoviana: Done!

owlas (approved):

Issue #9967, opened 2024-05-03
y-axis formatting has been removed (for table calculations)


TuringLovesDeathMetal:
I just ran into a similar issue (I think?) 

The y-axis labels seems to be showing up as numbers, not percentages 
<img width="1357" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/0a713346-06c2-4ac2-ac1a-b3b2622962a2">

https://lightdash-demo-thyme.onrender.com/share/9M-lkrnJfq1DoxgE7Yb3I

_Originally posted by @TuringLovesDeathMetal in https://github.com/lightdash/lightdash/issues/9904#issuecomment-2082091736_

PR #10002, opened 2024-05-07
fix: apply formatting to table calculations in axes

magnew:
Closes: #9967 

### Description:

Apply formatters to table calculations. We were skipping formatting because they aren't considered fields. I think these days they are more like fields and we *could* clean up some of these checks. But I think this is the right fix for this bug. 

### Reviewer actions

### DIFF
--- a/packages/common/src/types/field.ts
+++ b/packages/common/src/types/field.ts
@@ -167,7 +167,7 @@ export interface TableCalculationField extends Field {
 // This type check is a little fragile because it's based on
 // 'displayName'. Ideally these would all have fieldTypes.
 export const isTableCalculation = (
-    item?: Item | AdditionalMetric | TableCalculationField,
+    item: Item | AdditionalMetric | TableCalculationField,
 ): item is TableCalculation =>
     item
         ? !('binType' in item) &&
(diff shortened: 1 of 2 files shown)


packages/common/src/types/field.ts
```diff
@@ -167,7 +167,7 @@ export interface TableCalculationField extends Field {
 // This type check is a little fragile because it's based on
 // 'displayName'. Ideally these would all have fieldTypes.
 export const isTableCalculation = (
-    item: Item | AdditionalMetric | TableCalculationField,
```
rephus, line 170: instead of making this `item` optional, could we do:

`(axisItem && isTableCalculation(axisItem) && axisItem.format)`

This is like a bigger change that is not used anywhere else.

magnew: Sure, I'll do that. 

Just curious though, shouldn't it be generally safe to allow these typecheck functions to accept undefined? It seems like the alternative is to have checks for undefined all over the code, but they can be more centralized in these functions.

rephus: This is my personal opinion, so I'm happy to discuss if you think otherwise: 

> It seems like the alternative is to have checks for undefined all over the code, but they can be more centralized in these functions.
I agree with you , new methods should take undefined to centralize logic. 

But right now this method is used in 22 places in the code, this optional change only affects 1. So I think we should reduce the impact on these changes if possible. Are we 100% sure in all those 22 places `undefined === !tablecalculation` ?

rephus (approved):

Issue #9930, opened 2024-04-30
I want to be able to create a custom dimension using custom SQL


TuringLovesDeathMetal:
Context: https://www.notion.so/lightdash/67d3fd912f764d9eaf0f8e741d76fa00?v=805962b15e6a49958b88d1ec7dce15c4&p=f91f3d0a9c054d5ca68c5307354f2679&pm=s

### Description

There are groups and bin custom dimension types which are "commonly used" custom dimensions (and the shortcuts to build them are helpful). But sometimes I want to create dimensions using totally custom SQL. 

Here's an example use case:

I want to look at time between creating an account and using a bunch of different features in our product (e.g. time to first run query, time to first saved chart, etc.) and the average time for these events, each month. To do this today, I'd need to add these as additional dimensions in my .yml file, then create custom metrics which calculate the average. But, I don't have access to the .yml files (because I'm not on the data engineering team). So, I want to be able to create these dimensions in Lightdash (e.g. `DATE_DIFF(${my_table.first_saved_chart_created}, ${my_table.account_created}, DAY)`). Once I've created these dimensions, I want to then use them to create custom metrics! (e.g. the average time between creating an account and creating a saved chart, for each month this year). So, it's not just enough to be able to create these custom dimensions - I then want to be able to treat them like any other dimension. 

### Acceptance criteria

- [ ] I can create a custom dimension based on custom SQL
- [ ] The custom dimension definition can reference other dimensions from my table and any joined tables
- [ ] I can use the custom dimension to pivot (i.e. as a `column`) in a table visualization
- [ ] Only users with developer access or above can use custom SQL custom dimensions
- [ ] When I create a custom dimension using custom SQL, I need to select the `type` of my custom dimension (this should be the same types as table calculations)
- [ ] The modal for creating a custom dimension should be really similar to the table calculation modal
- [ ] There are two possible entry points to this:
    1. there’s an `+ Add` beside the dimension header in the sidebar (I think we should also do this for the `metrics` section, but we can do that separately)
  
<img width="248" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/2db2eacd-d50f-4443-9143-1bc9668a4147">

2. When you click `custom dimension` from a dimension in the sidebar, it opens a modal and you can select from the `types` : `bin` or `custom SQL`

### Documentation

add docs for custom dimensions using custom SQL to these docs: https://docs.lightdash.com/guides/how-to-create-custom-fields/

### Analytics event

[...]

PR #10014, opened 2024-05-08
feat: custom SQL dimension

ZeRego:
Closes: #9930 

### Description:

Screenshot 2024-05-08 at 13 11 34

Screenshot 2024-05-08 at 16 50 01
Screenshot 2024-05-08 at 16 49 49

#### What is working:
- add custom SQL dimension
- edit custom SQL dimension
- order by custom SQL dimension
- use custom SQL dimension in chart config
- save chart with custom SQL dimension
- backwards compatible with old URL with custom dimensions

#### What is not expected for this PR:
- reference dimensions
- filter by custom SQL dimension
- formatting options

### Reviewer actions


packages/frontend/src/utils/sortUtils.ts
```diff
@@ -94,6 +95,34 @@ export const getSortLabel = (
         }
     }
 
+    if (isCustomSqlDimension(item)) {
```
ZeRego, line 98: Show sort options based on the return type the user set for the custom dimension.

packages/frontend/src/providers/ExplorerProvider.tsx
```diff
@@ -239,6 +239,7 @@ export interface ExplorerReduceState {
         customDimension: {
             isOpen: boolean;
             isEditing?: boolean;
+            table?: string;
```
ZeRego, line 242: So we know to what table we need to add the custom SQL dimension. The Bin option gets that info from the selected dimension which is not available for SQL option.

packages/frontend/src/components/Explorer/CustomDimensionModal/index.tsx
```diff
@@ -1,431 +1,40 @@
...
-    Stack,
-    Text,
-    TextInput,
-    Title,
-} from '@mantine/core';
-import { useForm } from '@mantine/form';
-import { IconX } from '@tabler/icons-react';
-import { useEffect, useMemo } from 'react';
-import useToaster from '../../../hooks/toaster/useToaster';
+import { isCustomBinDimension, isDimension } from '@lightdash/common';
```
ZeRego, line 1: Moved this file content to `packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModel.tsx` 

This component only decides what custom dimension component to render.

packages/common/src/types/metricQuery.ts
```diff
@@ -38,7 +40,10 @@ export const getCustomDimensionId = (dimension: CustomDimension) =>
     dimension.id;
 
 export const isAdditionalMetric = (value: any): value is AdditionalMetric =>
-    value?.table && value?.name && !value?.fieldType && !value.binType;
+    value?.table &&
+    value?.name &&
+    !value?.fieldType &&
+    !isCustomDimension(value);
```
ZeRego, line 47: use `isCustomDimension` instead of checking for `binType` property

packages/common/src/types/field.ts
```diff
@@ -170,7 +203,7 @@ export const isTableCalculation = (
     item: Item | AdditionalMetric | TableCalculationField,
 ): item is TableCalculation =>
     item
-        ? !('binType' in item) &&
+        ? !isCustomDimension(item) &&
```
ZeRego, line 206: use `isCustomDimension` instead of checking for `binType` property

packages/common/src/index.ts
```diff
@@ -901,7 +900,7 @@ export function getItemMap(
     ].reduce(
         (acc, item) => ({
             ...acc,
-            [isAdditionalMetric(item) ? fieldId(item) : getItemId(item)]: item,
+            [getItemId(item)]: item,
```
ZeRego, line 903: no condition needed, getItem handles all types correctly

packages/common/src/authorization/types.ts
```diff
@@ -36,6 +36,7 @@ type Subject =
     | 'ChangeCsvResults'
     | 'ScheduledDeliveries'
     | 'DashboardComments'
+    | 'CustomSql'
```
ZeRego, line 39: Left as `CustomSql` so it can be used for table calculations.

packages/backend/src/database/entities/savedCharts.ts
```diff
@@ -190,6 +191,11 @@ export type DbSavedChartCustomDimensionInsert = Omit<
     custom_range: string | null;
 };
 
+export type SavedChartCustomDimensionsTable = Knex.CompositeTableType<
```
ZeRego, line 194: This table definition was missing.

packages/backend/src/queryBuilder.ts
```diff
@@ -756,14 +783,16 @@ export const buildQuery = ({
 
     const sqlSelect = `SELECT\n${[
         ...dimensionSelects,
-        ...(customDimensionSql?.selects || []),
+        ...(customBinDimensionSql?.selects || []),
```
owlas, line 786: Does the API have to select the dimensions as well as provide them as custom dimensions? 

That's how it works for custom metrics I believe:

- customDimensions: where you declare custom dimensions
- dimensions: where you select dimensions (including ones on the explore and custom dimensions)

packages/backend/src/services/CsvService/CsvService.ts
```diff
@@ -710,6 +711,21 @@ export class CsvService extends BaseService {
             throw new ForbiddenError();
         }
 
+        if (
```
owlas, line 714: how come this has to be declared here too? Isn't there a single run query function to check for permissions.

packages/backend/src/services/GdriveService/GdriveService.ts
```diff
@@ -68,6 +69,23 @@ export class GdriveService extends BaseService {
             throw new ForbiddenError();
         }
 
+        if (
```
owlas, line 72: is this also true for schedules?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1501,6 +1526,18 @@ export class ProjectService extends BaseService {
             throw new ForbiddenError();
         }
 
+        if (
+            metricQuery.customDimensions?.some(isCustomSqlDimension) &&
```
owlas, line 1530: concerned about how many times we have to write this down

packages/common/src/types/field.ts
```diff
@@ -88,17 +88,50 @@ export type BinRange = {
     from: number | undefined; // first range has from undefined
     to: number | undefined; // last range has to undefined
 };
-export interface CustomDimension {
+
+export enum CustomDimensionType {
```
owlas, line 92: type changes make sense! nice work

packages/backend/src/database/entities/savedCharts.ts
```diff
@@ -251,3 +257,27 @@ export type DBFilteredAdditionalMetrics = Pick<
...
     >;
+
+export const SavedChartCustomSqlDimensionsTableName =
+    'saved_queries_version_custom_sql_dimensions';
+
+export type DbSavedChartCustomSqlDimension = {
+    saved_queries_version_custom_sql_dimensions_uuid: number;
+    saved_queries_version_id: number;
+    id: string;
+    name: string;
```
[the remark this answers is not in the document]
ZeRego: yes

packages/backend/src/database/entities/savedCharts.ts
```diff
@@ -251,3 +257,27 @@ export type DBFilteredAdditionalMetrics = Pick<
...
+    table: string;
```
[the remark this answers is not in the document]
ZeRego: To add it under the correct table in the sidebar

packages/backend/src/queryBuilder.ts
ZeRego: No. Custom dimensions work like table calculations. If they are present in the metricquery they are applied to the sql :/

packages/backend/src/services/CsvService/CsvService.ts
ZeRego: > Isn't there a single run query function to check for permissions.

no :(

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: me too. Project service needs to be refactored and query logic extracted from it so it can be used across multiple services.

packages/backend/src/services/GdriveService/GdriveService.ts
ZeRego: this seemed to be the only one that receives a metricquery from the API instead of a dashboard/chart uuid.

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -849,16 +894,31 @@ export class SavedChartModel {
...
-                        table: cd.table,
-                        binType: cd.bin_type,
-                        binNumber: cd.bin_number,
-                        binWidth: cd.bin_width,
-                        customRange: cd.custom_range,
-                    })),
+                    customDimensions: [
+                        ...(
+                            customBinDimensionsRows || []
+                        ).map<CustomBinDimension>((cd) => ({
```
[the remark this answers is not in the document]
ZeRego: nop, thats in saved under `columnOrder` line 851

packages/backend/src/database/entities/savedCharts.ts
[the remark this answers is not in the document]
owlas: It might be defined across many tables right?

packages/backend/src/models/SavedChartModel.ts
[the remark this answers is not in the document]
owlas: What is the `order` field we have on the underlying table? Is order stored there or on another order table?

packages/backend/src/queryBuilder.ts
owlas: Does that make things more complicated than they need to be? Would it be simpler if they worked like custom metrics?

packages/backend/src/services/CsvService/CsvService.ts
owlas: a lot of work from what you've seen here?

packages/backend/src/services/ProjectService/ProjectService.ts
owlas: Do you think there are places here that might stop viewers from just "viewing" a chart with custom dimensions? It should only stop people if they're able to directly submit a metricQuery themselves

packages/backend/src/queryBuilder.ts
ZeRego: > Does that make things more complicated than they need to be?

It makes the code simpler actually but I agree the UX would be better if it worked like custom metrics. 
I don't think we should fix the custom dimension UX in this PR.

packages/backend/src/database/entities/savedCharts.ts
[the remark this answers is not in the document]
ZeRego: Not sure what you mean. 
We can define custom metrics under any of the explore tables.

packages/backend/src/models/SavedChartModel.ts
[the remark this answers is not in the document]
ZeRego: tech debt shenanigans. Instead of saving the column's order in the `saved query version` table we save the order index in the fields/custom dimension/table calculation tables.

packages/backend/src/services/CsvService/CsvService.ts
ZeRego: yes, a bit. There are multiple chains of functions in project service which I can't keep track of to have the full picture.

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: Yes. Underlying data, explore from here, drill by will throw errors as they generate new metric queries. 
We will have to work on a better UX in such cases. example: remove custom SQL dimensions if they have no permission instead of throwing an error.

owlas: Are the metric queries generated client side for these interactions?

packages/backend/src/database/entities/savedCharts.ts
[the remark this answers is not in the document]
owlas: When you click +add dimension, it could reference any of the tables right, so it doesn't live on one table alone.

e.g. `${table1.dim1} + ${table2.dim2}`

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: yes, client side. We would have to change the API to build them on the backend to make the underlying data and drill by work. Explore from here might be harder unless we create a "created from X" and then make sure they can only use the custom dimensions from that saved chart and not new ones.

packages/backend/src/database/entities/savedCharts.ts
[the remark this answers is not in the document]
ZeRego: correct, we will support them referencing for any table from that explore ( not included in this PR ). This value is only used to place the custom dimension in the UI and not for SQL generation.

packages/backend/src/services/ProjectService/ProjectService.ts
owlas: Yeah good point. Looks like underlying data should be able to work easily. But drill by (which is basically explore from here) and explore from here, are not going to work without much more complex permissions.

packages/backend/src/database/entities/savedCharts.ts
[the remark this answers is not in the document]
owlas: Can we remove it and have it live with custom metrics "outside" of the tables?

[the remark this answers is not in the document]
owlas: 🌴 oh dear. ok let's just make it extremely clear that it's only for the UI. I'm worried that devs are going to get misled by it

owlas (approved):

Issue #10008, opened 2024-05-08
If I update a filter value on a saved chart, it doesn't update in the export


TuringLovesDeathMetal:
### Description

If I create a saved chart with a filter, it doesn't update the filter int eh export if I change teh filter. This means I'm exporting data with values that don't match my filtered values. 

### Steps to Reproduce the Bug or Issue

1. Open a saved chart: https://analytics.lightdash.cloud/share/XB05qBja7g3_YNCvjWSOc
2. Click `export` from teh chart panel and export the results to a .csv 
3. See that the values in the export match the values in the filters/chart
4. Click `edit chart`
5. Update the filter value to something else (e.g. change the filter value to `comment_created` from the example above)
6. save the chart
7. click `export` from the chart panel and export the results to a .csv
8. See that the values in the export _**do not**_ match the values in the filters/chart. Instead, they match the previous filter values. 

### Cloud or self-hosting

cloud

almeidabbm:
**Steps to reproduce:**

1. Create chart with filter e.g. `event = song_played` & save
2. Edit chart, add another filter
3. Change root filter group to `Any`
4. Remove new filter
5. Save changes
6. Export CSV -> results will be correct, e.g. filtered to `song_played`
7. Edit chart
8. Change filter value
9. Save
10. Export CSV -> results will be incorrect, still filtered to previous filter value


https://github.com/lightdash/lightdash/assets/22939015/c1f86b2f-445c-4f0d-b8c5-cecd581beb66

PR #10019, opened 2024-05-08
fix: don't destructure root filter groups as they can be 'any' now

almeidabbm:
Closes: #10008

### Description:

The simplified filters form update function was always de-structuring the root filter groups into the new filters.
Since root filters can now be OR type, it would mean that we could end up with a root filter group having both `and` and `or`. The fix was to remove the de-structuring and just add the rules for the simplified form in a AND group - note: the simplified form only shows when theres only 1 filter rule or there's no subgroups, so it can always be AND in that case.

Before:

https://github.com/lightdash/lightdash/assets/22939015/412dcdf4-e8fc-46b7-8f0e-311c337761b5

After:

https://github.com/lightdash/lightdash/assets/22939015/0b2436bd-bf34-4ca6-a6a7-327162c1f533

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/index.tsx
+++ b/packages/frontend/src/components/common/Filters/index.tsx
@@ -1,8 +1,8 @@
 import {
     addFilterRule,
     deleteFilterRuleFromGroup,
-    getFilterRulesByFieldType,
     getFiltersFromGroup,
+    getInvalidFilterRules,
     getTotalFilterRules,
     hasNestedGroups,
     isAndFilterGroup,
@@ -49,14 +49,8 @@ const FiltersForm: FC<Props> = ({ filters, setFilters, isEditMode }) => {
     }, [fieldsMap]);
 
     const totalFilterRules = getTotalFilterRules(filters);
-    const { invalid: invalidFilterRulesPerType } = getFilterRulesByFieldType(
-        fields,
-        totalFilterRules,
-    );
-
-    const hasInvalidFilterRules = Object.values(invalidFilterRulesPerType).some(
-        (arr) => arr.length > 0,
-    );
+    const invalidFilterRules = getInvalidFilterRules(fields, totalFilterRules);
+    const hasInvalidFilterRules = invalidFilterRules.length > 0;
 
     const showSimplifiedForm: boolean =
         totalFilterRules.length < 2 && !hasNestedGroups(filters);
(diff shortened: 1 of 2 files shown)


TuringLovesDeathMetal (approved):
nice find! looks good to me :)

packages/frontend/src/components/common/Filters/index.tsx
```diff
@@ -49,18 +49,17 @@ const FiltersForm: FC<Props> = ({ filters, setFilters, isEditMode }) => {
     }, [fieldsMap]);
 
     const totalFilterRules = getTotalFilterRules(filters);
-    const {
-        valid: validFilterRulesPerType,
-        invalid: invalidFilterRulesPerType,
-    } = getFilterRulesByFieldType(fields, totalFilterRules);
+    const { invalid: invalidFilterRulesPerType } = getFilterRulesByFieldType(
```
joaoviana, line 52: this means we can remove the `valid` return from `getFilterRulesByFieldType`?

joaoviana:
One issue I've found with this change (unrelated to the problem, but a consequence of the changes)

See this PR: https://github.com/lightdash/lightdash/pull/9696 and the demo in https://github.com/lightdash/lightdash/pull/9696#issue-2233665284

Steps to reproduce: 
- Create chart like the demo
- Add same filters like the demo
- Add table calculation
- Use that table calculation as a filter 
- Remove table calculation
- See `tried to reference...` 
- Don't remove warning just yet(!)
- Keep deleting other filters
- The warning disappears. 
- This causes validation errors because it suddenly disappears and the user doesn't manually delete it. This was necessary when we rendered a different type of grouping on filters. I suggest looking into this validation error disappearing and checking if it's not stored in the data somewhere
I was able to verify that after deleting most filters and leaving one it didn't store the invalid reference filter: 
image


Could you check if you're able to reproduce the same?

Here is a screen-recording of what I tried in your branch: 

https://github.com/lightdash/lightdash/assets/7611706/3692c1fa-69db-4769-9a4f-784d14d8d1e0

joaoviana:
> One issue I've found with this change (unrelated to the problem, but a consequence of the changes)
> 
> See this PR: #9696 and the demo in [#9696 (comment)](https://github.com/lightdash/lightdash/pull/9696#issue-2233665284)
> 
> Steps to reproduce:
> 
> * Create chart like the demo
> * Add same filters like the demo
> * Add table calculation
> * Use that table calculation as a filter
> * Remove table calculation
> * See `tried to reference...`
> * Don't remove warning just yet(!)
> * Keep deleting other filters
> * The warning disappears.
> * This causes validation errors because it suddenly disappears and the user doesn't manually delete it. This was necessary when we rendered a different type of grouping on filters. I suggest looking into this validation error disappearing and checking if it's not stored in the data somewhere
>   I was able to verify that after deleting most filters and leaving one it didn't store the invalid reference filter:
> 
> image
> Could you check if you're able to reproduce the same?
> 
> Here is a screen-recording of what I tried in your branch:
> 
>  table-calculation-reference-issue.mov

This has been verified together that it doesn't throw a validation error

joaoviana (approved):

Issue #3554, opened 2022-10-23
Support For AWS Athena As Source


IstvanM:
### Description

Support for Athena 

### What problem does this solve?

Many startups don't have the budget to run Databricks or Snowflake as their Data Lake solution. Athena is currently the cheapest on the market.

TuringLovesDeathMetal:
Hey @IstvanM thanks for opening this issue!

We only support verified dbt adapters for now. You can find the list of verified adapters here: https://docs.getdbt.com/docs/supported-data-platforms#verified-adapters

Once Athena gets onto that list, we'll definitely work on adding support for it in Lightdash!

TuringLovesDeathMetal:
@all-contributors add @IstvanM for ideas

IstvanM:
Thank you very much!!! This is great!!! Sorry I was not responding fast enough 

🚀 🚀 🚀

jankatins:
Still relevant

nicor88:
The only blocker to make this happen is the verification of [dbt-athena-community](https://github.com/dbt-athena/dbt-athena) (the adapter with the most recent version of dbt).

Related to https://github.com/dbt-labs/dbt-core/discussions/7286 
Specifically the process of verifying a community-based adapter seems a bit unclear.

mrshu:
@nicor88 I guess the verification procedure has been settled now somewhat?

https://docs.getdbt.com/guides/dbt-ecosystem/adapter-development/7-verifying-a-new-adapter

nicor88:
@mrshu the verification process is still quite unclear, as in the case of athena the adapter is community based - and all  the verified adapters (e.g. see trino), have a company behind.

mrshu:
Thanks @nicor88. Would you estimate it might ever happen?

nicor88:
@mrshu Based on the conversations that I had with dbt seems like a dbt community adapter cannot be verified. Still, they are in the process of considering adding trusted adapters - that means that if Lightdash consider the support of "trusted" adapters and if dbt-athena becomes a trusted adapter, then this integration will be possible.

nicor88:
@TuringLovesDeathMetal `dbt-athena-community` became a trusted adapter: https://docs.getdbt.com/docs/trusted-adapters wondering if you plan to support not only verified but also trusted adapters.

cc @owlas

alfredHerdwatch:
would love to have this integration, would make lightdash an easy choice for us if it works with athena

[...]

Issue #4940, opened 2023-03-28
Athena integration


nicor88:
### Description
Athena is a query engine from AWS (based on presto/trino). It works with different type of data stored in S3, but with the recent integration  of Iceberg  (that allow full crud operation in the data lake) - it allows to build a serverless lakehouse easily if combined with [dbt-athena](https://github.com/dbt-athena/dbt-athena). 
I know that only "verified adapter are supported", but we can make dbt-athena compatible with [lightdash](https://github.com/lightdash/lightdash) if necessary.

### What problem does this solve?

Using Athena with lightdash

jankatins:
#789 seems to be a duplicate of this and has 3 more upvotes. https://github.com/lightdash/lightdash/issues/3554 is also probably a duplicate.

nicor88:
@jankatins we can close this one, I can put me message in the https://github.com/lightdash/lightdash/issues/3554

Specifically I think that the only "blocker" to make athena as source happening is the verification process, see my "answer" [here](https://github.com/dbt-labs/dbt-core/discussions/7286#discussioncomment-5583762)

Issue #789, opened 2021-11-09
Support AWS Athena


owlas:
_Originally posted by @owlas in https://github.com/lightdash/lightdash/issues/371#issuecomment-964516130_

mrshu:
@owlas would you mind briefly commenting on what this would entail? It would be very interesting for our usecase.

Thanks!

jankatins:
I had a *quick* look at this: seems this file has to be created for athena: https://github.com/lightdash/lightdash/blob/main/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts (+ tests, etc)

mrshu:
Thanks @jankatins, I guess it shouldn't be that difficult then.

jankatins:
Yes, it's still relevant.

jankatins:
https://github.com/lightdash/lightdash/issues/4940 and https://github.com/lightdash/lightdash/issues/3554 seem to duplicate this issue and have more information -> this can probably be closed.

mrshu:
@jankatins this can indeed be closed I believe.

TuringLovesDeathMetal:
Closing this - we only support dbt verified adapters. Once Athena is verified, we will support it!

ppriyranjan:
I see that Athena is verified adaptor now on dbt, https://docs.getdbt.com/docs/trusted-adapters
Can we open this issue?

PR #10020, opened 2024-05-08
feat: Add dbt-athena as a supported adapter

hhhonzik:
Closes: #3554 #789 #4940

### Description:

I've recently added a [custom adapter for Starrocks](https://github.com/hhhonzik/lightdash) which seems to be working well. With some minor time investment i was able to add **dbt-athena** support as well. If there is a will from contributors to merge this, i can cover the adapter with tests to ensure it's all ready to merge.

Otherwise feel free to close this!

Screenshot 2024-05-08 at 22 02 57

### Reviewer actions


TuringLovesDeathMetal:
Note to add docs for this once it's reviewed!

packages/backend/src/dbt/profiles.ts
```diff
@@ -174,6 +174,24 @@ const credentialsTarget = (
...
+                target: {
+                    type: WarehouseTypes.ATHENA,
+                    // all dbt config
+                    work_group: credentials.workgroup,
+                    region_name: credentials.awsRegion,
+                    s3_staging_dir: credentials.outputLocation,
+                    schema: credentials.schema,
+                    database: credentials.database,
+                    aws_access_key_id: credentials.awsAccessKeyId,
+                    aws_secret_access_key: credentials.awsSecretKey,
```
nicor88, line 188: can this be optional? for example if I run Lightdash in a self hosted environments (e.g. ECS) I would like to leverage the underlying role used by ECS to authenticate, without hard-coding credentials anywhere, simply because any call done by container are already authenticated.

packages/backend/src/dbt/profiles.ts
```diff
@@ -174,6 +174,24 @@ const credentialsTarget = (
...
+                },
+                environment: {
+                    [envVar('AWS_ACCESS_KEY_ID')]: credentials.awsAccessKeyId,
+                    [envVar('AWS_SECRET_KEY')]: credentials.awsSecretKey,
```
nicor88, line 192: the right ENV variable should be AWS_SECRET_ACCESS_KEY, see [this](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
```diff
@@ -0,0 +1,320 @@
...
+}: TableInfo) => `SELECT table_catalog
+                , table_schema
+                    , table_name
+                    , column_name
+                    , data_type
+            FROM ${database}.information_schema.columns
+            WHERE table_catalog = lower('${database}')
+                AND table_schema = lower('${schema}')
+                AND table_name = lower('${table}')
+            ORDER BY 1, 2, 3, ordinal_position`;
```
nicor88, line 67: this query is easy, but it's really really slow on scale. I've seen this query running ~2/3 minutes.
What we do in dbt-athena is to use glue apis. Also for example JDBC athena drivers uses the same concept, and in tools like Redash is possible to flag the usage of Glue APIs instead of using information_schema.

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
```diff
@@ -0,0 +1,320 @@
...
+
+        this.connectionOptions = {
+            outputLocation: credentials.outputLocation,
+            workgroup: credentials.workgroup,
+            database: credentials.database,
+            extraBotoArgs: credentials.extraBotoArgs,
+        };
+    }
+
+    private async checkQueryExequtionStateAndGetData(
```
nicor88, line 132: just a thought, isn't there any JS library that takes care of the below? for example in python there is PyAthena, that pretty much take care of all the operations around firing a query and returning the result. I'm not really familiar with the JS world, but did you do a quick search to see if we can reusable some library? it will definitely reduce complexity.

packages/backend/src/dbt/profiles.ts
hhhonzik: They are, in fact that's how i was testing the build. The only thing needed is `AWS_SDK_LOAD_CONFIG=true` as environment variable.

packages/cli/src/dbt/targets/athena.ts
```diff
@@ -0,0 +1,82 @@
...
+        },
+        catalog: {
+            type: 'string',
+        },
+        work_group: {
+            type: 'string',
+            nullable: true,
+        },
+    },
+    required: ['type', 'schema', 's3_staging_dir', 'region_name'],
```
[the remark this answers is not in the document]
hhhonzik: I took the required parameters from [dbt-athena adapter docs](https://github.com/dbt-athena/dbt-athena?tab=readme-ov-file#configuring-your-profile). I wanted to keep consistent config with the adapter. But happy to change that, it can be derived from work group afterall.

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
hhhonzik: There's [athena-express](https://www.npmjs.com/package/athena-express) - i would prefer that, but didn't want to add too many dependencies. I'll take this as a green light and use it :)

packages/cli/src/dbt/targets/athena.ts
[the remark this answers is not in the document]
nicor88: You are right, I checked the adatper implementation and `is_work_group_output_location_enforced` is used only for the data location, not for the staging result. As it is now for consistency we should do as you suggested, but this behvior can be changed in the future. In case of enforced workgroup configuration the staging location it's not necessary. I can't resolve comments, but this comment it's resolved on my side.

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
nicor88: Perfect, just double check with the lightdash team and let's check if `[athena-express](https://www.npmjs.com/package/athena-express)` is well mantained. I see that it was published the last time 2 years ago, that is not a good indicator tbh.

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
nicor88: @hhhonzik I'm happy to have a chat with you to change the behvior of the above, and make the retrival of metadata faster. The API to use should be simply glue `get_tables` https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue/client/get_tables.html for example in boto3. 

```
const getTables = (databaseName) => {
  const client = new GlueClient({});

  const command = new GetTablesCommand({
    DatabaseName: databaseName,
  });

  return client.send(command);
};
```
that supports only DatabaseName as filter, means that to to filter by a specific table we must do in memory.

packages/backend/src/dbt/profiles.ts
nicor88: Thanks for the clarification. So if I get right `AWS_SDK_LOAD_CONFIG ` should be enough when running from ECS or EKS/pods with authentication via roles right?

packages/cli/src/dbt/targets/athena.ts
```diff
@@ -0,0 +1,82 @@
...
+    WarehouseTypes,
+} from '@lightdash/common';
+import { JSONSchemaType } from 'ajv';
+import betterAjvErrors from 'better-ajv-errors';
+import { ajv } from '../../ajv';
+import { Target } from '../types';
+
+export type AthenaTarget = {
+    type: 'athena';
+
```
SimonGodefroid, line 13:
```suggestion
```

packages/common/src/compiler/translator.ts
```diff
@@ -66,6 +66,8 @@ const convertTimezone = (
             return timestampSql;
         case SupportedDbtAdapter.TRINO:
             return timestampSql;
+        case SupportedDbtAdapter.ATHENA:
```
SimonGodefroid, line 69:
```suggestion
        case SupportedDbtAdapter.TRINO:
        case SupportedDbtAdapter.ATHENA:
                    return timestampSql;
```

packages/common/src/utils/timeFrames.ts
```diff
@@ -276,6 +276,7 @@ const warehouseConfigs: Record<SupportedDbtAdapter, WarehouseConfig> = {
     [SupportedDbtAdapter.POSTGRES]: postgresConfig,
     [SupportedDbtAdapter.DATABRICKS]: databricksConfig,
     [SupportedDbtAdapter.TRINO]: trinoConfig,
+    [SupportedDbtAdapter.ATHENA]: trinoConfig, // Athena uses the same date functions as Trino
```
SimonGodefroid, line 279: Some day you'll find+replace `trino` for `whatever` and this will break 😄

packages/frontend/src/components/ProjectConnection/WarehouseForms/AthenaForm.tsx
```diff
@@ -0,0 +1,150 @@
+import { Anchor, NumberInput, Stack, TextInput } from '@mantine/core';
+import { type FC } from 'react';
```
SimonGodefroid, line 2:
```suggestion
import type { FC } from 'react';
```

packages/backend/src/dbt/profiles.ts
hhhonzik: Yup, I've added this to `.env` file, but it should definitely be mentioned in the docs.

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
```diff
@@ -0,0 +1,320 @@
...
+
+        const state = response.QueryExecution?.Status?.State;
+
+        if (
+            state === QueryExecutionState.QUEUED ||
+            state === QueryExecutionState.RUNNING
+        ) {
+            // In my case, queries run no faster than 800-900ms, which is why I set a 1000ms timeout
+            await new Promise<void>((res) => {
+                setTimeout(() => res(), 1000);
```
nicor88, line 146: Could we make the timeout configurable?
Also let me get it right, if a query takes more than 1 second will you timeout?

magnew (changes requested):
@hhhonzik, very cool. Thanks for posting this. I have a few questions to get it shipped. Bear with me, I am not a deep expert in Athena so there might be something I am not understanding 😄 

1. There is some confusion in my mind (and in this code) about the distinction between `database`, `schema` and `catalog`. It seems like Athena typically treats `database` and `schema` as the same, is that right? If so, can we choose one and require it and ignore the other? If they are subtly different and both are needed, which should we require and what do we do with the other? Also, the UI seems to send whatever is entered in the schema input box as `catalog`. Should it be sending that as schema or database? Can we hardcode catalog or does it need an input?

2. It looks like `AWS_SDK_LOAD_CONFIG` allows us to pick up role-based permissions from the aws credentials when using the lightdash CLI because you can specify the profile there. But I don't think that will work in the UI since there is no profile specified. So as far as I understand that means you can only set up the UI with a user that doesn't need to assume a role to use Athena. Does that sound right? If we want to support roles in the UI, should we add an advanced option to provide an amazon profile name (like you can in the DBT profile). 

FWIW, we can decide to only support a subset of possible AWS credential configurations for now. What kind of profile definitions do you expect this to work for for now?

Thanks again!

packages/warehouses/src/warehouseClients/AthenaWarehouseClient.ts
nicor88: Ah I got it, it's the waited time that determine how often we check when the query result is ready. In dbt-athena we call this **poll_interval**.

My suggestion would be to make the value configurable, with a default value of 0.5.
I've seen very optimized queries running below 1sec.
Also, as suggested by @magnew let's add a comment please, to give more context 🙏🏻

nicor88:
@magnew thanks for the review, as I have a clear interest in having this feature in, and as dbt-athena maintainer I could try to address few of the 2 questions.
1. Athena it's based on Trino (but with some different interface to communicate to it). Said so Athena has few concepts:
   * catalog: the most important catalog it's **awsdatacatalog**, that is pretty much an interface to glue catalog. It's possible to use athena as a federated system to query other sources, for example postgres. In this case the catalog name will be different. - most likely the catalog can be hardcoded, and its value should be **awsdatacatalog**. Given the fact that the user will use lightdash to query assets produced via dbt-athena, and that dbt-athena can only write to awsdatacatalog should be enough IMHO (I cannot foreseen edge cases now, but in the first iteration it's more than enough).
   * database/schema: there is no concept of schema in athena. If we pick the most relevant catalog `awsdatacatalog` (aka glue catalog) we have only the concept of database. In dbt-athena we use the keyword **schema** to refer to a database, and I believe that we should do the same in lightdash, as it's really dbt dependent
2. Regarding permissions I see only 2 most relevant use cases:
  a. The user provided a set of AWS_ credentials (most likely coming from an AWS IAM user). In this scenario the user it's required to provided the credentials in the UI. I believe that this PR address that properly. Specifically this will be the case in an environment where the user cannot control the containers where lightdash will be deployed (e.g. lightdash cloud)
  b. The user do not provide a set of AWS credentials, because for example has lightdash running in its own infra. In this case it will be nice to try to pick the underlying credentials used by the environment where the user deploy lightdash. For example if you deploy lightdash in ECS you can have an ecs role that have already the AWS_ credentials set, and lightdash could leverage that. The same apply to K8S with a role attached to a container, the AWS_ credentials will be available in the underlying container.

Specifically regarding point 2, the minimum requirement that should addressed by this PR, it's the support of AWS provided credentials by the user - that cover a ligthdash cloud deployment, as a self-hosted deployment (in this case, I would like to avoid to provide any credentials, but we can figure this out later).

owlas:
Hey @nicor88 - I'm going to close this PR. Not because we don't want to merge it but because we don't currently have bandwidth and it isn't a priority for us right now. 

To make communication with the team easier, we're making the open pull requests represent priority #1 for the team.

Let's keep the conversation going here and tag me in the slack community! If we have more push from the community (and when we have more bandwidth) we can revisit this!

Issue #10018, opened 2024-05-08
(Existing) big number comparisons on table calculation fields are broken


mhawkins-runa:
### Description

It looks like comparisons vs previous row on all existing big value tiles that use a table calculation field are broken, as they all show "n/a" vs previous row, with the reason apparently being that the previous row is not a number, even though it is a number. It's possible that this only applies to currency-formatted calculations, although I'm not sure.

I can fix these individually by changing the format of the table calc to default, saving the calc, then changing it back to currency and saving it again. This gets rid of the issue, even though all I've done is put everything back to how it was before.

I cannot reproduce this issue for any **new** big value charts. It only affects existing ones, and specifically those that are based on a table calculation.

### Steps to Reproduce the Bug or Issue

Look at any **existing** big value chart that meets the following criteria:
- Selected field is a table calculation
- Table calculation is formatted as currency
- Big value chart includes comparison vs previous row

### version

v0.1089.1

### Cloud or self-hosting

cloud

mhawkins-runa:
Actually, the fix is a bit quicker than what I've described above. All I need to do is open up the edit window for the table calculation and re-save it without any changes, and that fixes the issue.

mhawkins-runa:
It's also not just currency-formatted calculations. I also have percentages with the same issue.

PR #10031, opened 2024-05-10
fix: fix undefiled type table calculation on bignumber

rephus:
Closes: #10018

### What was the issue: 

new table calculations now have types, but existing table calculations has type as undefined. For undefined table calculations, we assume they are strings (for backwards compatibility) but this assumption is not correct on bignumber.

### Description:

Screenshot from 2024-05-10 09-08-10

Before:

Screenshot from 2024-05-10 09-02-22

Screenshot from 2024-05-10 08-55-23

After:

Screenshot from 2024-05-10 08-58-51

Screenshot from 2024-05-10 09-02-31

For strings: 

Screenshot from 2024-05-10 09-03-43

(new check type) For table calculation with `string` type:

Screenshot from 2024-05-10 09-05-20

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useBigNumberConfig.ts
+++ b/packages/frontend/src/hooks/useBigNumberConfig.ts
@@ -268,8 +268,12 @@ const useBigNumberConfig = (
     }, [item, firstRowValueRaw, selectedField, bigNumberStyle, resultsData]);
 
     const unformattedValue = useMemo(() => {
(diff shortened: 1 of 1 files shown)


rephus:
The change seems simple, but it can affect other places, I'm trying to run e2e tests to ensure everything works as expected. I'll also be doing some tests manually.

almeidabbm (approved):

Issue #10030, opened 2024-05-09
The week being shown in results table doesn't match with the week shown on the chart 


hamzahc1:
### Description

![image](https://github.com/lightdash/lightdash/assets/14341285/1663f619-ea28-426c-9734-e0cd27468941)

Charts and results tables are off by one day. 

Reported here: 
https://lightdash-community.slack.com/archives/C03MG2VTR08/p1715294687927269

### Steps to Reproduce the Bug or Issue

Reported here: 
https://lightdash-community.slack.com/archives/C03MG2VTR08/p1715294687927269
https://lightdash.slack.com/archives/C06PAQCH932/p1715290354141549

### Cloud or self-hosting

cloud

TuringLovesDeathMetal:
@magnew I wonder if this is related to `start of week` set to Monday vs. Sunday?

PR #10035, opened 2024-05-10
fix: week axes not respecting timezone correctly

magnew:
Closes: #10030

### Description:

In some timezones (US) the week axis wasn't showing the right dates. We were keeping timezone information, then converting to local time. This just forces the times to be treated as utc. 

Jose actually suggested this in code review, but I didn't make the change because I was misunderstanding a bit of a subtlety with dayjs. Since the values all have a timezone included, converting them with dayjs(value).utc() is the same as dayjs.utc(value), which is how you would tell dayjs to treat a value as already utc. What we have now is a bit semantically confusing (we are not really converting them), but it does the right thing. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -974,7 +974,7 @@ const getEchartAxes = ({
                 case TimeFrames.WEEK:
                     axisConfig.axisLabel = {
                         formatter: (value: any) => {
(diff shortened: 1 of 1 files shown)


magnew:
> The change seems simple, but it can affect other places, I'm trying to run e2e tests to ensure everything works as expected. I'll also be doing some tests manually.

Thank you! I did a fair amount of testing, but let me know if there is anything you'd like me to check particularly

rephus (approved):
Tested, LGTM

PR #10059, opened 2024-05-13
feat: enable sentry backend profiling

joaoviana:
Closes: N/A

### Description:

Follows this guide: https://docs.sentry.io/platforms/javascript/guides/node/profiling/

Enables profiling of 20% of the samples.

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -1,6 +1,7 @@
 import { LightdashMode, SessionUser } from '@lightdash/common';
 import { NodeSDK } from '@opentelemetry/sdk-node';
 import * as Sentry from '@sentry/node';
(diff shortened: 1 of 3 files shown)


packages/backend/src/App.ts
```diff
@@ -502,6 +504,7 @@ export default class App {
                 }
                 return 0.2;
             },
+            profilesSampleRate: 0.2, // 20% of samples will be profiled
```
ZeRego, line 507: > Set sampling rate for profiling - this is relative to tracesSampleRate

So we are going to get 20% of 20%. Double checking this is what we want.

ZeRego (approved):

packages/backend/src/App.ts
joaoviana: yup!

PR #10062, opened 2024-05-13
Chore/remove table calculation field type

joaoviana:
Closes: 

### Reviewer actions


packages/backend/src/queryCompiler.ts
```diff
@@ -1,17 +1,14 @@
 import {
     AdditionalMetric,
-    CompiledCustomDimension,
```
joaoviana, line 3: unused imports

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3052,7 +3054,7 @@ export class ProjectService extends BaseService {
             span?.finish();
         }
 
-        const allFilterableFields: FilterableField[] = [];
+        const allFilterableFields: FilterableDimension[] = [];
         const filterIndexMap: Record<string, number> = {};
 
         allFilters.forEach((filterSet) => {
```
joaoviana, line 3060: `allFilters` are `FilterableDimension`s

packages/common/src/types/field.ts
```diff
@@ -178,17 +178,16 @@ export enum TableCalculationType {
...
 export type TableCalculation = {
     index?: number;
     name: string;
+    // This is a unique property of the table calculation
     displayName: string;
     sql: string;
     format?: CustomFormat;
     type?: TableCalculationType;
+    // TODO: Check if this needs to be required - it's not in the metricQuery compiled
+    fieldType?: FieldType.TABLE_CALCULATION;
```
joaoviana, line 190: ❓ I believe I should not add this here - it was only in `TableCalculationField`

packages/common/src/types/field.ts
```diff
@@ -479,6 +446,33 @@ export interface Metric extends Field {
     requiredAttributes?: Record<string, string | string[]>; // Required attributes for the dimension this metric is based on
 }
 
+export const isFilterableDimension = (
```
joaoviana, line 449: Moved this here so we could use `isMetric` in `isFilterableField`

packages/common/src/utils/filters.ts
```diff
@@ -98,10 +98,6 @@ export const getFilterGroupItemsPropertyName = (
     return 'and';
 };
 
-export const filterableDimensionsOnly = (
```
joaoviana, line 101: Not used anywhere

packages/common/src/utils/filters.ts
```diff
@@ -139,6 +135,8 @@ export const getFilterTypeFromItem = (item: FilterableItem): FilterType => {
             return FilterType.BOOLEAN;
         default: {
             return assertUnreachable(
+                // @ts-expect-error - TODO: Fix this
+                // TODO: Fix this
```
joaoviana, line 139: ❓ Need help with this - this appeared after changes to `FilterableItem` type

packages/frontend/src/components/Explorer/CustomMetricModal/FilterForm.tsx
```diff
@@ -38,36 +39,40 @@ export const FilterForm: FC<{
     const isEditMode = useExplorerContext(
         (context) => context.state.isEditMode,
     );
-    const { fieldsMap } = useFiltersContext();
+    const { itemsMap } = useFiltersContext();
 
-    const dimensions = Object.values(fieldsMap).filter(isDimension);
+    const dimensionsMap = getFilterableDimensionsFromItemsMap(itemsMap);
+    const dimensions = Object.values(dimensionsMap);
```
joaoviana, line 45: Custom metric modal filters were only using dimensions

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
```diff
@@ -323,7 +323,7 @@ export const CustomMetricModal = () => {
                             <Accordion.Panel>
                                 <FiltersProvider
                                     projectUuid={projectUuid}
-                                    fieldsMap={fieldsMap}
+                                    itemsMap={fieldsMap}
```
joaoviana, line 326: perhaps the filtering in: https://github.com/lightdash/lightdash/pull/10062/files#r1599738715 should be done before here? 
```
itemsMap={dimensionsMap}
```
?

packages/frontend/src/components/Explorer/FiltersCard/useFieldsWithSuggestions.ts
```diff
@@ -54,9 +60,10 @@ export const useFieldsWithSuggestions = ({
                 // converts table calculation to filterable table calculation
                 // which is a sub-type of Field.
                 const cals = (tableCalculations || []).reduce<
-                    TableCalculationField[]
+                    ItemsMap[string][]
                 >((acc, cal) => {
-                    const tableCalculationFilters: TableCalculationField = {
+                    // TODO: Verify if this is OK
+                    const tableCalculationFilters = {
```
joaoviana, line 66: ❓ I removed the explicit typing. what do you think?

packages/frontend/src/components/common/Filters/FilterInputs/index.tsx
```diff
@@ -246,8 +246,8 @@ export const getConditionalRuleLabel = (
 
 export const getFilterRuleTables = (
     filterRule: ConditionalRule,
-    field: FilterableField,
-    filterableFields: FilterableField[],
+    field: FilterableDimension,
```
joaoviana, line 249: Only used in Dashboard filters

packages/frontend/src/components/common/Filters/index.tsx
```diff
@@ -41,12 +43,28 @@ type Props = {
     isEditMode: boolean;
 };
 
+const getInvalidFilterRules = (
```
joaoviana, line 46: Moved this util to this component because it's only used here

packages/common/src/types/field.ts
ZeRego: correct, we don't need this 👍 
We can also remove `fieldType` from `Field` as it was added alongside `TableCalculationField`

packages/common/src/utils/filters.ts
ZeRego: Do we need to handle custom dimensions? I was hoping to do this in the other ticket.

```
const type = isCustomBinDimension ? DimensionType.STRING : isCustomSqlDimension ? item.dimensionType : item.type;
switch (type)
...
```

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
ZeRego: ideally yes

packages/frontend/src/components/Explorer/FiltersCard/useFieldsWithSuggestions.ts
ZeRego: Can we remove the loop as we don't need to convert?

packages/common/src/types/field.ts
joaoviana: Done here: https://github.com/lightdash/lightdash/pull/10062/commits/2f8ca5ed0574d660d829e607fe9a0b6c3eaeb9ce

packages/frontend/src/components/Explorer/CustomMetricModal/index.tsx
joaoviana: done here: https://github.com/lightdash/lightdash/pull/10062/commits/6ec4497e9149073f79ff1042af5a0b6aa6e14d46

packages/frontend/src/components/Explorer/FiltersCard/useFieldsWithSuggestions.ts
joaoviana: done: https://github.com/lightdash/lightdash/pull/10062/commits/0797a01809c674faf5c70decac835d9a8e0252e8
thanks for the suggestion 🙏

packages/common/src/utils/filters.ts
joaoviana: Nope, we dont... `FilterableField` does not contain `CustomDimension`s...

joaoviana: https://github.com/lightdash/lightdash/pull/10062/commits/ec23c3cb8dd1acde51d62ea9d3f4ecf62a6e4a36

This removes the reference of `itemsMap` in `FilterableField` because that was include `Field`, and field as a `type: string` (**not an enum**) - hence why that `assertUnreachable` was complaining about a `string`

joaoviana:
For me to review 
- Filters in Filter form
- Filters in Dashboards (specially availableFilters endpoint - play around with tile targets)
- Filter dashboard to...
- Creating filter rules for any
- Propagation of filter rules when exploring from here...

joaoviana:
After testing: 
- Only issue I've found was with adding Table Calculations as filters on an explore. Testing here: https://lightdash-pr-10062.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/saved/6ec076a7-314b-42cf-aa2b-fedf6b5818fa/view

ZeRego (approved):

packages/backend/src/queryBuilder.mock.ts
```diff
@@ -1036,7 +1036,7 @@ export const METRIC_QUERY_WITH_TABLE_CALCULATION_FILTER: CompiledMetricQuery = {
                         fieldId: 'calc3',
                     },
                     operator: FilterOperator.EQUALS,
-                    values: [],
+                    values: ['my value'],
```
ZeRego, line 1039: This unit test was wrong from the beginning. 
The table calculation id would not match with the filter target id (`table_calculation_{table name}`) causing the test to fall into this condition.
`if (!field) return '1=1';`

joaoviana:
Tested the following again: https://github.com/lightdash/lightdash/pull/10062#issuecomment-2111896629 and it's working as expected
- Added TC
- Added it as a filter
- Then changed its type
- Added again as a filter

Tested Dashboard filters and the `availableFilters` endpoint
Confirmed as well that I don't see table calculations' filters where they shouldn't be - for example, in Dashboards & Custom metric modal

packages/backend/src/database/migrations/20240515102618_remove_table_calculation_filter_prefix.ts
```diff
@@ -0,0 +1,26 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    const versionsWithTableCalculationFilters = await knex(
+        'saved_queries_versions',
+    )
+        .select('saved_queries_version_uuid')
+        .whereRaw("jsonb_path_exists(filters, '$.tableCalculations')");
+
+    await knex('saved_queries_versions')
```
joaoviana, line 10:
```suggestion
 if (versionsWithTableCalculationFilters.length > 0) {
    await knex('saved_queries_versions')
```

I think this should be added, what do you think? ^

ZeRego: yeah, we can add it :)

packages/common/src/types/field.ts
```diff
@@ -199,26 +195,16 @@ export type TableCalculationMetadata = {
...
-export interface TableCalculationField extends Field {
-    fieldType: FieldType.TABLE_CALCULATION;
-    type: TableCalculationType;
-    index?: number;
-    name: string;
-    displayName: string;
-    sql: string;
 }
 
 // This type check is a little fragile because it's based on
```
magnew, line 200: I think this comment is no longer relevant.

Issue #10066, opened 2024-05-14
If I delete the dimension that is used to build my custom dimension, then there is an error in Lightdash I can't get rid of.


TuringLovesDeathMetal:
### Description

If I create a custom dimension, then I delete the dimension that the custom dimension is built from in my project, I get a persistent error that there's no way to get rid of (or fix). 

Instead of just having an error that can't be fixed, we should show the custom dimension in the sidebar, with a warning/error and allow the user to delete this custom dimension. 

### Steps to Reproduce the Bug or Issue

If I add a custom metric and then remove the dimension that it is based from the yaml then it still shows the error even if we remove the dimension from the results - same thing happens if we have a custom dimension, it's not removed and doesn't show up in the missing fields, just errors

If we remove the custom dimension then it works, not sure how we want to handle this, should it appear in the missing fields?

### Cloud or self-hosting

None

PR #10070, opened 2024-05-14
fix: display missing custom metrics and dimensions

almeidabbm:
Closes: #10066 

### Description:

https://github.com/lightdash/lightdash/assets/22939015/9206e91f-ac3f-4bba-9ffa-b3f323e466db

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/index.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreTree/index.tsx
@@ -1,6 +1,9 @@
 import {
+    convertFieldRefToFieldId,
+    getAllReferences,
     getItemId,
     isCustomBinDimension,
+    isCustomSqlDimension,
     type AdditionalMetric,
     type CompiledTable,
     type CustomDimension,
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/Explorer/ExploreTree/index.tsx
```diff
@@ -71,12 +72,30 @@ const ExploreTree: FC<ExploreTreeProps> = ({
...
-        const allTables = Object.keys(explore.tables);
-        return additionalMetrics.filter(
-            (metric) => !allTables.includes(metric.table),
-        );
+        return additionalMetrics.filter((metric) => {
+            const table = explore.tables[metric.table];
+            return (
+                !table ||
+                (metric.baseDimensionName &&
+                    !table.dimensions[metric.baseDimensionName])
```
rephus, line 80: why did you change this ? I thought the problem was only on custom dimensions

packages/frontend/src/components/Explorer/ExploreTree/index.tsx
```diff
@@ -71,12 +72,30 @@ const ExploreTree: FC<ExploreTreeProps> = ({
...
+        });
     }, [explore, additionalMetrics]);
 
+    const missingCustomDimensions = useMemo(() => {
+        return customDimensions?.filter((customDimension) => {
+            const table = explore.tables[customDimension.table];
+
+            return (
+                !table ||
+                (isCustomBinDimension(customDimension) &&
```
rephus, line 91: do we also need to check if `isCustomSqlDimension` ?

packages/frontend/src/components/Explorer/ExploreTree/index.tsx
almeidabbm: yeah, ended up noticing as well that if the base dimension for a custom metric was removed the custom metric wouldn't show as missing

https://github.com/lightdash/lightdash/assets/22939015/b396e478-3739-4077-aeb0-c6d6fe3b3436

packages/frontend/src/components/Explorer/ExploreTree/index.tsx
almeidabbm: good point, I had a look at that and `customSqlDimension` but since it doesn't have a base dimension and it's only sql I couldn't do it like this, probably can only do it after compiling?

almeidabbm: I believe I was able to do this in the latest [commit](https://github.com/lightdash/lightdash/pull/10070/commits/f8dd428263461d83a9f87cf9f99776c92496bbc1): 


https://github.com/lightdash/lightdash/assets/22939015/dfc2232b-fb50-4289-baca-4f5ebed702a7

rephus (approved):

PR #10076, opened 2024-05-14
fix: fix type error or undefined param.encode.x

joaoviana:
Closes: #9939

### Description:

[fix type error or undefined param.encode.x](https://github.com/lightdash/lightdash/commit/89eed7ce81c9b4496d81efc21a1e0f23b5b51752)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/SimpleChart/index.tsx
+++ b/packages/frontend/src/components/SimpleChart/index.tsx
@@ -180,10 +180,7 @@ const SimpleChart: FC<SimpleChartProps> = memo((props) => {
                                     // on `useEchartsCartesianConfig` to generate the tooltip
                                     if (eChartsOptions.tooltip.formatter) {
                                         const dim =
-                                            param.encode?.x &&
-                                            Array.isArray(param.encode?.x) &&
-                                            param.encode.x.length > 0 &&
-                                            param.encode?.x[0] !== undefined
+                                            param.encode?.x?.[0] !== undefined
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/SimpleChart/index.tsx
```diff
@@ -180,6 +180,9 @@ const SimpleChart: FC<SimpleChartProps> = memo((props) => {
                                     // on `useEchartsCartesianConfig` to generate the tooltip
                                     if (eChartsOptions.tooltip.formatter) {
                                         const dim =
+                                            param.encode?.x &&
+                                            Array.isArray(param.encode?.x) &&
+                                            param.encode.x.length > 0 &&
                                             param.encode?.x[0] !== undefined
```
rephus, line 186: This is good, 
Just wondering, this is the same as doing `encode.x?.[0] !== undefined` , right ? Not sure if there is any difference.

rephus (approved):

packages/frontend/src/components/SimpleChart/index.tsx
joaoviana: Thanks for the suggestion @rephus !
I've checked and seems like `encode.x` will be an array if not undefined. 
Pushed that change here: https://github.com/lightdash/lightdash/pull/10076/commits/9b535ba3b6bc64225b366eaccfec5ccaa4d75968

PR #10081, opened 2024-05-15
fix: disable chart link while editing dashboard

magnew:
### Description:

Disables the explore link in dashboard tile titles when editing a dashboard. When linking out we can't keep dashboard state and it's easy to lose dashboard changes. This makes it so users have to finish editing to explore. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
@@ -100,6 +100,8 @@ const TileBase = <T extends Dashboard['tiles'][number]>({
     const isMarkdownTileTitleEmpty =
         tile.type === DashboardTileTypes.MARKDOWN && !title;
 
+    console.log('it', titleHref);
+
     return (
         <Card
             component={Flex}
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
```diff
@@ -162,16 +162,22 @@ const TileBase = <T extends Dashboard['tiles'][number]>({
...
-                                        href={titleHref}
-                                        $hovered={titleHovered}
-                                        target="_blank"
-                                        className="non-draggable"
-                                        hidden={hideTitle}
-                                    >
-                                        {title}
-                                    </TileTitleLink>
+                                    {isEditMode ? (
+                                        <Text fw={600} fz="md">
```
joaoviana, line 166:
```suggestion
                                        <Text fw={600} fz="md" hidden={hideTitle}>
```

joaoviana (changes requested):
Could you add this prop? 
When I choose to hide a title, I want it to be hidden while on edit mode 👍

packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
magnew: Tested locally 👍

joaoviana (approved):

Issue #9985, opened 2024-05-06
Switching a date filter from month to year granularity results in incorrect SQL being used


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/1bc319a7a4ac48bd8c1e9c88e3df612f

### Steps to Reproduce the Bug or Issue

1. Add a filter for `date month`
2. Run the query and see that there are results for the current year
3. Switch the filter to be `date year`
4. See that the results are filtering for the incorrect SQL (it's still using the month on the right hand side of the equality)
5. See that the query returns no results

### Cloud or self-hosting

cloud

joaoviana:
Duplicate of: https://github.com/lightdash/lightdash/issues/7689

PR #10089, opened 2024-05-15
fix: convert year filter value to start of year when creating/changing filters

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/7689 https://github.com/lightdash/lightdash/issues/9985

### Description:

See tickets for description. 
Essentially, convert the date for a `Filter` of time interval `YEAR` to the start of the that year so `DATE_TRUNC` = filtering works as expected.

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/filters.ts
+++ b/packages/common/src/utils/filters.ts
@@ -218,14 +218,20 @@ export const getFilterRuleWithDefaultValue = <T extends FilterRule>(
                             ? defaultTimeIntervalValues[field.timeInterval]
                             : moment();
 
-                    const dateValue = valueIsDate
-                        ? formatDate(
-                              // Treat the date as UTC, then remove its timezone information before formatting
-                              moment.utc(value).format('YYYY-MM-DD'),
-                              undefined,
-                              false,
-                          )
-                        : formatDate(defaultDate, undefined, false);
+                    // If the field is a year filter, we want to convert it to the start of the year. Example: 2021-03-01 -> 2021-01-01
+                    const isYearFilter =
+                        isDimension(field) &&
+                        field.timeInterval === TimeFrames.YEAR;
+
+                    const dateValue =
+                        valueIsDate && !isYearFilter
+                            ? formatDate(
+                                  // Treat the date as UTC, then remove its timezone information before formatting
+                                  moment.utc(value).format('YYYY-MM-DD'),
+                                  undefined,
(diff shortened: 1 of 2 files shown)


magnew (commented):
I know we spent some time on this approach, but I don't think it's right. For two reasons:

1. I was testing this and it doesn't solve the whole problem. Things still break in the same way when you go from DAY filter to a MONTH filter. We kind of knew this cause we only did year in this code explicitly. But it makes me think we need a more generic approach.

2. I think maybe we are mis-diagnosing the issue. DATE_TRUNC is supposed to work with any date, not just the first of the year. I tested it in the SQL runner and it works with, for example, March 15 truncated to year

Screenshot 2024-05-16 at 14 24 15

Although, notably, it renders it as an ISO timestamp. I think something else is up here and we'll probably be better looking for the real cause.

joaoviana:
> I know we spent some time on this approach, but I don't think it's right. For two reasons:
> 
> 1. I was testing this and it doesn't solve the whole problem. Things still break in the same way when you go from DAY filter to a MONTH filter. We kind of knew this cause we only did year in this code explicitly. But it makes me think we need a more generic approach.
> 2. I think maybe we are mis-diagnosing the issue. DATE_TRUNC is supposed to work with any date, not just the first of the year. I tested it in the SQL runner and it works with, for example, March 15 truncated to year
> 
> Screenshot 2024-05-16 at 14 24 15
> Although, notably, it renders it as an ISO timestamp. I think something else is up here and we'll probably be better looking for the real cause.

On point 1: 
Yup, exactly, `DATE_TRUNC` works with any date, the problem is when the right hand-side of the condition is not YYYY-01-01

```sql
WHERE ((
  (DATE_TRUNC('YEAR', "customers".created)) = ('2017-02-01')
))
```

This will not work, because when you compare `DATE_TRUNC('YEAR', "customers".created)` to '2017-02-01', you are comparing `YYYY-01-01 (from the truncation) to 2017-02-01`, which will never be true.

joaoviana:
Not too sure on what the solution will be since it makes sense to discard this change, but we should look into: 

```sql
WHERE EXTRACT(YEAR FROM "customers".created) = EXTRACT(YEAR FROM DATE '2017-02-01')
```

joaoviana:
Closing this PR since @magnew 's point makes sense and this is not the desired solution

Issue #7564, opened 2023-10-20
You can filter by a custom sql dimension


rephus:
### Description

Note: this will only support custom SQL dimensions. Bin dimensions will be supported in https://github.com/lightdash/lightdash/issues/10092

### What problem does this solve?

Users can filter by the custom dimension.

PR #10103, opened 2024-05-17
feat: custom sql dimension filters

ZeRego:
Closes: #7564 

### Description:

Screenshot 2024-05-17 at 10 12 51

Tested:
- Add filter from sidebar
- Add filter from table header
- Add filter from table cell
- Add filter from filter select
- Add filter for unselected custom dimension ( joins necessary table )
- Add multiple filters for different types of custom dimensions

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/filters.mock.ts
+++ b/packages/common/src/utils/filters.mock.ts
@@ -1,9 +1,15 @@
 import { ConditionalOperator } from '../types/conditionalRule';
 import {
+    CustomDimensionType,
+    DimensionType,
+    type CustomSqlDimension,
+} from '../types/field';
+import {
     FilterOperator,
     type AndFilterGroup,
     type DashboardFilters,
     type FilterRule,
+    type Filters,
     type OrFilterGroup,
 } from '../types/filter';
 import type { MetricQuery } from '../types/metricQuery';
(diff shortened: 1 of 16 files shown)


rephus:
`random()` doesn't work in filters, because the random it generates in `where` is different from the one in `select`  , we should probably create a cte to fix this, but I don't think is worth it. 

image

rephus:
These custom dimension filters can't be used in dashboards, do we have plans to do it in the future ?

rephus:
I'm using `now()` as a custom sql dimension, this returns a timestamp, but I can only filter `dates`  

image

rephus:
I can't filter by `today` if custom sql dimension is `now()` with type `date` ,
 perhaps it is doing some timestamp formatting that not captured in the filter...

packages/common/src/utils/filters.mock.ts
```diff
@@ -201,3 +207,28 @@ export const expectedChartWithOverrideDashboardORFilters: MetricQuery = {
...
+export const expectedFiltersWithCustomSqlDimension: Filters = {
+    dimensions: {
+        and: [
+            {
+                id: 'uuid',
+                operator: ConditionalOperator.EQUALS,
+                target: {
+                    fieldId: 'custom-sql-dimension',
+                },
+                values: [],
```
rephus, line 229: should we add some value to test ?

rephus, line 229: should we add some value to test ?

packages/common/src/utils/filters.ts
```diff
@@ -102,7 +104,7 @@ export const getFilterGroupItemsPropertyName = (
 };
 
 export const getFilterTypeFromItem = (item: FilterableField): FilterType => {
-    const { type } = item;
+    const type = isCustomSqlDimension(item) ? item.dimensionType : item.type;
```
rephus, line 107: at this point we should probably create a `getDimensionType` method that takes a Dimension or CustomSqlDimension

ZeRego:
> Probably something out of scope, but random() doesn't work in filters, because the random it generates in where is different from the one in select , we should probably create a cte to fix this, but I don't think is worth it.

The same limitation exists with dimensions declared in YML using custom SQL. The method `NOW()` should return the same value for all the calls in the same query execution. 

> These custom dimension filters can't be used in dashboards, do we have plans to do it in the future?

We can considerate it when users present us with good use cases. Not planned at the moment.

> I can't filter by today if custom sql dimension is now() with type date ,
perhaps it is doing some timestamp formatting that not captured in the filter...

Going to investigate this as I must have missed something.

packages/common/src/utils/filters.ts
ZeRego: Was thinking the same, doing it after this PR

PR #10111, opened 2024-05-17
chore: get item type method

ZeRego:
Closes: 

### Description:

- Add a util function to get item type
- simplify switch case in sort method

### Reviewer actions


packages/common/src/utils/item.ts
```diff
@@ -43,29 +44,6 @@ export const isNumericType = (
     return numericTypes.includes(type);
 };
 
-export const isNumericItem = (
```
ZeRego, line 46: moved function bellow `getItemType`

packages/frontend/src/utils/sortUtils.ts
```diff
@@ -63,134 +62,48 @@ export const getSortLabel = (
     item: Field | TableCalculation | CustomDimension,
     direction: SortDirection,
 ) => {
-    if (isTableCalculation(item)) {
-        const type = item.type;
-        if (type === undefined)
+    const type = getItemType(item);
+    switch (type) {
```
ZeRego, line 66: We can have a single switch now.

rephus (approved):

PR #10115, opened 2024-05-20
feat: catalog page

rephus:
Closes: 

### Description:

image

### Default display:
[Sketch](https://www.figma.com/proto/rlqMIJ4pm3t8aAsiMd3CZJ/Lightdash-Product-New?node-id=2550-115315&t=eeMfJ6TrYhH4TyAU-0&scaling=min-zoom&page-id=2544%3A18282&starting-point-node-id=2550%3A115315)

- [x] By default You see a complete list of your tables , with descriptions in a column.
- [x] The descriptions should look consistent in terms of styling for now ( eg. its possible to have variation because of the markdown)
- [x] Display 'table groups' in a collapsible / expandable accordion.
- [x] If there is only one table, you should not be able to collapse / expand it
- [x] If you click a table -> You open an explore view with the table selected
- [ ] It should be clear when a table is made up of joins (by showing an icon and hover state. use a dummy icon for now)

### Search bar behaviour
[Sketch](https://www.figma.com/proto/rlqMIJ4pm3t8aAsiMd3CZJ/Lightdash-Product-New?node-id=2544-19305&t=eeMfJ6TrYhH4TyAU-0&scaling=min-zoom&page-id=2544%3A18282&starting-point-node-id=2550%3A115315)

- [x] If you use the search bar, your results will show tables that match, as well as fields/descriptions within that that match (use atree structure)
- [ ] If you click on Field X -> You open an explore view with the relevant table selected and Field X must be selected by default too

If you:

- [ ] ...search for a field that lives in a group, show the group name (like a tree structure. I want to work on this a lot more, so this doesn't have to be so refined)
- [ ] ...search for a group, I should be able to see all the fields within that group too (using the tree structure)
- [ ] ...search for field that appears in two tables because of a join, display it twice.

Implementation

- [x] New page: /catalog
- [x] No feature flag needed for now
- [x] Stying : Use pure mantine defaults and dont get too caught up trying to match the look of the wireframe. This is all going to change, so its all about functionality :)
- [x] Out of scope : Filters and Fields tab (as seen in some of the sketches)

Search API

    Implement backend search for search fields across the project, and tables across the project.
    We already have the SearchModel and SearchService containing search for fields. We need to refactor this to make the field search reusable, or simply write new SearchModels
    We must refactor out the field/table search, which currently relies on loading all explores and running in-memory regex. This could be a big performance bottleneck and doesn't provide semantic search. We want to ship this in a way that makes it very easy to iterate on the search algorithm/technology without affecting the search service / API / permission checks.
    User attributes are currently slowing down search massively, can we make that better?
    Can we index tables on project compile? For example creating tables, creating search vectors, or even embeddings to make it easier to search?
    New features
        We need syntax highlighting. Is this something we do on the backend or shall we do a frontend trick to show matches? The latter would be much easier

Open Questions

What happens when you go back? Does everything reset?

Backend request samples:
Screenshot from 2024-05-20 16-17-13
Screenshot from 2024-05-20 16-17-24
Screenshot from 2024-05-20 16-17-36

### Reviewer actions


examples/full-jaffle-shop-demo/dbt/models/tracks.yml
```diff
@@ -42,7 +42,7 @@ models:
               - HOUR_OF_DAY_NUM
               - MINUTE_OF_HOUR_NUM
       - name: timestamp_EST
-        description: "Timestamp, but converted to EST in BigQuery using DATETIME(timestamp, "America/New_York")"
+        description: "Timestamp, but converted to EST in BigQuery using DATETIME(timestamp, \"America/New_York\")"
```
magnew, line 45: This is fixed in main now with 

'Timestamp, but converted to EST in BigQuery using DATETIME(timestamp, "America/New_York")'

packages/backend/src/controllers/catalogController.ts
```diff
@@ -0,0 +1,52 @@
...
+
+@Route('/api/v1/projects/{projectUuid}/catalog')
+@Response<ApiErrorPayload>('default', 'Error')
+@Tags('Catalog')
+export class CatalogController extends BaseController {
+    /**
+     * Get catalog items
+     * @param projectUuid
+     * @param req contains filters for the catalog items
+     * - search: string
```
magnew, line 29: How do you think these args would change if when we want to paginate tables or filters separately? 

Instead of onlyTable and onlyFields, what do you think of something like `tablePage` and `fieldPage`; when undefined or null, you don't get that object, but if you set it to a number you get that page? There's a few ways we could handle page size with that pattern. 

And for filters, do you think that is separate arg we add later? I imagine the form of that will be different between table/field

packages/frontend/src/features/catalog/hooks/useCatalog.ts
```diff
@@ -0,0 +1,37 @@
...
+}: GetCatalogParams) =>
+    lightdashApi<ApiCatalogResults>({
+        url: `/projects/${projectUuid}/dataCatalog?allTables=${allTables}&allFields=${allFields}`,
+        method: 'GET',
+        body: undefined,
+    });
+
+export const useCatalog = (
+    projectUuid: string,
+    allTables: boolean,
```
rephus, line 25: I think the filters `alltables, allfields` and `search` should be like this 

```

//Initially dataResults only returns pages (allPages=true)
   const { dataResults, mutation } = useCatalog(projectUuid);
   onSearch(search) {
      mutation.mutate({search})
   } 
   onFieldTabSelected() {
       mutation.mutate({allFields=true})
   }
   onTableTabSelected() {
   //This should always return from cache, because it is the same request we load initially.
       mutation.mutate({allTables=true})
   }

```

packages/frontend/src/features/catalog/hooks/useCatalog.ts
```diff
@@ -0,0 +1,37 @@
...
+    allFields: boolean,
+) => {
+    const setErrorResponse = useQueryError();
+
+    return useQuery<ApiCatalogResults, ApiError>({
+        queryKey: ['comments', projectUuid, allTables, allFields],
+        queryFn: () =>
+            fetchCatalog({ projectUuid, allTables: true, allFields: true }),
```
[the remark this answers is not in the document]
magnew: Ok, I didn't realize that. I can make this change now.

magnew (approved):

PR #10118, opened 2024-05-20
feat: add timezone handling to "in the current" filter

owlas:
Closes: https://github.com/lightdash/lightdash-internal-work/issues/1760

test-frontend

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -23,6 +23,7 @@
         "liquidjs": "^10.0.0",
         "lodash": "^4.17.21",
         "moment": "^2.29.4",
+        "moment-timezone": "^0.5.45",
         "pegjs": "^0.10.0",
         "sanitize-html": "^2.12.1",
         "uuid": "^8.3.2",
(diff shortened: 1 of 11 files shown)


packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -376,7 +383,7 @@ export const renderFilterRuleSql = (
     escapeStringQuoteChar: string,
     startOfWeek: WeekDay | null | undefined,
     adapterType: SupportedDbtAdapter,
-    timezone?: string, // TODO replacde with enum
+    timezone: string,
```
owlas, line 386: Not sure why this was here already

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -263,12 +264,16 @@ export const renderDateFilterSql = (
                 filter.settings?.unitOfTime || UnitOfTime.days;
             const fromDate = dateFormatter(
                 getMomentDateWithCustomStartOfWeek(startOfWeek)
```
owlas, line 266: this is the logic change, related only to "in the current" for now

packages/common/package.json
```diff
@@ -23,6 +23,7 @@
         "liquidjs": "^10.0.0",
         "lodash": "^4.17.21",
         "moment": "^2.29.4",
+        "moment-timezone": "^0.5.45",
```
owlas, line 26: since we use moment for calculating filter comparisons

ZeRego (approved):

Issue #3593, opened 2022-10-26
I want to be able to add a description to my group_labels 


TuringLovesDeathMetal:
### Description

In my .yml, I'm able to add `group_labels` to my dimensions and metrics. 

I want to add a description to these group labels and have this description appear on-hover the same way that it does for dimension/metric descriptions. 

I imagine it could either live in the `model` `meta` or the `columns` `meta` tag and would work something like this:

```yaml
version: 2
models:
  - name: users_daily
        columns:
          meta:
            group_labels:
                product_usage:
                  description: "Stats on how our users are using Lightdash"
```

Then this would appear as a description on-hover of the group label (just like a metric or dimension description)
<img width="962" alt="image" src="https://user-images.githubusercontent.com/31848148/198063190-06f6c847-1ce9-40be-8401-91075a3140e5.png">

ZeRego:
I'm inclined to have this information in the model meta. There is more information about the group that we need to store like: label, description, order, sublabels.

note: I also propose a breaking change from `group_label` to `group` since we would now use ID's instead of labels.

Proposed yml
```
version: 2
models:
  - name: users_daily
    meta:
        groups:
            - product_usage:
               label: "Product usage"
               description: "Stats on how our users are using Lightdash"
               order: 1
               sub_groups:
                  - dates:
                     label: "Product usage"
                     description: "Stats on how our users are using Lightdash"
                     order: 1
```

How to use it in dimensions:

```
  columns:
    - name: product_created_date
      meta:
        dimension:
          group_label: product_usage.dates
```

donrudish:
+1. I'd like to see this feature added.

Issue #2686, opened 2022-07-12
Allow for nested grouping in `group_label` meta property


ZeRego:
-- Proposal updated on 2 May --

This approach covers:
- backwards compatible - we still support "group_label" but will mark it as "deprecated" in the docs and ignore it if "group" is available
- reusability - we have a flat list of group definitions which can be used in any order
- future proof: easy to add more group properties ( ordering, color, etc )
- fallback - when group reference is not found we automatically create a group with that label

```
version: 2
models:
  - name: example_model
    meta:
      groups:
        failed:
          label: Failed
          description: Grouping for failed percentiles
        over:
           label: Over
           description: Subgrouping for over percentiles
        under:
           label: Under
           description: Subgrouping for under percentiles
        success:
          label: Success
          description: Grouping for success percentiles
    columns:
      - name: failed_tdigest
        metrics:
          metric_a:
            type: number
            group: ["failed", "over"] # group value can be a string or an array 
          metric_b:
            type: number
            group: ["failed", "over", "third"] # ATM we will only support 2 levels so any extra levels will be ignored
          metric_c:
            type: number
            group: [ "failed", "custom" ] # if custom is not defined in groups, we will automatically show the sub group as "Custom"
          metric_d:
            type: number
            group_label: "My group label" # group label will be ignore in favour of group property
            group: [ "failed", "over" ]
          metric_e:
            type: number
            group_label: "My group label" # we will show the group as "My group label"
          metric_f:
            type: number
            group: [ "success", "over" ] # I can reuse the group "over" in any order I want
```

We are only supporting 1 sub group for now to keep the UI simple. We need to rethink our sidebar UI/UX to support more nesting. Feedback from a user on this:

_Originally posted by @Martin-Carlsson in https://github.com/lightdash/lightdash/issues/2680#issuecomment-1181877041_

Note that the sidebar already has a deep hierarchy with 5 levels, `Table -> dimensions/metrics group -> custom group -> dimension/metric -> date interval dimension` so perhaps we need to rethink this structure if we want to support nested grouping labels.

### Acceptance criteria
- [ ] I'm able to have up to 1 sub-groups in my table group labels (more than 1 level of sub-grouping doesn't get rendered in Lightdash)
- [ ] The sub-groups appear in the sidebar nested in the groups
- [ ] The sub groups are ordered alphabetically within a group

### Documentation
- [ ] update the docs on groups to include information on how to build subgroups 

### Analytics events

[...]

PR #10119, opened 2024-05-20
feat: Allow for nested grouping in group_label meta property

mmcordoba:
Closes: https://github.com/lightdash/lightdash/issues/3593
Closes: https://github.com/lightdash/lightdash/issues/2686

### Description:

https://github.com/lightdash/lightdash/assets/8274051/5c62d51b-d747-49e9-8d1d-95656161c6b6

NOTE: that there are changes to the **field.ts** (removed property group_label) Field Type and to **TreeGroupNode** to deal with the new way to represents groups and subgroups. This change was made so even if we do support still the group_label property on the model yamel definition, i wanted to clean up the UI part of it.

This change requires the DBT models to be refreshed by lightdash?. Not sure if this happens by default when a new version gets deployed.

Acceptance criteria
- [ ] I'm able to have up to 1 sub-groups in my table group labels (more than 1 level of sub-grouping doesn't get rendered in Lightdash)

We have implemented the feature by not bounding the depth of the subgroups, we have done this, because after testing, it feels like it should be up to the user to decide how many subgroups it needs.

- [x] The sub-groups appear in the sidebar nested in the groups
- [x] The sub groups are ordered alphabetically within a group
Documentation
- [x] update the docs on groups to include information on how to build subgroups

### Reviewer actions


packages/common/src/types/field.ts
```diff
@@ -234,7 +239,7 @@ export interface Field {
     compact?: CompactOrAlias;
     round?: number;
     format?: Format;
-    groupLabel?: string;
+    groups?: GroupType[];
```
ZeRego, line 242: To avoid duplicating groups information across dimensions and metrics we can keep this as group references. In the FE we can find the group label and description from the table.

`type Field` -> `groups?: string[];`
`type TableBase`-> `groupDetails?: Record<string, GroupType>`

packages/common/src/types/field.ts
```diff
@@ -292,11 +297,11 @@ export enum DimensionType {
 export interface Dimension extends Field {
     fieldType: FieldType.DIMENSION;
     type: DimensionType;
-    group?: string;
     requiredAttributes?: Record<string, string | string[]>;
     timeInterval?: TimeFrames;
     isAdditionalDimension?: boolean;
     colors?: Record<string, string>;
+    intervalBase?: boolean;
```
ZeRego, line 304:
```suggestion
    isIntervalBase?: boolean;
```

packages/common/src/types/field.ts
```diff
@@ -292,11 +297,11 @@ export enum DimensionType {
 export interface Dimension extends Field {
     fieldType: FieldType.DIMENSION;
     type: DimensionType;
-    group?: string;
```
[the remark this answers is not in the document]
mmcordoba: Done thanks

packages/common/src/types/field.ts
mmcordoba: Done as requested, thanks

packages/common/src/types/field.ts
mmcordoba: Done as requested, thanks

mmcordoba:
> > This change requires the DBT models to be refreshed by lightdash?. Not sure if this happens by default when a new version gets deployed.
> 
> @mmcordoba This doesn't happen by default so we would "break" current grouping. We would like to reduce the problem by following one of 2 options:
> 
> 1. Add backwards compatibility in the UI.
> 
> Adjust the frontend code to fallback to "group_label" and "group" when available and "groups" is empty. Create a ticket to remove tech debt in 1 week once most of the users have re-deployed their project and are stored with the `groups` values.
> 
> 2. Release in 2 phases
> 
> First we change the BE to store the `groups` property alongside the existing properties. After 1 week we remove the legacy properties and change the remaining of the code.
> 
> I'm inclined towards the first approach but you can decide.

Hi i have implemented option 1 as suggested.
Regarding the ticket for the tech debt, i think i can only open an issue , what type of issue you need me to create for tech debt? or could you open a ticket for me for the tech debt clean up?
Thanks

ZeRego:
> Regarding the ticket for the tech debt, i think i can only open an issue , what type of issue you need me to create for tech debt? or could you open a ticket for me for the tech debt clean up?

@mmcordoba you can open a feature request issue and share it with me. I can update the labels and milestone ;) 

I have pushed 2 changes:

- Fix mutation issue that caused date/time dimensions to have duplicated groups. 1 subgroup per time interval

Screenshot 2024-05-29 at 15 58 34

- Implement max group nesting of 2
If you require more nesting you can change `MAX_GROUP_DEPTH` in your fork or create a ticket to add an env var to configure this without a fork. 

Running e2e test before approving 🚀

ZeRego (approved):

Martin-Carlsson:
Wow! This is really cool 🤩

PR #10121, opened 2024-05-20
fix: validate custom dimensions if field is custom dim

joaoviana:
Closes: Incoming...

### Description:

To reproduce: 
- create custom dimension on a chart
- run validator
- see false errors `Dimensions not found...`

Before

After:
image

### Reviewer actions


packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -278,7 +278,9 @@ export class ValidationService extends BaseService {
                 (acc, field) =>
                     containsFieldId({
                         acc,
-                        fieldIds: availableDimensionIds,
+                        fieldIds: availableCustomDimensionIds.includes(field)
+                            ? availableCustomDimensionIds
+                            : availableDimensionIds,
```
ZeRego, line 283: why 
```suggestion
                        fieldIds: [...availableDimensionIds, ...availableCustomDimensionIds],
```

We need the same for `customMetricsErrors` validation as we can create a custom metric from a custom dimension.

joaoviana: And done here with your suggestion! https://github.com/lightdash/lightdash/pull/10121/commits/e39535df835814ec60411196ea3d1bf706ba9db4

ZeRego (approved):

PR #10123, opened 2024-05-20
fix: use http mode for slackbot

owlas:
- Removes socketmode for slack
- init slack app before our own routes
- remove app token (only needed for socket mode)
- never start the slack app manually (it's attached to the existing express instance)

Requires important to change to existing slack configurations:
- Disable socket mode
- Remove `/api/v1` prefix from oauth call back
- Add event subscription call back


rephus:
can we test it on render ?

IrakliJani (approved):

Issue #10107, opened 2024-05-17
Validate custom dimension id against all fields and table calculations


ZeRego:
### Description

Atm you can have a custom dimension and a table calculation with the same id.
This will throw and error when querying and you can't remove the table calculation.

### Steps to Reproduce the Bug or Issue

 add a custom dimension and a table calculation with the same id. eg: "test"

### Cloud or self-hosting

None

PR #10140, opened 2024-05-21
fix: check against tcs when creating custom sql dim

joaoviana:
Closes: #10107

### Description:

This also checks TCs when creating/editing a custom SQL dimension. Followed the issue's steps to replicate it.

Fix:

https://github.com/lightdash/lightdash/assets/7611706/e3fee4fd-7134-4827-bfe5-d35b243a5c98

Side note:

I also looked at `CustomBinDimensionModal` but the code below wouldn't cause an issue unless the TC was named with the same name and prefixed with the base table for the custom Bin dimension. 

Leaving this here if we want to add it in the future. I believe this is unrelated to the issue that the ticket raised though.

```ts

  const customDimensionName = sanitizeId(
                    label,
                    isEditing && isCustomDimension(item)
                        ? item.dimensionId
                        : item.name,
                );

                if (
                    isEditing &&
                    isCustomDimension(item) &&
                    customDimensionName === item.id
                ) {
                    return null;
                }

                const tableCalculationsIds = tableCalculations.map(
                    (tc) => tc.name,
                );

                const customDimensionsIds =
                    customDimensions?.map(
                        (customDimension) => customDimension.id,
                    ) ?? [];

               

                const isInvalid = [
                    ...tableCalculationsIds,
                    ...customDimensionsIds,
                ].some((id) => id === customDimensionName);

                return isInvalid
                    ? 'Dimension/Table calculation with this label already exists'
                    : null;

```

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
+++ b/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
@@ -1,6 +1,7 @@
 import {
     CustomDimensionType,
     DimensionType,
+    getItemId,
     snakeCaseName,
     type CustomSqlDimension,
 } from '@lightdash/common';
@@ -71,19 +72,10 @@ export const CustomSqlDimensionModal: FC<{
                     return null;
                 }
 
-                const tableCalculationsIds = tableCalculations.map(
-                    (tc) => tc.name,
-                );
-
-                const customDimensionsIds =
-                    customDimensions?.map(
-                        (customDimension) => customDimension.id,
-                    ) ?? [];
-
                 const isInvalid = [
-                    ...tableCalculationsIds,
-                    ...customDimensionsIds,
-                ].some((id) => id === customDimensionId);
+                    ...tableCalculations,
+                    ...(customDimensions ?? []),
+                ].some((i) => getItemId(i) === customDimensionId);
 
                 return isInvalid
                     ? 'Dimension/Table calculation with this label already exists'
--- a/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
+++ b/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
@@ -44,6 +44,10 @@ export const CustomSqlDimensionModal: FC<{
         (context) =>
             context.state.unsavedChartVersion.metricQuery.customDimensions,
     );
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
```diff
@@ -67,11 +71,22 @@ export const CustomSqlDimensionModal: FC<{
...
+                        (customDimension) => customDimension.id,
+                    ) ?? [];
+
+                const isInvalid = [
+                    ...tableCalculationsIds,
+                    ...customDimensionsIds,
+                ].some((id) => id === customDimensionId);
+
+                return isInvalid
+                    ? 'Dimension/Table calculation with this label already exists'
```
ZeRego, line 89: The problem also exists when I create a custom dimension and then a table calculation with the same name.

packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
```diff
@@ -67,11 +71,22 @@ export const CustomSqlDimensionModal: FC<{
...
+
+                const customDimensionsIds =
+                    customDimensions?.map(
+                        (customDimension) => customDimension.id,
+                    ) ?? [];
+
+                const isInvalid = [
+                    ...tableCalculationsIds,
+                    ...customDimensionsIds,
+                ].some((id) => id === customDimensionId);
```
ZeRego, line 86: Suggestion
```suggestion
                const isInvalid = [
                    ...tableCalculations,
                    ...(customDimensions ?? []),
                ].some((item) =>getItemId(item) === customDimensionId);
```

joaoviana: Thank you! nice nice

packages/frontend/src/components/Explorer/CustomDimensionModal/CustomSqlDimensionModal.tsx
joaoviana: Done here: https://github.com/lightdash/lightdash/pull/10140/commits/244720998940d9af3fd5d185a1a3b67f583e8810

also trimmed and lowercased on both ends of the comparison: https://github.com/lightdash/lightdash/pull/10140/commits/980677ce58ca092b4ce67e56230b15c341af881c

ZeRego (approved):

Issue #10128, opened 2024-05-20
Time-based fields are labelled incorrectly in chart config panel when not in GMT


magnew:
### Description

When you are in US timezones, time based series are labeled incorrectly in the chart config panel. They show up correctly in the chart and results. 

<img width="1337" alt="Screenshot 2024-05-20 at 12 51 39" src="https://github.com/lightdash/lightdash/assets/1864179/f606afab-664e-4343-ac48-e5f048085a8c">

### Steps to Reproduce the Bug or Issue

1. Set TZ to PST
2. Make a chart with months
3. Group by month
4. The series is labelled incorrectly

### Cloud or self-hosting

None

PR #10142, opened 2024-05-21
fix: use utc time display in series config

magnew:
Closes: #10128

### Description:

Treats time-based field values in the chart series config as UTC. There was an issue where the series config display wasn't matching the chart formatted values. 

The semantics here aren't great--the variable set to 'true' is called convertToUTC, and this it more treating them as utc, but it solves the problem. 

Before
image

After
Screenshot 2024-05-21 at 10 02 31

### Reviewer actions


joaoviana (approved):
Understood the issue and makes sense on the observation you've made on the concept of treating as UTC vs converting to UTC

PR #10143, opened 2024-05-21
fix: invite user without org

ZeRego:
Closes: 

### Description:

How to replicate: 
- set env var `ALLOW_MULTIPLE_ORGS=false` 
- register new user with email X
- see error "Cannot register user in a new organization. ..." and do **NOT** click to cancel registration
- login with demo user
- invite user with email X

Before: 
They would get the error:
Screenshot 2024-05-21 at 19 38 42

After:
user is invited

Added unit tests to cover invite logic:

- should create user and send invite when email is not found
- should send invite when email belongs to user without org
- should send invite when email belongs to inactive user in same org
- should throw error when email belongs to user in different org
- should throw error when email belongs to an active user in same org

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -333,18 +333,16 @@ export class UserService extends BaseService {
         const existingUserWithEmail = await this.userModel.findUserByEmail(
             email,
         );
-        if (
-            existingUserWithEmail &&
-            existingUserWithEmail.organizationUuid !== organizationUuid
-        ) {
-            throw new ParameterError(
-                'Email is already used by a user in another organization',
-            );
-        }
-        if (existingUserWithEmail && existingUserWithEmail.isActive) {
-            throw new ParameterError(
-                'Email is already used by a user in your organization',
-            );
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/UserService.ts
```diff
@@ -333,18 +333,16 @@ export class UserService extends BaseService {
...
-        ) {
-            throw new ParameterError(
-                'Email is already used by a user in another organization',
-            );
-        }
-        if (existingUserWithEmail && existingUserWithEmail.isActive) {
-            throw new ParameterError(
-                'Email is already used by a user in your organization',
-            );
+        if (existingUserWithEmail && existingUserWithEmail.organizationUuid) {
```
owlas, line 336: so this now means that existing users that haven't yet joined an org (i.e. organizationUuid is `undefined`) will fall through here to account setup

ZeRego: correct

rephus (approved):
That makes sense, good unit tests

Issue #10132, opened 2024-05-21
Can't create filter group when parent group is of type "Any"


ZeRego:
### Description

I want to create a nested filters where the main root type is "Any" 

### Steps to Reproduce the Bug or Issue

https://github.com/lightdash/lightdash/assets/9117144/1b6abbde-29c1-4550-8218-1f2191cadcd9

### Cloud or self-hosting

cloud

PR #10144, opened 2024-05-22
fix: filter group when type any

Inder782:
Closes: #10132 

### Description:

Modified the logic on isAndFilterGroup and it is working.

https://github.com/lightdash/lightdash/assets/124162738/61d52f2e-3511-48e1-9006-2af125fc986f

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterGroupForm.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterGroupForm.tsx
@@ -185,8 +185,8 @@ const FilterGroupForm: FC<Props> = ({
                         ]}
                         value={
                             isAndFilterGroup(filterGroup)
-                                ? FilterGroupOperator.and
-                                : FilterGroupOperator.or
+                                ? FilterGroupOperator.or
+                                : FilterGroupOperator.and
                         }
                         onChange={(operator: FilterGroupOperator) =>
                             onChangeOperator(operator)


packages/frontend/src/components/common/Filters/FilterGroupForm.tsx
```diff
@@ -185,8 +185,8 @@ const FilterGroupForm: FC<Props> = ({
                         ]}
                         value={
                             isAndFilterGroup(filterGroup)
-                                ? FilterGroupOperator.and
-                                : FilterGroupOperator.or
+                                ? FilterGroupOperator.or
+                                : FilterGroupOperator.and
```
ZeRego, line 189: Reading the condition, this fix make sense to me 🤔 
@almeidabbm I assume the bug is related to the new functions that "merge" groups. I will leave this PR for you to review

Inder782: can be waiting for him to review so that I can also understand where I went wrong

almeidabbm:
Hello @Inder782, thank you for taking the time to try and fix this. Just checked this PR and what @ZeRego mentioned is correct. By switching that condition, the filters will then be the inverse - as you'll be able to see from the screenshot below - the root filter is `ANY` in the select, but we're sending as if it were `ALL` - `and` keyword in the filter:

image

This bug is most likely coming from the flattening of groups done in the `flattenSameFilterGroupType` function, but haven't had the chance to dig any deeper.

Unfortunately the team is currently at capacity for reviews, therefore I will be closing this PR as it is not top priority for us. However, the branch will remain and we'll hopefully have time to review.

The prioritization is almost entirely determined by user interest, therefore if you know of any other Lightdash users that are in need of this fix, then sharing this information would help us to get to this sooner.

Inder782:
Thanks for pointing that out, @almeidabbm. I should have tried to dig deeper rather than just making it work. Next time, I will look more carefully and ensure the end-to-end bug is solved

Issue #10149, opened 2024-05-22
Catalog: It should be clear when a table is made up of joins


magnew:
### Description

It should be clear when a table is made up of joins (by showing an icon and hover state. use a dummy icon for now)

PR #10160, opened 2024-05-22
chore: show table join indicator in catalog

magnew:
Closes: #10149 

### Description:

Shows a number indicating joined tables in the catalog. Design for this will probably change and theres more info we could show, but this puts the information there.

Screenshot 2024-05-22 at 17 48 08

### Reviewer actions


rephus:
Adding more details to the api breaks the existing api tests, sorry

We can fix it only by checking some properties, not an exact match 

image

magnew:
> Adding more details to the api breaks the existing api tests, sorry
> 
> We can fix it only by checking some properties, not an exact match
> 
> image

All good. I didn't realize. I will fix up the test.

magnew:
@rephus I fixed this the fast way (added the field), but yea, we could relax these deep equal checks next time this is annoying 😄

rephus (approved):

PR #10161, opened 2024-05-22
chore: add partial word and multi-word search to catalog search

magnew:
### Description:

The current catalog search only searches whole words. This adds the ability to search partial words as well as multiple words. Note that without the multi word search, things like `customer table` cause an error. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/CatalogModel/CatalogModel.ts
+++ b/packages/backend/src/models/CatalogModel/CatalogModel.ts
@@ -68,6 +68,12 @@ export class CatalogModel {
         query: string,
         limit: number = 50,
     ): Promise<(CatalogTable | CatalogField)[]> {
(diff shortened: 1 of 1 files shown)


rephus (approved):
we already have some integration api tests on cypress, we should add these words there too. It can be done in a separate PR

rephus:
Tested, it look great.

One small UI detail, can we fix the table columns size when on hover ? 
[Screencast from 23-05-24 09:01:53.webm](https://github.com/lightdash/lightdash/assets/1983672/9b3f5aff-c48c-48e4-b26f-d48578d2c9f1)

magnew:
> we already have some integration api tests on cypress, we should add these words there too. It can be done in a separate PR

I'm going to merge this so it goes in the deploy this morning, but Ill put up a test PR next

Issue #9814, opened 2024-04-18
Bar charts appear outside of the y-axis


rephus:
### Description

![image](https://github.com/lightdash/lightdash/assets/1983672/e0e8abca-48dc-45ae-a56a-0022fe40f4d1)

This was fixed here https://github.com/lightdash/lightdash/issues/2943 https://github.com/lightdash/lightdash/pull/8715. 

But it sounds like there is a regression, or perhaps this use case was never covered. 

Thread: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1713401937122029

### Steps to Reproduce the Bug or Issue

- Create a bar chart
- adjust the min axis range to make a bar appear outside of the chart

### Cloud or self-hosting

None

joaoviana:
How to replicate this locally: 
```
[...]

PR #10162, opened 2024-05-22
fix: allow setting offset to avoid bars overlapping/falling outside of range

joaoviana:
Closes: #9814

### Description:

Screen recording: 

https://github.com/lightdash/lightdash/assets/7611706/49a88ceb-dab1-431f-b259-3076cb56afad

Requirements: 
- Only for x axis
- Only for bar charts, **not** horizontal
- Only for visualizations of type `value` on the x axis (basically, a numeric value like 0, 1, 2)

Adds `minOffset` and `maxOffset` to type `Axis`
User can set this offset when setting x-axis range to manual. 
The value set by default is `0.5` (-0.5 on the left and +0.5 on the right)
When the inputs (for max or min) are set to empty, then we apply the data min and data max from the `rows` data. If the user overrides to their own, the calculations are applied.

To replicate:

Locally:
- Go to `Tracks` table
- Select Timestamp month
- Select Event count
- Create table calculation, with this: ` (${tracks.event_count} - 4028) / 1000`
- X axis should be set to the table calculation
- Add filter for table calculation with: `is greater than` `-1`
- See overlap on Y axis
(now onto the fix)
- Open chart config
- Go to Axis tab
- Select manual x-axis range
- Overlap is no more an issue bc of the offset applied (feel free to play around with the mins and offsets)

On Render: 
- See this chart: https://lightdash-pr-10162.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/saved/7bfc2abb-eb66-45cf-a161-cf89af7ba945/view
- Play around with chart config > axis > x axis > turn off `auto x-axis range`

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
@@ -99,21 +99,19 @@ export const Axes: FC<Props> = ({ itemsMap }) => {
                                 dirtyLayout?.flipAxes ? 'y' : 'x'
                             }-axis range`}
                             min={dirtyEchartsConfig?.xAxis?.[0]?.min}
-                            minOffset={
-                                dirtyEchartsConfig?.xAxis?.[0]?.minOffset
-                            }
                             max={dirtyEchartsConfig?.xAxis?.[0]?.max}
-                            maxOffset={
-                                dirtyEchartsConfig?.xAxis?.[0]?.maxOffset
-                            }
                             setMin={(newValue) => setXMinValue(0, newValue)}
                             setMax={(newValue) => setXMaxValue(0, newValue)}
-                            setMinOffset={(newValue) =>
-                                setXMinOffsetValue(0, newValue)
-                            }
-                            setMaxOffset={(newValue) =>
-                                setXMaxOffsetValue(0, newValue)
-                            }
+                            {...(!dirtyLayout?.flipAxes && {
+                                minOffset:
+                                    dirtyEchartsConfig?.xAxis?.[0]?.minOffset,
+                                maxOffset:
+                                    dirtyEchartsConfig?.xAxis?.[0]?.maxOffset,
+                                setMinOffset: (newValue) =>
+                                    setXMinOffsetValue(0, newValue),
+                                setMaxOffset: (newValue) =>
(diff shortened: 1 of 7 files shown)


joaoviana:
I believe we want to turn off this offset setting for horizontal bar charts:

image

joaoviana:
> I believe we want to turn off this offset setting for horizontal bar charts:
> 
> image

Turned off here: https://github.com/lightdash/lightdash/pull/10162/commits/18ca716a1aeb21bf92238b66fdde4948a5055e25

hamzahc1:
@joaoviana to avoid this being blocked for longer, I'd suggest finding another reviewer for Monday!

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
```diff
@@ -102,6 +107,35 @@ export const Axes: FC<Props> = ({ itemsMap }) => {
...
+                                    dirtyEchartsConfig?.xAxis?.[0]
+                                        ?.minOffset !== undefined ||
+                                    dirtyEchartsConfig?.xAxis?.[0]
+                                        ?.maxOffset !== undefined
+                                }
+                                onChange={(e) => {
+                                    if (e.target.checked) {
+                                        setXMaxOffsetValue(
+                                            0,
+                                            DEFAULT_OFFSET_VALUE_FOR_MANUAL_RANGE_PERCENTAGE,
```
joaoviana, line 125: this is set here, to add it them to the chart config. I'm happy to change it to be set to _true_ or similar and use this constant in `useCartesianChartConfig`

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -902,6 +904,8 @@ const getEchartAxes = ({
         [true, true],
     );
 
+    console.log({ series });
```
ZeRego, line 907: todo: remove

ZeRego (approved):

joaoviana:
Applied logarithmic scaling that accommodates large and small ranges

**small**

before
Screenshot 2024-05-28 at 11 05 07

after
Screenshot 2024-05-28 at 11 05 35

**large**

before:
Screenshot 2024-05-28 at 11 05 31

after
Screenshot 2024-05-28 at 11 05 12

Issue #10148, opened 2024-05-22
investigate: copy link button is not working in Safari


TuringLovesDeathMetal:
### Description

Very, very minor thing I've been seeing for a week or two. When I'm using Lightdash with Safari (Version 17.4.1 (19618.1.15.11.14), Mac OS 14.4.1) and trying to use the "Copy Link" button in explores or elsewhere, it's not actually grabbing the URL to the clipboard. I don't have any clipboard managers installed or anything else that could be hijacking behavior as far as I know.

It does fire the "Link copied to clipboard" green visual, but inevitably, no actual link is copied, and whatever was on my clipboard prior remains.
 
This works great in Chrome/Arc/Firefox, so thinking it's probably something Safari specific. It's easy enough to use other browsers, but wanted to put this on your radar in the hopes that it's a quick fix.

### Steps to Reproduce the Bug or Issue

1. open Lightdash on a safari browser
2. click `copy link` button 
3. try to paste the link somewhere
4. see that the link is not actually being copied

### Cloud or self-hosting

None

joaoviana:
Reference: https://developer.apple.com/forums/thread/691873

PR #10178, opened 2024-05-23
fix: allow safari to copy to clipboard on async values

joaoviana:
Closes: [#10148](https://github.com/lightdash/lightdash/issues/10148)

### Description:

When dealing with async operations to retrieve text to then copy to clipboard, Safari blocks this operation because it doesn't consider it a direct user action: on click -> API call -> copy to clipboard. Since there's an API in the middle of the process, that is then blocked by Safari. 

Found this blog post that explained it nicely: https://wolfgangrittner.dev/how-to-use-clipboard-api-in-firefox/
And also it mentions this safari thread with the solution: https://developer.apple.com/forums/thread/691873

Also added a custom hook `useAsyncClipboard` which covers both major browsers: Chrome Firefox and Safari. 

> [!NOTE]  
> I checked other places in the code where we use `useClipboard` - and these are only used for non-async values. So all good! I also tested this in Safari: clicking `copy value` on a table cell. And it copied correctly!

Demo on Safari (I also tested in Chrome, all good): 

https://github.com/lightdash/lightdash/assets/7611706/e940a9d4-c33f-40c3-a8b6-e65c2bbd81a2

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/ShareShortLinkButton/index.tsx
+++ b/packages/frontend/src/components/common/ShareShortLinkButton/index.tsx
@@ -1,42 +1,36 @@
 import { ActionIcon } from '@mantine/core';
-import { useClipboard } from '@mantine/hooks';
 import { IconCheck, IconLink } from '@tabler/icons-react';
 import { type FC } from 'react';
 import { useLocation } from 'react-router-dom';
-import useToaster from '../../../hooks/toaster/useToaster';
+import { useAsyncClipboard } from '../../../hooks/useAsyncClipboard';
 import { useCreateShareMutation } from '../../../hooks/useShare';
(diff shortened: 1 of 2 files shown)


magnew (approved):
Nice. Because Safari is so...bad 😂 

Tested, works well

Wamy-Dev:
I was searching for a fix for this exact issue for my own code, thanks so much for this hook! Happy to say it solved my exact same issue too!

joaoviana:
> I was searching for a fix for this exact issue for my own code, thanks so much for this hook! Happy to say it solved my exact same issue too!

Oh I'm glad @Wamy-Dev! It was a tricky one, right? ✨

Issue #10034, opened 2024-05-10
test support for dbt 1.8


TuringLovesDeathMetal:
dbt has released v 1.8

Some users have reported that Lightdash does not work with this upgrade (specifically, `lightdash preview`). 

WE need to test this upgrade and see if we need to make any changes

TuringLovesDeathMetal:
Note: the error is that we only 'accept' up to dbt 1.7 - so we throw an error. Can we check to see if it breaks anything if we accept up to 1.8?

PR #10180, opened 2024-05-23
feat: dbt 1.8

ZeRego:
Closes: #10034

**This PR only offers partial support!**

- There is a [regression](https://github.com/dbt-labs/dbt-postgres/issues/96) in the postgres and redshift adaptor.
- Databricks adaptor hasn't released a version compatible with 1.8.

We are going to monitor these adaptors and support them when possible.

### Reviewer actions


packages/common/src/dbt/schemas/lightdashV12.json
```diff
@@ -0,0 +1,64 @@
+{
+    "definitions": {
+        "LightdashNodeConfig": {
+            "allOf": [
+                {
```
ZeRego, line 5: Notice that here we usually extend the dbt definition. 

They have changed the manifest json schema `https://schemas.getdbt.com/dbt/manifest/v12.json` to use draft 2020-12 `https://json-schema.org/draft/2020-12/schema` format.

I couldn't find a way to extend these definitions. 
Let me know if there is a way. 🙏

packages/common/src/dbt/validation.ts
```diff
@@ -11,31 +13,34 @@ import { ParseError } from '../types/errors';
...
-        lightdashSchemaV8,
-        lightdashSchemaV9,
-        lightdashSchemaV10,
-        lightdashSchemaV11,
-    ],
-});
-
+const ajv = new Ajv2020();
+ajv.addMetaSchema(draft7MetaSchema); // add backward compatibility with draft-07
+ajv.addSchema([
```
ZeRego, line 29: **draft 2020** is not compatible with **draft 07**
So we need to initialize AVJ2020 and add draft 07 meta schema.

joaoviana (approved):
🎉

Issue #10151, opened 2024-05-22
Catalog: searching for a group, a table within a group, or a field within a group should show the heirarchy


magnew:
### Description

- [x] ...search for a field that lives in a group, show the group name (like a tree structure. I want to work on this a lot more, so this doesn't have to be so refined)
- [ ] ...search for a group, I should be able to see all the fields within that group too (using the tree structure)
~~- [ ] ...search for field that appears in two tables because of a join, display it twice.--~~

PR #10197, opened 2024-05-26
chore: render catalog as a tree

magnew:
Closes: #10151 

### Description:

Renders the catalog as a tree instead of tables. This is necessary for some of the expansion capabilities and should give us more flexibility. When searching, tree nodes are opened (eg, if a field matches, it's table and group nodes are open). 

This integrates with the selection/metadata display, although the styles need some work. 

Screenshot 2024-05-26 at 15 44 39

### Reviewer actions


rephus:
recreating render

packages/backend/src/generated/routes.ts
```diff
@@ -3115,7 +3117,7 @@ const models: TsoaRoute.Models = {
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
     SupportedDbtVersions: {
         dataType: 'refEnum',
-        enums: ['v1.4', 'v1.5', 'v1.6', 'v1.7'],
+        enums: ['v1.4', 'v1.5', 'v1.6', 'v1.7', 'v1.8'],
```
rephus, line 3120: unrelated change

packages/backend/src/generated/swagger.json
```diff
@@ -91,14 +91,20 @@
...
+                                "type": "string"
+                            },
                             "basicType": {
                                 "type": "string"
                             },
                             "type": {
                                 "$ref": "#/components/schemas/CatalogType.Field"
                             }
                         },
-                        "required": ["basicType", "type"],
```
rephus, line 101: did you remove the basicType ? this was used in the metadata panel

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
```diff
@@ -63,14 +63,14 @@ export const CatalogMetadata: FC<React.PropsWithChildren<Props>> = ({
...
-                    </th>
+                    <tr>
+                        <th>field name</th>
+                        <th>type</th>
+                    </tr>
                 </thead>
                 <tbody>
                     {data.fields?.map((field) => (
-                        <tr key="">
+                        <tr key={field.name}>
```
rephus, line 73: table names are unique, field names are not... but I guess they are unique per table, so this should be ok... Thanks for fixing, I guess it was me who left this empty

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -41,53 +38,71 @@ export const CatalogPanel: FC<React.PropsWithChildren<Props>> = ({
...
-    }, [catalogResults]);
-
     const handleSearchChange = useCallback(
         (searchString: string) => {
             setSearch(searchString);
         },
         [setSearch],
     );
 
+    // TODO: should this transform be in the backend?
```
rephus, line 60: yeah, I'm not sure about this. Is it better to return the data formatted from the backend ? This means this endpoint will be less reusable and I dont' think this will make it more efficient though (we're just moving the complexity to the backend and making search slower...) What do you think ?

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -109,43 +124,68 @@ export const CatalogPanel: FC<React.PropsWithChildren<Props>> = ({
                 'ArrowDown',
                 () => {
                     //FIXME fix bug when multiple "fields" have the same name in search and you can't move
-                    if (selectedTable) {
+                    if (selection) {
                         //TODO move around grouped items
```
rephus, line 128: nice, this is done then, right ?

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -109,43 +124,68 @@ export const CatalogPanel: FC<React.PropsWithChildren<Props>> = ({
                 'ArrowDown',
                 () => {
                     //FIXME fix bug when multiple "fields" have the same name in search and you can't move
```
rephus, line 126: this is fixed too, right ?

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -109,43 +124,68 @@ export const CatalogPanel: FC<React.PropsWithChildren<Props>> = ({
...
             ],
             [
                 'ArrowUp',
                 () => {
-                    if (selectedTable) {
+                    if (selection) {
                         //TODO move around grouped items
-                        const selectedIndex = ungroupedCatalogItems.findIndex(
-                            (item) => item.name === selectedTable,
+                        const sortedGroupTables = Object.keys(
```
rephus, line 159: we could extract some of this logic outside the `arrow` methods and reuse

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <Group
+                noWrap
+                position="apart"
+                py="sm"
+                px="xs"
+                sx={(theme) => ({
+                    minHeight: 40,
+                    borderRadius: theme.radius.sm,
+                    padding: theme.spacing.md,
+                    backgroundColor: hovered
```
rephus, line 50: does the background color not change if selected ?

rephus (commented):
Nice refactoring, code look good

rephus:
Testing feedback:  https://lightdash-pr-10197.onrender.com/projects/891c2261-92ee-4db4-babf-7238a2047856/catalog 

- On pressing a key (up or down) it should select the first item if nothing is selected, but now it does nothing. Regression ? 
- Navigation doesn't work well for `generated_a` / `generated_b` are we selecting the next item in  sorted  list we display ? Regression ? 
[Screencast from 27-05-24 09:34:22.webm](https://github.com/lightdash/lightdash/assets/1983672/2df52dd2-b63e-4a5d-98ec-32099c1ad0cc)
- If you navigate to the last item and press down 1 more, the selected dissapear instead of selecting the first item, probably we should check length -1

rephus:
Why does `cust` search not return any results ? :thinking: but `custom` does.

rephus:
Are we not showing `field` description in the table ? 
This might be useful specially because when I `search` for `custom` I get `first_name` , and I know we are simply not showing all the fields from the table. THis is returned because there is something about `customer` in this field's description

image

rephus (changes requested):
This looks great, nice work with the tree structure. It doesn't seem slow so I'd not worry too much about optimization at this point (like moving the request to the backend) 
There are a few edge cases to handle, specially around keyboard navigation

packages/backend/src/generated/swagger.json
magnew: I didn't intend to. Maybe a bad merge?

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
magnew: Yea, I figured this would be unique enough here. Could always add table name too but 'shrug'

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: Not quite. I changed the selector type to do it, but didn't do work to make sure it was right yet. Im happy to do that next.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: This one should be, yea

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: I think it might depend on the perf we end up seeing with this. This tree function is expensive, but I put some effort into writing it as a single pass instead of multiple, so it might be fine. If it seems to hand the UI thread, we should move it to the BE. Otherwise, I agree that leaving it here makes things more flexible.

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
magnew: It gets a border for now. We could change the style more.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: Yea, probably a good call. Im actually thinking maybe the sort should happen on tree build...

magnew:
> * Navigating up does not loop to the bottom of the list, IMO we should be consistent with this behaviour. (either do both directions or do not loop at all)

I won't respond to these individually, but 🫡, I will fix them up. 

What the hell is going on with generated_a and generated_b? If you just do `['thing_b', 'thing_a'].sort(([a],[b]) => a.localCompare(b))` it returns `['thing_b', 'thing_a']`. That's not what I expect. I'll fix it though. 

I might do the sorting on tree build?

magnew:
> Why does `cust` search not return any results ? 🤔 but `custom` does.

I feel like when the partial word serach got moved, it broke. Need to investigate.

magnew:
> Are we not showing `field` description in the table ? This might be useful specially because when I `search` for `custom` I get `first_name` , and I know we are simply not showing all the fields from the table. THis is returned because there is something about `customer` in this field's description
> 
> image

Ah, cool, yea, lets add it.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: I made this an issue cause it might as well be done along with the fields #10202

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: I made it so the sort happens on the tree build. It adds more computation to building the tree, but we were doing that here and in render anyway.

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
magnew: Priyanka said she was going to work on styles.

magnew:
> Why does `cust` search not return any results ? 🤔 but `custom` does.

We need to look into whats up here. I dont think it has to do with this change.

packages/backend/src/generated/swagger.json
rephus: Irakli is adding it back on his PR , if you didn't change any types in this PR I'd suggest to revert these files

rephus:
I found a bug, I'm not sure if it is related with your changes. But here is how I can replicate: 

- Go into catalog https://lightdash-pr-10197.onrender.com/projects/8b7fec49-2e6f-413a-94a8-f63c7709622b/catalog
- navigate and show metadta for some tables
- Then do a search like `_day`
- Then try navigating with the arrows 
- :fire: metadata will not update and you will see `b[i.group] is undefined`  errors 
- You can't even update metadata by clicking...

[Screencast from 27-05-24 15:09:33.webm](https://github.com/lightdash/lightdash/assets/1983672/7831d198-c624-4c78-8c85-4ccb366fd74d)

rephus (approved):

Issue #10130, opened 2024-05-21
POC - Sidebar behaviour


PriPatel:
https://www.figma.com/proto/rlqMIJ4pm3t8aAsiMd3CZJ/Lightdash-Product-New?node-id=2646-46180&t=cd5OxnC0OuocwANS-0&scaling=min-zoom&page-id=2646%3A20890&starting-point-node-id=2646%3A46180

@joaoviana Is it possible to get a quick render environment for me to play with following:

- dummy block in the middle of the page.
- click on a button and a sidebar opens, resizing the dummy block. 
- I would love to see the 25% width variant of the drawer with a max width of 800px and a min of 600px. (I cant play with this in figma and its quite important as I need to resize browsers to get this right! ) 
- Smooth transition exploration. It would be nice if this resized nicely, rather than jumped.

PriPatel:
If you think this isnt going to work @joaoviana , happy to close it

joaoviana:
Only concern is that `Drawer` from mantine doesn't have an option to resize the rest of the layout. This is still no issue tho. Just have to create my own with some good ol CSS

PR #10212, opened 2024-05-28
feat: add right sidebar to page component

joaoviana:
Relates to: #10130 

### Description:

Adds `rightSidebar` to `Page`
Allows `Sidebar` to take a position - can be `right` or `left`
Ensures that the `useSidebarResize` hook can take a `position` arg so resizing works well. 
Adds a temporary example in `Catalog` to showcase how this works. 

Screen recording:

https://github.com/lightdash/lightdash/assets/7611706/29ec9957-e9f5-4361-9440-edc1c544c869

### Reviewer actions


joaoviana:
> Nice, the code looks good, but it is a lot of boiler plate just to make the panel be on the left or the right. do you think is worth to have this option ? I didn't see it on the requirements.

I see what you mean! Reason for adding this was so:
- we reuse the `Page` component and take advantage of the styling already applied when there's a sidebar - this helps with the height calculation of the page, its width, the animation that comes with it, and scrolling behaviours

Since the requirement included adding `animation`, I figured we could reuse the current `Sidebar` and `Page` and keep it consistent across the app.

Issue #10200, opened 2024-05-27
Tables in the catalog page should reflect the tables visible in the tables sidebar


PriPatel:
Currently, there are hidden tables that are showing.

PR #10213, opened 2024-05-28
fix: filter table selection

rephus:
Closes: #10200

### Description:

We apply the same filtering logic from explores?filtered=true

I filtered on table selection `generated_b` but not `generated_a`

Table
Screenshot from 2024-05-28 10-55-33

Search table 
Screenshot from 2024-05-28 11-17-40

Filter fields from generated_b on search

Screenshot from 2024-05-28 11-27-55

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CatalogService/CatalogService.ts
+++ b/packages/backend/src/services/CatalogService/CatalogService.ts
@@ -10,9 +10,13 @@ import {
     ExploreError,
     ForbiddenError,
     getBasicType,
+    hasIntersection,
     isDimension,
     isExploreError,
     SessionUser,
+    SummaryExplore,
+    TablesConfiguration,
+    TableSelectionType,
     UserAttributeValueMap,
 } from '@lightdash/common';
 import { LightdashAnalytics } from '../../analytics/LightdashAnalytics';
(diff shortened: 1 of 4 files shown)


packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -98,10 +102,32 @@ export class CatalogService<
         }, []);
     }
 
-    private static async getCatalogTables(
+    private static isExploreFiltered(
+        tablesConfiguration: TablesConfiguration,
+        explore: Pick<SummaryExplore, 'tags' | 'name'>,
+    ) {
+        // Optimization tip:
+        // Perhaps we should filter these in catalong on `index` rather than in real time (requires dbt refresh on table config changes)
```
magnew, line 110: You mean doing it that way **will** require dbt refresh on table config changes? Is that not true anyway? Or I guess tag changes don't require dbt refresh?

magnew (commented):
@rephus the code here makes sense, but I get an empty catalog when testing. Locally and on render. Are we filtering out too many tables 😄? Or do I need to set something up? I didn't get into debugging it but I can if you need.

packages/backend/src/services/CatalogService/CatalogService.ts
rephus: > tag changes don't require dbt refresh
This, tag changes don't require dbt refresh, but if we filter them out on index, it will be required. 

We can't do this anyway if we are going to `include` them on an advance filter.

rephus:
> the code here makes sense, but I get an empty catalog when testing. Locally and on render

Thanks, there was a bug when all tables were selected (default config) . Fixed.

magnew (approved):

Issue #10194, opened 2024-05-24
Frontend - Sidebar polish


PriPatel:
https://www.figma.com/design/S1dyYIYv7iLGgXOm7w6FVY/Lightdash-Data-Dictionary?node-id=98-3925&t=UKNfTVCyTHKK7gT7-0
**Kick off required to determine how much we can do in a day and a half!**

- [ ] 600px minimum width
- [ ] 800px maximum width
- [ ] Sidebar with markdown
- [ ] Sidebar with no markdown
- [ ] Sidebar with usage analytics
- [ ] Sidebar with 0 usage analytics
- [ ] Scalable metadata section
- [ ] Metrics and Dimensions list and behaviours
---
- [ ] Layout and responsiveness
- [ ] Dropshadows and strokes

PR #10222, opened 2024-05-28
feat: polish metadata sidebar 

joaoviana:
Closes: #10194 

### Description:

This PR covers the following: 
- Adds `SidebarWidthProps` to `Sidebar` to set the following styles: 
```ts
rightSidebarWidthProps={{
    defaultWidth: 600,
    minWidth: 600,
    maxWidth: 800,
 }}
```
- sets styles to the sidebar, according to the ticket mentioned above ^

Demo: 

https://github.com/lightdash/lightdash/assets/7611706/91d70f1f-ea9b-4c76-b27e-f5eec6783095

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/catalog/components/CatalogAnalyticCharts.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogAnalyticCharts.tsx
@@ -38,25 +38,9 @@ export const CatalogAnalyticCharts: FC<React.PropsWithChildren<Props>> = ({
     }
 
     return (
-        <Stack w="100%">
-            <Paper
-                maw="50%"
-                withBorder
-                p="xs"
-                px="lg"
-                sx={(theme) => ({
-                    backgroundColor: theme.colors.gray[0],
-                })}
-            >
-                <Text c="gray.6" fw={500} fz="xs">
-                    Charts
-                </Text>
-                <Text fw={700} fz="xl">
-                    {charts.length}
-                </Text>
-            </Paper>
+        <Stack>
             {chartsInSpace.length > 0 && (
-                <Stack w="100%">
+                <Stack>
                     {chartsInSpace.map((chart) => {
                         return (
                             <Paper key={chart.uuid} withBorder w="100%" p="xs">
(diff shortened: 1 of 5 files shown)


packages/frontend/src/features/catalog/components/CatalogAnalyticCharts.tsx
```diff
@@ -32,28 +38,76 @@ export const CatalogAnalyticCharts: FC<React.PropsWithChildren<Props>> = ({
     }
 
     return (
-        <Stack>
+        <Stack w="100%">
+            <Paper
+                maw="50%"
```
magnew, line 43: Is something else going in the other 50%? It looks a little funny at the moment

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
```diff
@@ -53,111 +59,302 @@ export const CatalogMetadata: FC = () => {
...
+                        paddingRight: theme.spacing.xs,
+                        paddingLeft: 0,
+                        fontSize: theme.fontSizes.sm,
+                        fontWeight: 500,
+                        '&[aria-selected="true"]': {
+                            color: theme.colors.gray[9],
+                        },
+                        '&[aria-selected="false"]': {
+                            color: theme.colors.gray[6],
+                        },
```
magnew, line 122: I think it's typically kind of frowned upon to style things based on aria-props. They have some of the most variable browser behavior.

This is probably fine for something not very critical, so leave it if you want. But could we use a class or data-prop instead? I know we arent controlling all this so if mantine wants it this way, so be it.

magnew (commented):
Beautiful! Much nicer. 

I had a couple minor suggestions but nothing blocking. One other thing that doesnt have to be part of this review is that the side bar opening for the first time has pretty noticeable lag. Like a couple seconds. Maybe it's fine, but it's something to keep an eye on. I know we have this concern with some of the other animating elements as well.

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
joaoviana: oops! 
used the `data-active` data attribute. Yea, thank you for checking this!! better now that we use a `data-prop` :+1:

packages/frontend/src/features/catalog/components/CatalogAnalyticCharts.tsx
joaoviana: Great point @magnew, I've actually removed this section as in the intended _final_ design we want 3 cards here: 
Views + Charts + (something else I cant remember)

Since we don't have all 3, then we remove this one.

Issue #10195, opened 2024-05-24
When browsing through 'fields', I want the option to see more 'metadata' for each one.


PriPatel:
Similar to the ticket below but this time, for fields.
https://github.com/lightdash/lightdash/issues/10073

- [ ] Copy over all interactions and tabs
- [ ] Usage analytics should cover the charts that the field is used in
- [ ] Show markdown/descriptions of the field

There will be more features to add here soon, but this provides the base!

PR #10223, opened 2024-05-28
feat: catalog field analytics

rephus:
Closes: #10195

### Description:

[Screencast from 29-05-24 09:28:17.webm](https://github.com/lightdash/lightdash/assets/1983672/c3d6442d-8cf5-43b4-8d43-c11bf5dd3bd6)

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CatalogService/CatalogService.ts
+++ b/packages/backend/src/services/CatalogService/CatalogService.ts
@@ -366,9 +366,7 @@ export class CatalogService<
         table: string,
         field: string,
     ): Promise<CatalogAnalytics> {
-        const { organizationUuid } = await this.projectModel.getSummary(
-            projectUuid,
-        );
+        const { organizationUuid } = user;
         if (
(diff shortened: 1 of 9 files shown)


packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -359,4 +359,58 @@ export class CatalogService<
         );
         return { charts: chartAnalytics };
     }
+
+    async getFieldAnalytics(
+        user: SessionUser,
+        projectUuid: string,
```
joaoviana, line 365: can we also get the `organizationUuid` from user? would remove one `getSummary` query - but all good if not

joaoviana (approved):
LGTM

I've tested: 
- viewing fields that are used in different charts and in different spaces
- viewing fields that aren't used anyway
- overview/description of field is working ok
- switching from field-> table and table-> field interaction is OK

Issue #10210, opened 2024-05-27
Fields should be selectable to show metadata


magnew:
### Description

Fields should be selectable to show metadata

PR #10224, opened 2024-05-28
chore: make fields selectable

magnew:
Closes: #10210

### Description:

Makes fields selectable and fixed up keyboard navigation to move across groups and tables. @rephus suggested this should make the same metadata requests as the containing table, so that's what it does. For now it doesn't display any field metadata specifically, just highlights the field in the table metadata. We can update the metadata view to show field data. 

**Note:** one main limitation here is that with keyboard navigation, selections do not scroll into view. I played with it some and I the scroll setup seems like it will sufficiently complex that we should add this once we have that finalized more. 

Screenshot 2024-05-28 at 18 46 43

### Reviewer actions

### DIFF
--- a/packages/common/src/types/catalog.ts
+++ b/packages/common/src/types/catalog.ts
@@ -15,6 +15,12 @@ export enum CatalogType {
     Field = 'field',
 }
 
+export type CatalogSelection = {
+    group: string;
+    table?: string;
+    field?: string;
+};
+
 export type ApiCatalogSearch = {
     search?: string;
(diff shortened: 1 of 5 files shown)


joaoviana (approved):
Tested the following scenarios:
- Search, go up and down with keyboard
- Use mouse to click - tried to click on the same button, all good. Perhaps we can skip the action if the field that is already selected is the same one as I'm clicking - a no-op?
- Search, removed search, searched again and went up and down with the keyboard + mouse

PR #10225, opened 2024-05-28
feat: add joined tables to metadata sidebar

joaoviana:
Closes: N/A

### Description:

Returns `joinedTables` from metadata endpoint;

Displays them: 

Screenshot 2024-05-28 at 17 55 45

### Reviewer actions

### DIFF
--- a/packages/common/src/types/catalog.ts
+++ b/packages/common/src/types/catalog.ts
@@ -57,6 +57,7 @@ export type CatalogMetadata = {
     modelName: string;
     source: string | undefined;
     fields: CatalogField[];
+    joinedTables: string[];
 };
 export type ApiCatalogMetadataResults = CatalogMetadata;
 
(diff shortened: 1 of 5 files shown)


packages/common/src/types/catalog.ts
```diff
@@ -57,6 +57,7 @@ export type CatalogMetadata = {
     modelName: string;
     source: string | undefined;
     fields: CatalogField[];
+    joinedTables: string[];
```
rephus, line 60: I thought we already had a property like this, to display joins in the list of tables. Right @magnew ?

rephus (approved):
Code looks good, just confirming we are not duplicating code

packages/common/src/types/catalog.ts
joaoviana: That's a great point @rephus, I checked that code but made more sense (to me) to add this to the response of the `metadata` - since it's a _simple_ map over the explore. 

I'm going ahead and merge this. I feel like this case is a great opportunity to make more refactors/improvements later 😄

PR #10230, opened 2024-05-29
chore: fix selection of tables not in results

magnew:
Closes: 

### Description:

This fixes a bug in catalog table selection where if a table wasn't returned in the results, we wouldn't fetch metadata for it. An example of this issue looks like: search 'customer', and a fields within tables get returned that match customer, but the table itself doesn't. We still build the tree showing the parent table, but if it wasn't in the results we weren't fetching it. 

This changes things to make the assumption that if the user was able to select a table, it's possible to fetch it's metadata.  

### Reviewer actions


magnew:
This got fixed elsewhere

Issue #10193, opened 2024-05-24
Frontend - Table and Page layout


PriPatel:
- [x] Empty state when no tables
- [x] Empty state when no search results
- [x] Table results with groups
- [x] Table results without groups
- [x] Hover states

- [x] Card design
- [x] Stroke and dropshadows
- [ ] Iconography

joaoviana:
Closing as `iconography` will be dealt with on #10260

PR #10247, opened 2024-05-30
style: update catalog tree list and group styles

joaoviana:
Relates to: https://github.com/lightdash/lightdash/issues/10193

### Description:

Updates List and Group components' styles

Since we don't use CSS modules, we need to track if the a component is first or last of the list to render styles conditionally like: 
- border radius
- border bottom

Before

Screenshot 2024-05-30 at 09 58 36

After

https://github.com/lightdash/lightdash/assets/7611706/8b5b45cf-3590-4f5c-93e1-ec6c7bf85b31

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
@@ -1,17 +1,15 @@
 import { type CatalogField, type CatalogTable } from '@lightdash/common';
 import {
-    ActionIcon,
     Box,
     Collapse,
     Group,
     Highlight,
-    Text,
     Tooltip,
+    UnstyledButton,
 } from '@mantine/core';
 import {
-    IconChevronRight,
     IconExternalLink,
-    IconLayersLinked,
+    IconLayersIntersect,
     IconTable,
 } from '@tabler/icons-react';
 import React, { useState, type FC } from 'react';
@@ -26,6 +24,8 @@ type Props = {
     isSelected?: boolean;
     url: string;
     onClick?: () => void;
+    isFirst: boolean;
+    isLast: boolean;
 };
 
 export const CatalogTableListItem: FC<React.PropsWithChildren<Props>> = ({
@@ -34,6 +34,8 @@ export const CatalogTableListItem: FC<React.PropsWithChildren<Props>> = ({
     searchString = '',
     isSelected = false,
     url,
+    isFirst,
+    isLast,
     onClick,
     children,
 }) => {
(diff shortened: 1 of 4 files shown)


rephus:
Keyboard navigation seems fine, but clicking on tables feels weird, kind of inconsistent. 

plus I can get into some invalid states when combining keyboard and click 

image

joaoviana:
> Keyboard navigation seems fine, but clicking on tables feels weird, kind of inconsistent.
> 
> plus I can get into some invalid states when combining keyboard and click
> 
> image

odd. I can't replicate it @rephus , could you share some steps to reach to this behaviour pls?

magnew:
> > Keyboard navigation seems fine, but clicking on tables feels weird, kind of inconsistent.
> > plus I can get into some invalid states when combining keyboard and click
> > image
> 
> odd. I can't replicate it @rephus , could you share some steps to reach to this behaviour pls?

I see it too. I think if you click on the table name itself, instead of anywhere else on the list item, it doesnt select.

rephus:
For example:  If I click on the table name, nothing happens, I need to click on the row, itself to select a table.

We do we have stopPropagation in place ?

joaoviana:
I see @rephus & @magnew, a click on the table text content doesn't open the `metadata`. Got it!

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
```diff
@@ -53,68 +55,68 @@ export const CatalogTableListItem: FC<React.PropsWithChildren<Props>> = ({
...
                     backgroundColor: hovered
-                        ? theme.colors.gray[2]
-                        : theme.colors.gray[1],
+                        ? theme.colors.gray[1]
+                        : 'transparent',
                     border: isSelected
                         ? `2px solid ${theme.colors.blue[6]}`
                         : undefined,
                     cursor: 'pointer',
+                    borderTopLeftRadius: isFirst ? theme.radius.lg : 0,
```
magnew, line 67: Could you do this with css or do you need the control these props get you?

Like:
```
'&:first-of-type': {
    borderTopLeftRadius: theme.radius.xl,
},
```

If you jsut like props better, that's fine too :D

rephus:
Nice, it works now, there is only one small glitch, 
If I click on a name, and then I use the keyboard, the old selected name keeps being selected
[Screencast from 30-05-24 13:16:06.webm](https://github.com/lightdash/lightdash/assets/1983672/46615b18-2d1f-490e-80bf-1c1d4f5e1141)

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
```diff
@@ -53,68 +55,68 @@ export const CatalogTableListItem: FC<React.PropsWithChildren<Props>> = ({
...
+                    borderTopRightRadius: isFirst ? theme.radius.lg : 0,
+                    borderBottomLeftRadius: isLast ? theme.radius.lg : 0,
+                    borderBottomRightRadius: isLast ? theme.radius.lg : 0,
```
magnew, line 70: There's a minor issue here with the last table radius when there is a child. 

Screenshot 2024-05-30 at 13 16 56

rephus:
On search, if I click on the tabl ename, the panel is collapsed, if I click outside, the table is selected (metadata) 

[Screencast from 30-05-24 13:17:55.webm](https://github.com/lightdash/lightdash/assets/1983672/a57bb439-025f-4774-a9ec-c82e2ddf52f3)

rephus:
If a table is collapsed, I can still navigate throught the files using the keyboard . (I think we should probably skip them) 

[Screencast from 30-05-24 13:19:27.webm](https://github.com/lightdash/lightdash/assets/1983672/86720cd6-504a-4bea-a916-d32d5b8feecf)

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
joaoviana: It works well for the `first-of-type` but doesn't work for the `last-of-type` because some siblings are also `div`s that aren't a `CatalogTableListItem`.Group. I would have to override the type of this element to achieve the same with this selector... I could also apply the styles on the parent, but I think that would make it less easier to read/track. Hope it's ok 👍

packages/frontend/src/features/catalog/components/CatalogTableListItem.tsx
joaoviana: Nice! Added a conditional check
image
https://github.com/lightdash/lightdash/pull/10247/commits/c288de2aaa170f87316508384f88b28e6299db2d

rephus (approved):
we can fix some of these navigation issues in a separate PR.Everything else look good

joaoviana:
> we can fix some of these navigation issues in a separate PR.Everything else look good

thanks @rephus ! I'll look into this later then. I was also able to replicate what you shared 👍

Issue #10215, opened 2024-05-28
I should be able to apply filters to this page to get more relevant results


PriPatel:
- [ ]  I should be able to limit by search by result type (metric / dimension / description). ie if I select only descriptions, my search results should only show fields where the description has matched my text search. 
- [ ] I should be able to hide base tables if they also appear as joined ones 
- [ ] I should be able to hide results from table groups (because sometimes these are set up to hide unwanted tables) 
---
- [ ] If filters have been applied, it should be obvious in the form of a filter pill
- [ ] I should be able to easily clear the filters. 

[Sketch](https://www.figma.com/design/S1dyYIYv7iLGgXOm7w6FVY/Lightdash-Data-Dictionary?node-id=163-25572&t=GI2WpnO078wuyEjl-0)

PR #10250, opened 2024-05-30
feat: catalog filters

magnew:
Closes: #10215 

### Description:

Add filters for search to the data catalog. A couple notes:
- This is done in this review by filtering the post-database-search results. Adding these filters to the database search can probably be done, but requires more
- Filtering is done on the FE, but it could be moved to the BE
- I mixed filtering into the tree build, which makes it harder to separate, but faster. We could also do it as a filter pass on the sorted tree. 
- There are some UX questions. Feedback welcome :D 

### Reviewer actions


joaoviana:
To address on this PR or later (up to you!)
1. I would like to be able to revert the filters back to default
2. I would like to know that there are filters applied when the popover is closed. (different color?)

rephus:
Should we filter out `staging` grouped table here ? 

image

magnew:
> To address on this PR or later (up to you!)
> 
> 1. I would like to be able to revert the filters back to default
> 2. I would like to know that there are filters applied when the popover is closed. (different color?)

Cool, Priyanka gave some ideas here. Implementing.

magnew:
> Should we filter out `staging` grouped table here ?
> 
> image

Hmm. Clearly I missed something. Looking.

magnew:
> Should we filter out `staging` grouped table here ?
> 
> image

Could this have been not clicking the apply button? We have now removed it

rephus:
> > Should we filter out `staging` grouped table here ?
> > image
> 
> Could this have been not clicking the apply button? We have now removed it

It's not the case. I can only see a description change. 

[Screencast from 30-05-24 17:00:10.webm](https://github.com/lightdash/lightdash/assets/1983672/f48eeb85-44e1-4c93-a75b-104764bc9dfd)

magnew:
> > Could this have been not clicking the apply button? We have now removed it
> 
> It's not the case. I can only see a description change.
> 
>  Screencast.from.30-05-24.17.00.10.webm

Cool, thanks. I got it. Wasnt hiding tables with only fields and no table returned

magnew:
> I'm not sure why these fields are still appearing here, if the description filter is disabled.
> 
> image

There was confusion about how descriptions should work and Priyanka and I decided to remove it for now.

packages/frontend/src/features/catalog/components/CatalogTree.tsx
```diff
@@ -127,6 +134,9 @@ export const CatalogTree: FC<React.PropsWithChildren<Props>> = ({
         return null;
     }
 
+    const hideGroupedTables =
+        Object.keys(tree).length === 1 && tree['Ungrouped tables'];
```
joaoviana, line 138: sorry I don't get this. Can you explain?

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -96,6 +100,20 @@ function getSelectionList(tree: CatalogTreeType): CatalogSelection[] {
     return selectionList;
 }
 
+enum FilterType {
+    Dimensions = 'dimensions',
+    Metrics = 'metrics',
+    HideBaseTables = 'hideBaseTables',
```
joaoviana, line 106: maybe we can remove this until implemented?

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -385,33 +472,176 @@ export const CatalogPanel: FC = () => {
...
+                                        );
+                                    }}
+                                    label={
+                                        <Text fz="xs" fw={500} c="gray.7">
+                                            Hide grouped tables
+                                        </Text>
+                                    }
+                                />
+                            </Stack>
+                            <Divider c="gray.1" />
```
joaoviana, line 627: I know it's in the designs but I would remove this - feel free to ignore - no change required tbh!

joaoviana (approved):
Working great for me! Some considerations but non-blocking

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: Sure!

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: Good catch

packages/frontend/src/features/catalog/components/CatalogTree.tsx
magnew: If there is only one group and it is the ungrouped tables group, we don't show the group name (which would be 'Ungrouped tables'). Hiding this wasn't in the designs but it felt weird to have one unnamed node there.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -385,33 +472,176 @@ export const CatalogPanel: FC = () => {
...
+                                                },
+                                            }}
+                                        >
+                                            Hide grouped tables
+                                        </Badge>
+                                    )}
+                                </Group>
+                            </Button>
+                        </Popover.Target>
+                        <Button
```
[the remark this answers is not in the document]
magnew: Yea, good call

PR #10255, opened 2024-05-30
style: initial catalog field styling

joaoviana:
Relates to: https://github.com/lightdash/lightdash/issues/10193

### Description:

Made the `CatalogFieldListItem` an  `unstyled` button
Added a transparent border to it regardless of it being selected or not - helps with content jumping when selecting/de-selecting
Added some spacings necessary for making it look nicer
Added temporary styles/icons to fields that are dimensions or metrics. This will change soon once I get @PriPatel 's new assets. 

Before

Screenshot 2024-05-30 at 14 56 58

After

Screenshot 2024-05-30 at 14 55 20

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
@@ -35,7 +35,7 @@ export const CatalogFieldListItem: FC<React.PropsWithChildren<Props>> = ({
                 onMouseEnter={() => setHovered(true)}
                 onMouseLeave={() => setHovered(false)}
                 onClick={onClick}
-                py="xs"
+                py="two"
             >
                 <Box miw={150}>
                     <Group
--- a/packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
@@ -1,6 +1,6 @@
-import { type CatalogField } from '@lightdash/common';
-import { Box, Group, Highlight } from '@mantine/core';
-import { IconAbc } from '@tabler/icons-react';
+import { FieldType, type CatalogField } from '@lightdash/common';
+import { Box, Group, Highlight, UnstyledButton } from '@mantine/core';
+import { IconAbc, IconNumber } from '@tabler/icons-react';
 import React, { useState, type FC } from 'react';
 import MantineIcon from '../../../components/common/MantineIcon';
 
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
```diff
@@ -22,40 +22,58 @@ export const CatalogFieldListItem: FC<React.PropsWithChildren<Props>> = ({
 
     return (
         <>
-            <Group
-                noWrap
-                py="xs"
+            <UnstyledButton
```
magnew, line 25: Why button? So you get focus and tabbing? The focus conflicts with our custom keyboard nav a bit, but maybe it's fine. If we want to put anything else interactive in these, it's going to need to become something else, but so far we dont.

packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
```diff
@@ -22,40 +22,58 @@ export const CatalogFieldListItem: FC<React.PropsWithChildren<Props>> = ({
...
+                        })}
                     >
-                        {field.name || ''}
-                    </Highlight>
+                        <MantineIcon
+                            icon={
+                                // TODO: Add icon for field type and for subtype
+                                field.fieldType === FieldType.DIMENSION
+                                    ? IconAbc
+                                    : IconNumber
```
magnew, line 56: Instead of number, should we use the same one we use in explores, filters, etc? `Icon123`

magnew (approved):
Beautiful! One design feedback I have is that the vertical margins between them feel a bit much to me. 
Screenshot 2024-05-30 at 17 54 15

Although I know you are working with @PriPatel more closely, so if you've talked about any of my suggestions already, feel free to ignore :D 

My question about buttons is not a blocker. Just a consideration.

packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
joaoviana: Absolutely, but these icons will be replaced tomorrow, so it's very very temporary 👍

packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
joaoviana: Good point! Let me change it back and see if things remain the same 👍 
My only reason for this was to make it an interactive element, but I understand the concern 💪

joaoviana:
> Beautiful! One design feedback I have is that the vertical margins between them feel a bit much to me. Screenshot 2024-05-30 at 17 54 15
> 
> Although I know you are working with @PriPatel more closely, so if you've talked about any of my suggestions already, feel free to ignore :D
> 
> My question about buttons is not a blocker. Just a consideration.

I was literally thinking about this and I'm glad you noticed it 😅  I've reduced the vertical spacing!

packages/frontend/src/features/catalog/components/CatalogFieldListItem.tsx
joaoviana: here! https://github.com/lightdash/lightdash/pull/10255/commits/24dbc17fa7055306df9029fa16f2bfd62bd80a5f

PriPatel:
> > Beautiful! One design feedback I have is that the vertical margins between them feel a bit much to me. Screenshot 2024-05-30 at 17 54 15
> > Although I know you are working with @PriPatel more closely, so if you've talked about any of my suggestions already, feel free to ignore :D
> > My question about buttons is not a blocker. Just a consideration.
> 
> I was literally thinking about this and I'm glad you noticed it 😅 I've reduced the vertical spacing!
> 
> here: [8ef1f9f](https://github.com/lightdash/lightdash/commit/8ef1f9fd5f9d7bebe61eb8b74ceca55aab77a828)

ohh actually the padding in the fields bits were actually larger! 
Screenshot 2024-05-31 at 11 52 32

PR #10256, opened 2024-05-30
fix: refactor validation error insert

ZeRego:
Closes: 

### Description:

Refactor code to do a bulk insert rather than one insert per row.

### Reviewer actions


rephus (approved):
Cool, good `(validations.length > 0` edge case handling, I normally forget about that. 

IIRC in postgres there is a soft limit of 1000 bulk insert rows. I recently had to do something similar where we insert in chunks to avoid this limit. I don't think we have 1000+ validation errors, but something to keep in mind.

Issue #10221, opened 2024-05-28
Using `explore from here` on a string with a NULL value applies an incorrect filter


TuringLovesDeathMetal:
### Description

When you try to `view underlying data` on a value that is a `ø` string, then click `explore from here`, we filter to `string = ø` when we should actually be filtering to `string IS NULL`. 

### Steps to Reproduce the Bug or Issue

1. Create a chart with a `null` dimension string (i.e. `ø`)
2. Click `view underlying data`
3. Click `explore from here`
4. see that there is no data shown because the filter is incorrect (it should be `string IS NULL`)

Here's an example:

https://www.loom.com/share/06b38a04ac094ea4a557e899334bbd5e

### Cloud or self-hosting

None

PR #10263, opened 2024-05-30
fix: underlying data with null filter

ZeRego:
Closes: #10221 

### Description:

For `null` values we pass the formatted value since echarts won't render them correctly otherwise. See more about bug here: https://github.com/apache/echarts/issues/15821

So when handling the echarts event we change it back to `null`.

https://github.com/lightdash/lightdash/assets/9117144/df263424-324e-4bfd-8d68-4b93b43ef176

### Reviewer actions


magnew (approved):
Nice

PR #10264, opened 2024-05-30
chore: add empty states for the catalog tree

magnew:
### Description:

Add empty states for the catalog tree. I used the existing sub optimal state compoenet cause it had pretty much everything. The style isn't exactly like the mockup but it's close. 

Note that this doesn't include the metadata empty state. 

No search results:
Screenshot 2024-05-30 at 19 32 25

No table in project:
Screenshot 2024-05-30 at 19 35 59

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/catalog/components/CatalogPanel.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogPanel.tsx
@@ -21,11 +21,14 @@ import {
     IconFilter,
     IconReportSearch,
     IconSearch,
+    IconTable,
     IconX,
 } from '@tabler/icons-react';
 import { useCallback, useMemo, useState, type FC } from 'react';
 import { useHistory } from 'react-router-dom';
+import LinkButton from '../../../components/common/LinkButton';
 import MantineIcon from '../../../components/common/MantineIcon';
+import SuboptimalState from '../../../components/common/SuboptimalState/SuboptimalState';
 import { useCatalogContext } from '../context/CatalogProvider';
 import { useCatalog } from '../hooks/useCatalog';
 import { useCatalogAnalytics } from '../hooks/useCatalogAnalytics';
@@ -107,6 +110,7 @@ export const CatalogPanel: FC = () => {
         selection,
         setSelection,
     } = useCatalogContext();
+
     // There are 3 search variables:
     // - search: the current search string
     // - completeSearch: the 3+ char search string that gets sent to the backend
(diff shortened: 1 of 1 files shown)


packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -414,13 +433,61 @@ export const CatalogPanel: FC = () => {
...
             </Group>
 
-            <CatalogTree
-                tree={catalogTree}
-                projectUuid={projectUuid}
-                searchString={debouncedSearch}
-                selection={selection}
-                onItemClick={selectAndGetMetadata}
-            />
+            {noResults ? (
```
rephus, line 436: In other components, we normally return `invalid states` in a different return , instead of adding this conditional state in the main return. What's the best react way to do this ?

rephus:
The link to `learn more` doesn't work :eyes: https://lightdash-pr-10264.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/catalog

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -414,13 +433,61 @@ export const CatalogPanel: FC = () => {
...
+                        <SuboptimalState
+                            icon={IconTable}
+                            title="No tables found in this project"
+                            description={
+                                "Tables are the starting point to any data exploration in Lightdash. They come from dbt models that have been defined in your dbt project's .yml files."
+                            }
+                            action={
+                                <LinkButton
+                                    color="dark"
+                                    href="https://docs.lightdash.com/guides/adding-tables-to-lightdash"
```
rephus, line 455: should we add a `target='_blank'` here ?

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: The thing is that this is below the search controls so we don't want to return a whole different tree, just this part. But this could go in the CatalogTree component and that could return this as an invalid state if you prefer that. I thought about doing it that way, but went with this do sort of centralize the fetch/empty state logic.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: Yea, good call. Done. I think that might fix the link out as well? It works locally.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
magnew: This is pretty standard react but if there is a pattern about this we like better, doing it differently is fine with me.

rephus: This is ok, I was wondering if there is a best practice out there for this

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
joaoviana: Styled the button here as well :https://github.com/lightdash/lightdash/pull/10264/commits/ce2436a5f3aad1bbfc7cb3b0574c41f7165d962e

joaoviana (approved):

PR #10265, opened 2024-05-30
fix: defer state when opening metadata sidebar

joaoviana:
Closes: N/A

### Description:

We could start opening the sidebar before waiting for the data to be returned. 
Added a temporary loading overlay (I'm sure this is being worked on a separate ticket)

This uses `useTransition` and we apply it to: 
- when we store the metadata

This unblocks the UI thread while the processing is happening 
Displays a loading indicator on the usage analytics + on the whole metadata sidebar

Demo: 

https://github.com/lightdash/lightdash/assets/7611706/46010d77-2c53-43d0-94a4-fea078aa595b

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
@@ -24,7 +24,7 @@ import {
     IconLink,
     IconTable,
 } from '@tabler/icons-react';
-import { useIsFetching } from '@tanstack/react-query';
+import { useIsMutating } from '@tanstack/react-query';
 import MarkdownPreview from '@uiw/react-markdown-preview';
 import { useEffect, useMemo, useState, type FC } from 'react';
 import { useHistory } from 'react-router-dom';
@@ -41,6 +41,7 @@ export const CatalogMetadata: FC = () => {
     const {
         projectUuid,
         metadata: metadataResults,
+        setMetadata,
         setSidebarOpen,
         analyticsResults,
         selection,
@@ -50,9 +51,11 @@ export const CatalogMetadata: FC = () => {
 
     const { reset: resetMetadata } = useCatalogMetadata(projectUuid);
 
-    const isFetchingAnalytics = useIsFetching({
-        queryKey: ['catalog_analytics', projectUuid],
-    });
+    const isMutatingAnalytics = useIsMutating([
+        'catalog_analytics',
+        projectUuid,
+    ]);
+    const isMutatingMetadata = useIsMutating(['catalog_metadata', projectUuid]);
 
     const history = useHistory();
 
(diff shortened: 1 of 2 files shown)


rephus:
If you navigate through items fast you can get some weird `fields` , can you replicate ? 

image

rephus:
Joins are no longer alligned, is this new ? 

image

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -97,6 +97,7 @@ function getSelectionList(tree: CatalogTreeType): CatalogSelection[] {
 }
 
 export const CatalogPanel: FC = () => {
+    const [, startTransition] = useTransition();
```
rephus, line 100: could you explain again what's the benefit of using this ? I can't notice any change even on slow 3g network settings.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
```diff
@@ -123,18 +124,18 @@ export const CatalogPanel: FC = () => {
...
+                setMetadata(data);
+            });
         }
     });
     const { mutate: getAnalytics } = useCatalogAnalytics(
         projectUuid,
         (data) => {
             if (data) {
-                setAnalyticsResults(data);
+                startTransition(() => {
```
rephus, line 136: I'm not exactly sure what's happening here, but my guess is that this analytics can take longer than `metadata` request . And we should not block the metadata panel while this analytic results is being requested, just so you know. We could add a `delay` on the backend to simulate what might happen.

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
```diff
@@ -90,7 +91,7 @@ export const CatalogMetadata: FC = () => {
         }
     }, [metadataResults, selection, selectedFieldInTable]);
 
-    if (!metadata) return null;
+    if (!metadata) return <LoadingOverlay visible />;
```
rephus, line 94: I can see this, only once, on the first metadata request, but not between metadata requests (on slow 3g network settings) , is this intended ? 

Perhaps we should `reset` the metadata when navigating, to force this overlay to appear between requests.

packages/frontend/src/features/catalog/components/CatalogPanel.tsx
joaoviana: Sure! 
This helps when opening/closing the sidebar. You can notice that when opening/closing the sidebar and testing out the before and after - specially on a 3g network. This is also visible in our own Analytics instance.

Since the value is deferred, the UI thread won't be blocked when processing the `metatada` - from the API and then the calculation in the `useMemo` in `CatalogMetadata`

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
joaoviana: I just tested this now https://github.com/lightdash/lightdash/pull/10265/commits/9d1ceba1ff7e9f4b165137227f6d9744e7cf6a5c and with a slow/fast 3g it will display the overlay. I recommend we try this on Analytics

rephus (approved):

Issue #10267, opened 2024-05-30
Pivot Table Row Grouping and Subtotals


agha4to:
Similar to non-Pivot Tables, provide a Chart configuration option to "Show subtotals" which will group rows on dimension columns and subtotal metrics as appropriate.

PR #10269, opened 2024-05-30
feat: pivot table row grouping and subtotals

agha4to:
Closes: https://github.com/lightdash/lightdash/issues/10267

### Description:
Adds row grouping and subtotal functionality to Pivot Tables. Pivot Table and non-Pivot Table components have distinct implementations, and ideally we'd like to combine them. This code does not do that. Instead, it keeps them separate. The underlying Pivot Table data structures have been "retrofitted" rather than rewritten at this time. I think that task is better done in the future when combining the two components, to avoid too much wasted effort.

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -53,6 +53,12 @@ const rowColumn: TableColumn = {
     enableGrouping: false,
 };
 
+const allPivotedSpacerColumn: TableColumn = {
+    id: 'all-pivoted-spacer',
+    cell: () => null,
+    enableGrouping: false,
+};
+
 const VirtualizedArea: FC<{
     cellCount: number;
     height: number;
@@ -94,6 +100,11 @@ const PivotTable: FC<PivotTableProps> = ({
     const hasRowTotals = data.pivotConfig.rowTotals;
 
     const { columns, columnOrder } = useMemo(() => {
+        // Pivoting all dimensions requires a spacer column under the pivoted headers.
+        const allDimensionsPivoted =
+            data.indexValueTypes.length === 0 &&
+            data.titleFields[0].length === 1;
+
         const indexPlaceholders: Record<string, ResultValue>[] = Array(
             data.indexValueTypes.length,
         ).fill({});
@@ -122,6 +133,8 @@ const PivotTable: FC<PivotTableProps> = ({
 
         const newColumnOrder: string[] = [];
         if (!hideRowNumbers) newColumnOrder.push(ROW_NUMBER_COLUMN_ID);
+        if (allDimensionsPivoted)
+            newColumnOrder.push(allPivotedSpacerColumn.id);
 
         let newColumns = data.retrofitData.pivotColumnInfo.map(
             (col, colIndex) => {
@@ -205,6 +218,8 @@ const PivotTable: FC<PivotTableProps> = ({
             },
         );
 
+        if (allDimensionsPivoted)
+            newColumns = [allPivotedSpacerColumn, ...newColumns];
         if (!hideRowNumbers) newColumns = [rowColumn, ...newColumns];
 
         return { columns: newColumns, columnOrder: newColumnOrder };
@@ -212,6 +227,7 @@ const PivotTable: FC<PivotTableProps> = ({
         data.retrofitData.pivotColumnInfo,
         data.indexValueTypes.length,
         data.headerValues,
+        data.titleFields,
         getField,
         hideRowNumbers,
     ]);
@@ -491,16 +507,6 @@ const PivotTable: FC<PivotTableProps> = ({
 
                     return (
                         <Table.Row key={`row-${rowIndex}`} index={rowIndex}>
-                            {/* renders empty rows if there are no index values but titles */}
-                            {data.indexValueTypes.length === 0 &&
-                                data.titleFields[0].map(
-                                    (_titleField, titleFieldIndex) => (
-                                        <Table.Cell
-                                            key={`empty-title-${rowIndex}-${titleFieldIndex}`}
-                                        />
-                                    ),
-                                )}
-
                             {row.getVisibleCells().map((cell, colIndex) => {
                                 const meta = cell.column.columnDef.meta;
                                 let item = meta?.item;
@@ -761,7 +767,7 @@ const PivotTable: FC<PivotTableProps> = ({
                                         {value.formatted}
                                     </Table.CellHead>
                                 ) : (
-                                    <Table.Cell
+                                    <Table.CellHead
                                         key={`footer-total-${totalRowIndex}-${totalColIndex}`}
                                     />
                                 );
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -94,20 +94,29 @@ const PivotTable: FC<PivotTableProps> = ({
     const hasRowTotals = data.pivotConfig.rowTotals;
 
     const { columns, columnOrder } = useMemo(() => {
-        let headerInfoForColumns = data.headerValues[0].map(
-            () => ({} as Record<string, ResultValue>),
-        );
-        data.headerValues.forEach((headerRow) => {
-            headerRow.forEach((headerColValue, headerColIndex) => {
-                if ('value' in headerColValue) {
-                    headerInfoForColumns[headerColIndex][
-                        headerColValue.fieldId
-                    ] = headerColValue.value;
-                }
-            });
-        });
-        headerInfoForColumns = [
-            ...Array(data.indexValueTypes.length),
+        const indexPlaceholders: Record<string, ResultValue>[] = Array(
+            data.indexValueTypes.length,
+        ).fill({});
+
+        const headerPlaceholders: Record<string, ResultValue>[] = Array(
+            data.headerValues.length,
+        ).fill({});
+
+        const headerInfoForColumns = data.headerValues.reduce<
+            Array<Record<string, ResultValue>>
+        >((acc, headerRow) => {
+            return headerRow.map((headerColValue, headerColIndex) =>
+                'value' in headerColValue
+                    ? {
+                          ...acc[headerColIndex],
+                          [headerColValue.fieldId]: headerColValue.value,
+                      }
+                    : acc[headerColIndex],
+            );
+        }, headerPlaceholders);
+
+        const finalHeaderInfoForColumns = [
+            ...indexPlaceholders,
             ...headerInfoForColumns,
         ];
 
@@ -179,8 +188,8 @@ const PivotTable: FC<PivotTableProps> = ({
                             item: item,
                             type: col.columnType,
                             headerInfo:
-                                colIndex < headerInfoForColumns.length
-                                    ? headerInfoForColumns[colIndex]
+                                colIndex < finalHeaderInfoForColumns.length
+                                    ? finalHeaderInfoForColumns[colIndex]
                                     : undefined,
                         },
                         aggregationFn: aggregationFunction,
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -608,11 +608,10 @@ const PivotTable: FC<PivotTableProps> = ({
                                         )}
                                     >
                                         {cell.getIsGrouped() ? (
-                                            <Group spacing="xxs">
+                                            <Group spacing="two" noWrap>
                                                 <Button
                                                     compact
                                                     size="xs"
-                                                    ff="Inter"
                                                     variant="subtle"
                                                     styles={(theme) => ({
                                                         root: {
@@ -623,6 +622,10 @@ const PivotTable: FC<PivotTableProps> = ({
                                                             paddingRight:
                                                                 theme.spacing
                                                                     .xxs,
+                                                            fontFamily:
+                                                                "'Inter', sans-serif",
+                                                            fontFeatureSettings:
+                                                                "'tnum'",
                                                         },
                                                         leftIcon: {
                                                             marginRight: 0,
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -103,87 +103,92 @@ const PivotTable: FC<PivotTableProps> = ({
         const newColumnOrder: string[] = [];
         if (!hideRowNumbers) newColumnOrder.push(ROW_NUMBER_COLUMN_ID);
 
-        let newColumns = data.retrofitData.firstRowOnly.map((col, colIndex) => {
-            newColumnOrder.push(col.fieldId);
+        let newColumns = data.retrofitData.pivotColumnInfo.map(
+            (col, colIndex) => {
+                newColumnOrder.push(col.fieldId);
 
-            const item = getField(col.underlyingId || col.baseId);
+                const item = getField(col.underlyingId || col.baseId);
 
-            const shouldAggregate =
-                col.meta?.type === 'rowTotal' ||
-                (item &&
-                    isField(item) &&
-                    isMetric(item) &&
-                    [MetricType.SUM, MetricType.COUNT].includes(item.type));
+                const shouldAggregate =
+                    col.meta?.type === 'rowTotal' ||
+                    (item &&
+                        isField(item) &&
+                        isMetric(item) &&
+                        [MetricType.SUM, MetricType.COUNT].includes(item.type));
 
-            // TODO: Remove code duplicated from non-pivot table version.
-            const aggregationFunction = shouldAggregate
-                ? (
-                      columnId: string,
-                      _leafRows: Row<ResultRow>[],
-                      childRows: Row<ResultRow>[],
-                  ) => {
-                      const aggregatedValue = childRows.reduce<number>(
-                          (agg, childRow) => {
-                              const cellValue = childRow.getValue(columnId) as
-                                  | ResultRow[number]
-                                  | undefined;
-                              const rawValue = cellValue?.value?.raw;
+                // TODO: Remove code duplicated from non-pivot table version.
+                const aggregationFunction = shouldAggregate
+                    ? (
+                          columnId: string,
+                          _leafRows: Row<ResultRow>[],
+                          childRows: Row<ResultRow>[],
+                      ) => {
+                          const aggregatedValue = childRows.reduce<number>(
+                              (agg, childRow) => {
+                                  const cellValue = childRow.getValue(
+                                      columnId,
+                                  ) as ResultRow[number] | undefined;
+                                  const rawValue = cellValue?.value?.raw;
 
-                              if (rawValue === null) return agg;
-                              const adder = Number(rawValue);
-                              if (isNaN(adder)) return agg;
+                                  if (rawValue === null) return agg;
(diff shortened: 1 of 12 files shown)


packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -92,24 +160,250 @@ const PivotTable: FC<PivotTableProps> = ({
...
-        (total: unknown, rowIndex: number): ResultValue | null => {
-            const value = last(data.indexValues[rowIndex]);
-            if (!value || !value.fieldId) throw new Error('Invalid pivot data');
-
-            const item = getField(value.fieldId);
-            if (!isSummable(item)) {
-                return null;
+    // TODO: indexValues, dataValues, and rowTotalFields should perhaps be combined in pivotQueryResults.ts
+    //  or elsewhere but doing it here for now.
+    //  Alternatively, this whole pivot table implementation should perhaps be combined with the non-pivot tables.
```
ZeRego, line 165: Yes, I think we should move this logic to `pivotQueryResults.ts` so all the data is calculated in that function and this component is only about rendering and actions.
If the outcome of a hook are values and those values are not updated anywhere else, it makes sense to be a `useMemo` and avoid multiple useState.

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -3,27 +3,84 @@ import {
...
+import {
+    columnHelper,
+    ROW_NUMBER_COLUMN_ID,
+    type TableColumn,
+} from '../Table/types';
 import TotalCellMenu from './TotalCellMenu';
 import ValueCellMenu from './ValueCellMenu';
 
+// TODO: Remove code duplicated from non-pivot table version.
+// Adapted from https://stackoverflow.com/a/45337588
```
ZeRego, line 56: We can export the methods from the other file and reuse them here.

ZeRego (commented):
I would appreciate if you rebase this PR after the updates 🙏 🙇

packages/frontend/src/components/common/PivotTable/index.tsx
agha4to: - Moved data structure "retrofit" to pivotQueryResults.ts
- Changed useState to useMemo where applicable
OK, I was thinking that pivotQueryResults.ts could be reworked to remove the "retrofit" when pivot and non-pivot components are combined.

agha4to:
@ZeRego, thank you for your patience. Refactored per your recommendations and fixed some tests. Thanks again!

almeidabbm:
Hi @agha4to 👋 

I have been testing this and stumbled across a small bug:
I cannot change the order of the row dimensions. Don't think this is related to this PR itself, but allowing to do it would help with displaying the correct row grouping

https://github.com/user-attachments/assets/0da44252-4952-4244-8be7-ddac206df1a4

agha4to:
Hi @almeidabbm, I've been thinking about **Rows** order and trying to implement something but I'm coming up against a few design questions.

Currently in Pivots, even without this PR:
- From the Results table, dragging columns reorders them in both the Results and Chart view
- From the pivot Configure chart sidebar, reordering items under Columns reorders the pivoted items at the top of the Chart

What's proposed:
- From the pivot Configure chart sidebar, reordering items under Rows should reorder columns in the Chart (It's somewhat confusing that reordering Rows would change column order, but it is standard for pivot tables.)

Questions:

In Configure chart, should **Rows** drags reflect in the Chart table only or also Results? A couple possibilities:

- A) Change only Chart because 1) it is after all the Chart config and 2) if dragging Rows changed Results, how would dragging Columns be handled? If we go this route, should drags of Results columns override the Configure chart Rows order and change Chart?
- B) Change both Chart and Results because it is confusing if the column orders differ between the Results and Charts. If so, again, how would dragging Columns in Chart config be handled to change Results?

What about Metrics? Shouldn't those columns also be draggable to reorder the Chart columns?

I think we need a design spec for all of this. There may be other loose ends to consider too.

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -53,63 +83,138 @@ const PivotTable: FC<PivotTableProps> = ({
...
+            ...headerInfoForColumns,
+        ];
+
+        const newColumnOrder: string[] = [];
+        if (!hideRowNumbers) newColumnOrder.push(ROW_NUMBER_COLUMN_ID);
+
+        let newColumns = data.retrofitData.firstRowOnly.map((col, colIndex) => {
+            newColumnOrder.push(col.fieldId);
+
+            const item = getField(col.underlyingId || col.baseId);
```
almeidabbm, line 109: @agha4to can `firstRowOnly` be typed in `pivot.ts`? I see it as `any[]` there.

This might cause issues when trying to access its properties - e.g. in this line `getField` needs a `string` argument but we might be trying to use a `column` object that doesn't have either `underlyingId` or `baseId` - seeing `packages/frontend/src/hooks/pivotTable/pivotQueryResults.ts:262`

packages/frontend/src/hooks/pivotTable/pivotQueryResults.ts
```diff
@@ -192,11 +203,161 @@ const getColSpanByKey = (
...
+                    },
+                };
+            }
+            return {
+                ...cell,
+                meta: {
+                    type: 'indexValue',
+                },
+            };
+        });
```
almeidabbm, line 283: related to comment above

almeidabbm (changes requested):

almeidabbm:
> Hi @almeidabbm, I've been thinking about **Rows** order and trying to implement something but I'm coming up against a few design questions.
> 
> Currently in Pivots, even without this PR:
> 
> * From the Results table, dragging columns reorders them in both the Results and Chart view
> * From the pivot Configure chart sidebar, reordering items under Columns reorders the pivoted items at the top of the Chart
> 
> What's proposed:
> 
> * From the pivot Configure chart sidebar, reordering items under Rows should reorder columns in the Chart (It's somewhat confusing that reordering Rows would change column order, but it is standard for pivot tables.)
> 
> Questions:
> 
> In Configure chart, should **Rows** drags reflect in the Chart table only or also Results? A couple possibilities:
> 
> * A) Change only Chart because 1) it is after all the Chart config and 2) if dragging Rows changed Results, how would dragging Columns be handled? If we go this route, should drags of Results columns override the Configure chart Rows order and change Chart?
> * B) Change both Chart and Results because it is confusing if the column orders differ between the Results and Charts. If so, again, how would dragging Columns in Chart config be handled to change Results?
> 
> What about Metrics? Shouldn't those columns also be draggable to reorder the Chart columns?
> 
> I think we need a design spec for all of this. There may be other loose ends to consider too.

I see, thank you for the detailed explanation
I agree that it might take a bit more thought on how to implement this, so since this isn't exactly tied to this PR I will create a new ticket from this comment 👍

packages/frontend/src/components/common/PivotTable/index.tsx
agha4to: Good suggestion. I got rid of the any-s and simplified things in the process. Please let me know of any other recommendations. Thank you 👍

almeidabbm (commented):
Couple things causing build errors related to types, otherwise types look good

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -53,63 +83,143 @@ const PivotTable: FC<PivotTableProps> = ({
...
+        if (!hideRowNumbers) newColumnOrder.push(ROW_NUMBER_COLUMN_ID);
+
+        let newColumns = data.retrofitData.pivotColumnInfo.map(
+            (col, colIndex) => {
+                newColumnOrder.push(col.fieldId);
+
+                const item = getField(col.underlyingId || col.baseId);
+
+                const shouldAggregate =
+                    col.meta?.type === 'rowTotal' ||
```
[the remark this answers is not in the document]
agha4to: Thank you. I don't know how I missed that 😞

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -53,63 +83,143 @@ const PivotTable: FC<PivotTableProps> = ({
...
+        ];
+
+        const newColumnOrder: string[] = [];
+        if (!hideRowNumbers) newColumnOrder.push(ROW_NUMBER_COLUMN_ID);
+
+        let newColumns = data.retrofitData.pivotColumnInfo.map(
+            (col, colIndex) => {
+                newColumnOrder.push(col.fieldId);
+
+                const item = getField(col.underlyingId || col.baseId);
```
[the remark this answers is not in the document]
agha4to: Thanks again. Changed it up a bit to leave `item` undefined if there's no `underlyingId` or `baseId` since `item` is not needed unless those IDs are available.

agha4to:
Found/fixed bug when drilling down when subtotals disabled and metrics as rows enabled. Header info needed to be analyzed and available to the drilldown function.

IrakliJani:
Hi @agha4to, amazing work as always. I am almost done with the code review, but in the meantime, there’s a small feature bug that I found:

As you can see here, I have expanded the row, and the actual value is `-` but it is still aggregated as $0.00. This might confuse someone, as they might think the underlying value is 0, but in fact, it’s null.

CleanShot 2024-07-31 at 21 03 13@2x

What I'm describing is:
- if all values are null, then the sum of those values should be displayed as null.
- If all values are numbers - sum them up.
- If it’s a mix (numbers and nulls) - also sum them up.

agha4to:
Thank you, @IrakliJani. I'll look into making the change. Note, I think we will have to also make this change in the non-pivot table code. Prior discussion: https://github.com/lightdash/lightdash/pull/9160#issuecomment-2027536401

agha4to:
Null handling change **_not_** made in the non-pivot implementation, but please let me know if I should in this PR. Changes required there would essentially be identical to changes done here. Thank you!

IrakliJani:
continuing review

agha4to:
Nice, @IrakliJani. Please note the wrapping after the number in parentheses is an issue in the **non**-pivot table implementation too. So I think we have 2 things to fix there, but I don't know if we want to sneak those into this PR:
1) null handling
2) no wrap after parentheses

IrakliJani:
@agha4to oh I see.. we can deal with that separately.

packages/frontend/src/hooks/pivotTable/pivotQueryResults.ts
```diff
@@ -192,11 +204,161 @@ const getColSpanByKey = (
...
+    const indexValues = data.indexValues.length ? data.indexValues : [[]];
+    const baseIdInfo = last(data.headerValues);
+    const uniqueIdsForDataValueColumns: string[] = Array(
+        data.headerValues[0].length,
+    );
+
+    data.headerValues.forEach((headerRow) => {
+        headerRow.forEach((headerColValue, colIndex) => {
+            uniqueIdsForDataValueColumns[colIndex] =
+                uniqueIdsForDataValueColumns[colIndex] +
```
IrakliJani, line 221:
```suggestion
                (uniqueIdsForDataValueColumns[colIndex] ?? '') + 
```
should this be

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -517,7 +690,7 @@ const PivotTable: FC<PivotTableProps> = ({
                             index={totalRowIndex}
                         >
                             {/* shows empty cell if row numbers are visible */}
-                            {hideRowNumbers ? null : <Table.Cell />}
+                            {hideRowNumbers ? null : <Table.CellHead />}
```
IrakliJani, line 693: iirc CellHead renders th instead of td which would be semantically inaccurate. what's the reason of using it here?

packages/frontend/src/hooks/pivotTable/pivotQueryResults.test.ts
```diff
@@ -91,12 +91,64 @@ describe('Should pivot data', () => {
             pivotConfig,
             titleFields: [[{ direction: 'header', fieldId: 'page' }], [null]],
             cellsCount: 7,
+            retrofitData: {
+                allCombinedData: [
+                    {
+                        undefinedpage__views__0: {
```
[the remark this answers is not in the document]
agha4to: Yes, using the string "undefined" which looks odd, but the idea is to have a unique key and "undefined" denotes the true value for that column. We could just use blank text `""` if undefined.

packages/frontend/src/hooks/pivotTable/pivotQueryResults.ts
agha4to: Hah, I just suggested that above without realizing you wrote this.

packages/frontend/src/components/common/PivotTable/index.tsx
agha4to: There was a reason, but I cannot remember. Something to do with formatting. Let me see if I can figure it out.

agha4to: OK, I think it is because the cell in the bottom left renders in white if we use Cell:
image
With CellHead:
image

agha4to: You're right about th instead of td, but the other row total "cells" also render as th:
image

IrakliJani:
Thank you for the changes, @agha4to. I’ve finished the code review, and it looks great. Tests are all ✅ as well 🙌  !!!

There are a few minor bugs that I noticed during further testing since we switched from no-library to react-table:

- There is an unnecessary hash sign and "TH gray color" issue. I believe this is new and doesn’t look nice when there’s an empty cell next to it (see the screenshot below).
CleanShot 2024-08-13 at 14 29 29@2x

- The styling of the TH (gray-colored cells) is inconsistent. I think 1 should be white, and 2, 3 and 4 should be gray.
CleanShot 2024-08-13 at 14 20 00@2x

- The UI input for subtotals/metrics as rows is correctly displayed, but the state values are incorrect. I checked the conditions when they are enabled or not, and they are correct. However, the state should also update once certain conditions change to avoid impossible UI states
CleanShot 2024-08-13 at 14 21 32@2x

Let me know if these make sense and if you need any help with the issues! 🙌

IrakliJani:
I found one major bug: row number appearing in the second column:
see the video so you can reproduce on your end:

https://github.com/user-attachments/assets/e6c92ff2-203d-47da-a698-9a397fb2bc0b

agha4to:
Awesome finds, @IrakliJani! I was able to fix these two but want to test so will push code here hopefully by eow:
1) Row number in wrong column
2) Checkbox UI impossible state

And regarding these other two:

3) Hash symbol background is grayed because it's the header for row number, and consistent with non-pivot table implementation. Please lmk and I can make it white, although then we should probably change the non-pivot version.
4) Current production build has the column total zebra shading (and lack of # header btw / see below) so not sure if there is some desired logic to it. Again please lmk and I can make it all gray, except the bottom cell under #. Although, perhaps just make the whole row gray including that first cell? Whatever you say 😉 np 👍 lmk
image

agha4to:
Here's a look at all gray column totals / bottom row:
image

Please also note that in the non-pivot version, the column totals / bottom row is all white, and the word "Total" is all the way to the left:
image

There are other variations between pivot and non-pivot implementations:
- non-pivot row number column is left-justified
- non-pivot Table name in header is not bold
- null handling, mentioned above
- no wrap after parentheses, mentioned above
- very likely more than this

IrakliJani:
> Here's a look at all gray column totals / bottom row: image
> 
> Please also note that in the non-pivot version, the column totals / bottom row is all white, and the word "Total" is all the way to the left: image
> 
> There are other variations between pivot and non-pivot implementations:
> 
>     * non-pivot row number column is left-justified
> 
>     * non-pivot Table name in header is not bold
> 
>     * null handling, mentioned above
> 
>     * no wrap after parentheses, mentioned above
> 
>     * very likely more than this

thanks @agha4to 

It's fine as it is now. no need to have parity between the two at the moment.

I created a ticket based on your comment here: https://github.com/lightdash/lightdash/issues/11187 feel free to add more there 🙌

IrakliJani:
> Awesome finds, @IrakliJani! I was able to fix these two but want to test so will push code here hopefully by eow:
> 
>     1. Row number in wrong column
> 
>     2. Checkbox UI impossible state
> 
> 
> And regarding these other two:
> 
>     3. Hash symbol background is grayed because it's the header for row number, and consistent with non-pivot table implementation. Please lmk and I can make it white, although then we should probably change the non-pivot version.
> 
>     4. Current production build has the column total zebra shading (and lack of # header btw / see below) so not sure if there is some desired logic to it. Again please lmk and I can make it all gray, except the bottom cell under #. Although, perhaps just make the whole row gray including that first cell? Whatever you say 😉 np 👍 lmk
>        image

1 & 2 awesome! 

3. A hash is not a requirement, and we don't have to be consistent with the non-pivot table. I don't think it plays any role or adds any visual aid, especially when pivoting on multiple levels.

4. You can ignore it if it's the same as the prod build. The idea here is to have no styling (white background + blank cell) on empty cells that have no value.

For example:
- A footer row with a total should be gray.
- A footer row with a null total should still be gray.
- A footer row with a label should still be gray.
- A footer row that displays nothing should be white.

The same goes for column totals, the index column, and headers.

@agha4to Did you push the code? I can't see a new commit to check the diff.

agha4to:
QA should be ✅ tomorrow I think. Will update w/ screenshots and we can go from there!
Thanks for https://github.com/lightdash/lightdash/issues/11187; would be nice to combine Pivot and non-Pivot implementations to avoid discrepancies
Thank you, @IrakliJani.

IrakliJani:
thanks @agha4to 🙌

agha4to:
Came across another possible issue with checkbox state so looking into that.

Regarding UI though:

1) I left the `#` but we can certainly change it. It may help users answer with "What are these numbers in the first column?" but then `#` is not an international symbol for "number" so perhaps confusing to some.
image

---

2) I think the white cell shown below between `#` and `Payments Total revenue` helps distinguish between headers referring to columns and headers referring to pivoted dimensions. Perhaps headers for pivoted items should have a colon after them like `Orders Order date year:` or other visual cue besides just right alignment.
image

---

3) Bottom row shading may still need some changes:
image

agha4to:
It turns out the 2 issues from QA are unrelated to any changes from this PR:
1) Pivot table does not update after filter(s) applied and query run
2) "Show row totals" & "Show column totals" checkboxes change values unexpectedly, due to mixup [here](https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx#L71-L77) which looks like an easy fix btw

IrakliJani:
@agha4to sorry for the long pause.. we didn’t have the capacity to give a final review and a QA

Is there anything remaining before we approve the code changes?

here are minor UI issues I noticed during QA:

CleanShot 2024-09-25 at 12 53 55@2x

IrakliJani:
render link: https://lightdash-pr-11681.onrender.com

agha4to:
I heard y'all were busy. I don't think there's a rush--no problem. Afaik, no remaining issues. Here's a look at the recent shading changes:

1. Last row of Row # column is **no longer** shaded
2. Last row of dimension columns are **no longer** shaded
3. Last row of metric columns that do not have column totals are **no longer** shaded
4. Row total columns, not including last row, **are** shaded

image

IrakliJani (approved):

IrakliJani:
@agha4to approving code changes ✅ great work 🙌

almeidabbm (approved):

Issue #10274, opened 2024-05-31
Cannot read properties of undefined (reading 'label') 


ZeRego:
### Description

Deploying a project throws this error for some models.

### Steps to Reproduce the Bug or Issue

still investigating reproduction steps 

### version

0.1112.0

### Cloud or self-hosting

None

PR #10275, opened 2024-05-31
fix: compile explore with join with fields and groups

ZeRego:
Closes: #10274 

### Description:

Reproduction steps:
- Have 2 models in your YML
- Table A joins table B with a subset of fields

``` 
       - join: table_b
          sql_on: ${table_a.user_id} = ${table_b.user_id}
          fields: [user_id, date]
```
- In your table B you need a date dimension with time intervals

```
      - name: date
        description: 'Date of the forecasting.'
        meta:
          dimension:
            type: date
            time_intervals: ['DAY', 'WEEK', 'MONTH', 'QUARTER']
```

### Reviewer actions

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.ts
+++ b/packages/common/src/compiler/exploreCompiler.ts
@@ -178,26 +178,24 @@ export class ExploreCompiler {
                                 requiredDimensionsForJoin.includes(
                                     dimensionKey,
                                 );
-                            const tableGroups =
-                                tables[join.table].groupDetails || {};
-                            const allGroupsIncluded =
+
+                            const isTimeIntervalBaseDimensionVisible =
+                                dimension.timeInterval &&
                                 dimension.groups &&
-                                dimension.groups.every(
-                                    (group) =>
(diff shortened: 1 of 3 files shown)


packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -178,26 +178,24 @@ export class ExploreCompiler {
...
+
+                            const isTimeIntervalBaseDimensionVisible =
+                                dimension.timeInterval &&
                                 dimension.groups &&
-                                dimension.groups.every(
-                                    (group) =>
-                                        join.fields !== undefined &&
-                                        (join.fields.includes(group) ||
-                                            join.fields.includes(
-                                                tableGroups[group].label,
```
ZeRego, line 190: The problem was that `tableGroups[group]` might be undefined. 
We only need to check the last value from `dimension.groups` which is the dimension id of the base dimension from a time interval dimension.
Refactored logic.

jpetey75 (approved):

Issue #10277, opened 2024-05-31
Rename group properties 


ZeRego:
### Description

We are also making 2 small changes to the YML groups syntax:
- field group will be named groups
- model groups will be named group_details

### What problem does this solve?

This will allow us, in the future, to use the same terminology for nested groups for models.
It also makes it clear to developers that it supports multiple values.

PR #10279, opened 2024-05-31
feat: rename YML group properties

ZeRego:
Closes: [#10277](https://github.com/lightdash/lightdash/issues/10277)

### Description:

Changes:
- field `group` renamed to `groups`
- update field docs
- model `groups` renamed to `group_details`
- update model docs
- update translator code + tests
- update demo yml

### Reviewer actions


ZeRego:
> Are these 2 properties `new` ? so nobody is using them at the moment and it is ok to change it without causing issues on existing customers, right ?
> 
> Apart from that, it seems ok.

Correct. Making this change before we announce it to everyone.

PR #10282, opened 2024-05-31
fix: missing time interval dimensions in joined tables

ZeRego:
Closes: [#10274](https://github.com/lightdash/lightdash/issues/10274)

### Description:

We now have a property to know what is the time interval dimension base dimension. 
This way we can easily know if the base dimension is available in a joined table. 
This property will also be useful in other places where we currently split the time interval dimension name to figure out the base dimension.

Reproduce steps:

Add join with a date field

```
       - join: table_b
          sql_on: ${table_a.user_id} = ${table_b.user_id}
          fields: [user_id, date]
```

Before:
time interval dimensions would not show up in joined tables

After:
time interval dimensions show in joined tables

### Reviewer actions


packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -181,12 +181,10 @@ export class ExploreCompiler {
 
                             const isTimeIntervalBaseDimensionVisible =
                                 dimension.timeInterval &&
-                                dimension.groups &&
+                                dimension.timeIntervalBaseDimensionName &&
                                 join.fields
                                     ? join.fields.includes(
-                                          dimension.groups[
-                                              dimension.groups.length - 1
-                                          ],
```
ZeRego, line 189: The problem with this logic is that the time interval dimension doesn't store the base dimension name in the `groups` but it's label. So this was never a match.

almeidabbm (approved):
Tested locally, LGTM 🚀 

Before:
Screenshot 2024-05-31 at 16 22 35

After:
Screenshot 2024-05-31 at 16 23 21

Issue #10271, opened 2024-05-31
Improve table icon + table name + joins icon alignment


PriPatel:
![image](https://github.com/lightdash/lightdash/assets/12776522/ec815a01-5e51-49b6-8222-c763bc6b21cb)
UI alignment issues, should we start using tables/or a grid of some sort?

PR #10283, opened 2024-05-31
fix: make TableListItem responsive

joaoviana:
Closes: #10271

### Description:

- Changes highlight colors from `violet` to `yellow`
- Truncates table name and if it's truncated, we display a tooltip
- Use Mantine's `Grid` to align all items accordingly
- `lineClamp`s the error messages
- Resets `selection` when user clicks on error table.

https://github.com/lightdash/lightdash/assets/7611706/53f7c22c-5b90-4528-b6d3-daeb3669f51d

### Reviewer actions


rephus:
> Resets selection when user clicks on error table. 

I don't think feels right if you're using keyboard navigation

[Screencast from 31-05-24 13:53:34.webm](https://github.com/lightdash/lightdash/assets/1983672/e7d9d0cf-35ce-478a-bd2e-5a98bd987868)

rephus:
Do we only show the first table description per group ?

image

no description in users or stg_orders

rephus:
I can se a bunch of useless scroll bars here 

image

rephus:
this join doesn't have a tooltip if too long, ri ght ? 

image

rephus:
Are we now using `Nº` instead of `123` ? 

image

rephus:
Not sure how bad this is, or what can we do to fix it .  but keeping pressed `up` or `down` it jumps a random number of items in the list.

rephus:
Are we not showing the description or  a tag `description matched` if a field desription matches the search ? 

image

rephus:
This is probably not in scope, but opening this page in mobile, redirects to /home. I don't think this is suppose to be mobile compatible, right ?

rephus (commented):
Code looks good, the only thing feels weird is the `unselect` of table errors

joaoviana:
> This is probably not in scope, but opening this page in mobile, redirects to /home. I don't think this is suppose to be mobile compatible, right ?

Not in scope, you're correct

> > Resets selection when user clicks on error table.
> 
> I don't think this feels right if you're using keyboard navigation
> 
> If you don't like the empty `metadata`panel when selected, we should show the error in the sidebar instead.

Done here: https://github.com/lightdash/lightdash/pull/10288




> Do we only show the first table description per group ?
> 
> 
> no description in users or stg_orders
> 
> and no descriptions here
> 

This is an issue with the `searchCatalog` on the BE - it's not returning `table`s so it doesn't carry the `description` from what I understand. It does for `customers`, doesn't do it for `crea` search. This should be address on a separate PR







> I can se a bunch of useless scroll bars here
> 

This tackles the first bit on this: https://github.com/lightdash/lightdash/pull/10286
I don't understand really how to manipulate the native scrollbars from appearing on the right sidebar for you. I need to investigate...






> this join doesn't have a tooltip if too long, ri ght ?
> 

Done! https://github.com/lightdash/lightdash/pull/10283/commits/62ef02ce27235cc35f6c883fe82da67346464abc





> Are we now using `Nº` instead of `123` ?
> 

This was just a temporary placeholder as I was waiting for the new icons, but this has been de-prioritised! I've replaced with the `123`




> Not sure how bad this is, or what can we do to fix it . but keeping pressed `up` or `down` it jumps a random number of items in the list.

I think we should throttle the API requests - which I think it's the reason why it randomly jumps some indices :/ 



> Are we not showing the description or a tag `description matched` if a field desription matches the search ?
> 
> image

I don't know about the requirements of displaying `description`s of fields, checking

rephus (approved):

PR #10293, opened 2024-06-03
chore: don't throw error for out of sync migrations

joaoviana:
Closes: [#10290](https://github.com/lightdash/lightdash/issues/10290)

### Description:

Triggered `Database has not been migrated yet` error. See the `after` where the event is discarded

Before 

Screenshot 2024-06-03 at 14 19 23

After 

Screenshot 2024-06-03 at 14 21 06

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/HealthService/HealthService.ts
+++ b/packages/backend/src/services/HealthService/HealthService.ts
@@ -4,7 +4,6 @@ import {
     LightdashInstallType,
     LightdashMode,
     SessionUser,
-    UnexpectedDatabaseError,
 } from '@lightdash/common';
 import { createHmac } from 'crypto';
 import { getDockerHubVersion } from '../../clients/DockerHub/DockerHub';
@@ -46,9 +45,8 @@ export class HealthService extends BaseService {
             await this.migrationModel.getMigrationStatus();
 
         if (!isComplete) {
-            throw new UnexpectedDatabaseError(
-                'Database has not been migrated yet',
-                { currentVersion },
+            console.warn(
+                `Database migrations are not up to date. Current version: ${currentVersion}`,
             );
         }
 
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/HealthService/HealthService.ts
```diff
@@ -46,9 +45,8 @@ export class HealthService extends BaseService {
             await this.migrationModel.getMigrationStatus();
 
         if (!isComplete) {
-            throw new UnexpectedDatabaseError(
-                'Database has not been migrated yet',
-                { currentVersion },
+            console.warn(
+                `Database migrations are not up to date. Current version: ${currentVersion}`,
```
joaoviana, line 49: Are we ok to keep this as a `warn` for devx @owlas ? 
I can remove it completely, but I'm thinking this could be useful for debugging 🤔

owlas: Useful!

owlas (changes requested):
Just looking at the code for `.status` in knex. It returns the difference between the number of migrations listed in the migrations table (i.e. migrations completed against the database), and the number of migrations defined as code in the repository.

So:
- 0 = all migrations up to date
- 1 or more = there are more DB migrations than there are defined in the code (you are running old code against a newer DB)
- -1 or less = there are more migrations defined in the code than have been run in the database (i.e. you are running new code against an older database).

We can actually be more targeted here. The new policy is that we can run old code against a new database. So that means that having 0 or above, is ok. But if it's below zero we actually should throw an error, because it violates the policy of having the database newer than the code.

https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/lib/migrations/migrate/Migrator.js#L248-L258

owlas (approved):

Issue #10296, opened 2024-06-03
Make sentry sampling configurable


owlas:
Allow `SENTRY_TRACE_SAMPLE_RATE` as an environment variable. This should be passed to the backend and this can then be propagated to the frontend through the health endpoint (the same way that we pick up the enviroment from the backend in the sentry config).

PR #10300, opened 2024-06-03
chore: make sentry traces and profiles sample rate configurable

joaoviana:
Relates to: [#10296](https://github.com/lightdash/lightdash/issues/10296)

Closes: [#10296](https://github.com/lightdash/lightdash/issues/10296)

### Description:

initially, the ticket mentioned to only allow traces sampling rate to be configurable, but I figured we would want the same for profiling in the backend.

Reads: 

```
SENTRY_TRACES_SAMPLE_RATE
SENTRY_PROFILES_SAMPLE_RATE
```

and parses them into: 

```
tracesSampleRate
profilesSampleRate
```

### Reviewer actions

### DIFF
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -419,6 +419,12 @@ const mergeWithEnvironment = (config: LightdashConfigIn): LightdashConfig => {
             release: VERSION,
             environment:
                 process.env.NODE_ENV === 'development' ? 'development' : mode,
+            tracesSampleRate: parseFloat(
+                process.env.SENTRY_TRACES_SAMPLE_RATE || '0.1',
+            ),
+            profilesSampleRate: parseFloat(
+                process.env.SENTRY_PROFILES_SAMPLE_RATE || '0.2',
+            ),
         },
         lightdashSecret,
         secureCookies: process.env.SECURE_COOKIES === 'true',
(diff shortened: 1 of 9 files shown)


packages/backend/src/config/parseConfig.ts
```diff
@@ -419,6 +419,12 @@ const mergeWithEnvironment = (config: LightdashConfigIn): LightdashConfig => {
             release: VERSION,
             environment:
                 process.env.NODE_ENV === 'development' ? 'development' : mode,
+            tracesSampleRate: parseFloat(
```
owlas, line 422: Additional error handling is needed, e.g. `getIntegerFromEnvironmentVariable` https://github.com/lightdash/lightdash/blob/1869560b80b22d0b6c950ef71a63d42a809cfa86/packages/backend/src/config/parseConfig.ts#L13-L13

owlas (commented):
looks good - but error handling needed on float parsing

packages/backend/src/config/parseConfig.ts
joaoviana: changed that method to `parseFloat` - https://github.com/lightdash/lightdash/pull/10300/commits/75be47b7a8e74716dc57a4892e257ff49c7d5cd7 would you prefer a separate method like `getFloatFromEnvironmentVariable`?

owlas: A separate method would be better otherwise we'll lose the current validation. i.e. I can have PGMAXCONNECTIONS=2.5

joaoviana: 💯

owlas (approved):

Issue #1735, opened 2022-04-11
I want to be able to create default filters in my Explores


TuringLovesDeathMetal:
### Description: What is it?

I want to add default filters to my Explores so that when people use them, they automatically have filters applied to them. 

These filters are either:
- empty, but the field is filtered to an empty value so users are pushed to filter on that field
- filled with one or many values 

### Problem: What problem does this solve?

Some datasets are really large or require specific filters to "make sense". For example, for 99% of queries I run using my `accounts` table, I want `account_closed = FALSE`. But, I still want to be able to query _**all**_ accounts (open or closed) for some queries. By default, I want to filter for `account_closed = FALSE` so other folks don't have to remember to add this filter each time, but I want the option to be able to remove it.

Another example from a user:
we are dealing with a massive amount of data partioned by date. We would like to force a filter on the date in order to force the user to apply a date. It will first avoid to have an error because you can not query a partioned table without a where condition on the date

### What: Roughly, what does this look like in the product?

- [ ] users can apply filters to Explores in their dbt projects. 
- [ ] The filters are applied on a model level (i.e. under `model` -->`meta` --> `filters`)
- [ ] the filter syntax is the same as what's used here: https://github.com/lightdash/lightdash/issues/3126
- [ ] there are three types of filters: 
    - `default` filters are applied to the table, but can be overridden in the UI (the same way as dashboard saved filters can be overriden, for example). 
    - `required` filters can have their values changed, but cannot be removed from the filters list in the UI
    - `sql_where` these are fixed filters cannot be changed and cannot be removed - these are actually hidden from the user in the UI. They'd only be visible in the compiled SQL. (this is the type we have already using `sql_where`)

```yaml
models:
  name: users
  meta:
      default_filters: 
          - created_date:
               field: ${created_at_date}
               value: >= '2022-05-21'
       required_filters:
         - created_date_required:
               field: ${created_at_date}
               value: >= '2022-05-21'
      sql_where: ${created_at_date} >= '2022-05-21' # we could change the name of `fixed filter` to `sql filter`
```

OR something like: 

```yaml
models:
  name: users
  meta:
    default_filters:
      - filter: ${created_at_date}
         value: >= '2022-05-21'
         required: no
     sql_filters: ${created_at_date} >= '2022-05-21' # This is a fixed filter - how can we fit this into the other format??
```

### Needs spec:

- what do `required` and `default` filters look like in the UI? 
[...]

PR #10338, opened 2024-06-07
feat: model required filters

mmcordoba:
Closes: 

### Description:

When a model is defined on a large data set, any query made to it can take hours to complete if the data accessed is not filtered by the right partitions/dimensions.

We needed a feature that allow us to define sensible filters for large data set models, that will guarantee that the model can only be queried when at least once filter on the needed dimensions is in place.

This feature implementation solves this problem for us.

Added the model meta property **required_filters**

```
version: 2
models:
  - name: events
    meta:
      group_details:
        events:
          label: Events
          description: Events that happened in the Jaffle Shop
      required_filters:
        - date: inThePast 14 days
 ```

This property follows the simplified filter language https://docs.lightdash.com/references/metrics#filters that has been extended to support the date filter for **in the last** and **in the next** interval

When a Dimension filter is added using this property, this filter will be added to the compiled query unless another filter is defined for the Dimension.
This feature allows us to force that when the model is used at least one filter will always be in place for the filtered Dimension.

Screenshot 2024-06-07 at 06 43 34

Screenshot 2024-06-07 at 06 44 54

### Reviewer actions

### DIFF
--- a/packages/backend/src/queryBuilder.ts
+++ b/packages/backend/src/queryBuilder.ts
@@ -4,6 +4,8 @@ import {
     CompiledCustomSqlDimension,
     CompiledDimension,
     CompiledMetricQuery,
+    CompiledTable,
+    convertFieldRefToFieldId,
     CustomBinDimension,
     CustomDimension,
     DbtModelJoinType,
@@ -28,13 +30,16 @@ import {
     isCompiledCustomSqlDimension,
     isCustomBinDimension,
     isFilterGroup,
+    isFilterRuleDefinedForFieldId,
     ItemsMap,
+    MetricFilterRule,
     parseAllReferences,
     renderFilterRuleSql,
     renderTableCalculationFilterRuleSql,
     SortField,
     SupportedDbtAdapter,
     TimeFrames,
+    UnitOfTime,
     UserAttributeValueMap,
     WarehouseClient,
     WeekDay,
@@ -998,6 +1003,81 @@ export const buildQuery = ({
         return undefined;
     };
 
+    const getNestedDimensionFilterSQLFromModelFilters = (
+        table: CompiledTable,
+        dimensionsFilterGroup: FilterGroup | undefined,
+    ): string | undefined => {
+        const modelFilterRules: MetricFilterRule[] | undefined =
+            table.required_filters;
+        if (!modelFilterRules) return undefined;
+        const reducedRules: string[] = modelFilterRules.reduce<string[]>(
+            (acc, filter) => {
+                // Convert filter to filter rule
+                const filterRule: FilterRule = {
+                    id: filter.id,
+                    target: {
+                        fieldId: convertFieldRefToFieldId(
+                            filter.target.fieldRef,
+                            table.name,
+                        ),
+                    },
+                    operator: filter.operator,
+                    values: filter.values,
+                    settings: {
+                        unitOfTime: filter.settings?.unitOfTime,
+                    },
+                };
+                let filterString: string | undefined;
+                // Required filter is only applied if the filterRule is not already present in the query filters
+                const dimension = Object.values(table.dimensions).find(
+                    (tc) => getItemId(tc) === filterRule.target.fieldId,
+                );
+                if (dimension) {
+                    // Check if the filter rule is already present in the query filters
+                    let dimensionFieldId = filterRule.target.fieldId;
+                    const timeDimension =
+                        dimension.isIntervalBase ||
+                        dimension.timeInterval !== undefined;
+                    // If its a time interval dimension, remove the time interval from the field id
+                    if (!dimension.isIntervalBase && dimension.timeInterval) {
+                        dimensionFieldId = dimensionFieldId.replace(
+                            `_${dimension.timeInterval.toLowerCase()}`,
+                            '',
+                        );
+                        const unitOfTime = dimension.timeInterval.toLowerCase();
+                    }
+                    if (
+                        !(
+                            dimensionsFilterGroup &&
+                            isFilterRuleDefinedForFieldId(
+                                dimensionsFilterGroup,
+                                dimensionFieldId,
+                                timeDimension,
+                            )
+                        )
+                    ) {
+                        filterString = `( ${sqlFilterRule(
+                            filterRule,
+                            FieldType.DIMENSION,
+                        )} )`;
+                    }
+                }
+                if (filterString) {
+                    return [...acc, filterString];
+                }
+                return [...acc];
+            },
+            [],
+        );
+        return reducedRules.join(' AND ');
+    };
+
+    const requiredDimensionFilterSql =
+        getNestedDimensionFilterSQLFromModelFilters(
+            explore.tables[explore.baseTable],
+            filters.dimensions,
+        );
+
     const baseTableSqlWhere = explore.tables[explore.baseTable].sqlWhere;
 
     const tableSqlWhere = baseTableSqlWhere
@@ -1015,8 +1095,15 @@ export const buildQuery = ({
         filters.dimensions,
         FieldType.DIMENSION,
     );
+    const requiredFiltersWhere = requiredDimensionFilterSql
+        ? [requiredDimensionFilterSql]
+        : [];
     const nestedFilterWhere = nestedFilterSql ? [nestedFilterSql] : [];
-    const allSqlFilters = [...tableSqlWhere, ...nestedFilterWhere];
+    const allSqlFilters = [
+        ...tableSqlWhere,
+        ...nestedFilterWhere,
+        ...requiredFiltersWhere,
+    ];
 
     const sqlWhere =
         allSqlFilters.length > 0 ? `WHERE ${allSqlFilters.join(' AND ')}` : '';
--- a/packages/common/src/compiler/translator.ts
+++ b/packages/common/src/compiler/translator.ts
@@ -455,6 +455,7 @@ export const convertTable = (
                 : OrderFieldsByStrategy.LABEL,
         groupLabel: meta.group_label,
         sqlWhere: meta.sql_filter || meta.sql_where,
+        required_filters: parseFilters(meta.required_filters),
         requiredAttributes: meta.required_attributes,
         groupDetails,
     };
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -387,4 +387,39 @@ export const compressDashboardFiltersToParam = (
         { dimensions: [], metrics: [], tableCalculations: [] },
     );
 
+export const isFilterRuleDefinedForFieldId = (
+    filterGroup: FilterGroup,
+    fieldId: string,
+    isInterval: boolean = false,
+): boolean => {
+    // Check if the filter group is an 'and' or 'or' group
+    const items = isAndFilterGroup(filterGroup)
+        ? filterGroup.and
+        : filterGroup.or;
+    // Iterate over each item in the filter group
+    for (let i = 0; i < items.length; i += 1) {
+        const item = items[i];
+        // If the item is a filter rule, check if its id matches the provided filter rule id
+        if (
+            !isFilterGroup(item) &&
+            isInterval &&
+            item.target.fieldId.startsWith(fieldId)
+        ) {
+            return true;
+        }
+        if (!isFilterGroup(item) && item.target.fieldId === fieldId) {
+            return true;
+        }
+        // If the item is a filter group, recursively call the function to check its items
+        if (
+            isFilterGroup(item) &&
+            isFilterRuleDefinedForFieldId(item, fieldId, isInterval)
+        ) {
+            return true;
+        }
+    }
+    // If the filter rule was not found in the filter group, return false
+    return false;
+};
+
 export { ConditionalOperator as FilterOperator };
--- a/packages/common/src/types/filterGrammar.ts
+++ b/packages/common/src/types/filterGrammar.ts
@@ -7,6 +7,7 @@ export type ParsedFilter = {
     type: string;
     values: any[];
     is?: boolean;
+    date_interval?: string;
 };
 
 const filterGrammar = `ROOT
@@ -20,7 +21,7 @@ EMPTY_STRING = '' {
   }
 
 EXPRESSION
-= NUMERICAL / LIST / TERM  
+= NUMERICAL / DATE_RESTRICTION / LIST / TERM  
 
 
 NUMERICAL = SPACE_SYMBOL* operator:OPERATOR SPACE_SYMBOL* value:NUMBER {
@@ -31,6 +32,17 @@ NUMERICAL = SPACE_SYMBOL* operator:OPERATOR SPACE_SYMBOL* value:NUMBER {
    }
 
 OPERATOR = '>=' / '<=' / '>' / '<'
+   
+DATE_RESTRICTION = SPACE_SYMBOL* operator:DATE_OPERATOR SPACE_SYMBOL* value:NUMBER SPACE_SYMBOL* interval:DATE_INTERVAL {
+    return {
+        type: operator,
+        values: [value],
+        date_interval: interval
+    }
+   }
+
+DATE_OPERATOR = 'inThePast' / 'inTheNext'
+DATE_INTERVAL = 'milliseconds' / 'seconds' / 'minutes' / 'hours' / 'days' / 'weeks' / 'months' / 'years'
 
 NUMBER 
   = FLOAT ([Ee] [+-]? INTEGER)?
@@ -189,6 +201,10 @@ export const parseOperator = (
             return FilterOperator.LESS_THAN;
         case '<=':
             return FilterOperator.LESS_THAN_OR_EQUAL;
+        case FilterOperator.IN_THE_PAST:
+            return FilterOperator.IN_THE_PAST;
+        case FilterOperator.IN_THE_NEXT:
+            return FilterOperator.IN_THE_NEXT;
         case 'null':
         case 'NULL':
             return isTrue ? FilterOperator.NULL : FilterOperator.NOT_NULL;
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

File: packages/backend/src/queryBuilder.ts [hunk 1]
```diff
@@ -998,6 +1004,91 @@ export const buildQuery = ({
...
+                    filterRule,
+                    FieldType.DIMENSION,
+                )} )`;
+                return [...acc, filterString];
+            },
+            [],
+        );
+
+        return reducedRules.join(' AND ');
+    };
```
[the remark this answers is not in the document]
mmcordoba: function already refactor and broken down into smaller methods

TuringLovesDeathMetal:
👋 Hey @mmcordoba - this is looking really great 🤩 

The one bit of feedback I have is around the filter not actually appearing in the `filters` tab. 

Today, we have `sql_filter` which adds a filter directly to the SQL (but not to the `filters` tab) and you cannot override it.

Then, the `required_filter` is different in that you can override the filter, but _**not**_ the field. 

Right now, this feels unclear to the user in the UI.

Instead, I think that if there was a way to add the `date` field and filter to the `filters` tab in your example, and it was possible to change the logic, but not remove the filter, then that could be a bit more clear? 

E.g. the normal `delete` option would be greyed out and on hover, it would say `This filter is a required filter. It can't be deleted, but the value can be changed.`

joaoviana:
hey @mmcordoba, I'll review this today :)

mmcordoba:
> hey @mmcordoba, I'll review this today :)

hi @joaoviana i am currently implementing @TuringLovesDeathMetal request, ill let you know once the review is updated with the requested changes. Thank you

mmcordoba:
> > hey @mmcordoba, I'll review this today :)
> 
> hi @joaoviana i am currently implementing @TuringLovesDeathMetal request, ill let you know once the review is updated with the requested changes. Thank you

hi @joaoviana , @TuringLovesDeathMetal requested changes have been implemented, now required filters are shown in the filter card of the ui, they can be changed but not removed.

if a require filter is remove from the meta model, the ui will reflect that as well when the dbt model is refreshed.

The query compiler as well checks the required filter is in place, in case existing charts are run when a new required filter is added.

@joaoviana please have a look into it once you get the time. 
Thank you

joaoviana:
Related issue: https://github.com/lightdash/lightdash/issues/1735

joaoviana (changes requested):
Tested the following scenarios

## Scenario 1

- Add chart using Tracks
- Add filters (target `event`)
- Save
- Add `required_filters`
```
meta:
      required_filters: 
        - event: bash_install_failed
```
- Refresh dbt
- See the filter still there and it took into account the value I had previously
 
------
‼️ Small UI quirk from the scenario above ^ (perhaps we can tackle this on a separate PR)
- The icon of the filter doesn't change from `...` to `X` disabled button when I refresh dbt and that becomes required

Steps to reproduce
1. Create chart with filter A
2. Save
3. Add `required_filter` that references A
4. Refresh dbt
5. See icon `...` 
6. Click on it
7. Turns into `X` and it's properly disabled

**I would consider this low priority though^**

------


## Scenario 2

OR filters 

- If I have `required_filter` A 
- And I have 2 filters A in chart
- Then I can only delete one - the other is marked as non-removable/required

## Scenario 3

Converting filter to filter group, changing from ANY to ALL, then trying to delete it somehow
- I didn't see any issues with this - it worked well 🙌 

## Scenario 4

Can change labels
Filter is updated

tested in `tracks.yml`

```yml
 meta:
      required_filters: 
        - event: bash_install_failed
....

 - name: event 
        description: ""
        meta:
          dimension:
            label: My Event
            type: string
....
```

packages/common/src/types/filterGrammar.ts
```diff
@@ -236,6 +252,9 @@ export const parseFilters = (
                         !!parsedFilter.is,
                     ),
                     values: parsedFilter.values || [1],
+                    settings: {
+                        unitOfTime: parsedFilter.date_interval,
+                    },
```
joaoviana, line 257:
```suggestion
                    ...(parsedFilter.date_interval
                        ? {
                              settings: {
                                  unitOfTime: parsedFilter.date_interval,
                              },
                          }
                        : {}),
```

that way you can remove the extra `undefined` additions in `filterGrammar.test.ts`

packages/common/src/types/filter.ts
```diff
@@ -387,4 +388,39 @@ export const compressDashboardFiltersToParam = (
         { dimensions: [], metrics: [], tableCalculations: [] },
     );
 
+export const isFilterRuleDefinedForFieldId = (
```
joaoviana, line 391: was thinking we could make the steps a bit more explicit. What do you think?

```ts

export const isFilterRuleDefinedForFieldId = (
    filterGroup: FilterGroup,
    fieldId: string,
    isInterval: boolean = false,
) => {
    // Determine if the filter group is an 'and' or 'or' group
    const filterGroupItem = isAndFilterGroup(filterGroup)
        ? filterGroup.and
        : filterGroup.or;

    const isMatchingFieldId = (item: FilterGroupItem) => {
        if (!isFilterGroup(item)) {
            // If the item is not a filter group, check if it matches the fieldId
            return isInterval
                ? item.target.fieldId.startsWith(fieldId)
                : item.target.fieldId === fieldId;
        }
        return false;
    };

    const isFilterRulePresent = (
        item: OrFilterGroup | AndFilterGroup | FilterRule,
    ): boolean => {
        if (isMatchingFieldId(item)) {
            return true;
        }
        if (isFilterGroup(item)) {
            // If the item is a filter group, recursively check its items
            return isFilterRuleDefinedForFieldId(item, fieldId, isInterval);
        }
        return false;
    };

    return filterGroupItem.some(isFilterRulePresent);
};
```

packages/common/src/utils/filters.ts
```diff
@@ -779,3 +784,87 @@ export const addDashboardFiltersToMetricQuery = (
...
+                !requiredFiltersRef.includes(item.target.fieldId)
+            ) {
+                // Remove the filter rule from the filter group
+                item.required = false;
+            } else if (isFilterGroup(item)) {
+                resetRequiredFilterRules(item, requiredFiltersRef);
+            }
+        }
+    }
+};
```
joaoviana, line 870:
```suggestion
export const resetRequiredFilterRules = (
    filterGroup: FilterGroup,
    requiredFiltersRef: string[],
) => {
    // Check if the input is a valid filter group
    if (isFilterGroup(filterGroup)) {
        const items = isAndFilterGroup(filterGroup) ? filterGroup.and : filterGroup.or;
        
        // Iterate over each item in the filter group
        for (const item of items) {
            // If the item is a filter rule, check if its id is not in the required filters reference
            if (isFilterRule(item) && !requiredFiltersRef.includes(item.target.fieldId)) {
                // Mark the filter rule as not required
                item.required = false;
            // If the item is a nested filter group, recursively call the function
            } else if (isFilterGroup(item)) {
                resetRequiredFilterRules(item, requiredFiltersRef);
            }
        }
    }
};
```

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
```diff
@@ -35,9 +40,66 @@ const FiltersCard: FC = memo(() => {
...
-    );
+    const { data } = useExplore(tableName);
+
+    const refreshRequiredFiltersProperty = (
+        allFilter: FilterGroup | undefined,
+    ) => {
+        if (!allFilter) return;
+        if (data && data.tables[tableName]) {
+            const requiredFilters =
+                data.tables[tableName].required_filters || [];
```
joaoviana, line 51:
```suggestion
                data.tables[tableName].requiredFilters || [];
```

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
```diff
@@ -35,9 +40,66 @@ const FiltersCard: FC = memo(() => {
...
+            const allRequiredFilters: FilterRule[] =
+                reduceRequiredDimensionFiltersToFilterRules(
+                    requiredFilters,
+                    data.tables[tableName],
+                    undefined,
+                );
+            const allFilterRefs = allRequiredFilters.map(
+                (filter) => filter.target.fieldId,
+            );
+            resetRequiredFilterRules(allFilter, allFilterRefs);
```
joaoviana, line 61: I would recommend adding a few comment before  these two steps:
- reduceRequiredDimensionFiltersToFilterRules
- resetRequiredFilterRules

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
```diff
@@ -35,9 +40,66 @@ const FiltersCard: FC = memo(() => {
...
+                        reducedRules,
+                    );
+            }
+        }
+        return unsavedQueryFilters;
+    };
+    const resetDimensionFiltersIfNoModelSelected = (
+        unsavedQueryFilters: Filters,
+    ) => {
+        if (tableName.length === 0) {
```
joaoviana, line 88: can you explain this pls? 😅

packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
```diff
@@ -129,9 +133,20 @@ const FilterRuleForm: FC<Props> = ({
 
             {isEditMode &&
                 (!onConvertToGroup ? (
-                    <ActionIcon onClick={onDelete}>
-                        <MantineIcon icon={IconX} size="sm" />
-                    </ActionIcon>
+                    <Tooltip
+                        label={isRequiredLabel}
+                        disabled={!isRequired}
```
joaoviana, line 138: can you add `variant="xs"` - makes it smaller that's all. 
and perhaps a `multiline`

packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
```diff
@@ -129,9 +133,20 @@ const FilterRuleForm: FC<Props> = ({
...
+                        withinPortal
+                    >
+                        <span>
```
joaoviana, line 141: does it need a span? 🤔

packages/common/src/utils/filters.ts
```diff
@@ -779,3 +784,87 @@ export const addDashboardFiltersToMetricQuery = (
...
+): FilterRule => ({
+    id: filter.id,
+    target: {
+        fieldId: convertFieldRefToFieldId(filter.target.fieldRef, tableName),
+    },
+    operator: filter.operator,
+    values: filter.values,
+    settings: {
+        unitOfTime: filter.settings?.unitOfTime,
+    },
```
joaoviana, line 800:
```suggestion
     ...(filter.settings?.unitOfTime && {
        settings: {
            unitOfTime: filter.settings.unitOfTime,
        },
    }),
```

packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
```diff
@@ -152,9 +167,21 @@ const FilterRuleForm: FC<Props> = ({
                             <Menu.Item onClick={onConvertToGroup}>
                                 Convert to group
                             </Menu.Item>
-                            <Menu.Item color="red" onClick={onDelete}>
-                                Remove
-                            </Menu.Item>
+                            <Tooltip
```
joaoviana, line 170: same for this tooltip: 
- variant
- multiline
- does it need a span

joaoviana (changes requested):
I've finished now reviewing this, great work @mmcordoba thank you for adding this feature 🙌 

LMK what you think

packages/common/src/utils/filters.ts
```diff
@@ -779,3 +784,87 @@ export const addDashboardFiltersToMetricQuery = (
...
+    requiredFilters: MetricFilterRule[],
+    table: Table,
+    filters: FilterGroup | undefined,
+): FilterRule[] =>
+    requiredFilters.reduce<FilterRule[]>((acc, filter): FilterRule[] => {
+        const filterRule = createFilterRuleFromRequiredMetricRule(
+            filter,
+            table.name,
+        );
+        const dimension = Object.values(table.dimensions).find(
```
joaoviana, line 838: nice-to-have 

Wondering if we can remove this find by utilising the keys of `table.dimensions` and getting the field of `filter.target.fieldId` 🤔 but it might not be straightforward - so up to you

packages/common/src/types/filter.ts
```diff
@@ -46,6 +46,7 @@ export interface FilterRule<
     target: T;
     settings?: S;
     disabled?: boolean;
+    required?: boolean;
```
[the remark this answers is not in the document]
mmcordoba: done thanks

packages/common/src/compiler/translator.ts
```diff
@@ -455,6 +455,7 @@ export const convertTable = (
                 : OrderFieldsByStrategy.LABEL,
         groupLabel: meta.group_label,
         sqlWhere: meta.sql_filter || meta.sql_where,
+        required_filters: parseFilters(meta.required_filters),
```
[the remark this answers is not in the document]
mmcordoba: done thanks

packages/common/src/types/table.ts
```diff
@@ -19,6 +21,7 @@ export type TableBase = {
     orderFieldsBy?: OrderFieldsByStrategy;
     groupLabel?: string;
     sqlWhere?: string;
+    required_filters?: MetricFilterRule[];
```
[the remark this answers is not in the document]
mmcordoba: done thanks

packages/common/src/types/filterGrammar.ts
mmcordoba: done both, thanks

packages/common/src/types/filter.ts
mmcordoba: updated as suggested, thanks

packages/common/src/utils/filters.ts
mmcordoba: updated as suggested, thanks

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
mmcordoba: done thanks

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
mmcordoba: added comments as suggested, thanks

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
mmcordoba: Sure! i had to add this code, as on the ui page when you are creating a chart/query is you selected a different model, or move back and forth between models, the required filters need to be cleaned out, as different models have different filters.

packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
mmcordoba: done thanks

packages/common/src/utils/filters.ts
mmcordoba: done thanks

packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
mmcordoba: done thanks (answered about the span above)

packages/common/src/utils/filters.ts
mmcordoba: Hi, gave it a try, but was not clear to me how to make it clean without manipulating the filter.target.fieldId

mmcordoba:
> Tested the following scenarios
> 
> ## Scenario 1
> * Add chart using Tracks
> * Add filters (target `event`)
> * Save
> * Add `required_filters`
> 
> ```
> meta:
>       required_filters: 
>         - event: bash_install_failed
> ```
> 
> * Refresh dbt
> * See the filter still there and it took into account the value I had previously
> 
> ‼️ Small UI quirk from the scenario above ^ (perhaps we can tackle this on a separate PR)
> 
> * The icon of the filter doesn't change from `...` to `X` disabled button when I refresh dbt and that becomes required
> 
> Steps to reproduce
> 
> 1. Create chart with filter A
> 2. Save
> 3. Add `required_filter` that references A
> 4. Refresh dbt
> 5. See icon `...`
> 6. Click on it
> 7. Turns into `X` and it's properly disabled
> 
> **I would consider this low priority though^**
> 
> ## Scenario 2
> OR filters
> 
> * If I have `required_filter` A
> * And I have 2 filters A in chart
> * Then I can only delete one - the other is marked as non-removable/required
> 
> ## Scenario 3
> Converting filter to filter group, changing from ANY to ALL, then trying to delete it somehow
> 
> * I didn't see any issues with this - it worked well 🙌
> 
> ## Scenario 4
> Can change labels Filter is updated
> 
> tested in `tracks.yml`
> 
> ```yaml
>  meta:
>       required_filters: 
>         - event: bash_install_failed
> ....
> 
>  - name: event 
>         description: ""
>         meta:
>           dimension:
>             label: My Event
>             type: string
> ....
> ```

Hi @joaoviana , the scenario 1 quirk you mentioned, i can explain why this happens:

Required filter over a dimension is added if the user has not already defined a filter over that dimension.
Required filter main porpoise is to guarantee that a filter will always exist over that dimension, but we allow the user to modify it, meaning the user defined filter over the dimension always takes precedence over the required one.
So as long as there is a user defined filter for the dimension we are happy! and no need to override it.
The required filter behaviour is to always add a filter on that dimension if the filter does not exist. But if the filter exists, we let it be. (we do not want to alter user defined charts if they already comply by having a filter on the dimension)

Hope this makes sense.

mmcordoba:
Good morning @joaoviana  I have updated the pr with all the required changes.
Please have a look at it again once you get some time.
Thank you!

joaoviana:
Thanks for pushing these @mmcordoba! I will also duplicate this PR so E2Es run and will finalise my review at the same time!

joaoviana (approved):
Great work @mmcordoba ! Merging now that E2Es have passed https://github.com/lightdash/lightdash/issues/10416 will work on this to update the docs with a few more use-cases

Issue #10321, opened 2024-06-05
When I pin an chart/dashboard, its views increase


joaoviana:
### Description

Pinning/unpinning an item increases its views

https://github.com/lightdash/lightdash/assets/7611706/285726d7-1345-4c40-852d-c3fecf8c1581

### Steps to Reproduce the Bug or Issue

1. Go to homepage
2. Register the number of views of a chart (let's say 40)
3. Pin it to the homepage
4. Notice that the chart views has incremented by 1

This is the same for dashboards

### Cloud or self-hosting

cloud

PR #10365, opened 2024-06-11
fix: chart and dashboard view count increase on pinning to homepage

maxdiplogit:
Relates to: #10321

### Description:

While pinning either a dashboard or a particular chart to the homepage, the view count increases because:
  1. `getById()` method in `DashboardService.ts` and
  2. `get()` method in `SavedChartService.ts`
call the following block of code, that is used to increase the views count:
```
await this.analyticsModel.addDashboardViewEvent(
    dashboard.uuid,
    user.userUuid,
);
```
To avoid this view count increase on `pinning/unpinning to homepage`, having a boolean value which tells whether the chart/dashboard is being pinned to homepage or not ( `isPinning` value ) helps in this case. This value will be `true` only when pinning occurs and hence will not increase the view count, the default value will be `false` if not provided otherwise.

https://github.com/lightdash/lightdash/assets/111004544/53591d2f-977f-41b6-ab23-b72c1b137a61

PS: Right click on the video and view in a new tab if not working over here

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/DashboardService/DashboardService.ts
+++ b/packages/backend/src/services/DashboardService/DashboardService.ts
@@ -18,6 +18,7 @@ import {
     SchedulerAndTargets,
     SchedulerFormat,
     SessionUser,
+    TogglePinnedItemInfo,
     UpdateDashboard,
     UpdateMultipleDashboards,
 } from '@lightdash/common';
@@ -189,8 +190,7 @@ export class DashboardService extends BaseService {
 
     async getById(
         user: SessionUser,
-        dashboardUuid: string,
-        isPinning: boolean = false
+        dashboardUuid: string
     ): Promise<Dashboard> {
         const dashboardDao = await this.dashboardModel.getById(dashboardUuid);
 
@@ -213,12 +213,11 @@ export class DashboardService extends BaseService {
             );
         }
 
-        if (!isPinning) {
-            await this.analyticsModel.addDashboardViewEvent(
-                dashboard.uuid,
-                user.userUuid,
-            );
-        }
+        await this.analyticsModel.addDashboardViewEvent(
+            dashboard.uuid,
+            user.userUuid,
+        );
+
         this.analytics.track({
             event: 'dashboard.view',
             userId: user.userUuid,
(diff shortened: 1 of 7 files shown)


joaoviana (changes requested):
Thank you for your contribution @maxdiplogit ! 
All info and work makes a lot of sense, but was wondering if we could improve it further! 

Seems to me that we don't actually need to return the whole `SavedChart` and `Dashboard` after successful pinning-toggle. This could save us a whole db query call. 

Here are my suggestions. LMK what you think and if you're going to action them. If not, I'm happy to pick it up :) 

in `pinning.ts`

```ts
export type TogglePinnedItemInfo = Pick<PinnedListAndItems, 'pinnedListUuid'> &
    Pick<ChartSummary, 'projectUuid' | 'spaceUuid'>;

export type ApiTogglePinnedItem = {
    status: 'ok';
    results: TogglePinnedItemInfo;
};
```

extend then `ApiResults` in `packages/common/src/index.ts`
and add this
```ts
ApiTogglePinnedItem['results']
```

In `SavedChartService`, lets then use these and change the return types for `togglePinning`

```
return {
            projectUuid,
            spaceUuid,
            pinnedListUuid: pinnedList.pinnedListUuid,
        };
```

In `DashboardService`, let's do the same: 
```ts
return {
            projectUuid,
            spaceUuid,
            pinnedListUuid: pinnedList.pinnedListUuid,
        };
```

Then for both, the return type is: `Promise<TogglePinnedItemInfo>`

Now after these changes...

On the `frontend`: 

`useChartPinningMutation`:

we should then use the same types we created, so: 
```ts
lightdashApi<TogglePinnedItemInfo>({
...
...
useMutation<TogglePinnedItemInfo, ApiError, { uuid: string }>
...
```

`useDashboardPinningMutation`:

```ts
lightdashApi<TogglePinnedItemInfo>({
...
...
useMutation<TogglePinnedItemInfo, ApiError, { uuid: string }>
...
```

maxdiplogit:
@joaoviana 
Yes you are absolutely right! This is an elegant solution saving precious DB calls. I will make the changes as suggested by you in the same PR :)

joaoviana:
Works great! Just pushed a formatting issue, and then once E2Es pass let's merge 💪

maxdiplogit:
> Works great! Just pushed a formatting issue, and then once E2Es pass let's merge 💪

Okay :smile:

joaoviana (approved):
Great work @maxdiplogit 🙌

PR #10383, opened 2024-06-12
fix: get all explores in one query for dashboard available filters

ZeRego:
Closes: 

### Description:

Query all explores in 1 query (removes n+1 query)

There is a separate [tech debt ticket](https://github.com/lightdash/lightdash/issues/10384) to replace all the old model methods with the new one.

test-frontend

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2456,54 +2456,103 @@ export class ProjectService extends BaseService {
             },
             async () =>
                 wrapOtelSpan('ProjectService.getExplore', {}, async () => {
-                    const project = organizationUuid
-                        ? { organizationUuid }
-                        : await this.projectModel.getSummary(projectUuid);
-                    if (
-                        user.ability.cannot(
-                            'view',
-                            subject('Project', {
-                                organizationUuid: project.organizationUuid,
-                                projectUuid,
-                            }),
-                        )
-                    ) {
-                        throw new ForbiddenError();
-                    }
-                    const explore = await this.projectModel.getExploreFromCache(
+                    const exploresMap = await this.findExplores({
+                        user,
                         projectUuid,
-                        exploreName,
-                    );
+                        exploreNames: [exploreName],
+                        organizationUuid,
+                    });
+                    const explore = exploresMap[exploreName];
 
-                    if (isExploreError(explore)) {
+                    if (!explore) {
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2457,62 +2457,97 @@ export class ProjectService extends BaseService {
             description: 'Gets a single explore from the cache',
```
joaoviana, line 2457: will review this soon! When you get latest Sentry, could you change this `Sentry` work here? Should be similar to: 
https://github.com/lightdash/lightdash/blob/main/packages/backend/src/services/ProjectService/ProjectService.ts#L2452

```ts
 return Sentry.startSpan(
            {
                op: 'ProjectService.getExplore',
                name: 'ProjectService.getExplore',
            },
            async () =>
                wrapOtelSpan('ProjectService.getExplore', {}, async () => {
....

```

no need to `.finish()` anymore as well

ZeRego: done ✅

joaoviana:
Could you trigger e2e tests for frontend @ZeRego ? 
I'm almost done with reviewing/testing this.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2669,39 +2718,30 @@ export class ProjectService extends BaseService {
...
+                    await Promise.all([
+                        this.spaceModel.getSpacesForAccessCheck(
+                            uniqueSpaceUuids,
+                        ),
+                        this.findExplores({
+                            user,
+                            projectUuid: savedCharts[0].projectUuid, // TODO: route should be updated to be project/dashboard specific. For now we pick it from first chart as they all should be from the same project
+                            exploreNames: savedCharts.map(
+                                (chart) => chart.tableName,
+                            ),
```
joaoviana, line 2736: can we pass `organizationUuid` from the `user.organizationUuid` here?

joaoviana (approved):

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: there is no `organizationUuid` in the code

ZeRego: ah, got it 👍 thanks, adding it now

PR #10402, opened 2024-06-13
test: promote unit test

rephus:
Closes: https://github.com/lightdash/lightdash/issues/10400
### Description:

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/PromoteService/PromoteService.ts
+++ b/packages/backend/src/services/PromoteService/PromoteService.ts
@@ -27,27 +27,27 @@ import { SavedChartModel } from '../../models/SavedChartModel';
 import { SpaceModel } from '../../models/SpaceModel';
 import { BaseService } from '../BaseService';
 
-type PromotedChart = {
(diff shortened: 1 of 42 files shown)


rephus:
The `new space` issue I had in the demo is fixed here 

image

packages/backend/src/services/PromoteService/PromoteService.ts
```diff
@@ -759,7 +759,7 @@ export class PromoteService extends BaseService {
         return space.data;
     }
 
-    private async getOrCreateDashboard(
+    async getOrCreateDashboard(
```
ZeRego, line 762: If you only want to expose this method for testing purposes we can extend this class.

First, we would need to convert it from `private` to `protected` so it is available in extended classes.

```
class PromoteServiceTest extends PromoteService {
   async getOrCreateDashboard(...) {
       return super.getOrCreateDashboard(...);
   }
}
```

ZeRego (approved):

Issue #10399, opened 2024-06-13
Express requests not being reported correctly in Sentry


joaoviana:
### Description

Go to Requests tab, then see requests not being captured
Verify `express` integration
Verify if we need to provide extra headers/context to Sentry events
Check if we are initialising Sentry too early or too late

### Steps to Reproduce the Bug or Issue

Run locally
`debug: true`
Check in Sentry's `Discover` and verify that no requests are being recorded, but spans are.

### Cloud or self-hosting

cloud

PR #10403, opened 2024-06-13
fix: remove otel and initiliaze sentry as early as possible

joaoviana:
Closes: #10399

### Description:

Express requests were not being recorded at all after v8 migration. This is because in v8 Sentry started using opentelemetry under the hood and it was clashing with our current `otel` set up + package versioning was broken. 

See here an example request of this broken: https://lightdash.sentry.io/performance/trace/f72f91ee8df246319835423a01b40736/?fov=1%2C1088419.000076294&node=txn-c12b89a667364bdca091ac9a2f9fdd85&statsPeriod=1h&timestamp=1718287851

What look for in the above? ⬆️ 
- See that only manual spans are being recorded, and the API request that triggered them is not being recorded. This is a correct link for reference: https://lightdash.sentry.io/performance/trace/33da0c6696c84837aabce0d588258641/?fov=0%2C13998.300048828125&node=txn-c608806159144eeb818521fb8b5a0ea0&statsPeriod=10m&timestamp=1718292349

To fix this:
- all I had to do was to delete `otel` and remove any reference to `wrapOtelSpan` and just utilise the Sentry ones. https://docs.sentry.io/platforms/javascript/guides/express/tracing/instrumentation/opentelemetry/
- initialised `sentry` before anything else on `App` to work with express integration https://docs.sentry.io/platforms/javascript/guides/express/migration/v7-to-v8/#updated-sdk-initialization

See these links after this fix: 
https://lightdash.sentry.io/performance/trace/24a3a435252245e79a779b34b9467910/?node=txn-3f1c07425e504db7b083e51b4f41012e&node=txn-afed5fcc7df4424bb69d5e2d1a9acd5e&statsPeriod=15m&timestamp=1718291880
https://lightdash.sentry.io/performance/?environment=development&project=5959292&query=&statsPeriod=10m

> [!NOTE]
> If we end up wanting to add custom otel instrumentation to Sentry, we can then update `sentry.ts` and follow this guide: https://docs.sentry.io/platforms/javascript/guides/express/migration/v7-to-v8/v8-opentelemetry/#custom-opentelemetry-instrumentation

### Reviewer actions


packages/backend/src/App.ts
```diff
@@ -230,6 +227,14 @@ export default class App {
             Logger.error('Error starting slack bot', e);
         });
 
+        Sentry.setTags({
```
joaoviana, line 230: verify tags

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -1015,7 +978,7 @@ export class UnfurlService extends BaseService {
 
                     Sentry.captureException(e);
                     hasError = true;
-                    span.recordException(e);
+                    span.addEvent(e);
```
owlas, line 981: Do you have docs for this change? I couldn't find a reference to `addEvent` for recording errors

owlas (approved):

packages/backend/src/services/UnfurlService/UnfurlService.ts
joaoviana: https://github.com/getsentry/sentry-javascript/blob/1201eb2d78425709293ee0aac180cbcc1f4e99d2/packages/types/src/span.ts#L247 saw this @owlas Perhaps we can even remove this because we already capture the exception on the line above, but I would like to attach this event to the `span`

PR #10424, opened 2024-06-14
fix: ensure all trino query result chunks are fetched and processed

Shifan-Gu:
Closes: 

Add a while loop in trino warehouse client to make sure all trino query result chunks are fetched and processed.

for reference, source code of trino-js-client `.next` method implementation:

https://github.com/regadas/trino-js-client/blob/main/src/index.ts#L378

`.next()` method terminates when first chunk of data is available and would make `done` property as false

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/TrinoWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/TrinoWarehouseClient.ts
@@ -187,7 +187,7 @@ export class TrinoWarehouseClient extends WarehouseBaseClient<CreateTrinoCredent
             }
             query = await session.query(alteredQuery);
 
-            const queryResult = await query.next();
+            let queryResult = await query.next();
 
             if (queryResult.value.error) {
                 throw new WarehouseQueryError(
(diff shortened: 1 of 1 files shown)


packages/warehouses/src/warehouseClients/TrinoWarehouseClient.ts
```diff
@@ -196,14 +196,21 @@ export class TrinoWarehouseClient extends WarehouseBaseClient<CreateTrinoCredent
...
-            const result: QueryData = queryResult.value.data ?? [];
-
+            let result: QueryData = [];
             const schema: {
                 name: string;
                 type: string;
                 typeSignature: { rawType: string };
             }[] = queryResult.value.columns ?? [];
 
+            // Using `await` in this loop ensures data chunks are fetched and processed sequentially.
```
ZeRego, line 206: Nice investigation! 🕵️ 
How many rows was necessary to have multiple chunks?

Shifan-Gu: I think it depends on how it's configured on trino side, this client calls the trino rest api to fetch the data.
https://trino.io/docs/current/develop/client-protocol.html
this endpoint returns the next uri address if needed.

ZeRego: Thanks for the info. 🙌 
Can you check the failing tests? I suspect this is because the first chunk has `done: true` so it never enters the `while` loop and concats the results. 
We can also add a new unit test that simulates multiple chunks. You can mock multiple calls like this
```
jest.fn()
     .mockReturnValueOnce({ done: false, value: queryResponse })
     .mockReturnValueOnce({ done: true, value: queryResponse });
```

Shifan-Gu: sure, will do 👍

ZeRego (approved):
lgtm 🥇

PR #10426, opened 2024-06-14
fix: remove n+1 queries when saving chart

ZeRego:
Remove n+1 queries when saving chart.

test-frontend

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -2,7 +2,11 @@
 // eslint-disable-next-line import/order
 import './sentry'; // Sentry has to be initialized before anything else
 
-import { LightdashMode, SessionUser } from '@lightdash/common';
+import {
+    LightdashMode,
+    SessionUser,
+    UnexpectedServerError,
+} from '@lightdash/common';
 import * as Sentry from '@sentry/node';
 import flash from 'connect-flash';
 import connectSessionKnex from 'connect-session-knex';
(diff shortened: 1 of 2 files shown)


packages/backend/src/App.ts
```diff
@@ -453,6 +457,9 @@ export default class App {
         expressApp.use(
             (error: Error, req: Request, res: Response, _: NextFunction) => {
                 const errorResponse = errorHandler(error);
+                if (error instanceof UnexpectedServerError) {
+                    Logger.error(error); // Log original error for debug purposes
```
rephus, line 461: Is this intended ? 
Recently, for debugging I've been using `console.error` instead of  `Logger.error` because with console we show the full stacktrace. Or add `console.trace(e)` on top.

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -180,44 +157,47 @@ const createSavedChartVersion = async (
                 timezone,
             })
             .returning('*');
-        const promises: Promise<any>[] = [];
-        dimensions.forEach((dimension) => {
-            promises.push(
-                createSavedChartVersionField(trx, {
+        if (dimensions.length > 0) {
```
rephus, line 160: nice, you could also check `if (dimensions.length > 0)` inside `createSavedChartVersionFields`

rephus (approved):
looks good

packages/backend/src/models/SavedChartModel.ts
ZeRego: yup, thats cleaner and reusable. Moved those checks inside the functions and triggered frontend tests to be sure nothing breaks.

Issue #10358, opened 2024-06-10
Custom metric `max` from `date` dimension should not be a type of `date`


IrakliJani:
### Description

Custom metric `max` from the `date` dimension should not be a type of `date`. Instead, it should be of type `max`.

⚠️ Keep in mind that the new solution should also include the fix linked below without breaking the above-described behavior

regression pr: https://github.com/lightdash/lightdash/pull/6940

### Steps to Reproduce the Bug or Issue

1. go to one of the explores
2. find date dimension and create `max`
3. add filter
4. run query
5. 💥 

### Cloud or self-hosting

None

rephus:
I'm not sure exactly what's wrong, the max of a date returns a date, not a number . Date filters seem to work (except `in the last` ) 

![image](https://github.com/lightdash/lightdash/assets/1983672/232f5f30-376b-4c6a-a639-8366d9472634)

IrakliJani:
@rephus try adding the filter in the custom metric itself

IrakliJani:
@rephus here's more context on this:

`packages/frontend/src/components/Explorer/CustomMetricModal/utils/index.ts`: in method `getTypeOverridesForAdditionalMetric`.

we should not be converting metric type from min,max to date because it breaks custom metric filters
but when we remove the conversion it breaks the formatting.

ideally - it should keep the min/max format of the metric but still use the value formatting of date.

IrakliJani:
relevant thread: https://lightdash.slack.com/archives/C03MCQ08UKS/p1717671947299319

PR #10442, opened 2024-06-17
fix: custom metric date max format

rephus:
Closes: https://github.com/lightdash/lightdash/issues/10358

### Description:

> ideally - it should keep the min/max format of the metric but still use the value formatting of date.

Keep `max` type on the custom metric, and add a new `date/timestamp` custom format .

- [x]  Keep in mind that the new solution should also include the fix linked below without breaking the above-described behavior

Screenshot from 2024-06-17 16-18-35

### New side effects

There are a couple of unexpected side effects when keeping this a `max` . We could create separate issues for these. 

- [ ] Max Date should not show totals 
Screenshot from 2024-06-17 16-21-29
- [ ] Max date tooltips should be formatted correctly 
image

### Reviewer actions


IrakliJani (approved):
this is great 👍👍

Issue #9024, opened 2024-02-21
Introduce alternative library to generate scheduled delivery images


ZeRego:
### Description

We currently use puppeteer to render the website and take screenshots for scheduled deliveries. 
It has proven to be unreliable as users from time to time, report images where some elements don't show or show incorrectly. 
We haven't been able to replicate most of these issues and logs haven't provided us a reason as to why this happens. 

We are now at a point where we exhausted our theories and fixes. 

We should start experimenting with other libraries, under a feature flag, and find a replacement. 

Here are some possible alternatives:

Library | Languages | Best For | Popularity | Browser Support | Ease of Use | Speed
-- | -- | -- | -- | -- | -- | --
Playwright | Python, JavaScript, TypeScript, .NET, Java | Cross-browser automation using a single API | Rapidly growing | Chrome, Firefox, Safari | Moderate | Moderate
Selenium | NodeJS, Python, Java, Ruby, Perl, R, Haskell, and Objective-C | Automated web testing and web scraping | Large user base | Chrome, Edge, Firefox, Safari | Can get complex as it requires additional setup | Resource-intensive and slow when running multiple instances
Cypress | NodeJS | Testing anything that runs in the browser | Rapidly growing | Chrome, Edge, Firefox, Safari | Beginner-friendly | Moderate
WebdriverIO | NodeJS | Web and mobile application testing | Growing user base | Chrome, Edge, Firefox, Safari | Moderate | Moderate

Issue #10457, opened 2024-06-19
Remove puppeteer completely 


joaoviana:
### Description

See `headlessBrowser.ts` and assess if we can replace it with `playwright` and then remove `puppeteer` from the package

### What problem does this solve?

Only one technology to deal with browser automation

ZeRego:
Already done

PR #10313, opened 2024-06-04
feat: increase max retries for dashboard image schedulers

joaoviana:
Closes: N/A

### Description:

Increases `maxAttempts` for dashboard + image screenshot + schedulers by `1` so it's `2` now. 

Also, upgrades `puppeteer` to latest, because it includes these fixes: 

https://github.com/puppeteer/puppeteer/pull/12352
https://github.com/puppeteer/puppeteer/pull/12225

This is based on the errors we would get from logs: 
ProtocolError
`Requesting main frame too early!`

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #10028, opened 2024-05-09
fix: retry screenshot for protocol and timeout errors

joaoviana:
Closes: #9947 

### Description:

Retries screenshot logic for a scheduled job when there's a `ProtocolError` or a `Timeout` error.
This will retry `saveScreenshot` recursively 3 times if the error is _retryable_

References: 
https://github.com/puppeteer/puppeteer/issues/10144#issuecomment-1918840262
And from `Test script` section in Browserless: https://www.browserless.io/blog/horizontally-scaling-chrome

To reproduce this retry/error: 
- Create a scheduled delivery for a dashboard
- Rename the `finalSelector` for dashboards from 
```ts
 finalSelector = '.react-grid-layout';
```
to 
```ts
 finalSelector = '.react-gri-layout';
```
- That will lead to a timeout error and you will see the retries happening.

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #10458, opened 2024-06-19
chore: use playwright to saves creenshots by default

joaoviana:
Closes: #9024

### Description:

We experimented using `playwright` to save screenshots and that has been quite reliable at getting them for scheduled deliveries and working with `browserless`. 
Besides these changes, we also:
- Added retries #10028 
- Increase `maxAttempts` of dashboard + image scheduled deliveries to 2 (#10313)
- Increased the replica count of some instances (see cloud)
- Reduced max number of concurrent jobs slightly for instances that heavily use scheduled deliveries (see cloud)
 
This PR removes: 
- usage of env var: `SCHEDULER_SCREENSHOT_WITH_PLAYWRIGHT`
- removes the puppeteer `saveScreenshot` util and replaces it with its playwright-equivalent

> [!NOTE]
> This doesn't remove `puppeteer` completely because I see some references in `headlessBrowser` and that should be handled in another PR to remove Puppeteer completely. Issue here: https://github.com/lightdash/lightdash/issues/10457

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -16,12 +16,6 @@ import { nanoid as useNanoid } from 'nanoid';
 import fetch from 'node-fetch';
 import { PDFDocument } from 'pdf-lib';
 import playwright from 'playwright';
-import puppeteer, {
-    ProtocolError,
-    TimeoutError,
-    type Browser,
-    type Page,
-} from 'puppeteer';
 import { S3Client } from '../../clients/Aws/s3';
 import { LightdashConfig } from '../../config/parseConfig';
 import { DashboardModel } from '../../models/DashboardModel/DashboardModel';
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -701,321 +673,6 @@ export class UnfurlService extends BaseService {
         );
     }
 
-    private async saveScreenshot({
```
rephus, line 704: I'd pin this commit somewhere, in case we need to compare playwright config against some old logic in puppeteer

rephus (approved):
Good refactoring, I don't need to test this

packages/backend/src/services/UnfurlService/UnfurlService.ts
rephus: https://github.com/lightdash/lightdash/releases/tag/before-playwright-puppeteer-replacement

PR #10463, opened 2024-06-19
fix: get correct response pattern for explore screenshots

joaoviana:
Closes: #10462 

### Description:

See Sentry issue in ticket ⬆️ 

- Waits for `/runQuery` instead of `/results` when sharing an unsaved chart link so it works with Slack image unfurling.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -492,14 +492,20 @@ export class UnfurlService extends BaseService {
                                     timeout: 60000,
                                 }); // NOTE: No await here
                             });
-                        } else if (
-                            lightdashPage === LightdashPage.CHART ||
-                            lightdashPage === LightdashPage.EXPLORE
-                        ) {
-                            // Wait for the visualization to load if we are in an explore page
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -492,15 +492,21 @@ export class UnfurlService extends BaseService {
...
                             const responsePattern = new RegExp(
                                 `${resourceUuid}/results`,
                             );
 
+                            chartResultsPromises = [
+                                page?.waitForResponse(responsePattern, {
+                                    timeout: 60000,
+                                }), // NOTE: No await here
+                            ];
+                        } else if (lightdashPage === LightdashPage.EXPLORE) {
```
joaoviana, line 506: more elegant ways to handle this are welcome! 

We need to define the `waitForResponse`s like this and then await them after - playwright way.

magnew: Seems pretty reasonable. Most alternatives I can think of would be much more intrusive.

magnew (approved):

PR #10467, opened 2024-06-20
fix: add queue size prometheus

rephus:
Closes:  https://github.com/lightdash/lightdash-cloud/issues/403

### Description:

We could add a group by on task and add the task as an extra argument to the gauge, but this will probably slow down the request, so not doing it for now. 

Screenshot from 2024-06-20 12-06-48

### Reviewer actions

### DIFF
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -226,6 +226,7 @@ export type LightdashConfig = {
         gcDurationBuckets?: number[];
         eventLoopMonitoringPrecision?: number;
         labels?: Object;
+        queueMonitoringFrequency?: number;
     };
     database: {
         connectionUri: string | undefined;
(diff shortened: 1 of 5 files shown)


packages/backend/src/prometheus.ts
```diff
@@ -37,6 +38,22 @@ export default class PrometheusMetrics {
...
+            const queueSizeGauge = new prometheus.Gauge({
+                name: 'queue_size',
+                help: 'Number of jobs in the queue',
+                ...rest,
+                collect() {},
+            });
+            setInterval(async () => {
+                const queueSize = await schedulerClient.getQueueSize();
+                queueSizeGauge.set(queueSize);
+            }, queueMonitoringFrequency || 30 * 1000); // 30 seconds
```
ZeRego: If we want to get the value at the time /metrics is requested we just need to have 

```
collect() {
   return schedulerClient.getQueueSize();
},
```

If we want to have a separate timer for this, we can remove the `collect() {},`

packages/backend/src/config/parseConfig.ts
```diff
@@ -641,6 +642,9 @@ const mergeWithEnvironment = (config: LightdashConfigIn): LightdashConfig => {
             labels: getObjectFromEnvironmentVariable(
                 'LIGHTDASH_PROMETHEUS_LABELS',
             ),
+            queueMonitoringFrequency: getIntegerFromEnvironmentVariable(
+                'LIGHTDASH_QUEUE_MONITORING_FREQUENCY',
```
ZeRego: Should we name it `LIGHTDASH_PROMETHEUS_QUEUE_MONITORING_FREQUENCY` and add it to the environment_variables.mdx ?

ZeRego (approved):

Issue #10441, opened 2024-06-17
Filter for week dimension is created with invalid value


ZeRego:
### Description

When created a filter for a week filter the default value is incorrect. 
User needs to change the filter value to filter correctly.

### Steps to Reproduce the Bug or Issue

(set project start of the week to be a different day. eg: tuesday)

1. go to week dimension
2. create filter for dimension
3. check default value

Expected:
- show tuesday from the current week

Current:
- show Monday/Sunday from the current week

### Cloud or self-hosting

None

PR #10471, opened 2024-06-20
fix: filter for week dimension is created with invalid value

magnew:
Closes: #10441 

### Description:

The bug here is that week filters were initially created with the wrong start of week when the start of was customised. This fix makes it so that week filters are initially created with an empty value, which is how most of the other fitler types work. The user has to fill in the value before the fitler is valid (again, like the other filters) and we use the right start of week.

Note: I also tried populating the value correctly initially, which is possible but a lot messier. The default filter values come from a utility, so creating the right default date required a lot of passing startOfWeek. 

### Reviewer actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [0.1135.0](https://github.com/lightdash/lightdash/compare/0.1134.0...0.1135.0) (2024-06-19)
+
+
+### Features
+
+* prometheus metrics ([#10459](https://github.com/lightdash/lightdash/issues/10459)) ([a92383b](https://github.com/lightdash/lightdash/commit/a92383bf6ffc90a47a44fec940f2a7d899d825f5))
+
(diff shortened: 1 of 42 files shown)


ZeRego:
@magnew problem still exists when updating a filter. eg: changing from day to week

bug changing filter

I think we need to change the logic in `getFilterRuleWithDefaultValue` to set the value to `null` when the dimension is week. Better to set null than an invalid value.
https://github.com/lightdash/lightdash/blob/0fe18e55ad327bd16df298e7cc00cb598669637c/packages/common/src/utils/filters.ts#L207

We can fix this in the packages/frontend/src/components/common/Filters/FilterInputs/FilterWeekPicker.tsx
by setting the value to the start of the week but doesn't feel intuitive. 
```
value={startOfWeek(date, firstDayOfWeek)}
```

Let me know what you think

12ian34:
Closing - will reopen with a different approach.

Issue #10328, opened 2024-06-06
The fields in the sidebar need to be displayed as labels / friendly-names


PriPatel:
Related to #10311 
The fields in the sidebar need to be displayed as labels / friendly-names

<img width="617" alt="Screenshot 2024-06-06 at 10 50 21" src="https://github.com/lightdash/lightdash/assets/12776522/ebf90264-c9e6-4def-84f0-490ff8d86637">
Right now, they are not very readable.

PR #10485, opened 2024-06-21
feat: display field labels in catalog metadata

joaoviana:
Closes: [#10328](https://github.com/lightdash/lightdash/issues/10328)

### Description:

- Display field labels on the panel **and** on the metadata sidebar table
- Created a separate component for the fields table as a lil refactor
- Added `tableLabel` to the CatalogTree so when we're searching we don't get _empty_ table labels 

Screenshot 2024-06-21 at 17 11 48

### Reviewer actions


magnew:
I have one follow up comment though: is there no nowhere to see the field id that was there before? It's often ugly but not showing it at all might be annoying for debugging things (for a user)

PriPatel:
@magnew 

Great point - I think it will be good to have in the metadata sidebar! (for later!) Or even on hover of the label

PR #10487, opened 2024-06-24
fix: queue monitoring sql

rephus:
Closes: https://github.com/lightdash/lightdash-cloud/issues/433
### How I tested this

set LIGHTDASH_SCHEDULER: False

Tried downloading a CSV, the gauge increased
Scheduled some deliveries in the future, the gauge didn't increase
### Description:

Screenshot from 2024-06-24 09-12-47

I also tried doing this: 

> add metric queue_concurrency_utilization where it filters rows by is_locked (as it is an indicator that is being processed) and divide by concurrency config to find the %.

However , when the jobs are `in progress` they are removed from the `jobs` table. so we can't track thoes jobs using SQL. We could do some manual tracking using `runner.events.on`

### Reviewer actions

### DIFF
--- a/packages/backend/src/scheduler/SchedulerClient.ts
+++ b/packages/backend/src/scheduler/SchedulerClient.ts
@@ -215,7 +215,7 @@ export class SchedulerClient {
         const graphileClient = await this.graphileUtils;
         const results = await graphileClient.withPgClient((pgClient) =>
             pgClient.query(
-                'select count(id) as count from graphile_worker.jobs where attempts <> max_attempts',
+                'select count(id) as count from graphile_worker.jobs where attempts <> max_attempts AND run_at < now()',
             ),
         );
         return parseInt(results.rows[0].count, 10);


ZeRego (approved):

ZeRego:
> However , when the jobs are in progress they are removed from the jobs table. so we can't track thoes jobs using SQL. We could do some manual tracking using runner.events.on

I don't think this is true as they need to track what worker is running a job and how many attempts they tried for that job.
When a job is in progress the `locked` columns are populated.
Example of a query to check how many jobs are in progress/waiting/etc.
https://github.com/graphile/worker/issues/380#issuecomment-1806155101

Also, we should filter by `task_identifier = 'handleScheduledDelivery'`. Otherwise, the "child" jobs we create might ruin this metric. 🤔

rephus:
> I don't think this is true as they need to track what worker is running a job and how many attempts they tried for that job.
> When a job is in progress the locked columns are populated.

That is also what I thought it was happening, however, when I was testing locally, I couldn't get any `running` job to be in the table (I tried refresh dbt and `export dashboard`) 

I even implemented the new SQL and metric, and then when I was testing I noticed the metric was always 0

rephus:
> Have you tried forcing the job to take 1 minute?

Nope, but let's do it on a separate PR anyway. I can create a ticket for that

Issue #10489, opened 2024-06-24
Do not show errored tables


joaoviana:
(get rid of the tables with error section, as users really dont' want to see this! The catalog should accurately reflect the tables sidebar)

PR #10491, opened 2024-06-24
feat: hide explore errors in catalog

joaoviana:
Closes: #10489

### Description:

Hide explore errors in Catalog if they have `NO_DIMENSIONS_FOUND` error - we do the same in the validator.
Reduce opacity of errors table list item (this will be revisited a bit more when we deal with the removal of this grouping)
Display tooltip if the error is truncated, but don't open the sidebar with the full error anymore

**note**: I see that the keyboard navigation could be handled differently. I believe this can be done on a separate PR, but lmk if pressing

To test, I changed this locally in `schema.yml`

image

before
Screenshot 2024-06-24 at 11 02 29

after

https://github.com/lightdash/lightdash/assets/7611706/2393cd12-1f54-4b62-9246-6daaea421eae

And now we don't see the errors group and we don't handle if there are errors in the explore because they don't get added to the payload

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CatalogService/CatalogService.ts
+++ b/packages/backend/src/services/CatalogService/CatalogService.ts
@@ -11,6 +11,7 @@ import {
     ExploreError,
     ForbiddenError,
     hasIntersection,
+    InlineErrorType,
     isExploreError,
     SessionUser,
     SummaryExplore,
@@ -127,6 +128,15 @@ export class CatalogService<
             await this.projectModel.getTablesConfiguration(projectUuid);
         return explores.reduce<CatalogTable[]>((acc, explore) => {
             if (isExploreError(explore)) {
+                // If no dimensions found, we don't show the explore error
+                if (
+                    explore.errors.every(
+                        (error) =>
+                            error.type === InlineErrorType.NO_DIMENSIONS_FOUND,
+                    )
+                )
+                    return acc;
+
                 return [
                     ...acc,
                     {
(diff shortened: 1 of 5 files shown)


magnew (approved):
Looks good. This is kinda a product question, but doesnt it seem useful to be able to find (and so potentially fix or remove) explores with errors? Maybe we should have an 'errors' section at some point. 

Obviously thats a future improvement.

joaoviana:
> Looks good. This is kinda a product question, but doesnt it seem useful to be able to find (and so potentially fix or remove) explores with errors? Maybe we should have an 'errors' section at some point.
> 
> Obviously thats a future improvement.

Thanks for pointing this out @magnew, I just spoke to @PriPatel and we're handling the errors differently. Putting this back into draft.

Main changes will be: 
- remove sidebar error display
- display whole error with a tooltip
- clarify what is an explore error that we want to show here. For example, on the Explorer, we hide `no dimensions found` error tables. Currently, on `main`, we display all errors in the catalog - we should filter and those out but keep the meaningful errors.

PriPatel:
Yeah thanks and sorry @magnew and @joaoviana <3   I was confused by the difference in what was being shown in our analytics sidebar vs analytics catalog. They should be the same , but I didnt realise that we were hiding certain errors from the sidebar!

joaoviana:
Thanks @magnew & @PriPatel for your comments, I've updated the description and the code!

Issue #10488, opened 2024-06-24
Funnel chart: add value and percentage display options


magnew:
### Description

A few of the early adopters asked to be able to show values and percentages without hover

PR #10497, opened 2024-06-24
feat: add label placement, value and percentage to funnel chart

magnew:
Closes: #10488 

### Description:

Add label properties to funnel chart:
- placement
- show value (without hover)
- show percentage (without hover)

Chart with label: right, value and percentage:
Screenshot 2024-06-24 at 19 06 12

Config:
Screenshot 2024-06-24 at 19 06 48

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsFunnelConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsFunnelConfig.ts
@@ -1,7 +1,10 @@
 import {
     formatItemValue,
+    FunnelChartLabelPosition,
+    type Metric,
     type ResultRow,
     type ResultValue,
+    type TableCalculation,
 } from '@lightdash/common';
 import { type EChartsOption, type FunnelSeriesOption } from 'echarts';
 import { round } from 'lodash';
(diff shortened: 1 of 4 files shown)


hamzahc1:
Shouldn't the label text always be black/grey? it's weird that it's yellow, blue, red etc - it means it's hard to read depending on the colourscheme you've picked...

packages/frontend/src/hooks/echarts/useEchartsFunnelConfig.ts
```diff
@@ -54,22 +73,39 @@ const useEchartsFunnelConfig = (isInDashboard: boolean) => {
...
+                            maxValue: chartConfig.maxValue,
+                        });
+
+                    return `${marker}<b>${name}</b><br /> Value: ${formattedValue} <br/> Percent: ${percentOfMax}%`;
                 },
             },
             label: {
                 show: true,
-                position: 'inside',
+                position: label?.position || FunnelChartLabelPosition.INSIDE,
```
joaoviana, line 88: we could add here `color: 'black'` if the position is different from `inside`?

joaoviana (changes requested):
I think we can address @hamzahc1's point - left a suggestion on how to add it. 

Great stuff, tested in Render 👍

packages/frontend/src/hooks/echarts/useEchartsFunnelConfig.ts
magnew: Done

joaoviana (approved):
Checked in render and label is `black` when it's not inside. 
If inside, echarts does a contrast check and applies the correct color

Issue #10150, opened 2024-05-22
Enable linking to a field in the explores page from the catalog


magnew:
### Description

Right now when you click on a table, it opens up a table. Similarly, when I click on a field, it should open up an explore with that field selected.

- [ ] If you click on Field X -> You open an explore view with the relevant table selected and Field X must be selected by default too. The query does not have to have been run.

PR #10509, opened 2024-06-25
feat: allow select field to explore in catalog

joaoviana:
Closes: [#10150](https://github.com/lightdash/lightdash/issues/10150)

### Description:

- Allows field selection when searching it (on hover) and also when viewing its metadata

Demo:

https://github.com/lightdash/lightdash/assets/7611706/4a653f98-8434-4784-8b55-d0b5c1203265

**NOTE**: this also filters out dimensions that have `isIntervalBase`: `true` - for example, we have a `Timestamp` dimension that has many granularities. Currently, we display its base dimension, but that isn't explorable. Instead, we should just filter them out from the catalog fields that we can search. If we want to bring back this functionality in the future, we should remove this filtering but handle these fields differently on the Catalog - we should have a tree basically.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
+++ b/packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
@@ -33,6 +33,7 @@ import MarkdownPreview from '@uiw/react-markdown-preview';
 import { useEffect, useMemo, useState, type FC } from 'react';
 import { useHistory } from 'react-router-dom';
 import MantineIcon from '../../../components/common/MantineIcon';
+import { getExplorerUrlFromCreateSavedChartVersion } from '../../../hooks/useExplorerRoute';
 import { useIsTruncated } from '../../../hooks/useIsTruncated';
 import { useCatalogContext } from '../context/CatalogProvider';
 import { useCatalogAnalytics } from '../hooks/useCatalogAnalytics';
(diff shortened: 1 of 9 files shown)


packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
```diff
@@ -428,12 +436,60 @@ export const CatalogMetadata: FC = () => {
                             },
                         })}
                         onClick={() => {
-                            history.push(
+                            if (metadata && isViewingField) {
+                                return history.push(
+                                    getExplorerUrlFromCreateSavedChartVersion(
```
magnew, line 441: This whole block is basically just a hardcoded default chart right? I feel like this should be a constant somewhere instead of inline in this handler. Maybe a util like getDefaultExplore? OR getExplorerUrlFromCreateSavedChartVersion could take null as a second argument and if so, it uses this default chart?

This is kinda hygiene so it doesn't have to be a blocker, but someday someone is gonna be confused about where this explore is getting initialized 😄

magnew (approved):
Looks good! I have a code suggestion that doesn't have to be blocking. And 2 UX suggestions:
- Can we put the field name (non-friendly name) in the metadata? I feel like it can be useful to map it to the actual data
- Even though I knew I was looking for it, I didn't see the "Select field" button, then I didn't expect it to be a redirect. IMO it's kinda hiding too far at the bottom of the page. And to me it would be a lot clearer if it said something like "Select field in explore". 

Those last two might be questions for @PriPatel. Just suggestions though, take it or leave it 😄

joaoviana:
> * Even though I knew I was looking for it, I didn't see the "Select field" button, then I didn't expect it to be a redirect. IMO it's kinda hiding too far at the bottom of the page. And to me it would be a lot clearer if it said something like "Select field in explore".

@PriPatel - something to consider on next steps!

packages/frontend/src/features/catalog/components/CatalogMetadata.tsx
joaoviana: 100% @magnew ! 

and here it is: https://github.com/lightdash/lightdash/pull/10509/commits/869b2bd7453946bc6b69e4308eca7d87dd28e33c

joaoviana:
> Can we put the field name (non-friendly name) in the metadata? I feel like it can be useful to map it to the actual data 

This appears on hover of the field in the metadata
image

Issue #10507, opened 2024-06-25
add `dashboard_id` to `query_executed` event


TuringLovesDeathMetal:
### Description

We should add `dashboard_id` to the `query_executed` event when a query has the context `dashboardView`

PR #10511, opened 2024-06-25
feat: analytics events dashboard id for dashboard query exec context

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/10507

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1719,7 +1719,7 @@ export class ProjectService extends BaseService {
                         );
                     }
 
-                    await this.analytics.track({
+                    this.analytics.track({
                         userId: user.userUuid,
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1719,7 +1719,7 @@ export class ProjectService extends BaseService {
                         );
                     }
 
-                    await this.analytics.track({
```
magnew, line 1722: Hmm, should these ever be awaited? I see a few other places we are doing this.

magnew (approved):
👍

packages/backend/src/services/ProjectService/ProjectService.ts
IrakliJani: no, its not a promise function and is redundant

Issue #10359, opened 2024-06-10
Adding flags to dbt project throws an error


jpetey75:
### Description

When `flags` config is added to `dbt_project.yml` an error is thrown:

>Failed to run "dbt1.8 deps" with dbt version "v1.8"

[Project flags docs](https://docs.getdbt.com/reference/global-configs/project-flags)

### Steps to Reproduce the Bug or Issue

1. add `flags` to `dbt_project.yml`
2. sync dbt project to Lightdash
3. error

### Cloud or self-hosting

None

jpetey75:
Internal threads for two users who his this error recently:

1. https://lightdash.slack.com/archives/C071TU6FZ3P/p1718044291245749
2. https://lightdash.slack.com/archives/C03MCQ08UKS/p1717497757923619

jpetey75:
came up again today: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1718892171890499

PR #10519, opened 2024-06-25
fix: support dbt_project flags

ZeRego:
Closes: [#10359](https://github.com/lightdash/lightdash/issues/10359)

### Description:

Problem:
> Runtime Error Do not specify both 'config' in profiles.yml and 'flags' in dbt_project.yml. Using 'config' in profiles.yml is deprecated."

Lightdash generates temporary profile files which had `config`. 
I have moved that config from the profile file to the dbt command env vars.

Note: `flags` was only supported from dbt >= 1.6
Note 2:
1.4 / 1.5 - should work without `flags`. Adding it would throw an error.
1.6 / 1.7 - should work with `flags`
1.8 - should work with flags but we don't support postgres atm so you need to test with BigQuery/Snowflake/Trino

### Reviewer actions

### DIFF
--- a/packages/backend/src/dbt/dbtCliClient.ts
+++ b/packages/backend/src/dbt/dbtCliClient.ts
@@ -2,14 +2,10 @@ import {
     assertUnreachable,
     DbtError,
     DbtLog,
-    DbtManifestVersion,
     DbtPackages,
-    DbtRpcDocsGenerateResults,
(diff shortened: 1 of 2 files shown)


rephus:
are we expecting then to see some  errors on the CLI e2e tests ?

rephus (approved):

PR #10523, opened 2024-06-25
fix: don't cache max value when fields change in funnel chart

magnew:
### Description:

Funnel chart was caching the max value when fields were changed. 

Repro:
1. Make a funnel chart with 2 numeric fields. 
2. Use steps are ROWS
3. Switch between the two fields
4. Notice the percentages are wrong in the second field

This just calculated the max along with the data. I don't remember why I had a useRef. 😅

### Reviewer actions


joaoviana (approved):
Yup this fixes it! tried here: https://lightdash-pr-10523.onrender.com/share/Cyvcw5dR4-mBJipreez65

Issue #10326, opened 2024-06-06
You should be able to pick what item type you are searching for before entering a search term 


PriPatel:
<img width="864" alt="Screenshot 2024-06-06 at 09 44 59" src="https://github.com/lightdash/lightdash/assets/12776522/5c9965cc-bb6d-4208-9108-99a72a73d83c">

**This will replace the current filters placeholder and makes for a much better and less confusing experience:** 
- [ ] By default you are searching through 'All item types'
- [ ] You can click on a dropdown to select which item type you would like to search by: For now (all , tables , metrics , dims) 
- [ ] If I've selected anything other than `all`, my search results should only be based on that item type
- [ ] This should not affect the search logic, just the result type that is being shown in the results (ie. I can still search for description matches)
- [ ] For styling use standard Mantine dropdown
<img width="218" alt="Screenshot 2024-06-06 at 09 59 10" src="https://github.com/lightdash/lightdash/assets/12776522/7a462b2c-08ac-45d2-a7d0-fddd39c1aa9c">

PR #10525, opened 2024-06-25
feat: enable searching catalog with filters (including search index)

joaoviana:
Closes: [#10326](https://github.com/lightdash/lightdash/issues/10326)

### Description:

- Enables searching just for `tables` or `metrics` or `dimensions`
- These are only enabled when search value is > 2

Overall changes
- Added migration file that includes `field_type` column to `catalog_search` - this is useful for filtering out fields by type: `dimension` or `metric`. These are added when the user compiles the project
- Search accepts `type` and `filter` which will take that into account when searching 
  - if the filter is `table` then it will search only by the table type
  - if the filter is `dimension` or `metric`, it will only search by fields **and** by `field_type`
- Removed previous filter logic on the FE because we changed a few things in its behaviour
  - we only accept one filter at a time
  - we process the search on the BE 

Demo:

https://github.com/lightdash/lightdash/assets/7611706/2c4cf04b-ab7e-4d46-aef2-06112c58925a

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/catalogController.ts
+++ b/packages/backend/src/controllers/catalogController.ts
@@ -42,11 +42,13 @@ export class CatalogController extends BaseController {
         @Request() req: express.Request,
         @Query() search?: ApiCatalogSearch['search'],
         @Query() type?: ApiCatalogSearch['type'],
+        @Query() filter?: ApiCatalogSearch['filter'],
     ): Promise<{ status: 'ok'; results: ApiCatalogResults }> {
         this.setStatus(200);
         const query: ApiCatalogSearch = {
             search,
             type,
+            filter,
         };
 
         const results = await this.services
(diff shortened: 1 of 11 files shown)


rephus:
Why is it showing up orders ? is it on the description ? 

image

rephus:
image
I don't think dimension filtering is workign as expected...

joaoviana:
For both comments raised @rephus its a description match

First one:

> Why is it showing up orders ? is it on the description ?

yes it is! We unfortunately can't highlight a line-clamp - not CSS supported... We need to wait for a bit more feedback from @PriPatel 's user feedback sessions to define the next step here

> I don't think dimension filtering is workign as expected...

It's a description match - each reference customer - I've pushed latest to this branch so Render will show the `field` description and the word will be highlighted 👍

rephus (approved):

Issue #10536, opened 2024-06-26
Group membership endpoint only returns the first 100 members


magnew:
### Description

Group membership endpoint only returns the first 100 members, it should return all of them

### Steps to Reproduce the Bug or Issue

1. Add 110 users to a group 💯 
2. Check the group membership, there are only 100

### Cloud or self-hosting

None

PR #10541, opened 2024-06-27
fix: increase limit on number of members shown per group

magnew:
Closes: #10536 

### Description:

Increase the number of members returned when viewing group membership. This is a partial/temporary fix. We should also make this a paginated list on the FE.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
+++ b/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
@@ -168,7 +168,7 @@ const GroupsView: FC = () => {
     const [search, setSearch] = useState('');
 
     const { data: groups, isInitialLoading: isLoadingGroups } =
-        useOrganizationGroups({ search, includeMembers: 100 }); // TODO: pagination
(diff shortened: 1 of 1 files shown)


joaoviana (approved):

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/GroupsView.tsx
```diff
@@ -168,7 +168,7 @@ const GroupsView: FC = () => {
     const [search, setSearch] = useState('');
 
     const { data: groups, isInitialLoading: isLoadingGroups } =
-        useOrganizationGroups({ search, includeMembers: 100 }); // TODO: pagination
+        useOrganizationGroups({ search, includeMembers: 2000 }); // TODO: pagination
```
joaoviana, line 171: non-blocking - we should have this as a constant here

magnew: Yea, I thought about that, but we don't really have any kind of constants location or management. Do you mean just at the top of the file? I can at least do that.

Issue #10537, opened 2024-06-26
Add stream method to warehouse clients


ZeRego:
Work needed for [#10515](https://github.com/lightdash/lightdash/issues/10515)

All WarehouseClients should have a method similar to `runQuery` with a callback argument to handle the data chunks.

Example:
```
    async streamQuery(
        sql: string,
        processBatch: (data: QueryData) => void,
    ): Promise<void> {
         ...
```

Refactor `runQuery` method to use the `streamQuery` and concatenate all the results.

PR #10545, opened 2024-06-27
feat: expose WarehouseClient method to stream results

ZeRego:
Closes: [#10537](https://github.com/lightdash/lightdash/issues/10537)

### Description:

Expose new WarehouseClient method `streamQuery` and deprecate `runQuery`.
Note that we are still using `runQuery` and all rows loaded in memory. We will incrementally refactor our code to use the new streaming method.

### Reviewer actions

test-api test-backend test-frontend

### DIFF
--- a/packages/common/src/types/warehouse.ts
+++ b/packages/common/src/types/warehouse.ts
@@ -14,6 +14,11 @@ export type WarehouseCatalog = {
     };
 };
 
+export type WarehouseResults = {
+    fields: Record<string, { type: DimensionType }>;
+    rows: Record<string, any>[];
+};
+
 export interface WarehouseClient {
     credentials: CreateWarehouseCredentials;
     getCatalog: (
@@ -24,14 +29,27 @@ export interface WarehouseClient {
         }[],
     ) => Promise<WarehouseCatalog>;
 
+    streamQuery(
+        query: string,
+        streamCallback: (data: WarehouseResults) => void,
+        options: {
+            tags?: Record<string, string>;
+            timezone?: string;
+        },
+    ): Promise<void>;
+
+    /**
+     * Runs a query and returns all the results
+     * @param sql
+     * @param tags
+     * @param timezone
+     * @deprecated Use streamQuery() instead to avoid loading all results into memory
+     */
     runQuery(
         sql: string,
         tags?: Record<string, string>,
         timezone?: string,
-    ): Promise<{
-        fields: Record<string, { type: DimensionType }>;
-        rows: Record<string, any>[];
-    }>;
+    ): Promise<WarehouseResults>;
 
     test(): Promise<void>;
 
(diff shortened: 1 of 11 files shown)


packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.test.ts
```diff
@@ -0,0 +1,32 @@
...
+                    close: jest.fn(async () => undefined),
+                })),
+                close: jest.fn(async () => undefined),
+            })),
+            close: jest.fn(async () => undefined),
+        })),
+    })),
+}));
+
+describe('DatabricksWarehouseClient', () => {
```
ZeRego, line 23: Added a sanity test for the Databricks client.

packages/warehouses/src/warehouseClients/WarehouseBaseClient.ts
```diff
@@ -39,13 +39,40 @@ export default class WarehouseBaseClient<T extends CreateWarehouseCredentials>
...
         throw new Error('Warehouse method not implemented.');
     }
 
+    async runQuery(
+        sql: string,
+        tags?: Record<string, string>,
+        timezone?: string,
+    ) {
+        let fields: WarehouseResults['fields'] = {};
+        const rows: WarehouseResults['rows'] = [];
```
rephus, line 59: nice wrapping, but I'm wondering if copying the `data.rows` into  a new `rows` means duplicating all the memory usage for each request now

packages/warehouses/src/warehouseClients/WarehouseBaseClient.ts
```diff
@@ -39,13 +39,40 @@ export default class WarehouseBaseClient<T extends CreateWarehouseCredentials>
...
+
+        await this.streamQuery(
+            sql,
+            (data) => {
+                fields = data.fields;
```
rephus, line 64: so you're overwritting fields with the last chunk of data. Is this compatible with all warehouses ?

packages/common/src/types/warehouse.ts
```diff
@@ -24,14 +29,27 @@ export interface WarehouseClient {
         }[],
     ) => Promise<WarehouseCatalog>;
 
+    streamQuery(
+        query: string,
+        streamCallback: (data: WarehouseResults) => void,
```
rephus, line 34: just curious. do we know roughly how big are the chunks per warehouse ? is it random ? is it a fixed number ? are they consistent per warehouse ?

rephus (commented):
Code look ok, nice wrapping. 

I'm triggering e2e tests to ensure these requests work for all warehouses

packages/warehouses/src/warehouseClients/WarehouseBaseClient.ts
ZeRego: It should be the same amount of memory as before as the current code does the same thing. 🤔 
`rows` memory will increase as it gets more entries and `data.rows` only exists while handling that chunk.

packages/warehouses/src/warehouseClients/WarehouseBaseClient.ts
ZeRego: Yes 👍 I opted to send them in all callback calls to be consistent even if the value is the same. In most warehouse clients we compute the `fields` only once at the start.

packages/common/src/types/warehouse.ts
ZeRego: If I'm not mistaken, this is configurable in the warehouse. Example: https://docs.databricks.com/en/structured-streaming/batch-size.html

rephus (approved):
Tested, everything seems to be working fine 
image


Good job

Issue #10551, opened 2024-06-27
Catalog sidebar


ZeRego:
Related to: Create new SQL runner page skeleton [#10550](https://github.com/lightdash/lightdash/issues/10550)

Create the warehouse catalog sidebar components 

- sidebar is collapsible
- Use "raw" mantine components
- Use new endpoints from [#10516](https://github.com/lightdash/lightdash/issues/10516) to populate the catalog sidebar.

This is just a sketch/ wireframe
- Sketch to show the schema of the table when the table has been selected.
- The schema bit should be draggable. I should be able to control the height of it to see more or fewer fields

![Image](https://github.com/lightdash/lightdash/assets/9117144/3d9ed31e-4228-4fe4-adaf-377085673482)

PR #10568, opened 2024-07-01
feat: sql runner table fields

joaoviana:
Closes: #10551 

### Description:

- uses new type `ApiWarehouseTableFields` when getting table fields 
- adds `react-resizable` so we can resize the table list field
- adds `TableFields` component 

(will add search feature on: https://github.com/lightdash/lightdash/issues/10569)
(will address stable sidebar-collapsing animation on: https://github.com/lightdash/lightdash/issues/10571)

Demo: 

https://github.com/lightdash/lightdash/assets/7611706/9c9ed044-7b00-493b-8545-556a957fb5c4

### Reviewer actions


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2657,15 +2658,23 @@ export class ProjectService extends BaseService {
...
-        return warehouseCatalog;
+        if (!schema) {
+            throw new NotFoundError(
+                'Schema not found in warehouse credentials',
+            );
+        }
+
+        return warehouseCatalog[warehouseClient.credentials.type][schema][
+            tableName
+        ];
```
joaoviana, line 2677: I would like to return only an object with the fields for a table. Would this be the correct way of doing this? 
Do you think we shouldn't  return this format instead?
If we should, what do you think about the handling of undefined `schema`?

packages/e2e/cypress/e2e/api/sqlRunner.cy.ts
```diff
@@ -64,32 +64,23 @@ Object.entries(warehouseConnections).forEach(
...
-                    const schemaResults = databaseResults[schema];
-                    expect(Object.keys(schemaResults)).to.include('orders');
-                    expect(Object.keys(schemaResults)).to.not.include(
-                        'customers',
-                    );
-                    const fieldResults = schemaResults.orders;
+                    const { results } = resp.body;
 
-                    expect(Object.keys(fieldResults)).to.have.length.gt(5);
+                    expect(Object.keys(results)).to.have.length.gt(5);
```
joaoviana, line 76: now we just return an object with the table fields.

packages/frontend/src/features/sqlRunner/hooks/useTableFields.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return useQuery<ApiWarehouseTableFields, ApiError>({
+        queryKey: ['sqlRunner', 'tables', tableName, projectUuid],
+        queryFn: () =>
+            fetchTableFields({
+                projectUuid,
+                tableName,
+            }),
+        retry: false,
+        enabled: !!tableName,
+        // TODO: handle error on no schema found
```
joaoviana, line 33: if the handling of `undefined` schema is OK, I think we should handle it at some point here...?

rephus: afaik the schema is needed to get the tables first, in the backend, so you might not reach this state if that happens...

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2656,16 +2676,28 @@ export class ProjectService extends BaseService {
...
+        const warehouseCatalog = await warehouseClient.getFields(
             tableName,
             schema,
             queryTags,
         );
 
         await sshTunnel.disconnect();
 
-        return warehouseCatalog;
+        return warehouseCatalog[database][schema][tableName];
```
joaoviana, line 2700: now we get the `database` depending on the warehouse. 
LMK what you think @rephus and if you would do it differently.

rephus (approved):
Looks good to me

packages/frontend/src/features/sqlRunner/hooks/useTableFields.tsx
joaoviana: removed comment

Issue #10515, opened 2024-06-25
Stream SQL runner results


rephus:
### Description

Goal: Build an endpoint that doesn't wait until the data is available from the warehouse. 

Acceptance criteria: 
- Main API endpoint should be fast to return (doesn't timeout or wait for results) 
- Memory should always be low in the backend ( do not load all results in memory) 

#### Worker approach
1 - Build an API endpoint to run the SQL query
2 - The backend creates a `job` that is going to be processed by a worker
3 - The worker is going to fetch the data and store it in cloud storage
4 - We have another API endpoint to stream the data from cloud storage

#### Websockets 
1 - Create a websocket to run the SQL query in the backend
2 - The backend will fetch the data  `by batches`  (make sure we don't use too much memory) 
3 - Each chunk will be returned via websocket

Make sure the websocket is stable and doesn't lose any data on `server restart` or `client restart` (at least it should be similar to our current worker implementation) 

### Open questions: 
- do we need to `formatRows` ? if so, when ?

Darshan813:
Hi,

I'm interested in the issue and would like to work on it

owlas:
Hey @Darshan813 the team have this as key focus and should have it fixed within the next few days!

Darshan813:
Alright, no worries. Thank you for getting back.

ZeRego:
# Implementation details

## Worker + pulling

1 - FE requests new results. eg: `[put] /sql-query-job`

2 - BE creates a job and returns job id (trigger warehouse job here and save warehouse query id, is it possible for all wharehouses? )

2.1 - The worker is going to fetch the data based on query id

2.2 - For each chunk: it formats, saves it in storage, updates job state

3 - FE fetches job state every X seconds `[get] /sql/query-job/{job id}`

4 - BE returns latest job state

```jsx
// initial state
{ error: null, isComplete: false, rowsSize: 0, lastUpdate: [date]}

// in progress state
{  error: null, isComplete: false, rowsSize: 100, lastUpdate: [date]}

// complete state
{  error: null, isComplete: true, rowsSize: 146, lastUpdate: [date]}

// error state
{ error: 'SQL has an error', isComplete: false, rowsSize: 0, lastUpdate: [date]}
```

- if job state `lastUpdate` is older than X minutes → update state as failed

5 - when job is complete, FE request the necessary around of results.

- For cartesian charts, we want to load all rows. Might require multiple requests.
- For table chart, we can load just the first page and load more on scroll/page change

[...]

Issue #10238, opened 2024-05-29
Enforce a limit in the SQL runner


jpetey75:
### Description

The SQL runner currently injects `LIMIT 25` at the end of queries, but it's easy to accidentally remove, and if you write a query from scratch there is no limit applied. We should add a limit similar to the one in the table explore that defaults to 500. 

<img width="412" alt="Screenshot 2024-05-29 at 12 43 54 PM" src="https://github.com/lightdash/lightdash/assets/5532776/cd4d8b1c-48db-41da-9ad3-ae878820c696">

### What problem does this solve?

If we enforce a limit, then the page will not crash if too much data is returned in the query results.

ZeRego:
@jpetey75 We don't parse the SQL so we can't inject the limit. Parsing the SQL and supporting all warehouses syntax would be an enormous effort.

Some other options:
- add basic validation to search for the word `limit` in the input
- wrap the user SQL in a CTE and apply the limit in the second query

PR #10575, opened 2024-07-01
feat: Sql runner job streaming

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10515

#### run (schedule job) 
Screenshot from 2024-07-01 14-52-21

#### Get job status 
Screenshot from 2024-07-01 15-16-36
#### On job error 
Screenshot from 2024-07-01 14-53-18

#### Results
Screenshot from 2024-07-01 15-37-36

### New e2e test
[Screencast from 01-07-24 16:50:04.webm](https://github.com/lightdash/lightdash/assets/1983672/375d5561-c143-4122-9c87-7681e8d75c75)

on max retries (reduced interval in snowflake) 

[Screencast from 01-07-24 16:33:40.webm](https://github.com/lightdash/lightdash/assets/1983672/40abf22b-fbbc-493a-8db8-d97e59128aec)

### Reviewer actions


packages/backend/src/controllers/sqlRunnerController.ts
```diff
@@ -69,4 +73,72 @@ export class SqlRunnerController extends BaseController {
...
+        @Request() req: express.Request,
+    ): Promise<any> {
+        this.setStatus(200);
+        this.setHeader('Content-Type', 'application/json');
+
+        const resultsFilePath = await this.services
+            .getProjectService()
+            .getResultsFile(req.user!, projectUuid, fileId);
+
+        // TODO set response headers
```
rephus, line 121: I've tried to add the right response headers, but they don't seem to work, I believe is a conflict with tsoa or sometihng. I can remove these comments and leave it for later.

packages/backend/src/controllers/sqlRunnerController.ts
```diff
@@ -69,4 +73,72 @@ export class SqlRunnerController extends BaseController {
...
+      this.setStatus(200);
+      this.setHeader('Content-Type', mime.lookup(filePath));
+      this.setHeader('Content-Length', stat.size.toString());
+      // Removing this line will cause to not launch the download, just serve the file as it
+      this.setHeader('Content-Disposition', `attachment; filename=${fileName}`);
+      */
+        const mystream = fs.createReadStream(resultsFilePath);
+        const { res } = req;
+        if (res) {
+            mystream.pipe(res);
```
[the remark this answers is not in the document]
rephus: we could also implement this endpoint outside of tsoa

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1,3 +1,4 @@
+import { NotFound } from '@aws-sdk/client-s3';
```
rephus, line 1: I added the new streaming and save file functions into projectService, we should probably create a new `sqlRunnerService` or `streamingService` . If so, I'd like to do it on aseparate PR and move all relevant methods there too (eg: tables/fields)

packages/e2e/cypress/e2e/api/sqlRunner.cy.ts
```diff
@@ -92,6 +93,97 @@ Object.entries(warehouseConnections).forEach(
...
+                        sql,
+                    }),
+                }).then((runResp) => {
+                    expect(runResp.status).to.eq(200);
+                    const { jobId } = runResp.body.results;
+
+                    const maxRetries = 20; // Snowflake is a bit slow compared to the others (normally takes ~3 seconds)
+
+                    // Poll request until job is `completed`
+                    const poll = (retries = 0) => {
```
rephus, line 122: this is a method to wrap the `job` status request, and call it recursively based on the `status` . 

Postgres (locally) takes <100ms to complete jobs, while snowflake takes ~3sec. I don't want to force cy.wait(5s) just to make it pass.

packages/backend/src/controllers/sqlRunnerController.ts
rephus: https://github.com/lukeautry/tsoa/issues/800

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1885,6 +1896,114 @@ export class ProjectService extends BaseService {
         return results;
     }
 
+    async streamResultsToLocalFile(
```
rephus, line 1899: function that creates stores the download file in the database, creates a file stream and pass a function to `write`. 

We should create something similar for s3

packages/backend/src/controllers/sqlRunnerController.ts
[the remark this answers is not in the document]
rephus: if we are going to keep this, I should probably return the readStream in `getResultsFile` instead of the path (I though tsoa should be able to read from a path like express.readFile)

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1885,6 +1896,114 @@ export class ProjectService extends BaseService {
...
+    async streamSqlQueryIntoFile(
+        userUuid: string,
+        projectUuid: string,
+        sql: string,
+    ): Promise<string> {
+        const { organizationUuid } = await this.projectModel.getSummary(
+            projectUuid,
+        );
+
+        await this.analytics.track({
```
rephus, line 1945: I'm keeping the same analytics track, should we make a new one ? / add context ?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1885,6 +1896,114 @@ export class ProjectService extends BaseService {
...
+            organization_uuid: organizationUuid,
+            user_uuid: userUuid,
+        };
+
+        // TODO upload to s3 if enabled
+        const fileId = await this.streamResultsToLocalFile(async (writter) => {
+            await warehouseClient.streamQuery(
+                sql,
+                async ({ rows, fields }) => {
+                    const formattedRows = formatRows(rows, {}); // TODO fields to itemmap
```
rephus, line 1967: we should convert fields to itemmap to properly do formatting... I'm sure how to do it here (and efficiently)

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1885,6 +1896,114 @@ export class ProjectService extends BaseService {
...
+                    formattedRows.forEach(writter);
```
rephus, line 1968: write every line using the file streaming callback

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1885,6 +1896,114 @@ export class ProjectService extends BaseService {
...
+                },
+                {
+                    tags: queryTags,
+                },
+            );
+        });
+
+        await sshTunnel.disconnect();
+        return `/api/v1/projects/${projectUuid}/sqlRunner/results/${fileId}`;
```
rephus, line 1977: points to the new endpoint in sqlRunnerController to return the file.

packages/e2e/cypress/e2e/api/sqlRunner.cy.ts
```diff
@@ -92,6 +93,97 @@ Object.entries(warehouseConnections).forEach(
...
+                                const { fileUrl } = resp.body.results.details;
+                                cy.request({
+                                    url: fileUrl,
+                                    method: 'GET',
+                                }).then((fileResp) => {
+                                    expect(fileResp.status).to.eq(200);
+                                    const lines = fileResp.body
+                                        .trim()
+                                        .split('\n');
+                                    const results = lines.map((line) =>
```
rephus, line 145: this is how we can consume a JSONL, not sure if there is a better way (JSON.load?)

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1885,6 +1896,114 @@ export class ProjectService extends BaseService {
...
+        } finally {
+            writeStream.end(() => {
+                this.logger.debug('File has been saved.');
+            });
+        }
+
+        return downloadFileId;
+    }
+
+    async streamSqlQueryIntoFile(
```
rephus, line 1936: perhaps we need to rename this function and add some code for s3

rephus:
I still need to test the memory usage on large files :eyes:

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: We can add a new `usingStreaming: true` property in case we want to track usage during Beta.
btw, it's not a promise so we don't need the await.

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: We should return the full URL with domain as we will do for S3 files.
We also need to store the `itemMap` in the job state as we need that info in the FE.

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: in the ~database~ temp folder

rephus:
Memory load testing (100k random results) 

### Saving files 
With streaming (not blocking API)
image
image

```
const fileId = await this.streamResultsToLocalFile(async (writter) => {
            await warehouseClient.streamQuery(
                sql,
                async ({ rows, fields }) => {
                    const formattedRows = formatRows(rows, {}); // TODO fields to itemmap
                    formattedRows.forEach(writter);
                },
                {
                    tags: queryTags,
                },
            );
        });
        ```

Without streaming (storing and formatting all rows in memory ) (blocking API)

Screenshot from 2024-07-03 10-01-35

```
const fileId = await this.streamResultsToLocalFile(async (writter) => {
            const { rows } = await await warehouseClient.runQuery(sql);
            const formattedRows = formatRows(rows, {}); // TODO fields to itemmap
            formattedRows.forEach(writter);

        });
        ```

rephus:
Memory load testing (50k rows) 

## Streaming results 

### With read buffer pipe 
Screenshot from 2024-07-03 10-20-02
Screenshot from 2024-07-03 10-20-11

```

 if (res) {
            readStream.pipe(res);
            await new Promise<void>((resolve, reject) => {
                readStream.on('end', () => {
                    res.end();
                    console.info('after readStream', process.memoryUsage())
                    resolve();
                });
            });
        }
```

packages/e2e/cypress/e2e/api/sqlRunner.cy.ts
```diff
@@ -83,6 +88,178 @@ Object.entries(warehouseConnections).forEach(
...
+        // This SQL generates 50k random results on Postgres
+        const sql = `
+            SELECT
+                gs.id AS payment_id,
+                (1000 + random() * 10000)::int AS amount,  -- Random amount between 1000 and 11000
+                (current_date - (random() * 365)::int)::date AS payment_date,  -- Random date within the last year
+                md5(random()::text) AS payment_method,  -- Random payment reference
+                (random() * 100000)::int AS customer_id  -- Random customer_id between 0 and 100000
+            FROM
+                generate_series(1, 50000) AS gs(id)  -- Generate 50k rows
```
ZeRego, line 205: Note that I'm about to enforce a limit of 5k on the SQL runner results. 
Enforce a limit in the SQL runner [#10238](https://github.com/lightdash/lightdash/issues/10238)

ZeRego (approved):

Issue #10571, opened 2024-07-01
Replace Sidebar's `Transition` with an equivalent animation component


joaoviana:
### Description

`Mantine`s Transition unmounts the component when the Transition is applied to a Sidebar (on close)
If we want to keep the component mounted and avoid excessive re-renders, we should replace it with another in-house solution. We could control the component's visibility with `display` or `visibility`. 

Also, `Transition` component has a bug where the animation isn't applied sometimes. Here is the relevant issue that was only fixed in **v7** : https://github.com/mantinedev/mantine/issues/5849

### What problem does this solve?

Avoid unnecessary re-renders and provides a more stable animation experience to the user

ZeRego:
Something we should support: 
- allow to partially collapse sidebar
- allow right sidebar that doesn't show the browser scroll bars during animation
https://github.com/lightdash/lightdash/pull/10578#discussion_r1661329775

jpetey75:
The frontend moved to Mantine 8 in [#26735](https://github.com/lightdash/lightdash/pull/26735) and is on 8.3.18 after [#27037](https://github.com/lightdash/lightdash/pull/27037), which includes the upstream fix for the Transition animation bug referenced here. The Sidebar still renders through Mantine's Transition and unmounts its content when closed, so this stays open for the keep mounted replacement.

PR #10578, opened 2024-07-01
Feat/sql runner new page skeleton

ZeRego:
Closes: 

### Reviewer actions


packages/frontend/src/pages/SqlRunnerNew.tsx
```diff
@@ -1,39 +1,75 @@
...
+                        >
+                            <ActionIcon size="xs">
+                                <MantineIcon
+                                    icon={IconDatabase}
+                                    onClick={() => setLeftSidebarOpen(true)}
+                                />
+                            </ActionIcon>
+                        </Tooltip>
+                    </Paper>
                 )}
```
ZeRego, line 62: todo: update <Page/> to config sidebar closed width and remove this from here

packages/frontend/src/components/common/Page/Page.tsx
```diff
@@ -118,7 +125,7 @@ const usePageStyles = createStyles<string, StyleProps>((theme, params) => {
 
             ...(params.withRightSidebar
                 ? {
-                      width: PAGE_CONTENT_WIDTH_LARGE,
+                      // width: PAGE_CONTENT_WIDTH_LARGE,
```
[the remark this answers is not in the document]
joaoviana: can you do this instead pls?

```tsx
...(params.withRightSidebar && !params.withSidebar
                ? {
                      width: PAGE_CONTENT_WIDTH_LARGE,
                  }
                : {}),
```

packages/frontend/src/pages/SqlRunnerNew.tsx
ZeRego: I don't think we can do it with the Transition component. We should do this work after "Replace Sidebar's Transition with an equivalent animation component [#10571](https://github.com/lightdash/lightdash/issues/10571)"

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -0,0 +1,273 @@
...
+                                }
+                                position="bottom"
+                            >
+                                <ActionIcon
+                                    size="xs"
+                                    onClick={() =>
+                                        setResultsHeight(
+                                            isResultsHeightMoreThanHalf
+                                                ? MIN_RESULTS_HEIGHT
+                                                : maxResultsHeight,
```
joaoviana, line 218: Not sure if we can decrease the `maxResultsHeight` to be something _smaller_ so that we can still see some of the sql when we expand the results.

joaoviana (changes requested):
Looking great! Just need to check the `ContentPanel` a bit more

packages/frontend/src/pages/SqlRunnerNew.tsx
```diff
@@ -1,39 +1,75 @@
...
-                <Box>Sql Runner new</Box>
+                <ContentPanel
+                    isChartConfigOpen={isRightSidebarOpen}
+                    openChartConfig={() => {
+                        setLeftSidebarOpen(false);
```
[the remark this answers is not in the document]
ZeRego: Yes, it is the intended behaviour
> The tables sidebar is collapsible and collapses automatically when the chart config panel opens

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
ZeRego: My interpretation from the wireframe and description was that the SQL panel collapses completely. Once the developer has the correct SQL he collapses the SQL panel and the screen would be all for the chart and chart config. 
We will adjust and refine these details once we get user feedback.

joaoviana (approved):

Issue #10508, opened 2024-06-25
add analytics event for `dashboard_auto_refresh_updated`


TuringLovesDeathMetal:
### Description

Event that is triggered when the `dashboard_auto_refresh` for a dashboard is updated (i.e. turned on or off or changed frequency). 

Parameters:

- `organization_id`
- `project_id`
- `user_id` - who updated the auto refresh
- `dashboard_id`
- `timestamp`
- `frequency` - this is the frequency e.g. `00:05:00` (or we could have it like `5 minutes`), or `off`

PR #10584, opened 2024-07-02
chore: add tracking event for auto refresh update

almeidabbm:
Closes: #10508 

### Description:

- Adds `dashboard_auto_refresh.updated` event

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
+++ b/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
@@ -202,12 +202,6 @@ const DashboardHeader = ({
         ],
     );
 
-    useEffect(() => {
-        // When unmounting the dashboard - navigate away - track the event with undefined interval
-        // Not doing it inside the button component because it get's triggered too often
-        return () => handleDashboardRefreshUpdateEvent(undefined);
-    }, [handleDashboardRefreshUpdateEvent]);
-
     return (
         <PageHeader
             cardProps={{
--- a/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
+++ b/packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
@@ -34,7 +34,7 @@ import {
     IconUpload,
 } from '@tabler/icons-react';
 import dayjs from 'dayjs';
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useHistory, useLocation, useParams } from 'react-router-dom';
 import { useToggle } from 'react-use';
 import { PromotionConfirmDialog } from '../../../features/promotion/components/PromotionConfirmDialog';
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
```diff
@@ -180,6 +180,34 @@ const DashboardHeader = ({
...
+        (intervalMin?: number) => {
+            track({
+                name: EventName.DASHBOARD_AUTO_REFRESH_UPDATED,
+                properties: {
+                    useId: user.data?.userUuid,
+                    dashboardId: dashboardUuid,
+                    organizationId: organizationUuid,
+                    projectId: projectUuid,
+                    frequency: intervalMin ? `${intervalMin} minutes` : 'off',
+                },
```
almeidabbm, line 193: I believe that timestamp is already included in the track correct? I didn't see it getting passed anywhere

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
```diff
@@ -180,6 +180,34 @@ const DashboardHeader = ({
         }),
     );
 
+    const handleDashboardRefreshUpdateEvent = useCallback(
+        (intervalMin?: number) => {
+            track({
+                name: EventName.DASHBOARD_AUTO_REFRESH_UPDATED,
+                properties: {
+                    useId: user.data?.userUuid,
```
joaoviana, line 188:
```suggestion
                    userId: user.data?.userUuid,
```

packages/frontend/src/components/common/Dashboard/DashboardRefreshButton.tsx
```diff
@@ -55,12 +61,16 @@ export const DashboardRefreshButton = () => {
         refreshInterval ? refreshInterval * 1000 * 60 : 0,
     );
 
+    useEffect(() => {
+        onIntervalChange(refreshInterval);
+    }, [onIntervalChange, refreshInterval]);
+
```
joaoviana, line 67: I think we could do this on the `Menu.Item`s so it's clearer, what do you think?

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
```diff
@@ -180,6 +180,34 @@ const DashboardHeader = ({
...
+            projectUuid,
+            track,
+            user.data?.userUuid,
+        ],
+    );
+
+    useEffect(() => {
+        // When unmounting the dashboard - navigate away - track the event with undefined interval
+        // Not doing it inside the button component because it get's triggered too often
+        return () => handleDashboardRefreshUpdateEvent(undefined);
```
joaoviana, line 208: 🤔 I think we should check if unmounted + if there was a refresh value previously...

Also, have you tried wrapping `DashboardRefreshButton` with `memo()`? I believe that reduces the unmounts and also allows you to move this to inside the component.

joaoviana (changes requested):
Requested changes 👍

packages/frontend/src/components/common/Dashboard/DashboardHeader.tsx
almeidabbm: I ended up fixing this in the last commit by having a memo variable called `hasInterval` since the interval changing was what was triggering the undefined call of `onIntervalChange`.
Added the memo but at least by itself didn't help a lot because of the above.

Before:
Screenshot 2024-07-03 at 12 07 12

After:
Screenshot 2024-07-03 at 12 07 44


These are for setting intervals and then navigating away from the dashboard 🎉

joaoviana (approved):
Works great! 💪

Issue #10595, opened 2024-07-03
Add `--only-tables` option to the CLI validate command


almeidabbm:
### Description

- Currently the `validate` command validates the whole project. We should add a `--only-tables` option so that it only validates tables when present. When the option is not present it should still validate the whole project keeping the current behaviour.

### What problem does this solve?

- Allow more flexible CI/CD validation when there are expected chart/dashboard errors.

PR #10597, opened 2024-07-03
feat: add table only validation to cli

almeidabbm:
Closes: #10595

### Description:

- Adds option to validate only tables to CLI

https://github.com/lightdash/lightdash/assets/22939015/fd717aa6-60e9-43c0-8fbc-d032c0b283fa

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.test.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.test.ts
@@ -205,6 +205,25 @@ describe('validation', () => {
             'Model "valid_explore" has a dimension reference: ${is_completed} which matches no dimension',
         );
     });
+
+    it.only('Should validate only tables in project', async () => {
+        (projectModel.getExploresFromCache as jest.Mock).mockImplementationOnce(
+            async () => [exploreError, exploreWithoutDimension],
+        );
+
+        const errors = await validationService.generateValidation(
+            'projectUuid',
+            undefined,
+            true, // onlyTables
+        );
+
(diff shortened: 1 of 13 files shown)


packages/common/src/types/scheduler.ts
```diff
@@ -406,6 +406,7 @@ export type ValidateProjectPayload = {
     userUuid: string;
     organizationUuid: string | undefined;
     explores?: (Explore | ExploreError)[];
+    onlyTables?: boolean;
```
ZeRego, line 409: Should we make this more flexible? An option where the user chooses what they want to validate. Possible options: tables, charts, dashboards.

almeidabbm: I have added a `--only` option which receives an array of `charts,dashboards,tables` + tests for this
Didn't use the `--include` option because there's already a `--exclude` one for models and it could be confusing 💪 

Also updated the video in the desc with the usage + validation check

packages/backend/src/services/ValidationService/ValidationService.test.ts
```diff
@@ -205,6 +205,90 @@ describe('validation', () => {
             'Model "valid_explore" has a dimension reference: ${is_completed} which matches no dimension',
         );
     });
+
+    it('Should validate only tables in project', async () => {
```
ZeRego, line 209: 👨‍🍳 💋 awesome tests

ZeRego (approved):

Issue #10552, opened 2024-06-27
Stream data to new SQL runner page


ZeRego:
Should be done after [#10550](https://github.com/lightdash/lightdash/issues/10550)

Use new endpoints from [#10515](https://github.com/lightdash/lightdash/issues/10515) to populate the results.

Acceptance criteria
- [ ] Install `Redux toolkit` - https://redux-toolkit.js.org/introduction/getting-started
- [ ] Update `immer` to latest version 
- [ ] Create a store in `store/index.ts`
- [ ] Add a sqlRunner reducer in `store/features/sqlRunner/sqlRunnerSlice.ts` - see tutorial here: https://redux-toolkit.js.org/tutorials/typescript

PR #10608, opened 2024-07-04
feat: stream data to sql runner page

joaoviana:
Closes: [#10552](https://github.com/lightdash/lightdash/issues/10552)

### Description:

- Creates hook `useSqlQueryRun` hook that does the following: 
- schedules run query job
- polls a query job status until it's complete or errors
- after a successful job it reads the streamed results, processes them into a `ResultRow[]` format

The component `ContentPanel` will keep track of what the user types on the `sql` + it will set the button as loading until the job settles. 

You can try with this: 

```sql

SELECT *
     FROM "postgres"."jaffle"."users" LIMIT 25

```

Demo:

https://github.com/lightdash/lightdash/assets/7611706/80735faa-eacc-4533-860a-fe846e9c2f8b

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
+++ b/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
@@ -6,7 +6,7 @@ import {
     type ResultRow,
     type SqlRunnerBody,
 } from '@lightdash/common';
-import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import { useMutation, useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 import { lightdashApi } from '../../../api';
 import useToaster from '../../../hooks/toaster/useToaster';
@@ -37,7 +37,6 @@ const scheduleSqlJob = async ({
 export const useSqlQueryRun = () => {
     const { showToastError } = useToaster();
     const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
-    const queryClient = useQueryClient();
     const sqlQueryRunMutation = useMutation<
         ApiJobScheduledResponse['results'],
         ApiError,
@@ -61,8 +60,9 @@ export const useSqlQueryRun = () => {
             return getSchedulerJobStatus(sqlQueryJob.jobId);
         },
         {
-            refetchInterval: (data) => {
+            refetchInterval: (data, query) => {
                 if (
+                    !!query.state.error ||
                     data?.status === SchedulerJobStatus.COMPLETED ||
                     data?.status === SchedulerJobStatus.ERROR
                 )
@@ -76,12 +76,6 @@ export const useSqlQueryRun = () => {
                         subtitle: data.details?.error,
                     });
                 }
-            },
-            onError: async () => {
-                await queryClient.cancelQueries([
-                    'jobStatus',
-                    sqlQueryJob?.jobId,
-                ]);
             },
             enabled: Boolean(sqlQueryJob && sqlQueryJob?.jobId !== undefined),
         },
--- a/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
+++ b/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
@@ -41,7 +41,6 @@ export const useSqlQueryRun = () => {
         ApiJobScheduledResponse['results'],
         ApiError,
         {
-            projectUuid: string;
             sql: SqlRunnerBody['sql'];
         }
     >(({ sql }) => scheduleSqlJob({ projectUuid, sql }), {
--- a/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
+++ b/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
@@ -67,7 +67,7 @@ export const useSqlQueryRun = () => {
                     data?.status === SchedulerJobStatus.ERROR
                 )
                     return false;
-                return 2000;
+                return 1000;
             },
             onSuccess: (data) => {
                 if (data?.status === SchedulerJobStatus.ERROR) {
--- a/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
+++ b/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
@@ -104,19 +104,22 @@ export const useSqlQueryRun = () => {
             const reader = rb?.getReader();
 
             const stream = new ReadableStream({
-                async start(controller) {
-                    async function push() {
-                        while (true) {
-                            const nextChunk = await reader?.read();
-                            if (nextChunk?.done) {
+                start(controller) {
+                    function push() {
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,168 @@
...
+    );
+
+    const { data: sqlQueryResults, isLoading: isResultsLoading } = useQuery<
+        ResultRow[] | undefined,
+        ApiError
+    >(
+        ['sqlQueryResults', sqlQueryJob?.jobId],
+        async () => {
+            const url = scheduledDeliveryJobStatus?.details?.fileUrl;
+            const response = await fetch(url, {
```
joaoviana, line 97: the `fileUrl` is the whole url, rather than the `fileId`. If I use `lightdashApi` it will prefix the url and I don't want that. 
Would it be ok to use fetch here or should we be better at this?
I can get the `fileId` from the `fileUrl` or I can look at the BE implementation to send back the `fileId` instead of the url.

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,171 @@
...
+                method: 'GET',
+                headers: {
+                    Accept: 'application/json',
+                },
+            });
+            const rb = response.body;
+            const reader = rb?.getReader();
+
+            const stream = new ReadableStream({
+                start(controller) {
```
joaoviana, line 107: from this example: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#fetch_stream

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
joaoviana: keeping this approach because the URL from `fileUrl` can include a _custom_ s3 endpoint

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,171 @@
...
+ */
+export const useSqlQueryRun = () => {
+    const { showToastError } = useToaster();
+    const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
+    const queryClient = useQueryClient();
+    const sqlQueryRunMutation = useMutation<
+        ApiJobScheduledResponse['results'],
+        ApiError,
+        {
+            projectUuid: string;
```
ZeRego, line 45: We can remove this param.

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,171 @@
...
+            return getSchedulerJobStatus(sqlQueryJob.jobId);
+        },
+        {
+            refetchInterval: (data) => {
+                if (
+                    data?.status === SchedulerJobStatus.COMPLETED ||
+                    data?.status === SchedulerJobStatus.ERROR
+                )
+                    return false;
+                return 2000;
```
ZeRego, line 70: Since we want this to be available as soon as possible I would reduce it to 1 second.

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,171 @@
...
+                        title: 'Could not run SQL query',
+                        subtitle: data.details?.error,
+                    });
+                }
+            },
+            onError: async () => {
+                await queryClient.cancelQueries([
+                    'jobStatus',
+                    sqlQueryJob?.jobId,
+                ]);
```
ZeRego, line 84: Curious, why does it need to cancel itself?

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,171 @@
...
+        ResultRow[] | undefined,
+        ApiError
+    >(
+        ['sqlQueryResults', sqlQueryJob?.jobId],
+        async () => {
+            const url = scheduledDeliveryJobStatus?.details?.fileUrl;
+            const response = await fetch(url, {
+                method: 'GET',
+                headers: {
+                    Accept: 'application/json',
```
ZeRego, line 100: fyi: There is still no standard for jsonl mime type. 
https://github.com/wardi/jsonlines/issues/19
We will keep json for now 👍

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -0,0 +1,171 @@
...
+            ),
+        },
+    );
+
+    const isLoading = useMemo(
+        () =>
+            scheduledDeliveryJobStatus?.status === SchedulerJobStatus.STARTED &&
+            isResultsLoading,
+        [scheduledDeliveryJobStatus?.status, isResultsLoading],
+    );
```
ZeRego, line 164: should it have `sqlQueryRunMutation.isLoading` to account for the first query and for scheduledDeliveryJobStatus to be undefined?

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
joaoviana: done: https://github.com/lightdash/lightdash/pull/10608/commits/0202e3ae712438b789159a3982813ece6418fdc1

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
joaoviana: Good point, I removed this and instead stop the `refetch`ing if there's an error on the `query`: https://github.com/lightdash/lightdash/pull/10608/commits/a670d2b3b647ae1f3acc3c19e4fa91f1c429ad47 lmk what you think!

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
joaoviana: oh yes! 
[chore: remove unused mutation param](https://github.com/lightdash/lightdash/pull/10608/commits/625fae41c50b55454b13d3bb717fe9a58882ff87)

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
joaoviana: added! https://github.com/lightdash/lightdash/pull/10608/commits/b56c5f0c38eb520eeb5fab66f54870e988d07f62 and tested

packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
joaoviana: had no idea! Thanks for sharing 👍

ZeRego (approved):

PR #10609, opened 2024-07-04
fix: fix duplicate copied tabbed dashboard error

Shifan-Gu:
Closes: 

when we duplicate the tabbed dashboard, we copied the tab over to new dashboard with old tab uuid, which breaks the 1-1 cardinally on the database side, later when we duplicate the whole project the copy logic breaks.

Steps to reproduce the issue:

1.  create a new empty dashboard
2. create a tab
3. create a tile inside the tab
4. duplicate the dashboard by hitting "Duplicate" button
5. run lightdash cli "start-preview" command to start the preview env
6. check error logs 

### Reviewer actions


packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -406,12 +412,17 @@ export class DashboardService extends BaseService {
                 }),
             );
 
+            const newTabbedTiles = updatedTiles.map((tile) => ({
+                ...tile,
+                tabUuid: newTabs.find((tab) => tab.uuid === tile.tabUuid)?.uuid,
```
ZeRego, line 417: Isn't this always `undefined` ?  the `tile.tabUuid` has uuids from the original dashboard and the `newTabs` has the newly generated uuids. So there is never a match 🤔 

We need a `tabUuidMap` that tracks the old and new uuids. 
We can also move this inside the `updatedTiles` loop.

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1659,9 +1659,11 @@ export class ProjectModel {
                     )
                     .returning('*');
             }
-            const dashboardTabsMapping = dashboardTabs.map((c, i) => ({
+            const dashboardTabsMapping = dashboardTabs.map((c) => ({
                 uuid: c.uuid,
-                newUuid: newDashboardTabs[i].uuid,
+                newUuid: newDashboardTabs.find(
+                    (t) => t.dashboard_version_id === c.dashboard_version_id,
+                )?.uuid!,
```
ZeRego, line 1666: All tabs have the same `dashboard_version_id` so this returns the same uuid on each loop.
We can have something like you did for the other file, generate the new uuids before the insert loop.

Shifan-Gu: Hi @ZeRego you are right this logic is broken, i was too rush on the fix :(

Intention of this line is to handle a special case that, we have some tabbed dashboards copied when this fix is not in place, so in our `dashboard_tab` database table is bit messed up, leads to a result that the `tab_uuid` is not unique. but in this line for each tab row we generate new `tab_uuid` for it, so the old `tab_uuid` is mapped to many new `tab_uuid`, in this case only when you use new `tab_uuid` in combination with `dashboard_version_id` can make sure you can map back to the old original tab, i'v updated the code to flip the mapping logic to incorporate this case, let me know if that make sense to you.

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1737,7 +1738,10 @@ export class ProjectModel {
                                               m.id === d.dashboard_version_id,
                                       )?.newId!,
                                   tab_uuid: dashboardTabsMapping.find(
-                                      (m) => m.uuid === d.tab_uuid,
+                                      (m) =>
+                                          m.uuid === d.tab_uuid &&
+                                          m.newDashboardVersionId ===
+                                              d.dashboard_version_id,
```
almeidabbm, line 1744: @Shifan-Gu from what I am seeing while testing, this will return undefined since `newDashboardVersionId` in the tabs mapping never matches the tile's `dashboard_version_id` and therefore the value of `tab_uuid` in the preview project `dashboard_tile` rows will always be `NULL`

I believe we can go around this by adding `dashboardVersionId: dashboardTabs[i].dashboard_version_id` to the tabs mapping and use it in this check instead of `m.newDashboardVersionId`.
After this we can probably remove `newDashboardVersionId` from the tabs mapping as it won't be used I believe.

almeidabbm (changes requested):

almeidabbm (approved):

packages/backend/src/models/ProjectModel/ProjectModel.ts
almeidabbm: Code looks good, thank you @Shifan-Gu 🚀

PR #10610, opened 2024-07-04
feat: save sql charts

rephus:
Closes:   #10639

### Description:

Adds SQL chart endpoints:
- get
- create
- update 
- delete

Screenshot 2024-07-11 at 15 18 20

### Reviewer actions


packages/backend/src/models/SavedSqlModel.ts
```diff
@@ -0,0 +1,312 @@
...
+        },
+    ): Promise<string> {
+        const [{ saved_sql_version_uuid: savedSqlVersionUuid }] = await trx(
+            SavedSqlVersionsTableName,
+        ).insert(
+            {
+                saved_sql_uuid: data.savedSqlUuid,
+                sql: data.sql,
+                config: data.config,
+                chart_kind: ChartKind.VERTICAL_BAR, // todo: get chart kind from config
```
magnew, line 229: I spent longer than I wanna admit thinking I was getting something wrong in the FE cause everything kept getting rendered as bar charts 😂 

That'll still be the case until we update these

packages/backend/src/services/SavedSqlService/SavedSqlService.ts
```diff
@@ -0,0 +1,226 @@
...
+        }
+        const space = await this.spaceModel.getSpaceSummary(sqlChart.spaceUuid);
+        const { isPrivate } = space;
+        const access = await this.spaceModel.getUserSpaceAccess(
+            user.userUuid,
+            sqlChart.spaceUuid,
+        );
+        if (
+            user.ability.cannot(
+                'create',
```
magnew, line 95: The FE lets them send these requests for now. This access check means its 'safe' but we can disable saving them if they done have this.

packages/common/src/types/sqlRunner.ts
```diff
@@ -14,3 +20,71 @@ export type SqlRunnerBody = {
...
+
+export type ApiSqlChart = {
+    status: 'ok';
+    results: SqlChart;
+};
+
+export type ApiCreateSqlChart = {
+    status: 'ok';
+    results: {
+        savedSqlUuid: string;
```
magnew, line 80: This is all we need, does it need a type? I just stubbed it out like this in the API call

magnew (approved):
🔥 Looks great. The best review I can give this is I wrote a bunch of code against it today. The only real issue I hit was the chart type default.

packages/backend/src/models/SavedSqlModel.ts
ZeRego: Fixed in the UI PR as we have the final config types there 👍

packages/common/src/types/sqlRunner.ts
ZeRego: Would it return the same chart kind type as the one sent in the body? My idea was for the FE to change the URL and fetch the saved SQL with the GET endpoint.

Issue #10166, opened 2024-05-23
Space group permissions don't show warnings on different roles


rephus:
This is one UI thing we should fix to avoid confusion

If I add a user , there is a warning message that appears if the user does not have enough `project` permissions to perform actions in the space

![Screenshot from 2024-05-21 08-41-16](https://github.com/lightdash/lightdash/assets/1983672/00448687-71ab-42dd-a7aa-aaa915614bff)

But When I add a group, this message doesn't appear, and I can't make changes in the space (so I don't have edit permissions) 
![Screenshot from 2024-05-21 08-41-35](https://github.com/lightdash/lightdash/assets/1983672/da66b9e8-97dc-4b1a-ad00-c37a0b1d723f)

_Originally posted by @rephus in https://github.com/lightdash/lightdash/pull/10082#pullrequestreview-2067682195_

PR #10644, opened 2024-07-08
fix: use project level role to show warning on frontend

almeidabbm:
Closes: #10166 

### Description:

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -809,6 +809,11 @@ export class SpaceModel {
                         ...spaceGroupAccessRoles,
                     ]);
 
+                    const highestProjectRole = getHighestProjectRole([
+                        inheritedOrgRole,
+                        inheritedProjectRole,
+                    ]);
+
                     // exclude users with no space role
                     if (!highestRole) {
                         return acc2;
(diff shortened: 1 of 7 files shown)


joaoviana:
Could you fix the type error on `build` @almeidabbm ? I will look at the PR soon!

packages/backend/src/models/SpaceModel.ts
```diff
@@ -846,6 +851,7 @@ export class SpaceModel {
                             hasDirectAccess: !!user_with_direct_access,
                             inheritedRole: highestRole.role,
```
joaoviana, line 852: I believe this means that we don't use `inheritedRole` anymore. 
Could you update the `inheritedRole` to be `highestProjectRole?.role` or something similar? We don't need `projectRole` being added I think 🤔

joaoviana (commented):
Everything else looks fine and tested! 
Just checking if we need an extra property, and then will approve 👍

packages/backend/src/models/SpaceModel.ts
almeidabbm: I am seeing `inheritedRole` being used in `packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceUserList.tsx` 😞

joaoviana: oops! ignore!

joaoviana (approved):

Issue #10635, opened 2024-07-08
Create table component and config mapping


ZeRego:
Based on last proposal in [Investigate chart config](https://github.com/lightdash/lightdash/issues/10567) #10567 

Goals:
- table component
- function/class that gets the table config and the results and returns all the data needed for the table component

PR #10648, opened 2024-07-08
chore: create table component and config

joaoviana:
Closes: #10635

### Description:

- Adds `TableDataTransformer`
- Adds `Table` component under `/visualizations`
- Gets style components already used in other tables' viz and applies these to the `Table` component

On `TableDataTransformer`

- Reads the results and provides all necessary methods to render a table
- Accounts for elements in the config (which will be done on a different issue) and will get columns depending on the config passed.

Demo:

https://github.com/lightdash/lightdash/assets/7611706/a582fef5-4485-426b-8a08-1f0c9d0aa514

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/visualizations/Table.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/visualizations/Table.tsx
@@ -1,0 +1,35 @@
+import { SqlRunnerResultsTransformer } from '@lightdash/common';
+import { type FC } from 'react';
+import CommonTable from '../../../../components/common/Table';
+import { getRawValueCell } from '../../../../hooks/useColumns';
+import { type useSqlQueryRun } from '../../hooks/useSqlQueryRun';
+
+type Props = {
+    data: NonNullable<ReturnType<typeof useSqlQueryRun>['data']>;
+};
+
+export const Table: FC<Props> = ({ data }) => {
+    const results = new SqlRunnerResultsTransformer({ data });
+
+    const rows = results.getRows();
+    const columns = results.getColumns();
+
+    return (
+        <CommonTable
+            status="success"
+            data={rows}
+            columns={columns.map((s) => ({
+                id: s,
+                accessorKey: s,
+                header: s.toLocaleUpperCase(),
+                cell: getRawValueCell,
+            }))}
+            pagination={{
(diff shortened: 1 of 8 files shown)


packages/frontend/src/features/sqlRunner/components/visualizations/Table.tsx
```diff
@@ -0,0 +1,148 @@
...
+    type TableChartSqlConfig,
+} from '../../transformers/TableDataTransformer';
+
+type Props = {
+    data: NonNullable<ReturnType<typeof useSqlQueryRun>['data']>;
+    config?: TableChartSqlConfig;
+};
+
+export const Table: FC<Props> = ({ data }) => {
+    const processor = new TableDataTransformer(data, undefined); // TODO: add config once we have it
```
magnew: I'm still a little fuzzy on why we want to use classes here 😅. I think its a pretty uncommon pattern to instantiate classes inside a functional component because kind of the whole point of functional components is that they don't hold onto their own state. Having class instances here skirts the suggested react state management approach where state centrally managed somewhere (redux, a context, etc). 

If we want to do this I think we have to at least put this (as well and the following `get`s) in `useMemo` calls. Looking around the internet I see people saying that's possible, but no one saying its a good idea. I think we are going to have state management problems. 

I think the `TableDataTransformer` class makes sense if we want to use it's functionality on the BE and FE, but the more reacty way to use it here is probably to instantiate it in the parent, then pass it to this component. In some ways that move the problem up the tree, so we can talk about if we want a different pattern. React encourages functional programming pretty heavily and haven't used classes in view components a lot so I don't have good instincts about it.

magnew (commented):
Im not gonna be nit picky about the details cause this is early stages, but I do have one big structural question.

packages/frontend/src/features/sqlRunner/components/visualizations/Table.tsx
joaoviana: All great points above @magnew, added a class that only takes care of the shared BE-FE methods and created a hook to consume the transformer + intialize React-UI-related code (like `useReactTable` and `useVirtualizer`)

magnew (approved):
Great work

Issue #10637, opened 2024-07-08
Create bar chart component


ZeRego:
Based on last proposal in https://github.com/lightdash/lightdash/issues/10567 https://github.com/lightdash/lightdash/issues/10567

Goals:

- echart component
- function/class that gets the bar config and the results and returns all the echarts object

PR #10656, opened 2024-07-09
chore: sqlrunner bar chart component

magnew:
Closes: #10637 

### Description:

Screenshot 2024-07-09 at 19 08 25

### Reviewer actions


packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1923,6 +1923,8 @@ const useEchartsCartesianConfig = (
         return undefined;
     }
 
+    console.log('eChartsOptions', eChartsOptions);
```
joaoviana, line 1926: can remove the logs everywhere 👍

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { type useSqlQueryRun } from '../../hooks/useSqlQueryRun';
+import { useBarChartDataTransformer } from '../../transformers/useBarChartDataTransformer';
+
+type BarChartProps = Omit<EChartsReactProps, 'option'> & {
+    data: NonNullable<ReturnType<typeof useSqlQueryRun>['data']>;
+};
+
+const BarChart: FC<BarChartProps> = memo(({ data, className, ...rest }) => {
+    // TODO: fix store type
+    const barChartConfig = useSelector(
```
joaoviana, line 14:
```suggestion
    const barChartConfig = useAppSelector(
```

we have 2 hooks that should be used instead of importing the ones from `react-redux` - types are inferred correctly

packages/common/src/transformers/BarChartDataTransformer.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// TODO: Should this just be ChartDataTransformer?
+export class BarChartDataTransformer {
+    private transformer: SqlRunnerResultsTransformer;
+
+    constructor(args: { data: ResultRow[] }) {
+        this.transformer = new SqlRunnerResultsTransformer(args);
+    }
+
+    public getEchartsSpec(config: BarChartConfig) {
```
joaoviana, line 13:
```suggestion
    public getEchartsSpec(config: BarChartConfig | undefined) {
```

This could be `undefined`

packages/common/src/transformers/BarChartDataTransformer.ts
```diff
@@ -0,0 +1,68 @@
...
+        const data = this.transformer.getRows();
+        const columns = this.transformer.getColumns();
+        const { axesConfig, seriesConfig } = config;
+
+        if (config.type !== 'barChart') return {};
```
joaoviana, line 18: I don't think we need this for now... let's treat this as a `barChart` until we decide what to do here

packages/common/src/types/visualizations.ts
```diff
@@ -0,0 +1,31 @@
+export type BarChartConfig = {
+    metadata: {
+        version: number;
+    };
+    // TODO: There was some back and forth about whether this should be 'defaultType'
+    // I'm not sure
+    type: 'barChart';
```
joaoviana, line 7: we can remove this - not needed for MVP

packages/common/src/types/visualizations.ts
```diff
@@ -0,0 +1,31 @@
...
+              }
+            | undefined;
+    };
+    axesConfig: {
+        x: {
+            reference: string;
+            label: string;
+        };
+        y: {
+            reference: string;
```
joaoviana, line 22: we don't need `reference` here - we should map from the `seriesConfig`

joaoviana (changes requested):
Looking great 👍

packages/common/src/types/visualizations.ts
magnew: Cause we don't know how we want to do this yet?

packages/common/src/types/visualizations.ts
magnew: I don't follow. These should be the axis configurations for the various series. So not just things like label, but also it could be things like tick spacing. I think we'll need it.

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
magnew: 👍 Cool, I didn't know that.

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
magnew: Definitley

packages/common/src/types/visualizations.ts
joaoviana: exactly 👍

packages/common/src/types/visualizations.ts
joaoviana: Sharing an example soon!

packages/common/src/types/visualizations.ts
```diff
@@ -0,0 +1,31 @@
+export type BarChartConfig = {
```
joaoviana, line 1:
```suggestion
export type BarChartSqlConfig = {
```

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { memo, type FC } from 'react';
+import { type useSqlQueryRun } from '../../hooks/useSqlQueryRun';
+import { useAppSelector } from '../../store/hooks';
+import { useBarChartDataTransformer } from '../../transformers/useBarChartDataTransformer';
+
+type BarChartProps = Omit<EChartsReactProps, 'option'> & {
+    data: NonNullable<ReturnType<typeof useSqlQueryRun>['data']>;
+};
+
+const BarChart: FC<BarChartProps> = memo(({ data, className, ...rest }) => {
```
joaoviana, line 12:
```suggestion
const BarChart: FC<BarChartProps> = memo(({ data }) => {
```

and we can update the props above ⬆️

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
```diff
@@ -0,0 +1,34 @@
...
+    // TODO: fix store type
+    const barChartConfig = useAppSelector(
+        (state: any) => state.sqlRunner.chartConfig,
```
joaoviana, line 15:
```suggestion
        (state) => state.sqlRunner.chartConfig,
```

joaoviana: And actually we can move back the type of 
```
chartConfig: BarChartConfig | undefined;
```
to just be

```
chartConfig: BarChartConfig;
```

makes sense what you had before, sorry

packages/common/src/types/visualizations.ts
magnew: I figured we might use these in another context than SQL runner, so leaving this generic is good. There's nothing about it that is tightly coupled to SQL. Thoughts?

joaoviana (approved):

packages/common/src/types/visualizations.ts
joaoviana: I called mine `TableChartSqlConfig` but I don't have a strong opinion so happy to change mine. You can keep this one sure 👍

Issue #10638, opened 2024-07-08
table/chart config panel


ZeRego:
Goals:
- allow user to change between table and bar chart
- allow user to configure table/bar chart

joaoviana:
I'm looking at how this would work with RTK/Redux and the current Table component config+viz.
I'm using this PR as a base: https://github.com/lightdash/lightdash/pull/10648

At the moment I'm caring about: 
- Consuming the config
- Dispatching an action
- Consuming the altered config

cc @owlas

owlas:
Finished, moving to https://github.com/lightdash/lightdash/issues/10769

PR #10663, opened 2024-07-09
feat: add table config panel

joaoviana:
Relates to: https://github.com/lightdash/lightdash/issues/10638

### Description:

- Dispatches action to set initial table config when fetching results (note: this is temporary and will be different once we can save sql-runner charts)
- Allows changing table label fields 
- This doesn't include changes to the config panel's copy and other interactions.

Demo:

https://github.com/lightdash/lightdash/assets/7611706/57ab915a-6ddb-4612-ae49-b6b65b12536b

### Reviewer actions


packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -18,6 +22,39 @@ export const sqlRunnerSlice = createSlice({
         setProjectUuid: (state, action: PayloadAction<string>) => {
             state.projectUuid = action.payload;
         },
+        setInitialResultsTableConfig: (
```
magnew, line 25: Why not have a default config in the store 'initialState'? Is this just so the initialisation pattern is always the same whether or not we are populating it from the database? 

I see that this is setting up all of the columns, but could the table render fine with a default config and not have entries for each column initially?

I don't have a strong opinion, just trying to understand.

packages/frontend/src/features/sqlRunner/transformers/TableDataTransformer.ts
```diff
@@ -4,30 +4,31 @@ import { getRawValueCell } from '../../../hooks/useColumns';
...
-                  visible: boolean;
-                  reference: string;
-                  label: string;
-                  frozen: boolean;
-                  order?: number;
-              }
-          >;
-      }
-    | undefined;
+export type TableChartSqlConfig = {
```
magnew, line 7: I put the bar chart type in packages/common/src/types/visualizations.ts

magnew (approved):
Cool

packages/frontend/src/features/sqlRunner/transformers/TableDataTransformer.ts
joaoviana: Sounds good!

magnew:
I merged main in here to fix the conflict

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
joaoviana: > I see that this is setting up all of the columns, but could the table render fine with a default config and not have entries for each column initially?

So since we rely on the `results` to render the columns config we need to set this after the query endpoint.

If we did set a default config in the store, then we would have to loop them somehow to render the table correctly 🤔 I'm not too sure on how this would look like but I think that at this moment we don't need to set a default because we still need to reference fields/columns

packages/frontend/src/features/sqlRunner/transformers/TableDataTransformer.ts
joaoviana: Done here: https://github.com/lightdash/lightdash/pull/10663/commits/a6beebcfcc7fa2fea6064b91e0dfc8410a2ea92e
didn't remove the word `Sql` because there's already a TableChartConfig
we should revisit this after being able to save charts IMO

Issue #10432, opened 2024-06-17
restrict scheduled delivery + google sheets sync frequency to no more than hourly


rephus:
Thread: https://lightdash.slack.com/archives/C02GQKJK84Q/p1718614092615079?thread_ts=1718607237.293609&cid=C02GQKJK84Q

- make this clear in the UI 
- throw an error when you try to create/save a scheduled delivery with a frequency ofmore than hourly.

rephus:
I believe this should be done on a separate PR 

- automatically disable existing deliveries that have a frequency that's more than hourly.

PR #10671, opened 2024-07-10
feat: limit schedulers hourly

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10432

### Description:

Extra fix: 

Fixed cron max width on schedulers ( I didn't fix it on sync) 
Before:
Screenshot from 2024-07-10 11-34-11

After:
Screenshot from 2024-07-10 11-34-18

#### Deliveries
Screenshot from 2024-07-10 11-40-12

#### Sync
Screenshot from 2024-07-10 11-45-03

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/scheduler.ts
+++ b/packages/common/src/utils/scheduler.ts
@@ -12,7 +12,7 @@ export function getHumanReadableCronExpression(cronExpression: string) {
     return valueWithUTC[0].toLowerCase() + valueWithUTC.slice(1);
 }
 
-export function isAllowedCronExpression(cronExpression: string): boolean {
+export function isValidFrequency(cronExpression: string): boolean {
     /** This function will return False if:
      * - the cronExpression is not valid (not 5 parts separated by spaces)
      * - the cronExpression frequency is less than 1 hour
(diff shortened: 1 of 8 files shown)


packages/common/src/utils/scheduler.ts
```diff
@@ -11,3 +11,30 @@ export function getHumanReadableCronExpression(cronExpression: string) {
...
+export function isAllowedCronExpression(cronExpression: string): boolean {
+    /** This function will return False if:
+     * - the cronExpression is not valid (not 5 parts separated by spaces)
+     * - the cronExpression frequency is less than 1 hour
+     */
+    const cronParts = cronExpression.trim().split(' ');
+    if (cronParts.length !== 5) {
+        // Invalid cron expression
+        return false;
+    }
```
almeidabbm, line 24: Nitpick: this could be replaced by using `isInvalidCronExpression` although it is currently in `packages/frontend/src/utils/fieldValidators.ts` it would need to be moved to `common`

almeidabbm (approved):

packages/common/src/utils/scheduler.ts
rephus: I've been looking at that method, but the type and arguments are different, `isInvalidCronExpression` returns an `error` string, while my method returns a boolean. 
The only thing those 2 methods have in common is the `split().length !== 5` , which in my method is even optional.

What I can do is use the method from `common` inside the validator, let me try.

rephus: updated

almeidabbm (approved):

Issue #10549, opened 2024-06-27
The users list in the organization settings should be paginated (or load incrementally)


TuringLovesDeathMetal:
### Description

One of our users has > 1500 users in the organization. This means that when the users page loads int he org settings, it freezes and takes a minute or two to actually load. 

Instead, this list should be paginated or should load incrementally to avoid this lag. 

<img width="967" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/c38fee17-156c-4a56-af99-607e9674b82e">

PR #10681, opened 2024-07-10
feat: user list pagination

almeidabbm:
Closes: #10549 

### Description:

- Adds pagination to user list
- Adds backend search (using regex matching) so that we're able to search paginated results instead of just results on each page

Note: tried using different libs that already implemented pagination but there was always some problem with typescript types coming from knex. Ended up deciding on our own pagination so that we can maintain it as we go and add functionality to it if needed

https://github.com/lightdash/lightdash/assets/22939015/251f428e-8e35-4d5e-831d-b80ea04be309

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/organizationController.ts
+++ b/packages/backend/src/controllers/organizationController.ts
@@ -10,6 +10,7 @@ import {
     ApiSuccessEmpty,
     CreateGroup,
     CreateOrganization,
+    IKnexPaginateArgs,
     OrganizationMemberProfileUpdate,
     UpdateAllowedEmailDomains,
     UpdateOrganization,
@@ -175,13 +176,24 @@ export class OrganizationController extends BaseController {
     async getOrganizationMembers(
         @Request() req: express.Request,
         @Query() includeGroups?: number,
+        @Query() pageSize?: number,
+        @Query() page?: number,
     ): Promise<ApiOrganizationMemberProfiles> {
         this.setStatus(200);
+        let paginateArgs: IKnexPaginateArgs | undefined;
+
+        if (pageSize && page) {
+            paginateArgs = {
+                page,
+                pageSize,
+            };
+        }
+
         return {
             status: 'ok',
             results: await this.services
                 .getOrganizationService()
-                .getUsers(req.user!, includeGroups),
+                .getUsers(req.user!, includeGroups, paginateArgs),
         };
     }
 
--- a/packages/backend/src/controllers/organizationController.ts
+++ b/packages/backend/src/controllers/organizationController.ts
@@ -10,7 +10,7 @@ import {
     ApiSuccessEmpty,
     CreateGroup,
     CreateOrganization,
-    IKnexPaginateArgs,
+    KnexPaginateArgs,
     OrganizationMemberProfileUpdate,
     UpdateAllowedEmailDomains,
     UpdateOrganization,
@@ -181,7 +181,7 @@ export class OrganizationController extends BaseController {
         @Query() searchQuery?: string,
     ): Promise<ApiOrganizationMemberProfiles> {
         this.setStatus(200);
-        let paginateArgs: IKnexPaginateArgs | undefined;
+        let paginateArgs: KnexPaginateArgs | undefined;
 
         if (pageSize && page) {
             paginateArgs = {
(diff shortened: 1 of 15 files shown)


packages/backend/src/controllers/organizationController.ts
```diff
@@ -175,13 +176,25 @@ export class OrganizationController extends BaseController {
     async getOrganizationMembers(
         @Request() req: express.Request,
         @Query() includeGroups?: number,
+        @Query() pageSize?: number,
+        @Query() page?: number,
+        @Query() searchQuery?: string,
     ): Promise<ApiOrganizationMemberProfiles> {
         this.setStatus(200);
+        let paginateArgs: KnexPaginateArgs | undefined;
```
rephus, line 184: we might want to reuse this on any endpoint that implements pagination, so it would be nice to extract this into a single function . 

But we can do it later, once this is reused.

packages/backend/src/database/pagination/errors.ts
```diff
@@ -0,0 +1,6 @@
+export class KnexPaginationError extends Error {
```
rephus, line 1: why not move it inside `common` `errors` next to the other error types , and `extends LightdashError`

packages/backend/src/models/OrganizationMemberProfileModel.ts
```diff
@@ -192,11 +218,25 @@ export class OrganizationMemberProfileModel {
...
-        const result: (DbOrganizationMemberProfile & {
+        if (searchQuery) {
+            orgMembersAndGroupsQuery = getColumnMatchRegexQuery(
+                orgMembersAndGroupsQuery,
+                searchQuery,
+                ['first_name', 'last_name', 'email', 'role'],
+            );
+        }
+
+        const { pagination, data } = await KnexPaginate.paginate(
```
rephus, line 229: nice

packages/backend/src/database/pagination/index.ts
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            if (pageSize < 1) {
+                throw new KnexPaginationError(
+                    'pageSize should be greater than 0',
+                );
+            }
+
+            const offset = (page - 1) * pageSize;
+            const count = await query.clone().clear('select').count();
```
rephus, line 23: this is making an extra query to the database to get the number of rsults, right ? 

Should we use `Promises.all` ?  here ?

packages/backend/src/models/SearchModel/utils/search.ts
```diff
@@ -45,6 +45,21 @@ export function getRegexFromUserQuery(query: string) {
...
 
+export function getColumnMatchRegexQuery(
+    queryBuilder: Knex.QueryBuilder,
+    searchQuery: string,
+    columns: string[],
+) {
+    const regex = getRegexFromUserQuery(searchQuery).source;
+
+    // use regexp_matches
+    return queryBuilder.where((builder) => {
```
rephus, line 56: I'm wondering how efficient this regex is...

rephus (changes requested):
It works great, just one bug we need to fix, when we make a new search, we should reset the `page` 

[Screencast from 11-07-24 13:36:12.webm](https://github.com/lightdash/lightdash/assets/1983672/0e26547f-75ca-476e-b7fb-9f35d0f7bdef)


Unit tests are not very useful here, but I'd love to see some `api` integration tests for this at some point. This can be done on a separate issue.

packages/backend/src/database/pagination/index.ts
almeidabbm: Yeah, to calculate the total number of pages
Good point, we can add that here yeah 🤔 shouldn't cause any issue

packages/backend/src/models/SearchModel/utils/search.ts
almeidabbm: This regex just splits the words in the query by spaces and then joins them using an OR conditional operator `|` so it needs to match full words in your search query - also it is case insensitive

I believe it should be fine as people will be searching just by name, email, role

Btw this is already being used to find matching tables and fields in the global search

rephus (approved):
Changes look good. :sheep: it !

PR #10685, opened 2024-07-10
chore: add bar chart config

magnew:
Closes: 

### Description:

Adds config options for bar chart
- rename axes
- rename series
- also updates chart style

### Reviewer actions

### DIFF
--- a/packages/common/src/transformers/BarChartDataTransformer.ts
+++ b/packages/common/src/transformers/BarChartDataTransformer.ts
@@ -59,10 +59,10 @@ export class BarChartDataTransformer {
                     return newRow;
                 }),
             },
-            series: series?.map((s) => ({
+            series: series?.map((s, idx) => ({
                 dimensions: [xField, s.reference],
                 type: 'bar',
-                name: s.name,
+                name: axes?.y[idx]?.label ?? s.reference,
                 encode: {
                     seriesName: s.reference,
                     x: xField,
--- a/packages/common/src/transformers/BarChartDataTransformer.ts
+++ b/packages/common/src/transformers/BarChartDataTransformer.ts
@@ -13,7 +13,7 @@ export class BarChartDataTransformer {
     public getEchartsSpec(config: BarChartConfig) {
         const data = this.transformer.getRows();
         const columns = this.transformer.getColumns();
-        const { axes } = config || { axes: undefined };
+        const { axes, series } = config || { axes: undefined };
 
         const xField = columns[0];
         const yField = columns[1];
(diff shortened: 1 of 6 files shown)


packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -56,10 +56,11 @@ export const sqlRunnerSlice = createSlice({
         setProjectUuid: (state, action: PayloadAction<string>) => {
             state.projectUuid = action.payload;
         },
-        setInitialResultsTableConfig: (
+        setInitialResultsAndSeries: (
```
magnew, line 59: I combined the chart and table reducers here to avoid a double-dispatch. There are other ways to do this if you'r prefer.

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -75,10 +62,42 @@ export const sqlRunnerSlice = createSlice({
...
+                        reference: Object.keys(action.payload[0])[0],
+                        label: Object.keys(action.payload[0])[0],
+                    },
+                    y: [
+                        {
+                            reference: Object.keys(action.payload[0])[1],
+                            label: Object.keys(action.payload[0])[1],
+                        },
+                    ],
+                },
```
joaoviana, line 92:
```suggestion
               const [xKey, yKey] = Object.keys(action.payload[0]);

const axes = {
    x: {
        reference: xKey,
        label: xKey,
    },
    y: [
        {
            reference: yKey,
            label: yKey,
        },
    ],
};
```

could do something like this, but just a nit-pick

packages/common/src/types/visualizations.ts
```diff
@@ -17,18 +17,19 @@ export type BarChartConfig = {
...
             reference: string;
-            position: 'left' | 'right';
+            position?: 'left' | 'right';
             label: string;
         }[];
     };
     series: {
         reference: string;
         yIndex: number;
+        name: string;
```
joaoviana, line 32: I believe the `label` of the `y` should be used instead of adding a `name` here.

So, for a serie X, I have a `yIndex`.
I can attribute a `label` to `y`
That should be updated and the `axes.y[serie.yIndex]` will get the `label` for us. That way you don't need to add `name`

LMK if you want to chat about this

packages/common/src/transformers/BarChartDataTransformer.ts
```diff
@@ -43,15 +59,16 @@ export class BarChartDataTransformer {
                     return newRow;
                 }),
             },
-            series: columns.slice(1).map((s) => ({
-                dimensions: [xField, s],
+            series: series?.map((s) => ({
```
joaoviana, line 62: 😍

joaoviana (commented):
Some considerations on the `series` type

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
magnew: I did something similar

packages/common/src/types/visualizations.ts
magnew: Updated.

I feel like we might end up going back to separating series and axis config cause this makes it so you can set series labels separately from the axes, but this is fine. Echarts separates axis and series config un propose so you can do things like name the axes differently than the series or have multiple series on an axis.

Issue #10602, opened 2024-07-04
You should be able to copy the SQL snippet when the SQL panel is closed


PriPatel:
> I regularly find myself wishing for the Copy SQL Code button to be available on the SQL banner when it's collapsed.
> I regularly copy the code and paste it in BigQuery to check how much the query costs, and I have to first expand the SQL section, then click on copy, and then collapse it again. So having it on the banner would save a few seconds every time

For users that don't want to actually examine the SQL in Lightdash, but rather copy it over into another tool , having to open the panel , click the copy button, then close the panel feels inefficient. 

**Possible solution**
- [ ] Show a 'copy' button when you hover over the closed SQL panel

cooldude6000:
hey,
To whoever it may concern,
I'm working on this issue.
I will SURELY comment back here if I'm unable to make it with this issue.
Thanks.

PR #10723, opened 2024-07-15
feat: added a copy sql button when card is closed

cooldude6000:
Closes: #10602

solves issue #10602 

### Description:

https://github.com/user-attachments/assets/f981297e-3ce8-48b6-abe4-32c66b921461

-----

This Is WORKING now. It WILL copy the SQL too, unlike the last time😂. (#10717)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/SqlCard/SqlCard.tsx
+++ b/packages/frontend/src/components/Explorer/SqlCard/SqlCard.tsx
@@ -1,5 +1,8 @@
 import { subject } from '@casl/ability';
+import { ActionIcon, CopyButton, Tooltip } from '@mantine/core';
+import { IconCheck, IconClipboard } from '@tabler/icons-react';
 import { memo, type FC } from 'react';
+import { useCompiledSql } from '../../../hooks/useCompiledSql';
 import { useApp } from '../../../providers/AppProvider';
 import {
     ExplorerSection,
@@ -26,6 +29,8 @@ const SqlCard: FC<SqlCardProps> = memo(({ projectUuid }) => {
     );
     const { user } = useApp();
 
+    const { data: sql } = useCompiledSql();
+
     const sqlIsOpen = expandedSections.includes(ExplorerSection.SQL);
     return (
         <CollapsableCard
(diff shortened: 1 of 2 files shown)


cooldude6000:
@joaoviana

joaoviana (changes requested):
Could you add these changes? 
- only appear on hover if the card is closed 
- should be next to the title `SQL`, instead of being on the far right 
- perhaps `title` is a `prop` that can take a `ReactNode`?

cooldude6000:
Hi @joaoviana ,



https://github.com/user-attachments/assets/28c89ba2-3819-46af-b98a-77bb3a3dc4dd




**NOW** :
> only appear on hover if the card is closed

The copy button will be visible ONLY when card is **CLOSED** && it is **HOVERED**

> should be next to the title SQL, instead of being on the far right

It is now😅


lmk if anything to change


Ps: Regarding the changes in SqlCard.tsx file : I removed all the changes made in it and it is now the exactly same file like (https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/components/Explorer/SqlCard/SqlCard.tsx) . I DID NOT make any changes to it (i mean, i _removed_ the ones I did in my previous commit and the file is back to default). I made changes ONLY in CollapsableCard.tsx.

cooldude6000:
Also, 
> perhaps `title` is a prop that can take a `ReactNode`?

Is it just changing:

**FROM**:

`title: string;` _(Line 55 of CollapsebleCard.tsx)_

TO:

`title: React.ReactNode` _(Line 55 of CollapsebleCard.tsx)_

?

cooldude6000:
@joaoviana

joaoviana:
made a few changes here @cooldude6000. 

For future reference, we should ensure that components in `common` have a general implementation and act as a skeleton for the other components that reuse them. 
So in this PR, I've added an optional `ref` prop that can be passed so that we can track the on-hover effect 
Then on the SQL card component I add `headerElement` and keep this logic isolated in the component that takes care of it

Thanks for contributing @cooldude6000 !

joaoviana (approved):

PR #10728, opened 2024-07-15
chore: make sql runner name and description updatable

magnew:
Closes: #10726 

### Description:

- Makes it possible to update the SQL runner name in place. 
- That name then gets used in the modal.
- Makes it possible to enter a description when saving
- Note that the description is currently not displayed anywhere. 

Note that once save, these can't be updated. We dont have them in the versioned types. I wasn't sure whether to add them or not -- the way that types are set up it looks intentional. Is there any reason these should be kept out of versioning? If so that's fine, but we have to change some things to plumb updated to name and description through. 

### Reviewer actions

### DIFF
--- a/packages/common/src/types/sqlRunner.ts
+++ b/packages/common/src/types/sqlRunner.ts
@@ -110,7 +110,7 @@ export type CreateSqlChart = {
     name: string;
     description: string | null;
     sql: string;
-    config: SqlRunnerChartConfig;
+    config: SqlRunnerChartConfig | {};
     spaceUuid: string;
 };
 
--- a/packages/common/src/types/sqlRunner.ts
+++ b/packages/common/src/types/sqlRunner.ts
@@ -46,7 +46,7 @@ export type BarChartConfig = {
         version: number;
     };
     type: ChartKind.VERTICAL_BAR;
-    style: {
+    style?: {
         legend:
             | {
                   position: 'top' | 'bottom' | 'left' | 'right';
@@ -54,7 +54,7 @@ export type BarChartConfig = {
               }
             | undefined;
     };
-    axes: {
+    axes?: {
         x: {
             reference: string;
             label?: string;
@@ -65,7 +65,7 @@ export type BarChartConfig = {
             label: string;
         }[];
     };
-    series: {
+    series?: {
         reference: string;
         yIndex: number;
         name: string;
@@ -110,7 +110,7 @@ export type CreateSqlChart = {
     name: string;
     description: string | null;
     sql: string;
-    config: SqlRunnerChartConfig | {};
+    config: SqlRunnerChartConfig;
     spaceUuid: string;
 };
 
--- a/packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
@@ -7,6 +7,7 @@ import {
     Text,
     Textarea,
     TextInput,
+    type ModalProps,
 } from '@mantine/core';
 import { useForm, zodResolver } from '@mantine/form';
 import { IconChartBar } from '@tabler/icons-react';
@@ -25,12 +26,9 @@ const validationSchema = z.object({
 
 type FormValues = z.infer<typeof validationSchema>;
 
-type Props = {
-    isOpen: boolean;
-    onClose: () => void;
-};
+type Props = Pick<ModalProps, 'opened' | 'onClose'>;
 
-export const SaveSqlChartModal: FC<Props> = ({ isOpen, onClose }) => {
+export const SaveSqlChartModal: FC<Props> = ({ opened, onClose }) => {
     const dispatch = useAppDispatch();
     const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
     const { data: spaces = [] } = useSpaceSummaries(projectUuid, true);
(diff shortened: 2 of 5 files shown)


packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -40,7 +44,8 @@ const initialState: SqlRunnerState = {
     projectUuid: '',
     activeTable: undefined,
     savedChartUuid: undefined,
-    name: 'Untitled SQL Query',
+    name: '',
```
joaoviana, line 47: do we want to use `DEFAULT_NAME` here?

packages/common/src/types/sqlRunner.ts
```diff
@@ -110,7 +110,7 @@ export type CreateSqlChart = {
     name: string;
     description: string | null;
     sql: string;
-    config: SqlRunnerChartConfig;
+    config: SqlRunnerChartConfig | {};
```
joaoviana, line 113: When creating a SQL chart, do we want it to be undefined instead?
Because in TS `{}` represents any value that isn't `null` or `undefined`, so I think we can use something else, like a `null`?

packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
```diff
@@ -1,40 +1,59 @@
...
     name: z.string().min(1),
+    description: z.string(),
 });
 
 type FormValues = z.infer<typeof validationSchema>;
 
-export const SaveSqlChartModal = () => {
+type Props = {
+    isOpen: boolean;
+    onClose: () => void;
```
joaoviana, line 30: we could pick this from `ModalProps` : `'opened' | 'onClose'`

packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
```diff
@@ -1,40 +1,59 @@
...
     const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
     const { data: spaces = [] } = useSpaceSummaries(projectUuid, true);
 
+    const name = useAppSelector((state) => state.sqlRunner.name);
+    const description = useAppSelector((state) => state.sqlRunner.description);
+
     const form = useForm<FormValues>({
         initialValues: {
             name: '',
+            description: '',
```
joaoviana, line 44: have we thought about having `name` and `description` from RTK in the `initialValues` of this form?
```
 const form = useForm<FormValues>({
        initialValues: {
            name,
            description,
```

Wonder if this would be a good approach 🤔

joaoviana (changes requested):
some Qs!

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
magnew: No, we want this to by undefined until someone saves. 

Maybe it should be called NAME_PLACEHOLDER instead of default

packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
magnew: They get put there in the useEffect below. I don't like that approach, but theres a race condition populating this. If you just put them here directly, they don't get set. If you know of a better way, let me know.

packages/common/src/types/sqlRunner.ts
magnew: Good call. I wasn't quite sure the best way to go here. Now that I'm looking, it might actually be best to default it to the most minimum set of config

```
    metadata: {
        version: number;
    };
    type:  string;
```

magnew: I will update

joaoviana: oooh nice, unless you have to handle that in a lot of places, then an `undefined`/`null` is probably better - but I like the minimal set of config approach as well 👍

packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
joaoviana: got it! ignore then :)

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
joaoviana: All good, you can ignore then 👍

packages/common/src/types/sqlRunner.ts
magnew: I did this and it seems to work. I put it in the create call for now, but we might want to think about if the defaulting should happen somewhere else. Like, should the version number get generated in the BE? I think thats a separate change.

joaoviana (approved):
just needs some extra optional chaining on `RightSidebar`

magnew:
> just needs some extra optional chaining on `RightSidebar`

What do you mean? 😄 Is this a change I should make here?

joaoviana:
> > just needs some extra optional chaining on `RightSidebar`
> 
> What do you mean? 😄 Is this a change I should make here?

here: https://github.com/lightdash/lightdash/actions/runs/9956057990/job/27505130321?pr=10728 `build` failed 👍

magnew:
> > > just needs some extra optional chaining on `RightSidebar`
> > 
> > 
> > What do you mean? 😄 Is this a change I should make here?
> 
> here: https://github.com/lightdash/lightdash/actions/runs/9956057990/job/27505130321?pr=10728 `build` failed 👍

AH, gotcha.

PR #10729, opened 2024-07-16
fix: auto join org to apply okta group

rephus:
Closes:  https://github.com/lightdash/lightdash-internal-work/issues/1889

### Description:

So, what's happening: 

- users on the org are not joining via invitation
- org has a whitelisted domain, so the user joins automatically the org
- however, when we check for groups, the user hasn't joined the group yet
- on /joinOrg we don't have information about the group

How this fixes it.

- Targeted fix: if `  !createdUser.organizationUuid && !inviteCode &&  hasGroups &&  allowedOrgs.length === 1`
- Join the org
- Then apply group

Screenshot from 2024-07-16 09-52-57

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -600,7 +600,7 @@ export class UserService extends BaseService {
             openIdUser.openId.groups.length;
         let { organizationUuid } = createdUser;
         if (
-            !createdUser.organizationUuid &&
+            !organizationUuid &&
             !inviteCode &&
             hasGroups &&
             allowedOrgs.length === 1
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/UserService.ts
```diff
@@ -589,21 +589,40 @@ export class UserService extends BaseService {
...
             this.lightdashConfig.groups.enabled === true &&
             this.lightdashConfig.auth.enableGroupSync === true &&
             Array.isArray(openIdUser.openId.groups) &&
-            openIdUser.openId.groups.length &&
-            createdUser.organizationUuid
-        )
+            openIdUser.openId.groups.length;
+        let { organizationUuid } = createdUser;
+        if (
+            !createdUser.organizationUuid &&
```
almeidabbm, line 603:
```suggestion
            !organizationUuid &&
```

almeidabbm (approved):
Small nitpick since `organizationUuid` has already been destructured, otherwise code looks good 🚀

PR #10730, opened 2024-07-16
chore: fix misc sql runner style and errors

magnew:
### Description:

Fix some styles and errors in new SQL runner:
- Remove double borders. Note that the one between the editor and the results is a drag handle. We could change the style if we want
- Remove redundant tooltip on Run Query button
- Fix input radius on field search
- Fix console error from resizable box. It can only have one child.
- Update results table layout

 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -84,7 +84,17 @@ export const ContentPanel: FC<Props> = ({
             ref={wrapperRef}
         >
             <Tooltip.Group>
-                <Paper shadow="none" radius={0} px="md" py="sm" withBorder>
+                <Paper
+                    shadow="none"
+                    radius={0}
+                    px="md"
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -84,7 +84,17 @@ export const ContentPanel: FC<Props> = ({
...
+                <Paper
+                    shadow="none"
+                    radius={0}
+                    px="md"
+                    py="sm"
+                    sx={(theme) => ({
+                        borderWidth: '0 0 1px 1px',
+                        borderStyle: 'solid',
+                        borderColor: theme.colors.gray[3],
+                    })}
```
magnew, line 96: These are a little annoying but I didnt see another great way to get only the borders we want

joaoviana (commented):
maybe some borders can be removed when SQL editor is collapsed?

image

joaoviana (approved):

PR #10735, opened 2024-07-16
feat: display sql chart with new endpoint to get results in dashboard

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/10692

### Description:

Adds `getSavedSqlChartWithResults` endpoint, which will: 
- return `chart` & jobID
- schedule a job with the `chart.sql` to get the results

How to test:
- Save a sql runner chart and get its `uuid`
- Call PATCH to update dashboard to add a `SQL_CHART` chart

screenshot:
Screenshot 2024-07-17 at 09 46 59

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/sqlRunnerController.ts
+++ b/packages/backend/src/controllers/sqlRunnerController.ts
@@ -3,6 +3,7 @@ import {
     ApiErrorPayload,
     ApiJobScheduledResponse,
     ApiSqlChart,
+    ApiSqlChartWithResults,
     ApiSuccessEmpty,
     ApiUpdateSqlChart,
     ApiWarehouseCatalog,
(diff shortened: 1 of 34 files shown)


joaoviana:
@rephus this endpoint would be called on the SQL chart tile so that we can use the results + config to render the chart/table

packages/backend/src/controllers/sqlRunnerController.ts
```diff
@@ -188,6 +189,40 @@ export class SqlRunnerController extends BaseController {
...
+    @Get('saved/{uuid}/chart-and-results')
+    @OperationId('getSavedSqlChartWithResults')
+    async getSavedSqlChartWithResults(
+        @Path() uuid: string,
+        @Path() projectUuid: string,
+        @Request() req: express.Request,
+    ): Promise<ApiSqlChartWithResults> {
+        this.setStatus(200);
+
+        const chart = await this.services
```
rephus, line 209: one `minimal` performance optimization. Since the `job` is going to take longer, perhaps we want to do `scheduleSqlJob` first. Or even do both in parallel with Promise.all

PR #10751, opened 2024-07-17
feat: allow toggle slack app to use socket mode

Shifan-Gu:
Closes: 

### Reviewer actions

### DIFF
--- a/packages/backend/src/clients/Slack/Slackbot.ts
+++ b/packages/backend/src/clients/Slack/Slackbot.ts
@@ -92,6 +92,7 @@ export class SlackBot {
                 ...slackOptions,
 
                 receiver: slackReceiver,
+                socketMode: !!slackOptions.appToken,
             });
 
             this.addEventListeners(app);
--- a/packages/backend/src/clients/Slack/Slackbot.ts
+++ b/packages/backend/src/clients/Slack/Slackbot.ts
@@ -74,28 +74,42 @@ export class SlackBot {
         }
 
         try {
-            const slackReceiver = new ExpressReceiver({
-                ...slackOptions,
-                installationStore: {
-                    storeInstallation: (i) =>
-                        this.slackAuthenticationModel.createInstallation(i),
-                    fetchInstallation: (i) =>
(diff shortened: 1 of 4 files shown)


packages/backend/src/clients/Slack/Slackbot.ts
```diff
@@ -74,27 +74,42 @@ export class SlackBot {
...
-                logLevel: LogLevel.INFO,
-                app: expressApp,
-            });
-
-            const app = new App({
-                ...slackOptions,
-
-                receiver: slackReceiver,
-            });
+            if (this.lightdashConfig.slack?.appToken) {
```
ZeRego, line 77: @Shifan-Gu Let's introduce a `SLACK_SOCKET_MODE` env var and use it for this condition. Two reasons for this: explicit configuration, prevent backwards compatibility issues as old instances might still have `SLACK_APP_TOKEN` env var.

packages/backend/src/clients/Slack/Slackbot.ts
```diff
@@ -74,27 +74,42 @@ export class SlackBot {
...
+                const app = new App({
+                    ...slackOptions,
+                    installationStore: {
+                        storeInstallation: (i) =>
+                            this.slackAuthenticationModel.createInstallation(i),
+                        fetchInstallation: (i) =>
+                            this.slackAuthenticationModel.getInstallation(i),
```
ZeRego, line 84: Can we handle the delete here?
```
                        deleteInstallation: (i) =>
                            this.slackAuthenticationModel.deleteInstallation(i),
```

packages/backend/src/clients/Slack/Slackbot.ts
Shifan-Gu: sure, will do 👍

packages/backend/src/clients/Slack/Slackbot.ts
Shifan-Gu: yes, its missing and should be in place, good catch 👍

ZeRego (approved):

Issue #10668, opened 2024-07-10
Share SQL Runner url


ZeRego:
When user clicks the "share url" button we save the current state in the database and return a state uuid to be used as a URL param. 

Unsaved SQL chart: `https:// ... /sql-runner-new?state={uuid}`
Saved SQL chart: `https:// ... /sql-runner-new/{uuid}?state={uuid}`

When the page loads it checks the URL param `state` and fetches/loads the state based on the uuid.

We don't want to add the entire state to the URL as this as caused issues before with url length limits.

rephus:
What is this `state` ? is this a new table with the sql config ? or just the `sharedUrl` we use on `explore charts`

PR #10753, opened 2024-07-17
feat: Share sql chart url

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10668

### Description:

[Screencast from 17-07-24 16:57:28.webm](https://github.com/user-attachments/assets/8366c19d-be63-4f84-8585-21d38654dcec)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/Header.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/Header.tsx
@@ -1,6 +1,6 @@
 import { ChartKind } from '@lightdash/common';
 import { ActionIcon, Group, Paper, Tooltip } from '@mantine/core';
-import { IconDeviceFloppy, IconLink } from '@tabler/icons-react';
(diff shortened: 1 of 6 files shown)


joaoviana (commented):
I'm assuming we don't want to run the SQL Query on load of URL, and I think that's fine bc we can iron out the UX and what Limit can be applied. 

Checking the code now 👍

joaoviana (changes requested):
Code looks good, but I believe we should add the `state` to the URL search params when the user clicks `share` button...

So when user clicks on an unsaved **or** saved chart:
- get shared url
- set the `state` search param to the shared URL uuid

joaoviana (approved):
works really well!

Perhaps there could be something we can address on a future PR once we're ready to release this. Like
- I save a chart
- I make a change and share
- I go to link and go back to original change
- there should be no `state` param bc it's back to normal 
But I think this is just a detail

Issue #10745, opened 2024-07-17
Refactor space page to use new v2 endpoint


ZeRego:
After https://github.com/lightdash/lightdash/issues/10694

Refactor Space.tsx to not use useChartSummaries and use the new endpoint /api/v2/content.
Refactor Space.tsx to not use the charts and dashboards from `useSpace` and use the new endpoint /api/v2/content?spaceUuids={uuid}.

PR #10758, opened 2024-07-17
feat: allow adding sql charts to dashboard

joaoviana:
Closes: #10744

### Description:

Allows adding tiles to a dashboard that are SQL charts

Uses the new v2 endpoint: `useChartSummariesV2` and allows `lightdashApi` to have a different prefix.
I've also added a chart icon to each select item when adding chart tiles to the dashboard

Improved loading state of `useSqlChartAndResults` hook and when setting the `chart` data if the job is cached.

Demo

https://github.com/user-attachments/assets/86030e9e-75a8-4c40-80cb-6e54790951d8

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

Issue #10616, opened 2024-07-05
New SQL runner miscellaneous UI/UX fixes


ZeRego:
Tracking small fixes while working on this milestone:
- [x] #10726
- [x] #10779
- [ ] #10829
- [x] #10683
- [ ] #10935
- [x] #10686
- [x] #10687
- [x] add default SQL. eg `SELECT * FROM lightdash-analytics.prod.github_daily LIMIT 25`
- [x] #10789
- [x] #10679
- [x] fix/remove double borders between panels 
- [x] ~reduce left sidebar padding (nice-to-have)~
- [x] #10781
- [x] debounce label field overrides and/or look into unblocking the UI thread when this happens
- [x] #10690
- [x] add content panel empty state
- [x] #10738
- [x] Add quoting when coping a table (e.g. when I click `pages` it should copy `` `database`.`table`.`pages` ``
- [x] It is not obvious to me that you can drag the results panel. I was stuck for a while with only full screen sql or full screen chart, I couldn't see both
- [x] adjust height of sql editor. the editor can go behind the `charts / results` panel, which means the last lines on the sql are hidden. 
- [x] can save charts in dashboards
- [x] #10783
- [x] ~Add autocomplete for fields. We can populate when a user selects a table and gets a list of fields. To get **all** fields it probably requires an extra endpoint~
- [ ] #10927
- [x] If I click on a sql in a dashboard, it takes me to the sql-runner-new but with a uuid instead of a slug, so it doesn't load correctly.
- [x] If I open this sql chart as a viewer, and try to run it, I get a backend error (403) as expected, but no feedback on the frontend.
- [x] #10844
- [ ] #10936
- [ ] #10937

Open a separate ticket/PR for each that requires multiple file/lines changes.

Faisal-imtiyaz123:
@ZeRego can I work on this?

ZeRego:
Hi @Faisal-imtiyaz123 👋 thanks for your interest in Lightdash. Tickets from this milestone are not open to external contributions as we are actively working on this milestone.

PR #10767, opened 2024-07-18
Space page v2

rephus:
Closes: [#10745](https://github.com/lightdash/lightdash/issues/10745)

### Description:

Screenshot from 2024-07-18 12-10-35

### Reviewer actions

### DIFF
--- a/packages/frontend/src/api.ts
+++ b/packages/frontend/src/api.ts
@@ -12,8 +12,6 @@ export const BASE_API_URL =
         ? `http://test.lightdash/`
         : import.meta.env.BASE_URL;
 
-const apiPrefix = `${BASE_API_URL}api/v1`;
-
 const defaultHeaders = {
     'Content-Type': 'application/json',
     [LightdashRequestMethodHeader]: RequestMethod.WEB_APP,
@@ -38,13 +36,17 @@ type LightdashApiProps = {
     url: string;
     body: BodyInit | null | undefined;
     headers?: Record<string, string> | undefined;
+    version?: 'v1' | 'v2';
 };
 export const lightdashApi = async <T extends ApiResponse['results']>({
     method,
     url,
     body,
     headers,
+    version = 'v1',
 }: LightdashApiProps): Promise<T> => {
+    const apiPrefix = `${BASE_API_URL}api/${version}`;
+
     let sentryTrace: string | undefined;
     // Manually create a span for the fetch request to be able to trace it in Sentry. This also enables Distributed Tracing.
     Sentry.startSpan(
--- a/packages/frontend/src/components/common/ResourceView/resourceUtils.ts
+++ b/packages/frontend/src/components/common/ResourceView/resourceUtils.ts
@@ -1,6 +1,7 @@
 import {
     assertUnreachable,
     ChartKind,
+    ChartSourceType,
     ResourceViewItemType,
     type ResourceViewItem,
 } from '@lightdash/common';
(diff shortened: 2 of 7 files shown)


rephus:
Before merging, we need toc onfirm what's the default behaviour with private spaces

rephus:
I've just tested with private spaces, and I can list all the charts inside a private space as admin 
image

packages/frontend/src/api.ts
```diff
@@ -38,13 +36,17 @@ type LightdashApiProps = {
...
+    version?: 'v1' | 'v2';
 };
 export const lightdashApi = async <T extends ApiResponse['results']>({
     method,
     url,
     body,
     headers,
+    version = 'v1',
 }: LightdashApiProps): Promise<T> => {
+    const apiPrefix = `${BASE_API_URL}api/${version}`;
```
joaoviana, line 48: I like this a lot more than I have on my PR: https://github.com/lightdash/lightdash/pull/10758/files

rephus:
INtersting, this e2e failed. I'll investigate

image

packages/frontend/src/components/common/ResourceView/resourceUtils.ts
```diff
@@ -54,6 +55,9 @@ export const getResourceUrl = (projectUuid: string, item: ResourceViewItem) => {
         case ResourceViewItemType.DASHBOARD:
             return `/projects/${projectUuid}/dashboards/${item.data.uuid}/view`;
         case ResourceViewItemType.CHART:
+            if (item.data.source === ChartSourceType.SQL) {
+                return `/projects/${projectUuid}/sql-runner-new/saved/${item.data.uuid}`;
```
joaoviana, line 59: perhaps `item.data.slug` instead?

joaoviana (commented):
I have this chart: 
https://lightdash-pr-10767.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/sql-runner-new/saved/tsting-out

And I can't see it on the `all saved charts` - should I expect to not see them yet?

packages/frontend/src/components/common/ResourceView/resourceUtils.ts
rephus: it seems slug is not currently exposed in `content` :eyes:

rephus:
> And I can't see it on the all saved charts - should I expect to not see them yet?
No, not yet, afaik this ticket was only for `space` page. 

I believe we'll update those eventually, but  at least we need to have a way to list thoes sql charts somewhere , and I think this was the first option.

packages/frontend/src/components/common/ResourceView/resourceUtils.ts
joaoviana: I see 👁️
Currently, links only work with `slug`, so having the `uuid` won't work - **yet**

Perhaps this should be a ticket

joaoviana:
> > And I can't see it on the all saved charts - should I expect to not see them yet?
> > No, not yet, afaik this ticket was only for `space` page.
> 
> I believe we'll update those eventually, but at least we need to have a way to list thoes sql charts somewhere , and I think this was the first option.

I see it! 
image


OK ignore me then. 

We should also be able to add this to a space on creation of the chart, which will be done here: https://github.com/lightdash/lightdash/issues/10616 - `allow to pick space on creation`

joaoviana:
Could you make this change in `useChartSummariesV2`?

```ts
const getChartSummariesInProjectV2 = async (projectUuid: string) => {
    return lightdashApi<ApiChartContentResponse['results']>({
        url: `/content?projectUuids=${projectUuid}&contentTypes=${ContentType.CHART}&pageSize=${Number.MAX_SAFE_INTEGER}`, // TODO: remove pageSize max once we have pagination
        method: 'GET',
        body: undefined,
        version: 'v2',
    });
};
```

joaoviana (approved):

Issue #10789, opened 2024-07-22
Add auto-complete for table names while typing on SQL editor


joaoviana:
Acceptance criteria
- [ ] When I type a table, I get an autocomplete/intellisense suggestion to use it - for example "postgres"."orders"
- [ ] I **don't** get autocomplete for fields available (needs an extra endpoint)

PR #10792, opened 2024-07-22
feat: add SQL editor autocompletion

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/10789

### Description:

Gets tables from server state (cached because the list on left side-bar is already loaded)
Generates quoted table autocompletion.

Demo:

https://github.com/user-attachments/assets/3c6f148f-c5b2-4907-b5a3-dee091cc7ebf

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
@@ -109,6 +109,7 @@ const LIGHTDASH_THEME = {
 const registerCustomCompletionProvider = (
     monaco: Monaco,
     language: string,
+    quoteChar: string,
     tables: string[],
 ) => {
     monaco.languages.registerCompletionItemProvider(language, {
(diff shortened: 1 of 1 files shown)


owlas (approved):

rephus:
We only autocomplete tables, right ? it would be useful to also autocomplete `fields` at least for the selected table 

image

rephus:
Field autocompletion is out of scope

-  I don't get autocomplete for fields available (needs an extra endpoint)
 
 But I don't think we need a new endpoint for it, right ? it would be nice to also have fields, at least for the `selected table` is this going to be done on a separate ticket ?

joaoviana:
> But I don't think we need a new endpoint for it, right ? it would be nice to also have fields, at least for the `selected table` is this going to be done on a separate ticket ?

It will be done 

> Field autocompletion is out of scope
> 
> * I don't get autocomplete for fields available (needs an extra endpoint)
> 
> But I don't think we need a new endpoint for it, right ? it would be nice to also have fields, at least for the `selected table` is this going to be done on a separate ticket ?

Yup, agree, it would be nice to include the fields when the user selects a table! But to get all we then need a different approach... I've added a separate task in: https://github.com/lightdash/lightdash/issues/10616

Issue #10749, opened 2024-07-17
Move sql vis config into transformers 


magnew:
### Description

Move sql vis config into transformers. And out of the slice. The config is stored in the slice but produced in the transformers. User changes get stored in the slice.

PR #10794, opened 2024-07-23
fix: extract bar chart config component

owlas:
Closes: #10749 

### Description:

Moves the bar chart config to its own slice

### Reviewer actions


packages/frontend/src/features/sqlRunner/components/visualizations/BarChartConfiguration.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const DEBOUNCE_TIME = 500;
+
+const BarChartConfiguration: FC<BarChartConfigurationProps> = ({
+    value,
+    onChange,
+}) => {
+    const onXAxisLabelChange = debounce((label: string) => {
+        onChange({
+            ...value,
```
magnew, line 20: Doing all of this spreading before sending the values to the store seems to me like skirting some of the point of using RTK. You were concerned about having too many specific setters for individual properties, no? But it seems like this way we have moved that complexity out of the slice and now we have to do this error-prone spreads in these configuration files for every property. I don't think this is less code than the way we were doing it before where there were actions per-property, right? And I think it's probably slower and more error prone than passing one property to the store and letting relying on RTK/Immer to set the nested properties. One of the things RTK is good at/good for is surgically alerting object without having to spread/copy the whole thing.

Maybe I missed a reason to do this 😅? If we do want to only have one setter in the store that takes the whole config, maybe we should use Immer here?

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
```diff
@@ -1,15 +1,15 @@
...
 import { useBarChartDataTransformer } from '../../transformers/useBarChartDataTransformer';
 
 type BarChartProps = {
     data: NonNullable<ReturnType<typeof useSqlQueryRun>['data']>;
-    config: BarChartConfig | undefined;
 } & Partial<Pick<EChartsReactProps, 'style'>>;
 
-const BarChart: FC<BarChartProps> = ({ data, config, style }) => {
+const BarChart: FC<BarChartProps> = ({ data, style }) => {
+    const config = useAppSelector((state) => state.barChartConfig.config);
```
joaoviana: Could you move to the parent component and keep the prop `config` pls? 
So it doesn't break sql charts in dashboards (we don't have RTK in any other part of the app)

packages/frontend/src/features/sqlRunner/store/index.ts
```diff
@@ -1,13 +1,14 @@
 // FIXES ts2742 issue with configureStore
 // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
-import type * as rtk from '@reduxjs/toolkit';
```
joaoviana, line 3: we need to add this back so the build passes pls 👍

packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
```diff
@@ -0,0 +1,77 @@
...
+                    },
+                    series: fieldIds.slice(1).map((reference, index) => ({
+                        reference,
+                        name: reference,
+                        yIndex: index,
+                    })),
+                };
+            }
+        });
+        builder.addCase(setSaveChartData, (state, action) => {
```
joaoviana, line 68: nice 👍

packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
```diff
@@ -0,0 +1,77 @@
+import { ChartKind, type BarChartConfig } from '@lightdash/common';
+import type { PayloadAction } from '@reduxjs/toolkit';
+import { createSlice } from '@reduxjs/toolkit';
+import { setInitialResultsAndSeries, setSaveChartData } from './sqlRunnerSlice';
+
+const initialState: { config: BarChartConfig | undefined } = {
+    config: undefined,
+};
+
+export const barChartSlice = createSlice({
```
joaoviana, line 10:
```suggestion
export const barChartConfigSlice = createSlice({
```

joaoviana: And if you could rename the file to `barChartConfigSlice` so everything matches 👍

joaoviana (approved):
Some small requested changes 👍

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
joaoviana: I believe this is done

packages/frontend/src/features/sqlRunner/store/index.ts
magnew: WFT, merges today

PR #10801, opened 2024-07-23
feat(wip): sql chart transformations

owlas:
Closes: 

### Reviewer actions


packages/frontend/src/features/sqlRunner/components/visualizations/barChartBizLogic.ts
```diff
@@ -0,0 +1,200 @@
...
+        columnId: string;
+        aggregation: string;
+        sort: string | null;
+    }[];
+    colorBy: {
+        columnId: string;
+    };
+};
+
+const transform = (config: BarChartConfig, rows: ResultRow[]) => {};
```
joaoviana, line 170: can you clarify what `transform` would do? Would it use a current transformer, like BarChartDataTransformer?

packages/frontend/src/features/sqlRunner/components/visualizations/barChartBizLogic.ts
```diff
@@ -0,0 +1,200 @@
...
+        columnId: string;
+        timeGrain: string;
+        type: 'time' | 'categorical';
+    };
+    y: {
+        columnId: string;
+        aggregation: string;
+        sort: string | null;
+    }[];
+    colorBy: {
```
joaoviana, line 165: Could you provide a usecase for this `colorBy`?

packages/frontend/src/features/sqlRunner/components/visualizations/barChartBizLogic.ts
```diff
@@ -0,0 +1,200 @@
...
+
+// available series
+type seriesLayoutOptions = {
+    columnId: string;
+};
+const seriesLayoutOptions = (columns: SqlColumn[]): seriesLayoutOptions[] => {
+    let options: seriesLayoutOptions[] = [];
+    for (const column of columns) {
+        switch (column.type) {
+            case DimensionType.STRING:
```
joaoviana, line 142: do we need more types for series layout options besides `STRING` & `BOOLEAN`?

packages/frontend/src/features/sqlRunner/components/visualizations/barChartBizLogic.ts
```diff
@@ -0,0 +1,259 @@
+import { DimensionType, friendlyName, type ResultRow } from '@lightdash/common';
+import { getPivotedResults } from './usePivotedResults';
+
+type SqlColumn = {
+    id: string;
+    type: DimensionType;
```
joaoviana, line 6: Not sure how we can get the `type` from current results endpoint. 
It returns a `ResultRow[]`, so that doesn't include the type.
So either the endpoint returns the results **and** the types for each, or we have to find this information somewhere else 🤔

packages/common/src/viz/sqlRunner.ts
```diff
@@ -0,0 +1,238 @@
...
+type SqlRunnerResultsTransformerDeps = {
+    duckDBSqlFunction: DuckDBSqlFunction;
+    rows: RowData[];
+    columns: SqlColumn[];
+};
+
+export class SqlRunnerResultsTransformer2
+    implements ResultsTransformerBase<SqlTransformBarChartConfig>
+{
+    private readonly duckDBSqlFunction: DuckDBSqlFunction;
```
magnew, line 106: Do we think this will be a more generic resutlsTransformationFuntion? It could be duck DB or a warehouse call or just an in-memory function? Or if we think we have duckDB available everywhere, this is fine.

packages/common/src/viz/barChart.ts
```diff
@@ -0,0 +1,86 @@
+import { friendlyName } from '../types/field';
+import { type ResultsTransformerBase } from './base';
+
+export type BarChartStyling = {
```
magnew, line 4: At least as this is set up, it's not really styling at the moment. It seems like this could all be part of the general BarChartConfig instead of splitting this out, no? Unless there is a plan to make this more separate kinds of properties.

packages/frontend/src/features/sqlRunner/components/visualizations/barChartBizLogic.ts
magnew, line 1: Is this file unused? It does seem like this is just a hook-wrapper around the other files. We probably don't need the convertToRowData in here, or it should go in the transformer.

magnew (commented):
Cool. Some small questions. I'll try to build a table with a similar pattern and see if it tracks.

owlas:
Closing for #10837

Issue #10714, opened 2024-07-12
I want to be able to override the snowflake warehouse configuration in my dbt .yml files. 


TuringLovesDeathMetal:
### Description

In dbt, I can configure which warehouse a model is run on (in snowflake), like this:

``` 
config {
snowflake­_warehouse: 'MY_WAREHOUSE'
}
```

In the project configuration in Lightdash, you also specify which data warehouse to use for your Snowflake connection. 

But, the Lightdash warehouse is the default, and this gets overridden by the dbt configuration if you've set any. 

For example, if I have a model called `katies_model` with a config like this:

```
config {
snowflake­_warehouse: 'WRONG_WAREHOUSE'
}
```

and then in Lightdash, I set my data warehouse to be `CORRECT_WAREHOUSE`, `katies_model` will be run on `WRONG_WAREHOUSE` (the Lightdash warehouse setting is overwritten by the dbt config). 

This was intentional for some users, but it's causing errors for some other users. 

In the proejct configuration settings in Lightdash, there should be an option (a toggle?) under the `warehouse` which says `always use this warehouse` or something and will _**not get overridden**_ by the dbt model warehouse config. 

More info in Slack here: https://lightdash.slack.com/archives/C06TWBVBN9F/p1720740043667169

### Documentation

Add this info to the project configuration docs

cooldude6000:
Hey,
should i add the info you gave here ? : https://docs.lightdash.com/get-started/setup-lightdash/connect-project

rephus:
> In the proejct configuration settings in Lightdash, there should be an option (a toggle?) under the warehouse which says always use this warehouse or something and will not get overridden by the dbt model warehouse config.

Should this be an option in the UI, or the CLI ? or both

PR #10803, opened 2024-07-23
fix: Override snowflake warehouse

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10714

### Description:

Before:
Screenshot from 2024-07-23 14-23-43

After:
Screenshot from 2024-07-23 14-23-26
Screenshot from 2024-07-23 14-08-10

Screenshot from 2024-07-23 14-23-10

### Reviewer actions

### DIFF
--- a/docs/docs/get-started/setup-lightdash/connect-project.mdx
+++ b/docs/docs/get-started/setup-lightdash/connect-project.mdx
@@ -467,8 +467,11 @@ access to use this warehouse. You can see al list of warehouses available in the
   style={{ display: 'block', margin: '0 auto 20px auto' }}
 />
 
-If `Always use this warehouse` is set to yes, then we'll be using this `warehouse` to all queries, even if this is overrwritten
-by the dbt config using `snowflake­_warehouse`
+:::info
+
+If “Always use this warehouse” is set to yes, this warehouse will be used for all queries, even if the dbt configuration specifies a different warehouse using snowflake_warehouse.
+
+:::
 
 #### Schema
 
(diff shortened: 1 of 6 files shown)


docs/docs/get-started/setup-lightdash/connect-project.mdx
```diff
@@ -467,6 +467,9 @@ access to use this warehouse. You can see al list of warehouses available in the
   style={{ display: 'block', margin: '0 auto 20px auto' }}
 />
 
+If `Always use this warehouse` is set to yes, then we'll be using this `warehouse` to all queries, even if this is overrwritten
+by the dbt config using `snowflake­_warehouse`
```
joaoviana, line 471:
```suggestion
:::info

If “Always use this warehouse” is set to yes, this warehouse will be used for all queries, even if the dbt configuration specifies a different warehouse using snowflake_warehouse.

:::
```

packages/common/src/types/projects.ts
```diff
@@ -153,6 +153,7 @@ export type CreateSnowflakeCredentials = {
     accessUrl?: string;
     startOfWeek?: WeekDay | null;
     quotedIdentifiersIgnoreCase?: boolean;
+    override?: string;
```
joaoviana, line 156: we should run a `generate-api`

joaoviana (changes requested):
2 changes requested 💪

packages/common/src/types/projects.ts
rephus: I already did, I believe this endpoint is old and not using tsoa, so it doesn't affect the docs.

joaoviana: Interesting, I got a change in `routes`

rephus: ups, you were right, sorry. I don't know what happened, I was probably in a different branch. 

Just pushed

joaoviana (approved):
💪

Issue #10807, opened 2024-07-23
Charts created from within dashboards have a warning icon on `explore from here` but no error


TuringLovesDeathMetal:
### Description

There's a warning icon, but no information, and no error when you click on explore from here. 

<img width="358" alt="image" src="https://github.com/user-attachments/assets/04dac590-1823-45c9-82e1-027ffc7ecf21">

### Steps to Reproduce the Bug or Issue

1. open a dashboard with a chart created from within the dashboard
2. on the chart created from within the dashboard, click the three-dot menu
3. see that `explore from here` has a warning icon, but no issue. 

### Cloud or self-hosting

None

PR #10808, opened 2024-07-23
fix: warning icon should be displayed when user can not use custom d…

Shifan-Gu:
…imensions

Closes:   #10807

### Description:

image (2)

this warning icon is showing up when i'm opening the chart as admin and gone when i'm opening the chart as viewer, this logic should be flipped.

### Reviewer actions


rephus (approved):
Good catch, thanks for your contribution

PR #10820, opened 2024-07-24
chore: return column types on sql runner results job

joaoviana:
Closes: N/A

### Description:

Why we need this:
https://github.com/lightdash/lightdash/pull/10801#discussion_r1688383916
So that we can apply dimensionType-specific logic to each config layout option

- Adds `columns` to sql runner job task
- Improves types and allows `logWrapper` to accept a generic (it will help with different response types for different jobs)
- Add `columns` on `select` of `useSqlQueryRun`
- Updated the transformers to reference `ResultRow[]` instead of the return type of the hook (initially I thought the hook type would be beneficial, but it's not that reliable)

Screenshot:

Screenshot 2024-07-24 at 11 05 29

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -103,8 +103,6 @@ export const ContentPanel: FC<Props> = ({
         },
     });
 
-    console.log(queryResults);
-
     // Run query on cmd + enter
     useHotkeys([
         [
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -94,14 +94,16 @@ export const ContentPanel: FC<Props> = ({
         isLoading,
     } = useSqlQueryRun({
         onSuccess: (data) => {
-            if (data) {
-                dispatch(setInitialResultsAndSeries(data));
+            if (data?.results) {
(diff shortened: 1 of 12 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -93,15 +93,17 @@ export const ContentPanel: FC<Props> = ({
         isLoading,
     } = useSqlQueryRun({
         onSuccess: (data) => {
-            if (data) {
-                dispatch(setInitialResultsAndSeries(data));
+            if (data?.results) {
+                dispatch(setInitialResultsAndSeries(data.results));
```
joaoviana, line 97: In my next PR I want to pass the whole data or dispatch 2 actions
one for setting initial results and series
another for setting the xLayoutOptions and yLayoutOptions (which are defined here: https://github.com/lightdash/lightdash/pull/10801/files) 
I believe this should be separate with the new slice you're creating @magnew

magnew: Makes sense. Although I think we figured out this afternoon that you're gonna want to add extraReducers instead of two dispatches. But you knew that :D

magnew (approved):
Nice

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
joaoviana: 100% - let's go with `extraReducers` 💪

Issue #10580, opened 2024-07-01
Filter value autocomplete not showing up for additional dimensions


jpetey75:
### Description

The filter value autocomplete did not show anything when trying to add a filter to a chart, specifically for additional dimensions.

![CleanShot 2024-07-01 at 12 58 38@2x](https://github.com/lightdash/lightdash/assets/5532776/2ee1ff3e-ba33-4fcd-8c3f-15dcfd25403b)

### Steps to Reproduce the Bug or Issue

1. Create a dimension that is defined in SQL (example `.yml` below, the `item_profit_margin_tier` dimension is the one that causes the bug)
2. Create a chart or dashboard with the explore that includes that new dimension
3. Try to add a filter using the SQL-defined dimension
4. No values appear as options in filter

```yml
  - name: item_profit
    description: 'Profit made for the individual item.'
    meta:
      dimension:
        format: 'usd'
        group: ['product_details', 'item_details']
      additional_dimensions:
        item_profit_margin:
          type: number
          description: Item profit divided by item price
          sql: ${item_profit} / ${item_price}
        item_profit_margin_tier:
          type: string
          description: True if item has profit margin greater than 12%
          sql: |
            case 
              when ${item_profit_margin} < .08 then 'Low profit margin' 
              when ${item_profit_margin} > .12 then 'High profit margin'
              else 'Mid profit margin' end
```

This code is added to our internal demo site with this PR: https://github.com/lightdash/lightdash-demo-data-gardening/pull/13

### Cloud or self-hosting

None

owlas:
Is this also happening for regular dimensions with the `sql` specified?

jpetey75:
I just tested and it looks like the filter autocomplete does work for regular dimension defined with a `sql` option @owlas

neo-angie-uesato:
Do you guys have an ETA for this issue?

rephus:
I can replicate, we get an error when trying to get results from these fields 

![image](https://github.com/user-attachments/assets/1795e638-3da9-4c49-b409-1978eae66e10)

>         "message": "Model \"events\" has a metric reference: ${event_id} which matches no metric",

rephus:
This also affects dimensions with custom sql like: 

```
      - name: another_dimension
        description: ""
        meta:
          dimension:
            type: string
            sql: ${event_id}  +1
 ```

Issue #8912, opened 2024-02-13
dimensions defined using custom SQL should have suggested filter values


TuringLovesDeathMetal:
### Description

Right now, if I have a dimension that's defined using custom SQL or any additional dimensions, then I dont' get any suggestions in the filter autocomplete. 

e.g. 
```yaml
name: status
  meta:
    dimension:
      sql: IF(${TABLE}.status = TRUE, 'completed', 'in progress')
```

The `status` field in Lightdash would have no autocomplete suggestions as a filter (since it's not just querying the data in the underlying table, it's applying some custom SQL to it).

PR #10823, opened 2024-07-24
fix: fix distinct value search on dimensions with custom sql

rephus:
Closes: #10580

### Description:

Screenshot from 2024-07-24 15-40-34
Screenshot from 2024-07-24 16-08-10

### Reviewer actions

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.ts
+++ b/packages/common/src/compiler/exploreCompiler.ts
@@ -654,7 +654,10 @@ export class ExploreCompiler {
         const { refTable, refName } = getParsedReference(ref, currentTable);
 
         const referencedMetric = tables[refTable]?.metrics[refName];
-        if (referencedMetric === undefined) {
+        // This refName can reference a dimension when using custom SQL in DBT and searching for distinct values on searchFieldUniqueValues
+        const referencedDimension = tables[refTable]?.dimensions[refName];
+        const referencedField = referencedMetric || referencedDimension;
+        if (referencedField === undefined) {
             throw new CompileError(
                 `Model "${currentTable}" has a metric reference: \${${ref}} which matches no metric`,
                 {},
@@ -663,7 +666,7 @@ export class ExploreCompiler {
 
         // eslint-disable-next-line @typescript-eslint/no-use-before-define
         const compiledMetric = this.compileMetricSql(
-            referencedMetric,
+            referencedField,
             tables,
             joinAliases,
         );
--- a/examples/full-jaffle-shop-demo/dbt/models/events.yml
+++ b/examples/full-jaffle-shop-demo/dbt/models/events.yml
@@ -69,6 +69,15 @@ models:
           dimension:
             type: number
             groups: ['events']
+          additional_dimensions: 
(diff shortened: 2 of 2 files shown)


packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -654,7 +654,10 @@ export class ExploreCompiler {
         const { refTable, refName } = getParsedReference(ref, currentTable);
 
         const referencedMetric = tables[refTable]?.metrics[refName];
-        if (referencedMetric === undefined) {
+        // This refName can reference a dimension when using custom SQL in DBT and searching for distinct values on searchFieldUniqueValues
+        const referencedDimension = tables[refTable]?.dimensions[refName];
+        const referencedField = referencedMetric || referencedDimension;
+        if (referencedField === undefined) {
```
ZeRego, line 660: `compileMetricReference` should not compile a dimension reference. 
This bug originates because custom metrics were the only way to inject custom SQL into a queryMetric and the `field.sql` would most likely return the "raw" column. This means the bug also happens for dimensions with custom SQL with references. https://github.com/lightdash/lightdash/issues/8912

Since we have custom dimensions now, we can change the metric query defined in the project service to have something like:
```
        const distinctField: CustomSqlDimension = {
            id: `${field.name}_distinct`,
            type: CustomDimensionType.SQL,
            dimensionType: isDimension(field) ? field.type : DimensionType.STRING,
            name: `${field.name}_distinct`,
            table: field.table,
            sql: `DISTINCT ${field.sql}`,
        };
```

ZeRego (changes requested):

packages/common/src/compiler/exploreCompiler.ts
rephus: yeah, that makes sense. I'll try using custom sql dimensions

Is there any reason (security?) why we build a fake metricQuery, rather than running the `select distinct from` directly against the warehouse ?

rephus: An issue about using custom dimensions is that the distinct doesn't like the extra () around it. 

This doesn't work: 

Screenshot from 2024-07-25 08-46-59

This works: 

image

rephus: I could run ' compileQuery' to ensure all permissions are in place (user attributes) and then run the distinct directly on the warehouse

rephus: Let me know what you think of the new approach

ZeRego: > Is there any reason (security?) why we build a fake metricQuery, rather than running the select distinct from directly against the warehouse ?

Was to make sure we generate the filters, joins, etc correctly.

rephus:
It even works for week names with sorting applied 

Screenshot from 2024-07-25 17-39-43
image

ZeRego (approved):

rephus:
This error is not related with this PR

image

I know what the issue is though, so I'll fix it on a separate PR

Issue #10773, opened 2024-07-18
Add `context` and `chart_id` to all `query_executed` events 


TuringLovesDeathMetal:
### Description

Right now, we have some pages where `query_executed` event is triggered, but we don't save the context of where the query executed event is coming from. 

Specifically, we're missing information about:

- SQL runner (does a sql runner query trigger the same `query_executed` event?)
- scheduled charts
- scheduled dashboards
- AI co-pilot queries
- automatic dashboard refresh (in case the query is triggered from the automatic dashboard refresh - is this possible to differentiate from `dashboardView` that we have today?)
- API-triggered query executions (are there multiple ways to do this? Do we need a secondary field wiht more context of which API call is executing this? Or is that overkill?)

We also have information about the project, dashboard, etc. but not about `chart_id`. 

Can we add `chart_id` as a parameter as well? And this value can be NULL (e.g. the query is not coming from a saved chart)

### Acceptance criteria:

- [ ] every query executed event has `context`
  - [ ] we might need to break this out into smaller tickets - e.g. add scheduled delivery to context
- [ ] query executed events for specific charts has a `chart_id` parameter which is filled

jpetey75:
What about API-triggered query executions @TuringLovesDeathMetal ? I assume there are multiple ways to trigger a query executed event via API

TuringLovesDeathMetal:
^yep! Good call!

rephus:
> SQL runner (does a sql runner query trigger the same query_executed event?)

SQL runner queries trigger the `lightdash_server.sql.executed` event, like this old sql runner: 

![image](https://github.com/user-attachments/assets/0801a71f-3ad9-496b-aff6-616f1a0e819f)

On the new sql runner we added `usingStreaming": true`

![image](https://github.com/user-attachments/assets/e9025d54-c321-4d65-851a-b21a199e89fb)

rephus:
I've noticed all scheduled deliveries where sending the wrong `GSHEET` context

PR #10834, opened 2024-07-25
chore: add api context analytic events

rephus:
Partially Closes: https://github.com/lightdash/lightdash/issues/10773

### Description:

This PR aims to add an `API / CLI` context to query.executed requests for most `chart results` endpoints. 

### Reviewer actions


rephus:
Once this is done, I can finally add the code to the ai-copilot

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -165,8 +168,23 @@ export enum QueryExecutionContext {
...
+export const getContextFromHeader = (req: Request) => {
+    const method = getRequestMethod(req.header(LightdashRequestMethodHeader));
+    switch (method) {
+        case RequestMethod.CLI:
+        case RequestMethod.CLI_CI:
+            return QueryExecutionContext.CLI;
+        case RequestMethod.UNKNOWN:
+            return QueryExecutionContext.API;
+        default:
+            return undefined;
```
IrakliJani, line 184: I'm wondering if this is actually the correct way to distinguish requests.

I don't think we usually set `UNKNOWN` as the request method. I've seen a couple of cases in the code, but I don't quite understand the reason behind it. I believe the `default:` use case is more suitable for API calls.

You can probably check for PAT as well. That, and having no request method would indicate an API call.

IrakliJani (commented):
added one comment which might be a improvement suggestion

packages/backend/src/analytics/LightdashAnalytics.ts
rephus: ok, I've just tried the following: 

- I made a request to get results from explore: that made a request with `method WEB_APP`

image

- I copied the same Request as CURL; and triggered the same request without modifying, using the same header, it obviously uses the same `method WEB_APP`

- Then I removed the `  -H 'Lightdash-Request-Method: WEB_APP' ` header, this method was `method UNKNOWN`


So for me it seems fair to use `UNKNOWN` to identify API requests. 
But if you are still not sure, or you think this can lead to false positives,  we can close this PR and tell @TuringLovesDeathMetal  it is not possible to reliably identify external `API` requests.

IrakliJani: oh I see, I think that covers it then. approving the PR

IrakliJani (approved):

PR #10836, opened 2024-07-25
fix: viewers can see saved SQL chart 

ZeRego:
Closes: 

### Description:

Problem: Only users that could manage CustomSQL could view the saved charts. 

Changes:

- Amend permission check when requesting results 
- Reuse permission check logic

Note: Not updating e2e tests in this PR. We need to seed a SQL chart to simplify e2e tests and cover these scenarios. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/sqlRunnerController.ts
+++ b/packages/backend/src/controllers/sqlRunnerController.ts
@@ -111,6 +111,7 @@ export class SqlRunnerController extends BaseController {
     /**
      * Get results from a file stored locally
      * @param fileId the fileId for the file
+     * @param projectUuid the uuid for the project
      * @param req express request
      */
     @Middlewares([allowApiKeyAuthentication, isAuthenticated])
@@ -206,20 +207,11 @@ export class SqlRunnerController extends BaseController {
     ): Promise<ApiSqlChartWithResults> {
         this.setStatus(200);
 
-        const chart = await this.services
-            .getSavedSqlService()
-            .getSqlChart(req.user!, projectUuid, uuid);
-
         return {
             status: 'ok',
-            results: {
-                jobId: (
-                    await this.services
-                        .getProjectService()
-                        .scheduleSqlJob(req.user!, projectUuid, chart.sql)
-                ).jobId,
-                chart,
-            },
+            results: await this.services
+                .getSavedSqlService()
+                .getChartWithResultJob(req.user!, projectUuid, uuid),
         };
     }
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/controllers/sqlRunnerController.ts
```diff
@@ -206,20 +207,11 @@ export class SqlRunnerController extends BaseController {
...
-                jobId: (
-                    await this.services
-                        .getProjectService()
-                        .scheduleSqlJob(req.user!, projectUuid, chart.sql)
-                ).jobId,
-                chart,
-            },
+            results: await this.services
+                .getSavedSqlService()
+                .getChartWithResultJob(req.user!, projectUuid, uuid),
```
ZeRego, line 214: Move this endpoint code to the service so we can have correct permission check.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2067,8 +2067,8 @@ export class ProjectService extends BaseService {
         );
         if (
             user.ability.cannot(
-                'manage',
-                subject('SqlRunner', { organizationUuid, projectUuid }),
+                'view',
+                subject('Project', { organizationUuid, projectUuid }),
```
ZeRego, line 2071: Allow viewers(and other roles) to get results.

packages/backend/src/services/SavedSqlService/SavedSqlService.ts
```diff
@@ -29,12 +31,53 @@ export class SavedSqlService extends BaseService {
...
     constructor(args: SavedSqlServiceArguments) {
         super();
         this.analytics = args.analytics;
         this.projectModel = args.projectModel;
         this.spaceModel = args.spaceModel;
         this.savedSqlModel = args.savedSqlModel;
+        this.schedulerClient = args.schedulerClient;
+    }
+
+    private async hasAccess(
```
ZeRego, line 45: Move duplicated code to private methods + Update all methods to use `hasSavedChartAccess` or `hasAccess`.

rephus (approved):
Code look good, I tested this, it works as expected

I also noticed a couple of unrelated bugs, perhaps we want to create tickest for these: 

- If I click on a sql in a dashboard, it takes me to the `sql-runner-new` but with a `uuid` instead of a slug, so it doesn't load correctly. 
- If I open this sql chart as a viewer, and try to run it, I get a backend error (403) as expected, but no feedback on the frontend.

PR #10837, opened 2024-07-25
feat: use duckdb to query sql runner and improve bar chart config

joaoviana:
Cherry-picked this commit: https://github.com/lightdash/lightdash/pull/10801/commits/589bdb585a92986a18b79c2545ead29ce720ed4f

Closes: N/A

Description: 

- Applies new types for config and styles/display of a chart
- Introduces new form to apply changes to a chart
- Can change aggregation of a y axis
- Allows for y and x field labels to change

> [!IMPORTANT]
> There are many UX changes that are not considered here, for example: users can have `none` aggregation; there's smart merges of config when things change on the results/user sql. This should be addressed separately on more PRs. For example, users can't add series to the chart **yet**


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -94,8 +94,8 @@ export const ContentPanel: FC<Props> = ({
         isLoading,
     } = useSqlQueryRun({
         onSuccess: (data) => {
-            if (data?.results) {
```
magnew, line 97: I might have changed this type this morning. Just FYI. So it *might* need a type guard or we can change how this works.

packages/frontend/src/features/sqlRunner/components/visualizations/BarChartConfiguration.tsx
```diff
@@ -1,65 +1,414 @@
...
+import { TableFieldIcon } from '../TableFields';
+import { BarChartAggregationConfig } from './BarChartAggregationConfig';
 
 const DEBOUNCE_TIME = 500;
 
-const BarChartConfiguration: FC = ({}) => {
+type Axes = SqlTransformBarChartConfig;
+
+// TODO: add y field feature
+// const YFieldAxisConfigForm: FC<{
```
magnew, line 30: Still need this commented section?

magnew (approved):
🎉 Very cool. 

Approved. I didn't ask for a lot of changes -- we've all been looking at this

packages/frontend/src/features/sqlRunner/components/visualizations/BarChartConfiguration.tsx
joaoviana: Going to remove the code but keep the todo, sounds good?

magnew: You can keep the code there too if it helps you. I was just checking

Issue #10839, opened 2024-07-25
chore: fix timezone e2e tests 


rephus:
There is something wrong in the timezone tests, fix or disable the failing tests so we can keep relying on those.

PR #10835, opened 2024-07-25
fix: space e2e test

rephus:
Closes: 

### Description:

What the issue is:

We have request happening in the modal to get the spaces, when the request finishes, the input is reset. 

On render the request takes longer than expected, and the typing is quick, which means some parts of the text gets deleted. 

See video: 

[Screencast from 25-07-24 16:48:33.webm](https://github.com/user-attachments/assets/858937e3-f6f5-4a3c-8493-0b95171d6b32)

This PR fixes this issue by waiting for some text to appear before typing.

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #10843, opened 2024-07-26
chore: fix timezone tests

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10839

### Description:

What was happening: 
The explore page does a reload (posthog?) so some of the actions were lost (adding a filter) 

How I fixed it: 

Wait until some of the content is loaded before doing any action. 

image

### Reviewer actions

### DIFF
--- a/packages/e2e/cypress/e2e/app/dates.cy.ts
+++ b/packages/e2e/cypress/e2e/app/dates.cy.ts
@@ -390,7 +390,7 @@ describe('Date tests', () => {
         cy.get('.tabler-icon-x').click({ multiple: true });
     });
 
-    it('Should keep value when changing date operator', () => {
+    it.only('Should keep value when changing date operator', () => {
         const todayDate = new Date();
 
         cy.visit(`/projects/${SEED_PROJECT.project_uuid}/tables/customers`);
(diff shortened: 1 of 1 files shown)


packages/e2e/cypress/e2e/app/dates.cy.ts
```diff
@@ -390,7 +390,7 @@ describe('Date tests', () => {
         cy.get('.tabler-icon-x').click({ multiple: true });
     });
 
-    it('Should keep value when changing date operator', () => {
+    it.only('Should keep value when changing date operator', () => {
```
magnew, line 393: only

magnew (approved):
Looks good aside from that 'only'. The tz tests passed on this review and another one failed :D

packages/e2e/cypress/e2e/app/dates.cy.ts
rephus: good catch, I always forget about reverting those...

rephus: The other e2e test is fixed here https://github.com/lightdash/lightdash/pull/10835 feel free to review that one too if you have time

rephus:
yeah, finally 
image

Issue #10854, opened 2024-07-26
Improve defaults for bar chart


owlas:
In order of preference

x = first string field -> first boolean field -> first date field -> first number field
y = sum first number field -> count first string field -> count first boolean field -> fail

joaoviana:
Similar to this ticket @owlas : [#10869](https://github.com/lightdash/lightdash/issues/10869)

Issue #10862, opened 2024-07-26
I can see if my chart needs extra configuration to get displayed


joaoviana:
### Description

Acceptance Criteria
- [ ] When I create a new chart, I see it has 2 defaults applies - one for the x axis and another for the y axis
- [ ] I can clear the x axis and the y axis
- [ ] If one or the other is completely empty: (x axis not defined, or y axis is an a empty array), then I should see an empty state, explaining why - "You're missing an X axis" / "You're missing a Y axis"
- [ ] If one column, that was previously referenced, isn't there anymore, the chart view could result into an error message like: "`column` does not exist. Choose another" as well as a message on the `Select` component (see `error` prop)

### What problem does this solve?

Clearer feedback to the user

PR #10891, opened 2024-07-29
fix: improve bar chart defaults

owlas:
Closes: #10854

## After

Resets configuration to default if the state matches the previous chart default

https://github.com/user-attachments/assets/aeb2ef24-1464-421c-96ba-89be4c9f7e96


joaoviana (approved):
Tested ✅ 

- the handling done in `defaultBarChartLayout` (returning undefined) will be useful to tackle this ticket: https://github.com/lightdash/lightdash/issues/10862

PR #10895, opened 2024-07-30
chore: remove old bar chart transformer

magnew:
### Description:

What was SqlRunnerResultsTransformer was very specific to bar chart and the old BarChartDataTransformer was just a thin wrapper around it. This removed the old BarChartDataTransformer and moves its functionality to the results transformer which gets renamed to BarChartResultsTransformer. 

This is in anticipation of adding a PieChartResultsTransformer and others. 

### Reviewer actions


magnew:
Discussed live and we aren't going to do this for now

PR #10899, opened 2024-07-30
feat: add pivot to bar charts

owlas:
Closes: #10769

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/BarChartFieldConfiguration.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/BarChartFieldConfiguration.tsx
@@ -191,10 +191,9 @@ export const BarChartFieldConfiguration = () => {
     const series = useAppSelector(
         (state) => state.barChartConfig.config?.display?.series,
     );
-    const groupByFields = useAppSelector(
-        (state) => state.barChartConfig.config?.fieldConfig?.groupBy,
+    const groupByField = useAppSelector(
+        (state) => state.barChartConfig.config?.fieldConfig?.groupBy?.[0],
     );
-    const groupByField = groupByFields?.[0];
     const groupByLayoutOptions = useAppSelector(
         (state) => state.barChartConfig.options.groupByOptions,
     );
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
```diff
@@ -53,6 +56,23 @@ export const barChartConfigSlice = createSlice({
...
+                reference: string;
+            }>,
+        ) => {
+            if (config?.fieldConfig) {
+                config.fieldConfig.groupBy = [
+                    { reference: action.payload.reference },
+                ];
+            }
+        },
+        unsetGroupByReference: ({ config }) => {
```
magnew, line 71: It seems like you did this very deliberately, but I'll ask: why have the unset specifically instead of passing undefined to setGroupByReference?

packages/frontend/src/features/sqlRunner/components/BarChartFieldConfiguration.tsx
```diff
@@ -149,6 +191,13 @@ export const BarChartFieldConfiguration = () => {
     const series = useAppSelector(
         (state) => state.barChartConfig.config?.display?.series,
     );
+    const groupByFields = useAppSelector(
+        (state) => state.barChartConfig.config?.fieldConfig?.groupBy,
+    );
+    const groupByField = groupByFields?.[0];
```
magnew, line 197: Could you just select this directly or do you need groupByFields for something else?

magnew (approved):
Looks good. I left a couple non-blocking questions.

packages/frontend/src/features/sqlRunner/components/BarChartFieldConfiguration.tsx
owlas: Fixed!

packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
owlas: To me it was more explicit but I don't feel strongly

magnew: I guess I don't either 😄. I was wondering if there was some gotcha, but if its jsut for readability, I'm cool with that.

Issue #10884, opened 2024-07-29
Table calculation modal should have the buttons for `save` and `cancel` appear in the modal (without scrolling)


TuringLovesDeathMetal:
### Description

Today, when I try to create/update a table calculation, the `save` and `cancel` buttons are hidden. I need to scroll to see them. 

These buttons should appear in the modal, by default. This may require us to shrink the SQL component. 

### Steps to Reproduce the Bug or Issue

![image](https://github.com/user-attachments/assets/97d82e55-9a7c-4582-873c-5785d3033923)
![image](https://github.com/user-attachments/assets/3816fb2d-07fb-4d41-b427-2604bad63623)

### Cloud or self-hosting

None

aayushi1995:
Hello @ZeRego, 
I want to contribute to this issue.
Could you please assign this issue to me?

PR #10901, opened 2024-07-30
fix: Table Calculation Modal (save and cancel) buttons

aayushi1995:
Closes: #10884 

### Description:
Hello,
Today, When we open the Table Calculation Modal, save and cancel button appears only when scrolled.
I added a wrapper element that has `position:sticky` as seen in the file:
packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
Additionally, there is one change done in the file:  
packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
The ace editor's left grey bar was overlapping on the Modals footer (Element with sticky position) as we scrolled down. So I had to adjust the z-index here in SQLForm.tsx

Current UI/UX:
Screenshot 2024-07-30 at 5 25 02 PM

New Behaviour:
Screenshot 2024-07-30 at 5 25 38 PM

cc: @TuringLovesDeathMetal @ZeRego 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
@@ -158,6 +158,9 @@ const TableCalculationModal: FC<Props> = ({
                 body: {
                     paddingBottom: 0,
                 },
+                content: {
+                    maxHeight: '70vh !important',
+                },
             }}
             fullScreen={isFullscreen}
         >
--- a/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
@@ -155,6 +155,9 @@ const TableCalculationModal: FC<Props> = ({
                     fontSize: theme.fontSizes.md,
                     fontWeight: 700,
                 },
+                body: {
+                    paddingBottom: 0,
+                },
             }}
             fullScreen={isFullscreen}
         >
@@ -224,7 +227,22 @@ const TableCalculationModal: FC<Props> = ({
                             data={Object.values(TableCalculationType)}
                         ></Select>
                     </Tooltip>
-                    <Group position="apart">
+                </Stack>
+
+                <Stack
+                    sx={() => ({
+                        position: 'sticky',
+                        backgroundColor: 'white',
+                        bottom: 0,
+                        zIndex: 2,
+                    })}
+                >
+                    <Group
+                        position="apart"
+                        sx={() => ({
+                            padding: theme.spacing.md,
+                        })}
+                    >
                         <ActionIcon
                             variant="outline"
                             onClick={toggleFullscreen}
--- a/packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
@@ -56,7 +56,7 @@ export const SqlForm: FC<Props> = ({ form, isFullScreen }) => {
 
     return (
         <>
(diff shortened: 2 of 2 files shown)


packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
```diff
@@ -224,7 +227,22 @@ const TableCalculationModal: FC<Props> = ({
...
+                    sx={() => ({
+                        position: 'sticky',
+                        backgroundColor: 'white',
+                        bottom: 0,
+                        zIndex: 2,
+                    })}
+                >
+                    <Group
+                        position="apart"
+                        sx={() => ({
```
rephus, line 242: most mantine components have most `style` properties as component props, so instead of dong `sx` yo ucould do `spacing={theme.spacing.md}`. afaik sx should only be used if the property is not directly available.

rephus (changes requested):
It is true that the buttons are now visible, but now the `type` dropdown is hidden behind he bottom, which is not ideal. I think we should try to display everything in the modal. We could try with the suggestion from the original ticket, and reduce the SQL textarea (normally you don't need more than 2 lines, and you can always go fullscreen) , what do you think ?

packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
aayushi1995: Noted.

aayushi1995:
Hi @rephus,
You can scroll in the modal and you'll be able to view the result type. 

But I acknowledge your comment, I can reduce the SQL textarea by 2 lines, However, if any new fields get added to the form in the future, the same issue will persist where buttons won't be visible. 

Making a Modal Footer (containing buttons) sticky at the bottom is a standard approach for better UI/UX hence I implemented it. Needless to say, I don't mind, removing the sticky behavior.

packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
aayushi1995: Hi @rephus,
<Group spacing={theme.spacing.md}/>
spacing attribute if for spacing between the elements inside the group wrapper. [Ref](https://v6.mantine.dev/core/group/?t=props)
I wanted to add padding around the element <Group/> element. Hence, I used sx, since I wasn't able to adding padding otherwise.
Thanks :)

rephus:
> Hi @rephus, You can scroll in the modal and you'll be able to view the result type.
> 
> But I acknowledge your comment, I can reduce the SQL textarea by 2 lines, However, you will still need to scroll and if any new fields get added to the form in the future, the same issue will persist where buttons won't be visible.
> 
> Making a Modal Footer (containing buttons) sticky at the bottom is a standard approach for better UI/UX hence I implemented it. Needless to say, I don't mind, removing the sticky behavior.

I think the sticky behaviour is a good solution, you can keep it, and I'd rather have the `type` field hidden than the save, as you suggested. But I think we can do it even better if we have everything visible without having to scroll.

> you will still need to scroll and if any new fields get added to the form in the future

You are right, that is expected, we'll have to redesign this modal if that happens again . That being said, we don't expect to add any more fields any time soon.

aayushi1995:
@rephus,
Surely, its better to have everything in visibility, changing the max height of Modal.content allows us to do so. 
Please refer to the image below for the updated Modal Design.  
Screenshot 2024-07-31 at 2 40 39 PM

rephus (approved):

rephus:
That looks good. Thanks for your contribution !

rephus:
@all-contributors add @aayushi1995  for code

aayushi1995:
Thank you. @rephus :)
It was my pleasure.

PR #10908, opened 2024-07-30
feat: change the position of bar y-axis

owlas:
https://github.com/user-attachments/assets/5fae1e6a-2be3-4dc8-8d31-f5ce44880d6c

### DIFF
--- a/packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
+++ b/packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
@@ -49,7 +49,7 @@ export type BarChartDisplay = {
     };
     yAxis?: {
         label?: string;
-        position?: 'left' | 'right';
+        position?: string;
     }[];
     series?: Record<string, { label?: string; yAxisIndex?: number }>;
     legend?: {
--- a/packages/frontend/src/features/sqlRunner/components/BarChartStyling.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/BarChartStyling.tsx
@@ -1,8 +1,11 @@
-import { XLayoutType } from '@lightdash/common';
-import { Stack, TextInput } from '@mantine/core';
+import { Select, Stack, TextInput } from '@mantine/core';
 import debounce from 'lodash/debounce';
 import { Config } from '../../../components/VisualizationConfigs/common/Config';
-import { setXAxisLabel, setYAxisLabel } from '../store/barChartSlice';
+import {
+    setXAxisLabel,
+    setYAxisLabel,
+    setYAxisPosition,
+} from '../store/barChartSlice';
 import { useAppDispatch, useAppSelector } from '../store/hooks';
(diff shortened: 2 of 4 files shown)


packages/frontend/src/features/sqlRunner/components/BarChartStyling.tsx
```diff
@@ -46,12 +55,25 @@ export const BarChartStyling = () => {
...
+                            onYAxisPositionChange(value || undefined)
+                        }
+                        placeholder={'Change the y-axis position'}
+                    />
+                </Config.Section>
+            </Config>
+            <Config>
+                <Config.Section>
+                    <Config.Heading>{`Series`}</Config.Heading>
+                    <TextInput radius="md" />
```
joaoviana, line 76: do you want to keep this here? The `series` bit

packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
```diff
@@ -49,7 +49,7 @@ export type BarChartDisplay = {
     };
     yAxis?: {
         label?: string;
-        position?: 'left' | 'right';
+        position?: string;
```
joaoviana, line 52: is this so we don't have an `echarts`-specific type?

joaoviana (approved):

packages/frontend/src/features/sqlRunner/components/BarChartStyling.tsx
```diff
@@ -1,8 +1,11 @@
-import { XLayoutType } from '@lightdash/common';
```
joaoviana, line 1: Needs this import back so `onXAxisLabelChange` can use it 👍

joaoviana (changes requested):

packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
owlas: It was unnecessarily strict, it meant having to type a lot of other places. I'm experimenting with more relaxed typing, this led to much less boilerplate ❤️

owlas: e.g. the value from a `Select` `onChange` is going to be a string, unless we do some type gymnastics

packages/frontend/src/features/sqlRunner/components/BarChartStyling.tsx
owlas: I'll remove this!

packages/frontend/src/features/sqlRunner/components/BarChartStyling.tsx
owlas: on it

owlas (commented):
making changes adn merging

joaoviana (approved):

PR #10913, opened 2024-07-30
chore: refactor sql runner redux state to have entire sql chart data

ZeRego:
Closes: 

### Description:

We need to store the entire saved SQL chart in the redux state as we need most (if not all) information for different parts of the page. The `savedSqlChart` property should be treated as "read only".

The properties that are editable(name, SQL, config) should also be stored as separate properties in the redux state as treated as the "unsaved/dirty" state.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
+++ b/packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
@@ -132,7 +132,7 @@ export const sqlRunnerSlice = createSlice({
                 state.activeSidebarTab = SidebarTabs.TABLES;
             }
         },
-        setSaveChartData: (state, action: PayloadAction<SqlChart>) => {
+        setSavedChartData: (state, action: PayloadAction<SqlChart>) => {
             state.savedSqlChart = action.payload;
             state.name = action.payload.name;
             state.description = action.payload.description || '';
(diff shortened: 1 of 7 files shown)


packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -24,23 +24,13 @@ export const DEFAULT_NAME = 'Untitled SQL Query';
...
     activeTable: string | undefined;
-    savedSqlUuid: string | undefined;
-    slug: string | undefined;
-    space:
-        | {
-              uuid: string;
-              name: string;
-          }
-        | undefined;
+    savedSqlChart: SqlChart | undefined;
```
joaoviana, line 27: can we then do this?
```suggestion
    readonly savedSqlChart: SqlChart | undefined;
```

joaoviana (approved):
One suggestion!

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
joaoviana: Ignore ^ it has to be set anyway

magnew (approved):
👍

PR #10915, opened 2024-07-30
chore: add pie chart to SQL runner

magnew:
Closes: 

### Description:

Here's the broad strokes of adding a the pie chart!

What's left to be done:
- ~~Config UI~~ -> Added a field picker, needs more config
- Clean up class hierarchy? I'm not super happy with how much this sits alongside the bar chart. We talked about this and this is what we decided to do today, but we might want to make these things share more
- Better/more data transformation. This works at the moment with the first dimension and first metric column. It could and should do more aggregation and handling data. 

Kapture 2024-07-30 at 19 03 55

### Reviewer actions


packages/frontend/src/features/sqlRunner/transformers/useBarChart.ts
```diff
@@ -1,48 +1,12 @@
...
-    return data.map((row) => {
-        return Object.fromEntries(
-            Object.entries(row).map(([key, value]) => {
-                return [key, value.value.raw];
-            }),
-        );
-    });
-};
-
-export class SqlRunnerResultsTransformerFE extends SqlRunnerResultsTransformer {
```
magnew, line 35: Moved to it's own file

packages/frontend/src/features/sqlRunner/transformers/usePieChart.ts
```diff
@@ -0,0 +1,36 @@
...
+    PieChartDataTransformer,
+    type PieChartSQLConfig,
+    type ResultRow,
+    type SqlColumn,
+} from '@lightdash/common';
+import { useMemo } from 'react';
+import { useAsync } from 'react-use';
+import { SqlRunnerResultsTransformerFE } from './SqlRunnerResultsTransformerFE';
+
+export const usePieChart = (
```
magnew, line 11: This is basically the same as useBarChart. We could probably be fine with one useSQLChart

packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
```diff
@@ -136,7 +159,11 @@ type SqlRunnerResultsTransformerDeps = {
 };
 
 export class SqlRunnerResultsTransformer
-    implements ResultsTransformerBase<SqlTransformBarChartConfig>
+    implements
+        ResultsTransformerBase<
+            SqlTransformBarChartConfig,
+            SqlTransformPieChartConfig
```
magnew, line 165: As discussed, I put the pie chart getters on this class, so it takes both generics. We could make sub classes if we don't like this. Or make the data types more shared? For example, instead of barChartXLayoutOptions, could it be dimensionOptions?

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -225,24 +230,48 @@ export const ContentPanel: FC = () => {
...
+                                            height: debouncedInputSectionHeight,
+                                            width: '100%',
+                                            flex: 1,
+                                        }}
+                                    />
+                                )}
+
+                            {queryResults?.results &&
+                                pieChartConfig &&
+                                selectedChartType === ChartKind.PIE && (
```
joaoviana, line 257: if you remove this `selectedChartType === ChartKind.PIE` you avoid the chart re-rendering every time you switch the type.
This can also be applied on the bar chart above ^

packages/frontend/src/features/sqlRunner/components/visualizations/PieChart.tsx
```diff
@@ -0,0 +1,47 @@
...
+                            width: 'auto',
+                            height: 'auto',
+                        }}
+                        style={style}
+                    />
+                )}
+            </>
+        );
+    },
+);
```
joaoviana, line 45: I think it's evident that we could have a shared `EchartsReact` component that uses one unified hook that takes a `ChartKind`

Do you think we can have a `useCartesianChart` that calls the content of `useBarChart` or `usePieChart` depending on the type?
For example

```
<EchartsVisualization data config style type/>
```

then inside this component we have everything as is (similar to this PieChart), but we call a hook, like: `useEchartsVisualization` that calls different transformers (or the same transformer that creates a different spec depending on the type)

packages/frontend/src/features/sqlRunner/transformers/usePieChart.ts
joaoviana: agree, got a similar comment here: https://github.com/lightdash/lightdash/pull/10915/files#r1698085158

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
magnew: Ah, because you have that conditional visibility? 👌

packages/frontend/src/features/sqlRunner/components/visualizations/PieChart.tsx
magnew: Yea, I think it makes sense. Let me see if I hit any snags making this more like that.

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
joaoviana: exactly 🙌

packages/frontend/src/features/sqlRunner/transformers/usePieChart.ts
magnew: Cool, added a ticket I will do next: https://github.com/lightdash/lightdash/issues/10926

PR #10929, opened 2024-07-31
feat: improved loading state for sql runner

owlas:
Closes: #10923 

### After

- keeps results while new results pending
- fixes loading state to watch for `isFetching` (`isLoading` is true from react query while data is undefined, even if not fetching)
- fixes loading state to include job status fetching (prevents blinking loading state)
- adds overlay to previous results on results fetching
- replaces bar chart loading state with same loading overlay

https://github.com/user-attachments/assets/415055ca-31c7-4c7e-a314-07e3a3e5cec9

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
@@ -27,7 +27,7 @@ const BarChart: FC<BarChartProps> = memo(
         return (
             <>
                 {!spec && <Skeleton h="100%" />}
-                <LoadingOverlay visible={loading} />
+                <LoadingOverlay visible={loading || !spec} />
                 {spec && (
                     <EChartsReact
                         option={spec}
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
```diff
@@ -27,10 +27,10 @@ const BarChart: FC<BarChartProps> = memo(
         return (
             <>
                 {!spec && <Skeleton h="100%" />}
+                <LoadingOverlay visible={loading} />
```
joaoviana, line 30:
```suggestion
                <LoadingOverlay visible={loading || !spec} />
```
since `error` is handled above ^

joaoviana (approved):
One small suggestion to remove the skeleton 

Testing locally quickly

packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
joaoviana: and we can remove the `Skeleton` ^

Issue #10931, opened 2024-07-31
Removing group by from SQL chart doesn't update series


ZeRego:
Add group by
Remove group by

Current:  
series still with pivot

Expected:
series without pivot

https://github.com/user-attachments/assets/d936e21a-7fff-42bb-b45b-d9f2c1d283b9

owlas:
Need `notMerge` in eCharts, doing this now

owlas:
Closed by #10938

PR #10938, opened 2024-07-31
fix: notMerge eCharts on barcharts

owlas:
Closses: https://github.com/lightdash/lightdash/issues/10931


packages/frontend/src/features/sqlRunner/components/visualizations/BarChart.tsx
```diff
@@ -30,6 +30,7 @@ const BarChart: FC<BarChartProps> = memo(
                 {spec && (
                     <EChartsReact
                         option={spec}
+                        notMerge
```
ZeRego, line 33: We prob need the same fix for the pie chart.

ZeRego (approved):

PR #10943, opened 2024-07-31
fix: use time axes on bar charts

owlas:
Closes: https://github.com/lightdash/lightdash/issues/10942

https://github.com/user-attachments/assets/9a7eb96b-fd33-4aa6-a9f5-fbdfe92971cd


magnew (approved):
Sweet

Issue #10819, opened 2024-07-24
When I'm in a dashboard, and edit a chart, then delete it, I should be redirected to the dashboard 


TuringLovesDeathMetal:
### Description

Right now, if I edit a chart that is saved to a space (i.e. _**not**_ a chart created from within a dashboard), then I delete the chart, I'm not redirected back to the dashboard I was working in. 

This is really annoying. 

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/0ff970d4fc1447dbb846b96e31b27e22

### Cloud or self-hosting

None

PR #10951, opened 2024-07-31
fix: If a chart is deleted, It should be redirected to the dashboard

anshul45:
Closes: (https://github.com/lightdash/lightdash/issues/10819)

### Description:
If we are editing chats from any dashboard, the history will include parameters like fromDashboard={dashboardId}. If we are editing chats from space, it will be null. I check if the search parameter is not null; 
If it isn’t, I redirect to the dashboard, Otherwise I redirect to the home page.

https://www.loom.com/share/2c58db6dba4647deaf8bb95685f40b1e?sid=a5f8d6a7-6f18-4caa-8c97-f16d4a82391d

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -976,7 +976,13 @@ const SavedChartsHeader: FC<SavedChartsHeaderProps> = ({
                             }
                         });
 
-                        history.push('/');
+                        if (search) {
+                            history.push(
+                                `/projects/${projectUuid}/dashboards/${dashboardUuid}`,
+                            );
+                        } else {
+                            history.push(`/`);
+                        }
 
                         deleteModalHandlers.close();
                     }}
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
```diff
@@ -976,7 +976,13 @@ const SavedChartsHeader: FC<SavedChartsHeaderProps> = ({
                             }
                         });
 
-                        history.push('/');
+                        if (search) {
```
rephus, line 979: Should we check `dashboardUuid` instead of `search` here ? 

For example, if the chart has another search (fromSpace), and I delete a chart, it throws an error because it takes me to `projects/3675b69e-8324-4110-bdca-059031aa8da3/dashboards/null` 

[Screencast from 06-08-24 11:15:36.webm](https://github.com/user-attachments/assets/ce1654aa-6661-4a77-b884-e11d94a849d3)

rephus (changes requested):

packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
anshul45: yes that's correct. We should check `dashboardUuid` because it redirecting to this `dashboardUuid` let me correct this.

IrakliJani:
tested here: https://github.com/lightdash/lightdash/pull/11084 works as expected. thanks for your contribution @anshul45 ! 🙌

IrakliJani (approved):

rephus (approved):

Issue #10936, opened 2024-07-31
Fix tile state when there is an error. It shows loading state instead of the error message.


ZeRego:
//packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx 

Error state should be handled before the loading state.

PR #10956, opened 2024-08-01
fix: dashboard SQL chart tile error state

ZeRego:
Closes: [#10936](https://github.com/lightdash/lightdash/issues/10936) 

### Description:

Changes:
- check `isFetched` and `isFetching` to catch when query is not enabled ( aka, savedSqlChartUuid is null )
- check if any of the queries returned an error
- update tile props 

Before: we would show the loading state if something was wrong

After: when saved chart no longer exists

Screenshot 2024-08-01 at 12 06 42

After: when there is an error fetching chart and results

Screenshot 2024-08-01 at 12 23 02

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx
@@ -1,7 +1,6 @@
 import {
     ChartKind,
     isTableChartSQLConfig,
-    type Dashboard,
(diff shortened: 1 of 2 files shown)


joaoviana (approved):
Works great! Got all states being rendered.

Only one thing to improve (later) is to reduce the overlay visibility when going from view to edit mode. 
I think this can be done on https://github.com/lightdash/lightdash/issues/10960

PR #10958, opened 2024-08-01
feat: allow styling bar chart series' labels

joaoviana:
Closes: 

### Description:

WIP:

https://github.com/user-attachments/assets/d630bfba-4ff7-4532-9475-a7f595086f3e

### Reviewer actions

### DIFF
--- a/packages/common/src/visualizations/BarChartDataTransformer.ts
+++ b/packages/common/src/visualizations/BarChartDataTransformer.ts
@@ -25,6 +25,31 @@ export class BarChartDataTransformer<TBarChartLayout, TPieChartConfig> {
 
         const DEFAULT_X_AXIS_TYPE = 'category';
 
+        const series = transformedData?.seriesColumns.map(
+            (seriesColumn, index) => {
+                const seriesLabel = Object.values(display?.series || {}).find(
+                    (s) => s.yAxisIndex === index,
+                )?.label;
+
+                return {
+                    dimensions: [
+                        transformedData.xAxisColumn.reference,
+                        seriesColumn,
+                    ],
+                    type: 'bar',
+                    name: seriesLabel || friendlyName(seriesColumn),
+                    encode: {
+                        x: transformedData.xAxisColumn.reference,
+                        y: seriesColumn,
+                    },
+                    yAxisIndex:
+                        (display?.series &&
+                            display.series[seriesColumn]?.yAxisIndex) ||
+                        0,
+                };
+            },
+        );
+
         return {
             tooltip: {},
             legend: {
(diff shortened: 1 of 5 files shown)


packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
```diff
@@ -195,6 +197,14 @@ export const barChartConfigSlice = createSlice({
...
 
             state.config.fieldConfig.y.splice(action.payload, 1);
+
+            // Remove series styling if only one y axis field left
+            if (
+                state.config.fieldConfig.y.length === 1 &&
+                state.config.display?.series
+            ) {
+                state.config.display.series = undefined;
+            }
```
joaoviana, line 207: does this make sense to you @owlas?
So when there's only one y axis, then there's no extra fields, so we can clear the series styling. LMK if you want this somewhere else

packages/common/src/visualizations/BarChartDataTransformer.ts
```diff
@@ -25,6 +25,31 @@ export class BarChartDataTransformer<TBarChartLayout, TPieChartConfig> {
 
         const DEFAULT_X_AXIS_TYPE = 'category';
 
+        const series = transformedData?.seriesColumns.map(
+            (seriesColumn, index) => {
+                const seriesLabel = Object.values(display?.series || {}).find(
```
joaoviana, line 30: this is where I'm a bit confused if I should do it this way @owlas 

We want to get the label, we know the series is an object where the key is a `reference` and the values include `yAxisIndex` 
Would you do this differently or potentially change the structure of series? 🤔

packages/common/src/visualizations/BarChartDataTransformer.ts
```diff
@@ -25,6 +25,31 @@ export class BarChartDataTransformer<TBarChartLayout, TPieChartConfig> {
...
+                    (s) => s.yAxisIndex === index,
+                )?.label;
+
+                return {
+                    dimensions: [
+                        transformedData.xAxisColumn.reference,
+                        seriesColumn,
+                    ],
+                    type: 'bar',
+                    name: seriesLabel || friendlyName(seriesColumn),
```
joaoviana, line 40: @owlas I'm not sure what to do with this `seriesColumn` because when there's no label applied, we default to this which is like "Sum of Customer id", but on the Styling panel we don't have access to that `seriesColumn`, so we default to reference, which will be something like `customer_id` - what would you do here? 
Would you keep this `friendlyName(seriesColumn)` and keep the input empty if no override on the styling panel?

joaoviana:
Closing due to lack of capacity

Issue #10945, opened 2024-07-31
Replace "sql.executed" with "query.executed" event


ZeRego:
slack thread: https://lightdash.slack.com/archives/C02FSS5HL2E/p1722434399383129?thread_ts=1722432671.482009&cid=C02FSS5HL2E

- [ ] replace 'sql.executed' with 'query.executed'
- [ ] update most of the properties to be optional as we won't have information about dim/metric/table calcs/etc
- [ ] apply context = 'sql_runner'

PR #10965, opened 2024-08-01
feat: endpoint to fetch SQL chart results from slug

ZeRego:
Closes: 

### Description:

Nothing changes in the UI.
The view page fetched the chart by slug and then the chart-and-results by uuid in sequence. Now we fetch the chart and the results independently and in parallel via the slug. 
They are handled independently so the page can load the chart details correctly independently of the results response.
This also helps setup the analytics events from [#10945](https://github.com/lightdash/lightdash/issues/10945)

### Reviewer actions


packages/frontend/src/features/sqlRunner/hooks/useSqlChartResults.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const job = await getCompleteJob();
+    const url =
+        job?.details && !isErrorDetails(job.details)
+            ? job.details.fileUrl
+            : undefined;
+    return getResultsStream(url);
+};
+
+export const useSqlQueryRun = (
```
joaoviana, line 137: It may be confusing to have two `useSqlQueryRun` hooks (on in `useSqlChartResults.tsx` and another in `useSqlQueryRun.tsx`)
Maybe this should be named `useSqlChartResults`?

packages/frontend/src/features/sqlRunner/hooks/useSqlChartResults.ts
```diff
@@ -0,0 +1,153 @@
...
+            ApiSqlRunnerJobStatusResponse['results']
+        >(scheduledJob.jobId);
+        if (
+            job.status === SchedulerJobStatus.SCHEDULED ||
+            job.status === SchedulerJobStatus.STARTED
+        ) {
+            return new Promise((resolve) => {
+                setTimeout(async () => {
+                    resolve(await getCompleteJob());
+                }, 3000);
```
joaoviana, line 109: we can decrease the timeout to something lower, like 1000?

joaoviana (approved):

Issue #10828, opened 2024-07-24
`sql_filter` that uses fields from other joined tables is not joining tables


TuringLovesDeathMetal:
### Description

I have a `sql_filter` which is using a field from a joined table. 

The joined table is not being included in my `from` bit of the SQL query. So, teh query errors when it's executed. 

This could be because it's using a bridge table to join it. Or, it could be an issue more generally wiht joined tables in the `sql_filter` we need to investigate!

### Steps to Reproduce the Bug or Issue

More context in this thread: https://lightdash.slack.com/archives/C071TU6FZ3P/p1721674501240779?thread_ts=1721674501.240779&cid=C071TU6FZ3P

### Cloud or self-hosting

None

IrakliJani:
workaround: https://docs.lightdash.com/references/joins/#always-join-a-table

ZeRego:
Note that we have logic to join all tables referenced in filters

Check `selectedTables` in `packages/backend/src/queryBuilder.ts`

PR #10967, opened 2024-08-01
fix: table `sql_filter` should join a referenced table

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/10828

test-frontend

### Description:

CleanShot 2024-08-01 at 22 25 05@2x
CleanShot 2024-08-01 at 22 24 41@2x

### Reviewer actions

### DIFF
--- a/packages/backend/src/queryBuilder.ts
+++ b/packages/backend/src/queryBuilder.ts
@@ -645,6 +645,32 @@ export const getCustomBinDimensionSql = ({
     return { ctes, joins, tables: [...new Set(tables)], selects };
 };
 
+const getJoinedTables = (explore: Explore, tableNames: string[]): string[] => {
+    if (tableNames.length === 0) {
+        return [];
+    }
+    const allNewReferences = explore.joinedTables.reduce<string[]>(
+        (sum, joinedTable) => {
+            if (tableNames.includes(joinedTable.table)) {
+                const newReferencesInJoin = parseAllReferences(
+                    joinedTable.sqlOn,
+                    joinedTable.table,
+                ).reduce<string[]>(
+                    (acc, { refTable }) =>
+                        !tableNames.includes(refTable)
+                            ? [...acc, refTable]
+                            : acc,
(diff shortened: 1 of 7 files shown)


packages/backend/src/queryBuilder.ts
```diff
@@ -807,34 +834,22 @@ export const buildQuery = ({
...
-                    return [...sum, ...newReferencesInJoin];
-                }
-                return sum;
-            },
-            [],
-        );
-        return [...allNewReferences, ...getJoinedTables(allNewReferences)];
-    };
+    const tableCompiledSqlWhere =
+        explore.tables[explore.baseTable].compiledSqlWhere;
```
IrakliJani, line 838: Since I'm using the new `compiledSqlWhere`, I realize this is a breaking change.

Before the change, we've been using `sqlWhere`, which is already compiled.

We need both the compiled and original versions to make the new filters work.

Something that I hadn't considered initially is that changes inside `exploreCompiler` happen at compile time, while `queryBuilder` changes occur at run time. So, whatever we end up compiling gets persisted in the DB.

Technically, the new solution adds the code to the right places, but since we only had the `sqlWhere` property, which is already a compiled version, it forces us to make a change that is not ideal to understand.

What we had before:
`sqlWhere` -> compiled SQL

Solution in the PR:
`sqlWhere` -> now contains uncompiled SQL (breaking change)
`compiledSqlWhere` -> contains compiled SQL

Proposed solution:
`sqlWhere` -> stays as compiled SQL
`uncompiledSqlWhere` -> will now hold uncompiled SQL

Let me know if you know any way to make this less confusing than it is.

almeidabbm: I agree with the proposed solution, seems to me like a good compromise to avoid a breaking change in the compiled tables 👌

almeidabbm (approved):

almeidabbm (approved):
LGTM 🚢

Issue #10914, opened 2024-07-30
Infinite popups during Google Sheets Export


IrakliJani:
### Description

When attempting to export data to Google Sheets, the system produces an infinite loop of pop-up dialogs.

this happens when exporting to google sheets from the explore in results table card

relevant thread with a video attached to it: https://lightdash.slack.com/archives/C069QQMERD2/p1722009132547149

### Steps to Reproduce the Bug or Issue

I have not been able to reproduce this.

### Cloud or self-hosting

None

PR #11008, opened 2024-08-05
fix: infinite google login popup

rephus:
Closes: https://github.com/lightdash/lightdash/issues/10914

### What the issue is 

At least on Firefox (popup blocker?) , the `window.close` api always returns true, which doesn't correspond what the documentation says (Related issue: https://stackoverflow.com/questions/5972164/window-open-returns-always-closed-in-true) . This was causing firefox to reopen inifnite popups. 

### Why we were checking for close ? 
This was mainly to `reopen` the modal automatically if someone closed the popup. Because once it was closed, it was not possible to reopen the modal again. 

###  How I fixed this instead 
WHen clicking on the `Select Google Sheet via Google drive`, I `force` close the modal, this opens a new modal. 

### Description:

Before:

[Screencast from 05-08-24 12:11:29.webm](https://github.com/user-attachments/assets/f04733f1-6f22-47c3-bdde-4a67bfde0514)

After:

[Screencast from 05-08-24 13:03:10.webm](https://github.com/user-attachments/assets/c887c9fb-0984-49c9-a77a-d8d3948ac81e)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sync/components/SelectGoogleSheetButton.tsx
+++ b/packages/frontend/src/features/sync/components/SelectGoogleSheetButton.tsx
@@ -49,7 +49,7 @@ export const SelectGoogleSheetButton: FC = () => {
         [methods],
     );
 
-    useGdriveAccessToken({
+    const { closePopup } = useGdriveAccessToken({
         enabled: isGoogleAuthQueryEnabled,
         onSuccess: (accessToken) => {
             if (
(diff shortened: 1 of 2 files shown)


rephus:
We should also confirm if this behaviour works for other browsers too, so do not test (only) firefox.

almeidabbm (approved):
Works well on both firefox and chrome!

Issue #10975, opened 2024-08-02
When I pick an x axis column that has all null values I get a into a bad visualization state


joaoviana:
### Description

We don't handle errors that well when a column  with only empty values is chosen as the x axis

The error gets displayed and it's hard for the user to know what to change

### Steps to Reproduce the Bug or Issue

1. Query x number of rows where a column has only null values
2. Select this column for the x axis 
3. You will get an error like this: Binder Error: Referenced column "YOUR_COLUMN" not found in FROM clause! Candidate bindings: "YOUR_COLUMN"

### Cloud or self-hosting

None

joaoviana:
`apache-arrow`'s `tableFromJSON` filters out columns that have `null` values, so it will not include them on the table, so when we query stuff it will throw the error. 

I have yet to find an alternative workaround.

Issue #10940, opened 2024-07-31
Add aggregation options for Pie Chart


magnew:
### Description

- [ ] Allow for metric selection for Pie chart
- [ ] Add aggregation options

PR #11009, opened 2024-08-05
fix: handle error when a column's rows are all null

joaoviana:
Closes: #10975

### Description:

`apache-arrow`'s `tableFromJSON` filters out columns that are all `null`. 
So when the user selects one of this type of columns, the duckdb query will fail because it's not part of the apache-arrow's table schema.

This PR: 
- checks which columns are null
- adds a `disabled` property to the xLayoutOptions so that the user knows they can't choose it
- alters the selection of the first candidate option for a default config: we filter out the disabled options and then carry on with the existing logic
- adds a shared component for Reference field selection with a custom `ItemComponent`

https://github.com/user-attachments/assets/9c68ace8-ec00-4052-bcc8-320e39b12cdc

### Reviewer actions

### DIFF
--- a/packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
+++ b/packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
@@ -22,6 +22,7 @@ export type AggregationOptions = typeof aggregationOptions[number];
 export type SqlColumn = {
     reference: string;
     type: DimensionType;
+    isNull?: boolean;
 };
 
 export enum XLayoutType {
@@ -32,15 +33,18 @@ export enum XLayoutType {
 export type XLayoutOptions = {
     type: XLayoutType;
     reference: string;
+    disabled: boolean;
 };
 
 export type YLayoutOptions = {
     reference: string;
     aggregationOptions: AggregationOptions[];
+    disabled: boolean;
 };
 
 export type GroupByLayoutOptions = {
     reference: string;
+    disabled: boolean;
 };
 
 export type CartesianChartDisplay = {
@@ -71,12 +75,7 @@ export type SqlTransformCartesianChartConfig = {
     groupBy: { reference: string }[] | undefined;
 };
 
-// TODO: Should pie chart share the X and Y layout options? They could be
-// called something else
-export type PieChartDimensionOptions = {
-    type: XLayoutType;
-    reference: string;
-};
+export type PieChartDimensionOptions = XLayoutOptions;
 
 export type PieChartMetricOptions = {
     reference: string;
@@ -182,6 +181,7 @@ export class SqlRunnerResultsTransformer
                 case DimensionType.BOOLEAN:
                     options.push({
                         reference: column.reference,
+                        disabled: !!column.isNull,
                     });
                     break;
                 default:
@@ -201,12 +201,14 @@ export class SqlRunnerResultsTransformer
                     options.push({
                         reference: column.reference,
                         type: XLayoutType.TIME,
+                        disabled: !!column.isNull,
                     });
                     break;
                 case DimensionType.TIMESTAMP:
                     options.push({
                         reference: column.reference,
                         type: XLayoutType.TIME,
+                        disabled: !!column.isNull,
                     });
                     break;
                 case DimensionType.STRING:
@@ -215,6 +217,7 @@ export class SqlRunnerResultsTransformer
                     options.push({
                         reference: column.reference,
                         type: XLayoutType.CATEGORY,
+                        disabled: !!column.isNull,
                     });
                     break;
                 default:
@@ -232,6 +235,7 @@ export class SqlRunnerResultsTransformer
                     options.push({
                         reference: column.reference,
                         aggregationOptions,
+                        disabled: !!column.isNull,
                     });
                     break;
                 case DimensionType.STRING:
@@ -243,6 +247,7 @@ export class SqlRunnerResultsTransformer
                                 option === MetricType.COUNT ||
                                 option === MetricType.COUNT_DISTINCT,
                         ),
+                        disabled: !!column.isNull,
                     });
                     break;
                 default:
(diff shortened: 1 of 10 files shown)


packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
```diff
@@ -349,18 +355,28 @@ export class SqlRunnerResultsTransformer
...
 
-        const groupFieldIds = firstCategoricalColumn?.reference
-            ? [firstCategoricalColumn.reference]
-            : undefined;
-        const metricId = firstNumericColumn?.reference;
+        let groupFieldIds: string[] | undefined;
+        if (firstCategoricalColumn?.reference) {
+            groupFieldIds = [firstCategoricalColumn.reference];
+        } else if (firstDateColumn?.reference) {
+            groupFieldIds = [firstDateColumn.reference];
```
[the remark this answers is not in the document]
ZeRego: yes. example: grouping by year/month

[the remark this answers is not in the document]
ZeRego: Even though possible to have that option, I wouldn't add a date column as the default group.

packages/frontend/src/features/sqlRunner/store/pieChartSlice.ts
```diff
@@ -35,9 +35,15 @@ export const pieChartConfigSlice = createSlice({
...
-                config.fieldConfig.groupFieldIds = [action.payload];
+            if (!config) return;
+
+            if (!config.fieldConfig) {
+                config.fieldConfig = {
+                    groupFieldIds: [action.payload],
+                };
             }
+
+            config.fieldConfig.groupFieldIds = [action.payload];
```
ZeRego, line 46: Since we don't use a `return` inside the `if` we should use a `else` here.

packages/frontend/src/features/sqlRunner/components/CartesianChartFieldConfiguration.tsx
```diff
@@ -221,18 +213,10 @@ const GroupByFieldAxisConfig = ({
...
-                    <TableFieldIcon
-                        fieldType={
-                            sqlColumns?.find(
-                                (x) => x.reference === field?.reference,
-                            )?.type ?? DimensionType.STRING
-                        }
-                    />
-                ) : null
+            fieldType={
+                sqlColumns?.find((x) => x.reference === field?.reference)
```
ZeRego, line 217: Why doesn't the option object have this information? Perhaps work for a separate PR.

packages/common/src/visualizations/SqlRunnerResultsTransformer.ts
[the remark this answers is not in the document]
joaoviana: Thanks, this makes sense! 

Since we can't **yet** add a metric: https://github.com/lightdash/lightdash/issues/10940 to a pie chart, I will then remove this default option, keep the select empty, but still allow it in case the user wants to select it

packages/frontend/src/features/sqlRunner/store/pieChartSlice.ts
joaoviana: looking!

packages/frontend/src/features/sqlRunner/components/CartesianChartFieldConfiguration.tsx
joaoviana: Absolutely. Adding a ticket: https://github.com/lightdash/lightdash/issues/11012

ZeRego (changes requested):
We have done a good job preventing/mitigating the error but we haven't handled the error. There are multiple ways to end up with the error. 

Example 1:

- Run query ( all columns have values )
- Edit chart x-axis to X
- Edit query ( column X only has null values )
- See chart error

Example 2:

- Run query ( all columns have values )
- Edit chart x-axis to X
- Save chart 
- After two days, I load the chart and column X only has null values
- See chart error

Screenshot 2024-08-05 at 14 42 23

packages/frontend/src/features/sqlRunner/store/pieChartSlice.ts
joaoviana: done here: https://github.com/lightdash/lightdash/pull/11009/commits/c153c6618947538ea57037dec7c3f46ccfa5f098

joaoviana:
> We have done a good job preventing/mitigating the error but we haven't handled the error. There are multiple ways to end up with the error.
> 
> Example 1:
> 
> * Run query ( all columns have values )
> * Edit chart x-axis to X
> * Edit query ( column X only has null values )
> * See chart error
> 
> Example 2:
> 
> * Run query ( all columns have values )
> * Edit chart x-axis to X
> * Save chart
> * After two days, I load the chart and column X only has null values
> * See chart error
> 
> Screenshot 2024-08-05 at 14 42 23

on it, thanks for sharing these examples!

Going to ingest the duckdb error and do some handling

packages/frontend/src/features/sqlRunner/duckDBQuery.ts
```diff
@@ -26,15 +26,24 @@ function initializeBundle() {
...
     const conn = await db.connect();
 
+    const columnReferences = columns.map((c) => c.reference);
+    const typedArrays = columnReferences.reduce((acc, c) => {
+        const col = rowData.map((r) => r[c]);
+        return {
+            ...acc,
+            [c]: col,
+        };
+    }, {});
```
joaoviana, line 44: by creating a typed array https://arrow.apache.org/docs/js/functions/Arrow_dom.tableFromArrays.html `null` all-null-values-tables won't be removed. 

I've kept the existing logic to disable these columns from being used - I believe this is fine. 

Regarding your point on handling the error @ZeRego, I believe this one is fixed - I tested example 1

joaoviana:
Removed `disabled` logic now that we can handle `null`s - perhaps we can keep track of columns that are all `null` in the future to have a smart default config candidate, but this can be thought about later

packages/frontend/src/features/sqlRunner/duckDBQuery.ts
ZeRego: Both example 1 and 2 are fixed. But I was able to reproduce the error with a different scenario:

1. Run query ( all columns have values )
2. Edit chart x-axis to X
3. Edit query to not return column X
4. See chart error

Screenshot 2024-08-06 at 10 08 49

joaoviana: will do that! 
I also tried to find an official error parser but no luck :/

joaoviana: Screenshot 2024-08-06 at 10 20 16

what do you think?

ZeRego (approved):

Issue #11007, opened 2024-08-05
Stop deleting orphan charts on dashboard save


IrakliJani:
### Description

Stop deleting orphan charts on dashboard save

requires both BE and FE changes

### Steps to Reproduce the Bug or Issue

-

### Cloud or self-hosting

None

PR #11016, opened 2024-08-05
fix: orphaned chart deletion

IrakliJani:
Closes: #11007 

### Description:

this PR fixes two things:
- exiting dashboard edit mode in the **frontend** is never going to delete orphan charts (or any charts)
- **backend** will only delete orphans if, and only if, they do not belong to any tile. This means that version reverting will now work for charts created within the dashboard, even if they get removed from the tile in the next dashboard version save.

### Reviewer actions


ZeRego (approved):

almeidabbm (approved):

almeidabbm:
Works well! There's just an e2e failing related to dashboards

rephus:
> There's just an e2e failing related to dashboards

It is possible that the error is because someone has deleted a chart in render for testing this manually. 
We can ignore if this is the case

Issue #11011, opened 2024-08-05
Sql Runner: table should update config when a new query is run


magnew:
### Description

Sql Runner: table should update config when a new query is run.

At the moment, if a different query is run, the table breaks. It should update it's config, like the bar chart. 

### Steps to Reproduce the Bug or Issue

1. run a query
2. select the table vis
3. run a different query
4. the table breaks

### Cloud or self-hosting

None

PR #11020, opened 2024-08-05
fix: reset table config when columns change

magnew:
Closes: #11011 

### Description:

**The problem was:** when a new SQL query was run, the table viz kept it's old config which didn't match any of the new columns and the chart would render nonsense (empty table)

**Now:** when a new query is run, the columns get reset to the columns from the new query. This is the same behavior as the bar chart

**Note:** in the future we could try harder to merge the old columns with the new or show some kind of warning, but this should at least un-break the case of totally new queries

**Before**
before

**After**
after

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/store/tableVisSlice.ts
+++ b/packages/frontend/src/features/sqlRunner/store/tableVisSlice.ts
@@ -1,5 +1,6 @@
 import {
     ChartKind,
+    deepEqual,
     isTableChartSQLConfig,
     type TableChartSqlConfig,
 } from '@lightdash/common';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/features/sqlRunner/store/tableVisSlice.ts
```diff
@@ -1,19 +1,24 @@
...
+    config: TableChartSqlConfig | undefined;
 };
 
 export const tableVisSlice = createSlice({
     name: 'tableVisConfig',
-    initialState,
+    initialState: {
+        defaultColumnConfig: undefined,
+        config: undefined,
+    } as InitialState,
```
joaoviana, line 21: do we need this? 🤔

joaoviana (approved):
Consistent! 💪

packages/frontend/src/features/sqlRunner/store/tableVisSlice.ts
magnew: Probably not.

magnew: Actually yes, there are ways to avoid it but none is clearer or less work than this.

PR #11036, opened 2024-08-06
chore: update dashboard e2e test selectors

ZeRego:
Closes: 

### Description:

A new checkbox was introduced in the filter configuration. https://github.com/lightdash/lightdash/pull/9505
Tabs are both mounted so the e2e selector was picking up that checkbox as well.
I added a test id to that tab and improved the e2e selector. Didn't want to change the tabs to have `keepMounted=false` because they store useful state when creating/editing a filter.

### Reviewer actions


rephus (approved):
makes sense, if tests pass, we can merge !

Issue #11044, opened 2024-08-07
Add option to raise error in `lightdash start-preview`


yu-iskw:
### Description

I want to add an option to handle whether or not a `lightdash start-preview` command raises an error, because the current implementation always ignores errors.

https://github.com/lightdash/lightdash/blob/acc641bccc9790751466eeab51334d0e0794c2eb/packages/cli/src/handlers/preview.ts#L190

```typescript
        await deploy(explores, {
            ...options,
            projectUuid: project.projectUuid,
            ignoreErrors: true,
        });
```

### What problem does this solve?

We run `lightdash start-preview` in GitHub Action. But, I have just realized we can't capture any errors in the automation process, because it returns the exit code 0 even if there are any errors.

PR #11045, opened 2024-08-07
feat: preview commands fail if there are compile errors

yu-iskw:
Closes: https://github.com/lightdash/lightdash/issues/11044

### Description:

Similar to `deploy` command, the preview commands will now fail if there are compile errors. 
To keep the existing behaviour, you can use the option `--ignore-errors`

Example:

```lightdash preview --ignore-errors```

### Reviewer actions

### DIFF
--- a/packages/cli/src/handlers/preview.ts
+++ b/packages/cli/src/handlers/preview.ts
@@ -25,6 +25,7 @@ type PreviewHandlerOptions = DbtCompileOptions & {
     name?: string;
     verbose: boolean;
     startOfWeek?: number;
+    raiseErrors: boolean;
 };
 
 type StopPreviewHandlerOptions = {
@@ -314,7 +315,7 @@ export const startPreviewHandler = async (
         await deploy(explores, {
             ...options,
             projectUuid: previewProject.projectUuid,
-            ignoreErrors: true,
+            ignoreErrors: !options.raiseErrors,
         });
         const url = await projectUrl(previewProject);
         console.error(`Project updated on ${url}`);
--- a/packages/cli/src/handlers/preview.ts
+++ b/packages/cli/src/handlers/preview.ts
@@ -25,7 +25,7 @@ type PreviewHandlerOptions = DbtCompileOptions & {
     name?: string;
     verbose: boolean;
     startOfWeek?: number;
-    raiseErrors: boolean;
+    ignoreErrors: boolean;
 };
 
 type StopPreviewHandlerOptions = {
--- a/packages/cli/src/handlers/preview.ts
+++ b/packages/cli/src/handlers/preview.ts
@@ -188,7 +188,6 @@ export const previewHandler = async (
         await deploy(explores, {
             ...options,
             projectUuid: project.projectUuid,
-            ignoreErrors: true,
         });
 
         await setPreviewProject(project.projectUuid, name);
@@ -246,7 +245,6 @@ export const previewHandler = async (
                     await deploy(await compile(options), {
                         ...options,
                         projectUuid: project.projectUuid,
-                        ignoreErrors: true,
                     });
                 }
 
@@ -315,7 +313,6 @@ export const startPreviewHandler = async (
         await deploy(explores, {
             ...options,
             projectUuid: previewProject.projectUuid,
-            ignoreErrors: options.ignoreErrors,
         });
         const url = await projectUrl(previewProject);
         console.error(`Project updated on ${url}`);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/cli/src/index.ts [hunk 1]
```diff
@@ -406,6 +406,7 @@ program
         parseUseDbtListOption,
         true,
     )
+    .option('--raise-errors', 'Raise errors if deploy fails', false)
```
yu-iskw, line 409: We can alsot use the `--ignore-errors` option as `lightdash deploy`. But, it changes the behaviors of the command. That's why I introduced `--raise-errors` to keep the current behavior.

ZeRego (changes requested):
A couple of changes required:

- For consistency with `deploy` command, we can use the same option but with a different default:
```
.option('--ignore-errors', 'Allows deploy with errors on compile', true)
``` 
- This option needs to be available for the commands `preview` and `start-preview`
- In the `packages/cli/src/handlers/preview.ts` we can update the `PreviewHandlerOptions` type to have `ignoreErrors: boolean` and remove `ignoreErrors: true,` from all 4 `deploy(...)` cases

yu-iskw:
@ZeRego thank you for the feedback. I was wondering how we can set it false with the `--ignore-errors` option, if we set the default value `true`. Does `--ignore-errors` option with the default value of `true` work?

I am not sure the feature to negate a boolean option works in our case. Another option can be to declare `--ignore-errors` and `--no-ignore-errors`. But I felt `--no-ignore-errors` looks difficult to understand.

https://www.npmjs.com/package/commander#other-option-types-negatable-boolean-and-booleanvalue
> If you define --foo first, adding --no-foo does not change the default value from what it would otherwise be.

```typescript
program
  .option('--no-sauce', 'Remove sauce')
  .option('--cheese <flavour>', 'cheese flavour', 'mozzarella')
  .option('--no-cheese', 'plain with no cheese')
  .parse();

const options = program.opts();
const sauceStr = options.sauce ? 'sauce' : 'no sauce';
const cheeseStr = (options.cheese === false) ? 'no cheese' : `${options.cheese} cheese`;
console.log(`You ordered a pizza with ${sauceStr} and ${cheeseStr}`);
```

ZeRego:
>  I was wondering how we can set it false with the --ignore-errors option, if we set the default value true.

We need to add a parser. Here is an example of an option with default `true`:

```
.option(
        '--use-dbt-list [true|false]',
        'Use `dbt list` instead of `dbt compile` to generate dbt manifest.json',
        parseUseDbtListOption,
        true,
    )
```

You would then be able to run ```lightdash start-preview --ignore-errors false```

yu-iskw:
Alright. I will do refactoring that part. Thanks.

yu-iskw:
@ZeRego let me clarify one thing. Do we need to change the behavior of `lightdash deploy --ignore-errors` as well? If so, that change might impact some Lightdash users. If not, the consistency between the option on `lightdash deploy` and `lightdash start-preview`.

```
# Need the argument
lightdash start-preview --ignore-errors false

# Don't need the argument
lightdash deploy --ignore-errors
```

ZeRego:
> @ZeRego let me clarify one thing. Do we need to change the behavior of `lightdash deploy --ignore-errors` as well? If so, that change might impact some Lightdash users. If not, lacks the consistency between the option on `lightdash deploy` and `lightdash start-preview`.
> 
> ```
> # Need the value
> lightdash start-preview --ignore-errors false
> 
> # Don't need the value
> lightdash deploy --ignore-errors
> ```

No changes to deploy. But you are correct, the option name is consistent but the default value would differ between deploy/preview commands and might cause some confusion
We can set default `false` for all of them and I will notify the community about this behaviour change and how to use the option to keep ignoring errors on preview.

yu-iskw:
@ZeRego I have modified the code. Can you review it again please?

https://github.com/lightdash/lightdash/pull/11045/commits/f265b9b70b4774ac9c536fd38feed0576e7e5e3f

packages/cli/src/handlers/preview.ts
```diff
@@ -314,7 +315,7 @@ export const startPreviewHandler = async (
         await deploy(explores, {
             ...options,
             projectUuid: previewProject.projectUuid,
-            ignoreErrors: true,
+            ignoreErrors: !options.raiseErrors,
```
ZeRego, line 318: We need to remove `ignoreErrors: !options.raiseErrors,` and `ignoreErrors: true` from all calls to `await deploy(...)`.

yu-iskw:
@ZeRego Sorry for bothering you. Can you review it again please?

ZeRego (approved):
lgtm 🙌  

Going to prep an announcement and release this tomorrow morning 🚀

yu-iskw:
@ZeRego thank you for testing this. I look forward to the release.

Issue #10831, opened 2024-07-24
`lightdash dbt run` overwrites custom description for state field


jpetey75:
### Description

I am creating a new model and ran into this, but my `state` is not the same as the the one in docs.

### Steps to Reproduce the Bug or Issue

1. In our internal analytics project, create a field named `state` and give it a custom description
2. Run `lightdash dbt run` on that model
3. The description will be overwritten with the docstring reference

Video: https://www.loom.com/share/f324aa4654074aca801b40f2225c1ced

### Cloud or self-hosting

None

joaoviana:
Been looking into this and will try to reproduce in our lightdash analytics repo.
You're right tho, why is it replacing the docs? 🤔 

What happens if you just run `lightdash dbt run`? 

From the code logic it should only replace if there's any docs reference, and I don't see that being the case here.

jpetey75:
@joaoviana there is a docs reference for `state` it's in the Stripe package that is used by our internal dbt project https://lightdash.slack.com/archives/C01788UVBUG/p1722028471349699?thread_ts=1721868961.135709&cid=C01788UVBUG

PR #11073, opened 2024-08-08
feat: allow for lightdash dbt run --assume-yes flag

joaoviana:
Closes: #10831

### Description:

- When users run `lightdash dbt run`, `assumeYes` is currently turned on by default. So all doc finds will overwrite the existing description for a dimension/metric.
- This adds the ability to pass `--assume-yes` and if not passed, it asks the user through a prompt if they want to replace the description field with a doc block. 

Demo:

https://github.com/user-attachments/assets/e10031b5-9f5f-4d0b-a338-0c1407e33829

### Reviewer actions

### DIFF
--- a/packages/cli/src/handlers/dbt/run.ts
+++ b/packages/cli/src/handlers/dbt/run.ts
@@ -54,7 +54,6 @@ export const dbtRunHandler = async (
     }
     await generateHandler({
         ...options,
-        assumeYes: options.assumeYes,
         excludeMeta: options.excludeMeta,
     });
 };
--- a/packages/cli/src/handlers/dbt/run.ts
+++ b/packages/cli/src/handlers/dbt/run.ts
@@ -9,13 +9,12 @@ import { DbtCompileOptions } from './compile';
 type DbtRunHandlerOptions = DbtCompileOptions & {
     excludeMeta: boolean;
     verbose: boolean;
+    assumeYes: boolean;
 };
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/handlers/dbt/run.ts
```diff
@@ -53,7 +54,7 @@ export const dbtRunHandler = async (
     }
     await generateHandler({
         ...options,
-        assumeYes: true,
+        assumeYes: options.assumeYes,
```
ZeRego, line 57: We can remove the line as it is passed in `...options,`

packages/cli/src/index.ts
```diff
@@ -243,7 +243,7 @@ ${styles.bold('Examples:')}
         false,
     )
     .option('--verbose', undefined, false)
-
+    .option('-y, --assume-yes', 'assume yes to prompts', false)
```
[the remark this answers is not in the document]
ZeRego: This makes sense to have the same default as the `generate` command.

ZeRego (approved):

packages/cli/src/handlers/dbt/run.ts
joaoviana: ty, done here: https://github.com/lightdash/lightdash/pull/11073/commits/868df8f55bf6fdef51a98ac7d0e8947fb8e5d93f

Issue #11080, opened 2024-08-08
Sort tables in filters the same as tables in the table explore left sidebar


jpetey75:
### Description

Right now filter pickers sort tables differently than the table list on the left inside of a table explore

### What problem does this solve?

Makes it more obvious what field you're picking and avoids mistakes picking the wrong field due to different orderings.

jpetey75:
Filters:

![image](https://github.com/user-attachments/assets/0a3559ab-eaa6-45fa-894a-e1bb096492f5)

Table explore:

![image](https://github.com/user-attachments/assets/155441da-e667-4890-becf-8e5091aae5a3)

jpetey75:
customer request from internal slack https://lightdash.slack.com/archives/C075JNX28GG/p1723118273985949

PR #11083, opened 2024-08-08
feat: sort field select similar to explore tree

joaoviana:
Closes: #11080

### Description:

Sorts fields in Filter select with the following priority: 
- base table first, and inside it dimensions > custom dimensions > metrics > additional metrics
- ...other tables
- then table calculations

This also generates a map of table names <-> labels so that we can group custom dimensions together with dimensions.

This won't 100% mimic the current explorer sidebar but at least has a more sensible sorting logic in place.

> [!NOTE]
> I wanted to move custom dims and custom metrics to be below dims + metrics, but custom metrics are compiled as metrics so I couldn't.

https://github.com/user-attachments/assets/ef1f9d95-d9a9-46d5-ba10-c60dafaa814b

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
+++ b/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
@@ -268,6 +268,7 @@ const FiltersCard: FC = memo(() => {
                     isEditMode={isEditMode}
                     filters={filters}
                     setFilters={setFilters}
+                    tableName={tableName}
                 />
             </FiltersProvider>
         </CollapsableCard>
--- a/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
+++ b/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
@@ -263,12 +263,12 @@ const FiltersCard: FC = memo(() => {
                 popoverProps={{
                     withinPortal: true,
                 }}
+                baseTable={data?.baseTable}
             >
                 <FiltersForm
                     isEditMode={isEditMode}
                     filters={filters}
                     setFilters={setFilters}
-                    baseTable={data?.baseTable}
                 />
             </FiltersProvider>
         </CollapsableCard>
--- a/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
+++ b/packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
@@ -268,7 +268,7 @@ const FiltersCard: FC = memo(() => {
                     isEditMode={isEditMode}
                     filters={filters}
                     setFilters={setFilters}
-                    tableName={tableName}
+                    baseTable={data?.baseTable}
                 />
             </FiltersProvider>
         </CollapsableCard>
--- a/packages/frontend/src/components/common/FieldSelect/index.tsx
+++ b/packages/frontend/src/components/common/FieldSelect/index.tsx
@@ -7,6 +7,7 @@ import {
     isField,
     isMetric,
     isTableCalculation,
+    sortTimeFrames,
     type Item,
 } from '@lightdash/common';
 import {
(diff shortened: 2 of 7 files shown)


packages/frontend/src/components/common/FieldSelect/index.tsx
```diff
@@ -86,13 +91,47 @@ const FieldSelect = <T extends Item = Item>({
...
     hasGrouping = false,
+    tableName,
     ...rest
 }: FieldSelectProps<T>): JSX.Element => {
     const sortedItems = useMemo(() => {
-        return items.sort((a, b) =>
-            getLabel(a, hasGrouping).localeCompare(getLabel(b, hasGrouping)),
-        );
-    }, [items, hasGrouping]);
+        // Give priority to items in the table === tableName
```
joaoviana, line 98: I can move this to a util function 

I also tried to mimic the logic in `ExploreTree/index`
```
const tableTrees = useMemo(() => {
        return Object.values(explore.tables)
            .sort((tableA, tableB) => {
                if (tableA.name === explore.baseTable) return -1;
                if (tableB.name === explore.baseTable) return 1;
                // Sorting explores by label
                return tableA.label.localeCompare(tableB.label);
            })
            .filter(
                (table) =>
                    !(isSearching && !searchHasResults(table)) && !table.hidden,
            );
    }, [explore, searchHasResults, isSearching]);
```

But these are diff structures.

packages/frontend/src/components/Explorer/FiltersCard/FiltersCard.tsx
```diff
@@ -268,6 +268,7 @@ const FiltersCard: FC = memo(() => {
                     isEditMode={isEditMode}
                     filters={filters}
                     setFilters={setFilters}
+                    baseTable={data?.baseTable}
```
ZeRego, line 271: could the base table be part of the filters provider and avoid the prop drilling?

joaoviana: good point, checking!

joaoviana: done here! https://github.com/lightdash/lightdash/pull/11083/commits/5dc7e9b8ea1fe523d42c126a4a63a5ca7aee3f69

packages/frontend/src/components/common/FieldSelect/index.tsx
```diff
@@ -86,13 +91,64 @@ const FieldSelect = <T extends Item = Item>({
...
+            if (isMetric(i)) return 3; // Additional metrics are compiled as metrics
+            return 4; // Table calculations have the lowest priority
+        };
+
+        return [
+            map,
+            [...items].sort((a, b) => {
+                /**
+                 * Sorting logic:
+                 * Sorts by table first
```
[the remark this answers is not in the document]
joaoviana: Done here: https://github.com/lightdash/lightdash/pull/11083/commits/f49c447e50915f0307ed12231b3a555246daa1ea

Also grouped time interval-based dims with the sort util by timeframe

Q for you: do we want to handle `groups` sorting as well?
(example would be in the events table with the `Event` group)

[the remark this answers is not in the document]
ZeRego: No, at the moment the important is to sort the tables correctly. Sorting the fields as they are in the sidebar involves more work as it is configurable at the table level. https://github.com/lightdash/lightdash/blob/2fb9b188f4e7041daea634a8b74d08c9a28497ce/packages/common/src/types/table.ts#L21
Sidebar field sorting logic: https://github.com/lightdash/lightdash/blob/fd2f9282b67a91ddd5a726b6733f3f5fa41118c2/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeNodes.tsx#L20

ZeRego (approved):

packages/frontend/src/components/common/FieldSelect/index.tsx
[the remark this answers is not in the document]
joaoviana: ahhh got it, thanks!

Issue #11059, opened 2024-08-07
SQL Runner schema improvements


magnew:
- [ ] Show all projects + schemas available to connection
- [ ] Cache schema in database or cloud storage for fast fetching
- [ ] get /tables should check cache and refresh if cache is empty
- [ ] introduce refresh endpoint to fetch the schema and give the option in the UI

PR #11099, opened 2024-08-09
feat: improved warehouse schemas

owlas:
Closes: #11059 

- [x] Adds db table to cache warehouse schemas (cached per project/user credentials)
- [x] Update bigquery `getTables` -> `getAllTables`
- [x] Update snowflake
- [x] Update redshift
- [x] Update postgres
- [ ] Update databricks
- [ ] Update trino
- [x] Update SqlRunnerService to fetch from cache or refresh cache depending on warehouse credentials
- [ ] Add new SqlRunnerService to trigger refresh cache


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2770,6 +2782,57 @@ export class ProjectService extends BaseService {
...
+        } else {
+            await this.warehouseAvailableTablesModel.createAvailableTablesForProjectWarehouseCredentials(
+                projectUuid,
+                warehouseTables,
+            );
+        }
+
+        await sshTunnel.disconnect();
+
+        return catalog;
```
magnew, line 2833: Do I have to query the database again to get this after adding it just above, or is it ok to return it here like this?

owlas (commented):
**Tested**

- ✅ Populating catalog on initial load with project credentials
- ✅ Populating catalog on initial load with user credentials

**Needed in another PR**

- Cannot test refresh behaviour without endpoint
- Better loading message in the sidebar, since loading the catalog could take a long time. e.g. spinner then after 1 second "please wait this could take a few minutes"
- No error message if sidebar `/tables` has an error, it's just blank. This happened to me when I enabled user warehouse credentials but hadn't provided any
- Databricks + Trino

magnew (approved):

Issue #11030, opened 2024-08-06
I can select dimensions/metrics from the sidebar


TuringLovesDeathMetal:
### Description

I can select metrics and dimensions from teh sidebar by clicking on them. When I click on them, they get added to the results table. 

- As you add metrics/dimensions to your results or filters, unavailable metrics/dimensions are disabled from the sidebar + filters list.
  - I don't think they should be hidden because that's really confusing that they just disappear. Maybe they go to a list at the bottom? under a single toggle? 
- Initially, we'll need to trigger an API call to get these metrics/dimensions (this will be a bit slow).

PR #11101, opened 2024-08-09
feat: hide fields on toggle

almeidabbm:
Relates to: #11030 

### Description:

- Set visibility of fields when selecting from sidebar

test-frontend
test-backend

### Reviewer actions

### DIFF
--- a/packages/common/src/types/semanticLayer.ts
+++ b/packages/common/src/types/semanticLayer.ts
@@ -52,3 +52,18 @@ export interface SemanticLayerTransformer<
     resultsToResultRows: (results: ResultsType) => SemanticLayerResultRow[];
     sqlToString: (sql: SqlType) => string;
 }
+
+export interface SemanticLayerClient {
+    getViews: () => Promise<SemanticLayerView[]>;
+    getFields: (
+        viewName: string,
+        selectedFields: Pick<
+            SemanticLayerQuery,
+            'dimensions' | 'timeDimensions' | 'metrics'
+        >,
+    ) => Promise<SemanticLayerField[]>;
+    getResults: (
+        query: SemanticLayerQuery,
+    ) => Promise<SemanticLayerResultRow[]>;
+    getSql: (query: SemanticLayerQuery) => Promise<string>;
+}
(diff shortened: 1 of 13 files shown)


packages/frontend/src/features/semanticViewer/api/hooks.ts
```diff
@@ -26,15 +26,31 @@ export const useSemanticLayerViews = ({
...
 }: SemanticLayerViewFieldsParams) =>
     useQuery<SemanticLayerField[], ApiError>({
-        queryKey: [projectUuid, 'semanticLayer', view, 'fields'],
-        queryFn: () => apiGetSemanticLayerViewFields({ projectUuid, view }),
+        queryKey: [
+            projectUuid,
+            'semanticLayer',
+            view,
+            'fields',
+            selectedFields,
```
IrakliJani, line 46: do we need to serialize this?

IrakliJani (approved):

packages/frontend/src/features/semanticViewer/api/hooks.ts
almeidabbm: I believe `useQuery` already does that for us: `When a query needs more information to uniquely describe its data, you can use an array with a string and any number of serializable objects to describe it`: https://tanstack.com/query/v4/docs/framework/react/guides/query-keys

packages/common/src/types/semanticLayer.ts
```diff
@@ -52,3 +52,18 @@ export interface SemanticLayerTransformer<
...
+        selectedFields: Pick<
+            SemanticLayerQuery,
+            'dimensions' | 'timeDimensions' | 'metrics'
+        >,
+    ) => Promise<SemanticLayerField[]>;
+    getResults: (
+        query: SemanticLayerQuery,
+    ) => Promise<SemanticLayerResultRow[]>;
+    getSql: (query: SemanticLayerQuery) => Promise<string>;
+}
```
almeidabbm, line 69: Interface that all semantic layer clients need to implement so they can work with the service

Issue #11018, opened 2024-08-05
All charts in dashboard deleted when editing multiple dashboards in the same tab


almeidabbm:
Steps to reproduce:
1. Create 2 dashboards
2. Create charts in dashboards for both dashboards
3. Click to edit the chart in one of the dashboards (let's call this, dashboard 1)
4. Copy the **edit** link of the other dashboard (dashboard 2) into the same tab where you were editing the chart of dashboard 1
5. Dashboard 2 opens in edit mode - notice that the title is of dashboard 2 but the charts are from dashboard 1
6. Click "Cancel"

Relates to: [#11016](https://github.com/lightdash/lightdash/pull/11016)

Video with reproduction steps:

https://github.com/user-attachments/assets/92a6f593-713b-4a1e-9931-41f3f4cdefeb

ZeRego:
We are closing this ticket:
- it is an edge case
- we stopped deleting the charts in the DB which was the biggest problem as we couldn't recover the charts
- we are planning to change this entire UI flow and stop relying on session storage

PR #11103, opened 2024-08-09
fix: clear dashboard storage if uuids dont match

ZeRego:
Closes: [#11018](https://github.com/lightdash/lightdash/issues/11018)

### Description:

Replication steps:
- go to dashboard A
- click edit chart
- change browser URL to dashboard B in edit mode

Before:

https://github.com/user-attachments/assets/e1c1a859-5388-4146-9bfc-0a28ab44deb9

After:

https://github.com/user-attachments/assets/2f3834eb-5529-4935-a133-24146a0c3dfb

### Reviewer actions

### DIFF
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -725,9 +725,20 @@ const Dashboard: FC = () => {
 };
 
 const DashboardPage: FC = () => {
-    const { projectUuid } = useParams<{ projectUuid: string }>();
+    const { projectUuid, dashboardUuid } = useParams<{
+        projectUuid: string;
+        dashboardUuid: string;
+    }>();
     const { user } = useApp();
     const dashboardCommentsCheck = useDashboardCommentsCheck(user?.data);
+    const { clearDashboardStorage, getDashboardUuid } = useDashboardStorage();
+
+    useEffect(() => {
+        if (dashboardUuid !== getDashboardUuid()) {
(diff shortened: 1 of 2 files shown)


joaoviana:
Can you check if you can save a chart in a dashboard?

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -725,9 +725,20 @@ const Dashboard: FC = () => {
...
     const { user } = useApp();
     const dashboardCommentsCheck = useDashboardCommentsCheck(user?.data);
+    const { clearDashboardStorage, getDashboardUuid } = useDashboardStorage();
+
+    useEffect(() => {
+        if (dashboardUuid !== getDashboardUuid()) {
+            // Clear storage if the dashboardUuid in storage doesn't match the current dashboardUuid
+            clearDashboardStorage();
+        }
+    }, [dashboardUuid, clearDashboardStorage, getDashboardUuid]);
```
joaoviana, line 741: this prevents us from saving charts within dashboards. Perhaps a different check needs to be done?

owlas: what about keying the storage on the dashboardUuid? So you have a cache per-id?

ZeRego: We would still need to store the dashboardUuid separately so when we go from dashboard to create chart it knows from what cache to fetch. I propose we close this ticket as it is an edge case, we stopped deleting the charts in the DB and we are planning to change this UI flow.

Issue #10023, opened 2024-05-09
Allow all users to see underlying data with custom SQL dimensions


ZeRego:
### Description

Not developers/admins will get an error when seeing the underlying data of a saved chart that has custom SQL dimension.

This is because the metric query is generated in the FE and passed to the BE. 
We need to change the API to build the metric query in the BE so the user can't tamper with the custom SQL.

"Explore from here" and "drill by"  buttons should exclude custom SQL dimensions to avoid errors.

### What problem does this solve?

Not developers/admins can see the underlying data of a saved chart that has custom SQL dimension.

jpetey75:
I heard from a customer that they want to create a saved chart with a custom SQL dimension, then share that chart with other Lightdash users as a template. These users may not have access to edit custom SQL dimensions, but they should still be able to edit other things by clicking "explore from here", like adjusting filters or adding other dimensions/metrics. Right now the "explore from here" throws an error.

PR #11104, opened 2024-08-09
fix: do not include custom dimensions for drill into

agha4to:
EDIT: Does not close. See comment below.
~Closes: https://github.com/lightdash/lightdash/issues/10023~

### Description:
Issue description states:
> "Explore from here" and "drill by" buttons should exclude custom SQL dimensions to avoid errors.

I can replicate, and this fixes, "Drill into" errors. However I cannot replicate errors with any of the three "Explore from here" functions:
1. In Dashboard > Chart 3 dot menu > Explore from here menu item
2. In Chart > Explore from here header blue button
3. After clicking a metric value > "View underlying data" menu item > Modal > Explore from here header link

Interestingly, one of the "Explore" implementations has a check for custom dimensions so perhaps this was at least partially fixed. But far as I can tell, it seems no changes needed for "Explore from here". Also, the issue description mentions moving logic from frontend to backend, but that seems unnecessary to fix this.

Finally, please note that I tried checking user permissions with the following, but this was returning `true` yet the error was still happening so I did not include this check:
```
user.data?.ability.can(
    'manage',
    subject('CustomSql', {
        organizationUuid: user?.data?.organizationUuid,
        projectUuid,
    }),
)
```

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/MetricQueryData/DrillDownModal.tsx
+++ b/packages/frontend/src/components/MetricQueryData/DrillDownModal.tsx
@@ -117,6 +117,18 @@ const drillDownExploreUrl = ({
     extraFilters,
     pivotReference,
 }: DrillDownExploreUrlArgs) => {
+    let metricQueryWithoutCustomDimensions = metricQuery;
+    const customDimensions = metricQuery.customDimensions;
+    if (customDimensions) {
+        const noncustomDimensions = metricQuery.dimensions
+            .filter(dim_id => !customDimensions.find(cd => cd.id === dim_id));
+        metricQueryWithoutCustomDimensions = {
+            ...metricQuery,
+            dimensions: noncustomDimensions,
+            customDimensions: [],
+        };
+    }
+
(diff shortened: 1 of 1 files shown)


ZeRego:
@agha4to That ticket is to eventually support custom dimensions in underlying data modal so this change wouldn't close it. I makes sense to leave the reference but not close it. 

On fixing the drill into option, I belive we should disable it based on the fix we did for the underlying data here:
https://github.com/lightdash/lightdash/pull/7699/files

ZeRego (changes requested):
Even though this update is related to the ticket it won't close it. The ticket is to support custom dimensions in the underlying data modal.

We can follow the same fix we did for underlying data: https://github.com/lightdash/lightdash/pull/7699/files

agha4to:
@ZeRego, I can use `hasCustomDimension` like in the code you linked. Is that what you mean? I feel like custom dimensions need to be filtered out still, but please lmk and thank you.

Issue #10818, opened 2024-07-24
Aliased joins are not correctly querying tables in Lightdash


TuringLovesDeathMetal:
### Description

When I use a metric defined in the .yml, the metric returns values from the non-aliased join and does not trigger the additional joins. When I use a custom metric based on the aliased join, the additional joins are triggered and the custom metric returns the correct results. The prebuilt metric, however, still returns values from the non-aliased join. 
 
The workaround where we use a view still works. I'll send screenshots of our SQL and .yml in a couple of messages:

### Steps to Reproduce the Bug or Issue

See more details in SLack thread here: https://lightdash.slack.com/archives/C06TWBVBN9F/p1721408349538969?thread_ts=1718391696.180329&cid=C06TWBVBN9F

### Cloud or self-hosting

cloud

PR #9842, opened 2024-04-22
fix: fix joined table with aliases in metric

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/9726

### Solution

 Generate a mapping with all the joins tables and aliases so we can evaluate the alias in the dimension reference method 

Screenshot from 2024-04-22 11-24-28

### Description:

This PR is only fixing the specified issue with nested joins , alias and sql metric. There are other scenarios that are not covered on this PR , like dimensions, user attributes, etc. THe approach should be similar to this fix though. 

Screenshot from 2024-04-22 12-21-47

Before
Screenshot from 2024-04-22 09-32-33

Screenshot from 2024-04-22 11-49-06

After:

Screenshot from 2024-04-22 12-21-03 (copy)

Screenshot from 2024-04-22 12-21-33

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #11128, opened 2024-08-13
fix: reference to join alias

ZeRego:
Closes: [#10818](https://github.com/lightdash/lightdash/issues/10818) 

### Description:

To reproduce:
- have 2 joins to the same model:
  - 1 without an alias
  - 1 with an alias 
- in the main model, add a metric with a reference to the join alias

To reproduce locally, change the orders model in schema.yml to:

```
    meta:
      joins:
        - join: customers
          sql_on: ${customers.customer_id} = ${orders.customer_id}
        - join: customers
          alias: customers_alias
          sql_on: ${customers_alias.customer_id} = ${orders.customer_id}
      metrics:
        test_alias:
          type: sum
          sql: ${customers_alias.customer_id}
```

**Important note**
Consideration when referencing joined models. If the model with the metric is joined to other models it will throw an error (see image bellow) unless that model also has the referred model.

Screenshot 2024-08-13 at 10 28 52

Example:

model A joins model B and has a metric referencing model B.
model C joins model A. This will result in an error unless we explicitly join model B to model C.

Best practices:
- If the metric only references the joined table, it is a bad practice and should be moved to that model instead.
- If the metric references multiple metrics from the base and joined table, and if that model is joined to other models they need to:
  - exclude that metric for the join fields
  - make sure the same join exists there 

### Reviewer actions


packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -83,22 +83,14 @@ export type UncompiledExplore = {
 const getReferencedTable = (
     refTable: string,
     tables: Record<string, Table>,
-    currentTable: string,
```
ZeRego, line 86: @rephus I rolled back most of your fix from https://github.com/lightdash/lightdash/pull/9842 since the tables arg already has the joined tables under the correct alias. 

eg:
```
tables {
   orders: ...
   customers: ...
   customers_alias: ...
}
```

rephus: >  Consideration when referencing joined models. If the model with the metric is joined to other models it will throw an error (see image bellow) unless that model also has the referred model.

mm, is it this the same use case though ? 
- I have a model Customers (A) 
- I have another model orders (B) joins a model (A) with an alias, and uses the alias from model A in a metric. 

I'm not using this `foo` reference on another table like payments or similar. 

> since the tables arg already has the joined tables under the correct alias.

In this case, I'd expect to see `foo` inside the `tables` when `currentTable`=== orders 

```
tables {
  payments: {},
  orders: {},
  customers: {},
  foo: {}, // same content as customers
 }
  ```

Comment (ZeRego) [reply packages/common/src/compiler/exploreCompiler.ts:86]: > mm, is it this the same use case though ?

Yes, let's use my example and replace the names with your code. 

Example:
model A (**orders**) joins model B (**foo**) and has a metric (**test_customers_foo**) referencing model B (**foo**).
model C (**payments**) joins model A(**orders**). This will result in an error unless we explicitly join model B(**foo**) to model C(**payments**).

> I'm not using this foo reference on another table like payments or similar.

When we compile `payments` explore we compile `orders` table which has the reference to `foo`. 

> In this case, I'd expect to see foo inside the tables when currentTable=== orders

`tables` arg has all the tables for the explore being compiled. 

`orders` is compiled without errors because `foo` is in `tables`.
`payments` throws an error because only `payments`, `orders`, `customers` are part of the `tables`.


Your example with payments only seemed to work because it was already joining `customers`. If `payments` doesn't join `customers` it would result in an error. The other assumption is that the join SQL would be the same between `foo` from orders join and `customers` from payments join

File: packages/common/src/compiler/exploreCompiler.test.ts [hunk 2]
```diff
@@ -159,6 +161,11 @@ describe('Explores with a base table and joined table', () => {
             compiler.compileExplore(joinedExploreWithSubsetOfFields),
         ).toStrictEqual(compiledJoinedExploreWithSubsetOfFields);
     });
+    test('should compile with 2 joins to the same table, one without alias and one with alias', () => {
```
rephus, line 164: :+1:

rephus (approved):

PR #11129, opened 2024-08-13
chore: refactor variable names for runners

owlas:
no logic changes


packages/common/src/visualizations/SqlResultsRunner.ts
```diff
@@ -383,11 +378,87 @@ export class SqlRunnerResultsTransformer
...
+    public async getPieChartData(): // config: SqlTransformPieChartConfig,
     Promise<PieChartData> {
         // TODO: NYI
         return {
             results: this.rows,
         };
     }
+
+    // TODO: merge with function below
+    getPieChartConfig({
```
magnew, line 390: My one concern is that this function and the one after it 'getChartConfig' don't seem to belong on the runner since they basically define the default chart config and how to update it. For example, the initial default config comes from the if at line 399. 

I know we talked about it a bit and you put these here because we didn't what the view code having the Sql types, but this doesn't seem like the right place. I kind of feel like the default config and config update logic should be in the chart models. Maybe we can change these not to rely on sql types, but a more abstract config type instead of moving them here?

Hmm. I will say, on the other hand, maybe this config update logic is going to be specific to SQL interactions and different on the semantic view page. So these could live here, but if we think the default config and config options are going to be the same in both pages (which seems likely), we should put at as much of this as possible in the Model layer.

magnew (approved):

magnew:
There are some build issue though. It's too bad the linter doesnt catch those. I wonder if we could fix that.

Issue #11086, opened 2024-08-08
Selected fields appear at the top of the field sidebar


owlas:
### Description

https://app.eraser.io/workspace/VE3IElerMSTDbEJzFEIf

PR #11136, opened 2024-08-13
feat: semantic viewer layout and UI

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/11030
Closes: https://github.com/lightdash/lightdash/issues/11086

### Reviewer actions

### DIFF
--- a/packages/backend/src/clients/cube/transformer.ts
+++ b/packages/backend/src/clients/cube/transformer.ts
@@ -70,7 +70,6 @@ const getSemanticLayerTimeGranularity = (
     }
 };
 
-// FIXME: we should not export from here. maybe common?
 export const getCubeTimeDimensionGranularity = (
     semanticGranularity: SemanticLayerTimeGranularity,
 ): TimeDimensionGranularity => {
--- a/packages/backend/src/clients/cube/transformer.ts
+++ b/packages/backend/src/clients/cube/transformer.ts
@@ -70,7 +70,8 @@ const getSemanticLayerTimeGranularity = (
     }
 };
 
-const getCubeTimeDimensionGranularity = (
+// FIXME: we should not export from here. maybe common?
+export const getCubeTimeDimensionGranularity = (
     semanticGranularity: SemanticLayerTimeGranularity,
 ): TimeDimensionGranularity => {
     switch (semanticGranularity) {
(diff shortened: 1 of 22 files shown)


packages/frontend/src/features/semanticViewer/components/SidebarViewFields.tsx
```diff
@@ -94,9 +91,22 @@ const SidebarViewFields = () => {
...
+        }
+
+        switch (field.kind) {
+            case FieldKind.DIMENSION:
+                return dispatch(toggleDimension(field));
+            case FieldKind.METRIC:
+                return dispatch(toggleMetric(field));
+            default:
+                return assertUnreachable(field.kind, 'Unknown field kind');
+        }
```
almeidabbm, line 109: `granularity` in `timeDimension` is nullable, should we just check if `kind === FieldKind.DIMENSION && type === SemanticLayerFieldType.TIME` so that time dimensions always get sent in that query param?

packages/backend/src/clients/cube/transformer.ts
```diff
@@ -65,7 +65,8 @@ const getSemanticLayerTimeGranularity = (
     }
 };
 
-const getCubeTimeDimensionGranularity = (
+// FIXME: we should not export from here. maybe common?
+export const getCubeTimeDimensionGranularity = (
```
almeidabbm, line 69: I don't think we can export from common since `TimeDimensionGranularity` is not available there, it comes from a package installed in backend

packages/frontend/src/features/semanticViewer/components/SidebarViewFields.tsx
IrakliJani: good point.. updating

packages/backend/src/clients/cube/transformer.ts
IrakliJani: ok, im gonna remove the comment

packages/backend/src/clients/dbtCloud/transformer.ts
```diff
@@ -109,11 +109,8 @@ export const dbtCloudTransfomers: SemanticLayerTransformer<
     semanticLayerQueryToQuery: (query) => {
         const { metrics, dimensions, timeDimensions } = query;
         return {
-            metrics: metrics.map((metric) => ({ name: metric })),
-            groupBy: [
-                ...dimensions.map((dimension) => ({ name: dimension })),
-                ...timeDimensions,
-            ],
+            metrics,
+            groupBy: [...dimensions, ...timeDimensions],
```
almeidabbm, line 113: One thing here, we should map `timeDimensions` to pass `granularity` as `grain` to `groupBy`: `{name: 'a', granularity: 'day'}` => `{name: 'a', grain: 'day'}`. Because the dbt type nullable then TS is not complaining

almeidabbm (approved):

PR #11137, opened 2024-08-14
chore: refactor pie chart layout

owlas:


packages/common/src/types/sqlRunner.ts
```diff
@@ -100,13 +100,13 @@ export type BarChartSqlConfig = SqlRunnerChartConfig & {
...
     type: ChartKind.LINE;
-    fieldConfig: SqlCartesianChartLayout | undefined;
+    fieldConfig: SqlPivotChartLayout | undefined; // PR NOTE: types are identical
     display: CartesianChartDisplay | undefined;
 };
 
 export type PieChartSqlConfig = SqlRunnerChartConfig & {
     type: ChartKind.PIE;
-    fieldConfig: SqlPieChartConfig | undefined;
+    fieldConfig: SqlPivotChartLayout | undefined; // PR NOTE: this will break serialization to the database (types are different)
```
owlas, line 109: Pie chart `fieldConfig` is now identical to cartesian chart so it can use the same runner methods.

Is this too generic? Since `fieldConfig` supports pivots but the pie chart doesn't? This is my only uncertainty but I'm down to try this and see how it scales.

packages/common/src/visualizations/CartesianChartDataTransformer.ts
```diff
@@ -1,39 +1,26 @@
...
-    constructor(args: {
-        transformer: ResultsRunnerBase<TCartesianChartLayout, TPieChartConfig>;
-    }) {
+    constructor(args: { transformer: ResultsRunnerBase<TPivotChartLayout> }) {
         this.transformer = args.transformer;
     }
 
-    async getEchartsSpec(config: SqlRunnerChartConfig) {
-        if (!isCartesianChartSQLConfig(config)) {
+    async getEchartsSpec(
```
owlas, line 13: Improve typing by exploding the config into `layout` (not used by chart models only runner) and `display` (not used by runner only chart models). i.e. `layout` is opaque to chart (since it depends on the runner context)

packages/common/src/visualizations/PieChartDataTransformer.ts
```diff
@@ -1,33 +1,23 @@
...
         }
-
-        const { fieldConfig, display } = config;
-
-        const transformedData = fieldConfig
-            ? await this.transformer.getPieChartData(
-                  fieldConfig as TPieChartConfig,
-              )
-            : undefined;
+        const transformedData = await this.transformer.getPivotChartData(
```
owlas, line 18: pie chart now uses the same runner method as cartesian charts

packages/common/src/visualizations/ResultsRunnerBase.ts
```diff
@@ -6,22 +6,12 @@ export type PivotChartData = {
     valuesColumns: string[];
 };
 
-export type PieChartData = {
-    results: RowData[];
-};
-
-// TODO: combine pivot chart + pie chart methods
-export interface ResultsRunnerBase<TPivotChartLayout, TPieChartLayout> {
+export interface ResultsRunnerBase<TPivotChartLayout> {
```
owlas, line 9: Runner has no pie chart specific config, only pivot type layouts

packages/common/src/visualizations/SqlResultsRunner.ts
```diff
@@ -337,128 +339,30 @@ export class SqlRunnerResultsTransformer
         };
     }
 
-    pieChartGroupFieldOptions(): PieChartDimensionOptions[] {
```
owlas, line 340: goodbye pie chart runners

packages/common/src/visualizations/SqlResultsRunner.ts
```diff
@@ -337,128 +339,30 @@ export class SqlRunnerResultsTransformer
...
+        let mergedLayout = currentConfig;
+
+        if (!currentConfig || !someFieldsMatch) {
+            mergedLayout = newDefaultLayout;
         }
 
         return {
-            newConfig,
-            newDefaultLayout,
+            defaultLayout: newDefaultLayout,
```
owlas, line 364: is this `mergedLayout` thing needed? Still unsure why. Either way, it's now generic to cartesian + pie

packages/common/src/visualizations/PieChartDataTransformer.ts
```diff
@@ -47,11 +37,9 @@ export class PieChartDataTransformer<TBarChartLayout, TPieChartConfig> {
                     radius: display?.isDonut ? ['30%', '70%'] : '50%',
                     center: ['50%', '50%'],
                     data: transformedData?.results.map((result) => ({
-                        name: result[
-                            fieldConfig?.groupFieldIds?.[0] || ''
-                        ] as string,
-                        groupId: fieldConfig?.groupFieldIds?.[0] || '',
-                        value: result[fieldConfig?.metricId || ''] as number,
+                        name: result[transformedData.indexColumn.reference],
```
owlas, line 40: note previously there could be many group field ids. Now we use the `index` (e.g. `x` on bar chart) property on `PivotLayout`and there can be only one (for now).

Alternative is to use the `pivot` property instead (e.g. colorBy on bar chart) but I don't think that makes as much sense

packages/frontend/src/features/sqlRunner/transformers/useSqlChart.ts
```diff
@@ -24,23 +25,23 @@ export const useSqlChart = (
             }),
         [rows, columns],
     );
-
-    const visTransformer = useMemo(() => {
-        if (isCartesianChartSQLConfig(config)) {
+    return useAsync(async () => {
```
owlas, line 28: refactor to handle `getEchartsSpec` having different signatures across pie vs. cartesian

packages/common/src/visualizations/SqlResultsRunner.ts
magnew: I think the merged layout is good because it keeps us from totally replacing the config when some fields match. 

BUT, I don't think we actually use defaultLayout anymore. It gets stored in redux after this return and seemingly never read again. Want to remove it? I could also take a not to follow up and look closer.

packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
```diff
@@ -20,16 +20,26 @@ export const barChartConfigSlice = createSlice({
...
+                    sqlRunnerResultsTransformer.mergePivotChartLayout({
+                        currentConfig: state.config?.fieldConfig,
                     });
 
-                state.config = newConfig;
-                state.defaultLayout = newDefaultLayout;
+                if (state.config) {
+                    state.config.fieldConfig = mergedLayout;
+                } else {
+                    state.config = {
```
magnew, line 33: I had moved this setting of default props out of the slice and into the 'Model' cause I thought it made more sense for the defaults to live on the models and the slices would just manage setting state. Which is why I was returning the new config from there and just setting it here. I don't have a strong reason for that except that the model seems like the right place for the defaults to be defined. At the moment the important one is just `type`, but it's definitely possible the display will also get a more complex default and having these view properties live in the slice doesn't seem quite right.

packages/frontend/src/features/sqlRunner/store/barChartSlice.ts
```diff
@@ -20,16 +20,26 @@ export const barChartConfigSlice = createSlice({
...
+                        metadata: {
+                            version: 1,
+                        },
+                        type: ChartKind.VERTICAL_BAR,
+                        fieldConfig: mergedLayout,
+                        display: undefined,
+                    };
+                }
+                state.defaultLayout = defaultLayout;
```
magnew, line 42: To speak to my other comment, I dont think we read this back anymore.

packages/frontend/src/features/sqlRunner/store/pieChartSlice.ts
```diff
@@ -53,19 +56,28 @@ export const pieChartConfigSlice = createSlice({
...
-                        currentConfig: state.config,
+                const { mergedLayout } =
+                    sqlRunnerResultsTransformer.mergePivotChartLayout({
+                        currentConfig: state.config?.fieldConfig,
                     });
-
-                state.config = newConfig;
+                if (state.config?.fieldConfig) {
+                    state.config.fieldConfig = mergedLayout;
+                } else {
```
magnew, line 71: Similar comment as to the bar chart.

packages/common/src/visualizations/PieChartDataTransformer.ts
magnew, line 1: Is this pre-rename? Did we already rename these to Models? If not, we can merge this first and do that next.

magnew (approved):
A few comments. Nothing blocking, but I vote for putting the default configs back in the Models. I commented with more details.

magnew (approved):

Issue #11092, opened 2024-08-09
SQL Runner: format fields as `%`


owlas:
### Description

Allow users to format a displayed value with `%`

Acceptance Criteria
- [ ] When I create a chart, I can see an option to format an x axis field or y axes on the Display/Styling section
- [ ] I can choose to format by `%` 
- [ ] Nice-to-have: allow formatting by a currency as well

PR #11152, opened 2024-08-14
feat: allow formatting by percentage

joaoviana:
Closes: #11092 

### Description:

Allows formatting of field by `percentage`

This is all done on the EchartsSpec.

https://github.com/user-attachments/assets/4cc41b9d-7582-4a5d-bf5a-ed3f05cf140c

### Reviewer actions

### DIFF
--- a/packages/common/src/visualizations/CartesianChartDataTransformer.ts
+++ b/packages/common/src/visualizations/CartesianChartDataTransformer.ts
@@ -1,9 +1,10 @@
-import { friendlyName } from '../types/field';
+import { CustomFormatType, Format, friendlyName } from '../types/field';
 import { ChartKind, ECHARTS_DEFAULT_COLORS } from '../types/savedCharts';
 import {
     isCartesianChartSQLConfig,
     type SqlRunnerChartConfig,
 } from '../types/sqlRunner';
+import { applyCustomFormat } from '../utils/formatting';
 import { type ResultsRunnerBase } from './ResultsRunnerBase';
 import { type IndexType } from './SqlResultsRunner';
 
@@ -56,6 +57,16 @@ export class CartesianChartDataTransformer<
         return this.colorMap.get(genericIdentifier)!;
     }
 
+    static getTooltipFormatter(format: Format | undefined) {
+        if (format === Format.PERCENT) {
+            return (value: number) =>
+                applyCustomFormat(value, {
+                    type: CustomFormatType.PERCENT,
+                });
+        }
+        return undefined;
+    }
+
     async getEchartsSpec(config: SqlRunnerChartConfig) {
         if (!isCartesianChartSQLConfig(config)) {
             return {};
@@ -80,8 +91,51 @@ export class CartesianChartDataTransformer<
             (row) => `${row[transformedData?.indexColumn.reference]}`,
         );
 
+        const series = transformedData?.valuesColumns.map(
+            (seriesColumn, index) => {
+                const seriesFormat = Object.values(display?.series || {}).find(
+                    (s) => s.yAxisIndex === index,
+                )?.format;
+
+                return {
+                    dimensions: [
+                        transformedData?.indexColumn.reference,
+                        seriesColumn,
+                    ],
+                    type: defaultSeriesType,
+                    stack: shouldStack ? 'stack-all-series' : undefined, // TODO: we should implement more sophisticated stacking logic once we have multi-pivoted charts
+                    name:
+                        (display?.series &&
+                            display.series[seriesColumn]?.label) ||
+                        friendlyName(seriesColumn),
+                    encode: {
+                        x: transformedData?.indexColumn.reference,
+                        y: seriesColumn,
+                    },
+                    yAxisIndex:
+                        (display?.series &&
+                            display.series[seriesColumn]?.yAxisIndex) ||
+                        0,
+                    tooltip: {
+                        valueFormatter: seriesFormat
+                            ? CartesianChartDataTransformer.getTooltipFormatter(
+                                  seriesFormat,
+                              )
+                            : undefined,
+                    },
+                    color: this.getSeriesColor(
+                        seriesColumn,
+                        possibleXAxisValues,
+                    ),
+                };
+            },
+        );
+
         return {
-            tooltip: {},
+            tooltip: {
+                trigger: 'axis',
+                appendToBody: true, // Similar to rendering a tooltip in a Portal
+            },
             legend: {
                 show: true,
                 type: 'scroll',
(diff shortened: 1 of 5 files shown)


packages/frontend/src/features/sqlRunner/components/CartesianChartStyling.tsx
```diff
@@ -19,6 +20,24 @@ export const CartesianChartStyling = ({
...
+        }
+        return state.barChartConfig.config.fieldConfig.y.map((f) => {
+            const format =
+                state.barChartConfig.config?.display?.series?.[f.reference]
+                    ?.format;
+            return {
+                reference: f.reference,
+                format,
+            };
+        });
```
joaoviana, line 38: should I create its own selector?

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -54,6 +54,7 @@ export const ContentPanel: FC = () => {
     const maxResultsHeight = useMemo(() => wrapperHeight - 56, [wrapperHeight]);
     const mantineTheme = useMantineTheme();
     const deferredInputSectionHeight = useDeferredValue(inputSectionHeight);
+    const deferredResultsHeight = useDeferredValue(resultsHeight);
```
joaoviana, line 57: small perf improvement

packages/common/src/visualizations/CartesianChartDataTransformer.ts
```diff
@@ -157,8 +202,12 @@ export type CartesianChartDisplay = {
     yAxis?: {
         label?: string;
         position?: string;
+        format?: Format;
     }[];
-    series?: Record<string, { label?: string; yAxisIndex?: number }>;
+    series?: Record<
+        string,
+        { label?: string; format?: Format; yAxisIndex?: number }
+    >;
```
joaoviana, line 210: thoughts on each having their own format? 🤔 I think it makes sense

packages/common/src/visualizations/CartesianChartDataTransformer.ts
```diff
@@ -80,8 +91,51 @@ export class CartesianChartDataTransformer<
             (row) => `${row[transformedData?.indexColumn.reference]}`,
         );
 
+        const series = transformedData?.valuesColumns.map(
+            (seriesColumn, index) => {
+                const seriesFormat = Object.values(display?.series || {}).find(
```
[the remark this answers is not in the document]
magnew: Hmm. Maybe I don't understand the question here. Is the question whether it is a good idea to track by index even though the values are stored as an object?

packages/common/src/visualizations/CartesianChartDataTransformer.ts
magnew: Each series? For sure. You might have some percentages and some raw numbers. Or whatever.

packages/frontend/src/features/sqlRunner/components/CartesianChartFormatConfig.tsx
```diff
@@ -0,0 +1,76 @@
...
+                value: option,
+                label: capitalize(option),
+            }))}
+            itemComponent={FormatItem}
+            icon={format && <FormatIcon format={format} />}
+            value={format ?? formatOptionsWithNone?.[0]}
+            onChange={(value) => value && onChangeFormat(value)}
+            styles={(theme) => ({
+                input: {
+                    width: '110px',
```
magnew, line 57: IMO, this looks a little funny truncated into the right side. Maybe it should be full width? That might be a little funny too. I suppose this can be a shout out to getting some design attention in here eventually 😄

packages/frontend/src/features/sqlRunner/components/CartesianChartStyling.tsx
magnew: Probably? The advantage of selectors is that they can be very specific and you can have them only update the selected value when that value changes. If you do it this way, I think this component will re-render more than necessary

packages/frontend/src/features/sqlRunner/store/cartesianChartBaseSlice.ts
```diff
@@ -194,10 +195,33 @@ export const cartesianChartConfigSlice = createSlice({
...
+                if (state.config.display.series) {
+                    Object.entries(state.config.display.series).forEach(
+                        ([key, series]) => {
+                            if (series.yAxisIndex === index) {
+                                // NOTE: Delete series from display object when it's removed from yAxis
+                                delete state.config?.display?.series?.[key];
+                            } else if (
+                                // NOTE: Decrease yAxisIndex for series that are after the removed yAxis
+                                series.yAxisIndex &&
+                                series.yAxisIndex > index
```
magnew, line 217: Can you do this safely? 🤔

packages/frontend/src/features/sqlRunner/store/cartesianChartBaseSlice.ts
```diff
@@ -206,5 +230,59 @@ export const cartesianChartConfigSlice = createSlice({
...
+            config.display.yAxis = config.display.yAxis || [];
+            config.display.yAxis[0] = {
+                ...config.display.yAxis[0],
+                format: validFormat,
+            };
+
+            // Update the format for series with yAxisIndex 0
+            if (config.display.series) {
+                Object.values(config.display.series).forEach((series) => {
+                    if (series.yAxisIndex === 0) {
```
magnew, line 253: Why the 0? Im just not seeing it

magnew (commented):
The approach seems good here. I had a few questions. Or share some of yours 😅.

Testing it I noticed a couple little bug that combine to be confusing (gif below):
- When I first run a query, it sets order_id on the y axis, which is a number, but it won't let me change to percent. 
- Then I change the y-axis, but it doesnt update the label in the style section. 
- If I then update the label manually, I can choose percent.

packages/frontend/src/features/sqlRunner/store/cartesianChartBaseSlice.ts
joaoviana: yes we can because the index will always be > 0 with the current state 👍

packages/frontend/src/features/sqlRunner/store/cartesianChartBaseSlice.ts
joaoviana: so that if you change the first series format, you also change the y axis format so that the label on the chart left is also _scaled_ to percentage. Hope that makes sense!

packages/frontend/src/features/sqlRunner/components/CartesianChartFormatConfig.tsx
joaoviana: 100%, it feels weird to have it full width, but I think this can be done with a custom select component that looks **better** 😅

packages/common/src/visualizations/CartesianChartDataTransformer.ts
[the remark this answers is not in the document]
joaoviana: The loop feels weird to me, but this is what I can do with our current config structure 😓

magnew (approved):

PR #11155, opened 2024-08-14
feat: semantic viewer field granularity selector

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/11125

### Description:
CleanShot 2024-08-14 at 23 21 53@2x

CleanShot 2024-08-14 at 23 20 50@2x

### Reviewer actions

### DIFF
--- a/packages/backend/src/generated/swagger.json
+++ b/packages/backend/src/generated/swagger.json
@@ -9041,7 +9041,7 @@
     },
     "info": {
         "title": "Lightdash API",
-        "version": "0.1216.0",
+        "version": "0.1218.0",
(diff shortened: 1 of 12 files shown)


almeidabbm (approved):
LGTM 🚀 

Works well with dbt, with the other semantic layer I got this error `Could not cast literal "UTC" to type TIME` but I tried running the same thing on their playground and got it as well

IrakliJani:
> LGTM 🚀
> 
> Works well with dbt, with the other semantic layer I got this error `Could not cast literal "UTC" to type TIME` but I tried running the same thing on their playground and got it as well

yeah thats what I observed but since the sql generation works I assume its not "on our end"

Issue #10969, opened 2024-08-02
Refactor results hooks


ZeRego:
We now have 3 places that fetch the results for the sql runner: sql runner (via sql), sql chart (via slug), dashboard (via uuid)
Chaining and managing to state of 3 useQuery is madness. We can combine the requests into a single react-query hook.

Example:
```
// function to be reused in all hooks that stream a file
const getResultsStream = async (url: string | undefined) => {
    ... handle stream data
};

const getSqlChartResults = async ({
                                      projectUuid,
                                      slug,
                                  }: {
    projectUuid: string;
    slug: string;
}) => {
    const scheduledJob = await lightdashApi<ApiJobScheduledResponse['results']>({
        url: `/projects/${projectUuid}/sqlRunner/saved/slug/${slug}/results-job`,
        method: 'GET',
        body: undefined,
    });

    // recursively check job status until it's complete
    async function getCompleteJob(): Promise<ApiSqlRunnerJobStatusResponse['results']> {
        ...
    }

    const job = await getCompleteJob();
    const url = job?.details && !isErrorDetails(job.details) ? job.details.fileUrl : undefined;
    return getResultsStream(url);
}

export const useSqlQueryRun = (
    projectUuid: string,
    slug: string | undefined,
) => {
    return useQuery<
        ResultRow[] | undefined,
        ApiError
    >(
        ['sqlChartResults', projectUuid, slug],
        () => {
            return getSqlChartResults({
                projectUuid,
                slug: slug!,
            });
        },
        {
            enabled: Boolean(slug),
        }
    );
};
```

PR #11156, opened 2024-08-15
refactor: simplify `useSqlQueryRun` hook

joaoviana:
Closes: [#10969](https://github.com/lightdash/lightdash/issues/10969)

### Description:

`useSqlQueryRun` only uses one mutation and has a `keepPreviousData`-like logic similar to `useQuery`'s prop. 
improved documentation of other hooks to be clearer

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
+++ b/packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
@@ -3,17 +3,16 @@ import {
     SchedulerJobStatus,
     type ApiError,
     type ApiJobScheduledResponse,
-    type ApiSqlRunnerJobStatusResponse,
     type ResultRow,
     type SqlColumn,
     type SqlRunnerBody,
 } from '@lightdash/common';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/hooks/useSqlQueryRun.tsx
```diff
@@ -50,169 +47,73 @@ export const useSqlQueryRun = ({
...
+                columns:
+                    job.details && !isErrorDetails(job.details)
+                        ? job.details.columns
+                        : [],
+            };
         },
         {
+            mutationKey: ['sqlRunner', 'run'],
             onError: () => {
                 showToastError({
```
joaoviana, line 97: do we want to clear previous data on error? 


In this PR:  
1. User types query with results
2. User types a query that leads to no results
3. Previous data is there, but an error toast message appears

Should we clear the previous data and have an empty state?
Perhaps it's a nice experience to keep the previous data, but unsure!

magnew (approved):

Issue #11159, opened 2024-08-15
Chore: refactor and unify components between sqlrunner and semantic viewer


rephus:
Based on the comments from: https://github.com/lightdash/lightdash/pull/11132

Rename these files: 

    SemanticeViewerResultsViewerFE -> SemanticViewerRunnerFE
    TableDataTransformer -> TableModel
    useTableDataTransformer -> useTableModel
    SqlRunnerChart -> ChartView

PR #11166, opened 2024-08-15
chore: refactor sql runner and semantic viewer shared componentes

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/11159

### Description:

Untitled Diagram drawio

### Main changes

- Move all the chart / configuration into its own DataViz
- Semantic/Sql is now using components from DataViz
- semantic/sql selectors inside DataViz is now taking props (eg: selectedChartType) 
- instead of triggering reducer on setSqlRunnerResults, we trigger it on OnResults (inside DataViz slice) . I think we can make it better to avoid making 2 dispatch calls. 

### Next steps

- [ ] DataViz is still using directly SqlRunnerResultsTransformerFE, we should probably pass this transfomer as a parameter, or add some abstraction 
- [ ] Some other components like TableDataTransfomers can also be extracted
- [ ] duckDbQuery can also be shared for now
- [ ] Probably we should also unify the hook for dealing with streaming results (run query -> check job -> get results -> parse jsonl) . This is not blocking any future work, but it would be nice. 
- [ ] Right now semanticviewer is converting semanticStuff into ResultRow/ SqlColumn for compatibility   with DataViz + transfomers (not even in the transfomer, but inside the react query hook) . We should probably create a new type that contains a more generic type or add some abstraction.

### Reviewer actions


packages/frontend/src/components/DataViz/store/barChartSlice.ts
```diff
@@ -15,7 +18,7 @@ export const barChartConfigSlice = createSlice({
         ...cartesianChartConfigSlice.caseReducers,
     },
     extraReducers: (builder) => {
-        builder.addCase(setSqlRunnerResults, (state, action) => {
+        builder.addCase(onResults, (state, action) => {
```
magnew, line 21: 👍 better name

packages/frontend/src/components/DataViz/store/hooks.ts
magnew, line 1: It looks like we have 3 of this file now: 1 in DataViz, one in sqlRunner and one in SemanticViewer. Can we jsut have one? Im not sure where it would go. maybe `src/hooks/storeHooks.ts`?

packages/frontend/src/components/DataViz/store/index.ts
magnew, line 1: We should delete this file, right? There is one on SemanticViewer and one in SqlRunner, but this one isn't used for real. A couple files import types from this, but they can get them from elsewhere, I think.

magnew (approved):
Nice work @rephus! This came together pretty well. I tested it pretty heavily in both SQL runner and Semantic Viewer. In semantic viewer I found a few issues, but I assume that's becuase it's more of a work in progress and less about this review (for exmaple, if you choose columns that don't work well for a bar chart, it hangs in data precessing). SQL looks good though, I think all of the functionality works as I expect. 

I put a couple comments about duplicated files I think should be cleaned up before we merge this, but aside from that, I agree with you'd next steps and I think we should get this it. Its a great leap forward.

packages/frontend/src/components/DataViz/store/hooks.ts
rephus: yes, you are right, but each file import the RootState (which is different) from their own folder , so they can't be reused. 
> import { type AppDispatch, type RootState } from '.';

We could get rid of this file and merge it with index.ts. What do you think ?

packages/frontend/src/components/DataViz/store/index.ts
rephus: same as above ? the store is different , and it is used in hooks .

magnew (changes requested):
Changing this to 'request changes'. There's a bug where the field selectors in the SQL runner don't update when a new selection is made.

IrakliJani (approved):

magnew (approved):

Issue #11047, opened 2024-08-07
SQL Runner: Virtual scrolling not working


magnew:
### Description

When a dataset is large, clicking the chart tab hangs a bit before switching to showing the chart. 

We haven't done much to optimize this. 

### Steps to Reproduce the Bug or Issue

1. Run a largish query, for a jaffle shop example: `SELECT * FROM "postgres"."jaffle"."generated_a" LIMIT 1000`, the main thing is the large limit
2. Click the chart tab

It's slow to render the first time.   

### Cloud or self-hosting

None

PR #11184, opened 2024-08-16
perf: track active configs and render on selection

joaoviana:
Related to: [#11047](https://github.com/lightdash/lightdash/issues/11047)

### Description:

We currently generate the data for all configs when the user hits run query. 

This gives a better experience between selecting different chart types, but slows down the initial load. 

This PR tracks the active chart config and _marks it as active_ so next time the user selects that same type, it won't re-mount.

Demo

https://github.com/user-attachments/assets/293aa5f8-5837-45ec-af4d-9736e9aec1d5

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -1,4 +1,7 @@
-import { isTableChartSQLConfig } from '@lightdash/common';
+import {
+    isTableChartSQLConfig,
+    type TableChartSqlConfig,
+} from '@lightdash/common';
 import {
     Box,
     getDefaultZIndex,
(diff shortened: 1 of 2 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -126,6 +116,30 @@ export const ContentPanel: FC = () => {
...
+                NonNullable<ReturnType<typeof selectChartConfigByKind>>,
+                TableChartSqlConfig
+            > => !isTableChartSQLConfig(c),
+        );
+
+        return {
+            chartConfigs,
+            tableConfig,
+        };
+    });
```
joaoviana, line 141: I could make this into its own selector separately, but would have to be a separate PR

owlas (approved):

Issue #11181, opened 2024-08-16
Syntax error: Expected \")\" but got \";\"


ZeRego:
### Description

SQL runner throws error if query contains a `;` 

### Steps to Reproduce the Bug or Issue

1. Go to new SQL runner page 
2. run a query that ends with `;` eg: `SELECT * FROM "example";`
3. see error

### Cloud or self-hosting

None

ZeRego:
So we need to update `applyLimitToSqlQuery()` to add our wrapping SQL in new lines, not inline with their SQL.
If they end their SQL with a comment it breaks atm.
`SELECT * from X; --`

Plus, change `runSqlQuery()` to use that method as well so we fix it in the old and new sql runner

PR #11190, opened 2024-08-19
fix: strip trailing semicolon on sql runners

owlas:
Closes: #11181

### DIFF
--- a/packages/backend/src/queryBuilder.ts
+++ b/packages/backend/src/queryBuilder.ts
@@ -307,7 +307,10 @@ export const applyLimitToSqlQuery = ({
     limit: number | undefined;
 }): string => {
     if (limit === undefined) return sqlQuery;
-    return `WITH user_sql AS (${sqlQuery}) select * from user_sql limit ${limit}`;
+    return `WITH user_sql AS (${sqlQuery.replace(
+        /;$/,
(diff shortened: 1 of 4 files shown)


packages/backend/src/queryBuilder.ts
```diff
@@ -307,7 +307,10 @@ export const applyLimitToSqlQuery = ({
     limit: number | undefined;
 }): string => {
     if (limit === undefined) return sqlQuery;
-    return `WITH user_sql AS (${sqlQuery}) select * from user_sql limit ${limit}`;
+    return `WITH user_sql AS (\n${sqlQuery.replace(
+        /;$/,
```
ZeRego, line 311: We need the multi-line option to capture the end of each line

```suggestion
        /;$/m,
```

Test with
```
with test as (SELECT * FROM "postgres"."jaffle"."customers")

select * from test;

-- test "postgres"."public"."dashboard_tile_comments"
```

ZeRego: We are not handling the edge case for now

> If we start stripping semicolons anywhere else but the last character, it might be in a string or something and not be a terminating semicolon

ZeRego (approved):

Issue #10666, opened 2024-07-09
Dashboard @ mentions not working sometimes


jpetey75:
### Description

Using @ to show a list of people to mention is not working in dashboards consistently. This was reported by a user and I reproduced it a few times, then after waiting a while it starts working. Is it possible there is a lag between the dashboard rendering and the list of users being available? Then once the list is cached it works fine?

Loom video of issue: https://www.loom.com/share/dfee07bef87d4831878381a3147a3d84

### Steps to Reproduce the Bug or Issue

1. Open a dashboard
2. Click on the comment icon in a tile
3. Type @ 
4. No list of users will appear initially

### Cloud or self-hosting

None

PR #11212, opened 2024-08-20
fix: Dashboard @ mentions not working sometimes

anshul45:
Closes: https://github.com/lightdash/lightdash/issues/10666

### Description:

Here in `packages/frontend/src/features/comments/components/CommentForm.tsx`
```
if ({isSuccess && userNames) <CommentWithMentions
                                suggestions={userNames}
                                shouldClearEditor={shouldClearEditor}
                                setShouldClearEditor={setShouldClearEditor}
                                onUpdate={setEditor}
                            />
```
"But when I console log suggestions, it shows values like [] then [{data}]. This happens because userNames is an array. Even if userNames has no data, it still passes the if condition. To fix this, we can do the following: 
```
{isSuccess && userNames && userNames.length > 0 && (
                            <CommentWithMentions
                                suggestions={userNames}
                                shouldClearEditor={shouldClearEditor}
                                setShouldClearEditor={setShouldClearEditor}
                                onUpdate={setEditor}
                            />
                        )}
```

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/comments/components/CommentForm.tsx
+++ b/packages/frontend/src/features/comments/components/CommentForm.tsx
@@ -94,7 +94,7 @@ export const CommentForm: FC<Props> = ({
                         </Avatar>
                     </Grid.Col>
                     <Grid.Col span={18} w={mode === 'reply' ? 300 : 350}>
-                        {isSuccess && userNames && (
+                        {isSuccess && userNames && userNames.length > 0 && (
                             <CommentWithMentions
                                 suggestions={userNames}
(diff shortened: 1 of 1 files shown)


ZeRego:
@anshul45 We should add a loading state to the `CommentForm` and wait for `useOrganizationUsers` and `useSpace` to be successful since we need both to have user suggestions.

ZeRego (changes requested):
We should add a loading state to CommentForm and wait for useOrganizationUsers and useSpace

Issue #11176, opened 2024-08-16
I would like to know why/where my SQL query has an error


joaoviana:
### Description

Currently, if there's the SQL query throws an error, we display a message `Couldn't fetch SQL query results`.

We should be more intentional with this error. We could read the error string, when it looks like this

> "Syntax error: Expected \")\" but got keyword WHERE at [3:1]"

And handle it better to aid the user. 

**Pre-requisites**
We should focus on 3 warehouses for now: 
- Postgres (easy to locally reproduce)
- BigQuery
- Snowflake

**Acceptance criteria**

On the Dataset/Table I'm referencing:
- [ ] When I reference a dataset that doesn't exist, and run a SQL query, I should know that it doesn't exist on the toast error
- [ ] I should also be able to see on the editor where the error is in. For example, if it's on line 3, I should see a marker on line 3 and a squiggly line under the wrong reference
- [ ] nice-to-have: I should get an alternative column that is close to what I'm referencing, like a `Did you mean xxx?`

On the column I'm referencing
- [ ] When I reference a table that doesn't exist, and run a SQL query, I should know that it doesn't exist on the toast error
- [ ] I should also be able to see on the editor where the error is in. For example, if it's on line 3, I should see a marker on line 3 and a squiggly line under the wrong reference
- [ ] nice-to-have: I should get an alternative column that is close to what I'm referencing, like a `Did you mean xxx?`

On the SQL I'm writing
- [ ] When I write _bad_ SQL code, I should have a linter-like experience. I should know what's wrong, and what should I do to correct it. 

### Implementation considerations/suggestions 

1. Go through Postgres, BigQuery, and Snowflake's APIs - do they have SQL linters or just error messaging support?
2. Adding an error marker to monaco editor _seems easy_, but how do we clear the error? After a succesfull SQL query?
3. Can we add `Sqlfmt`, `Sql fluff` or `Sql lint` to the monaco editor?
4. We have a shared warehouse client class, and we have methods for each, how would this method look like? 

### What problem does this solve?

Better SQL querying experience

PR #11238, opened 2024-08-22
fix: parse sql error message and highlight on editor

andrewdw:
Closes: #11176

### Description:

- In sqlRunner **BigQuery** and **Snowflake** syntax errors are are highlighted in the browser.
- Error message is more verbose; passing the db error to the frontend.

Screenshot 2024-08-22 at 2 40 10 PM

### Reviewer actions


packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -340,4 +350,60 @@ export class BigqueryWarehouseClient extends WarehouseBaseClient<CreateBigqueryC
...
+                        // return a new error with the line and character number in data object
+                        return new WarehouseQueryError(error.message, {
+                            lineNumber,
+                            charNumber,
+                        });
+                    }
+                    break;
+                }
+                break;
+            // case: 'accessDenied':
```
joaoviana, line 401: we can remove this comment

packages/common/src/types/sqlRunner.ts
```diff
@@ -225,3 +227,15 @@ export type ApiSqlChartWithResults = {
...
+export type MonacoHighlightChar = {
+    line: number;
+    char: number;
+};
+
+// monaco highlight line
+export type MonacoHighlightLine = {
+    start: MonacoHighlightChar;
+    end?: MonacoHighlightChar;
+};
```
joaoviana, line 241: I believe these should be in `SqlEditor`

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -40,6 +40,7 @@ import {
     DashboardProvider,
     useDashboardContext,
 } from '../providers/DashboardProvider';
+import '../styles/monaco.css';
```
joaoviana, line 43: this can be removed, let's add it to `ContentPanel`

packages/frontend/src/pages/MinimalDashboard.tsx
```diff
@@ -20,6 +20,7 @@ import {
 
 import { DashboardSqlChartTile as SqlChartTile } from '../components/DashboardTiles/DashboardSqlChartTile';
 import { useDateZoomGranularitySearch } from '../hooks/useExplorerRoute';
+import '../styles/monaco.css';
```
joaoviana, line 23: this should be removed here as well

packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -5,7 +5,15 @@ import {
...
     BigQueryTimestamp,
     Dataset,
+    JobResponse,
 } from '@google-cloud/bigquery';
+import {
+    common as bigqueryCommon,
+    JobsQueryResponse,
+    QueryResultsResponse,
+    SimpleQueryRowsCallback,
+    SimpleQueryRowsResponse,
```
joaoviana, line 15: are all of these being used?

packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -5,7 +5,15 @@ import {
     BigQueryTime,
     BigQueryTimestamp,
     Dataset,
+    JobResponse,
```
joaoviana, line 8: I don't think this is being used as well

packages/e2e/cypress/e2e/api/sqlRunner.cy.ts
```diff
@@ -47,6 +47,110 @@ Object.entries(warehouseConnections).forEach(
...
+                    headers: { 'Content-type': 'application/json' },
+                    method: 'POST',
+                    body: JSON.stringify({
+                        sql,
+                    }),
+                }).then((runResp) => {
+                    expect(runResp.status).to.eq(200);
+                    const { jobId } = runResp.body.results;
+
+                    const maxRetries = 20; // Snowflake is a bit slow compared to the others (normally takes ~3 seconds)
```
joaoviana, line 66: 🙌

packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
```diff
@@ -237,17 +259,73 @@ export const SqlEditor: FC<{
...
+                    inlineClassName: 'editorError',
+                },
+            },
+        ];
+
+        // Update decorations using the decorations collection
+        decorationsCollectionRef.current.set(newDecorations);
+    }, [sql, monaco, highlightText]);
+
+    useEffect(() => {
```
joaoviana, line 316: We can avoid using the `useEffect` and the internal state `value`. 

I recommend doing this instead (since we have RTK for the SQL state): 

```tsx
  onChange={onChange}
```

```tsx
 const onChange: OnChange = debounce((value: string | undefined) => {
        dispatch(setSql(value ?? ''));
        if (resetHighlightError && highlightText) {
            resetHighlightError();
        }
    }, DEBOUNCE_TIME);
```

As you start typing (and when there's an highlight text error) it will reset, removing then the need for this `useEffect`. It also has improved performance for the SQL typing

joaoviana (changes requested):
Works great! Just left a suggestion for the SqlEditor

packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
andrewdw: The issue with this is that the reset highlightError is debounced and will have a delay clearing, so as a user types (and fixing the error) it will remain highlighted until the debounce fires. The current useEffect will clear the highlight right away while still debouncing the dispatch.

joaoviana: oops! 

```tsx

    const debouncedSetSql = useMemo(
        () =>
            debounce((value: string) => dispatch(setSql(value)), DEBOUNCE_TIME),
        [dispatch],
    );

    const onChange: OnChange = useCallback(
        (value: string | undefined) => {
            if (highlightText && resetHighlightError) {
                resetHighlightError();
            }
            debouncedSetSql(value ?? '');
        },
        [debouncedSetSql, highlightText, resetHighlightError],
    );

```

This way we avoid use effect pitfalls when not providing all deps referenced in it. 

What do you think?

joaoviana (approved):
Tested:
- sql errors (syntax, reference etc)
- no sql errors
- saving a chart
- editing one

Issue #11031, opened 2024-08-06
There's a run query button that populates rows in a results table


TuringLovesDeathMetal:
### Description

I can manually trigger a query. 

- the run query button has a row limit (max limit is 5,000 rows)
- the results table from the query is default scroll (not paginated)
- We need to think about streaming results here (similar to the SQL runner).
- The metrics look different to dimensions (e.g. blue/yellow if no one has a better idea).

PR #11244, opened 2024-08-22
feat: semantic viewer run query button

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/11031

from the ticket:

- [x] I can manually trigger a query:
- [x] the run query button has a row limit (max limit is 5,000 rows)
- [x] (no pagination for now, loads full table) the results table from the query is default scroll (not paginated)
- [x] (done in a separate PR) We need to think about streaming results here (similar to the SQL runner).
- [ ] (not going to be implemented here) The metrics look different to dimensions (e.g. blue/yellow if no one has a better idea).

bonus:
- [x] abstract semantic layer configurations and have an api call to get information about the semantic layer client
- [x] fetch semantic layer limit separately
- [x] introduces better state initialization

### Reviewer actions


packages/common/src/types/semanticLayer.ts
```diff
@@ -112,14 +112,35 @@ export interface SemanticLayerTransformer<
...
 
+const SEMANTIC_LAYER_MAX_QUERY_LIMIT = 500;
+
 export function getDefaultedLimit(
     maxQueryLimit: number,
     queryLimit?: number,
 ): number {
-    return Math.min(queryLimit || 500, maxQueryLimit);
+    return Math.min(
+        queryLimit ?? SEMANTIC_LAYER_MAX_QUERY_LIMIT,
```
almeidabbm, line 122:
```suggestion
        queryLimit ?? SEMANTIC_LAYER_DEFAULT_QUERY_LIMIT,
```

packages/frontend/src/features/semanticViewer/components/RunSemanticQueryButton.tsx
```diff
@@ -100,20 +115,69 @@ export const RunSemanticQueryButton: FC = () => {
         sortBy,
     ]);
 
+    const handleSubmit = useCallback(
+        () =>
+            runSemanticViewerQuery({
+                ...allSelectedFieldsByKind,
+                sortBy,
```
almeidabbm, line 122: Missing limit here, when I change the limit I don't see it in the request payload

e.g. here I set the limit to 3 and I got all results

image


```suggestion
                sortBy,
                limit,
```

almeidabbm: Also on line 96 when we create the new transformer, should we be passing limit in the query?

almeidabbm (commented):
Couple minor things, otherwise looks super!! :D

packages/common/src/types/semanticLayer.ts
```diff
@@ -112,14 +112,35 @@ export interface SemanticLayerTransformer<
     sqlToString: (sql: SqlType) => string;
 }
 
+const SEMANTIC_LAYER_MAX_QUERY_LIMIT = 500;
```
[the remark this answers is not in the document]
IrakliJani: I see, now it makes sense.

almeidabbm (approved):

Issue #11218, opened 2024-08-21
I should be able to choose which dashboard tabs get included in my scheduled delivery


PriPatel:
-Insert sketches here-

(pr already in progress, so closing this )

PR #11266, opened 2024-08-26
feat: allow select tabs for scheduled delivery

Shifan-Gu:
Closes: 

### Description:

This change allows user to select the tabs of the contents which they want include in the scheduled delivery.

UX:
When include all checkbox is selected, select tabs dropdown would be disabled and scheduled delivery will include all tabs.
Screenshot 2024-08-26 at 14 35 01
When include all checkbox is not selected, then user can select the tabs they want include in the scheduled delivery
Screenshot 2024-08-26 at 14 35 12
Then scheduled delivery will filter content based on the tabs selected.
Screenshot 2024-08-26 at 14 35 29

### Reviewer actions


hamzahc1:
@Shifan-Gu do you have the link to the ticket that this PR relates to?

PriPatel:
@hamzahc1 The work for this [ticket](https://github.com/lightdash/lightdash/issues/11218) started before I finished speccing the milestone, so there is no spec for this. I can reopen the ticket. 

The spec was agreed over a figma file. If I can get a render environment, I'll be able to test this and make sure everything works as expected. 

All other changes in the future will have well-specced tickets. this is just an exception!

ZeRego:
@PriPatel 
PR env: https://lightdash-pr-11336.onrender.com/login 
Reminder that emails are in a sandbox and not sent to your email inbox. Let me know if you need help with this.

PriPatel (commented):
Screenshot 2024-09-02 at 14 26 12

- [ ] The checkbox should be selected by default. 
- [ ] Change `Dashboard Tabs` to `Tabs`

Shifan-Gu:
> Screenshot 2024-09-02 at 14 26 12
> * [ ]  The checkbox should be selected by default.
> * [ ]  Change `Dashboard Tabs` to `Tabs`
> * [ ]  !IMPORTANT : If i dont have multiple dashboard tabs, I should not see this option in the form at all. Right now, this is affecting users who don't have multiple tabs and it is confusing

Thank you for the comments! @PriPatel, i will update it.

packages/backend/src/database/migrations/20240822143117_add_scheduler_selected_tabs_column.ts
```diff
@@ -0,0 +1,14 @@
+import { Knex } from 'knex';
+import { SchedulerTableName } from '../entities/scheduler';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(SchedulerTableName, (table) => {
+        table.jsonb('selected_tabs');
```
rephus, line 6: can we specifiy this is `nullable` ?

rephus: also, it is probably better to use a `                    .specificType('selected_tabs', 'TEXT[]')` type instead of jsonb

packages/backend/src/database/entities/scheduler.ts
```diff
@@ -30,6 +30,7 @@ export type SchedulerDb = {
     thresholds: string | null;
     enabled: boolean;
     notification_frequency: string | null;
+    selected_tabs: string | null;
```
rephus, line 33: this does not match with the jsonb type on db, specified on the migration. I'd use `string[] | null` here and `text[]` type on db

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -191,6 +192,12 @@ export default class SchedulerTask {
                               JSON.stringify(sendNowSchedulerFilters),
                           )}`
                         : ''
+                }${
+                    selectedTabs
+                        ? `?selectedTabs=${encodeURI(
+                              JSON.stringify(selectedTabs),
```
rephus, line 198: you don't need to do stringify if you use text[] in db

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -235,13 +240,39 @@ const SchedulerForm: FC<Props> = ({
     isThresholdAlert,
     itemsMap,
 }) => {
+    const isDashboard = resource && resource.type === 'dashboard';
+    const isDashboardTabsEnabled = useFeatureFlagEnabled(
```
rephus, line 244: 👍 nice

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -235,13 +240,39 @@ const SchedulerForm: FC<Props> = ({
...
+        dashboard?.tabs !== undefined && dashboard.tabs.length > 0;
+
     const form = useForm({
         initialValues:
             savedSchedulerData !== undefined
-                ? getFormValuesFromScheduler(savedSchedulerData)
+                ? getFormValuesFromScheduler({
+                      ...savedSchedulerData,
+                      ...(isDashboardScheduler(savedSchedulerData) && {
+                          selectedTabs: isDashboardTabsAvailable
```
rephus, line 260: we can probably extract all this logic in a variable outside the method call, just for readability.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -770,6 +805,69 @@ const SchedulerForm: FC<Props> = ({
...
+                                                    marginBottom: '-4px',
+                                                }}
+                                            />
+                                        </Tooltip>
+                                    </Input.Label>
+                                    <Checkbox
+                                        size="xs"
+                                        label="Include all tabs"
+                                        labelPosition="right"
+                                        checked={allTabsSelected}
```
rephus, line 836: this should be enabled by default, right ?

rephus (changes requested):
I think we can improve the types on db, replace jsonb with string[].
The rest looks good, once the types are fixed I'll test this on render

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
Shifan-Gu: yes, but this is not done by setting a default value for "allTabsSelected", but instead by setting initial the form value "selectedTabs" to include all the tabs, then dynamically derive the value from this form value.

packages/backend/src/scheduler/SchedulerTask.ts
Shifan-Gu: i removed the stringify call in the model layer, but keep this call because it seems a straightforward way to encode the string array and decode it at the endpoint.

rephus: Ok, that works, another option is to pass an array using the query parameter multiple times like. 

`?selectedTabs=tab0&selectedTabs=tab1...`

Both ways are accepted

packages/backend/src/models/SchedulerModel/index.ts
```diff
@@ -317,6 +323,10 @@ export class SchedulerModel {
                         : null,
                     notification_frequency:
                         scheduler.notificationFrequency || null,
+                    selected_tabs:
+                        'selectedTabs' in scheduler && scheduler.selectedTabs
+                            ? (scheduler.selectedTabs as string[])
```
rephus, line 328: is this casting needed here ?

rephus (changes requested):
The code look ok, but when I was testing, I found some issues. 

### CSV export 
- not selecting any tab in the csv export, sends an email with nothing in it. Ideally we should prevent the scheduler from being created in the first place... This is a minor bug , which means we could merge this PR without this, and create a separate issue to fix it later. 

Screenshot from 2024-09-24 09-20-37

### Image export 

- It seems the tab filter is not working, if I try to filter a tab, it always shows me everything. I tested /minimal and hte parameter seems to work, perhaps we are not passing the paramenter correctly inside the headless browser. 
Screenshot from 2024-09-24 09-25-58
image
- selecting `no tabs` simply returns all tabs, I believe this is related with the previous point, where we don't do any filtering. Similar to the CSV UI, we should not allow the scheduler to be created. 


I also tested this without tabs, and it works as expected, so it doesn't break dashboards without tabs :+1: that's great.

Shifan-Gu:
> The code look ok, but when I was testing, I found some issues.
> 
> ### CSV export
> * not selecting any tab in the csv export, sends an email with nothing in it. Ideally we should prevent the scheduler from being created in the first place... This is a minor bug , which means we could merge this PR without this, and create a separate issue to fix it later.
> 
> Screenshot from 2024-09-24 09-20-37
> 
> ### Image export
> * It seems the tab filter is not working, if I try to filter a tab, it always shows me everything. I tested /minimal and hte parameter seems to work, perhaps we are not passing the paramenter correctly inside the headless browser.
>   Screenshot from 2024-09-24 09-25-58
>   image
> * selecting `no tabs` simply returns all tabs, I believe this is related with the previous point, where we don't do any filtering. Similar to the CSV UI, we should not allow the scheduler to be created.
> 
> I also tested this without tabs, and it works as expected, so it doesn't break dashboards without tabs 👍 that's great.

Hi @rephus thanks for the testing work! 

I have some issue with my local env to reproduce the issue, so haven't figure out the root cause for the issue. but i have a question for you is that, are you using standalone scheduler deployment to test this change?

rephus (approved):
Nice work

Issue #11085, opened 2024-08-08
Remove `formatted` field from the SQL Runner api results


owlas:
The backend currently returns `value` and `formatted` versions of the results but we're not using the `formatted` data - it's unnecessary work for the backend.

PR #11268, opened 2024-08-27
chore: remove formatted from rows

joaoviana:
Closes: [#11085](https://github.com/lightdash/lightdash/issues/11085)

### Description:

Removes formatting of rows. Now, we don't access the row value by the `formatted` key, but with the direct value instead.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2008,37 +2008,33 @@ export class ProjectService extends BaseService {
 
         const columns: VizSqlColumn[] = [];
 
-        const fileUrl = await this.downloadFileModel
-            .streamFunction(this.s3Client)(
-                `${this.lightdashConfig.siteUrl}/api/v1/projects/${projectUuid}/sqlRunner/results`,
-                async (writer) => {
-                    await warehouseClient.streamQuery(
-                        query,
-                        async ({ rows, fields }) => {
-                            if (!columns.length) {
-                                // Get column types from first row of results
-                                columns.push(
-                                    ...Object.keys(fields).map((fieldName) => ({
-                                        reference: fieldName,
-                                        type: fields[fieldName].type,
-                                    })),
-                                );
-                            }
+        const fileUrl = await this.downloadFileModel.streamFunction(
+            this.s3Client,
+        )(
+            `${this.lightdashConfig.siteUrl}/api/v1/projects/${projectUuid}/sqlRunner/results`,
+            async (writer) => {
+                await warehouseClient.streamQuery(
+                    query,
+                    async ({ rows, fields }) => {
+                        if (!columns.length) {
+                            // Get column types from first row of results
+                            columns.push(
+                                ...Object.keys(fields).map((fieldName) => ({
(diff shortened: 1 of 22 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2008,37 +2008,33 @@ export class ProjectService extends BaseService {
...
-                            formattedRows.forEach(writer);
-                        },
-                        {
-                            tags: queryTags,
-                        },
-                    );
-                },
-                this.s3Client,
-            )
-            // catch any errors
```
joaoviana, line 2038: removed unnecessary catch+throw

packages/frontend/src/features/semanticViewer/components/RunSemanticQueryButton.tsx
```diff
@@ -24,25 +23,6 @@ import {
 import { setLimit, setResults } from '../store/semanticViewerSlice';
 import { SemanticViewerResultsTransformer } from '../transformers/SemanticViewerResultsTransformer';
 
-const mapResultsToTableData = (
```
owlas, line 27: Let''s go!!!!

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -59,6 +60,16 @@ export const getRawValueCell = (
     return <span>{`${raw}`}</span>;
 };
 
+export const getValueCell = (
```
owlas, line 63: This looks like the 'formatter' method for the data. Lets abstract this out into a new file explicitly for formatting values from the warehouses, rather than having it buried in here with the `useColumns` info

joaoviana: 🤔 but this is only used when declaring `tanstack` table columns. Above you can see other similar methods like: `getFormattedValueCell` and `getRawValueCell`

Should I add it to be close the shared `DataViz` folder?

image


(btw I can't place it in `common` because of the types and it being out of the `FE-context`)

packages/common/src/visualizations/types/index.ts
```diff
@@ -58,12 +59,9 @@ export type VizPieChartDisplay = {
     isDonut?: boolean;
 };
 
-// ! TODO: Rename
-export type RowData = Record<string, unknown>;
-
 // TODO: what is `type`?
```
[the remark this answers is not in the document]
joaoviana: done here! https://github.com/lightdash/lightdash/pull/11268/commits/f9d2904714151fd3a0a519294e6adb939da812a8

packages/frontend/src/hooks/useColumns.tsx
owlas: Oh yeah I meant something like:

```
export const getValueCell = (tanstackStuff) => {
  const value = info.getValue()
  const formatted = format(value)
   return <span>{ formatted }</span>;
}
```

That way the function is all tanstack specific. But the formatting logic lives with the `dataviz` stuff

joaoviana: got itttttt

owlas (approved):

PR #10740, opened 2024-07-16
feat: run query on cmd+enter

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/10738

### Description:

allows to run a query with cmd+enter outside and inside the sql editor

Followed this suggestion to get to a working solution: https://github.com/suren-atoyan/monaco-react/issues/328#issuecomment-1079967898

demo:

https://github.com/user-attachments/assets/18fc4a0f-1756-4f2b-8b7b-3a69f2a0a5d3

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #11281, opened 2024-08-27
fix: allow cmd+enter when focusing sql editor

joaoviana:
Closes: N/A

### Description:

Bring back logic in `onMount` from here: https://github.com/lightdash/lightdash/pull/10740/files

To test: 

Click on Sql Editor 
use `Cmd`+`Enter` shortcut
See query run 

THis also works when the sql editor isn't focused.

demo:

https://github.com/user-attachments/assets/1dd89540-08c0-43ae-98ed-2b7616f4037c

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -132,7 +132,7 @@ export const ContentPanel: FC = () => {
         [
             'mod + enter',
             () => {
-                if (sql) runSqlQuery({ sql, limit: 7 });
+                if (sql) runSqlQuery({ sql, limit });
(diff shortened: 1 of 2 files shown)


andrewdw:
This looks good but shouldn't it be called "CMD+Enter"? The behaviour is that it submits on CMD + ENTER (which is consistent with other editors) instead of CMD + Space.

joaoviana:
> This looks good but shouldn't it be called "CMD+Enter"? The behaviour is that it submits on CMD + ENTER (which is consistent with other editors) instead of CMD + Space.

Why did I write `space` 🤦 thank you for pointing it out!

andrewdw (approved):
Works great! 🎉

Issue #11233, opened 2024-08-21
Handle SQL runner limit and improve the results table view


joaoviana:
### Description

Hide SQL Runner limit on the button.

Always show a `500` limit on the results table, unless the user applies a limit to it which will be lower(?) - this should be clear in the table! (e.g. `preview 500 rows` or something)

Allow for fine-tuning this limit on the visualization.

The Chart results table should mimic what's being chosen/configured on the chart. For example:
- I select 10 columns with my SQL Query
- But I only choose 3 to build my chart
- Then I should only see these on my chart's table

sanjay-sol:
Hey @joaoviana , Can I work on this ?

joaoviana:
thank you for your interest @sanjay-sol, but this is a task I'll be taking soon. You're more than welcome to look into other issues here: https://github.com/lightdash/lightdash/issues?q=is%3Aopen+is%3Aissue+label%3A%22%F0%9F%91%8B+open-contribution%22 😄

Issue #11210, opened 2024-08-20
Sql Runner charts should show pivoted results below


owlas:
The results that power a chart should be shown below the chart (since they can be different to the results table from the query). 

This is on the SQL runner page and the saved chart in view mode.

PR #11291, opened 2024-08-28
feat: remove limit and display chart results in chart viz

joaoviana:
Relates to: [#11233](https://github.com/lightdash/lightdash/issues/11233)

### Description:

- Removed `limit` override in button
- Allows for charts to have their own `results` table 

> [!NOTE]
> Don't have the `Allow fine-tuning the limit on the visualization` - this will be in a separate PR

Demo: 

https://github.com/user-attachments/assets/4d11f70a-11ba-40e0-9698-0935a58213e9

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2065,16 +2065,20 @@ export class ProjectService extends BaseService {
             ),
         ];
         const groupByQuery = `SELECT ${[
-            ...groupBySelectDimensions,
+            ...new Set(groupBySelectDimensions), // Remove duplicate columns
             ...groupBySelectMetrics,
-        ].join(
-            ', ',
-        )} FROM original_query group by ${groupBySelectDimensions.join(', ')}`;
+        ].join(', ')} FROM original_query group by ${Array.from(
+            new Set(groupBySelectDimensions),
+        ).join(', ')}`;
 
         const selectReferences = [
             indexColumn.reference,
             ...(groupByColumns || []).map((col) => col.reference),
-            ...(valuesColumns || []).map((col) => col.reference),
+            ...(valuesColumns || []).map((col) =>
+                groupByColumns && groupByColumns.length > 0
+                    ? `${col.reference}_${col.aggregation}`
+                    : `${col.reference}_${col.aggregation} as ${col.reference}`,
+            ),
         ];
 
         const pivotQuery = `SELECT ${selectReferences.join(
@@ -2204,7 +2208,7 @@ export class ProjectService extends BaseService {
                                 currentTransformedRow =
                                     currentTransformedRow ?? {};
                                 currentTransformedRow[valueColumnReference] =
-                                    row[col.reference];
+                                    row[`${col.reference}_${col.aggregation}`];
                             });
                         });
                     },
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2061,7 +2061,7 @@ export class ProjectService extends BaseService {
         const groupBySelectMetrics = [
             ...(valuesColumns ?? []).map(
                 (col) =>
-                    `${col.aggregation}(${col.reference}) as ${col.reference}`,
+                    `${col.aggregation}(${col.reference}) as ${col.reference}_${col.aggregation}`,
             ),
         ];
         const groupByQuery = `SELECT ${[
@@ -2222,7 +2222,9 @@ export class ProjectService extends BaseService {
             fileUrl,
             valuesColumns: groupByColumns
                 ? Array.from(valuesColumnReferences)
-                : valuesColumns.map((col) => col.reference),
+                : valuesColumns.map(
+                      (col) => `${col.reference}_${col.aggregation}`,
+                  ),
             indexColumn,
         };
     }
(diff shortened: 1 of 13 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2061,20 +2061,24 @@ export class ProjectService extends BaseService {
...
         ];
         const groupByQuery = `SELECT ${[
-            ...groupBySelectDimensions,
+            ...new Set(groupBySelectDimensions), // Remove duplicate columns
             ...groupBySelectMetrics,
-        ].join(
-            ', ',
-        )} FROM original_query group by ${groupBySelectDimensions.join(', ')}`;
+        ].join(', ')} FROM original_query group by ${Array.from(
+            new Set(groupBySelectDimensions),
```
joaoviana, line 2071: ensures there are no `column X is ambiguous` error

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2061,20 +2061,24 @@ export class ProjectService extends BaseService {
...
+        ).join(', ')}`;
 
         const selectReferences = [
             indexColumn.reference,
             ...(groupByColumns || []).map((col) => col.reference),
-            ...(valuesColumns || []).map((col) => col.reference),
+            ...(valuesColumns || []).map((col) =>
+                groupByColumns && groupByColumns.length > 0
+                    ? `${col.reference}_${col.aggregation}`
+                    : `${col.reference}_${col.aggregation} as ${col.reference}`,
```
joaoviana, line 2080: if it's pivoted, revert the aggregation - open to other solutions!

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2222,7 +2226,9 @@ export class ProjectService extends BaseService {
             fileUrl,
             valuesColumns: groupByColumns
                 ? Array.from(valuesColumnReferences)
-                : valuesColumns.map((col) => col.reference),
+                : valuesColumns.map(
+                      (col) => `${col.reference}_${col.aggregation}`,
```
joaoviana, line 2230: so that we have a unique identifier for a column that has an aggregation applied. 

For example:
- event
- event_sum

allows us to use both columns but on different axes, on different series, etc

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -210,6 +210,50 @@ export const ContentPanel: FC = () => {
         [currentVizConfig],
     );
 
+    const {
+        tableConfigByChartType,
+        resultsTableRunnerByChartType,
+        handlePivotData,
+    } = useChartResultsTableConfig(resultsRunner, activeConfigs);
```
joaoviana, line 217: this gets the Line + Bar + Pie charts' table configs. So that we get only the columns for the configured chart. 
For example: 
I choose one x, and on y (sum)

I should only see a table with x and another with y (sum), not all the columns like in the results table.

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -210,6 +210,50 @@ export const ContentPanel: FC = () => {
...
+        activeEditorTab,
+        activeConfigs.chartConfigs,
+        selectedChartType,
+    ]);
+
+    const showChartResultsTable = useMemo(() => {
+        return !!(
+            queryResults?.results &&
+            activeEditorTab === EditorTabs.VISUALIZATION &&
+            // if the chart is not pivoted, show the chart results table
```
joaoviana, line 245: we don't have pivoted tables yet, so this will do - the ticket to display this is here: https://github.com/lightdash/lightdash/issues/11210

rephus:
Does this `Close: https://github.com/lightdash/lightdash/issues/11233` ?

joaoviana:
> Does this Close: #11233 ?

Not fully, added to description https://github.com/lightdash/lightdash/pull/11291#issue-2491798482 the missing AC for this

rephus:
Is this weird table header expected ? 

image


Side question: Don't we have a link to share sql runner anymore ? I believe I implemented that some time ago 
https://lightdash-pr-11291.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/sql-runner

rephus:
Any idea why this is throwing errors ? 

[Screencast from 29-08-24 15:22:34.webm](https://github.com/user-attachments/assets/422f288b-1a08-4246-a788-51d23195f072)

joaoviana:
> Is this weird table header expected ?


REnder hasn't fully deployed yet, changes should be there soon and the table layout should be correct / similar to the results table! 😮‍💨 


> Side question: Don't we have a link to share sql runner anymore ? I believe I implemented that some time ago https://lightdash-pr-11291.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/sql-runner

Can't remember exactly, maybe bc the link was generating a large JSON which we don't want? We intend to save the draft state into the db etc IIRC.

joaoviana:
> Any idea why this is throwing errors ?
> 

Same issue as above, REnder isn't fully updated yet

joaoviana:
> Confirmed, the previous chart is working now.
> 
> Using first aggregation doesn't work though, not sure if this is expected
> 
> image

good point, checking if I can fix that - could be an issue with the SQL query generated, which if it's too much effort, I'm happy to disable this aggregation

rephus:
I'm not sure if this is supposed to be working on semantic viewer. But Even if it doesn't work, I don't think this is a blocker for merging this PR 

image

PR #11295, opened 2024-08-28
feat: change colors sql runner

rephus:
Closes: 

### Reviewer actions

### DIFF
--- a/packages/common/src/visualizations/CartesianChartDataTransformer.ts
+++ b/packages/common/src/visualizations/CartesianChartDataTransformer.ts
@@ -156,10 +156,10 @@ export class CartesianChartDataTransformer<TPivotChartLayout> {
                               )
                             : undefined,
                     },
-                    color: this.getSeriesColor(
-                        seriesColumn,
-                        possibleXAxisValues,
-                    ),
+                    color:
+                        (display?.series &&
+                            display.series[seriesColumn]?.color) ||
+                        this.getSeriesColor(seriesColumn, possibleXAxisValues),
                 };
             },
         );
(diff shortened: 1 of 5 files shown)


packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -257,10 +274,14 @@ export type CartesianChartDisplay = {
         position?: string;
         format?: Format;
     }[];
-    series?: Record<
```
rephus, line 260: I had to change this Record to be able to get the colors on the backend, I think this is related with tsoa not liking Record types and removing some content in the payload

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -176,10 +189,14 @@ export class CartesianChartDataModel
...
-                        possibleXAxisValues,
-                    ),
+                    color:
+                        (display?.series &&
+                            display.series[seriesColumn]?.color) ||
+                        CartesianChartDataModel.getDefaultColor(
+                            index,
+                            orgColors,
+                        ),
+                    // this.getSeriesColor( seriesColumn, possibleXAxisValues, orgColors),
```
joaoviana, line 199: do we need this then? can remove the comment, right?

joaoviana (approved):
Works great! 

Left some considerations on layout and on one Q on commented code

packages/common/src/visualizations/CartesianChartDataModel.ts
rephus: I replace this method with a new one. But the old one has some specific logic for saving so I was not sure if we should remove it

Issue #11276, opened 2024-08-27
For some tables, the schema appears to be empty


PriPatel:
<img width="435" alt="Screenshot 2024-08-27 at 13 54 29" src="https://github.com/user-attachments/assets/fa7a0475-56bb-48e8-92a8-b8ae83bd65aa">

This works for some tables, but not for others.

neehhaa06:
Hello @PriPatel 

Would you explain me a little bit, what is bug in this problem?

PR #11313, opened 2024-08-30
fix: sql runner fields on different schema

rephus:
Closes:  #11276

### Description:

Before:
Screenshot from 2024-08-30 09-42-26

Now: 

Screenshot from 2024-08-30 09-42-37

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -3080,6 +3080,7 @@ export class ProjectService extends BaseService {
         user: SessionUser,
         projectUuid: string,
         tableName: string,
+        schema?: string,
     ): Promise<WarehouseTableSchema> {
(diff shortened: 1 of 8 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3107,10 +3108,11 @@ export class ProjectService extends BaseService {
             project_uuid: projectUuid,
             user_uuid: user.userUuid,
         };
-        const schema = ProjectService.getWarehouseSchema(credentials);
+        const warehouseSchema =
```
rephus, line 3111: This was the main issue, we were assuming all tables were from the same `schema`, (eg: jaffle) but that is not correct if you select a table in a different schema, so I pass schema as a query parameter.

joaoviana (approved):
Tested a before and after with the `public` schema! works well 👍

Issue #11139, opened 2024-08-14
Add average as chart aggregation option


magnew:
### Description

Add average as chart aggregation option in SQL runner charts.

PR #11314, opened 2024-08-30
feat: avg aggregation sql runner

rephus:
Closes:   #11139

### Description:

Screenshot from 2024-08-30 14-54-59

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/SemanticLayerService/Pivoting.ts
+++ b/packages/backend/src/services/SemanticLayerService/Pivoting.ts
@@ -25,7 +25,12 @@ export function pivotResults(
         return results;
     }
 
-    const aggs: pl.Expr[] = values.map((v) => pl.col(v.name)[v.aggFunction]());
+    const aggs: pl.Expr[] = values.map((v) => {
+        if (v.aggFunction === 'avg') {
+            return pl.col(v.name).mean();
+        }
+        return pl.col(v.name)[v.aggFunction]();
+    });
 
     return df
         .groupBy(dimensionsToGroupBy)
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/DataViz/config/DataVizAggregationConfig.tsx
```diff
@@ -30,6 +30,7 @@ const AggregationIcon: FC<{ aggregation: string | undefined }> = ({
             icon = IconSum;
             break;
         case MetricType.AVERAGE:
+        case VizAggregationOptions.AVERAGE:
             icon = IconTrendingUp;
```
rephus, line 34: I don't think this is a great icon for `average` , but I can't find a find a better one to replace it

joaoviana: [IconMathAvg](https://tabler.io/icons/icon/math-avg)

This apparently 😅

rephus: You think ? that doesn't look right to me..

joaoviana: I mean the name of the icon has avg, but I don't see it being widely used as **the** symbol for average. what you have works for me, checking the code!

joaoviana (approved):
Just needs a type fix, but functionality works great!

packages/backend/src/services/SemanticLayerService/Pivoting.ts
```diff
@@ -44,9 +49,15 @@ export const groupResults = (
 ): SemanticLayerResultRow[] => {
     const df = pl.DataFrame(results);
 
-    const aliasedAggs: pl.Expr[] = values.map((v) =>
-        pl.col(v.name)[v.aggFunction]().alias(`${v.aggFunction}(${v.name})`),
-    );
+    const aliasedAggs: pl.Expr[] = values.map((v) => {
```
owlas, line 52: @almeidabbm and @IrakliJani why is this still here? I think it's confusing, we don't need this aggregation and developers shouldn't need to worry about it.

PR #11317, opened 2024-08-30
chore: handle script error + tech debt 

ZeRego:
### Description:

Changes:
- fetch user instead of user session (lighter method)
- amend method to update password to receive a user uuid instead of a user id.
- display errors nicely

Demo changing user password:

Screenshot 2024-08-30 at 16 14 09

Error before:

Screenshot 2024-08-30 at 16 15 30

Error after:
Screenshot 2024-08-30 at 16 16 52

### Reviewer actions

### DIFF
--- a/packages/backend/src/overrideUserPassword.ts
+++ b/packages/backend/src/overrideUserPassword.ts
@@ -6,35 +6,42 @@ import Logger from './logging/logger';
 import { UserModel } from './models/UserModel';
 
 (async function init() {
-    Logger.warn(`Override user password`);
-    const email: string | undefined = process.argv[2];
-    const newPassword: string | undefined = process.argv[3];
+    try {
+        Logger.warn(`Override user password`);
+        const email: string | undefined = process.argv[2];
+        const newPassword: string | undefined = process.argv[3];
 
-    if (!email) {
-        throw new Error('Email is undefined');
-    }
+        if (!email) {
+            throw new Error('Email is undefined');
+        }
 
-    if (!newPassword) {
-        throw new Error('New password is undefined');
-    }
-    const environment =
-        process.env.NODE_ENV === 'development' ? 'development' : 'production';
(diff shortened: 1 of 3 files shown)


packages/backend/src/overrideUserPassword.ts
```diff
@@ -6,35 +6,42 @@ import Logger from './logging/logger';
...
+                    : knexConfig.development,
+            ),
+        });
 
-    Logger.info(`Get user by email: ${email}`);
-    const user = await userModel.findSessionUserByPrimaryEmail(email);
-    if (user === undefined) {
-        throw new NotFoundError(`Cannot find user with uuid ${email}`);
+        Logger.info(`Get user by email: ${email}`);
+        const user = await userModel.findUserByEmail(email);
```
ZeRego, line 35: Main change was this method and wrapping everything in a try catch.

packages/backend/src/overrideUserPassword.ts
```diff
@@ -6,35 +6,42 @@ import Logger from './logging/logger';
 import { UserModel } from './models/UserModel';
 
 (async function init() {
-    Logger.warn(`Override user password`);
-    const email: string | undefined = process.argv[2];
-    const newPassword: string | undefined = process.argv[3];
+    try {
+        Logger.warn(`Override user password`);
```
rephus, line 10: I didn't know about this script. is this documented somewhere ?

rephus (approved):

packages/backend/src/overrideUserPassword.ts
ZeRego: In our public docs. 
Useful script for self hosted instances without a SMTP. So admins can reset passwords for users.
https://docs.lightdash.com/self-host/customize-deployment/use-sso-login-for-self-hosted-lightdash#passwords

PR #11324, opened 2024-09-02
fix: display error message if failed

joaoviana:
Closes: N/A

### Description:

Now that we get the pivot results through an API endpoint, we should get the error's `message` correctly to display under the `ChartView` `SuboptimalState`

How to test: 

- Select * any table
- Run query
- Select * another table
- See error states in chart configuration + `error generating chart` with a description from the Api error.

Similar screenshot:

Screenshot 2024-09-02 at 10 13 19

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DataViz/visualizations/ChartView.tsx
+++ b/packages/frontend/src/components/DataViz/visualizations/ChartView.tsx
@@ -60,8 +60,6 @@ const ChartView = memo(
             uuid,
         });
 
-        console.log({ error });
-
         if (!config?.fieldConfig?.x || config?.fieldConfig.y.length === 0) {
             return (
                 <SuboptimalState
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/DataViz/visualizations/ChartView.tsx
```diff
@@ -75,11 +76,9 @@ const ChartView = memo(
         }
         const loading = isLoadingProp || transformLoading;
 
-        // TODO: this could be more robust
-        const errorMessage = error?.message?.includes('Binder Error')
-            ? 'Some specified columns do not exist in the data'
+        const errorMessage = isApiError(error)
```
owlas, line 79: This error type checking should be deeper. At least in useChart or maybe deeper but that might need type changes. The `Error` type comes from `useAsync`

joaoviana: done here! https://github.com/lightdash/lightdash/pull/11324/commits/3d886839911049d468e5c50ad846d12775abd639

owlas (approved):

Issue #11335, opened 2024-09-02
Rename cancel button when editing dashboard chart


ZeRego:
### Description

> If i have gone to edit chart from a dash and made changes and saved them, it makes me think twice seeing Cancel as its indicating I will cancel my edits when it fact its just going to take me back to the dash

### Steps to Reproduce the Bug or Issue

1. create dashboard + chart in dashboard + save
2. click "edit chart"
3. make a change + save

Expected:
button in banner should say "Return to dashboard"

Current: 
button in banner says "Cancel" 

### Cloud or self-hosting

None

sanjay-sol:
Hey @ZeRego , I would love to work on this issue!

sanjay-sol:
Hey @ZeRego , What should be the TooltipText for the "Return to Dashboard" button ?

ZeRego:
@sanjay-sol We can use "Conclude chart editing and return to dashboard".

Ankush263:
Hello, I can contribute here, I'll raise a PR.

chidexebere:
Hello @ZeRego , 
this issue seems to still be open for contribution?  Can this be assigned to me?

ZeRego:
Hi @chidexebere 👋 
Thanks for reaching out. I removed the open-contribution tag as after reviewing 2 external pull requests we recognised this involves more changes than we anticipated.

PR #11346, opened 2024-09-02
fix: Rename cancel button tooltip for editing and viewing actions

sanjay-sol:
Closes: #11335 

### Description:
This pull request updates the button text in the dashboard banner when editing charts. Previously, the button displayed "Cancel" which might have been confusing, as it indicated that changes would be canceled, while it actually navigates back to the dashboard.

Before:
Screenshot 2024-09-02 at 8 36 02 PM

After:
Screenshot 2024-09-02 at 8 21 31 PM

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
+++ b/packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
@@ -38,9 +38,9 @@ export const DashboardExplorerBanner: FC<Props> = ({ projectUuid }) => {
     const cancelButtonText = useMemo(() => {
         switch (action) {
             case 'viewing':
+            case 'editing':
                 return 'Return to dashboard';
             case 'creating':
-            case 'editing':
                 return 'Cancel';
             default:
                 return assertUnreachable(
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/NavBar/DashboardExplorerBanner.tsx
```diff
@@ -38,9 +38,9 @@ export const DashboardExplorerBanner: FC<Props> = ({ projectUuid }) => {
     const cancelButtonText = useMemo(() => {
         switch (action) {
             case 'viewing':
+            case 'editing':
                 return 'Return to dashboard';
```
ZeRego, line 42: We are missing a bit of logic as we only want this if there are no unsaved changes. 

eg:
1. page load = Return to dashboard
2. I add a new metric = Cancel
3. I click save = Return to dashboard
4. I change the chart config = Cancel 
5. I save = Return to dashboard

sanjay-sol: Ohh !! Got it !

ZeRego:
Closing this PR for now, feel free to reopen when ready :)

Issue #11322, opened 2024-09-02
I'm seeing error icons next to most of my content, but the validator isn't showing any errors


PriPatel:
<img width="866" alt="Screenshot 2024-09-02 at 10 00 59" src="https://github.com/user-attachments/assets/3f98b5cf-c10a-49ba-8e06-e7eed552e2a1">
<img width="882" alt="Screenshot 2024-09-02 at 10 01 04" src="https://github.com/user-attachments/assets/95eb06a3-b98f-4ebf-84ba-47efeaa51199">

This affects most content in our Analytics project

jpetey75:
I don't know why this is happening but I can confirm it also happens in the Healthcare Demo project under the Lightdash org.

PR #11361, opened 2024-09-03
fix: dont display validation errors originated from CLI

ZeRego:
Closes: [#11322](https://github.com/lightdash/lightdash/issues/11322)

### Description:

When we run the validation CLI command, we store the errors in the database under a job id. This is because we may have multiple developers running validation on different branches. The CLI will check the job status every X seconds. Once it is ready, it fetches the validation errors via the job id. 

When we run the validation after a project is deployed or via the UI, we store the errors in the database without a job id (null). This is because there should only be 1 validation state for the project inside Lightdash. 

The validation page accurately fetches the validation state by filtering the errors where the job id is null.
The other pages that list chart/dashboard where NOT filtering the errors where the job id is null. This resulted in displaying errors related to in progress work from a developer branch. 

Reproduction steps:

1. rename a field in `schema.yml`. eg: order_id
2. run validate command `node ./packages/cli/dist/index.js validate --project-dir ./examples/full-jaffle-shop-demo/dbt --profiles-dir ./examples/full-jaffle-shop-demo/profiles` 

Actual:
See errors in home page, search, chart/dashboard pages

Expected:
No errors 

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -363,6 +363,7 @@ export class DashboardModel {
                                     SELECT json_agg(validations.*)
                                     FROM validations
                                     WHERE validations.dashboard_uuid = ${DashboardsTableName}.dashboard_uuid
+                                    AND validations.job_id IS NULL
                                 ), '[]'
                             ) as validation_errors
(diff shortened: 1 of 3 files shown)


rephus (approved):
this makes sense

Issue #11293, opened 2024-08-28
SQL Runner charts should have an action menu in the charts lists


magnew:
### Description

SQL Runner charts should have an action menu in the charts lists on the content browsing pages. It should at least be possible to:
- move them between spaces
- delete them

And maybe
- add them to dashboards

Other actions might take more infrastructure that we don't yet have around these types of charts

PR #11362, opened 2024-09-03
feat: allow move to space and delete for sql charts

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/11293

### Description:

Screenshot from 2024-09-03 12-13-04

- [ ] The rest of the actions are hidden 

- rename : needs new modal component
- add to dashboard: needs modifying hooks on add to dashboard component
- pin : needs new endpoint
- promote: needs new endpoint

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
+++ b/packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
@@ -129,16 +129,30 @@ const ResourceViewActionMenu: FC<ResourceViewActionMenuProps> = ({
         }),
     );
 
+    const isSqlChart =
+        item.type === ResourceViewItemType.CHART &&
+        item.data.source === ChartSourceType.SQL;
+
     switch (item.type) {
         case ResourceViewItemType.CHART: {
-            if (item.data.source === ChartSourceType.SQL) {
-                // Actions are not currently supported on SQL charts
-                // TODO implement
-                return null;
-            }
             const userAccess = spaces.find(
                 (space) => space.uuid === item.data.spaceUuid,
             )?.userAccess;
+
+            if (
+                isSqlChart &&
+                user.data?.ability?.cannot(
+                    'manage',
+                    subject('SqlRunner', {
+                        organizationUuid: user.data?.organizationUuid,
+                        projectUuid,
+                        access: userAccess ? [userAccess] : [],
+                    }),
+                )
+            ) {
+                return null;
+            }
+
             if (
                 user.data?.ability?.cannot(
                     'manage',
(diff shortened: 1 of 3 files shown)


joaoviana (changes requested):
in `ResourceActionMenu` L~365, there's a `Menu.Divider`, can you add this?
```tsx
 <Menu.Divider
                                display={isSqlChart ? 'none' : 'block'}
                            />
```

joaoviana (changes requested):
Moving to an existing space works ✅ 

Creating a space on the menu item + moving to it doesn't. 
Wonder if this is an existing bug though, checking

EDIT: Correct, this is a bug unrelated to this work, so I will ignore this in the review: https://github.com/lightdash/lightdash/issues/11363

packages/frontend/src/components/common/ResourceView/ResourceActionMenu.tsx
```diff
@@ -229,6 +243,7 @@ const ResourceViewActionMenu: FC<ResourceViewActionMenuProps> = ({
                                 item,
                             });
                         }}
+                        sx={{ display: isSqlChart ? 'none' : 'block' }}
```
joaoviana, line 246: could we use the `hidden` prop instead? 
```tsx
hidden={isSqlChart}
```

otherwise we have to deal with default styling if it's not a `sql chart`

image

rephus: hidden didn't work to me, let me try again

joaoviana: then we can do `disabled`?

rephus: https://github.com/mantinedev/mantine/issues/5521#issuecomment-1878444848

rephus: disabled without tooltip can be confusing, don't you think ? and I don't want to add many extra tooltips...

rephus: I can do this instead 
`                            sx={isSqlChart ?{ display:  'none'}: {}}`

joaoviana: then we can replace in the ternary the `block` with `flex` 👍

rephus: no need to add any flex, we can leave it empty 

> sx={isSqlChart ?{ display:  'none'}: {}}


image

joaoviana (approved):

rephus:
I'll create a ticket for the missing functions in the context menu (low priority)

Issue #11371, opened 2024-09-03
Reduce the API requests when viewing a non-table SQL chart


joaoviana:
### Description

When viewing a non-table SQL chart (in isolation or in a dashboard tile) we run 2 result jobs: 
1. the SQL query results' job
2. the pivot query results' job

Ensure we first one is not run and only the request `2` is run because that's the only one needed to render a chart.

Check `pivotQueryFn`, the hooks related to these 2 scenarios (`useSqlChartResults` and `useSqlChartAndResults`)

### Steps to Reproduce the Bug or Issue

1. Create SQL chart
2. Save
3. Enter view mode
4. See 2 requests 

(same for dashboard tiles)

### Cloud or self-hosting

cloud

PR #11384, opened 2024-09-04
fix: reduce api requests on chart view dashboard tile

joaoviana:
Closes: #11371 

### Description:

When viewing a non-Table viz dashboard tile, we don't need to get its results because the `ChartView` will deal with it. 
When viewing a table, we can get the results. 

This PR:
- Removed `chart-and-results` endpoint since it's unnecessary now and we reuse other utils from existing hooks. 
- Fetch results job only for `tables`
- Also fixed a width issue with a SQL Table dashboard tile, it's looking good now - was being cropped by overflow.
- Removed `data` prop from `ChartView` wasn't being used anywhere

### Reviewer actions


packages/frontend/src/features/sqlRunner/hooks/useDashboardSqlChart.tsx
```diff
@@ -0,0 +1,74 @@
...
+    savedSqlUuid: string;
+}): Promise<{ resultsAndColumns: ResultsAndColumns; chart: SqlChart }> => {
+    const chart = await fetchSavedSqlChart({
+        projectUuid,
+        uuid: savedSqlUuid,
+    });
+
+    const hasResultsHandledInChart = chart.config.type !== ChartKind.TABLE;
+
+    if (hasResultsHandledInChart) {
```
joaoviana, line 21: this is where we skip the results job @owlas

rephus:
Everything works as expected , I'll wait until the scroll on tables has been fixed and test it again

rephus:
This is probably out of scope, but if I add the same chart multiple times (same query) the request will be duplicated on the backend with the same query body. We should probably cache on react-query or similar somehow. 

This is probably a minor issue, since not many people will add the same chart , even if they do, they will probably contain different filters in the future, so it makes so make a different request. 

image

rephus (approved):
I can see scrollbars now. 

horizontal scrollbar only shows at the bottom. We can fix it on a separate PR 

[Screencast from 04-09-24 12:04:26.webm](https://github.com/user-attachments/assets/1aa72aa4-6903-443f-8f94-adbc68cb66a9)

PR #11387, opened 2024-09-04
fix: check for results and columns length

almeidabbm:
### Description:

Fixes: https://github.com/lightdash/lightdash/issues/11388

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
+++ b/packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
@@ -93,7 +93,10 @@ export const sqlRunnerSlice = createSlice({
             state,
             action: PayloadAction<ResultsAndColumns>,
         ) => {
-            if (!action.payload.results || !action.payload.columns) {
+            if (
+                action.payload.results.length === 0 ||
+                action.payload.columns.length === 0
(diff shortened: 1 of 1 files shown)


IrakliJani (approved):

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -93,7 +93,10 @@ export const sqlRunnerSlice = createSlice({
             state,
             action: PayloadAction<ResultsAndColumns>,
         ) => {
-            if (!action.payload.results || !action.payload.columns) {
+            if (
+                action.payload.results.length === 0 ||
```
joaoviana, line 97: besides `length` do we need to also check if they don't exist at all?

joaoviana (changes requested):
one Q!

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
almeidabbm: from the type that is there they should always exist,checking where they are coming from to understand if that's a possibility

joaoviana (approved):

PR #11398, opened 2024-09-04
fix: wait for sql before viewing sql chart

joaoviana:
Closes: N/A

### Description:

wait for `sql` before trying to display `ChartView` on `ViewSqlChart`. 

To replicate the issue **without the fix**
(keep cache, don't disable it)
- Go to chart A
- Click on breadcrumb to Space
- Go back to chart
image

Because the SQL was empty.
This is in between states and store initialisations. Since `ChartView`'s prop `sql` is intentionally optional, it led to this error.

Now all is well.

### Reviewer actions


rephus (approved):
makes sense, :ship: it !

Issue #10977, opened 2024-08-02
When I go to edit a saved SQL chart, I have to `run query` to open the chart configuration


joaoviana:
### Description

see steps to reproduce. But essentially we should enable the Chart tab when editing a saved SQL chart

### Steps to Reproduce the Bug or Issue

1. Save SQL chart with any type of vis and configure it
2. View it
3. Go to edit it
4. You can see the `Chart` tab/button disabled, and you have to run query to be able to use it

**Note:** the `chart` tab should be what's opened by default when a user clicks `edit` on a SQL runner chart. 

### Cloud or self-hosting

None

PR #11401, opened 2024-09-04
feat: fetch results on load of edit chart

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/10977

### Description:

When on `/edit` pass `isEditMode` 
Then that means that it should fetch results on load. If that is `true`, then select the tab to be `Chart` so that the user can view the chart on load

Screen recording:

https://github.com/user-attachments/assets/7b5a3ff7-fbcf-4fb2-8489-093027211d8f

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -75,14 +75,21 @@ export const ContentPanel: FC = () => {
         [resultsHeight, maxResultsHeight],
     );
 
-    const {
-        projectUuid,
-        sql,
-        limit,
-        activeEditorTab,
-        selectedChartType,
-        resultsTableConfig,
-    } = useAppSelector((state) => state.sqlRunner);
+    const fetchResultsOnLoad = useAppSelector(
+        (state) => state.sqlRunner.fetchResultsOnLoad,
+    );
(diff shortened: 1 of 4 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -287,18 +302,18 @@ export const ContentPanel: FC = () => {
...
                                     {
-                                        value: 'sql',
+                                        value: EditorTabs.SQL,
                                         label: (
                                             <Group spacing="xs" noWrap>
                                                 <MantineIcon
                                                     icon={IconCodeCircle}
                                                 />
-                                                <Text>Query</Text>
+                                                <Text>SQL</Text>
```
rephus, line 311: :+1:

rephus (approved):
Works well

Just 1 question, is it expected to re-run the `runPivotQuery` when going to edit mode ? I thought we were going to reuse the results, so we don't need to make extra requests when going into edit mode from view mode

joaoviana:
> Just 1 question, is it expected to re-run the runPivotQuery when going to edit mode ? I thought we were going to reuse the results, so we don't need to make extra requests when going into edit mode from view mode

Great question @rephus, @IrakliJani is adding caching to this function if I understand it correctly: https://github.com/lightdash/lightdash/pull/11409 so this will be a great perf/request improvement

PR #11404, opened 2024-09-04
fix: result cols mapping and query when fetching pivoted results

almeidabbm:
Closes: 

### Description:

This PR fixes 2 issues:
- Depending on the warehouse, the column names might not match the field name casing, so we need to match the column names to field names to avoid this
- We removed aggregations, so we cannot use the same query to fetch results for the visualizations, this is because it might contain more columns than the fields selected in the chart config and therefore it would brake the viz.

**Before**
Notice the lack of x-axis label
image
image

**After**
image
image
image

### Reviewer actions

### DIFF
--- a/packages/common/src/types/semanticLayer.ts
+++ b/packages/common/src/types/semanticLayer.ts
@@ -240,5 +240,5 @@ export function convertToResultsColumns(
 ) {
     return columns
         .map((value) => convertColumnToResultsColumn(value, resultsColumns))
-        .filter(Boolean) as string[];
+        .filter((value): value is string => !!value);
 }
--- a/packages/common/src/types/semanticLayer.ts
+++ b/packages/common/src/types/semanticLayer.ts
@@ -222,3 +222,23 @@ export function getFilterFieldNamesRecursively(filter: SemanticLayerFilter): {
         ...orFiltersFieldNames,
     ];
 }
+
+// Helper functions to convert between the column names in the query and the column names in the results
+
+export function convertColumnToResultsColumn(
+    column: string,
+    resultsColumns: string[],
+) {
+    return resultsColumns.find(
+        (resultCol) => resultCol.toLowerCase() === column.toLowerCase(),
+    );
+}
+
+export function convertToResultsColumns(
+    columns: string[],
+    resultsColumns: string[],
+) {
+    return columns
+        .map((value) => convertColumnToResultsColumn(value, resultsColumns))
+        .filter(Boolean) as string[];
+}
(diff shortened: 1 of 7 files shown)


packages/common/src/types/semanticLayer.ts
```diff
@@ -222,3 +222,23 @@ export function getFilterFieldNamesRecursively(filter: SemanticLayerFilter): {
         ...orFiltersFieldNames,
     ];
 }
+
+// Helper functions to convert between the column names in the query and the column names in the results
```
magnew, line 226: Why are they different? Just wanna understand

packages/common/src/types/semanticLayer.ts
```diff
@@ -222,3 +222,23 @@ export function getFilterFieldNamesRecursively(filter: SemanticLayerFilter): {
...
+
+export function convertColumnToResultsColumn(
+    column: string,
+    resultsColumns: string[],
+) {
+    return resultsColumns.find(
+        (resultCol) => resultCol.toLowerCase() === column.toLowerCase(),
```
magnew, line 233: So the results column names arent case sensitive?

magnew (approved):
Looks good. Tested with a few warehouses. Nice work!

I had a couple questions, but not really about the code.

packages/common/src/types/semanticLayer.ts
almeidabbm: I it to be warehouse related 🤔

packages/common/src/types/semanticLayer.ts
almeidabbm: Good question, I don't think so but perhaps this is something we should look into a bit better, but from my understanding the semantic layer only needs to receive the fields and they do that mapping to do the query on the warehouse, but then the results given to us are not mapped with column name === field name, so we kinda need to do it again on our side? 🤔

packages/common/src/types/semanticLayer.ts
```diff
@@ -222,3 +222,23 @@ export function getFilterFieldNamesRecursively(filter: SemanticLayerFilter): {
...
+    );
+}
+
+export function convertToResultsColumns(
+    columns: string[],
+    resultsColumns: string[],
+) {
+    return columns
+        .map((value) => convertColumnToResultsColumn(value, resultsColumns))
+        .filter(Boolean) as string[];
```
IrakliJani, line 243:
```suggestion
        .filter((value): check is string => !!value)
```

you can do this to avoid type casting

Issue #11364, opened 2024-09-03
The semantic Viewer chart results table can be different to the query results table. 


TuringLovesDeathMetal:
### Description

- the only place I can add metrics/dimensions to my query is in the `results` tab. 
- If I update the metrics/dimensions in my query, then my chart results table is **not** updated. 
- The way to update the chart results table is to add fields from my `results` tab to the chart configuration. 
- I can have different results tables in the `results` tab and in the `chart` tab. But, the `chart` tab is always a subset of the `results` tab.

PR #11409, opened 2024-09-04
chore: refactor chart viz hook

IrakliJani:
required for: #11210 
required for: #11364 

### Description:

major changes:
- `useChartViz` hook now uses react-query to cache pivoted queries.
- `ChartViz` component does not initialize `useChartViz` - instead it receives data as props

### Reviewer actions


joaoviana (changes requested):
Only requesting changes because of the queryKey variable. 

The rest is all tested and works really well! And it fixed a duplicate `runPivotQuery` request on Dashboard tiles that I wasn't able to fix prior to this 👍

joaoviana (approved):

Issue #11329, opened 2024-09-02
`First` aggregation method doesn't work (replace with `any value`)


joaoviana:
### Description

When selecting a column + its aggregation as `FIRST` then, my chart breaks.

Check what SQL is being generated against the warehouse.

### Steps to Reproduce the Bug or Issue

1. Create chart
2. Select a column that allows `First` aggregation
3. Select it
4. See chart breaking

### Implementation suggestion

Replace it with ANY_VALUE 
https://docs.snowflake.com/en/sql-reference/functions/any_value
https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#any_value
https://www.postgresql.org/docs/16/release-16.html

### Cloud or self-hosting

None

rephus:
any_value on postgres is not a avlid function (only available from v 16+)

https://www.postgresql.org/docs/15/functions-aggregate.html

PR #11420, opened 2024-09-05
fix: first value agg

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/11329

### Description:

Before

Screenshot from 2024-09-05 10-46-31

After

Postgres:

Screenshot from 2024-09-05 11-04-51

Bigquery:
Screenshot from 2024-09-05 10-48-34

### Reviewer actions


joaoviana (approved):
Could you pull latest and re-run `yarn generate-api`?

owlas (changes requested):
This aggregation should be called `any` value otherwise it's not accurate and for databases not supporting "first" it would be confusing.

Instead of the ternary `agg = first ? 'any' : agg` can we have a mapping that maps our internal agg names to the correct functions for each database. It could live on the warehouse clients

packages/common/src/utils/warehouse.ts
```diff
@@ -23,3 +24,38 @@ export const getFieldQuoteChar = (
...
+                const aggregationFunction =
+                    aggregation === VizAggregationOptions.ANY
+                        ? 'ANY_VALUE'
+                        : aggregation;
+                return `${aggregationFunction}(${q}${reference}${q})`;
+
+            case WarehouseTypes.POSTGRES:
+                if (aggregation === VizAggregationOptions.ANY) {
+                    // ANY_VALUE on Postgres is only available from version v16+
+                    return `(ARRAY_AGG(${q}${reference}${q}))[1]`;
```
owlas, line 50: fun workaround!

owlas (approved):

Issue #11425, opened 2024-09-05
I want to be able to change a series label in the SQL runner


TuringLovesDeathMetal:
### Description

I really want to be able to either hide my series, or rename my series. I can’t use special characters in defining SQL field names, so it’s definitely something I’d want to override in the chart!

PR #11436, opened 2024-09-05
fix: set series labels depending on field config `y`

joaoviana:
Closes: #11425

### Description:

- Hide legend if only on y axis fieldConfig
- When from 1 series to 2, then clear to default on series 1
- When only 1 series + when setting a label for `y` axis, then that will change the series label as well 

recording:

https://github.com/user-attachments/assets/f79fc029-130a-44c7-94c7-d91baee8a13f

### Reviewer actions

### DIFF
--- a/packages/common/src/visualizations/CartesianChartDataModel.ts
+++ b/packages/common/src/visualizations/CartesianChartDataModel.ts
@@ -210,7 +210,7 @@ export class CartesianChartDataModel
                 appendToBody: true, // Similar to rendering a tooltip in a Portal
             },
             legend: {
-                show: true,
+                show: !!(this.fieldConfig && this.fieldConfig.y.length > 1),
                 type: 'scroll',
             },
             xAxis: {
--- a/packages/common/src/visualizations/CartesianChartDataModel.ts
+++ b/packages/common/src/visualizations/CartesianChartDataModel.ts
@@ -162,9 +162,18 @@ export class CartesianChartDataModel
                 const seriesFormat = Object.values(display?.series || {}).find(
                     (s) => s.yAxisIndex === index,
                 )?.format;
-                const seriesLabel = Object.values(display?.series || {}).find(
-                    (s) => s.yAxisIndex === index,
-                )?.label;
+
+                const singleYAxisLabel =
+                    // NOTE: When there's only one y-axis left, set the label on the series as well
+                    this.fieldConfig?.y.length === 1 &&
+                    display?.yAxis?.[0]?.label
+                        ? display.yAxis[0].label
+                        : undefined;
+                const seriesLabel =
+                    singleYAxisLabel ??
+                    Object.values(display?.series || {}).find(
+                        (s) => s.yAxisIndex === index,
+                    )?.label;
 
                 const seriesColor = Object.values(display?.series || {}).find(
                     (s) => s.yAxisIndex === index,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -448,6 +440,9 @@ export const ContentPanel: FC = () => {
                                                                     error={
                                                                         chartVizQuery.error
                                                                     }
+                                                                    style={{
+                                                                        height: inputSectionHeight,
```
joaoviana, line 444: this is needed

rephus (changes requested):
Legend should also appear on pivot

image

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -210,7 +210,7 @@ export class CartesianChartDataModel
                 appendToBody: true, // Similar to rendering a tooltip in a Portal
             },
             legend: {
-                show: true,
+                show: !!(this.fieldConfig && this.fieldConfig.y.length > 1),
```
owlas, line 213: you should access the `transformedData.valuesColumns` which has type `PivotChartData` instead of the fieldConfig.

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
```diff
@@ -116,6 +116,16 @@ export const cartesianChartConfigSlice = createSlice({
             config.display = config.display || {};
             config.display.yAxis = config.display.yAxis || [];
 
+            // NOTE: If there's only one y-axis, set the label on the series as well
```
owlas, line 119: Could this be done in the CartesianChartModel so it resolves on render? Or is tied to state changes?

owlas (changes requested):

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
joaoviana: That's a good point, let me try to do it on the `CartesianChartModel`!

packages/common/src/visualizations/CartesianChartDataModel.ts
joaoviana: done!

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
joaoviana: And done! much better

owlas (approved):

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -162,9 +162,18 @@ export class CartesianChartDataModel
                 const seriesFormat = Object.values(display?.series || {}).find(
                     (s) => s.yAxisIndex === index,
                 )?.format;
-                const seriesLabel = Object.values(display?.series || {}).find(
-                    (s) => s.yAxisIndex === index,
-                )?.label;
+
+                const singleYAxisLabel =
+                    // NOTE: When there's only one y-axis left, set the label on the series as well
+                    this.fieldConfig?.y.length === 1 &&
```
owlas, line 168: is there a way to get this working without the field config or not? future refactor will mean it's not available here

rephus (approved):

Issue #11261, opened 2024-08-26
separate semantic viewer and sql runner chart configuration


owlas:
1. The X-axis and Group By in visualization config can only select dimensions.
2. The Y-axis can only select metrics.
3. You cannot add an extra Y-axis if there are not enough metrics to do so.
4. Metrics cannot have aggregation functions, so the UI for that is unnecessary.
5. Group By is only possible if you have more than one dimension.
6. You cannot Group By the same field you are using for the X-axis.

^^^ Double-check what can be shared with the SQL runner and what should be a separate component.

PR #11440, opened 2024-09-05
chore: separate semantic viewer and sql runner chart configuration

magnew:
Closes: #11261 

### Description:

With the change the Semantic Viewer will not share the config forms from the shared data viz. It still uses the shared viz slices and some of the components. The main change here is to move a significant amount of logic from the ResultsRunner to the SemanticViewerResultsRunner. Page specific logic should be in the page specific results runner. 

So for reviewers, this specifically does three things:
- Makes semantic-viewer-specific config forms in the semantic viewer dir
- Moves Semantic Viewer logic the the SemanticViewerRunner
- Implements the UI requirements 

There is some follow up work here to move more logic out of the ResultsRunner super class and into SemanticViewerResultsRunner and SQLRunnerResultsRunner (could the ResutlsRunner become abstract?). 

This also addresses adding some constraints to make the field picking UI better for building charts:
- [x] The X-axis and Group By in visualization config can only select dimensions.
- [x] The Y-axis can only select metrics.
- [x] You cannot add an extra Y-axis if there are not enough metrics to do so.
- [x] Metrics cannot have aggregation functions, so the UI for that is unnecessary.
- [x] Group By is only possible if you have more than one dimension.
- [x] You cannot Group By the same field you are using for the X-axis.

### Reviewer actions


packages/common/src/types/semanticLayer.ts
```diff
@@ -47,6 +48,10 @@ export type SemanticLayerField = {
     availableOperators: SemanticLayerStringFilterOperator[];
 };
 
+export type SemanticLayerColumn = VizSqlColumn & {
```
magnew, line 51: I added this type -- would we rather use `PivotChartData`? The name is confusing, but maybe we want a shared type? On the other hand, we have other semantic viewer specific types here.

packages/frontend/src/features/semanticViewer/components/SemanticViewerVizConfig/CartesianVizStyling.tsx
```diff
@@ -0,0 +1,5 @@
+import { CartesianChartStyling } from '../../../../components/DataViz/config/CartesianChartStyling';
+
+// NOTE: This file is included here, but the styling is shared from the
+// data viz folder for now. If the styling diverges, we can add it here.
+export { CartesianChartStyling as CartesianVizStyling };
```
magnew, line 5: The styling form is shared. It's called out that way in the eraser 'spec' and it doesnt currently have anything specific to semantic viewer, so I kept in in common for now. I put this file here to make it clear that we are generally not sharing the configs, but we could do this a different way if anyone has an opinion.

packages/frontend/src/features/semanticViewer/components/SemanticViewerVizConfig/CartesianVizFieldConfig.tsx
```diff
@@ -0,0 +1,300 @@
...
+    const groupByLayoutOptions = useVizSelector((state) =>
+        cartesianChartSelectors.getPivotLayoutOptions(state, selectedChartType),
+    );
+
+    // For axis layout options, we get the available options from the
+    // viz slice, then we have to filter them with the extra info (kind)
+    // from the columns.
+    const xLayoutOptions = useMemo(
+        () =>
+            indexOptions?.filter((option) => {
```
magnew, line 225: This filtering is a little bit awkward. We get generic fields (no 'kind') from the chart slice and have to filter them by the fields. I'm open to better ideas :D

packages/common/src/types/semanticLayer.ts
owlas: More of a red flag that uses `VizSqlColumn` which sounds like it's sql-runner specific

almeidabbm (approved):
LGTM 🚀

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -86,6 +90,84 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
+            indexLayoutOptions: this.columns.reduce((acc, column) => {
+                const field =
+                    SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
+                        this.fields,
+                        column.reference,
+                    );
+                if (field?.kind === FieldType.DIMENSION) {
+                    acc.push({
+                        reference: column.reference,
+                        type: this.getAxisType(column),
```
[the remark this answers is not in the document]
magnew: Good call. Updated

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -86,6 +90,85 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
+                SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
+                    this.fields,
+                    column.reference,
+                );
+            if (field?.kind === FieldType.DIMENSION) {
+                xColumn = column;
+            } else if (field?.kind === FieldType.METRIC) {
+                yColumn = column;
+            }
+        });
```
almeidabbm, line 152: This loops through all columns and will still try to get the field for each of those - setting x or y to be the last item in the columns list depending on type

was thinking of something very similar to what we had prior to the change we did during the call, e.g.
```suggestion
        const xColumn = this.columns.find((column) => {
            const field =
                SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
                    this.fields,
                    column.reference,
                );

             return field?.kind === FieldType.DIMENSION;
        });
        
        const yColumn = this.columns.find((column) => {
            const field =
                SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
                    this.fields,
                    column.reference,
                );

             return field?.kind === FieldType.METRIC;
        });
```

almeidabbm (approved):

Issue #11264, opened 2024-08-26
I want to be able to promote content from a restricted space (not just public spaces). 


TuringLovesDeathMetal:
### Description

I should be able to promote content from a restricted space. 

### Acceptance criteria

- [ ] the permissions in the upstream project are maintained (even if the permissions in the "promoting" project are different). 
- [ ] users can only promote content to a restricted space in the upstream project if they have at least `editor` access to the space in the upstream project
- [ ] update docs

PR #11453, opened 2024-09-06
feat: promote content from private spaces

ZeRego:
Closes: [#11264](https://github.com/lightdash/lightdash/issues/11264)

### Description:

Create/update content in private spaces.
Note that we don't update space permissions (public/private). 

Sep-06-2024 13-36-31

### Reviewer actions

### DIFF
--- a/docs/docs/guides/how-to-promote-content.md
+++ b/docs/docs/guides/how-to-promote-content.md
@@ -51,7 +51,7 @@ style={{ display: 'block', margin: '0 auto 20px auto' }}
 
 :::info
 
-You must be a `developer` and have access to the chart and space in both the `development project` and the `upstream project`. The space must also be public—promoting content in private or restricted access spaces is not supported yet.
+You must be a `developer` and have access to the chart and space in both the `development project` and the `upstream project`.
 
 :::
 
@@ -88,7 +88,7 @@ style={{ display: 'block', margin: '0 auto 20px auto' }}
 
 :::info
 
-You must be a `developer` and have access to the dashboard and space in both the `development project` and the `upstream project` as well as have access to all the charts in the dashboard. The space must also be public—promoting content in private or restricted access spaces is not supported yet.
+You must be a `developer` and have access to the dashboard and space in both the `development project` and the `upstream project` as well as have access to all the charts in the dashboard.
 
 :::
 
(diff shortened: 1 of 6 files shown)


packages/backend/src/services/PromoteService/PromoteService.ts
```diff
@@ -873,21 +838,33 @@ export class PromoteService extends BaseService {
...
+                const space = await this.spaceModel.createSpace(
                     data.projectUuid,
                     data.name,
                     user.userId,
                     data.isPrivate,
                     data.slug,
                 );
+                if (data.isPrivate) {
+                    // user who created the private space is set as a space admin by default
+                    await this.spaceModel.addSpaceAccess(
```
rephus, line 861: why not copying all the spaceAccess for this space ? , so everyone that already has access in dev also get access in prod ?

packages/backend/src/services/PromoteService/PromoteService.test.ts
```diff
@@ -494,6 +495,56 @@ describe('PromoteService promoting and mutating changes', () => {
...
+        expect(changes.spaces[0].action).toBe('update');
+        expect(changes.spaces[0].data.isPrivate).toBe(false); // Existing space is not private, so it should be kept that way
+
+        await service.upsertSpaces(user, changes);
+
+        expect(spaceModel.update).toHaveBeenCalledTimes(1);
+        expect(spaceModel.update).toHaveBeenCalledWith('upstream-space-uuid', {
+            name: promotedDashboardWithNewPrivateSpace.space.name,
+        });
+        expect(spaceModel.addSpaceAccess).toHaveBeenCalledTimes(0);
```
rephus, line 545: :+1:

docs/docs/guides/how-to-promote-content.md
```diff
@@ -51,7 +51,7 @@ style={{ display: 'block', margin: '0 auto 20px auto' }}
 
 :::info
 
-You must be a `developer` and have access to the chart and space in both the `development project` and the `upstream project`. The space must also be public—promoting content in private or restricted access spaces is not supported yet.
+You must be a `developer` and have access to the chart and space in both the `development project` and the `upstream project`.
```
rephus, line 54: perhaps instead of removing the private warning, we can add a new one saying private access will not be copied into prod

rephus (approved):
Tested, it works as described. 

This is a an improvement of what we currently have, so I'm happy to merge as it is, but I think this is going to cause confusion around who gets accesss, we should let users know this new limitation, or continue implementing promotion for spaces + project.

This can be a problem specially for people who are using prod as a read-only  copy and all the changes happen in dev.

packages/backend/src/services/PromoteService/PromoteService.ts
ZeRego: Created a separate ticket to copy all user access [#11469](https://github.com/lightdash/lightdash/issues/11469)

PR #11460, opened 2024-09-06
chore: upgrade @reduxjs/toolkit

joaoviana:
Closes: N/A

### Description:

Introduces a fix to: 

```
Type error: The inferred type of 'configureStore' cannot be named without a reference to '@reduxjs/toolkit/node_modules/redux'. This is likely not portable. A type annotation is necessary.
```

so we don't have to have a hacky fix for this anymore.

### Reviewer actions


joaoviana:
Not urgent to upgrade, so we can revisit later - the build is generated successfully on local dev, but not on github action

Issue #11463, opened 2024-09-08
AWS SES support for email client


gpdenny:
### Description
Introduce AWS SES Transport ([supported by nodemailer](https://nodemailer.com/transports/ses/)) to email client.

### What problem does this solve?
As a Platform Engineer deploying self-hosted Lightdash into an AWS environment.
I would like to use AWS SES _directly_ to send email.
So that I can avoid the need to generate long-lived SMTP credentials to use SES.

This feature would instead allow me to use an AWS IAM role assigned to my Lightdash container to send mail.

gpdenny:
@ZeRego @owlas 
Just want to check if there's a more appropriate tag to ensure this doesn't drop off the list to be prioritised since #11467
was closed for now?

ptdatta:
@ZeRego @gpdenny is this issue open to work?

ZeRego:
@ptdatta Updated the issue tags. Not open to external contrbutions at the moment.

PR #11467, opened 2024-09-09
feat: Add AWS SES support to email client

gpdenny:
Closes: #11463 

### Description:
Add support for AWS SES transport in email client.

Followed the same configuration / client convention as with S3.
(If there are further integrations with AWS I think it would be more appropriate to create a global configuration for AWS, instead repeating per service, as this is more likely to match how people write policies. Didn't want to potentially make breaking changes here to do that.)

Tested locally with Localstack and in a sandbox environment in AWS.
Also tested SMTP locally to ensure no regressions.

### Reviewer actions

### DIFF
--- a/docs/docs/references/sql-variables.mdx
+++ b/docs/docs/references/sql-variables.mdx
@@ -32,6 +32,5 @@ models:
       sql_filter: ${lightdash.user.email} = 'example@lightdash.com'
 ```
 
-If you are self hosting you can enable [SMTP](/self-host/customize-deployment/configure-smtp-for-lightdash-email-notifications) or [SSO](/self-host/customize-deployment/use-sso-login-for-self-hosted-lightdash) authentication to allow users to verify their email address.
-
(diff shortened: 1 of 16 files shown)


owlas:
Hey @gpdenny thanks for this PR, at a quick glance it looks excellent. I'm going to chat with the team about whether we want to maintain this. We intentionally chose to use SMTP credentials for the broadest compatibility. However I see the benefits of eliminating long lived SMTP credentials.

Let me get back to you next week!

gpdenny:
Hi @owlas just following up on this one :pray: Any thoughts on maintaining?

hamzahc1:
Hi @gpdenny this looks like a good PR! The team is a bit slammed at the moment so we haven't got bandwidth to review it but we will get to it 🙏

owlas:
Hey @gpdenny we're just trying to set the priority for this (I'm closing it in the meantime until we prioritize it). Are you already operating this on a fork of Lightdash? Are you updating to the latest version?

gpdenny:
@owlas 
We're simply not using the email features right now, so far I've not been asked to enable it, but I very much expect that to change in the future :sweat_smile:

Issue #11434, opened 2024-09-05
Should list all databases available to credentials


owlas:
The current sidebar only shows the database listed in the warehouse credentials. But the credentials could have access to multiple databases (e.g. multiple projects in gcp/bigquery)

e.g. here we request all datasets for the default project only, it should enemurate all datasets in all projects: https://github.com/lightdash/lightdash/blob/b23df73d2fd8221c670fbff9be3a70bff8cc3203/packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts#L292-L292

joaoviana:
Could be related: https://github.com/lightdash/lightdash/issues/11383

PR #11474, opened 2024-09-09
fix: check for empty rows when updating warehouse available tables

owlas:
Closes: #11383

Knex throws on empty insert rows: https://github.com/lightdash/lightdash/issues/11434

**Updated wiki**: https://www.notion.so/lightdash/Knex-6a3615a169044daeaa7a54465c631960

User complained about receiving this error, which is a combination of this bug and https://github.com/lightdash/lightdash/issues/11434


rephus (approved):
That makes sense, thanks for fixing

Issue #11270, opened 2024-08-27
UI changes to the SQL editor


PriPatel:
- [ ] Add a bit of spacing above the first line of SQL, just to give it breathing room. 
<img width="1037" alt="Screenshot 2024-08-27 at 13 12 20" src="https://github.com/user-attachments/assets/5afd9264-791e-48d9-b2ff-99c90613ffba">

- [ ] Maybe for later (depending on how quick it is
Its' easier in the old sql runner to know which line you are on at any time. Can the current line be highlighted similarly? It could be that the numbers on the side are highlighted more. 
<img width="140" alt="Screenshot 2024-08-27 at 13 15 16" src="https://github.com/user-attachments/assets/098c5b36-296e-4d37-b935-89949e73e8d7">

PR #11476, opened 2024-09-09
style: highlight on sql editor line and up padding top

joaoviana:
Closes: [#11270](https://github.com/lightdash/lightdash/issues/11270)

### Description:

Increased padding top for sql editor & subtle highlight on line
Screenshot 2024-09-09 at 15 18 43

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
@@ -115,7 +115,8 @@ const LIGHTDASH_THEME = {
     colors: {
         'editor.background': '#FFFFFF',
         'editor.foreground': '#333333',
-        'editor.lineHighlightBackground': '#ffffff',
+        'editor.lineHighlightBackground': '#f8f8f8',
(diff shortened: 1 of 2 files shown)


packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
```diff
@@ -115,7 +115,8 @@ const LIGHTDASH_THEME = {
     colors: {
         'editor.background': '#FFFFFF',
         'editor.foreground': '#333333',
-        'editor.lineHighlightBackground': '#ffffff',
+        'editor.lineHighlightBackground': '#f8f8f8',
```
rephus, line 118: in the previous sql runner, the line number was also highlighted, is this possible here ? if not, this should be enough..

rephus (approved):

packages/frontend/src/features/sqlRunner/components/SqlEditor.tsx
joaoviana: I'm not too sure what keyword on the theming I need to add, but this background should be enough for now 👍

PR #11482, opened 2024-09-09
chore: separate results runners, remove shared runner

magnew:
### Description:

WIP: there is a bit more to be worked out here, but I'm putting this up to see if people agree on the direction. 

Separate the SqlRunnerResultsRunner and SemanticLayerResultsRunner more. This makes is so that that don't share a superclass and instead implement a common interface. This way they can share fewer types and be more independent. 

### Reviewer actions


packages/common/src/visualizations/types/index.ts
```diff
@@ -33,6 +33,12 @@ export enum VizIndexType {
     CATEGORY = 'category',
 }
 
+export function getAxisType(column: VizColumn): VizIndexType {
+    return column.type === DimensionType.STRING
+        ? VizIndexType.CATEGORY
+        : VizIndexType.TIME;
+}
```
almeidabbm, line 40: This should probably be an exhaustive `switch/case` of `column.type`

With this code we're returning `VizIndexType.TIME` when `column.type` is `Number` for example

packages/frontend/src/features/semanticViewer/components/RunSemanticQueryButton.tsx
```diff
@@ -61,14 +61,8 @@ export const RunSemanticQueryButton: FC = () => {
...
-        const vizColumns =
-            SemanticViewerResultsRunner.convertColumnsToVizColumns(
-                fields,
-                resultsColumns,
-            );
-
-        dispatch(setResults({ results: resultsData, columns: vizColumns }));
-    }, [dispatch, resultsData, resultsColumns, fields]);
+        dispatch(setResults({ results: resultsData, columns }));
+    }, [dispatch, resultsData, resultsColumns, fields, columns]);
```
almeidabbm: By using `columns` here directly from the state the results table and table viz don't work correctly:

results table:
image

table viz:
image

this is because `state.columns` is only being set on `setResults` action, and therefore will keep being empty array.

With the previous code, where we converted the columns array that came in the response to `VizColumn` and passed that to the action the columns would always be just the ones in the results.

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -96,91 +116,49 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
+                reference: dimension.name,
+                type: getVizIndexTypeFromDimensionType(dimension.type),
+            })),
+            valuesLayoutOptions: this.metrics.map((metric) => ({
+                reference: metric.name,
+            })),
+            pivotLayoutOptions: this.dimensions.map((dimension) => ({
+                reference: dimension.name,
+                type: getVizIndexTypeFromDimensionType(dimension.type),
+            })),
```
almeidabbm, line 132: This allows us to select fields not present in the results (e.g. here I only selected `name` and `order total`)

image

We should filter them from the available columns

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -96,91 +116,49 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
-        const xColumn = this.columns.find((column) => {
-            const field =
-                SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
-                    this.fields,
-                    column.reference,
-                );
-            return field?.kind === FieldType.DIMENSION;
-        });
+        const xColumn = this.dimensions[0];
+        const yColumn = this.metrics[0];
```
almeidabbm, line 138: Something similar here, were setting the `default` config axis to `dimensions` and/or `metrics` that are not included in the results (e.g. here I selected `type` dimension and `orders` metric but the default config was `name` and `orders total` which are the first in the lists)

image

almeidabbm (commented):
Couple things

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: I believe this is fixed.

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: Now being done in the constructor

packages/common/src/visualizations/types/index.ts
magnew: Good call

packages/frontend/src/features/semanticViewer/components/RunSemanticQueryButton.tsx
magnew: Updated. I think this is fixed.

packages/frontend/src/components/DataViz/hooks/useChartViz.ts
```diff
@@ -7,14 +7,15 @@ import {
...
     type PivotChartData,
+    type VizChartLayout,
 } from '@lightdash/common';
 import { useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 import { useOrganization } from '../../../hooks/organization/useOrganization';
-import { type ResultsRunner } from '../transformers/ResultsRunner';
 
-type Args<T extends ResultsRunner> = {
+type Args<T extends IResultsRunner<VizChartLayout>> = {
```
owlas, line 18: `VizChartLayout` is runner specific so there shouldn't be an `type VizChartLayout` imported here. Are we able to have `T, V extends IResultsRunner<V>` or similar? It should be opaque here

packages/frontend/src/components/DataViz/hooks/useTableDataModel.ts
```diff
@@ -1,7 +1,9 @@
 import {
     ChartKind,
     TableDataModel,
+    type IResultsRunner,
     type RawResultRow,
+    type VizChartLayout,
```
owlas, line 6: shouldn't be needed

owlas: Same for all files in the DataViz directory

Is the goal to fix this next?

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -64,30 +65,52 @@ function getVizIndexTypeFromDimensionType(
     }
 }
 
-export class SemanticViewerResultsRunner extends ResultsRunner {
+export class SemanticViewerResultsRunner
+    implements IResultsRunner<VizChartLayout>
```
owlas, line 69: Here `VizChartLayout` should be `SqlRunnerPivotLayout` and it can be known to this object.

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -64,30 +65,52 @@ function getVizIndexTypeFromDimensionType(
...
+    private readonly dimensions: SemanticLayerField[];
+
+    private readonly metrics: SemanticLayerField[];
+
     constructor({
         query,
         projectUuid,
         fields,
-        ...args
+        rows,
```
owlas, line 89: what's fields vs columns names?

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -96,91 +119,49 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
-            }, [] as VizValuesLayoutOptions[]),
-            pivotLayoutOptions: this.columns.filter((column) => {
-                const field =
-                    SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
-                        this.fields,
-                        column.reference,
-                    );
-                return field?.kind === FieldType.DIMENSION;
-            }),
+            indexLayoutOptions: this.dimensions.map((dimension) => ({
```
owlas, line 122: This is much better, amazing

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -96,91 +119,49 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
     defaultPivotChartLayout(): VizChartLayout | undefined {
-        const xColumn = this.columns.find((column) => {
-            const field =
-                SemanticViewerResultsRunner.findSemanticLayerFieldFromColumn(
-                    this.fields,
-                    column.reference,
-                );
-            return field?.kind === FieldType.DIMENSION;
-        });
+        const xColumn = this.dimensions[0];
```
owlas, line 137: Make this pick the first date dimension if it exists otherwise the first dimension

packages/common/src/visualizations/types/index.ts
owlas: Also is this the same as `getVizIndexTypeFromDimensionType` ?

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -96,91 +119,49 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
...
-                    reference: yColumn.reference,
+                    reference: yColumn.name,
                 },
             ],
             groupBy: [],
         };
     }
 
-    static convertColumnsToVizColumns(
+    static convertColumnNamesToVizColumns(
```
owlas, line 158: This is because the backend doesnt know types?

We should fix this on the backend

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -231,17 +203,14 @@ export class SemanticViewerResultsRunner extends ResultsRunner {
 
         // The backend call has no knowledge of field types, so we need to map them to the correct types
         const vizColumns: VizColumn[] =
-            SemanticViewerResultsRunner.convertColumnsToVizColumns(
+            SemanticViewerResultsRunner.convertColumnNamesToVizColumns(
```
owlas, line 206: make this the backend's problem

packages/common/src/visualizations/types/index.ts
```diff
@@ -33,6 +33,19 @@ export enum VizIndexType {
     CATEGORY = 'category',
 }
 
+export function getColumnAxisType(column: VizColumn): VizIndexType {
```
owlas, line 36: is this the same as `getVizIndexTypeFromDimensionType`?

packages/frontend/src/features/sqlRunner/runners/SqlRunnerResultsRunner.ts
```diff
@@ -114,4 +132,195 @@ export class SqlRunnerResultsRunner extends ResultsRunner {
             columns,
         };
     }
+
```
owlas, line 135: moved back from ResultsRunner right? This is essentially a revert?

owlas (commented):
Looking better.

Primary feedback is that `VizChartLayout` was supposed to be a generic on the Runner so it can differ between each runner. This is still the same type on both runners.

To fix:
- Anything chart specific (runner agnostic) should have a runner with a generic/opaque layout type.
- Anything runner specific can implement a layout

packages/frontend/src/components/DataViz/hooks/useTableDataModel.ts
magnew: > Is the goal to fix this next?

Yes

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: I'll make it clearer with some renames. Fields is all of the semantic layer fields. columnNames is just the selected ones.

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: Agree. I'd like to do it in a follow up. I will add a comment.

packages/common/src/visualizations/types/index.ts
magnew: Annoyingly similar, but this is for the generic type instead of the Semantic runner one. I'd like to merge these.

packages/common/src/visualizations/types/IResultsRunner.ts
```diff
@@ -20,8 +22,10 @@ export interface IResultsRunner<TPivotChartLayout> {
 
     pivotChartOptions(): VizCartesianChartOptions;
 
+    // Sql specific?
```
magnew, line 25: Get dimensions

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -82,23 +83,81 @@ export class CartesianChartDataModel {
         const dimensions = this.resultsRunner.getDimensions();
         const metrics = this.resultsRunner.getMetrics();
 
-        const xColumn = dimensions[0];
-        const yColumn = metrics[0];
+        // TODO: the types could be cleaned up here to have fewer 'in' checks. 
+        // This works for
```
magnew, line 87: These types can be cleaned up, but this makes it work pretty well on both pages.

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: Resolving but leaving the other question about this open

magnew:
Closed, we've combined code in a new branch here: #11611

Issue #11368, opened 2024-09-03
I'm able to add semantic viewer charts to a dashboard


TuringLovesDeathMetal:
### Description

- I'm able to add a semantic viewer chart to a dashboard. 
- I can do this by clicking `new tile` in a dashboard and `selecting an existing chart`. But I **cannot** create a chart from within a dashboard. 
- I can do this by clicking `add to dashboard` in `view` mode of my saved chart. 
- **Note:** dashboards can only be created with charts from a single semantic layer + the SQL runner (that means they're all using the new components, not the old explore view!)
- We will need to turn off features on the dashboard based on the semantic viewer that you're using (e.g. date zoom, etc.). We'll need to figure this out and this information will be important when we build the semantic layer connections later on!

Issue #11366, opened 2024-09-03
I'm able to save charts from the Semantic Viewer


TuringLovesDeathMetal:
### Description

- Once I've created a chart in the Semantic Viewer, I'm able to save the chart. 
- the chart must be saved to a space (no charts within dashboards yet)

PR #11485, opened 2024-09-10
feat: save semantic viewer charts and add in the dashboards

IrakliJani:
Closes: #11366 
Closes: https://github.com/lightdash/lightdash/issues/11368

### Reviewer actions


packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -740,6 +741,32 @@ export type CreateSqlChartVersionEvent = BaseTrack & {
...
+    event: 'semantic_layer_chart_version.created';
+    userId: string;
+    properties: {
+        chartId: string;
+        versionId: string;
+        projectId: string;
+        organizationId: string;
+        chartKind: ChartKind;
+        semanticLayerQuery: SemanticLayerQuery;
+        barChart?: {
```
IrakliJani, line 754: this does not have table config. do we also need to keep all configuration blobs as the analytics event or the active one @TuringLovesDeathMetal

packages/backend/src/services/SavedSemanticLayerService/SavedSemanticLayerService.ts
```diff
@@ -0,0 +1,259 @@
...
+        semanticLayerChart: SemanticLayerCreateChart,
+    ): Promise<ApiSemanticLayerCreateChart['results']> {
+        const { organizationUuid } = await this.projectModel.getSummary(
+            projectUuid,
+        );
+        if (
+            user.ability.cannot(
+                'manage',
+                // TODO: add it's own ability
+                subject('CustomSql', { organizationUuid, projectUuid }),
```
IrakliJani, line 206: add `ability`

packages/frontend/src/components/DashboardTiles/DashboardSemanticLayerChartTile.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    const resultsRunner = useMemo(() => {
+        if (!data || !fields) return;
+
+        return new SemanticViewerResultsRunner({
+            projectUuid,
+            fields,
+            query: data.chart.semanticLayerQuery,
+            rows: chartData.results,
+            columns: [], // TODO: sqlRunnerChartData.columns,
```
IrakliJani, line 85: forgot column conversion here

packages/frontend/src/components/common/Page/Sidebar.tsx
```diff
@@ -107,20 +101,21 @@ const Sidebar: FC<React.PropsWithChildren<Props>> = ({
                 >
                     {(style) => (
                         <>
-                            <Card
```
IrakliJani, line 110: card is too opinionated to be a sidebar component

packages/backend/src/services/SavedSemanticLayerService/SavedSemanticLayerService.ts
IrakliJani: can be done as a separate ticket

packages/backend/src/analytics/LightdashAnalytics.ts
IrakliJani: @TuringLovesDeathMetalsSorry for providing less than a zero context 😂 

I noticed that with SQL Runner, when you save a chart, we fire an analytics event. The analytics event includes all of the configurations for all charts, not just the active chart that's being used. Additionally, we are not including the table visualization configuration at all. I copied this "blindly" from SQL Runner but left a comment not to forget to update both places if needed.

TuringLovesDeathMetal: >The analytics event includes all of the configurations for all charts, not just the active chart that's being used.


wut - that sounds wild 😂 

Ya, can we _**just**_ include the active chart configuration. Also, if there's a table visualization, then include it's configuration

almeidabbm (approved):

Issue #11142, opened 2024-08-14
Loading and error states for sql runner catalog


magnew:
### Description

Loading and error states for sql runner catalog:

- Better loading message in the sidebar, since loading the catalog could take a long time. e.g. spinner then after 1 second "please wait this could take a few minutes"
- No error message if sidebar /tables has an error, it's just blank. This happened to me when I enabled user warehouse credentials but hadn't provided any

PR #11491, opened 2024-09-10
Fix/loading error states for sql runner catalog

andrewdw:
Closes: #11142 

### Description:

- consolidated loading state into the TablesPanel.tsx
- added error message if useRefreshTables throws an error
- added a "Hang on, still loading" message if the wait is over 3s for panel to load

https://github.com/user-attachments/assets/a1c8a115-68fc-4209-bd1a-64e2ffc5e73e

https://github.com/user-attachments/assets/f0ec6d8e-00f0-4c95-b244-38db8934b008

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
@@ -2,7 +2,6 @@ import { ChartKind } from '@lightdash/common';
 import {
     ActionIcon,
     Group,
-    LoadingOverlay,
     ScrollArea,
     Stack,
     Title,
@@ -23,18 +22,18 @@ type Props = {
 
 export const Sidebar: FC<Props> = ({ setSidebarOpen }) => {
     const dispatch = useAppDispatch();
-
     const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
 
     const {
         mutate: updateTables,
-        // TODO: unify this loading with the one in TablesPanel, handle error
         isLoading,
+        error,
     } = useRefreshTables({ projectUuid });
 
     const { selectedChartType, activeSidebarTab, sqlColumns } = useAppSelector(
         (state) => state.sqlRunner,
     );
+
     return (
         <Stack spacing="xs" sx={{ flex: 1, overflow: 'hidden' }}>
             <Group position="apart">
@@ -75,8 +74,10 @@ export const Sidebar: FC<Props> = ({ setSidebarOpen }) => {
                 }
                 sx={{ flex: 1, overflow: 'hidden' }}
             >
-                <LoadingOverlay visible={isLoading} />
-                <TablesPanel />
+                <TablesPanel
+                    isLoading={isLoading}
+                    error={error?.message || null}
+                />
             </Stack>
 
             <ScrollArea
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/hooks/useTables.tsx
```diff
@@ -104,14 +104,16 @@ export const useRefreshTables = ({
...
 
-    return useMutation({
-        mutationFn: () => refreshTables({ projectUuid }),
-        onSuccess: async () => {
-            await queryClient.invalidateQueries([
-                'sqlRunner',
-                'tables',
-                projectUuid,
-            ]);
+    return useMutation<ApiWarehouseCatalog, { message: string }, void>(
```
joaoviana, line 107: It's most likely the error will be of type `ApiError`
```suggestion
     return useMutation<ApiWarehouseCatalog, ApiError>(
```

packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
```diff
@@ -75,8 +74,10 @@ export const Sidebar: FC<Props> = ({ setSidebarOpen }) => {
                 }
                 sx={{ flex: 1, overflow: 'hidden' }}
             >
-                <LoadingOverlay visible={isLoading} />
-                <TablesPanel />
+                <TablesPanel
+                    isLoading={isLoading}
+                    error={error?.message || null}
```
joaoviana, line 79: then here we can rewrite the message accessing:
```suggestion
                 error={error?.error.message || null}
```

packages/frontend/src/features/sqlRunner/components/TablesPanel.tsx
```diff
@@ -1,56 +1,109 @@
...
+    isLoading,
+    error,
+}) => {
     const initialPanelSizes = [80, 20];
     const activeTable = useAppSelector((state) => state.sqlRunner.activeTable);
+
+    // state for controlling the "still loading" message
+    const [showLoadingMessage, setShowLoadingMessage] = useState(false);
+
+    // effect to handle loading timeout
```
joaoviana, line 23: This is great and works really well, but we usually try to use `mantine` hooks that we have already available to us so we don't have to worry about the timeout clearing on unmount as well 👍 

Here's an implementation suggestion:

```tsx
 // state for controlling the "still loading" message
    const [showLoadingMessage, setShowLoadingMessage] = useState(false);

    const { start, clear } = useTimeout(
        () => setShowLoadingMessage(true),
        3000,
    );

    useEffect(() => {
        if (isLoading) {
            setShowLoadingMessage(false);
            start();
        } else {
            clear();
            setShowLoadingMessage(false);
        }
    }, [isLoading, start, clear]);
```

packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
joaoviana: Something to consider in the future: we could just have a generic error message to the user if there is an error, like `Couldn't get tables (or wtv)`

joaoviana (approved):
Looking good! 
Just needs a rebase + some Qs from my end 👍

joaoviana (changes requested):
(approved by accident)

packages/frontend/src/features/sqlRunner/components/TablesPanel.tsx
andrewdw: Nice. I need to get familiar with meantine hooks and utility funcs

PR #11501, opened 2024-09-10
chore: refetch data on sort or filter in semantic viewer

magnew:
Closes: #11490 

### Description:

- Refetch chart data on sort and query change. This updates query keys based on fitler and sort updates. Note that the hook itself doesn't get the sorts or filters directly, but now exposes a cache key that can be updated. This is because the useChartViz can be used in multiple places and shouldnt rely on page-specific types

This handles charts and and I'm posting it so we could use it for the demo. More to do:
- Tables don't go through this useQuery so don't get updated with this cache key. To update them, the query needs to be run again
- Filters get updated too often, it should only be on apply. I'm working on this. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DataViz/hooks/useChartViz.ts
+++ b/packages/frontend/src/components/DataViz/hooks/useChartViz.ts
@@ -71,7 +71,7 @@ export const useChartViz = <T extends ResultsRunner>({
             projectUuid,
             limit,
             JSON.stringify(config.fieldConfig),
-            additionalQueryKey,
+            ...(additionalQueryKey ?? []),
         ];
     }, [projectUuid, limit, config, additionalQueryKey]);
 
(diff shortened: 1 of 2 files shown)


almeidabbm (approved):

packages/frontend/src/components/DataViz/hooks/useChartViz.ts
```diff
@@ -61,8 +67,13 @@ export const useChartViz = <T extends ResultsRunner>({
         if (!config) return undefined;
         if (isVizTableConfig(config)) return undefined;
 
-        return [projectUuid, limit, JSON.stringify(config.fieldConfig)];
-    }, [projectUuid, limit, config]);
+        return [
+            projectUuid,
+            limit,
+            JSON.stringify(config.fieldConfig),
+            additionalQueryKey,
```
almeidabbm, line 74: I would probably destructure this here since additionalQueryKey should be array of items for the query:

```suggestion
            ...(additionalQueryKey ?? []),
```

Issue #4638, opened 2023-03-01
I want to be able to `export results` in the SQL runner.


TuringLovesDeathMetal:
### Description

Right now, you can't export results from the SQL runner. 

We should have the same export functionality in the results table for the SQL runner as we do for the Explore page. 

TO start, we can just allow exporting the results table to a .csv. 

### Acceptance criteria

- [ ] I can export the results from the SQL runner chart query to a .csv
- [ ] The option to export the results _**only**_ appears on the **chart** results table page. 
- [ ] Users can export results for the chart in `view` mode. 
- [ ] Users can export results for the chart in `edit` mode 
- [ ] There is the option to adjust the limit in the export: `results in table`, `all results`, `custom`
- [ ] there is the same export limit for number of cells as we have today (I think it's 1 million? - whatever the env variable is for the explore view, let's use the same for the SQL runner)

<img width="367" alt="image" src="https://github.com/user-attachments/assets/ff87f4dc-a492-404d-bb23-b1f2b385439c">

PR #11510, opened 2024-09-10
feat: download csv results sql chart

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/4638

### Description:

Screenshot from 2024-09-10 17-42-02

[Screencast from 10-09-24 16:36:45.webm](https://github.com/user-attachments/assets/6eb60e40-451b-410c-8941-e03a616f7fa8)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/pages/ViewSqlChart.tsx
+++ b/packages/frontend/src/pages/ViewSqlChart.tsx
@@ -200,14 +200,25 @@ const ViewSqlChart = () => {
                                                 columns.map(
                                                     (col) => col.reference,
                                                 );
+                                            console.log('results', results);
                                             const csvContent = [
                                                 columnReferences?.join(','),
                                                 ...results.map((row) =>
-                                                    Object.values(row).join(
-                                                        ',',
-                                                    ),
+                                                    columnReferences
+                                                        .map((reference) => {
+                                                            return (
+                                                                row[
+                                                                    reference
+                                                                ] || '-'
+                                                            );
+                                                        })
+                                                        .join(','),
                                                 ),
                                             ].join('\n');
+                                            console.log(
+                                                'csvContent',
+                                                csvContent,
+                                            );
 
                                             const blob = new Blob(
                                                 [csvContent],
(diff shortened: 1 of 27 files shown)


rephus:
@joaoviana afaik we don't currently store results anywhere in the slice, so I decided to keep things small to just store the dataUrl, and then download the data again and exporting the csv from the frontend. 

I have a couple of questions: 

- IS this the right results to download for the CSV ? or perhaps we only care about the pivot sql 
- Why do we have `sqlRunner/saved/slug/${slug}/results-job`  and `/sqlRunner/run` ? What's the difference ?

packages/frontend/src/pages/ViewSqlChart.tsx
```diff
@@ -110,6 +117,11 @@ const ViewSqlChart = () => {
         limit: sqlChart?.limit,
     });
 
+    useEffect(() => {
+        if (!chartVizQuery?.data?.url) return;
+        dispatch(setDataUrl(chartVizQuery.data.url));
```
rephus, line 122: I don't really use the dataUrl anymore, because I have direct acces sto chartVizQuery.data in this component , but I think it might be useful for downloading from other places. But let me know if you want me to remove it.

joaoviana: that's great! As long as it's managed in state somehow that's good 💪

packages/frontend/src/pages/ViewSqlChart.tsx
```diff
@@ -129,36 +141,75 @@ const ViewSqlChart = () => {
...
+                                value={activeTab}
+                                onChange={(val: TabOption) => setActiveTab(val)}
+                            />
+                        </Group>
+                        <Button
+                            leftIcon={<MantineIcon icon={IconDownload} />}
+                            variant="default"
+                            size="xs"
+                            disabled={!chartVizQuery?.data?.url}
+                            onClick={() => {
```
joaoviana, line 181: tried out adding the columns 

```ts
 onClick={() => {
                                if (chartVizQuery?.data?.url) {
                                    void getResultsFromStream<RawResultRow>(
                                        chartVizQuery.data.url,
                                    ).then((results) => {
                                        // Add column headers to the CSV
                                        const columns =
                                            chartVizQuery.data?.columns.map(
                                                (col) => col.reference,
                                            );
                                        const csvContent = [
                                            columns?.join(','),
                                            ...results.map((row) =>
                                                Object.values(row).join(','),
                                            ),
                                        ].join('\n');

                                     ...REST OF CODE
                       
```

joaoviana (approved):

PR #11564, opened 2024-09-16
feat: allow for two levels of filter groups

ZeRego:
Closes: 

### Description:

For more advanced filter logic, users need multiple levels of groups.
Example:
```
in the last 30 days 
AND (
  (
   day IN (1, 30)
   AND value > X
  )
  OR day NOT IN (1, 30)
)
```

We only allowed for 1 level which was a blocker for such logic. 
We are now allowing 2 levels as it solves most of the use cases and doesn't break the UI. 

Screenshot 2024-09-16 at 15 51 08

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterGroupForm.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterGroupForm.tsx
@@ -36,7 +36,7 @@ import FilterRuleForm from './FilterRuleForm';
 type Props = {
     hideButtons?: boolean;
     hideLine?: boolean;
-    allowConvertToGroup?: boolean;
+    groupDepth?: number;
     fields: FilterableField[];
     filterGroup: FilterGroup;
     isEditMode: boolean;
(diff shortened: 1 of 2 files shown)


rephus:
There is a build error

> 32 24.50 src/components/common/Filters/index.tsx(202,33): error TS2322: Type '{ hideLine: true; hideButtons: true; filterGroup: FilterGroup; fields: FieldWithSuggestions[]; isEditMode: boolean; onChange: (filterGroup: FilterGroup) => void; onDelete: () => void; allowConvertToGroup: true; }' is not assignable to type 'IntrinsicAttributes & Props'.
#32 24.50   Property 'allowConvertToGroup' does not exist on type 'IntrinsicAttributes & Props'.

rephus (approved):
Tested, it works as expected 

image

This is out of scope, but I'm missing some way of organizing the filters, like a drag and drop where I can place some filters in different groups.

Issue #11389, opened 2024-09-04
Automatically add a partitioning WHERE clause to default sql for bigquery in the sql runner


owlas:
The default query can query a lot of data in bigquery (sometimes hitting the billing limit and failing). This is because we default select everything.

If the user has a partition column in bigquery we can auto fill:

```sql
SELECT * 
  FROM `project_id.dataset_name.INFORMATION_SCHEMA.COLUMNS`
  WHERE TABLE_NAME = Table_Name 
    AND is_partitioning_column = "YES"
```

Then we should generate a query:

```sql
SELECT * FROM <table>
WHERE <partition_column> = -- <enter value>
```

PR #11576, opened 2024-09-17
feat: add partitioning bigquery sql runner

rephus:
Closes: #11389

### Description:

[Screencast from 17-09-24 11:24:37.webm](https://github.com/user-attachments/assets/dcb6e824-8690-4ed1-ad4a-9b3ff757c74c)

### Reviewer actions

### DIFF
--- a/packages/common/src/types/warehouse.ts
+++ b/packages/common/src/types/warehouse.ts
@@ -123,14 +123,8 @@ export enum PartitionType {
     DATE = 'DATE',
     RANGE = 'RANGE',
 }
-export type PartitionColumnDate = {
-    partitionType: PartitionType.DATE;
-    type: string;
+
+export type PartitionColumn = {
+    partitionType: PartitionType;
     field: string;
 };
-export type PartitionColumnRange = {
-    partitionType: PartitionType.RANGE;
-    field: string;
-    range: { start: string; end: string; interval: string };
-};
-export type PartitionColumn = PartitionColumnDate | PartitionColumnRange;
--- a/packages/common/src/types/warehouse.ts
+++ b/packages/common/src/types/warehouse.ts
@@ -15,10 +15,19 @@ export type WarehouseCatalog = {
     };
 };
 
+export type WarehouseTablesCatalog = {
+    [database: string]: {
+        [schema: string]: {
+            [table: string]: { partitionColumn?: PartitionColumn };
+        };
+    };
+};
+
 export type WarehouseTables = {
     database: string;
     schema: string;
     table: string;
+    partitionColumn?: PartitionColumn;
 }[];
 
 export type WarehouseResults = {
@@ -100,7 +109,28 @@ export type ApiWarehouseCatalog = {
     results: WarehouseCatalog;
 };
 
+export type ApiWarehouseTablesCatalog = {
+    status: 'ok';
+    results: WarehouseTablesCatalog;
+};
+
 export type ApiWarehouseTableFields = {
     status: 'ok';
     results: WarehouseTableSchema;
 };
+
+export enum PartitionType {
+    DATE = 'DATE',
+    RANGE = 'RANGE',
+}
+export type PartitionColumnDate = {
+    partitionType: PartitionType.DATE;
+    type: string;
+    field: string;
+};
+export type PartitionColumnRange = {
+    partitionType: PartitionType.RANGE;
+    field: string;
+    range: { start: string; end: string; interval: string };
+};
+export type PartitionColumn = PartitionColumnDate | PartitionColumnRange;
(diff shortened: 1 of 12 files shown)


packages/common/src/types/warehouse.ts
```diff
@@ -15,10 +15,19 @@ export type WarehouseCatalog = {
     };
 };
 
+export type WarehouseTablesCatalog = {
```
rephus, line 18: The same WarehouseCatalog was used for both fields and tables. 

Tables was returning an empty list of fields, but this was uncompatible with the new `partitionColumn` property , so I created a new type which is used only metadata on tables

packages/common/src/types/warehouse.ts
```diff
@@ -100,7 +109,28 @@ export type ApiWarehouseCatalog = {
...
 export type ApiWarehouseTableFields = {
     status: 'ok';
     results: WarehouseTableSchema;
 };
+
+export enum PartitionType {
+    DATE = 'DATE',
+    RANGE = 'RANGE',
+}
+export type PartitionColumnDate = {
```
rephus, line 126: These types are an extension of what bigquery returns on their metadata (+ partitionType for easy type checking)

packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -294,13 +296,30 @@ export class BigqueryWarehouseClient extends WarehouseBaseClient<CreateBigqueryC
...
-                database: t.bigQuery.projectId,
-                schema: t.dataset.id!,
-                table: t.id!,
-            })),
+        const tablesWithPartitions = await Promise.all(
+            datasetTablesResponses.flatMap(([tables]) =>
+                tables.map(async (t) => {
+                    const [metadata] = await t.getMetadata();
+                    const partitionColumn: PartitionColumn =
+                        metadata.timePartitioning || metadata.rangePartitioning
```
rephus, line 304: this is where we extract the metadata info from bigquery

owlas: This is parallel API calls one per table right? IF the user has 300 tables, it's 300 API calls. Did you stress test this yet?

rephus: Good point, I didn't stress test this, We have 14 tables on bigquery, and I didn't notice this was taking longer. 

Instead of making 1 query per table, I can make 1 query per schema, running this custom code. I don't see this information on the schema metadata

image

rephus: This query doesn't return the `range` or date granularity, which is useful, but perhaps less important than performance, what do you think ?

packages/frontend/src/features/sqlRunner/components/Tables.tsx
```diff
@@ -35,10 +36,31 @@ interface TableItemProps extends BoxProps {
...
 }
 
+const partitionFilter = (partitionColumn: PartitionColumn | undefined) => {
+    if (partitionColumn) {
+        const hint =
+            partitionColumn.partitionType === PartitionType.DATE
+                ? `This table has a date partition on this field of ${partitionColumn.type.toLowerCase()} granularity`
+                : `This table has a range partition on this field from ${partitionColumn.range.start} to ${partitionColumn.range.end} with an interval of ${partitionColumn.range.interval}`;
+
+        return `\nWHERE ${partitionColumn.field} = <add value> -- ${hint}`;
```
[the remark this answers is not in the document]
rephus: Done, because I'm doing now a query per dataset, so I don't know the range, I can't use it on the range default, I used 0 instead.

packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -294,12 +296,59 @@ export class BigqueryWarehouseClient extends WarehouseBaseClient<CreateBigqueryC
...
+                const tablePartitionInfo =
+                    datasetPartitionInfo?.partitionColumns.find(
+                        (pc) => pc.table_name === t.id,
+                    );
+                const partitionColumn: PartitionColumn | undefined =
+                    tablePartitionInfo
+                        ? {
+                              field: tablePartitionInfo.column_name,
+                              partitionType:
+                                  tablePartitionInfo.data_type === 'INT64'
```
joaoviana, line 339:
```suggestion
                                  tablePartitionInfo.data_type === BigqueryFieldType.INT64
```

joaoviana (approved):
Works great! Tested in POSTGRES (without partition columns) and bigquery with partition columns

joaoviana:
double-checked in db: 
image

Issue #11557, opened 2024-09-16
I would like to create an explore from the SQL runner [PoC]


joaoviana:
### Description

Allowing users to save their custom SQL queries as reusable models (explores) in the SQL Runner. This will enable users to access these queries within the Explorer UI, facilitating easier data exploration and visualization.

**Acceptance criteria**
- [ ] Add a button labeled "Create Reusable Model" to the SQL Runner UI 
- [ ] This button is only enabled if a query has been run and it's successful
- [ ] Clicking the button opens a modal prompting the user to enter: Explore Name (required)
- [ ] After saving, the user sees a green toast "Explore created <button>`Query from explore` </button>" - if the user clicks on the button, a new tab will open with the url `/tables/EXPLORE_NAME`
- [ ] The new custom explore will then be available for use in `query from tables` url
- [ ] The new custom explore is labeled as `custom` to indicate that it's not generated from a lightdash yml file
- [ ] When I save a custom explore, it gets translated as an explore to be saved in the db
- [ ] `sql` from the sql query would become the `sqlTable` field on the `Table` in the `Explore`
- [ ] When I land on the custom explore, I see `dimensions` which are the same as the columns in the SQL Query results table, but I see empty `metrics` 

> [!NOTE]
> Anybody can see the explore and query from it
> Users can only create metrics through custom metrics

> [!NOTE]
> Figure out how to edit a custom model - could it be on the explorer? If so, we should introduce a `edit` button that opens a modal, lets the user type a query, then lets them review the dimensions they would have available to them. What would this mean for saved charts?
> Figure out how to create persisted metrics when you're writing/saving the custom model - the user still needs to select an aggregation

### What problem does this solve?

Creates a bridge between the sql runner and the current explore

PR #11578, opened 2024-09-17
feat: enable saving sql runner model

joaoviana:
Closes: #11557

### Description:

- Hides this feature behind a feature flag
- Allows users to save sql runner queries as explores to be used when querying from tables
- Can go to explore and query dimensions, create custom metrics, and also save charts!

Limitations
- Can't edit 
- When clicking `refresh dbt`, it clears the saved explores
- Missing some core features

> [!IMPORTANT]
> This doesn't cover **everything** on the issue, but this is for demo purposes. The scope has been verified by @owlas.

Demo: 

https://github.com/user-attachments/assets/4d245988-7b78-4191-b0e8-518b18260c42

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -9,6 +9,7 @@ import {
     DbtProjectConfig,
     Explore,
     ExploreError,
+    friendlyName,
     generateSlug,
     isExploreError,
     NotExistsError,
@@ -1833,10 +1834,10 @@ export class ProjectModel {
         const toAddToCached: Explore = {
             name,
             tags: [],
-            label: name,
+            label: friendlyName(name),
             tables: translatedToExplore.tables,
             baseTable: name,
-            groupLabel: name,
+            groupLabel: 'Custom explores',
             joinedTables: [],
             targetDatabase: SupportedDbtAdapter.POSTGRES,
         };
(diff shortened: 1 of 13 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1802,4 +1806,60 @@ export class ProjectModel {
...
+            .onConflict(['project_uuid', 'name'])
+            .merge()
+            .returning(['name', 'cached_explore_uuid']);
+        const toAddToCached: Explore = {
+            name,
+            tags: [],
+            label: name,
+            tables: translatedToExplore.tables,
+            baseTable: name,
+            groupLabel: name,
```
rephus, line 1839: no need to add a gropuLabel, or perhaps we could use `sql runner` instead, so it doesn't look like this on the explore: 

image

rephus (approved):
This is amazing, it works perfectly on postgres, bigquery and snowflake, because of the way it is implemented, most of the features are available, including saving into a dashboard (as long as you don't refresh dbt)

packages/backend/src/models/ProjectModel/ProjectModel.ts
joaoviana: fixed the `groupLabel` and made some tiny UI/UX improvements with @PriPatel 👍

Issue #11498, opened 2024-09-10
sorting semantic viewer improvements


TuringLovesDeathMetal:
### Description

- [x] adding a sort should automatically trigger a query re-run. 
- [x] sorts should appear on the results table, not in pills above the results table. 
- [x] Clicking on the column should apply the sort to the table, based on that field. (same behaviour as we have right now with the pills, just applied to the column headers - including the arrows indicating the sort direction).
<img width="933" alt="image" src="https://github.com/user-attachments/assets/9ff1b116-689e-425a-9028-26d92417bace">

- [x] a chart table can't be sorted by fields that aren't included in the table.
  - [x]  If I have sorted by a field in my results table which is not included in my chart, then we should just not include that sort in the chart results table and in the query that generates the chart results table. 
  - [x] if there are sorts based on fields that _**are**_ in my chart results table, then these sorts should persist.

PR #11582, opened 2024-09-17
feat: sorts on table headers

almeidabbm:
Closes: #11498

### Description:

https://github.com/user-attachments/assets/a20fd968-9370-4202-aeb2-c6a8ec187074

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/semanticViewer/components/Content.tsx
+++ b/packages/frontend/src/features/semanticViewer/components/Content.tsx
@@ -12,7 +12,7 @@ import { SemanticViewerResultsRunner } from '../runners/SemanticViewerResultsRun
 import { useAppDispatch, useAppSelector } from '../store/hooks';
 import {
     selectAllSelectedFieldNames,
-    selectAllSelectedFieldsByKind,
+    selectAllSelectedFields,
     selectSemanticLayerInfo,
     selectSemanticLayerQuery,
 } from '../store/selectors';
@@ -20,6 +20,7 @@ import {
     EditorTabs,
     setActiveEditorTab,
     setResults,
+    updateSortBy,
 } from '../store/semanticViewerSlice';
 import ContentCharts from './ContentCharts';
 import ContentResults from './ContentResults';
@@ -42,14 +43,7 @@ const Content: FC = () => {
         selectChartConfigByKind(state, activeChartKind),
     );
 
-    const allSelectedFieldsByKind = useAppSelector(
-        selectAllSelectedFieldsByKind,
-    );
-
-    const selectedFieldsCount =
-        allSelectedFieldsByKind.dimensions.length +
-        allSelectedFieldsByKind.metrics.length +
-        allSelectedFieldsByKind.timeDimensions.length;
+    const allSelectedFields = useAppSelector(selectAllSelectedFields);
 
     const {
         data: requestData,
@@ -122,6 +116,19 @@ const Content: FC = () => {
         setHasClickedRunQueryButton(true);
     }, [runSemanticViewerQuery]);
 
+    const handleSortField = useCallback(
+        (fieldName: string) => {
+            const fieldToUpdate = allSelectedFields.find(
+                (f) => f.name === fieldName,
+            );
+
+            if (fieldToUpdate) {
+                dispatch(updateSortBy(fieldToUpdate));
+            }
+        },
+        [allSelectedFields, dispatch],
+    );
+
     return (
         <>
             <Group
@@ -188,7 +195,7 @@ const Content: FC = () => {
                         description="Please select a view from the sidebar to start building a query"
                     />
                 </Center>
-            ) : selectedFieldsCount === 0 ? (
+            ) : allSelectedFields.length === 0 ? (
                 <Center sx={{ flexGrow: 1 }}>
                     <SuboptimalState
                         title="Select a field"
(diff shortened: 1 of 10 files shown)


TuringLovesDeathMetal:
This looks great! The only thing that was weird was when I removed all the sorts, the query reran. 

I would expect that if I remove all of the sorting, the table stays in the same sort as whatever it was before (i.e. we shouldn't need to rerun the query).

packages/frontend/src/features/semanticViewer/store/semanticViewerSlice.ts
```diff
@@ -329,11 +338,15 @@ export const semanticViewerSlice = createSlice({
...
             }
-        },
 
+            state.shouldFetchResults = true;
+        },
         updateSaveModalOpen: (state, action: PayloadAction<boolean>) => {
             state.saveModalOpen = action.payload;
         },
+        setShouldFetchResults: (state, action: PayloadAction<boolean>) => {
+            state.shouldFetchResults = action.payload;
```
IrakliJani, line 348: hmm. im not super excited with this change. we had this in a previous explore and I can remember at least 3 times when we had to fix results not re-fetching after certain actions. I open to discuss how we can implement a "smart" auto re-fetch solution.

one of them being:
- any change that happens inside semantic query should run the refetch results:
- any change that happens inside semantic query and pivot config should refetch pivot results.

packages/frontend/src/features/semanticViewer/components/RunSemanticQueryButton.tsx
```diff
@@ -9,101 +9,30 @@ import {
...
-        projectUuid,
-        results,
-        semanticQuery,
-        fields,
-    ]);
-
-    const handleSubmit = useCallback(
-        () => runSemanticViewerQuery(semanticQuery),
-        [semanticQuery, runSemanticViewerQuery],
-    );
```
IrakliJani, line 106: :chefkiss:

packages/frontend/src/features/semanticViewer/components/Content.tsx
```diff
@@ -102,86 +176,13 @@ const Content: FC = () => {
 
                 {!!view && <Filters />}
 
-                <RunSemanticQueryButton ml="auto" />
+                <RunSemanticQueryButton
+                    ml="auto"
+                    onClick={runSemanticLayerQuery}
```
IrakliJani, line 181: not important. A useful pattern that I find easy to read is `onClick` -> `handleSomething`. `handle` prefix means it's defined within the same component. If it's `onClick` -> `onClick`, then the `on` prefix tells the actual handler is defined somewhere in the parent component. Something without the `handle` or `on` prefix means it's a utility function of some kind and is not related to the React lifecycle at all.

packages/frontend/src/features/semanticViewer/components/ContentCharts.tsx
```diff
@@ -75,6 +85,50 @@ const ContentCharts: FC = () => {
...
+            })),
+            ...selectedFieldsByKind.timeDimensions.map((d) => ({
+                ...d,
+                kind: FieldKind.DIMENSION,
+            })),
+            ...selectedFieldsByKind.metrics.map((m) => ({
+                ...m,
+                kind: FieldKind.METRIC,
+            })),
+        ];
```
IrakliJani, line 102: this can be a selector

packages/frontend/src/features/semanticViewer/components/ContentCharts.tsx
```diff
@@ -113,50 +168,55 @@ const ContentCharts: FC = () => {
...
+                                id={`semantic-viewer-panel-tab-${TabPanel.VISUALIZATION_TABLE}`}
+                                collapsible
+                                order={2}
+                                defaultSize={25}
+                                minSize={10}
+                                onCollapse={() => setOpenPanel(undefined)}
+                            >
+                                <Table
+                                    resultsRunner={pivotResultsRunner}
+                                    thSortConfig={thSortConfig}
```
IrakliJani, line 200: just throwing some ideas: this does not feel right. I feel like pivot results runner should be responsible to provide both sorting config and columns config. or maybe `TableDataModel` is better suited for that. but we only use TableDataModel for visualization... we might need two more models for raw results and pivot results. happy to discuss.

packages/frontend/src/features/semanticViewer/components/ContentResults.tsx
```diff
@@ -53,6 +60,50 @@ const ContentResults: FC = () => {
         fields,
     ]);
 
+    const thSortConfig = useMemo(() => {
```
IrakliJani, line 63: this duplication also makes me think weather this should live inside components and not runner or a model

almeidabbm: That's a great point, I don't like this very much either, but the reason why I did this was so I could pass a `onClick` handler to take care of mutating the sort state, this calls a dispatch so it isn't really doable in the runner 🤔 maybe we can figure out another solution?

packages/frontend/src/features/semanticViewer/store/semanticViewerSlice.ts
almeidabbm: Achieved this behavior by converting the hook into a `useQuery` with `sortBy` and `filters` on the query key. Also the hook is only enabled if any fields are selected and the user has clicked the run button 💪

almeidabbm:
> s great! The only thing that was weird was when I removed all the sorts, the query reran.
> 
> I would expect that if I remove all of the sorting, the table stays in the same sort as whatever it

Thank you 😄 This has been fixed in the latest commit 👌

packages/frontend/src/features/semanticViewer/components/ContentResults.tsx
IrakliJani: Yep, we can talk about it, but I think this PR is ready to be merged. I wouldn't block a PR for this.

IrakliJani (approved):

PR #11596, opened 2024-09-18
fix: reset chart configs on unmount of create and view

joaoviana:
Closes: N/A

### Description:

Reset chart-related slices on unmount. The current `resetState` doesn't reset chart-related slices on `unmount`

### Reviewer actions


rephus (approved):
Not sure what issue this is trying to solve, but I tested this and it looks good

joaoviana:
> Not sure what issue this is trying to solve, but I tested this and it looks good

When users create multiple sql charts, or interact with the sql runner (without refreshing the page) this could lead to configs not being correctly unmounted, leading then to weird config states
This was fixed with the sql runner slice previously by resetting on unmount, but now that we have more slices, we should do that as well

Issue #11486, opened 2024-09-10
You can save an invalid chart, and get an error later


rephus:
### Description

I switched tables and I ended up with an invalid chart, however, the chart was looking fine in edit mode, but in view mode, I got the ìnvalid chart` error 

![image](https://github.com/user-attachments/assets/fba4136f-8ea2-42a1-b778-50f59ce30df7)

![image](https://github.com/user-attachments/assets/c702c8f4-8fc8-483e-a6a2-88d09e7fae85)

Suggested solutions: 
- Do not allow saving chart with errors 
- Show chart error in edit mode. 

### Steps to Reproduce the Bug or Issue

1. generate  a chart from a table
2. change thesql to get errors on fields
3. save
4. get sql error

### Cloud or self-hosting

None

joaoviana:
- The `sql` sent to the backend should be the sql set **after** running a successful query. At the moment, if I just edit the sql, and not run query, it doesn't matter. That newly-typed sql will be sent to the backend
- When running a new / editing a `sql` query, and if there are configuration errors with the selected viz config, we should alert the user that there are errors and that they need to be fixed before saving

PR #11605, opened 2024-09-18
feat: store last successful sql queries

joaoviana:
Relates to: https://github.com/lightdash/lightdash/issues/11486

### Description:

Introduces a very simple `createHistoryReducer` that lets you store specific properties of a rtk reducer in _history_
At the moment this just stores them in a max 5 length. This can be extendable to allow redo-undo.

Screen recording: 

https://github.com/user-attachments/assets/3c54e2b6-0383-428a-9463-5a0420b20ace

### Reviewer actions


rephus:
Why is the default behaviour to copy to the clipboard and not replace the SQL ?

packages/frontend/src/features/sqlRunner/store/history.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export interface WithHistory<T> {
+    current: T;
+    past: HistoryItem<NonNullable<T>>[];
+}
+
+const DEFAULT_MAX_HISTORY_ITEMS = 10;
+
+export const createHistoryReducer = <T>(
```
rephus, line 15: Correct me if I'm wrong, but because history is just a redux component,a nd not persisted in database, this information will be lost if you refresh the page, and it is not shared across users. 

This is similar than doing CTRL-Z, right ? 

Are we considering persisting this info in database or even in session storage ?

rephus:
This popup is not very useful, all my changes were happening at the end of the line, so all the states look the same to me. We should make this popup bigger or introduce some breaking lines 

image

packages/frontend/src/features/sqlRunner/store/history.ts
rephus: I've just tested this and if I refresh the page I lose all the history.

rephus:
I don't think this feature is related with https://github.com/lightdash/lightdash/issues/11486 .. it is a cool feature though.

joaoviana:
Made changes based on feedback 
image

joaoviana:
> This popup is not very useful, all my changes were happening at the end of the line, so all the states look the same to me. We should make this popup bigger or introduce some breaking lines
> 
> image

Made it bigger @rephus 

The popover is useful to see what was written previously, the window is wider. I'm not too sure on how to make the lines break

packages/frontend/src/features/sqlRunner/store/history.ts
joaoviana: > Are we considering persisting this info in database or even in session storage ?

Ideally it should be stored in the database when we start implementing draft state for this particular feature.

> This is similar than doing CTRL-Z, right ?

_Could_ be similar, but this only stores run queries, not code that is written, if that makes sense 👍 

> I've just tested this and if I refresh the page I lose all the history.

I'm ok with that, we also lose all config if we refresh the sql runner page when creating a new one IIRC

rephus (approved):

PR #11606, opened 2024-09-18
feat: alert user if unsaved sql change

joaoviana:
Relates to: https://github.com/lightdash/lightdash/issues/11486

### Description:

Detects when the user has unsaved sql changes, and if so, alert them that there are unsaved changes when they save their chart. 

Screen recording

https://github.com/user-attachments/assets/b295ae1c-c744-414e-a893-3efd9546bc40

### Reviewer actions


packages/frontend/src/features/sqlRunner/components/Header/HeaderEdit.tsx
```diff
@@ -121,6 +130,20 @@ export const HeaderEdit: FC = () => {
...
+                'sqlRunner',
+                'savedSqlChart',
+                projectUuid,
+                savedSqlChart?.slug,
+            ],
+        });
+        history.push(
+            `/projects/${projectUuid}/sql-runner/${savedSqlChart?.slug}`,
+        );
+    }, [queryClient, history, savedSqlChart, projectUuid]);
```
joaoviana, line 145: ensures that if the user goes back, that we reset the queries of the savedSqlChart

packages/frontend/src/features/sqlRunner/components/Header/HeaderEdit.tsx
```diff
@@ -168,31 +191,35 @@ export const HeaderEdit: FC = () => {
...
+                                    loading={isLoading}
+                                    onClick={onSaveClick}
+                                >
+                                    Save
+                                </Button>
+                            </HoverCard.Target>
+                            <HoverCard.Dropdown p={0} bg="yellow.0">
+                                <SqlQueryBeforeSaveAlert />
+                            </HoverCard.Dropdown>
+                        </HoverCard>
```
joaoviana, line 209: if there are unrun changes, just alert the user so they're aware

packages/frontend/src/features/sqlRunner/components/Header/HeaderEdit.tsx
```diff
@@ -168,31 +191,35 @@ export const HeaderEdit: FC = () => {
...
-                                    history.push(
-                                        `/projects/${savedSqlChart.project.projectUuid}/sql-runner/${savedSqlChart.slug}`,
-                                    )
-                                }
+                                size="md"
+                                onClick={handleGoBackToViewPage}
                             >
-                                Cancel
-                            </Button>
+                                <MantineIcon icon={IconArrowBack} />
```
joaoviana, line 221: small ux improvement, `cancel` sounds ambiguous

packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
```diff
@@ -153,61 +230,41 @@ export const SaveSqlChartModal: FC<Props> = ({ opened, onClose }) => {
...
-                        padding: theme.spacing.md,
-                    })}
-                >
-                    <Button
-                        onClick={onClose}
-                        variant="outline"
-                        disabled={isCreatingSavedSqlChart}
+            {showWarning ? (
+                <Box>
+                    <SqlQueryBeforeSaveAlert />
```
joaoviana, line 243: if there are unrun changes, alert the user, but let them save the latest `sql` regardless.

packages/frontend/src/features/sqlRunner/hooks/useSqlChartResults.ts
```diff
@@ -79,7 +79,7 @@ export const useSqlChartResults = (
             });
         },
         {
-            enabled: Boolean(slug),
+            enabled: Boolean(slug) && Boolean(projectUuid),
```
joaoviana, line 82: was missing from this query, leading to duplicate requests

packages/frontend/src/pages/ViewSqlChart.tsx
```diff
@@ -114,6 +114,7 @@ const ViewSqlChart = () => {
         projectUuid,
         slug: params.slug,
         limit: sqlChart?.limit,
+        additionalQueryKey: [params.slug, sql],
```
joaoviana, line 117: This ensures that every time we create a sql chart, run a query, then change it (without running) then save, it won't get the cached results.

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -162,13 +178,20 @@ export const sqlRunnerSlice = createSlice({
...
             state.name = action.payload;
         },
         setSql: (state, action: PayloadAction<string>) => {
             state.sql = action.payload;
+
+            const normalizedNewSql = normalizeSQL(action.payload);
+            const normalizedCurrentSql = normalizeSQL(
+                state.successfulSqlQueries.current || '',
+            );
+            state.hasUnrunChanges = normalizedNewSql !== normalizedCurrentSql;
```
joaoviana, line 194: this was mentioned by you @rephus on another PR, this uses `sql-formatter` as the comparison tool between previous and current sql queries. If there's extra whitespace, then it shouldn't consider them as different. Updated the history adapter to accept a comparison function when removing duplicates.

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -20,6 +21,16 @@ export enum SidebarTabs {
...
 }
 
+/**
+ * Normalizes the SQL by removing all whitespace and formatting it consistently - helpful for comparing two SQL queries
+ * @param sql
+ * @returns formatted SQL
+ */
+const normalizeSQL = (sql: string): string =>
+    format(sql, {
+        language: 'sql',
```
rephus, line 31: is this compatible with all warehouses syntax ?

rephus: Tested on bigquery, it works well

rephus (approved):
Nice. 

design question: Should we be more consistent with the highlighting ? 
image

joaoviana:
> Nice.
> 
> design question: Should we be more consistent with the highlighting ? image

That's a good question, a `yellow` indicator instead?
Will get @PriPatel 's thoughts on this and we can define a way forward 💪

PriPatel:
> design question: Should we be more consistent with the highlighting ?

Sure, Can you explain what the red dot means and what the yellow text means? Then i can have a think - thanks!

PR #11607, opened 2024-09-19
refactor(wip): combine runners

owlas:
- merging runners into one (some subclassing just for constructors)
- query handling not finished for sql runner


packages/common/src/types/semanticLayer.ts
```diff
@@ -72,6 +72,9 @@ export type SemanticLayerQuery = {
     timezone?: string;
     pivot?: SemanticLayerPivot;
     filters: SemanticLayerFilter[];
+    customMetrics?: (Pick<SemanticLayerField, 'name' | 'aggType'> & {
```
owlas, line 75: This is the common interface for pivotQueries - it needs to support custom metrics

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -85,8 +86,8 @@ export class CartesianChartDataModel {
     }
 
     getDefaultLayout(): PivotChartLayout | undefined {
-        const dimensions = this.resultsRunner.getDimensions();
-        const metrics = this.resultsRunner.getMetrics();
+        const dimensions = this.resultsRunner.getPivotQueryDimensions();
```
owlas, line 89: `.getDimensions` and `.getMetrics` were redundant

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -3,37 +3,23 @@ import {
...
-    }
-
-    return {
-        on: [config.x.reference],
-        index: config.groupBy?.map((groupBy) => groupBy.reference) ?? [],
-        values: config.y.map((y) => y.reference),
-    };
-};
-
+// not useful - semantic layer field type should be source of truth
```
owlas, line 22: Almost all of these "conversion functions" are needed because we don't have unified types across both runners, let's make it a priority to remove them

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -64,15 +51,74 @@ function getVizIndexTypeFromDimensionType(
...
+            return {
+                reference: columnName,
+                type: dimType,
+            };
+        })
+        .filter((c): c is VizColumn => Boolean(c));
+};
+
+// This fields dependency should be fixed by fixing the API for semantic layer
+export const getPivotQueryFunctionForSemanticViewer = (
```
owlas, line 80: This function has to vary per runner because we need to refactor more of query execution

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -81,22 +127,20 @@ export class SemanticViewerResultsRunner implements IResultsRunner {
 
     private readonly metrics: SemanticLayerField[];
 
+    private readonly runPivotQuery: RunPivotQuery;
```
owlas, line 130: abstracted run operation

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -64,15 +51,74 @@ function getVizIndexTypeFromDimensionType(
...
+        return {
+            results,
+            indexColumn,
+            valuesColumns,
+            columns: vizColumns,
+        };
+    };
+};
 
+export class BaseResultsRunner implements IResultsRunner {
```
owlas, line 121: This is used by both runners!

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -112,119 +156,98 @@ export class SemanticViewerResultsRunner implements IResultsRunner {
         );
     }
 
-    getDimensions(): VizIndexLayoutOptions[] {
```
owlas, line 115: duplicate

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -112,119 +156,98 @@ export class SemanticViewerResultsRunner implements IResultsRunner {
...
-            ),
-        }));
-    }
-
-    getMetrics(): VizValuesLayoutOptions[] {
-        return this.metrics.map((metric) => ({
-            reference: metric.name,
-            aggregation: metric.aggType || VIZ_DEFAULT_AGGREGATION,
-        }));
+    async getPivotedVisualizationData(
```
owlas, line 159: most of this is abstracted by the pivotqueryfn - ideally it wouldn't be!

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -112,119 +156,98 @@ export class SemanticViewerResultsRunner implements IResultsRunner {
...
             reference: dimension.name,
-            axisType: getVizIndexTypeFromDimensionType(dimension.type),
+            axisType: getVizIndexTypeFromSemanticLayerFieldType(dimension.type),
             dimensionType: getDimensionTypeFromSemanticLayerFieldType(
                 dimension.type,
             ),
         }));
     }
 
     getPivotQueryMetrics(): VizValuesLayoutOptions[] {
```
owlas, line 184: we don't need this to vary per runner - it supports both. sql runner simply has no metrics

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -112,119 +156,98 @@ export class SemanticViewerResultsRunner implements IResultsRunner {
...
+        // returns empty for sql runner because there's no metrics
         return this.metrics.map((metric) => ({
             reference: metric.name,
             aggregation: metric.aggType || VIZ_DEFAULT_AGGREGATION,
-            aggregationOptions: [],
         }));
     }
 
     getPivotQueryCustomMetrics(): VizCustomMetricLayoutOptions[] {
```
owlas, line 192: mostly logic pulled from sql-runner. it works for semantic layer too, we just don't use it yet

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
```diff
@@ -239,3 +262,27 @@ export class SemanticViewerResultsRunner implements IResultsRunner {
         return this.rows;
     }
 }
+
+export class SemanticViewerResultsRunnerFrontend extends BaseResultsRunner {
```
owlas, line 266: So in the frontend we use react-query etc. to run the pivot. in the BE this might be different

packages/frontend/src/features/sqlRunner/runners/SqlRunnerResultsRunnerFrontend.ts
```diff
@@ -0,0 +1,234 @@
...
+        valuesColumns: values.map((value) => ({
+            reference: value.name,
+            aggregation: value.aggType ?? VIZ_DEFAULT_AGGREGATION,
+        })),
+        groupByColumns: groupBy?.map((f) => ({ reference: f.name })),
+    };
+};
+
+// TEMPORARY
+export const getPivotQueryFunctionForSqlRunner = ({
```
owlas, line 131: different execution for sql runner, it mostly maps the sql runner types to a semantic layer query

packages/frontend/src/features/sqlRunner/runners/SqlRunnerResultsRunnerFrontend.ts
```diff
@@ -0,0 +1,234 @@
...
+    }: {
+        columns: VizColumn[];
+        rows: RawResultRow[];
+        projectUuid: string;
+        limit: number;
+        sql: string;
+        slug?: string;
+        uuid?: string;
+    }) {
+        const fields: SemanticLayerField[] = columns.map((column) => ({
```
owlas, line 208: Creates the fields - makes all the base logic shared for semantic viewer + sql runner

packages/frontend/src/features/sqlRunner/runners/SqlRunnerResultsRunnerFrontend.ts
```diff
@@ -0,0 +1,234 @@
...
+            kind: FieldType.DIMENSION,
+            name: column.reference,
+            type: getSemanticLayerFieldTypeFromDimensionType(
+                column.type || DimensionType.STRING,
+            ),
+            visible: true,
+            label: column.reference,
+            // why are these required?
```
owlas, line 216: these should be optional types

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: Agreed. This seems like the hardest part, no?

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: We should put it in it's own file

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: Fine by me. My only counter thought is that having it in both places makes it very clear what you'd override in the future if the implementations diverge. But we can get to that then. 

This is fine for now since we are combining the runners.

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: Cool

magnew (approved):

packages/frontend/src/features/semanticViewer/runners/SemanticViewerResultsRunner.ts
magnew: But do you anticipate this always working the same way on both pages? If they always share this, any time we change one, we'll be changing the other.

packages/frontend/src/features/sqlRunner/runners/SqlRunnerResultsRunnerFrontend.ts
```diff
@@ -0,0 +1,234 @@
...
+            columns,
+        };
+    };
+};
+
+export class SqlRunnerResultsRunnerFrontend
+    extends BaseResultsRunner
+    implements IResultsRunner
+{
+    constructor({
```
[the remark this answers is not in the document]
magnew: So this runner would get instantiated with a SemanticLayerQuery (or whatever we rename that to)? 

Then we'd remove `convertSemanticLayerQueryToSqlRunnerPivotQuery` and use SemanticLayerQuery for sql queries? 

I think we'll need a converter from SQL runner pivot queries to SemanticLayerQueries. OR, we'll have to move more logic into this runner so it is the only place that runPivotQuery gets called. At the moment there is a hook (used on both pages) that calls the pivot query function, so the SQL runner page will need to do that same thing that this runner is doing to create the RunPivotQuery function. Ideally we get rid of that hook and these runners handle that.

magnew:
Closing in favor of #11611

Issue #11496, opened 2024-09-10
There is an empty users page included in the `users` list


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/e8af8b79783440c7b28ccf918b0582cb

### Steps to Reproduce the Bug or Issue

1. go to the `users` page in hte org settings
2. go to the last page
3. it's empty

### Cloud or self-hosting

None

PR #11608, opened 2024-09-19
fix: total count pagination bug

rephus:
Closes:  #11496

#### What the issue is: 
By removing select and group, we are not filtering out duplicated data on the results, so the count can be innacurate. The worst thing is that this bug is hard to catch, and it will depend on the `query` que pass as parameter. We need a solution that is consistent and error-proneless

#### How I fixed this: 
Replace the count custom query with a CTE that simply counts all the results in the original subquery. Hopefully, this is more generic and less hacky. 

#### Considerations: 
This is virtually running the same query twice, one just for counting results. However, this is not different to what we had previosly anyway...

An alternative option is to get all the results from the query, count and filter on javascript, this is potentially more memory heavy and not a big improvement... 

### Reviewer actions

### DIFF
--- a/packages/backend/src/database/pagination/index.ts
+++ b/packages/backend/src/database/pagination/index.ts
@@ -26,24 +26,24 @@ export default class KnexPaginate {
             const totalRecordsCountPromise = query.client.raw(
                 `
                 WITH count_cte AS (?)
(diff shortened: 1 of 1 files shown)


almeidabbm (approved):
LGTM 🚀

Issue #11629, opened 2024-09-20
Similar charts from SQL runner override each other


jpetey75:
### Description

I'm not sure exactly what's happening, but it looks like the most recent chart I create in SQL runner appears when I create a new query that is similar. Also, if both similar charts are in a dashboard, the same chart is shown twice rather than showing both charts correctly.

### Steps to Reproduce the Bug or Issue

Video: https://www.loom.com/share/5390141831d54006bc244e5060d135f7

Also see internal customer convo here: https://lightdash.slack.com/archives/C04KD95TF0U/p1726676859450469

### Cloud or self-hosting

cloud

PR #11634, opened 2024-09-23
fix: add additional query key to dashboard sql tile

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/11629

### Description:

Add query keys to `useChartViz` dashboard sql tiles: uuid & slug
Add query keys to `useChartViz` to Content panel charts: `sql` 

This ensures we don't return cached data for different charts, by applying unique query keys.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx
@@ -100,6 +100,7 @@ const SqlChartTile: FC<Props> = ({ tile, isEditMode, ...rest }) => {
         sql: data?.chart.sql,
         slug: data?.chart.slug,
         limit: data?.chart.limit,
+        additionalQueryKey: [data?.chart.slug, data?.chart.sql, savedSqlUuid],
     });
 
     if (isLoading) {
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx
```diff
@@ -100,6 +100,7 @@ const SqlChartTile: FC<Props> = ({ tile, isEditMode, ...rest }) => {
         sql: data?.chart.sql,
         slug: data?.chart.slug,
         limit: data?.chart.limit,
+        additionalQueryKey: [data?.chart.slug, data?.chart.sql, savedSqlUuid],
```
IrakliJani, line 103: This is not related to the PR, and I know this prop already exists. To me, it feels like a hack. Besides, we are already passing `slug` and `sql` to the `viz` method. We might need to consider caching from a different perspective rather than hacking in additional keys every time we use this hook. However, I think we should first wait for this PR to get merged: https://github.com/lightdash/lightdash/pull/11611

IrakliJani (approved):

packages/frontend/src/components/DashboardTiles/DashboardSqlChartTile.tsx
magnew: Yea, this was my bad hack. Sorry. It looks like I'll take it out in the linked PR

Issue #11529, opened 2024-09-12
Download image of charts made in SQL runner


jpetey75:
### Description

Add image download for charts in SQL runner, just like we have in charts from the query builder.

PR #11639, opened 2024-09-23
feat: allow download charts as image,pdf

joaoviana:
Closes: [#11529](https://github.com/lightdash/lightdash/issues/11529)

### Description:

Allows downloading charts as image,pdf,json.etc - uses the same component as the Explorer

Demo: 

https://github.com/user-attachments/assets/53d131c8-f5e2-4a4a-aab1-771f60dcd61b

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/ChartDownload.tsx
+++ b/packages/frontend/src/components/ChartDownload.tsx
@@ -34,7 +34,7 @@ import { useVisualizationContext } from './LightdashVisualization/VisualizationP
 
 const FILE_NAME = 'lightdash_chart';
 
-enum DownloadType {
+export enum DownloadType {
     JPEG = 'JPEG',
     PNG = 'PNG',
     SVG = 'SVG',
@@ -121,9 +121,11 @@ function downloadPdf(base64: string, width: number, height: number) {
 
 type DownloadOptions = {
     getChartInstance: () => EChartsInstance | undefined;
+    unavailableOptions?: DownloadType[];
 };
 export const ChartDownloadOptions: React.FC<DownloadOptions> = ({
     getChartInstance,
+    unavailableOptions,
 }) => {
     const [type, setType] = useState<DownloadType>(DownloadType.PNG);
     const [isBackgroundTransparent, setIsBackgroundTransparent] =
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/ChartDownload.tsx
```diff
@@ -215,18 +209,14 @@ const ChartDownloadOptions: React.FC<DownloadOptions> = ({
...
-                    size="xs"
-                    ml="auto"
-                    leftIcon={<MantineIcon icon={IconDownload} />}
-                    onClick={onDownload}
-                >
-                    {' '}
-                    Download
-                </Button>
-            )}
+            <Button
```
joaoviana, line 212: we would never reach this statement. It's moved down below before rendering this component.

packages/frontend/src/features/sqlRunner/components/Download/ChartDownload.tsx
```diff
@@ -0,0 +1,83 @@
...
+                                label={capitalize(DownloadFileType.CSV)}
+                                my="xs"
+                            />
+                            <Radio
+                                value={DownloadFileType.IMAGE}
+                                label={capitalize(DownloadFileType.IMAGE)}
+                            />
+                        </Radio.Group>
+                        {downloadFormat === DownloadFileType.IMAGE && (
+                            <ChartDownloadOptions
```
joaoviana, line 64: use the same chart download options when choosing `image` download format.

rephus (approved):
it works great. 

Can we remo the JSON option though ? 

it looks like it is the results in JSON format, but it looks more like the echart configa, and also, it is not an "image" format...
Instead, we could add a proper JSON format , next to CVS and image, to download the results

image

joaoviana:
> it works great.
> 
> Can we remo the JSON option though ?
> 
> it looks like it is the results in JSON format, but it looks more like the echart configa, and also, it is not an "image" format... Instead, we could add a proper JSON format , next to CVS and image, to download the results (on a separate ticket perhaps)
> 
> image

Great point @rephus , removed that option on sql runner charts 👍 

We can add the option later in a separate label when we feel the need from user feedback 🙌 


Screenshot 2024-09-23 at 15 01 16
Removed in sql runner 🆙

Issue #11638, opened 2024-09-23
I would like to change the chart type/style of a series


joaoviana:
### Description

Example to implement: 
<img width="247" alt="image" src="https://github.com/user-attachments/assets/9efa252c-b705-4142-bde1-0f68c4773301">

Acceptance criteria
- [ ] When I have multiple series, I can set the chart type for each: currently, we only have `line` and `bar`, so let's implement for those first
- [ ] The series chart type, by default, is set to the selected chart type

### What problem does this solve?

Feature parity

PR #11648, opened 2024-09-23
feat: allow changing the chart type of each series

joaoviana:
Closes: [#11638](https://github.com/lightdash/lightdash/issues/11638)

### Description:

- Can change the chart type of each series
- It's quick 
- When you have 2 series and then delete one, it resets the chosen chart type to ensure the visualization matches. The user can just switch the base chart type as they wish. 

Demo:

https://github.com/user-attachments/assets/70e699ce-3ccb-4222-a25d-5a1ea3e9c88a

### Reviewer actions

### DIFF
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -584,13 +584,11 @@ export const getChartKind = (
 };
 
 export const getEChartsChartTypeFromChartKind = (
-    chartKind: Extract<
-        ChartKind,
+    chartKind:
         | ChartKind.VERTICAL_BAR
         | ChartKind.LINE
         | ChartKind.AREA
-        | ChartKind.SCATTER
-    >,
+        | ChartKind.SCATTER,
 ): CartesianSeriesType => {
     switch (chartKind) {
         case ChartKind.VERTICAL_BAR:
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -583,6 +583,32 @@ export const getChartKind = (
     }
 };
 
+export const getEChartsChartTypeFromChartKind = (
+    chartKind: Extract<
+        ChartKind,
+        | ChartKind.VERTICAL_BAR
+        | ChartKind.LINE
+        | ChartKind.AREA
+        | ChartKind.SCATTER
+    >,
+): CartesianSeriesType => {
+    switch (chartKind) {
+        case ChartKind.VERTICAL_BAR:
+            return CartesianSeriesType.BAR;
+        case ChartKind.LINE:
+            return CartesianSeriesType.LINE;
+        case ChartKind.AREA:
+            return CartesianSeriesType.AREA;
+        case ChartKind.SCATTER:
+            return CartesianSeriesType.SCATTER;
+        default:
+            return assertUnreachable(
+                chartKind,
+                `Unknown chart kind: ${chartKind}`,
+            );
+    }
+};
+
 export type ChartSummary = Pick<
     SavedChart,
     | 'uuid'
--- a/packages/common/src/visualizations/CartesianChartDataModel.ts
+++ b/packages/common/src/visualizations/CartesianChartDataModel.ts
@@ -4,7 +4,11 @@ import {
     Format,
     friendlyName,
 } from '../types/field';
-import { ChartKind, ECHARTS_DEFAULT_COLORS } from '../types/savedCharts';
+import {
+    ChartKind,
+    ECHARTS_DEFAULT_COLORS,
+    getEChartsChartTypeFromChartKind,
+} from '../types/savedCharts';
 import { applyCustomFormat } from '../utils/formatting';
 import {
     type VizCartesianChartConfig,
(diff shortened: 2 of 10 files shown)


packages/common/src/types/savedCharts.ts
```diff
@@ -583,6 +583,32 @@ export const getChartKind = (
     }
 };
 
+export const getEChartsChartTypeFromChartKind = (
```
joaoviana, line 586: our `ChartKind` types have a different value from the `Echarts`' ones. I've created a util bc I haven't seen one currently being used.

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -168,35 +172,41 @@ export class CartesianChartDataModel
...
+                        : undefined;
 
                 return {
                     dimensions: [
                         transformedData.indexColumn?.reference,
                         seriesColumn,
                     ],
-                    type: defaultSeriesType,
+                    type:
+                        seriesChartType && !isSingleAxis
```
joaoviana, line 203: if singleAxis, just ignore the custom chart type

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -306,6 +317,7 @@ export type CartesianChartDisplay = {
             format?: Format;
             yAxisIndex?: number;
             color?: string;
+            type?: Extract<ChartKind, ChartKind.LINE | ChartKind.VERTICAL_BAR>;
```
joaoviana, line 320: we only support these types for now

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
```diff
@@ -39,67 +39,94 @@ export const CartesianChartSeries: React.FC<SeriesColorProps> = ({
...
-                                <Group spacing="two" noWrap>
-                                    <Tooltip
-                                        label="Format"
-                                        variant="xs"
-                                        withinPortal
-                                    >
-                                        <MantineIcon
-                                            color="gray.5"
-                                            icon={IconBrush}
+                    {series.map(
```
joaoviana, line 42: this `.entries` wasn't necessary at all

packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
```diff
@@ -30,9 +30,13 @@ export const Sidebar: FC<Props> = ({ setSidebarOpen }) => {
...
 
-    const { selectedChartType, activeSidebarTab, sqlColumns } = useAppSelector(
-        (state) => state.sqlRunner,
+    const selectedChartType = useAppSelector(
+        (state) => state.sqlRunner.selectedChartType,
     );
+    const activeSidebarTab = useAppSelector(
+        (state) => state.sqlRunner.activeSidebarTab,
+    );
+    const sqlColumns = useAppSelector((state) => state.sqlRunner.sqlColumns);
```
joaoviana, line 39: small optimisation

joaoviana:
`tsoa` needed to be generated so the API knows what is the payload, response etc.
`tsoa` doesn't like `Extract`s so I've replaced them here: https://github.com/lightdash/lightdash/pull/11648/commits/ba7de34e0508733cdd362a3356e860852f392c8b

packages/frontend/src/components/DataViz/config/CartesianChartStyling.tsx
```diff
@@ -45,23 +45,22 @@ export const CartesianChartStyling = ({
...
+            const seriesLabel = foundSeries?.label;
+            const seriesColor = foundSeries?.color;
+            const seriesType = foundSeries?.type;
             return {
                 reference: f.reference,
                 format: seriesFormat,
-                label: seriesLabel,
+                label:
+                    seriesLabel ??
+                    friendlyName(`${f.reference}_${f.aggregation}`),
```
andrewdw, line 61: f.aggrigation is optional so would it need a fallback if it ends up undefined?

packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
andrewdw:
```suggestion
    const { selectedChartType, activeSidebarTab, sqlColumns } = useAppSelector(
        (state) => ({
            selectedChartType: state.sqlRunner.selectedChartType,
            activeSidebarTab: state.sqlRunner.activeSidebarTab,
            sqlColumns: state.sqlRunner.sqlColumns,
        }),
    );
```

By selecting only the necessary parts of the state, we can avoid potential unnecessary re-renders. The original was just missing the selection of child objects

andrewdw: On second thought it now no longer fully couples the param name to the type which could cause issues later if we rename things. so feel free to ignore my suggestion if that is a concern

andrewdw (approved):
Looks good! Just left a couple comments but approving since I don't believe they could/would break anything

packages/frontend/src/features/sqlRunner/components/Sidebar.tsx
joaoviana: > By selecting only the necessary parts of the state, we can avoid potential unnecessary re-renders. The original was just missing the selection of child objects

the suggested approach is returning always a new reference after an action dispatch, causing unnecessary rerenders after action dispatches. This was really useful: https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization see the first: `completedTodos` example. 
```
The default comparison is a strict === reference comparison.
```
https://react-redux.js.org/api/hooks#equality-comparisons-and-updates

We can use another package like `reselect` to change the comparison logic tho 👍

packages/frontend/src/components/DataViz/config/CartesianChartStyling.tsx
joaoviana: Good spot, I believe this could be `undefined` for the semantic viewer, but will be defined for the sql runner. Since this is shared, I'm going to leave it, if that's ok! :)

andrewdw: Makes sense, sounds good!

Issue #11643, opened 2024-09-23
Allow custom limit on results table


joaoviana:
### Description

Currently we preview the first 500 rows when a user runs a sql query. We should allow them to export the csv results with a custom limit, up to the max limit allowed in their org.

<img width="1555" alt="image" src="https://github.com/user-attachments/assets/b7149150-ee01-414b-8065-b4bb0006772a">

Acceptance criteria
- When creating/editing: 
- [ ] When I'm on the sql  click on the download icon and I'm previewing the first 500 rows, I can customise the limit applied to be over that default limit
- [ ] When I view a Table chart, I want to be able to do the same 

**Nice-to-haves:**
- When I'm editing: 
- [ ] When I view a Saved chart/table, I want to be able to export the underlying data results (non-aggregated) with a custom sql 
- [ ] + As a separate feature request, one thing that is helpfullis to see how many rows of data a query results to. This would allow users to ensure they are pulling and self serving with a complete data set! Ex: I run sql, and I see that my results yielded to 250k data points (if too hard, this can be done on a separate PR)

PR #11666, opened 2024-09-24
feat: allow setting custom sql on results download

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/11643

### Description:

When on the SQL tab or when visualising a Table, users can: 
- Apply a custom limit to their query so that they can download the results as csv

Recording:

https://github.com/user-attachments/assets/400bc98d-4c6b-4044-a2bf-2124387421da

### Reviewer actions


rephus (approved):
This is good for now, and it solves the ticket. 

image

ideally , we should add a toggle to downloda all the data, without having to specify the limit yourself. A workaround is to huse a ridicously high number every time

joaoviana:
> This is good for now, and it solves the ticket.
> 
> image
> 
> ideally , we should add a toggle to downloda all the data, without having to specify the limit yourself. A workaround is to huse a ridicously high number every time

So could we this high number be the LIGHTDASH_CSV_CELLS_LIMIT?

Issue #11284, opened 2024-08-28
add `context` to `query executed` analytics events for scheduled deliveries


TuringLovesDeathMetal:
More context here: https://lightdash.slack.com/archives/C02GQKJK84Q/p1724338405375889

- [ ] add `context` = `scheduledDashboard` for queries that are executed because of a dashboard scheduled delivery. (I'd actually be happy to just have _**all**_ charts + dashboards using a `scheduledDelivery` context instead of having `scheduledChart` and `scheduledDashboard` as separate events if that's easier!)
- [ ] add `context` = `alert` for queries that are executed because they are used for threshold alerting

PR #11683, opened 2024-09-25
feat: add context to deliveries

rephus:
Closes: https://github.com/lightdash/lightdash/issues/11284

### Description:

On charts: 
Screenshot from 2024-09-26 15-49-35
On dashboards: 
Screenshot from 2024-09-26 16-24-54

### Reviewer actions


packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -185,6 +187,24 @@ export const getContextFromHeader = (req: Request) => {
...
+    req: Request,
+): QueryExecutionContext | undefined => {
+    const context: QueryExecutionContext | undefined =
+        typeof req.query.context === 'string'
+            ? (req.query.context as QueryExecutionContext)
+            : undefined;
+    if (context) {
+        for (const [key, value] of Object.entries(QueryExecutionContext)) {
+            if (value.toLowerCase() === context.toLowerCase()) {
+                return value as QueryExecutionContext;
```
rephus, line 200: only return valid QueryExecutionContext

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -935,8 +955,8 @@ export type DownloadCsv = BaseTrack & {
         context?:
```
rephus, line 955: we should probably standarize this a bit more and reuse the QueryExecutionContext type

packages/frontend/src/hooks/dashboard/useDashboardChart.ts
```diff
@@ -18,6 +20,7 @@ const useDashboardChart = (tileUuid: string, savedChartUuid: string | null) => {
         invalidateCache,
         granularity,
         isAutoRefresh,
+        context,
```
joaoviana, line 23: should we also include a similar change to `useDashboardSqlChart`? Since these can be added and scheduled in dashboards?

rephus:
> do we want to add this context to dashboard.views?
That was not part of this PR, we could, for sure, how useful is that @TuringLovesDeathMetal  ? 
In any case, this is something it can be added on top of this change, so I'd suggest to do it on a seaprate ticket/PR

packages/frontend/src/hooks/dashboard/useDashboardChart.ts
rephus: I'll do it in this PR, but first I'll have to refactor query context and move it to common because it has some dependency issues.

rephus:
> set up an alert for a chart

You were right, the condition for the alert was wrong. I fixed it

joaoviana (approved):
Works for alerts 💪 
image

Issue #11675, opened 2024-09-25
Missing space with slug "(...)" to promote


ceyzeriat:
### Description

I was promoting a dashboard and had a red toast pop-up failure `Failed to promote dashboard -- Missing space with slug "company-kpis" to promote`.
The promotion needed to create the space, the dashboard, and the multiple charts.
Interestingly, the Space got created. A second promotion attempt led to the same error, and creating a duplicate space.

### Steps to Reproduce the Bug or Issue

1. Create a dashboard in a Space, with empty spaces in their names
2. Promote the dashboard where the Space + Dashboard need to be created
3. Error

### version

v0.1269.1

### Cloud or self-hosting

cloud

rephus:
Thanks for creating this issue, I've identified the problem and we will fix it soon.

PR #11684, opened 2024-09-25
fix: force same slug on spaces in different project

rephus:
Closes:  #11675

### Description:

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -1371,7 +1371,7 @@ export class SpaceModel {
         userId: number,
         isPrivate: boolean,
         slug: string,
-        forceSameSlug?: boolean,
+        forceSameSlug: boolean = false,
     ): Promise<Space> {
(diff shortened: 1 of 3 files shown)


packages/backend/src/models/SpaceModel.ts
```diff
@@ -1383,7 +1384,9 @@ export class SpaceModel {
                     is_private: isPrivate,
                     name,
                     created_by_user_id: userId,
-                    slug: await generateUniqueSlug(trx, SpaceTableName, slug),
+                    slug: forceSameSlug
```
rephus, line 1387: generateUniqueSlug doesn't check if the slug are unique within a project

On promotion, sometimes we create spaces when promoting, this means, the new space in the upstream project had  adifferent slug. We need to force the slug to be the same, otherwise promotion is useless. 

Another option is to have ` generateUniqueSlug` to search for unique slugs within a project, and throw errors on the service if they are not the same, but this seems a bit overcomplicated.

joaoviana (approved):

Issue #11685, opened 2024-09-25
`is not` is cut off in the filter modal


TuringLovesDeathMetal:
### Description

<img width="521" alt="image" src="https://github.com/user-attachments/assets/7d39ce92-8d4d-4e41-854c-e1ff02cd64ec">

### Steps to Reproduce the Bug or Issue

see above

### Cloud or self-hosting

None

Issue #11438, opened 2024-09-05
`nested filter` copy is confusing 


TuringLovesDeathMetal:
### Description

Right now, there are two places where I can `add nested filter`. But, the two buttons have two different actions. 

<img width="913" alt="image" src="https://github.com/user-attachments/assets/ea2dd6ea-a965-44dd-b553-370937c078de">

https://www.loom.com/share/4278d78436ad44378c75faf119173408

We should rename the blue `add nested filter` to just say `+ add filter`

PR #11688, opened 2024-09-25
fix: semantic viewer filter ui bugs

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/11685
Closes: https://github.com/lightdash/lightdash/issues/11438
Bonus: fixed text alignment inconsistencies

fyi, this padding should be enough on all fonts/os variations I think:

CleanShot 2024-09-25 at 19 25 21@2x

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/semanticViewer/components/FiltersModal/Filter/MultiStringFilter.tsx
+++ b/packages/frontend/src/features/semanticViewer/components/FiltersModal/Filter/MultiStringFilter.tsx
@@ -4,7 +4,7 @@ import {
     type SemanticLayerFilter,
     type SemanticLayerStringFilter,
 } from '@lightdash/common';
-import { Group, Select, type SelectItem } from '@mantine/core';
+import { Group, rem, Select, type SelectItem } from '@mantine/core';
 import { useMemo, type FC } from 'react';
 import FilterMultiStringInput from '../../../../../components/common/Filters/FilterInputs/FilterMultiStringInput';
 import FilterFieldSelect from '../FilterFieldSelect';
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/semanticViewer/components/FiltersModal/Filter/MultiStringFilter.tsx
```diff
@@ -54,19 +54,15 @@ const MultiStringFilter: FC<MultiStringFilterProps> = ({
...
+                styles={{ input: { paddingRight: rem(20) } }}
                 withinPortal
                 data={operatorsOpts ?? []}
                 value={filter.operator}
-                onChange={(value: SemanticLayerFilter['operator'] | null) => {
-                    if (!value) {
-                        return;
-                    }
-
+                onChange={(value: SemanticLayerFilter['operator']) => {
```
almeidabbm, line 63: The default argument for `onChange` in the select component is `string | null`, not sure if removing it won't allow some error to happen

image

packages/frontend/src/features/semanticViewer/components/FiltersModal/Filter/TimeFilter.tsx
```diff
@@ -80,19 +80,15 @@ const RelativeTimeFilter: FC<RelativeTimeFilterProps> = ({
...
+                styles={{ input: { paddingRight: rem(20) } }}
                 withinPortal
                 data={operatorsOpts ?? []}
                 value={filter.operator}
-                onChange={(value: SemanticLayerFilter['operator'] | null) => {
-                    if (!value) {
-                        return;
-                    }
-
+                onChange={(value: SemanticLayerFilter['operator']) => {
```
almeidabbm, line 89: same as above

almeidabbm (approved):
Approved with a small question 💪

packages/frontend/src/features/semanticViewer/components/FiltersModal/Filter/MultiStringFilter.tsx
IrakliJani: can't be a null because it does not have `clearable`. unfortunately, mantine is not strongly typed that way

almeidabbm: Gotcha, makes sense 👌 
We can disregard this comment then

Issue #11627, opened 2024-09-20
Copy SQL runner content when creating preview environments.


rephus:
### Description

We copy most of the content from prod to preview, but this is not happening for SQL runner charts. 

This is what you get on a dashboard on preview environments

![image](https://github.com/user-attachments/assets/3518d438-ae23-457b-a1b4-856ecbe295e6)

See thread https://lightdash.slack.com/archives/C0797GG4TPH/p1726840818270339

PR #11694, opened 2024-09-26
fix: copy sql runner content with preview env

andrewdw:
Closes: #11627

### Description:
- Saved sql-runner charts now properly get duplicated in to the preview environment.

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -1371,16 +1371,17 @@ export class ProjectModel {
                         );
                     }
                     // Generate the slug asynchronously
-                    const uniqueSlug = await generateUniqueSlug(
-                        trx,
-                        'saved_sql',
-                        d.slug, // using the existing slug as a base - preventing naming duplicates
-                    );
+                    // const uniqueSlug = await generateUniqueSlug(
+                    //     trx,
+                    //     'saved_sql',
+                    //     d.slug, // using the existing slug as a base - preventing naming duplicates
+                    // );
                     // Map the saved SQL to the new saved SQL
                     const createSavedSQL: CloneSavedSQL = {
                         ...d,
+                        project_uuid: previewProjectUuid,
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1334,6 +1336,182 @@ export class ProjectModel {
                           .returning('*')
                     : [];
 
+            // .dP"Y8    db    Yb    dP 888888 8888b.      .dP"Y8  dP"Yb  88
```
rephus, line 1339: :sweat_smile: original

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1334,6 +1336,182 @@ export class ProjectModel {
...
+                }
+                // Create an array of promises for generating slugs and mapping saved SQLs
+                const mappedSavedSQLsPromises = savedSQLList.map(async (d) => {
+                    if (!d.space_uuid) {
+                        throw new Error(
+                            `Chart ${d.saved_sql_uuid} has no space_uuid`,
+                        );
+                    }
+                    // Generate the slug asynchronously
+                    const uniqueSlug = await generateUniqueSlug(
```
rephus, line 1374: I don't think this will work, since the `generateUniqueSlug` doesn't get unique slugs per project. We need to force these slugs to be the same (in different projects) in order for the promotion to work. (see how we force the slug on spaces)

rephus (commented):
The code looks great, I don't think you're missing anything, we just need to fix the slug in different project. I'll test it once that is done.

rephus (approved):
This is perfect, great job. 
Screenshot from 2024-10-01 09-16-29

Slugs are the same on DB

Screenshot from 2024-10-01 09-17-16

I recorded a video testing promotion, it worked great

[Screencast from 01-10-24 09:17:44.webm](https://github.com/user-attachments/assets/562996c1-24d7-4584-81ce-5c4379eb2d62)

PR #11659, opened 2024-09-24
feat: semantic viewer new/edit/view modes

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/11423

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #11703, opened 2024-09-26
feat: semantic viewer slug

IrakliJani:
Missing feature from: #11659 

### Description:

adds slug support for urls

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/v2/SemanticLayerController.ts
+++ b/packages/backend/src/controllers/v2/SemanticLayerController.ts
@@ -218,42 +218,47 @@ export class SemanticLayerController extends BaseController {
      */
     @Middlewares([allowApiKeyAuthentication, isAuthenticated])
     @SuccessResponse('200', 'Success')
-    @Get('/saved/{uuid}')
+    @Get('/saved')
     @OperationId('getSavedSemanticViewerChart')
     async getSavedSemanticViewerChart(
-        @Path() uuid: string,
-        @Path() projectUuid: string,
         @Request() req: express.Request,
+        @Path() projectUuid: string,
+        @Query() slug?: string,
+        @Query() uuid?: string,
     ): Promise<ApiSemanticViewerChartGet> {
         this.setStatus(200);
         return {
             status: 'ok',
             results: await this.services
                 .getSavedSemanticViewerChartService()
-                .getSemanticViewerChart(req.user!, projectUuid, uuid),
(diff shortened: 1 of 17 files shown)


packages/backend/src/controllers/v2/SemanticLayerController.ts
```diff
@@ -218,42 +218,47 @@ export class SemanticLayerController extends BaseController {
...
-    @Get('/saved/{uuid}')
+    @Get('/saved')
     @OperationId('getSavedSemanticViewerChart')
     async getSavedSemanticViewerChart(
-        @Path() uuid: string,
-        @Path() projectUuid: string,
         @Request() req: express.Request,
+        @Path() projectUuid: string,
+        @Query() slug?: string,
+        @Query() uuid?: string,
```
almeidabbm, line 227: Curious about the reason to support both `slug` and `uuid` via query params instead of just `slug` via url param 🤔 is it because the tiles are working with `uuid` and therefore we need to support both?

packages/common/src/types/errors.ts
```diff
@@ -123,6 +123,7 @@ export class UnexpectedServerError extends LightdashError {
         message = 'Something went wrong.',
         data: { [key: string]: any } = {},
     ) {
+        console.log(message, data);
```
almeidabbm, line 126: Just the log 🫣

almeidabbm (approved):
LGTM 🚀

packages/backend/src/controllers/v2/SemanticLayerController.ts
IrakliJani: correct. its mix of both in all over the code. same with sql runner

packages/common/src/types/errors.ts
IrakliJani: oh good catch. altho, a very necessary log 🤣

packages/backend/src/controllers/v2/SemanticLayerController.ts
IrakliJani: do you have a better suggestion? I was not 100% onboard with using query params. but also was lazy enough to split two routes into 4

almeidabbm: I think that it looks cleaner this way, also it feels like it's extra work just to do the same thing, we can probably "standardize" the way we do this between `semantic-viewer` and `sql-runner` a bit later 👌

Issue #11367, opened 2024-09-03
Add missing functions for sql runner charts in `spaces` list view 


rephus:
We added move to other space and delete in the context menu, but we are still missing some functions because they require extra components/endpoints: 

![Screenshot from 2024-09-03 12-13-04](https://user-images.githubusercontent.com/1983672/363980356-c859535b-551d-4e78-bcfb-d6a77e110f79.png)

- [ ]  rename : needs new modal component
- [ ] add to dashboard: needs modifying hooks on add to dashboard component
- [ ] pin : needs new endpoint
- [ ] promote: needs new endpoint

![Screenshot from 2024-09-03 14-13-19](https://user-images.githubusercontent.com/1983672/363980370-6545bef2-a877-46dd-9dc6-e25b74907f34.png)

_Originally posted by @rephus in https://github.com/lightdash/lightdash/issues/11362#issuecomment-2326389507_

jpetey75:
No one has upvoted or commented on this in over six months. If it's still something you want, let us know and we'll reopen.

PR #11704, opened 2024-09-27
fix: get correct uuid column for resource actions

joaoviana:
Relates to: #11367

### Description:

Fixes the `uuid` column reference so users can add charts to dashboards from the resource actions (which get `uuid` and not the `slug`)

Before

https://github.com/user-attachments/assets/80e28361-432b-46e8-a0fd-82812cef2ae1

After

https://github.com/user-attachments/assets/b7939b04-5701-409b-a8a5-91b89f677e64

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -997,7 +997,7 @@ export class SpaceModel {
                     slug: string;
                 }[]
             >([
-                `${chartTable}.${uuidColumnName}`,
+                `${chartTable}.${uuidColumnName} as uuid`,
                 `${chartTable}.name`,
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/SpaceModel.ts
```diff
@@ -997,7 +997,7 @@ export class SpaceModel {
                     slug: string;
                 }[]
             >([
-                `${chartTable}.${uuidColumnName}`,
+                `${chartTable}.${uuidColumnName} as uuid`,
```
joaoviana, line 1000: this will fix for semantic viewer charts 🆙

almeidabbm (approved):
Works great!

PR #11731, opened 2024-09-30
feat: write back create PR

rephus:
Closes: 

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
+++ b/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
@@ -6,6 +6,7 @@ import {
     DimensionType,
     findAndUpdateModelNodes,
     ForbiddenError,
+    friendlyName,
     GitIntegrationConfiguration,
     isUserWithOrg,
     lightdashDbtYamlSchema,
@@ -14,11 +15,14 @@ import {
     SavedChart,
     SessionUser,
     UnexpectedServerError,
+    VizColumn,
 } from '@lightdash/common';
 import Ajv from 'ajv';
 import * as yaml from 'js-yaml';
 import {
     createBranch,
+    createFile,
+    createPullRequest,
     getFileContent,
     getLastCommit,
     getOrRefreshToken,
@@ -39,6 +43,14 @@ type GitIntegrationServiceArguments = {
     githubAppInstallationsModel: GithubAppInstallationsModel;
 };
 
+type GithubProps = {
+    owner: string;
+    repo: string;
+    branch: string;
+    token: string;
+    path: string;
+    quoteChar: `"` | `'`;
+};
 // TODO move this to common and refactor cli
 type YamlColumnMeta = {
     dimension?: {
@@ -300,7 +312,8 @@ Affected charts:
             );
         const [owner, repo] = project.dbtConnection.repository.split('/');
         const { branch } = project.dbtConnection;
-        return { owner, repo, branch };
+        const path = project.dbtConnection.project_sub_path?.replace(/^\//, '');
+        return { owner, repo, branch, path };
     }
 
     async getOrUpdateToken(organizationUuid: string) {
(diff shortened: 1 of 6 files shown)


packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+    private static async createYmlFile({
+        githubProps,
+        name,
+        columns,
+    }: {
+        githubProps: GithubProps;
+        name: string;
+        columns: VizColumn[];
+    }): Promise<any> {
+        const fileName = `models/${friendlyName(name)}.yml`;
```
joaoviana, line 520: maybe you want to use `snakeCase` instead?

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+        const content = yaml.dump(
+            {
+                version: 2,
+                models: [
+                    {
+                        name: friendlyName(name),
```
joaoviana, line 526: same here

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+
+    private static async createYmlFile({
+        githubProps,
+        name,
+        columns,
+    }: {
+        githubProps: GithubProps;
+        name: string;
+        columns: VizColumn[];
+    }): Promise<any> {
```
joaoviana, line 519: here we can remove this explicit type definition and apply the return type of `createFile` as referenced above ⬆️

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+    private static async createSqlFile({
+        githubProps,
+        name,
+        sql,
+    }: {
+        githubProps: GithubProps;
+        name: string;
+        sql: string;
+    }): Promise<any> {
+        const fileName = `models/${friendlyName(name)}.sql`;
```
joaoviana, line 492: I think `snakeCase` is intended 🤔

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+        await GitIntegrationService.createSqlFile({
+            githubProps,
+            name,
+            sql,
+        });
+        await GitIntegrationService.createYmlFile({
+            githubProps,
+            name,
+            columns,
+        });
```
joaoviana, line 590: probably for later, but do we want to verify if both are created and do some error handling or is that for later?

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+        const pullRequest = await createPullRequest({
+            owner,
+            repo,
+            title: `Write back ${name} SQL and YML model`,
```
joaoviana, line 594:
```suggestion
            title: `Write back `${name}` SQL and YML model`,
```
if we add the backticks like this I think it will code-format it so it looks nicer, wdyt? 
Maybe it needs a unicode or wtv

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+            body: `Created by Lightdash, this PR adds a new SQL file to the dbt model
```
joaoviana, line 595: Perhaps we can add some extra copy here on the files added:
```
Created by Lightdash, this PR adds a new SQL file and generates a Lightdash .yml file to the dbt model.
```

I didn't word this right, but something like this? Mentioning that there are 2 files being generated basically.

joaoviana (commented):
Looks great! Some Qs/suggestions. I'll also need help setting this up locally 👀

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +479,131 @@ Affected charts:
...
+
+    private static async createSqlFile({
+        githubProps,
+        name,
+        sql,
+    }: {
+        githubProps: GithubProps;
+        name: string;
+        sql: string;
+    }): Promise<any> {
```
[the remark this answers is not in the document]
rephus: I'll remove the promise, and let the compiler detect it automatically, but let me know if you still prefer to be more explicit

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
rephus: I'd say we should handle error cases later. 
So far I know this can throw errors if the file already exists.

rephus:
Updated

image

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -466,4 +481,131 @@ Affected charts:
...
+        });
+        await GitIntegrationService.createYmlFile({
+            githubProps,
+            name,
+            columns,
+        });
+        const pullRequest = await createPullRequest({
+            owner,
+            repo,
+            title: `Write back \`${name}\` SQL and YML model`,
```
joaoviana, line 596:
```suggestion
            title: `Write back \`${snakeCase(name)}\` SQL and YML model`,
```

can be done later/it's not that important

joaoviana (approved):

PR #11737, opened 2024-09-30
feat: improve get slack channels performance for large workspace

Shifan-Gu:
Closes: 

### Description:

Issue 1: For large slack workspace, slack client `getChannels` method does not work very well with reduce and sorting the channels, for given example when mocking the workspace with `160k` channels 

```
        const allChannels: SlackChannel[] = [];

        for(let i = 0; i < 160000; i+=1) {
            allChannels.push({ id: `C${i}`, name: `Channel ${(Math.random() + 1).toString(36).substring(7)}` });
        }

        Logger.debug('start sorting channels ');

        const sortedChannels =  allChannels
        .reduce<SlackChannel[]>(
            (acc, { id, name }) =>
                id && name ? [...acc, { id, name: `#${name}` }] : acc,
            [],
        )
        .map(channel => ({ id: channel.id, name: `#${channel.name}` }))
        .sort((a, b) => a.name.localeCompare(b.name));

        Logger.debug('sorting channels completed');
```
the reduce and sorting is taking about 2 mins
Screenshot 2024-09-30 at 21 39 46

and during this period of time, the server main process is completed blocked, result in the situation where all the k8s incoming health check call is failing (gateway timeout) and eventually trigger the pod restart and service downtime.

Issue 2: 

When you passing a super large channels data array (e.g. 200k) to the `MultiSelect`, the component will crash with following error.
Screenshot 2024-09-30 at 21 34 44

change list:

1. replaced reduce with filter & map.
2. added a `search` query parameter to allow responsive search on cached channels.
3. setup a threshold for slack channels responsive search, if channels number below `100000` then let `MultiSelect` handle the search, otherwise use the api.
4. debounce the search in 0.5s.
5. make slack channels cache time configurable, to be able to extend the cache time and reduce the slack api calls that fetches all the slack channels & users.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/SlackSettingsPanel/index.tsx
+++ b/packages/frontend/src/components/UserSettings/SlackSettingsPanel/index.tsx
@@ -58,7 +58,7 @@ const SlackSettingsPanel: FC = () => {
 
     const [search, setSearch] = useState('');
 
-    const debounceSetSearch = debounce((val) => setSearch(val), 500);
+    const debounceSetSearch = debounce((val) => setSearch(val), 1500);
 
     const { data: slackChannels, isInitialLoading: isLoadingSlackChannels } =
         useSlackChannels(search, {


packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -383,6 +385,10 @@ const SchedulerForm: FC<Props> = ({
 
     const [showFormatting, setShowFormatting] = useState(false);
 
+    const [search, setSearch] = useState('');
+
+    const debounceSetSearch = debounce((val) => setSearch(val), 500);
```
rephus, line 390: This is missing `debounce` import

rephus:
Why am I getting channels that are not related with my search ? 

image

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -422,6 +428,9 @@ const SchedulerForm: FC<Props> = ({
             .concat(privateChannels);
     }, [slackChannelsQuery?.data, privateChannels]);
 
+    let responsiveChannelsSearchEnabled =
+        slackChannels.length >= MAX_SLACK_CHANNELS || search.length > 0; // enable responvive channels search if there are more than MAX_SLACK_CHANNELS defined channels
```
rephus, line 432:
```suggestion
        slackChannels.length >= MAX_SLACK_CHANNELS || search.length > 0; // enable responsive channel search if there are more than MAX_SLACK_CHANNELS defined channels
```

Typo

packages/backend/src/controllers/slackController.ts
```diff
@@ -39,6 +40,7 @@ export class SlackController extends BaseController {
     @OperationId('getSlackChannels')
     async get(
         @Request() req: express.Request,
+        @Query() search?: string,
```
rephus, line 43: You need to `yarn generate-api` to start using this parameter, otherwise tsoa will ignore this argument on the query (I was getting undefined all the time)

rephus (changes requested):
This is a nice improvement, but we still need a few changes, see comments for more details 
- generate-api
- remove limit=10

packages/backend/src/controllers/slackController.ts
rephus: image

packages/backend/src/clients/Slack/SlackClient.ts
```diff
@@ -22,7 +22,8 @@ type SlackClientArguments = {
     lightdashConfig: LightdashConfig;
 };
 
-const CACHE_TIME = 1000 * 60 * 10; // 10 minutes
+const DEFAULT_CACHE_TIME = 1000 * 60 * 10; // 10 minutes
+const MAX_CHANNELS_LIMIT = 100000;
```
rephus, line 26: Considering we can now use search, I'd suggest to reduce this to 1000 , what do you think ?

rephus:
> Issue 1: For large slack workspace, slack client getChannels method does not work very well with reduce and sorting the channels, for given example when mocking the workspace with 160k channels

You are not addressing  this issue on this PR, right ? we still make the same API call to slack to get all the channels in a loop (becuase of the cursor parameter) 

Having a large workspace is still going to take a long time before it is cached.

rephus:
Also make sure you run github prehooks before committing, that would have captured the `debounce` issue.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -983,6 +992,7 @@ const SchedulerForm: FC<Props> = ({
                                                         data={slackChannels}
                                                         searchable
                                                         creatable
+                                                        limit={10}
```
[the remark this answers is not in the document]
Shifan-Gu: Hi, i put this because caz i remember it was not very smooth when rendering the dropdown list without this limit set when deal with the large data set, yea but let me confirm that, if its still the case i will expand the limit to a bigger number.

packages/backend/src/clients/Slack/SlackClient.ts
Shifan-Gu: Hi, sure we can do that, from my experience 100k is around the edge of causing that maximum call stack size exceed error, but i want make sure before search is fully tested and ready the users with small workspace is not impacted, for small workspace it would just using the `MultiSelect` build-in search to handle this.

Shifan-Gu:
> > Issue 1: For large slack workspace, slack client getChannels method does not work very well with reduce and sorting the channels, for given example when mocking the workspace with 160k channels
> 
> You are not addressing this issue on this PR, right ? we still make the same API call to slack to get all the channels in a loop (becuase of the cursor parameter)
> 
> Having a large workspace is still going to take a long time before it is cached.

yea, i replace with reduce with map and filter which is making sorting able to complete in seconds, but yes fetch channels is still very slow and unfortunately this `conversation.list` endpoint doesn't support filter entries, so remediate action i can do is cache it longer, so i extracted it as part of configuration, in our deployment we are caching it for a 24 hours right now.

but still we can have a separate ticket to explore a better caching and storage solution.

Shifan-Gu:
> Also make sure you run github prehooks before committing, that would have captured the `debounce` issue.

sorry, that's my bad, we have another internal repo of this copy which is used for internal review, i think this import is gone when i'm resolving the conflicts when merging.

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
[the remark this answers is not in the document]
Shifan-Gu: yea, i tested with 99999 channels without `limit` the input box is not responding smoothly.

in this clip i was keep clicking the input box and this is what it looks like.

https://github.com/user-attachments/assets/8b3a3cf0-90d8-4135-bf5c-ac30832c1bba

rephus (approved):
Tested, it works as expected .Thanks for these changes.

rephus:
We have 2 failing e2e tests, but I don't think they are related 

  │ ✖  downloadCsv.cy.ts                        01:14        4        3        1        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✖  explore.cy.ts                            01:30       10        9        1        -        - │

PR #11746, opened 2024-10-01
fix: branch name write back and installationId

rephus:
Closes: 

### Description:

Using installationId instead of user token will use the bot to create commits / prs instead of the user 

Updated branch name

Screenshot from 2024-10-01 09-54-53

### Reviewer actions


packages/backend/src/clients/github/Github.ts
```diff
@@ -87,7 +87,7 @@ export const getLastCommit = async ({
     owner: string;
     repo: string;
     branch: string;
-    token: string;
+    token: string; // TODO use installationId instead, to act as a bot
```
joaoviana, line 90: do we need to make this change here or irrelevant to this PR?

joaoviana (approved):

packages/backend/src/clients/github/Github.ts
rephus: Maybe later, this method is used in other places (custom metrics), so I don't want to refactor it now

Issue #11747, opened 2024-10-01
I can see to which repo my sql will be written back to


joaoviana:
### Description

- Replace badge content with a link to the repo 
- Add the Github icon as well to make it clear
- Opens link on a new tab

<img width="431" alt="image" src="https://github.com/user-attachments/assets/9ae968b5-53db-4ad9-aa84-55026a6868d4">

### What problem does this solve?

Helps user know what repo it will affect

PR #11763, opened 2024-10-02
feat: write back preview

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/11747

### Description:

image

### Reviewer actions


packages/frontend/src/features/sqlRunner/hooks/useGithubDbtWritePreview.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export const useGithubDbtWritePreview = () => {
+    const { showToastError } = useToaster();
+
+    return useMutation<
+        ApiGithubDbtWritePreview['results'],
+        ApiError,
+        CreatePrParams
+    >(createPullRequest, {
+        mutationKey: ['sqlRunner', 'createPullRequest'],
```
joaoviana, line 41:
```suggestion
        mutationKey: ['sqlRunner', 'githubDbtWritePreview'],
```

packages/frontend/src/features/sqlRunner/hooks/useGithubDbtWritePreview.tsx
```diff
@@ -0,0 +1,50 @@
...
+const createPullRequest = async ({
+    projectUuid,
+    name,
+    sql,
+    columns,
+}: CreatePrParams) =>
+    lightdashApi<ApiGithubDbtWritePreview['results']>({
+        url: `/projects/${projectUuid}/sqlRunner/preview`,
+        method: 'POST',
+        body: JSON.stringify({ name, sql, columns }),
```
joaoviana, line 26: do we need `sql` and `columns`?

packages/backend/src/controllers/sqlRunnerController.ts
```diff
@@ -424,8 +452,8 @@ export class SqlRunnerController extends BaseController {
     ])
     @SuccessResponse('200', 'Success')
     @Post('pull-request')
-    @OperationId('createCustomExplore')
-    async createPr(
+    @OperationId('writeBackCreatePr')
```
joaoviana, line 455: 🙏

packages/frontend/src/features/sqlRunner/hooks/useGithubDbtWritePreview.tsx
```diff
@@ -0,0 +1,50 @@
...
+        ApiError,
+        CreatePrParams
+    >(createPullRequest, {
+        mutationKey: ['sqlRunner', 'createPullRequest'],
+
+        onError: () => {
+            showToastError({
+                title: 'Failed to get github preview',
+                subtitle: 'Please check your Github settings.',
+            });
```
joaoviana, line 47: I think we can pass `name` to this mutation key so we get some caching benefits 🤔 

```tsx

export const useGithubDbtWritePreview = (name: string) => {
    const { showToastError } = useToaster();

    return useMutation<
        ApiGithubDbtWritePreview['results'],
        ApiError,
        CreatePrParams
    >(createPullRequest, {
        mutationKey: ['sqlRunner', 'githubDbtWritePreview', name],
        onError: () => {
            showToastError({
                title: 'Failed to get github preview',
                subtitle: 'Please check your Github settings.',
            });
        },
    });
};
```

packages/frontend/src/features/sqlRunner/components/WriteBackToDbtModal.tsx
```diff
@@ -54,22 +58,45 @@ export const WriteBackToDbtModal: FC<Props> = ({ opened, onClose }) => {
...
+                void getWritePreview({
+                    projectUuid,
+                    name: form.values.name,
+                    sql,
+                    columns,
+                });
+            }
+        }, 300);
+        return () => clearTimeout(debounceTimer);
+    }, [opened, form.values.name, projectUuid, sql, columns, getWritePreview]);
```
joaoviana, line 99: This is really, cool, but we can do two things here: 
- have only 1 `useEffect`
- use `mantine/hooks` for the debounced value
- keep the mutation's previous data so the UI doesn't jump on load

```tsx
// NOTE: This keeps a mutation's previous data (similar to useQuery's `keepPreviousData`
 const [writePreviewData, setWritePreviewData] =
        useState<ApiGithubDbtWritePreview['results']>();

    const form = useForm<FormValues>({
        initialValues: {
            name: '',
        },
        validate: zodResolver(validationSchema),
    });
    const { mutateAsync: getWritePreview, isLoading: isPreviewLoading } =
        useGithubDbtWritePreview(form.values.name);
    const [debouncedName] = useDebouncedValue(form.values.name, 300);
```

then the unified `useEffect` will be this

```tsx
  useEffect(() => {
        if (!opened || !projectUuid || !sql || !columns) return;

        const loadPreview = async () => {
            const data = await getWritePreview({
                projectUuid,
                name: debouncedName || 'custom view',
                sql,
                columns,
            });
            setWritePreviewData(data);
        };

        void loadPreview();
    }, [opened, projectUuid, debouncedName, sql, columns, getWritePreview]);

```

then, after the `Badge` we could do something like

```tsx
</Badge>
                            {isPreviewLoading && <Loader size="xs" />}
```

joaoviana (changes requested):
Suggested a few ways to remove a `useEffect` and use some mantine hooks + caching strategy, lmk what you think!

packages/frontend/src/features/sqlRunner/hooks/useGithubDbtWritePreview.tsx
rephus: no, but since we already have `CreatePrParams` I decided to use the same type for the body, we might use it in the future. Happy to remove it if you don't like it

packages/frontend/src/features/sqlRunner/hooks/useGithubDbtWritePreview.tsx
rephus: good idea

packages/frontend/src/features/sqlRunner/components/WriteBackToDbtModal.tsx
rephus: I tried with 1 useEffect only, but then it was making multiple requests because it was triggered on `isLoading`but I'll try this approach. Thanks for the suggestion

packages/frontend/src/features/sqlRunner/hooks/useGithubDbtWritePreview.tsx
rephus: good idea, but we can't use name there... it should be on the async method..

joaoviana (approved):

Issue #11749, opened 2024-10-01
When I refresh dbt, my custom virtual views shouldn't be cleared


joaoviana:
### Description

Currently, when users create a custom virtual view, from the SQL runner, it gets added to the `/tables` view:
<img width="430" alt="image" src="https://github.com/user-attachments/assets/994c2373-bd9d-447c-9066-b404cc88db77">
<img width="301" alt="image" src="https://github.com/user-attachments/assets/d05bd079-a79e-4225-bc13-dab4fa7ca9ad">

ℹ️  However, when users click on `refresh dbt` (or compile their project) the explores will be re-populated, and virtual views/explores won't be included in that because they don't _persist_

### Implementation suggestion

We could flag these compiled explores with an optional property: `isVirtual` or include a `type` property `virtual|default`
Then, whenever the cached_explores and cached_explore tables get cleared, we ensure that the virtual ones are kept in memory before removal and added back

### Steps to Reproduce the Bug or Issue

See ℹ️ above ^ 

### Cloud or self-hosting

None

PR #11772, opened 2024-10-02
fix: persist virtual views when refreshing dbt

joaoviana:
Closes: [#11749](https://github.com/lightdash/lightdash/issues/11749)

### Description:

This PR allows the persistence of `virtual views`/`custom explores` when the user refreshes dbt.
It also moved these below all _default_ explores

Screenshot 2024-10-02 at 15 32 48

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -948,7 +948,7 @@ export class ProjectModel {
 
                 // We don't support multiple explores with the same name at the moment
                 const uniqueExplores = uniqWith(
-                    explores,
+                    [...explores, ...customExplores.map((e) => e.explore)],
                     (a, b) => a.name === b.name,
                 );
 
@@ -984,10 +984,7 @@ export class ProjectModel {
                 )
                     .insert({
                         project_uuid: projectUuid,
-                        explores: JSON.stringify([
-                            ...uniqueExplores,
-                            ...customExplores.map((e) => e.explore),
-                        ]),
+                        explores: JSON.stringify(uniqueExplores),
                     })
                     .onConflict('project_uuid')
                     .merge()
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -939,7 +939,7 @@ export class ProjectModel {
                 )
                     .select('explore')
                     .where('project_uuid', projectUuid)
-                    .whereRaw("explore->>'type' = ?", [ExploreType.CUSTOM]);
+                    .whereRaw("explore->>'type' = ?", [ExploreType.VIRTUAL]);
 
                 // delete previous individually cached explores
                 await this.database(CachedExploreTableName)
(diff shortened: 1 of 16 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -2042,23 +2050,13 @@ export class ProjectModel {
...
-        const toAddToCached: Explore = {
-            name,
-            tags: [],
-            label: friendlyName(name),
-            tables: translatedToExplore.tables,
-            baseTable: name,
-            groupLabel: 'Custom explores',
-            joinedTables: [],
-            targetDatabase: SupportedDbtAdapter.POSTGRES,
-        };
```
joaoviana, line 2061: this was unnecessary - also wrong 😅

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -933,14 +933,22 @@ export class ProjectModel {
             'ProjectModel.saveExploresToCache',
             {},
             async () => {
+                // Get custom explores/virtual views before deleting them
+                const customExplores = await this.database(
+                    CachedExploreTableName,
+                )
+                    .select('explore')
+                    .where('project_uuid', projectUuid)
+                    .whereRaw("explore->>'type' = ?", [ExploreType.CUSTOM]);
```
rephus, line 942: we'll need now a new ticket to allow users `delete` custom explores in the UI

rephus (approved):
This works great, 

There is only one weird thing if you have some filtered tables on settings. We should probably create a separate ticket for this. 

When creating the custom explore, you can see it on the explore, but it doesn't show up on the list

I think we should `add it to the list of selected explores` on custom explore creation so they appear automatically. Forcing custom explores to always appear on the list might not be ideal, specially since we can't remove them 

Screenshot from 2024-10-03 08-37-29
Screenshot from 2024-10-03 08-36-35
[Screencast from 03-10-24 08:36:56.webm](https://github.com/user-attachments/assets/dfc46e9a-3e0f-45c4-b678-16fbc52987ac)

joaoviana:
No filtering anymore for custom explores/virtual views

Screenshot 2024-10-03 at 10 58 33

**before adding:**
Screenshot 2024-10-03 at 10 58 39

**after adding:**
Screenshot 2024-10-03 at 10 59 02

rephus:
is the e2e error on catalog.cy.ts related ?

packages/common/src/types/explore.ts
```diff
@@ -38,6 +38,11 @@ export type CompiledTable = TableBase & {
     uncompiledSqlWhere?: string;
 };
 
+export enum ExploreType {
+    CUSTOM = 'custom',
```
rephus, line 42: should we use `virtual` here ?

joaoviana: Agree, I should change! ty

joaoviana:
> is the e2e error on catalog.cy.ts related ?
> 
> Is this because you manually changed the selected tables on render, and that broke some e2e tests ? If that's the case, we can merge

This was exactly why @rephus ! 
Reset the db to have a clean slate and all are passing 😄

Issue #11770, opened 2024-10-02
Allow upserting credentials


almeidabbm:
When you change the environment id and not the service token, the “save” button is still greyed out, so I can’t upsert changes

PR #11775, opened 2024-10-02
fix: make update accept partial

almeidabbm:
Closes: #11770 

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -2216,11 +2216,9 @@ export class ProjectModel {
 
         const [updatedProject] = await this.database(ProjectTableName)
             .update({
-                semantic_layer_connection: updatedSemanticLayerConnection
-                    ? this.encryptionUtil.encrypt(
-                          JSON.stringify(updatedSemanticLayerConnection),
-                      )
-                    : null,
+                semantic_layer_connection: this.encryptionUtil.encrypt(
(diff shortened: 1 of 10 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -2082,15 +2083,59 @@ export class ProjectModel {
...
+            ? merge(currentSemanticLayerConnection, connectionUpdate)
+            : connectionUpdate;
+
+        if (
+            !ProjectModel.isSemanticLayerConnectionValid(
+                updatedSemanticLayerConnection,
+            )
+        ) {
+            throw new ParameterError('Invalid semantic layer connection');
+        }
```
almeidabbm, line 2132: It should only merge the current connection with the partial update when:
- There is no current connection since it will take all properties from the partial
- When type is the same, this is so that we don't end up with rogue properties being saved to the wrong type

Before saving we always check if the `updatedSemanticLayerConnection` is valid (has all properties needed for the semantic layer type), this is to ensure that we never create a connection with missing properties

almeidabbm:
Sql runner new tests are failing because db name changed after closing/reopening the PR

IrakliJani:
needs rebasing @almeidabbm

packages/backend/src/models/ProjectModel/ProjectModel.ts
IrakliJani: makes sense. nice work!

IrakliJani (approved):
🚢

PR #11777, opened 2024-10-02
fix: Improve domain check in emails 

SamKingDev:
Closes: https://github.com/lightdash/lightdash-internal-work/issues/2099

### Description:
This fixes a security vulnerability where malicious users could sign up with an account which could be classed as another organisations email, therefore joining that organization if the "let users with this domain" setting is enabled. As emails can contain more than one emails an email such as `"@organization.com@"@gmail.com` would with the old code be classed as organization.com

This bug hasn't been tested as I don't have a spare email server but in theory looking at this code there is no reason why this wouldn't work.

https://stackoverflow.com/questions/12355858/how-many-symbol-can-be-in-an-email-address

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/email.ts
+++ b/packages/common/src/utils/email.ts
@@ -4,7 +4,7 @@ export const getEmailDomain = (email: string): string => {
     }
 
     const domains = email.split('@');
-    if (domains.length !== 2) {
+    if (domains.length !== 2 || !domains[1]) {
         throw new Error(`Invalid email: ${email}`);
     }
     return domains[1].toLowerCase();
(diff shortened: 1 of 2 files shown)


packages/common/src/utils/email.ts
```diff
@@ -3,11 +3,11 @@ export const getEmailDomain = (email: string): string => {
         throw new Error(`Invalid email, contains whitespace: ${email}`);
     }
 
-    const domain = email.split('@')[1];
-    if (!domain) {
+    const domains = email.split('@');
+    if (domains.length !== 2) {
```
rephus, line 7: The new code doesn't check if there is a domain, so this new code would accept "user@" as a valid email. 
Perhaps we add `|| !domains[1]`  here to ensure this case is covered, like in the previous version.

rephus (changes requested):
Thanks for the code ! you successfuly fixed the issue. We only need to cover one condition that was previously covered in the old code

rephus (approved):

Issue #11765, opened 2024-10-02
Disable `write-back to dbt` option if project does not have github integration or no github project connection


joaoviana:
### Description

See steps below ⬇️ 

ref: 
<img width="264" alt="image" src="https://github.com/user-attachments/assets/bd441f54-9333-4a30-81f4-3436c36ac01b">

### Steps to Reproduce the Bug or Issue

1. Go to a project with no github connection
2. Go to SQL runner creation page
3. Run query
4. Try to write back to dbt (this shouldn't be possible at all)

### Cloud or self-hosting

None

PR #11778, opened 2024-10-03
fix: disable github button if not enabled

rephus:
Closes: https://github.com/lightdash/lightdash/issues/11765

### Description:

Screenshot from 2024-10-03 09-55-06

For testing, you can force this error by blocking the /list endpoint in dev tools.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/UserSettings/GithubSettingsPanel/index.tsx
+++ b/packages/frontend/src/components/UserSettings/GithubSettingsPanel/index.tsx
@@ -27,7 +27,7 @@ const getGithubRepositories = async () =>
         body: undefined,
     });
 
-const useGitHubRepositories = () =>
+export const useGitHubRepositories = () =>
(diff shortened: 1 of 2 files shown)


rephus:
Should we `hide` the menu.item if feature flag is not enable,d and show error if `github not installed` (aka error on list errors)

joaoviana:
> Should we `hide` the menu.item if feature flag is not enable,d and show error if `github not installed` (aka error on list errors)
> 
> I don't want peope saying: "This is disabled for me, but I don't even see github integrations on settings "

Agree! Maybe we can do a conditional rendering of that Menu.Item, good point 💪

joaoviana (approved):
Looks great! But I agree with your point above, I think we should ship that in this PR @rephus

rephus:
image
Done

joaoviana (approved):

Issue #10355, opened 2024-06-10
It should be easy to see to which tab my dashboard filter is being applied


joaoviana:
![image](https://github.com/user-attachments/assets/b644e196-2d8b-4335-ba72-ca051a6393be)

**If I have turned the filter toggle off for a particular tab (tab 1 above) in the dashboard:**
- [ ]  When I am viewing Tab 1 in the dashboard, the filter should appear to be disabled
- [ ]  When I am viewing Tab s or any other tab where the filter is on, I should be able to use the filter normally
- [ ]  When I hover over the disabled filter, there should be a tooltip telling me why it is disabled. `This filter only applies to 'affected tab name', 'affected tab name'. `

Example:
![image](https://github.com/user-attachments/assets/1bc309d5-ce00-49af-8c70-714f77f4db39)

PR #11803, opened 2024-10-03
feat: deactivate unused filters for tab

agha4to:
Closes: https://github.com/lightdash/lightdash/issues/10355

### Description:
For each dashboard filter, if the filter does not apply to any Tiles on the current Tab, show the filter as inactive and include an info icon with tooltip about where the filter applies.

Screenshot below shows:
1. Normal filter tag/button
2. Dashed border w/ gray text and icon/tooltip: filter deactivated for currently active tab
3. Normal temporary filter tag/button
4. Blue dashed border w/ gray text and icon/tooltip: temporary filter deactivated for currently active tab
image

Please note the new function `isTileFilterable` to check whether tiles are filterable or not (MARKDOWN and LOOM). There may be other places, unrelated to this PR, where bugs are caused because this is not done.

Also please note that any Required filter will stop all tabs from loading even if the Required filter does not apply to all tabs. This is an existing issue, and I feel it is probably outside the scope of this PR.

### Reviewer actions


packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
```diff
@@ -37,6 +45,45 @@ const ActiveFilters: FC<ActiveFiltersProps> = ({
         (c) => c.updateDimensionDashboardFilter,
     );
 
+    const makeTabGetter = useCallback(
```
ZeRego, line 48: > if the filter does not apply to any Tiles on the current Tab, show the filter as inactive

The initial ticket doesn't mention this behaviour. Was this brought up by someone?
Even though this could be a clever way to identify which filters are being used in the tab, as a user, I was confused. 
The filter rules were disabled and said they were only applied to the first tab but when I went to edit them, it was marked as enabled.

Screenshot 2024-10-21 at 14 37 45

ZeRego:
> Also please note that any Required filter will stop all tabs from loading even if the Required filter does not apply to all tabs. This is an existing issue, and I feel it is probably outside the scope of this PR.

👍

packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
agha4to: The issue mentions "disabled" rather than "inactive" but I think I see the main issue you're mentioning. This PR currently checks if each filter is being applied to any tile on the current tab. I think you're suggesting that it should check if each filter is enabled for a tab, even if there are no tiles on the current tab. That makes sense. I'll take a look.

ZeRego:
> Dashed border w/ gray text and icon/tooltip: filter deactivated for currently active tab
> Blue dashed border w/ gray text and icon/tooltip: temporary filter deactivated for currently active tab

@PriPatel Any comments on the styling for the disabled filters?

PriPatel:
@ZeRego I dont think the dashes are necessary as it adds an extra layer of 'stuff' to understand and the pattern generally indicates an 'empty state' . The greyed out appearance should be enough

agha4to:
Fyi thinking was that Gray can mean unable to click, but since these are clickable we want to differentiate as "inactive".
Trying to differentiate from invalid filters for example:
image

agha4to:
I removed the text color change for inactive filters, so now it is the same as active filters.
image

PriPatel:
@agha4to The border colour of the dashed line should be the same as an undashed line. Basically we want the _only difference_ between an active and unactive tab is the dashed border. No colour changes etc. 
Now the border is black and it stands out more than the active tabs, its having the opposite effect.

PriPatel:
@agha4to Sorry for the back and forth but as this deviated from the original design in figma (rightly so!), its been a bit hard to get right. The problem with the latest screenshot shared is that the 'unactive tab' now stands out more than the active tab. I've just mocked up a solution here.
Screenshot 2024-10-29 at 13 27 46


- [ ] Stroke colour should not change 
- [ ] Add dashes
- [ ] Add background opacity 50% to the unactive pill (this makes it look less solid, but the unaffected text still makes it look clickable)
- [ ] The info icon feels distracting - can the info text just show on hover over the filter? It can override the current tooltip.

PriPatel:
@agha4to - great feedback and suggestion, lets go with that!

packages/frontend/src/components/DashboardFilter/ActiveFilters/index.tsx
```diff
@@ -121,6 +130,74 @@ const ActiveFilters: FC<ActiveFiltersProps> = ({
...
+                    );
+                });
+            };
+        },
+        [dashboardTiles, sortedTabUuids],
+    );
+
+    const getTabsUsingFilter = useCallback(
+        (filterId: string) => makeTabGetter(dashboardFilters)(filterId),
+        [makeTabGetter, dashboardFilters],
```
conv22, line 192: minor: I think wrapping getTabsUsing filter and other tab getters within the useCallback is redundant here, it's used within the same react component and gets called on every render

ZeRego (approved):

Issue #11817, opened 2024-10-04
Posthog feature flag check in the BE should not fail silently


IrakliJani:
### Description

Posthog feature flag check in the BE should not fail silently and return false

### Steps to Reproduce the Bug or Issue

-

### Cloud or self-hosting

None

PR #11818, opened 2024-10-04
fix: posthog silent fails

IrakliJani:
Closes: #11817 

### Reviewer actions


rephus (approved):

packages/backend/src/postHog.ts
```diff
@@ -47,12 +77,16 @@ export async function isFeatureFlagEnabled(
...
                 : {},
-        ),
-        new Promise<boolean>((resolve) => {
-            setTimeout(() => {
-                resolve(false);
-            }, FLAG_CHECK_TIMEOUT);
-        }),
+        );
+
+        clearTimeout(timeout);
```
almeidabbm, line 82: This will make it so that the `timeoutPromise` never rejects or resolves

IrakliJani: it does not matter because its in a `Promise.race`

almeidabbm: true, what I meant is that it'll hang in the background, but probably will be garbage collected anyway and it is not important 🤔

almeidabbm (approved):

.env.development
```diff
@@ -36,7 +36,7 @@ DBT_PROJECT_DIR=.
 
 SCHEDULER_ENABLED=true
 
-GROUPS_ENABLED=true
+# GROUPS_ENABLED=true
```
almeidabbm, line 39: Not sure if intended, but it's for dev so not important

packages/backend/src/postHog.ts
IrakliJani: 🤞

rephus (approved):

PR #11832, opened 2024-10-07
fix: check if file exists in github

rephus:
Closes: #11831

### Description:

Screenshot from 2024-10-07 14-01-30

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
+++ b/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
@@ -23,6 +23,7 @@ import Ajv from 'ajv';
 import * as yaml from 'js-yaml';
 import { nanoid } from 'nanoid';
 import {
+    checkFileDoesNotExist,
     createBranch,
     createFile,
     createPullRequest,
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -537,6 +538,7 @@ Affected charts:
             name,
             'sql',
         );
+        await checkFileDoesNotExist({ ...githubProps, path: fileName });
```
joaoviana, line 541: in `GitIntegrationService` > `writeBackPreview` this is the path:

```
 path: `${path}/models/lightdash`,
```

Could you remind me if the `/lightdash` is intended? 
I was basically trying to cause this duplicate file issue, by creating an `events` write back in my `jaffle_shop` repo. I was expecting it to error because there's already an `events.yml` in the `/models` folder.

rephus: context: https://lightdash.slack.com/archives/C07P77EG3KR/p1727951317786839?thread_ts=1727950310.393599&cid=C07P77EG3KR

We can add subfolders to models, to easily filter them, and reduce conflicts. 

if you add an events from the sql runner, it will not error the first time, because it writes in /lightdash, but it will if you do it a second time.

joaoviana (approved):
got the error when there was already a file
image

Issue #11834, opened 2024-10-07
Disable write back for non-github project connections


joaoviana:
### Description

When I have a non-GitHub connection project, I shouldn't be able to write back to dbt because there's no way for us to know what repo to make changes to. 

Consider hiding this option in the SQL runner, or disable with an appropriate tooltip

Reference in `HeaderCreate.tsx`

<img width="227" alt="image" src="https://github.com/user-attachments/assets/79912922-2ca8-4d7c-8a56-cc36356b311a">

### Steps to Reproduce the Bug or Issue

See above ^

### Cloud or self-hosting

cloud

PR #11836, opened 2024-10-07
fix: disable write back if project is not github

rephus:
Closes: #11834
### Description:

Screenshot from 2024-10-07 17-05-30
Screenshot from 2024-10-07 17-05-42
Screenshot from 2024-10-07 17-06-52

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/Header/HeaderCreate.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/Header/HeaderCreate.tsx
@@ -1,4 +1,4 @@
-import { FeatureFlags } from '@lightdash/common';
+import { DbtProjectType, FeatureFlags } from '@lightdash/common';
 import {
     Button,
     Group,
@@ -21,6 +21,7 @@ import { useGitHubRepositories } from '../../../../components/UserSettings/Githu
 import { EditableText } from '../../../../components/VisualizationConfigs/common/EditableText';
 import useHealth from '../../../../hooks/health/useHealth';
 import { useFeatureFlagEnabled } from '../../../../hooks/useFeatureFlagEnabled';
+import { useProject } from '../../../../hooks/useProject';
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import {
     DEFAULT_NAME,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/features/sqlRunner/components/Header/HeaderCreate.tsx
```diff
@@ -50,7 +54,8 @@ export const HeaderCreate: FC = () => {
 
     const isGithubIntegrationEnabled =
         useFeatureFlagEnabled(FeatureFlags.CustomSQLEnabled) &&
-        health?.data?.hasGithub;
+        health?.data?.hasGithub &&
+        project?.dbtConnection.type === DbtProjectType.GITHUB;
```
joaoviana, line 58: I believe this boolean should be kept as is, because if someone has a github connection type but github app/bot hasn't been set up yet, then we show the tooltip `Please enable Github integration to write back to dbt in Settings > Integrations > Github'`

Instead, I would create a separate var like `canWriteBackToDbt = project?.dbtConnection.type === DbtProjectType.GITHUB`

and then if that's not true, we don't render the option

What do you think?

joaoviana (changes requested):
one q

packages/frontend/src/features/sqlRunner/components/Header/HeaderCreate.tsx
rephus: This `isGithubIntegrationEnabled` flag is what controls the `render` of the `menu item`

So. if doesn't have github ENV variables (hasGithub), or the feature flag is Flase,  or the project is not  on GITHUB, it will not show the menu item, which is likely a thing you can't change easily. 

To check if you have a github integration (aka bot) , we use ` const { isError: githubIsNotInstalled } = useGitHubRepositories();`  If this is false (and `isGithubIntegrationEnabled` is true) , we show the `Please enable Github integration to write back to dbt in Settings > Integrations > Github'`

joaoviana: ahhhhh `hasGithub` checks an env that isn't affected by the installation - sorry for adding confusion to this. 

Makes sense the explanation, approving!

joaoviana: And tested all scenarios

joaoviana (approved):

Issue #11795, opened 2024-10-03
When I edit a virtual view, if existing fields are changed, then I get a warning.


TuringLovesDeathMetal:
### Description

If I change the existing fields in a virtual view, then we trigger a warning when the user tries to save. 

- [ ] we need to have a copy of the schema of the virtual view when the user opens `edit`. This schema should have information about the field names and their types that are generated from the query. 
- [ ] If a field name no longer exists (it is deleted, or the name is changed), then we trigger the warning message on save
- [ ] If the type of a field name changes, then we trigger the warning message on save (e.g. `user_id` changes from `string` to `integer`)

<img width="836" alt="image" src="https://github.com/user-attachments/assets/d672a186-ec98-40b6-a0c1-2369480243ac">

- the warning message should contain information about which fields are affected (e.g. if I delete or rename `katies_cool_column`, then I should see a message saying `katies_cool_column` as one of the fields affected)
- if more than 3 fields are affected, we should only list the first 3 fields, then have the option to `show more` which loads the list of fields in a scroll (?) (something that makes it so the modal isn't massive! I'll leave the implementation up to you! 👀 )

PR #11837, opened 2024-10-07
feat: alert users on changes to virtual view

joaoviana:
Closes: #11795

### Description:

Alerts the user when there are changes to the schema when editing their virtual view
This PR also moves the modal to the features/sqlRunner to be easier to manage

Demo:

https://github.com/user-attachments/assets/4db061af-3aab-4695-a796-91e562654d78

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/Header/HeaderVirtualView.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/Header/HeaderVirtualView.tsx
@@ -1,20 +1,184 @@
 import { type VizColumn } from '@lightdash/common';
-import { Button, Group, LoadingOverlay, Text } from '@mantine/core';
-import { IconTableAlias } from '@tabler/icons-react';
-import { type FC } from 'react';
+import {
+    Button,
+    Collapse,
+    Group,
+    List,
+    LoadingOverlay,
+    Modal,
+    Stack,
+    Text,
+    type ModalProps,
+} from '@mantine/core';
+import {
+    IconAlertCircle,
+    IconAlertHexagon,
+    IconPlus,
+    IconTableAlias,
+    IconTrash,
+} from '@tabler/icons-react';
+import { memo, useEffect, useState, type FC } from 'react';
 import { useHistory } from 'react-router-dom';
 import MantineIcon from '../../../../components/common/MantineIcon';
 import { useSqlQueryRun } from '../../hooks/useSqlQueryRun';
 import { useUpdateVirtualView } from '../../hooks/useVirtualView';
-import { useAppSelector } from '../../store/hooks';
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
+import { setSqlRunnerResults } from '../../store/sqlRunnerSlice';
+import { compareColumns, type ColumnDiff } from '../../utils/compareColumns';
+
+const DiffListItem: FC<{ diff: ColumnDiff }> = memo(({ diff }) => {
+    return (
+        <List.Item
+            fz="xs"
+            icon={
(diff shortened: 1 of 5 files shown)


packages/frontend/src/pages/SqlRunnerNew.tsx
```diff
@@ -164,10 +164,12 @@ const SqlRunnerNew = ({
                                 label={'Open sidebar'}
                                 position="right"
                             >
-                                <ActionIcon size="sm">
+                                <ActionIcon
+                                    size="sm"
```
joaoviana, line 168: small fix here - should be on ActionIcon instead of the mantine icon

packages/frontend/src/features/sqlRunner/components/Header/HeaderVirtualView.tsx
```diff
@@ -25,32 +189,59 @@ export const HeaderVirtualView: FC<{
     const { mutateAsync: updateVirtualView, isLoading: isUpdatingVirtualView } =
         useUpdateVirtualView(projectUuid);
 
-    const handleUpdateVirtualView = async () => {
+    useEffect(() => {
         if (!columns) {
             return;
         }
+        if (!initialColumns) {
+            setInitialColumns(columns);
```
joaoviana, line 197: useful for the final diffing

rephus:
If there is a SQL error, pressing save doesn't do anything

image
We should show a tooltip with the error

rephus:
should we tell which columns were added ?

image

joaoviana:
> should we tell which columns were added ?
> 
> image

this wasn't part of the initial requirement (additions) so I just added as a one-liner to avoid info overload

joaoviana:
> If there is a SQL error, pressing save doesn't do anything
> 
> image We should show a tooltip with the error

hmmm checking!

rephus:
Changing the table in this sql does not throw any popup, I guess that is expected... as long as the columns are the same 

> SELECT order_id FROM "jaffle_db_pg_13_pr_11837"."jaffle"."payments"  limit 5

rephus:
not sure if we want to force to `run query` if there are unsaved changes before `allowing save` :thinking:

joaoviana:
> not sure if we want to force to `run query` if there are unsaved changes before `allowing save` 🤔

You mean when the query is exactly the same or when there is a different query but it hasn't been run? 

Regarding the latter point, here is the internal thread https://lightdash.slack.com/archives/C07P77EG3KR/p1728055744424379?thread_ts=1728048409.083809&cid=C07P77EG3KR

>Allow users to save changes without needing to run the query, especially if they know the SQL is correct and don’t need to validate it.

So we the run the query if there are unrun changes (diffing could be better tho) so we can get the correct columns + their types

rephus:
image
nice

packages/frontend/src/features/sqlRunner/utils/compareColumns.ts
```diff
@@ -0,0 +1,53 @@
+import { type VizColumn } from '@lightdash/common';
+
+export type ColumnDiff = {
+    type: 'deleted' | 'typeChanged' | 'added';
+    reference: VizColumn['reference'];
+    oldType?: VizColumn['type'];
+    newType?: VizColumn['type'];
+};
+export const compareColumns = (
```
rephus, line 9: :+1:

rephus (approved):
It works great

joaoviana:
> It works great
> 
> If we can fix the tooltip on errors, then great, but not a blocker IMO

done!
Screenshot 2024-10-08 at 10 35 32

Issue #11830, opened 2024-10-07
I am warned that my current explore work will be discarded if I decide to edit a virtual view


joaoviana:
### Description

Acceptance Criteria 
- [ ] When I'm querying from a table, create a chart, etc, I'm warned that my changes will be lost if I decide to edit this virtual view
- [ ] I can close the modal and none of my changes will be lost
- [ ] If I make a change by saving, then my page will refresh and the `unsaved chart version` will be empty

PR #11838, opened 2024-10-07
feat: warn users that changes will be discarded before editing

joaoviana:
Closes: #11830 

### Description:

Refactored modal to listen to modal step changes (now there's 3)
Center modals that provide warnings
Clear unsaved chart version if the user wants to discard them and continue editing the virtual view

Demo:

https://github.com/user-attachments/assets/5a0f94fb-5ebd-4630-a9fa-c575a1dcad5c

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/ExplorePanel/index.tsx
+++ b/packages/frontend/src/components/Explorer/ExplorePanel/index.tsx
@@ -12,7 +12,6 @@ import { IconDots, IconPencil } from '@tabler/icons-react';
 import { memo, useMemo, useState, useTransition, type FC } from 'react';
 import { EditVirtualViewModal } from '../../../features/sqlRunner/components/EditVirtualViewModal';
 import { useExplore } from '../../../hooks/useExplore';
-import useSearchParams from '../../../hooks/useSearchParams';
 import { useExplorerContext } from '../../../providers/ExplorerProvider';
 import MantineIcon from '../../common/MantineIcon';
 import PageBreadcrumbs from '../../common/PageBreadcrumbs';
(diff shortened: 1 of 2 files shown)


joaoviana:
> any idea why I can't save virtual explores on render ? adblock is off
> 
> image

unrelated to this PR, I'm going to check why the postog feature is returning `false` and not `true` @rephus

rephus (approved):
This works, but would it be possible to : 

- do not show popup if there are `no` changes
- do not unmount the component, it makes sense to keep in the background and  avoid making new requests

If they are not possible or difficult, I'm happy to merge this as it is

[Screencast from 08-10-24 10:34:57.webm](https://github.com/user-attachments/assets/abf95290-08d1-4fa4-8da4-67ec487f5990)

joaoviana:
> * do not show popup if there are `no` changes

Ah, this got me blocked. 
I tried to select from the sql runner slice, but no luck bc it was selecting outside the store provider. 
Then, I tried using this new react hook, like this
```tsx
export function useHasUnrunChanges() {
  return useSyncExternalStore(
    store.subscribe,
    () => selectHasUnrunChanges(store.getState()),
    () => false // Server-side initial state
  );
}
```
Which listens to the redux store (even though it's not provided) - it was getting the value correctly, but every time the store changed, it would trigger a re-render - this obviously led to a performance hit + the sql on the editor was being cleared every time, bc the modal was instantiating the sql runner on every render. 

Since this is not immediately possible, I will skip this just for now...

> * do not unmount the component, it makes sense to keep in the background and  avoid making new requests

Done with `ConditionalVisibility` - great suggestion ty

Issue #11850, opened 2024-10-08
Dashboard chart Y Axis Label gets cut


almeidabbm:
![image](https://github.com/user-attachments/assets/c8b7f860-918e-47be-8bfe-d51eae251655)

PR #11852, opened 2024-10-08
fix: add 25px padding to grid instead of 5%

almeidabbm:
Closes: #11850

### Description:

Before:
image

After:
image

Note: this makes the padding of fixed instead of %, which means that when a chart tile is bigger the padding will be the same, previously it got bigger

### Reviewer actions


IrakliJani (approved):
love the comment: // small padding

PR #11858, opened 2024-10-09
fix: hide edit/delete virtual view if no permission to

joaoviana:
Closes: N/A

### Description:

Manage permissions by adding a new ability 
`>= developers can manage virtual views`

## Admin - can see 3-dot menu

Screenshot 2024-10-09 at 10 22 43

## Editor - cannot see 3-dot menu

Screenshot 2024-10-09 at 10 28 30

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -4382,6 +4382,17 @@ export class ProjectService extends BaseService {
         projectUuid: string,
         payload: CreateVirtualViewPayload,
     ) {
+        const { organizationUuid } =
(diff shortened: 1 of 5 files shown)


rephus (approved):
Changes are  good, is there any reason why we need a separate subject ? is this because we don't want interactive_viewer to delete virtualv iews ?

joaoviana:
> Changes are good, is there any reason why we need a separate subject ? is this because we don't want interactive_viewer to delete virtualv iews ?

basically yea! 
I also wondered this (a need for a separate subject)
thinking more about it, I wonder if we could use the `manage` `project` ability. Essentially, the users who can update explores/compile a project are the ones that can manage a project, am I correct @rephus ?

Issue #11876, opened 2024-10-10
403 error when running `lightdash preview` with org member and project admin


almeidabbm:

LauriLanni:
I strongly think project previews is a functionality that should be given to the project developers, not just admins. It is a core part of the Lightdash development flow

almeidabbm:
> I strongly think project previews is a functionality that should be given to the project developers, not just admins. It is a core part of the Lightdash development flow

Hello @LauriLanni 👋 

Thank you for sharing your feedback, we've just created a ticket to do this: https://github.com/lightdash/lightdash/issues/11881

PR #11877, opened 2024-10-10
fix: project admin can create preview projects

almeidabbm:
Closes: #11876 

### Description:

From the docs page, a project admin should be able to create project previews: https://docs.lightdash.com/references/roles/#organization-roles

**Before**
image

**After**
image

**Steps to reproduce:**
1. Create a user that has `member` role at the org level and `admin` at the project level
2. Run `lightdash preview` with that user

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -424,7 +424,10 @@ export class ProjectService extends BaseService {
             ) ||
             user.ability.cannot(
                 'create',
-                subject('Project', { organizationUuid }),
+                subject('Project', {
+                    organizationUuid,
+                    type: data.type,
+                }),
             )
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -424,7 +424,10 @@ export class ProjectService extends BaseService {
             ) ||
             user.ability.cannot(
                 'create',
-                subject('Project', { organizationUuid }),
+                subject('Project', {
```
rephus, line 427: I can't replicate this issue locally as an admin. :thinking: I don't think this type is needed, since it was only checking if you can `crate project in the org` , and an admin can `manage projects`

rephus (approved):
Tested locally, nice fix.

PR #11883, opened 2024-10-10
chore: migrate legacy dashboard endpoints from project router to controller

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/11882

### Description:

- Migrates dashboard GET, POST and PATCH to project controller

### Reviewer actions


rephus:
We need to fix the build and run all e2e tests before we can merge this

rephus:
I've seen this `    1) Should download a CSV from dashboard` failing in another PR, so I don't think this PR is causing the issue.

packages/backend/src/controllers/projectController.ts
```diff
@@ -505,4 +513,109 @@ export class ProjectController extends BaseController {
             results: undefined,
         };
     }
+
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Get('{projectUuid}/dashboards')
```
rephus, line 519: Get methods should be safe to translate. 
Post methods, with Record<> can be dangerous and we need to manually test its backwards compatibility

packages/backend/src/controllers/projectController.ts
```diff
@@ -505,4 +513,109 @@ export class ProjectController extends BaseController {
...
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('201', 'Created')
+    @Post('{projectUuid}/dashboards')
+    @OperationId('createDashboard')
+    async createDashboard(
+        @Path() projectUuid: string,
+        @Body() body: DuplicateDashboardParams | CreateDashboard,
```
rephus, line 554: this CreateDAshboard type is quite complex, we need to be careful. I didn't see any recrod in the nested type, though

packages/backend/src/controllers/projectController.ts
```diff
@@ -505,4 +513,109 @@ export class ProjectController extends BaseController {
...
+        @Request() req: express.Request,
+        @Query() duplicateFrom?: string,
+    ): Promise<ApiCreateDashboardResponse> {
+        const dashboardService = this.services.getDashboardService();
+        this.setStatus(201);
+
+        let results: ApiCreateDashboardResponse['results'];
+
+        if (duplicateFrom) {
```
rephus, line 563: This is ok because you copied it from the existing browser, but we should move this logic to the service IMO. We can do it on a separate PR or make a ticket.

packages/backend/src/controllers/projectController.ts
```diff
@@ -505,4 +513,109 @@ export class ProjectController extends BaseController {
...
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('200', 'Updated')
+    @Patch('{projectUuid}/dashboards')
+    @OperationId('updateDashboards')
+    async updateDashboards(
+        @Path() projectUuid: string,
+        @Body() body: UpdateMultipleDashboards[],
```
rephus, line 606: Update multiple dashboards extends Dashboard, which contains DashboardFilters :fire: This type has caused issues in the past because of the Record (eg: /calculate-total simply uses any instead)

rephus (commented):
The code itself looks good, but I'd like to test these endpoints manually first.

packages/backend/src/controllers/projectController.ts
rephus: IT seems this endpoint is only used when moving dashboards to a space 
image

rephus (approved):
I manually tested these endpoints on render and tehey seem to be working fine, even with dashboard filters :+1: it should be safe to merge.

PR #11886, opened 2024-10-10
fix: check for pgversion before querying for matviews

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/11885

### Description:

Previously we were querying `pg_catalog.pg_matviews` without checking first for the PG version.
`pg_matviews` is only available from PostgreSQL >= 12: https://www.postgresql.org/docs/current/view-pg-matviews.html

Logs added for debug purposes:

**PG 11 (doesn't support matviews)**

image

**PG 15 (supports matviews)**
image (1)

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
@@ -314,6 +314,16 @@ export class PostgresClient<
         if (databases.size <= 0 || schemas.size <= 0 || tables.size <= 0) {
             return {};
         }
+
+        const { rows: pgVersionRows } = await this.runQuery('SELECT version()');
+        const pgVersionString = pgVersionRows[0]?.version ?? '';
+        const versionRegex = /PostgreSQL (\d+)\./;
+        const versionMatch = pgVersionString.match(versionRegex);
+        const supportsMatviews =
+            versionMatch && versionMatch[1]
+                ? parseInt(versionMatch[1], 10) >= 12
+                : false;
+
         const query = `
             SELECT table_catalog,
                    table_schema,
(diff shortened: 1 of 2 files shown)


examples/full-jaffle-shop-demo/dbt/models/metrics.yml
```diff
@@ -1,35 +0,0 @@
-version: 2
```
rephus, line 1: could you rollback these  unrelated changes ? not sure if this is still needed somewhere...

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -314,6 +314,13 @@ export class PostgresClient<
         if (databases.size <= 0 || schemas.size <= 0 || tables.size <= 0) {
             return {};
         }
+
+        const { rows: supportsMatViewsRows } = await this.runQuery(
+            `SELECT current_setting('server_version_num')::integer >= 120000 as supports_matviews`,
+        );
+        const supportsMatviews =
+            supportsMatViewsRows[0]?.supports_matviews ?? '';
```
rephus, line 322:
```suggestion
            supportsMatViewsRows[0]?.supports_matviews ?? false;
```

Should this be a boolean ? rather than an empty string

rephus (approved):

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -314,6 +314,13 @@ export class PostgresClient<
         if (databases.size <= 0 || schemas.size <= 0 || tables.size <= 0) {
             return {};
         }
+
+        const { rows: supportsMatViewsRows } = await this.runQuery(
+            `SELECT current_setting('server_version_num')::integer >= 120000 as supports_matviews`,
```
joaoviana, line 319: so this is >= v12?

almeidabbm: yep yep, `pg_matviews` is only available on and after 12: https://www.postgresql.org/docs/current/view-pg-matviews.html

Issue #11710, opened 2024-09-27
Explore is creating very long URLs when selecting fields


rephus:
### Description

There is a bug , where selecting dimensions and columns generates a really long URL 

This is particularly bad if someone for whatever reason refresh their browser, because they get this error

![image](https://github.com/user-attachments/assets/fed61415-1388-42bb-85ad-57fe29c3ace6)

### Steps to Reproduce the Bug or Issue

1. go to explore (eg orders)
2. select some dimensions and metrics
3. Copy the generated URL :fire: 

![image](https://github.com/user-attachments/assets/2eef2af1-1d4d-46d0-926a-aa24e9829f34)

> The most repetitive part is inside the chartConfig.config.eChartsConfig.series array. Each item in this array represents a series in the chart and follows this structure:

> The repeating patterns are:
a. There are 15 series objects, each representing a combination of:
3 metric fields: "orders_average_order_size", "orders_completed_order_count", "orders_date_of_first_order"
5 status values: "placed", "shipped", "return_pending", "completed", "returned"

### Cloud or self-hosting

None

rephus:
The issue is mainly with pivotValues, I guess it creates one entry per serie 

![image](https://github.com/user-attachments/assets/c0afc68b-f8d7-49b7-a15c-a6664a1eebb7)

PR #11889, opened 2024-10-11
fix: reduce url on explore

rephus:
Closes: https://github.com/lightdash/lightdash/issues/11710

### Description:

What was happening: adding pivots was causing a really long metricquery, and we were adding all the series config into the URL 

What's new: We check the URL, if it is bigger than `3000` chars (just on queyr parameters) (limit seems to be ~4000 , including the host+path) then we remove the `series from the metric query config`

Expected side effect: CUstom configuration on series (eg: color, value labels, mixed chart types...) will be lost after refreshing. 
Screenshot from 2024-10-11 10-20-06
image

Alternative solutions: If we don't want to lose information (custom serie config), we will have to store it somewhere, perhaps in local storage, or save a draft in the backend . I don't think this needs to be done, the side effect is very small. 

How to test: 
- go to an explore
- select dimensions and columns
- add a big pivot value (eg: orders day) 
- if you refresh now the page or copy the URL into a new tab, it should load fine (except the custom serie config) 

Screenshot from 2024-10-11 10-20-06

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useExplorerRoute.ts
+++ b/packages/frontend/src/hooks/useExplorerRoute.ts
@@ -2,6 +2,7 @@ import {
     ChartType,
     CustomDimensionType,
     DateGranularity,
+    isCartesianChartConfig,
     type CreateSavedChartVersion,
     type CustomBinDimension,
     type CustomDimension,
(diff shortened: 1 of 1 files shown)


almeidabbm (approved):
👨‍🍳 🤌

PR #11911, opened 2024-10-14
fix: databricks sql runner catalog

rephus:
Closes: https://github.com/lightdash/lightdash/issues/11574

### Description:

Before:

Screenshot from 2024-10-14 09-15-11

After:

Screenshot from 2024-10-14 09-18-59
Screenshot from 2024-10-14 09-28-24
Screenshot from 2024-10-14 09-30-35
Screenshot from 2024-10-14 09-30-47

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
@@ -396,6 +396,21 @@ export class DatabricksWarehouseClient extends WarehouseBaseClient<CreateDatabri
         }
     }
 
+    async getAllTables() {
+        const query = `
+            SELECT table_catalog, table_schema, table_name
+            FROM information_schema.tables
+            WHERE table_type = 'MANAGED' 
+            ORDER BY 1,2,3
+        `;
+        const { rows } = await this.runQuery(query, {}, undefined, undefined);
+        return rows.map((row) => ({
+            database: row.table_catalog,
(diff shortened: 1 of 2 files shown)


packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
```diff
@@ -396,6 +396,21 @@ export class DatabricksWarehouseClient extends WarehouseBaseClient<CreateDatabri
         }
     }
 
+    async getAllTables() {
+        const query = `
+            SELECT table_catalog, table_schema, table_name
+            FROM information_schema.tables
+            WHERE table_type = 'MANAGED'
```
joaoviana, line 403: I see, so we can include more tables depending on types here: https://docs.databricks.com/en/sql/language-manual/information-schema/tables.html#table-types


Will review this with the credentials you've set up 👍

rephus: yeah, is there any other you think we should include ? like `materialized_view` I'd like to avoid `view` because that's what the system table uses

joaoviana: Perhaps let's leave it for now? I like what you have already

joaoviana (approved):
can see the tables of databricks catalog

Screenshot 2024-10-14 at 10 09 55

Issue #11913, opened 2024-10-14
Update explore share URL every time metric query changes


almeidabbm:
Currently we only update the explore URL when the query runs successfully. This means that if I am updating an explore and encounter an error, the link I share doesn't include my changes.

PR #11916, opened 2024-10-14
fix: update explore when metric query changes instead of on success

almeidabbm:
Closes: [#11913](https://github.com/lightdash/lightdash/issues/11913)

### Description:

**Steps to test**
1. Click on "Query from tables"
2. Select a table and some fields (don't click "Run Query")
3. Copy the link at the top, when opened it should have the same state

### Reviewer actions


joaoviana (changes requested):
I would expect my url params to clear when click back to tables

image

image

almeidabbm:
> I would expect my url params to clear when click back to tables
> 
> image image

Now it should only update when there is table name, this way the url params are cleared 💪

joaoviana (approved):

Issue #11793, opened 2024-10-03
(OIDC) There is issue when i try to cancel registration


nesh-ua:
### Description

I have one issue with login via OIDC.
I allow registration only using invite
When people login through my OIDC provider.
They get this message on page, and I think it is ok.
But when I click button cancel registration: 
![page](https://github.com/user-attachments/assets/a073858b-0194-40ab-a541-bf6f5843a8b6)

there is 500 error
![500 error](https://github.com/user-attachments/assets/dcf80881-8fc5-4ae1-9adc-79f4c1e1e259)

### Steps to Reproduce the Bug or Issue

1. Login through OIDC provider (maybe another SSO provider have the same issue)
2. Try to cancel registration

### version

v0.1282.0

### Cloud or self-hosting

self-hosting

PR #11923, opened 2024-10-14
fix: delete user on cancel registration

rephus:
Closes:  #11793

### Description:

Before:
Screenshot from 2024-10-14 16-38-24

After:
Screenshot from 2024-10-14 16-40-43

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -294,7 +294,9 @@ export class UserService extends BaseService {
         const [orgForUser] = await this.userModel.getOrganizationsForUser(
             userUuidToDelete,
         );
-        if (orgForUser.organizationUuid) {
+        // The user might not have an org yet
+        // This is expected on the "Cancel registration" flow on single org instances.
+        if (orgForUser?.organizationUuid) {
             // We assume only one org per user
 
             if (


packages/backend/src/services/UserService.ts
```diff
@@ -294,7 +294,9 @@ export class UserService extends BaseService {
         const [orgForUser] = await this.userModel.getOrganizationsForUser(
             userUuidToDelete,
         );
-        if (orgForUser.organizationUuid) {
+        // The user might not have an org yet
+        // This is expected on the "Cancel registration" flow on single org instances.
+        if (orgForUser?.organizationUuid) {
```
joaoviana, line 299: Yes, we do the same on the frontend on `join organization` error

joaoviana (approved):

packages/backend/src/services/UserService.ts
joaoviana: One question though, maybe the boolean needs to be switched?

rephus: mm, I don't think so ? 

this IF only makes sense if the user has an org. If not, we simply delete the user

joaoviana (approved):

PR #11950, opened 2024-10-16
chore: debug posthog client

almeidabbm:
### Reviewer actions


packages/backend/src/postHog.ts
```diff
@@ -11,9 +11,27 @@ const FLAG_CHECK_TIMEOUT = process.env.POSTHOG_CHECK_TIMEOUT
...
+              if (!response.ok) {
+                  try {
+                      Logger.error('PostHog Error', response.json());
+                  } catch (e) {
+                      Logger.error('PostHog Error', response.body);
+                  }
+              }
+
+              return response;
+          },
```
almeidabbm, line 26: We should discuss the possibility of overwriting the request method used by posthog client
Currently they're using axios this would replace it with fetch, it probably won't make a difference, but we can log the errors ourselves

image

IrakliJani (approved):

Issue #11318, opened 2024-08-30
Introduce new permission to create preview projects


ZeRego:
### Description

Add a new permission `ProjectPreview` that is available to:
- org: admin, developer, editor, int viewer
- project: admin, developer

The permission check should include the `upstreamProjectUuid` value. 

```
// check project creation
user.ability.cannot(
  'create', 
  subject('Project', { organizationUuid }),
); 

// check preview project creation
user.ability.cannot(
  'create',
  subject('PreviewProject', { organizationUuid, projectUuid: upstreamProjectUuid }),
) 
```

When a preview is created from another project, all project admins and developers should be copied. 

### What problem does this solve?

At the moment we need to assign a user to have access to all projects (org int viewer or higher) or be a project admin to be able to create a preview. This is not suitable for most organisations where they want to limit users to 1 project but don't want them to delete projects by mistake. Project developers should be able to create previews since it is a crucial part of their day-to-day.

PR #11955, opened 2024-10-16
feat: interactive viewers can create previews

rephus:
Closes:  [11318](https://github.com/lightdash/lightdash/issues/11318)

### Description:

Interactive viewers can't create projects but they can create preview projects
Screenshot 2024-10-16 at 16 39 03

Screenshot 2024-10-16 at 16 41 01

copied content
Screenshot 2024-10-16 at 16 54 09

Copy same users / groups in the previevw project 

Screenshot from 2024-10-17 09-14-25

Because of that :point_up: we now can copy also private spaces, withouth letting the user access those

Screenshot from 2024-10-17 09-19-56

Screenshot from 2024-10-17 09-20-50

Can't create a default project as org editor/project developer: 

>         console.log('createWithoutCompile', organizationUuid, data.upstreamProjectUuid, data.type)
Screenshot from 2024-10-17 15-54-23
Screenshot from 2024-10-17 15-54-28

### Reviewer actions

test-frontend test-backend

### DIFF
--- a/docs/docs/references/dbt-write-back.mdx
+++ b/docs/docs/references/dbt-write-back.mdx
@@ -4,9 +4,9 @@ import TablesConfigurationWriteBack from './assets/tables-configuration-write-ba
 
 # dbt Write-Back
 
-dbt Write-Back allows you to create models in dbt from queries you've built in the [SQL Runner](/references/sql-runner). You can then explore these models as Tables in Lightdash. Unlike [Virtual Views](/references/virtual-views), these models will be saved to and managed in your dbt project. 
+dbt Write-Back allows you to create models in dbt from queries you've built in the [SQL Runner](/references/sql-runner). You can then explore these models as Tables in Lightdash. Unlike [Virtual Views](/references/virtual-views), these models will be saved to and managed in your dbt project.
 
-We recommend using dbt Write-Back if you're planning to use a query regularly so it can be governed and managed like other dbt models available in your Lightdash project. 
+We recommend using dbt Write-Back if you're planning to use a query regularly so it can be governed and managed like other dbt models available in your Lightdash project.
 
 :::info
 
@@ -16,7 +16,7 @@ dbt Write-Back is currently only available for dbt projects hosted in GitHub.
 
 ## Getting started with dbt Write-Back
 
-Writing back to dbt will create a new branch and start a pull request in your dbt project to create a new .sql and .yml file based off of your SQL Runner query. 
+Writing back to dbt will create a new branch and start a pull request in your dbt project to create a new .sql and .yml file based off of your SQL Runner query.
 
 To get started, select the `Write back to dbt` option from the `save` drop-down.
 
@@ -27,7 +27,7 @@ To get started, select the `Write back to dbt` option from the `save` drop-down.
   style={{ display: 'block', margin: '0 auto 20px auto' }}
 />
 
-You'll be asked to enter a name that will be used as the model name and file names in your dbt project. Clicking `open pull request` will open a pull request created by Lightdash against your dbt project in GitHub. 
+You'll be asked to enter a name that will be used as the model name and file names in your dbt project. Clicking `open pull request` will open a pull request created by Lightdash against your dbt project in GitHub.
 
 <img
   src={WriteBackPullRequest}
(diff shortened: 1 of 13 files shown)


docs/docs/references/dbt-write-back.mdx
```diff
@@ -4,9 +4,9 @@ import TablesConfigurationWriteBack from './assets/tables-configuration-write-ba
 
 # dbt Write-Back
 
-dbt Write-Back allows you to create models in dbt from queries you've built in the [SQL Runner](/references/sql-runner). You can then explore these models as Tables in Lightdash. Unlike [Virtual Views](/references/virtual-views), these models will be saved to and managed in your dbt project. 
+dbt Write-Back allows you to create models in dbt from queries you've built in the [SQL Runner](/references/sql-runner). You can then explore these models as Tables in Lightdash. Unlike [Virtual Views](/references/virtual-views), these models will be saved to and managed in your dbt project.
```
rephus, line 7: file not changed, but it automatically applied so formatting after updating roles.mdx

owlas (commented):
My issue is with these lines when we're doing permission checks: https://github.com/lightdash/lightdash/pull/11955/files#diff-dba02b88a0bd4a4fdcf7ee49daaaa481189bc9ad76a683c357d0f59e568370a5L420-L434

The permissions system should be `if USER cannot ACTION` - it should be a single check. If we're checking multiple things then that means there's permission logic encoded outside of CASL. 

We want all the permissions in CASL.

This check should just be:
```
if (user.abality.cannot('create', subject('PreviewProject', {organizationUuid, baseProjectUuid}))
```

The actual permission should be, if you're an admin/dev of a project (either because of org or project level permissions, then you can create a preview). 

Let me know if this makes sense. We shouldn't need to add the ability to create projects to interactive viewers and then rely on developers to check that permission **AND** another permission

joaoviana:
**Admin/default user can create projects & previews**
admin-org-project-preview

**Interactive viewer can't create previews if they're an interactive viewer in the project as well**
interactive-viewer-cannnot-create-project

**Interactive viewer can create previews if they're >=developer in the project**
Screenshot 2024-10-17 at 16 23 26

joaoviana:
org editor can't create default projects

Screenshot 2024-10-17 at 16 31 50

packages/common/src/authorization/projectMemberAbility.test.ts
```diff
@@ -1309,4 +1310,93 @@ describe('Project member permissions', () => {
             ).toEqual(true);
         });
     });
+
+    describe('test project preview permissions', () => {
+        it('developers can not create preview or regular projects', () => {
```
joaoviana, line 1315: maybe you mean `viewers`?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3916,6 +3976,10 @@ export class ProjectService extends BaseService {
             },
             async () => {
                 const spaces = await this.spaceModel.find({ projectUuid }); // Get all spaces in the project
+
+                // TODO we do not need to filter allowed space content here,
```
joaoviana, line 3980: Since we don't need, should we remove? Or should this be raised as a separate issue?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3901,6 +3903,64 @@ export class ProjectService extends BaseService {
...
+        upstreamProjectUuid: string,
+        previewProjectUuid: string,
+    ): Promise<void> {
+        this.logger.info(
+            `Copying access from project ${upstreamProjectUuid} to preview project ${previewProjectUuid}`,
+        );
+        await wrapSentryTransaction<void>(
+            'duplicateUserAccess',
+            {
+                previewProjectUuid,
```
joaoviana, line 3921: should we add the `upstreamProjectUuid` here as well?

joaoviana (changes requested):
works as expected with groups + different roles in org, then project

some qs though!

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: I was going to leave it commeted out, but we can always check the code in the git history.

joaoviana (approved):

Issue #11869, opened 2024-10-09
When I hide a field in table chart types, the field should appear to be disabled.


PriPatel:
<img width="362" alt="Screenshot 2024-10-09 at 17 12 53" src="https://github.com/user-attachments/assets/cbcd91ee-1afb-4831-9ed6-e4d2445bae02">

Right now, it still looks active

- [ ] When I hide it, it should be disabled and I shouldn't be able to rename it (it feels misleading)

Ankush263:
Hello @PriPatel , I have made the changes, and now it should look like this (with `not-alowed` cursor)
![Screenshot (159)](https://github.com/user-attachments/assets/76120452-f3b9-4175-8016-c15948062add)
I'll raise a PR.

PR #11962, opened 2024-10-16
fix: style hidden column as read-only

Ankush263:
Closes: #11869

### Description:

- When someone hides a field in the table chart, it should appear to be disabled
- The cursor pointer is set to be `not-allowed`

### Reviewer actions

- [✅] I have manually tested the changes in the preview environment
- [✅] I have reviewed the code
- [✅] I understand that "request changes" will block this PR from merging


joaoviana (approved):
Thank you @Ankush263 for contributing again to Lightdash and making it better 🙌 

I've made a small change on removing an unnecessary rightSection style override and used a theme color for the hidden inputs 

Just running E2E tests now 👍

Issue #11960, opened 2024-10-16
Analytics events for virtual views


TuringLovesDeathMetal:
### Description

- `virtual_view_created` - this event should be triggered when a virtual view is first created. 
  - `virtual_view_id` - unique identifier for the virtual view
  - `name` - name of the virtual view (this is PII, so I'll strip it out in our data modelling layer, but we should track it)
  - `user_id` - user ID that created the virtual view
  - `project_id` - project_id that the virtual view was created in
  - `organization_id` - org ID that the virtual view was created in. 
- `virtual_view_updated` - this event should be triggered when a virtual view is updated.
  - `virtual_view_id` - unique identifier for the virtual view
  - `name` - name of the virtual view (if the name is changed, this should change)
  - `user_id` - user ID that updated the virtual view
  - `project_id` - project_id that the virtual view is in
  - `organization_id` - org ID that the virtual view is in
- `virtual_view_deleted` - this event should be triggered when a virtual view is deleted.
  - `virtual_view_id` - unique identifier for the virtual view
  - `name` - name of the virtual view (if the name is changed, this should change)
  - `user_id` - user ID that deleted the virtual view
  - `project_id` - project_id that the virtual view is in
  - `organization_id` - org ID that the virtual view is in

- `query_executed` --> add a parameter called `virtual_view_id` which is empty if the query was executed from a regular explore or the SQL runner, but is non-empty if the query was executed  from a virtual view.

PR #11969, opened 2024-10-17
chore: add CUD virtual views & query_executed analytics events

joaoviana:
Closes: #11960 

### Description:

> [!NOTE]
> I'm doing the `saved_chart_created` and `saved_chart_updated` on a separate PR since it demands a wider change

**virtual_view_created - this event should be triggered when a virtual view is first created.**

Screenshot 2024-10-17 at 10 25 36

**virtual_view_updated - this event should be triggered when a virtual view is updated.**

Screenshot 2024-10-17 at 10 26 27

**virtual_view_deleted - this event should be triggered when a virtual view is deleted.**

Screenshot 2024-10-17 at 10 27 21

**query_executed --> add a parameter called virtual_view_id which is empty if the query was executed from a regular explore or the SQL runner, but is non-empty if the query was executed from a virtual view.**

Screenshot 2024-10-17 at 10 36 49

### Reviewer actions

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -208,6 +208,7 @@ type MetricQueryExecutionProperties = {
     numCustomSqlDimensions: number;
     dateZoomGranularity: string | null;
     timezone?: string;
+    virtualViewId?: string;
 };
 
 type SqlExecutionProperties = {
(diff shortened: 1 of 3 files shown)


packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -1056,6 +1057,38 @@ export type SemanticLayerView = BaseTrack & {
     };
 };
 
+export type VirtualViewCreatedEvent = BaseTrack & {
+    event: 'virtual_view.created';
```
rephus, line 1061: should we unify all these events and make `name` optional ?

rephus (approved):

packages/backend/src/analytics/LightdashAnalytics.ts
joaoviana: done!

Issue #11716, opened 2024-09-30
time_intervals property is not working correctly on Additional dimensions 


rephus:
### Description

I created an additional dimension of a date , I used time_intervals: 'OFF' and it is creating 2 dimensions, one without time intervals, and another one with the default interval, there should be only 1 . 

Also adding only `['YEAR']` doesn't work as expected

### Steps to Reproduce the Bug or Issue

1. add this to the yml

```
      - name: created
        description: Timestamp (UTC) when customer was created
        meta:
          additional_dimensions:
            created_year:
              type: date
              time_intervals: 'OFF'
              label: "Enrollment Season Year (Date)"
              sql: to_date(created || '-01-01', 'YYYY-MM-DD')
```

2. Refresh dbt
3. Get 2 dimensions

![image](https://github.com/user-attachments/assets/1f86f9f7-00a9-40c7-b6de-e1e888fb466b)

### Cloud or self-hosting

None

PR #11970, opened 2024-10-17
fix: fix additional_dimensions allowing time_intervals

almeidabbm:
Closes: #11716 

### Description:

After this PR `additional_dimensions` now support `time_intervals`!
It follows the same logic of `dimensions` where:
- if `time_intervals` is not specified, it uses the default
- if `time_intervals` is specified, it uses those intervals
- if `time_intervals` are "OFF", the value is raw

```yaml
      - name: event
        description: ""
        meta:
          dimension:
            type: string
            groups: ["events"]
          additional_dimensions:
            timestamp_est:
              type: timestamp
              group_label: "Timestamp test"
              sql: ${TABLE}.timestamp_ltz
              time_intervals: ["YEAR"]
```

### Reviewer actions

### DIFF
--- a/packages/common/src/compiler/translator.ts
+++ b/packages/common/src/compiler/translator.ts
@@ -78,12 +78,15 @@ const convertTimezone = (
 const isInterval = (
     dimensionType: DimensionType,
     { meta }: DbtModelColumn,
+    isAdditionalDimension: boolean | undefined,
 ): boolean =>
     [DimensionType.DATE, DimensionType.TIMESTAMP].includes(dimensionType) &&
     meta.dimension?.time_intervals !== false &&
-    ((meta.dimension?.time_intervals &&
-        meta.dimension.time_intervals !== 'OFF') ||
-        !meta.dimension?.time_intervals);
+    Boolean(
+        meta.dimension?.time_intervals &&
+            meta.dimension.time_intervals !== 'OFF',
+    ) &&
+    !isAdditionalDimension;
 
 const convertDimension = (
     index: number,
@@ -115,13 +118,16 @@ const convertDimension = (
         sql = convertTimezone(sql, 'UTC', 'UTC', targetWarehouse);
     }
     const isIntervalBase =
-        timeInterval === undefined && isInterval(type, column);
+        timeInterval === undefined &&
+        isInterval(type, column, isAdditionalDimension);
 
     let timeIntervalBaseDimensionName: string | undefined;
+
     const groups: string[] = convertToGroups(
         column.meta.dimension?.groups,
         column.meta.dimension?.group_label,
     );
+
     if (timeInterval) {
         timeIntervalBaseDimensionName = name;
         sql = timeFrameConfigs[timeInterval].getSql(
@@ -299,7 +305,7 @@ export const convertTable = (
             let extraDimensions = {};
 
             const processIntervalDimension = (dim: Dimension) => {
-                if (isInterval(dim.type, column)) {
+                if (isInterval(dim.type, column, dim.isAdditionalDimension)) {
                     let intervals: TimeFrames[] = [];
                     if (
                         column.meta.dimension?.time_intervals &&
--- a/packages/common/src/compiler/translator.ts
+++ b/packages/common/src/compiler/translator.ts
@@ -5,6 +5,7 @@ import {
     convertToGroups,
     isV9MetricRef,
     SupportedDbtAdapter,
+    type DbtColumnLightdashDimension,
     type DbtMetric,
     type DbtModelColumn,
     type DbtModelNode,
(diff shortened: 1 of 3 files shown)


packages/common/src/compiler/translator.ts
```diff
@@ -115,13 +115,17 @@ const convertDimension = (
         sql = convertTimezone(sql, 'UTC', 'UTC', targetWarehouse);
     }
     const isIntervalBase =
-        timeInterval === undefined && isInterval(type, column);
+        timeInterval === undefined &&
+        isInterval(type, column) &&
+        !isAdditionalDimension;
```
almeidabbm, line 120: Additional dimensions can never be a interval base (this is what was causing the dimensions to not show in the frontend)

packages/common/src/compiler/translator.ts
```diff
@@ -378,10 +380,14 @@ export const convertTable = (
...
+
                 return {
                     ...acc,
-                    [subDimensionName]: additionalDim,
-                    ...processIntervalDimension(additionalDim),
+                    // When the additional dim is interval AND the base dimension is a interval base then we want to compute all additional dims with the parent intervals otherwise just set the additional dim
+                    ...(isInterval(additionalDim.type, column) &&
+                    dimension.isIntervalBase
+                        ? processIntervalDimension(additionalDim)
+                        : { [subDimensionName]: additionalDim }),
```
almeidabbm, line 390: This logic fixes 2 things:

1. Previously we were always trying to process interval dimensions on the additional dim + adding the additional dim as is, this was causing duplicates the duplicate mentioned in the ticket and the bug where we took into account the `time_intervals` property in the yaml for the additional dimension
2. Makes it so that if the base dimension is not an interval base (e.g. when base dimension is 'string' but somehow we want to add an additional dimension which is type 'date' - see mock data `MODEL_WITH_ADDITIONAL_DIMENSIONS`) then we won't generate intervals and the dimension will be just the raw value.

Overall this makes it so that:
1. if base dimension is date/time and my additional dimension is date/time then it will respect the intervals of the base dimension
2. if base dimension is date/time with intervals set to 'OFF' and my additional dimension is date/time, both will be raw value (no intervals)
3. if base dimension is NOT date/dime and additional dimension is date/time, my additional dimension will not have intervals
4. if base dimension is date/time and additional dimension is NOT date/time, my additional dimension will not have intervals

Point 3 can be changed to add the intervals to the additional dimension, but I believe that would be unexpected since they cannot be controlled via the yaml

packages/common/src/compiler/translator.mock.ts
```diff
@@ -1297,145 +1297,7 @@ export const LIGHTDASH_TABLE_WITH_ADDITIONAL_DIMENSIONS: Omit<
...
-            requiredAttributes: undefined,
-            round: undefined,
-            source: undefined,
-            sql: "DATE_TRUNC('MONTH', ${metadata}-->'created_at')",
-            table: 'myTable',
-            tableLabel: 'My table',
-            timeInterval: TimeFrames.MONTH,
-            timeIntervalBaseDimensionName: 'created_at',
-            type: DimensionType.DATE,
-        },
```
almeidabbm, line 1347: Removed the mock for additional dimension intervals because in this case of the mock the base dimension is `string` and the additional dimension is `timestamp`, we no longer generate intervals additional dimensions in this case.

packages/common/src/compiler/translator.ts
almeidabbm: After the last commit this is no longer true, see description

packages/common/src/compiler/translator.ts
```diff
@@ -296,18 +299,26 @@ export const convertTable = (
...
 
-            let extraDimensions = {};
-
-            const processIntervalDimension = (dim: Dimension) => {
-                if (isInterval(dim.type, column)) {
+            const processIntervalDimension = (
+                dim: Dimension,
+                overrideTimeIntervals: DbtColumnLightdashDimension['time_intervals'],
+            ) => {
+                if (dim.isIntervalBase) {
```
almeidabbm, line 306: This fixes the issue where when a `additional_dimension` had `time_intervals: "OFF"` then it would still show as both `raw` value + default time frames. 

A dimension with `time_intervals: "OFF"` is NOT `isIntervalBase` and therefore no intervals are generated.

joaoviana:
tested with override: 
image
image


and without override: 

image

image

packages/common/src/schemas/json/lightdash-dbt-2.0.json
```diff
@@ -383,6 +383,79 @@
                                                     "maxItems": 2
                                                 }
                                             }
+                                        },
+                                        "additional_dimensions": {
+                                            "type": "object",
```
joaoviana, line 388: 🙌

joaoviana (approved):
🚢 it !

Issue #11859, opened 2024-10-09
Allow float numbers in custom range inputs for custom dimensions


IrakliJani:
### Description

Enable users to input float values when defining custom ranges for custom dimensions.

![CleanShot 2024-10-09 at 15 06 07@2x](https://github.com/user-attachments/assets/a1a89618-5792-4c15-a0d0-431671ad9db2)

PR #11981, opened 2024-10-17
feat: custom dimensions custom range float values

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/11859

### Description:

mantine 7 offers the option to have floats in the number input. but mantine 6 does not. changed to text input instead of number input and I also added a zod schema

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
+++ b/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
@@ -150,7 +150,7 @@ export const CustomBinDimensionModal: FC<{
     }, [setFieldValue, item, isEditing]);
 
     const handleOnSubmit = form.onSubmit((unparsedValues) => {
-        // manitne form does not produce zod parsed values
+        // mantine form does not produce zod parsed values
         // so, number({ coerce: true }) does not work
         // that's why we need to parse the values manually
         const values = formSchema.parse(unparsedValues);
--- a/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
+++ b/packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
@@ -149,7 +149,12 @@ export const CustomBinDimensionModal: FC<{
         }
     }, [setFieldValue, item, isEditing]);
 
-    const handleOnSubmit = form.onSubmit((values) => {
+    const handleOnSubmit = form.onSubmit((unparsedValues) => {
+        // manitne form does not produce zod parsed values
+        // so, number({ coerce: true }) does not work
+        // that's why we need to parse the values manually
+        const values = formSchema.parse(unparsedValues);
+
         if (item) {
             const sanitizedId = sanitizeId(
                 values.customDimensionLabel,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
```diff
@@ -59,29 +60,13 @@ export const CustomBinDimensionModal: FC<{
...
-                customRange: DEFAULT_CUSTOM_RANGE,
-            },
-        },
-        validate: {
-            customDimensionLabel: (label) => {
-                if (!label) return null;
-
-                if (!item) return null;
+    const formSchema = z.object({
+        customDimensionLabel: z.string().refine(
```
IrakliJani, line 64: since I moved the validation to zod, I also checked if it works the same way it used to. it does:
CleanShot 2024-10-17 at 20 35 04@2x

packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
```diff
@@ -94,16 +79,49 @@ export const CustomBinDimensionModal: FC<{
...
+            fixedNumber: z.object({
+                binNumber: z.number().positive(),
+            }),
+            fixedWidth: z.object({
+                binWidth: z.number().positive(),
+            }),
+            customRange: z.array(
+                z.object({
+                    from: z.number({ coerce: true }).or(z.undefined()),
+                    to: z.number({ coerce: true }).or(z.undefined()),
```
IrakliJani, line 102: CleanShot 2024-10-17 at 20 35 56@2x

rephus (commented):
This is a great solution, and it works as expected. 

However, there is only 1 thing I'm worried about, now the backend query is using `string` instead of `number` (actually, a mix) which means, this type is wrong, and in the backend we are still assuming this is a number. I'm surprised tsoa is not rejecting the request. Do you think we should change it ?
 
Screenshot from 2024-10-21 08-51-26

image

IrakliJani:
@rephus really good point! let me check that

IrakliJani:
@rephus fixed that. included a comment in the code

packages/frontend/src/components/Explorer/CustomDimensionModal/CustomBinDimensionModal.tsx
```diff
@@ -129,7 +149,12 @@ export const CustomBinDimensionModal: FC<{
         }
     }, [setFieldValue, item, isEditing]);
 
-    const handleOnSubmit = form.onSubmit((values) => {
+    const handleOnSubmit = form.onSubmit((unparsedValues) => {
+        // manitne form does not produce zod parsed values
```
rephus, line 153:
```suggestion
        // mantine form does not produce zod parsed values
```

IrakliJani:
@rephus Technically, they can be strings. We don't perform any calculations with them, as far as I remember... But it’s usually not an issue unless someone has an issue with it 😁 

Let me know if it's worth changing!

IrakliJani:
Releasing this as is. @rephus, let me know if I should create a ticket and change it to a string. Although, if we do that, it would mean that we have both (number and string) saved in the DB.

rephus:
no, that's ok for now, as long as we don't get weird like bugs like 0.49999999 in queries when using floats :sweat_smile:

Issue #11213, opened 2024-08-21
Change the appearance of tabs


PriPatel:
- [ ] Use [Mantine Outline](https://mantine.dev/core/tabs/)
- [ ] If the user has tabs, the tile background should be white
- [ ] If the user has tabs, the tile border should change to `gray2`
- [ ] (!IMPORTANT) If the user has no tabs, the appearance of their dashboard should **not** be affected

<img width="806" alt="Screenshot 2024-08-21 at 10 06 09" src="https://github.com/user-attachments/assets/f2591065-b96a-4a2e-ad32-5e4f6aa4cd3b">

note: we're using this as an experiment to see what the white background will feel like

inverts:
Is the `gray2` in the description here referencing the Mantine gray 2?
<img width="470" alt="image" src="https://github.com/user-attachments/assets/cb257cdc-b254-4ece-99c2-a30fd3025932">
Just checking since it doesn't look as dark as the example image in this issue description, but it is close to the same color as the tab border:
<img width="270" alt="image" src="https://github.com/user-attachments/assets/b1e9a902-7ea8-49a7-8d93-7909d26c065b">
The tab border is Mantine's `gray 3`.

PR #11984, opened 2024-10-17
feat: updated dashboard tab appearance

inverts:
Closes: 
https://github.com/lightdash/lightdash/issues/11213

### Description:

Updated tabs according to the specs described in the issue. Appearance remains unchanged when tabs are not in use.

image
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -232,7 +232,7 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                                     });
                                 }
                             }}
-                            mt="xs"
+                            mt={tabsEnabled ? 'sm' : 'xs'}
                             styles={
                                 tabsEnabled
                                     ? {
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -17,6 +17,7 @@ import {
 } from '../../pages/Dashboard';
 import { useDashboardContext } from '../../providers/DashboardProvider';
 import { TrackSection } from '../../providers/TrackingProvider';
+import '../../styles/droppable.css';
 import { SectionName } from '../../types/Events';
 import MantineIcon from '../common/MantineIcon';
 import { LockedDashboardModal } from '../common/modal/LockedDashboardModal';
@@ -209,7 +210,7 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                 setHaveTabsChanged(true);
             }}
         >
-            <Droppable droppableId="tabs" direction="horizontal">
+            <Droppable droppableId="dashboard-tabs" direction="horizontal">
                 {(provided) => (
                     <div {...provided.droppableProps} ref={provided.innerRef}>
                         <Tabs
@@ -237,6 +238,7 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
                                     ? {
                                           root: {
                                               backgroundColor: 'white',
+                                              flexGrow: 1,
                                           },
                                       }
                                     : undefined
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -4,7 +4,7 @@ import {
     type DashboardTab,
     type DashboardTile,
 } from '@lightdash/common';
-import { Button, Group, Tabs } from '@mantine/core';
+import { ActionIcon, Group, Tabs } from '@mantine/core';
 import { IconPlus } from '@tabler/icons-react';
 import cloneDeep from 'lodash/cloneDeep';
 import { useMemo, useState, type FC } from 'react';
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
```diff
@@ -111,13 +111,19 @@ const TileBase = <T extends Dashboard['tiles'][number]>({
...
-                overflow: 'unset',
-
-                border: isEditMode
-                    ? `1px dashed ${theme.colors.blue[5]}`
-                    : `1px solid ${theme.colors.gray[1]}`,
-            })}
+            sx={(theme) => {
+                let border = `1px solid ${theme.colors.gray[1]}`;
+                if (tabs && tabs.length > 1) {
+                    border = `1px solid ${theme.colors.gray[3]}`;
```
ZeRego, line 117: @PriPatel can you confirm what gray we should use? ticket mentioned gray 2 but gray 3 is closer to the screenshot you provided.

packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -232,122 +232,126 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
...
-                                                  padding: 5,
-                                                  borderRadius: 3,
-                                              }
-                                            : undefined
-                                    }
+                                <Tabs.List
+                                    style={{
+                                        backgroundColor:
+                                            'var(--mantine-color-gray-0)',
+                                        paddingLeft: '1.25em',
```
ZeRego, line 253: Where possible use Mantine [props](https://v6.mantine.dev/styles/style-props/#supported-props) and abstract values (xs, sm, md, lg, xl) that can easily be changed in the theme.

packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -232,122 +232,126 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
...
+                                                    borderWidth: 0,
+                                                }}
+                                            >
+                                                Add
+                                            </Button>
+                                        </Group>
+                                    )}
+                                </Tabs.List>
+                            )}
+                            <div style={{ padding: '0 1.25em 1.25em 1.25em' }}>
```
ZeRego, line 303: Could we move the spacing to the `<ResponsiveGridLayout` component or the above component?

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -685,23 +685,25 @@ const Dashboard: FC = () => {
...
-                    handleUpdateTiles={handleUpdateTiles}
-                    handleDeleteTile={handleDeleteTile}
-                    handleBatchDeleteTiles={handleBatchDeleteTiles}
-                    handleEditTile={handleEditTiles}
-                    setGridWidth={setGridWidth}
-                    activeTab={activeTab}
-                    setActiveTab={setActiveTab}
-                    setAddingTab={setAddingTab}
-                />
+                <div style={{ margin: '0 -1.25em 0 -1.25em' }}>
```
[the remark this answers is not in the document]
inverts: Some background first:
The new white background design requires us to expand the grid layout section to the edges of the page, otherwise the individual grid components go to the edge:
image
We could add another layer to pad this and reduce the overall grid layout area, but I decided to follow the design as proposed in https://github.com/lightdash/lightdash/issues/11213, which does go to the edge (and I agree this looks better).
The gray padding here is defined up at the Page component, however. I first experimented with removing the padding here, which subsequently required adding padding back to individual elements like the filter section, for example. The Page component is used all across the app, however. As expected, I started noticing padding issues elsewhere throughout the app:
image
I started trying to track down all the places where we would have to make an adjustment here, but it started feeling very messy. I am also of the opinion that negative margins are to be avoided where possible to avoid confusion, but I gave it a shot here to help avoid what might be much more complexity.

Totally open to suggestions, but I hope that helps explain!

packages/frontend/src/components/DashboardTabs/index.tsx
inverts: I believe it breaks the ResponsiveGridLayout when added on that element, I remember trying that. I remember having a similar experience a long time ago using react-grid-layout, where it seems whatever spacing/padding calculations they're doing in there for organizing the grid items depends on there not being padding on the containing div. I also don't like adding an extra div here so I think I must have tried it on the above component as well but I can't remember. I can play around with it again and double-check.

packages/frontend/src/components/DashboardTabs/index.tsx
inverts: Thanks, will look into this. Still getting familiar with Mantine.

packages/frontend/src/components/DashboardTabs/index.tsx
inverts: Current:
image


When placed on the closest above component (`<Tabs>`):
image

When placed on `ResponsiveGridLayout`:
image

PriPatel (commented):
Amazing work - this is looking much better! 


Feedback here on a couple of visual tweaks:

1. There should be more spacing between the bottom of the tab button and the top of the first tile (so, more padding. The padding at the bottom of the tiles container looks good, so you could just match that!)  
Screenshot 2024-10-23 at 10 36 09

---
2.  There should be no stroke on the bottom of the tab. (border-bottom). 
Screenshot 2024-10-23 at 10 36 53



It should look more like this: 
Screenshot 2024-10-23 at 10 28 12


---


3. The tiles container can take up the full height of the page. Right now, Im not sure what the height is based on, but it feels a bit confusing to see some white and some grey. I expect everything below the tabs to feel like one whole container on the page.
Screenshot 2024-10-23 at 10 37 12

packages/frontend/src/components/DashboardTiles/TileBase/index.tsx
PriPatel: I think gray3 is correct! (looks good in the PR)

inverts:
Resolved the filter spacing and the whitespace expanding to the bottom.
image

Note there is still a small padding of grey at the very bottom coming from the parent Page container, but it is less noticeable. We can resolve this but likely only reasonably with more negative margins for the same reasons we had to above in this PR, which was discouraged. I barely notice that extra grey at the bottom and it doesn't stand out, unlike before - so I decided to leave it as-is. Let me know if we'd prefer to just go ahead and fill that gap with negative margins on the tab container.

ZeRego (approved):

ZeRego:
@all-contributors add @inverts for code

Issue #6705, opened 2023-08-11
Ability to reorder dashboard filters


IrakliJani:
### Description

Ability to Reorder Filters in Dashboard

Users want to reorder how the filters appear at the top of the dashboard (instead of them being stuck in the order you applied them in).

PR #11985, opened 2024-10-17
feat: reorderable dashboard filters

agha4to:
Closes: https://github.com/lightdash/lightdash/issues/6705

### Description:
When editing a dashboard, filters may be reordered by dragging. Blue indicator denotes drop location.
image

Please note:
1. Using `dnd-kit` as it supports dragging of components that wrap to multiple lines.
2. The blue "drop zone" indicator only occupies the existing space between filters because trying to expand it causes glitches when filters wrap on multiple lines. This is also the reason the 6-dot drag handle appears on all filters when in edit mode rather than only on mouseover.
3. While dragging, filter is slightly translucent to help see what's under it. Lmk if you'd like to change that.
4. A dragged filter appears under filters to its right. I looked into fixing this, and the code changes are considerable. Since this is sort of nice to have, I did not continue working on that, but will if you'd like. Lmk.
5. Fyi there is a bug unrelated to this PR where temporary filters disappear on save, although they are saved. Steps to repro: add temporary filter, enter edit mode, save. Temp filter disappears. Reload page to see it was saved, and it displays as a non-temporary filter.

### Reviewer actions

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -27,6 +27,8 @@
     "dependencies": {
         "@casl/ability": "^5.4.4",
         "@casl/react": "^3.0.0",
+        "@dnd-kit/core": "^6.1.0",
+        "@dnd-kit/sortable": "^8.0.0",
         "@emotion/react": "^11.10.6",
         "@hello-pangea/dnd": "^16.5.0",
         "@hookform/error-message": "^2.0.0",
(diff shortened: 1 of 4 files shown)


hamzahc1:
@ZeRego can you create a Render environment for this to test out?

IrakliJani (approved):
this looks great @agha4to!
I pushed a zIndex fix and minor code change.
approving and waiting for the e2e to pass before merging 👍 
thanks!

agha4to:
Thank you, @IrakliJani. Very nice changes on your part too 👍

Issue #11995, opened 2024-10-18
Editors cannot view SQL runner charts


almeidabbm:
Editors cannot see a **saved** sql runner chart. It should have the same permissions as for `SavedCharts`

![Image](https://github.com/user-attachments/assets/0712c9ea-d79f-4cee-a182-4ce748ce6c59)

PR #11997, opened 2024-10-18
fix: check for saved chart view access when running pivot query

almeidabbm:
Closes: #11995 

### Description:

**Steps to test**
1. Create a chart using the `Sql Runner` & save
2. Create a new user with `editor` role
3. Open the saved chart using the `editor` user

Note: `viewers` & `interactive_viewers` should also be able to see the saved chart, as this should have the same permissions as normal charts

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/SavedSqlService/SavedSqlService.ts
+++ b/packages/backend/src/services/SavedSqlService/SavedSqlService.ts
@@ -412,9 +412,26 @@ export class SavedSqlService extends BaseService {
         body: SqlRunnerPivotQueryBody,
         context?: QueryExecutionContext,
     ): Promise<{ jobId: string }> {
+        const { savedSqlUuid } = body;
         const { organizationUuid } = await this.projectModel.getSummary(
             projectUuid,
         );
+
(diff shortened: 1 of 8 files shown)


packages/backend/src/services/SavedSqlService/SavedSqlService.ts
```diff
@@ -412,23 +412,41 @@ export class SavedSqlService extends BaseService {
         body: SqlRunnerPivotQueryBody,
         context?: QueryExecutionContext,
     ): Promise<{ jobId: string }> {
+        const { savedSqlUuid } = body;
```
joaoviana, line 415: there might be just the `slug` on view mode btw, could we account for that as well?

joaoviana: or is it handled by the FE?

joaoviana: I see it now, ignore!

joaoviana (approved):

PR #12002, opened 2024-10-18
feat: use listener middleware to run pivot query

joaoviana:
Closes: N/A

### Description:

This PR ensures that we listen to:
- the sql query run 
- If it runs, ensures that the pivot query is run (on first load)
- If the user changes the config, it will re-run the pivot query and checks for errors
- If the user changes the chart type, it will re-run the pivot query

This introduces a `createListenerMiddleware`: https://redux-toolkit.js.org/api/createListenerMiddleware

The listeners are listed in: `sqlRunnerListeners.ts`
Since we need to listen to `async` actions, there are 2 thunks here: 
- one for running the query (we don't use `react-query`)
- one for running the pivot query against the warehouse

This is helpful to keep the `ContentPanel` unaware of the necessary pieces to build a chart. 

### Reviewer actions


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -143,53 +123,21 @@ export const ContentPanel: FC = () => {
...
-                // reset error highlighting
-                setHightlightError(undefined);
-            },
-            onError: ({ error }) => {
-                showToastError({
-                    title: 'Could not fetch SQL query results',
-                    subtitle: error.message,
-                });
-
-                if (error?.data) {
```
joaoviana, line 159: add this back!

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -143,53 +123,21 @@ export const ContentPanel: FC = () => {
...
 
-    const { mutateAsync: runSqlQuery, isLoading } = useSqlQueryRun(
-        projectUuid,
-        {
-            onSuccess: (_data) => {
-                // reset error highlighting
-                setHightlightError(undefined);
-            },
-            onError: ({ error }) => {
-                showToastError({
```
joaoviana, line 154: add this back!

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
joaoviana: done

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
joaoviana: done

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -348,14 +366,96 @@ export const sqlRunnerSlice = createSlice({
...
+                    state.hasUnrunChanges = false;
+                }
+                state.sqlRows = action.payload.results;
+                state.fileUrl = action.payload.fileUrl;
+            })
+            .addCase(runSqlQuery.rejected, (state, action) => {
+                state.queryIsLoading = false;
+                state.queryError = action.payload ?? undefined;
+
+                state.editorHighlightError = action.payload?.data
```
joaoviana, line 445: when there's an error, set the highlight on editor error.

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -383,15 +484,14 @@ export const {
...
-        sqlRunnerSlice.selectors.selectSavedSqlChart,
         sqlRunnerSlice.selectors.selectColumns,
         sqlRunnerSlice.selectors.selectRows,
         selectProjectUuid,
         selectLimit,
         selectSql,
         (_state, sortBy?: VizSortBy[]) => sortBy,
     ],
-    (sqlChart, columns, rows, projectUuid, limit, sql, sortBy) => {
+    (columns, rows, projectUuid, limit, sql, sortBy) => {
```
joaoviana, line 494: unused variable `sqlChart`

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -594,112 +503,109 @@ export const ContentPanel: FC = () => {
...
+                                                                    }
+                                                                >
+                                                                    <ChartView
+                                                                        config={
+                                                                            c
                                                                         }
-                                                                        isVisible={
-                                                                            selectedChartType ===
-                                                                            c.type
+                                                                        spec={pivotedChartInfo?.data?.getChartSpec(
```
joaoviana, line 535: can use this function any time that the `display` changes

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -143,55 +126,34 @@ export const ContentPanel: FC = () => {
...
+                    limit: DEFAULT_SQL_LIMIT,
+                    projectUuid,
+                }),
+            );
         },
-        [runSqlQuery, dispatch],
+        [dispatch, projectUuid],
     );
 
+    useEffect(() => {
```
joaoviana, line 146: think this is fine. and if there's no error (if it changed and there is a successful query, it clears that toaster)

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
```diff
@@ -38,8 +59,79 @@ const initialState: CartesianChartState = {
...
+            resultsRunner,
+            config,
+            selectedChartType ?? ChartKind.VERTICAL_BAR,
+        );
+
+        const chartData = await dispatch(
+            fetchPivotChartData({ vizDataModel, limit, sql }),
+        ).unwrap();
+
+        const getChartSpec = (orgColors: string[]) => {
```
joaoviana, line 115: this allows each chart to trigger a spec retrieval when the current display changes.

packages/frontend/src/features/sqlRunner/store/thunks.ts
```diff
@@ -0,0 +1,181 @@
...
+            | TableDataModel
+            | PieChartDataModel
+            | CartesianChartDataModel;
+        limit: number;
+        sql: string;
+    }
+>(
+    'cartesianChartBaseConfig/fetchPivotChartData',
+    async ({ vizDataModel, limit, sql }) => {
+        const chartData = await vizDataModel.getPivotedChartData({
```
owlas, line 115: within this function there is secretly api calls happening using the react-query query client directly (i.e. without hooks). We might want to pull those out eventually

packages/frontend/src/features/sqlRunner/store/sqlRunnerListeners.ts
```diff
@@ -0,0 +1,133 @@
...
+        },
+    });
+};
+
+/**
+ * Add a listener for when the chart type is changed.
+ * This listener will fetch the chart (runPivotQuery) data if the chart type has changed.
+ * @param startListening - The startAppListening function from listenerMiddleware.ts
+ */
+export const addChartTypeListener = (
```
owlas, line 107: in future this can be smarter

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
owlas: Could this happen when `sqlRun` is fulfilled instead of in a useeffect?

owlas (approved):
Looks good! Achieves the goal of removing the orchestration from ContentPanel and having it in redux. 

Let's keep building on this pattern going forward, I think we can end up simplifying a lot

packages/frontend/src/features/sqlRunner/store/sqlRunnerListeners.ts
joaoviana: 💯

packages/frontend/src/features/sqlRunner/store/thunks.ts
joaoviana: Oh I had not seen this - it really goes deep in - we can use something different if you want.

joaoviana:
Removed one thunk @owlas  because it was unnecessary: https://github.com/lightdash/lightdash/pull/12002/commits/9c2f92768638778233f4a3792e290665c58ff847 and now `prepareandFetchdata` is the only thunk to get the chart data 

I've also ignored the serialisation error from the `getChartSpec` function - ideally no functions/classes are stored in a store but this is an exception. This is because the viz model needs to have the pivot chart assigned in the class so there's no straightforward way with a selector 
because the selector needs the pivot data

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
joaoviana: no it shouldn't, because this is a react hook 
it happens when `runSqlQuery.reject` happens - an error is set in the store

Issue #11993, opened 2024-10-18
Fix Failed to get GitHub integration on sql runner


rephus:
### Description

An `org editor` and `project developer` get this error on sql runner

![image](https://github.com/user-attachments/assets/24be4981-782c-412a-8704-b80f58e41352)

### Steps to Reproduce the Bug or Issue

.

### Cloud or self-hosting

None

rephus:
Unable to reproduce after the recent refactoring

PR #12012, opened 2024-10-21
fix: permission github dev

rephus:
Closes: https://github.com/lightdash/lightdash/issues/11993
### Description:

image

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/GithubAppService/GithubAppService.ts
+++ b/packages/backend/src/services/GithubAppService/GithubAppService.ts
@@ -160,9 +160,13 @@ export class GithubAppService extends BaseService {
         if (!isUserWithOrg(user)) {
             throw new Error('User is not part of an organization');
         }
+
+        // This endpoint is also used for developers on projects
+        // when using the sql runner, so we should allow access
+        // However github app is an organization property, so we can't check projects
         if (
             user.ability.cannot(
-                'update',
+                'view',
                 subject('Organization', {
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/GithubAppService/GithubAppService.ts
```diff
@@ -160,9 +160,13 @@ export class GithubAppService extends BaseService {
...
             throw new Error('User is not part of an organization');
         }
+
+        // This endpoint is also used for developers on projects
+        // when using the sql runner, so we should allow access
+        // However github app is an organization property, so we can't check projects
         if (
             user.ability.cannot(
-                'update',
+                'view',
```
rephus, line 169: if you have a better solution, let me know.

joaoviana:
Just to confirm, this is so that >= `developer`s can write back, right? because `org editor`s can't - mentioning this because of the main ticket

image

packages/backend/src/services/GithubAppService/GithubAppService.ts
joaoviana: hmmm good point - we can use `virtualView`? or is that too misleading?

joaoviana (approved):

rephus:
> Just to confirm, this is so that >= developers can write back, right? because org editors can't - mentioning this because of the main ticket

Yes, you need to be a developer on the project to access the sql runner

Issue #4958, opened 2023-03-29
Configure timezone on scheduled deliveries 


rephus:
### Description
right now all schedulers are in UTC, I want to be able to configure the timezone that the scheduled delivery is sent in.

The main reason for this is to deal with daylight savings time. 

The other reason for this is that if you're working in a timezone, it's easier to think about the timing there (instead of needing to convert the time to UTC).

![image](https://user-images.githubusercontent.com/1983672/228577997-763021fd-57c3-4aa5-9431-9bfd58419aa7.png)

jpetey75:
This issue had an additional upvote and an open PR a while back (https://github.com/lightdash/lightdash/pull/9738) on a duplicate issue #9387 that I just closed.

Issue #5921, opened 2023-06-15
As a project admin, I want to be able to set the default time zone for my project


TuringLovesDeathMetal:
### Description

As a project admin, I want to set the default timezone that the results in my project are shown in. 

### Acceptance criteria

Miro sketch: https://miro.com/app/board/uXjVM-HOwUI=/?moveToWidget=3458764557197579930&cot=14

- [ ] There is a new section in the project settings which **ONLY project admins** have access to called `time zone settings`
- [ ] In the time zone settings, there is an option to set the default project time zone
- [ ] By default, this is set to UTC
- [ ] There is a drop-down list of standard time zone options 
- [ ] If set to anything but UTC, then we change the time zone in the project
- [ ] Any timestamp fields are affected by the time zone in the project (**?? I think that today, we also change no-timezone timestamps to be in UTC. Should we keep this behaviour and always convert no timezone timestamps to be in the project timezone? I'm not actually sure how we treat no timezone timestamps - we'll need to look into this some more!**)
- [ ] When a user runs a query with timestamp data, the results are shown in the specified timezone. E.g. if I've set my project timezone as Europe/Paris, then these timestamps would show in my results table as `2024-05-29 09:00:15 (+02:00)`
<img width="353" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/cd0fb81d-8613-499d-8cf4-4fb3a02d57c0">

- [ ] When a user uses a filter using timestamp data, we use the correct timezone in the filter and this is shown in the time zone label.
  - [ ] in dashboards:
<img width="391" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/7ed4dc7b-fad0-4916-89b7-1f3a6d12bd99">
  - [ ] in the explore view:
<img width="1050" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/2d2e6a29-1bb9-4fff-960c-3b8bc6e95222">

- [ ] When a user tries to schedule a delivery and there is a project time zone specified, we schedule the delivery based on the project timezone (and this is shown in the scheduled delivery!)
<img width="639" alt="image" src="https://github.com/lightdash/lightdash/assets/31848148/9c1aa2dd-b64d-4a70-9364-297edeb300d4">

### Documentation

- Depending on where we put this configuration, either add a section to the `update your project connection` docs with `time zone settings` (if we add it to the warehouse connection page). Or, create a new `reference` page called `Time zone settings` and add it in there.
[...]

PR #12015, opened 2024-10-21
feat: add scheduled deliveries timezone

almeidabbm:
Closes: [#4958](https://github.com/lightdash/lightdash/issues/4958)

### Description:

- Adds timezone picker to scheduler create/update modal
image
- Converts scheduler dates considering the selected timezone
**Selected time was 18:00 with Europe/Paris timezone**
image
- Displays correct timezone on scheduler lists
image

test-frontend
test-backend
test-cli

### Reviewer actions


packages/frontend/src/components/ReactHookForm/CronInput/index.tsx
```diff
@@ -78,6 +86,7 @@ export const CronInternalInputs: FC<
                     error={error}
                 />
             )}
+            {children}
```
almeidabbm, line 89: Should we hide the timezone picker for hourly frequency?
If so, should we set timezone to `undefined`?

ZeRego: I would keep the timezone picker.
Fun fact time... https://www.timeanddate.com/time/time-zones-interesting.html

almeidabbm: Oh that's good to know!!! 🙏

packages/backend/src/database/migrations/20241021110921_add-timezone-to-scheduler.ts
```diff
@@ -0,0 +1,17 @@
+import { Knex } from 'knex';
+
+const SchedulerTableName = 'scheduler';
+
+export async function up(knex: Knex): Promise<void> {
+    if (await knex.schema.hasTable(SchedulerTableName)) {
+        await knex.schema.alterTable(SchedulerTableName, (table) => {
+            table.string('timezone').nullable();
```
ZeRego, line 8: We know null/undefined represents UTC timezone. We should define it as the default value in the DB and remove all the `timezone ? timezone : 'UTC'` logic from the codebase.

ZeRego (changes requested):

packages/backend/src/database/migrations/20241021110921_add-timezone-to-scheduler.ts
```diff
@@ -0,0 +1,17 @@
+import { Knex } from 'knex';
+
+const SchedulerTableName = 'scheduler';
+
+export async function up(knex: Knex): Promise<void> {
+    if (await knex.schema.hasTable(SchedulerTableName)) {
+        await knex.schema.alterTable(SchedulerTableName, (table) => {
+            table.string('timezone').defaultTo('UTC').notNullable();
```
rephus, line 8: if we do this https://lightdash.slack.com/archives/C02GQKJK84Q/p1729579976689549?thread_ts=1729517981.947559&cid=C02GQKJK84Q
, this should be nullable()

TuringLovesDeathMetal:
@almeidabbm This looks great! 

Two bits of feedback from me:

1. Could we please add the offset in the timezone picker/text that shows? 

This is what I mean: 
image

Otherwise, the names of the timezones are tricky (e.g. I'm in Morocco right now, but I don't know which timezone I'd pick. But I know that I'm UTC -1 from a quick google search)

2. Could we please add an option to search from the timezone picker? 

The search should be by the city name, or by the timezone offset. Ideally, we'd have a bunch of other lists that would map to these options in the background 😅 (like [this guy points out](https://www.gregoryschmidt.ca/writing/timezone-ux-problems#:~:text=2.%20%20Ways%20to%20search%20via%20time%20zone)) but I think just the city name + timezone offset (e.g. `-3` or `UTC - 3`) would be great!

TuringLovesDeathMetal:
sorry, one last thing 😅 

The whole reason that we're doing this is because of daylight savings. Just to double check: does the time zone picker that we're using account for that? 

E.g. will EST change to UTC - 5 on November 3rd? (since it's daylight savings then). Or does the time zone picker not consider daylight savings?

TuringLovesDeathMetal:
> All customers will have to update all their schedulers in order to respect this timezone before the next daylight saving, right ? Perhaps another approach would be to set the timezone at the `project` level, so they only need to change it once ? (we can also reuse this in the future for queries) I mean, this is a great solution which solves the issue, and still good to have, but perhaps we can add the project config to the user UX this and fix the `timezone` issue at a project level. We could create s eparate ticket for that
> 
> More details in here: https://lightdash.slack.com/archives/C02GQKJK84Q/p1729579976689549?thread_ts=1729517981.947559&cid=C02GQKJK84Q

There's an issue for this here @rephus : https://github.com/lightdash/lightdash/issues/5921

I agree, that would be the _**best**_ solution. But, I also think that it could be a bit weird to have a project-level time zone and it _**only**_ affect the scheduled deliveries?

rephus:
>  Do you think it would be a bit weird to have a project-level time zone and it only affect the scheduled deliveries?
@TuringLovesDeathMetal I agree the UI might not be obvious (we can add some comments/labels/tooltips around that input) , but it will be useful for people that want to fix this daylight saving that is coming soon. 

An alternative UI would be to add a `timezone` input in the scheduled delivieres admin panel (we can still store it in the project table) 

What I'm trying to avoid is releasing a feature like "here, you can add timezones to schedulers", and customers saying: "I'm not goingt o update all my schedulers one by one". Just for context, **collectors has ~270 active scheduled deliveries**...

I was thinking we could clearly announce this on a `headway` update, like : 

> Dailyight savings are coming soon, and we've just shipped a new feature to keep your scheduled deliveries in sync. You can update individual deliveries, or update your project to reflect this timezone on all scheduled deliveries

And in the future we can still use it for running queries, or anything that is project related.

TuringLovesDeathMetal:
This looks great @almeidabbm !

- love the search! ✅ 
- definitely more easy to search with the UTC offset ✅ 

Just some small changes:

- can we make the `custom cron expression` option just say `custom`. Then, can we make that first toggle a bit smaller? That way we can stretch the time zone toggle a bit more? so that the names aren't overlapped. 

The other option is to have the UTC offset on one line, then the place on the second line? Which one do you think would be better?? 


- Can we make this say `default time zone` instead of `timezone` (there's actually a space between `time` and `zone` ...which I find super weird, but it's true) 

image

- Can we make the `update` button grey until there's a change? (then it becomes blue)? 

image

- The project default is cropped and there's no indication of what the project default is from the dropdown list. 

image

I think there are two things that we could improve here: 

1. Add `utc -x` as the first thing that appears after `project default` 
2. There's a check mark or some indication beside the item in my list for the time zone that's set (even if it's the default one)

image

^Bonus points if when you open the list, it automatically scrolls to the time zone that's being used in the list 👀

hamzahc1:
@almeidabbm - one thing I noticed is that cities have an underscore in them. Is it possible to have these appear as they would normally with spaces. 
 
CleanShot 2024-10-23 at 09 52 05

almeidabbm:
> This looks great @almeidabbm !
> 
> * love the search! ✅
> * definitely more easy to search with the UTC offset ✅
> 
> Just some small changes:
> 
> * can we make the `custom cron expression` option just say `custom`. Then, can we make that first toggle a bit smaller? That way we can stretch the time zone toggle a bit more? so that the names aren't overlapped.
> 
> The other option is to have the UTC offset on one line, then the place on the second line? Which one do you think would be better??
> 
> * Can we make this say `default time zone` instead of `timezone`. I feel like adding `default` makes it a bit more clear that it can be changed later on? Also, weirdly, there's a space between `time` and `zone` (which I always also get wrong haha)
> 
> * Can we make the `update` button grey until there's a change? (then it becomes blue)?
> 
> * The project default is cropped and there's no indication of what the project default is from the dropdown list.
> 
> I think there are two things that we could improve here:
> 
> 1. Add `utc -x` as the first thing that appears after `project default`
> 2. There's a check mark or some indication beside the item in my list for the time zone that's set (even if it's the default one)
> 
> ^Bonus points if when you open the list, it automatically scrolls to the time zone that's being used in the list as the first row shown in the drop-down 👀

@TuringLovesDeathMetal thank you! I have committed the suggestions with the exception of 2:

- "Add `utc -x` as the first thing that appears after `project default`" - I believe it will take more space and probably end up cropping the text in the select
- "There's a check mark or some indication beside the item in my list for the time zone that's set (even if it's the default one)" - This is because it is only using the `project default` if you unset the time zone in a scheduled delivery. Basically you can select the same time zone as the one set in the `project settings` but that will be an override, which means that the cron won't get updated when you change the project default.

packages/backend/src/services/SchedulerService/SchedulerService.ts
```diff
@@ -425,4 +443,59 @@ export class SchedulerService extends BaseService {
             undefined,
         );
     }
+
+    async updateDefaultTimezoneSchedulers(
```
magnew, line 447: Small thing, but the name similarity between `updateDefaultTimezoneSchedulers` and `updateDefaultSchedulerTimezone` is a little confusing. Could this one be `updateSchedulersWithDefaultTimezone`? Ignore if you disagree :D

magnew (commented):
I'm still testing/reviewing cron scenarios, but I think I found one issue to address: using project default doesn't seem to work for me. See if you can repro:

1. set project TZ to New York
2. set up an hourly delivery for a time in NY timezone (like, 2 mins in the future)
3. set up an hourly deliver in UTC for the equivalent minute
4. which one do you get? I only get the UTC one. 
Screenshot 2024-10-23 at 17 12 48

packages/backend/src/services/SchedulerService/SchedulerService.ts
almeidabbm: Updated 👌

almeidabbm: Also, bug found was fixed in [fix: timezone being passed as null](https://github.com/lightdash/lightdash/pull/12015/commits/452fd87b2a2535b6afb7dd88bd5747d55badf9d3)

magnew: Yup, fixed

magnew (approved):
I did a lot of testing and this seems to be working great now. Nice work!

A couple little UX things that don't need to block this, but I thought I'd call them out cause they were a little confusing (take these suggestions or leave them):

1. When changing project TZ, the default text stays when you focus the input t search. Which is just kind of annoying cause you have to delete what is there to search:
Kapture 2024-10-23 at 16 01 43

2. This input for the individual scheduler looks disabled even though you can interact with it. Maybe th
Screenshot 2024-10-23 at 15 58 27
is was intentional, but it's a little unclear.

almeidabbm:
> I did a lot of testing and this seems to be working great now. Nice work!
> 
> A couple little UX things that don't need to block this, but I thought I'd call them out cause they were a little confusing (take these suggestions or leave them):
> 
> 1. When changing project TZ, the default text stays when you focus the input t search. Which is just kind of annoying cause you have to delete what is there to search:
>    Kapture 2024-10-23 at 16 01 43
>          
>            [
>              
>            
>                Kapture 2024-10-23 at 16 01 43
>              ](https://private-user-images.githubusercontent.com/1864179/379690393-fa82ce71-dcfb-4840-a07d-24f5a5c41010.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mjk3NjI5NTksIm5iZiI6MTcyOTc2MjY1OSwicGF0aCI6Ii8xODY0MTc5LzM3OTY5MDM5My1mYTgyY2U3MS1kY2ZiLTQ4NDAtYTA3ZC0yNGY1YTVjNDEwMTAuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTAyNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEwMjRUMDkzNzM5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OTY5ZjMyYzg3NTA2NmMzODI2NzMyNWJmYjg5OGU1ZDIzZWU0MjFmY2YwYjgxMjFhMGNlYjlkMmJjZWMzMjI1NCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.aOhgeqGNI8Yp2X2bxu73VttdRaTCgtwGNXLaydHjw_M)
>            
>            
>              
>                
>                  
>                
>                
>                  
>                  
>                
>              
>              [
>                
>                  
>                
>              ](https://private-user-images.githubusercontent.com/1864179/379690393-fa82ce71-dcfb-4840-a07d-24f5a5c41010.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mjk3NjI5NTksIm5iZiI6MTcyOTc2MjY1OSwicGF0aCI6Ii8xODY0MTc5LzM3OTY5MDM5My1mYTgyY2U3MS1kY2ZiLTQ4NDAtYTA3ZC0yNGY1YTVjNDEwMTAuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MTAyNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDEwMjRUMDkzNzM5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OTY5ZjMyYzg3NTA2NmMzODI2NzMyNWJmYjg5OGU1ZDIzZWU0MjFmY2YwYjgxMjFhMGNlYjlkMmJjZWMzMjI1NCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.aOhgeqGNI8Yp2X2bxu73VttdRaTCgtwGNXLaydHjw_M)
> 2. This input for the individual scheduler looks disabled even though you can interact with it. Maybe th
> 
> Screenshot 2024-10-23 at 15 58 27 is was intentional, but it's a little unclear.

Changed the variant to 'default'.
Regarding the other behavior we will leave it as is since other searchable selects are working the same way and it had to be a custom implementation

ZeRego (approved):

PR #12022, opened 2024-10-21
fix: compare sqls on change to edit virtual view

joaoviana:
Closes: N/A

### Description:

Found this issue when the user types a SQL query, runs it, but can't save its edit. 
Only they can type the sql and click save. 

reuses a compare sql function. 

**before**

https://github.com/user-attachments/assets/35ee226e-dde9-4cd9-9882-f97d9e56d2d3

**after**

https://github.com/user-attachments/assets/8ac2574f-9fbb-483e-aadc-7da76bfe7a3e

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
+++ b/packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
@@ -66,6 +66,17 @@ const normalizeSQL = (
     }
 };
 
+export const compareSqlQueries = (
+    previousSql: string,
+    currentSql: string,
+    warehouseConnectionType: WarehouseTypes | undefined,
+): boolean => {
+    return (
+        normalizeSQL(previousSql, warehouseConnectionType) !==
+        normalizeSQL(currentSql, warehouseConnectionType)
+    );
+};
+
 export const DEFAULT_NAME = 'Untitled SQL Query';
 
 export interface SqlRunnerState {
@@ -282,7 +293,11 @@ export const sqlRunnerSlice = createSlice({
                 state.successfulSqlQueries.current || '',
                 state.warehouseConnectionType,
             );
-            state.hasUnrunChanges = normalizedNewSql !== normalizedCurrentSql;
+            state.hasUnrunChanges = compareSqlQueries(
+                normalizedNewSql,
+                normalizedCurrentSql,
+                state.warehouseConnectionType,
+            );
         },
         setSqlLimit: (state, action: PayloadAction<number>) => {
             state.limit = action.payload;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -282,7 +293,11 @@ export const sqlRunnerSlice = createSlice({
                 state.successfulSqlQueries.current || '',
                 state.warehouseConnectionType,
             );
-            state.hasUnrunChanges = normalizedNewSql !== normalizedCurrentSql;
```
rephus, line 285: I don't understand why this was not working as expected...

packages/frontend/src/features/virtualView/components/HeaderVirtualView.tsx
```diff
@@ -337,8 +338,17 @@ export const HeaderVirtualView: FC<{
     const history = useHistory();
     const sql = useAppSelector((state) => state.sqlRunner.sql);
     const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
-    const hasUnrunChanges = useAppSelector(
-        (state) => state.sqlRunner.hasUnrunChanges,
```
rephus, line 341: why can't you rely on `state.sqlRunner.hasUnrunChanges` ? is this not duplicating the logic ?

rephus (approved):
Works as expected, I only have 2 questions about the change.

packages/frontend/src/features/virtualView/components/HeaderVirtualView.tsx
joaoviana: can't rely on this because this was the wrong selector for this behaviour. 
This just checks if there are unrunChanges, not if there's a change in original sql vs current sql. 
A similar behaviour is done on the `save` behaviour in `HeaderEdit` 
```tsx
disabled={!config || !sql || !hasChanges}
```
where `hasChanges` is calculated on the component: 

```tsx

    const hasChanges = useMemo(() => {
        if (!initialSavedSqlChart) return false;
        const changedSql = sql !== initialSavedSqlChart.sql;
        const changedLimit = limit !== initialSavedSqlChart.limit;
        const changedConfig = !isEqual(config, initialChartConfig);

        return changedSql || changedLimit || changedConfig;
    }, [initialSavedSqlChart, initialChartConfig, sql, limit, config]);
```

you can see here that we also check the sql against the initial sql

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
joaoviana: see https://github.com/lightdash/lightdash/pull/12022#discussion_r1810216561

Issue #11470, opened 2024-09-09
Display aggregations and X-axis ordering in line with the fields themselves. 


PriPatel:
**Before**
<img width="403" alt="Screenshot 2024-09-09 at 13 12 35" src="https://github.com/user-attachments/assets/ba1187d5-131c-4754-b545-f6bd94a394da">
**After**
<img width="482" alt="Screenshot 2024-10-17 at 16 14 34" src="https://github.com/user-attachments/assets/ec837538-7ea8-49c3-82a8-ad2c6bc41565">

---

**Aggregations**
- [x]  The aggregation should be displayed next to the field itself (and there is no need to specify the word aggregation
- [x] Clicking on the aggregation should open the dropdown (no changes to the dropdown appearancefor now)
- [x] Change `Any` to `Any Value` in the aggregations dropdown
- [x] Aggregation pill should be bg -  indigo0 50% transparency. Check figma

**X-axis ordering**
- [x] Move the ordering option into a similar pill in the x-axis field input
- [x] Default should be `Ascending` with an arrow pointing to the right
- [x] Clicking on this will simply reverse the direction and show `Descending`
- [x]  Ordering pill should be bg - gray0.  Check figma

PriPatel:
@joaoviana I've asked about what should happen by default and if the options are truly ever just `ascending` and `descending` ! x

PR #12054, opened 2024-10-23
style: apply new designs to field configuration (x, y)

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/11470

### Description:

See main ticket and request figma link

Screenshot 2024-10-23 at 16 34 56

THis PR: 
- Introduces `usePillSelectStyles` - allows us to create PillSelect's with Mantine hooks
- Allows the user to update the sort and aggregation inside the fieldReferenceSelect
- Replaces the `+ add ` button with an action icon
- Replaces the `clearable` prop with a different approach (with the - action icon)
- Refactors the `setSortBy` action. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DataViz/config/DataVizAggregationConfig.tsx
+++ b/packages/frontend/src/components/DataViz/config/DataVizAggregationConfig.tsx
@@ -3,7 +3,14 @@ import {
     VizAggregationOptions,
     type VizValuesLayoutOptions,
 } from '@lightdash/common';
-import { Box, Group, Select, Text, useMantineTheme } from '@mantine/core';
+import {
+    Box,
+    Group,
+    Select,
+    Text,
+    Tooltip,
+    useMantineTheme,
+} from '@mantine/core';
 import {
     IconAsterisk,
     IconMathFunction,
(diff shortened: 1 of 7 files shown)


joaoviana:
If this meets the acceptable requirements @PriPatel , could we have @magnew review this please? 

So essentially after your design review -> we can get Marshall to review if the remaining bits can be done separately? What do you think?

PriPatel:
Review here! 
https://www.notion.so/lightdash/Review-128a63207a7a80bc8ea8c4f061bd865a

- There is one big logic change that needs to take a place and a few styling bits that need to match Figma (less important than the logic change which might break charts if we proceed this way!)

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
```diff
@@ -284,10 +284,23 @@ export const cartesianChartConfigSlice = createSlice({
...
+            if (action.payload.direction) {
+                fieldConfig.sortBy = [
+                    {
+                        reference: action.payload.reference,
+                        direction: action.payload.direction,
+                    },
+                ];
+            } else {
+                fieldConfig.sortBy = undefined;
+            }
```
joaoviana, line 303: can clear now the sortBy

packages/frontend/src/components/DataViz/hooks/usePillSelectStyles.tsx
```diff
@@ -0,0 +1,58 @@
+import { createStyles } from '@mantine/core';
+
+export const usePillSelectStyles = createStyles(
```
joaoviana, line 3: this helps with creating a pill select
image

PriPatel:
@joaoviana Just remembered(!), as we're not doing the full milestone, the tooltip should be added here! 'Aggregation type' when you hover over the y pill and 'Sort by' on the x axis pill

joaoviana:
@PriPatel added tooltips!

@magnew  - updated e2e tests and added test ids where necessary

packages/frontend/src/components/DataViz/config/DataVizAggregationConfig.tsx
```diff
@@ -72,40 +80,46 @@ export const DataVizAggregationConfig: FC<Props> = ({
...
+                    value && onChangeAggregation(value)
+                }
+                classNames={{
+                    item: classes.item,
+                    dropdown: classes.dropdown,
+                    input: classes.input,
+                    rightSection: classes.rightSection,
+                }}
+                styles={{
+                    input: {
```
magnew, line 115: I can't remember: is the classNames styling no good when a parameter is required (aggregation here)? It's a little odd that we are stying input with styles and classnames. But maybe that's life in modern react 😂 . I'm sure this isn't the only place we are doing things like that.

packages/frontend/src/components/DataViz/hooks/usePillSelectStyles.tsx
magnew: Cool!

magnew (approved):
✅ Looks great
✅ Works great - I don't see any issues
✅ Code looks good!

Issue #12055, opened 2024-10-23
When I try to create a second 'Chart from within a dashboard', without saving the dashboard, the first chart gets deleted. 


PriPatel:
1. Go to edit a dashboard
2. Create new chart from the dashboard
3. Save the chart
4. Now you're back in the dashboard. The new chart shows up fine
5. Now try to create a new chart
6. You are now directed away from the unsaved dashboard and you lose all your changes.

PriPatel:
This might be 'by design'.
I would expect at the very least to be warned and prompted to save my changes if this is the case.

PR #12083, opened 2024-10-24
fix: warn user when leaving dashboard to create chart within dashboard on unsaved changes

rephus:
…d on unsaved changes

Closes: https://github.com/lightdash/lightdash/issues/12055

### Description:

This behaviour was intentional, but I agree it doesn't make much sense. 

I'm throwing a warning if you're trying to leave the dashboard to create a chart with unsaved changes. 

[Screencast from 24-10-24 16:49:49.webm](https://github.com/user-attachments/assets/50b8a4e7-4a46-4dcc-bbf6-c055a248c752)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -517,9 +517,7 @@ const Dashboard: FC = () => {
                 if (
                     !prompt.pathname.includes(
                         `/projects/${projectUuid}/dashboards/${dashboardUuid}`,
-                    ) &&
-                    // Allow user to add a new table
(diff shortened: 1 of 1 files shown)


PriPatel:
@rephus I can't replicate this now, so it does feel like it was a bug and not by design! Can you double check? THis was taking place at the same time as this https://lightdash.slack.com/archives/C02DGV9RWKV/p1729693089665969 . I wonder if it was connected?

rephus:
the popup works, but the real issue happens when canceling the chart in dashboard, where the dashboard storage is cleared.

Issue #11842, opened 2024-10-08
SQL Runner Chart doesn't update when changing SQL


almeidabbm:
### Description

Chart doesn't update when updating the sql, this is most likely because of react `useQuery` and we need to update the query key

https://github.com/user-attachments/assets/42d0a291-621c-4f79-812a-9c860ae200d2

### Steps to Reproduce the Bug or Issue

1. Write SQL
2. Click "Run Query"
3. Open chart tab
4. Change SQL
5. Click "Run Query"
6. Open chart tab - chart is only updated once you change the chart config

### Cloud or self-hosting

None

joaoviana:
this might be fixed already because of the react-query replacement with a `thunk` @magnew, I think this won't be replicable anymore

magnew:
Yea, this is different now. There might be some opportunity to improve what we have now, but this specific issue is fixed.

magnew:
Actually re-opened after some edge case testing.

PR #12085, opened 2024-10-24
fix: update chart when sql changes

magnew:
Closes: #11842 

### Description:

Fixes keeping old results around when a new SQL query is run. This does 2 things:

1. Always reset the chart config when a new query is run. We had been trying to keep some of it around, but that wasn't working right anyway, so I think just updating it is fine. It seems like the way the UI is set up, it won't be common for people to work on a chart, change the SQL, then work on the chart again. 
2. Remove the old results from the store when an error occurs. They were sticking around and it was confusing. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DataViz/store/barChartSlice.ts
+++ b/packages/frontend/src/components/DataViz/store/barChartSlice.ts
@@ -27,6 +27,7 @@ export const barChartConfigSlice = createSlice({
         });
         builder.addCase(prepareAndFetchChartData.rejected, (state, action) => {
             state.chartDataLoading = false;
+            state.chartData = undefined;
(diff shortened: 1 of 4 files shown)


magnew:
> E2E tests are failing, could you check them before I start reviewing? seems like an issue with error states when sql changes @magnew

For sure

joaoviana (commented):
should we also clear the data in the sql runner slice?

magnew:
> should we also clear the data in the sql runner slice?

In this case we are clearing in on prepareAndFetchChartData, which the SQL runner slice doesn't listen to, so this should be sufficient. There might be a reason to revisit error handling in the sql query run, but I don't thin it affects this issue.

joaoviana (approved):

PR #12100, opened 2024-10-25
fix: improve ux for registering users

rephus:
Closes: https://github.com/lightdash/lightdash/issues/12097

### Description:

Sometimes we allow user registering when they are in single org instances, and without invitations. We can detect these cases earlier on and avoid creating the user . 

This not only improves the overall UX for the user registering, but also, it reduces errosr in the long term. We recently discovered  some of these users were created some time ago (they got an error, and they forgot about it), and then, when they try to register again, their user is already taken. 

My fix is already taken into account special register cases like `whitelisted domains`, empty user/orgs, and enabled OIDC for SSO. 

### User and password flow 

Before:

[Screencast from 25-10-24 13:53:07.webm](https://github.com/user-attachments/assets/c3d3fa5a-6490-4669-aba3-4a7c84567f00)

After:

[Screencast from 25-10-24 13:57:06.webm](https://github.com/user-attachments/assets/83f2c1f4-bf87-4ad3-a271-4f2985e5cec2)

### SSO 

Before:

[Screencast from 25-10-24 14:04:48.webm](https://github.com/user-attachments/assets/2e9af50a-88f2-4b90-8a94-f8eda845ddde)

AFter:

[Screencast from 25-10-24 14:04:18.webm](https://github.com/user-attachments/assets/3fa1ae0f-26be-4153-9d6c-06f1c2341441)

### Generic messages for user taken 

This can still happen on shared instances

Before:
Screenshot from 2024-10-25 14-06-33

After:
the blue text with the red background hurt my eyes... but this is the default, perhaps we can improve it on a separate PR 
Screenshot from 2024-10-25 14-08-12

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -509,7 +509,7 @@ export class UserService extends BaseService {
                 await this.organizationModel.getAllowedOrgsForDomain(
                     getEmailDomain(email),
                 )
-            ).length !== 1
+            ).length < 1
         ) {
             throw new ForbiddenError(
                 `You can't register a new user with email ${email} on this instance. Please contact your organization administrator and ask for an invite`,
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -488,6 +488,35 @@ export class UserService extends BaseService {
         }
     }
 
+    /**
+     * In this method we check if the user is allowed to register in this instance before registerUser happens.
+     * If the instance is a single org, and there are already users and orgs,
+     * and the user's email domain is allowed in the org whitelisted,
+     * we throw a ForbiddenError.
+     * @param inviteCode invite code
+     * @param email email for the user registering
+     */
+    async checkNewUserRegistrationAllowed(
+        inviteCode: string | undefined,
+        email: string,
+    ) {
+        if (
+            inviteCode === undefined &&
+            !this.lightdashConfig.allowMultiOrgs &&
+            (await this.userModel.hasUsers()) &&
+            (await this.organizationModel.hasOrgs()) &&
+            (
+                await this.organizationModel.getAllowedOrgsForDomain(
+                    getEmailDomain(email),
+                )
+            ).length !== 1
+        ) {
+            throw new ForbiddenError(
+                `You can't register a new user on this instance. Please contact your organization administrator and ask for an invite`,
+            );
+        }
+    }
+
     async loginWithOpenId(
         openIdUser: OpenIdUser,
         authenticatedUser: SessionUser | undefined,
@@ -673,6 +702,10 @@ export class UserService extends BaseService {
         }
 
         // Create user
+        await this.checkNewUserRegistrationAllowed(
+            inviteCode,
+            openIdUser.openId.email,
+        );
         const createdUser = await this.activateUserWithOpenId(
             openIdUser,
             inviteCode,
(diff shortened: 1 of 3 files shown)


packages/backend/src/services/UserService.ts
```diff
@@ -673,6 +702,10 @@ export class UserService extends BaseService {
         }
 
         // Create user
+        await this.checkNewUserRegistrationAllowed(
```
rephus, line 705: This code in `loginWithOpenId` is triggered after the `OIDC` linking logic

packages/backend/src/services/UserService.ts
```diff
@@ -488,6 +488,35 @@ export class UserService extends BaseService {
...
+     * @param email email for the user registering
+     */
+    async checkNewUserRegistrationAllowed(
+        inviteCode: string | undefined,
+        email: string,
+    ) {
+        if (
+            inviteCode === undefined &&
+            !this.lightdashConfig.allowMultiOrgs &&
+            (await this.userModel.hasUsers()) &&
```
rephus, line 506: short cirtuit evaluation, if there is an invite code or it is a shared instance, we don't execute the following queries.

rephus:
@PriPatel I'm not sure if you are interested in reviewing these changes for the `bad` user registering flow

packages/backend/src/services/UserService.ts
```diff
@@ -488,6 +488,35 @@ export class UserService extends BaseService {
...
+            (await this.organizationModel.hasOrgs()) &&
+            (
+                await this.organizationModel.getAllowedOrgsForDomain(
+                    getEmailDomain(email),
+                )
+            ).length !== 1
```
ZeRego, line 512: Better check if bellow 1 in case `allowMultiOrgs` is disabled after a couple of orgs are created. 
```suggestion
            ).length < 1
```

rephus: This check is for `allowedOrgsForAdmin` where we should only return 1 org if the email is whitelisted by an org. 

In here we are checking `===1` , because we only should have 1 matching org https://github.com/lightdash/lightdash/blob/main/packages/backend/src/services/UserService.ts#L695. Do you think we  should we change that logic too ?

packages/backend/src/services/UserService.ts
```diff
@@ -488,6 +488,35 @@ export class UserService extends BaseService {
...
+        ) {
+            throw new ForbiddenError(
+                `You can't register a new user on this instance. Please contact your organization administrator and ask for an invite`,
```
owlas, line 515: We need to include the email in here too. Some users have been confused why they don't have access - it could be because the email we got from their SSO provider doesn't match what they have in the database.

rephus: image
Done

packages/backend/src/services/UserService.ts
ZeRego: > This check is for allowedOrgsForAdmin where we should only return 1 org if the email is whitelisted by an org.

You meant `getAllowedOrgsForDomain` ? It can return multiple orgs. 

> In here we are checking ===1 , because we only should have 1 matching org

That assumption is incorrect. 
If the user only has 1 matching org, we auto join them to the org.
If the user has more than 1, they will have the option in the UI to decide which one to join.
If there is no matching org, one will be created (logic for this is in FE 😢 ) or throw an error when they can't (flow we are trying to fix).

Having `< 1` would prevent the third flow.

ZeRego (approved):

PR #12102, opened 2024-10-25
fix(auth): show actionable error message when Okta authentication has…

ZeRego:
… access denied

Closes: [#12014](https://github.com/lightdash/lightdash/issues/12014)

### Description:

Show better error message when user doesn't have access to Lightdash. 
Other errors ([see list](https://developer.okta.com/docs/reference/error-codes/#example-errors-for-openid-connect-and-social-login)) are mostly related to bad integration and not relevant to the user authenticating so we just log them properly.

Before:
okta error

After:

okta fix

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/authentication/strategies/oktaStrategy.ts
+++ b/packages/backend/src/controllers/authentication/strategies/oktaStrategy.ts
@@ -4,7 +4,9 @@ import {
     LightdashError,
     OpenIdIdentityIssuerType,
     OpenIdUser,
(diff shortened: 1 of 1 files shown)


almeidabbm (approved):
If there's an error in `setupOktaIssuerClient` we don't show an error message:

image

Approving since this fixes the issue for `access_denied` and this one is mostly related to bad environment variables, perhaps can be done on a separate ticket?

PR #12109, opened 2024-10-25
fix: check if y is defined

ZeRego:
Closes: #12108 

### Description:

`y` is possibly undefined so we need to check it before accessing the value inside

### Reviewer actions


rephus (approved):
Looks good, and it seems, if `dim` === '', this should not cause any other issues further down in the code

PR #12136, opened 2024-10-28
feat: add pagination to /dataCatalog/metrics endpoint

almeidabbm:
Closes: #12135 

### Description:

- introduces `required_attributes` column to `catalog_search` (`null` by default, and then user has to refresh dbt / compile project for all to be populated for each field and table - this is handled on the db query)
- this also ensures that the filtering around required_attributes (or user attributes) and table access (like model selection, tag selection) is happening on the the db query instead of post-db-query

 
### Reviewer actions

test-frontend

### DIFF
--- a/packages/backend/src/models/CatalogModel/CatalogModel.ts
+++ b/packages/backend/src/models/CatalogModel/CatalogModel.ts
@@ -7,10 +7,19 @@ import {
     FieldType,
     NotFoundError,
     UnexpectedServerError,
+    type KnexPaginateArgs,
+    type KnexPaginatedData,
 } from '@lightdash/common';
 import { Knex } from 'knex';
-import { CatalogTableName } from '../../database/entities/catalog';
-import { CachedExploreTableName } from '../../database/entities/projects';
+import {
+    CatalogTableName,
+    type DbCatalog,
+} from '../../database/entities/catalog';
+import {
+    CachedExploreTableName,
+    type DbCachedExplore,
+} from '../../database/entities/projects';
+import KnexPaginate from '../../database/pagination';
 import { wrapSentryTransaction } from '../../utils';
 import {
     getFullTextSearchQuery,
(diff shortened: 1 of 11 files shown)


packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -524,17 +542,21 @@ export class CatalogService<
...
-        const catalogMetrics = await this.searchCatalog(
+        // ! Pagination here might not work as expected because there's some filtering being applied after the data is fetched
+        const paginatedCatalogMetrics = await this.searchCatalog(
             projectUuid,
             userAttributes,
             search,
             CatalogType.Field,
             CatalogFilter.Metrics,
+            paginateArgs,
         );
```
almeidabbm, line 553: This is important, we paginate on the query, but since we then apply some filters on the JS side (e.g. to check user attributes) the page might return less results than the `pageSize`
Not sure how we can solve this yet, but it is also a problem in other places of the app

packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -118,15 +129,23 @@ export class CatalogModel {
             .orderBy('search_rank', 'desc')
             .limit(limit ?? 50);
 
-        const catalogItems = await catalogItemsQuery;
+        const paginatedCatalogItems = await KnexPaginate.paginate(
+            catalogItemsQuery.select<(DbCatalog & { explore: Explore })[]>(),
+            paginateArgs,
```
joaoviana, line 134: if it's undefined, will it return everything?

packages/backend/src/models/CatalogModel/utils/index.ts
```diff
@@ -29,6 +30,10 @@ export const convertExploresToCatalog = (
             description: field.description || null,
             type: CatalogType.Field,
             field_type: field.fieldType,
+            required_attributes:
+                field.requiredAttributes ??
+                baseTable.requiredAttributes ??
+                null,
```
almeidabbm, line 36: This has the same order as we previously had in `packages/backend/src/models/CatalogModel/utils/parser.ts` line 27, field attributes come before base table

Issue #12003, opened 2024-10-18
the project list should be ordered differently


TuringLovesDeathMetal:
### Description

- production projects should be at the top, listed by the most users who have viewer+ access to them (i.e. project with 900 users is at the top, project with 400 users is below that)
- preview projects should be below the production projects, listed in order of most recently created (most recent at the top)

PR #12140, opened 2024-10-28
feat: project list ordering

IrakliJani:
Closes: #12003

### OG description from the ticket:

- production projects should be at the top, listed by the most users who have viewer+ access to them (i.e. project with 900 users is at the top, project with 400 users is below that)
- preview projects should be below the production projects, listed in order of most recently created (most recent at the top)

### Description:

first task mentions users who have viewer+ access. Since we are optimizing for a large number of project lists, querying permissions for all users is going to be very resource-intensive (though I'm open to suggestions if you think this is achievable).

I decieded to order by three components:

- if project is default it goes first, preview goes last
- then we order by direct project members + group inherited members (descending)
- and in the last we order by created_at (descending)

second sort is the most controversial, but let me clarify:
- organization-inherited users do not matter because they are the same for all projects.
- project members and groups might have overlapping users, but this isn't a major issue, as duplication also contributes to the "popularity" of the project.

CleanShot 2024-10-28 at 22 42 01@2x

CleanShot 2024-10-28 at 22 42 37@2x

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -64,7 +64,10 @@ import {
 } from '../../database/entities/organizations';
 import { PinnedListTableName } from '../../database/entities/pinnedList';
 import { ProjectGroupAccessTableName } from '../../database/entities/projectGroupAccess';
-import { DbProjectMembership } from '../../database/entities/projectMemberships';
+import {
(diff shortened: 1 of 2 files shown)


TuringLovesDeathMetal:
ughhhhhh amaaaaazing 😚 👌 so much better already

ZeRego (approved):

Issue #12075, opened 2024-10-24
analytics for scheduled deliveries time zones


TuringLovesDeathMetal:
### Description

- add `time zone` parameter to `scheduler_created`, `scheduled_updated` events which is in the format like `(UTC - 4:00) - America Los Angeles` 
- add new event `default_project_time_zone_updated` which is triggered when the default proejct time zone is updated. 
  - user_id
  - organization_id
  - project_id
  - time_zone  (like `(UTC - 4:00) - America Los Angeles` )

rephus:
@TuringLovesDeathMetal we decided to not update the timezone at the project level (yet),  this is only for schedulers. So I'm renaming the  event to `default_scheduler_time_zone.updated` FYI


![image](https://github.com/user-attachments/assets/f9b17680-88e7-47c1-95d1-684aee96e5cf)

PR #12142, opened 2024-10-29
chore: add analytic events to scheduler timezone changes

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12075

### Description:

Screenshot from 2024-10-29 09-43-53

Screenshot from 2024-10-29 09-50-43

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -4704,7 +4704,7 @@ export class ProjectService extends BaseService {
             userId: user.userUuid,
             properties: {
                 projectId: projectUuid,
-                organizatoinUuid: project.organizationUuid,
+                organizationUuid: project.organizationUuid,
                 timeZone: getTimezoneLabel(schedulerTimezone),
             },
         });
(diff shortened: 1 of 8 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -4698,6 +4699,16 @@ export class ProjectService extends BaseService {
                 schedulerTimezone,
             );
 
+        this.analytics.track({
+            event: 'default_scheduler_timezone.updated',
+            userId: user.userUuid,
+            properties: {
+                projectId: projectUuid,
+                organizatoinUuid: project.organizationUuid,
```
almeidabbm, line 4707:
```suggestion
                organizationUuid: project.organizationUuid,
```

almeidabbm (changes requested):
Tests are failing because in `packages/backend/src/models/SchedulerModel/index.ts` we're not doing a fallback for `timezone` to `scheduler.timezone ?? undefined` so it is `null` but being cast as `Scheduler`

we can solve this in a separate PR but it should be a single line change

almeidabbm (approved):

Issue #12047, opened 2024-10-23
my `number` metric type is showing the incorrect total value in a pivot table. 


TuringLovesDeathMetal:
### Description

Slack thread here: https://lightdash.slack.com/archives/C04FQQU4RL7/p1729681989919299

If I have a `number` metric type defined like this:

```yaml
- name: revenue
    meta:
      metrics:
        total_revenue:
          type: sum
- name: cost
    meta:
      metrics:
        total_cost:
          type: sum
        total_percent_profit: # this is my number metric!
          type: number
          sql: 100 * ${total_revenue}/${total_cost}
```

In my unpivoted table, the column total shows me the correct answer (it takes `SUM(total_revenue across column)/SUM(total_cost across column)`). But, when I pivot the data, it sums each row (`SUM(total_revenue / total_cost)`). This gives me the wrong answer. 

Either we should remove these metric types (number) or we should make them work properly in pivoted tables. I think that if we know the metric type = `number`, then we should just calculate the row total based on the other metric row totals. This would only work for `number` metric types that are built from other metrics that are sums or counts.

rephus:
@TuringLovesDeathMetal IIRC there were some recent discussions around this, did we decide what solution should we apply here ?

Issue #9981, opened 2024-05-06
I want totals to appear in my pivot tables for all metric types 


TuringLovesDeathMetal:
### Description

Today, totals only appear for `sum` and `count` metric types in pivot tables.

I want totals to be calculated for all of my metric types (e.g. `count distinct`, `average`, etc.)

This includes both column totals and row totals

TuringLovesDeathMetal:
Is this related to #2907 ?

rephus:
For context, we don't calculate totals on pivot tables for metric type `number`, because the `sum` of numbers is not a correct total. 

Let's take this example: 

I have a metric which is a type:number calculated from `sums`

```
      - name: amount
        description: Total amount (AUD) of the individual payment
        meta:
          metrics:
            total_revenue:
              type: sum
              description: Sum of all payments
            total_cost:
                type: number
                sql: ${total_revenue} - 10
            total_percent_profit: 
                type: number
                sql: 100 * ${total_revenue}/${total_cost}
```

WHen I calculate totals in a non-pivot table, the result is correct, we calculate this on a special backend endpoint, by grouping the right columns.

![image](https://github.com/user-attachments/assets/d4cb8a4c-cef9-4b67-a7e2-3079bec76925)

When I pivot the table, the frontend is `aggregating` all the rows, to calculate the total, (because we allowed NUMBER on `isSummable` function) , the total result is wrong.

![image](https://github.com/user-attachments/assets/8592655a-fccc-44aa-bd9c-8bcced03f6a7)

FOr this use case, we could `manually` calculate the total, by running the `100 * ${total_revenue}/${total_cost}` operation on the `total of the individual metrics`, like: 

> 100 * (994.16 /	854.16) = 116

However, the main issue is that in the `frontend`, this total_revenue and total_cost might not be present in the result of the query, so we'll have to move this calculation to the backend somehow, per dimension pivot, which is not an easy thing to do. 

So we decided to temporarily remove this total calculation from number types

clarajdavison:
Seconding this issue, we have a lot of problems with count_distinct columns not totaling up, and it would be really helpful to make that possible

jesshitchcock:
Explanation of the limitation in this loom [video](https://www.loom.com/share/73273037f57e417ca9fde18299b78ede).

paul-bridi:
[...]

PR #12152, opened 2024-10-29
fix: remove number metric total calculation on pivot

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12047
Related: https://github.com/lightdash/lightdash/issues/9981

### Why we have decided to not support number metrics on totals 

https://github.com/lightdash/lightdash/issues/9981#issuecomment-2443881249

For context, we don't calculate totals on pivot tables for metric type `number`, because the `sum` of numbers is not a correct total. 

Let's take this example: 

I have a metric which is a type:number calculated from `sums`

```
      - name: amount
        description: Total amount (AUD) of the individual payment
        meta:
          metrics:
            total_revenue:
              type: sum
              description: Sum of all payments
            total_cost:
                type: number
                sql: ${total_revenue} - 10
            total_percent_profit: 
                type: number
                sql: 100 * ${total_revenue}/${total_cost}
```

WHen I calculate totals in a non-pivot table, the result is correct, we calculate this on a special backend endpoint, by grouping the right columns.

image

When I pivot the table, the frontend is `aggregating` all the rows, to calculate the total, (because we allowed NUMBER on `isSummable` function) , the total result is wrong.

image

FOr this use case, we could `manually` calculate the total, by running the `100 * ${total_revenue}/${total_cost}` operation on the `total of the individual metrics`, like: 

> 100 * (994.16 /	854.16) = 116

However, the main issue is that in the `frontend`, this total_revenue and total_cost might not be present in the result of the query, so we'll have to move this calculation to the backend somehow, per dimension pivot, which is not an easy thing to do. 

So we decided to temporarily remove this total calculation from number types 

### Description

before:
Screenshot from 2024-10-29 11-49-33

after:

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useColumnTotals.ts
+++ b/packages/frontend/src/hooks/useColumnTotals.ts
@@ -31,15 +31,14 @@ export const isSummable = (item: Item | undefined) => {
     if (isCustomDimension(item)) {
         return false;
     }
-    const numericTypes: string[] = [
-        DimensionType.NUMBER,
-        MetricType.NUMBER,
-        MetricType.COUNT,
-        MetricType.SUM,
-    ];
+    const numericTypes: string[] = [MetricType.COUNT, MetricType.SUM];
+    const isNumberDimension =
+        isDimension(item) && item.type === DimensionType.NUMBER;
+    const isNumbericType =
+        numericTypes.includes(item.type) || isNumberDimension;
     const isPercent = item.format === 'percent';
     const isDatePart = isDimension(item) && item.timeInterval;
-    return numericTypes.includes(item.type) && !isPercent && !isDatePart;
+    return isNumbericType && !isPercent && !isDatePart;
 };
 
 const getResultColumnTotals = (


packages/frontend/src/hooks/useColumnTotals.ts
```diff
@@ -31,15 +31,14 @@ export const isSummable = (item: Item | undefined) => {
...
         return false;
     }
-    const numericTypes: string[] = [
-        DimensionType.NUMBER,
-        MetricType.NUMBER,
-        MetricType.COUNT,
-        MetricType.SUM,
-    ];
+    const numericTypes: string[] = [MetricType.COUNT, MetricType.SUM];
+    const isNumberDimension =
```
rephus, line 35: Keep DimensionType.NUMBER support, which I believe is correct, and not related with this bugfix.

DimensionType.NUMBER === 'string' so we can't leave it in the numericType, as it matches `MetricType.NUMBER`

IrakliJani (approved):

Issue #8307, opened 2023-12-14
Preview environments in the project selector should only show up for users with developer access + above for a project


TuringLovesDeathMetal:
### Description

By default, when you build a preview environment, everyone inherits the access that they have for the project that the preview environment was built from. 

The issue with this is that viewers and editors end up accidentally working in preview environments, trying to share links/explore things/build charts without realizing that their work will be deleted once the preview environment is removed. 

But, there are use cases where people want to have editors/viewers with access to preview environments (e.g. building a new product launch chart/dashboard in staging and wanting to share it with a business user to make sure that it's correct). 

Instead of having preview environments appear in the project list for _**everyone**_, they should only be visible to developers + admins. But, everyone else should still have access to these projects, but they should be hidden by default from the project selector (i.e. as a developer, I can share the project with you, a viewer, but you can't see it in the selector list by default).

PR #12158, opened 2024-10-29
feat: project list no previews for members and below

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/8307

### Description:

test-frontend
test-backend
test-cli

internal discussion: https://lightdash.slack.com/archives/C07TP0EN8PP/p1730197569059179

developer:

CleanShot 2024-10-29 at 19 49 18@2x

member:

CleanShot 2024-10-29 at 19 49 44@2x

member able to access preview via url:

CleanShot 2024-10-29 at 19 50 20@2x

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -322,9 +322,9 @@ export class ProjectService extends BaseService {
                     const upstreamProject = await this.projectModel.get(
                         data.upstreamProjectUuid,
                     );
-                    if (upstreamProject.type === ProjectType.DEFAULT) {
+                    if (upstreamProject.type === ProjectType.PREVIEW) {
                         throw new ForbiddenError(
-                            'Cannot create a preview project from a default project',
+                            'Cannot create a preview project from a preview project',
                         );
                     }
                 }
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -533,6 +533,10 @@ export class ProjectService extends BaseService {
                 const projectSummary = await this.projectModel.getSummary(
                     data.upstreamProjectUuid,
                 );
+                await this.copyUserAccessOnPreview(
+                    data.upstreamProjectUuid,
+                    projectUuid,
+                );
                 await this.copyContentOnPreview(
                     data.upstreamProjectUuid,
                     projectUuid,
@@ -673,6 +677,65 @@ export class ProjectService extends BaseService {
         return {
             jobUuid: job.jobUuid,
         };
+    }
+
+    /*
+        Copy user permissions from upstream project
+        if the user is a viewer in the org, but an editor in a project
+        we want the user to also be an editor in the preview project
+    */
+    async copyUserAccessOnPreview(
+        upstreamProjectUuid: string,
+        previewProjectUuid: string,
+    ): Promise<void> {
+        this.logger.info(
+            `Copying access from project ${upstreamProjectUuid} to preview project ${previewProjectUuid}`,
+        );
+        await wrapSentryTransaction<void>(
+            'duplicateUserAccess',
+            {
+                previewProjectUuid,
+                upstreamProjectUuid,
+            },
+            async () => {
+                const projectAccesses =
+                    await this.projectModel.getProjectAccess(
+                        upstreamProjectUuid,
+                    );
+                const groupAccesses =
+                    await this.projectModel.getProjectGroupAccesses(
+                        upstreamProjectUuid,
+                    );
+
+                this.logger.info(
+                    `Copying ${projectAccesses.length} user access on ${previewProjectUuid}`,
+                );
+                this.logger.info(
+                    `Copying ${groupAccesses.length} group access on ${previewProjectUuid}`,
+                );
+                const insertProjectAccessPromises = projectAccesses.map(
+                    (projectAccess) =>
+                        this.projectModel.createProjectAccess(
+                            previewProjectUuid,
+                            projectAccess.email,
+                            projectAccess.role,
+                        ),
+                );
+                const insertGroupAccessPromises = groupAccesses.map(
+                    (groupAccess) =>
+                        this.groupsModel.addProjectAccess({
+                            groupUuid: groupAccess.groupUuid,
+                            projectUuid: previewProjectUuid,
+                            role: groupAccess.role,
+                        }),
+                );
+
+                await Promise.all([
+                    ...insertGroupAccessPromises,
+                    ...insertProjectAccessPromises,
+                ]);
+            },
+        );
     }
 
     async setExplores(
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -286,24 +286,47 @@ export class ProjectService extends BaseService {
         this.groupsModel = groupsModel;
     }
 
-    static hasProjectOrPreviewProjectCreateAccess(
+    static validateProjectCreationPermissions(
         user: SessionUser,
         data: CreateProject,
     ) {
+        if (!data.type) {
+            throw new ParameterError('Project type must be provided');
+        }
+
+        if (data.type === ProjectType.DEFAULT && data.upstreamProjectUuid) {
+            throw new ParameterError(
+                'upstreamProjectUuid must not be provided for default projects',
+            );
+        }
+
+        if (data.type === ProjectType.PREVIEW && !data.upstreamProjectUuid) {
+            throw new ParameterError(
+                'upstreamProjectUuid must be provided for preview projects',
+            );
+        }
+
         switch (data.type) {
             case ProjectType.DEFAULT:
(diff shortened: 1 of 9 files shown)


packages/common/src/authorization/organizationMemberAbility.test.ts
```diff
@@ -761,7 +811,18 @@ describe('Organization member permissions', () => {
         it('can use the SemanticViewer', () => {
             expect(ability.can('manage', 'SemanticViewer')).toEqual(true);
         });
+
+        it('cannot create a project in organization they belong to', () => {
```
ZeRego, line 815: [non blocker] Are we missing the test were they can create a preview in the same org?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -286,6 +286,69 @@ export class ProjectService extends BaseService {
...
+                        'create',
+                        subject('Project', {
+                            upstreamProjectUuid: data.upstreamProjectUuid,
+                            type: ProjectType.PREVIEW,
+                        }),
+                    ) ||
+                    user.ability.can(
+                        'create',
+                        subject('Project', {
+                            organizationUuid: user.organizationUuid,
```
rephus, line 334: I'm not 100% sure about this, but I think this condition is kind of wrong. 

A user who doesn't have access  to the project, can still create a preview with an upstreamProjectUuid (aka copying content) if you can create a project at the org level. 

This might still work because give our project role inheritance and our ability config , if you are a developer on the org, you are a dev on all the projects. 
But This is wrong from the permission point of view. Does that make sense ? 

An easy fix for this would be to simply check the ability inside an if like 

```
if (data.upstreamProjectUuid) {
  // check project ability
} else {
// check organization ability
}
```

packages/common/src/authorization/organizationMemberAbility.test.ts
```diff
@@ -49,6 +49,45 @@ describe('Organization member permissions', () => {
                 true,
             );
         });
+
+        it('can create a project in their own organization', () => {
```
rephus, line 53: nice tests

packages/common/src/authorization/organizationMemberAbility.test.ts
```diff
@@ -382,6 +421,16 @@ describe('Organization member permissions', () => {
             expect(ability.can('manage', 'Organization')).toEqual(false);
         });
 
+        it('cannot create a project in organization they belong to', () => {
```
rephus, line 424: should we check the type here ?

packages/common/src/authorization/organizationMemberAbility.test.ts
```diff
@@ -1638,97 +1699,4 @@ describe('Organization member permissions', () => {
             ).toEqual(false);
         });
     });
-
-    describe('test project preview permissions', () => {
```
rephus, line 1642: are these tests no longer relevant ? should we write new ones ?

packages/common/src/authorization/projectMemberAbility.test.ts
```diff
@@ -1371,19 +1371,22 @@ describe('Project member permissions', () => {
                 ability.can(
                     'create',
                     subject('Project', {
-                        projectUuid,
+                        upstreamProjectUuid: projectUuid,
```
rephus, line 1374: do we want to create DEFAULT projects with upstreamProjectUuid ?

IrakliJani:
@rephus technically, `ProjectService.copyUserAccessOnPreview` function is not necessary anymore, right?

packages/frontend/src/components/NavBar/ProjectSwitcher.tsx
```diff
@@ -240,13 +245,49 @@ const ProjectSwitcher = () => {
...
-    }, [activeProjectUuid, projects]);
+        return projects
+            .filter((p) => p.projectUuid !== activeProjectUuid)
+            .filter((project) => {
+                switch (project.type) {
+                    case ProjectType.DEFAULT:
+                        return true;
+                    // check if user has permission to create preview project on an organization level (developer, admin)
+                    // or check if user has permission to create preview project on a project level
+                    // - they should have permission (developer, admin) to the upstream project
```
ZeRego, line 269: But we don't have that information (the upstream project) here. 
We are assuming the user role was copied from the upstream project to the preview. As in, if they were developers in the upstream project they are developers in the preview so they are allowed to create preview project from the preview project.

packages/e2e/cypress/e2e/api/organizationPermissions.cy.ts
```diff
@@ -251,14 +251,16 @@ describe('Lightdash API tests for viewer org user', () => {
         });
     });
 
-    it('Should get a forbidden (403) from POST project', () => {
+    it('Should get a forbidden error (403) from POST project', () => {
```
rephus, line 254: viewers shoudl also get a 403 error for previews, right ? should we "duplicate" this test then ?

packages/common/src/authorization/projectMemberAbility.test.ts
rephus: in fact, we throw an error here https://github.com/lightdash/lightdash/pull/12158/files#diff-dba02b88a0bd4a4fdcf7ee49daaaa481189bc9ad76a683c357d0f59e568370a5R299

I guess this test is not very relevant then.

rephus (commented):
The code looks good, I'll test it now

rephus:
I can still see `create preview` from previews, not sure if render is up to date 

image

rephus:
Not related with this PR (just something I noticed I did wrong) , we can create a separate ticket if you like. 

We should say `preview project` if there are errors

image

packages/backend/src/services/ProjectService/ProjectService.ts
IrakliJani: I had it set up as you showed in your code example, but I had to change it because of exactly this:

> This might still work because give our project role inheritance and our ability config , if you are a developer on the org, you are a dev on all the projects.

There are developers who inherit permissions from the organization level. If you are a developer at the organization level and a viewer at the project level, you are still considered a developer on the project because the highest role takes precedence.

packages/frontend/src/components/NavBar/ProjectSwitcher.tsx
IrakliJani: > they are allowed to create preview project from the preview project.
goo dpoint, we do not handle this yet. I'll make the change!

IrakliJani:
> I can still see `create preview` from previews as an org developer , not sure if render is up to date



> Not related with this PR (just something I noticed I did wrong) , we can create a separate ticket if you like.
> 
> We should say `Failed to create preview project` if there are errors

pushed both of the frontend changes here

rephus (approved):

packages/common/src/authorization/projectMemberAbility.test.ts
IrakliJani: Yeah, I think it's still good to keep it, since this only tests the project member's ability without needing to know what is happening within services or organization permissions.

btw, this check returns false so that permission check also forbids the action

packages/common/src/authorization/organizationMemberAbility.test.ts
IrakliJani: improved specs including this one

packages/common/src/authorization/organizationMemberAbility.test.ts
IrakliJani: updated

packages/common/src/authorization/organizationMemberAbility.test.ts
IrakliJani: updated

packages/e2e/cypress/e2e/api/organizationPermissions.cy.ts
IrakliJani: updated

packages/e2e/cypress/e2e/api/organizationPermissions.cy.ts
```diff
@@ -210,137 +213,169 @@ describe('Lightdash API organization permission tests', () => {
...
 
-        cy.request({
-            url: endpoint,
-            headers: { 'Content-type': 'application/json' },
-            method: 'GET',
-        }).then((resp) => {
-            expect(resp.status).to.eq(200);
-        });
-    });
+                cy.request({
```
rephus, line 330: is this creating `projects` for those who are allowed to create projects ?

rephus: ahh, this is for DEFAULT and UPSTREAM, so it should faill even for admins, right ?

IrakliJani: correct!

IrakliJani:
> @rephus technically, `ProjectService.copyUserAccessOnPreview` function is not necessary anymore, right?

This was an incorrect assumption for an edge case where we have an organization-level, non-admin, non-developer user with developer permissions on a project.

We need to copy the permissions as they were before. Code has been reverted!

Issue #11198, opened 2024-08-19
Update error message stacking


jpetey75:
### Description

The warnings that appear in the query builder sometimes pile up and block the user from editing results table (like fixing the table calculation that caused the error). Ideally the warnings would not block any features, or offer a single "dismiss all" option rather than having to click each one.

Part of the problem is that often times multiple errors get thrown when a single field is removed.

### Acceptance criteria:

- [ ] When there is one error message, the behaviour should be exactly like it is today
- [ ] When there is more than one error message, they should collapse into a collapsible list. 
- [ ] There should be a way to open this list
- [ ] There should be a way to `dismiss all errors` on the collapsed version of the list
- [ ] There should be a way to open the list and dismiss the errors one-by-one
- [ ] When the list is collapsed, it should tell you how many errors there are (e.g. `errors (5)`)
- [ ] This number should decrease if you dismiss errors. 

![image](https://github.com/user-attachments/assets/d0f2bf82-46df-4490-b113-02f4216e2612)

rephus:
We could unify these errors by using the `key` argument in the toaster. Some of them (I think this one in particular) are just duplicates that are triggered multiple times.

TuringLovesDeathMetal:
@magnew Small frontend changes: 

- [ ] add margins to the top and bottom of the multiple error box
- [ ] collapse the errors by default if there are more than 1
- [ ] add scrolling within the error messages list (if there are multiple)
- [ ] keep the "collapse" and title pinned at the top

PR #12162, opened 2024-10-29
feat: stacking error list

magnew:
Closes: #11198 

### Description:

Makes it possible to add errors to a list instead of each in it's own toast. That list is collapsable and the errors can be dismissed individually. 

Note that there is a new method `addToastError` that will add to the list and I kept the old `showToastError` for now. This way we can keep the existing behavior in some places, but use error lists in others. For now I have only updated a few places where multiple errors were most problematic to use the error list, primarily the API error reporting. 

The collapsed errors now look like:
Screenshot 2024-10-30 at 09 58 29

And the expanded list looks like:
Screenshot 2024-10-30 at 09 58 44

 One consistent way to test this is to add a table calculation with invalid columns or no sql, then do it again. The first toast will be the same as always, the second one will add to the list.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
@@ -111,7 +111,10 @@ const TableCalculationModal: FC<Props> = ({
     const handleSubmit = form.onSubmit((data) => {
         const { name, sql } = data;
         if (sql.length === 0)
-            return addToastError({ title: 'SQL cannot be empty' });
+            return addToastError({
+                title: 'SQL cannot be empty',
+                key: 'table-calculation-modal',
+            });
 
         try {
             onSave({
@@ -125,6 +128,7 @@ const TableCalculationModal: FC<Props> = ({
             addToastError({
                 title: 'Error saving',
                 subtitle: e.message,
+                key: 'table-calculation-modal',
             });
         }
     });
--- a/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/TableCalculationModal.tsx
@@ -52,7 +52,7 @@ const TableCalculationModal: FC<Props> = ({
     const theme = useMantineTheme();
     const [isFullscreen, toggleFullscreen] = useToggle(false);
 
-    const { showToastError } = useToaster();
+    const { addToastError } = useToaster();
 
     const tableCalculations = useExplorerContext(
         (context) =>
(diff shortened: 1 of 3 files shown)


rephus (commented):
Nice, this works great, I tested a few errors and it is working as expected. Just a small UI suggestion, could we get rid of the weird scrollbar when there are exactly 3 items ? it seems uneccesary, perhaps by increasing the size of the panel a few pixels. 
[Screencast from 30-10-24 10:58:15.webm](https://github.com/user-attachments/assets/f3bd3d5f-c79e-4d33-a6db-b4bdae192eee)


My suggestion was to reduce the amount of duplicated errors by reusing the `key` in the tooptip, I believe this is what we do in sql runner. But this is a useful approach. 


[Screencast from 30-10-24 11:00:45.webm](https://github.com/user-attachments/assets/f57f1029-a7da-40bf-8137-bfedbb95250e)

magnew:
> Nice, this works great, I tested a few errors and it is working as expected. Just a small UI suggestion, could we get rid of the weird scrollbar when there are exactly 3 items ? it seems uneccesary, perhaps by increasing the size of the panel a few pixels.
> 
>  Screencast.from.30-10-24.10.58.15.webm 
> My suggestion was to reduce the amount of duplicated errors by reusing the `key` in the tooptip, I believe this is what we do in sql runner. But this is a useful approach.
> 
>  Screencast.from.30-10-24.11.00.45.webm

Thanks for looking! Yea, let me see if I can make it better in the 3-items case

rephus (approved):

TuringLovesDeathMetal:
niiiiice! Looks great 😍 agreed with Javi - two tiny things: 

1. the scroll when there are three items feels a bit funny

image

2. when there are two items, I wish it resized to have a fixed margin (instead of having a bunch of empty space)

image


If this is going to take more than like an hour to figure out though, then I think we can skip them and open them as a separate issue and go ahead and merge this! ✅

magnew:
> niiiiice! Looks great 😍 agreed with Javi - two tiny things:
> 
> 1. the scroll when there are three items feels a bit funny
> 
> image
> 2. when there are two items, I wish it resized to have a fixed margin (instead of having a bunch of empty space)
> 
> image
> If this is going to take more than like an hour to figure out though, then I think we can skip them and open them as a separate issue and go ahead and merge this! ✅

I fixed part of this. Getting the thing to grow vertically, then scroll has been a bit problematic, but I'll spend a few more minutes. I agree it looks dumb.

magnew:
> > niiiiice! Looks great 😍 agreed with Javi - two tiny things:
> > 
> > 1. the scroll when there are three items feels a bit funny
> > 
> > image
> > 2. when there are two items, I wish it resized to have a fixed margin (instead of having a bunch of empty space)
> > image
> > If this is going to take more than like an hour to figure out though, then I think we can skip them and open them as a separate issue and go ahead and merge this! ✅
> 
> I fixed part of this. Getting the thing to grow vertically, then scroll has been a bit problematic, but I'll spend a few more minutes. I agree it looks dumb.

Fixed!

magnew:
Render got messed up on this one and I had issues creating new ones. SO, I ran the e2e tests locally and I'm going to merge this.

PR #12172, opened 2024-10-30
feat: add sorting to dataCatalog/metrics endpoint

almeidabbm:
Closes: [#12168](https://github.com/lightdash/lightdash/issues/12168)

### Description:

New query params added:
- `sort`: name of the column to sort by
- `order`: either `asc`, `desc` or `undefined`

### Reviewer actions

### DIFF
--- a/packages/common/src/types/api/sort.ts
+++ b/packages/common/src/types/api/sort.ts
@@ -1,0 +1,2 @@
+export type ApiSortDirection = 'asc' | 'desc';
+export type ApiSort = { sort: string; order?: ApiSortDirection };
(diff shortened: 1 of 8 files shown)


packages/common/src/types/api/sort.ts
```diff
@@ -0,0 +1,2 @@
+export type ApiSortDirection = 'asc' | 'desc';
+export type ApiSort = { sort: string; order?: ApiSortDirection };
```
joaoviana, line 2: we should probably add the possible strings `sort` can be for now? like the columns of the catalog or `'name'` for now? wdyt?

packages/common/src/types/api/sort.ts
```diff
@@ -0,0 +1,2 @@
+export type ApiSortDirection = 'asc' | 'desc';
```
joaoviana, line 1: I think here we should reference the SORTBY enum we have already

joaoviana (approved):
works great! Did a little test with the ui

image

PR #12174, opened 2024-10-30
fix: add group members in a single insert

ZeRego:
Closes: 

### Description:

Remove loop that adds members on group creation. Prevent issues with database pool max out.
Should have no impact on existing behaviour.

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/GroupsModel.ts
+++ b/packages/backend/src/models/GroupsModel.ts
@@ -318,7 +318,7 @@ export class GroupsModel {
                 group.organization_id,
             )
             .whereIn('user_uuid', memberUuids)
-            .select('user_id', 'user_uuid');
+            .select('users.user_id', 'user_uuid');
 
         if (users.length === 0) {
             return [];
--- a/packages/backend/src/models/GroupsModel.ts
+++ b/packages/backend/src/models/GroupsModel.ts
@@ -41,6 +41,63 @@ export class GroupsModel {
         this.database = args.database;
     }
 
+    static async addGroupMembers(
+        trx: Knex,
+        groupUuid: string,
+        memberUuids: string[],
+    ): Promise<GroupMembership[]> {
+        // Check if the group exists
+        const group = await trx('groups')
+            .where('group_uuid', groupUuid)
+            .first('group_uuid', 'organization_id');
+        if (group === undefined) {
+            throw new NotFoundError(`No group found`);
+        }
+
+        // Check what members exist and are part of the organization
+        const users = await trx('users')
+            .innerJoin(
+                'organization_memberships',
+                'users.user_id',
+                'organization_memberships.user_id',
+            )
+            .where(
+                'organization_memberships.organization_id',
+                group.organization_id,
+            )
+            .whereIn('user_uuid', memberUuids)
+            .select('users.user_id', 'user_uuid');
+
+        if (users.length === 0) {
+            return [];
+        }
+
+        const membershipsToAdd = users.map((user) => ({
+            group_uuid: groupUuid,
+            user_id: user.user_id,
+            organization_id: group.organization_id,
+        }));
+
+        // Add members to the group
+        const membershipsAdded = await trx('group_memberships')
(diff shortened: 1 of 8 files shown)


packages/backend/src/models/GroupsModel.ts
```diff
@@ -194,31 +197,81 @@ export class GroupsModel {
         await this.database('groups').where('group_uuid', groupUuid).del();
     }
 
-    async addGroupMember(
+    async maybeAddGroupMember(
```
ZeRego, line 200: Rename so it's clear it might fail but won't throw an error.

packages/backend/src/models/GroupsModel.ts
```diff
@@ -194,31 +197,81 @@ export class GroupsModel {
...
+            return members[0];
+        } catch (e) {
+            if (e instanceof NotFoundError) {
+                return undefined;
+            }
+            throw e;
+        }
+    }
+
+    async addGroupMembers(
```
ZeRego, line 221: Add method to add multiple members in a single insert.

packages/backend/src/models/GroupsModel.ts
```diff
@@ -270,31 +273,81 @@ export class GroupsModel {
...
+        groupUuid: string,
+        memberUuids: string[],
+    ): Promise<GroupMembership[]> {
+        // Check if the group exists
+        const group = await this.database('groups')
```
rephus, line 302: we could have all these queries inside a single transaction, it will not have much impact on this particular query (apart from race conditions when updating groups/users at the same time), but overall I think is a nice thing to have. 

Not a blocker.

packages/backend/src/services/OrganizationService/OrganizationService.ts
```diff
@@ -495,27 +495,11 @@ export class OrganizationService extends BaseService {
...
+        const groupWithMembers = await this.groupsModel.createGroup({
             organizationUuid: actor.organizationUuid,
             ...createGroup,
         });
 
-        if (createGroup.members === undefined) {
-            return group;
-        }
-
-        await Promise.all(
```
rephus, line 507: :+1:

packages/frontend/src/hooks/useOrganizationGroups.ts
```diff
@@ -110,7 +110,7 @@ export const useInfiniteOrganizationGroups = (
 };
 
 const createGroupQuery = async (data: CreateGroup) =>
-    lightdashApi<Group>({
+    lightdashApi<GroupWithMembers>({
```
rephus, line 113: I assume nothing else needs to change on the frontned to handle this type change. Perhaps the response was not even used

packages/backend/src/models/GroupsModel.ts
rephus: should we add a deprecated tag ? or maybe replace its usage with the new method and remove this one.

rephus (approved):
works as expected :+1: 

image

packages/backend/src/models/GroupsModel.ts
```diff
@@ -155,30 +155,33 @@ export class GroupsModel {
...
-            uuid: row.group_uuid,
-            name: group.name,
-            createdAt: row.created_at,
-            organizationUuid: group.organizationUuid,
-        };
+
+        if (createGroup.members && createGroup.members.length > 0) {
+            await this.addGroupMembers(
+                row.group_uuid,
+                createGroup.members.map((m) => m.userUuid),
```
almeidabbm, line 180: can't we just use `createGroup.memberUuids` here?

ZeRego: `CreateGroup` argument type doesn't have that property.

packages/frontend/src/hooks/useOrganizationGroups.ts
ZeRego: Correct, I could find no usage in the frontend.

Issue #12156, opened 2024-10-29
If I'm already in the SQL runner, and I click on `new query from SQL runner`, a new tab should open with a new SQL runner


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/2afb9746938f424faa451be4551814c0?from_recorder=1&focus_title=1

Right now, if I'm in the SQL runner and I want to quickly build another SQL runner query, I click `new` --> `query using sql runer` and it just keeps the window open that I already have. 

Instead, if I'm in the SQL runner, and I click `new` --> `query using sql runner`, a new tab should open with the SQL runner opened up.

PR #12180, opened 2024-10-30
feat: open sql runner in a new tab if already on sql runner

magnew:
Closes: #12156 12156

### Description:

If already on the SQL runner page, open New > Query using SQL runner in a new browser tab.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/NavBar/ExploreMenu.tsx
+++ b/packages/frontend/src/components/NavBar/ExploreMenu.tsx
@@ -119,8 +119,9 @@ const ExploreMenu: FC<Props> = memo(({ projectUuid }) => {
                                 to={`/projects/${projectUuid}/sql-runner`}
                                 onClick={(event) => {
                                     if (
-                                        history.location.pathname ===
-                                        `/projects/${projectUuid}/sql-runner`
+                                        history.location.pathname.startsWith(
+                                            `/projects/${projectUuid}/sql-runner`,
+                                        )
                                     ) {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/NavBar/ExploreMenu.tsx
```diff
@@ -117,6 +117,18 @@ const ExploreMenu: FC<Props> = memo(({ projectUuid }) => {
                                 title="Query using SQL runner"
                                 description="Access your database to run ad-hoc queries."
                                 to={`/projects/${projectUuid}/sql-runner`}
+                                onClick={(event) => {
+                                    if (
+                                        history.location.pathname ===
+                                        `/projects/${projectUuid}/sql-runner`
```
rephus, line 123: if I'm editing an existing chart, my pathname is different (https://lightdash-pr-12180.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/sql-runner/saved-sql-chart/edit) I think we should also open these URLs in a different path. 

Perhaps we can see if pathname startsWith `/projects/${projectUuid}/sql-runner`

apart from that, it works well.

rephus (changes requested):

packages/frontend/src/components/NavBar/ExploreMenu.tsx
magnew: Good point! Updated

rephus (approved):

Issue #12175, opened 2024-10-30
Store a field's chart usage in catalog 


almeidabbm:
- add counter of `chart_usage` for each field - metric and dimension
- update counter on chart creation/update/deletion

PR #12182, opened 2024-10-30
feat: get a field's chart usage and store in catalog

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/12175

### Description:

- Adds `chart_usage` column to `catalog_search`
- After the catalog is indexed, a new task will be scheduled to calculate the chart usage per each field in the catalog with the charts where this field is used
- Then, the column `chart_usage` will be updated with the new count

> [!NOTE]
> A separate PR will be created to update this counter on chart creation/update/deletion. 

### Reviewer actions

test-frontend

### DIFF
--- a/packages/backend/src/models/CatalogModel/utils/index.ts
+++ b/packages/backend/src/models/CatalogModel/utils/index.ts
@@ -13,6 +13,7 @@ export const convertExploresToCatalog = (
     catalogInserts: DbCatalogIn[];
     catalogFieldMap: CatalogFieldMap;
 } => {
+    // Track fields' ids and names to calculate their chart usage
     const catalogFieldMap: CatalogFieldMap = {};
 
     const catalogInserts = cachedExplores.reduce<DbCatalogIn[]>(
--- a/packages/backend/src/models/CatalogModel/utils/index.ts
+++ b/packages/backend/src/models/CatalogModel/utils/index.ts
@@ -1,38 +1,68 @@
-import { CatalogType, Explore } from '@lightdash/common';
+import {
+    CatalogType,
+    Explore,
+    getItemId,
+    type CatalogFieldMap,
+} from '@lightdash/common';
 import { DbCatalogIn } from '../../../database/entities/catalog';
 
 export const convertExploresToCatalog = (
     projectUuid: string,
     cachedExplores: (Explore & { cachedExploreUuid: string })[],
-): DbCatalogIn[] =>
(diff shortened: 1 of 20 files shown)


packages/backend/src/models/CatalogModel/utils/index.ts
```diff
@@ -1,38 +1,68 @@
...
-            name: explore.name,
-            description: baseTable?.description || null,
-            type: CatalogType.Table,
-            required_attributes: baseTable.requiredAttributes ?? {}, // ! Initializing as {} so it is not NULL in the database which means it can't be accessed
-        };
+): {
+    catalogInserts: DbCatalogIn[];
+    catalogFieldMap: CatalogFieldMap;
+} => {
+    const catalogFieldMap: CatalogFieldMap = {};
```
joaoviana, line 16: I would add a comment here
```
// Track fields' ids and names to calculate their chart usage 
```

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -619,6 +620,45 @@ export class SavedChartModel {
         return savedChart;
     }
 
+    async getChartWithFieldSummaries(
```
joaoviana, line 623: can we rename to `getChartUsageByField`?

packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -535,14 +518,41 @@ export class CatalogService<
...
+                        name: metric.name,
+                        table: metric.tableName,
+                    }),
                 ),
             }));
 
         return {
             pagination,
-            data: await Promise.all(analyticsPromises),
+            data: await Promise.all(analyticsPromises), // ! This is very bad practive, temporary until we have `chart_usage` count
```
joaoviana, line 530: do you think we can remove this?

joaoviana: And wait until we add differently ? I want to basically check how the scheduled task does in analytics for example

packages/backend/src/scheduler/SchedulerClient.ts
```diff
@@ -710,6 +712,31 @@ export class SchedulerClient {
         return { jobId };
     }
 
+    async updateCatalogChartUsages(
```
joaoviana, line 715: I would document this util 
```
// Updates catalog with chart usages after the catalog is indexed - for example, metric_1 is used by 2 charts, so its chart_usage will be 2
```

joaoviana (approved):

Issue #12201, opened 2024-10-31
clicking `view underlying data` on a date value that's null causes an error


TuringLovesDeathMetal:
### Description

https://www.loom.com/share/b953b611092542b2bc1aa0bb57b0f391

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/b953b611092542b2bc1aa0bb57b0f391

### Cloud or self-hosting

None

PR #12219, opened 2024-11-01
fix: view underlying data with a null date

magnew:
Closes: #12201 

### Description:

This fixes an issue where clicking 'view underlying data' with a null date caused an error in the dialog (see video). The fix is small and a seems a little unrelated from the bug, but it makes our date handling more correct. The fix is to return null when a date field is null; before were were trying to parse null as a date and returning 'Invalid date' as a date. The underlying data dialog was then trying to use 'Invalid date' as a filter value, causing the error. 

This fix will change the way null dates are returned from runQuery. I don't know if there are other places that were relying on the old (broken) behavior. This is definitely more correct though. 

To test:
- use the orders model and choose oder date day as one of the fields
- filter such that orders date day isNull (in this branch there will be a null value)
- view underlying data
- this throws an error in main, it should be fixed in this branch

Repro WITH the bug (main)
Kapture 2024-11-01 at 14 46 30

Behavior with this fix
Kapture 2024-11-01 at 14 44 41

test-frontend

### Reviewer actions

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/data/raw_orders.csv
+++ b/examples/full-jaffle-shop-demo/dbt/data/raw_orders.csv
@@ -89,7 +89,6 @@ id,user_id,order_date,status
 88,91,2018-03-27,shipped
 89,21,2018-03-28,placed
 90,66,2018-03-30,shipped
-100,85,null,placed
 91,47,2018-03-31,placed
 92,84,2018-04-02,placed
 93,66,2018-04-03,placed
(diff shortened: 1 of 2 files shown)


examples/full-jaffle-shop-demo/dbt/data/raw_orders.csv
```diff
@@ -98,3 +98,4 @@ id,user_id,order_date,status
 97,89,2018-04-07,placed
 98,41,2018-04-07,placed
 99,85,2018-04-09,placed
+100,85,null,placed
```
magnew, line 101: Added for testing. I don't know if this will break some other test.

ZeRego (approved):

examples/full-jaffle-shop-demo/dbt/data/raw_orders.csv
ZeRego: Running e2e tests to make sure is doesn't

Issue #12044, opened 2024-10-23
Developers should be able to delete preview projects that they've created


TuringLovesDeathMetal:
### Description

Today, only admins can delete projects. But, now that we're adding the option to create preview projects in the UI, we need to have a way for developers to delete their preview projects they've created. 

Can project developers even see the projects list page in the org settings? (if they're not an org developer)? 

If yes, then we should just add the `delete` option to preview projects that they've created.

PR #12229, opened 2024-11-02
feat: developers should be able to delete previews 

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/12044

test-backend
test-frontend
test-cli

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -929,13 +929,19 @@ export class ProjectService extends BaseService {
     }
 
     async delete(projectUuid: string, user: SessionUser): Promise<void> {
-        const { organizationUuid, type } =
-            await this.projectModel.getWithSensitiveFields(projectUuid);
+        const project = await this.projectModel.getWithSensitiveFields(
+            projectUuid,
+        );
 
         if (
             user.ability.cannot(
                 'delete',
-                subject('Project', { organizationUuid, projectUuid }),
+                subject('Project', {
+                    type: project.type,
+                    projectUui: project.projectUuid,
+                    organizationUuid: project.organizationUuid,
+                    createdByUserUuid: project.createdByUserUuid,
+                }),
             )
         ) {
             throw new ForbiddenError();
@@ -948,7 +954,7 @@ export class ProjectService extends BaseService {
             userId: user.userUuid,
             properties: {
                 projectId: projectUuid,
-                isPreview: type === ProjectType.PREVIEW,
+                isPreview: project.type === ProjectType.PREVIEW,
             },
         });
     }
@@ -4016,6 +4022,7 @@ export class ProjectService extends BaseService {
             dbtConnection: project.dbtConnection,
             upstreamProjectUuid: data.copyContent ? projectUuid : undefined,
             dbtVersion: project.dbtVersion,
+            createdByUserUuid: user.userUuid,
         };
 
         const previewProject = await this.createWithoutCompile(
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -938,7 +938,7 @@ export class ProjectService extends BaseService {
                 'delete',
                 subject('Project', {
                     type: project.type,
-                    projectUui: project.projectUuid,
+                    projectUuid: project.projectUuid,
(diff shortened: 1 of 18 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -927,13 +929,19 @@ export class ProjectService extends BaseService {
...
+            projectUuid,
+        );
 
         if (
             user.ability.cannot(
                 'delete',
-                subject('Project', { organizationUuid, projectUuid }),
+                subject('Project', {
+                    type: project.type,
+                    projectUui: project.projectUuid,
```
rephus, line 941:
```suggestion
                    projectUuid: project.projectUuid,
```

packages/frontend/src/components/UserSettings/ProjectManagementPanel/index.tsx
```diff
@@ -73,8 +77,10 @@ const ProjectListItem: FC<ProjectListItemProps> = ({
                     <Can
                         I="delete"
                         this={subject('Project', {
+                            type: project.type,
+                            projectUuid: project.projectUuid,
                             organizationUuid: user.data?.organizationUuid,
-                            projectUuid,
+                            createdByUserUuid: project.createdByUserUuid,
```
rephus, line 83: should this be 
`createdByUserUuid: user.data?.uuid` ? 
otherwise, the other check is redundant, right ?

packages/frontend/src/pages/Settings.tsx
```diff
@@ -189,124 +189,99 @@ const Settings: FC = () => {
...
+                                        exact
+                                        icon={
+                                            <MantineIcon icon={IconUserPlus} />
+                                        }
+                                    />
+                                )}
+                                {user.ability.can(
+                                    'manage',
+                                    subject('Organization', {
+                                        organizationUuid:
```
rephus, line 230: reundant check, this should be 
`organizationUuid: user.data?.organizationUuid` I think.

packages/frontend/src/pages/Settings.tsx
```diff
@@ -189,124 +189,99 @@ const Settings: FC = () => {
...
-                                            icon={
-                                                <MantineIcon
-                                                    icon={IconUserShield}
-                                                />
-                                            }
-                                        />
-                                    )}
+                                {user.ability.can(
+                                    'update',
+                                    'OrganizationMemberProfile',
```
rephus, line 212: These abilities are not checking organizationUuids, which is fine, for now, since we only allow 1 org per user

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: These typos are very annoying, if you don't type it corrently, then the ability check is useless. We should enforce this typing somehow.

rephus (commented):
only 1 change needed

rephus (approved):
I've approved so you can merge whenever this is ready

packages/frontend/src/pages/Settings.tsx
IrakliJani: I didn’t touch this chunk (the indentation diff might make it seem otherwise), but I believe it's correct because this is the permission definition in CASL:

```ts
can('manage', 'Organization', {
    organizationUuid: member.organizationUuid,
});
```

To match that, it should be `organization.organizationUuid`

or am I wrong?

packages/frontend/src/pages/Settings.tsx
IrakliJani: same as above, not in the scope of the change. but I agree!

packages/frontend/src/components/UserSettings/ProjectManagementPanel/index.tsx
IrakliJani:
```ts
        can('delete', 'Project', {
            type: ProjectType.PREVIEW,
            createdByUserUuid: member.userUuid,
        });
```

this is the permission definition so we have to match `project.createdByUserUuid` to the member's uuid

IrakliJani: we have similar for `CsvJobResult`

```ts
        can('view', 'CsvJobResult', {
            createdByUserUuid: member.userUuid,
        });
```

Issue #6879, opened 2023-08-24
Document user API


ZeRego:
### Description

Refactor `packages/backend/src/routers/userRouter.ts` to /controllers

PR #12235, opened 2024-11-04
feat: change user 'isActive' definition

ZeRego:
Related to: https://github.com/lightdash/lightdash-commercial/issues/324

### Description:

We want user accounts marked with isActive = false to not have access to Lightdash. 

At the moment, is_active is being used to track what users were invited but haven't logged in yet aka don't have an authentication method.

The variable/value is used in these places:

https://github.com/lightdash/lightdash/blob/547ad8083fc47a848b33110ea6a09f440c7b160f/packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx#L65

https://github.com/lightdash/lightdash/blob/1dd45b035b9485c6f646f68a4a4a5520eca15010/packages/frontend/src/features/comments/components/CommentForm.tsx#L51

https://github.com/lightdash/lightdash/blob/fa0c44aff40c8a86a5dd1ae0dda91a58be09c8a1/packages/backend/src/services/UserService.ts#L369

https://github.com/lightdash/lightdash/blob/f87a352b7f19b1af9fce1fa414bb84e49d6de185/packages/backend/src/models/OrganizationMemberProfileModel.ts#L86

First, we need to adjust our logic to not rely on isActive to determine if a user is pending (aka hasn't accepted the invite ).
We would have to update the logic to check if the user already has an authentication method.

What I changed:
- add migration so `is_active` is `true` for all users
- update models to fetch if the user has authentication methods and based on that determine their `isPending` status.
- update logic that depending on `isActive` to now use `isPending` 
- misc of fixes (check PR comments)

(note that you can only set a user to be inactive by changing the DB)
Screenshot 2024-11-04 at 13 21 29

### Test:
✅ authenticate with user
✅ invite user
✅ accept invite 
✅ tested migration rollback and forward with pending users

## Next PRs

In the second PR, we can update the code to block non active users from using Lightdash.

In the third PR, have a way for admins to activate/deactivate users in the UI.

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/OrganizationMemberProfileModel.ts
+++ b/packages/backend/src/models/OrganizationMemberProfileModel.ts
@@ -162,14 +162,19 @@ export class OrganizationMemberProfileModel {
                 userUuids: data.map((m) => m.user_uuid),
             });
 
+        const usersHaveAuthenticationMap = new Map(
+            usersHaveAuthenticationRows.map((row) => [
+                row.user_uuid,
+                row.has_authentication,
+            ]),
+        );
+
         return {
             pagination,
             data: data.map((m) =>
                 OrganizationMemberProfileModel.parseRow(
                     m,
-                    usersHaveAuthenticationRows.find(
-                        (row) => row.user_uuid === m.user_uuid,
-                    )?.has_authentication,
+                    usersHaveAuthenticationMap.get(m.user_uuid) || false,
                 ),
             ),
         };
@@ -285,15 +290,19 @@ export class OrganizationMemberProfileModel {
             await UserModel.findIfUsersHaveAuthentication(this.database, {
                 userUuids: updatedMembers.map((m) => m.user_uuid),
             });
+        const usersHaveAuthenticationMap = new Map(
+            usersHaveAuthenticationRows.map((row) => [
+                row.user_uuid,
+                row.has_authentication,
+            ]),
+        );
 
         return {
             pagination,
             data: updatedMembers.map((m) => ({
                 ...OrganizationMemberProfileModel.parseRow(
                     m,
-                    usersHaveAuthenticationRows.find(
-                        (row) => row.user_uuid === m.user_uuid,
-                    )?.has_authentication,
+                    usersHaveAuthenticationMap.get(m.user_uuid) || false,
                 ),
                 groups: m.groups,
             })),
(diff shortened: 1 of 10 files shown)


packages/frontend/src/features/comments/components/CommentForm.tsx
```diff
@@ -48,7 +48,7 @@ export const CommentForm: FC<Props> = ({
     const userNames: SuggestionsItem[] | undefined = useMemo(() => {
         if (!listUsers || !space?.access) return undefined;
         return listUsers.reduce<SuggestionsItem[]>((acc, user) => {
-            if (!user.isActive) return acc;
+            if (user.isPending) return acc;
```
ZeRego, line 51: Don't show pending users in dashboard comments suggestions.

packages/frontend/src/components/UserSettings/UsersAndGroupsPanel/UsersView.tsx
```diff
@@ -62,28 +62,27 @@ const UserNameDisplay: FC<{
 }> = ({ user, showInviteLink, hasEmail, onGetLink }) => {
     return (
         <Flex justify="space-between" align="center">
-            {user.isActive ? (
-                <Stack spacing="xxs">
-                    <Title order={6}>
-                        {user.firstName} {user.lastName}
+            {!user.isActive ? (
```
ZeRego, line 65: Flip logic to check first if user is inactive before checking pending status.
+ show email if they don't have names.

packages/backend/src/services/UserService.ts
```diff
@@ -1014,7 +1014,13 @@ export class UserService extends BaseService {
         const updatedUser = await this.userModel.updateUser(
             user.userUuid,
             user.email,
-            data,
+            {
```
ZeRego, line 1017: User routes are not using TSOA to validate body properties so I'm made sure it doesn't update properties it shouldn't. Related: https://github.com/lightdash/lightdash/issues/6879

packages/backend/src/models/UserModel.ts
```diff
@@ -246,7 +284,11 @@ export class UserModel {
         if (user === undefined) {
             throw new NotFoundError(`Cannot find user with uuid ${userUuid}`);
         }
-        return mapDbUserDetailsToLightdashUser(user);
+
+        return mapDbUserDetailsToLightdashUser(
+            user,
+            await this.hasAuthentication(userUuid),
```
ZeRego, line 290: A bigger refactor is needed to avoid calling the same static/private functions in most methods.

packages/backend/src/models/OrganizationMemberProfileModel.ts
```diff
@@ -148,9 +157,21 @@ export class OrganizationMemberProfileModel {
...
+                userUuids: data.map((m) => m.user_uuid),
+            });
+
         return {
             pagination,
-            data: data.map(OrganizationMemberProfileModel.parseRow),
+            data: data.map((m) =>
+                OrganizationMemberProfileModel.parseRow(
+                    m,
+                    usersHaveAuthenticationRows.find(
```
andrewdw, line 170: Not optimal to have a loop within a loop, could this be broken out to something like:

```
const usersHaveAuthenticationMap = new Map(
        usersHaveAuthenticationRows.map(row => [row.user_uuid, row.has_authentication])
    );

    return {
        pagination,
        data: data.map((m) =>
            OrganizationMemberProfileModel.parseRow(
                m,
                usersHaveAuthenticationMap.get(m.user_uuid) || false,
            ),
        ),
    };
```

ZeRego: Nice one :) refactored the code with that 👍

andrewdw (approved):

PR #12242, opened 2024-11-04
feat: block deactivated users from accessing Lightdash

ZeRego:
Related to: https://github.com/lightdash/lightdash-commercial/issues/324

### Description:

We want user accounts marked with isActive = false to not have access to Lightdash. The experience should be:

- if they aren’t signed in: they see an error trying to sign in
- if they are signed in: on the next action they are signed out and redirected to /login

Note: we need to change the value in the DB to test.

Fail with password

https://github.com/user-attachments/assets/e54afa0f-9007-42e3-8db2-851f23beb1d7

Fail with oauth

https://github.com/user-attachments/assets/767316e1-3725-40c8-b4f4-4b30b6166af5

Session destroyed & redirect to /login page ( the error message might show for a brief moment depending on the error handling in the page. Depriotized this as it is an edge case and haven't found of a solution that covers all places )

https://github.com/user-attachments/assets/ab6f57f4-80dc-4d9b-833f-c2921bb600aa

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SchedulerModel/index.ts
+++ b/packages/backend/src/models/SchedulerModel/index.ts
@@ -140,7 +140,15 @@ export class SchedulerModel {
     }
 
     async getAllSchedulers(): Promise<SchedulerAndTargets[]> {
-        const schedulers = this.database(SchedulerTableName).select();
+        const schedulers = this.database(SchedulerTableName)
+            .select()
+            .join(
+                UserTableName,
+                `${UserTableName}.user_uuid`,
+                `${SchedulerTableName}.created_by`,
+            )
+            .where(`${SchedulerTableName}.enabled`, true)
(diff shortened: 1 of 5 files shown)


packages/backend/src/models/SchedulerModel/index.ts
```diff
@@ -140,7 +140,15 @@ export class SchedulerModel {
     }
 
     async getAllSchedulers(): Promise<SchedulerAndTargets[]> {
-        const schedulers = this.database(SchedulerTableName).select();
+        const schedulers = this.database(SchedulerTableName)
```
ZeRego, line 143: Avoid running scheduled deliveries from deactivated users. 
We should inform admins about this when deactivating users.

andrewdw (approved):

Issue #12260, opened 2024-11-05
Analytics Events


PriPatel:
- [ ] User creates tag (server with tag name) 
- [ ] User tags metrics (ui click) 
- [ ] User filters by tag (ui search)

PR #12295, opened 2024-11-06
chore: add catalog-category-related events

joaoviana:
Closes: #12260

### Description:

> [!NOTE]
> Added as well project id and org id to all metrics-catalog-related events

 ### User creates tag (server with tag name)

image

 ### User tags metrics (ui click)

image

 ### User filters by tag (ui search)

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/MetricsCatalogPanel.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/MetricsCatalogPanel.tsx
@@ -1,15 +1,27 @@
 import { Group, Stack, Title } from '@mantine/core';
+import { useEffect } from 'react';
 import { useParams } from 'react-router-dom';
 import { useMount } from 'react-use';
 import RefreshDbtButton from '../../../components/RefreshDbtButton';
+import { useProject } from '../../../hooks/useProject';
 import { useAppDispatch, useAppSelector } from '../../sqlRunner/store/hooks';
-import { setActiveMetric, setProjectUuid } from '../store/metricsCatalogSlice';
+import {
+    setActiveMetric,
+    setOrganizationUuid,
+    setProjectUuid,
+} from '../store/metricsCatalogSlice';
 import { MetricChartUsageModal } from './MetricChartUsageModal';
 import { MetricsTable } from './MetricsTable';
 
 export const MetricsCatalogPanel = () => {
-    const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
+    const projectUuid = useAppSelector(
+        (state) => state.metricsCatalog.projectUuid,
+    );
+    const organizationUuid = useAppSelector(
+        (state) => state.metricsCatalog.organizationUuid,
+    );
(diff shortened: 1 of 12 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricsCatalogPanel.tsx
```diff
@@ -1,15 +1,27 @@
...
+import {
+    setActiveMetric,
+    setOrganizationUuid,
+    setProjectUuid,
+} from '../store/metricsCatalogSlice';
 import { MetricChartUsageModal } from './MetricChartUsageModal';
 import { MetricsTable } from './MetricsTable';
 
 export const MetricsCatalogPanel = () => {
-    const projectUuid = useAppSelector((state) => state.sqlRunner.projectUuid);
```
joaoviana, line 11: oops

packages/frontend/src/types/Events.ts
```diff
@@ -113,7 +113,12 @@ export enum EventName {
...
     NOTIFICATIONS_COMMENTS_ITEM_CLICKED = 'notifications_comments_item.clicked',
     DASHBOARD_AUTO_REFRESH_UPDATED = 'dashboard_auto_refresh.updated',
+
+    // Metrics Catalog
     METRICS_CATALOG_CHART_USAGE_CLICKED = 'metrics_catalog_chart_usage.clicked',
     METRICS_CATALOG_CHART_USAGE_CHART_CLICKED = 'metrics_catalog_chart_usage_chart.clicked',
     METRICS_CATALOG_EXPLORE_CLICKED = 'metrics_catalog_explore.clicked',
+    // TODO: Rename from tag to category
+    METRICS_CATALOG_TAG_ADDED = 'metrics_catalog_category_added',
+    METRICS_CATALOG_TAG_FILTER_APPLIED = 'metrics_catalog_category_filter_applied',
```
almeidabbm, line 123: Shouldn't these have `.` instead of `_` to separate the `metrics_catalog_category` action?

```suggestion
    METRICS_CATALOG_TAG_ADDED = 'metrics_catalog_category.added',
    METRICS_CATALOG_TAG_FILTER_APPLIED = 'metrics_catalog_category.filter_applied',
```

almeidabbm (approved):

packages/frontend/src/types/Events.ts
almeidabbm: + I think we should perhaps also track when a user removes a category and when a tag is deleted from the project

joaoviana: > I think we should perhaps also track when a user removes a category and when a tag is deleted from the project

good point! After a discussion with @PriPatel it seems that this is not an immediate need 👍

joaoviana: https://github.com/lightdash/lightdash/pull/12295/commits/1955830a27ab9e0bb00405f1d56534c60553b448

renamed here!

PR #12314, opened 2024-11-07
feat: added exact match function to search utils

andrewdw:
### Description:
- This adds a function that allows search to search by exact instead of a fuzzy search
- This will be used by SCIM `eq` filter when looking for specific users

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SearchModel/utils/search.ts
+++ b/packages/backend/src/models/SearchModel/utils/search.ts
@@ -60,6 +60,18 @@ export function getColumnMatchRegexQuery(
     });
 }
 
+export function getColumnMatchExactQuery(
+    queryBuilder: Knex.QueryBuilder,
+    searchQuery: string,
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/SearchModel/utils/search.ts
```diff
@@ -60,6 +60,18 @@ export function getColumnMatchRegexQuery(
     });
 }
 
+export function getColumnMatchExactQuery(
```
ZeRego, line 63: I don't think we need this type of abstraction for exact match. 
It is useful for search terms, so we can search across name/description, etc
For the exact match we know the column and the name so I think we just need to update the `GroupModel` `find` to have more filter options like organizationUuid. 

```
        if (filters.name) {
            query = query.where(
                'groups.name',
                filters.name,
            );
        }
```

PR #12330, opened 2024-11-08
feat: force unique group names at the db level

andrewdw:
Closes: https://github.com/lightdash/lightdash-commercial/issues/362

### Description:
- Creates a unique constraint on the groups table by (name, organization_id) but first changes any identically named groups in an org to "$NAME", "$NAME 2", "$NAME...
- On group update and delete it checks for a Postgres unique violation error and throws an appropriate Lightdash error

Screenshot 2024-11-11 at 10 15 18 AM

### Reviewer actions

### DIFF
--- a/packages/backend/src/database/migrations/20241108145015_group-names-unique.ts
+++ b/packages/backend/src/database/migrations/20241108145015_group-names-unique.ts
@@ -27,7 +27,7 @@ export async function up(knex: Knex): Promise<void> {
             // eslint-disable-next-line no-await-in-loop
             await knex(GroupsTableName)
                 .where({ group_uuid: rows[i].group_uuid })
-                .update({ name: `${name} ${i}` });
+                .update({ name: `${name} ${i + 1}` });
         }
     }
 
--- a/packages/backend/src/database/migrations/20241108145015_group-names-unique.ts
+++ b/packages/backend/src/database/migrations/20241108145015_group-names-unique.ts
@@ -1,0 +1,44 @@
+import { Knex } from 'knex';
+
+const GroupsTableName = 'groups';
+
+export async function up(knex: Knex): Promise<void> {
+    // Ensure no duplicate names for organization_id by appending suffixes
+    const duplicates = await knex(GroupsTableName)
+        .select('name', 'organization_id')
+        .count('* as duplicate_count')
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/GroupsModel.ts
```diff
@@ -226,31 +228,41 @@ export class GroupsModel {
...
-            if (row === undefined) {
-                throw new UnexpectedDatabaseError(`Failed to create group`);
-            }
-
-            if (createGroup.members && createGroup.members.length > 0) {
-                await GroupsModel.addGroupMembers(
-                    trx,
-                    row.group_uuid,
-                    createGroup.members.map((m) => m.userUuid),
+                    [createGroup.name.trim(), createGroup.organizationUuid],
```
andrewdw, line 241: I noticed that we were allowing trailing whitespace in group names so I'm stripping it out here before save.

packages/backend/src/database/migrations/20241108145015_group-names-unique.ts
```diff
@@ -0,0 +1,44 @@
...
+            .where({ name, organization_id })
+            .orderBy('created_at', 'asc');
+
+        // Rename duplicates with suffixes
+        // eslint-disable-next-line no-plusplus
+        for (let i = 1; i < rows.length; i++) {
+            // eslint-disable-next-line no-await-in-loop
+            await knex(GroupsTableName)
+                .where({ group_uuid: rows[i].group_uuid })
+                .update({ name: `${name} ${i}` });
```
andrewdw, line 30: If an org has 2 groups named the same "Testing" it will migrate them "Testing 1" and "Testing 2"

ZeRego: Correction. In the case of two groups with the same name, the first one keeps the original while the rest gets a numeric suffix.
before migration
Screenshot 2024-11-11 at 10 07 44

after migration
Screenshot 2024-11-11 at 10 08 25

andrewdw: You are correct. I've modified it so that the original stays the same and the next one gets "Name 2" and so on

ZeRego (approved):

Issue #12204, opened 2024-10-31
Filters for dates/timestamps on scheduled deliveries are not overriding the dashboard filter value


TuringLovesDeathMetal:
### Description

If I have a filter in my scheduled delivery, the value I apply there is not overriding what I have in my dashboard. 

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/7d183132a6154ed781041838ee496870

### Cloud or self-hosting

None

PR #12332, opened 2024-11-08
fix: send now filters weren't working for images, some filters 

magnew:
…isabled on creation

Closes: #12204 

### Description:

The SD filters were being set as disabled. This fixes that by using the existing fitler setting for disabled.

Also, send now wasn't using the Scheduled delivery filters because of a typo in URL parsing. This was adding to confusing about whether SD filters worked at all. Also fixed here. 

To repro the issue:

Add a fitler to a dashboard for something like 'date is after...'
Set up a scheduled delivery with a filter for a different 'date is after...'
Send now. You will get data filtered by the dashboard filter.

### Reviewer actions


joaoviana (approved):

hamzahc1:
is a second reviewer needed here @magnew ?

PR #12347, opened 2024-11-11
chore: introduce shared emoji picker and metric icon

joaoviana:
Close: #12344 

### Description:

Adds `emoji-picker-react`
Creates shared emoji picker (one for all cells with `Portal`) 

DEMO: 

https://github.com/user-attachments/assets/92e49064-65be-48b4-afb8-9aac07694ba0

> [!NOTE]
> If we had a popover with the picker for each cell, that would lead to a very slow ux because the emoji picker would be mounted on every open. If we kept the popover mounted, it would lead to a nr_of_cells X pickers on the UI which is a no-no. So keeping a shared instance and moving the position as requested is an easier experience. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
@@ -12,7 +12,7 @@ import {
 import { useClickOutside } from '@mantine/hooks';
 import EmojiPicker, { EmojiStyle } from 'emoji-picker-react';
 import { type MRT_Row, type MRT_TableInstance } from 'mantine-react-table';
-import { forwardRef, useCallback, useState, type FC } from 'react';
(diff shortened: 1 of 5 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    left={position.left}
+                    sx={{
+                        zIndex: getDefaultZIndex('overlay'),
+                    }}
+                >
+                    <Paper shadow="md" pos="relative">
+                        <Text fw={500} size="sm" ta="center">
+                            Coming soon
+                        </Text>
+
```
joaoviana, line 46: added a coming soon until users can actually customise their metrics

almeidabbm (approved):
LGTM 🚀

Issue #12394, opened 2024-11-13
Smoother loading experience of metrics


PriPatel:
- [x] Increase default number fetched from 20-50
- [x] Only show spinner when loading more metrics
- [x] Do not show spinner when adding/removing tags and/or changing icons
- [x] Show spinner when searching, filtering, sorting

PR #12409, opened 2024-11-13
chore: improve loading ux and increase page size to 50

joaoviana:
Closes: #12394

### Description:

Improves loading experience
increases page size to `50`

> [!NOTE]
> on the ticket it says to show the spinner on scroll for next page, but I can't seem to be able to show it. I still think that these changes are better than what we have now.

demo:

https://github.com/user-attachments/assets/9eb68259-e5a3-4624-a467-1d3e9eb4a7a0

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/MetricsTable.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/MetricsTable.tsx
@@ -1,5 +1,6 @@
 import { type CatalogItem } from '@lightdash/common';
 import { Box, Divider, Group, Text, useMantineTheme } from '@mantine/core';
+import { useIsMutating } from '@tanstack/react-query';
 import {
     MantineReactTable,
     useMantineReactTable,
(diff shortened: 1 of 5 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricsTable.tsx
```diff
@@ -63,6 +71,25 @@ export const MetricsTable = () => {
...
+                'remove-category',
+                'update-catalog-item-icon',
+            ];
+            return Boolean(
+                mutation.options.mutationKey?.some((key) =>
+                    mutationKeys.includes(key as string),
+                ),
+            );
+        },
+    });
```
joaoviana, line 91: can be moved to a separate hook? 

wonder how cache keys can be maintained better. constants?

packages/frontend/src/features/metricsCatalog/hooks/useCatalogItemIcon.tsx
```diff
@@ -0,0 +1,96 @@
+import {
```
joaoviana, line 1: moved to separate file

packages/frontend/src/features/metricsCatalog/components/MetricsTable.tsx
```diff
@@ -95,6 +122,11 @@ export const MetricsTable = () => {
         return lastPage.pagination?.totalResults ?? 0;
     }, [data]);
 
+    const showLoadingOverlay = useMemo(
+        () => isFetching && isPreviousData && !isMutating,
```
almeidabbm, line 126: Can't we just use `isFetchingNextPage` here?

packages/frontend/src/features/metricsCatalog/components/MetricsTable.tsx
almeidabbm: I think having the keys as constants would be a very nice addition here. Don't think we would need to extract this to a separate hook for now, if we end up reusing it then I would do that

almeidabbm (approved):

packages/frontend/src/features/metricsCatalog/components/MetricsTable.tsx
joaoviana: apparently not lol...

PR #12411, opened 2024-11-13
feat: allow to rotate personal access token

ZeRego:
### Description:

Allow to rotate a personal access token.

Changes:
- migrate database to track when a token was rotated and last used
- amend personal access tokens types
  - make uuid required
  - made properties nullable instead of optional
  - delete duplicate type
- create a controller since existing routes are in legacy code
- add service and model methods
- register when token is used to authenticate a user
- display rotation, last usage and uuid in tokens page
- update docs

Tokens page
Screenshot 2024-11-13 at 16 48 16

Docs:
Screenshot 2024-11-13 at 17 14 59

### Reviewer actions


docs/sidebars.js
```diff
@@ -270,17 +270,30 @@ module.exports = {
...
-                            label: 'Project Settings',
-                        },
                         'references/usage-analytics',
                         'references/roles',
                         'references/groups',
                         'references/user-attributes',
                     ],
                 },
+                {
+                    type: 'category',
```
ZeRego, line 280: Fix document sidebar location

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
```diff
@@ -28,22 +38,87 @@ import {
...
+                        <Tooltip
+                            withinPortal
+                            position="top"
+                            maw={350}
+                            label={formatTimestamp(lastUsedAt)}
+                        >
+                            <Text>{formatDate(lastUsedAt)}</Text>
+                        </Tooltip>
+                    ) : (
+                        <Text color="gray.6">Never used</Text>
```
rephus, line 88: will this cause confusion for existing tokens ? since this lastUsedAtt is new and nullable, existing tokens will all show `never used` (until they are used again)

packages/backend/src/models/DashboardModel/PersonalAccessTokenModel.ts
```diff
@@ -39,10 +43,67 @@ export class PersonalAccessTokenModel {
...
     }
 
+    async getUserToken({
+        userUuid,
+        tokenUuid,
+    }: {
+        userUuid: string;
+        tokenUuid: string;
+    }): Promise<PersonalAccessToken> {
+        const row = await this.database('personal_access_tokens')
```
rephus, line 53: should we use `PersonalAccessTokenTableName` here ? we normally use constants on talbes, I think the only benefit is to avoid typos.

packages/backend/src/controllers/personalAccessTokensController.ts
```diff
@@ -0,0 +1,52 @@
...
+} from '@tsoa/runtime';
+import express from 'express';
+import { allowApiKeyAuthentication, isAuthenticated } from './authentication';
+import { BaseController } from './baseController';
+
+@Route('/api/v1/me/personal-access-tokens')
+@Response<ApiErrorPayload>('default', 'Error')
+@Tags('Personal Access Tokens')
+export class PersonalAccessTokensController extends BaseController {
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
```
rephus, line 25: unauthorisedInDemo ? (Requires import, so I can't suggest a change here)

packages/backend/src/services/PersonalAccessTokenService.ts
```diff
@@ -58,4 +60,42 @@ export class PersonalAccessTokenService extends BaseService {
...
+        if (data.expiresAt.getTime() < Date.now()) {
+            throw new ParameterError('Expire time must be in the future');
+        }
+
+        const existingToken = await this.personalAccessTokenModel.getUserToken({
+            userUuid: user.userUuid,
+            tokenUuid: personalAccessTokenUuid,
+        });
+
+        if (!existingToken.expiresAt) {
```
rephus, line 78: since this is a business decision, and not a security/tecnical issue, I 'd add a comment here, in case someone has a use case that forces us to remove it. So we don't need to remember why we do this in the first place in a few months.

packages/backend/src/services/PersonalAccessTokenService.ts
```diff
@@ -58,4 +60,42 @@ export class PersonalAccessTokenService extends BaseService {
...
+            throw new ParameterError('Token can only be rotated once per hour');
+        }
+
+        const newToken = await this.personalAccessTokenModel.rotate({
+            personalAccessTokenUuid,
+            expiresAt: data.expiresAt,
+        });
+        this.analytics.track({
+            userId: user.userUuid,
+            event: 'personal_access_token.rotated',
```
rephus, line 97: no properties ?

rephus (commented):
look good , just a few comments

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
ZeRego: @rephus Fair point 💡  Options:
- keep the same
- don't show any value. 
- have a condition with a hardcoded date `if was created before 14 Nov 2024 and used date is null don't show any value otherwise "Never used"` 
Any preference or other option comes to mind? I'm inclined to the second.

packages/backend/src/services/PersonalAccessTokenService.ts
ZeRego: None of the personal access tokens has properties. Didn't want to make the PR bigger.

docs/sidebars.js
ZeRego: @jpetey75 Let me know if you agree with this.

packages/frontend/src/components/UserSettings/AccessTokensPanel/TokensTable.tsx
ZeRego: Went with second option

rephus (approved):

docs/sidebars.js
jpetey75: @ZeRego maybe it can just go under `References > Lightdash Development`? I don't think there are very many items to put into a `User` category 

The other option is just to put a `User Settings` under the `Admin` category, or even at the top level without a category, then we just put all user settings there. 

I might move to `Lightdash Development` next time I make a docs change, but it's not super important IMO

Issue #12316, opened 2024-11-07
Can't invite project users to private spaces


jpetey75:
### Description

When I try to invite project users to a private space in that project, I cannot select their name to invite. 

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/c02eb4432315404bb2f1c55e7297576e

In the video, the user I am trying to invite only has access to this project, not the whole organization. Inviting org users to the space works fine. 

### Cloud or self-hosting

cloud

PR #12413, opened 2024-11-13
fix: cant invite project users to space

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/12316

### Description:

this bug happens because of react query caching + infinite pagination combination. I can discuss this further but the short story is, react query produces different cache spaces for search queries because search query is in the cache key.

https://github.com/user-attachments/assets/da43b4f2-25fa-4cf6-98cf-e9cd132e3631

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceAddUser.tsx
+++ b/packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceAddUser.tsx
@@ -1,4 +1,8 @@
-import { OrganizationMemberRole, type Space } from '@lightdash/common';
+import {
+    OrganizationMemberRole,
+    type OrganizationMemberProfile,
+    type Space,
+} from '@lightdash/common';
 import {
     Avatar,
     Badge,
@@ -16,6 +20,7 @@ import {
 } from '@mantine/core';
 import { useDebouncedValue } from '@mantine/hooks';
 import { IconUsers } from '@tabler/icons-react';
+import { uniq, uniqBy } from 'lodash';
 import {
     forwardRef,
     useEffect,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceAddUser.tsx
```diff
@@ -86,25 +94,33 @@ export const ShareSpaceAddUser: FC<ShareSpaceAddUserProps> = ({
...
 
-    const organizationUsers = useMemo(
-        () => infiniteOrganizationUsers?.pages.map((p) => p.data).flat(),
-        [infiniteOrganizationUsers?.pages],
-    );
+    useEffect(() => {
+        const allPages =
+            infiniteOrganizationUsers?.pages.map((p) => p.data).flat() ?? [];
+
+        setAllSearchedOrganizationUsers((previousState) =>
```
rephus, line 101: I'm not fully understand what this is doing, but perhaps we should at least add acomment explaining what this is fixing.

IrakliJani: @rephus Code did not take into account how React Query caches data. Since the search query is part of the react query's cache key, every time the search query changes, the results returned by the hook exclude those from the previous search query.

This causes an issue where the autocomplete can display search results, but when you click on a result, the query resets to an empty string, loading a completely new set of results. As a result, the selected value disappears, as described in the original issue ticket.

This change aggregates all values returned by the hook into a state, keeping track of all users that have ever appeared in the search results, regardless of the search query.

I’ve included the updated comment.
hope it’s clearer now!

Issue #12391, opened 2024-11-13
When I filter by a category, all the correct metrics come up, but I should still be able to see all the categories the metrics has. 


PriPatel:
This causes a lot of weird confusing cases where I tried to add a tag, then it disappears. 

- There is a metric tagged A and B. I can see tag A and B in the category column
- Search for A
- Metric comes up but only displays tag A in the column
- I try to add a tag but it 'disappears'

**Expected behaviour**
- [ ] You can filter by a tag and the metrics should be filters by that tag, but I should still be able to see every one of that metrics' categories.

PR #12426, opened 2024-11-14
fix: get all tags when filtering

joaoviana:
Closes: [#12391](https://github.com/lightdash/lightdash/issues/12391)

### Description:

Now it returns the catalog item will all tags if one of them is included in the filtering query

demo: 

https://github.com/user-attachments/assets/3c331329-156d-46a6-a959-19e00dcb5ca5

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/CatalogModel/CatalogModel.ts
+++ b/packages/backend/src/models/CatalogModel/CatalogModel.ts
@@ -316,10 +316,18 @@ export class CatalogModel {
         }
 
         if (catalogTags) {
-            catalogItemsQuery = catalogItemsQuery.andWhere(
-                `${CatalogTagsTableName}.tag_uuid`,
-                'in',
-                catalogTags,
+            catalogItemsQuery = catalogItemsQuery.whereExists(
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -316,10 +316,18 @@ export class CatalogModel {
...
+                    void this.select('*')
+                        .from(CatalogTagsTableName)
+                        .whereRaw(
+                            `${CatalogTagsTableName}.catalog_search_uuid = ${CatalogTableName}.catalog_search_uuid`,
+                        )
+                        .whereIn(
+                            `${CatalogTagsTableName}.tag_uuid`,
+                            catalogTags,
+                        );
+                },
```
joaoviana, line 330: can you think of a different alternative @almeidabbm or is this solution ok?

almeidabbm: I think this is a good approach 💪

almeidabbm (approved):

Issue #12336, opened 2024-11-11
Long filter values break dashboard page layout


magnew:
### Description

Very long filter values cause the dashboards to have a side-scroll and push the data zoom controls off the page

These filters should get truncated/overflowed.

<img width="1504" alt="Screenshot 2024-11-11 at 10 06 05" src="https://github.com/user-attachments/assets/ab2ba0dc-1567-444e-9341-ecc6db8a8480">

### Steps to Reproduce the Bug or Issue

1. Make a very long filter on a dashboard

### Cloud or self-hosting

None

PriPatel:
@magnew Im not sure if we truncate or use a number of values. eg. what if the values are numbers?  I think that would annoy people? 
https://github.com/lightdash/lightdash/issues/6484
The suggestion was in this ticket here. Im thinking that this should be more 'width' based, rather than number of values based.  I suppose you could have one veryyy long value, and still have this issue?

andrewdw:
I may have started working on this before reading the comments. In my draft PR I'm working on a width based system. 
It gets the page width via a hook and sets the max-width of the text. Still todo: visually showing the value is truncated, and dealing with it being aware of other elements nearby that make the page-width a bit smaller

PR #12439, opened 2024-11-14
fix: dashboard breaking width if too many filters

andrewdw:
Closes: #12336 

### Description:
- filters now contained in a box that won't push the date filters off the page
- filters are set with a max width (arbitrary 800px for now) and truncated if larger
- filter container scrolls horizontally if the area is shrunk smaller than the filter width

https://github.com/user-attachments/assets/b353c980-bdcc-43e6-aa8d-3b3de9927b06

### Reviewer actions


magnew (approved):
Nice!
✅ Code looks good!
✅ This works well, I don't see anywhere where it breaks. And I think it's a big improvement. 
🤔 I wonder if we can't do something a little more elegant in the future -- it feels a little awkward with a few big filters, but that might be for whenever we can do a bit more dashboard design work. I don't have a better short-term suggestion.

Issue #12297, opened 2024-11-07
Deadlock detected when deleting a project


IrakliJani:
### Description

`delete from "projects" where "project_uuid" = $1 - deadlock detected`

internal thread: https://lightdash.slack.com/archives/C02GQKJK84Q/p1730971063842759

### Steps to Reproduce the Bug or Issue

-

### Cloud or self-hosting

None

PR #12449, opened 2024-11-15
fix: delete deadlock

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12297

### Steps

I'm not 100% sure this will fix the issue, but here is my proposal to minimize time spent. 

- Make this quick fix
- release
- flag sentry issue as resolved https://lightdash.sentry.io/issues/5486173409/events/ab060ab4715b4334a8c29a5c806993cd/?project=5959292
- if it happens again , then we keep investigating

### Description:

The proposal by claude makes sense though 

Based on the error message and the code provided, the deadlock is occurring during project deletion. The main deletion happens in the ProjectModel.delete method:

```
    async delete(projectUuid: string): Promise<void> {
        await this.database('projects')
            .where('project_uuid', projectUuid)
            .delete();
    }
```

The deadlock is likely occurring because there are multiple foreign key relationships with CASCADE DELETE constraints that are trying to delete related records simultaneously. Looking at the migrations and schema, we can see several tables with CASCADE DELETE constraints on project_uuid:
1. Cached explores:
2. DBT Cloud integrations:
3. Preview content:

The deadlock likely occurs when multiple cascading deletes try to acquire locks on related tables in different orders. To fix this, you could:
Wrap the deletion in a transaction to ensure consistent locking order
Explicitly delete related records in a specific order before deleting the project
Consider changing some CASCADE DELETE constraints to SET NULL if appropriate
Here's a suggested fix using a transaction:

```
async delete(projectUuid: string): Promise<void> {
    await this.database.transaction(async (trx) => {
        await trx('projects')
            .where('project_uuid', projectUuid)
            .delete();
    });
}
```

This ensures all cascading deletes happen within a single transaction, which should help prevent deadlocks by maintaining a consistent lock order.

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -455,9 +455,11 @@ export class ProjectModel {
     }
 
     async delete(projectUuid: string): Promise<void> {
-        await this.database('projects')
-            .where('project_uuid', projectUuid)
-            .delete();
+        await this.database.transaction(async (trx) => {
+            await this.database('projects')
+                .where('project_uuid', projectUuid)
+                .delete();
+        });
     }
 
     async getWithSensitiveFields(
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -456,7 +456,7 @@ export class ProjectModel {
 
     async delete(projectUuid: string): Promise<void> {
         await this.database.transaction(async (trx) => {
-            await this.database('projects')
+            await trx('projects')
                 .where('project_uuid', projectUuid)
                 .delete();
         });
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -456,9 +456,7 @@ export class ProjectModel {
 
     async delete(projectUuid: string): Promise<void> {
         await this.database.transaction(async (trx) => {
-            await trx('projects')
-                .where('project_uuid', projectUuid)
-                .delete();
+            await trx('projects').where('project_uuid', projectUuid).delete();
         });
     }
 


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -455,9 +455,11 @@ export class ProjectModel {
     }
 
     async delete(projectUuid: string): Promise<void> {
-        await this.database('projects')
-            .where('project_uuid', projectUuid)
-            .delete();
+        await this.database.transaction(async (trx) => {
+            await this.database('projects')
```
IrakliJani, line 459:
```suggestion
            await trx('projects')
```

IrakliJani (commented):
I think it's worth a try. But this needs change because transaction is being created but not used in the block.

packages/backend/src/models/ProjectModel/ProjectModel.ts
rephus: yes, of course, good catch, I didn't notice

IrakliJani (approved):

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -455,9 +455,11 @@ export class ProjectModel {
     }
 
     async delete(projectUuid: string): Promise<void> {
-        await this.database('projects')
-            .where('project_uuid', projectUuid)
-            .delete();
+        await this.database.transaction(async (trx) => {
+            await trx('projects')
+                .where('project_uuid', projectUuid)
+                .delete();
```
IrakliJani, line 461: patch to fix prettier
```suggestion
            await trx('projects').where('project_uuid', projectUuid).delete();
```

ZeRego:
> This ensures all cascading deletes happen within a single transaction, which should help prevent deadlocks by maintaining a consistent lock order.

@rephus @IrakliJani 
Using a transaction is helpful when there are multiple statements. 
I don't understand how this helps as they are already being deleted in a since transaction since we only trigger 1 statement 🤔

Issue #12427, opened 2024-11-14
I want to be able to sort the chart results table in the SQL runner


TuringLovesDeathMetal:
### Description

Right now, there is no way to sort the results table in the SQL runner chart tab. 

This is especially important when I add a pivot (because I can't get the sorting in my query results table to apply to the pivoted chart results table). 

There should be a way to sort the results in the chart tab. 

### Acceptance criteria:

- [ ] I am able to sort the columns in my chart results table. 
- [ ] if I click on a column, it sorts the column. 
- [ ] There is an arrow indicating the direction of sort on the column header
- [ ] for now, I can only sort by one column 
- [ ] **I should be able to sort by a pivoted column!**
- [ ] if I click on another column, then teh first sort is removed, and a new sort is applied only to that other column (see the semantic viewer for the sorting behaviour we want to replicate). 
- [ ] sorting that I apply to the results table is linked to the sorting I apply to the chart configuration. i.e. if I add `sort descending` to my x-axis field in the chart results table, then in the chart config, teh x-axis field should have `sort descending` also applied. If I change teh sort in the chart config panel, it should change the sort in the results table. 
- [ ] The sorting I apply in the chart results table should "override" the sorting from the sql query (i.e. if I have sorting in the sql query, then I add sorting int he results table, I should basically just be adding a `sort` to the chart results table query - not trying to append the sort to the SQL query or something weird).

PR #12463, opened 2024-11-15
feat: sorting in the SQL runner

magnew:
Closes: #12427 

### Description:

Adds sorting on any column to the SQL runner.

Kapture 2024-11-15 at 16 14 57

### Reviewer actions


TuringLovesDeathMetal (changes requested):
change that we chatted about in slack - I don't think this should interact with the x-axis ordering (unlike the spec in the ticket)

magnew:
> change that we chatted about in slack - I don't think this should interact with the x-axis ordering (unlike the spec in the ticket)

Removed sorting in the side bar

magnew:
> > change that we chatted about in slack - I don't think this should interact with the x-axis ordering (unlike the spec in the ticket)
> 
> Removed sorting in the side bar

Added back after discussing more.

TuringLovesDeathMetal (approved):
nice! Looks great :)

joaoviana (approved):
Works great! 

Wondering if we should add a `light/subtle` ⬆️ ⬇️  button with a tooltip label `Sort`

(can be done on a separate PR btw!)

magnew:
> Works great!
> 
> Wondering if we should add a `light/subtle` ⬆️ ⬇️ button with a tooltip label `Sort`
> 
> (can be done on a separate PR btw!)

We talked about it a little and left if off for now. I agree something would be nice though. Will follow up with something.

magnew:
Failing test is failing in main. Im going to merge this.

Issue #12496, opened 2024-11-19
On embedding, when scrolling, clicking on a URL cell creates the popup in a different place


rephus:
### Description

![image](https://github.com/user-attachments/assets/910150ca-09f8-4c18-84eb-df9ab0858d16)

### Steps to Reproduce the Bug or Issue

.

### Cloud or self-hosting

None

PR #12497, opened 2024-11-19
fix: fix cell context menu embedding

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12496

### Alternative solution: 

I think the issue is only happening on embedding because other dashboards/pages are wrapped by perhaps a scrollable component , with relative position, where `window.scrollY` is always 0. We could use the same wrapper on embedding to fix the issue, however, I'm not sure exactly how this is done, or if that will cause other issues (eg: when printing the page) 

I belive this solution (given we are already hacking the position using absolute) it is more reusable, since it doesn't depend on the wrapper. 

### Description:

Added scroll to absolute position of the menu

Before

Screenshot from 2024-11-19 08-53-25

After

Screenshot from 2024-11-19 09-13-00

I also tessted this on `dashboard tables` , `explore table` and `results` 

Screenshot from 2024-11-19 09-13-10

### Reviewer actions


IrakliJani (approved):
tested for non-embedded tables

Issue #12501, opened 2024-11-19
The colour in the selector is wrong in the chart config and saves as the colour in the selector (SQL Runner)


TuringLovesDeathMetal:
### Description

If I add two series to my chart, the multiple series are shown as a different colour in the series configuration, but then they're saved as teh colour in teh series configuration (instead of the colour shown on the chart). 

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/dd2531634a064328b1dfc12a9fa63cfa

### Cloud or self-hosting

None

PR #12510, opened 2024-11-19
fix: color picker not matching colors in sql runner

magnew:
Closes: #12501 

### Description:

The SQL color picker wasn't matching the colors displayed on the chart by default. We were passing an empty array instead of undefined, which was causing the colors to default incorrectly

To repro:
- Run a query
- Make a bar chart with multiple series
- Notice that the color picker is a different color than the second series by default 

Before:
before

With this fix:
Screenshot 2024-11-19 at 15 50 02

### Reviewer actions


rephus (approved):
Works as expected :+1:

Issue #12500, opened 2024-11-19
Changing my second series in the SQL runner doesn't actually change the chart type. 


TuringLovesDeathMetal:
### Description

If I have two series in the SQL runner, and I change the chart style of one of the series, it doesn't actually change on the SQL runner. 

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/93aa086e0d1a4fb0baabdf10cae11bfc

### Cloud or self-hosting

None

PR #12512, opened 2024-11-19
fix: enable setting second series type in SQL runner

magnew:
Closes: #12500 

### Description:

Setting a second series to a different type wasn't working in SQL runner. We were ignoring the series type in rendering. 

To repro:
- Make a bar chart with two series
- Try to change one series to 'line'
- Before: no effect

Now:
Screenshot 2024-11-19 at 16 44 44

### Reviewer actions


packages/common/src/types/savedCharts.ts
```diff
@@ -600,10 +596,7 @@ export const getEChartsChartTypeFromChartKind = (
         case ChartKind.SCATTER:
             return CartesianSeriesType.SCATTER;
         default:
-            return assertUnreachable(
```
owlas, line 603: why are we removing this?

owlas (commented):
It's working well. But what's the intended behaviour with stacking? Currently it's stacking the lines and bars together

packages/common/src/types/savedCharts.ts
magnew: Having this function only take a subset of the enum meant that you couldnt call it anywhere you had a ChartKind, so you have to add a typeguard everywhere you want to call it. Letting it take any ChartKind and returning a reasonable default seems a little cleaner. Especially since it's only ever called by our configuration forms. 

If you think it's better to keep the assert and typeguard it, I can put it back.

magnew:
> It's working well. But what's the intended behaviour with stacking? Currently it's stacking the lines and bars together

Updated to only stack bars. This is what we do in the explorer

owlas (approved):
works well!

PR #12529, opened 2024-11-20
feat(wip): clearable axis labels in explorer

owlas:
discussion: it's not possible to clear the y-axis label beacuse it's registered as an empty string, which is overridden when we render echarts code.

**Workaround** - set the field to a single whitespace `' '` - but quite annoying UX

**Suggestion** - have a clear button that sets the value to `' '`

**Altenative** - we could change the default render behaviour so that having `''` as the default name doesn't render the default name. But this would remove the default label from all existing charts for users that didn't specify this label.

**Future learning** - we should have a clearer `undefined` state

https://github.com/user-attachments/assets/50df60ab-6200-40e1-8b18-89cd7da2dc06


owlas:
Closing for now. Workaround is `' '` as the label

Issue #12528, opened 2024-11-20
Spacing in dashboards has changed 


PriPatel:
<img width="1433" alt="Screenshot 2024-11-20 at 09 51 56" src="https://github.com/user-attachments/assets/d3cb36f3-b4b6-4003-afa6-2f0f685944b9">

- [ ] Spacing on left hand side and right hand side of the tiles is uneven
- [ ] The 'add filter' button is offset . It should normally line up with the dashboard title and the start of the tiles.

I think a new change might have accidentally introduced this, so the behaviour needs to go back to how it was

ZeRego:
This was likely introduced with the changes to Dashboard Tabs. Make sure the fix works with both.

PriPatel:
Hmm - I look out for these changes when reviewing tabs. This feels very new, I wonder if there was anything else pushed this week?

PR #12533, opened 2024-11-20
fix: dashboard X spacing to be consistent with headers

andrewdw:
Closes: #12528 

### Description:
- makes dashboard spacing consistent with headers

CleanShot 2024-11-20 at 16 23 32

### Reviewer actions

### DIFF
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -690,18 +690,24 @@ const Dashboard: FC = () => {
                 withFullHeight={true}
             >
                 <Group position="apart" align="flex-start" noWrap px={'lg'}>
-                    <Box style={{ flexGrow: 1, overflow: 'auto' }}>
-                        {/* This Group will take up remaining space (and not push DateZoom) */}
-                        <Group position="apart" align="flex-start" noWrap>
-                            {dashboardChartTiles &&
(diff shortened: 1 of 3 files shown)


magnew (approved):
Tested with and without tabs, looks good ✅ 
Code makes sense if we want tabs going to the edges, which it looks like we do ✅

packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -692,12 +692,7 @@ const Dashboard: FC = () => {
                 <Group position="apart" align="flex-start" noWrap px={'lg'}>
                     <Box style={{ flexGrow: 1, overflow: 'auto' }}>
                         {/* This Group will take up remaining space (and not push DateZoom) */}
-                        <Group
-                            position="apart"
-                            align="flex-start"
-                            noWrap
-                            px={'lg'}
-                        >
+                        <Group position="apart" align="flex-start" noWrap>
```
joaoviana, line 695: I believe you can remove the Box above and just have the group like this

```
<Group
                        position="apart"
                        align="flex-start"
                        noWrap
                        grow
                        sx={{
                            overflow: 'auto',
                        }}
                    >


```

joaoviana (changes requested):
Once PR is rebased, let's wait for the e2es to pass 👍

joaoviana (approved):
shipit

Issue #11976, opened 2024-10-17
Change tabs to `Data`, `Series` (for cartesian) and `Display`


PriPatel:
**Cartesian chart types ( Bar / Line)** 
- [x] Data tab should pretty much be the same in terms of content
- [x] Replace the `Styling` wording with `Series`. 
- [x] Add a new tab 'Display'
- [x] Move axes labels from the `Series`tab  into `Display` tab

**Other charts types** (so far)
- [x] Leave `Table` and `Pie` charts as they are. They don't have enough options in there for multiple tabs to make sense

PR #12538, opened 2024-11-20
feat: update sql runner config tabs

magnew:
Closes: #11976 

### Description:

Updates the SQL editor tab layout to be more similar to explorer:
- Move axis configs to an Axes tab 
- Add a 'Series' tab

Note that the logic in the slice changes a bit as now series config is less dependent on axis config. 

Also note that this doesn't change at all the way the props are *read*. So it should have no effect on existing charts. New charts will have slightly different logic for where 'format' gets written. 

### Reviewer actions


packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
```diff
@@ -459,15 +450,15 @@ export const cartesianChartConfigSlice = createSlice({
...
-                        valueLabelPosition: action.payload.valueLabelPosition,
-                    };
-                }
+            if (
+                action.payload.index !== undefined &&
+                action.payload.reference !== undefined
+            ) {
+                display.series[action.payload.reference] = {
+                    ...display.series[action.payload.reference],
+                    yAxisIndex: action.payload.index, // TODO: Why set the index here?
```
joaoviana, line 459: so there's a way to reference the axis the series belongs to.

joaoviana: Without having to .find it - perf issue that is happening on current explorer

packages/frontend/src/components/DataViz/config/CartesianChartConfiguration.tsx
```diff
@@ -28,7 +29,8 @@ export const CartesianChartConfig = ({
             <Tabs color="gray" defaultValue="data" keepMounted>
                 <Tabs.List>
                     <Tabs.Tab value="data">Data</Tabs.Tab>
-                    <Tabs.Tab value="styling">Styling</Tabs.Tab>
+                    <Tabs.Tab value="series">Series</Tabs.Tab>
+                    <Tabs.Tab value="axes">Axes</Tabs.Tab>
```
joaoviana, line 33: Main ticket said to name it Display - any reason for renaming it?

joaoviana (approved):
Works great!

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
magnew: My question is more: why set it again when we are setting these other properties? presumably it gets set when the series is created, and this whole if block could be

`display.series[action.payload.reference].valuePosition = action.payload.valueLabelPosition`

No?

packages/frontend/src/components/DataViz/config/CartesianChartConfiguration.tsx
magnew: Because it has the same options now as 'Axes' in the Explorer and totally different options than 'Display' in the explorer. I was going for consistency. I can check with Katie.

magnew: I made it 'Display' 👍

packages/e2e/cypress/e2e/app/sqlRunnerNew.cy.ts
```diff
@@ -283,7 +283,7 @@ describe('SQL Runner (new)', () => {
         );
 
         // Perform the styling change
-        cy.contains('Styling').click();
+        cy.contains('Axes').click();
```
joaoviana, line 286:
```suggestion
        cy.contains('Display').click();
```

joaoviana (approved):

Issue #10419, opened 2024-06-14
Reference line not working when x-axis is a week dimension


jpetey75:
### Description

When creating a chart on data that is grouped by week, the reference line feature does not work (the line never appears).

https://www.loom.com/share/6807a66cfa8f4ff48ad9a0e2b7b5928b

### Steps to Reproduce the Bug or Issue

1. create a query that has a week dimension
2. configure a line or bar chart with week on the x-axis
3. add a reference line and choose one of the dates
4. no line appears

### Context on available options:

This is caused be the fact that we moved weeks to use a categorical axis. We suspected there would be some tradeoffs and this is the most painful tradeoff we've seen. Reference lines can't be added to any value on a categorical axis, then have to line up with a category. So you can have a reference line that corresponds with week-start, but not Wednesday, for example. A couple of options here:

- Keep the limitation that lines can only be on weeks, but make it more clear
- Use a numeric axis when there is a reference line is added. This will be a little weird with the labels changing.
- Try using a separate axis for the reference lines and hide everything but the line. This might work, might not. And might be very hard to get right.
- Go back to the drawing board with trying to customize numeric axes in such a way that we can get the labels we want and keep the lines.

Talk to @magnew for more details!

PR #12553, opened 2024-11-21
fix: reference line weeks

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10419

### Description:

Before

Screenshot from 2024-11-21 12-52-26

After
Screenshot from 2024-11-21 12-52-32

Screenshot from 2024-11-21 12-55-44

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -84,7 +84,10 @@ const getLabelFromField = (fields: ItemsMap, key: string | undefined) => {
     }
 };
 
-const getAxisTypeFromField = (item?: ItemsMap[string]): string => {
+const getAxisTypeFromField = (
+    item?: ItemsMap[string],
+    hasReferenceLine?: boolean,
+): string => {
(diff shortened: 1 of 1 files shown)


rephus:
Note this is not enough for extending the axis and making reference lines show int he past or the future. 

This looks like it is a different bug, I've been investigating a bit and it seems we return an empty list of fields for the bottom axis https://github.com/lightdash/lightdash/blob/main/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts#L1050 , which is used in `getMinAndMaxReferenceLines` , I'm not sure if this is only affecting weeks, all dates, or all types. 

I'd suggest to create a separate issue for this and prioritize it accordingly.

magnew (approved):
This seems better than what we have. In that it works 😄. Someone may complain about the labels changing, but its better than this just not working. 

Do you know why the line is offeset from the bar? Is it a 'week start' mismatch?

rephus:
> Do you know why the line is offeset from the bar? Is it a 'week start' mismatch?
Because the bar is on 15th, the reference line is on the 14th. And because we hide labels on overlap, I guess we show the axis label for the reference line

Issue #12488, opened 2024-11-18
User list in space permissions shows duplicate users 


jpetey75:
### Description

The user list in space permissions is showing the same users multiple times if they are in multiple groups with access to the space 

### Steps to Reproduce the Bug or Issue

https://www.loom.com/share/2bd12f39f3ab4c4aa870774e7022ad5d

1. add the same user to multiple user groups
2. give permission to a private space to those same user groups
3. user will appear multiple times in the user list for the space 

### Cloud or self-hosting

cloud

PR #12555, opened 2024-11-21
fix: user list in space permissions shows duplicates

andrewdw:
Closes: #12488 

### Description:
- Adds deduplication logic to not show the same user more than once in space permissions. This is caused when the user belongs to more than one group.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceUserList.tsx
+++ b/packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceUserList.tsx
@@ -529,8 +529,7 @@ export const ShareSpaceUserList: FC<ShareSpaceUserListProps> = ({
     );
 
     const accessByType = useMemo<SpaceAccessByType>(() => {
-        // function to get the highest role for a user
-        const getHighestAccess = (
+        const getDirectOrHighestAccess = (
             existing: SpaceShare,
             current: SpaceShare,
         ) => {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceUserList.tsx
```diff
@@ -529,7 +529,19 @@ export const ShareSpaceUserList: FC<ShareSpaceUserListProps> = ({
     );
 
     const accessByType = useMemo<SpaceAccessByType>(() => {
-        return space.access.reduce<SpaceAccessByType>(
+        const seenUsers = new Set<string>();
+
+        // don't show duplicate users - this can happen if a user belongs to multiple groups
+        const deduplicateUsers = (spaceShares: SpaceShare[]) =>
```
ZeRego, line 535: I believe this will cause issues if they have different roles in each group. Also, should the `seenUsers` be inside the function or is intentional that is filtering across project/org/direct?

ZeRego: The org and project collapse section should list all the users that inherit access from them.
The users list at the bottom should list all users with access to the space. That includes all users with direct access and inherited access from org/project/group. 
At the moment, we dedupe across all the sections, causing users to show in only 1 of those places.

I don't understand why users with inherited access are coming up with `hasDirectAccess = true`. Something is wrong up the chain :/

andrewdw: Updated it to only dedup users + show their highest role 🤙

packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceUserList.tsx
```diff
@@ -529,7 +529,27 @@ export const ShareSpaceUserList: FC<ShareSpaceUserListProps> = ({
...
+        ) => {
+            if (existing.hasDirectAccess) return existing; // direct access takes priority
+            const existingRoleNumber = roleOrder[existing.role];
+            const currentRoleNumber = roleOrder[current.role];
+            return currentRoleNumber > existingRoleNumber ? current : existing;
+        };
+
+        const deduplicatedUsersMap = new Map<string, SpaceShare>();
+
+        const result = space.access.reduce<SpaceAccessByType>(
```
ZeRego, line 552: We can update the reducer type and remove unused logic (push to direct array) and avoid a separate variable to track the dedupe user map.

Example new type 
```
reduce<{
    project: SpaceShare[];
    organisation: SpaceShare[];
    directMap: Map<string, SpaceShare>;
}>
```

packages/frontend/src/components/common/ShareSpaceModal/ShareSpaceUserList.tsx
```diff
@@ -529,13 +529,59 @@ export const ShareSpaceUserList: FC<ShareSpaceUserListProps> = ({
...
+            organisation: SpaceShare[];
+            direct: SpaceShare[];
+        }>(
             (acc, spaceShare) => {
                 if (spaceShare.hasDirectAccess) {
                     acc.direct.push(spaceShare);
-                } else if (spaceShare.inheritedFrom === 'project') {
+                } else if (
+                    spaceShare.inheritedFrom === 'project' ||
+                    spaceShare.inheritedFrom === 'group'
```
ZeRego, line 581: Amend case where user inherits access from project group

ZeRego (approved):

Issue #11405, opened 2024-09-04
Move project creation & updates to be a scheduled job


ZeRego:
### Description

Move project creation & updates to be a scheduled job.
Should cover the following cases from both UI and CLI:
- project create
- project deploy

### What problem does this solve?

These actions should be moved to a scheduled job for multiple reasons:
- Resolve issues when a project is being create/updated during a Lightdash update or when a server is shut down for other reasons(eg: error).
- reduce server memory spikes
- have a retry system
- better debug log

ZeRego:
TODO: /{projectUuid}/createPreview

PR #12570, opened 2024-11-21
chore: move project creation to scheduler

ZeRego:
Closes: [#11405](https://github.com/lightdash/lightdash/issues/11405)

### Description:

We don't need to update other endpoints as the project creation is the last one to still compile a project in the main server instead of the worker.

Changes:
- move the async work to a scheduled task
- update all project tasks to have 1 max attempts

#### On error (job 168):

Screenshot 2024-11-21 at 17 36 04

#### On success (job 169):

Screenshot 2024-11-21 at 17 36 52
Screenshot 2024-11-21 at 17 37 00

### Reviewer actions

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -779,6 +779,8 @@ export default class SchedulerTask {
                     error: e,
                 },
             });
+            // Update legacy job
+            await this.projectService._markJobAsFailed(payload.jobUuid);
             throw e;
         }
     }
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -34,6 +34,7 @@ import {
     QueryExecutionContext,
     ScheduledDeliveryPayload,
     SchedulerAndTargets,
+    SchedulerCreateProjectWithCompilePayload,
     SchedulerFilterRule,
     SchedulerFormat,
     SchedulerJobStatus,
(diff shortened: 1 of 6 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -627,7 +627,7 @@ export class ProjectService extends BaseService {
         return { jobId: schedulerJob.jobId };
     }
 
-    async create(
+    async scheduleCreate(
```
ZeRego, line 630: These changes are unreadable in Github. Here is a TLDR: 

Split `create` method into 2: `scheduleCreate` and `_create` 

`scheduleCreate` responsibilities:
- creates the legacy job&steps in the database which are used by the UI
- schedule the creation of the project, which will call `_create`

`_create` responsibilities:
- has the same logic as before, tests the adaptor, compiles, etc
- updates the job&steps status

IrakliJani (commented):
I see that you removed scheduler job polling in the FE from the original PR, but since we do return the (legacy) job ID, it’s fine most of the time. However, I still think it should be there because if the scheduler job fails to execute, there’s no way to report the error back to the user.

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -779,6 +779,8 @@ export default class SchedulerTask {
                     error: e,
                 },
             });
+            // Update legacy job
+            await this.projectService._markJobAsFailed(payload.jobUuid);
```
IrakliJani, line 783: nice!

IrakliJani (approved):

Issue #11590, opened 2024-09-18
When adding a filter, move the cursor to the input field


rephus:
### Description
when you add a filter it would be nice to move the cursor straight to the input field.

- [ ] Optionally, it would be nice you could hit CMD+ENTER to run the query while you are in the input field. (currently it captures the enter key)
### What problem does this solve?

 A very common pattern for usersz is to type a string into the filter, scroll down, hit enter. If the cursor then moved to the input, it would be super clean

sanjay-sol:
Hey @rephus ! Can I work on this ?

Josephine19001:
I will like to work on this

chidexebere:
@ZeRego seems this is no longer an issue because following the screen recording I shared above...except I am missing something

PR #12588, opened 2024-11-22
feat: focus filter input on open

andrewdw:
Closes: #11590 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/FieldSelect/index.tsx
+++ b/packages/frontend/src/components/common/FieldSelect/index.tsx
@@ -18,7 +18,14 @@ import {
     Tooltip,
     type SelectProps,
 } from '@mantine/core';
-import { forwardRef, useCallback, useMemo } from 'react';
+import {
+    forwardRef,
+    useCallback,
+    useEffect,
+    useMemo,
+    useRef,
+    useState,
+} from 'react';
 import FieldIcon from '../Filters/FieldIcon';
 
 interface ItemComponentProps extends React.ComponentPropsWithoutRef<'div'> {
@@ -77,6 +84,8 @@ type FieldSelectProps<T extends Item = Item> = Omit<
     onClosed?: () => void;
     hasGrouping?: boolean;
     baseTable?: string;
+    focusOnRender?: boolean;
+    onCmdEnter?: () => void;
 };
 
 const getLabel = (item: Item, hasGrouping: boolean) => {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/common/FieldSelect/index.tsx
```diff
@@ -203,6 +226,7 @@ const FieldSelect = <T extends Item = Item>({
 
     return (
         <Select
+            ref={inputRef}
```
IrakliJani, line 229: if I remember correctly, there should be `autofocus` prop on the component. can you try that out?

andrewdw: CleanShot 2024-11-25 at 18 19 35@2x

It seems to have the prop but it doesn't do anything... am I missing something?

andrewdw: hmm I found and tested this solution: https://stackoverflow.com/questions/69456631/how-do-i-set-autofocus-to-an-input-thats-inside-a-select-dropdown

but it re-focuses on every render which is not going to work

IrakliJani: gotcha

packages/frontend/src/components/common/FieldSelect/index.tsx
```diff
@@ -93,8 +101,23 @@ const FieldSelect = <T extends Item = Item>({
...
+    useEffect(() => {
+        setIsMounted(true);
+    }, []);
+    useEffect(() => {
+        if (isMounted && focusOnRender && inputRef.current) {
+            inputRef.current.focus();
+        }
+    }, [isMounted, focusOnRender]);
+
+    // Create table label mapping and sort items
```
IrakliJani, line 120: unnecessary comment

packages/frontend/src/components/common/FieldSelect/index.tsx
```diff
@@ -93,8 +101,23 @@ const FieldSelect = <T extends Item = Item>({
     inactiveItemIds = [],
     hasGrouping = false,
     baseTable,
+    focusOnRender = false,
     ...rest
 }: FieldSelectProps<T>): JSX.Element => {
+    const inputRef = useRef<HTMLInputElement | null>(null); // Input ref for focus handling
+
+    // Focus on input when mounted
+    const [isMounted, setIsMounted] = useState(false);
```
IrakliJani, line 110: `isMounted` state is unnecessary. React’s useEffect already runs after the component mounts, so you can directly focus the input inside a useEffect without the additional state.

packages/frontend/src/components/common/FieldSelect/index.tsx
```diff
@@ -93,8 +101,23 @@ const FieldSelect = <T extends Item = Item>({
...
+    useEffect(() => {
+        setIsMounted(true);
+    }, []);
+    useEffect(() => {
+        if (isMounted && focusOnRender && inputRef.current) {
+            inputRef.current.focus();
+        }
+    }, [isMounted, focusOnRender]);
```
IrakliJani, line 118:
```tsx
useEffect(() => {
    if (focusOnRender && inputRef.current) {
        inputRef.current.focus();
    }
}, [focusOnRender]);
```

this should be enough here

packages/frontend/src/components/common/FieldSelect/index.tsx
andrewdw: Hmm removing the isMounted state doesn't seem to trigger the focus anymore. but throwing it to the bottom of the event loop with this works. How do you feel about this?

```
 useEffect(() => {
        if (focusOnRender) {
            const interval = setInterval(() => {
                if (inputRef.current) {
                    inputRef.current.focus();
                    clearInterval(interval);
                }
            }, 10);
        }
    }, [focusOnRender]);```

IrakliJani: @andrewdw 

`setInterval` can be replaced with `setTimeout` and there is no need for `clearInterval`

but also, can you just try this instead? I tested locally and it worked just fine

```ref={(ref) => ref?.focus()}```

IrakliJani (approved):

Issue #12591, opened 2024-11-25
Allow peeking metric and get info in Modal


joaoviana:
### Access
- [ ] You can access the catalog from any metric in the metrics catalog by clicking on the metric name
- [ ] There is a 'close' button to get out of this view
- [ ] When users click on a metric, the URL changes
- [ ] When the user visits the same URL, it opens the chart automatically

PR #12593, opened 2024-11-25
feat: display metric peek modal on click of metric column name

joaoviana:
Closes: [#12591](https://github.com/lightdash/lightdash/issues/12591)

### Description:

- adds new endpoint to get a metric through /dataCatalog/metrics/{tableName}/{metricName}
- it gets the filtered explore, which accounts for user attributes 
- gets metric with hook

on the UI:
- when the user clicks on a metric name, it will open the modal (tooltips and improvements in interaction shall be done separately)
- when the modal opens, the url changes as well to: `/metrics/peek/orders/average_order_size`
- if the user visits this url, the modal will open 

demo: 

https://github.com/user-attachments/assets/682fff88-dc14-417f-80a5-6b06291ca321

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
@@ -9,6 +9,7 @@ import {
     Paper,
     Portal,
     Tooltip,
+    UnstyledButton,
 } from '@mantine/core';
 import { useClickOutside } from '@mantine/hooks';
 import { IconTable, IconTrash } from '@tabler/icons-react';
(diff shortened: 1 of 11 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricsCatalogPanel.tsx
```diff
@@ -160,6 +172,20 @@ export const MetricsCatalogPanel = () => {
         [user.data, dispatch, projectUuid],
     );
 
+    useEffect(
+        function openMetricPeekModal() {
```
IrakliJani, line 176: im so used to inline functions that I was confused with this 😆

packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
```diff
@@ -241,15 +259,19 @@ export const MetricsCatalogColumnName = forwardRef<HTMLDivElement, Props>(
...
                         openDelay={300}
                     >
-                        <Highlight
-                            highlight={table.getState().globalFilter || ''}
-                            c="dark.9"
-                            fw={500}
-                            fz="sm"
-                            lh="150%"
+                        <UnstyledButton
+                            onClick={() => handlePeekMetric(row.original)}
```
IrakliJani, line 263: is it okay to have it exposed like this or should we hide it better?

IrakliJani (approved):

packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
joaoviana: I think it's ok from chatting with @PriPatel

Issue #11925, opened 2024-10-14
I would like to be able to have 2 series one on the left and one on the right axis


almeidabbm:
Blocked by: https://github.com/lightdash/lightdash/issues/11939

### Description

![image](https://github.com/user-attachments/assets/7703330a-2365-48af-a723-fdc778382509)

joaoviana:
Blocked by #11611

magnew:
Blocked by #12286

hamzahc1:
if this is blocked - do you want to move it back and work on the blocking ticket instead?

PR #12608, opened 2024-11-25
feat: I would like to be able to have 2 series one on the left and one on the right axis

magnew:
Closes: #11925 

### Description:

Enables multiple axes SQL runner charts. 

Notes on implementation:
- introduces a new `whichYAxis` prop on series, which maps to the echarts `yAxisIndex` prop determining which axis should be used for the series
- adds UI to set `whichYAxis` per series
- a lot of the code here is for determining which axes to show in the edge cases and determining what the axis label should be
- removes the ability to set which side an axis is on

Back compat:
- This changes *slightly* the way we read already saved charts in 2 ways: 1) We will no longer check the saved yAxisIndex when determining which axis a series is on. This should be ok since we had a bug before where that value was always 0. 2) We do still check axis position for the primary axis, but you can no longer set it. For older saved charts, they should render the same, but if they are edited, there is no way to change the axis back so they might need to be re-created. 

Kapture 2024-11-25 at 18 01 52

### Reviewer actions


hamzahc1:
What happened with this PR? Did we make any progress on it in the last day?

almeidabbm:
> What happened with this PR? Did we make any progress on it in the last day?

Yes, there was progress and @magnew was trying to make some changes which would make this logic a bit clearer but there was a problem regarding the pivoted column the series refers to

I am approving this as is since it is working, will just try to get the pipeline passing

almeidabbm (approved):

almeidabbm (changes requested):
@ZeRego spotted 2 bugs which are possibly related to the same thing where post aggregation column names are not being taken into account when setting the display key (reference) and ends up overwriting the object

- flipping axis position of series with the same reference will change the axis position of the second series as well


https://github.com/user-attachments/assets/e94f3f1f-2a68-4816-a75f-3afc245db3b7

- left axis doesn't appear in display when first series has right position

Screenshot 2024-11-26 at 15 29 24
Screenshot 2024-11-26 at 15 29 29

magnew:
@almeidabbm thank you for the thorough review! Looking into both of the bugs you mentioned, they both only repro when you have duplicate fields in the chart. For example `amount` as an average and `amount` as a sum. This seems to already be a problem with duplicate fields in the SQL runner -- if you have two fields with different aggregations, changing their properties causes changes to both. 

I was talking to @ZeRego about this and we think since this is already a problem, we could merge this review. I will file a ticket to fix duplicate fields. I looked into it a bit and it's going to be a fairly large change, so I don't want to jam it in here. In the mean time we *could* add a restriction so that people can't have multiple series of the same field for now. 

The other issue you both mentioned (correctly) is that the usage or `yAxisIndex` and `whichYAxis` is a bit confusing. I played around with renaming `yAxisIndex` and it's definitely going to break pretty much all existing charts. SO, I think if we are going add this feature now, there is not a much better way to do it. I think we have two options: 1) Punt this work and not do it for now so we don't introduce more confusion, 2) Merge this and address the `yAxisIndex` rename with a future migration or data format update. This could be a little painful, but I think there is going to have to be a migration and/or update to the format to address pivotting correctly anyway.

almeidabbm (approved):

almeidabbm:
> @almeidabbm thank you for the thorough review! Looking into both of the bugs you mentioned, they both only repro when you have duplicate fields in the chart. For example `amount` as an average and `amount` as a sum. This seems to already be a problem with duplicate fields in the SQL runner -- if you have two fields with different aggregations, changing their properties causes changes to both.
> 
> I was talking to @ZeRego about this and we think since this is already a problem, we could merge this review. I will file a ticket to fix duplicate fields. I looked into it a bit and it's going to be a fairly large change, so I don't want to jam it in here. In the mean time we _could_ add a restriction so that people can't have multiple series of the same field for now.
> 
> The other issue you both mentioned (correctly) is that the usage or `yAxisIndex` and `whichYAxis` is a bit confusing. I played around with renaming `yAxisIndex` and it's definitely going to break pretty much all existing charts. SO, I think if we are going add this feature now, there is not a much better way to do it. I think we have two options: 1) Punt this work and not do it for now so we don't introduce more confusion, 2) Merge this and address the `yAxisIndex` rename with a future migration or data format update. This could be a little painful, but I think there is going to have to be a migration and/or update to the format to address pivotting correctly anyway.

I would say let's go with option 2, this is something that should be available and we need to fix the underlying issue anyway, so it's better to ship it as is and create the tickets to fix the actual problem preventing the duplicate references from working

PR #12617, opened 2024-11-26
feat: charts as code cli

rephus:
Closes: 

### USage

requirements: 
- have a project on CLI, you can change it by doing `lightdash config set-project`
- on CLI, type `lighthdash download` to download all charts as JSON in a `target` folder
- UPdate chat files locally
- on CLI, type `lightdash upload` to update or create these charts in lightdash 

[Screencast from 2024-11-26 16-23-24.webm](https://github.com/user-attachments/assets/311d789c-36d1-4fae-bca0-1a3484a64abe)

image

### How it works 

We created 2 new endpoints, one for get all charts, and another for creating/updating a chart

Created `coderController` + `coderService` : where I use existing methods from savedChartModel. 

These endpoints are using a new chart type called `ChartAsCode`. This new type *contains all chart details a user can modify* (we exclude fields like projectUuid, orgUuid,  upbatedByUser, acces, or readOnly details) 

*download*
CLI makes a call to the backend to get all charts and configs 

*upload*
CLI makes a call to the backend `per chat` to create if the chart with the slug does not exist, or update if the chart with slug exists

### How to test

Checkout this code and build CLI locally 

> yarn cli-build && ./packages/cli/dist/index.js download

### Not included in this PR (out of scope) 

- ~permissions~ added basic permissions for admins, since this endpoint is going to be public
- space creation if missing 
- analytic events
- database optimizations 
- scheduler job . Although this is going to be using the API heavily, these endpoints should be quick, so I'm not too worried about locking the API
- CLI update paralelization (I'd rather have a sequential method in the CLi that updates the charts one by one, than making 100 requests to the backend at the same time and locking the API) 
- Logic to check if we even need to update charts (based on updated time?)
- CLI waiting spinners
- Warn users when chart is overwritten

### Reviewer actions


packages/backend/src/controllers/coderController.ts
```diff
@@ -0,0 +1,64 @@
...
+        };
+    }
+
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Post('/chart')
+    @OperationId('upsertChartAsCode')
+    async upsertChartAsCode(
+        @Path() projectUuid: string,
+        @Body() chart: Omit<ChartAsCode, 'metricQuery'> & { metricQuery: any },
```
rephus, line 53: metricQuery is a type too compliex for tsoa to handle, we already do something similar in other endpoints. We  can add a few more details to this type though.

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.lightdashConfig = lightdashConfig;
+        this.analytics = analytics;
+        this.projectModel = projectModel;
+        this.savedChartModel = savedChartModel;
+        this.dashboardModel = dashboardModel;
+        this.spaceModel = spaceModel;
+        this.schedulerClient = schedulerClient;
+    }
+
+    private static transformChart(
```
rephus, line 64: converts a complex SavedChartDAO into a `ChartAsCode` simple type, by removing some extra details, uuids , and adding some chartAsCode specific stuff like version or spaceSlug.

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,180 @@
...
+        user: SessionUser,
+        projectUuid: string,
+        chartAsCode: ChartAsCode,
+    ) {
+        // TODO handle permissions
+
+        const [chart] = await this.savedChartModel.find({
+            slug: chartAsCode.slug,
+        });
+        if (chart === undefined) {
```
rephus, line 129: if chart with slug does not exist, we create, if it exists, we update

packages/backend/src/controllers/coderController.ts
```diff
@@ -0,0 +1,64 @@
...
+                .getCoderService()
+                .getCharts(req.user!, projectUuid),
+        };
+    }
+
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Post('/chart')
+    @OperationId('upsertChartAsCode')
+    async upsertChartAsCode(
```
rephus, line 51: This endpoint is currently taking only 1 chart at a time, we can easily modify this to take an array of chartsAsCode. The chart to create/update is referenced using the `slug` inside the Body

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,180 @@
...
+            name: chartAsCode.name,
+            description: chartAsCode.description,
+        });
+
+        const updatedChart = await this.savedChartModel.createVersion(
+            chart.uuid,
+            chartAsCode,
+            user,
+        );
+        const spaces = await this.spaceModel.find({
```
rephus, line 172: we could simply return `created:boolean` without the chart, to avoid this extra request.

packages/cli/src/handlers/download.ts
```diff
@@ -0,0 +1,132 @@
...
+import * as path from 'path';
+import { getConfig } from '../config';
+import GlobalState from '../globalState';
+import { checkLightdashVersion, lightdashApi } from './dbt/apiClient';
+
+const DOWNLOAD_FOLDER = 'target';
+export type DownloadHandlerOptions = {
+    verbose: boolean;
+};
+export const downloadHandler = async (
```
rephus, line 18: Code for the CLI , it ony makes a single request to the API, and does file manipulation.

packages/cli/src/handlers/download.ts
```diff
@@ -0,0 +1,132 @@
...
+    AuthorizationError,
+    SavedChart,
+} from '@lightdash/common';
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import { getConfig } from '../config';
+import GlobalState from '../globalState';
+import { checkLightdashVersion, lightdashApi } from './dbt/apiClient';
+
+const DOWNLOAD_FOLDER = 'target';
```
rephus, line 14: I picked this folder name because it is already included in our gitignore, but we can change it.

packages/cli/src/handlers/download.ts
```diff
@@ -0,0 +1,132 @@
...
+export type DownloadHandlerOptions = {
+    verbose: boolean;
```
rephus, line 16: we can add some extra properties like --project_uuid, or --chart_uuid

packages/backend/src/controllers/coderController.ts
```diff
@@ -0,0 +1,64 @@
...
+    Response,
+    Route,
+    SuccessResponse,
+    Tags,
+} from '@tsoa/runtime';
+import express from 'express';
+import { allowApiKeyAuthentication, isAuthenticated } from './authentication';
+import { BaseController } from './baseController';
+
+@Route('/api/v1/projects/{projectUuid}/coder')
```
owlas, line 25: This makes more sense attached to the existing routes in the project controller

```
GET /api/v1/projects/{projectUuid}/charts <- already exists

GET /api/v1/projects/{projectUuid}/charts/code <- to get all charts as code

POST /api/v1/projects/{projectUuid}/charts/{slug)/code <- consistent

File: packages/backend/src/services/CoderService/CoderService.ts [hunk 10]
```diff
@@ -0,0 +1,180 @@
...
+        spaceSummary: Pick<SpaceSummary, 'uuid' | 'slug'>[],
+    ): ChartAsCode {
+        const spaceSlug = spaceSummary.find(
+            (space) => space.uuid === chart.spaceUuid,
+        )?.slug;
+        if (!spaceSlug) {
+            throw new NotFoundError(`Space ${chart.spaceUuid} not found`);
+        }
+
+        return {
```
owlas, line 75: making this type "opt in" makes sense!

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,180 @@
...
+        };
+    }
+
+    async getCharts(user: SessionUser, projectUuid: string) {
+        const project = await this.projectModel.get(projectUuid);
+        if (!project) {
+            throw new NotFoundError(`Project ${projectUuid} not found`);
+        }
+
+        // TODO handle permissions
```
owlas, line 98: should add this before shipping there's already an endpoint that gets all charts for a project: https://github.com/lightdash/lightdash/blob/259bb325630bebcf8cd26b6cf6dffbc98e82afbb/packages/backend/src/services/ProjectService/ProjectService.ts#L3841-L3841

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,180 @@
...
+        // We need to get the charts and all the chart config
+        // At the moment we are going to fetch them all in individual queries
+        // But in the future we should fetch them in a single query for optimiziation purposes
+        const chartSummaries = await this.savedChartModel.find({ projectUuid });
+        const chartPromises = chartSummaries.map((chart) =>
+            this.savedChartModel.get(chart.uuid),
+        );
+        const charts = await Promise.all(chartPromises);
+
+        // get all spaces to map  spaceSlug
```
owlas, line 111: is this so we can manage which space the chart is in? We don't need this if not

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,180 @@
...
+        const spaceUuids = charts.map((chart) => chart.spaceUuid);
+        const spaces = await this.spaceModel.find({ spaceUuids });
+        return charts.map((chart) =>
+            CoderService.transformChart(chart, spaces),
+        );
+    }
+
+    async upsertChart(
```
owlas, line 119: Can we resuse logic from the promoting charts etc. here?

packages/cli/src/handlers/download.ts
owlas: `target` is a throwaway folder in a dbt project, so if this is run in a dbt directory the .yml will likely get deleted. 

What about `lightdash`?

packages/cli/src/handlers/download.ts
```diff
@@ -0,0 +1,132 @@
...
+    try {
+        await fs.mkdir(outputDir, { recursive: true });
+    } catch (error) {
+        // Directory already exists
+    }
+
+    for (const chart of chartsAsCode) {
+        const chartPath = path.join(outputDir, `${chart.slug}.json`);
+        GlobalState.debug(`> Writing chart to ${chartPath}`);
+        await fs.writeFile(chartPath, JSON.stringify(chart, null, 2));
```
owlas, line 58: let's serialize as yml now, it's more consistent with a dbt project

packages/backend/src/services/CoderService/CoderService.ts
rephus: In the future we might want to create a space if it doesn't exist, just from the chart (similar how promotion works) , but because we do'nt want to have uuids, I added a new `spaceSlug` property to ChartsAsCode

packages/backend/src/controllers/coderController.ts
rephus: Done, removed this controller and moved to projectController

packages/backend/src/services/CoderService/CoderService.ts
rephus: Done, as Katie suggested, right now it is only avaialable for admins

packages/cli/src/handlers/download.ts
rephus: Done

rephus: We could add in the future a --json format on the donwload command to allow both formats

packages/backend/src/services/CoderService/CoderService.ts
rephus: It is 100% doable, and we should  reuse that code soon. 

I think it requires a bit of work to adapt thos emethods and fake some data, so it would look like:  

```
        let promotionChanges: PromotionChanges =
        PromoteService.getChartChanges(chartAsCode, existingChart);

    promotionChanges = await this.promoteService.upsertSpaces(user, promotionChanges);

    promotionChanges = await this.promoteService.upsertCharts(user, promotionChanges);

```

I would prioritize this over having dashboards, because then having dashboards will be really easy. And spaces will be handled automatically

docs/docs/guides/filtering-dashboard-in-url.mdx
```diff
@@ -8,7 +8,7 @@ For example, you might have a sales dashbaord that shows your most profitable Pa
```
owlas, line 8: the mdx files were added by accident

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    async getCharts(user: SessionUser, projectUuid: string) {
+        const project = await this.projectModel.get(projectUuid);
+        if (!project) {
+            throw new NotFoundError(`Project ${projectUuid} not found`);
+        }
+
+        // TODO handle permissions
+        // Filter charts based on user permissions (from private spaces)
+        if (user.ability.cannot('manage', subject('Project', project))) {
```
owlas, line 101: does this project definitely include the organization uuid?

owlas (approved):
Looks good!

packages/backend/src/services/CoderService/CoderService.ts
rephus: it does, but I can change it to be explicit

PR #12639, opened 2024-11-28
feat: metric trees backend

almeidabbm:
Closes: #12638 

### Description:

- Creates `metrics_tree_edges` table
- Adds endpoints to `CREATE`, `DELETE` and `GET` the edges

What this PR doesn't do:
- handle permissions at the backend level
- node positions - this can come later since we first would like to just save edges and then use a layout lib to display everything

### Reviewer actions


packages/backend/src/database/migrations/20241128121723_add-metric-tree-edges-table.ts
```diff
@@ -0,0 +1,33 @@
...
+                .references('catalog_search_uuid')
+                .inTable(CATALOG_SEARCH_TABLE)
+                .onDelete('CASCADE');
+
+            table.primary([
+                'source_catalog_search_uuid',
+                'target_catalog_search_uuid',
+            ]);
+        });
+    }
```
IrakliJani, line 28: does this table gets refreshed every time catalog refreshes? I guess yes but im still conflicted with the idea of using catalog search uuids in any form and just rely on (table, metric) pairs as primary source of links

IrakliJani (approved):

packages/common/src/types/catalog.ts
```diff
@@ -94,6 +94,11 @@ export type CatalogTable = Pick<
 
 export type CatalogItem = CatalogField | CatalogTable;
 
+export type CatalogMetricsTreeEdge = {
+    source: Pick<CatalogField, 'catalogSearchUuid' | 'name' | 'tableName'>;
+    target: Pick<CatalogField, 'catalogSearchUuid' | 'name' | 'tableName'>;
```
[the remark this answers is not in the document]
almeidabbm: 1. Yes the catalog search uuid is the reference that we use for the `CatalogSearch` table
2. The `tableName` here is the base of the explore where this metric is found, the catalog only contains rows for metrics/dimensions in the base table explore `packages/backend/src/models/CatalogModel/CatalogModel.ts:706`

packages/backend/src/database/migrations/20241128121723_add-metric-tree-edges-table.ts
almeidabbm: I completely agree, I think we should get rid of that ID and use `tableName`, `name` as the primary key, but for now we gotta use these UUIDs

Not sure when we should do this tbh

packages/backend/src/database/migrations/20241129140659_add-table-name-to-catalog-search.ts
```diff
@@ -0,0 +1,35 @@
...
+    await knex(CATALOG_SEARCH_TABLE).update({
+        // @ts-ignore - we're doing something similar in 20240604094543_add_views_columns_to_chart_and_dashboards_tables.ts
+        table_name: knex
+            .select(knex.raw(`explore->>'baseTable'`))
+            .from(CACHED_EXPLORE_TABLE)
+            .whereRaw(
+                `${CACHED_EXPLORE_TABLE}.cached_explore_uuid = ${CATALOG_SEARCH_TABLE}.cached_explore_uuid`,
+            )
+            .limit(1),
+    });
```
IrakliJani, line 22: will this be a problematic because of slow migration for app cloud?

packages/backend/src/models/CatalogModel/utils/index.ts
```diff
@@ -30,6 +30,7 @@ export const convertExploresToCatalog = (
                 type: CatalogType.Table,
                 required_attributes: baseTable.requiredAttributes ?? {}, // ! Initializing as {} so it is not NULL in the database which means it can't be accessed
                 chart_usage: null, // Tables don't have chart usage
+                table_name: explore.baseTable,
```
IrakliJani, line 33: 👌

IrakliJani (commented):
code looks good. approving with the concern of data migration

packages/backend/src/database/migrations/20241129140659_add-table-name-to-catalog-search.ts
owlas: It should be possible to do this update with a single query and not `N` subqueries, that would make it significantly faster! I'm not sure it's supported with knex syntax but with knex raw you would do:

```sql
UPDATE catalog_search
SET table_name = cached_explore.explore->>baseTable
FROM cached_explore
WHERE catalog_search.cached_explore_uuid = cached_explore.cached_explore_uuid
```

This assumes:
1. That `catalog_search` is `m:1` with `cached_explore`
2. And that `cached_explore` has all the entries for `catalog_search` maybe that's not true?

almeidabbm: Amazing! I have updated the code using `knex.raw` as `updateFrom` PostgreSQL specific `updateFrom` only got introduced in knex 2.5.0 https://github.com/knex/knex/releases/tag/2.5.0

owlas (approved):

Issue #12574, opened 2024-11-22
when we wrap a query with `limit` in the SQL runner, the ordering isn't persisted


TuringLovesDeathMetal:
### Description

Right now, when we load 500 rows in the SQL runner, we do this by adding a CTE and a `limit` clause wrapped around the SQL query. 

The issue with this is that the ordering isn't guaranteed to be persisted. 

So, the ordering can become wrong in the results/look jumbled. So far, this has only been reported in Snowflake, but it could also be an issue in other data warehouses.

Some options to resolve this:

1. use the snowflake SDK and the `limit` parameter (?)
2. add a `limit` clause to the end of the query (we would need to make sure there isn't an existing `limit` clause int he query already). 

### Steps to Reproduce the Bug or Issue

1. add a query in the SQL runner using the snowflake warehouse connection
2. add ordering to your SQL query (e.g. order by date)
3. see that the results are in the wrong order
4. run the same query in teh Snowflake query executor
5. see that the query is ordered correctly
6. add a CTE and a limit clause to the query
7. see that the query is ordered incorrectly

### Cloud or self-hosting

None

andrewdw:
@TuringLovesDeathMetal Do you have a sample query I can try in our staging Snowflake instance that I can test against? I'm having a bit of trouble replicating

PR #12647, opened 2024-11-28
fix: strip subquery query limits and append to end

andrewdw:
Closes: #12574 

### Description:
- modifies the `applyLimitToSqlQuery` function to strip out limits and semicolons from subqueries and applies limit to end of query
- It also looks for the smallest outer limit applied and will use the smaller one

// keyword to include frontend tests
test-frontend

### Reviewer actions

### DIFF
--- a/packages/backend/src/queryBuilder.test.ts
+++ b/packages/backend/src/queryBuilder.test.ts
@@ -993,7 +993,7 @@ describe('Time frame sorting', () => {
 });
 
 describe('applyLimitToSqlQuery', () => {
-    // strip out formatting so we just test content
+    // strip out white space formatting so we just test content
     const normalizeSQL = (sql: string) => sql.replace(/\s+/g, ' ').trim();
 
     it('should return the original query if limit is undefined', () => {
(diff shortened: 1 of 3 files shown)


packages/backend/src/queryBuilder.ts
```diff
@@ -307,10 +386,14 @@ export const applyLimitToSqlQuery = ({
...
-        '',
-    )}\n) select * from user_sql limit ${limit}`;
+    const existingLimit = extractOuterLimitFromSQL(sqlQuery);
+    if (existingLimit !== undefined && existingLimit <= limit) {
+        // existing limit is less than or equal to the provided limit
+        return sqlQuery;
+    }
+    // remove existing limits and apply the new limit
+    const sqlWithoutCommentsAndLimits = removeCommentsAndLimits(sqlQuery);
+    return `WITH user_sql AS (\n${sqlWithoutCommentsAndLimits}\n) select * from user_sql limit ${limit}`;
```
ZeRego, line 396: The initial order problem is caused because of this logic that wraps their SQL in a CTE. 
So we need to add a `LIMIT ${limit}` at the end of their query instead.

packages/backend/src/queryBuilder.test.ts
```diff
@@ -1038,9 +1051,126 @@ describe('applyLimitToSqlQuery', () => {
...
+    it('should handle queries with LIMIT in strings', () => {
+        const sqlQuery = `
+            SELECT * FROM users WHERE name = 'LIMIT 10';
+        `;
+        const limit = 10;
+
+        // Adjust the expected query to exclude the semicolon
+        const expectedQuery = `WITH user_sql AS (\n${sqlQuery
+            .trim()
+            .replace(/;$/, '')}\n) select * from user_sql limit ${limit}`;
```
ZeRego, line 1143: If possible, hardcode the expected result. Otherwise, we might fall into a problem of coding the same logic that we are testing.

ZeRego (approved):
lgtm 👍 good job

PR #12648, opened 2024-11-28
style: add new offWhite color & fix layout issue

joaoviana:
Closes: N/A

### Description:

Adds new Offwhite color - there's a theory that we could add an _off_ color pallette (thank you @PriPatel for the suggestion)
but just one color is ok for now. 

Fixed a layout issue where the _sidebar_ of the modal wasn't taking full height 
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/mantineTheme.ts
+++ b/packages/frontend/src/mantineTheme.ts
@@ -16,6 +16,10 @@ export const getMantineThemeOverride = (overrides?: {
     // Without it things look a little darker than before.
     black: '#111418',
 
+    colors: {
+        offWhite: ['#FDFDFD'],
+    },
+
     spacing: {
         one: rem(1),
         two: rem(2),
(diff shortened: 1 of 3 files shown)


packages/frontend/src/mantineTheme.ts
```diff
@@ -16,6 +16,10 @@ export const getMantineThemeOverride = (overrides?: {
     // Without it things look a little darker than before.
     black: '#111418',
 
+    colors: {
+        offWhite: ['#FDFDFD'],
```
joaoviana, line 20: when I tried to override `white` it impacted a lot of other parts of the app.

IrakliJani: oh good, I think white should stay as is at least for now

IrakliJani (approved):

Issue #12658, opened 2024-11-29
Add data zoom


joaoviana:
Experiment with 'zooming' in and out. The back up is date zoom and how it works in dashboards.

PR #12662, opened 2024-11-29
feat: add data zoom to metric visualization

joaoviana:
Closes: #12658 

### Description:

- creates visualization/ folder
- moves types to that folder, along with the component
- adds new hook useChartZoom which includes a bunch of handlers to be used in recharts charts

Demo

https://github.com/user-attachments/assets/b5cf908a-038c-4ea9-ac12-fd915ea10109

> [!NOTE]
> Git is not smart enough to detect a file that was moved when it contains significant changes. 

### Reviewer actions


packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
```diff
@@ -0,0 +1,248 @@
...
+            resetZoom,
+        },
+        activeData,
+    } = useChartZoom({
+        data: timeSeriesData || [],
+    });
+
+    useEffect(() => {
+        resetZoom();
+    }, [data.comparisonRows, resetZoom]);
```
IrakliJani, line 120: why only comparison rows and not both?

IrakliJani (commented):
approved code changes with one question

packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
```diff
@@ -0,0 +1,241 @@
...
+            handleMouseUp,
+            resetZoom,
+        },
+        activeData,
+    } = useChartZoom({
+        data: timeSeriesData || [],
+    });
+
+    useEffect(() => {
+        resetZoom();
```
joaoviana, line 112: now, every time the data changes, the zoom will reset @IrakliJani . Thank you for pointing it out. I feel like this is ok bc `data` is a stable reference, if I understand this correctly

IrakliJani: makes sense!

Issue #12635, opened 2024-11-28
Empty state:  Data is unavailable because the selected range goes too far back


PriPatel:
https://www.figma.com/design/AFhGBAKR4kreLZR5uqaSKE/Metrics-First?node-id=1088-3932&t=XIJjQ6RROwquK3G9-0
Same layout for the empty state where the user doesn't have a default time dimension defined. 

- [ ] Should affect users of all permissions
- [ ] Nice to have - a CTA to reset the range?

PR #12677, opened 2024-12-02
feat: show empty state if no data in range

joaoviana:
Closes: [#12635](https://github.com/lightdash/lightdash/issues/12635)

### Description:

- if the query is successful, but there's no data, then show an empty state
- did *not* do the reset-filters button logic - can be done later.

> [!NOTE]
> The main ticket specifically targets "...the selected range goes too far back" but there could also be no data for a specific range - like a very recent date? so chatted with @PriPatel and updated the copy on the subheading to be less specific.

Demo

image

### Reviewer actions


packages/frontend/src/features/metricsCatalog/components/MetricsVisualizationEmptyState.tsx
```diff
@@ -0,0 +1,41 @@
+import { Center, Paper, Stack, Text } from '@mantine/core';
+import { IconClockCancel } from '@tabler/icons-react';
+import MantineIcon from '../../../components/common/MantineIcon';
+import MetricsVisualizationEmptyStateImage from '../../../svgs/metricsCatalog/metrics-visualization-empty-state.svg';
+
+export const MetricsVisualizationEmptyState = () => {
+    return (
+        <Paper
```
IrakliJani, line 8: does it have composition with `component` prop? if so you can probably do it with <Paper component={Flex} or {Center}

IrakliJani (approved):
approving code changes. screenshot looks 🤌 <3

packages/frontend/src/features/metricsCatalog/components/MetricsVisualizationEmptyState.tsx
joaoviana: Ah, I tried this but it still requires a few more props even after composition, so keeping it like this 👍

PR #12643, opened 2024-11-28
feat: metric trees frontend save edges

almeidabbm:
Closes: #12641

⚠️ Based on: https://github.com/lightdash/lightdash/pull/12639

### Description:

- Creates edges on connect
- Deletes edges on backspace

https://github.com/user-attachments/assets/45578ac8-f1f9-4d7e-bc92-c3405fdaa005

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #12682, opened 2024-12-02
feat: limit metric node count

almeidabbm:
Closes: #12679 

⚠️ Depends on https://github.com/lightdash/lightdash/pull/12643

### Description:

- Adds limit of 30 metrics to display on the tree
- When there are more than 30 metrics (by changing filters) and user is on the tree, show suboptimal state

https://github.com/user-attachments/assets/64f8258e-5739-4439-9f79-27607403d138

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/CatalogModel/CatalogModel.ts
+++ b/packages/backend/src/models/CatalogModel/CatalogModel.ts
@@ -15,6 +15,7 @@ import {
     type CatalogItem,
     type CatalogItemWithTagUuids,
     type CatalogMetricsTreeEdge,
+    type CatalogMetricsTreeNode,
     type ChartUsageIn,
     type KnexPaginateArgs,
     type KnexPaginatedData,
@@ -697,11 +698,23 @@ export class CatalogModel {
 
     async getMetricsTree(
         projectUuid: string,
+        metricNodes: CatalogMetricsTreeNode[],
     ): Promise<{ edges: CatalogMetricsTreeEdge[] }> {
-        const edges = await this.database(MetricsTreeEdgesTableName).where(
-            'project_uuid',
-            projectUuid,
-        );
+        const metricNodeEntries = metricNodes.map((n) => [n.name, n.tableName]);
+        const edges = await this.database(MetricsTreeEdgesTableName)
+            .where('project_uuid', projectUuid)
+            .andWhere(function sourceNodeWhere() {
+                void this.whereIn(
+                    ['source_metric_name', 'source_metric_table_name'],
+                    metricNodeEntries,
+                );
+            })
+            .andWhere(function targetNodeWhere() {
+                void this.whereIn(
+                    ['target_metric_name', 'target_metric_table_name'],
+                    metricNodeEntries,
+                );
+            });
 
         return {
             edges: edges.map((e) => ({
--- a/packages/backend/src/services/CatalogService/CatalogService.ts
+++ b/packages/backend/src/services/CatalogService/CatalogService.ts
@@ -884,10 +884,10 @@ export class CatalogService<
     async getMetricsTree(
         user: SessionUser,
         projectUuid: string,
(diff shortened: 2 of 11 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricsTableTopToolbar.tsx
```diff
@@ -252,6 +235,17 @@ export const MetricsTableTopToolbar: FC<Props> = memo(
             FeatureFlags.MetricTrees,
         );
 
+        const isValidMetricsTree =
+            totalResults > 0 && totalResults <= MAX_METRICS_TREE_NODE_COUNT;
+
+        // If the metrics tree is not valid, set the view to list
+        // can happen when changing filters while in tree view
```
magnew, line 242: Good enough for now, but the experience is a little odd. You mess with filters and the view changes without warning. Something to think about.

magnew (approved):
Nice work!

✅ Works as expected
✅ Code looks good

I left a couple thoughts about UI, but they are more to think about as we move forward.

packages/frontend/src/features/metricsCatalog/components/MetricsTableTopToolbar.tsx
magnew: I mean something for all of us to think about 😄. Should we disable filters on the tree? It would be cool if they work, but at the moment itll almost always pop you out of the tree experience to mess with them. Unless you are being very careful to only narrow down.

packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -697,11 +698,23 @@ export class CatalogModel {
...
+        const metricNodeEntries = metricNodes.map((n) => [n.name, n.tableName]);
+        const edges = await this.database(MetricsTreeEdgesTableName)
+            .where('project_uuid', projectUuid)
+            .andWhere(function sourceNodeWhere() {
+                void this.whereIn(
+                    ['source_metric_name', 'source_metric_table_name'],
+                    metricNodeEntries,
+                );
+            })
+            .andWhere(function targetNodeWhere() {
```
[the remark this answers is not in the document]
almeidabbm: Yep, for now since we're sure that we'll always know the nodes displayed on the canvas we're just getting the edges containing source/target included in that set (to limit the amount of things we get in the frontend)

But it should be easy to change if we decide to have a more complex feature just like the one you mentioned 👌

packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -877,9 +879,23 @@ export class CatalogService<
...
+        // Using metricIds instead of filters and searching metrics because we might want to have the ability to select specific metrics rather than filter by catalog tags
+        metricIds: string[],
+    ) {
         // TODO: check permissions
-        return this.catalogModel.getMetricsTree(projectUuid);
+        if (metricIds.length > MAX_METRICS_TREE_NODE_COUNT) {
+            throw new ParameterError(
+                `Cannot get more than ${MAX_METRICS_TREE_NODE_COUNT} metrics in the metrics tree`,
+            );
+        }
```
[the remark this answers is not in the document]
almeidabbm: No this is just a safeguard in the backend, the frontend is preventing this error from happening by disabling the UI component

packages/frontend/src/features/metricsCatalog/components/MetricsTableTopToolbar.tsx
almeidabbm: I completely agree, I was going to leave this as is until we have a better experience but this can be annoying, I will try to use the suboptimal state component, hopefully it will work fine here 🙏

magnew: Much better!

Issue #12667, opened 2024-11-29
I should still be able to use the metrics explorer if I dont have any defaults defined in the .yml


PriPatel:
### Which time-based dimension to use

- [ ] Logic in order of priority: 

1. If a date dimension has been defined in the .yml on the **metric** level, use that. IF NOT; 
2. If a date dimension has been defined in the .yml on the **model** level, use that. IF NOT; 
3. If there is only one time dimension in the table, use that. IF NOT;
4. If the table is joined and the joined tables have time dimensions show a dropdown with the date dims of the joined tables.  IF NOT;
5. There is a dropdown where you can select which time dimension to use to a chart plotted. (this dropdown will include all time dimensions available in the table). This input will be called 'X-axis'

--

- [ ] If 1,2,3 above, do not show the 'X-axis' input in the sidebar. Instead, show the dimension name on hover of the X-axis label

**OPEN QUESTION**
- [ ]  If 4/5 above, the first thing the user will see is everything apart from the input. Maybe theres an animated blinker here? Something we can pair on. Im not sure how greyed out the chart needs to be.

Issue #12659, opened 2024-11-29
Date Zoom (intervals / granularities)


PriPatel:
### Frontend
- [x] Add a dropdown/pill next the time filter where the user can update granularity (`daily` `weekly` `monthly` `yearly` etc)
- [x] Behaviour should be like dashboard 'date zoom'

### Calendar picker / filters should change config depending on the selected granularity. 
https://miro.com/app/board/uXjVL-eF9po=/
We can have a discussion about this one

PriPatel:
@joaoviana 
Two updates here: 

- Can you make this a dropdown instead of a pill inside the other input. 
- Can you change copy to daily , weekly , monthly , yearly

PR #12701, opened 2024-12-03
feat: render metric visualization even if no default & config

joaoviana:
Closes: [#12667](https://github.com/lightdash/lightdash/issues/12667)

### Description:

- Generates a default time dimension for a metric if defined on table yml - test with `Average order size`
- metric yml - test with `Total order amount`
- if the table only has ONE date dim - test with `Unique customer count`

If no defaults available ⬆️ , we then get the available time dimensions for that metric and allow the changing of the field and/or interval - you can test this with anything in `payments` table, `events` table etc

demo:

https://github.com/user-attachments/assets/d5c669d7-30ef-417d-af62-41eb61e22769

> [!IMPORTANT]
> The granularity/interval changes the chart as it should, as well as the filtered range - however, the filter component is not in sync with the change. I want to tackle this on this ticket: https://github.com/lightdash/lightdash/issues/12659

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/metricsExplorerController.ts
+++ b/packages/backend/src/controllers/metricsExplorerController.ts
@@ -1,11 +1,13 @@
 import {
     ApiErrorPayload,
     MetricExplorerComparison,
+    TimeDimensionConfig,
     type ApiMetricsExplorerQueryResults,
     type MetricExplorerComparisonType,
     type MetricExplorerDateRange,
 } from '@lightdash/common';
 import {
+    Body,
     Middlewares,
     OperationId,
     Path,
@@ -45,6 +47,10 @@ export class MetricsExplorerController extends BaseController {
         @Query() endDate: string,
         @Query() compareToPreviousPeriod?: boolean,
         @Query() compareToMetric?: string,
+        @Body()
+        body?: {
+            timeDimensionOverride?: TimeDimensionConfig;
+        },
     ): Promise<ApiMetricsExplorerQueryResults> {
         this.setStatus(200);
 
@@ -79,6 +85,7 @@ export class MetricsExplorerController extends BaseController {
                 metric,
                 dateRange,
                 compare,
+                body?.timeDimensionOverride,
             );
 
         return {
--- a/packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
+++ b/packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
@@ -1,12 +1,12 @@
 import { subject } from '@casl/ability';
 import {
     assertUnreachable,
-    CompiledMetric,
     ForbiddenError,
     getFieldIdForDateDimension,
     getGrainForDateRange,
     getItemId,
     getMetricExplorerDateRangeFilters,
+    getTimeDimensionConfig,
     MetricExplorerComparison,
     MetricExplorerComparisonType,
     oneYearBack,
@@ -15,6 +15,7 @@ import {
     type MetricsExplorerQueryResults,
     type ResultRow,
     type SessionUser,
+    type TimeDimensionConfig,
 } from '@lightdash/common';
 import { v4 as uuidv4 } from 'uuid';
 import type { LightdashConfig } from '../../config/parseConfig';
@@ -92,8 +93,7 @@ export class MetricsExplorerService<
         const { defaultTimeDimension } = metric;
 
         const timeDimensionConfig =
-            timeDimensionOverride ??
-            MetricsExplorerService.getTimeDimensionConfig(metric);
+            timeDimensionOverride ?? getTimeDimensionConfig(metric);
 
         if (!timeDimensionConfig) {
             throw new Error(
@@ -203,38 +203,6 @@ export class MetricsExplorerService<
             rows: currentResults,
             comparisonRows: comparisonResults,
             fields: allFields,
-        };
-    }
-
-    /**
-     * Helper method to determine the correct time dimension configuration
-     * It covers the case where the metric has a default time dimension
-     * and the case where the metric has available time dimensions (could be multiple and from different tables that are joined)
-     */
-    private static getTimeDimensionConfig(
-        metric: CompiledMetric,
-    ): TimeDimensionConfig | undefined {
-        if (metric.defaultTimeDimension) {
-            // Use the metric's table when default time dimension is specified
-            return {
-                table: metric.table,
-                field: metric.defaultTimeDimension.field,
-                interval: metric.defaultTimeDimension.interval,
-            };
-        }
-
-        // Fall back to the first available time dimension
-        const firstAvailableTimeDimension =
-            getFirstAvailableTimeDimension(metric);
-        if (!firstAvailableTimeDimension) {
-            return undefined;
-        }
-
-        // Use the dimension's own table in case it's joined
-        return {
-            table: firstAvailableTimeDimension.table,
-            field: firstAvailableTimeDimension.field,
-            interval: firstAvailableTimeDimension.interval,
         };
     }
 }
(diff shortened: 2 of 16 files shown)


packages/backend/src/controllers/metricsExplorerController.ts
```diff
@@ -45,6 +47,10 @@ export class MetricsExplorerController extends BaseController {
         @Query() compareToMetric?: string,
         @Query() startDate?: string,
         @Query() endDate?: string,
+        @Body()
+        body?: {
+            timeDimensionOverride?: TimeDimensionConfig;
+        },
```
joaoviana, line 53: body or 3 separate query params instead?

packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
```diff
@@ -218,4 +226,36 @@ export class MetricsExplorerService<
             fields: allFields,
         };
     }
+
+    /**
+     * Helper method to determine the correct time dimension configuration
+     * It covers the case where the metric has a default time dimension
+     * and the case where the metric has available time dimensions (could be multiple and from different tables that are joined)
+     */
+    private static getTimeDimensionConfig(
```
joaoviana, line 235: I was thinking of moving this to the catalogService but maybe not - seems tied to this more. 
Would a util for this, and the other 2 utils in CatalogService be a better placement? As in, a separate file?

packages/frontend/src/features/metricsCatalog/components/MetricPeekModal.tsx
```diff
@@ -84,21 +88,48 @@ export const MetricPeekModal: FC<Props> = ({ opened, onClose }) => {
...
     const metricResultsQuery = useRunMetricExplorerQuery({
         projectUuid,
         exploreName: tableName,
         metricName,
         comparison: comparisonParams,
         dateRange,
+        timeDimensionOverride,
     });
 
+    const timeDimensionBaseField: TimeDimensionConfig | undefined =
```
joaoviana, line 103: this is so I can get the field's field name, interval and table so some things can be manipulated: 
- field 
- interval/granularity
the `table` is so that we can query from joined fields

packages/frontend/src/features/metricsCatalog/components/MetricPeekModal.tsx
```diff
@@ -149,6 +180,41 @@ export const MetricPeekModal: FC<Props> = ({ opened, onClose }) => {
                 >
                     <Stack p="xl" bg="offWhite.0" miw={360}>
                         <Stack spacing="xl">
+                            {metricQuery.data?.availableTimeDimensions && (
```
joaoviana, line 183: user can only pick from date dims if there are available time dimensions - and these are only defined **if there is no default time dimension**

packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
IrakliJani: this can be inside metric explorer utils (common) and maybe reused in both catalog and here?

packages/common/src/types/catalog.ts
```diff
@@ -104,7 +104,11 @@ export type ApiMetricsCatalog = {
 };
 
 export type MetricWithAssociatedTimeDimension = CompiledMetric &
-    Pick<CompiledTable, 'defaultTimeDimension'>;
+    Pick<CompiledTable, 'defaultTimeDimension'> & {
+        availableTimeDimensions?: (CompiledDimension & {
```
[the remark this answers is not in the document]
IrakliJani: looks good!

packages/frontend/src/features/metricsCatalog/components/visualization/TimeDimensionPicker.tsx
```diff
@@ -0,0 +1,99 @@
...
+                        backgroundColor: theme.colors.gray[2],
+                    },
+                    '&[data-selected="true"]:hover': {
+                        backgroundColor: theme.colors.gray[3],
+                    },
+                    '&:hover': {
+                        backgroundColor: theme.colors.gray[1],
+                    },
+                },
+            }}
```
IrakliJani, line 95: should this be a style hook?

packages/frontend/src/features/metricsCatalog/hooks/useRunMetricExplorerQuery.ts
```diff
@@ -60,7 +63,11 @@ const postRunMetricExplorerQuery = async ({
             queryString ? `?${queryString}` : ''
         }`,
         method: 'POST',
-        body: undefined,
+        body: timeDimensionOverride
+            ? JSON.stringify({
+                  timeDimensionOverride,
+              })
+            : undefined,
```
IrakliJani, line 70: Oh, I see. Okay, maybe we should include a dimension id override with the grain as part of the string, and then just use it in the backend.

So far, what I see is that `TimeDimensionConfig` includes the name of the table, the field ID, and the grain.

This can be combined into a single field ID before making the request.

The override can be named `timeDimensionId` or `timeDimensionIdOverride` as a query parameter.

wdyt?

packages/frontend/src/features/metricsCatalog/components/visualization/TimeDimensionPicker.tsx
IrakliJani: very low priorityt

joaoviana: yes it should but not doing on this PR if that's ok!

packages/frontend/src/features/metricsCatalog/hooks/useRunMetricExplorerQuery.ts
joaoviana: so it would be like
'orders_order_id_month'?

joaoviana: good shout, on it

packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
joaoviana: [refactor: move utils to metricsExplorer util](https://github.com/lightdash/lightdash/pull/12701/commits/ded6c6b9d70703fa878a7127d1714802016f3a42)

joaoviana:
- always set a `timeDimension` for a metric and utilise it wherever possible
- metrics explorer query returns metric as well now and includes the time dimension (which can be overriden)
- date range in date picker is controlled now. Every time we change the granularity, it will clear the preset if selected and apply the default range for that granularity. this will be refined in: https://github.com/orgs/lightdash/projects/11/views/1?pane=issue&itemId=89572515&issue=lightdash%7Clightdash%7C12659

packages/backend/src/controllers/metricsExplorerController.ts
IrakliJani: I take my words back, body is fine

packages/frontend/src/features/metricsCatalog/components/MetricPeekModal.tsx
```diff
@@ -95,14 +83,57 @@ export const MetricPeekModal: FC<Props> = ({ opened, onClose }) => {
...
+
+        if (!timeDimension) return;
+
+        if (timeDimension.interval !== timeDimensionBaseField?.interval) {
+            setDateRange(
+                getDefaultDateRangeFromInterval(timeDimension.interval),
+            );
+        }
+    }, [timeDimensionOverride, metricQuery.data, timeDimensionBaseField]);
+
```
IrakliJani, line 136: okay this moved here, nice!

IrakliJani (approved):

Issue #12676, opened 2024-12-02
Compare to another metric


IrakliJani:
### Compare to another metric
- [ ] You can compare your metric to **another metric**
- [ ] You cannot compare to last year AND another metric. Only one type of comparison is allowed.
- [ ]  **Which metrics can you compare to?** If there are no dimensions selected (i.e. only a metric over time) then you can compare literally any other metric (as long as that metric also has a time grain available).
- [ ] Once the user clicks on the new metric, it is also plotted on the chart. 
- [ ] User can easily clear this comparison

PR #12703, opened 2024-12-03
feat: metric explorer compare with other metric - part 1

IrakliJani:
Part 1 of: #12676 

### Description:

CleanShot 2024-12-03 at 23 50 35@2x

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/metricsExplorer.ts
+++ b/packages/common/src/utils/metricsExplorer.ts
@@ -15,9 +15,11 @@ import {
     type FieldTarget,
     type FilterRule,
 } from '../types/filter';
-import type {
-    MetricExploreDataPoint,
-    MetricExplorerDateRange,
+import {
+    MetricExplorerComparison,
+    type MetricExploreDataPoint,
+    type MetricExplorerComparisonType,
+    type MetricExplorerDateRange,
 } from '../types/metricsExplorer';
 import type { ResultRow } from '../types/results';
 import { TimeFrames, type DefaultTimeDimension } from '../types/timeFrames';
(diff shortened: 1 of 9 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricPeekModal.tsx
```diff
@@ -73,22 +85,21 @@ export const MetricPeekModal: FC<Props> = ({ opened, onClose }) => {
             case MetricExplorerComparison.DIFFERENT_METRIC:
                 return {
                     type: MetricExplorerComparison.DIFFERENT_METRIC,
-                    metricName: metricQuery.data.name,
+                    // TODO: this is hardcoded for now, should be a dropdown in the UI
+                    metricTable: 'orders',
+                    metricName: 'total_non_completed_order_amount',
```
IrakliJani, line 79: upcoming work for part 2 (ui)

joaoviana: and I'm assuming that the legends + tooltips will be addressed in part 2, right?

packages/common/src/utils/metricsExplorer.ts
```diff
@@ -185,20 +191,31 @@ export const getMetricExplorerDataPointsWithCompare = (
...
         ...Object.keys(offsetGroupByCompareMetricRows),
     ]);
 
+    const compareMetricId =
+        comparison.type === MetricExplorerComparison.PREVIOUS_PERIOD
+            ? metricId
+            : getItemId({
+                  table: comparison.metricTable,
+                  name: comparison.metricName,
+              });
```
joaoviana, line 211: maybe we can be more explicit with the enum for this case? so it's not the _default_ if we add more 

can be part 2

packages/frontend/src/features/metricsCatalog/components/MetricPeekModal.tsx
IrakliJani: yes

packages/common/src/utils/metricsExplorer.ts
IrakliJani: +

Issue #12671, opened 2024-12-02
Changing to `year` granularity in date zoom on a dashboard breaks my chart


TuringLovesDeathMetal:
### Description

If the charts in my dashboard have data only from one year (e.g. from January 2024 - June 2024), and I change the date zoom to be `year`, then the chart says `no data available`. 

If the data spans more than a year (e.g. from December 2023 - January 2024), and I change the date zoom to be `year`, then the charts work. 

### Steps to Reproduce the Bug or Issue

1. go to a dashboard 
2. add a date filter for some time period that only covers a year (e.g. in the last 3 weeks) 
3. change the date zoom to `year`
4. see that all of the charts break. 

### Cloud or self-hosting

None

PR #12711, opened 2024-12-04
fix: add better handling for year granularity in dashboard filter

andrewdw:
Closes: #12671

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -97,6 +97,7 @@ import {
     SummaryExplore,
     TablesConfiguration,
     TableSelectionType,
+    TimeFrames,
     UnexpectedServerError,
     UpdateMetadata,
     UpdateProject,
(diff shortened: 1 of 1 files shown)


andrewdw:
Having trouble duplicating locally in jaffle. Generated 1000 events between 2020 and now and can't get it to break

```
INSERT INTO "jaffle"."events" (event_id, timestamp_tz, timestamp_ntz, timestamp_ltz, date, event)
SELECT
    300 + row_number() OVER (), -- Event IDs starting from 301
    ts,                        -- Random timestamp with timezone
    ts,                        -- Random timestamp without timezone (same as tz for simplicity)
    ts,                        -- Random timestamp (localized)
    ts::date,                  -- Date part of the timestamp
    events.event_type          -- Random event type
FROM (
    SELECT 
        generate_series(
            '2019-01-01 00:00:00'::timestamp, 
            now()::timestamp, 
            '1 second'::interval
        ) AS ts
) ts_series
CROSS JOIN (
    SELECT unnest(ARRAY['song_played', 'playlist_created', 'playlist_deleted']) AS event_type
) events
ORDER BY random() -- Shuffle rows
LIMIT 1000;

```

PR #12758, opened 2024-12-06
fix: check if project has saved charts

ZeRego:
Closes: 

### Description:

Changes:
- remove unnecessary call in home page
- change request in empty dashboard to check if project has charts
- update endpoint that checks if project has charts to only request 1 chart from the database

Demo: new request working when project has charts or not.

Screenshot 2024-12-06 at 10 19 34

Screenshot 2024-12-06 at 10 20 17

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -10,8 +10,10 @@ import {
     assertUnreachable,
     CacheMetadata,
     CalculateTotalFromQuery,
+    ChartSourceType,
     ChartSummary,
     CompiledDimension,
+    ContentType,
     convertCustomMetricToDbt,
     countCustomDimensionsInMetricQuery,
     countTotalFilterRules,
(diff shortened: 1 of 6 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3627,8 +3635,20 @@ export class ProjectService extends BaseService {
...
         }
         try {
-            const charts = await this.savedChartModel.find({ projectUuid });
-            return charts.length > 0;
+            const charts = await this.contentModel.findSummaryContents(
+                {
+                    projectUuids: [projectUuid],
+                    contentTypes: [ContentType.CHART],
+                    chart: {
+                        sources: [ChartSourceType.DBT_EXPLORE],
```
almeidabbm, line 3643: Previously we also only checked for charts in `SavedChartModel` so this is the same check but we can change this to be looking for all `ChartSourceType` since they can all be added to dashboards (sql_runner, semantic_viewer)

almeidabbm (approved):

Issue #10514, opened 2024-06-25
Lightdash should cache the filter autocomplete values


TuringLovesDeathMetal:
### Description

Right now, when a user tries to add a filter and the tables are really big, the filter autocomplete is really slow. 

To help with this, we should cache teh filter autocomplete values. 

We would need some way to refresh the cache here.

PR #12780, opened 2024-12-09
feat: cache results in filters

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/10514

### Description:

image

[Screencast from 2024-12-09 13-21-55.webm](https://github.com/user-attachments/assets/8ee4c493-8e44-443f-ba61-f1e2b60759a4)

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -39,6 +39,7 @@ import {
     Explore,
     ExploreError,
     ExploreType,
+    FieldValueSearchResult,
     FilterableDimension,
     FilterGroupItem,
     FilterOperator,
@@ -179,6 +180,11 @@ type RunQueryTags = {
     chart_uuid?: string;
     dashboard_uuid?: string;
 };
+
+const cachedUniqueSearchResults: Record<
+    string,
+    FieldValueSearchResult<string>
+> = {};
 
 type ProjectServiceArguments = {
     lightdashConfig: LightdashConfig;
@@ -2476,7 +2482,8 @@ export class ProjectService extends BaseService {
         search: string,
         limit: number,
         filters: AndFilterGroup | undefined,
-    ): Promise<Array<unknown>> {
+        forceRefresh: boolean,
+    ) {
         const { organizationUuid } = await this.projectModel.getSummary(
             projectUuid,
         );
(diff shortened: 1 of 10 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2490,6 +2497,16 @@ export class ProjectService extends BaseService {
...
         }
 
+        const cachekey = `${projectUuid}-${table}-${initialFieldId}-${search}-${JSON.stringify(
+            filters,
+        )}`;
+        if (
+            !forceRefresh &&
+            cachedUniqueSearchResults[cachekey] !== undefined
+        ) {
+            return cachedUniqueSearchResults[cachekey];
```
rephus, line 2507: refresh cache idea: 
if refreshedAt is too old, we could return whatever is in the cache, and refresh the cache in the background, for the next time this is called.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2490,6 +2497,16 @@ export class ProjectService extends BaseService {
             throw new ForbiddenError();
         }
 
+        const cachekey = `${projectUuid}-${table}-${initialFieldId}-${search}-${JSON.stringify(
+            filters,
+        )}`;
+        if (
```
rephus, line 2503: Should we add an environment variable to control this feature ? 
I don't want to use `feature flags` in the backend.

packages/frontend/src/hooks/useFieldValues.ts
```diff
@@ -87,14 +92,23 @@ export const useFieldValues = (
         },
         [filters, initialData],
     );
+    const cachekey = [
+        'project',
+        projectId,
+        tableName,
+        fieldName,
+        'search',
+        debouncedSearch,
```
rephus, line 101: extracted cacheKey in a separate variable, but I didn't change it. We don't need to include forceRefresh here since we'll be using `refetch`
Note: Adding forceRefresh here will trigger 2 queries (one with true, and another with false)

packages/common/src/types/fieldMatch.ts
```diff
@@ -1,4 +1,6 @@
 export type FieldValueSearchResult<T = unknown> = {
     search: string;
     results: T[];
+    cached: boolean;
```
rephus, line 4: added new properties to this type.
This type is only used for this `filter search`

rephus:
UI suggestion: 

Try pinning the `cache message` on the dropdown to the bottom

rephus:
TODO: fix user attributes

owlas (approved):

PR #12792, opened 2024-12-09
perf: get metrics with batched attribute-filter

joaoviana:
Closes: #12772

### Description:

Adds `getMetrics` service method to get all metrics in batch - allows receiving user attributes to fix a n+1 query
`getMetric` calls `getMetrics` with just one metric in the array

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CatalogService/CatalogService.ts
+++ b/packages/backend/src/services/CatalogService/CatalogService.ts
@@ -1035,6 +1035,10 @@ export class CatalogService<
             timeIntervalOverride,
         );
 
+        if (metrics.length === 0) {
+            throw new NotFoundError('Metric not found');
+        }
+
         return metrics[0];
     }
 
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -915,82 +919,123 @@ export class CatalogService<
...
+        timeIntervalOverride?: TimeFrames,
+    ) {
+        const metrics = await this.getMetrics(
+            user,
+            projectUuid,
+            [{ tableName, metricName }],
+            timeIntervalOverride,
+        );
+
+        return metrics[0];
```
almeidabbm, line 1038: We should check if metric exists and throw error when it doesn't. Since we're using the `getMetrics` function, ts type inference doesn't show it can be `undefined`

almeidabbm (approved):

packages/backend/src/services/CatalogService/CatalogService.ts
joaoviana: Good point, done here! https://github.com/lightdash/lightdash/pull/12792/commits/6c358ae17ac4daf8d9b9e4f912726806e390925b

Issue #12810, opened 2024-12-10
Series config is not kept when sharing url


ZeRego:
### Description

Series config is not kept when sharing url to unsaved query

### Steps to Reproduce the Bug or Issue

https://github.com/user-attachments/assets/e6db1cc4-4603-47c9-af71-936eb587ddc1

### Cloud or self-hosting

cloud

PR #12811, opened 2024-12-10
fix: share unsaved chart config

ZeRego:
Closes: [#12810](https://github.com/lightdash/lightdash/issues/12810)

### Description:

Generate URL with entire chart config rather than using current URL which removes chart config to keep URL valid.

Same fix as https://github.com/lightdash/lightdash/pull/12400

Sharing from explore

https://github.com/user-attachments/assets/86e6e9bd-cc0c-45f6-9e59-748c0979f78c

### Reviewer actions


magnew (approved):
🎉

Issue #12789, opened 2024-12-09
fix user attributes on cached filter search


rephus:
Cached filter results should not be shared if there are user attibutes applied to a field. 

_Originally posted by @rephus in https://github.com/lightdash/lightdash/issues/12780#issuecomment-2528097602_

PR #12821, opened 2024-12-11
fix: improve backend autocomplete

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12789

## Changes: 

- cache entire hashed query, this includes user attributes (if any) . Note, this is secure, but less efficient,s ince we need to get all metadata from explore and metricquery to compute the query to get the cache key to know if it is cached or not;
- use s3cacheClient to store results in s3, rather than in memory
- change in env variables  

### Description:

Same UI: 

[Screencast from 2024-12-11 10-04-29.webm](https://github.com/user-attachments/assets/fdbf3eaa-9837-4844-8755-db19237d5651)

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -181,11 +181,6 @@ type RunQueryTags = {
     dashboard_uuid?: string;
 };
 
-const cachedUniqueSearchResults: Record<
-    string,
-    FieldValueSearchResult<string>
-> = {};
-
 type ProjectServiceArguments = {
     lightdashConfig: LightdashConfig;
     analytics: LightdashAnalytics;
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2679,7 +2679,12 @@ export class ProjectService extends BaseService {
             refreshedAt: new Date(),
         };
         if (isCacheEnabled) {
-            const buffer = Buffer.from(JSON.stringify(searchResults));
+            const buffer = Buffer.from(
+                JSON.stringify({
+                    ...searchResults,
+                    cached: true,
+                }),
+            );
             // fire and forget
             this.s3CacheClient
                 .uploadResults(queryHash, buffer, queryTags)
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -2598,10 +2598,10 @@ export class ProjectService extends BaseService {
             userAttributes,
             this.lightdashConfig.query.timezone || 'UTC',
         );
-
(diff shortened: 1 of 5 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2616,6 +2599,42 @@ export class ProjectService extends BaseService {
...
+            .createHash('sha256')
+            .update(queryHashKey)
+            .digest('hex');
+
+        const isCacheEnabled =
+            this.lightdashConfig.resultsCache.autocompleteEnabled &&
+            this.s3CacheClient.isEnabled();
+
+        if (!forceRefresh && isCacheEnabled) {
+            const isCached = await this.s3CacheClient.getResultsMetadata(
```
rephus, line 2615: getResultsMetadata does not throw an error if results does not exist, so this needs to be called before getResults. It works like `hasResults`

packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.tsx
```diff
@@ -161,7 +161,7 @@ const FilterStringAutoComplete: FC<Props> = ({
                     <Tooltip
                         withinPortal
                         position="left"
-                        label={`Click here to cache results now`}
+                        label={`Click here to refresh cache filter values`}
```
rephus, line 164: quick small copy change suggested by Jake. More UI changes will be added on a separate PR

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2616,6 +2599,42 @@ export class ProjectService extends BaseService {
             this.lightdashConfig.query.timezone || 'UTC',
         );
 
+        const queryHashKey = metricQuery.timezone
```
rephus, line 2602: same hashkey logic as results

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2615,6 +2598,42 @@ export class ProjectService extends BaseService {
...
+
+            if (isCached !== undefined) {
+                const cacheEntry = await this.s3CacheClient.getResults(
+                    queryHash,
+                );
+                const stringResults =
+                    await cacheEntry.Body?.transformToString();
+                if (stringResults) {
+                    try {
+                        return JSON.parse(stringResults);
```
ZeRego, line 2645: Response doesn't have`cached: true,` anymore. is that expected?

rephus: When we upload the results to s3, we include all the metadata , like cached, or refreshedAt. When  fetching from s3, we simply return the same object we previously store. 

To avoid conflicts with normal cached results,  thes are stored in a different cacheKey

ZeRego (approved):

Issue #12854, opened 2024-12-12
Add `selector` for charts and dashboards to `upload` and `download` commands. 


TuringLovesDeathMetal:
### Description

I want to be able to select specific charts and dashboards to download/upload as code from the CLI (instead of all content). 

### Acceptance criteria

- [ ] There's a way to select specific charts from the CLI using `lightdash download --charts my-chart-slug`
- [ ] you can select charts using their UUID, the share URL, or the SLUG
- [ ] You can only download and upload saved charts (e.g. you can't copy a URL from an unsaved chart and try to download it as code)
- [ ] You can select many charts in a row by separating their selection with a space (see example shell below)
- [ ] This selection works for `lightdash download` and `lightdash upload`

- [ ] There's a way to select specific dashboards fromt eh CLI using `lightdash download --dashboards my-dashboard-slug`
- [ ] you can select dashboards using their UUID, the share URL, or the SLUG
- [ ] You can select many dashboards in a row by separating their selection with a space (see example shell below)
- [ ] This selection works for `lightdash download` and `lightdash upload`

```bash
# download a list of charts `revenue-daily`, `total-sales-weekly` and `orders-in-emea`
$ lightdash download --charts (or -c) revenue-daily total-sales-weekly orders-in-emea

# download a list of dashboards `kpi-dashboard` and `customer-success-overview`
$ lightdash download --dashboards (or -d) kpi-dashboard customer-success-overview

# download content from the `finance` and `marketing` spaces, and also a 
# dashboard called `kpi-dashboard`
$ lightdash download --spaces finance marketing --dashboards kpi-dashboard
```

PR #12872, opened 2024-12-12
feat: chart selector as code

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12854

### Description:

- This introduces a breaking change in the API (they neecd to upgrade their CLI) local files are still compatible, as now we return `missingIds` in results 

Download: 

image

Another example: 
image

Upload: 
image

#### on download, if any filter is set, we don't donwload anything else 

image

#### on upload, if a filter is specified but the file didn't have local changes, we show a warning 
image

### Reviewer actions


magnew (approved):
🚀 This is super cool. Approved, with a couple observations

✅ Code makes sense
✅ Can download with slug filter
✅ Can download with UUID
✅ Can download with url
✅ Can upload by slug

❓ When I upload a chart by slug twice in a row with no changes, I don't see the no changes warning. Am I doing this wrong? It also says it will update the chart.
(also, this line should be 'updated' instead of 'update' to match 'skipped')
Screenshot 2024-12-16 at 14 05 23

🤔 I don't know if there is anything we can do about this, but I got tripped a little by the behavior when no `-c` or `-d` is specified, but you try to use a fitler, like
`root@50b0ff10e176:/usr/app# yarn cli-build && ./packages/cli/dist/index.js upload f8`

It treats it as no filter, which is reasonable, but it feels like an easy way to accidentally update everything when you only meant to update one thing. Maybe we should fail if an arg is provided but no `-c` or `-d`? Could be a future UX improvement.

Issue #12886, opened 2024-12-13
I shouldnt be able to click on the metric name (for everyone!)


PriPatel:
@almeidabbm we're just going to hide it for everyone 🗡

PR #12891, opened 2024-12-13
chore: disable opening old explore on metric name click (temp)

almeidabbm:
Closes: [#12886](https://github.com/lightdash/lightdash/issues/12886)

### Description:

- Deletes the code for opening the old explore on metric name click, not sure if we will bring it back at some point

### Reviewer actions


packages/frontend/src/features/metricsCatalog/components/MetricsCatalogColumnName.tsx
```diff
@@ -298,7 +296,12 @@ export const MetricsCatalogColumnName = forwardRef<HTMLDivElement, Props>(
                         variant="xs"
                         openDelay={300}
                     >
-                        <UnstyledButton onClick={handleMetricClick}>
+                        <UnstyledButton
```
joaoviana, line 299: I think we should just remove all of this code - we can reference this PR later 👍 
And maybe let's not keep the UnstyledButton - it does nothing

IrakliJani (approved):
> Comments out the code for opening
I only see removals but its all good 😆 ship it!

PR #12900, opened 2024-12-13
fix: only set border bottom to none when theres scroll and is last row

almeidabbm:
Closes: 

### Description:

- when there's no scroll it means that there won't be overlap with footer, so we don't set border bottom to `none`
image

- when there's scroll, we set it to `none` so there's no double borders
image

### Reviewer actions


IrakliJani (approved):
oh that's very smart. well done!

Issue #12829, opened 2024-12-11
Share SQL Runner button doesn't contain chart configuration


ZeRego:
### Description

If I share a table chart from the SQL runner, everything loads well.
If I share a cartesian chart from the SQL runner, the chart doesn't load.

### Steps to Reproduce the Bug or Issue

https://github.com/user-attachments/assets/40634be6-eab3-4565-b8fe-ae2c85e5c847

### Cloud or self-hosting

cloud

TuringLovesDeathMetal:
we need to auto-run the query when you land on that URL! (maybe)

PR #12918, opened 2024-12-16
fix: SQL runner share url

ZeRego:
Closes: [#12829](https://github.com/lightdash/lightdash/issues/12829) 

### Description:

Problems:
- we were storing results and other state properties that should not be part of a share url
- we had to rerun the query on load
- we were not storing the viz config in the share url

Changes:
- move all logic around share URL to hooks
- only save relevant properties in share url
- save viz config in share url
- handle legacy URLs

https://github.com/user-attachments/assets/3ccb5ea1-7678-427b-bb86-b8f94c8b6234

### Reviewer actions


magnew (approved):
👍 Nice! Works great

Issue #12856, opened 2024-12-12
Optimize methods for fetching charts/dashboards on  `lightdash download`


TuringLovesDeathMetal:
### Description

We already have a method specific for charts/download as code, but this method is just fetching all full charts/dashboards in parallel (in Promise.all) . We should update the query to fetch all objects in a single query instead. 

Upload is ok

PR #12919, opened 2024-12-16
fix: content as code download in batches

rephus:
Closes: #12856

### Description:

[Screencast from 2024-12-17 09-23-58.webm](https://github.com/user-attachments/assets/96a20ce6-c8a4-473e-a7b1-18765778fe87)

I tried stress loading this a bit, it seems to be working as expected 

We can increase the max concurrent charts if we like. 

Screenshot from 2024-12-16 16-51-53

178 charts + 39 dashboards takes 2 seconds to download in batches 

image

### Load tests

I created thousand of charts and dashboards, some charts were charts within dashboards

I allowed 1000 simultaneous charts (aka 1000 queries to the database) 

The old listing of charts and dashboards in the UI breaks with a few hundreds

Some charts were as big as 119kb (4351 lines) 

Memory increased  a small amount per call 

image

Command took 12 seconds 

image

Conclusion: 1000 simultaneous requests are manageable, however, I think having a limit of 100 charts per query makes sense. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/config/lightdashConfig.mock.ts
+++ b/packages/backend/src/config/lightdashConfig.mock.ts
@@ -163,4 +163,7 @@ export const lightdashConfigMock: LightdashConfig = {
     headlessBrowser: {
         internalLightdashHost: 'https://test.lightdash.cloud',
     },
+    contentAsCode: {
+        maxDownloads: 100,
+    },
 };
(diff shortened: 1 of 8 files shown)


magnew (approved):
> 178 charts + 39 dashboards takes 2 seconds to download in batches
This seems pretty good. 

Code make sense. I didn't stress test this, but I can if you think it would be helpful. 

To me, this is a good compromise given the constraints.

magnew (approved):
Spinner 😵‍💫 !

Issue #12939, opened 2024-12-18
Add feature flag endpoint


ZeRego:
### Description

At the moment, to control access to certain features, we rely on Posthog and env vars.
The env vars are available to the frontend via the health endpoint and posthog via their library. 

New endpoint:
```GET /api/v1/feature-flag/{identifier}```

This endpoint has authentication as optional since some feature flags are used before users login.

### What problem does this solve?

Browser/extension blocking requests to posthog disabling features to users.
Devx figuring out where the feature flag is.
Self-hosted users can't enable features because they don't use posthog.

PR #12947, opened 2024-12-18
chore: feature flag endpoint

ZeRego:
Closes: [#12939](https://github.com/lightdash/lightdash/issues/12939) 

### Description:

Centralize feature flag source logic in the Model so it can be used by all services. In the near future, we will also have a table in the DB to control feature flags at the org/user level.

Changes:
- create feature flag controller/service/model
- move user groups feature flag logic to new model
- remove users group flag from health endpoint
- update frontend to use new hook/endpoint

## Demo

`GROUPS_ENABLED=false` + false Posthog feature flag

Screenshot 2024-12-18 at 13 27 28

`GROUPS_ENABLED=true` + false Posthog feature flag
OR
`GROUPS_ENABLED=false` + true Posthog feature flag
Screenshot 2024-12-18 at 13 24 49

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/v2/FeatureFlagController.ts
+++ b/packages/backend/src/controllers/v2/FeatureFlagController.ts
@@ -1,0 +1,41 @@
+import { ApiErrorPayload, FeatureFlag } from '@lightdash/common';
+import {
+    Get,
+    OperationId,
+    Path,
+    Request,
+    Response,
+    Route,
+    SuccessResponse,
+    Tags,
+} from '@tsoa/runtime';
+import express from 'express';
+import { BaseController } from '../baseController';
+
+@Route('/api/v2/feature-flag')
+@Response<ApiErrorPayload>('default', 'Error')
+@Tags('v2', 'Feature Flag')
+export class FeatureFlagController extends BaseController {
+    /**
+     * Get feature flag
+     */
+    @SuccessResponse('200', 'Success')
+    @Get('/{featureFlagId}')
+    @OperationId('Get feature flag')
+    async getFeatureFlag(
+        @Request() req: express.Request,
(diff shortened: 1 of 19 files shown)


packages/backend/src/controllers/v2/FeatureFlagController.ts
```diff
@@ -0,0 +1,41 @@
...
+export class FeatureFlagController extends BaseController {
+    /**
+     * Get feature flag
+     */
+    @SuccessResponse('200', 'Success')
+    @Get('/{featureFlagId}')
+    @OperationId('Get feature flag')
+    async getFeatureFlag(
+        @Request() req: express.Request,
+        @Path() featureFlagId: string,
```
ZeRego, line 27: Note that I typed featureFlagId as a string instead of FeatureFlag enum so we can easily extend commercial flags.

packages/frontend/src/hooks/useFeatureFlagEnabled.ts
```diff
@@ -1,9 +1,36 @@
...
 
 /**
  * Thin wrapper around posthog's useFeatureFlagEnabled hook that is aware
  * of our FeatureFlags enum.
  */
 export const useFeatureFlagEnabled = (featureFlag: FeatureFlags) =>
     useFeatureFlagEnabledPosthog(featureFlag) === true;
+
+/**
+ * Use our own endpoint to get the feature flag from multiple sources.
```
almeidabbm, line 18: Should we add a comment mentioning that not all feature flags are implemented in the new endpoint? this way we don't risk people trying to use it for things that aren't there yet

almeidabbm (approved):

PR #12952, opened 2024-12-18
chore: check for isIntervalBase being present

joaoviana:
Closes: [#12938](https://github.com/lightdash/lightdash/issues/12938)

### Description:

For the reported issue, the following happened
1. User peeks a metric
2. Metric info is returned with `undefined` `timeDimension` 
3. If it's `undefined`, then the date range is never set, which will never trigger the metric explorer query

So after investigating a bit further, realised that time/date dimensions **might** not have `isIntervalBase` so we should check if that exists, then check if truthy. otherwise, ignore

### Reviewer actions


joaoviana:
not necessary/nothing wrong with current implementation

Issue #12903, opened 2024-12-13
downloading a specific dashboard also downloads all associated charts


TuringLovesDeathMetal:
### Description

If I run `download -d my-dashboard`, it should also download all related charts

PR #12955, opened 2024-12-18
fix: downloading a specific dashboard also downloads all associated charts

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/12903

### Description:

Screenshot from 2024-12-18 17-09-52

Another example (creating folders) 

image

i’d expect that the dashboard selector would always download charts linked to the dashboard (even if I added a chart selector as well).

image

### Reviewer actions

### DIFF
--- a/packages/cli/src/handlers/download.ts
+++ b/packages/cli/src/handlers/download.ts
@@ -119,6 +119,43 @@ const readCodeFiles = async <T extends ChartAsCode | DashboardAsCode>(
     return items;
 };
 
+export const downloadCharts = async (
+    chartIds: string[], // slug, uuid or url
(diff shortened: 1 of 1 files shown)


packages/cli/src/handlers/download.ts
```diff
@@ -119,6 +119,66 @@ const readCodeFiles = async <T extends ChartAsCode | DashboardAsCode>(
     return items;
 };
 
+export const downloadContent = async (
```
rephus, line 122: unified download chart/dashboard into a single method that we can reuse

magnew (approved):
🔥 Tested, looks good!

Issue #12969, opened 2024-12-19
Cache Metrics Explorer query


joaoviana:
### Description

Cache Metrics Explorer queries 

Acceptance Criteria

- [x] 24h cache
- [x] Figure out if stored in Postgres or extend/manipulate the current results cache enabled flag - understand which orgs have it enabled. 
- [x] Generate a key that includes the capabilities that the metrics explorer allows - date filter, metric info, time override, and comparison type 
- [x] Decide if we want a quick solution, for example, improving on-metric-explore load time or **all** interactions

### open questions
1. how often are people exploring a metric - if org A explores a metric on a Monday, will they view it on Tuesday? If not, how relevant is this caching strategy?
2. how often are users exploring a metric on the same day?
3. should we ignore the caching if the model contains user attributes?

almeidabbm:
We decided to use the current caching mechanism which proved to be beneficial during our tests. This is enabled via env variable.

PR #13004, opened 2024-12-23
chore: allow force cache for metric explorer queries

almeidabbm:
Closes: #12969 

### Description:

Forces current caching mechanism on metrics explorer queries, cache by default expires within 24h

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1613,6 +1613,7 @@ export class ProjectService extends BaseService {
         explore: validExplore,
         granularity,
         chartUuid,
+        forceUseCache,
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1783,6 +1789,7 @@ export class ProjectService extends BaseService {
         metricQuery: MetricQuery;
         queryTags: RunQueryTags;
         invalidateCache?: boolean;
+        forceUseCache?: boolean;
```
ZeRego, line 1792: Cache is a opt-in feature and should only be controlled/enabled via the feature flag.

ZeRego (changes requested):

PR #13006, opened 2024-12-23
feat: format metric explorer and tree values on the frontend instead of backend

almeidabbm:
Closes: #12997 

### Description:

- Formats result values on the frontend instead of backend
- Fix tree total query to use data range query params, previously they were being ignored

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/MetricTree/MetricTreeConnectedNode.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/MetricTree/MetricTreeConnectedNode.tsx
@@ -2,8 +2,8 @@ import {
     applyCustomFormat,
     ComparisonFormatTypes,
     CustomFormatType,
+    formatItemValue,
     friendlyName,
-    getCustomFormat,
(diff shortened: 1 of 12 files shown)


packages/frontend/src/features/metricsCatalog/components/MetricTree/MetricTreeConnectedNode.tsx
```diff
@@ -84,6 +84,14 @@ const MetricTreeConnectedNode: React.FC<
         [totalQuery.data, data.timeFrame],
     ); // TODO: will it always be prev month?
 
+    const formattedValue = useMemo(() => {
+        if (totalQuery.data) {
+            const customFormat = getCustomFormat(totalQuery.data.metric);
+            return applyCustomFormat(totalQuery.data.value, customFormat);
```
ZeRego, line 90: Could we have used formatItemValue ?
`formatItemValue(metric, totalQuery.data.value)`

ZeRego (approved):

Issue #12987, opened 2024-12-20
SQL Runner should limit the initial query results, not just the pivoted results


magnew:
### Description

SQL Runner should limit the initial query results, not just the pivoted results. 

We currently send this for a select query
```
WITH original_query AS (
    SELECT * 
    FROM "postgres"."jaffle"."customers"
), 
group_by_query AS (
    SELECT 
        "created", 
        SUM("customer_id") AS "customer_id_sum" 
    FROM original_query 
    GROUP BY "created"
)
SELECT * 
FROM group_by_query 
ORDER BY "created" ASC 
LIMIT 500;
```

Which could scan a huge number of results in the orignal query. We should limit the original query and produce something like

```
WITH original_query AS (
    SELECT * 
    FROM "postgres"."jaffle"."customers"
    LIMIT 500 // Put this here
), 
group_by_query AS (
    SELECT 
        "created", 
        SUM("customer_id") AS "customer_id_sum" 
    FROM original_query 
    GROUP BY "created"
)
SELECT * 
FROM group_by_query 
ORDER BY "created" ASC 
LIMIT 500; // Do we care if this one is here? 
```

It seems ok and consistent to keep the outter limit in addition to the original query limit. 

### Steps to Reproduce the Bug or Issue

n/a

### Cloud or self-hosting

None

magnew:
There is a draft PR associated with this (currently closed) which is probably a reasonable approach for setting this inner limit, but if we do that we need to make it possible to change the limit.

jpetey75:
Is this still happening @magnew? I know there have been some changes so I'm not sure if this is still relevant.

PR #13011, opened 2024-12-23
fix: apply limit to inner SQL runner query

magnew:
Closes: #12987 
### Description:

Apply limit to SQL runner queries before pivot. 

WIP: needs more testing

### Reviewer actions


magnew:
This approach probably works and we can revive it, but before doing this (or with it) we need to make it possible to change this inner limit.

PR #13012, opened 2024-12-23
fix: fetch segments to prevent long running query when segmenting

almeidabbm:
### Description:

Fetches top `MAX_SEGMENT_DIMENSION_UNIQUE_VALUES` segments by metric before running the query so that a filter can be applied to reduce amount of results.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
+++ b/packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
@@ -261,18 +261,17 @@ export class MetricsExplorerService<
         projectUuid: string,
         exploreName: string,
         segmentDimension: string | null,
-        filters: MetricQuery['filters'],
+        metricQuery: MetricQuery,
     ) {
         if (!segmentDimension) {
             return [];
         }
 
         const getSegmentsMetricQuery: MetricQuery = {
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
```diff
@@ -254,6 +256,36 @@ export class MetricsExplorerService<
...
+    ) {
+        if (!segmentDimension) {
+            return [];
+        }
+
+        const getSegmentsMetricQuery: MetricQuery = {
+            ...metricQuery,
+            exploreName,
+            dimensions: [segmentDimension],
+            sorts: [{ fieldId: segmentDimension, descending: true }],
```
IrakliJani, line 274: I believe the sort here should be by a metric

IrakliJani (changes requested):
just one requested change

packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts
almeidabbm: Yep, that's true, changed in most recent commit! Thank you 🙏

IrakliJani (approved):
nice!

Issue #13014, opened 2024-12-24
Migrate react-router to v6


ZeRego:
### Description

v5 to v6 
https://github.com/remix-run/react-router/discussions/8753
https://blog.logrocket.com/migrating-react-router-v6-guide/
https://remix.run/blog/react-router-v6 (edited)

PR #13020, opened 2024-12-24
chore: react router v6 part4

ZeRego:
Related to: https://github.com/lightdash/lightdash/issues/13014

### Description:

Migrates routes to CompatRoute which are v5 components using the context from v6. And all useLocation and useParams.
Kept the same assumption in most of the code. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/pages/Home.tsx
+++ b/packages/frontend/src/pages/Home.tsx
@@ -27,14 +27,13 @@ const Home: FC = () => {
     const project = useProject(selectedProjectUuid);
     const onboarding = useOnboardingStatus();
     const pinnedItems = usePinnedItems(
-        selectedProjectUuid!,
+        selectedProjectUuid,
         project.data?.pinnedListUuid,
     );
-    console.log('home', params, selectedProjectUuid);
     const {
         data: mostPopularAndRecentlyUpdated,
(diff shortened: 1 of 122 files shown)


packages/frontend/src/pages/Home.tsx
```diff
@@ -27,14 +27,14 @@ const Home: FC = () => {
     const project = useProject(selectedProjectUuid);
     const onboarding = useOnboardingStatus();
     const pinnedItems = usePinnedItems(
-        selectedProjectUuid,
+        selectedProjectUuid!,
         project.data?.pinnedListUuid,
     );
-
+    console.log('home', params, selectedProjectUuid);
```
joaoviana, line 33: extra log

packages/frontend/src/pages/Space.tsx
```diff
@@ -44,7 +45,10 @@ const Space: FC = () => {
     const { projectUuid, spaceUuid } = useParams<{
         projectUuid: string;
         spaceUuid: string;
-    }>();
+    }>() as {
+        projectUuid: string;
+        spaceUuid: string;
+    };
```
joaoviana, line 51: I like this, wonder if we should have a reusable hook like `useProjectParams` that does not return `undefined` so we can avoid the non-null assertion operator

joaoviana (approved):

PR #13021, opened 2024-12-24
chore: frontend stack upgrade with vite+react component exports

IrakliJani:
Closes: 

### Description:

motivation: improve hot reloading performance in frontend
problem: [Consistent components exports](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react#consistent-components-exports)
how does HMR works: https://www.gatsbyjs.com/docs/reference/local-development/fast-refresh/#how-it-works

- [adds `react-refresh` eslint plugin](https://github.com/ArnaudBarre/eslint-plugin-react-refresh) to identify and prevent files with inconsistent exports
- slightly bumps eslint to fix an [issue with `react-refresh` plugin](https://github.com/ArnaudBarre/eslint-plugin-react-refresh/issues/56#issuecomment-2444932899)
- These changes only address errors reported by `react-refresh` and resolve them. They do not modify business logic or affect any package other than the frontend.
- inconsistent files are split between `utils.ts`, `constants.ts` and `types.ts`. hooks are named after the function itself.

### Reviewer actions


packages/frontend/src/providers/Tracking/TrackingProvider.tsx
```diff
@@ -0,0 +1,200 @@
...
+import {
+    type EventData,
+    type IdentifyData,
+    type PageData,
+    type SectionData,
+    type TrackingData,
+} from './types';
+import useTracking from './useTracking';
+
+const LIGHTDASH_APP_NAME = 'lightdash_webapp';
```
joaoviana, line 23: move to /contants.ts?

packages/frontend/src/providers/Tracking/context.ts
```diff
@@ -0,0 +1,6 @@
+import { createContext } from 'react';
+import { type TrackingContextType } from './types';
+
+const TrackingContext = createContext<TrackingContextType>(undefined as any);
```
joaoviana, line 4:
```suggestion
const TrackingContext = createContext<TrackingContextType | undefined>(undefined);
```

this?

joaoviana (approved):
was able to trigger the error that has been added to ESLint
image

almeidabbm (approved):

packages/frontend/src/providers/Tracking/context.ts
IrakliJani: I am going to avoid making the change, as it triggers an error in two other files. In this PR, I am trying to avoid any changes that are out of scope.

joaoviana: cool!

packages/frontend/src/providers/Tracking/TrackingProvider.tsx
joaoviana: done

PR #13024, opened 2024-12-26
fix: null error in SQL runner save dialog

magnew:
### Description:

Fix an error from setting at text area value to null

### Reviewer actions


packages/frontend/src/features/sqlRunner/components/SaveSqlChartModal.tsx
magnew, line 1: I changed this to pass '' when description is null. I saw that's how we handle it elsewhere. But in this same file we are handle deal with 'name' differently -- it gets defaulted to '' instead of null.

IrakliJani (approved):

PR #13028, opened 2024-12-26
fix: optimistically set time dimension interval picker

joaoviana:
Closes: N/A

### Description:

Optimistically-update set interval in dropdown and let the query take care of the rest
Ensure we keep a stable previous ref to the chosen date range/label so that the segmented control doesnt jump to custom date ranges when changing the interval. 

https://github.com/user-attachments/assets/c9cc6ee5-c3fa-4fb7-ba76-a96f649bd95a

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/visualization/MetricPeekDatePicker.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/visualization/MetricPeekDatePicker.tsx
@@ -141,7 +141,7 @@ export const MetricPeekDatePicker: FC<Props> = ({
             <Popover.Target>
                 <Group position="apart" w="fill-available" noWrap>
                     <SegmentedControl
-                        disabled={isFetching}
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/metricsCatalog/components/visualization/MetricPeekDatePicker.tsx
```diff
@@ -77,11 +83,26 @@ export const MetricPeekDatePicker: FC<Props> = ({
         });
     }, [organizationUuid, projectUuid, track]);
 
-    const matchingPresetLabel = getMatchingPresetLabel(dateRange, timeInterval);
+    const lastStableMatchingPresetLabel = useRef(
```
joaoviana, line 86: could use useState, but useRef is actually a simpler solution that doesn't trigger re-renders

almeidabbm (approved):

Issue #11518, opened 2024-09-10
If I've closed the sidebar before changing to the chart tab, I have no idea how to configure the chart


PriPatel:
(for those using the sql runner for the first time)
1. Open SQL runner
2. Close tables sidebar
3. Write sql and run query
4. Change to chart tab
5. "How do I edit my chart configuration?"

magnew:
Added an improvement here where if the sidebar was close and you switch to viz, the sidebar opens. 

We can potentially further improve the behavior, but I am closing this ticket since the behavior is now different.

PR #13029, opened 2024-12-26
fix: show the sidebar when switching to the viz tab

magnew:
Relates to: #11518 

### Description:

This does two things:
- Moves sidebar state to the slice, making it possible to control it in other parts of the page
- Opens the sidebar (if it was closed) when switching to the viz tab. This mitigates the issue mentioned in the linked ticket (#11518) where it's confusing to switch to the viz tab and see no cahrt config. We could potentially improve this behavior in the future, but this seems better than what we have. And moving sidebar state to the slice will be necessary for whatever else we do there. 

### Reviewer actions


joaoviana:
Tested in Render, works great!

Issue #13013, opened 2024-12-24
[SQL runner] Can't save when there is only one column


ZeRego:
### Description

Can't save when there is only one column. 
I need to select the chart config and set it to table to be able to save.

### Steps to Reproduce the Bug or Issue

https://github.com/user-attachments/assets/74f82119-45ea-4e76-9ca6-dab9d4eacac1

### Cloud or self-hosting

cloud

PR #13035, opened 2024-12-27
feat: enable SQL runner to save to a table when no viz is selected

magnew:
Closes: #13013 

### Description:

This fixes a bug and makes a behavior change:
- **bug:** sql runner cant save when there is only one column
- **change:** when saving from the SQL tab, the viz will be saved as a table, the current view

The implementation does the following
- Populates both table and chart config when data loads (it used to be just the chart config)
- If the user saves while viewing the results table, the table viz will be saved

To test:
- Run a query with one column
- Save from the SQL tab
- A table is saved

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -580,7 +580,7 @@ export const ContentPanel: FC = () => {
                                                                             }
                                                                             style={{
                                                                                 height: inputSectionHeight,
-                                                                                flex: 1,
+                                                                                flex: inputSectionWidth,
                                                                             }}
                                                                             onChartReady={(
                                                                                 instance,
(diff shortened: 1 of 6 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -580,7 +580,7 @@ export const ContentPanel: FC = () => {
                                                                             }
                                                                             style={{
                                                                                 height: inputSectionHeight,
-                                                                                flex: 1,
+                                                                                flex: inputSectionWidth,
```
magnew, line 583: Sometimes the chart wasn't always rendering right. We have a lot of transition animations on this page and it seemed to render halfway through one. This is pretty explicit.

joaoviana (approved):
works great!

tested 
- quick save, saw a table being saved
- edited

tested also: 
- saving a bar chart 
- editing back to table

tested also: 
- resizes of height of the chart

Issue #11894, opened 2024-10-11
The orange state on the word SQL is misleading


PriPatel:
<img width="210" alt="Screenshot 2024-10-11 at 10 29 15" src="https://github.com/user-attachments/assets/621f3e61-c014-4c41-bb01-0cc79550fbf0">

The orange text here feels like an error and is a bit misleading. When I hover over it, it tells me I need to run a query, whcih means I need to move the mouse all the way over to the Run Query - its better nearer the Run Query button.
Furthermore this clashes with the red warning dot icon

- [ ] For now, remove the orange state. Theres a pattern to be implemented near the Run query button later.

PR #13036, opened 2024-12-27
fix: remove orange state from SQL runner

magnew:
Closes: #11894 

### Description:

I saw this ticket when looking for something else in the SQL runner and it was easy enough to pick up. 

Removes the orange warning color from unrun SQL as the linked ticket calls for. 

Before
Screenshot 2024-12-27 at 16 55 58

After
Screenshot 2024-12-27 at 16 56 31

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -349,6 +349,7 @@ export const ContentPanel: FC = () => {
                                                 >
                                                     <Group spacing={4} noWrap>
                                                         <MantineIcon
+                                                            color="gray.6"
                                                             icon={
                                                                 IconCodeCircle
                                                             }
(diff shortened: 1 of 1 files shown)


joaoviana (changes requested):
can you readd the gray.6 colors?

magnew:
> can you readd the gray.6 colors?

Sure. It looked a little disabled to me 😄. But you're right I shouldn't change more style than requested when fixing a style thing.

magnew:
@joaoviana I put it back on the icon and not the text, so it matches the chart button -- what do you think?
Screenshot 2024-12-27 at 17 09 22

Or should it be different than the chart button on purpose? When the text is also gray, it looks odd to me next to the chart button
Screenshot 2024-12-27 at 17 09 56

joaoviana (approved):

joaoviana:
> @joaoviana I put it back on the icon and not the text, so it matches the chart button -- what do you think? Screenshot 2024-12-27 at 17 09 22
> 
> Or should it be different than the chart button on purpose? When the text is also gray, it looks odd to me next to the chart button Screenshot 2024-12-27 at 17 09 56

what you pushed makes more sense 👍 great!

PR #13055, opened 2024-12-30
refactor: replace useHistory with useNavigate (part 6)

joaoviana:
Relates to: #13014 

### Description:

Replace `useHistory` with `useNavigate`

### Reviewer actions


joaoviana:
has everything except for the 2 files that use history.block. 
We need to update the router to be a data router so that we can use `useBlocker` and that will be the replacement of `history.block`

Issue #13057, opened 2024-12-30
When a metric has markdown description, the markdown characters are shown in table view, making it hard to read


PriPatel:
- [ ] This should just read like plain text (like all the other descriptions)
<img width="554" alt="Screenshot 2024-12-30 at 17 59 14" src="https://github.com/user-attachments/assets/3e025b05-ecda-46c3-b4d0-504a8acd5cab" />

- [ ] Markdown should only be rendered when the user **clicks** to view full descriptions

ptdatta:
@PriPatel is it open to work?

PR #13066, opened 2025-01-02
fix: render markdown on clamped description

almeidabbm:
Closes: #13057 

### Description:

- Adds ability to preview clamped markdown

**Note:** Removed ability to display highlight for description when searching 

**Before**
image

**After**
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/MetricCatalogCellOverlay.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/MetricCatalogCellOverlay.tsx
@@ -6,7 +6,9 @@ import {
     px,
     useMantineTheme,
 } from '@mantine/core';
-import MarkdownPreview from '@uiw/react-markdown-preview';
+import MarkdownPreview, {
+    type MarkdownPreviewProps,
+} from '@uiw/react-markdown-preview';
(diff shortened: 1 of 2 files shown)


magnew (approved):
Looks good! 

Probably not a blocker, but I noticed something a little strange is happening with `completion_percentage`. Since there is no label or description (and maybe since it's type: number?), we used to render a `-`, now it gets rendered as a bullet. Which is a little odd. But I can't repro this with most fields, I'm guess cause most metrics are not 'number'?

In main:
Screenshot 2025-01-02 at 14 13 55

This branch:
Screenshot 2025-01-02 at 14 14 01

almeidabbm:
> Looks good!
> 
> Probably not a blocker, but I noticed something a little strange is happening with `completion_percentage`. Since there is no label or description (and maybe since it's type: number?), we used to render a `-`, now it gets rendered as a bullet. Which is a little odd. But I can't repro this with most fields, I'm guess cause most metrics are not 'number'?
> 
> In main: Screenshot 2025-01-02 at 14 13 55
> 
> This branch: Screenshot 2025-01-02 at 14 14 01

Nice catch! should be fixed in the latest commit

Issue #12959, opened 2024-12-18
Incomplete periods on charts look misleading 


PriPatel:
When Im looking at the values for an incomplete period, it gives the wrong impression because the values are much lower.

- The line should look different when its covering an incomplete period. Maybe a decreased transparency.

PR #13077, opened 2025-01-02
feat: show different color for incomplete period

almeidabbm:
Closes: #12959 

### Description:

- Adds dashed line for incomplete period
- Calculate uniq legend payload since now we duplicate the line name for the incomplete period to keep legend interactions

https://github.com/user-attachments/assets/62f590a4-afa5-4cd8-b6bb-dc47e934e41f

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
@@ -4,6 +4,7 @@ import {
     friendlyName,
     getCustomFormat,
     MetricExplorerComparison,
+    type MetricExploreDataPointWithDateValue,
     type MetricExplorerDateRange,
     type MetricExplorerQuery,
     type MetricsExplorerQueryResults,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
```diff
@@ -444,6 +449,42 @@ const MetricsVisualization: FC<Props> = ({
...
+            // Add the last completed period data to the incomplete period data, this will fill in the gap between the last completed period and the first incomplete period
+            incompletePeriodData.push(lastCompletedPeriodData);
+
+            return {
+                ...rest,
+                completedPeriodData,
+                incompletePeriodData,
+            };
+        });
+    }, [results?.metric.timeDimension?.interval, segmentedData]);
```
joaoviana: would be interesting to move this to the react-query hook as a `select` - but an improvement for later

packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
```diff
@@ -122,7 +126,8 @@ const MetricsVisualization: FC<Props> = ({
         (state) => state.metricsCatalog.abilities.canManageExplore,
     );
 
-    const { colors } = useMantineTheme();
+    const theme = useMantineTheme();
+    const colors = theme.colors;
```
joaoviana, line 130: do you need this change? 🤔

joaoviana:
hmm I think this needs some rethinking 

when I have a daily granularity, then today or yesterday shouldn't be considered as incomplete, bc there's no incomplete day in this case

image

for granularity Week, I should only see the last incomplete week, I think I see 2...
image


for granularity month, i see the same as the week example

image


Year makes sense! 

I wonder if dashed is the right style for this... did you have something else in mind?

IrakliJani (approved):
peer-reviewed with Bruno. looks great!

Issue #13081, opened 2025-01-03
Legend doesn't match the tooltip when comparing to previous period


almeidabbm:
### Description

When comparing to previous year the legend doesn't match the values displayed on the tooltip which can be misleading.

This happens because we show the last 12 months.

![image](https://github.com/user-attachments/assets/92bd3b4b-fb59-4c0e-b41b-294a59d8efe3)

### Steps to Reproduce the Bug or Issue

1. Open the metrics explorer
2. Select any time granularity except for year
3. Compare to previous period

### Cloud or self-hosting

None

PR #13085, opened 2025-01-03
fix: adds month to prev period comparison legend when years in date range are different

almeidabbm:
Closes: #13081 

### Description:

- Adds month and year in date range when years are different

**Before**
image

**After**
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
+++ b/packages/frontend/src/features/metricsCatalog/components/visualization/MetricsVisualization.tsx
@@ -263,21 +263,57 @@ const MetricsVisualization: FC<Props> = ({
                 if (!dateRange || !results?.metric.timeDimension?.interval)
                     return null;
 
-                const currentPeriodYear = dateRange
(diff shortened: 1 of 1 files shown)


hamzahc1:
@almeidabbm - only going off the "after" image, the tooltip and legend still show different values 🤔 - although the legend is now much clearer as to what's happening. Is this intentional or do we want to also update the tool tip to reflect the legend values?

almeidabbm:
> @almeidabbm - only going off the "after" image, the tooltip and legend still show different values 🤔 - although the legend is now much clearer as to what's happening. Is this intentional or do we want to also update the tool tip to reflect the legend values?

Good point, I believe this to be intentional as the value in the tooltip changes as you navigate to more recent data points.

https://github.com/user-attachments/assets/bf38fcd4-c873-4950-a81e-2825cfd2370b

In this video, when we hover the last data point, the years are 2024 and 2025, while the ones before that we get 2023 and 2024

joaoviana (approved):

Issue #13095, opened 2025-01-06
When I save a dashboard's changes, I don't get redirected to view mode


joaoviana:
### Description

Edit a dashboard
Make changes
See redirect not happening correctly, user does not get redirected to path /view

### Steps to Reproduce the Bug or Issue

see description for steps

### Cloud or self-hosting

cloud

PR #13100, opened 2025-01-06
fix: redirect to view mode when dashboard is saved

joaoviana:
Closes: #13095 

### Description:

To replicate see issue ^ 

Underlying issue: 
- When we save a dashboard, we navigate back to /view
- The use Effect in dashboard provider for temp filters was running unnecessarily bc the reference was not stable (dashboardTempFilters). I introduced the useDeepCompareEffect to avoid unnecessary re-renders. Before, it would not wait for the navigate in Dashboard.tsx to update the pathname, and a redirect to back to previous pathname (edit mode) would happen. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/Routes.tsx
+++ b/packages/frontend/src/Routes.tsx
@@ -206,7 +206,7 @@ const DASHBOARD_ROUTES: RouteObject[] = [
                 element: <DashboardPageWrapper keyParam={'dashboardUuid'} />,
             },
             {
-                path: '/projects/:projectUuid/dashboards/:dashboardUuid:mode/tabs/:tabUuid?',
+                path: '/projects/:projectUuid/dashboards/:dashboardUuid/:mode/tabs/:tabUuid?',
                 element: <DashboardPageWrapper keyParam={'tabUuid'} />,
             },
         ],
--- a/packages/frontend/src/pages/Dashboard.tsx
+++ b/packages/frontend/src/pages/Dashboard.tsx
@@ -273,7 +273,7 @@ const Dashboard: FC = () => {
                 );
             } else {
                 void navigate(
-                    `/projects/${projectUuid}/dashboards/${dashboardUuid}/view/`,
+                    `/projects/${projectUuid}/dashboards/${dashboardUuid}/view`,
                     { replace: true },
                 );
             }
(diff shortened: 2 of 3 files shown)


packages/frontend/src/pages/Dashboard.tsx
```diff
@@ -273,7 +273,7 @@ const Dashboard: FC = () => {
                 );
             } else {
                 void navigate(
-                    `/projects/${projectUuid}/dashboards/${dashboardUuid}/view/`,
+                    `/projects/${projectUuid}/dashboards/${dashboardUuid}/view`,
```
joaoviana, line 276: I think this makes sense and dont see why it should end with a trailing /view

packages/frontend/src/Routes.tsx
```diff
@@ -206,7 +206,7 @@ const DASHBOARD_ROUTES: RouteObject[] = [
                 element: <DashboardPageWrapper keyParam={'dashboardUuid'} />,
             },
             {
-                path: '/projects/:projectUuid/dashboards/:dashboardUuid:mode/tabs/:tabUuid?',
+                path: '/projects/:projectUuid/dashboards/:dashboardUuid/:mode/tabs/:tabUuid?',
```
joaoviana, line 209: missing /

packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
```diff
@@ -210,8 +210,8 @@ const DashboardProvider: React.FC<
         }
     }, [dashboard, dashboardFilters, overridesForSavedDashboardFilters]);
 
-    // Updates url with temp and overridden filters
-    useEffect(() => {
+    // Updates url with temp and overridden filters and deep compare to avoid unnecessary re-renders for dashboardTemporaryFilters
```
joaoviana, line 213: dashboardTemporaryFilters were unstable even when there were no changes to them. 
After saving the dashboard, this useEffect would run with an outdated `pathname,` (edit pathname) in 245, which would do the following: user navigated back to view -> useeffect runs -> navigated back to edit

packages/frontend/src/Routes.tsx
magnew: doh

magnew (approved):
Looks good! I tested this in a lot of ways and it works. I guess its hard to know for sure with a useEffect race condition, but this seems to fix the issue. 

Nice work -- this is a good fix for a tricky problem.

Issue #11967, opened 2024-10-17
CTAs in the top right corner of a SQL chart in edit Mode have incorrect states. 


PriPatel:
Im pretty these were fixed here https://github.com/lightdash/lightdash/pull/11471 , but Im seeing the same issues again today: 
<img width="215" alt="Screenshot 2024-10-17 at 10 37 22" src="https://github.com/user-attachments/assets/9aaf12d5-724b-439d-916b-a9834c14b2ba">

These buttons should match edit mode of an explore :
- [ ] The Save button should only be green if changes have been made
- [ ] The icon button should be a text button saying 'Cancel'

PR #13109, opened 2025-01-06
fix: sql runner edit-mode buttons get the right state

magnew:
Closes: #11967 
### Description:

Makes the SQL runner edit mode buttons consistently match spec as per #11967, with one exception: when there are no edits, the icon button remains. It felt a bit odd to have a 'Cancel' button with no alternative. Now it looks like

**Edit mode, no changes**
Screenshot 2025-01-06 at 18 09 25

**Edit mode with changes**
Screenshot 2025-01-06 at 18 09 34

The missing green state was only a problem on tables. Now fixed. 

### Reviewer actions

### DIFF
--- a/packages/common/src/visualizations/TableDataModel.ts
+++ b/packages/common/src/visualizations/TableDataModel.ts
@@ -75,7 +75,6 @@ export class TableDataModel {
                     reference: key,
                     label: this.columnsConfig?.[key]?.label ?? key,
                     frozen: this.columnsConfig?.[key]?.frozen ?? false,
-                    order: this.columnsConfig?.[key]?.order,
                 },
             }),
             {},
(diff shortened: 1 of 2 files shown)


packages/common/src/visualizations/TableDataModel.ts
```diff
@@ -75,7 +75,6 @@ export class TableDataModel {
                     reference: key,
                     label: this.columnsConfig?.[key]?.label ?? key,
                     frozen: this.columnsConfig?.[key]?.frozen ?? false,
-                    order: this.columnsConfig?.[key]?.order,
```
joaoviana, line 78: why this change sorry?

magnew: My bad, I meant to mention this in the PR. This is what was causing the save button to be green no matter what for tables. We set order to undefined here, but it is never used so we don't save it. The saved table then has no 'order' field and the logic that checks whether there are changes (to make the save button green) sees that as a change.

joaoviana (approved):

Issue #10384, opened 2024-06-12
Remove legacy explore model methods


ZeRego:
### Description

Remove `getExploreFromCache`, `findExploreByTableName`, `getExploresFromCache` and any query to the table `CachedExploresTableName`.
Instead, use `findExploresFromCache`.

(Create a new ticket) In a following PR add a migration to delete `CachedExploresTableName` table and the remaining code (entity, knex).

yshplsngh:
Hello @ZeRego, Can you please assign me to this issue?

yshplsngh:
@ZeRego After merging the PR
I will start working on removing `CachedExploresTableName` table

PR #13113, opened 2025-01-06
chore: remove legacy getExploreFromCache method

yshplsngh:
#10384

### Description:
This PR provides a partial fix for issue #10384. As per @ZeRego guidance, to keep PRs focused when they impact >3 files, additional PRs will be submitted for the remaining methods listed in this issue.

Changes in this PR include:
- removed the `getExploreFromCache` method and instead used `findExploresFromCache`

### Reviewer actions


packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -1227,10 +1227,11 @@ export class CatalogService<
             throw new ForbiddenError();
         }
 
-        const explore = await this.projectModel.getExploreFromCache(
+        const explores = await this.projectModel.findExploresFromCache(
             projectUuid,
-            tableName,
+            [tableName],
         );
+        const explore = explores[tableName];
```
ZeRego, line 1234: To avoid applying the same logic in all services. Let's keep the methods `getExploreFromCache` and `findExploreByTableName` but refactor them to use `findExploresFromCache`. 
That way, no changes are required to services and we deduplicate queries for easier maintenance. 

Eg:
```
    async getExploreFromCache(
        projectUuid: string,
        exploreName: string,
    ): Promise<Explore | ExploreError> {
        const cachedExplores = await this.findExploresFromCache(projectUuid, [exploreName]);
        const cachedExplore = cachedExplores[exploreName];
        if (cachedExplore === undefined) {
            throw new NotExistsError(
                `Explore "${exploreName}" does not exist.`,
            );
        }
        return cachedExplore;
    }
    
    async findExploreByTableName(
        projectUuid: string,
        tableName: string,
    ): Promise<Explore | ExploreError | undefined> {
        const cachedExplores = await this.findExploresFromCache(projectUuid, [tableName]);
        return cachedExplores[tableName];
    }
```

ZeRego (changes requested):

packages/backend/src/services/CatalogService/CatalogService.ts
ZeRego: For `getExploresFromCache` we should replace it with `findExploresFromCache` as it had an incorrect return type.

yshplsngh: @ZeRego 
Sorry for the inconvenience. The code of this branch became a mess due to my mistake
So I created a new PR with the latest changes #13211.
From next this will not happen again.

Issue #13118, opened 2025-01-07
Pylon widget not loading 


joaoviana:
### Description

Figure out why it's not loading - probably due to the React 19 upgrade and how the script is injected(?)

### Steps to Reproduce the Bug or Issue

no steps

### Cloud or self-hosting

cloud

ptdatta:
hey @joaoviana Can I work on it?

ptdatta:
> hi @ptdatta I've assigned myself to it and am actively working on it, thank you tho! :)

Oo ok thank you.

PR #13120, opened 2025-01-07
fix: load pylon script through hook

joaoviana:
Closes: #13118 

### Description:

Followed this example: https://github.com/rilldata/rill/pull/4861/files

Added hook `usePylon` to load script with ts rather than using `script` element - couldn't make it work

How to test: 
- set app id env var
- go to update HelpMenu.tsx to have {isCloudCustomer || true && (
 
Screenshot 2025-01-07 at 13 50 08

### Reviewer actions

### DIFF
--- a/packages/frontend/src/providers/ThirdPartyServicesProvider.tsx
+++ b/packages/frontend/src/providers/ThirdPartyServicesProvider.tsx
@@ -1,26 +1,70 @@
 import { LightdashMode } from '@lightdash/common';
 import { PostHogProvider, usePostHog } from 'posthog-js/react';
-import { useEffect, type FC } from 'react';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/providers/ThirdPartyServicesProvider.tsx
```diff
@@ -1,26 +1,70 @@
...
+                    initPylonWidget(health.data.pylon.appId);
+                }
+            }
+        },
+        [user, health, initPylonWidget],
+    );
+
+    useEffect(
+        function setNewIssueCustomFields() {
+            if (health.data?.pylon?.appId && user.data) {
```
ZeRego, line 67: We should check if `window.Pylon` is defined. And with that fix we can have a single useEffect.

Issue #13051, opened 2024-12-30
The 'expanded' version of the tile should show when it is outside of the grey box


PriPatel:
Right now, the expanded version of the tile only shows when the tile is connected. This means I don't really know whats going on and don't see the value of this until I actually make a connection. (but i probably won't do this because I don't know that I'll eventually see a big number!) 

- [ ] It makes more sense to show the expanded version of the tile with the big number when it is outside of the grey box. (even if its unconnected) 
- [ ] Moving it back to the grey box collapses it. 

This requires a bit of a logic change and playing around with the 'grey box' and its boundaries. It might need to become responsive.

PR #13130, opened 2025-01-07
feat: show expanded node when dragging out

almeidabbm:
Closes: #13051 

### Description:

**Before**

https://github.com/user-attachments/assets/301e2bf6-e7c2-4837-82e1-5632eb06424a

**After**

https://github.com/user-attachments/assets/8f1c69eb-66d7-4999-becb-4319df7b2676

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/metricsCatalog/hooks/useTreeNodePosition.tsx
+++ b/packages/frontend/src/features/metricsCatalog/hooks/useTreeNodePosition.tsx
@@ -10,26 +10,19 @@ export const useTreeNodePosition = (nodeId: string) => {
             const intersectingNode = getNode(intersectingNodeId);
             if (!intersectingNode || !node) return false;
 
-            const {
-                position: intersectingNodePosition,
-                measured: intersectingNodeMeasured,
-            } = intersectingNode;
-
-            const {
-                width: intersectingNodeWidth,
-                height: intersectingNodeHeight,
-            } = intersectingNodeMeasured ?? {};
-
+            const { position: intersectingNodePosition } = intersectingNode;
             const { position: nodePosition, measured: nodeMeasured } = node;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/features/metricsCatalog/hooks/useTreeNodePosition.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+            return !(
+                (
+                    intersectingNodePosition.x + 100 <= nodePosition.x || // intersecting node is to the left, here we don't use the node width because node expands to the rigth, so we use a fixed value
+                    intersectingNodePosition.x >=
+                        nodePosition.x + (nodeWidth ?? 0) || // intersecting node is to the right
+                    intersectingNodePosition.y + 20 <= nodePosition.y || // intersecting node is above, here we don't use height because node expands to the bottom, so we use a fixed value
+                    intersectingNodePosition.y >=
+                        nodePosition.y + (nodeHeight ?? 0)
+                ) // intersecting node is below
```
joaoviana, line 25: can you clarify the usage of `100` and `20` here?

almeidabbm: Yep, definitely, so when a node expands (from just the name from the name + metric value) it grows vertically (expanding down) and horizontally (expanding right), because of this we cannot rely on the position of `x + (width | heigh)` as the base point for checking if there's a overlap since those values will change. By using static values this is more consistent at the expense of not being an instant (as soon as the node touches the free nodes box) change from expanded to not expanded.

If we use `width` or `height` calculation when node is touching the box when coming in from the top (touching with the bottom edge) or from the left (touching with the right edge) the node will flicker between both expanded and not expanded state as it will reduce in size and briefly the edges will not touch the box

Issue #13119, opened 2025-01-07
Upload images to slack from deliveries


rephus:
### Description

Now we can upload images to Slack on scheduled deliveries

This was done for `image` chart deliveries, but we're still missing some other use cases like.

- threshold alerting 
- dashboards

Once they are all done , we can remove the footer about expiration (not for csvs) 
Follow the implementation for image charts and upload images to slack for those other different scenarios.

PR #13138, opened 2025-01-08
fix: threshold alert upload image to slack

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13119

### Description:

Dashboards: 

Screenshot from 2025-01-08 10-20-26

Threshold alert 

Screenshot from 2025-01-08 10-17-44

### Reviewer actions


packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -557,10 +563,8 @@ export default class SchedulerTask {
 
                     const blocks = getChartThresholdAlertBlocks({
                         ...getBlocksArgs,
-                        footerMarkdown: `This is a ${thresholdFooter} sent by Lightdash. For security reasons, delivered files expire after ${
-                            this.s3Client.getExpirationWarning()?.days || 3
-                        } days`,
-                        imageUrl,
+                        footerMarkdown: `This is a ${thresholdFooter} sent by Lightdash.`,
+                        imageUrl: slackImageUrl,
```
IrakliJani, line 567: should we keep an expiration text if the image is hosted on s3?

this can be a return type for `tryUploadingImageToSlack`

`{ url: string, expiring: boolean }`

IrakliJani (approved):

IrakliJani:
@rephus im curious, when does the upload fail?

rephus:
> @rephus im curious, when does the upload fail?

I didn't get any image errors, but considering slack takes some time to process the image before it can be used (and we have a limited number of retries), I believe this could fail at some point. So it everything goes well, the image should always be uploaded, but becuase I don't trust slack, I keep the s3 url as fallback. What do you think ? 

> also, why do we upload to s3 and then upload to slack? can't we directly upload to slack?

We first upload to s3, because the `handle` method is `target` agnostic, the same file is uploaded once, and then it can be used on email or slack . WE could upload directly to slack,but we'll have to change the parent method, I believe it is easier this way. 

let me know if you hayve any other implementation suggestions.

packages/backend/src/scheduler/SchedulerTask.ts
rephus: done

PR #13154, opened 2025-01-08
Chore/improve catalog search query

almeidabbm:
### Description:

- Improves the performance of the search catalog query by doing the tags catalog tags aggregation on a second query after the catalog items query has finished. This helps because when aggregating we would have to group by which became very expensive.

Tests were done with 200000+ metrics:

**Before**
image

**After**
image

### Reviewer actions


rephus (approved):
:rocket:

Issue #13166, opened 2025-01-09
support argument --project on lightdash download/upload


rephus:
### Description

So instead of switching projects using `config set-project`, you can choose a different project using the uuid

zhyd1997:
Hi @rephus 

I am open to contribute!

PR #13185, opened 2025-01-10
feat(cli): support argument `--project` on lightdash download/upload.

zhyd1997:
Closes: #13166 

### Reviewer actions

### DIFF
--- a/packages/cli/src/handlers/download.ts
+++ b/packages/cli/src/handlers/download.ts
@@ -26,6 +26,7 @@ export type DownloadHandlerOptions = {
     dashboards: string[]; // These can be slugs, uuids or urls
     force: boolean;
     path?: string; // New optional path parameter
+    project?: string;
 };
 
 const getDownloadFolder = (customPath?: string): string => {
@@ -213,7 +214,7 @@ export const downloadHandler = async (
         );
     }
 
-    const projectId = config.context.project;
+    const projectId = config.context.project || options.project;
     if (!projectId) {
         throw new Error(
             'No project selected. Run lightdash config set-project',
@@ -478,7 +479,7 @@ export const uploadHandler = async (
         );
     }
 
-    const projectId = config.context.project;
+    const projectId = config.context.project || options.project;
     if (!projectId) {
         throw new Error(
             'No project selected. Run lightdash config set-project',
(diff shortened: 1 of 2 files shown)


packages/cli/src/index.ts
```diff
@@ -56,6 +57,20 @@ function parseUseDbtListOption(value: string | undefined): boolean {
     return value.toLowerCase() !== 'false';
 }
 
+function parseProjectArgument(value: string | undefined): string | undefined {
```
rephus, line 60: nice validation

packages/cli/src/handlers/download.ts
```diff
@@ -213,7 +214,7 @@ export const downloadHandler = async (
         );
     }
 
-    const projectId = config.context.project;
+    const projectId = config.context.project || options.project;
```
rephus, line 217: I don't think this is the right way to get the projectId, it should be the other way arround, the  `options` should override the context if provided.

rephus (changes requested):
I tested this and it is not working as expected 



I asked to download a dashboard from `--project 3675b69e-8324-4110-bdca-059031aa8da3` , however it is trying to fetch from project `bd9e6480-7c15-45e5-b279-eb812af00db7` which is the one on my context (config set-project). 

You can see it on the verbose logs 


image

packages/cli/src/index.ts
```diff
@@ -475,6 +496,12 @@ program
         'specify a custom path to upload charts and dashboards from',
         undefined,
     )
+    .option(
+        '--project <project uuid>',
```
rephus, line 500: we could also accept `-p` as a parameter, like the rest of the arguments (They all have a short and a long version)

zhyd1997: `-p` is used for `—path`.

packages/cli/src/handlers/download.ts
zhyd1997: I see.

Fixed: https://github.com/lightdash/lightdash/pull/13185/commits/427f9e97d3a2462af62e23c95ae99d11deb16372

rephus (approved):
Works as expected. Thanks !

rephus:
@all-contributors add @zhyd1997 for code

PR #13188, opened 2025-01-10
fix: display field is used incorrectly

yzqzy:
The display field is used incorrectly. The value should not be used.


ZeRego:
@yzqzy Thank you for the contribution.
Can you provide more context? is there an open ticket about the issue?
Can you provide a screenshot of the problem and the fix?

yzqzy:
In an internationalized scenario, this label is in Chinese, and the list needs to be displayed in Chinese rather than in value.

ZeRego:
@yzqzy In that case, a fix is required where the label is displayed. It should find the explore/table label based on the name.
The code you changed needs to be the values so they can be referenced in other places.

PR #13212, opened 2025-01-11
Feat: Added explore_name to query metadata

### LINKED ISSUE #10258: Add explore name to the Lightdash query metadata
### Description

Each warehouse query has metadata from Lightdash (eg: project uuid, user uuid, etc ) 
We want to have the explore name/id to that metadata. 

Tech note:
Look for the type `RunQueryTags` and check where it is used

> Would be great if in Lightdash query tag you'd also include information about the entry explore model that's being used.

Easier for admins to relate queries to specific explores.

ptdatta:
Closes: #10258 

### Description:

Fixes 

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -3520,7 +3520,6 @@ export class ProjectService extends BaseService {
             organization_uuid: user.organizationUuid,
             project_uuid: projectUuid,
             user_uuid: user.userUuid,
-            explore_name: tableName,
         };
         let database = ProjectService.getWarehouseDatabase(credentials);
         if (!database) {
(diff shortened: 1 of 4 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3513,6 +3520,7 @@ export class ProjectService extends BaseService {
             organization_uuid: user.organizationUuid,
             project_uuid: projectUuid,
             user_uuid: user.userUuid,
+            explore_name: tableName,
```
ZeRego, line 3523: We should remove this one.

ZeRego (changes requested):
Just a small fix needed.

ptdatta:
> Just a small fix needed.

Fixed ✅

ZeRego (approved):

PR #13243, opened 2025-01-13
fix: add pivot info to sql runner

magnew:
Closes: 

### Description:

WIP: return origin information with pivoted data

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -107,6 +107,7 @@ import {
     UpdateVirtualViewPayload,
     UserAttributeValueMap,
     UserWarehouseCredentials,
+    VizAggregationOptions,
     VizColumn,
     WarehouseClient,
     WarehouseCredentials,
@@ -2448,6 +2449,17 @@ export class ProjectService extends BaseService {
         let currentRowIndex = 0;
         let currentTransformedRow: ResultRow | undefined;
         const valuesColumnReferences = new Set<string>(); // NOTE: This is used to pivot the data later with the same group by columns
+        const valuesColumnMetadata = new Set<{
+            referenceField: string;
+            id: string;
+            aggregation: VizAggregationOptions;
+            pivotValues: {
+                field: string;
+                value: string;
+            }[];
+        }>();
+
+        // here
 
         const fileUrl = await this.downloadFileModel.streamFunction(
             this.s3Client,
@@ -2461,6 +2473,8 @@ export class ProjectService extends BaseService {
                             rows.forEach(writer);
                             return;
                         }
+
+                        // columns appears unused
                         if (!columns.length) {
                             // Get column types from first row of results
                             columns.push(
@@ -2470,6 +2484,8 @@ export class ProjectService extends BaseService {
                                 })),
                             );
                         }
+
+                        console.log('rows>>>>>>>>>>>>>', rows);
 
                         rows.forEach((row) => {
                             // Write rows to file in order of row_index. This is so that we can pivot the data later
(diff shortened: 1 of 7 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2513,14 +2537,21 @@ export class ProjectService extends BaseService {
 
         await sshTunnel.disconnect();
 
+        console.log(
```
magnew, line 2540: :p

packages/frontend/src/features/queryRunner/semanticViewerPivotQueries.ts
```diff
@@ -108,9 +102,19 @@ export const getPivotQueryFunctionForSemanticViewer = (
...
+                // but the extra information isn't used in the Semantic viewer
+                acc.push({
+                    referenceField: col,
+                    id: col,
+                    aggregation: VizAggregationOptions.ANY,
+                    pivotValues: [],
+                });
+            }
+            return acc;
+        }, [] as PivotValuesColumn[]);
```
IrakliJani, line 117:
```ts
const valuesColumns: PivotValuesColumn[] = pivotedResults.columns.reduce<PivotValuesColumn[]>((acc, col) => {
    if (!query.pivot?.index.includes(col)) {
        // Create a new object for the current column
        const newColumn: PivotValuesColumn = {
            referenceField: col,
            id: col,
            pivotValues: [],
        };

        // Use the spread operator to create a new array with the new column
        return [...acc, newColumn];
    }

    return acc;
}, []);
```

a bit safer alternative to avoid `as` and `push`

IrakliJani: I was going to suggest making this optional or somehow trimming this information from the result metadata for the semantic viewer, but since this can be helpful in the semantic viewer and does not add a lot of data transfer, I am okay to include this metadata for sem viewer as well.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2481,7 +2483,7 @@ export class ProjectService extends BaseService {
         const columns: VizColumn[] = [];
         let currentRowIndex = 0;
         let currentTransformedRow: ResultRow | undefined;
-        const valuesColumnReferences = new Set<string>(); // NOTE: This is used to pivot the data later with the same group by columns
+        const valuesColumnData = new Map<string, PivotValuesColumn>();
```
IrakliJani, line 2486: ok cool. so we used to have uniq values of pivot column. but now the value is a key so it's still "uniq" 👍

packages/frontend/src/components/DataViz/store/cartesianChartBaseSlice.ts
```diff
@@ -34,7 +35,7 @@ export type CartesianChartState = {
         | undefined;
     chartDataLoading: boolean;
     chartDataError: Error | null | undefined;
-    series?: string[];
+    series?: PivotValuesColumn[];
```
IrakliJani, line 38: okay rich! that's what you'll use for UI I guess

packages/common/src/visualizations/PieChartDataModel.ts
```diff
@@ -386,7 +386,9 @@ export class PieChartDataModel {
                             ? result[transformedData.indexColumn.reference]
                             : '-',
                         groupId: transformedData.indexColumn?.reference,
-                        value: result[transformedData.valuesColumns[0]],
+                        value: result[
+                            transformedData.valuesColumns[0].pivotColumnName
```
IrakliJani, line 390: we should probably leave a note for our future selves explaining why we have an array here.

magnew: Will do.

IrakliJani (approved):
code looks good, types look great. tested locally with old version + new version to check backward compatibility on persisted charts - all good there too!

- old
CleanShot 2025-01-21 at 17 23 21@2x

CleanShot 2025-01-21 at 17 23 37@2x

- new
CleanShot 2025-01-21 at 17 21 53@2x
CleanShot 2025-01-21 at 17 21 00@2x

Issue #11939, opened 2024-10-15
You should be able to customize grouped series in the sql runner


PriPatel:
It should be possible to customize series individually for any property that is available in the main series. There are two main cases

- Right now if I `group by`, I don't see the series show up anywhere in the chart config panel and am unable to customise the colours / labels etc.

joaoviana:
Blocked by: #11611

joaoviana:
Reference: https://github.com/lightdash/lightdash/issues/10994#issuecomment-2273447007

Look at local branch: `discussion/chart-series-types` for type changes when we start grouped series

First, focus on UI changes to single series

PR #13281, opened 2025-01-14
feat: sql runner customize pivoted series

magnew:
Closes: #11939

### Description:

Makes it so pivoted series can be customized. Before this, grouping by a column didn't allow styling or customization of the resulting pivoted series. Now it does. 

The style of the forms will probably be updated, but the functionality is there and we can iterate on it. 
Screenshot 2025-01-20 at 14 25 19

### Reviewer actions


packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
```diff
@@ -46,217 +41,175 @@ export const CartesianChartSeries = ({
...
+                    label: seriesLabel ?? friendlyName(s.id),
+                    color: seriesColor ?? colors[index],
+                    type: seriesType,
+                    valueLabelPosition: seriesValueLabelPosition,
+                    whichYAxis: seriesWhichYAxis,
+                    // TODO: for now tracking the index in the series array.
+                    // This makes it so we don't need to make a breaking change to
+                    // the saved format, which applies styles by index, but is less
+                    // readable and less performant than if we used the reference.
+                    seriesIndex: index,
```
magnew, line 79: I want to remove this before checkin. It was for back compat but it doesn't work right. Making this a draft till then.

magnew: To add a little more detail here: I'm going to remove this since after some more testing it's clear it won't work to use an index. These configs will refer to the series by name instead of index, since there is no way to keep the saved indices synced with the data. The result will be a subtle braking change for existing charts, but only charts that aren't really working right at the moment. If you have pivoted, then styled a pivoted series, the styles will be off. 

There will be no change to the data on the chart, just the display properties.

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -512,12 +512,15 @@ export class CartesianChartDataModel {
...
-                );
-
                 const seriesColumnId = seriesColumn.pivotColumnName;
 
+                // NOTE: seriesColumnId is the post pivoted column name and we now store the display based on that.
+                // If there is no display object for the seriesColumnId, we also referenceField for compatibility with
+                // the old display object that stored display info by the field, not the ID.
+                const seriesDisplay =
+                    display?.series?.[seriesColumnId] ??
+                    display?.series?.[seriesColumn.referenceField];
```
magnew, line 522: There is a small breaking change to existing charts here that Oliver and I talked about and decided was OK. We used to store the field reference and an index and look series up by index. Index was broken since indices can change, especially when pivoting, so now we look them up by the pivoted column id which includes the pivotted names. But we also fallback to looking for the field ID we used to save. The new code will result in slightly different rendering for charts that were pivotted and had styles applied to the pivotted series (which didn't work well anyway). 

Here is a dashboard saved from **main**
Screenshot 2025-01-21 at 14 15 04

**And here is what it looks like with this change**
Screenshot 2025-01-21 at 14 35 03

IrakliJani:
@magnew 

I visualized the same data in old explorer and new sql runner.
only noticeable difference between them is the UI groups.

CleanShot 2025-01-21 at 17 53 47@2x

CleanShot 2025-01-21 at 17 54 10@2x

I think the first one (old explorer) is the proper way.
It should be grouped on a pivot field so I think top level group should be "Total order amount" and then then sub items should be "avg_placed", "avg_returned", "avg_return_pending", etc..

it even looks a bit strange with two levels of the same header:

CleanShot 2025-01-21 at 17 57 44@2x
CleanShot 2025-01-21 at 17 57 37@2x

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
```diff
@@ -46,217 +41,227 @@ export const CartesianChartSeries = ({
...
+                    label: seriesLabel ?? friendlyName(s.pivotColumnName),
+                    color: seriesColor ?? colors[index],
+                    type: seriesType,
+                    valueLabelPosition: seriesValueLabelPosition,
+                    whichYAxis: seriesWhichYAxis,
+                };
+
+                if (!acc[s.pivotColumnName]) {
+                    acc[s.pivotColumnName] = [];
+                }
```
IrakliJani, line 78: this check can be way higher to halt earlier

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
```diff
@@ -46,217 +41,227 @@ export const CartesianChartSeries = ({
...
+
+                acc[s.pivotColumnName].push(config);
+
+                return acc;
```
IrakliJani, line 82: monir, can be immutable:
```suggestion
                return {
                    ...acc,
                    [s.pivotColumnName]: [...(acc[s.pivotColumnName] || []), config],
                };
```

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
```diff
@@ -46,217 +41,227 @@ export const CartesianChartSeries = ({
...
+                                    {
+                                        value: 'Stacked',
+                                        label: 'Stacked',
+                                    },
+                                ]}
+                                defaultValue={
+                                    currentConfig?.display?.stack
+                                        ? 'Stacked'
+                                        : 'None'
+                                }
```
IrakliJani, line 178: (I know it's not part of a change)
this is going to raise a warning -  since this is a controlled react input it should be `value`

packages/frontend/src/components/DataViz/config/SingleSeriesConfiguration.tsx
```diff
@@ -0,0 +1,142 @@
...
+                                        <MantineIcon icon={IconAlignRight} />
+                                    </Group>
+                                ),
+                            },
+                        ]}
+                        value={whichYAxis === 1 ? 'right' : 'left'}
+                        onChange={(value) =>
+                            onAxisChange(
+                                reference,
+                                value === 'left' ? 'left' : 'right',
```
IrakliJani, line 122: I noticed this later is converted to 0 and 1

can be an enum

```
enum AxisDirection {
   LEFT
   RIGHT
}
```

might help with the typing a bit too instead of using literal values

magnew:
> @magnew
> 
> I visualized somewhat similar data in old explorer and new sql runner. only noticeable difference between them is the UI groups.
> 
> CleanShot 2025-01-21 at 17 53 47@2x
> 
> CleanShot 2025-01-21 at 17 54 10@2x
> 
> I think the first one (old explorer) is the proper way. It should be grouped on a pivot field so I think top level group should be "Total order amount" and then then sub items should be "avg_placed", "avg_returned", "avg_return_pending", etc..
> 
> it even looks a bit strange with two levels of the same header:
> 
> CleanShot 2025-01-21 at 17 57 44@2x CleanShot 2025-01-21 at 17 57 37@2x

Good catch, thank you! Fixed.

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
magnew: Done

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
magnew: Done

packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx
magnew: Done

packages/frontend/src/components/DataViz/config/SingleSeriesConfiguration.tsx
magnew: Done. Still a little gross cause mantine wants a string, but adding the Enum makes it clearer elsewhere.

IrakliJani: awesome

magnew:
merged in a linked PR

PR #13320, opened 2025-01-16
chore: rename metric tree to canvas

almeidabbm:
### Description:

- Renames `MetricTree` component to `Canvas`
- Replaces `/tree` route to `/canvas`

This is so we follow new naming.
Canvas will still render metric trees (it only renders that for now) so the backend endpoints to fetch the tree remains the same, but new naming makes it more generic and future proof as the canvas might be able to render other things.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/Routes.tsx
+++ b/packages/frontend/src/Routes.tsx
@@ -371,13 +371,13 @@ const METRICS_ROUTES: RouteObject[] = [
         ),
     },
     {
-        path: '/projects/:projectUuid/metrics/tree',
+        path: '/projects/:projectUuid/metrics/canvas',
         element: (
             <>
                 <NavBar />
                 <TrackPage name={PageName.METRICS_CATALOG}>
                     <MetricsCatalog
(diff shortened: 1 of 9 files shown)


joaoviana (changes requested):
could you look at the queryKeys' renames? 
I think that would be beneficial. 
And are you ok with no renames in lightdash/common? 

because there are some methods that still mention "tree" like: "getDefaultMetricTreeNodeDateRange"

would be nice to change these as well

almeidabbm:
> could you look at the queryKeys' renames? I think that would be beneficial. And are you ok with no renames in lightdash/common?
> 
> because there are some methods that still mention "tree" like: "getDefaultMetricTreeNodeDateRange"
> 
> would be nice to change these as well

I left the `tree` name behind on anything related to the metric trees (backend and common) because I believe we should think that we're rendering a `metrics tree` on a `canvas`. Which is true, the canvas is the space where things are rendered and trees are the actual relationship between metrics

joaoviana (approved):

Issue #13323, opened 2025-01-16
Amend error message when joined model has required attributes


ZeRego:
### Description

We display different message depending if the model with the required attribute is the main table or a joined table. Error message from the latter should be similar to the first one. 

Tech note:
the error logic should also be moved inside the _compileQuery method

### Steps to Reproduce the Bug or Issue

<img width="1116" alt="Image" src="https://github.com/user-attachments/assets/9a2b5399-1a4f-49de-815f-7983cbfc9232" />

### Cloud or self-hosting

None

PR #13353, opened 2025-01-21
fix: identify filtered explore error on missing dimension

rephus:
Closes:  #13323

### Description:

Before:

Screenshot from 2025-01-21 09-16-14

After

Screenshot from 2025-01-21 10-30-50

unfilterdTables is not exposed on the explore api

Screenshot from 2025-01-21 10-36-00

### Reviewer actions


packages/backend/src/queryBuilder.ts
```diff
@@ -79,6 +80,22 @@ const getDimensionFromId = (
                 };
         }
 
+        // At this point, we couldn't find the dimension with the given id in the explore
+        // it is possible that the explore is a joined table and is filtered by user_attributes
+        // So we check if the dimension exists in the unfiltered tables
+        if (
```
rephus, line 86: _compileQuery already has the explore filtered out, that happens on the `getExplores` generic method at the beggining of `get chart results` .

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -75,6 +75,9 @@ export const getFilteredExplore = (
         joinedTables: explore.joinedTables.filter((joinedTable) =>
             filteredTableNames.includes(joinedTable.table),
         ),
+        ...(filteredTableNames.length > 0
```
rephus, line 78: no need to add this argument if there are no filtered tables, since this will be redundant

packages/common/src/index.ts
```diff
@@ -494,7 +494,7 @@ export type ApiCompiledQueryResults = string;
 
 export type ApiExploresResults = SummaryExplore[];
 
-export type ApiExploreResults = Explore;
+export type ApiExploreResults = Omit<Explore, 'unfilteredTables'>;
```
ZeRego, line 497: Have you run `generate-api` to check there aren't other affected response types?

ZeRego (approved):

packages/common/src/index.ts
rephus: on swagger, all affected endpoints replaced `Explore` with `Pick_Explore.Exclude_keyofExplore.unfilteredTables__`

PR #13356, opened 2025-01-21
fix: do not filter explores early on for chart results

rephus:
Closes:  #13323

Alternative solution 

### Description:

Before:

Screenshot from 2025-01-21 09-16-14

After

Screenshot from 2025-01-21 10-30-50

unfilterdTables is not exposed on the explore api

Screenshot from 2025-01-21 10-36-00

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1164,9 +1164,6 @@ export class ProjectService extends BaseService {
         );
 
         try {
-            if (isExploreError(filteredExplore)) {
-                throw new UnexpectedServerError('Filtered explore is an error');
-            }
             const exploreWithOverride =
                 ProjectService.updateExploreWithGranularity(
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1154,29 +1158,55 @@ export class ProjectService extends BaseService {
         timezone: string,
         granularity?: DateGranularity,
     ): Promise<CompiledQuery> {
-        const exploreWithOverride = ProjectService.updateExploreWithGranularity(
+        const filteredExplore = ProjectService.filterExplore(
```
rephus, line 1161: do the filtering inside the _compilequery method, if there is a FieldReferenceError here, then we check if the file exists in the unfiltered "explore"

ZeRego:
It was not clear for devx when to pass `false`. closing this in favour of the other PR.

PR #13216, opened 2025-01-13
fix: --noUncheckedIndexedAccess on PromoteService

ptdatta:
Closes: 

### Description:

Added strict undefined check for `PromoteService`

Goal: Enable noUncheckedIndexedAccess  all over backend.

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #13364, opened 2025-01-22
chore: Throw error any tsconfig

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/3790

### Description:

Added: 

-  "useUnknownInCatchVariables": true,
-  "noImplicitAny": true
- "@typescript-eslint/no-explicit-any": "warn"

I added `no-explicit-any` as a warn, since there are too many to handle. Just in common there are 86 errors, 154 in backend. 

image

I do not recommend enabling `fixToUnknown` since it will automatically convert all anys into unkonwns, causing a lot of different errors: 

image

To the main tsconfig, we rely on this file on all packages, I don't think there is any point of making a separate file. But let me know if you think otherwise 

Note: adding failing files to `exclude` will not remove hte error, it will remove the entire file from the ts compilation, causing more errors

("noUncheckedIndexedAccess": true, will be added soon, probably after  https://github.com/lightdash/lightdash/pull/13216) 

Once enabled, I saw most of the `any`  errors come from the `unknown in catch` . 

- Fixed all `catch` error in common and warehouses
- Kept ` "useUnknownInCatchVariables": false` on backend and frontend, so we can fix them separately. 

I believe there are a couple of extra `any`s  on commercial, we'll have to make another PR once it is merged

Next steps: 

- Create a ticket to solve all 42 `any` errors on backend
- Create a ticket to solve all 7 `any` errors on frontend
- fix all `unexpected any` errors: just in common there are 86 errors. 

### Reviewer actions


packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -215,9 +218,10 @@ export class BigqueryWarehouseClient extends WarehouseBaseClient<CreateBigqueryC
             });
 
             await streamPromise;
-        } catch (e) {
-            const response = e?.response as bigquery.IJob;
-            const responseError = response?.status?.errorResult || e;
+        } catch (e: unknown) {
+            const response = (e as { response?: bigquery.IJob })?.response;
+            const responseError =
+                response?.status?.errorResult || (e as bigquery.IErrorProto);
```
IrakliJani, line 224: this is still assuming with `as` and can be problematic when we update lib version
I suggest a small object checker that complies with IErrorProto

packages/frontend/src/features/sqlRunner/store/thunks.ts
```diff
@@ -69,7 +70,10 @@ export const runSqlQuery = createAsyncThunk<
                 return rejectWithValue(job.error);
             }
         } catch (error) {
-            return rejectWithValue(error?.error as ApiErrorDetail);
+            if (isApiError(error)) {
+                return rejectWithValue(error.error);
+            }
+            throw error;
```
IrakliJani, line 76: nice!

.eslintrc.js
```diff
@@ -10,5 +10,6 @@ module.exports = {
         // Allows the use of 'void' operator for promises we deliberately don't wait for,
         // providing a way to bypass the 'no-floating-promises' rule when necessary.
         'no-void': ['error', { allowAsStatement: true }],
+        "@typescript-eslint/no-explicit-any": "warn"
```
IrakliJani, line 13: warn or error? I think tsc is erroring after the config change right?

IrakliJani (approved):

.eslintrc.js
rephus: for now, warn, once we fix the issues, we can change it to warn (see other related tickets)

packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
rephus: yeah, I agree, we should not use `as` anywhere. I fixed this

Issue #13378, opened 2025-01-22
I want to be able to control the metrics that are shown in Spotlight


joaoviana:
### Description

Acceptance Criteria
- [ ] I can control the default visibility of my metrics in Spotlight through .yml 
- [ ] The default visibility is set to "show", meaning that all metrics will be shown in the metrics catalog.
- [ ] I can change this configuration by setting it to "hide" 
- [ ] This global configuration should live in a project-level file called "lightdash.config.yml"
- [ ] This file is optional, and can be empty if it exists - we should configure it back to defaults if thats the case.
- [ ] I can be even more specific and control the visibility of metrics on a model level. This should override the default visibility set in the global yml config just for this model and its metrics. 
- [ ] I can also override this config on a metric level. This will override the global + model -level configs. 
- [ ] The values supported are: "show" and "hide"
- [ ] Users can make these changes through the CLI and/or Refreshing dbt on the UI, depending on their project. 

Internal slack thread:  https://lightdash.slack.com/archives/C02GQKJK84Q/p1737460425203049?thread_ts=1737039661.531999&cid=C02GQKJK84Q
Internal final proposal notion doc: https://www.notion.so/lightdash/Proposals-17da63207a7a80ac953dd2da5f9c388d?pvs=4#17da63207a7a8068afb0d74674e2ff0f

PR #13381, opened 2025-01-22
feat: add visibility config to metrics via yaml

almeidabbm:
Closes: #13378 

### Description:

Allows users to control a metric's visibility in Spotlight through yml files

#### How to test

Create `lightdash.config.yml` (this is not required and can be empty) in `/examples/full-jaffle-shop-demo/dbt/lightdash.config.yml`

```yml
spotlight:
  default_visibility: hide # Optional, defaults to "show"

```

Next time you refresh on UI/deploy through the cli it will hide all metrics on the catalog

Once setting all to `hide`, go to a model yml file, like `events.yml` and set this model's visibility to "show":

```yml
version: 2
models:
  - name: events
    meta:
      spotlight:
        visibility: show
      group_details:
        events:
          label: Events
          description: Events that happened in the Jaffle Shop
        nested_group:
          label: nested group
        another_nested_group:
          label: another nested group
    columns:
      - name: timestamp_tz
        description: ""
...
```

You see only event's metrics shown.

image

Now, set the model's visibility from 'show' to 'hide', and just show one metric "in_gbp"

```yml
...
  in_gbp:
              type: sum
              format: "gbp"
              spotlight:
                visibility: show
... 
```

once refreshed, only this metric will be visible: 
image

 
### Reviewer actions


examples/full-jaffle-shop-demo/dbt/lightdash.config.yml
```diff
@@ -0,0 +1,2 @@
+spotlight:
+  default_visibility: show # Optional, defaults to "show"
```
rephus, line 2: :+1:

packages/backend/src/database/migrations/20250122161125_add_spotlight_visibility_column_to_catalog_search.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from 'knex';
+
+const CatalogTableName = 'catalog_search';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(CatalogTableName, (table) => {
+        table.boolean('spotlight_show').defaultTo(true).notNullable();
```
rephus, line 7: optimization suggestion: 
if you're planning to filter by this, we should also add an index

if you're not planning to filter by this, we could make/use a jsonb column, so we don't need to add migrations for all the flags

packages/backend/src/database/migrations/20250122161125_add_spotlight_visibility_column_to_catalog_search.ts
```diff
@@ -0,0 +1,15 @@
...
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(CatalogTableName, (table) => {
```
rephus, line 12: we normally check if column exists first, just to avoid unnecesary errors 

eg:
```
  if (await knex.schema.hasColumn(dashboards, 'slug')) {
        await knex.schema.alterTable(dashboards, (table) => {
            table.dropColumn('slug');
        });
    }
    ```

File: packages/backend/src/models/CatalogModel/CatalogModel.ts [hunk 4]
```diff
@@ -305,6 +313,13 @@ export class CatalogModel {
                 );
             });
 
+        if (context === CatalogSearchContext.SPOTLIGHT) {
+            catalogItemsQuery = catalogItemsQuery.where(
```
rephus, line 317: I can see a filer here, we should add an index for this column

packages/backend/src/models/CatalogModel/utils/index.ts
```diff
@@ -1,34 +1,49 @@
...
 import { DbCatalogIn } from '../../../database/entities/catalog';
+import { DbTag } from '../../../database/entities/tags';
+
+const getSpotlightShow = (
+    spotlight?: Explore['spotlight'] | Metric['spotlight'],
+) => {
+    const visibility =
+        spotlight?.visibility || DEFAULT_SPOTLIGHT_CONFIG.default_visibility;
+
+    return visibility === 'show';
```
rephus, line 28: should this show be an enum ?

packages/backend/src/projectAdapters/dbtBaseProjectAdapter.ts
```diff
@@ -69,6 +78,35 @@ export class DbtBaseProjectAdapter implements ProjectAdapter {
...
+    private async getLightdashProjectConfig(): Promise<LightdashProjectConfig> {
+        if (!this.dbtProjectDir) {
+            return {
+                spotlight: DEFAULT_SPOTLIGHT_CONFIG,
+            };
+        }
+
+        const configPath = path.join(
+            this.dbtProjectDir,
+            'lightdash.config.yml',
```
rephus, line 90: oh, I see, so we're adding a new lightdash config file. This seems  a bit overcomplicated. Have you considered using ENV variables instead ?

joaoviana: We gathered this info from a proposal sent to like ~6 external customers, and they like the managing being done globally (on a separate yml file) + model + metric overrides. 
We are also considering letting them define directories (on the global yml config file) to configure all models in that said directory as most have lots of models/metrics. 
We considered just doing model / level overrides but people want a better devx. You can see more info on this as well on the links shared on the main issue 💪

packages/backend/src/models/CatalogModel/utils/index.ts
joaoviana: We considered this but the type is already good enough for intellisense /validation

packages/backend/src/database/migrations/20250122161125_add_spotlight_visibility_column_to_catalog_search.ts
joaoviana: TODO: Add this

packages/backend/src/models/CatalogModel/CatalogModel.ts
joaoviana: TODO: set this column to an index

joaoviana: done here: https://github.com/lightdash/lightdash/pull/13381/commits/1dbdbde9b4e4f416cdcdaeea2726f79ae43a15b5

packages/backend/src/database/migrations/20250122161125_add_spotlight_visibility_column_to_catalog_search.ts
joaoviana: done here! https://github.com/lightdash/lightdash/pull/13381/commits/1dbdbde9b4e4f416cdcdaeea2726f79ae43a15b5

packages/backend/src/database/migrations/20250122161125_add_spotlight_visibility_column_to_catalog_search.ts
joaoviana: oops same commit as above but included this change request https://github.com/lightdash/lightdash/commit/1dbdbde9b4e4f416cdcdaeea2726f79ae43a15b5

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -245,7 +250,17 @@ export class ExploreCompiler {
             this.compileJoin(j, includedTables),
         );
 
+        const spotlightVisibility =
+            meta.spotlight?.visibility ?? spotlightConfig?.default_visibility;
```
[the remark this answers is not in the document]
joaoviana: this is a great point @rephus 💯 since we're doing a similar logic to other type of config (categories) then I'll do this util/reusable function on that PR 👍

joaoviana (approved):

Issue #13358, opened 2025-01-21
I want to add Danish Krone as a currency format option


jpetey75:
### Description

The format value can be `dkk` and the currency symbol follows the number, e.g. `865kr`. 

- also need to update [dimension](https://docs.lightdash.com/references/dimensions#format) and [metric](https://docs.lightdash.com/references/metrics/#format) docs when the change is made

jpetey75:
There is a recent PR to add Japanese Yen so I expect the update will be similar: https://github.com/lightdash/lightdash/pull/13346/

PR #13397, opened 2025-01-23
feat: add support for danish krone

magnew:
Closes: #13358

### Description:

Adds support for Danish Krone. This is currently the most straightforward approach, which doesn't localize Krone correctly. With this we will get the international version of the krone format, which looks like `DKK 1.50` instead of the Danish local version which looks like `1.50 kr.`. ~~We have a bit of an issue where we allow 'custom' separators, but using the localizer that would create the Danish version of the currency would override selected separators.~~

In standup we decided to ship this as `DKK <number>` 

### Reviewer actions


packages/e2e/cypress/e2e/app/formats.cy.ts
```diff
@@ -5,7 +5,7 @@ describe('Explore', () => {
         cy.login();
     });
 
-    it('Should query in explore with formats and rounds', () => {
+    it.only('Should query in explore with formats and rounds', () => {
```
rephus, line 8: leftover only

rephus (approved):
Remove the .only in the test, then it is good to go. 

My thought is that we don't need to add all possible currencies to the YML, since we support a more custom formatting in the UI . Still, we can merge this, people might wonder why we support DKK and not other more popular rcurrencies. 

image

We should probably update the docs too

packages/e2e/cypress/e2e/app/formats.cy.ts
magnew: Ooops! Good catch, thank you

magnew:
> My thought is that we don't need to add all possible currencies to the YML, since we support a more custom formatting in the UI . Still, we can merge this, people might wonder why we support DKK and not other more popular rcurrencies.

Yea, agreed. I brought that concern up but it sounds like we wanna do it. I'll look into docs.

magnew:
There are two failing tests that pass locally. One is a flakey one I will post a PR for. The other is known. Merging this.

PR #13418, opened 2025-01-27
chore: no unchecked index values

rephus:
Closes:  https://github.com/lightdash/lightdash-internal-work/issues/2800
### Description:

Add an "warning" eslint rule so we can incrementaly fix these errors before we can apply `noUncheckedIndexedAccess` to tsconfig compile . 

On common and backend, I listed all existing warnings, and excluded from the rule, all other "new" files should error

image

image

I fixed only a few, as an example, but there are more. IMO we should fix them in batches, it is a shame we can't simply ignore the existing errors, to prevent these in the future.

Errors in packages: 

- [ ] Found 409 errors in backend 
- [ ] Found 458 errors in 122 files frontend
- [ ] Found 93 errors. in common (163 with tests) 
- [ ]  Found 29 errors. in warehouses (35 with tests)
- [ ]  Found 23 errors. in cli

### Reviewer actions


rephus:
Once approved, write some estile guide for code

packages/backend/tsconfig.index.json
```diff
@@ -0,0 +1,20 @@
+{
+    // This file will extend the tsconfig.json file for the test environment
+    // This means we will not apply the noUncheckedIndexedAccess option to the test files
+    // Change build method in package.json to use this file
+    // It is temporary disabled in package.json until we fix all the errors
```
rephus, line 5: IN order to permanently change this file in package.json, we'll have to modify also the dockerfile, since that file is `COPY`ed separately.

rephus: unfortunately, I don't think  there is any way to partilaly activate this rule (or just warn) . So until we fix all errors, we can't use it.

packages/backend/tsconfig.index.json
```diff
@@ -0,0 +1,20 @@
...
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "noUncheckedIndexedAccess": true
+    },
+    "include": [
+        "src/**/*.ts",
+        "src/**/*.json"
+    ],
+    "exclude": [
```
rephus, line 14: Adding the files to exclude will exclude them from all the tsconfig rather than just that rule 

How I tested this:
- I added an unused variable to a test file, which is excluded in my noUncheckedIndexedAccess file
- I don't get an error,
- I remove the `        "src/*/.test.ts",` exclusion from my extra file
- I get the error now

packages/backend/src/routers/headlessBrowser.ts
```diff
@@ -13,13 +18,14 @@ export const getAuthenticationToken = (value: string) =>
...
-        const { userUuid } = req.params;
+        const userUuid = getObjectValue(req.params, 'userUuid');
         const hash = getAuthenticationToken(userUuid);
 
         if (hash !== req.body.token) {
             throw new ForbiddenError();
         }
-        const sessionUser = await req.services
+
+        const sessionUser = await (req as AnyType).services
```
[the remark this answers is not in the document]
almeidabbm: Interesting you had to do this, I checked out this branch to see what might be happening but I do have services defined if I remove the `req as AnyType`
image

almeidabbm (approved):

packages/backend/src/routers/headlessBrowser.ts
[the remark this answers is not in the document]
rephus: I can try reverting this back, now we are using lint, this might be a compile thing...

rephus:
I believe e2e errors are unrelated

PR #13451, opened 2025-01-29
fix: sql runner error serialization error

magnew:
### Description:

Fixes an error in the SQL runner where non-serializable objects (Error) were being stored. This makes it use the Serializable error type.

To test:
- Run a query like `SELECT * FROM "postgres"."jaffle"."orders" `
- Switch to chart tab
- Switch back to SQL tab
- Run a query on a different table, causing a chart error `SELECT * FROM "postgres"."jaffle"."events"`
- The console will show and error about the chart being invalid, but NOT the below serialization error. 

Error before:
Screenshot 2025-01-29 at 15 24 58

### Reviewer actions


rephus (approved):
Tested on render, look good

Issue #13410, opened 2025-01-24
Validation should skip charts that are not saved and not on any active dashboard


jpetey75:
### Description

From Jose: Looks like the bug in on validation that should exclude charts not in the latest dashboard version.

https://lightdash.slack.com/archives/C02DGV9RWKV/p1737746357851409?thread_ts=1737742617.730939&cid=C02DGV9RWKV

### Steps to Reproduce the Bug or Issue

1. create dashboard and put some charts on it
2. delete the dashboard
3. update or remove some dimensions used in those charts
4. validation throws errors even though those charts on not used anywhere 

### Cloud or self-hosting

None

VikramNagwal:
Can I join the Slack workspace, or is it restricted to members only?

rephus:
Hi @VikramNagwal that link points to a private conversation in our private slack workspace. But you are welcome to join our open slack community, some of our github links points there. Here is the link: https://join.slack.com/t/lightdash-community/shared_invite/zt-2ehqnrvqt-LbCq7cUSFHAzEj_wMuxg4A

VikramNagwal:
thanks @rephus

PR #13453, opened 2025-01-29
fix: validation orphaned charts

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13410

### Description:

Before:

image

After:

Screenshot from 2025-01-29 17-20-15

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -1189,7 +1189,7 @@ export class DashboardModel {
     }
 
     async getOrphanedCharts(
-        dashboardUuid: string,
+        dashboardUuids: string[],
     ): Promise<Pick<SavedChart, 'uuid'>[]> {
         const getChartsInTilesQuery = this.database(DashboardTileChartTableName)
             .distinct('saved_chart_id')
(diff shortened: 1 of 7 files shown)


packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -1189,7 +1189,7 @@ export class DashboardModel {
     }
 
     async getOrphanedCharts(
-        dashboardUuid: string,
+        dashboardUuids: string[],
```
rephus, line 1192: changed spec, this method was only used in non-critical 2 places.

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -1203,11 +1203,18 @@ export class DashboardModel {
                 `${DashboardsTableName}.dashboard_id`,
                 `${DashboardVersionsTableName}.dashboard_id`,
             )
-            .where(`${DashboardsTableName}.dashboard_uuid`, dashboardUuid);
-
+            .whereIn(`${DashboardsTableName}.dashboard_uuid`, dashboardUuids)
+            .where(
```
rephus, line 1207: I believe this filter by last version was missing, and it was getting wrong results, since it was checking the chart across all versions.

ZeRego: @rephus I think we need to add a comment to the top of this method.
Check this PR: https://github.com/lightdash/lightdash/pull/11016
> backend will only delete orphans if, and only if, they do not belong to any tile. This means that version reverting will now work for charts created within the dashboard, even if they get removed from the tile in the next dashboard version save.

rephus:
Closed in favor of https://github.com/lightdash/lightdash/pull/13466

Issue #13398, opened 2025-01-24
Investigate Google Sheet Syncs being disabled


oli-rmsy:
### Description

Google sheet sync is being created but at some point is turned off. This can be a day or two later.

Link to internal slack thread: https://lightdash.slack.com/archives/C071TU6FZ3P/p1737560364578869

### Steps to Reproduce the Bug or Issue

No repro

### Cloud or self-hosting

cloud

PR #13461, opened 2025-01-30
feat: handle sheets transient err and send email if disabled

joaoviana:
Closes: [#13398](https://github.com/lightdash/lightdash/issues/13398)

### Description:

If the error is 500, we can retry the scheduled delivery, considered a Transient error
If the scheduler is disabled, we will notify the user that created the google sheets sync

we're disabling google sheets syncs even if the error is to do with google api. We should retry 
image

let's also monitor how this goes.

### Reviewer actions


almeidabbm (approved):

joaoviana:
image

Issue #13464, opened 2025-01-30
Dashboard charts color is inconsistent


almeidabbm:
### Description

Currently when there is no color assigned via `yaml` we calculate the colors in `packages/frontend/src/hooks/useChartColorConfig/useChartColorConfig.tsx` by relying on indexes of already used colors and iterate that way.

This indexing is highly dependent on which visualization is rendered first.
For example, if a chart has 3 series and is the first to get its results, it will get colors from the palette with indexes 0, 1 and 2
When the second chart gets its results (assuming they are all different series), it will be assigned colors 3, 4 and 5

When a user reloads there's the chance that the chart that got the results last in the previous render, could get its results first now, so the colors will be 0, 1 and 2 for chart 2 and 3, 4 and 5 for chart 1.

### Steps to Reproduce the Bug or Issue

1. Create a dashboard with 2 charts that have different series
2. Reload the chart (this is by chance, but there will be times that the colors switch around)

https://github.com/user-attachments/assets/c53c6c22-6b33-4d77-aa53-cda34e0eb8bc

### Cloud or self-hosting

None

PR #13465, opened 2025-01-30
fix: chart color inconsistent on reload

almeidabbm:
Closes: #13464

### Description:

- Uses a more deterministic approach to color selection
- Include group in color selection so that it can bring colors in the same group to closer indexes

### Reviewer actions


almeidabbm:
Closing as this needs more thought. Current proposed solution is more deterministic at the expense of higher chance of repeating colors within the same chart.

Issue #13474, opened 2025-01-31
Filtering by a boolean False dimension doesn't work on dashboards


rephus:
### Description

I believe this is because we check for `!value` , where this field can be undefined, but also, `false`. It works well with `true` values

![Image](https://github.com/user-attachments/assets/d9860827-05b5-4217-acab-ee5ef7ee5b37)
![Image](https://github.com/user-attachments/assets/9a1bb31b-08b8-4a63-8d33-7fbaabe947c1)

### Steps to Reproduce the Bug or Issue

.

### Cloud or self-hosting

None

zhyd1997:
Hi @rephus 

I can work on it if the issue has low/no priority.

PR #13484, opened 2025-02-01
fix(frontend): handled false boolean value for filters.

zhyd1997:
Closes: #13474  

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
@@ -5,6 +5,7 @@ import {
 } from '@lightdash/common';
 import { Menu, Text } from '@mantine/core';
 import { IconFilter } from '@tabler/icons-react';
+import isNil from 'lodash/isNil';
 import { type FC } from 'react';
 import MantineIcon from '../common/MantineIcon';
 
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
```diff
@@ -32,7 +33,7 @@ export const FilterDashboardTo: FC<Props> = ({ filters, onAddFilter }) => {
                             null
                         </Text>
                     )}
-                    {filter.values && filter.values[0] && (
+                    {filter.values && !isNil(filter.values[0]) && (
```
mhtiti, line 36: I think using open chaining will be better: ```filter.values?.[0] ```

zhyd1997: Hi @mhtiti 

1. `foo?.[0]` will be transpiled as `foo && foo[0]` **ES5** code.
2. for any _type_ of `foo`, it will not throw an error when using `[0]` access, it will just return `undefined`.

Here, we just need to filter out `undefined` and `null` values, except for `0`, `false` and `''`, so I use `lodash/isNil` method`.

Basit-Balogun10:
Hey @zhyd1997 
The changes LGTM, well done 🙌🏾 

Is there a way to quickly test this in a preview environment or deployment without pulling the branch locally? I've been looking around but haven't found an option.

Cc: @ZeRego @joaoviana – if anyone has suggestions on a quick way to test PR changes, that would be super helpful for faster reviews!

notgiorgi (approved):
tested, LGTM

notgiorgi:
@all-contributors add @zhyd1997 for code

Issue #11855, opened 2024-10-08
`lightdash generate` does not work with models materialized as `materialized_view` on PG warehouse


almeidabbm:
**Steps to reproduce**

⚠️ Using Postgres

1. Go to `dbt_project.yml` and change the models `materialized` to `materialized_view`
2. Run `lightdash run dbt --full-refresh` so that the models get re-materialized
3. Run `lightdash generate -s {your_materialized_view_model}`
4. You should see the following error: ![Image](https://github.com/user-attachments/assets/03ab6df5-f86d-4151-9e1c-e6c304e6e1c1)

PR #13494, opened 2025-02-03
fix: use current_database instead of matviewowner

almeidabbm:
Closes: #11855

### Description:

- Check current database against list of databases instead of `matviewowner`

Previously we had a where clause when listing models materialized as `materialized_view` that checked if `database` matched `matviewowner` - this is the user that materialized the model and not the database where the materialized view exists. Therefore if `user` was different from `database` name, it would fail finding the mat views.

To reproduce:
1. Change db that has contains jaffle schema to something other than `postgres` so that it is different from the user
2. in `dbt_project.yml` change `staging` to materialize as `materialized_view`
3. run `lightdash dbt run`
4. run `lightdash generate`

**Before**

image

**After**
image

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
@@ -354,7 +354,7 @@ export class PostgresClient<
 
             UNION ALL
 
-            SELECT mv.matviewowner AS table_catalog,
+            SELECT current_database() AS table_catalog,
                 n.nspname AS table_schema,
                 c.relname AS table_name,
                 a.attname AS column_name,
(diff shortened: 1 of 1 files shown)


packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -364,7 +364,7 @@ export class PostgresClient<
             JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
             JOIN pg_catalog.pg_matviews mv ON n.nspname = mv.schemaname AND c.relname = mv.matviewname
             WHERE c.relkind = 'm'
-            AND mv.matviewowner IN (${Array.from(databases)})
+            AND current_database() IN (${Array.from(databases)})
```
almeidabbm, line 367: This might be a redundant check 🤔

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
```diff
@@ -354,7 +354,7 @@ export class PostgresClient<
 
             UNION ALL
 
-            SELECT mv.matviewowner AS table_catalog,
+            SELECT current_database() AS table_catalog,
```
rephus, line 357: SOoooo, this is a nice approach, BUT, we need to keep in mind not all postgres versions might support this function. 

After some research , it seems this function was introduced on postgres 7 https://www.postgresql.org/docs/7.3/functions-misc.html, which was released on 2002, so 21 years ago !. 

I think it is safe to use :smile:

rephus (approved):

packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
almeidabbm: Nice point, we actually check for PG version to be >= 12 to query matviews so it should be fine yeah 🙇

PR #13503, opened 2025-02-04
fix: convert custom metric filters

rephus:
Closes: 

### Description:

image

image

### Reviewer actions

test-backend test-frontend


packages/common/src/compiler/filtersCompiler.test.ts
```diff
@@ -806,3 +812,197 @@ describe('Filter SQL', () => {
...
+        expect(convertMetricFilterToDbt(filters)).toEqual(expected);
+    });
+
+    it('should convert IN_THE_NEXT and IN_THE_PAST filters correctly', () => {
+        const filters: MetricFilterRule[] = [
+            {
+                target: { fieldRef: 'field1' },
+                id: '1',
+                operator: FilterOperator.IN_THE_NEXT,
+                values: ['14 days'],
```
ZeRego, line 962: The unit of time is in the `settings`:
  eg:
  ```
              {
                "id": "08d5781b-203a-43b9-a34a-01e14016699c",
                "operator": "inThePast",
                "values": [
                    1
                ],
                "settings": {
                    "unitOfTime": "years",
                    "completed": false
                },
                "target": {
                    "fieldRef": "events.date"
                }
            },
```
Should we return error when `completed: true` since our YML syntax doesn't support it?

File: packages/common/src/compiler/filtersCompiler.test.ts [hunk 2]
```diff
@@ -806,3 +812,197 @@ describe('Filter SQL', () => {
...
+        ];
+        expect(convertMetricFilterToDbt(filters)).toEqual(expected);
+    });
+    it('should convert NULL filters correctly', () => {
+        const filters: MetricFilterRule[] = [
+            {
+                target: { fieldRef: 'field2' },
+                id: '1',
+                operator: FilterOperator.EQUALS,
+                values: [null],
```
ZeRego, line 860: Should we test the IS_NULL operator? I don't think we usually use EQUALS + null value

```
{
                "id": "15204703-20d0-4b20-9a3c-277963bc2356",
                "operator": "isNull",
                "values": [
                    1
                ],
                "target": {
                    "fieldRef": "events.event"
                }
            },
```

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -485,3 +493,88 @@ export const renderFilterRuleSql = (
...
+        throw new ParameterError(
+            `Filter values are undefined for filter: ${JSON.stringify(filter)}`,
+        );
+    }
+    // Value can be null
+
+    switch (operator) {
+        case FilterOperator.EQUALS:
+        case FilterOperator.NULL:
+            return `${values[0]}`;
```
ZeRego, line 518: What happens when there are multiple values?
eg:
```
{
    "id": "3c0ca671-26b9-4cd1-b15e-5786740273a3",
    "operator": "equals",
    "values": [
        "1",
        "2",
        "3"
    ],
    "target": {
        "fieldRef": "orders.order_id"
    }
}
```

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -485,3 +493,88 @@ export const renderFilterRuleSql = (
         }
     }
 };
+
+/**
+ * Converts a filter rule from custom metrics to dbt meta tags in yml
+ * More info about dbt filters: https://docs.lightdash.com/references/metrics/#available-filter-types
+ */
+const convertFilterOperatorToDbt = (filter: MetricFilterRule): string => {
```
ZeRego, line 501: Should we move this to `packages/common/src/types/filterGrammar.ts`? Have a method to encode and another to decode. 
We can amend the unit tests to convert back and forward? yml -> filter object -> yml

```
it('Should parse date in the next operator with interval', () => {
        const filtersYml = [{ name: 'inTheNext 14 years' }];
        const filtersJson = [{
                id: undefined,
                operator: FilterOperator.IN_THE_NEXT,
                settings: {
                    unitOfTime: 'years',
                },
                target: {
                    fieldRef: 'name',
                },
                values: [14],
            }]
        expect(removeIds(parseFilters(filtersYml)).toStrictEqual(filtersJson);
        expect(convert(filtersJson).toStrictEqual(filtersYml);
    });
```

packages/common/src/compiler/filtersCompiler.ts
rephus: I forgot about that, I'll fix it

packages/common/src/compiler/filtersCompiler.test.ts
rephus: or add +1 to the value ..

packages/common/src/compiler/filtersCompiler.ts
rephus: this should be the right syntax to allow `OR` values  https://docs.lightdash.com/references/metrics/#available-filter-types
image

packages/common/src/types/filterGrammarConversion.ts
```diff
@@ -0,0 +1,130 @@
...
+                    settings.completed &&
+                    operator === FilterOperator.IN_THE_NEXT
+                ) {
+                    totalValue += 1;
+                }
+                if (
+                    settings.completed &&
+                    operator === FilterOperator.IN_THE_PAST
+                ) {
+                    totalValue -= 1;
```
ZeRego, line 84: I would throw `NotImplementedError` when complete=true rather than returning the wrong filter logic.
Filter in UI
`date is in the last 1 completed year` (2024.01.01 - 2024.12.31)
converted to YML (1-1=0)
`date is in the last 0 year` 🤔

packages/common/src/types/filterGrammarConversion.test.ts
```diff
@@ -0,0 +1,298 @@
...
+});
+
+describe('convert from filterGrammar', () => {
+    it('should return undefined if filters are undefined', () => {
+        expect(convertMetricFilterToDbt(undefined)).toBeUndefined();
+    });
+
+    it('should return an empty array if filters are an empty array', () => {
+        expect(convertMetricFilterToDbt([])).toEqual([]);
+    });
```
ZeRego, line 253: duplicate tests

packages/common/src/types/filterGrammarConversion.test.ts
```diff
@@ -0,0 +1,298 @@
...
+                values: ['value3', 'value4'],
+            },
+        ];
+        expect(() => convertMetricFilterToDbt(filters)).toThrow(
+            'No function implemented to convert custom metric filter to dbt: inBetween',
+        );
+    });
+});
+
+describe('convert from filterGrammar', () => {
```
ZeRego, line 246: Love these tests! ❤️ testing both functions at the same time!

ZeRego (approved):
lgtm
Just left a note about the completed=true scenario

PR #13505, opened 2025-02-04
fix: update manifestV12.json 

joaoviana:
Closes: N/A

### Description:

From: https://schemas.getdbt.com/

### Reviewer actions

test-backend
test-cli

### DIFF
--- a/packages/common/src/dbt/schemas/manifestV12.json
+++ b/packages/common/src/dbt/schemas/manifestV12.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft/2020-12/schema#",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
     "type": "object",
     "title": "WritableManifest",
     "properties": {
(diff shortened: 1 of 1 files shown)


packages/common/src/dbt/schemas/manifestV12.json
```diff
@@ -8108,13 +8975,7 @@
                                             "offset_to_grain": {
                                                 "anyOf": [
                                                     {
-                                                        "enum": [
```
rephus, line 8111: string > enum

Should be ok to replace

rephus (approved):
Changes look good, afaik this is what dbt is supposed to do . 

However, I can imagine some cases where the manifest is valid for dbt 18, but it will not work.  I'm wondering if we should have an early v12 version for dbt 18 and have an incremental v12 with the latest changes. 

This is not a blocker, we can always rollback this file and make a new v12 manifest.

joaoviana:
> Changes look good, afaik this is what dbt is supposed to do .
> 
> However, I can imagine some cases where the manifest is valid for dbt 18, but it will not work. I'm wondering if we should have an early v12 version for dbt 18 and have an incremental v12 with the latest changes.
> 
> This is not a blocker, we can always rollback this file and make a new v12 manifest.

sure, understood! Can always refer back to this PR to create a v12-1.9-and-later or something 😅

Issue #13480, opened 2025-01-31
On DBT refresh, replace all custom metrics with metrics if compatible


rephus:
### Description

Proposal: https://www.notion.so/lightdash/Write-back-custom-metrics-to-dbt-189a63207a7a8092900deb2a300cfe16

Technical implementation suggestion: 

- on refresh dbt method, get all content (see more details about content below) 
- find any custom metric reference, 
- check if the sanitized sql of the custom metric matches any sql on the explore metric
- If so, replace that custom metric with the metric
- save a version (if available) 

 Content that should be updated is: 

- charts (metricquery, filters) 
- scheduled deliveries (filters)
- threshold alerting (threshold) 
- any other ?

### Open questions

- What do we do if names are different ? 
- What happens if custom metrics has filters/formatting ?

TuringLovesDeathMetal:
Things to test:

- if you've changed the axis labels, do they stay customized? (e.g. I use a custom metric, I create a chart and update the y-axis label, my metric gets replaced by a YAML metric, is my y-axis labelling still customized?)

PR #13511, opened 2025-02-04
feat: replace custom metrics with matched model metric after project compile

ZeRego:
Closes: [#13480](https://github.com/lightdash/lightdash/issues/13480)

### Description:

After a successful project compile, we replace custom metrics with a matching model metric.
Notes:
- feature is behind a feature flag for testing purposes
- metric needs to match: id, label, SQL, base dimension, table, metric type, format, percentile, filters
- at the moment we only support metrics with no format & percentile
- the replacement happens asynchronously so it might take some time after the project is compiled
- because of the previous point, it does not invalidate frontend cache

Follow-up work:
- Explorer page replaces custom metrics automatically
- Send notification/email/slack message to admin/developers about charts updated and suggestions

Demo:
https://github.com/user-attachments/assets/eef7fed4-1fc9-4d3c-8802-a35ea4b14876

### Reviewer actions


packages/common/src/utils/fields.ts
```diff
@@ -78,3 +79,149 @@ export const getFieldsFromMetricQuery = (
...
+        (condition) => condition.isMatch,
+    );
+    const isSuggestedMatch = Object.values(conditions)
+        .filter((condition) => condition.requiredForSuggestion)
+        .every((condition) => condition.isMatch);
+    if (
+        conditions.fieldIdMatch.isMatch &&
+        getItemId(metric) === 'customers_customer_id_average_of_customer_id'
+    ) {
+        console.log('isExactMatch', isExactMatch);
```
ZeRego, line 157: todo: remove debug console logs

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -450,6 +451,38 @@ export class SavedChartModel {
         };
     }
 
+    static convertDbSavedChartAdditionalMetricToAdditionalMetric(
```
ZeRego, line 454: Extracted from line 905

packages/common/src/utils/fields.ts
```diff
@@ -78,3 +79,149 @@ export const getFieldsFromMetricQuery = (
         ...additionalMetrics,
     };
 };
+
+export function compareMetricAndCustomMetric({
```
ZeRego, line 83: TODO: add unit tests

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -896,6 +903,27 @@ export default class SchedulerTask {
...
                 });
             }
+            const canReplaceCustomMetrics = await isFeatureFlagEnabled(
+                FeatureFlags.ReplaceCustomMetricsOnCompile,
+                {
+                    userUuid: user.userUuid,
+                    organizationUuid: user.organizationUuid,
+                },
+                {
+                    // because we are checking this in the health check, we don't want to throw an error
```
rephus, line 913: we might need to update this comment, this does not happen on healthcheck

packages/common/src/utils/charts.ts
```diff
@@ -0,0 +1,64 @@
+import {
+    type CreateSavedChartVersion,
+    type ReplaceCustomFields,
+    type SkippedReplaceCustomFields,
+} from '../types/savedCharts';
+import { getItemId } from './item';
+
+export function maybeReplaceFieldsInChartVersion({
```
rephus, line 8: why is this common ?

packages/common/src/utils/fields.ts
```diff
@@ -78,3 +83,167 @@ export const getFieldsFromMetricQuery = (
...
+                    }),
+                ),
+            requiredForSuggestion: true,
+        },
+    };
+
+    const isExactMatch = Object.values(conditions).every(
+        (condition) => condition.isMatch,
+    );
+    const isSuggestedMatch = Object.values(conditions)
```
rephus, line 158: :ok_hand:

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -5373,4 +5381,111 @@ export class ProjectService extends BaseService {
...
+        userUuid: string;
+        organizationUuid: string;
+        projectUuid: string;
+        replaceFields: ReplaceCustomFields;
+        skipChartsUpdatedAfter: Date;
+    }): Promise<Array<Pick<SavedChartDAO, 'uuid' | 'name'>>> {
+        const updatedChartPromises = Object.entries(replaceFields).map(
+            async ([chartUuid, fieldsToReplace]) => {
+                const chart = await this.savedChartModel.get(chartUuid);
+                if (chart.updatedAt > skipChartsUpdatedAfter) {
```
rephus, line 5441: nice catch

rephus:
Rebuilding render

rephus (commented):
Code look good, let me test i

rephus (approved):
It works really well. Nicely done Screenshot from 2025-02-05 09-13-22

My only problem is that the code was a bit hard to follow, with different methods acrosso backend/common service/model/scheduler... 

Perhaps we might want to create a new service to unify these  `replacement` logic. Not a blocker

packages/common/src/utils/charts.ts
ZeRego: We need to have the same logic in the explore page/provider. 
When creating/editing a chart, if the explore has a metric that matches a custom metric we replace it in memory.

packages/backend/src/scheduler/SchedulerTask.ts
ZeRego: removed comment and timeout 👍

Issue #13481, opened 2025-01-31
Implement UI for custom metric write back


rephus:
### Description

full proposal: https://www.notion.so/lightdash/Write-back-custom-metrics-to-dbt-189a63207a7a8092900deb2a300cfe16

Design: https://www.notion.so/lightdash/Write-back-custom-metrics-to-dbt-189a63207a7a8092900deb2a300cfe16?pvs=4#189a63207a7a80fb81b9cea23d9b0f77

![Image](https://github.com/user-attachments/assets/a950d5ef-8ecb-48b3-b11e-0c2881387893)

PR #13535, opened 2025-02-05
feat: frontend custom metric write back

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13481

### Description:

[Screencast from 2025-02-05 13-36-24.webm](https://github.com/user-attachments/assets/1f1d5c7c-d017-4d10-83cf-32f30ee30a08)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomMetricWriteBackModal/FilterForm.tsx
+++ b/packages/frontend/src/components/Explorer/CustomMetricWriteBackModal/FilterForm.tsx
@@ -1,0 +1,131 @@
+import {
+    createFilterRuleFromField,
+    getFieldRef,
+    isDimension,
+    type ConditionalOperator,
+    type FieldTarget,
+    type FilterableDimension,
+    type FilterRule,
+} from '@lightdash/common';
+import { Button, Stack } from '@mantine/core';
+import {
+    useCallback,
+    type Dispatch,
+    type FC,
+    type SetStateAction,
+} from 'react';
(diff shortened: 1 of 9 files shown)


rephus:
STill missing: 
- check feature flags
- check github integration

TuringLovesDeathMetal:
Nice! This is looking really great, Javi :) 

Just a few small things:

1. Can we add `write back to dbt` in the same list, but before the `remove` item? So not under a separator, in a new place. 
image

2. clicking on `open pull request` should open the pull request, in a new tab. (instead of needing to click on the success toast to open the PR). 
image

3. I think that the YAML generator thing we're using is also linting the YAML? So it's replacing single quotes with double quotes in an unrelated field in the PR that it opened. I don't think we want this to happen!: 
image

4. The padding around the buttons seems really big (e.g. comparing it to the SQL runner write-back, I think the buttons are a lot closer to the borders?) 
image

5. I think that we can make some changes to teh copy here! Can we say 
```
Your pull request #10293 was successfully created on Github.

Once it is merged, refresh your dbt connection to see your updated metrics.
```

^The #10293 should be hyperlink which links to the open PR!

image

6. ^Can we also change the copy on the button from `back` to `close`

ZeRego (approved):

TuringLovesDeathMetal:
@rephus I think that (1), (3), and (6) from above are still there after the changes! But there might be a reason why they weren't changed, so let me know if that's teh case :)

rephus:
e2e errors don't seem related

Issue #13542, opened 2025-02-06
feat: allow for lightdash dbt run --assume-no flag


almeidabbm:
### Description

Currently we allow for `--assume-yes` when running `lightdash dbt run` which will overwrite all docs in `yaml` files.
When that flag is not used, the user will be prompted wether they would like to update the `yaml` files.

### What problem does this solve?

User having to go through each prompt to deny updating docs.

PR #13543, opened 2025-02-06
feat: add --assume-no option

almeidabbm:
Closes:  #13542

### Description:

- Adds `--assume-no` option to `dbt run` to skip all prompts to update yaml files

AC:
- `--assume-yes` should still work as is (write all files without prompting)
- default should still ask for files
- `--assume-no` should not prompt and not write files

Related docs PR: https://github.com/lightdash/lightdash-docs/pull/40

### Reviewer actions

### DIFF
--- a/packages/cli/src/handlers/dbt/run.ts
+++ b/packages/cli/src/handlers/dbt/run.ts
@@ -1,5 +1,5 @@
 import { getErrorMessage, ParseError } from '@lightdash/common';
-import { Command } from 'commander';
+import { Command, InvalidArgumentError } from 'commander';
 import execa from 'execa';
 import { LightdashAnalytics } from '../../analytics/analytics';
 import GlobalState from '../../globalState';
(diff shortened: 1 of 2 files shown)


packages/cli/src/handlers/dbt/run.ts
```diff
@@ -54,8 +60,11 @@ export const dbtRunHandler = async (
         });
         throw new ParseError(`Failed to run dbt:\n  ${msg}`);
     }
-    await generateHandler({
-        ...options,
-        excludeMeta: options.excludeMeta,
-    });
+
```
joaoviana, line 63: perhaps this would be useful?

```suggestion
if (options.assumeYes && options.assumeNo) {
    throw new Error("Cannot use both --assume-yes and --assume-no flags");
}
```

joaoviana (approved):

joaoviana:
Would be nice to have a little headway update 😉

packages/cli/src/handlers/dbt/run.ts
almeidabbm: Added in [chore: throw when both options are used](https://github.com/lightdash/lightdash/pull/13543/commits/176a82c7c063419204f89056fa147bae94b3d764)

joaoviana (approved):

Issue #13527, opened 2025-02-05
Replace custom metric with matching metrics in the explore page


ZeRego:
### Description

Related to: https://github.com/lightdash/lightdash/issues/13480

After project compile, the explore page should compare the custom metrics against the explore metrics.
If there is a match, it should replace them. 

If they have the same id but not a match it should show an error and not let them save the chart. In those cases, the user should rename the custom metric or remove it in favour of the explore metric.

### What problem does this solve?

Remove unnecessary custom metrics when I'm creating/updating a chart.
Inform user of potential issue:

PR #13550, opened 2025-02-06
feat: replace custom metric with matching metric during chart creatio…

ZeRego:
…n/update

Relates to: [#13527](https://github.com/lightdash/lightdash/issues/13527)

### Description:

When creating/editing a chart, if the chat has a custom metric that is an exact match of a model metric, it will be replaced.
The following PR will add error messages when metric and custom metric have the same id but aren't an exact match.

https://github.com/user-attachments/assets/ad5980e1-a8c0-42e1-b8e3-09eb5327b7bb

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/fields.ts
+++ b/packages/common/src/utils/fields.ts
@@ -215,28 +215,34 @@ export function findReplaceableCustomMetrics({
     );
 }
 
+export function convertReplaceableFieldMatchMapToReplaceFieldsMap(
+    replaceableFieldMap: ReplaceableFieldMatchMap,
+): ReplaceCustomFields[string]['customMetrics'] {
+    return Object.entries(replaceableFieldMap).reduce<
+        ReplaceCustomFields[string]['customMetrics']
+    >((acc2, [customFieldId, customField]) => {
+        if (customField.match) {
+            return {
+                ...acc2,
+                [customFieldId]: {
+                    replaceWithFieldId: customField.match.fieldId,
+                },
+            };
+        }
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/fields.ts
```diff
@@ -215,28 +215,34 @@ export function findReplaceableCustomMetrics({
     );
 }
 
+export function convertReplaceableFieldMatchMapToReplaceFieldsMap(
```
ZeRego, line 218: extract logic from util below.

packages/common/src/utils/fields.ts
```diff
@@ -215,28 +215,34 @@ export function findReplaceableCustomMetrics({
...
+                [customFieldId]: {
+                    replaceWithFieldId: customField.match.fieldId,
+                },
+            };
+        }
+        return acc2;
+    }, {});
+}
+
+// todo: rename+breakdown this util and all the replace/replaceable types to be more descriptive
```
ZeRego, line 236: something for later as it needs to update lots of files

packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
```diff
@@ -973,6 +975,24 @@ function reducer(
                 },
             };
         }
+        case ActionType.REPLACE_FIELDS: {
+            console.log(
```
rephus, line 979: leftover log ?

rephus (approved):

PR #13558, opened 2025-02-06
feat: custom metric duplicate id error

ZeRego:
Closes: [#13527](https://github.com/lightdash/lightdash/issues/13527) 

### Description:

Show error when there is a metric with the same ID, independently if the rest(sql, etc) is a match.

Screenshot 2025-02-06 at 15 27 14
Screenshot 2025-02-06 at 15 49 40
Screenshot 2025-02-06 at 18 40 22

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/types.ts
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/types.ts
@@ -31,6 +31,13 @@ export type TreeProviderProps = {
     itemsMap: Record<string, NodeItem>;
     selectedItems: Set<string>;
     missingCustomMetrics?: AdditionalMetric[];
+    itemsAlerts?: {
+        [id: string]: {
+            errors?: { message: string }[];
+            warnings?: { message: string }[];
+            infos?: { message: string }[];
+        };
+    };
     missingCustomDimensions?: CustomDimension[];
     groupDetails?: Record<string, GroupType>;
     isGithubIntegrationEnabled?: boolean; // For displaying the write back on custom metrics in TreeSingleNodeActions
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/Explorer/SaveChartButton/index.tsx
```diff
@@ -40,21 +55,40 @@ const SaveChartButton: FC<{ isExplorer?: boolean }> = ({ isExplorer }) => {
...
+                        isExplorer ? (
+                            <MantineIcon icon={IconDeviceFloppy} />
+                        ) : undefined
+                    }
+                    {...(isDisabled && {
+                        'data-disabled': true,
+                    })}
+                    sx={{
+                        '&[data-disabled="true"]': {
+                            pointerEvents: 'all',
```
ZeRego, line 84: Change to show tooltip when button is disabled.

rephus:
it would be useful to know what is similar and different.  I'll have to go to the SQL or YML to see the underlying sql and filter, and still it is not obvious. 

image


But we can improve it l ater

packages/frontend/src/components/Explorer/SaveChartButton/index.tsx
rephus: I hate we need to need to add a lot of boilerplate just to add a tooltip, it would be nice to have something like `title` , wihin the button component.

packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/types.ts
```diff
@@ -31,6 +31,13 @@ export type TreeProviderProps = {
     itemsMap: Record<string, NodeItem>;
     selectedItems: Set<string>;
     missingCustomMetrics?: AdditionalMetric[];
+    itemsAlerts?: {
```
rephus, line 34: nice, we could refactor the missing custom dimensions/metrics and other warnings to this generic alert

rephus (approved):
I like the new alert functionality

ZeRego:
--Update--
Remove suggestions for now. PR only shows error.

Issue #13615, opened 2025-02-11
Lightdash generate queries the wrong table information when using dbt cloud cli


almeidabbm:
### Description

Relates to: https://github.com/lightdash/lightdash-internal-work/issues/2840

When using dbt cloud cli we don't have access to credentials, therefore to get the table informations and `generate` the yaml files we perform a request to `api/v1/projects/<project_uuid>/sqlRunner/fields` which uses the current project warehouse connection - https://docs.lightdash.com/references/lightdash-cli/#known-limitations-when-using-dbt-cloud-cli

This means that if the current project doesn't have the expected model materialized, the `generate` command will fail with the following error:

![Image](https://github.com/user-attachments/assets/bdf2f066-7034-4c24-8a21-92d3fcd97724)

We have tried to go over this issue by adding a database override in this endpoint: https://github.com/lightdash/lightdash/pull/13546

For snowflake, this endpoint makes the following query:

```sql
SELECT TABLE_CATALOG as "table_catalog",
                   TABLE_SCHEMA  as "table_schema",
                   TABLE_NAME    as "table_name",
                   COLUMN_NAME   as "column_name",
                   DATA_TYPE            as "data_type"
FROM information_schema.columns
WHERE TABLE_NAME = ?
AND TABLE_SCHEMA = ?
AND TABLE_CATALOG = ?
ORDER BY 1, 2, 3;
```

And since it relies on where clauses instead of a fully qualified name - which solves the quoted identifiers issue - we can't be sure that the `schema`, `database` or `table` will match the casing in the database.

### Steps to Reproduce the Bug or Issue

1. Create a snowflake db without events table (prod)
2. Deploy using prod db to lightdash using the cli
3. Create a snowflake db with the events model (staging)
4. Using the dbt cloud project that has the staging db, run `lightdash generate -s events`

### Possible solution
- Use `show columns in {table}` with fully qualified name instead of  where clauses

### Cloud or self-hosting

None

PR #13617, opened 2025-02-11
fix: use getCatalog to get columns information when all arguments of get fields are available

almeidabbm:
Closes: #13615 

### Description:

- When `database`, `schema` and `table` exist, we can use `getCatalog` to get the table fields using a fully qualified name that will not fail because of casing - query with where clauses fails because of that.

### Reviewer actions


almeidabbm:
Note: this fixes the dbt cloud cli issue in the ticket, but still it uses the current project connection to make the query, wondering if we should try and find another way to make dbt cloud cli work more seamlessly, not sure if possible because we cannot get the credentials

ZeRego (approved):

Issue #13517, opened 2025-02-04
Enable overriding SDK component styles


magnew:
Enable overriding SDK component styles

IrakliJani:
see suggestion: https://github.com/lightdash/lightdash/pull/13455#discussion_r1949319970

PR #13627, opened 2025-02-11
feat: enable overriding fonts and bg in sdk

magnew:
Closes: #13517 
### Description:

Enable override of fonts and background in sdk

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Table/Table.styles.ts
+++ b/packages/frontend/src/components/common/Table/Table.styles.ts
@@ -25,14 +25,14 @@ interface TableContainerProps {
 }
 
 export const TableContainer = styled.div<
-    TableContainerProps & { children: ReactNode }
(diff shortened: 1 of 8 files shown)


packages/frontend/src/components/common/Table/Table.styles.ts
```diff
@@ -25,14 +25,14 @@ interface TableContainerProps {
 }
 
 export const TableContainer = styled.div<
-    TableContainerProps & { children: ReactNode }
+    TableContainerProps & { children: ReactNode; $tableFont: string }
```
IrakliJani, line 28: this should be optional right? can be in the TableContainerProps

IrakliJani (approved):

packages/frontend/src/components/common/Table/Table.styles.ts
magnew: Good call

PR #13628, opened 2025-02-11
fix: add slash to lightdash URL when not present

magnew:
Closes: 

### Description:

Adds a slash to lightdash url when not present. Both of our initial uses had this problem.

### Reviewer actions

### DIFF
--- a/packages/sdk/src/Lightdash.tsx
+++ b/packages/sdk/src/Lightdash.tsx
@@ -39,6 +39,10 @@ const decodeJWT = (token: string) => {
 };
 
 const persistInstanceUrl = (instanceUrl: string) => {
+    if (!instanceUrl.endsWith('/')) {
+        instanceUrl = `${instanceUrl}/`;
+    }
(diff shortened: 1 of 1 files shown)


packages/sdk/src/Lightdash.tsx
```diff
@@ -39,6 +39,10 @@ const decodeJWT = (token: string) => {
 };
 
 const persistInstanceUrl = (instanceUrl: string) => {
+    if (!instanceUrl.endsWith('/')) {
+        instanceUrl = `${instanceUrl}/`;
```
IrakliJani, line 43: strange that typescript is not complaining on mutating this

IrakliJani (approved):

packages/sdk/src/Lightdash.tsx
magnew: 😑 Agree. I thought about changing it anyway, but it works.

Issue #13635, opened 2025-02-12
Apply user attributes to table FROM clause


owlas:
Allows the table location to be templated with user attributes

e.g.

```
select * FROM `database`.`${attribute.schema}`.name
```

PR #13663, opened 2025-02-13
feat: apply user attributes to sql from

owlas:
Closes: #13635

### DIFF
--- a/packages/common/src/compiler/translator.ts
+++ b/packages/common/src/compiler/translator.ts
@@ -552,12 +552,13 @@ export const convertTable = (
         });
     }
 
+    const sqlTable = model.meta.sql_from || model.relation_name;
     return {
         name: model.name,
         label: tableLabel,
         database: model.database,
         schema: model.schema,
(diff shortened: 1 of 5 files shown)


almeidabbm (changes requested):
Interpolation of attributes works well:

image

But I believe we would need to wrap the `sql_from` part in parentheses for it to work, getting this error:

image

In sql runner, I had to wrap it:

image

packages/common/src/compiler/translator.ts
```diff
@@ -552,12 +552,13 @@ export const convertTable = (
         });
     }
 
+    const sqlTable = model.meta.sql_from || model.relation_name;
```
almeidabbm, line 555: possibly here
```suggestion
    const sqlTable = model.meta.sql_from ? `(${model.meta.sql_from})` : model.relation_name;
```

almeidabbm (approved):

almeidabbm:
👨‍🍳 🤌

Issue #13638, opened 2025-02-12
disable pull request to write back if you use filter logic that we don't support in YAML


TuringLovesDeathMetal:
### Description

Right now there are more options for filtering metrics from the custom metric builder in the UI that aren't supported in our YAML metric definitions. This means that teh write back won't actually work. 

We should disable `open pull request` with a message saying `This metric has filters which are not yet supported in YAML.`

PR #13665, opened 2025-02-13
fix: disable write back on error

rephus:
Closes: https://github.com/lightdash/lightdash/issues/13638

### Description:

Screenshot from 2025-02-13 11-19-20
Screenshot from 2025-02-13 11-19-29

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/CustomMetricWriteBackModal/index.tsx
+++ b/packages/frontend/src/components/Explorer/CustomMetricWriteBackModal/index.tsx
@@ -1,5 +1,7 @@
 import {
     convertCustomMetricToDbt,
+    getErrorMessage,
+    NotImplementedError,
     type AdditionalMetric,
 } from '@lightdash/common';
 import {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/CustomMetricWriteBackModal/index.tsx
```diff
@@ -92,9 +94,25 @@ const SingleCustomMetricModalContent = ({
...
     const previewCode = useMemo(() => {
-        return yaml.dump({ [item.name]: convertCustomMetricToDbt(item) });
+        try {
+            const code = yaml.dump({
+                [item.name]: convertCustomMetricToDbt(item),
+            });
+            setError(undefined);
+            return code;
+        } catch (e) {
+            if (e instanceof NotImplementedError) {
```
notgiorgi, line 107: [non-blocker] could it also throw this error? 

https://github.com/lightdash/lightdash/blob/a821783c614cf5f3649705f5ffad93ba454ed467/packages/common/src/types/filterGrammarConversion.ts#L30

notgiorgi (approved):

packages/frontend/src/components/Explorer/CustomMetricWriteBackModal/index.tsx
rephus: yes !, it could, as well as other LightdashError , 
If that happens, we will still show the error message on `setError(getErrorMessage(e));`

`NotImplementedError` is just the most "expected" error and what we're trying to solve here, therefore, the error message is more detailed.

That being said, I'm going to make a small change to also show the `error type`if not NotImpementedError

rephus: Updated: 

image

Issue #13458, opened 2025-01-30
Dashboard exports are cropped/don't respect dashboard layout (only markdown tiles)


joaoviana:
### Description

internal dashboard: https://analytics.lightdash.cloud/projects/21eef0b9-5bae-40f3-851e-9554588e71a6/dashboards/039f8b7e-afc0-4842-9448-4fc0fdf76de8/view

<img width="1887" alt="Image" src="https://github.com/user-attachments/assets/4d120b76-bd00-463f-a8a4-4f0e83265ce9" />

export doesn't move markdown tile to bottom or find another way to keep both in view **(only when there are only markdown tiles)**

<img width="691" alt="Image" src="https://github.com/user-attachments/assets/5dd3af17-4ced-474e-a115-c82991c0277d" />

when there is one saved chart tile added, it will export correctly, not cropping the layout

see `saveScreenshot` in `unfurlService` and see how markdown tiles are being ignore when calculating the layout / page width before screenshotting

### Steps to Reproduce the Bug or Issue

see description ^

### Cloud or self-hosting

cloud

PR #13676, opened 2025-02-13
fix: wait for markdown tiles before screenshot

joaoviana:
Closes: [#13458](https://github.com/lightdash/lightdash/issues/13458)

### Description:

wait for markdown tiles to be attached to dom before screenshotting.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -671,6 +671,15 @@ export class UnfurlService extends BaseService {
                     }
 
                     if (lightdashPage === LightdashPage.DASHBOARD) {
+                        // Wait for markdown tiles specifically
+                        const markdownTiles = await page
+                            .locator('.markdown-tile')
+                            .all();
+                        await Promise.all(
+                            markdownTiles.map((tile) =>
+                                tile.waitFor({ state: 'attached' }),
+                            ),
+                        );
                         const loadingChartOverlays = await page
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -671,6 +671,15 @@ export class UnfurlService extends BaseService {
                     }
 
                     if (lightdashPage === LightdashPage.DASHBOARD) {
+                        // Wait for markdown tiles specifically
+                        const markdownTiles = await page
+                            .locator('.markdown-tile')
+                            .all();
+                        await Promise.all(
+                            markdownTiles.map((tile) =>
+                                tile.waitFor({ state: 'attached' }),
```
joaoviana, line 680: "attached" is better than "visible" - tile could be empty, so if there's no content it wouldn't be considered as "visible"

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -708,12 +717,13 @@ export class UnfurlService extends BaseService {
                         finalSelector = '.react-grid-layout';
                     }
 
-                    const fullPage = await page.$(finalSelector);
+                    const fullPage = await page.locator(finalSelector);
```
joaoviana, line 720: locator is the way to go, that way we accurately get the fullpage boundingbox

almeidabbm (approved):
👨‍🍳 🤌 
image

PR #13695, opened 2025-02-14
feat: remove legacy sql runner page

ZeRego:
### Description:

Removes legacy SQL runner page. API endpoint is being deprecated instead of removed to give more time for developers to stop relying on it.

We had a banner for more than 3 months to warn users to stop using it.
Screenshot 2025-02-14 at 14 17 11

### Reviewer actions


joaoviana (changes requested):
can you remove sqlRunner.cy.ts in e2e/app?
and rename sqlRunnerNew.cy.ts to sqlRunner.cy.ts?

joaoviana (commented):
something that can be done later: 
Rename SqlRunnerNew.tsx to just SqlRunner.tsx

joaoviana (commented):
I think there's also a prop in `VisualizationProvider` we don't need anymore: 'isSqlRunner' - I think we could remove it here

joaoviana (approved):

Issue #12577, opened 2024-11-22
Exporting image don't work for pie charts


rephus:
### Description

![image](https://github.com/user-attachments/assets/d5018d38-3af5-46a4-83e0-e01d1e6a9745)

### Steps to Reproduce the Bug or Issue

https://demo.lightdash.com/projects/2014e038-ff4b-4761-ae6f-fbf551e7b468/dashboards/665be298-b79e-456f-9ef3-03f9e068a2d0/view

### Cloud or self-hosting

None

rephus:
There is something weird on these charts, If I try to export the image, it throws an error, but if I change the label ,then it works fine. It must be an invalid state. 

![image](https://github.com/user-attachments/assets/0b7f8b8b-6f81-4eb1-8933-8f6f0b6a7b94)

rephus:
Duplicated issue https://github.com/lightdash/lightdash/issues/7414

rephus:
Duplicated issue https://github.com/lightdash/lightdash/issues/7414

ptdatta:
Hey @rephus is this issue fixed? As in my case pie charts are downloading correctly.

HMaan0:
Hey @ptdatta, It looks like this issue still exists. I would love to work on fixing it if it’s still open. Let me know if there’s anything I should consider before getting started.

PR #13707, opened 2025-02-15
fix: exporting image for pie chart

HMaan0:
Closes: #12577

### Description:
This PR introduces changes to how the chart instance is handled for pie charts. Instead of directly extracting the URL (svgBase64) from  chart instance, the URL is now extracted from the DOM.
Works for both downloading (all format supported) and copying to clipboard.

https://github.com/user-attachments/assets/8126cb4b-3060-425e-978c-3809215535cd

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/ChartDownload/chartConfigs.ts
+++ b/packages/frontend/src/components/common/ChartDownload/chartConfigs.ts
@@ -1,0 +1,53 @@
+export const chartConfigs = (chartInstance: any) => {
+    const isPieChat = chartInstance._chartsViews.some(
+        (view: { constructor: { name: string } }) =>
+            view.constructor.name === 'PieView2',
+    );
+    if (!isPieChat) {
+        const svgBase64 = chartInstance.getDataURL();
+        const width = chartInstance.getWidth();
+        const height = chartInstance.getHeight();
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/ChartDownload/chartConfigs.ts
```diff
@@ -0,0 +1,54 @@
+import { type EChartsInstance } from 'echarts-for-react';
+export const chartConfigs = (chartInstance: EChartsInstance) => {
+    const isPieChat = chartInstance._chartsViews.some(
```
joaoviana, line 3: there's a typo here ^

packages/frontend/src/components/common/ChartDownload/chartConfigs.ts
```diff
@@ -0,0 +1,54 @@
...
+        (view: { constructor: { name: string } }) =>
+            view.constructor.name === 'PieView2',
```
joaoviana, line 5: not sure if there's a better way to do this - perhaps from a prop in ChartDownloadOptions

joaoviana:
Thank you for your contribution @HMaan0 but since I have concerns over the implementation I'm going to close this.

Issue #13689, opened 2025-02-14
Headless Browser authentication fails when internal host is http and SECURE_COOKIES is `true`


joaoviana:
### Description

The headless browser authentication fails when `headlessBrowser.internalLightdashHost` and `siteUrl` use different domains. This occurs because the `connect.sid` cookie domain isn't being set on the internal domain.

Affected Components

1. `UnfurlService.getUserCookie()` - Cookie domain set from internal host
2. `App.initExpress()` - CSP settings might block cross-domain cookies
3. Headless browser login flow

Reason:
1. In UnfurlService.ts, the cookie is set for parsedUrl.hostname from the internal Lightdash host
2. When internal/public domains differ (e.g., internal: http://localhost:8080, public: https://lightdash.example.com), the cookie domain becomes invalid for the actual site URL
3. The current implementation in parseUrl() uses internalLightdashHost for minimal URLs but doesn't account for public domain cookie requirements

> [!IMPORTANT]
> This works well if `SECURE_COOKIES` is set to **false**, but that's not an appropriate workaround. 

### Steps to Reproduce the Bug or Issue

See above  ^ 

### Cloud or self-hosting

self-hosting

PR #13723, opened 2025-02-17
fix: set signed cookie on login of headless browser

joaoviana:
Closes: #13689 

### Description:

How to test this
- INTERNAL_LIGHTDASH_HOST=http://lightdash-dev:3000
- SITE_URL=http://localhost:3000 or create a https tunnel with `cloudflared` (this ensures you can login when cookies are set to be **secure**!)
- SECURE_COOKIES=true
- RUN "dashboard image export" - 'Generate preview' is a way to test - anything that calls the screenshot functionality and triggers a headlessBrowser login.
 
Main changes:
1. Implemented secure cookie setting logic for cross-domain scenarios
2. Allows separate internal routing for headless browser requests
3. Added `cookie-signature` dependency for secure cookie handling
4. Simplified error message when headlessBrowser login fails in `UnfurlService` 

Tests:
1. New test for invalid token handling (Render envs will have INTERNAL_LIGHTDASH_HOST always set, based on the service name)
2. Added dashboard image export preview test

> [!IMPORTANT]
> Setting a signed session cookie manually only happens if :  secure cookies are enabled, the internal host is http (can't set session cookie because of policies set in App.ts). This is because 'express-session' middleware does not allow us to set multiple domain's cookie sessions

> [!IMPORTANT]
> headlessBrowser's login endpoint requires a token that is compared with the private method that calls it when unfurling. 

Reasoning:

### Reviewer actions

test-cli
test-backend
test-frontend


packages/backend/src/routers/headlessBrowser.ts
```diff
@@ -16,6 +17,13 @@ export const getAuthenticationToken = (value: string) =>
         .update(value)
         .digest('hex');
 
+// TEST:
+// try with Docker compose; then run api login the headless browser login with Postman - different origin, should throw error
+// set internal host to point to the name of the container in docker compose
+
+// have e2e tests for this to catch possible regressions
+// might have to have render use the internal host env var
```
almeidabbm, line 25: Should this be removed?

almeidabbm (approved):

packages/backend/src/routers/headlessBrowser.ts
joaoviana: 🤦 yes! Will push, wait for build and merged. tyyyy!

Issue #13730, opened 2025-02-18
Support between and not between filter for numeric fields


ZeRego:
### Description

We have a `between` filter for dates but not for numeric fields.

### What problem does this solve?

At the moment users need multiple filters to achieve this logic.

PR #13757, opened 2025-02-19
feat: Support between and not between filter for numeric fields

notgiorgi:
Closes: #13730

### Description:

Support between and not between filter for numeric fields

Demo: https://cln.sh/JqVrqcHb

### Reviewer actions


packages/frontend/src/components/DashboardFilter/FilterConfiguration/utils/index.ts
```diff
@@ -35,10 +35,11 @@ export const hasFilterValueSet = (filterRule: DashboardFilterRule) => {
         case FilterOperator.NOT_IN_THE_CURRENT:
             return filterRule.settings && filterRule.settings.unitOfTime;
         case FilterOperator.IN_BETWEEN:
+        case FilterOperator.NOT_IN_BETWEEN:
             return (
                 filterRule.values &&
                 filterRule.values.length === 2 &&
-                filterRule.values.every(Boolean)
+                filterRule.values.every((val) => val != null && val !== '')
```
notgiorgi, line 42: For just dates `Boolean` would've worked, but we need to properly check for `0` now for numers

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -144,6 +144,14 @@ export const renderNumberFilterSql = (
...
         case FilterOperator.LESS_THAN_OR_EQUAL:
             return `(${dimensionSql}) <= (${filter.values?.[0] || 0})`;
+        case FilterOperator.IN_BETWEEN:
+            return `(${dimensionSql}) >= (${
+                filter.values?.[0] || 0
+            }) AND (${dimensionSql}) <= (${filter.values?.[1] || 0})`;
+        case FilterOperator.NOT_IN_BETWEEN:
+            return `(${dimensionSql}) <= (${
+                filter.values?.[0] || 0
+            }) OR (${dimensionSql}) >= (${filter.values?.[1] || 0})`;
```
ZeRego, line 154: @TuringLovesDeathMetal Can you confirm this logic pls
We have implemented the between to be inclusive (same as the date between). Based on that, the "not between" should be `< min OR > max`, so the numbers mentioned are not part of the results. 
I belive `<= min OR >= max` is a bug. 

Example between:
Screenshot 2025-02-24 at 14 22 08

TuringLovesDeathMetal: agreed! I think it should be "exclusive" of the numbers listed (so what you said: < min or > max)

notgiorgi: Makes sense, making the change

ZeRego (approved):

Issue #8761, opened 2024-01-29
Be able to set the conditional formatting on a field based on a value in a different field.


TuringLovesDeathMetal:
### Description

I want to be able to set the conditional formatting on a field based on a value in a different field.

So In a table under the conditional formatting configure, it would be really nice to have the 'select field' be the field that the condition is based on and the 'apply' field be the field that the color will apply to.

![image](https://github.com/lightdash/lightdash/assets/31848148/5087e1b2-e639-491d-93fc-ce4b22bdcb97)

^This example would have the colouring applied to the `Baskets Average Baskets Total` field, but the logical operator would be based on the value of another field

ZeRego:
@almeidabbm @TuringLovesDeathMetal 
The fields should be selected for each condition! So we can create the following logic:
> If status = completed and amount > 100, display green color

For backwards compatibility, if there is no field value it defaults to the field that is being coloured.

almeidabbm:
> [@almeidabbm](https://github.com/almeidabbm) [@TuringLovesDeathMetal](https://github.com/TuringLovesDeathMetal) The fields should be selected for each condition! So we can create the following logic:
> 
> > If status = completed and amount > 100, display green color
> 
> For backwards compatibility, if there is no field value it defaults to the field that is being coloured.
> 
> This feature is not supported for range colouring.

@ZeRego I'm not sure I understood what you meant by "if there is no field value it defaults to the fields that is being coloured"

If there is no field to compare to, then the comparison will return `false` and therefore the coloring will be "white" by default

Is this what you meant?

ZeRego:
@almeidabbm I haven't looked into this code but I imagine the data looks something like 
```
conditional_formatting: [
   {
        field: 'metric_a' // metric I want to color
        type: 'single_color'
        color: 'blue'
        conditions: [
             {
                  operator: 'equals',
                  value: '100'
             }
       ]
   }
]
```

And we need to add an optional property to the condition object. Example:
```
 {
    field: 'another_metric' // metric I want to use for the condition
    operator: 'equals',
    value: '100'
}
```

For backwards compatibility if `field` is undefined / null, it should use 'metric_a'.

PR #13766, opened 2025-02-20
feat: conditional formatting against another field and with another field's values

almeidabbm:
Closes: [#8761](https://github.com/lightdash/lightdash/issues/8761)

### Description:

**Allows to compare field to another field - only when other field is same type**
image

**Allows to color field based on another fields values - doesn't have to be same type**
image

V2:

Added segmented control:
image

### Reviewer actions

### DIFF
--- a/packages/common/src/types/conditionalFormatting.ts
+++ b/packages/common/src/types/conditionalFormatting.ts
@@ -99,7 +99,8 @@ export type ConditionalFormattingRowFields = Record<
     }
 >;
 
-export enum CompareTargetComparisonType {
-    Field = 'field',
-    Values = 'values',
+export enum ConditionalFormattingComparisonType {
+    VALUES = 'values',
+    TARGET_FIELD = 'target_field',
+    TARGET_TO_VALUES = 'target_to_values',
 }
--- a/packages/common/src/types/conditionalFormatting.ts
+++ b/packages/common/src/types/conditionalFormatting.ts
@@ -37,8 +37,7 @@ export const isConditionalFormattingWithValues = (
 
 export const isConditionalFormattingWithCompareTarget = (
     rule: ConditionalFormattingWithConditionalOperator,
-): rule is ConditionalFormattingWithCompareTarget =>
-    'compareTarget' in rule && Boolean(rule.compareTarget);
+): rule is ConditionalFormattingWithCompareTarget => 'compareTarget' in rule;
 
 export type ConditionalFormattingConfigWithSingleColor = {
     target: FieldTarget | null;
--- a/packages/common/src/types/conditionalFormatting.ts
+++ b/packages/common/src/types/conditionalFormatting.ts
@@ -24,6 +24,7 @@ export type ConditionalFormattingWithValues<T = number | string> =
 export type ConditionalFormattingWithCompareTarget<T = number | string> =
     ConditionalRule<ConditionalOperator, T> & {
         compareTarget: FieldTarget | null;
+        values?: T[];
     };
 
 export type ConditionalFormattingWithConditionalOperator<T = number | string> =
@@ -97,3 +98,8 @@ export type ConditionalFormattingRowFields = Record<
         value: unknown;
     }
 >;
+
+export enum CompareTargetComparisonType {
+    Field = 'field',
+    Values = 'values',
+}
--- a/packages/common/src/types/conditionalFormatting.ts
+++ b/packages/common/src/types/conditionalFormatting.ts
@@ -15,10 +15,27 @@ export type ConditionalFormattingColorRange = {
     steps: number;
 };
 
-export type ConditionalFormattingWithConditionalOperator<T = number | string> =
+export type ConditionalFormattingWithValues<T = number | string> =
     ConditionalRule<ConditionalOperator, T> & {
         values: T[];
     };
+
+export type ConditionalFormattingWithCompareTarget<T = number | string> =
+    ConditionalRule<ConditionalOperator, T> & {
+        compareTarget: FieldTarget | null;
+    };
+
+export type ConditionalFormattingWithConditionalOperator<T = number | string> =
+    | ConditionalFormattingWithValues<T>
+    | ConditionalFormattingWithCompareTarget<T>;
+
+export const isConditionalFormattingWithValues = (
+    rule: ConditionalFormattingWithConditionalOperator,
+): rule is ConditionalFormattingWithValues => 'values' in rule;
+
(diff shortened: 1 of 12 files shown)


packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormattingRule.tsx
```diff
@@ -106,36 +163,64 @@ const ConditionalFormattingRule: FC<ConditionalFormattingRuleProps> = ({
...
-                            if (!value) return;
-                            onChangeRuleOperator(value as ConditionalOperator);
+                <Stack spacing="xs">
+                    <Switch
+                        label="Compare to another field"
+                        disabled={!field}
+                        labelPosition="right"
+                        checked={isConditionalFormattingWithCompareTarget(rule)}
+                        onChange={(event) => {
+                            console.log(event.currentTarget.checked, rule);
```
joaoviana, line 173: remove

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormattingItem.tsx
```diff
@@ -113,6 +116,30 @@ export const ConditionalFormattingItem: FC<Props> = ({
                     } else if (newField) {
                         // Update the target if the field is changed
                         draft.target = { fieldId: getItemId(newField) };
+
+                        if (
+                            isConditionalFormattingConfigWithSingleColor(draft)
```
joaoviana, line 121: can you add comments here?

packages/common/src/types/conditionalFormatting.ts
```diff
@@ -15,11 +16,30 @@ export type ConditionalFormattingColorRange = {
...
+    | ConditionalFormattingWithCompareTarget<T>;
+
+export const isConditionalFormattingWithValues = (
+    rule: ConditionalFormattingWithConditionalOperator,
+): rule is ConditionalFormattingWithValues => 'values' in rule;
+
+export const isConditionalFormattingWithCompareTarget = (
+    rule: ConditionalFormattingWithConditionalOperator,
+): rule is ConditionalFormattingWithCompareTarget =>
+    'compareTarget' in rule && Boolean(rule.compareTarget);
```
almeidabbm, line 41: This is the fallback that mentioned here https://github.com/lightdash/lightdash/issues/8761#issuecomment-2672050448 by @ZeRego

almeidabbm: Had to remove this since when we're selecting a compare target, the initial value is null, and will have no values property on the condition (type will be compare conditional format target to compare target) so it would throw `notImplemented` because it was none of the 3 types of conditions:
1. Compare field to values (only has values array in the condition)
2. Compare field to compare target value (has compareTarget property in the condition, but no values property)
3. Compare compare target to values (has compare target property and values property)

I don't believe this fallback will be needed since I don't think there will be any regression, previous conditional formattings only had `values` property (no `compareTarget`) which is still a valid condition and will compare it to the field. To compare to a different field, the user has to explicitly enable the switch.

This would be a problem with regression if we ALWAYS added the `compareTarget` property and decided on wether comparing it to the target or to the field depending on if it was `null` or not.

packages/frontend/src/components/common/Filters/FilterInputs/DefaultFilterInputs.tsx
```diff
@@ -36,6 +37,10 @@ const DefaultFilterInputs = <T extends ConditionalRule>({
             : undefined,
     });
 
+    if (!field) {
```
joaoviana, line 40: we should test this in other places - let's see how this goes!

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormattingRule.tsx
```diff
@@ -106,36 +255,99 @@ const ConditionalFormattingRule: FC<ConditionalFormattingRuleProps> = ({
...
+                                disabled={!field}
+                                labelPosition="right"
+                                checked={comparisonSwitchValues.includes(
+                                    CompareTargetComparisonType.Field,
+                                )}
+                                onChange={handleChangeComparisonType(
+                                    CompareTargetComparisonType.Field,
+                                )}
+                            />
+                            <Checkbox
```
joaoviana, line 272: hide if the switch above is false - I think this is how it works!

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormattingRule.tsx
```diff
@@ -106,36 +255,99 @@ const ConditionalFormattingRule: FC<ConditionalFormattingRuleProps> = ({
...
+                                )
+                            ) : (
+                                <FiltersProvider projectUuid={projectUuid}>
+                                    <FilterInputComponent
+                                        filterType={filterType}
+                                        field={field}
+                                        rule={rule}
+                                        onChange={handleChangeRule}
+                                    />
+                                </FiltersProvider>
```
joaoviana, line 347: potential refactor here I think!

joaoviana (commented):
tested on table and on pivot table

Screenshot 2025-02-21 at 11 12 43
Screenshot 2025-02-21 at 11 12 11
Screenshot 2025-02-21 at 11 12 04
Screenshot 2025-02-21 at 11 06 03
Screenshot 2025-02-21 at 11 03 17

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormattingRule.tsx
almeidabbm: Done in: [chore: remove optional field in filter inputs](https://github.com/lightdash/lightdash/pull/13766/commits/213cc6a5ae3689bf2caadaab855d6c12746f93bb)

packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/ConditionalFormattingRule.tsx
almeidabbm: Done in: [chore: hide checkbox if switch is false](https://github.com/lightdash/lightdash/pull/13766/commits/2e4321f437ba6228bfdb7c70341200880421eb72)

joaoviana (approved):

Issue #13764, opened 2025-02-20
Reimplement formatting in the frontend to avoid crashes when loading a lot of results from a warehouse.


rephus:
### Description

Just skip the entire formatted method on results, then return results only with raw without formatted, and do the formatting entirely in the frontend.

Thread: https://lightdash.slack.com/archives/C02DGV9RWKV/p1740059311960229

Related PR: https://github.com/lightdash/lightdash/pull/13006

Later on, we can add streaming to the request to make it more efficient  and reduce memory consumption.

PR #13768, opened 2025-02-20
feat: format rows frontend

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13764

### Description:

Before

Screenshot from 2025-02-20 16-11-41

After

Screenshot from 2025-02-20 16-27-41
Screenshot from 2025-02-20 16-56-46

Screenshot from 2025-02-20 17-05-20

### Reviewer actions

test-frontend test-backend

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -1178,6 +1178,24 @@ export function formatRows(
     });
 }
 
+export function rowsWithoutFormatting(rows: { [col: string]: AnyType }[]) {
+    return rows.map((row) => {
+        const resultRow: ResultRow = {};
+        const columnNames = Object.keys(row || {});
+
+        for (const columnName of columnNames) {
+            const value = row[columnName];
+            resultRow[columnName] = {
+                value: {
+                    raw: value,
+                    formatted: null,
+                },
+            };
+        }
+        return resultRow;
+    });
+}
+
 const isObject = (object: AnyType) =>
     object != null && typeof object === 'object';
 export const removeEmptyProperties = (object: Record<string, AnyType>) => {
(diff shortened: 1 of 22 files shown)


packages/common/src/index.ts
```diff
@@ -1178,6 +1178,24 @@ export function formatRows(
...
+                value: {
+                    raw: value,
+                    formatted: null,
+                },
+            };
+        }
+        return resultRow;
+    });
+}
+
```
almeidabbm, line 1198: This is still duplicating the rows and creating new objects for each row column, we might be bypassing formatting here, but the memory overhead should still be high.

Can't we just return the raw rows? instead of formatting them to be `ResultRow[]`?

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -64,6 +73,11 @@ export const formatCellContent = (data?: { value: ResultValue }) => {
             : value?.formatted ?? value?.raw;
     }
 
+    if (value?.formatted === null && item) {
+        // Null formatting means the formatting was skipped by the backend
+        // so we need to handle formatting in the frontend based on the raw value
+        return formatItemValue(item, value?.raw);
+    }
```
almeidabbm, line 80: continuing on the comment above, we could create a type check function `isRawValue` that just checks `!('raw' in value && 'formatted' in value)` and use it here to make sure we use the value as is

packages/common/src/index.ts
rephus: yeah, I agree, however the memory impact should not be that big (it should be at least half) , and in terms of speed, it is a lot faster. 

The issue with  just returning raw rows, is that we need to change the respose TYpe of our result request, which is cascading through all the components. Even worse, if we want to make it conditional, we should return multiple types for this response, which is really annoying. 

I agree in the future we sshould skip this loop completely, but for this POC (while we keep this behind an env variable), I believe this is the most practical solution.

almeidabbm (approved):

rephus:
errors in e2e don't seem relevant

PR #13873, opened 2025-02-28
fix: check if custom metric filter exists before validating

joaoviana:
Closes: #13872

### Description:

custom metric filter could be `null`.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.ts
@@ -364,23 +364,21 @@ export class ValidationService extends BaseService {
                         [],
                     );
 
-                    const customMetricFilterErrors =
-                        customMetricsFilters.reduce<CreateChartValidation[]>(
-                            (acc, filter) => {
-                                const fieldId = convertFieldRefToFieldId(
-                                    filter.target.fieldRef,
-                                );
-                                return containsFieldId({
-                                    acc,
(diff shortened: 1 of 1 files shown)


packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -364,23 +364,21 @@ export class ValidationService extends BaseService {
...
-                                    fieldId,
-                                    error: `Custom metric filter error: the field '${fieldId}' no longer exists`,
-                                    errorType: ValidationErrorType.CustomMetric,
-                                    fieldName: fieldId,
-                                });
-                            },
-                            [],
-                        );
+                    const customMetricFilterErrors = customMetricsFilters
+                        .filter((f) => !!f?.target)
```
joaoviana, line 368: happy to do something different or extending the MetricFilterRule[] to be (MetricFilterRule|null)[]

notgiorgi, line 368: curios in which cases does `filter.target` doesn't exist? shouldn't that fail the validation?

joaoviana: oops, meant **just** `filter`

joaoviana: So perhaps the query returns nulls?

notgiorgi (approved):

Issue #5422, opened 2023-05-10
Keep original quotations and comments in generated property YAML files


yu-iskw:
### Description
`lightdash generate` is awesome to automatically add metadata for lightdash. But, it can be break YAML files, if they contains prohibit characters without quotations as `[` and `]`. Aside from that, we would like to keep comments in original YAML files in the generated YAML files.

#### Example YAML
If the original looks like below,
```yaml
  columns:
    # The column will be deprecated on xxx.
    - name: column1
      description: "Acceptable only [1, 2, 3]"
```

the generated YAML is invalid because `[` and `]` is 
```yaml
  columns:
    - name: column1
      description: Acceptable only [1, 2, 3]
```

### What problem does this solve?

In python, there is a package which enables us to keep comments and quotations in an original YAML. However, I don't know if there is any similar module which is available in node for the lightdash CLI.
https://pypi.org/project/ruamel.yaml/

owlas:
Hey @yu-iskw thanks for opening this! we've been trying to find a similar yaml (de)serializer that preserves comments/whitespace/quotes like you've requested but it seems we don't have as good an option as python (which is frustrating).

It's been a few months since we've searched for a solution, we'll take a look again

yu-iskw:
@owlas I totally understand. I will look for a solution even to appropriately quote strings for descriptions and so on.

yu-iskw:
@owlas It seems that https://github.com/eemeli/yaml enables us to deal with comments in YAML.

https://eemeli.org/yaml/#comments-and-blank-lines

I found the package in the issue.
https://github.com/nodeca/js-yaml/issues/689

Issue #13896, opened 2025-03-03
Refactor YamlSchema types conversion to DbtSchemaEditor


ZeRego:
> > That way this class only cares about YamlSchema types
>
> If the conversions are not part of the class, we are going to spread the YamlSchema types across other util functions or classes. 
> Instead, I'm going to move the util function into the class.
> Reasons:
> - We need the table and dimension where the metric is inserted on
> - Centralize dbt schema YML logic in this class

_Originally posted by @ZeRego in https://github.com/lightdash/lightdash/pull/13894#discussion_r1977530293_

PR #13901, opened 2025-03-03
feat: preserve YAML formatting and comments when running `lightdash generate` command

ZeRego:
Closes: #5422 
Closes: [#13896](https://github.com/lightdash/lightdash/issues/13896)

### Description:

Changes:
- move schema validation to class constructor (+ test)
- refactor `lightdash generate` command to use the new class. Preserves YAML formatting and comments.

Update meta:

https://github.com/user-attachments/assets/fb53d90d-092d-4b25-ae87-0b296df6b4d5

Generate columns:

https://github.com/user-attachments/assets/635209e9-5e7d-430b-a490-3550dd8d26ac

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
+++ b/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
@@ -2,7 +2,6 @@
 import { subject } from '@casl/ability';
 import {
     AdditionalMetric,
-    AnyType,
     ApiGithubDbtWritePreview,
     CustomSqlDimension,
     DbtProjectType,
@@ -12,7 +11,6 @@ import {
     getErrorMessage,
     GitIntegrationConfiguration,
     isUserWithOrg,
-    lightdashDbtYamlSchema,
     ParameterError,
     ParseError,
     PullRequestCreated,
@@ -22,11 +20,8 @@ import {
     snakeCaseName,
     UnexpectedServerError,
     VizColumn,
-    YamlSchema,
 } from '@lightdash/common';
-import Ajv from 'ajv';
 import { nanoid } from 'nanoid';
-import { parse } from 'yaml';
 import {
     LightdashAnalytics,
     WriteBackEvent,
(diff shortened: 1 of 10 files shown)


packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -121,24 +121,6 @@ export class GitIntegrationService extends BaseService {
         };
     }
 
-    protected static async loadYamlSchema(
```
ZeRego, line 124: Moved to DbtSchemaEditor constructor

packages/cli/src/handlers/generate.ts
```diff
@@ -138,13 +136,6 @@ export const generateHandler = async (options: GenerateHandlerOptions) => {
                     }
                 }
 
-                const existingHeadComments = await getFileHeadComments(
```
ZeRego, line 141: The new YAML library doesn't affect comments so we can remove this workaround

packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.ts
```diff
@@ -57,13 +96,38 @@ export default class DbtSchemaEditor {
...
+            return undefined;
+        }
+        const columns = model.getIn(['columns']);
+        if (!isSeq(columns)) {
+            // node is not an array
+            return undefined;
+        }
+        return columns.items
+            .filter(isMap)
+            .map<YamlColumn>((column) => column.toJSON());
```
notgiorgi, line 127: [non-blocker] I think it would make sense for all public methods to return `.toJSON()` value with a proper type and only private ones to use types from `yaml`

notgiorgi (approved):
lgtm, turning out great!

packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.ts
ZeRego: That was my intention at the start, but I'm finding that it's not possible because the "proper" types are created from the dbt manifest, which contains much more information than what's in the schema YAML.

"Proper" types contain all the information and can be converted to YAML types.
YAML types lack information about SQL and can't be converted back to "proper" types.

ZeRego:
CLI tests failing for unrelated reason.
Test pass locally. 
Screenshot 2025-03-04 at 11 37 32

Issue #13882, opened 2025-03-03
Rows with new line delimiters in snowflake in a table visualization are flickering


TuringLovesDeathMetal:
### Description

When I have a cell in a table visualization with a new row delimeter (i.e. `\n`) the cells in the table are flickering. 

This is on Snowflake

### Steps to Reproduce the Bug or Issue

see slack thread here: https://lightdash.slack.com/archives/C065J7AG5GQ/p1740761414059659?thread_ts=1738255212.326699&cid=C065J7AG5GQ

### Cloud or self-hosting

None

rephus:
This is not an issue of snowflake, this is an issue on virtual tables. We dynamically display rows to reduce memory in the frontend. If one of thoes rows is bigger than the rest, it might get removed. We need to investigate how to fix this

rephus:
I can replicate on postgres with this customd imension 

```
case 
when MOD(${orders.order_id},10) = 0 THEN 'line















'
else 
'small line'
end

```

PR #13904, opened 2025-03-04
fix: fix dynamic height virtual table

rephus:
Closes:  #13882

### Description:

Before

[Screencast from 2025-03-04 09-11-11.webm](https://github.com/user-attachments/assets/dc866a41-2e4b-4fed-8085-9b2ece9aadd8)

After

[Screencast from 2025-03-04 09-34-29.webm](https://github.com/user-attachments/assets/23c1f20e-f860-4e3c-9da0-834b220aed22)

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
+++ b/packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
@@ -212,7 +212,15 @@ const TableRow: FC<TableRowProps> = ({
         </Tr>
     );
 };
-
+const getMaxRowHeight = (row: Row<ResultRow>) => {
+    return row.getVisibleCells().reduce((acc, cell) => {
+        const lines = (
+            cell.getValue() as ResultRow[0] | undefined
+        )?.value?.formatted?.split('\n');
+        if (!lines) return acc;
+        return Math.max(acc, lines?.length * ROW_HEIGHT_PX);
+    }, ROW_HEIGHT_PX);
+};
 const VirtualizedTableBody: FC<{
     tableContainerRef: React.RefObject<HTMLDivElement | null>;
 }> = ({ tableContainerRef }) => {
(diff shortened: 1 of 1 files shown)


notgiorgi (approved):
lgtm

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
```diff
@@ -212,7 +212,15 @@ const TableRow: FC<TableRowProps> = ({
         </Tr>
     );
 };
-
+const getMaxRowHeight = (row: Row<ResultRow>) => {
+    return row.getVisibleCells().reduce((acc, cell) => {
+        const lines = (
+            cell.getValue() as ResultRow[0] | undefined
+        )?.value?.formatted?.split('\n');
```
rephus, line 219: this only takes into account new lines (which wsa the issue the customer reported)
This will not take into account long "line-break" text (we'll have to calculate the width of the cell, I believe this is not in scope)

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
```diff
@@ -223,7 +231,14 @@ const VirtualizedTableBody: FC<{
     const rowVirtualizer = useVirtualizer({
         getScrollElement: () => tableContainerRef.current,
         count: rows.length,
-        estimateSize: () => ROW_HEIGHT_PX,
+        estimateSize: (index) => {
+            try {
+                return getMaxRowHeight(rows[index]);
+            } catch (e) {
+                console.error('Error getting row height', e);
```
rephus, line 238: I don't expect this method to fail, but just in case, I added a try-catch so we don't throw a runtime error if this fails

packages/frontend/src/components/common/Table/ScrollableTable/TableBody.tsx
rephus: added a comment in code

PR #13922, opened 2025-03-04
fix: newlines in cells issue with react removeChild

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/13919

### Description:

I could not reproduce this locally, but I assume this is due to rendering fragments and multiple dom nodes instead of one. I updated the code to render just one DOM entry, as the error could be coming from Virtualizer or React Table itself.

Additionally, I removed the styled component when the cell field contains URLs, as it introduced an additional wrapper. I also removed the entire `RichBodyCell` component since I found it redundant and incorporated its logic into `BodyCell` itself.

before:
CleanShot 2025-03-04 at 19 27 58@2x

after:
CleanShot 2025-03-04 at 19 32 25@2x

### Reviewer actions


almeidabbm (approved):
LGTM 🚀 
image

Also tested: https://github.com/lightdash/lightdash/pull/13904

image

almeidabbm:
With URLS also works:
image

Issue #13910, opened 2025-03-04
Implement batch results for snowflake


almeidabbm:
### Description

Note: this should use the query id that `snowflake` uses in their API

**Relevant links**
- https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-execute#execute-queries-asynchronously
- https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-consume#batch-processing-results

PR #13927, opened 2025-03-04
feat: pagination interface for warehouse and snowflake implementation

almeidabbm:
Closes: [#13910](https://github.com/lightdash/lightdash/issues/13910)

### Description:

- Initial work to support paginated results.
- Only implemented for snowflake so far, querying using a queryId isn't working for other warehouses since we need to save metadata which will be done on a separate PR.
    - For now, when performing a query on a warehouse that doesn't have pagination implemented, we're returning the full result set and no more pages.
- Formats the rows as they come in instead of formatting the whole result set at once

#### How it works:
-  first request using a `metricQuery` or `chartUuid` returns a `queryId` and the requested page, subsequent requests should use the `queryId` to fetch more pages.

### Reviewer actions


packages/backend/src/controllers/v2/ProjectController.ts
```diff
@@ -0,0 +1,137 @@
...
+import { getContextFromHeader } from '../../analytics/LightdashAnalytics';
+import { allowApiKeyAuthentication, isAuthenticated } from '../authentication';
+import { BaseController } from '../baseController';
+import type { ApiRunPaginatedQueryResponse } from '../runQueryController';
+
+@Route('/api/v2/projects')
+@Response<ApiErrorPayload>('default', 'Error')
+@Tags('v2', 'Projects')
+export class V2ProjectController extends BaseController {
+    @Hidden()
```
almeidabbm, line 33: Hides the endpoint while it is WIP

packages/backend/src/controllers/runQueryController.ts
```diff
@@ -36,6 +38,11 @@ export type ApiRunQueryResponse = {
     };
 };
 
+export type ApiRunPaginatedQueryResponse = {
```
ZeRego, line 41: todo: move to packages/backend/src/controllers/v2/ProjectController.ts

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1756,6 +1768,592 @@ export class ProjectService extends BaseService {
...
+        if (
+            user.ability.cannot(
+                'manage',
+                subject('Explore', { organizationUuid, projectUuid }),
+            )
+        ) {
+            throw new ForbiddenError();
+        }
+
+        // TODO paginate: we need metadata to be able enforce this permission
```
ZeRego, line 2086: We need to change the `manage` Explore check for this to work for viewers. And we only validate the query content on creation (aka runPaginatedMetricQuery )

We should validate:
- query project uuid match URL project uuid
- user that created query is the same as user authenticated 

These are only possible once we store the metadata. For now, we can check if the user can view project uuid.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1756,6 +1768,592 @@ export class ProjectService extends BaseService {
...
+        page,
+        pageSize,
+        context = QueryExecutionContext.DASHBOARD,
+    }: PaginateDashboardChartArgs): Promise<ApiPaginatedQueryResults> {
+        if (!isUserWithOrg(user)) {
+            throw new ForbiddenError('User is not part of an organization');
+        }
+
+        const savedChart = await this.savedChartModel.get(chartUuid);
+        const { organizationUuid, projectUuid: dashboardProjectUuid } =
```
ZeRego, line 2234: chartProjectUuid rather than dashboardProjectUuid ?

ZeRego: Or do we want to get both chart and dashboard and compare both project uuids with the one from the URL?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1756,6 +1768,592 @@ export class ProjectService extends BaseService {
...
+        };
+
+        const queryTags: RunQueryTags = {
+            organization_uuid: organizationUuid,
+            project_uuid: projectUuid,
+            user_uuid: user.userUuid,
+            chart_uuid: chartUuid,
+            dashboard_uuid: dashboardUuid,
+            explore_name: explore.name,
+            query_context: autoRefresh
```
ZeRego, line 2317: non blocker: We should remove this arg from the endpoint and let the FE apply the correct context

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1756,6 +1768,592 @@ export class ProjectService extends BaseService {
...
+                                projectUuid,
+                                user.userUuid,
+                            ),
+                            {
+                                snowflakeVirtualWarehouse: explore.warehouse,
+                                databricksCompute: explore.databricksCompute,
+                            },
+                        );
+
+                    span.setAttribute('lightdash.projectUuid', projectUuid);
```
ZeRego, line 1845: non blocker: add page, page size to span metadata

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2082,6 +2680,27 @@ export class ProjectService extends BaseService {
         );
     }
 
+    private validatePagination({
+        pageSize,
+        page,
+    }: Required<ResultsPaginationArgs>) {
+        if (page < 1) {
+            throw new ResultsPaginationError('page should be greater than 0');
```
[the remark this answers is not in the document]
almeidabbm: Mostly because of the status code, `ParameterError` uses 400 while this one is 422, actually I should probably deduplicate it with `KnexPaginationError`, wdyt?

ZeRego (approved):

Issue #13929, opened 2025-03-05
Exporting a pivoted CSV doesn't respect "show table names"


rephus:
### Description

![Image](https://github.com/user-attachments/assets/d2f42164-bdbc-428b-9019-b43000d7ba77)

### Steps to Reproduce the Bug or Issue

- create a table viz 
- pivot the table
- select "show table names"
- export csv

### Cloud or self-hosting

None

rephus:
You may notice some columns name are different when exporting, I created a separte issue for this, and this is unrelated to this PR https://github.com/lightdash/lightdash/issues/13929 (afaik we already pass this argument, but it is ignored on the pivot, need to investigate)

rubiagatra:
Can I help this?

virendrapatil24:
Hey @rubiagatra, thanks for your interest! I’ve already started working on this and will raise a PR soon.

rubiagatra:
Sure go ahead! @virendrapatil24

virendrapatil24:
Hi @rephus, could you please assign this issue to me? I have raised a PR for it. Let me know if anything else needs to be addressed.

virendrapatil24:
Hi @rephus, hope you doing good! I just wanted to follow up on this. Raised PR #13993 for this fix.

jpetey75:
related to https://github.com/lightdash/lightdash/issues/17570

PR #13930, opened 2025-03-05
fix: metrics as rows in csv

rephus:
Closes: 

### Description:

You may notice some columns name are different when exporting, I created a separte issue for this, and this is unrelated to this PR https://github.com/lightdash/lightdash/issues/13929 (afaik we already pass this argument, but it is ignored on the pivot, need to investigate)

Before

Screenshot from 2025-03-05 08-46-18

After

Explore
Screenshot from 2025-03-05 09-04-31

Saved chart
Screenshot from 2025-03-05 09-05-01

dashboard

Screenshot from 2025-03-05 09-08-06

### Reviewer actions


packages/backend/src/controllers/exploreController.ts
```diff
@@ -143,7 +143,7 @@ export class ExploreController extends BaseController {
             columnOrder: string[];
             hiddenFields?: string[];
             chartName?: string;
-            pivotColumns?: string[];
+            pivotConfig?: PivotConfig;
```
rephus, line 146: I'm refactoring the pivotConfig api, to use pivotCOnfig instead of separate parameters for pivotColumns and metricsAsRows.

packages/backend/src/controllers/exploreController.ts
```diff
@@ -189,11 +180,11 @@ export class ExploreController extends BaseController {
                 csvLimit,
                 showTableNames,
                 customLabels,
-                columnOrder,
-                hiddenFields,
                 chartName: body.chartName,
                 fromSavedChart: false,
-                pivotConfig: csvPivotConfig,
+                columnOrder,
```
rephus, line 185: columnOrder and hiddenFields are included in pivotConfig, but since pivtoConfig is optional, we use it separately, as it is needed for non pivoted tables.

magnew (approved):
✅ Makes sense
✅ Works as expected

rephus:
e2e errors seem unrelated. It is complaining about custom dimensions in explorer

PR #13937, opened 2025-03-05
chore: set correct sentry tags on saveScreenshot errors

joaoviana:
Closes: N/A

### Description:

Better grouping of errors since now there is a ScreenshotError class

Set tags correctly and remove some unnecessary ones

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -11,9 +11,11 @@ import {
     LightdashPage,
     LightdashRequestMethodHeader,
     RequestMethod,
+    ScreenshotError,
     SessionUser,
     snakeCaseName,
 } from '@lightdash/common';
+import * as Sentry from '@sentry/node';
 import * as fsPromise from 'fs/promises';
 import { nanoid as useNanoid } from 'nanoid';
 import fetch from 'node-fetch';
@@ -273,6 +275,7 @@ export class UnfurlService extends BaseService {
         const details = await this.unfurlDetails(url);
 
         const buffer = await this.saveScreenshot({
+            authUserUuid,
             imageId,
             cookie,
             url,
@@ -376,6 +379,7 @@ export class UnfurlService extends BaseService {
         lightdashPage,
         chartType,
         organizationUuid,
+        authUserUuid,
         gridWidth = undefined,
         resourceUuid = undefined,
         resourceName = undefined,
@@ -392,6 +396,7 @@ export class UnfurlService extends BaseService {
         lightdashPage?: LightdashPage;
         chartType?: string;
         organizationUuid?: string;
+        authUserUuid: string;
         gridWidth?: number | undefined;
         resourceUuid?: string;
         resourceName?: string;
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -416,10 +421,22 @@ export class UnfurlService extends BaseService {
...
-            'UnfurlService.saveScreenshot',
-            {},
+        return Sentry.startSpan(
+            {
+                name: 'UnfurlService.saveScreenshot',
+                op: 'saveScreenshot',
+            },
             async (span) => {
+                Sentry.setTags({
+                    'page.type': lightdashPage ?? 'undefined',
```
joaoviana, line 431: tags need to be like this for them to be set on the issue level 'TAG'

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -779,22 +787,12 @@ export class UnfurlService extends BaseService {
                                 e,
                             )}`,
                         );
-                        span.addEvent(getErrorMessage(e));
```
joaoviana, line 782: unnecessary

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -779,22 +787,12 @@ export class UnfurlService extends BaseService {
...
+
                         span.setAttributes({
-                            'page.type': lightdashPage,
-                            url,
-                            chartType: chartType || 'undefined',
-                            organization_uuid: organizationUuid || 'undefined',
-                            uuid: resourceUuid ?? 'undefined',
-                            title: resourceName ?? 'undefined',
                             is_retrying: true,
-                            custom_width: `${gridWidth}`,
```
joaoviana, line 791: we don't care about custom_width at this point because nobody complained.

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -392,6 +396,7 @@ export class UnfurlService extends BaseService {
         lightdashPage?: LightdashPage;
         chartType?: string;
         organizationUuid?: string;
+        authUserUuid: string;
```
joaoviana, line 399: this saveScreenshot can come from a scheduled delivery OR a dashboard export on the UI

packages/backend/src/services/UnfurlService/UnfurlService.ts
notgiorgi: @joaoviana it needs to be a string `"undefined"`?

notgiorgi (approved):

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -947,7 +949,7 @@ export class UnfurlService extends BaseService {
             body: JSON.stringify({ token }),
         });
         if (response.status !== 200) {
-            throw new Error(
+            throw new AuthorizationError(
```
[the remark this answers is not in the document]
notgiorgi: are we ignoring it as in do-not-track or do-not-alert?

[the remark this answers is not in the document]
joaoviana: do-not-track in Sentry, which won't alert

packages/backend/src/services/UnfurlService/UnfurlService.ts
joaoviana: otherwise the tag isn't sent so I would rather have it set to undefined, hope that's ok!

Issue #7021, opened 2023-09-06
Lightdash CLI with BigQuery should work with GitHub OIDC in GitHub Actions


yu-iskw:
### Description
We use GitHub OIDC with Google Cloud to get credentials for BigQuery. We tried to execute Lightdash CLI commands in GitHub Actions with OIDC. However, that doesn't work. If we use a credentials of Google Cloud created by GitHub OIDC with the Lightdash CLI as `lightdash deploy`, I got the error message. 

```
Cannot get credentials from oauth
```

According to my research, the current implementation to deal with BigQuery credentials doesn't support the external service account type of Google Cloud as we get

https://github.com/lightdash/lightdash/blob/88c37ffef5340a802e1f24f08c3c4189fb202a1b/packages/cli/src/dbt/targets/Bigquery/oauth.ts

The contents in `GOOGLE_APPLICATION_CREDENTIALS` created by GitHub OIDC looks as the below.

```
{
  "type": "external_account",
  "audience": "//iam.googleapis.com/projects/xxxxxxxx/locations/global/workloadIdentityPools/my-pool/providers/my-provider",
  "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
  "token_url": "https://sts.googleapis.com/v1/token",
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/***@****.iam.gserviceaccount.com:generateAccessToken",
  "credential_source": {
    "url": "https://pipelines.actions.githubusercontent.com/********/00000000-0000-0000-0000-000000000000/_apis/distributedtask/hubs/Actions/plans/*********/jobs/*******/idtoken?api-version=2.0&audience=https%3A%2F%2Fiam.googleapis.com%2Fprojects%2F265863795685%2Flocations%2Fglobal%2FworkloadIdentityPools%2Fmy-pool%2Fproviders%2Fmy-provider",
    "headers": {
      "Authorization": "***"
    },
    "format": {
      "type": "json",
      "subject_token_field_name": "value"
    }
  }
}
```

### To Reproduce

1. Create a dbt project to be deployed to Lightdash like `jaffle_shop`

2. Create a profile with the `oauth` type to execute a Lightdash CLI command in GitHub Actions

```yaml
jaffle_shop:
  outputs:
    github:
      type: bigquery
      method: oauth
      project: "*******************"
      dataset: "jaffle_shop"
      threads: 5
      location: "asia-northeast1"
```

[...]

PR #13955, opened 2025-03-06
Add OAuth Token based method for dbt BiqQuery client

weeezes:
Closes: https://github.com/lightdash/lightdash/issues/7021 because you can now get the access token and set it to an environment variable and use it as such in a `profiles.yml`:

```
xyz:
  target: prd
  outputs:
    prd:
      dataset: ...
      location: ...
      method: oauth-secrets
      priority: interactive
      project: ...
      type: bigquery
      token: "{{ env_var('DBT_ENV_SECRET_ACCESS_TOKEN') }}"
```

### Description:

This adds support for the [dbt OAuth Token-Based](https://docs.getdbt.com/docs/core/connect-data-platform/bigquery-setup#oauth-token-based) authentication method.

`google-auth-library` and `@google-cloud/bigquery` had to be updated for versions that support the `authClient` parameter.

### Reviewer actions


packages/cli/src/dbt/targets/Bigquery/oauthSecrets.ts
```diff
@@ -0,0 +1,102 @@
...
+        location: {
+            type: 'string',
+            nullable: true,
+        },
+        execution_project: {
+            type: 'string',
+            nullable: true,
+        },
+    },
+    required: ['type', 'dataset', 'method'],
```
ZeRego, line 75: To fix, threads is required in type and optional in schema.

packages/cli/src/dbt/targets/Bigquery/index.ts
```diff
@@ -79,43 +82,58 @@ export const convertBigquerySchema = async (
...
             maximumBytesBilled: target.maximum_bytes_billed,
             location: target.location,
             executionProject: target.execution_project,
         };
+
+        if (getKeyfileContents) {
+            const keyfile = await getKeyfileContents(target);
+
+            if ('credentials' in keyfile) {
+                creds.authClient = new OAuth2Client(keyfile);
```
ZeRego, line 130: Could we store the OAuth2ClientOptions instead of the OAuth2Client instance? 
I'm surprised saving the instanced worked when creating projects/previews as the credentials are sent to the server/database, to later be used in the UI.

ZeRego (commented):
This looks promising! 🙌  
Does this work with both a [Refresh token](https://docs.getdbt.com/docs/core/connect-data-platform/bigquery-setup#refresh-token) and a [Temporary token](https://docs.getdbt.com/docs/core/connect-data-platform/bigquery-setup#temporary-token)?

Can you share a video of this working in the CLI and UI?

weeezes:
Darn, didn't consider the UI side at all. I made a few fixes based on your feedback but I need to take another - bit deeper - look at how this should work.

owlas:
Hey @weeezes as a coincidence we've just merged https://github.com/lightdash/lightdash/pull/15026 which adds support for the full SSO flow, would that solve the use case for you? So instead of generating your own token and including it, you can use a browser based sign-in with google, and we'll obtain and store the access token for you.

owlas:
hey we haven't been able to prioritize this review yet and are closing this PR. we'll pick it back up as soon as we can!

Issue #13943, opened 2025-03-06
Upgrade to Sentry v9


notgiorgi:
### Description

Upgrade to Sentry v9

Front-end: https://docs.sentry.io/platforms/javascript/guides/react/migration/v8-to-v9/
Back-end: https://docs.sentry.io/platforms/javascript/guides/node/migration/v8-to-v9/

notgiorgi:
# Back-end (Node.js)

Guide https://docs.sentry.io/platforms/javascript/guides/node/migration/v8-to-v9/

- [x] Minimum Compatible ECMAScript Version: ES2020
- [x] Minimum supported Node.js version: 18.0.0
- [x] minimum required TypeScript version is 5.0.4

### Behavioral/API changes

- [x] ~Prisma version 5.x and below are no longer supported (minimum supported version is 6.x).~
- [ ] Dropping spans in the beforeSendSpan hook is no longer possible. This means you can no longer return null from the beforeSendSpan hook. This hook is intended to be used to add additional data to spans or remove unwanted attributes (for example for PII stripping). To control which spans are recorded, we recommend configuring [integrations](https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/) instead.
- [ ] The beforeSendSpan hook now receives the root span as well as the child spans. We recommend checking your beforeSendSpan to account for this change.
- [ ] The request property on the samplingContext argument passed to the tracesSampler and profilesSampler options has been removed. samplingContext.normalizedRequest can be used instead. Note that the type of normalizedRequest differs from request.
- [ ] The startSpan behavior was changed if you pass a custom scope: While in v8, the passed scope was set active directly on the passed scope, in v9, the scope is cloned. This behavior change does not apply to @sentry/node where the scope was already cloned. This change was made to ensure that the span only remains active within the callback and to align behavior between @sentry/node and all other SDKs. As a result of the change, span hierarchy should be more accurate. However, modifying the scope (for example, set tags) within the startSpan callback behaves a bit differently now.
[...]

PR #13956, opened 2025-03-06
chore: upgrade sentry to v9

notgiorgi:
Closes: #13943

### Description:

Upgrade guide: https://docs.sentry.io/platforms/javascript/guides/node/migration/v8-to-v9/

#### Platform requirements

- [x] Minimum Compatible ECMAScript Version: ES2020
- [x] Minimum supported Node.js version: 18.0.0
- [x] minimum required TypeScript version is 5.0.4

#### Behavioral/API changes

- [x] ~Prisma version 5.x and below are no longer supported (minimum supported version is 6.x).~
- [x] Dropping spans in the beforeSendSpan hook is no longer possible. This means you can no longer return null from the beforeSendSpan hook. This hook is intended to be used to add additional data to spans or remove unwanted attributes (for example for PII stripping). To control which spans are recorded, we recommend configuring [integrations](https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/) instead.
- [x] The beforeSendSpan hook now receives the root span as well as the child spans. We recommend checking your beforeSendSpan to account for this change.
- [x] The request property on the samplingContext argument passed to the tracesSampler and profilesSampler options has been removed. samplingContext.normalizedRequest can be used instead. Note that the type of normalizedRequest differs from request.
- [x] The startSpan behavior was changed if you pass a custom scope: While in v8, the passed scope was set active directly on the passed scope, in v9, the scope is cloned. This behavior change does not apply to @sentry/node where the scope was already cloned. This change was made to ensure that the span only remains active within the callback and to align behavior between @sentry/node and all other SDKs. As a result of the change, span hierarchy should be more accurate. However, modifying the scope (for example, set tags) within the startSpan callback behaves a bit differently now.
- [x] Passing undefined as a tracesSampleRate option value will now be treated the same as if the attribute was not defined at all. In previous versions, it was checked whether the tracesSampleRate property existed in the SDK options to decide if trace data should be propagated for tracing. Consequentially, this sometimes caused the SDK to propagate negative sampling decisions when tracesSampleRate: undefined was passed. This is no longer the case and sampling decisions will be deferred to downstream SDKs for distributed tracing. This is more of a bugfix rather than a breaking change, however, depending on the setup of your SDKs, an increase in sampled traces may be observed.
- [x] If you use the optional captureConsoleIntegration and set attachStacktrace: true in your Sentry.init call, console messages will no longer be marked as unhandled (handled: false) but as handled (handled: true). If you want to keep sending them as unhandled, configure the handled option when adding the integration:
- [x] The tracesSampler hook will no longer be called for every span. Root spans may however have incoming trace data from a different service, for example when using distributed tracing.
- [x] The requestDataIntegration will no longer automatically set the user from request.user when express is used. Starting in v9, you'll need to manually call Sentry.setUser() e.g. in a middleware to set the user on Sentry events.
- [x] The processThreadBreadcrumbIntegration was renamed to childProcessIntegration.
- [x] The childProcessIntegration's (previously processThreadBreadcrumbIntegration) name value has been changed from "ProcessAndThreadBreadcrumbs" to "ChildProcess". Any filtering logic for registered integrations should be updated to account for the changed name.
- [x] ~The vercelAIIntegration's name value has been changed from "vercelAI" to "VercelAI" (capitalized). Any filtering logic for registered integrations should be updated to account for the changed name.~
- [x] ~The Prisma integration no longer supports Prisma v5 and supports Prisma v6 by default. As per Prisma v6, the previewFeatures = ["tracing"] client generator option in your Prisma Schema is no longer required to use tracing with the Prisma integration.~
- [x] When skipOpenTelemetrySetup: true is configured, httpIntegration({ spans: false }) will be configured by default. You no longer have to specify this manually. With this change, no spans are emitted once skipOpenTelemetrySetup: true is configured, without any further configuration being needed.
- [x] The transactionContext property on the samplingContext argument passed to the tracesSampler and profilesSampler options has been removed. All object attributes are available in the top-level of samplingContext:
- [x] The enableTracing option was removed. Instead, set tracesSampleRate: 1 or tracesSampleRate: 0.
- [x] The autoSessionTracking option was removed.
- [x] The getCurrentHub(), Hub and getCurrentHubShim() APIs have been removed. They were on compatibility life support since the release of v8 and have now been fully removed from the SDK.
- [x] The addOpenTelemetryInstrumentation method has been removed. Use the openTelemetryInstrumentations option in Sentry.init() or your custom Sentry Client instead.
- [x] The debugIntegration has been removed. To log outgoing events, use [Hook Options](https://docs.sentry.io/platforms/javascript/guides/node/configuration/options/#hooks) (beforeSend, beforeSendTransaction, ...).
- [x] The sessionTimingIntegration has been removed. To capture session durations alongside events, use [Context](https://docs.sentry.io/platforms/javascript/guides/node/enriching-events/context/) (Sentry.setContext()).
- [x] The addOpenTelemetryInstrumentation method was removed. Use the openTelemetryInstrumentations option in Sentry.init() or your custom Sentry Client instead.
- [x] registerEsmLoaderHooks now only accepts true | false | undefined. The SDK will default to wrapping modules that are used as part of OpenTelemetry Instrumentation.
- [x] ~The nestIntegration was removed. Use the NestJS SDK (@sentry/nestjs) instead.~
- [x] ~The setupNestErrorHandler was removed. Use the NestJS SDK (@sentry/nestjs) instead.~

#### Type Changes
- [x] Scope usages now always expect Scope instances.
- [x] Client usages now always expect BaseClient instances. The abstract Client class was removed. Client classes now have to extend from BaseClient.

#### Package removals
- [x] `@sentry/utils` package will no longer be published.
- [x] Removed `@sentry/types` package (will continue to be published, however, it is deprecated and its API will not be extended. It will not be published as part of future major versions.)

### Reviewer actions


packages/backend/src/App.ts
```diff
@@ -494,6 +494,16 @@ export default class App {
...
 
+        expressApp.use((req, res, next) => {
+            if (req.user) {
+                Sentry.setUser({
+                    id: req.user.userUuid,
+                    organization: req.user.organizationUuid,
+                });
+            }
+            next();
+        });
```
notgiorgi, line 505: > The requestDataIntegration will no longer automatically set the user from request.user when express is used. Starting in v9, you'll need to manually call Sentry.setUser() e.g. in a middleware to set the user on Sentry events.

packages/backend/src/sentry.ts
```diff
@@ -40,13 +43,14 @@ Sentry.init({
...
-            context.request?.url?.endsWith('/favicon.ico') ||
-            context.request?.url?.endsWith('/robots.txt') ||
-            context.request?.url?.endsWith('livez') ||
-            context.request?.headers?.['user-agent']?.includes('GoogleHC')
+            request?.url?.endsWith('/status') ||
+            request?.url?.endsWith('/health') ||
+            request?.url?.endsWith('/favicon.ico') ||
+            request?.url?.endsWith('/robots.txt') ||
+            request?.url?.endsWith('livez') ||
+            request?.headers?.['user-agent']?.includes('GoogleHC')
```
notgiorgi, line 53: > The request property on the samplingContext argument passed to the tracesSampler and profilesSampler options has been removed. samplingContext.normalizedRequest can be used instead. Note that the type of normalizedRequest differs from request.

packages/backend/src/utils.ts
```diff
@@ -4,8 +4,8 @@ import {
     SshKeyPair,
     validateEmail,
 } from '@lightdash/common';
+import { CustomSamplingContext } from '@sentry/core';
```
notgiorgi, line 7: > All exports and APIs of @sentry/utils and @sentry/types (except for the ones that are explicitly called out in this migration guide to be removed) have been moved into the @sentry/core package.

packages/backend/src/App.ts
joaoviana: can you update the `SchedulerTaskTracer` `setUser` to include 'organization': 'payloadTags['organization.uuid']'?

joaoviana (approved):
left 2 requests above ^ on setting the user. works great

Issue #13957, opened 2025-03-06
Address `getAuthenticationToken` error when unfurling urls in Slackbot


notgiorgi:
https://lightdash.sentry.io/issues/6013870246/?environment=cloud_beta&project=5959292&referrer=issue-stream&statsPeriod=14d&stream_index=4

PR #13958, opened 2025-03-06
fix: Address `getAuthenticationToken` error when unfurling urls in Slackbot

notgiorgi:
Closes: #13957

### Description:

`getAuthenticationToken` function was receiving `null` as an argument and failing
https://github.com/lightdash/lightdash/blob/85bc1adad69b46070f8dd7976cf6ffde7b1780dc/packages/backend/src/routers/headlessBrowser.ts#L15

The only way for this to happen is for `getUserUuid` function to return null https://github.com/lightdash/lightdash/blob/3d04a286ff909d5a27fa734bf8775354bb33fc6d/packages/backend/src/clients/Slack/Slackbot.ts#L193

This could happen in cases when user which installed the slack integration got deleted, row got cascaded and `user_uuid` points to `NULL`

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SlackAuthenticationModel.ts
+++ b/packages/backend/src/models/SlackAuthenticationModel.ts
@@ -102,7 +102,7 @@ export class SlackAuthenticationModel {
             )
             .select<(DbSlackAuthTokens & DbUser)[]>('*')
             .where('slack_team_id', teamId);
-        if (isNil(row) || isNil(row.user_uuid)) {
+        if (isNil(row?.user_uuid)) {
             throw new Error(`Could not find user uuid id ${teamId}`);
         }
         return row.user_uuid;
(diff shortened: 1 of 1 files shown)


notgiorgi:
https://lightdash.sentry.io/issues/6013870246/?environment=cloud_beta&project=5959292&referrer=issue-stream&statsPeriod=14d&stream_index=4

joaoviana (approved):
Great find!

packages/backend/src/models/SlackAuthenticationModel.ts
```diff
@@ -101,7 +102,7 @@ export class SlackAuthenticationModel {
             )
             .select<(DbSlackAuthTokens & DbUser)[]>('*')
             .where('slack_team_id', teamId);
-        if (row === undefined) {
+        if (isNil(row) || isNil(row.user_uuid)) {
```
BackEndTea, line 105: This could probably be shortened with optional chaining: 

```suggestion
        if (isNil(row?.user_uuid)) {
```

notgiorgi: Thank you for your suggestion!

PR #13965, opened 2025-03-07
fix: subtotals check if null or undefined

almeidabbm:
Fixes: https://lightdash.sentry.io/issues/6363892358/?project=4507168845529088&referrer=github-pr-bot
Related PR: https://github.com/lightdash/lightdash/pull/13847#issuecomment-2704200370

image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -190,38 +190,32 @@ const PivotTable: FC<PivotTableProps> = ({
                                 const subtotal = data.groupedSubtotals?.[
                                     subtotalGroupKey
                                 ]?.find((sub) => {
-                                    if (
-                                        typeof sub !== 'object' ||
-                                        sub === null ||
-                                        typeof groupingValues !== 'object' ||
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -190,6 +190,18 @@ const PivotTable: FC<PivotTableProps> = ({
...
+                                        typeof groupingValues !== 'object' ||
+                                        groupingValues === null ||
+                                        typeof pivotedHeaderValues !==
+                                            'object' ||
+                                        pivotedHeaderValues === null
+                                    ) {
+                                        return false;
+                                    }
+
                                     return (
```
rephus, line 205: we could also wrap this function inside a try-catch. IMO is better to not display 1 particular record, than to break the UI

almeidabbm: Done in the latest commit!
I added an error log as well so that we can check what is happening if this happens, but can remove if you don't think it is necessary

rephus (approved):

PR #13967, opened 2025-03-07
fix: optimize sql runner search

rephus:
Closes: 

### Description:

issues 

- because of the search hook, we were doing the `select` on every rerender : fixed by moving the filtering and select into a memo 
- table was using search, not debounced search, so it was rendering too often 
- when search happens, we expand all tables, but when characters were deleted, we were showing all items without filtering, causing too many items to render 
- 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/Tables.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/Tables.tsx
@@ -31,6 +31,8 @@ import { useTables, type TablesBySchema } from '../hooks/useTables';
 import { useAppDispatch, useAppSelector } from '../store/hooks';
 import { setSql, toggleActiveTable } from '../store/sqlRunnerSlice';
 
+const limitTableResults = 100;
+
 interface TableItemProps extends BoxProps {
     table: string;
     search: string;
@@ -186,15 +188,16 @@ const Table: FC<{
     }, [tables, schema, search]);
 
     useEffect(() => {
-        if (
+        const isTableSelected =
             activeTable &&
             Object.keys(tables).includes(activeTable) &&
-            schema === activeSchema
-        ) {
+            schema === activeSchema;
+        if (hasMatchingTable || isTableSelected) {
             setIsExpanded(true);
-        }
-        if (hasMatchingTable) {
-            setIsExpanded(true);
+        } else {
+            // Autoclose when search is empty and no matching table is selected
+            // to avoid rendering all tables after a search
+            setIsExpanded(false);
         }
     }, [activeTable, tables, hasMatchingTable, activeSchema, schema]);
     return (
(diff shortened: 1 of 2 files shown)


packages/frontend/src/features/sqlRunner/components/Tables.tsx
```diff
@@ -24,13 +24,16 @@ import {
...
 import dayjs from 'dayjs';
+import Fuse from 'fuse.js';
 import { memo, useEffect, useMemo, useState, type FC } from 'react';
 import MantineIcon from '../../../components/common/MantineIcon';
 import { useIsTruncated } from '../../../hooks/useIsTruncated';
 import { useTables, type TablesBySchema } from '../hooks/useTables';
 import { useAppDispatch, useAppSelector } from '../store/hooks';
 import { setSql, toggleActiveTable } from '../store/sqlRunnerSlice';
 
+const limitTableResults = 100;
```
magnew, line 35: If there are more, they just have to get more specific with the search?

packages/frontend/src/features/sqlRunner/hooks/useTables.tsx
```diff
@@ -50,13 +48,8 @@ export const useTables = ({ projectUuid, search }: GetTablesParams) => {
...
         retry: false,
         enabled: !!projectUuid,
-        select(data) {
-            const tablesBySchema = Object.entries(data).flatMap(([, schemas]) =>
-                Object.entries(schemas).map(([schema, tables]) => ({
-                    schema,
-                    tables,
-                })),
-            );
+        /*select(data) {
```
magnew, line 51: You wanna keep this?

packages/frontend/src/features/sqlRunner/components/Tables.tsx
```diff
@@ -118,17 +121,21 @@ const TableItem: FC<TableItemProps> = memo(
...
-                            component={Text}
-                            highlight={search || ''}
-                            truncate
-                            sx={{
-                                flex: 1,
-                            }}
-                        >
-                            {table}
-                        </Highlight>
+                        {search.length > 2 ? (
```
magnew, line 124: I know its a common pattern to require 3 chars, but even knowing that I was slightly confused when I typed 2. Could we just render a help text like "Enter at least 3 characters to search" when its < 3?

magnew (approved):
Looks good. I made a UX suggestion. 

Good thinking fixing the FE up first.

packages/frontend/src/features/sqlRunner/components/Tables.tsx
rephus: how about this ? 
image

magnew (approved):

rephus:
tested manually the e2e error , seems to work well on render

image

Issue #13884, opened 2025-03-03
Address Sentry reporting of `invalid_grant` error in Worker - Google Sheets


joaoviana:
This is a transient error that shouldn't be reported in Sentry. 

There should be some checking of the error messaging on `SchedulerTask` or similar. 
Perhaps we don't want to propagate the error when it happens, but maybe we catch it to show on scheduler errors/slack notifications.

PR #13982, opened 2025-03-07
fix: handle unauthorized and invalid_grant errors from google

notgiorgi:
Closes: #13884

### Description:

The problem was that:

1.  [`GoogleDriveClient#getCredentials` method](https://github.com/lightdash/lightdash/blob/eeb6b20382a475c8ee27d2e949640353d46c5a5f/packages/backend/src/clients/Google/GoogleDriveClient.ts#L39-L55) is not actually checking wether existing credentials are valid, only checks if they exist
2. Each different request to google could throw an 401/400 error
3. The errors thrown were not `ForbiddenError`s  thus they weren't caught by [`shouldDisableSync`](https://github.com/lightdash/lightdash/blob/ab042358d45223ef15492715417ab47d6c19cb4c/packages/backend/src/scheduler/SchedulerTask.ts#L1992-L1995) and sync was not disabled

As a solution I wrapped each google api call in a try/catch that properly handles the errors

Before:
image

After:
image

To reproduce:
1. Setup `Google Sheets Sync` on a chart
2. Manually revoke Lightdash app from the google accounts
3. Wait for sync worker to run and you'll get `invalid_grant` error
### Reviewer actions

### DIFF
--- a/packages/backend/src/clients/Google/GoogleDriveClient.ts
+++ b/packages/backend/src/clients/Google/GoogleDriveClient.ts
@@ -54,32 +54,27 @@ export class GoogleDriveClient {
         }
     }
 
-    private static async changeTabTitle(
-        sheets: sheets_v4.Sheets,
-        fileId: string,
-        title: string,
-    ) {
-        const spreadsheet = await sheets.spreadsheets.get({
-            spreadsheetId: fileId,
-        });
-        await sheets.spreadsheets.batchUpdate({
-            spreadsheetId: fileId,
-            requestBody: {
-                requests: [
-                    {
-                        updateSheetProperties: {
-                            properties: {
-                                sheetId:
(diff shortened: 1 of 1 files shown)


packages/backend/src/clients/Google/GoogleDriveClient.ts
```diff
@@ -54,32 +54,27 @@ export class GoogleDriveClient {
         }
     }
 
-    private static async changeTabTitle(
```
rephus, line 57: why was this method removed ?

rephus (approved):
LGTM

packages/backend/src/clients/Google/GoogleDriveClient.ts
notgiorgi: it was unused

PR #13987, opened 2025-03-07
fix(refactor): query results hook

ZeRego:
Closes: 

### Description:

Refactor query handling for Explorer, Chart and ChartHistory  

Main changes:
- Unify hook/query to get results for explore, chart and chart version.
- Switch `useQueryResults` from using `useMutation` to use `useQuery`.
- Removed `useQueryResults` `ExplorerProvider` property in favor of directly passing required props like `dateZoomGranularity` and `viewModeQueryArgs`.

Demo create chart:

https://github.com/user-attachments/assets/17523d5d-d003-4d7d-abdb-f8eddbe9dffd

Demo chart history:

https://github.com/user-attachments/assets/6bf3ecdb-d627-4eaf-915b-3a5541113bf5

Demo create dashboard chart:

https://github.com/user-attachments/assets/875a6937-1146-401f-8fc6-7d8086239f10

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

Issue #3136, opened 2022-09-02
Don't cut off long-running queries


owlas:
### Description

the `/runQuery` endpoint in syncronous meaning a lot of network setups will eventually cut off the http request for long running queries. We should make this async.

This would need two things in the UI to help:
1. Polling with a timer showing how long the query is taking
2. The ability to cancel a query

Basit-Balogun10:
Ability to cancel a query is being tracked further in #1406 and has the client-side implementation done in #13991

jpetey75:
we have rebuilt the query engine since this issue was created, if it comes up again under the new infrastructure we can reopen this issue

Issue #1406, opened 2022-02-17
You should be able to cancel a query


TuringLovesDeathMetal:
### Description: What is it?

When you hit `run query`, it can be really annoying if you have to wait for it to run, but you know you made a mistake. You should be able to hit `cancel query` when you've hit `run query` and it's still loading.

### Problem: What problem does this solve?

 it can be really annoying if you have to wait for it to run, but you know you made a mistake.

### (Optional) Why: How do we know this is a real problem and worth solving?

https://getdbt.slack.com/archives/C026WJE4A69/p1645039472102019?thread_ts=1644794904.784429&cid=C026WJE4A69

### Success: How will we know if we’ve solved this problem?

You're able to cancel your query before it's completed

### Audience: Who are we building this for?

Data builders - people adding things to the results tables and building charts.

### What: Roughly, what does this look like in the product?

When the `run query` button is running, if you hover over the button, it would say `cancel` with an `x`. Clicking on this would cancel the query. 

A query that's cancelled would ideally keep the previous results that were in the table (it wouldn't leave you with a blank table, or load up the new results if they had already been fetched)

**This would also cancel the query in the data warehouse!** not just in the UI in Lightdash.

ZeRego:
Docs on how to implement this https://tanstack.com/query/v4/docs/framework/react/guides/query-cancellation#manual-cancellation

yileeZAVA:
Hello team, hope you all are well. Just checking if this feature is still planned to be enabled? Thank you.

cyclonstep:
Hello team, 
because this issues have been requested since two years ago and I found out there isn't any movements at the moment, let me upvote this one too. 

Also, for the background and motivation of why it would be better to have this feature (at least in my team): 
- People are making mistakes while writing queries. (and mistakes not always in wrong query syntax)
- Not everyone appointed as data builders are understand or even aware exact situations in the connected data warehouse.
- Adhoc queries for quick analysis on new business process will often need to querying and joining some unoptimized tables or models.
- For time-based data warehouse solutions (like snowflake), wasted time means wasted costs. 

Thank you very much 🙏

Basit-Balogun10:
## Progress Update

Work in progress in #13991, Requirements checked so far:

- ✅ User can cancel a query via the UI
- ✅ Previous results are preserved after cancellation

### Implementation Plan

I propose splitting this into two PRs:

1. **Current PR**: Client-side implementation (UI and query cancellation)
[...]

PR #13991, opened 2025-03-08
feat(frontend): allow users to cancel a query in explore

Basit-Balogun10:
Closes: 

### Description:

Client-side implementation for #1406 and #3136, branched off the refactor PR https://github.com/lightdash/lightdash/pull/13987 to utilize the new `useQuery` hook implementation in `packages/frontend/src/hooks/useQueryResults.ts` (might be better to merge #13987 before this)

**Implementation Details:**
- We're leveraging [React Query's built-in cancellation support](https://tanstack.com/query/v4/docs/framework/react/guides/query-cancellation#manual-cancellation) through the `cancelQueries` method 
- When "Cancel query" is clicked in the UI, we call queryClient.cancelQueries() with the appropriate query key. Cancellation is also triggered by pressing the `esc` key while a query is running.
- The UI is improved to only show the cancel button after a user has manually run a query, not during initial page load
- We maintain the last successful result in the cache so users still see data when a query is canceled instead of an empty stat

**Testing Notes**
For anyone testing this functionality:
- I personally have to throttle the runQuery request in the browser DevTools Network tab to a "Very slow" speed (3kb/s - just a random value) to properly test the cancellation. This is because, at normal network speeds, the queries and requests usually complete too quickly to be canceled in time.
- I'm assuming most users don’t experience this in real business use cases, given how frequently this issue is raised and the number of upvotes it has received (also mentioned in #3136).
- This throttling mimics real-world scenarios where queries take longer to complete (which seems to be common given the number of upvotes/requests for this feature)
- 

[lightdash-allow-users-to-cancel-query-in-explore.webm](https://github.com/user-attachments/assets/ba1e79fb-062b-4bb2-b48d-a7013ae69c24)

### Reviewer actions


Basit-Balogun10:
Sounds good, no problem!

Issue #13830, opened 2025-02-26
Dashboard filter tile targets break on dashboard reupload


oli-rmsy:
### Description

When uploading a dashboard as code dashboard yml files, tile target config is ignored.

### Steps to Reproduce the Bug or Issue

1. download a dashboard with some filters pointing to specific tiles.
2. reupload that dashboard immediately (use the force flag to bypass the skip)
3. Observe that tile target are incorrect (everything is default included with every filter)

### Cloud or self-hosting

None

PR #14000, opened 2025-03-10
fix: filters on dashboards as code

rephus:
Closes: https://github.com/lightdash/lightdash/issues/13830

### Description:

Fixed by replacing the tile uuid with a chart slug on `download`, then matching the filter back on `upload`

If there are multiple charts with the same slug, we assign them an index based on their position in the tile list. 

Before:

Screenshot from 2025-03-10 11-02-55

After:

Screenshot from 2025-03-10 11-46-17

Dashboard UI after upload on a different project: 

Screenshot from 2025-03-10 12-21-16

dashboard filters (uploaded vs original) 
Screenshot from 2025-03-10 12-21-50

### Edge case:

Note: There is an edge case where this will not work , since we are relying on the position of the tiles when they are saved. 

Edge case not covered (replication steps) : 
- Have a dashboard with the same `chart` duplicated in different tiles
- Add a dashboard filter
- Deselect one of the filters on one of the duplicated tiles 
- Download code 
- Update the code, and remove one of the duplicated charts from the list of tiles 
- Upload the code
- :fire: now the chart that didn't have a filter applied **might** have the filter applied (since we apply filters by default)  

I belive this is a extreme edge case, and it is not worth fixing.  The fix will force us to persist some index in db when saving the dashboard, since we can't rely on anything else 

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.test.ts
+++ b/packages/backend/src/services/CoderService/CoderService.test.ts
@@ -1,0 +1,149 @@
+// CoderService.test.ts
+import {
+    AnyType,
+    DashboardAsCode,
+    DashboardChartTileProperties,
+    DashboardDAO,
+    DashboardFilterRule,
+    DashboardTile,
+    DashboardTileTarget,
+} from '@lightdash/common';
+import { CoderService } from './CoderService';
+
+describe('CoderService.getFiltersWithTileSlugs', () => {
+    it('should convert tile UUIDs to slugs in filters', () => {
+        const mockDashboard: DashboardDAO = {
+            filters: {
+                dimensions: [
+                    {
+                        tileTargets: {
+                            'uuid-1': {
+                                fieldId: 'field-1',
+                            } as DashboardTileTarget,
+                            'uuid-2': {
+                                fieldId: 'field-2',
+                            } as DashboardTileTarget,
+                        },
+                    },
+                ],
+            },
+            tiles: [
+                { uuid: 'uuid-1', properties: { chartSlug: 'slug-1' } },
+                { uuid: 'uuid-2', properties: { chartSlug: 'slug-2' } },
+            ],
+        } as AnyType; // Use 'as any' to bypass type checks for simplicity
+
+        const result = CoderService.getFiltersWithTileSlugs(mockDashboard);
+
+        expect(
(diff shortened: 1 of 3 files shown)


IrakliJani (commented):
Code looks good, but I’m worried about the edge case. I still think a dedicated `index` column for `dashboard_tiles` table is a safer approach, with a unique constraint on (`dashboard_version_id`, `index`). --- (it can be a slug column, or some kind of unique identifier that is persisted when cloning project) 

That way we can reference this index in `tileTargets`, avoiding potential slug collisions and the hacky logic of matching tile order in the code.

If you are comfortable with the current approach, feel free to merge, but I’m personally worried that we'd need to introduce a  breaking change to fix the edge case mentioned in the PR description.

rephus:
I agree the index is a safer approach, but I still think the edge case impact is very low. And we still need to populate this index for all existing dashboards, which is going to be a pain. 

I have a proposal: let's merge this as it is, and if someone complains, we can introduce the index, and check the `version` of the DashboardAsCode file to apply the correct version on upload.

magnew:
I think I agree with both of you 😬. It does seem like a very edge-case. I'm not to concerned about that particular case, but I'm wondering if we might want that index anyway for anything in the future that might also care about tile indices (for example, the translation in the SDK will have this problem also). 

One more thing to consider is maybe how heavily this feature (content as code) will be used. This seems like an edge case in an environment where people are curating dashboards by hand, but in cases like kraken where they plan to basically create and destroy dashboards programatically, I can imagine they might regularly have the same chart with multiple filters and remove one sometimes. OR never. It's hard to know. 

Anyway, @rephus, nice work, code looks good. I agree with Irakli that if you are up for it, merge it, but we may have to add the index and deal with that later.

magnew:
> I agree the index is a safer approach, but I still think the edge case impact is very low. And we still need to populate this index for all existing dashboards, which is going to be a pain.
> 
> I have a proposal: let's merge this as it is, and if someone complains, we can introduce the index, and check the `version` of the DashboardAsCode file to apply the correct version on upload.

Woah, I think we commented at the same time 😄. And came to similar conclusions.

magnew (approved):

rephus:
I merged this and created a new ticket in the milestone to add the index

Issue #13935, opened 2025-03-05
Create paginated query metadata model


almeidabbm:
### Description

We need to save some metadata related to the query that is being paginated.
This is because whenever we want to use just the query id to fetch a second page, we won't have information like:
1. `metricQuery`
2. `fields` - `itemsMap`
3. `compiledSql`

Additionally this metadata can save information about the pagination itself:
1. total number of rows
2. number of pages
3. page size

PR #14016, opened 2025-03-10
Feat/add query history table

almeidabbm:
Closes: [#13935](https://github.com/lightdash/lightdash/issues/13935)

### Description:
- Adds new `query_history` table which will hold the metadata for a query
- Fetch the query history when fetching more pages using a query id
- If there is no `warehouseQueryId` then we just run the query again as the default behavior from warehouse client, otherwise we fetch more results from the warehouse query id (snowflake for now)
- Adds new env var for max page size which is 500 (default limit) * 5

### Reviewer actions


packages/backend/src/services/ProjectService/resultsPagination.ts
```diff
@@ -0,0 +1,37 @@
+import type { ResultsPaginationArgs } from '@lightdash/common';
+
+import { PaginationError } from '@lightdash/common';
+
+export function getNextAndPreviousPage(page: number, totalPageCount: number) {
+    const nextPage =
+        page === totalPageCount || totalPageCount === 0 ? undefined : page + 1;
```
ZeRego, line 7: non blocker: check if page is not greater than total

packages/backend/src/models/QueryHistoryModel.ts
```diff
@@ -0,0 +1,71 @@
...
+                total_row_count: queryHistory.totalRowCount,
+            })
+            .returning('query_uuid');
+
+        return {
+            queryUuid: result.query_uuid,
+        };
+    }
+
+    async get(queryUuid: string, projectUuid: string) {
```
ZeRego, line 63: Naming convention we should aim for: 
getX = throws error if there is no result
findX = returns undefined if there is no result

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1768,17 +1777,134 @@ export class ProjectService extends BaseService {
...
+        if (!queryHistory) {
+            throw new NotFoundError(`Query with uuid ${queryUuid} not found`);
+        }
+
+        if (
+            user.ability.cannot(
+                'view',
+                subject('Project', { organizationUuid, projectUuid }),
+            ) ||
+            projectUuid !== queryHistory.projectUuid ||
```
ZeRego, line 1808: We already filter by project uuid in the `queryHistoryModel`. We could also search by user uuid and remove this logic.

almeidabbm: Added in: [chore: get query history using userUuid](https://github.com/lightdash/lightdash/pull/14016/commits/a5b19d07065af3f8361762ec7ce5261997a15ec3)

packages/backend/src/models/QueryHistoryModel.ts
almeidabbm: Done in: [fix: throw error when QueryHistory not found](https://github.com/lightdash/lightdash/pull/14016/commits/56a42afe057f3c0ecea0adfd7ea79b1e303a3c7e)

packages/backend/src/services/ProjectService/resultsPagination.ts
almeidabbm: Here: [chore: check if page is not greater than totalPageCount](https://github.com/lightdash/lightdash/pull/14016/commits/8438d1b8b77bf9f7884ca06bb2f3d60376096a62)

ZeRego (approved):
lgtm 🏅

Issue #13950, opened 2025-03-06
Some series are rendering with transparent colors


magnew:
### Description

Some users are seeing some series rendered as transparent. 

Here are two internal threads:
https://lightdash.slack.com/archives/C075JNX28GG/p1741260050661739?thread_ts=1741184477.004629&cid=C075JNX28GG

https://lightdash.slack.com/archives/C049LQZGVD3/p1741255522099489

### Steps to Reproduce the Bug or Issue

Find a repro

### Cloud or self-hosting

None

rephus:
I was able to replicate 

![Image](https://github.com/user-attachments/assets/d00699b0-9698-4756-a288-878c6e1e6a2f)

the series with blank colors were series that were added after the chart was saved, so they don't existing in the chart config.

How to replicate:
- make a chart
- add grouping
- add a filter, to remove most of the groups
- without runing query again, hit save

PR #14020, opened 2025-03-11
fix: empty colors on dashboards

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13950
### Description:

What was the issue

On charts, we compute cartesian chart config from echarts on explore, by running a method called `getExpectedSeriesMap` from the results. 

So the chartConfig we pass to the visualization provider is complete. 
Screenshot from 2025-03-11 08-42-09

On dashboards, we don't pass the right chartConfig to the visprovider, only the initial chart config, which is missing the "new" series. 
Screenshot from 2025-03-11 08-42-43

This is causing fallbackColors to be empty for some series. 

Solution: compute series on dashboarcharttile before visProvider. 

### Before

Screenshot from 2025-03-11 09-54-35

### After

Screenshot from 2025-03-11 09-56-04

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -1,8 +1,9 @@
 import { subject } from '@casl/ability';
 import {
+    CartesianSeriesType,
     ChartType,
-    DashboardTileTypes,
     createDashboardFilterRuleFromField,
+    DashboardTileTypes,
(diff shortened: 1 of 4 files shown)


magnew (approved):
Niiiiiice work. This makes sense. Its pretty hard to rigorously test, but it the implementation squares with the issues I've seen. 

I tried to find anything new breaking and couldn't ✅ 
I think I found one actual repro that it fixes ✅

rephus:
pivo ttable works for me on render 
image

PR #14022, opened 2025-03-11
feat: Enable API call to operate warehouse credentials

yu-iskw:
Closes: NA

Please let me know if I should open an issue.

### Description:

Currently, we can not call the APIs for workflow credentials as `getWarehouseCredentials`. But if the APIs are available, we can implement the seamlesslly integrate Lightdash with the warehouse service.

For instance, we can create service account key with terraform. If the `updateWarehouseCredentials` API is available, we can seamlessly register the credentials in Lightdash. We won't need manual operations any more. 

### Reviewer actions

### DIFF
--- a/packages/backend/src/controllers/userController.ts
+++ b/packages/backend/src/controllers/userController.ts
@@ -235,7 +235,7 @@ export class UserController extends BaseController {
     /**
      * Get user warehouse credentials
      */
-    @Middlewares([isAuthenticated])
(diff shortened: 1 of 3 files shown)


rephus:
That makes sense @yu-iskw , I'm happy to add these. 

Could you run `pnpm generate-api` to ensure the files are uptodate ?

yu-iskw:
I have executed `pnpm generate-api` locally to generate the schema.

rephus (approved):

Issue #14018, opened 2025-03-10
CSV Download from SQL Runner returns 0 as '-'


oli-rmsy:
### Description

When downloading as a CSV, the sql runner returns a - instead of a 0 when 0 is shown in the results table.

### Steps to Reproduce the Bug or Issue

Open SQL Runner
Enter query: `select 0 as example`
Run download CSV
Observe incorrect CSV value

### Cloud or self-hosting

None

PR #14026, opened 2025-03-11
fix: don't treat all falsey values as undefined in sql-runner

magnew:
Closes: #14018 

### Description:

SQL runner was rendering all falsey values as `-`. This makes it render 0, null, false and undefined consistently with other places we download csvs. 

### Reviewer actions

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -1444,8 +1444,5 @@ describe('Formatting', () => {
             expect(applyDefaultFormat(1234)).toBe('1,234');
             expect(applyDefaultFormat(-1234)).toBe('-1,234');
         });
-        it('should handle arrays', () => {
-            expect(applyDefaultFormat([1, 2, 3])).toBe('1,2,3');
-        });
     });
 });
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -12,6 +12,7 @@ import {
 import { TimeFrames } from '../types/timeFrames';
 import {
     applyCustomFormat,
+    applyDefaultFormat,
     convertCustomFormatToFormatExpression,
     currencies,
     formatItemValue,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/features/sqlRunner/hooks/useDownloadResults.tsx
```diff
@@ -22,7 +22,19 @@ export const downloadCsv = async (
...
+            if (value === null) {
+                return '∅';
+            }
+            if (value === undefined) {
+                return '-';
+            }
+            if (value === false) {
+                return 'false';
+            }
+            return value;
```
IrakliJani, line 36: `applyDefaultFormat` can be used here I think

magnew: Ah, nice, thanks. I was looking for something like that. It's not exported, but I think it makes sense to export it instead of using one of the exported functions from formatting -- those are all kind of specific to our explorey types. 

Updated.

packages/frontend/src/features/sqlRunner/hooks/useDownloadResults.tsx
```diff
@@ -1,4 +1,5 @@
 import { type RawResultRow, type VizColumnsConfig } from '@lightdash/common';
+import { applyDefaultFormat } from '@lightdash/common/src';
```
IrakliJani, line 2:
```suggestion
import { applyDefaultFormat } from '@lightdash/common';
```

packages/common/src/utils/formatting.test.ts
```diff
@@ -1423,4 +1424,28 @@ describe('Formatting', () => {
...
+        it('should handle strings', () => {
+            expect(applyDefaultFormat('foo')).toBe('foo');
+        });
+        it('should handle numbers', () => {
+            expect(applyDefaultFormat(1234567)).toBe('1,234,567');
+            expect(applyDefaultFormat(1234)).toBe('1,234');
+            expect(applyDefaultFormat(-1234)).toBe('-1,234');
+        });
+        it('should handle arrays', () => {
+            expect(applyDefaultFormat([1, 2, 3])).toBe('1,2,3');
```
IrakliJani, line 1448: I think this is the default `toString`(arrays->string) behavior and not part of the original implementation of `applyDefaultFormat`

packages/frontend/src/features/sqlRunner/hooks/useDownloadResults.tsx
magnew: Thanks. I didn't check my IDE's work well enough. My bad.

IrakliJani (approved):

Issue #14029, opened 2025-03-11
Fix sourcemaps uploading


joaoviana:
### Description

https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/vite/

### Steps to Reproduce the Bug or Issue

n/a

### Cloud or self-hosting

None

PR #14038, opened 2025-03-11
fix: sentry release and sourcemaps on both projects

joaoviana:
Closes: #14029 

### Description:

Fixes sourcemaps upload issue and path issues when checking errors in Sentry

Appends `~` on BE

Screenshot 2025-03-12 at 09 57 34

And `~/assets` on FE

Screenshot 2025-03-12 at 09 57 27

So that the unminify path is correctly recognised.

Uses `sentry-cli` to upload both sourcemaps on build time. Since the addition of Debug IDs it isn't recommended to have 2 separate builds (like we currently have a separate sentry release action) and upload the sourcemaps on one that won't be used by our users. So I've removed that action.

Sentry plugin for vite has been added. This handles the setting of debug IDs etc but does **not** handle the sourcemaps upload. I would rather have that centralised in one place for the BE and FE in the Dockerfile. If the necessary env vars are set, then it will run the builds accordingly. 

 
### Reviewer actions


packages/frontend/vite.config.ts
```diff
@@ -16,10 +17,9 @@ export default defineConfig({
         svgrPlugin(),
         reactPlugin(),
         compression({
-            include: [/\.(js)$/, /\.(css)$/, /\.js\.map$/],
```
joaoviana, line 19: dont compress map files.

notgiorgi (approved):
Great work! 🚀

Issue #12724, opened 2024-12-04
I want to show images from URL in fields


12ian34:
### Description

Like [Looker's implementation here](https://cloud.google.com/looker/docs/reference/param-field-html), I would like to use html image tags to show images within Lightdash, which are to be called from external URLs and shown to the users in navigation.

Ideally we wouldn't require users to add custom HTML, but instead just show an image, for example in a table cell, when the data coming form the database is a URL that hosts that image.

This may be useful in charts too, like a bar chart, but most commonly it's used in tables.

### What problem does this solve?

e.g. if a product team is looking at sales data for many products, it can help identify the product easier if there is an actual image loaded, rather than having to read the title.

rephus:
POC Implemented on https://github.com/lightdash/lightdash/pull/18019

Docs available: https://docs.lightdash.com/references/dimensions#image-display

PR #14058, opened 2025-03-13
feat: allow showing html images in fields (#12724)

freiondrej:
Closes: #12724

### Description:

Hello!

This is my first contribution to this project. It's in a rough draft state, as I would first like to get your feedback if this is in general the right direction to tackle the given issue (its Slack link is older than 90 days so I cannot access the thread, sadly).

image

The configuration is very similar to the Looker example suggested by the issue, the one from the screenshot is using this:
```yml
        meta:
          dimension:
            type: html
            html: <img src="https://placehold.co/150x100?text=${value.raw}"/> ;
```

I'll highly appreciate your guidance on which other places should reflect this change (probably using the `dangerouslySetInnerHtml` prop, like I tried in the BodyCell component).

Then, I'll polish up the code and update / add tests, and documentation.

Thank you, this project is incredible and the DX is *really* good, I was able to draft this in ~1 hour after starting playing with it 🚀 👏 

### Reviewer actions

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.ts
+++ b/packages/common/src/compiler/filtersCompiler.ts
@@ -438,6 +438,7 @@ export const renderFilterRuleSql = (
         : field.compiledSql;
 
     switch (fieldType) {
+        case DimensionType.HTML:
         case DimensionType.STRING:
         case MetricType.STRING: {
             return renderStringFilterSql(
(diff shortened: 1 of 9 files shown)


joaoviana:
Closing due to inactivity

Issue #13990, opened 2025-03-07
Saving New Chart Redirects to First Tab


olaonikosi:
Description:
"When a user creates and saves a new chart within a specific tab in Lightdash, the application automatically redirects them to the first tab. This disrupts the user's workflow and requires them to manually navigate back to the tab where they were working.

Steps to Reproduce:
Navigate to a tab other than the first tab in a Lightdash dashboard or exploration.
Create a new chart within the currently active tab.
Save the newly created chart.
Observe that Lightdash navigates back to the first tab.

Expected Behavior:
After saving a new chart, Lightdash should maintain the user's context within the currently active tab.

Impact:
Disrupted workflow
Increased navigation burden
Disrupted iteration
Inconsistency and frustration
Redundant steps."

PR #14067, opened 2025-03-13
fix: saving new chart redirects to first tab

magnew:
Closes: #13990 

### Description:

Primarily this makes it so that when you edit a chart on a dashboard tab and go back to the dashboard, you'll be in the tab you were editing. This part is pretty straightforward. 

This required a change to the way the dashboard gets mounted. We were unmounting and remounting the whole dashboard on tab change, when the URL changed. But this meant the context was lost every time you switched tabs, so we couldn't support proper editing across tabs. This 'worked' before because in edit mode we didn't look at the URL for tabs. But it made it impossible to deep link to the tabs during edit. 

This PR changes the router to NOT unmount and remount the dashboard on tab change. This will help perf, and also fixes a couple bugs I saw when editing across tabs, but is definitely a change. As far as I can tell it works well. The old behavior may have been necessary with the old react router, which we have now upgraded. 

To test:
- Make a dashboard with multiple tabs
- Add a new (not saved) chart to a tab other than the default
- Save your chart and go back to the dashboard. 
- You should land on the new tab with the new chart.
- Also note that switching tabs during edit now changes the tabID in the url  

### Reviewer actions

### DIFF
--- a/packages/frontend/src/Routes.tsx
+++ b/packages/frontend/src/Routes.tsx
@@ -1,6 +1,6 @@
 import { Stack } from '@mantine/core';
 import { type FC } from 'react';
-import { Navigate, Outlet, useParams, type RouteObject } from 'react-router';
+import { Navigate, Outlet, type RouteObject } from 'react-router';
 import AppRoute from './components/AppRoute';
 import ForbiddenPanel from './components/ForbiddenPanel';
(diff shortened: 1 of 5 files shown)


ZeRego (approved):
lgtm 👍

Issue #14041, opened 2025-03-12
Support underlying data query in new paginated endpoint


ZeRego:
### Description

At the moment the FE builds a metric query and requests data via the endpoint:
`/projects/${projectUuid}/explores/${tableId}/runUnderlyingDataQuery`

We should only pass the necessary information(eg: explore id, sort, limit) to the new endpoint `/api/v2/projects/{projectUuid}/query` and let the BE generate the metric query.

### What problem does this solve?

Migrate all queries endpoints to the new endoint with pagination.

Issue #13916, opened 2025-03-04
Optimise UI to only fetch necessary results 


almeidabbm:
### Description

The results table and the visualisation should fetch results independently. 
This will allow each to only load the necessary results pages.

Example:

Results page only loads the first page. Loads more pages as the user scrolls or changes page.
Bar chart loads all pages.

### Context

What components display results:
- viz (charts/table/etc)
- results table
- underlying data table

What pages/endpoints use those components but don't support pagination yet:
- embed dashboard
- underlying data ( waiting for [#14041](https://github.com/lightdash/lightdash/issues/14041))

What warehouses ignore `pageSize` endpoint param and return a single page with all the rows:
- trino
- postgres
- databricks

Challenges:
- keep UI pagination pagination decoupled from API+warehouse pagination
- How to incrementally add lazy load support since the explorer provider, dashboard, viz and table components are all expecting all the rows.

PR #14142, opened 2025-03-19
feat: lazy load rows for results table

ZeRego:
Related to: [#13916](https://github.com/lightdash/lightdash/issues/13916)

### Context:

What components display results:
- viz (charts/table/etc)
- results table
- underlying data table

What pages/endpoints use those components but don't support pagination yet:
- embed dashboard
- underlying data ( waiting for [#14041](https://github.com/lightdash/lightdash/issues/14041))

What warehouses ignore `pageSize` endpoint param and return a single page with all the rows:
- trino
- postgres
- databricks

Challenges/Goals:
- keep UI pagination pagination decoupled from API+warehouse pagination
- How to incrementally add lazy load support since the explorer provider, dashboard, viz and table components are all expecting all the rows.

### Changes:

For this PR I'm focusing on the results table in the explorer page:
- refactor `common/Table` component to have manual pagination
- split results hook into 2:
  - one that creates the query and waits for it to be ready
  - one that gets the rows based on UI needs
- lazy load rows for results table
- load all rows for viz 

### Demo

Results table loading prefetching rows for next pages and loads all rows when in scroll mode.

https://github.com/user-attachments/assets/db6d2a30-2249-4316-9b71-83dc8e005b5e

Viz loads all rows for now

https://github.com/user-attachments/assets/1401c2a0-c23d-4795-a033-3426c67ebd52

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -476,16 +476,18 @@ export type ApiExecuteAsyncQueryResults = {
     appliedDashboardFilters: DashboardFilters | null;
 };
 
+export type ReadyQueryResultsPage = ResultsPaginationMetadata<ResultRow> & {
+    queryUuid: string;
+    rows: ResultRow[];
+    fields: ItemsMap;
+    initialQueryExecutionMs: number;
+    resultsPageExecutionMs: number;
+    metricQuery: MetricQuery;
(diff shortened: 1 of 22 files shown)


almeidabbm:
Getting this error 
image

suspect it is tied to the infinite results but haven't found the culprit

Issue #14146, opened 2025-03-20
Fix underlyingData hook 'explore must be used within explore provider'


almeidabbm:
### Description

https://lightdash.sentry.io/issues/6431631740/?referrer=issue_alert-slack&notification_uuid=0df2fc7c-c836-4d6c-9652-6ce0f3240d90&environment=cloud_beta&alert_rule_id=15154466&alert_type=issue

### Steps to Reproduce the Bug or Issue

1. Open dashboard
2. Click `view underlying data` in one of the charts

### Cloud or self-hosting

None

PR #14147, opened 2025-03-20
fix: dashboard chart underlying data queryUuid

almeidabbm:
Closes: #14146 

AC:
- should still work for chart view
- explorer

### Description:

**Before:**

https://github.com/user-attachments/assets/1579feee-b5fc-4cd7-b999-53d3a23c9144

**After:**

https://github.com/user-attachments/assets/3eafe811-4093-4a1f-a94c-c3481b058d24

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -1212,7 +1212,7 @@ type DashboardChartTileProps = Omit<
 export const GenericDashboardChartTile: FC<
     DashboardChartTileProps & {
         isLoading: boolean;
-        data: ApiChartAndResults | undefined;
+        data: (ApiChartAndResults & { queryUuid?: string }) | undefined;
         error: ApiError | null;
     }
 > = ({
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
```diff
@@ -1212,7 +1212,7 @@ type DashboardChartTileProps = Omit<
 export const GenericDashboardChartTile: FC<
     DashboardChartTileProps & {
         isLoading: boolean;
-        data: ApiChartAndResults | undefined;
+        data: (ApiChartAndResults & { queryUuid?: string }) | undefined;
```
joaoviana, line 1215: wonder why "queryUuid" could always be optional. Can you explain that?

joaoviana (approved):
tested

packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
almeidabbm: Yes, this is because of the new pagination work, on the frontend we're requesting either the new pagination endpoint or the old one, depending on a feature flag. The new endpoint adds a queryUuid to the returned results.

This should be cleaned up once we remove the feature flag and old code

PR #14152, opened 2025-03-20
fix: simplify color on pivots, unify logic with pie charts

rephus:
Closes: 

### What the goal is 
For calculate color method, make logic consistent for barchars and pie charts for the same groups

Before

Screenshot from 2025-03-20 09-44-54

AFter

Screenshot from 2025-03-20 14-19-07

### Description:

The first thing I noticed, is that the method that was calculating the color (based on the serie/pivot) was different for pie charts and bar charts 

Pie charts

image

Screenshot from 2025-03-20 14-23-23

Bar charts 

Screenshot from 2025-03-20 14-23-14

This is because we were using on barcharts pivot field, rather than pivotValue[].field. After updating this, the logic was a lot simpler. 

Screenshot from 2025-03-20 14-26-38

It even works for charts with multigropus

Screenshot from 2025-03-20 14-29-29
Screenshot from 2025-03-20 14-26-58

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useChartColorConfig/utils.ts
+++ b/packages/frontend/src/hooks/useChartColorConfig/utils.ts
@@ -14,59 +14,17 @@ export const isGroupedSeries = (series: SeriesLike) => {
 };
 
 export const calculateSeriesLikeIdentifier = (series: SeriesLike) => {
-    const baseField =
-        (series as EChartSeries).pivotReference?.field ??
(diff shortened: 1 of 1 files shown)


joaoviana (approved):
tested many series types, stacked and not. Flipped axis and colors weren't affected. Tested as well in dashboards and embedded dashboards

PR #14168, opened 2025-03-24
fix: support ENCRYPTED PRIVATE KEY format in profiles.yml

TrsNium:
Closes: 

### Description:
- Updated regex pattern for private key detection
  - Previously only matched `-----BEGIN PRIVATE KEY-----`
  - Now also supports `-----BEGIN ENCRYPTED PRIVATE KEY-----`

## Testing Checklist
- [ ] Standard private keys (BEGIN PRIVATE KEY) are processed correctly as before
- [ ] Encrypted private keys (BEGIN ENCRYPTED PRIVATE KEY) are processed correctly

### Reviewer actions

### DIFF
--- a/packages/cli/src/dbt/templating.ts
+++ b/packages/cli/src/dbt/templating.ts
@@ -31,7 +31,7 @@ export const renderProfilesYml = (
     // Fix multiline privatekey strings
     // Prevents error: Error: error:1E08010C:DECODER routines::unsupported
     const privateKeyRegex =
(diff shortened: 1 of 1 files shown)


rephus:
Thanks for your contribution. For context, this is only used when multiline private keys is not correctly parsed in the YML file. Another option woudl be to use the private_key_path property in profiles.

rephus (approved):

PR #14182, opened 2025-03-25
feat: lazy load table viz

ZeRego:
Closes: 

### Description:

lazy load rows on table viz. 

todo: load all rows for pivot tables.

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/SimpleTable/index.tsx
+++ b/packages/frontend/src/components/SimpleTable/index.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex } from '@mantine/core';
+import { noop } from '@mantine/utils';
 import { IconAlertCircle } from '@tabler/icons-react';
-import { type FC } from 'react';
+import { type FC, useCallback, useMemo } from 'react';
 import { isTableVisualizationConfig } from '../LightdashVisualization/types';
 import { useVisualizationContext } from '../LightdashVisualization/useVisualizationContext';
 import { LoadingChart } from '../SimpleChart';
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/SimpleTable/index.tsx
```diff
@@ -29,21 +34,94 @@ const SimpleTable: FC<SimpleTableProps> = ({
...
+
+    useEffect(() => {
+        if (shouldPaginateResults) return;
+
+        // Load all the rows
+        resultsData?.setFetchAll(true);
+        return () => {
+            resultsData?.setFetchAll(false);
+        };
+    }, [shouldPaginateResults, resultsData]);
```
almeidabbm, line 119: We paginate results when:
- Is pivoted table
- Is showing subtotals

packages/frontend/src/components/SimpleTable/index.tsx
```diff
@@ -29,21 +34,94 @@ const SimpleTable: FC<SimpleTableProps> = ({
...
+            return 'loading';
+        }
+
+        // When paginated, it's success as soon as there are rows
+        // When not paginated, it's success as soon as all rows have been fetched
+        const isSuccess = shouldPaginateResults
+            ? resultsData.rows.length > 0
+            : resultsData.hasFetchedAllRows;
+
+        return isSuccess ? 'success' : 'loading';
```
magnew, line 61: We don't have a type for this?

magnew (approved):
Very nice! 🚀 

Code looks good. I don't see anything that needs a change. I focussed on testing a lot of things. I have two questions, neither needs to block, although I wonder if the one about "Rows in table". Tested:

✅ Loads on scroll for normal table
✅ Loads all rows when there are subtotals
✅ Column total shown before lazy load
✅ Loads all rows when pivoted
✅ Load all rows for cartesian chart
✅ Export all rows looks good
✅ Scheduled delivery returns all rows
✅ Dashboard Scheduled delivery sends all rows

❓Export -> Results in table. The behavior has changed here. On main it does all results from the search (because we load them all). Now it exports a different number depending on where you have scrolled to. I wonder if "Results in table" should fetch each page? Or if we should re-think the UI between "Results in Table" and "All results". It is less meaningful with this work.

❓Dashboard tile loads all rows, expected?

magnew:
Retested, looks good!

PR #14204, opened 2025-03-28
fix: fix snowflake password and encryption override

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/14205

### Description:

What the issue was: 

It is possible to have a private key already, and if that exists, we prioritize private_key over password, even if password authentication type was selected 

Before
Screenshot from 2025-03-28 14-59-36

After

Screenshot from 2025-03-28 15-03-51

### Reviewer actions

### DIFF
--- a/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
@@ -148,15 +148,11 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
 
         let authenticationOptions: Partial<ConnectionOptions> = {};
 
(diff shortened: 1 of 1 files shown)


packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -148,7 +148,15 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
 
         let authenticationOptions: Partial<ConnectionOptions> = {};
 
-        if (credentials.privateKey) {
+        if (
+            credentials.password &&
+            credentials.authenticationType === 'password'
```
rephus, line 153: Now we check for authenticationType, even if privateKey is defined

IrakliJani (approved):

PR #14207, opened 2025-03-28
fix: use only valid pivot dimensions when computing chart series

almeidabbm:
Closes: #14206
Relates to: https://github.com/lightdash/lightdash/issues/13950

### Description:
- After fixing series with transparent colors, we were passing pivot dimensions before sanitization by the usePivotDimensions hook which caused a bug when a chart was saved with a pivot dimension but didn't have it in the results.

**Steps to reproduce:**
1. Create cartesian chart with grouped dimension
2. Remove the dimension in the group and save chart (don't run query) - this will cause the config to have a pivot dimension but no matching selected dimension
3. Add chart to dashboard
4. Open dashboard

**Before**
image

**After**
image

### Reviewer actions


rephus (approved):
Nice fix.

This is still compatible with the transparent color fix

PR #14215, opened 2025-03-31
chore: disable timezone tests

rephus:
Closes: 

### Description:

They are normally failing, and we ignore them most of the time

### Reviewer actions

test-timezone


rephus:
Without test-timezone

image

magnew (approved):
Makes sense. I think we are going to address timezones again in the next few months so maybe we can bring them back and fix them then.

Issue #14223, opened 2025-03-31
Failure to fetch query-pagination flag can result in a "No data available" error


magnew:
### Description

Failure to fetch query-pagination flag can result in a "No data available" error.

A customer was seeing "No data available" on dashboards and investigations show that we aren't correctly handling the error thrown when Lightdash couldn't fetch the query-pagination flag. 

We should ignore those errors and fall back to running the warehouse query the old way. 

Investigation notes in this internal thread:
https://lightdash.slack.com/archives/C03MCQ08UKS/p1742826925414249

### Steps to Reproduce the Bug or Issue

n/a

### Cloud or self-hosting

None

PR #14224, opened 2025-03-31
fix: enable useDashboardChart regardless of feature flag value, the c…

almeidabbm:
…heck is done internally

Closes: #14223 

### Description:

- The hook for dashboard charts was disabled when result of feature flag check was empty. This happens when there's an error making a request to `api/v2/feature-flag/query-pagination`.

**Steps to reproduce**
- Force the 404 error when getting the feature flag value in FeatureFlagModel
```typescript
    public async get(args: FeatureFlagLogicArgs): Promise<FeatureFlag> {
        const handler = this.featureFlagHandlers[args.featureFlagId];
        // if (handler) {
        //     return handler(args);
        // }
        // // Default to check Posthog feature flag
        // if (args.user && isFeatureFlags(args.featureFlagId)) {
        //     return FeatureFlagModel.getPosthogFeatureFlag(
        //         args.user,
        //         args.featureFlagId,
        //     );
        // }
        throw new NotFoundError(`Feature flag ${args.featureFlagId} not found`);
    }
```
- Open a dashboard

**Before**
image

**After**
image

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/dashboard/useDashboardChart.ts
+++ b/packages/frontend/src/hooks/dashboard/useDashboardChart.ts
@@ -185,7 +185,7 @@ const useDashboardChart = (tileUuid: string, chartUuid: string | null) => {
                 ? queryKey.concat([granularity])
                 : queryKey,
         queryFn: fetchChartAndResults,
-        enabled: !!chartUuid && !!dashboardUuid && !!queryPaginationEnabled,
+        enabled: !!chartUuid && !!dashboardUuid,
         retry: false,
         refetchOnMount: false,
     });


packages/frontend/src/hooks/dashboard/useDashboardChart.ts
```diff
@@ -185,7 +185,7 @@ const useDashboardChart = (tileUuid: string, chartUuid: string | null) => {
                 ? queryKey.concat([granularity])
                 : queryKey,
         queryFn: fetchChartAndResults,
-        enabled: !!chartUuid && !!dashboardUuid && !!queryPaginationEnabled,
```
magnew, line 188: Why was this here? 😄

magnew (approved):
LGTM

packages/frontend/src/hooks/dashboard/useDashboardChart.ts
almeidabbm: Not sure, but it was completely blocking the fallback to legacy endpoint, I should've noticed it earlier on the review 😬

Issue #14256, opened 2025-04-02
Value drop down in filter group not collapsing


almeidabbm:
### Description

When I have a filter group and open the value dropdown. sometimes it doesn't collapse.
This happens on both outside click and value selection

https://github.com/user-attachments/assets/62e7fc79-29dc-4330-990d-36acad3ef446

### Steps to Reproduce the Bug or Issue

1. Create chart
2. Add filter
3. Convert to group
4. Add second filter
5. Try to change the value. drop down doesn't close

### Cloud or self-hosting

None

PR #14258, opened 2025-04-02
fix: in the current filter stop autofocus when there's a value

almeidabbm:
Closes: #14256 

### Description:

- Only autofocus select when there's no value selected. This is because there are multiple components with autofocus and there are conflicts

**After**
https://github.com/user-attachments/assets/9a9f9ac3-d992-4e78-9e5d-c5f96ef8a3ec

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
@@ -170,6 +170,7 @@ const FilterRuleForm: FC<Props> = ({
                         closeOnItemClick
                         withArrow
                         arrowPosition="center"
+                        withinPortal
                     >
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
```diff
@@ -170,6 +170,7 @@ const FilterRuleForm: FC<Props> = ({
                         closeOnItemClick
                         withArrow
                         arrowPosition="center"
+                        withinPortal
```
almeidabbm, line 173: This makes the filter group menu not to be hidden by portal boundaries. it was like 

image

and now is

image

notgiorgi (approved):
tested, LGTM

PR #14268, opened 2025-04-03
feat: add nested spaces db / be structure (proposal)

joaoviana:
Closes: 

## Areas of concern

# 🟢 **CONTROLLER** 

### `Get space by uuid - @OperationId('GetSpace')`

Currently returns the space (regardless if it's nested or not)
    **Questions**:
    - Should we support getting the nested spaces of a space in a tree structure?

### `Create space - @OperationId('CreateSpaceInProject')`

Currently creates a space and supports creating a nested space within that space

### `Delete space - @OperationId('DeleteSpace')`

Currently deletes the space (regardless if it's nested or not)
    **Questions**:
    - How would the deletion of a space affect the nested spaces within that space?

### `Update space - @OperationId('UpdateSpace')`

Currently updates the space (regardless if it's nested or not)
    **Actions**:
    - Block updates to the nested space because it will be inherited from root space (MVP)

### `Update Space Access - @OperationId('AddSpaceUserAccess')`

Currently updates a space's access list (regardless if it's nested or not)
    **Actions**:
    - Block updates to the nested space's access list because it will be inherited from root space (MVP)

### `Remove Space Access - @OperationId('RevokeSpaceAccessForUser')`

Currently removes a space's access list (regardless if it's nested or not)
    **Actions**:
    - Block removal of the nested space's access list because it will be inherited from root space (MVP)

### `Grant space access to group - @OperationId('AddSpaceGroupAccess')`

Currently grants a space's access list to a group (regardless if it's nested or not)
    **Actions**:
    - Block granting of the nested space's access list to a group because it will be inherited from root space (MVP)

### `Revoke space access from group - @OperationId('RevokeGroupSpaceAccess')`

Currently revokes a space's access list from a group (regardless if it's nested or not)
    **Actions**:
    - Block revoking of the nested space's access list from a group because it will be inherited from root space (MVP)

# 🟡 **SERVICE**

### getSpace

Used only by the `SpaceController` to get a space by uuid (`operationId: GetSpace`)

### createSpace

Used by the `SpaceController` to create a space in a project (`operationId: CreateSpaceInProject`)

### updateSpace

Used by the `SpaceController` to update a space (`operationId: UpdateSpace`)
    Actions: 
    - Block updates to the nested space because it will be inherited from root space (MVP)

### deleteSpace

Used by the `SpaceController` to delete a space (`operationId: DeleteSpace`)
    Actions: 
    - Handle nested space deletion if parentSpaceUuid is provided

### addSpaceUserAccess

Used by the `SpaceController` to add a user to a space (`operationId: AddSpaceUserAccess`)
    Actions: 
    - Block adding a user to the nested space because it will be inherited from root space (MVP)

### removeSpaceUserAccess

Used by the `SpaceController` to remove a user from a space (`operationId: RevokeSpaceAccessForUser`)
    Actions: 
    - Block removing a user from the nested space because it will be inherited from root space (MVP)

### addSpaceGroupAccess

Used by the `SpaceController` to add a group to a space (`operationId: AddSpaceGroupAccess`)
    Actions: 
    - Block adding a group to the nested space because it will be inherited from root space (MVP)

### removeSpaceGroupAccess

Used by the `SpaceController` to remove a group from a space (`operationId: RevokeGroupSpaceAccess`)
    Actions: 
    - Block removing a group from the nested space because it will be inherited from root space (MVP)

### togglePinning

Used by the `ProjectController` to toggle pinning a space (`operationId: TogglePinning`)
    Actions: 
    - Block pinning of the nested spaces (MVP)

# 🟣 **MODEL**

### getSpaceIdAndName

Used by the `SavedChartModel`:
 - createSavedChart
 - update
 - updateMultiple

Used by the `DashboardModel`:
 - update
 - updateMultiple

This seems fine and doesn't need to change

### getFirstAccessibleSpace

Used by the `DashboardService` to get the first accessible space for a user when creating a dashboard:
 - create
Actions: 
- Return a root space 

Used by the `SavedChartModel` to get the first accessible space for a user when creating a saved chart:
 - createSavedChart
Actions: 
- Return a root space 

Used by `SpaceModel`'s `getSpaceWithQueries` for a migration `03_saved_dashboards.ts`

### getSpaceWithQueries

for a migration `03_saved_dashboards.ts`

### find

Used by `CatalogService`'s `filterChartsWithAccess` to get chart usage of a table in Lightdash

Used by `ValidationService`'s `hidePrivateContent` to hide private content from a user

Used by `SchedulerService`'s `checkViewResource` to check if a user has access to a resource before sending

Used by `PromoteService`'s `getPromotedDashboard` and `getPromotedChart` to get the upstream spaces of a promoted dashboard and chart

Used by `PinningService`'s `getPinnedItems` to get the pinned items a user has access to 

Used by `ContentService`'s `find` to get content a user has access to

Used by `CoderService`'s `getDashboards` and `getCharts` and `getOrCreateSpace`

Used by `ProjectService`'s `getCharts` to get charts a user has access to. `getChartSummaries` to get chart summaries a user has access to. `getMostPopularAndRecentlyUpdated` to get the most popular and recently updated charts a user has access to. `getSpaces` to get spaces a user has access to. `copyContentOnPreview` to duplicate spaces from one project to its preview project - see `duplicateContent` in `ProjectModel`

### get

Used by `UnfurlService`'s `exportDashboard` to check if user can export a dashboard in a private space. 

Used by `DashboardService`'s `create` to check if user can save a dashboard in a particular space.

Used by `SpaceService`'s `togglePinning` to check if user can pin an item in a space.

### getSpaceDashboards

Used by `ProjectService`'s `getMostPopular` to get the most popular dashboards of a space and also `getRecentlyUpdated` to get the recently updated dashboards of a space

### getUserSpaceAccess

Used by 
- `CommentService`
- `DashboardService`
- `ProjectService`
- `PromoteService`
- `SavedChartService`
- `SavedSemanticViewerChartService`
- `SavedSqlService`
- `SchedulerService`
- `SpaceService`
- `UnfurlService`

### getUserSpacesAccess

Used by
- `CatalogService`
- `CoderService`
- `ContentService`
- `DashboardService`
- `PinningService`
- `ProjectService`
- `SearchService`
- `ValidationService`

### getSpaceSqlCharts

Used by `ProjectService`'s `getCharts` to get the saved sql charts of a space. As well as `getMostPopular` and `getRecentlyUpdated` to get the most popular and recently updated saved sql charts of a space

### getSpaceSemanticViewerCharts

Used by `ProjectService`'s `getCharts` to get the saved semantic viewer charts of a space. As well as `getMostPopular` and `getRecentlyUpdated` to get the most popular and recently updated saved semantic viewer charts of a space

### getSpaceQueries

Used by `ProjectService`'s `getMostPopular` and `getRecentlyUpdated` to get the most popular and recently updated saved queries of a space

### getSpaceSummary

Used By
- `CommentService`
- `DashboardService`
- `ProjectService`
- `PromoteService`
- `SavedChartService`
- `SavedSemanticViewerChartService`
- `SavedSqlService`
- `SearchService`
- `SpaceService`

### getSpacesForAccessCheck

Used by `ProjectService`'s `getAvailableFiltersForSavedQueries` to check space access to the charts that belong to a dashboard so that filters are not shown to users that don't have access to the dashboard

### getFullSpace

Used by `PromoteService`'s `upsertSpaces` to create spaces when promoting charts (promoteChart) and dashboards (promoteDashboard)

Used by `SpaceService`'s `getSpace` to get a space by uuid

### createSpace

Used by `SpaceService`'s `createSpace` to create a space

Used by `PromoteService`'s `upsertSpaces` to create spaces when promoting charts (promoteChart) and dashboards (promoteDashboard)
    Actions: 
    - Outline when a space is created by upsertSpaces
    - Handle nested space creation if parentSpaceUuid is provided

Used by `CoderService`'s `upsertDashboard` (`@OperationId('upsertDashboardAsCode')`) to create a space when promoting a dashboard
    Actions: 
    - Outline when a space is created by upsertDashboard 
    - Handle nested space creation if parentSpaceUuid is provided

Used by `CoderService`'s `upsertChart` (`@OperationId('upsertChartAsCode')`) to create a space when promoting a chart
    Actions: 
    - Outline when a space is created by upsertChart 
    - Handle nested space creation if parentSpaceUuid is provided

### deleteSpace

Used by `SpaceService`'s `deleteSpace` to delete a space

### update

Used by `SpaceService`'s `updateSpace` to update a space

Used by `PromoteService`'s `upsertSpaces` to update a space's name when promoting a chart (promoteChart) or dashboard (promoteDashboard)

### addSpaceAccess

Used by `SpaceService`'s `addSpaceUserAccess` to add a user to a space and when `createSpace` is used

### removeSpaceAccess

Used by `SpaceService`'s `removeSpaceUserAccess` to remove a user from a space 

### addSpaceGroupAccess

Used by `SpaceService`'s `addSpaceGroupAccess` to add a group to a space

Used by `PromoteService`'s `upsertSpaces` to add a group to a space when promoting a chart (promoteChart) or dashboard (promoteDashboard)
 Actions: 
    - Outline when a group is added to a space by upsertSpaces
    - Block adding group access to the nested space because it will be inherited from root space (MVP)

### removeSpaceGroupAccess

Used by `SpaceService`'s `removeSpaceGroupAccess` to remove a group from a space

# ⚫️ **User flows**

### Getting all spaces in a project 

Requires endpoint from `ProjectController`'s - `@Get('{projectUuid}/spaces') @OperationId('ListSpacesInProject')`
With hook `useSpaceSummaries` we are getting a flat structure of spaces.
Below you can see the user flows that use this hook and the changes that need to be made to support nested spaces.
Usage with actions: 
 - Creating a Dashboard - we get all spaces a user has access to - do not flat; retrieve nested spaces
 - When moving a resource (chart/dashboard) to a different space - do not flat; retrieve nested spaces
 - When moving a chart to a different space - do not flat; retrieve nested spaces
 - When moving a dashboard to a different space - do not flat; retrieve nested spaces
 - When adding a resource to a space - do not flat; retrieve nested spaces
 - When viewing all spaces in `Browse` on the nav bar - do not flat; **do not** retrieve nested spaces
 - When viewing all spaces in `All Spaces` page - do not flat; **do not** retrieve nested spaces
 - When adding a chart to a dashboard - do not flat; retrieve nested spaces
 - When saving Semantic Viewer, SQL Runner, and changing a SQL Chart's space - do not flat; retrieve nested spaces

### Checking if user can create in any space

`useCreateInAnySpaceAccess` calls `useSpaceSummaries` and that probably doesn't need to be a flat structure, since space permissions of nested spaces are inherited from root spaces.

### Viewing a space

Requires endpoint from `SpaceController`'s - `@Get('{projectUuid}/spaces/{spaceUuid}') @OperationId('GetSpace')`

With hook `useSpace` we are getting a space by uuid.

Actions: 
- When viewing a space 
    - do not flat; retrieve nested spaces but only their uuids & names
    - retrieve current space's access from root space

### Searching for a space (search bar)

Actions:
- Review how nested spaces would be handled in the search bar - will they be included or only root ones?

### Creating previews

`ProjectService`s `copyContentOnPreview` retrieves all spaces from the source project
It then calls `projectModel.duplicateContent` to:
 - duplicate those spaces to the preview project (with a new UUID), while maintaining other properties (except for the project_id which is updated to the new preview project's ID)
 - duplicate space permissions (both user and group access) for the new spaces

Actions: 

- Update `duplicateContent` to handle nested spaces, by creating all root spaces, then iterating over them to create nested ones, and recalculate ltree paths using the new UUIDs - this will be tricky 

### Updating previews

From what I can tell, the current code doesn't recopy content.

### Promotion of content

How spaces are affected during promotion:
- Space Identification:
    - During promotion, spaces are identified by their slug
    - When promoting content, the system checks if a space with the same slug exists in the target project
- Space Creation in Target Project:
    - If a space with the same slug doesn't exist in the target project, a new space is created (`PromoteService.upsertSpaces`)
- Space Handling During Promotion:
    - The source space metadata (name, privacy settings) is carried over to the target
    - If the space already exists, only its name may be updated (permissions are not changed)
    - For new spaces, permissions from the source space are copied

Actions: 
  - Account for hierarchy of spaces during promotion so that nested spaces and relationships between spaces are preserved
  - Promotion process needs to make sure spaces are created in the correct order, starting from the root spaces and then creating nested spaces
  - Path calculation for nested spaces needs to be handled correctly
  - Consider the full path of slugs to identify a spaces position in the hierarchy. 🙋 Does it mean we need to use the slug instead of the uuid as the identifier for a space in `path` column? 

### Content as code

Spaces are identified by their slug only when using content as code. Most of the actions are the same as promotion, but there are some differences:

Actions: 
- Decide on a format for representing nested space paths in slugs (e.g., parent/child/grandchild) and possibly do the same as last action above ⬆️
- If a space with the same slug doesn't exist, a new space is created
- If the space already exists, only its name may be updated (permissions are not changed)
- For new spaces, permissions from the source space are copied
- Shared util functions for both CoderService and PromoteService that handle parsing of paths, creating spaces in hierarchical order, handling permission inheritance (even though permissions of nested spaces are inherited from root spaces, so we might be ok with not needing to handle copying of spaces permissions if we are smart about it)

=================================================================

### Database Schema Changes

1. **Add to spaces table:**

   - `parent_space_uuid UUID NULL` - References parent space (NULL for root spaces)
   - `path ltree NULL` - Stores the hierarchical path in ltree format

2. **Create new indexes:**

   - `CREATE INDEX spaces_path_idx ON spaces USING gist (path)` - For efficient hierarchical queries
   - `CREATE INDEX spaces_parent_space_uuid_index ON spaces USING btree (parent_space_uuid)` - For parent-child lookups

3. **Foreign key constraint:**
   - `ALTER TABLE spaces ADD CONSTRAINT spaces_parent_space_uuid_foreign FOREIGN KEY (parent_space_uuid) REFERENCES spaces(space_uuid) ON DELETE CASCADE`

### Endpoint

`POST` /spaces endpoint now allows creating nested spaces, if a `parentSpaceUuid` is provided

```
POST http://localhost:8080/api/v1/projects/3675b69e-8324-4110-bdca-059031aa8da3/spaces
Content-Type: application/json

{
  "name": "Child Space",
  "isPrivate": false,
  "parentSpaceUuid": "YOUR_PARENT_SPACE_UUID"
}

```

 

### Reviewer actions


packages/backend/src/database/migrations/20250403090130_add_nested_paces_ability.ts
```diff
@@ -0,0 +1,50 @@
+import { Knex } from 'knex';
+
+const tableName = 'spaces';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw('CREATE EXTENSION IF NOT EXISTS ltree');
+
+    await knex.schema.alterTable(tableName, (table) => {
```
joaoviana, line 8: https://www.mermaidchart.com/raw/3a34a3e6-cd49-4213-897e-d9faaef9a852?theme=light&version=v0.1&format=svg

packages/backend/src/database/migrations/20250403090130_add_nested_paces_ability.ts
```diff
@@ -0,0 +1,50 @@
...
+        // Easier to find a space's immediate parent
+        table.uuid('parent_space_uuid').nullable();
+        // Enables much faster queries for finding all descendants or ancestors
+        // Like:
+        // SELECT * FROM spaces WHERE path @> '123'; - This will return all spaces that are descendants of space 123
+        // SELECT * FROM spaces WHERE path <@ '123'; - This will return all spaces that are ancestors of space 123
+        table.specificType('path', 'ltree').nullable();
```
joaoviana, line 15: https://www.postgresql.org/docs/current/ltree.html#LTREE

packages/backend/src/database/migrations/20250403090130_add_nested_paces_ability.ts
```diff
@@ -0,0 +1,50 @@
+import { Knex } from 'knex';
+
+const tableName = 'spaces';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw('CREATE EXTENSION IF NOT EXISTS ltree');
```
joaoviana, line 6: to discuss

packages/backend/src/services/SpaceService/SpaceService.ts
```diff
@@ -155,12 +155,50 @@ export class SpaceService extends BaseService {
...
+                throw new ForbiddenError(
+                    'Parent space must be in the same project',
+                );
+            }
+
+            // Check if user has manage access to parent space
+            const parentSpaceAccess = await this.spaceModel.getUserSpaceAccess(
+                user.userUuid,
+                space.parentSpaceUuid,
+            );
```
joaoviana, line 177: in here, I would call a `getRootSpace` so I could check the root space's permissions before going ahead

packages/backend/src/database/migrations/20250403090130_add_nested_paces_ability.ts
owlas: so it looks like this in in the standard library, so no problem: https://www.postgresql.org/docs/current/ltree.html

owlas (commented):
I can see the paths are made from uuids. This has a couple of cases where this might not work:
- duplicating a space - do we duplicate all children?
- previews + promoting content - how would this work?

owlas:
Treating the controller and service as effectively the same:

> Get space by uuid - @OperationId('GetSpace')
> Currently returns the space (regardless if it's nested or not)
> Questions:
> - Should we support getting the nested spaces of a space in a tree structure?

i don't think this needs to return any tree structure, it just gets a single space


> Delete space - @OperationId('DeleteSpace')
> Currently deletes the space (regardless if it's nested or not)
> Questions:
> - How would the deletion of a space affect the nested spaces within that space?

Cascade makes sense if we inherit all permissions. If permissions could be revoked on nested spaces then this might be more surprising


> togglePinning
> Used by the ProjectController to toggle pinning a space (operationId: TogglePinning)
> Actions:
> - Block pinning of the nested spaces (MVP)

why not have pinning?

> getSpaceIdAndName

curious we have this function that's only used by a few methods

> getSpaceWithQueries

can this get deleted then? we can modify the migration if it's not needed

> getUserSpaceAccess

needs updating to include inherited permissions

> getSpaceQueries

not sure why we need this - can we reduce the number of methods on this model?

> getSpacesForAccessCheck

definitely needs to change to check parent spaces

> Search

How can we implement this without tanking performance? v1 we could exclude nested spaces



### Ideas

**New SpacePermissions type**

Should we have 3 different types:

- `SpaceSummary` - minimum info about a Space (used in lists)
- `Space` - most information about a Space (used in detail view)
- `SpacePermissions` - all detail needed for permission checks

**Reduce methods in SpaceModel**

Some redundant and similar looking methods

**User flows: creating dashboards / charts - this needs to exist**

The list endpoint should return all spaces under a given path (e.g. under root, or under another space). This would power the space browser as well as an input selector to add a dashboard/chart to a space.

Any downside for existing components if we make this change? I don't think so?

**User flows: Viewing a space**

We should probably always include (at least one) parent space and name for showing breadcrumbs. What about the space detail view, would it show children? Probably but we would use the list endpoint

**Slugs include parents**

I like having it such that slugs include all parents `/parent/child/grand-child` - that works great for promoting and as-code because you only have to search for a space with the exact slug (path) and sync it, or if it doesn't exist, create it and maybe it's parents.

In this world, every space has a uuid (which is globally unique across all orgs/projects) and a `slug` which is unique within a project (as it is today) but also encodes some hierarchical info.

This would just be a little annoying to update when the parent slug changes, because you would have to manually update all rows with a slug referencing the parent  but that's not so bad. And actually I'm not even sure that slugs ever change right now

owlas:
Let's focus on having root permissions only, but we should be careful not to over-optimise for that. We should build it and consider that we will soon have more complex permissions checks that need all the parents above the current space

joaoviana:
closing, but using this as a reference on next steps. First part is here: #14285

Issue #10696, opened 2024-07-11
I want to be able to customize what's shown in my tooltip (include extra fields)


TuringLovesDeathMetal:
### Description

Right now, what's shown in the tooltip is fixed. 

Instead, sometimes, I want to show a field in the tooltip that's in the results table, but not in the chart. For example, if I have a % conversion rate, I want to show the total number of users in that rate in the tooltip. 

### Potential solution

In the chart config, there should be an option to customize the tooltips. This could open a box where you can add HTML or Markdown? And the fields can be referenced as variables. See this video for an example of something similar in Tableau: https://www.youtube.com/watch?v=HBSjEwm-r78

jpetey75:
another request for this in customer slack: https://lightdash.slack.com/archives/C075JNX28GG/p1723043412070209

>Is there any way to get "tooltip" type behaviour on hover? For example, if we have a by-date look at subscriptions, or trials, or so on, and then if we hover over we can see (for example) conversion rate for that day. Is this something that can be done?


(wants to show metrics/data in tooltip that isn't in the chart)

jpetey75:
mentioned in community slack https://lightdash-community.slack.com/archives/C067S48BHRP/p1723212623476719

>I have using scatter plots lately and it would be great to be able to adjust what is said within the label when you add a group.
Essentially I want to use the chart to show a marker for each data point but then sub divide into groups. I want to group by market region but still show the value for the dimensions within the label.
given how a scatter plot works, it would be best to have the option to select which dimensions to show within the label - as you would with a table chart option

jpetey75:
customer request for tooltips in table charts: https://lightdash.slack.com/archives/C06B2PP2478/p1723569986110339

>Able to add description / tooltip for a Table.

PR #14292, opened 2025-04-04
feat: tooltip config

rephus:
Closes: #10696

### Description:
Tooltip "fields" are only limited to those who are part of the chat config. See code for more details. 
This should work with both pivot and no pivot fields. 

Data is raw, we can do formatting once we process data from results directly. 

Screenshot from 2025-04-07 10-17-15

### Reviewer actions


rephus:
TODO: Fix monaco editor positionnig issues 

Screenshot from 2025-04-07 10-28-02

packages/frontend/src/styles/monaco.css
```diff
@@ -2,3 +2,11 @@
     background-color: yellow;
     border-bottom: 2px solid red;
 }
+
+.tooltip-editor {
+    border: 1px solid #ccc;
+    overflow: visible;
+}
+.tooltip-editor .suggest-widget {
```
rephus, line 10: Fix for suggestions

Before

Screenshot from 2025-04-07 10-28-02

After

image

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1860,6 +1863,24 @@ const useEchartsCartesianConfig = (
                     })
                     .join('');
 
+                // At the moment, we only correctly filter fields that are
```
rephus, line 1866: Check comment, we only can rely on the data from the chart, which is already prefiltered. If we want to use other fields, we'll have to rely on resultsData, and I believe that will impact performance. 

We can release this version first, and then wait for feedback (I want to see other use cases)

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Legend/TooltipConfig.tsx
```diff
@@ -0,0 +1,124 @@
...
+    fields: string[];
+    //items: (Field | TableCalculation | CustomDimension | CompiledDimension)[];
+};
+
+const registerCustomCompletionProvider = (
+    monaco: Monaco,
+    language: string,
+    quoteChar: string,
+    fields: string[],
+    //items: (Field | TableCalculation | CustomDimension | CompiledDimension)[],
```
rephus, line 38: commented code is intended, for when we add all fields from resultsdata

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -1883,13 +1904,13 @@ const useEchartsCartesianConfig = (
                             itemsMap,
                         );
 
-                        return `${tooltipHeader}<br/><table>${tooltipRows}</table>`;
+                        return `${tooltipHeader}<br/>${tooltipHtml}<table>${tooltipRows}</table>`;
```
rephus, line 1907: we add the tooltip to the header. 

Could this generate a vulnerability issue ?

rephus: Fixed by sanitizing input using dompurify 

Screenshot from 2025-04-07 13-20-38

magnew (approved):
Very cool! Some users are going to love this. I have 2 usability things that could be done as follow-ups, but would make this feature better. 

1. Could we have an option to fully replace the tooltip content? At the moment the custom html gets mounted between the header and the rows, which will be useful, but I think some users will want to only show their own custom tooltip. We can wait and see if anyone asks for this, but I think it would be useful 😄 

2. I know we will have full docs for this, but some prompting in the UI would be helpful for users who havent seen the docs. It doens't have to say much, but something as simple as "Enter custom HTML to display in chart tooltips" would at least prompt people with the right thing to try. We can point them to docs for details, but as it is now it's a blank box. I would have guess to use MD and that is not right. 

I tested as many injection vectors as I could. DOMpurify seems to be working ✅

rephus:
> Could we have an option to fully replace the tooltip content? 
We had a chat about this https://lightdash.slack.com/archives/C02NPFHF4F6/p1743755989221129?thread_ts=1743752558.449839&cid=C02NPFHF4F6

For now we're going to keep it, but we can change it we get feedback about this.

Issue #14284, opened 2025-04-04
Invite links don't work when invites are deleted and recreated


notgiorgi:
### Description

Invite links don't work when invites are deleted and recreated

### Steps to Reproduce the Bug or Issue

1. Cannot be easily reproduced locally, use staging
2. Create an invite
3. Copy the invite link 
4. In incognito window accept the invite, fill in the user data, register but do not complete registration, leave the window open
5. On the admin side, remove the user and regenerate the invitation
6. In the same incognito window paste the link

### Cloud or self-hosting

None

PR #14293, opened 2025-04-04
fix: invite links dont work when invites are deleted and recreated

notgiorgi:
Closes:  #14284

### Description:
Demo: https://cln.sh/Mq8yQPYs

### Reviewer actions

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -40,6 +40,7 @@ import {
     createAzureAdPassportStrategy,
     createGenericOidcPassportStrategy,
     googlePassportStrategy,
+    invalidUserErrorHandler,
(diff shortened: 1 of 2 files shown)


packages/backend/src/App.ts
```diff
@@ -635,6 +632,37 @@ export default class App {
...
+                            );
+                            res.redirect(req.path);
+                            return;
+                        }
+
+                        Logger.info('Invalid user, redirecting to login');
+                        res.redirect('/login');
+                    });
+                    return;
+                }
```
almeidabbm, line 664: Couldn't this logic be handled in the `inviteLinkErrorHandler` instead of deleting it? from the order it seems like it would pick it up no?

PR #14343, opened 2025-04-09
fix: nested space promotions

notgiorgi:
Closes: N/A

### Description:

Fixes the bug with promoting charts in nested spaces leading to only the last space being created on top level

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/PromoteService/PromoteService.ts
+++ b/packages/backend/src/services/PromoteService/PromoteService.ts
@@ -860,9 +860,10 @@ export class PromoteService extends BaseService {
             .filter((change) => change.action === PromotionAction.CREATE)
             .map(async (spaceChange) => {
                 const { data: spaceChangeFromBase } = spaceChange;
-                const isNestedSpace = await this.spaceModel.isRootSpace(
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/PromoteService/PromoteService.ts
```diff
@@ -860,9 +860,10 @@ export class PromoteService extends BaseService {
             .filter((change) => change.action === PromotionAction.CREATE)
             .map(async (spaceChange) => {
                 const { data: spaceChangeFromBase } = spaceChange;
-                const isNestedSpace = await this.spaceModel.isRootSpace(
+                const isRootSpace = await this.spaceModel.isRootSpace(
```
joaoviana, line 863: 🤦 

good spot. I just tested this with and without this change. 
confirmed that with your fix the hierarchy is built correctly.

joaoviana: Tested also with one root being already in the upstream project. and saw the child and grandchild being promoted in the upstream

joaoviana (approved):

joaoviana (approved):

PR #14376, opened 2025-04-11
feat: add updated_at column to results cache and return in cache metadata

almeidabbm:
Closes: [#14375](https://github.com/lightdash/lightdash/issues/14375)

### Description:

- Adds `updated_at` column to `results_cache` table
- Returns cache metadata in `executeAsyncQuery`

https://github.com/user-attachments/assets/e4381746-28ef-4eb8-aa08-c81de300faa9

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -229,6 +229,7 @@ import {
 import {
     type ExecuteAsyncDashboardChartQueryArgs,
     type ExecuteAsyncMetricQueryArgs,
+    type ExecuteAsyncQueryReturn,
     type ExecuteAsyncSavedChartQueryArgs,
     type ExecuteAsyncUnderlyingDataQueryArgs,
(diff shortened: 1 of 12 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2559,7 +2564,11 @@ export class ProjectService extends BaseService {
 
                     return {
                         queryUuid: queryHistoryUuid,
-                    };
+                        cacheMetadata: {
```
magnew, line 2567: You think we should return expiresAt as well?

I guess I don't know exactly what we would use it for, so we can do that later if we have a use case. But it seems like it will be useful.

magnew (approved):
🎉

packages/backend/src/services/ProjectService/ProjectService.ts
almeidabbm: yep we can do that, adding

PR #14388, opened 2025-04-11
fix: react sdk jsx runtime

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/14389

### Description:

test-frontend
test-backend
test-cli

Notable changes:
- Fixes JSX runtime issues in the production environment (caused by incorrect bundling of the package).
- The SDK package now lives inside the frontend package (this reduces build complexity).
- You now only need to build the frontend as SDK once (prior to that it was 1. build frontend as lib - 2. build sdk from frontend lib).
- Separates Vite configs into two: one for the SDK and one for build/dev (like it was before the SDK was introduced).
- Adds a Next.js demo app.

### Reviewer actions


magnew (approved):
Nice work! This seems like a big improvement. I don't think I have any useful comments on code to change. I tested various paths of building things and it all seems to work ✅ 

I see we have the test app only in Next now, which is fine by me. Do we expect that to be the primary use case? I guess so if that's how Kraken is going.

Issue #14425, opened 2025-04-15
Cache explores in memory on find explore


almeidabbm:
Currently we always query the application db for explores which can become expensive.
Cache explores in memory using `node-cache` on `findExploresFromCache` calls

PR #14427, opened 2025-04-15
feat: cache explores in memory

almeidabbm:
Closes: #14425 

### Description:

- Adds in memory cache to `findExploresFromCache`

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/ProjectModel/ExploreCache.ts
+++ b/packages/backend/src/models/ProjectModel/ExploreCache.ts
@@ -1,93 +1,46 @@
 import { Explore, ExploreError } from '@lightdash/common';
 import NodeCache from 'node-cache';
 
-type CachedExplore = Explore | ExploreError;
+type CachedExplores = Record<string, Explore | ExploreError>;
 
 export class ExploreCache {
-    private readonly cache: NodeCache;
+    private readonly cache: NodeCache | undefined;
 
     constructor() {
         // Initialize cache with 30 seconds TTL
-        this.cache = new NodeCache({
-            stdTTL: 30, // time to live in seconds
-            checkperiod: 60, // cleanup interval in seconds
-        });
+        this.cache =
+            process.env.EXPERIMENTAL_CACHE === 'true'
+                ? new NodeCache({
+                      stdTTL: 30, // time to live in seconds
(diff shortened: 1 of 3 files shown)


packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -855,24 +857,86 @@ export class ProjectModel {
...
+                    return acc;
+                }, {});
+
+                // If no exploreNames provided, cache all explores for the project
+                if (!exploreNames?.length) {
+                    this.exploreCacheByName.setAllExplores(
+                        projectUuid,
+                        exploresMap,
+                    );
+                }
```
almeidabbm, line 936: Need to track `all` explores on a separate key, this is because we can't be sure that all explores are cached when we try to get them from the in memory cache.

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -855,24 +857,86 @@ export class ProjectModel {
...
+                const missingExploreNames: string[] = [];
+
+                // If exploreNames is provided, try to get each explore from cache
+                if (exploreNames?.length) {
+                    for (const name of exploreNames) {
+                        const cachedExplore =
+                            this.exploreCacheByName.getIndividualExplore(
+                                projectUuid,
+                                name,
+                            );
```
almeidabbm, line 870: Explore cache is by `name` since that's how we query them from the db.
Also, looking at `saveExploresToCache` we use a `map` to make them unique by name, not sure how much we can trust this, any clue?

packages/backend/src/models/ProjectModel/ExploreCache.ts
```diff
@@ -0,0 +1,93 @@
+import { Explore, ExploreError } from '@lightdash/common';
+import NodeCache from 'node-cache';
```
ZeRego, line 2: do we need to update package.json?

packages/backend/src/models/ProjectModel/ProjectModel.ts
ZeRego: we can trust it. explore name should be unique per project

ZeRego (approved):

Issue #14423, opened 2025-04-15
SQL Runner has a limit of 10 series per x axis value


ZeRego:
### Description

Example query: https://analytics.lightdash.cloud/projects/21eef0b9-5bae-40f3-851e-9554588e71a6/sql-runner/broken-sql-runner/edit

Problem is that we hardcoded a limit of 10 in the pivot SQL. 
https://github.com/lightdash/lightdash/blob/main/packages/backend/src/services/ProjectService/ProjectService.ts#L3798

Potential solution:
- increase limit
- Add an OR condition in case the total columns are under 100(?)

Independently of the solution, if there is a limit, we should display that information in the UI. Similar to explorer page showing the row limit was reached and results might be incomplete.

### Steps to Reproduce the Bug or Issue

1. go to sql runner
2. write sql with lots of combinations between columns
3. configure chart to pivot/group

Expected:
See bars for all values in results

Actual:
Can only see 10 bars for each x axis value.

### Cloud or self-hosting

None

PR #14430, opened 2025-04-15
feat: warn user of max column limit (sql runner)

joaoviana:
Closes: https://github.com/lightdash/lightdash/issues/14423

### Description:

Increased from `10` to `100`
Checks column count
Frontend displays the warning if that happens.

How to test:
```
SELECT * FROM "postgres"."jaffle"."customers" 
```

a chart like this

image

### Reviewer actions

### DIFF
--- a/packages/e2e/cypress/e2e/app/sqlRunner.cy.ts
+++ b/packages/e2e/cypress/e2e/app/sqlRunner.cy.ts
@@ -125,7 +125,7 @@ describe('SQL Runner (new)', () => {
         cy.get('div[role="option"]').contains('first_name').click();
         cy.get('.echarts-for-react')
             .find('text')
-            .contains('Customer id sum aaron')
+            .contains('Customer id sum frances')
             .should('be.visible');
 
         // Verify that the chart is not displayed when the configuration is incomplete


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3791,11 +3792,26 @@ export class ProjectService extends BaseService {
...
+                 group_by_query AS (${groupByQuery}), 
+                 pivot_query AS (${pivotQuery}),
+                 total_columns AS (${totalColumnsQuery}) as total_columns
+            FROM group_by_query
+            )`;
+
+            // Add a max column limit to avoid too many columns and performance issues.
+            // Warn the user if we exceed it.
+
+            pivotedSql += `\nSELECT p.*, t.total_columns FROM pivot_query p CROSS JOIN total_columns t WHERE p.${q}row_index${q} <= ${
```
joaoviana, line 3811: adds total_columns per row
I didn't want to make this a separate query so this cross join helped. lmk if you think otherwise

magnew (approved):
LGTM!

packages/e2e/cypress/e2e/app/sqlRunner.cy.ts
```diff
@@ -125,7 +125,7 @@ describe('SQL Runner (new)', () => {
         cy.get('div[role="option"]').contains('first_name').click();
         cy.get('.echarts-for-react')
             .find('text')
-            .contains('Customer id sum aaron')
+            .contains('Customer id sum frances')
```
joaoviana, line 128: now that we get more columns, the legend doesn't include aaron on the first page 
so I updated this assertion.

Compared the same chart with this PR's changes and without

with
Screenshot 2025-04-16 at 12 00 40
You can test here with querying customers and doing the config as on the screenshot

https://lightdash-pr-14434.onrender.com/projects/3675b69e-8324-4110-bdca-059031aa8da3/sql-runner

without
Screenshot 2025-04-16 at 12 00 35

joaoviana: wdyt @magnew ? It's definitely a change we need to make but could cause confusion to users. Does it make sense to have a headway announcement?

Issue #9533, opened 2024-03-26
Improvements to multi-sort behaviour -  POC 1 


PriPatel:
No multi-sort needed
- [ ] Show the default sort applied in the pill
- [ ] If the user sorts another column, the first sort is overridden. The sort pill shows the updated sort.

User wants multisort
- [ ] The user has to click into the pill to add a multisort. They have to pick which field to sort they can sort by. The dropdown shows all available columns
- [ ] The user can delete the sort completely

[Walkthrough here](https://www.figma.com/proto/73mLRDEqv8zsJeGOjTJnEu/Untitled?type=design&node-id=18-2&t=I7pPVJ17E86kaI9b-0&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=1%3A6)

PR #14437, opened 2025-04-16
feat: improve sorting behaviour ux

joaoviana:
Closes: #9533 

### Description:

When a sort is applied to from the column header, then replace the sorting
When the user wants to multisort, then they have to click on the badge to add more

Ticket above includes mockups and relevant thread is here: https://lightdash.slack.com/archives/C03MCQ08UKS/p1744369704613779?thread_ts=1744369543.999679&cid=C03MCQ08UKS

https://github.com/user-attachments/assets/b55f52db-f6ad-4596-961b-13803d94d111

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderSortMenuOptions.tsx
+++ b/packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderSortMenuOptions.tsx
@@ -30,8 +30,8 @@ const ColumnHeaderSortMenuOptions: FC<Props> = ({ item, sort }) => {
             : SortDirection.ASC
         : undefined;
 
-    const addSortField = useExplorerContext(
-        (context) => context.actions.addSortField,
+    const setSortFields = useExplorerContext(
+        (context) => context.actions.setSortFields,
     );
     const removeSortField = useExplorerContext(
         (context) => context.actions.removeSortField,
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderSortMenuOptions.tsx
```diff
@@ -56,10 +56,14 @@ const ColumnHeaderSortMenuOptions: FC<Props> = ({ item, sort }) => {
...
-                                          sortDirection === SortDirection.DESC,
-                                  })
+                                : setSortFields([
+                                      {
+                                          fieldId: itemFieldId,
+                                          descending:
+                                              sortDirection ===
+                                              SortDirection.DESC,
+                                      },
+                                  ])
```
joaoviana, line 66: clicking on a column header overrides sorts

magnew (approved):
🚀 Nice! Works great. Makes more sense than what we had before. 

One tiny thing, this popover tail + border radius looks a little funny to me. But I'm no designer, so if this is intentional, ship it
Screenshot 2025-04-17 at 10 56 08

joaoviana:
> 🚀 Nice! Works great. Makes more sense than what we had before.
> 
> One tiny thing, this popover tail + border radius looks a little funny to me. But I'm no designer, so if this is intentional, ship it Screenshot 2025-04-17 at 10 56 08

good point - resetting radius back to default

PR #14439, opened 2025-04-16
fix: use single transaction when creating an upstream space

notgiorgi:
Closes: 

### Description:

uses single transaction when creating an upstream space

### Reviewer actions

### DIFF
--- a/packages/backend/src/models/SpaceModel.ts
+++ b/packages/backend/src/models/SpaceModel.ts
@@ -351,20 +351,23 @@ export class SpaceModel {
         };
     }
 
-    async find(filters: {
-        projectUuid?: string;
-        projectUuids?: string[];
-        spaceUuid?: string;
-        spaceUuids?: string[];
-        slug?: string;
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/SpaceModel.ts
```diff
@@ -351,20 +351,23 @@
...
-    }): Promise<Omit<SpaceSummary, 'userAccess'>[]> {
+    async find(
+        filters: {
+            projectUuid?: string;
+            projectUuids?: string[];
+            spaceUuid?: string;
+            spaceUuids?: string[];
+            slug?: string;
+        },
+        { trx = this.database }: { trx?: Knex } = { trx: this.database },
```
IrakliJani, line 362: does the selects need to be inside the transaction?

notgiorgi: in the context of `createNestedSpaceFromBaseProject` it does

IrakliJani: ah I see it now

PR #14444, opened 2025-04-16
fix: respect row limit for column limit warning

joaoviana:
Closes: N/A

### Description:

After: https://github.com/lightdash/lightdash/pull/14430
I realised that if the row limit is like 5, then the columns retrieved are small, but the query was still counting the columns regardless of the row limit

To test

have a chart like this on customers

image

play with the row limit (5 will generate less columns)
Since it generates less columns, it should also know how many columns are for that row limit (and then we can check how many columns are actually in the query before showing the warning to the user)

### Reviewer actions


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3866,19 +3866,25 @@ export class ProjectService extends BaseService {
...
+                    SELECT *
+                    FROM group_by_query
+                    LIMIT ${limit ?? 500}
+                )`;
             let pivotedSql = `
             WITH original_query AS (${userSql}), 
                  group_by_query AS (${groupByQuery}), 
                  pivot_query AS (${pivotQuery}),
-                 total_columns AS (${totalColumnsQuery}) as total_columns
-            FROM group_by_query
```
joaoviana, line 3880: unnecessary

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3866,19 +3866,24 @@ export class ProjectService extends BaseService {
...
+                    SELECT * FROM pivot_query WHERE "row_index" <= ${
+                        limit ?? 500
+                    }
+                 ),
+                 total_columns AS (
+                    SELECT (COUNT(DISTINCT ${groupByColumns
+                        .map((col) => `filtered_rows.${q}${col.reference}${q}`)
+                        .join(', ')}) * ${
+                valuesColumns?.length || 1
+            }) as total_columns FROM filtered_rows
```
joaoviana, line 3884: counts columns based on the filtered rows

magnew (approved):
Works as advertised ✅

Issue #14398, opened 2025-04-14
allow customizing CSV export row limit for embedded dashboards


IrakliJani:
### Description

When exporting a CSV from a chart on an embedded dashboard, only 500 rows are exported even though the full dataset is larger and under the 100,000 cell limit. Users can't control or change this limit.

- This limit seems hardcoded and not configurable per project or dashboard
- Lightdash UI allows setting a row limit, but embedded dashboards don't expose this
- Users with access to the full chart expect to download the full data they see

### What problem does this solve?

Makes embedded exports more useful by matching user expectations and enabling full access to the data behind a chart.

ℹ  check if the limit was intentional to prevent end users of embedded dashboards from downloading more data than developers intended to expose. We'll need to assess if allowing customization here aligns with our security and access control goals before proceeding.

magnew:
There is now a PR WIP that handles a lot of this: https://github.com/lightdash/lightdash/pull/14458

There is still one major chunk of work to do to finish it. Details in the PR.

ZeRego:
This work is unblocked by https://github.com/lightdash/lightdash/issues/15017

PR #14458, opened 2025-04-17
feat: enable downloading all rows in embedded charts

magnew:
Closes: #14398 

### Description:

WIP: enable exporting all rows in embedded dashbaords.

Where this is an what it still needs:
- The FE is largely done. It includes the download menu in embed dialogs and an option on the embed page.
- In the backend, the EmbedService is set up to *schedule* jobs. At the moment it is doing it in the normal scheduler queue. It could be that it should be in a different embed queue, I couldn't decide. 
- The logic to return the CSV link from the BE is mostly there, although it's hard to test without the workers done. There may need to be a little more here. 
- **The main thing remaining** is having workers actually process the CSV jobs. I went back and forth between trying to get the regular scheduler workers to process the jobs and making an EE specific handler. Making something EE specific would be a cleaner split, but it will require a lot of code duplication and probably a new table. I could use advice from anyone who has an opinion here 😄 

### Reviewer actions

### DIFF
--- a/packages/backend/src/ee/analytics/index.ts
+++ b/packages/backend/src/ee/analytics/index.ts
@@ -14,6 +14,7 @@ export type EmbedDashboardViewed = BaseTrack & {
         context: 'preview' | 'production';
         dashboardFiltersInteractivity?: DashboardFilterInteractivityOptions;
         canExportCsv?: boolean;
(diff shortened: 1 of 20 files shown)


magnew:
Closing for now. Will re-open if needed.

Issue #14468, opened 2025-04-18
Different series colors in dashboard charts


almeidabbm:
### Description

When I have 2 charts on a dashboard containing the same series but in a different order, these will have different colors for same series.

### Steps to Reproduce the Bug or Issue

1. Create dashboard
2. create 2 charts with same series but different series order

![Image](https://github.com/user-attachments/assets/1708230c-9982-489f-b85b-1c9578dc9e21)

### Cloud or self-hosting

None

PR #13923, opened 2025-03-04
fix:  disable calculate series color

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13831
Closes: https://github.com/lightdash/lightdash/issues/13464

### Description:

I was able to replicate by removing the color on the groups (after looking at the chartconfig form the customer) . I even copied their org palette

THe colors were a bit random, when `calculateSeriesColorAssignment` was enabled. 

The good colors: 

image

Random colors: 

Screenshot from 2025-03-04 16-27-43

Sometimes even the color in series were reversed: 

Screenshot from 2025-03-04 16-38-45

### Reviewer actions

- [ ] I have manually tested the changes in the preview environment
- [ ] I have reviewed the code
- [ ] I understand that "request changes" will block this PR from merging

PR #14469, opened 2025-04-18
fix: sort series for fallback colors

almeidabbm:
Closes: #14468

This happens when calculated series colors is disabled. Calculated series colors introduces other issues related to chart render order, but fixes this.

This approach will sort the `fallback` colors so that colors always match if 2 charts contain the same series and are using fallback colors

Related PR: https://github.com/lightdash/lightdash/pull/13923 

### Description:

**Before**
image

**After**
image

### Reviewer actions


rephus:
This fix is compatible with the "invisible " color bugfix  :+1: 

image

rephus (approved):
Works great. I was afraid this could cause any other issues, but everything seems to work as expected. 

image

Issue #14659, opened 2025-05-05
Fix dashboard charts fetching twice when clicking on refresh button


almeidabbm:
### Description

Related PRs:
- https://github.com/lightdash/lightdash/pull/14658
- https://github.com/lightdash/lightdash/pull/14613

We invalidate the full query key and we have `invalidateCache` as part of the query key, so it will trigger both `useQuery` (`true` and `false` invalidate cache) when clicking the refresh button.

We cannot remove `invalidateCache` from the queryKey because otherwise it will never send the `true` to invalidate the cache

### Steps to Reproduce the Bug or Issue

1. Open dashboard with cache enabled
2. Click the refresh button up top
3. Check the network tab, you should see 2 requests for each chart

### Cloud or self-hosting

None

PR #14661, opened 2025-05-05
fix: dashboard double fetch on first refresh

almeidabbm:
Closes: [#14659](https://github.com/lightdash/lightdash/issues/14659)

### Description:

- We were updating `invalidateQueries` state + invalidating manually the query. This caused double fetch on first refresh because `invalidateQueries` went from `false` to `true` which invalidates the query itself, but then it was also getting invalidated by the refresh button.

- This PR fixes this by splitting queries into invalidate between related queries (which always get fetched and are not dependent on `invalidateCache`) and results queries (dependent on `invalidateCache`). Then we only manually invalidate the results manually when `invalidateCache` is true and the button is clicked, otherwise the `invalidateCache` state change invalidates the query by itself

**Before**

https://github.com/user-attachments/assets/60b684d1-c168-4714-993a-1c6e34d3b2c6

**After**

https://github.com/user-attachments/assets/f3371697-a31e-48ff-9def-75da8a1cd6b5

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Dashboard/DashboardRefreshButton.tsx
+++ b/packages/frontend/src/components/common/Dashboard/DashboardRefreshButton.tsx
@@ -56,20 +56,36 @@ export const DashboardRefreshButton: FC<DashboardRefreshButtonProps> = memo(
             [refreshInterval],
         );
 
-        const { isFetching, invalidateDashboardRelatedQueries } =
-            useDashboardRefresh();
+        const {
+            isFetching,
+            invalidateDashboardRelatedQueries,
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#14661** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/14661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/14661?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #14672, opened 2025-05-06
If there is an error in a tile on a dashboard, it just loads infinitely, so I can't edit the dashboard tile


TuringLovesDeathMetal:
### Description

There is a bug that's appeared recently where if there is an error in a tile on a dahsboard (e.g. incorrect field name or field name doesn't exist), then the tile just loads infinitely and there's no way to edit the tile. 

Instead, the tile should load, show an error, and the user should be able to edit the tile. 

This also appears to be affecting charts in `view` mode! (see example here: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1746027698103129)

### Steps to Reproduce the Bug or Issue

See an example here: https://analytics.lightdash.cloud/projects/21eef0b9-5bae-40f3-851e-9554588e71a6/dashboards/60805581-99ee-4137-b3c1-aab3c89868e7/view

### Cloud or self-hosting

None

PR #14691, opened 2025-05-06
fix: show error for visualizations on paginated query error

magnew:
Closes: #14672 

### Description:

This fixes an issue where charts with errors in the query never got out of the loading state. We weren't handling the error status on paginated queries completely. This was especially problematic on dashbaords, because the edit controls were hidden.

To repro the issue, you can add a bad table calculation on the Explore page. You'll see it loads forever:
Kapture 2025-05-06 at 18 49 48

If you save that chart it will also load forever. 

Now, it shows an error:
after

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/useQueryResults.ts
+++ b/packages/frontend/src/hooks/useQueryResults.ts
@@ -180,6 +180,7 @@ export type InfiniteQueryResults = Partial<
     fetchAll: boolean;
     hasFetchedAllRows: boolean;
     totalClientFetchTimeMs: number | undefined;
+    error: ApiError | null;
 };
 
 // This hook lazy load results has they are needed in the UI
@@ -248,6 +249,11 @@ export const useInfiniteQueryResults = (
         hasFetchedAllRows,
     ]);
 
+    const queryClient = useQueryClient();
+
+    // Initial backoff time in ms
+    const backoffRef = useRef(250);
+
     const nextPage = useQuery<
         ApiGetAsyncQueryResults & { clientFetchTimeMs: number },
         ApiError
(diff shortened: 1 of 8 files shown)


packages/frontend/src/hooks/useQueryResults.ts
```diff
@@ -238,6 +239,7 @@ export const useInfiniteQueryResults = (
             !!projectUuid &&
             !!queryUuid &&
             (isFetchingPage || (fetchAll && !hasFetchedAllRows))
+            // thing
```
ZeRego, line 242: Missing something here?

magnew: Will remove

packages/frontend/src/hooks/useQueryResults.ts
```diff
@@ -262,6 +268,66 @@ export const useInfiniteQueryResults = (
...
+                    break;
+                }
+                case QueryHistoryStatus.READY: {
+                    backoffRef.current = 250;
+                    setFetchedPages((prevState) => [
+                        ...prevState,
+                        {
+                            ...results,
+                            clientFetchTimeMs: performance.now() - startTime,
+                        },
```
almeidabbm, line 323: Can we set this to a variable and both call `setFetchedPages` here + return?

packages/frontend/src/hooks/useQueryResults.ts
```diff
@@ -262,6 +268,66 @@ export const useInfiniteQueryResults = (
...
+                    ]);
+                    break;
+                }
+                default:
+                    return assertUnreachable(status, 'Unknown query status');
+            }
+
             return {
```
almeidabbm, line 331: If above is done, remove this, it duplicates and calculates `clientFetchTimeMs` twice

almeidabbm (changes requested):

almeidabbm (approved):

Issue #14407, opened 2025-04-14
SQL runner queries always with 500 limit


ZeRego:
### Description

SQL runner queries are always limited to 500 rows. 

There is logic to apply a limit to SQL runner queries to prevent long running queries.
The problem is the FE is always passing the default row limit instead of max limit.

https://github.com/lightdash/lightdash/blob/5474e4b7caf2bf138d06669172f24b41a4bb317d/packages/backend/src/services/ProjectService/ProjectService.ts#L3665

### Steps to Reproduce the Bug or Issue

1. go to sql runner page
2. run `SELECT * FROM "example" limit 1000`
3. See warehouse query history

Expected:

SELECT * FROM "example" limit 1000

Actual: 

SELECT * FROM "example" limit 500

### Cloud or self-hosting

None

magnew:
I think this was partially intentional, but the current behavior is both confusing, and not quite right. There are two parts to this:

1. In the SQL runner SQL tab (not the chart tab), setting the limit in the UI was removed. I think the intention was that is would always run with 500 as a 'preview'. At the time we didn't have a query limit Env variable, now we do. We should set this limit to that var. 
2. In the Charts tab, we are just not sending the limit in the UI. We should. 

In both cases it's still a little confusing what happens the limits set in SQL, but these changes will still be less confusing.

PR #14773, opened 2025-05-12
fix: SQL Runner always sends limit 500

magnew:
Closes: #14407 

### Description:

There are still some limit-related UX questions in the SQL runner, but this is an improvement that specifically solves the stated issue that the SQL runner ALWAYS sends limit=500. 

On the initial SQL page (where limit can't be changed), we now use the `LIGHTDASH_QUERY_DEFAULT_LIMIT` for the query limit and the default results download. 
Screenshot 2025-05-12 at 15 15 15

On the chart page, we now respect the user-input limit
Screenshot 2025-05-12 at 15 16 34

What this *doesn't* address:
- Limits entered in the SQL are still overridden if they are higher than the limit

To test:
- Set `LIGHTDASH_QUERY_DEFAULT_LIMIT` to something other than 500
- Check that the initial query page uses that limit
- Set the limit manually on the chart page and check that that limit is used. 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
+++ b/packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
@@ -54,6 +54,7 @@ import { Table } from '../../../components/DataViz/visualizations/Table';
 import RunSqlQueryButton from '../../../components/SqlRunner/RunSqlQueryButton';
 import { useOrganization } from '../../../hooks/organization/useOrganization';
 import useToaster from '../../../hooks/toaster/useToaster';
+import useApp from '../../../providers/App/useApp';
 import { DEFAULT_SQL_LIMIT } from '../constants';
 import { useAppDispatch, useAppSelector } from '../store/hooks';
 import {
@@ -100,6 +101,7 @@ export const ContentPanel: FC = () => {
 
     // Get organization colors to generate chart specs with a color palette defined by the organization
     const { data: organization } = useOrganization();
+    const { health } = useApp();
 
     const { showToastError } = useToaster();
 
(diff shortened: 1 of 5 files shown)


packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -258,6 +260,20 @@ export const ContentPanel: FC = () => {
         }
     }, [queryResults, panelSizes]);
 
+    useEffect(() => {
+        if (!limit) {
+            dispatch(
+                setSqlLimit(
+                    health.data?.query.defaultLimit ?? DEFAULT_SQL_LIMIT,
```
IrakliJani, line 267: I always wondered if sql limit and query limit variables were synced? we should completely remove `DEFAULT_SQL_LIMIT` at some point and only read it from `defaultLimit`

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -258,6 +260,20 @@ export const ContentPanel: FC = () => {
...
+                ),
+            );
+        }
+    }, [health.data?.query.defaultLimit, dispatch, limit]);
+
+    const defaultQueryLimit = useMemo(() => {
```
IrakliJani, line 273: you can reuse this variable to set it in `setSqlLimit` above. will avoid some react hook checks

IrakliJani (approved):
good stuff!

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
magnew: Yea, it would be better. At the moment we need some fallback on page load, then this useEffect runs. To do it right we should initialise the store differently and we won't need this. 

I'm a little waiting to see where we go with SQL runner before doing bigger work to change the patterns.

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
magnew: Good call! Updated

PR #14787, opened 2025-05-13
fix: Fix Trino client to return a more descriptive error message

hhhonzik:
### Description:

In case of error `trino-client` do not return an `Error` instance, but just a typed `Object`. Common method `getErrorMessage` was unable to parse such object. 

This PR updates Trino warehouse to return Error object in case of an error.

Before:
Screenshot 2025-05-13 at 14 47 11

After:
Screenshot 2025-05-13 at 14 44 00

### Reviewer actions


ZeRego (approved):
thanks for the contribution 🙌

PR #14800, opened 2025-05-14
fix: handle empty rows in CSV export

notgiorgi:
Closes: #ISSUE_NUMBER

### Description:
Fix CSV export when rows array contains null values by adding a null check before accessing the first row. This prevents the application from crashing when trying to export empty data sets.

The change adds a check using Lodash's `isNil` function to safely handle cases where the first row might be null or undefined, returning an empty array for field IDs in such cases.

### Reviewer actions

### DIFF
--- a/packages/backend/src/services/CsvService/CsvService.ts
+++ b/packages/backend/src/services/CsvService/CsvService.ts
@@ -49,6 +49,7 @@ import { stringify } from 'csv-stringify';
 import * as fs from 'fs';
 import * as fsPromise from 'fs/promises';
 
+import isNil from 'lodash/isNil';
 import moment, { MomentInput } from 'moment';
 import { nanoid } from 'nanoid';
 import { pipeline, Readable, Transform, TransformCallback } from 'stream';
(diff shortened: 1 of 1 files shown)


notgiorgi:
* **#14800** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/14800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/14800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/CsvService/CsvService.ts
```diff
@@ -283,9 +284,13 @@ export class CsvService extends BaseService {
         const fileId = CsvService.generateFileId(fileName, truncated);
         const writeStream = fs.createWriteStream(`/tmp/${fileId}`);
 
-        const sortedFieldIds = Object.keys(rows[0])
-            .filter((id) => selectedFieldIds.includes(id))
-            .sort((a, b) => columnOrder.indexOf(a) - columnOrder.indexOf(b));
+        const sortedFieldIds = isNil(rows[0])
```
rephus, line 287: just curious, what is the advantage of using isNil ? (Checks if value is null or undefined)
Is this not the same as doing `rows[0] ? []: ... `

rephus (approved):

packages/backend/src/services/CsvService/CsvService.ts
notgiorgi: You're right! in this case there's not much difference

I use `isNil` or `!= null` to check for `null | undefined` exclusively, `rows [0] ? ...` would also return false for empty string, zero and `false`

notgiorgi:
### Merge activity

* **May 14, 5:45 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/14800).

PR #14808, opened 2025-05-14
fix: dashboard tile infinite load on error

magnew:
### Description:

**The problem**: dashboard tiles weren't handling errors from `/saved/:savedQueryID` and would infinitely load on error. This came up for users accessing dashboards with charts that they didn't have access to (because of space permissions).

**Solution**: the hook now handles errors and you see them reported
fix

To test:
- Make a dashboard in a shared space with a chart in a private space
- Access that dashboard as another user
- You should see the space access error 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/dashboard/useDashboardChartReadyQuery.ts
+++ b/packages/frontend/src/hooks/dashboard/useDashboardChartReadyQuery.ts
@@ -10,7 +10,7 @@ import {
     type SavedChart,
 } from '@lightdash/common';
 import { useQuery } from '@tanstack/react-query';
-import { useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { lightdashApi } from '../../api';
 import useDashboardContext from '../../providers/Dashboard/useDashboardContext';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/dashboard/useDashboardChartReadyQuery.ts
```diff
@@ -61,19 +63,27 @@ export const useDashboardChartReadyQuery = (
...
+    const chartQuery = useSavedQuery({
         id: chartUuid ?? undefined,
     });
 
-    const { data: explore } = useExplore(chart?.metricQuery?.exploreName);
+    useEffect(() => {
+        if (chartQuery.error) {
+            setError(chartQuery.error);
+        }
+    }, [chartQuery.error]);
```
notgiorgi, line 74: I don't think you need this `useEffect` or `useState` directly use the error from `chartQuery.error`?

magnew: My bad, you're right! I iterated on this a couple ways and never went back and simplified. Thanks for the push.

Updated

notgiorgi (approved):

Issue #14784, opened 2025-05-13
Navigation is enabled when selecting items


tatianainama:
### Description

When users are selecting items, navigation should be disabled to prevent accidental redirection to another page. This prevents user frustration that occurs when trying to select multiple items and accidentally missing the checkbox, which currently triggers navigation.

https://github.com/user-attachments/assets/b17e48a5-a07c-46a7-9148-f35cbb9f838e

### Steps to Reproduce the Bug or Issue

1. Select an item by using the checkbox
2. Click on another item outside the select checkbox
3. I'm redirected to the item page

### Cloud or self-hosting

None

tatianainama:
With the fix, we will also update the behavior in selection mode. If a bulk selection starts, clicking a row will toggle the selection. Navigation is still available if the name of the item is clicked.

PR #14817, opened 2025-05-14
fix: Navigation enabled when selecting rows

tatianainama:
Closes: #14784 

### Description:
- `InfiniteResourceTable`: Disable item navigation when selecting rows in table
  - Navigation is still enabled in the item title

### Reviewer actions


notgiorgi:
UX suggestion: If we are disabling navigation when selected, the click on item should unselect it probably. Check similar flow Gmail app for example

notgiorgi (approved):
Tested, LGTM, one UX comment

Issue #14718, opened 2025-05-08
Pivoted table shows loading state twice


magnew:
### Description

Pivoted table shows loading state twice in Explorer. 

Link in analytics: 
https://analytics.lightdash.cloud/share/ps5jhtUXfIlK23z6AYh36

### Steps to Reproduce the Bug or Issue

See link

### Cloud or self-hosting

None

PR #14819, opened 2025-05-14
fix: pivot table loading state in explorer

pimmey:
…ueryService

Closes: #14718 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
+++ b/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
@@ -302,11 +302,11 @@ const useTableConfig = (
             return;
         }
 
-        setPivotTableData({
+        setPivotTableData((prevState) => ({
+            ...prevState,
             loading: true,
-            data: undefined,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -302,11 +302,11 @@ const useTableConfig = (
             return;
         }
 
-        setPivotTableData({
+        setPivotTableData((prevState) => ({
+            ...prevState,
```
pimmey, line 306: Not sure about this one. It works because of [how we render the PivotTable](https://github.com/lightdash/lightdash/blob/issues/14718/packages/frontend/src/components/SimpleTable/index.tsx#L140), if we data is set and even if loading is true.

Let me know, if this is a lucky fix or if we should take a deeper look into it :)

almeidabbm (approved):

almeidabbm:
@all-contributors please add @pimmey for bug

Issue #14857, opened 2025-05-16
Missing check indicator when moving Spaces into root level


tatianainama:
### Description

If I select a space in where to created my new nested space, the space has a check mark beside it. If I click the spaces space, there is no check mark 

| Space | Root |
|--------|--------|
| ![Image](https://github.com/user-attachments/assets/d92f3726-81e7-4e28-97c7-3f828b3bf9ff) | ![Image](https://github.com/user-attachments/assets/d511c430-7036-419c-8052-ee7ee37f8e6d) | 

### Steps to Reproduce the Bug or Issue

1. Select a Space to move
2. Click any Space from the tree selection. A check indicates the selected Space
3. Click the root Space from the tree selection. The check is missing 

### Cloud or self-hosting

None

PR #14868, opened 2025-05-19
fix: Missing check indicator when moving Spaces into root level 

tatianainama:
Closes: #14857

### Description:

Display check ✓ icon when selecting root + some other style improvements:
- Unify tree box padding
- Balance paddings and margins of selected elements in tree (see screenshots comparing hover & selected states)
- Always destination space, even when moving into Root Space

| Before | After |
|--------|--------|
| image | image | 

### Reviewer actions

### DIFF
--- a/packages/frontend/src/components/common/Tree/TreeItem.tsx
+++ b/packages/frontend/src/components/common/Tree/TreeItem.tsx
@@ -58,12 +58,7 @@ const TreeItem: React.FC<Props> = ({
             w="100%"
             gap={rem(4)}
             h={rem(32)}
-            // This component isn't optimized for the top-level root item,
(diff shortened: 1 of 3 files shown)


joaoviana:
Should we make sure that the "summary" message is always visible? 
image

Or should this summary message be smaller and somewhere else?

packages/frontend/src/components/common/Tree/TreeItem.tsx
```diff
@@ -109,7 +104,7 @@ const TreeItem: React.FC<Props> = ({
                 {stringLabel}
             </Highlight>
 
-            {!isRoot && selected && (
+            {selected && (
```
joaoviana, line 107: image

Should we highlight the row as well when 'All spaces' is selected?

Issue #14969, opened 2025-05-23
Create dashboard filter from SQL column


ZeRego:
### What problem does this solve?

Add dashboard filter when there are only SQL chart tiles

PR #14983, opened 2025-05-23
chore: update SQL chart column metadata to include type information

ZeRego:
Related to: https://github.com/lightdash/lightdash/issues/14969

### Description:
This PR updates the SQL chart tile metadata to include column type information. Previously, only column references were stored, but now we're storing both the reference and type for each column. This change ensures that dashboard filters have access to the correct column type information when configuring filters.

Key changes:
- Updated `SqlChartTileMetadata` type to store `ResultColumn` objects (with reference and type) instead of just strings
- Modified `DashboardSqlChartTile` to extract both reference and type from query dimensions
- Updated filter configuration components to properly access column references from the new metadata structure

### Reviewer actions

### DIFF
--- a/packages/common/src/index.test.ts
+++ b/packages/common/src/index.test.ts
@@ -1,7 +1,7 @@
 import moment from 'moment';
 import {
     getDateGroupLabel,
-    getFilterRuleWithDefaultValue,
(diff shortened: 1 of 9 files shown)


pimmey (approved):
Let's get them filters going 🚀

PR #15000, opened 2025-05-26
fix: filter dashboard versions by dashboard ID

rephus:
Closes: #15001 

### Description:
Fixed a bug in the SQL query that filters dashboard versions. The previous query was selecting the latest dashboard version across all dashboards, rather than the latest version for each specific dashboard. This change ensures that we correctly filter by the last version of each dashboard by adding a where clause to match the dashboard_id.

before:

image

after: 

image


almeidabbm (approved):
Before:
image

After:
image

PR #15008, opened 2025-05-27
feat: copy table configuration from prod to preview projects

rephus:
Closes: #ISSUE_NUMBER

### Description:
Added the ability to copy table configuration from production projects to preview projects. This feature allows users to specify a `--table-configuration prod` flag when creating preview projects, which will copy the table configuration from the upstream production project. This ensures that preview projects maintain the same table settings as their production counterparts.

The implementation adds a new enum `CreateProjectTableConfiguration` with options `PROD` and `ALL`, and extends the project creation process to handle copying table configurations when specified.

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -26,6 +26,7 @@ import {
     CreateJob,
     CreateProject,
     CreateProjectMember,
+    CreateProjectTableConfiguration,
     CreateSnowflakeCredentials,
     CreateVirtualViewPayload,
     CreateWarehouseCredentials,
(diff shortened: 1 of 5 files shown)


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -845,6 +846,32 @@ export class ProjectService extends BaseService {
             }
         }
 
+        if (
+            data.tableConfiguration === CreateProjectTableConfiguration.PROD &&
```
rephus, line 850: similar structure : https://github.com/lightdash/lightdash/pull/15008/files#diff-dba02b88a0bd4a4fdcf7ee49daaaa481189bc9ad76a683c357d0f59e568370a5R830

PR #15023, opened 2025-05-27
feat: deprecate RESULTS_CACHE variables

magnew:
Closes:


almeidabbm (approved):
🚀 LGTM

Tested docker default with another env variable reference to make sure it works:
image
image
image

PR #15026, opened 2025-05-27
feat: add BigQuery SSO controller for database access

rephus:
Closes: #XXXX

### Description:
Added BigQuery SSO support to allow users to connect to BigQuery using their Google account. This PR introduces a new controller `BigquerySSOController` with an endpoint to fetch BigQuery databases using the user's Google credentials. The implementation leverages the Google OAuth refresh token to authenticate with BigQuery and retrieve available datasets.

The PR also adds a "Connect with Google" button to the BigQuery connection form in the frontend, laying the groundwork for the SSO flow.


rephus:
* **#15026** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15026?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15026?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
unrelated api test error

image

rephus:
Fixed this e2e test. Other don't seem relevant

image

packages/frontend/src/components/ProjectConnection/WarehouseForms/BigQueryForm.tsx
```diff
@@ -51,20 +64,104 @@ export const BigQuerySchemaInput: FC<{
...
+                variant="default"
+                color="gray"
+                disabled={disabled}
+                leftIcon={
+                    <Image
+                        width={16}
+                        src={
+                            'data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTgiIGhlaWdodD0iMTgiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PGcgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIj48cGF0aCBkPSJNMTcuNiA5LjJsLS4xLTEuOEg5djMuNGg0LjhDMTMuNiAxMiAxMyAxMyAxMiAxMy42djIuMmgzYTguOCA4LjggMCAwIDAgMi42LTYuNnoiIGZpbGw9IiM0Mjg1RjQiIGZpbGwtcnVsZT0ibm9uemVybyIvPjxwYXRoIGQ9Ik05IDE4YzIuNCAwIDQuNS0uOCA2LTIuMmwtMy0yLjJhNS40IDUuNCAwIDAgMS04LTIuOUgxVjEzYTkgOSAwIDAgMCA4IDV6IiBmaWxsPSIjMzRBODUzIiBmaWxsLXJ1bGU9Im5vbnplcm8iLz48cGF0aCBkPSJNNCAxMC43YTUuNCA1LjQgMCAwIDEgMC0zLjRWNUgxYTkgOSAwIDAgMCAwIDhsMy0yLjN6IiBmaWxsPSIjRkJCQzA1IiBmaWxsLXJ1bGU9Im5vbnplcm8iLz48cGF0aCBkPSJNOSAzLjZjMS4zIDAgMi41LjQgMy40IDEuM0wxNSAyLjNBOSA5IDAgMCAwIDEgNWwzIDIuNGE1LjQgNS40IDAgMCAxIDUtMy43eiIgZmlsbD0iI0VBNDMzNSIgZmlsbC1ydWxlPSJub256ZXJvIi8+PHBhdGggZD0iTTAgMGgxOHYxOEgweiIvPjwvZz48L3N2Zz4='
+                        }
+                        alt={`Google logo}`}
```
notgiorgi:
```suggestion
                        alt="Google logo"
```

theres an extra brace

packages/frontend/src/components/ProjectConnection/WarehouseForms/BigQueryForm.tsx
```diff
@@ -51,20 +64,104 @@ export const BigQuerySchemaInput: FC<{
...
+
+    useEffect(() => {
+        // Reset authenticationType when savedConnection changes, trigger only when savedConnection is loaded
+        if (savedConnection)
+            form.setFieldValue(
+                'warehouse.authenticationType',
+                savedConnection?.authenticationType ||
+                    BigqueryAuthenticationType.PRIVATE_KEY,
+            );
+    }, [savedConnection]); // eslint-disable-line react-hooks/exhaustive-deps
```
notgiorgi: I have a small proposal for a better way to solve this [optional suggestion].

We seem to run into this pattern often:

1. We introduce a new field in the warehouse connection config.
2. We want to default this new field (e.g., `authenticationType`, with a new value like SSO).
3. We want to ensure that all existing projects (which don’t have this field in the DB) default to the old value.
4. Our global default (in front-end form) is set to the new value, but we have to implement special handling to maintain backward compatibility.

Instead, what if we:

1. Modify the backend `/projects/{projectUuid}` route to handle this logic directly.
   - For example, for legacy BigQuery projects, use [this method](https://github.com/lightdash/lightdash/blob/1ca88bca20c07dd0f79333c6efe6ab75723b5eb2/packages/backend/src/models/ProjectModel/ProjectModel.ts#L762-L806) to apply `authenticationType === undefined ? PRIVATE_KEY : SSO`.
2. Let the frontend receive fully resolved config values, so we don’t need complex logic to infer defaults client-side.
3. When the form is submitted, the backend receives data in the correct shape—no need for fallback/backward-compat logic.

notgiorgi (commented):
👏 took a look at the code, going to test it now

packages/frontend/src/components/ProjectConnection/WarehouseForms/BigQueryForm.tsx
rephus: yeah, that's not a bad idea to improve consistency. 

Would that solve this use case though ? 

- when form loads, we load a mantine `select` with a defaultValue `sso` 
- we get saved chart from hook 
- when savedchart is loaded, defaultValue is useless now, we can only change it using `form.setFieldValue` (if exists) 
- AND we can't wait for savedchat to load ( like `if (!savedchart) return null`) because for new projects, the hook will always return undefined

rephus: Your solution will sitll improve consistency and remove some other checks, so we can do it anyway

notgiorgi: I assume you meant `savedConnection`.

- The form is only rendered if `project` is already loaded: https://github.com/lightdash/lightdash/blob/1aae6809a7a43155d8b20f7ccd3e564083f0213c/packages/frontend/src/components/ProjectConnection/UpdateProjectConnection.tsx#L171-L181

- We then do `{ ...defaultValues, ...project.warehouseConnection }`: https://github.com/lightdash/lightdash/blob/1aae6809a7a43155d8b20f7ccd3e564083f0213c/packages/frontend/src/components/ProjectConnection/UpdateProjectConnection.tsx#L57-L76

- So default form value `authenticationType: SSO`  is only going to be set, if `project.authenticationType` is set to `undefined`; if we do the proposed back-end change, that will basically never happen

Maybe I'm missing something but I'm not sure when does this edge case happen?

rephus: I added the default, and simplified the default logic in the FE

notgiorgi: ✨ Will re-review and test shortly

notgiorgi (approved):
Tested, LGTM

Issue #15031, opened 2025-05-28
Dashboard charts failing with "Columns not found"


almeidabbm:
Some dashboard charts fail with the following error:

![Image](https://github.com/user-attachments/assets/22d89046-429f-4c52-a060-2938c89578a0)

joaoviana:
https://lightdash.sentry.io/issues/6620409161/?environment=cloud_beta&query=id%3Ad41986ff43cf45e284c8dc21ac03b9cb&referrer=issue-stream&sort=date&stream_index=0

Issue #14728, opened 2025-05-08
Rename `results_cache` and ResultsCacheModel


almeidabbm:
- Rename `results_cache` table to `query_results_file`
- Remove `expires_at` column, as this can be calculated in the cache service and is not related to paginated results
- Rename `cache_key` to `file_name`?
- Rename `ResultsCacheModel` to `QueryResultsFileModel`

ZeRego:
Also take into account this slack thread: https://lightdash.slack.com/archives/C08N5JUL57V/p1746799759056689

PR #15034, opened 2025-05-28
fix: race condition for results_cache entry pending while query history ready

almeidabbm:
Closes: #15031
Relates to: https://github.com/lightdash/lightdash/issues/14728

- Fixes race condition by overriding the `query_history` status when getting results page if the `results_cache` row is `PENDING`

**Before**
image

**After**
Log
image
Dashboards
image

**Steps to reproduce**
1. Use a warehouse that is not local - longer execution times help reproducing this race condition
2. Use a chart that has more than 500 results - more pages help reproducing this race condition - I used events, with `event_id` and limit of 5000
3. Create dashboard and duplicate the tile a couple times - duplicating instead of adding the same chart will lead to more requests of the same results
4. Open the dashboard twice in separate browser windows at the same time - found it was easier to work with windows rather than tabs because of how chrome fires the requests

test-frontend


almeidabbm:
* **#15034** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15034?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15034?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #15059, opened 2025-05-28
feat: migrate results file information to query history table

ZeRego:
Closes: #14728

### Description:

There should be no impact on UI. Tested with explore and sql charts and with and without results caching enabled.

Key changes:

- Simplify state management
  - `cache_key` is now set when query is created 
  - Results metadata is now set when the query state is READY. This simplifies the logic as we don't need to retroactively update the query status based on the cache state.
- Separate cache key from results file
  - fixes issue where queries with the same cache_key tried to update the same result row. We can have multiple query with the same cache key being triggered at the same and won't impact each other.

### AI Description:
This PR refactors the results caching system by moving results metadata from the `results_cache` table directly into the `query_history` table. This simplifies the data model and reduces the need for joins between tables.

Key changes:
- Added new columns to `query_history` table to store results metadata (file name, timestamps, columns)
- Made `cache_key` non-nullable in `query_history` table
- Removed the `results_cache` table and its model
- Updated the `S3ResultsFileStorageClient` with a utility method to sanitize file extensions
- Refactored the `AsyncQueryService` to work with the new data model
- Updated the `CommercialCacheService` to use query history for cache lookups

This change improves the reliability of the caching system by eliminating race conditions between query history and cache status updates.


ZeRego:
* **#15059** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15059?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15059?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #15024, opened 2025-05-27
dashboard refresh button doesn't invalidate SQL charts


ZeRego:
### Description

dashboard refresh button doesn't invalidate SQL charts

### Steps to Reproduce the Bug or Issue

1. go to dashboard
2. add SQL chart
3. click refresh button

Expect:
SQL chart to refresh and invalidate cache

Current:
Doesn't refresh or invalidate cache 

### Cloud or self-hosting

None

PR #15060, opened 2025-05-28
fix: refresh SQL charts on dashboards

magnew:
Closes: #15024 

### Description:

Refetches SQL charts on dashboards. This was pretty tricky to get right because it ran contrary to a fix we made a couple weeks ago not to *double* fetch charts. What was done and why:

- Add `savedSqlChartResults` to the query key for refetching results. Seemed like it should be enough. But then on the **first** refresh, the sql charts weren't getting fetched because `invalidateCache` in the refresh button wasn't true yet
- So, remove the check for invalidate cache in the refresh button. But then the Explore charts were double-fetching. 
- Remove invalidateCache from the explore charts query key. Explore charts still work because we now have `dashboard_chart_ready_query` as part of their key. 
- We **couldn't** just add `invalidateCache` to the SQL charts query key because that hook is used outside of dashboards. 

I verified that both types of charts refetch every time now and to not fetch twice. Here's some videos

Explore charts only fetch once on each refresh
Explore1

SQL charts only fetch once on each refresh
Sql1Fetch

After this fix, all charts refresh

After

### DIFF
--- a/packages/frontend/src/components/common/Dashboard/DashboardRefreshButton.tsx
+++ b/packages/frontend/src/components/common/Dashboard/DashboardRefreshButton.tsx
@@ -65,26 +65,20 @@ export const DashboardRefreshButton: FC<DashboardRefreshButtonProps> = memo(
         const clearCacheAndFetch = useDashboardContext(
             (c) => c.clearCacheAndFetch,
         );
-        const invalidateCache = useDashboardContext((c) => c.invalidateCache);
 
         const setIsAutoRefresh = useDashboardContext((c) => c.setIsAutoRefresh);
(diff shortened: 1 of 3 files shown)


magnew:
* **#15060** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15060?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15060?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #15100, opened 2025-05-30
fix: system prompts not working for web ui

tatianainama:
Fixed agent settings retrieval when handling Slack or web app prompts. Before we were only getting agent settings from Slack channel prompts.


notgiorgi (approved):
Great catch 🚢

Issue #2510, opened 2022-06-22
Add ability to authenticate Snowflake via SSO


hamzahc1:
### Description

- [ ] add ability to authenticate Snowflake via SSO
- [ ] Add docs for this here

magants:
The inability of LD to provide SSO authentication is preventing our organization from using this product.

hamzahc1:
@all-contributors add @magants for ideas

gamezober:
> The inability of LD to provide SSO authentication is preventing our organization from using this product.

This is the same for our organization!

jw-aleksandr-zolotukhin:
+1
pretty serious limitation that could prevent adoption for companies that are tough on their data privacy protection.

menno-vandebron:
Agree with the comments above, we would need this to adhere to our new security guidelines

PR #15126, opened 2025-06-02
feat: snowflake project creation SSO

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/2510

### Description:
image

[Screencast from 2025-06-04 09-55-54.webm](https://github.com/user-attachments/assets/956be3d3-963b-418d-9b13-ab8a2aa83dea)

Added Snowflake OAuth integration to enable users to authenticate with their Snowflake credentials. This implementation includes:

- Created a new Snowflake passport strategy using OpenID Connect
- Added configuration options for Snowflake OAuth (client ID, client secret, endpoints)
- Set up login and callback routes for the Snowflake authentication flow
- Updated the OpenIdIdentityIssuerType enum to include Snowflake as an identity provider

The integration follows the same pattern as other OAuth providers, leveraging the existing OIDC handler for user authentication.


rephus:
* **#15126** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15126?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
Unrelated failing test 
image

lightdash local.session.sql
rephus, line 1: revert commit or add to gitignore ?

.vscode/settings.json
```diff
@@ -6,5 +6,17 @@
...
+  ],
+  "sqltools.connections": [
+    {
+      "previewLimit": 50,
+      "server": "localhost",
+      "port": 5432,
+      "driver": "PostgreSQL",
+      "name": "lightdash local",
+      "database": "postgres",
+      "username": "stevegardner",
```
rephus: this might be useful for everyone if you make it generic

packages/frontend/src/components/ProjectConnection/WarehouseForms/SnowflakeForm.tsx
```diff
@@ -34,12 +45,49 @@ export const SnowflakeSchemaInput: FC<{
...
 const SnowflakeForm: FC<{
     disabled: boolean;
 }> = ({ disabled }) => {
     const [isOpen, toggleOpen] = useToggle(false);
     const { savedProject } = useProjectFormContext();
     const form = useFormContext();
+    // TODO: These are stubs for the actual hooks
+    const {
+        data,
+        error: bigqueryAuthError,
```
rephus: rename bigquery variables

packages/frontend/src/components/ProjectConnection/WarehouseForms/SnowflakeForm.tsx
```diff
@@ -108,43 +110,53 @@ const SnowflakeForm: FC<{
...
+        : isSsoEnabled
+        ? SnowflakeAuthenticationType.SSO
+        : hasPrivateKey
+        ? SnowflakeAuthenticationType.PRIVATE_KEY
+        : SnowflakeAuthenticationType.PASSWORD;
 
-    const authenticationType: string =
-        form.values.warehouse.authenticationType ??
-        (hasPrivateKey ? 'private_key' : 'password');
+    if (!form.isDirty()) {
```
stevelikesmusic: We need react to re-render if the default value is different than what we started with. This can be because results coming back from SSO auth.

packages/frontend/src/components/ProjectConnection/WarehouseForms/SnowflakeForm.tsx
```diff
@@ -166,16 +178,10 @@ const SnowflakeForm: FC<{
...
-                                ? SnowflakeAuthenticationType.SSO
-                                : hasPrivateKey
-                                ? SnowflakeAuthenticationType.PRIVATE_KEY
-                                : SnowflakeAuthenticationType.PASSWORD
-                        }
+                        defaultValue={defaultAuthType}
                         label="Authentication Type"
                         description={
-                            isAuthenticated ? (
+                            isLoadingAuth ? null : isAuthenticated ? (
```
stevelikesmusic, line 185: Returning`null` so we don't have a "Loading...." flash on the screen.

packages/frontend/src/components/ProjectConnection/WarehouseForms/defaultValues.ts
```diff
@@ -104,7 +104,7 @@ export const SnowflakeDefaultValues: CreateSnowflakeCredentials = {
     queryTag: '',
     accessUrl: '',
     startOfWeek: undefined,
-    authenticationType: 'private_key',
+    authenticationType: undefined,
```
stevelikesmusic, line 107: We want to handle this in the component rather than refactor how we load default values into the Warehouse forms.

packages/frontend/src/components/ProjectConnection/WarehouseForms/util.ts
```diff
@@ -0,0 +1,6 @@
+import { capitalize, type WarehouseTypes } from '@lightdash/common';
+
+export const getSsoLabel = (warehouse: WarehouseTypes) =>
+    `User Account (Sign in with ${capitalize(warehouse)})`;
+export const PRIVATE_KEY_LABEL = `Service Account (JSON key file)`;
+export const PASSWORD_LABEL = `Password`;
```
stevelikesmusic, line 6: A reusable way to get the labels shaped how @owlas proposes in #15159

packages/backend/src/config/parseConfig.ts
```diff
@@ -964,6 +974,15 @@ export const parseConfig = (): LightdashConfig => {
...
                 scopes: process.env.AUTH_OIDC_SCOPES,
             },
+            snowflake: {
+                clientId: process.env.SNOWFLAKE_OAUTH_CLIENT_ID,
+                clientSecret: process.env.SNOWFLAKE_OAUTH_CLIENT_SECRET,
+                authorizationEndpoint:
+                    process.env.SNOWFLAKE_OAUTH_AUTHORIZATION_ENDPOINT,
+                tokenEndpoint: process.env.SNOWFLAKE_OAUTH_TOKEN_ENDPOINT,
+                loginPath: '/login/snowflake',
+                callbackPath: '/oauth/redirect/snowflake',
```
stevelikesmusic, line 984: Are these two fields not dynamic from Snowflake? We do we reference these values rather than set the router to the string endpoints if they're hardcoded?

https://github.com/lightdash/lightdash/blob/95258402bb2f0ad89953ad6470588015a61ed37d/packages/backend/src/routers/apiV1Router.ts#L178-L192

rephus: no, these 2 are lightdash specific endpoints

rephus: > We do we reference these values rather than set the router to the string endpoints if they're hardcoded

We could, but this is how we do it for other sso providers, so this way it is consistent. 
I imagine the original reason was to keep it centralized in config, in case we want to change it in the future to another ENV variable.

rephus:
could you take a look at this failing test @stevelikesmusic  ? 
image

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -793,13 +794,14 @@ export class ProjectModel {
                 if (rawCredentials.privateKey === undefined) {
                     return {
                         ...nonSensitiveCredentials,
-                        authenticationType: 'password',
+                        authenticationType:
+                            SnowflakeAuthenticationType.PASSWORD,
```
stevelikesmusic, line 798: @rephus do we need to change anything here? I'm not quite following the different states `rawCredentials` might have.

stevelikesmusic:
> could you take a look at this failing test @stevelikesmusic ?
@rephus fixed in the last commit 👍

packages/backend/src/models/ProjectModel/ProjectModel.ts
rephus: that is ok, I don't think we need to add anything for SSO in here. This code was added recently to deal with `undefined` authenticationType. 

when adding  sso throught the new form, it should always have `nonSensitiveCredentials.authenticationType === sso`

rephus:
unrelated e2e error

image

Issue #15200, opened 2025-06-05
Leave tab mode if there's only 1 tab


stevelikesmusic:
### Description

When deleting the penultimate tab in a dashboard, we should leave tab mode.

### What problem does this solve?

It's a bit cumbersome to be left on one tab and remain in tab mode. If I've deleted all but one tab, it's likely I don't want tabs at all. We won't delete the last tab in the DB just in case we want to keep the tab name—though it's likely the tab name isn't all that important. Perhaps we could prompt for the last tab?

PR #15204, opened 2025-06-06
feat: exit tab mode if we're down to a single tab

stevelikesmusic:
Closes: #15200

### Description:

This is a UX improvement where we only show tabs if there is more than one. There's no reason to have a single tab, since we have just one dashboard.

When exiting tab mode from having one tab, we keep that tab in the database. For now that means a user cannot delete the final tab so that we keep the tab name. Maybe this isn't all that important and we can revist later.

### DIFF
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -87,9 +87,14 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
     const [isDeletingTab, setDeletingTab] = useState<boolean>(false);
 
     const defaultTab = dashboardTabs?.[0];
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -183,7 +188,14 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
...
             });
-            return; // keep all tiles if its the last tab
+            // If this is the last tab, navigate to the non-tab URL.
+            // See `const = sortedTabs` for more context.
+            void navigate(
+                `/projects/${projectUuid}/dashboards/${dashboardUuid}/edit`,
+                { replace: true },
+            );
+
+            return;
```
joaoviana, line 198: tested many ways and works great 👍

joaoviana (approved):

PR #15223, opened 2025-06-06
feat: store thread context when summoning ai slack bot

joaoviana:
Closes:

### DIFF
--- a/packages/backend/src/ee/clients/Slack/SlackBot.ts
+++ b/packages/backend/src/ee/clients/Slack/SlackBot.ts
@@ -5,6 +5,7 @@ import {
     KnownBlock,
     SlackEventMiddlewareArgs,
 } from '@slack/bolt';
+import { type MessageElement } from '@slack/web-api/dist/response/ConversationsHistoryResponse';
 import { SlackBot, SlackBotArguments } from '../../../clients/Slack/Slackbot';
 import Logger from '../../../logging/logger';
 import { AiAgentModel } from '../../models/AiAgentModel';
@@ -276,10 +277,12 @@ export class CommercialSlackBot extends SlackBot {
         });
     }
 
+    // WARNING: Needs - channels:history scope for all slack apps
     private async handleAppMention({
         event,
         context,
         say,
+        client,
     }: SlackEventMiddlewareArgs<'app_mention'> & AllMiddlewareArgs) {
         Logger.info(`Got app_mention event ${event.text}`);
 
@@ -309,6 +312,44 @@ export class CommercialSlackBot extends SlackBot {
 
             name = agentConfig.name;
 
+            let threadMessages:
+                | Array<Required<Pick<MessageElement, 'text' | 'user' | 'ts'>>>
+                | undefined;
+            if (event.thread_ts) {
+                try {
+                    const threadHistory = await client.conversations.replies({
+                        channel: event.channel,
+                        ts: event.thread_ts,
+                        limit: 100, // TODO: What should be the limit?
+                    });
+
+                    if (
+                        threadHistory.messages &&
+                        threadHistory.messages.length > 1
+                    ) {
+                        // Extract previous messages (excluding the current mention) for database storage
+                        threadMessages = threadHistory.messages
+                            .filter((msg) => msg.ts !== event.ts)
+                            .map((msg) => ({
+                                text: msg.text || '[message]',
+                                user: msg.user || 'unknown', // TODO: use the user uuid from the message?
+                                ts: msg.ts || '',
+                            }))
+                            .filter((msg) => msg.ts); // Only include messages with valid timestamps
+
+                        Logger.info(
+                            `Found thread context with ${threadMessages.length} previous messages`,
+                        );
+                    }
+                } catch (error) {
+                    Logger.warn(
+                        'Failed to fetch thread history, using original message only:',
+                        error,
+                    );
+                    // Fall back to original behavior if thread fetch fails
+                }
+            }
+
             [slackPromptUuid, createdThread] =
                 await this.aiService.createSlackPrompt({
                     userUuid,
@@ -319,6 +360,7 @@ export class CommercialSlackBot extends SlackBot {
                     prompt: event.text,
                     promptSlackTs: event.ts,
                     agentUuid: agentConfig.uuid ?? null,
+                    threadMessages,
                 });
         } catch (e) {
             if (e instanceof AiDuplicateSlackPromptError) {
--- a/packages/backend/src/ee/models/AiModel.ts
+++ b/packages/backend/src/ee/models/AiModel.ts
@@ -405,4 +405,64 @@ export class AiModel {
             })
             .returning('ai_prompt_uuid');
     }
+
+    async existsSlackPromptsByChannelAndTimestamps(
+        slackChannelId: string,
+        timestamps: string[],
+    ): Promise<string[]> {
+        if (timestamps.length === 0) return [];
+
+        const existingPrompts = await this.database(AiSlackPromptTableName)
+            .select('prompt_slack_ts')
+            .where('slack_channel_id', slackChannelId)
+            .whereIn('prompt_slack_ts', timestamps);
+
+        return existingPrompts.map((p) => p.prompt_slack_ts);
+    }
+
+    // TODO: reuse this?
+    async bulkCreateSlackPrompts(
+        threadUuid: string,
+        promptsData: Pick<
+            SlackPrompt,
+            | 'createdByUserUuid'
+            | 'prompt'
+            | 'slackUserId'
+            | 'slackChannelId'
+            | 'promptSlackTs'
+            | 'createdAt'
+        >[],
+    ): Promise<string[]> {
+        if (promptsData.length === 0) return [];
+
+        return this.database.transaction(async (trx) => {
+            const promptRows = await trx(AiPromptTableName)
+                .insert(
+                    promptsData.map((data) => ({
+                        ai_thread_uuid: threadUuid,
+                        created_by_user_uuid: data.createdByUserUuid,
+                        prompt: data.prompt,
+                        ...(data.createdAt
+                            ? { created_at: data.createdAt }
+                            : {}),
+                    })),
+                )
+                .returning('ai_prompt_uuid');
+
+            if (promptRows.length !== promptsData.length) {
+                throw new Error('Failed to create all prompts');
+            }
+
+            await trx(AiSlackPromptTableName).insert(
+                promptRows.map((row, index) => ({
+                    ai_prompt_uuid: row.ai_prompt_uuid,
+                    slack_user_id: promptsData[index].slackUserId,
+                    slack_channel_id: promptsData[index].slackChannelId,
+                    prompt_slack_ts: promptsData[index].promptSlackTs,
+                })),
+            );
+
+            return promptRows.map((row) => row.ai_prompt_uuid);
+        });
+    }
 }
--- a/packages/backend/src/ee/services/AiService/AiService.ts
+++ b/packages/backend/src/ee/services/AiService/AiService.ts
@@ -25,6 +25,7 @@ import {
     isField,
     isSlackPrompt,
 } from '@lightdash/common';
+import { type MessageElement } from '@slack/web-api/dist/response/ConversationsHistoryResponse';
 import { CoreMessage } from 'ai';
 import { pick } from 'lodash';
 import slackifyMarkdown from 'slackify-markdown';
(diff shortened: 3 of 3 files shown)


joaoviana:
* **#15223** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15223?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/clients/Slack/SlackBot.ts
```diff
@@ -309,6 +312,44 @@ export class CommercialSlackBot extends SlackBot {
...
 
+            let threadMessages:
+                | Array<Required<Pick<MessageElement, 'text' | 'user' | 'ts'>>>
+                | undefined;
+            if (event.thread_ts) {
+                try {
+                    const threadHistory = await client.conversations.replies({
+                        channel: event.channel,
+                        ts: event.thread_ts,
+                        limit: 100, // TODO: What should be the limit?
```
joaoviana: ❓

packages/backend/src/ee/clients/Slack/SlackBot.ts
```diff
@@ -309,6 +312,44 @@ export class CommercialSlackBot extends SlackBot {
...
+                    if (
+                        threadHistory.messages &&
+                        threadHistory.messages.length > 1
+                    ) {
+                        // Extract previous messages (excluding the current mention) for database storage
+                        threadMessages = threadHistory.messages
+                            .filter((msg) => msg.ts !== event.ts)
+                            .map((msg) => ({
+                                text: msg.text || '[message]',
+                                user: msg.user || 'unknown', // TODO: use the user uuid from the message?
```
joaoviana: ❓ don't know what to do here. Advice pls

packages/backend/src/ee/clients/Slack/SlackBot.ts
```diff
@@ -309,6 +312,44 @@ export class CommercialSlackBot extends SlackBot {
...
+                                ts: msg.ts || '',
+                            }))
+                            .filter((msg) => msg.ts); // Only include messages with valid timestamps
```
joaoviana: almost feels like this won't happen, but maybe more robust this way?

packages/backend/src/ee/models/AiModel.ts
```diff
@@ -405,4 +405,64 @@ export class AiModel {
...
+        timestamps: string[],
+    ): Promise<string[]> {
+        if (timestamps.length === 0) return [];
+
+        const existingPrompts = await this.database(AiSlackPromptTableName)
+            .select('prompt_slack_ts')
+            .where('slack_channel_id', slackChannelId)
+            .whereIn('prompt_slack_ts', timestamps);
+
+        return existingPrompts.map((p) => p.prompt_slack_ts);
```
joaoviana: sufficient filtering?

packages/backend/src/ee/models/AiModel.ts
```diff
@@ -405,4 +405,64 @@ export class AiModel {
...
+    }
+
+    // TODO: reuse this?
```
joaoviana: should this be reused to handle single-slack prompts as well?

packages/backend/src/ee/services/AiService/AiService.ts
```diff
@@ -734,6 +735,55 @@ export class AiService {
...
+        if (newMessages.length === 0) return;
+
+        // Convert Slack timestamp to Date (Slack ts is Unix timestamp with microseconds)
+        const convertSlackTsToDate = (ts: string): Date =>
+            new Date(parseFloat(ts) * 1000); // Convert to milliseconds
+
+        // Prepare data for bulk insert
+        const promptsData = newMessages
+            .map((msg) => ({
+                createdByUserUuid: fallbackUserUuid, // TODO: use the user uuid from the message
```
joaoviana: similar to question above, not sure 🤔

packages/backend/src/ee/services/AiService/AiService.ts
```diff
@@ -734,6 +735,55 @@ export class AiService {
...
+                promptSlackTs: msg.ts,
+                createdAt: convertSlackTsToDate(msg.ts),
+            }))
+            .sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
+
+        try {
+            await this.aiModel.bulkCreateSlackPrompts(threadUuid, promptsData);
+        } catch (error) {
+            Logger.error('Failed to store thread context messages:', error);
+            // TODO: handle this error?
```
joaoviana: I think we can remove the trycatch here, wdyt?

packages/backend/src/ee/clients/Slack/SlackBot.ts
```diff
@@ -276,10 +277,12 @@ export class CommercialSlackBot extends SlackBot {
         });
     }
 
+    // WARNING: Needs - channels:history scope for all slack apps
```
notgiorgi, line 283: should it be added here? 
https://github.com/lightdash/lightdash/blob/f71e79f5ae58646605e9a13a8935df885b6f8f4d/packages/common/src/types/slackSettings.ts#L15-L27

Also how does adding scope work? they'd have to re-install the slack app right? Which means they have to re-configure all agents?

joaoviana: had no idea this type existed! 

> Also how does adding scope work? they'd have to re-install the slack app right? 

We add the scope to each slack app (tedious) and they would have to reinstall in Lightdash

> Which means they have to re-configure all agents?

I don't think this is the case...? I would have to test, but I don't think the slack installation id would change 🤔 so it's just a matter of reinstalling...

notgiorgi: also, does this become default behavior? Or maybe it should only happen if user explicitly says "read the thread?" 🤔

joaoviana: I believe from a user pov context needs to be passed no matter what if they prompt the bot, so my thinking is, if someone tags the bot, then the bot should have access to what was talked about 🤔

notgiorgi: I think that does make sense 💯  

But what I meant is that it's a "breaking change"

joaoviana:
1. If summoned in thread, add:
`Yes, Yes and remember, No` options so that slack bot either stores all threads, remembers or doesn't get context

packages/backend/src/ee/clients/Slack/SlackBot.ts
joaoviana: Added consent screen when summoned: 




this info (consent) is stored on a separate table - we can chat about this implementation

packages/backend/src/ee/database/migrations/20250610102119_slack_installation_ai_thread_consent.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from 'knex';
+
+const SlackAuthTokensTableName = 'slack_auth_tokens';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(SlackAuthTokensTableName, (table) => {
+        table.boolean('ai_thread_access_consent').defaultTo(false);
```
joaoviana, line 7: slack installation now tracks ai thread consent - only for ee

joaoviana:
final result (consent granted)

Screenshot 2025-06-10 at 15 58 55
Screenshot 2025-06-10 at 15 59 05

packages/backend/src/ee/clients/Slack/SlackBot.ts
```diff
@@ -378,4 +410,68 @@ export class CommercialSlackBot extends SlackBot {
...
+        channelId,
+        threadTs,
+        excludeMessageTs,
+        hasConsent,
+    }: {
+        client: WebClient;
+        channelId: string;
+        threadTs: string;
+        excludeMessageTs: string;
+        hasConsent: boolean;
```
notgiorgi: Is this redundant? Method is not called w/o permissions

packages/backend/src/ee/services/AiAgentService.ts
```diff
@@ -1056,6 +1057,58 @@ export class AiAgentService {
...
+        if (newMessages.length === 0) return;
+
+        // Convert Slack timestamp to Date (Slack ts is Unix timestamp with microseconds)
+        const convertSlackTsToDate = (ts: string): Date =>
+            new Date(parseFloat(ts) * 1000); // Convert to milliseconds
+
+        // Prepare data for bulk insert
+        const promptsData = newMessages
+            .map((msg) => ({
+                createdByUserUuid: fallbackUserUuid, // TODO: use the user uuid from the message
```
notgiorgi, line 1098: We have a method in slack client that gets actual slack users info, you can do it like this:

https://github.com/lightdash/lightdash/blob/ac1cd0fafa86374e0a176a5b1e3a7adea31f5316/packages/backend/src/ee/services/AiAgentService.ts#L355-L366

packages/frontend/src/components/UserSettings/SlackSettingsPanel/index.tsx
```diff
@@ -248,20 +251,47 @@ const SlackSettingsPanel: FC = () => {
...
+                                        Configure Slack channel project mappings{' '}
+                                        <Anchor
+                                            component={Link}
+                                            to="/generalSettings/aiAgents"
+                                        >
+                                            here
+                                        </Anchor>
+                                        .
+                                    </Alert>
+                                </Stack>
```
notgiorgi: Let's make the copy consistent here, should it be `AI` or `AI Agent` or `AI Analyst?

Should we have a disclaimer here saying that longer threads decrease agent performance?

notgiorgi (commented):
Code looking great, testing it now

packages/frontend/src/components/UserSettings/SlackSettingsPanel/index.tsx
notgiorgi: Should there be a visual indication I have to re-install the app in order to use the new feature? I toggled the switch and tried to summon without installing

notgiorgi: I tried `re-install`: and got this in the terminal:

`warn: Failed to fetch thread history, using original message only: An API error occurred: missing_scope`

notgiorgi: Okay actually re-install works, you just need to add `channels:history` scope to `slackSettings.ts`

It also does display a warning on integrations page. Not sure if we want to display that warning to users without AI/Thread access enabled though

packages/backend/src/ee/clients/Slack/SlackBot.ts
```diff
@@ -378,4 +410,68 @@ export class CommercialSlackBot extends SlackBot {
...
+        if (!hasConsent || !threadTs) {
+            return undefined;
+        }
+
+        try {
+            const threadHistory = await client.conversations.replies({
+                channel: channelId,
+                ts: threadTs,
+                limit: 100, // TODO: What should be the limit?
+            });
```
notgiorgi, line 471: this has a bug where it also re-saves messages from our bot:





You can test this by going to the old ui (new ui will be broken for this)

notgiorgi (commented):
LGTM! Just a few issues that need to be fixed

packages/backend/src/ee/clients/Slack/SlackBot.ts
joaoviana: :100: yes !

packages/backend/src/ee/clients/Slack/SlackBot.ts
joaoviana: ah! will look into this now

packages/frontend/src/components/UserSettings/SlackSettingsPanel/index.tsx
joaoviana: Made this change based on your comments

image


And yes to the scope - since we have to update the scope for them, we can help with that onboarding

packages/backend/src/ee/clients/Slack/SlackBot.ts
joaoviana: Improved filtering - this is now working with "subtype" and "bot id" filtering before sending the messages - there's a clear check before calling the aiAgentService

packages/backend/src/ee/services/AiAgentService.ts
joaoviana: Looking into this now

Issue #10214, opened 2024-05-28
I want to be able to specify my csv exports to use UTF-8 BOM


TuringLovesDeathMetal:
### Description

Right now, we allow users to export files as csv types. 

This is great, but for some languages (e.g. Japanese) exporting to CSV in UTF-8 actually causes issues with characters. 

Excel isn't very good at recognizing UTF-8, so, to fix this, you need to actually import the .csv to excel as UTF-8 BOM. 

There's a discussion here which makes it more clear what's going on: https://stackoverflow.com/questions/6002256/is-it-possible-to-force-excel-recognize-utf-8-csv-files-automatically

Basically, Lightdash should offer the option to users to import .csv files to excel using UTF-8 BOM (or maybe we just use this by default?)

PR #15116, opened 2025-06-01
Deprecate legacy download endpoints

Divyansh200102:
Closes: #15013

### Description:

image

image

PR #15226, opened 2025-06-07
feat: add UTF-8 BOM support to CSV exports

Divyansh200102:
Closes: #10214 

### Description:

This PR updates the CSV export functionality to include a UTF-8 BOM at the beginning of exported files. This ensures proper character rendering—especially for non-Latin scripts like Japanese—when opening CSVs in Microsoft Excel, which otherwise fails to correctly interpret UTF-8 encoding.

Before:
Screenshot 2025-06-07 172639

After:
Screenshot 2025-06-07 174443


pimmey:
Fix in #15269, applied to the new /download endpoints. We're going to deprecate the old ones, to which these changes apply (#15116)

pimmey:
Actually, we still use this when exporting pivot tables 😄 reopening!

pimmey (approved):
⚡️

Issue #13715, opened 2025-02-17
Analytics events


PriPatel:
Add Rudderstack usage events to track adoption. This should be used to create a dashboard for feature adoption. 

We'll track:
- Tab creation
- Total tabs
-

stevelikesmusic:
@PriPatel is there any more context for this one?

PR #15240, opened 2025-06-09
feat: add tabs count analytics when saving a dashbaord

stevelikesmusic:
Closes: #13715 

### Description:

Adds analytics for tabs usage. 

### Related

[analytics PR
](https://github.com/lightdash/lightdash-analytics/pull/628)

### DIFF
--- a/packages/backend/src/services/DashboardService/DashboardService.ts
+++ b/packages/backend/src/services/DashboardService/DashboardService.ts
@@ -154,6 +154,7 @@ export class DashboardService
             loomTilesCount: dashboard.tiles.filter(
                 ({ type }) => type === DashboardTileTypes.LOOM,
             ).length,
+            tabsCount: dashboard.tabs.length,
         };
     }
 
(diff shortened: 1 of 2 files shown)


rephus (approved):
Perfect

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -154,6 +154,7 @@ export class DashboardService
             loomTilesCount: dashboard.tiles.filter(
                 ({ type }) => type === DashboardTileTypes.LOOM,
             ).length,
+            tabsCount: dashboard.tabs.length || 0,
```
rephus: If `tabs` can't be undefined, then we don't need `|| 0` , right ? According to the `DashboardDAO` type, tabs is not optional

 dashboard.tabs.length will throw an error if tabs is undefined

perhaps a safer option would be to do: `dashboard.tabs?.length || 0`
or `(dashboard.tabs || []).length`

rephus (changes requested):

packages/backend/src/services/DashboardService/DashboardService.ts
stevelikesmusic: That's true

Issue #15246, opened 2025-06-10
Agents list/chat should be project-dependant


joaoviana:
### Description

AC
- [ ] projects/UUID/ai-agents/...
- [ ] Should only fetch ai agents for the project

### What problem does this solve?

Clarity based on the project switcher and search functionality on the nav bar

PR #15263, opened 2025-06-10
feat: show agents only in project

joaoviana:
Closes: #15246

### Description:

This PR adds project-specific AI agents endpoints and UI. It introduces a new endpoints in `AiAgentController` to fetch AI agents scoped to a specific project. The UI has been updated to use project-specific routes for AI agents, changing from `/ai-agents` to `/projects/{projectUuid}/ai-agents`.

The implementation includes:
- New API endpoints for listing and getting project-specific AI agents
- Updated model and service layers to support filtering agents by project UUID
- Refactored frontend hooks to separate organization-wide and project-specific AI agent operations
- Updated navigation and routing to use project-scoped paths for AI agents

### DIFF
--- a/packages/backend/src/ee/controllers/aiAgentController.ts
+++ b/packages/backend/src/ee/controllers/aiAgentController.ts
@@ -331,3 +331,52 @@ export class AiAgentController extends BaseController {
         return this.services.getAiAgentService<AiAgentService>();
     }
 }
+
+@Route('/api/v1/projects/{projectUuid}/aiAgents')
+@Hidden()
+@Response<ApiErrorPayload>('default', 'Error')
+export class ProjectAiAgentController extends BaseController {
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Get('/')
+    @OperationId('listProjectAgents')
+    async listProjectAgents(
+        @Request() req: express.Request,
+        @Path() projectUuid: string,
+    ): Promise<ApiAiAgentSummaryResponse> {
+        this.setStatus(200);
+
+        return {
+            status: 'ok',
+            results: await this.getAiAgentService().listAgents(
+                req.user!,
+                projectUuid,
+            ),
+        };
+    }
+
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Get('/{agentUuid}')
+    @OperationId('getProjectAgent')
+    async getProjectAgent(
+        @Request() req: express.Request,
+        @Path() projectUuid: string,
+        @Path() agentUuid: string,
+    ): Promise<ApiAiAgentResponse> {
+        this.setStatus(200);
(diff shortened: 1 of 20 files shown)


joaoviana:
* **#15263** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15263?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15263?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/controllers/aiAgentController.ts
```diff
@@ -331,3 +331,52 @@ export class AiAgentController extends BaseController {
...
+            ),
+        };
+    }
+
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Get('/{agentUuid}')
+    @OperationId('getProjectAgent')
+    async getProjectAgent(
+        @Request() req: express.Request,
```
joaoviana, line 363: only need these 2 for now

tatianainama: we cannot deprecate/delete the others because we use them in organization settings right? 🤔

packages/frontend/src/ee/features/aiCopilot/hooks/useOrganizationAiAgents.ts
tatianainama, line 1: nit: I'm leaning towards making these useQuery functions more strict and enforce the existence of uuids, wdyt?
Imo, the consumer should take care of making sure the id is available before calling this function

For example, in the following, it will simplify the whole logic of the function if any of the params is missing
```ts
export const useGenerateAgentThreadResponseMutation = (
    projectUuid: string | undefined,
    agentUuid: string | undefined,
    threadUuid: string | undefined,
) => {

// ...

  mutationFn: (data) =>
      agentUuid && threadUuid // no need for ternaries
          ? generateAgentThreadResponse(agentUuid, threadUuid, data)
          : Promise.reject(), 

}
```

packages/backend/src/ee/controllers/aiAgentController.ts
joaoviana: correct! We should keep both endpoints for now

packages/frontend/src/ee/features/aiCopilot/hooks/useOrganizationAiAgents.ts
joaoviana: hmm I see! I've been following this pattern for some time but I completely understand what you mean. 🤔 

> Imo, the consumer should take care of making sure the id is available before calling this function

I like this, but how would you check the existence of the uuid on mutation?
Maybe this mutation hook shouldn't receive any args, and the mutate function should receive all the args necessary instead? 

Let's continue this chat and refactor on a separate PR (just to move this arch change forward if that's ok 🙏 )

Issue #15247, opened 2025-06-10
I should be able to change agent while I'm on the chat view


joaoviana:
### Description

AC

- [ ] Add an agent switcher on the chat view

### What problem does this solve?

Allows quick context changes

PR #15286, opened 2025-06-11
feat: Introduce Agent Switcher

tatianainama:
Closes: #15247

### Description:

This PR introduces an Agent Switcher to easily navigate between project agents.
I've also moved things around, like reducing the settings button to just display an icon, and moved the `new thread` button down to the threads list

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
@@ -1,0 +1,90 @@
+import { type AiAgent } from '@lightdash/common';
+import { type ComboboxItem, Group, Select, Text } from '@mantine-8/core';
+import { IconPointFilled } from '@tabler/icons-react';
+import { useNavigate } from 'react-router';
+import { LightdashUserAvatar } from '../../../../components/Avatar';
+import MantineIcon from '../../../../components/common/MantineIcon';
+
+type Props = {
+    agents: Pick<AiAgent, 'name' | 'uuid' | 'imageUrl'>[];
+    selectedAgent: AiAgent;
+    projectUuid: string;
+};
+
+interface AgentSelectOption extends ComboboxItem {
+    imageUrl?: AiAgent['imageUrl'];
+}
+
+const renderSelectOption = ({
+    option,
+    checked,
+}: {
+    option: AgentSelectOption;
+    checked?: boolean;
+}) => (
+    <Group gap="xs" wrap="nowrap" miw={0} flex={1}>
+        <LightdashUserAvatar
+            size={20}
+            variant="filled"
+            name={option.label}
+            src={option.imageUrl}
+        />
+        <Text size="xs" truncate="end" flex={1}>
+            {option.label}
+        </Text>
+
+        {checked && <MantineIcon icon={IconPointFilled} size={12} />}
+    </Group>
+);
+
+export const AgentSwitcher = ({
+    agents,
+    selectedAgent,
+    projectUuid,
+}: Props) => {
+    const navigate = useNavigate();
+    const handleSelect = (value: string | null) => {
+        if (value)
+            void navigate(`/projects/${projectUuid}/ai-agents/${value}`, {
+                viewTransition: true,
+            });
+    };
+    const agentOptions = agents.map(
+        ({ name, uuid, imageUrl }) =>
+            ({
+                label: name,
+                value: uuid,
+                imageUrl: imageUrl,
+            } satisfies AgentSelectOption),
+    );
+    return (
+        <Select
+            data={agentOptions}
+            defaultValue={selectedAgent.uuid}
+            onChange={handleSelect}
+            checkIconPosition="right"
+            w="100%"
+            allowDeselect={false}
+            renderOption={renderSelectOption}
+            withScrollArea={false}
+            leftSection={
+                <LightdashUserAvatar
+                    size={22}
+                    variant="filled"
+                    name={selectedAgent.name}
+                    src={selectedAgent.imageUrl}
+                />
+            }
+            styles={(theme) => ({
+                input: {
+                    borderColor: theme.colors.gray[2],
+                    boxShadow: `var(--mantine-shadow-subtle)`,
+                    color: theme.colors.gray[7],
+                },
+                dropdown: {
+                    paddingRight: 4,
+                },
+            })}
+        />
+    );
+};
--- a/packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
+++ b/packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
@@ -1,11 +1,11 @@
 import { type AiAgent, type AiAgentThreadSummary } from '@lightdash/common';
 import {
+    ActionIcon,
     Box,
     Button,
     Center,
     Divider,
     Group,
-    List,
     Loader,
     NavLink,
     Paper,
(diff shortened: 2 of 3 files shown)


tatianainama:
* **#15286** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15286?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
```diff
@@ -205,7 +174,18 @@ const AgentPage = () => {
 
                     {projectUuid && threads && threads.length > 0 && (
                         <Stack gap="xs">
-                            <Title order={6}>Threads</Title>
+                            <Group justify="space-between">
+                                <Title order={6}>Threads</Title>
+                                <Button
+                                    size="compact-xs"
+                                    variant="subtle"
+                                    leftSection={<IconPlus size={16} />}
```
joaoviana: what about a write icon or something?

packages/frontend/src/mantine8Theme.ts
```diff
@@ -153,6 +154,22 @@ export const getMantine8ThemeOverride = (
...
+                styles: (theme) => ({
+                    input: {
+                        borderColor: theme.colors.gray[2],
+                        boxShadow: `var(--mantine-shadow-subtle)`,
+                        color: theme.colors.gray[7],
+                    },
+                    dropdown: {
+                        paddingRight: 4,
+                    },
+                }),
```
joaoviana: Do we want these styles to be applied to all `Select`s under this mantine provider? I don't think so 🤔

packages/frontend/src/mantine8Theme.ts
```diff
@@ -141,9 +142,9 @@ export const getMantine8ThemeOverride = (
                     radius: 'md',
                     shadow: 'subtle',
                     withBorder: true,
-                    sx: (theme: MantineTheme) => ({
-                        '&[data-with-border]': {
-                            border: `1px solid ${theme.colors.gray[2]}`,
+                    styles: (theme: MantineTheme) => ({
+                        root: {
+                            borderColor: theme.colors.gray[2],
```
joaoviana, line 147: this will apply to all `Paper`s - I would like to have the option to not set a border - 🤔

packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
```diff
@@ -0,0 +1,81 @@
...
+            ({
+                label: name,
+                value: uuid,
+                imageUrl: imageUrl,
+            } satisfies AgentSelectOption),
+    );
+    return (
+        <Select
+            data={agentOptions}
+            defaultValue={selectedAgent.uuid}
```
joaoviana, line 63: can't we just control with `value`?

packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
```diff
@@ -0,0 +1,81 @@
...
+            variant="filled"
+            name={option.label}
+            src={option.imageUrl}
+        />
+        <Text size="xs" truncate="end">
+            {option.label}
+        </Text>
+
+        {checked && (
+            <IconPointFilled style={{ marginInlineStart: 'auto' }} size={12} />
```
joaoviana: > style={{ marginInlineStart: 'auto' }}

what does this do? And could you wrap this icon with `<MantineIcon icon={...}`?

packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
tatianainama: I thought about leaving this as an uncontrolled component as we are using it for navigation (and not actually a form select). If we want to use value, we will need to create state for it (and update it on selection)

packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
tatianainama: sounds good! Just the icon instead of the text?

packages/frontend/src/mantine8Theme.ts
tatianainama: This will only change the border color, so if you create a Paper without borders, it should not add them!

packages/frontend/src/mantine8Theme.ts
tatianainama: good point. I opted for this (instead of variants) because I like the idea of consistency and having all selects with the same style by default. We can always create specialized ones by updating the styles in the component itself.
But let me know, I can move these into the switcher instead!

packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
joaoviana: I think icon and text are fine? just a different icon?

And I wonder if the positioning makes it harder now to start a chat while viewing an existing thread

what about somewhere here and a bit bigger? 

image

packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
tatianainama: oops, I brought this from the Mantine example 😬 
I'll update!

packages/frontend/src/mantine8Theme.ts
joaoviana: no, it adds by default bc it's targeting `root`, see example

```tsx
    <Paper w={100} h={100} bg="gray.0">
                            j
                        </Paper>
```

image

packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
tatianainama: Good point. I'm gonna play around with it!

packages/frontend/src/mantine8Theme.ts
joaoviana: actually, mb - this is completely wrong. 

it adds a border bc `defaultProps` are `withBorder true` closing!

packages/frontend/src/ee/features/aiCopilot/components/AgentSwitcher.tsx
joaoviana: good point - defaultValue it is 👍

packages/frontend/src/mantine8Theme.ts
tatianainama: ah yes! Should I remove the defaultProp `withBorder` to make it more clear?

joaoviana: I like the "withBorder" being a default prop - we do it Spotlight's theming so these changes are great 👍

packages/frontend/src/mantine8Theme.ts
joaoviana: I'm just wondering if these would look a lot different if on a form for example - leading to making the changes to those components as well, like `TextInput`, `TagInput` etc
if you could move then the `styles` to the component itself that would be great - and then we can move the theme if we want to have it for all input-like components. wdyt?

tatianainama: sounds good! Will move! 🙏

packages/frontend/src/ee/pages/AiAgents/AgentPage.tsx
tatianainama: I kinda like the dark button better, but it might be too much since we reduce the amount of information here.
The light one works, but I'll need to work on the borders as it's a shade darker than the selects and it kinda annoys me 😅 wdyt?



(lighter borders)

tatianainama:
### Merge activity

* **Jun 12, 1:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15286).
* **Jun 12, 1:15 PM UTC**: @tatianainama merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15286).

PR #15321, opened 2025-06-13
feat: Add UI for managing Service Accounts

stevelikesmusic:
Closes: https://github.com/lightdash/lightdash-internal-work/issues/3456
Tied with #15261

### Description:

Adds a panel for managing service accounts. Admins can create and delete service accounts. 

### Create with multiple scopes

Screen Cast 2025-06-11 at 6 02 31 PM

### Create with a single scope

Screen Cast 2025-06-11 at 6 03 05 PM

### Delete an account

Screen Cast 2025-06-11 at 6 03 22 PM

## Feature Flag

This is shipping behind a [feature flag (service-accounts)](https://us.posthog.com/project/16377/feature_flags/147431) until we're ready to launch

### DIFF
--- a/packages/frontend/src/ee/features/serviceAccounts/ServiceAccountsCreateModal.tsx
+++ b/packages/frontend/src/ee/features/serviceAccounts/ServiceAccountsCreateModal.tsx
@@ -12,11 +12,13 @@ import { type FC } from 'react';
 
 import { ServiceAccountScope } from '@lightdash/common';
 
-const AVAILABLE_SCOPES = Object.values(ServiceAccountScope).map((scope) => ({
-    label: scope,
-    value: scope,
-    group: scope.split(':')[0],
-}));
+const AVAILABLE_SCOPES = Object.values(ServiceAccountScope)
+    .map((scope) => ({
+        label: scope,
+        value: scope,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/api.ts
```diff
@@ -47,15 +47,25 @@ const handleError = (err: any): ApiError => {
...
+    method: 'GET' | 'DELETE';
+    body?: BodyInit | null | undefined;
+};
+
+type LightdashApiPropsWrite = LightdashApiPropsBase & {
+    method: 'POST' | 'PATCH' | 'PUT';
+    body: BodyInit | null | undefined;
+};
+
+type LightdashApiProps = LightdashApiPropsGetOrDelete | LightdashApiPropsWrite;
```
rephus, line 67: why is needed this change ?

stevelikesmusic: This isn't needed per se. It just makes using the client a little easier type-wise. GET requests shouldn't have a body, and DELETE often doesn't have a body. Before, `body` was a required field for every requests. This change makes `body` optional—so as to be backwards compatible with all the `body: undefined` usage from the required field.

PR #15350, opened 2025-06-15
fix: prevent login page refresh when clicking Continue button

### LINKED ISSUE #15346: Login page refreshes when hitting 'Continue' button
## Bug Description

When users click the 'Continue' button on the login page, the page refreshes instead of proceeding with the login flow. This creates a frustrating user experience.

## Steps to Reproduce
1. Navigate to the login page
2. Enter login credentials 
3. Click the 'Continue' button
4. Observe that the page refreshes instead of continuing

## Expected Behavior
Clicking 'Continue' should proceed with the login flow without refreshing the page.

## Actual Behavior  
The page refreshes when 'Continue' is clicked, which is annoying and interrupts the login process.

## Impact
- Poor user experience during login
- Potential confusion for users expecting smooth login flow
- May appear as if the login system is broken

## Environment
- Affects login flow
- Reproducible on development environment

owlas:
## Summary
Fixes #15346 - Resolves multiple UI issues causing jarring page refresh when clicking the "Continue" button on the login page.

## Changes
- **Remove full page spinner during login options fetch** - Prevents jarring page replacement that caused the "refresh" effect
- **Fix form stage transition logic** - Wait for both `preCheckEmail` AND `loginOptions` before switching stages to prevent intermediate broken states
- **Add delayed loading spinner (400ms)** - Only show loading feedback if request takes longer than 400ms to avoid flashy UI on fast networks  
- **Preserve SSO options during loading** - Prevent Okta/SSO buttons from disappearing and reappearing during transitions

## Test plan
- [ ] Navigate to `/login` page
- [ ] Enter an email address and click "Continue"
- [ ] Verify no page flash/refresh occurs
- [ ] Verify smooth transition to password field
- [ ] Verify SSO buttons remain visible throughout transition
- [ ] Test on both fast and slow network conditions

## Before/After
**Before:** Jarring page replacement with PageSpinner when clicking Continue
**After:** Smooth transition with contextual loading states only when needed

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/frontend/src/features/users/components/LoginLanding.tsx
+++ b/packages/frontend/src/features/users/components/LoginLanding.tsx
@@ -57,6 +57,7 @@ const Login: FC<{}> = () => {
     }, [flashMessages.data, showToastError]);
 
     const [preCheckEmail, setPreCheckEmail] = useState<string>();
+    const [showDelayedLoading, setShowDelayedLoading] = useState(false);
 
     const redirectUrl = location.state?.from
         ? `${location.state.from.pathname}${location.state.from.search}`
@@ -92,6 +93,27 @@ const Login: FC<{}> = () => {
         }
     }, [loginOptionsSuccess, loginOptions]);
 
+    // Delayed loading state - only show loading if request takes longer than 400ms
+    useEffect(() => {
+        let timeoutId: NodeJS.Timeout;
+
+        if (isInitialLoadingLoginOptions) {
+            // Set a timer to show loading after 400ms
+            timeoutId = setTimeout(() => {
+                setShowDelayedLoading(true);
+            }, 400);
+        } else {
+            // Request completed, hide loading immediately
+            setShowDelayedLoading(false);
+        }
+
+        return () => {
+            if (timeoutId) {
+                clearTimeout(timeoutId);
+            }
+        };
+    }, [isInitialLoadingLoginOptions]);
+
     const { mutate, isLoading, isSuccess, isIdle } = useLoginWithEmailMutation({
         onSuccess: (data) => {
             identify({ id: data.userUuid });
@@ -116,7 +138,7 @@ const Login: FC<{}> = () => {
         }
     }, [isDemo, mutate, isIdle]);
 
-    const formStage = preCheckEmail ? 'login' : 'precheck';
+    const formStage = preCheckEmail && loginOptions ? 'login' : 'precheck';
 
     const isEmailLoginAvailable =
         loginOptions?.showOptions &&
@@ -136,6 +158,14 @@ const Login: FC<{}> = () => {
     }, [form.values, formStage, isEmailLoginAvailable, mutate]);
 
     const disableControls =
+        isInitialLoadingLoginOptions ||
+        loginOptionsLoading ||
+        (loginOptionsSuccess && loginOptions.forceRedirect === true) ||
+        isLoading ||
+        isSuccess;
+
+    const showButtonLoading =
+        showDelayedLoading ||
         loginOptionsLoading ||
         (loginOptionsSuccess && loginOptions.forceRedirect === true) ||
         isLoading ||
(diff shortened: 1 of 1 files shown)


joaoviana:
Build's failing: 

```
Type 'string' is not assignable to type 'OpenIdIdentityIssuerType'.ts(2322)
index.tsx(14, 5): The expected type comes from property 'providerName' which is declared here on type 'IntrinsicAttributes & { inviteCode?: string | undefined; intent?: "add" | "signin" | "signup" | undefined; providerName: OpenIdIdentityIssuerType; redirect?: string | undefined; } & ButtonProps'
```

packages/frontend/src/features/users/components/LoginLanding.tsx
```diff
@@ -92,6 +93,27 @@ const Login: FC<{}> = () => {
...
+            setShowDelayedLoading(false);
+        }
+
+        return () => {
+            if (timeoutId) {
+                clearTimeout(timeoutId);
+            }
+        };
+    }, [isInitialLoadingLoginOptions]);
+
```
joaoviana: Suggestion: 
We can use `useTimeout` from mantine/hooks

```
import { useTimeout } from '@mantine/hooks';
```

```suggestion
// Delayed loading state - only show loading if request takes longer than 400ms
    const { start: startDelayedLoading, clear: clearDelayedLoading } =
        useTimeout(() => setShowDelayedLoading(true), 400);

    useEffect(() => {
        if (isInitialLoadingLoginOptions) {
            // Start timer to show loading after 400ms
            startDelayedLoading();
        } else {
            // Request completed, hide loading immediately and clear timer
            setShowDelayedLoading(false);
            clearDelayedLoading();
        }
    }, [
        isInitialLoadingLoginOptions,
        startDelayedLoading,
        clearDelayedLoading,
    ]);
```

packages/frontend/src/features/users/components/LoginLanding.tsx
```diff
@@ -136,6 +158,14 @@ const Login: FC<{}> = () => {
     }, [form.values, formStage, isEmailLoginAvailable, mutate]);
 
     const disableControls =
+        isInitialLoadingLoginOptions ||
```
joaoviana: 👍

packages/frontend/src/features/users/components/LoginLanding.tsx
```diff
@@ -148,7 +178,20 @@ const Login: FC<{}> = () => {
...
+    );
+    useEffect(() => {
+        if (ssoOptions.length > 0) {
+            setPreservedSsoOptions(ssoOptions);
+        }
+    }, [ssoOptions]);
+
+    const displaySsoOptions =
+        ssoOptions.length > 0 ? ssoOptions : preservedSsoOptions;
+
```
joaoviana: I believe we don't need this at all. All we need is `ssoOptions` to be set-state-able

`ssoOptions` is memoed and computed whenever `loginOptions` changes. 

So remove this code, as well as `ssoOptions` `useMemo` and have this (from line 79-> 102)

```
    const {
        data: loginOptions,
        isInitialLoading: isInitialLoadingLoginOptions,
        isLoading: loginOptionsLoading,
        isSuccess: loginOptionsSuccess,
    } = useFetchLoginOptions({
        email: preCheckEmail,
    });
    const [ssoOptions, setSsoOptions] = useState<string[]>([]);

    // Disable fetch once it has succeeded
    useEffect(() => {
        if (loginOptions && loginOptionsSuccess) {
            if (loginOptions.forceRedirect && loginOptions.redirectUri) {
                window.location.href = loginOptions.redirectUri;
            }
            const ssoOptionsAvailable = loginOptions.showOptions.filter(
                isOpenIdIdentityIssuerType,
            );
            if (ssoOptionsAvailable.length > 0) {
                setSsoOptions(ssoOptionsAvailable);
            }
        }
    }, [loginOptionsSuccess, loginOptions]);
```

joaoviana (changes requested):
left comment suggestions and there's a build type fail

owlas:
thanks @joaoviana I updated this and simplified the logic, it was a bit of a mess. I've also removed the cloned state for sso options and using `keepPreviousData` instead on react query - this is basically so that the sso option doesn't weirdly disappear when the user first hits "continue" to load new sso options. 

The main goal here was for the page loading spinner not to show when pressing continue (which caused the page to kinda reload. That's still working. 

And I opted to add a 400ms debounce to the loading state from getting login optinos, because it's pretty fast most of the time and was a bit of a loading flash.

PR #15395, opened 2025-06-17
chore: ability to switch llm providers and models

IrakliJani:
Closes:


IrakliJani:
* **#15395** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15395?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15395?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

notgiorgi (approved):

packages/backend/src/ee/services/AiAgentService.ts
```diff
@@ -1109,8 +1102,13 @@ export class AiAgentService {
                 description: string;
             }>;
         }) => {
+            if (!this.lightdashConfig.ai.copilot.providers?.openai?.apiKey) {
+                throw new Error('OpenAI API key not found');
+            }
+
```
tatianainama: don't we need to check for both here? as you could only have anthropic enabled for example

tatianainama (commented):
I think I'm missing something, but this enables the possibility of switching providers, but it still defaults to gpt (doesn't work with anthropic if set for example)

packages/backend/src/ee/services/AiAgentService.ts
IrakliJani: good point, this is definitely incorrect. changing

IrakliJani: oh actually, this is an embedding model. claude does not have an embedding model so we are kinda stuck with it

IrakliJani: Nevertheless, this feature is currently not being used and is susceptible to changes.

tatianainama: > oh actually, this is an embedding model. claude does not have an embedding model so we are kinda stuck with it

ah got it! Maybe we could update the error to better describe and mention that this feature is only allowed for the openAi provider instead?
Btw, does the same applies to the updates of `generateExploreResult` ?

IrakliJani: updated!

re generateExploreResult, nope, its hard-coded at the moment , but it will change

IrakliJani:
### Merge activity

* **Jun 17, 3:14 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15395).
* **Jun 17, 3:14 PM UTC**: @IrakliJani merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15395).

Issue #15377, opened 2025-06-16
Support compact formatting for bytes


agha4to:
### Description

Support the following compact Formats:

- SI: KB, MB, GB, TB, PB
- IEC: KiB, MiB, GiB, TiB, PiB

### What problem does this solve?

These additional formats allow for easy application of accurate compact representations for bytes, dividing by multiples 1024 or 1000 depending on the desired byte standard. This avoids having to create columns and custom formats when analyzing large byte counts.

PR #15403, opened 2025-06-17
feat: SI and IEC byte formats

agha4to:
Closes: [15377](https://github.com/lightdash/lightdash/issues/15377)

### Description:
Support for compact byte formats:
- SI: KB, MB, GB, TB, PB
- IEC: KiB, MiB, GiB, TiB, PiB

Divides by multiples of 1000 or 1024 depending on the desired byte standard. Note that KiB, MiB, etc. which divide by 1024^n require special handling since standard formats are base 10.

**New Types**
image

**New Compact Formats**
image

**Example output: Formatted vs. raw value for KiB**
image

**Config menu also uses these values**
image


almeidabbm (approved):
LGTM 🚀 

Pushed a small change to add a `useMemo` for the compact value override
Failing e2e test is not related as it is also happening in `main`, fixing in a separate PR

Issue #12897, opened 2024-12-13
I would like a treemap chart type 


jpetey75:
### Description

Treemap visualization to chart types 

https://en.wikipedia.org/wiki/Treemapping

![image](https://github.com/user-attachments/assets/4dc964f5-bee0-46f3-8fad-73c6c39ed1df)

PR #15407, opened 2025-06-18
feat: adding a new treemap chart type

inverts:
Closes: https://github.com/lightdash/lightdash/issues/12897

### Description:

Adds support for a treemap chart type. Utilizes the eCharts treemap: https://echarts.apache.org/examples/en/index.html#chart-type-treemap 

image

### Generated Budgets Data 

Includes a new example data source, "Generated budgets". This is a randomly generated collection of 100 Teams, each belonging to a Department, each belonging to an Organization. Each Team has the following metrics:
- Total Employee Count (uniform 1-50)
- Total Budget (normal, mean=employee_count * 1000, stddev=employee_count*200)
- Total Budget Used (uniform, 0.7-1.2 * total_budget)
- Average Budget Use Percentage 

The data source also includes 
- one row for each Department with a null Team, summarizing the metrics above for the Department, and 
- one row for each Organization with a null Department and Team, summarizing the metrics above for the Organization

Ideally, the data can automatically be summarized in order to provide accurate parent group sizes. However, if it cannot, I wanted the treemap visualization to be able to understand a data format that provided it with the parent aggregated values. This is expressed with null children as described above.

image

#### Layout

There are three main sections for the **Layout**:
- Dimension Hierarchy: This is where the user orders the dimensions (drag n drop) to tell the treemap the hierarchy
- Size Metric: Which metric determines how large each block is?
- Color Metric: [optional] Which metric determines the color of the block?

By default, the color metric is disabled, and each block is given a color from the template based on the index of the block within its parent (eg. block colors are only unique at a per-level basis). Block colors should inherit the project-specified custom theme colors, but I'm not extremely familiar with this feature and only did a happy path test of this.

image

The **Auto-sum parent size** toggle is on by default and will automatically generate parent size values by summing all the rows for that parent. If it is off, the block sizes will be determined by a row where the child dimension of the parent is equal to null. If no such row exists, the size will be 0 and the parent and any descendants will not be visible (!).

image

The individual colors used can be specified - currently only a linear range between two colors is supported:
image

If no thresholds are specified, it automatically determines the range - **but only on a per-level basis**. This can be useful to see the "worst"/"best" node for each section, but it can also be deceiving if the user doesn't understand how it works. Take this example where Engineering is using 96% of the budget, which one might expect to be good, but it is bright red because it's the highest of all the departments in Tech.

image

#### Display

The second section has two options:

**Minimum Pixels Visible** - allows you to reduce the number of blocks by not rendering blocks smaller than this. Note that if you zoom in to the chart, the blocks become visible once they are large enough.

1,000sq px:
image

10,000 sq px:
image

**Max Leaf Depth** - How many levels deep we should show the hierarchy. If this is less than the total dimensions, the treemap "drilldown" mode is activated, and clicking a leaf at the deepest level will expose the children of that leaf.

image

### Known bugs, ideas for features, and observations:

I didn't want to go too far down the road of making decisions without consulting the team, so I started keeping a list of things that would be best to have some feedback or discussion on first - sometimes because they may be design decisions, and sometimes because the fix was not clear, so it may be helpful to understand priority before investing time in it.

- Rendering issues:
  - Bug: Initial graph zoom with results pane open does not show the whole treemap
  I think the chart renders as if the Results pane was collapsed, and then when the results pane is expanded on page load, the chart doesn’t re-render to the new bounds. 

https://github.com/user-attachments/assets/9759cb2e-c19b-4613-bd2b-30c27a93551f
  - Bug: Sometimes chart renders weird after drilling “out” (overlapping blocks), and a zoom change fixes it **-Edit: ** I'm able to reproduce this by zooming any arbitrary amount, and then collapsing/expanding a section of the page. Triggering zoom again fixes it. 

https://github.com/user-attachments/assets/0f5cf17a-c5c7-46c7-b87a-01989f7e920c
  - Bug: Drilling into a category with the results pane open sometimes doesn't produce breadcrumbs to allow you to drill out. Collapsing the results pane automatically drills out.
- Colors
  - ~~Feature: Auto-sum (or auto-aggregate if possible) color metric from leaf nodes~~
  - Feature: Option to show thresholds in chart as a legend
  - Misc: Color thresholds do not clamp values. For example, with data ranging -1 to 1, setting min to 0 and max to 0.001 will not yield binary colored nodes. Eg. even if the max color is red and the value is 0.1, the - color may not be red, only redd-ish.
  - Feature: Binary/discrete color rules
  - Feature: If not using static thresholds, user should be able to pick global min/max vs level min/max (default is level it seems)
  - Bug/Feature: Color or way to indicate when there is no data for the color metric for that node.
- Feature: Guess the order of the dimensions by frequency in the data
- Feature: Better user feedback/display for when there is no automatic aggregation and the parents have size 0, resulting in an empty chart.
- Feature: Option to show legend values on the nodes/blocks
- Feature: SQL Runner chart support
- Misc: Tooltip is nice but looks a little different from other chart tooltips, LD may want to format it differently
- Misc: Clicking a category with a header label only zooms to that category. If you drill into that category’s children, then click the category on the breadcrumb ux (at bottom), you get a fullscreen view of that - category (personally I prefer the latter view). There is an eCharts option that modifies this behavior
- ~~Bug: “Move” hand cursor doesn’t show up for dimension reordering until you move a dimension once (doesn’t have to reorder)~~
- Bug/Misc: Dimension order is not preserved when you change the dimensions (add/remove) in the query
- Misc: Reordering dimensions in the wrong order can make the chart blank, may not be clear to the user what is wrong.
- Misc: With some Min Pixels Visible values, the chart can render some intermediate nodes on their own, making them look like leaf nodes.
- ~~Misc: We should consider ignoring null dimension values when auto-sum parent size is on~~


.gitignore
```diff
@@ -78,7 +78,6 @@ lightdash
 
 # Prometheus
 prometheus.yml
-data/
```
inverts: This was preventing the new example data source CSV from being committed. I'm not sure if that was intentional, and removing this did not reveal any other files to me, but I'm not sure if it is meant to hide some prometheus related files that could be generated during development.

packages/frontend/src/hooks/useTreemapChartConfig.ts
```diff
@@ -0,0 +1,376 @@
...
+                                getEmptyTreemapNode(dimensionValueStr);
+                        }
+                        if (i === groupFieldIds.length - 1) {
+                            parent.children[dimensionValueStr].value = [
+                                rowSizeMetricValue,
+                                rowColorMetricValue,
+                            ];
+                        }
+                        if (shouldSumSizeMetric) {
+                            parent.value[0] += rowSizeMetricValue;
```
almeidabbm: This custom implementation doesn't support all metric types - because we're always **adding** children values to calculate the parent's value.

See this example with average:
Screenshot 2025-06-24 at 17 45 09

We used to have a similar issue with `subtotals` which was fixed by calculating the subtotals in the backend. I believe that endpoint might be used here instead of doing all the calculations in the frontend. Here's the PR that added the `subtotals` endpoint: https://github.com/lightdash/lightdash/pull/13847

The way it works is that you pass in the `metricQuery` and the `columnOrder` and the backend will run a query for each subgroup of dimensions leaving the last one in the `columnOrder` out as it is the "detailed" subgroup (matches results).

Then it returns all the results in a map of `key : rows`, where `key` is an identifier for the subgroup.

e.g.

consider the following metric query

```json
{
  "exploreName": "events",
  "dimensions": [
    "events_date_month",
    "events_date_week",
    "events_date_day",
    "events_event"
  ],
  "metrics": [
    "events_count"
  ],
  "filters": {},
  "sorts": [
    {
      "fieldId": "events_count",
      "descending": true
    }
  ],
  "limit": 500,
  "metricOverrides": {},
  "tableCalculations": [],
  "additionalMetrics": [],
  "customDimensions": []
}
```

and column order

```json
[
  "events_date_month",
  "events_date_week",
  "events_date_day",
  "events_event",
  "events_count"
]
```

the subtotals endpoint will return the following results:
```
{
  "events_date_month": [
    {
      "events_date_month": "2020-07-01T00:00:00Z",
      "events_count": "1467"
    },
    // ...rest
  ],
  "events_date_month:events_date_week": [
    {
      "events_date_month": "2020-07-01T00:00:00Z",
      "events_date_week": "2020-07-20T00:00:00Z",
      "events_count": "331"
    },
    // ...rest
  ],
  "events_date_month:events_date_week:events_date_day": [
    {
      "events_date_month": "2020-07-01T00:00:00Z",
      "events_date_week": "2020-07-27T00:00:00Z",
      "events_date_day": "2020-07-30T00:00:00Z",
      "events_count": "36"
    },
    // ...rest
  ]
}
```

There's a frontend hook that already takes care of calling the endpoint: `packages/frontend/src/hooks/useCalculateSubtotals.ts`

With this we're sure that all metric types will be supported in treemap charts respecting things like "average", etc 🙌

.gitignore
almeidabbm: You can remove this, commit the new file and then re-add `data/` to `.gitignore` so that nothing is lost 😄

packages/frontend/src/hooks/useTreemapChartConfig.ts
inverts: Thank you for all the details! I knew this was a shortcoming of my implementation but didn't realize we had something to address it. Hopefully I have implemented it as intended. I noticed the `pivotDimensions` were sometimes undefined in the table subtotal usage, and that supplying undefined there still produced the aggregate values I needed, so I hardcoded that to `undefined`. 
I was able to leverage the function to also update the color metric as well from my original comment:
> Feature: Auto-sum (or auto-aggregate if possible) color metric from leaf nodes

I have also updated the verbiage for the toggle + tooltip and updated the variables in the code to reflect the toggle's functionality.

image


I suspect there might be more conversation to be had about this functionality and what kind of user experience is desired. I think it most cases the auto-aggregation is probably enough, and there's some value in not over-complicating the options, but maybe there also still may be users who would want to supply their aggregate data in the null-dimension format.

almeidabbm: Just had another look at the PR and the `subtotals` implementation works super well!! 🚀 

About the values displayed in the `treemap` itself, I have found that these are not formatted at the moment - using just `value?.raw`

(check results table and `treemap` chart below)
image

Would it be possible to display the formatted string?

Regarding the `auto-aggregate` toggle, not sure if this is necessary 🤔 
I believe we should always `auto-aggregate`, would there be a scenario where users would be inclined to provide their own aggregated values? This could be possibly adding a bit of complexity. 
Checking with the team and will reply here 😄 

EDIT: after discussing with the team, we believe it would simplify the feature if we could just have it always `auto-aggregate` (using subtotals endpoint) rather than having an option that depends on having a specific data format (null-dimension format)

packages/backend/src/services/SavedChartsService/SavedChartService.ts
```diff
@@ -42,6 +42,7 @@ import {
 } from '@lightdash/common';
 import cronstrue from 'cronstrue';
 import { Knex } from 'knex';
+import { max } from 'lodash';
```
almeidabbm: Pls remove unused import

packages/frontend/src/components/VisualizationConfigs/TreemapConfig/TreemapDisplayConfig.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (!isTreemapVisualizationConfig(visualizationConfig)) return null;
+
+    const { visibleMin, setVisibleMin, leafDepth, setLeafDepth } =
+        visualizationConfig.chartConfig;
+
+    return (
+        <Stack>
+            <Config>
+                <Group>
+                    <Config.Heading>Minimum Pixels Visible</Config.Heading>
```
almeidabbm: Can we change to `Minimum section size`
Can we not use camel casing for the config titles? e.g. use `Minimum section size` instead of `Minimum Section Size`

**Note:** we understand that we are not consistent with this in some places, but we trying to get it more consistent

packages/frontend/src/components/VisualizationConfigs/TreemapConfig/TreemapDisplayConfig.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    </Tooltip>
+                    <NumberInput
+                        value={visibleMin}
+                        onChange={setVisibleMin}
+                        min={0}
+                        step={500}
+                        formatter={(value) => (value ? `${value}px\u00B2` : '')}
+                        parser={(value) => value.replace(/px\u00B2\s?$/, '')}
+                    />
+                    <Config.Heading>Max Leaf Depth</Config.Heading>
```
almeidabbm: Same here regarding camel case

packages/frontend/src/components/VisualizationConfigs/TreemapConfig/TreemapLayoutConfig.tsx
```diff
@@ -0,0 +1,293 @@
...
+                      </Draggable>
+                  );
+              });
+
+    return (
+        <Stack>
+            <Config>
+                <Config.Section>
+                    <Group spacing="xs">
+                        <Config.Heading>Dimension Hierarchy</Config.Heading>
```
almeidabbm: Camel case

almeidabbm (changes requested):
Hi @inverts 👋 
Sorry this is taking some time!

We just had another look and there's a couple other small changes added, mostly nitpicks
There's 2 things that are blocking this from getting merged:
1. Merge conflicts against `main`
2. The values displayed in the chart are the `raw` values and not the `formatted`

inverts:
No problem! All makes sense to me. Some things came up and wasn't able to make updates last week, but next week after Tuesday I'll be having a look 👍

packages/frontend/src/hooks/useTreemapChartConfig.ts
inverts: FYI, after removing it I had to decide how to interpret `null` values in the data.
I figured that it would be better to assume the user wants all their data displayed unless they filter it out from the query, so they are displayed in the visualization. Also, the subtotals calculation includes the null dimension values, so showing the null categories ensures that data adds up correctly.

image
image

packages/frontend/src/components/VisualizationConfigs/TreemapConfig/TreemapDisplayConfig.tsx
inverts: Updated all config labels accordingly:
image
image

packages/frontend/src/hooks/useTreemapChartConfig.ts
inverts: Updated screenshot after applying formatting:
image

packages/frontend/src/components/SimpleTreemap/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+    useEffect(() => {
+        // Load all the rows
+        resultsData?.setFetchAll(true);
+    }, [resultsData]);
+
+    useEffect(() => {
+        const listener = () => chartRef.current?.getEchartsInstance().resize();
+        window.addEventListener('resize', listener);
+        return () => window.removeEventListener('resize', listener);
+    });
```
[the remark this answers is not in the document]
inverts: This is the same as the pie chart example it was taken from, minimal impact as well. I started playing around with this `getEchartsInstance().resize()` and adding new listeners on the parent element to try and resolve some of the display issues enumerated in the PR description, but didn't have any luck. I think a change to this code could be done when a fix is identified for those issues.

almeidabbm:
@all-contributors add @inverts for code

PR #15450, opened 2025-06-20
fix: service account model no license

rephus:
Closes:

### DIFF
--- a/packages/backend/src/models/ModelRepository.ts
+++ b/packages/backend/src/models/ModelRepository.ts
@@ -1,5 +1,6 @@
 import { Knex } from 'knex';
 import { LightdashConfig } from '../config/parseConfig';
+import { ServiceAccountModel } from '../ee/models/ServiceAccountModel';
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/ModelRepository.ts
```diff
@@ -515,6 +516,16 @@ export class ModelRepository
...
 
+    public getServiceAccountModel(): ServiceAccountModel {
+        return this.getModel(
+            'serviceAccountModel',
+            () =>
+                new ServiceAccountModel({
+                    database: this.database,
+                }),
+        );
+    }
```
almeidabbm, line 527: This means we also don't need the override in `packages/backend/src/ee/index.ts:247` correct?
How about the `serviceAccountService`? this is also in ee, but seems like it is only used in `ee` so it's fine?

rephus: good catch, I'll remove it from ee. 
ServiceAccount is ok in ee

Issue #15054, opened 2025-05-28
Support to use Application Default Credentials(ADC) for BigQuery connection


to-lz1:
### Description

This issue proposes adding an option to use [ADC](https://cloud.google.com/docs/authentication/application-default-credentials) when we create a new BigQuery connection from the Web UI.

### What problem does this solve?

Currently, creating a BigQuery connection from the Web UI requires a service account key file. 
This method is less secure and adds operational overhead. On compute platforms like  Cloud Run or Google Compute Engine, 
it's possible to use the environment's service account credentials directly without generating a key. 
By supporting ADC, we can enhance security and reduce operational effort.

#### External Design

- Add an `AUTH_ENABLE_GCLOUD_ADC` environment variable.
  - If this variable is set to `true`, we add `Application Default Credentials` option to BigQuery authentication type dropdown.

- If `Application Default Credentials` is selected, users can create a new project without uploading a key file.

- The connection test will fail if ADC is not configured correctly on the environment

#### Internal Design

- During project creation or update, if the Application Default Credentials method is selected, the backend will validate that the `AUTH_ENABLE_GCLOUD_ADC` variable is enabled.

- Subsequently, sensitive credential fields for the project will be cleared.
  - This prevents retaining credentials from the previous private key method when updating the authentication type.

- During the dbt connection test, the profile will be written using the oauth method.
  - ref: https://docs.getdbt.com/docs/core/connect-data-platform/bigquery-setup#service-account-impersonation
  - With this setting, dbt (and google-auth-library, more specifically) will search for ADC.

### Non Goal

- While service account impersonation would offer stricter security, supporting it is out of scope for this issue.
- Lightdash CLI seems to have supported `oauth` method([ref](https://github.com/lightdash/lightdash/blob/41d4dcbc5bd546c73e7884e4b3cb710e287a77fb/packages/cli/src/dbt/targets/Bigquery/index.ts#L85)). Therefore, we need no change to CLI in this issue.

-----

If this proposal is accepted, I'm happy to submit a PR!

to-lz1:
As we discussed on community Slack, I agree that we should finish https://github.com/lightdash/lightdash/issues/13720 first.

Then we can add a new option “ADC” to the newly implemented input field “Authentication Type”.

ref: https://github.com/lightdash/lightdash/pull/15026

PR #15476, opened 2025-06-23
feat: add Application Default Credentials auth method for BigQuery

to-lz1:
Closes: #15054

### Description:

This pull request introduces support for BigQuery Application Default Credentials (ADC) authentication.

### Configuration Changes

* Added `enableGCloudADC` to the `AuthGoogleConfig` type and updated `parseConfig` to parse the `AUTH_ENABLE_GCLOUD_ADC` environment variable. [[1]](diffhunk://#diff-3e67382c265162c53535687712b5435a6f262f6917f02a313942c772d355e4e0R725) [[2]](diffhunk://#diff-3e67382c265162c53535687712b5435a6f262f6917f02a313942c772d355e4e0R1037)
* Added `AUTH_ENABLE_GCLOUD_ADC` as an environment variable in `docker/docker-compose.dev.yml` to enable ADC authentication.
* Provided a commented-out line for mounting Google Cloud credentials in the development container.

### Frontend Changes

* Added `useHealth` hook to check if ADC is enabled and dynamically updated the authentication type options in `BigQueryForm`. [[1]](diffhunk://#diff-00c89f9b26c65ebe71d8205edfffed5b82853827d1a295d20aed9442eb7e88eeR25) [[2]](diffhunk://#diff-00c89f9b26c65ebe71d8205edfffed5b82853827d1a295d20aed9442eb7e88eeR121-R122) [[3]](diffhunk://#diff-00c89f9b26c65ebe71d8205edfffed5b82853827d1a295d20aed9442eb7e88eeR180-R197)

### Backend Changes

* Added `BigqueryAuthenticationType.ADC` to the list of supported authentication types.
* Updated `credentialsTarget` in `packages/backend/src/dbt/profiles.ts` to handle ADC authentication.
* Modified `mergeMissingWarehouseSecrets` in `packages/backend/src/models/ProjectModel/ProjectModel.ts` to account for ADC authentication, which does not require credentials.

### Screenshot

image


packages/backend/src/dbt/profiles.ts
```diff
@@ -26,31 +27,47 @@ const credentialsTarget = (
...
+            if (
+                credentials.authenticationType ===
+                BigqueryAuthenticationType.ADC
+            ) {
+                // With oauth method and no keyfile contents, dbt will use the
+                // application default credentials (ADC) to authenticate
+                bqResult.target.method = 'oauth';
+                return bqResult;
+            }
+            throw new Error(
```
rephus: Normally we do `switch` and `assertUnreachable` on default, this will also helps us during compile, rather than throwing runtime errors.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1211,9 +1211,9 @@ export class ProjectService extends BaseService {
     }
 
     /* When editing a project, most fields are optional
-    but if the user switches from one authentication type to another, 
+    but if the user switches from one authentication type to another,
     we need to validate the secrets are present */
-    static validateConfigSecrets(project: UpdateProject) {
+    validateConfigSecrets(project: UpdateProject) {
```
rephus, line 1216: 👍

packages/warehouses/src/warehouseClients/BigqueryWarehouseClient.ts
```diff
@@ -530,7 +530,7 @@ export class BigqueryWarehouseClient extends WarehouseBaseClient<CreateBigqueryC
             )}`,
         );
         return new WarehouseQueryError(
-            `Bigquery warehouse error: ${error?.reason}`,
+            `Bigquery warehouse error: ${error?.reason} - ${error?.message}`,
```
rephus: we intentionally didn't include `error?.message` in case that returned some private info. Could you remove it ? 
This info is already returned on a console.error, so it can be reviewed on the backend logs

rephus (changes requested):
Code looks amazing, my only blocker is the error.message . The other changes are optional

packages/backend/src/dbt/profiles.ts
to-lz1: @rephus Thank you for your comment!

I noticed that we need to handle `undefined` case, which may happen when we 'test & deploy' old BigQuery project without any update. For now, I've decided to treat this case the same way as private key. Please let me know if you have a different opinion.

rephus (approved):

to-lz1:
@rephus 

> When running tests, these failed, can you fix them ?

I've fixed it, and confirmed that the backend unit test to pass on my machine!

rephus (approved):
Tests are running fine, tested it locally. It should not have any impact to any custoemrs if this ENV variables is disabled.

PR #15502, opened 2025-06-23
fix: validate undefined filter groups

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/15501

### Description:
Fix filter rule flattening logic to handle undefined filter groups. The previous implementation wasn't correctly accumulating filter rules from nested groups and didn't handle the case where a filter group could be undefined. This change refactors the `flattenFilterGroup` function to use `flatMap` for cleaner code and adds explicit handling for undefined filter groups.

Root cause for the filters to have been saved incorrectly hasn't been found, but this fixes the validation problems

### DIFF
--- a/packages/common/src/types/filter.ts
+++ b/packages/common/src/types/filter.ts
@@ -242,20 +242,22 @@ export const isMetricFilterTarget = (
 export const getFilterRules = (filters: Filters): FilterRule[] => {
     const rules: FilterRule[] = [];
     const flattenFilterGroup = (filterGroup: FilterGroup): FilterRule[] => {
-        const groupRules: FilterRule[] = [];
-
-        (isAndFilterGroup(filterGroup)
+        // Explicitly checking for undefined filter groups (and || or), saved filter group somehow was undefined when saving
+        const groupItems: FilterGroupItem[] | undefined = isAndFilterGroup(
+            filterGroup,
+        )
             ? filterGroup.and
-            : filterGroup.or
-        ).forEach((item) => {
+            : filterGroup.or;
+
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#15502** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15502?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15502?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ZeRego (approved):

ZeRego:
@almeidabbm Did you find where the `and`/`or` was being set as `undefined`? Typescript type shows it should not be possible so even though we can fix this place, it may happen elsewhere. 

```
type OrFilterGroup = {
    id: string;
    or: Array<FilterGroup | FilterRule>;
};

type AndFilterGroup = {
    id: string;
    and: Array<FilterGroup | FilterRule>;
};
```

almeidabbm:
> @almeidabbm Did you find where the `and`/`or` was being set as `undefined`? Typescript type shows it should not be possible so even though we can fix this place, it may happen elsewhere.
> 
> ```
> type OrFilterGroup = {
>     id: string;
>     or: Array<FilterGroup | FilterRule>;
> };
> 
> type AndFilterGroup = {
>     id: string;
>     and: Array<FilterGroup | FilterRule>;
> };
> ```

I time-boxed reproducing this, but haven't been able to, it's a weird state for it to be in 🤔 merging as it should fix the issue when validating, but keeping an eye out in case this happens again

PR #15569, opened 2025-06-25
refactor: use database timestamp functions and consolidate response update methods

IrakliJani:
Closes: 

### Description:
This PR refactors the AI agent model to use database functions for timestamps instead of JavaScript Date objects. This ensures that timestamps are consistently set by the database server rather than the application.

Key changes:
- Replaced `new Date()` instances with `trx.fn.now()` or `database.fn.now()` for more accurate timestamp handling
- Fixed TypeScript issues with proper type annotations for database queries
- Consolidated `updateWebAppResponse` into the existing `updateModelResponse` method to reduce code duplication
- Added AiAgentUserPreferences to knex tables definition
- Improved the message construction in the getThreadMessages method

This change improves timestamp consistency and reduces redundant code in the AI agent implementation.


IrakliJani:
* **#15571** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15571?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15570** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15569** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/models/AiAgentModel.ts
```diff
@@ -378,7 +376,8 @@ export class AiAgentModel {
                 const [result] = await trx(AiAgentInstructionVersionsTableName)
                     .insert({
                         ai_agent_uuid: agent.ai_agent_uuid,
-                        created_at: updatedAt,
+                        // @ts-expect-error - knex types are not correct
+                        created_at: trx.fn.now(),
```
tatianainama: is this going to be the same `now` as the one from agent's `updated_at` value? [Docs](https://knexjs.org/guide/utility.html#now) don't say anything about this, which makes me think it's just a `now` utility
Just asking because this will have a slightly different value than the one from updated.

packages/backend/src/ee/models/AiAgentModel.ts
```diff
@@ -286,7 +285,8 @@ export class AiAgentModel {
             if (args.instruction) {
                 await trx(AiAgentInstructionVersionsTableName).insert({
                     ai_agent_uuid: agent.ai_agent_uuid,
-                    created_at: createdAt,
+                    // @ts-expect-error - knex types are not correct
+                    created_at: trx.fn.now(),
```
tatianainama: I think we should update the Db types that use Date to also be `Knex.Raw`
so we won't need to bypass ts
something like:

```ts
type DatabaseTimestamp = Date | Knex.Raw;

export type DbAiAgentInstructionVersions = {
    ai_agent_instruction_version_uuid: string;
    ai_agent_uuid: string;
    instruction: string;
    created_at: DatabaseTimestamp;
};
```

wdyt?

tatianainama (approved):
left some comments but nothing blocking!

packages/backend/src/ee/models/AiAgentModel.ts
IrakliJani: this is going to use PG's NOW() function instead of js new Date()
do they have to match?

tatianainama: it's not required that they match, but from previous experience, it's nice to have all the same timestamps when updating values in the db, specially when debugging incidents in one year 😅

IrakliJani: good point, let me check if transaction is taking care of that (I don't speak good DB)

IrakliJani: @tatianainama this is so cool!

packages/backend/src/ee/models/AiAgentModel.ts
IrakliJani: yeah, good suggestion, I'll update it now!

IrakliJani: but it should be only for insert and update types

packages/backend/src/ee/models/AiAgentModel.ts
tatianainama: niiiice then we are good to go!

packages/backend/src/ee/models/AiAgentModel.ts
tatianainama: right, the return type is the same as now 🙏

IrakliJani:
### Merge activity

* **Jun 26, 2:16 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15569).
* **Jun 26, 2:17 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15569) rebased this pull request as part of a merge.
* **Jun 26, 2:18 PM UTC**: @IrakliJani merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15569).

PR #15576, opened 2025-06-26
feat: Extract JWT util

stevelikesmusic:
Closes: #15538

### Description:
Extracts JWT utility functions from EmbedService for encoding and decoding embed JWT tokens. The implementation includes:

- `encodeJwt`: Creates signed JWT tokens with configurable expiration
- `decodeJwt`: Validates and decodes JWT tokens with proper error handling

Added comprehensive test coverage for both functions, including tests for:
- Valid token encoding/decoding
- Expired token handling
- Invalid token structure
- Error cases with malformed secrets

PR #15579, opened 2025-06-26
feat: Use JWT Middleware in EmbedController

stevelikesmusic:
Closes:


stevelikesmusic:
* **#15648** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15599** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15579** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15579?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/types/auth.ts
```diff
@@ -6,3 +10,19 @@ export enum AuthTokenPrefix {
...
     PERSONAL_ACCESS_TOKEN = 'ldpat_',
 }
+
+/**
+ * The account object is used to store the account information for the user.
+ * It's meant to be agnostic of the authentication method.
+ *
+ * AuthData defaults to EmbedJwt for now as we learn more.
+ */
+export type Account<AuthData extends EmbedJwt | Record<string, unknown>> = {
```
owlas: would it be simpler to use if we had multiple fields instead of type templating?

packages/common/src/types/auth.ts
```diff
@@ -6,3 +10,19 @@ export enum AuthTokenPrefix {
...
+    organization: Pick<Organization, 'organizationUuid'>;
+    authentication: {
+        data: AuthData;
```
owlas: What would be here in the case of `Record<string,unknown>`?

packages/backend/src/middlewares/jwtAuthMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export async function jwtAuthMiddleware(
+    req: Request,
+    res: Response,
+    next: NextFunction,
+): Promise<void> {
+    try {
+        // Extract project UUID from URL path since middleware is applied to /api/v1/embed path
+        // The path will be like /api/v1/embed/{projectUuid}/dashboard
+        const projectUuid = parseProjectUuid(req.path);
```
owlas, line 47: this feels like we're throwing away strict tsoa typing for path parameters in order to set the project value earlier, what's the reason that we need this here? Can the auth middleware function without knowledge of the project?

packages/backend/src/middlewares/jwtAuthMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        const embedService = req.services.getEmbedService() as EmbedService;
+        if (!embedService) {
+            Logger.error('Services are not initialized for embed token');
+            next();
+            return;
+        }
+
+        // Get embed configuration from database
+        const { encodedSecret, organization } =
```
owlas, line 69: is the problem that decoding the token requires a secret scoped to projects? Will this be different once embedding works like service accounts?

packages/common/src/types/auth.ts
stevelikesmusic: I suppose it could be. I was trying to avoid encoding JWT specific concerns in the type and use generics to give flexibility.

packages/common/src/types/auth.ts
stevelikesmusic: I was unsure. I used the generic for now so as not to make for a harder refactor later if we added something. The goal was to avoid having something like `jwtData?` as a field in the type.

packages/backend/src/middlewares/jwtAuthMiddleware.ts
stevelikesmusic: We need the project to get the embedding which has the JWT secret. I don't think we're throwing away tsoa typing because this is running before routes are applied. The middleware functions differently than the route handlers. 

A better long-term approach would probably be to sign JWTs with a consistent key set as an ENV_VAR. And then encode the current secret as a field in the payload (one key for signing the JWT and one key for validating the projectId). The two secrets would perform two functions:

1. Verify that we have a valid JWT (no trip to the DB needed)
2. Confirm that this JWT is valid for the given projectId (go to the DB for this one)

packages/backend/src/middlewares/jwtAuthMiddleware.ts
stevelikesmusic: Yes, the JWT secret is bound to the embedding. No, service accounts won't fix this. 

I think we should have [two](https://github.com/lightdash/lightdash/pull/15579/files#r2172161893) as we have two concerns: signing and validating the project + JWT.

packages/common/src/types/auth.ts
owlas: So this is primarily a container for the SessionUser + active organization?

everything else is just metadata?

packages/backend/src/ee/controllers/embedController.ts
```diff
@@ -175,15 +174,15 @@ export class EmbedController extends BaseController {
     @Post('/dashboard')
     @OperationId('getEmbedDashboard')
     async getEmbedDashboard(
+        @Request() req: express.Request,
         @Path() projectUuid: string,
-        @Header('Lightdash-Embed-Token') embedToken: string,
```
owlas, line 176: this would also remove this from the documentation for the endpoint (in open ai docs) is the goal to eventually use the `@Security` decorator from tsoa?

packages/backend/src/middlewares/jwtAuthMiddleware.ts
owlas: Could we also fix this by having the embed tokens sit at the org level?

If the embed tokens become synonymous with service accounts (aka oauth clients) then they will probably move to the org level right? And then we don't have this dependency.

I don't love the approach here, can we document that it's temporary and the reason we have to do this is because JWT keys are signed with project level keys

owlas: Could we also fix this by having the embed tokens sit at the org level?

If the embed tokens become synonymous with service accounts (aka oauth clients) then they will probably move to the org level right? And then we don't have this dependency.

I don't love the approach here, can we document that it's temporary and the reason we have to do this is because JWT keys are signed with project level keys

packages/backend/src/ee/services/EmbedService/EmbedService.ts
```diff
@@ -316,13 +314,14 @@ export class EmbedService extends BaseService {
 
     async getDashboard(
         projectUuid: string,
-        embedToken: string,
+        account: Account<EmbedJwt>,
```
owlas: can you explain how the controller knows the type is `Account<EmbedJwt>` and not `Account<Record<string,unknown>>`? I couldn't see the narrowing anywhere!

owlas (approved):
direction looks good, just need to fix the build and can you answer my qusetion about type narrowing in the controller?

packages/common/src/types/auth.ts
stevelikesmusic: Those are the two big ones. It also will include permissions associated with the JWT (dashboard and available filters)—and ideally service account.

packages/backend/src/middlewares/jwtAuthMiddleware.ts
stevelikesmusic: > Could we also fix this by having the embed tokens sit at the org level?

We could. I assumed that we wanted a secret mapped specifically to an embed, but I suppose the embed ID would work just as well. 

Was there a reason to make the embedding secret be the JWT secret rather than have it be a top-level/org secret?

packages/backend/src/ee/controllers/embedController.ts
stevelikesmusic: Ah that's a great catch! Would adding the header decorator to the class indicate all routes require the header? What if we just added JSDocs to explain these endpoints require the JWT? Long-term, I think we want to deprecate this controller altogether and have embeds talk to standard endpoints. 

I think the goal is we should use the `@Security` decorator when scopes are ready—although, we could go with roles sooner just as easily and for simplicity. That's a separate concern in my mind (authorization vs. authentication). Right now these endpoints use the Token for both `who am I` and `what can I do`. The JWT middleware parses that out so that when requests hit controllers, those distinctions have already been made.

packages/backend/src/middlewares/jwtAuthMiddleware.ts
owlas: no I think it got tied to the idea that embedding is enabled at a project level

packages/backend/src/middlewares/jwtAuthMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+        // The path will be like /api/v1/embed/{projectUuid}/dashboard
+        const projectUuid = parseProjectUuid(req.path);
+
+        if (!projectUuid) {
+            next();
+            return;
+        }
+        req.project = { projectUuid };
+
+        const embedToken = req.headers[JWT_HEADER_NAME] as string;
```
owlas: Don't force typing here, you're checking below for undefined anyway. just handle the cases - it'll be a `string | string[] | undefined`

packages/backend/src/ee/services/EmbedService/EmbedService.ts
stevelikesmusic: For this case right now, `EmbedJwt` is the only variant. We could narrow the type in the controller though since these endpoints (and the services) can only operate with a token. Thinking about it more now, perhaps the better solution is more well defined union types in Account rather than the generic 🤔

packages/backend/src/middlewares/jwtAuthMiddleware.ts
stevelikesmusic: I'll leave a TODO here and create a ticket to move the secret to the org level. 

Do you imagine we'd continue to:
1. Show the secret on the Embed page (I think yes)
2. Allow secret rotation from the Embed (we'd need to callout that this would change the secret for all embeds in the org, not just this project)

packages/backend/src/middlewares/jwtAuthMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+        req.account = {
+            authentication: {
+                type: 'embed',
+                data: decodedToken,
+                source: embedToken,
+            },
+            organization,
+            // FIXME: This is temporary until we parse user and abilities
+            user: {} as SessionUser,
+        } as Account<EmbedJwt>;
```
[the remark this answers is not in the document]
stevelikesmusic: Will address this is union types rather than generics

packages/backend/src/middlewares/jwtAuthMiddleware.ts
stevelikesmusic: https://github.com/lightdash/lightdash/issues/15661

packages/common/src/types/auth.ts
stevelikesmusic: Got rid of the generics

packages/common/src/types/auth.ts
stevelikesmusic: Handled with the new typing

packages/backend/src/ee/controllers/embedController.ts
stevelikesmusic: @owlas these routes are actually not even documented in the API docs. The whole controller is annotated with `@Hidden`

packages/backend/src/ee/services/EmbedService/EmbedService.ts
stevelikesmusic: Updated the types and using a type guard function to discriminate the union

packages/backend/src/ee/controllers/embedController.ts
owlas: Great spot. Let's leave it then!

packages/common/src/types/auth.ts
```diff
@@ -6,3 +20,41 @@ export enum AuthTokenPrefix {
...
+ * The account object is used to store the account information for the user.
+ * It's meant to be agnostic of the authentication method.
+ */
+export type Account = {
+    organization: Pick<Organization, 'organizationUuid'>;
+    authentication: LightdashAccountAuth;
+    user: SessionUser;
+};
+
+export function getEmbeddedAuth(account: Account): EmbeddedAccount {
```
owlas: Could also be a typeguard:

```
isEmbeddedAccountAuth(auth: LightdashAccountAuth): auth is EmbeddedAccount {
  ...
};
```

owlas: this allows you to use this in places where you don't want to throw error and narrow types

packages/common/src/types/auth.ts
```diff
@@ -6,3 +20,41 @@ export enum AuthTokenPrefix {
...
+    type: 'session';
+    source: string;
+}>;
+
+type ServiceAccount = LightdashAuth<{
+    type: 'service-account';
+    source: string;
+}>;
+
+type LightdashAccountAuth =
```
owlas, line 46: More consistent naming for these variables like `EmbeddedAccountAuth` - `LightdashUserAccountAuth` - `ServiceAccountAuth`

owlas (approved):
looks good - couple non blocking comments

packages/common/src/types/auth.ts
stevelikesmusic: What if the typeguard function evaluates `false`? Would we not throw?

Issue #15347, opened 2025-06-15
Using CLI with `--skip-dbt-compile` fails without working profiles.yml


owlas:
### Description

When the lightdash cli is used to compile with `lightdash compile --skip-dbt-compile` it should read the existing `target/manifest.json` to compile the lightdash metadata.

This shouldn't need a valid `profiles.yml` setup, since we only need to read in the `target/manifest.json`. This is because the lightdash compile needs a full `WarehouseClient` with valid credentials. This dependency isn't necessary and it should be possible to run:

`lightdash compile --skip-warehouse-catalog --skip-dbt-compile` without any valid credentials. In fact it should be usable with only the manifest.json

### Steps to Reproduce the Bug or Issue

1. Get a dbt project
2. Have broken or non-existent profiles.yml
3. Have a valid manifest.json
4. Then run `lightdash compile --skip-warehouse-catalog --skip-dbt-compile`
5. You will see an error

### Cloud or self-hosting

self-hosting

PR #15593, opened 2025-06-27
chore: compile explores without credentials

owlas:
Working towards #15347

This makes `WarehouseSqlBuilder` a drop in replacement for `WarehouseClient` when the credentials aren't required.

I've replaced the type in the `compileExplore` and `ExploreCompiler` so that the CLI can start using them to compile projects without needing warehouse credentials


owlas:
* **#15595** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15594** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15594?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15593** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15593?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15593?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/warehouses/src/warehouseClients/WarehouseBaseSqlBuilder.ts
```diff
@@ -10,8 +10,6 @@ import { getDefaultMetricSql } from '../utils/sql';
 export default abstract class WarehouseBaseSqlBuilder
     implements WarehouseSqlBuilder
 {
-    abstract readonly type: WarehouseTypes;
```
owlas, line 13: This was causing trouble because it doesn't exist on `WarehouseClient` plus it's not needed!

packages/warehouses/src/warehouseClients/WarehouseBaseSqlBuilder.ts
```diff
@@ -24,6 +22,10 @@ export default abstract class WarehouseBaseSqlBuilder
 
     abstract getAdapterType(): SupportedDbtAdapter;
 
+    getFieldQuoteChar(): string {
```
owlas, line 25: By moving this onto the clients, consuming code doesn't need to determine the warehouse type to get the field quote char

magnew (approved):
This code is pretty straightforward -- what's the use case for running `lightdash compile --skip-warehouse-catalog --skip-dbt-compile`? Just updating LD metadata without changing data/models? Is that for tooling? Nothing dangerous can happen?

owlas:
@magnew so it actually does update the table definitions in lightdash, it just assumes the user has previously run `dbt complie` already. The lightdash CLI wraps the dbt CLI and exposes most functionality but we've had a couple of edge cases recently where the lightdash CLI didn't have the features the user needed.

So this flow allows the user to run `dbt compile` in their own special way (producing the `manifest.json`) and then lightdash can pick up the work from there to translate into Lightdash objects and push to LD servers

owlas:
### Merge activity

* **Jun 30, 8:14 PM UTC**: @owlas merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15593).

Issue #15580, opened 2025-06-26
Apply CASL abilities based on permissions from the JWT and limited access as a viewer


stevelikesmusic:
Now that we have middleware parsing the JWT and adding to the request, we need to create CASL abilities based on the embedding and JWT capabilities.

PR #15599, opened 2025-06-27
feat: hydate req.account in jwt middleware

stevelikesmusic:
Closes: #15580

### Description:
Refactors the JWT authentication middleware to improve embedded dashboard functionality. This PR:

- Introduces a new `hydrateEmbeddedAccount` utility to properly populate user session data and abilities
- Adds CASL ability rules for embedded users with proper scoping
- Replaces `EmbedJwt` with `CreateEmbedJwt` type in relevant places due to zod and tsoa not being on friendly terms
- Enhances the embed model to include organization name
- Extracts dashboard UUID resolution logic to a dedicated method
- Adds comprehensive test coverage for embedded user abilities
- Slight refactor to CASL verbs and subjects so their sorted for easier reading

This change improves the security model for embedded dashboards by properly scoping user permissions based on the JWT token contents.

### DIFF
--- a/packages/backend/src/auth/lightdashJwt.ts
+++ b/packages/backend/src/auth/lightdashJwt.ts
@@ -1,7 +1,6 @@
 /* eslint-disable */
 import {
     CreateEmbedJwt,
-    EmbedJwt,
     EmbedJwtSchema,
     ForbiddenError,
     getErrorMessage,
@@ -45,7 +44,7 @@ export function encodeLightdashJwt(
 export function decodeLightdashJwt(
     token: string,
     encodedSecret: string | Buffer,
-): EmbedJwt {
+): CreateEmbedJwt {
     try {
         const encryptionUtil = new EncryptionUtil({ lightdashConfig });
         const secret = encryptionUtil.decrypt(
@@ -53,7 +52,7 @@ export function decodeLightdashJwt(
                 ? encodedSecret
                 : Buffer.from(encodedSecret),
         );
-        const decodedToken = verify(token, secret) as EmbedJwt;
+        const decodedToken = verify(token, secret) as CreateEmbedJwt;
 
         // Alert if the token is not in the expected format so we can inform the org before enforcing validation
         try {
--- a/packages/backend/src/ee/controllers/embedController.ts
+++ b/packages/backend/src/ee/controllers/embedController.ts
@@ -5,6 +5,7 @@ import {
     ApiCalculateTotalResponse,
     ApiErrorPayload,
     ApiSuccessEmpty,
+    assertEmbeddedAuth,
     CacheMetadata,
     CreateEmbed,
     CreateEmbedJwt,
@@ -16,8 +17,6 @@ import {
     EmbedUrl,
     Explore,
     FieldValueSearchResult,
-    FilterInteractivityValues,
-    ForbiddenError,
     Item,
     MetricQueryResponse,
     SavedChart,
@@ -176,9 +175,9 @@ export class EmbedController extends BaseController {
         @Path() projectUuid: string,
     ): Promise<ApiEmbedDashboardResponse> {
         this.setStatus(200);
-        if (!req.account) {
-            throw new ForbiddenError('Account is missing');
-        }
+
+        assertEmbeddedAuth(req.account);
+
         return {
             status: 'ok',
             results: await this.getEmbedService().getDashboard(
@@ -197,9 +196,9 @@ export class EmbedController extends BaseController {
         @Body() body: SavedChartsInfoForDashboardAvailableFilters,
     ): Promise<ApiEmbedDashboardAvailableFiltersResponse> {
         this.setStatus(200);
-        if (!req.account) {
-            throw new ForbiddenError('Account is missing');
-        }
+
+        assertEmbeddedAuth(req.account);
+
         return {
             status: 'ok',
             results:
@@ -226,9 +225,9 @@ export class EmbedController extends BaseController {
         },
     ): Promise<ApiEmbedChartAndResultsResponse> {
         this.setStatus(200);
-        if (!req.account) {
-            throw new ForbiddenError('Account is missing');
-        }
+
+        assertEmbeddedAuth(req.account);
+
         return {
             status: 'ok',
             results: await this.getEmbedService().getChartAndResults(
@@ -256,9 +255,9 @@ export class EmbedController extends BaseController {
         },
     ): Promise<ApiCalculateTotalResponse> {
         this.setStatus(200);
-        if (!req.account) {
-            throw new ForbiddenError('Account is missing');
-        }
+
+        assertEmbeddedAuth(req.account);
+
         return {
             status: 'ok',
             results: await this.getEmbedService().calculateTotalFromSavedChart(
@@ -292,9 +291,7 @@ export class EmbedController extends BaseController {
         this.setStatus(200);
         const { search, limit, filters, forceRefresh } = body;
 
-        if (!req.account) {
-            throw new ForbiddenError('Account is missing');
-        }
+        assertEmbeddedAuth(req.account);
 
         const results = await this.getEmbedService().searchFilterValues({
             account: req.account,
--- a/packages/backend/src/ee/models/EmbedModel.ts
+++ b/packages/backend/src/ee/models/EmbedModel.ts
@@ -15,7 +15,17 @@ export class EmbedModel {
         const [embed] = await this.database('embedding')
             .select()
             .leftJoin('users', 'embedding.created_by', 'users.user_uuid')
-            .where('project_uuid', projectUuid);
+            .leftJoin(
+                'projects',
+                'projects.project_uuid',
+                'embedding.project_uuid',
+            )
+            .leftJoin(
+                'organizations',
+                'organizations.organization_id',
+                'projects.organization_id',
+            )
+            .where('embedding.project_uuid', projectUuid);
 
         if (!embed) {
             throw new NotFoundError(
@@ -41,7 +51,8 @@ export class EmbedModel {
         return {
             projectUuid: embed.project_uuid,
             organization: {
-                organizationUuid: embed.user_organization_uuid,
+                organizationUuid: embed.organization_uuid,
+                name: embed.organization_name,
             },
             encodedSecret: embed.encoded_secret,
             dashboardUuids: validDashboardUuids,
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -1,6 +1,5 @@
 import { subject } from '@casl/ability';
 import {
-    Account,
     addDashboardFiltersToMetricQuery,
     AndFilterGroup,
     CommercialFeatureFlags,
@@ -17,13 +16,13 @@ import {
     EmbedUrl,
     Explore,
     ExploreError,
+    ExternalAccount,
     FieldValueSearchResult,
     FilterableDimension,
     ForbiddenError,
     formatRows,
     getDashboardFiltersForTileAndTables,
     getDimensions,
-    getEmbeddedAuth,
     getFilterInteractivityValue,
     getItemId,
     InteractivityOptions,
@@ -289,7 +288,7 @@ export class EmbedService extends BaseService {
         if (!isEnabled.enabled) throw new ForbiddenError('Feature not enabled');
     }
 
-    private async getDashboardUuidFromContent(
+    async getDashboardUuidFromJwt(
         decodedToken: CreateEmbedJwt,
         projectUuid: string,
     ) {
@@ -315,15 +314,15 @@ export class EmbedService extends BaseService {
 
     async getDashboard(
         projectUuid: string,
-        account: Account,
+        account: ExternalAccount,
         // TODO: WHY IS THIS OPTIONAL??
         checkPermissions: boolean = true,
     ): Promise<Dashboard & InteractivityOptions> {
         const { data: decodedToken, source: embedToken } =
-            getEmbeddedAuth(account);
+            account.authentication;
         const { dashboardUuids, allowAllDashboards, user } =
             await this.embedModel.get(projectUuid);
-        const dashboardUuid = await this.getDashboardUuidFromContent(
+        const dashboardUuid = await this.getDashboardUuidFromJwt(
             decodedToken,
             projectUuid,
         );
@@ -398,11 +397,11 @@ export class EmbedService extends BaseService {
 
     async getAvailableFiltersForSavedQueries(
         projectUuid: string,
-        account: Account,
+        account: ExternalAccount,
         savedChartUuidsAndTileUuids: SavedChartsInfoForDashboardAvailableFilters,
         checkPermissions: boolean = true,
     ): Promise<DashboardAvailableFilters> {
-        const { data: decodedToken } = getEmbeddedAuth(account);
+        const { data: decodedToken } = account.authentication;
         const { dashboardUuids, allowAllDashboards } =
             await this.embedModel.get(projectUuid);
 
@@ -423,7 +422,7 @@ export class EmbedService extends BaseService {
             filters: CompiledDimension[];
         }[] = [];
 
-        const dashboardUuid = await this.getDashboardUuidFromContent(
+        const dashboardUuid = await this.getDashboardUuidFromJwt(
             decodedToken,
             projectUuid,
         );
@@ -743,7 +742,7 @@ export class EmbedService extends BaseService {
 
     async getChartAndResults(
         projectUuid: string,
-        account: Account,
+        account: ExternalAccount,
         tileUuid: string,
         dashboardFilters?: DashboardFilters,
         dateZoomGranularity?: DateGranularity,
@@ -751,11 +750,11 @@ export class EmbedService extends BaseService {
         checkPermissions: boolean = true,
     ) {
         const { data: decodedToken, source: embedToken } =
-            getEmbeddedAuth(account);
+            account.authentication;
         const { dashboardUuids, allowAllDashboards, user } =
             await this.embedModel.get(projectUuid);
 
-        const dashboardUuid = await this.getDashboardUuidFromContent(
+        const dashboardUuid = await this.getDashboardUuidFromJwt(
             decodedToken,
             projectUuid,
         );
@@ -865,16 +864,15 @@ export class EmbedService extends BaseService {
     }
 
     async calculateTotalFromSavedChart(
-        account: Account,
+        account: ExternalAccount,
         projectUuid: string,
         savedChartUuid: string,
         dashboardFilters?: DashboardFilters,
         invalidateCache?: boolean,
     ) {
-        const { data: decodedToken, source: embedToken } =
-            getEmbeddedAuth(account);
+        const { data: decodedToken } = account.authentication;
 
-        const dashboardUuid = await this.getDashboardUuidFromContent(
+        const dashboardUuid = await this.getDashboardUuidFromJwt(
             decodedToken,
             projectUuid,
         );
@@ -957,10 +955,7 @@ export class EmbedService extends BaseService {
             explore,
             queryTags: {
                 embed: 'true',
-                external_id: EmbedService.getExternalId(
-                    decodedToken,
-                    embedToken,
-                ),
+                external_id: account.user.id,
                 project_uuid: projectUuid,
                 organization_uuid: chart.organizationUuid,
                 chart_uuid: chart.uuid,
@@ -989,7 +984,7 @@ export class EmbedService extends BaseService {
         filters,
         forceRefresh,
     }: {
-        account: Account;
+        account: ExternalAccount;
         projectUuid: string;
         filterUuid: string;
         search: string;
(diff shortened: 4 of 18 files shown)


stevelikesmusic:
* **#15648** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15599** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15599?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15579** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15579?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/backend/src/auth/lightdashJwt.ts
```diff
@@ -45,15 +45,15 @@ export function encodeLightdashJwt(
...
+): CreateEmbedJwt {
     try {
         const encryptionUtil = new EncryptionUtil({ lightdashConfig });
         const secret = encryptionUtil.decrypt(
             Buffer.isBuffer(encodedSecret)
                 ? encodedSecret
                 : Buffer.from(encodedSecret),
         );
-        const decodedToken = verify(token, secret) as EmbedJwt;
+        const decodedToken = jwt.verify(token, secret) as CreateEmbedJwt;
```
owlas: do we need schema checking here? Are we sure it has `CreateEmbedJwt` ? These payloads are user created

packages/backend/src/ee/models/EmbedModel.ts
```diff
@@ -15,6 +15,11 @@ export class EmbedModel {
         const [embed] = await this.database('embedding')
             .select()
             .leftJoin('users', 'embedding.created_by', 'users.user_uuid')
+            .leftJoin(
```
owlas, line 18: is the organization supposed to be the organization that the embedding belongs to? 

In that case it would be better to go from embed -> projects -> organizations and not from users -> organizations (which also needs consideration of the organization_memberships table)

packages/backend/src/ee/services/EmbedService/EmbedService.ts
```diff
@@ -553,8 +556,8 @@ export class EmbedService extends BaseService {
                       let sanitizedValue: string[];
                       if (typeof value === 'string') {
                           sanitizedValue = [value];
-                      } else if (isArray(value)) {
-                          sanitizedValue = value.map((v) =>
+                      } else if (Array.isArray(value)) {
+                          sanitizedValue = (value as string[]).map((v) =>
```
owlas: why do we have `as string[]` is it guaranteed?

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
```diff
@@ -1,14 +1,9 @@
...
-    SessionUser,
-} from '@lightdash/common';
+import { ForbiddenError, ParameterError } from '@lightdash/common';
 import { NextFunction, Request, Response } from 'express';
-import { EmbedService } from '../ee/services/EmbedService/EmbedService';
-import Logger from '../logging/logger';
-import { decodeJwt, JWT_HEADER_NAME } from '../utils/JwtUtil';
+import { EmbedService } from '../../ee/services/EmbedService/EmbedService';
+import Logger from '../../logging/logger';
+import { decodeJwt, JWT_HEADER_NAME } from '../../utils/JwtUtil';
```
owlas: this is missing

owlas: could you also avoid the name `utils` which to me usually says "miscellaneous" - is there a better name that makes these functions easier to find?

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
```diff
@@ -66,17 +61,23 @@ export async function jwtAuthMiddleware(
         const { encodedSecret, organization } =
             await embedService.getEmbeddingByProjectId(projectUuid);
         const decodedToken = decodeJwt(embedToken, encodedSecret);
+        const dashboardUuid = await embedService.getDashboardUuidFromContent(
```
owlas: why do we check for dashboard uuid here? is that an essential part of the jwt schema? If yes should it be part of a specific schema check. If not, should this be handled by the services?

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+    MemberAbility,
+    Organization,
+    OrganizationMemberRole,
+} from '@lightdash/common';
+import { applyEmbeddedAbility } from '@lightdash/common/src/authorization/scopeAbility';
+
+function getExternalId(decodedToken: CreateEmbedJwt, embedToken: string) {
+    return (
+        decodedToken.user?.externalId ||
+        decodedToken.iat?.toString() ||
```
owlas: these don't seem like valid fallback ids? why are we using a timestamp or the entire token? The return type of this is `any`

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+export function hydrateEmbeddedAccount(
+    organization: Pick<Organization, 'organizationUuid' | 'name'>,
+    embedJwt: CreateEmbedJwt,
+    rawToken: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Account<CreateEmbedJwt> {
+    // Build abilities for the embedded user
+    const dashboardId = getDashboardId(embedJwt);
+
+    if (!dashboardId) {
```
owlas: this was also checked in the jwt decode but makes more sense here

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -0,0 +1,80 @@
...
+) => EmbeddedAbilityBuilderPayload;
+
+const addBaseAbilities: EmbeddedAbilityBuilder = ({
+    embedUser,
+    dashboardUuid,
+    builder,
+}) => {
+    const { content } = embedUser;
+    const { can } = builder;
+    can('view', 'Dashboard', {
```
owlas, line 24: should we add organizationUuid to be safe?

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -0,0 +1,80 @@
...
+        dashboardUuid,
+    });
+    can('view', 'SavedChart', {
```
owlas: this allows viewing all public saved charts,  that doesn't seem right

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -0,0 +1,80 @@
+import { type AbilityBuilder } from '@casl/ability';
```
owlas, line 1: i like the approach here

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const dashboardAbilities: EmbeddedAbilityBuilder = ({
+    embedUser,
+    dashboardUuid,
+    builder,
+}) => {
+    const { content } = embedUser;
+    const { can } = builder;
+
+    can('view', 'Dashboard', { dateZoom: content.canDateZoom });
```
owlas: does this mean they can view any dashboard that has datezoom enabled?

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -0,0 +1,80 @@
...
+const exportAbilities: EmbeddedAbilityBuilder = ({
+    embedUser,
+    dashboardUuid,
+    builder,
+}) => {
+    const { content } = embedUser;
+    const { can } = builder;
+
+    if (content.canExportCsv) {
+        can('export', 'Dashboard', 'csv');
```
owlas: not sure if this is formatted correctly

Maybe something like:

```typescript
can('create', 'DashboardExport', {
  dashboardUuid: dashboardUuid,
  type: 'csv'
});
```

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+    embedJwt: CreateEmbedJwt,
+    dashboardUuid: string,
+): MemberAbility {
+    const builder = new AbilityBuilder<MemberAbility>(Ability);
+    applyEmbeddedAbility(embedJwt, dashboardUuid, builder);
+
+    return builder.build();
+}
+
+function getDashboardId(embedJwt: CreateEmbedJwt) {
```
owlas: how do consumers of this know if they got the slug or uuid?

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Account<CreateEmbedJwt> {
+    // Build abilities for the embedded user
+    const dashboardId = getDashboardId(embedJwt);
+
+    if (!dashboardId) {
+        throw new Error('Dashboard UUID is required for dashboard embedding');
+    }
+
+    const abilities = buildEmbedAbilities(embedJwt, dashboardId);
```
owlas: this will fail for slugs

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+                {
+                    id: dashboardId,
+                    filters: allowedFilters ?? [],
+                },
+            ],
+        },
+        organization,
+        // Create the fields we're able to set from the JWT
+        user: {
+            userUuid: getExternalId(embedJwt, rawToken),
```
owlas: seems this will cause errors given the range of types `getExternalId` can return

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+            ability: abilities,
```
owlas: these seem really permissive based on the `buildEmbedAbilities` functions

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+            abilityRules: abilities.rules,
+            email: embedJwt.user?.email,
+            firstName: '',
+            lastName: '',
+            organizationUuid: organization.organizationUuid,
+            organizationName: organization.name,
+            isTrackingAnonymized: false,
+            isMarketingOptedIn: false,
+            isSetupComplete: false,
+            userId: 0,
```
owlas: could this cause problems?

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+            dashboards: [
+                {
+                    id: dashboardId,
+                    filters: allowedFilters ?? [],
+                },
+            ],
+        },
+        organization,
+        // Create the fields we're able to set from the JWT
+        user: {
```
owlas: if this is only used by embed services right now why constrain ourselves to the SessionUser schema and why not create a more permissive structure like one with an optional `userId` etc.

packages/backend/src/auth/lightdashJwt.ts
stevelikesmusic: We make the schema check right below. I was trying not to refactor this too much from what the EmbedSevice was doing.

packages/backend/src/ee/models/EmbedModel.ts
stevelikesmusic: Yeah, that's true. Good callout. I'll make that change

packages/backend/src/ee/services/EmbedService/EmbedService.ts
stevelikesmusic: It should be guaranteed as far as TS is concerned as this operates on: 

```ts
export type UserAttributeValueMap = Record<string, string[]>;
```

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
stevelikesmusic: I can definitely update the name. The intent with utils is viewing `jwtAuthMiddleware` as a module and `utils` being scoped just to this module and not intended for re-use elsewhere.

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
stevelikesmusic: The JWT schema defines a `dashboardUuid` or `dashboardSlug`. We want to give proper CASL permissions—which we do in the next stack PR. This was a check we were doing in EmbedService and have now moved it into the middleware.

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: This was lifted from EmbedService. What would you suggest as a better fallback? Something like `orgId + embedToken`

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: Do they need to care? I understood the slug as just an easier/human-readable way to identify a dashboard. By the time we get here, we just care about the dashboardId to fetch from the DB. 

FWIW, `req.account.authentication.data` is the decoded JWT

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: That's true, I'll remove it from the middleware

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: By the time we get here, we've translated the slug to the dashboard ID

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: That's a great callout and something I'm changing in the downstack PR. TLDR is that `SessionUser` will be a union of something like `LightdashUser` and `ExternalUser`

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: I'm addressing this in the downstack where `externalId` and `userUuid` will be optional fields depending on the user type.

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: Permissive how? I'm still wrapping my head around the CASL abilities. I tried to just add what let async-queries pass.

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: I'm addressing this with two User types.

packages/common/src/authorization/scopeAbility.ts
stevelikesmusic: We can!

packages/common/src/authorization/scopeAbility.ts
stevelikesmusic: This should be that they can use date zoom filtering. Maybe the verb should be different like `use` or `filter`?

packages/backend/src/auth/lightdashJwt.ts
stevelikesmusic: The other thing to mention is that even though we do validate against the schema, we don't actually enforce it. I had updated this to throw if validation failed and it was a breaking change for customers.

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
stevelikesmusic: @owlas thinking more about this: Long-term it makes a lot of sense to re-think the overall type strategy for users. Making something like `userUuid` would be a pretty large re-factor of the code to type guard all the places that we assume `user.userUuid` exists. Ideally, we'd have a generic `user.id` as part of a base user. 

For this PR—and I would say this body of work for EmbeddedExplores—would you open to compromising leveraging the current SessionUser? I can open an issue to come back and add stronger typing around users.

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
stevelikesmusic: Updated the name

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
owlas: Yes it matters, for example in the `SavedSqlModel` you have two methods `.getBySlug` which takes a slug and a projectuuid, since slug is only unique within a project. there's a second method `.getByUuid` which takes just the globally unique Uuid. So it's always important to know whether you're handling a slug or uuid

packages/backend/src/middlewares/jwtAuthMiddleware/hydrateEmbeddedAccount.ts
```diff
@@ -0,0 +1,63 @@
+// eslint-disable
```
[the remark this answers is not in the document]
stevelikesmusic: We've seen this in a few places now 🤔

packages/backend/src/auth/lightdashJwt.ts
```diff
@@ -45,15 +44,15 @@ export function encodeLightdashJwt(
 export function decodeLightdashJwt(
     token: string,
     encodedSecret: string | Buffer,
-): EmbedJwt {
```
stevelikesmusic, line 48: `EmbedJwt` is inferred from zod, which does not play nicely with tsoa.

packages/common/src/authorization/scopeAbility.ts
rephus: I agree, we should be dynamically checking the abilities based on the input variables. Otherwise, this will return true even for a user who is trying to export a csv from a different org.

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const dashboardAbilities: EmbeddedAbilityBuilder = ({
+    embedUser,
+    dashboardUuid,
+    builder,
+}) => {
+    const { content } = embedUser;
+    const { can } = builder;
+
+    can('view', 'Dashboard', { dateZoom: content.canDateZoom ?? false });
```
rephus: Same, use `dashboardUuid` to filter abilities based on the dashboard

packages/common/src/types/user.ts
```diff
@@ -4,38 +4,42 @@ import { type AnyType } from './any';
...
 }
 
-export type ExternalUser = BaseUser &
-    Partial<LightdashUser> & {
-        externalId: string;
-        type: 'external';
-    };
+export interface ExternalUser extends AccountUser {
+    id: string;
+    type: 'external';
```
rephus, line 59: Where is the interface that extends AccountUser and does type: 'lightdash' ?

packages/backend/src/ee/controllers/embedController.ts
```diff
@@ -176,9 +175,9 @@ export class EmbedController extends BaseController {
         @Path() projectUuid: string,
     ): Promise<ApiEmbedDashboardResponse> {
         this.setStatus(200);
-        if (!req.account) {
-            throw new ForbiddenError('Account is missing');
-        }
+
+        assertEmbeddedAuth(req.account);
```
rephus, line 179: nice

packages/common/src/authorization/types.ts
```diff
@@ -2,12 +2,13 @@ import { type Ability, type ForcedSubject } from '@casl/ability';
...
-    | 'update'
-    | 'view'
     | 'create'
     | 'delete'
-    | 'promote';
+    | 'export'
+    | 'manage'
+    | 'promote'
+    | 'update'
+    | 'view';
```
stevelikesmusic, line 11: 🔤 only

packages/common/src/authorization/types.ts
```diff
@@ -19,37 +20,37 @@ interface Organization {
 }
 
 export type CaslSubjectNames =
```
stevelikesmusic, line 22: Just 🔤

packages/common/src/types/user.ts
stevelikesmusic: I haven't added that one yet. I was just focussed on ExternalUser. Happy to add that here if it'd help.

stevelikesmusic: @rephus Added an example of SessionAccount now in [auth.ts](https://github.com/lightdash/lightdash/pull/15599/files#diff-fcb90d1217cc668b911f997ec11ffd44aebe96e1e09d0855330b77178499fd01R72-R77)

PR #15616, opened 2025-06-30
fix: lightdash jwt lint

rephus:
Closes: 

image


packages/backend/src/auth/lightdashJwt.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
rephus, line 1: I can't replicate these eslint errors locally, but they keep failing on CI

stevelikesmusic (approved):
So weird

PR #15627, opened 2025-06-30
fix: improve error logs sso

rephus:
Closes:

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -1533,14 +1533,22 @@ export class UserService extends BaseService {
                     if (err || !accessToken) {
                         // Make sure you are passing a google's refresh token, and not a snowflake refresh token by mistake
                         // othwerise this will throw a `invalid_grant` error
+                        const lastFourDigits = refreshToken.slice(-4);
+
+                        // Extract meaningful error message from the error object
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/UserService.ts
```diff
@@ -1533,14 +1533,22 @@ export class UserService extends BaseService {
...
+                        const lastFourDigits = refreshToken.slice(-4);
+
+                        // Extract meaningful error message from the error object
+                        const errorMessage =
+                            err?.data?.error ||
+                            err?.message ||
+                            err?.error ||
+                            'Unknown error';
+
                         console.error(
```
stevelikesmusic, line 1545: When do we decide whether to use `console` or the winston logger? We only expect these logs to make to GCP right? These wouldn't appear in Sentry?

rephus: > When do we decide whether to use console or the winston logger

Yes, we normally use winston whenever available (eg: can't be used in common). But I'm using console.error here, because I'm using `err` as a second parameter, that works well with console, but not with winston. 

> We only expect these logs to make to GCP right

Yes, they should appear on GCP and also on kubectl logs

> These wouldn't appear in Sentry

Yes, console.error don't appear on sentry afaik.

PR #15621, opened 2025-06-30
feat: use filters and userAttributes from req.account in EmbedService

stevelikesmusic:
Closes: #15622

### Description:
Refactored the JWT authentication middleware for embedded dashboards to improve security and maintainability. Key changes include:

1. Restructured the account access controls to properly separate user attributes and filtering permissions
2. Fixed the organization join in the EmbedModel to correctly retrieve organization data
3. Improved the dashboard permissions checking with a more consistent method
4. Extracted JWT verification logic to use the proper decodeLightdashJwt function
5. Added better error handling for missing dashboard UUIDs in embed tokens
6. Simplified the external ID handling by using consistent user identifiers

PR #15648, opened 2025-07-01
feat: use JWT auth user throughout async-query functionality

stevelikesmusic:
Closes:


stevelikesmusic:
* **#15648** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15599** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15579** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15579?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -1431,6 +1431,10 @@ export class LightdashAnalytics extends Analytics {
     }
 
     track<T extends BaseTrack>(payload: TypedEvent | UntypedEvent<T>) {
+        // FIXME: Create a separate trackExternal for embeds
+        const identity = payload.userId
```
stevelikesmusic, line 1514: We'll want a better long-term fix. For now this should suffice.

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -88,20 +89,23 @@ export function hydrateEmbeddedAccount(
         organization,
         // Create the fields we're able to set from the JWT
         user: {
-            userUuid: getExternalId(embedJwt, rawToken),
+            userUuid: '',
```
stevelikesmusic: I'm open to alternatives here—wondering if it's better to allow for userUuid to be `undefined`. I opted for adding `externalId` as its own field to indicate this is not a Lightdash user.

packages/backend/src/middlewares/jwtAuthMiddleware/utils.ts
```diff
@@ -88,20 +89,23 @@ export function hydrateEmbeddedAccount(
...
+            externalId: getExternalId(embedJwt, rawToken),
             ability: abilities,
             abilityRules: abilities.rules,
             email: embedJwt.user?.email,
             firstName: '',
             lastName: '',
             organizationUuid: organization.organizationUuid,
             organizationName: organization.name,
+            organizationCreatedAt: organization.createdAt,
```
stevelikesmusic: This was the last field needed to pass `isUserWithOrg`
https://github.com/lightdash/lightdash/blob/c7a79539574773a25766c42b6af802b254fa07b0/packages/common/src/types/user.ts#L33-L39

packages/backend/src/models/EmailModel.ts
```diff
@@ -55,6 +55,14 @@ export class EmailModel {
     }
 
     async getPrimaryEmailStatus(userUuid: string): Promise<EmailStatus> {
+        if (!userUuid) {
```
stevelikesmusic, line 58: We won't be doing anything email related right now for external users.

packages/backend/src/models/QueryHistoryModel/QueryHistoryModel.ts
```diff
@@ -157,8 +157,14 @@ export class QueryHistoryModel {
         const query = this.database(QueryHistoryTableName)
             .where('query_uuid', queryUuid)
             .andWhere('project_uuid', projectUuid)
-            .andWhere('created_by_user_uuid', userUuid)
             .update(update);
+
+        if (userUuid) {
```
stevelikesmusic, line 162: These types of checks work whether userUuid is `''` or `undefined`

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1268,18 +1269,23 @@ export class AsyncQueryService extends ProjectService {
         if (!isUserWithOrg(user)) {
             throw new ForbiddenError('User is not part of an organization');
         }
-        const userAttributes =
-            await this.userAttributesModel.getAttributeValuesForOrgMember({
-                organizationUuid: user.organizationUuid,
-                userUuid: user.userUuid,
-            });
+        let { userAttributes, intrinsicUserAttributes } = user.controls || {};
```
stevelikesmusic: We load these up for JWT auth when authenticating in the middleware:
https://github.com/lightdash/lightdash/blob/e480c9903b3e92c946aae38d2319f088177bc3d9/packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts#L64-L66

That effectively is the orgMember values because those attributes are provided at runtime via the JWT.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -724,7 +728,7 @@ export class ProjectService extends BaseService {
             credentials = await this.refreshCredentials(credentials, userUuid);
 
             userWarehouseCredentialsUuid = userWarehouseCredentials.uuid;
-        } else {
+        } else if (userUuid) {
             credentials = await this.refreshCredentials(credentials, userUuid);
```
stevelikesmusic: We don't want to try to refresh credentials for a user that isn't in the DB

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2554,22 +2558,8 @@ export class ProjectService extends BaseService {
...
-
-                    const emailStatus =
-                        await this.emailModel.getPrimaryEmailStatus(
-                            user.userUuid,
-                        );
-                    const intrinsicUserAttributes = emailStatus.isVerified
-                        ? getIntrinsicUserAttributes(user)
-                        : {};
-
+                    const { userAttributes, intrinsicUserAttributes } =
```
stevelikesmusic, line 2609: We have the class method, so why not get attributes all from the same place.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3741,11 +3731,14 @@ export class ProjectService extends BaseService {
         if (!explores) {
             return [];
         }
-        const userAttributes =
-            await this.userAttributesModel.getAttributeValuesForOrgMember({
-                organizationUuid,
-                userUuid: user.userUuid,
-            });
+        let { userAttributes } = user.controls || {};
```
stevelikesmusic: These last two are one-offs because we don't use intrinsic values which is another call to the DB. Alternatively, we could join the two queries.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
ZeRego, line 1: @almeidabbm @magnew tagging you to double-check the changes on async queries.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1268,18 +1269,23 @@ export class AsyncQueryService extends ProjectService {
...
-            ? getIntrinsicUserAttributes(user)
-            : {};
+        if (!intrinsicUserAttributes) {
+            const emailStatus = await this.emailModel.getPrimaryEmailStatus(
+                user.userUuid,
+            );
+            intrinsicUserAttributes = emailStatus.isVerified
+                ? getIntrinsicUserAttributes(user)
+                : {};
+        }
```
almeidabbm: `AsyncQueryService` extends `ProjectService`, can't we use `projectService.getUserAttributes` here instead of duplicating logic?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1441,7 +1447,9 @@ export class AsyncQueryService extends ProjectService {
                         await this.queryHistoryModel.create({
                             projectUuid,
                             organizationUuid,
-                            createdByUserUuid: user.userUuid,
+                            createdByUserUuid: user.externalId
```
almeidabbm: From what I can see in `packages/common/src/ee/embed/index.ts` the embed token schema says that `externalId` is optional, therefore doesn't it mean that in here a jwt auth user could be running a query and it will try to set `user.userUuid` if there's no `externalId` in the jwt?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: 🔥 Absolutely! I missed that.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: For a [JWT user](https://github.com/lightdash/lightdash/pull/15648/files#diff-dbccf159688738b24a7565cb87abc0a4ed7a044915eead44abcc0e3e0bfe2a23R93), we'll get a fallback externalId so that we set _something_. If coming in via JWT, they will always have `externalId`. I'm still trying to get the union type to express this elegantly and will probably need to come back in a separate PR to get them right which will require a larger refactor.

almeidabbm: Gotcha! Thank you, sounds good 💪

packages/e2e/cypress/e2e/api/explore.cy.ts
```diff
@@ -0,0 +1,37 @@
...
+            // Test the GetExplore endpoint using JWT authentication
+            cy.request({
+                url: `${apiUrl}/projects/${SEED_PROJECT.project_uuid}/tables/customers`,
+                method: 'GET',
+                headers: {
+                    'lightdash-embed-token': jwtToken,
+                },
+            }).then((resp) => {
+                expect(resp.status).to.eq(200);
+            });
```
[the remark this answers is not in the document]
stevelikesmusic: explores and tables in the context of the Lightdash UI are synonymous

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1617,22 +1642,26 @@ export class AsyncQueryService extends ProjectService {
...
         const { organizationUuid } = await this.projectModel.getSummary(
             projectUuid,
         );
 
+        const ability = getAbilityOrThrow(account, user);
         if (
-            user.ability.cannot(
-                'manage',
+            ability.cannot(
+                'view',
```
stevelikesmusic, line 1713: Requiring `manage` permissions on `Explore` seemed a bit of overkill and would give external JWT users too much access for what we want. @almeidabbm any reason this _shouldn't_ change to `view`?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1657,21 +1686,22 @@ export class AsyncQueryService extends ProjectService {
...
-            metricQuery.exploreName,
+            exploreName: metricQuery.exploreName,
             organizationUuid,
-        );
+        });
 
         const warehouseConnection = await this._getWarehouseClient(
             projectUuid,
-            await this.getWarehouseCredentials(projectUuid, user.userUuid),
+            await this.getWarehouseCredentials(projectUuid, user?.userUuid),
```
stevelikesmusic, line 1753: @almeidabbm there's nothing for us to pass here if we don't have a real user right?

packages/backend/src/services/AsyncQueryService/types.ts
```diff
@@ -18,15 +18,16 @@ import {
     type SortField,
 } from '@lightdash/common';
 
-export type CommonAsyncQueryArgs = {
-    user: SessionUser;
+export type CommonAsyncQueryArgs<UserOverride = SessionUser> = {
```
stevelikesmusic, line 21: This generic is to lower the blast radius of making `user` optional.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -556,19 +559,33 @@ export class ProjectService extends BaseService {
...
     }
 
-    private async getUserAttributes(
-        user: SessionUser,
-        organizationUuid: string,
-    ) {
-        const userAttributes =
-            await this.userAttributesModel.getAttributeValuesForOrgMember({
-                organizationUuid,
+    async getUserAttributes({
```
stevelikesmusic, line 562: This method will be used to handle getting user attributes either from the req.account or the req.user.

rephus (approved):

almeidabbm (approved):

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -368,34 +369,46 @@ export class AsyncQueryService extends ProjectService {
...
             )
         ) {
             throw new ForbiddenError();
         }
 
-        if (user.userUuid !== queryHistory.createdByUserUuid) {
+        if (
+            queryHistory.createdByUserUuid &&
+            user?.userUuid !== queryHistory.createdByUserUuid
+        ) {
```
almeidabbm, line 411: Small question here, does this mean that queries triggered by service accounts can get it's results fetched by anyone?

stevelikesmusic: 🤔 That's a good question @almeidabbm! I suppose technically this is true—any query triggered by a caller that is not in the `user` table could be fetched so long as the someone had the query ID. 

I'm not sure what the best path for this is:
1. We don't do anything additional: For now the application assumes that there's a userUuuid for the query history in most places. We also don't make query IDs searchable/available for anyone to browse.  
2. We add a new `callerId` (or similar) column to the `query_history` that would be the caller (not a user from the `users` table) to identify queries by a particular JWT external user or service account.

almeidabbm: I think we should do option 2 possibly on a separate PR 🤔 even though query uuid's are difficult to "guess" I am not sure how comfortable we would be with someone being able to view results from a query they didn't trigger, e.g. if the JWT has more permissions than the user requesting the results

wdyt @lightdash/glitch ?

stevelikesmusic: Alternatively, what would you all think of moving towards a `createdBy` field rather than a column that was a FK to `users`? Basically rename `createdByUserUuid` to `createdBy` of type `string` without the FK. Then _any_ user identify themselves with the query in a single way (Lightdash user, JWT user, or service account).

I assume there's a cascade involved to delete query history if the user is deleted. I haven't looked through other considerations just yet.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: According to `packages/common/src/authorization/projectMemberAbility.ts` & `packages/common/src/authorization/organizationMemberAbility.ts` changing this CASL check to `view` would give `viewers` access to executing metric queries (these come from the explore page)

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: This should be fine, perhaps the only change I would do is in `getWarehouseCredentials` instead of passing `user?.userUuid ?? ''` when user credentials are required we should throw an error saying that there's no user in that case

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -693,19 +710,26 @@ export class ProjectService extends BaseService {
...
+                userWarehouseCredentialsUuid: undefined,
+            };
+        }
+
         if (credentials.requireUserCredentials) {
             const userWarehouseCredentials =
                 await this.userWarehouseCredentialsModel.findForProjectWithSecrets(
                     projectUuid,
-                    userUuid,
+                    userUuid ?? '',
```
almeidabbm, line 732: Comment above relates to this. We should perhaps just throw an error instead of even trying to get user credentials when there's no user

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: Yeah there's a cascade that doesn't delete it but that sets it to `null` IIRC
Probably a second column would be good? That way we should be able to maintain the FK to users and also one for account?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: Is that not what we want—should Explore viewers not be able to execute metric queries? As I understand, if we give JWT users `manage` perms, they would pass any check on the `Explore` subject.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: In the case of JWTs, there would be no userUuid, but we still want to get the project warehouse creds without throwing.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: Currently `viewers` cannot access the explore page and by consequence they also can't run explore queries, they can only view saved charts

image

~So if I am getting this correctly, if we change this CASL to `view` then `viewers` would be able to execute explore queries 🤔~

Just had a second look at the ability files and we only give permissions for "Explore" for roles > interactive_viewers 🤔 so viewer doesn't even have "view" on explore, I think this change is correct!

packages/backend/src/services/ProjectService/ProjectService.ts
almeidabbm: @stevelikesmusic here's what I mean, if `credentials` has `requireUserCredentials` and there is no user shouldn't it fail instead of passing empty string?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: Awesome. Thank you!!

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: Is there a TTL for these records? JWT users won't have any FK to cascade off of.

packages/backend/src/analytics/LightdashAnalytics.ts
owlas: @stevelikesmusic if user id is missing, this assumes that it's an embed?

packages/backend/src/models/EmailModel.ts
owlas: @stevelikesmusic should this throw instead since we shouldn't be getting primary email status if it's not defined? or do you expect to enter this path without a userUuid

packages/common/src/authorization/scopeAbility.ts
```diff
@@ -25,6 +25,14 @@ const addBaseAbilities: EmbeddedAbilityBuilder = ({
         dashboardUuid,
         organizationUuid: organization.organizationUuid,
     });
+    can('view', 'Explore', {
```
owlas, line 28: nice, eventually it would be cool for casl to also handle the user-attributes checking for whether an explore can be viewed

packages/common/src/types/auth.ts
```diff
@@ -76,12 +76,13 @@ export type AccountAccess = {
      * */
     dashboardId: string;
     filtering?: DashboardFilterInteractivityOptions;
-    controls?: AccountAccessControls;
+    controls: AccountAccessControls;
```
owlas, line 79: from the name I'm not clear what this is

owlas: Is `AccountAccess` specific to embedding dashboards?

packages/common/src/types/auth.ts
```diff
@@ -76,12 +76,13 @@ export type AccountAccess = {
...
 };
 
 type AccountOrganization = Pick<Organization, 'organizationUuid' | 'name'>;
 
 type ILightdashAccount = {
+    type: 'external' | 'lightdash';
```
owlas, line 85: how come we use the `I` syntax on this type only?

packages/common/src/types/auth.ts
```diff
@@ -98,6 +100,7 @@ export type ExternalAccount = LightdashAccount<{
 }>;
 
 export type SessionAccount = LightdashAccount<{
+    type: 'lightdash';
```
owlas, line 103: should this be `LightdashAccount { type: 'lightdash' }` or `SessionAccount { type: 'session' }` for consistency

packages/common/src/types/queryHistory.ts
```diff
@@ -29,7 +29,7 @@ export enum QueryHistoryStatus {
 export type QueryHistory = {
     queryUuid: string;
     createdAt: Date;
-    createdByUserUuid: string | null;
+    createdByUserUuid: string | null | undefined;
```
owlas, line 32: what's the intended difference between null & undefined on this property?

packages/e2e/cypress.config.ts
```diff
@@ -2,6 +2,10 @@ import { defineConfig } from 'cypress';
 import cypressSplit from 'cypress-split';
 import { unlinkSync } from 'fs';
 
+// If running natively, we want to use environment variables from the host machine
+// to be added to Cypress.env()
+const env = process.env.RUNTIME === 'native' ? process.env : {};
```
owlas, line 7: this should be in another PR so it's not blocked - could this leak secrets into cypress? when is RUNTIME===native?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3739,17 +3764,26 @@ export class ProjectService extends BaseService {
...
     }
 
-    async getAllExploresSummary(
-        user: SessionUser,
-        projectUuid: string,
-        filtered: boolean,
-        includeErrors: boolean = true,
-    ): Promise<SummaryExplore[]> {
+    async getAllExploresSummary({
+        user,
```
owlas, line 3768: Why are we keeping `user` in here if we've migrated the endpoint to use `account`?

I'm definitely up for sending `user` and `account` on requests. But once we migrate an endpoint, can't we completely remove the dependency on `SessionUser` ?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3739,17 +3764,26 @@ export class ProjectService extends BaseService {
...
+        account?: Account;
+        projectUuid: string;
+        filtered: boolean;
+        includeErrors?: boolean;
+    }): Promise<SummaryExplore[]> {
         const { organizationUuid } = await this.projectModel.getSummary(
             projectUuid,
         );
+
+        const ability = account?.user.ability ?? user.ability;
```
owlas, line 3784: if we removed the `user` type then we wouldn't need these checks all around

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -368,34 +369,46 @@ export class AsyncQueryService extends ProjectService {
 
     async getAsyncQueryResults({
         user,
```
owlas, line 371: similar to `ProjectService` - what stops us removing `user` as an argument for these?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -368,34 +369,46 @@ export class AsyncQueryService extends ProjectService {
...
         projectUuid,
         queryUuid,
         page = 1,
         pageSize,
-    }: GetAsyncQueryResultsArgs): Promise<ApiGetAsyncQueryResults> {
-        if (!isUserWithOrg(user)) {
+    }: GetAsyncQueryResultsArgs<
+        SessionUser | undefined
+    >): Promise<ApiGetAsyncQueryResults> {
+        if (!user?.organizationUuid && !account?.organization) {
```
owlas, line 380: should account always be scoped to an organization?

owlas: Also if `account?.organization` is `{}` it'll be truthy! is that expected?

packages/backend/src/analytics/LightdashAnalytics.ts
stevelikesmusic: That's right. We assume embed.

packages/backend/src/models/EmailModel.ts
stevelikesmusic: That's a better approach to avoid this call path altogether.

stevelikesmusic:
Closing this as it was broken up into many more PRs

PR #15659, opened 2025-07-02
feat: show sql instead of metric query

joaoviana:
Closes: N/A

### Description:

- Adding a `sql` field to the `VisualizationMetadataSchema` to store generated SQL queries
- Making filters nullable in chart generation tool args schemas (this was necessary for non-filtered queries!)
- Adding a SQL viewer with syntax highlighting that appears when clicking the "SQL" button **instead** of the metric query

image

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AiChartVisualization.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AiChartVisualization.tsx
@@ -23,6 +23,7 @@ import ErrorBoundary from '../../../../../features/errorBoundary/ErrorBoundary';
 import { type EChartSeries } from '../../../../../hooks/echarts/useEchartsCartesianConfig';
 import useHealth from '../../../../../hooks/health/useHealth';
 import { useOrganization } from '../../../../../hooks/organization/useOrganization';
+import { useCompiledSqlFromMetricQuery } from '../../../../../hooks/useCompiledSql';
 import { useExplore } from '../../../../../hooks/useExplore';
 import { type InfiniteQueryResults } from '../../../../../hooks/useQueryResults';
 import useApp from '../../../../../providers/App/useApp';
@@ -79,6 +80,12 @@ export const AiChartVisualization: FC<Props> = ({
     );
 
     const toolCalls = message.toolCalls;
+
+    const { data: compiledSql } = useCompiledSqlFromMetricQuery({
+        tableName,
+        projectUuid,
+        metricQuery,
+    });
 
     const resultsData = useMemo(
         () => ({
@@ -232,7 +239,10 @@ export const AiChartVisualization: FC<Props> = ({
                                 <DrillDownModal />
                             </>
                         ) : activeTab === 'calculation' ? (
-                            <AiChartToolCalls toolCalls={toolCalls} />
+                            <AiChartToolCalls
+                                toolCalls={toolCalls}
+                                compiledSql={compiledSql}
+                            />
                         ) : (
                             assertUnreachable(activeTab, 'Invalid active tab')
                         )}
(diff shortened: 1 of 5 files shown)


joaoviana:
* **#15659** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15659?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15659?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IrakliJani (changes requested):
@joaoviana  SQL generation should come from the metric query, not from the AI. The AI can't guarantee the same results that we get from running the metric query- I bet they don't match up in 99% of cases.

You need to send the metric query to this endpoint instead to get the raw SQL: `/api/v1/projects/{projectUuid}/explores/orders/compileQuery`

joaoviana:
> @joaoviana SQL generation should come from the metric query, not from the AI. The AI can't guarantee the same results that we get from running the metric query- I bet they don't match up in 99% of cases.
> 
> You need to send the metric query to this endpoint instead to get the raw SQL: `/api/v1/projects/{projectUuid}/explores/orders/compileQuery`

cool, I've grabbed the `metriQuery` from running the metric query and have created a new hook to get the compiled sql

packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AiChartVisualization.tsx
```diff
@@ -23,6 +23,7 @@ import ErrorBoundary from '../../../../../features/errorBoundary/ErrorBoundary';
 import { type EChartSeries } from '../../../../../hooks/echarts/useEchartsCartesianConfig';
 import useHealth from '../../../../../hooks/health/useHealth';
 import { useOrganization } from '../../../../../hooks/organization/useOrganization';
+import { useCompiledSqlFromMetricQuery } from '../../../../../hooks/useCompiledSql';
```
IrakliJani, line 26: im surprised we did not have this hook already

IrakliJani: nvm, I double checked and existing `useCompiledSql` is using explorer context.................

IrakliJani (approved):

packages/frontend/src/hooks/useCompiledSql.ts
```diff
@@ -77,3 +77,19 @@ export const useCompiledSql = (
...
+    tableName,
+    projectUuid,
+    metricQuery,
+}: {
+    tableName: string;
+    projectUuid: string;
+    metricQuery: MetricQuery;
+}) => {
+    return useQuery<ApiCompiledQueryResults, ApiError>({
+        queryKey: ['compiledQuery', tableName, metricQuery, projectUuid],
```
IrakliJani, line 91: I think this should be {projectUuid} / "compiledQuery" / {tableName} / {MetricQuery}

so if you invalidate project, this also gets invalidated

IrakliJani (commented):
one minor comment

packages/frontend/src/hooks/useCompiledSql.ts
joaoviana: if I invalidate a project I probably want this query to be invalidated. I am ok with this (it's following the same metric query structure as the `useCompiledSql`

IrakliJani: @joaoviana I meant that when queryClient.invalidateQueries(['project', uuid]); runs somewhere, it won’t invalidate this query. Unfortunately, we don’t have a clear guide on how to structure query keys..

PR #15671, opened 2025-07-02
fix: Fix validation modal doesn't update custom metrics

tatianainama:
Closes:


tatianainama:
* **#15671** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15671?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15671?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/RenameService/RenameService.ts
```diff
@@ -207,15 +207,15 @@ export class RenameService extends BaseService {
...
-                        `Replacing field "${to}" with "${from}" from the chart "${chart.name}"`,
+                        `Replacing field "${from}" with "${to}" from the chart "${chart.name}"`,
                     );
                     throw new NotImplementedError(
-                        `Replacing fields on custom chart fields is not currently supported`,
+                        `Renaming to a custom field (table calculation, custom metric, or custom dimension) is not currently supported. You can rename FROM a custom field, but not TO one.`,
                     );
                 } else {
-                    // Otherwise, it will keep failing
+                    // The target field should be a standard field from the explore
```
tatianainama, line 218: updated docs around this, I found the previous a bit confusing, so let me know if this makes sense 🙏

packages/backend/src/services/RenameService/RenameService.ts
tatianainama, line 1: I've only updated docs in this file. I found the previous ones confusing, so let me know if this makes sense 🙏

packages/backend/src/services/RenameService/rename.ts
```diff
@@ -628,13 +636,15 @@ export const renameSavedChart = ({
...
-            `${nameChanges.from}${isPrefix ? '_' : ''}`,
-        );
 
+    const searchTerms = Object.values({
+        from: addSuffixIfPrefix(nameChanges.from, isPrefix),
+        fromReference: addSuffixIfPrefix(nameChanges.fromReference, isPrefix),
+        fromFieldName: addSuffixIfPrefix(nameChanges.fromFieldName, isPrefix),
+    });
+
+    const containsModelName = buildModelNameChecker(searchTerms);
```
tatianainama: this is the main change. 
We were not considering field and references values. Skipping that omitted changing fields in charts containing custom metrics

ZeRego: The matching rules seem week and might catch cases that we don't want 🤔 I believe this uncovers other issues with the replace logic.

For example, add the following mock and unit test where there is a metric query using a dimension with a similar name "another_Customer_ID":

```
export const chartWithCustomMetricWithSimilarName = {
    uuid: '1382da19-a74e-4e32-afd6-1c5c9ac46ee3',
    projectUuid: '3675b69e-8324-4110-bdca-059031aa8da3',
    name: 'min customer id 1',
    description: '',
    tableName: 'customers',
    updatedAt: new Date('2025-07-01T15:21:02.181Z'),
    updatedByUser: {
        userUuid: 'b264d83a-9000-426a-85ec-3f9c20f368ce',
        firstName: 'David',
        lastName: 'Attenborough',
    },
    metricQuery: {
        exploreName: 'customers',
        dimensions: ['customers_first_name'],
        metrics: ['customers_Customer_ID_min_of_Customer_ID_10'],
        filters: {},
        sorts: [{ fieldId: 'customers_first_name', descending: false }],
        limit: 500,
        metricOverrides: {},
        tableCalculations: [],
        additionalMetrics: [
            {
                name: 'Customer_ID_min_of_Customer_ID_10',
                label: 'Min of Customer id 10',
                description:
                    'Min of Min of Customer id on the table Customers with filters customers.another_Customer_ID',
                uuid: '595f9b6e-0d75-4c1f-a39a-f0b734695607',
                sql: '${TABLE}.another_Customer_ID',
                table: 'customers',
                type: 'min',
                baseDimensionName: 'another_Customer_ID',
                filters: [
                    {
                        id: '0ce8a9c0-be16-46ac-b428-cd1ed4003ab0',
                        target: { fieldRef: 'customers.another_Customer_ID' },
                        values: [10],
                        operator: 'greaterThan',
                    },
                ],
                formatOptions: { type: 'default', separator: 'default' },
            },
        ],
        customDimensions: [],
    },
    chartConfig: {
        type: 'cartesian',
        config: {
            layout: {
                xField: 'customers_first_name',
                yField: ['customers_Customer_ID_min_of_Customer_ID_10'],
            },
            eChartsConfig: {
                series: [
                    {
                        type: 'bar',
                        encode: {
                            xRef: { field: 'customers_first_name' },
                            yRef: {
                                field: 'customers_Customer_ID_min_of_Customer_ID_10',
                            },
                        },
                        yAxisIndex: 0,
                        isFilteredOut: false,
                    },
                ],
            },
        },
    },
    tableConfig: {
        columnOrder: [
            'customers_first_name',
            'customers_Customer_ID_min_of_Customer_ID_10',
        ],
    },
    organizationUuid: '172a2270-000f-42be-9c68-c4752c23ae51',
    spaceUuid: '85ca5b1e-f950-4b32-b701-2d864dd091df',
    spaceName: 'Testint space',
    pinnedListUuid: null,
    pinnedListOrder: null,
    dashboardUuid: null,
    dashboardName: null,
    colorPalette: [
        '#5470c6',
        '#fc8452',
        '#91cc75',
        '#fac858',
        '#ee6666',
        '#73c0de',
        '#3ba272',
        '#9a60b4',
        '#ea7ccc',
    ],
    slug: 'min-customer-id-1',
} as SavedChartDAO;
```
And test
```
    test('should not update chart with custom metric with similar name', () => {
        const { updatedChart, hasChanges } = renameSavedChart({
            type: RenameType.FIELD,
            chart: chartWithCustomMetricWithSimilarName,
            nameChanges: {
                from: 'customers_Customer_ID',
                to: 'customers_customer_id',
                fromReference: 'customers.Customer_ID',
                toReference: 'customers.customer_id',
                fromFieldName: 'Customer_ID',
                toFieldName: 'customer_id',
            },
            validate: false,
        });

        expect(hasChanges).toBe(false);
        expect(updatedChart).toEqual(chartWithCustomMetricWithSimilarName);
    });
```

packages/backend/src/services/RenameService/RenameService.ts
ZeRego: Much clearer now. thank you 🙏

packages/backend/src/services/RenameService/rename.ts
```diff
@@ -616,6 +616,14 @@ export const getNameChanges = ({
     };
 };
 
+const addSuffixIfPrefix = (value: string | undefined, isPrefix: boolean) =>
+    value ? `${value}${isPrefix ? '_' : ''}` : '';
```
rephus: how can value be undefined ? if undefined, this will return '', which means, it will match all models.

packages/backend/src/services/RenameService/rename.ts
```diff
@@ -628,13 +636,15 @@ export const renameSavedChart = ({
...
     const isPrefix = type === RenameType.MODEL;
-    const containsModelName = (object: Object) =>
-        JSON.stringify(object).includes(
-            `${nameChanges.from}${isPrefix ? '_' : ''}`,
-        );
 
+    const searchTerms = Object.values({
+        from: addSuffixIfPrefix(nameChanges.from, isPrefix),
+        fromReference: addSuffixIfPrefix(nameChanges.fromReference, isPrefix),
+        fromFieldName: addSuffixIfPrefix(nameChanges.fromFieldName, isPrefix),
```
rephus: I understand looking at fromReference, but why fieldName ? if we don't need it, I'd suggest to remove it to avoid false positives.

packages/backend/src/services/RenameService/rename.ts
tatianainama: good catch 🙏

packages/backend/src/services/RenameService/rename.ts
tatianainama: I thought of using `fieldName` too as it's only available if the `RenameType` is `Field`. I think that's enough to avoid false positives when updating models?

packages/backend/src/services/RenameService/rename.mock.ts
```diff
@@ -21,6 +21,210 @@ export const fieldRename = createRenameFactory({
...
+        ],
+        limit: 500,
+        metricOverrides: {},
+        tableCalculations: [],
+        additionalMetrics: [
+            {
+                name: 'Customer_ID_min_of_Customer_ID_10',
+                label: 'Min of Customer id 10',
+                description:
+                    'Min of Min of Customer id on the table Customers with filters customers.Customer_ID',
```
rephus, line 151: this could be missleading

tatianainama: that's true and intentional! This description was autogenerated, and I wanted to include it here in the tests to make sure we only update the required values and not user defined things, like titles and descriptions, does it make sense?

packages/backend/src/services/RenameService/rename.ts
tatianainama: I've added this test case!
This was also failing in production, without the changes included here. I think I cover this now with a new check

packages/backend/src/services/RenameService/rename.ts
```diff
@@ -84,14 +84,20 @@ export const createRenameFactory = ({
...
-                    toReferenceWithoutTable,
-                );
+                if (strWithoutTable === fromReferenceWithoutTable) {
+                    return str.replace(
+                        fromReferenceWithoutTable,
+                        toReferenceWithoutTable,
+                    );
+                }
+                return str;
             }
```
tatianainama, line 101: @ZeRego This now covers the case you discover while reviewing.
This function was updating the value no matter what. We are now checking that the string to update actually matches the `fromReference` value before updating

packages/backend/src/services/RenameService/rename.ts
```diff
@@ -684,6 +697,8 @@ export const renameSavedChart = ({
     if (validate)
         validateRename(chart, updatedChart, chart.name, 'chart', nameChanges);
 
+    // Only mark changes if there was actually an update. There are renameFunctions that won't change the values if not needed
+    const hasChanges = JSON.stringify(chart) !== JSON.stringify(updatedChart);
```
tatianainama, line 701: Another false positive here. The `hasChanges` flag was always set to true whenever the replace function was called, regardless of whether it actually changed anything. We only want to change this flag if there was actually a change, so we do a final string-to-string comparison to set it instead.

tatianainama:
### Merge activity

* **Jul 3, 2:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15671).
* **Jul 3, 2:43 PM UTC**: @tatianainama merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15671).

PR #15688, opened 2025-07-03
feat: add Slack OAuth integration

rephus:
Closes: #XXXX

### How to use with AI bot

- open a browser on `/api/v1/auth/slack` to start authentication
- poll `/api/v1/slack/is-authenticated` with user credentials until this returns 200 (returns 404 when not authenticated)
- openid_identities will contain a `user_id` and `slack id` (subject) 

### Changes required on slack apps. 

We need to add a new extra callback to the slack app, for this flow to work. 

Slack app > oauth & permissions > redirect URLs

https://api.slack.com/apps/A04EJP8LZPD/oauth

image

### Demo:

[Screencast from 2025-07-03 09-25-30.webm](https://github.com/user-attachments/assets/3d875dbf-af74-485b-8ca0-c33f9b263e1a)

On slack settings you can see who have logged in, and revoke access if we need to. 
https://lightdash.slack.com/marketplace/A04EJP8LZPD-lightdash-share-local?next_id=0&tab=settings

image

Also added team to the database

image

In the future, we could pass the teamId they want to login into in the request, most likely, we will wrap the `auth` api request and pass an extra argument with the `team` to the slackstrategy ( not doing this now because that will depend on the UI,) 

### AI Description:
Add Slack OAuth integration to allow users to link their Lightdash accounts with Slack. This PR:

- Adds Slack OAuth2 authentication strategy
- Creates type definitions for the Slack passport strategy
- Implements API endpoints for checking if a user has linked their Slack account
- Adds necessary dependencies (`passport-slack-oauth2` and `@slack/oauth`)
- Registers the Slack strategy in the application
- Updates the OpenID identity issuer types to include Slack

### DIFF
--- a/packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
+++ b/packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
@@ -1,0 +1,69 @@
+/// <reference path="../../../@types/passport-slack-oauth2.d.ts" />
+import {
+    ForbiddenError,
+    getErrorMessage,
+    OpenIdIdentityIssuerType,
+    OpenIdUser,
+    UnexpectedDatabaseError,
+} from '@lightdash/common';
+import { Strategy as SlackStrategy } from 'passport-slack-oauth2';
+import { lightdashConfig } from '../../../config/lightdashConfig';
+
+export const slackPassportStrategy = !(
+    lightdashConfig.slack?.clientId && lightdashConfig.slack?.clientSecret
+)
+    ? undefined
+    : new SlackStrategy(
+          {
+              scope: ['openid', 'identity.basic'],
+              clientID: lightdashConfig.slack.clientId,
+              clientSecret: lightdashConfig.slack.clientSecret,
+              callbackURL: `${lightdashConfig.siteUrl}/api/v1/auth/slack/callback`,
+              state: true,
+              passReqToCallback: true,
+          },
+          async (req, accessToken, refreshToken, profile, done) => {
+              if (!req.user || req.user.email === undefined) {
+                  return done(new ForbiddenError('Invalid user'));
+              }
+
+              const openIdUser: OpenIdUser = {
+                  openId: {
+                      subject: profile.id,
+                      issuer: 'slack-oauth2',
+                      issuerType: OpenIdIdentityIssuerType.SLACK,
+                      email: req.user.email,
+                      firstName: profile.user.name.split(' ')[0],
+                      lastName: profile.user.name.split(' ')[1],
+                      groups: undefined,
+                      teamId: profile.team.id,
+                  },
+              };
+
+              try {
+                  // This might throw an error if the user is already linked to the identity
+                  await req.services
+                      .getUserService()
+                      .linkOpenIdIdentityToUser(req.user, openIdUser);
+                  return done(null);
+              } catch (e: unknown) {
+                  if (
+                      e &&
+                      typeof e === 'object' &&
+                      'code' in e &&
+                      e.code === '23505'
+                  ) {
+                      // Postgres duplicate key error code
+                      console.warn(
+                          `Trying to link open id to user ${req.user.userUuid} but user is already linked`,
+                      );
+                      return done(null); // Silent error if user is already linked to the identity
+                  }
+                  console.error(
+                      `Unable to link open id to user ${req.user.userUuid}`,
+                      e,
+                  );
+                  return done(new UnexpectedDatabaseError(getErrorMessage(e)));
+              }
+          },
+      );
--- a/packages/common/src/types/openIdIdentity.ts
+++ b/packages/common/src/types/openIdIdentity.ts
@@ -5,6 +5,7 @@ export enum OpenIdIdentityIssuerType {
     AZUREAD = 'azuread',
     GENERIC_OIDC = 'oidc',
     SNOWFLAKE = 'snowflake',
+    SLACK = 'slack',
 }
 
 export const isOpenIdIdentityIssuerType = (
@@ -21,6 +22,7 @@ export type CreateOpenIdIdentity = {
     userId: number;
     email: string;
     refreshToken?: string; // Used in google to access google drive files
+    teamId?: string;
 };
 
 export type UpdateOpenIdentity = Pick<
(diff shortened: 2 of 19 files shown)


rephus:
* **#15723** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15688** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15688?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15688?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
I have updated the notion guide https://www.notion.so/lightdash/Adding-Slack-to-cloud-instances-9d9d9e7699064d769effb4d5386a81e8#119a63207a7a80e989a8dc7058439bb8 to include the new callback in the manifest

rephus:
We could add a popup redirection on successful login, like we do with google

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
```diff
@@ -0,0 +1,71 @@
...
+                      e &&
+                      typeof e === 'object' &&
+                      'code' in e &&
+                      e.code === '23505'
+                  ) {
+                      // Postgres duplicate key error code
+                      console.warn(
+                          `Trying to link open id to user ${req.user.userUuid} but user is already linked`,
+                      );
+                      return done(null); // Silent error if user is already linked to the identity
```
rephus, line 60: you can get this if you try to authenticate multiple times, and we already created the entry in openid. This should not happen in the AI flow,

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Strategy as SlackStrategy } from 'passport-slack-oauth2';
+import { lightdashConfig } from '../../../config/lightdashConfig';
+
+export const slackPassportStrategy = !(
+    lightdashConfig.slack?.clientId && lightdashConfig.slack?.clientSecret
+)
+    ? undefined
+    : new SlackStrategy(
+          {
+              scope: ['openid'],
```
notgiorgi:
```suggestion
              scope: ['openid', 'identity.basic'],
```

I was getting an error: `{"ok":false,"error":"missing_scope","needed":"identity.basic","provided":"openid"}` during the callback, then added this change and it started working. Do I have some missing configuration?

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
```diff
@@ -0,0 +1,71 @@
...
+                  'http://',
+                  'https://',
+              )}/api/v1/auth/slack/callback`,
+              state: true,
+              passReqToCallback: true,
+          },
+          async (req, accessToken, refreshToken, profile, done) => {
+              if (!req.user || req.user.email === undefined) {
+                  return done(new Error('Invalid user'));
+              }
```
notgiorgi: Edge case here:

1. The user is not logged into Lightdash on their browser.
2. They prompt the AI on Slack.
3. They receive a link for `/api/v1/auth/slack`.
4. When they click the link, they encounter an error 500.

Currently, we lack the logic to save the user’s state (url) and redirect them back to it after logging in. Therefore, we can consider this a known limitation unless you have any suggestions.

We could just redirect to login, so that they can log-in and then go to slack and try again

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
notgiorgi: - I logged in from demo@ligtdash.com (user_id = 1)
- then I logged in with demo2@lightdash.com (user_id = 2)

The slack accounts were the same, `user_id` stayed `1`

It’s likely a well-known limitation that we don’t need to address.

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
rephus: mm, I need to double check, I initially used `identity.basic` but then I changed it to `openid` following slack oauth docs . We can add both if we need to.

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
rephus: good point, I'll take alook

packages/common/src/types/openIdIdentity.ts
```diff
@@ -5,6 +5,7 @@ export enum OpenIdIdentityIssuerType {
     AZUREAD = 'azuread',
     GENERIC_OIDC = 'oidc',
     SNOWFLAKE = 'snowflake',
+    SLACK = 'slack',
```
notgiorgi, line 8: Kind of confusing that we have this enum for `issuer_type` but we always query with `issuer` which is just a string and differs from `issuer_type`

rephus: yeah, issuer type is an enum, but issuer can be anything, dpeending on the type (for google, it is an url)

rephus: afaik we don't use the issuer for anything at the moment.

notgiorgi: okay no worries!

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
rephus: Fixed by adding a redirection to /login?redirect=/auth/slack 
if user is not authenticated.

[Screencast from 2025-07-04 16-36-27.webm](https://github.com/user-attachments/assets/eee5519d-e7ed-4e9d-80e6-394ff197459d)

packages/backend/src/controllers/authentication/strategies/slackStrategy.ts
rephus: after revoking my user, I got this error. I added `identity.basic` as well as `openid` (identity.basic looks like a limitation on passport, but not a requirement on slack)

notgiorgi (approved):
LGTM! 🚢

PR #15707, opened 2025-07-04
chore: migrate instance initialization to new InstanceConfigurationService 

rephus:
Closes: #

## Why ? 

Initial setup was added initially to the organization service, we had to add a few extra dependencies for this to work,  when we added ServiceModel (in ee) we had to initialize also on the EE services, which makes everything more confusing. 
This approach is cleaner.
Soon we'll be adding a new method `updateInstance`  to this service too.

Related: https://github.com/lightdash/lightdash/pull/15464

## Demo

Without license (silent error) 

image

WIth license but no ENV config  (skip)

image

With license and ENV config but with existing org/users  (skip)

image

WIth license and ENV config and no orgs/users (apply initial setup) 

image

### Description:
Created a dedicated `InstanceConfigurationService` to handle instance initialization, moving this functionality out of the `OrganizationService`. This improves separation of concerns and error handling during initialization.

The new service properly handles missing configuration errors and provides better logging during the initialization process. The implementation includes graceful error handling with appropriate debug and error messages.

diff --git packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts

PR #15710, opened 2025-07-04
feat: add updateSetup config for API key and dbt token updates

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/15840

image

### Description:
Added the ability to update instance configuration on server restart through environment variables. This allows updating admin API keys and dbt personal access tokens without requiring a full initial setup. The implementation includes a new `updateSetup` configuration section and a corresponding method in the `InstanceConfigurationService` to apply these updates.

### DIFF
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -413,7 +413,7 @@ const getInitialSetupConfig = (): LightdashConfig['initialSetup'] => {
  * This method is some kind of subset of the initial setup config
  * and it is used to update some of the values on server restart
  */
-const getUpdateSetupConfig = (): LightdashConfig['updateSetup'] => {
+export const getUpdateSetupConfig = (): LightdashConfig['updateSetup'] => {
     if (
         process.env.LD_SETUP_ADMIN_API_KEY &&
         !process.env.LD_SETUP_ADMIN_EMAIL
--- a/packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.test.ts
+++ b/packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.test.ts
@@ -12,6 +12,7 @@ import {
 } from '@lightdash/common';
 import { analyticsMock } from '../../analytics/LightdashAnalytics.mock';
 import { lightdashConfigMock } from '../../config/lightdashConfig.mock';
+import { getUpdateSetupConfig } from '../../config/parseConfig';
 import { InstanceConfigurationService } from './InstanceConfigurationService';
 
 // Mock data
@@ -120,13 +121,19 @@ describe('InstanceConfigurationService.updateInstanceConfiguration', () => {
     });
 
     describe('validation scenarios', () => {
-        test('Do not throw error if there is no update setup config', async () => {
+        test('Do not throw error with default update setup config', async () => {
             service = createMockService({
                 organizationModel: {
                     getOrgUuids: jest
                         .fn()
                         .mockResolvedValue(['org-1', 'org-2']),
                 },
+                projectModel: {
+                    getProjectUuids: jest
+                        .fn()
+                        .mockResolvedValue(['project-1', 'project-2']),
+                },
+                updateSetup: getUpdateSetupConfig(),
             });
 
             await expect(


rephus:
* **#15710** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15710?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15707** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rephus:
This can be reviewed now, I still want ot do a few more tests locally before merging

packages/backend/src/config/parseConfig.ts
```diff
@@ -291,35 +291,32 @@ export const parseOrganizationMemberRoleArray = (
...
+    if (!value) return undefined;
+
+    const enumValues = enumObj ? Object.values(enumObj) : [];
+    if (enumValues.length > 0 && !enumValues.includes(value)) {
+        throw new ParameterError(
+            `Invalid value "${value}". Must be one of ${enumValues.join(', ')}`,
+        );
+    }
+    return value as T;
+};
```
stevelikesmusic, line 317: Since we're returning a provided type, we can infer that type rather than having to pass in as a type arg:

```suggestion
function parseEnum<T extends Record<string, string | number>>(
    value: string | undefined,
    enumObj: T,
): T[keyof T] | undefined {
    if (!value || !enumObj) return undefined;

    const enumValues = enumObj ? Object.values(enumObj) : [];
    if (enumValues.length > 0 && !enumValues.includes(value)) {
        throw new ParameterError(
            `Invalid value "${value}". Must be one of ${enumValues.join(', ')}`,
        );
    }
    return value as T[keyof T];
}
```

One question I have is when would `enumObj` ever be `undefined`?  Why not make `enumObject` required if we're trying to assert the value is in it? I would expect us to returned `undefined` for an empty `enumObject`.

packages/backend/src/config/parseConfig.ts
```diff
@@ -412,6 +409,53 @@ const getInitialSetupConfig = (): LightdashConfig['initialSetup'] => {
     }
 };
 
+/**
```
stevelikesmusic, line 412: Would it be possible for this function to be called in the initial setup function? Then if we need to add more update fields, we only need to remember to add it in one place.

packages/backend/src/ee/models/ServiceAccountModel.ts
```diff
@@ -59,14 +59,14 @@ export class ServiceAccountModel {
...
-        user: SessionUser,
+        user: SessionUser | undefined,
         data: CreateServiceAccount,
         token: string,
     ): Promise<ServiceAccountWithToken> {
         const tokenHash = ServiceAccountModel._hash(token);
         const [row] = await this.database('service_accounts')
             .insert({
-                created_by_user_uuid: user.userUuid,
+                created_by_user_uuid: user?.userUuid || null,
```
stevelikesmusic, line 69: How would this work without being linked to an identity? Is it because of the middleware that gets any admin if we don't have a valid createBy user?

packages/backend/src/App.ts
```diff
@@ -286,6 +286,7 @@ export default class App {
             const instanceConfigurationService =
                 this.serviceRepository.getInstanceConfigurationService<InstanceConfigurationService>();
             await instanceConfigurationService.initializeInstance();
+            await instanceConfigurationService.updateInstanceConfiguration();
```
stevelikesmusic, line 289: If the server restarts, will `initializeInstance` still run? When would the server start and `updateInstance` runs but not `initializeInstance`?

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
```diff
@@ -286,4 +254,164 @@ export class InstanceConfigurationService extends BaseService {
             throw error;
         }
     }
+
+    async updateInstanceConfiguration() {
```
stevelikesmusic: Could this method be broken up into additional methods to group the operations being performed?

stevelikesmusic (approved):

packages/backend/src/App.ts
rephus: both will run, it is up to the methods to decide what to do, depending on the config, and if orgs/projects exist. 

In initializeInstance we will check if there are ogs already, if there is at least one, nothing will happen.

packages/backend/src/ee/models/ServiceAccountModel.ts
rephus: exactly, created_by_user_uuid can be null, so, if no user is provider, we'll set it to null, and then delegate how to handle it to the middleware

packages/backend/src/config/parseConfig.ts
rephus: at the moment, updateSetup is a subset of initialSetup, but I can imagine this can change in the future, and updateSetup can properties that are not part of the initialsetup, for example, Org Color palette (this was discussed at some point)

packages/backend/src/config/parseConfig.ts
rephus: mm, not sure, let me take a look

rephus: I don't think this new type is compatible :thinking: 

image

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
rephus: I created different private methods inside the service

packages/backend/src/config/parseConfig.ts
stevelikesmusic: With the new type, you wouldn't need the type args. You would use:

```ts
parseEnum(process.env.SOME_VAR, WeekDay)
```

The new type will infer the type of `WeekDay` properly and return a value of `some key in WeekDay` since we sometimes return a string if we don't have an enum.

Issue #15683, opened 2025-07-02
I want my queries to run in a graphile worker


almeidabbm:
### Description

Currently queries are running in the backend, this overloads the event loop which could cause it to be unresponsive or even OOM

Running queries in a worker should be behind a feature flag

PR #15729, opened 2025-07-04
feat: add SCHEDULER_INCLUDE_TASKS and SCHEDULER_EXCLUDE_TASKS environment variables

almeidabbm:
Relates to: #15683
Closes: [#15738](https://github.com/lightdash/lightdash/issues/15738)

### Description:
Add ability to include or exclude specific scheduler tasks via environment variables. This allows users to selectively enable only the scheduler tasks they need.

Two new environment variables are introduced:
- `SCHEDULER_INCLUDE_TASKS`: Comma-separated list of tasks to include
- `SCHEDULER_EXCLUDE_TASKS`: Comma-separated list of tasks to exclude

When both variables are set, it will throw a `ParseError`

The implementation includes validation to ensure only valid task names are processed, with warnings for invalid entries.

test-frontend


almeidabbm:
* **#15730** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15730?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15729** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15729?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15729?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/config/parseConfig.ts
```diff
@@ -479,6 +481,76 @@ export const parseResultsS3Config = (): LightdashConfig['results']['s3'] => {
...
+            (task) => !validExcludeTasks.includes(task),
+        );
+    }
+
+    if (validIncludeTasks.length > 0 && validExcludeTasks.length > 0) {
+        // Both set: remove excluded from the included
+        return validIncludeTasks.filter(
+            (task) => !validExcludeTasks.includes(task),
+        );
+    }
```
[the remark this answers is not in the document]
almeidabbm: Changed the behavior. Now it throws an error when both are set as it is redundant, it can be set just by using include

packages/backend/src/config/parseConfig.test.ts
```diff
@@ -418,3 +419,124 @@ describe('process.env.LIGHTDASH_IFRAME_EMBEDDING_DOMAINS', () => {
...
+        // Mock console.warn to capture warning messages
+        jest.spyOn(console, 'warn').mockImplementation(() => {});
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    describe('Default behavior', () => {
+        test('should return all tasks when scheduler tasks environment variables are not set or empty', () => {
```
magnew, line 437: So if you don't set anything, every worker does everything?

almeidabbm: Yes! That's the default behavior. If we want to have multiple worker deployments (at the infra level) we just need to create those in kubernetes and pass the different env variables

magnew (approved):
Great 🎉

PR #15788, opened 2025-07-09
chore: add ai_artifacts and ai_artifact_versions tables

joaoviana:
Closes:


joaoviana:
* **#16483** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16483?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15788** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15788?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15788?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,137 @@
...
+            .references('ai_thread_uuid')
+            .inTable(AiThreadTable)
+            .onDelete('CASCADE');
+        table
+            .uuid('created_by_user_uuid')
+            .nullable()
+            .references('user_uuid')
+            .inTable(UsersTable)
+            .onDelete('SET NULL');
+        table.text('artifact_type').notNullable().defaultTo('chart'); // 'chart', 'visualization', etc.
```
joaoviana, line 29: 'text', 'file'?

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,134 @@
...
+        table.text('title').nullable();
+        table.text('description').nullable();
+        table
+            .uuid('saved_query_uuid')
+            .nullable()
+            .references('saved_query_uuid')
+            .inTable(SavedQueriesTable)
+            .onDelete('SET NULL');
```
joaoviana, line 37: do we want to keep the linking? I think so

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,131 @@
...
+        table
+            .uuid('created_by_user_uuid')
+            .nullable()
+            .references('user_uuid')
+            .inTable(UsersTable)
+            .onDelete('SET NULL');
+        table.integer('version_number').notNullable();
+        table.jsonb('viz_config_output').nullable();
+        table.jsonb('filters_output').nullable();
+        table.jsonb('metric_query').nullable();
```
joaoviana, line 72: we can deprecate an `ai_prompt`'s `viz_config_output` if we use this

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,131 @@
...
+        table.unique(['ai_artifact_uuid', 'version_number']);
+
+        table.index('ai_artifact_uuid');
+        table.index('ai_prompt_uuid');
+        table.index('created_by_user_uuid');
+        table.index('created_at');
+    });
+
+    /*
+    Could we do this?
```
joaoviana, line 83: we could go further and create artifacts + versions for each chart in existing threads - but perhaps we don't want to support that?

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,131 @@
...
+            viz_config_output: prompt.viz_config_output,
+            filters_output: prompt.filters_output,
+            metric_query: prompt.metric_query,
+        });
+    }
+
+    3. After migration is complete and verified:
+       - Update application code to use artifacts instead of ai_prompt.viz_config_output
+       - Eventually drop viz_config_output, filters_output, metric_query from ai_prompt
+       - Remove backwards compatibility code
```
joaoviana, line 123: note ^

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,131 @@
...
+            .defaultTo(knex.raw('uuid_generate_v4()'));
+        table.timestamp('created_at').notNullable().defaultTo(knex.fn.now());
+        table
+            .uuid('ai_thread_uuid')
+            .notNullable()
+            .references('ai_thread_uuid')
+            .inTable(AiThreadTable)
+            .onDelete('CASCADE');
+        table
+            .uuid('created_by_user_uuid')
```
tatianainama, line 24: We can get this information from the prompt, so I think we can avoid duplicating the user reference here

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
tatianainama: I guess we will not allow changing the type of an artifact during versioning? Once you start a chart, you can update chart configurations/filters/etc but not change it to something different. Otherwise, this should be added to the version table instead

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,131 @@
...
+            .nullable()
+            .references('user_uuid')
+            .inTable(UsersTable)
+            .onDelete('SET NULL');
+        table.text('artifact_type').notNullable().defaultTo('chart'); // 'chart', 'visualization', etc.
+        table.text('title').nullable();
+        table.text('description').nullable();
+        table
+            .uuid('saved_query_uuid')
```
tatianainama, line 33: so we link the saved query to the artifact, and then the saved queries versions are attached to it right? something like:
- User creates chart
- User saves query
- User updates chart in chat
  - A new artifact is created
  - A new saved query version is created (linked to the previously created query
- User edits chart in Lightdash outside chat
  - New saved query version is created
  - Saved query link from thread is broken (this column is null now)

Is this correct?

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
tatianainama: yeah, I think users could delete the saved chart but leave this artifact intact

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
```diff
@@ -0,0 +1,131 @@
...
+            .onDelete('CASCADE');
+        table
+            .uuid('ai_prompt_uuid')
+            .nullable()
+            .references('ai_prompt_uuid')
+            .inTable(AiPromptTable)
+            .onDelete('SET NULL');
+        table.timestamp('created_at').notNullable().defaultTo(knex.fn.now());
+        table
+            .uuid('created_by_user_uuid')
```
tatianainama, line 64: same as before, prompts and threads are private, only one user can interact with them, so I think we need to track this in this table 🤔

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
tatianainama, line 72: When could `viz_config_output` and `metric_query` be null? 🤔 Is this based on the type of artifact? If so, then it makes sense to move it to this table and then add a constraint based on that to get better data integrity checks!

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
tatianainama: I think this makes sense! I've never done migrations with data updates like this, so I'm not sure about the possible problems, but this looks good to me

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
tatianainama: do we need to support both flows where we read from prompt and from artifact? no right? once we move to the new table, that's it 🤔

tatianainama (approved):
This looks great!

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
joaoviana: I agree, exactly! I see an artifact type as immutable

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
joaoviana: sounds like a plan - got it 🤔

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
joaoviana: > same as before

this comment? https://github.com/lightdash/lightdash/pull/15788#discussion_r2196860506

So we're ok to get the `user_uuid` from the `ai_artifacts` table -> `ai_web_app_thread`? (because `ai_thread` does not have a user uuid info - sorry so confusing 😓 ) 

or are we saying that we would get it from a `ai_prompt`? 🤔

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
joaoviana: good point! I don't think it should? so yea should be removed - the artifact row shouldn't be stored if it's nullable in cases like these - unless we want to support many other types of ai_artifacts 😓

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
joaoviana: >Saved query link from thread is broken (this column is null now)

yup!

packages/backend/src/ee/database/migrations/20250709140139_ai_artifacts_and_versions.ts
joaoviana: I don't think we have a crazy amount of rows to make this impossible to do. If we convert each prompt output to an artifact we should be in the clear 💪 

But I wonder about the strategy! 
Should the creation of tables + endpoints etc be created first, then we move all existing prompts to populate the new tables?

Issue #13368, opened 2025-01-22
create a `--skip-space-create` flag for `lightdash upload` command


TuringLovesDeathMetal:
### Description

Right now, we create a space on `lightdash upload` if there isn't an existing space in your project for one of the charts/dashboards. 

This is great for creating new projects, but sometimes, I don't want to create a new space because I want to make sure our permissions are set up properly and overwriting these by creating spaces from code can get messy. 

### Acceptance criteria

- there is a flag `--skip-space-create` 
- If a space that a chart is in doesn’t exist - it won’t create the space, and it won’t upload the chart.which would skip creating charts + dashboards in a space that doesn't exist in the target project. 
- The upload succeeds, and the `success` message would list the spaces + content that it didn’t upload.
e.g. 
```bash
success: 
dashboards_num = 30

skipped: 30 charts
list the names of spaces
list the names of charts
```

PR #15801, opened 2025-07-10
feat: add skip-space-create option to CLI upload command

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/13368

Before:

Screenshot from 2025-07-10 12-51-29

After:

Screenshot from 2025-07-10 13-21-01

### Description:
Added a new `--skip-space-create` flag to the CLI upload command that prevents automatic space creation when uploading charts and dashboards. This allows users to upload resources only to existing spaces, avoiding unintended space creation.

The implementation:
1. Added a `skipSpaceCreate` parameter to the chart and dashboard upsert endpoints
2. Modified the CoderService to check this flag before creating a space
3. Updated the CLI to pass this flag through the upload command chain
4. Added appropriate error handling to skip resources when their spaces don't exist

This feature is useful for CI/CD pipelines where you want to ensure resources are only uploaded to pre-defined spaces.

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.ts
+++ b/packages/backend/src/services/CoderService/CoderService.ts
@@ -20,6 +20,7 @@ import {
     getLtreePathFromContentAsCodePath,
     getLtreePathFromSlug,
     NotFoundError,
+    ParameterError,
     Project,
     PromotionAction,
     PromotionChanges,
@@ -677,6 +678,7 @@ export class CoderService extends BaseService {
         projectUuid: string,
         slug: string,
         chartAsCode: ChartAsCode,
+        skipSpaceCreate?: boolean,
     ) {
         const project = await this.projectModel.get(projectUuid);
 
@@ -706,6 +708,7 @@ export class CoderService extends BaseService {
                     projectUuid,
                     chartAsCode.spaceSlug,
                     user,
+                    skipSpaceCreate,
                 );
 
             console.info(
@@ -805,7 +808,9 @@ export class CoderService extends BaseService {
             projectUuid,
             chartAsCode.spaceSlug,
             user,
+            skipSpaceCreate,
         );
+
         const { promotedChart, upstreamChart } =
             await this.promoteService.getPromoteCharts(
                 user,
@@ -846,6 +851,7 @@ export class CoderService extends BaseService {
         projectUuid: string,
         spaceSlug: string,
         user: SessionUser,
+        skipSpaceCreate?: boolean,
     ): Promise<{ space: Omit<SpaceSummary, 'userAccess'>; created: boolean }> {
         const [existingSpace] = await this.spaceModel.find({
             path: getLtreePathFromContentAsCodePath(spaceSlug),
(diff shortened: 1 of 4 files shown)


rephus:
* **#15801** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15801?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/handlers/download.ts
```diff
@@ -509,33 +512,56 @@ const upsertResources = async <T extends ChartAsCode | DashboardAsCode>(
...
-            changes[`${type} with errors`] =
-                (changes[`${type} with errors`] ?? 0) + 1;
-            console.error(
-                styles.error(
-                    `Error upserting ${type}: ${getErrorMessage(error)}`,
-                ),
-            );
+        } catch (error: unknown) {
+            if (
+                error instanceof LightdashError &&
```
rephus, line 528: the backend will throw an error for a missing space, but in the CLi, we transform that error into a "skipped" warning. Charts/dashboards are uploaded individually

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -870,7 +876,11 @@ export class CoderService extends BaseService {
                 "You don't have access to a private space",
             );
         }
-
+        if (skipSpaceCreate) {
+            throw new NotFoundError(
+                `Space ${spaceSlug} does not exist, skipping creation`,
+            );
+        }
```
stevelikesmusic, line 883: Why do we throw a 404 here instead of just returning early if we're opting out of creating a space?

packages/cli/src/handlers/download.ts
```diff
@@ -509,33 +512,56 @@ const upsertResources = async <T extends ChartAsCode | DashboardAsCode>(
...
                 method: 'POST',
                 url: `/api/v1/projects/${projectId}/${type}/${item.slug}/code`,
-                body: JSON.stringify(item),
+                body: JSON.stringify({
+                    ...item,
+                    skipSpaceCreate,
+                }),
             });
 
+            if (upsertData[type].length === 0) {
```
stevelikesmusic: Would this only happen for spaces right now?

packages/cli/src/handlers/download.ts
stevelikesmusic: What if the space does exist but the skip flag is on?

rephus: then we use that space, this is only to avoid creating "public" spaces when they don't exist.

packages/backend/src/services/CoderService/CoderService.ts
rephus: because this is the part where we "create" the space, because it doesn't exist, if there is no flag, we will create an empty new space with public access (some people want to avoid this)

packages/cli/src/handlers/download.ts
rephus: no, for charts and dashboards. Since there was no upserData because space didn't exist

rephus: in fact, this should not happen, since in that case, it will go into the `catch`

rephus: removed

Issue #15803, opened 2025-07-10
detect token expired errors for snowflake and improve UX for refreshing


rephus:
### Description

<img width="873" height="293" alt="Image" src="https://github.com/user-attachments/assets/9dbea2ff-b520-4af1-9238-0f83b51ff4f9" />

At the moment, these errors return an unexpected server error, 

We should use instead a different error type, capture that in the UI and offer the user a way to refresh their token 

See thread: https://lightdash.slack.com/archives/C08TPMFGGTX/p1752137755275789

### Steps to Reproduce the Bug or Issue

.

### Cloud or self-hosting

None

PR #15804, opened 2025-07-10
fix: add Snowflake token refresh error handling with reauthentication UI

rephus:
Closes: https://github.com/lightdash/lightdash/issues/15803

[Screencast from 2025-07-10 16-22-51.webm](https://github.com/user-attachments/assets/331e32f8-a537-4601-aa39-7317b37d06d3)

### Description:
Improved error handling for Snowflake SSO token refresh by adding a specific `SnowflakeTokenError` class that triggers a re-authentication UI in the frontend. When a Snowflake token refresh fails, users will now see a prompt to re-authenticate instead of a generic error message.

The implementation:
1. Adds try/catch block around Snowflake token refresh
2. Creates a new `SnowflakeTokenError` class in common error types
3. Enhances the API error display component to show a Snowflake login form when this specific error occurs
4. Exports the `SnowflakeFormInput` component for reuse in the error display

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -122,6 +122,7 @@ import {
     SavedChartsInfoForDashboardAvailableFilters,
     SessionUser,
     snakeCaseName,
+    SnowflakeTokenError,
     SortByDirection,
     SortField,
     SpaceQuery,
@@ -596,21 +597,29 @@ export class ProjectService extends BaseService {
             args.type === WarehouseTypes.SNOWFLAKE &&
             args.authenticationType === 'sso'
         ) {
-            const token = await this.userModel.getRefreshToken(
-                userUuid,
-                OpenIdIdentityIssuerType.SNOWFLAKE,
-            );
-            this.logger.debug(
-                `Refreshing snowflake token for user ${userUuid}`,
-            );
-            const accessToken = await UserService.generateSnowflakeAccessToken(
-                token,
-            );
-            return {
-                ...args,
-                authenticationType: 'sso',
-                token: accessToken,
-            };
+            try {
+                const token = await this.userModel.getRefreshToken(
+                    userUuid,
+                    OpenIdIdentityIssuerType.SNOWFLAKE,
+                );
+                this.logger.debug(
+                    `Refreshing snowflake token for user ${userUuid}`,
+                );
+                const accessToken =
+                    await UserService.generateSnowflakeAccessToken(token);
+                return {
+                    ...args,
+                    authenticationType: 'sso',
+                    token: accessToken,
+                };
+            } catch (e: unknown) {
+                this.logger.error(
+                    `Error refreshing snowflake token: ${JSON.stringify(e)}`,
+                );
+                throw new SnowflakeTokenError(
+                    `Error refreshing snowflake token`,
+                );
+            }
         }
         return args;
     }
--- a/packages/common/src/types/errors.ts
+++ b/packages/common/src/types/errors.ts
@@ -534,6 +534,20 @@ export class AiAgentNotFoundError extends LightdashError {
     }
 }
 
+/* This specific error will be used in the frontend
+to show a "reauthenticate" button in the UI
+*/
+export class SnowflakeTokenError extends LightdashError {
+    constructor(message: string) {
+        super({
+            message,
+            name: 'SnowflakeTokenError',
+            statusCode: 401,
+            data: {},
+        });
+    }
+}
+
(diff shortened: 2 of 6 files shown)


rephus:
* **#15807** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15807?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15804** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15804?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15804?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
Note this will only update the token for the user and it works for user specific credentials. However, it will not run or compile the project again, to fix the project connection if the token expired is the one that was used to connect the project. We might need to fix this use case on a separate ticket, once it is properly specced 

For that, the user should go to project settings and reauthenticate in there first, then saving the project.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -589,21 +590,29 @@ export class ProjectService extends BaseService {
...
+                    await UserService.generateSnowflakeAccessToken(token);
+                return {
+                    ...args,
+                    authenticationType: 'sso',
+                    token: accessToken,
+                };
+            } catch (e: unknown) {
+                this.logger.error(
+                    `Error refreshing snowflake token: ${JSON.stringify(e)}`,
+                );
```
stevelikesmusic, line 618: Will this error get logged anyway by the generic error handler in App?

packages/common/src/types/errors.ts
```diff
@@ -534,6 +534,20 @@ export class AiAgentNotFoundError extends LightdashError {
...
 
+/* This specific error will be used in the frontend
+to show a "reauthenticate" button in the UI
+*/
+export class SnowflakeTokenError extends LightdashError {
+    constructor(message: string) {
+        super({
+            message,
+            name: 'SnowflakeTokenError',
+            statusCode: 500,
```
stevelikesmusic: Would it be worth making this a 401 if they need to authenticate? Or is it not a guarantee this is an auth error?

packages/frontend/src/hooks/toaster/MultipleToastBody.tsx
```diff
@@ -76,6 +76,7 @@ const MultipleToastBody = ({
                             {toastData.apiError ? (
                                 <ApiErrorDisplay
                                     apiError={toastData.apiError}
+                                    onClose={() => onCloseError?.(toastData)}
```
owlas, line 79: are these changes still needed?

packages/frontend/src/hooks/toaster/useToaster.tsx
```diff
@@ -273,13 +273,14 @@ const useToaster = () => {
                     toastsData={currentErrors.current[key]}
                     onCloseError={(error) => onCloseError(error)}
                 />
-            ) : currentErrors.current[key][0].apiError ? (
+            ) : currentErrors.current[key]?.[0]?.apiError ? (
```
owlas: are these changes still needed?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -589,21 +590,29 @@ export class ProjectService extends BaseService {
...
-            );
-            const accessToken = await UserService.generateSnowflakeAccessToken(
-                token,
-            );
-            return {
-                ...args,
-                authenticationType: 'sso',
-                token: accessToken,
-            };
+            try {
```
owlas, line 600: This whole PR looks great, it would just be a little nicer if we caught the specific errors that need refresh, this would catch any error right?

packages/frontend/src/hooks/toaster/useToaster.tsx
rephus: not anymore, no. Will revert

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: I have triggered this for: 
- expired tokens
- invalid client id 
- invalid token

They all required a re-login to fix it, and probably for some other cases too (eg: snowflake permissions?) . We can add some exceptions if we get them, but right now I don't know any.

rephus:
rerunning tests again, okteto seems to work now

rephus:
Errors seem unrelated, https://lightdash.slack.com/archives/C02GQKJK84Q/p1752507504744589

Issue #15824, opened 2025-07-11
add prometheus metric for query execution


ZeRego:
We should have metrics for:
- `ready`, `cancelled` and `error` queries
- these should be send on `execute` rather than on get results

**Goal:** I want to be able to track the ratio of failed queries

PR #15833, opened 2025-07-11
feat: track query statuses on execution in prometheus

pimmey:
Closes: #15824

### Description:
Add Prometheus monitoring for query statuses on execution to track the ratio of failed queries.

### DIFF
--- a/.claude/commands/add-context-file.md
+++ b/.claude/commands/add-context-file.md
@@ -30,3 +30,7 @@ A diagram is not always necessary but if it helps explain a complex flow, includ
 Link to external documentation and specs or files rather than duplicating them here. If it covers complex engineering
 concepts, link to tutorials for engineers to learn more.
 </content>
+
+<linting>
+After writing the CLAUDE.md file, run `pnpm format` on that single file to ensure proper formatting.
+</linting>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## [0.1830.1](https://github.com/lightdash/lightdash/compare/0.1830.0...0.1830.1) (2025-07-23)
+
+
+### Bug Fixes
+
+* authentication/CLAUDE.md formatting ([#15960](https://github.com/lightdash/lightdash/issues/15960)) ([983fd58](https://github.com/lightdash/lightdash/commit/983fd588c1cf9c947456132c1faf5f2c0233d3d5))
+
+# [0.1830.0](https://github.com/lightdash/lightdash/compare/0.1829.1...0.1830.0) (2025-07-23)
+
+
+### Features
+
+* add JWT header to all UI network requests if present ([#15914](https://github.com/lightdash/lightdash/issues/15914)) ([f9326a4](https://github.com/lightdash/lightdash/commit/f9326a4db1e0e654f2c2dc0c2a536ad2ad854b8b))
+
 ## [0.1829.1](https://github.com/lightdash/lightdash/compare/0.1829.0...0.1829.1) (2025-07-23)
 
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,52 +9,49 @@ Lightdash is an open-source business intelligence tool (Looker alternative) that
 ## Architecture
 
 **Monorepo Structure** (pnpm workspaces):
-- `packages/common/` - Shared utilities, types, and business logic
-- `packages/backend/` - Node.js/Express API server with database layer
-- `packages/frontend/` - React web application with Vite build system
-- `packages/warehouses/` - Data warehouse client adapters (BigQuery, Snowflake, Postgres, etc.)
-- `packages/cli/` - Command-line interface for dbt project management
-- `packages/e2e/` - Cypress end-to-end tests
+
+-   `packages/common/` - Shared utilities, types, and business logic
+-   `packages/backend/` - Node.js/Express API server with database layer
+-   `packages/frontend/` - React web application with Vite build system
+-   `packages/warehouses/` - Data warehouse client adapters (BigQuery, Snowflake, Postgres, etc.)
+-   `packages/cli/` - Command-line interface for dbt project management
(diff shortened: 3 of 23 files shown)


pimmey:
* **#15833** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15833?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15833?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/prometheus.ts
```diff
@@ -52,6 +55,32 @@ export default class PrometheusMetrics {
         }
     }
 
+    public initializeQueryMetrics() {
```
ZeRego: We can move this to the `start` method.

packages/backend/src/prometheus.ts
```diff
@@ -52,6 +55,32 @@ export default class PrometheusMetrics {
...
+                name: 'query_status_total',
+                help: 'Total number of queries by status',
+                labelNames: ['status', 'warehouse_type', 'context'],
+                ...rest,
+            });
+        }
+    }
+
+    public incrementQueryStatus(
+        status: string,
```
ZeRego: Should we enforce the arguments from `string` to the enum types?

packages/backend/src/prometheus.ts
```diff
@@ -52,6 +55,32 @@ export default class PrometheusMetrics {
         }
     }
 
+    public initializeQueryMetrics() {
+        const { enabled, ...rest } = this.config;
+        if (enabled) {
+            this.queryStatusCounter = new prometheus.Counter({
```
ZeRego: The counter doesn't reset on each ping. Is that intended?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1311,6 +1330,13 @@ export class AsyncQueryService extends ProjectService {
                     error: getErrorMessage(e),
                 },
             );
+
+            // Track error query in Prometheus
+            this.prometheusMetrics?.incrementQueryStatus(
+                QueryHistoryStatus.ERROR,
```
ZeRego, line 1341: Was able to get metrics for ready, error and cancelled.
```
# HELP query_status_total Total number of queries by status
# TYPE query_status_total counter
query_status_total{status="ready",warehouse_type="postgres",context="dashboardView"} 5
query_status_total{status="ready",warehouse_type="postgres",context="chartView"} 1
query_status_total{status="ready",warehouse_type="postgres",context="sqlRunner"} 4
query_status_total{status="cancelled",warehouse_type="unknown",context="exploreView"} 1
query_status_total{status="ready",warehouse_type="postgres",context="exploreView"} 1
query_status_total{status="error",warehouse_type="postgres",context="exploreView"} 1
```

Only edge case is in the SQL Runner, since there is an error when we fetch 1 row before running the async query.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1292,6 +1304,13 @@ export class AsyncQueryService extends ProjectService {
                     original_columns: originalColumns,
                 },
             );
+
+            // Track successful query in Prometheus
+            this.prometheusMetrics?.incrementQueryStatus(
+                QueryHistoryStatus.READY,
+                warehouseClient.credentials.type,
+                queryTags.query_context,
+            );
```
almeidabbm, line 1318: On line `1594` there's another update to `QueryHistoryStatus.READY` - when cache is hit
We should also track that one with `prometheus`

packages/backend/src/prometheus.ts
pimmey: @ZeRego it was intended, as I thought we are looking to track cumulative totals over time. So the only time it would reset would be on process restart. Is it not the case?

ZeRego: I don't have a strong opinion, just wondering if it was intentional and if we can calculate error ratio changes. We can experiment with this 👍

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
pimmey: @ZeRego Nice! Shall we add some kind of annotation for the SQL Runner edge case to recognise it?

ZeRego: No changes required in this PR. This flow was also causing problems for dashboard params.
@almeidabbm would it be possible to run this query after creating the row in query_history?

almeidabbm: Don't think so as we need it to be able to get the columns returned by the query to save them to query history 🤔 

Not sure if I understand the edge case you described, that `limit: 1` query runs directly using the warehouse client **before** calling`executeAsyncQuery` so no prometheus metrics will be recorded for it

ZeRego: > so no prometheus metrics will be recorded for it

But shouldn't we? If there is a problem in the SQL compilation or similar, we won't know about it.
This is not a blocker right now, but it creates all these edge cases which might be a pain in the future.

almeidabbm: Yep 100%, we would need to move where we create the `query_history` so that we can save the error there if it happens 🤔

almeidabbm: @ZeRego created an issue to fix this: https://github.com/lightdash/lightdash/issues/15931

almeidabbm (changes requested):
There is an issue with formatting:

Image

PR #15843, opened 2025-07-14
feat: add create_by_account to query_history

stevelikesmusic:
Closes: #15847

### Description:
Added support for tracking queries by account in query history. `account` encapsulates identity that is not userUuid—which would include an external ID. 

These changes are needed so that embedded users can make async queries and have query history associated to their external ID. 

This change:

- Added a new `created_by_account` column to the query history table
- Created a migration to add the column and appropriate indexes
- Updated the QueryHistoryModel to support filtering by account instead of user UUID
- Modified method signatures in AsyncQueryService to accommodate the new account parameter
- Updated type definitions to include the new field

This enhancement allows queries to be associated with either a user UUID or an account identifier, enabling better tracking of query history for different authentication methods.

Moving forward, we'll also be able to run a clean up job on `query_history` with the new index to remove expired history.

### DIFF
--- a/packages/backend/src/ee/database/migrations/20250713101551_add_query_history_account.ts
+++ b/packages/backend/src/ee/database/migrations/20250713101551_add_query_history_account.ts
@@ -1,0 +1,28 @@
+import { Knex } from 'knex';
+
+const QUERY_HISTORY_TABLE = 'query_history';
+const CREATED_BY_ACCOUNT_COLUMN = 'created_by_account';
+
+export async function up(knex: Knex): Promise<void> {
+    // 1. Add a new nullable "createdByAccount" column of type string
+    await knex.schema.alterTable(QUERY_HISTORY_TABLE, (table) => {
+        table.string(CREATED_BY_ACCOUNT_COLUMN).nullable();
+    });
+
+    // 2. Add indexes on the new column
+    await knex.schema.alterTable(QUERY_HISTORY_TABLE, (table) => {
+        table.index([CREATED_BY_ACCOUNT_COLUMN]);
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    // Remove index
+    await knex.schema.alterTable(QUERY_HISTORY_TABLE, (table) => {
+        table.dropIndex([CREATED_BY_ACCOUNT_COLUMN]);
+    });
+
+    // Remove the created_by_account column
+    await knex.schema.alterTable(QUERY_HISTORY_TABLE, (table) => {
+        table.dropColumn(CREATED_BY_ACCOUNT_COLUMN);
+    });
+}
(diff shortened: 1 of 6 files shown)


stevelikesmusic:
* **#15843** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15843?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15843?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/QueryHistoryModel/QueryHistoryModel.ts
```diff
@@ -111,12 +112,16 @@ export class QueryHistoryModel {
...
+            | 'createdByAccount'
+        > & {
+            createdByAccount?: string | null;
+        },
     ) {
         const [result] = await this.database(QueryHistoryTableName)
             .insert({
                 status: QueryHistoryStatus.PENDING,
-                created_by_user_uuid: queryHistory.createdByUserUuid,
+                created_by_user_uuid: queryHistory?.createdByUserUuid ?? null,
```
owlas, line 123: is this one already nullable and it weren't handling it?

packages/backend/src/models/QueryHistoryModel/QueryHistoryModel.ts
```diff
@@ -151,27 +156,55 @@ export class QueryHistoryModel {
     async update(
         queryUuid: string,
         projectUuid: string,
-        userUuid: string,
         update: DbQueryHistoryUpdate,
+        userUuid?: string,
+        account?: string,
```
owlas: what's the name of this now? I've seen `accountId` `externalId` and just `account` is used here.

packages/backend/src/ee/database/migrations/20250713101551_add_query_history_account.ts
```diff
@@ -0,0 +1,31 @@
...
+export async function up(knex: Knex): Promise<void> {
+    // 1. Add a new nullable "createdByAccount" column of type string
+    await knex.schema.alterTable(QUERY_HISTORY_TABLE, (table) => {
+        table.string(CREATED_BY_ACCOUNT_COLUMN).nullable();
+    });
+
+    // 2. Add indexes on the new column and the expired column
+    await knex.schema.alterTable(QUERY_HISTORY_TABLE, (table) => {
+        table.index([CREATED_BY_ACCOUNT_COLUMN]);
+        table.index([RESULTS_EXPIRES_AT_COLUMN]);
```
owlas: indexing the existing expires_at column is the only expensive migration. if you omit this you can make these changes and move on independently of scheduling in the bigger migration for expires at and handling deletes

packages/backend/src/models/QueryHistoryModel/QueryHistoryModel.ts
stevelikesmusic: This column was nullable, but in practice we always have a userUuid. Once we add account, it will be getting passed as null.

packages/backend/src/models/QueryHistoryModel/QueryHistoryModel.ts
stevelikesmusic: Fair enough. I'll make it `accountId` for consistency with `userUuid`.

owlas (approved):

PR #15851, opened 2025-07-14
feat: setup Cypress for JWT async-queries

stevelikesmusic:
Closes: #15848

### Description:
This PR improves the E2E testing infrastructure with several enhancements:

1. Added a new `e2e-open-native` script that allows running Cypress tests using the host machine's environment variables
2. Enhanced error logging in the backend to include stack traces in development mode
3. Added a new `getJwtToken` Cypress command to generate JWT tokens for testing embedded dashboards
4. Improved type safety in Cypress commands by adding TypeScript type annotations
5. Made database connection parameters configurable via environment variables (PGUSER, PGPASSWORD, etc.)
6. Added support for specifying a custom dbt project directory via environment variables

These changes make it easier to run E2E tests natively and sets up functionality for the changes coming to AsyncQuery Service to take JWTs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
         "sdk-typecheck": "pnpm -F frontend typecheck-sdk",
         "sdk-build": "pnpm common-build && pnpm frontend-build-sdk",
         "e2e-open": "pnpm -F e2e cypress:open",
+        "e2e-open-native": "pnpm -F e2e cypress:open:native",
         "e2e-run": "pnpm -F e2e cypress:run",
         "e2e-lint": "pnpm -F e2e lint",
         "e2e-format": "pnpm -F e2e format",
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -624,12 +624,18 @@ export default class App {
                     error instanceof UnexpectedServerError ||
                     !(error instanceof LightdashError)
                 ) {
-                    console.error(error); // Log original error for debug purposes
+                    // This intentionally uses console vs. winston because of problems from some error/JSON payloads.
+                    console.error(error);
                 }
                 Logger.error(
                     `Handled error of type ${errorResponse.name} on [${req.method}] ${req.path}`,
                     errorResponse,
                 );
+
+                if (process.env.NODE_ENV === 'development') {
+                    Logger.error(error.stack);
+                }
+
                 this.analytics.track({
                     event: 'api.error',
                     userId: req.user?.userUuid,
--- a/packages/e2e/cypress.config.ts
+++ b/packages/e2e/cypress.config.ts
@@ -2,6 +2,10 @@ import { defineConfig } from 'cypress';
 import cypressSplit from 'cypress-split';
 import { unlinkSync } from 'fs';
 
+// If running natively, we want to use environment variables from the host machine
+// to be added to Cypress.env()
+const env = process.env.RUNTIME === 'native' ? process.env : {};
+
 export default defineConfig({
     viewportWidth: 1920,
     viewportHeight: 1080,
@@ -65,6 +69,7 @@ export default defineConfig({
             return config;
         },
     },
+    env,
 
     video: true,
     videoCompression: true,
--- a/packages/e2e/cypress/support/commands.ts
+++ b/packages/e2e/cypress/support/commands.ts
@@ -62,20 +62,24 @@ declare global {
 
             registerNewUser(): Chainable<Element>;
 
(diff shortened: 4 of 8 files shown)


stevelikesmusic:
* **#15851** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15851?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15851?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/App.ts
```diff
@@ -624,12 +624,17 @@ export default class App {
                     error instanceof UnexpectedServerError ||
                     !(error instanceof LightdashError)
                 ) {
-                    console.error(error); // Log original error for debug purposes
+                    Logger.error(error); // Log original error for debug purposes
```
rephus: this was intentionally using console, because Logger doesn't like some error types or json, but if it works for you, that's ok.

package.json
```diff
@@ -82,6 +82,7 @@
         "sdk-typecheck": "pnpm -F frontend typecheck-sdk",
         "sdk-build": "pnpm common-build && pnpm frontend-build-sdk",
         "e2e-open": "pnpm -F e2e cypress:open",
+        "e2e-open-native": "pnpm -F e2e cypress:open:native",
```
rephus, line 85: what does it do ?

packages/backend/src/App.ts
```diff
@@ -624,12 +624,17 @@ export default class App {
...
                 }
                 Logger.error(
                     `Handled error of type ${errorResponse.name} on [${req.method}] ${req.path}`,
                     errorResponse,
                 );
+
+                if (process.env.NODE_ENV === 'development') {
+                    Logger.error(error.stack);
```
rephus, line 636: :+1:

rephus: should this happen inside the `error instanceof UnexpectedServerErro` if ?

packages/e2e/cypress.config.ts
```diff
@@ -2,6 +2,10 @@ import { defineConfig } from 'cypress';
 import cypressSplit from 'cypress-split';
 import { unlinkSync } from 'fs';
 
+// If running natively, we want to use environment variables from the host machine
+// to be added to Cypress.env()
+const env = process.env.RUNTIME === 'native' ? process.env : {};
```
rephus, line 7: oh, is this because cypress doesn't take ENV variables unless they are prefixed with CYPRESS_ ?

packages/e2e/cypress/support/commands.ts
```diff
@@ -636,3 +653,94 @@ Cypress.Commands.add(
             });
     },
 );
+
+Cypress.Commands.add(
+    'getJwtToken',
```
rephus: cool command, but do you use it anywhere ? could we have at least 1 example of this being used ?

packages/backend/src/App.ts
stevelikesmusic: I didn't know that had been a problem. Will change it back with a comment.

packages/e2e/cypress.config.ts
stevelikesmusic: yes, we'd have to explicitly set values for things like the DBT project. without this, we can only run local e2e tests for things that create a project against docker—or we explicitly pass in those configs

packages/e2e/cypress/support/commands.ts
stevelikesmusic: It will be used in the async-query PR when it lands. I copied this over from the larger [PR](https://github.com/lightdash/lightdash/pull/15648/files#diff-4b7f83632cad891227cff059c8134a8044c807e4e990d99d183c23b2c4fcd74f) that I'm breaking apart.

I can save it for that upcoming PR if you prefer.

packages/backend/src/App.ts
stevelikesmusic: I think we want it for any error that hits this handler. We still would catch things like ForbiddenError error here. It's helpful because then we have a stack trace to exactly where the error was thrown—rather than only getting a HTTP status + endpoint in the terminal.

package.json
stevelikesmusic: This loads copies the local environment from process.env to Cypress' env. We can then run all e2e tests locally running a native setup rather than only being able to run with docker.

rephus (approved):

Issue #15547, opened 2025-06-25
Ensure we're emitting all the proper analytics for embedded users


stevelikesmusic:
We currently have a separate API for the embedded dashboards. Those analytics are a bit different than our current model that includes a real user in its context. We need to have a way to distinguish between both in the Explore controllers and publish events with the correct config.

PR #15852, opened 2025-07-14
feat: add trackAccount to LightdashAnalytics

stevelikesmusic:
Closes: #15547

### Description:
Added a new `trackAccount` method to `LightdashAnalytics` class to centralize tracking events from an account. This method automatically extracts common properties like `userId`, `anonymousId`, `externalId`, and `organizationId` from the account object, reducing code duplication.

Updated embed analytics to use this new method, simplifying the tracking code in `EmbedService` by removing redundant properties that are now handled automatically.

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -1,6 +1,6 @@
 /// <reference path="../@types/rudder-sdk-node.d.ts" />
-import { Type } from '@aws-sdk/client-s3';
 import {
+    Account,
     AnyType,
     CacheMetadata,
     CartesianSeriesType,
@@ -20,11 +20,9 @@ import {
     QueryHistoryStatus,
     RequestMethod,
     SchedulerFormat,
-    SqlRunnerQuery,
     TableSelectionType,
     ValidateProjectPayload,
     WarehouseTypes,
-    getErrorMessage,
     getRequestMethod,
 } from '@lightdash/common';
 import Analytics, {
@@ -33,7 +31,6 @@ import Analytics, {
 import { Request } from 'express';
 import { v4 as uuidv4 } from 'uuid';
 import { LightdashConfig } from '../config/parseConfig';
-import Logger from '../logging/logger';
 import { VERSION } from '../version';
 
 type Identify = {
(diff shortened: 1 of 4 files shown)


stevelikesmusic:
* **#15852** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15852?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15852?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -1564,4 +1561,33 @@ export class LightdashAnalytics extends Analytics {
...
+        };
+
+        if (!payload.properties) {
+            payload.properties = {};
+        }
+
+        if (account.authentication.type === 'session') {
+            payload.userId = account.user.id;
+        } else {
+            payload.anonymousId = 'embed';
```
rephus, line 1583: is this safe to assume all "non session" will be used for embed ? will we have to remember this when we start using jwt tokens for other purposes ?

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -1564,4 +1561,33 @@ export class LightdashAnalytics extends Analytics {
...
+            payload.properties.externalId = account.user.id;
+        }
+
+        // TODO: Could these be top-level fields rather than properties?
```
rephus, line 1587: I agree, but we need to confirm first how this is used and how easy will be to migrate on analtyics

rephus (approved):
Looks good, in the future we might want to abstract also `track` to pass an "sessionUser | account" and populate the ids, regardless of their authentication method.

packages/backend/src/analytics/LightdashAnalytics.ts
stevelikesmusic: Ah that's a good call. I think I landed on the right type name to user for the user: `known` which implies this is a user in the users table. A better check here will be `account.user.type === 'known'`. That's what we care about for this because `userId` if used will be joined to the users table in dbt.

Issue #15850, opened 2025-07-14
Wrap req.user in req.account


stevelikesmusic:
Add the session user to req.account after passport has handled its auth flow.

PR #15787, opened 2025-07-09
feat: use JWT token in explores for React SDK

stevelikesmusic:
Closes: #15540

### Description:
This PR adds the ability to explore charts from embedded dashboards. When a user clicks "Explore from here" on a chart in an embedded dashboard, they can now navigate to an embedded explore view for that chart. 

For now, this is enabled just for the React SDK, but we can easily enable this with an additional field in the JWT payload. 

### Key changes:
- Added an `Explore` component to the SDK that can be used alongside the existing `Dashboard` component
- Added an `onExplore` callback to the Dashboard component that gets triggered when a user clicks "Explore from here"
- Moved the `JWT_HEADER_NAME` constant from backend to common package for better sharing
- Updated the Explorer components to work with project UUIDs from the embed context using `useEmbed`. 
- Added example implementations in both the React and Next.js test apps

Screen Cast 2025-07-09 at 4 39 23 PM

### Caveats

1. We are showing everything right now in the UI. The next PR will handle using permissions from the JWT to hide elements like the dbt button and SQL Runner.
2. When using the Next app, the viewport is too narrow and we have to scroll. We'll need to make the sidebar responsive to the screen width for narrow views. 
3. I have plans to come back and make the SDK DX a little better. Right now we have to build the whole package to use it in the test apps, which is a bit of a pain when developing. 
4. I haven't tested user-attributes yet, but will be a part of testing before releasing. 

### Testing

Use the [example apps](https://www.notion.so/lightdash/Lightdash-React-SDK-Developer-Guide-208a63207a7a80368a82c2fe5a6dfe9f#208a63207a7a80cabc41d9ab58629fd5) to test Explores

1. Start your regular Lightdash dev server
2. From `packages/frontend` run `pnpm build-sdk` 
3. Get an embed token from the embed page of User Settings. Click `Generate and copy URL` in the UI. 

 

5. Run either example app—the vanilla React app is a little easier because you can paste the embed URL and get the parsed token from it. 
  a. Vanilla React - `packages/frontend/sdk-test-app
  b. Next App       - `packages/frontend/sdk-next-test-app`

PR #15853, opened 2025-07-14
feat: add acccountMiddleware to centralize identity in the app

stevelikesmusic:
Closes: #15850

### Description:
Adds an `accountMiddleware` to centralize identity management across the application. This middleware creates a standardized `Account` object from the authenticated user and attaches it to the request object.

The middleware:
- Only processes requests with an authenticated user
- Preserves existing account information if already present
- Extracts organization details into a dedicated structure
- Normalizes user identity information
- Captures authentication source from cookies

This sets us up to replace req.user with req.account in Project and AsyncQuery services.


stevelikesmusic:
* **#16002** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16002?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15999** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15999?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15893** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15904** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15904?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15853** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
```diff
@@ -0,0 +1,52 @@
+import { SessionAccount } from '@lightdash/common';
+import { NextFunction, Request, Response } from 'express';
+
+/**
+ * Middleware to attach the account to the request.
+ * This is used to centralize identity across the application.
+ */
+export function accountMiddleware(
```
owlas: should it session account middleware since it explicitly converts sessions authentication into accounts? similar to how `jwtAuthMiddleware` is for jwts

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
```diff
@@ -0,0 +1,52 @@
...
+    const account: SessionAccount = {
+        authentication: {
+            type: 'session',
+            source: req.headers.cookie || '',
+        },
+        organization,
+        user: {
+            ...user,
+            organization,
+            type: 'lightdash',
```
owlas: i still find `type: lightdash` quite generic

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    const {
+        organizationCreatedAt,
+        organizationName,
+        organizationUuid,
+        ...user
+    } = req.user;
+
+    const account: SessionAccount = {
```
owlas: is this missing abilities? The type for `SessionAccount` doesn't include them

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
stevelikesmusic: What about `native`? I think references to something like `internal` are typically more specific

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
stevelikesmusic: This does have abilities. Session users get their abilities from the model when their fetched and parsed from the DB. 

There's some inheritance/type-extension at work:
1. SessionAccount has a [LightdashSessionUser](https://github.com/lightdash/lightdash/blob/4bb54b999facce404f1bf13a85703e01723529a2/packages/common/src/types/user.ts#L39-L55)
2. That LightdashSessionUser extends [AccountUser](https://github.com/lightdash/lightdash/blob/4bb54b999facce404f1bf13a85703e01723529a2/packages/common/src/types/user.ts#L8-L16) which is what every Account should use for its user implementation
3. We use a [generic](https://github.com/lightdash/lightdash/blob/4bb54b999facce404f1bf13a85703e01723529a2/packages/common/src/types/auth.ts#L84-L90) to enforce these interfaces. These interfaces are used to compose what [Account](https://github.com/lightdash/lightdash/blob/4bb54b999facce404f1bf13a85703e01723529a2/packages/common/src/types/auth.ts#L110) is.

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
stevelikesmusic: That's true. Maybe something like `native`? I'd prefer not to use something like `internal` as it suggests someone with an `@lightdash.com` domain to me.

packages/backend/src/middlewares/accountMiddleware/accountMiddleware.ts
stevelikesmusic: Yeah something like `sessionAccountMiddleware`?

packages/backend/src/auth/account/account.ts
```diff
@@ -0,0 +1,110 @@
+import { Ability, AbilityBuilder } from '@casl/ability';
```
stevelikesmusic, line 3: This keeps all account creation logic in one place rather than spread between middlewares.

packages/backend/src/middlewares/jwtAuthMiddleware/hydrateEmbeddedAccount.ts
```diff
@@ -1,73 +0,0 @@
-/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call */
```
stevelikesmusic, line 1: This all moved to `fromJwt`

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
```diff
@@ -79,20 +88,20 @@ export async function jwtAuthMiddleware(
         ]);
 
         if (!dashboardUuid) {
-            res.status(404).json({
-                status: 'error',
-                message: 'Cannot verify JWT. Dashboard ID not found',
-            });
+            next(
```
stevelikesmusic, line 91: Gives us better insights into why/where the 403 occurred.

packages/common/src/types/user.ts
```diff
@@ -12,7 +11,7 @@ export type AccountUser = {
     isActive: boolean;
     abilityRules: AbilityBuilder<MemberAbility>['rules'];
     ability: MemberAbility;
-    type: 'lightdash' | 'external';
+    type: 'customer' | 'external';
```
stevelikesmusic: What do you think of this name @owlas?

stevelikesmusic: Actually, I think maybe `known` is the way to go—a user that's in our user table we can make relations to.

packages/common/src/types/user.ts
```diff
@@ -45,7 +44,6 @@ export interface LightdashSessionUser extends AccountUser {
     userId: number;
     firstName: string;
     lastName: string;
-    organization: Pick<Organization, 'organizationUuid' | 'name'>;
```
stevelikesmusic, line 48: Dropping this in favor of `account.organization`

packages/common/src/types/user.ts
stevelikesmusic: Went with `'known'`

packages/backend/src/auth/account/account.ts
```diff
@@ -0,0 +1,118 @@
...
+    ExternalAccount,
+    ForbiddenError,
+    MemberAbility,
+    Organization,
+    SessionAccount,
+    SessionUser,
+} from '@lightdash/common';
+import { Account } from '@lightdash/common/src/types/auth';
+
+function getExternalId(
```
owlas: prefer to use `const = ()` instead of `function`

packages/backend/src/auth/account/account.ts
```diff
@@ -0,0 +1,118 @@
...
+): string {
+    return (
+        decodedToken.user?.externalId ||
+        `anonymous-jwt::${organization.organizationUuid}_${embedToken}`
+    );
+}
+
+function createAccount(account: Partial<SessionAccount>): SessionAccount;
+function createAccount(account: Partial<ExternalAccount>): ExternalAccount;
+function createAccount<T extends Account>(account: Partial<T>): T {
```
owlas: I don't think we have function overloading anywhere in the app. The type for `createAccount` looks good enough without the overloads

```
const createAccount = <T extends Account>(account: Partial<T>): T => {}

File: packages/backend/src/auth/account/account.ts [hunk 11]
```diff
@@ -0,0 +1,118 @@
...
+    return {
+        ...account,
+        isAuthenticated: () => !!account.authentication && !!account.user?.id,
+        isKnownUser: () => account.user?.type === 'known',
```
owlas: What is the difference between these?

packages/backend/src/auth/account/account.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (!organizationUuid || !organizationName || !organizationCreatedAt) {
+        const missing: string[] = [];
+        if (!organizationUuid) missing.push('organizationUuid');
+        if (!organizationName) missing.push('organizationName');
+        if (!organizationCreatedAt) missing.push('organizationCreatedAt');
+
+        throw new ForbiddenError(
+            `Missing required organization fields: ${missing.join(', ')}.`,
+        );
+    }
```
stevelikesmusic: Replaces all the checks for isUserWithOrg throughout the app:
https://github.com/lightdash/lightdash/blob/8e03c6bc9cda1aa41a57730ca6dd000e8a1ff8c9/packages/common/src/types/user.ts#L65-L71

packages/backend/src/App.ts
```diff
@@ -92,7 +93,7 @@ declare global {
              * @deprecated Clients should be used inside services. This will be removed soon.
              */
             clients: ClientRepository;
-            account?: Account;
+            account: Account;
```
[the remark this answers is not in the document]
owlas: seems legit

packages/backend/src/auth/account/account.ts
owlas: @ZeRego added `isUserWithOrg` - do you remember why this was needed?

packages/common/src/types/user.ts
owlas: `'registered'` vs. `'anonymous'`?

packages/backend/src/auth/account/account.ts
stevelikesmusic: I'm going to remove this check from Oliver's [Slack context](https://lightdash.slack.com/archives/C08L4N6GU1J/p1752751409411389?thread_ts=1752744894.132299&cid=C08L4N6GU1J).

@ZeRego does CASL not catch what this check is looking for?

packages/backend/src/App.ts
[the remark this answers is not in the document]
stevelikesmusic: yeah that's fair.

packages/backend/src/auth/account/account.ts
stevelikesmusic: The idea was to return the discriminate type from `createAccount`. But we could just also return the generic `Account`.

packages/backend/src/auth/account/account.ts
stevelikesmusic: `isAuthenticated` -> analogous to `req.user.isAuthenticated()` which just says yes we have a user
`isKnownUser` -> Is this a user in the database. Will be replaced with `isRegisteredUser`. We need this distinction because of the route we went with `query_history`.

owlas (commented):
deleted

owlas (approved):

packages/backend/src/middlewares/accountMiddleware/sessionAccountMiddleware.test.ts
```diff
@@ -0,0 +1,59 @@
...
+                user: defaultSessionUser,
+            } as express.Request;
+
+            sessionAccountMiddleware(request, mockResponse, mockNext);
+
+            expect(request.account).toBeDefined();
+            expect(mockNext).toHaveBeenCalledWith();
+        });
+    });
+});
```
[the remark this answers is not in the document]
stevelikesmusic: We have unit tests to verify Account.fromSession. We only want to verify that we add to the req object or not.

PR #15854, opened 2025-07-14
chore: AI Agent tests

tatianainama:
Closes:


tatianainama:
* **#15883** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15883?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15854** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15854?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15854?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/test-and-build.yml
```diff
@@ -51,3 +51,36 @@ jobs:
...
+                      echo "should_run=${{contains(steps.PR.outputs.pr_body, 'test-ai')}}" >> "$GITHUB_OUTPUT"
+                  else
+                      echo "should_run=false" >> "$GITHUB_OUTPUT"
+                  fi
+
+            - name: Run AI Agent Tests
+              if: steps.ai_changes.outputs.ai == 'true' || steps.ai_override.outputs.should_run == 'true'
+              run: pnpm --filter backend test -- --testPathPattern="src/ee/services/ai/agents/tests/agent.test.ts"
+              env:
+                  OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
tatianainama, line 84: I've created a dedicated key for tests, it's in a new project in the organization with credits limits in place 🦺

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
```diff
@@ -0,0 +1,424 @@
...
+};
+
+export const createArgs = ({
+    agentSettings,
+    messageHistory,
+    ...options
+}: Partial<Omit<AiAgentArgs, 'agentSettings'>> & {
+    agentSettings?: Partial<AiAgentArgs['agentSettings']>;
+} = {}) => {
+    const { OPENAI_API_KEY, OPENAI_MODEL_NAME = 'gpt-4o' } = process.env;
```
IrakliJani: gpt-4.1 should be used. I also advise you to switch from 4o to 4.1 locally

tatianainama: oh good! will update!
Side note: would it be valuable to test responses with different models too?

IrakliJani: yes, eventually, but not now

.github/workflows/test-and-build.yml
```diff
@@ -51,3 +51,36 @@ jobs:
 
             - name: Run unit tests
               run: pnpm test
+
+            # AI Agent Tests to run only after file changes
+            - name: Check for AI files changes
```
IrakliJani, line 56: I think it'll be good to extract this out of test and build and have a separate workflow. wdyt?

.github/workflows/test-and-build.yml
```diff
@@ -51,3 +51,36 @@ jobs:
...
+                  if [ "${{ github.event_name }}" = "pull_request" ]; then
+                      echo "AI override: ${{contains(steps.PR.outputs.pr_body, 'test-ai')}}"
+                      echo "should_run=${{contains(steps.PR.outputs.pr_body, 'test-ai')}}" >> "$GITHUB_OUTPUT"
+                  else
+                      echo "should_run=false" >> "$GITHUB_OUTPUT"
+                  fi
+
+            - name: Run AI Agent Tests
+              if: steps.ai_changes.outputs.ai == 'true' || steps.ai_override.outputs.should_run == 'true'
+              run: pnpm --filter backend test -- --testPathPattern="src/ee/services/ai/agents/tests/agent.test.ts"
```
IrakliJani: would be nice to have it:
- under pnpm `test:ai`
- include everything under `services/ai`
- exclude services/ai from `pnpm test`

.github/workflows/test-and-build.yml
IrakliJani: cool. suggestion from graphite is tempting, but I’m not super set on it either way.

.github/workflows/test-and-build.yml
```diff
@@ -51,3 +51,36 @@ jobs:
...
+              env:
+                  OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+                  OPENAI_MODEL_NAME: 'gpt-4o'
```
IrakliJani: gpt-4.1

.github/workflows/test-and-build.yml
IrakliJani: scrap that... I think this triggers it? and does not actually run the test?

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
```diff
@@ -0,0 +1,197 @@
...
+
+        describe('Basic functionality', () => {
+            it(
+                'correctly returns all the models it can use',
+                async () => {
+                    const message = createMessage('What models can you use?');
+                    const args = createArgs({
+                        messageHistory: [message],
+                        agentSettings: {
+                            instruction: `When asked about models you can use, only respond with a list of the models you can see`,
```
IrakliJani: nice use of instructions!

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
IrakliJani: I feel like we are dragging this env variable a lot. should this be something that you pass  to createArgs maybe?

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
```diff
@@ -0,0 +1,424 @@
...
+export const createMessage = (
+    content: string,
+    role: 'user' | 'assistant' = 'user',
+) => ({
+    role,
+    content,
+});
+
+const capitalize = (str: string): string =>
+    str.charAt(0).toUpperCase() + str.slice(1);
```
IrakliJani: what do we use in actual code?

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
```diff
@@ -0,0 +1,424 @@
...
+    MetricType,
+} from '@lightdash/common';
+import { getOpenaiGptmodel } from '../../models/openai-gpt';
+import type { AiAgentArgs, AiAgentDependencies } from '../../types/aiAgent';
+import { AiAgentExploreSummary } from '../../types/aiAgentExploreSummary';
+
+export type ConversationHistory = Array<{
+    role: 'user' | 'assistant';
+    content: string;
+}>;
```
IrakliJani: I think you should reuse types from `getChatHistoryFromThreadMessages`

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
```diff
@@ -0,0 +1,197 @@
...
+        });
+    });
+};
+
+if (!OPENAI_API_KEY) {
+    describe('skipping AI agent integration tests - OPENAI_API_KEY is not set', () => {
+        it('skips', () => {
+            expect(true).toBe(true);
+        });
+    });
```
IrakliJani: should be removed I guess?

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
```diff
@@ -0,0 +1,197 @@
...
+                    promptTestUtils.expectCorrectToolUsage(deps, [
+                        'getExplores',
+                        'getExplore',
+                        'searchFields',
+                        'getPrompt',
+                    ]);
+                    promptTestUtils.expectResponseWithContent(
+                        thirdResponse,
+                        'bar chart',
+                    );
```
IrakliJani: is this checking if text contains bar chart or not?

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
```diff
@@ -0,0 +1,197 @@
...
+                        args: thirdArgs,
+                        dependencies: deps,
+                    });
+
+                    promptTestUtils.expectCorrectToolUsage(deps, [
+                        'getExplores',
+                        'getExplore',
+                        'searchFields',
+                        'getPrompt',
+                    ]);
```
IrakliJani: If the full conversation history is available here — including tool calls — then it really shouldn't be calling `getExplores`, `getExplore`, `searchFields`, or `getPrompt`, since all that info should already be accessible from the message history.  

It’s probably not behaving that way because we’re not including tool call results in the history.

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
IrakliJani: Ideally, this should be a test case where you check that bar chart tool has been called

.github/workflows/test-and-build.yml
tatianainama: I [thought](https://app.graphite.dev/github/pr/lightdash/lightdash/15883/chore-split-build%2C-unit-tests-and-ai-tests-into-distintic-workflows) of separating this into a different workflow, and I started playing around with it:

- build workflow -> creates artifact with build
  - test workflow -> runs all unit tests
  - test ai workflow -> runs AI tests if needed

It has the benefit of separation of concerns and avoids making this file huge. In terms of performance, it only allows us to run the unit + AI tests in parallel, but don't think it gains us much, as the build process is the the one that takes the longest

.github/workflows/test-and-build.yml
tatianainama: That means also using a different key for local development? so we will have two, unless if there is no `OPENAI_TEST_API_KEY`, use `OPENAI_API_KEY` ?

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: yeah it only checks if there is a bar chart text 😞 
Agree! calling the tool is a better approach, I'll work on that in a followup!

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: all tests files need to contain a test run, otherwise Jest complains 😬 I could remove this and let it crash if you are running this locally without the variable!

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
tatianainama: another capitalize, I'll update to use that!

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: good point, this is because I'm not storing the tool calls/results.
I'll update the `storeToolCall/Results` mock to keep track of this, so I can re-feed the history correctly

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: removed now!

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: updated!

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: storing tool calls now. I found that the second question is still using find explores and fields 🤔 need to investigate more

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
tatianainama: created a factory to solve this

.github/workflows/test-and-build.yml
IrakliJani: actually, this will be confusing for everyone. let's just use `OPENAI_API_KEY`

packages/backend/jest.config.dev.js
```diff
@@ -1,21 +1,43 @@
-module.exports = {
+const common = {
     preset: 'ts-jest',
     testEnvironment: 'node',
     automock: false,
     setupFiles: ['./setupJest.ts'],
-    testPathIgnorePatterns: ['/node_modules/', '/dist/'],
+    moduleNameMapper: {
+        '^@/(.*)$': '<rootDir>/src/$1',
```
IrakliJani: what's this for?

IrakliJani: oh it's not new. I don't think we've used @/aliasing

.github/workflows/test-and-build.yml
IrakliJani: gotcha, thanks for the explanation 👍

.github/workflows/test-and-build.yml
```diff
@@ -51,3 +51,36 @@ jobs:
...
+            # AI Agent Tests to run only after file changes
+            - name: Check for AI files changes
+              uses: dorny/paths-filter@v3
+              id: ai_changes
+              with:
+                  token: ${{ github.token }}
+                  filters: |
+                    ai:
+                      - 'packages/backend/src/ee/services/ai/**/*'
+                      - 'packages/common/**/*'
```
IrakliJani: should this also be in ee ai folder in common instead of **/* ?

.github/workflows/test-and-build.yml
```diff
@@ -51,3 +51,36 @@ jobs:
...
+            - uses: 8BitJonny/gh-get-current-pr@3.0.0
+              id: PR
+              if: github.event_name == 'pull_request'
+
+            - name: Check for AI test override if the PR description contains "test-ai"
+              id: ai_override
+              run: |
+                  if [ "${{ github.event_name }}" = "pull_request" ]; then
+                      echo "AI override: ${{contains(steps.PR.outputs.pr_body, 'test-ai')}}"
+                      echo "should_run=${{contains(steps.PR.outputs.pr_body, 'test-ai')}}" >> "$GITHUB_OUTPUT"
```
IrakliJani, line 75: this is really good 👍

packages/backend/jest.config.js
```diff
@@ -1,7 +1,29 @@
-module.exports = {
+const common = {
     preset: 'ts-jest',
     testEnvironment: 'node',
     automock: false,
     setupFiles: ['./setupJest.ts'],
-    testPathIgnorePatterns: ['/node_modules/', '/dist/'],
+    moduleNameMapper: {
+        '^@/(.*)$': '<rootDir>/src/$1',
```
IrakliJani: okay that means we do use it.. but I don't see it anywhere

packages/backend/package.json
```diff
@@ -154,9 +154,11 @@
...
         "scheduler": "node dist/scheduler.js",
         "scheduler-dev": "tsx watch --clear-screen=false src/scheduler.ts",
-        "test": "TZ=UTC jest --config jest.config.js",
-        "test-sequential": "TZ=UTC NODE_ENV=test jest --config jest.config.js --runInBand",
-        "test:dev": "TZ=UTC jest --config jest.config.dev.js --onlyChanged --watch",
+        "test": "TZ=UTC jest --config jest.config.js --selectProjects main",
+        "test-sequential": "TZ=UTC NODE_ENV=test jest --config jest.config.js --runInBand --selectProjects main",
+        "test:dev": "TZ=UTC jest --config jest.config.dev.js --onlyChanged --watch --selectProjects main",
+        "test:ai": "TZ=UTC jest --config jest.config.js --selectProjects ai",
+        "test:ai:dev": "TZ=UTC jest --config jest.config.dev.js --onlyChanged --watch --selectProjects ai",
```
IrakliJani, line 161: nice!

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    createMockDepsFactory,
+    createToolMessage,
+    createToolMessages,
+    mockExploresSummary,
+    promptTestUtils,
+} from './testUtils';
+
+const TIMEOUT = 30_000;
+const createArgs = createArgsFactory({
+    apiKey: process.env.OPENAI_API_KEY!,
```
IrakliJani: you can throw somewhere above instead of !

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    describe('Basic functionality', () => {
+        it(
+            'correctly returns all the models it can use',
+            async () => {
+                const message = createMessage('What models can you use?');
+                const args = createArgs({
+                    messageHistory: [message],
+                    agentSettings: {
+                        instruction: `When asked about models you can use, only respond with a list of the models you can see`,
```
IrakliJani: I am curious to remove all special instructions and see how agent performs. we'll never have that kind of instructions in real scenarios.

packages/backend/src/ee/services/ai/agents/tests/testUtils.ts
```diff
@@ -0,0 +1,572 @@
...
+            type: 'number',
+            fieldType: 'metric',
+        },
+    },
+    cacheMetadata: {},
+}));
+
+export const createMockDepsFactory = () => {
+    const toolCalls = new Map<string, Parameters<StoreToolCallFn>[0]>();
+    const toolResults = new Map<string, Parameters<StoreToolResultsFn>[0][0]>();
```
IrakliJani, line 478: ok this is great, so you keep tool calls in between loops.

I think in the follow up PRs we should do the same for prompts and mock implementation of conversation history better.

we should also use tool call args (json generated by the llm) to verify selecting proper metrics and dimensions, proper chart visualization config and more.

IrakliJani: Ideally, it should not check text responses and should instead rely entirely on tool calls.

IrakliJani (approved):
Approved, left some minor suggestions (no blockers), questions, and some opinions on the upcoming work.

packages/backend/jest.config.js
tatianainama: I did a quick search, it doesn't look like we actively use it, but it's also added in the tsconfig. Maybe it was a default project configuration.

packages/backend/src/ee/services/ai/agents/tests/agent.test.ts
tatianainama: that's a good point! It should still reply with all the models but more verbose. I'm gonna remove and see

.github/workflows/test-and-build.yml
tatianainama: great catch 🙏

packages/backend/jest.config.js
tatianainama: I removed it, if we needed in the future, we can add it 🧹

tatianainama:
### Merge activity

* **Jul 22, 8:56 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15854).
* **Jul 22, 8:56 AM UTC**: @tatianainama merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/15854).

PR #15873, opened 2025-07-15
feat: lightdash parameters

almeidabbm:
### Description:

Lightdash parameters main work. This PR makes parameters available in YML and causes the UI to pick them up in explores and dashboards. There is more to do, but this is essentially auto-feature-flagged. If no parameters are included, the UI doesn't show them.


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1710,11 +1730,16 @@ export class ProjectService extends BaseService {
     async compileQuery(
         args: {
             user: SessionUser;
-            metricQuery: MetricQuery;
+            // ! TODO: we need to fix this type
+            body: MetricQuery & { parameters?: ParametersValuesMap };
```
magnew, line 1734: Does this comment mean we should add `parameters?` to Metric query, or is there some other fix needed?

packages/common/src/utils/loadLightdashProjectConfig.ts
```diff
@@ -35,6 +35,10 @@ export const loadLightdashProjectConfig = async (
         );
     }
 
+    if (configFile.parameters == null) {
```
magnew, line 38: Looks like this will come back null from ajv. I'm just making it match the expected TS type.

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego: Parameters should NOT go in the MetricQuery type. This union was a way to keep the endpoint backwards compatible. 😢

magnew (approved):
💥

PR #1234, opened 2022-01-25
feat: disable password authentication

nathaliabj:
### Pre requirements:
- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/lightdash/lightdash/blob/main/.github/CONTRIBUTING.md#sending-a-pull-request).

### Reference:
Closes: #1150 

### Description:
- remove login option with email/password 
- remove create user option with email/password 
- hide password tb from user settings
- throw error if a user is created wit email/password

Preview:

Screenshot 2022-01-26 at 11 56 34

### Scope of the changes (select all that apply):
- [x] Frontend  
- [x] Backend  
- [x] Common  
- [ ] E2E tests  
- [ ] Docs  
- [ ] CI/CD

PR #15889, opened 2025-07-17
feat: add OAuth2 server for CLI authentication

rephus:
Closes: #1234

### Description:
Implements OAuth2 server for CLI authentication, replacing the current API key copy-paste approach with a more secure and user-friendly method. This PR:

- Adds OAuth2 server implementation with Authorization Code flow and PKCE support
- Creates database tables for OAuth tokens, authorization codes, and clients
- Implements OAuth endpoints for authorization, token exchange, introspection, and revocation
- Updates CLI to support OAuth-based login with browser authentication
- Adds middleware to validate OAuth tokens for API requests
- Generates Personal Access Tokens automatically after OAuth authentication

The new flow allows users to authenticate by opening a browser window rather than copying and pasting tokens, improving security and user experience.


rephus:
* **#15889** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15889?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 7  dependent PRs ([#15902](https://github.com/lightdash/lightdash/pull/15902) <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15902?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#15906](https://github.com/lightdash/lightdash/pull/15906) <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15906?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#15908](https://github.com/lightdash/lightdash/pull/15908) <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15908?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 4 others) 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15889?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/controllers/authentication/middlewares.ts
```diff
@@ -47,6 +48,14 @@ export const unauthorisedInDemo: RequestHandler = (req, res, next) => {
     }
 };
 
+export const allowOauthAuthentication: RequestHandler = (req, res, next) => {
```
rephus, line 51: new middleware used to create PAT

packages/backend/src/models/OAuthModel.ts
```diff
@@ -0,0 +1,261 @@
...
+export class OAuthModel {
+    private readonly database: Knex;
+
+    constructor({ database }: OAuthModelArguments) {
+        this.database = database;
+    }
+
+    // There is going to be a single client
+    // for all instances and customers
+    async getClient(): Promise<OAuthClient | null> {
```
rephus: To simplify some Oauth process, we hardcoded a Client, I did this on the code, rather than creating a migration and adding a default value, in case we need to change some of the specification, it will be easier to update the code than to create migrations to update the default app in db. 

We don't expetc to support multiple clients anytime soon.

packages/backend/src/App.ts
```diff
@@ -651,6 +652,18 @@ export default class App {
                     },
                 });
 
+                // Check if this is an OAuth endpoint and return OAuth2-compliant error response
+                if (error instanceof OauthAuthenticationError) {
```
rephus, line 658: Oauth2 has a very specific error typ errors, so if the error is from OauthAuthenticationError (new type) , then we will change the response type

Another option is to do this based on the req URL, rather than the error type

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+
+const oauthRouter = express.Router({ mergeParams: true });
+
+// Get OAuth service from request
+function getOAuthService(req: express.Request): OAuthService {
+    return req.services.getOauthService();
+}
+
+// Get authorization - use OAuth2Server
+oauthRouter.get('/authorize', async (req, res, next) => {
```
rephus, line 19: I built a router directly on express and not a tsoa controller because this oauth2server depends on the req,res which are not easily available on tsoa

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (!req.user) {
+        const loginUrl = `/login?redirect=${encodeURIComponent(
+            req.originalUrl || req.url,
+        )}`;
+        return res.redirect(loginUrl);
```
rephus, line 24: if user is not authenticated, we redirect to /login page

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        state,
+        code_challenge,
+        code_challenge_method,
+    } = req.query;
+    if (!client_id || !redirect_uri || !scope) {
+        return res.status(400).send('Missing required parameters');
+    }
+    // Render authorize page
+    res.set('Content-Type', 'text/html');
+    return res.send(`
```
rephus: it is a good practice to show a authorization page, even if the user is already authenticated. 

Alternatively, We could pass an argument to skip this window for the client

Screenshot from 2025-07-21 12-46-24

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+                            <button type="submit" name="approve" value="false" class="oauth-btn deny">Deny</button>
+                        </div>
+                    </form>
+                </div>
+            </body>
+        </html>
+    `);
+});
+
+oauthRouter.post('/authorize', async (req, res, next) => {
```
rephus, line 85: the real oauth2 authorization happens after the access has been authorized, in a post endpoint.

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        url.searchParams.set('error', 'access_denied');
+        if (req.body.state) {
+            url.searchParams.set('state', req.body.state);
+        }
+        return res.redirect(url.toString());
+    }
+});
+
+// Post token - use OAuth2Server
+oauthRouter.post('/token', async (req, res, next) => {
```
rephus, line 145: with the authorization token, we use this endpoint to get access token or even refresh token. That logic is all handled in the oauth2server

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+            errorMessage.includes('Invalid') ||
+            errorMessage.includes('required')
+                ? 401
+                : 400;
+        res.status(statusCode).json({ error: errorMessage });
+    }
+});
+
+// Token introspection endpoint
+oauthRouter.post('/introspect', async (req, res) => {
```
rephus, line 180: standard endpoint on oauth2 to validate an existing token.

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
```diff
@@ -0,0 +1,96 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS oauth2_clients (
```
rephus, line 5: I find this migration more readable in raw format.

rephus: Since we only need 1 client, we could harcode getClient, and avoid writting a new table and a single client in the database. 

However, it was more complex to manually adapt the rest of the calls and types to not use a client_id in the database, than to keep it.

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
```diff
@@ -0,0 +1,96 @@
...
+        CREATE INDEX IF NOT EXISTS oauth2_access_tokens_expires_at_idx ON oauth2_access_tokens(expires_at);
+        
+        CREATE INDEX IF NOT EXISTS oauth2_refresh_tokens_client_id_idx ON oauth2_refresh_tokens(client_id);
+        CREATE INDEX IF NOT EXISTS oauth2_refresh_tokens_user_id_idx ON oauth2_refresh_tokens(user_id);
+        CREATE INDEX IF NOT EXISTS oauth2_refresh_tokens_expires_at_idx ON oauth2_refresh_tokens(expires_at);
+
+        INSERT INTO oauth2_clients (client_id, client_secret, redirect_uris, grants, scopes, client_name) 
+        VALUES (
+            'lightdash-cli',
+            'secret',
```
rephus: we could keep this `secret` empty, but it seems some oauth methdos like "refresh_token" enforce a secret, otherwise, client can't be validated. So I added this one here, still, most actions don't need a secret.

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
```diff
@@ -0,0 +1,96 @@
...
+            ARRAY[${Array.from(
+                { length: 11 },
+                (_, i) => `'http://localhost:${8100 + i}/callback'`,
```
rephus: whitelist ports between 8100 and 8110. Wildwards don't work

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
```diff
@@ -0,0 +1,96 @@
...
+            ).join(', ')}],
+            ARRAY['authorization_code', 'refresh_token'],
```
rephus: For our flow, we don't need refresh_token, but since we want to make this as compatible as possible, I added, implemented and tested refresh_token too

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+        super();
+        this.userModel = userModel;
+        this.oauthModel = oauthModel;
+        this.lightdashConfig = lightdashConfig;
+        this.initializeOAuthServer();
+    }
+
+    private initializeOAuthServer(): void {
+        this.oauthServer = new OAuth2Server({
+            model: this.oauthModel,
```
rephus, line 37: this is how we initialize the oauth2server, the model is a file with some endpoints implemented, for the library to use. https://node-oauthoauth2-server.readthedocs.io/en/master/model/overview.html

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+        user: UserWithOrganizationUuid,
+    ): Promise<OAuth2Server.AuthorizationCode> {
+        return this.oauthServer.authorize(request, response, {
+            authenticateHandler: {
+                handle: () => user,
+            },
+        });
+    }
+
+    public async token(
```
rephus, line 66: this methods are a simple wrapper to be used on the `oauthRouter` , without using directly the oauth Server. We could simplify the code by removing this abstraction.

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public async revokeToken(token: string): Promise<boolean> {
+        // Try to revoke as access token first
+        const refreshToken = await this.oauthModel.getRefreshToken(token);
+        if (refreshToken) return this.oauthModel.revokeToken(refreshToken);
+        return false;
+    }
+
+    public async authenticateWithOauthToken(
```
rephus: This is a custom fucntion to use appToken on the middleware for authentication. At the moment we only allow this on the create PAT endpoint

packages/cli/src/handlers/login.ts
```diff
@@ -11,16 +11,26 @@ import { configFilePath, setContext, setDefaultUser } from '../config';
...
 import { checkLightdashVersion } from './dbt/apiClient';
+import { loginWithOauth } from './oauthLogin';
 import { setFirstProject, setProjectCommand } from './setProject';
 import { buildRequestHeaders } from './utils';
 
 type LoginOptions = {
     /** Associated with a Personal Access Token or Service Account Token */
     token?: string;
+    /** Use OAuth2 flow instead of password/token */
+    oauth?: boolean;
```
rephus, line 19: small changes to allow a new login method on CLI ( --oauth)

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+
+    if (availablePort === null || availablePort === undefined) {
+        throw new Error('No available ports found');
+    }
+
+    return availablePort;
+};
+
+// Helper function to open browser
+const openBrowser = async (url: string): Promise<void> => {
```
rephus, line 19: platform dependant, Tested on linux (ubuntu) and mac. If this fails, we still show the URL, so users can open it manually.

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+            const callbackUrl = new URL(req.url, `http://localhost:${port}`);
+            const code = callbackUrl.searchParams.get('code');
+            const returnedState = callbackUrl.searchParams.get('state');
+            const error = callbackUrl.searchParams.get('error');
+
+            if (error) {
+                rejectAuth(new AuthorizationError(`OAuth error: ${error}`));
+                res.writeHead(400, { 'Content-Type': 'text/html' });
+                res.end(`
+                    <html>
```
rephus: html with error response 

Screenshot from 2025-07-18 14-53-31

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+
+            if (!code || !returnedState) {
+                rejectAuth(
+                    new AuthorizationError(
+                        'Missing authorization code or state',
+                    ),
+                );
+                res.writeHead(400, { 'Content-Type': 'text/html' });
+                res.end(`
+                    <html>
```
rephus, line 177: html for successful oauth flow

Screenshot from 2025-07-18 14-53-57

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+            res.writeHead(200, { 'Content-Type': 'text/html' });
+            res.end(`
+                <html>
+                    <head>
+                        <style>${oauthPageStyles}</style>
+                    </head>
+                    <body>
+                        <div class="stack">
+                           
+                            <div class="container success">
```
rephus: html for successful oauth flow 

Screenshot from 2025-07-18 14-53-57

packages/common/src/utils/oauth.ts
```diff
@@ -0,0 +1,60 @@
+export const oauthPageStyles = `
```
rephus, line 3: custom styles to be used in both the router and the cli page.

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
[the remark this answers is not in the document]
stevelikesmusic: Should this be an env var?

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
stevelikesmusic: Fantastic

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
stevelikesmusic: I appreciate this approach. Are we only anticipating using the OAuth with the CLI for now? Would the MCP server count as another client? Would there be a scenario where embedding customers might want OAuth?

packages/backend/src/generated/routes.ts
```diff
@@ -1826,6 +1826,7 @@ const models: TsoaRoute.Models = {
             isAdditionalDimension: { dataType: 'boolean' },
             colors: { ref: 'Record_string.string_' },
             isIntervalBase: { dataType: 'boolean' },
+            aiHint: { dataType: 'string' },
```
stevelikesmusic: Do we need a precommit or CI check to generate a new schema if a Controller (or its underlying types) has changed?

packages/backend/src/routers/oauthRouter.ts
stevelikesmusic: Good ol' Sir Attenborough

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        code_challenge,
+        code_challenge_method,
+    } = req.query;
+    if (!client_id || !redirect_uri || !scope) {
+        return res.status(400).send('Missing required parameters');
+    }
+    // Render authorize page
+    res.set('Content-Type', 'text/html');
+    return res.send(`
+        <html>
```
stevelikesmusic: Would it be worth pulling this out as its own HTML file? Or just thinking it's not worth installing a template engine this one page?

packages/backend/src/models/OAuth2Model.ts
```diff
@@ -0,0 +1,300 @@
+/* eslint-disable class-methods-use-this */
+import { AuthTokenPrefix, UserWithOrganizationUuid } from '@lightdash/common';
+import type {
+    AuthorizationCode,
+    AuthorizationCodeModel,
+    Client,
+    Token,
+    User,
+} from '@node-oauth/oauth2-server';
+import { createHash } from 'crypto';
```
stevelikesmusic: Is this used?

packages/backend/src/models/OAuth2Model.ts
```diff
@@ -0,0 +1,300 @@
...
+    async getClient(
+        clientId: string,
+        clientSecret?: string | null,
+    ): Promise<Client | false> {
+        const query = this.database('oauth2_clients')
+            .select('*')
+            .where('client_id', clientId);
+
+        if (clientSecret) {
+            void query.andWhere('client_secret', clientSecret);
```
stevelikesmusic, line 28: Right now we have 1 client with a plaintext secret: `'secret`'. Would it be worth encrypting these values now should we want to add new clients in the future and not need to remember that secrets are stored plaintext?

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (!req.user) {
+        return res.status(401).send('Unauthorized');
+    }
+    if (!req.user.userId || !req.user.organizationUuid) {
+        return res.status(400).send('Missing userUuid or organizationUuid');
+    }
+    if (req.body.approve === 'true') {
+        const oauthService = getOAuthService(req);
+        const oauthReq = new OAuth2Server.Request(req);
+        const oauthRes = new OAuth2Server.Response(res);
```
stevelikesmusic: Is this the bulk of what `Oauth2Server` does?

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+            });
+        }
+        const url = new URL(redirectUri);
+        url.searchParams.set('error', 'access_denied');
+        if (req.body.state) {
+            url.searchParams.set('state', req.body.state);
+        }
+        return res.redirect(url.toString());
+    }
+});
```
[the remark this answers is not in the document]
stevelikesmusic: Are we deferring rate limiting to GCP?

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+
+// Post token - use OAuth2Server
+oauthRouter.post('/token', async (req, res, next) => {
+    const oauthService = getOAuthService(req);
+    const oauthReq = new OAuth2Server.Request(req);
+    const oauthRes = new OAuth2Server.Response(req);
```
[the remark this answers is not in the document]
stevelikesmusic: That's a nice catch

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        console.error('Token endpoint error:', error);
+        const errorMessage =
+            error instanceof Error ? error.message : 'Unknown error';
+
+        // Return 401 for authentication errors, 400 for other errors
+        const statusCode =
+            errorMessage.includes('Invalid') ||
+            errorMessage.includes('required')
+                ? 401
+                : 400;
```
stevelikesmusic, line 174: Would it be worth throwing an error that returns the status so we get a stack trace?

packages/backend/src/routers/oauthRouter.ts
stevelikesmusic: Is this endpoint being called by the client or by OAuth2Server?

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+                iat: Math.floor(
+                    tokenData.accessTokenExpiresAt.getTime() / 1000 - 3600, // issued at: token cretaed timestamp
+                ),
+                sub: tokenData.user.userUuid, // subject: Unique user identifier
+                aud: tokenData.client.id, // audience: Client identifier
+                iss: 'lightdash', // issuer
+                jti: tokenData.accessToken, // JWT ID: Unique token identifier
+                client_id: DEFAULT_OAUTH_CLIENT_ID,
+                username: tokenData.user.userUuid,
+            };
```
stevelikesmusic, line 216: We should require these claims in the embedded JWT ❤️

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        await oauthService.revokeToken(token);
+    } catch (error) {
+        console.error('Revocation error:', error);
+    }
+    return res.status(200).send();
+});
+
+// OAuth2 Discovery endpoint
+oauthRouter.get('/.well-known/oauth-authorization-server', (req, res) => {
+    const baseUrl = `${req.protocol}://${req.get('host')}`;
```
stevelikesmusic: Do we support `http`? Or is that up to customers that might be hosting themselves?

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+        req: Request,
+        next: NextFunction,
+    ): Promise<void> {
+        try {
+            const token = req.headers.authorization;
+            if (!token || typeof token !== 'string' || token.length === 0) {
```
stevelikesmusic: Wouldn't `!token` evaluate `false` for an empty string? No need to have `token.length === 0` in that case?

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+    let rejectAuth: (reason: Error) => void;
+    const authPromise = new Promise<{ code: string; state: string }>(
+        (resolve, reject) => {
+            resolveAuth = resolve;
+            rejectAuth = reject;
+        },
+    );
+
+    // Create HTTP server to handle the callback
+    const server = http.createServer((req, res) => {
```
stevelikesmusic, line 55: Very cool

packages/cli/src/handlers/oauthLogin.ts
stevelikesmusic: Is it worth using a templating engine like handlebars for these messages?

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+                    reject();
+                });
+            });
+            return port;
+        } catch {
+            return null;
+        }
+    };
+
+    const results = await Promise.all(ports.map(checkPort));
```
stevelikesmusic: Should we use `.allSettled`? Promise.all will fail on the first rejected promise it receives.

packages/e2e/cypress/e2e/api/oauthLogin.cy.ts
```diff
@@ -0,0 +1,805 @@
...
+describe('OAuth API Integration Tests', () => {
+    const apiUrl = '/api/v1/oauth';
+
+    // PKCE test values - these are pre-computed for testing
+    const PKCE_TEST_VALUES = {
+        codeVerifier: 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk',
+        codeChallenge: 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM',
+    };
+
+    // No global login - each test will handle its own authentication state
```
stevelikesmusic, line 12: Nice

stevelikesmusic (commented):
So much good stuff here Javi! 👏 👏 👏 

Just a few questions

rephus:
Hashing on DB: 
- authorization_code
- app_token
- refresh_token

Tested to login using the CLI, and it seems to be working as expected

Tests are still pàssing, they are high level enough to not care about the value in db

rephus:
> Is it worth using a templating engine like handlebars for these messages?

Yes, I just did it before reading your comment :brain:

packages/cli/src/handlers/oauthLogin.ts
rephus: Promise.all seems to work..  :eyes: I think because we are doing a try-catch

image

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
rephus: At the moment, we don't support new clients, if we need it, this will be done on a separate PR if possible

packages/backend/src/routers/oauthRouter.ts
rephus: neither, this is a oauth2 standard endpoint used to validate tokens, we don't use it on the CLI

Same as .well-known endpoint.

packages/cli/src/handlers/oauthLogin.ts
rephus: I've just exctracted the error message in a simple function. I believe the HTML is simple enough so we don't need any library.

packages/backend/src/services/OAuthService/OAuthService.ts
owlas: Is there a library we can use for this? Since validating the token provided (including checking expiry etc.) is standardised? Once the token is validated etc. then the call to getSessionUser would be lightdash specific

packages/backend/src/controllers/authentication/middlewares.ts
```diff
@@ -3,6 +3,7 @@
 import {
     ApiError,
     AuthorizationError,
+    AuthTokenPrefix,
```
owlas: unused param

packages/backend/src/controllers/authentication/middlewares.ts
```diff
@@ -47,6 +48,14 @@ export const unauthorisedInDemo: RequestHandler = (req, res, next) => {
     }
 };
 
+export const allowOauthAuthentication: RequestHandler = (req, res, next) => {
+    if (req.isAuthenticated()) {
+        next();
+        return;
+    }
+
+    void req.services.getOauthService().authenticateWithOauthToken(req, next);
```
owlas: can we make this so that http specific handling is kept in the function and the oauth service logic operates without needing to know about the http framework - usually services don't handle raw requests

owlas: I like the pattern of getting services off the request though! Tidy

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+        this.userModel = userModel;
+        this.oauthModel = oauthModel;
+        this.lightdashConfig = lightdashConfig;
+        this.initializeOAuthServer();
+    }
+
+    private initializeOAuthServer(): void {
+        this.oauthServer = new OAuth2Server({
+            model: this.oauthModel,
+            accessTokenLifetime: 60 * 60,
```
owlas: Let's add these as env vars

packages/backend/src/services/OAuthService/OAuthService.ts
owlas: i like this because we could replace the underlying library if we needed to!

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+};
+
+export class OAuthService extends BaseService {
+    protected oauthServer!: OAuth2Server;
+
+    private userModel: UserModel;
+
+    private oauthModel: OAuth2Model;
+
+    private lightdashConfig: LightdashConfig;
```
owlas, line 21: not being used but I guess you'll want it to parameterise based on the env vars

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+        throw new AuthorizationError(
+            `Failed to create PAT: ${patResponse.status} ${patResponse.statusText}`,
+        );
+    }
+    const patResponseBody = await patResponse.json();
+    const { token } = patResponseBody.results;
+    return token;
+};
+// Helper function to find an available port
+const findAvailablePort = async (
```
owlas: i'm pretty sure the standard node server starts on a random high number port if it's not specified! we shouldn't need this custom port finding

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+import { promisify } from 'util';
+import GlobalState from '../globalState';
+import * as styles from '../styles';
+
+// Shared CSS styles for OAuth response pages - matching Lightdash login design
+
+const execAsync = promisify(exec);
+
+// PKCE helper functions
+const generateCodeVerifier = (): string =>
```
owlas: we should use a library for all of this, we typically use the openid-client library, which contains all of this logic

packages/cli/src/handlers/oauthLogin.ts
```diff
@@ -0,0 +1,374 @@
...
+    crypto.randomBytes(32).toString('base64url');
+
+const generateCodeChallenge = (verifier: string): string =>
+    crypto.createHash('sha256').update(verifier).digest('base64url');
+
+const generatePersonalAccessToken = async (
```
owlas: my assumption was that the OAuth flow would return the personal access token AS the access token directly, is that not possible?

packages/backend/src/models/OAuth2Model.ts
owlas: ideally this would operate on hashes too, so we never need to store the plain secret

packages/backend/src/models/OAuth2Model.ts
```diff
@@ -0,0 +1,300 @@
...
+                grants: result.grants,
+            },
+            user: {
+                userId: result.user_id,
+                organizationUuid: result.organization_uuid,
+            },
+        };
+    }
+
+    async generateAccessToken(
```
owlas, line 267: Is this where we could conditionally generate a personal access token instead of something else depending on the scopes provided/

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+    ): Promise<void> {
+        try {
+            const token = req.headers.authorization;
+            if (!token || typeof token !== 'string' || token.length === 0) {
+                next();
+                return;
+            }
+            const tokenParts = token.split(' ');
+            const oauthToken = tokenParts[1];
+            const isValidOauthToken =
```
owlas: How come this looks inverted? It's checking if it doesn't start with OAUTH_APP ? is that right?

packages/backend/src/routers/oauthRouter.ts
owlas: We should use a templating engine yeah, even a lightweight one because otherwise this has XSS vulnerability, currently anything can be injected into the HTML through the client_id - since we use raw string formatting

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
owlas: +1 to having the getClients model in the backend return both fixed clients (where the client_secret is generated from env vars) and dynamic clients, which are stored here

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
owlas: which part of the stack doesn't support wild cards? In the oauth standard it should be allowed

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
owlas: If you change the getClient model to inject fixed clients, you don't need to seed the table here with a value, which is good. I don't see a good dev experience for securely inserting a secret into the database here

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -0,0 +1,126 @@
...
+import { UserModel } from '../../models/UserModel';
+import { BaseService } from '../BaseService';
+
+type OAuthServiceArguments = {
+    userModel: UserModel;
+    oauthModel: OAuth2Model;
+    lightdashConfig: LightdashConfig;
+};
+
+export class OAuthService extends BaseService {
```
owlas, line 14: We should log these service calls in detail!

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        scope,
+        state,
+        code_challenge,
+        code_challenge_method,
+    } = req.query;
+    if (!client_id || !redirect_uri || !scope) {
+        return res.status(400).send('Missing required parameters');
+    }
+    // Render authorize page
+    res.set('Content-Type', 'text/html');
```
owlas, line 38: this is missing a lot of the security headers we apply when we server static content from the UI (see the express root router which has the security headers)

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+        const oauthService = getOAuthService(req);
+        await oauthService.revokeToken(token);
+    } catch (error) {
+        console.error('Revocation error:', error);
+    }
+    return res.status(200).send();
+});
+
+// OAuth2 Discovery endpoint
+oauthRouter.get('/.well-known/oauth-authorization-server', (req, res) => {
```
owlas, line 245: is this not something the library provides for us?

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,302 @@
...
+    const baseUrl = `${req.protocol}://${req.get('host')}`;
+
+    res.json({
+        issuer: baseUrl,
+        authorization_endpoint: `${baseUrl}/api/v1/oauth/authorize`,
+        token_endpoint: `${baseUrl}/api/v1/oauth/token`,
+        introspection_endpoint: `${baseUrl}/api/v1/oauth/introspect`,
+        revocation_endpoint: `${baseUrl}/api/v1/oauth/revoke`,
+        response_types_supported: ['code'],
+        grant_types_supported: [
```
owlas, line 255: we aren't supporting client_credentials currently.

owlas (changes requested):

packages/cli/src/handlers/oauthLogin.ts
rephus: I tried already, but let me try again if there is an argument I was missing...

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
rephus: all other tables for oauth reference a client to be in db. Returning afixed client on the service will not work with all these foreign keys (I tried). It is possible to remove the client reference, but that will make adding more clients harder/not compatible with oauth.

packages/backend/src/database/migrations/20250717082932_oauth_server_tokens.ts
rephus: this is the code inside the library 

```
  async validateRedirectUri (redirectUri, client) {
    if (this.model.validateRedirectUri) {
      return this.model.validateRedirectUri(redirectUri, client);
    }

    return client.redirectUris.includes(redirectUri);
  }
  ```
  
  I'll add  a new method in the model to validate redirect uris taking wilcards into account

File: packages/backend/src/controllers/authentication/middlewares.ts (hunk 38 above)
Comment (rephus) [reply packages/backend/src/controllers/authentication/middlewares.ts]: Method removed in another PR

File: packages/backend/src/services/OAuthService/OAuthService.ts (hunk 39 above)
Comment (rephus) [reply packages/backend/src/services/OAuthService/OAuthService.ts]: Fixed in another PR

File: packages/backend/src/services/OAuthService/OAuthService.ts (hunk 16 above)
Comment (rephus) [reply packages/backend/src/services/OAuthService/OAuthService.ts]: fixed in another PR , I switched to oauthserver.authenticate

File: packages/backend/src/routers/oauthRouter.ts (hunk 47 above)
Comment (rephus) [reply packages/backend/src/routers/oauthRouter.ts:38]: Are you talking about the helmet headers on App ? Those are applied to the main express, so I imagine they are inherited by all the endpoints: 

I tested it here and it seems the headers are applied as expected. Am I missing anything else ? 


image

image

File: packages/cli/src/handlers/oauthLogin.ts (hunk 41 above)
Comment (rephus) [reply packages/cli/src/handlers/oauthLogin.ts]: Fixed in another PR

File: packages/cli/src/handlers/oauthLogin.ts (hunk 42 above)
Comment (rephus) [reply packages/cli/src/handlers/oauthLogin.ts]: Done in another PR

File: packages/cli/src/handlers/oauthLogin.ts (hunk 43 above)
Comment (rephus) [reply packages/cli/src/handlers/oauthLogin.ts]: It is possible, by updating the oauth server model method that saves/gets the access token and use our existing PAT database model instead, but I find that confusing, and it would mean mixing 2 different authentication methods, so I'm not sure that will be 100% compatible with oauth without making changes to PAT . 

For example, an access token, requires a client and has scopes, while our PAT has a rotation mechanism, so that means both are not directly compatible. 

I'd rather accept access token from oauth as a new way to authenticate people on the middleware for all endpoints

Another option would be to `generate` a new PAT at the time we create the access token, and return the PAT token as a non-oauth extra value in the oauth response. This way we don't break the oauth implementation, and we skip 1 request. 

But I don't mind keepitn this implementation, this is similar to what we do with the user/password login on the CLI , where we first login with the user/password, then, using the cookie, we generate a PAT which we save on the config file.

File: packages/backend/src/services/OAuthService/OAuthService.ts (hunk 45 above)
Comment (rephus) [reply packages/backend/src/services/OAuthService/OAuthService.ts]: method removed on another PR

File: packages/backend/src/routers/oauthRouter.ts (hunk 48 above)
Comment (rephus) [reply packages/backend/src/routers/oauthRouter.ts:245]: unfortunately not

File: packages/backend/src/services/OAuthService/OAuthService.ts [hunk 50]
```diff
@@ -0,0 +1,93 @@
...
+    ): Promise<OAuth2Server.Token> {
+        return this.oauthServer.authenticate(request, response);
+    }
+
+    public async revokeToken(token: string): Promise<boolean> {
+        // Try to revoke as access token first
+        const refreshToken = await this.oauthModel.getRefreshToken(token);
+        if (refreshToken) return this.oauthModel.revokeToken(refreshToken);
+        return false;
+    }
```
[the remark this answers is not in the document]
rephus: the model spec for node-server says this is only for refresh tokens

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,271 @@
...
+                    value: (code_challenge || '') as string,
+                },
+                {
+                    name: 'code_challenge_method',
+                    value: (code_challenge_method || '') as string,
+                },
+            ],
+        }),
+    );
+});
```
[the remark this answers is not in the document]
rephus: we apply rate limiting on the cloud

packages/backend/src/config/parseConfig.ts
```diff
@@ -1272,6 +1276,16 @@ export const parseConfig = (): LightdashConfig => {
                 loginPath: '/login/snowflake',
                 callbackPath: '/oauth/redirect/snowflake',
             },
+            oauthServer: {
+                accessTokenLifetime:
+                    getIntegerFromEnvironmentVariable(
+                        'AUTH_OAUTH_SERVER_ACCESS_TOKEN_LIFETIME',
+                    ) || 60 * 60, // 1 hour
```
stevelikesmusic, line 1283: Is it worth enforcing a max on this so customers don't set it to something like dangerous like a week or more.

packages/backend/src/controllers/authentication/middlewares.ts
stevelikesmusic: Could you add `req.account?.isAuthenticated()` to the check?

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -0,0 +1,271 @@
...
+    } else {
+        // Denied
+        res.set('Content-Type', 'text/html');
+        const redirectUri = req.body.redirect_uri;
+        if (!redirectUri) {
+            return res.status(400).json({
+                error: 'invalid_request',
+                error_description: 'Missing redirect_uri',
+            });
+        }
```
stevelikesmusic, line 127: Would it be worth moving this above the try and conditional so you don't need to check in both conditions—before `req.body.approve === 'true'`?

packages/backend/src/services/OAuthService/OAuthService.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    });
+
+    it('calls revokeToken and returns true if access token revoked', async () => {
+        mockOAuthModel.revokeToken.mockResolvedValue(true);
+        const result = await oauthService.revokeToken('token');
+        expect(result).toBe(true);
+        expect(mockOAuthModel.revokeToken).toHaveBeenCalled();
+    });
+
+    it('calls authorize', async () => {
```
stevelikesmusic: Are these tests providing much value? For the most part, they seem to just test that the mock returns the fixture it was given. Are we just verifying that a certain method is being called from the library? There's not really any business logic in the service to test. 

I think the Cypress CLI test provides the higher signal for OAuth?

packages/cli/src/handlers/login.ts
stevelikesmusic: Should OAuth be the default login mechanism for CLI and we have --password be the new option?

stevelikesmusic (approved):

packages/backend/src/config/parseConfig.ts
rephus: this is from an ENV variable. On the cloud, we set this value, although I don't think we'll change it very often.

packages/backend/src/services/OAuthService/OAuthService.test.ts
rephus: I agree, since our service doesn't do much, and we use the library heavily, these tests are now very dumb. I'll remove these tests, but keep the edge case ones.

Issue #15539, opened 2025-06-25
Update Explore Endpoints to use Account


stevelikesmusic:
For the JWT auth to work for Explores, we need to be able to execute queries without a user. Typically this user is only there for access control and analytics tracking. Update the endpoints to handle these situations. 

`/projects` and `/orgs` also need to be updated for where they get their user.

This should:

- [ ] Update Explore endpoints for explores to get the new req.account user
- [ ] Update Explore endpoints to operate without a user UUD
- [ ] Update `/projects` and `/orgs` to use the new req.account user
- [ ] Update `/user` to return the new req.account user

PR #15893, opened 2025-07-17
feat: refactor AsyncQuery and Project service to use account where needed for Embedded Explores

stevelikesmusic:
Closes: #15539

### Description:

This is the big one. 

This PR adds support for the new Account abstraction to handle both session users and embedded users consistently throughout the backend. It replaces direct `req.user` references with `req.account` in controllers and middleware, and updates service methods to accept an Account object instead of SessionUser. This enables better handling of authentication across different user types (session, JWT, anonymous) and improves security by properly checking permissions based on account type.

The changes ensure that API endpoints and query execution work correctly for both regular users and embedded users, with appropriate permission checks for each account type.

## Key changes:
- Replace `user` parameters with `account` in service methods
- Update middleware to check `req.account.isAuthenticated()` so e2e tests pass
- Refactor query history model to work with different account types
- Add JWT authentication support for async queries
- Add e2e tests for JWT authentication with async queries

### DIFF
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -1346,7 +1346,7 @@ export class AsyncQueryService extends ProjectService {
         parameters,
     }: Pick<
         ExecuteAsyncMetricQueryArgs,
-        'user' | 'metricQuery' | 'dateZoom' | 'parameters'
+        'account' | 'metricQuery' | 'dateZoom' | 'parameters'
     > & {
         warehouseSqlBuilder: WarehouseSqlBuilder;
         explore: Explore;
@@ -1639,6 +1639,7 @@ export class AsyncQueryService extends ProjectService {
                                     ', ',
                                 )}`,
                             },
+                            account,
                         );
 
                         return {
@@ -1753,9 +1754,9 @@ export class AsyncQueryService extends ProjectService {
         };
 
         const queryTags: RunQueryTags = {
+            ...this.getUserQueryTags(account),
             organization_uuid: organizationUuid,
             project_uuid: projectUuid,
-            user_uuid: account.user.id,
             explore_name: metricQuery.exploreName,
             query_context: context,
         };
@@ -1900,9 +1901,9 @@ export class AsyncQueryService extends ProjectService {
                 : metricQuery;
 
         const queryTags: RunQueryTags = {
+            ...this.getUserQueryTags(account),
             organization_uuid: savedChartOrganizationUuid,
             project_uuid: projectUuid,
-            user_uuid: account.user.id,
             chart_uuid: chartUuid,
             explore_name: savedChartTableName,
             query_context: context,
@@ -2110,9 +2111,9 @@ export class AsyncQueryService extends ProjectService {
         };
 
         const queryTags: RunQueryTags = {
+            ...this.getUserQueryTags(account),
             organization_uuid: organizationUuid,
             project_uuid: projectUuid,
-            user_uuid: account.user.id,
             chart_uuid: chartUuid,
             dashboard_uuid: dashboardUuid,
             explore_name: explore.name,
@@ -2270,9 +2271,9 @@ export class AsyncQueryService extends ProjectService {
         };
 
         const queryTags: RunQueryTags = {
+            ...this.getUserQueryTags(account),
             organization_uuid: organizationUuid,
             project_uuid: projectUuid,
-            user_uuid: account.user.id,
             explore_name: exploreName,
             query_context: context,
         };
@@ -2455,9 +2456,9 @@ export class AsyncQueryService extends ProjectService {
         );
 
         const queryTags: RunQueryTags = {
+            ...this.getUserQueryTags(account),
             organization_uuid: organizationUuid,
             project_uuid: projectUuid,
-            user_uuid: account.user.id,
             query_context: context,
         };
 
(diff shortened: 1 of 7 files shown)


stevelikesmusic:
* **#16002** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16002?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15999** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15999?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15893** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15904** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15904?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15853** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15853?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -176,11 +175,11 @@ export class AsyncQueryService extends ProjectService {
     ): Promise<{ hasAccess: boolean; userAccess: SpaceShare | undefined }> {
         const space = await this.spaceModel.getSpaceSummary(spaceUuid);
         const access = await this.spaceModel.getUserSpaceAccess(
-            user.userUuid,
+            account.user.id,
```
ZeRego, line 188: Do we need to check for `account.isRegisteredUser()` everywhere we replace `user.userUuid` or can we trust `account.user.id` ?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1607,7 +1589,10 @@ export class AsyncQueryService extends ProjectService {
                         await this.schedulerClient.scheduleTask(
                             SCHEDULER_TASKS.RUN_ASYNC_WAREHOUSE_QUERY,
                             {
-                                userUuid: user.userUuid,
+                                userUuid: account.user.id,
+                                userId: account.user.id,
```
ZeRego, line 1661: why both with the same value?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1701,30 +1687,31 @@ export class AsyncQueryService extends ProjectService {
         const queryTags: RunQueryTags = {
             organization_uuid: organizationUuid,
             project_uuid: projectUuid,
-            user_uuid: user.userUuid,
+            user_uuid: account.user.id,
```
ZeRego: fyi: in EmbedService, we apply different query tags

```
queryTags: Omit<Required<RunQueryTags>, 'user_uuid' | 'chart_uuid'> & {
            embed: 'true';
            external_id: string;
            chart_uuid?: string; // optional because query for filter autocomplete doesn't have chart uuid
        };
```

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: I assumed we use `getUserSpaceAccess` because we don't trust. If we're not registered, won't the function return not found?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: That is probably a typo 😅

stevelikesmusic: I remember now. It's due to the types. We'd need to make a lot more changes to the shared types:
https://github.com/lightdash/lightdash/blob/0a1e82e840535343b602efbdfc5966dc55850738/packages/common/src/types/scheduler.ts#L357-L362

Changing this to adapt to both user types would mean updating all scheduled tasks. Do you think the duplicate is fine for now and then we can update the types as more move over? We can make the change now, but it would mean changing many more unrelated types.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: Would something like this suffice?

```ts 
private static getUserQueryTags(account: Account) {
        if (account.isJwtUser()) {
            return {
                embed: 'true',
                external_id: account.user.id,
            };
        }
        return {
            user_uuid: account.user.id,
        };
    }
```

And then we use that wherever we'd be using query tags rather than explicitly setting `user_uuid` in each place?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
ZeRego: Already PM'ed you about prefixing external ids to avoid problems.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
ZeRego: makes sense 👍

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
ZeRego: makes sense 👍

PR #15894, opened 2025-07-17
Logo change

Joeson-Suomie:
The Lightdash icon logo was changed to a different image, but the file name remained the same.
No changes to any component.


joaoviana:
Hi @Joeson-Suomie what is the reason for a logo change? Was there a GitHub issue that required this work?

PR #15914, opened 2025-07-21
feat: add JWT header to all UI network requests if present

stevelikesmusic:
Closes: 

### Description:
This PR refactors how embed tokens are handled in the application to improve the embedding experience. The key changes include:

1. Moving the `JWT_HEADER_NAME` constant from backend to common package for better sharing
2. Storing embed tokens in in-memory in a dedicated module that can be accessed from the UI API
3. Refactoring the API client to automatically include embed tokens in requests
4. Improving project UUID handling by supporting it as a query parameter

These changes simplify the embedding flow and make it more robust, particularly when navigating between pages in an embedded dashboard.

### DIFF
--- a/packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
+++ b/packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
@@ -2,12 +2,13 @@
 // This rule is failing in CI but passes locally
 import {
     ForbiddenError,
+    JWT_HEADER_NAME,
     NotFoundError,
     ParameterError,
 } from '@lightdash/common';
 import { NextFunction, Request, Response } from 'express';
 import * as Account from '../../auth/account';
-import { decodeLightdashJwt, JWT_HEADER_NAME } from '../../auth/lightdashJwt';
+import { decodeLightdashJwt } from '../../auth/lightdashJwt';
 import { EmbedService } from '../../ee/services/EmbedService/EmbedService';
 import Logger from '../../logging/logger';
 
@@ -15,8 +16,12 @@ import Logger from '../../logging/logger';
  * We don't have the parsed routes yet, so we get the path params in a
  * cheap and dirty way. Long-term, we'll use org-level JWTs rather than project-level.
  */
-const parseProjectUuid = (path: string) => {
-    const pathParts = path.split('/');
+const parseProjectUuid = (req: Pick<Request, 'query' | 'path'>) => {
+    if (req.query.projectUuid) {
+        return req.query.projectUuid as string;
+    }
+
+    const pathParts = req.path.split('/');
     const projectParams = ['embed', 'projects'];
     const projectUuidIndex =
         pathParts.findIndex((part) => projectParams.includes(part)) + 1;
@@ -40,7 +45,9 @@ export async function jwtAuthMiddleware(
     next: NextFunction,
 ): Promise<void> {
     try {
-        if (req.account) {
+        // There are some situations where we'll already have a user and need to still create a
+        // JWT account. One example is when an admin is previewing an embed URL.
+        if (req.account?.isAuthenticated()) {
             next();
             return;
         }
@@ -49,7 +56,7 @@ export async function jwtAuthMiddleware(
         // The path will be like /api/v1/embed/{projectUuid}/dashboard
         // TODO: This is a hack to get the project UUID because JWT secrets are bound to the project rather than the organization.
         //       https://github.com/lightdash/lightdash/issues/15661
-        const projectUuid = parseProjectUuid(req.path);
+        const projectUuid = parseProjectUuid(req);
 
         if (!projectUuid) {
             next();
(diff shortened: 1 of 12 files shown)


stevelikesmusic:
* **#15914** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15914?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15914?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
We need to fix tests 

image

packages/backend/src/middlewares/jwtAuthMiddleware/jwtAuthMiddleware.ts
```diff
@@ -2,21 +2,26 @@
...
 import { EmbedService } from '../../ee/services/EmbedService/EmbedService';
 import Logger from '../../logging/logger';
 
 /**
  * We don't have the parsed routes yet, so we get the path params in a
  * cheap and dirty way. Long-term, we'll use org-level JWTs rather than project-level.
  */
-const parseProjectUuid = (path: string) => {
-    const pathParts = path.split('/');
+const parseProjectUuid = (req: Request) => {
```
rephus: perhaps 

```suggestion
const parseProjectUuid = (req: Pick<Request, 'query' | 'path'>) => {
```

rephus: in case we want to reuse it in some other places where request object is not available. but this is a small personal preference, not a blocker.

packages/frontend/src/ee/features/embed/EmbedDashboard/hooks.ts
```diff
@@ -10,21 +10,17 @@ import useDashboardFiltersForTile from '../../../../hooks/dashboard/useDashboard
...
-    projectUuid: string | undefined,
-    embedToken: string | undefined,
-) => {
+export const useEmbedDashboard = (projectUuid: string | undefined) => {
     return useQuery<Dashboard & InteractivityOptions, ApiError>({
         queryKey: ['embed-dashboard'],
-        queryFn: () => postEmbedDashboard(projectUuid!, embedToken!),
-        enabled: !!embedToken && !!projectUuid,
+        queryFn: () => postEmbedDashboard(projectUuid!),
+        enabled: !!projectUuid,
```
rephus: will these changes be compatible with existing saved embedded dashboards tokens ?

stevelikesmusic: That's a good question. My assumption was that we wouldn't even make it here because of early returns in the components using these hooks. I can add them back to be safe.

stevelikesmusic: I brought back all the checks for embedToken to enable these React Query calls.

packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
```diff
@@ -18,18 +24,38 @@ const EmbedProvider: FC<React.PropsWithChildren<Props>> = ({
...
+
+    // There is method to this madness:
+    // When we get an embedded URL, the JWT token is added as a hash to the URL location.
+    // We immediately redirect somewhere else to a URL without the hash. Consequently, if we make
+    // this initialization in a useEffect, we will not have the hash token in the URL by the time
+    // the effect runs.
+    if (!isInitialized) {
+        setToInMemoryStorage(EMBED_KEY, { projectUuid, token: embedToken });
+        setIsInitialized(true);
+    }
```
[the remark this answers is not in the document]
stevelikesmusic: There's a comment above the conditional addressing this

packages/frontend/src/utils/inMemoryStorage.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+const inMemoryStore = new Map<string, unknown>();
+
+export const getFromInMemoryStorage = <T>(key: string): T | undefined => {
+    return inMemoryStore.get(key) as T;
+};
+
+export const setToInMemoryStorage = <T>(key: string, value: T): void => {
+    inMemoryStore.set(key, value);
+};
```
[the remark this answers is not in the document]
stevelikesmusic: This seems like overkill. We currently store the token in-memory within a React Provider. Any malicious user will still have access even with encryption as they can read all our client-side code.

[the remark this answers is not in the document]
rephus: > While this avoids localStorage persistence across browser sessions, it still keeps tokens in memory during the entire session, potentially exposing them to XSS attacks.

Is this easy to exploit somehow ?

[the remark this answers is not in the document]
stevelikesmusic: From all I've read, any token stored in-memory (via React or otherwise) is vulnerable to being exploited by a malicious actor—the token kept in React in-memory context is just as available as the Map module. 

The security mitigation is in taking steps to ensure the application is protected against XSS attacks. So perhaps this is really highlighting the need to lock down our embedding security more than anything. 

@owlas any opinions? What were the security considerations in the initial embed design? 

I'm not exactly sure how to adequately assess the security risk of our current approach. My initial research suggests that our current approach is not very secure. We only have one place in the app where we use `dangerouslySetInnerHtml` and that uses an HTML sanitizer. To start, we won't be allowing any raw SQL queries for embedded users. So maybe our security risk is low in practice, but technically high given how we use the tokens.

PR #15916, opened 2025-07-21
feat: add role management to SCIM user provisioning

ZeRego:
Closes: #XXXX

### Description:
Added support for user roles in SCIM provisioning. This enhancement allows identity providers to specify and manage user roles (admin, member, etc.) when provisioning users through SCIM.

The implementation:
- Adds a Lightdash extension schema to the SCIM user object
- Enables setting roles during user creation
- Supports updating roles through SCIM PATCH operations
- Includes validation to ensure only valid organization member roles are used
- Maintains backward compatibility with existing SCIM implementations

This makes it easier for organizations to manage user permissions directly from their identity provider without requiring additional steps in Lightdash.

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.ts
@@ -5,6 +5,7 @@ import {
     ForbiddenError,
     getErrorMessage,
     GroupWithMembers,
+    isOrganizationMemberRole,
     isValidEmailAddress,
     LightdashUser,
     NotFoundError,
@@ -137,7 +138,8 @@ export class ScimService extends BaseService {
         const updatedAt =
             'updatedAt' in user ? user.updatedAt : user.userUpdatedAt;
 
-        return {
+        // Create the base SCIM user
+        const scimUser: ScimUser = {
             schemas: [ScimSchemaType.USER],
             id: user.userUuid,
             userName: user.email || '',
@@ -162,6 +164,16 @@ export class ScimService extends BaseService {
                 ).href,
             },
         };
+
+        // Add the Lightdash extension schema if the user has a role
+        if (user.role) {
+            scimUser.schemas.push(ScimSchemaType.LIGHTDASH_USER_EXTENSION);
+            scimUser[ScimSchemaType.LIGHTDASH_USER_EXTENSION] = {
+                role: user.role,
+            };
+        }
+
+        return scimUser;
     }
 
     static getScimUserEmail(user: Pick<ScimUser, 'userName'>): string {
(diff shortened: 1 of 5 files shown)


ZeRego:
* **#15916** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/15916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joaoviana:
**Creating new **admin****

Screenshot 2025-07-21 at 15 48 00
Screenshot 2025-07-21 at 15 47 44


**Trying to create new user with an invalid role**

Screenshot 2025-07-21 at 15 48 29


**Editing existing user**



Screenshot 2025-07-21 at 15 54 21
Screenshot 2025-07-21 at 15 54 16

**Getting existing user to verify role change**
```json
{"schemas":["urn:ietf:params:scim:schemas:core:2.0:User","urn:lightdash:params:scim:schemas:extension:2.0:User"],"id":"d5048ee7-3012-4ade-bb1e-5ac2169e3454","userName":"newuser@example.com","name":{"givenName":"New","familyName":"User"},"active":true,"emails":[{"value":"newuser@example.com","primary":true}],"meta":{"resourceType":"User","created":"2025-07-21T14:47:30.817Z","lastModified":"2025-07-21T14:54:05.033Z","location":"http://localhost:3000/api/v1/scim/v2/Users/d5048ee7-3012-4ade-bb1e-5ac2169e3454"},"urn:lightdash:params:scim:schemas:extension:2.0:User":{"role":"editor"}}
```

**Creating user with role empty string**

> [1] If the role property is missing when creating a user, the user will be created with the role `member`.

Screenshot 2025-07-21 at 15 47 44


**Direct path (PATCH) of role of existing user when the role is empty - goes to member**

Screenshot 2025-07-21 at 16 10 28

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -336,12 +348,31 @@ export class ScimService extends BaseService {
...
                 user.active,
             );
+            // Extract role from extension schema if available
+            const extensionData = user[ScimSchemaType.LIGHTDASH_USER_EXTENSION];
+            let role = OrganizationMemberRole.MEMBER; // Default role
+
+            // If a role is provided in the extension schema, validate and use it
+            if (extensionData?.role) {
+                // Validate that the role is a valid OrganizationMemberRole
+                if (!isOrganizationMemberRole(extensionData.role)) {
```
joaoviana, line 358: tested ✅

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -418,18 +449,47 @@ export class ScimService extends BaseService {
                     isActive: user.active ?? dbUser.isActive,
                 },
             );
+
+            // Update user's organization role if provided in the extension schema
```
joaoviana, line 453: wonder if we should consider the edge case where it could be updating the last admin user as non-admin, leaving no admins in the org 🤔

ZeRego: I thought about it and didn't want to add that restriction to SCIM provisioning. That check was put in place so users using the UI wouldn't be locked out of their org. 
In this case, I expect them to notice and use SCIM to assign a new admin.
In the case of the token expiring, they would reach out to support.

Issue #15903, opened 2025-07-18
Inconsistent validator behavior with explores + virtual views


joaoviana:
A customer is experiencing long-running compile jobs in Lightdash. Investigation revealed:

- The worker pod is being shut down by Kubernetes after fetching table metadata for 502 tables, causing the compile job to wait indefinitely.
- There is a difference in behavior between running `lightdash validate` in the CLI and using the UI. Specifically, fields from virtual views are not pulled in when running `lightdash validate` locally, leading to warnings in the CLI but not in the UI.

**Steps to reproduce:**
1. Run a compile job in Lightdash with a project containing a large number of tables (e.g., 500+).
2. Observe that the worker pod may be killed by Kubernetes after fetching table metadata, causing the compile job to hang.
3. Run `lightdash validate` locally and in the UI; note the difference in warnings/errors related to virtual views.

**Expected behavior:**
- Validator behavior should be consistent between CLI and UI, especially regarding virtual views.

PR #15923, opened 2025-07-21
fix: get virtual views when cli-validate project

joaoviana:
Closes: #15903

### Description:

Get compiled explores + virtual views with new method so that these can be merged and checked when validating with the CLI

How to test:
- Go to SQL runner and create a virtual view
- Validate project with the CLI (see cli/README)
- See that `compiledExplores` has no virtual views, so we add them

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.ts
@@ -607,9 +607,7 @@ export class ValidationService extends BaseService {
             const virtualViews =
                 await this.projectModel.findVirtualViewsFromCache(projectUuid);
 
-            explores = Array.from(
-                compiledExplores.concat(Object.values(virtualViews)),
-            );
+            explores = compiledExplores.concat(Object.values(virtualViews));
             this.logger.debug(
                 `Merged ${compiledExplores.length} compiled explores with ${
                     Object.values(virtualViews).length
--- a/packages/backend/src/services/ValidationService/ValidationService.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.ts
@@ -601,14 +601,25 @@ export class ValidationService extends BaseService {
             }`,
         );
 
-        const explores =
-            compiledExplores !== undefined
-                ? compiledExplores
-                : Object.values(
-                      await this.projectModel.findExploresFromCache(
-                          projectUuid,
-                      ),
-                  );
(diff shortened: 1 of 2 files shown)


packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -601,14 +601,25 @@ export class ValidationService extends BaseService {
...
+            );
+            this.logger.debug(
+                `Merged ${compiledExplores.length} compiled explores with ${
+                    Object.values(virtualViews).length
+                } virtual views for validation`,
+            );
+        } else {
+            explores = Object.values(
+                await this.projectModel.findExploresFromCache(projectUuid),
+            );
```
almeidabbm, line 619: In here we don't need to concat virtual views?

packages/backend/src/services/ValidationService/ValidationService.ts
```diff
@@ -601,14 +601,25 @@ export class ValidationService extends BaseService {
...
-                  );
+        let explores: (Explore | ExploreError)[];
+        if (compiledExplores !== undefined) {
+            // For CLI validation, when compiled Explores are provided, merge them with virtual views from cache
+            const virtualViews =
+                await this.projectModel.findVirtualViewsFromCache(projectUuid);
+
+            explores = Array.from(
+                compiledExplores.concat(Object.values(virtualViews)),
+            );
```
almeidabbm: No need to create a new array when calling `concat`

packages/backend/src/services/ValidationService/ValidationService.ts
joaoviana: No because `findExploresFromCache` gets virtual views, as they are _explores_

packages/backend/src/services/ValidationService/ValidationService.ts
joaoviana: 🤦

joaoviana: after the change, all good: 

Image

joaoviana: https://github.com/lightdash/lightdash/pull/15923/commits/a9beca052231794ed3046c8b616f3dce5d750bda

almeidabbm (approved):
LGTM 🚀 

Image

PR #15706, opened 2025-07-04
feat: add group access to ai agents

joaoviana:
Relates to: #15385

### Description:
This PR adds group-based access control for AI Agents, allowing administrators to restrict agent access to specific user groups. 

Key changes:
- Added a `groupAccess` field to AI Agents that stores which groups can access an agent
- Implemented permission checks throughout the AI Agent service to verify user access
- Added UI components in the agent creation/edit form to select which groups can access an agent
- Improved error handling for permission-related errors in the frontend
- Added database schema and model methods to support group-based access control

When no groups are specified for an agent, it remains accessible to all users(non-viewers) in the organization. When groups are specified, only users in those groups (or users with admin/developer permissions) can access the agent.

Screenshot 2025-07-04 at 13 53 51

Issue #15385, opened 2025-06-17
I would like to control who gets to interact with the AI Agents


joaoviana:
### Description

Acceptance criteria
- [ ] As an Admin, I can control who gets to interact with my AI Agent
- [ ] I can set a role for interaction with AI Agent (e.g. editor and above, developers and above,etc)

> [!NOTE]
> Currently, explores can be filtered with "tag"s but there's no way to have an AI agent that is _unfiltered_/no tags but only accessible to a specific set of users  

### What problem does this solve?

Some AI Agents need to have access to a lot more information, but that information needs to be restricted

## Technical considerations

Current AI Permissions:
- `interactive_viewer` and above can view AI agents and create threads
- `developer` + `admin` can manage AI agents
- anyone that can interact with the AI Agents can see 'Ask AI'

### Suggested role controls
1. Role selector (e.g. `editor` and above)
2. User selector (stretch goal?)
3. Group selector (stretch goal?)

### Global Ask AI button visibility

Display Ask AI button if current user has at least one AI Agent that they have access to on their current project.

### AI-Agent-level permission changes

When creating/editing an AI Agent, admins/developers can: 
1. Set visibility to all 
or
1. Override permissions to a specific minimum role 
2. to a specific user list
3. to a specific group 

To implement:
1. On each endpoint, check if current user can view or interact with said agent. 

### Questions

1. Should there be a global setting in "organization settings" to set a default permissions set for everything AI Agents?
2. Should there be a way for admins to hide the "Ask AI" button for their org?

PR #15937, opened 2025-07-22
feat: add group access to ai agents 

joaoviana:
Relates to: #15385 
Re-opens: #15706 after being reverted in https://github.com/lightdash/lightdash/pull/15849

### Description:

This PR adds group-based access control for AI Agents, allowing administrators to restrict agent access to specific user groups. 

Key changes:
- Added a `groupAccess` field to AI Agents that stores which groups can access an agent
- Implemented permission checks throughout the AI Agent service to verify user access
- Added UI components in the agent creation/edit form to select which groups can access an agent
- Improved error handling for permission-related errors in the frontend
- Added database schema and model methods to support group-based access control

When no groups are specified for an agent, it remains accessible to all users(non-viewers) in the organization. When groups are specified, only users in those groups (or users with admin/developer permissions) can access the agent.

Screenshot 2025-07-04 at 13 53 51


packages/frontend/src/components/NavBar/AiAgentsButton.tsx
```diff
@@ -24,7 +24,10 @@ export const AiAgentsButton = () => {
     });
     const aiCopilotFlagQuery = useFeatureFlag(CommercialFeatureFlags.AiCopilot);
     const aiAgentFlagQuery = useFeatureFlag(CommercialFeatureFlags.AiAgent);
-    const agents = useProjectAiAgents(projectUuid);
+    const agents = useProjectAiAgents({
+        projectUuid,
+        redirectOnUnauthorized: false,
+    });
```
joaoviana: this was redirecting everyone that had the feature off

notgiorgi (approved):

Issue #15932, opened 2025-07-22
YAML Filters - Negative Numbers Not Recognised Correctly


jesshitchcock:
### Summary
When using negative numbers in YAML filter conditions, the SQL generation is incorrect. The system fails to properly parse negative number comparisons and generates malformed SQL syntax.

### Expected Behaviour
Filter condition amount: ">=-1" should generate valid SQL:
```SUM(CASE WHEN ((`deals`.amount) >= (-1)) THEN (`deals`.seats) ELSE NULL END)```

### Actual Behavior
Filter condition amount: ">=-1" generates invalid SQL syntax:
```SUM(CASE WHEN ((`deals`.amount) IN (>=-1)) THEN (`deals`.seats) ELSE NULL END)```

### Steps to Reproduce the Bug or Issue

1. In a model sql yaml file, add a filter to a ">=-1" filter to a metric. For example: 
``` yaml
version: 2
models:
  - name: deals
    columns:
      - name: deal_id
        description: "l"
        meta:
          dimension:
            type: string
      - name: account_id
        description: ""
        meta:
          dimension:
            type: number  
          
      - name: stage
        description: "The deal stage (New, Qualified, PoC, Negotiation, Won, Lost)"
        meta:
          dimension:
            type: string
     
      - name: plan
        description: "The plan the prospect is interested in (Basic or Professional)"
        meta:
          dimension:
            type: string
            colors:
              'Basic': "#C0C0C0"
              'Professional': "#FFD700"
      - name: seats
        description: "The number of seats they are purchasing"
        meta:
          dimension:
            type: number
          metrics:
            total_seats:
              type: sum
              description: The sum of seats for selected deals
              filters: 
                - amount: ">=-1"  #<--filter applied 
```
2. Create a preview project 
3. In the Lightdash explore, select the filtered metric. In this case `total seats` and run the SQL 
4. Change the filter to: 
``` yaml
              filters: 
                - amount: ">=1"  #<--filter applied 
```
5. Recompile or wait for the project to automatically recompile and then reselect the metric. 

### version

v0.1825.0

### Cloud or self-hosting

cloud

PR #15942, opened 2025-07-22
fix: correct compile for greater or less than filters with negative numbers

jesshitchcock:
Closes: [#15932](https://github.com/lightdash/lightdash/issues/15932)

### Description:
Corrects issue where negative numbers used with greater than, less than, greater than or equal to and less than or equal to is compiled as: `SUM(CASE WHEN ((deals.amount) IN (>=-1)) THEN (deals.seats) ELSE NULL END)` instead of `SUM(CASE WHEN ((deals.amount) >= (-1)) THEN (deals.seats) ELSE NULL END)` 

Also added a test for this use case. 

Note I have not tested locally as I haven't had tome to set this up yet. Would appreciate a thorough review even though the code change is small.

### DIFF
--- a/packages/common/src/types/filterGrammar.test.ts
+++ b/packages/common/src/types/filterGrammar.test.ts
@@ -144,15 +144,27 @@ describe('Parse grammar', () => {
         // Positive floats
         expect(parser.parse('> 3.14')).toEqual({ type: '>', values: [3.14] });
         expect(parser.parse('< 0.5')).toEqual({ type: '<', values: [0.5] });
-        expect(parser.parse('>= 100.001')).toEqual({ type: '>=', values: [100.001] });
-        expect(parser.parse('<= 2.718')).toEqual({ type: '<=', values: [2.718] });
-        
-        // Negative floats  
+        expect(parser.parse('>= 100.001')).toEqual({
+            type: '>=',
+            values: [100.001],
+        });
+        expect(parser.parse('<= 2.718')).toEqual({
+            type: '<=',
+            values: [2.718],
+        });
+
+        // Negative floats
         expect(parser.parse('< -3.14')).toEqual({ type: '<', values: [-3.14] });
         expect(parser.parse('> -0.5')).toEqual({ type: '>', values: [-0.5] });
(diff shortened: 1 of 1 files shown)


packages/common/src/types/filterGrammar.ts
```diff
@@ -63,7 +63,7 @@ FLOAT
   = INTEGER '.'? INTEGER?
 
 INTEGER
-  = [0-9]+
+  = '-'?[0-9]+
```
tatianainama: I would move this to the `NUMBER` definition like:
```diff
NUMBER
++  = '-'? FLOAT ([Ee] [+-]? INTEGER)?
    { return Number(text()) }

INTEGER
--  = [0-9]+
```

`INTEGER` is used as part of `FLOAT`, so adding this negative here allows expressions like `-10.-5`
(We could add another test case for that!)

jesshitchcock: Thanks!

packages/common/src/types/filterGrammar.test.ts
```diff
@@ -134,6 +140,39 @@ describe('Parse grammar', () => {
...
+        expect(() => parser.parse('< -3.-14')).toThrow();
+    });
+
+    it('Negative integers work correctly', async () => {
+        // Test that negative integers still work after moving negative sign to NUMBER
+        expect(parser.parse('> -42')).toEqual({ type: '>', values: [-42] });
+        expect(parser.parse('< -1')).toEqual({ type: '<', values: [-1] });
+        expect(parser.parse('>= -100')).toEqual({ type: '>=', values: [-100] });
+        expect(parser.parse('<= -999')).toEqual({ type: '<=', values: [-999] });
+    });
```
tatianainama: I thinks these are already covered by the ones in L128!

jesshitchcock: Removed

jesshitchcock:
Tested it is working in development environment 
image

ZeRego:
@all-contributors add @jesshitchcock for code

Issue #15829, opened 2025-07-11
Add basic evaluation framework for AI Agents


tatianainama:
### Description

Implement a basic evaluation framework to test AI Agents performance. Start with some automated tests that validates agent responses against expected outputs (models, defined instructions, chart generation)

This can be done with unit and e2e testing for now

### What problem does this solve?

At the moment we can't reliably detect when model changes cause performance regressions or when the agent goes cuckoo. This makes development subjective and risky, so we need tools to know if changes improve or degrade the agent's ability to choose correct metrics, dimensions, and visualizations.

PR #15989, opened 2025-07-24
test: score find explores response with llm-as-a-judge

joaoviana:
Relates to: #15829 

### Description:

`llmAsAJudge.ts` utility uses `autoevals` prompt for `factuality` scoring and `JSONDiff` as another scorer 

image

### DIFF
--- a/packages/backend/src/ee/services/ai/agents/tests/agent.integration.test.ts
+++ b/packages/backend/src/ee/services/ai/agents/tests/agent.integration.test.ts
@@ -7,6 +7,8 @@ import {
 } from '../../../../../vitest.setup.integration';
 import { DbAiAgentToolCall } from '../../../../database/entities/ai';
 import { AiAgentService } from '../../../AiAgentService';
+import { getOpenaiGptmodel } from '../../models/openai-gpt';
+import { llmAsAJudge } from './utils/llmAsAJudge';
 
 // Skip if no OpenAI API key
 const hasApiKey = !!process.env.OPENAI_API_KEY;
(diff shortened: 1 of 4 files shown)


packages/backend/src/ee/services/ai/agents/tests/utils/llmAsAJudge.ts
```diff
@@ -0,0 +1,60 @@
...
+
+                Assistant Response: "${response}"
+
+                Evaluate the response on three dimensions:
+                1. Relevance - Does it answer what was asked?
+                2. Accuracy - Is the information correct and trustworthy?
+                3. Completeness - Does it cover all aspects of the query?
+
+                Be strict but fair in your evaluation.`,
+    });
```
IrakliJani: I don't know what the best practice is around having 1 system prompt vs separating system prompt and user message. might be worth experimenting

packages/backend/src/ee/services/ai/agents/tests/agent.integration.test.ts
```diff
@@ -79,8 +82,20 @@ describeOrSkip('agent integration tests', () => {
             expect(toolCalls[0].tool_name).toBe('findExplores');
 
             expect(typeof response).toBe('string');
-            // TODO: check if response contains the expected data
-            // TODO: add evaluation
+
+            const model = getOpenaiGptmodel({
+                apiKey: process.env.OPENAI_API_KEY!,
+                modelName: 'gpt-4o-mini',
```
IrakliJani: why gpt 4o mini?

joaoviana: only because it's quicker - will go for gpt-4.1 in favour of consistency

packages/backend/src/ee/services/ai/agents/tests/utils/llmAsAJudge.ts
joaoviana: I see - I'm following this yml template now: https://github.com/braintrustdata/autoevals/blob/5aa20a0a9eb8fc9e07e9e5722ebf71c68d082f32/templates/factuality.yaml

and also use JSON diff for the score diff algo 

wdyt of the new changes?

PR #16016, opened 2025-07-25
feat: apply req.account to apiKey and serviceAccounts

stevelikesmusic:
Closes: #ISSUE_NUMBER

### Description:
Adds support for API Key and Service Account authentication types in the account system.  This should give us broader coverage of req.account and prevent unexpected breaks in endpoints that now expect a req.account to be there. Previously, we only were creating req.account for cookies and JWT auth. 

Context: 
- https://github.com/lightdash/lightdash/pull/16007
- [Slack](https://lightdash.slack.com/archives/C02GQKJK84Q/p1753376104001839)

### AI Breakdown
- Extends the `Account` type to include `ApiKeyAccount` and `ServiceAcctAccount`
- Adds helper methods to identify service accounts and personal access tokens
- Implements `fromApiKey` and `fromServiceAccount` functions to create accounts from these auth types
- Refactors the account creation logic to handle all account types consistently
- Adds a shared helper function `extractOrganizationFromUser` to reduce code duplication
- Updates middleware to properly set account information for API key authentication
- Enhances type definitions in the auth system

### DIFF
--- a/packages/backend/src/auth/account/account.ts
+++ b/packages/backend/src/auth/account/account.ts
@@ -6,12 +6,14 @@ import {
     AccountHelpers,
     AccountOrganization,
     AnonymousAccount,
+    ApiKeyAccount,
     applyEmbeddedAbility,
     CreateEmbedJwt,
     Embed,
     ForbiddenError,
     MemberAbility,
     Organization,
+    ServiceAcctAccount,
     SessionAccount,
     SessionUser,
     UserAccessControls,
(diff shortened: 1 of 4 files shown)


stevelikesmusic:
* **#16016** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16016?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16016?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stevelikesmusic:
ugh graphite....

packages/backend/src/ee/authentication/middlewares.ts
```diff
@@ -150,6 +151,7 @@ export const authenticateServiceAccount: RequestHandler = async (
             createdAt: serviceAccount.createdAt,
             updatedAt: serviceAccount.createdAt,
         };
+        req.account = fromServiceAccount(req.user!, token);
```
rephus, line 157: ideally, we should generate the `userless` account from the service account directly, not the hacky made user.

packages/backend/src/auth/account/account.ts
```diff
@@ -100,21 +124,51 @@ export const fromJwt = ({
     });
 };
 
+export const fromServiceAccount = (
+    sessionUser: SessionUser,
```
rephus, line 131: we should get the account from an userless service account. 

I'm ok with this for now, but once this account is in place we should consider removing that tech debt from service accounts. We should add a comment in here about  it.

stevelikesmusic: Totally agree! Added a comment in both places with a link to the ticket you created

PR #16047, opened 2025-07-28
test: add relevancy test for time series

joaoviana:
Closes: N/A

### Description:

Adds
- `mastra`
- `@mastra/core`
- `@mastra/evals`

And creates a relevancy test for time-series prompt - the fields retrieved should be relevant to the prompt


packages/backend/src/vitest.setup.integration.ts
```diff
@@ -152,6 +154,30 @@ export const setupIntegrationTest =
             abilityRules: [],
         };
 
+        const testUserSessionAccount: SessionAccount = {
```
joaoviana, line 157: `findExplore/getExplore` expects new type `Account` (which can be `SessionAccount`)

packages/backend/src/ee/services/ai/agents/tests/agent.integration.test.ts
```diff
@@ -256,4 +247,74 @@ describeOrSkip('agent integration tests', () => {
...
+
+            const toolCalls = await context
+                .db<DbAiAgentToolCall>('ai_agent_tool_call')
+                .leftJoin(
+                    'ai_agent_tool_result',
+                    'ai_agent_tool_call.tool_call_id',
+                    'ai_agent_tool_result.tool_call_id',
+                )
+                .where('ai_agent_tool_call.ai_prompt_uuid', prompt!.promptUuid)
+                .select('*');
```
IrakliJani, line 271: there is a    ` async getToolCallsAndResultsForPrompt(promptUuid: string) {` method that returns both tool calls and tool results with inner join. might be helpful to use here

IrakliJani (approved):
there's an error in api tests. not sure how it's relevant to the changes made in this PR

joaoviana:
> there's an error in api tests. not sure how it's relevant to the changes made in this PR

no, this must have been from adding a new yml file/new explore

packages/backend/src/ee/services/ai/agents/tests/agent.integration.test.ts
joaoviana: I want to use that method when we get to remove the feature flag for the _experimental_ find fields tool you've built 🙌 so will skip this suggestion for now 🙏

Issue #16021, opened 2025-07-25
Save dashboard parameters


ZeRego:
### Description

Similar to saving parameters in the chart. We should be able to save parameters in the dashboard.

### What problem does this solve?

slack thread: https://lightdash.slack.com/archives/C08D77QLD47/p1753437448357069?thread_ts=1753203421.142489&cid=C08D77QLD47

> I know you can set a default in code, but with the nature of the parameter we are using, it would be nice for the dashboard to remember the previous default value we set

PR #16066, opened 2025-07-28
chore: add parameters to dashboard creation

magnew:
Relates to: #16021 

### Description:

Adds the ability to save parameter values with dashboards. This adds a dashboard parameters type and a column to the `dashboard_views` table to follow the pattern we have with filters.

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -8,6 +8,7 @@ import {
     DashboardDAO,
     DashboardLoomTile,
     DashboardMarkdownTile,
+    DashboardParameterValue,
     DashboardSqlChartTile,
     DashboardTab,
     DashboardTileTypes,
@@ -28,6 +29,7 @@ import {
     sanitizeHtml,
     type DashboardBasicDetailsWithTileTypes,
     type DashboardFilters,
+    type DashboardParameters,
 } from '@lightdash/common';
 import { Knex } from 'knex';
 import { v4 as uuidv4 } from 'uuid';
@@ -145,6 +147,7 @@ export class DashboardModel {
                 metrics: [],
                 tableCalculations: [],
             },
+            parameters: version.parameters || null,
         });
 
         if (version.tabs.length > 0) {
@@ -299,6 +302,7 @@ export class DashboardModel {
                                 : undefined,
                         })) ?? [],
                 },
+                parameters: version.parameters || null,
             })
             .where({ dashboard_version_id: versionId.dashboard_version_id });
     }
@@ -499,6 +503,7 @@ export class DashboardModel {
             dashboardUuid: string;
             name: string;
             filters: DashboardFilters;
+            parameters: DashboardParameters | null;
             chartUuids: string[];
         }>
     > {
@@ -541,6 +546,7 @@ export class DashboardModel {
                     dashboardUuid: `${cteName}.dashboard_uuid`,
                     name: `${cteName}.name`,
                     filters: `${DashboardViewsTableName}.filters`,
+                    parameters: `${DashboardViewsTableName}.parameters`,
                     chartUuids: this.database.raw(
                         "COALESCE(ARRAY_AGG(DISTINCT saved_queries.saved_query_uuid) FILTER (WHERE saved_queries.saved_query_uuid IS NOT NULL), '{}')",
                     ),
@@ -561,7 +567,7 @@ export class DashboardModel {
                     `${DashboardTileChartTableName}.saved_chart_id`,
                     `${SavedChartsTableName}.saved_query_id`,
                 )
-                .groupBy(1, 2, 3)
+                .groupBy(1, 2, 3, 4)
         );
     }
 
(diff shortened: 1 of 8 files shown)


magnew:
* **#16079** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16079?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16077** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16077?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16066** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/types/parameters.ts
```diff
@@ -1 +1,13 @@
 export type ParametersValuesMap = Record<string, string | string[]>;
+
+export type DashboardParameterValue = {
+    id: string;
+    parameterName: string;
+    value: string | string[];
+    tileTargets: string[];
```
magnew: I put this tileTargets here even though it's not used. I assume it will be. Should we keep it or add it later?

I wasn't totally sure about the type -- unlike filters this could probably just be an array of tile ids. I couldn't think of a great case for a more complex object, but maybe it exists.

packages/common/src/types/parameters.ts
```diff
@@ -1 +1,13 @@
...
+};
+
+export type DashboardParameters = {
+    // This array is wrapped in a top-level object to avoid a serialization issue
+    values: DashboardParameterValue[];
```
magnew: Like the comment says, I wrapped this in an object. I couldnt get knex to save an array as a top-level object in a JSONB column. I can take another approach it this seem janky.

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -541,6 +565,7 @@ export class DashboardModel {
                     dashboardUuid: `${cteName}.dashboard_uuid`,
                     name: `${cteName}.name`,
                     filters: `${DashboardViewsTableName}.filters`,
+                    parameters: `${DashboardViewsTableName}.parameters`,
                     chartUuids: this.database.raw(
```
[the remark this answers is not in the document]
ZeRego: @magnew, important fix.

packages/common/src/types/parameters.ts
ZeRego: Could be a `Record<string, string>` with tile uuid and the value for that tile. Or make it `Record<string, {value: string}>` so it's extendable to more configs.
I would say add it later once we have specific requirements.

packages/common/src/types/parameters.ts
ZeRego: You can see the `customRange` jsonb array in SavedChartModel.

```JSON.stringify(customDimension.customRange)```

Is also fine to leave it like this 👍

Comment (ZeRego) [reply packages/common/src/types/parameters.ts]: Do we want to have an array or a record? unlike filters that can have multiple filters for the same dimension, parameters have to be unique. 🤔 
Also, do we need an ID?  

```
        const testParameters = {
            values: [
                {
                    id: 'test-param-1',
                    parameterName: 'time_zoom',
                    value: 'weekly',
                    tileTargets: [],
                },
                {
                    id: 'test-param-2',
                    parameterName: 'region',
                    value: ['US', 'EU'],
                    tileTargets: [],
                },
            ],
        };
```
vs
```
        const testParameters = {
            time_zoom: {
                parameterName: 'time_zoom',
                value: 'weekly',
                tileTargets: [],
            },
            region: {
                parameterName: 'region',
                value: ['US', 'EU'],
                tileTargets: [],
            },
        };
```

magnew: Discussed in Slack

Marshall
> Do we want to have an array or a record? unlike filters that can have multiple filters for the same dimension, parameters have to be unique. :thinking_face:
> Also, do we need an ID?

From your review. Slack might be faster to chat. I couldn't really decide and I just went one way. For array vs. record: I went with arrays since we talked at some point about having different values for different tiles in a dashboard. If we do it with a record, it's going to be really hard to ever do that. Its possible the UI for that would be crazy anyway and we want to shut it down. If it doensnt matter, the array may let us be more flexible later. What do you think?
Yea, don't need ID. I was following filters, but it's not really even used there.

Jose Rego
If we do it with a record, it's going to be really hard to ever do that.
That's why in my mind the tileTargets is a record of tile uuid + value.

Marshall
Yea, I like that. Still keep a value at the top level?

Jose Rego
yes

packages/backend/src/models/DashboardModel/DashboardModel.ts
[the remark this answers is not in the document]
magnew: Does this seem right now, @ZeRego? I can see how I missed it. List of numbers if very opaque 😬

ZeRego (approved):

Issue #10080, opened 2024-05-15
I want to be able to not re-run a query when I'm adding filters/sorting in Lightdash


TuringLovesDeathMetal:
### Description

Today, every time I add a new filter or a new sort to my results, a query gets rerun in Lightdash. 

This is fine if I'm working with a tiny table, but if my table is say 1TB, then this is _**really**_ time-consuming and takes a really long time for me to do anything. 

I don't know the right answer here. It could be a combination of backend changes which result in us caching data smartly so that we can update ordering adn filtering quite quickly as well as UX/UI changes where there's a way to stop queries from running automatically (and you have to manually click `run query` each time instead). 

It's also an odd pattern today, so it's hard to know when to click `run query` vs. not.

- adding a dimension/metric to my table **doesn't** run a new query
- adding a table calculation runs a query
- adding a new filter **doesn't** run a query
- adding sorting runs a query (with each new change I make to sorting).

TuringLovesDeathMetal:
Potential solution: Adding a button near `run query` which you can set to `auto refresh` or `not auto refresh`. 

By default, it's set to `auto refresh`. We'd need to change the current behaviour so that adding metrics/dimensions + filters to your query would automatically re-run the query (so it's more consistent)

You can change it to `no auto refresh` which would make it so that _**nothing**_ would re-run a query unless you clicked `run query`. 

Ideally, I think that sorting would be something we could do using cache-ing 👀 (that's the one thing that feels weird that it doesn't auto-update with this change).

TuringLovesDeathMetal:
^This could actually be hidden in the drop-down of `run query`!

PR #16086, opened 2025-07-29
feat: add auto run results setting in explore

almeidabbm:
Closes: #10080

### Description:
Added an auto-fetch toggle button that allows users to enable/disable automatic query execution when making changes to the explorer. This gives users more control over when queries are run, especially useful for complex or resource-intensive queries.

The auto-fetch button appears in the Explorer header, with a tooltip indicating the current status. When enabled (default), queries run automatically as before. When disabled, users must manually click the refresh button to execute queries.

Auto fetch results setting is saved in the local storage.
Complete removed `shouldFetchResults` boolean from `exploreProvider`

image

### DIFF
--- a/packages/frontend/src/components/AutoFetchResultsButton.tsx
+++ b/packages/frontend/src/components/AutoFetchResultsButton.tsx
@@ -1,0 +1,37 @@
+import { ActionIcon, Tooltip, type MantineSize } from '@mantine/core';
+import { IconRefresh, IconRefreshOff } from '@tabler/icons-react';
+import { memo, type FC } from 'react';
+import useExplorerContext from '../providers/Explorer/useExplorerContext';
+import MantineIcon from './common/MantineIcon';
+
+export const AutoFetchResultsButton: FC<{ size?: MantineSize }> = memo(
+    ({ size }) => {
+        const autoFetchEnabled = useExplorerContext(
+            (context) => context.state.autoFetchEnabled,
+        );
+        const setAutoFetchEnabled = useExplorerContext(
+            (context) => context.actions.setAutoFetchEnabled,
+        );
+
+        return (
+            <Tooltip
+                label={`Auto-fetch results ${
+                    autoFetchEnabled ? 'enabled' : 'disabled'
+                }`}
+                position="bottom"
+                withArrow
+                withinPortal
+            >
+                <ActionIcon
+                    size={size}
+                    variant="default"
+                    onClick={() => setAutoFetchEnabled(!autoFetchEnabled)}
+                >
+                    <MantineIcon
+                        icon={autoFetchEnabled ? IconRefresh : IconRefreshOff}
+                    />
+                </ActionIcon>
+            </Tooltip>
+        );
+    },
+);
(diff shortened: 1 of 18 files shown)


almeidabbm:
* **#16086** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16086?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16086?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/AutoFetchResultsButton.tsx
```diff
@@ -0,0 +1,35 @@
+import { ActionIcon, Tooltip, type MantineSize } from '@mantine/core';
+import { useLocalStorage } from '@mantine/hooks';
+import { IconRefresh, IconRefreshOff } from '@tabler/icons-react';
+import { memo, type FC } from 'react';
+import MantineIcon from './common/MantineIcon';
+
+export const AutoFetchResultsButton: FC<{ size?: MantineSize }> = memo(
+    ({ size }) => {
+        const [autoFetchEnabled, setAutoFetchEnabled] = useLocalStorage({
```
magnew, line 9: We really need redux 😂

packages/frontend/src/components/common/Filters/index.tsx
```diff
@@ -39,7 +39,7 @@ import useFiltersContext from './useFiltersContext';
 
 type Props = {
     filters: Filters;
-    setFilters: (value: Filters, shouldFetchResults: boolean) => void;
```
magnew, line 42: 🔥

packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
```diff
@@ -862,6 +854,11 @@ const ExplorerProvider: FC<
     dateZoomGranularity,
     projectUuid: propProjectUuid,
 }) => {
+    const [autoFetchEnabled] = useLocalStorage({
+        key: 'lightdash-explorer-auto-fetch-enabled',
+        defaultValue: true,
+    });
```
magnew, line 860: Why use local storage instead of making it part of the explore provider state? Do you need it in places where there is no provider?

magnew (commented):
This is great! 

For some reason it works fro me locally, but not on Okteto. Does Okteto work? 😂 

I put a question about local storage -- if possible I think it would be better to store the boolean in the provider instead of local storage. But if there is a reason for it to be there that I am missing, ok. 

This doesn't have to block checking, but, in my opinion, this should be a toggle or this kind of segmented control. I find the button confusing. 
Screenshot 2025-07-31 at 10 20 02

packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
almeidabbm: Yep yep, the reason for local storage is so that this value persists across user sessions. I believe it wouldn't be a great experience if a user likes to have it as autofetch: off and needs to click it every time 🤔

almeidabbm: btw the preview environment wasn't deployed, I have retriggered 😅 that's probably why it didn't work for you there?

magnew: > yep yep, the reason for local storage is so that this value persists across user sessions. I believe it wouldn't be a great experience if a user likes to have it as autofetch: off and needs to click it every time

Ok, yea, cool, agree! That makes sense.

magnew (approved):

Issue #16014, opened 2025-07-25
Build MCP server for Lightdash


rephus:
## Description
Add basic MCP (Model Context Protocol) support to Lightdash as a remote server with OAuth authentication.

## Acceptance Criteria

**AC1: Remote MCP Server with OAuth**
- Implement MCP server with HTTP transport
- Integrate with existing Express middleware
- OAuth authentication for secure access
- Must pass all tests using the MCP Inspector tool (https://github.com/modelcontextprotocol/inspector)

**AC2: Version Tool**
- Implement a single MCP tool: `get_lightdash_version`
- Returns the current version of Lightdash being used

## Implementation Notes
- Use existing Lightdash OAuth infrastructure
- Integrate as Express middleware in the existing application
- Follow MCP specification for HTTP transport
- Keep scope minimal - just authentication and version tool
- Future iterations can add query execution, dashboard management, etc.

## Things to Look Out For
- MCP OAuth flow might require dynamic client registration

PR #16088, opened 2025-07-29
feat: add Model Context Protocol (MCP) support

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/16014

https://www.loom.com/share/015efa0447fa4d9ca537d863cd5703ba

Screenshot from 2025-07-29 17-09-04

image

Oauth authentication 

Screenshot from 2025-07-30 11-32-01
Screenshot from 2025-07-30 11-31-57

### Description:
Adds Model Context Protocol (MCP) support to Lightdash, enabling AI assistants to interact with Lightdash data. This PR:

- Implements a new MCP router with JSON-RPC endpoints for the protocol
- Creates a McpService to handle version information and site URL retrieval
- Adds security measures like origin validation to prevent DNS rebinding attacks
- Implements initial tool functionality to retrieve Lightdash version
- Includes tests for the McpService

The implementation follows the MCP specification with support for initialize, capabilities/get, tools/list, and tools/call methods, allowing AI assistants to discover and use Lightdash functionality.


rephus:
* **#16088** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16088?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16088?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/vite.config.ts
```diff
@@ -121,6 +121,10 @@ export default defineConfig({
                 target: 'http://localhost:8080',
                 changeOrigin: true,
             },
+            '/.well-known': { // MCP inspector requires .well-known to be on the root, but according to RFC 9728 (OAuth 2.0 Protected Resource Metadata) the .well-known endpoint is not required to be at the root level.
```
rephus, line 124: Related open issue https://github.com/modelcontextprotocol/inspector/issues/634

packages/backend/src/services/OAuthService/OAuthService.ts
```diff
@@ -83,4 +84,32 @@ export class OAuthService extends BaseService {
         if (refreshToken) return this.oauthModel.revokeToken(refreshToken);
         return false;
     }
+
+    public async registerClient({
```
rephus, line 95: Added new support for registering clients 

It should not be a requirement for MCP inspector, because of this https://github.com/modelcontextprotocol/inspector/pull/345 but I didn't manage to make it work 

Screenshot from 2025-07-30 11-15-47

owlas (commented):
This looks like it re-implements a lot of logic that's provided by the MCP server out of the box. Like exposing an endpoint for listing tools.

```
const mcpServer = new McpServer({...})

 app.post('/api/v1/mcp', async (req, res) => {
    const transport = new StreamableHTTPServerTransport();
    await mcpServer.connect(transport);
    await transport.handleRequest(req, res, req.body);
  });
```

packages/backend/src/ee/services/McpService/McpService.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        it('should handle server close operations', async () => {
+            const mcpServer = service.getServer();
+            await mcpServer.connect(transport);
+
+            // Test server close
+            await mcpServer.close();
+            expect(mcpServer.isConnected()).toBe(false);
+        });
+    });
+});
```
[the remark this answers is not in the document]
owlas: I guess this came from the claude file? Anyway absolutely excellent, all tests are very low value

packages/backend/src/routers/mcpRouter.ts
```diff
@@ -0,0 +1,71 @@
+import {
+    LightdashError,
+    MissingConfigError,
+    NotImplementedError,
+} from '@lightdash/common';
+// eslint-disable-next-line import/extensions
+import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
```
owlas: why import from `.js` ?

packages/frontend/vite.config.ts
owlas: do we need to make the target the `SITE_URL` in prod?

packages/backend/src/models/OAuth2Model.ts
```diff
@@ -308,6 +308,52 @@ export class OAuth2Model implements AuthorizationCodeModel {
         return false;
     }
 
+    async createClient({
+        clientName,
+        redirectUris,
+        grantTypes = ['authorization_code', 'refresh_token'],
+        scopes = ['read', 'write'],
```
owlas: let's default to no scopes if not provided

packages/backend/src/models/OAuth2Model.ts
```diff
@@ -308,6 +308,52 @@ export class OAuth2Model implements AuthorizationCodeModel {
         return false;
     }
 
+    async createClient({
```
owlas, line 311: I feel these have to be scoped to a specific organization, it doesn't matter yet because the tool is only getLightdashVersion

packages/backend/src/routers/oauthRouter.ts
```diff
@@ -241,6 +253,52 @@ oauthRouter.post('/revoke', async (req, res) => {
...
 });
 
+// Dynamic Client Registration endpoint (RFC 7591)
+oauthRouter.post('/register', async (req, res) => {
+    try {
+        const {
+            client_name,
+            redirect_uris,
+            grant_types = ['authorization_code', 'refresh_token'],
+            scope = 'read write',
```
owlas: let's introduce a specific mcp scope and have it such that _only_ mcp scopes can be accessed via DCR (this endpoint) - since this allows anybody to register an oauth client. we should also make sure we have logging for this on the AUDIT endpoint. also open a ticket on the infra repo for rate limiting

packages/backend/src/ee/services/McpService/McpService.ts
```diff
@@ -0,0 +1,54 @@
...
+            }),
+        );
+    }
+
+    public getServer(): McpServer {
+        return this.mcpServer;
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    public getLightdashVersion(): string {
```
owlas: lets add at least one permission check here:
- The instance should have an environment variable set to enable MCP
- The user (token) should have the CASL ability (oauth scope) for `mcp`

packages/backend/src/routers/mcpRouter.ts
rephus: It is required, see related open issue https://github.com/modelcontextprotocol/typescript-sdk/issues/440

packages/backend/src/models/OAuth2Model.ts
rephus: yeah, maybe,  clients can't be linked to organizations at the moment, only the oauth tokens. We can do it once we need it and add a db migration 

How can we get the organization when creating a client ? if the register client is no authenticated

packages/backend/src/ee/services/McpService/McpService.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        const user = context.authInfo.extra?.user;
+        const account = context.authInfo.extra?.account;
+        const { scopes } = account.authentication;
+
+        // TODO replace with CASL ability check
+        if (!scopes.includes('mcp:read')) {
+            throw new ForbiddenError('You are not allowed to access MCP');
+        }
```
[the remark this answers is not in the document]
rephus: that ability doesn't exist yet. This is work in progress

packages/backend/src/models/OAuth2Model.ts
owlas: oh yeah when the clients are dynamically created, there is no organization context (unless the client lets us make the DCR endpoint authenticated with a token). 

Then the best thing I think is to make clients created via DCR have very narrow scopes they can grant. you can mark this resolved!

packages/backend/src/routers/mcpRouter.ts
```diff
@@ -0,0 +1,92 @@
...
+                return await transport.handleRequest(authReq, res, req.body);
+            }
+        } catch (error) {
+            Logger.error(`MCP endpoint error: ${error}`);
+            if (error instanceof LightdashError) {
+                res.status(error.statusCode).json({ error: error.message });
+            } else {
+                Logger.error(`Unknown MCP endpoint error: ${error}`);
+                res.status(500).json({ error: 'Internal server error' });
+            }
```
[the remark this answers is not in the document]
rephus: this should follow mcp standards not Lightdash

packages/backend/src/auth/account/account.ts
```diff
@@ -174,3 +175,32 @@ export const fromSession = (
...
+    token: {
+        accessToken: string;
+        accessTokenExpiresAt?: Date;
+        refreshToken?: string;
+        refreshTokenExpiresAt?: Date;
+        scope?: string[];
+        client: { id: string };
+    },
+): OauthAccount => {
+    const [organization, user] = extractOrganizationFromUser(sessionUser);
```
owlas, line 190: can't this be extracted from the token?

owlas: @stevelikesmusic can you review?

packages/backend/src/routers/mcpRouter.ts
```diff
@@ -0,0 +1,95 @@
...
+        throw new MissingConfigError('MCP service not available');
+    }
+}
+
+// MCP endpoint - supports Streamable HTTP
+// Keep the MCP router as raw Express because:
+// - MCP protocol requirements don't align with REST/TSOA patterns
+// - We need full control over HTTP streaming and headers
+// - It follows the same pattern as other protocol-specific endpoints (OAuth)
+mcpRouter.all(
```
stevelikesmusic, line 37: Would it be worth adding rate limiting to these knowing that all the requests calling it will be automated?

packages/backend/src/auth/account/account.ts
rephus: We could, because the token contains the userUuid and orgUuid. in fact, we use those to find the session user in the middleware https://github.com/lightdash/lightdash/blob/main/packages/backend/src/controllers/authentication/middlewares.ts#L61 . The only downside is that we'll have to make an extra query to get org/user details to populate the account, I don't think there is any benefit on doing that.

stevelikesmusic: I agree with @rephus. We'll already have the user loaded up with org details so why not treat it like any other session user? Is there a possibility that the org on the token wouldn't match the org for the user herer?

stevelikesmusic (approved):

rephus:
I tested this integration with other MCP tools like cursor or claude. It seems there are still some authentication redirection issues these MCP clients . But we'll fix it on a separate PR. Here is a list of compatible clients https://modelcontextprotocol.io/clients

PR #16106, opened 2025-07-30
fix: dark mode, tree style display abnormal

yzqzy:
Description:
When the computer is in dark mode, the tree node body of the tree component is black due to style priority issues.

Changes:
* Add the !importmant attribute to your style
* Local docker development environment, you need to ignore the .pnpm-store directory

move_chart_style_error


yzqzy:
@IrakliJani  Could you please review my PR?

.gitignore
```diff
@@ -45,8 +45,11 @@ yarn-debug.log*
 yarn-error.log*
 
 .idea
+.pnpm-store
 node_modules
 package-lock.json
+pnpm-lock.yaml
+yarn.lock
```
joaoviana, line 52: let's please revert this change @yzqzy

packages/frontend/src/components/common/Tree/TreeItem.module.css
```diff
@@ -1,16 +1,16 @@
...
         cursor: pointer;
 
         &:hover {
-            background-color: var(--mantine-color-gray-0);
+            background-color: var(--mantine-color-gray-0) !important;
         }
 
         &:is([data-selected='true']) {
-            background-color: var(--mantine-color-blue-0);
+            background-color: var(--mantine-color-blue-0) !important;
```
joaoviana, line 13: I would pull from latest `main` and re-check if this is needed @yzqzy - I believe we fixed this CSS specificity issue and !important is not recommended

yzqzy: Ok, the problem is that there is an issue with the dark mode tree component color.

Issue #13176, opened 2025-01-09
Long chart titles need to be truncated when viewing editing charts that have been created in dashboards


PriPatel:
<img width="342" alt="Screenshot 2025-01-09 at 17 45 19" src="https://github.com/user-attachments/assets/8fbc5c8e-a338-40ff-a610-4509385311e6" />

The title name here should be truncated so that the space breadcrumbs can fit on the same line.

Abhijay007:
Hi, I am working on this issue will open a PR soon :)

PR #16142, opened 2025-07-31
fix : long chart titles need to be truncated when viewing editing charts

Abhijay007:
Closes : #13176 

### Description:

This PR adds text truncation to the chart title in the SavedChartsHeader component to ensure that long chart names don't break the layout and can fit on the same line as the breadcrumbs.

### Changes Made

File Modified: `packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx`
Added: CSS truncation properties to the chart title

### Preview : 

Before :  (_as you can see, the breadcrumb is hidden and the edit button is nearly inaccessible._)

image

After : 

image

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -334,7 +334,17 @@ const SavedChartsHeader: FC = () => {
                                     dashboardUuid={savedChart.dashboardUuid}
                                     dashboardName={savedChart.dashboardName}
                                 />
-                                <Title c="dark.6" order={5} fw={600}>
+                                <Title 
+                                    c="dark.6" 
+                                    order={5} 
+                                    fw={600}
+                                    sx={{
+                                        maxWidth: '300px',
+                                        whiteSpace: 'nowrap',
+                                        overflow: 'hidden',
+                                        textOverflow: 'ellipsis',
+                                    }}
+                                >
                                     {savedChart.name}
                                 </Title>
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
```diff
@@ -334,7 +334,19 @@ const SavedChartsHeader: FC = () => {
...
+                                    c="dark.6" 
+                                    order={5} 
+                                    fw={600}
+
+                                    sx={{
+                                        maxWidth: '300px',
+                                        whiteSpace: 'nowrap',
+                                        overflow: 'hidden',
+                                        textOverflow: 'ellipsis',
+                                    }}
```
stevelikesmusic: As part of our migration from Mantine v6 to v8 ([style guide](https://github.com/lightdash/lightdash/blob/main/packages/frontend/STYLE_GUIDE.md)), prefer to use the Mantine component and [style](https://mantine.dev/styles/style-props/) props:

```tsx
<Title 
  c="dark.6" 
  order={5} 
  fw={600}
  truncate={true}
  maw={300}
>
```

Not suggesting to migrate anything over to v8 in this PR, but avoid using `style`, `styles`, or `sx` on Mantine components. When we migrate this to v8, we'll need to add a CSS module for truncating—Matine dropped the `Title.truncate` in the latest version.

Abhijay007: Thanks for the sugggestion :) @stevelikesmusic I will keep that in mind while working on future PRs

Issue #16043, opened 2025-07-25
Explorer: Empty filters apply placeholder values


jesshitchcock:
### Description

When running queries with empty filters in the explorer, the SQL compiler sometimes applies placeholder values e.g. ('%1%') instead of ignoring the empty filter entirely.

<img width="2084" height="560" alt="Image" src="https://github.com/user-attachments/assets/4ba69116-4305-440c-b9d2-2f63534f0038" />

The expected behaviour would be either: 
- cannot apply a filter without a value OR 
- the filter does not get applied i.e. no defaults should be used. 

### Steps to Reproduce the Bug or Issue

1. In a Lightdash Project, Open the Explorer
2. Select some fields
3. Apply filters but do not enter in values e.g. 

<img width="743" height="581" alt="Image" src="https://github.com/user-attachments/assets/cf3c1153-8350-43ef-bb56-a17adb8cce57" />

4. Review the compiled SQL, note that default values are applied for the following conditions: 
- string: `is` and `is not`
- string: `includes` & `does not include`
- string: `starts with` & `ends with`
- numeric/metric:  `is less than` & `is greater than` & `is less than or equal to`
- numeric/metric: `is between` & `is not between`

``` sql 
WITH metrics AS (
SELECT
  `accounts`.account_name AS `accounts_account_name`,
  `accounts`.account_id AS `accounts_account_id`,
  AVG(`deals`.amount) AS `deals_average_amount`,
  SUM(`deals`.amount) AS `deals_total_amount`
FROM `lightdash-healthcare-demo`.`lightdash_demo_saas`.`accounts` AS `accounts`
INNER JOIN `lightdash-healthcare-demo`.`lightdash_demo_saas`.`deals` AS `deals`
  ON (`accounts`.account_id) = (`deals`.account_id)
WHERE ('true' = 'true' or segment != 'Enterprise') AND ((
  (`accounts`.industry) IN ('1')
) AND (
  ((`accounts`.industry) NOT IN ('1' ) OR (`accounts`.industry) IS NULL)
) AND (
  LOWER(`accounts`.industry) LIKE LOWER('%1%')
) AND (
  LOWER(`accounts`.industry) NOT LIKE LOWER('%1%')
) AND (
  (`accounts`.industry) LIKE '1%'
) AND (
  (`accounts`.industry) LIKE '%1'
))
GROUP BY 1,2
)
SELECT
  *
FROM metrics
WHERE ((
  true
) AND (
  true
) AND (
[...]

Issue #7083, opened 2023-09-16
Query generated in number filters with empty values on where clause of less than or greater than is incorrect


ujwalkumar1995:
### Description

When applying a number filter with greater than or less than in the where clause, the query generated does not seem correct. It would  return unexpected results even though nothing was selected in the values.

### To Reproduce

- Apply a number filter.
- Use a where clause of less than or greater than.
- Don't select any value for the where clause and see the query generated.
- The query generated does not seem correct. Without any values in the where clause, the query should not have been impacted.

Have attached video for reference.

https://github.com/lightdash/lightdash/assets/20976813/0e0d945c-734c-41dc-89f1-b5e46b891edc

### App Version

Running latest code from master in local machine.

PR #16151, opened 2025-07-31
fix: Empty filters apply placeholder values

tatianainama:
Closes: #16043, #7083

### Description:
Improved handling of empty filter values in SQL generation. 

1. String filters with empty values now return `true` instead of generating invalid SQL
2. String filters with mixed empty and non-empty values properly filter out the empty strings
3. Number filters with empty or insufficient values (e.g., only one value for IN_BETWEEN) now return `true`

~~todo: Validate correct usage of `''` strings, as it can be a valid filter~~ We don't support filter by empty strings atm

| Before | After |
|--------|--------|
|  |  |
| image | image |

Follow-up: #16152 tackling range input improvements

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.ts
+++ b/packages/common/src/compiler/filtersCompiler.ts
@@ -57,7 +57,7 @@ export const renderStringFilterSql = (
     filter: FilterRule<FilterOperator, unknown>,
     stringQuoteChar: string,
 ): string => {
-    const filterValues = filter.values;
+    const filterValues = filter.values?.filter((v) => v !== '');
 
     switch (filter.operator) {
         case FilterOperator.EQUALS:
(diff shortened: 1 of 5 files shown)


tatianainama:
* **#16152** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16152?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16151** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16151?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -57,7 +57,7 @@ export const renderStringFilterSql = (
     filter: FilterRule<FilterOperator, unknown>,
     stringQuoteChar: string,
 ): string => {
-    const filterValues = filter.values;
+    const filterValues = filter.values?.filter((v) => v !== '');
```
almeidabbm, line 60: Does this mean we can't filter a dimension value by empty string? 🤔

tatianainama: That's right. This logic was parsing all the empty strings into unwanted sql like `%1%`. With these changes, now they are filtered out.
I think if we want to allow filter by `''` we might want to introduce a new filter operator like `IS_EMPTY` / `NOT_EMPTY`, wdyt?

almeidabbm: Yep, had a second look think this is good, I just checked and our string filter input doesn't seem to allow for empty strings we might not need those 🤔

almeidabbm (approved):

tatianainama:
### Merge activity

* **Aug 5, 8:27 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/16151).
* **Aug 5, 8:28 AM UTC**: @tatianainama merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/16151).

PR #16143, opened 2025-07-31
feat: refactor Project and Organaization services to enable iframe explores

stevelikesmusic:
Closes: #N/A

### Description:
Pulled out of the [PR](https://github.com/lightdash/lightdash/pull/16118) for enabling explores in embedded iframes 

Refactored authentication to use `req.account` instead of `req.user` in organization and project controllers. 

test-frontend
test-backend

PR #16118, opened 2025-07-30
feat: enable 'Explore from here' for iframe and embed URLs

stevelikesmusic:
Closes:

PR #16159, opened 2025-07-31
chore: refactor embed provider and hooks for iframe explores

stevelikesmusic:
Closes: 

### Description:
Refactored the embedded app experience to prepare for embed iframe explores. Created a new `EmbeddedApp` component that wraps the `EmbedProvider` and handles navigation between dashboard and explore views. This will let us add a new route that can facilitate navigation between Dashboard and what will be the iframe Explore. 

Some hooks were tweaked to disable if we're in an embedded context because we know those requests will fail—and they're not needed for the embedded experience.

### DIFF
--- a/packages/frontend/src/hooks/useSpaces.ts
+++ b/packages/frontend/src/hooks/useSpaces.ts
@@ -13,7 +13,7 @@ import {
 } from '@tanstack/react-query';
 import { lightdashApi } from '../api';
 import useToaster from './toaster/useToaster';
-import useUser from './user/useUser';
+import { useAccount } from './user/useAccount';
 
 const getSpaceSummaries = async (projectUuid: string) => {
     return lightdashApi<SpaceSummary[]>({
@@ -35,7 +35,7 @@ export const useSpaceSummaries = (
     includePrivateSpaces: boolean = false,
     queryOptions?: UseQueryOptions<SpaceSummary[], ApiError>,
 ) => {
-    const { data: user } = useUser(true);
+    const { data: account } = useAccount();
     return useQuery<SpaceSummary[], ApiError>(
         ['projects', projectUuid, 'spaces'],
         () => getSpaceSummaries(projectUuid!),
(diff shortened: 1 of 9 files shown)


stevelikesmusic:
* **#16118** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16118?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16159** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16143** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16143?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/useSpaces.ts
```diff
@@ -49,7 +49,7 @@ export const useSpaceSummaries = (
                               !!user &&
                               hasDirectAccessToSpace(space, user.userUuid),
                       ),
-            enabled: !!projectUuid,
+            enabled: !!projectUuid && !!user,
```
rephus: sometimes you do 

>  enabled: !!projectUuid && !!user

And sometimes 

> enabled: ability?.can('manage', 'Explore'),

why is that ? should we standandarize this ?

stevelikesmusic: In this case, we already had `user` in scope. I suppose we use an ability. Right now, spaces require a registered user, so it seemed simpler to use `user` in this case. Most times, definitely prefer CASL.

stevelikesmusic: I made a change to use account here instead of useUser. Good call-out!

Issue #16005, opened 2025-07-24
Regular Expression Support for Explore Filters


jesshitchcock:
### Description
Users need the ability to filter for literal special characters (like underscores) in the Explore interface, but currently have no way to escape wildcard characters in filters.

### Current Problem
When a user wants to filter for addresses containing the literal string "_us_", the underscore characters are treated as SQL wildcards (matching any single character) rather than literal underscores. This returns unwanted results that match patterns like "aus", "bus", "cus", etc.
<img width="1527" height="657" alt="Image" src="https://github.com/user-attachments/assets/c5ff4072-de38-4525-a021-4a70030717d9" />

### Proposed Solution
Add regular expression filter support to the Explore interface. This would allow users to specify regex patterns that compile to REGEXP_CONTAINS() functions in the generated SQL.
Example Implementation:

User selects filter: addresses.street_address includes `r'_us_'`
Generated SQL: `REGEXP_CONTAINS(addresses.street_address, r'_us_')`
This works: 
``` sql
SELECT
  `addresses`.street_address AS `addresses_street_address`
FROM `lightdash-analytics`.`dbt_jhitchcock`.`addresses` AS `addresses`
WHERE ((
  REGEXP_CONTAINS(`addresses`.street_address, r'_us_')
))
GROUP BY 1
ORDER BY `addresses_street_address`
LIMIT 500
```
### What problem does this solve?

- Enables filtering for literal special characters (_, %, etc.) in the Explore interface
- Provides more powerful pattern matching capabilities for advanced users
- Eliminates the need to switch to SQL Runner for basic regex filtering
- Improves user experience when working with data containing special characters

### Current Workaround
Users must use the SQL Runner with REGEXP functions, which breaks the self-service nature of the Explore interface.

Abhijay007:
Hi @jesshitchcock,  I'm currently working on this issue and plan to open a PR soon. Before that, I had a quick question regarding the SQL generation part.

In the proposed solution, you mentioned:

> User selects filter: addresses.street_address includes r'_us_'
> Generated SQL: REGEXP_CONTAINS(addresses.street_address, r'_us_')
> This works:
> 
> ```
> SELECT
>   `addresses`.street_address AS `addresses_street_address`
> FROM `lightdash-analytics`.`dbt_jhitchcock`.`addresses` AS `addresses`
> WHERE (
>   REGEXP_CONTAINS(`addresses`.street_address, r'_us_')
> )
> GROUP BY 1
[...]

PR #16177, opened 2025-08-01
feat: added regular expression support for explore filters 

Abhijay007:
Closes: #16005 

### Description:

This PR enables filtering for literal special characters (_, %) in `INCLUDE/NOT_INCLUDE` filters across all supported databases.

Changes Made: 

Problem: Users couldn't filter for literal special characters - `_us_` would match aus, bus, cus instead of literal underscores.

- Smart Detection: Auto-detects special characters and uses appropriate SQL
- Database-Specific: REGEXP_CONTAINS (BigQuery), REGEXP_LIKE (Snowflake), ~* (PostgreSQL/Redshift)
- Fallback: LIKE with escaping for databases without regex (Databricks, Trino)
- Backward Compatible: No breaking changes

Preview : 

As you can see in the lightdash demo mode here, when I try to search for `_ut` in the filters, it is also giving me emails with `ut` as well : 

includeFilter-before

And as you can see in the localhost now, if I search for `_pe`, I will only get `return_pending` but not `shipped` as both contain `pe` : 

includeFilter-after

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.test.ts
+++ b/packages/common/src/compiler/filtersCompiler.test.ts
@@ -77,6 +77,7 @@ import {
     stringFilterDimension,
     stringFilterRuleMocks,
 } from './filtersCompiler.mock';
+import { SupportedDbtAdapter } from '../types/dbt';
 
 const formatTimestamp = (date: Date): string =>
     moment(date).format('YYYY-MM-DD HH:mm:ss');
@@ -658,6 +659,7 @@ describe('Filter SQL', () => {
                 stringFilterDimension,
                 stringFilterRuleMocks.includeFilterWithSingleVal,
                 "'",
+                SupportedDbtAdapter.BIGQUERY,
             ),
         ).toBe(stringFilterRuleMocks.includeFilterWithSingleValSQL);
     });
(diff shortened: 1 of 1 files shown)


packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -75,16 +139,40 @@ export const renderStringFilterSql = (
...
-            const includesQuery = filterValues.map(
-                (v) => `LOWER(${dimensionSql}) LIKE LOWER('%${v}%')`,
-            );
+            
+            const includesQuery = filterValues.map((v) => {
+                const stringValue = String(v);
+                
+                // Use regex if database supports it and special characters are present
+                if (adapterType && supportsRegexFunctions(adapterType) && containsLikeSpecialCharacters(stringValue)) {
+                    return generateRegexSql(dimensionSql, stringValue, adapterType);
```
ZeRego, line 148: Is there a problem if we use this SQL even if there are no special char? 
Thinking if we can simplify this PR to a single function with switch case. Example:

```
    switch (adapterType) {
        case SupportedDbtAdapter.BIGQUERY:
            return `REGEXP_CONTAINS(${dimensionSql}, r'${stringValue}')`;
        case SupportedDbtAdapter.SNOWFLAKE:
            return `REGEXP_LIKE(${dimensionSql}, '${stringValue}', 'i')`;
        case SupportedDbtAdapter.POSTGRES:
        case SupportedDbtAdapter.REDSHIFT:
            return `${dimensionSql} ~* '${stringValue}'`;
        default:
            const escapedValue = escapeLikeSpecialCharacters(stringValue);
            return `LOWER(${dimensionSql}) LIKE LOWER('%${escapedValue}%')`;
    }
```

Issue #16180, opened 2025-08-03
Document Icon in Settings Not Visible Until Hover


Abhijay007:
### Description

Issue:

In the settings section, the document icon is not visible initially due to its background color matching the container. It only becomes visible when hovering over it, which makes it seem like it’s missing.

<img width="493" height="128" alt="Image" src="https://github.com/user-attachments/assets/90626adf-a1db-49c0-83de-239422abb962" />

Expected Behavior:

The icon should be clearly visible at all times, not just on hover.

Suggestion:

Adjust the icon’s default color or background so that it contrasts properly with the surrounding UI.

### Steps to Reproduce the Bug or Issue

1. Go to the Settings page.
2. Scroll down to the section where the version is displayed.
3. Hover over the icon next to the version number (document icon).
4. Notice that the icon is only visible on hover due to its background color blending with the page.

PR #16187, opened 2025-08-04
fix: document icon in settings not visible until hover 

Abhijay007:
Closes: #16180

### Description:

The PR fixes the documentation icon visibility issue in the footer, where the book icon was only visible on hover due to poor contrast.

### Changes made

`packages/frontend/src/components/AboutFooter.tsx`: Updated icon color and button variants
change: Removed custom hover styles in favor of consistent light variant behavior

### Preview

| Before | After |
|--------|-------|
| Before | After |

### DIFF
--- a/packages/frontend/src/components/AboutFooter.tsx
+++ b/packages/frontend/src/components/AboutFooter.tsx
@@ -82,7 +82,12 @@ const AboutFooter: FC<{ minimal?: boolean; maxWidth?: number }> = ({
                             href="https://docs.lightdash.com/"
                             target="_blank"
                         >
-                            <ActionIcon color="gray.7" p="xs" size="lg">
+                            <ActionIcon
+                                color="gray.7"
+                                p="xs"
+                                size="lg"
+                                variant="subtle"
+                            >
                                 <MantineIcon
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/AboutFooter.tsx
```diff
@@ -85,7 +85,7 @@ const AboutFooter: FC<{ minimal?: boolean; maxWidth?: number }> = ({
                                 href="https://docs.lightdash.com/"
                                 target="_blank"
                             >
-                                <ActionIcon color="gray.7" p="xs" size="lg">
+                                <ActionIcon variant="light" p="xs" size="lg">
```
stevelikesmusic: Try this:
```suggestion
                                <ActionIcon
                                    variant="subtle"
                                    color="gray.7"
                                    p="xs"
                                    size="lg"
                                >
```

stevelikesmusic (changes requested):
Nice catch. Looks like this may have been a regression from a recent v8 migration.

magnew:
I made a bad rebase. I'll fix it.

PR #16195, opened 2025-08-04
chore: update style guide with m8 suggestions

magnew:
Updates the style guide with a few rules I found helpful to get Claude to upgrade components a little better.

### DIFF
--- a/packages/frontend/STYLE_GUIDE.md
+++ b/packages/frontend/STYLE_GUIDE.md
@@ -174,7 +174,7 @@ import styles from './Component.module.css';
 
 Do not include a `<file>.css.d.ts` file for css modules. We are using Vite and don't need them.
 
-In CSS modules, avoid using `!imporant` where possible. It's better to use the `classNames` API for increased specificity. 
+In CSS modules, avoid using `!important` where possible. It's better to use the `classNames` API for increased specificity. 
 
 ### 3. Important Guidelines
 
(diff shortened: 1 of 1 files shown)


magnew:
* **#16195** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16195?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/STYLE_GUIDE.md
```diff
@@ -210,6 +214,26 @@ Before moving styles to CSS modules, analyze if they're actually needed:
...
+```tsx
+// ❌ Bad - Text just for styles
+<Text fw={500} c="dimmed" fz="sm" mt="sm">
+  Parameters
+</Text>
+
+// ✅ Good - inherits local text styles
+<Box component="span" fw={500} c="dimmed" fz="sm" mt="sm">
+  Parameters
+</Box>
```
joaoviana: i'm not too sure on this use-case. Wouldn't this cause issues with text elements not being a `p` tag anymore? 🤔

packages/frontend/STYLE_GUIDE.md
```diff
@@ -172,6 +172,10 @@ import styles from './Component.module.css';
 </Card>
 ```
 
+Do not include a `<file>.css.d.ts` file for css modules. We are using Vite and don't need them.
+
+In CSS modules, avoid using `!important` where possible. It's better to use the `classNames` API for increased specificity.
```
joaoviana, line 177: 🙌

packages/frontend/STYLE_GUIDE.md
magnew: You might be right, I can take this out and we can see. In practice I've been finding that when <Text> elements are there, things look shitty cause the test styles override the styles that would otherwise be inherited. In every case I've found, removing the text element made things work better. BUT, this might be too aggressive. I'll remove it and we can revisit.

joaoviana: yea "aggressive" is the right word here 👍 agree with your points but just thinking about edge cases not covered 🤔

Issue #16280, opened 2025-08-07
DB Migration for roles


stevelikesmusic:
## Overview 

Create a migration that creates three database tables for enabling custom roles. This should also implement types and entities but not models or services. 

### `roles` table

This is the shell for roles and part of the join table we'll add. Records here are user-defined.

```sql
CREATE TABLE IF NOT EXISTS roles (
    role_uuid UUID PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    description TEXT,
    organization_uuid UUID NOT NULL REFERENCES organizations,
    created_by UUID NOT NULL REFERENCES users,
    created_at TIMESTAMP DEFAULT NOW(),

    UNIQUE(name, organization_uuid)
);
```

### `scoped_roles` table

This is our join table to associate scopes (defined in the app code) to a custom role. 

```sql
CREATE TABLE IF NOT EXISTS scoped_roles (
    role_uuid UUID REFERENCES roles(id) ON DELETE CASCADE,
    scope_name VARCHAR(100) NOT NULL,
    granted_at TIMESTAMP DEFAULT NOW(),
    granted_by UUID NOT NULL REFERENCES users,  
    PRIMARY KEY (role_uuid, scope_uuid)
);

CREATE INDEX idx_scoped_roles_role ON scoped_roles(role_uuid);
CREATE INDEX idx_scoped_roles_scope ON scoped_roles(scope_name);
```

### `user_roles` table

```sql
CREATE TABLE IF NOT EXISTS user_roles (
    user_id UUID NOT NULL,
    role_id UUD NOT NULL,
    project_uuid UUID,
    organization_uuid UUID,
    PRIMARY KEY (user_uuid, role_uuid),
    
    -- If user is removed from org, cascade delete their role assignments
    FOREIGN KEY (user_id, org_id) REFERENCES user_orgs(user_id, org_id) 
        ON DELETE CASCADE,
    
    -- Prevent role deletion if assigned to users
    FOREIGN KEY (role_id, org_id) REFERENCES roles(id, org_id) 
        ON DELETE RESTRICT
)

CREATE INDEX idx_user_roles_organization ON user_roles(organization_uuid);
CREATE INDEX idx_user_roles_project ON user_roles(project_uuid);
```

PR #16319, opened 2025-08-08
feat: add migration for custom roles

stevelikesmusic:
Closes: #16280 (The parent issue has the related sub-issues that will be a part of this stack)

### Description:
Adds database schema for custom roles and permissions system with three new tables:

1. `roles` - Stores organization-specific roles with name and description
2. `scopes` - Defines available permissions (resource + action combinations)
3. `scoped_roles` - Join table linking roles to their granted permissions

The migration initializes the system with common permission scopes for organization management, projects, dashboards, charts, spaces, exports, and AI agents.

This is the first go at adding roles. We'll have a subsequent migration as well to add more roles, as well as an update to the seed the database correctly for e2e tests. 

I move scope definitions out to a separate PR: https://github.com/lightdash/lightdash/pull/16353

[Context](https://www.notion.so/lightdash/Custom-roles-permission-233a63207a7a8027a6d4d68a4b210045)

### DIFF
--- a/packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
+++ b/packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
@@ -1,0 +1,109 @@
+import { Knex } from 'knex';
+
+const RolesTableName = 'roles';
+const ScopedRolesTableName = 'scoped_roles';
+const OrgMembershipsTableName = 'organization_memberships';
+const ProjectMembershipsTableName = 'project_memberships';
+const ProjectGroupAccessTableName = 'project_group_access';
+
+const addRoleUuidTo = async (knex: Knex, tableName: string) => {
+    if (!(await knex.schema.hasColumn(tableName, 'role_uuid'))) {
+        await knex.schema.alterTable(tableName, (table) => {
+            table
+                .uuid('role_uuid')
+                .nullable()
+                .references('role_uuid')
+                .inTable('roles')
+                // Prevent deleting a role if assigned to a user
+                .onDelete('RESTRICT');
+        });
+    }
+};
+
+export async function up(knex: Knex): Promise<void> {
+    // Create roles table
+    await knex.schema.createTable(RolesTableName, (rolesTable) => {
+        rolesTable
+            .uuid('role_uuid')
+            .primary()
+            .defaultTo(knex.raw('uuid_generate_v4()'));
+        rolesTable.string('name').notNullable();
+        rolesTable.text('description').nullable();
+        rolesTable
+            .uuid('organization_uuid')
+            .nullable()
+            .references('organization_uuid')
+            .inTable('organizations')
+            .onDelete('CASCADE');
+
+        rolesTable
+            .uuid('created_by')
+            .nullable()
+            .references('user_uuid')
+            .inTable('users')
+            .onDelete('SET NULL');
+        rolesTable.string('owner_type').notNullable().defaultTo('user');
+        rolesTable.timestamp('created_at').defaultTo(knex.fn.now());
+        rolesTable.timestamp('updated_at').defaultTo(knex.fn.now());
+
+        rolesTable.unique(['name', 'organization_uuid']);
+        rolesTable.index('organization_uuid');
+    });
+
+    // 'user' roles must have an organization_uuid, 'system' roles defined by Lightdash must not
+    await knex.raw(`
+        ALTER TABLE roles
+            ADD CONSTRAINT user_roles_need_org
+                CHECK (
+                    (owner_type = 'user' AND organization_uuid IS NOT NULL) OR
+                    (owner_type = 'system' AND organization_uuid IS NULL)
+                )
+    `);
+
+    await Promise.all([
+        addRoleUuidTo(knex, OrgMembershipsTableName),
+        addRoleUuidTo(knex, ProjectMembershipsTableName),
+        addRoleUuidTo(knex, ProjectGroupAccessTableName),
+    ]);
+
+    // Create scoped_roles join table
+    await knex.schema.createTable(ScopedRolesTableName, (scopedRolesTable) => {
+        scopedRolesTable
+            .uuid('role_uuid')
+            .notNullable()
+            .references('role_uuid')
+            .inTable('roles')
+            .onDelete('CASCADE');
+        scopedRolesTable.string('scope_name').notNullable();
+        scopedRolesTable.timestamp('granted_at').defaultTo(knex.fn.now());
+        scopedRolesTable
+            .uuid('granted_by')
+            .nullable()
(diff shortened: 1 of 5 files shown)


stevelikesmusic:
* **#16319** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16319?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#16323](https://github.com/lightdash/lightdash/pull/16323) <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16323?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#16380](https://github.com/lightdash/lightdash/pull/16380) <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16380?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
```diff
@@ -0,0 +1,184 @@
...
+const ScopedRolesTableName = 'scoped_roles';
+
+export async function up(knex: Knex): Promise<void> {
+    // Create roles table
+    await knex.schema.createTable(RolesTableName, (table) => {
+        table
+            .uuid('role_uuid')
+            .primary()
+            .defaultTo(knex.raw('uuid_generate_v4()'));
+        table.string('name', 100).notNullable();
```
rephus: why 100 ? is 255 the default ?

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
```diff
@@ -0,0 +1,184 @@
...
+            .uuid('organization_uuid')
+            .notNullable()
+            .references('organization_uuid')
+            .inTable('organizations')
+            .onDelete('CASCADE');
+        table
+            .uuid('created_by')
+            .references('user_uuid')
+            .inTable('users')
+            .onDelete('SET NULL');
```
rephus, line 44: SET NULL makes sense here

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
```diff
@@ -0,0 +1,184 @@
...
+        table.timestamp('created_at').defaultTo(knex.fn.now());
+        table.timestamp('updated_at').defaultTo(knex.fn.now());
+
+        table.unique(['name', 'organization_uuid']);
+    });
```
rephus, line 51: are we planning to do filtering by org_uuid ? for example, to get all roles per org, in that case. I'd suggest to add an index on `organization_uuid`

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
```diff
@@ -0,0 +1,184 @@
...
+        table
+            .uuid('scope_uuid')
+            .notNullable()
+            .references('scope_uuid')
+            .inTable('scopes')
+            .onDelete('CASCADE');
+        table.timestamp('granted_at').defaultTo(knex.fn.now());
+        table
+            .uuid('granted_by')
+            .notNullable()
```
rephus:
```suggestion
            .onDelete('SET NULL')
```

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
```diff
@@ -0,0 +1,184 @@
...
+            .references('user_uuid')
+            .inTable('users');
+
+        table.index('role_uuid', 'idx_role_scopes_role');
```
rephus: I don't think we need to add `idx_role_scopes_role`  , it will be created automatically

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
```diff
@@ -0,0 +1,184 @@
...
+            resource: 'Space',
+            action: 'manage',
+            description: 'Manage spaces and associated content',
+            is_commercial: false,
+        },
+
+        // Export permissions
+        {
+            resource: 'Export',
+            action: 'csv',
```
rephus: is this not the other way arround ? resource should be `csv` and action `export`

packages/common/src/authorization/types.ts
```diff
@@ -8,7 +8,10 @@ export type AbilityAction =
     | 'manage'
     | 'promote'
     | 'update'
-    | 'view';
+    | 'view'
+    | 'csv'
```
rephus: these 3 should be `exports`

packages/common/src/types/roles.ts
```diff
@@ -0,0 +1,80 @@
...
+    status: 'ok';
+    results: Role[];
+};
+
+export type ApiGetRoleResponse = {
+    status: 'ok';
+    results: RoleWithScopes;
+};
+
+export type ApiCreateRoleResponse = {
```
rephus: perhaps we can reuse some of these types ? unless you are planning to change them

rephus (commented):
OVerall look good, I made a few suggestions on the migration, let's make sure it works before merging to avoid patching the migration later.

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
stevelikesmusic: That's right. I suppose it can simply be the default. No strong reason to limit the length

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
stevelikesmusic: Ah yep! I have an update coming.

packages/backend/src/database/migrations/20250807212731_add_custom_roles.ts
stevelikesmusic: I was trying to figure out the best way to express this. Since we're moving scopes to be defined in JS, then we can model them a little better:

```ts
const scope: Scope = {
  name: 'export:dashboard',
  getModifer: () => ({ type: 'csv' }),
  isEnterprise: false,
}
```

stevelikesmusic: I added a specific subject for each in the scopes PR: so DashboardCsv, DashboardImage, etc. 
https://github.com/lightdash/lightdash/blob/d70bf7c8f7af5aeb994aeb87e7551bcc7cc1e7f1/packages/common/src/authorization/types.ts#L32-L34

rephus (approved):

Issue #16259, opened 2025-08-07
I want my current dashboard filters to appear when I create a scheduled delivery


jpetey75:
### Description

I want my current dashboard filters to appear when I create a scheduled delivery. Maybe it should be an optional toggle because sometimes I don't want my current filters applied in my scheduled delivery, but I think most of the time this workflow makes the most sense (to copy the filter values current state into the new scheduled delivery).

### What problem does this solve?

I can QA a set of filters to make sure it looks correct, then pass them directly into a scheduled delivery setting without re-creating the filters in those settings.

PR #16357, opened 2025-08-12
feat: scheduled delivery filters default to current values

magnew:
Closes: #16259

### Description:

Makes it so that, when creating a scheduled delivery, the filters default to the current live view in the dashboard. 

This means that now temporary filters get saved with the delivery, which is probably what people want. When viewing/editing the delivery later, we show all of the filters, along with a message about the older (now saved) filter values.

Screenshot 2025-08-13 at 18 06 04

Things to test:
- New schedulers default to existing filters - saved and temporary
- Deliveries have all filters (including temporary) applied
- Editing filters shows all values.


magnew:
* **#16449** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16357** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/CsvService/CsvService.ts
```diff
@@ -901,7 +901,6 @@ export class CsvService extends BaseService {
         const dashboardFilters = overrideDashboardFilters || dashboard.filters;
 
         if (schedulerFilters) {
-            // Scheduler filters can only override existing filters from the dashboard
```
magnew, line 904: I changed this assumption. Can we think of any reason this is bad?

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -399,6 +419,26 @@ const SchedulerFilters: FC<SchedulerFiltersProps> = ({
...
+                                    key={filter.id}
+                                    dashboardFilter={filter}
+                                    schedulerFilter={filter}
+                                    onChange={handleUpdateSchedulerFilter}
+                                    onRevert={() => revertFilter(filter.id)}
+                                    hasChanged={false}
+                                />
+                            </Box>
+                        );
+                    })}
```
joaoviana: Image


```tsx
  <Stack spacing="xs">
                        {savedFiltersNotInDashboard.length > 0 && (
                            <Text fz="sm" c="dimmed" fw={500}>
                                The following filters are applied to this
                                scheduled delivery but no longer exist in the
                                dashboard
                            </Text>
                        )}
                        {savedFiltersNotInDashboard?.map((filter) => {
                            return (
                                <Box key={filter.id} pl="md">
                                    <FilterItem
                                        key={filter.id}
                                        dashboardFilter={filter}
                                        schedulerFilter={filter}
                                        onChange={handleUpdateSchedulerFilter}
                                        onRevert={() => revertFilter(filter.id)}
                                        hasChanged={false}
                                    />
                                </Box>
                            );
                        })}
                    </Stack>
```

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -306,22 +307,40 @@ const SchedulerFilters: FC<SchedulerFiltersProps> = ({
...
-    const originalDashboardFilters = dashboard?.filters;
-    const dashboardFilterIds = useMemo(
-        () => new Set(dashboard?.filters.dimensions.map((f) => f.id)),
-        [dashboard?.filters.dimensions],
-    );
 
     const [schedulerFiltersData, setSchedulerFiltersData] = useState<
         SchedulerFilterRule[] | undefined
-        // NOTE: Filter out any filters that are not in the dashboard anymore
-    >(schedulerFilters?.filter((sf) => dashboardFilterIds.has(sf.id)));
```
joaoviana, line 318: This filtering was great when only overrides could be applied (existing filters)
I wonder if something smarter could be applied here to de-dupe filters that target the same id and same value.

Steps
1. Create filter in edit mode and save
2. Create similar filter in view mode and open scheduled delivery filter

I see that it makes sense to have both because they show similarly on the dashboard, but worth considering

Image

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -399,6 +419,26 @@ const SchedulerFilters: FC<SchedulerFiltersProps> = ({
...
+                        return (
+                            <Box key={filter.id} pl="md">
+                                <FilterItem
+                                    key={filter.id}
+                                    dashboardFilter={filter}
+                                    schedulerFilter={filter}
+                                    onChange={handleUpdateSchedulerFilter}
+                                    onRevert={() => revertFilter(filter.id)}
+                                    hasChanged={false}
+                                />
```
joaoviana, line 514: wish there was a way to remove this scheduled delivery filter - I can't remove it from the scheduled delivery after this case

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -306,22 +307,40 @@ const SchedulerFilters: FC<SchedulerFiltersProps> = ({
...
+        if (
+            allFilters?.dimensions &&
+            allFilters.dimensions.length > 0 &&
+            (!schedulerFilters || schedulerFilters.length === 0)
+        ) {
+            // Convert live filters to scheduler filter format and initialize the form
+            const liveSchedulerFilters = allFilters.dimensions.map(
+                (filter): SchedulerFilterRule => ({
+                    ...filter,
+                    tileTargets: undefined, // Remove tile targets for scheduler context
```
joaoviana: `tileTargets` were made `undefined` because scheduled delivery filters were initially implemented to only override existing filters
The util to `applyDimensionOverrides` would cross-check the filter saved with the scheduled delivery filter to apply the correct `tileTargets`

So if we're now adding "scheduler filters that don't exist in dashboard saved filters" we need to send the `tileTargets` 

This means that the type for `SchedulerFilter` needs to change so, if someone adds a new filter and sets specific tileTargets, the logic knows how to handle that. 
Perhaps `SchedulerFilter` could be either `tileTargets: undefined` (if SchedulerFilterOverride) or defined if "SchedulerFilter"

LMK what you think

joaoviana (commented):
Left a few Qs - also would be cool to experiment the toggle as described on the original ticket: 

```
Maybe it should be an optional toggle because sometimes I don't want my current filters applied in my scheduled delivery,
```

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Yea, I saw this, but since they can be duplicated on the dashboard, I figured we should just keep it the same here. The goal is to make it more obvious how the filters you have applied to the dashboard will apply to the delivery, so I think we should keep mirroring whats there.

Screenshot 2025-08-12 at 12 10 05

There's a separate question about if the dashboard should let you do this 😄

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Yea, good point. I'll update the types and we can also think about whether this information needs to be displayed.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Fair enough. I can look into a remove button.

magnew: Added removal! 
Screenshot 2025-08-13 at 18 10 15

Could be added to the other filters too, but I agree it was most important here.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Cool. Very updated. I made it so the tile targets are saved and `applyDimensionOverrides`.

I also made it so tiles that are no longer on the dashboard list the tiles the apply to. 

There is more that could be done UI-wise there. Maybe all of the scheduler filters should say their tiles, but those ones seemed most important cause the info is nowhere else.

magnew:
Closed and opened this just to restart Okteto. Not sure if that's the best way :D

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: This is perfect

packages/common/src/types/filter.ts
```diff
@@ -348,27 +347,36 @@ export const removeFieldFromFilterGroup = (
...
+    );
+
+    // Add scheduler filters that don't exist in dashboard saved filters
+    const existingIds = new Set(dashboardFilters.dimensions.map((d) => d.id));
+    const newDimensions = overrideArray.filter(
+        (schedulerFilter) => !existingIds.has(schedulerFilter.id),
+    );
+    overriddenDimensions.push(...newDimensions);
+
+    return overriddenDimensions;
```
joaoviana, line 378: this is great

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -177,17 +183,46 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
...
+                                    ', ',
+                                )}`}
+                                fz="xs"
+                                multiline
+                                w={200}
+                            >
+                                <Text fz="xs" color="gray.6" span>
+                                    {`Applies to ${tilesWithFilter.length} tiles`}
+                                </Text>
+                            </Tooltip>
```
joaoviana, line 208: Idk how to replicate this 🤔 can you give me a step list?

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -177,17 +183,46 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
...
                         <ActionIcon
                             size="xs"
-                            disabled={isEditing || hasChanged}
                             onClick={() => {
-                                setIsEditing(true);
+                                if (isEditing) {
+                                    setIsEditing(false);
+                                } else {
+                                    setIsEditing(true);
+                                }
```
joaoviana:
```suggestion
                               setIsEditing(!isEditing);
                                
```

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -200,11 +235,17 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
...
                                     ...dashboardFilter,
                                     operator,
-                                    tileTargets: undefined,
-                                });
+                                    values:
+                                        operator === FilterOperator.NOT_NULL ||
+                                        operator === FilterOperator.NULL
+                                            ? undefined
+                                            : dashboardFilter.values,
+                                };
```
joaoviana: I don't understand this change 🤔

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -214,10 +255,7 @@ const FilterItem: FC<SchedulerFilterItemProps> = ({
                             field={field}
                             rule={schedulerFilter ?? dashboardFilter}
                             onChange={(newFilter) => {
-                                onChange({
-                                    ...newFilter,
-                                    tileTargets: undefined,
-                                });
+                                onChange(newFilter);
```
joaoviana, line 249: cool this makes sense - but this makes me think about current Schedule Deliveries that have Filters with `tileTargets` `undefined` -> could we keep using `SchedulerFilter` type but have it equal to `DashboardFilterRule` with its property `tileTargets: undefined | DashboardFilterRule['tileTargets']`? This way we can tackle both already saved ones and future ones. 

To summarise:
 1. Scheduler Override Filters (existing behavior)
  - tileTargets: `undefined`
  - Only override values (operator, values, disabled state)
  - Inherit tile applicability from dashboard
  2. Inherited Dashboard Filters (new feature)
  - Include tileTargets from current dashboard state
  - Full filter definition including tile applicability
  - Used when no override exists

This way we can better identify these 2 types of filters.

packages/frontend/src/features/scheduler/components/SchedulerPreview.tsx
```diff
@@ -14,7 +14,7 @@ import { CUSTOM_WIDTH_OPTIONS } from '../constants';
 
 type Props = {
     dashboard: Dashboard;
-    schedulerFilters: SchedulerFilterRule[] | undefined;
```
joaoviana, line 17: Type suggestion related to another comment somewhere in my review that requested to keep this type for better distinguishing

packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
```diff
@@ -838,6 +838,27 @@ const DashboardProvider: React.FC<
         [dashboardFilters.dimensions, allFilterableFieldsMap],
     );
 
+    // Memoized mapping of tile UUIDs to their display names
+    const tileNamesById = useMemo(() => {
```
joaoviana, line 841: I wonder if we need this at all? 🤔

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
```diff
@@ -302,46 +328,81 @@ const SchedulerFilters: FC<SchedulerFiltersProps> = ({
...
+        ) {
+            onChange(currentDashboardFilters.dimensions);
+        }
+    }, [
+        currentDashboardFilters.dimensions,
+        savedFilters,
+        onChange,
+        draftFilters,
+        isEditMode,
+    ]);
```
joaoviana, line 367: this works great for new schedulers - I like it! I would test with Jake the solution and ask about the necessity of a toggle to inherit filters, rather than them being applied automatically 💪

joaoviana (changes requested):
works great! 

One issue with types + we should keep `tileTargets` `undefined` for overrides so there's a good semantic distinction.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Yea! Make a temp fitler that applies to some tiles. Save a scheduler. Then remove the temp filter from the dashbaord and edit the scheduler. The fitler will be there, and this shows you which tiles it applies to. 

For filters that still exist in the dashboard, you can go look at the checklist but this seemed a little necessary cause this information is otherwise lost. We could add all of the tile info in the shcheduler interface, but I say we do it in the future.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Its an annoyance from the fact that we are re-implementing some of the fitler logic. Null and not null have no values, so we remove the values. All others retain their value. 

In the current filters if you update to null or not null, the values get retained, but never used, so its not a problem. In this change we display them, so I had to remove them.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Discussed in Slack -- we are going to remove this type.

packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
magnew: I use it to show the tile names that a fitler is applied to. Its really convoluted and will likely be useful in the future.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Will do. People are pretty clear that this should be the default, but maybe it can be turned off.

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: I see, you can use `isWithValueFilter(
                                                        filter.operator,
                                                    )`

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
joaoviana: Got it! 

Image

magnew: I see at least one style improvement to make :D

packages/frontend/src/features/scheduler/components/SchedulerFilters.tsx
magnew: Thanks -- I didn't see that

PR #16359, opened 2025-08-12
fix: add frontend SDK to release process

rephus:
Closes: 

image

image

Tested semantic release in Alpha branch https://github.com/lightdash/lightdash/pull/16358

### Description:
Add frontend SDK to the release process by including it in the build and release scripts. The changes include:

1. Adding a comment in the release workflow to indicate that configuration is in release.config.js
2. Updating the build-published-packages script to also build the frontend SDK
3. Adding the frontend package to the release-packages script
4. Creating a release script in the frontend package.json for SDK publishing
5. Including the frontend SDK package.json in the release configuration

These changes ensure that the frontend SDK is properly built and published as part of the release process.

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -19,7 +19,8 @@
         "fix-format": "pnpm run formatter ./src --write",
         "unused-exports": "ts-unused-exports ./tsconfig.json --allowUnusedEnums --allowUnusedTypes --showLineNumber --ignoreFiles='build/.*' --ignoreFiles='sdk/dist/.*' --ignoreFiles='.*/__mocks__/.*' --ignoreFiles='src/stories/.*'",
         "storybook": "storybook dev -p 6006 --no-open",
-        "build-storybook": "storybook build"
+        "build-storybook": "storybook build",
+        "release": "(cd sdk && pnpm publish --no-git-checks)"
     },
     "dependencies": {
         "@casl/ability": "^5.4.4",
--- a/release.config.js
+++ b/release.config.js
@@ -48,6 +48,7 @@ module.exports = {
                     'packages/e2e/package.json',
                     'packages/frontend/package.json',
                     'packages/warehouses/package.json',
+                    'packages/frontend/sdk/package.json',
                 ],
                 message:
                     'chore(release): ${nextRelease.version} \n\n${nextRelease.notes}',
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -31,6 +31,7 @@ jobs:
                   NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
 
             - name: Semantic Release
+              # Config is on release.config.js
               id: semantic
               uses: cycjimmy/semantic-release-action@v4
               env:
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
     "license": "MIT",
     "private": true,
     "workspaces": [
-        "packages/*"
+        "packages/*",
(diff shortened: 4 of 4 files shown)


rephus:
* **#16359** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16359?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16359?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/package.json
```diff
@@ -19,7 +19,9 @@
         "fix-format": "pnpm run formatter ./src --write",
         "unused-exports": "ts-unused-exports ./tsconfig.json --allowUnusedEnums --allowUnusedTypes --showLineNumber --ignoreFiles='build/.*' --ignoreFiles='sdk/dist/.*' --ignoreFiles='.*/__mocks__/.*' --ignoreFiles='src/stories/.*'",
         "storybook": "storybook dev -p 6006 --no-open",
-        "build-storybook": "storybook build"
+        "build-storybook": "storybook build",
+        "update-sdk-version": "sed -i \"s/\\\"version\\\": \\\".*\\\"/\\\"version\\\": \\\"$(grep '\"version\"' package.json | cut -d'\"' -f4)\\\"/\" sdk/package.json",
```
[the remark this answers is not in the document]
rephus: this is supposed to run only in linux

rephus: semantic-release updates automatically all the versions for all packages, but sdk is not a package, so we need to update this manually , by copying the version from the frontend package (already updated)

release.config.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
         [
             '@semantic-release/exec',
             {
-                prepareCmd: 'pnpm build-published-packages',
+                prepareCmd: 'pnpm -F frontend update-sdk-version && pnpm build-published-packages',
```
rephus: Other packages are updated here https://github.com/lightdash/lightdash/pull/16359/files#diff-ac88fd8e5da48b4c325de01dc00bc9609325e9bdb489c5a1abc781d538a1579eR28

packages/frontend/package.json
stevelikesmusic: _Should_ the sdk be its own package?

stevelikesmusic (approved):

packages/frontend/package.json
rephus: maybe, @IrakliJani what do you think ?

IrakliJani: @stevelikesmusic @rephus 
Initially, the SDK was a separate package, but it was difficult to build it outside of the frontend package. That’s why it was moved into the frontend package.

It is still its own package, but now resides under the frontend package because the SDK has its own package.json.

rephus: > It is still its own package

But this is not a core package, right ? because this  `version` for the SDK is not being updated  like the others, that is why I need this "extra" script. And we can't use `pnpm -F sdk build` right ? 

or perhaps I'm missing something else on the package configuration or the release config.

IrakliJani: yeah.. correct.

in the root package.json:
```
"workspaces": [
        "packages/*"
    ],
```

I wonder, if you add sdk path here and update `assets` in `release.config.js`

it might work with release script directly without the need for custom version bumper. worth trying

rephus: I think that worked, thanks  ! 
image

PR #16361, opened 2025-08-12
fix: re-rendering and re-execution of dashboard chart on initial load

almeidabbm:
Closes:

### DIFF
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -1940,6 +1940,22 @@ export class AsyncQueryService extends ProjectService {
         };
     }
 
+    private async getDashboardParameters(
+        dashboardUuid: string,
+    ): Promise<ParametersValuesMap | undefined> {
+        const dashboard = await this.dashboardModel.getById(dashboardUuid);
+        const { parameters: rawDashboardParameters } = dashboard;
+
+        // Convert dashboard parameters to ParametersValuesMap format
+        return rawDashboardParameters
+            ? Object.fromEntries(
+                  Object.entries(rawDashboardParameters).map(
+                      ([key, dashboardParam]) => [key, dashboardParam.value],
+                  ),
+              )
+            : undefined;
+    }
+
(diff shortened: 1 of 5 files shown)


almeidabbm:
* **#16361** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16361?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16361?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -2782,10 +2788,16 @@ export class AsyncQueryService extends ProjectService {
...
 
+        const dashboardParameters = await this.getDashboardParameters(
+            dashboardUuid,
+        );
+
         // Combine default parameter values with request parameters first
         const combinedParameters = await this.combineParameters(
             projectUuid,
             args.parameters,
+            undefined, // Explicitly empty saved chart parameters, only dashboard parameters are used
```
magnew, line 2799: This is different, but it makes sense.

packages/frontend/src/features/parameters/components/Parameters.tsx
```diff
@@ -62,23 +64,6 @@ export const Parameters: FC<Props> = ({
     // Calculate selected parameters count
     const selectedParametersCount = Object.values(parameters ?? {}).length;
 
-    // Apply defaults
-    useEffect(() => {
```
magnew, line 66: Getting rid of this is key, I think.

almeidabbm: Yep yep, exactly, this was what was taking in the defaults and setting them as if they were manually set by the user thus triggering another query.

Removing this behavior meant that now `dashboards` could sent `undefined` parameters to let the backend handle the default. Which is why we had to change/remove the `saved_chart` parameters from the combined ones. Otherwise charts could be rendered with different parameters. But also, now the behavior is consistent and predictable 🙌

Issue #16297, opened 2025-08-08
define parameters at model level


ZeRego:
### Description

Ideally parameters can be defined in either in the lightdash.config.yml or at the model level

Note: how to avoid name conflicts? should we throw an error if two models use the same param name? 

https://lightdash.slack.com/archives/C08TPMFGGTX/p1754583212805709?thread_ts=1753346630.351489&cid=C08TPMFGGTX

### What problem does this solve?

Not having a massive lightdash.config.yml file.

ps: this partially solved the problem for dbt cloud users since the receive the model meta from dbt cloud.

PR #16378, opened 2025-08-12
feat: model level parameters

almeidabbm:
Closes: #16297

### Description:
This PR adds support for model-level parameters in dbt models, allowing parameters to be defined directly in the model YAML files. These parameters work alongside project-level parameters defined in `lightdash.config.yml`.

Key changes:
- Added support for defining parameters in model metadata
- Added validation to prevent duplicate parameter names between model and project
- Updated the frontend to display and use parameters from both sources
- Modified the SQL join condition in the example to demonstrate parameter usage

The implementation ensures that model-specific parameters are properly loaded and displayed in the UI, while maintaining backward compatibility with existing project-level parameters.

**Model parameter with same name as project parameter error**

**Model parameter**

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.ts
+++ b/packages/common/src/compiler/exploreCompiler.ts
@@ -86,6 +86,7 @@ export type UncompiledExplore = {
     aiHint?: string | string[];
     meta: DbtRawModelNode['meta'];
     databricksCompute?: string;
+    projectParameters?: LightdashProjectConfig['parameters'];
 };
 
 const getReferencedTable = (
@@ -123,6 +124,7 @@ export class ExploreCompiler {
         meta,
         databricksCompute,
         aiHint,
+        projectParameters,
     }: UncompiledExplore): Explore {
         // Check that base table and joined tables exist
         if (!tables[baseTable]) {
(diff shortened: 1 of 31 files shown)


almeidabbm:
* **#16378** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16378?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16378?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/features/scheduler/components/SchedulerParameters.tsx
```diff
@@ -230,7 +217,8 @@ const SchedulerParameters: FC<SchedulerParametersProps> = ({
             <MantineProvider theme={getMantine8ThemeOverride()}>
                 {Object.entries(availableParameters).map(
                     ([paramKey, parameter]) => {
```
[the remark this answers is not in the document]
almeidabbm: It wont, there's an IF block before it:

```typescript
    if (!availableParameters || Object.keys(availableParameters).length === 0) {
        return (
            <Center component={Stack} h={100}>
                <Text color="dimmed">
                    No parameters defined for this project.
                </Text>
            </Center>
        );
    }
```

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -149,6 +151,22 @@ export class ExploreCompiler {
...
+        const hasDuplicateParameterDefinitions = Object.keys(
+            projectParameters || {},
+        ).some((projectParamName) =>
+            Object.keys(meta.parameters || {}).some(
+                (modelParamName) => modelParamName === projectParamName,
+            ),
+        );
+
+        if (hasDuplicateParameterDefinitions) {
+            throw new CompileError(
```
pimmey, line 164: Just a Q: how comes we wanna throw an error here, instead of maybe overriding?

almeidabbm: Having an error for name duplicates brings it to the attention of the developer that there is in fact a duplicate instead of silently overriding which could lead to confusing behavior

pimmey (approved):
🛳️🛳️🛳️

Issue #16371, opened 2025-08-12
Frontend SQL viewer crashes with parameters


jesshitchcock:
### Description

The Fontend SQL Viewer is crashing when I try to view the SQL in the explore, even though I am able to run the query and get results. 
This seems to only be happening when a parameter is included in the SQL. 

<img width="1103" height="564" alt="Image" src="https://github.com/user-attachments/assets/381bdac7-aded-4377-bc9e-910f21be677c" />

This [Loom video](https://www.loom.com/share/379d26b48e134ff882e84e30da7cb850) demonstrates the issue. 

### Steps to Reproduce the Bug or Issue

1. Define parameter in lightdash.config.yml
``` yaml
parameters:
  date_format:
    label: "Date"
    description: "Date format"
    options:
      - "American Date"
      - "Standard Date"
    default: ["American Date"]
    multiple: false
  date_range:
    label: "Date Range"
    description: "Filter data by date range"
    options: 
      - "2025-01-01"
    default: ["2025-01-01"]
```
2. Define an additional dimension for a column in a model.yml file. 
``` yaml 
      - name: created_date
        description: ""
        meta:
          dimension:
            type: date
          additional_dimensions:
            after_2025:
              type: boolean
              sql: ${TABLE}.created_date >= CAST(${lightdash.parameters.date_range} AS TIMESTAMP)
```
3. Run `lightdash preview`
4. Open the explorer and select the dimension that has the parameter applied
5. Hit `Run Query` to test that the SQL runs and provides results. 
6. Try to open the SQL at the bottom of the explore, you should see the error message. 

### Cloud or self-hosting

None

PR #16379, opened 2025-08-13
fix: catch error in sql formatter

magnew:
Closes: #16371 

### Description:

Formatting fails with invalid (not normal) SQL. One way to cause this is to have a Parameter in your query. We are also seeing it in user reports. This just catches the error on formatting and returns the original error. 

To test
- Select Subscriptions model
- Select 'Filtered Subscriptions'
- Open SQL card
- On main it will error

Screenshot 2025-08-13 at 07 59 58

### DIFF
--- a/packages/frontend/src/components/RenderedSql.tsx
+++ b/packages/frontend/src/components/RenderedSql.tsx
@@ -50,7 +50,7 @@ export const RenderedSql = () => {
                 language: getLanguage(project?.warehouseConnection?.type),
             });
         } catch (e) {
-            console.error(e);
+            console.error('Error rendering SQL:', e.message || 'Unknown error occurred');
             return data.query;
         }
     }, [data?.query, project?.warehouseConnection?.type]);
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/RenderedSql.tsx
```diff
@@ -43,6 +43,18 @@ export const RenderedSql = () => {
...
     );
 
+    const formattedSql = useMemo(() => {
+        if (!data?.query) return '';
+        try {
+            return format(data.query, {
+                language: getLanguage(project?.warehouseConnection?.type),
+            });
+        } catch (e) {
+            console.error('Error rendering SQL:', e.message || 'Unknown error occurred');
```
joaoviana: wonder if it needs its own error instance logged in case we need to tackle some cases but I'm happy with the change and we do the same in `sqlRunnerSlice.ts`

joaoviana (approved):

almeidabbm (approved):
LGTM 🚀 

**Before**

Image

**After**

Image

Issue #16347, opened 2025-08-11
Create Project roles API


stevelikesmusic:
Now that we have the DB and business logic in place. Create a controller for roles. We'll also want testing at the e2e API level and possibly at the service level. 

These would be nested under `/api/v2` namespace:

## API

  - `GET /projects/:projectId/access` - Get access and given roles
  - `PATCH /groups/:groupId/projects/:projectId` - Assign group role to project
  - `DELETE /groups/:groupId/projects/:projectId` - Remove group role from project
  - `PATCH /projects/:projectId/access/:accessId` - Assign existing user access to new role
  - `DELETE /projects/:projectId/access/:accessId` - Remove existing user access from project
  - `POST /projects/:projectId/access` - Create user assignment to project

PR #16323, opened 2025-08-08
feat: add roles models and services

stevelikesmusic:
Closes: #16282

### Description:
Adds models and service for role-based access control:

- Created `RolesModel`, `ScopesModel`, and `ScopedRolesModel` to manage roles and permissions
- Implemented `RolesService` with methods for creating, updating, and deleting roles
- Added functionality to assign and remove scopes (permissions) from roles

This implementation provides the foundation the API in the next PR

[Context](https://www.notion.so/lightdash/Custom-roles-permission-233a63207a7a8027a6d4d68a4b210045)

PR #16380, opened 2025-08-13
feat: add custom roles and permissions management

rephus:
Closes: #1234

### Description:
This PR adds a new Roles API to manage custom roles and permissions in Lightdash. The implementation includes:

- New RolesController with endpoints for creating, updating, and deleting roles
- RolesModel for database interactions related to roles management
- RolesService to handle business logic and authorization
- Support for assigning roles to users and groups
- Project access management through roles
- Role scopes management for fine-grained permissions
- E2E tests for the new API endpoints
- Disables CSP report-only mode by setting it to false

The PR also updates database entities to support role-based access control by adding role_uuid fields to organization memberships, project memberships, and project group access tables.


rephus:
* **#16380** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16380?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16380?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16319** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16319?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#16323](https://github.com/lightdash/lightdash/pull/16323) <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16323?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/backend/src/models/RolesModel.ts
```diff
@@ -0,0 +1,452 @@
...
+
+            const existingAccess = await this.database('project_group_access')
+                .where('group_uuid', groupUuid)
+                .where('project_id', project.project_id)
+                .first();
+
+            throw new NotImplementedError('Group access is not implemented');
+            /* 
+
+            if (existingAccess) {
```
rephus, line 284: TODO

packages/backend/src/models/UserModel.ts
```diff
@@ -487,7 +487,7 @@ export class UserModel {
 
         return projectMemberships.map((membership) => ({
             projectUuid: membership.project_uuid,
-            role: membership.role,
+            role: membership.role || ProjectMemberRole.VIEWER,
```
rephus, line 490: since db role can be now null, but this type can't be

packages/backend/src/services/RolesService.ts
```diff
@@ -0,0 +1,462 @@
...
+        this.analytics = analytics;
+        this.rolesModel = rolesModel;
+        this.userModel = userModel;
+        this.organizationModel = organizationModel;
+        this.groupsModel = groupsModel;
+        this.projectModel = projectModel;
+    }
+
+    async getRolesByOrganizationUuid(
+        user: SessionUser,
```
rephus: should we be using accounst ?

packages/backend/src/database/entities/organizationMemberships.ts
```diff
@@ -8,15 +8,23 @@ export type DbOrganizationMembership = {
     user_id: number;
     created_at: Date;
     role: OrganizationMemberRole;
+    role_uuid?: string | null;
```
rephus, line 11: @stevelikesmusic I created this branch from your migration branch, not the mode.s, it is possible some of this code is already duplicated.

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+import { BaseController } from '../baseController';
+
+@Route('/api/v2/roles')
+@Response<ApiErrorPayload>('default', 'Error')
+@Tags('v2', 'Roles')
+export class RolesController extends BaseController {
+    /**
+     * Get roles for organization
+     */
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
```
stevelikesmusic: @rephus is it worth putting the middleware at the class level for this controller? I would expect auth to be the same for all routes.

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+    @OperationId('GetRolesByOrganization')
+    async getRolesByOrganization(
+        @Request() req: express.Request,
+        @Path() orgUuid: string,
+        @Query() load?: string,
+    ): Promise<ApiGetRolesResponse | ApiRoleWithScopesResponse> {
+        const loadScopes = load === 'scopes';
+        const roles = await this.services
+            .getRolesService()
+            .getRolesByOrganizationUuid(req.user!, orgUuid, loadScopes);
```
stevelikesmusic: Could these pass in `req.account` rather than `req.user`

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+        };
+    }
+
+    /**
+     * Create a new role
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
```
stevelikesmusic: How do we determine what is/isn't allowed in demo-mode?

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+    /**
+     * Update role
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('200', 'Success')
+    @Patch('/{roleUuid}')
```
stevelikesmusic: What would you think of namespacing roles under `/org/:orgId` so we can verify that a role from one org isn't changed by someone in another org?

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+    /**
+     * Assign role to group
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('200', 'Success')
+    @Patch('/groups/{groupId}/role/{roleId}')
```
stevelikesmusic: Should groups also be namespaced under org like `users`?

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+            results: undefined,
+        };
+    }
+
+    /**
+     * Get project access
+     */
+    @Middlewares([allowApiKeyAuthentication, isAuthenticated])
+    @SuccessResponse('200', 'Success')
+    @Get('/projects/{projectId}/access')
```
stevelikesmusic: Project access is the superset of user and group roles?

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+    @OperationId('GetProjectAccess')
+    async getProjectAccess(
+        @Request() req: express.Request,
+        @Path() projectId: string,
+    ): Promise<{
+        status: 'ok';
+        results: {
+            users: unknown[];
+            groups: unknown[];
```
stevelikesmusic: Why are these types `unknown`—just placeholders?

packages/backend/src/controllers/v2/RolesController.ts
```diff
@@ -0,0 +1,427 @@
...
+    /**
+     * Update user project access
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
+    ])
+    @SuccessResponse('200', 'Success')
+    @Patch('/projects/{projectId}/access/{accessId}')
```
stevelikesmusic: Two questions:

1. We namespace groups in the API, but not users. Any reason not to use the same pattern?
2. What does `accessId` refer to? From what I could tell the project access/membership tables use compound keys.

packages/backend/src/models/RolesModel.ts
```diff
@@ -0,0 +1,471 @@
...
+
+    constructor(database: Knex) {
+        this.database = database;
+    }
+
+    get db(): Knex {
+        return this.database;
+    }
+
+    private async getOrganizationId(organizationUuid: string): Promise<number> {
```
stevelikesmusic, line 38: 😭

packages/backend/src/models/RolesModel.ts
```diff
@@ -0,0 +1,471 @@
...
+                `${RolesTableName}.*`,
+                this.database.raw(
+                    `STRING_AGG(${ScopedRolesTableName}.scope_name, ',') as scopes`,
+                ),
+            )
+            .where(`${RolesTableName}.organization_uuid`, organizationUuid)
+            .orWhere(`${RolesTableName}.type`, 'system')
+            .groupBy(`${RolesTableName}.role_uuid`);
+
+        const roles = await query;
```
stevelikesmusic: Why create the `query` var instead of just awaiting it into `roles`?

packages/backend/src/models/RolesModel.ts
```diff
@@ -0,0 +1,471 @@
...
+
+    async assignRoleToUser(
+        userUuid: string,
+        roleUuid: string,
+        organizationUuid?: string,
+        projectUuid?: string,
+    ): Promise<void> {
+        const userId = await this.getUserId(userUuid);
+
+        if (organizationUuid) {
```
stevelikesmusic, line 208: Should we either return early if we have an org? Or make the types mutually exclusive between org and project.

packages/backend/src/models/UserModel.ts
stevelikesmusic: I updated the migration so that `role` is not `nullable` after all. We just use the default while we're transitioning.

packages/backend/src/services/RolesService.ts
stevelikesmusic: Absolutely!

packages/backend/src/services/RolesService.ts
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    private static validateOrganizationAccess(
+        user: SessionUser,
+        organizationUuid: string,
+    ): void {
+        if (
+            !user.organizationUuid ||
+            user.organizationUuid !== organizationUuid
+        ) {
```
stevelikesmusic: Could this be simpler with optional chaining?

```ts
user?.organizationUuid !== organizationUuid
```

packages/backend/src/services/RolesService.ts
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    private static validateRoleName(name: string): void {
+        if (!name) {
+            throw new ParameterError('Role name is required');
+        }
+        if (name.trim().length === 0) {
+            throw new ParameterError('Role name cannot be empty');
+        }
+        if (name.length > 255) {
```
stevelikesmusic: Please someone some day trigger this one 😂

packages/backend/src/services/RolesService.ts
```diff
@@ -0,0 +1,429 @@
...
+            throw new ParameterError(
+                'Role name must be 255 characters or less',
+            );
+        }
+    }
+
+    async getRolesByOrganizationUuid(
+        user: SessionUser,
+        organizationUuid: string,
+        loadScopes?: boolean,
```
stevelikesmusic: Very nice!

packages/e2e/cypress/e2e/api/roles.cy.ts
```diff
@@ -0,0 +1,458 @@
...
+                expect(resp.body).to.have.property('status', 'ok');
+                expect(resp.body.results).to.have.property('users');
+                expect(resp.body.results).to.have.property('groups');
+                expect(resp.body.results.users).to.be.an('array');
+                expect(resp.body.results.groups).to.be.an('array');
+            });
+        });
+
+        it('should create user project access', () => {
+            let testUserUuid: string;
```
stevelikesmusic: Using [aliases](https://docs.cypress.io/app/core-concepts/variables-and-aliases#Aliases) we wouldn't need to keep `testUserUuid` state like this—I remember this way can be somewhat flakey in Cypress. 

Aliases would also help to make the results of requests available later on without having to nest Cypress `.then` calls.

stevelikesmusic (commented):
Nice work Javi 👏

packages/backend/src/controllers/v2/RolesController.ts
rephus: normally we only allow view actions on demo

packages/e2e/cypress/e2e/api/roles.cy.ts
rephus: I didn't know that ! cool, I'll change it

packages/backend/src/controllers/v2/RolesController.ts
rephus: fixed

packages/backend/src/controllers/v2/RolesController.ts
rephus: fixed

packages/backend/src/controllers/v2/RolesController.ts
rephus: good question, I don't know, I'm used to specify this on each route :thinking:

packages/backend/src/controllers/v2/RolesController.ts
rephus: to do this, I'd not require the orguuid in the path, but instead, get the org for that role, and check if the user has the right casl abilities. I haven't implemented permissions, let me take a look now

packages/backend/src/controllers/v2/RolesController.ts
rephus: I consider groups their own namespace, I don't think we need to be that explicit. What do you think ?

packages/backend/src/controllers/v2/RolesController.ts
rephus: this endpoint returns both 

        return {
            users: userAccess,
            groups: groupAccess,
        };

packages/backend/src/controllers/v2/RolesController.ts
rephus: yeah , so to be fair, this was claude, he built the endpoinst following the ticket spec https://github.com/lightdash/lightdash/issues/16347

> What does accessId refer to

accessId is the userUuid in the service, I'll change that

packages/backend/src/models/RolesModel.ts
rephus: You might want to unnasign a role both from a org and a project. I don't think we'll be using both at the same time in the UI though

packages/backend/src/controllers/v2/RolesController.ts
rephus: I implemented casl permission checks  now, and added some integration tests for them too

rephus:
moving RoleService into EE

rephus:
I'm building a roles.http to manually call these endpoints an follow the user journey

packages/backend/src/models/RolesModel.ts
```diff
@@ -0,0 +1,457 @@
...
+            const organizationId = await this.getOrganizationId(
+                organizationUuid,
+            );
+            await this.database(OrganizationMembershipsTableName)
+                .where('user_id', userId)
+                .where('organization_id', organizationId)
+                .update({ role_uuid: null });
```
rephus, line 251: by updating this to null, it will default to the old role system.

stevelikesmusic:
If the controller/service is in EE, will it load without a license? Would would OSS deploys get and set roles?

examples/api/roles.http
```diff
@@ -0,0 +1,190 @@
+
+### Login
```
stevelikesmusic, line 2: I love these

packages/backend/src/ee/controllers/OrganizationRolesController.ts
```diff
@@ -0,0 +1,289 @@
...
+ * - Managing role assignments within an organization
+ *
+ * /api/v2/orgs/{orgId}/roles/assignments
+ * - Listing role assignments for an organization
+ * - Creating role assignments for an organization
+ * - Managing role assignments within an organization
+ *
+ * /api/v2/orgs/{orgId}/roles/assignments/user/{userId}
+ * - Listing role assignments for a user
+ */
```
stevelikesmusic, line 54: Do we want these comments to render with the Swagger docs?

packages/backend/src/ee/controllers/OrganizationRolesController.ts
```diff
@@ -0,0 +1,289 @@
...
+        };
+    }
+
+    /**
+     * Delete role from organization
+     */
+    @Middlewares([
+        allowApiKeyAuthentication,
+        isAuthenticated,
+        unauthorisedInDemo,
```
stevelikesmusic, line 158: You may have answered this in another version comment, but I can't find it 😅 Why mark this one as `unauthroisedInDemo`?

rephus:
no, it will not load, so you can't set or get custom roles. For backwards compatibility, I'm not sure if we are going to build some new endpoints on OSS or add some extra logic to the existing user service.

packages/backend/src/ee/services/RolesService.ts
```diff
@@ -0,0 +1,766 @@
...
+        orgUuid: string,
+    ): Promise<RoleAssignment[]> {
+        RolesService.validateOrganizationAccess(account, orgUuid);
+
+        // Get organization role assignments from model
+        const userAssignments =
+            await this.rolesModel.getOrganizationRoleAssignments(orgUuid);
+
+        // Format user assignments
+        const formattedUserAssignments: RoleAssignment[] = userAssignments.map(
```
stevelikesmusic, line 243: What's the purpose of this? Don't we map the object we get back from the database in the model?

packages/backend/src/ee/services/RolesService.ts
```diff
@@ -0,0 +1,766 @@
...
+        if (loadScopes) {
+            return this.rolesModel.getRolesWithScopesByOrganizationUuid(
+                organizationUuid,
+            );
+        }
+
+        return this.rolesModel.getRolesByOrganizationUuid(organizationUuid);
+    }
+
+    async createRole(
```
stevelikesmusic, line 161: One question I have now that I missed in the initial design: should we allow passing scopes to the create role method? Do we want a UI where admins create a role and add scopes in one step? Or do we want to create the role first and then apply scopes in a different step? Same question for updating.

packages/common/src/types/roles.ts
```diff
@@ -50,3 +65,45 @@ export type ApiDeleteRoleResponse = ApiSuccessEmpty;
...
+export type ApiGetProjectAccessResponse = {
+    status: 'ok';
+    results: {
+        users: ProjectAccess[];
+        groups: GroupProjectAccess[];
+    };
+};
+
+// Unified Assignment Types
+export type RoleAssignment = {
```
stevelikesmusic, line 78: Is it worth having two separate types? They're used quite distinctly and it might make the service methods a little cleaner. 

At the least, if we have a unified type, we should make project and org IDs mutually exclusive so that we couldn't have a `RoleAssignment` where both are defined.

packages/backend/src/ee/services/RolesService.ts
```diff
@@ -0,0 +1,766 @@
...
+            assignments.push({
+                roleId: userAccess.role, // This might be role name for legacy, need to handle
+                roleName: userAccess.role,
+                assigneeType: 'user',
+                assigneeId: userAccess.userUuid,
+                assigneeName: `${userAccess.firstName} ${userAccess.lastName}`,
+                projectId: userAccess.projectUuid,
+                createdAt: new Date(), // TODO: Get actual dates from DB
+                updatedAt: new Date(),
+            });
```
stevelikesmusic, line 393: Does the current UI have a unified definition of assignments? What's the benefit to having a unified approach if the UI is still separated by user and group?

packages/backend/src/ee/services/RolesService.ts
```diff
@@ -0,0 +1,766 @@
...
+                projectId: groupAccess.projectUuid,
+                createdAt: new Date(), // TODO: Get actual dates from DB
+                updatedAt: new Date(),
+            });
+        }
+
+        return assignments;
+    }
+
+    async createProjectRoleAssignment(
```
stevelikesmusic, line 413: Why not have separate methods for group and user roles?

packages/backend/src/ee/services/RolesService.ts
```diff
@@ -0,0 +1,766 @@
...
+        account: Account,
+        projectId: string,
+        request: CreateRoleAssignmentRequest,
+    ): Promise<RoleAssignment> {
+        const { roleId, assigneeType, assigneeId } = request;
+
+        // Validate role ownership and get role details
```
stevelikesmusic, line 420: What's our typical criteria for tracking analytics usage?

packages/backend/src/ee/services/RolesService.ts
```diff
@@ -0,0 +1,766 @@
...
+        const formattedUserAssignments: RoleAssignment[] = userAssignments.map(
+            (assignment) => ({
+                roleId: assignment.roleId,
+                roleName: assignment.roleName || 'Unknown Role',
+                assigneeType: 'user' as const,
+                assigneeId: assignment.assigneeId,
+                assigneeName: assignment.assigneeName,
+                organizationId: assignment.organizationId,
+                createdAt: assignment.createdAt,
+                updatedAt: assignment.createdAt, // Use createdAt since updatedAt doesn't exist
```
stevelikesmusic, line 252: Which table is missing `updatedAt`? If we need that field, a migration will be better than re-using `createdAt`.

stevelikesmusic:
Okteto seems to be having a hard time lately 😩

rephus:
yeah, I've been running these e2e tests locally, they all pass, and I'm only introducing new endpoints, not updating existing ones, so I don't expect anything to fail. You can checkou this branch if you want to test it locally

stevelikesmusic (approved):

Issue #16411, opened 2025-08-14
Explore from here on a table chart causes infinite load


magnew:
### Description

Explore from here on a table chart causes infinite load

### Steps to Reproduce the Bug or Issue

Add a table chart to a dashboard
Explore from here
Table will never finish loading

 

### Cloud or self-hosting

None

PR #16417, opened 2025-08-14
fix: table stuck in loading state on explore from here

magnew:
Closes: #16411

### Description:

Clicking explore from here on a table in a dashboard was showing an infinite loading state in the explore. 

This was partly due to the infinitely loading logic, which made the loader logic different for tables than other charts, and partly do to the new logic about when to auto-load queries. The current behavior for *all* charts is to not auto-run the query on explore from here. But most charts show a no data message. 

This just makes the table chart also show a no-data message instead of an incorrect loading message. 

To test: 
- explore from here from a table on a dashboard
- the viz should show a no-data message until you manually run the query

### DIFF
--- a/packages/frontend/src/components/SimpleTable/index.tsx
+++ b/packages/frontend/src/components/SimpleTable/index.tsx
@@ -34,8 +34,13 @@ const SimpleTable: FC<SimpleTableProps> = ({
     minimal = false,
     ...rest
 }) => {
-    const { columnOrder, itemsMap, visualizationConfig, resultsData } =
-        useVisualizationContext();
+    const {
+        columnOrder,
+        itemsMap,
+        visualizationConfig,
+        resultsData,
+        isLoading,
+    } = useVisualizationContext();
 
     const shouldPaginateResults = useMemo(() => {
         return Boolean(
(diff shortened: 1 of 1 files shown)


magnew:
* **#16417** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16417?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

almeidabbm (approved):
This fixes the infinite loading issue, but for some reason the state says there's no data:

Image

If we try and edit other charts this is the status:

Image

magnew:
Thanks @almeidabbm! Fixed to differentiate between 'idle' state and no data. 

I used the default idle state in the explore -- we probably could make it look better, but I think this at least fixes the issue right. 

Screenshot 2025-08-18 at 10 16 23

almeidabbm:
> Thanks @almeidabbm! Fixed to differentiate between 'idle' state and no data.
> 
> I used the default idle state in the explore -- we probably could make it look better, but I think this at least fixes the issue right.
> 
> Screenshot 2025-08-18 at 10 16 23

Agreed, this works well to fix this issue. We might want find out at some point why `<EmptyChart/>` is not rendering in this specific case. 🚀

magnew:
> > Thanks @almeidabbm! Fixed to differentiate between 'idle' state and no data.
> > I used the default idle state in the explore -- we probably could make it look better, but I think this at least fixes the issue right.
> > Screenshot 2025-08-18 at 10 16 23
> 
> Agreed, this works well to fix this issue. We might want find out at some point why `<EmptyChart/>` is not rendering in this specific case. 🚀

Its basically cause that's defined in SimpleChart (which is everything echarts) and this is SimpleTable. We could make them look more similar.

Issue #16351, opened 2025-08-11
Hydrate scopes from role for session users


stevelikesmusic:
When a session request is handled, we load the [session user via the UserModel](https://github.com/lightdash/lightdash/blob/9ed2fd73a43adea74ff7f292e86fe1f30b0ab1eb/packages/backend/src/models/UserModel.ts#L733-L775). That model also loads the associated roles assigned to that user. Using the environment variable config `isCustomRolesEnabled`, we should delegate behavior in the UserModel to fetch roles from the current tables, or get the role and scopes for the user in the custom roles model being implemented. 

## Acceptance

- [ ] `findSessionUserAndOrgByUuid` delegates between current role lookup and custom role lookup via config
- [ ] If loading custom roles, we apply CASL permissions with the `scopeAbilityBuilder`. 

## Question

Is it worth adding [TestContainers](https://node.testcontainers.org/quickstart/usage/) to test models directly rather than implicitly via e2e API tests?

PR #16456, opened 2025-08-18
fix: apply custom role abilities to user

rephus:
Closes: #16351

- [x] load role_uuid and scopes into ability builder
- [x] Apply abilities for scopes
- [x] apply if env variable is enabled
- [x] Fix tests to ensure parity with old named roles

### Description:
Added support for custom roles by introducing a `roleUuid` field to user and member profiles. This change enables the system to identify and apply custom role-based permissions through the authorization system.

The implementation includes:
- Added `roleUuid` field to `OrganizationMemberProfile` and `ProjectMemberProfile` types
- Updated database queries to fetch role UUIDs from memberships
- Enhanced the user ability builder to handle custom role scopes
- Added functionality to fetch and apply custom role permissions

This lays the groundwork for the custom roles feature, allowing for more granular permission control beyond the standard role types.


rephus:
* **#16456** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16456?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16456?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/UserModel.ts
```diff
@@ -519,9 +535,75 @@ export class UserModel {
...
+            if (!scopesRecord[roleUuid]) {
+                scopesRecord[roleUuid] = [];
+            }
+            scopesRecord[roleUuid].push(scopeName);
+        });
+
+        return scopesRecord;
+    }
+
+    private async generateUserAbilityBuilder(user: DbUserDetails): Promise<{
```
rephus, line 568: extracted all the role builder logic into a single private method, it is reused in many places in this model

packages/common/src/authorization/index.ts
```diff
@@ -27,35 +33,60 @@ export const getUserAbilityBuilder = ({
...
-                userUuid: user.userUuid,
-            },
-            builder,
-            permissionsConfig,
-        });
-        projectProfiles.forEach((projectProfile) => {
-            projectMemberAbilities[projectProfile.role](
-                projectProfile,
+        if (user.roleUuid) {
+            // TODO apply custom role abilities
```
rephus: @stevelikesmusic this is where we should load your ability builder for scopes

rephus: do we have a single method for scopes ? or do we still have to load org and projects separately

packages/common/src/authorization/index.ts
```diff
@@ -27,35 +33,60 @@ export const getUserAbilityBuilder = ({
...
+        }
+        projectProfiles.forEach((projectProfile) => {
+            if (projectProfile.roleUuid) {
+                const scopes = customRoleScopes?.[projectProfile.roleUuid];
+                if (scopes) {
+                    throw new NotFoundError(
+                        `Custom role with uuid ${user.roleUuid} was not found`,
+                    );
+                }
+                // TODO apply custom role abilities
```
rephus: @stevelikesmusic I think the new `buildAbilityFromScopes` is not equivalent to our existing builder configuration, or I don't know how to apply this correctly. 

on the existing implementation. First we create a builder, then we apply organization roles using the org context (https://github.com/lightdash/lightdash/pull/16456/files#diff-9b14bc11988e1adf91ede01e83a58e5a9ef468b1ba744a3f0cdabc63f5bb84dcR49), then, we apply the abilities to the same builder. 

This new `buildAbilityFromScopes` takes a single org and project, and returns a build. So we can't apply org permissions separately, or return abilities for different projects. 

For example: A user is a  `viewer` in the org, and an admin in project-1, and developer in project-2. 

Even if we build it on the parent method (userModel), I think we are missing something. What do you think ? 

On solution that might work is: 
- export applyScopeAbilities : so we can use it separately, with a single build
- `ScopeContext` in applyScopeAbilities has "projectUuid" optional: for building org abilities
this way the API will be similar to our existing org/project builder

rephus: Also, a reason why we need to implement the buildAbilityFromScopes inside this method, is because, until we migrate everything, old role names and new custom roles will coexist for a while. So we need t oapply this conditionally if role_uuid exists, so we need to use the same builder. 

example: an user has an old `admin` role on the org, but a new  custom role  in project-1

stevelikesmusic: You're right. I updated the API to receive a builder rather than create a new one itself.

packages/common/src/authorization/index.ts
stevelikesmusic: Good point! I'll create a new method in a separate PR to be the entry similar to getUserAbilityBuilder. Similar API in that it would receive projectProfiles and apply both org and project permissions? The only difference is that we'd have a projectUuid instead of an orgUuid.

rephus:
unrelated e2e test

image

stevelikesmusic (approved):

Issue #16284, opened 2025-08-07
Add `scopeAbilityBuilder` for applying scopes to CASL


stevelikesmusic:
Create a new ability builder for scopes: `scopeAbilityBuilder`. The builder will take a list of scope names and context. We will use those scopes to apply CASL abilities for the user and add context to the CASL subject similar to what we do now in builders like organization and projects. Use the current abilityBuilders to understand the what all to include in the context object. 

Scopes will be passed in as a single name that encapsulate a CASL action/subject pair: `view:dashboard` that will need to be parsed into something like `{action: 'view', subject: 'Dashboard'}`. 

We should have test coverage to ensure:
1. Scope names are properly parsed in to abilities
2. Context is properly applied for things like: org, project, and space access

PR #16460, opened 2025-08-18
feat: add scopeAbilityBuilder for scoped roles

stevelikesmusic:
Closes: #16284

### Description:
Implements a CASL-based permission system for fine-grained access control using scopes. 

One caveat is `PersonalAccessToken` which is currently config-based rather than role based. I kept the config-based approach for now, but open to adding it as a scope, since that would be a better UX for admins anyway. 

The abilities were copied over from org + project ability builders. 

### AI Description
- A comprehensive scope definition system with detailed conditions for each permission
- A scope ability builder that converts scopes into CASL abilities
- Extensive test coverage for all permission scenarios

The implementation supports organization-level, project-level, and user-specific permissions with proper access control for private resources. It handles different user roles (viewer, editor, admin) and includes enterprise-specific permissions.


stevelikesmusic:
* **#16460** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/authorization/scopeAbilityBuilder.test.ts
```diff
@@ -0,0 +1,1474 @@
...
+            isEnterprise: false,
+            organizationRole: 'admin',
+            projectUuid: 'project-123',
+            scopes: new Set(),
+        };
+
+        it('should build ability with organization view permissions', () => {
+            const ability = buildAbilityFromScopes({
+                ...baseContext,
+                scopes: new Set(['view:Organization']),
```
rephus: is casing relevant ?

packages/common/src/authorization/scopeAbilityBuilder.ts
```diff
@@ -0,0 +1,75 @@
...
+} from './types';
+
+/**
+ * Parse a scope name into action and CASL subject
+ * @param scope - The scope name (e.g., 'view:dashboard', 'manage:saved_chart')
+ * @returns Object with action and subject properties
+ */
+const parseScope = (scope: ScopeName) => {
+    const [action, subjectPart] = scope.split(':');
+    const subject = upperFirst(camelCase(subjectPart));
```
rephus: why upperFirst ?

packages/common/src/authorization/scopeAbilityBuilder.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+    });
+
+    // TODO:Do we need this if admins can manage more easily in the settings UI now?
+    const { pat } = context?.permissionsConfig || {};
+    if (
+        pat?.enabled &&
+        pat?.allowedOrgRoles.includes(context.organizationRole)
+    ) {
+        builder.can('manage', 'PersonalAccessToken');
```
rephus: what's this for ? this is not being returned, or evaluated.

packages/common/src/authorization/scopeAbilityBuilder.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
```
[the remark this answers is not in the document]
rephus: should we apply this suggestion ?

packages/common/src/authorization/scopeAbilityBuilder.test.ts
stevelikesmusic: Yes and no. I initially thought to keep everything snake_case: `export:dashboard_csv`. Thinking more, I thought that changing the casing from PascalCase (CaslSubjects) to snake_case would make it harder to search and harder for an LLM to parse search through. No very strong opinions, though.

packages/common/src/authorization/scopes.ts
```diff
@@ -0,0 +1,873 @@
...
+            // Public dashboards
+            const conditions: Record<string, unknown>[] = [
+                {
+                    organizationUuid: context.organizationUuid,
+                    isPrivate: false,
+                },
+            ];
+
+            // User's accessible dashboards via space access
+            if (context.userUuid) {
```
rephus: if userUuid is not defined, should we throw an error ? or is it because thsi could be happening on a service account.. is there a way to enforce this context when using this ?

packages/common/src/authorization/scopeAbilityBuilder.ts
stevelikesmusic: The idea was to match our CASL subject types which are all PascalCase.

packages/common/src/authorization/scopes.ts
```diff
@@ -0,0 +1,873 @@
...
+import { type Scope, ScopeGroup, type ScopeName } from '../types/scopes';
+import { SpaceMemberRole } from '../types/space';
+
+const scopes: Scope[] = [
+    {
+        name: 'view:Dashboard',
+        description: 'View dashboards',
+        isEnterprise: false,
+        group: ScopeGroup.CONTENT,
+        getConditions: (context) => {
```
rephus, line 29: I have the feeling we are rebuilding CASL abilities here, is there another better way to do this ? 🤔 I can't think of one

packages/common/src/authorization/scopeAbilityBuilder.ts
stevelikesmusic: We mutate the `builder` here similar to what's being currently being done:
https://github.com/lightdash/lightdash/blob/8e284bc04141efff8270368ec0a66a6474455761/packages/common/src/authorization/organizationMemberAbility.ts#L15-L21

I'm thinking we keep this here and also include a scope—existing permissions work with config like they do today, but then can also assign PAT via scopes. What do you think?

packages/common/src/authorization/scopes.ts
```diff
@@ -0,0 +1,873 @@
...
+                            role: SpaceMemberRole.EDITOR,
+                        },
+                    },
+                },
+                {
+                    organizationUuid,
+                    access: {
+                        $elemMatch: {
+                            userUuid,
+                            role: SpaceMemberRole.ADMIN,
```
rephus: should editor not be enough if EDITOR < ADMIN ?

packages/common/src/authorization/scopes.ts
```diff
@@ -0,0 +1,873 @@
...
+        name: 'manage:SavedChart',
+        description: 'Create, edit, and delete saved charts',
+        isEnterprise: false,
+        group: ScopeGroup.CONTENT,
+        getConditions: (context) => {
+            const { organizationUuid, userUuid } = context;
+            if (context.scopes.has('manage:Organization')) {
+                return [{ organizationUuid }];
+            }
+            return [
```
rephus, line 104: I think repeated conditions should be extracted into a function, to improve readability

packages/common/src/authorization/scopes.ts
```diff
@@ -0,0 +1,873 @@
...
+                },
+            ];
+
+            // User's accessible spaces
+            if (context.userUuid) {
+                conditions.push({
+                    organizationUuid: context.organizationUuid,
+                    projectUuid: context.projectUuid,
+                    access: {
+                        $elemMatch: { userUuid: context.userUuid },
```
rephus: I've seen this condition in another place, perhaps we should extract this into a reusable function

packages/common/src/authorization/scopes.ts
stevelikesmusic: This was in the context of a JWT or service account where we might not have a registered userUuid. This is specific to spaces access. 

> is there a way to enforce this context on a call

It should be that if this check is missing and we make a query against CASL without the needed conditions this will fail. But I don't think there's a typesafe way of enforcing this because abilities are assigned at runtime.

packages/common/src/authorization/scopes.ts
stevelikesmusic: We're using CASL kind of as the foundation of a framework. All abilities are being mapped to static scopes. I can't think of a better way to do this either.

packages/common/src/authorization/scopes.ts
stevelikesmusic: In org/project abilities, we grant EDITOR _and_ ADMIN 😅 
https://github.com/lightdash/lightdash/blob/8e284bc04141efff8270368ec0a66a6474455761/packages/common/src/authorization/organizationMemberAbility.ts#L130-L166

Issue #404, opened 2021-09-02
table calculations can reference other table calculations


TuringLovesDeathMetal:
Right now, you can only build table calculations from metrics/dimensions. 

You should be able to have table calculations that reference other table calculations. 

e.g. 
```
users_active_50_plus = COUNT(DISTINCT(IF(${users.average_spend} > 50, ${users.user_id}, NULL)))

users_active_50_plus_value_per_user = SUM(${users.average_spend}) / NULLIF(${users_active_50_plus}, 0)
```

JohnRomanski:
A table calculation traditionally (in Looker) does not generate new SQL or modify the results of the query.
> [Table calculations](https://cloud.google.com/looker/docs/table-calculations) operate on the results from your query after it has run. 

note: In Looker, SQL can be modified through the use of [custom fields](https://cloud.google.com/looker/docs/custom-fields) but this is presented a different functionality and requires different permissions than table calculations. Custom fields did not come along until much later, suggesting to me that they are a less universal need, more complicated to develop, or both.

yileeZAVA:
Hello team, just to understand the last auto updates from canny (bot). Does it mean that now we can refer table calculations in table calculations? From previous messages, this ticket seems to be dropped? Thanks.

TuringLovesDeathMetal:
Hey Zi - this was just me testing an integration. This is still an open issue

yileeZAVA:
Ah right. I got it.  Thank you!

jranks123:
This feature would be SO useful!

jparkington:
+1 for this feature!

PR #16465, opened 2025-08-18
feat: enable table calcs referencing other table calcs

magnew:
Closes: #404 

### Description:

Makes it possible for table calculations to reference other table calculations. The approach is:
- Make a dependency graph for TCs
- Look for cycles
- Sort them based on dependencies so we know which to compile first
- Compile them in order by injecting the SQL from upstream calcs into the downstream ones

To test:
- Create a table calc
- Create another one that references the first one
- You can continue to chain them

What this **doesn't** solve:
- Auto complete in the FE (implemented up the stack)
- Issues with TCs getting renamed when edited

Here's what it looks like (the last TC is defined like `${rate_plus_1} + ${rate_plus_1_plus_1}`:
Screenshot 2025-08-18 at 15 43 40

### DIFF
--- a/packages/backend/src/queryCompiler.ts
+++ b/packages/backend/src/queryCompiler.ts
@@ -16,10 +16,127 @@ import {
     type WarehouseSqlBuilder,
 } from '@lightdash/common';
 
+interface TableCalculationDependency {
+    name: string;
+    dependencies: string[];
+}
+
+const getTableCalculationReferences = (sql: string): string[] => {
+    const matches = sql.match(lightdashVariablePattern) || [];
+    return matches.map((match) => match.slice(2, -1)); // Remove ${ and }
+};
+
+const buildTableCalculationDependencyGraph = (
+    tableCalculations: TableCalculation[],
+): TableCalculationDependency[] =>
+    tableCalculations.map((calc) => ({
+        name: calc.name,
+        dependencies: getTableCalculationReferences(calc.sql),
+    }));
+
+const detectCircularDependencies = (
+    dependencies: TableCalculationDependency[],
+): void => {
+    const visited = new Set<string>();
+    const recursionStack = new Set<string>();
+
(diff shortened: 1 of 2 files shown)


magnew:
* **#16467** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16465** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16465?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16465?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/queryCompiler.ts
```diff
@@ -16,36 +16,215 @@ import {
...
 
+interface TableCalculationDependency {
+    name: string;
+    dependencies: string[];
+}
+
+const getTableCalculationReferences = (sql: string): string[] => {
+    const matches = sql.match(lightdashVariablePattern) || [];
+    return matches.map((match) => match.slice(2, -1)); // Remove ${ and }
+};
```
almeidabbm, line 27: This will take into account all references in the table calculation. I believe this is fine as only TableCalculations are taken into account for `dependencies` when performing dfs, just pointing it out

packages/backend/src/queryCompiler.ts
```diff
@@ -16,36 +16,215 @@ import {
...
+
+    for (const dep of dependencies) {
+        if (!visited.has(dep.name)) {
+            dfs(dep.name, []);
+        }
+    }
+};
+
+const sortTableCalcs = (
+    dependencies: TableCalculationDependency[],
```
almeidabbm: can we rename this to dependencyGraph? it will make it easier to read the code, specially the nested loop on line 93

PR #16485, opened 2025-08-19
chore: add and clean run specific schemas for e2e CLI

almeidabbm:
### Description:
Added run id specific schemas to remove concurrency issues + cleanup job after

This change will help prevent test failures due to using the same db when multiple workflows are triggered at the same time.

image

test-cli

### DIFF
--- a/.github/workflows/cli-tests.yml
+++ b/.github/workflows/cli-tests.yml
@@ -47,7 +47,7 @@ jobs:
     # This is used to test the CLI commands that only use Lightdash API
     # like login, content as code, etc.
     # We don't need to install dbt dependencies for this.
-    cli-tests-without-dbt: 
+    cli-tests-without-dbt:
         timeout-minutes: 30
         runs-on: ubuntu-latest
         needs: prepare-preview
@@ -83,7 +83,7 @@ jobs:
     # CLI tests with a single dbt version
     # They require simple dbt interaction, for building manifest
     # But in these tests we are more interested about the integration wiht Lightdash
-    # For commands like preview, validate, etc. 
+    # For commands like preview, validate, etc.
     cli-tests:
         timeout-minutes: 30
         runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#16485** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16485?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16485?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/cli-tests.yml
```diff
@@ -166,6 +168,8 @@ jobs:
                 # Make sure we support the first and 2 latest versions
                 dbt: ['1.4', '1.9', '1.10']
         name: dbt ${{ matrix.dbt }}
+        concurrency:
+            group: cli-dbt-tests-${{strategy.job-index}}-concurrency-group
```
rephus: should we use the same group, regardless of the job-index  ? 

I'm not fully sure how these groups works, but if we use different group names, they will still run in parallel, right ?

almeidabbm: Yes, these groups will only block execution while another job with the same group is running across the repository. This is so that we don't run 2 of the same job at the same time in different PRs, it will slow down CLI jobs but should make them a lot more stable

If it doesn't we can revert 👌

rephus: > This is so that we don't run 2 of the same job at the same time in different PRs

Why do we care about this ? these github actions will be running against different okteto preview environments, so they should not affect performance , or risk race conditions

I thought the goal of this PR was to create groups between CLI tests for the same PR, to avoid tests running on the same okteto preview env.

almeidabbm: Sorry I misunderstood the question, having the same group will block these jobs from running in parallel.
These jobs use the `index` for the schema, so they should be safe to run in parallel, I believe the issue is when running the job with the same index on different PRs

almeidabbm: The error we were seeing was mainly when running DBT,

image

Notice the schema in the error `jaffle-1`. I believe what was causing this was another CI Job running at the same time which targeted the same index schema.

almeidabbm:
Dropped schemas after run: https://github.com/lightdash/lightdash/actions/runs/17291383821/job/49080654800

image

PR #16490, opened 2025-08-19
fix: add create and delete scopes for VirtualView

rephus:
Closes: 

We added 2 new scopes : create and delete virtualviews

Existing old roles will keep using manage:virtualView, so they will be able to create or delete virtual views

Tested updating old roles:

With create only (but not delete):

Screenshot from 2025-08-20 09-43-07

With manage: 
Screenshot from 2025-08-20 09-35-08

### Description:
Added two new authorization scopes for virtual views: `create:VirtualView` and `delete:VirtualView`. These granular permissions will allow for more fine-grained access control when working with virtual views, complementing the existing `manage:VirtualView` scope.

### DIFF
--- a/packages/common/src/authorization/scopeAbilityBuilder.test.ts
+++ b/packages/common/src/authorization/scopeAbilityBuilder.test.ts
@@ -1365,6 +1365,188 @@ describe('scopeAbilityBuilder', () => {
             });
         });
 
+        describe('virtual view permissions', () => {
+            it('should handle create:virtual_view permissions', () => {
+                const ability = buildAbilityFromScopes({
+                    ...baseContext,
+                    scopes: ['create:virtual_view'],
+                });
+
+                expect(
+                    ability.can(
+                        'create',
+                        subject('VirtualView', {
+                            organizationUuid: 'org-123',
+                            projectUuid: 'project-123',
+                        }),
+                    ),
+                ).toBe(true);
+
+                // Should not be able to delete with create scope
+                expect(
+                    ability.can(
+                        'delete',
+                        subject('VirtualView', {
+                            organizationUuid: 'org-123',
+                            projectUuid: 'project-123',
+                        }),
+                    ),
+                ).toBe(false);
+            });
+
+            it('should handle delete:virtual_view permissions', () => {
+                const ability = buildAbilityFromScopes({
+                    ...baseContext,
+                    scopes: ['delete:virtual_view'],
+                });
+
+                expect(
+                    ability.can(
+                        'delete',
+                        subject('VirtualView', {
+                            organizationUuid: 'org-123',
+                            projectUuid: 'project-123',
+                        }),
+                    ),
+                ).toBe(true);
+
+                // Should not be able to create with delete scope
+                expect(
+                    ability.can(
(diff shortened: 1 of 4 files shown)


rephus:
* **#16490** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16490?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16490?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/authorization/scopeAbilityBuilder.test.ts
```diff
@@ -1365,6 +1365,188 @@ describe('scopeAbilityBuilder', () => {
             });
         });
 
+        describe('virtual view permissions', () => {
```
rephus, line 1368: added tests to make sure create and delete are 2 se parate scopes, but manage can do both.

packages/common/src/authorization/scopes.ts
```diff
@@ -514,6 +514,20 @@ const scopes: Scope[] = [
...
+        isEnterprise: false,
+        group: ScopeGroup.DATA,
+        getConditions: addDefaultOrgIdCondition,
+    },
+    {
+        name: 'delete:VirtualView',
+        description: 'Delete virtual views',
+        isEnterprise: false,
+        group: ScopeGroup.DATA,
+        getConditions: addDefaultOrgIdCondition,
```
rephus, line 529: TODO allow deleting your own created scope, can be done  on a separate PR

stevelikesmusic: I just added a modifier to scopes for this use-case to power Spaces a bit better. What do you think of defining those like `delete:VirtualView@self` or `manage:Space@self`. 

https://github.com/lightdash/lightdash/pull/16456/files#diff-1776342f7cd2898b6a4b47331c26e9fdb385fc3b180f8ac72e52aeb925ebd28eR114-R132

packages/frontend/src/components/Explorer/ExplorePanel/index.tsx
```diff
@@ -251,7 +259,7 @@ const ExplorePanel: FC<ExplorePanelProps> = memo(({ onBack }) => {
             >
                 <Group position="apart">
                     <PageBreadcrumbs size="md" items={breadcrumbs} />
-                    {canManageVirtualViews &&
+                    {canCreateVirtualViews &&
```
stevelikesmusic: Could you use the CASL `<Can>` component? This would make it work with accounts as well.

rephus: done

packages/common/src/authorization/scopes.ts
rephus: I think it would be nice if we can assume for now that `create = create + delete self` and  `delete = delete from others`  and `manage = create + delete` 

Not sure if this is possible though 

I'm trying to prioritize the `user/dev` experience, over the permisison internal logic

PR #16517, opened 2025-08-20
chore: add logging for account edges

stevelikesmusic:
Closes: #N/A

### Description:
Part of a series of follow-ups to the [Embedded AuthZ Incident ](https://www.notion.so/lightdash/Authorization-error-on-embedded-dashboards-255a63207a7a80a59b70ef3075833b48)

Prevents multiple authentication methods from overriding each other by adding checks to ensure an account is not already authenticated before setting a new authentication method. Added warning logs when an account is already authenticated via OAuth, API key, service account, or session to help debug authentication issues.

### DIFF
--- a/packages/backend/src/controllers/authentication/middlewares.ts
+++ b/packages/backend/src/controllers/authentication/middlewares.ts
@@ -13,6 +13,7 @@ import { ErrorRequestHandler, Request, RequestHandler } from 'express';
 import passport from 'passport';
 import { URL } from 'url';
 import { fromApiKey, fromOauth } from '../../auth/account/account';
+import { buildAccountExistsWarning } from '../../auth/account/warnAccountExists';
 import { lightdashConfig } from '../../config/lightdashConfig';
 import { authenticateServiceAccount } from '../../ee/authentication';
 import Logger from '../../logging/logger';
@@ -64,8 +65,15 @@ export const allowOauthAuthentication: RequestHandler = (req, res, next) => {
                     organization: token.user.organizationUuid,
                 })
                 .then((user) => {
-                    req.account = fromOauth(user, token);
-
+                    if (req.account?.isAuthenticated()) {
+                        Logger.warn(
+                            buildAccountExistsWarning('OAuth'),
+                            req.account?.authentication?.type,
+                        );
+                    }
+                    if (user) {
+                        req.account = fromOauth(user, token);
+                    }
                     req.user = user;
                     next();
                 })
@@ -107,6 +115,13 @@ export const allowApiKeyAuthentication: RequestHandler = (req, res, next) => {
             req,
             res,
             () => {
+                if (req?.account?.isAuthenticated()) {
+                    Logger.warn(
+                        buildAccountExistsWarning('ApiKey'),
+                        req.account?.authentication?.type,
+                    );
+                }
+
                 if (req.user) {
                     req.account = fromApiKey(
                         req.user!,
--- a/packages/backend/src/auth/account/warnAccountExists.ts
+++ b/packages/backend/src/auth/account/warnAccountExists.ts
@@ -1,0 +1,6 @@
+/**
+ * Common util for warning about an account already being authenticated.
+ * Centralizes messages to enable creating metrics for this case.
+ */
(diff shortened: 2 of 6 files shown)


stevelikesmusic:
* **#16517** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/controllers/authentication/middlewares.ts
```diff
@@ -64,7 +64,14 @@ export const allowOauthAuthentication: RequestHandler = (req, res, next) => {
                     organization: token.user.organizationUuid,
                 })
                 .then((user) => {
-                    req.account = fromOauth(user, token);
+                    if (!req.account?.isAuthenticated()) {
+                        req.account = fromOauth(user, token);
+                    } else {
+                        Logger.warn(
+                            'OAuth Account is already authenticated',
+                            req.account,
```
almeidabbm: Can we just log the account authentication type on the new logs? We should probably avoid logging the whole account

almeidabbm (changes requested):

packages/backend/src/controllers/authentication/middlewares.ts
stevelikesmusic: Agreed

almeidabbm (approved):

packages/backend/src/controllers/authentication/middlewares.ts
```diff
@@ -107,11 +114,16 @@ export const allowApiKeyAuthentication: RequestHandler = (req, res, next) => {
             req,
             res,
             () => {
-                if (req.user) {
+                if (req.user && !req?.account?.isAuthenticated()) {
```
almeidabbm: Do we need these added checks in here `!req?.account?.isAuthenticated()` ?

stevelikesmusic: For isAuthenticated, or the optional chaining? I suppose if we have an account, then we have `isAuthenticated`. The additional check is a sanity check from seeing the Account is not embedded error. Maybe it's worth removing and only logging if we have an account before making the wrapping call.

almeidabbm: The `isAuthenticated`, asking just because we didn't have it previously, but probably this is fine?

stevelikesmusic: 🤔 I added this check to the other middleware functions so as not to overwrite the account if it already exists. But maybe we'll learn more if we keep the behavior and log errors. I think it'd be fine, but I like the intention of moving cautiously with these middleware changes.

almeidabbm (changes requested):
CI is failing now
Also, we should look at this graphite comment and understand if it is relevant or not 👌

packages/backend/src/controllers/authentication/middlewares.ts
```diff
@@ -64,8 +65,12 @@ export const allowOauthAuthentication: RequestHandler = (req, res, next) => {
...
                 })
                 .then((user) => {
-                    req.account = fromOauth(user, token);
-
+                    if (req.account?.isAuthenticated()) {
+                        warnAccountExists(req.account, 'OAuth');
+                    }
+                    if (req.user) {
+                        req.account = fromOauth(user, token);
+                    }
```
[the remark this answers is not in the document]
stevelikesmusic: Is this what we want? If we return early here we change the current behavior. I understood that we want to determine what is happening. If returning early here is acceptable, then I think we need a different approach.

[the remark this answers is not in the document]
almeidabbm: This PR is just for logging, if this changes behavior then I would say we should keep it 👌 
Was just wondering if the comment was valid or not, graphite has the tendency to loose context and make suggestions that don't really make sense 😓

stevelikesmusic:
@almeidabbm I think we're good now?

PR #16518, opened 2025-08-20
feat: use projectUuid in embed for JWT abilities

stevelikesmusic:
Closes: #N/A

### Description:

Part of follow-ups to the [Embedded AuthZ Incident](https://www.notion.so/lightdash/Authorization-error-on-embedded-dashboards-255a63207a7a80a59b70ef3075833b48)

@ZeRego suggested that customers may not be sending a proper JWT with all the claims we expect. This seems correct. I made an assumption that `projectUuid` would always be a claim in the JWT, but we don't enforce the JWT schema for embeds. The theory is that customers that were affected do not mint tokens with `projectUuid`.

We use `projectUuid` when applying CASL abilities. This would then make sense because we have authenticated users that get 403 Forbidden errors when accessing resources we expect to be available.

The error occurred because "View Underlying Data" moved embedded dashboards to use endpoints outside of the Embed Controller. This meant that permissions are more important, because they're checked in those services—but not in EmbedService. If `projectUuid` is missing, we don't apply the correct abilities when resolving the identity.

This change uses `projectUuid` from the embed record we get back from the database rather than the `projectUuid` from the JWT. That will give us a guaranteed `projectUuid` and not rely on the token. We only check the JWT for claims related to granting permissions (ex. canExplore, canDateZoom, etc.)

### DIFF
--- a/packages/backend/src/auth/account/account.test.ts
+++ b/packages/backend/src/auth/account/account.test.ts
@@ -1,9 +1,9 @@
 import { AbilityBuilder } from '@casl/ability';
 import {
     CreateEmbedJwt,
-    Embed,
     MemberAbility,
     OrganizationMemberRole,
+    OssEmbed,
     SessionUser,
     UserAccessControls,
 } from '@lightdash/common';
@@ -13,9 +13,21 @@ import { fromJwt, fromSession } from './account';
 
 describe('account', () => {
     describe('fromJwt', () => {
-        const mockOrganization: Embed['organization'] = {
-            organizationUuid: 'test-org-uuid',
-            name: 'Test Organization',
+        const mockEmbed: OssEmbed = {
+            organization: {
+                organizationUuid: 'test-org-uuid',
+                name: 'Test Organization',
+            },
+            projectUuid: 'test-project-uuid',
+            encodedSecret: 'test-encoded-secret',
+            dashboardUuids: ['test-dashboard-uuid'],
+            allowAllDashboards: false,
+            createdAt: '2021-01-01',
+            user: {
+                userUuid: 'test-user-uuid',
+                firstName: 'Test',
+                lastName: 'User',
+            },
         };
 
         const mockDecodedToken: CreateEmbedJwt = {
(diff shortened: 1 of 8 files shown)


stevelikesmusic:
* **#16518** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16518?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/types/auth.ts
```diff
@@ -56,6 +57,16 @@ export type OauthAuth = {
     resource?: URL;
 };
 
+export type OssEmbed = {
```
rephus, line 60: will this be backwards compatible ? as in,  can an existing token have access to the fucntionality, without throwing errors about a missing projectuuid ?

packages/backend/src/auth/account/account.test.ts
```diff
@@ -13,9 +14,21 @@ import { fromJwt, fromSession } from './account';
 
 describe('account', () => {
```
rephus, line 14: we should probably add a test for a minimal jwt config, to ensure all changes are backwards comaptible with existing generated jwt tokens, for that, new fields might remain optional.

packages/common/src/types/auth.ts
stevelikesmusic: Naming is hard. OssEmbed is taken from the Embed DB type where `projectUuid` is a required. I just copied it out of the EE folder so that we could use the type in OSS. It's then aliased as Embed so as not to need to change many more files that are unrelated to the scope of this PR. 

https://github.com/lightdash/lightdash/blob/5fd12eed8d9b4b784257586697ade51d03b9af1e/packages/common/src/ee/embed/index.ts#L6-L14

The JWT type has the optional `projectUuid` that I missed as being required:
https://github.com/lightdash/lightdash/blob/5fd12eed8d9b4b784257586697ade51d03b9af1e/packages/common/src/ee/embed/index.ts#L103-L116

packages/backend/src/auth/account/account.test.ts
stevelikesmusic: I now remember "fixing" an e2e test that had a missing `projectUuid` to get underlying data to pass—thinking it was required 😅 . I'll leave that test as is when we add underlying data again. 

The tests here actually don't even have `projectUuid` in the mock token. I removed the projectUuid from the jwtAbility.test.

Issue #16252, opened 2025-08-06
Enable integer (numeric) parameters


magnew:
### Description

Parameters are always strings at the moment. It would be powerful to also allow numbers. 

### Plan of work (PRs open)
- [x] Phase 1: Update type definitions in common package to support parameter types
- [x] Phase 2: Update JSON schema validation for number parameters
- [x] Phase 3: Modify SQL replacement logic to handle numbers without quotes
- [x] Phase 4: Integration with Query Builder
- [x] Phase 5: Add frontend components for number input with validation
- [x] Phase 6: Add parameter type validation and conversion utilities
- [x] Phase 7: [Update documentation and add tests](https://github.com/lightdash/mintlify-docs/issues/133)

PR #16543, opened 2025-08-21
feat: add number to parameter value types

pimmey:
Closes: #16252 

### Description:
Add support for number parameters in SQL queries and dashboards. This enhancement allows parameters to be defined with a `type: number` property, which will:

1. Pass numeric values directly to SQL without quotes
2. Support both single number values and arrays of numbers
3. Properly validate number parameters to prevent SQL injection
4. Update parameter definitions in project config schema

The implementation includes:
- Type-aware parameter replacement in SQL queries
- Updated parameter type definitions throughout the codebase
- Frontend support for displaying and selecting numeric parameters
- Validation to ensure number parameters contain valid numeric values

This change maintains backward compatibility with existing string parameters while providing better support for numeric data types.

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -236,11 +236,7 @@ import {
     getFilteredExplore,
 } from '../UserAttributesService/UserAttributeUtils';
 import { UserService } from '../UserService';
-import { ValidationService } from '../ValidationService/ValidationService';
-import {
-    combineProjectAndExploreParameters,
-    getAvailableParameterDefinitions,
-} from './parameters';
+import { getAvailableParameterDefinitions } from './parameters';
 
 export type ProjectServiceArguments = {
     lightdashConfig: LightdashConfig;
@@ -1796,7 +1792,6 @@ export class ProjectService extends BaseService {
             userAttributes,
             timezone,
             parameters,
-            availableParameters,
             parameterDefinitions: availableParameterDefinitions,
         });
 
@@ -1819,7 +1814,7 @@ export class ProjectService extends BaseService {
             projectUuid,
         );
 
-        return combineProjectAndExploreParameters(projectParameters, explore);
+        return getAvailableParameterDefinitions(projectParameters, explore);
     }
 
     async compileQuery(
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
@@ -78,8 +78,7 @@ export type BuildQueryProps = {
     warehouseSqlBuilder: WarehouseSqlBuilder;
     userAttributes?: UserAttributeValueMap;
     parameters?: ParametersValuesMap;
-    availableParameters: string[];
-    parameterDefinitions?: ParameterDefinitions;
+    parameterDefinitions: ParameterDefinitions;
     intrinsicUserAttributes: IntrinsicUserAttributes;
     timezone: string;
 };
@@ -809,7 +808,7 @@ export class MetricQueryBuilder {
         const compiledMetric = exploreCompiler.compileMetricSql(
             { ...metric, sql: processedSql }, // use preprocessed SQL with CTE references resolved
             explore.tables,
-            this.args.availableParameters,
+            Object.keys(this.args.parameterDefinitions),
         );
 
         return compiledMetric.sql;
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -61,7 +61,7 @@ import { SavedChartModel } from '../../../models/SavedChartModel';
 import { UserAttributesModel } from '../../../models/UserAttributesModel';
 import { BaseService } from '../../../services/BaseService';
 import {
-    combineProjectAndExploreParameters,
+    getAvailableParameterDefinitions,
     getDashboardParametersValuesMap,
 } from '../../../services/ProjectService/parameters';
 import { ProjectService } from '../../../services/ProjectService/ProjectService';
@@ -642,7 +642,7 @@ export class EmbedService extends BaseService {
         const projectParameters =
             await this.projectService.projectParametersModel.find(projectUuid);
 
-        return combineProjectAndExploreParameters(projectParameters, explore);
+        return getAvailableParameterDefinitions(projectParameters, explore);
     }
 
     private async _runEmbedQuery({
--- a/packages/backend/src/services/ProjectService/parameters.ts
+++ b/packages/backend/src/services/ProjectService/parameters.ts
@@ -57,15 +57,3 @@ export const getAvailableParameterDefinitions = (
         ...exploreParameterDefinitions,
     };
 };
-
-/**
- * Combine project and explore parameters
- * @param projectParameters - The project parameters
- * @param explore - The explore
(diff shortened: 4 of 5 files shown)


pimmey:
* **#16543** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16543?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16543?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/providers/Explorer/types.ts
```diff
@@ -163,7 +164,10 @@ export type Action =
       }
     | {
           type: ActionType.SET_PARAMETER;
-          payload: { key: string; value: string | string[] | null };
+          payload: {
+              key: string;
+              value: string | number | string[] | number[] | null;
```
almeidabbm: We can use `ParameterValue` in here as well

packages/frontend/src/pages/ViewSqlChart.tsx
```diff
@@ -73,7 +73,7 @@ const ViewSqlChart = () => {
     });
 
     const handleParameterChange = useCallback(
-        (key: string, value: string | string[] | null) => {
+        (key: string, value: string | number | string[] | number[] | null) => {
```
almeidabbm: Same `ParameterValue`

packages/frontend/src/features/sqlRunner/utils/monaco.ts
```diff
@@ -108,7 +108,11 @@ export const registerCustomCompletionProvider = (
     settings?: SqlEditorPreferences,
     availableParameters?: Record<
         string,
-        { label: string; description?: string; default?: string | string[] }
+        {
+            label: string;
+            description?: string;
+            default?: string | number | string[] | number[];
```
almeidabbm: `ParameterValue`

packages/frontend/src/features/sqlRunner/store/sqlRunnerSlice.ts
```diff
@@ -231,7 +231,7 @@ export const sqlRunnerSlice = createSlice({
             state,
             action: PayloadAction<{
                 key: string;
-                value: string | string[] | null;
+                value: string | number | string[] | number[] | null;
```
almeidabbm: Basically everywhere we define these we should use the type 💪

packages/frontend/src/features/parameters/components/ParameterInput.tsx
```diff
@@ -311,12 +313,27 @@ export const ParameterInput: FC<ParameterInputProps> = ({
...
+            if (parameter.type === 'number' && newValues !== null) {
+                if (Array.isArray(newValues)) {
+                    finalValue = newValues.map((v) => {
+                        const num = Number(v);
+                        return isNaN(num) ? v : num;
+                    }) as string[] | number[];
+                } else {
+                    const num = Number(newValues);
+                    finalValue = isNaN(num) ? newValues : num;
+                }
```
almeidabbm: should we be returning the value in here if `Number()` produces `NaN`? Perhaps we should error or filter out? Specially because we know the type is `number` so it should never be string?

packages/frontend/src/features/parameters/components/ParameterInput.tsx
```diff
@@ -167,7 +165,10 @@ export const ParameterInput: FC<ParameterInputProps> = ({
             if (parameter.allow_custom_values && newValue.trim()) {
                 if (parameter.multiple) {
                     const updatedValues = [...currentValues, newValue.trim()];
-                    onParameterChange(paramKey, updatedValues);
+                    onParameterChange(
+                        paramKey,
+                        updatedValues as string[] | number[],
```
almeidabbm: We should not need for type cast in here `currentValues` `useMemo` above should produce the correct type, main issue, we should introduce a type guard there to make sure the inference will be `string[]` or `number[]`

packages/backend/src/utils/QueryBuilder/utils.ts
```diff
@@ -152,7 +152,7 @@ const getWrapChars = (wrapChar: string): [string, string] => {
 export const replaceLightdashValues = (
     regex: RegExp,
     sql: string,
-    valuesMap: Record<string, string | string[]>,
+    valuesMap: Record<string, string | number | string[] | number[]>,
```
almeidabbm, line 155: Here it shouldn't be ParameterValue because this is generic function, this type is correct

packages/backend/src/utils/QueryBuilder/parameters.ts
```diff
@@ -93,3 +108,86 @@ export const unsafeReplaceParametersAsRaw = (
...
+        return allParametersResult;
+    }
+
+    // Split parameters by type
+    const stringParameters: ParametersValuesMap = {};
+    const numberParameters: ParametersValuesMap = {};
+
+    Object.entries(parameterValuesMap).forEach(([key, value]) => {
+        const paramDef = parameterDefinitions?.[key];
+        const isNumberType = paramDef?.type === 'number';
```
almeidabbm, line 150: 🧑‍🍳 🤌

packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
```diff
@@ -1297,11 +1300,18 @@ export class MetricQueryBuilder {
...
+                  sql: query,
+                  parameterValuesMap: this.args.parameters ?? {},
+                  parameterDefinitions: this.args.parameterDefinitions,
+                  sqlBuilder: this.args.warehouseSqlBuilder,
+              })
+            : safeReplaceParametersWithSqlBuilder(
+                  query,
+                  this.args.parameters ?? {},
+                  this.args.warehouseSqlBuilder,
+              );
```
almeidabbm: Why do we need to have this check? shouldn't it always be with types?

almeidabbm (changes requested):

packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
pimmey: Good call! There was indeed a fallback within `safeReplaceParametersWithTypes` :)

packages/frontend/src/features/parameters/components/ParameterInput.tsx
pimmey: Good catch! Filtering out it is ✅

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1877,6 +1884,14 @@ export class ProjectService extends BaseService {
             explore,
         );
 
+        const projectParameters = await this.projectParametersModel.find(
+            projectUuid,
+        );
+        const parameterDefinitions = getCombinedParameterDefinitions(
+            projectParameters,
+            explore,
+        );
```
almeidabbm: We are making another query to project in here. Can't we just call `_compileQuery` with the `availableParameterDefinitions` instead of passing both keys and definitions so we can query only once?
Also we should check everywhere that `_compileQuery` is called to update the function call with these

Also `getCombinedParameterDefinitions` should be called `getAvailableParameterDefinitions`

~In `_compileQuery` we are already passing the `availableParameters` from line `1882` so most likely we won't need to duplicate this logic in here.~

almeidabbm (approved):
Lovely 🚀

PR #16614, opened 2025-08-27
fix: prevent overriding existing sorts with default sort

pimmey:
Closes: #ISSUE_NUMBER

### Description:
Fix default sort field logic to only apply when no sorts exist in the unsaved chart version. Previously, the default sort was being applied even when the user had already set custom sorts, which would override their selection.

The fix adds a check to ensure that `unsavedChartVersion.metricQuery.sorts.length` is zero before applying the default sort, and adds this value to the dependency array of the useEffect hook.


pimmey:
* **#16614** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IrakliJani (approved):

Issue #16291, opened 2025-08-08
I want to hide field labels on my funnel chart


jpetey75:
### Description

I want an option to hide labels on funnel charts completely (right now if I show my funnel step count values or percentages the labels show up automatically).

For example, in the image below I want to show the step values, but not the field names (only show the numbers):

<img width="1666" height="736" alt="Image" src="https://github.com/user-attachments/assets/2ffbb2cd-a333-4f06-9e20-cdadc46b6131" />

In the chart configuration if I choose Position: Hidden and check the boxes for Show value and/or Show percentage, the value and percentage should still show, even if the text label is hidden. 

<img width="2454" height="950" alt="Image" src="https://github.com/user-attachments/assets/83caf498-79d3-457e-9930-c6f853d2b012" />

### What problem does this solve?

Decrease clutter, increase customizability

nehaaprasad:
@jpetey75 
now this PR addresses the issue raised by users who want to hide field labels on funnel charts while keeping values and percentages visible, reducing clutter and increasing customizability.

PR #16644, opened 2025-08-29
feat: hide field labels on funnel charts when position is set to hidden

naaa760:
### Description

- This PR adds the ability to completely hide field labels on funnel charts while keeping values and percentages visible. 

fixes: #16291

**What it solves:**

- Users can now hide field names to reduce clutter and increase customizability
- Values and percentages remain visible when configured, even when labels are hidden
- Maintains all existing functionality for other label positions

**How it works:**

- When `Position: Hidden` is selected in the funnel chart configuration, the field names are excluded from the label display. However, if `Show value` and/or `Show percentage` are checked, those numerical values will still appear on the chart.

**Before:** Field names always showed alongside values/percentages
**After:** Field names can be hidden while preserving the numerical data display


joaoviana:
Closing as this doesn't fix the issue

PR #16648, opened 2025-08-29
feat: add custom manifest upload for preview projects

rephus:
Closes: #XXXX

### Description:
Adds support for custom manifest.json uploads in preview projects. This feature allows users to upload pre-compiled dbt manifests instead of generating them during the preview project creation process.

The implementation includes:
- A new textarea field in the CreatePreviewProjectModal for pasting manifest JSON
- JSON validation to ensure the manifest has required fields
- Backend support for passing the custom manifest through the API
- Added manifest parameter to relevant types and interfaces
- Simplified the events.yml example file by removing unnecessary metric examples

This feature helps users who already have compiled manifests to create preview projects more efficiently without requiring Lightdash to recompile their dbt project.

image


rephus:
* **#16648** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ZeRego:
@rephus We went with a textarea instead of a file upload because dbt cloud provides the manifest as text?

> Disable refresh ?
Refreshing a dbt project makes no sense when this config is enabled, since the manifest will be the same, perhaps we should disable the dbt button (can happen on a separate PR)

Makes sense for us to create the preview with type `DbtProjectType.NONE`, which disables the refresh button and other functionality that should not be supported in this case.

I also have concerns about the payload size since the manifest can have multiple MB and how slow the dbt connection de/encryption is going to take. This can be mitigated if the FE removes unnecessary data (eg: exposures) from the manifest before making the request. 

If we use `DbtProjectType.NONE`, we don't need to store the manifest in the database. It is only necessary during creation.

rephus:
> @rephus We went with a textarea instead of a file upload because dbt cloud provides the manifest as text?

Yes, apparently, on dbt cloud is easier to copy the manifest (they don't have an option to export) than uploading a file. 

> This can be mitigated if the FE removes unnecessary data (eg: exposures) from the manifest before making the request.

I'll take a look. 

> Makes sense for us to create the preview with type `DbtProjectType.NONE`, which disables the refresh button and other functionality that should not be supported in this case.

Good idea, I'll make the change.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1197,6 +1197,7 @@ export class ProjectService extends BaseService {
                     try {
                         // There's no project yet, so we don't track
                         const trackingParams = undefined;
+                        // Get the custom manifest from the create project configuration
```
ZeRego: These comments seem out of place 🤔

packages/frontend/src/components/ProjectConnection/DbtForms/DbtManifestForm copy.tsx
ZeRego, line 1: filename typo " copy"

packages/frontend/src/components/ProjectConnection/DbtForms/DbtManifestForm copy.tsx
```diff
@@ -0,0 +1,53 @@
...
+                        }
+                        target="_blank"
+                        rel="noreferrer"
+                    >
+                        change your connection type
+                    </Anchor>
+                    , setup a{' '}
+                    <Anchor
+                        href={
+                            'https://docs.lightdash.com/guides/cli/how-to-use-lightdash-deploy#automatically-deploy-your-changes-to-lightdash-using-a-github-action'
```
ZeRego: These instructions don't seem accurate here since there is no way to update a manifest project via CLI/github actions atm.

Should we have a textarea input here for updates?

packages/common/src/types/projects.ts
```diff
@@ -231,6 +232,7 @@ export const maybeOverrideWarehouseConnection = <
 
 export interface DbtProjectConfigBase {
     type: DbtProjectType;
+    customManifest?: string;
```
ZeRego: Do we still use this property?

Issue #16637, opened 2025-08-28
Improve dashboard Parameters loading time


pimmey:
### Description

Right now we seem to wait for all of the charts to load before we display parameters in the select and results in a spinning wheel that is shown for excessively long time. If a dashboard has many charts, that becomes a bit inefficient. 

[Internal discussion](https://lightdash.slack.com/archives/C0653KLKUMA/p1756324474984739)

### What problem does this solve?

This improves UX for parameters users who have heavy dashboards.

nehaaprasad:
@pimmey 
resolved the dashboard parameters loading delay by making parameters display immediately when fetched from the API instead of waiting for all charts to finish loading.

PR #16657, opened 2025-08-29
feat: dashboard parameters loading time by showing parameters immediately

naaa760:
Closes: #16637

### Description:
Fixes dashboard parameters loading time by showing parameters immediately when available instead of waiting for all charts to load.

**Before:** Parameters were only displayed after ALL chart tiles finished loading, causing excessive wait times for dashboards with many charts.

**After:** Parameters are displayed as soon as they're fetched from the API, allowing users to interact with them immediately while charts continue loading in the background.

**Changes:**
- Added `parametersReady` state to track when parameters are available
- Modified Parameters component to use `parametersReady` instead of `areAllChartsLoaded`
- Parameters now display immediately, improving UX for heavy dashboards


joaoviana:
Thank you @naaa760 but this issue isn't open to contribution and therefore does not adhere to community guidelines. ty

Issue #16502, opened 2025-08-20
Bar chart labels not visible for large labels


IrakliJani:
When generating a bar chart to display blog page views (using the Pages explore, filtered to paths starting with "/blog"), the resulting chart does not display the labels for each bar. This makes it difficult to interpret which blog page each bar represents, reducing the usefulness of the visualization.

nehaaprasad:
@IrakliJani 
users couldn't see which blog page each bar represented in their page views charts because the labels were hidden.

PR #16659, opened 2025-08-31
fix: bar chart labels not visible for large labels (blog URLs)

naaa760:
### Description:

fixes: #16502

- When generating bar charts with long x-axis labels (like blog page URLs), the labels were being completely hidden due to the `hideOverlap: true` setting, making the charts unreadable.

**What was changed:**
- Modified `getAxisFormatter` function in `useEchartsCartesianConfig.ts` to intelligently handle long labels
- Added automatic 45° rotation for labels longer than 80px instead of hiding them
- Maintained `hideOverlap: true` for short labels to preserve performance
- Preserved user manual rotation settings

**Result:**
- Blog page URLs and other long labels are now visible and readable
- Charts remain performant for short labels
- No breaking changes to existing functionality

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -1035,17 +1035,27 @@ const getEchartAxes = ({
         }
 
         axisConfig.nameGap = defaultNameGap || 0;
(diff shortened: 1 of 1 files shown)


joaoviana:
Thank you for your contribution @naaa760 but this is not open to external contributions

Issue #10672, opened 2024-07-10
I want to see labels instead of field slugs in Funnel charts


rephus:
Currently the field IDs show up in the funnel chart legend and field labels on the chart (see image) rather than the `label` that should be shown. In the image you can see the correct labels should be: Unique deals, Qualified deals, Poc deals 

<img width="1650" height="1186" alt="Image" src="https://github.com/user-attachments/assets/9645965b-0f35-4114-8cd0-1db02143e603" />

ongdisheng:
Hi @jesshitchcock, I’d be happy to take this issue and work on it. Thanks!

jpetey75:
This works now

ZeRego:
@jpetey75 This is not working for me 👀 

<img width="1652" height="1286" alt="Image" src="https://github.com/user-attachments/assets/58992822-95c7-4524-af69-ecd9eea14cfd" />

PR #16712, opened 2025-09-03
feat: Display field labels instead of IDs in funnel chart

ongdisheng:
Closes:  #10672

### Description:

This PR fixes an issue where funnel charts were displaying raw field IDs like `customers_customer_name` instead of human-readable labels like `Customer Name`. The fix updates `useFunnelChartConfig.ts` to use `getItemLabelWithoutTableName()` to properly extract field labels from the `itemsMap` when building the chart data. This ensures users see meaningful field names in their funnel visualizations rather than technical database identifiers.

image


joaoviana:
Thank you for your contribution @ongdisheng but it doesn't work properly: 
image

Feel free to reopen once you find the fix for it 

```
hook.js:608 TypeError: Cannot convert undefined or null to object
    at Object.keys (<anonymous>)
    at useEchartsFunnelConfig.ts:101:22
    at useFunnelChartConfig (useEchartsFunnelConfig.ts:100:32)
```

ongdisheng:
@joaoviana Thanks for catching this! Could you share the steps you took to encounter this error? It seems to be working fine on my end as shown in the screenshot below.
image

joaoviana:
It broke when I started creating a chart (bar by default) and then changed it to funnel 👀

ongdisheng:
Hi @joaoviana, I've created a new PR at #17368 with the fix for the crash. Could you take a look when you get a chance? Thanks!

Issue #16665, opened 2025-09-01
Mobile dashboards calculate tile sizes incorrectly


magnew:
### Description

Mobile dashboards are calculating tile sizes incorrectly with the improvements to dashboard layouts

### Steps to Reproduce the Bug or Issue

1. Make a dashboard with a row of 3 - 4 tiles
2. Size it very narrow
3. Reload so you get the mobile version
4. Some tiles will be squished or lost

### Cloud or self-hosting

None

PR #16666, opened 2025-09-01
fix: mobile dashbaord layout

magnew:
Closes: #16665

### Description:

We recently updated dashboards to scale tiles based on the grid columns, but it's not working right in mobile dashboards. In mobile dashboards, we override the number of grid columns with `stackVerticallyOnSmallestBreakpoint`. 

This removes the column scaling for the smallest size (mobile) minimal dashboards. 

1. Make a dashboard with a row of 3 - 4 tiles
2. Size it very narrow
3. Reload so you get the mobile version
4. All tiles should be there, stacked.

PR #16616, opened 2025-08-27
fix: dashboard responsive layout depending on columns

almeidabbm:
Relates to: #4161

### Description:
Implemented responsive grid layouts for dashboards to properly scale tiles across different screen sizes. The changes include:

1. Enhanced `getReactGridLayoutConfig` to scale tile dimensions based on the number of columns in the current breakpoint
2. Added support for medium and small breakpoints in dashboard layouts
3. Created a TypeScript type for `ResponsiveGridLayoutProps` to improve type safety
4. Optimized grid layout configuration by reusing the same grid props instance
5. Applied responsive layouts to all dashboard views (standard, minimal, and embedded)

This ensures dashboard tiles maintain proper proportions when viewed on different devices, providing a better user experience across various screen sizes.

**Before**

**After**

PR #16853, opened 2025-09-11
fix: mobile layout calculating incorrectly

almeidabbm:
Closes: #16665

### Description:
Improved dashboard grid layout calculations by:
- Added a `DEFAULT_COLS` constant to standardize column references
- Removed unnecessary `Math.round()` calls for x and w coordinates to prevent rounding errors in tile positioning
- Fixed minW to be a constant value of 6 instead of scaling it
- Removed the small breakpoint layout generation which would cause issues with scheduled deliveries.

**Steps to test**
1. Create dashboard with 2 rows, top one with 3 tiles and the one below with 1 tile covering the row
2. Go to minimal view and resize window as you go, it should be responsive and stack vertically on the smallest size
3. Try to export dashboard as png

**Before**

**After**

@magnew had fixed this same issue in https://github.com/lightdash/lightdash/pull/16666 but it ended up causing issues with scheduled deliveries.

The reason for this was that in https://github.com/lightdash/lightdash/pull/16616 we added a layout for `sm` in `minimalDasboard`. But since we stack vertically this was breaking. Fix was to remove that layout and leave only `md` and `lg`, the smallest breakpoint will stack.

**Scheduled deliveries before previous attempt**

**Scheduled deliveries with this fix**

### DIFF
--- a/packages/frontend/src/components/DashboardTabs/gridUtils.ts
+++ b/packages/frontend/src/components/DashboardTabs/gridUtils.ts
@@ -10,20 +10,22 @@ export type ResponsiveGridLayoutProps = {
     rowHeight: number;
 };
 
+const DEFAULT_COLS = 36;
+
 export const getReactGridLayoutConfig = (
     tile: DashboardTile,
     isEditMode = false,
     cols = 36,
 ): Layout => {
     // Scale factor based on the number of columns (36 is the default for lg)
-    const scaleFactor = cols / 36;
+    const scaleFactor = cols / DEFAULT_COLS;
 
     return {
         minH: 1,
-        minW: Math.max(1, Math.round(6 * scaleFactor)),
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#16853** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joaoviana:
Tested
1. minimal dashboard full screen size and other breakpoints
2. minimal dashboard on <768px to render mobile route and tested 
All good

Issue #16594, opened 2025-08-26
Add missing SCIM endpoints


ZeRego:
### Description

Our current SCIM integration supports `/Users` and `/Groups`, but the following **required SCIM 2.0 discovery endpoints** are not implemented:

- `/ServiceProviderConfig`
- `/Schemas`
- `/ResourceTypes`

According to the [SCIM 2.0 Protocol (RFC 7644)](https://datatracker.ietf.org/doc/html/rfc7644) and [SCIM Core Schema (RFC 7643)](https://datatracker.ietf.org/doc/html/rfc7643), these endpoints are mandatory for a compliant SCIM service.

PR #16862, opened 2025-09-12
feat: add SCIM discovery endpoints for identity provider integration

rephus:
Closes: #16594

### /ServiceProviderConfig

Spec 
image

our implementation:

image

### /Schemas sample 

Spec sample: 
image

Our implementation
image

### Resource types 

Spec sample

image

Our implementation

image

### User resource type

image

### Description:
Adds SCIM discovery endpoints to support identity provider integration. This PR implements three new endpoints:

1. `/api/v1/scim/v2/ServiceProviderConfig` - Returns configuration details about Lightdash's SCIM implementation
2. `/api/v1/scim/v2/Schemas` - Returns schema definitions for User and Group resources
3. `/api/v1/scim/v2/ResourceTypes` - Returns available resource types and their endpoints

These endpoints follow the SCIM 2.0 specification and provide identity providers with the necessary metadata to properly integrate with Lightdash's SCIM API. The implementation includes comprehensive test coverage and example HTTP requests for testing.

### DIFF
--- a/packages/backend/src/ee/controllers/scimRootController.ts
+++ b/packages/backend/src/ee/controllers/scimRootController.ts
@@ -1,8 +1,15 @@
+/* eslint-disable class-methods-use-this */
+import {
+    ScimResourceType,
+    ScimSchema,
+    ScimServiceProviderConfig,
+} from '@lightdash/common';
 import {
     Get,
     Hidden,
     Middlewares,
     OperationId,
+    Path,
     Request,
     Response,
     Route,
@@ -11,6 +18,7 @@ import {
 import express from 'express';
 import { BaseController } from '../../controllers/baseController';
 import { isScimAuthenticated } from '../authentication';
+import { ScimService } from '../services/ScimService/ScimService';
 
 @Route('/api/v1/scim/v2')
 @Hidden()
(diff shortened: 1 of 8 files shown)


rephus:
* **#16862** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16862?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/controllers/scimRootController.ts
```diff
@@ -28,4 +35,44 @@ export class ScimRootController extends BaseController {
...
+     */
+    @Middlewares([])
+    @Get('/ResourceTypes')
+    @OperationId('GetScimResourceTypes')
+    @Response<ScimResourceType[]>('200', 'Success')
+    async getResourceTypes(
+        @Request() req: express.Request,
+    ): Promise<ScimResourceType[]> {
+        return ScimService.getResourceTypes();
+    }
```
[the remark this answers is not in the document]
rephus: the RFCs don’t require /Schemas to be authenticated. SCIM allows anonymous (unauthenticated) requests in some deployments, and the spec describes /Schemas simply as a discovery endpoint without mandating auth. So it can be public

Several implementations keep discovery endpoints (/ServiceProviderConfig, /ResourceTypes, /Schemas) open;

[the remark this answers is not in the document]
stevelikesmusic: Any reason to omit the Middlewares decorator altogether? Or decorate the class and let each route method inherit?

packages/backend/src/ee/controllers/scimRootController.test.ts
```diff
@@ -0,0 +1,222 @@
+import { ScimSchemaType } from '@lightdash/common';
+import { ScimService } from '../services/ScimService/ScimService';
+
+describe('SCIM Discovery Endpoints', () => {
```
stevelikesmusic: Should these tests simply live in ScimService.test? Looks like some of these tests are duplicated there.

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -1053,4 +1057,701 @@ export class ScimService extends BaseService {
...
+
+    static async getServiceProviderConfig(): Promise<ScimServiceProviderConfig> {
+        return {
+            schemas: [ScimSchemaType.SERVICE_PROVIDER_CONFIG],
+            documentationUri: 'https://docs.lightdash.com/guides/scim',
+            patch: {
+                supported: true,
+            },
+            bulk: {
+                supported: false,
```
stevelikesmusic, line 1069: I thought we did support bulk?

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -1053,4 +1057,701 @@ export class ScimService extends BaseService {
...
+                        mutability: 'readWrite',
+                        returned: 'default',
+                        uniqueness: 'none',
+                    },
+                ],
+            },
+        };
+
+        // Return all attributes as array - TypeScript ensures all ScimUser fields are covered
+        return Object.values(fieldAttributeMap);
```
stevelikesmusic, line 1214: Why not define this as an array rather than an object?

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -1053,4 +1057,701 @@ export class ScimService extends BaseService {
...
+    }
+
+    static async getSchemas(): Promise<ScimSchema[]> {
```
stevelikesmusic: Why return a Promise? This all looks synchronous.

packages/common/src/ee/scim/types.ts
```diff
@@ -131,3 +131,77 @@ export type ApiCreateScimServiceAccountRequest = Pick<
     ServiceAccount,
     'expiresAt' | 'description'
 >;
+
+export interface ScimServiceProviderConfig {
```
stevelikesmusic, line 135: Do we prefer `interface` or `type` for defining types?

packages/backend/src/ee/services/ScimService/ScimService.ts
rephus: >   The reason we use an object is for type safety - the ScimUserFieldMap mapped type ensures that we
  define attributes for exactly the fields that exist in the ScimUser interface. If we used an array
  directly, we could miss fields or add extra ones without TypeScript catching it.

packages/backend/src/ee/services/ScimService/ScimService.ts
rephus: > Since bulk operations are optional in the SCIM spec and we don't currently implement them, the
  configuration correctly shows bulk: { supported: false }. Steve might have been confused about
  whether we implemented bulk operations, but it looks like we haven't and that's appropriate for an
  initial implementation.

packages/common/src/ee/scim/types.ts
rephus: >   The general TypeScript convention is:
  - Use interface for object shapes that might be extended
  - Use type for unions, computed types, or when you need specific TypeScript features

>  Since all the new types I added (ScimServiceProviderConfig, ScimSchema, ScimResourceType,
  ScimSchemaAttribute) are interfaces that define object shapes and could potentially be extended,
  they correctly use interface and are consistent with the existing SCIM types.

Issue #16855, opened 2025-09-11
Color and series order mismatch on dashboard


pimmey:
### Description

As reported by one of the customers with early access:

> The colours and series ordering of one of my dashboard charts has changed and some point in the last 24 hours, though the version history says no-one has edited it.

More details in [internal thread](https://lightdash.slack.com/archives/C08D77QLD47/p1757581113731629).

### Dashboard

<img width="625" height="542" alt="Image" src="https://github.com/user-attachments/assets/fd5a8f28-53ce-465b-96e9-0fccd09f0865" />

### Explore from here

<img width="1087" height="358" alt="Image" src="https://github.com/user-attachments/assets/2521aa66-4b74-438c-8324-329734a0b2f2" />

### Steps to Reproduce the Bug or Issue

1. Set USE_SQL_PIVOT_RESULTS to `false`
2. Create a chart with pivot and stack
3. Add it to a dashboard
4. Set USE_SQL_PIVOT_RESULTS to `true`
5. Click "Explore from here" from the dashboard chart
6. Notice different order

### version

0.1992.4

### Cloud or self-hosting

cloud

PR #16876, opened 2025-09-12
fix: explore from here consistency

pimmey:
Closes: #16855

### Description:
Auto-fetch results for charts with pivot configurations when loaded from dashboards or as saved charts. This ensures that charts requiring backend pivot processing are properly rendered without requiring manual user interaction.

The changes include:
- Renamed `explore` to `exploreData` to avoid naming conflicts
- Added detection for charts with pivot configurations that need backend processing
- Enhanced the auto-fetch logic to consider pivot configuration requirements
- Added safeguards to wait for explore data when backend pivot processing is needed
- Improved dependency tracking in the useEffect hook for more reliable fetching

### Recording

https://github.com/user-attachments/assets/0b899958-b600-4f39-bcf8-d6fd743c0c23

### Note
This addresses only the sorting, I haven't been able to reproduce different colors.

### DIFF
--- a/packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
+++ b/packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
@@ -1532,7 +1532,7 @@ const ExplorerProvider: FC<
             const metricQuery = unsavedChartVersion.metricQuery;
             let pivotConfiguration: PivotConfiguration | undefined;
 
-            if (!useSqlPivotResults?.enabled || !explore) {
+            if (!explore) {
                 return;
             }
 
--- a/packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
+++ b/packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
@@ -1532,6 +1532,10 @@ const ExplorerProvider: FC<
             const metricQuery = unsavedChartVersion.metricQuery;
             let pivotConfiguration: PivotConfiguration | undefined;
 
+            if (!useSqlPivotResults?.enabled || !explore) {
+                return;
+            }
+
             if (useSqlPivotResults?.enabled && explore) {
                 const items = getFieldsFromMetricQuery(metricQuery, explore);
                 pivotConfiguration = derivePivotConfigurationFromChart(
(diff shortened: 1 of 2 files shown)


pimmey:
* **#16876** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16876?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/16876?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
```diff
@@ -1532,6 +1532,10 @@ const ExplorerProvider: FC<
             const metricQuery = unsavedChartVersion.metricQuery;
             let pivotConfiguration: PivotConfiguration | undefined;
 
+            if (!useSqlPivotResults?.enabled || !explore) {
+                return;
+            }
+
```
almeidabbm: This will block running the query when `useSqlPivotResults` is `false`.

e.g. edit chart and try to run query

packages/frontend/src/providers/Explorer/ExplorerProvider.tsx
```diff
@@ -1532,6 +1532,10 @@ const ExplorerProvider: FC<
             const metricQuery = unsavedChartVersion.metricQuery;
             let pivotConfiguration: PivotConfiguration | undefined;
 
+            if (!explore) {
+                return;
+            }
```
almeidabbm, line 1537: Is this early return needed here?

pimmey: Good question. Perhaps not the most elegant solution, but it does fix the racing condition when the explorer is still not available and we're auto-fetching. Without this early return, it will request metric-query without `pivotConfig`. Do you feel like there's a better alternative?

almeidabbm: Makes sense, I was not understanding the reason initially but this is good 🙇

Issue #16866, opened 2025-09-12
Error when exporting the results from a pivoted table viz


almeidabbm:
### Description

<img width="2541" height="1249" alt="Image" src="https://github.com/user-attachments/assets/58025844-8e79-4c90-b71a-7a85e55180b8" />

### Steps to Reproduce the Bug or Issue

1. Go to explore
2. Add 2 dimensions and 1 metric
3. Set the visualization to be table
4. Move one of the dimensions to columns
5. Run query
6. Click on **viz** csv export

### Cloud or self-hosting

None

nehaaprasad:
@almeidabbm 
fixed the CSV export bug for pivoted tables, you should now be able to export without getting the cannot get key from object error

PR #16886, opened 2025-09-14
fix: pivot table CSV export error when fields are missing

naaa760:
Closes: #16866

### Description:
Fixed a bug where CSV export for pivoted tables would fail with "Cannot get key 'X' from object" error when expected field IDs were missing from the formatted row data.

**Changes:**
- Added safe field accessor function in `pivotQueryResults()` to handle missing fields gracefully
- Returns default values (`{ raw: null, formatted: '' }`) when fields don't exist instead of throwing errors
- Maintains all existing functionality while preventing crashes during pivot CSV export

**Testing:**
Users can now successfully export CSV files from pivoted tables without encountering the accessor error.

### DIFF
--- a/packages/common/src/pivot/pivotQueryResults.ts
+++ b/packages/common/src/pivot/pivotQueryResults.ts
@@ -595,6 +595,17 @@ export const pivotQueryResults = ({
 
     const rowIndices = {};
     const columnIndices = {};
+    const safeGetFieldValue = (
+        row: ResultRow,
+        fieldId: string,
+    ): ResultValue => {
+        const field = row[fieldId];
+        if (field && field.value) {
+            return field.value;
+        }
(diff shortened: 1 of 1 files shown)


almeidabbm:
Hello @naaa760 👋 

Thank you so much for your contribution! 🎉

We’ve reviewed your changes, and while they do resolve the issue, they rely on a fallback field. The root cause is actually deeper: when exporting pivoted tables where the pivoting happens in the backend, we need to use a different function to correctly generate the pivot data. Specifically, this should go through `convertSqlPivotedRowsToPivotData`.

We will be closing this PR in favor of: https://github.com/lightdash/lightdash/pull/16928

nehaaprasad:
@almeidabbm 
okay, thankyou!

Issue #16863, opened 2025-09-12
Switch Debug Panel to Use Artifact System and Display Metric Query/Filters


notgiorgi:
The Debug panel currently does not use the artifact system, which results in incorrect display of metric queries and filters. This issue proposes:

- Refactoring the Debug panel to use the artifact system for accurate representation of metric queries and filters.
- Adding the ability to display tool call results and the ECharts configuration within the Debug panel for improved debugging and transparency.

**Why it's needed:**
- Ensures the Debug panel accurately reflects the current state of metric queries and filters.
- Provides developers and users with more comprehensive debugging information by including tool call results and ECharts config.

nehaaprasad:
@notgiorgi 
the Debug panel now properly uses the artifact system and shows the real metric queries, filters, and ECharts config instead of null values. Much better for debugging

PR #16887, opened 2025-09-14
fix: debug panel to use artifact system for accurate metric query and filter display

naaa760:
Closes: #16863

### Description:
Fixed the Debug panel to properly integrate with the artifact system instead of displaying null values. The panel now shows:

- **Actual metric queries** with real dimensions, metrics, filters, and sorts from the artifact system
- **ECharts configuration** generated from the visualization config (instead of null)
- **Tool call results** (already working, maintained)

**Before:** Debug panel showed `null` for metric queries and visualization config because it wasn't connected to the artifact system.

**After:** Debug panel displays the real data that powers the AI-generated charts, providing developers with accurate debugging information.

**Changes:**
- Updated `AgentChatDebugDrawer` to accept artifact and visualization query data
- Modified `AgentChatAssistantBubble` to fetch artifact data and generate ECharts config
- Integrated with existing artifact hooks (`useAiAgentArtifact`, `useAiAgentA


joaoviana:
Thank you for your contribution @naaa760 but this doesn't align with our guidelines

PR #16917, opened 2025-09-16
fix: upgrade axios dependencies

rephus:
Closes:  https://github.com/lightdash/lightdash-internal-work/issues/3697

### Description:


rephus:
image

Unrelated snyk error, we get this error everytime we change pnpm-lock since we don't use yarn anymore

joaoviana (approved):

package.json
```diff
@@ -15,7 +15,13 @@
         "typescript": "5.5.4"
     },
     "overrides": {
-        "@mantine/core@8.0.0>@mantine/hooks": "$@mantine-8/hooks"
+        "@mantine/core@8.0.0>@mantine/hooks": "$@mantine-8/hooks",
+        "axios": "^1.12.0"
+    },
+    "pnpm": {
+        "overrides": {
+            "trino-client>axios": "^1.12.0"
```
rephus, line 24: The reason we need both is:

  1. "overrides" with "axios": "^1.12.0": This handles direct axios dependencies and some transitive
  ones
  2. "pnpm.overrides" with "trino-client>axios": "^1.12.0": This is pnpm-specific syntax that forces
  the override for nested dependencies

  The > syntax in pnpm overrides is more specific than the general override and ensures that axios
  within trino-client specifically gets updated to ^1.12.0, fixing any security vulnerabilities in
  axios 1.7.7.

Issue #16913, opened 2025-09-15
I want to see the "Apply" button for dashboard filters after I've selected values from the drop down


jpetey75:
### Description

Right now the Apply button gets blocked by the filter drop-down that shows values

<img width="912" height="806" alt="Image" src="https://github.com/user-attachments/assets/efe08bec-7cb0-4dc3-b1f7-a8484b865a58" />

Ideally the Apply button will appear to the right of the values list so it can be pressed without hiding the list first.

nehaaprasad:
Hi @jpetey75 
the Apply button is now positioned to the right of the filter dropdown so it stays accessible when selecting values.

let me know!
thanks!

PR #16941, opened 2025-09-16
fix: position Apply button to prevent blocking by filter dropdown

nehaaprasad:
Closes: #16913

### Description:
Fixed Apply button positioning in dashboard filters to prevent it from being blocked by the filter dropdown. The Apply button now appears to the right of the filter values input, allowing users to apply filters without having to close the dropdown first.

**Changes:**
- Repositioned Apply and Revert buttons to horizontal layout next to filter input
- Apply button remains accessible when filter dropdown is open
- Improved user experience for selecting and applying filter values

**Before:** Apply button was below the filter input and got hidden by dropdown
**After:** Apply button is positioned to the right and always visible


joaoviana:
does not adhere to community guidelines

PR #16977, opened 2025-09-18
chore: fix axios  and linkifyjs dependencies 

rephus:
Closes: 

image

### Description:

test-frontend test-backend test-cli

PR #16979, opened 2025-09-18
fix: fix axios dependency on posthog-node

rephus:
Closes: 
image


package.json
```diff
@@ -24,6 +24,7 @@
     "pnpm": {
         "overrides": {
             "trino-client>axios": "^1.12.0",
+            "posthog-node>axios": "^1.12.0",
```
rephus, line 27: I tried upgrading posthog first to a major version, however, that included some breaking changes

● ⚠️ BREAKING CHANGES DETECTED:

  The posthog-node v4→v5 upgrade has significant breaking changes that WILL break your code:

  1. groupProperties removed - Your code uses groupProperties (line 87-93) which was deprecated and
  removed in v5
  2. Node.js 20+ required ✅ (you have v20.12.0)
  3. API signature changes for feature flags

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -376,10 +376,12 @@ const SchedulerForm: FC<Props> = ({
         validateInputOnBlur: ['options.customLimit'],
 
         validate: {
+            // @ts-ignore
```
rephus, line 379: Ignoring these for now, these types will be fixed in a separate PR https://github.com/lightdash/lightdash/pull/16981

rephus:
Close in favor of https://github.com/lightdash/lightdash/pull/16977

PR #16982, opened 2025-09-18
chore: fix form-data dependencies

rephus:
Closes:


package.json
```diff
@@ -18,14 +18,15 @@
...
-        "linkifyjs": "^4.3.2",
-        "form-data": "^2.5.4"
+        "linkifyjs": "^4.3.2"
     },
     "pnpm": {
         "overrides": {
             "trino-client>axios": "^1.12.0",
-            "@slack/web-api>form-data": "^2.5.4",
-            "@types/request>form-data": "^2.5.4"
+            "form-data@2.5.3": "2.5.5",
```
rephus, line 26: instead of overriding individual dependencies, we override the old version with the fixed version of the package.

rephus: This approach is better because it:
  - Targets exact vulnerable versions regardless of dependency path
  - Is more maintainable than package-specific overrides
  - Provides universal coverage across the entire dependency tree
  - Makes it easy to add new vulnerable versions as they're discovered

PR #16995, opened 2025-09-18
chore: remove danger dependency

rephus:
Closes: 

    - Low risk: No active danger rules are currently implemented
    - No functionality loss: Empty danger checks provide no current value
    - CI improvement: Removes an unused CI step that could fail unnecessarily
    - Dependency reduction: Removes 1 direct dependency + its transitive dependencies


stevelikesmusic (approved):
🔥 🔥 🔥

PR #17001, opened 2025-09-18
fix: AI agent bug when streaming long text

notgiorgi:
### Description:

Improves AI agent response handling by collecting the complete response from all steps instead of just the last text fragment.

The changes:

- In the backend, we now aggregate text from all steps to create a complete response
- In the frontend, we collect and combine all text parts from the message before updating the UI
- This ensures the full context of the AI's response is preserved and displayed to the user

This fix prevents truncated or incomplete responses that could occur when the agent's output was split across multiple steps.


notgiorgi:
* **#17001** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17001?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17001?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

notgiorgi:
It's not actually reasoning, it's normal agent output. At least there's nothing in stream to indicate that those initial messages are something different than the rest of the text (also GPT-4.1 doesnt support reasoning)

tatianainama:
> It's not actually reasoning, it's normal agent output. At least there's nothing in stream to indicate that those initial messages are something different than the rest of the text (also GPT-4.1 doesnt support reasoning)

yeah got it! I was wrong with the words. The thing I'm trying to say is that before this change, the stream was overwriting itself:

1. created a checklist
2. started calling the tools
3. once it finished the checklist, it deleted it from the text and only displayed the response instead


Now, with this change, all the text streamed is maintained, so we see the checklist (multiple times, as it changed while calling tools) and then the final response

IrakliJani (approved):
nice catch!

notgiorgi:
@tatianainama ah yes, definitely something to consider when we support reasoning models! we should ahve a dedicated UI to render reasoning output

tatianainama:
yeah got it! I think I'm confusing us all here 😅 look at the **final state** of these answers:

| Before | This PR |
|--------|--------|
|  |  |


Before this change, the final answer only contains the result, whereas now, it contains the whole stream. Is this something we want?

notgiorgi:
@tatianainama  Yea, this is very confusing, was hard to debug 😅 

All of that text is the output of the LLM, before my fix it was just kind of swallowing the first part because of the bug. if we want shorter response, we need to adjust the system prompt (or I think the llms also have a configuration option for that)

We only caught this bug, because the system prompt changes made it generate a lot of text. if you ask it to generate 10000 words essay on BI tools, it will have a same issue.

This is totally separate from reasoning feature of LLMs

notgiorgi:
### Merge activity

* **Sep 19, 9:49 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17001).
* **Sep 19, 9:49 AM UTC**: @notgiorgi merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17001).

Issue #16790, opened 2025-09-09
dbt write back does not correctly apply the new `config:` option in dbt 1.10


oli-rmsy:
### Description

When using write back, and also using dbt 1.10 where the meta option has been migrated under config, you get a non-valid output.

We need to update write back to work as expected in 1.10

See image below for example output which is not valid Lightdash config.
![](https://files.slack.com/files-pri/T03942PB7E3-F09E96PPG3U/screenshot_2025-09-09_at_17.41.24.png)

### Steps to Reproduce the Bug or Issue

1. write back a dimension from the UI to a project that users `config: meta:`

### Cloud or self-hosting

None

PR #17025, opened 2025-09-22
feat: add dbt version compatibility for schema editor

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/16790

Screenshot from 2025-09-22 10-49-40

Screenshot from 2025-09-22 11-16-35

### Description:
This PR adds support for dbt v1.10+ compatibility in the schema editor. In dbt v1.10+, custom metrics and dimensions need to be placed under `config.meta` instead of directly under `meta`. 

The changes include:
- Added a method to detect if the dbt version is 1.10 or higher
- Modified the schema editor to place custom metrics and dimensions in the correct location based on dbt version
- Added comprehensive tests to verify the correct behavior for different dbt versions
- Updated the GitIntegrationService to pass the project's dbt version to the schema editor

These changes ensure that Lightdash can properly generate YAML files that are compatible with both older dbt versions and the newer v1.10+ format.

### DIFF
--- a/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
+++ b/packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
@@ -8,9 +8,11 @@ import {
     DbtGitlabProjectConfig,
     DbtProjectType,
     DbtSchemaEditor,
+    DbtVersionOptionLatest,
     ForbiddenError,
     friendlyName,
     getErrorMessage,
+    getLatestSupportDbtVersion,
     GitIntegrationConfiguration,
     isUserWithOrg,
     ParameterError,
@@ -20,6 +22,7 @@ import {
     SavedChart,
     SessionUser,
     snakeCaseName,
+    SupportedDbtVersions,
     UnexpectedServerError,
     VizColumn,
 } from '@lightdash/common';
@@ -58,6 +61,7 @@ type GitProps = {
     quoteChar: `"` | `'`;
     hostDomain?: string; // For GitLab or GitHub Enterprise
     type: DbtProjectType.GITHUB | DbtProjectType.GITLAB;
+    dbtVersion?: SupportedDbtVersions;
 };
 
 // Keep backward compatibility
@@ -252,7 +256,18 @@ Affected charts:
             hostDomain,
         });
 
-        const yamlSchema = new DbtSchemaEditor(fileContent, fileName);
+        // Get the dbt version from the project
+        const project = await this.projectModel.get(projectUuid);
+        const dbtVersion =
+            project.dbtVersion === DbtVersionOptionLatest.LATEST
+                ? getLatestSupportDbtVersion()
+                : project.dbtVersion;
+
+        const yamlSchema = new DbtSchemaEditor(
+            fileContent,
+            fileName,
+            dbtVersion,
+        );
 
         if (!yamlSchema.hasModels()) {
             throw new ParseError(`No models found in ${fileName}`);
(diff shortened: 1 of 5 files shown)


rephus:
* **#17025** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17025?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17025?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
```diff
@@ -475,6 +485,12 @@ Affected charts:
         )}${snakeCaseName(user.lastName)}`;
         const branchName = `lightdash-${userName}-${nanoid(4)}`;
 
+        // Get the dbt version from the project
+        const project = await this.projectModel.get(projectUuid);
+        const dbtVersion = project.dbtVersion === DbtVersionOptionLatest.LATEST
```
stevelikesmusic: Why can't we simply use the project dbt version?

packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.ts
```diff
@@ -67,6 +75,19 @@ export default class DbtSchemaEditor {
         }
     }
 
+    isDbtVersion110OrHigher(): boolean {
```
stevelikesmusic, line 78: Not for this one. But thinking ahead when there might be another breaking dbt version, what if we structured `SupportedDbtVersions` so that it was a `const` instead of an enum:
```typescript
const SupportedDbtVersions = {
  V1_4: { value: 1.4, label: 'v1.4' }
  ...
} as const;
```
then we could more easily make comparisons:

```typescript
SupportedDbtVersions[currentVersion].value >= SupportedDbtVersions.V1_10.value
```

It looks like this is the only comparison we're doing now. Just something to think about for the future if we end up doing more of these version comparisons.

packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.test.ts
```diff
@@ -61,3 +68,138 @@ describe('DbtSchemaEditor', () => {
...
+                table: 'test_table',
+                baseDimensionName: 'test_column',
+            },
+        ]);
+
+        const result = editor.toString().replace(/\s+/g, '');
+        // Check that meta is NOT inside config
+        expect(result).not.toContain('config:meta:metrics:');
+        // Check that meta is directly at column level
+        expect(result).toContain('description:Atestcolumnmeta:metrics:');
```
stevelikesmusic, line 118: This this asserting that meta is at the same level as description without any linebreaks?

stevelikesmusic (approved):

packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.test.ts
rephus: yeah, I removed extra lines and spaces to make it more consistent

packages/backend/src/services/GitIntegrationService/GitIntegrationService.ts
rephus: because dbtVersion can be "latest"

```
export enum DbtVersionOptionLatest {
    LATEST = 'latest',
}

export type DbtVersionOption = SupportedDbtVersions | DbtVersionOptionLatest;

```

We want to translate latest to a version (in this case, 1.10 or newer)

stevelikesmusic: Any reason to have have the model translate to the actual version—in the case it's `'latest'`?

rephus: I guess that field is used directly by the UI , 
image
we could have an extra function on the model that returns the version translated. But I'd rather do it on a separate PR

PR #17116, opened 2025-09-28
feat: migrate ParametersCard to Redux store

pimmey:
Closes: #17115

### Description:
This PR migrates the Parameters Card component to use Redux selectors and actions instead of the Explorer Context. Key changes include:

1. Added new Redux selectors for parameters:
   - `selectParameters`
   - `selectParameterDefinitions`
   - `selectTableName`

2. Added new Redux actions:
   - `clearAllParameters`
   - `setParameterDefinitions`

3. Updated ParametersCard to use Redux selectors while maintaining backward compatibility with Context for computed state (missingRequiredParameters)

4. Added synchronization between Context and Redux state for parameters to ensure proper transitions during migration

5. Fixed parameter handling in ParameterInput to avoid Redux immutability issues by creating copies of arrays

6. Added an effect to trigger query updates when parameters change

7. Fixed a condition in useCompiledSql to properly check for tableId existence

PR #17109, opened 2025-09-26
feat: migrate ResultsCard to use Redux store

pimmey:
Closes: #17106

### Description:
Migrates explorer state from Context to Redux for the ResultsCard component. This change moves several selectors including tableName, sorts, metricQuery, and columnOrder from the Explorer Context to Redux state. 

The PR also adds a new Redux action `setSortFields` to manage sort state and implements synchronization between Context and Redux by adding an effect in ExplorerProvider that keeps Redux sorts in sync with Context sorts.

Additionally, the useColumns hook has been updated to use the Redux selectors instead of Context for state that's now available in Redux.

PR #17104, opened 2025-09-26
refactor: migrate table name from context to Redux store

pimmey:
Closes: #17105

### Description:
Refactored SqlCard component to use the Redux store selector `selectTableName` instead of accessing the table name through the Explorer context. This change aligns with the pattern of using Redux selectors for state management in the Explorer component.

PR #17121, opened 2025-09-29
feat: add useExplorerQuery hook with tests

pimmey:
Closes: #17122

### Description:
Adds new hooks for managing Explorer query execution: `useExplorerQuery` and `useExplorerQueryActions`. These hooks provide a unified interface for handling query state, execution, and results in the Explorer component.

`useExplorerQuery` handles:
- Reading query configuration from Redux state
- Managing query execution lifecycle
- Tracking active fields and query validity
- Supporting pivoted and unpivoted query results
- Parameter validation and change detection
- Auto-fetching based on user preferences

`useExplorerQueryActions` provides utility functions for:
- Manually fetching results
- Canceling running queries

Added comprehensive tests to verify the hooks' functionality and ensure they return the expected properties and behaviors.

### DIFF
--- a/packages/frontend/src/hooks/useExplorerQuery.test.tsx
+++ b/packages/frontend/src/hooks/useExplorerQuery.test.tsx
@@ -1,0 +1,100 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { renderHook } from '@testing-library/react';
+import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { explorerStore } from '../features/explorer/store';
+import { useExplorerQuery, useExplorerQueryActions } from './useExplorerQuery';
+
+// Mock the hooks that depend on external APIs
+vi.mock('./useExplore', () => ({
+    useExplore: vi.fn(() => ({ data: null })),
+}));
+
+vi.mock('./useFeatureFlagEnabled', () => ({
+    useFeatureFlag: vi.fn(() => ({ data: { enabled: false } })),
+}));
+
+vi.mock('./parameters/useParameters', () => ({
+    useParameters: vi.fn(() => ({ data: {} })),
+}));
+
+vi.mock('../providers/Explorer/useExplorerQueryManager', () => ({
+    useQueryManager: vi.fn(() => [
+        {
+            query: { isFetched: false, isFetching: false },
+            queryResults: { queryUuid: null, totalResults: 0 },
+        },
+        vi.fn(),
+    ]),
+}));
+
+const createWrapper = () => {
+    const queryClient = new QueryClient({
+        defaultOptions: {
+            queries: { retry: false },
+            mutations: { retry: false },
+        },
+    });
+
+    return ({ children }: { children: React.ReactNode }) => (
+        <QueryClientProvider client={queryClient}>
+            <Provider store={explorerStore}>
+                <MemoryRouter>{children}</MemoryRouter>
+            </Provider>
+        </QueryClientProvider>
+    );
+};
+
+describe('useExplorerQuery', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+    });
+
+    it('should return query state and actions', () => {
+        const { result } = renderHook(() => useExplorerQuery(), {
+            wrapper: createWrapper(),
+        });
+
+        expect(result.current).toHaveProperty('query');
+        expect(result.current).toHaveProperty('queryResults');
+        expect(result.current).toHaveProperty('runQuery');
+        expect(result.current).toHaveProperty('resetQueryResults');
+        expect(result.current).toHaveProperty('getDownloadQueryUuid');
+        expect(result.current).toHaveProperty('activeFields');
+        expect(result.current).toHaveProperty('isValidQuery');
+    });
+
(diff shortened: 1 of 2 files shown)


pimmey:
* **#17129** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17129?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17121** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17121?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17116** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17116?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17109** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17109?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17104** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17104?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/useExplorerQuery.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+                wrapper: createWrapper(),
+            },
+        );
+
+        expect(result.current).toHaveProperty('fetchResults');
+        expect(result.current).toHaveProperty('cancelQuery');
+        expect(typeof result.current.fetchResults).toBe('function');
+        expect(typeof result.current.cancelQuery).toBe('function');
+    });
+});
```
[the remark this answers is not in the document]
pimmey: This was just to avoid unused exports at this point. But good point, will add more realistic tests afterwards.

packages/frontend/src/hooks/useExplorerQuery.ts
```diff
@@ -0,0 +1,374 @@
...
+    const projectUuid = projectUuidProp || projectUuidFromParams;
+
+    // Get explore data and pivot configuration
+    const { data: explore } = useExplore(tableName);
+    const { data: useSqlPivotResults } = useFeatureFlag(
+        FeatureFlags.UseSqlPivotResults,
+    );
+
+    // Compute the complete metric query (including Redux filters)
+    const computedMetricQuery = useMemo(
```
magnew, line 80: Do you think this will eventually just come from redux? Either we'd store the computed query or this would happen in a selector? Right now we are selecting both from the store, then combining them here, but the info is already in the store. Maybe there is a reason I'm not seeing that this has to happen here. 

This particular thing probably doesn't matter too much, but I'm thinking about what patterns we want to move towards. If we can store or compute as much as possible in redux it will probably be easier to follow.

packages/frontend/src/hooks/useExplorerQuery.ts
```diff
@@ -0,0 +1,374 @@
...
+        // Query configuration
+        validQueryArgs,
+        missingRequiredParameters,
+
+        // State derived from Redux
+        activeFields,
+        isValidQuery,
+
+        // Parameter management
+        setParameterReferences,
```
magnew, line 333: This hook returns a lot of things. And retains a lot of state. It starts to seem like a mini context -- which it kind of is cause I know a lot of this is getting pulled out of the ExploreProvider. Some of this is necessary cause it's about managing reactquery, but some of it feels like it should get moved (eventually). Like the setParamterReferences feels a little out of place. As a consumer of this hook, it's not clear when I would use that or if I should be setting it in the store. 

I wonder if, after the migration, we couldnt make this hook have a clearer separation of concerns.

packages/frontend/src/hooks/useExplorerQuery.ts
```diff
@@ -0,0 +1,374 @@
...
+        } else {
+            setUnpivotedQueryArgs(null);
+        }
+    }, [validQueryArgs, needsUnpivotedData, isResultsOpen]);
+
+    // Auto-fetch logic
+    useEffect(() => {
+        if ((!autoFetchEnabled || !query.isFetched) && isEditMode) return;
+        runQuery();
+    }, [runQuery, autoFetchEnabled, isEditMode, query.isFetched]);
```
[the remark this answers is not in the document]
magnew: I *think* this is wrong and will break things 😄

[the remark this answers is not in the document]
pimmey: Did you mean the AI suggestion or actual logic @magnew? 😄

packages/frontend/src/hooks/useExplorerQuery.ts
pimmey: Yeah absolutely, if we can extract it from the store, we will do that, instead of keeping it in the hook. I haven't quite reached the "full picture" stage, where I can clean up and move things around 😄

packages/frontend/src/hooks/useExplorerQuery.ts
pimmey: 💯 being addressed in the upcoming PR #17129 


> I wonder if, after the migration, we couldnt make this hook have a clearer separation of concerns.

Again, agree.

Issue #17123, opened 2025-09-29
Adding tabs to a dashboard can break existing tab layout


magnew:
### Description

1. Create a dashboard with a few tiles, no tabs
2. Save
3. Edit the dashboard and create a second tab
4. When clicking back to the first tab while still in edit mode the charts disappear
5. If you click the first tab again, they reappear, but layout is broken

Note: this only happens the first time tabs are added

Also, there is some weirdness with clicking to other tabs -- it seems to require clicking twice while in edit mode. 

### Steps to Reproduce the Bug or Issue

See above

### Cloud or self-hosting

None

PR #17140, opened 2025-09-30
fix: adding dashboard tabs can break layout

magnew:
Closes: #17123 

### Description:

**The bug:** adding new tabs and navigating between them and existing tabs had lots of issues. One manifestation was that when tabs were first created, navigating to the first tab before saving broke the layout there since new tabs weren't set up properly. 

**Repro steps from the ticket:**
1. Create a dashboard with a few tiles, no tabs
2. Save
3. Edit the dashboard and create a second tab
4. When clicking back to the first tab while still in edit mode the charts disappear
5. If you click the first tab again, they reappear, but layout is broken

**What was wrong:** 1) We were validating tab navigation only against saved tabs (now we validate against new and saved) 2) We were setting the tab in the URL separately from in the context (now we always get the active tab from the URL)

**Before (bug)**
before

**Fixed**
after

### DIFF
--- a/packages/frontend/src/components/DashboardTabs/index.tsx
+++ b/packages/frontend/src/components/DashboardTabs/index.tsx
@@ -41,9 +41,6 @@ type DashboardTabsProps = {
     handleDeleteTile: (tile: IDashboard['tiles'][number]) => Promise<void>;
     handleBatchDeleteTiles: (tile: IDashboard['tiles'][number][]) => void;
     handleEditTile: (tiles: IDashboard['tiles'][number]) => void;
-    setActiveTab: (
-        value: React.SetStateAction<DashboardTab | undefined>,
-    ) => void;
     setAddingTab: (value: React.SetStateAction<boolean>) => void;
     setGridWidth: (value: React.SetStateAction<number>) => void;
 };
@@ -54,7 +51,6 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
     addingTab,
     dashboardTiles,
     activeTab,
-    setActiveTab,
     handleAddTiles,
     handleUpdateTiles,
     handleDeleteTile,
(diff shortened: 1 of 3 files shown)


magnew:
* **#17140** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/DashboardTabs/index.tsx
```diff
@@ -144,6 +140,21 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
...
+            {
+                pathname: isEditMode
+                    ? `/projects/${projectUuid}/dashboards/${dashboardUuid}/edit/tabs/${tab?.uuid}`
+                    : `/projects/${projectUuid}/dashboards/${dashboardUuid}/view/tabs/${tab?.uuid}`,
+                search: newParams.toString(),
+            },
+            { replace: true },
+        );
+    };
+
```
almeidabbm, line 157: What happens to `activeTab` now that we're no longer updating it via `setActiveTab`

almeidabbm (approved):
Works very very well!! 🚀 just have a question

**Before**

https://github.com/user-attachments/assets/54042bff-0b9f-44a3-b4ad-725559692e7a

**After**

https://github.com/user-attachments/assets/7915ed57-e222-464d-9cf8-00e1615ddd87

packages/frontend/src/components/DashboardTabs/index.tsx
magnew: It gets set in the provider (line 234 in this PR), which syncs it to the URL. I think the original bugs here were because we were setting it, then re-directing, then it was getting set again there.

almeidabbm: Gotcha!! just saw it 🚀

Issue #17134, opened 2025-09-30
Re-authenticating to snowflake re-creates warehouse credentials for user


owlas:
When users re-auth into snowflake it's creating a duplicate entry in their warehouse connections, we should only have one.

PR #17148, opened 2025-10-01
fix: duplicate snowflake warehouse credentials

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/17134

Before
Screenshot from 2025-10-01 10-28-26

After
Screenshot from 2025-10-01 10-31-56
Screenshot from 2025-10-01 10-27-50

### Description:
Added a method to delete all warehouse credentials for a user by warehouse type and implemented it in the Snowflake authentication flow to prevent duplicate credentials. The new `deleteAllByUserAndWarehouseType` method is called before creating new Snowflake credentials during authentication, ensuring users don't accumulate multiple sets of credentials for the same warehouse type.

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -1715,6 +1715,12 @@ export class UserService extends BaseService {
         user: SessionUser,
         refreshToken: string,
     ) {
+        // Remove old Snowflake credentials to prevent duplicates on re-authentication
+        await this.userWarehouseCredentialsModel.deleteAllByUserAndWarehouseType(
+            user.userUuid,
+            WarehouseTypes.SNOWFLAKE,
+        );
+
         const snowflakeCredentials: UpsertUserWarehouseCredentials = {
             name: 'Default',
             credentials: {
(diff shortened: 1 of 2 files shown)


rephus:
* **#17148** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/UserService.ts
```diff
@@ -1715,6 +1715,12 @@ export class UserService extends BaseService {
         user: SessionUser,
         refreshToken: string,
     ) {
+        // Remove old Snowflake credentials to prevent duplicates on re-authentication
+        await this.userWarehouseCredentialsModel.deleteAllByUserAndWarehouseType(
```
stevelikesmusic, line 1719: Do we care if the credentials creation fails after we've deleted everything? I supposed they'd just try to create new creds again? Is it worth deleting everything but the newly created creds on successful creation—maybe overkill?

rephus: I don't think this should cause any issues. If you are trying to relogin it is likely because the credentials were not working or expired in the first place, so we don't need to keep them anyway.

Issue #17160, opened 2025-10-01
Updating tile target before dashboard upload results tile not found


almeidabbm:
### Description

If we update the `tileSlug` (both in tile targets and tiles) in a dashboard yaml file and keep the `chartSlug` the same, it results in `Tile with slug {slug} not found in tilesWithUuids` error.

This is because we're trying to find the tile by checking the target `tileSlug` against `chartSlug`

### Steps to Reproduce the Bug or Issue

1. Download a dashboard with filters
2. Update tile target `tileSlug` and tile `tileSlug` but keep `chartSlug` the same.
3. Run `lightdash upload --force`

### Cloud or self-hosting

None

PR #17161, opened 2025-10-01
fix: find tile targets first by tileSlug

almeidabbm:
Closes: #17160

### Description:
Added support for matching dashboard tiles by `tileSlug` in addition to `chartSlug` when converting dashboard filters from code to database format. This fixes an issue where filters weren't properly associated with tiles that changed slug.

This would effectively remove the tile target from the filter.

The changes include:
- Created a new type `DashboardTileWithSlug` that extends `DashboardTile` with a `tileSlug` property
- Updated the filter matching logic to check both `tileSlug` and `chartSlug` when finding the corresponding tile
- Updated return types in relevant functions to use the new type

**Steps to reproduce**

1. Download a dashboard with filters
2. Update tile target `tileSlug` and tile `tileSlug` but keep `chartSlug` the same.
3. Run `lightdash upload --force`

**Test yaml**
Notice `tileSlug` being `saved-in-dashboard-how-much-revenue-do-we-have-per-payment-method-1` while `chartSlug` is different

```yaml
name: Dashboard with dashboard charts
description: null
updatedAt: "2025-10-01T12:23:21.376Z"
tiles:
  - x: 0
    "y": 0
    h: 9
    w: 24
    tabUuid: null
    type: saved_chart
    properties:
      title: ""
      hideTitle: false
      chartSlug: saved-in-dashboard-how-much-revenue-do-we-have-per-payment-method
      chartName: "[Saved in dashboard] How much revenue do we have per payment method?"
    tileSlug: saved-in-dashboard-how-much-revenue-do-we-have-per-payment-method-1
filters:
  metrics: []
  dimensions:
    - target:
        fieldId: payments_payment_method
        fieldName: payment_method
        tableName: payments
      values:
        - bank_transfer
      disabled: false
      operator: equals
      tileTargets:
        saved-in-dashboard-how-much-revenue-do-we-have-per-payment-method-1:
          fieldId: payments_payment_method
          tableName: payments
  tableCalculations: []
tabs: []
slug: dashboard-with-dashboard-charts
spaceSlug: jaffle-shop
version: 1
downloadedAt: "2025-10-01T12:23:32.454Z"
```

**Before**
image

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.ts
+++ b/packages/backend/src/services/CoderService/CoderService.ts
@@ -30,6 +30,7 @@ import {
     SpaceMemberRole,
     SpaceSummary,
     UpdatedByUser,
+    type DashboardTileWithSlug,
 } from '@lightdash/common';
 import { v4 as uuidv4 } from 'uuid';
 import { LightdashAnalytics } from '../../analytics/LightdashAnalytics';
@@ -209,7 +210,7 @@ export class CoderService extends BaseService {
      */
     static getFiltersWithTileUuids(
         dashboardAsCode: DashboardAsCode,
-        tilesWithUuids: DashboardTile[],
+        tilesWithUuids: DashboardTileWithSlug[],
     ): DashboardDAO['filters'] {
         const dimensionFiltersWithUuids: DashboardDAO['filters']['dimensions'] =
             dashboardAsCode.filters.dimensions.map((filter) => {
@@ -220,7 +221,9 @@ export class CoderService extends BaseService {
                         const tileUuid = tilesWithUuids.find(
                             (t) =>
                                 isAnyChartTile(t) &&
-                                t.properties.chartSlug === tileSlug,
+                                // Match first by tileSlug, then by chartSlug (for the case of tile not having a slug)
+                                (t.tileSlug === tileSlug ||
+                                    t.properties.chartSlug === tileSlug),
                         )?.uuid;
                         if (!tileUuid) {
                             console.error(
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#17161** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
That `content.cy.ts` api test error seems unrelated

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -220,7 +221,9 @@ export class CoderService extends BaseService {
                         const tileUuid = tilesWithUuids.find(
                             (t) =>
                                 isAnyChartTile(t) &&
-                                t.properties.chartSlug === tileSlug,
+                                // Match first by tileSlug, then by chartSlug (for the case of tile not having a slug)
+                                (t.tileSlug === tileSlug ||
```
rephus, line 225: :+1:

packages/common/src/types/coder.ts
```diff
@@ -71,6 +71,10 @@ export type DashboardTileAsCode = Omit<DashboardTile, 'properties' | 'uuid'> & {
         | DashboardLoomTileProperties['properties'];
 };
 
+export type DashboardTileWithSlug = DashboardTile & {
```
rephus, line 74: why not adding an optional `tileSlug?: string` to DashboardTile instead ?  

as long as we keep it optional so it is  backwards compatible, it should be fine to modify existing types.

Whatever you find more readable and scalable. this also works.

almeidabbm: Yeah I thought about that but felt like it could be misinterpreted since `tileSlugs` don't exist at all outside the coder service. This makes it a bit more contained 🚀

PR #17203, opened 2025-10-02
feat: handle table label & description updates

joaoviana:
Closes: N/A

### Description:
Added support for table entity updates in the CatalogModel. Previously, the model only handled dimension and metric updates, but now it can also process table-level changes. This includes:

1. Adding `CompiledTable` as a possible field type to update
2. Adding a check for table entity type
3. Restructuring the update logic to handle tables separately from dimensions and metrics
4. Adding a type filter to the database query to distinguish between table and field updates

This enhancement allows the catalog to properly update table metadata such as labels and descriptions.

### DIFF
--- a/packages/backend/src/models/CatalogModel/CatalogModel.ts
+++ b/packages/backend/src/models/CatalogModel/CatalogModel.ts
@@ -6,12 +6,14 @@ import {
     ChangesetWithChanges,
     CompiledDimension,
     CompiledMetric,
+    CompiledTable,
     Explore,
     FieldType,
     NotFoundError,
     TableSelectionType,
     UNCATEGORIZED_TAG_UUID,
     UnexpectedServerError,
+    assertUnreachable,
     convertToAiHints,
     isExploreError,
     type ApiCatalogSearch,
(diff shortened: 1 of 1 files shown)


joaoviana:
* **#17214** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17203** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -245,31 +246,52 @@ export class CatalogModel {
...
+                                        | 'metrics'
+                                        | undefined;
+                                    if (change.entityType === 'metric') {
+                                        key = 'metrics';
+                                    } else if (
+                                        change.entityType === 'dimension'
+                                    ) {
+                                        key = 'dimensions';
+                                    } else {
+                                        key = undefined;
```
tatianainama: _*Irakli voice coming from another dimension*: use `assertUnreachable`_

packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -245,31 +246,52 @@ export class CatalogModel {
...
-                                | CompiledMetric;
+                                | CompiledMetric
+                                | CompiledTable;
+                            const isTable = change.entityType === 'table';
+
                             if (
                                 cachedExploreTable.tables[
                                     change.entityTableName
                                 ]
                             ) {
```
notgiorgi: reverse this and early-return to simplify

packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -245,31 +246,52 @@ export class CatalogModel {
...
+                                    }
+
+                                    const table =
+                                        cachedExploreTable.tables[
+                                            change.entityTableName
+                                        ];
+
+                                    if (!table) {
+                                        return null;
+                                    }
```
notgiorgi: this is the same check as the above if, can share variable/check (early return)

packages/backend/src/models/CatalogModel/CatalogModel.ts
```diff
@@ -245,31 +246,52 @@ export class CatalogModel {
...
+                                        key = 'metrics';
+                                    } else if (
+                                        change.entityType === 'dimension'
+                                    ) {
+                                        key = 'dimensions';
+                                    } else {
+                                        key = undefined;
+                                    }
+                                    if (!key) {
+                                        return null;
```
notgiorgi: NP

```suggestion
                               const entityTypeToFieldKey: Record<
                                    string,
                                    'metrics' | 'dimensions'
                                > = {
                                    metric: 'metrics',
                                    dimension: 'dimensions',
                                };
                                const fieldKey =
                                    entityTypeToFieldKey[change.entityType];

                                if (!fieldKey) {
                                    return null;
                                }
```

packages/backend/src/models/CatalogModel/CatalogModel.ts
tatianainama, line 1: I think we can simplify this logic a lot and make it more legible with something like the following:
``` ts
let fieldToUpdate:
    | CompiledDimension
    | CompiledMetric
    | CompiledTable;
const isTable = change.entityType === 'table';
const table = cachedExploreTable.tables[
    change.entityTableName
]
if (!table) return null
switch(change.entityType) {
    case 'table': {
        fieldToUpdate = table
        break;
    }
    case 'dimension': {
        fieldToUpdate = table.dimensions[change.entityName]
        break;
    }
    case 'metric': {
        fieldToUpdate = table.metrics[change.entityName]
        break
    }
    default: 
        assertUnreachable(change.entityType, `Unknown entity type ${change.entityType}`)
}

```

wdyt?

tatianainama (approved):
it works! just left a nit to make this a bit more legible

PR #17267, opened 2025-10-07
feat: add embed configuration to instance setup

rephus:
Closes: #1234

### Description:
Adds support for configuring embedded dashboards during instance setup using environment variables. This PR introduces two new environment variables:

- `LD_SETUP_EMBED_ALLOW_ALL_DASHBOARDS`: When set to 'true', allows embedding of all dashboards
- `LD_SETUP_EMBED_SECRET`: Sets the secret key used for embedding authentication

The implementation updates the `InstanceConfigurationService` to handle these new settings during initial setup and instance configuration updates, with appropriate validation to ensure an admin email is provided when setting an embed secret.


rephus:
* **#17267** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17267?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17267?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/config/parseConfig.ts
```diff
@@ -477,6 +477,11 @@ export const getUpdateSetupConfig = (): LightdashConfig['updateSetup'] => {
         dbt: {
             personal_access_token: process.env.LD_SETUP_GITHUB_PAT,
         },
+        embed: {
+            allowAllDashboards:
+                process.env.LD_SETUP_EMBED_ALLOW_ALL_DASHBOARDS === 'true',
+            secret: process.env.LD_SETUP_EMBED_SECRET,
```
stevelikesmusic, line 483: Is this a different value than the `LIGHTDASH_SECRET`? Can we not leverage the existing secret?

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
```diff
@@ -501,6 +520,64 @@ export class InstanceConfigurationService extends BaseService {
...
+        if (allowAllDashboards === undefined && !secret) return;
+
+        try {
+            const projectUuid = await this.getSingleProject();
+
+            // If config embed secret is provided, we need to call .save to override the config
+            // This requires a user UUID, we get it from the admin email
+            if (secret) {
+                let userUuid: string | undefined;
+                const adminEmail = config.organization?.admin?.email;
```
stevelikesmusic, line 538: In our ideal world, would this identity be a service account rather than the admin? 

Also, `created_by` in the embed table is a nullable field. Do we _need_ to pass a user ID?

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
```diff
@@ -501,6 +520,64 @@ export class InstanceConfigurationService extends BaseService {
...
+                );
+                await this.embedModel.updateDashboards(projectUuid, {
+                    dashboardUuids: [],
+                    allowAllDashboards,
+                });
+            }
+        } catch (error) {
+            this.logger.error(`Error updating embed settings: ${error}`);
+        }
+    }
```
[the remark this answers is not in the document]
stevelikesmusic: Should we make some kind of permissions check here?

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
```diff
@@ -501,6 +520,64 @@ export class InstanceConfigurationService extends BaseService {
...
+    ) {
+        if (!config.embed || !this.embedModel) return;
+
+        const { allowAllDashboards, secret } = config.embed;
+        if (allowAllDashboards === undefined && !secret) return;
+
+        try {
+            const projectUuid = await this.getSingleProject();
+
+            // If config embed secret is provided, we need to call .save to override the config
```
stevelikesmusic: Are we overriding the config or the embed record?

stevelikesmusic (approved):
Looking good Javi! Just left some general questions.

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
rephus: > In our ideal world, would this identity be a service account rather than the admin?

Maybe, if we change this table to also use an account id rather than an user id

> Also, created_by in the embed table is a nullable field. Do we need to pass a user ID?

on .save, the userUuid is mandatory, and I'd rather not change this, even if the table accepts null ,we don't know if some service logic might break if this user is null (it should be ok, but...)

packages/backend/src/services/InstanceConfigurationService/InstanceConfigurationService.ts
rephus: by config I mean the embed record config... is that confusing ? Sorry, I wrote that not claude

packages/backend/src/config/parseConfig.ts
rephus: It is a different value, and this one can be exposed in the UI, while the secret is used for encryption and can't be exposed, so different purposes.

Issue #17273, opened 2025-10-07
AI agent visualization errors with "Tried to reference a metric with unknown field ID"


IrakliJani:
Some users report that sometimes visualization fails to load AI agent visualization with error - "Tried to reference a metric with unknown field ID"

PR #17281, opened 2025-10-07
feat: validate field entity types in AI visualization tools

IrakliJani:
Closes: https://github.com/lightdash/lightdash/issues/17273

### Description:
Improves field validation in AI tools by using more specific validation functions. This PR replaces generic field existence checks with type-specific validation (`validateFieldEntityType`) to ensure dimensions and metrics are used correctly in their respective contexts. The changes affect multiple visualization tools (table, bar, time series) and the propose change functionality.

The main improvement is that we now validate that fields are of the correct type (dimension or metric) rather than just checking if they exist, which prevents misuse of fields in inappropriate contexts.

### DIFF
--- a/packages/backend/src/ee/services/ai/tools/proposeChange.ts
+++ b/packages/backend/src/ee/services/ai/tools/proposeChange.ts
@@ -18,7 +18,6 @@ import { toModelOutput } from '../utils/toModelOutput';
 import { toolErrorHandler } from '../utils/toolErrorHandler';
 import {
     validateFieldEntityType,
-    validateSelectedFieldsExistence,
     validateTableNames,
 } from '../utils/validators';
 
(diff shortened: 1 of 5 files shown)


IrakliJani:
* **#17281** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

IrakliJani:
before:

IrakliJani:
after:

packages/backend/src/ee/services/ai/tools/proposeChange.ts
```diff
@@ -140,13 +139,17 @@ export const getProposeChange = ({
...
+                                    explore,
+                                    [
+                                        getItemId({
+                                            table: entityTableName,
+                                            name: change.value.value.metric
+                                                .baseDimensionName,
+                                        }),
+                                    ],
+                                    'dimension',
+                                );
```
tatianainama, line 152: I'm a bit lost here, we only support metric creation at the moment, don't we? we should remove the `create` case from here, and I don't think we need to validate the entity type, as it should not exist?

IrakliJani: Great question! Yes, but those metrics are based on dimensions, so we need to validate that the dimension the metric is based on exists in the explore.

tatianainama: ah got it! Thanks for the explanation 🙏

IrakliJani:
### Merge activity

* **Oct 7, 2:51 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17281).
* **Oct 7, 2:51 PM UTC**: @IrakliJani merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17281).

PR #17292, opened 2025-10-07
feat: allow fetching dashboards and charts by slug

stevelikesmusic:
Closes: CENG-69

### Description:
This PR enhances dashboard and saved chart access by supporting slug-based URLs in addition to UUIDs. The implementation replaces `getById` with `getByIdOrSlug` across the codebase, which intelligently determines whether the provided identifier is a UUID or slug and queries the database accordingly.

This change enables more user-friendly URLs for dashboards and charts, allowing them to be accessed via their slugs (e.g., `/dashboards/my-sales-dashboard`) rather than requiring UUIDs.

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -8,7 +8,6 @@ import {
     DashboardDAO,
     DashboardLoomTile,
     DashboardMarkdownTile,
-    DashboardParameterValue,
     DashboardSqlChartTile,
     DashboardTab,
     DashboardTileTypes,
@@ -32,7 +31,7 @@ import {
     type DashboardParameters,
 } from '@lightdash/common';
 import { Knex } from 'knex';
-import { v4 as uuidv4 } from 'uuid';
+import { validate as isValidUuid, v4 as uuidv4 } from 'uuid';
 import {
     DashboardTable,
     DashboardTabsTableName,
@@ -659,8 +658,8 @@ export class DashboardModel {
         );
     }
 
-    async getById(dashboardUuid: string): Promise<DashboardDAO> {
-        const [dashboard] = await this.database(DashboardsTableName)
+    async getByIdOrSlug(dashboardUuidOrSlug: string): Promise<DashboardDAO> {
+        const query = this.database(DashboardsTableName)
             .leftJoin(
                 DashboardVersionsTableName,
                 `${DashboardsTableName}.dashboard_id`,
@@ -722,9 +721,29 @@ export class DashboardModel {
                 `${DashboardsTableName}.views_count`,
                 `${DashboardsTableName}.first_viewed_at`,
             ])
-            .where(`${DashboardsTableName}.dashboard_uuid`, dashboardUuid)
             .orderBy(`${DashboardVersionsTableName}.created_at`, 'desc')
             .limit(1);
+
+        if (isValidUuid(dashboardUuidOrSlug)) {
+            void query.where((builder) => {
+                void builder
+                    .where(
+                        `${DashboardsTableName}.dashboard_uuid`,
+                        dashboardUuidOrSlug,
+                    )
+                    .orWhere(
+                        `${DashboardsTableName}.slug`,
+                        dashboardUuidOrSlug,
+                    );
+            });
+        } else {
+            void query.where(
+                `${DashboardsTableName}.slug`,
+                dashboardUuidOrSlug,
+            );
+        }
+
+        const [dashboard] = await query;
 
         if (!dashboard) {
             throw new NotFoundError('Dashboard not found');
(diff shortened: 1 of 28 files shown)


stevelikesmusic:
* **#17292** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -238,11 +238,13 @@ export class DashboardService
         });
     }
 
-    async getById(
+    async getByIdOrSlug(
```
almeidabbm, line 241: Good use for renaming, forcing to use `getByIdOrSlug` everywhere so that we don't miss a spot 🙌

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -786,6 +787,9 @@ export class SavedChartModel {
                 name: 'SavedChartModel.get',
             },
             async () => {
+                const isUuid = isValidUuid(savedChartUuidOrSlug);
+                const filterField = isUuid ? 'saved_query_uuid' : 'slug';
```
almeidabbm, line 791: Low chance of happening but what if the chartSlug is a uuid?
Wouldn't it be better to just do `orWhere`?

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -722,7 +724,7 @@ export class DashboardModel {
                 `${DashboardsTableName}.views_count`,
                 `${DashboardsTableName}.first_viewed_at`,
             ])
-            .where(`${DashboardsTableName}.dashboard_uuid`, dashboardUuid)
+            .where(`${DashboardsTableName}.${filterField}`, dashboardUuidOrSlug)
```
almeidabbm: Same thing as with the chart one 🤔

packages/backend/src/models/SavedChartModel.ts
stevelikesmusic: Valid point! Yeah, that's a safer route with no drawbacks.

packages/frontend/src/features/comments/hooks/useComments.ts
```diff
@@ -47,7 +49,13 @@ export const useCreateComment = () => {
     >((data) => createDashboardTileComment(data), {
         mutationKey: ['create-comment'],
         onSuccess: async (_, { dashboardUuid }) => {
-            await queryClient.invalidateQueries(['comments', dashboardUuid]);
+            await Promise.all([
```
stevelikesmusic, line 52: We're adding an additional change to the query client because of our dual state. Sometimes we're using the route params (which might be `slug`), and sometimes we defer to a standard identifier (`dashboard.uuid`). We need to handle both in the hook for the UI to properly update after a successful mutation.

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -876,13 +880,28 @@ export class SavedChartModel {
...
+                                `${SavedChartsTableName}.slug`,
+                                savedChartUuidOrSlug,
+                            );
+                    });
+                } else {
+                    void chartQuery.where(
+                        `${SavedChartsTableName}.slug`,
+                        savedChartUuidOrSlug,
+                    );
+                }
```
almeidabbm, line 903: [nitpick] wouldn't this be the same as just having:

```suggestion
                void chartQuery.where((builder) => {
                    void builder
                        .where(
                            `${SavedChartsTableName}.saved_query_uuid`,
                            savedChartUuidOrSlug,
                        )
                        .orWhere(
                            `${SavedChartsTableName}.slug`,
                            savedChartUuidOrSlug,
                        );
                });
```

without the need for checking for uuid validity?

packages/backend/src/models/DashboardModel/DashboardModel.ts
```diff
@@ -1070,14 +1089,25 @@ export class DashboardModel {
...
+
+        if (isValidUuid(dashboardUuidOrSlug)) {
+            void query.where((builder) => {
+                void builder
+                    .where('dashboard_uuid', dashboardUuidOrSlug)
+                    .orWhere('slug', dashboardUuidOrSlug);
+            });
+        } else {
+            void query.where('slug', dashboardUuidOrSlug);
+        }
```
almeidabbm, line 1106: same nit

almeidabbm (approved):

packages/backend/src/models/SavedChartModel.ts
stevelikesmusic: I wish! The problem is that the database does a UUID validation on filters that are a UUID field. So if we pass `some-dashboard-slug` to the `where uuid =` clause, postgres will throw an type error. We have to only send the value to a UUID field in the DB if we're certain it's that type.

almeidabbm: I see!! That makes total sense, didn't take that into consideration, merge away 🚀

stevelikesmusic:
### Merge activity

* **Oct 9, 1:08 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17292).
* **Oct 9, 1:08 PM UTC**: @stevelikesmusic merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17292).

PR #17329, opened 2025-10-08
feat: add bar chart display style for table columns

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-64/cell-visualization-bar-in-table

This is a POC: the currenct configuration is very simple, just a list of fields, in the future, we can add different color, values, etc. 

PR for making this work with pivot tables https://github.com/lightdash/lightdash/pull/17366

Compatible with charts as code
Can be visualized in dashboards too

Screenshot from 2025-10-09 09-10-47

Screenshot from 2025-10-09 09-17-13

Screenshot from 2025-10-09 09-09-55

Negative numbers simply don't show up (I think that works for now), we can change this later

image

### Description:
Added bar chart display style for table columns. This feature allows numeric columns to be visualized as horizontal bars alongside their text values.

Key changes:
- Extended `VizColumnConfig` type to include `displayStyle` option ('text' or 'bar')
- Added bar configuration options for customizing min/max values and colors
- Implemented column statistics calculation to automatically determine appropriate scaling
- Enhanced cell rendering to support bar visualization while preserving the numeric value

Bar chart in table columns

### DIFF
--- a/packages/frontend/src/components/DataViz/hooks/useTableDataModel.ts
+++ b/packages/frontend/src/components/DataViz/hooks/useTableDataModel.ts
@@ -13,6 +13,7 @@ import { useVirtualizer } from '@tanstack/react-virtual';
 import { useCallback, useMemo, useRef } from 'react';
 import { getValueCell } from '../../../hooks/useColumns';
 import { ROW_HEIGHT_PX } from '../../common/Table/Table.styles';
+import { calculateColumnStats } from '../utils/columnStats';
 
 // TODO: this name could change or we could replace this with useVirtualTable.
 // It's not really clear what is doing with the table data model for a consumer.
@@ -33,6 +34,20 @@ export const useTableDataModel = ({
     const columns = useMemo(() => tableModel.getVisibleColumns(), [tableModel]);
     const rows = useMemo(() => tableModel.getRows(), [tableModel]);
 
+    // Calculate stats for columns with bar display style
+    const columnStats = useMemo(() => {
+        if (!config?.columns) return {};
+
+        // Find columns that need bar chart display
+        const barColumns = columns.filter(
+            (col) => config?.columns?.[col]?.displayStyle === 'bar',
+        );
+
+        if (barColumns.length === 0) return {};
+
+        return calculateColumnStats(rows, barColumns);
+    }, [rows, columns, config?.columns]);
+
     const tanstackColumns: ColumnDef<RawResultRow, any>[] = useMemo(() => {
         return columns.map((column) => ({
             id: column,
@@ -49,6 +64,10 @@ export const useTableDataModel = ({
         data: rows,
         columns: tanstackColumns,
         getCoreRowModel: getCoreRowModel(),
+        meta: {
+            columnStats,
+            columnsConfig: config?.columns ?? undefined,
+        },
     });
 
     const getRowHeight = useCallback(() => ROW_HEIGHT_PX, []);
--- a/packages/frontend/src/components/DataViz/utils/columnStats.ts
+++ b/packages/frontend/src/components/DataViz/utils/columnStats.ts
@@ -1,0 +1,44 @@
+import { type RawResultRow } from '@lightdash/common';
+
+export type ColumnStats = {
+    min: number;
+    max: number;
+};
+
+export type ColumnStatsMap = Record<string, ColumnStats>;
+
+/**
+ * Calculate min/max statistics for numeric columns
+ * Used for scaling bar chart displays
+ */
+export function calculateColumnStats(
+    rows: RawResultRow[],
+    columnReferences: string[],
+): ColumnStatsMap {
+    const stats: ColumnStatsMap = {};
+
+    if (rows.length === 0) return stats;
+
+    // Initialize with first row values
+    columnReferences.forEach((colRef) => {
+        // Initialize stats for all columns with default values
+        stats[colRef] = { min: Infinity, max: -Infinity };
+    });
+
+    // Scan all rows to find min/max
+    rows.forEach((row) => {
+        columnReferences.forEach((colRef) => {
+            const value = Number(row[colRef]);
+            if (!Number.isNaN(value) && stats[colRef]) {
+                if (value < stats[colRef].min) {
+                    stats[colRef].min = value;
+                }
+                if (value > stats[colRef].max) {
+                    stats[colRef].max = value;
+                }
+            }
+        });
+    });
+
+    return stats;
+}
--- a/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
+++ b/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
@@ -4,7 +4,6 @@ import {
     isCustomDimension,
     isDimension,
     isField,
-    isFilterableItem,
     isMetric,
     isNumericItem,
     isTableCalculation,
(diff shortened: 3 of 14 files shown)


rephus:
* **#17367** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17367?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17366** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17366?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17329** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17329?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17329?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hamzahc1:
@rephus, can the charts and numbers be left-aligned? They look a bit odd centred! 
example of what I mean it should look like: 
CleanShot 2025-10-09 at 17 35 37@2x

rephus:
> @rephus, can the charts and numbers be left-aligned? They look a bit odd centred! example of what I mean it should 

Yeah, done 
image

packages/frontend/src/components/DataViz/hooks/useTableDataModel.ts
```diff
@@ -49,6 +64,10 @@ export const useTableDataModel = ({
         data: rows,
         columns: tanstackColumns,
         getCoreRowModel: getCoreRowModel(),
+        meta: {
```
stevelikesmusic, line 67: Is meta just "everything else"?

packages/frontend/src/components/DataViz/utils/columnStats.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    if (rows.length === 0) return stats;
+
+    // Initialize with first row values
+    columnReferences.forEach((colRef) => {
+        const firstValue = rows[0]?.[colRef];
+        if (typeof firstValue === 'number') {
+            stats[colRef] = { min: firstValue, max: firstValue };
+        }
+    });
```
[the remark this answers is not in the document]
stevelikesmusic: I agree with this one for initialization. If we don't set a value on init, then all checks will fail when comparing a number to `undefined`. That will be a tricky bug because of how JavaScript handles number comparison with `undefined` (always false) rather than throw a type error.

packages/frontend/src/components/DataViz/utils/columnStats.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    // Scan all rows to find min/max
+    rows.forEach((row) => {
+        columnReferences.forEach((colRef) => {
+            const value = row[colRef];
+            if (typeof value === 'number' && stats[colRef]) {
```
stevelikesmusic: Would it be worth parsing a number? What if for some reason we get stringified numbers?

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -464,22 +463,41 @@ const useTableConfig = (
...
 
     const minMaxMap = useMemo(() => {
-        if (
-            !itemsMap ||
-            !resultsData ||
-            resultsData.rows.length === 0 ||
-            !conditionalFormattings ||
-            conditionalFormattings.length === 0
-        ) {
+        // Calculate min/max for columns that need it (conditional formatting or bar charts)
```
stevelikesmusic: This comment seems incorrect? Maybe it's for the whole block—if that's the case, this may be better pulled into a function?

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -464,22 +463,41 @@ const useTableConfig = (
...
             .filter(
-                ([_, field]) => isNumericItem(field) && isFilterableItem(field),
+                ([fieldId, field]) =>
+                    fieldsNeedingMinMax.has(fieldId) &&
+                    isNumericItem(field) &&
+                    isColumnVisible(fieldId) &&
+                    fieldId in resultsData.rows[0],
             )
-            .filter(([fieldId]) => isColumnVisible(fieldId))
-            .filter(([fieldId]) => fieldId in resultsData.rows[0])
```
stevelikesmusic, line 482: Do we prefer one complex filter or chained filters? An earlier portion of the PR used chained filters. No strong opinions, just curious.

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -62,10 +62,125 @@ export const formatCellContent = (data?: { value: ResultValue }) => {
 
 export const getFormattedValueCell = (
     info: CellContext<ResultRow, { value: ResultValue }>,
-) => formatCellContent(info.getValue());
+) => {
+    const cellValue = info.getValue();
```
stevelikesmusic, line 83: This block should be its own function or a component that can be returned.

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -62,10 +62,125 @@ export const formatCellContent = (data?: { value: ResultValue }) => {
...
+
+    return formatCellContent(cellValue);
+};
 
 export const getValueCell = (info: CellContext<RawResultRow, string>) => {
     const value = info.getValue();
+    const columnId = info.column.id;
+
+    // Check if this column has bar display style
+    const columnConfig = info.table.options.meta?.columnsConfig?.[columnId];
```
stevelikesmusic: This should get pulled into its own function or component.

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
rephus: I think I prefer nested filters... I'll update it

rephus:
> Bar type don't update with conditional formatting right away

@stevelikesmusic 
The chart was pivoted, by switching to another chart type, you removed the pivoting, making it work on this PR. I have another PR that adds support to pivot tables https://github.com/lightdash/lightdash/pull/17366 But I don't expect it to work here

rephus:
As I mentioned, at the moment we don't support negative numbers, because it is visually confusing, but we might support it in the future if we find a way to make it different 
Screenshot from 2025-10-10 09-58-27

packages/frontend/src/components/DataViz/utils/columnStats.ts
rephus: well spotted, the issue you mentioned about the last column, is related, for some reasons, "metric counts" are returned as a string. 


                },
                "payments_unique_payment_count": {
                    "value": {
                        "raw": "1",
                        "formatted": "1"
                    }
                }

stevelikesmusic:
Certainly not for this PR, but it'd be cool to center 0 if we have negative numbers. Then the bar chart functions closer to a number line so it's easy to see negative and positive.

packages/frontend/src/hooks/useColumns.tsx
rephus: I'm not sure I understand the change you want, is this for readability ? or performance, could you provide an example ?

stevelikesmusic: For readability. The bulk of this function was very simple. Now the function is primarily for building the bar cell, but usage will primarily be the default at the end.

packages/frontend/src/hooks/useColumns.tsx
rephus: done

packages/frontend/src/hooks/useColumns.tsx
rephus: done

rephus:
We'll need to modify the looker2dbt script to support this new config like:

```

chartConfig:
  type: table
  config:
    columns:
      payments_total_revenue:
        displayStyle: bar
```

packages/frontend/src/components/DataViz/utils/columnStats.ts
rephus: Done

packages/frontend/src/components/DataViz/utils/columnStats.ts
```diff
@@ -0,0 +1,44 @@
...
+    // Initialize with first row values
+    columnReferences.forEach((colRef) => {
+        // Initialize stats for all columns with default values
+        stats[colRef] = { min: Infinity, max: -Infinity };
+    });
+
+    // Scan all rows to find min/max
+    rows.forEach((row) => {
+        columnReferences.forEach((colRef) => {
+            const value = Number(row[colRef]);
```
stevelikesmusic, line 31: ❤️

stevelikesmusic (approved):

PR #17356, opened 2025-10-09
fix: error when filtering by metric with fanout

magnew:
Closes: #17014 
Relates to: #16829 

### Description:

There is a SQL when filtering by a metric not included in the query with fanouts protection.

Repro:
- Select a dimension
- Select a metric from a joined table
- Filter by a metric from the joined table, but don't select it
- Run the query
- Error

There are a couple localhost repros that match customer issues:
[Repro 1](http://localhost:3000/projects/3675b69e-8324-4110-bdca-059031aa8da3/tables/fanouts_accounts?create_saved_chart_version=%7B%22tableName%22%3A%22fanouts_accounts%22%2C%22metricQuery%22%3A%7B%22exploreName%22%3A%22fanouts_accounts%22%2C%22dimensions%22%3A%5B%5D%2C%22metrics%22%3A%5B%22fanouts_accounts_inflated_industry_count%22%5D%2C%22filters%22%3A%7B%22metrics%22%3A%7B%22id%22%3A%22b836ed72-c9cc-438b-91c6-0dacb35591d7%22%2C%22and%22%3A%5B%7B%22id%22%3A%22a5e8a611-24b8-4a98-98aa-e9d5e762fb52%22%2C%22target%22%3A%7B%22fieldId%22%3A%22fanouts_deals_inflated_deal_count%22%7D%2C%22operator%22%3A%22notNull%22%2C%22values%22%3A%5B%5D%7D%5D%7D%7D%2C%22sorts%22%3A%5B%5D%2C%22limit%22%3A500%2C%22tableCalculations%22%3A%5B%5D%2C%22additionalMetrics%22%3A%5B%5D%2C%22metricOverrides%22%3A%7B%7D%7D%2C%22tableConfig%22%3A%7B%22columnOrder%22%3A%5B%22fanouts_accounts_inflated_industry_count%22%5D%7D%2C%22chartConfig%22%3A%7B%22type%22%3A%22big_number%22%2C%22config%22%3A%7B%22selectedField%22%3A%22fanouts_accounts_inflated_industry_count%22%2C%22showBigNumberLabel%22%3Atrue%2C%22showComparison%22%3Afalse%2C%22comparisonFormat%22%3A%22raw%22%2C%22flipColors%22%3Afalse%7D%7D%7D)

[...]

PR #17360, opened 2025-10-09
fix: filters can break fanned out charts in dashbaords

magnew:
### Problem

There is a problem where *dashboard filters* on charts generated from fanned out queries can fail because the fanout ctes don't include references to the filter fields. I tried to solve this a couple ways, but it seems like it might come down to solving the cascading fanout problem. Bellow is a claude-explanation after working on this problem a few ways. I think we have a couple options to solve the issue:
1. Do what this PR does and add fanout protection for every table in the join chain in a query (pros and cons below)
2. Try to actually detect tables in the chain the need protection (pros and cons below)
3. Try to selectively disable filters that can't be used on a chart tile
4. Any other ideas? Maybe there is something more elegant. 

**Dashboard filters can reference metrics that aren't selected in the chart.** When these filter-only metrics come from tables involved in fanout joins, the query builder needs to:

  1. Include the metric in the SQL (for the WHERE clause)
  5. Create fanout protection CTEs for the metric's table
  6. Join that CTE into the final query

  **The bug:** `findTablesWithMetricInflation` only detects tables that are directly inflated by being on the "one" side of a one-to-many join. It misses tables that are inflated **transitively** through the join
  chain.

  **Example from the bug:**
  fanouts_accounts (base)
    → fanouts_deals (one-to-many)
      → fanouts_industry_sales_targets (many-to-one with complex condition)

  Dashboard filter: fanouts_deals.inflated_deal_count IS NOT NULL
  Selected metric: fanouts_accounts.inflated_industry_count

  - Chart only selects `fanouts_accounts.inflated_industry_count`
  - Dashboard applies filter on `fanouts_deals.inflated_deal_count`
  - `findTablesWithMetricInflation` marks `fanouts_accounts` as needing protection
  - But it **doesn't** mark `fanouts_deals` as needing protection
  - Result: `fanouts_deals.inflated_deal_count` is placed in `cte_unaffected` instead of its own CTE
  - The filter references a column that doesn't exist → query fails

  ### Why Cascading Inflation Detection Fixes This

  The core issue is that **any table in a join chain with fanout joins has potentially inflated metrics**, even if it's on the "many" side:

  fanouts_accounts (100 rows)
    → fanouts_deals (1:3 relationship, produces 300 rows)

  If you calculate COUNT(deal_id) on those 300 rows, you get 300.
  But if there's ALSO a fanout from accounts → sales_targets,
  those 300 rows might actually be 900 rows after the second fanout.
  Now COUNT(deal_id) = 900, which is wrong.

  When there's ANY fanout join in the query, ALL tables with metrics in that join chain need CTEs to:
  1. Deduplicate their rows using primary keys
  2. Calculate metrics on clean data
  3. Join the deduplicated metrics together at the end

  **The fix:** When `tablesWithMetricInflation` has any tables, add ALL tables that have metrics in the query. This ensures filter-only metrics get their own CTEs and can be properly referenced in WHERE clauses.

  ### The Tradeoff

  **This approach is intentionally conservative:**

  ✅ **Fixes dashboard filters** - Filter-only metrics are properly protected and available in WHERE clauses
  ✅ **Prevents cascading inflation bugs** - All metrics calculated on correct row counts
  ✅ **Simpler to reason about** - Clear rule: "any fanout → protect all metrics"

  ⚠️ **May create unnecessary CTEs** - Tables without actual inflation still get protected
  ⚠️ **Potentially less efficient** - More CTEs means more subqueries to execute

  **Alternative:** Implement graph-based transitive fanout detection to identify exactly which tables are inflated through the join path. More accurate but significantly more complex.

  ### Additional Fixes

  1. **Filter-only metrics in metric CTEs**: Changed line 1202-1203 to use `allMetricsObjects` instead of `metricsObjects` so filter-only metrics are included in the final metrics CTE SELECT
  2. **Empty CTE prevention**: Fixed `hasUnaffectedCte` condition to prevent creating `cte_unaffected` with no columns

  ---

  **Bottom line:** Dashboard filters with metrics require that those metrics have proper fanout protection. The simplest correct solution is to protect all tables when any fanout exists.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
@@ -894,6 +894,22 @@ export class MetricQueryBuilder {
         }, []);
 
         const allMetricsObjects = [...metricsObjects, ...metricsFromFilters];
+
+        // Add all tables with metrics to inflation set if there are any inflated tables
+        // This handles cascading inflation where a table's metrics are inflated by a chain of joins
+        if (tablesWithMetricInflation.size > 0) {
+            allMetricsObjects.forEach((metric) => {
+                if (
+                    joinedTables.has(metric.table) &&
+                    !tablesWithMetricInflation.has(metric.table)
+                ) {
+                    // This table has metrics and is joined, but not marked as inflated
+                    // In a query with ANY fanout joins, all tables with metrics need protection
+                    tablesWithMetricInflation.add(metric.table);
+                }
+            });
+        }
+
         const metricsWithCteReferences: Array<CompiledMetric> = [];
         const referencedMetricObjects = metricsObjects.reduce<CompiledMetric[]>(
             (acc, metricObject) => {
@@ -965,7 +981,7 @@ export class MetricQueryBuilder {
                 tables: [tableName],
             });
             const metricsFromTable = [
-                ...metricsObjects,
+                ...allMetricsObjects,
                 ...referencedMetricObjects,
             ].filter((metric) => metric.table === tableName);
             metricsFromTable.forEach((metric) => {
@@ -985,7 +1001,7 @@ export class MetricQueryBuilder {
         tablesWithMetricInflation.forEach((tableName) => {
             const table = explore.tables[tableName];
             const metricsFromTable = [
-                ...metricsObjects,
+                ...allMetricsObjects,
                 ...referencedMetricObjects,
             ].filter((metric) => metric.table === tableName);
             const metricsInCte: CompiledMetric[] = [];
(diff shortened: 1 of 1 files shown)


magnew:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17360?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#17360** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17360?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17360?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17356** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17356?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

magnew:
Discussed live. This will probably work and probably cause bad perf.

PR #17368, opened 2025-10-10
feat: Display field labels instead of IDs in funnel chart

ongdisheng:
Closes: #10672 

### Description:

This PR fixes the issue where field IDs were displayed instead of field labels in funnel charts by importing the method `getItemLabelWithoutTableName` from common and uses it to retrieve the proper field label when building funnel chart data points. Additionally, a null check was added for `numericFields` to prevent a crash that occurred when switching from other chart types to funnel charts, where `numericFields` could be undefined during the initial render transition.

#### Screenshot
image


ZeRego:
@all-contributors add @ongdisheng for code

PR #17386, opened 2025-10-10
feat: add set fields to model schema

stevelikesmusic:
Closes: CENG-66

### Description:
Adds support for field sets in explores, allowing users to define reusable groups of fields that can be referenced in joins and metrics.

Field sets can be defined in a model's YAML and referenced using the `*` suffix. For example:

```yaml
models:
  - name: payments
    meta:
      sets:
        base_set:
          fields:
            - orders.status
            - orders.shipping_method
```

These sets can then be referenced in joins or `show_underlying_values` with:
```
- base_set*
```

The implementation includes:
- Support for one level of nesting (sets can reference other sets)
- Exclusion syntax with `-field_name`
- Validation of fields
- Comprehensive test coverage

This PR replaces the explicit field list in the payments model with field sets to demonstrate usage.


stevelikesmusic:
* **#17386** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17386?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
Tests are failing :eyes:

packages/common/src/types/explore.ts
```diff
@@ -49,6 +50,8 @@ export type CompiledTable = TableBase & {
     uncompiledSqlWhere?: string;
     parameterReferences?: string[];
     parameters?: LightdashProjectConfig['parameters'];
+    sets?: Record<string, FieldSetDefinition>;
```
rephus, line 53: what's the difference between these 2 ? should we add some comment explaining ?

packages/common/src/compiler/translator.ts
```diff
@@ -339,6 +339,86 @@ function normalizePrimaryKey(
...
+        setDef.fields.forEach((field) => {
+            if (typeof field !== 'string') {
+                throw new ParseError(
+                    `Set "${setName}" in model "${model.name}" contains non-string field: ${field}`,
+                );
+            }
+
+            // Allow field references ending with * (set references)
+            // Allow field references starting with - (exclusions)
+            const cleanField = field.replace(/^-/, '').replace(/\*$/, '');
```
rephus, line 386: if this syntax gets more complex, we might consider implementing a grammar, like we do for dbt filters. But at this point, this is probably easier. 

Should we add some unit tests for this function too ? (not just the one in exploreCompiler)

packages/common/src/compiler/fieldSetExpander.ts
```diff
@@ -0,0 +1,202 @@
+import { type FieldSetDefinition } from '../types/dbt';
+import { CompileError } from '../types/errors';
+import { type Table } from '../types/explore';
+
+export const FIELD_SET_LIMITS = {
+    MAX_SET_SIZE: 100,
```
rephus: why do we need limits ?

examples/full-jaffle-shop-demo/dbt/models/payments.yml
```diff
@@ -37,15 +47,16 @@ models:
...
-                  - customers.date_of_most_recent_created_customer
-                  # Dimensions
-                  - amount
-                  - orders.status
-                  - orders.customer_id
-                  - orders.order_date
-                  - orders.order_id
-                  - orders.order_source
-                  - orders.shipping_method
+                  - test_set*
```
rephus, line 52: should this be base_set ?

examples/full-jaffle-shop-demo/dbt/models/payments.yml
```diff
@@ -5,6 +5,16 @@ models:
     config:
       tags: ["ai"]
       meta:
+        sets: 
+          base_set:
+            fields:
```
rephus, line 10: why do we need fields object ? and not directly a list of strings ?

examples/full-jaffle-shop-demo/dbt/models/payments.yml
rephus: actually, I think this is great for testting the feature

packages/common/src/compiler/translator.ts
stevelikesmusic: That's a good idea for some tests. I hope that the syntax stays like this—just two operators.

packages/common/src/compiler/fieldSetExpander.ts
stevelikesmusic: Ah this is a leftover from an earlier iteration that could have had circular references. I can remove the limits. Before, we recursively went through all sets.

examples/full-jaffle-shop-demo/dbt/models/payments.yml
stevelikesmusic: I thought giving `fields` a namespace, we leave room for any API development. If we pass a list of strings directly to `<set_name>`, any change to the `set` API would be a breaking change. 

Our `join` API goes even further where we have to define `join` within the `joins` namespace:

```yaml
meta:
  joins:
    - join: foo
      sql_on: ...
```

I don't think we need to get that specific for `sets`. But declaring `fields` keeps us from locking the API.

packages/common/src/types/explore.ts
stevelikesmusic: 🔥

packages/common/src/compiler/translator.ts
stevelikesmusic: Also, I had considered doing a check with more definite methods like `field.startsWith('-')` and `field.endsWith('*')`. But in this case, the regex seemed more concise/straightforward. What do you think?

packages/e2e/cypress/e2e/app/dashboard.cy.ts
```diff
@@ -89,7 +89,7 @@ describe('Dashboard', () => {
         cy.url().should('not.include', '?');
     });
 
-    it('views underlying data with dimensions and metrics', () => {
+    it.only('views underlying data with dimensions and metrics', () => {
```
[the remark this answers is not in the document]
rephus: good point, we should remove .only

rephus (approved):
remove the .only and then it is good to go !

stevelikesmusic:
Ugh. I need to add an eslint rule for me 😭

packages/common/src/compiler/translator.ts
```diff
@@ -339,6 +339,111 @@ function normalizePrimaryKey(
...
+                    if (hasNestedSetReferences) {
+                        throw new ParseError(
+                            `Set "${setName}" in model "${model.name}" references set "${referencedSetName}", which itself contains set references. Only one level of set nesting is allowed.`,
+                        );
+                    }
+                }
+            }
+        });
+    });
+}
```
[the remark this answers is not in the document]
stevelikesmusic: I think this would be pre-mature optimization. These are all sync processes with reasonable limits from how the SQL queries work and perform.

stevelikesmusic:
### Merge activity

* **Oct 15, 5:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17386).
* **Oct 15, 5:40 PM UTC**: @stevelikesmusic merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17386).

PR #17398, opened 2025-10-13
fix: preview from cli can copy organization warehouse credentials

rephus:
Closes: https://linear.app/lightdash/issue/CENG-63/apply-this-to-previews-on-ui-and-cli-copy-the-org-uuid-credentials-on

Screenshot from 2025-10-13 09-33-17

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/dbt_project.yml
+++ b/examples/full-jaffle-shop-demo/dbt/dbt_project.yml
@@ -12,7 +12,7 @@ seed-paths:
   - data
 macro-paths:
   - macros
-target-path: "{{ env_var('DBT_ENV__TARGET_FOLDER', 'target') }}"
+target-path: target
 clean-targets:
   - target
   - dbt_modules
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1121,6 +1121,18 @@ export class ProjectService extends BaseService {
         await this.validateProjectCreationPermissions(user, data);
 
         const newProjectData = data;
+
+        // If type preview and has upstream project, we first link the preview to the same organization warehouse credentials (if exists)
+        if (
+            newProjectData.type === ProjectType.PREVIEW &&
+            newProjectData.upstreamProjectUuid
+        ) {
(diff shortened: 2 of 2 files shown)


rephus:
* **#17398** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17398?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17398?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/full-jaffle-shop-demo/dbt/dbt_project.yml
```diff
@@ -12,7 +12,7 @@ seed-paths:
   - data
 macro-paths:
   - macros
-target-path: "{{ env_var('DBT_ENV__TARGET_FOLDER', 'target') }}"
+target-path: target
```
stevelikesmusic, line 15: It strikes again...

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1121,6 +1121,18 @@ export class ProjectService extends BaseService {
         await this.validateProjectCreationPermissions(user, data);
 
         const newProjectData = data;
+
+        // If type preview and has upstream project, we first link the preview to the same organization warehouse credentials (if exists)
+        if (
```
stevelikesmusic, line 1126: Would/could we have an org that has set org credentials, but hasn't created a project yet? Are there any other scenarios we would want a compile to use the org credentials?

rephus: > could we have an org that has set org credentials, but hasn't created a project yet

Actually no, since in order to access settings, you first needt o have a project. 


This makes me think, we might want to improve the `onboarding` flow, and let them create org credentials even before the first project is created. But that might require a bit of design or extra spec, so leaving it for now until someone asks for it.

But this would not matter for this PR, since we only copy the org credentials from the upstream project (if exists) . 

Did I answer your question ?

stevelikesmusic: Yes thanks Javi! And the upstream project is just any other project in the org?

Issue #17372, opened 2025-10-10
Chart scheduled delivery results are NOT pivoted


almeidabbm:
### Description

Related PR: https://github.com/lightdash/lightdash/pull/17351

PR #17425, opened 2025-10-13
fix: pivot results on scheduled deliveries

almeidabbm:
Closes: #17372

### Description:
Add `featureFlagService` to scheduler tasks to enable pivot results in scheduled deliveries. This allows the scheduler to check if the `UseSqlPivotResults` feature flag is enabled for a user when generating scheduled chart results.

The change passes the pivot flag to query execution, ensuring that scheduled deliveries respect the same pivot settings as interactive queries.

**This PR also:**
- Fixes scheduled deliveries form that was always generating csv/xsls with formatted values
- Deduplicates logic between CSV and XSLS so that they both use the same underlying functions

**After this PR:**
- Formatting options should work regardless of feature flag value
- `USE_SQL_PIVOT_RESULTS: true`
  - If chart has a group by then the resulting CSV/XSLS should be pivoted
  - If table viz is pivoted, the resulting csv/xsls should be pivoted
  - When nothing is pivoted or grouped, the csv/xsls should be simple
- `USE_SQL_PIVOT_RESULTS: false`
  - If table is pivoted, scheduled delivery csv/xsls should be pivoted
  - Cartesian charts NEVER return pivoted csv/xsls

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -119,6 +119,7 @@ import {
 } from '../services/CsvService/CsvService';
 import { DashboardService } from '../services/DashboardService/DashboardService';
 import { ExcelService } from '../services/ExcelService/ExcelService';
+import type { FeatureFlagService } from '../services/FeatureFlag/FeatureFlagService';
 import { getDashboardParametersValuesMap } from '../services/ProjectService/parameters';
 import { ProjectService } from '../services/ProjectService/ProjectService';
 import { RenameService } from '../services/RenameService/RenameService';
@@ -152,6 +153,7 @@ export type SchedulerTaskArguments = {
     msTeamsClient: MicrosoftTeamsClient;
     renameService: RenameService;
     asyncQueryService: AsyncQueryService;
+    featureFlagService: FeatureFlagService;
 };
 
 export default class SchedulerTask {
@@ -193,6 +195,8 @@ export default class SchedulerTask {
 
     protected readonly asyncQueryService: AsyncQueryService;
 
+    private readonly featureFlagService: FeatureFlagService;
+
     constructor(args: SchedulerTaskArguments) {
         this.lightdashConfig = args.lightdashConfig;
         this.analytics = args.analytics;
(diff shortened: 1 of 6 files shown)


almeidabbm:
* **#17425** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17425?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17425?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -400,26 +405,38 @@ export default class SchedulerTask {
                 // We don't generate CSV files for Google sheets on handleNotification task,
                 // instead we directly upload the data from the row results in the uploadGsheets task
                 throw new Error("Don't fetch csv for gsheets");
-            case SchedulerFormat.XLSX: {
+            case SchedulerFormat.CSV:
```
almeidabbm, line 408: reuse the same switch case between CSV and XSLS, this makes sure that we're using the same queries and functions to generate both files

packages/backend/src/services/PivotTableService/PivotTableService.ts
```diff
@@ -200,10 +197,18 @@ export class PivotTableService extends BaseService {
...
+                column.pivotColumnName,
+                column,
+            ]) ?? [],
+        );
+
         // PivotQueryResults expects a formatted ResultRow[] type, so we need to convert it first
         // TODO: refactor pivotQueryResults to accept a Record<string, any>[] simple row type for performance
-        const formattedRows = formatRows(rows, itemMap);
-
+        const formattedRows = formatRows(rows, itemMap, pivotValuesColumnsMap);
```
almeidabbm, line 211: This allows the pivoted csv's to have the formatted results

packages/frontend/src/features/scheduler/components/SchedulerForm.tsx
```diff
@@ -424,7 +426,7 @@ const SchedulerForm: FC<Props> = ({
                 )
             ) {
                 options = {
-                    formatted: values.options.formatted,
+                    formatted: values.options.formatted === Values.FORMATTED,
```
almeidabbm, line 429: We were passing the string `'formatted'` or `'raw'` to the backend rather than the boolean options needs. All scheduled deliveries were formatted regardless of the option

magnew: Verified this works now -- you can send raw

PR #17436, opened 2025-10-14
fix: useExporerQuery not running pivot query when USE_SQL_PIVOT_RESULTS is enabled

almeidabbm:
## Summary

  Refactored `useExplorerQuery` to fix pivot queries not running when `USE_SQL_PIVOT_RESULTS` feature flag is enabled.

  ## Problem

  The hook was using `unsavedChartVersion` from Redux, which only contains Redux-managed fields. This meant `pivotConfig` (stored in Explorer Context) was missing when building query arguments, preventing pivot queries from executing.

  ## Solution

  **1. Use merged chart version**
  - Changed to use `mergedUnsavedChartVersion` from Explorer Context
  - Ensures both Redux fields and context-only fields (`chartConfig`, `pivotConfig`) are included in query arguments

  **2. Extracted query management into three focused hooks**
  - `useExplorerQueryManager` - Query state and `runQuery` logic (single source of truth)
  - `useExplorerQueryEffects` - Auto-fetch and orchestration effects (call once at root)
  - `useExplorerQuery` - Public API for components (action functions + state)

  **3. Cleaned up auto-fetch logic**
  - Removed manual auto-fetch `useEffect` from `Explorer/index.tsx`
  - Now handled automatically by `useExplorerQueryEffects`

  ## Testing

  - Verify pivot queries execute when `USE_SQL_PIVOT_RESULTS` is enabled
  - Confirm auto-fetch behavior works for saved charts, dashboards, and pivot configs
  - Check that unpivoted data displays correctly in results table

Note: check that https://github.com/lightdash/lightdash/pull/17370 and https://github.com/lightdash/lightdash/pull/17322 still work + query autofetch logic + make sure that minimal pages and embed always invalidate cache

PR #17437, opened 2025-10-14
feat: add post-calculation metrics for running totals and percentages

ZeRego:
Relates to: https://linear.app/lightdash/issue/CENG-77

### Description:
This PR adds support for post-calculation metrics, which are metrics that perform calculations on other metrics after aggregation. Three new metric types are introduced:

1. `percent_of_previous` - Calculates the percentage change from the previous value
2. `percent_of_total` - Calculates the percentage of the total
3. `running_total` - Calculates a running total of values

The implementation includes:
- New metric types in the common package
- Validation to ensure post-calculation metrics only reference numeric metrics
- SQL compilation logic for each post-calculation type
- Support for these metrics in filters, sorting, and other operations
- Integration with pivot tables

The PR also includes an example in the jaffle shop demo with three new metrics that demonstrate these capabilities.

**Note:** for pivot charts, these metrics are only calculated correctly if we have env var `USE_SQL_PIVOT_RESULTS: true`. This is why these metrics are marked as Experimental.
**Note2:** There might some weird behaviour in the FE until https://github.com/lightdash/lightdash/pull/17436 is merged

### Demo
Screenshot 2025-10-14 at 11 17 37

### Docs PR - https://github.com/lightdash/mintlify-docs/pull/183
Screenshot 2025-10-14 at 14 38 44
Screenshot 2025-10-14 at 14 38 53
Screenshot 2025-10-14 at 14 39 09

### DIFF
--- a/packages/common/src/dbt/schemas/lightdashMetadata.json
+++ b/packages/common/src/dbt/schemas/lightdashMetadata.json
@@ -224,7 +224,10 @@
                         "number",
                         "string",
                         "date",
-                        "boolean"
+                        "boolean",
+                        "percent_of_previous",
+                        "percent_of_total",
+                        "running_total"
                     ]
                 },
                 "description": {
--- a/packages/common/src/schemas/json/lightdash-dbt-2.0.json
+++ b/packages/common/src/schemas/json/lightdash-dbt-2.0.json
@@ -148,7 +148,10 @@
                                                     "min",
                                                     "number",
                                                     "string",
-                                                    "sum"
+                                                    "sum",
+                                                    "percent_of_previous",
+                                                    "percent_of_total",
+                                                    "running_total"
                                                 ]
                                             },
                                             "label": {
@@ -630,7 +633,10 @@
                                                                 "min",
                                                                 "number",
                                                                 "string",
-                                                                "sum"
+                                                                "sum",
+                                                                "percent_of_previous",
+                                                                "percent_of_total",
+                                                                "running_total"
                                                             ]
                                                         },
                                                         "label": {


ZeRego:
* **#17437** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17437?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1469,6 +1471,44 @@ export class AsyncQueryService extends ProjectService {
         }
     }
 
+    private async getMetricQueryFields({
```
ZeRego, line 1474: Compiling the metric query now needs the pivot configuration, causing a catch-22* scenario.

> catch-22 is a paradoxical situation from which an individual cannot escape because of contradictory rules or limitations.


I need the pivot configuration to compile the metric query.
I need the compiled metric query to derive the pivot configuration from saved chart.


I had to copy this logic from the ProjectService._compileMetric service, so we can call `derivePivotConfigurationFromChart` before compiling the metric query.

packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
```diff
@@ -322,51 +421,20 @@ export class MetricQueryBuilder {
             });
         }
 
-        // Find metrics from metric query
```
ZeRego, line 325: Extracted most of the logic to `getSelectedAndReferencedMetricIds`. This made it clear we had duplicate loops and logic. Refactor to first understand all the metrics that need to be queried and then loop over them to generate the SQL and collect the tables.

examples/full-jaffle-shop-demo/dbt/models/orders.yml
```diff
@@ -146,6 +146,19 @@ models:
                 default_time_dimension:
                   field: order_date
                   interval: DAY
+              amount_percent_of_previous:
+                type: percent_of_previous
+                sql: ${total_order_amount}
+                format: "#,##0.00%"
```
rephus, line 152: should we apply this format by default ? 

Image

rephus:
NaN ? 
image

I guess it is trying to convert ∅ into a number

rephus:
:kissing: :ok_hand: 

image

rephus:
This filter doesn't make any sense, but it works :+1: 

image

packages/backend/src/queryCompiler.ts
```diff
@@ -250,6 +253,66 @@ const compileAdditionalMetric = ({
...
+        groupByColumns.length > 0
+            ? `PARTITION BY ${groupByColumns
+                  .map((col) => `${q}${col.reference}${q}`)
+                  .join(', ')}`
+            : undefined;
+
+    const finalOrderByClause = orderByClause ?? `ORDER BY (SELECT NULL)`;
+
+    if (type === MetricType.RUNNING_TOTAL) {
+        return `SUM(${sql}) OVER (${
```
rephus, line 288: I tested these in postgres, snowflake and bigquery. It works on all 3, syntax seems pretty generic.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: Why can't we make `_compileMetric` a public method? Doesn't AsyncQueryService extend ProjectService?

examples/full-jaffle-shop-demo/dbt/models/orders.yml
ZeRego: New ticket: https://github.com/lightdash/lightdash/issues/17488

Issue #17433, opened 2025-10-14
[CLI] Authenticate to Snowflake with "externalbrowser"


ZeRego:
### Description

dbt profile docs: https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#sso-authentication

```
my-snowflake-db:
  target: dev
  outputs:
    dev:
      type: snowflake
      account: [account id] # Snowflake <account_name>
      user: [username] # Snowflake username
      role: [user role] # Snowflake user role

      # SSO config
      authenticator: externalbrowser

      database: [database name] # Snowflake database name
      warehouse: [warehouse name] # Snowflake warehouse name
      schema: [dbt schema]

```

Note that this authentication type might have some limitations. Ex: the user might need to authenticate multiple times (1 for dbt and 1 for Lightdash)

### What problem does this solve?

Even with limitations, it unblocks orgs that only support this authentication option. 
At the moment, they can only use the Lightdash CLI with the option "--no-warehouse-credentials"

PR #17455, opened 2025-10-15
feat: add Snowflake external browser authentication support

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/17433

This is an MVP for testing externalbrowser authentication ont he CLI.  We should ship this, and let the customers test it before working on the following bugs/fixes

Authenticating on the CLI 

[Screencast from 2025-10-15 12-13-28.webm](https://github.com/user-attachments/assets/72783adb-3d31-4e34-81b9-fa82838438a0)

Running queries on the project 

[Screencast from 2025-10-15 12-14-29.webm](https://github.com/user-attachments/assets/e35f4b9e-4984-42db-a8c4-e15df8b66315)

I had to install some pythong dependencies, and add create some system folders before authentication was working even on dbt. Let me know if you have any issues testing this. 

### Next steps  (create new issues)

- [ ] Add externalbrowser authentication enum to the snowflake form 
Screenshot from 2025-10-15 12-15-25

- [ ] Using externalbrowser authentcation opens 2 tabs (I guess we initialize snowflake client twice)

### Description:
Add support for Snowflake external browser authentication in the CLI and warehouse client. This allows users to authenticate to Snowflake using their browser, which is useful for environments that require SSO or MFA.

Key changes:
1. Added `external_browser` as a new authentication type in the Snowflake enum
2. Updated the Snowflake warehouse client to support the new authentication method
3. Added caching for Snowflake connections with external browser auth to avoid opening multiple browser tabs
4. Updated the Snowflake SDK dependency from ~2.1.3 to ~2.3.1 to support the new authentication method
5. Fixed the order of success/error enums in the generated routes


rephus:
* **#17455** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17455?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/handlers/dbt/getWarehouseClient.ts
```diff
@@ -21,6 +22,12 @@ import {
 import GlobalState from '../../globalState';
 import { lightdashApi } from './apiClient';
 
+// Cache Snowflake client with external browser auth to avoid opening multiple browser tabs
```
rephus: without this cachedclient, we'll be creating a lot of warehouse clients, and trigger authentication multiple times, opening hundres of browsr tabs, and causing 429 error: (too many requests)

packages/cli/src/handlers/dbt/getWarehouseClient.ts
```diff
@@ -264,12 +272,36 @@ export default async function getWarehouseClient(
...
         GlobalState.debug(`> Using target ${target.type}`);
         credentials = await warehouseCredentialsFromDbtTarget(target);
-        warehouseClient = warehouseClientFromCredentials({
-            ...credentials,
-            startOfWeek: isWeekDay(options.startOfWeek)
-                ? options.startOfWeek
-                : undefined,
-        });
+
+        // Check if we can reuse cached Snowflake client with external browser auth
```
rephus: we could cache the client for all warehouses, not just snowflake. That would simplify the code

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -271,13 +279,36 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
...
             connection = createConnection({
                 ...this.connectionOptions,
                 ...connectionOptionsOverrides,
             });
-            await Util.promisify(connection.connect.bind(connection))();
+
+            // External browser authentication requires connectAsync()
+            if (this.connectionOptions.authenticator === 'EXTERNALBROWSER') {
+                await Util.promisify(
+                    connection.connectAsync.bind(connection),
```
rephus: this is where the authentication happens, it is all handled by the library, also on Lightdash UI

packages/warehouses/package.json
```diff
@@ -15,7 +15,7 @@
         "lodash": "^4.17.21",
         "pg": "^8.13.1",
         "pg-cursor": "^2.10.0",
-        "snowflake-sdk": "~2.1.3",
+        "snowflake-sdk": "~2.3.1",
```
rephus, line 18: there was a bug about `connectAsync` on 2.1.3 (version from 2020)  

>  When we tried to use external browser authentication with
  connectAsync(), we got the error: "External browser and Okta are not compatible with connection
  process"

I don't see any breaking changes https://docs.snowflake.com/en/release-notes/clients-drivers/nodejs-2025

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -710,14 +750,17 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
             // Ignore error and let UI show invalid table
             return undefined;
         } finally {
-            await new Promise((resolve, reject) => {
-                connection.destroy((err, conn) => {
-                    if (err) {
-                        reject(new WarehouseConnectionError(err.message));
-                    }
-                    resolve(conn);
+            // Only destroy connection if it's not cached
```
ZeRego: third time we duplicate this code. We can move it to a method

ZeRego (approved):

ZeRego:
> Add externalbrowser authentication enum to the snowflake form

For now, I would only display that option if the value is external_browser. Not sure if we should support this for new projects in the UI if they need to authenticate on each query. 🤔

packages/cli/src/handlers/dbt/getWarehouseClient.ts
ZeRego: makes sense to cache all 👍 this will be cached once per Lightdash command so there should be no side effects when switching accounts or projects.

rephus: done . I confirmed the cache is recreated on each command. Still, I added a `cache key` based on the credentials per config (is this too much?) 

Screenshot from 2025-10-20 09-27-06

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
rephus: updated

rephus:
I made some improvements, cleanup code. I remove dthe cache from the warehouse client, it was redudant considering we do it on the CLI now

packages/cli/src/handlers/dbt/getWarehouseClient.ts
```diff
@@ -264,12 +292,32 @@ export default async function getWarehouseClient(
...
+
+            warehouseClient = warehouseClientFromCredentials({
+                ...credentials,
+                startOfWeek: isWeekDay(options.startOfWeek)
+                    ? options.startOfWeek
+                    : undefined,
+            });
+
+            // Cache client if appropriate
+            if (cacheKey) {
```
stevelikesmusic: Is there a problem if we don't have a cacheKey?

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -210,6 +214,10 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
                 token: credentials.token,
                 authenticator: 'OAUTH',
             };
+        } else if (credentials.authenticationType === 'external_browser') {
```
stevelikesmusic: Why not use the constant for comparison?

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
```diff
@@ -278,22 +286,85 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
...
+    private async createExternalBrowserConnection(
+        connectionOptionsOverrides?: Partial<ConnectionOptions>,
+    ): Promise<Connection> {
+        // Return cached promise if one exists (handles both in-flight and completed connections)
+        if (this.externalBrowserConnectionPromise) {
+            return this.externalBrowserConnectionPromise;
+        }
+
+        // Create and cache the connection promise
+        this.externalBrowserConnectionPromise = (async () => {
```
stevelikesmusic, line 339: Why use the IIFE here but a Promise constructor for destroy—just duplicating the existing pattern?

stevelikesmusic (commented):
Very cool!

packages/cli/src/handlers/dbt/getWarehouseClient.ts
rephus: ah, no, before cachekey was optional, now it is not

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
rephus: yes, we want to cache the promise, to avoid opening multiple tabs for externalbrowser authentication, this is similar ot the `createConnection` but without the cache.

packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
rephus: yeah, makes sense. I'll update this and all other authenticationtypes

PR #17477, opened 2025-10-15
fix: improve explore table and modal render perf

magnew:
### Description:

Small improvements to render perf on explores page
- Don't call getBoundingClientRect 2x for every cell in the table (actually seems to make a difference)
- Memoize modals


packages/frontend/src/components/common/Table/ScrollableTable/BodyCell.tsx
```diff
@@ -73,6 +74,16 @@ const BodyCell: FC<React.PropsWithChildren<CommonBodyCellProps>> = ({
...
 
+    // Calculate bounds when menu/tooltip opens, not during every render
+    useEffect(() => {
+        if ((shouldRenderMenu || shouldRenderTooltip) && elementRef.current) {
+            setElementBounds(elementRef.current.getBoundingClientRect());
+        } else if (!isMenuOpen && !isTooltipOpen) {
+            // Clear bounds when closed to free memory
+            setElementBounds(null);
+        }
+    }, [shouldRenderMenu, shouldRenderTooltip, isMenuOpen, isTooltipOpen]);
```
almeidabbm, line 85: 🧑‍🍳 🤌

PR #17482, opened 2025-10-15
chore: improve tsconfig for incremental builds

stevelikesmusic:
### Description:

I'd been hitting some gnarly tsc waits in Webstorm and sometimes in Cursor. tsserver would run and run compiling all files. This attempts to configure TS to build more efficiently and not constantly rebuild the entire project. 

Enables TypeScript project references across packages to improve build performance and type checking. This PR:

- Enables declaration file generation with source maps for backend, cli, and warehouses packages
- Sets up project references between packages to establish proper dependency relationships
- Configures composite builds for better incremental compilation
- Standardizes tsBuildInfoFile locations to dist/.tsbuildinfo across packages
- Enables composite builds in common package's CJS and ESM configurations

These changes will improve the development experience by allowing TypeScript to better understand dependencies between packages and enable more efficient incremental builds.


stevelikesmusic:
* **#17482** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17482?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/clients/github/Github.ts
```diff
@@ -49,7 +51,7 @@ export const getOctokitRestForUser = (authToken: string) => {
     };
 };
 
-export const getOctokitRestForApp = (installationId: string) => {
+export const getOctokitRestForApp = (installationId: string): OctokitRest => {
```
rephus, line 54: this changed because we need to add now return types everywhere ? is this mandatory ?

packages/backend/tsconfig.json
```diff
@@ -67,5 +67,13 @@
...
     },
     "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.json"],
-    "exclude": ["dist", "node_modules"]
+    "exclude": ["dist", "node_modules"],
+    "references": [
+        {
+            "path": "../common"
+        },
+        {
+            "path": "../warehouses"
```
rephus, line 76: warehouses doesn't change that often, if this has some negative impact, we can consider to keep this out.

packages/common/src/index.ts
```diff
@@ -958,32 +948,14 @@ type ApiResults =
...
-    | ApiAiAgentEvaluationSummaryListResponse['results']
-    | ApiAiAgentEvaluationResponse['results']
-    | ApiAiAgentEvaluationRunResponse['results']
-    | ApiAiAgentEvaluationRunSummaryListResponse['results']
-    | ApiAiAgentEvaluationRunResultsResponse['results']
-    | ApiCreateEvaluationResponse['results']
-    | ApiAppendInstructionResponse['results']
-    | ApiGetChangeResponse['results']
-    | ApiAiOrganizationSettingsResponse['results']
-    | ApiUpdateAiOrganizationSettingsResponse['results'];
```
rephus, line 984: dont' we need these api types ?

rephus:
I think these tests are unrelated 
image
, but still, it might be worth checking manually

stevelikesmusic:
Fair point. Yes there were multiple files changed for imports—the reason being how tsc compiles. It can't navigate layers of barrel exports files when building declaration files, which we need for `composite: true`.

packages/backend/src/clients/github/Github.ts
stevelikesmusic: This is because of how TypeScript infers return types. If we use function declarations (ie. `export function getOctokitRestForApp`), then tsc has a better idea of types. With function expressions (ie. `export const getOctokitRestForApp`), tsc needs more help in getting complex or nested types. 

This is only a concern with `composite: true` because tsc has stricter requirements for writing declaration files. When we have `composite: false`, tsc keeps everything in memory, so it has more context of complex types. This is a tradeoff of having efficient tsc compilation in a monorepo.

packages/backend/tsconfig.json
stevelikesmusic: I don't think this is a negative impact. It just indicates that backend should look for a declaration file in `warehouses` rather than doing everything in memory when a file in `warehouses` does change.

packages/common/src/index.ts
stevelikesmusic: I'm not so sure. These types are only used for `ApiResponse` which is only used as the return type for our two API clients ([frontend](https://github.com/lightdash/lightdash/blob/61e31d511d5906ce0109f630d06b5833f0da82a8/packages/frontend/src/api.ts#L116) and [cli](https://github.com/lightdash/lightdash/blob/61e31d511d5906ce0109f630d06b5833f0da82a8/packages/cli/src/handlers/dbt/apiClient.ts#L26-L30)). In both cases, we end up casting types anyway with the generic, so I don't believe these types are event used. 

If we do want to keep these, I think we'll need to pull API types into a separate package so we don't have circular dependencies between `common/index` and `common/ee` which is the reason for this change.

stevelikesmusic:
It looks like these tests are failing on `main` as well. Same suites but different tests. The failures don't appear to be related.

packages/common/src/index.ts
stevelikesmusic: I ended up moving these to avoid circular dependencies between ee and this index. All API types are now in common/types/api.ts

stevelikesmusic:
### Merge activity

* **Oct 23, 8:56 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17482).
* **Oct 23, 8:56 AM UTC**: @stevelikesmusic merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17482).

PR #17494, opened 2025-10-16
feat: remove project access and group memberships when deactivating SCIM users

ZeRego:
Closes: https://linear.app/lightdash/issue/CENG-62/disabled-user-should-drop-their-role-to-member

### Description:
Enhances SCIM user deactivation to properly clean up user permissions when a user is deactivated through SCIM. When a user is set to inactive:

1. Their organization role is downgraded to MEMBER
2. They are removed from all project roles
3. They are removed from all groups in the organization

This ensures that deactivated users lose all access privileges while maintaining their account for potential future reactivation.

Also fixes the `removeUserProjectAccess` method to properly filter by project UUID, and adds a new `removeUserAccessFromAllProjects` method to remove a user from all projects at once.

### Demo

```
# Deactivate jose@lightdash.com via SCIM
EMAIL="..."
BASE_URL="http://localhost:3000/"
TOKEN="..."
USER_ID="..."

# PATCH the user to set active=false
curl -X PATCH \
  "$BASE_URL/api/v1/scim/v2/Users/$USER_ID" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/scim+json" \
  -d '{
    "schemas": ["urn:ietf:params:scim:api:messages:2.0:PatchOp"],
    "Operations": [
      { "op": "Replace", "path": "active", "value": false }
    ]
  }'
```

https://github.com/user-attachments/assets/371b046e-37de-4c4a-aa0f-f56fea1596ba

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.mock.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.mock.ts
@@ -75,10 +75,13 @@ export const ScimServiceArgumentsMock: ConstructorParameters<
     userModel: userModelMock,
     emailModel: emailModelMock,
     analytics: analyticsMock,
-    groupsModel: {} as GroupsModel,
+    groupsModel: {
+        removeUserFromAllGroups: jest.fn().mockResolvedValue(2),
+    } as unknown as GroupsModel,
     serviceAccountModel: {} as ServiceAccountModel,
     commercialFeatureFlagModel: {} as CommercialFeatureFlagModel,
     rolesModel: {
         removeUserProjectAccess: jest.fn().mockResolvedValue(undefined),
+        removeUserAccessFromAllProjects: jest.fn().mockResolvedValue(3),
     } as unknown as RolesModel,
 };
--- a/packages/backend/src/ee/services/ScimService/ScimService.test.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.test.ts
@@ -272,6 +272,63 @@ describe('ScimService', () => {
     });
 
     describe('updateUser', () => {
+        test('should downgrade role to MEMBER and remove project and group memberships when deactivating user', async () => {
+            const { organizationMemberProfileModel, rolesModel, groupsModel } =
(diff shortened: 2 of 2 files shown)


ZeRego:
* **#17494** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17494?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/RolesService/RolesService.ts
```diff
@@ -806,7 +806,7 @@ export class RolesService extends BaseService {
         );
         await this.validateProjectAccess(account, projectUuid);
 
-        await this.rolesModel.removeUserProjectAccess(userUuid);
+        await this.rolesModel.removeUserProjectAccess(userUuid, projectUuid);
```
ZeRego, line 809: Fix issue where it was remove access from all projects.

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -525,6 +531,75 @@ export class ScimService extends BaseService {
                 );
             }
 
+            // If setting user to inactive, drop org role to MEMBER and remove project roles
+            if (user.active === false) {
```
stevelikesmusic, line 535: Should these DB operations be in a transaction? What would the effect be if some operations failed? There doesn't seem to be any indication to admins that trigger this function that any part has failed.  

Also, would it be worth pulling this logic into its own function?

ZeRego: Not worth imo. Most operations in the service layer are executed in separate transactions, and errors are handled case by case.
These are non-blocking operations and very unlikely to throw an error. It logs errors on failure in case there needs to be an investigation.

PR #17507, opened 2025-10-16
fix: improve fitler card init

magnew:
### Description:

WIP Filter card loading is very slow. This is a POC to move the processing to a worker. It seems to help a lot, but this approach has some problems. I'm working on:

- seeing if I can clean up the implementation (theres a lot to support the worker)
- Or just optimizing the processing to begin with -- this might be the right answer.


magnew:
Closing in favor of a better PR

PR #17524, opened 2025-10-17
fix: add role to snowflake sso authentication

rephus:
Thread: https://lightdash.slack.com/archives/C07KMC3A0BZ/p1760701246803189?thread_ts=1759417019.733669&cid=C07KMC3A0BZ

This fixes roles when default_role for the user doesn't match the role on the project , causing access issues

I also added some non sensitive logging to make debugging a lot easier. 
image

*NOTE* : This might break some existing snowflake sso authenticated users if they don't have the right role (I guess this should be the intended behaviour) .YOu will see something like this : 

image

### Description:
Fixed Snowflake authentication by correctly setting the username parameter for different authentication types. Previously, username was always included in the connection options, which could cause conflicts when using SSO authentication. Now, username is only included when appropriate for each authentication method (JWT, password) and excluded for SSO. 

Added detailed connection logging that safely excludes sensitive information, making it easier to debug connection issues while maintaining security.

### DIFF
--- a/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/SnowflakeWarehouseClient.ts
@@ -206,6 +206,7 @@ export class SnowflakeWarehouseClient extends WarehouseBaseClient<CreateSnowflak
                 );
             }
             authenticationOptions = {
+                // Do not include username when doing SSO authentication
                 token: credentials.token,
                 authenticator: 'OAUTH',
             };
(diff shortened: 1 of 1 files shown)


rephus:
* **#17524** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17524?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17524?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #14576, opened 2025-04-28
Error: Maximum update depth exceeded.


bramrodenburg:
### Description

When applying a filter to a dashboard, sometimes (randomly) the following error shows up:

<img width="1497" alt="Image" src="https://github.com/user-attachments/assets/f0845207-6288-4fd6-8f05-3a0b1a8f6725" />

Minified React [error #185](https://react.dev/errors/185) corresponds to:
```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

### Steps to Reproduce the Bug or Issue

1. Go to a dashboard
2. Apply some filters
3. Check if you get the error. If not, go to step 2.

### version

v0.1601.

### Cloud or self-hosting

self-hosting

bramrodenburg:
This (closed) issue seems to be related: https://github.com/lightdash/lightdash/issues/14032

Issue #185, opened 2021-07-06
Boolean series becomes true/false in chart (but it's yes/no in table)


TuringLovesDeathMetal:
![image](https://user-images.githubusercontent.com/31848148/124600085-a74ffe00-de66-11eb-94bd-46313cb5b784.png)

When you add a yes/no dimension and it becomes a series, it gets labelled as `true`, (but in the table it's yes/no) 

The series label should match the values in the table (i.e. should read `yes`/ `no`, no `true` / `false`)

PR #17582, opened 2025-10-22
fix: [LD-14576] - Prevent infinite re-render loop in <TrackPage> component

afranzi:
Closes: [Error: Maximum update depth exceeded. #14576
](https://github.com/lightdash/lightdash/issues/14576)

### 🐛 Bug Fix: React Infinite Re-render Loop (Error #185)

This PR resolves an issue where the `<TrackPage>` component could trigger an infinite re-render loop, leading to the "Maximum update depth exceeded" error (React Minified Error #185) on certain pages, particularly dashboards.

The issue occurred because the `useEffect` hook inside `<TrackPage>` was using the object destructured from props (`...rest`) as a dependency:

```typescript
useEffect(() => {
    page(rest);
}, [page, rest]); // 👈 'rest' is a new object on every re-render
```

Screenshot 2025-10-22 at 11 56 00
Screenshot 2025-10-22 at 11 55 54

### DIFF
--- a/packages/frontend/src/providers/Tracking/TrackingProvider.tsx
+++ b/packages/frontend/src/providers/Tracking/TrackingProvider.tsx
@@ -178,11 +178,14 @@ export const TrackPage: FC<React.PropsWithChildren<PageData>> = ({
 }) => {
     const { page } = useTracking();
 
-    const pageData = useMemo(() => rest, [
-        rest.name,
-        rest.category,
-        rest.type,
-    ]);
+    const pageData = useMemo(
+        () => ({
+            name: rest.name,
+            category: rest.category,
+            type: rest.type,
+        }),
+        [rest.name, rest.category, rest.type],
+    );
 
     useEffect(() => {
         page(pageData);


packages/frontend/src/providers/Tracking/TrackingProvider.tsx
```diff
@@ -178,12 +178,18 @@ export const TrackPage: FC<React.PropsWithChildren<PageData>> = ({
 }) => {
     const { page } = useTracking();
 
+    const pageData = useMemo(() => rest, [
+        rest.name,
+        rest.category,
+        rest.type,
+    ]);
```
notgiorgi:
```suggestion
    const pageData = useMemo(
        () => ({
            name: rest.name,
            category: rest.category,
            type: rest.type,
        }),
        [rest.name, rest.category, rest.type],
    );
```

This should fix the build error, otherwise LGTM

notgiorgi:
@all-contributors add @afranzi for code

Issue #16669, opened 2025-09-01
Update lightdash generate-exposures - compatible with dbt 1.10.0


jesshitchcock:
## Problem:
`lightdash generate-exposures` currently generates exposures with the old dbt YAML configuration that is incompatible with dbt 1.10. 
Starting in dbt 1.10, certain properties like tags have moved from top-level to nested under the config property for better validation and consistency.

This is becoming a blocker for teams upgrading to dbt 1.10, as their exposure generation workflow breaks during compilation.

### Current Issue:
The command generates exposures with tags at the top level:
``` yaml
exposures:
  - name: ld_project_35e2838d_72d2_4e3b_b336_258b2cf8e40f
    ...
    tags:
      - lightdash
      - project
```

This causes dbt 1.10 compilation to fail with:
``` error: dbt1060: Ignored unexpected key "tags". YAML path: tags.```

### Required Fix:

Tags need to be moved under the config property (see docs on this [here](https://docs.getdbt.com/reference/exposure-properties)): 
``` yaml
exposures:
  - name: ld_project_35e2838d_72d2_4e3b_b336_258b2cf8e40f
    ...
    config:
      tags:
        - lightdash
        - project
```

### Proposed Solution:
Update the lightdash generate-exposures command to:

- Generate exposures with the dbt 1.10+ compliant structure (tags under config)
- Optionally detect the dbt version and generate the appropriate format
- Ensure backward compatibility for teams still on older dbt versions

jesshitchcock:
@rephus feature request updated this new info.

PR #17584, opened 2025-10-22
feat: update dbt exposures format for dbt 1.10+ compatibility

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/16669
Screenshot from 2025-10-22 13-13-22

Tested with 1.9 

image

### Description:
Update dbt exposure generation to support dbt 1.10+ format

This PR updates the `generateExposures` handler to detect the dbt version and format exposures accordingly. For dbt 1.10+, tags are now placed under a `config` section rather than at the top level of each exposure.

The PR also updates the example Jaffle Shop demo exposures to use the new format, moving tags under the config section and reorganizing the chart and dashboard exposures.

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/models/lightdash_exposures.yml
+++ b/examples/full-jaffle-shop-demo/dbt/models/lightdash_exposures.yml
@@ -8,79 +8,17 @@ exposures:
     label: Lightdash - Jaffle shop
     description: Lightdash project
     url: http://localhost:3000/projects/3675b69e-8324-4110-bdca-059031aa8da3/home
-    tags:
-      - lightdash
-      - project
     depends_on:
+      - ref('orders')
+      - ref('payments')
+      - ref('customers')
       - ref('plan')
-      - ref('customers')
       - ref('membership')
-      - ref('orders')
-      - ref('payments')
-  - name: ld_chart_01962af0_d72a_4428_913b_2673298d752d
-    type: analysis
-    owner:
-      name: David Attenborough
-      email: ""
-    label: How many users were created each month ?
-    description: A pivot table sample
-    url: >-
-      http://localhost:3000/projects/3675b69e-8324-4110-bdca-059031aa8da3/saved/01962af0-d72a-4428-913b-2673298d752d/view
-    tags:
-      - lightdash
-      - chart
-    depends_on:
-      - ref('plan')
-      - ref('customers')
-      - ref('membership')
-  - name: ld_chart_c53c4c5b_10d3_4b79_a616_0f7797c8fc21
-    type: analysis
-    owner:
-      name: David Attenborough
-      email: ""
-    label: What's our total revenue to date?
-    description: A single number showing the sum of all historical revenue
-    url: >-
-      http://localhost:3000/projects/3675b69e-8324-4110-bdca-059031aa8da3/saved/c53c4c5b-10d3-4b79-a616-0f7797c8fc21/view
-    tags:
-      - lightdash
-      - chart
-    depends_on:
-      - ref('orders')
-      - ref('payments')
-      - ref('customers')
-  - name: ld_chart_be15e7e2_2ad4_49c7_b1fe_40d334e17373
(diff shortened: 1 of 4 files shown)


rephus:
* **#17584** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17584?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17584?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/full-jaffle-shop-demo/dbt/models/lightdash_exposures.yml
```diff
@@ -8,171 +8,381 @@ exposures:
...
-      - ref('plan')
-      - ref('customers')
-      - ref('membership')
       - ref('orders')
       - ref('payments')
-  - name: ld_chart_01962af0_d72a_4428_913b_2673298d752d
+      - ref('customers')
+      - ref('plan')
+      - ref('membership')
+    config:
```
rephus, line 17: most of the changes in this PR are for this example file

packages/cli/src/handlers/generateExposures.ts
```diff
@@ -12,13 +13,21 @@ import { getConfig } from '../config';
...
 import { checkLightdashVersion, lightdashApi } from './dbt/apiClient';
+import { getDbtVersion } from './dbt/getDbtVersion';
 
 type GenerateExposuresHandlerOptions = {
     projectDir: string;
     verbose: boolean;
     output?: string;
 };
 
+const isDbtVersion110OrHigher = (version: string): boolean => {
```
stevelikesmusic: Could we not do something like:

```typescript
const versionNum = parseFloat(version.slice(1))
return versionNum >= 1.10
```

packages/cli/src/handlers/generateExposures.ts
```diff
@@ -53,6 +62,12 @@ export const generateExposuresHandler = async (
     try {
         const absoluteProjectPath = path.resolve(options.projectDir);
 
+        // Detect dbt version to determine format
+        const dbtVersionInfo = await getDbtVersion();
```
stevelikesmusic, line 60: Could `getDbtVersion` return a float field of the actual version in its response? I suppose this breaks if they introduce a patch version number though...

packages/cli/src/handlers/generateExposures.ts
rephus: in fact, we already have a similar method for `isDbtVersion110OrHigher`let me refactor the code to unify the usage

packages/cli/src/handlers/generateExposures.ts
rephus: we could return it in float format, by parsing the dbt response (sometimes they include alpha, so we'll have to use some regex or simliar)... but I'd rather not do this on this PR anyway, since that could cause issues with existing methods. Feel free to open a ticket for that if you want.

packages/cli/src/handlers/generateExposures.ts
rephus: I'll keep the same logic we already use DbtSchemaEditor, which is based on the enum. Your solution should also work though

rephus: hahah :facepalm: 

image

stevelikesmusic: You win this round arithmetic... 

If we wanted, we could use a version enum without string values:
```typescript
enum DbtVersion {
  1_9
  1_10
}
```
which would let us do something like
```typescript
function parseVersion(version: string) {
  return DbVersion[version.slice(1).replace('.', '_')]
}

parseVersion(version) >= DbtVersion.1_10
```

But that would be a bigger refactor than simply re-using the function that already exists.

rephus:
tests passing on different dbt versions 

image

Issue #16595, opened 2025-08-26
Make it possible to let users change parameters in embedded dashboards


magnew:
### Description

Include parameters in the "users can change" setting for embedded dashboards, which would make it possible for dashboard developers let users change parameters on demand. 

Right now parameters work in embedded dashboards, but the ability to change them is disabled there. You can build charts and dashboards with saved parameters, and the saved values will be used in the embedded results.

PR #17588, opened 2025-10-22
feat: parameter support to embedded dashboards backend

pimmey:
Relates to: #16595

### Description:
Adds support for changing parameters in embedded dashboards. This PR introduces a new `canChangeParameters` flag in the JWT content that controls whether users can modify parameters in embedded dashboards. When enabled, user-provided parameters are accepted and combined with dashboard parameters in the embed service.

The implementation includes:
- Adding `canChangeParameters` to the account access object
- Updating the embed controller to accept parameters in API requests
- Modifying the embed service to respect the parameter permissions
- Adding parameter support to chart results, totals, and results table endpoints

This enhancement gives more flexibility to embedded dashboard users while maintaining proper permission controls.

### DIFF
--- a/packages/backend/src/auth/account/account.ts
+++ b/packages/backend/src/auth/account/account.ts
@@ -108,6 +108,7 @@ export const fromJwt = ({
         access: {
             dashboardId: dashboardUuid,
             filtering: decodedToken.content.dashboardFiltersInteractivity,
+            canChangeParameters: decodedToken.content.canChangeParameters,
             controls: userAttributes,
         },
         // Create the fields we're able to set from the JWT
--- a/packages/backend/src/ee/controllers/embedController.ts
+++ b/packages/backend/src/ee/controllers/embedController.ts
@@ -22,6 +22,7 @@ import {
     FieldValueSearchResult,
     Item,
     MetricQueryResponse,
+    ParametersValuesMap,
     SavedChart,
     SavedChartsInfoForDashboardAvailableFilters,
     SortField,
@@ -227,6 +228,7 @@ export class EmbedController extends BaseController {
             dashboardFilters?: DashboardFilters;
             dateZoomGranularity?: DateGranularity;
             dashboardSorts?: SortField[];
+            parameters?: ParametersValuesMap;
         },
     ): Promise<ApiEmbedChartAndResultsResponse> {
         this.setStatus(200);
@@ -242,6 +244,7 @@ export class EmbedController extends BaseController {
                 body.dashboardFilters,
                 body.dateZoomGranularity,
                 body.dashboardSorts,
+                body.parameters,
             ),
         };
     }
@@ -299,6 +302,7 @@ export class EmbedController extends BaseController {
         @Body()
         body: {
             dashboardFilters?: AnyType; // DashboardFilters; temp disable validation
+            parameters?: ParametersValuesMap;
             invalidateCache?: boolean;
         },
     ): Promise<ApiCalculateTotalResponse> {
@@ -313,6 +317,7 @@ export class EmbedController extends BaseController {
                 projectUuid,
                 savedChartUuid,
                 body.dashboardFilters,
+                body.parameters,
                 body.invalidateCache,
             ),
         };
@@ -328,6 +333,7 @@ export class EmbedController extends BaseController {
         @Body()
         body: {
             dashboardFilters?: DashboardFilters;
+            parameters?: ParametersValuesMap;
             columnOrder: string[];
             pivotDimensions?: string[];
             invalidateCache?: boolean;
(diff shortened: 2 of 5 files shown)


pimmey:
* **#17674** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17674?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17616** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17591** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17588** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17588?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17588?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/services/EmbedService/EmbedService.ts
```diff
@@ -959,11 +960,14 @@ export class EmbedService extends BaseService {
 
         const dashboardParameters = getDashboardParametersValuesMap(dashboard);
 
-        // No parameters are passed in embed requests, just combine the saved parameters
+        const acceptedUserParameters =
+            account.access.canChangeParameters && userParameters
```
stevelikesmusic, line 964: Is there a CASL permission that gates Params?

packages/backend/src/auth/account/account.ts
```diff
@@ -108,6 +108,7 @@ export const fromJwt = ({
         access: {
             dashboardId: dashboardUuid,
             filtering: decodedToken.content.dashboardFiltersInteractivity,
+            canChangeParameters: decodedToken.content.canChangeParameters,
```
stevelikesmusic, line 111: Do you expect parameters to evolve similarly to filters? For example, with filters, the embed can define a subset of all available dashboard filters? 

The `access` object is defining behavior like what filters are available vs. are filters enabled at all. One way you could define the API so it could be extended into something more flexible later would be:

```typescript
access: {
  ...
  parameters: {
    enabled: boolean,
  }
}
```

This is the same design that we currently have for filters. 

For the future reference (not a suggestion here), I'm looking at how we can evolve the design of our JWT into something more extensible and secure ([doc](https://linear.app/lightdash/document/embed-single-chart-spec-25f9e6074962)).

packages/common/src/ee/embed/index.ts
```diff
@@ -100,6 +101,7 @@ type CommonEmbedJwtContent = {
         enabled: FilterInteractivityValues | boolean;
         allowedFilters?: string[] | null;
     };
+    canChangeParameters?: boolean;
```
stevelikesmusic, line 104: Would it be worth defining this a parameter interactivity?

packages/backend/src/ee/controllers/embedController.ts
```diff
@@ -227,6 +228,7 @@ export class EmbedController extends BaseController {
             dashboardFilters?: DashboardFilters;
             dateZoomGranularity?: DateGranularity;
             dashboardSorts?: SortField[];
+            parameters?: ParametersValuesMap;
```
stevelikesmusic, line 231: Is it worth adding an e2e test to `embed.cy.ts`?

stevelikesmusic (commented):
If we don't have a CASL check for Parameters, I think it'd helpful to add Parameters to the JWT under interactivity similar to filters.

packages/backend/src/ee/services/EmbedService/EmbedService.ts
pimmey: I believe not!

packages/backend/src/auth/account/account.ts
pimmey: Perhaps eventually we'll make it more granular, but as it stood now, the customers just wanted to ability to change the parameters and not decide what parameters they can change. Pretty sure we'll have to do that in the future 🙂

packages/common/src/ee/embed/index.ts
pimmey: Follow up in the [topmost PR](https://github.com/lightdash/lightdash/pull/17674) 🙂

packages/backend/src/ee/controllers/embedController.ts
pimmey: It might be, but after a quick research it seems like something that might take a bit of an effort (time). I can follow up in a separate PR with e2e test for parameters ✅

packages/backend/src/auth/account/account.ts
stevelikesmusic: I see the change in the top stack!

PR #17603, opened 2025-10-23
feat: add chart embedding support to JWT authentication

rephus:
Closes: https://linear.app/lightdash/issue/CENG-102/update-jwt-to-accept-chart

Creating jwt for chart
Screenshot from 2025-10-23 10-56-36

Attempt to use it . we don't support it on the method yet, but at least it knows it is not a dashboard.

Screenshot from 2025-10-23 10-56-25

### Description:
Add support for chart embedding in the API. This enhancement allows users to generate embed URLs for individual charts, similar to how dashboard embedding works. The implementation includes:

- New JWT content types for chart embedding with appropriate scopes and permissions
- Updated authentication middleware to handle chart content types
- Modified ability checks to properly handle chart-specific permissions
- Added HTTP examples for generating chart embed URLs and calculating chart totals

This change maintains backward compatibility with existing dashboard embedding while extending the embedding functionality to individual charts.

### DIFF
--- a/packages/backend/src/auth/account/account.ts
+++ b/packages/backend/src/auth/account/account.ts
@@ -76,21 +76,30 @@ export const fromJwt = ({
     decodedToken,
     embed,
     source,
+    contentUuid,
+    contentType = 'dashboard',
     dashboardUuid,
     userAttributes,
 }: {
     decodedToken: CreateEmbedJwt;
     embed: OssEmbed;
     source: string;
-    dashboardUuid: string;
+    contentUuid?: string;
+    contentType?: 'dashboard' | 'chart';
+    dashboardUuid?: string;
     userAttributes: UserAccessControls;
 }): AnonymousAccount => {
     const builder = new AbilityBuilder<MemberAbility>(Ability);
     const externalId = getExternalId(decodedToken, source, embed.organization);
 
+    // Support legacy dashboardUuid parameter for backward compatibility
+    const resolvedContentUuid = contentUuid || dashboardUuid;
+    const resolvedContentType = contentType || 'dashboard';
+
     applyEmbeddedAbility(
         decodedToken,
-        dashboardUuid,
+        resolvedContentUuid,
+        resolvedContentType,
         embed,
         externalId,
         builder,
@@ -106,8 +115,17 @@ export const fromJwt = ({
         organization: embed.organization,
         embed,
         access: {
-            dashboardId: dashboardUuid,
-            filtering: decodedToken.content.dashboardFiltersInteractivity,
+            contentId: resolvedContentUuid,
+            contentType: resolvedContentType,
+            // Legacy field for backward compatibility
+            dashboardId:
+                resolvedContentType === 'dashboard'
+                    ? resolvedContentUuid
+                    : undefined,
+            filtering:
+                decodedToken.content.type === 'dashboard'
+                    ? decodedToken.content.dashboardFiltersInteractivity
+                    : undefined,
             controls: userAttributes,
         },
         // Create the fields we're able to set from the JWT
--- a/packages/backend/src/ee/controllers/embedController.ts
+++ b/packages/backend/src/ee/controllers/embedController.ts
@@ -9,6 +9,7 @@ import {
     ApiSuccessEmpty,
     assertEmbeddedAuth,
     CacheMetadata,
+    CommonEmbedJwtContent,
     CreateEmbedJwt,
     CreateEmbedRequestBody,
     Dashboard,
@@ -54,7 +55,7 @@ export type ApiEmbedDashboardResponse = {
     status: 'ok';
     results: Dashboard & {
         // declare type as TSOA doesn't understand zod type InteractivityOptions
-        dashboardFiltersInteractivity?: CreateEmbedJwt['content']['dashboardFiltersInteractivity'];
+        dashboardFiltersInteractivity?: CommonEmbedJwtContent['dashboardFiltersInteractivity'];
         canExportCsv?: boolean;
         canExportImages?: boolean;
     };
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -30,7 +30,9 @@ import {
     getItemId,
     InteractivityOptions,
     IntrinsicUserAttributes,
+    isChartContent,
     isDashboardChartTileType,
+    isDashboardContent,
     isDashboardSlugContent,
     isDashboardSqlChartTile,
     isExploreError,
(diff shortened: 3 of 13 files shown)


rephus:
* **#17603** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/controllers/embedController.ts
```diff
@@ -54,7 +54,10 @@ export type ApiEmbedDashboardResponse = {
     status: 'ok';
     results: Dashboard & {
         // declare type as TSOA doesn't understand zod type InteractivityOptions
-        dashboardFiltersInteractivity?: CreateEmbedJwt['content']['dashboardFiltersInteractivity'];
+        dashboardFiltersInteractivity?: {
```
rephus: not sure why this changed

packages/backend/src/ee/services/EmbedService/EmbedService.ts
```diff
@@ -370,13 +440,18 @@ export class EmbedService extends BaseService {
...
+        const canExportCsv =
+            'canExportCsv' in content ? content.canExportCsv : undefined;
+        const canExportImages =
+            'canExportImages' in content ? content.canExportImages : undefined;
+        const canExportPagePdf =
+            'canExportPagePdf' in content
+                ? content.canExportPagePdf
+                : undefined;
+        const canDateZoom =
+            'canDateZoom' in content ? content.canDateZoom : undefined;
```
stevelikesmusic: Is this to make tsc happy? It seems overkill.

packages/common/src/types/auth.ts
```diff
@@ -80,8 +80,12 @@ export type UserAccessControls = {
 };
 
 export type DashboardAccess = {
```
stevelikesmusic: Can you change the name of the type to something like EmbedAccess?

packages/backend/src/ee/services/EmbedService/EmbedService.ts
```diff
@@ -427,6 +502,13 @@ export class EmbedService extends BaseService {
         checkPermissions: boolean = true,
     ): Promise<DashboardAvailableFilters> {
         const dashboardUuid = account.access.dashboardId;
```
stevelikesmusic: We should probably change these accessors. Makes me think though if we should add another field like `content`? 

Something like:
```typescript
account.access.content.id
```

What do you think? Extensible or too complicated?

packages/common/src/types/auth.ts
```diff
@@ -80,8 +80,12 @@ export type UserAccessControls = {
...
-    /** The dashboard ID the account has access to */
-    dashboardId: string;
+    /** The content ID the account has access to (dashboard or chart) */
+    contentId?: string;
+    /** The type of content (dashboard or chart) */
+    contentType?: 'dashboard' | 'chart';
+    /** @deprecated Use contentId instead. The dashboard ID the account has access to */
+    dashboardId?: string;
```
stevelikesmusic: Do we need to deprecate since this object is only used internally? I wonder if we could create a new `content` object with `content: { type, id|idOrSlug }`?

packages/common/src/ee/embed/index.ts
```diff
@@ -116,8 +145,20 @@ type EmbedJwtContentDashboardSlug = CommonEmbedJwtContent & {
     dashboardSlug: string;
 };
 
+type EmbedJwtContentChartUuid = CommonChartEmbedJwtContent & {
+    contentId: string;
+};
+
+type EmbedJwtContentChartSlug = CommonChartEmbedJwtContent & {
+    contentId: string;
+};
```
stevelikesmusic: Do we need two types for this?

packages/common/src/ee/embed/index.ts
```diff
@@ -116,8 +145,20 @@ type EmbedJwtContentDashboardSlug = CommonEmbedJwtContent & {
...
+
 export type CreateEmbedJwt = {
```
stevelikesmusic, line 152: I think this type must be used in the Settings UI for showing JWT minting snippets. Those fields are all in the context of Dashboards, so it's throwing in the test because charts are missing some of the fields we use in the snippet. I think we'll need to have one more layer of abstraction like `DashboardEmbedJwt` and `ChartEmbedJwt`. Then we can use the Dashboard type in the snippet. I'll use the Chart type to add a chart-specific snippet.

https://github.com/lightdash/lightdash/blob/ec6ad8f30a1b3c7fec5ca39f4af33534bb5de528/packages/frontend/src/ee/features/embed/SettingsEmbed/EmbedCodeSnippet.tsx#L189-L200

rephus:
jwt tokens should be valid for non embed urls now, right ? For all the endpoints that support account

packages/common/src/authorization/jwtAbility.ts
```diff
@@ -6,7 +6,8 @@ import type { MemberAbility } from './types';
 
 type EmbeddedAbilityBuilderPayload = {
     embedUser: CreateEmbedJwt;
-    dashboardUuid: string;
+    contentUuid: string;
+    contentType: 'dashboard' | 'chart';
```
stevelikesmusic, line 10: What if we grouped this into `content`?

packages/common/src/authorization/jwtAbility.ts
```diff
@@ -18,42 +19,59 @@ type EmbeddedAbilityBuilder = (
...
-        projectUuid: embed.projectUuid,
-        isPrivate: false,
-    });
+    if (contentType === 'chart') {
+        can('view', 'SavedChart', {
+            uuid: contentUuid,
+            organizationUuid: organization.organizationUuid,
+            projectUuid: embed.projectUuid,
+        });
+    }
```
stevelikesmusic, line 80: I might put this in a separate `chartAbilities` function rather than in `dashboardAbilities` now that we're specifically handling charts.

packages/common/src/authorization/jwtAbility.ts
```diff
@@ -90,11 +115,26 @@ const exportAbilities: EmbeddedAbilityBuilder = ({
...
-        can('export', 'Dashboard', {
-            organizationUuid: organization.organizationUuid,
-            type: 'csv',
-        });
+    const canExportCsv =
+        'canExportCsv' in content ? content.canExportCsv : undefined;
+    const canExportPagePdf =
+        'canExportPagePdf' in content ? content.canExportPagePdf : undefined;
+    const canExportImages =
+        'canExportImages' in content ? content.canExportImages : undefined;
```
stevelikesmusic: Would it be worth creating a typeguard for dashboard vs. chart?

packages/backend/src/auth/account/account.ts
```diff
@@ -76,21 +76,30 @@ export const fromJwt = ({
...
+    contentType?: 'dashboard' | 'chart';
+    dashboardUuid?: string;
     userAttributes: UserAccessControls;
 }): AnonymousAccount => {
     const builder = new AbilityBuilder<MemberAbility>(Ability);
     const externalId = getExternalId(decodedToken, source, embed.organization);
 
+    // Support legacy dashboardUuid parameter for backward compatibility
+    const resolvedContentUuid = contentUuid || dashboardUuid;
+    const resolvedContentType = contentUuid ? contentType : 'dashboard';
```
stevelikesmusic: Could this simply be `contentType || 'dashboard'`?

packages/backend/src/ee/services/EmbedService/EmbedService.ts
rephus: Agree, this is because the type can be a dashboard or chart (without datezoom properties and others). I fixed this by checking the type first.

rephus:
Closed in favor of https://github.com/lightdash/lightdash/pull/17615

PR #17641, opened 2025-10-24
fix: prevent refetch loop on email verification page

pimmey:
## Summary
  Fixes infinite request loop on `/verify-email` page when session expires.

  ## The Issue
  When users are on the `/verify-email` page and their session expires, three endpoints were looping infinitely:
  1. `/api/v1/user/me/email/status`
  2. `/api/v1/org/projects`
  3. `/api/v1/org`

  **The loop mechanism:**
  1. Query gets 401 → global error handler invalidates `['health']`
  2. Health refetches → returns `isAuthenticated: true` (session cookie still present but expired)
  3. Component re-renders → queries become enabled again → refetch → 401 → repeat

  ## Changes Made

  ### 1. Page.tsx - Conditional project fetching
  Don't fetch `activeProjectUuid` and `projects` when `withNavbar={false}`.

  **Fixes:** The projects and organization endpoint loops.

  ### 2. useEmailVerification.ts - Prevent aggressive refetching
  Added `refetchOnMount: false` and `refetchOnReconnect: false` to `useEmailStatus` query.

  **Fixes:** The email/status endpoint loop.

  **Trade-off:** Loses automatic refetch-on-mount, but this is acceptable because:
  - Query only needs to run once on page load
  - Updates happen via manual `invalidateQueries` when user submits verification code
  - Defensive safeguard against loops on a critical auth page

  ### 3. VerifyEmailForm.tsx - Remove duplicate query call
  Pass email status data from parent instead of calling `useEmailStatus` twice.

  **Why:** Cleaner code, prevents multiple query observers.

  ## Important Note

  **This is a workaround, not a root cause fix.**

  The architectural issue is that the global 401 handler invalidates health for ALL 401 errors, but:
  - The `/health` endpoint doesn't require authentication and never returns 401
  - It relies on the presence of a session cookie, which may still exist even when expired
  - This creates a scenario where `isAuthenticated` stays `true` even though the session is invalid

  A proper fix would require:
  - Rethinking the global 401 handling strategy (query-specific error handling vs global)
  - OR making the `/health` endpoint actually return 401 when sessions are invalid
  - OR using a different mechanism to detect session expiry

  This workaround prevents the immediate production issue while we can architect a better solution.

### DIFF
--- a/packages/frontend/src/hooks/useEmailVerification.ts
+++ b/packages/frontend/src/hooks/useEmailVerification.ts
@@ -32,6 +32,11 @@ export const useEmailStatus = (enabled: boolean) =>
         queryKey: ['email_status'],
         queryFn: () => getEmailStatusQuery(),
         enabled,
+        // Prevent infinite loop on /verify-email page when session has issues
+        // This query only needs to run once on mount - verification updates via
+        // manual invalidation (useVerifyEmail onSuccess)
+        refetchOnMount: false,
+        refetchOnReconnect: false,
     });
 
 export const useOneTimePassword = () => {
--- a/packages/frontend/src/components/RegisterForms/VerifyEmailForm.tsx
+++ b/packages/frontend/src/components/RegisterForms/VerifyEmailForm.tsx
@@ -1,3 +1,4 @@
+import { type EmailStatusExpiring } from '@lightdash/common';
 import {
     Alert,
     Anchor,
@@ -12,7 +13,6 @@ import { IconAlertCircle } from '@tabler/icons-react';
 import { useEffect, type FC } from 'react';
 import Countdown, { zeroPad } from 'react-countdown';
 import {
-    useEmailStatus,
     useOneTimePassword,
     useVerifyEmail,
 } from '../../hooks/useEmailVerification';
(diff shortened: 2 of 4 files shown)


pimmey:
* **#17641** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17641?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17641?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/useEmailVerification.ts
```diff
@@ -32,6 +32,11 @@ export const useEmailStatus = (enabled: boolean) =>
         queryKey: ['email_status'],
         queryFn: () => getEmailStatusQuery(),
         enabled,
+        // Prevent infinite loop on /verify-email page when session has issues
+        // This query only needs to run once on mount - verification updates via
+        // manual invalidation (useVerifyEmail onSuccess)
+        refetchOnMount: false,
+        refetchOnReconnect: false,
```
notgiorgi, line 39: Does this also need `staleTime`?

pimmey: We don't strictly need it, but adding `staleTime: Infinity` would make the intent clearer, I guess. But the behaviour should be the same :)

PR #17652, opened 2025-10-24
fix: remove jwt from URL hash

stevelikesmusic:
Closes: CENG-116

### Description:

We had been relying on implicit behavior in DashboardProvider to remove the JWT from the URL hash. Something changed recently and we no longer implicitly make that redirect.

This PR explicitly removes the JWT from the URL. We use the browser directly rather than react-router since we're already using window.location to get the URL hash.

### DIFF
--- a/packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
+++ b/packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
@@ -1,7 +1,7 @@
 import { type LanguageMap, type SavedChart } from '@lightdash/common';
 import get from 'lodash/get';
 import { useEffect, useMemo, useState, type FC } from 'react';
-import { useLocation, useParams } from 'react-router';
+import { useLocation, useNavigate, useParams } from 'react-router';
 import { useAccount } from '../../../hooks/user/useAccount';
 import { useAbilityContext } from '../../../providers/Ability/useAbilityContext';
 import {
@@ -41,10 +41,20 @@ const EmbedProvider: FC<React.PropsWithChildren<Props>> = ({
     const { data: account, isLoading } = useAccount();
     const ability = useAbilityContext();
     const params = useParams();
+    const navigate = useNavigate();
     const projectUuid = projectUuidFromProps || params.projectUuid;
     const location = useLocation();
     const { dispatchEmbedEvent } = useEmbedEventEmitter();
     const mode: EmbedMode = encodedToken ? 'sdk' : 'direct';
+
+    // Remove the token from the URL.
+    useEffect(() => {
+        if (mode === 'direct' && location.hash) {
+            void navigate(location.pathname + location.search, {
+                replace: true,
+            });
+        }
+    }, [mode, location, navigate]);
 
     // We sync embed UI changes with the URL just as with the main app.
     // For iframe embeds only, we emit messages to the parent window.


stevelikesmusic:
* **#17652** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
```diff
@@ -46,6 +46,11 @@ const EmbedProvider: FC<React.PropsWithChildren<Props>> = ({
     const { dispatchEmbedEvent } = useEmbedEventEmitter();
     const mode: EmbedMode = encodedToken ? 'sdk' : 'direct';
 
+    // Remove the token from the URL.
+    if (window.location.hash === `#${embedToken}`) {
```
owlas: Is this provider used in the react SDK or in any setting where we don't want to manipulate the URL? 

In the case it's only used in the app do we need to assert the token? Or just wipe the hash no matter what the value?

stevelikesmusic: Yes, this provider would render in both SDK and iframe. I can add more specificity to the check. 

The token should only be there when loading the iframe. I recently added a `mode` field to distinguish between SDK and /embed in the app (iframe or direct usage of get-embed-url)

packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
```diff
@@ -41,11 +41,21 @@ const EmbedProvider: FC<React.PropsWithChildren<Props>> = ({
...
     const mode: EmbedMode = encodedToken ? 'sdk' : 'direct';
 
+    // Remove the token from the URL.
+    useEffect(() => {
+        if (mode === 'direct' && location.hash) {
+            void navigate(location.pathname + location.search, {
+                replace: true,
+            });
+        }
+    }, [mode, location, navigate]);
```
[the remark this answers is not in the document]
rephus: I don't think this is relevant

packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
```diff
@@ -41,11 +41,21 @@ const EmbedProvider: FC<React.PropsWithChildren<Props>> = ({
...
     const ability = useAbilityContext();
     const params = useParams();
+    const navigate = useNavigate();
     const projectUuid = projectUuidFromProps || params.projectUuid;
     const location = useLocation();
     const { dispatchEmbedEvent } = useEmbedEventEmitter();
     const mode: EmbedMode = encodedToken ? 'sdk' : 'direct';
 
+    // Remove the token from the URL.
+    useEffect(() => {
```
rephus, line 51: is not done already somewhere else ?

packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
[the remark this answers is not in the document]
stevelikesmusic: Agreed. `location` is a memoized object from react-router. if one field changes the whole object is new.

packages/frontend/src/ee/providers/Embed/EmbedProvider.tsx
stevelikesmusic: From what I could tell, this was being done implicitly with a `navigate` in DashboardProvider. At some point, that was refactored and we no longer did the navigate, so the JWT remains in the URL. This explicitly removes the JWT from the URL in iframe and /embed

stevelikesmusic:
### Merge activity

* **Nov 3, 11:00 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17652).
* **Nov 3, 11:00 AM UTC**: @stevelikesmusic merged this pull request with [Graphite](https://app.graphite.dev/github/pr/lightdash/lightdash/17652).

PR #17654, opened 2025-10-24
fix KnexSessionStore configuration keys

kgeis:
Closes: n/a

### Description:
Update KnexSessionStore configuration keys to match case in the [connect-session-knex documentation](https://www.npmjs.com/package/connect-session-knex).


ZeRego:
The configuration is correct, as we are using v3.0.1. https://github.com/gx0r/connect-session-knex/tree/v3.0.1

PR #17673, opened 2025-10-27
feat: add --organization-credentials when creating project on the CLI

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-112/add-organization-credentials-in-the-cli

Creating a project using org credential on the CLI 

Screenshot from 2025-10-27 15-51-23
Screenshot from 2025-10-27 15-52-36

When trying to use a org credential that doesn't exist

Screenshot from 2025-10-27 15-51-09

### Description:
This PR adds a new endpoint to fetch organization warehouse credentials summaries that are accessible to all organization members, not just those with manage permissions. The summaries include only non-sensitive information (name, description, type, and UUID).

Key changes:
- Added a new `getAllSummaries` method to `OrganizationWarehouseCredentialsService`
- Updated the controller to accept a `summary` query parameter
- Added permission for all organization members to view credential summaries
- Enhanced the CLI to support using organization credentials by name with the new `--organization-credentials` flag
- Fixed error message to distinguish between view and manage permissions

This enables regular organization members to see available credential options without exposing sensitive connection details, and allows them to reference credentials by name in the CLI.

### DIFF
--- a/packages/backend/src/ee/controllers/OrganizationWarehouseCredentialsController.ts
+++ b/packages/backend/src/ee/controllers/OrganizationWarehouseCredentialsController.ts
@@ -2,6 +2,7 @@ import {
     ApiErrorPayload,
     ApiOrganizationWarehouseCredentialsListResponse,
     ApiOrganizationWarehouseCredentialsResponse,
+    ApiOrganizationWarehouseCredentialsSummaryListResponse,
     ApiSuccessEmpty,
     CreateOrganizationWarehouseCredentials,
     UpdateOrganizationWarehouseCredentials,
@@ -15,6 +16,7 @@ import {
     Patch,
     Path,
     Post,
+    Query,
     Request,
     Response,
     Route,
@@ -37,14 +39,30 @@ export class OrganizationWarehouseCredentialsController extends BaseController {
      * Get all warehouse credentials for the current organization
      * @summary List warehouse credentials
      * @param req express request
+     * @param summary If true, returns only summaries (name, type) accessible to all members. If false/undefined, returns full credentials requiring manage permission.
      */
     @Middlewares([allowApiKeyAuthentication, isAuthenticated])
     @Get()
     @OperationId('ListOrganizationWarehouseCredentials')
     async getAll(
         @Request() req: express.Request,
-    ): Promise<ApiOrganizationWarehouseCredentialsListResponse> {
+        @Query() summary?: boolean,
+    ): Promise<
+        | ApiOrganizationWarehouseCredentialsListResponse
+        | ApiOrganizationWarehouseCredentialsSummaryListResponse
+    > {
         this.setStatus(200);
+
+        if (summary) {
+            return {
+                status: 'ok',
+                results:
+                    await this.getOrganizationWarehouseCredentialsService().getAllSummaries(
+                        req.account!,
+                    ),
+            };
+        }
+
         return {
             status: 'ok',
             results:
--- a/packages/backend/src/ee/services/OrganizationWarehouseCredentialsService.ts
+++ b/packages/backend/src/ee/services/OrganizationWarehouseCredentialsService.ts
@@ -7,6 +7,7 @@ import {
     NotImplementedError,
     OpenIdIdentityIssuerType,
     OrganizationWarehouseCredentials,
+    OrganizationWarehouseCredentialsSummary,
     SnowflakeAuthenticationType,
     UpdateOrganizationWarehouseCredentials,
     WarehouseTypes,
@@ -57,7 +58,7 @@ export class OrganizationWarehouseCredentialsService extends BaseService {
             )
         ) {
             throw new ForbiddenError(
-                'You do not have permission to view organization warehouse credentials',
+                'You do not have permission to manage organization warehouse credentials',
             );
         }
     }
@@ -70,6 +71,47 @@ export class OrganizationWarehouseCredentialsService extends BaseService {
         return this.organizationWarehouseCredentialsModel.getAllByOrganizationUuid(
             organizationUuid,
         );
+    }
+
+    /**
+     * Get summaries of all organization warehouse credentials
+     * This is accessible to all organization editor members (no manage permission required)
+     * Returns only non-sensitive information: name, description, and warehouse type
+     */
+    async getAllSummaries(
+        account: Account,
+    ): Promise<OrganizationWarehouseCredentialsSummary[]> {
+        const { organizationUuid } = account.organization;
+        if (!organizationUuid) {
+            throw new ForbiddenError('User must be in an organization');
+        }
+
+        if (
+            account.user.ability.cannot(
+                'view',
+                subject('OrganizationWarehouseCredentials', {
+                    organizationUuid,
+                }),
+            )
+        ) {
+            throw new ForbiddenError(
+                'You do not have permission to view organization warehouse credentials',
+            );
+        }
+
+        const allCredentials =
+            await this.organizationWarehouseCredentialsModel.getAllByOrganizationUuid(
+                organizationUuid,
+            );
+
+        // Return only the summary fields
+        return allCredentials.map((cred) => ({
+            organizationWarehouseCredentialsUuid:
+                cred.organizationWarehouseCredentialsUuid,
+            name: cred.name,
+            description: cred.description,
+            warehouseType: cred.warehouseType,
+        }));
     }
 
     async get(
@@ -171,9 +213,8 @@ export class OrganizationWarehouseCredentialsService extends BaseService {
 
         // Verify ownership before update
         const existing =
-            await this.organizationWarehouseCredentialsModel.getByUuid(
+            await this.organizationWarehouseCredentialsModel.getByUuidWithSensitiveData(
                 credentialsUuid,
-                true, // withSensitiveData
             );
         if (existing.organizationUuid !== organizationUuid) {
             throw new ForbiddenError(
--- a/packages/backend/src/models/OrganizationWarehouseCredentialsModel.ts
+++ b/packages/backend/src/models/OrganizationWarehouseCredentialsModel.ts
@@ -85,21 +85,9 @@ export class OrganizationWarehouseCredentialsModel {
         );
     }
 
-    async getByUuid(
+    private async _getByUuid(
         uuid: string,
-        withSensitiveData?: false,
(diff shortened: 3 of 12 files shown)


rephus:
* **#17673** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/services/OrganizationWarehouseCredentialsService.ts
```diff
@@ -72,6 +73,47 @@ export class OrganizationWarehouseCredentialsService extends BaseService {
...
+        account: Account,
+    ): Promise<OrganizationWarehouseCredentialsSummary[]> {
+        const { organizationUuid } = account.organization;
+        if (!organizationUuid) {
+            throw new ForbiddenError('User must be in an organization');
+        }
+
+        if (
+            account.user.ability.cannot(
+                'view',
```
rephus, line 91: new "view" permission for editors

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -745,13 +745,13 @@ export class ProjectService extends BaseService {
...
         },
-    >(args: T, userUuid: string): Promise<T> {
+    >(args: T, userUuid: string, organizationUuid: string): Promise<T> {
         // If using organization credentials, load them from the organization table
         const organizationWarehouseCredentialsUuid =
-            args.warehouseConnection.type === WarehouseTypes.SNOWFLAKE
-                ? args.organizationWarehouseCredentialsUuid ||
-                  args.warehouseConnection.organizationWarehouseCredentialsUuid
-                : undefined;
+            args.organizationWarehouseCredentialsUuid ||
```
rephus, line 756: we need to confirm this args.organizationWarehouseCredentialsUuid belongs to the user

packages/backend/src/ee/controllers/OrganizationWarehouseCredentialsController.ts
```diff
@@ -37,14 +39,30 @@ export class OrganizationWarehouseCredentialsController extends BaseController {
...
     @Middlewares([allowApiKeyAuthentication, isAuthenticated])
     @Get()
     @OperationId('ListOrganizationWarehouseCredentials')
     async getAll(
         @Request() req: express.Request,
-    ): Promise<ApiOrganizationWarehouseCredentialsListResponse> {
+        @Query() summary?: boolean,
+    ): Promise<
+        | ApiOrganizationWarehouseCredentialsListResponse
+        | ApiOrganizationWarehouseCredentialsSummaryListResponse
```
rephus, line 52: I'm not a big fan of the composed  response... but not sure how to make it better

packages/backend/src/ee/controllers/OrganizationWarehouseCredentialsController.ts
```diff
@@ -37,14 +39,30 @@ export class OrganizationWarehouseCredentialsController extends BaseController {
...
      * @param req express request
+     * @param summary If true, returns only summaries (name, type) accessible to all members. If false/undefined, returns full credentials requiring manage permission.
      */
     @Middlewares([allowApiKeyAuthentication, isAuthenticated])
     @Get()
     @OperationId('ListOrganizationWarehouseCredentials')
     async getAll(
         @Request() req: express.Request,
-    ): Promise<ApiOrganizationWarehouseCredentialsListResponse> {
+        @Query() summary?: boolean,
```
rephus, line 49: We need a new endpoint that only returns the list of organization credentials, even for editors, so they can use them on the 

The problem is that any other endpoint like `/summary` will conflict with `@Get('{credentialsUuid}')` endpoint. 

We can't modify this existing endpoint to return simply the summary because we need the credentials in the UI

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: fixed

packages/backend/src/models/OrganizationWarehouseCredentialsModel.ts
```diff
@@ -132,6 +132,57 @@ export class OrganizationWarehouseCredentialsModel {
         return this.convertToOrganizationWarehouseCredentials(result);
     }
 
+    async getByName(
```
rephus: crazy JS overload method . I don't like it

rephus:
Unrelated snowflake sql runner error 
image https://lightdash.slack.com/archives/C02GQKJK84Q/p1761651240301579

packages/backend/src/ee/controllers/OrganizationWarehouseCredentialsController.ts
stevelikesmusic: Would it be possible to make certain required fields optional and have a single type? It looks like the summary is just picking fields from the original list type?

packages/backend/src/ee/services/OrganizationWarehouseCredentialsService.ts
stevelikesmusic: Does this need to be added to scopes?

packages/backend/src/models/OrganizationWarehouseCredentialsModel.ts
```diff
@@ -85,21 +85,61 @@ export class OrganizationWarehouseCredentialsModel {
         );
     }
 
-    async getByUuid(
+    private async _getByUuid(
```
stevelikesmusic, line 88: Why the refactor throughout this file? We added `getByNameWithSensitiveData`, but it doesn't look to be used in this PR?

packages/cli/src/handlers/createProject.ts
```diff
@@ -65,6 +66,51 @@ const askPermissionToStoreWarehouseCredentials = async (): Promise<boolean> => {
...
+        if (!credential) {
+            const availableNames =
+                credentialsList.length > 0
+                    ? credentialsList
+                          .map((c) => `  - ${c.name} (${c.warehouseType})`)
+                          .join('\n')
+                    : '  (none available)';
+            throw new Error(
+                `Organization warehouse credentials "${name}" not found.\n\nAvailable credentials:\n${availableNames}`,
+            );
```
stevelikesmusic, line 96: Why is it that in the ProjectService, we give an opaque error message and log orgId details, but here we provide the whole list of acceptable names? Is the assumption that an attacker would only use the HTTP API but not the CLI?

packages/cli/src/handlers/createProject.ts
```diff
@@ -155,7 +230,9 @@ export const createProject = async (
         spinner?.start();
     }
 
-    const project: CreateProjectOptionalCredentials = {
+    const project: CreateProjectOptionalCredentials & {
+        organizationWarehouseCredentialsUuid?: string;
```
stevelikesmusic: Why add this one-off field here instead of adding it to the CreateProjectOptionalCredentials type?

packages/backend/src/ee/services/OrganizationWarehouseCredentialsService.ts
rephus: I added this to organizationAbilities,  https://github.com/lightdash/lightdash/pull/17673/files#diff-04939cb1d2a7a996362405be79b800f680e36e6ac7e8b0268743ed3b42965b61R219

Since this is an **organization** ability, this doesn't need to go anywhere else for now, right ?

packages/cli/src/index.ts
```diff
@@ -438,6 +438,10 @@ program
         'Skip copying content from the source project',
         false,
     )
+    .option(
+        '--organization-credentials <name>',
+        'Use organization warehouse credentials with the specified name (Enterprise Edition feature)',
```
stevelikesmusic, line 443: I always forget, EE is Cloud Pro and up or just Enterprise?

packages/backend/src/ee/services/OrganizationWarehouseCredentialsService.ts
stevelikesmusic: Ah you're right! This is an org ability, not project. My mistake.

packages/backend/src/ee/controllers/OrganizationWarehouseCredentialsController.ts
rephus: yeah, but that would mean we'll have to refactor all the places where the List type is used, to take an optional credentials. Let me see how dificult that is.

packages/backend/src/models/OrganizationWarehouseCredentialsModel.ts
rephus: I didn't like the way it was overloading the method, claude tried to do the same thing on another method, so I made it refactor both. What do you think ? do you prefer the old way ?

stevelikesmusic: Mostly curious about the reason. I often find it a little harder to reason about with `publicMethod` and `_publicMethod`. But I don't have a strong opinion honestly.

packages/cli/src/index.ts
rephus: EE is "you need an license to run this" on localhost, but it works on our cloud

packages/cli/src/handlers/createProject.ts
rephus: this method in the CLI is only to get the org credential uuid from the credential name (and confirm it exists) . Org credential names are available for all editors in the org. Then, we pass the org credential uuid into the create project method; in there we confirm that org crednetial uuid matches the right org. 

I don't think there is any vulnerability here, but let me know if you found one.

packages/cli/src/handlers/createProject.ts
rephus: YOu were right, organizationWarehouseCredentialsUuid was already part of the PRoject type, so not needed here.

packages/cli/src/handlers/createProject.ts
stevelikesmusic: All good. I didn't spot a vulnerability. Just curious about difference in how we handled errors

PR #17691, opened 2025-10-28
feat: add parameter-based conditional formatting for metrics

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-55/use-lightdash-parameters-in-field-formatting

```
with_currency_parameter: 
              type: sum 
              format: '${ld.parameters.currency=="usd"?"$":""}${ld.parameters.currency=="eur"?"€":""}${ld.parameters.currency=="gbp"?"£":""}0,0.00'
              sql: |
                CASE 
                  WHEN ${ld.parameters.currency} = 'usd' THEN ${event_id} * 1
                  ELSE ${event_id} * 0.86
                END
            with_multiple_parameters: 
              type: sum 
              # only shows $ currency
              format: '${ld.parameters.currency=="usd"?"$":""}0,0${ld.parameters.decimals=="true"?".00":""}'
            with_decimals_format: 
              type: sum 
              format: '0,0${ld.parameters.decimals=="true"?".00":""}'
            with_mixed_parameter: 
              type: sum 
              # Invalid parameter, using the wrong value for decimals
              format: '${ld.parameters.decimals=="usd"?"$":"-"}0,0.00'
```

[Screencast from 2025-10-28 16-52-40.webm](https://github.com/user-attachments/assets/7078d3ab-345e-43bf-8fe3-a5232be43288)

With simple symbol replacement
```
            with_simple_currency_symbol: 
              type: sum 
              # only shows plain parameter symbol (it needs to be compatible with numfmt syntax)
              format: '${ld.parameters.currency_symbol}0,0.00'
```
image

### Description:
Added support for conditional formatting with parameters in chart visualizations. This allows users to dynamically change formatting based on parameter values, such as currency symbols and decimal places.

The implementation:
- Added new parameters for currency and decimal formatting in the Jaffle Shop demo
- Enhanced the formatting system to evaluate conditional expressions like `${ld.parameters.currency=="usd"?"$":"€"}0,0.00`
- Updated all visualization components to pass parameter values to formatting functions
- Added example metrics in the demo that use parameter-based formatting

This enables users to create more dynamic and interactive visualizations that respond to parameter changes without requiring chart reconfiguration.

### DIFF
--- a/packages/common/src/utils/conditionalFormatExpressions.ts
+++ b/packages/common/src/utils/conditionalFormatExpressions.ts
@@ -11,11 +11,14 @@ import { LightdashParameters } from '../compiler/parameters';
  *   - "currency" -> "currency"
  */
 function stripParameterPrefix(paramName: string): string {
-    if (paramName.startsWith(LightdashParameters.PREFIX_SHORT)) {
-        return paramName.substring(LightdashParameters.PREFIX_SHORT.length);
+    const shortPrefix = `${LightdashParameters.PREFIX_SHORT}.`;
+    const longPrefix = `${LightdashParameters.PREFIX}.`;
+
+    if (paramName.startsWith(shortPrefix)) {
+        return paramName.substring(shortPrefix.length);
     }
-    if (paramName.startsWith(LightdashParameters.PREFIX)) {
-        return paramName.substring(LightdashParameters.PREFIX.length);
+    if (paramName.startsWith(longPrefix)) {
+        return paramName.substring(longPrefix.length);
     }
     return paramName;
 }
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -7,6 +7,7 @@ import {
     isTextFormat,
     isValidFormat,
 } from 'numfmt';
+import { LightdashParameters } from '../compiler/parameters';
 import {
     CompactConfigMap,
     CustomFormatType,
@@ -768,36 +769,44 @@ export function formatItemValue(
     if (value === null) return '∅';
     if (value === undefined) return '-';
     if (item) {
-        // Check if format uses parameter placeholders
-        const hasParameterPlaceholders =
-            hasValidFormatExpression(item) &&
-            item.format.includes('${ld.parameters');
-
         if (hasValidFormatExpression(item)) {
-            // Only apply format expression if parameters are provided when needed
-            if (!hasParameterPlaceholders || parameters) {
-                // Evaluate conditional expressions in format string if parameters are provided
-                const formatExpression = parameters
-                    ? evaluateConditionalFormatExpression(
-                          item.format,
-                          parameters,
-                      )
-                    : item.format;
+            // Check if format uses parameter placeholders
+            const hasParameterPlaceholders =
+                item.format.includes(
+                    `\${${LightdashParameters.PREFIX_SHORT}`,
+                ) || item.format.includes(`\${${LightdashParameters.PREFIX}`);
 
-                try {
-                    const result = formatValueWithExpression(
-                        formatExpression,
-                        value,
-                    );
-                    return result;
-                } catch (error) {
-                    // Fall through to default formatting below
+            // NEW: Handle parameter-based formats separately
+            if (hasParameterPlaceholders) {
+                // If parameters are provided, evaluate and apply the format
+                if (parameters) {
+                    const formatExpression =
+                        evaluateConditionalFormatExpression(
+                            item.format,
+                            parameters,
+                        );
+                    try {
+                        const result = formatValueWithExpression(
+                            formatExpression,
+                            value,
+                        );
+                        return result;
+                    } catch (error) {
+                        // If evaluation fails, fall back to default formatting
+                        return applyDefaultFormat(value);
+                    }
+                } else {
+                    // No parameters provided but format needs them - use default formatting
+                    return applyDefaultFormat(value);
                 }
             }
-            // If format uses parameters but none are provided, use default formatting
-            // and skip CustomFormat extraction from the format string
-            if (hasParameterPlaceholders && !parameters) {
-                return applyDefaultFormat(value);
+
+            // EXISTING: Handle non-parameter formats (unchanged behavior)
+            try {
+                const result = formatValueWithExpression(item.format, value);
+                return result;
+            } catch (error) {
+                // Fall through to custom format handling below
             }
         }
 


rephus:
* **#17691** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17691?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17691?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -487,8 +488,11 @@ export class ExploreCompiler {
 
         const compiledSql = compiledMetric.sql;
 
-        // Extract parameter references from metric sql
-        const parameterReferences = getParameterReferences(compiledSql);
+        // Extract parameter references from both SQL and format string
+        const parameterReferences = getParameterReferencesFromSqlAndFormat(
```
rephus, line 492: Get parameter references from format, this will appear the parameter int he UI even if the parameter is only used for `format` , not `sql`

packages/common/src/compiler/parameters.test.ts
```diff
@@ -114,18 +114,20 @@ describe('getParameterReferences', () => {
             expect(result).toEqual([]);
         });
 
-        it('should reject parameters with invalid characters', () => {
+        it('should extract valid portion when parameter has invalid characters', () => {
```
rephus: reverted

packages/common/src/utils/formatting.test.ts
```diff
@@ -1566,4 +1566,160 @@ describe('Formatting', () => {
...
+                    formatItemValue(
+                        metricWithConditionalFormat,
+                        1234.56,
+                        false,
+                        parameters,
+                    ),
+                ).toBe('€1,234.56');
+            });
+
+            it('should handle simple currency parameter substitution', () => {
```
rephus, line 1618: bonus feature. without any conditional , just use the parameter on the format

packages/common/src/visualizations/helpers/getCartesianAxisFormatterConfig.ts
```diff
@@ -69,7 +72,7 @@ export const getCartesianAxisFormatterConfig = ({
     if (axisItem && (hasFormattingConfig || axisMinInterval)) {
         axisConfig.axisLabel = {
             formatter: (value: AnyType) =>
-                formatItemValue(axisItem, value, true),
+                formatItemValue(axisItem, value, true, parameters),
```
rephus, line 75: now we need to pass paramteres to all the formatItemValue method and similar, on both the frontend and backend

examples/full-jaffle-shop-demo/dbt/models/events.yml
```diff
@@ -94,6 +95,26 @@ models:
...
+              type: sum 
+              # only shows $ currency
+              format: '${ld.parameters.currency=="usd"?"$":""}0,0${ld.parameters.decimals=="true"?".00":""}'
+            with_decimals_format: 
+              type: sum 
+              format: '0,0${ld.parameters.decimals=="true"?".00":""}'
+            with_mixed_parameter: 
+              type: sum 
+              # Invalid parameter, using the wrong value for decimals
+              format: '${ld.parameters.decimals=="usd"?"$":"-"}0,0.00'
```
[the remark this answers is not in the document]
rephus: This is intentional

examples/full-jaffle-shop-demo/dbt/lightdash.config.yml
```diff
@@ -53,3 +53,26 @@ parameters:
...
+    description: "Conditional formatting for decimals"
+    default: "true"
+    options:
+      - "true"
+      - "false"
+  currency:
+    label: "Currency"
+    description: "Conditional formatting for currency"
+    default: "usd"
+    options:
```
stevelikesmusic, line 67: Decided against a format field for params?

packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
```diff
@@ -1838,6 +1838,15 @@ export class MetricQueryBuilder {
             sqlBuilder: this.args.warehouseSqlBuilder,
         });
 
+        // Also collect parameter references from fields (e.g., format strings)
+        Object.values(fields).forEach((field) => {
+            if ('parameterReferences' in field && field.parameterReferences) {
```
stevelikesmusic, line 1843: Sometimes TypeScript just makes things more difficult...

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -487,8 +488,11 @@ export class ExploreCompiler {
...
+            compiledSql,
+            typeof metric.format === 'string' ? metric.format : undefined,
```
stevelikesmusic, line 494: Is this check in case the type is the deprecated Format type?

packages/common/src/compiler/parameters.ts
```diff
@@ -6,30 +6,59 @@ import { CompileError } from '../types/errors';
 import type { CompiledTable, Table } from '../types/explore';
 import type { LightdashProjectParameter } from '../types/lightdashProjectConfig';
 
+// Regex for SQL parameter replacement - requires full ${} syntax
```
stevelikesmusic, line 9: I appreciate the comment

packages/common/src/compiler/parameters.ts
```diff
@@ -6,30 +6,59 @@ import { CompileError } from '../types/errors';
...
     return [...new Set(parameterNames)];
 };
 
+/**
+ * Extracts and combines parameter references from both SQL and format strings
+ * @param compiledSql - The compiled SQL to extract parameters from
+ * @param format - Optional format string to extract parameters from
+ * @returns An array of unique parameter names from both sources
+ */
+export const getParameterReferencesFromSqlAndFormat = (
```
stevelikesmusic, line 50: Could the name be shortened to something like `formatParameterReferences` or would it make sense to add the `format` option to `getParameterReferences` instead of adding a regex option?

packages/common/src/compiler/parameters.ts
```diff
@@ -6,30 +6,59 @@ import { CompileError } from '../types/errors';
...
+    compiledSql: string,
+    format?: string,
+): string[] => {
+    const sqlParameterReferences = getParameterReferences(compiledSql);
+
+    const formatParameterReferences =
+        format && typeof format === 'string'
```
stevelikesmusic, line 57: Why check that format is type string? We check when calling and the TS type for the signature is string.

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Utilities for evaluating conditional expressions in format strings with parameters
+ */
+
+/**
+ * Strips the ld.parameters. prefix from a parameter name if present
+ * Examples:
+ *   - "ld.parameters.currency" -> "currency"
+ *   - "currency" -> "currency"
+ */
+function stripParameterPrefix(paramName: string): string {
```
stevelikesmusic, line 13: Would it be worth having unit tests for these utils?

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+    const ldPrefix = 'ld.parameters.';
+    const lightdashPrefix = 'lightdash.parameters.';
```
stevelikesmusic: Should these be separate constants in their own module constants? At least the `ld|lightdash.parameters` parts

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+    value: string,
+    parameters: Record<string, unknown>,
+): unknown {
+    // Check if it's a quoted string literal
+    if (
+        (value.startsWith('"') && value.endsWith('"')) ||
+        (value.startsWith("'") && value.endsWith("'"))
+    ) {
+        return value.slice(1, -1);
+    }
```
stevelikesmusic, line 54: Why not use the function above instead of re-implementing the logic?

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+    return value;
+}
+
+/**
+ * Gets the value from either a parameter or a literal (quoted string)
+ */
+function getConditionValue(
+    value: string,
+    parameters: Record<string, unknown>,
+): unknown {
```
stevelikesmusic, line 47: Why wouldn't the return type be `string | undefined`?

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+    const [conditionPart, valuesPart] = expression
+        .split('?')
+        .map((s: string) => s.trim());
+
+    if (!valuesPart) {
+        return expression; // Invalid ternary, return as-is
+    }
+
+    // Split values by : to get trueValue and falseValue
+    const colonIndex = valuesPart.indexOf(':');
```
stevelikesmusic: Why not use the same .split method we've been using with the `:`?

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    // Otherwise, treat as parameter name - strip ld.parameters. prefix if present
+    const paramName = stripParameterPrefix(value);
+    return parameters[paramName];
+}
+
+/**
+ * Evaluates a condition like: paramName=="value" or paramName!="value"
+ */
+function evaluateCondition(
```
stevelikesmusic, line 64: Do we also support greater/less than comparisons?

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    // Check for == operator
+    if (condition.includes('==')) {
+        const [left, right] = condition
+            .split('==')
+            .map((s: string) => s.trim());
+        const leftValue = getConditionValue(left, parameters);
+        const rightValue = getConditionValue(right, parameters);
+        return leftValue === rightValue;
```
stevelikesmusic, line 85: It seems like abstracting a `parseCondition` might be helpful?

packages/common/src/utils/formatting.ts
```diff
@@ -524,6 +529,27 @@ export function applyCustomFormat(
...
+
+    if (!hasPlaceholders) {
+        return isValidFormat(formatString);
+    }
+
+    // Strip out ${...} placeholders and validate what remains
+    // This handles formats like: '${ld.parameters.currency=="usd"?"$":""}0,0.00'
+    // After stripping: '0,0.00' which is valid
+    // Use bounded quantifier to prevent ReDoS - max 500 chars per placeholder
+    const withoutPlaceholders = formatString.replace(/\$\{[^}]{0,500}\}/g, '');
```
stevelikesmusic: Is this the same regex from the utils?

packages/common/src/utils/formatting.ts
```diff
@@ -717,12 +747,38 @@ export function formatItemValue(
...
     if (item) {
         if (hasValidFormatExpression(item)) {
-            return formatValueWithExpression(item.format, value);
+            // Check if format uses parameter placeholders (supports both ${ld.parameters and ${lightdash.parameters)
+            const hasParameterPlaceholders =
+                item.format.includes('${ld.parameters') ||
+                item.format.includes('${lightdash.parameters');
+
+            // Only apply format expression if parameters are provided when needed
+            if (!hasParameterPlaceholders || parameters) {
```
stevelikesmusic: I don't understand this one. We evaluate parameters if there aren't parameter placeholders?

packages/common/src/utils/formatting.test.ts
```diff
@@ -1566,4 +1566,195 @@ describe('Formatting', () => {
             expect(applyDefaultFormat(-1234)).toBe('-1,234');
         });
     });
+
+    describe('formatItemValue with parameters', () => {
```
stevelikesmusic, line 1580: Should we add tests for potentially malicious inputs (ie. really long strings, parameters without closing braces, etc.) to test things like the ReDoS protection?

stevelikesmusic (commented):
So much good stuff 👏 👏 👏

packages/common/src/compiler/parameters.ts
rephus: This method is a wrapper for getting parameter both from sql and format,which use different regex. This method is only used in 2 places, but I think the abstraction is useful. I don't mind renaming it though.

packages/common/src/utils/conditionalFormatExpressions.ts
rephus: Yes !

packages/common/src/utils/conditionalFormatExpressions.ts
rephus: good idea

packages/common/src/compiler/parameters.ts
rephus: format can be multiple types, `string | Format`

packages/common/src/utils/conditionalFormatExpressions.ts
rephus: that remove quotes from the values, not the prefix (eg: ${ld.parameters.currency=="usd"?"$":""}) 

So `"usd"` -> `usd`

packages/common/src/utils/conditionalFormatExpressions.ts
rephus: parameters: Record<string, unknown>,
 If we return `parameters[paramName]` the result can be unknown

packages/common/src/utils/conditionalFormatExpressions.ts
rephus: That is also an alternative valid implementation. 
The main reason why it is used ":" here is because first we want to know if ":" and see if it is a valid ternary expression. (it is easier than checking lenght of the split)

packages/common/src/utils/conditionalFormatExpressions.ts
rephus: Not yet. @magnew could you confirm if parameter can take numbers ? or just strings. 

This feature would not be hard to add in the future, we just need to update some methods from this file.

packages/common/src/compiler/parameters.ts
stevelikesmusic: I get that metric.format has `string | Format`, but the signature for this function only accepts `string` and we make a typecheck before calling it. 

Can we either add `Format` to the signature types or remove the typecheck? It makes the reader stop and question what's going on when in the function `format` is typed as `string` and then we check the type again.

packages/common/src/utils/formatting.ts
rephus: No, this is supposed to be the oppsite, so we remove the parameter sytanx from the format, so for 

`${ld.parameters.currency=="usd"?"$":""}0,0.00'` -> we'll get '0,0.00'

packages/common/src/utils/conditionalFormatExpressions.ts
stevelikesmusic: I see the difference now:
- One removes quotes if they're there
- The other removes quotes if they're there, otherwise does the parameter substitution 

Would it be worth pulling out an `isQuotedString` function?

packages/common/src/utils/formatting.ts
stevelikesmusic: I love|hate regex

packages/common/src/utils/conditionalFormatExpressions.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // Split values by : to get trueValue and falseValue
+    const colonIndex = valuesPart.indexOf(':');
+    if (colonIndex === -1) {
+        return expression; // Invalid ternary, return as-is
+    }
+
+    const trueValue = valuesPart.substring(0, colonIndex).trim();
+    const falseValue = valuesPart.substring(colonIndex + 1).trim();
```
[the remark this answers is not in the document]
stevelikesmusic: @rephus do we allow passing in time values this way?

packages/common/src/utils/formatting.ts
```diff
@@ -717,12 +747,37 @@ export function formatItemValue(
...
     if (item) {
         if (hasValidFormatExpression(item)) {
-            return formatValueWithExpression(item.format, value);
+            // Check if format uses parameter placeholders
+            const hasParameterPlaceholders =
+                item.format.includes('${ld.parameters');
+
+            // Only apply format expression if parameters are provided when needed
+            if (!hasParameterPlaceholders || parameters) {
+                // Evaluate conditional expressions in format string if parameters are provided
```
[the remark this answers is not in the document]
stevelikesmusic: This seems valid

packages/common/src/utils/conditionalFormatExpressions.ts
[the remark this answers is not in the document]
rephus: good question, let me check

rephus:
Errors in tests seem unrelated. I manually tested existing formatting on preview

Screenshot from 2025-11-03 12-03-04

Issue #17707, opened 2025-10-28
Jinja templating in dbt_project.yml fails can for missing dbt context variables


almeidabbm:
### Description

Related PR: https://github.com/lightdash/lightdash/pull/17167

dbt allows Jinja templating within dbt_project.yml.
This means users can define configuration values dynamically using variables that exist only within the dbt runtime context — for example:
```yaml
models:
  +schema: "{{ target.name }}"
```
Currently, when Lightdash attempts to render `Jinja` templates in `dbt_project.yml`, these context-specific variables are not available.
As a result, they render as empty strings, which can lead to invalid YAML syntax or misconfigured projects when we attempt to parse the file post-render.

<img width="1222" height="294" alt="Image" src="https://github.com/user-attachments/assets/2113702c-cc35-4ce5-9570-253e8b5de006" />

### Steps to Reproduce the Bug or Issue

1. Copy the example project yaml into `dbt_project.yml` in `full-jaffle-shop-demo`
2. Try to run a `cli` command, e.g. `lightdash start-preview`

Example `dbt_project.yml`

```yaml
name: jaffle_shop
version: "0.1"
profile: jaffle_shop
config-version: 2
model-paths:
  - models
analysis-paths:
  - analysis
test-paths:
  - tests
seed-paths:
  - data
macro-paths:
  - macros
target-path: target
clean-targets:
  - target
  - dbt_modules
  - logs
models:
  jaffle_shop:
    +schema: |
      {%- if target.name in ["prod"] -%}snapshots
      {%- else -%}{{ target.schema }}_snapshots
      {%- endif -%}
    materialized: table
    staging:
      materialized: view
    fanouts_examples:
      +meta:
        group_label: fanouts
    campaigns:
      +meta:
        group_label: campaigns
    marketing_touchpoints:
      +meta:
        group_label: campaigns
    patient_health_scores:
      +meta:
        group_label: healthcare
    visit_analytics:
      +meta:
        group_label: healthcare
seeds:
  jaffle_shop:
    raw_customers:
      +column_types:
        created: timestamp
    raw_orders:
      +column_types:
        order_date: date
        estimated_delivery_days: integer
        shipping_cost: numeric
        tax_rate: numeric
    raw_plan:
      +column_types:
        metadata: jsonb
    raw_tracks:
      +column_types:
        timestamp: timestamp 
```

### Cloud or self-hosting

None

PR #17167, opened 2025-10-01
fix: process env_var in project yml

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/17165

> target-path: "{{ env_var('DBT_ENV__TARGET_FOLDER', 'target') }}"

> export DBT_ENV__TARGET_FOLDER='new_target'

image

PR #17708, opened 2025-10-28
fix: jinja replacement conflicts with missing jinja context

almeidabbm:
Closes: #17707

### Description:
Improved Jinja templating handling in dbt context by rendering individual config keys instead of the entire file. This approach provides better error isolation, allowing the system to continue processing even if some Jinja expressions fail to render. The code now separately renders `target-path`, `models-path`, `name`, and `profile` keys, maintaining the original fallback behavior when rendering fails.

**Steps to reproduce:** https://github.com/lightdash/lightdash/issues/17707

### DIFF
--- a/packages/cli/src/dbt/context.ts
+++ b/packages/cli/src/dbt/context.ts
@@ -16,6 +16,22 @@ export type DbtContext = {
     modelsDir: string;
 };
 
+function tryRenderVariable(variableName: string, raw: string): string {
+    try {
+        return renderTemplatedYml(raw);
+    } catch (e) {
+        GlobalState.debug(
+            `> Warning: Failed to render Jinja in dbt_project.yml ${variableName}: ${getErrorMessage(
+                e,
+            )}`,
+        );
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#17708** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

almeidabbm:
> Code looks great. IT seems this change is also compatible with the tests implemented in this other PR https://github.com/lightdash/lightdash/pull/17167/files should we add some tests to ensure we don't render yml for other parts ?

This function just returns a subset of the `dbt_context.yml` file so it's not possible to verify that other fields don't get their Jinja rendered

PR #17560, opened 2025-10-21
feat(redux): finish redux migration

almeidabbm:
# Migrate Explorer Context to Redux

### Description:
This PR migrates the Explorer Context to Redux, moving chartConfig and pivotConfig state management from Context to Redux. Key changes include:

- Replaced `useExplorerContext` with Redux selectors and actions in multiple components
- Added new Redux selectors like `selectChartConfig`, `selectPivotConfig`, and `selectSavedChart`
- Added new Redux actions for chart configuration management
- Implemented `selectHasUnsavedChanges` selector to replace the Context comparison function
- Enhanced `useExplorerInitialization` hook to handle more initialization scenarios
- Removed the ExplorerProvider component and related Context implementation
- Updated components to use Redux for state management instead of Context

This change improves performance by reducing unnecessary re-renders and simplifies the codebase by consolidating state management in Redux.

test-perf

PR #17710, opened 2025-10-28
fix: store unsaved configs by type

magnew:
### Description:

Store chart config changes by type so that they aren't lost when switching charts. 

To see the issue this is solving:
- Create a bar chart
- Change a config (color for example)
- Switch to big number
- Switch back
- Your config will be lost

### DIFF
--- a/packages/frontend/src/features/explorer/store/explorerSlice.ts
+++ b/packages/frontend/src/features/explorer/store/explorerSlice.ts
@@ -38,11 +38,28 @@ import {
     type ExplorerReduceState,
     type ExplorerSection,
 } from '../../../providers/Explorer/types';
-import { getValidChartConfig } from '../../../providers/Explorer/utils';
+import {
+    getCachedPivotConfig,
+    getValidChartConfig,
+} from '../../../providers/Explorer/utils';
 
 import { calcColumnOrder } from './utils';
 
 export type ExplorerSliceState = ExplorerReduceState;
+
+// Type-safe helper to save config to cache
+// Ensures that chartType and config match up correctly
+function saveConfigToCache<T extends ChartType>(
+    cache: Partial<ConfigCacheMap>,
+    chartType: T,
+    config: ConfigCacheMap[T]['chartConfig'],
+    pivotConfig?: { columns: string[] },
+): void {
+    cache[chartType] = {
+        chartConfig: config,
+        pivotConfig,
+    } as ConfigCacheMap[T];
+}
 
 const initialState: ExplorerSliceState = defaultState;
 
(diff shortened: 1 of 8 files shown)


magnew:
* **#17710** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17710?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17756** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17756?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17560** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/features/explorer/store/explorerSlice.ts
```diff
@@ -373,32 +374,38 @@ const explorerSlice = createSlice({
...
-                cachedConfigs: Partial<ConfigCacheMap>;
-            }>,
+            action: PayloadAction<{ chartType: ChartType }>,
         ) => {
+            const before = state.unsavedChartVersion.chartConfig;
+
+            // save the current config to the cache
+            state.cachedChartConfigs[before.type] = current(
+                before.config,
+            ) as any;
```
almeidabbm: Would using some sort of type narrowing function like

```typescript
function saveConfigToCache<T extends ChartType>(
    cache: Partial<ConfigCacheMap>,
    chartType: T,
    config: ConfigCacheMap[T],
) {
    cache[chartType] = config;
}
```

help here?

almeidabbm (changes requested):
I am seeing a strange behavior when trying to go back to a cached config. It is not keeping the same group by

https://github.com/user-attachments/assets/86534933-13bb-40a6-a894-dd7643d8de4e

packages/frontend/src/features/explorer/store/explorerSlice.ts
magnew: Good call

almeidabbm (approved):
WORKS SUPER!! 🚀 
Minor issue is that tables now are not taking into consideration the pivot config in the case that there's no pivot config cached (see video) but this might be intended behavior or even fixed later, non blocking feedback

https://github.com/user-attachments/assets/c92998d9-f643-49e9-989f-9651c812e996

magnew:
> WORKS SUPER!! 🚀 Minor issue is that tables now are not taking into consideration the pivot config in the case that there's no pivot config cached (see video) but this might be intended behavior or even fixed later, non blocking feedback
> 
>  Screen.Recording.2025-10-31.at.16.05.32.mov

Yea, intentional. I believe this is how it used to work.

PR #17756, opened 2025-10-30
fix: keep correct pivot dimension when switching charts

magnew:
### Description:

[redux branch] When switching between charts, we were resetting the pivot dimension to the default. This makes it so we keep the pivot config when switching between chart types. 

To test:
- create a bar chart with 2 pivot dimension options
- choose the non-default option
- switch to big number
- switch back to bar chart
- the pivot dimension should now be maintained

Note that one upshot from this change is that switching to and from a table will maintain the pivot across types. This is actually kind of nice, but we can change the behavior back to what we had before when we cache each chart type separately.

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
@@ -44,7 +44,6 @@ const VisualizationCardOptions: FC = memo(() => {
         setStacking,
         isLoading,
         resultsData,
-        setPivotDimensions,
         pivotDimensions,
     } = useVisualizationContext();
     const disabled = isLoading || !resultsData || resultsData.rows.length <= 0;
@@ -336,7 +335,6 @@ const VisualizationCardOptions: FC = memo(() => {
                     }
                     icon={<MantineIcon icon={IconChartPie} />}
                     onClick={() => {
-                        setPivotDimensions(undefined);
                         setStacking(undefined);
                         setCartesianType(undefined);
                         setChartType(ChartType.PIE);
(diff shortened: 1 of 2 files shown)


magnew:
* **#17710** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17756** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17756?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17756?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17560** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17560?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

magnew:
> Works super!! 🚀 I do have a question tho, if you save a chart that was initially configured to have a group by but then doesn't need it (e.g. first save cartesian and then change config to big number) the saved chart configuration will always keep the `pivotConfiguration`, wondering if this will have any side effects 🤔
> 
>  Screen.Recording.2025-10-30.at.14.47.01.mov 
> I actually believe this to be fine since we always call `derivePivotConfigurationFromChart` which returns `undefined` for anything other than `table` or `cartesian` so those will be ignored. Just wanted to raise the question

That's a good callout, but it should be fine. I'm kind of planning to have the pivot config saved per chart soon, so maybe it will be a non-issue.

almeidabbm:
Failing test is unrelated 😄

PR #17779, opened 2025-10-31
fix: clicking on metrics becomes slow because of expensive useColumns recalculation

almeidabbm:
### Description:
Optimized the `useColumns` hook by splitting the item map calculation into two layers:

1. A base layer that contains fields that rarely change
2. An override layer for metric overrides that change frequently

This prevents full recalculation when only metric overrides change. Additionally:

- Improved the filtering of active fields by using a direct loop instead of Array.reduce
- Added early return when there are no metric overrides
- Simplified the logic for applying overrides to only items that have them

These changes should reduce unnecessary re-renders and improve performance when working with tables that have many columns or frequent metric override changes.

**Before**

[Screen Recording 2025-10-31 at 12.16.01.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/user-attachments/video/0e33adbb-72aa-4816-976e-44a5b0da31bf.mov)

**After**

[Screen Recording 2025-10-31 at 12.14.52.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/user-attachments/video/7590c521-60d2-4042-916f-2f6c470516a3.mov)

### DIFF
--- a/packages/frontend/src/hooks/useColumns.tsx
+++ b/packages/frontend/src/hooks/useColumns.tsx
@@ -180,29 +180,50 @@ export const useColumns = (): TableColumn[] => {
 
     const hasNoActiveFields = activeFields.size === 0;
 
-    const itemsMap = useMemo<ItemsMap | undefined>(() => {
+    // Split itemsMap into base map (rarely changes) and override layer (frequently changes)
+    // This prevents full recalculation when only metricOverrides change
+    const baseItemsMap = useMemo<ItemsMap | undefined>(() => {
         if (!exploreData || hasNoActiveFields) return;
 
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#17779** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17779?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17779?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pimmey (approved):

PR #17782, opened 2025-10-31
chore: optimize table calc form initial load

magnew:
### Description:

The problem: table calculation modal is slow to open and slow to save.

All this does is some of the obvious react perf best practices:
- Use memos
- Use callbacks
- Lazy load SQL form

The good news is that this helps a fair amount -- I see a consistent **20-30%** improvement in opening the dialog. So this is worth doing. 

This less good news is that there is more to do. Will keep digging.

### DIFF
--- a/packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
+++ b/packages/frontend/src/features/tableCalculation/components/SqlForm.tsx
@@ -6,7 +6,7 @@ import {
     useMantineTheme,
 } from '@mantine/core';
 import { IconSparkles } from '@tabler/icons-react';
-import { type FC } from 'react';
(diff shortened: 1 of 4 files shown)


magnew:
* **#17782** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17782?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pimmey (approved):
Nice, can see it work faster as well 💪

PR #17831, opened 2025-11-04
fix: key viz provider off of chart ID

magnew:
### Description:

Re-mount the visualisation provider whenever the chart ID changes. This avoids caching any chart state in the viz provider.

PR #17832, opened 2025-11-04
fix: recreate store for each explore instance

magnew:
Closes: GLITCH-63

### Description:

Recreates the explore store for each instance of the explore. This will prevent state from bleeding between explores as we have seen when editing multiple charts from a dashboard. 

It seems like this pattern is generally deemed safe IF you are only using the store on a single page in your app, which is currently the case for us. We don't need any of the explore page state outside of the explore page. And we do want to reset it per instance.

### DIFF
--- a/packages/frontend/src/features/explorer/store/index.ts
+++ b/packages/frontend/src/features/explorer/store/index.ts
@@ -1,11 +1,17 @@
 import { configureStore } from '@reduxjs/toolkit';
 import { explorerReducer } from './explorerSlice';
 
-export const explorerStore = configureStore({
-    reducer: {
-        explorer: explorerReducer,
-    },
-});
+// Factory function to create a new store instance
+// This prevents state leaking between different explorer instances
+export const createExplorerStore = () =>
+    configureStore({
+        reducer: {
+            explorer: explorerReducer,
+        },
+    });
+
+// Keep singleton for backward compatibility (Explorer.tsx uses it)
+export const explorerStore = createExplorerStore();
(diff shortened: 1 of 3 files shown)


magnew:
* **#17832** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17832?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17832?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17831** <a href="https://app.graphite.dev/github/pr/lightdash/lightdash/17831?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/features/explorer/store/index.ts
```diff
@@ -1,11 +1,17 @@
...
-    },
-});
+// Factory function to create a new store instance
+// This prevents state leaking between different explorer instances
+export const createExplorerStore = () =>
+    configureStore({
+        reducer: {
+            explorer: explorerReducer,
+        },
+    });
```
almeidabbm, line 11: This basically means that we can also initialize the store with a `preloadedState` instead of relying on the useEffects?

magnew: Yes. I have that mostly working, but it has it's own challenges. Ill put a PR on top of this stack

PR #17858, opened 2025-11-04
chore: populate explore store explicitly

magnew:
### Description:

Populate the explore store more explicitly. Instead of using using a hook with 2 useEffects. 

Should help with perf and stability. This seems like an improvement in clarity to me, but I am open to other opinions. 

I tested this a lot, but it's a little risky. Any more eyes on "Does this work right?" would be good. 

Things to try:
- Creating charts
- Editing
- Explore from here
- Share links
- Embed
- History

All of which I have tried various ways.

### DIFF
--- a/packages/frontend/src/pages/SavedExplorer.tsx
+++ b/packages/frontend/src/pages/SavedExplorer.tsx
@@ -1,4 +1,4 @@
-import { lazy, memo, Suspense, useEffect, useMemo } from 'react';
+import { lazy, memo, Suspense, useEffect, useState } from 'react';
 import { Provider } from 'react-redux';
 import { useParams } from 'react-router';
 import ErrorState from '../components/common/ErrorState';
@@ -11,7 +11,6 @@ import {
     buildInitialExplorerState,
     createExplorerStore,
     explorerActions,
-    useExplorerDispatch,
 } from '../features/explorer/store';
 import useDashboardStorage from '../hooks/dashboard/useDashboardStorage';
 import { useExplorerQueryEffects } from '../hooks/useExplorerQueryEffects';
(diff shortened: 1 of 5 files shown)


magnew:
* **#17858** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17858?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/features/explorer/store/useExplorerInitialization.ts
```diff
@@ -1,159 +0,0 @@
...
-        expandedSections,
-        customInitialState,
-        defaultLimit,
-        dispatch,
-    ]);
-
-    // Set savedChart in Redux (separate from initialState for reactivity)
-    // This handles the case where data loads after component mounts
-    useEffect(() => {
-        dispatch(explorerActions.setSavedChart(savedChart));
```
pimmey, line 155: We used to update `savedChart` here as well, will that reside someplace (more appropriate) else? 😄

packages/frontend/src/pages/SavedExplorer.tsx
```diff
@@ -101,10 +105,7 @@ const SavedExplorer = () => {
 
     return (
         <Provider store={store} key={`saved-${savedQueryUuid}-${mode}`}>
```
pimmey: Maybe we can only rely on `saved-${savedQueryUuid}` here, cause the `useEffect` above has `isEditMode`. Would probably prevent some unnecessary unmount? 🤔

pimmey (approved):
LGTM, tested it for a while -- works the same, but architecturally it's much nicer 💅 
Just left those two comments, but pre-approving ✅

packages/frontend/src/pages/SavedExplorer.tsx
magnew: Good call.

packages/frontend/src/features/explorer/store/useExplorerInitialization.ts
magnew: It's still happening in a useEffect :(. In the SavedExplorer. 

Not sure we can totally avoid useeffects unless we use thunks or RTK Query. But I figured it was still better than having both of these useEffects in a hook. I think.

PR #17863, opened 2025-11-05
ci: Proposal to optimize the Dockerfile used for preview environments

rberrelleza:
Closes: 

### Description:
This is a proposal to potentially improve the time it takes to build a preview environment for the typical Pull Request. This is not fully tested yet, but wanted to send it early in order to get feedback. 

The main changes are:
1. create dedicated `.dockerignore` and `.oktetoignore` files to reduce the files sent from the github worker to the Okteto cluster. 
2. Only install pnpm once and create a 'base' layer
3. Create different stages for the `common`, `warehouses`, `backend` and `frontend` projects to reduce cache misses and increase potential parallelization.
4. Only copy the `dist` folders to the final image, instead of dist + source, to reduce image size and total number of files. 

Looking forward to your feedback

### DIFF
--- a/dockerfile-prs
+++ b/dockerfile-prs
@@ -159,7 +159,6 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
 FROM prod-builder AS build-common
 COPY packages/common/tsconfig*.json packages/common/
 COPY packages/common/src/ packages/common/src/
-#RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --offline
 RUN pnpm -F @lightdash/common build
 # output: /usr/app/packages/common/dist
 


.dockerfile.dockerfile-prs
```diff
@@ -0,0 +1,28 @@
+# .dockerfile.dockerfile-prs
```
stevelikesmusic, line 1: @rberrelleza Is this file used via an okteto convention? I don't see it referenced anywhere else in the PR. Should this be a `.dockerignore` like `dockerfile-prs.dockerignore`?

rberrelleza: yes, this is meant to be a .dockerignore for the dockerfile-prs file.  it should be named `dockerfile-prs.dockerignore`, I put the wrong filename.

PR #17851, opened 2025-11-04
refactor: make getUserProjectRoles public and use UUID instead of ID

ZeRego:
Closes: 

### Description:
Made `getUserProjectRoles` method public and updated it to use `user_uuid` instead of `user_id`. The method now joins with the users table to look up project memberships by UUID rather than ID, which is more consistent with our UUID-based approach elsewhere in the codebase.

PR #17872, opened 2025-11-05
feat: include project roles in SCIM user responses

ZeRego:
Closes: #XXXX

### Description:
Enhances SCIM user endpoints to include project roles in user responses. This change ensures that when SCIM users are retrieved, their project roles are properly included alongside organization roles, providing a complete view of user permissions across the platform.

The implementation adds a new `getUserScimRoles` method that combines organization roles with project roles from the user model, and updates all SCIM user endpoints to fetch and include these roles in responses.

Screenshot 2025-11-05 at 09 42 01
Screenshot 2025-11-05 at 09 40 57

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.ts
@@ -1528,7 +1528,7 @@ export class ScimService extends BaseService {
 
     private async getUserScimRoles(
         user: Pick<LightdashUser, 'userUuid' | 'role'>,
-        allScimRoles: ScimRole[],
+        availableScimRoles: ScimRole[],
     ): Promise<ScimUserRole[]> {
         try {
             const allRoles: ScimUserRole[] = [];
@@ -1556,7 +1556,7 @@ export class ScimService extends BaseService {
             );
 
             // Filter SCIM roles to only include those the user has and convert to ScimUserRole
-            const projectScimRoles = allScimRoles
+            const projectScimRoles = availableScimRoles
                 .filter((scimRole) => userScimRoleIds.includes(scimRole.id))
                 .map((scimRole) => ({
                     value: scimRole.value,


ZeRego:
* **#17925** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17925?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17914** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17914?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17899** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17895** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17895?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17894** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17879** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17879?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17872** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17872?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17872?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17851** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17851?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -1485,6 +1526,58 @@ export class ScimService extends BaseService {
...
+            const userScimRoleIds = userProjectRoles.map((role) =>
+                ScimService.generateRoleId({
+                    roleUuid: role.roleUuid || role.role, // Check first for custom role uuid and then system role name
+                    projectUuid: role?.projectUuid,
+                }),
+            );
+
+            // Filter SCIM roles to only include those the user has and convert to ScimUserRole
+            const projectScimRoles = allScimRoles
+                .filter((scimRole) => userScimRoleIds.includes(scimRole.id))
```
stevelikesmusic, line 1560: If a user has an org role, do they automatically get that same org role for all projects? Are we filtering out the possibility that a user has only explicit project roles and no inherited roles from their org assignment?

At most, we would have only assigned a single role above, so do we need the filter/map chain? I don't understand why we push a value into `allScimRoles` only to map it to a new object later.

ZeRego: > If a user has an org role, do they automatically get that same org role for all projects? 

Internally, yes. SCIM only exposes explicit roles and not inherited ones. 

> Are we filtering out the possibility that a user has only explicit project roles and no inherited roles from their org assignment?

In theory, the user will have an org role because it is required for a org membership. If it hasn't we will throw an error later because we cant connect the user to the org.
Should I add an explicit error here?

> At most, we would have only assigned a single role above, so do we need the filter/map chain? I don't understand why we push a value into allScimRoles only to map it to a new object later.

I don't follow. Can you explain again with an example?

stevelikesmusic: I take it back. I got confused with the names `allRoles` and `allScimRoles`

ZeRego: renaming consts

Issue #17739, opened 2025-10-29
I want 100% stacked bar chart value labels to show up as percent with a single decimal


jpetey75:
I want 100% stacked bar chart value labels to show up as percent with a single decimal.

Right now when you enable value labels the chart shows the percent value, but no % symbol, and it includes three decimals: `45.287`, `3.915`, `12.028`, etc. 

It would look better rounded to a single decimal and a % on the end: `45.3%`, `3.9%`, `12.0%`, etc.

torimwhaley:
I think I found a bug with %'s in a stacked bar chart. When I stack a table calculation which is a percentage, stacking works perfectly, except when stacking 100%, in which case the % labels shown in the bar are inflated. 

Here is a link to the [SaaS Demo chart in question](https://analytics.lightdash.cloud/projects/2cc7b1db-c5b5-40c2-b35b-9dfc88d5494f/saved/3262c8a9-b52a-4002-9cb1-1c50bfa30586/edit?fromDashboard=554fb84c-13e4-4500-b137-4164188733ee).

<img width="1913" height="884" alt="Image" src="https://github.com/user-attachments/assets/b2ccebb1-f553-467d-a4a1-ecef6a1546d1" />
<img width="1911" height="913" alt="Image" src="https://github.com/user-attachments/assets/11d9e09a-7eea-4719-9b56-55ea3f174395" />

PR #17882, opened 2025-11-05
fix: format values as percentages in 100% stacked charts

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-132/fix-stacked-bar-chart-value-labels
Closes: https://linear.app/lightdash/issue/CENG-131/fix-100percent-bar-chart-issues
Closes: https://github.com/lightdash/lightdash/issues/17739

Before

Screenshot from 2025-11-05 12-44-44

Numbers were not formatted (https://github.com/lightdash/lightdash/issues/17739) : 

image

SQL runner 

image

After
Screenshot from 2025-11-05 12-46-28

SQL runner

image

### Added new e2e tests

image

### Description:
Improves formatting for 100% stacked charts by displaying values with a percentage sign and proper decimal rounding. This ensures that values in 100% stacked charts are consistently displayed as percentages, making the visualization more intuitive for users.

The implementation adds special formatting logic for stack100 charts in both the CartesianChartDataModel and the ECharts configuration hooks, ensuring that percentage values are properly formatted across all chart components.

### DIFF
--- a/packages/common/src/visualizations/CartesianChartDataModel.ts
+++ b/packages/common/src/visualizations/CartesianChartDataModel.ts
@@ -16,7 +16,7 @@ import {
     ECHARTS_DEFAULT_COLORS,
 } from '../types/savedCharts';
 import { type SqlRunnerQuery } from '../types/sqlRunner';
-import { applyCustomFormat } from '../utils/formatting';
+import { applyCustomFormat, formatNumberValue } from '../utils/formatting';
 import {
     getAxisLabelStyle,
     getAxisLineStyle,
@@ -119,7 +119,25 @@ export class CartesianChartDataModel {
 
     // Get the formatter for the value label,
     // which has more complex inputs
-    static getValueFormatter(format: Format | undefined) {
+    static getValueFormatter(
+        format: Format | undefined,
+        isStack100: boolean = false,
+    ) {
+        // For 100% stacked charts, always format as percentage
+        // because all values are already converted to 0-100 range
+        if (isStack100) {
+            return (params: AnyType) => {
+                const value =
+                    params.value[params.dimensionNames[params.encode.y[0]]];
+                return typeof value === 'number'
+                    ? `${formatNumberValue(value, {
+                          type: CustomFormatType.NUMBER,
+                          round: 1,
+                      })}%`
+                    : `${value}%`;
+            };
+        }
+
         if (format === Format.PERCENT) {
             // Echarts doesn't export the types for this function
             return (params: AnyType) => {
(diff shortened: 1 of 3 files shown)


rephus:
* **#17882** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17882?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17882?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -625,6 +628,19 @@ const seriesValueFormatter = (
     }
 };
 
+/**
+ * Format value for 100% stacked charts
+ * For stack100, values are already percentages (0-100) so we just append %
+ */
+const formatStack100Value = (value: unknown): string => {
```
rephus, line 635: These changes are for explore charts

packages/common/src/visualizations/CartesianChartDataModel.ts
```diff
@@ -119,7 +119,25 @@ export class CartesianChartDataModel {
 
     // Get the formatter for the value label,
     // which has more complex inputs
-    static getValueFormatter(format: Format | undefined) {
+    static getValueFormatter(
```
rephus, line 122: these changes are for sql runner charts

stevelikesmusic: Would it make sense to call this something like `usePercents` rather than `isStack100`? If ever we wanted to use this for something similar.

packages/e2e/cypress/e2e/app/pivotTables.cy.ts
```diff
@@ -83,3 +83,24 @@ describe('Pivot Tables', () => {
         cy.contains('$1.00');
     });
 });
+
+describe.only('100% stacked bar chart', () => {
```
stevelikesmusic: I feel seen with this `.only`.

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -675,6 +691,7 @@ const getPivotSeries = ({
     cartesianChart,
     pivotValuesColumnsMap,
     parameters,
+    isStack100,
```
stevelikesmusic, line 694: Same comment as above about not being specific to stack100

stevelikesmusic (commented):
Only blocker is remove the .only from the test

packages/e2e/cypress/e2e/app/pivotTables.cy.ts
rephus: thanks, I always forget to remove that

packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
rephus: mm, I'm not sure, stack100 has its specific logic on different places. and even a custom format on the tooltips. Making this generic would be confusing IMO, like, if we rename this `usePercents`, is this logic for stack100? for table calculation with percent format ? for the new metric types ? do they all share he same logic ? 

If all are going to share the same logic , I'd rather to do something like 

```
if (isStack100 || isTableCalculationPercent) formatPercent(value) 
``` 

What do you think ? I'm not 100% sure about this

stevelikesmusic: That works!

stevelikesmusic (approved):

rephus:
Unrelated e2e error 
image

Issue #17266, opened 2025-10-07
I want "Filter dashboard to..." filter to update my default dashboard filter rather than adding a new filter


jpetey75:
Right now when I use the `Filter dashboard to...` option after clicking on a chart segment, the filter gets added to the end of my default dashboard filters. Even if I have a default filter that uses the same field as the filter I just added:

<img width="2412" height="366" alt="Image" src="https://github.com/user-attachments/assets/20c78531-b7a9-4e68-af92-08b76e66056d" />

Ideally it would just input the value I chose into that default filter.

lakshz:
I can solve this. Please assign.

PR #17887, opened 2025-11-05
feat: dashboard filter update should update the existing filter instead of adding new

lakshz:
Closes: #17266 

### Description:

https://github.com/user-attachments/assets/b8fa12bb-c046-46c9-8d22-80882ac1c104

### DIFF
--- a/packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
+++ b/packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
@@ -851,7 +851,7 @@ const DashboardProvider: React.FC<
         );
     }, [dashboardTiles]);
 
-    const addDimensionDashboardFilter = useCallback(
+    const upsertDimensionDashboardFilter = useCallback(
         (filter: DashboardFilterRule, isTemporary: boolean) => {
             const existingSavedFilter = dashboardFilters.dimensions.find(
                 (d) => d.target.fieldId === filter.target.fieldId,
@@ -907,7 +907,7 @@ const DashboardProvider: React.FC<
                     }));
                 }
             } else {
-                setDashboardTemporaryFilters((prev) => ({
+                setDashboardFilters((prev) => ({
                     ...prev,
                     dimensions: [...prev.dimensions, filter],
                 }));
@@ -1159,7 +1159,7 @@ const DashboardProvider: React.FC<
         setDashboardTemporaryFilters,
         dashboardFilters,
         dashboardTemporaryFilters,
-        addDimensionDashboardFilter,
+        upsertDimensionDashboardFilter,
         updateDimensionDashboardFilter,
         removeDimensionDashboardFilter,
         addMetricDashboardFilter,
--- a/packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
+++ b/packages/frontend/src/components/DashboardFilter/FilterDashboardTo.tsx
@@ -17,13 +17,13 @@ type Props = {
 };
 
 export const FilterDashboardTo: FC<Props> = ({ filters, onAddFilter }) => {
-    const addDimensionDashboardFilter = useDashboardContext(
-        (c) => c.addDimensionDashboardFilter,
+    const upsertDimensionDashboardFilter = useDashboardContext(
+        (c) => c.upsertDimensionDashboardFilter,
     );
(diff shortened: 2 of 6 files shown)


packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
```diff
@@ -856,11 +856,30 @@ const DashboardProvider: React.FC<
             const setFunction = isTemporary
                 ? setDashboardTemporaryFilters
                 : setDashboardFilters;
-            setFunction((previousFilters) => ({
-                dimensions: [...previousFilters.dimensions, filter],
-                metrics: previousFilters.metrics,
-                tableCalculations: previousFilters.tableCalculations,
-            }));
+            if (isTemporary) {
```
ZeRego, line 879: If I follow correctly, this would remove the saved filter which could cause issues.

I suggest updating logic to:

(if isTemporary is true)
1. detect if has a matching dashboard filter
2. if yes, override filter
3. if no, detect if has a matching dashboard temporary filter
4. if yes, override temporary filter
5. if no, add temporary filter

(if isTemporary is false)
1. detect if has a matching dashboard filter
2. if yes, override filter
3. if no, add filter

ZeRego (changes requested):
left a comment with suggestion

packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
lakshz: Umm OK. I think the key question I want to ask is how we want to handle the override logic when a temporary filter matches an existing saved dashboard filter. 

Let me explain my current implementation:

The bottom three steps (in your logic explained above) can be bundled regardless of whether `isTemporary` is true or false. Those 3 steps say the same:

1. Check if temporary matches any temporary filter, or saved filters matches any saved filter. If yes, override. If not, add. Only the `setFunction` has to change, which is handled.

3. This can always be executed, regardless of `isTemporary`, to avoid `temporary<>temporary` and `saved<>saved` collisions.                  
                

The special case is when a temporary filter conflicts with an existing saved filter.
In that situation, I:
1. Remove the saved filter from the frontend state (it can still be restored via "reset filters")
2. Add the temporary filter instead

Is this override logic a problem? Another option is, I:
1. Remove the older saved filter.
2. Add the new saved filter.

My reasoning was simply that in non-edit mode, the “Add Filter” action should always create a temporary filter. However, in the latter logic, a temporary filter would keep a styling of a saved filter. I can switch to that if it aligns better with the expected behavior.

ZeRego: Perhaps we should rename this to `upsertDimensionDashboardFilter` to better represent the logic. 

> I want to ask is how we want to handle the override logic when a temporary filter matches an existing saved dashboard filter.

It should update the existing saved dashboard filter operator and value. 
Note that it should NOT remove the filter as it may contain other configuration (eg: tile target overrides).
`isTemporary` var would only be used when there is no match, to create the correct filter type.

packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
```diff
@@ -853,17 +853,73 @@ const DashboardProvider: React.FC<
...
+                    setDashboardTemporaryFilters((prev) => ({
+                        ...prev,
+                        dimensions: [...prev.dimensions, filter],
+                    }));
+                }
+            } else {
+                setDashboardTemporaryFilters((prev) => ({
+                    ...prev,
+                    dimensions: [...prev.dimensions, filter],
+                }));
```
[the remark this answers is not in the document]
lakshz: Fixed.

packages/frontend/src/providers/Dashboard/DashboardProvider.tsx
```diff
@@ -851,19 +851,75 @@ const DashboardProvider: React.FC<
...
+                                operator: filter.operator,
+                            };
+                        }
+                        return d;
+                    }),
+                };
+                setDashboardFilters(newDashboardFilters);
+                setHaveFiltersChanged(true);
+                return;
+            }
```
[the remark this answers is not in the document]
ZeRego: intended behaviour

PR #17894, opened 2025-11-05
feat: add project role management to SCIM user updates

ZeRego:
Closes: #XXXX

### Description:
Adds support for managing user project roles through the SCIM API. This implementation enables:

1. Parsing and validation of role IDs in the format `projectUuid:roleUuid`
2. Updating both system roles and custom roles for projects when a user is updated via SCIM
3. Proper validation of role arrays to ensure:
   - Exactly one organization role is provided
   - No duplicate project roles exist
   - All role values are valid

The changes include comprehensive tests for the new functionality, covering role parsing, validation, and the update process for both organization and project-level roles.

#### Update - add project editor role

Screenshot 2025-11-05 at 15 03 26

#### Update without `roles` property. Success and keeps existing roles
Screenshot 2025-11-05 at 15 04 14

#### Errors
Screenshot 2025-11-05 at 15 03 59
Screenshot 2025-11-05 at 15 03 48
Screenshot 2025-11-05 at 15 02 56

PR #17879, opened 2025-11-05
fix: improve SCIM role display names and update role counts

ZeRego:
Closes: #SCIM-role-display-names

### Description:
This PR improves the display names for SCIM roles by using proper capitalization. It adds `OrganizationMemberRoleLabels` to map role values to their human-readable display names (e.g., "admin" → "Admin"). The PR updates the SCIM service to use these labels when returning role information, making the interface more user-friendly. Tests have been updated to reflect these changes, including adjusting the expected number of roles from 19 to 20.

Using Org roles so it includes the 'member' role, which doesn't exist in project system roles.

Screenshot 2025-11-05 at 11 32 51

PR #17895, opened 2025-11-05
feat: add role validation and assignment during SCIM user creation

ZeRego:
Closes: #XXXX

### Description:
Refactors SCIM user role management by extracting the role update logic into a reusable `upsertUserRoles` method. This improves code organization and adds role validation when creating users. The changes also ensure that project roles are properly assigned during user creation, fixing an issue where roles were only being applied during user updates.

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.ts
@@ -457,7 +457,7 @@ export class ScimService extends BaseService {
                 organizationUuid,
                 userUuid: dbUser.userUuid,
                 roles: user.roles,
-                currentOrgRole: role, // will skip org role update since we just created it
+                currentOrgRole: role,
             });
 
             // verify user email on create if coming from scim


ZeRego:
* **#17925** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17925?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17914** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17914?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17899** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17895** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17895?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17895?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17894** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17879** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17879?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17872** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17872?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17851** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17851?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -737,6 +728,57 @@ export class ScimService extends BaseService {
...
+                                projectUuid,
+                                userUuid,
+                                roleUuid,
+                            );
+                        }
+                    } else if (
+                        isOrganizationMemberRole(roleUuid) &&
+                        roleUuid !== currentOrgRole
+                    ) {
+                        await this.organizationMemberProfileModel.updateOrganizationMember(
```
rephus, line 769: is this not conflicting with `createOrganizationMembershipByUuid` ?

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -737,6 +728,57 @@ export class ScimService extends BaseService {
...
+    }: {
+        organizationUuid: string;
+        userUuid: string;
+        roles: ScimUser['roles'];
+        currentOrgRole: string | undefined;
+    }) {
+        if (roles !== undefined && roles.length > 0) {
+            await Promise.all(
+                roles.map(async (role) => {
+                    const { roleUuid, projectUuid } = ScimService.parseRoleId(
```
rephus, line 748: will this throw an error if parsing fails ? if so, are we handling the error correctly ?

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -737,6 +728,57 @@ export class ScimService extends BaseService {
...
+                        role.value,
+                    );
+                    if (projectUuid) {
+                        if (isSystemRole(roleUuid)) {
```
rephus, line 752: we don't need to add system role for a project if the organization role is the same , since it inherits the org role.  I still think it is good  to do now, in case they change the org role in the future (project role should remain unmodified). 

And if orgRole > projectRole, we might want to throw a warning or return an error directly. What do you think ?

rephus (approved):
LGTM, small comment to think about, no blockers.

packages/backend/src/ee/services/ScimService/ScimService.ts
ZeRego: > I still think it is good to do now, in case they change the org role in the future (project role should remain unmodified).

exactly ;) I think this will always be allowed and is up to them.

> And if orgRole > projectRole, we might want to throw a warning or return an error directly. What do you think ?

There is no mechanism in SCIM to provide warnings and errors are too harsh. Will let this be flexible.

packages/backend/src/ee/services/ScimService/ScimService.ts
ZeRego: We marked the user extension schema as deprecated in the docs so they shouldn't be using both the extension role + roles properties. Worst case the role is overwritten.
Normal case, if the role match, it will skip the update.

PR #17980, opened 2025-11-10
fix: respect LIGHTDASH_QUERY_DEFAULT_LIMIT

magnew:
### Description:

For new explores we were adding a 500 row limit before checking the default. This makes it so new explores start with the limit undefined and it gets set to the `LIGHTDASH_QUERY_DEFAULT_LIMIT` when populating the store. 

Repro:
- set `LIGHTDASH_QUERY_DEFAULT_LIMIT=1000` in your instance
- Create a new explore
- Query limit (in the refresh button) should match the limit you set

### DIFF
--- a/packages/frontend/src/hooks/useExplorerRoute.ts
+++ b/packages/frontend/src/hooks/useExplorerRoute.ts
@@ -26,6 +26,7 @@ import {
     useExplorerDispatch,
     useExplorerSelector,
 } from '../features/explorer/store';
+import useApp from '../providers/App/useApp';
 import {
     defaultQueryExecution,
     defaultState,
@@ -212,6 +213,7 @@ export const useExplorerRoute = () => {
 export const useExplorerUrlState = (): ExplorerReduceState | undefined => {
     const { showToastError } = useToaster();
     const { search } = useLocation();
+    const { health } = useApp();
     const pathParams = useParams<{
         projectUuid: string;
         tableId: string | undefined;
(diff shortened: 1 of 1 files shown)


magnew:
* **#17980** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17980?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17980?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/useExplorerRoute.ts
```diff
@@ -227,27 +227,29 @@ export const useExplorerUrlState = (): ExplorerReduceState | undefined => {
...
+                    parsedChart ||
+                    ({
+                        tableName: pathParams.tableId,
+                        metricQuery: {
+                            exploreName: pathParams.tableId,
+                            dimensions: [],
+                            metrics: [],
+                            filters: {},
+                            sorts: [],
+                            limit: undefined,
```
magnew: This is the key

almeidabbm: Instead of setting `undefined` which breaks types and requires us to cast to unknown, can't we use the `useHealth` hook to fetch that value and default to 500 if `undefined`?

`limit: health.data?.query.defaultLimit ?? 500`

almeidabbm (approved):
LGTM 🚀

PR #17989, opened 2025-11-10
fix: Restore funnel chart color overrides by using field IDs as keys

tatianainama:
### Description:
Fixes a regression introduced in #17368 where funnel  chart color overrides stopped working when using column configuration. After the changes, funnel charts displayed human-readable field labels  instead of IDs (e.g., "Total Amount" instead of "payments_total_amount") but this caused color and label  overrides to break because:

  1. Color/label overrides were stored with field IDs as keys
  2. The data points now used labels as their `name` property
  3. When looking up `colorOverrides[name]`, the label didn't match the stored ID keys
  4. Result: All funnel steps used default colors instead of custom colors

This pr includes **both** an `id` and a `name` property, using ID for stable key for color/label overrides and name for labels in the chart

**Caveat**: Funnels created with custom colors between #17368 and this will not retain the colors, which I think is a valid tradeoff without needing to introduce a migration function to cover this short timeframe.

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/FunnelChartConfig/FunnelChartConfigTabs.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/FunnelChartConfig/FunnelChartConfigTabs.tsx
@@ -229,18 +229,15 @@ export const ConfigTabs: FC = memo(() => {
                                     .map((step) => {
                                         return (
                                             <StepConfig
-                                                key={step.name}
+                                                key={step.id}
+                                                id={step.id}
                                                 defaultColor={
-                                                    colorDefaults[step.name]
+                                                    colorDefaults[step.id]
                                                 }
                                                 defaultLabel={step.name}
                                                 swatches={[]}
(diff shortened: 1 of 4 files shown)


tatianainama:
* **#17989** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17989?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tatianainama:
### Merge activity

* **Nov 10, 4:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/17989).
* **Nov 10, 4:43 PM UTC**: @tatianainama merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/17989).

PR #18011, opened 2025-11-10
fix: new props causing 'save' in existing charts

magnew:
### Description:

**The problem:** 
New properties (showAxisTicks) were causing old charts to show the 'save changes' button, even when nothing was changed. This was because we diff the savedChart with the unsavedChartVersion and the diff was including new properties in the unsaved version. 

**The fix:**
Include the defaults for any new properties in the diff with the saved chart. 

**Repro:**
- Open a chart created before Nov 6
- Edit
- It will show the save button, even if nothing has changed (there is a `showAxisTicks: false in the unsaved version)`

### DIFF
--- a/packages/frontend/src/providers/Explorer/utils.ts
+++ b/packages/frontend/src/providers/Explorer/utils.ts
@@ -3,7 +3,7 @@ import omit from 'lodash/omit';
 import { EMPTY_CARTESIAN_CHART_CONFIG } from '../../hooks/cartesianChartConfig/useCartesianChartConfig';
 import { type ConfigCacheMap } from './types';
 
-const DEFAULTS: Record<ChartType, () => unknown> = {
+const DEFAULTS = {
     [ChartType.CARTESIAN]: () => ({ ...EMPTY_CARTESIAN_CHART_CONFIG }), // factory to avoid shared refs
     [ChartType.BIG_NUMBER]: () => ({ showTableNamesInLabel: false }),
     [ChartType.TABLE]: () => ({ showTableNames: false }),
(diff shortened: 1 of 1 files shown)


magnew:
* **#18011** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18011?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18011?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/providers/Explorer/utils.ts
```diff
@@ -41,23 +46,41 @@ export const getCachedPivotConfig = (
...
+// and merge with defaults to handle backwards compatibility when new properties are added
 export const cleanConfig = (config: ChartConfig): ChartConfig => {
     if (
         config.type === ChartType.CARTESIAN &&
         config.config?.eChartsConfig?.series
     ) {
+        // Merge with defaults to ensure old saved charts have new default properties
+        const defaults = DEFAULTS[ChartType.CARTESIAN]() as Record<
+            string,
+            AnyType
```
magnew: Is there a better way to type this?

almeidabbm: Replied in approval message

magnew:
> Works well! 🚀 As for the type question, if you remove the type from the `DEFAULTS` definition, then you can remove the casts below!
> 
> ```ts
> const DEFAULTS = {
>     [ChartType.CARTESIAN]: () => ({ ...EMPTY_CARTESIAN_CHART_CONFIG }), // factory to avoid shared refs
>     [ChartType.BIG_NUMBER]: () => ({ showTableNamesInLabel: false }),
>     [ChartType.TABLE]: () => ({ showTableNames: false }),
>     [ChartType.PIE]: () => ({ showLegend: false, valueLabel: 'outside' }),
>     [ChartType.FUNNEL]: () => ({}),
>     [ChartType.TREEMAP]: () => ({}),
>     [ChartType.CUSTOM]: () => ({}),
> };
> ```

Done. You think this is fine cause that type wasn't doing much anyway?

almeidabbm:
> > Works well! 🚀 As for the type question, if you remove the type from the `DEFAULTS` definition, then you can remove the casts below!
> > ```ts
> > const DEFAULTS = {
> >     [ChartType.CARTESIAN]: () => ({ ...EMPTY_CARTESIAN_CHART_CONFIG }), // factory to avoid shared refs
> >     [ChartType.BIG_NUMBER]: () => ({ showTableNamesInLabel: false }),
> >     [ChartType.TABLE]: () => ({ showTableNames: false }),
> >     [ChartType.PIE]: () => ({ showLegend: false, valueLabel: 'outside' }),
> >     [ChartType.FUNNEL]: () => ({}),
> >     [ChartType.TREEMAP]: () => ({}),
> >     [ChartType.CUSTOM]: () => ({}),
> > };
> > ```
> 
> Done. You think this is fine cause that type wasn't doing much anyway?

Yeah that type was not allowing for the actual types to be inferred and therefore would make them more generic which is what was causing the error, it couldn't determine which value type would be for each of the keys

PR #18019, opened 2025-11-11
feat: add image dimension support with URL-based rendering

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-130/investigate-images-on-table

Scrolling on virtual table rendering feels "ok". Having bigger images (eg: 48px) made it worse. 

Screenshot from 2025-11-11 11-15-50
Screenshot from 2025-11-11 11-16-28

## on image errors

image

### Description:
Added support for image dimensions in Lightdash. This feature allows users to define image URLs in their dbt models and display them in the UI. Images are rendered with tooltips that show larger versions on hover.

The implementation includes:
- Added `image` property to dimension types in dbt.ts and field.ts
- Updated the translator to convert image metadata from dbt models
- Added image cell formatting in useColumns.tsx with tooltip functionality
- Added an example in the jaffle-shop demo showing how to create image URLs in SQL

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/models/events.sql
+++ b/examples/full-jaffle-shop-demo/dbt/models/events.sql
@@ -1,1 +1,1 @@
-SELECT * FROM {{ ref('raw_events') }}
+SELECT *, 'https://placehold.co/300x100?text=id%20' || event as image_url  FROM {{ ref('raw_events') }}
--- a/examples/full-jaffle-shop-demo/dbt/models/events.yml
+++ b/examples/full-jaffle-shop-demo/dbt/models/events.yml
@@ -79,6 +79,14 @@ models:
           metrics:
             count:
               type: count
+      - name: image_url
+        description: "Image URL for the event from events.sql"
+        meta:
+          dimension:
+            type: string
+            groups: [ 'image' ]
+            image: 
+                url: "${value.raw}" 
       - name: event_id
         description: ""
         meta:
@@ -86,6 +94,28 @@ models:
             type: number
             groups: [ 'events', 'nested_group', 'another_nested_group' ]
           additional_dimensions:
+            image_event_id:
+              groups: [ 'image' ]
+              description: "Additional dimension of image URL for event_id. Uses event_id from SQL, but image URL template combines with event name from row"
+              type: string
+              # creates url in sql
+              # intentionally introduce some missing images to test the image cell
+              sql: |
+                case
+                  when ${event_id} > 100 then 'lightdash.com/missing-image.jpg'
+                  else 'placehold.co/300x100?text=id%20' || ${event_id}
+                  end
+              image:
+                # Template combines value (the URL from SQL) with other row fields
+                # Use liquidjs to combine the URL with the event name and upcase
+                url: "https://${value.raw}-${row.events.event.raw | upcase}" 
+            image_invalid_protocol:
+              groups: [ 'image' ]
+              description: "Invalid URL protocol"
+              type: string
+              sql: ${event_id}
+              image:
+                url: "file://${value.raw}" 
             event_tier:
               type: string
               description: Ratio of event_id to event
--- a/packages/common/src/types/dbt.ts
+++ b/packages/common/src/types/dbt.ts
@@ -168,6 +168,9 @@ export type DbtColumnLightdashDimension = {
     urls?: FieldUrl[];
     required_attributes?: Record<string, string | string[]>;
     ai_hint?: string | string[];
+    image?: {
+        url: string;
+    };
 } & DbtLightdashFieldTags;
 
 export type DbtColumnLightdashAdditionalDimension = Omit<
--- a/packages/frontend/src/components/common/Table/ImageCell.tsx
+++ b/packages/frontend/src/components/common/Table/ImageCell.tsx
@@ -1,0 +1,69 @@
+import { type Dimension } from '@lightdash/common';
+import { Tooltip } from '@mantine/core';
+import { IconPhotoOff } from '@tabler/icons-react';
+import { useState } from 'react';
+import MantineIcon from '../MantineIcon';
+
+export const BrokenImageCell = ({
+    imageUrl,
+    error,
+}: {
+    imageUrl: string;
+    error?: string;
+}) => {
+    return (
+        <Tooltip
+            withinPortal
+            w="400px"
+            multiline
+            label={`Could not load image "${imageUrl}" ${
+                error ? `: ${error}` : ''
+            }`}
+        >
+            <span style={{ display: 'inline-block', lineHeight: 0 }}>
+                <MantineIcon icon={IconPhotoOff} size="xxl" color="gray" />{' '}
+                {/* xxl =32px same size as image cell */}
+            </span>
+        </Tooltip>
+    );
+};
+
+export const ImageCell = ({
+    imageUrl,
+}: {
+    item: Dimension;
+    imageUrl: string;
+}) => {
+    const [isBroken, setIsBroken] = useState(false);
+
+    if (isBroken) {
+        return <BrokenImageCell imageUrl={imageUrl} />;
+    }
+
+    return (
+        <Tooltip
+            withinPortal
+            label={
+                // Full image in tooltip
+                <img
+                    src={imageUrl}
+                    alt=""
+                    style={{ maxWidth: '400px', maxHeight: '400px' }}
+                />
+            }
+        >
+            {/* Small thumbnail in table cell */}
+            <img
+                src={imageUrl}
+                alt=""
+                style={{
+                    height: '32px',
+                    width: 'auto',
+                    display: 'block',
+                    objectFit: 'contain',
+                }}
+                onError={() => setIsBroken(true)}
+            />
+        </Tooltip>
+    );
+};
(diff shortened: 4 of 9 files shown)


rephus:
* **#18019** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18019?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18019?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/full-jaffle-shop-demo/dbt/models/events.yml
```diff
@@ -86,6 +86,11 @@ models:
             type: number
             groups: [ 'events', 'nested_group', 'another_nested_group' ]
           additional_dimensions:
+            image: 
+              type: string
+              sql: "'https://placehold.co/300x100?text=id%20' || ${event_id}" # creates url in sql
```
rephus: format string on postgres with URL.

packages/common/src/types/dbt.ts
```diff
@@ -168,6 +168,9 @@ export type DbtColumnLightdashDimension = {
     urls?: FieldUrl[];
     required_attributes?: Record<string, string | string[]>;
     ai_hint?: string | string[];
+    image?: {
+        url: string;
```
rephus, line 172: later on, we'll add extra properties about dimensions, etc.

rephus: actually, it is so straightforward, we could add dimensions now , or on a subPR

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -156,6 +157,38 @@ const formatBarDisplayCell = (
     );
 };
 
+const formatImageCell = (item: Dimension, value: string) => {
```
rephus: item is not used at the moment, but we'll use it soon to add dimensions, etc.

examples/full-jaffle-shop-demo/dbt/models/events.yml
rephus: placehold.co generates dynamic images using URLs (great for labels and debugging)

rephus: If service goes down/blocks us, it will not break any implementation or  tests

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -156,6 +158,22 @@ const formatBarDisplayCell = (
...
+        if (!['http:', 'https:'].includes(imageUrl.protocol)) {
+            console.error(`Invalid image protocol "${value}"`);
+            return '-'; // Reject non-http(s) URLs
+        }
+
+        return <ImageCell item={item} imageUrl={imageUrl.href} />;
+    } catch (error) {
+        console.error(`Invalid image URL "${value}":  ${error}`);
+    }
+    return '-';
```
rephus: we could also return the `Could not load image`  component here instead. This is for invalid URLs, more than missing images, so I think this is ok.

packages/frontend/src/components/common/Table/ImageCell.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useState } from 'react';
+import MantineIcon from '../MantineIcon';
+
+export const ImageCell = ({
+    imageUrl,
+}: {
+    item: Dimension;
+    imageUrl: string;
+}) => {
+    const [isBroken, setIsBroken] = useState(false);
```
rephus, line 37: I tried to handle missing images without states , for performance, but that made everything worse and unreadable. I think this is ok.

packages/frontend/src/components/common/Table/ImageCell.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    if (isBroken) {
+        return (
+            <Tooltip withinPortal label={`Could not load image: ${imageUrl}`}>
+                <span style={{ display: 'inline-block', lineHeight: 0 }}>
+                    <MantineIcon icon={IconPhotoOff} size="xxl" color="gray" />
```
rephus: xxl=32px. This will ensure even missing images have the same cell height.

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -156,6 +158,22 @@ const formatBarDisplayCell = (
     );
 };
 
+const formatImageCell = (item: Dimension, value: string) => {
+    try {
+        // Validate image url, otherwise return -
+        const imageUrl = new URL(value);
```
rephus: image URL handling outside component rendering

rephus:
@ZeRego are we ok with just supporting images on "additional dimensions" at the moment ? I don't think an image should be a metric or even table calculation, right ?

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -156,6 +158,22 @@ const formatBarDisplayCell = (
...
+            console.error(`Invalid image protocol "${value}"`);
+            return '-'; // Reject non-http(s) URLs
+        }
+
+        return <ImageCell item={item} imageUrl={imageUrl.href} />;
+    } catch (error) {
+        console.error(`Invalid image URL "${value}":  ${error}`);
+    }
+    return '-';
+};
```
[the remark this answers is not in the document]
rephus: we should not limit domains, checking the protocol should be enough to reject most of the vulnerabilities

image

examples/full-jaffle-shop-demo/dbt/models/events.sql
```diff
@@ -1 +1 @@
-SELECT * FROM {{ ref('raw_events') }}
+SELECT *, 'https://placehold.co/300x100?text=id%20' || event as image_url  FROM {{ ref('raw_events') }}
```
rephus, line 1: adds full image URL as a column in sql. So we can test it directly with dimensions, not just with additional dimensions

packages/common/src/types/dbt.ts
ZeRego: Don't forget to update the schemas:
- packages/common/src/schemas/json/lightdash-dbt-2.0.json
- packages/common/src/dbt/schemas/lightdashMetadata.json

packages/frontend/src/components/common/Table/ImageCell.tsx
```diff
@@ -0,0 +1,52 @@
+import { type Dimension } from '@lightdash/common';
+import { Tooltip } from '@mantine/core';
+import { IconPhotoOff } from '@tabler/icons-react';
+import { useState } from 'react';
+import MantineIcon from '../MantineIcon';
+
+export const ImageCell = ({
+    imageUrl,
+}: {
+    item: Dimension;
```
ZeRego, line 34: Is `item` used?

packages/frontend/src/hooks/useColumns.tsx
ZeRego: I expected this method to receive `item: Dimension, value: ResultValue` and generate the final URL using LiquidJS. At the moment, does it matter what I define in YML `image.url` 🤔

packages/frontend/src/components/common/Table/ImageCell.tsx
rephus: not at the moment, but we will use it soon when we add the custom dimensions and other properties

packages/frontend/src/hooks/useColumns.tsx
rephus: Yes, I missed the liquidJS part, I added the logic to formatImageCell, keeping he ImageCell component the same. 

I updated my events.yml to use some of the liquidjs templating 

>                 url: "https://${value.raw}-${row.events.event.raw | upcase}" 


Screenshot from 2025-11-12 09-18-46

I'm also showing brokenImage with a useful tooltip when liquidjs has invalid formatting

image

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -156,6 +163,71 @@ const formatBarDisplayCell = (
...
+            `Invalid image URL template "${item.image?.url}": ${error}`,
+        );
+        return (
+            <BrokenImageCell
+                imageUrl={item.image?.url || (value.raw as string)}
+                error={getErrorMessage(error) as string}
+            />
+        );
+    }
+};
```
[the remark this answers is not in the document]
rephus: Yeah, I know, but this is required for properly formatting the url using liquidjs. I can't think of another way to optimize this.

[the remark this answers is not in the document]
rephus: Ok, I have 2 possible optimizations: 

- only extract if the row contains liquidJS syntax, to skip processing all this data. 

Or 

- extract this at a row level , not at a cell level, if the row contains at least 1 image.  (this will optimize rows with more than 2 images) . This might potentially not impact many use cases, and we'll have to refactor the existing code to pass this parsed row info. So I'll add a comment about this potential improvement, but skip this for now.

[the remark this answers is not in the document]
rephus: image

ZeRego (approved):

PR #18018, opened 2025-11-11
fix: project admins to view organization roles

ZeRego:
Closes: 

### Description:
This PR enhances the roles and permissions system by allowing project admins to view organization roles. Previously, only organization admins could view roles, but now project admins can also access this information.

#### before

Screenshot 2025-11-10 at 13 52 17

#### after 
CleanShot 2025-11-11 at 10 09 57@2x

PR #17984, opened 2025-11-10
fix: use role value instead of label for project role matching

ZeRego:
Closes: https://github.com/lightdash/lightdash/issues/17973
Closes: #17956

### Description:
Fixed a bug in project access management where user roles were incorrectly matched. The code now compares role values instead of labels when finding project roles, and properly uses role IDs instead of role names when processing project role assignments.

#### Before (wrong roles)

Screenshot 2025-11-07 at 15 23 51

#### Before (missing roles)
Screenshot 2025-11-10 at 13 57 11

#### After (correct roles)
Screenshot 2025-11-10 at 13 43 02
Screenshot 2025-11-10 at 14 00 11

Screenshot 2025-11-10 at 13 43 22

PR #18020, opened 2025-11-11
fix: support multiple group access roles in project access component

ZeRego:
Closes: #[issue number]

### Description:
Updated the project access component to support users belonging to multiple groups with different access levels. Previously, the code only considered the first group a user belonged to, but now it properly evaluates all groups and determines the highest role across them.

The changes include:
- Refactored `userGroupAccess` to `userGroupAccesses` to store an array of all groups a user belongs to
- Updated the role calculation logic to consider all groups when determining the highest role
- Modified the UI to display all groups a user belongs to instead of just the first one
- Enhanced the access warning system to evaluate permissions across multiple groups

This fixes the issue where users belonging to multiple groups weren't having all their group permissions properly considered and displayed.

### before

It was only handling the first group. 

### after

CleanShot 2025-11-11 at 10 31 58@2x
CleanShot 2025-11-11 at 10 31 16@2x

### DIFF
--- a/packages/frontend/src/components/ProjectAccess/ProjectAccessRowV2.tsx
+++ b/packages/frontend/src/components/ProjectAccess/ProjectAccessRowV2.tsx
@@ -91,8 +91,8 @@ const ProjectAccessRowV2: FC<Props> = ({
     );
 
     // Find groups the user belongs to that have project access
-    const userGroupAccess = useMemo((): UserGroupAccess | null => {
-        if (!organizationGroups || !groupRoles) return null;
+    const userGroupAccesses = useMemo((): UserGroupAccess[] => {
+        if (!organizationGroups || !groupRoles) return [];
 
         // Find groups where the user is a member
         const userGroups = organizationGroups.filter((group) => {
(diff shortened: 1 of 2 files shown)


ZeRego:
* **#18020** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18020?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18020?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#18018** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18018?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17984** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/17984?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/components/ProjectAccess/ProjectAccessRowV2.tsx
```diff
@@ -116,8 +116,7 @@ const ProjectAccessRowV2: FC<Props> = ({
             })
             .filter((item): item is UserGroupAccess => item !== null);
 
-        // TODO can an user be in multiple groups ? what happens then with permisisons
```
ZeRego, line 119: This PR addresses this TODO

stevelikesmusic: The thing with this was that we don't have a way create a role hierarchy when using custom roles. What is the highest role if someone is assigned a custom role? The messaging that was there is to indicate that a custom role could be more or less permissive than the other assigned system roles. But now, we don't communicate that concern, so it's less obvious that a user might have a different set of permissions than expected without looking at the scopes of the custom role.

ZeRego: The message is still the same. I solved the problem where the messaging logic was only considering the first group role.

PR #18017, opened 2025-11-11
feat: add Chart to embed SDK

stevelikesmusic:
Closes: CENG-104

### Description:

Adds a Chart component to the Lightdash Embed SDK. This is a two part PR with stricter permissions coming in a [follow up](https://linear.app/lightdash/issue/CENG-148/add-stricter-permissions-for-chart-embeds).

### Testing

I added an http-client env file. This works for both VSCode and WebStorm HTTP clients. You can create an http-client.env.private.json file to override values in the public file. 

Create an embed URL for the charts you need and use that in the SDK test app.

#### Render Embedded Chart

JWTs for charts work only for charts. Dashboards will not load. 

#### Allow chart slugs in addition to UUIDs

Allow both UUIDs and slugs. Debugging breakpoints are optional. 

#### Only allow chart UUID in JWT

Only load embedded charts for a chart UUID or slug that matches what's bound to the JWT

PR #18022, opened 2025-11-11
feat: add stricter embed chart permissions

stevelikesmusic:
Closes: CENG-148

### Description:

Refactored the JWT authentication to use a more structured `EmbedContent` type instead of separate `contentUuid` and `contentType` fields. This change improves type safety and allows for more detailed content information to be passed around, including chart UUIDs, dashboard UUID, and available explores.

The new structure also enables more granular permissions checking, particularly for chart embeds where we can now properly check permissions against specific explores.

This lays the foundation for front-loading all dashboard charts so we can further implement strict permissions for dashboards and eventually AI Agents.

These changes should only affect single chart embeds and not dashboards, dashboard charts, underlydata, or explores (all dashboard related concerns)

### DIFF
--- a/packages/backend/src/auth/account/account.ts
+++ b/packages/backend/src/auth/account/account.ts
@@ -10,6 +10,7 @@ import {
     applyEmbeddedAbility,
     buildAccountHelpers,
     CreateEmbedJwt,
+    EmbedContent,
     ForbiddenError,
     MemberAbility,
     OauthAccount,
@@ -76,28 +77,19 @@ export const fromJwt = ({
     decodedToken,
     embed,
     source,
-    contentUuid,
-    contentType = 'dashboard',
     userAttributes,
+    content,
 }: {
     decodedToken: CreateEmbedJwt;
     embed: OssEmbed;
     source: string;
-    contentUuid?: string;
-    contentType?: 'dashboard' | 'chart';
     userAttributes: UserAccessControls;
+    content: EmbedContent;
 }): AnonymousAccount => {
     const builder = new AbilityBuilder<MemberAbility>(Ability);
     const externalId = getExternalId(decodedToken, source, embed.organization);
 
-    applyEmbeddedAbility(
-        decodedToken,
-        contentUuid,
-        contentType,
-        embed,
-        externalId,
-        builder,
-    );
+    applyEmbeddedAbility(decodedToken, content, embed, externalId, builder);
     const abilities = builder.build();
 
     return createAccount({
@@ -109,8 +101,7 @@ export const fromJwt = ({
         organization: embed.organization,
         embed,
         access: {
-            contentId: contentUuid,
-            contentType,
+            content,
             filtering: decodedToken.content.dashboardFiltersInteractivity,
             parameters: decodedToken.content.parameterInteractivity,
             controls: userAttributes,
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -464,12 +464,24 @@ export class AsyncQueryService extends ProjectService {
             account,
         );
 
-        if (
+        const isForbidden =
             account.user.ability.cannot(
                 'view',
-                subject('Project', { organizationUuid, projectUuid }),
-            )
-        ) {
+                subject('Project', {
+                    organizationUuid,
+                    projectUuid,
+                }),
+            ) &&
+            account.user.ability.cannot(
+                'view',
+                subject('Explore', {
+                    organizationUuid,
+                    projectUuid,
+                    exploreNames: [queryHistory.metricQuery.exploreName],
+                }),
+            );
+
+        if (isForbidden) {
             throw new ForbiddenError();
         }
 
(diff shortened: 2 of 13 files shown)


stevelikesmusic:
* **#18022** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18022?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18022?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#18017** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18017?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/common/src/types/auth.ts
```diff
@@ -82,11 +82,22 @@ export type UserAccessControls = {
...
+/**
+ * Details about the content bound to the given JWT
+ */
+export type EmbedContent = {
+    /** The dashboard UUID the JWT may have access to */
+    dashboardUuid?: string;
+    /** The chart IDs the JWT dashboard may have access to */
+    chartUuids: string[];
+    /** Explores available to the embedded dashboard */
+    explores: string[];
```
stevelikesmusic, line 94: One question here design wise: would we prefer to pass a list of charts rather than break up chartUuids and explore/table names? Maybe we start here and if we need to, we could have a Picked version of charts and/or dashboards.

rephus:
Failing tests

packages/backend/src/auth/account/account.ts
```diff
@@ -76,28 +77,19 @@ export const fromJwt = ({
     decodedToken,
     embed,
     source,
-    contentUuid,
-    contentType = 'dashboard',
```
rephus, line 80: Most of the chat properties or contentUuid is new, but I believe this contentType already existed since embed v1. So we need to be careful and assume people will keep using this (eg: do not throw validation errors if this property exists) . 

I'd rather flag is a deprecated if yo uthink this might cause issues.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -464,12 +464,24 @@ export class AsyncQueryService extends ProjectService {
...
             account.user.ability.cannot(
                 'view',
-                subject('Project', { organizationUuid, projectUuid }),
-            )
-        ) {
+                subject('Project', {
+                    organizationUuid,
+                    projectUuid,
+                }),
+            ) &&
```
rephus, line 474: should this be an OR ?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -1695,16 +1707,24 @@ export class AsyncQueryService extends ProjectService {
...
+                    const isForbidden =
                         account.user.ability.cannot(
                             'view',
                             subject('Project', {
                                 organizationUuid,
                                 projectUuid,
                             }),
-                        )
-                    ) {
+                        ) &&
```
rephus, line 1717: OR ?

packages/common/src/authorization/jwtAbility.ts
```diff
@@ -70,37 +67,44 @@ const chartAbilities: EmbeddedAbilityBuilder = ({
     can('view', 'SavedChart', {
         access: {
             $elemMatch: {
-                chartUuid: contentUuid,
+                chartUuid: content.chartUuids[0],
```
rephus, line 70: what happens if multiple `content.chartUuids` are provided (or none), should we loop them ?

stevelikesmusic: We should only have one chartUuid for a chart. I'll add a check when parsing. The types don't allow for optional values

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
stevelikesmusic: OR would fail because we're not giving chart embeds the `view Project` permission. OR would work if we're checking a positive condition: `can view project OR can view explore`. Would that be more readable rather than `cannot view project AND cannot view explore`?

packages/backend/src/auth/account/account.ts
stevelikesmusic: I thought contentType was new for charts as well, because we never needed to check before—everything was dashboard. By the time we're here, we're past the API schema, so can't we rename so long as we make sure everything is updated? The inputs via the API are still the same.

ZeRego (approved):
lgtm 👍 but Javi should also approve this since is working closer with embed charts.

rephus (approved):

stevelikesmusic:
### Merge activity

* **Nov 13, 3:56 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/18022).
* **Nov 13, 3:56 PM UTC**: @stevelikesmusic merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/18022).

PR #18030, opened 2025-11-11
fix: filter out hidden fields when getting available explores

joaoviana:
Closes: #N/A

### Description:

Filter out hidden dimensions and metrics in the AI agent service when retrieving explores, rather than filtering them in the response generation. This moves the filtering logic to a single place in the codebase, making it more maintainable and consistent.

### DIFF
--- a/packages/backend/src/ee/services/AiAgentService.ts
+++ b/packages/backend/src/ee/services/AiAgentService.ts
@@ -364,6 +364,14 @@ export class AiAgentService {
         return false;
     }
 
+    /**
+     * Get all available explores for an AI agent
+     * Filters out hidden dimensions and metrics from all tables
+     * @param user - The user making the request
+     * @param projectUuid - The project UUID
+     * @param availableTags - The available tags
+     * @returns The available explores
+     */
     async getAvailableExplores(
         user: SessionUser,
         projectUuid: string,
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#18030** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/services/AiAgentService.ts
```diff
@@ -411,7 +419,36 @@ export class AiAgentService {
...
+                                        metrics: Object.fromEntries(
+                                            Object.entries(
+                                                table.metrics,
+                                            ).filter(
+                                                ([, metric]) => !metric.hidden,
+                                            ),
+                                        ),
+                                    },
+                                ],
+                            ),
```
IrakliJani, line 449: Could we extract this the same way we have `filterExploreByTags`? It would be nice to have a couple of tests as well. you can also use lodash filter on object to filter out an object properties. see `filterExploreByTags` as a reference

Issue #18029, opened 2025-11-11
Databricks queries infinitely pending


pimmey:
### Description

There doesn't seem to be timeout protection for Databricks connection. With small resources, like a small number of max concurrent queries, we won't know if something went wrong and the query will stay with `pending` status indefinitely. 

### Steps to Reproduce the Bug or Issue

1. Have a Databricks warehouse configured with smaller resoures
2. Try to execute multiple queries in parallel, such as loading a dashboard or exporting a dashboard

### Cloud or self-hosting

None

PR #18035, opened 2025-11-11
feat: add timeout to databricks config under feature flag

pimmey:
Closes: #18029

### Description:
Making use of internal Databricks timeout configs, they also have default re-try mechanism built-in:
  - retryMaxAttempts: 5 
  - retriesTimeout: 15 minutes

PR #18050, opened 2025-11-12
fix: properly close Databricks connection on session creation error

pimmey:
Relates to: #18029 

### Description:
Fix connection handling in DatabricksWarehouseClient to properly close connections when session creation fails. This prevents connection leaks by ensuring that if an error occurs during session creation, any established connection is properly closed before the error is propagated.

The changes include:
- Making the connection variable nullable by adding `| undefined` type
- Adding error handling to close the connection if it exists when session creation fails
- Using optional chaining when closing the connection in the close method

### DIFF
--- a/packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
@@ -277,27 +277,28 @@ export class DatabricksWarehouseClient extends WarehouseBaseClient<CreateDatabri
 
     private async getSession() {
         const client = new DBSQLClient({});
-        let connection: IDBSQLClient | undefined;
+        let connection: IDBSQLClient;
         let session: IDBSQLSession;
 
         try {
             connection = await client.connect(this.connectionOptions);
+        } catch (e: AnyType) {
+            throw new WarehouseConnectionError(getErrorMessage(e));
+        }
 
+        try {
             session = await connection.openSession({
                 initialCatalog: this.catalog,
                 initialSchema: this.schema,
             });
         } catch (e: AnyType) {
-            if (connection) {
-                try {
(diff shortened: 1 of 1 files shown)


pimmey:
* **#18050** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18050?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#18035** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
```diff
@@ -277,7 +277,7 @@ export class DatabricksWarehouseClient extends WarehouseBaseClient<CreateDatabri
 
     private async getSession() {
         const client = new DBSQLClient({});
-        let connection: IDBSQLClient;
+        let connection: IDBSQLClient | undefined;
```
almeidabbm: Instead of changing this to be optional we could use 2 try catches, 1 around `client.connect` that will handle what happens when the connection fails and throws out a `WarehouseConnectionError` and then another one around `openSession` which will handle the `connection.close()` and also throw a `WarehouseConnectionError`

almeidabbm (approved):

packages/warehouses/src/warehouseClients/DatabricksWarehouseClient.ts
pimmey: Good call!

PR #18063, opened 2025-11-12
fix: show explore error on tile instead of loading

magnew:
### Description:

**The problem:** errors in models resulted in dashboard tiles showing loading forever. This caused scheduled deliveries to show loading spinners when there were model errors. Not that errors in the query were already displayed properly. 

Screenshot 2025-11-12 at 14 34 15

**Fix:** propagate errors from useExplore to the tiles

**Repro:**
- Create a chart
- Add it to a dashboard
- Change the model to break the chart and refresh dbt
- Chart tile will load forever


magnew:
Tests passed on the first run? 🤨

PR #18075, opened 2025-11-12
feat: improve oauth page styling

owlas:
**AFTER**

CleanShot 2025-11-12 at 17 13 10@2x

CleanShot 2025-11-12 at 17 13 36@2x

---

**BEFORE**

CleanShot 2025-11-12 at 17 19 59@2x

CleanShot 2025-11-12 at 17 20 06@2x

### DIFF
--- a/packages/common/src/utils/oauth.ts
+++ b/packages/common/src/utils/oauth.ts
@@ -308,7 +308,7 @@ export interface OAuthUser {
     organizationName: string;
 }
 
-export interface OAuthScope {
+export interface OAuthScopeDescription {
     title: string;
     description: string;
 }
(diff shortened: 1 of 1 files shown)


packages/common/src/utils/oauth.ts
```diff
@@ -61,6 +61,23 @@ export const oauthPageStyles = `
     }
 `;
 
+// Lightdash logo SVG
+const LIGHTDASH_LOGO_SVG = `
```
rephus, line 65: can't we use an existing image in the repo, or upload the file as svg directly , rather than in code ?

packages/common/src/utils/oauth.ts
```diff
@@ -238,6 +328,54 @@ export interface OAuthRedirectParams {
...
+};
+
+/**
+ * Gets a friendly client name from the client_id
+ */
+export const getClientName = (clientId: string): string => {
+    const clientNames: Record<string, string> = {
+        'lightdash-cli': 'Lightdash CLI tool',
+    };
+    return clientNames[clientId] || clientId;
```
rephus, line 376: this looks like an overcomplicated way of formatting a single string... I'd used a switch instead.

PR #18098, opened 2025-11-13
feat: add Databricks OAuth U2M authentication support

rephus:
Closes: #7890

Internal docs: https://www.notion.so/lightdash/Databricks-Oauth-2aaa63207a7a804bb4c9fbe9ae845b53

Before:

only M2M was implemented

Now: 

We support 3 different U2M auth flows: 
- no client (we'll use dbt-databricks as default client) 
- client_id
- client_id and client_secret (requires extra ENV variable) 

More info:  https://www.notion.so/lightdash/Databricks-Oauth-2aaa63207a7a804bb4c9fbe9ae845b53?source=copy_link#2aba63207a7a806e8e5bd600f9299f63

https://lightdash.slack.com/archives/C08UQLRU947/p1763111427889149?thread_ts=1762957465.551329&cid=C08UQLRU947

image

Screenshot from 2025-11-14 13-25-14

### Description:
Add support for Databricks OAuth User-to-Machine (U2M) authentication in the CLI. This allows users to authenticate with Databricks using their browser, similar to how the Databricks CLI works.

The implementation:
1. Adds a new `oauth_u2m` authentication type to the Databricks credentials
2. Creates a local callback server to handle the OAuth flow
3. Opens the user's browser for authentication
4. Exchanges the authorization code for access and refresh tokens
5. Keeps tokens in memory only (not stored in profiles.yml)

This makes it easier for users to connect to Databricks without having to create and manage personal access tokens.

### DIFF
--- a/packages/cli/src/dbt/targets/Databricks/oauth.ts
+++ b/packages/cli/src/dbt/targets/Databricks/oauth.ts
@@ -1,30 +1,10 @@
 import { AuthorizationError } from '@lightdash/common';
-import { exec } from 'child_process';
 import * as http from 'http';
 import fetch from 'node-fetch';
 import { generators } from 'openid-client';
 import { URL } from 'url';
-import { promisify } from 'util';
 import GlobalState from '../../../globalState';
-
-const execAsync = promisify(exec);
-
-// Helper function to open browser
-const openBrowser = async (url: string): Promise<void> => {
-    try {
-        const { platform } = process;
-
-        if (platform === 'darwin') {
-            await execAsync(`open "${url}"`);
-        } else if (platform === 'win32') {
-            await execAsync(`start "${url}"`);
-        } else {
-            await execAsync(`xdg-open "${url}"`);
-        }
-    } catch (error) {
-        GlobalState.debug(`> Could not open browser automatically: ${error}`);
-    }
-};
+import { openBrowser } from '../../../handlers/login/oauth';
 
 /**
  * Databricks OAuth tokens result
@@ -122,33 +102,11 @@ export const performDatabricksOAuthFlow = async (
     });
 
     // Start the server on port 8020 (standard for Databricks CLI)
-    // If that fails, try random port
     const preferredPort = 8020;
     await new Promise<void>((resolve, reject) => {
         server.on('error', (err: NodeJS.ErrnoException) => {
-            if (err.code === 'EADDRINUSE') {
-                // Port 8020 is busy, try random port
-                GlobalState.debug(
-                    `> Port ${preferredPort} is busy, using random port`,
-                );
-                server.listen(0, () => {
-                    const address = server.address();
-                    if (address === null)
-                        throw new Error('Failed to get server address');
-
-                    if (typeof address === 'object') {
-                        port = address.port;
-                    } else {
-                        port = parseInt(address.toString(), 10);
-                    }
-                    GlobalState.debug(
-                        `> OAuth callback server listening on port ${port}`,
-                    );
-                    resolve();
-                });
-            } else {
-                reject(err);
-            }
+            // perhaps 8020 port is busy, but we can't use a random port
+            reject(err);
         });
 
         server.listen(preferredPort, () => {
--- a/packages/cli/src/handlers/login/oauth.ts
+++ b/packages/cli/src/handlers/login/oauth.ts
@@ -1,0 +1,23 @@
+// Helper function to open browser
+
+import { exec } from 'child_process';
+import { promisify } from 'util';
+import GlobalState from '../../globalState';
+
+const execAsync = promisify(exec);
+
+export const openBrowser = async (url: string): Promise<void> => {
+    try {
+        const { platform } = process;
+
+        if (platform === 'darwin') {
+            await execAsync(`open "${url}"`);
+        } else if (platform === 'win32') {
+            await execAsync(`start "${url}"`);
+        } else {
+            await execAsync(`xdg-open "${url}"`);
+        }
+    } catch (error) {
+        GlobalState.debug(`> Could not open browser automatically: ${error}`);
+    }
+};
--- a/packages/cli/src/handlers/oauthLogin.ts
+++ b/packages/cli/src/handlers/oauthLogin.ts
@@ -3,34 +3,14 @@ import {
     generateOAuthErrorResponse,
     generateOAuthSuccessResponse,
 } from '@lightdash/common';
-import { exec } from 'child_process';
 import * as http from 'http';
 import fetch from 'node-fetch';
 import { generators, Issuer } from 'openid-client';
 import { URL } from 'url';
-import { promisify } from 'util';
 import GlobalState from '../globalState';
 import * as styles from '../styles';
+import { openBrowser } from './login/oauth';
 import { generatePersonalAccessToken } from './login/pat';
-
-const execAsync = promisify(exec);
-
-// Helper function to open browser
-const openBrowser = async (url: string): Promise<void> => {
-    try {
-        const { platform } = process;
-
-        if (platform === 'darwin') {
-            await execAsync(`open "${url}"`);
-        } else if (platform === 'win32') {
-            await execAsync(`start "${url}"`);
-        } else {
-            await execAsync(`xdg-open "${url}"`);
-        }
-    } catch (error) {
-        GlobalState.debug(`> Could not open browser automatically: ${error}`);
-    }
-};
 
 export const loginWithOauth = async (
     url: string,


rephus:
* **#18102** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18098** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18098?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/dbt/targets/Databricks/oauth.ts
```diff
@@ -0,0 +1,259 @@
...
+    expiresAt: number; // Unix timestamp in seconds
+}
+
+/**
+ * Perform Databricks U2M OAuth flow
+ * Opens browser for user authentication and exchanges authorization code for tokens
+ * @param host Databricks workspace host
+ * @param clientId OAuth client ID (defaults to 'databricks-cli')
+ */
+export const performDatabricksOAuthFlow = async (
```
rephus, line 24: our U2M code implements full oauth flow, including a redirect server

Databricks offers an automatic authentication method, using their databricks CLI , but when using SDK's authType: 'databricks-oauth', we have no access to the tokens. The SDK handles everything internally, but doesn't persist across CLI process restarts. Each lightdash command =new process = new OAuth flow.

I tested that approach and the UX was worse since we can't have access to the tokens.

packages/cli/src/handlers/dbt/getWarehouseClient.ts
```diff
@@ -309,12 +311,53 @@ export default async function getWarehouseClient(
...
+                credentials.token = accessToken;
+            } catch (e) {
+                GlobalState.debug(
+                    `> Failed to exchange Databricks OAuth credentials for access token: ${getErrorMessage(
+                        e,
+                    )}`,
+                );
+                console.warn(
+                    styles.error(
+                        `\nFailed to authenticate with Databricks using M2M OAuth (client_id and client_secret). ` +
```
rephus, line 330: edge case, we assumed this was M2M becuase there was a client and secret, but this auth method doesn't work for U2M, use an ENV variable . 

Based on customer feedback, we can default to U2M instead.

packages/cli/src/dbt/targets/Databricks/oauth.ts
```diff
@@ -0,0 +1,269 @@
...
+            res.end(
+                '<html><body><h1>Authentication Failed</h1><p>Access denied. You can close this window.</p></body></html>',
+            );
+            return;
+        }
+        if (error) {
+            rejectAuth(new AuthorizationError(`OAuth error: ${error}`));
+            res.writeHead(400);
+            res.end(
+                `<html><body><h1>Authentication Failed</h1><p>Error: ${error}</p></body></html>`,
```
stevelikesmusic, line 66: Is this all a common pattern we can abstract for all our browser based Oauth flows in the CLI?

packages/cli/src/dbt/targets/Databricks/oauth.ts
```diff
@@ -0,0 +1,269 @@
...
+            await execAsync(`open "${url}"`);
+        } else if (platform === 'win32') {
+            await execAsync(`start "${url}"`);
+        } else {
+            await execAsync(`xdg-open "${url}"`);
+        }
+    } catch (error) {
+        GlobalState.debug(`> Could not open browser automatically: ${error}`);
+    }
+};
```
stevelikesmusic: Is this something that can be abstracted? I assume all CLI Oauth flows would use this mapping?

packages/cli/src/dbt/targets/databricks.ts
```diff
@@ -107,22 +123,36 @@ export const convertDatabricksSchema = (
...
+    // OAuth authentication
     if (authType === 'oauth') {
-        if (!target.client_id || !target.client_secret) {
-            throw new ParseError(
-                'Databricks OAuth authentication requires client_id and client_secret in profiles.yml',
-            );
+        // Determine authentication type: check env var first, then auto-detect
+        let authenticationType: DatabricksAuthenticationType;
+
+        const databricksOAuthEnv = process.env.DATABRICKS_OAUTH?.toLowerCase();
```
stevelikesmusic, line 131: Does this get parsed in lightdashConfig?

packages/cli/src/dbt/targets/databricks.ts
```diff
@@ -23,10 +23,14 @@ export type DatabricksTarget = {
     http_path: string;
     // PAT authentication
     token?: string;
-    // OAuth M2M authentication
+    // oauth can reference either M2M or U2M
     auth_type?: 'token' | 'oauth';
```
stevelikesmusic, line 27: Do these map as?

- M2M: 'token'
- U2M: 'oauth'

packages/cli/src/dbt/targets/Databricks/oauth.ts
```diff
@@ -0,0 +1,269 @@
...
+
+        GlobalState.debug(
+            `> OAuth access token: ${accessToken.substring(0, 10)}...`,
+        );
+        GlobalState.debug(
+            `> OAuth refresh token: ${refreshToken.substring(0, 10)}...`,
+        );
+
+        // Calculate expiration timestamp (current time + expires_in)
+        const expiresAt = Math.floor(Date.now() / 1000) + expiresIn;
```
stevelikesmusic, line 216: Is this expiry consistent with our other Oauth methods?

packages/cli/src/dbt/targets/Databricks/oauth.ts
```diff
@@ -0,0 +1,269 @@
...
+    // If that fails, try random port
+    const preferredPort = 8020;
+    await new Promise<void>((resolve, reject) => {
+        server.on('error', (err: NodeJS.ErrnoException) => {
+            if (err.code === 'EADDRINUSE') {
+                // Port 8020 is busy, try random port
+                GlobalState.debug(
+                    `> Port ${preferredPort} is busy, using random port`,
+                );
+                server.listen(0, () => {
```
stevelikesmusic: I remember you having a different solution if the port was busy that cycles through a list of other ports if the default port is busy. Is this a different case?

packages/cli/src/handlers/dbt/getWarehouseClient.ts
stevelikesmusic: So what's the edge case here? Why would the auth fail for M2M when there's client_id and secret? Is this related to a setting the customer makes in databricks?

packages/cli/src/handlers/dbt/getWarehouseClient.ts
```diff
@@ -309,12 +311,53 @@ export default async function getWarehouseClient(
             GlobalState.debug(
                 `> Exchanging Databricks OAuth credentials for access token`,
             );
-            const { accessToken } = await exchangeDatabricksOAuthCredentials(
+            try {
+                const { accessToken } =
+                    await exchangeDatabricksOAuthCredentials(
+                        credentials.serverHostName,
+                        credentials.oauthClientId,
+                        credentials.oauthClientSecret,
```
stevelikesmusic, line 319: Are these two values null for U2M?

stevelikesmusic (approved):
Very cool! Nothing blocking, but I do have a few questions:  
  
This adds a bit more complexity to the Oauth flow. Would it be worth having some kind of tests and a CLAUDE.md file for Oauth? I would be worried about creating regressions to Oauth if I had to make changes to it.   
  
Is the Databricks Oauth that different from our other Oauth flows that its own version of the Oauth flow pieces?   
  
I'm surprised the the databricks SDK isn't persiting the token itself.

packages/cli/src/handlers/dbt/getWarehouseClient.ts
rephus: Yes and No . 

This is explained here https://www.notion.so/lightdash/Databricks-Oauth-2aaa63207a7a804bb4c9fbe9ae845b53?source=copy_link#2aba63207a7a806e8e5bd600f9299f63

U2M can have no client, client but no secret AND client and secret. So it depends on your configuration

packages/cli/src/dbt/targets/Databricks/oauth.ts
rephus: actually, databrics can't use a random port, I'll remove this part of the code

packages/cli/src/dbt/targets/databricks.ts
rephus: nope. token, is just toke authentication, like PAT

oauth includes both m2m and u2m, and there is no way of telling which one is

packages/cli/src/dbt/targets/Databricks/oauth.ts
rephus: yes, if we add more methods. I don't think we should add this premature optimization right now. And I'd like to keep this separate from the oauth server on lightdash.

packages/cli/src/dbt/targets/Databricks/oauth.ts
rephus: yes, extracted into a separate function, this was also used on lightdash oauth server, so we can reuse it already

Issue #18101, opened 2025-11-13
Backend pivoted queries should respect LIGHTDASH_PIVOT_TABLE_MAX_COLUMN_LIMIT and show warning


almeidabbm:
Currently backend pivot queries respect a limit of `100` columns.
This came in from the previous implementation of the SQL Runner which shows a warning whenever there are more columns than displayed.

Both `SQL Runner` and `Explorer` BE pivot queries should respect `LIGHTDASH_PIVOT_TABLE_MAX_COLUMN_LIMIT` and show a warning when results are limited.

The default for this value should also be increased from `60` to `100` to respect the current limits of the `SQL Runner`

PR #18112, opened 2025-11-13
fix: respect max column limit config for backend pivoted results

almeidabbm:
Closes: #18101

### Description:
Increases the default pivot table column limit from 60 to 200 and makes the limit configurable throughout the application. The PR:

- Increases the default `LIGHTDASH_PIVOT_TABLE_MAX_COLUMN_LIMIT` from 60 to 200
- Removes the hardcoded `MAX_PIVOT_COLUMN_LIMIT` constant
- Passes the configured column limit to `PivotQueryBuilder.toSql()` in all relevant services
- Updates the `PivotQueryBuilder` to conditionally apply column limiting only when a limit is provided
- Updates the frontend to use the dynamic column limit from the health endpoint
- Updates documentation to reflect the configurable nature of the column limit

This change allows users to work with larger pivot tables while maintaining the ability to limit columns for performance reasons.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
@@ -1,6 +1,5 @@
 import {
     getAggregatedField,
-    MAX_PIVOT_COLUMN_LIMIT,
     normalizeIndexColumns,
     ParameterError,
     type PivotConfiguration,
@@ -73,9 +72,10 @@ export class PivotQueryBuilder {
      */
     private static calculateMaxColumnsPerValueColumn(
         valuesColumns: PivotConfiguration['valuesColumns'],
+        columnLimit: number,
     ): number {
         const valueColumnsCount = valuesColumns?.length || 1;
-        const remainingColumns = MAX_PIVOT_COLUMN_LIMIT - 1; // Account for the index column
+        const remainingColumns = columnLimit - 1; // Account for the index column
         return Math.floor(remainingColumns / valueColumnsCount);
     }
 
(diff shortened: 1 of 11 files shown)


almeidabbm:
* **#18113** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18113?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18112** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18112?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18112?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
```diff
@@ -527,8 +528,19 @@ export class PivotQueryBuilder {
...
+        let maxColumnsPerValueColumnSql = '';
+
+        if (columnLimit) {
+            const maxColumnsPerValueColumn =
+                PivotQueryBuilder.calculateMaxColumnsPerValueColumn(
+                    valuesColumns,
+                    columnLimit,
+                );
+
+            // Keep trailing space to avoid SQL syntax errors
```
magnew: Leading space, yea? Looks fine

almeidabbm: >❓ We decided the limit will apply to bars too? Which is what happens

About this, yep, this will make it so that bars are also limited, charts with a lot of series were slow and we were already limiting in the sql runner, this will make this consistent across charts and also configurable, currently we're stuck at 100 when `USE_SQL_PIVOT_RESULTS` is enabled

PR #18125, opened 2025-11-14
feat: add image dimension and sizing controls

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-172/adjust-image-on-table-size-on-yml

### Fit

- cover (default)

image

- none

image

- Fill /contain (will depend on the size)

image

### Description:
Added support for configuring image dimensions and object-fit properties in table cells. This enhancement allows users to specify width, height, and fit options for images displayed in tables.

The PR adds:
- New schema properties for image configuration (width, height, objectFit)
- Updated types to support these new properties
- Enhanced ImageCell component to respect the configured dimensions and fit options
- Added example in the jaffle-shop demo showing how to use the new image sizing options

This makes image display more flexible, allowing for consistent sizing and proper image fitting within table cells.

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/models/events.yml
+++ b/examples/full-jaffle-shop-demo/dbt/models/events.yml
@@ -87,6 +87,17 @@ models:
             groups: [ 'image' ]
             image: 
                 url: "${value.raw}" 
+          additional_dimensions:
+            image_with_size:
+              groups: [ 'image' ]
+              description: "Image URL with size"
+              type: string
+              sql: ${image_url}
+              image:
+                url: "${value.raw}" 
+                width: 128 # squared, do not preserve aspect ratio
+                height: 128
+                fit: cover
       - name: event_id
         description: ""
         meta:
@@ -108,7 +119,8 @@ models:
               image:
                 # Template combines value (the URL from SQL) with other row fields
                 # Use liquidjs to combine the URL with the event name and upcase
-                url: "https://${value.raw}-${row.events.event.raw | upcase}" 
+                url: "https://${value.raw}-${row.events.event.raw | upcase}"
+
             image_invalid_protocol:
               groups: [ 'image' ]
               description: "Invalid URL protocol"
--- a/packages/frontend/src/components/common/Table/ImageCell.tsx
+++ b/packages/frontend/src/components/common/Table/ImageCell.tsx
@@ -29,6 +29,7 @@ export const BrokenImageCell = ({
 };
 
 export const ImageCell = ({
+    item,
     imageUrl,
 }: {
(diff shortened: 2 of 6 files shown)


rephus:
* **#18129** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18129?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18125** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18125?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/full-jaffle-shop-demo/dbt/models/events.yml
```diff
@@ -87,6 +87,17 @@ models:
...
+            image_with_size:
+              groups: [ 'image' ]
+              description: "Image URL with size"
+              type: string
+              sql: ${image_url}
+              image:
+                url: "${value.raw}" 
+                width: 128 # squared, do not preserve aspect ratio
+                height: 128
+                fit: cover
```
ZeRego, line 100: `objectFit` typo ?

packages/frontend/src/components/common/Table/ImageCell.tsx
```diff
@@ -40,6 +41,29 @@ export const ImageCell = ({
...
 
+    // Get dimensions and objectFit from item.image configuration
+    const width = item.image?.width;
+    const height = item.image?.height;
+
+    const size =
+        width || height
+            ? {
+                  width: width ? `${width}px` : 'auto',
+                  height: height ? `${height}px` : '32px',
```
ZeRego: should height default to 'auto' if width is defined to keep the ratio?

examples/full-jaffle-shop-demo/dbt/models/events.yml
rephus: No, this is intentional. On YML I think we can simplify and avoid the objectFit property. But if you prefer to be 1:1 to css properties, we can use objectFit instead.

ZeRego: I don't get it, everywhere in this PR we use `objectFit`. So how is this mapping from `fit` to `objectFit`?

rephus: ah, yes , my bad. I should rename also other definitions, and keep `objectFit` only on the css property. Great catch.

ZeRego (approved):

PR #18129, opened 2025-11-14
feat: add table column width support for image dimensions

rephus:
Closes: 

Before

Screenshot from 2025-11-14 10-17-37

After

Screenshot from 2025-11-14 11-18-49

### Description:
Added support for dynamic width images in table visualizations. This PR introduces a new `getImageSize` function that extracts width information from dimension items with image URLs and applies appropriate styling constraints to the column. The implementation includes:

1. A new example in the jaffle-shop demo with a dynamic width image field
2. Logic to set fixed width constraints for image columns with explicit width values
3. Proper type annotation for the row parameter in the accessor function

This enhancement allows for better control over image sizing in table visualizations, improving the overall display of image content.

### DIFF
--- a/packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
+++ b/packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
@@ -2,6 +2,7 @@ import {
     formatItemValue,
     getSubtotalKey,
     isCustomDimension,
+    isDimension,
     isField,
     type ItemsMap,
     type ParametersValuesMap,
@@ -75,6 +76,23 @@ export function getSubtotalValueFromGroup(
     return subtotal?.[columnId];
 }
 
+const getImageSize = (item: ItemsMap[string] | undefined) => {
+    if (isDimension(item) && item.image?.url) {
+        const defaultWidth = 100;
+        const defaultPadding = 8 * 2;
+        const width = (item.image?.width || defaultWidth) + defaultPadding;
+
+        return {
+            style: {
+                width,
+                minWidth: width,
+                maxWidth: width,
+            },
+        };
+    }
+    return {};
+};
+
 const getDataAndColumns = ({
     itemsMap,
     selectedItemIds,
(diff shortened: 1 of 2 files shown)


rephus:
* **#18129** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18129?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#18125** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
```diff
@@ -75,6 +76,24 @@ export function getSubtotalValueFromGroup(
     return subtotal?.[columnId];
 }
 
+const getImageSize = (item: ItemsMap[string] | undefined) => {
+    if (isDimension(item) && item.image?.url) {
+        const defaultWidth = 100;
+        const defaultPadding = 8 * 2;
+        const width = (item.image?.width || defaultWidth) + defaultPadding;
```
rephus, line 83: if width is specified, we adjust the column to that image size. Otherwise we will apply a default width of 100 (default image has a fixed height, so we can't say if this is enough) 

We might weant to change these default values, based on feedback. 

Or switch to a default fixed width, rather than height , if table virtualization is not a big issue.

rephus:
Ideally, we should be able to deleate thiscolumn size to table or tanstack, but I couldn't find where the size is dynamically calculated. We should replace the underlying URL with a box of the image size. 

Somewhere in `TableBody.tsx`  or `useTableDataModel`

We might need to change this as well if we want to improve the table virtualization

Issue #18120, opened 2025-11-13
Min Max Range Conditional Formatting Not Displaying in Pivot Tables


jesshitchcock:
### Description

Conditional formatting with range-based rules (using "Min value in table" and "Max value in table") fails to display in pivot table visualizations in Explore.

<img width="1277" height="506" alt="Image" src="https://github.com/user-attachments/assets/096a7ad9-d6e8-4982-85a7-841e31f4a58a" />

### Steps to Reproduce the Bug or Issue

1. In Explore, select two dimensions and one metric
2. Execute the query
3. Open the chart configuration and select "Table" as the chart type
4. In the "General" tab, move one dimension to the columns section to create a pivot table
5. Navigate to "Conditional formatting" tab
6. Add a new rule and select "Range" as the type
7. Set "Min value type" to "Min value in table"
8. Set "Max value type" to "Max value in table"
9. Observe the results

NB: Note this is not reproducible if: `use-sql-pivot-results` flag is on so would be good to check if this is solved by new code changes not yet released. 

### version

v0.2165.1

### Cloud or self-hosting

cloud

PR #18133, opened 2025-11-14
fix: conditional formatting on pivoted tables with USE_SQL_PIVOT_RESULTS enabled

almeidabbm:
Closes: #18120

### Description:
Enhances conditional formatting to support backend-pivoted data by adding a function to find pivot columns that correspond to each metric. The implementation now correctly handles SQL pivot tables by collecting values from all relevant pivot columns rather than assuming a direct field match. This also adds additional validation to filter out NaN values and handle cases where no valid values are found.

**Config**

**Before**

**After**

### DIFF
--- a/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
+++ b/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
@@ -489,22 +489,49 @@ const useTableConfig = (
             return undefined;
         }
 
+        // For backend-pivoted data (SQL pivot), we need to find the pivot columns
+        // that correspond to each metric
+        const getColumnsForField = (fieldId: string): string[] => {
+            if (!resultsData.pivotDetails) {
+                // No backend pivot, use the field directly
+                return [fieldId];
+            }
+
+            // Find all pivot columns that reference this field
+            return resultsData.pivotDetails.valuesColumns
+                .filter((col) => col.referenceField === fieldId)
+                .map((col) => col.pivotColumnName);
+        };
+
         return Object.entries(itemsMap)
             .filter(([_, field]) => isNumericItem(field))
             .filter(([fieldId]) => fieldsNeedingMinMax.has(fieldId))
             .filter(([fieldId]) => isColumnVisible(fieldId))
-            .filter(([fieldId]) => fieldId in resultsData.rows[0])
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#18144** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18144?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18133** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18133?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18133?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
```diff
@@ -489,22 +489,49 @@ const useTableConfig = (
...
-                    .map((value) => Number(value))
-                    .map((value) => convertFormattedValue(value, field));
+                const columnsToCheck = getColumnsForField(fieldId);
+
+                if (columnsToCheck.length === 0) {
+                    return acc;
+                }
+
+                // Collect values from all columns that correspond to this field
+                const columnValues = resultsData.rows.flatMap((row) =>
```
magnew, line 518: I know this was copied over, but 3 maps and 2 filters here means going through these columns a lot more times than we need to. And it could be a lot of columns. These things are on my mind since that last milestone 😵‍💫 

This is all in a reduce and we are spreading the accumulator too 😬 

This could be a follow up, but this whole function could be a lot faster. I noticed that changing the conditional formatting is really slow. It might fix it.

magnew (commented):
Code looks good ✅ 
Tested and works great ✅ 

Left a comment that's kinda unrelated, but I had already written it out. We could address separately.

packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
almeidabbm: That's a very good point, having a look 🙇

almeidabbm: The previous approach already looped over rows so this shouldn't make it worse, will see if improving this actually makes conditional formatting better, but from what I've seen so far calculating the min max map is not the point of failure 😢

Issue #4437, opened 2023-02-10
I would like a progress bar / gauge chart


owlas:
### Description

These can be much more engaging for users compared to a big number. Especially if there's a goal we're aiming for

PR #18150, opened 2025-11-14
feat: add gauge chart visualization type

ZeRego:
Closes: #4437

### Description:
Adds a new gauge chart type to the visualization options. This chart type displays a single metric value on a gauge with customizable min/max values and appearance options.

Known limitations/following features:
- cannot disable "hover/pointer" behaviour
- no tooltip
- customise inner label
- customise section labels/legend
- customise formatting
- reference lines?
- no options to see underlying data, drill by, filter by 

CleanShot 2025-11-18 at 09 11 48@2x

CleanShot 2025-11-18 at 09 13 10@2x

### DIFF
--- a/packages/frontend/src/components/SimpleGauge/index.tsx
+++ b/packages/frontend/src/components/SimpleGauge/index.tsx
@@ -89,7 +89,7 @@ const calculateRadius = (boundWidth: number, boundHeight: number) => {
 const EchartOptions: Opts = { renderer: 'svg' };
 
 const SimpleGauge: FC<SimpleGaugeProps> = memo((props) => {
-    const { chartRef, isLoading, resultsData } = useVisualizationContext();
+    const { chartRef, isLoading } = useVisualizationContext();
     const [chartWidth, setChartWidth] = useState(0);
     const [chartHeight, setChartHeight] = useState(0);
     const sizes = useMemo(() => {
@@ -131,11 +131,6 @@ const SimpleGauge: FC<SimpleGaugeProps> = memo((props) => {
     });
 
     useEffect(() => {
-        // Load all the rows
-        resultsData?.setFetchAll(true);
-    }, [resultsData]);
-
-    useEffect(() => {
         const listener = () => chartRef.current?.getEchartsInstance().resize();
         const observer = new ResizeObserver(([entry]) => {
             const { width, height } = entry.contentRect;
@@ -146,7 +141,7 @@ const SimpleGauge: FC<SimpleGaugeProps> = memo((props) => {
         if (chartRef.current?.getEchartsInstance().getDom()) {
             observer.observe(chartRef.current?.getEchartsInstance().getDom());
         }
-
+        window.addEventListener('resize', listener);
         return () => {
             window.removeEventListener('resize', listener);
             observer.disconnect();


ZeRego:
* **#18150** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18150?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/database/migrations/20241114000000_add_gauge_to_chart_type.ts
```diff
@@ -0,0 +1,10 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex('chart_types').insert({ chart_type: 'gauge' });
```
stevelikesmusic, line 4: Any reason why is chart_types a table instead of a Postgres enum or a TS enum?

packages/common/src/types/savedCharts.ts
```diff
@@ -23,6 +23,7 @@ export enum ChartKind {
     FUNNEL = 'funnel',
```
stevelikesmusic, line 23: <p class='graphite__hidden'><i>[Re: line +13]</i></p>
What's the difference between ChartType and ChartKind? 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18150?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/frontend/src/components/LightdashVisualization/VisualizationConfigGauge.tsx
```diff
@@ -0,0 +1,63 @@
...
+        }, {});
+        return { ...metrics, ...numericTableCalculations };
+    }, [itemsMap]);
+
+    const gaugeConfig = useGaugeChartConfig(
+        initialChartConfig as any,
+        itemsMap,
+    );
+
+    useEffect(() => {
```
stevelikesmusic, line 42: What's the purpose of this useEffect?

packages/frontend/src/components/SimpleGauge/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+        };
+    }, [chartWidth, chartHeight]);
+    const gaugeOptions = useEchartsGaugeConfig({
+        isInDashboard: props.isInDashboard,
+        ...sizes,
+    });
+
+    useEffect(() => {
+        // Load all the rows
+        resultsData?.setFetchAll(true);
```
stevelikesmusic: What does this call do?

packages/frontend/src/components/SimpleGauge/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+            setChartWidth(width);
+            setChartHeight(height);
+        });
+
+        if (chartRef.current?.getEchartsInstance().getDom()) {
+            observer.observe(chartRef.current?.getEchartsInstance().getDom());
+        }
+
+        return () => {
+            window.removeEventListener('resize', listener);
```
stevelikesmusic, line 146: Where do we add this event listener?

packages/frontend/src/components/SimpleGauge/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+        const availableHeightForFontSizeCalculation = chartHeight ?? 0;
+
+        const boundHeight = clamp(
+            availableHeightForFontSizeCalculation,
+            BOX_MIN_HEIGHT,
+            BOX_MAX_HEIGHT,
+        );
+
+        return {
+            tileFontSize: calculateFontSize(
```
stevelikesmusic, line 107: Is this still too large for large values? The example with percentages and decimals seems to overflow the gauge display?

packages/backend/src/database/migrations/20241114000000_add_gauge_to_chart_type.ts
ZeRego: Didn't know Postgres had enums. Seems like the con of using them is that you cannot update/remove values from it. So rollbacks would be very difficult.
We also use a TS enum, but this guarantees correct values at run time.

packages/common/src/types/savedCharts.ts
ZeRego: At the start, we only had ChartType but as the CARTESIAN type logic grew and included a lot of subtypes ("bars, lines, scatter, etc"), we found the need to have a more granular type to help display correct icons and such.

packages/frontend/src/components/LightdashVisualization/VisualizationConfigGauge.tsx
ZeRego: Updates the provider. Follows the pattern from the other chart hooks.

packages/frontend/src/components/SimpleGauge/index.tsx
ZeRego: Fetches all the results pages. Perhaps not needed here, will check in the following commit.

packages/frontend/src/components/SimpleGauge/index.tsx
ZeRego: It seems to have a good balance of space and size. In cases where the number is too big and overlapping the bar, users can update the metric formatting to be in the thousands, or round to 0, etc.

packages/frontend/src/components/SimpleGauge/index.tsx
ZeRego: I must have deleted it by mistake. nice catch! fixing now

packages/backend/src/database/migrations/20241114000000_add_gauge_to_chart_type.ts
stevelikesmusic: I've always understood enums like a FK. It's flexible to add and rename values with [ALTER TYPE ](https://www.postgresql.org/docs/current/sql-altertype.html). Removing is tricky, but similar to removing a FK. 

The table is definitely more flexible if we have to update this whenever we add charts. Have we removed charts from being supported in Lightdash?

PR #18191, opened 2025-11-18
feat: make browser oauth default for login 

owlas:
make oauth login default, it's the best

### DIFF
--- a/packages/cli/src/handlers/login.ts
+++ b/packages/cli/src/handlers/login.ts
@@ -1,5 +1,4 @@
 import { AuthorizationError } from '@lightdash/common';
-import inquirer from 'inquirer';
 import fetch from 'node-fetch';
 import { URL } from 'url';
 import { LightdashAnalytics } from '../analytics/analytics';
@@ -7,7 +6,6 @@ import { configFilePath, setContext, setDefaultUser } from '../config';
 import GlobalState from '../globalState';
 import * as styles from '../styles';
 import { checkLightdashVersion } from './dbt/apiClient';
-import { generatePersonalAccessToken } from './login/pat';
 import { loginWithOauth } from './oauthLogin';
 import { setFirstProject, setProjectCommand } from './setProject';
 import { buildRequestHeaders } from './utils';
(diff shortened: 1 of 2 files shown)


owlas:
* **#18192** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18192?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18191** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18191?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/cli/src/handlers/login.ts
```diff
@@ -144,12 +77,10 @@ export const login = async (url: string, options: LoginOptions) => {
     }
 
     let loginResult;
-    if (options.oauth) {
-        loginResult = await loginWithOauth(url);
-    } else if (options.token) {
+    if (options.token) {
         loginResult = await loginWithToken(url, options.token);
     } else {
-        loginResult = await loginWithPassword(url);
```
rephus, line 152: are you intentionally removing the password auth ? not just defaulting to oauth. Some users might still need this on interactive docker/headless environments. 

I'd rather switch the default, but keep password with a new --password flag.

rephus (approved):
Oauth works great, this is what I use every day. However, I find annoying the PAT expires in 8 hours . We  should add a new argument to change this default behaviour, can be done on a separate PR or when a customer asks for it

rephus:
Your first (and only) commit didn't use the right commit message prefix format, that is why this test fails:

> Validate PR Title / main (pull_request_target)

packages/cli/src/handlers/login.ts
owlas: it's not possible to use headless because we asked for the password interactively not via a flag

Issue #18195, opened 2025-11-18
Allow spaces created via CLI to be public by default


rephus:
## Problem
Currently, when uploading charts to a new space via the CLI (e.g., `lightdash upload`), the created space is private by default. This causes friction for teams deploying content across multiple instances, as they must manually make spaces public after creation, often via API calls or scripts. This is especially problematic in CI/CD workflows where spaces are not created via the UI.

## Proposed Solution
- Add a `--public` flag (or similar) to the CLI upload command, allowing users to specify that any new spaces created during the upload should be public by default.
- Alternatively, provide a deployment configuration option to set the default visibility of newly created spaces.

## Why it's needed
- Reduces manual work and potential for error when deploying content to multiple instances.
- Supports workflows where most spaces should be public, with only a few exceptions (e.g., HR or financial data).
- Improves the developer experience and aligns with user expectations for content visibility.

## Acceptance Criteria
- When using the CLI to upload charts, users can specify that new spaces should be public.
- Spaces created via CI/CD pipelines can be public by default if configured.
- Documentation is updated to reflect the new flag/option.

## Labels
- Customer Support
- ✨ feature-request
- ⚙️ backend
- Ops⚙️

## Milestone
- 💜 Customer Support

PR #18200, opened 2025-11-18
feat: add public space creation option for charts and dashboards

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-190/allow-spaces-created-via-cli-to-be-public-by-default

Before:

Screenshot from 2025-11-18 16-27-06

After:
Screenshot from 2025-11-18 16-30-09

Screenshot from 2025-11-18 16-30-25

### Description:
Added support for creating public spaces when uploading charts and dashboards via the CLI. This PR introduces a new `publicSpaceCreate` parameter to the backend API and a corresponding `--public` flag to the CLI upload command.

The changes include:
1. Added a new `publicSpaceCreate` parameter to the chart and dashboard creation endpoints
2. Updated the space creation logic to respect the public/private setting
3. Added a new `--public` flag to the CLI upload command
4. Removed unused properties from the GaugeChart model (gapSectionColor, showProgress)
5. Fixed the order of success/error enums in the generated routes

This feature allows users to specify whether newly created spaces should be public or private when uploading content via the CLI.

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.ts
+++ b/packages/backend/src/services/CoderService/CoderService.ts
@@ -683,6 +683,7 @@ export class CoderService extends BaseService {
         slug: string,
         chartAsCode: ChartAsCode,
         skipSpaceCreate?: boolean,
+        publicSpaceCreate?: boolean,
     ) {
         const project = await this.projectModel.get(projectUuid);
 
@@ -713,6 +714,7 @@ export class CoderService extends BaseService {
                     chartAsCode.spaceSlug,
                     user,
                     skipSpaceCreate,
+                    publicSpaceCreate,
                 );
 
             console.info(
@@ -856,6 +858,7 @@ export class CoderService extends BaseService {
         spaceSlug: string,
         user: SessionUser,
         skipSpaceCreate?: boolean,
+        publicSpaceCreate?: boolean,
     ): Promise<{ space: Omit<SpaceSummary, 'userAccess'>; created: boolean }> {
         const [existingSpace] = await this.spaceModel.find({
             path: getLtreePathFromContentAsCodePath(spaceSlug),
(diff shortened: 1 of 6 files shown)


rephus:
* **#18200** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18200?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -914,7 +917,9 @@ export class CoderService extends BaseService {
 
             const newSpace = await this.spaceModel.createSpace(
                 {
-                    isPrivate: closestAncestorSpace?.isPrivate ?? true,
+                    isPrivate:
+                        closestAncestorSpace?.isPrivate ??
+                        publicSpaceCreate !== true,
```
stevelikesmusic, line 922: I'm not sure I understand the intent with --public. If the user passes the public flag, how do we determine the space to use or create? Should we return a user-facing message if we get the --public flag for a private space? Should we prevent the operation altogether to get confirmation that this operation is what the user really wants (creating public content in a private space).

rephus: at the moment, by default, when a new space doesn't exist, we create them as private, this is blocking some customers because they can't share the content on a brand new instance. 

With this flag, we change that default behaviour, and create those spaces as public by default. 

Does that answer your question ?

ZeRego (approved):

rephus:
unrelated e2e error 
image

stevelikesmusic:
Closes: #18195

PR #18117, opened 2025-11-13
chore: separate deps image from docker build

stevelikesmusic:
Closes: CENG-139

### Description:

Break up the Dockerfile to create a base image with dbt and other dependencies pre-built. 

test-frontend test-backend

PR #18201, opened 2025-11-18
chore: parallelize docker PR builds

stevelikesmusic:
Closes: 

### Description:

Optimize PR builds to run in parallel stages where we can. If this is successful, we can roll it up to the main dockerfile—and even consolidate.

It looks like the mount caches added in this PR are successfully bring down build times on cache hits. We should monitor build times to see if further optimizations are neccessary or not. 

test-frontend test-backend

### DIFF
--- a/dockerfile-prs
+++ b/dockerfile-prs
@@ -1,7 +1,15 @@
-# -----------------------------
+# syntax=docker/dockerfile:1.7
+
+# Optimized Multi-Stage Build for PR Previews
+# Uses parallel build stages and cache mounts for maximum caching
+# Okteto uses a cluster of buildkit instances to persist caches across PRs
+# New builds when the cluster scales will have a cold cache, but subsequent
+# builds will warm the cache again.
+
+# ------------------------------
 # Stage 0: install dependencies
-# -----------------------------
-FROM node:20-bookworm-slim AS base
+# ------------------------------
+FROM node:20-bookworm-slim AS pnpm-base
 
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
@@ -11,6 +19,8 @@ RUN corepack prepare pnpm@9.15.5 --activate
 RUN pnpm config set store-dir /pnpm/store
 
 WORKDIR /usr/app
+
+FROM pnpm-base AS base
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
@@ -37,7 +47,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Installing multiple versions of dbt
 # dbt 1.4 is the default
-RUN python3 -m venv /usr/local/dbt1.4 \
+# Use pip cache to speed up subsequent builds
+RUN --mount=type=cache,target=/root/.cache/pip \
+    python3 -m venv /usr/local/dbt1.4 \
     && /usr/local/dbt1.4/bin/pip install \
     "dbt-postgres~=1.4.0" \
     "dbt-redshift~=1.4.0" \
@@ -48,7 +60,8 @@ RUN python3 -m venv /usr/local/dbt1.4 \
     "dbt-clickhouse~=1.4.0" \
     "psycopg2-binary==2.9.6"
 
-RUN ln -s /usr/local/dbt1.4/bin/dbt /usr/local/bin/dbt\
+RUN --mount=type=cache,target=/root/.cache/pip \
+    ln -s /usr/local/dbt1.4/bin/dbt /usr/local/bin/dbt\
     && python3 -m venv /usr/local/dbt1.5 \
     && /usr/local/dbt1.5/bin/pip install \
     "dbt-postgres~=1.5.0" \
@@ -104,7 +117,7 @@ RUN ln -s /usr/local/dbt1.4/bin/dbt /usr/local/bin/dbt\
     "dbt-databricks~=1.9.0" \
     "dbt-trino~=1.9.0" \
     "dbt-clickhouse~=1.9.0" \
-    && ln -s /usr/local/dbt1.9/bin/dbt /usr/local/bin/dbt1.9 \ 
+    && ln -s /usr/local/dbt1.9/bin/dbt /usr/local/bin/dbt1.9 \
     && python3 -m venv /usr/local/dbt1.10 \
     && /usr/local/dbt1.10/bin/pip install \
     "dbt-core~=1.10.0" \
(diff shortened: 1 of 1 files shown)


stevelikesmusic:
* **#18201** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18201?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#18117](https://github.com/lightdash/lightdash/pull/18117) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18117?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#18209](https://github.com/lightdash/lightdash/pull/18209) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18209?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
This PR built in 23m 
image
 
 Compared to another PR without optimization, where it built in 22m
  
image

dockerfile-prs
```diff
@@ -135,52 +142,58 @@ EXPOSE 8080
...
 COPY .pnpmfile.cjs .
-COPY packages/common/package.json ./packages/common/
-COPY packages/warehouses/package.json ./packages/warehouses/
-COPY packages/backend/package.json ./packages/backend/
-COPY packages/frontend/package.json ./packages/frontend/
 
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
-    pnpm install --frozen-lockfile --prefer-offline
+COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
+COPY packages/common/package.json packages/common/
```
rephus, line 156: package.json is pretty much going to change on every PR (and when pulling master) becuase of the version

dockerfile-prs
```diff
@@ -235,23 +287,15 @@ RUN rm -rf node_modules \
 
 # Install production dependencies
 ENV NODE_ENV production
+
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
-    pnpm install --prod --frozen-lockfile --prefer-offline
+    pnpm fetch --prod && pnpm install --prod --frozen-lockfile --offline
```
rephus: > Fetch packages from a lockfile into virtual store, package manifest is ignored. WARNING! This is an experimental command. Breaking changes may be introduced in non-major versions of the CLI

Should we use experimental ? what's the advantage of using fetch over install from the lockfile ?

dockerfile-prs
```diff
@@ -1,7 +1,12 @@
+# syntax=docker/dockerfile:1.7
```
rephus, line 1: is it possible github actions by default is not using this buildkit ? 

https://docs.docker.com/build/ci/github-actions/configure-builder/

rephus (commented):
The changes make sense, but at the moment I don't see any advantage for this. If there are no improvements on the build time, I'd suggest to keep it as it is, to avoid potential issues.

dockerfile-prs
stevelikesmusic: That's true. I'm not sure how we can get around that one. At the least though, using cache mounts for the dbt installs seem to have drastically cut things down with okteto.  

[This build](https://github.com/lightdash/lightdash/runs/55797473782) in the subsequent PR used the warmed cache from here and skipped right over the dbt installs.

dockerfile-prs
stevelikesmusic: I don't think so. Okteto is running the build outside of GH and those nodes are dedicated to buildkit. This directive helps when running locally. 

I do think we need to explicitly enable buildkit for GH actions—which we do for the `main` build. 

After talking with Ramiro, I think we have a working solution using their buildkit cluster.

dockerfile-prs
stevelikesmusic: Good call—where do you see that warning? 

We're already fetching to hydrate the cache mount. I'm dropping the additional fetch here so that we simply get packages from the warmed cache.

rephus: > where do you see that warning

This is not a warning, this is the `pnpm fetch --help` , where it sais this is experimental.

rephus:
keep in mind doing `chore` will not build a new image . So you will have to wait until someone merges a `fix` or `feat` on main to see if this is working

stevelikesmusic:
of course 😅 I'll amend the commit message

dockerfile-prs
```diff
@@ -135,52 +148,61 @@ EXPOSE 8080
...
 
-# Build backend
-COPY packages/backend/tsconfig.json ./packages/backend/
-COPY packages/backend/tsconfig.sentry.json ./packages/backend/
-COPY packages/backend/src/ ./packages/backend/src
+COPY packages/backend/tsconfig*.json packages/backend/
+COPY packages/backend/src/ packages/backend/src/
 
-# Conditionally build backend with sourcemaps if Sentry environment variables are set
+ARG SENTRY_AUTH_TOKEN=""
```
rephus, line 197: why do we need to specify this sentry args on all the layers ?

dockerfile-prs
```diff
@@ -37,7 +47,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # Installing multiple versions of dbt
 # dbt 1.4 is the default
-RUN python3 -m venv /usr/local/dbt1.4 \
+# Use pip cache to speed up subsequent builds
+RUN --mount=type=cache,target=/root/.cache/pip \
```
rephus, line 51: I'd say we do not need the dbt dependencies for building, only on run time, so we could also paralalize this step. But we can do it on a separate PR , just to keep this incremental and easily rollbackable

dockerfile-prs
stevelikesmusic: I see. I think pnpm fetch is not even needed. We're caching with BuildKit's cache mount. So we can probably drop the fetch for pnpm's virtual cache.

dockerfile-prs
stevelikesmusic: I'll consolidate those in the base layer

stevelikesmusic: Actually reading more, we need to redeclare ARG instructions in each build stage. We didn't need to before because we had a single build stage.

dockerfile-prs
stevelikesmusic: Agreed

stevelikesmusic:
### Merge activity

* **Nov 19, 2:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/18201).
* **Nov 19, 2:45 PM UTC**: @stevelikesmusic merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/18201).

PR #18227, opened 2025-11-19
chore: use runs endpoint for schedulers run list

magnew:
### Description:

Uses the new `runs` endpoint to render the list of scheduler runs. Then the `runs/:runId/logs` endpoint to get details. This improves the fetching semantics and stability of the schedulers run history. Before we were fetching all logs and filtering and grouping on the FE. Now we fetch a paginated list of scheduler runs which include a computed status at the top level. When a user clicks to expand a run, we fetch its details. 

Improves performance by:
- Fetching a paginated list of runs initially, instead of all logs
- Relying on filtering and computed status from the BE

Also updates the UI a bit:
- Fixes incorrect/confusing run sort order
- Lists the jobs in a run in chronological order

Test by looking at the Run History. Runs will be listed and can be expanded to see child jobs. 

Screenshot 2025-11-19 at 18 54 11

PR #18206, opened 2025-11-18
chore: unify + improve SD run history api

magnew:
### Description:

Improves the schedulers list api by making it possible to fetch latest run information in the same call as the schedulers. Before we were having to make a separate call for last run status, now it is (optionally) included in the list. 

Specifically:
- adds a flag (`includeLatestRun`) to the schedulers `/list` endpoint. This makes it possible to fetch the list with latest runs for each scheduler without making another call + query to get the run information
- adds a schedulerUuids filter to `getSchedulerForProject` so that we can get information about specific schedulers instead of loading them all into memory when we process `/runs`
- adds a `schedulerUuids` filter to `getSchedulerRuns` so that we can filter by multiple schedulers

### **/list** payload now looks like

```
"data": [
    {
        "schedulerUuid": "d8dc574f-7b33-4bd7-9b5a-6c7d1d3d5ee5",
        "name": "Meow - image",
        "message": "",
        "createdAt": "2025-11-18T12:24:24.809Z",
        "updatedAt": "2025-11-18T18:09:40.170Z",
        "createdBy": "b264d83a-9000-426a-85ec-3f9c20f368ce",
        "createdByName": "David Attenborough",
        "cron": "11 * * * *",
        "savedChartUuid": null,
        "savedChartName": null,
        "dashboardUuid": "f4234f20-b9d4-4c18-9ff9-96dbc5dd184d",
        "dashboardName": "A dashboard",
        "format": "image",
        "options": {
            "withPdf": false
        },
        "filters": [],
        "parameters": null,
        "customViewportWidth": null,
        "thresholds": [],
        "enabled": true,
        "notificationFrequency": null,
        "selectedTabs": null,
        "includeLinks": true,
        "targets": [
            {
                "schedulerSlackTargetUuid": "7da23fee-3e44-4592-9a38-5c6ad1f0bf4e",
                "createdAt": "2025-11-18T18:09:40.170Z",
                "updatedAt": "2025-11-18T18:09:40.172Z",
                "schedulerUuid": "d8dc574f-7b33-4bd7-9b5a-6c7d1d3d5ee5",
                "channel": "U05F97EE6BV"
            },
            {
                "schedulerSlackTargetUuid": "f67b8faa-f020-492f-80e5-45e208e8c15e",
                "createdAt": "2025-11-18T18:09:40.170Z",
                "updatedAt": "2025-11-18T18:09:40.172Z",
[...]

PR #18139, opened 2025-11-14
feat: show latest SD run status

magnew:
### Description:

Show the last run status for each scheduler on the main scheduler list.

Screenshot 2025-11-17 at 20 28 50

PR #18228, opened 2025-11-19
fix: remove uneeded database calls for schedulers

magnew:
### Description:

Remove unneeded DB calls
- We were fetching a row to check loading -> moved down
- We were fetching a scheduler for each delete dialog.

### DIFF
--- a/packages/frontend/src/components/SchedulersView/LogsTable.tsx
+++ b/packages/frontend/src/components/SchedulersView/LogsTable.tsx
@@ -53,7 +53,9 @@ import {
 } from '../../features/scheduler/hooks/useScheduler';
 import useHealth from '../../hooks/health/useHealth';
 import { useGetSlack } from '../../hooks/slack/useSlack';
+import LoadingState from '../common/LoadingState';
 import MantineIcon from '../common/MantineIcon';
+import ResourceEmptyState from '../common/ResourceView/ResourceEmptyState';
 import { LogsTopToolbar } from './LogsTopToolbar';
 import {
     formatTaskName,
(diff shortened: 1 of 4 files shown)


magnew:
* **#18287** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18287?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18174** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18228** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18228?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#18227** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18139** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18206** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

owlas (commented):
investigating an infinite loop when fetching runs

packages/frontend/src/components/SchedulersView/LogsTable.tsx
pimmey, line 191: Pretty sure this is causing that infinite loop Oliver mentioned. Maybe we should just run it once, with empty deps array. But it's not even part of this PR 😄

packages/frontend/src/components/SchedulersView/LogsTable.tsx
```diff
@@ -206,42 +208,47 @@ const LogsTable: FC<LogsTableProps> = ({ projectUuid }) => {
...
+            // Fetch logs for newly expanded rows
+            newlyExpandedRowIds.forEach((rowId) => {
+                const rowIndex = parseInt(rowId, 10);
+                const run = schedulerRunsData?.[rowIndex];
+
+                if (run && !childLogsMap.has(run.runId)) {
+                    void fetchRunLogsMutation
+                        .mutateAsync(run.runId)
+                        .then((childLogs) => {
+                            setChildLogsMap((prev) => {
```
pimmey, line 238: First I thought maybe this is spinning something infinitely, cause we're setting the child logs map to a `new Map`, but it's also a dependency of this callback 🤔 but it seems like we don't call it downstream. `MantineReactTable` just stores the reference and doesn't react to it changing, but it's a fragile one 😄 maybe something to keep an eye for.

magnew: Good call. Im actually going to change it substantially in a follow up.

packages/frontend/src/components/SchedulersView/LogsTable.tsx
magnew: But I can't make it happen 😑

PR #18258, opened 2025-11-20
feat: group SD connection errors, improve message

magnew:
### Description:

Users currently see a cryptic call stack when scheduled deliveries fail because the headless browser crashes. We were already adding a more user-friendly message in some cases, but this adds a few more cases and makes the error message a bit more clear for most people. 

This is still not an especially useful error message and I am open to suggestions for improvement. Generally there is not much a user can do about this kind of error, so it doesn't have any call to action. Ideally very few people will see this.

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -1142,13 +1142,19 @@ export class UnfurlService extends BaseService {
                         code: 2, // Error
                     });
 
+                    // All of these errors are related to connection issues and usually
+                    // arise because the headless browser either crashed while handling the request
+                    // or crash before and can't be reached. We are grouping them togther here
+                    // for a user-friendly error message, but we keep the original error message in the logs.
                     const isConnectionError =
-                        errorMessage.includes('ECONNREFUSED');
+                        errorMessage.includes('ECONNREFUSED') ||
+                        errorMessage.includes('ENOTFOUND') ||
+                        errorMessage.includes('ECONNRESET');
 
                     throw new ScreenshotError(
                         `Screenshot ${errorType}: ${
                             isConnectionError
-                                ? 'There was a connection error while capturing the screenshot. Please contact your admin or support team.'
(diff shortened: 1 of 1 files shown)


magnew:
* **#18258** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18258?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -1142,13 +1142,19 @@ export class UnfurlService extends BaseService {
...
-                        errorMessage.includes('ECONNREFUSED');
+                        errorMessage.includes('ECONNREFUSED') ||
+                        errorMessage.includes('ENOTFOUND') ||
+                        errorMessage.includes('ECONNRESET');
 
                     throw new ScreenshotError(
                         `Screenshot ${errorType}: ${
                             isConnectionError
-                                ? 'There was a connection error while capturing the screenshot. Please contact your admin or support team.'
+                                ? 'There was a connection error while capturing the screenshot. This often indicates a heavy load on the screenshot service and the delivery cannot be completed at this time.'
```
owlas, line 1157: This feels as cryptic as the original message to an end user and leaves them with no way to solve it.

When we group these messages together (like the spinners, and the connection not available) can we capture these as a subset of fixed tags like:

- worker_unavailable
- content_error
- unexpected_error
- auth_error


Ideally in the app we could give a count of each type

owlas: to be clear i'm approving the copy change but I think the solution is that something that allows people (lightdash devs **and** the org admins) to see a clear breakdown of error types, so we can spot when there are resource problems on the workers

magnew: Agreed, this is a band aid, but it will keep more stack traces from bubbling up to the user-facing error toast

PR #18291, opened 2025-11-21
feat: allow switching agents on multi-agent slack channel

joaoviana:
Closes: #N/A

### Description:

Added the ability to switch AI agents within an existing Slack thread. Users can now select a different agent from a dropdown menu that appears at the bottom of agent responses. When switching agents:

- The dropdown groups agents by project when multiple projects are available
- Any running query is automatically cancelled
- A confirmation message replaces the dropdown after switching
- Analytics events track agent switches

This feature enhances flexibility by allowing users to change the AI agent they're interacting with without starting a new conversation thread.

**ignore the preview project there, that was before I pushed some changes**

### DIFF
--- a/packages/backend/src/ee/models/AiAgentModel.ts
+++ b/packages/backend/src/ee/models/AiAgentModel.ts
@@ -3334,6 +3334,30 @@ export class AiAgentModel {
             });
     }
 
+    async getLatestPromptInThread(threadUuid: string): Promise<string | null> {
+        const result = await this.database(AiPromptTableName)
+            .select('ai_prompt_uuid')
+            .where('ai_thread_uuid', threadUuid)
+            .orderBy('created_at', 'desc')
+            .first();
+
+        return result?.ai_prompt_uuid ?? null;
+    }
+
+    async updateThreadAgent({
+        threadUuid,
+        agentUuid,
+    }: {
+        threadUuid: string;
+        agentUuid: string;
+    }): Promise<void> {
+        await this.database(AiThreadTableName)
+            .where('ai_thread_uuid', threadUuid)
+            .update({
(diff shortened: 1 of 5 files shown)


joaoviana:
* **#18291** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/models/AiAgentModel.ts
```diff
@@ -3334,6 +3334,30 @@ export class AiAgentModel {
...
+        threadUuid,
+        agentUuid,
+    }: {
+        threadUuid: string;
+        agentUuid: string;
+    }): Promise<void> {
+        await this.database(AiThreadTableName)
+            .where('ai_thread_uuid', threadUuid)
+            .update({
+                agent_uuid: agentUuid,
```
tatianainama, line 3357: uhm, do you think there will be issues with updating a thread's agent? 
We will be losing 'historical' context of a thread, and I think we might run into some access control/instructions edge cases if you can retry older messages? 🤔 
For the history, we could add either a new table to keep track of switches, but nothing too critical for now.
As for the thread relationship, I think ideally we would need to have this `agent_uuid` moved to the `ai_prompt` table instead.

This is a big architectural change, so I don't think we should go with this if you don't see any critical issues with this current approach for slack only, wdyt?

joaoviana:
Closing as this requires a migration + considerations as @tatianainama mentioned above ⬆️

PR #18375, opened 2025-11-26
fix: disable OAuth option when org GitHub integration not configured

almeidabbm:
This fix addresses a UX issue where users could select OAuth authorization during project creation even when organization-level GitHub integration was not configured, leading to confusion and inability to complete the setup.

The OAuth option is now disabled when organization GitHub integration is unavailable, with clear messaging directing users to Settings → Integrations to set it up first. The form also now defaults to Personal Access Token when OAuth is not available for new projects.

### DIFF
--- a/packages/frontend/src/components/ProjectConnection/DbtForms/GithubForm.tsx
+++ b/packages/frontend/src/components/ProjectConnection/DbtForms/GithubForm.tsx
@@ -295,8 +295,8 @@ const GithubForm: FC<{ disabled: boolean }> = ({ disabled }) => {
         savedProject?.dbtConnection?.personal_access_token !== undefined
             ? 'personal_access_token'
             : githubConfig?.enabled
-              ? 'installation_id'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/ProjectConnection/DbtForms/GithubForm.tsx
```diff
@@ -329,6 +335,12 @@ const GithubForm: FC<{ disabled: boolean }> = ({ disabled }) => {
                                         Click here to use another account
                                     </Anchor>
                                 </Text>
+                            ) : !isOAuthAvailable ? (
```
rephus, line 338: Normally when a feature is not available, we don't render it.

almeidabbm:
Internal thread: https://lightdash.slack.com/archives/C03MCQ08UKS/p1764944448444229?thread_ts=1764065307.464209&cid=C03MCQ08UKS

PR #18376, opened 2025-11-26
feat: apply field ordering within groups

### LINKED ISSUE #7125: I want `order_fields_by: index` to work within groups
### Description

Ordering fields in yml only works for non-hierarchical fields. When there are groups or sub-fields (see dates), the order can't be changed by a user. 

For a concrete use case, a user has asked to be able to reorder time fields so they can put them in order of scale instead of alphabetical

image

For the first version of this we don't need to support a new ordering system that would work across multiple groups. It should just respect the ordering in the YAML file, but translate that to the order within the group as well as top-level dimensions. 

It will give users control of the order of fields below the top level.

almeidabbm:
This change enables the `order_fields_by` configuration to control the ordering of fields within groups, not just at the top level. When users set `order_fields_by: 'index'` in their YAML configuration, fields within groups will now be ordered according to their definition sequence in the YAML file. Similarly, `order_fields_by: 'label'` will sort fields alphabetically within groups.

The fix modifies the tree building logic to apply the sorting strategy recursively to all children within group nodes, ensuring consistent ordering at every level of the field hierarchy.

[Source](https://lightdash.slack.com/archives/C03MCQ08UKS/p1763998427714749)

Closes #7125

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/utils.ts
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/utils.ts
@@ -3,10 +3,12 @@ import {
     isCustomDimension,
     isDimension,
     isField,
+    OrderFieldsByStrategy,
     type GroupType,
 } from '@lightdash/common';
 import Fuse from 'fuse.js';
 import { MAX_GROUP_DEPTH } from './constants';
+import { sortNodes } from './sortNodes';
 import {
     isGroupNode,
     type GroupNode,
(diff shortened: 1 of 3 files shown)


pimmey:
## Before
Just picking up the order defined in YML (neither label, nor index):

| order_fields_by: index | order_fields_by: label  |
|--------------------|---|
|             Screenshot 2025-12-03 at 20 01 59 |  Screenshot 2025-12-03 at 20 04 17 |

## After
Respecting `label` and `index`:

| order_fields_by: index | order_fields_by: label  |
|--------------------|---|
|        Screenshot 2025-12-03 at 20 08 25  |  Screenshot 2025-12-03 at 20 06 33  |

Issue #3015, opened 2022-08-19
I want to make geo/map plots as native Lightdash charts


TuringLovesDeathMetal:
### Description

https://echarts.apache.org/examples/en/editor.html?c=map-usa

> [!NOTE]
> Remove env var to enable maps beta as well as feature flag, once beta phase is complete

TuringLovesDeathMetal:
Recommendation from @olly-writes-code to check out deck.gl 👀 

https://github.com/lightdash/lightdash/discussions/1175#discussioncomment-3765512

olly-writes-code:
[Kepler.gl ](https://kepler.gl/)is best in class for map visualizations. It's powered by deck.gl. Kepler is probably overkill for lightdash's needs but it gives a good demo for what world class map visualization look like / work.

olly-writes-code:
Still relevant!

olly-writes-code:
not stale!

olly-writes-code:
not stale!

12ian34:
related to 
https://github.com/lightdash/lightdash/issues/8260

jpetey75:
Feedback from a [customer in slack](https://lightdash.slack.com/archives/C06D1TVN6PM/p1722273995036249?thread_ts=1722272577.720519&cid=C06D1TVN6PM):
> we are interested in post code, countries and regions as we do have a lot of jurisdiction use cases

12ian34:
requested by another prospect yesterday

12ian34:
requested by another prospect today

12ian34:
Another request from a logistics startup wanted postcode level heat map in the uk

pragunbhutani:
I'd love to see this feature as well, speaking from the perspective of a OTT video platform in India. We'd love to be able to see the geographical performance for content we produce in different parts of India.

benjdworx:
I'd like to see maps and support for PostGIS columns with filters. i.e. ability to check if a point is in a polygon associated with a name

benjdworx:
Will this include the ability to display custom POSTGIS polygons on a Google Map or Mapquest Map or is this limited to US states?

benjdworx:
We would like the ability to display a series of latitude/longitudes on a map representing a route traveled by a vehicle. We currently do this using the Google Maps API but it would be valuable to have it integrated into LightDash

kilian-sg:
[...]

PR #18394, opened 2025-11-28
feat: leaflet-based maps

magnew:
Closes: #3015

### Description:

Switches the map base-renderer to Leaflet. Also:
- Lazy load leaflet
- Include heat maps
- Add option to save current map extent
- Adds an optional legend
- Updates tooltip info
- Removes some area options (counties)
- Styles leaflet zoom controls
- Maybe small UX tweaks

#### Scatter plot 
Screenshot 2025-12-01 at 15 01 59

#### Bubble map
Screenshot 2025-12-01 at 15 02 41

#### Area map
Screenshot 2025-12-01 at 15 05 11

#### Heat map
Screenshot 2025-12-01 at 15 03 10

#### Custom shapes
Screenshot 2025-12-01 at 15 00 46

### DIFF
--- a/packages/frontend/src/components/SimpleMap/index.tsx
+++ b/packages/frontend/src/components/SimpleMap/index.tsx
@@ -254,9 +254,162 @@ const SimpleMap: FC<SimpleMapProps> = memo((props) => {
     }, [mapConfig?.geoJsonUrl, mapConfig?.isLatLong]);
 
     // Get location type from visualization config
-    const locationType = isMapVisualizationConfig(visualizationConfig)
-        ? visualizationConfig.chartConfig.validConfig.locationType
-        : undefined;
+    const locationType = useMemo(
+        () =>
+            isMapVisualizationConfig(visualizationConfig)
+                ? visualizationConfig.chartConfig.validConfig.locationType
+                : undefined,
+        [visualizationConfig],
+    );
+
+    // Memoize scatter/lat-long mode calculations
+    const scatterData = useMemo(
+        () => mapConfig?.scatterData || [],
+        [mapConfig?.scatterData],
+    );
+
+    const scatterValueRange = useMemo(() => {
+        const values = scatterData.map((d) => d.value);
+        return {
+            min: values.length > 0 ? Math.min(...values, 0) : 0,
+            max: values.length > 0 ? Math.max(...values, 1) : 1,
+        };
+    }, [scatterData]);
+
+    const sizeScale = useMemo(() => {
+        if (!mapConfig) return null;
+        const sizeMin = mapConfig.sizeRange?.min ?? scatterValueRange.min;
+        const sizeMax = mapConfig.sizeRange?.max ?? scatterValueRange.max;
+        return scaleSqrt()
+            .domain([sizeMin, sizeMax])
+            .range([mapConfig.minBubbleSize, mapConfig.maxBubbleSize]);
+    }, [mapConfig, scatterValueRange.min, scatterValueRange.max]);
+
+    const heatmapPoints = useMemo((): [number, number, number][] => {
+        const { min, max } = scatterValueRange;
+        return scatterData.map((point) => {
+            const intensity =
+                max === min ? 0.5 : (point.value - min) / (max - min);
+            return [point.lat, point.lon, intensity];
+        });
+    }, [scatterData, scatterValueRange]);
+
+    const heatmapGradient = useMemo((): Record<number, string> => {
+        if (!mapConfig) return {};
+        const gradient: Record<number, string> = {};
+        mapConfig.colors.scale.forEach((color, index) => {
+            const position = index / (mapConfig.colors.scale.length - 1);
+            gradient[position] = color;
+        });
+        return gradient;
+    }, [mapConfig]);
+
+    const isHeatmap = mapConfig?.locationType === MapChartType.HEATMAP;
+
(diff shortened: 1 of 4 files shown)


magnew:
* **#18394** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18394?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/vite-env.d.ts
```diff
@@ -5,3 +5,8 @@ declare const __APP_VERSION__: string;
 declare const REACT_SCAN_ENABLED: boolean;
 declare const REACT_QUERY_DEVTOOLS_ENABLED: boolean;
 declare const ECHARTS_V6_ENABLED: boolean;
+
+declare module '*.module.css' {
```
magnew, line 9: I'm not sure if this should go in this PR, but it was needed for the CSS module. Could split out or handle a different way.

packages/frontend/src/components/SimpleMap/index.tsx
```diff
@@ -35,84 +102,450 @@ const LoadingChart = () => (
...
 
-    console.log(
-        'SimpleMap render - isLoading:',
-        isLoading,
-        'mapOptions:',
-        !!mapOptions,
-    );
-
+    // Load GeoJSON data for choropleth maps
     useEffect(() => {
```
magnew, line 170: @rephus This useEffect is my biggest open question in this PR. Do we want to move this or do it differently? It's a lot of network processing for a component. But I didn't want to spend a lot of time on it if we aren't sure really what we need. I could put most of this in a utility -- or streamline it? I think we are using it to process both the json form our server as well as custom ones. Which is efficient, but might not be best.

packages/common/src/types/savedCharts.ts
```diff
@@ -143,7 +143,6 @@ export type GaugeChart = {
 
 export enum MapChartLocation {
```
rephus, line 144: so, when choosing world, we show all contries across the world.
when we choose europe, we only show europe countries. 
and when we choose usa, we only show usa **states**

A bit inconsistent, but I guess this is what probably most people want. 

How about a single `world` with `usa states` ? then we can get rid of thi sselection

 
Image

packages/frontend/src/components/VisualizationConfigs/MapConfig/MapDisplayConfig.tsx
```diff
@@ -216,10 +191,66 @@ export const Display: FC = memo(() => {
...
+                            { value: MapTileBackground.LIGHT, label: 'Light' },
+                            { value: MapTileBackground.DARK, label: 'Dark' },
+                            {
+                                value: MapTileBackground.SATELLITE,
+                                label: 'Satellite',
+                            },
+                        ]}
+                        value={
+                            validConfig.tileBackground ??
+                            MapTileBackground.OPENSTREETMAP
```
rephus: Should we use LIGHT as default map ? perhaps it will be more efficient than openstreetmap

packages/frontend/src/components/SimpleMap/index.tsx
rephus: I agree this can be optimized. We can do it on a separate ticket.

First of all, we should not need to reload the loaded json if latlon changes. 

Then, I was thinking doing the topojson transformation on the backend if needed

It would be cool if we could move this to reactquery

packages/frontend/src/components/SimpleMap/index.tsx
```diff
@@ -35,84 +102,450 @@ const LoadingChart = () => (
...
+        const sizeMin = mapConfig.sizeRange?.min ?? minValue;
+        const sizeMax = mapConfig.sizeRange?.max ?? maxValue;
+
+        // Create d3 sqrt scale for proportional circle sizing (scatter mode)
+        const sizeScale = scaleSqrt()
+            .domain([sizeMin, sizeMax])
+            .range([mapConfig.minBubbleSize, mapConfig.maxBubbleSize]);
+
+        // Prepare heatmap points: [lat, lng, intensity]
+        const heatmapPoints: [number, number, number][] = scatterData.map(
```
rephus: should we memoize these values ?

packages/common/src/types/savedCharts.ts
magnew: Yea, its is a little asymmetric. I think these are the popular ones when you look at demos, but maybe I should look around a little more. We could also add a bunch more options.

packages/frontend/src/components/VisualizationConfigs/MapConfig/MapDisplayConfig.tsx
magnew: I was wondering the same thing. Partly just cause it looks better. Since we both thought so, I'll change it.

packages/frontend/src/components/SimpleMap/index.tsx
magnew: Ok, I'm gonna file this as a ticket in the milestone.

packages/frontend/src/components/SimpleMap/index.tsx
magnew: Yes

PR #18438, opened 2025-12-01
feat: add tooltip for save chart button when no public spaces available

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-224/users-cannot-save-charts-when-all-spaces-are-restricted-no-public

Improve UI when saving charts with no public spaces 

Before: 

Even editors don't get the `save chart` button

Screenshot from 2025-12-01 15-06-11

Now:

## can save spaces (aka editors) 

editors can create new spaces, so there are no limitations
Screenshot from 2025-12-01 15-18-16

Screenshot from 2025-12-01 15-18-31

## Can save charts but can't create spaces (aka interactive viewers)

If no public spaces are available, and you can't create spaces, we disable the action and show the reason on a tooltip

Screenshot from 2025-12-01 15-21-49

### Description:
Improves the SaveChartButton UX by showing a tooltip when there are no public spaces to save a chart to. The button is now disabled when a user can't create charts, and a tooltip explains why. This change also adds a check for whether the user can create a space, which affects how the button is displayed.

### DIFF
--- a/packages/frontend/src/components/Explorer/ExplorerHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/ExplorerHeader/index.tsx
@@ -21,6 +21,7 @@ import { useExplorerQuery } from '../../../hooks/useExplorerQuery';
 import { getExplorerUrlFromCreateSavedChartVersion } from '../../../hooks/useExplorerRoute';
 import useCreateInAnySpaceAccess from '../../../hooks/user/useCreateInAnySpaceAccess';
 import { Can } from '../../../providers/Ability';
+import { useAbilityContext } from '../../../providers/Ability/useAbilityContext';
 import useApp from '../../../providers/App/useApp';
 import { RefreshButton } from '../../RefreshButton';
 import RefreshDbtButton from '../../RefreshDbtButton';
@@ -34,6 +35,7 @@ const ExplorerHeader: FC = memo(() => {
     const { projectUuid } = useParams<{ projectUuid: string }>();
     const { user } = useApp();
     const { onBackToDashboard } = useEmbed();
+    const ability = useAbilityContext();
 
     // Get state from Redux and new hook
     const limit = useExplorerSelector(selectQueryLimit);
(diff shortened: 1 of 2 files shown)


rephus:
* **#18438** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18438?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18438?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/Explorer/ExplorerHeader/index.tsx
```diff
@@ -67,6 +67,14 @@ const ExplorerHeader: FC = memo(() => {
         'SavedChart',
     );
 
+    const userCanCreateSpace = user?.data?.ability?.can(
```
stevelikesmusic: Could you leverage `useAbilityContext`​ to get the ability for a user and/or account?

packages/frontend/src/components/Explorer/ExplorerHeader/index.tsx
```diff
@@ -162,9 +170,24 @@ const ExplorerHeader: FC = memo(() => {
...
+                            label="There are no public spaces to save this chart to"
+                        >
+                            <div>
+                                <SaveChartButton
+                                    isExplorer
+                                    disabled={!userCanCreateCharts}
+                                />
+                            </div>
+                        </Tooltip>
+                    ))}
```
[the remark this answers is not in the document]
stevelikesmusic: Can we add an ability check for being able to save charts at all?

packages/frontend/src/components/Explorer/SaveChartButton/index.tsx
```diff
@@ -83,7 +86,7 @@ const SaveChartButton: FC<{ isExplorer?: boolean }> = ({ isExplorer }) => {
                 maw={300}
             >
                 <Button
-                    disabled={isDisabled}
+                    disabled={isDisabled || disabled}
```
stevelikesmusic: Why not add to the current `isDisabled`​ variable? It might help to change the name of the prop to something explicit to distinguish it from `isDisabled`​

rephus:
Investigating e2e error 
image

**Update** fixed, we should not show that button for explore from here on embed

Issue #18426, opened 2025-12-01
Comparison colors on big numbers are faded in dark mode


tatianainama:
In dark mode, the comparison colors used for big numbers are too faded, reducing their effectiveness and visibility. 

<img width="472" height="392" alt="Image" src="https://github.com/user-attachments/assets/6a71f4b6-509d-4abc-b131-8f6499f8c053" />

**Steps to reproduce:**
- Enable dark mode
- View a big number visualization with comparison colors
- Observe that the comparison colors are faded

**Expected:**
- Comparison colors should be vibrant and clearly visible in dark mode.

**Notes**
This was created using css filters, consider improving colors by using Mantine's components or defining new ones with css using `@mixin dark`

PR #18525, opened 2025-12-03
fix: simple statistic comparison colors in dark theme

tatianainama:
Closes: #18426

### Description:
Updated the color variables in SimpleStatistic component to use Mantine theme variables for better dark mode support. Removed the custom dark mode filter as it's no longer needed with the new color variables.

The changes ensure consistent styling for trend pills (up, down, neutral) across both light and dark themes by using the appropriate Mantine color variables.

_after_

_before_

### DIFF
--- a/packages/frontend/src/components/SimpleStatistic/SimpleStatistic.module.css
+++ b/packages/frontend/src/components/SimpleStatistic/SimpleStatistic.module.css
@@ -6,21 +6,18 @@
     border-radius: 100px;
     border: 1px solid;
     font-weight: 500;
-    @mixin dark {
-        filter: brightness(0.5) contrast(2);
-    }
 }
 
 .trendPillUp {
-    color: var(--mantine-color-green-7);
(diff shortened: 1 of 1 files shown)


tatianainama:
* **#18525** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18525?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18525?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hamzahc1:
The comparisons looks better @tatianainama but the big number and text still seem too dull. My suggestion is the subtitle text should be as bright as the current big number and then the big number should be as bright as the tile headers?

tatianainama:
@hamzahc1 agree!
What do you think of something like these?

_new_


_old_

Issue #18420, opened 2025-12-01
conditional formatting from old chart built in light mode looks off in dark mode


12ian34:
<img width="225" height="709" alt="Image" src="https://github.com/user-attachments/assets/a09935f3-afee-4c23-99ce-8ba25557ac6f" />

tatianainama:
Another example of tables with conditional formatting 

<img width="5036" height="1064" alt="Image" src="https://github.com/user-attachments/assets/dd0b863a-0739-425d-8c83-ca8528f8cc96" />

PR #18534, opened 2025-12-03
fix: conditional formatting support for dark theme

tatianainama:
Closes: #18420

### Description:

Added dark theme support for table cell background colors. This PR introduces a new `isDarkTheme` prop to the `BodyCell` component that is used to apply a color inversion filter when dark theme is active. 

> [!NOTE]
> This is an intermediate solution. Ideally we would have two different palettes for light/dark themes

> [!NOTE]
> This hack doesn't work well when configuring a chart, as colors are inverted

### DIFF
--- a/packages/frontend/src/components/common/Table/ScrollableTable/BodyCell.tsx
+++ b/packages/frontend/src/components/common/Table/ScrollableTable/BodyCell.tsx
@@ -40,6 +40,7 @@ interface CommonBodyCellProps {
     isLargeText?: boolean;
     tooltipContent?: string;
     minimal?: boolean;
+    isDarkTheme?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


tatianainama:
* **#18534** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tatianainama:
@IrakliJani I reproduced this using the default values after adding a rule 🤔  but agree with your point!  I don't feel very confident with this workaround, it's very hacky

PR #18547, opened 2025-12-04
feat: add event loop blocking test to livez endpoint

rephus:
Closes: 

Also in alpha https://github.com/lightdash/lightdash/pull/18549

image

I tried to enforce event loop on analytics, adding extra user attributes/groups, loading multiple dashboards, running refresh... But I can't replicate, I can't also find any code that is particuarlly expensive to cause an event loop. 

If we want to at least fix the platform, we need a consistent way to reproduce this issue. This new option argument in livez does exactly that. 

### Description:
Added a testing feature to the `/livez` endpoint that allows simulating event loop blocking. By passing a `blockSeconds` query parameter, the endpoint will block the event loop for the specified number of seconds (up to 60) using a synchronous busy-wait loop. This is useful for testing how the application behaves under event loop blocking conditions.

### DIFF
--- a/packages/backend/src/routers/apiV1Router.ts
+++ b/packages/backend/src/routers/apiV1Router.ts
@@ -24,6 +24,16 @@ import { userRouter } from './userRouter';
 export const apiV1Router: Router = express.Router();
 
 apiV1Router.get('/livez', async (req, res, next) => {
+    // For testing event loop blocking - pass ?blockSeconds=X to block for X seconds
+    const blockSeconds = parseInt(req.query.blockSeconds as string, 10);
+    if (!Number.isNaN(blockSeconds) && blockSeconds > 0 && blockSeconds <= 60) {
+        const startTime = Date.now();
+        const blockUntil = startTime + blockSeconds * 1000;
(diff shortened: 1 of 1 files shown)


rephus:
* **#18547** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

almeidabbm:
Since this is just on the `livez` endpoint the application will always restart because of the kubernetes `liveness` probe and doesn't allow to test how both readiness and liveness interact (which will fail first) under load 🤔

rephus:
@almeidabbm blockSeconds is an optional argument, it will onlly block used, liveness probe will not be using this parameter.  I could build a separate endpoint just for this, but I thought using an argument on livez would make sense 

This is a simple way, to block the event loop on demand.

Issue #18503, opened 2025-12-03
Don't rely embedding config on user's existence


pimmey:
### Description

Currently, if a user who created an embedding config is deleted - that config is not available anymore. We throw this:
```ts
// embed table does not cascade when the user gets deleted
// so we need to check if the user still exists and throw an error if not
// in the frontend this prompts the user to create a new embed
if (!embed.user_uuid) {
    throw new NotFoundError(`User not found for embed`);
}
```

However, this appears to be confusing to customers. The embedded dashboards stop working and it's not 100% clear why, even with the error shown to the client.

### Steps to Reproduce the Bug or Issue

1. Create embed config with a user
2. Delete that user
3. Try to access the embed 

### version

v0.2218.1

### Cloud or self-hosting

None

PR #18565, opened 2025-12-04
fix: handle embeds with deleted users

pimmey:
Closes: #18503

### Description:
This PR fixes an issue with embedded dashboards and charts when the user who created the embed has been deleted. Previously, the system would throw a "User not found" error, but now it properly handles embeds with missing users by making the user field nullable and using the account's user ID as a fallback when checking feature enablement.

The changes:
1. Remove the check that throws an error when a user is not found for an embed
2. Make the user field in the OssEmbed type nullable
3. Add fallback to account.user.id when checking feature enablement if the embed's user is null (this is used solely as a synthetic identifier for PostHog's `distinct_id`)

This allows embeds to continue working even after the original creator's account has been deleted.

### DIFF
--- a/packages/backend/src/ee/models/EmbedModel.ts
+++ b/packages/backend/src/ee/models/EmbedModel.ts
@@ -47,13 +47,6 @@ export class EmbedModel {
             );
         }
 
-        // embed table does not cascade when the user gets deleted
-        // so we need to check if the user still exists and throw an error if not
-        // in the frontend this prompts the user to create a new embed
-        if (!embed.user_uuid) {
-            throw new NotFoundError(`User not found for embed`);
-        }
-
         const dashboards = await this.database('dashboards')
             .select()
             .whereIn('dashboard_uuid', embed.dashboard_uuids);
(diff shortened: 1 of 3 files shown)


pimmey:
* **#18565** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/models/EmbedModel.ts
```diff
@@ -47,13 +47,6 @@ export class EmbedModel {
             );
         }
 
-        // embed table does not cascade when the user gets deleted
-        // so we need to check if the user still exists and throw an error if not
-        // in the frontend this prompts the user to create a new embed
-        if (!embed.user_uuid) {
-            throw new NotFoundError(`User not found for embed`);
-        }
-
```
stevelikesmusic, line 56: @pimmey I'm not too sure about the effect of allowing embeds to exist if the user that created them is deleted—also not sure why we have this check but not the cascade if the embed user that created the row is important. I _think_ it's safe, as the embed isn't really coupled to a user. I would expect the related user to be there just for auditing and troubleshooting.   
  
I assume we don't do anything to content when the creating user is deleted. Embeds should be the same, as it's really project-level content.

pimmey:
@stevelikesmusic during the stand up a few days back, @ZeRego  mentioned we used to have that cause we used to send requests via that user. However now that we are using JWT, that is no longer the case. I've followed the codepaths, and indeed it seems that it was used only as a meta data for PH's `distinct_id`. But happy to hear the confirmation once again, given it's a EE feature 😅

packages/backend/src/ee/models/EmbedModel.ts
pimmey: Yeah agree it seems a bit sensitive, but also like nothing bad should happen once we remove this 😄 I've tested this scenario locally as well, it just works 🤷

Issue #18400, opened 2025-11-28
Bar chart x-axis sorting not working


owlas:
### Description

[...]

PR #18403, opened 2025-11-28
fix: bar total sorting

almeidabbm:
Relates to: #18400

### Description:
Fixed bar chart sorting when using BAR_TOTALS sort type by preserving the sorted category values in the axis configuration. Previously, ECharts would override our custom sorting, but now we explicitly set the axis data property with our sorted category values to maintain the correct order.

The implementation extracts the sorted category values during the sorting process and applies them to the appropriate axis (x or y depending on flipAxes setting) in the chart configuration.

[Localhost chart for testing
[...]

PR #18581, opened 2025-12-04
feat: introduce x-axis default sort and fix categorical sorting

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/18400

 ### Summary

  - Added a new "Default" sorting option that preserves query result order
  - Made "Ascending/Descending" options actually sort categorical axes alphabetically (previously they only affected the inverse property)
  - Renamed internal variables for clarity

###  New Sort Options

  | Option          | Behavior                                          |
  |-----------------|---------------------------------------------------|
  | Default         | Display in query result order (no forced sorting) - for backwards compatibility |
  | Default (reversed)        | Display in reversed query result order - for backwards compatibility |
  | Ascending       | Sort categorical axis A→Z                         |
  | Descending      | Sort categorical axis Z→A                         |
  | Bars ascending  | Sort by stacked bar totals (lowest first)         |
  | Bars descending | Sort by stacked bar totals (highest first)        |

###  Test Scenarios

  1. Default sorting: Create a bar chart with a categorical x-axis → Select "Default" sort → Bars should display in query result order
  2. Default (reverse) sorting: Create a bar chart with a categorical x-axis → Select "Default (reversed)" sort → Bars should display in reversed query result order
  3. Alphabetical sorting: Create a bar chart with categorical x-axis → Select "Ascending" → Categories should be sorted A→Z → Select "Descending" → Categories should be sorted Z→A
  4. Bar totals sorting: Create a stacked bar chart → Select "Bars ascending" → Bars should be ordered by total height (lowest to highest) → Select "Bars descending" → Bars should be ordered highest to lowest
  5. Bins: Create a bar chart with a binned dimension on x-axis → Test all sort options above

test-frontend
test-backend
test-cli

### DIFF
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -379,10 +379,13 @@ export type XAxis = Axis & {
 
 export enum XAxisSortType {
     DEFAULT = 'default',
+    CATEGORY = 'category',
     BAR_TOTALS = 'bar_totals',
 }
 
 export enum XAxisSort {
+    DEFAULT = 'default',
+    DEFAULT_REVERSED = 'default_reversed',
     ASCENDING = 'ascending',
     DESCENDING = 'descending',
     BAR_TOTALS_ASCENDING = 'bar_totals_ascending',
@@ -392,15 +395,20 @@ export enum XAxisSort {
 export function getXAxisSort(
     xAxis: Pick<XAxis, 'sortType' | 'inverse'> | undefined,
 ): XAxisSort {
-    if (!xAxis) return XAxisSort.ASCENDING;
+    if (!xAxis) return XAxisSort.DEFAULT;
 
     switch (xAxis.sortType) {
+        case XAxisSortType.CATEGORY:
+            return xAxis.inverse ? XAxisSort.DESCENDING : XAxisSort.ASCENDING;
         case XAxisSortType.BAR_TOTALS:
             return xAxis.inverse
                 ? XAxisSort.BAR_TOTALS_DESCENDING
                 : XAxisSort.BAR_TOTALS_ASCENDING;
+        case XAxisSortType.DEFAULT:
         default:
-            return xAxis.inverse ? XAxisSort.DESCENDING : XAxisSort.ASCENDING;
+            return xAxis.inverse
+                ? XAxisSort.DEFAULT_REVERSED
+                : XAxisSort.DEFAULT;
     }
 }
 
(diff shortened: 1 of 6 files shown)


almeidabbm:
* **#18581** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18581?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18581?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#18403** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18403?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rephus:
Open question: This issue was tagged as a regression. Should a regression get fixed by adding a new feature ?

packages/common/src/types/savedCharts.ts
```diff
@@ -392,15 +394,18 @@ export enum XAxisSort {
...
     switch (xAxis.sortType) {
+        case XAxisSortType.CATEGORY:
+            return xAxis.inverse ? XAxisSort.DESCENDING : XAxisSort.ASCENDING;
         case XAxisSortType.BAR_TOTALS:
             return xAxis.inverse
                 ? XAxisSort.BAR_TOTALS_DESCENDING
                 : XAxisSort.BAR_TOTALS_ASCENDING;
+        case XAxisSortType.DEFAULT:
         default:
-            return xAxis.inverse ? XAxisSort.DESCENDING : XAxisSort.ASCENDING;
```
rephus, line 403: Will the change the behaviour of already saved charts from customers ?

almeidabbm:
> Open question: This issue was tagged as a regression. Should a regression get fixed by adding a new feature ?

Don't think this was an actual regression, the x-axis sorting has always been kinda unpredictable.
This stack fixes that by:
- Fixing the totals sorting - trunk PR
- Adding a new sorting type so that we can have predictable sorting on the x-axis - this one

The previous behavior would be that `ascending/descending` would just use the sorting in the results and either invert the order of the x-axis or not which is not actually `ascending/descending` category sorting.

Internal thread for more context: https://lightdash.slack.com/archives/C08N5JUL57V/p1764342180771819

packages/common/src/types/savedCharts.ts
almeidabbm: Good point, this will, should we add `Default` and `Default inverse`? so that it keeps the same sorting for already existing charts? wdyt?

almeidabbm: Just updated the PR with this, 
https://github.com/user-attachments/assets/8299b922-71ba-450b-b79e-3c8238028f5c

Also changed the icons to better reflect this order: 
image

Issue #10245, opened 2024-05-30
testing zapier connection


TuringLovesDeathMetal:
### Description

blah

### Steps to Reproduce the Bug or Issue

blah

### Cloud or self-hosting

None

PR #18622, opened 2025-12-05
feat: add Databricks OAuth authentication support

rephus:
Closes: #10245

Unlike the CLI, i is required to have a databricks app, so we can whitelist the redirect uri : 

```
{"error_description":"redirect_uri 'http:\/\/localhost:3000\/api\/v1\/oauth\/redirect\/databricks' not registered for OAuth application 'databricks-cli'","error":"invalid_request"}
```

- allow users to identify with databricks using oauth in the UI 
- **New functionality** If user has warehouse credentials, even if we don't have `require user credentials` enabled, we'll use their user credentials to authenticate against the warehouse

I tested with a secondary user (javi) rather than my user (javier) 

#### Before making oauth

Screenshot from 2025-12-10 09-07-34

[Screencast from 2025-12-10 09-07-53.webm](https://github.com/user-attachments/assets/ec89e7c0-159e-4b07-b95a-f80b7981fdb6)

#### After making oauth request 

Screenshot from 2025-12-10 09-08-36

### Description:
This PR adds Databricks SSO authentication support to the Databricks warehouse connection form. Users can now choose between Personal Access Token and OAuth authentication methods when connecting to Databricks.

Key changes:
- Added authentication type selector in the Databricks form
- Implemented "Sign in with Databricks" button for OAuth flow
- Created new hooks for Databricks authentication status and login popup
- Added support for requiring user credentials
- Reordered enum values in routes.ts to maintain consistent API behavior

databricks-sso-support


rephus:
* **#18622** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18622?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18622?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1081,48 +1089,65 @@ export class ProjectService extends BaseService {
...
-            if (!isRegisteredUser) {
-                throw new ForbiddenError(
-                    'Embedded users cannot use personal warehouse credentials',
-                );
-            }
+        // Check if user has their own credentials for this project's warehouse type
+        // Only fetch user credentials when:
+        // 1. requireUserCredentials is enabled (user credentials are mandatory)
+        // 2. Databricks warehouse (supports optional user OAuth credentials)
+        const shouldFetchUserCredentials =
```
rephus, line 1096: This is a behaviour change for user warehouse credentials. 

I'd like to test it first with the customer before enabling this feature for other warehouses.

packages/backend/src/controllers/authentication/strategies/databricksStrategy.ts
```diff
@@ -11,27 +11,32 @@ import { URL } from 'url';
...
-      new OAuth2Strategy(
+    : new OAuth2Strategy(
           {
               authorizationURL:
                   lightdashConfig.auth.databricks.authorizationEndpoint,
               tokenURL: lightdashConfig.auth.databricks.tokenEndpoint,
               clientID: lightdashConfig.auth.databricks.clientId,
-              clientSecret: lightdashConfig.auth.databricks.clientSecret,
+              // U2M OAuth uses a public client without a secret, but passport-oauth2 requires this field
+              clientSecret: lightdashConfig.auth.databricks.clientSecret || '',
```
stevelikesmusic, line 30: Is the client secret used at all? If not, why not simply pass the empty string?

packages/backend/src/routers/apiV1Router.ts
```diff
@@ -233,21 +234,35 @@ apiV1Router.get(
     },
 );
 
-apiV1Router.get(
-    lightdashConfig.auth.databricks.loginPath,
-    storeOIDCRedirect,
-    passport.authenticate('databricks'),
-);
+// Databricks OAuth routes - only register if strategy is configured
+// (requires DATABRICKS_OAUTH_CLIENT_ID, DATABRICKS_OAUTH_AUTHORIZATION_ENDPOINT, DATABRICKS_OAUTH_TOKEN_ENDPOINT)
```
stevelikesmusic, line 238: Do we need these comments? This will be in the passport strategy. Maybe better in the CLAUDE.md for databricks/OAuth or docs?

packages/backend/src/controllers/authentication/strategies/databricksStrategy.ts
```diff
@@ -11,27 +11,32 @@ import { URL } from 'url';
...
 import Logger from '../../../logging/logger';
 
+// Databricks U2M OAuth doesn't require a client secret (public client)
+// and requires PKCE (Proof Key for Code Exchange)
+// https://docs.databricks.com/en/dev-tools/auth/oauth-u2m.html
 export const databricksPassportStrategy = !(
     lightdashConfig.auth.databricks.clientId &&
-    lightdashConfig.auth.databricks.clientSecret &&
     lightdashConfig.auth.databricks.authorizationEndpoint &&
     lightdashConfig.auth.databricks.tokenEndpoint
```
stevelikesmusic, line 20: We make this check in a couple places for these three fields. Since they all come from the config, maybe we could set `auth.databricks.enabled` there and use it across the app where we're checking for these?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -832,10 +833,17 @@ export class ProjectService extends BaseService {
                     `Refreshing databricks U2M OAuth token for user ${userUuid}`,
                 );
 
+                // Use client ID from config (same as used during OAuth authorization)
+                // Default to pre-registered Databricks public client
+                const clientId =
+                    this.lightdashConfig.auth.databricks.clientId ||
+                    args.oauthClientId ||
+                    DATABRICKS_DEFAULT_OAUTH_CLIENT_ID;
```
stevelikesmusic, line 841: Is this an environment variable?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -832,10 +833,17 @@ export class ProjectService extends BaseService {
...
+
                 const { accessToken, refreshToken: newRefreshToken } =
                     await refreshDatabricksOAuthToken(
                         args.serverHostName,
-                        args.oauthClientId || 'databricks-cli',
```
stevelikesmusic, line 838: I don't understand this change. Why did we have CLI before? Is this change only scoped for the CLI, or is this a general change to work with the Lightdash UI?

packages/backend/src/services/ProjectService/ProjectService.ts
stevelikesmusic: What's the plan for testing this with the customer? Once this merges, there's only one code path.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1081,48 +1089,65 @@ export class ProjectService extends BaseService {
...
-                    'Embedded users cannot use personal warehouse credentials',
-                );
-            }
+        // Check if user has their own credentials for this project's warehouse type
+        // Only fetch user credentials when:
+        // 1. requireUserCredentials is enabled (user credentials are mandatory)
+        // 2. Databricks warehouse (supports optional user OAuth credentials)
+        const shouldFetchUserCredentials =
+            credentials.requireUserCredentials ||
+            credentials.type === WarehouseTypes.DATABRICKS;
```
stevelikesmusic, line 1098: Since we're in a big Databricks conditional, when would we get here and `credentials.type` not be Databricks?

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1081,48 +1089,65 @@ export class ProjectService extends BaseService {
...
+                      projectUuid,
+                      userId,
+                      credentials.type,
+                  )
+                : undefined;
 
-            const userWarehouseCredentials =
-                await this.userWarehouseCredentialsModel.findForProjectWithSecrets(
-                    projectUuid,
+            if (userWarehouseCredentials !== undefined) {
```
stevelikesmusic: Why an exact check for `undefined` rather than a simple truthy `(userWarehouseCredentials)` like we do below on line 1133 with `else if (credentials.requireUserCredentials)`? Maybe just me, but for JS/TS, I start looking around to see if we're handling other falsey-ish cases vs. other languages where the empty state is stricter. Not a blocker at all, just curious.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1081,48 +1089,65 @@ export class ProjectService extends BaseService {
...
+                        ...credentials,
+                        ...userWarehouseCredentials.credentials,
+                    } as CreateWarehouseCredentials; // force type as typescript doesn't know the types match
+                } else {
+                    throw new UnexpectedServerError(
+                        'User warehouse credentials are not compatible',
+                    );
+                }
+                this.logger.debug(
+                    `Using user warehouse credentials for user ${userId}`,
```
stevelikesmusic, line 1126: Why change the language? "Refreshing" seemed more precise for the next call.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1081,48 +1089,65 @@ export class ProjectService extends BaseService {
...
+                );
+                credentials = await this.refreshCredentials(
+                    credentials,
                     userId,
-                    credentials.type,
                 );
-            if (userWarehouseCredentials === undefined) {
+                userWarehouseCredentialsUuid = userWarehouseCredentials.uuid;
+            } else if (credentials.requireUserCredentials) {
```
stevelikesmusic, line 1133: I don't understand. Why does `requireUserCredentials` indicate there are none? Do we set `requireUserCredentials`to false if we were successful?

packages/frontend/src/components/ProjectConnection/WarehouseForms/DatabricksForm.tsx
```diff
@@ -60,12 +95,70 @@ const DatabricksForm: FC<{
...
+        refetch: refetchAuth,
+    } = useIsDatabricksAuthenticated();
+
+    const isSso =
+        form.values.warehouse?.type === WarehouseTypes.DATABRICKS &&
+        form.values.warehouse.authenticationType ===
+            DatabricksAuthenticationType.OAUTH_U2M;
+
+    const isAuthenticated =
+        data !== undefined && databricksAuthError === null && isSso;
```
stevelikesmusic, line 111: For readability, I find something like this a bit easier:

```typescript
const isAuthenticated = data && !databricksAuthError && isSso
```

packages/frontend/src/components/UserSettings/MyWarehouseConnectionsPanel/CreateCredentialsModal.tsx
```diff
@@ -157,20 +154,20 @@ export const CreateCredentialsModal: FC<Props> = ({
...
+                            those warehouse credentials will be saved during the oauth process
+                            so we don't need the save button */}
+                        {![
+                            WarehouseTypes.BIGQUERY,
+                            WarehouseTypes.SNOWFLAKE,
+                            WarehouseTypes.DATABRICKS,
+                        ].includes(
+                            warehouseType ?? form.values.credentials.type,
+                        ) && (
+                            <Button size="xs" type="submit" disabled={isSaving}>
```
stevelikesmusic, line 167: I wonder if this could be simplified with a variable above the JSX like:

```typescript
const isOauth = [...WarehouseTypes].includes(warehouseType...)

...

!isOauth && <SaveButton />
```

packages/frontend/src/hooks/useDatabricks.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            channel.removeEventListener('message', handleMessage);
+            channel.close();
+            popupWindow.close();
+        };
+
+        channel.addEventListener('message', handleMessage);
+    });
+};
```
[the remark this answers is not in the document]
stevelikesmusic: I'm guessing this is the same for the other warehouse hooks?

stevelikesmusic (commented):
Mostly stylistic/readability comments and questions.

One thing was the mention of a behavioral change that you'd test with the customer. It doesn't look like there's a fork in the current behavior though. So once this merges, will the change be for everyone?

packages/backend/src/controllers/authentication/strategies/databricksStrategy.ts
rephus: it can be, depending on the client, you can have a client with or without secrets.

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: No, this is a constant specified in warehouse client

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: clientId is extracted in a variable. This can be used in both, to refresh the token from the user before making queries.

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: This is different, we still use refreshing below. This is an extra log to know we are using user warehouse credentials

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: if credentials.requireUserCredentials === true, then it means we fetched the user warehouse credentials, if they are present. it should go into userWarehouseCredentials !== undefined), if not, we throw an error because they are required.

packages/frontend/src/hooks/useDatabricks.ts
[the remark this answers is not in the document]
rephus: yeah, we can ignore

PR #18638, opened 2025-12-06
build(dockerfile): reinstall pnpm to fix workspace symlinks (main)

tlinkin:
### Description:
Reinstall pnpm in the build-frontend stage to restore workspace symlinks that are disrupted by COPY. Use a cache mount for the pnpm store to keep rebuilds fast and deterministic, and run with --frozen-lockfile and --prefer-offline.

Also tidy up a comment line in the Dockerfile.

Furthermore I successfully built the docker image on ARM to support Graviton instances.


ZeRego:
Hi @tlinkin 👋 Thanks for the contribution. Can you provide a bit more context on what error you got? Was this problem introduced recently, or never worked? Are you building ARM Docker images (if so, how)?

tlinkin:
Hello 👋

As you can see in this small change, the `pnpm install` step is missing for the frontend, even though it’s present in all the other build steps. Without it, building the Docker image fails because the `COPY` command can’t find the expected files.

We’re building the image for ARM, since our EKS cluster runs on Graviton instances. I’m happy to share that with this fix the image builds and runs stably on our cluster: https://hub.docker.com/r/teodorlinkin/lightdash

It would be great if official ARM images could be published, so we don’t have to maintain a separate pipeline 🙂

Thanks for all your work on Lightdash!

tlinkin:
> Are you building ARM Docker images (if so, how)?

`docker build . -f dockerfile` ?

owlas:
Hey @tlinkin we can't reproduce the problem on our side with the symlinks, i'm closing the PR

PR #18640, opened 2025-12-06
chore: fix various minor bash script issues

Spissable:
### Description:

When running the first time setup I noticed a few teeny tiny issues, e.g. the colors that weren't used correctly. 
In addition I ran [shellcheck](https://github.com/koalaman/shellcheck) which I highly recommend!

Shellcheck issues found and corrected: 
https://www.shellcheck.net/wiki/SC2034 -- Unused variables
https://www.shellcheck.net/wiki/SC2046 -- Unquoted variables
https://www.shellcheck.net/wiki/SC2028 -- echo may not expand escape sequence

Before and after the color fix:
image

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -15,26 +15,18 @@ port=8080
 Started="bash_install.started"
 Failed="bash_install.failed"
 Successful="bash_install.successful"
-Support="bash_install.support"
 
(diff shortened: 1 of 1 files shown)


owlas:
nice thanks @Spissable !

Issue #18598, opened 2025-12-04
Dashboards as code does not support SQL runner charts


jpetey75:
### Description

When I use `lightdash download` on a dashboard with a SQL runner chart, the chart reference shows up in my dashboard YAML, but there is no chart YAML associated to it. 

Then, if I `lightdash upload` that dashboard, it will fail because of that missing reference. 

Lightdash should either skip SQL runner charts, or (ideally) it should support download and upload of those charts. 

### Steps to Reproduce the Bug or Issue

See description

### Cloud or self-hosting

None

PR #18643, opened 2025-12-08
feat: add SQL charts as code 

rephus:
Closes: https://github.com/lightdash/lightdash/issues/18598

Screenshot from 2025-12-08 10-24-26

Screenshot from 2025-12-08 10-32-31
Screenshot from 2025-12-08 10-33-28

### Load test (pagination on sql charts)

image

image

### Description:
This PR adds support for SQL charts in the content-as-code workflow, allowing users to download and upload SQL charts using the CLI. The implementation includes:

1. New API endpoints for getting and upserting SQL charts as code
2. Support in the CLI for downloading and uploading SQL charts
3. Integration with the existing content-as-code infrastructure

This enables users to version control their SQL charts alongside regular charts and dashboards, making it easier to manage and deploy SQL-based visualizations across environments.

### DIFF
--- a/packages/backend/src/models/SavedSqlModel.ts
+++ b/packages/backend/src/models/SavedSqlModel.ts
@@ -37,7 +37,7 @@ type SelectSavedSql = Pick<
     | 'last_viewed_at'
 > &
     Pick<DbSavedSqlVersion, 'sql' | 'limit' | 'config' | 'chart_kind'> &
-    Pick<DbSpace, 'space_uuid'> &
+    Pick<DbSpace, 'space_uuid' | 'path'> &
     Pick<DbProject, 'project_uuid'> &
     Pick<DbOrganization, 'organization_uuid'> & {
         updated_at: Date;
@@ -117,7 +117,7 @@ export class SavedSqlModel {
 
     async find(options: {
         uuid?: string;
-        slug?: string;
+        slugs?: string[];
         projectUuid?: string;
     }) {
         return this.database
@@ -190,6 +190,7 @@ export class SavedSqlModel {
                 `${SpaceTableName}.space_uuid`,
                 `${SpaceTableName}.name as spaceName`,
                 `${SpaceTableName}.is_private as space_is_private`,
+                `${SpaceTableName}.path`,
             ])
             .where((builder) => {
                 if (options.uuid) {
@@ -199,10 +200,10 @@ export class SavedSqlModel {
                     );
                 }
 
-                if (options.slug) {
-                    void builder.where(
+                if (options.slugs && options.slugs.length > 0) {
+                    void builder.whereIn(
                         `${SavedSqlTableName}.slug`,
-                        options.slug,
+                        options.slugs,
                     );
                 }
 
@@ -232,7 +233,7 @@ export class SavedSqlModel {
     }
 
     async getBySlug(projectUuid: string, slug: string) {
-        const results = await this.find({ slug, projectUuid });
+        const results = await this.find({ slugs: [slug], projectUuid });
         const [result] = results;
         if (!result) {
             throw new NotFoundError('Saved sql not found');
@@ -292,15 +293,16 @@ export class SavedSqlModel {
         savedSqlVersionUuid: string;
     }> {
         return this.database.transaction(async (trx) => {
+            // Use provided slug or generate one from the name
+            const finalSlug =
+                data.slug ??
+                (await generateUniqueSlug(trx, SavedSqlTableName, data.name));
+
             const [{ saved_sql_uuid: savedSqlUuid, slug }] = await trx(
                 SavedSqlTableName,
             ).insert(
                 {
-                    slug: await generateUniqueSlug(
-                        trx,
-                        SavedSqlTableName,
-                        data.name,
-                    ),
+                    slug: finalSlug,
                     name: data.name,
                     description: data.description,
                     created_by_user_uuid: userUuid,
--- a/packages/backend/src/services/CoderService/CoderService.ts
+++ b/packages/backend/src/services/CoderService/CoderService.ts
@@ -29,6 +29,7 @@ import {
     Space,
     SpaceMemberRole,
     SpaceSummary,
+    SqlChartAsCode,
     UpdatedByUser,
     type DashboardTileWithSlug,
 } from '@lightdash/common';
@@ -38,6 +39,7 @@ import { LightdashConfig } from '../../config/parseConfig';
 import { DashboardModel } from '../../models/DashboardModel/DashboardModel';
 import { ProjectModel } from '../../models/ProjectModel/ProjectModel';
 import { SavedChartModel } from '../../models/SavedChartModel';
+import { SavedSqlModel } from '../../models/SavedSqlModel';
 import { SpaceModel } from '../../models/SpaceModel';
 import { SchedulerClient } from '../../scheduler/SchedulerClient';
 import { BaseService } from '../BaseService';
@@ -49,6 +51,7 @@ type CoderServiceArguments = {
     analytics: LightdashAnalytics;
     projectModel: ProjectModel;
     savedChartModel: SavedChartModel;
+    savedSqlModel: SavedSqlModel;
     dashboardModel: DashboardModel;
     spaceModel: SpaceModel;
     schedulerClient: SchedulerClient;
@@ -72,6 +75,8 @@ export class CoderService extends BaseService {
 
     savedChartModel: SavedChartModel;
 
+    savedSqlModel: SavedSqlModel;
+
     dashboardModel: DashboardModel;
 
     spaceModel: SpaceModel;
@@ -85,6 +90,7 @@ export class CoderService extends BaseService {
         analytics,
         projectModel,
         savedChartModel,
+        savedSqlModel,
         dashboardModel,
         spaceModel,
         schedulerClient,
@@ -95,6 +101,7 @@ export class CoderService extends BaseService {
         this.analytics = analytics;
         this.projectModel = projectModel;
         this.savedChartModel = savedChartModel;
+        this.savedSqlModel = savedSqlModel;
         this.dashboardModel = dashboardModel;
         this.spaceModel = spaceModel;
         this.schedulerClient = schedulerClient;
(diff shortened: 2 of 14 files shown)


rephus:
* **#18643** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18643?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18643?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/SavedSqlModel.ts
```diff
@@ -118,6 +118,7 @@ export class SavedSqlModel {
     async find(options: {
         uuid?: string;
         slug?: string;
+        slugs?: string[];
```
stevelikesmusic, line 120: Why not change the `slug` field to `slugs`? The model is only used internally and not exposed to a public API.

packages/backend/src/models/SavedSqlModel.ts
```diff
@@ -292,15 +301,16 @@ export class SavedSqlModel {
         savedSqlVersionUuid: string;
     }> {
         return this.database.transaction(async (trx) => {
+            // Use provided slug or generate one from the name
+            const finalSlug =
+                data.slug ??
+                (await generateUniqueSlug(trx, SavedSqlTableName, data.name));
```
stevelikesmusic, line 299: Is it possible a chart is added via code and has a duplicate slug? If someone has nested downloads, they could easily add a duplicated slug between spaces/folders.

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -322,30 +359,72 @@ export class CoderService extends BaseService {
...
+
+        // Query both regular charts and SQL charts in parallel
+        const [charts, sqlChartRows] = await Promise.all([
+            this.savedChartModel.find({
+                slugs: chartSlugs,
+                projectUuid,
+                excludeChartsSavedInDashboard: false,
+                includeOrphanChartsWithinDashboard: true,
+            }),
+            this.savedSqlModel.find({
```
stevelikesmusic, line 378: I know developers have access to SQL Runner. But can non-developers run `lightdash download/upload`? If so, it might be a good idea to check permissions for SqlRunner?

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -322,30 +359,72 @@ export class CoderService extends BaseService {
...
         return tiles.map((tile) => {
             if (isAnyChartTile(tile)) {
-                const savedChart = charts.find(
-                    (chart) => chart.slug === tile.properties.chartSlug,
-                );
+                const { chartSlug } = tile.properties;
+
+                // Check if it's a SQL chart or regular chart
+                const sqlChartUuid = sqlChartSlugToUuid.get(chartSlug);
+                const regularChartUuid = regularChartSlugToUuid.get(chartSlug);
```
stevelikesmusic: I find this a bit confusing. Why not have a single Map with an `isSql` field or similar? Then it's a bit more obvious what we're checking.

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -637,7 +716,6 @@ export class CoderService extends BaseService {
             this.savedChartModel.get(chart.uuid),
         );
         const charts = await Promise.all(chartPromises);
-        const missingIds = CoderService.getMissingIds(chartIds, charts);
```
stevelikesmusic, line 640: Why move this?

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -677,6 +757,122 @@ export class CoderService extends BaseService {
...
+    }> {
+        const project = await this.projectModel.get(projectUuid);
+        if (!project) {
+            throw new NotFoundError(`Project ${projectUuid} not found`);
+        }
+
+        if (
+            user.ability.cannot(
+                'manage',
+                subject('ContentAsCode', {
```
stevelikesmusic, line 781: Could someone with `ContentAsCode` not have `SqlRunner`access?

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -677,6 +757,122 @@ export class CoderService extends BaseService {
...
+            )
+        ) {
+            throw new ForbiddenError(
+                'You are not allowed to download SQL charts',
+            );
+        }
+
+        // For SQL charts, we use slugs directly (no UUID to slug conversion needed)
+        // since SQL charts are only identified by slug in the as-code workflow
+        const slugs = chartIds;
```
stevelikesmusic, line 794: If we know these are really slugs, why call them `chartIds` in the function params only to rename them here in the function?

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -677,6 +757,122 @@ export class CoderService extends BaseService {
         };
     }
 
+    async getSqlCharts(
```
stevelikesmusic, line 762: Is most of this a duplicate of `getCharts`? Could the logic be consolidated? Whoever updates or the other in the future will need to have context to update both functions.

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -677,6 +757,122 @@ export class CoderService extends BaseService {
...
+                    chartKind: row.chart_kind,
+                    lastUpdatedAt: row.last_version_updated_at,
+                },
+                row.path,
+            ),
+        );
+
+        // Calculate missing IDs
+        const foundSlugs = new Set(sqlChartRows.map((c) => c.slug));
+        const missingIds = chartIds
```
stevelikesmusic, line 866: Why `chartsIds` and not `slugs`?

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -853,6 +1049,134 @@ export class CoderService extends BaseService {
...
+        sqlChartAsCode: SqlChartAsCode,
+        skipSpaceCreate?: boolean,
+        publicSpaceCreate?: boolean,
+    ): Promise<PromotionChanges> {
+        const project = await this.projectModel.get(projectUuid);
+
+        if (
+            user.ability.cannot(
+                'manage',
+                subject('ContentAsCode', {
```
stevelikesmusic, line 1067: Same question about SqlRunner permissions

packages/backend/src/services/CoderService/CoderService.ts
```diff
@@ -853,6 +1049,134 @@ export class CoderService extends BaseService {
         return promotionChanges;
     }
 
+    async upsertSqlChart(
+        user: SessionUser,
+        projectUuid: string,
+        slug: string,
+        sqlChartAsCode: SqlChartAsCode,
```
stevelikesmusic, line 1058: Why pass an explicit `slug` if `sqlChartAsCode` has a slug field? I don't get why we use one over the other in this method.

packages/cli/src/handlers/download.ts
```diff
@@ -107,12 +114,32 @@ const createDirForContent = async (
...
+    }
+};
+
 const writeContent = async (
     contentAsCode: ContentAsCodeType,
     outputDir: string,
     languageMap: boolean,
 ) => {
-    const itemPath = path.join(outputDir, `${contentAsCode.content.slug}.yml`);
+    const extension = getFileExtension(contentAsCode.type);
```
stevelikesmusic, line 138: Why not something simpler like:

```typescript
const extension = contentAsCode.type === 'sqlChart' ? '.sql.yml' : '.yml'
```

packages/cli/src/handlers/download.ts
```diff
@@ -569,6 +688,11 @@ const logUploadChanges = (changes: Record<string, number>) => {
     });
 };
 
+// SQL charts have 'sql' field instead of 'tableName'/'metricQuery'
+const isSqlChart = (
```
stevelikesmusic, line 701: We could even use this function when getting the extension above.

packages/cli/src/handlers/download.ts
```diff
@@ -355,7 +462,19 @@ export const downloadContent = async (
         offset = contentAsCode.offset;
     } while (contentAsCode.offset < contentAsCode.total);
 
-    return [contentAsCode.total, [...new Set(chartSlugs)]];
+    // Download SQL charts separately (only when downloading charts, not dashboards)
+    if (type === 'charts') {
+        sqlChartsTotal = await downloadSqlCharts(
```
stevelikesmusic: Is it possible to that a savedQuery and saveSql both have the same slugs? Is that an edge case we're catching?

packages/e2e/cypress/e2e/api/contentAsCode.cy.ts
```diff
@@ -205,3 +206,166 @@ describe('Dashboards as Code API', () => {
...
+        );
+        cy.login();
+    });
+
+    it('make sure the SQL chart is loaded from YML', () => {
+        cy.log('sqlChartAsCode', sqlChartAsCode);
+        cy.wrap(sqlChartAsCode).should('exist');
+        cy.wrap(sqlChartAsCode.sql).should('exist');
+        cy.wrap(sqlChartAsCode.chartKind).should('exist');
+    });
```
stevelikesmusic: Is this test necessary? It tests Cypress rather than our application

packages/e2e/cypress/e2e/api/contentAsCode.cy.ts
```diff
@@ -205,3 +206,166 @@ describe('Dashboards as Code API', () => {
...
+
+    it('should download SQL charts as code', () => {
```
stevelikesmusic, line 223: Do we need to read from the filesystem in the before for this test?

packages/e2e/cypress/cli/api/contentAsCode.cy.ts
```diff
@@ -103,4 +103,153 @@ describe('Content as Code CLI', () => {
...
+  columns: {}
+chartKind: table
+spaceSlug: jaffle-shop
+version: 1
+updatedAt: "${new Date().toISOString()}"
+downloadedAt: "${new Date(Date.now() - 60000).toISOString()}"
+`;
+
+        cy.exec(
+            `echo '${sqlChartContent}' > ${lightdashDir}/charts/${sqlChartSlug}.sql.yml`,
```
stevelikesmusic, line 135: Why not use the same chart we added in support for the API tests?

packages/backend/src/models/SavedSqlModel.ts
rephus: I believe this is an existing pattern in other find methods. But I can take a look at how many usages of this method there are

packages/backend/src/models/SavedSqlModel.ts
rephus: uniqueness is per type, so you can have the same slug for a saved chart and a sql chart and they should not override. 

If the same type has the same slug, we will consider both the same ,so we overwrite them, that is the existing behaviour

packages/backend/src/services/CoderService/CoderService.ts
rephus: with custom permissions, maybe. In that case, you should be able to donwload/upload sql charts, but not run them

packages/backend/src/services/CoderService/CoderService.ts
rephus: It is almost the same yes, initially it was combined, but I found it more confusing, so I decided to split the logic and create separate endpoints for both. 

I could extract some common logic if that helps

packages/cli/src/handlers/download.ts
rephus: Downloading is not a problem, because we store it in a separate sql.yml file
Uploading is also not a problem, because we use different endpoint, depending on the chart type.

Both can coexist in the database, they only need to be unique per type.

Issue #18627, opened 2025-12-05
CLI deploy conflicts with changeset metrics causing explore failures


tatianainama:
### Description

When creating metrics using changesets and later deploy via CLI with dbt files containing metrics with the same name, the explores becomes broken due to entity conflicts during changeset application

### Steps to Reproduce the Bug or Issue

1. Use AI agent to create a metric 
2. Add a metric with the same name in dbt files
3. run `lightdash deploy` with the updated dbt files
4. try access the explore of the table with the metric

<img width="945" height="219" alt="Image" src="https://github.com/user-attachments/assets/5695656b-65b0-46d7-b099-f15663e22cd9" />

https://github.com/user-attachments/assets/566fe1f9-30cd-4011-9bd8-2161308044e2

### Cloud or self-hosting

None

PR #18663, opened 2025-12-08
fix: changeset conflicts causes explore failures

tatianainama:
Closes: #18627

### Description:
Implemented automatic conflict resolution for changesets by adding a retry mechanism that handles conflicts gracefully. When a conflict occurs during changeset application, the system now:

1. Identifies the conflicting change using the new `ChangesetConflictError` class
2. Automatically removes the conflicting change
3. Retries applying the remaining changes (up to 3 times)
4. Logs detailed information about the conflict resolution process

This improves system resilience by preventing changeset conflicts from causing failures in the project model.

_logs on conflict resolution_

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -2,6 +2,8 @@ import {
     AlreadyExistsError,
     AnyType,
     BigqueryAuthenticationType,
+    ChangeBase,
+    ChangesetConflictError,
     ChangesetUtils,
     CompiledTable,
     CreateProject,
@@ -1078,9 +1080,10 @@ export class ProjectModel {
                 );
 
                 if (changeset && applyChangeset) {
-                    finalExplores = ChangesetUtils.applyChangeset(
+                    finalExplores = await this.applyChangesetWithRetry(
                         changeset,
                         finalExplores,
+                        projectUuid,
                     );
                 }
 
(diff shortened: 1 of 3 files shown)


tatianainama:
* **#18663** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18663?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18663?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1096,6 +1099,85 @@ export class ProjectModel {
...
 
+    /**
+     * Apply changeset to explores with automatic conflict resolution
+     * Deletes conflicting changes and retries up to 3 times
+     */
+    private async applyChangesetWithRetry(
+        changeset: { changes: Array<ChangeBase & { changeUuid: string }> },
+        explores: Record<string, Explore | ExploreError>,
+        projectUuid: string,
+        maxRetries: number = 3,
```
IrakliJani, line 1110: Isn't this a bit too much? If the changeset can't be applied, what will make it succeed on the next attempt?

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1096,6 +1099,85 @@ export class ProjectModel {
...
+                        },
+                    );
+
+                    // If we cannot remove the conflicted change, bubble up error
+                    if (!error.data.changeUuid) throw error;
+
+                    // eslint-disable-next-line no-await-in-loop
+                    await this.changesetModel.revertChange(
+                        error.data.changeUuid,
+                    );
```
IrakliJani, line 1141: Do we actually need to revert, or is silently failing (with logging) enough for now?

packages/backend/src/models/ProjectModel/ProjectModel.ts
tatianainama: I think the current approach fails as soon as a change fails, but it doesn't cover all the possible failures, so you might have multiple conflicting changes, even after fixing one.  
I thought of this as a quick solution, but ideally we would collect all conflicts before throwing

packages/backend/src/models/ProjectModel/ProjectModel.ts
tatianainama: I think it might be more confusing having the changeset in the UI that might not actually be applied. The conflict issue comes from the name, so it could be that the change created doesn't do the same thing as the new metric. What you see is actually different from what the changeset UI says 🤔   
I get your point though, so I'm not sure what would be best tbh!

tatianainama:
Closing in favor of simpler solution #18705

PR #18742, opened 2025-12-10
feat: add screenshot ready indicator for more reliable dashboard captures

pimmey:
Closes: GLITCH-95 

### Description:
This PR improves dashboard and chart screenshot reliability by adding a new indicator mechanism that explicitly signals when content is ready to be captured. 

The implementation adds a hidden DOM element (`#lightdash-ready-indicator`) that appears only when all tiles have finished loading. Each visualization component now tracks its loading state and signals when it's ready for screenshot. The backend can be configured to use this new approach via the `SCHEDULER_USE_SCREENSHOT_READY_INDICATOR=true` environment variable.

This approach is more reliable than the previous method which relied on waiting for loading overlays to disappear, as it provides a positive signal when content is ready rather than waiting for the absence of loading indicators.

### DIFF
--- a/packages/frontend/src/pages/MinimalSavedExplorer.tsx
+++ b/packages/frontend/src/pages/MinimalSavedExplorer.tsx
@@ -66,10 +66,21 @@ const MinimalExplorerContent = memo(() => {
 
     useEffect(() => {
         if (hasSignaledReady.current) return;
-        if (!savedChart || isLoadingQueryResults) return;
+        if (
+            !savedChart ||
+            isLoadingQueryResults ||
+            health.isInitialLoading ||
+            !health.data
+        )
+            return;
         setIsScreenshotReady(true);
         hasSignaledReady.current = true;
-    }, [savedChart, isLoadingQueryResults]);
+    }, [
+        savedChart,
+        isLoadingQueryResults,
+        health.isInitialLoading,
+        health.data,
+    ]);
 
     if (!savedChart || health.isInitialLoading || !health.data) {
         return null;


pimmey:
* **#18765** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18765?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18742** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18742?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18742?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pimmey:
Nice catch!

Repro'd and fixed.

<details>

<summary>Before/After</summary>

| Before | After  |
|--------------------|---|
|          slack-image_jaffle_dashboard_CidlScG-stMBQZeXdxE1I |  slack-image_jaffle_dashboard_ogoh_lFljmOZKw3g3RKyq      |

</details>

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -266,8 +266,24 @@ export class UnfurlService extends BaseService {
             page.on('response', responseHandler);
         });
 
-        // This is needed because tables are lazy loaded so we have no way of knowing when the chart is displayed
+        // Wait for dashboard/chart to be ready for screenshot
+        const self = this;
         async function waitForAllLoaded() {
```
almeidabbm, line 271: `waitForPaginatedResultsResponse` - for minimal charts calls on this function
This means that if we enable the feature flag and have a single chart scheduled delivery it might not be delivered because we only add the `ScreenshotReadyIndicator` in the upstack PR

almeidabbm (changes requested):
Can no longer reproduce the issue I had previously on the dashboard export.
Believe we might need to either change the logic so that this PR only listens for the presence of `ScreenshotReadyIndicator` for dashboards or bring in the upstack PR code into it

PR #18744, opened 2025-12-10
ci: build and push pr docker images using depot

owlas:


owlas:
closing because it's slower

PR #18757, opened 2025-12-11
fix: extract and improve CSV limit handling logic

rephus:
Closes:  https://linear.app/lightdash/issue/CENG-270/gcp-alert-oom-log-event-occurred

Unify logic when downloading CSV on underlyingdata and dashboards. 

Both were passing MAXINT for all results. But instead, we now copy the logic from explore CSV, where we apply the MAX_CSV_ROWS limit based on the ENV variable 

Refactored the limit to be reused on different places

### Description:
Refactored query limit handling by extracting the logic into a dedicated utility module. Created a new `csvLimitUtils.ts` file with two functions:

1. `calculateCsvRowLimit` - Calculates the appropriate row limit based on column count and cell limits
2. `metricQueryWithLimit` - Applies limits to metric queries with proper validation

This change standardizes how we handle query limits across the application, particularly for CSV exports and underlying data queries. The implementation now consistently handles three limit scenarios:
- `undefined`: Use existing query limit with validation
- `null`: Apply maximum rows based on cell limit configuration
- `number`: Apply specific limit capped by cell limit configuration

The code now properly handles the `limit` parameter as `number | null | undefined` throughout the codebase, improving type safety.

AI: I've analyzed the diff and provided a concise PR description that focuses on the key changes: extracting query limit logic into a dedicated utility module, standardizing limit handling across the application, and improving type safety. I've maintained the requested template format while avoiding generalities about changing multiple files or updating parameters.

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -17189,7 +17189,7 @@ const models: TsoaRoute.Models = {
         },
     },
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
-    'PartialObjectDeep___91_x-string_93__58__name_63_-string-or-undefined--description_63_-string-or-undefined--tiles_63__58__40__type-DashboardTileTypes.SAVED_CHART-or-DashboardTileTypes.SQL_CHART--properties_58__title-string--chartName-string__-or-_type-DashboardTileTypes.MARKDOWN--properties_58__title-string--content-string__-or-_type-DashboardTileTypes.LOOM--properties_58__title-string___41_-Array-or-undefined__._recurseIntoArrays-true__':
+    'PartialObjectDeep___91_x-string_93__58__name_63_-string-or-undefined--tiles_63__58__40__type-DashboardTileTypes.SAVED_CHART-or-DashboardTileTypes.SQL_CHART--properties_58__title-string--chartName-string__-or-_type-DashboardTileTypes.MARKDOWN--properties_58__title-string--content-string__-or-_type-DashboardTileTypes.LOOM--properties_58__title-string___41_-Array-or-undefined--description_63_-string-or-undefined__._recurseIntoArrays-true__':
         {
             dataType: 'refAlias',
             type: {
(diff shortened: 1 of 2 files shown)


rephus:
* **#18757** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18757?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18757?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2362,39 +2363,6 @@ export class ProjectService extends BaseService {
         };
     }
 
-    private metricQueryWithLimit(
```
rephus, line 2390: extracted into a utils function

rephus:
The way we download CSV on explore/dashboard feel very intesive for the backend. We could  create a single endpoint that downloads the data on the worker, right ?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -2688,11 +2682,18 @@ export class AsyncQueryService extends ProjectService {
...
             tableCalculations: [],
             additionalMetrics: [],
         };
 
+        const underlyingDataMetricQueryWithLimit = applyMetricQueryLimit(
+            underlyingDataMetricQuery,
+            limit,
+            this.lightdashConfig.query?.csvCellsLimit,
+            this.lightdashConfig.query?.maxLimit,
+        );
```
almeidabbm, line 2695: So in this case we will be assuming the `underlying-data` query will always have a `500` rows limit unless we explicitly pass a limit from the frontend?

almeidabbm (approved):
LGTM, just a question below❓

almeidabbm:
> The way we download CSV on explore/dashboard feel very intesive for the backend. We could create a single endpoint that downloads the data on the worker, right ?

IIRC the downloads are already scheduled to run on the worker via the `schedule-download` endpoint 🤔 Most likely the issue is that the query still runs in the backend, downloading just receives a query uuid to process the results from the query

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
rephus: That's a great question, I looked again into `metricQueryWithLimit` and we override the `limit:500` with whatever new limit is from `applyMetricQueryLimit` , so hardcoding a 500 should have no impact (we just need to build a valid metricquery before calling this method). Right ? 

image

almeidabbm: Absolutely! just wanted to confirm that this was the intended behavior, it makes sense!

PR #18830, opened 2025-12-15
chore: add beta badge to maps

magnew:
### Description:

- Adds the beta badge to the map chart type
- I decided this made more sense in the selection than on the map itself -- this way developers selecting the chart will see it, but viewers will not. We could add it elsewhere if necessary. 
- I re-used the existing beta badge

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
@@ -5,7 +5,7 @@ import {
     FeatureFlags,
     isSeriesWithMixedChartTypes,
 } from '@lightdash/common';
-import { Button, Menu } from '@mantine/core';
+import { Button, Group, Menu } from '@mantine/core';
 import {
     IconChartArea,
     IconChartAreaLine,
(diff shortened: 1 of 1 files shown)


magnew:
* **#18830** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18830?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18830?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
```diff
@@ -188,8 +189,8 @@ const VisualizationCardOptions: FC = memo(() => {
                 };
             case ChartType.MAP:
                 return {
-                    text: 'Map',
-                    icon: <MantineIcon icon={IconMap} color="gray" />,
+                    text: 'Map (Beta)',
```
rephus, line 192: we can probably remove this (Beta) if we already have the official badge below 

Image

rephus (approved):

packages/frontend/src/components/Explorer/VisualizationCardOptions/index.tsx
magnew: Sure, yea, I wasn't sure how heavy-handed to be

magnew: Well, now that I played with it, I think we should keep it. Mostly because if you are editing an existing map, it's the only indication that its a beta. I want editors to see it, but not necessarily viewers so I dont want to put it on the map itself. 

Let me know if you disagree 😄 -- I don't have a strong opinion, but it seems like a reasonable compromise at the moment.

PR #18854, opened 2025-12-16
fix: normalize warehouse name comparison in mergeWarehouseCredentials

rephus:
Closes: #ISSUE_NUMBER

### Description:
Fix Snowflake warehouse comparison in mergeWarehouseCredentials function by adding case-insensitive and whitespace-trimming logic. This ensures that warehouses with the same name but different casing or whitespace are properly recognized as identical.

### DIFF
--- a/packages/common/src/types/projects.ts
+++ b/packages/common/src/types/projects.ts
@@ -283,6 +283,10 @@ export const mergeWarehouseCredentials = <T extends CreateWarehouseCredentials>(
 ): T => {
     // If types don't match, return newCredentials as-is (can't merge different warehouse types)
     if (baseCredentials.type !== newCredentials.type) {
+        // eslint-disable-next-line no-console
+        console.info(
+            `Skipping merge of warehouse credentials due to differing types: ${baseCredentials.type} and ${newCredentials.type}`,
+        );
         return newCredentials;
     }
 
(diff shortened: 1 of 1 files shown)


rephus:
* **#18854** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18854?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18854?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/types/projects.ts
```diff
@@ -291,7 +291,8 @@ export const mergeWarehouseCredentials = <T extends CreateWarehouseCredentials>(
     if (
         baseCredentials.type === WarehouseTypes.SNOWFLAKE &&
         newCredentials.type === WarehouseTypes.SNOWFLAKE &&
-        baseCredentials.warehouse !== newCredentials.warehouse
+        baseCredentials.warehouse.toLowerCase().trim() !==
```
rephus, line 298: This is related to the issue we were discussing during the call. Perhaps is a case sensitivity issue that is preventing from copying credentials

We can't simply remove this limitation, since some customers don't want to override the config becuase they are very different...

PR #18856, opened 2025-12-16
chore: limit map extent

magnew:
Closes: PROD-1084

### Description:

Limits the extent of maps so that you can't zoom out infinitely. This is a bit of a tricky problem since we don't have set container sizes. We need to allow zooming out and try not to show weirdness at the edges. This uses some suggested settings from leaflet to:
- Limit the amount a user can zoom out
- Only repeat the map at the edges once

So fully zoomed out it looks like
Screenshot 2025-12-16 at 16 25 02

I tested a version where we only show one map, but it looks a little...bad. 
Screenshot 2025-12-16 at 16 37 59

I suppose we could make a user setting, but this seems like a reasonable start.

### DIFF
--- a/packages/frontend/src/components/SimpleMap/index.tsx
+++ b/packages/frontend/src/components/SimpleMap/index.tsx
@@ -440,6 +440,12 @@ const SimpleMap: FC<SimpleMapProps> = memo((props) => {
                     zoom={mapConfig.zoom}
                     style={{ height: '100%', width: '100%' }}
                     scrollWheelZoom
+                    minZoom={2}
+                    maxBounds={[
+                        [-90, -180],
+                        [90, 180],
+                    ]}
+                    maxBoundsViscosity={1.0}
                 >
                     <MapRefUpdater mapRef={leafletMapRef} />
                     <MapExtentTracker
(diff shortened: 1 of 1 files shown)


magnew:
* **#18865** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18865?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18860** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18860?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18856** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18856?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus (approved):
I agree the only 1 map version look bad. Perhaps there is a way to use like a white background instead ? We can investigate other options on a separate PR

magnew:
> I agree the only 1 map version look bad. Perhaps there is a way to use like a white background instead ? We can investigate other options on a separate PR

Yea, we could try some other options. I think in the basic case it depends on the tile provider. FWIW, its a bit of an edge case to show whole world maps with no bounds set, I think. Most of the requests we have are for specific areas

PR #18866, opened 2025-12-16
feat: release ai generation of chart metadata to all

joaoviana:
Closes: N/A

### Description:

Adds support for wide-purpose AI tasks using a shared Anthropic API key. This implementation:

1. Creates a new `getWidePurposeModel` method in `AiService` that:
   - First checks for a shared Anthropic API key (claude-haiku-4-5)
   - Falls back to the user's AI Copilot configuration if enabled

2. Refactors the chart metadata generation to use this new method instead of directly checking for AI Copilot access

3. Adds `hasSharedAnthropicApiKey` flag to the health endpoint response

4. Creates a new React hook `useAiWidePurposeFeaturesCheck` that checks if wide-purpose AI features are available (either through shared key or user's AI Copilot)

This change allows certain AI features to work for all users when a shared Anthropic API key is configured, without requiring individual AI Copilot access.

### DIFF
--- a/packages/backend/src/ee/services/AiService/AiService.ts
+++ b/packages/backend/src/ee/services/AiService/AiService.ts
@@ -16,6 +16,7 @@ import {
     isDashboardChartTileType,
     isField,
 } from '@lightdash/common';
+import { LanguageModel } from 'ai';
 import { LightdashAnalytics } from '../../../analytics/LightdashAnalytics';
 import { fromSession } from '../../../auth/account';
 import { LightdashConfig } from '../../../config/parseConfig';
@@ -33,6 +34,8 @@ import OpenAi from '../../clients/OpenAi';
 import { DashboardSummaryModel } from '../../models/DashboardSummaryModel';
 import { generateChartMetadata as generateChartMetadataFromContext } from '../ai/agents/chartMetadataGenerator';
 import { getModel } from '../ai/models';
+import { getAnthropicModel } from '../ai/models/anthropic-claude';
+import { getModelPreset } from '../ai/models/presets';
 import {
     fieldDesc,
     formatSummaryArray,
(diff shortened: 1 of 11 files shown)


joaoviana:
* **#18892** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#18866** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18866?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/ee/features/aiCopilot/hooks/useAiWidePurposeFeaturesCheck.ts
```diff
@@ -0,0 +1,13 @@
+import { CommercialFeatureFlags } from '@lightdash/common';
+import useHealth from '../../../../hooks/health/useHealth';
+import { useFeatureFlag } from '../../../../hooks/useFeatureFlagEnabled';
+
+/**
+ * Checks if the ai wide purpose features are available.
+ * It checks if the shared anthropic api key is available or if the ai copilot feature flag is enabled
+ */
+export const useAiWidePurposeFeaturesCheck = () => {
```
joaoviana: would you move this hook somewhere else? I think I could but don't know if it should be be in features/ai or something?

packages/backend/src/ee/services/AiService/AiService.ts
```diff
@@ -88,6 +91,49 @@ export class AiService {
...
     }
 
+    /**
+     * Gets a language model for "wide-purpose" tasks.
+     * 1. Checks anthropic shared key
+     * 2. Falls back to AI Copilot if the feature flag is enabled for the user,
+     *    using their configured model.
+     *
+     * @returns The language model
+     */
```
joaoviana, line 101: what do you think of this order?

packages/backend/src/services/HealthService/HealthService.ts
```diff
@@ -203,6 +203,9 @@ export class HealthService extends BaseService {
                     this.lightdashConfig.ai.analyticsProjectUuid,
                 analyticsDashboardUuid:
                     this.lightdashConfig.ai.analyticsDashboardUuid,
+                hasSharedAnthropicApiKey:
+                    !!this.lightdashConfig.ai.copilot.providers.anthropic
+                        ?.apiKey,
```
notgiorgi: I'd say let's not use `anthropic` here as it might as well be any other provider.

packages/backend/src/ee/services/AiService/AiService.ts
notgiorgi: The logic sounds looks great!

I'm wondering if we should name this better

packages/frontend/src/ee/features/aiCopilot/hooks/useAiWidePurposeFeaturesCheck.ts
notgiorgi: Yea, probably better not to put it under copilot

joaoviana: yea, I will create features/ai and also move the dashboard summary (not widely used) feature under there. Would make more sense

packages/backend/src/services/HealthService/HealthService.ts
joaoviana: yea I did it like this because
- feature flag service isn't used here - but probably should? so I can do the check in one go **or** I can import aiService and check if ambient ai is enabled. wdyt?

notgiorgi: I think the logic you have here is good! Just wasn't sure about name, `ambientAiEnabled` would work better I think.

joaoviana: ahhhh :man-facepalming: sorry, got it! But ambientAi should be enabled even if there's no anthropic api key no?

joaoviana: but lmk if that extra check (to see if copilot is enabled) isn't necessary for now

PR #364, opened 2021-08-24
Update README.md

TuringLovesDeathMetal:
adds product roadmap to readme

PR #18894, opened 2025-12-17
security: update hono override to ^4.10.2 (CVE-2025-62610)

owlas:
## Summary
- Updates the hono pnpm override from ^4.9.6 to ^4.10.2
- Fixes CVE-2025-62610 (Hono Improper Authorization vulnerability in JWT Audience Validation)
- Resolves Dependabot alert #364

## Context
hono is a transitive dependency of `@mastra/core` which is used by the backend. Since @mastra/core@0.24.8 still depends on hono@^4.9.7 (vulnerable), we need the override to enforce the patched version.

## Test plan
- [x] pnpm install succeeds
- [ ] CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     },
     "pnpm": {
         "overrides": {
-            "hono": "^4.9.6",
+            "hono": "^4.10.2",
             "linkifyjs": "^4.3.2",
             "axios": "^1.12.0",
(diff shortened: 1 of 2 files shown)


owlas:
Closing - @mastra/core is a dead dependency. Will remove it entirely instead of using an override.

PR #18919, opened 2025-12-17
fix: use refreshToken consistently in ProjectService

rephus:
Closes: 

List of changes: 

- `improvement`: UserService: new snowflake user warehouse credentials will be using `refreshToken` instead of `token`
- `improvement`: Added extra `clearSecretsFromCredentials` security method that removes all secrets from the project/org credentials before overriding the details with the user (in case there is a missmatch, the authentication will fail) 
- `bugfix`: Remove bug: args.refreshToken || args.token on `refreshToken` method.  We are only expecting `refreshToken` now 
- `bugfix`: UserWarehouseCredentialsModel: We throw errors when snowflake user oauth does not have refresh token , and force users to reauthenticate

Screenshot from 2025-12-18 11-23-51

Added tests:

  1. User has refreshToken → works correctly, uses user token instead of project token
  2. User has token (old bug) → throws 'Error refreshing snowflake token'
  3. requireUserCredentials is false → uses project credentials

### Description:
Fixes a bug in the ProjectService where the refreshToken was incorrectly being set from either args.refreshToken or args.token. This PR removes the fallback to args.token and ensures we consistently use the refreshToken property when refreshing warehouse credentials.

### DIFF
--- a/packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
+++ b/packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
@@ -1,5 +1,8 @@
 import {
     NotFoundError,
+    SnowflakeAuthenticationType,
+    snowflakeSsoUserCredentialsSchema,
+    SnowflakeTokenError,
     UnexpectedServerError,
     UpsertUserWarehouseCredentials,
     UserWarehouseCredentials,
@@ -176,11 +179,32 @@ export class UserWarehouseCredentialsModel {
             userUuid,
             warehouseType,
         );
+
         if (credentials) {
-            return this.convertToUserWarehouseCredentialsWithSecrets(
-                credentials,
-            );
+            const credentialsWithSecrets =
+                this.convertToUserWarehouseCredentialsWithSecrets(credentials);
+
+            // Validate Snowflake SSO credentials with Zod schema
+            // This ensures refreshToken is present and token field is not allowed
+            if (
+                credentialsWithSecrets.credentials.type ===
+                    WarehouseTypes.SNOWFLAKE &&
+                credentialsWithSecrets.credentials.authenticationType ===
+                    SnowflakeAuthenticationType.SSO
+            ) {
+                const result = snowflakeSsoUserCredentialsSchema.safeParse(
+                    credentialsWithSecrets.credentials,
+                );
+                if (!result.success) {
+                    throw new SnowflakeTokenError(
+                        `Please reauthenticate to access snowflake`,
+                    );
+                }
+            }
+
+            return credentialsWithSecrets;
         }
+
         return undefined;
     }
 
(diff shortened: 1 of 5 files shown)


rephus:
* **#18919** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18919?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
```diff
@@ -176,11 +177,33 @@ export class UserWarehouseCredentialsModel {
...
+            // Fixes backwards compatibility token for snowflake
+            // Without this, refreshCredentials will not have the refreshToken
+            // So it will try to fetch the refresh token from the user's OpenID table
+            // Instead of the user warehouse credentials table
+            if (
+                credentialsWithSecrets.credentials.type ===
+                WarehouseTypes.SNOWFLAKE
+            ) {
+                if (!credentialsWithSecrets.credentials.refreshToken) {
+                    throw new SnowflakeTokenError(
```
rephus, line 199: This will throw an error and force users to reauthenticate if snowflake user credentials is using `token` instead of `refreshToken` 

Image

packages/common/src/types/userWarehouseCredentials.ts
```diff
@@ -36,7 +36,12 @@ export type UserWarehouseCredentialsWithSecrets = Pick<
         | Pick<CreatePostgresCredentials, 'type' | 'user' | 'password'>
         | Pick<
               CreateSnowflakeCredentials,
-              'type' | 'user' | 'password' | 'authenticationType' | 'token'
+              | 'type'
+              | 'user'
+              | 'password'
+              | 'authenticationType'
+              | 'refreshToken'
+              | 'token'
```
rephus: I kept `token` here so we can check it here https://github.com/lightdash/lightdash/pull/18919/changes#diff-31e83ba5002599e953eb94d38816460bf2352d4651e2836c0757854a00e05592R198

packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
owlas: Nice! 

Can we update this to use zod or something which throws on ANY extra property? So we get a validated payload only by this point

rephus: Updated

Image

Image

packages/common/src/types/userWarehouseCredentials.ts
owlas: @rephus i can't see this anymore in this diff - i think it's not needed ?

packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
owlas: Awesome!! Only change is to leave the technical part out of the error message. Just:

"Please reauthenticate to access snowflake"

PR #18961, opened 2025-12-18
fix: add user session validation in getSpaces endpoint

pimmey:
Closes: PROD-2098 / LIGHTDASH-BACKEND-BFH

### Description:
Added proper error handling for missing user sessions in the ProjectController's getSpaces endpoint. Now, when a user session is not found, the API will throw an AuthorizationError with a clear message instead of potentially causing a runtime error. Also imported the AuthorizationError class to support this change.

### DIFF
--- a/packages/backend/src/controllers/projectController.ts
+++ b/packages/backend/src/controllers/projectController.ts
@@ -18,6 +18,7 @@ import {
     ApiSqlChartAsCodeUpsertResponse,
     ApiSqlQueryResults,
     ApiSuccessEmpty,
+    AuthorizationError,
     CalculateTotalFromQuery,
     ChartAsCode,
     CreateProjectMember,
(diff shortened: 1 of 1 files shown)


pimmey:
* **#18961** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18961?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18961?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pimmey:
Definitely would be cleaner, but the way our middleware is structured now -- checking for `if (req.account?.isAuthenticated() || req.user?.userUuid) {` -- it's not so obvious how to do that. That said, it's worth taking a deeper look into it, if we keep getting these ✅

PR #18995, opened 2025-12-19
fix: improve error handling for expired S3 cache results

pimmey:
Closes: GLITCH-110 / LIGHTDASH-BACKEND-AFB

### Description:
Improved error handling for expired results in S3 storage. Now properly throws and handles `NotFoundError` when results are not found or have expired, providing clearer error messages to users. Added specific handling for `NoSuchKey` and `NotFound` errors in the S3 cache client, and updated the scheduler to avoid sending expected `NotFoundError` instances to Sentry. Also fixed the export results component to properly hide loading notifications when errors occur.

### DIFF
--- a/packages/common/src/types/errors.ts
+++ b/packages/common/src/types/errors.ts
@@ -503,7 +503,7 @@ export class S3Error extends LightdashError {
 
 export class ResultsExpiredError extends LightdashError {
     constructor(
-        message = 'Your results have expired. Please refresh the page and try again.',
+        message = 'Your results have expired. Please refresh the page or re-run the query.',
     ) {
         super({
             message,


pimmey:
* **#18995** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/18995?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/scheduler/SchedulerTaskTracer.ts
```diff
@@ -291,12 +292,17 @@ export const traceTask = <T extends SchedulerTaskName>(
                             });
 
                             // Only capture to Sentry if this is the final attempt or it's not a retryable error
+                            // NotFoundError is expected (e.g., expired cache) and should not go to Sentry
                             const isRetryableError =
                                 e instanceof GoogleSheetsTransientError;
+                            const isExpectedError = e instanceof NotFoundError;
```
almeidabbm: Not found errors can be thrown in other scenarios, not just when the results file no longer exists 🤔 
Probably most `NotFoundErrors` should fall in this category, but wondering if there's any other we might want to get sentry visibility on

pimmey: You're right, that was a tad too broad. Narrowed down to a custom `ResultsExpiredError` :heavy_check_mark:

packages/common/src/types/errors.ts
```diff
@@ -529,6 +529,19 @@ export class S3Error extends LightdashError {
     }
 }
 
+export class ResultsExpiredError extends LightdashError {
+    constructor(
+        message = 'Your results have expired. Please refresh the page and try again.',
```
ZeRego: Why refresh the page? Can they rerun the query to make it work?

pimmey: Good point! But also it might happen on the chart view page where we don't have run query button. Maybe we could go for:
1. "Your results have expired. Please try again."
2. "Your results have expired. Please refresh or re-run the query."

PR #19033, opened 2025-12-23
fix: prevent pivot mismatch warning when groupByColumns is undefined

notgiorgi:
### Description:

Fixed a potential null reference error in the VisualizationWarning component by adding a null check for `resultsData.pivotDetails?.groupByColumns` before attempting to map over it. Also improved code readability by removing unnecessary optional chaining and array fallback.

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationWarning.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationWarning.tsx
@@ -62,16 +62,17 @@ const VisualizationWarning: FC<PivotMismatchWarningProps> = ({
 
     // Determine if configured pivot dimensions are different from the ones used to compute the results
     const shouldShowPivotMismatch = useMemo(() => {
-        if (!resultsData.pivotDetails?.groupByColumns) return false;
-
-        // Determine pivot used to compute current results
-        const resultsPivotDimensions =
-            resultsData.pivotDetails.groupByColumns.map(
-                (c: { reference: string }) => c.reference,
-            );
-
         // Only show when using SQL pivot results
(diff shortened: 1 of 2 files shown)


notgiorgi:
* **#19033** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19033?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19033?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/Explorer/VisualizationCard/VisualizationWarning.tsx
```diff
@@ -62,10 +62,14 @@ const VisualizationWarning: FC<PivotMismatchWarningProps> = ({
 
     // Determine if configured pivot dimensions are different from the ones used to compute the results
     const shouldShowPivotMismatch = useMemo(() => {
+        if (!resultsData?.pivotDetails?.groupByColumns) return false;
```
ZeRego: This will not display the warning when they are editing the charts. Example:  
1\. run a query with no pivot  
2\. update chart config to have pivot

Expected: warning for user to run the query again  
Actual: no warning

almeidabbm: @notgiorgi solution here could be to use `const validQueryArgs = useExplorerSelector(selectValidQueryArgs)` instead of the `dirtyPivotDimensions` and pass them to the `VisualizationWarning` component as `pivotDimensions` props instead of the chart config one

PR #19047, opened 2025-12-23
fix: conditional formatting w/ field references  for "metrics as rows"

notgiorgi:
Closes: [PROD-2154](https://linear.app/lightdash/issue/PROD-2154/column-totals-only-display-a-single-metric-when-show-metrics-as-rows)

### Description:

When `metricsAsRows: true`, metrics become rows instead of columns:

```
|                         | 2025-01/standard | 2025-01/overnight | 2024-06/standard |
|-------------------------|------------------|-------------------|------------------|
| Row 0: Total order amt  |     $461.85      |      $353.53      |     $410.80      |
| Row 1: Total completed  |     $160.00      |      $196.50      |     $227.50      |
```

This causes two problems:

### Problem A: Wrong Field ID

Column metadata is set using `baseId`:

```ts
const itemId = col.underlyingId || col.baseId || col.fieldId;
const item = itemId ? getField(itemId) : undefined;
```

When `metricsAsRows: true`, `baseId` for data columns is the **pivoted dimension** (e.g., `orders_shipping_method`), NOT the actual metric.

Result: ALL data columns have `itemId: "orders_shipping_method"` instead of the actual metric field ID.

### Problem B: Metric Values on Different Rows

Even if we fixed the field ID, comparing one metric to another requires cross-row access:

- Row 0 contains `orders_total_order_amount` values
- Row 1 contains `orders_total_completed_order_amount` values

When evaluating a cell on Row 1, the value from Row 0 is not available in `rowFields`.

### Example

Rule: "Highlight Total Completed when <= Total Order Amount"

Evaluating cell `$160.00` (Row 1, 2025-01/standard):

- `rowFields` only contains: `{ "orders_shipping_method": <some value> }`
- Lookup `rowFields["orders_total_order_amount"]` → **undefined**
- Rule silently fails - no highlighting applied

PR #19039, opened 2025-12-23
fix: conditional formatting in pivoted tables w/ field comparison

notgiorgi:
### Description:

When a table is pivoted, the same metric appears in **multiple columns** (one per pivot dimension value):

```
|                         | 2025-01/standard | 2025-01/overnight | 2024-06/standard |
|-------------------------|------------------|-------------------|------------------|
| Total order amount      |     $461.85      |      $353.53      |     $100.80      |
| Total completed amount  |     $160.00      |      $196.50      |     $27.50      |
```

### What Goes Wrong

When building `rowFields`, the code iterates through ALL columns left-to-right:

```ts
acc[getItemId(meta.item)] = { field, value }; // Same key for all columns with same metric!
```

Since multiple columns share the same `meta.item` (via `baseId`), each column **overwrites** the previous value:

1. Column 0: `rowFields["orders_total_order_amount"] = 461.85`
2. Column 2: `rowFields["orders_total_order_amount"] = 353.53` ← overwrites!
3. Column 4: `rowFields["orders_total_order_amount"] = 100.80` ← overwrites again!

Only the **last column's value** survives.

### Example

Rule: "Highlight Total Completed when <= Total Order Amount"

Evaluating cell `$160.00` (2025-01/standard column):

- **Should compare**: `$160.00 <= $461.85` (same pivot context) → TRUE ✓
- **Actually compares**: `$160.00 <= $100.80` (rightmost column!) → Different result

With many columns, the final value could be from a completely unrelated pivot context.

To reproduce:  
  
  
1\.

PR #19064, opened 2025-12-23
fix: conditional formatting for field references for metrics-as-rows

notgiorgi:
Closes: [PROD-2154](https://linear.app/lightdash/issue/PROD-2154/column-totals-only-display-a-single-metric-when-show-metrics-as-rows)

### Description:

When `metricsAsRows: true`, metrics become rows instead of columns:

```
|                         | 2025-01/standard | 2025-01/overnight | 2024-06/standard |
|-------------------------|------------------|-------------------|------------------|
| Row 0: Total order amt  |     $461.85      |      $353.53      |     $410.80      |
| Row 1: Total completed  |     $160.00      |      $196.50      |     $227.50      |
```

This causes two problems:

### Problem A: Wrong Field ID

Column metadata is set using `baseId`:

```ts
const itemId = col.underlyingId || col.baseId || col.fieldId;
const item = itemId ? getField(itemId) : undefined;
```

When `metricsAsRows: true`, `baseId` for data columns is the **pivoted dimension** (e.g., `orders_shipping_method`), NOT the actual metric.

Result: ALL data columns have `itemId: "orders_shipping_method"` instead of the actual metric field ID.

### Problem B: Metric Values on Different Rows

Even if we fixed the field ID, comparing one metric to another requires cross-row access:

- Row 0 contains `orders_total_order_amount` values
- Row 1 contains `orders_total_completed_order_amount` values

When evaluating a cell on Row 1, the value from Row 0 is not available in `rowFields`.

### Example

Rule: "Highlight Total Completed when <= Total Order Amount"

Evaluating cell `$160.00` (Row 1, 2025-01/standard):

- `rowFields` only contains: `{ "orders_shipping_method": <some value> }`
- Lookup `rowFields["orders_total_order_amount"]` → **undefined**
- Rule silently fails - no highlighting applied

### DIFF
--- a/packages/frontend/src/components/common/PivotTable/index.tsx
+++ b/packages/frontend/src/components/common/PivotTable/index.tsx
@@ -32,6 +32,7 @@ import {
     getExpandedRowModel,
     useReactTable,
     type GroupingState,
+    type Row,
 } from '@tanstack/react-table';
 import { useVirtualizer } from '@tanstack/react-virtual';
 import isEqual from 'lodash/isEqual';
@@ -93,6 +94,18 @@ const VirtualizedArea: FC<{
         </Table.Row>
     );
 };
+
+type IndexDimension = { fieldId: string; value: unknown };
+
+const extractIndexDimensions = (
+    indexValueRow: PivotData['indexValues'][number],
+): IndexDimension[] =>
+    indexValueRow
+        .filter((v) => v.type === 'value')
+        .map((v) => ({
+            fieldId: v.fieldId,
+            value: 'value' in v ? v.value?.raw : undefined,
+        }));
 
 type PivotTableProps = BoxProps & // TODO: remove this
     React.RefAttributes<HTMLTableElement> & {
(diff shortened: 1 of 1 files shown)


notgiorgi:
* **#19064** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19064?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19064?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#19039** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19039?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#19047](https://github.com/lightdash/lightdash/pull/19047) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19047?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/frontend/src/components/common/PivotTable/index.tsx
```diff
@@ -110,6 +113,7 @@ type PivotTableProps = BoxProps & // TODO: remove this
 
 const PivotTable: FC<PivotTableProps> = ({
     data,
+    unpivotedRows,
```
almeidabbm: Can't we do the same with just the `pivotedData` we receive above? it should have all information needed, e.g. valuesColumns, etc.

Asking because `unpivotedRows` can be a couple different things - considering you're passing `resultsData.rows`:
- Backend pivoted rows (flagEnabled)
- Original rows with no pivot (flag disabled)

`pivotData` should be a unified data structure with all the information needed to process the rows in the frontend 🤔

notgiorgi: Will take a look thanks!

notgiorgi:
### Merge activity

* **Dec 24, 2:47 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/19064).
* **Dec 24, 2:50 PM UTC**: [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/19064) rebased this pull request as part of a merge.
* **Dec 24, 2:50 PM UTC**: @notgiorgi merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/19064).

Issue #19126, opened 2025-12-29
Improve inline label rendering in bar charts when there is insufficient space


tatianainama:
### Description

When there is insufficient space to render stacked bar labels inline, the chart configuration defaults to not rendering them:

<img width="1122" height="847" alt="Image" src="https://github.com/user-attachments/assets/d2f34bfe-5df3-4569-829f-4ed72dedc17f" />

<img width="359" height="610" alt="Image" src="https://github.com/user-attachments/assets/3b7b3861-ebd1-40ad-b180-6d161cefa7c3" />

We should find a way to improve this:
- Always render very small values (including 0) with minimal visual space, such as a thin line or minimal bar
  - Consider using transparency or a dashed background?
- Always render the label values, adjusting other section positioning if needed to accommodate them

### What problem does this solve?

This prevents confusion when sections are not rendered at all. When values are too small to display, they create the false impression that the section doesn't exist or has a value of zero.

PR #19148, opened 2025-12-30
feat: add option to show overlapping labels in stacked bar charts

pimmey:
Closes: #19126

### Description:
Added support for showing overlapping labels in stacked bar charts. This feature allows users to display labels even when they would normally be hidden due to space constraints.

The implementation:
- Adds a new `showOverlappingLabels` property to series label configuration
- Creates a dynamic label layout function that reduces font size for labels that don't fit in their segments
- Adds a checkbox in the Series configuration panel to toggle this feature for stacked bar charts
- Only shows the checkbox when stacked bar series are present in the chart

This improves data visibility in stacked bar charts by ensuring all labels can be seen, even in smaller segments.

<details>
<summary>Quick demo</summary>

### Acceptable
This is also the customer's case, with only 1 isolated bar segment.
| Off | On  |
|--------------------|---|
|  |  |

### Not so good
Many stacked ones. Needs either horizontal spacing/shifting or cascading, in case there are many labels. But really, there are so many cases where this will not work - if the label is very long for example. We always are limited by the dimensions of the chart and can only do so much about it. Leading lines could be a solution, but if the space is limited, where do we render those? 🤔 
| Off | On  |
|--------------------|---|
|  |  |
</details>

### DIFF
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -310,6 +310,7 @@ export type Series = {
     label?: {
         show?: boolean;
         position?: 'left' | 'top' | 'right' | 'bottom' | 'inside';
+        showOverlappingLabels?: boolean;
     };
     hidden?: boolean;
     areaStyle?: Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


pimmey:
* **#19148** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #19153, opened 2025-12-30
fix: logging individual scheduler targets when batch jobs are created

almeidabbm:
### Description:
Added a undefined check for target in the scheduler task to prevent errors when processing undefined targets. The code now safely skips processing when a target is undefined instead of attempting to log it, which could cause runtime errors.

**Why we need this check?**
After introducing batch jobs we log the notifications differently and return an explicit `undefined` target.
Without this check it is throwing: `error: Missing target for scheduler format {format}`

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -2966,8 +2966,12 @@ export default class SchedulerTask {
 
             // Create scheduled jobs for targets
             await Promise.all(
-                scheduledJobs.map(({ target, jobId: targetJobId }) =>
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#19153** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19153?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

magnew (approved):

PR #19166, opened 2025-12-30
fix: prevent screenshot ready indicator when query results are loading

pimmey:
### Description:
Fixed a bug in the MinimalSavedExplorer component where the ScreenshotReadyIndicator was being displayed even when query results were still loading.

<details>
<summary>How it went</summary>

### Debugging
Started by taking a look at the loading states - there are 2 spinners and a message about missing required parameters in the mix.

https://github.com/user-attachments/assets/39801e3a-c1ce-469d-9b84-07b1a3ad44f1

Also reproduced locally:
Screenshot 2025-12-30 at 16 55 44

OK, so the indicator would appear after the first "loading" is done. But then there's another when we resolve the parameters.

Yay, works (don't mind the temp debugging bar):
Screenshot 2025-12-30 at 16 58 20
</details>

### DIFF
--- a/packages/frontend/src/pages/MinimalSavedExplorer.tsx
+++ b/packages/frontend/src/pages/MinimalSavedExplorer.tsx
@@ -59,7 +59,10 @@ const MinimalExplorerContent = memo(() => {
     const savedChart = useExplorerSelector(selectSavedChart);
 
     const isLoadingQueryResults =
-        query.isFetching || queryResults.isFetchingRows;
+        query.isFetching ||
+        queryResults.isFetchingRows ||
+        !query.data?.queryUuid ||
+        queryResults.queryUuid !== query.data.queryUuid;
 
     const [isScreenshotReady, setIsScreenshotReady] = useState(false);
     const hasSignaledReady = useRef(false);
@@ -113,7 +116,7 @@ const MinimalExplorerContent = memo(() => {
                 </Box>
             </MantineProvider>
 
-            {isScreenshotReady && !isLoadingQueryResults && (
+            {isScreenshotReady && (
                 <ScreenshotReadyIndicator
                     tilesTotal={1}
                     tilesReady={1}


pimmey:
* **#19166** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19166?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/pages/MinimalSavedExplorer.tsx
```diff
@@ -112,7 +113,7 @@ const MinimalExplorerContent = memo(() => {
                 </Box>
             </MantineProvider>
 
-            {isScreenshotReady && (
+            {isScreenshotReady && !isLoadingQueryResults && (
```
almeidabbm: isn't `isScreenshotReady` already taking `isLoadingQueryResults` into account? 🤔

pimmey: It is, it's used in `useEffect` where we set `isScreenshotReady`. But I thought `isLoadingQueryResults` would only go `true` -> `false` once, which isn't the case. If it goes `true` -> `false` -> `true` ... again (like in the case with parameters), we need to guard it here additionally.

almeidabbm: I see 🤔 but then does this fix it? doesn't it happen that `isLoadingQueryResults` goes true -> false -> true which causes the render of the ready component and screenshot to be taken anyway

Wondering if the issue isn't the calculation of `isLoadingQueryResults` itself

pimmey: > but then does this fix it?

It does 😄 

> which causes the render of the ready component and screenshot to be taken anyway

Also true! I think there is a brief moment when the ready component renders, but then `isLoadingQueryResults ` changes again - so it is removed from DOM, before the screenshot is taken. So far it worked, but I can smell "race condition" in the air with this approach.

Let me see, if I can make it more reliable.

almeidabbm: Yeah, it's like even if it is a brief switch locally we can't be sure of it in a prod env. Therefore the `unfurlService` will see the component and stop waiting for it to be rendered and take a screenshot, even though it has unmounted and shows a loading spinner 💪

almeidabbm (approved):
Great debugging!!! 🚀

PR #19174, opened 2025-12-31
feat: allow choosing join keys in maps

magnew:
### Description:

Makes it possible to choose the field to use as the join key with area maps. Geo json typically has properties for each shape that include various keys, but there is not a lot of standardization around which properties to use to join to data. This PR makes it possible to chose which key to use to join data with an area on the map.

For example, a standard property object for Indonesia could look like
```
{ "name": "Indonesia", "ISO3166-1-Alpha-3": "IDN", "ISO3166-1-Alpha-2": "ID" }
```

This makes it possible to choose which column in your dataset contains the ISO3 value. 

In this PR, we have standard properties hardcoded for our built in maps
- For the world, you can choose an ISO-3 column
- For the US, you can choose a state code 

**Next:** chose the map prop AND field to link in custom maps. And look for automatic matches. 

**Demo**
Kapture 2026-01-05 at 18 50 12


magnew:
* **#19211** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19174** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19174?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19174?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#19210** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

magnew:
> When changing a join key the tooltips seem to still render with the previous one 🤔 If we re-run query it fixes it, but this is most likely just a re-rendering issue?
> 
>  Screen.Recording.2026-01-08.at.17.04.01.mov

🔥 Good catch. This thing is like, outside of the normal react pipeline so it needs a lot of manual refreshing

PR #19196, opened 2026-01-05
docs: add info about container host mapping

Spissable:
### Description:
While setting up the project locally, @almeidabbm and I noticed that a little info was missing about mapping `minio` to `localhost` in the hosts file.

I experimented a little bit with some settings, but editing the hosts file seems to be the best way in the end.


almeidabbm (approved):

almeidabbm:
@all-contributors please add @Spissable for docs

PR #19204, opened 2026-01-05
fix: log error instead of throwing when scheduler is disabled

ZeRego:
Closes: PROD-2258

### Description:

Improved error handling for disabled schedulers in the `handleScheduledDelivery` task. Instead of throwing an error when a scheduler is disabled, the task now logs the error and gracefully exits, preventing job failures in the queue.

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -2860,8 +2860,21 @@ export default class SchedulerTask {
                   );
 
         if (!scheduler.enabled) {
-            // This should not happen, if schedulers are not enabled, we should remove the scheduled jobs from the queue
-            throw new Error('Scheduler is disabled');
+            await this.schedulerService.logSchedulerJob({
+                task: SCHEDULER_TASKS.HANDLE_SCHEDULED_DELIVERY,
+                schedulerUuid,
+                jobId,
+                jobGroup: jobId,
+                scheduledTime,
+                status: SchedulerJobStatus.ERROR,
+                details: {
+                    error: `Scheduler is disabled, skipping scheduled delivery.`,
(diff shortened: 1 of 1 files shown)


ZeRego:
* **#19204** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -2860,8 +2860,21 @@ export default class SchedulerTask {
                   );
 
         if (!scheduler.enabled) {
-            // This should not happen, if schedulers are not enabled, we should remove the scheduled jobs from the queue
```
stevelikesmusic, line 2863: > This should not happen

Except when it does :sweat_smile:  
  
Why would this happen when we expect it to be unreachable?

stevelikesmusic (approved):

packages/backend/src/scheduler/SchedulerTask.ts
ZeRego: I haven't found any issue in our disable logic so I assume this is a race condition because the job deletion and disable are not in the same transaction.

```
        // Remove scheduled jobs, even if the scheduler is not enabled
        await this.schedulerClient.deleteScheduledJobs(schedulerUuid);
        await this.schedulerModel.deleteScheduledLogs(schedulerUuid);

        const scheduler = await this.schedulerModel.setSchedulerEnabled(
            schedulerUuid,
            enabled,
        );
```

PR #19209, opened 2026-01-05
fix: Update who's allowed to manage scheduled deliveries

Spissable:
### Description:
Scheduled deliveries shall only be editable by admins or owners of the scheduled delivery.

#### TODO:
- [x] project members
- [x] organization members
- [x] fix e2e tests (scheduler)
- [x] service accounts (needs further clarification)
- [x] UI changes (hide buttons)

Co-authored-by: bruno@lightdash.com

### DIFF
--- a/packages/backend/src/services/SchedulerService/SchedulerService.ts
+++ b/packages/backend/src/services/SchedulerService/SchedulerService.ts
@@ -163,8 +163,8 @@ export class SchedulerService extends BaseService {
         scheduler: Scheduler;
         resource: ChartSummary | DashboardDAO;
     }> {
-        // editors can "manage" their own scheduled deliveries,
-        // however, interactive users can only "create" scheduled deliveries
+        // admins can manage all scheduled deliveries,
+        // everyone below can only manage their own scheduled deliveries
         const scheduler = await this.schedulerModel.getScheduler(schedulerUuid);
         const resource = await this.getSchedulerResource(scheduler);
         const { organizationUuid, projectUuid } = resource;
--- a/packages/common/src/authorization/organizationMemberAbility.test.ts
+++ b/packages/common/src/authorization/organizationMemberAbility.test.ts
@@ -545,18 +545,6 @@ describe('Organization member permissions', () => {
                     ).toEqual(true);
                 });
 
-                it('can view all ScheduledDeliveries', () => {
-                    expect(
-                        ability.can(
-                            'view',
-                            subject('ScheduledDeliveries', {
-                                organizationUuid:
-                                    ORGANIZATION_ADMIN.organizationUuid,
-                            }),
-                        ),
-                    ).toEqual(true);
-                });
-
                 it('can manage all ScheduledDeliveries', () => {
                     expect(
                         ability.can(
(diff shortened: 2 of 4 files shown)


Spissable:
@almeidabbm we overlooked 1 detail so far:
Our interpretation was that the on/off toggle, the edit button and the delete button are the "edit operations". However, as it currently stands, the backend also checks the manage permission on the "send now" operations: https://github.com/lightdash/lightdash/blob/de4b88f64a7dc1df06e3e80f657304193a2db077/packages/backend/src/services/SchedulerService/SchedulerService.ts#L850

I think it's up to interpretation whether this makes sense or not, but in either case we need to update on side:
- we also hide the "send now" button when you're not allowed to manage
- or we allow the send operation in the backend


Edit: the agreement is to hide ALL the buttons to reflect the current permissions.

packages/backend/src/services/SchedulerService/SchedulerService.ts
```diff
@@ -163,10 +163,8 @@ export class SchedulerService extends BaseService {
         scheduler: Scheduler;
         resource: ChartSummary | DashboardDAO;
     }> {
-        // editors can "manage" scheduled deliveries,
-        // which means they can edit scheduled deliveries created from other users, even admins
-        // however, interactive users can only "create" scheduled deliveries,
-        // which means they can only edit their own scheduled deliveries
+        // editors can "manage" their own scheduled deliveries,
+        // however, interactive users can only "create" scheduled deliveries
```
almeidabbm: `interative_viewers` can also manage scheduled deliveries they own (same as before)

packages/common/src/authorization/projectMemberAbility.test.ts
```diff
@@ -1118,6 +1225,61 @@ describe('Project member permissions', () => {
...
+                });
+
+                it('cannot view all ScheduledDeliveries', () => {
+                    expect(
+                        ability.can(
+                            'view',
+                            subject('ScheduledDeliveries', { projectUuid }),
+                        ),
+                    ).toEqual(false);
+                });
```
almeidabbm: We can remove these `view` permissions checks since they're not used anywhere in the code

packages/common/src/authorization/organizationMemberAbility.test.ts
```diff
@@ -531,6 +531,68 @@ describe('Organization member permissions', () => {
...
+                    expect(
+                        ability.can(
+                            'view',
+                            subject('ScheduledDeliveries', {
+                                organizationUuid:
+                                    ORGANIZATION_ADMIN.organizationUuid,
+                            }),
+                        ),
+                    ).toEqual(true);
+                });
```
almeidabbm: We can remove these view permissions checks since they're not used anywhere in the code

packages/common/src/authorization/roleToScopeMapping.test.ts
```diff
@@ -127,9 +128,9 @@ describe('roleToScopeMapping', () => {
 
             // Define scopes that should be equivalent to an editor role
             // Based on projectMemberAbility.ts, an editor can:
-            // - Do everything interactive_viewer can do (view content, export, create deliveries)
+            // - Do everything interactive_viewer can do (view content, export, create deliveries & manage own deliveries)
             // - Create spaces
-            // - Manage non-private spaces, jobs, pinned items, scheduled deliveries, dashboard comments, tags
+            // - Manage non-private spaces, jobs, pinned items, dashboard comments, tags and view scheduled deliveries
```
almeidabbm: we can remove the `view` mentioned here

almeidabbm (approved):
Super job 🚀 
Pre-approved, there are a couple minor comments and tests regarding `view` permissions which we should remove to reduce confusion (`view` isn't checked anywhere in code) + we should let e2e tests pass

PR #19222, opened 2026-01-06
fix: take nullsFirst sort into account when pivoting in the backend

almeidabbm:
Closes: https://linear.app/lightdash/issue/GLITCH-130/sql-pivot-pipeline-does-not-correctly-sort-null-values

### Description:
Added support for NULLS FIRST/LAST in query sorting when using backend pivoting.

Key changes:
- Added `nullsFirst` property to `VizSortBy` type
- Implemented `getNullsFirstLast` static method in `PivotQueryBuilder` to generate appropriate SQL clauses
- Updated SQL generation to include NULLS FIRST/LAST clauses in ORDER BY statements
- Modified anchor CTEs to use LEFT JOINs to preserve rows with NULL values
- Added comprehensive test coverage for the new functionality

**Before**
[before.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/c2086961-5c05-47ea-adaa-fb0ea13f539b.mov)

**After**

[after.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/49ca1bfb-d86f-4ae3-b95e-d16c1a3b1a67.mov)

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
@@ -8,7 +8,6 @@ import {
     TimeFrames,
     VizSortBy,
     WarehouseSqlBuilder,
-    type CompiledDimension,
     type ItemsMap,
     type PivotConfiguration,
 } from '@lightdash/common';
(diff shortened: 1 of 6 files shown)


almeidabbm:
* **#19222** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19222?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
```diff
@@ -504,7 +539,7 @@ export class PivotQueryBuilder {
                             `g.${q}${col.reference}${q} = ${cteName}.${q}${col.reference}${q}`,
                     )
                     .join(' AND ');
-                joins.push(`JOIN ${cteName} ON ${joinConditions}`);
+                joins.push(`LEFT JOIN ${cteName} ON ${joinConditions}`);
```
magnew, line 542: You don't think this will cause any issues? Now we could potentially be keeping rows we weren't keeping before?

magnew (approved):
✅ This definitely works better in the main cases. I left a potentially dumb question just cause this is such a core code path. 

Should the join type be dependent on the sort type. Maybe not. I think the SQL produced looks good. But testing this made me agree with the users who want to be able to see the pivoted sql somewhere

packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts
almeidabbm: This should be fine because:
- Anchor values are only used for sorting - They appear in `ORDER BY` clauses for `DENSE_RANK()`, not in the final `SELECT`
- `NULL` anchor values are handled - The `nullsFirst` parameter controls `NULLS FIRST/LAST` in the sort (lines 287-289, 389-394)
- Anchor row/column sql is using `SELECT DISTINCT` so there will not be any duplicates

PR #19247, opened 2026-01-07
feat: add roleUuid to project memberships and improve permission checks

rephus:
Closes: #N/A

### Description

Before

Screenshot from 2026-01-07 14-07-34

After

Screenshot from 2026-01-07 15-48-46

There were a few issues that were preventing a  `member/viewer` on the org, to `create previews` using `custom roles`

#### first issue: 

> we were not copying the custom role uuid to the new project

Before
Screenshot from 2026-01-07 14-30-39

After

Screenshot from 2026-01-07 14-30-25

#### second issue: 

> casl checkcs were checking `projectuuid,organizationUuid` , and this is not true at the org level. 

It is possible this issue is present on other parts of the code, outside the `create preview` .

See comment for example.

#### Third issue: 

we were missing update@self scope for users to update their own previews. 

### changes:

This PR enhances project permissions by adding more granular access controls:

1. Adds a new `update:Project@self` scope that allows users to update projects they created themselves
2. Adds support for `roleUuid` in project memberships
3. Improves permission checks in ProjectService by:
   - Using more specific project attributes for authorization checks
   - Adding more descriptive error messages when permissions are denied
4. Fixes permission checks for managing YAML tags by requiring both 'manage' and 'view' permissions

These changes ensure that permissions are properly enforced while providing clearer feedback when access is denied.

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -1444,6 +1444,7 @@ export class ProjectModel {
         projectUuid: string,
         email: string,
         role: ProjectMemberRole,
+        roleUuid?: string,
     ): Promise<void> {
         try {
             const [project] = await this.database('projects')
@@ -1471,6 +1472,7 @@ export class ProjectModel {
             await this.database('project_memberships').insert({
                 project_id: project.project_id,
                 role,
+                role_uuid: roleUuid || null,
                 user_id: user.user_id,
             });
         } catch (error: AnyType) {
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -1739,16 +1739,23 @@ export class ProjectService extends BaseService {
         projectUuid: string,
         explores: (Explore | ExploreError)[],
     ): Promise<void> {
-        const { organizationUuid } = await this.projectModel.getSummary(
+        const project = await this.projectModel.getWithSensitiveFields(
             projectUuid,
         );
         if (
             user.ability.cannot(
                 'update',
-                subject('Project', { organizationUuid, projectUuid }),
+                subject('Project', {
+                    projectUuid,
+                    organizationUuid: project.organizationUuid,
+                    type: project.type,
+                    createdByUserUuid: project.createdByUserUuid,
+                }),
             )
         ) {
-            throw new ForbiddenError();
+            throw new ForbiddenError(
+                `User does not have permission to update project`,
+            );
         }
 
         // TODO: Do not hardcode CLI information here
@@ -1765,7 +1772,7 @@ export class ProjectService extends BaseService {
             userUuid: user.userUuid,
             projectUuid,
             context: 'cli',
-            organizationUuid,
+            organizationUuid: project.organizationUuid,
         });
     }
 
@@ -2072,13 +2079,15 @@ export class ProjectService extends BaseService {
                 'delete',
                 subject('Project', {
                     type: project.type,
-                    projectUuid: project.projectUuid,
                     organizationUuid: project.organizationUuid,
+                    projectUuid: project.projectUuid,
                     createdByUserUuid: project.createdByUserUuid,
                 }),
             )
         ) {
-            throw new ForbiddenError();
+            throw new ForbiddenError(
+                `User does not have permission to delete project`,
+            );
         }
 
         await this.projectModel.delete(projectUuid);
@@ -5611,6 +5620,7 @@ export class ProjectService extends BaseService {
                             previewProjectUuid,
                             projectAccess.email,
                             projectAccess.role,
+                            projectAccess.roleUuid,
                         ),
                 );
                 const insertGroupAccessPromises = groupAccesses.map(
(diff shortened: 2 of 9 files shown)


rephus:
* **#19247** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19247?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19247?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1444,6 +1444,7 @@ export class ProjectModel {
         projectUuid: string,
         email: string,
         role: ProjectMemberRole,
+        roleUuid?: string,
```
rephus, line 1447: first issue

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -1739,16 +1739,22 @@ export class ProjectService extends BaseService {
...
-        const { organizationUuid } = await this.projectModel.getSummary(
+        const project = await this.projectModel.getWithSensitiveFields(
             projectUuid,
         );
         if (
             user.ability.cannot(
                 'update',
-                subject('Project', { organizationUuid, projectUuid }),
+                subject('Project', {
+                    projectUuid,
```
rephus, line 1749: second and third issue

second: had organizationUuid, this is not true for a `member/viewer` at the org level 

third: we were not checking update@self (new scope)

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2073,12 +2079,13 @@ export class ProjectService extends BaseService {
...
                     type: project.type,
                     projectUuid: project.projectUuid,
-                    organizationUuid: project.organizationUuid,
                     createdByUserUuid: project.createdByUserUuid,
                 }),
             )
         ) {
-            throw new ForbiddenError();
+            throw new ForbiddenError(
+                `User does not have permission to delete project`,
```
rephus, line 2089: Improve error copy on some errors

Before

Image

After


Image

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -6654,11 +6654,14 @@ export class ProjectService extends BaseService {
                 'update',
                 subject('Project', {
                     projectUuid,
-                    organizationUuid,
+                    type: project.type,
```
rephus: second and third issue

packages/common/src/authorization/roleToScopeMapping.ts
```diff
@@ -76,6 +76,7 @@ const BASE_ROLE_SCOPES = {
         'create:Project@preview', // Preview projects
         'delete:Project@self', // Preview projects created by user
         'update:Project',
+        'update:Project@self',
```
rephus, line 79: third issue

packages/common/src/authorization/projectMemberAbility.ts
```diff
@@ -220,6 +220,10 @@ export const projectMemberAbilities: Record<
         can('update', 'Project', {
             projectUuid: member.projectUuid,
         });
+        can('update', 'Project', {
```
rephus, line 224: ensure parity with new scope

packages/backend/src/services/ProjectService/ProjectService.ts
rephus: removing `organizationUuid`  could be problematic if we had the same projectUuid under different organizationUuids, but since that is not possible, it should be safe.

packages/backend/src/services/ProjectService/ProjectService.ts
stevelikesmusic: This doesn't really fix the issue. Trying this in the preview env, selecting `Update Project Self` still failed this check for `lightdash deploy`. `Update Project` succeeded though—after I removed the Self flavor. I'm wondering if the createdBy check is looking at who created the parent project—not the preview.

If I add `Update Project` this works because we only add {projectUuid} in that check. CASL ignores `type` and `createdBy` if they don't exist in the ability condition.

This also still works with org roles that can update the project because they also don't add the createdBy condition. 

This doesn't break anything, but does steer someone to think that it needs @self, when that modifier won't pass this check.

rephus: This is how I tested it: 

- create a member on the org
- create a custom role that inherits from editor 
- add create preview, delete@self,  update@self
- As the new user, do `lightdash preview`

lightdash deploy will still not work if the user didn't create the project, or it is not a preview, right ? 

This is needed, not to access the parent project, but for the replaceProjectParameters executed on the preview

stevelikesmusic: I tested by cloning a `Viewer` role. If I clone Editor, I don't even need `update Project`oddly enough.

packages/common/src/authorization/scopes.ts
```diff
@@ -261,6 +274,7 @@ const scopes: Scope[] = [
         group: ScopeGroup.PROJECT_MANAGEMENT,
         getConditions: (context) => [
             {
+                projectUuid: context.projectUuid,
```
stevelikesmusic, line 277: Nice!

packages/cli/src/handlers/deploy.ts
```diff
@@ -110,8 +111,32 @@ export const deploy = async (
         options.projectUuid,
     );
 
-    await replaceProjectYamlTags(options.projectUuid, lightdashProjectConfig);
-    await replaceProjectParameters(options.projectUuid, lightdashProjectConfig);
+    // These two methods are not critical to the deployment process, so we can ignore errors and show warnings instead
```
stevelikesmusic, line 114: Very cool. This means they don't need `manage:Tags` and also `update:Project` or just `manage:Tags`?

rephus:
apparently unrelated e2e error  
image

Issue #19214, opened 2026-01-06
Emoji in tile title is not rendered in scheduled delivery screenshot


Spissable:
### Description

When the title of a tile contains an emoji, the emoji isn't rendered properly in a scheduled delivery screenshot:

<img width="1400" height="660" alt="Image" src="https://github.com/user-attachments/assets/f6f54913-33bf-4b86-bca2-f5da604aa8f0" />

### Steps to Reproduce the Bug or Issue

1. Create a dashboard containing an emoji in the title (e.g. 🥸)
2. Create a scheduled delivery using the `Image` format
3. Trigger the scheduled delivery via `Send now`

Note: It doesn't seem to be affecting every emoji. So far the one confirmed broken one is 🥸.

### version

v0.2290.1

### Cloud or self-hosting

cloud

PR #19250, opened 2026-01-07
chore: update headless browser to chromium 2.38.2

Spissable:
Closes: https://github.com/lightdash/lightdash/issues/19214

### Description:
After inspecting the headless-browser docker container I noticed that it was fairly outdated and therefore stuck with some older fonts (e.g. fonts-noto-color-emoji from September 2020).
I re-tested the issue with the most up-to-date version and the emoji started rendering:

image

P.S. I figured it's more convenient that the docker-compose files reference our own dockerfile that we already have in place, this way we only need to adjust the chromium version in 1 place.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -125,7 +125,8 @@ services:
             - db-data:/var/lib/postgresql/data
 
     headless-browser:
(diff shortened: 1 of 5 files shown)


almeidabbm:
@all-contributors please add @Spissable for code

PR #19277, opened 2026-01-08
feat: allow changing date/timestamp dimensions' formatting in results table

joaoviana:
Relates to: [PROD-904](https://linear.app/lightdash/issue/PROD-904/formatting-date-type-dimensionsmetrics-in-the-results-table)

> [!IMPORTANT]
> Wanted to only do it just for date/timestamp dimensions first, so we agree on implementation. A PR on top of this one would cover metrics of date type, like MIN/MAX of a date dimension.

### Description:

This PR adds support for dimension formatting overrides, allowing users to customize how dimension values are displayed in charts and tables. The implementation includes:

- Added a new `dimensionOverrides` field to the metric query model
- Created a database migration to add the `dimension_overrides` column to saved queries versions
- Updated controllers to pass dimension overrides to the query service
- Enhanced the format modal to work with both metrics and dimensions
- Improved the format form with date/timestamp specific formatting options
- Added support for custom date formats with examples and preview

This feature is particularly useful for date and timestamp dimensions, allowing users to customize date formats (e.g., "dd/mm/yyyy", "mmmm d, yyyy") directly from the UI without having to modify the underlying data model.

### DIFF
--- a/packages/backend/src/database/migrations/20260108110951_add_dimension_overrides_column_to_saved_queries_versions.ts
+++ b/packages/backend/src/database/migrations/20260108110951_add_dimension_overrides_column_to_saved_queries_versions.ts
@@ -1,0 +1,16 @@
+import { Knex } from 'knex';
+
+const SavedQueriesVersionsTableName = 'saved_queries_versions';
+const DimensionOverridesColumnName = 'dimension_overrides';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(SavedQueriesVersionsTableName, (table) => {
+        table.jsonb(DimensionOverridesColumnName).nullable();
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(SavedQueriesVersionsTableName, (table) => {
+        table.dropColumn(DimensionOverridesColumnName);
+    });
+}
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -718,7 +718,7 @@ export function convertCustomFormatToFormatExpression(
         case CustomFormatType.ID:
         case CustomFormatType.DATE:
         case CustomFormatType.TIMESTAMP: {
-            // No implementation yet
+            // No format expression needed
             break;
         }
         default: {
--- a/packages/frontend/src/components/Explorer/FormatForm/index.tsx
+++ b/packages/frontend/src/components/Explorer/FormatForm/index.tsx
@@ -1,19 +1,40 @@
 import {
     CompactConfigMap,
     CustomFormatType,
+    DimensionType,
+    MetricType,
     NumberSeparator,
-    applyCustomFormat,
+    TableCalculationType,
     convertCustomFormatToFormatExpression,
     currencies,
     findCompactConfig,
+    formatValueWithExpression,
     getCompactOptionsForFormatType,
     type CustomFormat,
 } from '@lightdash/common';
-import { Anchor, Flex, Select, Stack, Text, TextInput } from '@mantine-8/core';
+import {
+    Anchor,
+    Box,
+    Flex,
+    Group,
+    Paper,
+    SegmentedControl,
+    Select,
+    Stack,
+    Text,
+    TextInput,
+} from '@mantine-8/core';
 import { NumberInput } from '@mantine/core';
 import { type GetInputProps } from '@mantine/form/lib/types';
+import {
+    IconCalendar,
+    IconClockHour4,
+    IconExternalLink,
+} from '@tabler/icons-react';
 import { useMemo, type FC } from 'react';
 import { type ValueOf } from 'type-fest';
+import MantineIcon from '../../common/MantineIcon';
+import { PolymorphicPaperButton } from '../../common/PolymorphicPaperButton';
 
 type Props = {
     formatInputProps: (
@@ -24,9 +45,10 @@ type Props = {
         path: keyof CustomFormat,
         value: ValueOf<CustomFormat>,
     ) => void;
+    itemType?: DimensionType | MetricType | TableCalculationType;
 };
 
-const formatTypeOptions = [
+const numericFormatTypeOptions = [
     CustomFormatType.DEFAULT,
     CustomFormatType.PERCENT,
     CustomFormatType.CURRENCY,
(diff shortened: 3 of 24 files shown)


joaoviana:
* **#19277** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19277?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/database/migrations/20260108110951_add_dimension_overrides_column_to_saved_queries_versions.ts
```diff
@@ -0,0 +1,16 @@
...
+    await knex.schema.alterTable(SavedQueriesVersionsTableName, (table) => {
+        table.jsonb(DimensionOverridesColumnName).nullable();
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(SavedQueriesVersionsTableName, (table) => {
+        table.dropColumn(DimensionOverridesColumnName);
+    });
+}
```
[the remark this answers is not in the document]
joaoviana: hmm ignoring this

packages/common/src/utils/formatting.ts
ZeRego, line 1: Did you get errors when using the existing code? because it should fall in the "case CustomFormatType.CUSTOM:" and the other cases is to handle the other explicit format types/options.

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -68,6 +69,10 @@ export const formatCellContent = (
 ) => {
     if (!data) return '-';
 
+    // // Check if item has formatOptions (from metric/dimension overrides)
+    // const hasFormatOptions =
+    //     item && 'formatOptions' in item && item.formatOptions !== undefined;
+
```
ZeRego: todo: delete comments

packages/frontend/src/hooks/useColumns.tsx
```diff
@@ -508,18 +528,24 @@ export const useColumns = (): TableColumn[] => {
...
-                        // For DATE and TIMESTAMP types, use the pre-formatted value from backend
-                        // to avoid timezone issues when re-parsing UTC date strings on the frontend
+                        // Check if item has formatOptions (from metric/dimension overrides)
+                        const hasFormatOptions =
+                            currentItem &&
+                            'formatOptions' in currentItem &&
+                            currentItem.formatOptions !== undefined;
+
+                        // For DATE and TIMESTAMP types WITHOUT format overrides,
+                        // use the pre-formatted value from backend to avoid timezone issues
```
ZeRego: Have you confirmed there are no timezone issues when using formatOptions?

packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
```diff
@@ -92,6 +93,15 @@ const ContextMenu: FC<ContextMenuProps> = ({
                 <ColumnHeaderSortMenuOptions item={item} sort={sort} />
 
                 <Menu.Divider />
+
+                {/* Format menu for date/timestamp dimensions */}
+                {isTimeBasedDimension(item) && (
```
ZeRego: extra points if we support overrides on numeric dimensions

packages/frontend/src/components/Explorer/FormatModal/index.tsx
```diff
@@ -51,6 +59,29 @@ export const FormatModal = memo(() => {
...
+        [dispatch],
+    );
+
+    // Determine if the item is a date/timestamp field
+    const isDateField = useMemo(() => {
+        if (!item) return false;
+        const type = getItemType(item);
+        return [
+            'date',
+            'timestamp',
```
ZeRego: Why do we need the hardcoded values?

packages/frontend/src/components/Explorer/FormatModal/index.tsx
```diff
@@ -129,13 +181,15 @@ export const FormatModal = memo(() => {
...
             }
         >
-            <form id="format-metric-form" onSubmit={handleOnSubmit}>
+            <form id="format-field-form" onSubmit={handleOnSubmit}>
                 <FormatForm
                     formatInputProps={getFormatInputProps}
                     format={form.values.format}
                     setFormatFieldValue={setFormatFieldValue}
+                    isDateField={isDateField}
+                    itemType={item ? getItemType(item) : undefined}
```
ZeRego, line 170: Since we are passing the itemType, should the isDateField check be in FormatForm?

ZeRego:
@joaoviana You don't need to address here, but fyi: CSV exports currently enforce a format for date/timestamp cells. https://github.com/lightdash/lightdash/blob/2e9d8f32f4eff59c8e239ee91aca168360533963/packages/backend/src/services/CsvService/CsvService.ts#L251

(both dim and metric have format overrides)
CleanShot 2026-01-08 at 16 54 32@2x

I know for googlesheets it is possible to set the raw value and configure the cell format separately, but not for CSVs.

packages/common/src/utils/formatting.ts
joaoviana: reverted my changes and custom-format support works as expected! Ty

packages/frontend/src/hooks/useColumns.tsx
joaoviana: done

packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
joaoviana: will do on a separate PR 👍

packages/frontend/src/components/Explorer/FormatModal/index.tsx
joaoviana: not needed, simplified!

packages/frontend/src/components/Explorer/FormatModal/index.tsx
joaoviana: good shout, logic is done in `FormatForm`

joaoviana:
> @joaoviana You don't need to address here, but fyi: CSV exports currently enforce a format for date/timestamp cells.
> 
> https://github.com/lightdash/lightdash/blob/2e9d8f32f4eff59c8e239ee91aca168360533963/packages/backend/src/services/CsvService/CsvService.ts#L251
> 
> I know for googlesheets it is possible to set the raw value and configure the cell format separately, but not for CSVs.

will keep that in mind 👍

joaoviana:
Made all changes, addressing/checking the timezone issue now @ZeRego. Will put it back to review once I get a clearer picture.

packages/frontend/src/hooks/useColumns.tsx
joaoviana: ah! Now that we don't do extra/different logic in `useColumns`, the format will happen from the backend, so it will fall into the case

```ts

                        // For DATE and TIMESTAMP types, use the pre-formatted value from backend
                        // to avoid timezone issues when re-parsing UTC date strings on the frontend
                        if (
                            isField(currentItem) &&
                            (currentItem.type === DimensionType.DATE ||
                                currentItem.type === DimensionType.TIMESTAMP)
                        ) {
                            return cellValue.value.formatted;
                        }

```

packages/frontend/src/components/Explorer/FormatForm/index.tsx
```diff
@@ -92,140 +175,308 @@ export const FormatForm: FC<Props> = ({
...
+
+    // Generate preview for date formats
+    const datePreview = useMemo(() => {
+        if (!isDateType) return null;
+        if (!format.custom) return null;
+        try {
+            return applyCustomFormat(sampleDate, {
+                type: CustomFormatType.CUSTOM,
+                custom: format.custom,
+            });
```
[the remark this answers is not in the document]
joaoviana: applied this change @ZeRego 

only change pushed just now was this to improve the "preview" results

Screenshot 2026-01-09 at 10 31 49
Screenshot 2026-01-09 at 10 31 45

joaoviana:
applied this change @ZeRego 

only change pushed just now was this to improve the "preview" results

Screenshot 2026-01-09 at 10 31 49
Screenshot 2026-01-09 at 10 31 45

Issue #19268, opened 2026-01-08
Allow partial support for explores with field errors by disabling problematic metrics


rephus:
Currently, if an explore contains a field with errors, the entire explore fails to load, making it difficult to test or use other working parts of the explore. 

<img width="677" height="448" alt="Image" src="https://github.com/user-attachments/assets/fac07477-01bb-436a-bd99-b34878db9ebd" />

**Proposed solution:**
- Update the explore compiler to be more flexible:
  - If an explore contains fields with errors, disable only the problematic metrics/fields instead of failing the entire explore.
  - The rest of the explore should remain accessible and usable.
- This behavior should be controlled by an ENV variable, so it can be enabled or disabled as needed.

**Why this is needed:**
- Makes it easier to test and use explores, even if some fields are broken.
- Improves the developer and user experience by not blocking access to otherwise functional explores.

**Suggested implementation steps:**
- Update the explore compiler logic to skip/disable fields with errors.
- Add an ENV variable to toggle this behavior.
- Ensure clear UI indication for disabled metrics/fields.

**Label:**
- Customer Support
- Feature
- ⚙️ backend
- ✍️ needs spec

**Milestone:**
- 💜 Customer Support

PR #19278, opened 2026-01-08
feat: add partial compilation mode for fields with errors

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/19268

Before

Screenshot from 2026-01-08 15-48-22

After
image

image

### Description:
Adds a new partial compilation feature that allows explores to load even when some fields fail to compile. When enabled, fields with compilation errors are marked with a `compilationError` property instead of causing the entire explore to fail.

This feature is controlled by a new configuration option `partialCompilation.enabled` which can be set via the `PARTIAL_COMPILATION_ENABLED` environment variable.

The implementation:
- Adds the new configuration option to `LightdashConfig`
- Updates the `ExploreCompiler` to handle field compilation errors gracefully
- Passes the partial compilation flag through the project adapter chain
- Adds error handling for dimensions, metrics, and joins that fail to compile

### DIFF
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -987,6 +987,14 @@ export type LightdashConfig = {
     editYamlInUi: {
         enabled: boolean;
     };
+    /**
+     * When enabled, fields that fail to compile will be marked with a
+     * compilationError instead of causing the entire explore to fail.
+     * This allows users to still access other fields in the explore.
+     */
+    partialCompilation: {
+        enabled: boolean;
+    };
 };
 
 export type SlackConfig = {
@@ -1762,6 +1770,9 @@ export const parseConfig = (): LightdashConfig => {
         },
         editYamlInUi: {
             enabled: process.env.EDIT_YAML_IN_UI_ENABLED === 'true',
+        },
+        partialCompilation: {
+            enabled: process.env.PARTIAL_COMPILATION_ENABLED === 'true',
         },
     };
 };
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -128,7 +128,8 @@ type RawSummaryRow = {
     tags: Explore['tags'];
     groupLabel: Explore['groupLabel'] | null;
     type: Explore['type'] | null;
-    errors: ExploreError['errors'] | null;
+    errors: ExploreError['errors'] | null; // Fatal errors from ExploreError
+    warnings: Explore['warnings'] | null; // Non-fatal warnings from partial compilation
     baseTable: Explore['baseTable'];
     baseTableDatabase: Explore['tables'][string]['database'];
     baseTableSchema: Explore['tables'][string]['schema'];
@@ -1135,6 +1136,7 @@ export class ProjectModel {
                     explore->'groupLabel' as "groupLabel",
                     explore->'type' as type,
                     explore->'errors' as errors,
+                    explore->'warnings' as warnings,
                     explore->'baseTable' as "baseTable",
                     explore->'tables'->(explore->>'baseTable')->>'database' as "baseTableDatabase",
                     explore->'tables'->(explore->>'baseTable')->>'schema' as "baseTableSchema",
@@ -1157,7 +1159,8 @@ export class ProjectModel {
             type: row.type ?? undefined,
             baseTableRequiredAttributes:
                 row.baseTableRequiredAttributes ?? undefined,
-            ...(row.errors ? { errors: row.errors } : {}), // Only add errors key if errors are present
+            ...(row.errors ? { errors: row.errors } : {}), // Fatal errors from ExploreError
+            ...(row.warnings ? { warnings: row.warnings } : {}), // Non-fatal warnings from partial compilation
         }));
     }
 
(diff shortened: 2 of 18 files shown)


rephus:
* **#19278** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19278?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19278?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/config/parseConfig.ts
```diff
@@ -1760,5 +1768,8 @@ export const parseConfig = (): LightdashConfig => {
         editYamlInUi: {
             enabled: process.env.EDIT_YAML_IN_UI_ENABLED === 'true',
         },
+        partialCompilation: {
+            enabled: process.env.PARTIAL_COMPILATION_ENABLED === 'true',
```
rephus, line 1775: new env variable for the backend

packages/cli/src/handlers/compile.ts
```diff
@@ -95,6 +95,7 @@ const getExploresFromLightdashYmlProject = async (
         warehouseSqlBuilder,
         lightdashProjectConfig,
         disableTimestampConversion,
+        process.env.PARTIAL_COMPILATION_ENABLED === 'true',
```
rephus, line 98: we also use ENV variable on the CLI (we could replace it with an --arg or automatically add it on `--ignore-errors` )

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -138,24 +159,46 @@ export class ExploreCompiler {
...
+                      });
+                      return false;
+                  }
+                  return true;
+              })
+            : joinedTables;
+
+        // Validate all joined tables exist (only when partial compilation is disabled)
+        if (!this.options.allowPartialCompilation) {
+            joinedTables.forEach((join) => {
```
rephus, line 189: existing behaviour for ! allow partial completion

packages/common/src/compiler/exploreCompiler.ts
```diff
@@ -113,11 +116,29 @@ const getReferencedTable = (
...
+
+    constructor(
+        warehouseClient: WarehouseSqlBuilder,
+        options: ExploreCompilerOptions = {},
+    ) {
         this.warehouseClient = warehouseClient;
+        this.options = options;
     }
 
     compileExplore({
```
rephus, line 144: this is where we check if allowPartialCompilation is enabled, and return a list of warnings on the explore, instead of throwing an error and generating a exploreError. 

Parts of this code could be simplified/unified if we get rid of the ENV variable and enable this by default.

packages/common/src/types/explore.ts
```diff
@@ -59,6 +59,21 @@ export enum ExploreType {
     DEFAULT = 'default',
 }
 
+export enum InlineErrorType {
+    // Fatal error types (ExploreError - explore is broken)
+    METADATA_PARSE_ERROR = 'METADATA_PARSE_ERROR',
+    NO_DIMENSIONS_FOUND = 'NO_DIMENSIONS_FOUND',
+    // Warning types (Explore.warnings - explore still usable)
+    SKIPPED_JOIN = 'SKIPPED_JOIN',
```
rephus, line 67: extened existing type to include warnings

packages/common/src/types/explore.ts
```diff
@@ -81,23 +96,23 @@ export type Explore = {
...
-
-export type InlineError = {
-    type: InlineErrorType;
-    message: string;
+    /**
+     * Non-fatal warnings from partial compilation.
+     * Present when some joins or fields failed to compile but the explore is still usable.
+     * Only populated when PARTIAL_COMPILATION_ENABLED=true.
+     */
+    warnings?: InlineError[];
```
rephus, line 104: reuse InlineError type, we could make an InlineWarning instead, it doesn't have much impact at the moment.

This warnings on explore goes then to the `warnings` field in the exploreSummary (instead of overloading errors )

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -128,7 +128,8 @@ type RawSummaryRow = {
     tags: Explore['tags'];
     groupLabel: Explore['groupLabel'] | null;
     type: Explore['type'] | null;
-    errors: ExploreError['errors'] | null;
+    errors: ExploreError['errors'] | null; // Fatal errors from ExploreError
+    warnings: Explore['warnings'] | null; // Non-fatal warnings from partial compilation
```
rephus, line 132: added new field for explore summaries. At the beggining, I thought about reusing "errors", but eventually it gets more difficult to  process in the frontend, when you have a mix of exploreErrors and explore with warnings (so I rolled back and refactored the code). So this feels more scalable and error-prone

packages/frontend/src/components/Explorer/ExploreSideBar/ExploreNavLink.tsx
```diff
@@ -144,6 +150,64 @@ const ExploreNavLink: React.FC<ExploreNavLinkProps> = ({
...
+                withinPortal
+                position="right"
+                withArrow
+                radius="md"
+                shadow="subtle"
+                variant="xs"
+            >
+                <HoverCard.Target>
+                    <Box>
+                        <NavLink
```
ZeRego: Almost an exact duplicate of the existing return. Could the hovercard be just on the rightSection to avoid duplicating NavLink? 
What happens if I try to use a field that has a warning?

rephus:
error looks unrelated 
image

PR #19269, opened 2026-01-08
fix: use null coalescing check for useSqlPivotResults feature flag

almeidabbm:
### Description:
Makes `useSqlPivotResults` config property optional by:
- Changing the type from `boolean` to `boolean | undefined` in the `LightdashConfig` type
- Updating the config parsing to set the value to `undefined` when the environment variable is not set
- Modifying the feature flag logic to use the nullish coalescing operator (`??`) instead of logical OR (`||`) to properly handle the undefined case

This fixes an issue where we were always falling back to posthog for the pivot sql flag when it was `false`

test-frontend
test-backend

PR #19284, opened 2026-01-08
fix: groups enabled feature flag checking

almeidabbm:
### Description:
Updated the `groups.enabled` configuration to be optional by changing its type from `boolean` to `boolean | undefined`. 

The environment variable `GROUPS_ENABLED` now sets the value only when explicitly defined, otherwise it remains `undefined`. 

Updated the feature flag logic to use the nullish coalescing operator (`??`) instead of logical OR (`||`) when checking if groups are enabled.

This fixes the fact that we were always falling back to `posthog` even when the environment variable was explicitly set to false

### DIFF
--- a/packages/backend/src/config/parseConfig.test.ts
+++ b/packages/backend/src/config/parseConfig.test.ts
@@ -573,3 +573,16 @@ test('should set useSqlPivotResults only when the environment variable is set', 
     const falseConfig = parseConfig();
     expect(falseConfig.query.useSqlPivotResults).toBe(false);
 });
+
+test('should set groups.enabled only when the environment variable is set', () => {
+    const undefinedConfig = parseConfig();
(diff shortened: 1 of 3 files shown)


almeidabbm:
* **#19284** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19284?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19284?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#19269** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Issue #19215, opened 2026-01-06
Loom thumbnail is not rendered in scheduled delivery screenshot


Spissable:
### Description

A loom video thumbnail isn't rendered properly in a scheduled delivery screenshot:

<img width="1400" height="660" alt="Image" src="https://github.com/user-attachments/assets/f6f54913-33bf-4b86-bca2-f5da604aa8f0" />

Expected behavior:
- Thumbnail is shown, otherwise there is just a title with nothing else

### Steps to Reproduce the Bug or Issue

1. Add a loom video to a dashboard
2. Create and fire a scheduled delivery

### version

v0.2290.1

### Cloud or self-hosting

cloud

PR #19318, opened 2026-01-09
fix: load loom thumbnail for screenshot functionality

Spissable:
Closes: https://github.com/lightdash/lightdash/issues/19215

### Description:
Loom videos are embedded via an iframe. This iframe doesn't show up at all in the screenshots (e.g. scheduled delivery preview).
I tried various different approaches with delays, waiting for the iframe, disabling some security flags in the headless-browser etc, but none of them worked.

In the end, the solution here feels a bit overkill but it's the only one I got working. In essence, it will display a minimal loom tile which is basically just the thumbnail image. To get this image, I had to implement a little proxy in the backend, as the image doesn't have a predictable filename (it adds a hash at the end).

#### Example screenshot:
image

### DIFF
--- a/packages/common/src/types/api.ts
+++ b/packages/common/src/types/api.ts
@@ -36,6 +36,7 @@ import {
     type ApiGetComments,
 } from './api/comments';
 import { type Email } from './api/email';
+import { type LoomThumbnailResult } from './api/loom';
 import {
     type ApiGetProjectParametersListResults,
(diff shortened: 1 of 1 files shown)


Spissable:
The main issue was actually that the browserless container was throwing some crypto error when loading the iframe. We managed to get a hacky workaround working. closing this pr and trying the other approach again

PR #19352, opened 2026-01-12
chore: add organizationId and projectId to scheduler events

Spissable:
### Description:
In an effort to improve the internal observability, we decided to add a few extra fields in the scheduler tracking events:
- `userId`
- `organizationId`
- `projectId`

Note: initially I started by adding it to the sub-properties, but finally decided that it might make more sense to have it into the root Track, especially since the `userId` was already there. However, we can easily revert back by reverting the last commit.

Special note:
I noticed that a bunch of extra props were being passed in some of the `SchedulerNotificationJobEvent` events, but the type didn't include them yet, so I added them.

In an ideal world, I would even go further and change the base `properties` field to `unknown` [here](https://github.com/lightdash/lightdash/blob/400ccb2a5b4d6b58b285f21420b73f185bec07f6/packages/backend/src/%40types/rudder-sdk-node.d.ts#L31). This would make things much more strict and avoid the issue mentioned above. But it would also require some changes in the [trackAccount](https://github.com/lightdash/lightdash/blob/400ccb2a5b4d6b58b285f21420b73f185bec07f6/packages/backend/src/analytics/LightdashAnalytics.ts#L1698) function.


Spissable:
> This looks good, but I would prefer if we kept consistency across `events` 🤔 wdyt?
> 
> To do that we should have:
> 
>     * `userId` at the root
> 
>     * `organizationId` and `projectId` in properties

Not sure tbh. To me it feels like these properties belong together, being generic information, rather than being very event specific. The comment [here](https://github.com/lightdash/lightdash/blob/400ccb2a5b4d6b58b285f21420b73f185bec07f6/packages/backend/src/analytics/LightdashAnalytics.ts#L1717) also suggested perhaps re-thinking the old approach.

So while I definitely favour consistency, maybe it's the right time to move these properties to the root and (slowly) adopt that across the code-base? I'd understand if that's maybe too ambitious though 😅

Issue #19312, opened 2026-01-09
Add a validator warning for pivoted charts with extra dimensions


magnew:
### Description

Add a validator warning for pivoted charts with extra dimensions. We already show a warning when creating the chart, but people are going to want to be able to find existing charts with this problem. 

### What problem does this solve?

It will make it possible to find existing charts with incorrect pivot config.

PR #19357, opened 2026-01-12
feat: add validation errors for extra fields

magnew:
Closes: #19312, GLITCH-136

### Description:

Adds a validator error for charts with extra dimensions that may cause incorrect rendering with the new pivoting pipeline. 
- There is a new validator rule and it will show an error for each extra dimension in charts
- The new validation only runs when `USE_SQL_PIVOT_RESULTS` is enabled
- For now, the fix button is hidden - there are a couple options for how that could work and I will follow up, but this part is always necessary
- Works for the CLI and UI

**Note**
This is the first time we are adding validation for chart config. All of the other validations have to do with project compilation. Does anyone see any issues with this? I'm not 100% sure how customers use the validator. If we need to we could make it optional to validate these. 

**UI**
Screenshot 2026-01-12 at 16 53 21

**CLI**
Screenshot 2026-01-12 at 17 06 19

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationWarning.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationWarning.tsx
@@ -1,8 +1,6 @@
 import {
-    ChartType,
     FeatureFlags,
-    isCustomDimension,
-    isDimension,
+    hasUnusedDimensions,
     type ChartConfig,
     type ItemsMap,
     type MetricQuery,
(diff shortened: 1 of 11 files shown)


magnew:
* **#19451** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19451?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19450** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19450?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19370** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19366** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19366?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19367** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19367?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19357** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/Explorer/VisualizationCard/VisualizationWarning.tsx
```diff
@@ -88,36 +86,14 @@ const VisualizationWarning: FC<PivotMismatchWarningProps> = ({
...
-        const unusedQueryDimensions = metricQueryDims.filter(
-            (d) => !usedDims.has(d),
-        );
-        return unusedQueryDimensions.length > 0;
+        return hasUnusedDimensions({
+            chartType: chartConfig.type,
+            chartConfig: chartConfig.config,
+            pivotDimensions: dirtyPivotDimensions,
+            queryDimensions: resultsData?.metricQuery?.dimensions ?? [],
+        });
```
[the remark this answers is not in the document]
magnew: This is intentional

packages/frontend/src/components/SettingsValidator/ValidatorTable/ErrorMessage.tsx
```diff
@@ -28,6 +29,20 @@ const ErrorMessageByType: FC<{
...
+            validationError.errorType === ValidationErrorType.ChartConfiguration
+        ) {
+            return (
+                <Text>
+                    <CustomMark>{validationError.fieldName}</CustomMark> is
+                    included in the query but not used in the chart
+                    configuration (x-axis, y-axis, or group by). This can cause
+                    incorrect rendering. We recommend removing unused dimensions
+                    from the query.
+                </Text>
```
pimmey, line 43: How comes we don't use the error message from backend here? 🤔

magnew: Good question. I had originally had just this one as more of a help-text than an error. I was thinking we would eventually link to some docs. 

But maybe I should try to unify them better.

pimmey: Right, just seemed like this from [ValidationService.ts‎](https://github.com/lightdash/lightdash/pull/19357/changes#diff-94f63e1efcc3a8e4cc7d4d18d794bae35239044eb1550c9b0ac963032541dc3fR486) one is not used on the FE :-)

magnew: I ended up not doing this 😬. Because I wanted the FE error to be a little more friendly and I wanted to keep the BE error more like a log error. And I couldn't find a reasonable middle ground. But I could do a follow up.

PR #19362, opened 2026-01-12
feat: add SCIM API examples and auto-verify emails for SCIM users

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/19363

## What the issue is

Some users are updating their okta email, but they get an error, because the new email gets unverified, causing this error when using tries to login again using OKTA

image

On SCIM, we create new emails if the email changes, but by default, those emails are `unverified`

Added some code to `automatically` verify emails when they are created from SCIM (existing pattern)

### How to test
- create a new user using scim.http
- Update scim user with a new username(email) using scim.http
- New email is unverified

* Before
Screenshot from 2026-01-12 17-07-59

* After

Screenshot from 2026-01-12 17-25-02

### Description:
Added automatic email verification for SCIM users when their email is updated. This ensures that when a user's email is changed through the SCIM API, the new email is automatically verified without requiring additional steps.

Also added an HTTP example file for SCIM API operations, including endpoints for creating and updating users, which serves as documentation for how to interact with the SCIM API.

### DIFF
--- a/examples/api/scim.http
+++ b/examples/api/scim.http
@@ -1,0 +1,57 @@
+
+### Login
+// @no-log
+POST http://localhost:8080/api/v1/login
+Content-Type: application/json
+
+{
+    "email": "demo@lightdash.com",
+    "password": "demo_password!"
+}
+
+# To get a new scim token, check scim-org-tokens.http
+
+### Create a new user
+
+POST http://localhost:8080/api/v1/scim/v2/Users
+Content-Type: application/json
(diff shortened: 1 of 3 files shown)


rephus:
* **#19362** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19362?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19362?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #19438, opened 2026-01-14
feat: duplicate charts when duplicating dashboard tabs

rephus:
Closes:  [https://linear.app/lightdash/issue/PROD-944/duplicate-dashboard-tab-charts-are-mirroring-charts-from-the-original](https://linear.app/lightdash/issue/PROD-944/duplicate-dashboard-tab-charts-are-mirroring-charts-from-the-original)

### what the issue is

duplicating a tab is duplicating the same charts, pointing to the same "chart within dashboard".
For charts within dashboards, we want to to duplicate the chart as well, so we can introduce changes.

This is how it works now:

1. Tab duplication (frontend) → Flags chart tiles that have belongsToDashboard: true with shouldDuplicateChart: true
2. Dashboard save (frontend → backend) → Sends tiles array, some with the flag
3. Backend processing → For each tile with shouldDuplicateChart: true:
\- Clones the chart via savedChartModel.create()
\- Updates catalog field usage
\- Tracks analytics event
\- Returns tile with new savedChartUuid (flag stripped)
4. Frontend state sync → onSuccessCallback updates local dashboardTiles with the response containing new UUIDs

Regular chart tiles (charts from spaces just referenced in the dashboard) won't be duplicated - they'll continue pointing to the same chart, which is the expected behavior.

### Demo

See how URL at the bottom is different for duplicated  chart within dashboards

[Screencast from 2026-01-15 10-18-59.webm](https://github.com/user-attachments/assets/605b4e5b-ca2f-4578-b4eb-d2e5b67514bd)

### Description:

Implemented chart duplication when duplicating dashboard tabs. When a user duplicates a tab containing charts that belong directly to the dashboard (not saved in spaces), the system now creates copies of those charts rather than having multiple tiles reference the same chart.

This ensures that changes to charts in one tab don't affect identical charts in other tabs, providing a more intuitive user experience when duplicating dashboard content.

The implementation:

1. Adds a `shouldDuplicateChart` flag to chart tile properties
2. Marks charts for duplication during tab duplication in both dashboard tab implementations
3. Handles chart duplication on the backend during dashboard save
4. Updates local state with the server response to ensure UI consistency


rephus:
* **#19438** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19438?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19438?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/features/dashboardTabs/index.tsx
```diff
@@ -301,11 +302,31 @@ const DashboardTabs: FC<DashboardTabsProps> = ({
...
+                        tile.properties.belongsToDashboard &&
+                        tile.properties.savedChartUuid
+                    ) {
+                        return {
+                            ...tile,
+                            uuid: uuid4(),
+                            tabUuid: newTab.uuid,
+                            properties: {
+                                ...tile.properties,
+                                shouldDuplicateChart: true,
```
rephus: add a flag in the frontend when duplicating tabs for charts within dashboards

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -459,83 +531,20 @@ export class DashboardService
...
-                                dashboardId:
-                                    duplicatedChart.dashboardUuid ?? undefined,
-                                duplicated: true,
-                                virtualViewId:
-                                    cachedExplore?.type === ExploreType.VIRTUAL
-                                        ? cachedExplore.name
-                                        : undefined,
-                            },
-                        });
+                        const newChartUuid =
```
rephus, line 539: Existing `duplicate dashboard` , where we were correctly duplicating charts within dashboards. 

I extracted this logic into a method, to be reused also when duplicating tabs.

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -670,10 +679,48 @@ export class DashboardService
                 new Set(dashboard.tiles.map((t) => t.type)),
             );
 
+            // Handle chart duplication for tiles marked with shouldDuplicateChart
```
rephus: This happens inside dashboard.update

packages/frontend/src/features/dashboardTabsV2/index.tsx
```diff
@@ -451,6 +452,25 @@ const DashboardTabsV2: FC<DashboardTabsProps> = ({
                 const duplicatedTiles = tilesToDuplicate.map((tile) => {
                     const newUuid = uuid4();
                     tileUuidMapping.set(tile.uuid, newUuid);
+
+                    // For chart tiles that belong to the dashboard, mark them for duplication
```
rephus: duplicate logic on tabs v2

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -670,10 +679,48 @@ export class DashboardService
...
+            const tilesToSave = await Promise.all(
+                dashboard.tiles.map(async (tile) => {
+                    // Check if this is a chart tile that needs duplication
+                    if (
+                        tile.type === DashboardTileTypes.SAVED_CHART &&
+                        tile.properties.shouldDuplicateChart &&
+                        tile.properties.savedChartUuid
+                    ) {
+                        const newChartUuid =
+                            await this.duplicateChartForDashboard({
```
rephus: reuse duplicate chart method, extracted from duplicate dashboard

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -193,6 +193,78 @@ export class DashboardService
...
+        projectUuid,
+        dashboardUuid,
+        user,
+    }: {
+        chartUuid: string;
+        projectUuid: string;
+        dashboardUuid: string;
+        user: SessionUser;
+    }): Promise<string> {
+        const chartToDuplicate = await this.savedChartModel.get(chartUuid);
```
IrakliJani, line 211: could you throw an error if chart does not belong to a dashboard? to avoid duplicating charts that does not belong to it in case we change something accidentally in the FE?

packages/common/src/types/dashboard.ts
```diff
@@ -58,6 +58,7 @@ export type DashboardChartTileProperties = {
         chartName?: string | null;
         lastVersionChartKind?: ChartKind | null;
         chartSlug?: string;
+        shouldDuplicateChart?: boolean; // Set when duplicating a tab - signals backend to create a copy of the chart
```
IrakliJani: can we use `belongsToDashboard`flag and only rely on that in the BE?

rephus: afaik all charts within dashboards have `belongsToDashboard` flag. So how do we know if this charts was recently added and needs to be duplicated ?

I guess we could group all those charts in the backend, and the chart uuids are not unique,  duplicate them in the backend. What do you think ?

rephus: Done:

  When duplicating a tab:
  1. Frontend copies tiles with the same savedChartUuid
  2. Backend detects when the same chart UUID appears multiple times for belongsToDashboard: true tiles
  3. First occurrence keeps the original chart
  4. Subsequent occurrences get new duplicated charts created automatically

So frontend code was simplified, but backend was slightly more complicated, which is fine.

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -670,10 +684,96 @@ export class DashboardService
...
+                        ...tile,
+                        properties: {
+                            ...tile.properties,
+                            savedChartUuid: newChartUuid,
+                        },
+                    };
+                }
+                return tile;
+            });
+
```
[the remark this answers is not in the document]
rephus: forgot to remove frontend flags :+1:

rephus:
Unrelated e2e errors 
image

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -670,10 +684,96 @@ export class DashboardService
...
+            // Handle chart duplication for dashboard charts that appear multiple times
+            // This happens when duplicating a dashboard tab with charts saved directly to the dashboard
+            // We detect duplicates by finding chart UUIDs that appear more than once
+            // Step 1: Count occurrences of each chart UUID for dashboard charts
+            const chartUuidOccurrences = new Map<string, number>();
+            dashboard.tiles.forEach((tile) => {
+                if (
+                    tile.type === DashboardTileTypes.SAVED_CHART &&
+                    tile.properties.belongsToDashboard &&
+                    tile.properties.savedChartUuid
```
IrakliJani, line 696: nice!

PR #19440, opened 2026-01-14
chore: drop `nodemon` in favour of `tsx watch`

Spissable:
### Description:
Just noticed that there is only 1 remaining instance of `nodemon` being used. Replacing it in order to drop the dependency altogether.


ZeRego (approved):

Spissable:
TIL: `tsx watch` treats stdin as a change event and therefore reloads, which makes a REPL rather useless.

Closing until a better alternative shows up.

PR #19449, opened 2026-01-14
feat: delete OpenID identities by email in SCIM service to prevent login conflicts

rephus:
Closes: https://linear.app/lightdash/issue/PROD-2483/remove-conflicting-openid-records-during-scim-user-creationupdate

## What the issue is

- when user changes the email on okta. 
- scim disables previous user
- scim creates a new user / or updates existing one

But an existing openid might still be there, pointing to the old deactivated user, causing issues on login

This PR deletes existing openids with the same email on SCIM create  and update (when active flag changes). 

### Before

Screenshot from 2026-01-15 09-32-50
Screenshot from 2026-01-15 09-33-16
Screenshot from 2026-01-15 09-32-58

### After
Screenshot from 2026-01-15 09-54-45
Screenshot from 2026-01-15 09-56-18

### Description:
Added functionality to delete existing OpenID identities for users when they are created via SCIM. This prevents login conflicts when a user's email is changed through SCIM and an old OpenID identity record exists pointing to a deactivated account.

The implementation:
1. Added `openIdIdentityModel` to the ScimService constructor arguments
2. Created a new method `deleteIdentitiesByEmail` in the OpenIdIdentityModel
3. Added a call to delete existing identities before creating a new user via SCIM

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.ts
@@ -645,7 +645,7 @@ export class ScimService extends BaseService {
             // If active status changes, either true or false
             // We delete all openid identities for the user's email and user uuid
             // to prevent login conflicts
-            if (user.active !== undefined) {
+            if (user.active && user.active !== dbUser.isActive) {
                 this.logger.debug(
                     `SCIM: Updating active user ${emailToUpdate} to ${user.active}`,
                 );


rephus:
* **#19449** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19449?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19449?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -417,6 +423,13 @@ export class ScimService extends BaseService {
                 ScimService.validateRolesArray(user.roles, validRoleValues);
             }
             const email = ScimService.getScimUserEmail(user);
+            // Delete any existing openid identities for this email to prevent login conflicts
+            // This handles the case where a user's email changed via SCIM and an old
+            // openid identity record exists pointing to a deactivated account
+            this.logger.debug(
+                `SCIM: deleting openid identities for email ${email}`,
+            );
+            await this.openIdIdentityModel.deleteIdentitiesByEmail(email);
```
ZeRego: Pseudo code:
```suggestion
            const deletedIdentities = await this.openIdIdentityModel.deleteIdentitiesByEmail(email);
            if(deletedIdentities > 0) {
              this.logger.debug(
                `SCIM: deleted ${deletedIdentities} openid identities for email ${email}`,
              );
            }
```

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -569,6 +582,18 @@ export class ScimService extends BaseService {
                 true, // automatically verify email
             );
 
+            if (user.active && user.active !== dbUser.isActive) {
```
ZeRego: Should we move this to `if (user.active === false) {` block?
Also, we should delete all oidc identities for this user, not just the ones matching the email.

rephus: I think we should remove it if both goes active and inactive (in fact, the issue we had is when the openid was still pointing to an inactive user) 

We could just do `if (user.active !== undefined)`

> Also, we should delete all oidc identities for this user, not just the ones matching the email.

Yeah, that makes sense

rephus: one thing might cause issues is that any refresh tokens are used from openid for scheduled deliveries. But afaik okta doesn't use refresh tokents at the moment. So it should be fine.

packages/backend/src/ee/services/ScimService/ScimService.ts
```diff
@@ -599,6 +642,19 @@ export class ScimService extends BaseService {
                 roles: user.roles,
             });
 
+            // If active status changes, either true or false
+            // We delete all openid identities for the user's email and user uuid
+            // to prevent login conflicts
+            if (user.active !== undefined) {
```
ZeRego: I prefered the previous approach where you compare what is in the db vs the new value. I know there will be a provider that always passes a value on each update and we will keep deleting authentication

ZeRego (approved):

PR #19466, opened 2026-01-15
fix: apply correct conditional formatting for pivoted columns

Spissable:
### Description:
When a dimension is pivoted from rows to columns, its values are stored in `headerInfo`. Before this fix, conditional formatting rules that compared against the pivoted dimension couldn't find the field in `rowFields`. Now, the pivoted dimension values are extracted from `headerInfo` and included in `rowFields`, allowing conditional formatting rules to evaluate correctly.

### Example Conditional formatting:
image

### Before:
before

### After:
after


Spissable:
> @Spissable
> 
> Mostly looking great, didn't work with: unpivoted dimension + metrics as rows

Thanks for spotting @notgiorgi! I replicated your scenario and it's looking good with my 2nd commit:
image

PR #19469, opened 2026-01-15
chore: update map tiles

magnew:
### Description:

Updates the map tiles to use map servers with clearer fair use policies. We were using tiles that are commonly used, but don't explicitly allow light commercial usage. These new tile layers should be ok to use until we have a lot of usage, then we will need to host or buy tiles. (there is a ticket for tracking usage)

We can decide we don't need so many layer options, but for now this is at least move them to ones we can use. 

Updated to be only OSM and satellite tiles. We can add more as needed, but we have to get registered

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/MapConfig/MapDisplayConfig.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/MapConfig/MapDisplayConfig.tsx
@@ -390,12 +390,10 @@ export const Display: FC = memo(() => {
                     <Select
                         data={[
                             { value: MapTileBackground.NONE, label: 'None' },
-                            { value: MapTileBackground.LIGHT, label: 'Light' },
                             {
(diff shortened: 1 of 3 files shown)


magnew:
* **#19469** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19469?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

magnew:
> Getting some odd issues.
> 
> **With dark**
> 
> Image
> **With light**
> 
> Image
> With openstreetmap it works well
> 
> Image

Ok, looking into it. It's possible we need an account even if it's free. Maybe we limit it to OSM for now.

rephus (approved):

PR #19488, opened 2026-01-15
fix: dont include chart warnings in error count

magnew:
### Description:

Filter chart config warnings out of validation error count in the header. 

These are warnings that don't need to be included in the error count.

### DIFF
--- a/packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
+++ b/packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
@@ -1,4 +1,7 @@
-import { NotificationResourceType } from '@lightdash/common';
+import {
+    NotificationResourceType,
+    ValidationErrorType,
+} from '@lightdash/common';
 import { Button, Indicator, Menu } from '@mantine/core';
 import { IconBell } from '@tabler/icons-react';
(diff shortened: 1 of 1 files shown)


magnew:
* **#19488** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/NavBar/NotificationsMenu/index.tsx
```diff
@@ -93,10 +100,10 @@ export const NotificationsMenu: FC<{ projectUuid: string }> = ({
                 </Button>
             </Menu.Target>
             <Menu.Dropdown maw="400px">
-                {hasValidationNotifications && (
+                {hasValidationNotifications && validationErrors && (
```
almeidabbm: [nitpick] don't think we need the double check for `validationErrors` here, we already check for it in `hasValidationNotifications`

PR #19495, opened 2026-01-16
fix: do not run and disable scheduled deliveries when there are no targets

almeidabbm:
Closes: https://linear.app/lightdash/issue/GLITCH-144/scheduled-deliveries-with-no-targets-shouldnt-run

### Description:
Added support for a new partial failure type `MISSING_TARGETS` to handle cases where scheduled deliveries have no recipients configured. The scheduler will now:

1. Detect when a scheduler has no targets and disable it automatically
2. Log this as a specific failure type rather than silently failing
3. Display a user-friendly message in the UI about missing recipients

This improves the user experience by providing clear feedback when scheduled deliveries fail due to configuration issues rather than data problems.

### DIFF
--- a/packages/backend/src/scheduler/SchedulerTask.ts
+++ b/packages/backend/src/scheduler/SchedulerTask.ts
@@ -2913,6 +2913,59 @@ export default class SchedulerTask {
             return;
         }
 
+        const {
+            createdBy: userUuid,
+            savedChartUuid,
+            dashboardUuid,
+            thresholds,
+            notificationFrequency,
+            targets,
+        } = scheduler;
+
+        const sessionUser = await this.userService.getSessionByUserUuid(
+            userUuid,
+        );
+        const account = Account.fromSession(sessionUser);
+
+        // If the scheduler is not a gsheets and has no targets, we skip the delivery
+        if (
+            scheduler.format !== SchedulerFormat.GSHEETS &&
+            targets.length === 0
+        ) {
+            Logger.warn(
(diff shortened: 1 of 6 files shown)


almeidabbm:
* **#19495** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19495?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19495?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -2913,6 +2913,56 @@ export default class SchedulerTask {
...
+            targets,
+        } = scheduler;
+
+        const sessionUser = await this.userService.getSessionByUserUuid(
+            userUuid,
+        );
+        const account = Account.fromSession(sessionUser);
+
+        // If the scheduler is not a gsheets and has no targets, we skip the delivery
+        if (scheduler.format !== SchedulerFormat.GSHEETS && !targets.length) {
```
Spissable: Nitpick: reading a negation with a numeric value is not super clear to me. I would suggest this version:

```suggestion
        if (scheduler.format !== SchedulerFormat.GSHEETS && targets.length === 0) {
```

almeidabbm: Good suggestion! Updated 🙇

PR #19532, opened 2026-01-19
chore: add `DEV` prefix in document title in `DEV` mode

Spissable:
### Description:
Small QoL improvement for local development. Since I constantly have our prod instance and a local instance open in my browser, I've been finding it hard to distinguish between the two.

As a simple fix for that, I just add the postfix ` (DEV)` in the document title, whenever we're in DEV mode.

image

### DIFF
--- a/packages/frontend/src/components/common/DocumentTitle.tsx
+++ b/packages/frontend/src/components/common/DocumentTitle.tsx
@@ -5,11 +5,11 @@ type Props = {
 };
 
 const APP_NAME = 'Lightdash';
-const SUFFIX = import.meta.env.DEV ? ' (DEV)' : '';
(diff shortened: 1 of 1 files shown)


pimmey:
Love it! Do you think it's better as postfix or maybe prefix could be more universal, including boomers like me who still use browsers with horizontal tabs? 😄

Spissable:
> Love it! Do you think it's better as postfix or maybe prefix could be more universal, including boomers like me who still use browsers with horizontal tabs? 😄

@pimmey on second thought I kinda like the prefix idea! I was navigating to some pages with long titles, and the "DEV" disappeared since the title became too long!


Pushed:
image

PR #19558, opened 2026-01-19
fix: support additional dimensions in dbt 1.10+ config.meta structure

pimmey:
Closes: PROD-2558

### Description:
Fix additional dimensions support for dbt 1.10+ where metadata is stored in `config.meta` instead of `meta`. This ensures that interval dimensions of additional dimensions properly use the additional dimension's properties (label, sql, etc.) instead of being overwritten by the base dimension's properties.

The fix adds support for the new metadata structure by setting `config.meta.dimension` to prevent the base dimension's properties from overwriting the additional dimension's properties.

<details>
<summary>Demo</summary>

### Before
Screenshot 2026-01-19 at 17 54 14

### After
Screenshot 2026-01-19 at 17 57 11
</details>

### DIFF
--- a/packages/common/src/compiler/translator.ts
+++ b/packages/common/src/compiler/translator.ts
@@ -594,6 +594,16 @@ export const convertTable = (
                         intervals = getDefaultTimeFrames(dim.type);
                     }
 
+                    const dimensionMeta = {
+                        ...columnMeta.dimension,
+                        type: dim.type,
+                        label: dim.label,
+                        groups: dim.groups,
+                        sql: dim.sql,
+                        description: dim.description,
+                        hidden: dim.hidden,
+                    };
+
                     return intervals.reduce(
                         (acc, interval) => ({
                             ...acc,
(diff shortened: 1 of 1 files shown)


pimmey:
* **#19558** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19558?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19558?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/compiler/translator.ts
```diff
@@ -621,6 +621,23 @@ export const convertTable = (
                                                           hidden: dim.hidden,
                                                       },
                                                   },
+                                                  // In dbt 1.10+, config.meta takes precedence over meta
+                                                  // so we must set config.meta.dimension to prevent
+                                                  // the base dimension's properties from overwriting
+                                                  config: {
+                                                      meta: {
```
tatianainama, line 629: nit: should we store this `meta` configuration so we ensure it's always the same if it's both at root level or nested under config? just to make sure if we update anything here in the future, we don't forget updating both locations, keeping it a bit DRY

pimmey: Well spotted 👌

PR #19571, opened 2026-01-20
feat: add `my scheduled deliveries` user settings page

Spissable:
Closes: https://linear.app/lightdash/issue/GLITCH-112/my-schedules-page-in-user-settings

## Description:
This PR adds a new user settings page called "My scheduled deliveries". It's basically a simplified version of the project settings "Syncs & Scheduled deliveries" page and with a different main scope.

Main differences:
- No filtering for the owner and no owner column, because it's always scoped to the current user.
- The main endpoint difference is that the new page isn't scoped to the project anymore, but rather to the user.
- No run history, except for the "Latest Run" column. This could be implemented but the PR is already way too big as it is.
- FE and BE check permissions for "create ScheduledDeliveries". If you can't create any, you also can't view any.

## Testing:
For easier testing I made sure to reference the user and project in the scheduled deliveries name, so you can take that as a reference in the screenshot below.

### demo3@lightdash.com (viewer) doesn't have the setting:
image

### demo2@lightdash.com (editor):
✅ Sees ALL their scheduled deliveries across different projects and different types
✅ Delete works
✅ Filtering works

image

### demo@lightdash.com (admin):
✅ All the functionality is preserved. Table is scoped by project rather than users.

image

### DIFF
--- a/packages/frontend/src/components/SchedulersView/index.tsx
+++ b/packages/frontend/src/components/SchedulersView/index.tsx
@@ -83,12 +83,10 @@ const SchedulersView: FC<{ projectUuid?: string; isUserScope?: boolean }> = ({
     };
 
     const handleRefresh = async () => {
-        const invalidateQueries = isUserScope
-            ? [queryClient.invalidateQueries(['userPaginatedSchedulers'])]
-            : [
-                  queryClient.invalidateQueries(['paginatedSchedulers']),
-                  queryClient.invalidateQueries(['schedulerRuns']),
-              ];
+        const invalidateQueries = [
+            queryClient.invalidateQueries(['paginatedSchedulers']),
+            queryClient.invalidateQueries(['schedulerRuns']),
+        ];
 
         await Promise.all(invalidateQueries);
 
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/SchedulersView/SchedulersTable.tsx
```diff
@@ -156,9 +162,13 @@ const SchedulersTable: FC<SchedulersTableProps> = ({
     const totalDBRowCount = data?.pages?.[0]?.pagination?.totalResults ?? 0;
     const totalFetched = flatData.length;
 
+    // For user scope, get projectUuid from first scheduler item for fetching project details
+    const projectUuid = isUserScope
+        ? (flatData[0]?.projectUuid ?? undefined)
+        : projectUuidProp;
+    const { data: project } = useProject(projectUuid);
```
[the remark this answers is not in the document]
almeidabbm: @Spissable this is true here, in user mode you can get schedulers from different projects

packages/backend/src/services/SchedulerService/SchedulerService.ts
almeidabbm, line 400: This function is calling `getSchedulerForUser` internally in the model.
At the service level we already call `this.schedulerModel.getScheduler` on line 379, so we're kinda running 2 queries to get schedulers.
Can't we use `getSchedulerRuns` directly here and pass the schedulers from line 379 so there wouldn't be the need to add the new `getUserSchedulers` function? 🤔

packages/backend/src/services/SchedulerService/SchedulerService.ts
```diff
@@ -319,7 +319,7 @@ export class SchedulerService extends BaseService {
             return schedulers;
         }
 
-        const runs = await this.schedulerModel.getSchedulerRuns({
```
almeidabbm, line 322: Same thing here 🤔

Issue #19570, opened 2026-01-20
Google Sync scheduler fails when sync is disabled; toggle label unclear


jesshitchcock:
*Issue summary:*

A customer reported that their Google Sync scheduler was failing. Upon investigation, it was found that the sync was disabled. The customer was confused by the toggle, believing it only controlled notifications, not the sync status itself.

*Steps to reproduce:*
- Disable Google Sync for a chart.
- Attempt to sync to sheets and observe that the scheduler fails, but the UI toggle does not clearly indicate that sync is disabled (customer thought it was for notifications).

*Expected behavior:*
- The toggle should clearly indicate that it controls the sync status, not just notifications.
- The UI should provide clear feedback when sync is disabled and explain the consequences (e.g., scheduler will not run).

*Suggested improvements:*
- Update the toggle label and/or add a tooltip to clarify its function.
- Consider adding a warning or info message when sync is disabled, explaining that scheduled syncs will not run.
- Add a toggle to enable or disable syncs and deliveries from the admin panel

PR #19580, opened 2026-01-20
feat: improve visual feedback on scheduled delivery toggle

Spissable:
Closes: https://github.com/lightdash/lightdash/issues/19570

### Description:
This PR implements 2 of the 3 considerations of the issue linked above:

Suggestion 1
> Update the toggle label and/or add a tooltip to clarify its **function.**

✅ Commit 1

> Consider adding a warning or info message when sync is disabled, explaining that scheduled syncs will not run.

✅ Commit 2

Suggestion 3
> Add a toggle to enable or disable syncs and deliveries from the admin panel

❌ I think we should treat this one more as a feature request rather than a quick hotfix. I think it's not immediately clear where to integrate this functionality. E.g. maybe it should be part of the edit dialog, or just in the menu drop down.

### Testing:

Scheduled Deliveries:

https://github.com/user-attachments/assets/9f030b09-8484-49e1-87b1-49c8ca0dee14

Google Sync:

https://github.com/user-attachments/assets/d9adc121-e2f1-4d0b-a2bd-94a9f7d0efd7

### DIFF
--- a/packages/frontend/src/features/scheduler/components/ConfirmPauseSchedulerModal.tsx
+++ b/packages/frontend/src/features/scheduler/components/ConfirmPauseSchedulerModal.tsx
@@ -9,6 +9,7 @@ type ConfirmPauseSchedulerModalProps = {
     loading?: boolean;
     onClose: () => void;
     onConfirm: () => void;
+    description: string;
 };
 
 const ConfirmPauseSchedulerModal: FC<ConfirmPauseSchedulerModalProps> = ({
@@ -17,6 +18,7 @@ const ConfirmPauseSchedulerModal: FC<ConfirmPauseSchedulerModalProps> = ({
     loading,
     onClose,
     onConfirm,
+    description,
 }) => {
     return (
         <MantineModal
@@ -30,7 +32,7 @@ const ConfirmPauseSchedulerModal: FC<ConfirmPauseSchedulerModalProps> = ({
                     Pause
                 </Button>
             }
-            description="This will pause the scheduled delivery. It will not run until it is enabled again."
+            description={description}
         />
     );
 };
--- a/packages/frontend/src/features/sync/components/SyncModalView.tsx
+++ b/packages/frontend/src/features/sync/components/SyncModalView.tsx
@@ -50,8 +50,8 @@ const ToggleSyncEnabled: FC<{ scheduler: Scheduler }> = ({ scheduler }) => {
                 maw={130}
                 label={
                     scheduler.enabled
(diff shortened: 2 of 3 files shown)


jesshitchcock:
@Spissable can we add different text for the toggle depending on whether it is a syncs or scheduled delivery?

jesshitchcock:
Suggestion 3 is a nice to have - let's leave it for now

packages/frontend/src/features/scheduler/components/ConfirmPauseSchedulerModal.tsx
```diff
@@ -0,0 +1,38 @@
...
+            onClose={onClose}
+            title={`Pause "${schedulerName}"?`}
+            icon={IconPlayerPause}
+            size="md"
+            actions={
+                <Button onClick={onConfirm} loading={loading}>
+                    Pause
+                </Button>
+            }
+            description="This will pause the scheduled delivery. It will not run until it is enabled again."
```
almeidabbm: We should allow passing this as a `required` prop so that for `syncs` it doesn't mention `scheduled delivery`

packages/frontend/src/features/sync/components/SyncModalView.tsx
```diff
@@ -39,28 +40,45 @@ const ToggleSyncEnabled: FC<{ scheduler: Scheduler }> = ({ scheduler }) => {
...
-            </Box>
-        </Tooltip>
+        <>
+            <Tooltip
+                withinPortal
+                variant="xs"
+                maw={130}
+                label={
+                    scheduler.enabled
+                        ? 'Toggle off to temporarily pause scheduled delivery'
```
almeidabbm: here as well, it should mention `sync` rather than scheduled delivery

almeidabbm (approved):
Woot woot!! 🚀 
Just some small nit picks

Spissable:
Updated the wording on the tool-tip depending on whether it's a scheduled delivery or google syc. e.g.
image

packages/frontend/src/features/scheduler/components/ConfirmPauseSchedulerModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+    onClose,
+    onConfirm,
+    description,
+}) => {
+    return (
+        <MantineModal
+            opened={opened}
+            onClose={onClose}
+            title={`Pause "${schedulerName}"?`}
+            icon={IconPlayerPause}
```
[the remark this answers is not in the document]
Spissable: That was actually incorrect. The custom component already takes care of this

Issue #19444, opened 2026-01-14
Enable choosing geo join keys for custom maps


magnew:
### Description

Make it possible to choose which geojson property to use for joining maps shapes to query data. 

### What problem does this solve?

This will enable users to use custom map layers

PR #19583, opened 2026-01-20
feat: add support for joins in custom maps

magnew:
Closes: #19444, PROD-2484

### Description:

Adds support for setting up joins with custom map shapes. With this change, users can join the shapes in their geojson to columns in their data. 

A typical geojson file includes a property section for each shape (area), for example:

```
"properties": {
        "cod_prov": "07",
        "name": "Illes Balears",
        "cod_ccaa": "03",
}
```
This PR parses custom JSON provided by the user and allows them to chose which of these properties maps to a column in their data. 

**To test**
- Build a query with `Tables/Geo`
- Include `State province name` and `Customer count sum` fields
- Switch the viz to map
- Choose Map type = Area
- Use 'custom region'
- Add [this geo data URL](https://raw.githubusercontent.com/codeforgermany/click_that_hood/refs/heads/main/public/data/spain-provinces.geojson)(Spains provinces)
- The 'Map join field' should be populated by the properties in the json ✅ 
- Choose 'name'
- Chose 'State province name' in the 'Data join field'
- You should see one province in Spain get colored in ✅ (at the moment we only have US data which accidentally matches here, I'm going to add more)

Screenshot 2026-01-20 at 14 58 25

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/MapConfig/MapLayoutConfig.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/MapConfig/MapLayoutConfig.tsx
@@ -10,14 +10,18 @@ import {
 } from '@lightdash/common';
 import {
     Group,
+    Loader,
     SegmentedControl,
     Select,
     Stack,
     Switch,
-    Text,
     TextInput,
 } from '@mantine/core';
 import { memo, useEffect, useMemo, type FC } from 'react';
+import {
+    findMatchingProperty,
+    useGeoJsonProperties,
+} from '../../../hooks/useGeoJsonProperties';
 import { isMapVisualizationConfig } from '../../LightdashVisualization/types';
 import { useVisualizationContext } from '../../LightdashVisualization/useVisualizationContext';
 import FieldSelect from '../../common/FieldSelect';
@@ -33,6 +37,12 @@ const getRegionFieldConfig = (
             return {
                 label: 'State field',
                 description: 'Field containing state codes (e.g., "CA", "NY")',
+            };
+        case MapChartLocation.CUSTOM:
+            return {
+                label: 'Data join field',
+                description:
+                    'Field containing values that match the selected map join field',
             };
         case MapChartLocation.WORLD:
         default:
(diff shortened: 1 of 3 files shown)


magnew:
* **#19588** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19587** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19587?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19583** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19583?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/hooks/useGeoJsonProperties.ts
```diff
@@ -0,0 +1,184 @@
...
+};
+
+/**
+ * Extracts unique property keys from all features in a GeoJSON FeatureCollection.
+ * For TopoJSON, converts to GeoJSON first.
+ */
+const extractPropertiesFromGeoJson = (
+    data: unknown,
+): GeoJsonPropertiesResult => {
+    let features: GeoJSON.Feature[] = [];
```
almeidabbm, line 37: Wondering where `GeoJSON` comes from? 🤔 thought it would need to be imported from leaflet?

packages/frontend/src/components/VisualizationConfigs/MapConfig/MapLayoutConfig.tsx
```diff
@@ -95,8 +105,46 @@ export const Layout: FC = memo(() => {
         );
     }, [itemsMap]);
 
-    // Extract chart config values (only valid when isMapConfig is true)
+    // Extract chart config for use in hooks (before early return)
     const chartConfig = isMapConfig ? visualizationConfig.chartConfig : null;
```
almeidabbm: This should probably be a `useMemo` as the chart config might change quite often

packages/frontend/src/hooks/useGeoJsonProperties.ts
magnew: I guess leaflet adds it as a global namespace. Does that seem like a bad idea?

almeidabbm: Not at all, it's just a type, was mostly curious 🤔

PR #19585, opened 2026-01-20
feat: map analytics events to prometheus metrics

Shifan-Gu:
Closes: 

### Context

In our self-hosted deployment, we need app-level metrics to autoscale our Trino cluster using KEDA. Since KEDA uses Prometheus queries as autoscaling triggers, we need a way to expose Lightdash analytics events as Prometheus metrics meanwhile keep the flexibility without touching the existing prometheus.ts file and application code.

### Solution

This PR introduces a `CustomMetricManager` that converts Lightdash analytics events to Prometheus counters through JSON configuration. This approach allows us to:

- Select which analytics events to expose as Prometheus metrics
- Configure metrics without modifying `prometheus.ts` or application code
- Maintain flexibility for deployment-specific autoscaling needs

### Implementation

- **Configuration-driven**: Define metrics in JSON (see `packages/backend/src/metrics/custom-metrics.config.example.json`)
- **Automatic event subscription**: Hooks into existing analytics system
- **Dynamic label extraction**: Extracts labels from event payloads based on configuration
- **Seamless integration**: Works with existing Prometheus setup

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -19,6 +19,7 @@ import {
 import * as Sentry from '@sentry/node';
 import flash from 'connect-flash';
 import connectSessionKnex from 'connect-session-knex';
+import { EventEmitter } from 'events';
 import express, { Express, NextFunction, Request, Response } from 'express';
 import expressSession from 'express-session';
 import expressStaticGzip from 'express-static-gzip';
@@ -73,13 +74,12 @@ import {
 } from './services/ServiceRepository';
 import { UtilProviderMap, UtilRepository } from './utils/UtilRepository';
 import { VERSION } from './version';
-import PrometheusMetrics from './prometheus';
+import PrometheusMetrics from './prometheus/PrometheusMetrics';
 import { snowflakePassportStrategy } from './controllers/authentication/strategies/snowflakeStrategy';
 import { databricksPassportStrategy } from './controllers/authentication/strategies/databricksStrategy';
 import { jwtAuthMiddleware } from './middlewares/jwtAuthMiddleware';
 import { InstanceConfigurationService } from './services/InstanceConfigurationService/InstanceConfigurationService';
 import { slackPassportStrategy } from './controllers/authentication/strategies/slackStrategy';
-import { SlackClient } from './clients/Slack/SlackClient';
 import { sessionAccountMiddleware } from './middlewares/accountMiddleware';
 
 // We need to override this interface to have our user typing
@@ -181,10 +181,13 @@ export default class App {
 
     private readonly customExpressMiddlewares: Array<(app: Express) => void>;
 
+    private readonly analyticsEventEmitter: EventEmitter;
+
     constructor(args: AppArguments) {
         this.lightdashConfig = args.lightdashConfig;
         this.port = args.port;
         this.environment = args.environment || 'production';
+        this.analyticsEventEmitter = new EventEmitter();
         this.analytics = new LightdashAnalytics({
             lightdashConfig: this.lightdashConfig,
             writeKey: this.lightdashConfig.rudder.writeKey || 'notrack',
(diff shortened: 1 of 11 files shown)


packages/backend/src/App.ts
```diff
@@ -293,6 +310,66 @@ export default class App {
         }
     }
 
+    /**
+     * Initialize CustomMetricManager from configuration file if available
+     */
+    private initializeCustomMetricManager(): CustomMetricManager | null {
+        const configPath =
+            process.env.LIGHTDASH_CUSTOM_METRICS_CONFIG_PATH ||
+            process.env.CUSTOM_METRICS_CONFIG_PATH;
```
ZeRego: Let's move this to `parseConfig.ts` and here we can use `this.lightdashConfig.X`

packages/backend/src/App.ts
```diff
@@ -224,6 +231,10 @@ export default class App {
         this.prometheusMetrics = new PrometheusMetrics(
             this.lightdashConfig.prometheus,
         );
+
+        // Initialize CustomMetricManager if config is available
+        this.customMetricManager = this.initializeCustomMetricManager();
```
ZeRego: Would it be possible to have this logic inside the existing prometheus class? and have the folder/files structure:

/prometheus/PrometheusMetrics.ts
/prometheus/PrometheusEventMetricManager.ts // instead of CustomMetricManager
/prometheus/...

This way we don't have 2 classes related to prometheus metrics in App.ts and we can easily identify what they are from the file name. "Custom Metric" its a term already used inside Lightdash so it could be confusing.

packages/backend/src/metrics/CustomMetricManager.ts
```diff
@@ -0,0 +1,261 @@
...
+        // Override the track method to emit events before calling the original
+        this.analytics.track = ((
+            payload: Parameters<LightdashAnalytics['track']>[0],
+        ) => {
+            // Call the original track method first
+            this.originalTrackMethod!(payload);
+
+            // Emit event for our listeners
+            this.eventEmitter.emit('analytics:track', payload);
+        }) as typeof this.analytics.track;
```
ZeRego: iirc, this will force the type rather than assert the function matches the type. 
So if we change the signature, we won't catch the problem here.

ZeRego: We could create the event emitter in App.ts, so it can be used in LightdashAnalytics (to emit) and here (subscribe). This way, we avoid overwriting the method! 
We can also adopt more specific events so this class can subscribe to specific events instead processing all of them.
The structure `<source>.<method>.<event>` provides a clear hierarchy:
-   `source`: `analytics` (The module/class)
-   `method`: `track` (The specific action being performed)
-   `event`: `user.created` (The payload's event name)

🚀

packages/backend/src/metrics/CustomMetricManager.ts
```diff
@@ -0,0 +1,261 @@
...
+        payload: Parameters<LightdashAnalytics['track']>[0],
+    ): Record<string, string> {
+        // Use custom extractor if provided
+        if (metricConfig.extractLabels) {
+            return metricConfig.extractLabels({
+                event: payload.event,
+                userId: payload.userId,
+                anonymousId: payload.anonymousId,
+                properties: payload.properties,
+            });
```
ZeRego: We need validation here, in case the function returns an unexpected data format.

packages/backend/src/prometheus/README.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+### Configuration Fields
+
+- **eventName**: The Lightdash analytics event name to listen for (e.g., `user.logged_in`, `query.executed`)
+- **metricName**: The Prometheus metric name (must follow Prometheus naming conventions)
+- **help**: Help text for the Prometheus metric
+- **labelNames**: Array of label names to extract from the event payload. These must match the exact property keys in `payload.properties` (e.g., `loginProvider`, `projectId`, `context`)
+- **extractLabels** (optional): Custom function to extract labels (not supported in JSON, would need to be added programmatically)
```
ZeRego, line 50: > (not supported in JSON, would need to be added programmatically)

Can you expand on this?

Shifan-Gu:
thanks for the valuable feedback @ZeRego! i will re-work on it and let you know how it goes.

PR #19587, opened 2026-01-20
fix: zoom to data on maps

magnew:
### Description:

Zooms map extent to the data in the map. Before we were not changing the map bounds when the map area changed. Now we change the map bounds based on the area loaded.

**Note** the perf isn't amazing when switched between these heavy maps (to and from 'World'). We can work on perf separately.
 
Kapture 2026-01-20 at 16 56 55

### DIFF
--- a/packages/frontend/src/components/SimpleMap/index.tsx
+++ b/packages/frontend/src/components/SimpleMap/index.tsx
@@ -1,4 +1,4 @@
-import { MapChartType } from '@lightdash/common';
+import { MapChartLocation, MapChartType } from '@lightdash/common';
 import { Box, Divider, Stack, Text, UnstyledButton } from '@mantine/core';
 import { useClipboard } from '@mantine/hooks';
 import { IconCopy, IconMap } from '@tabler/icons-react';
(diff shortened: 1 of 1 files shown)


magnew:
* **#19588** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19587** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19587?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#19583** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19583?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/components/SimpleMap/index.tsx
almeidabbm: since we removed `mapConfig` from the dep array we can probably just convert this into `mapConfig?.isLatLong` and remove the `eslint-disable-next-line react-hooks/exhaustive-deps` below

packages/frontend/src/components/SimpleMap/index.tsx
```diff
@@ -314,6 +314,96 @@ const MapExtentTracker: FC<{
...
+                        hasFitted: true,
+                    };
+                }
+            } catch (error) {
+                console.warn('Failed to fit map to scatter bounds:', error);
+            }
+        }
+        // Note: geoJsonData and scatterData are intentionally in deps to re-fit when data loads,
+        // but the ref check above prevents re-fitting on subsequent renders with same config
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
almeidabbm: No need for this, removed it locally and it didn't complain, it seems like all deps are used within the hook

packages/frontend/src/components/SimpleMap/index.tsx
magnew: Good call

PR #19625, opened 2026-01-21
fix: respect the color gradient in maps

almeidabbm:
Closes: https://linear.app/lightdash/issue/PROD-1065/maps-should-have-more-color-picking-control

### Description:
Refactored map color handling to improve consistency and maintainability:

- Created a new constants file with `MAP_FILL_OPACITY` (0.7) to standardize opacity across regions, markers, and legend
- Implemented a more robust color scale system using `createColorScale` function that properly interpolates between colors
- Added memoized color scales for both scatter points and choropleth regions to improve performance
- Replaced direct color calculation with the new scale functions
- Fixed edge cases for single color or empty color arrays

This change ensures consistent coloring across the map visualization while making the code more maintainable.

### DIFF
--- a/packages/frontend/src/components/SimpleMap/constants.ts
+++ b/packages/frontend/src/components/SimpleMap/constants.ts
@@ -1,0 +1,4 @@
+// Opacity for map fills - used for regions, markers, and legend
+export const MAP_FILL_WITH_DATA_OPACITY = 0.7;
+export const MAP_FILL_NO_DATA_OPACITY = 0.5;
+export const MAP_FILL_NO_BASE_MAP_OPACITY = 1;
--- a/packages/frontend/src/components/SimpleMap/index.tsx
+++ b/packages/frontend/src/components/SimpleMap/index.tsx
@@ -2,7 +2,7 @@ import { MapChartLocation, MapChartType } from '@lightdash/common';
 import { Box, Divider, Stack, Text, UnstyledButton } from '@mantine/core';
 import { useClipboard } from '@mantine/hooks';
 import { IconCopy, IconMap } from '@tabler/icons-react';
-import { scaleSqrt } from 'd3-scale';
+import { scaleLinear, scaleSqrt } from 'd3-scale';
 import L from 'leaflet';
 import 'leaflet/dist/leaflet.css';
 import {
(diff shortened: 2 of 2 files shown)


almeidabbm:
* **#19649** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19626** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19626?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19625** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19625?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/SimpleMap/constants.ts
```diff
@@ -0,0 +1,2 @@
+// Opacity for map fills - used for regions, markers, and legend
+export const MAP_FILL_OPACITY = 0.7;
```
magnew: Good call. We have to remember to merge logic with the last color change I made. In that one we use 1.0 when there is no background tiles.

packages/frontend/src/components/SimpleMap/index.tsx
```diff
@@ -340,20 +341,18 @@ type SimpleMapProps = {
...
+    if (colors.length === 1) return () => colors[0];
+    if (max === min) return () => colors[Math.floor(colors.length / 2)];
+
+    // Create domain points that map evenly across the value range
+    const domain = colors.map(
+        (_, i) => min + (i / (colors.length - 1)) * (max - min),
     );
-    return colors[index];
+
+    return scaleLinear<string>().domain(domain).range(colors).clamp(true);
```
magnew, line 450: I think this linear interpolation is known to produce 'ugly' colors when interpolating in HEX space (I kinda ran into this before). 

Usually people use d3 interpolate here, but we don't have that and it might not be worth bringing in just for this. BUT, it looks like we have a colorUtils which uses ColorIO. Can we use the range functions in there? Added bonus is that it will give the same results as table calculations.

packages/frontend/src/components/SimpleMap/constants.ts
almeidabbm: Done in the latest version

packages/frontend/src/components/SimpleMap/index.tsx
almeidabbm: Sounds good!! Have created a new [PR](https://github.com/lightdash/lightdash/pull/19649) which uses oklab instead of srgb which d3 is using 🙇

PR #19630, opened 2026-01-21
refactor: create custom metrics out of period-over-period enablement 

joaoviana:
Relates to: [PROD-1164](https://linear.app/lightdash/issue/PROD-1164/refactor-pop-backend-so-that-previous-period-fields-are-metrics)

### Description:

Refactors period-over-period (PoP) metrics implementation to use explicit additional metrics instead of relying on naming conventions with suffixes. This approach:

1. Generates proper additional metrics for PoP previous-period values with appropriate metadata
2. Fixes formatting issues with PoP metrics by preserving the original metric's formatting options
3. Makes the code more maintainable by removing string manipulation and suffix-based logic

## Questions: 

1. Are we aligned that additionalMetrics are generated server‑side and should not be created in the UI for the global PoP flow? Currently, users click a button, the configuration gets sent to `metricQuery` and we get all metrics generated with a comparison metric. An alternative would be to have this on a metric level in the UI, for example:
  * User clicks on a metric in the results table, sets a PoP config, we compile that additional metric -> send to metricQuery -> query runs, all good

3. Are we ok with augmenting the response `metricQuery` to insert pop siblings into metrics + additionalMetrics so the frontend can treat them as selected without mutating user intent? Code would be in `prepareMetricQueryAsyncQueryArgs` which calls `addPopAdditionalMetricsToMetricQueryForResponse`. 

**Finally**, I think that creating a custom metric in the ui of type PoP would be good and more predictable, basically emulating what we have for additional metrics already. **Here is the PR that experiments with that and shows that it actually simplifies the code a lot**: https://app.graphite.com/github/pr/lightdash/lightdash/19634/refactor(wip)-allow-creation-of-additional-metrics-pop-for-each-column-metric I know we chatted about this alternative, but wasn't sure if there is some gotchas I need to think about. This PR also introduces a few more properties to `AdditionalMetric` which would mean that we can support in the YML in the future. 

global PoP demo

[CleanShot 2026-01-21 at 18.02.44.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/607613a8-39e1-474b-bdf8-c55382b7f4f5.mp4)

per-metric PoP demo

[CleanShot 2026-01-21 at 19.41.50.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/0d6fb282-8e69-419e-9997-c2a2cb556179.mp4)


joaoviana:
* **#19634** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19630** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19630?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19630?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/queryCompiler.ts
```diff
@@ -292,16 +293,11 @@ export const compileMetricQuery = ({
...
-        ...metricQuery.dimensions,
-        ...metricQuery.metrics.reduce<string[]>((acc2, metric) => {
-            acc2.push(metric);
-            if (metricQuery.periodOverPeriod) {
-                acc2.push(`${metric}${POP_PREVIOUS_PERIOD_SUFFIX}`);
-            }
-            return acc2;
-        }, []),
-    ];
+    const popMetricIds = (metricQuery.additionalMetrics ?? [])
```
ZeRego, line 296: is this const being used?

PR #19638, opened 2026-01-22
fix: small e2e test fixes

owlas:


owlas:
moved here: https://github.com/lightdash/lightdash/pull/19884

PR #19655, opened 2026-01-22
feat: use AsyncQueryService for AI agent metric queries

notgiorgi:
Closes: [PROD-2605: Agent should use new Async Query api](https://linear.app/lightdash/issue/PROD-2605/agent-should-use-new-async-query-api)

### Description:
Refactors AI agent query execution to use a new `executeMetricQueryAndGetResults` method in AsyncQueryService. This method handles the complete query lifecycle: execution, polling for completion, and fetching all result pages.

The implementation extracts raw values from the ResultRow format, making the data more directly usable by AI agents. The new polling mechanism includes exponential backoff, timeout handling, and proper error propagation.

### DIFF
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -12,6 +12,7 @@ import {
     type ApiGetAsyncQueryResults,
     assertIsAccountWithOrg,
     assertUnreachable,
+    type CacheMetadata,
     type CompiledCustomSqlDimension,
     CompiledDimension,
     type CompiledMetric,
@@ -110,7 +111,10 @@ import { compileMetricQuery } from '../../queryCompiler';
 import type { SchedulerClient } from '../../scheduler/SchedulerClient';
 import { wrapSentryTransaction } from '../../utils';
 import { metricQueryWithLimit as applyMetricQueryLimit } from '../../utils/csvLimitUtils';
-import { processFieldsForExport } from '../../utils/FileDownloadUtils/FileDownloadUtils';
+import {
+    processFieldsForExport,
+    streamJsonlData,
+} from '../../utils/FileDownloadUtils/FileDownloadUtils';
 import { safeReplaceParametersWithSqlBuilder } from '../../utils/QueryBuilder/parameters';
 import { PivotQueryBuilder } from '../../utils/QueryBuilder/PivotQueryBuilder';
 import {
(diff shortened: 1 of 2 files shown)


notgiorgi:
* **#19656** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19656?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19655** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19655?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -718,40 +719,7 @@ export class AsyncQueryService extends ProjectService {
...
-                    return undefined;
-                default:
-                    return assertUnreachable(status, 'Unknown query status');
-            }
-        };
-
-        // Wait for results to be ready
-        await pollForAsyncChartResults();
-
+        await this.pollForQueryCompletion({ account, projectUuid, queryUuid });
```
notgiorgi, line 725: What'd be a reasonable default timeout here?

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -3399,4 +3367,105 @@ export class AsyncQueryService extends ProjectService {
...
+
+        // 1. Execute
+        const { queryUuid, cacheMetadata, fields } =
+            await this.executeAsyncMetricQuery(args);
+
+        // 2. Poll
+        await this.pollForQueryCompletion({ account, projectUuid, queryUuid });
+
+        // 3. Fetch all pages
+        const allRows: ResultRow[] = [];
```
notgiorgi: this could become a potential memory bottleneck, but we have maxLimit set to 1000 rows by default

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: I would say the 5 min default seems good, specially since this is only being used @ the agent level at the moment. If we need it to be configurable we can add something in the future

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: Seems good for now, but you could also tap into the `resultsStream` like SQL Runner is doing. This way you won't need to paginate yourself in the agent, less memory usage, and bigger limits if needed.

Async query function is `getResultsStream`

notgiorgi:
### Merge activity

* **Jan 26, 12:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/19655).
* **Jan 26, 12:10 PM UTC**: @notgiorgi merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/19655).

PR #19668, opened 2026-01-22
fix: map render performance

magnew:
Closes: GLITCH-150

### Description:

**The problem**
Maps felt super laggy in the explore page when zooming and panning. The page was re-rendering on zoom and pan. 

**Solution**
We need to store the map extent so we can save the map with it, but we don't want to rerender when it the user move the map. We jsut want to pick up the current extent on save. 

**Specifically**
Map Extent Save Flow:
  1. MapExtentTracker captures pan/zoom → stores in Redux tempMapExtent
  2. At save time, selectUnsavedChartVersionForSave enriches with the extent
  3. After save, SimpleMap reads extent directly from savedChart (source of truth)

  Save Button Behavior:
  - Panning/zooming activates the save button (lightweight comparison of 3 numbers)
  - Other map config changes also activate it (normal deep comparison, minus extent fields)
  - No re-renders of the visualization on pan/zoom
  
 **To test**
Is perf reasonable when using the world map? 
Can you save it with a specific area centered?

### DIFF
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -208,6 +208,7 @@ export type MapChart = {
     noDataColor?: string;
     // Field configuration (controls tooltip visibility and custom labels)
     fieldConfig?: Record<string, MapFieldConfig>;
+    saveMapExtent?: boolean;
 };
 
 export enum FunnelChartDataInput {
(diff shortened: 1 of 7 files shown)


magnew:
* **#19668** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19668?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #19696, opened 2026-01-23
chore: add db + saved chart persistence for pop additional metrics

joaoviana:
Relates to: [PROD-1164: Refactor PoP backend so that previous period fields are metrics](https://linear.app/lightdash/issue/PROD-1164/refactor-pop-backend-so-that-previous-period-fields-are-metrics)

### Description:
Added support for Period-over-Period (PoP) metadata in additional metrics. This PR:

- Adds new columns to the `saved_queries_version_additional_metrics` table to store PoP-specific metadata:
  - `generated_by`: Identifies metrics generated by PoP functionality
  - `base_metric_id`: References the original metric used to create the PoP metric
  - `time_dimension_id`: Stores the time dimension used for comparison
  - `granularity`: Stores the time granularity (day, month, etc.)
  - `period_offset`: Stores the offset for comparison periods

- Creates a new migration file to add these columns to the database

- Updates the SavedChartModel to handle these new fields when creating and retrieving saved charts

- Removes the deprecated `period_over_period_config` field from saved charts, as PoP is now implemented through explicit additional metrics


joaoviana:
* **#19697** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19697?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#19698](https://github.com/lightdash/lightdash/pull/19698) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19698?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#19699](https://github.com/lightdash/lightdash/pull/19699) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19699?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#19696** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19696?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19696?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/database/entities/savedCharts.ts
```diff
@@ -237,6 +237,12 @@ export type DbSavedChartAdditionalMetric = {
     base_dimension_name: string | null;
     uuid: string;
     format_options?: CustomFormat | null; // JSONB
+    // PoP metadata (optional)
+    generated_by?: string | null;
```
joaoviana: instead of having a `type` "pop", we get the base metric's `type` because the pop metric derives from a metric with a defined type

packages/backend/src/models/SavedChartModel.ts
```diff
@@ -103,7 +103,6 @@ type DbSavedChartDetails = {
     pinned_list_uuid: string;
     dashboard_uuid: string | null;
     timezone: TimeZone | null;
-    period_over_period_config: PeriodOverPeriodComparison | null;
```
joaoviana, line 106: this will be deprecated. Should I drop the column, or add a comment within the migration?

tatianainama: I think we could drop the column directly, given this is beta, clean state :star:

packages/backend/src/database/entities/savedCharts.ts
tatianainama: makes sense! I found the `generated_by` name a bit confusing, I was expecting a "who generated this" first, rather than the feature that did it.  
What do you think of something like `generation_type` / `origin` instead? just to remove the 'actor' factor

PR #19791, opened 2026-01-29
feat: add dashboard tile and tab limits with configuration

ZeRego:
Closes: PROD-2751

### Description:

This PR adds dashboard tile and tab limits to prevent performance issues with large dashboards. It introduces two new configuration options:

- `maxTilesPerTab`: Limits the number of tiles per dashboard tab (default: 50)
- `maxTabsPerDashboard`: Limits the number of tabs per dashboard (default: 20)

The PR implements UI validations that prevent users from adding more tiles or tabs when limits are reached, showing helpful error messages. It also updates the AddChartTilesModal to display available slots and validate selections against the limit.

These limits can be configured via environment variables:

- `LIGHTDASH_DASHBOARD_MAX_TILES_PER_TAB`
- `LIGHTDASH_DASHBOARD_MAX_TABS_PER_DASHBOARD`

Note: Only adding validation in the FE, dont wan't to affect users using download/upload commands on existing dashboards. 

CleanShot 2026-01-29 at 10 47 20@2x
CleanShot 2026-01-29 at 12 02 53@2x
CleanShot 2026-01-29 at 12 02 35@2x

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
+++ b/packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
@@ -63,11 +63,23 @@ const AddTileButton: FC<Props> = ({
 
     // Get limits from health config
     const maxTilesPerTab = health.data?.dashboard?.maxTilesPerTab || 50;
+    const maxTabsPerDashboard =
+        health.data?.dashboard?.maxTabsPerDashboard || 20;
+
+    // Calculate current tabs count
+    const currentTabsCount = useMemo(() => {
+        return dashboardTabs ? dashboardTabs.length : 1; // Default dashboard has 1 tab
+    }, [dashboardTabs]);
 
     // Check if we can add a tile to current tab
     const canAddTile = useMemo(() => {
         return currentTabTilesCount < maxTilesPerTab;
     }, [currentTabTilesCount, maxTilesPerTab]);
+
+    // Check if we can add a new tab
+    const canAddTab = useMemo(() => {
+        return currentTabsCount < maxTabsPerDashboard;
+    }, [currentTabsCount, maxTabsPerDashboard]);
 
     const onAddTile = useCallback(
         (tile: Dashboard['tiles'][number]) => {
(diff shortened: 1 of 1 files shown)


ZeRego:
* **#19791** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19791?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19791?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/config/parseConfig.ts
```diff
@@ -1576,6 +1580,16 @@ export const parseConfig = (): LightdashConfig => {
                     ) || 3,
             },
         },
+        dashboard: {
+            maxTilesPerTab:
+                getIntegerFromEnvironmentVariable(
+                    'LIGHTDASH_DASHBOARD_MAX_TILES_PER_TAB',
+                ) || 50,
```
rephus, line 1590: good defaults. What would happen if an existing customer already has more than 50 ? would he not be allowed to add more (good) ? or would he throw errros on saving (bad)

*Update* I can see this is only on the AddTileButton component, in the frontend, so I assume is only when adding :+1:

packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
```diff
@@ -63,96 +93,131 @@ const AddTileButton: FC<Props> = ({
...
+                            onClick={() =>
+                                setAddTileType(DashboardTileTypes.HEADING)
+                            }
+                            leftSection={<MantineIcon icon={IconHeading} />}
+                        >
+                            Heading
+                        </Menu.Item>
+                    </Menu.Dropdown>
+                </Menu>
+            ) : canAddTab ? (
```
rephus, line 191: I thought our linter didn't like ternary expressions...

packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
```diff
@@ -63,96 +93,131 @@ const AddTileButton: FC<Props> = ({
...
-            <Menu
-                position="bottom"
-                withArrow
-                withinPortal
-                shadow="md"
-                width={200}
-            >
-                <Menu.Target>
+            {canAddTile ? (
+                <Menu
```
rephus, line 97: we should probably extract this into a separate component or function to make the logic more readable.

packages/backend/src/config/parseConfig.ts
ZeRego: correct 👍 only when adding tab/tile

Issue #19792, opened 2026-01-29
Remove ACL requirement for S3 bucket operations


jesshitchcock:
## Description 
We've had a customer report that dashboard CSV exports are failing with the error "The bucket does not allow ACLs" on their self-hosted instance. Individual chart downloads work fine, but full dashboard exports don't.

AWS disabled ACLs by default for new S3 buckets in 2023, treating them as a legacy feature in favor of more secure policy-based access control. Our current implementation appears to require ACL support for dashboard exports specifically, which creates compatibility issues for customers using modern AWS security practices.

## Expected behavior:
Dashboard CSV exports should work on S3 buckets that have ACLs disabled (the AWS default since 2023)

Requested change:
Remove the ACL dependency and use policy-based access control for all S3 operations, particularly dashboard exports.

### Steps to Reproduce the Bug or Issue

N/A

### Cloud or self-hosting

self-hosting

PR #19797, opened 2026-01-29
fix: remove file-level ACLs in s3 uploads

magnew:
Closes: #19792, PROD-2796

### Description:

The problem:
- Self-hosted customers are seeing an error when exporting dashboards "The bucket does not allow ACLs". It looks like this is because we still have a couple legacy uploads that set ACL: 'private', but GCP (and AWS) security policies now suggest setting uniformBucketLevelAccess, which disables ACLs.

**Solution**
- remove the ACL: 'private' settings from our s3 uploads. Our newer uploads dont include it anyway, and it's frowned upon by GCP and AWS.

**Also**
- We are reviewing security settings on Lightdash buckets to make sure they are in line with current GCP recommendations

### DIFF
--- a/packages/backend/src/clients/Aws/S3Client.ts
+++ b/packages/backend/src/clients/Aws/S3Client.ts
@@ -53,7 +53,6 @@ export class S3Client extends S3BaseClient {
                 Key: fileId,
                 Body: file,
                 ContentType: fileOpts.contentType,
-                ACL: 'private',
                 ContentDisposition: createContentDispositionHeader(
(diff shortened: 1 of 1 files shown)


magnew:
* **#19797** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19797?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19797?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #17538, opened 2025-10-20
add `lightdash config get-project` CLI command


oli-rmsy:
### Description

Users would like to see the current selected project without having the use the set command and have to re-select from a list of project (which can be quite long!)

A new CLI command that is similar to set-project would enable this, but instead it just return the current project.

### What problem does this solve?

An easier way to determine the current project you are building from

nathan-glean:
This would also be super helpful for when we write automations or make/taskfile commands too, as we can then use the current set project as a value / variable 
e.g. if you're using project X run this otherwise run that
Without being able to see what the current project is, we can't really make these kinds of scripts

ZeRego:
@nathan-glean for automation and CI/CD we recommend using the following environment variables to authenticate yourself and specify a project on each command:
- LIGHTDASH_API_KEY a personal access token you can generate in the app under the user settings
- LIGHTDASH_URL address for your running Lightdash instance
- LIGHTDASH_PROJECT a project uuid you want to use on that command

Example:
```
LIGHTDASH_API_KEY=946fedb74003405646867dcacf1ad345 LIGHTDASH_URL="https://your_subdomain_here.lightdash.cloud" LIGHTDASH_PROJECT="<uuid>" lightdash preview
```

ZeRego:
Done: https://github.com/lightdash/lightdash/pull/19809

PR #19805, opened 2026-01-29
feat: add list-projects to CLI

owlas:
Enable listing project

PR #19803, opened 2026-01-29
feat: display current project info in CLI commands

ZeRego:
Closes: PROD-2777

### Description:

Added project information logging to CLI commands to improve user experience. Now users can clearly see which project they are working with when running commands like deploy, download, upload, validate, refresh, generate-exposures, and preview. This helps prevent accidental operations on the wrong project.

The implementation adds a new `logProjectInfo` method to the GlobalState class that displays the current project UUID in a consistent format across all commands.

PR #19806, opened 2026-01-29
feat: add show-project command to CLI

ZeRego:
Closes: https://github.com/lightdash/lightdash/issues/17538

### Description:

Added a new `lightdash config show-project` command that displays information about the currently active project. The command shows the project name, UUID, type, and URL, fetching the most up-to-date information from the API when possible. If the API call fails, it falls back to showing the information available in the local configuration.

This makes it easier for users to identify which project they're currently working with without having to check the configuration files manually.

### DIFF
--- a/packages/cli/src/handlers/showProject.ts
+++ b/packages/cli/src/handlers/showProject.ts
@@ -1,0 +1,78 @@
+import { AuthorizationError } from '@lightdash/common';
+import { getConfig, setProject } from '../config';
+import GlobalState from '../globalState';
+import * as styles from '../styles';
+import { checkLightdashVersion } from './dbt/apiClient';
+import { getProject } from './dbt/refresh';
+
+type ShowProjectHandlerOptions = {
+    verbose: boolean;
+};
+
+export const showProjectHandler = async (
+    options: ShowProjectHandlerOptions,
+) => {
+    GlobalState.setVerbose(options.verbose);
+    await checkLightdashVersion();
+
+    const config = await getConfig();
+
+    if (!config.context?.apiKey || !config.context.serverUrl) {
+        throw new AuthorizationError(
+            `Not logged in. Run 'lightdash login --help'`,
+        );
+    }
+
+    const { project: projectUuid, serverUrl } = config.context;
+
+    if (!projectUuid) {
+        console.error(styles.warning('No active project configured.'));
+        console.error('');
+        console.error('To set an active project, run:');
+        console.error(`  ${styles.bold('lightdash config set-project')}`);
+        console.error('');
+        return;
+    }
+
+    try {
+        // Fetch project details from the API to get the most up-to-date information
+        const project = await getProject(projectUuid);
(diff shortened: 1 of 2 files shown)


ZeRego:
* **#19806** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19806?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#19803** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19803?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

rephus:
Are you aware of https://github.com/lightdash/lightdash/pull/19805 ? perhaps we could unify some logic...

packages/cli/src/handlers/showProject.ts
```diff
@@ -0,0 +1,78 @@
...
+
+type ShowProjectHandlerOptions = {
+    verbose: boolean;
+};
+
+export const showProjectHandler = async (
+    options: ShowProjectHandlerOptions,
+) => {
+    GlobalState.setVerbose(options.verbose);
+    await checkLightdashVersion();
```
rephus, line 16: probably no need to check lightdash version for this read-only command..

packages/cli/src/handlers/showProject.ts
```diff
@@ -0,0 +1,78 @@
...
+        console.error(`  UUID: ${styles.secondary(projectUuid)}`);
+        console.error(
+            `  URL:  ${styles.bold(`${serverUrl}/projects/${projectUuid}/home`)}`,
+        );
+        console.error('');
+
+        if (error instanceof Error) {
+            console.error(
+                styles.warning(
+                    `Warning: Unable to fetch current project details: ${error.message}`,
```
rephus, line 66: you can use getErrorMessage method and skip the if clause.

packages/cli/src/handlers/showProject.ts
```diff
@@ -0,0 +1,78 @@
...
+        console.error(`  Name: ${styles.success(project.name)}`);
+        console.error(`  UUID: ${styles.secondary(projectUuid)}`);
+        console.error(`  Type: ${styles.secondary(project.type)}`);
+        console.error(
+            `  URL:  ${styles.bold(`${serverUrl}/projects/${projectUuid}/home`)}`,
+        );
+        console.error('');
+
+        // update project name in config
+        await setProject(projectUuid, project.name);
```
rephus, line 52: :+1:

rephus: I hope this would not throw any error, otherwise the output is going to look weird..

ZeRego:
Already done in https://github.com/lightdash/lightdash/pull/19809

PR #19825, opened 2026-01-29
feat: map analytics events to prometheus metrics

Shifan-Gu:
Closes: 

### Context

In our self-hosted deployment, we need app-level metrics to autoscale our Trino cluster using KEDA. Since KEDA uses Prometheus queries as autoscaling triggers, we need a way to expose Lightdash analytics events as Prometheus metrics meanwhile keep the flexibility without touching the existing prometheus.ts file and application code.

### Solution

This PR introduces a `CustomMetricManager` that converts Lightdash analytics events to Prometheus counters through JSON configuration. This approach allows us to:

- Select which analytics events to expose as Prometheus metrics
- Configure metrics without modifying `prometheus.ts` or application code
- Maintain flexibility for deployment-specific autoscaling needs

### Implementation

- **Configuration-driven**: Define metrics in JSON (see `packages/backend/src/metrics/custom-metrics.config.example.json`)
- **Automatic event subscription**: Hooks into existing analytics system
- **Dynamic label extraction**: Extracts labels from event payloads based on configuration
- **Seamless integration**: Works with existing Prometheus setup

## Updates: 2026-01-29

@ZeRego 

- Moved all prometheus related files to prometheus folder.
- Renamed `CustomMetricManager` to `PrometheusEventsMetricsManager`.
- Moved prometheus events metrics manager init logic to PrometheusMetricsClass
- Add a dedicated event emitter for to subscribe to the specific configured of analytic events.
- Added validation for `extractLabels`
- Some other minor fixes

### DIFF
--- a/packages/backend/src/prometheus/PrometheusMetrics.ts
+++ b/packages/backend/src/prometheus/PrometheusMetrics.ts
@@ -549,37 +549,35 @@ export default class PrometheusMetrics {
         }
 
         try {
-            // Validate and sanitize the config path to prevent path traversal attacks
-            const sanitizedPath = path
-                .normalize(configPath)
-                .replace(/^(\.\.[/\\])+/, '');
-            const fullPath = path.isAbsolute(sanitizedPath)
-                ? sanitizedPath
-                : path.join(process.cwd(), sanitizedPath);
+            // Validate the config path to prevent path traversal attacks
+            // Resolve to absolute path first, then validate against base directory
+            const resolvedPath = path.resolve(process.cwd(), configPath);
+            const basePath = path.resolve(process.cwd());
 
             // Ensure the resolved path doesn't escape the working directory
-            const resolvedPath = path.resolve(fullPath);
-            const basePath = path.resolve(process.cwd());
-            if (!resolvedPath.startsWith(basePath)) {
+            if (
+                !resolvedPath.startsWith(basePath + path.sep) &&
+                resolvedPath !== basePath
+            ) {
                 throw new Error(
                     'Invalid configuration path: path traversal detected',
                 );
             }
 
-            if (!fs.existsSync(fullPath)) {
+            if (!fs.existsSync(resolvedPath)) {
                 Logger.warn(
-                    `PrometheusEventMetricManager config file not found: ${fullPath}`,
+                    `PrometheusEventMetricManager config file not found: ${resolvedPath}`,
                 );
                 return;
             }
 
-            const configContent = fs.readFileSync(fullPath, 'utf-8');
+            const configContent = fs.readFileSync(resolvedPath, 'utf-8');
             let jsonConfig: unknown;
             try {
                 jsonConfig = JSON.parse(configContent);
             } catch (parseError) {
                 Logger.error(
-                    `Failed to parse PrometheusEventMetricManager config JSON from ${fullPath}`,
+                    `Failed to parse PrometheusEventMetricManager config JSON from ${resolvedPath}`,
                     parseError,
                 );
                 return;
@@ -589,7 +587,7 @@ export default class PrometheusMetrics {
                 prometheusEventMetricsConfigSchema.safeParse(jsonConfig);
             if (!parsedConfig.success) {
                 Logger.error(
-                    `Invalid PrometheusEventMetricManager config from ${fullPath}`,
+                    `Invalid PrometheusEventMetricManager config from ${resolvedPath}`,
                     {
                         errors: parsedConfig.error.errors.map((issue) => ({
                             message: issue.message,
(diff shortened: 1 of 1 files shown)


packages/backend/src/prometheus/PrometheusEventMetricManager.ts
```diff
@@ -0,0 +1,281 @@
...
+     * Optional function to extract label values from the event payload
+     * If not provided, labels will be extracted directly from payload.properties
+     * using the labelNames as keys
+     */
+    extractLabels?: (payload: {
+        event: string;
+        userId?: string;
+        anonymousId?: string;
+        properties?: Record<string, unknown>;
+    }) => Record<string, string>;
```
joaoviana: `extractLabels` is declared as a function in the type, but configs are loaded from JSON and you can never provide a function via JSON 🤔 . The README acknowledges this too.

packages/backend/src/prometheus/PrometheusEventMetricManager.ts
```diff
@@ -0,0 +1,281 @@
...
+                try {
+                    prometheus.register.registerMetric(counter);
+                } catch (error) {
+                    Logger.warn(
+                        `Prometheus counter already registered: ${metricConfig.metricName}`,
+                        error,
+                    );
+                }
+
+                this.counters.set(metricConfig.eventName, counter);
```
joaoviana: if two metrics share the same `eventName` (e.g., total queries + queries by project), the second overwrites the first. Then `handleTrackEvent` at line 188 looks up by `eventName` and always gets the last one. Key should be `metricName` instead. wdyt? 🤔

packages/backend/src/prometheus/PrometheusEventMetricManager.ts
```diff
@@ -0,0 +1,281 @@
...
+
+                // Register with global Prometheus registry
+                try {
+                    prometheus.register.registerMetric(counter);
+                } catch (error) {
+                    Logger.warn(
+                        `Prometheus counter already registered: ${metricConfig.metricName}`,
+                        error,
+                    );
+                }
```
joaoviana: `new prometheus.Counter(...)` already auto-registers with the default registry. The manual `prometheus.register.registerMetric(counter)` on line 124 will always throw `"already registered"`, which the `try/catch` silently swallows. Either pass `registers: []` to the `Counter` constructor and register manually, or remove the manual registration call.

packages/backend/src/prometheus/PrometheusMetrics.ts
```diff
@@ -509,9 +518,74 @@ export default class PrometheusMetrics {
...
+            const configContent = fs.readFileSync(fullPath, 'utf-8');
+            const jsonConfig = JSON.parse(configContent);
+
+            // Validate config structure
+            if (!jsonConfig.metrics || !Array.isArray(jsonConfig.metrics)) {
+                Logger.error(
+                    'PrometheusEventMetricManager config must have a "metrics" array',
+                );
+                return;
+            }
```
joaoviana: The JSON config is parsed and checked for a metrics array, but individual metric entries aren't validated. A config like `{"metrics": [{"foo": "bar"}]}` would create a counter with name: `"undefined"`. Consider validating with Zod (which we have on the backend) or check minimum required fields (eventName, metricName, help, labelNames)

packages/backend/src/analytics/LightdashAnalytics.ts
```diff
@@ -1700,22 +1701,27 @@ type LightdashAnalyticsArguments = {
...
     options?: ConstructorParameters<typeof Analytics>[1];
+    eventEmitter?: EventEmitter;
 };
 
 export class LightdashAnalytics extends Analytics {
     private readonly lightdashConfig: LightdashConfig;
 
     private readonly lightdashContext: Record<string, AnyType>;
 
+    private readonly eventEmitter?: EventEmitter;
```
joaoviana, line 1712: `SchedulerApp` don't pass have an `eventEmitter`, so the I wonder if you also want scheduler-triggered events (scheduled query runs, etc.) to also be exposed as Prometheus metrics 🤔

packages/backend/src/prometheus/PrometheusMetrics.ts
```diff
@@ -509,9 +518,72 @@ export default class PrometheusMetrics {
...
+
+            if (!fs.existsSync(fullPath)) {
+                Logger.warn(
+                    `PrometheusEventMetricManager config file not found: ${fullPath}`,
+                );
+                return;
+            }
+
+            const configContent = fs.readFileSync(fullPath, 'utf-8');
+            const jsonConfig = JSON.parse(configContent);
```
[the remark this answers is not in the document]
joaoviana: dismissing this.

packages/backend/src/prometheus/PrometheusEventMetricManager.ts
Shifan-Gu: oh, that's a good point 👍

packages/backend/src/prometheus/PrometheusEventMetricManager.ts
Shifan-Gu: yea you are right, this is bit over engineered here, i will remove this part and mapping the payload values as labels should be good enough here.

Issue #19851, opened 2026-01-30
Allow customization and independent hiding of Y-axes in charts


jesshitchcock:
Currently, when attempting to hide a Y-axis in a chart with dual Y-axes, both axes are hidden simultaneously. Users need the ability to hide or customize only one Y-axis (e.g., hide the right Y-axis while keeping the left/secondary axis visible).

**Feature details:**
- Users should be able to independently show/hide each Y-axis in charts that support dual Y-axes.
- This will improve flexibility for users who want to display only one axis for clarity or design reasons.

**Why it's needed:**
- Enhances chart customization and user control.
- Addresses user feedback where hiding one axis currently hides both, which is not always desirable.

PR #19861, opened 2026-01-30
feat: Add independent left/right Y-axis visibility controls

hamzahc1:
### Description:

Closes: #19851 
This PR adds independent visibility controls for left (primary) and right (secondary) Y-axes in Cartesian charts, while maintaining backward compatibility with the legacy `showYAxis` property.

#### Changes:
- **Type definitions**: Added `showLeftYAxis` and `showRightYAxis` properties to `CompleteCartesianChartLayout` with `showYAxis` marked as legacy
- **UI controls**: Updated the Axes configuration panel to display separate checkboxes for left and right Y-axes when a secondary axis exists
- **State management**: Added `setShowLeftYAxis` and `setShowRightYAxis` callbacks in the chart config hook
- **Chart rendering**: Modified ECharts configuration to use independent axis visibility settings, with fallback to legacy `showYAxis` for backward compatibility

#### Key Features:
- Users can now toggle left and right Y-axes independently
- The right Y-axis control only appears when there are series on the secondary axis
- Full backward compatibility: existing charts using `showYAxis` will continue to work as before
- Legacy `showYAxis` acts as a fallback when new properties are not explicitly set

#### Backward Compatibility:
Charts created before this change will continue to work as expected. The legacy `showYAxis` property will control both axes until explicitly overridden by the new independent controls.

https://claude.ai/code/session_018TaQJoEL44sToVHhQZhhS3

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
@@ -130,8 +130,16 @@ export const Axes: FC<Props> = ({ itemsMap }) => {
         dirtyLayout?.showRightYAxis !== undefined
             ? dirtyLayout?.showRightYAxis
             : legacyShowYAxis;
-    // Determine if there are series on the secondary Y-axis (right axis)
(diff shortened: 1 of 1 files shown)


rephus:
Tested  main usecase  and backwards compatibility


Before

Screenshot from 2026-02-02 08-38-46

After


image

Issue #19858, opened 2026-01-30
Metrics appear in Spotlight despite global 'hide' setting in config during 'lightdash preview'


notgiorgi:
**Issue:**
When running `lightdash preview`, all metrics from the project are visible in the Spotlight, even though the global Spotlight visibility is set to `hide` in the Lightdash config file.

**Steps to Reproduce:**
1. Set the global Spotlight visibility to `hide` in the Lightdash config file.
2. Run `lightdash preview`.
3. Observe that all metrics are still visible in the Spotlight.

**Expected Behavior:**
Metrics should be hidden from the Spotlight as per the global config setting.

**Actual Behavior:**
All metrics are visible in the Spotlight regardless of the config setting.

nehaaprasad:
Hello! @notgiorgi 
please check out the PR!

PR #19868, opened 2026-02-01
fix: respect global spotlight default_visibility in preview

nehaaprasad:
Closes: #19858

### Description:

- When running `lightdash preview` with global Spotlight visibility set to hide in `lightdash.config.yml`, metrics were still appearing in Spotlight. 
- This fix ensures the compiler always applies the project’s default visibility (show or hide) to metrics and explores, so a global “hide” setting is respected and those items no longer show in Spotlight during preview.


joaoviana:
Thank you for your contribution @naaa760 but this is not open to external contribution.

Issue #19838, opened 2026-01-30
Metric row ordering doesn't respect user selection order in pivoted tables


jesshitchcock:
### Description

I'm seeing an issue with metric ordering when using "metrics as rows" in pivoted table charts.

## Expected behavior:
Metrics should display in the order I selected them (Profit → Revenue → Sales Count)

## Actual behavior:
The metrics get reordered automatically, and this order doesn't match my selection sequence. There's no way to control this - neither the sort controls nor drag-and-drop reordering in the field list affects the metric row order.

## Why this matters:
Users often need metrics in a specific logical sequence. For example, showing base metrics like Profit and Revenue before calculated ones like "Profit %" - having them auto-sorted breaks that narrative flow.

## Possible solutions:
Respect the original selection order from the metrics list, OR
Allow drag-and-drop reordering of metrics to control row order, OR
Add explicit sort controls for metric rows

### Steps to Reproduce the Bug or Issue

1. Add dimensions/metrics (e.g., Order Date Day + Profit, Revenue, Sales Count in that specific order)
2. Switch to table chart and pivot by the date dimension
3. Enable "metrics as rows" and note the order of the rows

Example: https://analytics.lightdash.cloud/share/SyMYXBFe2puRhnPapIar0
Loom: https://www.loom.com/share/aadd577563224b4591821e5fc3612928

### version

v0.2399.3

### Cloud or self-hosting

None

PR #19910, opened 2026-02-02
fix: respect metric order in pivoted table

magnew:
Closes: #19838, #19919

### Description:

When using 'metrics as rows', the pivoted table viz was not respecting the order specified by users in the results table. Now it does. 

Test steps:
- Pivot a table
- Use metrics as rows
- Order in the pivot table (vertical) should match order in the results table (horizontal)

Kapture 2026-02-02 at 18 42 41

### DIFF
--- a/packages/common/src/pivot/pivotQueryResults.ts
+++ b/packages/common/src/pivot/pivotQueryResults.ts
@@ -921,7 +921,8 @@ export const convertSqlPivotedRowsToPivotData = ({
     // Get unique base metrics from valuesColumns, preserving order
     const baseMetricsArray = filteredValuesColumns
         .map((col) => col.referenceField)
-        .filter((field, index, self) => self.indexOf(field) === index);
+        .filter((field, index, self) => self.indexOf(field) === index)
+        .sort((a, b) => columnOrder.indexOf(a) - columnOrder.indexOf(b));
 
     const headerValueTypes = getHeaderValueTypes({
         metricsAsRows: pivotConfig.metricsAsRows,


magnew:
* **#19912** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19912?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#19910** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/19910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/pivot/pivotQueryResults.ts
```diff
@@ -918,10 +918,10 @@ export const convertSqlPivotedRowsToPivotData = ({
...
     );
 
-    // Get unique base metrics from valuesColumns
-    const baseMetricsArray = Array.from(
-        new Set(filteredValuesColumns.map((col) => col.referenceField)),
-    );
+    // Get unique base metrics from valuesColumns, preserving order
+    const baseMetricsArray = filteredValuesColumns
+        .map((col) => col.referenceField)
+        .filter((field, index, self) => self.indexOf(field) === index);
```
almeidabbm: This didn't fix the issue when I tried, the set should be keeping order 🤔 the actual issue is that we need to sort the base metrics array with the column order that was set by the user like we do for the header values on line `1030`

https://github.com/user-attachments/assets/3aa10293-26b8-4427-8ac4-bf75b8368521

```suggestion
        .filter((field, index, self) => self.indexOf(field) === index)
        .sort((a, b) => columnOrder.indexOf(a) - columnOrder.indexOf(b));
```

almeidabbm (approved):

PR #19984, opened 2026-02-05
fix: Include custom dimensions in isStringDimension check

### LINKED ISSUE #19962: Bug: String custom dimensions missing from conditional formatting dropdown
## Description

String custom dimensions (custom SQL dimensions with string type, and bin dimensions) do not appear in the conditional formatting field dropdown in table visualisations. Numeric custom dimensions work correctly.

## Root cause

`isStringDimension` in `packages/common/src/utils/item.ts` gates on `isDimension()`, which checks for `field.fieldType === FieldType.DIMENSION`. `CustomDimension` types don't have a `fieldType` property, so they always fail this check.

`isNumericItem` does not have this gate — it just calls `getItemType()` directly — which is why numeric custom dimensions work fine.

## Expected behaviour

String custom dimensions should appear in the conditional formatting dropdown and use string operators, the same as regular string dimensions.

## Fix

Add an `isCustomDimension` check alongside `isDimension` in `isStringDimension`.

bikalsiwakoti:
Closes: #19962

## Summary
- Added `isCustomDimension` check alongside `isDimension` in `isStringDimension` function
- Fixes string custom dimensions not appearing in conditional formatting dropdown

## Root Cause
`isStringDimension` gated on `isDimension()`, which fails for `CustomDimension` types since they lack the `fieldType` property.

## Fix
Changed the check from:
`isDimension(item) && getItemType(item) === DimensionType.STRING`

To:
`(isDimension(item) || isCustomDimension(item)) && getItemType(item) === DimensionType.STRING`

image

### DIFF
--- a/packages/common/src/utils/item.ts
+++ b/packages/common/src/utils/item.ts
@@ -121,7 +121,10 @@ export const isStringDimension = (
     if (!item) {
         return false;
     }
-    return isDimension(item) && getItemType(item) === DimensionType.STRING;
+    return (
+        (isDimension(item) || isCustomDimension(item)) &&
+        getItemType(item) === DimensionType.STRING
+    );
 };
 
 export const getItemIcon = (


ZeRego:

Issue #20001, opened 2026-02-05
Table calculations used in grouped charts have overlapped series


jesshitchcock:
### Description

When using a table calculation in a grouped chart the series values overlap when stacking the chart. 

Here is an example: 

<img width="1508" height="654" alt="Image" src="https://github.com/user-attachments/assets/920f2049-3943-4fbe-93bf-3ce808e3d8ba" />

### Steps to Reproduce the Bug or Issue

Internal repro [here](https://analytics.lightdash.cloud/share/CwMM9bKHy8jfYlsvEg1_5)

### version

v0.2422.1

### Cloud or self-hosting

cloud

PR #19916, opened 2026-02-02
fix: scatter chart pivot config with metrics on both axes

hamzahc1:
Closes: #19911

### Description:
This PR fixes an issue where scatter charts with metrics on both the x and y axes were not properly generating pivot configurations. 

**The Problem:**
Scatter charts can have metrics on both axes (e.g., Order Count on x-axis, Revenue on y-axis). Previously, the `derivePivotConfigFromChart` function only handled dimensions as index columns and ignored metrics used on the x-axis, resulting in incomplete pivot configurations.

**The Solution:**
Added logic to recognize metrics used as x-axis fields in scatter charts and include them as index columns in the pivot configuration. This ensures that:
- The x-axis metric is properly included as an index column with `VizIndexType.CATEGORY`
- The y-axis metric is included as a values column
- Any groupBy dimensions are properly categorized as group by columns

**Changes:**
- Updated `getIndexColumn` function to handle metrics in addition to dimensions
- Added import for `isMetric` utility function and `VizIndexType` enum
- Added comprehensive test case covering scatter charts with metrics on both axes

https://claude.ai/code/session_01Y9tMVwLqVXRQQgu5VztNBF

PR #20013, opened 2026-02-05
fix: stacking not working with extra metrics

magnew:
<!-- Thanks so much for your PR, your contribution is appreciated! ❤️ -->

Closes: #20001 

### Description:
                                                                                                              
  **Summary**                                                                                                                      
Fixes a regression introduced in #19916 where stacked bar charts with pivoted data would display overlapping bars instead of properly stacking them.                                                                                                      
                                                                                                                               
  **The Problem**                                                                                                                  
                                                                                                                               
  PR #19916 added support for scatter charts with metrics on both axes by allowing metrics to become index columns. However,   
  the fix was too broad - it added all metrics to the index columns if they weren't explicitly in valuesColumns.               
                                                                                                                               
  This broke pivoted bar charts in a subtle way. Consider a stacked bar chart with:                                            
  - x-axis: shipping_method (dimension)                                                                                        
  - y-axis: % of total (table calculation that uses total_order_amount metric)                                                 
  - pivot: order_source                                                                                                        
                                                                                                                               
  The metric total_order_amount is in the query (the table calc needs it), but it's not in valuesColumns (only the table calc %
   of total is). So the buggy code added it as an extra index column.                                                          
                                                                                                                               
  Having multiple index columns changes how the pivot groups data - instead of one row per shipping method, you get one row per
   (shipping_method + total_order_amount) combination. This multiplies the data rows and breaks stacking because ECharts places
   the values at different array positions.                                                                                    
                                                                                                                               
  **The Fix**                                                                                                                      
                                                                                                                               
  Only treat a metric as an index column if it's explicitly the x-axis field (the original scatter chart use case)

### DIFF
--- a/packages/common/src/pivot/derivePivotConfigFromChart.test.ts
+++ b/packages/common/src/pivot/derivePivotConfigFromChart.test.ts
@@ -21,6 +21,7 @@ import {
 import {
     BinType,
     CustomDimensionType,
+    DimensionType,
     FieldType,
     MetricType,
     type ItemsMap,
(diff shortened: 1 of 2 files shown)


magnew:
* **#20013** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20013?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20013?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20127, opened 2026-02-10
fix: address custom dimension overlap on cartesian chart

joaoviana:
Closes: 

### Description:
This PR fixes an issue where vertical bar charts with numeric X-axis fields would display with overlapping bars instead of even spacing.

**Changes:**
- Modified the bottom axis type logic in `useEchartsCartesianConfig.ts` to force 'category' type for the X-axis when:
  - The chart is a vertical bar chart (not flipped)
  - The X-axis field is numeric (inferred as 'value' type)
  - The series contains bar chart elements

This ensures that numeric X-axis values are treated as discrete categories rather than continuous values, preventing bar overlap and maintaining consistent spacing across the chart.

The logic mirrors the existing behavior for horizontal bar charts, which already enforce 'category' type on the Y-axis for similar reasons.

https://claude.ai/code/session_01AhAKgFkLAefev54XzC6kWe


joaoviana:
don't like the solution provided.

Issue #20144, opened 2026-02-10
Allow unique colors per bar in single-series bar/horizontal bar charts


oli-rmsy:
## Problem

When a bar or horizontal bar chart has a single dimension and a single metric (no grouping), all bars are the same color. Users want to visually differentiate bars by having each bar a different color, but the only current workaround is adding a second dimension at the same grain to force a group-by — which is hacky.

**Customer request**: "I have a horizontal bar chart with just provider name and count. I want the bars to be colored differently."

## Proposed Solution

Add a chart setting (e.g. "Color bars by dimension value") that assigns a unique color from the org color palette to each bar in a single-series bar chart.

### Implementation sketch

ECharts supports per-data-point coloring via `itemStyle.color` on individual data entries:

```js
series: [{
  type: 'bar',
  data: [
    { value: 10, itemStyle: { color: '#5470c6' } },
    { value: 22, itemStyle: { color: '#91cc75' } },
    { value: 28, itemStyle: { color: '#fac858' } },
  ]
}]
```

In the Lightdash codebase:

1. **Add a config option** to `CartesianChart` or `EChartsConfig` in `packages/common/src/types/savedCharts.ts` (e.g. `colorByDimensionValue?: boolean`).
2. **Apply per-bar colors** in `packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts` around where bar-specific `itemStyle` is already set (~line 2450). When the option is enabled and there's a single series, map each data point to an object with `itemStyle.color` drawn from the org color palette — similar to how `applyRoundedCornersToStackData()` already wraps data points with per-item `itemStyle`.
3. **Expose the toggle** in the chart series config UI (`packages/frontend/src/components/DataViz/config/CartesianChartSeries.tsx`).

This approach is low-risk since it only changes data-point formatting for a single series when explicitly enabled.

PR #20146, opened 2026-02-10
feat: add color by value option for bar charts

oli-rmsy:
Closes: #20144

Screenshots of changes in comment below.

I have added a new button to control echarts config - this allows the user to change the way that colour is applied to charts with a single series.

It took a few iterations - but this button should now only show when a user has one series selected, and only for bar charts or horizontal bar charts. I've tested all of the above and it is working well.

## Summary
- Adds a **"Color by value"** toggle for bar chart series that assigns a unique color from the org color palette to each bar
- Enables visual differentiation in single-series bar/horizontal bar charts where all bars would otherwise be the same color
- Uses ECharts `itemStyle.color` function to apply per-data-point coloring based on palette index

## Changes
- `packages/common/src/types/savedCharts.ts` — Added `colorByValue?: boolean` to `Series` type
- `packages/common/src/visualizations/types/index.ts` — Added `colorByValue` to `EChartsSeries` type and updated `itemStyle.color` to accept function values
- `packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts` — When `colorByValue` is enabled on a non-stacked bar series, applies a color function that cycles through the org color palette per data point
- `packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SingleSeriesConfiguration.tsx` — Added "Color by value" checkbox for bar-type series in the series config panel

## Test plan
- [x] Create a bar chart with a single dimension and single metric (no grouping)
- [x] Open the series config panel and verify the "Color by value" checkbox appears for bar series
- [x] Enable the toggle and verify each bar gets a unique color from the org palette
- [x] Test with a horizontal bar chart (flip axes) to confirm it works in both orientations
- [x] Verify the toggle does NOT appear for line/area chart types
- [x] Verify stacked bar charts are unaffected (colorByValue only applies to non-stacked bars)
- [x] Save the chart and reload to confirm the setting persists

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -7658,11 +7658,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
@@ -8140,11 +8140,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
@@ -8159,11 +8159,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
@@ -8178,11 +8178,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
@@ -8197,11 +8197,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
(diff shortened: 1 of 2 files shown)


oli-rmsy:
See below images for the setting being off (default), and then on.

Screenshot 2026-02-10 at 16 41 37
Screenshot 2026-02-10 at 16 41 48

joaoviana:
1. Can you tell claude to run `pnpm generate-api`? It updates the api responses/payload types 
2. I don't really like how I can change a color (circle selector) and then have a checkbox that affects color on a separate area 
image

packages/common/src/types/savedCharts.ts
```diff
@@ -346,6 +346,7 @@ export type Series = {
     smooth?: boolean;
     markLine?: MarkLine;
     isFilteredOut?: boolean;
+    colorByValue?: boolean;
```
joaoviana, line 493: I would add a comment here

packages/common/src/visualizations/types/index.ts
```diff
@@ -391,6 +388,7 @@ export type EChartsSeries = {
         /** The field ID of the base metric this PoP series compares against */
         baseFieldId: string;
     };
+    colorByValue?: boolean;
```
joaoviana, line 395: same here

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SingleSeriesConfiguration.tsx
```diff
@@ -359,6 +359,23 @@ const SingleSeriesConfiguration: FC<Props> = ({
                             />
                         </Group>
                     )}
+                    {type === CartesianSeriesType.BAR &&
+                        !isGrouped &&
+                        isSingle && (
```
joaoviana, line 373: this is great, and intentional!

oli-rmsy:
@joaoviana made a change to the colour selector when the new option is selected. Not sure I'm a big fan, maybe it should just be grey with a line through it? Let me know what you think.
Screenshot 2026-02-10 at 17 17 43

Screenshot 2026-02-10 at 17 17 51

Possible option something like: 
image

joaoviana:
> @joaoviana made a change to the colour selector when the new option is selected. Not sure I'm a big fan, maybe it should just be grey with a line through it? Let me know what you think. <img alt="Screenshot 2026-02-10 at 17 17 43" width="385" height="128" src="https://private-user-images.githubusercontent.com/191847248/547809155-48c4e1da-7b75-41a8-a0ee-867feb003ef1.png?

Good point! Could you check if the component accepts a "disabled" option? Maybe there's already default styles for it. If not, go with that approach you just shared :) 
And finally, can you request it to add a `Tooltip` with a `IconHelpCircle` and some info on what it means? Worried it's not clear to the user of what it does 👍 @oli-rmsy

packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Series/SingleSeriesConfiguration.tsx
```diff
@@ -359,6 +360,32 @@ const SingleSeriesConfiguration: FC<Props> = ({
...
+                            <Group spacing="xs">
+                                <Tooltip
+                                    withinPortal
+                                    maw={300}
+                                    multiline
+                                    label="Assign a unique color to each bar based on its dimension value"
+                                >
```
joaoviana: this tooltip should be added to `Checkbox`'s `label` prop. 
label prop value should be something like
```tsx
label={<Group spacing='xs'> <Text fz='xs'> Color by value </Text> <Tooltip code...> <MantineIcon icon={<IconHelpCircle/>} color='ldGray.6' /> </Tooltip></Group>
```

magnew:
Talked to @torimwhaley -- closing this in favor of https://github.com/lightdash/lightdash/pull/20845

Issue #19606, opened 2026-01-21
Add export button for table charts


jesshitchcock:
### Description

There is no export button for table charts which means currently users can only export the results table not the configured chart. 

This is especially important when I want to export pivoted charts. 

It's also relevant when I want to export data from a chart and not include things like hidden fields.

Problem and workaround in this [loom](https://www.loom.com/share/b8cf45e7ba8b4a0299d23e0a6c9ea1bb) 

### What problem does this solve?

I can export table charts as tables as I have configured them in the chart config

PR #20215, opened 2026-02-11
fix: should be able to export table viz on chart view

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/19606

### Description:
Fixed image export for table visualizations by adding a check to ensure tables can be exported as CSV or XSLS on chart view page.

**Before**

**After**

### DIFF
--- a/packages/frontend/src/components/common/ChartDownload/ChartDownloadMenu.tsx
+++ b/packages/frontend/src/components/common/ChartDownload/ChartDownloadMenu.tsx
@@ -21,7 +21,10 @@ import {
 } from '../CollapsableCard/constants';
 import MantineIcon from '../MantineIcon';
 import ChartDownloadOptions from './ChartDownloadOptions';
-import { CHART_TYPES_WITHOUT_IMAGE_EXPORT } from './chartDownloadUtils';
+import {
+    CHART_TYPES_WITHOUT_DATA_EXPORT,
+    CHART_TYPES_WITHOUT_IMAGE_EXPORT,
+} from './chartDownloadUtils';
 
 export type ChartDownloadMenuProps = {
     getDownloadQueryUuid: (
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#20215** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20215?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20215?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20237** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20237?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/frontend/src/components/common/ChartDownload/ChartDownloadMenu.tsx
```diff
@@ -87,7 +87,8 @@ const ChartDownloadMenu: React.FC<ChartDownloadMenuProps> = memo(
         if (
             CHART_TYPES_WITHOUT_IMAGE_EXPORT.includes(
```
tatianainama, line 91: I think we need to remove the `ChartType.TABLE` from this list instead? 
I'm confused by the conditions 😅 because with this change is saying:
`if ['table', 'custom', ...].includes('table') && !('table' === 'table')` right?

almeidabbm: Yes that's correct and a good point, the reason why I decided not to do that was because the constant was called `CHART_TYPES_WITHOUT_IMAGE_EXPORT` for which table is in fact one of those 🤔 but this component also handles CSV/XLSX exports

tatianainama: ah got it! might be worth creating a new conditional for this case, but a comment can do it! it's a nit, but I'm sure I'm going to forget about this in a week :sweat_smile:

almeidabbm: Yeah, I ended up adding a new list `CHART_TYPES_WITHOUT_DATA_EXPORT` so it is clearer 🙇 ty @tatianainama

PR #20317, opened 2026-02-14
feat: support default credentials for Bedrock

dtaniwaki:
### Description:

We want to use Bedrock AI features in ECS/EKS deployments that rely on Task Roles or IRSA for authentication. Currently,
Bedrock requires explicit credentials (BEDROCK_API_KEY or BEDROCK_ACCESS_KEY_ID), which means we have to extract and pass keys manually even when the container already has IAM role-based access.

This PR adds BEDROCK_USE_DEFAULT_CREDENTIALS=true, which skips explicit credential configuration and lets the AWS SDK resolve credentials through its [credential provider change](https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html#credentialProviderChain). This covers Task Roles, IRSA, Instance Profiles, and other standard AWS authentication methods.

[AWS Default Credential Provider Chain]

When no explicit credentials are passed to the AWS SDK, it resolves credentials automatically by checking these sources in
 order:

1. Environment variables (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY)
2. Shared credentials file (~/.aws/credentials)
3. ECS Task Role
4. EC2 Instance Profile
5. EKS IRSA / Pod Identity
6. SSO credentials

Before: Only explicit BEDROCK_API_KEY or BEDROCK_ACCESS_KEY_ID + BEDROCK_SECRET_ACCESS_KEY were supported, so deployments
using Task Roles or IRSA had to extract and pass credentials manually.

After: BEDROCK_USE_DEFAULT_CREDENTIALS=true delegates credential resolution to the AWS SDK, enabling Task Roles, IRSA,
Instance Profiles, and any future credential sources without explicit key management.


ZeRego:

PR #20318, opened 2026-02-14
fix(athena): Athena warehouse in My Warehouse Connections

dtaniwaki:
## Summary

Athena warehouse type was missing from the My Warehouse Connections edit modal — trying to edit Athena credentials would just throw an error. This adds the missing form inputs (access key ID and secret access key) and the credential placeholders so the modal works properly for Athena connections.


ZeRego:
If this is a critical feature, please reach out to support.

dtaniwaki:
I see. Although this is a critical issue for Athena users, I'm using my own fork with this fix.

PR #20319, opened 2026-02-15
fix: build relation_name fallback when dbt ls does not populate it

dtaniwaki:
## Summary

Could you please review this when you get a chance?

When compiling a project using `dbt ls`, the `relation_name` field is not always populated even though the `database` and `schema` values resolved by `generate_schema_name` / `generate_database_name` macros are correctly set. This causes generated SQL queries to miss custom schema names, leading to incorrect table references on adapters like Athena.

This PR adds a `buildRelationName` helper to `convertTable()` that constructs a fully qualified table reference from the model's `database`, `schema`, and `alias` fields using the appropriate adapter quote character when `relation_name` is missing.

- Backwards-compatible: when `relation_name` is present, it is used as before
- Handles adapter-specific quoting (`` ` `` for BigQuery/Databricks, `"` for all others)

I've added tests as well — would appreciate it if you could take a look at those too. Thanks!


ZeRego:
If this is a critical feature, please reach out to support.

PR #20357, opened 2026-02-17
perf(athena): use ListTableMetadataCommand instead of GetTableMetadataCommand

dtaniwaki:
### Description:

The Athena getCatalog() method currently calls GetTableMetadataCommand once per table, which means a project with hundreds of tables triggers hundreds of individual AWS Glue API calls. This leads to rate limiting and slow catalog loading.

Replace per-table GetTableMetadataCommand calls with per-schema ListTableMetadataCommand calls. Requests are grouped by {database, schema}, and a single ListTableMetadataCommand retrieves all table metadata for  each schema in one paginated call, filtering to only the requested tables. For a project with 200 tables  across 5 schemas, this reduces API calls from 200 to 5.

### DIFF
--- a/packages/warehouses/src/warehouseClients/AthenaWarehouseClient.test.ts
+++ b/packages/warehouses/src/warehouseClients/AthenaWarehouseClient.test.ts
@@ -121,4 +121,204 @@ describe('AthenaWarehouseClient', () => {
             });
         });
     });
+
+    describe('getCatalog', () => {
+        const mockSend = jest.fn();
+
+        beforeEach(() => {
+            mockAthenaClient.mockImplementation(() => ({
+                send: mockSend,
+            }));
+        });
+
+        test('should use ListTableMetadataCommand to fetch catalog', async () => {
+            mockSend.mockResolvedValueOnce({
+                TableMetadataList: [
+                    {
+                        Name: 'orders',
+                        Columns: [
+                            { Name: 'id', Type: 'integer' },
+                            { Name: 'status', Type: 'varchar' },
+                        ],
+                    },
+                    {
+                        Name: 'users',
+                        Columns: [
+                            { Name: 'id', Type: 'integer' },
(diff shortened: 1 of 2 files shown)


ZeRego:
Thanks a lot for this contribution, @dtaniwaki — the intent (collapsing per-table Glue calls to avoid rate limiting) is a good one, and the tests are thorough. After review, though, we're going to hold off on merging this as-is due to a few concerns:

**1. Possible performance regression on large Glue schemas.** The previous code made exactly one `GetTableMetadataCommand` per *requested* table. This change paginates through *every* table in a schema via `ListTableMetadataCommand` (`MaxResults: 50`, the API max) and filters client-side. The "200 → 5 calls" win assumes most tables in each schema are actually requested. Athena/Glue is frequently used over data lakes with schemas containing thousands of tables where Lightdash only references a handful — in that case a schema with, say, 3000 tables but 3 requested goes from 3 calls to ~60 paginated calls, i.e. strictly worse. `ListTableMetadataCommand` does support a server-side `Expression` name filter, which could bound the worst case, but the sparse-vs-dense access pattern needs validation before we can be confident this is a net win across deployments.

**2. Changed partial-failure semantics.** Previously a `MetadataException` on a single table skipped only that table while the rest of the schema resolved. Now any list-level error drops the *entire schema* to `[]`. That's reasonable for a genuinely missing schema, but it's a behavior change that can silently hide otherwise-valid tables.

**3. Case-sensitivity in the client-side filter.** `tableFilter.has(tableMeta.Name)` is an exact-string match, whereas the old direct lookup matched by the requested `TableName`. If requested casing differs from what Glue returns (Glue normalizes to lowercase), tables could be dropped where the old path matched.

None of these are unfixable, but they need warehouse-side validation against real large-schema deployments before we'd take the change.

If this rate-limiting issue is impacting you in production and is important to prioritize, please reach out to our support team so we can track it properly and look into the right fix. Thanks again for taking the time to put this together.

PR #20367, opened 2026-02-17
feat: add support for required filters defined in model configuration

rephus:
Closes:  https://linear.app/lightdash/issue/PROD-2641/required-default-date-filter-remains-applied-in-sql-but-is-hidden-in

When a required filter is updated , the filter is still enforced in the backend. -> not expected
When a non required filter is updated, the old filter is replaced -> expected

So what we need to do is to disable these required filters on the UI, to prevent the user from updating

## After

image

### Description:
Adds support for required filters defined in model configuration. This PR:

1. Updates the `subscriptions.yml` model to include a sample required filter for `plan_name`
2. Improves the UI for required filters by:
   - Disabling field selection, operator selection, and value inputs for required filters
   - Updating the tooltip message to clarify that required filters cannot be modified
   - Adding a disabled menu icon with tooltip for required filters
   - Removing the conditional "Remove" menu item since required filters now have a separate UI treatment

This enhancement allows model owners to define filters that must be applied when querying the model, ensuring consistent and valid data access patterns.

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterRuleForm.tsx
@@ -87,7 +87,7 @@ const FilterRuleForm: FC<Props> = memo(
         );
         const isRequired = filterRule.required;
         const isRequiredLabel = isRequired
-            ? "This filter is a required filter.\n It can't be deleted, but the value can be changed."
+            ? 'This is a required filter defined in the model configuration and cannot be modified.'
             : '';
 
         if (!activeField) {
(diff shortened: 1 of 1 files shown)


rephus:
* **#20367** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20367?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20363, opened 2026-02-17
feat: add `Default user spaces` to project settings

Spissable:
Relates to: https://linear.app/lightdash/issue/GLITCH-170/by-default-users-with-interactive-viewer-access-and-above-should-have

### Description:
This PR adds the ability to toggle the "Default user spaces" via the project settings.

### Feature Overview:
- Admins can enable the feature on a project basis
- Default spaces are created lazily, when we fetch user info (when cache is missed)
- Anyone who can "manage SavedChart" gets the default space, this includes interactive viewer, but excludes viewer.
- The user gets explicit admin permissions on their space.

#### Admin

https://github.com/user-attachments/assets/f8078f57-6dc4-4952-8982-15f968cee209

#### Editor

https://github.com/user-attachments/assets/dd37812f-589d-4031-8b94-82bc71a07f65

#### Interactive Viewer

https://github.com/user-attachments/assets/285700c3-8e6d-4e6f-bc73-36eefa24de7c

#### Viewer

https://github.com/user-attachments/assets/e8cbd625-76b7-441d-8539-cc5ccaa05698

PR #20345, opened 2026-02-16
feat: allow managing permissions of nested default space

Spissable:
Relates to: https://linear.app/lightdash/issue/GLITCH-170/by-default-users-with-interactive-viewer-access-and-above-should-have

### Description:
This is the frontend part of the "getRootSpace" hack, that basically allows managing the permissions of default spaces, even though they are nested spaces.

PR #20335, opened 2026-02-16
feat: treat default user spaces as root

Spissable:
Relates to: https://linear.app/lightdash/issue/GLITCH-170/by-default-users-with-interactive-viewer-access-and-above-should-have

### Description:
Currently only root spaces can have dedicated permissions, all children spaces inherit their permissions.
While the nested space permissions are WiP this temporary workaround of treating default space permissions as roots, this should allow us to move forward already while still using the older permissions model.

PR #20334, opened 2026-02-16
feat: add migrations for user default spaces

Spissable:
Relates to: https://linear.app/lightdash/issue/GLITCH-170/by-default-users-with-interactive-viewer-access-and-above-should-have

### Description:
These are the base migrations needed to facilitate the new feature of "default user spaces".

PR #20369, opened 2026-02-17
feat: lazily create default user spaces on login

Spissable:
Relates to: https://linear.app/lightdash/issue/GLITCH-170/by-default-users-with-interactive-viewer-access-and-above-should-have

### Description:
This PR introduces the creation of the default user space as part of the log-in flow.
The fact that it's being done during the log-in flow is because it will handle some nice cases such as "user was created with a lower permission and later upgraded". Once the user then logs in with a permission high enough, they will automatically get their default space.

### DIFF
--- a/packages/backend/src/services/UserService.ts
+++ b/packages/backend/src/services/UserService.ts
@@ -1574,20 +1574,22 @@ export class UserService extends BaseService {
                     );
                 span.setAttribute('cacheHit', cacheHit);
 
-                if (!cacheHit) {
-                    void this.ensureDefaultUserSpaces(sessionUser).catch(
-                        (err) => {
-                            this.logger.error(
-                                'Failed to ensure default user spaces',
-                                err,
-                            );
-                        },
-                    );
-                }
-
                 return sessionUser;
             },
         );
+    }
+
+    async onLogin(user: {
+        userUuid: string;
+        organizationUuid?: string;
+    }): Promise<void> {
+        if (!user.organizationUuid) return;
+
+        const sessionUser = await this.findSessionUser({
+            id: user.userUuid,
+            organization: user.organizationUuid,
+        });
+        await this.ensureDefaultUserSpaces(sessionUser);
     }
 
     private async ensureDefaultUserSpaces(
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -803,6 +803,11 @@ export default class App {
                 id: user.userUuid,
                 organization: user.organizationUuid,
             });
+
+            // Fire-and-forget: run once-per-login setup tasks
+            void userService.onLogin(user).catch((err) => {
+                Logger.error('Failed to run onLogin tasks', err);
+            });
         });
 
         // Before each request handler we read `sess.passport.user` from the session store


Spissable:
* **#20369** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20369?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#20345](https://github.com/lightdash/lightdash/pull/20345) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20345?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#20363](https://github.com/lightdash/lightdash/pull/20363) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20363?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20343** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20343?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20380** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#20334** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20334?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#20335](https://github.com/lightdash/lightdash/pull/20335) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20335?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1607,6 +1607,112 @@ export class ProjectModel {
...
+        },
+    ): Promise<void> {
+        const existing = await this.database(SpaceTableName)
+            .where('is_default_user_space', true)
+            .where('created_by_user_id', user.userId)
+            .where('project_id', projectId)
+            .whereNull('deleted_at')
+            .first('space_uuid');
+
+        if (existing) return;
```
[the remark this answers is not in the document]
Spissable: The current code is optimized for the more common case "space already exists", rather than the un-common case "race condition of two executions trying to create the space". So the current code is the better trade-off imo.

packages/backend/src/services/UserService.ts
```diff
@@ -1567,11 +1573,78 @@ export class UserService extends BaseService {
...
+                                'Failed to ensure default user spaces',
+                                err,
+                            );
+                        },
+                    );
+                }
+
                 return sessionUser;
             },
         );
```
almeidabbm, line 1579: Might lead to perf issues to do this in every `findSessionUser`

packages/backend/src/models/ProjectModel/ProjectModel.ts
```diff
@@ -1607,6 +1607,112 @@ export class ProjectModel {
...
+                SpaceTableName,
+                `${SpaceTableName}.project_id`,
+                `${ProjectTableName}.project_id`,
+            )
+            .where(
+                `${OrganizationTableName}.organization_uuid`,
+                organizationUuid,
+            )
+            .where(`${ProjectTableName}.has_default_user_spaces`, true)
+            .where(`${SpaceTableName}.name`, 'Default User Spaces')
```
almeidabbm: Could use the same constant as the prev PR

packages/backend/src/services/UserService.ts
Spissable: handled in the `passport.deserializeUser` as discussed.

packages/backend/src/services/UserService.ts
```diff
@@ -1567,11 +1573,83 @@ export class UserService extends BaseService {
...
+        userUuid: string;
+        organizationUuid?: string;
+    }): Promise<void> {
+        if (
+            !this.lightdashConfig.defaultUserSpaces.enabled ||
+            !user.organizationUuid
+        )
+            return;
+
+        const sessionUser = await this.userModel.findSessionUserByUUID(
```
almeidabbm: This might get all sessions regardless of org. we can use `findSessionUser` which is what's used in the deserialize function and accepts the org? Since you already explicitly check for the org uuid to exist we should use that one

PR #20400, opened 2026-02-18
fix: retry query on network errors

Shifan-Gu:
Closes: 

### Description:
In our self-hosted deployment when the trino cluster is not scaled to a desired size during the peak hours we kept getting 5xx errors from the queries requests, to mitigate this problem, this PR adds retries to react-query. I tested it a bit and seems to work well. Current implementation retries 3 times with increasing delays between retries.

### DIFF
--- a/packages/frontend/src/features/sqlRunner/hooks/useSavedSqlChartResults.tsx
+++ b/packages/frontend/src/features/sqlRunner/hooks/useSavedSqlChartResults.tsx
@@ -17,6 +17,7 @@ import { useQuery } from '@tanstack/react-query';
 import { useCallback } from 'react';
 import getChartDataModel from '../../../components/DataViz/transformers/getChartDataModel';
 import { useOrganization } from '../../../hooks/organization/useOrganization';
+import { useQueryRetryConfig } from '../../../hooks/useQueryRetry';
 import {
     getDashboardSqlChartPivotChartData,
     getSqlChartPivotChartData,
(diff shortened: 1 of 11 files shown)


packages/frontend/src/features/sqlRunner/hooks/useSavedSqlChartResults.tsx
```diff
@@ -17,6 +17,7 @@ import { useQuery } from '@tanstack/react-query';
 import { useCallback } from 'react';
 import getChartDataModel from '../../../components/DataViz/transformers/getChartDataModel';
 import { useOrganization } from '../../../hooks/organization/useOrganization';
+import { defaultRetryConfig } from '../../../hooks/useQueryRetry';
```
joaoviana: @Shifan-Gu could you please check if it's possible to turn this off/on? 
I'm thinking something like `env var` -> `/health` endpoint -> `useHealth()`

I'll test this in the meantime

Also, do you only want this on dashboard + dashboard charts + sql  runner charts? Asking because I believe this doesn't cover standard charts

Shifan-Gu: Hi @joaoviana, 

It can't be turned off looks like, i can add a toggle to it as part of the config loading from env variable, 
I think it should be fine for now to just cover the sql charts, we already have this change running on our self hosted deployment now, haven't heard of any user complains yet, if in future we need add it to other charts i can send another PR request.

joaoviana: Awesome, thank you @Shifan-Gu 🙏

PR #20402, opened 2026-02-18
chore: use inheritance chain to resolve space access

magnew:
### Description:

  - Replace root-space-only permission resolution in SpacePermissionService.getSpacesCaslContext() with inheritance-chain-aware
   resolution:                                                                                                               
    - Direct access is aggregated from all spaces in the chain                                                     
    - Project/org access is only included when the chain reaches a root that inherits                            
    - isPrivate is derived from !inheritsFromOrgOrProject (CASL rules unchanged)

Specifically 

- `getSpacesCaslContext` in `SpacePermissionService.ts`:

  - Gets inheritance chain per space via getInheritanceChain()
  - Collects all unique space UUIDs from chains for batch direct access query
  - Identifies root spaces that inherit from org/project
  - Batch-fetches direct, project, and org access in parallel
  - Computes isPrivate = !inheritsFromOrgOrProject
  - Passes resolved access to CASL context

- `SpacePermissionModel.getInheritanceChain()` walks ancestors using ltree, stops at first inherit_parent_permissions=false, returns SpaceInheritanceChain with chain + inheritsFromOrgOrProject

 With these inputs to `getSpacesCaslContext`
 
```
[
    "220096c5-c0d7-4586-912b-6d8e18b17e00",
    "d0b363c6-9cb6-4656-a605-516d06e044d7",
    "34426ca8-9426-4e24-a288-1ac9de4e33d6",
    "194b4a70-fb78-43b6-b988-05e7c497be1c",
    "7e8a7c8b-e768-4f16-b445-70c0263e1d01",
    "d5b3be0f-8189-4194-abf6-7666eb38c005",
    "5b73b419-6466-4267-8e82-3f52bb5e9c9a",
    "726a7aae-7e40-44bd-b76b-b7749468d35a",
    "7e0eabe4-fc21-4db5-930d-5c55b6b7b06d",
    "e112f243-9316-4326-911f-190fc4ce6bb5",
    "525bea73-0b94-4db5-ba5d-2dfee1c52649",
    "8128806d-3f2a-4fae-8021-d3e6222e0e7b",
    "addca29f-e273-442c-895d-85b83274796c",
    "2828e787-c0cf-46f7-88dd-6fc340908575",
    "9b37addc-61d6-45a8-af99-4a8c5306e30c",
    "b9281408-1923-4cb5-91da-23776cfdb21b",
    "70f981ae-2773-4a88-a2c4-956452ae3c19",
[...]

PR #20398, opened 2026-02-18
chore: bring back getInheritanceChain

magnew:
### Description:

Bring back getInheritanceChain

PR #20407, opened 2026-02-18
chore: use inherit instead of isPrivate in space checks

magnew:
### Description:

Rename isPrivate to `inheritParentPermissions`

### DIFF
--- a/packages/backend/src/services/SpaceService/SpacePermissionService.ts
+++ b/packages/backend/src/services/SpaceService/SpacePermissionService.ts
@@ -17,7 +17,7 @@ import { BaseService } from '../BaseService';
 export type SpaceAccessContextForCasl = {
     organizationUuid: string;
     projectUuid: string;
-    isPrivate: boolean;
+    inheritParentPermissions: boolean;
     access: SpaceAccess[];
 };
 
@@ -241,7 +241,7 @@ export class SpacePermissionService extends BaseService {
             result[spaceUuid] = {
                 organizationUuid: space.organizationUuid,
                 projectUuid: space.projectUuid,
-                isPrivate,
+                inheritParentPermissions: inheritsFromOrgOrProject,
                 access,
             };
         }
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -231,7 +231,7 @@ export class AsyncQueryService extends ProjectService {
                 subject('SavedChart', {
                     organizationUuid: savedChart.organization.organizationUuid,
                     projectUuid: savedChart.project.projectUuid,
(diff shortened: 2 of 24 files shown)


magnew:
* **#20407** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#20402** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20402?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#20448](https://github.com/lightdash/lightdash/pull/20448) <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20448?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#20398** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20398?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

magnew:
Closing for something better

packages/backend/src/services/SpaceService/SpacePermissionService.ts
```diff
@@ -17,7 +17,7 @@ import { BaseService } from '../BaseService';
 export type SpaceAccessContextForCasl = {
     organizationUuid: string;
     projectUuid: string;
-    isPrivate: boolean;
+    inheritParentPermissions: boolean;
```
almeidabbm, line 20: This should be called `inheritsFromOrgOrProject` so it isn't confused with space `inheritParentPermission` boolean.

This is because the `inheritParentPermission` in a space means that this space inherits from the space above, but the `inheritFromOrgOrProject` boolean we return from the chain will mean that this chain inherits will end up inheriting from the org or project which is different from the space boolean.

For example: if I have a child space that inherits from the parent, but the parent (which is root) doesn't inherit. Then:
- the child space's inherit_parent_permission will be `true`
- the **chain** `inheritFromOrgOrProject` will be false, because the root space broke inheritance before getting to org/project perms

packages/backend/src/services/SpaceService/SpacePermissionService.ts
```diff
@@ -241,7 +241,7 @@ export class SpacePermissionService extends BaseService {
             result[spaceUuid] = {
                 organizationUuid: space.organizationUuid,
                 projectUuid: space.projectUuid,
-                isPrivate,
+                inheritParentPermissions: inheritsFromOrgOrProject,
```
almeidabbm, line 244: This logic is correct, since we're using `inheritsFromOrgOrProject`. Believe it is just a naming thing

PR #20419, opened 2026-02-18
fix: update AI agent test cases to use specific date ranges instead of relative dates

notgiorgi:
### Description:

Updated AI agent test cases to use specific date ranges instead of relative time periods. This change replaces expressions like "last 3 months" or "last 5 years" with explicit date ranges like "Q3 2024" or "between 2023-01-01 and 2024-12-31". Also removed the moment.js dependency from the test file since it's no longer needed for calculating relative dates.

The update ensures that test cases have consistent, predictable date ranges that won't change as time passes, making tests more stable and deterministic.

### DIFF
--- a/packages/backend/src/ee/services/ai/agents/tests/testCases.ts
+++ b/packages/backend/src/ee/services/ai/agents/tests/testCases.ts
@@ -1,5 +1,4 @@
 import { AiAgent, CatalogType } from '@lightdash/common';
-import moment from 'moment';
 import { CatalogSearchContext } from '../../../../../models/CatalogModel/CatalogModel';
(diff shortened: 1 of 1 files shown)


notgiorgi:
* **#20419** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20435, opened 2026-02-19
fix: maintain pivot series order based on query sort

rephus:
Closes: https://linear.app/lightdash/issue/PROD-2927/sorted-grouped-dimension-does-not-respect-newly-introduced-category

  Problem: mergeExistingAndExpectedSeries always preserved the existing series order from the
  saved chart config. When the query is sorted by a grouped dimension, the series in the chart
  legend/config panel appeared in whatever order they were first encountered, not in the sorted
  order. Users had to manually reorder series every time.

  Fix (packages/frontend/src/hooks/cartesianChartConfig/utils.ts):
  1. Simplified the new series insertion — just append new series to the merged list (with
  inherited yAxisIndex for pivot series in the same group)
  2. Added a final sort — after merging, all series are reordered to match expectedSeriesMap key
  order. Since expectedSeriesMap is built from pivotDetails.valuesColumns (which the backend sorts
   by columnIndex from SQL DENSE_RANK()), the series now respect the query sort order

  This also handles the original PROD-2927 issue (new categories appearing at the bottom) because
  the sort at the end places everything in the correct position.

  Files changed:
  - packages/frontend/src/hooks/cartesianChartConfig/utils.ts — simplified merge + added sort
  - packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.mock.ts — updated
  expected output to reflect new ordering
  - packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.test.ts — added test
  for new category sorted position

--- 

### Before

Screenshot from 2026-02-19 09-16-18

### After
Screenshot from 2026-02-19 09-26-06

Screenshot from 2026-02-19 09-26-12

image

image

### Description:
Fixed chart series ordering to respect query sort order. Previously, when new pivot categories were added to a chart, they would be appended to the end of their group rather than inserted in the correct sorted position.

The PR refactors the `mergeExistingAndExpectedSeries` function to:
1. First collect all series (existing and new)
2. Then sort them according to the order defined in the expected series map

This ensures that when a new pivot category (like "b" between "a" and "c") is added, it appears in the correct position rather than at the end of the series group.

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/common/GrabIcon.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/common/GrabIcon.tsx
@@ -26,7 +26,6 @@ export const GrabIcon: FC<Props> = ({
                 }),
                 opacity: disabled ? 0.3 : 0.6,
                 cursor: disabled ? 'default' : 'grab',
-                '&:hover': { opacity: disabled ? 0.3 : 1 },
             }}
         >
             <MantineIcon color="ldGray.6" icon={IconGripVertical} />


rephus:
* **#20435** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/VisualizationConfigs/common/GrabIcon.tsx
```diff
@@ -1,26 +1,45 @@
...
+    const icon = (
+        <Box
+            {...dragHandleProps}
+            style={{
+                ...(hovered !== undefined && {
+                    visibility: hovered ? 'visible' : 'hidden',
+                }),
+                opacity: disabled ? 0.3 : 0.6,
+                cursor: disabled ? 'default' : 'grab',
+                '&:hover': { opacity: disabled ? 0.3 : 1 },
```
tatianainama:
```suggestion
```

nit: we could remove this, as the new Mantine's Box doesn't work with inline pseudo selectors

tatianainama (approved):
works great! 💯

PR #20440, opened 2026-02-19
fix: return the full direct user access list instead of list filtered to the user

almeidabbm:
Closes: https://linear.app/lightdash/issue/GLITCH-173/fix-access-lists-in-space-summaries

### Description:
Added a new `getDirectAccessUserUuids` method to the `SpacePermissionService` that efficiently retrieves user UUIDs with direct access to spaces.

This fixes an issue where the `access` list in `SpaceSummaries` was being populated only with the user uuid if the user had direct access to the space - this happened because we were using the full context filtered just to the user. 

This array should contain the full direct access list (all user uuids with direct access) so that we can show the correct tooltip on the frontend.

**Space**

**Before**

**After**

### DIFF
--- a/packages/backend/src/services/ContentService/ContentService.ts
+++ b/packages/backend/src/services/ContentService/ContentService.ts
@@ -31,10 +31,7 @@ import { BaseService } from '../BaseService';
 import { DashboardService } from '../DashboardService/DashboardService';
 import { SavedChartService } from '../SavedChartsService/SavedChartService';
 import { SavedSqlService } from '../SavedSqlService/SavedSqlService';
-import type {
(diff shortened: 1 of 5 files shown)


almeidabbm:
* **#20440** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20468, opened 2026-02-19
feat: skip content type check on lightdash upload --force

joaoviana:
Closes: n/A

### Description:
Added a `force` parameter to chart, SQL chart, and dashboard creation endpoints to allow forcing updates even when no changes are detected. This enables users to override the default behavior where resources with no changes are skipped during promotion.

The implementation:
1. Added the `force` parameter to the controller methods
2. Passed the parameter through to the CoderService methods
3. Modified the promotion logic to convert NO_CHANGES actions to UPDATE actions when force is enabled
4. Updated the CLI to pass the force flag to the API

```

  1. Regular charts (upsertChart): promoteService.getChartChanges() could return NO_CHANGES based on timestamp/name/description comparison → now overridden to UPDATE when force=true
  2. Charts inside dashboards (upsertDashboard): promoteService.getPromotionDashboardChanges() could return NO_CHANGES for individual chart tiles → now also overridden to UPDATE when
  force=true
  3. Dashboards themselves: Already always UPDATE (no change needed — there's even a TODO comment about this at line 1534)
  4. SQL charts (upsertSqlChart): Already always CREATE or UPDATE directly — never goes through promoteService, so NO_CHANGES was never possible

  So with --force, every content type will always be written to the database — no server-side "nothing changed" skipping.
```

### DIFF
--- a/packages/backend/src/controllers/projectController.ts
+++ b/packages/backend/src/controllers/projectController.ts
@@ -1067,6 +1067,7 @@ export class ProjectController extends BaseController {
         chart: Omit<ChartAsCode, 'chartConfig' | 'description'> & {
             skipSpaceCreate?: boolean;
             publicSpaceCreate?: boolean;
+            force?: boolean;
             chartConfig: AnyType;
             description?: string | null; // Allow both undefined and null
         },
@@ -1085,6 +1086,7 @@ export class ProjectController extends BaseController {
                 },
                 chart.skipSpaceCreate,
                 chart.publicSpaceCreate,
+                chart.force,
             ),
         };
     }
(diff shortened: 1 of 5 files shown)


joaoviana:
* **#20468** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20468?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20504, opened 2026-02-20
fix: replace n+1 query for inheritance chain

Spissable:
Related to: https://linear.app/lightdash/issue/GLITCH-177/investigate-performance-bottlenecks-of-inherited-access-resolution

### Description:
Refactoring the N+1 queries in the inheritance chain resolution with 1 batch version.
Refactoring the 2 queries into 1 single query.

### Benchmarks:
Setup: 200 spaces (100 deep and 100 wide), 100 users, direct permissions etc...

AVG over 100 sequential requests:

Inheritance original: 73ms
Legacy (root perm): 64ms
Refactoring (2 queries): 47ms
Refactoring (single query): 40ms

### DIFF
--- a/packages/backend/src/models/SpacePermissionModel.ts
+++ b/packages/backend/src/models/SpacePermissionModel.ts
@@ -439,32 +439,21 @@ export class SpacePermissionModel {
                     name: string;
                     inherit_parent_permissions: boolean;
                     parent_space_uuid: string | null;
-                }[] = await this.database
-                    .raw(
-                        `SELECT
-                        leaf.space_uuid AS requested_space_uuid,
-                        a.space_uuid,
-                        a.name,
-                        a.inherit_parent_permissions,
-                        a.parent_space_uuid
-                    FROM ${SpaceTableName} leaf
-                    JOIN ${SpaceTableName} a
-                        ON a.path @> leaf.path
-                        AND a.project_id = leaf.project_id
-                    WHERE leaf.space_uuid IN (${spaceUuids.map(() => '?').join(', ')})
-                    ORDER BY leaf.space_uuid, nlevel(a.path) DESC`,
-                        spaceUuids,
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/SpacePermissionModel.ts
```diff
@@ -416,68 +416,119 @@ export class SpacePermissionModel {
...
-                    .orderByRaw(`nlevel(${SpaceTableName}.path) DESC`);
+                const result: Record<string, SpaceInheritanceChain> = {};
+                for (const requestedUuid of spaceUuids) {
+                    const myAncestors =
+                        ancestorsBySpace.get(requestedUuid) ?? [];
+                    if (myAncestors.length === 0) {
+                        // Space not found in DB — skip silently (caller
+                        // will notice the missing key in the result map).
+                        // eslint-disable-next-line no-continue
+                        continue;
```
Spissable, line 477: To be discussed

magnew: Did we make a decision here? I think you said there were cases to be made for either approach, but I don't remember if we decided. I agree that throwing here seems a little aggressive -- we would basically fail the whole permission check, yea?

In this continue case, I think we would never be *adding* permissions the user shouldn't have, so it seems safe. They might not get some extra permissions they should have. But this is a pretty rare case, right?

Spissable: I think there was never a proper discussion and the previous fail-approach might've just been a random choice by Claude.

I'm also leaning more towards not failing for the reasons you mentioned. The main case that I can think of why we even run into this scenario would be space deletion with an unlucky timing, in which case failing the whole request doesn't really make sense.

magnew: Yea, definitely an edge case. As long as it's true that the missing entry will never result in added access.

magnew (approved):
Nice

PR #20514, opened 2026-02-20
fix: cache field conversion in PostgresClient query streaming

pimmey:
Closes: 

Optimized PostgresWarehouseClient performance by implementing field conversion caching and refactoring the field conversion method. 

The `convertQueryResultFields` method now uses `Object.fromEntries` and `map` instead of `reduce` for cleaner code. More importantly, added caching logic in the query stream processing to avoid redundant field conversions - since `result.fields` maintains the same array reference across all rows in a query, the converted fields are now cached and reused rather than being recalculated for every row.

This reduces unnecessary computation during large query result processing by caching the field type conversions.

### DIFF
--- a/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/PostgresWarehouseClient.ts
@@ -296,8 +296,6 @@ export class PostgresClient<
                         string,
                         { type: DimensionType }
                     > | null = null;
-                    let cachedFieldsRef: QueryResult<AnyType>['fields'] | null =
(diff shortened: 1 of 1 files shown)


pimmey:
* **#20514** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20534, opened 2026-02-23
feat: pre-aggregate virtual explores

notgiorgi:
### Description:

This PR introduces pre-aggregate virtual explore generation and visibility controls for pre-aggregate explores.

**Key Changes:**

- **Pre-aggregate Virtual Explore Generation**: Added functionality to automatically generate internal pre-aggregate explores when `ENABLE_PRE_AGGREGATE_VIRTUAL_EXPLORES` is enabled. These explores are built from pre-aggregate definitions and include only supported metric types (SUM, COUNT, MIN, MAX).

- **Visibility Controls**: Implemented environment variable-based visibility controls for pre-aggregate explores:

  - `ENABLE_PRE_AGGREGATE_DEBUG_VIEW` - Shows pre-aggregate explores in the UI for debugging
  - `ENABLE_PRE_AGGREGATE_INTERNAL_VISIBILITY` - Internal visibility flag for pre-aggregate explores
  - Pre-aggregate explores are filtered out by default unless these flags are enabled

- **Pre-aggregate Query Processing**: Extended async query service to process pre-aggregates when `ENABLE_PRE_AGGREGATES` or `ENABLE_PRE_AGGREGATE_DRY_RUN` is enabled.

- **Explore Filtering**: Updated explore filtering logic in both SearchModel and ProjectService to respect pre-aggregate visibility settings while ensuring they're included when debug flags are enabled.

- **Metric Rewriting**: Pre-aggregate explores automatically rewrite metrics to use appropriate aggregate functions (e.g., SUM for pre-aggregated values) and exclude unsupported metric types like AVERAGE and custom SQL metrics.

- **Time Dimension Handling**: Added logic to handle time dimensions in pre-aggregate explores, filtering out time intervals finer than the rollup granularity and mapping to appropriate column names.

### DIFF
--- a/packages/backend/src/models/SearchModel/index.ts
+++ b/packages/backend/src/models/SearchModel/index.ts
@@ -1249,21 +1249,36 @@ export class SearchModel {
             .limit(1);
 
         if (explores.length > 0 && explores[0].explores) {
+            const includePreAggregateDebugExplores =
+                process.env.ENABLE_PRE_AGGREGATE_DEBUG_VIEW === 'true' ||
+                process.env.ENABLE_PRE_AGGREGATE_INTERNAL_VISIBILITY === 'true';
             return explores[0].explores.filter(
                 (explore: Explore | ExploreError) => {
+                    if (
+                        !includePreAggregateDebugExplores &&
+                        explore.type === ExploreType.PRE_AGGREGATE
+                    ) {
+                        return false;
+                    }
                     if (tableSelection.type === TableSelectionType.WITH_TAGS) {
                         return (
                             hasIntersection(
                                 explore.tags || [],
                                 tableSelection.value || [],
-                            ) || explore.type === ExploreType.VIRTUAL
+                            ) ||
+                            explore.type === ExploreType.VIRTUAL ||
+                            (includePreAggregateDebugExplores &&
+                                explore.type === ExploreType.PRE_AGGREGATE)
                         );
                     }
                     if (tableSelection.type === TableSelectionType.WITH_NAMES) {
                         return (
                             (tableSelection.value || []).includes(
                                 explore.name,
-                            ) || explore.type === ExploreType.VIRTUAL
+                            ) ||
+                            explore.type === ExploreType.VIRTUAL ||
+                            (includePreAggregateDebugExplores &&
+                                explore.type === ExploreType.PRE_AGGREGATE)
                         );
                     }
                     return true;
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -2114,10 +2114,10 @@ export class AsyncQueryService extends ProjectService {
 
         let preAggregateMetadata: Pick<CacheMetadata, 'preAggregate'> | null =
             null;
-        if (
-            process.env.ENABLE_PRE_AGGREGATE_DRY_RUN === 'true' &&
-            (explore.preAggregates || []).length > 0
-        ) {
+        const preAggregatesEnabled =
+            process.env.ENABLE_PRE_AGGREGATES === 'true' ||
+            process.env.ENABLE_PRE_AGGREGATE_DRY_RUN === 'true';
+        if (preAggregatesEnabled && (explore.preAggregates || []).length > 0) {
             const matchResult = findMatch(metricQuery, explore);
 
             preAggregateMetadata = {
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -4897,28 +4897,40 @@ export class ProjectService extends BaseService {
             projectUuid,
             includeErrors,
         );
+        const includePreAggregateDebugExplores =
+            process.env.ENABLE_PRE_AGGREGATE_DEBUG_VIEW === 'true' ||
+            process.env.ENABLE_PRE_AGGREGATE_INTERNAL_VISIBILITY === 'true';
+        const visibleExploreSummaries = includePreAggregateDebugExplores
+            ? allExploreSummaries
+            : allExploreSummaries.filter(
+                  (explore) => explore.type !== ExploreType.PRE_AGGREGATE,
+              );
 
         if (filtered) {
             const {
                 tableSelection: { type, value },
             } = await this.getTablesConfiguration(account, projectUuid);
             if (type === TableSelectionType.WITH_TAGS) {
-                return allExploreSummaries.filter(
+                return visibleExploreSummaries.filter(
                     (explore) =>
                         hasIntersection(explore.tags || [], value || []) ||
-                        explore.type === ExploreType.VIRTUAL, // Custom explores/Virtual views are included by default
+                        explore.type === ExploreType.VIRTUAL || // Custom explores/Virtual views are included by default
+                        (includePreAggregateDebugExplores &&
+                            explore.type === ExploreType.PRE_AGGREGATE),
                 );
             }
             if (type === TableSelectionType.WITH_NAMES) {
-                return allExploreSummaries.filter(
+                return visibleExploreSummaries.filter(
                     (explore) =>
                         (value || []).includes(explore.name) ||
-                        explore.type === ExploreType.VIRTUAL, // Custom explores/Virtual views are included by default
+                        explore.type === ExploreType.VIRTUAL || // Custom explores/Virtual views are included by default
+                        (includePreAggregateDebugExplores &&
+                            explore.type === ExploreType.PRE_AGGREGATE),
                 );
             }
         }
 
-        return allExploreSummaries;
+        return visibleExploreSummaries;
     }
 
     async getExplore(
@@ -6574,7 +6586,9 @@ export class ProjectService extends BaseService {
         const allExplores = Object.values(cachedExplores);
 
         const validExplores = allExplores?.filter(
-            (explore) => explore.type !== ExploreType.VIRTUAL,
+            (explore) =>
+                explore.type !== ExploreType.VIRTUAL &&
+                explore.type !== ExploreType.PRE_AGGREGATE,
         );
 
         if (!validExplores) {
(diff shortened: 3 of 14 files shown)


packages/backend/src/models/SearchModel/index.ts
```diff
@@ -1122,21 +1122,36 @@ export class SearchModel {
             .limit(1);
 
         if (explores.length > 0 && explores[0].explores) {
+            const includePreAggregateDebugExplores =
+                process.env.ENABLE_PRE_AGGREGATE_DEBUG_VIEW === 'true' ||
+                process.env.ENABLE_PRE_AGGREGATE_INTERNAL_VISIBILITY === 'true';
```
notgiorgi, line 1254: Info: this has been reworked in the upstack we just have 1 feature flag

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -2117,10 +2117,10 @@ export class AsyncQueryService extends ProjectService {
             CacheMetadata,
             'preAggregateHit' | 'preAggregateName' | 'preAggregateMissReason'
         > | null = null;
-        if (
-            process.env.ENABLE_PRE_AGGREGATE_DRY_RUN === 'true' &&
-            (explore.preAggregates || []).length > 0
-        ) {
+        const preAggregatesEnabled =
+            process.env.ENABLE_PRE_AGGREGATES === 'true' ||
+            process.env.ENABLE_PRE_AGGREGATE_DRY_RUN === 'true';
```
notgiorgi, line 2119: same here

packages/backend/src/services/ProjectService/ProjectService.test.ts
```diff
@@ -638,6 +639,40 @@ describe('ProjectService', () => {
...
+                expect(
+                    result.find(
+                        (explore) => explore.name === preAggregateExplore.name,
+                    ),
+                ).toBeDefined();
+            } finally {
+                process.env.ENABLE_PRE_AGGREGATE_DEBUG_VIEW = previousDebugFlag;
+            }
+        });
+
```
notgiorgi: redundant test

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -4774,28 +4774,40 @@ export class ProjectService extends BaseService {
             projectUuid,
             includeErrors,
         );
+        const includePreAggregateDebugExplores =
+            process.env.ENABLE_PRE_AGGREGATE_DEBUG_VIEW === 'true' ||
+            process.env.ENABLE_PRE_AGGREGATE_INTERNAL_VISIBILITY === 'true';
```
notgiorgi, line 4902: Info: this is simplifed in upstack into a single env var

packages/common/src/compiler/translator.ts
```diff
@@ -1195,7 +1197,60 @@ export const convertExplores = async (
             }
         });
 
-        return [...acc, ...newExplores];
+        const preAggregatesEnabled =
+            process.env.ENABLE_PRE_AGGREGATES === 'true' ||
+            process.env.ENABLE_PRE_AGGREGATE_VIRTUAL_EXPLORES === 'true';
```
notgiorgi: Same here

packages/common/src/compiler/translator.ts
```diff
@@ -1195,7 +1197,60 @@ export const convertExplores = async (
...
+                    warnings: [
+                        ...(compiled.warnings || []),
+                        ...warningMessages,
+                    ],
+                },
+                ...generatedPreAggregateExplores,
+            ];
+        });
+
+        return [...acc, ...exploresWithGeneratedPreAggregates];
```
notgiorgi: would it be possible to extract this, move under `preAggregates/` and concat with returned result

packages/common/src/preAggregates/buildPreAggregateExplore.ts
```diff
@@ -0,0 +1,383 @@
...
+import {
+    getJoinedDimensionColumnName,
+    getMetricColumnName,
+    getPreAggregateExploreName,
+    getTimeDimensionColumnName,
+} from './naming';
+
+export const PRE_AGGREGATE_SQL_TABLE_PLACEHOLDER = '{{PRE_AGG_SQL_TABLE}}';
+
+const GRANULARITY_ORDER: TimeFrames[] = [
```
notgiorgi: this already exists in `timeFrames.ts`

packages/common/src/preAggregates/buildPreAggregateExplore.ts
notgiorgi, line 1: there's a good change some of the utils defined in this file already exist, in which case we want to reuse them  
  
check:  
  
packages/common/src/utils/fields.ts  
packages/common/src/utils/item.ts

packages/common/src/preAggregates/buildPreAggregateExplore.ts
```diff
@@ -0,0 +1,383 @@
...
+            });
+            return acc;
+        },
+        new Map<FieldId, CompiledMetric>(),
+    );
+
+const isSupportedMetricType = (metricType: MetricType): boolean =>
+    [MetricType.SUM, MetricType.COUNT, MetricType.MIN, MetricType.MAX].includes(
+        metricType,
+    );
```
notgiorgi, line 80: add a todo comment for avg

joaoviana:
Do you want to update `findExploresFromCache` method in `ProjectModel` to filter out pre aggregate explores by default or all good?

packages/common/src/preAggregates/buildPreAggregateExplore.ts
```diff
@@ -0,0 +1,367 @@
...
+        // TODO: Support AVERAGE by materializing SUM + COUNT and deriving AVG at query time.
+        if (
+            metric.type === MetricType.AVERAGE ||
+            metric.type === MetricType.COUNT_DISTINCT ||
+            metric.type === MetricType.MEDIAN ||
+            metric.type === MetricType.PERCENTILE ||
+            metric.type === MetricType.NUMBER ||
+            metric.type === MetricType.SUM_DISTINCT ||
+            PostCalculationMetricTypes.includes(metric.type) ||
+            !isSupportedMetricType(metric.type)
```
joaoviana, line 254: Since `isSupportedMetricType` only returns `true` for `SUM | COUNT | MIN | MAX`, this final `!isSupportedMetricType`  makes all preceding conditions _dead_ _code_ do you agree?

packages/common/src/preAggregates/buildPreAggregateExplore.ts
```diff
@@ -0,0 +1,367 @@
...
+    return {
+        name: tableName,
+        label: tableName,
+        database: '',
+        schema: '',
+        sqlTable,
+        dimensions: {},
+        metrics: {},
+        lineageGraph: {},
+    };
```
joaoviana: `includedTableNames` is built from tables that all come from validated dimensions/metrics (which throw if their table isn't found). So do you think it makes sense to check if `sourceTable` exists?

packages/common/src/preAggregates/buildPreAggregateExplore.ts
```diff
@@ -0,0 +1,367 @@
...
+        tables[dimension.table].dimensions[dimension.name] = {
+            ...dimension,
+            sql: compiledSql,
+            compiledSql,
+            tablesReferences: [sourceExplore.baseTable],
+        };
+    });
+
+    includedMetrics.forEach(({ fieldId, metric }) => {
+        const metricColumnName = getMetricColumnName(fieldId);
```
joaoviana, line 320: hmm this function just returns the same argument no? I think this is unnecessary

packages/common/src/preAggregates/generatePreAggregateExplores.ts
```diff
@@ -0,0 +1,68 @@
...
+            return [compiled, ...generatedPreAggregateExplores];
+        }
+
+        return [
+            {
+                ...compiled,
+                warnings: [
+                    ...(compiled.warnings || []),
+                    ...generationErrors.map((message) => ({
+                        type: InlineErrorType.FIELD_ERROR,
```
joaoviana, line 60:
```
METADATA_PARSE_ERROR
```

instead? Or a completely new `InlineErrorType` for better tracking/ux

packages/common/src/preAggregates/buildPreAggregateExplore.ts
```diff
@@ -0,0 +1,367 @@
...
+    getPreAggregateExploreName,
+    getTimeDimensionColumnName,
+} from './naming';
+import {
+    getDimensionBaseName,
+    getDimensionReferences,
+    getMetricReferences,
+} from './references';
+
+export const PRE_AGGREGATE_SQL_TABLE_PLACEHOLDER = '{{PRE_AGG_SQL_TABLE}}';
```
joaoviana: not used yet, would this be needed in the upstack?

notgiorgi: removed

packages/common/src/preAggregates/buildPreAggregateExplore.ts
notgiorgi: https://app.graphite.com/github/pr/lightdash/lightdash/20534/feat-pre-aggregate-virtual-explores#lines-packages/common/src/preAggregates/buildPreAggregateExplore.ts-R295-R295

packages/common/src/preAggregates/buildPreAggregateExplore.ts
notgiorgi: I left it as a naming-boundary helper so we can change metric column naming in one place later (without touching builder logic)

packages/common/src/preAggregates/generatePreAggregateExplores.ts
notgiorgi: I kept FIELD_ERROR intentionally because this path is a non-fatal generation warning on an otherwise valid explore; METADATA_PARSE_ERROR is used for fatal metadata parse failures afaik

joaoviana: gotcha! makes sense ty

PR #20537, opened 2026-02-23
fix: preserve SQL LIMIT clause when no limit parameter provided

owlas:
## Summary
When wrapping user SQL as a subquery, the `SqlQueryBuilder` was stripping the original LIMIT clause even when no limit parameter was passed. This caused queries like `SELECT * FROM table LIMIT 1` to return unlimited rows when executed through the async SQL API.

**Fix:**
- When `limit` is `undefined`: preserves original LIMIT/OFFSET from user's SQL
- When `limit` is provided: existing behavior unchanged (uses `Math.min` of original and provided limit)

## Test plan
- [x] Unit tests added for preserving LIMIT when no limit parameter provided
- [x] Unit tests added for preserving LIMIT with OFFSET
- [ ] Test via SQL Runner UI: run `SELECT * FROM table LIMIT 1` and verify 1 row returned
- [ ] Test via API: POST to `/api/v2/projects/{uuid}/query/sql` without `limit` param

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.ts
@@ -170,10 +170,10 @@ export class SqlQueryBuilder {
             }
             if (limitOffset) {
                 // No provided limit but original SQL had limit/offset - preserve it
-                const offsetString = limitOffset.offset
-                    ? ` OFFSET ${limitOffset.offset}`
-                    : '';
-                return `LIMIT ${limitOffset.limit}${offsetString}`;
+                return SqlQueryBuilder.buildLimitOffsetSql(
+                    limitOffset.limit,
+                    limitOffset,
+                );
             }
             // No provided limit and no original limit
             return undefined;


packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.ts
```diff
@@ -156,16 +156,33 @@ export class SqlQueryBuilder {
...
+                    this.limit,
+                    limitOffset,
+                );
+            }
+            if (limitOffset) {
+                // No provided limit but original SQL had limit/offset - preserve it
+                const offsetString = limitOffset.offset
+                    ? ` OFFSET ${limitOffset.offset}`
+                    : '';
+                return `LIMIT ${limitOffset.limit}${offsetString}`;
```
almeidabbm: This could be been done by `SqlQueryBuilder.buildLimitOffsetSql` so we don't repeat logic, `SqlQueryBuilder.buildLimitOffsetSql(limitOffset.limit, limitOffset)` 🤔

owlas: done!

Issue #19705, opened 2026-01-23
Editing filter values causes UI to jump to last filter when multiple filter groups are present


torimwhaley:
### Description

When a chart has multiple filter groups, editing the value of a filter near the top of the filter panel causes the UI to jump to the bottom (last filter) after each character is added or removed.

Although the filter value is applied correctly, the jumpy behavior makes editing filters frustrating and error-prone. This issue reproduces consistently and does not appear to be configuration-specific.

The behavior was observed on a self-hosted instance (v0.2370.2) and is also reproducible on Lightdash Cloud.

see loom [here](https://www.loom.com/share/841296ecfe1f49feacf97c2038791682)
see example chart [here](https://analytics.lightdash.cloud/projects/2cc7b1db-c5b5-40c2-b35b-9dfc88d5494f/saved/1cbd7d21-369e-4640-b247-d99910431fdd/) - edit and go to filter pane

### Steps to Reproduce the Bug or Issue

Steps to Reproduce the Bug or Issue
1. Create or open a chart with several groups of filters.
2. Ensure multiple filters are visible in the filter panel.
3. Click into a filter input near the top of the panel.
4. Start typing or deleting characters.

Observe that after each character change, the UI scrolls/jumps to the last filter input.

### version

0.2370.2

### Cloud or self-hosting

cloud

bikalsiwakoti:
Hey @torimwhaley 
Can you please look into this :)

PR #20564, opened 2026-02-24
fix: stop filter autoFocus from stealing focus to last filter

charliedowler:
## Summary
- Replaced hardcoded `autoFocus={true}` on every filter input with a conditional system
- New rules get autofocused via `autoFocusRuleId` state threaded through the component tree
- Existing rules retain focus after value changes via `onFocusCapture` tracking + post-render restoration
- Added `data-rule-id` attribute to `FilterRuleForm` for DOM-based focus targeting

## Test plan
- [x] Playwright tests validate all 3 scenarios (`filter-autofocus-repro.spec.ts`):
  - Changing an existing filter value keeps focus on the edited filter (not the last one)
  - Adding a new filter via "Add filter" autofocuses the new rule's input
  - Adding a group rule via "Add group rule" autofocuses the new rule's input
- [ ] Manual: verify filters still work normally in chart edit mode
- [ ] Manual: verify dashboard filter inputs still work

Closes: https://github.com/lightdash/lightdash/issues/19705

### DIFF
--- a/packages/frontend/src/components/common/Filters/index.tsx
+++ b/packages/frontend/src/components/common/Filters/index.tsx
@@ -35,11 +35,11 @@ import {
 } from 'react';
 import { useToggle } from 'react-use';
 import { v4 as uuidv4 } from 'uuid';
+import { useIsFilterAutofocusEnabled } from '../../../hooks/useIsFilterAutofocusEnabled';
 import {
     type FieldsWithSuggestions,
     type FieldWithSuggestions,
 } from '../../Explorer/FiltersCard/useFieldsWithSuggestions';
-import { useIsFilterAutofocusEnabled } from '../../hooks/useIsFilterAutofocusEnabled';
 import FieldSelect from '../FieldSelect';
 import MantineIcon from '../MantineIcon';
 import { FILTER_SELECT_LIMIT } from './constants';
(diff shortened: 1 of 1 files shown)


charliedowler:
## 🤖 Implementation notes from Claude Code

This fix was developed iteratively with a Playwright test suite (`filter-autofocus-repro.spec.ts`) covering 3 scenarios that validated each approach. Here's what was tried:

### What worked

**Two-layer focus management:**

1. **`autoFocusRuleId` prop threading (for new rules)** — When "Add filter" or "Add group rule" is clicked, the new rule's ID is captured and passed down from `FiltersForm` → `FilterGroupForm` → `FilterRuleForm`. Only the matching rule gets `autoFocus={true}` on mount. A `useEffect` clears the ID after one render cycle. For group rules, an `onAutoFocusRule` callback bubbles the ID up from nested `FilterGroupForm` to `FiltersForm`.

2. **`onFocusCapture` + post-render restore (for existing rules)** — `onFocusCapture` on the form container tracks which rule was last focused via `data-rule-id`. After every render, if focus has fallen to `document.body`, a `useEffect` restores it to the input inside that rule.

### What didn't work (and why)

1. **Local `autoFocusRuleId` state inside `FilterGroupForm`** — Failed for "Add group rule" because `getFiltersFromGroup` in `@lightdash/common` generates **new UUIDs** for every group container on each call. This causes nested `FilterGroupForm` components to get new React `key` props → unmount/remount → local state destroyed before consumed.

2. **`requestAnimationFrame` refocus inside `BooleanFilterInputs`** — Failed because the same UUID regeneration issue means the component is remounted (not re-rendered) after a value change, so the ref and callback are destroyed.

3. **`useState` + `useEffect` refocus inside `BooleanFilterInputs`** — Same remount issue — state is lost.

4. **Wrapper `div` ref with `querySelector('input')` inside `BooleanFilterInputs`** — Same remount issue.

The key insight: because `getFiltersFromGroup` recreates group container IDs, the entire component tree below `FiltersForm` is destroyed and recreated on every filter change. Any approach that stores state or refs inside the remounting components will fail. The fix had to live at the `FiltersForm` level, above the reconstruction.

All 3 Playwright test scenarios pass:
- ✅ Changing first filter value → focus stays on rule 0
- ✅ Adding a new filter → focus lands on the new rule
- ✅ Adding a group rule → focus lands on the new rule (at the correct index)

PR #20580, opened 2026-02-24
feat: add CSV import to multi-number filter inputs

almeidabbm:
Closes: https://linear.app/lightdash/issue/PROD-3578/i-want-to-use-csv-filter-values-for-number-type-dimensions

## Summary

- Create `FilterMultiNumberInput` component that wraps `TagInput` with `ManageFilterValuesModal`, providing CSV import for number filters
- Replace the inline `TagInput` in `DefaultFilterInputs` for NUMBER EQUALS/NOT_EQUALS operators with the new component
- Manage button (list icon) appears on hover in the right section, matching the string filter UX pattern
- CSV import validates values with a number regex, rejecting non-numeric entries
- Value truncation matching `FilterStringAutoComplete` behavior:
  - Summary mode (>500 values): shows "N values selected"
  - Inline truncation (>50 values): shows first 50 pills + "+N more" clickable pill
  - Max-height scroll for the input container (350px)

https://github.com/user-attachments/assets/2be23efb-af61-4641-8f4e-c2ec032978d9

- Remove redundant "Manage values" button from string filter summary mode (be6b79852d)

**Before**
CleanShot 2026-02-24 at 18 02 46

**After**
CleanShot 2026-02-24 at 18 02 28

## Test plan

- [ ] Open a number filter with EQUALS or NOT_EQUALS operator
- [ ] Verify the manage/edit icon appears on hover in the right section
- [ ] Click it to open the ManageFilterValuesModal
- [ ] Import a CSV file with mixed values — only valid numbers should be imported
- [ ] Verify paste still works (paste "1,2,3" → three tags)
- [ ] Verify the clear button still works
- [ ] Import >500 values — verify summary mode shows "N values selected"
- [ ] Import 50-500 values — verify truncation with "+N more" pill
- [ ] Verify string filters still work unchanged (including summary mode without "Manage values" button)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.test.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.test.tsx
@@ -195,10 +195,8 @@ describe('FilterStringAutoComplete', () => {
                 screen.getByDisplayValue('600 values selected'),
             ).toBeInTheDocument();
 
-            // Should show "Manage values" button
-            expect(
-                screen.getByRole('button', { name: 'Manage values' }),
-            ).toBeInTheDocument();
+            // Should not render individual value pills
+            expect(screen.queryByText('value_0')).not.toBeInTheDocument();
         });
     });
 
@@ -229,7 +227,7 @@ describe('FilterStringAutoComplete', () => {
             });
         });
 
-        it('opens modal from summary mode manage values button', async () => {
+        it('opens modal from summary mode by clicking the input', async () => {
             const user = userEvent.setup({ pointerEventsCheck: 0 });
             // Use > 500 values to trigger summary mode
             const values = createValues(600);
@@ -245,11 +243,11 @@ describe('FilterStringAutoComplete', () => {
                 />,
             );
 
-            // In summary mode, there's a visible "Manage values" button
-            const manageButton = screen.getByRole('button', {
-                name: 'Manage values',
-            });
-            await user.click(manageButton);
+            // In summary mode, clicking the input opens the modal
+            const summaryInput = screen.getByDisplayValue(
+                '600 values selected',
+            );
+            await user.click(summaryInput);
 
             // Modal should open
             await waitFor(() => {


packages/frontend/src/components/common/Filters/FilterInputs/FilterMultiNumberInput.tsx
```diff
@@ -0,0 +1,75 @@
+import { ActionIcon, Tooltip } from '@mantine/core';
```
joaoviana: if this is new, then it should be in mantine-8

almeidabbm: The component now uses `@mantine-8/core` and `@mantine-8/hooks` (pushed in v2). 

One thing to note: the `Tooltip` wrapping the manage button uses v8 styling, while the same tooltip in `FilterStringAutoComplete` uses v6. This causes a slight font size/weight difference between the two. What's your preference: keep v8 styling (will be consistent once `FilterStringAutoComplete` is migrated), or temporarily use the v6 Tooltip to match the current look?

**V8**
CleanShot 2026-02-24 at 17 17 59

**V6**
CleanShot 2026-02-24 at 17 18 03

joaoviana: I would say to keep v8 styling on `FilterMultiNumberInput` because I'm doing a batch of migrations just now ! :D

joaoviana:
I believe it needs truncation and a max height, similar to how the string filter autocomplete does 
image

almeidabbm:
Thanks for the feedback! Updated in v3 — the multi-number filter now matches the string filter behavior:

- **Summary mode** (>500 values): shows "N values selected" read-only input
- **Inline truncation** (>50 values): shows first 50 pills + "+N more" clickable pill that opens the manage modal
- **Max-height scroll** (≤50 values): input container capped at 350px with overflow scroll

Reused the same truncation utilities from `FilterStringAutoComplete.utils`.

@joaoviana here are the screenshots:
**truncate**
CleanShot 2026-02-24 at 17 55 50
**summary**
CleanShot 2026-02-24 at 17 56 40

PR #20611, opened 2026-02-25
chore: updated nested permission text

magnew:
### Description:

Text and terminology update around nested space permissions. The goal is to make terminology clearer -- if this doesn't make sense we can iterate. With this there are two main permission categories
- Inherited access - the space inherits permissions from its parents and project
- Restricted access - the space defines it's own permissions, which it's children inherit

I chose 'inherited permissions' here since it makes it clear what to expect. We tried some others (Parent, Public, Project) which were all trying to convey something specific that needed to be explained, but wasn't accurate. 'Inherited' is clear about what to expect and works at multiple levels in the tree. 

'Restricted' is also more clear than some of the alternatives we tried (Custom, private). It makes it clear that access isn't open while leaving space for it to be private or a little less than private. 

There was some question about whether *top-level* spaces should have their own term, like 'Project access', but after trying it, I found it more confusing to try to explain the top-level space access differently. Partly because we don't have different visuals or iconography to show that you are at a top-level space. The distinction only feels meaningful to those of us who are familiar with the system. I think other users will understand 'Inherited access' better, even at the root.

### DIFF
--- a/packages/frontend/src/components/common/ResourceView/ResourceAccessInfo.tsx
+++ b/packages/frontend/src/components/common/ResourceView/ResourceAccessInfo.tsx
@@ -10,15 +10,15 @@ import { getResourceAccessLabel, getResourceAccessType } from './utils';
 const ResourceAccessInfoData = {
     [ResourceAccess.Private]: {
         Icon: IconLock,
-        status: 'Private',
+        status: 'Restricted access',
     },
     [ResourceAccess.Public]: {
         Icon: IconUsers,
-        status: 'Public',
+        status: 'Inherited access',
     },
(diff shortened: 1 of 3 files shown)


magnew:
* **#20611** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20618, opened 2026-02-25
feat: Add semi-additive metrics support to Lightdash

hamzahc1:
## Summary

This PR introduces support for **semi-additive metrics** in Lightdash — metrics that can be summed across most dimensions but require special handling on the time dimension (taking the first or last value instead of summing). Common examples include account balances, inventory levels, and MRR.

Semi-additive metrics are configured via YAML metadata and generate specialized SQL using window functions (ROW_NUMBER) or subquery filters to select the appropriate time period's value before aggregation.

## Key Changes

- **Type definitions** (`packages/common/src/types/field.ts`):
  - Added `SemiAdditiveConfig` type with `timeDimension`, `windowChoice` (min/max), and optional `windowGroupings`
  - Extended `Metric` interface with optional `semiAdditive` property
  - Added `CompiledMetric.compiledSemiAdditive` for compiled dimension references
  - Added `isSemiAdditiveMetric()` helper predicate

- **dbt YAML support** (`packages/common/src/types/dbt.ts`):
  - Added `semi_additive` field to `DbtColumnLightdashMetric` type
  - Wired conversion logic in `convertModelMetric()` to map YAML config to `Metric.semiAdditive`

- **JSON schema validation** (`packages/common/src/dbt/schemas/lightdashMetadata.json`):
  - Added `semi_additive` schema definition with required fields (`time_dimension`, `window_choice`) and optional `window_groupings`

- **Metric query types** (`packages/common/src/types/metricQuery.ts`):
  - Added `semiAdditive` property to `AdditionalMetric` interface for UI-created custom metrics

- **SQL compilation** (`packages/common/src/compiler/exploreCompiler.ts`):
  - Added compilation of semi-additive dimension references in `compileMetricSql()`
  - Resolves time dimension and window grouping SQL expressions for use in query building

- **Query building** (`packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts`):
  - Added `buildSemiAdditiveCtes()` method to generate semi-additive CTEs
  - Implements two patterns:
    - **With window_groupings**: ROW_NUMBER PARTITION BY approach for multi-entity tables
    - **Without window_groupings**: Subquery MIN/MAX filter for single-entity tables
  - Integrated into `compileQuery()` pipeline to generate CTEs before final SELECT
  - Updated `getMetricsSQL()` to skip semi-additive metrics from regular aggregation (handled via CTE instead)

## Implementation Details

- **CTE-based approach**: Follows the same pattern as existing `sum_distinct` metrics, using CTEs to isolate the semi-additive logic and ensure correct row filtering before aggregation
- **Orthogonal to metric type**: Semi-additivity is a configuration property, not a metric type, allowing any aggregate metric (SUM, COUNT, etc.) to be semi-additive
- **Dimension compilation**: Time dimension and window grouping references are compiled during explore compilation to resolve table aliases and generate correct SQL
- **Multiple metrics support**: Query builder handles multiple semi-additive metrics in a single query, each with its own CTE

## Testing Considerations

The implementation includes validation for:
- Time dimension must be a valid DATE/TIMESTAMP dimension
- Window groupings must reference valid dimensions
- Semi-additive cannot be combined with `distinct_keys` (sum_distinct)
- Proper handling of edge cases (no dimensions, time dimension in GROUP BY, metric filters)

https://claude.ai/code/session_01Bxr6cgNi6mnJ9nEa11B5x6


PLAN-semi-additive-metrics.md
```diff
@@ -0,0 +1,663 @@
...
+### Example queries and results
+
+**Scenario**: `daily_account_balances` table with columns `date_day`, `account_id`, `region`, `balance`
+
+**Query 1**: Metric `current_balance` grouped by `region`
+```
+| region | current_balance |
+|--------|-----------------|
+| East   | 150             |  ← Sum of latest balances for East accounts
+| West   | 250             |  ← Sum of latest balances for West accounts
```
jesshitchcock, line 161: If the time dim is not included do we return the value corresponding to the first date in the partition or the last depending on config. 

Assumption: if not all dates are present we would take the max or min (depending on config) date available within the partition. E.g for East we might have data until 20 Jan and West until 29 Jan. The `current_balance` metric based on the max date would reflect the data for these max dates per region.

jesshitchcock:
Hey @hamzahc1,  

I had a thought on the semi-additive plan re: `Case A` vs `Case B`.

If we assume the end user is responsible for data completeness (i.e. every entity has a row on the first (or last) date of each time bucket), then Case A with window_groupings doesn't actually need the ROW_NUMBER approach. 

We can just use the same global MAX date filter as Case B:
``` sql
-- instead of row_number() to work out max per entity, we use the global max per time partition
WITH sa_max_dates AS (                                                                                                                                                      
    SELECT                                                                                                                                                                    
      DATE_TRUNC('month', date_day) AS date_month,                                                                                                                            
      MAX(date_day) AS max_date                                                                                                                                               
    FROM daily_account_balances                                                                                                                                               
    WHERE date_day IS NOT NULL                                                                                                                                                
    GROUP BY DATE_TRUNC('month', date_day)
  ),
  sa_current_balance AS (
    SELECT
      m.date_month,
      t.account_id,
      t.region,
      SUM(t.balance) AS current_balance
    FROM daily_account_balances t
    INNER JOIN sa_max_dates m ON t.date_day = m.max_date
    WHERE t.date_day IS NOT NULL
    GROUP BY m.date_month, t.account_id, t.region
  )
  SELECT * FROM sa_current_balance
``` 

This works because if no data is missing, every entity's latest row IS the global max date — so the results are identical to ROW_NUMBER with rn = 1. I think this was the original design and I perhaps overcomplicated things by trying to handle edge cases. 

### Suggestion 
Let's simplify Case A to also use the global MAX/MIN subquery approach (same as Case B), with window_groupings still declared in the YAML for documentation/validation but not changing the SQL generation strategy. 
Pros: 
- One code path instead of two
- Simpler generated SQL (no window functions)
- Better query performance on large tables

Cons: 
- If data IS missing, entities are silently dropped, but that's the user's responsibility to ensure data completeness. And this is something I would expect in the Bi layer

What do you think? I have pushed the suggested changes so you can see them, but we can revert if we decide against them.

PR #20639, opened 2026-02-26
feat: introduce parameter labels

### LINKED ISSUE #15959: Allow to define a label for each parameter option
### Description

At the moment, we can define parameter options values. 
We want to be able to define a label for each option and have a data structure that allows for future configuration. For example, disabled option based on user attribute. 

Example:
```
parameters:
  subscription_status:
    label: "Subscription Status"
    description: "Filter subscriptions by their current status"
    default: "all"
    options:
      - value: "all"
        label: "All status"
      - value: "pending"
        label: "Pending"
      - value: "success"
        label: "Success"
```

Have the label be different from the value. 
The label can contain uppercase letters, white spaces or special charts but we might not want in the value.

dtaniwaki:
Thanks for maintaining this project! I'd love to get a review on this feature addition that extends Lightdash parameters with label support.

Closes: #15959
 
### Description:

This PR introduces **parameter labels** — the ability to show human-readable display text for parameter option values, separate from the underlying SQL-substituted value.

There are two ways to use this:

**1. `label_dimension` in `options_from_dimension`**

When fetching options from a dbt dimension, an optional `label_dimension` can now be specified to display a different field as the label while the `dimension` field is used as the actual value:

```yaml
parameters:
  employer:
    label: Employer
    options_from_dimension:
      model: employer_lookup
      dimension: employer_id          # value used in SQL substitution
      label_dimension: employer_name  # display label shown in the UI
```

When `label_dimension` is set, the API fetches both fields, searches/sorts by the label field, and returns `{ value, label }` pairs to the frontend.

**2. Labeled static `options`**

Static options can now be specified as `{ label, value }` objects instead of plain strings/numbers:

```yaml
parameters:
  status:
    label: Status
    options:
      - label: Active
        value: active_id_1
      - label: Inactive
        value: inactive_id_2
```

**Changes summary:**
- `common`: Added `LightdashParameterOption` type and `isLightdashParameterOption` type guard; extended `LightdashProjectParameter` to accept labeled options; updated JSON schema accordingly.
- `backend`: `_getFieldValuesMetricQuery` resolves an optional label dimension, queries both value and label fields, and returns `{ value, label }` pairs when a label field is present.
- `frontend`: `useFieldValues` accumulates a `labelMap` (value → label) from labeled API responses; `ParameterInput` uses this map to display labels in the dropdown while submitting the underlying value.
- Tests added for both new config shapes.

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -7310,6 +7310,25 @@ const models: TsoaRoute.Models = {
         },
     },
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
+    LightdashParameterOption: {
+        dataType: 'refAlias',
+        type: {
+            dataType: 'nestedObjectLiteral',
+            nestedProperties: {
+                value: {
+                    dataType: 'union',
+                    subSchemas: [
+                        { dataType: 'string' },
+                        { dataType: 'double' },
+                    ],
+                    required: true,
+                },
+                label: { dataType: 'string', required: true },
+            },
+            validators: {},
+        },
+    },
+    // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
     LightdashProjectParameter: {
         dataType: 'refAlias',
         type: {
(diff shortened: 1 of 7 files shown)


ZeRego:
Thanks for this contribution, @dtaniwaki! 🙏

I've rebased this PR onto the latest `main` and resolved the conflicts directly on your branch — your commit and `Signed-off-by` are preserved. I also scoped it down to **labeled static parameter options**: `options:` entries given as `{ label, value }` pairs. That part is clean and ready.

I've **removed the `label_dimension` changes** (the label field for `options_from_dimension`) from this PR. Since this PR was first opened, `main` gained a `FilterAutocompleteValue { value, label }` model for dimension autocomplete, so the cleanest, backwards-compatible way to do dimension-sourced labels is now different from the original approach here. We've opened #24931 to implement that separately, building on that model.

What this PR contains now:
- `LightdashParameterOption` type + JSON-schema support for `options: [{ label, value }]`
- The parameter input renders the labels (showing the label, submitting the value)
- Config-load tests for labeled options
- Regenerated OpenAPI types

Happy to adjust if anything looks off — and thanks again for kicking this off!

dtaniwaki:
@ZeRego Thank you for your review. I totally agree your modification and will make a PR for #24931 separately.

PR #20656, opened 2026-02-26
fix: throw AuthorizationError when pending user tries to sign up via OpenID without invite code

dtaniwaki:
Closes: 

### Description:

When a user has been invited to Lightdash (i.e. a pending user record exists for their email), they should complete sign-up via the invite link rather than through an OpenID/SSO flow. Previously, if a pending user attempted to sign up via OpenID without using their invite link, the system would proceed into the database transaction and either fail with an unhelpful error or create an inconsistent state.

This fix adds an early check before entering the database transaction: if the email associated with the OpenID login already belongs to a pending user, an `AuthorizationError` is thrown with a clear, actionable message directing the user to use the invite link sent to their email.

As part of this change, the `email` extraction logic (previously duplicated inside the transaction) is moved above it so it can be shared by both the pre-flight check and the transaction body.

### DIFF
--- a/packages/backend/src/models/UserModel.ts
+++ b/packages/backend/src/models/UserModel.ts
@@ -2,6 +2,7 @@ import { AbilityBuilder } from '@casl/ability';
 import {
     ActivateUser,
     AlreadyExistsError,
+    AuthorizationError,
     CreateUserArgs,
     CreateUserWithRole,
     ForbiddenError,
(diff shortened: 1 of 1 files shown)


ZeRego:
If this is a critical fix, please reach out to support.

dtaniwaki:
This is a very critical UX issue of the invitation flow. However, I'm using my own fork with this fix.

Issue #20670, opened 2026-02-26
CLI --defer does not use production schema for non-selected models pulled in via joins


almeidabbm:
#### Summary
When using `lightdash preview` (or `lightdash compile`) with `--defer --state`, models that are not in the user's original selection but get pulled in because of Lightdash joins incorrectly point to the dev schema instead of the production schema.

This defeats the purpose of `--defer`, which is to only build selected models in dev while referencing production for everything else.

#### Root cause
dbt's `--defer` flag only affects ref() resolution inside compiled SQL. The manifest node metadata (relation_name, schema, database) always reflects the current target (dev), even for non-compiled/deferred models.

Lightdash reads relation_name from the manifest to build the FROM clause of explores (in `convertTable()` via `translator.ts`). So when `maybeCompileModelsAndJoins()` detects missing joined models and re-runs `dbt ls` with them included, those joined models get their `relation_name` from the dev manifest — pointing to the wrong schema.

```
  Expected: FROM "postgres"."jaffle"."membership" (production)
  Actual: FROM "postgres"."jaffle_dev"."membership" (dev)
```

### Steps to Reproduce the Bug or Issue

1. **Add a second profile target to `examples/full-jaffle-shop-demo/profiles/profiles.yml`:**

```yaml
jaffle_dev:
  type: postgres
  host: "{{ env_var('PGHOST') }}"
  port: "{{ env_var('PGPORT') | as_number }}"
  user: "{{ env_var('PGUSER') }}"
  pass: "{{ env_var('PGPASSWORD') }}"
  dbname: "{{ env_var('PGDATABASE') }}"
  sslmode: disable
  schema: "jaffle_dev"
```

2. **Seed and run dbt for both targets:**

```bash
cd examples/full-jaffle-shop-demo/dbt

# Production
dbt seed --profiles-dir ../profiles/ --target jaffle
dbt run --profiles-dir ../profiles/ --target jaffle

# Dev
dbt seed --profiles-dir ../profiles/ --target jaffle_dev
dbt run --profiles-dir ../profiles/ --target jaffle_dev
```

3. **Add a `lightdash` tag to a few models that have joins (e.g. `customers.yml`, `orders.yml`, `payments.yml`):**

In each model's config block, add:

```yaml
config:
  tags: ["core", "ai", "lightdash"]
```

_Note: `customers` joins to `membership` and `plan` — these will be the deferred models._

4. **Generate production artifacts:**

```bash
[...]

PR #20672, opened 2026-02-26
fix: patch deferred model schemas when using --defer

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/20670

### Description:

Fixed schema resolution for deferred models when using `--defer` flag with joins. Previously, non-selected models that were pulled in via joins would incorrectly use the development target schema instead of the production schema from the state manifest.

The fix introduces patching logic that:

- Tracks originally selected model IDs separately from models added through join expansion
- Loads the state manifest when `--defer` and `--state` options are provided
- Updates the `relation_name`, `schema`, and `database` properties for non-selected models to match their production counterparts from the state manifest
- Provides debug logging for patched models and graceful error handling if the state manifest cannot be loaded

**Before**

**After**

**Note:** Reproduction steps in github issue

### DIFF
--- a/packages/cli/src/handlers/compile.ts
+++ b/packages/cli/src/handlers/compile.ts
@@ -4,10 +4,12 @@ import {
     convertLightdashModelsToDbtModels,
     DbtManifest,
     DbtManifestVersion,
+    DbtModelNode,
     Explore,
     ExploreError,
     getCompiledModels,
     getDbtManifestVersion,
+    getErrorMessage,
     getModelsFromManifest,
     getSchemaStructureFromDbtModels,
     isExploreError,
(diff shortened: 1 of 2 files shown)


packages/cli/src/handlers/compile.ts
```diff
@@ -171,6 +172,52 @@ export const compile = async (options: CompileHandlerOptions) => {
...
+                let patchedCount = 0;
+                compiledModels.forEach((model) => {
+                    if (!originallySelectedModelIds.includes(model.unique_id)) {
+                        const stateModel = stateModelMap.get(model.unique_id);
+                        if (stateModel) {
+                            /* eslint-disable no-param-reassign */
+                            model.relation_name = stateModel.relation_name;
+                            model.schema = stateModel.schema;
+                            model.database = stateModel.database;
+                            /* eslint-enable no-param-reassign */
```
almeidabbm: We shouldn't be reassigning params

PR #20693, opened 2026-02-26
feat: add case sensitivity configuration for string filters

ZeRego:
Closes: PROD-5660

### Description:

This PR adds case sensitivity control for string filters in Lightdash. The feature allows users to configure whether string filters should be case sensitive or case insensitive at both the explore level and individual dimension level.

**Key changes:**

- Added `caseSensitive` property to explore and dimension configurations in dbt YAML files
- Modified string filter SQL generation to apply `UPPER()` transformations when case insensitive filtering is enabled
- Implemented a hierarchy where field-level `caseSensitive` settings override explore-level settings, with a default of `true` (case sensitive)
- Updated filter operators `EQUALS`, `NOT_EQUALS`, `STARTS_WITH`, and `ENDS_WITH` to support case insensitive matching
- Added comprehensive test coverage for case sensitivity behavior across different filter operators
- Extended type definitions to include the new `caseSensitive` property in explore and dimension interfaces

When `caseSensitive` is set to `false`, string filters will convert both the database column values and filter values to uppercase before comparison, enabling case insensitive matching. The feature maintains backward compatibility by defaulting to case sensitive behavior when not explicitly configured.

### DIFF
--- a/packages/backend/src/ee/services/ai/utils/validators.ts
+++ b/packages/backend/src/ee/services/ai/utils/validators.ts
@@ -272,6 +272,8 @@ function validateFilterRule(
                 (string: string) => string.replaceAll('\\', '\\\\'),
                 WeekDay.SUNDAY,
                 SupportedDbtAdapter.BIGQUERY,
+                'UTC',
+                true, // Default to case sensitive for validation
             );
         }
     } catch (e) {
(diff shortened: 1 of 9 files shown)


ZeRego:
* **#20693** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/20693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20716, opened 2026-02-27
feat(cli): add generate-docs command to inject Lightdash metrics into dbt lineage

oli-rmsy:
## Summary

- Adds a new `lightdash generate-docs` CLI command that generates dbt docs and then hijacks dbts metric and semantic layer nodes to display Lightdash nodes.
- Injects semantic_model and metric nodes into `manifest.json` so Lightdash metrics (defined in dbt YAML under `meta.metrics`) appear as purple-colored nodes on the dbt docs lineage graph
- Supports both server mode (`dbt docs serve`) and static mode (`--static` for a self-contained HTML file)

## How it works

1. Compiles the dbt project into an isolated `target/lightdash_docs/` directory (via `dbt compile --target-path`), leaving the project's real `target/` untouched
2. Injects Lightdash metric nodes into the isolated `manifest.json`:
   - One **semantic_model** node per model that has Lightdash metrics (light purple)
   - One **metric** node per Lightdash metric definition (full Lightdash purple)
   - Correct `parent_map`/`child_map` lineage, including derived metric dependencies
   - Cross-model metric references (e.g. `${dbt_orders.count_distinct_order_id}`) are resolved so derived metrics connect to metrics from multiple semantic models
   - Each metric description includes the full Lightdash YAML definition as a code block, and `type_params.expr` is populated with the SQL expression
3. Runs `dbt docs generate --no-compile --target-path target/lightdash_docs/` which picks up the already-injected manifest (since `--no-compile` skips writing `manifest.json`), including for `--static` HTML embedding
4. Serves docs or prints the static file path. All dbt output is suppressed behind spinners for clean CLI output

All injected nodes are tagged with `_lightdash_injected` in their meta for idempotent re-runs.

## Usage

```bash
# Generate docs and serve locally
lightdash generate-docs

# Generate a self-contained static HTML file
lightdash generate-docs --static
```

## Test plan

- [x] `pnpm -F cli build` compiles without errors
- [x] `pnpm -F cli typecheck` passes
- [x] Injection produces correct semantic_model and metric nodes in manifest.json
- [x] Derived metrics have correct lineage (depend on input metrics, not semantic model)
- [x] Cross-model derived metrics connect to metrics from multiple semantic models
- [x] Idempotent: running twice produces identical output
- [x] `--static` flag produces static_index.html with injected manifest embedded
- [x] `--no-serve` flag skips dbt docs serve
- [x] Node colors render correctly in dbt docs lineage graph
- [x] Project's real `target/` directory is never modified

Below is an example of this running locally against Thyme to Shine

image

### DIFF
--- a/packages/cli/src/handlers/generateDocs.ts
+++ b/packages/cli/src/handlers/generateDocs.ts
@@ -17,8 +17,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { LightdashAnalytics } from '../analytics/analytics';
 import GlobalState from '../globalState';
 import {
-    getManifestPath,
-    getStaticIndexPath,
+    getTargetFilePath,
     injectLightdashLineage,
     patchStaticIndex,
 } from './injectLightdashLineage';
@@ -34,6 +33,31 @@ type GenerateDocsHandlerOptions = {
     serve: boolean;
     port: number;
 };
+
+/** Build the common dbt CLI flags shared across generate and serve steps */
+function buildCommonDbtArgs(
+    absoluteProjectDir: string,
+    options: Pick<
+        GenerateDocsHandlerOptions,
+        'projectDir' | 'profilesDir' | 'target'
+    >,
+    absoluteTargetPath: string | undefined,
+): string[] {
+    const args: string[] = [];
+    if (options.projectDir) {
+        args.push('--project-dir', absoluteProjectDir);
+    }
+    if (options.profilesDir) {
+        args.push('--profiles-dir', options.profilesDir);
+    }
+    if (options.target) {
+        args.push('--target', options.target);
+    }
+    if (absoluteTargetPath) {
+        args.push('--target-path', absoluteTargetPath);
+    }
+    return args;
+}
 
 export const generateDocsHandler = async (
     options: GenerateDocsHandlerOptions,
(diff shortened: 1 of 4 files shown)


packages/cli/src/handlers/generateDocs.ts
```diff
@@ -0,0 +1,207 @@
...
+    const commonArgs = buildCommonDbtArgs(
+        absoluteProjectDir,
+        options,
+        absoluteTargetPath,
+    );
+
+    // Step 1: Run dbt docs generate
+    // dbt output streams directly to the terminal via stdio: 'inherit'
+    try {
+        const generateArgs = ['docs', 'generate'];
```
owlas: runs dbt docs for the user, are there any dbt args they might want to pass in here like variables etc.?

packages/cli/src/handlers/generateDocs.ts
```diff
@@ -0,0 +1,207 @@
...
+                properties: {
+                    executionId,
+                    step: 'inject_lineage',
+                    error: msg,
+                },
+            });
+            throw new Error(`Failed to inject Lightdash lineage:\n  ${msg}`);
+        }
+
+        // Step 2b: If --static, the HTML file was generated BEFORE we injected,
```
owlas: what's static? why do we allow this?

packages/cli/src/handlers/generateDocs.ts
```diff
@@ -0,0 +1,207 @@
...
+            durationMs: Date.now() - startTime,
+            skipInject: options.skipInject,
+            serve: options.serve,
+        },
+    });
+
+    // Step 3: Either print the static file path or start a local server.
+    // --static and --serve are mutually exclusive: static outputs a file,
+    // serve starts dbt's built-in HTTP server for interactive browsing.
+    if (options.static) {
```
owlas, line 193: do we need static **and** serve?

oli-rmsy: Yup, generally serve for dev and static for prod!

packages/cli/src/handlers/generateDocs.ts
```diff
@@ -0,0 +1,217 @@
...
+            String(options.port),
+            '--target-path',
+            ldDocsDir,
+            ...commonArgs,
+        ];
+
+        GlobalState.debug(`> Running: dbt ${serveArgs.join(' ')}`);
+        await execa('dbt', serveArgs, {
+            stdio: ['pipe', 'pipe', 'pipe'],
+        });
```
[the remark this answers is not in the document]
oli-rmsy: Good catch on the buffering concern — piping stdout/stderr on a long-running server would leak memory. However `stdio: 'inherit'` isn't the right fix here since we intentionally suppress dbt's noisy serve output (request logs, version info, etc.) and show our own clean message with a clickable URL instead.

Fixed by using `stdio: ['inherit', 'ignore', 'ignore']` which discards stdout/stderr entirely without buffering, while keeping stdin connected.

owlas (approved):

PR #20743, opened 2026-03-02
perf: replace correlated subquery with derived table JOIN in chart CTE

joaoviana:
Closes: N/A

### Description:

`getProjectChartsLastVersionCTE` used a correlated scalar subquery (`SELECT MAX(...) WHERE sqv.saved_query_id = sq.saved_query_id`) that executed once per chart row. For projects with 1000s of charts, this means 1000s of separate index lookups.

Replace with a derived table JOIN that computes all max version IDs in a single HashAggregate pass, then joins. Same results, significantly less I/O

### DIFF
--- a/packages/backend/src/models/SavedChartModel.ts
+++ b/packages/backend/src/models/SavedChartModel.ts
@@ -1315,19 +1315,31 @@ export class SavedChartModel {
         qb: Knex.QueryBuilder,
         projectUuid: string,
     ) {
-        const maxVersionSubquery =
-            '(SELECT MAX(sqv.saved_queries_version_id) FROM saved_queries_versions sqv WHERE sqv.saved_query_id = sq.saved_query_id)';
+        // Use a derived table for MAX version instead of a correlated subquery per row.
+        // This lets PostgreSQL compute all max versions in a single pass (HashAggregate)
+        // rather than doing N index lookups for N charts.
+        const latestVersions = this.database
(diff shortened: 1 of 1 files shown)


packages/backend/src/models/SavedChartModel.ts
```diff
@@ -1315,19 +1315,31 @@ export class SavedChartModel {
...
-            '(SELECT MAX(sqv.saved_queries_version_id) FROM saved_queries_versions sqv WHERE sqv.saved_query_id = sq.saved_query_id)';
+        // Use a derived table for MAX version instead of a correlated subquery per row.
+        // This lets PostgreSQL compute all max versions in a single pass (HashAggregate)
+        // rather than doing N index lookups for N charts.
+        const latestVersions = this.database
+            .select('saved_query_id')
+            .max('saved_queries_version_id as max_version_id')
+            .from(SavedChartVersionsTableName)
+            .groupBy('saved_query_id')
+            .as('latest');
```
almeidabbm, line 1326: Question: should this be scoped to the project? it would need a couple joins but currently this is calculating the max version id for all the charts in the DB

joaoviana: The `INNER JOIN` already filters to only project-relevant charts. Adding project-scoping joins inside the subquery would add complexity without it being needed @almeidabbm

PR #20773, opened 2026-03-03
fix: only show tiles panel when tiles tab is selected

rephus:
Closes: 

Screenshot from 2026-03-03 10-47-02

Before

[Screencast from 2026-03-03 10-05-10.webm](https://github.com/user-attachments/assets/5bf7d671-dbc7-4ef4-925f-e3f97e10fce9)

[Screencast from 2026-03-03 10-14-28.webm](https://github.com/user-attachments/assets/88fa60b8-4eeb-4765-a421-1c0ad017e95f)

After

[Screencast from 2026-03-03 10-05-30.webm](https://github.com/user-attachments/assets/0d32ef78-6f2c-4475-8c48-7c581810f728)

[Screencast from 2026-03-03 10-16-34.webm](https://github.com/user-attachments/assets/b48a490a-5a68-4118-9b6c-3e054a33c0f4)

### Description:
Fixed the tiles panel in FilterConfiguration to only render when both a draft filter rule exists and the tiles tab is selected. This prevents the panel from being displayed when it shouldn't be visible.

### DIFF
--- a/packages/frontend/src/features/dashboardFilters/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/features/dashboardFilters/FilterConfiguration/index.tsx
@@ -480,7 +480,7 @@ const FilterConfiguration: FC<Props> = ({
                     </Stack>
                 </Tabs.Panel>
 
-                {draftFilterRule && (
+                {draftFilterRule && selectedTabId === FilterTabs.TILES && (
                     <Tabs.Panel
                         value={FilterTabs.TILES}
                         w={500}


packages/frontend/src/features/dashboardFilters/FilterConfiguration/index.tsx
```diff
@@ -480,7 +480,7 @@ const FilterConfiguration: FC<Props> = ({
                     </Stack>
                 </Tabs.Panel>
 
-                {draftFilterRule && (
+                {draftFilterRule && selectedTabId === FilterTabs.TILES && (
```
rephus, line 483: The FilterConfiguration component has two tabs internally:                                            
  - "Filter Settings" (default) — the operator/value picker you see when clicking a filter              
  - "Chart tiles" — a list of ALL tiles with checkboxes to configure which tiles the filter applies to  
   
  Mantine's <Tabs.Panel> renders its children regardless of whether the tab is active — it just uses CSS
   display: none to hide inactive panels. So when you click a filter button:

  1. The popover opens on the "Settings" tab (lightweight — just an operator select and value input)
  2. But React also mounts the hidden "Chart tiles" panel, which renders TileFilterConfiguration — 120
  Checkbox + Tooltip + FieldSelect components, one per tile

  Adding selectedTabId === FilterTabs.TILES makes React skip mounting TileFilterConfiguration entirely
  until the user actually navigates to that tab. Since in view mode the tab bar isn't even shown (line
  349 — it only shows in edit mode), this component was being rendered for nothing.

  So the fix isn't really obscure — it's just "don't render 120 heavy components for a tab the user
  can't even see."

rephus: This is the component that was ausing the filter slowdown when there are many tiles, even from different tabs

Image

The fix does not render this part if the tab panel is not selected. 

---

The FilterConfiguration component has two tabs internally:

"Filter Settings" (default) — the operator/value picker you see when clicking a filter
"Chart tiles" — a list of ALL tiles with checkboxes to configure which tiles the filter applies to

PR #20793, opened 2026-03-03
refactor: replace isPrivate with inheritsFromOrgOrProject in CASL authorization layer

Spissable:
Relates to: https://linear.app/lightdash/issue/GLITCH-178/remove-isprivate-completely

### Description:
This is Phase 1 of dropping references to the old `isPrivate` property. Phase 1 revolves around CASL checks, which had certain rules based on this boolean.
Semantically the rules have been inverted and the abstraction makes sure that:

if inheritance chain reaches the root:
- and the root space inherits from parent (project/org)
- then `inheritsFromOrgOrProject` = `true`

otherwise:
- `inheritsFromOrgOrProject` = `false`

### DIFF
--- a/packages/common/src/authorization/organizationMemberAbility.ts
+++ b/packages/common/src/authorization/organizationMemberAbility.ts
@@ -40,9 +40,19 @@ const applyOrganizationMemberStaticAbilities: Record<
     },
     viewer(member, { can }) {
         applyOrganizationMemberStaticAbilities.member(member, { can });
+        // TODO: remove once we're confident that nobody is stuck on an old frontend version
+        can('view', 'Dashboard', {
+            organizationUuid: member.organizationUuid,
+            isPrivate: false,
+        });
         can('view', 'Dashboard', {
             organizationUuid: member.organizationUuid,
             inheritsFromOrgOrProject: true,
+        });
+        // TODO: remove once we're confident that nobody is stuck on an old frontend version
+        can('view', 'SavedChart', {
+            organizationUuid: member.organizationUuid,
+            isPrivate: false,
         });
         can('view', 'SavedChart', {
             organizationUuid: member.organizationUuid,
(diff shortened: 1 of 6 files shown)


packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -2402,16 +2402,18 @@ export class AsyncQueryService extends ProjectService {
             // TODO: Get all chartUuids for a given dashboard in the middleware.
             //       https://linear.app/lightdash/issue/CENG-110/front-load-available-charts-for-dashboard-requests
             access = [{ chartUuid: savedChart.uuid }];
-            const space =
```
Spissable, line 2437: The fact that we were still using `getSpaceSummary(savedChartSpaceUuid)` was actually a bug. If I understand it right, that basically resulted in the inheritance chain not taking any effect for this edge-case of JwtUser.

packages/frontend/src/features/sqlRunner/components/Header/HeaderView.tsx
```diff
@@ -72,7 +72,6 @@ export const HeaderView: FC = () => {
         subject('SavedChart', {
             organizationUuid: user.data?.organizationUuid,
             projectUuid,
-            isPrivate: savedSqlChart?.space.isPrivate,
```
Spissable, line 75: Note: this was removed because there wasn't any rule for `manage, subject('SavedChart'...` that relied on `isPrivate` anyways.

packages/frontend/src/features/sqlRunner/components/Header/HeaderView.tsx
```diff
@@ -84,7 +83,6 @@ export const HeaderView: FC = () => {
...
```
Spissable, line 87: Note: this was removed because there wasn't any rule for `promote, subject('SavedChart'...` that relied on `isPrivate` anyways.

packages/backend/src/services/SavedChartsService/SavedChartService.ts
```diff
@@ -932,6 +940,7 @@ export class SavedChartService
         return {
             ...savedChart,
             isPrivate: space.isPrivate,
+            inheritsFromOrgOrProject: !space.isPrivate,
```
Spissable: ⚠️ This is wrong and it's potentially used in the frontend CASL check ⚠️

almeidabbm (approved):
LGTM 🚀

PR #20843, opened 2026-03-04
feat: add Liquid template support for dynamic SQL dimensions

rephus:
Closes: https://linear.app/lightdash/issue/PROD-5654/unsupported-liquidjs

### Description:

There was a small blocker: liquidjs syntax {% %} was in conflict with jinja dbt syntax, so we need to find a way to support without reinventing the liquidjs syntax. For that, I suggested to detect liquidjs syntax and wrap it in a compatible {% raw %} block so dbt doesn't break

Screenshot from 2026-03-04 16-58-55

Screenshot from 2026-03-04 16-53-12

Screenshot from 2026-03-04 16-56-13

---

This PR adds Liquid template support for dynamic SQL in Lightdash dimensions and metrics, enabling conditional SQL generation based on parameter values.

**Key Features:**
- **Liquid Template Engine**: Introduces a new SQL templating system using Liquid syntax (`{% if %}`, `{% elsif %}`, `{% else %}`, `{% endif %}`) for conditional SQL blocks
- **Dynamic Date Granularity**: Adds a `date_granularity` parameter to the jaffle shop demo with options for Day, Week, Month, Quarter, and Year
- **Dynamic Dimensions**: Implements `dynamic_date` dimension that changes SQL based on the `date_granularity` parameter value
- **Dynamic Metrics**: Adds `dynamic_metric` that switches between counting events or summing event IDs based on parameter configuration

**Technical Implementation:**
- **dbt Integration**: Preprocesses YAML files before dbt compilation by wrapping Lightdash Liquid blocks in `{% raw %}` tags to prevent dbt's Jinja engine from processing them
- **Query-time Rendering**: Evaluates Liquid templates during query execution when parameter values are known, before standard parameter substitution
- **Parameter Context**: Builds Liquid context with parameters accessible via both `ld.parameters.x` and `lightdash.parameters.x` syntax
- **Chart Stability**: Improves pivot series ordering to be deterministic regardless of backend value column ordering, with proper handling of dotted values like version strings

**Example Usage:**
```sql
{% if ld.parameters.date_granularity == 'Day' %} 
  ${TABLE}.date
{% elsif ld.parameters.date_granularity == 'Week' %} 
  DATE_TRUNC('week', ${TABLE}.date)
{% else %} 
  ${TABLE}.date
{% endif %}
```

This enables powerful dynamic SQL patterns similar to Looker's Liquid templating, allowing dimensions and metrics to adapt their behavior based on user-selected parameter values.

test-frontend test-backend

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/parameters.ts
+++ b/packages/backend/src/utils/QueryBuilder/parameters.ts
@@ -1,5 +1,6 @@
 import {
     parameterRegex,
+    renderLiquidSql,
     UnexpectedServerError,
     type ParameterDefinitions,
     type ParametersValuesMap,
@@ -168,10 +169,15 @@ export const safeReplaceParametersWithTypes = ({
     sqlBuilder: WarehouseSqlBuilder;
     wrapChar?: string;
 }) => {
+    // Render Liquid template blocks ({% if %}/{% elsif %}/{% else %}/{% endif %})
+    // before parameter substitution. This evaluates conditional SQL based on parameter values,
+    // e.g., {% if ld.parameters.grain == "day" %} ${snapshot_date} {% endif %}
+    const liquidRenderedSql = renderLiquidSql(sql, parameterValuesMap);
+
     // First, get all parameter references from the original SQL using the standard function
     // This ensures we capture ALL parameter references, not just the ones we have values for
     const allParametersResult = safeReplaceParameters({
-        sql,
+        sql: liquidRenderedSql,
         parameterValuesMap,
         escapeString: sqlBuilder.escapeString.bind(sqlBuilder),
         quoteChar: sqlBuilder.getStringQuoteChar(),
@@ -214,7 +220,7 @@ export const safeReplaceParametersWithTypes = ({
     });
 
     // First replace string parameters (with quotes and escaping)
-    let processedSql = sql;
+    let processedSql = liquidRenderedSql;
     if (Object.keys(stringParameters).length > 0) {
         const stringResult = safeReplaceParameters({
             sql: processedSql,
--- a/packages/common/src/templating/liquidSql.ts
+++ b/packages/common/src/templating/liquidSql.ts
@@ -1,0 +1,101 @@
+import { Liquid } from 'liquidjs';
+
+/**
+ * Liquid template engine for SQL rendering.
+ * Uses default Liquid delimiters ({% %} for tags, {{ }} for output)
+ * to support conditional SQL syntax like:
+ *   {% if ld.parameters.grain == "day" %} ${snapshot_date} {% endif %}
+ *
+ * This is consistent with the existing Lightdash parameter syntax
+ * (${ld.parameters.x}) — just using Liquid delimiters instead of ${}.
+ *
+ * This is separate from the URL template engine (template.ts) which
+ * uses custom delimiters (${} for output).
+ */
+const liquidSqlEngine = new Liquid({
+    cache: true,
+    strictVariables: false, // Allow missing variables to resolve to falsy
+    strictFilters: false,
+});
+
+// Only activate Liquid rendering when the SQL contains Lightdash parameter
+// references inside {% %} tags. A bare {%  without ld.parameters/lightdash.parameters
+// could be unrelated syntax and should not be processed.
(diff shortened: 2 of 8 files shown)


packages/backend/src/dbt/dbtCliClient.ts
```diff
@@ -63,6 +64,77 @@ export const getDbtConfig = async (
     };
 };
 
+/**
+ * Find all .yml/.yaml files recursively in a directory,
+ * skipping node_modules, target, and dbt_packages.
+ */
+const findYamlFiles = async (dir: string): Promise<string[]> => {
```
rephus: This is the tricky part of the feature. 

liquidjs syntax {% %} was in conflict with jinja dbt syntax, so we need to find a way to support without reinventing the liquidjs syntax. 

For that, I suggested to detect liquidjs syntax and wrap it in a compatible {% raw %} block so dbt doesn't break

packages/common/src/templating/liquidSql.ts
```diff
@@ -0,0 +1,139 @@
...
+ * const sql = `{% if ld.parameters.grain == "day" %} DATE_TRUNC('day', "events".report_date)
+ *   {% elsif ld.parameters.grain == "week" %} DATE_TRUNC('week', "events".report_date)
+ *   {% else %} "events".report_date
+ *   {% endif %}`;
+ *
+ * renderLiquidSql(sql, { grain: 'day' });
+ * // => ' DATE_TRUNC(\'day\', "events".report_date)\n'
+ * ```
+ */
+export const renderLiquidSql = (
```
rephus, line 83: The key insight is that renderLiquidSql uses the full LiquidJS engine — it's   
  not regex-based. So any Liquid syntax that LiquidJS supports already works at query time out of the   
  box:                                                                                                  
                                                                                                        
  - {% case %} / {% when %} — switch statements                                                         
  - {% for %} — loops (e.g., generate UNION ALL per value)              
  - {% assign %} — variables
  - {{ ld.parameters.x }} — output tags (interpolation)
  - Filters like {{ value | upcase }}, | split, | join

packages/common/src/templating/liquidSql.ts
```diff
@@ -0,0 +1,139 @@
...
+        (match: string, offset: number, fullString: string) => {
+            // Don't double-wrap if already wrapped (check preceding text for {% raw %})
+            const preceding = fullString.slice(
+                Math.max(0, offset - 12),
+                offset,
+            );
+            if (preceding.includes('{% raw %}')) {
+                return match;
+            }
+            return `{% raw %}${match}{% endraw %}`;
```
rephus: wrap liquidjs in a jinja syntax to allow dbt compile

packages/backend/src/dbt/dbtCliClient.ts
```diff
@@ -265,28 +337,39 @@ export class DbtCliClient implements DbtClient {
...
-                const selector = this.selector?.trim();
+                // Wrap Lightdash Liquid blocks in {% raw %} before dbt compilation
+                const restoreYaml = await preprocessYamlForLiquid(
+                    this.dbtProjectDirectory,
+                );
 
-                if (selector) {
-                    if (!DbtCliClient.validateSelector(selector)) {
-                        throw new ParseError('Invalid dbt selector format');
+                try {
```
rephus: we are trying to not break any existing dbt build by adding try/catch guards around the new liquidjs code paths.

packages/backend/src/utils/QueryBuilder/parameters.ts
```diff
@@ -168,10 +169,15 @@ export const safeReplaceParametersWithTypes = ({
     sqlBuilder: WarehouseSqlBuilder;
     wrapChar?: string;
 }) => {
+    // Render Liquid template blocks ({% if %}/{% elsif %}/{% else %}/{% endif %})
+    // before parameter substitution. This evaluates conditional SQL based on parameter values,
+    // e.g., {% if ld.parameters.grain == "day" %} ${snapshot_date} {% endif %}
+    const liquidRenderedSql = renderLiquidSql(sql, parameterValuesMap);
```
rephus, line 175: this is where we convert liquidjs syntax

rephus: we could extract it to a separate method

PR #20901, opened 2026-03-06
fix: subtotal rows show '-' when all underlying values have no result

### LINKED ISSUE #20900: Subtotal rows should show '-' when all underlying values are '-' (no result)
## Bug Description

When a pivot table has subtotal rows, and the underlying raw values for a metric are `-` (indicating no result/null), the subtotal row should also display `-` instead of attempting to aggregate the missing values.

## Steps to Reproduce

1. Create a pivot table with subtotals enabled
2. Have a metric where some pivoted columns have no data (displayed as `-` in the raw rows)
3. Observe the subtotal row for that group

## Expected Behavior

If all underlying values in a subtotal group are `-` (no result), the subtotal should also display `-` to indicate there is no result to aggregate.

## Actual Behavior

The subtotal row shows a blank or calculated value even when the underlying raw value is `-`.

## Screenshot

In this example, the "Sprout Society - Unique Customers" column shows `-` for the raw row under 2026-Q1, but the subtotal row for `(1) 2026-Q1` does not show `-`:

Image

## Labels

🐛 bug

oli-rmsy:
## Summary

Fixes #20900 - the issue being end users were confused when sub total values were just blank. this introduces a '-' instead. See the screenshot below where there is completely blank field that leads to confusion:

image

- `getSubtotalValueFromGroup` previously returned `null` for both "column not applicable for subtotals" (e.g. the last grouped dimension column) and "no data for this combination" (no matching subtotal record, or SQL null from aggregating all nulls). The PivotTable treated all `null` returns as "don't render", producing blank cells.
- Now the function distinguishes the two cases:
  - `null` → column not applicable (render nothing, preserving existing behavior)
  - `undefined` → no data (`formatItemValue` renders `'-'`, matching regular row behavior)

## Test plan

- [ ] Create a pivot table with subtotals enabled where some pivoted columns have no data (shown as `-` in raw rows)
- [ ] Verify subtotal rows now show `-` for those columns instead of blank
- [ ] Verify subtotal rows still show correct aggregated values where data exists
- [ ] Verify the last grouped dimension column still shows nothing (not `-`) when expanding groups

### DIFF
--- a/packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
+++ b/packages/frontend/src/hooks/tableVisualization/getDataAndColumns.tsx
@@ -65,8 +65,14 @@ export function getGroupingValuesAndSubtotalKey(
 export function getSubtotalValueFromGroup(
     subtotal: Record<string, number> | undefined,
     columnId: string,
-) {
-    const subtotalColumnIds = Object.keys(subtotal ?? {});
+): number | null | undefined {
+    // No matching subtotal record exists (no data for this combination)
+    // Return undefined so formatItemValue will show '-'
+    if (subtotal === undefined) {
+        return undefined;
+    }
+
+    const subtotalColumnIds = Object.keys(subtotal);
 
     // If the subtotal column is not in the subtotalsGroup, return null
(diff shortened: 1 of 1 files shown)


oli-rmsy:
Tested and this is now working as expected: 
image

PR #21010, opened 2026-03-10
chore: add endpoints for setting impersonation switch

magnew:
Relates to: GLITCH-221

### Description:

Add endpoints to get and set `impersonation_enabled` on an org. Can only be changed by org admins

### DIFF
--- a/packages/backend/src/services/OrganizationService/OrganizationService.ts
+++ b/packages/backend/src/services/OrganizationService/OrganizationService.ts
@@ -7,6 +7,7 @@ import {
     CreateColorPalette,
     CreateGroup,
     CreateOrganization,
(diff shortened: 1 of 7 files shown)


packages/backend/src/services/OrganizationService/OrganizationService.ts
```diff
@@ -758,4 +758,42 @@ export class OrganizationService extends BaseService {
 
         return palette;
     }
+
+    async getImpersonationEnabled(user: SessionUser): Promise<boolean> {
+        const { organizationUuid } = user;
+        if (
+            user.ability.cannot(
+                'update',
```
magnew, line 773: The permissions to get this probably will be relaxed.

Issue #21023, opened 2026-03-10
Space with inherited permissions only visible to 3 users instead of all project members


torimwhaley:
## Bug Description

A root-level space configured with inherited permissions is only visible to 3 users, despite the project having 100+ members. Other spaces with the same inherited permission setup in the same project are accessible to all project members as expected.

The user cannot toggle the space to public — the UI doesn't allow forcing it.

## Suspected Root Cause

Likely a data inconsistency between `is_private` and `inherit_parent_permissions` on this space. During the migration (`20260129134608_add_inherit_parent_permissions_to_spaces`), root spaces were set to `inherit_parent_permissions = NOT is_private`. If `is_private` was incorrectly `true` for this space, the migration would have set `inherit_parent_permissions = false`, making it behave as a restricted space with only direct access grants.

## Steps to Investigate

Query the space state in production:

```sql
SELECT space_uuid, name, is_private, inherit_parent_permissions, parent_space_uuid
FROM spaces WHERE space_uuid = '<space_uuid>';
```

If `inherit_parent_permissions = false` (or `is_private = true`) when it shouldn't be, that confirms the data inconsistency.

Also check if there's a broader pattern — are there other spaces in the same state across other orgs?

Customer identifiers are in Pylon #11299.

## Relevant Code

- Space permission model: `packages/backend/src/models/SpacePermissionModel.ts`
- Access resolver: `packages/common/src/authorization/space/spaceAccessResolver.ts`
- Migration: `packages/backend/src/database/migrations/20260129134608_add_inherit_parent_permissions_to_spaces.ts`

PR #21041, opened 2026-03-11
fix: account for duplicate paths in space inheritance

magnew:
Closes: #21023

### Description:

A root-level space was showing only 3 users in "Who has access" instead of the expected 100+. The root cause is in how getInheritanceChains() determines whether a space inherits permissions from the org/project level. It uses an ltree @> (ancestor-of) join to walk from a space up to its root. The problem is that ltree paths are generated getLtreePathFromSlug(), which is lossy (dashes become underscores, so two distinct slugs like expert-unit and expert_unit), but also produces duplicate paths for duplicate slugs, which can come up in a few ways. 

When two root-level spaces in the same project share a path, the @> join matches both. If the unrelated space has inherit_parent_permissions: false and gets processed first (due to nlevel DESC ordering), the chain walker breaks early and incorrectly concludes the space doesn't inherit from org/project — making it appear private and hiding most users.

**Fix**
Replace the ltree @> join with a recursive CTE that walks the parent_space_uuid foreign key. This follows the actual space hierarchy rather than relying on the denormalized (and potentially duplicate) ltree path.

Performance is bounded by tree depth (~4-5 levels max), not width, and each recursion step is an indexed PK lookup — so this is effectively the same cost as the original query.

The output shape is identical (same columns, same leaf-to-root ordering), so no changes are needed in the downstream chain-walking logic or callers.

### DIFF
--- a/packages/backend/src/models/SpacePermissionModel.ts
+++ b/packages/backend/src/models/SpacePermissionModel.ts
@@ -428,27 +428,44 @@ export class SpacePermissionModel {
             async () => {
                 if (spaceUuids.length === 0) return {};
 
+                // NOTE: use a CTE instead of ltree path @> joins to walk the space
+                // hierarchy via parent_space_uuid FK. We have duplicate paths in the
+                // database due to legacy spaces and content as code and an ltree path
+                // @> join matches unrelated spaces due to those duplicates,
+                // breaking the inheritance chain and hiding users from "Who has access".
                 const ancestorRows: {
                     requested_space_uuid: string;
                     space_uuid: string;
                     name: string;
                     inherit_parent_permissions: boolean;
                     parent_space_uuid: string | null;
-                }[] = await this.database(`${SpaceTableName} as leaf`)
-                    .select({
-                        requested_space_uuid: 'leaf.space_uuid',
-                        space_uuid: 'ancestor.space_uuid',
-                        name: 'ancestor.name',
-                        inherit_parent_permissions:
-                            'ancestor.inherit_parent_permissions',
-                        parent_space_uuid: 'ancestor.parent_space_uuid',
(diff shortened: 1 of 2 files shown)


packages/backend/src/models/SpacePermissionModel.ts
magnew, line 1: This could also be knex, but its not necessarily easier to read

```js
const ancestorRows = await this.database                                                                                    
      .withRecursive('chain', (qb) => {                                                                                       
          qb.select({                                                                                                         
              requested_space_uuid: 'space_uuid',                                                                             
              space_uuid: 'space_uuid',                                       
              name: 'name',
              inherit_parent_permissions: 'inherit_parent_permissions',
              parent_space_uuid: 'parent_space_uuid',
          })
          .select(this.database.raw('1 AS depth'))
          .from(SpaceTableName)
          .whereIn('space_uuid', spaceUuids)
          .whereNull('deleted_at')
          .unionAll((qb2) => {
              qb2.select({
                  requested_space_uuid: 'c.requested_space_uuid',
                  space_uuid: 's.space_uuid',
                  name: 's.name',
                  inherit_parent_permissions: 's.inherit_parent_permissions',
                  parent_space_uuid: 's.parent_space_uuid',
              })
              .select(this.database.raw('c.depth + 1'))
              .from(`${SpaceTableName} as s`)
              .join('chain as c', 's.space_uuid', 'c.parent_space_uuid')
              .whereNull('s.deleted_at');
          });
      })
      .select('requested_space_uuid', 'space_uuid', 'name', 'inherit_parent_permissions', 'parent_space_uuid')
      .from('chain')
      .orderBy('requested_space_uuid')
      .orderBy('depth', 'asc');
```

PR #21060, opened 2026-03-11
feat: add includeDateZoomCapabilities to saved chart GET endpoint

almeidabbm:
Closes:


almeidabbm:
Folded into 03-11-feat_date-zoom-applied-backend — date zoom capabilities are now computed frontend-side

PR #21067, opened 2026-03-12
feat: add async query queue lifecycle statuses

joaoviana:
## Summary
- add explicit async query lifecycle states for queued, executing, and expired queries
- expire stale queued jobs in the NATS worker and report them to Sentry
- update frontend polling and loading copy to distinguish queued vs executing queries

## Validation
- pnpm --filter backend run generate-api
- pnpm --filter frontend run typecheck
- pnpm --filter backend run test-sequential -- AsyncQueryService.test.ts
- targeted eslint on edited backend/frontend files

## Notes
- full backend typecheck still fails on an unrelated pre-existing error in packages/backend/src/models/ProjectCompileLogModel.ts

### DIFF
--- a/packages/frontend/src/features/queryRunner/executeQuery.test.ts
+++ b/packages/frontend/src/features/queryRunner/executeQuery.test.ts
@@ -50,9 +50,10 @@ describe('executeQuery', () => {
             executeSqlPivotQuery('project-uuid', {
                 sql: 'select 1',
                 pivotConfiguration: {
-                    columns: [],
-                    rows: [],
-                    values: [],
+                    indexColumn: undefined,
+                    valuesColumns: [],
+                    groupByColumns: undefined,
+                    sortBy: undefined,
                 },
             }),
         ).rejects.toThrow('Query expired in queue');


packages/backend/src/nats/NatsWorker.ts
```diff
@@ -217,6 +225,15 @@ export class NatsWorker {
             jobMetadata,
         );
 
+        const canRun = await this.prepareQueryForExecution(
+            parsed.payload.queryUuid,
+            workerLabel,
+        );
+        if (!canRun) {
+            message.ack();
```
notgiorgi:
```suggestion
            message.term();
```

packages/backend/src/models/QueryHistoryModel/QueryHistoryModel.ts
```diff
@@ -186,11 +186,49 @@ export class QueryHistoryModel {
 
         // only update pending queries to ready
         if (update.status === QueryHistoryStatus.READY) {
-            void query.andWhere('status', QueryHistoryStatus.PENDING);
+            void query.andWhere('status', QueryHistoryStatus.EXECUTING);
         }
```
notgiorgi, line 190: Let's double check this

notgiorgi: https://github.com/lightdash/lightdash/blob/634222e9debb891420d8e4507d5736476881fa7a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts#L2715-L2735

https://github.com/lightdash/lightdash/blob/634222e9debb891420d8e4507d5736476881fa7a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts#L2044-L2068

joaoviana: Double-checked this. Both `READY` paths already move the row to `EXECUTING` first:

* cache-hit path calls `updateStatusToExecuting()` before `update(... READY ...)` 
* main execution path calls `updateStatusToExecuting()` before `runAsyncWarehouseQuery()`, and that later does `update(... READY ...)`

So the `READY` guard on `EXECUTING` is correct. 

Also added test coverage for I also added test coverage for both cache-hit and non-NATS cache-miss paths to assert that `updateStatusToExecuting()` happens before `READY`.

packages/backend/src/nats/NatsWorker.ts
```diff
@@ -217,6 +225,15 @@ export class NatsWorker {
             jobMetadata,
         );
 
+        const canRun = await this.prepareQueryForExecution(
+            parsed.payload.queryUuid,
+            workerLabel,
+        );
```
notgiorgi: I think this should live in here, to keep the free of the query-related logic  
  
https://github.com/lightdash/lightdash/blob/634222e9debb891420d8e4507d5736476881fa7a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts#L1762-L1776

notgiorgi:
```
updateProcessingStartedAt
```

and `prepareQueryForExecution` could be merged probably, because it just updates a timestamp

PR #21072, opened 2026-03-12
feat: add configurable row limit for pre-aggregate materializations

tatianainama:
### Description:

Added a new `materializationRowLimit` configuration option to the pre-aggregates settings. This allows limiting the number of rows processed during pre-aggregate materialization by setting the `PRE_AGGREGATES_ROW_LIMIT` environment variable. When not set, we will still use `MAX_SAFE_INTEGER`


packages/backend/src/ee/services/PreAggregateMaterializationService/buildMaterializationMetricQuery.ts
```diff
@@ -305,19 +319,27 @@ export const buildMaterializationMetricQuery = ({
 
     const metricFieldIds = Array.from(selectedMetricFieldIds);
 
+    const SYSTEM_MAX_ROWS = 10_000_000;
```
tatianainama, line 322: @IrakliJani using 10M as system max rows. This can be overwritten by env variables / pre-aggregate definition in yml

PR #21076, opened 2026-03-12
fix: reuse nats client in both worker and backend

IrakliJani:
Refactored NATS client architecture by replacing `NatsJobClient` with a more robust `NatsClient` that handles connection management, stream setup, and graceful shutdown.


packages/backend/src/clients/NatsClient.ts
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        const conn = await connect({
+            servers: url,
+            maxReconnectAttempts: -1,
+        });
+        this.connection = conn;
+        Logger.info(`NATS #${this.connectionId} connected to ${url}`);
+
+        await this.ensureStreams();
```
notgiorgi, line 188: this seems wrong: back-end pods will now connect to streams

packages/backend/src/clients/NatsClient.ts
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    // ── Stream infrastructure ───────────────────────────────────
+
+    /** Idempotently create all streams + consumers from STREAM_CONFIGS. */
+    async ensureStreams(): Promise<void> {
+        const conn = this.getConnection();
+        const jsm = await conn.jetstreamManager();
+        await Promise.all(
+            Object.values(STREAM_CONFIGS).map(
```
notgiorgi, line 114: this doesn't respect `--stream` anymore  
https://github.com/lightdash/lightdash/blob/2bd7b57f052a03f0ce57853818c0c20b2ca48ee1/packages/backend/src/nats/NatsWorker.ts#L70

Issue #21086, opened 2026-03-12
Fetching dashboards and charts by slug can fail if there are duplicates in other projects


magnew:
### Description

Fetching dashboards and charts by slug can fail if there are duplicates in other projects.

We do not pass project ID when fetching dashboards and charts by slug, so those fetches can fail when a slug is duplicated across projects. We should pass the project uuid to these queries and use it to scope correctly. 

### Steps to Reproduce the Bug or Issue

n/a

### Cloud or self-hosting

None

PR #21088, opened 2026-03-12
fix: disambiguate dashboard and chart slug lookups by project

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/21086

## Summary

Dashboard and chart slugs are not unique across projects — they're just a nullable indexed text column with no cross-project uniqueness constraint.

When two projects contain content with the same slug (e.g. via `createPreview` with `copyContent`), `GET /api/v1/dashboards/:slug` or `GET /api/v1/saved/:slug` returns whichever has the most recent version, which may belong to a different project.
If the user doesn't have access to that project's space, they get a 403.

The fix adds an optional `projectUuid` query parameter to both the dashboard and chart GET-by-slug endpoints. The frontend already knows the project UUID from route params (`/projects/:projectUuid/dashboards/:dashboardUuid`) and now passes it as `?projectUuid=X`. The model uses it to scope the slug lookup with `WHERE project_uuid = ?`.

The parameter is optional so existing API consumers (public API, schedulers, etc.) continue to work unchanged — UUID-based lookups are unaffected since UUIDs are globally unique.

#### Steps to reproduce
1. Create a dashboard in a project (jaffle shop)
2. Create a preview from that project (copy content)
3. Verify that the dashboards both have the same slug (this is expected cross project)
4. Go to one of the projects and request the dashboard via slug `projects/<projectuuid>/dashboards/<slug>`
5. Verify that regardless of the project you're in, you will always get the latest created/changed dashboard with the slug

**Before**

https://github.com/user-attachments/assets/ca51f344-821d-4f18-bbea-d00524f74323

**After**

https://github.com/user-attachments/assets/d0f42f36-74c7-4c20-a954-f46e88c2cd8e

### DIFF
--- a/packages/frontend/src/hooks/dashboard/useDashboard.ts
+++ b/packages/frontend/src/hooks/dashboard/useDashboard.ts
@@ -161,7 +161,10 @@ export const useDashboardQuery = ({
  * @param dashboardUuid The dashboard uuid
  * @returns The latest dashboard or null if the dashboard is up to date
  */
-export const useDashboardVersionRefresh = (dashboardUuid: string) => {
+export const useDashboardVersionRefresh = (
+    dashboardUuid: string,
+    projectUuid?: string,
+) => {
     const queryClient = useQueryClient();
 
     return useMutation<Dashboard | null, ApiError, Dashboard | undefined>({
@@ -172,7 +175,10 @@ export const useDashboardVersionRefresh = (dashboardUuid: string) => {
                     throw new Error('Current dashboard is undefined');
                 }
 
-                const latestDashboard = await getDashboard(dashboardUuid);
+                const latestDashboard = await getDashboard(
+                    dashboardUuid,
+                    projectUuid,
+                );
 
                 const currentTime = new Date(
                     currentDashboard.updatedAt,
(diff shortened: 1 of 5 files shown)


packages/frontend/src/hooks/dashboard/useDashboard.ts
```diff
@@ -130,12 +136,13 @@ export const useDashboardsAvailableFilters = (
 
 export const useDashboardQuery = (
```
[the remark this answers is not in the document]
almeidabbm: Fixed in 4bce61c — added `projectUuid` to both `useDashboardQuery` and `useSavedQuery` query keys.

packages/frontend/src/hooks/dashboard/useDashboard.ts
```diff
@@ -128,14 +134,19 @@ export const useDashboardsAvailableFilters = (
         },
     );
 
-export const useDashboardQuery = (
-    id?: string,
-    useQueryOptions?: UseQueryOptions<Dashboard, ApiError>,
-) => {
+export const useDashboardQuery = ({
+    id,
```
almeidabbm: this should be called uuidOrSlug

almeidabbm: Fixed — renamed `id` to `uuidOrSlug` in both `useDashboardQuery` and `useSavedQuery` hooks, plus all call sites.

PR #21127, opened 2026-03-13
fix: introduce slug redirect so chart/dashboard lookup by slug is deduplicated

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/21086

## Summary

Fixes a bug where dashboard/chart slug lookups can return wrong content when slugs collide across projects (e.g. via `createPreview` with `copyContent`).

- **Slug redirect**: Added `useSlugRedirect` hook that replaces slug URLs with UUID URLs after the initial resource load, preventing subsequent API calls from using ambiguous slugs
- **Project-scoped lookup**: Backend GET endpoints for dashboards (`/dashboards/:slug`) and charts (`/saved/:slug`) now accept an optional `?projectUuid` query param to disambiguate slug lookups. Frontend passes `projectUuid` from URL params on the initial fetch.
- **Hook refactor**: `useDashboardQuery` and `useSavedQuery` refactored to named args (`{ uuidOrSlug, projectUuid, useQueryOptions }`) with `projectUuid` included in TanStack Query cache keys

**Steps to reproduce**

1. Create a dashboard in a project (jaffle shop)
2. Create a preview from that project (copy content)
3. Verify that the dashboards both have the same slug (this is expected cross project)
4. Go to one of the projects and request the dashboard via slug `projects/<projectuuid>/dashboards/<slug>`
5. Verify that you get redirected to the correct dashboard/chart in the project

https://github.com/user-attachments/assets/d7c476a8-c413-4e16-bbc6-59ac407e77df

### DIFF
--- a/packages/frontend/src/hooks/useSlugRedirect.ts
+++ b/packages/frontend/src/hooks/useSlugRedirect.ts
@@ -24,7 +24,11 @@ export function useSlugRedirect(
         if (urlParam === resolvedUuid) return;
         if (isValidUuid(urlParam)) return;
 
-        const newPath = location.pathname.replace(urlParam, resolvedUuid);
+        const segments = location.pathname.split('/');
+        const idx = segments.lastIndexOf(urlParam);
+        if (idx === -1) return;
+        segments[idx] = resolvedUuid;
+        const newPath = segments.join('/');
         void navigate(newPath + location.search + location.hash, {
             replace: true,
         });


packages/frontend/src/hooks/useSlugRedirect.ts
```diff
@@ -0,0 +1,32 @@
...
+) {
+    const navigate = useNavigate();
+    const location = useLocation();
+
+    useEffect(() => {
+        if (!urlParam || !resolvedUuid) return;
+        if (urlParam === resolvedUuid) return;
+        if (isValidUuid(urlParam)) return;
+
+        const newPath = location.pathname.replace(urlParam, resolvedUuid);
```
pimmey: A small one, but if someone has a dashboard named `dash` the `/dashboards/` part of the url will be replaced (before it reaches `dash`). Could also match the instance name 🤷 

This guy Claude suggests:
```
 Consider a more targeted replacement:

  const segments = location.pathname.split('/');
  const idx = segments.lastIndexOf(urlParam);
  if (idx !== -1) {
      segments[idx] = resolvedUuid;
  }
  const newPath = segments.join('/');
```

almeidabbm: Fixed — switched to segment-based replacement so it only targets the exact path segment matching the slug.

almeidabbm: Fixed in: [chore: replace by segment instead of whole URL](https://github.com/lightdash/lightdash/pull/21127/changes/de9d1db1d7a19c1e2efe76048cb62bb556876b21)

**Before**

https://github.com/user-attachments/assets/b38c1d78-0d7f-4a34-983d-0a413f3b98e1



**After**

https://github.com/user-attachments/assets/fbe6536b-1228-4f42-b104-d441ddc4a1fb

Issue #21153, opened 2026-03-16
Chart query SQL is unformatted when copied or opened in SQL Runner


jesshitchcock:
## Bug

When copying or opening the **chart query** SQL from the explorer's SQL card (via the copy button or "Open in SQL Runner"), the SQL is unformatted. The **base query** is not affected because the backend already builds it with newlines.

### Steps to reproduce

1. Open any explore and run a query with a pivot (so the "Chart Query" tab appears)
2. Expand the SQL section and select "Chart Query"
3. Either copy the SQL using the clipboard icon or click "Open in SQL Runner"
4. The SQL is unformatted — a dense string with minimal line breaks

### Expected behavior

The copied/opened SQL should be formatted the same way it appears in the SQL card display.

### Fix

1. **Frontend**: Format the SQL in `SqlCard.tsx` with `sql-formatter` before passing to copy and "Open in SQL Runner", matching what `RenderedSql` already does for display

PR #21156, opened 2026-03-16
chore: fix chart query formatting

jesshitchcock:
Closes: https://github.com/lightdash/lightdash/issues/21153

### Description:
Previously, copying SQL from the SQL card or opening it in the SQL Runner gave you one long unformatted line which was difficult to work with. 

### Changes
The SQL is now automatically formatted in the SQL card before it hits the `open in SQL runner` or `copy` button, using the correct syntax for your warehouse (BigQuery, Postgres, Snowflake, etc.). If formatting ever fails for any reason, it falls back to the raw SQL so nothing breaks but logs an error. 

buttons_for_copying_sql

[Loom explainer](https://www.loom.com/share/0b6a8f8aa6c945938a4f2dcd89b89344)

### DIFF
--- a/packages/frontend/src/components/Explorer/SqlCard/SqlCard.tsx
+++ b/packages/frontend/src/components/Explorer/SqlCard/SqlCard.tsx
@@ -81,8 +81,8 @@ const SqlCard: FC<SqlCardProps> = memo(({ projectUuid }) => {
                 language: getLanguage(project?.warehouseConnection?.type),
             });
         } catch (e) {
-            console.error(
-                'Error rendering SQL:',
+            console.warn(
+                'Error formatting SQL:',
                 e instanceof Error ? e.message : 'Unknown error occurred',
             );
             return selectedSql;


packages/frontend/src/components/Explorer/SqlCard/SqlCard.tsx
```diff
@@ -62,6 +74,21 @@ const SqlCard: FC<SqlCardProps> = memo(({ projectUuid }) => {
...
         selectedView === 'pivotQuery' ? data?.pivotQuery : data?.query;
 
+    const formattedSql = useMemo(() => {
+        if (!selectedSql) return '';
+        try {
+            return format(selectedSql, {
+                language: getLanguage(project?.warehouseConnection?.type),
+            });
+        } catch (e) {
+            console.error(
```
Spissable: Nitpick: I would change it to `console.warn` since it's not really essential functionality and instead of `'Error rendering SQL:'` -> `'Error formatting SQL: '`

jesshitchcock: fixed in next commit

Issue #20674, opened 2026-02-26
fix: metrics-with-time-dimensions endpoint silently truncates results to 50


almeidabbm:
## Current endpoint issues

The `GET /api/v1/projects/{projectUuid}/dataCatalog/metrics-with-time-dimensions` endpoint returns a non-deterministic, incomplete subset of metrics. Results change after every project refresh, and entire models can appear/disappear from the response.

## Root Cause

`CatalogService.getAllCatalogMetricsWithTimeDimensions()` calls `catalogModel.search()` **without pagination args**. However, `CatalogModel.search()` always constructs a pagination object with hardcoded defaults (`{ page: 1, pageSize: 50 }`), which means results are silently capped at 50 rows — even though `KnexPaginate.paginate()` supports returning all results when `paginateArgs` is `undefined`.

The 50 metrics selected depend on `search_rank` ordering, which changes on every catalog re-index (project refresh). After the 50-row cap, `getAvailableCompareMetrics()` further filters to only metrics with valid time dimensions, reducing the final count even more.

**Code path:**
1. `CatalogService.ts:1407-1418` — calls `catalogModel.search()` with no `paginateArgs`
2. `CatalogModel.ts:1208-1211` — always constructs `{ page: 1, pageSize: 50 }` even when `paginateArgs` is `undefined`
3. `KnexPaginate (pagination/index.ts:13-52)` — has a `return all` path when `paginateArgs` is `undefined`, but it's never reached from `search()`

## Reproduction

Using the jaffle shop demo project (345 spotlight metrics, 179 with valid time dimensions):

```bash
# Returns only 25-48 metrics (varies per refresh) instead of 179
curl -H 'Authorization: ApiKey $TOKEN' \
  "$SITE_URL/api/v1/projects/$PROJECT_UUID/dataCatalog/metrics-with-time-dimensions" \
  | jq '.results | length'

# Trigger a refresh, then query again — completely different set of metrics
curl -X POST -H 'Authorization: ApiKey $TOKEN' \
  "$SITE_URL/api/v1/projects/$PROJECT_UUID/refresh"

sleep 35

curl -H 'Authorization: ApiKey $TOKEN' \
  "$SITE_URL/api/v1/projects/$PROJECT_UUID/dataCatalog/metrics-with-time-dimensions" \
  | jq '.results | length'
```

Three consecutive refreshes produced three different result sets (27, 48, 25 metrics) with almost no overlap.

[...]

PR #21172, opened 2026-03-16
feat: add v2 API endpoint for paginated metrics with time dimensions

tatianainama:
Closes: #20674

### Description:

Added paginated v2 endpoint for metrics with time dimensions. Supports sorting for deterministic results

#### Context

The v1 endpoint can silently return incomplete results, as it currently filters in memory after db pagination. This returns a subset of the expected results. This new endpoint moves the filtering to the db and exposes proper pagination results so api consumers get predictable results.

### DIFF
--- a/packages/backend/src/services/CatalogService/CatalogService.ts
+++ b/packages/backend/src/services/CatalogService/CatalogService.ts
@@ -1415,6 +1415,7 @@ export class CatalogService<
             },
             tablesConfiguration:
                 await this.projectModel.getTablesConfiguration(projectUuid),
+            hasTimeDimension: true,
         });
 
         const filteredMetrics = allCatalogMetrics.data.filter(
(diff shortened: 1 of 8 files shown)


packages/backend/src/services/CatalogService/CatalogService.ts
```diff
@@ -1435,6 +1436,50 @@ export class CatalogService<
...
+                filter: CatalogFilter.Metrics,
+            },
+            tablesConfiguration:
+                await this.projectModel.getTablesConfiguration(projectUuid),
+            hasTimeDimension: true,
+            paginateArgs,
+            sortArgs,
+        });
+
+        return results as KnexPaginatedData<CatalogField[]>;
```
tatianainama, line 1507: safe to cast, we are filtering query with `CatalogType.Field` but `.search()` returns `CatalogItem` anyways. We need to improve type narrowing in search to avoid this casting

PR #21213, opened 2026-03-17
fix: stop dropping models without config.materialized

almeidabbm:
## Summary

- Remove redundant filter in `getModelsFromManifest` that required `config.materialized` to be truthy
- The first filter already excludes ephemeral models (`config?.materialized !== 'ephemeral'`), making the second filter unnecessary
- Models without `config.materialized` set (e.g. views using default materialization) were silently dropped after #21064 unified all code paths to go through `getModelsFromManifest`

## Test plan

- [x] Added unit tests for `getModelsFromManifest` covering: tables, views, missing materialized, ephemeral exclusion, non-model exclusion, and mixed scenarios
- [x] Verified tests fail without the fix and pass with it
- [x] `pnpm -F common typecheck` passes

### DIFF
--- a/packages/common/src/types/dbt.ts
+++ b/packages/common/src/types/dbt.ts
@@ -787,13 +787,7 @@ export const getModelsFromManifest = (
         );
     }
     const adapterType = manifest.metadata.adapter_type;
-    return models
-        .filter(
-            (model) =>
-                model.config?.materialized &&
-                model.config.materialized !== 'ephemeral',
-        )
-        .map((model) => normaliseModelDatabase(model, adapterType));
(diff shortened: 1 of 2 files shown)


packages/common/src/types/dbt.ts
```diff
@@ -787,13 +787,7 @@ export const getModelsFromManifest = (
         );
     }
     const adapterType = manifest.metadata.adapter_type;
-    return models
-        .filter(
-            (model) =>
-                model.config?.materialized &&
-                model.config.materialized !== 'ephemeral',
```
almeidabbm, line 794: Filter already exists above

Issue #19635, opened 2026-01-21
Improve search behavior for deeply nested field groups in Tables UI


torimwhaley:
### Description

In the Query From Tables UI, users can organize metrics and dimensions into up to three nested group levels. However, the search experience breaks down at the third nesting level.

When searching for fields, matches inside deeply nested (third-level) groups are often not returned. This makes it hard to discover fields unless users already know exactly where they live in the hierarchy

This is especially noticeable in projects with heavily structured field groupings.

### What problem does this solve?

As field hierarchies grow, search becomes the primary way users navigate metrics and dimensions. If search does not reliably surface fields in deeply nested groups, users may assume fields don’t exist, duplicate logic, or abandon structured grouping altogether.

PR #21224, opened 2026-03-18
feat: improve explore search for nested field groups

tatianainama:
Closes: #19635

### Description:
Fixed search not finding fields in deeply nested groups (was limited to 2 levels) by making the visibility check recursive, matching the non-virtualized tree.
I also bumped the `MAX_GROUP_DEPTH` from 3 to 5, to display in the UI the tree. I included an indicator that shows if the found field exceeded the UI max depth limit

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
@@ -24,6 +24,7 @@ import {
 import {
     IconAlertTriangle,
     IconFilter,
+    IconHierarchyOff,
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
```diff
@@ -254,7 +256,7 @@ const TreeSingleNodeComponent: FC<Props> = ({ node }) => {
                     withinPortal
                     maw={300}
                     multiline
-                    label={messages.join('\n')}
+                    label={messages.map((m) => m.message).join('\n')}
```
tatianainama, line 259: current bug :grimacing:

Issue #21233, opened 2026-03-18
nested aggregate metrics can be emitted twice when combined with fanout-protection CTEs


notgiorgi:
## Summary

Nested aggregate metrics can be compiled twice when a query triggers both:

1. the fanout-protection / deduplication CTE path
2. the nested-aggregate CTE path

When that happens, the same metric alias is selected once into `na_base` and then selected again from `nested_agg_results`, producing duplicate output columns. This later surfaces as an ambiguous or unresolved column error depending on the warehouse.

## Impact

Queries that use a nested aggregate metric together with dimensions that activate fanout-protection joins can fail at runtime.

Observed failure mode in repro:

- `ORDER BY "<metric_alias>" is ambiguous`

Other warehouses may surface this as unresolved or duplicate-column analysis errors instead.

### 1. Add a repro metric to the jaffle-shop demo model

File:

- `examples/full-jaffle-shop-demo/dbt/models/fanouts_examples/fanouts_accounts.yml`

Under `models[0].meta.metrics`, add:

```yml
arr_viability_score_repro:
    type: number
    label: 'Repro: ARR Viability Score'
    sql: |
        CASE
          WHEN ${total_estimated_annual_recurring_revenue} IS NULL THEN NULL
          ELSE 1 + (
            0.99 * COALESCE(
              100.0 * (
                MAX(${total_estimated_annual_recurring_revenue}) OVER () - ${total_estimated_annual_recurring_revenue}
              ) / NULLIF(
                MAX(${total_estimated_annual_recurring_revenue}) OVER () - MIN(${total_estimated_annual_recurring_revenue}) OVER (),
                0
              ),
              0
            )
          )
        END
```

### 2. Sync the demo project

Refresh the project so the new metric is available in Lightdash.

### 3. Run this query shape

Explore:

- `fanouts_accounts`

Dimension:

- `fanouts_users.job_title`

Metric:

- `arr_viability_score_repro`

Sort:

- `arr_viability_score_repro DESC`

Equivalent query payload:

```json
{
    "exploreName": "fanouts_accounts",
    "dimensions": ["fanouts_users_job_title"],
    "metrics": ["fanouts_accounts_arr_viability_score_repro"],
    "filters": {},
    "sorts": [
        {
            "fieldId": "fanouts_accounts_arr_viability_score_repro",
            "descending": true
        }
    ],
    "limit": 500,
    "tableCalculations": [],
    "additionalMetrics": [],
    "metricOverrides": {}
}
```

[...]

PR #21236, opened 2026-03-18
fix: prevent duplicate nested aggregate metrics in fanout queries

notgiorgi:
Closes: https://github.com/lightdash/lightdash/issues/21233

### Description:

Fixed a bug where nested aggregate metrics were being emitted multiple times in generated SQL queries when fanout CTEs were also present. The issue occurred because nested aggregate outer metrics were being processed both by the nested aggregation CTE flow and the regular metric processing flow.

The fix introduces a check to exclude nested aggregate outer metrics from being processed in the regular metric flows, ensuring they are only materialized once through the nested_agg CTE. This prevents duplicate metric definitions in the final SQL query.

Added comprehensive test coverage for both regular nested aggregate metrics and cross-table nested aggregate metrics when combined with fanout scenarios to verify the metrics are emitted exactly once.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
@@ -1698,6 +1698,11 @@ export class MetricQueryBuilder {
         const metricsObjects = allMetricsToProcess.map((field) =>
             this.getMetricFromId(field),
         );
+        const nestedAggOuterIds = new Set(
+            this.getMetricsWithNestedAggregates().map(
+                ({ outerMetricId }) => outerMetricId,
+            ),
+        );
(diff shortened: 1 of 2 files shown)


packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
```diff
@@ -2251,24 +2262,31 @@ export class MetricQueryBuilder {
...
-                                name: unaffectedMetricsCteName,
-                                metrics: unaffectedMetrics.map((m) =>
-                                    getItemId(m),
-                                ),
-                            },
-                        ]);
+                ...metricsWithCteReferences
+                    .filter(
+                        (metric) => !nestedAggOuterIds.has(getItemId(metric)),
+                    )
```
notgiorgi, line 2268: just adds the filter

PR #21259, opened 2026-03-19
feat(cli): make project selection optional during login

Spissable:
Closes: https://linear.app/lightdash/issue/GLITCH-252/improve-project-creation-flow-for-agents

### Description:
One slightly uncomfortable flow has been that the cli's login command forced you to select a project. However, when all your intention is to CREATE a new project, it feels wrong to select a project first and adds risk that we end up overwriting the selected project if we're not careful.

This PR introduces a few things to improve that situation:

### 1) At the project selection after logging in you now have an option to not select a project:
image

To test run `pnpm -F @lightdash/cli dev login http://localhost:3000`

### 2) Same thing when trying to set a project explicitly

To test run `pnpm -F @lightdash/cli dev config set-project`

### 3) Extra command to quickly unset the project

To test run `pnpm -F @lightdash/cli dev config unset-project`

### 4) User friendly error when trying to run a command that requires a project, but not project is being provided:

image

Note: to verify that these commands work as expected you can keep an eye on `cat ~/.config/lightdash/config.yaml` and verify that `project` and `projectName` update accordingly.

### DIFF
--- a/packages/cli/src/handlers/setProject.ts
+++ b/packages/cli/src/handlers/setProject.ts
@@ -2,7 +2,7 @@ import { OrganizationProject } from '@lightdash/common';
 import inquirer from 'inquirer';
 import { URL } from 'url';
 import { LightdashAnalytics } from '../analytics/analytics';
-import { getConfig, setProject } from '../config';
+import { getConfig, setProject, unsetProject } from '../config';
 import GlobalState from '../globalState';
 import { lightdashApi } from './dbt/apiClient';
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/handlers/setProject.ts
```diff
@@ -36,17 +39,29 @@ export const setProjectCommand = async (name?: string, uuid?: string) => {
...
             {
                 type: 'list',
                 name: 'project',
-                choices: projects.map((project) => ({
-                    name: project.name,
-                    value: project.projectUuid,
-                })),
+                choices: [
+                    {
+                        name: "Don't select a project",
```
magnew, line 49: Does anything work if you don't select a project? I guess you can create a project.

I'm basically wondering if we should push people (not agents) NOT to use this option. If this isn't a very useful state for users, maybe it shouldn't be the default. 

If you are going to need a project to do anything useful, this could be "Select a project later"

Spissable: > Does anything work if you don't select a project? I guess you can create a project.

That's exactly the use-case that I built this for. Because I ended up forgetting the `--create` in the deployment and overwrote the previously selected project, because I was forced to select one.

Spissable: > If you are going to need a project to do anything useful, this could be "Select a project later"

Imo the output afterwards clarifies that pretty well!

image

Issue #21267, opened 2026-03-20
I would like to be able to sync my sql runner charts to Google Sheets


almeidabbm:
### Description

Google sheets sync is already available for explorer charts but it's a lacking feature on sql runner charts.

### What problem does this solve?

Users would like to be able to sync their SQL Runner charts/results to google sheets.

PR #21269, opened 2026-03-20
feat: add Google Sheets sync backend for SQL runner charts

almeidabbm:
Relates to: https://github.com/lightdash/lightdash/issues/21267

## Summary

Wires SQL runner saved charts into the existing scheduler infrastructure so they can be synced to Google Sheets on a recurring schedule.

- Adds `SqlChartScheduler` type to the `Scheduler` union and `savedSqlUuid`/`savedSqlName` fields to `SchedulerBase`
- Updates `SchedulerModel` to query, create, and join SQL chart schedulers
- Extends the `uploadGsheets` scheduler task with a new branch that loads the saved SQL chart, executes the query against the warehouse, and writes raw results to Google Sheets via `appendCsvToSheet`
- Adds `GET/POST /sqlRunner/saved/{uuid}/schedulers` endpoints to the SQL runner controller
- Updates `SchedulerService.getSchedulerResource` to resolve SQL chart project context

The existing scheduler infrastructure (cron, retries, error handling, auto-disable) is fully reused. SQL runner results are written as raw values since they don't go through Lightdash's formatting layer.

### DIFF
--- a/packages/backend/src/controllers/sqlRunnerController.ts
+++ b/packages/backend/src/controllers/sqlRunnerController.ts
@@ -669,7 +669,7 @@ export class SqlRunnerController extends BaseController {
             status: 'ok',
             results: await this.services
                 .getSavedSqlService()
-                .getSchedulers(req.user!, savedSqlUuid),
+                .getSchedulers(req.user!, projectUuid, savedSqlUuid),
         };
     }
 
@@ -696,7 +696,7 @@ export class SqlRunnerController extends BaseController {
             status: 'ok',
             results: await this.services
                 .getSavedSqlService()
-                .createScheduler(req.user!, savedSqlUuid, body),
+                .createScheduler(req.user!, projectUuid, savedSqlUuid, body),
         };
     }
 }
--- a/packages/backend/src/generated/swagger.json
+++ b/packages/backend/src/generated/swagger.json
@@ -28605,7 +28605,7 @@
     },
     "info": {
         "title": "Lightdash API",
-        "version": "0.2648.5",
+        "version": "0.2649.0",
         "description": "Open API documentation for all public Lightdash API endpoints. # Authentication Before you get started, you might need to create a Personal Access Token to authenticate via the API. You can create a token by following this guide: https://docs.lightdash.com/references/personal_tokens\n",
         "license": {
             "name": "MIT"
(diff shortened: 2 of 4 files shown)


packages/backend/src/scheduler/SchedulerTask.ts
```diff
@@ -2990,7 +3051,7 @@ export default class SchedulerTask {
                     groupId: notification.jobGroup,
                     type: 'gsheets',
                     format,
-                    resourceType: savedChartUuid ? 'chart' : 'dashboard',
+                    resourceType: dashboardUuid ? 'dashboard' : 'chart',
```
magnew: We don't want to distinguish between charts and sql charts here? That's probably fine. We could probably use this if we wanted to know the difference `getSchedulerResourceTypeAndId`

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -4854,6 +4854,51 @@ export class AsyncQueryService extends ProjectService {
...
+        }
+
+        const resultsStream = await this.getResultsStorageClientForContext(
+            queryHistory.context,
+        ).getDownloadStream(queryHistory.resultsFileName);
+
+        const rows: Record<string, unknown>[] = [];
+        await streamJsonlData<void>({
+            readStream: resultsStream,
+            onRow: (rawRow) => {
```
magnew, line 4918: This is loading everything into memory, right? I guess we have to in this case since we don't control the gsheets client -- did we fix it so that there is always a row-limit in SQL? 😅

packages/backend/src/scheduler/SchedulerTask.ts
almeidabbm: Fixed in the latest commits using `getSchedulerResourceTypeAndId`

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
almeidabbm: This is true, seems like all chart syncs do the same exact thing 🤔 could possibly be a bigger refactor?

- `executeSavedChartQueryAndGetResults`
- `executeMetricQueryAndGetResults`
- `executeDashboardChartQueryAndGetResults`
and now
- `executeSqlChartQueryAndGetResults`

PR #21320, opened 2026-03-23
feat: add large chart performance optimizations with canvas renderer and label hiding

rephus:
Screenshot from 2026-03-23 16-05-31

- [x] Behind a feature flag
- [x] Respect themes and colors
- [x] cOmpatible with dark mode
- [x] tested with deliveries
- [x] interactivity (drill intoy, underlying data)
- [ ] flag is not used in minimal, so this shouldn't affect deliveries (do we want to?) 
- [ ] Can't apply theme when swtiching from dark/light (pending) 

### Description:

Adds a new feature flag `LargeChartPerformance` to optimize chart rendering performance when dealing with large datasets. When enabled, the system automatically switches from SVG to canvas renderer when the total data points exceed 500, resolves CSS variables for canvas compatibility, and enables label overlap hiding to reduce visual clutter.

The optimization includes:
- Automatic renderer switching based on data point threshold (series count × dataset rows)
- CSS variable resolution for canvas renderer compatibility since canvas cannot access DOM-based CSS variables
- Label overlap prevention to improve readability in dense charts

test-frontend test-backend

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
@@ -2467,6 +2467,7 @@ const useEchartsCartesianConfig = (
                                 computedColor,
                             ),
                         },
+                        labelLayout: { hideOverlap: true },
                     }),
                     // Apply reference line styling with readable colors
                     ...(serie.markLine && {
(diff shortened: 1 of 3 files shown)


packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts
```diff
@@ -2467,6 +2467,7 @@ const useEchartsCartesianConfig = (
                                 computedColor,
                             ),
                         },
+                        labelLayout: { hideOverlap: true },
```
rephus, line 2470: bugfix, unrelated to canvas model. Aapply overlap to existing pivot series

Issue #21142, opened 2026-03-15
PDF format for scheduled deliveries via page.pdf()


ota2000:
## Summary

Add a **PDF format option** for scheduled deliveries that renders the dashboard page directly as a native PDF using Playwright's `page.pdf()`.

## Motivation

Users want to receive dashboard reports as PDF attachments in their scheduled deliveries. PDFs are easier to archive, print, and share compared to inline images.

## Implementation

Leverages the existing headless browser infrastructure — adds `page.pdf()` alongside the existing `screenshot()` path. ~240 lines changed across 8 files.

- **common**: Add `SchedulerFormat.PDF` enum value
- **backend/UnfurlService**: Add `outputFormat='pdf'` option to `saveScreenshot`, using `page.pdf()` with `@page` CSS for single-page output
- **backend/SchedulerTask**: Add PDF case to format dispatch, treat PDF like IMAGE for notification paths
- **frontend**: Add `.pdf` option to scheduler format selector, hide formatting options for PDF

## PDF output characteristics

- Text (BigNumbers, tables, headings) is **selectable and searchable**
- Charts are rendered at **browser quality** (same as dashboard)
- Links are **clickable**
- Layout **matches the dashboard exactly**
- **Single page output** (no page breaks) via `@page` CSS
- Emoji and CJK text render correctly

## Follow-ups

- [ ] Dedicated email template for PDF delivery (currently reuses IMAGE template)
- [ ] Consider deprecating "Also include image as PDF attachment" checkbox on IMAGE format in favor of the new PDF format

PR #21438, opened 2026-03-26
feat: improve PDF quality in scheduled deliveries using native browser rendering

rephus:
Closes: https://github.com/lightdash/lightdash/issues/21142

## Summary
- Replaces the old `createImagePdf` method (which embedded a PNG screenshot into a PDF via pdf-lib) with native browser PDF rendering via Playwright's `page.pdf()`
- PDFs now have selectable text, clickable links, and vector graphics instead of a flat rasterized image
- Simplifies the approach from #21143 by applying the improved PDF rendering to the existing "image + PDF" delivery flow rather than adding a new `SchedulerFormat.PDF` option

Based on work by @OTA2000 in #21143 — thank you for the contribution! 🙏

## Test plan
- [x] Create a scheduled delivery with "Image" format and "Also include image as PDF attachment" checked
- [x] Verify the PDF attachment has selectable text and clickable links
- [x] Verify the image preview still works correctly in Slack/email

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -768,13 +768,7 @@ export class UnfurlService extends BaseService {
         sendNowSchedulerParameters?: ParametersValuesMap | undefined;
         outputFormat?: 'image' | 'pdf';
         withPdf?: boolean;
-    }): Promise<
-        | {
-              imageBuffer?: Buffer;
-              pdfBuffer?: Buffer;
-          }
-        | undefined
-    > {
+    }): Promise<{ imageBuffer?: Buffer; pdfBuffer?: Buffer } | undefined> {
         this.logger.info(
             `with tiles ${JSON.stringify(chartTileUuids)} and ${JSON.stringify(
                 sqlChartTileUuids,


packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -1117,27 +1242,92 @@ export class UnfurlService extends BaseService {
...
-                    // Full page screenshot for charts
-                    const imageBuffer = await page.screenshot({
-                        path,
-                        fullPage: true,
-                        animations: 'disabled',
-                    });
+                    // Also generate PDF in the same browser session
+                    if (withPdf) {
+                        const pdfBuffer = await generatePdf();
+                        return { imageBuffer, pdfBuffer };
```
rephus: change output, we can now return a pdf buffer on the same browser call

packages/backend/src/services/UnfurlService/UnfurlService.ts
```diff
@@ -1117,29 +1246,92 @@ export class UnfurlService extends BaseService {
...
-                    const imageBuffer = await page.screenshot({
-                        path,
-                        fullPage: true,
-                        animations: 'disabled',
-                    });
+                    // Also generate PDF in the same browser session
+                    const pdfBuffer = withPdf ? await generatePdf() : undefined;
 
-                    return imageBuffer;
+                    return { imageBuffer, pdfBuffer };
```
rephus, line 1328: now we return 2 buffers from a single chrome instance.

PR #21453, opened 2026-03-26
feat: add configurable DuckDB memory limit for pre-aggregates

tatianainama:
### Description:

The shared DuckDB instance used for pre-aggregate queries has no memory cap. Under concurrent load, multiple queries scanning parquet files can grow memory past the container limit, OOMKilling the pod and losing all inflight queries:skull:

This pr adds `PRE_AGGREGATE_DUCKDB_QUERY_MEMORY_LIMIT` env var to bound DuckDB's memory. When the limit is hit, individual queries error and fall back to the warehouse instead of crashing the entire pod. Only affects the pre-aggregate query path, materializations use isolated instances with their own limits for now

Sizing guide:

```
memory_limit >= (per_query_memory × max_concurrent_queries) + overhead
```

- `per_query_memory`: depends on the parquet size and query complexity (~120-140MB for a HASH_GROUP_BY over 4.3M rows)
- `max_concurrent_queries`: controlled by `NATS_WORKER_CONCURRENCY`
- `overhead`: bootstrap, parquet metadata cache, etc

### DIFF
--- a/packages/warehouses/src/warehouseClients/DuckdbWarehouseClient.ts
+++ b/packages/warehouses/src/warehouseClients/DuckdbWarehouseClient.ts
@@ -86,7 +86,7 @@ export type DuckdbWarehouseClientArgs = {
     databasePath?: string;
     s3Config?: DuckdbS3SessionConfig;
     resourceLimits?: DuckdbResourceLimits;
-    bufferPoolSize?: string; // e.g. '256MB' — controls DuckDB's buffer_pool_size for parquet/HTTP caching
+    memoryLimit?: string; // e.g. '256MB', '1GB' — sets DuckDB's memory_limit for the shared instance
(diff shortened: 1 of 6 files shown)


packages/warehouses/src/warehouseClients/DuckdbWarehouseClient.ts
```diff
@@ -483,7 +483,7 @@ export class DuckdbWarehouseClient extends WarehouseBaseClient<CreatePostgresCre
 
                 if (this.bufferPoolSize) {
                     await db.run(
-                        `SET buffer_pool_size = '${this.bufferPoolSize}';`,
```
tatianainama: this was a dormant bug. we didn't have this set anywhere and this prop is not valid for the current api :skull:

PR #21505, opened 2026-03-30
fix: use dbt run_results to track compiled models for deferred preview builds

notgiorgi:
## Summary

- read `run_results.json` after `dbt compile` when `--use-dbt-list=false`
- use the fresh `unique_id` list from the current invocation instead of relying on stale `manifest.compiled` flags
- avoid misclassifying deferred models as selected models during preview compilation, which can leave BigQuery queries pointing at missing PR tables
- fall back gracefully when `run_results.json` is unavailable

**How I validated the fix**

I reproduced the `--use-dbt-list=false` artifact issue using the local dbt example

The problem with the old behavior is that after `dbt compile`, Lightdash had no explicit `compiledModelIds`, so it fell back to `manifest.compiled`. Those flags can be stale from previous runs, which means a slim compile can incorrectly look like “many models were compiled in this run.”

To validate the fix, I checked both the “nothing modified” and “one model modified” cases using `run_results.json`, which is written fresh by dbt for each compile.

**Validation steps**

Create the state manifest:

```bash
cd examples/full-jaffle-shop-demo/dbt

dbt seed --profiles-dir ../profiles --target jaffle --target-path target-prod
dbt run --profiles-dir ../profiles --target jaffle --target-path target-prod
dbt compile --profiles-dir ../profiles --target jaffle --target-path target-prod
```

Verify the “nothing modified” case:

```bash
export SEED_SCHEMA=jaffle_pr

dbt compile \
  --profiles-dir ../profiles \
  --target jaffle \
  --target-path target-pr \
  --select 'state:modified+' \
  --defer \
  --state ./target-prod

jq '.results | map(.unique_id)' target-pr/run_results.json
```

Result:

- `run_results.json` returned `[]`
- this confirms dbt selected nothing in that invocation
- under the old behavior, Lightdash would still have fallen back to stale `manifest.compiled`

Verify the “one model modified” case:

```bash
printf '\n-- test change %s\n' "$(date +%s)" >> models/orders.sql

dbt compile \
  --profiles-dir ../profiles \
  --target jaffle \
  --target-path target-pr \
  --select 'state:modified+' \
  --defer \
  --state ./target-prod

jq '[.results[] | .unique_id | select(startswith("model."))]' target-pr/run_results.json
```

Result:

- dbt wrote a fresh `run_results.json`
- filtering to `model.` ids returned only:

```json
[
  "model.jaffle_shop.orders"
]
```

This is the expected behavior and demonstrates the fix:

- before: `useDbtList=false` meant `compiledModelIds` was `undefined`, so selection was inferred from stale `manifest.compiled`
- after: `useDbtList=false` reads `run_results.json`, so `compiledModelIds` reflects the models actually compiled in that invocation

That makes `originallySelectedModelIds` accurate again, which allows defer patching to correctly rewrite non-selected models to the state/prod relation.

### DIFF
--- a/packages/cli/src/handlers/dbt/compile.ts
+++ b/packages/cli/src/handlers/dbt/compile.ts
@@ -7,7 +7,9 @@ import {
     SupportedDbtVersions,
 } from '@lightdash/common';
 import execa from 'execa';
+import { promises as fs } from 'fs';
 import { xor } from 'lodash';
+import * as path from 'path';
 import { loadManifest, LoadManifestArgs } from '../../dbt/manifest';
 import GlobalState from '../../globalState';
 import { getDbtVersion } from './getDbtVersion';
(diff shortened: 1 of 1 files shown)


notgiorgi:
* **#21505** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #21543, opened 2026-03-31
feat: wip data app domain set-up

Spissable:
### Description:
🚧 The goal is to embed the data apps from a different domain for proper isolation. Right now we're relying on an opaque origin.


Spissable:
Replaced with a new PR: https://github.com/lightdash/lightdash/commit/fd08c48bfc37d0c539ca7a2fc5c3d2aa27d69a77

PR #21570, opened 2026-03-31
fix: pre aggregate deployment from cli

IrakliJani:
This PR fixes pre-aggregate explore generation from CLI


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -2150,11 +2156,17 @@ export class ProjectService extends BaseService {
...
+        const exploresWithPreAggregates = enhanceExploresForPreAggregates({
+            explores,
+            enabled: this.lightdashConfig.preAggregates.enabled,
+        });
+
         await this.saveExploresToCacheAndIndexCatalog(
             user.userUuid,
             projectUuid,
-            explores,
+            exploresWithPreAggregates,
```
IrakliJani, line 2162: real fix, the rest is reusability

PR #21588, opened 2026-03-31
fix: period-over-period previous period metric not inheriting format

### LINKED ISSUE #21207: Period over period "previous period" metric is not inheriting metric format after chart is saved
### Description

Period over period "previous period" metric is not inheriting metric format after chart is saved

[bug: previous period metric format is lost after saving chart - Watch Video](https://www.loom.com/share/7b417f7702aa401eb0929cbca3461303)

### version

_No response_

### Cloud or self-hosting

None

jpetey75:
## Summary

- `buildPopAdditionalMetric` was copying legacy format fields (`round`, `compact`, `format`) but not `formatOptions` — the modern custom format field
- The "previous period" metric was created without its format, so it appeared correctly before saving but lost formatting after save/reload

## Fix

Added `formatOptions` to the `Pick<Metric, ...>` type and copied it to the constructed `additionalMetric` in `periodOverPeriodComparison.ts`.

Fixes #21207

## Test plan

- [ ] Open a chart with a metric that has custom number formatting (e.g. currency, percentage with decimals)
- [ ] Add a period-over-period comparison
- [ ] Verify the "previous period" metric shows the same format as the base metric
- [ ] Save the chart and reload — verify the format is still applied to the previous period metric


packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -2562,12 +2549,30 @@ export class AsyncQueryService extends ProjectService {
             pivotDimensions: metricQuery.pivotDimensions,
         });
 
+        // Build a map from PoP metric IDs to their base metric IDs
+        // so we can inherit format overrides from the base metric
+        const popToBaseMetricId: Record<string, string> = {};
```
magnew: This probably shouldn't be here in the AsyncQueryService. It's pretty feature-specific. By the time we are here, we should basically have normalised the information we need to build and run a query without specific logic for Pop. 

The fact that this is duplicated across services makes that more apparent.

magnew (commented):
This spreads Pop-specific logic into the query service, which we should avoid if possible. 

I think it could be changed to extract the popToBaseMetricId + override resolution into a shared utility in common (or something similar), called once before entering the query service. This also fixes the existing duplication between AsyncQueryService and ProjectService

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
jpetey75: Good call — extracted into `getMetricOverridesWithPopInheritance` in `common/src/types/periodOverPeriodComparison.ts`. Both services and `useColumns` now call it once and read from the resolved map, removing the duplication. c4da0b1dc7

jpetey75:
@magnew how's it looking with the update from last week?

magnew (approved):
Nice 🚀

jpetey75:
tested again just now and looks good still after the last update, both YAML formatting and frontend "edit format" still work, and they are both applied to the period comparison successfully

PR #21592, opened 2026-03-31
feat(cli): support spark dbt adapter in deploy command

jessicabowden:
## Summary

- Adds `SupportedDbtAdapter.SPARK` so the CLI and compiler can handle dbt projects using the `spark` adapter
- Auto-detects Spark adapter in `deploy --create` and skips warehouse credential extraction, since Spark `session`-based projects have no warehouse server to connect to
- Users configure their actual query engine (e.g., Athena) in the Lightdash UI after project creation
- Maps Spark to Databricks SQL behavior throughout (same dialect)

## Motivation

dbt projects using the `spark` adapter (common with PySpark on EMR/Glue writing Iceberg tables) were rejected by the CLI with "dbt adapter not supported". These projects typically query data through a separate engine like Athena, so Lightdash doesn't need a Spark warehouse client — it just needs to accept the adapter type for manifest parsing and SQL compilation.

## Test plan

- [x] `lightdash deploy --create` works with a `type: spark, method: session` dbt project
- [x] All packages typecheck cleanly (`common`, `backend`, `frontend`, `warehouses`, `cli`)
- [x] `pnpm -F common test` passes (57 suites, 1637 tests)
- [x] `pnpm -F backend test:dev:nowatch` relevant tests pass (`DbtMetadataApiClient`, `MetricQueryBuilder`)
- [x] Verify project is accessible in Lightdash UI after deploy
- [ ] Verify Athena warehouse connection can be configured in project settings

### DIFF
--- a/packages/backend/src/dbt/DbtMetadataApiClient.ts
+++ b/packages/backend/src/dbt/DbtMetadataApiClient.ts
@@ -21,6 +21,7 @@ const quoteChars: Record<SupportedDbtAdapter, string> = {
     trino: `"`,
     clickhouse: `"`,
     athena: `"`,
+    spark: '`',
 };
 
 const PAGE_SIZE = 500;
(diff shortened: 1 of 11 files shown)


charliedowler:
Hi @jessicabowden — thank you so much for this contribution! We really appreciate you taking the time to submit it.

We've rebased and landed it as #23253. Closing this one in favor of that. Thanks again!

PR #21611, opened 2026-04-01
fix: mock worker_threads in ProjectService test

almeidabbm:
## Summary

- The `should get results with 501 rows` test in `ProjectService.test.ts` fails when run standalone (`pnpm -F backend test`) because it spawns a `Worker` thread that requires the compiled `dist/services/ProjectService/formatRows.js` artifact
- This file only exists after a full `pnpm build`, which CI runs but local dev doesn't
- Mocks `worker_threads` so `formatRows` runs synchronously in the main thread, removing the build dependency

## Test plan

- [x] Confirmed test fails without the fix (`Cannot find module .../dist/services/ProjectService/formatRows.js`)
- [x] Confirmed all 35 tests pass with the fix

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.test.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.test.ts
@@ -81,6 +81,41 @@ import {
     virtualExplore,
 } from './ProjectService.mock';
 
+// Mock worker_threads so the >500 rows test doesn't need a compiled
+// dist/services/ProjectService/formatRows.js artifact. In production,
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#21611** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #21637, opened 2026-04-01
fix: apply project timezone to all relative date filters

almeidabbm:
## Summary

- Add `.tz(timezone)` to `IN_THE_PAST`, `NOT_IN_THE_PAST`, and `IN_THE_NEXT` filter operators
- `IN_THE_CURRENT` and `NOT_IN_THE_CURRENT` already used `.tz(timezone)` — this makes all 5 relative operators consistent
- Fix DST bug in completed-path date arithmetic: `.subtract()`/`.add()` on `completedDate`/`fromDate` were missing `.tz(timezone)`, causing 1-hour errors around DST transitions (e.g. spring-forward producing a 24h range instead of 23h for a calendar day)
- Add timezone-specific unit tests for completed day filters with `America/New_York`
- Add DST spring-forward regression tests for completed `IN_THE_PAST` and `IN_THE_NEXT` paths
- Add non-completed path timezone tests for `IN_THE_PAST` and `IN_THE_NEXT`

**No feature flag gating** — when timezone is `'UTC'` (the default), `.tz('UTC')` produces identical results to the previous code (no `.tz()` call) in a UTC process. The behavioral change only activates when a non-UTC project timezone is explicitly configured, which means the user opted in via the existing project timezone setting (GLITCH-235).

Closes GLITCH-286

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.mock.ts
+++ b/packages/common/src/compiler/filtersCompiler.mock.ts
@@ -672,6 +672,39 @@ export const filterInTheNextNonCompletedDayTimezoneMocks = [
     ],
 ];
 
+// DST spring forward: March 8, 2020 at 2:00 AM EST → 3:00 AM EDT (America/New_York)
+// System time: 09 Mar 2020 05:00:00 GMT (= 01:00 EDT, post-DST)
+// IN_THE_PAST 1 day (non-completed) crossing DST:
+// UTC:            from = subtract 1 day = 2020-03-08 05:00:00, until = 2020-03-09 05:00:00 (24h)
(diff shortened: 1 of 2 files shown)


packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -320,14 +319,18 @@ export const renderDateFilterSql = (
                     : '';
 
             if (completed) {
-                const completedDate = moment(
-                    getMomentDateWithCustomStartOfWeek(effectiveStartOfWeek)
-                        .startOf(unitOfTime)
-                        .format(unitOfTimeFormat[unitOfTime]),
```
almeidabbm, line 326: Not needed since now we're doing `startOf(unitOfTime)` to handle the truncation

Issue #20840, opened 2026-03-04
Add 'project' as a group that can be given permissions to a space


magnew:
### Description

Add a special group of all project members that can be granted access to a space. This would enable a restricted space to grant only 'view' access to project members. 

### What problem does this solve?

This would make it possible to have broad (but limited) permissions to a space while also granting more access to some users. A specific customer scenario looks like:

- Everyone can see the Space and its content
- Only some users or groups can edit the contents

PR #21649, opened 2026-04-01
feat: make project users a group for permissions

magnew:
Closes: #20840

## Summary

This is an alternative implementation to https://github.com/lightdash/lightdash/pull/21645

Adds an "All project members" option to the space share modal, allowing admins to grant every project member a role on a restricted-access space without adding them individually.

## How it works

A sentinel group record (`00000000-...`) is created via migration in the `groups` table. When a user selects "All project members" in the share modal, a row is inserted into `space_group_access` referencing this sentinel UUID — same flow as adding any other group.

The permission system's `getDirectSpaceAccess` query has two new UNION legs that recognize the sentinel and expand it to actual users:
- One joins through `project_memberships` (explicit project members)
- One joins through `organization_memberships` for non-`member` roles (org members who implicitly have project access)

This means the sentinel behaves like a dynamic group whose membership is always in sync with who can access the project — no background sync jobs, no stale membership.

The existing group UNION leg excludes the sentinel (`WHERE group_uuid != sentinel`) so it doesn't try to expand it via `group_memberships` (which would return zero rows).

On the frontend, the sentinel is filtered out of the regular org groups list to avoid duplicates, and instead shown as a synthetic "All project members" entry at the top of the Groups section in the dropdown.

## Use case

Restricted-access spaces (`inherit_parent_permissions = false`) currently require adding users or groups one by one. This makes it possible to create a space hierarchy like:

- **Top-level space** (inherits from project — everyone sees it)
  - **Child space** (restricted — only specific people)
    - **Grandchild space** (restricted, but all project members granted viewer access)

## Test plan

- [ ] Open a restricted space's share modal — "All project members" appears in the Groups dropdown
- [ ] Add it — all org/project members gain access at the selected role
- [ ] Remove it — access is revoked
- [ ] Verify it doesn't appear as a duplicate (not shown twice in the dropdown)
- [ ] Verify regular groups still work as before
- [ ] Verify spaces without it are unaffected (no perf impact — the UNION legs return zero rows when the sentinel isn't in `space_group_access`)

### DIFF
--- a/packages/backend/src/database/migrations/20260330133142_add_project_members_sentinel_group.ts
+++ b/packages/backend/src/database/migrations/20260330133142_add_project_members_sentinel_group.ts
@@ -1,0 +1,31 @@
+import { Knex } from 'knex';
+
+const PROJECT_MEMBERS_GROUP_UUID = '00000000-0000-0000-0000-000000000000';
+
+export async function up(knex: Knex): Promise<void> {
+    // Insert a sentinel group record so the FK on space_group_access is
+    // satisfied. The sentinel has no members in group_memberships — the
+    // permission system expands it dynamically to all project members.
+    const [firstOrg] = await knex('organizations')
+        .select('organization_id')
+        .limit(1);
+    if (!firstOrg) return;
+
+    await knex('groups')
+        .insert({
+            group_uuid: PROJECT_MEMBERS_GROUP_UUID,
+            organization_id: firstOrg.organization_id,
+            name: 'All project members',
+        })
+        .onConflict('group_uuid')
+        .ignore();
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex('space_group_access')
+        .where('group_uuid', PROJECT_MEMBERS_GROUP_UUID)
+        .delete();
+    await knex('groups')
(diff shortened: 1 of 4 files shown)


packages/backend/src/database/migrations/20260330133142_add_project_members_sentinel_group.ts
```diff
@@ -0,0 +1,31 @@
...
+const PROJECT_MEMBERS_GROUP_UUID = '00000000-0000-0000-0000-000000000000';
+
+export async function up(knex: Knex): Promise<void> {
+    // Insert a sentinel group record so the FK on space_group_access is
+    // satisfied. The sentinel has no members in group_memberships — the
+    // permission system expands it dynamically to all project members.
+    const [firstOrg] = await knex('organizations')
+        .select('organization_id')
+        .limit(1);
+    if (!firstOrg) return;
```
almeidabbm, line 12: This will not work for multi tenant instances 🤔

magnew: Yea

PR #21732, opened 2026-04-03
fix: allow null meta in dbt manifest schemas for Fusion

torimwhaley:
## Summary

- dbt Fusion (preview.164+) outputs `config.meta: null` instead of `config.meta: {}` in its manifest, and reports `dbt_schema_version` as `v12` (older Fusion previews reported `v20`)
- Our `lightdashV12.json` schema requires `meta` to be an object, causing validation to fail with `Field at "/config/meta" must be object` for every model without explicit Lightdash meta
- Updates all lightdash validation schemas (V7–V12) to accept `null` for `meta` fields using `oneOf`, matching the existing V20 schema behavior
- Also adds `isDbtFusion` flag to `DbtVersion` type for future use

**Root cause:** Fusion preview.164 changed `dbt_schema_version` from `v20.json` back to `v12.json` in the manifest metadata. Our V20 schema already allowed null meta, but V12 did not — so the CLI started rejecting Fusion manifests.

**Reproduced locally** with dbt-fusion 2.0.0-preview.164 and confirmed the fix resolves the validation error.

## Test plan

- [x] `pnpm -F common test` — all 60 suites, 1680 tests pass
- [x] `pnpm -F @lightdash/cli lint` — clean
- [x] Local repro: `dbt compile` with Fusion preview.164 → `lightdash compile --skip-dbt-compile --skip-warehouse-catalog` → SUCCESS
- [ ] CI passes

### DIFF
--- a/packages/cli/src/handlers/dbt/getDbtVersion.ts
+++ b/packages/cli/src/handlers/dbt/getDbtVersion.ts
@@ -45,6 +45,9 @@ const getDbtCLIVersion = async () => {
 const isDbtCloudCLI = (version: string): boolean =>
     version.match(DBT_CLOUD_CLI_REGEX) !== null;
 
+const isDbtFusion = (version: string): boolean =>
+    version.match(DBT_FUSION_VERSION_REGEX) !== null;
+
 const getSupportedDbtVersionOption = (
     version: string,
 ): DbtVersionOption | null => {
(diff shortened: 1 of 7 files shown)


packages/cli/src/handlers/dbt/getDbtVersion.ts
IrakliJani, line 16:
```
$ dbt --version
dbt-fusion 2.0.0-preview.164
```

looks good!

IrakliJani (approved):
tested, this is perfect!

```
$ dbt --version
dbt-fusion 2.0.0-preview.164
```

before the fix:

Image

after the fix:

Image

PR #21740, opened 2026-04-03
fix: move app generation pipeline to Graphile Worker for deploy resilience

Spissable:
Closes: https://linear.app/lightdash/issue/GLITCH-314/make-the-app-builder-backend-recoverable

### Description:
The app generation pipeline (up to 55 minutes) previously ran as a fire-and-forget promise. If the backend pod died during a deployment, the results were lost. Now the pipeline runs inside a Graphile Worker job that persists in PostgreSQL and survives deploys.

The main tricky part is the stale detection and resumption:
Basically, we're running now a cronjob every 2 minutes and check for any app versions that haven't finished (not ready/error) and that haven't received any status update in 5 minutes. These we then consider stale, so we unlock the jobs so it can be picked up again, which will gracefully resume it and finish the job.

### DIFF
--- a/packages/backend/src/ee/scheduler/SchedulerWorker.ts
+++ b/packages/backend/src/ee/scheduler/SchedulerWorker.ts
@@ -12,15 +12,18 @@ import { SchedulerTaskArguments } from '../../scheduler/SchedulerTask';
 import { SchedulerWorker } from '../../scheduler/SchedulerWorker';
 import { TypedEETaskList } from '../../scheduler/types';
 import { AiAgentService } from '../services/AiAgentService/AiAgentService';
+import { AppGenerateService } from '../services/AppGenerateService/AppGenerateService';
 import type { EmbedService } from '../services/EmbedService/EmbedService';
 import { ManagedAgentService } from '../services/ManagedAgentService/ManagedAgentService';
 
-const AI_AGENT_EVAL_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes in milliseconds
+const AI_AGENT_EVAL_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes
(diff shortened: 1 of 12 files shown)


packages/backend/src/ee/services/AppGenerateService/AppGenerateService.ts
```diff
@@ -799,13 +842,65 @@ export class AppGenerateService extends BaseService {
         return durationMs;
     }
 
-    private async runNewAppPipeline(
+    /**
+     * Ordered pipeline stages. Used to skip completed stages on retry.
+     */
+    private static readonly STAGE_ORDER: AppVersionStatus[] = [
```
magnew: Use the shared type? This list is repeated a couple times.

packages/backend/src/ee/scheduler/SchedulerWorker.ts
```diff
@@ -164,6 +184,35 @@ export class CommercialSchedulerWorker extends SchedulerWorker {
...
+                        await this.appGenerateService.markError(
+                            payload.appUuid,
+                            payload.version,
+                            e,
+                            'Build timed out. Please try again.',
+                        );
+                    },
+                );
+            },
+            [EE_SCHEDULER_TASKS.SWEEP_STALE_APP_LOCKS]: async () => {
```
magnew, line 213: Do we run EE schedulers on all deployments? I can't remember.

magnew (approved):
Nice approach. I the graphite comments might have some Merritt. Approving and I will test it locally.

packages/backend/src/ee/scheduler/SchedulerWorker.ts
Spissable: Affirm, moving it.

Spissable: Update on this: not an issue actually, as our cloud deployments include this. So EE is fine.

PR #21792, opened 2026-04-07
feat: add formula test harness

owlas:
Closes:

### DIFF
--- a/packages/formula-tests/package.json
+++ b/packages/formula-tests/package.json
@@ -11,12 +11,12 @@
   },
   "dependencies": {
     "@lightdash/formula": "workspace:*",
-    "duckdb": "^1.1.0",
-    "pg": "^8.11.0",
-    "tsx": "^4.7.0"
+    "duckdb": "1.4.2",
+    "pg": "8.13.1",
+    "tsx": "4.19.2"
   },
   "devDependencies": {
-    "@types/pg": "^8.10.0",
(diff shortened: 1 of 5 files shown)


packages/formula/src/codegen/base.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        // Dispatch to specific function handlers
+        const handler = this.getFunctionHandler(name);
+        if (handler) {
+            return handler(args, node);
+        }
+
+        // SECURITY: Unknown functions must be rejected — never pass through to SQL.
+        // Passing through would allow users to call arbitrary warehouse functions
+        // (e.g. pg_sleep, read_csv_auto, dblink) bypassing the semantic layer.
```
tatianainama, line 100: I don't think this should be handled here, but in the layer before calling the warehouse when sanitizing the sql, no need to repeat here

packages/formula/src/codegen/base.ts
tatianainama, line 1: would it worth looking into reusing the warehouse sql builder to reuse what we currently do for handling these?

packages/formula/src/codegen/base.ts
owlas: yeah the warning about warehouse-specific stuff is a bit misleading - this throws if a function name is not registered - is that what you meant?

PR #21800, opened 2026-04-07
fix: format relative date filter boundaries in project timezone

almeidabbm:
## Summary

- Fix relative date filter boundaries (`inThePast`, `inTheCurrent`, `inTheNext`) for positive-offset timezones on date/day dimensions
- The boundary date was formatted in UTC instead of the project timezone, causing a one-day shift for timezones like Asia/Tokyo (+9)
- Add `formatBoundary` helper in `renderDateFilterSql` that formats in the project timezone for date-level comparisons (YYYY-MM-DD) and in UTC for timestamp-level comparisons

## Problem

When `inThePast 1 completed day` targets a day dimension with Asia/Tokyo (+9):
1. Boundary computed correctly: midnight Apr 4 Tokyo
2. Converted to UTC: `.utc()` → Apr 3 15:00 UTC  
3. `formatDate` strips time → `'2020-04-03'` (wrong — should be `'2020-04-04'`)

The UTC calendar date is one day behind the local date for positive offsets. Negative offsets (Americas, Pacific) are unaffected because midnight local is still the same UTC calendar day.

## Fix

`formatBoundary` formats the moment directly in the project timezone for date-level formatting:
```typescript
if (dateFormatter === formatDate) {
    return m.format('YYYY-MM-DD'); // formats in current tz (project TZ)
}
return dateFormatter(m.utc().toDate()); // timestamp path unchanged
```

Closes GLITCH-323

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.ts
+++ b/packages/common/src/compiler/filtersCompiler.ts
@@ -287,7 +287,7 @@ export const renderDateFilterSql = (
         if (dateFormatter === formatDate) {
             return m.format('YYYY-MM-DD');
         }
-        return dateFormatter(m.utc().toDate());
+        return dateFormatter(m.clone().utc().toDate());
     };
 
     const castValue = (value: string): string => {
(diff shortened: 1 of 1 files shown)


almeidabbm:
Folded into #21638 — the boundary fix only manifests with timezone-aware DATE_TRUNC (NTZ output), so they need to ship together.

PR #21809, opened 2026-04-07
feat: add dbt version detection and display installed versions in UI

IrakliJani:
### Description:

Added functionality to detect and display installed dbt patch versions in the project connection form. The system now automatically discovers which dbt versions are available in the environment by scanning PATH for dbt executables and reading their corresponding virtual environment metadata.

Key changes include:

- Created `getDbtExecName()` function to map dbt versions to executable names and extracted existing switch logic from `DbtCliClient`
- Added `getInstalledDbtVersions()` function that detects installed dbt versions by finding virtual environment paths and reading package metadata from site-packages directories
- Implemented new API endpoint `/api/v1/dbt/installed-versions` that returns detected dbt versions with caching for performance
- Enhanced the dbt version selector in the frontend to show actual installed patch versions (e.g., "v1.11 (running 1.11.2)") alongside the base version numbers
- Added React hook `useInstalledDbtVersions` to fetch and cache the installed version data

The version detection works by resolving symlinks in PATH to find the actual dbt installation directories, then parsing the `dbt_core-*.dist-info` metadata to extract precise version numbers. This provides users with better visibility into which dbt versions are actually available in their environment.

### DIFF
--- a/packages/backend/src/routers/apiV1Router.ts
+++ b/packages/backend/src/routers/apiV1Router.ts
@@ -20,6 +20,7 @@ import {
     getDatabricksOidcEndpointsFromHost,
     getDatabricksStrategyName,
 } from '../controllers/authentication/strategies/databricksStrategy';
+import { getInstalledDbtVersions } from '../dbt/dbtVersions';
 import { AiAgentService } from '../ee/services/AiAgentService/AiAgentService';
 import Logger from '../logging/logger';
 import { UserModel } from '../models/UserModel';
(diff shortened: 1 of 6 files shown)


packages/backend/src/routers/apiV1Router.ts
```diff
@@ -210,6 +211,17 @@ apiV1Router.get('/health', async (req, res, next) => {
         .catch(next);
 });
 
+apiV1Router.get('/dbt/installed-versions', async (req, res, next) => {
```
IrakliJani, line 214: this should not be a a old api router, instead it should be a controller + service :white_frowning_face:

PR #21812, opened 2026-04-07
feat: support dbt seeds as join targets

Spissable:
Closes: https://linear.app/lightdash/issue/PROD-6187/joins-with-dbt-seedssources-silently-fail

### Description:
Seeds are now included in the compilation pipeline so they can be used as join targets from models, eliminating the need for wrapper models that just `SELECT * FROM {{ ref('seed') }}`. Seeds are compiled as tables for join resolution but do not create standalone explores.

Screenshot 2026-04-07 at 16 50 08

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/data/raw_order_statuses.csv
+++ b/examples/full-jaffle-shop-demo/dbt/data/raw_order_statuses.csv
@@ -1,0 +1,6 @@
+status,status_label,status_priority,is_active
+placed,Order Placed,1,true
+shipped,Shipped,2,true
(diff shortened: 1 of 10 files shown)


jesshitchcock:
My test case ended up being pretty complex but I got it working in the end. Note it was a bit annoying to have to manually add the YAML for my seed (lightdash generate only works for models).

jesshitchcock (approved):
My test case ended up being pretty complex but I got it working in the end. Note it was a bit annoying to have to manually add the YAML for my seed (lightdash generate only works for models).

PR #21889, opened 2026-04-09
fix(skills): cli-reference.md references non-existing cli command 'lightdash delete'

schwericke:
cli-reference.md features a full section about 'lightdash delete', a non-existing CLI command. 
- lightdash delete -> error: unknown command 'delete'
- no mentions in the docs https://docs.lightdash.com/references/lightdash-cli 

In my first commit, I updated the section to cover deletions via REST API https://docs.lightdash.com/api-reference/v2/delete-dashboard.

In my second commit, instead I removed the whole section, to keep the skill clean with (existing :P) CLI commands only.


ZeRego:

PR #21900, opened 2026-04-09
feat: thread resolved timezone through formatting paths

almeidabbm:
closes: https://linear.app/lightdash/issue/GLITCH-291/update-result-formatting-to-use-resolved-project-timezone

### Problem

Timezone-aware DATE_TRUNC creates a display mismatch. With project timezone `America/New_York (UTC -4:00)`, a raw timestamp showing `2024-01-15, 02:00 (+00:00)` gets grouped under `2024-01-14` — because it was still January 14th in New York. The user sees two different dates for the same row with no explanation.

### Summary

Threads the resolved project timezone through all formatting paths so timestamps display in the correct timezone. 100% gated behind `EnableTimezoneSupport`.

- **Backend:** resolve display timezone in AsyncQueryService, ProjectService, EmbedService; pass to `formatRows`/`formatRow`
- **Frontend:** add `resolvedTimezone` to VisualizationProvider context; pass to axis labels, tooltips, series labels, stack totals, and pivot display values
- **DATE normalization:** strip the time component from DATE values to midnight UTC in `formatRawValue`, so echarts positioning and pivot keys work correctly

### DIFF
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -23,6 +23,7 @@ import {
     ExecuteAsyncDashboardChartRequestParams,
     Explore,
     ExploreError,
+    FeatureFlags,
     FieldValueSearchResult,
     FilterableDimension,
     ForbiddenError,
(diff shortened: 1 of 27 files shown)


almeidabbm:
Closing this — we're taking a different approach that keeps the existing `convertToUTC` parameter and adds `timezone` as an additive optional parameter instead. This reduces the diff significantly and avoids touching existing callers. New PRs incoming.

PR #21902, opened 2026-04-09
feat: auto-detect CJK language for correct font rendering in screenshots

### LINKED ISSUE #21499: fix: CJK text in screenshots/PDFs renders with wrong font (Chinese instead of Japanese)
## Problem

Screenshots and PDF scheduled deliveries render CJK (Japanese, Chinese, Korean) text using an incorrect font. Specifically, Japanese text appears with Chinese-style glyph shapes (WenQuanYi Zen Hei) instead of the correct Japanese font (Noto Sans CJK JP).

This affects:
- Scheduled email/Slack deliveries (both image and PDF format)
- Dashboard/chart thumbnails (Slack unfurls)

## Root Cause

The headless browser container (`browserless/chromium:v2.38.2`) **does** have CJK fonts installed — Noto Sans CJK JP, SC, TC, KR, IPAGothic, and WenQuanYi Zen Hei. The container also has fontconfig rules (`70-fonts-noto-cjk.conf`) that select the correct Noto variant per language:

- `lang=ja` → Noto Sans CJK JP
- `lang=zh-cn` → Noto Sans CJK SC
- `lang=ko` → Noto Sans CJK KR

**However**, `index.html` hardcodes `lang="en"`, so fontconfig's language-based rules never activate. The actual font resolution path is:

1. `Inter` → not installed in the container
2. Other system fonts in font-family → not installed
3. `sans-serif` → resolves to `DejaVu Sans` (no CJK glyphs)
4. CJK fallback → **WenQuanYi Zen Hei** (Chinese font, wrong glyph shapes for Japanese/Korean)

Verified with fontconfig inside the headless browser container:

```
$ fc-match "sans-serif:lang=ja"  →  WenQuanYi Zen Hei (not Noto Sans CJK JP!)
$ fc-match "sans-serif"          →  DejaVu Sans
$ fc-match "Inter"               →  DejaVu Sans (Inter not installed)
```

## Proposed Solution: Inject `lang` attribute during headless browser rendering

Add a `lang` option to the scheduled delivery / screenshot rendering process. When the headless browser takes a screenshot or generates a PDF, inject the correct `<html lang="...">` attribute via `page.evaluate()` after `page.goto()`.

This allows fontconfig's existing `70-fonts-noto-cjk.conf` rules to automatically select the correct font variant per language — no CSS changes, no Dockerfile changes needed.

### Implementation approach

1. **Add a `locale` (or `lang`) option** to the scheduler configuration (e.g., `SchedulerOptions` type in `packages/common/src/types/scheduler.ts`)
2. **Pass it through** to `UnfurlService.saveScreenshot()`
3. **Inject it** into the page after navigation:

```typescript
// In UnfurlService.ts, after page.goto()
if (lang) {
    await page.evaluate((l) => {
        document.documentElement.lang = l;
    }, lang);
}
```

4. **UI**: Add a language selector to the scheduled delivery settings modal (defaults to `en`)

This is related to #5026 (i18n internationalization support), though this change is scoped specifically to the headless browser rendering pipeline.

### Affected files

- `packages/common/src/types/scheduler.ts` — add `lang` option type
- `packages/backend/src/services/UnfurlService/UnfurlService.ts` — inject `lang` after `page.goto()`
- `packages/backend/src/scheduler/SchedulerTask.ts` — pass `lang` through to UnfurlService
- `packages/frontend/` — scheduled delivery settings UI (if adding explicit selector)

### ~~Approach A: Add CJK font to CSS `font-family` stack~~

Add `'Noto Sans CJK JP'` to `mantineTheme.ts` before `sans-serif`. This was verified locally and does fix Japanese rendering. However, it **hardcodes a single CJK language variant** (JP), which has the same structural bias as the current state — just favoring Japanese instead of Chinese. This approach was rejected in favor of the `lang` injection approach above.

### ~~Approach B: Install fonts in headless browser Dockerfile~~

Not needed — the container already has all required CJK fonts installed. The issue is font *selection*, not font *availability*.

Spissable:
Closes: https://github.com/lightdash/lightdash/issues/21499

### Description:
Chromium on Linux ignores fontconfig for CJK font fallback, rendering all CJK text with Chinese-style glyphs. Instead of requiring a manual project-level language setting, auto-detect the language from page content using Unicode script ranges (hiragana/katakana → Japanese, hangul → Korean, character frequency heuristics → SC/TC Chinese) and set the html lang attribute so CSS :lang() rules select the correct Noto Sans CJK font variant.

### Testing:
Used the following example inputs in a Markdown tile:
```
Japanese:
直接売上の骨格 — 今月の角度分析レポート

Korean:
직접 매출 골격 — 이번 달 각도 분석 보고서

Simplified Chinese:
直接销售的骨架 — 这个月的角度分析报告

Traditional Chinese:
直接銷售的骨架 — 這個月的角度分析報告
```

Each was correctly detected, validated via logs.

#### Example with Japanese:

Before:
japanese-before

After:
japanese-after

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -1251,6 +1251,46 @@ export class UnfurlService extends BaseService {
                         'Screenshot ready indicator found - page is ready',
                     );
 
+                    // Auto-detect CJK language from page content and set
+                    // <html lang="..."> so CSS :lang() rules select the
+                    // correct Noto Sans CJK font variant for screenshots.
+                    const detectedLang = await page.evaluate(() => {
+                        const text = document.body.innerText;
+                        if (/[\u3040-\u309F\u30A0-\u30FF]/.test(text))
+                            return 'ja';
+                        if (/[\uAC00-\uD7AF\u1100-\u11FF]/.test(text))
+                            return 'ko';
+                        if (/[\u4E00-\u9FFF]/.test(text)) {
+                            // Distinguish Simplified vs Traditional Chinese
+                            // by checking for script-specific character variants
+                            const simplified = (
+                                text.match(
+                                    /[\u4EEC\u56FD\u5B66\u5BF9\u8FD9\u8BA9\u8BF4\u4E1C\u7ECF\u5F00]/g,
+                                ) ?? []
+                            ).length;
+                            const traditional = (
+                                text.match(
+                                    /[\u5011\u570B\u5B78\u5C0D\u9019\u8B93\u8AAA\u6771\u7D93\u958B]/g,
+                                ) ?? []
+                            ).length;
+                            return traditional > simplified ? 'zh-TW' : 'zh-CN';
+                        }
+                        return null;
+                    });
+                    if (detectedLang) {
+                        await page.evaluate(
+                            (lang) =>
+                                document.documentElement.setAttribute(
+                                    'lang',
+                                    lang,
+                                ),
+                            detectedLang,
+                        );
+                        this.logger.info(
+                            `Auto-detected CJK language: ${detectedLang}`,
+                        );
+                    }
+
                     const path = `/tmp/${imageId}.png`;
 
                     let finalSelector = selector;
--- a/packages/frontend/src/mantineTheme.ts
+++ b/packages/frontend/src/mantineTheme.ts
@@ -376,6 +376,22 @@ export const getMantineThemeOverride = (
             '.wmde-markdown img': {
                 backgroundColor: 'transparent',
             },
+            // CJK font selection for headless browser screenshots/PDFs.
+            // Chromium on Linux ignores fontconfig, so we use :lang() CSS
+            // selectors to explicitly pick the correct Noto Sans CJK variant.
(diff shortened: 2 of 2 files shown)


magnew (approved):

ota2000:
Thanks for tackling this! The auto-detection approach is clever and works well for single-language dashboards.

A few edge cases I wanted to flag for consideration:

- **Mixed CJK text**: If a dashboard contains both Japanese and Chinese text (e.g. a multi-region report), the first regex match wins and applies to the entire page. Japanese text with even one hiragana character would set `lang="ja"`, rendering Chinese text with Japanese font glyphs.

- **Kanji-only Japanese**: Japanese text without hiragana/katakana (e.g. `売上分析`, `月次報告`) would fall through to the CJK Unified Ideographs check and be detected as Chinese.

- **Short text detection**: The simplified vs. traditional Chinese heuristic relies on 10 characteristic characters each. For short text (a chart title, a single metric label), there may not be enough signal for an accurate distinction.

These might be rare in practice, but worth keeping in mind. A project-level setting (as discussed in #21541) could complement auto-detection as a user override for cases where the heuristic doesn't get it right.

PR #21923, opened 2026-04-10
fix: cross-model type:number referencing sum_distinct metrics

rephus:
Closes: PROD-6869

CleanShot 2026-04-10 at 11 57 56

### Description:

Fixes cross-model type:number metrics that reference sum_distinct metrics. Previously, when a type:number metric referenced a sum_distinct metric from a joined model, the query would fail with "invalid identifier" because the sum_distinct's deduplication CTE metadata was lost during cross-model reference compilation.

The fix identifies non-aggregate metrics that reference sum_distinct/average_distinct metrics and handles them in the deduplication CTE's outer SELECT instead of the regular SELECT. This allows their SQL templates to be recompiled so that metric references resolve to the correct CTE aliases rather than inlined fallback SQL.

Added comprehensive test coverage including:
- Cross-model type:number referencing sum_distinct with and without dimensions
- Same-model type:number referencing both sum_distinct and regular aggregates
- New test fixture `deduped_revenue_per_order` in the jaffle shop demo that references `customers.total_order_amount_deduped` to validate the fix

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/models/orders.yml
+++ b/examples/full-jaffle-shop-demo/dbt/models/orders.yml
@@ -53,6 +53,17 @@ models:
             drivers:
               - total_order_amount
               - total_completed_order_amount
+          # PROD-6869: Cross-model type:number referencing sum_distinct
+          # This metric references customers.total_order_amount_deduped (sum_distinct)
+          # from the joined customers model. Before the fix, this would fail
+          # with "invalid identifier" because the sum_distinct's deduplication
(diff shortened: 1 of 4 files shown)


rephus:
* **#21923** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21923?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #21940, opened 2026-04-10
fix: date filter boundaries no longer depend on server timezone

almeidabbm:
## Summary

DATE dimension filters (`in the current day`, `in the past N days`, etc.) used `moment(date)` to format boundary dates, which formats in the **server's local timezone**. On servers with a non-UTC timezone, `endOf('day')` at 23:59 UTC gets shifted to the next calendar day, producing a 2-day filter range instead of 1.

**Fix:** When timezone-aware `DATE_TRUNC` is off, force boundary computation and formatting to UTC — matching the warehouse's UTC-based `DATE_TRUNC`. When the flag is on, both use the project timezone.

**Before**
CleanShot 2026-04-10 at 16 11 01

**After**
CleanShot 2026-04-10 at 16 09 28

## Context

**Before timezone work:** All filters used `timezone = 'UTC'`. Boundaries computed in UTC, formatted with `formatDate` (server-local time). Worked because servers run in UTC and the timezone was always UTC.

**PR #21637:** Made all relative date operators consistently use `.tz(timezone)`. Once non-UTC project timezones started flowing through, `formatDate` (which formats in server-local time) became a problem for DATE dimensions. The boundary dates could land on different calendar days depending on the server's timezone.

**This PR:** When no explicit boundary formatter is provided, `renderDateFilterSql` now defaults both boundary computation and formatting to UTC. This restores the original UTC-based behavior, makes it server-timezone-independent, and keeps filter boundaries aligned with the warehouse's UTC-based `DATE_TRUNC`. When timezone-aware `DATE_TRUNC` is enabled #21874 , callers pass an explicit formatter and both sides use the project timezone instead.

## Test plan

- [x] Added tests that simulate non-UTC server timezones via `moment.tz.setDefault()` and verify single-day boundaries
- [x] Verified tests fail without the fix (5 failures on positive-offset server TZs)
- [ ] Manual: set project timezone to non-UTC, apply `in the current day` filter on a date dimension, verify SQL shows single-day range

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.test.ts
+++ b/packages/common/src/compiler/filtersCompiler.test.ts
@@ -1,11 +1,14 @@
+import momentTz from 'moment-timezone';
 import moment from 'moment/moment';
 import { SupportedDbtAdapter } from '../types/dbt';
+import { DimensionType } from '../types/field';
 import { FilterOperator, UnitOfTime, type FilterRule } from '../types/filter';
 import { WeekDay } from '../utils/timeFrames';
 import {
     createBoundaryDateFormatter,
     renderBooleanFilterSql,
     renderDateFilterSql,
+    renderFilterRuleSql,
     renderFilterRuleSqlFromField,
     renderNumberFilterSql,
     renderStringFilterSql,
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#21940** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21940?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/21940?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #21948, opened 2026-04-10
feat: migrate filter autocomplete to async query service

charliedowler:
## Summary
- Routes filter autocomplete queries through `AsyncQueryService` instead of the synchronous `ProjectService` path, unifying caching via `query_history` and enabling non-blocking warehouse execution
- Extracts `_getFieldValuesMetricQuery` into a shared utility (`fieldValuesQueryBuilder.ts`) to avoid circular service dependencies between `AsyncQueryService` and `ProjectService`
- Adds `AsyncFilterAutocomplete` feature flag (disabled by default) so the new path can be gradually rolled out per-org

## Changes

**Backend:**
- New `executeAsyncFieldValueSearch` method in `AsyncQueryService` — builds MetricQuery via shared utility, compiles SQL, checks cache, fires async warehouse query
- New `POST /api/v2/projects/:projectUuid/query/field-values` endpoint in `QueryController`
- Shared `getFieldValuesMetricQuery` utility extracted from `ProjectService._getFieldValuesMetricQuery` (existing method now delegates to it)

**Frontend:**
- `useFieldValues` hook branches on `AsyncFilterAutocomplete` feature flag
- New `getFieldValuesAsync` function: submit → poll → adapt response to `FieldValueSearchResult` shape
- Query key includes flag value (`v1`/`v2`) to prevent cache cross-contamination

**Common:**
- New `ExecuteAsyncFieldValueSearchRequestParams` and `ApiExecuteAsyncFieldValueSearchResults` types
- New `AsyncFilterAutocomplete` feature flag enum value

## Context
SPK-375 — first of three planned sync→async migrations (SPK-373 Calculate Total, SPK-374 Calculate Subtotals). The pattern established here will be reused.

## Test plan
- [x] `pnpm -F backend typecheck` passes
- [x] `pnpm -F frontend typecheck` passes
- [x] `pnpm -F backend lint` passes (no new errors)
- [x] `pnpm -F frontend lint` passes
- [x] `pnpm -F backend test:dev:nowatch` — 135 tests pass
- [ ] Manual: enable `async-filter-autocomplete` flag, type in a filter dropdown, verify results load via v2 endpoint
- [ ] Manual: disable flag, verify v1 path still works
- [ ] Manual: verify embed filter autocomplete is unaffected

### DIFF
--- a/packages/backend/src/controllers/v2/QueryController.ts
+++ b/packages/backend/src/controllers/v2/QueryController.ts
@@ -3,6 +3,7 @@ import {
     ApiErrorPayload,
     ApiExecuteAsyncDashboardChartQueryResults,
     ApiExecuteAsyncDashboardSqlChartQueryResults,
+    ApiExecuteAsyncFieldValueSearchResults,
     ApiExecuteAsyncSqlQueryResults,
     ApiGetAsyncQueryResults,
     ApiSuccess,
@@ -21,6 +22,7 @@ import {
     type ApiJobScheduledResponse,
     type ExecuteAsyncDashboardChartRequestParams,
     type ExecuteAsyncDashboardSqlChartRequestParams,
+    type ExecuteAsyncFieldValueSearchRequestParams,
     type ExecuteAsyncMetricQueryRequestParams,
     type ExecuteAsyncSavedChartRequestParams,
     type ExecuteAsyncSqlChartRequestParams,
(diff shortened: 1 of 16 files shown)


charliedowler:
**Re: Graphite security comment** — The permission check is in the service layer (`executeAsyncFieldValueSearch` checks `ability.cannot('view', subject('Project', ...))`), matching the exact same pattern as `executeAsyncMetricQuery` and all other methods in `QueryController`. No controller in this file does controller-level auth — it's all delegated to the service. Adding a redundant `hasUserAccessToProject` call at the controller would be inconsistent with the rest of the file.

charliedowler:
**Re: Graphite poll timeout comment** — Fixed in commit f7902c2. Added `MAX_POLL_ATTEMPTS = 30` (~30s with exponential backoff). Throws `'Field value search timed out. Please try again.'` on timeout. Covered by 4 unit tests in `useFieldValues.test.ts`.

PR #21974, opened 2026-04-14
fix: parse date strings as UTC in formatDate/formatTimestamp to prevent 1-day offset in stacked bar tooltip

### LINKED ISSUE #21962: Stacked bar charts show dates shifted by 1 day during DST transitions
## Description

When a bar chart is stacked by a dimension, date values can appear shifted by 1 day compared to the totals. This only affects stacked charts — the same data rendered as a non-stacked chart shows correct dates.

The issue started appearing after the UK DST change on March 29, 2026.

## Root cause

In `packages/common/src/pivot/pivotQueryResults.ts`, the `convertSqlPivotedRowsToPivotData` function calls `formatItemValue` with inconsistent `convertToUTC` parameters:

* **Pivot/groupBy dimension values** (line \~1001): `convertToUTC: true`
* **Row totals** (line \~1376): `convertToUTC: false`
* **Index values**: raw values, no conversion

This means when a chart is stacked, the pivot transformation converts groupBy dates to UTC but leaves index dates as-is. During DST transitions, the 1-hour offset is enough to push dates across the midnight boundary, causing records to appear on the wrong day in the stacked segments.

Non-stacked charts don't hit this code path — they render rows directly from the SQL result without pivoting, so they're unaffected.

## Steps to reproduce

1. Create a bar chart with a `TIMESTAMP` dimension truncated to day on the x-axis
2. Stack by another dimension (e.g. user ID)
3. View the chart while in a DST-affected timezone (e.g. Europe/London after March 29)
4. Compare the dates in the stacked segments vs the totals — stacked segments show dates shifted by 1 day

## Expected behavior

## Suggested fix

Make the `convertToUTC` parameter consistent across all `formatItemValue` calls in `convertSqlPivotedRowsToPivotData`. The change is \~3 parameter updates, but needs test coverage for date dimensions in pivoted queries.

Reported via Pylon [PROD-3978](https://linear.app/lightdash/issue/PROD-3978/i-want-to-be-able-to-add-filters-to-my-scheduled-delivery-that-arent).

shu22203:
Relates to: https://github.com/lightdash/lightdash/issues/21962

### Description:

**Bug**: In stacked bar charts, the individual group hover tooltip showed dates 1 day behind the correct date (e.g. showed "2026-03-13" when the correct date was "2026-03-14"). The X-axis and the cumulative tooltip were both correct; only the per-series item-mode tooltip was wrong. The bug only appeared in UTC+ timezones (e.g. JST/UTC+9).

**Root cause**: When `convertToUTC=true`, `formatDate` and `formatTimestamp` used `moment(date).utc()`. For date-only strings like `"2026-03-14"`:

1. `moment("2026-03-14")` parses as **local midnight** (in JST: `2026-03-14T00:00:00+09:00`)
2. `.utc()` switches to UTC view → `2026-03-13T15:00:00Z`
3. Formats as `"2026-03-13"` — 1 day behind

ECharts with `useUTC: true` treats `"2026-03-14"` as UTC midnight (JavaScript's `new Date("2026-03-14")` parses ISO date-only strings as UTC), so the axis and cumulative tooltip showed the correct date.

**Fix**: Change `moment(date).utc()` → `moment.utc(date)` in both `formatDate` and `formatTimestamp`. `moment.utc(date)` parses the string **as UTC** from the start, matching ECharts' `useUTC: true` behaviour.

The cumulative tooltip was unaffected because ECharts provides a numeric timestamp via `axisPointer.label.formatter`, and `moment(numericTimestamp).utc()` always gives the correct UTC date regardless of local timezone.

### DIFF
--- a/packages/common/src/utils/formatting.ts
+++ b/packages/common/src/utils/formatting.ts
@@ -137,11 +137,14 @@ export function formatDate(
     convertToUTC: boolean = false,
     timezone?: string,
 ): string {
+    // moment.utc(date) parses date-only strings as UTC. moment(date).utc()
+    // parses in the local timezone first, shifting the date back a day
+    // in UTC+ browsers (e.g. JST).
     let momentDate;
     if (timezone) {
         momentDate = moment.utc(date).tz(timezone);
     } else if (convertToUTC) {
-        momentDate = moment(date).utc();
+        momentDate = moment.utc(date);
     } else {
         momentDate = moment(date);
     }
@@ -163,7 +166,7 @@ export function formatTimestamp(
     if (timezone) {
         momentDate = moment.utc(value).tz(timezone);
     } else if (convertToUTC) {
-        momentDate = moment(value).utc();
+        momentDate = moment.utc(value);
     } else {
         momentDate = moment(value);
     }


almeidabbm (approved):
Looks great!! Thank you so much for the contribution @shu22203, approved ✅ 

**Before**

Image

**After**

Image

almeidabbm:
@all-contributors please add @shu22203 for code

Issue #21917, opened 2026-04-09
CLI: lightdash dbt run reformats long descriptions by adding line breaks


torimwhaley:
## Bug

`lightdash dbt run -s <model>` adds line breaks to long description strings in YAML files. This happens because the `yaml` library (v2.x) used in `DbtSchemaEditor.toString()` defaults to a `lineWidth` of 80 characters, wrapping any string that exceeds that length.

This is cosmetic but unexpected -- the CLI shouldn't be reformatting existing content when adding missing columns.

## Steps to reproduce

1. Have a dbt model YAML file with a description longer than 80 characters on a single line
2. Run `lightdash dbt run -s <model>`
3. Observe that the description now has line breaks inserted

## Expected behavior

Existing descriptions should be preserved as-is. The CLI should only add missing columns, not reformat existing content.

## Fix

One-line change in `packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.ts` -- add `lineWidth: 0` to the `this.doc.toString()` options to disable line wrapping:

```ts
return this.doc.toString({
    singleQuote:
        options?.quoteChar && options.quoteChar === "'" ? true : null,
    lineWidth: 0,
});
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #21977, opened 2026-04-14
fix: preserve long YAML descriptions without line wrapping in CLI

charliedowler:
## Bug
`lightdash dbt run -s <model>` reformats long description strings in dbt model YAML files by inserting line breaks at 80 characters when adding missing columns.

Closes #21917

## Expected
Existing descriptions should be preserved exactly as-is, with no reformatting.

## Reproduction
Failing tests in `packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.test.ts` — the "long description preservation" describe block contains two tests that parse YAML with >80 char descriptions and assert they survive a round-trip through `toString()`.

## Evidence (before)
- Failing test output shows descriptions wrapped at 80 chars:
  ```
  - description: This is a very long description that exceeds eighty characters and
  +   should not be reformatted or wrapped by the yaml serializer at all
  ```

## Fix
**Root cause:** `DbtSchemaEditor.toString()` called `this.doc.toString()` without a `lineWidth` option. The `yaml` v2 library defaults to `lineWidth: 80`, causing all plain scalar strings longer than 80 characters to be folded with line breaks.

**Change:** Added `lineWidth: 0` to the `toString()` options in `packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.ts`, which disables line wrapping entirely.

## Evidence (after)
- All 15 DbtSchemaEditor tests passing, including 2 new regression tests
- typecheck / lint: ✅

### DIFF
--- a/packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.mock.ts
+++ b/packages/common/src/dbt/DbtSchemaEditor/DbtSchemaEditor.mock.ts
@@ -180,19 +180,15 @@ models:
             fixed_width_id:
               label: Fixed width name
               type: string
-              sql: CONCAT(FLOOR(\${TABLE}.dim_a / 10) * 10, ' - ', (FLOOR(\${TABLE}.dim_a / 10)
-                + 1) * 10 - 1)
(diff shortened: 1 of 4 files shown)


charliedowler:
Closing — jarvis instance-3 reset

PR #21981, opened 2026-04-14
feat: add 'Start at zero' toggle for percentage y-axis on line/area charts

### LINKED ISSUE #21730: Percentage-formatted line charts should default y-axis to start at zero
## Description

When a line chart displays percentage-formatted values, the y-axis auto-scaling heuristic can zoom into a narrow range (e.g. 85%-90%), making small fluctuations look dramatic. Percentages have a natural 0-100 scale, so the y-axis should default to starting at 0 to give proper visual context.

Bar charts already handle this correctly -- the auto-scaling logic is disabled for bar series, so they always start at zero. Line and area charts with percentage values should follow the same principle.

## Current behavior

A line chart showing values like 85%, 87%, 89% may render with a y-axis range of \~84%-90%, exaggerating visual variation.

## Expected behavior

When the y-axis displays percentage-formatted values, the axis should default to starting at 0 (and ideally cap at 100 if all values fall within 0-100).

## Implementation considerations

Changing the default auto-scaling behavior will affect how existing saved charts render. A chart that currently zooms into a tight percentage range would suddenly show the full 0-100 scale, which could make some charts less readable for users who prefer the zoomed view.

Options to handle this:

1. **Migration approach**: When applying the new default, backfill existing charts that rely on the current heuristic with an explicit min/max value matching their current rendered range, so they don't visually change. New charts get the zero-based default.
2. **Opt-in toggle**: Add an explicit "Start at zero" axis option (defaulting to on for new percentage charts) so users have direct control rather than relying on heuristics.
3. **Version the default**: Apply the new behavior only to newly created charts, leaving existing charts on the old heuristic.

## Workaround

Users can manually set the y-axis min to 0 in the Axes config panel.

## Related

* [PROD-553](https://linear.app/lightdash/issue/PROD-553/axis-auto-range-is-confusing-and-inconsistent) (broader axis auto-range inconsistency)

charliedowler:
## Summary
- Adds a **"Start axis at zero"** toggle to the Y-axis section of the Axes config panel for line and area charts with percentage-formatted fields
- When enabled, forces the y-axis minimum to 0 instead of auto-scaling to a narrow range around the data (e.g. 84–90%), which exaggerates small fluctuations
- **New charts** with percentage-formatted line/area series default the toggle to **ON**
- **Existing saved charts** are unaffected — the toggle defaults to OFF when `startAtZero` is not present in the saved config

## Changes
- `packages/common/src/types/savedCharts.ts` — Added optional `startAtZero` property to the `Axis` type
- `packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.ts` — When `startAtZero` is true, forces y-axis min to 0 (for both left and right axes), taking priority over the default range heuristic
- `packages/frontend/src/hooks/cartesianChartConfig/useCartesianChartConfig.ts` — Added `setYAxisStartAtZero` setter and a one-time effect that auto-defaults `startAtZero: true` for new (unsaved) charts with percentage-formatted line/area series
- `packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx` — Added the "Start axis at zero" Switch control, shown only when the axis has percentage-formatted fields and chart type is line or area

## Test plan
- [ ] Create a new line chart with a percentage-formatted metric → verify y-axis starts at 0 by default
- [ ] Toggle "Start axis at zero" OFF → verify y-axis auto-scales to data range
- [ ] Open an existing saved line chart with percentage y-axis → verify no visual change (toggle shows as OFF)
- [ ] Verify the toggle does not appear for bar charts or non-percentage fields
- [ ] Verify the toggle works on both left and right y-axes
- [ ] Save a chart with the toggle ON, reload → verify setting persists

Closes #21730

### DIFF
--- a/packages/common/src/types/savedCharts.ts
+++ b/packages/common/src/types/savedCharts.ts
@@ -621,6 +621,8 @@ type Axis = {
     inverse?: boolean;
     /** Rotation angle for axis labels */
     rotate?: number;
+    /** Force y-axis to start at zero (useful for percentage-formatted axes) */
+    startAtZero?: boolean;
 };
 
 export type XAxis = Axis & {
--- a/packages/e2e/cypress/cli/dbt/cli.cy.ts
+++ b/packages/e2e/cypress/cli/dbt/cli.cy.ts
@@ -6,7 +6,7 @@ describe('CLI', () => {
     // Scale timeout based on the number of models and thread count (both
     // read dynamically in cypress.config.ts). dbt runs models in parallel,
     // so we estimate batches rather than sequential model count.
-    const TIMEOUT_PER_BATCH_MS = 3000;
+    const TIMEOUT_PER_BATCH_MS = 6000;
     const BASE_TIMEOUT_MS = 30000;
     const modelCount = Number(Cypress.env('MODEL_COUNT')) || 50;
     const dbtThreads = Number(Cypress.env('DBT_THREADS')) || 4;
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/index.tsx
@@ -4,6 +4,7 @@ import {
     getDateGroupLabel,
     getItemLabelWithoutTableName,
     getXAxisSort,
+    hasPercentageFormat,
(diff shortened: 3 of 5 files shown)


charliedowler:
Closing — jarvis instance-2 reset

PR #22004, opened 2026-04-14
fix: stacked bar chart dates shifted by 1 day during DST transitions

Spissable:
## Bug
In stacked bar charts, date values in pivot segments are shifted by 1 day compared to totals during DST transitions. The `convertSqlPivotedRowsToPivotData` function calls `formatItemValue` with `convertToUTC: true` for pivot/groupBy header values, but `false` for row totals and no conversion for index values. During DST, the UTC conversion pushes dates across the midnight boundary.

## Expected

## Reproduction
- Failing test: `packages/common/src/pivot/pivotQueryResults.test.ts` — "should not shift date dimensions during DST transitions"
- Uses `TZ=Europe/London` with a BST timestamp (`2026-03-29T00:30:00.000+01:00`) as a pivot groupBy value
- Header value incorrectly shows `2026-03-28` instead of `2026-03-29`

## Evidence (before)
- Failing test asserts `headerValue.value.formatted` equals `'2026-03-29'` but receives `'2026-03-28'`

## Fix
Changed `convertToUTC` from `true` to `false` in the `formatItemValue` call for pivot header values (`packages/common/src/pivot/pivotQueryResults.ts:1027`). This makes the parameter consistent with row totals (same function, line 1405) and the non-SQL pivot path (lines 262, 277) which all use `false`.

Root cause: `moment(date).utc()` shifts timestamps to UTC, causing dates near midnight in DST-affected timezones to cross day boundaries (e.g. `2026-03-29T00:30:00+01:00` → `2026-03-28T23:30:00Z` → formatted as March 28 instead of March 29).

## Evidence (after)
- Test now passing: `packages/common/src/pivot/pivotQueryResults.test.ts`
- typecheck / lint / test: ✅

### DIFF
--- a/packages/common/src/pivot/pivotQueryResults.mock.ts
+++ b/packages/common/src/pivot/pivotQueryResults.mock.ts
@@ -4,10 +4,12 @@ import {
     DimensionType,
     FieldType,
     MetricType,
+    type Dimension,
 } from '../types/field';
 import { type MetricQuery } from '../types/metricQuery';
 import { type PivotData } from '../types/pivot';
 import { type ResultRow } from '../types/results';
+import { TimeFrames } from '../types/timeFrames';
 import {
     SortByDirection,
     VizAggregationOptions,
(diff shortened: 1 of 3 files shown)


Spissable:
Just a test

PR #22006, opened 2026-04-14
fix: stacked bar charts show dates shifted by 1 day during DST transitions

### LINKED ISSUE #21962: Stacked bar charts show dates shifted by 1 day during DST transitions

Spissable:
## Bug

Stacked bar charts show dates shifted by 1 day compared to totals during DST transitions. When a chart is stacked by a dimension, the pivot transformation converts groupBy dates using `convertToUTC: true`, which causes `moment().utc()` to shift timestamps with timezone offsets (e.g. `+01:00` during BST) across midnight boundaries.

Closes #21962

## Expected

## Fix

**Root cause:** In `convertSqlPivotedRowsToPivotData` (`packages/common/src/pivot/pivotQueryResults.ts`), the `formatItemValue` call for pivot header values (groupBy dimension values) passed `convertToUTC: true` (line ~1027), while row totals (line ~1405) and the non-SQL pivot path (lines 262, 277) all used `false`. During DST transitions, this inconsistency caused dates like `2026-04-01T00:00:00+01:00` (midnight BST) to be formatted as `2026-03-31` (the previous day in UTC).

**Fix:** Changed `convertToUTC` from `true` to `false` in the single affected `formatItemValue` call, making it consistent with all other paths.

## Evidence (after)

- Test: `packages/common/src/pivot/pivotQueryResults.test.ts` — "should not UTC-convert date dimension values in pivot headers (DST fix)"
- The test uses `TZ=Europe/London` with a post-DST date (`2026-04-01T00:00:00.000+01:00`) and verifies headers show `2026-04-01`, not `2026-03-31`
- typecheck / lint / test: ✅

### DIFF
--- a/packages/common/src/pivot/pivotQueryResults.test.ts
+++ b/packages/common/src/pivot/pivotQueryResults.test.ts
@@ -5,6 +5,7 @@ import {
     type ItemsMap,
 } from '../types/field';
 import { TimeFrames } from '../types/timeFrames';
+import * as formatting from '../utils/formatting';
 import { VizAggregationOptions, VizIndexType } from '../visualizations/types';
 import {
     convertSqlPivotedRowsToPivotData,
@@ -1467,8 +1468,9 @@ describe('convertSqlPivotedRowsToPivotData', () => {
         };
 
         const dstDate = '2026-04-01T00:00:00.000+01:00';
+        const spy = jest.spyOn(formatting, 'formatItemValue');
 
-        const result = convertSqlPivotedRowsToPivotData({
+        convertSqlPivotedRowsToPivotData({
             rows: [
                 {
                     orders_created_day: {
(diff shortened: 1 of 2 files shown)


Spissable:
Just a test for the hackathon

PR #22014, opened 2026-04-14
fix: validate table and filters in field value search

charliedowler:
## Bug
`POST /api/v1/projects/:projectUuid/field/:fieldId/search` returns 500 UnexpectedServerError when:
1. Request body is missing `table` → Knex undefined binding error
2. Request body has `filters` object without `.and` property → TypeError

Both should return 4xx validation errors instead. Same code path is reachable from the v2 endpoint via `AsyncQueryService.executeAsyncFieldValueSearch`.

## Expected
- Missing/empty `table` → 400 ParameterError
- Malformed `filters` (no `.and`) → handled gracefully (ignored)

## Reproduction
- `handles filters object missing .and property gracefully`

## Evidence (before)
- Failing test: `packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts` — see CI run

## Fix
**Root cause:** `fieldValuesQueryBuilder.ts` passed `table` directly to `exploreResolver.findExploreByTableName()` without validating it's non-empty — `undefined`/`""` reaches Knex as an undefined binding. Separately, `filters.and.filter(...)` at line 110 assumed `.and` exists whenever `filters` is truthy.

**Changes:**
- Added early `!table` guard that throws `ParameterError` with message `'Field value search requires a non-empty "table"'`
- Changed `if (filters)` to `if (filters && Array.isArray(filters.and))` so malformed filter objects are safely skipped

Both v1 and v2 endpoints are hardened because they share `getFieldValuesMetricQuery`.

## Evidence (after)
- All 10 tests passing: `packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts`
- typecheck / lint / test: ✅

### DIFF
--- a/packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts
+++ b/packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts
@@ -2,6 +2,7 @@ import {
     FilterOperator,
     NotFoundError,
     ParameterError,
+    type AndFilterGroup,
     type Explore,
 } from '@lightdash/common';
 import { getFieldValuesMetricQuery } from './fieldValuesQueryBuilder';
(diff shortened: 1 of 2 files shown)


charliedowler:

PR #22015, opened 2026-04-14
fix: validate table and filters in field value search to prevent 500 errors

charliedowler:
## Bug
`POST /api/v1/projects/:projectUuid/field/:fieldId/search` returns 500 UnexpectedServerError when:
1. Request body is missing `table` — causes Knex undefined binding error
2. Request body has `filters` object without `.and` property — causes TypeError

Sentry: LIGHTDASH-BACKEND-CF9, LIGHTDASH-BACKEND-CF8

## Expected
Both cases should return 4xx validation errors, not 500.

## Reproduction
- `handles filters object without .and property gracefully`

## Evidence (before)
- `table: undefined` → unhandled Knex error (`Undefined binding(s) detected when compiling SELECT`)
- `table: ''` → silently resolves instead of rejecting (hits Knex downstream)
- `filters` without `.and` → `TypeError: Cannot read properties of undefined (reading 'filter')` at `fieldValuesQueryBuilder.ts:110`

## Fix
**Root cause:** `getFieldValuesMetricQuery` in `fieldValuesQueryBuilder.ts` did not validate that `table` is a non-empty string before passing it to `exploreResolver.findExploreByTableName()`, and accessed `filters.and` without checking `.and` exists.

**Changes:**
- Added early `!table` check that throws `ParameterError` with message `"Field value search requires a non-empty "table""` (line 54)
- Changed `if (filters)` to `if (filters?.and)` so a `filters` object missing `.and` is silently skipped instead of crashing (line 113)

Both v1 (`projectRouter`) and v2 (`AsyncQueryService.executeAsyncFieldValueSearch`) code paths use the same `getFieldValuesMetricQuery` helper, so both are hardened by this fix.

## Evidence (after)
- All 10 tests passing in `fieldValuesQueryBuilder.test.ts` — including 3 new regression tests
- typecheck / lint / test: ✅

### DIFF
--- a/packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts
+++ b/packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts
@@ -2,6 +2,7 @@ import {
     FilterOperator,
     NotFoundError,
     ParameterError,
+    type AndFilterGroup,
     type Explore,
 } from '@lightdash/common';
 import { getFieldValuesMetricQuery } from './fieldValuesQueryBuilder';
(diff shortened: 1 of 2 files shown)


charliedowler:

PR #22023, opened 2026-04-14
fix: validate table and filters in field value search to prevent 500s

charliedowler:
## Bug
POST `/api/v1/projects/:projectUuid/field/:fieldId/search` and `POST /api/v2/projects/:projectUuid/query/field-values` return **500 UnexpectedServerError** when:
1. Request body is missing `table` field — Knex throws `Undefined binding(s)` error
2. `filters` object is present but missing `.and` property — TypeError: `Cannot read properties of undefined (reading 'filter')`

Sentry: LIGHTDASH-BACKEND-CF9, LIGHTDASH-BACKEND-CF8

## Expected
Both cases should return 400 ParameterError with a meaningful message.

## Reproduction
- `throws ParameterError when filters is truthy but missing .and`

## Evidence (before)
- Failing tests: `packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts` — see CI run for repro commit

## Fix
**Root cause**: `getFieldValuesMetricQuery` in `fieldValuesQueryBuilder.ts` had no input validation for `table` or `filters.and`:
- `table` was passed directly to `exploreResolver.findExploreByTableName()` → Knex query with undefined binding → 500
- `filters.and.filter(...)` was called when `filters` was truthy but `.and` was undefined → TypeError → 500

**Change**: Added two early validation guards at the top of `getFieldValuesMetricQuery`:
1. `if (!table)` → throws `ParameterError('Field value search requires a non-empty "table" parameter')`
2. `if (filters && !Array.isArray(filters.and))` → throws `ParameterError('Field value search "filters" must include an "and" array')`

Both v1 and v2 endpoints flow through this shared function, so both are covered.

## Evidence (after)
- All 10 tests passing: `packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts`
- typecheck / lint: ✅

### DIFF
--- a/packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts
+++ b/packages/backend/src/services/ProjectService/fieldValuesQueryBuilder.test.ts
@@ -2,6 +2,7 @@ import {
     FilterOperator,
     NotFoundError,
     ParameterError,
+    type AndFilterGroup,
     type Explore,
 } from '@lightdash/common';
 import { getFieldValuesMetricQuery } from './fieldValuesQueryBuilder';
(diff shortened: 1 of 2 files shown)


charliedowler:

PR #22083, opened 2026-04-17
refactor: extract navbar root element function and optimize Mantine8Provider theme handling

tatianainama:
stabilize `Mantine8Provider`props so it stops invalidating memos and ref pipelines on every render. Also updated NavBar to hoist a constant for getting the root element instead of using inline arror that generates new reference on every render 

this is part of a work to trying to solve PROD-7019

### DIFF
--- a/packages/frontend/src/components/NavBar/index.tsx
+++ b/packages/frontend/src/components/NavBar/index.tsx
@@ -57,6 +57,9 @@ const NavBarContent = ({
     );
 };
 
+const getNavBarRootElement = () =>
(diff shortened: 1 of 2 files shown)


tatianainama:
* **#22083** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22083?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22083?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22090, opened 2026-04-17
feat: Athena Warehouse ReuseResultConfiguration 

### LINKED ISSUE #21978: Athena connector: support ResultReuseConfiguration to enable query result caching
### Description

Amazon Athena supports query result reuse — repeated identical queries can return cached results instantly, avoiding unnecessary scan costs. However, this feature must be explicitly enabled per query via a ResultReuseConfiguration block in the StartQueryExecution API call. If this block is omitted, Athena defaults to `Enabled: false`.

Currently, the Lightdash Athena connector does not include `ResultReuseConfiguration` in its query execution calls, meaning result caching is never activated.

### Proposed solution

Add support for `ResultReuseConfiguration` in the Athena connector's query execution call. This could be exposed as an optional connection-level setting, for example:

```
{
  "ResultReuseConfiguration": {
    "ResultReuseByAgeConfiguration": {
      "Enabled": true,
      "MaxAgeInMinutes": 60 * 5
    }
  }
}
```

When not configured, the connector should continue its current behaviour (no block sent, caching disabled).

- **Cost** — without result reuse, every Lightdash query scans data in S3 even when results are identical to a recent execution, leading to unnecessary charges.
- **Performance** — cached results are returned in milliseconds vs. seconds or minutes for a full scan, which significantly improves dashboard load times.

hhhonzik:
Closes: #21978

### Description:

Adds additional controls to Athena Warehouse for ReuseResultConfiguration. 

Official docs: https://docs.aws.amazon.com/athena/latest/ug/reusing-query-results.html

Screenshot 2026-04-17 at 16 15 35

### DIFF
--- a/packages/common/src/types/projects.ts
+++ b/packages/common/src/types/projects.ts
@@ -209,6 +209,8 @@ export type CreateAthenaCredentials = {
     requireUserCredentials?: boolean;
     startOfWeek?: WeekDay | null;
     dataTimezone?: string;
+    resultReuseEnabled?: boolean;
+    resultReuseMaxAgeInMinutes?: number;
 };
 
 export type AthenaCredentials = Omit<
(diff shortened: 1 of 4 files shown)


joaoviana:
Thank you for the contribution, closing as we can pick this work in the near future.

Issue #22069, opened 2026-04-16
Dashboard tile shows "No data available" when query fails with empty error message


jesshitchcock:
## Bug

When a dashboard chart tile's query errors with a missing or empty message, the UI falls back to rendering `"No data available"` as the error description. This is misleading: it looks identical to a successful query that returned zero rows, but the tile actually failed.

## Where

`packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx`:

```ts
description={
    getDashboardTileErrorMessage(error) || 'No data available'
}
```

`getDashboardTileErrorMessage` returns `error.error?.message` (or `error.message` for native `Error`). Whenever that resolves to `undefined` or empty string, the fallback string `"No data available"` is shown — in an error-state UI (`SuboptimalState` with `IconAlertCircle`), but the text makes it look like a valid empty result.

## When does this happen?

Several paths can produce a tile error with no message:

- Query polling hits `EXPIRED` / `CANCELLED` states where the returned `error` field is undefined
- Network-level request aborts (e.g. a fast refresh cancels in-flight polls)
- Embed JWT expires mid-poll and the auth bounce doesn't carry a user-facing message
- Any future error path that forgets to populate `error.message`

This is especially painful for embedded dashboards where multiple tile queries run concurrently — intermittent failures surface as "a different tile shows 'No data available' on each refresh", which is indistinguishable from genuine empty datasets and makes root-cause investigation very difficult in the field.

## Expected

- `"No data available"` should only be shown when the query succeeded with zero rows.
- On an error with missing message, show a clearly-error label such as `"Failed to load chart"` and log the raw error object (including status code / error name / any upstream details) to the browser console so engineers can diagnose.
- Optionally surface an error ID / status code in the UI when available.

## Impact

Makes embed and iframe chart-loading bugs very hard to diagnose from customer reports — support receives "charts aren't loading" tickets with no error signal to work from, because the UI has silently swallowed the real failure.

PR #22121, opened 2026-04-20
fix: show non-empty error message on failed dashboard tile queries

pimmey:
Closes: PROD-7014 / #22069

### Description:

When a query fails with an empty error message from the backend, the dashboard tile was silently showing "No data available" instead of indicating an actual error occurred. This masked real query failures and made it difficult to distinguish between a query returning no results and a query that errored.

- `getErrorMessage` now guards against empty error messages, falling back to a descriptive unknown error string when the message is falsy
- `useInfiniteQueryResults` uses a logical OR (`||`) instead of nullish coalescing (`??`) when constructing the error message, so empty strings correctly fall back to `'Query failed'`
- Dashboard chart tiles (both standard and SQL) now display `'Error running query'` instead of `'No data available'` when an error is present
- A regression test has been added to verify that an empty backend error string surfaces as a non-empty message on the error object

### DIFF
--- a/packages/common/src/types/errors.ts
+++ b/packages/common/src/types/errors.ts
@@ -510,8 +510,10 @@ export class NotImplementedError extends LightdashError {
         });
     }
 }
-export const getErrorMessage = (e: unknown) =>
-    e instanceof Error ? e.message : `Unknown ${typeof e} error`;
+export const getErrorMessage = (e: unknown) => {
+    if (e instanceof Error && e.message) return e.message;
+    return `Unknown ${typeof e} error`;
+};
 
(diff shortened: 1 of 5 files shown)


pimmey:
* **#22121** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22121?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22108, opened 2026-04-20
refactor: add timezone parameter to formatting functions

almeidabbm:
Relates to: https://linear.app/lightdash/issue/GLITCH-291

## Summary

- Add optional `timezone?: string` parameter to core formatting functions (`formatDate`, `formatTimestamp`, `formatItemValue`, `formatRow`, `formatRows`, `getFormattedValue`, `valueFormatter`)
- When `timezone` is provided, it takes precedence over the existing `convertToUTC` boolean — uses `moment.utc(value).tz(timezone)` for correct timezone-aware formatting
- Existing callers are completely unaffected — all new parameters are optional, no signature changes to existing call sites
- This is the foundation for GLITCH-291 (timezone-aware result formatting). Backend and frontend threading come in follow-up PRs.

## Test plan

- [x] New tests for `formatDate` and `formatTimestamp` with timezone parameter
- [x] Tests verify timezone takes precedence over `convertToUTC`
- [x] Tests verify existing behavior unchanged when no timezone provided
- [x] All existing formatting tests pass (80/80)
- [x] Typecheck passes for common, backend, and frontend

PR #22142, opened 2026-04-21
fix(formatting): remove timezone param from formatDate

almeidabbm:
Reverts the double-shift on DATE values introduced by #22108.

## Problem

#22108 added an optional `timezone` parameter to `formatDate`. When supplied, the function did `moment.utc(date).tz(timezone)` — which double-shifts DATE values that were already truncated in the project timezone at the SQL layer (timezone-aware `DATE_TRUNC`). E.g. with project timezone `America/New_York`, a Jan 15 bucket would display as Jan 14.

A `DATE` has no time component — a calendar day has no instant to shift — so the parameter was semantically meaningless.

## Flag-off parity

When no timezone is supplied, behavior is the same as pre-#22108.

No production callers passed a timezone to `formatDate` directly — only tests and the `formatItemValue` dispatcher (which is updated here).

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -1741,24 +1741,6 @@ describe('Formatting', () => {
         const utcTimestamp = '2020-04-04T02:00:00.000Z';
 
         describe('formatDate', () => {
-            test('formats date in project timezone', () => {
-                // 2020-04-04 02:00 UTC = 2020-04-03 22:00 New York
-                expect(
-                    formatDate(
-                        utcTimestamp,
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#22142** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22142?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22142?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22150, opened 2026-04-21
fix: route sum_distinct refs through dd CTE when fanout protection is active (SPK-333)

rephus:
Closes: [SPK-333](https://linear.app/lightdash/issue/SPK-333/investigate-sum-distinct-metrics-referenced-by-other-metrics)

### Before

CleanShot 2026-04-21 at 13 24 06

### After

CleanShot 2026-04-21 at 13 22 52

### Issue:   

SPK-333 — Investigate sum_distinct metrics referenced by other metrics

### Symptom: 

When a type:number metric references a sum_distinct metric AND another metric in the
query triggers fanout protection (cte_keys_/cte_metrics_/cte_unaffected), the
generated SQL projects the type:number metric inside dd_base with the sum_distinct
reference expanded to its inlined fallback `SUM("orders".amount)`. That expression
references a base table that is not in scope of the dd_base CROSS JOIN cte_unaffected
context, producing either invalid SQL or a duplicate-column error (the same metric
also showed up in the outer SELECT via nonAggDdSelects, correctly routed).
###  Root cause: 

getExperimentalMetricsCteSQL (the fanout-protection code path) pushed every
non-aggregate metric with cross-table references into metricsWithCteReferences and
then inlined their compiledSql via replaceMetricReferencesWithCteReferences. When the
referenced metric was a sum_distinct, no CTE contained it yet (the dd_* CTE is built
later), so the reference fell through to compileMetricSql and was inlined as the raw
fallback SUM. The existing fix (PR #21923) only applied to the non-fanout path via
getMetricsSQL; the fanout path was never taught about nonAggReferencingDd.
###  Fix:

packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts:1900, 1921, 2324 —
inside getExperimentalMetricsCteSQL, compute nonAggReferencingDd and skip adding
those metrics to metricsWithCteReferences; also exclude them from unaffectedMetrics.
The referenced dependencies (e.g. unique_customer_count) are still collected into
referencedMetricObjects so they land in cte_unaffected/dd_base where the later
nonAggDdSelects logic can resolve the reference via dd_base.
###  Test:

Added unit test `non-aggregate referencing sum_distinct should route through dd CTE
even with fanout protection` + mock `EXPLORE_WITH_FANOUT_AND_DD_REFERENCE`. Test
fails without the fix (inlined SUM in dd_base body) and passes with it. Also
manually verified via the live Jaffle shop — both SPK-333 scenarios return
mathematically consistent data ($4,149.12 deduped = orders total; $41.08 CLV =
4149.12 / 101 customers). All 167 existing MetricQueryBuilder tests + 59 SQL
snapshots pass unchanged.
### Regression risk: LOW 

the change is scoped to the fanout-protection branch and gated on
`nonAggReferencingDd.has(id)`, a set already consumed correctly by the later dd CTE
flow; existing fanout-only and distinct-only paths are byte-identical via snapshots.
### Status:

          DONE

---

### Description:

Fixes a bug where a `type: number` metric referencing a `sum_distinct` metric would produce invalid SQL when another metric on the same query triggered fanout protection (i.e. the `cte_keys_`/`cte_metrics_`/`cte_unaffected` CTE flow).

Previously, when fanout protection was active, the query builder would inline the non-aggregate metric's `${sum_distinct}` reference as a raw fallback `SUM(...)` directly into `dd_base`. This was invalid because the underlying table (e.g. `"orders"`) is not in scope inside `dd_base`, which projects from `cte_unaffected CROSS JOIN ...` rather than the raw joined tables.

The fix identifies non-aggregate metrics that reference a `sum_distinct` or `average_distinct` metric and defers their projection to the outer `dd_base` SELECT, where the `dd_*` CTE aliases are in scope. Their transitive dependencies (e.g. a `count_distinct` in the denominator) are still collected so they continue to land correctly in `cte_unaffected`/`dd_base`. Additionally, these metrics are excluded from `cte_unaffected` to prevent their broken `compiledSql` from being emitted there.

A new test case (`EXPLORE_WITH_FANOUT_AND_DD_REFERENCE` / `METRIC_QUERY_FANOUT_AND_DD_REFERENCE`) covers the repro scenario: `average_customer_lifetime_value` (a `type: number` metric dividing `total_order_amount_deduped` by `unique_customer_count`) selected alongside `orders.total_order_amount` and `payments.payment_count`, both of which trigger fanout protection. The test asserts that:
- The `dd_customers_total_order_amount_deduped` CTE alias is used in the final expression rather than the raw `SUM("orders".amount)`
- `customers_average_customer_lifetime_value` is aliased exactly once in the output, avoiding "column specified more than once" errors

deploy-preview frontend-test backend-test

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.mock.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.mock.ts
@@ -3272,3 +3272,213 @@ export const METRIC_QUERY_CROSS_MODEL_SUM_DISTINCT_NO_DIMS: CompiledMetricQuery 
         compiledAdditionalMetrics: [],
         compiledCustomDimensions: [],
     };
+
+// Repro for SPK-333: a type:number metric referencing a sum_distinct metric
+// is selected alongside a simple sum metric on a joined (one-to-many) table
+// that triggers fanout protection. The non-aggregate metric's ${sum_distinct}
+// reference must resolve to the dd CTE alias; inlining the fallback SUM breaks
+// because the raw table is not in scope of the dd_base outer SELECT.
+export const EXPLORE_WITH_FANOUT_AND_DD_REFERENCE: Explore = {
+    targetDatabase: SupportedDbtAdapter.POSTGRES,
+    name: 'customers',
+    label: 'customers',
+    baseTable: 'customers',
+    tags: [],
+    tables: {
+        customers: {
+            name: 'customers',
+            label: 'customers',
+            database: 'mydb',
+            schema: 'public',
+            sqlTable: 'customers',
+            primaryKey: ['customer_id'],
+            lineageGraph: {},
+            dimensions: {
+                customer_id: {
+                    type: DimensionType.STRING,
+                    name: 'customer_id',
+                    label: 'Customer ID',
+                    table: 'customers',
+                    tableLabel: 'customers',
+                    fieldType: FieldType.DIMENSION,
+                    sql: '${TABLE}.customer_id',
+                    compiledSql: '"customers".customer_id',
+                    tablesReferences: ['customers'],
+                    hidden: false,
+                },
+            },
+            metrics: {
+                unique_customer_count: {
+                    type: MetricType.COUNT_DISTINCT,
+                    name: 'unique_customer_count',
+                    label: 'Unique Customer Count',
+                    table: 'customers',
+                    tableLabel: 'customers',
+                    fieldType: FieldType.METRIC,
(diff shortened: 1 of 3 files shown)


rephus:
* **#22150** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22150?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22166, opened 2026-04-21
fix: defer SDK filter conversion until tile metadata and available filters are ready

### LINKED ISSUE #22148: SDK filters dropped on cold load when dashboard tiles span multiple explores
## Description

When the React SDK passes `filters` to an embedded dashboard whose tiles span multiple explores, the filter is silently dropped from any tile whose explore differs from the filter's `model` on the **initial browser load** (cold render). The previous fix in [PROD-5897](https://linear.app/lightdash/issue/PROD-5897/sdk-filters-dont-auto-map-across-tiles-using-different-explores) / lightdash/lightdash#21412 only kicks in when `availableFilters` data is already present at the time `convertSdkFilterToDashboardFilter` runs — and on a cold load it isn't.

This is a regression of the symptom from [PROD-5897](https://linear.app/lightdash/issue/PROD-5897/sdk-filters-dont-auto-map-across-tiles-using-different-explores) caused by a load-order race, not by the matcher itself.

## Reproduction

1. Embed a dashboard via the React SDK whose tiles use two or more explores (e.g. one tile on `model_a`, another on `model_b`).
2. Pass an SDK filter targeting `model_a`'s field, where `model_b` also has a same-named field of the same type:

   ```jsx
   <Dashboard filters={[{ model: 'model_a', field: 'shared_field', operator: FilterOperator.EQUALS, value: 'x' }]} />
   ```
3. Cold-refresh the embedding page in the browser.
4. Observe: only the `model_a` tiles are filtered. Tiles on `model_b` ignore the filter.
5. Click the dashboard's "Reset all filters" button (or "Refresh dashboard"), then re-apply the filter via the SDK / UI flow.
6. Observe: now all tiles are correctly filtered.

## Real-world report

Customer (GoStudent), dashboard `4b14aeb8-4962-4428-8054-f545c19731aa`. Three tiles, two on `fct_gocampus_centers_report`, one on `fct_gocampus_subscriptions`. SDK filter on `full_center_name`. The `fct_gocampus_subscriptions` table tile renders unfiltered on cold load; the two chart tiles on `fct_gocampus_centers_report` filter correctly.

HAR confirms: on cold load, all three `query/dashboard-tile` POST bodies include the filter with `tileTargets: {}`. After UI reset, the same three requests include `tileTargets` with all three tile UUIDs correctly mapped (including `fct_gocampus_subscriptions_full_center_name` for the table tile).

## Root cause

`useDashboardsAvailableFilters` (`packages/frontend/src/hooks/dashboard/useDashboard.ts:120-137`) is `enabled: savedChartUuidsAndTileUuids.length > 0`. On the first render of `DashboardProvider`, `dashboardTiles` is undefined, so `savedChartUuidsAndTileUuids ?? []` is empty and the query is **disabled**. React Query reports `isInitialLoading: false` for disabled queries.

The SDK-filter branch in `DashboardProvider.tsx:671-689` guards with:

```ts
if (sdkFilters.length > 0) {
    if (isLoadingDashboardFilters) return;
    updatedDashboardFilters.dimensions = sdkFilters.map(
        (sdkFilter) => convertSdkFilterToDashboardFilter(sdkFilter, filterableFieldsByTileUuid),
    );
}
```

Because `isLoadingDashboardFilters` is already `false` (the query hasn't started — it's disabled), the guard doesn't fire. `convertSdkFilterToDashboardFilter` is called with `filterableFieldsByTileUuid = undefined` and falls back to `tileTargets: {}` (`packages/frontend/src/ee/features/embed/EmbedDashboard/utils.ts:67-69`).

Once `dashboardFilters` is set to a non-empty value, the outer `if (dashboardFilters === emptyFilters)` sentinel on line 665 prevents the effect from re-running when the real available-filters data later arrives — locking in the empty `tileTargets`.

## Proposed fix

Tighten the SDK-filter gate in `DashboardProvider.tsx` to also wait for tile metadata and available-filters data to actually exist:

```ts
if (sdkFilters.length > 0) {
    if (!savedChartUuidsAndTileUuids) return;
    if (savedChartUuidsAndTileUuids.length > 0 && !filterableFieldsByTileUuid) return;
    // ...convert as before
}
```

`isLoadingDashboardFilters` is insufficient because it stays `false` for disabled queries; gating on `filterableFieldsByTileUuid` covers both "tiles haven't loaded yet" and "available-filters request still in flight".

## Workaround

Pass one SDK filter per model, so that each filter's `target.fieldId` matches its tile's explore directly (the backend `getDashboardFilterRulesForTables` matches by `fieldId`, not by `tileTargets`):

```js
const filters = ['model_for_chart', 'model_for_table'].flatMap(model => [
  { model, field: 'shared_field', operator: FilterOperator.EQUALS, value },
]);
```

This avoids depending on cross-explore `tileTargets` mapping entirely.

## Related

* [PROD-5897](https://linear.app/lightdash/issue/PROD-5897/sdk-filters-dont-auto-map-across-tiles-using-different-explores) — original cross-explore SDK filter fix (matcher logic, shipped in lightdash/lightdash#21412)
* lightdash/lightdash#21095 — embed filter autocomplete fallback fix

pimmey:
Closes: #22148

### Description:

Fixes a bug where SDK filters applied to embedded dashboards would silently drop filter targets for tiles on different explores during a cold load.

The root cause was that the available-filters query is disabled until dashboard tile metadata has loaded. React Query reports a disabled query as `isInitialLoading: false`, so gating SDK filter conversion on `isLoadingDashboardFilters` allowed the converter to fire too early — before `filterableFieldsByTileUuid` was populated — resulting in `tileTargets: {}` and the filter only applying to the primary explore's tiles.

The fix introduces a `shouldDeferSdkFilters` utility that correctly gates conversion on the underlying data rather than the loading flag:

- Defers if tile metadata (`savedChartUuidsAndTileUuids`) hasn't arrived yet
- Proceeds immediately if there are no filterable chart tiles (nothing to wait for)
- Defers if chart tiles exist but `filterableFieldsByTileUuid` hasn't resolved yet
- Proceeds once both tile metadata and available filters are ready (including the edge case where the query succeeds but returns no filterable fields)

Unit tests covering all five cases are included.

### DIFF
--- a/packages/frontend/src/ee/features/embed/EmbedDashboard/utils.test.ts
+++ b/packages/frontend/src/ee/features/embed/EmbedDashboard/utils.test.ts
@@ -5,7 +5,10 @@ import {
     type FilterableDimension,
 } from '@lightdash/common';
 import { describe, expect, it } from 'vitest';
-import { convertSdkFilterToDashboardFilter } from './utils';
+import {
+    convertSdkFilterToDashboardFilter,
+    shouldDeferSdkFilters,
+} from './utils';
 
 const makeDimension = (
     table: string,
@@ -122,5 +125,53 @@ describe('convertSdkFilterToDashboardFilter', () => {
                 tableName: 'payments',
             },
         });
+    });
+});
+
+describe('shouldDeferSdkFilters', () => {
+    it('defers when chart tile metadata has not loaded yet', () => {
+        // Tiles haven't arrived from POST /dashboard yet, so we cannot know
+        // whether to wait on the available-filters query or not.
+        expect(shouldDeferSdkFilters(undefined, undefined)).toBe(true);
+    });
+
+    it('proceeds when there are no filterable chart tiles', () => {
+        // Dashboard has only markdown / non-chart tiles. There is nothing
+        // for the available-filters query to fetch (it stays disabled), so
+        // there is nothing to wait for. Apply with empty tileTargets.
+        expect(shouldDeferSdkFilters([], undefined)).toBe(false);
+    });
+
+    it('defers when chart tiles exist but available filters have not resolved', () => {
+        // useDashboardsAvailableFilters reports isInitialLoading=false for
+        // disabled queries, so we must gate on the data itself, not the
+        // loading flag. Without filterableFieldsByTileUuid we cannot build
+        // cross-explore tileTargets and would silently drop the filter
+        // from tiles on other explores.
+        expect(
+            shouldDeferSdkFilters(
+                [{ tileUuid: 't1', savedChartUuid: 'c1' }],
+                undefined,
+            ),
+        ).toBe(true);
+    });
+
+    it('proceeds once chart tiles and available filters are both ready', () => {
+        expect(
+            shouldDeferSdkFilters([{ tileUuid: 't1', savedChartUuid: 'c1' }], {
+                t1: [makeDimension('payments', 'payment_method')],
+            }),
+        ).toBe(false);
+    });
+
+    it('proceeds when available filters resolved to an empty mapping', () => {
+        // Edge case: query succeeded but returned no filterable fields for
+        // any tile. We treat that as "ready" so we don't deadlock the load.
+        expect(
+            shouldDeferSdkFilters(
+                [{ tileUuid: 't1', savedChartUuid: 'c1' }],
+                {},
+            ),
+        ).toBe(false);
     });
 });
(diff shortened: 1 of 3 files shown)


pimmey:
* **#22166** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22166?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22176, opened 2026-04-21
refactor: migrate 7 single-call services to audited ability

sdolidze:
## Summary

Migrates all 7 services that had a single direct `.ability.` call site to `this.createAuditedAbility()` with type-prefixed metadata for ITGC-compliant audit logging:

| File | Subject | Metadata |
|------|---------|----------|
| `CsvService.ts` | `ExportCsv` | `dashboardUuid`, `dashboardName` |
| `DeployService.ts` | `DeployProject` | `projectUuid`, `projectName` |
| `SlackIntegrationService.ts` | `Organization` | — (converted from string subject to object form) |
| `UserAttributeUtils.ts` | `Organization` | — |
| `GitlabAppService.ts` | `Organization` | — |
| `ProjectParametersService.ts` | `Project` | `projectUuid`, `projectName` |
| `FunnelService.ts` | `SqlRunner` | `projectUuid`, `projectName` |

Promotes `no-direct-ability-check` from `warn` → `error` for all 6 fully-migrated folders (CsvService, DeployService, FunnelService, GitlabAppService, ProjectParametersService, SlackIntegrationService). `UserAttributesService/` is not promoted because `UserAttributesService.ts` still has 4 unmigrated calls.

## Notable change: `validateUserAttributeOverrides` signature

`UserAttributeUtils.ts` is a standalone utility file (not a class), so it can't call `this.createAuditedAbility()`. Refactored `validateUserAttributeOverrides` to accept a `CaslAuditWrapper<Ability>` parameter; both `McpService` call sites now create the wrapper via `this.createAuditedAbility(user)` and pass it in. Also added an explicit `organizationUuid` guard since the audited subject type requires it to be a `string`.

Two other minor behavior changes:

- `ShareService.getShareUrl` now throws `NotFoundError` when `shareUrl.organizationUuid` is missing (the left-join in `ShareModel.getSharedUrl` can produce `null` if the org row is gone). Before, this path silently ran an ability check with `undefined` — which would always deny.
- `GitlabAppService.canManageOrg` lost its `// eslint-disable-next-line class-methods-use-this` comment since the method now uses `this`.

## Test plan

- [x] `pnpm -F backend typecheck` clean
- [x] `pnpm -F backend lint` — 0 errors, only warnings on the other unmigrated services
- [x] `pnpm -F backend exec jest src/services/CsvService src/services/ShareService` — 15/15 pass, audit events visible
- [ ] Manual smoke: schedule CSV dashboard export, trigger a deploy, delete Slack installation, use header attribute overrides via MCP, connect/disconnect GitLab, list project parameters, open Funnel Builder — confirm audit events in PM2 logs for each

deploy-preview test-frontend test-backend

### DIFF
--- a/packages/backend/.eslintrc.js
+++ b/packages/backend/.eslintrc.js
@@ -102,10 +102,17 @@ module.exports = {
             files: [
                 'src/services/CatalogService/**/*.ts',
                 'src/services/ContentService/**/*.ts',
+                'src/services/CsvService/**/*.ts',
                 'src/services/DashboardService/**/*.ts',
+                'src/services/DeployService.ts',
+                'src/services/FunnelService/**/*.ts',
+                'src/services/GitlabAppService/**/*.ts',
+                'src/services/ProjectParametersService.ts',
                 'src/services/SavedChartsService/**/*.ts',
                 'src/services/SavedSqlService/**/*.ts',
                 'src/services/SchedulerService/**/*.ts',
+                'src/services/ShareService/**/*.ts',
+                'src/services/SlackIntegrationService/**/*.ts',
                 'src/services/SpaceService/**/*.ts',
             ],
             rules: {
--- a/packages/backend/src/services/ShareService/ShareService.ts
+++ b/packages/backend/src/services/ShareService/ShareService.ts
@@ -2,6 +2,7 @@ import { subject } from '@casl/ability';
 import {
     ForbiddenError,
     isUserWithOrg,
+    NotFoundError,
     SessionUser,
     ShareUrl,
 } from '@lightdash/common';
@@ -46,12 +47,17 @@ export class ShareService extends BaseService {
             throw new ForbiddenError('User is not part of an organization');
         }
         const shareUrl = await this.shareModel.getSharedUrl(nanoid);
+        if (!shareUrl.organizationUuid) {
+            throw new NotFoundError('Shared link does not exist');
+        }
 
+        const auditedAbility = this.createAuditedAbility(user);
         if (
-            user.ability.cannot(
+            auditedAbility.cannot(
                 'view',
                 subject('OrganizationMemberProfile', {
                     organizationUuid: shareUrl.organizationUuid,
+                    metadata: { shareNanoid: shareUrl.nanoid },
                 }),
             )
         ) {
--- a/packages/backend/src/services/SlackIntegrationService/SlackIntegrationService.ts
+++ b/packages/backend/src/services/SlackIntegrationService/SlackIntegrationService.ts
@@ -1,3 +1,4 @@
+import { subject } from '@casl/ability';
 import {
     ForbiddenError,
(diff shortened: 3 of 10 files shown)


packages/backend/src/services/SlackIntegrationService/SlackIntegrationService.ts
```diff
@@ -107,7 +108,13 @@ export class SlackIntegrationService<
         const organizationUuid = user?.organizationUuid;
         if (!organizationUuid) throw new ForbiddenError();
 
-        if (user.ability.cannot('manage', 'Organization')) {
+        const auditedAbility = this.createAuditedAbility(user);
+        if (
+            auditedAbility.cannot(
+                'manage',
+                subject('Organization', { organizationUuid }),
+            )
```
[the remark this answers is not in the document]
rephus: yeah, this is a bit redundant, you are checking that organizationUuid is the same organization as your user, which is always going to be true. This is a known limitation , and only acceptable becuase users can only have 1 org. 

I'll keep the code as it is

packages/backend/src/services/UserAttributesService/UserAttributeUtils.ts
```diff
@@ -30,14 +31,18 @@ export const userAttributeOverridesSchema = z
  */
 export const validateUserAttributeOverrides = (
     user: SessionUser,
+    auditedAbility: CaslAuditWrapper<Ability>,
     headerAttributes: UserAttributeValueMap,
     dbAttributes: UserAttributeValueMap,
 ): void => {
     const { organizationUuid } = user;
+    if (!organizationUuid) {
```
rephus, line 39: why was this removed ? all users should have an organization (and only 1) organizationUuid. I'd keep it for now.

packages/backend/.eslintrc.js
```diff
@@ -102,10 +102,17 @@ module.exports = {
             files: [
                 'src/services/CatalogService/**/*.ts',
                 'src/services/ContentService/**/*.ts',
+                'src/services/CsvService/**/*.ts',
```
rephus, line 105: nice

packages/backend/src/services/ShareService/ShareService.ts
```diff
@@ -46,12 +47,17 @@ export class ShareService extends BaseService {
             throw new ForbiddenError('User is not part of an organization');
         }
         const shareUrl = await this.shareModel.getSharedUrl(nanoid);
+        if (!shareUrl.organizationUuid) {
+            throw new NotFoundError('Shared link does not exist');
```
rephus, line 51: makes sense

rephus (commented):
Code looks good, only a few minor non-blocking comments.  just to doublecheck all permissions are still working, let's make sure it runs at least all Api tests

rephus:
Adding `deploy-preview test-frontend test-backend` to trigger tests

PR #22186, opened 2026-04-21
refactor: migrate 4-call services to audited ability

sdolidze:
## Summary

- Migrates `OAuthService`, `PersonalAccessTokenService`, `UserAttributesService`, and `ChangesetService` from direct `user.ability.can/cannot(...)` calls to `createAuditedAbility()` for ITGC-compliant audit logging.
- Type-prefixed metadata keys (`projectUuid`, `changeUuid`, `userAttributeUuid`, `userAttributeName`, `clientId`, etc.) are included in the subject metadata so audit logs carry enough context for reviewers.
- `OAuthService` previously used a bare `'Organization'` string subject — converted to `subject('Organization', { organizationUuid })`. Note this is a subtle semantic tightening: bare-string CASL checks match any `manage:Organization` rule regardless of conditions, whereas the conditional form now evaluates the rule's `organizationUuid` condition. For a typical org admin managing their own org this is equivalent, but cross-org admin paths (if any existed implicitly) are now denied. The `!user.organizationUuid` short-circuit is moved before the audited check to avoid passing `undefined` to the subject.
- `updateClient`/`deleteClient` also attach `metadata: { clientId }` so audit logs identify which OAuth client was touched.
- Adds the four folders/files to the ESLint `no-direct-ability-check` error list (and alphabetizes/dedupes the existing entries as a drive-by).

## Test plan

- [x] `pnpm -F backend lint` — 0 errors for the four migrated services
- [x] Manual: list / create / update / delete OAuth clients as an org admin.
- [x] Manual: create / rotate / delete a personal access token.
- [x] Manual: list / create / update / delete user attributes.
- [ ] Manual: view, revert, and revert-all changesets for a project.


packages/backend/src/services/PersonalAccessTokenService.ts
```diff
@@ -86,7 +95,19 @@ export class PersonalAccessTokenService extends BaseService {
         user: SessionUser,
         personalAccessTokenUuid: string,
     ): Promise<void> {
-        if (user.ability.cannot('delete', subject('PersonalAccessToken', {}))) {
+        const auditedAbility = this.createAuditedAbility(user);
+        if (
+            auditedAbility.cannot(
+                'delete',
+                subject('PersonalAccessToken', {
+                    organizationUuid: user.organizationUuid ?? '',
```
rephus: I don't know if this fallback would cause issues... since undefined !== ''

rephus (commented):
I'd suggest to test these features on the UI/CLI and see if they keep working, Claude can suggest some repro steps...

PR #22193, opened 2026-04-21
perf(auth): throttle session touch, enable session/creds caches by default

zuiris:
## Summary

Three small, independent quick wins on the authenticated request path. Together they deliver **+35–50% throughput** and **−25–37% latency** on a CONC=5..25 sweep of the async query pipeline (`POST /metric-query` → poll → results) against a local Jaffle Shop, with zero change to external behavior.

1. **Throttle `connect-session-knex` `store.touch()` to one UPDATE per sid per 60s.**
   Before this change, every authenticated request fired `UPDATE sessions SET expired=…` — **44.7% of DB time** in `pg_stat_statements` profiles taken against the query pipeline. With the throttle, the `expired` column trails real last-activity by at most 60s, which is well inside the session maxAge granularity (24h by default). No user-visible change to session lifetime semantics.

2. **Flip the `EXPERIMENTAL_CACHE` default for `sessionUserCache` from off to on**, and pass `useClones: false`. This skips the ~5 DB roundtrips auth middleware otherwise fires per request to resolve the `SessionUser` + CASL `Ability`. The `Ability` is immutable after `build()`, so sharing the reference across requests is safe. With the default `useClones: true`, NodeCache's `cloneDeep` of the full rule set on every hit was actively harmful — measured **−71% throughput vs cache-off at CONC=5** — which is likely why deployments never opted in. The env var is preserved as an escape hatch: set `EXPERIMENTAL_CACHE=false` to disable.

3. **Same treatment for `warehouseCredentialsCache`** — skips a DB roundtrip plus AES decrypt + `JSON.parse` on every query execute. Same `EXPERIMENTAL_CACHE=false` opt-out.

## Measured impact

Small aggregation query (`orders.status × SUM(amount)`) against local Postgres, N=200 per concurrency level, single Node process:

| CONC | p50 total      | p95 total      | throughput          |
| ---- | -------------- | -------------- | ------------------- |
| 5    | 131 → **94 ms** (−28%)  | 145 → **116 ms** (−20%) | 38 → **51 rps** (+35%)  |
| 10   | 203 → **152 ms** (−25%) | 218 → **179 ms** (−18%) | 47 → **64 rps** (+36%)  |
| 25   | 470 → **330 ms** (−30%) | 557 → **351 ms** (−37%) | 50 → **75 rps** (+50%)  |

Broader context and the next bottlenecks after this change are available on request.

## Why `useClones: false` is safe

- `SessionUser.ability` is a `@casl/ability` `Ability` instance; after `build()` it exposes only `.can()`/`.cannot()` and related read-only accessors. No code path mutates the cached object.
- `warehouseCredentials` is used as a plain read-only value by warehouse clients.
- TTL is short (30s) so even a theoretical rogue mutation would self-heal quickly.

## Test plan

- [ ] `pnpm -F backend typecheck` passes
- [ ] `pnpm -F backend lint` — no new errors
- [ ] Log in as demo user, run a query, verify results render
- [ ] Confirm session stays alive past `SESSION_TOUCH_THROTTLE_MS` (60s) without logging users out
- [ ] Confirm under concurrent load `pg_stat_statements` no longer shows `UPDATE sessions` in the top-10 consumers
- [ ] Confirm `sessionUserCache` is populated after first authenticated request and subsequent requests skip the CASL ability rebuild
- [ ] Confirm `EXPERIMENTAL_CACHE=false` still disables both caches (regression check for any deployment that relied on it being off)

## Notes for reviewers

- The `EXPERIMENTAL_CACHE` env var is preserved but its semantic is flipped: previously `=true` opted in, now any value other than `=false` keeps the cache on. Deployments that had explicitly set `EXPERIMENTAL_CACHE=true` see no behavior change; deployments that relied on it being off by default will need to set `EXPERIMENTAL_CACHE=false` to preserve old behavior.
- The `store.touch` override is intentionally a small in-memory `Map<sid, lastTouchMs>`. In a multi-process deployment each process has its own throttle window, which is fine — the worst case is one UPDATE per process per sid per 60s, which is still O(processes) not O(requests).

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -359,6 +359,22 @@ export default class App {
             sidfieldname: 'sid',
         });
 
+        const SESSION_TOUCH_THROTTLE_MS = 60_000;
+        const lastTouchBySid = new Map<string, number>();
+        const originalTouch = store.touch?.bind(store);
+        if (originalTouch) {
+            store.touch = function throttledTouch(sid, sess, fn) {
+                const now = Date.now();
+                const last = lastTouchBySid.get(sid) ?? 0;
+                if (now - last < SESSION_TOUCH_THROTTLE_MS) {
+                    if (fn) fn();
+                    return;
+                }
+                lastTouchBySid.set(sid, now);
+                originalTouch(sid, sess, fn);
+            };
+        }
+
         // Use custom middlewares if provided
         this.customExpressMiddlewares.forEach((middleware) =>
             middleware(expressApp),
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -126,14 +126,16 @@ export type ProjectModelArguments = {
 
 const CACHED_EXPLORES_PG_LOCK_NAMESPACE = 1;
 
-// Initialize cache for warehouse credentials with 30 seconds TTL
+// useClones:false — callers treat the decrypted credentials as read-only.
+// Set EXPERIMENTAL_CACHE=false to opt out.
 const warehouseCredentialsCache =
-    process.env.EXPERIMENTAL_CACHE === 'true'
-        ? new NodeCache({
-              stdTTL: 30, // time to live in seconds
-              checkperiod: 60, // cleanup interval in seconds
-          })
-        : undefined;
+    process.env.EXPERIMENTAL_CACHE === 'false'
+        ? undefined
+        : new NodeCache({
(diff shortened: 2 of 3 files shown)


tatianainama:
Thanks for the PR. We've changed our [contribution policy](https://github.com/lightdash/lightdash/blob/main/.github/CONTRIBUTING.md), so I'm closing this one. If you’re happy to share what you found, please open a [feature request](https://github.com/lightdash/lightdash/issues/new?template=feature-request.yml) and include the measurements. That’s the best way for us to gauge demand!

PR #22194, opened 2026-04-21
feat(compiler): warn when a metric name is defined in multiple sources

torimwhaley:
## Summary

When the same metric name is defined in more than one place on a single dbt model, `convertTable` was silently merging via spread/reduce — keeping only one definition with no feedback. This surfaces a `DUPLICATE_FIELD_NAME` warning naming the metric, the model, and all conflicting sources, so users get an actionable message instead of losing definitions silently.

The new check covers metric names colliding across:
- multiple columns' `meta.metrics`
- model-level `meta.metrics` and any column's `meta.metrics`
- a dbt metric and a Lightdash `meta.metrics` definition

Aggregation behavior is unchanged. The warning matches the existing metric-vs-dimension duplicate pattern (warn rather than fail). Surfaces in the UI alongside other table warnings.

Out of scope: literal duplicate YAML keys within a single `meta.metrics` dict (dbt rejects those at parse time, before the translator runs).

## Test plan

- [x] New unit test in `translator.test.ts` covering cross-column metric collision (asserts a single `revenue` metric remains and the warning names both source columns)
- [x] All 64 existing common test suites pass (1784 tests, 2 skipped)
- [x] `pnpm -F common typecheck` clean
- [x] `pnpm -F common lint` clean (only pre-existing warnings)

### DIFF
--- a/packages/common/src/compiler/translator.mock.ts
+++ b/packages/common/src/compiler/translator.mock.ts
@@ -781,6 +781,30 @@ export const MODEL_WITH_WRONG_METRICS: DbtModelNode = {
     },
 };
 
+export const MODEL_WITH_DUPLICATE_METRIC_ACROSS_COLUMNS: DbtModelNode = {
+    ...model,
+    columns: {
+        amount: {
+            name: 'amount',
+            data_type: DimensionType.NUMBER,
+            meta: {
+                metrics: {
+                    revenue: { type: MetricType.SUM },
+                },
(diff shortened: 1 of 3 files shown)


torimwhaley:
Closing — wrong layer. The duplicate metrics were generated by the Sigma → Lightdash migration tool, so the fix belongs in the migration repo, not in Lightdash core.

PR #22203, opened 2026-04-22
fix: stage uploaded images for data apps

magnew:
## Description

- Frontend now only receives an opaque `imageId` (UUID) instead of an S3 key
- Images are staged at a deterministic path (`apps/{appUuid}/uploads/{imageId}`) with no extension — MIME type stored as S3 `ContentType`
- At pipeline time, backend reconstructs the staging key from `appUuid` + `imageId`, copies to version assets folder, and writes to sandbox
- Removes `AppImageAttachment` type entirely — no S3 paths, MIME types, or filenames cross the API boundary
- No database changes — storage path is fully deterministic from values the backend already has

### DIFF
--- a/packages/backend/src/ee/controllers/appGenerateController.ts
+++ b/packages/backend/src/ee/controllers/appGenerateController.ts
@@ -52,7 +52,7 @@ export class AppGenerateController extends BaseController {
             req.user!,
             projectUuid,
             body.prompt,
-            body.image,
+            body.imageId,
             body.appUuid,
             body.chartUuids,
         );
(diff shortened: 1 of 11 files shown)


magnew:
* **#22203** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/ee/controllers/appGenerateController.ts
```diff
@@ -52,7 +52,7 @@ export class AppGenerateController extends BaseController {
             req.user!,
             projectUuid,
             body.prompt,
-            body.image,
+            body.imageId,
```
Spissable, line 55: Just to be extra safe, we should validate that the imageId is a valid uuidv4, since a user could pass arbitrary strings and do some funky path traversal potentially. Applies both here in the `generateApp` and the `iterateApp` below

Spissable (approved):

PR #22206, opened 2026-04-22
fix: prevent dropdown remount resetting scroll on chart tile selection

### LINKED ISSUE #3169: When a user selects a saved chart to add to a dashboard, we shouldn't scroll back up to the top of the list
### Description

When a user selects a saved chart to add to a dashboard, we automatically scroll back up to the top of the list. This can be frustrating because typically, you'll be scrolling through a list top-to-bottom and adding charts along the way. So, each time you add a chart, you lose your place in the list. 

https://www.loom.com/share/1d0ece7e09394f9ca8f0b27959991ec0

Proposed changes: when a user selects an item from the list, the list remains in that position (we don't scroll back to the top each time)

pimmey:
Closes: #3169

### Description:

Fixes a bug in the "Add Chart Tiles" modal where selecting a chart or triggering any re-render would reset the dropdown scroll position back to the top, making it difficult to use the "Load more" pagination when browsing large chart lists.

The root cause was that the `dropdownComponent` prop was defined inline, causing React to treat it as a new component type on every render. This unmounted and remounted the dropdown, resetting `scrollTop` to 0.

The fix extracts the dropdown into a stable `DropdownComponent` created once via `useMemo` with empty dependencies. Pagination state (`hasNextPage`, `isFetching`, `fetchNextPage`) is accessed through a ref (`paginationRef`) so the component always reads the latest values without needing to be recreated.

Additionally, after clicking "Load more", a `pendingScrollToEndRef` flag is set and consumed in a `useLayoutEffect` that fires after the new items render, automatically scrolling to the bottom so the next "Load more" button remains visible.

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
+++ b/packages/frontend/src/components/DashboardTiles/TileForms/AddChartTilesModal.tsx
@@ -21,6 +21,7 @@ import uniqBy from 'lodash/uniqBy';
 import React, {
     forwardRef,
     useEffect,
+    useLayoutEffect,
     useMemo,
     useRef,
     useState,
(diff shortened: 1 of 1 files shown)


pimmey:
* **#22206** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22206?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22206?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22207, opened 2026-04-22
feat(timeframes): add toUTC inverse to dateTruncTimezoneConversions

almeidabbm:
Relates to: https://linear.app/lightdash/issue/GLITCH-291/update-result-formatting-to-use-resolved-project-timezone

### The problem

When a project is configured with a non-UTC timezone and `EnableTimezoneSupport` is on, truncated date/timestamp columns were leaving SQL as **naive wall-clock values that looked like UTC on the wire**.

The frontend then re-shifted them — and we had two formatter special cases papering over the resulting double-shift.

### The fix

Root-cause the problem in SQL. Each warehouse adapter now wraps the `DATE_TRUNC` output back into a real UTC instant representing the project-TZ wall clock, so values stay honest end-to-end.

- **SQL** — `DATE_TRUNC` output is re-tagged to a real UTC instant per warehouse
- **Filters** — literals are tagged in the project TZ so `WHERE` clauses compare like-for-like
- **Tooltips** — the cartesian tooltip header is threaded with the project TZ
- **API** — stops reporting a timezone name when the flag is off (previously it lied about what the data actually was)

With the SQL honest, the knock-on cleanups fall out:

- Formatter collapses to a **single uniform policy** — no more DATE vs TIMESTAMP branching
- The frontend's local feature-flag gate goes away

### Behaviour matrix

| State | Before | After |
|---|---|---|
| Flag **off** | — | Byte-identical |
| Flag **on** + UTC project TZ | — | No-op |
| Flag **on** + non-UTC project TZ | Wrong buckets, wrong day east of UTC | Correct buckets across all truncatable intervals, including positive-offset TZs (Tokyo/Sydney) that today display the wrong day |

**Before - notice timestamp hour offset**
CleanShot 2026-04-22 at 11 01 47

**After**
CleanShot 2026-04-22 at 11 03 31

### Test plan

- [x] Flag off: SQL and API responses unchanged vs main
- [x] Flag on + UTC: no-op vs today's flag-on behavior
- [x] Flag on + `Asia/Tokyo` / `Pacific/Pago_Pago`: buckets match documented per-TZ counts on `timezone_test`
- [x] Filter operators (`equals`, `inBetween`, `inThePast`) return correct rows across UTC / Pago_Pago / Tokyo


almeidabbm:
* **#22234** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22234?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22207** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22207?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22207?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22218, opened 2026-04-22
fix(formula): default window function ORDER BY to query sort fields

joaoviana:
## Summary

Window functions (`LAG`, `LEAD`, `ROW_NUMBER`, `RANK`/`DENSE_RANK`, `FIRST`/`LAST`, `RUNNING_TOTAL`, `MOVING_SUM`/`MOVING_AVG`, `NTILE`) with no explicit `ORDER BY` in the formula now inherit the containing query's sort fields as a fallback. Fixes BigQuery's `Window ORDER BY is required for analytic function lag` error and makes `LAG(x)` semantically pick the row immediately above the current one in the rendered table on every warehouse.

Closes **PROD-7080**.

### Before / After

Given a query sorted by `users_user_created_ts_month ASC`, a formula table calc like `=LAG(Total users)` previously compiled to:

```sql
LAG(`users_count_user_ids`) OVER ()
-- BigQuery: Window ORDER BY is required for analytic function lag
```

It now compiles to:

```sql
LAG(`users_count_user_ids`) OVER (ORDER BY `users_user_created_ts_month` ASC)
```

Explicit clauses in the formula still win — `=LAG(Total users, PARTITION BY App version, ORDER BY User created month)` compiles unchanged. The fallback only fills in the missing `ORDER BY`, so `=LAG(Total users, PARTITION BY App version)` composes the user's partition with the injected order:

```sql
LAG(`users_count_user_ids`) OVER (
    PARTITION BY `users_app_version_latest`
    ORDER BY `users_user_created_ts_month` ASC
)
```

### Design

A new compile option on `@lightdash/formula` — `defaultOrderBy: { column: string; direction?: 'ASC' | 'DESC' }[]` — lets callers supply a fallback ORDER BY. The formula compiler uses it only when the formula's OVER clause has no explicit `ORDER BY`; an explicit one always wins. Data shape (not a callback) because this is descriptive input to the compiler, not an embedding decision the caller needs to make — same pattern as the existing `columns: Record<string, string>` option. Quoting and direction formatting stay internal to the formula package so they follow the dialect's rules consistently.

The wiring:

1. `packages/formula/src/types.ts` — new `defaultOrderBy` on `CompileOptions`.
2. `packages/formula/src/codegen/generator.ts` — `appendOverClause` emits the default ORDER BY into every window function's OVER clause when the formula didn't specify one. Flows through every dialect-specific LAG/LEAD hook (Redshift's 3-arg COALESCE wrapper, ClickHouse's `lagInFrame`) automatically — they all route back through `emitWindow → generateWindowFunction → appendOverClause`.
3. `packages/backend/src/queryCompiler.ts` — the only caller that sets `defaultOrderBy`; it passes `metricQuery.sorts`, mirroring the existing `MetricQueryBuilder.buildWindowOrderByClause` that's already used for post-calculation metrics (`RUNNING_TOTAL`, `PERCENT_OF_PREVIOUS`). Custom bin dimensions get their `_order` suffix so the window order matches the table's numeric-bin order.

### Scope notes

- Sorts on table calculations are filtered out of the default. Formula table calcs and their siblings are projected in the same `SELECT` (the `table_calculations` CTE), and referencing a sibling in an OVER clause self-references within that SELECT — every warehouse rejects it. Users sorting by a table calc still need an explicit `ORDER BY` in the formula; this PR doesn't make that case worse than today.
- `nullsFirst` on sorts is ignored in the injected clause (matches `buildWindowOrderByClause`). Follow-up if anyone hits it — ClickHouse's syntax diverges and deserves a focused change.
- No change to formulas that had explicit `ORDER BY` or to the `renderAggregate` `OVER ()` wrap for whole-result-set aggregates (intentionally unordered).

## Test plan

- [x] `pnpm -F formula test` — 149 unit tests pass, including 11 new `defaultOrderBy` cases (fallback, user-ORDER-BY-wins precedence, partition + default, multi-field, empty option, BigQuery quoting, Redshift 3-arg COALESCE wrapper, `renderAggregate` untouched, `ROW_NUMBER`/`FIRST`/`RUNNING_TOTAL`).
- [x] `pnpm formula:test:fast` — 311/311 on DuckDB (4 new end-to-end cases: `window/lag-default-order`, `window/lag-user-partition-default-order`, `window/lag-user-order-wins-over-default`, `window/row-number-default-order`).
- [x] `pnpm formula:test:tier2 --warehouse bigquery` — 311/311 on real BigQuery, end-to-end verification on the warehouse whose strict ORDER-BY requirement surfaced the bug.
- [x] `pnpm -F backend typecheck` — clean.
- [x] Manual verification on a real BigQuery project with `=ROW_NUMBER()`, `=LAG(Total users)`, and `=ROUND((A - LAG(A, PARTITION BY V)) / COALESCE(LAG(A, PARTITION BY V), 1), 2)` — compiled SQL shows the injected `ORDER BY` exactly as expected, composes correctly with explicit `PARTITION BY`.
- [x] CI tier1 + tier2 across Postgres / Redshift / Snowflake / Databricks / ClickHouse.

### DIFF
--- a/packages/backend/src/queryCompiler.ts
+++ b/packages/backend/src/queryCompiler.ts
@@ -184,6 +184,21 @@ const compileTableCalculation = (
                     columns[dep.name] = dep.name;
                 }
             }
+            // Filter out sorts on table calculations: a formula table calc
+            // and its siblings are projected in the same SELECT, so ordering
+            // by a sibling alias inside its OVER clause self-references
+            // within that SELECT and every warehouse rejects it.
+            const validFieldIdsForSort = new Set(validFieldIds);
+            const defaultOrderBy = sortFields
+                .filter((s) => validFieldIdsForSort.has(s.fieldId))
+                .map((s) => ({
+                    column: customBinDimensionIds.has(s.fieldId)
+                        ? `${s.fieldId}_order`
+                        : s.fieldId,
+                    direction: (s.descending ? 'DESC' : 'ASC') as
+                        | 'ASC'
+                        | 'DESC',
+                }));
             // Table calcs land in a post-aggregation SELECT alongside non-
             // aggregate dimension columns, so bare SQL aggregates would be
             // rejected by the warehouse. Wrapping as `AGG(x) OVER ()` turns
@@ -193,6 +208,7 @@ const compileTableCalculation = (
                 dialect,
                 columns,
                 renderAggregate: (inner) => `${inner} OVER ()`,
+                defaultOrderBy,
             });
             return {
                 ...tableCalculation,
--- a/packages/formula-tests/cases/window.cases.ts
+++ b/packages/formula-tests/cases/window.cases.ts
@@ -406,6 +406,126 @@ export const windowCases: TestCase[] = [
         tags: ['window'],
     },
 
+    // ── defaultOrderBy fallback ───────────────────────────────────
+    // These exercise the compile-time `defaultOrderBy` option, which
+    // queryCompiler.ts passes from the containing query's sort fields.
+    // BigQuery and Snowflake reject analytic functions with no ORDER BY
+    // ("Window ORDER BY is required for analytic function lag"), so these
+    // cases regress-guard the fallback that keeps `=LAG(A)` working when
+    // the user hasn't written an explicit ORDER BY in the formula.
+
+    {
+        id: 'window/lag-default-order',
+        formula: '=LAG(A)',
+        description:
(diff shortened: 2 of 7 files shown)


joaoviana:
* **#22218** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22218?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22218?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22224, opened 2026-04-22
fix: stop explore jitter when AI generates table calc formula

joaoviana:
## Summary

Two changes that together stop the Explore page behind the Table Calculation modal from visibly jittering when the AI generates a formula:

1. **Stop remounting the formula editor on AI apply.** Previously `handleFormulaAiApply` bumped a key prop on `FormulaForm` so Tiptap would re-read `initialContent`, which destroyed and rebuilt the ProseMirror view on every AI result. Replaced with an in-place `editor.commands.setContent(...)` triggered by an `initialContent` effect in `FormulaEditor`. An equality guard (`editor.getText() === initialContent`) keeps normal typing a no-op.
2. **Anchor the modal to the top (`centered: false`).** When the AI result includes a `format`, the `FormatForm` reveals additional controls and the modal body grows. With `centered: true` (Mantine's default), the modal re-centered vertically on every growth — which is what reads as "Explore jittering" behind it. With the modal anchored at the top, growth only pushes content down; the modal frame's position stays put.

Closes **PROD-7077**.

## Follow-up

The format section itself still expands/contracts when AI sets a non-default format — that shift is visible _inside_ the modal. This PR only stops it from rippling into the Explore behind. Making the format section's reveal feel smoother is worth doing separately.

## Test plan

- [ ] Manual: AI-generate a formula, including one that sets a format (currency / percent) — Explore no longer jitters; modal body grows downward without re-centering.
- [ ] Manual: type in the formula editor — text flows normally, no flicker or sync loops.
- [ ] Manual: edit an existing formula calc — formula seeds as before.
- [ ] Manual: other modals (unchanged) still center vertically.

### DIFF
--- a/packages/frontend/src/components/Explorer/FormatForm/getFormatSummary.ts
+++ b/packages/frontend/src/components/Explorer/FormatForm/getFormatSummary.ts
@@ -1,0 +1,78 @@
+import {
+    CustomFormatType,
+    findCompactConfig,
+    type CustomFormat,
+} from '@lightdash/common';
+
+export const getFormatTypeLabel = (type: CustomFormatType): string => {
+    switch (type) {
+        case CustomFormatType.BYTES_SI:
+            return 'Bytes (SI)';
+        case CustomFormatType.BYTES_IEC:
+            return 'Bytes (IEC)';
+        case CustomFormatType.DATE:
+            return 'Date';
+        case CustomFormatType.TIMESTAMP:
+            return 'Timestamp';
+        case CustomFormatType.DEFAULT:
+            return 'Default';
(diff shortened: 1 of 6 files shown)


joaoviana:
* **#22224** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22224?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22224?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22238, opened 2026-04-22
fix: upload image to s3 with correct credentials

magnew:
### Description:
Summary
  - Image uploads for data apps fail in production with SignatureDoesNotMatch (HTTP 403) from GCS
  - Root cause: AppGenerateService.getS3Client() only uses explicit accessKey/secretKey for credentials, ignoring the useCredentialsFrom credential chain that production relies on (e.g., env, container_metadata,
  instance_metadata)
  - Tarball uploads work because runPipeline() executes on the scheduler worker pod, while uploadImage() runs on the API server pod — the two pods have different credential environments
  - Extracted resolveS3Credentials() from S3BaseClient into a reusable function and wired it into AppGenerateService.getS3Client() so it resolves credentials the same way all other S3 clients do

  Why only images?

  The app generation pipeline (runPipeline) runs on the scheduler worker pod via Graphile Worker, while image uploads (uploadImage) run on the API server pod handling the HTTP request. Both use the same getS3Client()
  code, but the scheduler worker pod has credentials that happen to work with the old code path, while the API server pod requires the useCredentialsFrom credential chain.

  Changes

  - packages/backend/src/clients/Aws/S3BaseClient.ts — Extract credential resolution logic into a standalone resolveS3Credentials() function. S3BaseClient now calls it internally (pure refactor, no behavior change).
  - packages/backend/src/ee/services/AppGenerateService/AppGenerateService.ts — getS3Client() now uses resolveS3Credentials() instead of only checking explicit keys.

### DIFF
--- a/packages/backend/src/clients/Aws/S3BaseClient.ts
+++ b/packages/backend/src/clients/Aws/S3BaseClient.ts
@@ -28,6 +28,80 @@ export type S3BaseConfiguration =
     | undefined;
 
 /**
+ * Resolves S3 credentials from explicit keys, a credential chain, or SDK defaults.
+ * Returns the credentials property to assign to an S3ClientConfig, or undefined
+ * to let the SDK use its default resolution.
+ */
+export function resolveS3Credentials(config: {
+    accessKey?: string;
+    secretKey?: string;
+    useCredentialsFrom?: string[];
+}): S3ClientConfig['credentials'] | undefined {
+    if (config.accessKey && config.secretKey) {
+        Logger.debug('Using S3 storage with access key credentials');
+        return {
+            accessKeyId: config.accessKey,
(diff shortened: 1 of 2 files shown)


magnew:
* **#22238** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22238?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

magnew:
Fixed a different way, but this still might be a better approach: https://github.com/lightdash/lightdash/pull/22241/changes

The other fix buffers images to memory.

PR #22247, opened 2026-04-22
refactor: migrate GroupService to audited ability

sdolidze:
## Summary
- Replaced all direct `actor.ability.can/cannot()` calls with `auditedAbility.can/cannot()` in `GroupService`
- Added `groupUuid` and `groupName` metadata to all Group subject checks for audit logging
- Added `GroupService.ts` to the `no-direct-ability-check` error list in `.eslintrc.js`

## Test plan
- [x] `eslint` passes with 0 errors on GroupService.ts
- [ ] Verify group CRUD operations still work (create, read, update, delete members)
- [ ] Verify project group access operations work (add, remove, update)


packages/backend/src/services/GroupService.ts
```diff
@@ -61,12 +61,17 @@ export class GroupsService extends BaseService {
             throw new ForbiddenError('Group service is not enabled');
         }
 
+        const auditedAbility = this.createAuditedAbility(actor);
         const group = await this.groupsModel.getGroup(member.groupUuid);
         if (
-            actor.ability.cannot(
```
rephus, line 66: in this case, actor is a SessionUser. But in some cases, actor can be a User or a Service account (no user attached), FYI.

rephus (approved):

PR #22270, opened 2026-04-23
refactor: migrate ProjectService to audited ability

sdolidze:
## Summary

- Migrate 77 direct CASL ability checks (`user.ability.can/cannot(...)` and `account.user.ability.can/cannot(...)`) in `ProjectService` to `auditedAbility.can/cannot(...)` for ITGC-compliant audit logging.
- Attach type-prefixed metadata (`savedChartUuid`, `savedChartName`, `dashboardUuid`, `exploreName`, `spaceUuid`, `spaceName`, `jobUuid`) to CASL subjects where applicable. Omit metadata for subjects where `projectUuid`/`organizationUuid` would duplicate the subject body.
- Use explicit keys in `subject()` bodies instead of object spreads, keeping only the fields that CASL rules actually evaluate:
  - `SavedChart` / `Space` subjects: `organizationUuid`, `projectUuid`, `inheritsFromOrgOrProject`, `access` (the full `SpaceAccessContextForCasl` shape).
  - `Job` subject (view check): only `userUuid` (the sole field referenced by any `Job` CASL condition, see [organizationMemberAbility.ts:112](https://github.com/lightdash/lightdash/blob/main/packages/common/src/authorization/organizationMemberAbility.ts#L112)) plus `organizationUuid` required by `AuditableCaslSubject`.
  - Dropped the `...savedChart` spread in `getAvailableFiltersForSavedQuery{,ies}` — its fields were either shadowed by `spaceCtx` or unused by any rule.
- Use `user.organizationUuid!` non-null assertion where the audited wrapper's stricter types require `organizationUuid: string` (instead of adding early `throw new ForbiddenError()` guards).
- Add `src/services/ProjectService/**/*.ts` to the `no-direct-ability-check` ESLint error list.

## Test plan

- [x] \`pnpm -F backend lint\` — 0 errors
- [x] \`pnpm -F backend typecheck\` — passes
- [ ] No unit tests exist for ProjectService


packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -3530,12 +3547,19 @@ export class ProjectService extends BaseService {
...
         ]);
 
+        const auditedAbility = this.createAuditedAbility(account);
         if (
-            account.user.ability.cannot(
+            auditedAbility.cannot(
                 'view',
-                subject('SavedChart', spaceCtx),
+                subject('SavedChart', {
+                    ...spaceCtx,
```
rephus: should we be explicit with these fields ? 
Like: `spaceUuid: spaceCtx.spaceUuid`

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -5087,7 +5129,16 @@ export class ProjectService extends BaseService {
             ) {
                 throw new NotFoundError(`Cannot find job`);
             }
-        } else if (user.ability.cannot('view', subject('Job', job))) {
+        } else if (
+            auditedAbility.cannot(
+                'view',
+                subject('Job', {
+                    ...job,
+                    organizationUuid: user.organizationUuid!,
```
rephus, line 5143: like in previous PRs, this is redundant, but let's keep it for now.

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -6003,12 +6066,17 @@ export class ProjectService extends BaseService {
...
                     );
 
+                const auditedAbility = this.createAuditedAbility(account);
                 if (
-                    account.user.ability.cannot(
+                    auditedAbility.cannot(
                         'view',
                         subject('SavedChart', {
                             ...savedChart,
                             ...spaceCtx,
```
rephus: I believe this should be explicit, and use the right  fields like`projectUuid` rather than adding some objects that might return undefined fields.

rephus (approved):
Overall looks good, we could consider being more explicit when passing fields to auditLogs, rather than passing the entire object. But I'll leave that to you.

PR #22298, opened 2026-04-23
docs: custom sql field materialization design note

notgiorgi:
Closes:

### Description:

Adds a design document outlining how pre-aggregates can support derived dimensions and metrics defined with `sql:`.

The document covers:

- The distinction between **query-invariant** fields (whose values are fully determined by row data) and **query-dependent** fields (whose values change based on runtime context such as parameters or user attributes)
- Which field types and expressions are unsupported for pre-aggregate matching (e.g. `count_distinct`, window functions, volatile functions, subqueries) and should fall back to warehouse execution
- Four materialization strategies:
  - Materializing derived dimension values directly (query-invariant dimensions)
  - Materializing dependency columns instead of the derived value (query-dependent dimensions)
  - Materializing metrics directly (query-invariant additive metrics)
  - Materializing metric components for recomputation at query time (averages, query-dependent metrics)
- How stored metadata (original SQL, dependency mappings, runtime dependency list) enables correct query-time rewrites for parameterized fields
- How pre-aggregate matching should verify that enough information was materialized to correctly evaluate a requested derived field
- How queries against pre-aggregates should rewrite query-dependent expressions (e.g. substituting resolved parameter values into CASE expressions) and translate source-warehouse SQL into DuckDB-compatible SQL when needed
- Concrete SQL examples for each strategy covering dimensions, additive metrics, average-like metrics, and query-dependent metrics

### DIFF
--- a/docs/preagg-derived-sql-materialization.md
+++ b/docs/preagg-derived-sql-materialization.md
@@ -1,0 +1,538 @@
+# Pre-Aggregates for Derived `sql:` Fields
+
+This note describes how pre-aggregates could support derived dimensions and metrics defined with `sql:`.
+
+## Definitions
+
+Before anything else, split derived `sql:` fields into two categories.
+
+### Query-invariant
+
+A field is **query-invariant** when its value is fully determined by row data already in the warehouse and does not depend on runtime context.
+
+Examples:
+
+```yaml
+order_value_bucket:
+  type: string
+  sql: CASE WHEN ${amount} > 100 THEN 'high' ELSE 'low' END
+```
+
+```yaml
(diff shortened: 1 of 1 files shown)


notgiorgi:
* **#22298** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22298?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22321, opened 2026-04-23
feat(content): surface verification on Space overview

torimwhaley:
## Summary

The verified badge appears on charts/dashboards everywhere except the Space overview page. PR #20943 added the indicator to tiles, resource browser, search, and omnibar — this fills the remaining gap.

The frontend already renders `ResourceVerifiedIndicator` in `InfiniteResourceTableColumnName` when `item.data.verification` is populated. The Space page uses `InfiniteResourceTable` → which reads from `/api/v2/content` → which runs through `ContentModel`. The gap was in the backend: `ContentConfiguration`s hardcoded \`verification: null\`.

This left-joins \`content_verification\` (and the verifying user) into the dashboard, dbt-explore chart, and SQL chart configurations, and extends \`SummaryContentRow\` with four new verification columns so all five UNION branches share the same shape (\`space\` and \`data_app\` configs emit NULLs).

Closes PROD-7090.

## Test plan

- [x] Backend typecheck, lint, \`test:dev:nowatch\` (138 tests)
- [x] Verified a chart via the chart menu → Space overview shows the green check indicator on its icon; API response includes \`verification: { verifiedBy, verifiedAt }\`
- [x] Non-verified charts on the same page return \`verification: null\` and render no badge

### DIFF
--- a/packages/backend/src/models/ContentModel/ContentConfigurations/SqlChartContentConfiguration.ts
+++ b/packages/backend/src/models/ContentModel/ContentConfigurations/SqlChartContentConfiguration.ts
@@ -5,6 +5,7 @@ import {
     ContentType,
 } from '@lightdash/common';
 import { Knex } from 'knex';
+import { ContentVerificationTableName } from '../../../database/entities/contentVerification';
 import { DashboardsTableName } from '../../../database/entities/dashboards';
 import { OrganizationTableName } from '../../../database/entities/organizations';
 import { ProjectTableName } from '../../../database/entities/projects';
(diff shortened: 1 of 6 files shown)


torimwhaley:
* **#22321** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22321?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22321?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/models/ContentModel/ContentConfigurations/SqlChartContentConfiguration.ts
```diff
@@ -236,7 +260,17 @@ export const sqlChartContentConfiguration: ContentConfiguration<SelectSavedSql>
...
                 views: value.views,
                 firstViewedAt: value.first_viewed_at,
-                verification: null,
+                verification: value.verified_at
+                    ? {
+                          verifiedBy: {
+                              userUuid: value.verified_by_user_uuid ?? '',
+                              firstName:
+                                  value.verified_by_user_first_name ?? '',
+                              lastName: value.verified_by_user_last_name ?? '',
```
ZeRego: @torimwhaley can we have it default to NULL rather than empty strings? Same for the other files.

torimwhaley: Done in the force-push. I went with returning the whole `verification` as `null` when user-join data is missing (rather than changing the public `ContentVerificationInfo` type), since the frontend renders `verifiedBy.firstName`/`lastName` directly via template strings — making those fields nullable would propagate to several detail-page renderings that would then read "Verified by null null". The empty-string fallbacks are gone in all three configs (Sql / Dashboard / DbtExploreChart).

PR #22345, opened 2026-04-24
fix: make feature flag `default_enabled` nullable

rephus:
Closes:

### Description:
Makes the `default_enabled` column on the `feature_flags` table nullable. When `default_enabled` is `null`, the feature flag evaluation returns `null` instead of a flag object, allowing callers to distinguish between a flag that is explicitly disabled and one that has no default set.

The migration handles rollback by setting any `null` values back to `false` before restoring the `NOT NULL` constraint.

### DIFF
--- a/packages/backend/src/database/entities/featureFlags.ts
+++ b/packages/backend/src/database/entities/featureFlags.ts
@@ -5,7 +5,7 @@ export const FeatureFlagOverridesTableName = 'feature_flag_overrides';
 
 export type DbFeatureFlag = {
     flag_id: string;
-    default_enabled: boolean;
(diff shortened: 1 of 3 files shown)


rephus:
* **#22345** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22345?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22345?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22357, opened 2026-04-24
fix: delegate screenshot readiness to chart-level callbacks in Explorer

pimmey:
Closes:

### Description:

Previously, the screenshot readiness signal for Explore pages (used by Slack unfurls) was determined at the container level by watching query loading state. This created a race condition where the container-level gate could fire before the chart had a chance to call `setFetchAll(true)` or complete chart-type-specific work (e.g. map tile loading).

The readiness signal is now driven by the chart itself via two new callbacks — `onScreenshotReady` and `onScreenshotError` — passed down through `VisualizationCard`. The `Explorer` component listens to these callbacks and flips `isScreenshotReady` only when the chart signals back, ensuring all chart-level work is complete before the screenshot is taken.

A fallback `useEffect` is retained to handle the case where a query error prevents any chart from mounting, meaning neither callback would ever fire. In that case, the error state is captured directly so the unfurl still reflects the correct outcome.

### DIFF
--- a/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationCard.tsx
+++ b/packages/frontend/src/components/Explorer/VisualizationCard/VisualizationCard.tsx
@@ -68,9 +68,16 @@ export type EchartsClickEvent = {
 
 type Props = {
     projectUuid?: string;
+    onScreenshotReady?: () => void;
+    onScreenshotError?: () => void;
 };
 
-const VisualizationCard: FC<Props> = memo(({ projectUuid: fallBackUUid }) => {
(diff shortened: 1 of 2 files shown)


pimmey:
* **#22357** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22371, opened 2026-04-27
fix: prevent preview projects from being set as default in `set-project`

rephus:
Closes: [https://linear.app/lightdash/issue/PROD-7126/cli-shows-the-same-preview-project-as-both-preview-and-production-in](https://linear.app/lightdash/issue/PROD-7126/cli-shows-the-same-preview-project-as-both-preview-and-production-in)

Before

set-project --uuid \<preview-uuid>:  silently overwrites context.project with the preview UUID

After

### Description:

Prevents preview projects from being stored as the default project via `set-project`, which previously caused the download prompt to display the same project under both "Preview" and "Production" labels.

**`setProject`:**

- Preview projects are now filtered out of the interactive selection list.
- When using `--uuid` or `--name`, an error is thrown if the matched project is a preview project.
- In non-interactive mode, the first non-preview project is selected instead of the first project overall.

**`selectProject`:**

- Adds a guard for misconfigured state where the stored main project UUID and preview project UUID are identical (produced by older versions of `set-project`). In this case, the project is treated as a preview only, avoiding the duplicate label issue.

Tests are added for both handlers covering the new filtering/validation behaviour and the misconfigured UUID edge case.

### DIFF
--- a/packages/cli/src/handlers/selectProject.test.ts
+++ b/packages/cli/src/handlers/selectProject.test.ts
@@ -1,0 +1,75 @@
+import { ProjectType } from '@lightdash/common';
+import { Config } from '../config';
+import { lightdashApi } from './dbt/apiClient';
+import { selectProject } from './selectProject';
+
+jest.mock('inquirer');
+jest.mock('../analytics/analytics');
+jest.mock('./dbt/apiClient', () => ({
+    lightdashApi: jest.fn(),
+}));
+jest.mock('../config', () => ({
+    unsetPreviewProject: jest.fn().mockResolvedValue(undefined),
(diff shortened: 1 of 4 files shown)


rephus:
* **#22371** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22371?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22371?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22376, opened 2026-04-27
fix: calculate total using cached results in explore edit

### LINKED ISSUE #22332: [Bug] Column totals serve stale cached values in explores after underlying data changes
### Bug description

Column totals in a results table show stale values after the underlying warehouse data is updated (for example, after a `dbt run` on the model). Row values refresh correctly, but the totals row continues to display values from the previous query execution. Recreating the chart, toggling "Show column total" off/on, or dropping and re-adding metrics/dimensions does not fix it. Reproduces across browsers, so this is a backend cache issue, not React Query.

### Steps to reproduce

1. Open a project with results caching enabled (the bug is particularly easy to hit in a preview project).
2. Build a table chart in the explore view with "Show column total" enabled on a metric.
3. Run the query and note the totals value.
4. Update the underlying data in the warehouse (for example, `dbt run` to refresh the model with new rows).
5. Re-run the explore query.
6. Row values update to reflect the new data, but the column total still shows the value from step 3.

### Expected behavior

Column totals should reflect the currently displayed rows after the explore is re-run.

### Actual behavior

Column totals return stale values from the backend results cache.

### Workaround

Add any dimension filter in the explore (even one with no filter clause). This changes the generated SQL, producing a new cache key and forcing a cache miss.

### Root cause

The `/calculate-total` endpoint hits the shared results cache, and there is no way to invalidate it from the explore view:

1. `getCalculateTotalMetricQuery` (`AsyncQueryService.ts:5352`) strips `dimensions`, `sorts`, and `tableCalculations` from the query. For a data-only model change, the resulting SQL is byte-identical before and after, so the cache key is identical.
2. `calculateTotalFromQuery` (`AsyncQueryService.ts:5913`) — the explore path — does not forward `invalidateCache` to `calculateMetricQueryTotal`. Compare with `calculateTotalFromSavedChart` (`AsyncQueryService.ts:5792`), which does forward it. This asymmetry means explore totals have no escape hatch.
3. `findResultsCache` (`AsyncQueryService.ts:451`) does not exclude `QueryExecutionContext.CALCULATE_TOTAL` or explore queries, so stale entries are returned.
4. The main explore query is unaffected when dimensions or sorts are modified while debugging, which changes the SQL and produces fresh cache keys. Totals are immune to dimension changes.

### Suggested fix

Thread `invalidateCache` through `calculateTotalFromQuery` → `calculateMetricQueryTotal` so the explore "Run query" action can force a refresh of totals, matching the saved-chart path. A follow-up consideration: whether `QueryExecutionContext.CALCULATE_TOTAL` should bypass the results cache entirely in explore contexts, given how often totals SQL collides after data-only warehouse updates.

almeidabbm:
Closes: https://github.com/lightdash/lightdash/issues/22332

### Description:

Propagates the `invalidateCache` flag through the calculate-total flow so that cache invalidation behaviour is consistent between the main query and the totals row.

Previously, the `calculateTotal` endpoint always ignored the caller's cache preference. Now `invalidateCache` is included in the `CalculateTotalFromQuery` type and forwarded through `useCalculateTotal`, `useTableConfig`, `useColumns`, and both the `AsyncQueryService` and `EmbedService` backend handlers.

The `buildQueryArgs` logic was also corrected so that edit mode always forces `invalidateCache: true` (re-running a query in the explorer is an explicit user action that should bypass the cache), while view mode respects the `minimal` flag as before. The hardcoded `invalidateCache: true` that was previously applied unconditionally to all explore queries has been removed in favour of this computed value flowing from `buildQueryArgs`.

### How to test
1. Enable `RESULTS_CACHE_ENABLED`
2. Open explore
2. Select 1 dimension and 1 metric, choose table viz
3. Add column totals
4. Check the table data
5. Mutate the data and call `dbt run`
6. Check the table data again

**Data mutation**
```diff
diff --git a/examples/full-jaffle-shop-demo/dbt/models/payments.sql b/examples/full-jaffle-shop-demo/dbt/models/payments.sql
index cbe6a96b76..d556f59530 100755
--- a/examples/full-jaffle-shop-demo/dbt/models/payments.sql
+++ b/examples/full-jaffle-shop-demo/dbt/models/payments.sql
@@ -1 +1 @@
-select * from {{ ref('stg_payments') }}
+select * from {{ ref('stg_payments') }} where amount > 10
```

**Before**

Initial query
before_data_change

After changing data - notice total is the same even though the rows have changed
after_data_change

**After**

Initial query
before_data_change

After changing data
CleanShot 2026-04-27 at 12 26 41

### DIFF
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -1617,6 +1617,7 @@ export class EmbedService extends BaseService {
                     query_context: QueryExecutionContext.CALCULATE_TOTAL,
                 },
                 parameters: combinedParameters,
+                invalidateCache: data.invalidateCache,
                 userAccessControls: {
                     userAttributes,
                     intrinsicUserAttributes,
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -6802,6 +6802,7 @@ const models: TsoaRoute.Models = {
         type: {
             dataType: 'nestedObjectLiteral',
             nestedProperties: {
+                invalidateCache: { dataType: 'boolean' },
                 parameters: { ref: 'ParametersValuesMap' },
                 explore: { dataType: 'string', required: true },
                 metricQuery: { ref: 'MetricQueryRequest', required: true },
@@ -8747,11 +8748,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
@@ -9267,11 +9268,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['success'],
+                                                            enums: ['error'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['error'],
+                                                            enums: ['success'],
                                                         },
                                                     ],
                                                     required: true,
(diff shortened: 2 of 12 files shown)


almeidabbm:
* **#22376** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22376?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #20164, opened 2026-02-10
feat: improver tooltip hover behavior on mixed charts

joaoviana:
Closes: [PROD-2599: Bars intermittently disappear on hover in mixed bar + line chart](https://linear.app/lightdash/issue/PROD-2599/bars-intermittently-disappear-on-hover-in-mixed-bar-line-chart)

### Description:
Fixed tooltip flickering in mixed charts by implementing a state tracking mechanism that prevents redundant setOption calls. The solution:

1. Added a ref to track when item-tooltip mode is active
2. Added debouncing for mouseOut events to prevent rapid tooltip mode switching
3. Implemented cleanup for pending timeouts to avoid race conditions when moving between chart elements quickly
4. Removed unnecessary emphasis state changes that were contributing to the flickering

This change significantly improves the user experience when hovering over chart elements in mixed chart types.

PR #22384, opened 2026-04-27
fix: clear pending highlight timer and dispatch downplay on chart mouseout

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7121/grouped-and-stacked-bar-charts-focus-on-a-single-series-value-gets

Before

CleanShot 2026-04-27 at 12 46 49

After

can't reproduce

 Where the bug is: packages/frontend/src/components/SimpleChart/index.tsx, in handleOnMouseOver (line 419).

  The mouseover handler explicitly dispatches eCharts.dispatchAction({ type: 'highlight', seriesIndex }) via a 100ms setTimeout (it's needed to keep
  the hover focus visible after switching the tooltip from axis mode to item mode). But there are two race conditions:

  1. The pending highlight fires after mouseout. If you hover briefly (<100ms) and leave, the setTimeout is never cancelled — it fires the highlight
  AFTER the cursor has already left.
  2. No mirror downplay. Manually-dispatched highlight bypasses ECharts' automatic mouseout downplay. So even when the highlight fires while the
  cursor is on the chart and you leave, nothing tells ECharts "release the focus".

  This was a regression from PR #20164 (commit 19123ec162, Feb 2026) — the previous code had emphasis: { disabled: true } on mouseout that cleared
  this. That PR removed it to fix tooltip flicker on mixed charts but didn't replace the highlight-clearing path.

  My fix in handleOnMouseOut:
  1. Cancel the pending highlight setTimeout (kills race #1).
  2. Dispatch { type: 'downplay' } (kills race #2).
  
---

## # Description:

Fixes a bug where a chart series would remain in a highlighted/blurred state after the cursor leaves the chart area.

When a `highlight` action is manually dispatched via `setTimeout` in the mouseover handler, ECharts' automatic `downplay` on mouseout no longer fires correctly, leaving the focused series visually stuck in an emphasized state.

To fix this:
- A `highlightTimer` ref is introduced to track the pending `highlight` dispatch. If the cursor leaves before the timer fires, the pending highlight is cancelled so it never applies.
- On mouseout, a `downplay` action is explicitly dispatched to clear any emphasis state that was already applied by a previous `highlight` dispatch.

### DIFF
--- a/packages/frontend/src/components/SimpleChart/index.tsx
+++ b/packages/frontend/src/components/SimpleChart/index.tsx
@@ -286,6 +286,10 @@ const SimpleChart: FC<SimpleChartProps> = memo(
         const mouseOverTimer = useRef<
             ReturnType<typeof setTimeout> | undefined
         >(undefined);
+        const highlightTimer = useRef<
+            ReturnType<typeof setTimeout> | undefined
+        >(undefined);
+        const hasDispatchedHighlight = useRef(false);
 
         const handleOnMouseOver = useCallback(
             (params: any) => {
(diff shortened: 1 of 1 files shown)


rephus:
* **#22384** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
Tested with mixed chart to ensure there are no regressions https://github.com/lightdash/lightdash/pull/20164

PR #22395, opened 2026-04-27
feat(audit): emit authentication audit events for ITGC compliance

sdolidze:
## Summary

Adds authentication audit logging at the auth boundary (login, logout, impersonation, per-request auth failures) so that successful and rejected access attempts are captured for ITGC compliance. These events sit outside the existing CASL audit wrapper because authentication happens before any CASL permission check.

- **Session lifecycle (success + failure):** password and OpenID login emit \`login\`/Session events; \`/logout\` emits \`logout\`/Session.
- **Per-request auth (failures only):** PAT, service-account bearer, and embed JWT failures emit \`login\` denied events. Successes are not duplicated — the CASL audit wrapper already records the resource access.
- **Impersonation:** \`startImpersonation\`/\`stopImpersonation\` emit \`impersonation_start\`/\`impersonation_stop\` events including denial reasons (feature disabled, not session auth, nested, self, no perms).

Wires \`{ ip, userAgent }\` from each entry point through to the service so the events land with request context.

## Implementation notes

- New helper \`createUnknownAuthActor()\` (in \`logging/auditLog.ts\`) builds the actor for failed-login cases where we know the email but not the user.
- \`UserService\` gets a small private \`emitAuthAuditEvent()\` helper that wraps \`logAuditEvent\` with try/catch so an audit-write failure can never break login.
- All passport strategies updated to forward IP/user-agent. \`localPassportStrategy\` now uses \`passReqToCallback: true\` (already supported by \`@types/passport-local\`).

## Test plan

- [x] \`pnpm -F backend typecheck\` clean
- [x] \`pnpm -F backend lint\` clean
- [x] \`pnpm -F backend test:dev:nowatch\` — 1073 tests pass; new tests cover allowed/denied login events for password and OpenID, and denied PAT auth
- [ ] Manual smoke: log in with password, log in with Google/Okta, log out, fail login with wrong password — confirm audit log entries
- [ ] Manual smoke: start/stop impersonation — confirm allowed and denied events
- [ ] Manual smoke: hit an endpoint with a bogus PAT / Bearer token / embed JWT — confirm denied events

### DIFF
--- a/packages/backend/src/controllers/authentication/strategies/apiKeyStrategy.ts
+++ b/packages/backend/src/controllers/authentication/strategies/apiKeyStrategy.ts
@@ -12,10 +12,17 @@ export const apiKeyPassportStrategy = ({
     new HeaderAPIKeyStrategy(
         { header: 'Authorization', prefix: 'ApiKey ' },
         true,
-        async (token, done) => {
+        async (token, done, req) => {
             try {
-                const user =
-                    await userService.loginWithPersonalAccessToken(token);
+                const user = await userService.loginWithPersonalAccessToken(
+                    token,
+                    req
+                        ? {
+                              ip: req.ip,
+                              userAgent: req.get('user-agent'),
+                          }
+                        : undefined,
+                );
                 return done(null, user);
             } catch {
                 return done(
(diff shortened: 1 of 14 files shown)


sdolidze:
Splitting into two stacked PRs for easier review.

ZeRego (approved):

PR #22400, opened 2026-04-27
fix: swallow DB errors in `FeatureFlagModel` for non-UUID userUuids

pimmey:
### Description:

Fixes a regression affecting embed accounts where a non-UUID `userUuid` (e.g. `external::…`) causes Postgres to throw a `22P02` invalid UUID syntax error during feature flag database lookups. Previously, this would result in a 500 error rather than gracefully resolving the flag as disabled.

The fix extracts the database error handling into a shared `tryGetFromDatabase` helper that catches any database errors, logs a warning, and returns `null` (resolving to disabled) instead of propagating the exception. This replaces the previously inconsistent error handling where some code paths had try/catch and others did not.

Tests have been added to verify that `EnableTimezoneSupport` and `EnableDataApps` flag lookups both return `enabled: false` rather than throwing when the database layer fails.

### DIFF
--- a/packages/backend/src/models/FeatureFlagModel/FeatureFlagModel.test.ts
+++ b/packages/backend/src/models/FeatureFlagModel/FeatureFlagModel.test.ts
@@ -7,9 +7,17 @@ import { FeatureFlagModel } from './FeatureFlagModel';
 // Minimal stub — tests below don't exercise the database layer
 const databaseStub = {} as Knex;
 
-const buildModel = (configOverrides: Partial<LightdashConfig> = {}) =>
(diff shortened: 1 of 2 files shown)


pimmey:
* **#22400** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22400?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22400?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22405, opened 2026-04-27
fix: adopt SCIM orphan users instead of wiping openid_identity on 409

### LINKED ISSUE #22402: SCIM `createUser` corrupts orphan users on conflict, causing 409 retry loops
### Summary

  When SCIM `POST /api/v1/scim/v2/Users` is called for an email that already exists in the database without an organization membership (an "orphan"
  user), the request fails with `409 uniqueness` **and** permanently deletes the user's `openid_identity` rows. Because Okta retries indefinitely
  (~every 30 min), the user is locked out of SSO and the loop never recovers on its own.

  ### Reproduction

  1. A user logs in via OIDC JIT but does not meet the auto-join conditions in `UserService.ts:885–893` (requires `groups.enabled && enableGroupSync
  && groups.length > 0 && allowedOrgs.length === 1`). Result: a user row + verified email row + openid_identity row, but **no
  `organization_memberships` row**.
  2. SCIM is enabled / Okta later attempts to provision the same email via `POST /Users`.
  3. `ScimService.createUser` is called.

.

### version

_No response_

### Cloud or self-hosting

None

rephus:
Closes:  https://github.com/lightdash/lightdash/issues/22402

Before
CleanShot 2026-04-27 at 16 19 01

After
CleanShot 2026-04-27 at 16 30 20

### Description:

Fixes a bug where SCIM `POST /Users` would permanently wipe a user's `openid_identity` rows when the request resulted in a 409 conflict, breaking SSO for affected users.

The root cause was that `deleteOpenIdIdentitiesForUser` was called before `userModel.createUser`, with no transaction wrapping both operations. If `createUser` threw an `AlreadyExistsError` (e.g. for a user with a verified email but no organization membership — an "orphan" created by OIDC JIT when auto-join conditions failed), the OpenID identity deletion was not rolled back. Okta would then retry the SCIM request indefinitely, never succeeding, while the user could no longer log in via SSO.

The fix introduces an upfront lookup via `findUserByEmail` before any destructive operations:

- If the email belongs to a user **with** an existing organization membership, a 409 is returned immediately — no OpenID identities are touched.
- If the email belongs to an **orphan** user (verified email, no organization membership), the user is adopted into the SCIM-provisioned organization directly, mirroring the behavior of `UserService.createPendingUserAndInviteLink`. No new user is created and no OpenID identities are deleted.
- Only for genuinely new emails does the original flow (delete stale OpenID identities, then create user) proceed.

Tests are added covering both the orphan adoption path and the 409-without-OpenID-wipe path. A reproduction script is also included in `examples/api/scim.http` with seeding instructions to verify the bug locally.

### DIFF
--- a/examples/api/scim.http
+++ b/examples/api/scim.http
@@ -34,6 +34,50 @@ Authorization: Bearer scim_9d95455ef43b7dd9077d9001e1c5b112
 }
 
 
+### --- BUG REPRO: SCIM 409 wipes openid_identity for orphan users ---
+# An "orphan" is a user with verified email + openid_identity but NO organization_membership
+# (e.g. created by OIDC JIT when auto-join conditions failed).
+# When SCIM POST /Users runs for that email, ScimService.createUser deletes the openid_identity
+# row outside any transaction (ScimService.ts:470), then userModel.createUser throws
+# AlreadyExistsError on the global email check (UserModel.ts:756). The 409 returns to Okta
+# but the openid wipe is NOT rolled back. Okta retries forever.
+#
+# Steps to reproduce locally:
+#   1) Pick an email below, e.g. orphan-repro@example.com
+#   2) Seed the orphan in psql:
+#        BEGIN;
+#        INSERT INTO users (first_name, last_name, is_active)
+#          VALUES ('Orphan', 'Repro', true) RETURNING user_id \gset
+#        INSERT INTO emails (user_id, email, is_primary, is_verified)
+#          VALUES (:'user_id', 'orphan-repro@example.com', true, true);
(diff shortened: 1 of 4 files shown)


rephus:
* **#22405** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22405?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22405?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22081, opened 2026-04-17
fix: sort-only table calculations appearing as chart series (#21965)

owlas:
## Summary

- Adds a dedicated `sortOnlyColumns` field to `PivotConfiguration` to separate metrics needed for SQL sort anchor CTEs from metrics that should be displayed as chart series
- `PivotQueryBuilder.toSql()` merges both lists for SQL generation, but only `valuesColumns` flows through to `pivotDetails` in the API response
- Removes the `visibleMetricFieldIds` workaround in `getCartesianPivotConfig` that was added as a downstream patch for this issue

## Context

PR #21883 fixed pivot chart sorting when the sort field isn't on the Y-axis by adding sort-only metrics to `valuesColumns`. But `valuesColumns` serves double duty — it's both the list for SQL generation AND the list that flows to the frontend as chart series. This caused sort-only metrics to render as unwanted chart series.

## Test plan

- [x] `pnpm -F common typecheck` passes
- [x] `pnpm -F backend typecheck` passes (only pre-existing formula errors)
- [x] `pnpm -F common test -- --testPathPattern=pivot` — all 64 tests pass
- [ ] Manual: create a pivoted cartesian chart with a table calc sort field — the table calc should NOT appear as a series, sort should still work

Issue #10229, opened 2024-05-29
We should validate references in custom SQL dimensions


TuringLovesDeathMetal:
### Description

Right now, custom SQL dimensions are not validated on `lightdash validate`. We should make sure these are included in validation.

PR #22421, opened 2026-04-27
fix: cartesian x-axis collapsing when table calculation is also a sort field

### LINKED ISSUE #22389: Cartesian chart x-axis only renders first value when x-axis is a table calculation that is also a sort field
## Symptom

When a Cartesian chart uses a **table calculation** as the **x-axis field**, and that same table calculation is **also used to sort the results**, the chart only renders the first distinct x-axis value. All other rows collapse into a single point/bar even though the underlying query returns multiple distinct values for that field.

Switching the x-axis to the raw dimension that the table calc is built from (or sorting by a different field) restores the expected chart with all distinct x-axis values.

## Reproduction

1. Create a saved chart (Cartesian: bar/line/etc.) with a query containing:
   * At least one dimension
   * At least one metric (used on y-axis)
   * A string-typed table calculation derived from the dimension (e.g. a `CASE` or formatted label of the raw dimension)
2. Set the table calculation as the **x-axis field**.
3. Sort the results by that same table calculation.
4. Run the chart.

Expected: chart shows one x-axis tick per distinct table-calc value.<br>Actual: chart shows only the first table-calc value with data; remaining values are missing from the rendered series.

If you change the sort to reference the raw dimension instead of the table calc (keeping the table calc on the x-axis), the chart renders correctly. Likewise, if you keep the sort on the table calc but switch the x-axis to the raw dimension, the chart also renders correctly. Both conditions together (table calc as x-axis AND as sort field) are required to trigger the bug.

## Scope

* Affects Cartesian charts (bar, line, etc.) that pivot via `derivePivotConfigFromChart`.
* Field type of the table calc does not matter — confirmed with `string`-typed table calculations.
* Reproduces when the table calc is just a relabelling of an existing dimension, but is not limited to that case.

## Regression source

This is a regression introduced by PR lightdash/lightdash#22081 ("fix: sort-only table calculations appearing as chart series", merged 2026-04-17, released in v0.2761.4), which fixed issue [PROD-6952](https://linear.app/lightdash/issue/PROD-6952/sort-only-table-calculation-appears-as-unwanted-chart-series-in).

That PR added a `sortOnlyMetrics` collection step to `packages/common/src/pivot/derivePivotConfigFromChart.ts` so that sort-only metrics/table calcs are surfaced to `PivotQueryBuilder` rather than being silently dropped. Before this PR, sorting by the x-axis table calc worked fine.

## Where the problem is

`packages/common/src/pivot/derivePivotConfigFromChart.ts`

Inside `getCartesianPivotConfiguration`, the `sortOnlyMetrics` filter (around lines 284-296) collects every sort field that is a metric or a table calculation and is not already in `valuesColumns`. It does **not** exclude the configured `xField`.

```ts
const sortOnlyMetrics = metricQuery.sorts
    .filter(
        (sort) =>
            !valuesRefs.has(sort.fieldId) &&
            (metricQuery.metrics.includes(sort.fieldId) ||
                (metricQuery.tableCalculations || []).some(
                    (tc) => tc.name === sort.fieldId,
                )),
    )
    .map((sort) => ({
        reference: sort.fieldId,
    }));
```

When the x-axis table calc is also a sort field, it is captured by this filter and merged into `allValuesColumns` (line 300), which is then passed to `getIndexColumn` as the `valuesColumns` argument:

```ts
const allValuesColumns = [...valuesColumns, ...sortOnlyMetrics];
const indexColumn = getIndexColumn(
    groupByColumns,
    allValuesColumns,
    fields,
    metricQuery,
    xField,
);
```

`getIndexColumn` (lines \~107-115) then filters out any name present in `valuesColumnsReferences`, which now includes the x-axis table calc:

```ts
const indexColumnNames = [
    ...metricQuery.dimensions,
    ...metricQuery.metrics,
    ...tableCalcNames,
].filter(
    (dim) =>
        !groupByColumnsReferences.includes(dim) &&
        !valuesColumnsReferences.includes(dim),
);
```

Net effect: the x-axis table calc never makes it into the pivot's `indexColumn`, so the resulting `PivotConfiguration` has no x-axis dimension for that field. Downstream, the chart series collapses onto a single index value.

The raw dimension does not hit this bug because the `sortOnlyMetrics` filter only considers metrics and table calculations, not dimensions.

## Why this matters

* This is a silent rendering bug. The query runs successfully and the results table shows all rows correctly — only the chart is wrong, which makes it easy to miss and easy to mistrust.
* Using a table calc to derive a display label for an x-axis (and sorting by that label to control axis order) is a common pattern.

## Notes for whoever picks this up

* Any fix must preserve the behaviour from PR lightdash/lightdash#22081 / issue [PROD-6952](https://linear.app/lightdash/issue/PROD-6952/sort-only-table-calculation-appears-as-unwanted-chart-series-in): a table calc that is **only** used for sorting (not on x-axis, not in y-axis values, not as a pivot column) must still be passed through to `PivotQueryBuilder` so its sort is honoured and an anchor CTE can be generated.
* Regression test should cover all four combinations: (table calc on x-axis, sort by table calc), (table calc on x-axis, sort by other), (sort-only table calc not on chart), (table calc on y-axis as value, also sort).
* Worth checking whether the same shape of bug exists for **metrics** used as the x-axis on scatter charts — the `sortOnlyMetrics` filter captures metrics too, and `getIndexColumn` only re-admits a metric as an index column when `dim === xField` (line 156). The same `valuesColumnsReferences` exclusion may strip a sort-anchored metric x-axis before that branch can run.

charliedowler:
## Summary

Fixes #22389 — Cartesian charts with a table calculation on the x-axis that
is also used as a sort field were collapsing onto a single x-axis tick. This
was a regression introduced by #22081, which over-broadly captured every
sort field as a "sort-only" value column.

- Preserve pivot index fields when the same metric/table calculation is also
  used for sorting (the #22389 fix).
- Keep sort-only metrics/table calculations available for ordering without
  leaking them into chart or export output (preserves #22081's intent).
- Avoid false unused-dimension validator warnings when a chart-used table
  calculation references hidden dimensions, including window-function and
  formula table calculations (closes the table-calculation half of #10229;
  custom SQL dimensions still pending).
- Add API, Cypress, and common utility coverage for the pivot sorting blast
  radius. Cypress selector now matches both `<svg>` and `<canvas>` echarts
  renderers (canvas became the default in #22422).

## Testing

- pnpm -F common test -- derivePivotConfigFromChart.test.ts
- pnpm -F common test -- chartValidation.test.ts
- pnpm -F common typecheck
- pnpm -F frontend typecheck
- SITE_URL=http://localhost:8100 pnpm -F api-tests test:api tests/pivotSorting.test.ts
- pnpm -F e2e cypress:run --spec cypress/e2e/app/pivotTables.cy.ts --config baseUrl=http://localhost:3020
- Targeted lint/format checks for touched packages

### DIFF
--- a/packages/api-tests/tests/pivotSorting.test.ts
+++ b/packages/api-tests/tests/pivotSorting.test.ts
@@ -369,7 +369,6 @@ describe('Pivot sorting', () => {
 
         expect(customerLabels.length).toBeGreaterThan(1);
         expect(new Set(customerLabels).size).toBeGreaterThan(1);
-        expect(customerLabels).toEqual([...customerLabels].sort());
     });
 
     it('returns x-axis metric values when the chart is sorted by that metric', async () => {
--- a/packages/api-tests/tests/pivotSorting.test.ts
+++ b/packages/api-tests/tests/pivotSorting.test.ts
@@ -1,14 +1,23 @@
 import {
     CartesianSeriesType,
     ChartType,
+    DashboardTileTypes,
+    DownloadFileType,
     FilterOperator,
     QueryHistoryStatus,
     SEED_PROJECT,
     TableCalculationType,
+    type ApiDownloadAsyncQueryResultsAsCsv,
     type ApiGetAsyncQueryResults,
+    type CreateChartInDashboard,
     type CreateChartInSpace,
+    type CreateDashboard,
+    type Dashboard,
+    type DashboardChartTile,
+    type DashboardFilters,
     type ResultRow,
     type SavedChart,
+    type UpdateDashboard,
 } from '@lightdash/common';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 import { ApiClient, Body } from '../helpers/api-client';
@@ -34,6 +43,16 @@ const getColumnIds = (results: ReadyQueryResults): string[] => [
     ...results.rows.flatMap((row) => Object.keys(row)),
 ];
 
+const getPivotValues = (
+    results: ReadyQueryResults,
+    fieldId: string,
+): unknown[] =>
+    results.pivotDetails?.valuesColumns.flatMap((column) =>
+        column.pivotValues
+            .filter((pivotValue) => pivotValue.referenceField === fieldId)
+            .map((pivotValue) => pivotValue.value),
+    ) ?? [];
+
 const booleanEqualsFilter = (fieldId: string, value: boolean) => ({
     dimensions: {
         id: `${fieldId}-filter`,
@@ -50,6 +69,12 @@ const booleanEqualsFilter = (fieldId: string, value: boolean) => ({
 
 const fieldReference = (fieldId: string) => `\${${fieldId}}`;
 
+const emptyDashboardFilters: DashboardFilters = {
+    dimensions: [],
+    metrics: [],
+    tableCalculations: [],
+};
+
 const customerLabelTableCalculation = {
     name: 'customer_label',
     displayName: 'Customer label',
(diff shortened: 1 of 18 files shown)


charliedowler:
* **#22421** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22421?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22421?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #19274, opened 2026-01-08
fix: table calculations cause extra series

magnew:
Closes: #19252 

### Description:

**The issue**
With SQL pivoting enabled, table calculations were causing incorrect rendering for grouped charts. We were always treating them as index columns, but usually the are more like value columns. 

[...]

PR #22429, opened 2026-04-28
fix: preserve pivot x-axis sort fields

### LINKED ISSUE #22389: Cartesian chart x-axis only renders first value when x-axis is a table calculation that is also a sort field

charliedowler:
## Summary

Fixes #22389.

This keeps Cartesian pivot x-axis fields in their index role when the same metric or table calculation is also used as a sort field. Previously, sorted metrics/table calculations that were not displayed as y-values were collected as sort-only columns before index derivation. When the sorted field was also the x-axis table calculation, it was excluded from `indexColumn`, causing the chart to collapse onto the wrong x-axis bucket.

The fix derives `indexColumn` from the displayed chart roles first, then computes `sortOnlyColumns` only for sort fields that are not already represented as values, group-by columns, or index columns.

### Unit-test repro (deterministic, ~2s)

The new test case `keeps x-axis table calculation as an index column when sorted by itself` in `derivePivotConfigFromChart.test.ts` fails on `main` and passes on this branch:

```
✕ keeps x-axis table calculation as an index column when sorted by itself
  Expected value: {"reference": "payment_method_label", "type": "category"}
  Received array: [{"reference": "payments_payment_method", "type": "category"}]
```

The buggy `indexColumn` substitutes the underlying raw dimension instead of the table calc, so the chart loses its x-axis variation.

### Live UI repro (Jaffle shop demo)

Build a Cartesian (bar) chart matching the unit test shape:
- Table: `orders`
- Dimensions: `order_priority`, `status`
- Metric: `total_order_amount`
- Table calculation: `priority_label` = `upper(${orders.order_priority})`
- Pivot config: `columns: ["orders_status"]`
- x-axis field: `priority_label` (the table calc)
- Sort: `priority_label` (same table calc)

Both x-axis-as-table-calc AND sort-by-same-table-calc are required to trigger the regression.

**Before fix:** chart collapses to a single x-axis position with all status series stacked at one location, even though the result table shows three distinct priority values (HIGH/NORMAL/URGENT). A `Results may be incorrect` warning is raised.

**After fix:** chart correctly renders three x-axis groups (HIGH, NORMAL, URGENT) with stacked status pivot series.

To reproduce in seconds without clicking through the UI, hydrate an explore URL with `create_saved_chart_version=<urlencoded JSON>` matching the shape above (see test file for exact field IDs).

## Testing

- `pnpm -F common test -- derivePivotConfigFromChart.test.ts`
- `pnpm -F common typecheck`

## Notes

This is intentionally narrower than #22421: it does not add formula-table-calculation validation changes, pivot-aware table-calculation sort errors, sort-label UI changes, or support for metric/table-calculation pivot columns.

### DIFF
--- a/packages/common/src/pivot/derivePivotConfigFromChart.ts
+++ b/packages/common/src/pivot/derivePivotConfigFromChart.ts
@@ -164,37 +164,6 @@ const getIndexColumn = (
         .filter((col): col is NonNullable<typeof col> => col !== undefined);
 };
 
-const isMetricOrTableCalculation = (
-    fieldId: string,
-    metricQuery: MetricQuery,
-) =>
-    metricQuery.metrics.includes(fieldId) ||
-    (metricQuery.tableCalculations || []).some((tc) => tc.name === fieldId);
-
-const getSortOnlyColumns = (
-    metricQuery: MetricQuery,
-    displayedValuesColumns: PivotConfiguration['valuesColumns'],
-    groupByColumns: PivotConfiguration['groupByColumns'],
-    indexColumn: PivotConfiguration['indexColumn'],
-): NonNullable<PivotConfiguration['sortOnlyColumns']> => {
-    const roleFieldIds = new Set([
-        ...displayedValuesColumns.map((column) => column.reference),
-        ...(groupByColumns ?? []).map((column) => column.reference),
-        ...normalizeIndexColumns(indexColumn).map((column) => column.reference),
-    ]);
-
-    return metricQuery.sorts
-        .filter(
-            (sort) =>
-                !roleFieldIds.has(sort.fieldId) &&
-                isMetricOrTableCalculation(sort.fieldId, metricQuery),
-        )
-        .map((sort) => ({
-            reference: sort.fieldId,
-            aggregation: VizAggregationOptions.ANY,
-        }));
-};
-
 function getTablePivotConfiguration(
     savedChart: Pick<SavedChartDAO, 'chartConfig' | 'pivotConfig'>,
     metricQuery: MetricQuery,
@@ -316,15 +285,28 @@ function getCartesianPivotConfiguration(
             xField,
         );
 
-        // Include metrics/table calculations that are used in sorts but not
-        // displayed in the chart or otherwise represented by the pivot roles.
-        // Without these, PivotQueryBuilder can't generate anchor CTEs for sort keys
-        // that are not part of the chart output.
-        const sortOnlyColumns = getSortOnlyColumns(
-            metricQuery,
-            valuesColumns,
-            groupByColumns,
-            indexColumn,
+        // Metrics/table calculations used in sorts but not displayed in the
+        // chart still need anchor CTEs in PivotQueryBuilder, so list them as
+        // sort-only columns. xField TCs/metrics are already represented by
+        // indexColumn; groupBy fields are dimensions, so the metric/TC check
+        // below excludes them.
+        const valuesRefs = new Set(valuesColumns.map((c) => c.reference));
+        const sortOnlyColumns: NonNullable<
+            PivotConfiguration['sortOnlyColumns']
+        > = metricQuery.sorts.flatMap((sort) =>
+            sort.fieldId !== xField &&
+            !valuesRefs.has(sort.fieldId) &&
+            (metricQuery.metrics.includes(sort.fieldId) ||
+                (metricQuery.tableCalculations || []).some(
+                    (tc) => tc.name === sort.fieldId,
+                ))
+                ? [
+                      {
+                          reference: sort.fieldId,
+                          aggregation: VizAggregationOptions.ANY,
+                      },
+                  ]
+                : [],
         );
 
         const partialPivotConfiguration: Omit<PivotConfiguration, 'sortBy'> = {
--- a/packages/common/src/pivot/derivePivotConfigFromChart.ts
+++ b/packages/common/src/pivot/derivePivotConfigFromChart.ts
@@ -94,14 +94,13 @@ const getIndexColumn = (
         groupByColumns?.map((c) => c.reference) ?? [];
     const valuesColumnsReferences =
         valuesColumns?.map((c) => c.reference) ?? [];
-    const sortFieldIds = new Set(metricQuery.sorts.map((s) => s.fieldId));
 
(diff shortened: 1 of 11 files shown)


packages/common/src/pivot/derivePivotConfigFromChart.ts
```diff
@@ -94,14 +94,13 @@ const getIndexColumn = (
...
 
     // Find any columns that are not groupBy or value columns (these become index columns)
-    // Table calculations are only included if they are the xField (used as the x-axis dimension)
-    // or if they are used in sorting. Otherwise we don't want to include them as multiple
-    // index columns cause multiple series.
+    // Table calculations are only included when they are the xField. Sort-only table
+    // calculations are handled separately so they can order the result without
+    // changing the chart's index fields.
     const tableCalcNames = (metricQuery.tableCalculations || [])
-        .filter((tc) => tc.name === xField || sortFieldIds.has(tc.name))
```
rephus, line 104: This ` || sortFieldIds.has(tc.name)`. you removed was introduced here  https://github.com/lightdash/lightdash/pull/19274

We should check what the issue was and makes sure we are not reintroducing it.

charliedowler: Good shout. This is not re-introduced, have tested manually and have added regression tests.

PR #22454, opened 2026-04-28
fix(pivot): respect explicit sortBy order over groupByColumns declaration

almeidabbm:
Relates to: https://github.com/lightdash/lightdash/issues/20117

### Description:

When a pivot query has multiple group-by columns and a `sortBy` referencing only one of them, the `column_index` `DENSE_RANK()` window function was previously ordering by the group-by columns in their declaration order, ignoring the user-specified sort. This meant that if `status` was declared before `status_priority`, the ranking would be driven by `status` alphabetically, with `status_priority` only acting as a tiebreaker — even when the user explicitly sorted by `status_priority`.

The fix reorders the `OVER (ORDER BY ...)` clause so that group-by columns referenced in `sortBy` come first (in the order the user specified them), followed by any remaining group-by columns in their original declaration order. This ensures the pivot column ordering reflects the user's intent.

**Before**

```sql
WITH
  original_query AS (
    SELECT
      "raw_order_statuses".status_priority AS "raw_order_statuses_status_priority",
      "orders".status AS "orders_status",
      "orders".shipping_method AS "orders_shipping_method",
      COUNT(DISTINCT "orders".order_id) AS "orders_unique_order_count"
    FROM
      "postgres"."jaffle"."orders" AS "orders"
      LEFT OUTER JOIN "postgres"."jaffle"."raw_order_statuses" AS "raw_order_statuses" ON ("orders".status) = ("raw_order_statuses".status)
    GROUP BY
      1,
      2,
      3
    ORDER BY
      "raw_order_statuses_status_priority"
  ),
  group_by_query AS (
    SELECT
      "orders_status",
      "raw_order_statuses_status_priority",
      "orders_shipping_method",
      (ARRAY_AGG("orders_unique_order_count")) [1] AS "orders_unique_order_count_any"
    FROM
      original_query
    group by
      "orders_status",
      "raw_order_statuses_status_priority",
      "orders_shipping_method"
  ),
  pivot_query AS (
    SELECT
      g."orders_shipping_method",
      g."orders_status",
      g."raw_order_statuses_status_priority",
      g."orders_unique_order_count_any",
      DENSE_RANK() OVER (
        ORDER BY
          g."orders_shipping_method" ASC
      ) AS "row_index",
      DENSE_RANK() OVER (
        ORDER BY
          g."orders_status" ASC,
          g."raw_order_statuses_status_priority" ASC
      ) AS "column_index"
    FROM
      group_by_query g
  ),
  filtered_rows AS (
    SELECT
      *
[...]

PR #22489, opened 2026-04-29
fix: revert "fix(pivot): respect explicit sortBy order over groupByColumns declaration"

almeidabbm:
Reverts #22454.

The pivot table renderer treats `groupByColumns` declaration order as the **column hierarchy** (first entry = outermost header) and merges adjacent value columns that share the same outer value. The SQL has to emit columns in the same order for header merging to be correct.

#22454 put the `sortBy` group-by first in the column rank's `ORDER BY`, which restructures the SQL hierarchy without telling the renderer. With M:N group-bys this produces wrong header merging — e.g. `groupByColumns: [payment_method, status]` sorted by `status` ends up rendering "bank_transfer" merged across four columns that are actually four different payment methods, all `completed`.

#22454's own example worked only because its two group-bys were 1:1, so flipping their order in the rank had no observable effect on the column clustering — just on the visible sort.

**Chart config**
CleanShot 2026-04-29 at 10 52 47

**Before (bug)**
CleanShot 2026-04-29 at 10 50 30

**After**
CleanShot 2026-04-29 at 10 52 19

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.test.ts
+++ b/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.test.ts
@@ -754,43 +754,6 @@ describe('PivotQueryBuilder', () => {
             expect(result).toContain('ORDER BY "date" ASC, "event" DESC');
         });
 
-        test('Should lead column_index ORDER BY with the sorted group-by column, not declaration order', () => {
-            // Without this, DENSE_RANK orders columns by the first declared
-            // group-by alphabetically, with the user-sorted column only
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#22489** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22489?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22489?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22488, opened 2026-04-29
feat(permissions): strip SQL bodies from explore-from-here URLs

joaoviana:
### Summary

Stops "Explore from here" URLs from leaking the `sql` body of every custom SQL dimension and SQL table calculation into browser history, share-link records, and the URL bar — for users without `manage:CustomFields`. Admin / Developer URLs still round-trip the SQL so shareable links remain reproducible end-to-end.

Closes [PROD-7180](https://linear.app/lightdash/issue/PROD-7180). Builds on the saved-version exemption that landed via [#22451](https://github.com/lightdash/lightdash/pull/22451).

### Architecture (strip → carry → exempt → rehydrate)

1. **Strip on encode.** `getExplorerUrlFromCreateSavedChartVersion` accepts a `cannotAuthorCustomSql` option. When true, the URL builder runs the new `stripSqlBodiesFromMetricQuery` helper from `@lightdash/common` to blank every SQL field's `sql` body before serialising into `create_saved_chart_version`.
2. **Carry the source UUID.** A new `savedChartUuid` query param holds the originating chart's UUID so the destination can identify the saved version.
3. **Saved-version exemption.** The diff helper already exempted unchanged SQL fields. It now treats *empty incoming SQL against a non-empty saved value as preserved* — so the strip-on-encode round-trip doesn't fire the gate.
4. **Rehydrate before execution.** `assertCanRunSqlAuthoredFields` returns a resolved metric query with `mergeSavedSqlBodiesIntoMetricQuery` overlaying saved SQL bodies onto empty incoming ones. `executeAsyncMetricQuery`, `calculateTotalFromQuery`, and `calculateSubtotalsFromQuery` use the resolved query for execution so the warehouse SQL still compiles.

### Where the source UUID flows in

A new `useSourceSavedChartUuid` hook centralises the lookup with three fallbacks:

1. Redux `selectSavedChart` — loaded by the `/saved/{uuid}` route.
2. `savedChartUuid` URL search param — modern explore-from-here URLs.
3. **Backward compat:** `uuid` field inside `create_saved_chart_version` JSON — pre-PROD-7180 share links don't carry the new query param but do carry the chart UUID inside the payload, so the saved-version exemption still fires for existing share links.

`useColumns`, `useExplorerQueryManager`, and `VisualizationCard` consume the hook so run / totals / subtotals all forward `savedChartUuid` consistently.

### URL builder callers updated

- `components/ExploreFromHereButton/index.tsx` (main toolbar button)
[...]

PR #22492, opened 2026-04-29
feat(permissions): hide "Edit calculation" in column header for SQL table calcs

joaoviana:
### Summary

The column-header context menu showed **"Edit calculation"** for every table calc regardless of variant or scope. For SQL table calcs that exposed the SQL editor — and the authored SQL — to users without `manage:CustomFields`. The backend rejects on save, but the editor wastes time getting there.

Mirrors the existing custom-dim gate in the same file: hide the menu item when `isSqlTableCalculation(item) && cannotAuthorCustomSql`. Formula and Template table calcs are unaffected because they don't expose a `sql` body.

This is a follow-up that came out of [PROD-7180](https://linear.app/lightdash/issue/PROD-7180) manual testing — separate from that PR ([#22488](https://github.com/lightdash/lightdash/pull/22488)) since the affordance lives in a different surface.

### What changed

`packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx`:
- Import `isSqlTableCalculation` from `@lightdash/common`.
- Wrap the existing "Edit calculation" `Menu.Item` (and the divider that follows it) in `{!(isSqlTableCalculation(item) && cannotAuthorCustomSql) && (…)}` — same predicate already used for the custom-dim Edit just above.

### Test plan

- [ ] As an Editor (no `manage:CustomFields`), open a saved chart that has a SQL table calc → expand the Results card → click the column header menu on the SQL table calc → "Edit calculation" is **not** in the menu (only Filter, Sort, Remove visible).
- [ ] Same flow on a **Formula** or **Template** table calc → "Edit calculation" is visible (those variants don't expose a sql body, so they're not gated).
- [ ] As an Admin → "Edit calculation" is visible for the SQL table calc.
- [ ] Existing custom-dim Edit gate (immediately above this block) still works for both personas.

### DIFF
--- a/packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
+++ b/packages/frontend/src/components/Explorer/ResultsCard/ColumnHeaderContextMenu.tsx
@@ -12,6 +12,7 @@ import {
     isMetricWithDateValue,
     isNumericItem,
     isPeriodOverPeriodAdditionalMetric,
+    isSqlTableCalculation,
     isTableCalculation,
     isTimeBasedDimension,
     type TableCalculation,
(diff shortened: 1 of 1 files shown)


joaoviana:
* **#22492** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22492?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22507, opened 2026-04-29
feat(permissions): gate SQL-authored field writes

joaoviana:
Closes part 1 of [PROD-7028](https://linear.app/lightdash/issue/PROD-7028).

## What this fixes

Three persistence paths today write a chart's `metricQuery` without checking that the caller holds `manage:CustomFields` — meaning a user with the right combination of lower-tier scopes can author or upload charts containing custom SQL dimensions or SQL table calculations and bypass the intended gate.

| Path | Today | After this PR |
| --- | --- | --- |
| `SavedChartService.create()` | Gates SQL table calcs only; **misses custom SQL dimensions** | Gates both via the shared guard |
| `CoderService.upsertChart()` | Only checks `manage:ContentAsCode` | Also requires `manage:CustomFields` if payload contains custom SQL fields |
| `PromoteService.upsertCharts()` | No `manage:CustomFields` check at all | Walks the whole batch, fails atomically with all offending chart names enumerated |

## How

New shared utility `assertCanWriteSqlAuthoredFields` in `packages/backend/src/utils/SqlAuthoredFieldsGuard.ts`. Wraps `hasSqlAuthoredFields()` from `@lightdash/common` (covers both custom SQL dims and SQL table calcs) and the `manage:CustomFields` CASL check. Throws `CustomSqlQueryForbiddenError` with a write-context message.

Strict semantics — any SQL-authored field requires the scope, no diff-based exemption. The existing `assertCanRunSqlAuthoredFields` in `AsyncQueryService` keeps its diff-based exemption for the run-side (re-running an existing saved chart's SQL is intentionally allowed without the scope; modifying it is not).

## Who's affected

| User class | Holds `manage:CustomFields`? | Affected? |
| --- | --- | --- |
| Service accounts (CI/CD, default org-admin) | Yes | No |
| Developer system role | Yes | No |
| Editor / Viewer system roles | No, but no `manage:ContentAsCode` either | No (cannot reach Coder/Promote) |
| Custom role: ContentAsCode + CustomFields | Yes | No |
| Custom role: ContentAsCode **without** CustomFields | No | **Yes** — cannot upload/promote charts with SQL fields. *This is the bypass we are closing.* |
| Cross-project promotion: scope in source, not destination | Mixed | **Yes** — cannot promote SQL fields into a destination where authoring would be denied. Correct-by-spec. |

[...]

PR #22510, opened 2026-04-29
feat(permissions): strip or fail on SQL-authored field reads

joaoviana:
Closes: [PROD-7028](https://linear.app/lightdash/issue/PROD-7028). Stacked on top of #22507.

## What this fixes

Three read-side paths today return chart definitions including raw `customDimensions[].sql` and SQL table calculation bodies to callers who do not hold `manage:CustomFields`. PR 1 (#22507) closed the *write* side; this PR closes the *read* side.

| Endpoint | Behaviour | Why this strategy |
| --- | --- | --- |
| `SavedChartService.get()` | **Strip** SQL bodies via `stripSqlBodiesFromMetricQuery()` for callers without the scope | General-purpose chart fetch hit by every chart load (incl. dashboard tiles, which are references-only and resolve through this method). Strip-on-read mirrors the URL-encoding pattern already used on the frontend; `mergeSavedSqlBodiesIntoMetricQuery` rehydrates on the run-side so editing/running still works for callers with the scope. |
| `CoderService.getCharts()` | **Fail** with `CustomSqlQueryForbiddenError`, listing offending chart names | Authoring-tier path. `lightdash download` is realistically only run by Developers or service accounts (already hold the scope). Returning a YAML with empty SQL would be silently corrupt and the user could not re-upload it anyway (PR 1 blocks that). Fail loud. |
| `PromoteService.getPromoteChartDiff()` and `getPromoteDashboardDiff()` | **Fail** with `CustomSqlQueryForbiddenError`, listing offending chart names | Diff endpoints exist to support the privileged promote operation, which PR 1 blocks at the destination scope. Showing a diff the user cannot act on is just confusing — fail at the same layer. |

The dashboard endpoint is unchanged: `DashboardChartTileProperties` is references-only (just `savedChartUuid` + display metadata), and the chart fetch flows through `SavedChartService.get()` which is now stripped. No separate dashboard-side leak.

## Helper changes (`utils/SqlAuthoredFieldsGuard.ts`)

- Renamed `assertCanWriteSqlAuthoredFields` → `assertCanAccessSqlAuthoredFields`. The underlying check ("metricQuery has SQL fields AND caller lacks the scope") applies equally to write and read contexts, so the neutral name is more accurate. Call sites from PR 1 updated to the new name.
- Added `stripSqlBodiesIfForbidden` — pure helper that returns the metricQuery unchanged when the caller holds the scope, or with empty SQL bodies otherwise. Used by `SavedChartService.get()`.

## Who's affected (read-side)

| Caller class | Has `manage:CustomFields`? | Effect |
| --- | --- | --- |
| Service accounts (org-admin default) | Yes | No change |
| Developer / Editor / Admin system roles | Yes | No change |
| Viewer-equivalent custom roles | Often no | Chart-fetch responses now contain empty `sql` fields. The chart still renders normally; the empty body is only visible when inspecting the API response or opening the SQL editor (which `CustomDimensionModal` already hides for these users). |
| Custom role: ContentAsCode without CustomFields | No | Cannot download YAML — gets 403 with chart names |
| Custom role attempting promotion without destination CustomFields | No | Cannot view promotion diff — gets 403 with chart names |

No known customer configurations rely on these read paths returning SQL bodies to callers without the scope.

## Out of scope (deferred)

The Linear comment flagged that other endpoints returning chart definitions inline (catalog, search results, expanded spaces, v1 list endpoints) may also leak SQL bodies. This PR sticks to the three call sites the comment explicitly identified. A comprehensive sweep of all chart-returning endpoints is the natural follow-up.

## Test plan

- [x] `pnpm -F backend lint` — clean
- [x] `pnpm -F backend typecheck` — clean
- [x] `pnpm -F backend exec jest --testPathPattern='SqlAuthoredFieldsGuard|sqlAuthoredFields|readSqlAuthoredFields|SavedChartService|CoderService|PromoteService'` — all 62 tests pass
- [ ] Manual: as a Developer (has `manage:CustomFields`), opening a chart with a custom SQL dim still shows the full SQL body in the editor
- [ ] Manual: as a Viewer, opening the same chart renders correctly but the API response contains empty `sql` fields (run path uses `mergeSavedSqlBodiesIntoMetricQuery` to rehydrate from the saved chart)
- [ ] Manual: `lightdash download` for a custom role granting only `ContentAsCode` returns 403 with chart names
- [ ] Manual: `GET /api/v1/saved/{uuid}/promoteDiff` for a user lacking destination `manage:CustomFields` returns 403

test-all

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.sqlAuthoredFields.test.ts
+++ b/packages/backend/src/services/CoderService/CoderService.sqlAuthoredFields.test.ts
@@ -6,8 +6,10 @@ import {
     DimensionType,
     OrganizationMemberRole,
     PossibleAbilities,
+    TableCalculationType,
     type ChartAsCode,
     type CustomSqlDimension,
+    type SqlTableCalculation,
 } from '@lightdash/common';
 import { analyticsMock } from '../../analytics/LightdashAnalytics.mock';
 import { lightdashConfigMock } from '../../config/lightdashConfig.mock';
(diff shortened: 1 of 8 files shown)


joaoviana:
* **#22511** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22511?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22510** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22510?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22510?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22507** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22507?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

joaoviana:
Tested
Embedding, SDK
CLI, API (curl) 
Visual tests

PR #22523, opened 2026-04-29
fix: drop silent 50-row default in CatalogModel.search

### LINKED ISSUE #22497: Metrics Catalog: some dimensions missing from filter options
## Summary

Dimensions defined in the dbt model aren't showing up as filter options in the Metrics Catalog, even though they're available and working in the Table Explore view for the same model. Looks systemic rather than a config issue — affects every metric in the affected model.

## Steps to reproduce

1. Open the Metrics Catalog and pick a metric from a model with multiple dimensions
2. Try to filter by one of the model's dimensions
3. Some fields arw missing from the filter list
4. Switch to Table Explore on the same model — the dimension is there and works as expected

## Expected

Dimensions defined in the dbt model should be available as filters in the Metrics Catalog, matching what's offered in Table Explore.

## Actual

A subset of dimensions is missing from the filter list. All metrics in the affected model appear impacted.

almeidabbm:
## Summary

`CatalogModel.search` defaulted `pageSize` to 50 when callers omitted `paginateArgs`, silently truncating per-table dim lists. The Metrics Catalog filter dropdown lost dimensions on tables with more than 50 catalog rows.

The truncation was made worse by `getFilterDimensions` applying its in-memory `string|boolean only / exclude time-derived` filter *after* the DB pagination — so up to 50 catalog rows came back, and the type filter could shrink the visible result further.

Drops the silent default and adds explicit `{ pageSize: 50 }` to the project-wide callers that relied on it (`getCatalog` search, `getMetricsCatalog`, v1 metrics-with-time-dimensions). `getFilterDimensions` and `getSegmentDimensions` now return the full per-table list, bounded by their `exploreName` filter.

## Test plan

- [x] New API regression tests in `catalog.test.ts` against the seeded `events` explore (the only seed wide enough to hit the cap)
- [x] Backend lint + typecheck clean

Closes: https://github.com/lightdash/lightdash/issues/22497

### DIFF
--- a/packages/api-tests/tests/catalog.test.ts
+++ b/packages/api-tests/tests/catalog.test.ts
@@ -875,12 +875,6 @@ describe('Lightdash catalog filter / segment dimensions', () => {
         admin = await login();
     });
 
-    // Regression for PROD-7206: CatalogModel.search defaulted paginateArgs
-    // to pageSize 50, so callers that wanted the full per-table dim list
-    // silently lost any rows past position 50. The events explore is the
-    // only seed explore wide enough (> 50 indexable dimensions) to expose
-    // the truncation. Compute the expected set from the explore directly
-    // so this test stays correct as the YAML evolves.
     const expectedFilterableDimNames = (explore: {
         baseTable: string;
         tables: Record<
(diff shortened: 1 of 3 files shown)


almeidabbm:
* **#22523** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22549, opened 2026-04-30
chore: small cleanup

joaoviana:
Small cleanup pass.


joaoviana:
Closing — opened against the wrong base. The actual change (editor gate on the custom SQL dimension modal) is already present on main; the diff here was bloated by an unrelated local rollback branch.

PR #22569, opened 2026-04-30
fix(pivot): preserve NULL groups when pivoting (PROD-7075)

### LINKED ISSUE #22208: Null dimension values are dropped when the dimension is pivoted to a column in table charts
### Description

When a dimension containing null values is used as a **row**, null renders as its own row in a table chart. When the same dimension is moved to **columns** (pivoted), the null group is silently dropped — it does not appear as a null column.

The expected behaviour is that pivoted columns behave symmetrically with pivoted rows: a null group should render as a column (optionally labelled "null" or blank) alongside the non-null groups.

### Steps to reproduce

1. Create a table chart with:
    - A metric (any aggregation)
    - A non-null dimension in rows
    - A dimension that contains some `NULL` values (e.g. a date dimension) in columns
2. Observe that no column is rendered for the null group — the data for those rows is missing from the pivot.
3. Move the same dimension from columns to rows. The null group now renders as a row as expected.

### Expected

Null pivot groups render as a column (similar to how they render as a row).

### Actual

Null pivot groups are dropped from the pivoted result set.

### Root cause (investigation)

**Primary** — `packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts:1574-1588`:

```ts
const valueSuffix =
    pivotValues.length > 0
        ? pivotValues.map((p) => p.value).join('_')
        : '';

valuesColumns.forEach((col) => {
    const valueColumnField = PivotQueryBuilder.getValueColumnFieldName(
        col.reference,
        col.aggregation,
    );
    const valueColumnReference = valueSuffix
        ? `${valueColumnField}_${valueSuffix}`
        : valueColumnField;
    ...
});
```

- `[null].join('_')` returns `""`, which is falsy.
- `valueColumnReference` then falls back to the un-suffixed `valueColumnField`, colliding with the base column key.
- The null pivot group's data either overwrites or is overwritten by the base column and is not distinguishable in the transformed row.

**Secondary** — `packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.ts` (precomputed-rankings path, lines 1060-1090 and 1341):

- LEFT JOINs between the pivot result and the column-ranking CTE use `=` comparisons on the group-by columns.
- In SQL, `NULL = NULL` is NULL (not true), so null pivot groups fail to join.
- `column_index` comes back NULL and is filtered out by `WHERE column_index <= maxColumnsPerValueColumn`.

### Suggested fix

- Generate a deterministic, non-empty key for null pivot values (e.g. `p.value ?? '__null__'`) when building `valueSuffix`, and preserve the null-ness in the column metadata so the frontend can render the label appropriately.
- In `PivotQueryBuilder`, change JOIN conditions that compare potentially-null group-by columns to use `IS NOT DISTINCT FROM` (or warehouse-specific equivalents) so null groups match.

### Workaround

Coalesce null values in the dbt model before they reach Lightdash (e.g. `COALESCE(my_date::text, 'unknown')`).

almeidabbm:
Closes https://github.com/lightdash/lightdash/issues/22208

NULL pivot groups were being silently dropped from table charts. Two bugs:

**1. Column-key collision in `AsyncQueryService`**
Pivot column keys were built with `pivotValues.map(p => p.value).join('_')`. When the groupBy value was NULL, `[null].join('_')` is `''`, the falsy-suffix branch kicked in, and the NULL group's column collided with the unsuffixed base column. NULL is now mapped through a sentinel and the falsy fallback is gone.

**2. NULL-unsafe joins in `PivotQueryBuilder`**
`row_ranking` / `column_ranking` (and the anchor CTEs they depend on) joined with plain `=`. Since `NULL = NULL` is `NULL`, NULL groups missed their rank, came back with `row_index = NULL`, and got filtered out by `WHERE row_index <= limit`. Added `getNullSafeEqualSql` to the warehouse SQL builder (default: `(a = b OR (a IS NULL AND b IS NULL))`, dialects can override) and routed every pivot join through it.

NULL groups now render as a column or row alongside the rest, labelled `∅`.

**Before**
before_rows_omitted
before_columns_omitted

**After**
after_as_rows
after_as_columns

### DIFF
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -196,6 +196,12 @@ import {
 } from './types';
 
 const SQL_QUERY_MOCK_EXPLORER_NAME = 'sql_query_explorer';
+
+// Sentinel used when building the suffix on pivoted value-column references for
+// NULL groupBy values. `[null].join('_')` collapses to '' and would collide with
+// the unsuffixed base column key. The sentinel is only used in internal column
+// keys — the frontend reads `pivotValues[].value === null` for label rendering.
+const PIVOT_NULL_KEY = '__lightdash_null__';
 export const QUEUED_QUERY_EXPIRED_MESSAGE =
     'Your query expired while waiting in the queue. Please try again.';
 
@@ -1671,10 +1677,20 @@ export class AsyncQueryService extends ProjectService {
 
                       // Suffix the value column with the group by columns to avoid collisions.
                       // E.g. if we have a row with the value 1 and the group by columns are ['a', 'b'],
-                      // then the value column will be 'value_1_a_b'
+                      // then the value column will be 'value_1_a_b'.
+                      // NULL pivot values are mapped to a sentinel so that
+                      // `[null].join('_')` doesn't collapse to '' and collide with the
+                      // unsuffixed base column.
                       const valueSuffix =
                           pivotValues.length > 0
-                              ? pivotValues.map((p) => p.value).join('_')
+                              ? pivotValues
+                                    .map((p) =>
+                                        p.value === null ||
+                                        p.value === undefined
+                                            ? PIVOT_NULL_KEY
+                                            : p.value,
+                                    )
+                                    .join('_')
                               : '';
 
                       // eslint-disable-next-line @typescript-eslint/no-loop-func -- forEach is synchronous, executes within current loop iteration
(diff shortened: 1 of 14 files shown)


almeidabbm:
* **#22569** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22585, opened 2026-05-01
fix: improve partial compilation error messages with field names and model-prefix hints

rephus:
Closes: PROD-6943

### Description:

When a parameter is referenced using the short form (e.g. `${lightdash.parameters.region}`) but only exists as a model-prefixed parameter (e.g. `a.region`), the compiler now detects this mismatch and surfaces a hint pointing to the correct full reference (e.g. `${lightdash.parameters.a.region}`) instead of silently compiling to `NULL` or throwing a generic missing parameter error.

Partial compilation warnings for dimensions and metrics now include the field name in the error message (e.g. `Dimension "selected_team" failed to compile: ...`) so it's immediately clear which field is broken and why.

The CLI compile output for partial successes now displays each warning on its own indented line with a `⚠` prefix, making it easier to read when multiple warnings are present.

### DIFF
--- a/packages/cli/src/handlers/compile.ts
+++ b/packages/cli/src/handlers/compile.ts
@@ -368,7 +368,12 @@ export const compile = async (options: CompileHandlerOptions) => {
             e.warnings.length > 0
         ) {
             status = styles.warning('PARTIAL_SUCCESS');
-            messages = `: ${styles.warning(e.warnings.map((warning) => warning.message).join(', '))}`;
+            messages = `\n${e.warnings
+                .map(
+                    (warning) =>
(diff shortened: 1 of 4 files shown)


rephus:
* **#22585** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22585?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22585?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22632, opened 2026-05-04
fix(timezone): coerce inner expr to TIMESTAMP for BigQuery 3-arg TIMESTAMP_TRUNC

almeidabbm:
### What
BigQuery rejects timezone-aware `TIMESTAMP_TRUNC` when the inner expression is a `DATETIME` instead of a `TIMESTAMP`. This happens when a dimension declared `type: timestamp` actually emits `DATETIME` at runtime — e.g. `timestamp_add(...)` over a `DATE` column.

CleanShot 2026-05-04 at 11 21 39

### Fix
Wrap the inner expression in `TIMESTAMP(...)` for the 3-arg `TIMESTAMP_TRUNC` path. No-op for real timestamps, valid coercion for DATETIME (interpreted as UTC, which matches the truncation zone). Other adapters already absorb the same drift via their explicit casts (`::timestamptz`, `convert_timezone`, `CAST AS timestamp`); this just brings BigQuery in line.

### Flag scope
Fully gated behind `EnableTimezoneSupport`. The new `TIMESTAMP(...)` wrap only fires inside the `if (timezone)` branch, and `timezone` is only ever forwarded when `useTimezoneAwareDateTrunc` is true (set from the flag). Flag-off SQL is byte-identical to before.

### DIFF
--- a/packages/common/src/utils/timeFrames.test.ts
+++ b/packages/common/src/utils/timeFrames.test.ts
@@ -92,6 +92,32 @@ describe('TimeFrames', () => {
             );
         });
 
+        test('BigQuery 3-arg TIMESTAMP_TRUNC wraps inner expr in TIMESTAMP() so DATETIME inputs coerce', () => {
+            expect(
+                getSqlForTruncatedDate(
+                    SupportedDbtAdapter.BIGQUERY,
+                    TimeFrames.DAY,
+                    '${TABLE}.created',
+                    DimensionType.TIMESTAMP,
+                    undefined,
+                    'UTC',
+                ),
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#22632** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22655, opened 2026-05-05
test: add service account authentication, scope matrix, and identity regression tests

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7407/add-api-tests-to-avoid-regressions

  - Authentication negatives — missing/invalid/expired bearer (4 cases) plus a positive smoke.
  - Scope matrix — ORG_READ / ORG_EDIT / ORG_ADMIN / SCIM_MANAGE × representative endpoints. Captures two surprising current behaviors (commented
  inline so the refactor diff makes them obvious): ORG_READ unconditionally allows manage Space (broader than the scope name implies), and SCIM_MANAGE
   tokens can read /org/projects (no per-subject scope filter at the controller layer).
  - Identity (current admin-spoofing behavior) — GET /api/v1/user with an SA token returns userUuid = SEED_ORG_1_ADMIN.user_uuid, email =
  service-account@lightdash.com, firstName = service account, lastName = <description>. These are the assertions that should flip after the refactor.
  - Content attribution (current admin-spoofing behavior) — a space created by an SA grants the seeded admin's UUID into the space access list
  (because the FK is currently the admin's user uuid, not the SA's).

### Description:

Adds a comprehensive regression test baseline for the service account authentication system ahead of a planned refactor.

The new tests cover:

- **Authentication negatives** — verifying that requests with no token, an invalid token, or an expired token are rejected with `401`, while a freshly minted token is accepted.
- **Scope matrix** — exercising `ORG_READ`, `ORG_EDIT`, `ORG_ADMIN`, and `SCIM_MANAGE` scopes against relevant endpoints to confirm which operations are permitted or denied. Notably documents current over-broad behavior (e.g. `ORG_READ` allowing space creation, `SCIM_MANAGE` not being filtered at the controller level) so any future tightening surfaces as an intentional failure.
- **Identity spoofing baseline** — pins down the current behavior where `authenticateServiceAccount` borrows the seeded admin user's UUID and overwrites email/name fields with service-account placeholders. These assertions are explicitly marked to be updated when the refactor introduces a dedicated SA user record.
- **Content attribution baseline** — verifies that content created via a service account token is currently attributed to the seeded admin user, again flagged for update after the refactor replaces the borrowed identity with a real SA user.

Helper utilities (`createServiceAccountToken`, `bearerClient`) are introduced to reduce boilerplate across all new test cases.

### DIFF
--- a/packages/api-tests/tests/serviceAccounts.test.ts
+++ b/packages/api-tests/tests/serviceAccounts.test.ts
@@ -1,8 +1,62 @@
-import { SEED_PROJECT } from '@lightdash/common';
+import {
+    SEED_ORG_1,
+    SEED_ORG_1_ADMIN,
+    SEED_PROJECT,
+    ServiceAccountScope,
+} from '@lightdash/common';
 import { ApiClient, Body } from '../helpers/api-client';
 import { login } from '../helpers/auth';
 
 const apiUrl = '/api/v1';
+
+const inOneHour = () => new Date(Date.now() + 60 * 60 * 1000).toISOString();
+const oneSecondAgo = () => new Date(Date.now() - 1000).toISOString();
+
+const createServiceAccountToken = async (
+    admin: ApiClient,
+    scopes: ServiceAccountScope[],
+    {
+        descriptionPrefix = 'api-test',
+        expiresAt = inOneHour(),
+    }: { descriptionPrefix?: string; expiresAt?: string } = {},
+): Promise<{ token: string; uuid: string }> => {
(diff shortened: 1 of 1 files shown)


rephus:
* **#22655** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22655?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22710, opened 2026-05-05
feat: include metric-level tags in catalog queries

### LINKED ISSUE #22196: Bug: metric-level tags defined in dbt YAML are not returned or filterable via the data catalog API
## Bug report

Metric-level `tags` defined under `meta.metrics.<metric_name>.tags` in dbt YAML are silently dropped from the data catalog API response, and the `tags=` query parameter on the v2 `metrics-with-time-dimensions` endpoint does not match them.

This is the feature documented in:
- https://docs.lightdash.com/references/metrics#tags
- https://docs.lightdash.com/api-reference/v1/recipes#filter-metrics-by-tag-using-the-data-catalog-api

## Steps to reproduce

1. Define a metric-level tag in dbt YAML:

```yaml
columns:
  - name: feedback_id
    meta:
      metrics:
        count_of_feedback:
          type: count
          tags:
            - period_metric
```

2. Refresh dbt and re-index the catalog.
3. Call `GET /api/v2/projects/{uuid}/dataCatalog/metrics-with-time-dimensions?tableName=...&tags=period_metric`.

## Expected

- Each metric's `tags` array in the response includes its YAML-defined metric-level tags (e.g., `period_metric`).
- The `tags=period_metric` query param returns metrics that have `period_metric` defined at the metric level.

## Actual

- The `tags` array on each metric only contains node/table-level values inherited from the explore (`package:...`, model `config.tags`, etc.). Metric-level tags are missing.
- The `tags=` filter returns no results when the tag is only defined at the metric level.

## Root cause

Two locations in the catalog read path drop metric-level tags:

1. `packages/backend/src/models/CatalogModel/utils/parser.ts:167` passes `dbCatalog.explore.tags` (table-level) into `parseFieldFromMetricOrDimension` instead of the field's own `tags`. Line 74 in `parseFieldsFromCompiledTable` similarly hardcodes `tags: []`.

2. `packages/backend/src/models/CatalogModel/CatalogModel.ts:1010-1015` only filters against `${CachedExploreTableName}.explore->'tags'` (table-level), never checking per-field tags.

The metric tags ARE correctly parsed during dbt translation (`packages/common/src/types/dbt.ts:650-656` populates `Metric.tags` for both `convertModelMetric` and `convertColumnMetric`), so the fix is in the catalog read/filter path, not the dbt parser.

Note: the v1 `metrics-with-time-dimensions` endpoint doesn't accept a `tags` filter param at all (`packages/backend/src/controllers/catalogController.ts:581-585`), so any fix should be on v2.

## Affected version

0.2741.0+ (the version that introduced metric-level tag support)

magnew:
Closes: #22196, PROD-7064 

### Description:

Metric-level tags defined under meta.metrics.<metric_name>.tags in dbt YAML are now included in catalog responses. 

You can also fitler by those tags in the API


magnew:
* **#22710** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22710?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22713, opened 2026-05-05
feat: required parameters & decouple parameters from dimensions

oli-rmsy:
[PROD-7476](https://linear.app/lightdash/issue/PROD-7476/required-parameters-and-decouple-parameters-from-dimensions)

## Summary

- Adds `required: true` for **model-level** parameters and lets parameters surface in the explore and dashboard without needing a dimension that references them.
- Required parameters always show on their model's base explore (and on dashboards containing a chart from that explore) and block query execution until a value is set or a default is provided.
- Required is rejected for **project-level** parameters (validation runs in `replaceProjectParameters`).
- Required is **scoped to the explore's base table** — joined-table required params surface only when their SQL is actually referenced. Multi-tile dashboards OR-merge `required` so any tile that considers a parameter required wins, regardless of load order.
- Demo example added under `examples/full-jaffle-shop-demo/dbt/models/customers.yml` (`required_tenant`, `required_region_with_default`).

## Motivating use case

A user is parameterising the `sql_from` clause of a model — e.g. `sql_from: ${TABLE_PREFIX}_events_${ld.parameters.region}` — so the explore points at a different physical table per parameter value. The parameter is `required: true` with a default, so the explore is always queryable but the user can switch tables by changing the parameter.

Before this change, the parameter only appeared in the explore/dashboard UI if a dimension that referenced it was added to the results table — which is awkward when the parameter drives `sql_from` and has no natural dimension to expose. With this change the required parameter shows up in the explore and on dashboards without any dimension being selected, while still being blocked-on or default-resolved at run time.

## Changed surfaces

- **Type/schema**: `LightdashProjectParameter.required?: boolean`, JSON schema (`lightdash-dbt-2.0.json`)
- **Validation**: `validateParameterConfiguration(parameters, scope: 'project' | 'model')` — rejects `required: true` at project scope; wired into `ProjectService.replaceProjectParameters`
- **Compiler helper**: `getAvailableParametersFromTables(tables, baseTable?)` — strips `required` from non-base tables when `baseTable` is provided
- **Frontend explore**: `ParametersCard` filter, `useExplorerQueryManager` missing-required computation, render condition in `Explorer/index.tsx`
- **Frontend dashboard**: `Dashboard.tsx` filter, `DashboardProvider` (OR-merge in `addParameterDefinitions` + `missingRequiredParameters`), per-tab filter in `dashboardTabs/index.tsx`, embed dashboard parameters

## Test plan
- [x] Unit tests for parameter validation and `getAvailableParametersFromTables` (45 tests passing)
- [x] Manual: customers explore shows both required params with no fields selected
- [x] Manual: orders explore (which joins customers) does **not** show customers' required params
- [x] Manual: events explore (no customers relation) shows no Parameters card
- [x] Manual: required + default ("EMEA") doesn't block query
- [x] Manual: required without default red-outlines the input and blocks "Run query"
- [x] Manual: dashboard with customers chart shows both required params in the filter bar
- [x] Manual: API rejects project-level `required: true` with a clear `ParameterError`
- [x] Frontend + common typecheck clean

### DIFF
--- a/examples/full-jaffle-shop-demo/dbt/models/customers.yml
+++ b/examples/full-jaffle-shop-demo/dbt/models/customers.yml
@@ -11,6 +11,23 @@ models:
             options_from_dimension:
               dimension: "first_name"
               model: "customers"
+          required_tenant:
+            label: "Tenant (required, no dimension reference)"
+            description: "Demo of a required parameter that has no underlying dimension reference. It is always shown in the Customers explore and the query is blocked until a value is selected."
+            required: true
+            options:
+              - "acme"
+              - "globex"
+              - "stark-industries"
+          required_region_with_default:
+            label: "Region (required with default)"
+            description: "Demo of a required parameter that has a default — always shown but does not block the query."
+            required: true
+            default: "EMEA"
+            options:
+              - "AMER"
+              - "EMEA"
+              - "APAC"
         primary_key: customer_id
         spotlight:
           categories: ["experimental"]
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -159,6 +159,7 @@ import {
     UserAccessControls,
     UserAttributeValueMap,
     UserWarehouseCredentials,
+    validateParameterConfiguration,
     VizColumn,
     WarehouseClient,
     WarehouseConnectionError,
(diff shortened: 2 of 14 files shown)


oli-rmsy:
A couple of screenshots showing the feature working end-to-end:

**Customers explore — required parameters surface immediately, no dimensions selected**

The `Tenant` parameter (no default) red-outlines and blocks the run; `Region` resolves to its `EMEA` default and doesn't block.

Customers explore with required parameters

**Jaffle dashboard filter bar — required parameters from the customers chart appear alongside existing filters**

Dashboard tile based on the customers explore contributes its required params; they show in the bar even though no dimension references them.

Dashboard filter bar with required parameters

owlas:
Fix here: https://github.com/lightdash/lightdash/pull/23369

PR #22732, opened 2026-05-06
chore: migrate `DashboardComments` to config-based flag

pimmey:
### Description:

Migrates the dashboard comments feature flag from the dynamic `FeatureFlagModel` / `LIGHTDASH_DISABLE_FEATURE_FLAGS` system to a static server-side config field (`dashboardComments.enabled`), controlled by the existing `DISABLE_DASHBOARD_COMMENTS` environment variable.

Previously, the `DISABLE_DASHBOARD_COMMENTS` env var was translated into a disabled feature flag entry and checked at runtime via `FeatureFlagModel.get()` on every comment operation. Now, the value is parsed once at startup into `lightdashConfig.dashboardComments.enabled` and exposed through the health endpoint, removing the need for `FeatureFlagModel` in `CommentService` entirely.

On the frontend, `useDashboardCommentsCheck` no longer calls `useServerFeatureFlag` and instead reads `health.data?.dashboardComments.enabled` directly, keeping the default-on behaviour for self-hosters who have not set the env var.

The `DashboardComments` entry has been removed from the `FeatureFlags` enum as it is no longer backed by the feature flag infrastructure.

### DIFF
--- a/packages/backend/src/config/lightdashConfig.mock.ts
+++ b/packages/backend/src/config/lightdashConfig.mock.ts
@@ -336,6 +336,9 @@ export const lightdashConfigMock: LightdashConfig = {
         enabled: false,
         retentionDays: 30,
     },
+    dashboardComments: {
+        enabled: true,
+    },
     preAggregates: {
         enabled: false,
         parquetEnabled: false,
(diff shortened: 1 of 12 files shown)


pimmey:
* **#22732** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22732?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22732?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22737, opened 2026-05-06
feat(managed-agent): group activity log by run

joaoviana:
Closes: 

### Description

The Autopilot activity log was a flat reverse-chronological list. After a few scheduled runs it became hard to tell where one run ended and the next began. This groups actions by `session_id` (already on the action shape — no backend change) so each Autopilot run renders as a collapsible section with a header.

- Frontend-only. The actions API and types are untouched.
- The first 3 runs render expanded by default; older runs collapse behind a "Show N older runs" affordance.
- Run header shows relative time + per-action-type count pills + total. Clicking the header toggles collapse.
- Clicking an action inside a hidden run auto-expands the run *and* reveals the older bucket so the detail sidebar context isn't orphaned.
- Open/closed state resets on reload (intentional — no localStorage churn).
- Polling: new runs arriving via the 30s refetch don't push previously-opened runs out of their default-open state.
- Visual cleanup along the way: dropped the now-vestigial timestamp column inside groups (run header carries the time), softened the table card so it doesn't clash with the page bg (transparent wrapper, action rows explicit white, headers transparent), neutralised the previously-blue selected-row colour with a small left accent.

### Out of scope (deferred)

- Showing runs that produced 0 actions ("agent ran cleanly"). Needs a small backend addition (last-heartbeat timestamp on the settings response). Easy follow-up; orthogonal to grouping.

### Test plan

- [ ] Land on the activity page with multiple runs — first 3 expanded, older collapsed
- [ ] Click a run header — collapses; click again — expands
- [ ] Click "Show N older runs" — reveals older run headers (collapsed by default)
- [ ] Click an action inside an older (hidden) run — older bucket reveals + the run auto-expands
- [ ] Wait 30s for the polled refetch — open runs stay open, new runs land at the top
- [ ] Toggle dark mode — table colours read correctly in both schemes
- [ ] Run a heartbeat with the Run-now button — UI updates after the polled refetch

PR #22738, opened 2026-05-06
feat(managed-agent): animate the home-card pixel grid

joaoviana:
Closes: 

### Description

Stacked on #22737. The Autopilot home card has a static SVG pixel grid on the right edge — purely decorative. This makes it feel alive with two CSS-only effects, and fixes a pre-existing rendering issue along the way.

**Per-cell twinkle.** Each rect now exposes its base opacity as a CSS custom property (`--rect-opacity`), and a keyframe animates from `var(--rect-opacity)` to `calc(var(--rect-opacity) * 3.2)` and back. Four overlapping `nth-child` selectors (`5n+1`, `7n+3`, `11n+4`, `13n+7`) with different durations (4.5–7s) and delays put each affected cell on its own cycle, so there's no flat strobe. Cells matching multiple selectors take the last one's animation, which adds extra de-sync for free.

**Whole-grid drift.** The SVG slowly wanders on a 22s cycle (`translate(0,0) → (-4,2) → (3,-2) → (0,0)`) — small enough to be subliminal, big enough to make the card feel like it's breathing.

**Fixed pixel rendering.** Previously the SVG used `preserveAspectRatio="xMaxYMid slice"` with `height="100%"`, which scaled cells up dramatically in tall containers and cropped them aggressively in narrow ones. Switched to fixed dimensions (252×270, no `preserveAspectRatio`) so cells always render at their native 7×7 size and 9px column spacing, regardless of card aspect ratio. Increased row count from 12 to 30 so the pattern reaches further down without empty space below in taller card variants.

Both animations honour `prefers-reduced-motion`. No JS, no per-frame work, no layout thrash — pure compositor properties (opacity + transform).

### Test plan

- [ ] Open the home page with Autopilot enabled — pixel grid twinkles softly *and* slowly drifts
- [ ] Confirm twinkle isn't synchronised across cells (no flat strobe pulse)
- [ ] Confirm cells render at consistent native size regardless of card width/height
- [ ] Toggle dark mode — both effects still subtle
- [ ] Set OS reduced-motion preference — both effects stop, cells stay at base opacity in their resting positions

### DIFF
--- a/packages/frontend/src/ee/features/managedAgent/ManagedAgentHomeCard.module.css
+++ b/packages/frontend/src/ee/features/managedAgent/ManagedAgentHomeCard.module.css
@@ -66,20 +66,66 @@
     position: absolute;
     top: 0;
     right: 0;
-    bottom: 0;
     color: light-dark(
         var(--mantine-color-violet-4),
         var(--mantine-color-violet-6)
     );
     pointer-events: none;
-    transition:
-        transform 300ms var(--ease-out-quart),
-        opacity 300ms var(--ease-out-quart);
+    transition: opacity 300ms var(--ease-out-quart);
     opacity: 0.6;
+    animation: pixelDrift 22s ease-in-out infinite;
 }
 
 .card:hover .pixelGrid {
     opacity: 1;
+}
+
+@keyframes pixelDrift {
+    0%,
+    100% {
+        transform: translate(0, 0);
+    }
+    33% {
+        transform: translate(-4px, 2px);
+    }
+    66% {
+        transform: translate(3px, -2px);
+    }
+}
+
+.pixelGrid rect {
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#22738** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22737** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22737?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #22739, opened 2026-05-06
fix: throw better errors on Athena client failures

magnew:
### Description:
Athena auth/credential failures now return a classified, actionable error instead of a generic WarehouseQueryError wrapping the bare AWS message.

  - All four AWS SDK catch sites (streamQuery, getCatalog, getAllTables, getFields) route through a new translateAthenaError helper.
  - Known auth errors (UnrecognizedClientException, InvalidSignatureException, ExpiredTokenException, CredentialsProviderError, etc.) become WarehouseConnectionError — matching what
  Snowflake/BigQuery/Databricks already do for connection-time failures.
  - Messages now include the AWS error code, HTTP status, and a one-line hint:
  [UnrecognizedClientException 400] The security token included in the request is invalid. AWS rejected the access key ID. Check the access key in your project settings is correct and the IAM user
  is enabled.

  On fragility

  Branching on error.name is the standard pattern across our warehouse clients. Three safeguards:

  1. error.name is part of AWS SDK v3's documented error contract; SDK version is pinned.
  2. HTTP 401 fallback catches future auth-error names we haven't catalogued. We deliberately don't fall back on 403 (also covers IAM permission gaps, which are query-time issues).
  3. If both signals miss, behavior degrades to the previous baseline — we never lose information, only fail to upgrade the classification.

### DIFF
--- a/packages/warehouses/src/warehouseClients/AthenaWarehouseClient.test.ts
+++ b/packages/warehouses/src/warehouseClients/AthenaWarehouseClient.test.ts
@@ -1,6 +1,8 @@
 import {
     AthenaAuthenticationType,
     CreateAthenaCredentials,
+    WarehouseConnectionError,
+    WarehouseQueryError,
     WarehouseTypes,
 } from '@lightdash/common';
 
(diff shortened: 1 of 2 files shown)


magnew:
* **#22739** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22739?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22739?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22740, opened 2026-05-06
feat(managed-agent): add action category and previous-version metadata helpers

joaoviana:
## Summary

Foundation PR for the revert-actions work. Adds two helpers to `@lightdash/common` for the Autopilot (managed agent) feature:

1. `getManagedAgentActionCategory(actionType)` → `'undo' | 'dismiss'` — single source of truth for which action types are real reversals (mutate world state) vs log-only dismissals.
2. `getFixedBrokenMetadata(metadata)` → typed accessor for the `{ previousVersionUuid }` shape stored on `FIXED_BROKEN` actions, returns `null` for legacy actions that predate revert support.

## Why

Today `reverseAction` switch-cases over action types and conflates two semantically different operations: undoing a state change (e.g. restoring a soft-deleted chart) vs dismissing a log entry (e.g. dismissing a "stale" flag). The next PRs in this stack split that conceptual model and need a shared helper to derive category, plus a typed reader for the metadata field used to revert chart fixes.

This PR is pure additions — no callers change, no behavior changes. Sets up the next two PRs.

## Categories

| Action type | Category |
|---|---|
| `SOFT_DELETED` | undo |
| `CREATED_CONTENT` | undo |
| `FIXED_BROKEN` | undo |
| `FLAGGED_STALE` | dismiss |
| `FLAGGED_BROKEN` | dismiss |
| `FLAGGED_SLOW` | dismiss |
| `INSIGHT` | dismiss |

## Regression analysis

| Group | Effect |
|---|---|
| All users | Pure additions to `@lightdash/common` exports. Nothing imports them yet. |

## Test plan

- [ ] CI build of `@lightdash/common` succeeds
- [ ] No-op runtime: PR doesn't change any behavior, only exposes new symbols

PR #22741, opened 2026-05-06
feat(managed-agent): make fix-broken chart actions reversible

joaoviana:
Closes PROD-7453

## Summary

Makes Autopilot's `FIXED_BROKEN` actions reversible. Stacks on top of #22740.

Before this PR, when a user clicked "Revert" on a chart-fix action, the agent's switch-case fell through to the "log-only" branch and only flipped `reversedAt` — the chart kept the agent's version. Workaround was to manually use the chart history dropdown.

## How it works

Two coordinated changes in `ManagedAgentService`:

**Write path (`handleFixBrokenChart`):**
- Before `createVersion` is called, fetch the chart's current latest version via `savedChartModel.getLatestVersionSummary(chartUuid)`.
- Store its UUID in the action metadata as `{ previousVersionUuid }`.

**Read path (`reverseAction` → `restorePreviousChartVersion`):**
- Read `previousVersionUuid` via the typed `getFixedBrokenMetadata` helper from common.
- Fetch the prior chart payload via `savedChartModel.get(chartUuid, previousVersionUuid)`.
- Apply it as a new version via `createVersion` — preserves history rather than rewriting.

## Why store the UUID instead of computing from DB

Initial intuition was "use the existing `rollbackToVersionAtTimestamp` API with `action.createdAt - 1ms`." That's racy: `createVersion` runs *before* `createAction`, so `action.createdAt` lands *after* the agent's version's `created_at` — the rollback target is the agent's own version. Computing from session+user+chart in DB is fragile too. Capturing the version UUID at write time is one extra field, deterministic, and survives subsequent edits.

## Behavioral changes

1. **Legacy `FIXED_BROKEN` actions throw a clear error on revert** instead of silently no-oping. Existing actions in the DB don't have `previousVersionUuid`. `restorePreviousChartVersion` throws a `ParameterError` with a message pointing the user to the chart's version history.
2. **`FLAGGED_SLOW` is now an explicit dismiss case** instead of falling through to the default branch. Same behavior, more obvious intent.
3. **Analytics `managed_agent.action_reversed` event gets a new `actionCategory` property** (`'undo' | 'dismiss'`) so we can distinguish real reversals from dismissals.

## Out of scope

- Warning the user if there are newer human-authored versions between the agent fix and now (worth its own UX design).
- Re-flagging the chart as broken after a successful revert (`validateProject` will rediscover on next run).
- Diff visualization between fixed and pre-fix versions (separate Slack feedback item, not in PROD-7453's scope).

## Regression analysis

| Group | Effect |
|---|---|
| Users with newly-created `FIXED_BROKEN` actions | Revert button now actually restores the previous version (previously a silent no-op). |
| Users with legacy `FIXED_BROKEN` actions | Revert now throws a typed error with a clear "use chart history" message instead of silently succeeding. |
| Users with other action types | No change. |
| Users without Autopilot enabled | No change. |
| Service accounts / cross-project | No change — per-project authorization unchanged. |

The legacy-action error path is acceptable because the prior behavior was actively misleading.

## Test plan

- [ ] Trigger a real `FIXED_BROKEN` action; verify `metadata.previousVersionUuid` is populated in the DB
- [ ] Click "Revert" → chart's latest version reverts to the pre-fix payload (new version with old content)
- [ ] Manually clear `metadata` on an action and click "Revert" → toast shows "predates revert support" error
- [ ] Reverse a `SOFT_DELETED` action → still restores content
- [ ] Reverse a `CREATED_CONTENT` action → still soft-deletes the agent-created chart
- [ ] PostHog: confirm new `actionCategory` property appears on `managed_agent.action_reversed` events

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -1391,6 +1391,7 @@ export type ManagedAgentActionReversedEvent = BaseTrack & {
         organizationId: string;
         projectId: string;
         actionType: string;
+        actionCategory: 'undo' | 'dismiss';
         targetType: string;
         sessionId: string;
         actionAgeMs: number;
--- a/packages/backend/src/ee/services/ManagedAgentService/ManagedAgentService.ts
+++ b/packages/backend/src/ee/services/ManagedAgentService/ManagedAgentService.ts
@@ -2,10 +2,13 @@ import { subject } from '@casl/ability';
 import {
     FeatureFlags,
     ForbiddenError,
+    getFixedBrokenMetadata,
+    getManagedAgentActionCategory,
     getManagedAgentScheduleCron,
     ManagedAgentActionType,
     ManagedAgentTargetType,
     NotFoundError,
+    ParameterError,
     ProjectType,
     ServiceAccountScope,
     type ChartConfig,
@@ -517,11 +520,14 @@ export class ManagedAgentService extends BaseService {
                     );
                 }
                 break;
+            case ManagedAgentActionType.FIXED_BROKEN:
+                await this.restorePreviousChartVersion(action);
+                break;
             case ManagedAgentActionType.FLAGGED_STALE:
             case ManagedAgentActionType.FLAGGED_BROKEN:
+            case ManagedAgentActionType.FLAGGED_SLOW:
             case ManagedAgentActionType.INSIGHT:
-            case ManagedAgentActionType.FIXED_BROKEN:
-                // These are log-only entries — marking as reversed dismisses them
+                // Log-only entries — marking as reversed dismisses them
                 break;
             default:
                 break;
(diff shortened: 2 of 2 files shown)


joaoviana:
* **#22746** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22742** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22741** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22741?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22741?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22740** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22740?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #22753, opened 2026-05-06
feat(timezone): consult disableTimestampConversion in shouldSkipTimezoneConversion

almeidabbm:
Linear: https://linear.app/lightdash/issue/GLITCH-413

## Why

`shouldSkipTimezoneConversion` hardcodes `effectiveInputTz='UTC'` for Snowflake, assuming the compile-time `CONVERT_TIMEZONE` wrap normalized the input. That assumption breaks when a project sets `disableTimestampConversion: true` on Snowflake credentials - values flow through in `dataTimezone`, not UTC. The skip-check can then either skip a needed DATE_TRUNC localization (silent misdisplay) or emit a redundant identity conversion.

## What

Plumb `whSkipUtcNormalization?: boolean` through `BuildQueryProps` → `ProjectService._compileQuery`, sourced from `credentials.disableTimestampConversion` (Snowflake-only). The skip-check now falls back to `dataTimezone` for Snowflake when the flag is set, mirroring the non-Snowflake branch.

## Test plan

- [x] Unit tests for the four config combinations (Layer 1 active vs disabled × project tz === or !== input tz)
- [x] Backend typecheck passes
- [x] Backend lint passes
- [ ] Manual curl verification on a Snowflake project with `disableTimestampConversion: true` + non-UTC `dataTimezone` (sanity-check on a real warehouse — multi-warehouse tests are not in CI)

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -3105,6 +3105,15 @@ export class ProjectService extends BaseService {
         return { explore, dateZoomApplied: false };
     }
 
+    private static getSnowflakeDisableConvertTimezoneWrap(
+        credentials: CreateWarehouseCredentials,
+    ): boolean {
+        return (
+            credentials.type === WarehouseTypes.SNOWFLAKE &&
+            credentials.disableTimestampConversion === true
+        );
+    }
+
     static async _compileQuery({
         metricQuery,
         explore,
@@ -3120,7 +3129,7 @@ export class ProjectService extends BaseService {
         continueOnError,
         useTimezoneAwareDateTrunc,
         dataTimezone,
-        whSkipUtcNormalization,
+        snowflakeDisableConvertTimezoneWrap,
     }: {
         metricQuery: MetricQuery;
         explore: Explore;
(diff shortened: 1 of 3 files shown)


almeidabbm:
* **#22753** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

almeidabbm:
Closing — superseded by GLITCH-414.

Manual end-to-end verification on a Snowflake project with `disableTimestampConversion: true` + non-UTC `dataTimezone` showed the skip-check fix in this PR is insufficient. The deeper bug is that `dateTruncTimezoneConversions[SNOWFLAKE].toProjectTz` (and the EXTRACT equivalent) hardcodes the source timezone as `'UTC'`, so even with the skip-check corrected, the SELECT-list wrap still emits structurally wrong SQL in `disable=true` configs.

GLITCH-413 is canceled. The full end-to-end fix is scoped under GLITCH-414 (Backlog), which addresses the source-tz hardcoding, the skip-check, and the credential plumbing together.

PR #22802, opened 2026-05-07
fix: align BASE_ROLE_SCOPES with static ability builders, add scope parity tests

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7469/bug-creating-a-role-from-admin-doesnt-inherit-all-the-scopes
Closes: https://linear.app/lightdash/issue/PROD-7480/confusing-scope-duplicate

### Description:

Fixes drift between `BASE_ROLE_SCOPES` (the source for custom roles created via "Duplicate <system role>") and the static ability builders. Before this change, duplicating system Admin gave a custom role missing ~23 scopes — primarily the org-management bundle (`manage:Group`, `manage:Organization`, `impersonate:User`, etc.) plus a handful of granular project-level scopes. After this change, the admin tier of `BASE_ROLE_SCOPES` is the union of every static ability granted at admin, at both project and org level (91 of 91 scopes in the vocabulary). Adds two new parity tests so this drift is caught in CI.

### What changes

**`BASE_ROLE_SCOPES` aligned with static ability builders for each tier:**

- `viewer`: + `view:Organization`, `view:OrganizationMemberProfile`
- `interactive_viewer`: + `create:Job`, `view:Job`, `view:Job@self`, `manage:SemanticViewer@space`
- `editor`: + `manage:SemanticViewer`, `view:OrganizationWarehouseCredentials`
- `developer`: + `create:VirtualView`, `delete:VirtualView`, `promote:Dashboard`, `promote:Dashboard@space`, `promote:SavedChart`, `promote:SavedChart@space`
- `admin`: + the full org-management bundle (`manage:OrganizationMemberProfile`, `manage:Group`, `manage:InviteLink`, `manage:GitIntegration`, `manage:OrganizationWarehouseCredentials`, `manage:Organization`, `impersonate:User`, `manage:PersonalAccessToken`)

**New scope:** `view:OrganizationWarehouseCredentials` (enterprise) — the org editor static ability granted it but no scope existed in the vocabulary, leaving editors with an unrepresented capability.

**Vocabulary cleanup:** removed `export:DashboardCsv`, `export:DashboardImage`, `export:DashboardPdf` from `scopes.ts`, `CaslSubjectNames`, and the viewer tier. They were registered ~9 months ago in #16460 but never enforced at any endpoint — `manage:ExportCsv` is the actual gate (`CsvService.scheduleExportCsvDashboard()`, `GdriveService.scheduleUploadGsheet()`, plus 6 frontend usages). They only existed as dead toggles in the role-builder UI. Per-format granularity can be re-introduced later by splitting `manage:ExportCsv` together with matching endpoint enforcement.

**Three parity tests in `roleToScopeParity.test.ts`:**

1. **Project parity (per role)** — every action+subject the project ability builder grants must be reachable through some scope in `BASE_ROLE_SCOPES`.
2. **Org parity (per role)** — same check against `applyOrganizationMemberStaticAbilities`. The test that was missing — would have caught the org-management drift early.
3. **Vocabulary coverage** — every scope in `scopes.ts` must appear in at least one tier of `BASE_ROLE_SCOPES`.

Comparison logic switched from strict rule-set equivalence (brittle against condition-shape differences) to "every role-builder rule reachable through scopes" (extras allowed).

### Runtime impact

- **Existing system-role users** (no `roleUuid`): unchanged — they go through the static ability builders, not `BASE_ROLE_SCOPES`.
- **Existing custom roles**: unchanged — their scope snapshot is stored in `scoped_roles` at creation time and is not migrated.
- **New custom roles via "Duplicate <system role>"**: pick up the expanded set on creation.
- **Service accounts assigned `roleUuid: system:admin` (etc.)**: read `getSystemRoles()` → `BASE_ROLE_SCOPES` at runtime, so they immediately see the new scopes. Intended outcome — makes SAs assigned a system role behave as the role's name implies.

### ⚠️ PAT bypass — reviewer note

`manage:PersonalAccessToken` is now in `BASE_ROLE_SCOPES.admin`. Self-hosted deployments use `PAT_ALLOWED_ORG_ROLES` to restrict PAT creation to specific tiers — but **CASL is additive**, so any custom role with this scope ticked bypasses that gate. Trade-off accepted because the alternative was leaving PAT off the admin scope set entirely, which would block SAs with system Admin role from managing PATs. Operators concerned about the bypass should untick the scope when they create lower-privilege custom roles. Documented inline in `roleToScopeMapping.ts`.

### Documentation

`docs/authentication-and-roles.md`:

- Drift safeguards section now describes all three parity tests and what each one catches.
- "Project vs organization assignment" section gets a concrete example table, explanation of CASL additivity as an architectural safeguard against project-level escalation, and a structured list of options for the known UX gap (org-only scopes show as ticked at project assignment but silently no-op).

User-facing docs follow-up needed in `lightdash/mintlify-docs` (separate PR):

- `references/workspace/custom-roles.mdx` still says custom roles are project-only — outdated now that org-level assignment is real.
- Project-vs-org silent-no-op pitfall needs a Warning callout.
- PAT bypass needs a note where `PAT_ALLOWED_ORG_ROLES` is documented.


rephus:
* **#22802** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22802?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22802?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/authorization/scopes.ts
```diff
@@ -621,29 +628,6 @@ const scopes: Scope[] = [
         getConditions: addDefaultUuidCondition,
     },
 
-    // Sharing Scopes
-    {
-        name: 'export:DashboardCsv',
```
rephus, line 626: they were not used anywhere.

rephus:
Investigating API test error

> PR #22804 (unrelated Snowflake timezone fix) also fails on the exact same test at line 243. So this isn't caused by #22802 — it's a regression already on main.

PR #22804, opened 2026-05-07
fix(timezone): thread column timezone through Snowflake DATE_TRUNC/EXTRACT wrap

almeidabbm:
## Problem

Snowflake projects with `disableTimestampConversion: true` and a non-UTC `dataTimezone` produced wrong timezone-aware `DATE_TRUNC` / `EXTRACT` SQL — the wrap hardcoded `UTC` as the column source TZ, but with the dbt translator wrap disabled, columns are actually in `dataTimezone`.

## Fix

Replace the `dataTimezone` + `columnInDataTimezone` pair on `BuildQueryProps` with a single derived `columnTimezone: string`. Compute it once at the service boundary via `getColumnTimezone(credentials)`, co-located with the credential types in `projects.ts`.

- Snowflake: returns `UTC` (translator wraps timestamps with `CONVERT_TIMEZONE('UTC', col)`) unless `disableTimestampConversion` opts out — then returns `dataTimezone`.
- Other adapters: returns `dataTimezone` (defaulting to `UTC`).

`MetricQueryBuilder` no longer carries the Snowflake `if`. The conversion maps in `timeFrames.ts` accept `sourceTimezone` as an explicit parameter.

Also in this PR:
- New unit tests cover the non-UTC `sourceTimezone` path through both `dateTruncTimezoneConversions` / `dateExtractsTimezoneConversions` and end-to-end via `getSqlForTruncatedDate` / `getSqlForDatePart`.
- `docs/timezone-handling.md` updated to describe the `getColumnTimezone`-derived source TZ and the `disableTimestampConversion` case.

## Test plan

- [x] `pnpm -F common typecheck`
- [x] `pnpm -F common test -- timeFrames` (79 pass)
- [x] `pnpm -F backend test -- MetricQueryBuilder` (175 pass)
- [x] Manual repro on Snowflake — see below

## Manual repro

Setup: two Snowflake projects pointing at the same warehouse, both `dataTimezone: America/New_York`. One has `disableTimestampConversion: false` (control), the other `true` (the broken case). Same query on `timezone_test`: count rows by day on the NTZ column, sorted by day.

**Bug, in one row:** the NTZ column has a row at `2024-01-01 00:00:00` — that's midnight NY because the project's `dataTimezone` is NY. With NY as the project tz too, that row should land in the `2024-01-01` bucket.

| | bucket for `2024-01-01 00:00:00` (NTZ, NY local) |
|---|---|
| control project | `2024-01-01` ✅ |
| broken project, **pre-fix** (main) | `2023-12-31` ❌ — treats NTZ as UTC, shifts 5h back |
| broken project, **post-fix** (this PR) | `2024-01-01` ✅ |

Why: the inner `CONVERT_TIMEZONE` source argument was hardcoded to `UTC`. Fix sets it to the column's actual timezone (`dataTimezone`):

```diff
- CONVERT_TIMEZONE('UTC',              'America/New_York', event_timestamp_ntz)
+ CONVERT_TIMEZONE('America/New_York', 'America/New_York', event_timestamp_ntz)
```

Full row-by-row counts (project tz = `America/New_York`):

| day bucket | control | broken (pre-fix) | broken (post-fix) |
|---|---|---|---|
| 2023-12-31 | — | 1 | — |
| 2024-01-01 | 1 | — | 1 |
| 2024-01-14 | 1 | 3 | 1 |
| 2024-01-15 | 7 | 6 | 7 |
| 2024-01-16 | 4 | 3 | 4 |

When project tz = `UTC`, the buggy SQL collapses to a no-op (`CONVERT_TIMEZONE('UTC', 'UTC', col)`), so all three columns match and the bug is hidden — that's why this only surfaces for non-UTC users.

Closes: https://linear.app/lightdash/issue/GLITCH-414/snowflake-disabletimestampconversion-non-utc-datatimezone-produces

### DIFF
--- a/docs/timezone-handling.md
+++ b/docs/timezone-handling.md
@@ -90,9 +90,11 @@ flowchart LR
 
 With `useTimezoneAwareDateTrunc` on, truncation is timezone-aware. The base dimension SQL is round-tripped through the project timezone so boundaries fall on project-local wall-clock midnights, but the returned value is still a real UTC instant:
 
-1. Shift the UTC value into project wall-clock
+1. Shift the column value from its source TZ into project wall-clock
 2. Truncate on that wall-clock
 3. Convert the truncated wall-clock back to UTC
+
+The source TZ for step 1 is derived once per query at the service boundary via `getColumnTimezone(credentials)` (in `packages/common/src/types/projects.ts`) and threaded through `timeFrames.ts` as `sourceTimezone`. It returns `'UTC'` for Snowflake when the translator wrap is active, `dataTimezone` when Snowflake's `disableTimestampConversion` opts out of that wrap, and `dataTimezone` (defaulting to UTC) for every other adapter. Most warehouses ignore it because their `toProjectTz` doesn't take a source TZ; Snowflake threads it into the inner `CONVERT_TIMEZONE`.
 
 The SQL differs per warehouse (some have native TZ-aware truncation, others compose `AT TIME ZONE` / `CONVERT_TIMEZONE` / `to_utc_timestamp`), but the shape is identical everywhere. Flag off → falls back to raw `DATE_TRUNC` grouping in UTC (old behavior).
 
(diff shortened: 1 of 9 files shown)


almeidabbm:
* **#22804** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22804?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22804?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22661, opened 2026-05-05
feat: provision dedicated user row per service account with tombstone-on-delete semantics

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7403/create-per-service-account-user-record
Closes: https://linear.app/lightdash/issue/PROD-7421/filter-out-service-account-users-from-listings-eg-list-users

### Done

Schema — 4 migrations applied, no behavior change at runtime yet:
  - users.is_service_account BOOLEAN NOT NULL DEFAULT FALSE (OSS)
  - service_accounts.service_account_user_uuid UUID UNIQUE NOT NULL FK→users (EE)
  - Backfill: every existing SA has a linked dedicated user + organization_memberships row, role mapped from scopes (admin/member)
  - NOT NULL promotion after backfill

  Provisioning — ServiceAccountModel.save creates users + organization_memberships + service_accounts in one transaction. Verified end-to-end via API.

  Tombstone delete — ServiceAccountModel.delete drops only the service_accounts row; user record persists so historical FKs (created_by_user_uuid)
  keep resolving.

  Filtered from human-facing surfaces — users.is_service_account = false enforced in:
  - OrganizationMemberProfileModel.queryBuilder (covers admin org-member listings + SCIM /Users)
  - OrganizationMemberProfileModel.findOldestAdminUserUuid (admin-fallback in getAdminUser)
  - OrganizationMemberProfileModel.getOrganizationMembersAndGroups
  - UserModel.findUserByEmail / findSessionUserByPrimaryEmail / getUserByPrimaryEmailAndPassword

  Tests — 10/10 in serviceAccounts.test.ts, 179/179 across organizationPermissions/roles/projectPermissions/groups. New api-tests cover:
  create→auth→delete lifecycle, duplicate descriptions, /org/users no-leak, SCIM /Users no-leak.

  Verified by direct curl: /org/users returns 0 SA rows; SCIM /Users returns 79 users with 0 SA-typed.

###  Not done

  Phase C — auth middleware cutover. Today, when an SA hits the API:
  - Middleware still calls getAdminUser and spoofs the admin's identity onto req.user.
  - Live requests and content updates stamp the seeded admin's UUID on every created_by_user_uuid / updated_by_user_uuid column.
  - API responses still show David Attenborough (or whichever admin is the fallback) as the actor — not the service account.
  - Audit log still emits the "attribution caveat" naming the admin.

[...]

PR #22809, opened 2026-05-07
feat: set `service_account_user_uuid` NOT NULL (PROD-7427)

rephus:
Closes: #22661

### Description:

Adds a follow-up migration to enforce `NOT NULL` on the `service_account_user_uuid` column in the `service_accounts` table. The column was intentionally left nullable when it was first introduced to allow safe rolling deploys and rollbacks. Now that the stability window has passed, this migration tightens the constraint.

The migration is wrapped in a `try/catch` so that if any stray `NULL` rows exist (e.g. created by an old-code replica during the deploy window), the migration will log a warning and continue rather than blocking the deploy chain. Affected rows can be backfilled manually and the migration re-run.

The `down` migration reverts the column back to nullable.

### DIFF
--- a/packages/backend/src/ee/database/migrations/20260507151700_set_service_account_user_uuid_not_null.ts
+++ b/packages/backend/src/ee/database/migrations/20260507151700_set_service_account_user_uuid_not_null.ts
@@ -1,0 +1,39 @@
+import { Knex } from 'knex';
+
+const ServiceAccountsTableName = 'service_accounts';
+const ColumnName = 'service_account_user_uuid';
+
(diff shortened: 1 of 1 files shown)


rephus:
* **#22809** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22809?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22815, opened 2026-05-07
feat(managed-agent): diff and revert CTA for auto-fixed charts

joaoviana:
Closes [PROD-7391](https://linear.app/lightdash/issue/PROD-7391/show-diff-and-revert-cta-for-auto-fixes).

When Autopilot auto-fixes a broken chart, surface what changed and give a prominent revert path. Mirrors the existing soft-deleted "Restore" banner pattern at the top of the action sidebar.

## What's in

- **Auto-fixed banner**: teal-tinted banner with primary `Revert` button (uses the existing `reverseAction` wiring — no new mutation). Hidden once the action is reversed; the existing "Undone" label takes over.
- **Legacy fix fallback**: actions recorded before `previousVersionUuid` was tracked already fail revert with a friendly error (`ManagedAgentService.ts:674`). The button is disabled with a tooltip pointing at chart history for those.
- **`FixedBrokenDiff` component**: semantic diff alongside `ChartDetails` / `BrokenDetails` in the sidebar:
    - **Dimensions / Metrics / Table calculations**: added (green) and removed (red strikethrough) field-pill chips, reusing the existing `fieldPill` style with new `fieldPillAdded` / `fieldPillRemoved` colour variants.
    - **Chart type**, **explore name**: before → after text rows when changed.
    - **Filters**: rule count delta only — no filter-tree rendering for v1 (intentional — filter trees get unreadable fast).
    - **Empty diff fallback**: if nothing surface-level changed, shows a "see chart history for details" message.
    - Always renders a **View full chart history** link as a punch-through to the existing version-history page.
- **Attribution fix** (`ManagedAgentService.ts`): `handleFixBrokenChart` and `restorePreviousChartVersion` previously passed `undefined` as the user to `savedChartModel.createVersion`, leaving `last_version_updated_by_user_uuid` NULL. The frontend rendered that as "Last modified by null null" (visible the moment we exposed it in the new sidebar). Auto-fixes now attribute to the human admin who enabled Autopilot (`getAutopilotActor`); reverts attribute to the human clicker.

## Out of scope

- Dashboard auto-fixes (the agent only fixes charts today via `fix_broken_chart`)
- Filter-rule-level diffing (count delta only)
- Rendered-output comparison (would require a query)
- Replacing the existing chart version-history page

## Regression analysis

Purely additive frontend change in EE plus a one-line attribution fix in the existing `fix_broken_chart` / revert flow:
[...]

PR #22818, opened 2026-05-07
fix(managed-agent): refresh runs and latest-run caches after revert

joaoviana:
Follow-up to #22815.

The action-sidebar's revert mutation only invalidated the `managed-agent-actions` cache, leaving the `managed-agent-runs` and `managed-agent-latest-run` caches stale. After clicking **Undo** or **Dismiss**, the action row in the run-grouped table didn't update (no dimming, no strikethrough) until the user manually refreshed the page, because `RunGroup` also reads from the runs list (action counts, trigger badges) which never re-fetched.

Every other mutation in `ManagedAgentActivityPage.tsx` already invalidates all three caches (run-now at lines 1789-1793, settings at lines 1883-1891) — the revert path was the only outlier. This PR brings it into line.

## Regression analysis

Purely additive — adds two more `invalidateQueries` calls in the existing `revertMutation.onSuccess` handler:
- For users without managed-agent enabled — never executed, no impact.
- For users with managed-agent who never click revert — never executed, no impact.
- For users clicking revert — they now also pay one tick of refetch on the runs list and latest-run query (both already polling on a 30s/3s interval anyway). No new endpoints, no schema changes.

## Test plan

- [ ] Open the Autopilot activity page, expand a run, click an action → click **Undo** (or **Dismiss** for FLAGGED_*/INSIGHT) → action row dims with strikethrough immediately, no manual refresh required
- [ ] Run-row counts/dots reflect the post-revert state without refresh

### DIFF
--- a/packages/frontend/src/ee/features/managedAgent/ManagedAgentActivityPage.tsx
+++ b/packages/frontend/src/ee/features/managedAgent/ManagedAgentActivityPage.tsx
@@ -790,6 +790,12 @@ const DetailSidebar: FC<{
             void queryClient.invalidateQueries({
                 queryKey: ['managed-agent-actions', action.projectUuid],
             });
+            void queryClient.invalidateQueries({
+                queryKey: ['managed-agent-runs', action.projectUuid],
+            });
+            void queryClient.invalidateQueries({
+                queryKey: ['managed-agent-latest-run', action.projectUuid],
+            });
         },
         onError: ({ error }) => {
             const isHardDeleted =


joaoviana:
* **#22819** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22819?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22818** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22818?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joaoviana:
Closing — turns out this fix is a no-op. React Query v4's invalidateQueries already fuzzy-matches by prefix, so 'managed-agent-actions' invalidation cascades to per-run keys. And the runs-list 'actionCountsByType' is COUNT(*) GROUP BY action_type without any reversed_at filter, so revert can't change those counts anyway. The original 'UI doesn't update after revert' bug is actually a stale React state reference in the parent's selectedAction — fix coming in a separate PR.

PR #22819, opened 2026-05-07
feat(managed-agent): track run started, run completed, and action created

joaoviana:
Adds three new analytics events to fill the run-cycle gap. Existing events only covered settings changes, manual run-now triggers, and reverts — nothing fired when a run actually **started**, **finished**, or recorded an **agent action**.

## New events

| Event | Fires | Key properties |
|---|---|---|
| `managed_agent.run_started` | top of `runHeartbeat` | `triggeredBy` (cron/manual/on_enable), `schedule`, `hasSlackChannel` |
| `managed_agent.run_completed` | both `failRunSafely` (config errors) and the main heartbeat `finally` (runtime errors + success) | `status`, `durationMs`, `actionCount`, `actionCountsByType`, `slackPosted`, `error` |
| `managed_agent.action_created` | after every successful `managedAgentModel.createAction` (5 call sites) | `actionType`, `targetType`, `runUuid`, `sessionId` |

Tool-call-level events were intentionally skipped — too high cardinality and the tool-name list shifts with prompt iterations. Run-level + action-level is the right resolution.

## What this unlocks

- **Adoption** — orgs/projects with active runs.
- **Activity** — runs per project per day, by trigger type.
- **Effectiveness** — actions per run, action-type distribution, what tool the agent leans on.
- **Reliability** — error rate, run duration distribution.
- **Engagement loop** — pair `action_created` with the existing `action_reversed` event to compute "human override rate" per `actionType`.

## Refactor

- Extracted a `HeartbeatContext` type + four small private helpers from `runHeartbeat`: `loadHeartbeatContext`, `trackRunStarted`, `trackRunCompleted`, `maybePostHeartbeatSlackSummary`. The heartbeat function now reads as a clean orchestration — load context, fire started, run, finalise (track completed inside `finally`).
- `failRunSafely` now also threads through `trackRunCompleted`, so config errors produce the same event shape as runtime errors (with `actionCount: 0` and the relevant error string).
- Slack post wrapped in `try/catch` so a Slack failure can never swallow the finally-block analytics or `finishRun` — the `slackPosted` field on `run_completed` reflects actual delivery.
- New model methods: `getRun(runUuid)` for the run row, and `getActionCountsByTypeForRun(runUuid)`. The latter replaces the old `countActionsForRun` call inside heartbeat (sum of values gives the total, no extra query).

## Regression analysis

Backend-only, additive analytics work in EE:
- For users **without managed-agent enabled** — `runHeartbeat` is never called, no impact.
- For users with managed-agent enabled — every run already paid the cost of `getSettings`, `getSummary`, `countActionsForRun`, `finishRun`. We add one extra `getRun(runUuid)` (single-row indexed read) at the top of heartbeat. `countActionsForRun` is replaced by `getActionCountsByTypeForRun` (one DB query instead of one — same network cost).
- Behavioural change to runHeartbeat is limited to: (a) Slack-post failures no longer abort the finally block, and (b) `failRunSafely` now also emits `run_completed`. Both desirable.
- `analyticsUserId = settings?.enabledByUserUuid ?? ''` — empty string fallback exists for the edge case where `getSettings` returns null. In practice, settings always exist when a run row exists; the fallback is purely a TypeScript guard.

## Test plan

- [ ] Trigger a Run-Now → `run_started` fires immediately, `run_completed` fires when the agent finishes (or after the 5 min `MANAGED_AGENT_SESSION_TIMEOUT_MS`)
- [ ] Disable Autopilot mid-flight → next cron tick fires `run_started` then `run_completed` with `status:'error'`, `error:'Autopilot disabled'`, `actionCount:0`
- [ ] Project with no service-account token → `run_started` then `run_completed` with `error:'No service account token'`
- [ ] During a run, every action the agent takes (flag, fix, soft-delete, create, insight) emits `action_created` with the right `actionType` / `targetType`
- [ ] Slack-enabled project → `run_completed.slackPosted` is `true` on success, `false` on Slack-post failure (and the failure does NOT prevent the rest of the finally block)

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -1398,11 +1398,57 @@ export type ManagedAgentActionReversedEvent = BaseTrack & {
     };
 };
 
+export type ManagedAgentRunStartedEvent = BaseTrack & {
+    event: 'managed_agent.run_started';
+    userId: string;
+    properties: {
+        organizationId: string;
+        projectId: string;
+        runUuid: string;
+        triggeredBy: 'cron' | 'manual' | 'on_enable';
+        schedule: string;
+        hasSlackChannel: boolean;
+    };
+};
+
+export type ManagedAgentRunCompletedEvent = BaseTrack & {
+    event: 'managed_agent.run_completed';
+    userId: string;
+    properties: {
+        organizationId: string;
+        projectId: string;
+        runUuid: string;
+        triggeredBy: 'cron' | 'manual' | 'on_enable';
+        status: 'completed' | 'error';
+        durationMs: number;
+        actionCount: number;
+        actionCountsByType: Record<string, number>;
+        slackPosted: boolean;
+        error: string | null;
+    };
+};
+
+export type ManagedAgentActionCreatedEvent = BaseTrack & {
+    event: 'managed_agent.action_created';
+    userId: string;
+    properties: {
+        organizationId: string;
+        projectId: string;
+        runUuid: string | null;
+        sessionId: string;
+        actionType: string;
+        targetType: string;
+    };
+};
+
 export type ManagedAgentEvent =
     | ManagedAgentSettingsCreatedEvent
     | ManagedAgentSettingsUpdatedEvent
     | ManagedAgentRunNowTriggeredEvent
-    | ManagedAgentActionReversedEvent;
+    | ManagedAgentActionReversedEvent
+    | ManagedAgentRunStartedEvent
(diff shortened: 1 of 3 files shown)


joaoviana:
* **#22819** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22820, opened 2026-05-07
fix(managed-agent): keep action sidebar in sync after revert

joaoviana:
After clicking **Revert/Undo/Dismiss** in the action sidebar, the table row updated immediately (cache invalidation propagates via React Query's prefix match) but the **sidebar kept showing the pre-revert state** — auto-fixed banner with active Revert button, no "Undone" label — until manual page refresh.

## Root cause

`ManagedAgentActivityPage` stores the selected action as a state reference:

```ts
const [selected, setSelected] = useState<ManagedAgentAction | null>(null);
// ...
const handleSelectAction = useCallback((action) => setSelected(action), []);
```

When the actions cache refetches after the revert mutation succeeds, the parent re-renders with a fresh actions array — but the stored `selected` reference still points at the **old** object (with `reversedAt: null`). `<DetailSidebar action={selected} />` receives stale data.

The table rows look reactive because they're rendered via `actions.map((action) => <ActionRow action={action} />)` — every refetch produces fresh objects mapped into fresh rows.

## Fix

Inside `DetailSidebar`, derive the live action from the mutation's response:

```ts
const liveAction: ManagedAgentAction = revertMutation.data ?? action;
```

`revertMutation.data` is the authoritative post-revert action (the API endpoint returns it). Use `liveAction` for every reactive computation: `isReversed`, `showFixedBanner`, `showRestoreBanner`, `isTargetDeleted`, `reversedLabel`, `previousVersionUuid`, etc. Stable fields read from the prop directly (no behavioural change).

Add `key={selected.actionUuid}` on the `<DetailSidebar>` in the parent so React fully remounts when the user picks a different action — otherwise the previous mutation's `data` would leak across selections.

## Type plumbing

- `reverseAction` is now typed `Promise<ManagedAgentAction>` and returns the API response.
- New `ApiManagedAgentActionResponse = ApiSuccess<ManagedAgentAction>` in common, added to the `ApiResults` union so `lightdashApi<ManagedAgentAction>` typechecks.
- Removed the duplicate frontend-local `ManagedAgentAction` type. The only difference was `string` vs `Date` for `reversedAt`/`createdAt` — moot in practice because every consumer wraps in `new Date(...)` which accepts both. Both call sites (`ManagedAgentActivityPage.tsx`, `useManagedAgentActions.ts`) now import from `@lightdash/common`.

## Regression analysis

Frontend-only, narrow surface:

- For users **without managed-agent enabled** — the page never renders, no impact.
- For users who never click revert/undo/dismiss — `revertMutation.data` is undefined, `liveAction` falls back to `action`, behaviour identical to before.
- For users who click revert and then immediately switch to a different action — the new `key` prop forces a fresh mount of the sidebar, so mutation state can't leak.
- For users who click revert and stay — sidebar updates instantly from `mutation.data` instead of waiting for the round-trip cache invalidation to thread through stale parent state.

## Test plan

- [ ] Open a FIXED_BROKEN action → click **Revert** → without refreshing, sidebar should immediately show "Undone X seconds ago by …" banner; auto-fixed banner + diff section should hide.
- [ ] Open a SOFT_DELETED action → click **Restore** → "Deleted by Autopilot" banner replaced by "Undone …" label without refresh.
- [ ] Open a FLAGGED_* action → click **Dismiss** → "Dismissed …" label appears without refresh.
- [ ] Click revert on Action A → switch to Action B (still un-reversed) → sidebar shows Action B's correct unreversed state, not Action A's mutation result.


joaoviana:
* **#22820** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22820?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22846, opened 2026-05-08
fix(oauth): persist rotated Snowflake & Databricks refresh tokens (#22831)

### LINKED ISSUE #22831: Snowflake OAuth: rotated refresh tokens are discarded, scheduled jobs fail with invalid_grant
## Summary

When a Snowflake OAuth security integration is configured with `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = TRUE` (or in any other scenario where the OAuth server returns a *new* refresh token alongside a fresh access token — e.g. a federated external IdP), the previous refresh token is invalidated as soon as Snowflake issues the new one. Lightdash currently **discards the rotated refresh token** returned by the Snowflake token endpoint, so the stored refresh token in `user_warehouse_credentials` (and project / org credentials) becomes stale immediately after the first refresh.

The next refresh attempt — typically a scheduler run a few minutes later — fails with:

```
SnowflakeTokenError: Error refreshing snowflake token: The provided grant or refresh token is invalid.
```

This is independent of the configured `OAUTH_REFRESH_TOKEN_VALIDITY` (TTL) — the token is invalidated on rotation, not on TTL expiry.

## Symptoms

- Google Sheets / Slack / email syncs (any scheduler-owned job) start failing with `Error refreshing snowflake token: The provided grant or refresh token is invalid.` shortly after the owner triggers a token refresh.
- Interactive queries can also fail, but are masked because the user can simply re-login.
- Re-authenticating the user fixes it — until the next rotation.

## Root cause

`UserService.generateSnowflakeAccessToken` (`packages/backend/src/services/UserService.ts`) explicitly discards the rotated refresh token returned by `passport-oauth2-refresh`:

```ts
refresh.requestNewAccessToken(
    'snowflake',
    refreshToken,
    (err, accessToken, _refreshToken, result) => { // <-- _refreshToken is the rotated token, ignored
        if (err || !accessToken) { reject(err); return; }
        resolve(accessToken);
    },
);
```

The underlying [`oauth` lib](https://github.com/ciaranj/node-oauth/blob/master/lib/oauth2.js#L212-L214) splits `access_token` and `refresh_token` out of the token response and surfaces both in the callback — so the rotated token is available, just thrown away.

`ProjectService.refreshCredentials` (`packages/backend/src/services/ProjectService/ProjectService.ts`) — the runtime refresh path — then sets only `token: accessToken` on the returned credentials and never updates the `refreshToken` field on the source row.

For comparison, the Databricks refresh path captures the rotated refresh token correctly (see `refreshDatabricksOAuthToken` callers), but doesn't persist it back to the credential row either — so this same class of bug exists for Databricks U2M/M2M, just less commonly observed.

## When does rotation happen?

For Snowflake's *Snowflake OAuth* integration:

- `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = TRUE` — refresh tokens are single-use, every refresh rotates and invalidates the previous token. The bug fires on every refresh.
- `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = FALSE` (default) — refresh tokens are reusable across multiple refreshes. The bug only fires when a new refresh token is issued, e.g. on user re-authorization.

For Snowflake's *External OAuth* integration (federated through Okta / AzureAD / etc.) the IdP controls rotation behavior; many IdPs rotate by default.

`DESCRIBE SECURITY INTEGRATION <name>` and inspect the listed parameters to determine which mode an integration is in.

## Reproduction

1. Configure a Snowflake OAuth security integration with `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = TRUE`.
2. Set up Lightdash with this integration as the Snowflake SSO provider.
3. Have a user authenticate via Snowflake OAuth and run any query — the access token is refreshed, Snowflake rotates and invalidates the previous refresh token.
4. Wait for the access token to expire (default 10 min) or trigger another refresh from a scheduled job.
5. The refresh fails with `invalid_grant` because the stored refresh token has been invalidated by Snowflake.

## Affected paths

All three storage locations for Snowflake SSO refresh tokens are affected:

- `user_warehouse_credentials` — when project has `requireUserCredentials = true`
- `warehouse_credentials` (project-level) — used by `ProjectService.buildAdapter` and the project-credentials branch of `getWarehouseCredentials`
- `organization_warehouse_credentials` — when project uses org-level credentials

## Fix

Capture the rotated refresh token and persist it back to whichever row sourced the original token, with an atomic compare-and-swap so concurrent refreshes don't overwrite each other.

PR: #22846

## Notes

- Existing GitHub issue [#16683](https://github.com/lightdash/lightdash/issues/16683) is about *notifying* users before tokens expire — this is a different bug (rotation discard), and fixing this should reduce — though not eliminate — the need for that notification.
- Databricks U2M/M2M has the same latent gap and is covered by the same PR.

ZeRego:
## Summary

- Capture the rotated OAuth refresh token returned by warehouse providers (currently discarded for Snowflake; for Databricks it was kept in memory but never written back) and persist it back to whichever row sourced it — `user_warehouse_credentials`, `warehouse_credentials` (project), or `organization_warehouse_credentials`.
- Single generic `rotateRefreshToken(uuid, oldToken, newToken)` method on each of the three credential models, with `forUpdate()` row lock + compare-and-swap so concurrent refreshes don't overwrite each other; cache invalidation on the project path.
- Covers Snowflake SSO and Databricks OAuth U2M / M2M with the same code path.
- Closes #22831.

## Why

Snowflake's OAuth integration can rotate refresh tokens — the previous token is invalidated as soon as a new one is issued. Two common scenarios:

- **Snowflake OAuth (custom client) with `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = TRUE`** — every refresh rotates, every refresh invalidates the prior token.
- **Snowflake OAuth (custom client) with `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = FALSE` (default)** — refresh tokens are reusable across refreshes, but a new refresh token is issued on user re-authorization, which can also invalidate the prior token depending on integration semantics.
- **External OAuth (federated through Okta / AzureAD)** — the IdP controls rotation; many IdPs rotate by default.

Lightdash was discarding the new token in all of these cases, so the next refresh hit `invalid_grant`. Symptom: scheduled jobs (Google Sheets / Slack / email syncs) fail shortly after the owner uses Lightdash interactively. Detailed root-cause analysis in #22831.

Databricks OAuth U2M/M2M has the same rotation behavior (depending on workspace config) and the same latent persistence gap — the `refreshDatabricksOAuthToken` callers correctly captured the rotated `refreshToken` in memory but never wrote it back to the credentials store. Now folded into the same fix.

## What changed

| File | Change |
|---|---|
| `services/UserService.ts` | `generateSnowflakeAccessToken` now returns `{ accessToken, refreshToken }` instead of only the access token. |
| `services/ProjectService/ProjectService.ts` | `refreshCredentials` Snowflake SSO branch propagates the rotated `refreshToken` on the returned credentials (mirrors the Databricks pattern). New helpers `refreshCredentialsAndPersistRotation` + `persistRefreshTokenRotation` write the rotated token back to the source row via the appropriate model — generic over warehouse type, picks up Snowflake SSO and Databricks U2M/M2M structurally. `getWarehouseCredentials` (org / user / project / embed branches) and `buildAdapter` use the new helper. Persistence failures log structured context (`sourceKind`, `sourceUuid`) without failing the in-flight query. |
| `models/UserWarehouseCredentialsModel.ts` | New generic `rotateRefreshToken(uuid, oldToken, newToken)` — `forUpdate()` row lock + compare-and-swap on the stored `refreshToken` field. |
| `models/OrganizationWarehouseCredentialsModel.ts` | Same generic method on `warehouse_connection`. |
| `models/ProjectModel/ProjectModel.ts` | Same generic method on `warehouse_credentials.encrypted_credentials`; also invalidates `warehouseCredentialsCache` after a successful swap so the next reader picks up the rotated token. |
| `services/ProjectService/ProjectService.test.ts` | Existing mocks updated for the new return shape; four new tests cover (a) Snowflake rotation persisted via the user model, (b) Snowflake no-op when the same token is returned, (c) Databricks U2M rotation persisted via the user model, (d) Databricks U2M no-op when the same token is returned. |

The CAS guard on the stored `refreshToken` value is what protects against credential type changes (the old token won't match) and concurrent rotations (whoever wins the lock + CAS check writes). No warehouse-specific branching in the model methods — they're structural.

## Test plan

- [x] `pnpm -F backend lint` on touched files
- [x] `pnpm -F backend test -- ProjectService.test` — 49/49 (includes 4 new tests for rotated-token persistence + no-op)
- [ ] Manual: configure a Snowflake OAuth integration with `OAUTH_SINGLE_USE_REFRESH_TOKENS_REQUIRED = TRUE`, log in, run a query, wait for access-token TTL, run another query — should succeed instead of failing with `invalid_grant`.
- [ ] Manual: same scenario for Databricks OAuth U2M with rotation enabled at the workspace.

## Out of scope / follow-ups

- Initial Databricks OAUTH_M2M token issuance from `oauthClientId + oauthClientSecret` (when no refresh token has been saved yet) is still re-exchanged on every request rather than persisted. The new helper only persists rotations of an existing refresh token. Worth a separate optimization PR.
- `openid_identities.refresh_token` is not rotated by this change; it's only used by `UserService.getAccessToken('snowflake')` for the EE `/snowflake/sso/is-authenticated` check, which is not on the query path.

### DIFF
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -996,6 +996,21 @@ export class ProjectService extends BaseService {
             : undefined;
     }
 
+    private static getRotationSourceUuid(
+        source: RefreshTokenRotationSource,
+    ): string {
+        switch (source.kind) {
+            case 'project':
+                return source.projectUuid;
+            case 'organization':
+                return source.organizationWarehouseCredentialsUuid;
+            case 'user':
+                return source.userWarehouseCredentialsUuid;
+            default:
+                return assertUnreachable(source, 'Unknown source kind');
+        }
+    }
+
     private async persistRefreshTokenRotation({
         source,
         oldRefreshToken,
@@ -1036,11 +1051,11 @@ export class ProjectService extends BaseService {
             }
         } catch (error) {
             // Don't fail the in-flight query: the freshly minted access token is still usable.
-            this.logger.error(
-                `Failed to persist rotated OAuth refresh token for ${
-                    source.kind
-                }: ${getErrorMessage(error)}`,
-            );
+            this.logger.error('Failed to persist rotated OAuth refresh token', {
+                sourceKind: source.kind,
+                sourceUuid: ProjectService.getRotationSourceUuid(source),
+                error: getErrorMessage(error),
+            });
         }
     }
 
--- a/packages/backend/src/services/ProjectService/ProjectService.ts
+++ b/packages/backend/src/services/ProjectService/ProjectService.ts
@@ -270,6 +270,14 @@ import { UserService } from '../UserService';
 import { getFieldValuesMetricQuery } from './fieldValuesQueryBuilder';
 import { getAvailableParameterDefinitions } from './parameters';
 
+type RefreshTokenRotationSource =
+    | { kind: 'project'; projectUuid: string }
+    | {
+          kind: 'organization';
+          organizationWarehouseCredentialsUuid: string;
+      }
+    | { kind: 'user'; userWarehouseCredentialsUuid: string };
+
 export type ProjectServiceArguments = {
     lightdashConfig: LightdashConfig;
     analytics: LightdashAnalytics;
@@ -768,7 +776,7 @@ export class ProjectService extends BaseService {
                 );
                 // If we try to generate access token from token instead of refreshToken
                 // it will throw an error: The request was invalid.
-                const accessToken =
+                const { accessToken, refreshToken: newRefreshToken } =
                     await UserService.generateSnowflakeAccessToken(
                         refreshToken,
                     );
@@ -776,6 +784,7 @@ export class ProjectService extends BaseService {
                     ...args,
                     authenticationType: SnowflakeAuthenticationType.SSO,
                     token: accessToken,
+                    refreshToken: newRefreshToken,
                 };
             } catch (e: unknown) {
                 if (e instanceof LightdashError) {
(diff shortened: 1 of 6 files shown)


ZeRego:
* **#22855** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22846** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22846?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/services/ProjectService/ProjectService.ts
```diff
@@ -948,6 +957,93 @@ export class ProjectService extends BaseService {
...
+        const oldRefreshToken = ProjectService.getCredentialsRefreshToken(args);
+
+        const refreshed = await this.refreshCredentials(args, userUuid);
+
+        const newRefreshToken =
+            ProjectService.getCredentialsRefreshToken(refreshed);
+
+        if (
+            oldRefreshToken &&
+            newRefreshToken &&
```
rephus, line 976: yes, do not store new refreshtoken if they are returned undefined.

PR #22850, opened 2026-05-08
fix(data-apps): use fully qualified e2b template name

Spissable:
### Description:
Self-hosters are currently either required to upload their own template or set the 
`E2B_TEMPLATE_NAME` to `lightdash/lightdash-data-app`.
To make it easier, I'm changing the default value to be the fully qualified name of our public template. Self-hosters can still opt to use their own, but the default value will work out-of-the-box with our template.


Spissable:
Closing, because the default value will still not work universally because we have `lightdash-eu/lightdash-data-app` for EU.

Spissable:
The cloud envs now properly define the template based on US/EU region so it's safe to now add the US default value here.

PR #22853, opened 2026-05-08
feat: hide dashbaord tabs

magnew:
Closes: #20563

### Description:

Makes it possible to hide dashbaord tabs. When not in edit mode, hidden tabs will be hidden.

PR #22870, opened 2026-05-08
fix: apply truncation in case statement for rolling period buckets

### LINKED ISSUE #20812: Metric canvas rolling period CASE statement doesn't truncate timestamps, causing boundary-date records to be excluded
## Bug Description

In the Metric Canvas "total" comparison view with rolling period date filtering, the SQL CASE statement that buckets data into current/previous periods uses raw timestamp comparisons, while the WHERE clause uses `TIMESTAMP_TRUNC` to truncate to day boundaries. This mismatch causes records on boundary dates to fall through to `NULL` in the CASE statement and be excluded from results.

## Example

Given a metric like **count of user_ids** with a rolling 7-day comparison, the generated SQL produces:

**WHERE clause** (truncates to day):
```sql
WHERE TIMESTAMP_TRUNC(table.created_at, DAY) >= '2025-01-08'
  AND TIMESTAMP_TRUNC(table.created_at, DAY) <= '2025-01-14'
```

**CASE statement** (raw timestamp, no truncation):
```sql
CASE
  WHEN table.created_at >= '2025-01-08' AND table.created_at <= '2025-01-14' THEN 0
  WHEN table.created_at >= '2025-01-01' AND table.created_at <= '2025-01-07' THEN 1
  ELSE NULL
END
```

A record with `created_at = '2025-01-14 15:30:00'` passes the WHERE clause (because `TIMESTAMP_TRUNC('2025-01-14 15:30:00', DAY) = '2025-01-14'`), but fails the CASE condition (because `'2025-01-14 15:30:00' > '2025-01-14'`), resulting in `NULL` and the record being excluded from both current and previous period counts.

## Root Cause

In `packages/common/src/utils/metricsExplorer.ts`, the `buildRollingPeriodCustomDimension` function generates the CASE SQL using raw `timeDimensionFieldRef` without applying `TIMESTAMP_TRUNC`:

```typescript
sql: `CASE
  WHEN ${timeDimensionFieldRef} >= ${cast(current.start)} AND ${timeDimensionFieldRef} <= ${cast(current.end)} THEN 0
  ...
END`
```

Meanwhile, in `packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts`, the filter uses `interval: TimeFrames.DAY` which triggers `TIMESTAMP_TRUNC` in the filter compiler:

```typescript
const filterTimeDimension = {
    table: metricTimeDimension.table,
    field: metricTimeDimension.field,
    interval: TimeFrames.DAY,  // triggers TIMESTAMP_TRUNC in WHERE clause
};
```

## Steps to Reproduce

1. Open Metric Canvas
2. Select a metric that uses a timestamp field as its time dimension (e.g. count of user_ids by created_at)
3. Switch to "total" comparison view
4. Use rolling period date filtering (e.g. last 7 days)
5. Check the generated SQL - CASE boundaries won't match WHERE boundaries
6. Records with timestamps after midnight on boundary dates will be silently excluded

## Expected Behavior

The CASE statement should apply the same `TIMESTAMP_TRUNC` to the time dimension field as the WHERE clause, so that all records passing the filter are correctly bucketed.

## Suggested Fix

Apply `TIMESTAMP_TRUNC` (or equivalent per warehouse) to `timeDimensionFieldRef` in the CASE conditions within `buildRollingPeriodCustomDimension`, matching the truncation used in the WHERE clause filter.

## Affected Files

- `packages/common/src/utils/metricsExplorer.ts` (lines ~691-716) - `buildRollingPeriodCustomDimension`
- `packages/backend/src/services/MetricsExplorerService/MetricsExplorerService.ts` (lines ~264-322) - service calling the function

magnew:
Closes: #20812 

### Description:

Fixes rolling comparisons in metric canvas by applying the same date truncation to the case statement as the where clause. 

See linked ticket for a detailed explanation of the issue.

### DIFF
--- a/packages/backend/src/database/entities/dashboards.ts
+++ b/packages/backend/src/database/entities/dashboards.ts
@@ -154,6 +154,7 @@ export type DbDashboardTabs = {
     dashboard_id: number;
     dashboard_version_id: number;
     order: number;
+    hidden: boolean;
     created_at: Date;
     updated_at: Date;
 };
--- a/packages/backend/src/database/migrations/20260507143028_add_hidden_to_dashboard_tabs.ts
+++ b/packages/backend/src/database/migrations/20260507143028_add_hidden_to_dashboard_tabs.ts
@@ -1,0 +1,20 @@
+import { Knex } from 'knex';
+
+const tableName = 'dashboard_tabs';
+const columnName = 'hidden';
+
+export async function up(knex: Knex): Promise<void> {
+    if (!(await knex.schema.hasColumn(tableName, columnName))) {
+        await knex.schema.alterTable(tableName, (table) => {
+            table.boolean(columnName).notNullable().defaultTo(false);
+        });
+    }
+}
+
+export async function down(knex: Knex): Promise<void> {
+    if (await knex.schema.hasColumn(tableName, columnName)) {
+        await knex.schema.alterTable(tableName, (table) => {
+            table.dropColumn(columnName);
+        });
+    }
+}
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -2341,6 +2341,7 @@ const models: TsoaRoute.Models = {
         type: {
             dataType: 'nestedObjectLiteral',
             nestedProperties: {
+                hidden: { dataType: 'boolean', required: true },
                 order: { dataType: 'double', required: true },
                 name: { dataType: 'string', required: true },
                 uuid: { dataType: 'string', required: true },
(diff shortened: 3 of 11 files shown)


magnew:
* **#22870** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22870?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22870?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22853** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22853?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #22880, opened 2026-05-11
fix: Enable handleExceptions and handleRejections on Winston Console transport

### LINKED ISSUE #22824: [Logging] Enable handleExceptions / handleRejections on Winston Console transport
## Use case

When running Lightdash with a log aggregator (e.g. Datadog, Loki), unhandled exceptions and promise rejections produce raw multi-line stack traces on stderr. Because the Winston Console transport does not have `handleExceptions: true` / `handleRejections: true`, the aggregator splits each line of the stack trace into a separate error-level log event (10-20 events per exception instead of one structured entry).

## Proposed solution

Add `handleExceptions: true` and `handleRejections: true` to the Console transport in `packages/backend/src/logging/winston.ts`.

## Workaround

Self-hosted operators can add a multi-line aggregation rule in their log pipeline to group stack trace lines into a single event.

karthikkarnam:
Closes: [[Logging] Enable handleExceptions / handleRejections on Winston Console transport](https://github.com/lightdash/lightdash/issues/22824)

### Description:

The Winston Console transport was not configured with `handleExceptions: true` or `handleRejections: true`, so uncaught exceptions and unhandled promise rejections bypassed Winston entirely and wrote raw plain-text stack traces to stderr.

This means one unhandled exception produces 10-20 separate log events — each a single stack frame with no standalone diagnostic value — rather than a single structured JSON entry. For deployments using a log aggregator (e.g. Datadog), this creates significant noise and makes it difficult to identify real issues.

This PR adds `handleExceptions: true` and `handleRejections: true` to the Console transport so all errors are captured by Winston, formatted consistently with the rest of the log output, and emitted as a single structured entry.

### DIFF
--- a/packages/backend/src/logging/winston.ts
+++ b/packages/backend/src/logging/winston.ts
@@ -124,6 +124,8 @@ if (lightdashConfig.logging.outputs.includes('console')) {
             level:
                 lightdashConfig.logging.consoleLevel ||
                 lightdashConfig.logging.level,
+            handleExceptions: true,
+            handleRejections: true,
         }),
     );
 }


ZeRego:
Closing in favour of https://github.com/lightdash/lightdash/commit/12b0f8520782f43d921e1bfe734b6fa361565ed1

PR #22881, opened 2026-05-11
fix: resolve short-form metric refs against outer metric's table to prevent same-named joined metric collision (PROD-7503)

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7503/nested-aggregate-cte-resolves-metric-reference-to-wrong-table-when

### Description:

Fixes a bug where a short-form metric reference (e.g. `${met_active_customers}`) inside a mixed nested aggregate metric was incorrectly resolving to a same-named aggregate metric on a joined table instead of the raw column helper on the outer metric's own table.

The root cause was that the previous substitution logic matched `${name}` purely by metric name, without considering which table the reference belonged to. When a joined table had an aggregate metric with the same name as a base table's raw helper, the joined table's pre-aggregated CTE column would be substituted into the `MAX_BY` expression instead of the base table's raw column — producing incorrect SQL.

The fix replaces the name-only string substitution with a `getParsedReference`-based approach that resolves each `${ref}` token using the outer metric's own table as the default table context. This ensures short-form refs like `${met_active_customers}` correctly map to `base_tbl.met_active_customers`, while qualified refs like `${joined_tbl.met_active_customers}` still correctly resolve to the joined table's CTE column.

A new test fixture (`EXPLORE_NESTED_AGG_NAME_COLLISION` / `METRIC_QUERY_NESTED_AGG_NAME_COLLISION`) and corresponding test case are added to cover this scenario and prevent regressions.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
@@ -3176,33 +3176,29 @@ export class MetricQueryBuilder {
 
             if (hasRawDep) {
                 // Mixed metric: replace aggregate dep refs with CTE 1 column
-                // refs manually, then let compileMetricSql resolve raw refs
-                // and ${TABLE} refs against the base table (which is in scope
-                // in CTE 3).
-                let mixedSql = preSql;
-                for (const [depId, depMetric] of aggregateInnerDeps) {
-                    const escapedName = depMetric.name.replace(
-                        /[.*+?^${}()|[\]\\]/g,
-                        '\\$&',
-                    );
-                    const escapedTable = depMetric.table.replace(
-                        /[.*+?^${}()|[\]\\]/g,
-                        '\\$&',
-                    );
-                    const cteRef = `${naCteName}.${fieldQuoteChar}${depId}${fieldQuoteChar}`;
-                    mixedSql = mixedSql
-                        .replace(
-                            new RegExp(`\\$\\{${escapedName}\\}`, 'g'),
-                            cteRef,
-                        )
-                        .replace(
-                            new RegExp(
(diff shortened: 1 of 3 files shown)


rephus:
* **#22881** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22881?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22881?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
```diff
@@ -3176,33 +3176,29 @@ export class MetricQueryBuilder {
...
-                                `\\$\\{${escapedTable}\\.${escapedName}\\}`,
-                                'g',
-                            ),
-                            cteRef,
+                // refs, then let compileMetricSql resolve raw refs and
+                // ${TABLE} refs against the base table (which is in scope in
+                // CTE 3). Resolve each ${ref} via getParsedReference so the
+                // short form ${name} maps to the outer metric's own table —
+                // not to a same-named metric on a joined table (PROD-7503).
+                const mixedSql = preSql.replace(
```
rephus, line 3184: The PR #21509 tests cover only same-table raw + aggregate mixing (single-table EXPLORE_WITH_NESTED_AGG). They never exercised a cross-table  aggregate dep in the mixed CTE — which is exactly the gap that let PROD-7503 slip through. My new test fills that gap.

rephus:
CleanShot 2026-05-11 at 10 39 46

tests seem unrelated with our code change.

PR #22910, opened 2026-05-11
feat(spk-447): gate custom roles via DB feature flag

sdolidze:
## Summary

Adds a per-organization DB-backed gate for custom roles, layered on top of the existing `CUSTOM_ROLES_ENABLED` env var. Either signal enables the feature.

Closes [SPK-447](https://linear.app/lightdash/issue/SPK-447).

## Why

Service accounts already work with both env var and DB-backed flag overrides; custom roles only honored the env var, so cloud orgs couldn't be opted in without a backend redeploy. This brings custom roles into line.

## Approach

- New EE flag: `CommercialFeatureFlags.CustomRoles = 'custom-roles'`.
- New handler `CommercialFeatureFlagModel.getCustomRolesFlag`: returns `true` if env is set, else falls back to `feature_flags` + `feature_flag_overrides` (user → org → default). Mirrors `getEnableDataAppsEnabled` / `getAiCopilotFlag`.
- `UserModel.generateUserAbilityBuilder` resolves the flag per-session via `featureFlagModel.get(...)` and passes the resolved boolean into `getUserAbilityBuilder`. Service-account branch is **untouched** — role bindings still apply regardless of the flag.
- Frontend (`Settings.tsx`, `TableCalculationModal.tsx`): OR `health.isCustomRolesEnabled` (env-only) with `useServerFeatureFlag(CommercialFeatureFlags.CustomRoles)` — same pattern as service-accounts.

## Behavior matrix

| `CUSTOM_ROLES_ENABLED` env | DB flag for `'custom-roles'` | Frontend (menu visible?) | Backend (scopes applied?) |
|---|---|---|---|
| false | false | hidden | no |
| false | true | shown | yes |
| true | false | shown | yes |
| true | true | shown | yes |

Also gated by `isEnterprise` (license key) on the backend — non-EE installs always resolve to no regardless of env/DB.

## Test plan

- [x] \`pnpm -F backend test:dev:nowatch\` passes
- [x] \`pnpm -F common typecheck\` / \`pnpm -F backend typecheck\` / \`pnpm -F frontend typecheck\` clean
- [x] Manual: with \`CUSTOM_ROLES_ENABLED\` unset, insert a \`feature_flag_overrides\` row for a test org → Settings menu shows Custom Roles and a custom role works end-to-end
- [x] Manual: with neither env nor DB row, menu is hidden and API rejects custom-role operations
- [x] Manual: with \`CUSTOM_ROLES_ENABLED=true\`, behavior matches today's main
- [x] Verify service-account flows are unaffected

## Notes

\`/api/v1/health.isCustomRolesEnabled\` remains env-only by design; per-org resolution is exposed via \`/api/v2/feature-flag/custom-roles\`, which the frontend ORs with the health value. SPK-447's third acceptance criterion ("health reflects the resolved value") was intentionally dropped to avoid an extra DB round-trip on every health probe.

### DIFF
--- a/packages/backend/src/ee/models/CommercialFeatureFlagModel.ts
+++ b/packages/backend/src/ee/models/CommercialFeatureFlagModel.ts
@@ -1,4 +1,4 @@
-import { CommercialFeatureFlags } from '@lightdash/common';
+import { CommercialFeatureFlags, FeatureFlag } from '@lightdash/common';
 import { Knex } from 'knex';
 import { LightdashConfig } from '../../config/parseConfig';
 import {
@@ -14,6 +14,8 @@ export class CommercialFeatureFlagModel extends FeatureFlagModel {
             // Add new commercial handlers
             [CommercialFeatureFlags.AiCopilot]:
                 this.getAiCopilotFlag.bind(this),
+            [CommercialFeatureFlags.CustomRoles]:
+                this.getCustomRolesFlag.bind(this),
         };
     }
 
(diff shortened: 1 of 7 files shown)


sdolidze:
* **#22910** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22911, opened 2026-05-11
feat(timezone): user-level timezone preference (PROD-7528)

ZeRego:
## Summary

Adds a per-user timezone preference (Looker `timezone: user_timezone` compatibility). Charts that haven't pinned a timezone now render in each viewer's preferred zone — between the chart and project layers in the resolution chain.

```
chart.timezone → user.timezone → project.queryTimezone → 'UTC'
```

Closes: [PROD-7528](https://linear.app/lightdash/issue/PROD-7528/support-user-level-timezone)

## Implementation

- Migration adds nullable \`timezone\` column to \`users\`; threaded through \`LightdashUser\` / \`UpdateUserArgs\` and validated server-side via \`isValidTimezone\` on update.
- \`resolveQueryTimezone\` takes an optional \`userTimezone\` arg; new \`getAccountUserTimezone(account)\` helper returns \`null\` for anonymous (embed/JWT) accounts so they fall through to the project default.
- The resolved chart > user > project timezone is stamped onto \`metricQuery.timezone\` before persisting to \`query_history\`. Worker paths and retrieval paths read it back via a lean \`getDisplayTimezoneForQueryHistory\` helper without re-resolving.
- **Pre-aggregate materializations skip user-level resolution** (\`materializationRole !== undefined → userTimezone: null\`) so shared pre-aggregate tables aren't compiled against the triggering user's zone.
- Profile UI: gated behind \`EnableTimezoneSupport\`, adds a clearable \`TimeZonePicker\` to \`ProfilePanel\`. Empty → project default.

### Demo

CleanShot 2026-05-12 at 11 36 44@2x

## Test plan

- [x] Unit tests for new \`resolveQueryTimezone\` precedence cases pass
- [x] Migration applies and rolls back cleanly
- [x] Typecheck + lint pass across common / backend / frontend
- [x] Existing test suites pass (UserService, AsyncQueryService, EmbedService, ScimService, account, UserModel)
- [x] End-to-end precedence verified via 6 live queries against local stack — persisted \`metric_query.timezone\` in \`query_history\` matches expected for every combination:
  | chart | user | project | resolved | ✓ |
  |---|---|---|---|---|
  | — | Tokyo | — | Asia/Tokyo | ✅ |
  | NY | Tokyo | — | America/New_York | ✅ |
  | — | — | — | UTC | ✅ |
  | — | — | London | Europe/London | ✅ |
  | — | Sydney | London | Australia/Sydney | ✅ |
  | Berlin | Sydney | London | Europe/Berlin | ✅ |
- [x] Profile UI flow tested in browser (save / reload / clear)
- [ ] CI green


packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
ZeRego, line 1: @almeidabbm worth reviewing

packages/common/src/utils/resolveQueryTimezone.ts
ZeRego, line 1: @almeidabbm worth reviewing

packages/backend/src/services/ProjectService/ProjectService.ts
ZeRego, line 1: @almeidabbm worth reviewing

packages/frontend/src/components/UserSettings/ProfilePanel/index.tsx
```diff
@@ -37,6 +51,11 @@ const ProfilePanel: FC = () => {
     } = useApp();
     const { showToastSuccess, showToastApiError } = useToaster();
 
+    const { data: userTimezonesFlag } = useServerFeatureFlag(
+        FeatureFlags.EnableUserTimezones,
```
almeidabbm, line 55: Should this be gated behind this flag or the `EnableTimezoneSupport`? 🤔 
If it's gated behind this at the UI level then it means both `EnableTimezoneSupport` and `EnableUserTimezones` will need to be enabled for this to work

almeidabbm (commented):
LGTM 🚀 the pattern of saving the timezone in the metric query sounds good and should stand caching

Also confirmed that the hot paths for `DATE_TRUNC`/`EXTRACTS` are behind the feature flag and that is also true, no regression!

packages/frontend/src/components/UserSettings/ProfilePanel/index.tsx
ZeRego: both needed 👍

PR #22921, opened 2026-05-12
fix(dashboards): prevent FK violation and orphan-cleanup crash on save (PROD-5905)

sdolidze:
## Summary

Two related backend fixes for dashboard save failures in production. Linear: [PROD-5905](https://linear.app/lightdash/issue/PROD-5905/dashboard-save-fails-with-fk-violation-when-editing-multi-tab).

### Bug 1 — FK violation on dashboard save

`DashboardModel.addVersion` inserts tiles with `tab_uuid` without verifying the value matches any tab being written in the same version. When a caller (frontend, CLI, promotion, concurrent edit) sends `tabs: []` (or any tab list that doesn't include a tile's `tabUuid`), Postgres rejects the tile insert with:

```
violates foreign key constraint "dashboard_tiles_tab_uuid_dashboard_version_id_foreign"
Key (tab_uuid, dashboard_version_id)=(…, N) is not present in table "dashboard_tabs".
```

The user sees a generic "Something went wrong" toast; the actual FK error only appears in **stderr**, not structured logs, which is why this went undetected for ~28 days across both multi-tenant instances and 5 self-hosted ones (~37 failed saves).

**Fix:** intersect each tile's `tabUuid` with the set of tab UUIDs being inserted in this version; stale references become `NULL`. The tile lands without a tab — the same state it would be in on a no-tab dashboard. No behavior change for valid payloads.

### Bug 2 — Orphan cleanup crash hides successful saves

`deleteOrphanedChartsInDashboards` runs *after* `addVersion` succeeds and calls `permanentDelete` on each orphaned chart. If a chart is already gone (e.g. race with a retried save), `NotFoundError` bubbles up and fails the save response — even though the version was already created. Surfaced in prod as 121 occurrences across 5 instances since March 11.

**Fix:** swallow `NotFoundError` per chart and log it. Other errors still propagate.

## Test plan

- [x] New unit test \`should null out tile.tabUuid that does not match any tab in this version (PROD-5905)\` in \`DashboardModel.test.ts\` — fails without Fix 1, passes with it.
- [x] \`pnpm -F backend lint\` clean.
- [x] \`pnpm -F backend typecheck\` clean.
- [x] Full \`DashboardModel\` + \`DashboardService\` jest suites pass (36 tests).
- [x] End-to-end repro: crafted bad payload (\`tabs: []\` + tiles carrying \`tabUuid\`) sent to \`PATCH /api/v2/projects/.../dashboards/...\` returns **500 on main**, **200 with \`tabUuid: null\` on this branch**.

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.test.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.test.ts
@@ -744,4 +744,58 @@ describe('DashboardModel', () => {
             ),
         ).rejects.toThrowError(NotFoundError);
     });
+
+    test('should null out tile.tabUuid that does not match any tab in this version (PROD-5905)', async () => {
+        // Simulates the bad payload that triggered the FK violation in
+        // production: tabs[] empty (or missing the tile's tab) while a tile
+        // still carries a stale tabUuid.
+        const staleTabUuid = '00000000-0000-0000-0000-000000000000';
+        const versionWithStaleTabUuid: typeof addDashboardVersion = {
+            ...addDashboardVersion,
+            tabs: [],
+            tiles: [
+                {
+                    ...addDashboardVersion.tiles[0],
+                    tabUuid: staleTabUuid,
+                },
+            ],
+        };
+
+        tracker.on.select(DashboardsTableName).responseOnce([dashboardEntry]);
+        tracker.on.select(SpaceTableName).responseOnce([spaceEntry]);
+        tracker.on.insert(DashboardsTableName).responseOnce([dashboardEntry]);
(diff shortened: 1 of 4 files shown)


sdolidze:
* **#22921** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22921?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22921?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22884, opened 2026-05-11
fix(echarts): snap category date axis in project timezone

almeidabbm:
Bar charts using WEEK / MONTH / QUARTER / YEAR category axes render no bars on non-UTC project timezones.

After #22207, truncated intervals come back as UTC instants aligned to project-TZ midnights (e.g. Pago_Pago Nov 1 → `2024-11-01T11:00:00Z`). `getCategoryDateAxisConfig` still snapped axis boundaries to UTC midnights → string mismatch → ECharts category mapping drops every row.

Fix: snap WEEK / MONTH / QUARTER / YEAR boundaries in `resolvedTimezone`, then format back to UTC ISO so the strings match. Falls back to UTC when the flag is off — pre-#22207 behavior unchanged.

**Before**
CleanShot 2026-05-11 at 09 27 03

**After**
after_fix

Closes [GLITCH-428](https://linear.app/lightdash/issue/GLITCH-428/category-axis-time-dimensions-render-no-bars-on-non-utc-project)

PR #22922, opened 2026-05-12
fix(echarts): pad continuous category axis for all bar charts

almeidabbm:
Closes GLITCH-435.

## Summary

Cartesian bar charts on a category date axis can drop bars (or render nothing at all) whenever the snap-generated category strings don't byte-for-byte equal the warehouse row strings. ECharts matches dataset rows to categories by strict string equality, so any mismatch is silent — the bar just disappears.

While debugging GLITCH-435 we found that the same change unblocks three distinct bugs, two of which pre-date #22884:

1. **Non-UTC project tz + DST (the GLITCH-435 trigger).** `getCategoryDateAxisConfig` chains `.add()` on a `dayjs.tz` object, but the dayjs timezone plugin doesn't recompute the zone offset across DST inside a chain. The continuous range stays at the start-of-range offset (e.g. EST), while the warehouse correctly emits DST-aware UTC instants for some months (e.g. Nov 1 in EDT). Off by one hour → bar dropped.
2. **UTC project tz, sub-day string drift (pre-existing).** Even with no DST in play, the row's x-field and the snap's category can disagree on sub-day formatting (ms precision, `Z` vs `+00:00`, etc.), dropping bars whenever rowLimit is off.
3. **Flipped axes + row limit, latent since #21480.** The padding call site selected `yField[0]` (the metric) when `flipAxes` was true, so each row's metric column got overwritten with a date string. The chart rendered nothing. The bug was masked behind the rowLimit gate; removing the gate exposed it, then the selector was corrected.

## Fix

`padDatasetForContinuousAxis` already normalizes both sides to `YYYY-MM-DD` and rewrites rows to the canonical category string. The call site has two changes:

- Drop the `rowLimit` half of the guard so padding runs whenever there is a continuous range. Per #21480's commit log the gate was stylistic alignment with surrounding row-limit code, not correctness-bearing.
- Always use `layout.xField` as the date field. The flip only swaps which visual axis renders the date; the dimension's field name doesn't change.

For UTC charts where rows and categories already match exactly, the rewrite is byte-for-byte identical to the input — zero behavioral change.

Also seeds a `monthly_dst` series on `timezone_test` (Jan 2024 → Jan 2025, one row mid-month at NYC noon) so all three failure shapes are locally reproducible.

### Before
CleanShot 2026-05-12 at 11 23 45

**Limit rows**
CleanShot 2026-05-12 at 11 30 38

**Flipped axes**
CleanShot 2026-05-12 at 11 27 31

**Flipped axes, limit rows**
CleanShot 2026-05-12 at 11 30 15

### After
CleanShot 2026-05-12 at 11 25 43

**Limit rows**
CleanShot 2026-05-12 at 11 31 00

**Flipped axes**
CleanShot 2026-05-12 at 11 27 04

**Flipped axes, limit rows**
CleanShot 2026-05-12 at 11 29 18


almeidabbm:
* **#22952** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#22922** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22922?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22922?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22912, opened 2026-05-11
feat(scheduler): show last-run status badge on dashboard/chart scheduler list

ZeRego:
Closes: SPK-448
Refs: SPK-422

### Description

Editors can now see at a glance whether each scheduled delivery's last run succeeded, failed, is running, or has never run, directly from the "Scheduled deliveries" modal on dashboard and chart pages — replacing the previously planned deep-link-to-settings approach with an inline badge.

**Backend** — Adds an `includeLatestRun` query param to the v2 resource-scoped scheduler endpoints (`/api/v2/dashboards/:uuid/schedulers`, `/api/v2/saved/:uuid/schedulers`). Permission gates are unchanged (`create ScheduledDeliveries` + space access via `checkCreateScheduledDeliveryAccess`). The latest-run enrichment is centralised in a new `SchedulerModel.attachLatestRunToSchedulers` helper, and a `latestOnly` flag on `getRunsForSchedulers` collapses results to one row per scheduler via a per-scheduler `ROW_NUMBER()` filter instead of fetching all runs and discarding them in JS.

**Frontend** — `SchedulersListItem` renders a status badge (Completed / Failed / Partial failure / Running / Scheduled) with a tooltip showing the relative timestamp; falls back to a muted "Not run yet" pill when there is no run. `getRunStatusConfig` is lifted from `SchedulersTable` into the shared `SchedulersViewUtils` so the admin view and the modal use the same source of truth.

### Screenshot

spk-448-all-states

| State | Visual |
|---|---|
| Brand new delivery | *Not run yet* (italic gray) |
| Daily revenue alert | ❗ **Failed** (red) |
| Partial-failure delivery | ⚠️ **Partial failure** (yellow) |
| Running right now | 🏃 **Running** (blue) |
| Weekly sales summary | ✓ **Completed** (green) |

Tooltip on hover: *Last run 2 hours ago (2026/05/11 15:53)*

PR #22947, opened 2026-05-12
fix(e2e): enable USE_SQL_PIVOT_RESULTS in preview env

almeidabbm:
### Description

`pivotTables.cy.ts` aliases the metric-query POST only when its body carries `pivotConfiguration`. `buildQueryArgs.ts` only sets that field when the `use-sql-pivot-results` flag is enabled.

Cloud helm sets `USE_SQL_PIVOT_RESULTS=true` (per the comment in `parseConfig.ts:1574-1578`), but the okteto preview env deploys via `docker/docker-compose.preview.yml` and otherwise falls back to PostHog targeting, which doesn't reach the ephemeral preview org. With the flag effectively `false` in preview, the 4 pivot chart specs in shard 3 deterministically time out at 60s × 3 retries × 4 specs (~12 min) on every PR — observed on #22912, #22920, #22922.

This mirrors the existing `EMBEDDING_ENABLED: true` block, which exists for the same reason.

test-all

### DIFF
--- a/docker/docker-compose.preview.yml
+++ b/docker/docker-compose.preview.yml
@@ -81,6 +81,13 @@ services:
             # enabledFeatureFlags allowlist (step 1a).
             EMBEDDING_ENABLED: true
 
+            # Force-enable the SQL-pivot path in preview envs. Cloud helm sets
+            # this true; preview deploys via docker-compose and otherwise
+            # falls back to PostHog targeting, which doesn't reach the
+            # ephemeral test org — pivotTables.cy.ts then times out waiting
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#22947** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22947?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22947?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22952, opened 2026-05-12
refactor(echarts): consume padded dataset directly for stack decorations

almeidabbm:
Follow-up to #22922 (DST padding fix on continuous category axes).

That PR turned padding on for every continuous-range bar chart but only normalized `dataset.source`. Stack decorations (rounded corners on top segments, stack-total labels in 100% mode) were still built from the raw, drifted-cat rows — so on a non-UTC project timezone with DST, stacked bars kept dropping even after #22922.

This PR routes those decorations through the same padded dataset. Two follow-on cleanups bundled in:

- Decorations consume the canonicalized flat dataset (`paddedSortedResults`) directly, so cat strings match `xAxis.data` exactly.
- Padding runs once, upstream of the 100%-stack transform, instead of producing two parallel padded datasets. Composition equivalence is pinned by new tests.

### Tests
- `transformToPercentageStacking` was untested — added basic / gap-tolerance / no-op coverage.
- New `pad ∘ transform` test asserts the two composition orders agree on non-gap rows.
- All tests pass against both the pre- and post-collapse implementations.

### DIFF
--- a/packages/common/src/visualizations/helpers/styles/barChartStyles.test.ts
+++ b/packages/common/src/visualizations/helpers/styles/barChartStyles.test.ts
@@ -1,11 +1,10 @@
-import { type ResultRow } from '../../../types/results';
 import { CartesianSeriesType } from '../../../types/savedCharts';
 import { type EChartsSeries } from '../../types';
 import { applyRoundedCornersToStackData } from './barChartStyles';
 
-const makeRow = (date: string, value: number): ResultRow => ({
-    date_month: { value: { raw: date, formatted: date } },
-    count: { value: { raw: value, formatted: String(value) } },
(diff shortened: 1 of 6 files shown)


almeidabbm:
* **#22952** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#22922** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #22967, opened 2026-05-12
feat: migrate results cache flag to per-org feature flag with env fallback

rephus:
### Description

Converts `RESULTS_CACHE_ENABLED` from a static environment-variable check into a feature flag (`ResultsCacheEnabled`) backed by `FeatureFlagModel`. The DB value (user override → org override → flag default) takes precedence; if no DB row exists, falls back to the existing `RESULTS_CACHE_ENABLED` env var. This unblocks shared-instance customers (eu1/app) who currently can't opt in to caching without a redeploy.

### Resolution priority

`user override → org override → flag default → RESULTS_CACHE_ENABLED env → false`

### Compatibility matrix

| DB state | ENV | Result | Who this is |
| --- | --- | --- | --- |
| no row | `true` | cache **on** | self-hosters and existing cloud (today's behavior) |
| no row | unset/`false` | cache **off** | self-hosters and shared instances (today's behavior) |
| `default_enabled=false`, no override | anything | cache **off** | shared instances opting in per-org |
| `default_enabled=false`, org override `=true` | anything | cache **on** for that org only | shared-instance customer enabled |
| `default_enabled=true` | anything | cache **on** | DB-enforced "always on" |

Existing customers with the env var set keep working with no DB changes.

### Backend changes

- New `ResultsCacheEnabled` flag + `getWithEnvFallback` helper in `FeatureFlagModel` (reusable for any future DB-first/env-fallback flag).
- `FeatureFlagModel.getFromDatabase` now skips the user-override lookup when `userUuid` isn't a valid UUID. Embed accounts use a non-UUID `external::…` ID; without this guard the user query raises a Postgres uuid type error and the catch silently swallows the org-override lookup too.
- `CommercialCacheService.findCachedResultsFile(projectUuid, cacheKey, user)` self-protects: gates internally on the FF, and exposes `isResultsCacheEnabled(user)` for the pagination path that previously read the sync `isEnabled` getter.
- `ICacheService.isEnabled` (sync boolean) replaced with `isResultsCacheEnabled(user)` (async). `findCachedResultsFile` now requires a `user` arg.
- `ProjectService.getResultsFromCacheOrWarehouse` widened to accept the user context. The embed path (`EmbedService._runEmbedQuery`) and the regular path both pass `account.user` / `account.organization` through.
- `AsyncQueryService.findResultsCache` accepts the `Account` and forwards user context to the cache service.

### Frontend changes

- `useFieldValues` reads the flag via `useServerFeatureFlag(ResultsCacheEnabled)` instead of `useHealth().hasResultsCaching`.
- `hasResultsCaching` removed from `HealthState` and `/api/v1/health` (it can't carry per-org values; the flag endpoint already exists).

### Breaking change

- `/api/v1/health` no longer returns `hasResultsCaching`. Anything scraping that field should switch to `GET /api/v2/feature-flag/results-cache-enabled`.

### Tested

- PAT path: flag on → cache hit on second run; flag off → no hit even with prior matching row.
- Embed path: with org override `=false` → no cache; remove override → cache hits resume. Confirms the UUID-skip fix is required for embed.
- Env-fallback: DB row deleted with env unset → resolves false (env fallback works).

### Out of scope

- `AUTOCOMPLETE_CACHE_ENABLED` and the autocomplete cache path are unchanged from main.

### DIFF
--- a/packages/backend/src/ee/index.ts
+++ b/packages/backend/src/ee/index.ts
@@ -381,6 +381,7 @@ export async function getEnterpriseAppArguments(): Promise<EnterpriseAppArgument
                     queryHistoryModel: models.getQueryHistoryModel(),
                     lightdashConfig: context.lightdashConfig,
                     storageClient: clients.getResultsFileStorageClient(),
+                    featureFlagModel: models.getFeatureFlagModel(),
                 }),
             mcpService: ({ context, repository, models }) =>
                 new McpService({
--- a/packages/backend/src/ee/services/CommercialCacheService.ts
+++ b/packages/backend/src/ee/services/CommercialCacheService.ts
@@ -1,13 +1,19 @@
+import { FeatureFlags } from '@lightdash/common';
 import { type S3ResultsFileStorageClient } from '../../clients/ResultsFileStorageClients/S3ResultsFileStorageClient';
 import type { LightdashConfig } from '../../config/parseConfig';
+import { FeatureFlagModel } from '../../models/FeatureFlagModel/FeatureFlagModel';
 import { QueryHistoryModel } from '../../models/QueryHistoryModel/QueryHistoryModel';
-import type { ICacheService } from '../../services/CacheService/ICacheService';
+import type {
+    CacheServiceUser,
+    ICacheService,
+} from '../../services/CacheService/ICacheService';
 import { type CacheHitCacheResult } from '../../services/CacheService/types';
 
 type CacheServiceDependencies = {
     queryHistoryModel: QueryHistoryModel;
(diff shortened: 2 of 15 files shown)


rephus:
* **#22967** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22967?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22988, opened 2026-05-13
fix: dedupe identical SCIM roles before validation to handle Entra PATCH Add duplicates

### LINKED ISSUE #23037: SCIM PATCH Add from Entra fails validation with "found 2 organization roles" when adding existing role
## Description

When Microsoft Entra runs an "Update changed user" sync job, it sometimes emits a SCIM PATCH `Add` operation for `roles` containing a role that the user already has assigned. Per RFC 7644 §3.5.2.1, `Add` on a multi-valued attribute appends, so the post-patched roles array ends up with identical duplicate entries.

Lightdash's `validateRolesArray` then rejects the request with one of:

- `Roles array must contain exactly one organization role, found 2`
- `Roles array can only contain one role per project. Duplicate project UUIDs: …`

The user's role hasn't actually changed — Entra is re-asserting the same role — but the SCIM update fails.

## Steps to reproduce

1. Provision a user via Entra SCIM with an organization role (e.g. `admin`).
2. Trigger an Entra "Update changed user" sync that emits a PATCH with `op: Add`, `path: roles`, value containing the user's existing role.
3. Lightdash returns 400 `invalidValue` instead of treating the duplicate Add as a no-op.

## Expected behaviour

Per RFC 7644 §3.5.2.1, "if the target location already contains the value specified, no changes SHOULD be made to the resource, and a success response MUST be returned." Identical duplicate roles should be collapsed silently; only genuinely conflicting roles (e.g. `admin` + `editor` at org level, or two different roles for the same project) should fail validation.

## Fix

PR #22988 — dedupes roles by `value` inside `validateRolesArray` before running conflict checks.

rephus:
Closes: https://github.com/lightdash/lightdash/issues/23037

### Description:

When Microsoft Entra runs "Update changed user" sync jobs, it sometimes emits a SCIM PATCH `Add` operation for `roles` on top of a user's already-assigned role. This caused identical duplicate role entries to appear in the roles array, which would incorrectly trigger validation errors like "found 2 organization roles" or "duplicate project roles".

`validateRolesArray` now deduplicates roles by value before running validation, silently collapsing identical duplicates. Conflicting organization roles (e.g. both `admin` and `editor`) or conflicting per-project roles still throw as expected. The method now returns the deduped roles array, which is then passed through to `upsertUserRoles` in both the create and update user flows.

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.test.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.test.ts
@@ -853,6 +853,92 @@ describe('ScimService', () => {
             );
         });
 
+        test('should dedupe identical organization roles (Entra PATCH Add on top of existing role)', () => {
+            const roles = [
+                {
+                    value: 'admin',
+                    display: 'Admin',
+                    type: 'Organization',
+                    primary: true,
+                },
+                {
+                    value: 'admin',
+                    display: 'Admin',
+                    type: 'Organization',
+                    primary: false,
+                },
+            ];
+
+            const result = ScimService.validateRolesArray(
+                roles,
+                validRoleValues,
+            );
+            expect(result).toHaveLength(1);
+            expect(result[0].value).toBe('admin');
+        });
+
+        test('should dedupe identical project roles', () => {
+            const roles = [
+                {
+                    value: 'admin',
+                    display: 'Admin',
+                    type: 'Organization',
+                    primary: true,
+                },
+                {
+                    value: 'project-1-uuid:viewer',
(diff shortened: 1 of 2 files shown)


rephus:
* **#22988** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22988?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22988?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #22995, opened 2026-05-13
fix(echarts): scope continuousDateRange to the actual X axis

almeidabbm:
Closes: GLITCH-439

## Summary

Fixes a chart rendering bug where the axes display `1970-01-01` (epoch zero) when a date dimension is placed on the Y axis, in both flipped and non-flipped layouts.

## Root cause

`continuousDateRange` was computed via an unconditional fallback chain:

```ts
bottomAxisExtraConfig.data ?? topAxisExtraConfig.data ?? leftAxisExtraConfig.data ?? rightAxisExtraConfig.data
```

introduced in #21480. The left/right fallbacks were intended to handle **flipped** (horizontal bar) layouts, where the X field visually sits on the left axis. But the fallback fired unconditionally — so in a normal layout, a date dim on Y would leak into the chain via `leftAxisExtraConfig.data`. The same problem mirrored in flipped layouts via `rightAxisExtraConfig.data` (the secondary Y-field metric).

That value then flowed into `padDatasetForContinuousAxis(rows, range, xField)`, which rewrote every row's `xField` column with date strings from the range — corrupting `dataset.source` and producing the epoch-zero signature in ECharts.

PR #22952 did not introduce the corruption, but unmasked it by routing stack decorations (rounded corners, stack totals) through the padded dataset instead of the un-padded `rows` source.

## Fix

`continuousDateRange` now reads only from the axis that actually holds the X field:

- `flipAxes === false` → `bottomAxisExtraConfig.data`
- `flipAxes === true` → `leftAxisExtraConfig.data`

The top/right secondary axes are dropped from the chain — `topAxisXId` is `undefined` in non-flipped (so `topAxisExtraConfig.data` was always `undefined` there anyway), and `rightAxisYId` in flipped falls back to `layout.yField?.[1]` which is a Y-field metric, not the X field.

**Before**
CleanShot 2026-05-13 at 10 46 38

Flipped
CleanShot 2026-05-13 at 10 46 50

**After**

CleanShot 2026-05-13 at 10 46 13

Flipped
CleanShot 2026-05-13 at 10 45 05

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.test.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.test.ts
@@ -14,6 +14,7 @@ import {
     getCategoryDateAxisConfig,
     getMinAndMaxValues,
     padDatasetForContinuousAxis,
+    selectContinuousDateRange,
 } from './useEchartsCartesianConfig';
 
 vi.mock('./../../providers/TrackingProvider');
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#22995** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/22995?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23009, opened 2026-05-13
feat: add `write_slack_summary` tool for dedicated agent Slack summaries

notgiorgi:
Closes:

### Description:

Introduces a dedicated `write_slack_summary` tool for the managed agent to explicitly capture the Slack summary it wants to post, rather than relying on the agent's last text message as the summary.

The agent now calls `write_slack_summary` once at the end of its run with the formatted Slack message. The client intercepts this tool call, stores the summary, and returns a synthetic success response so the agent can continue. The captured `slackSummary` is then returned from the session and used when posting to Slack and persisting the run result.

The agent system prompt has been updated to include this as a final checklist step (step 6), instructing the agent to call `write_slack_summary` exactly once after completing its work. The previous instruction to use the Slack tone-of-voice skill when writing the final summary has been moved into this new step.

`write_slack_summary` is also excluded from updating the run's `current_activity` display, since it is an internal bookkeeping call rather than a meaningful user-facing action.


notgiorgi:
* **#23009** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23009?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23029, opened 2026-05-13
feat(ai-agent): polish tool-call activity card

joaoviana:
## Summary

UI improvements to the AI agent tool-call activity card. Pure visual + ergonomic polish — no behaviour changes to the agent itself, no dependency on the discoverFields subagent work that's stacked on top.

- Shared `ToolCallChip` component that consolidates the six description sites where the same `<Badge color="gray" variant="light" size="xs" radius="sm" style={{textTransform:'none',fontWeight:400}}>` block was duplicated. Wraps `TruncatedText` so long values ellipsize at a consistent width with a hover tooltip.
- Guard `getToolCallChipLabel` and `ToolCallDescription` against undefined `toolArgs` mid-stream — was crashing `<LatestRow>` with `Cannot read properties of undefined (reading 'title')` when a tool call arrived before its input chunks.
- Right→down chevron on both `ToolCallRow` and `LiveActivityCard` (`IconChevronRight` + `rotate(90deg)`).
- Explicit `font-family: var(--mantine-font-family)` on `.label` and `.latestLabel` — Mantine v8's `<Text>` was applying its own `font-family: sans-serif` baseline that shadowed parent inheritance and broke Inter.
- `font-weight: 500` on activity labels.
- Card horizontal padding `7px 12px` (was `7px 0`, content was flush to the card edge).
- Icon chip: white background + 1px subtle border + soft shadow instead of the previous heavy gray fill.
- Shimmer fades opacity at the loop boundaries — was visibly snapping back from `125%` to `-25%` at full opacity.
- Tightened expand-body padding `10px / 8px` → `2px / 6px`.
- Removed `.rowButton:hover` background.
- 20px breathing room for markdown tables (`.aiMarkdown > * + table` and `> table + *`) — the rounded outline makes them read as cards, so the default 12px sibling gap felt cramped.
- New `extraBody` slot on `ToolCallRow` so any tool group can plug nested content into the expanded body. Generic; consumed by the subagent PR on top of this for the trace, but useful for any future tool that wants nested detail.

## Regression analysis

Visible to all users on web AI chat — no flag. Affects only the activity-card chrome and styling, not any tool-call routing or backend behaviour. The title-undefined guard makes the activity card *more* robust during streaming (was crashing in some races). Description-component refactors preserve the exact rendered output of each existing component; the only change is the chip's shared styling and tooltip behaviour for long labels.

## Test plan

- [x] `pnpm -F frontend typecheck`
- [x] Manual: existing AI thread renders correctly, no console errors
- [x] Activity card expand/collapse works with new chevron direction
- [x] Chips ellipsize with tooltip on hover for long field names


joaoviana:
Closing to squash overhaul in one commit.

PR #23030, opened 2026-05-13
chore(ai-agent): data-discovery subagent for the AI analyst, behind flag

joaoviana:
## Summary

Introduces a `discoverFields` tool that replaces `findExplores` + `findFields` in the parent analyst's toolset **when the flag `AiDiscoverFieldsSubagent` is on**. Internally the tool runs a focused subagent over the same model with just those two tools and an extracted ambiguity-checker system prompt, returning a schema-validated handoff (`resolved | ambiguous | no_match`) with the chosen explore + a filtered list of fields.

**Goals** — address two compounding pains in the analyst's discovery flow:
- **(A) context bloat** — parent no longer carries full field metadata (descriptions + aiHints for 200+ fields) across every turn.
- **(C) accuracy** — the ambiguity-checker prompt lives in a dedicated, focused window instead of competing with query/viz reasoning in the parent.

Stacked on top of #23062 (`parent_tool_call_id` column).

Full RFC including motivation, schema design, alternatives, and findings from the trial: `~/Desktop/ai-discover-fields-subagent-rfc.md`.

## What's in this PR

- **`FeatureFlags.AiDiscoverFieldsSubagent`** (default off). `AiAgentService` evaluates per-user/per-org and threads `useDiscoverFieldsSubagent` through `AiAgentArgs`. In `agentV2.getAgentTools`: ON → register `discoverFields`, omit `findExplores`/`findFields`; OFF → register the legacy pair, omit `discoverFields`.
- **System prompt split.** `getSystemPromptV2` picks `SYSTEM_PROMPT_TEMPLATE` (discoverFields wording — the new default once the flag rolls out) when ON, or `SYSTEM_PROMPT_TEMPLATE_LEGACY` (a **verbatim snapshot of current `main`**) when OFF.
- **New `agents/discoverFields/` directory:** `schema.ts` (Zod; the discriminated union is wrapped in `{ handoff: union }` because Anthropic rejects unions at the top level of a tool input schema), `systemPrompt.ts` (the extracted ambiguity checker + decision procedure), `agent.ts` (the inner `generateText`), `tool.tsx` (parent-facing tool).
- **`stripAnthropicThinking()`** removes the `thinking` provider option for the subagent only — Anthropic rejects extended-thinking + forced `tool_choice`, and `experimental_output` forces tool choice to `"any"`. Parent keeps thinking enabled.
- **Step cap = 10** on the subagent. Realistic worst path is ~5 steps; gives slack without enabling runaway loops.
- **Distinct telemetry `functionId: 'discoverFieldsSubagent'`** on the subagent's `generateText` call so latency, token usage, and step counts surface independently of the parent. Telemetry-config factory extracted to a shared `agents/telemetry.ts` module (avoids an `agentV2` ↔ subagent import cycle).
- **Subagent persistence** uses `parentToolCallId` (the new column from #23062) to link internal `findExplores`/`findFields` calls back to the parent's `discoverFields` toolCallId.
- **`ToolNameSchema`** and the two `TOOL_DISPLAY_MESSAGES` records in `@lightdash/common` gain `discoverFields` so persisted tool calls parse correctly on thread fetch and surface a friendly label.
- **Frontend:** `DiscoverFieldsTrace` renders the subagent's internal trace under the parent `Discovered fields` row in the activity card. Trace data is read from `metadata.trace` on the tool result (JSONB) — switching the UI to query by `parent_tool_call_id` is a follow-up PR.

## Regression analysis

**Flag OFF (default for every org):** byte-for-byte equivalent to current `main`. The legacy `findExplores` + `findFields` tools are registered and `SYSTEM_PROMPT_TEMPLATE_LEGACY` is a verbatim snapshot of today's `systemV2Template.ts`. No behavioural change for any org until the flag is flipped.

**Flag ON (no orgs yet — dogfood gated):** the parent agent calls `discoverFields` as a single tool replacing the previous `findExplores` → `findFields` pair. Verified in the trial across:
- *Resolved* — single-explore unambiguous queries pick the right explore + 5–20 relevant fields, parent runs `runQuery`.
- *Ambiguous* — generic metric queries return candidate explores + a `suggestedQuestion`; parent does **not** call `runQuery`.
- *Thread pivot* — different topic mid-thread triggers a re-call; subagent switches explore.
- *Same-topic follow-up* — parent reuses fields without re-calling `discoverFields`.

## Findings during the trial (also in RFC §10)

- **Anthropic `thinking` × forced `tool_choice` is invalid.** `experimental_output` forces tool choice to `"any"`, which Anthropic refuses to combine with `thinking: 'enabled'`. `stripAnthropicThinking()` removes `thinking` for the subagent only.
- **Zod discriminated unions need an object wrapper** for `experimental_output: Output.object({ schema })` because Anthropic requires `type: "object"` at the top level. Wrap with `z.object({ handoff: union })` and unwrap in the agent.
- **`ToolNameSchema` gating** is mandatory — every persisted tool call is parsed against this enum on thread fetch; without the addition, every subsequent thread `GET` 500s.
- **Three frontend switch sites** for any new agent tool: `ToolCallDescription.tsx` (uses `assertUnreachable` — hard crash on missing case), `toolIcons.ts` (`Record<ToolName, ...>` — TS error), `getActivityTitle.ts` (`SEARCH_TOOLS` / `VIZ_TOOLS` classifier sets). Worth a "checklist when adding a new tool" doc.
- **Live progress streaming is its own thing.** The AI SDK 6 subagents pattern supports yielding `UIMessage`s from an async-generator `execute` (via `readUIMessageStream` over `result.toUIMessageStream()`), surfaced in the UI as `part.state === 'output-available' && part.preliminary === true`. Plumbing this requires changing the tool's output shape and threading partial outputs through `agentV2.ts`'s `onChunk` handler. Follow-up PR on top of this one.

## Follow-ups (separate PRs)

- Switch the UI trace nesting from `metadata.trace` (JSONB) to a query by `parent_tool_call_id` (DB column populated by this PR).
- Live progress streaming via the AI SDK preliminary-tool-results pattern.
- `storeToolCall` error suppression cleanup — `AiAgentService.ts` does `void wrapSentryTransaction(...)` so the inner promise is fire-and-forget; `wrapSentryTransaction` only logs the catch when Sentry is enabled. Not subagent-specific; separate PR.
- Curated eval reference set (~20–30 questions) before flipping the flag default to on. Coverage criteria in RFC §5.

## Migration

```bash
pnpm -F backend migrate
```

(Applied by stacked parent #23062 — but listing here so reviewers running the PR locally remember to run migrations before testing.)

## Test plan

- [x] `pnpm -F common typecheck && pnpm -F backend typecheck && pnpm -F frontend typecheck`
- [x] `pnpm -F backend lint && pnpm -F frontend lint`
- [x] Manual browser tests across resolved / ambiguous / pivot / follow-up cases (with flag ON)
- [x] DB column populated for subagent calls; trace renders in activity card
- [ ] **Flag-OFF smoke test** — verify with the flag unset that the analyst's flow is byte-equivalent to current `main` (reviewer)
- [ ] Curated eval reference set (follow-up, before flipping flag default)


joaoviana:
* **#23070** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23070?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23030** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23062** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #23035, opened 2026-05-13
feat(database): Make migration verification configurable

DonoA:
When practicing an emergency rollback, we discovered that it's much easier to simply disable the migration verification in some cases instead of attempting to downgrade the database to match the currently migrated version. 

To make this as flexible as possible, I've introduced an environment variable flag which allows these migrations to be disabled in some cases.


ZeRego:
Solved in https://github.com/lightdash/lightdash/pull/23139/changes

PR #23054, opened 2026-05-14
fix(dashboards): preserve filter overrides when duplicating a chart tile

sdolidze:
Closes: PROD-7487

## Summary

Duplicating a chart tile on a dashboard silently reset that tile's per-filter configuration. Filters the user had disabled on the source tile re-enabled themselves on the duplicate, and any custom field-mapping override was dropped.

Only the "Duplicate chart" action on a dashboard tile is affected. Adding a new chart, editing a chart, and duplicating an entire dashboard are unchanged.

## Root cause

Per-tile filter overrides live in `DashboardFilterRule.tileTargets`, keyed by tile UUID. When a chart is duplicated, a fresh tile UUID is minted via `uuid4()` and the tile is appended through `onAddTiles`, but `tileTargets` is never touched — so the new UUID is absent from the map.

```ts
// getDashboardFilterRulesForTile — packages/common/src/utils/filters.ts
const tileConfig = filter.tileTargets?.[tileUuid];
if (tileConfig === undefined) {
    // ...falls through to auto-apply the filter with default targeting
    return filter;
}
```

A missing entry is treated as "apply this filter to the tile with default targeting," so every matching filter snaps back on and any custom mapping is discarded. The source tile's entry is still in `tileTargets` — it just isn't being carried over to the new UUID.

```
source tile A  (uuid = a)
filter F.tileTargets = { a: { fieldId: "orders.status", isEnabled: false } }

Duplicate A -> new tile B (uuid = b)

Before:
  filter F.tileTargets = { a: { ... } }
  b not in tileTargets  ->  default auto-apply  ->  override lost

After:
  filter F.tileTargets = { a: { ... }, b: { ... } }   (b copied from a)
                          ->  duplicate inherits source's override
```

## Fix

`onAddTiles` now accepts an optional `tileUuidMapping: Record<newUuid, sourceUuid>`. When `Dashboard.tsx` receives a non-empty mapping, it walks every dimension filter and copies the source tile's `tileTargets` entry onto the new tile's UUID, then flags filters as changed. The chart-duplicate flow is the only caller that passes the mapping today; every other caller forwards `undefined` and behaves identically.

- `packages/frontend/src/pages/Dashboard.tsx` — `handleAddTiles` rewrites `dashboardFilters.dimensions[*].tileTargets` for the mapped UUIDs and sets `haveFiltersChanged`.
- `packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx` — duplicate-chart effect mints the new tile UUID up front and passes `{ [newUuid]: props.tile.uuid }`.
- `AddTileButton`, `AddChartTilesModal`, `DashboardHeader`, `EmptyStateNoTiles`, `GridTile`, `dashboardTabs/index` — widened `onAddTiles` signature to forward the optional mapping; no behaviour change.
- Out of scope: `metrics` and `tableCalculations` filter arrays. The dashboard model only writes `tileTargets` on `dimensions`, which matches the existing reducer surface.

## Test plan

- [x] `pnpm -F frontend typecheck && pnpm -F frontend lint`
- [ ] Manual: on a dashboard with a filter disabled for chart A, duplicate A and confirm the duplicate also has the filter disabled
- [ ] Manual: on a dashboard with a filter whose field is remapped for chart A, duplicate A and confirm the duplicate inherits the same remap
- [ ] Manual: add a brand-new chart tile (not a duplicate) and confirm filters still auto-apply as before

### DIFF
--- a/packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
+++ b/packages/frontend/src/components/DashboardTiles/AddTileButton.tsx
@@ -32,7 +32,11 @@ import AddChartTilesModal from './TileForms/AddChartTilesModal';
 import { TileAddModal } from './TileForms/TileAddModal';
 
 type Props = {
-    onAddTiles: (tiles: Dashboard['tiles'][number][]) => void;
+    onAddTiles: (
+        tiles: Dashboard['tiles'][number][],
+        // Map of new tile UUID → source tile UUID, so dashboard filter `tileTargets` are copied from the source.
+        tileUuidMapping?: Record<string, string>,
+    ) => void;
     setAddingTab: (value: React.SetStateAction<boolean>) => void;
     activeTabUuid?: string;
     dashboardTabs?: Dashboard['tabs'];
--- a/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
+++ b/packages/frontend/src/components/DashboardTiles/DashboardChartTile.tsx
@@ -544,7 +544,11 @@ interface DashboardChartTileMainProps extends Pick<
     tile: IDashboardChartTile;
     dashboardChartReadyQuery: DashboardChartReadyQuery;
     resultsData: InfiniteQueryResults;
-    onAddTiles?: (tiles: Dashboard['tiles'][number][]) => void;
+    onAddTiles?: (
+        tiles: Dashboard['tiles'][number][],
+        // Map of new tile UUID → source tile UUID, so dashboard filter `tileTargets` are copied from the source.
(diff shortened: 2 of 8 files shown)


sdolidze:
* **#23054** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23054?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23054?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23060, opened 2026-05-14
fix(timezone): skip date-trunc/extract wrap when target tz matches source

almeidabbm:
Closes GLITCH-444

## Summary

When the project query timezone matches the warehouse column timezone (the default UTC↔UTC case for vanilla BigQuery with `EnableTimezoneSupport` on), the timezone-aware DATE_TRUNC / EXTRACT paths still emit a tz round-trip that is semantically a no-op but **defeats partition pruning** on BigQuery — turning scoped queries into full-table scans and ballooning billed bytes.

This consolidates the no-op short-circuit at the boundary so every caller is gated by the same rule.

### DATE_TRUNC — DAY dim, project tz `UTC`

Before:
```sql
SELECT TIMESTAMP_TRUNC(TIMESTAMP(`t`.event_timestamp), DAY, 'UTC') AS `day` ...
```
After:
```sql
SELECT TIMESTAMP_TRUNC(`t`.event_timestamp, DAY) AS `day` ...
```

### DATE_TRUNC — DAY dim, project tz `America/New_York` (preserved)

```sql
SELECT TIMESTAMP_TRUNC(TIMESTAMP(`t`.event_timestamp), DAY, 'America/New_York') AS `day` ...
```

### EXTRACT — HOUR_OF_DAY dim, project tz `UTC`

Before:
```sql
SELECT EXTRACT(HOUR FROM TIMESTAMP(`t`.event_timestamp) AT TIME ZONE 'UTC') AS `hour` ...
```
After:
```sql
SELECT EXTRACT(HOUR FROM `t`.event_timestamp) AS `hour` ...
```

### EXTRACT — HOUR_OF_DAY dim, project tz `America/New_York` (preserved)

```sql
SELECT EXTRACT(HOUR FROM TIMESTAMP(`t`.event_timestamp) AT TIME ZONE 'America/New_York') AS `hour` ...
```

### FORMAT — MONTH_NAME dim, project tz `UTC`

Before / After (now stable at the byte level):
```sql
SELECT FORMAT_DATETIME('%B', `t`.event_timestamp) AS `month_name` ...
```

## Root cause

The equality short-circuit (`columnTimezone === args.timezone → undefined`) lived only in `MetricQueryBuilder.timezoneForDateTrunc`, but two call sites destructured `args.timezone` directly and bypassed the getter — and `getSqlForTruncatedDate` / `getSqlForDatePart` / `getSqlForDatePartName` only bailed on `!timezone || type !== TIMESTAMP`, never on `target === source`. So the wrap fired even when source and target both defaulted to UTC.

## What changed

- **`packages/common/src/utils/timeFrames.ts`** — new `resolveTimezoneWrap(type, timezone, sourceTimezone)` helper that returns the resolved wrap pair or `null` (treating missing source as UTC). Applied at all three boundary functions: `getSqlForTruncatedDate`, `getSqlForDatePart`, `getSqlForDatePartName`.
- **`packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts`** — `timezoneForDateTrunc` getter simplified to its feature-flag gate; the equality check now lives at the boundary so call sites that bypass the getter are still covered.
- **`packages/common/src/utils/timeFrames.test.ts`** — updated the stale BQ wrap test (was asserting `'UTC'` which is now the no-op) to use a non-UTC tz. Added regression tests asserting UTC-on-UTC and matching non-UTC source/target short-circuit for all three paths.

## Test plan

- [x] Typecheck + lint clean (`common`, `backend`)
- [x] Unit tests pass: `timeFrames` (83/83), `MetricQueryBuilder` (179/179, 60 snapshots), all tz-adjacent backend suites (276/276)
- [x] Pre-existing common failures (`filtersCompiler`, `formatting`) verified identical on `main` — not introduced here
- [x] Live `/compileQuery` sanity across BigQuery, Postgres, ClickHouse, Databricks, Trino, Snowflake (default + `disableTimestampConversion` with non-UTC `dataTimezone`) — UTC tz drops the wrap, non-UTC preserves it, on both TRUNC and EXTRACT paths

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts
@@ -319,11 +319,14 @@ export class MetricQueryBuilder {
         CompiledDimension
     > = {};
 
-    /** Query timezone when timezone-aware DATE_TRUNC is active, undefined otherwise. */
+    /** Query timezone when timezone-aware DATE_TRUNC is active, undefined otherwise.
+     *  The target-vs-source equality short-circuit lives in `getSqlForTruncatedDate`
+     *  (and its EXTRACT siblings) so any call site that passes a timezone gets the
+     *  same no-op treatment — see `needsTimezoneWrap` in `timeFrames.ts`. */
     private get timezoneForDateTrunc(): string | undefined {
-        if (!this.args.useTimezoneAwareDateTrunc) return undefined;
-        if (this.columnTimezone === this.args.timezone) return undefined;
-        return this.args.timezone;
+        return this.args.useTimezoneAwareDateTrunc
+            ? this.args.timezone
+            : undefined;
     }
 
     private get columnTimezone(): string {
--- a/packages/common/src/utils/timeFrames.test.ts
+++ b/packages/common/src/utils/timeFrames.test.ts
@@ -104,11 +104,72 @@ describe('TimeFrames', () => {
                     '${TABLE}.created',
                     DimensionType.TIMESTAMP,
                     undefined,
-                    'UTC',
+                    'America/New_York',
                 ),
             ).toEqual(
-                "TIMESTAMP_TRUNC(TIMESTAMP(${TABLE}.created), DAY, 'UTC')",
+                "TIMESTAMP_TRUNC(TIMESTAMP(${TABLE}.created), DAY, 'America/New_York')",
             );
+        });
+
+        test('UTC target with default (UTC) source short-circuits the wrap', () => {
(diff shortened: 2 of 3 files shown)


almeidabbm:
* **#23060** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23060?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23060?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23079, opened 2026-05-14
fix: clear stale field values on filter autocomplete force-refresh

### LINKED ISSUE #23039: [Filters] Force-refresh on filter autocomplete does not clear stale values from dropdown
## Bug description

When a user clicks the refresh button on the filter autocomplete dropdown, the backend correctly bypasses the S3 cache and returns fresh values from the warehouse. However, the frontend `useFieldValues` hook accumulates results in a `Set` that is never cleared on force-refresh - only when the field or filters change. This means stale values (e.g. a renamed dimension value) persist in the dropdown even after clicking refresh.

## Steps to reproduce

1. Have a dimension with value "Old Name" visible in a filter dropdown
2. Rename the value to "New Name" in the source data
3. Open the filter dropdown - "Old Name" still appears (expected, cached)
4. Click the refresh icon at the bottom of the dropdown
5. "Old Name" still appears in the dropdown alongside "New Name"
6. A full page reload (F5) after the refresh correctly shows only "New Name"

## Expected behavior

After clicking the refresh icon, the dropdown should show only the fresh values returned from the warehouse. The old value should no longer appear.

## Actual behavior

The old value persists because `handleUpdateResults` in `useFieldValues.ts` (line 253) merges new results into the existing Set:
```js
setResults((oldSet) => new Set([...oldSet, ...data.results].sort(...)))
```

The `results` Set is only cleared when the field name changes or filters change (lines 235-238, 349-356), not on force-refresh.

## Workaround

Click the refresh icon in the filter dropdown (this updates the server-side cache), then do a full page reload (F5 / Cmd+R). The next page load will fetch the updated cached values without the stale entries.

## Environment

- Affects Lightdash Cloud (and any instance with autocomplete caching enabled)
- Frontend code: `packages/frontend/src/hooks/useFieldValues.ts`
- Trigger component: `packages/frontend/src/components/common/Filters/FilterInputs/FilterStringAutoComplete.tsx`

Tarunswamy-Muralidharan:
## Bug description

When a user clicks the refresh button on the filter autocomplete dropdown, the backend correctly bypasses the S3 cache and returns fresh values from the warehouse. However, the frontend `useFieldValues` hook accumulated results in a Set that was never cleared on force-refresh -- stale values persisted alongside fresh ones.

## Root cause

The `useEffect` that resets `searches` / `results` / `resultCounts` already had `forceRefresh` in its dependency array, but only checked for field name changes. This meant clicking refresh fetched fresh data but merged it into the existing accumulated Set instead of replacing it.

## Fix

Added `forceRefresh` to the reset condition so the accumulated state is cleared when force-refresh is triggered.

## How I verified

1. Traced the flow in `FilterStringAutoComplete.tsx` -> `useFieldValues.ts`
2. Confirmed `forceRefresh` goes `true` -> `refetch()` -> `setForceRefresh(false)` before `onSuccess` fires
3. The existing `useEffect` with `forceRefresh` in deps fires on the prop change, clearing state BEFORE `handleUpdateResults` merges new results

Fixes #23039

### DIFF
--- a/packages/frontend/src/hooks/useFieldValues.ts
+++ b/packages/frontend/src/hooks/useFieldValues.ts
@@ -348,10 +348,12 @@ export const useFieldValues = (
         [search],
     );
 
-    // reset state when field changes
+    // reset state when field changes or force-refresh is triggered
     useEffect(() => {
         if (!!fieldName && field.name !== fieldName) {
             setFieldName(field.name);
+        }
+        if ((!!fieldName && field.name !== fieldName) || forceRefresh) {
             setSearches(new Set<string>());
             setResults(new Set(initialData));
             setResultCounts(new Map());


ZeRego:
Closing in favour of https://github.com/lightdash/lightdash/pull/23085

PR #23090, opened 2026-05-14
feat(filters): allow dashboard editors to lock filters (PROD-7564)

charliedowler:
## Summary

Adds the ability for dashboard editors to **lock** a dashboard filter. Locked filters are read-only for viewers, and URL/embed overrides targeting a locked field are ignored. Shipped behind the `lock-dashboard-filters` feature flag.

- New optional `locked?: boolean` on `DashboardFilterRule` (stored in the existing JSONB column — no migration).
- Hover-revealed lock/unlock toggle in the filter chip's right side (in edit mode). Already-locked chips show the lock indicator at rest in both edit and view mode.
- Override stripping at three input boundaries: URL `?filters=` / `?tempFilters=` (frontend `DashboardProvider`), sessionStorage replay, and embed SDK `externalFilters` (EE `EmbedService`).
- View-mode interaction: clicking a locked chip is a no-op; the existing tooltip explains the lock state.
- One-time toast when overrides are dropped because of a lock; toast guard resets on dashboard navigation.

Relates to PROD-7564.

## Safety

Designed for safe deploy independent of release:

- Type change is additive optional — old saved filters parse as `locked: undefined` (falsy).
- The strip helper (`stripOverridesForLockedFilters`) is a pure identity when no rule has `locked: true`. Backend `_getAppliedDashboardFilters` rewrite preserves bit-for-bit behaviour when no locked rules exist.
- Flag off → chip toggle button never renders, `draftFilterRule.locked` is `undefined`, no UI surface visible.
- Flag on per-org → that org's editors get the chip toggle; only locked dashboards diverge in behaviour, everywhere else is unchanged.
- No DB migration. No new API endpoints. No `generate-api` regeneration needed.

## Test plan

- [ ] In an org with the flag on, hover any filter chip in dashboard **edit mode** — an open-padlock button slides in next to the X.
- [ ] Click the padlock — the icon switches to closed, stays visible at rest, and `locked: true` is persisted on save.
- [ ] Switch to view mode — the locked chip shows a closed-lock indicator on the right; clicking the chip is a no-op with a tooltip explaining why.
- [ ] Append `?tempFilters=...` to the URL targeting a locked field — verify (a) the locked value is honoured (URL is cleaned of the override on load), (b) a one-time toast says the URL override was ignored.
- [ ] In edit mode, open a locked filter's settings popover, clear its value while `Required` is on — `Apply` is disabled with "A locked, required filter must have a value".
- [ ] Generate an embed JWT with `externalFilters` targeting a locked field — server-side queries use the locked saved value (verify via Spotlight trace or scheduler logs).
- [ ] Turn the flag off → chip toggle disappears, but existing locked filters keep being enforced server-side (safety valve).
- [ ] All 46 common unit tests pass (`pnpm -F common test`).

### DIFF
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -56,6 +56,7 @@ import {
     SavedChartsInfoForDashboardAvailableFilters,
     SessionAccount,
     SortField,
+    stripOverridesForLockedFilters,
     UpdateEmbed,
     UserAccessControls,
     UserAttributeValueMap,
@@ -966,7 +967,6 @@ export class EmbedService extends BaseService {
         return chart;
     }
 
-    // eslint-disable-next-line class-methods-use-this
     private async _getAppliedDashboardFilters(
         account: AnonymousAccount,
         explore: Explore,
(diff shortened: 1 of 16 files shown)


charliedowler:
Closing — opened from wrong head branch (prod-5931 instead of prod-7564). Reopening from prod-7564.

PR #23092, opened 2026-05-15
fix(ee): interpolate column name in ai_agent_tool_call length check

IrakliJani:
## Summary

The `20260513165925_add_parent_tool_call_id_to_ai_agent_tool_call` migration uses Knex's `??` identifier placeholder inside `table.check()`, but `table.check(predicate, bindings, name)` treats the `bindings` array as **value** placeholders, not identifier placeholders. Knex expands `??` as two positional binds → `\$1\$2`, producing invalid SQL:

\`\`\`
alter table "ai_agent_tool_call" add constraint ai_agent_tool_call_parent_tool_call_id_length_check
  check(length(\$1\$2) <= 128)
                ^ syntax error at or near "\$2"
\`\`\`

Result: `pnpm migrate-production` exits 1 on any fresh DB, so every PR preview deploy hangs in `CrashLoopBackOff` (server never starts → readiness probe never goes green → Okteto `--wait` times out at 15m).

## Fix

Interpolate the column name directly. `ColumnName` is an in-file constant, not user input, so there's no injection risk.

\`\`\`ts
table.check(\`length(\${ColumnName}) <= 128\`, [], CheckConstraintName);
\`\`\`

Same SQL semantics as intended, generates `check(length(parent_tool_call_id) <= 128)`.

### DIFF
--- a/packages/backend/src/ee/database/migrations/20260513165925_add_parent_tool_call_id_to_ai_agent_tool_call.ts
+++ b/packages/backend/src/ee/database/migrations/20260513165925_add_parent_tool_call_id_to_ai_agent_tool_call.ts
@@ -12,7 +12,11 @@ export async function up(knex: Knex): Promise<void> {
     await knex.schema.alterTable(AiAgentToolCallTableName, (table) => {
         table.text(ColumnName).nullable();
         table.index(['ai_prompt_uuid', ColumnName], IndexName);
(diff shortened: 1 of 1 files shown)


joaoviana:
Verified
1. with current migration -> error
2. with your fix -> all good

(this came from a refactoring from .raw -> using knex built in methods, which was not tested by me entirely)

PR #23100, opened 2026-05-15
fix(formula): cast dividend so int/int no longer truncates on Redshift

joaoviana:
Closes: PROD-7754

## Summary

Customer-reported via [PROD-7754](https://linear.app/lightdash/issue/PROD-7754): on Redshift, `=Total Orders / Total Accounts` in a table calc came back as `0` or `1` instead of the intended ratio. Reproduced with two `COUNT(*)` metrics — `5 / 3 → 1`, `2 / 3 → 0`.

The formula language targets Google-Sheets-style semantics where `/` is always real-valued. Postgres-family and Trino-family warehouses treat `int / int` as integer division (truncates). Fix: cast the dividend per dialect so all 9 warehouses agree.

## Root cause

`BaseGenerator.generateBinaryOp` emitted `(left / NULLIF(right, 0))` with no cast. On Postgres / Redshift / Trino / Athena, `int / int = int` — fractional part dropped silently. Other warehouses (BigQuery, Snowflake, DuckDB, Databricks, ClickHouse) auto-widen and were never broken.

## Fix

Add a `generateDivision` dialect hook (mirrors the existing `generateModulo` precedent in the same file). Default keeps the bare emission; affected dialects override:

| Dialect | Emission |
|---------|----------|
| Postgres | `((x)::numeric / NULLIF(y, 0))` |
| Redshift | `((x)::numeric(38, 10) / NULLIF(y, 0))` — bare `::numeric` is `numeric(18, 0)` on RS (zero scale, would re-truncate); same precision pin as the existing ROUND emitter |
| Trino / Athena | `(CAST(x AS DOUBLE) / NULLIF(y, 0))` |
| BigQuery / Snowflake / DuckDB / Databricks / ClickHouse | unchanged — bare default |

### Why \`::numeric\` and not \`::DOUBLE PRECISION\`

\`postgresStyleAvg\` already uses \`::DOUBLE PRECISION\` so that was the obvious first attempt — but it silently flips \`=ROUND(A / B)\` from half-away-from-zero (Google Sheets, Excel) to banker's rounding. \`round(250.5::numeric) = 251\` but \`round(250.5::double precision) = 250\` on Postgres. \`::numeric\` preserves the user-visible rounding semantics; the existing \`math/round-no-precision\` integration case caught this when I first tried DOUBLE PRECISION.

## Regression analysis

- **`int / int`** (was broken): now returns real-valued result. **Behavior change is the whole point of the PR.**
- **`decimal / int` and `int / decimal`** (was correct): unchanged at the user-visible level. The dividend gets an explicit `::numeric` cast that's a no-op for already-numeric inputs. SQL differs but values match — covered by the existing `arithmetic/divide` case still passing.
- **`decimal / decimal`** (was correct): same as above.
- **`=ROUND(A / B)`** on Postgres: rounding mode unchanged because the dividend lands as \`numeric\`, not double — covered by `math/round-no-precision` still passing.
- **Division by zero**: still short-circuits to NULL via NULLIF. Pinned in `division-by-zero short-circuits to NULL via NULLIF on all dialects` codegen test.
- **Other dialects** (BigQuery, Snowflake, DuckDB, Databricks, ClickHouse): zero SQL change — they don't override the new hook.
- **Metric-level SQL**: untouched. This only changes formula-package codegen (table calculations). Metric SQL paths in \`packages/warehouses\` are unaffected.

## Verification

| Suite | Result |
|-------|--------|
| Unit (\`packages/formula/tests\`) | 367/367 pass |
| Integration tier1 (DuckDB, Postgres, Redshift) | 360/360 on Redshift; 360/361 on Postgres (the one pre-existing failure — \`aggregation/count-distinct-over-partition-by\` — is unrelated, fails on `main` too) |
| Integration tier2 (BigQuery, Snowflake, Databricks, ClickHouse, Athena, Trino) | 722/722 |

New integration case: `arithmetic/divide-integer-columns` — fails on `main` (Postgres returns `0` instead of `0.5`), passes on this branch.
New codegen cases under `describe('division (/)')` lock the per-dialect SQL so future changes can't silently revert this.

## Linear

Closes [PROD-7754](https://linear.app/lightdash/issue/PROD-7754).


joaoviana:
* **#23100** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23100?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23100?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23104, opened 2026-05-15
perf(validation): scope project row-lock to avoid streaming dbt_connection bytea

joaoviana:
## Problem

`ValidationModel.insertValidations` and `ValidationModel.replaceProjectValidations` acquire a row lock on `projects` to serialise concurrent validation runs:

```ts
await trx(ProjectTableName)
    .where('project_uuid', projectUuid)
    .forUpdate();
```

Knex has no `.select(...)` to narrow the projection here, so the emitted SQL is `SELECT * FROM projects ... FOR UPDATE`. Postgres streams every column on the row back over the wire — including `dbt_connection`, an encrypted `bytea` blob that can be up to ~2 MB per row — and the result is then immediately discarded. The caller only wants the row lock.

Same class of bug as #22767, which narrowed the auth-path joins in `UserModel` / `RolesModel` for the same reason.

## Change

Add `.select(trx.raw('1'))` so the lock acquisition emits `SELECT 1 FROM projects ... FOR UPDATE`:

- `ValidationModel.ts:84` — `insertValidations`
- `ValidationModel.ts:154` — `replaceProjectValidations`

Lock semantics are unchanged — `FOR UPDATE` on a `WHERE project_uuid = ?` predicate locks the same single row regardless of the projection. Only the bytes Postgres ships back change.

## Regression analysis

- **Lock behaviour unchanged**: PG's row-level lock is keyed off the row identity matched by the `WHERE`, not the column list. `SELECT 1 ... FOR UPDATE` and `SELECT * ... FOR UPDATE` acquire identical locks.
- **Result unused**: in both call sites the awaited result is not assigned or read. Confirmed by reading both function bodies end-to-end.
- **No transactional side effects**: still inside the same `this.database.transaction(...)` block; lock continues to be held until commit/rollback as before.
- **Other models audited**: scanned every other model that joins `projects` (UserFavoritesModel, ResourceViewItemModel, SavedChartModel, SpaceModel, AppModel, AnalyticsModel, SavedSqlModel, SpacePermissionModel, UserWarehouseCredentialsModel, all five ContentConfigurations, WarehouseAvailableTablesModel, SearchModel, DashboardModel, SchedulerModel, EmbedModel, PreAggregateModel, AiAgentModel). All other `projects` joins already narrow their column list explicitly. These two `forUpdate()` calls were the only remaining offenders after #22767.

## Who's affected

- **All deployments** running validation — every project compile and every scheduled validation tick previously transferred the full `projects` row (including encrypted credentials blob) per validation run, just to throw it away.
- **Self-hosted instances with large `dbt_connection` payloads** benefit most. Cloud SaaS instances see a smaller absolute win but the same shape.

## Test plan

- [x] `pnpm -F backend test:dev:nowatch` — 223 tests pass (all suites for changed files)
- [x] `pnpm exec eslint src/models/ValidationModel/ValidationModel.ts` — clean
- [ ] Backend typecheck is failing on `main` for an unrelated reason (`anthropic-claude.ts` AI SDK type mismatch). This PR does not touch that file.

### DIFF
--- a/packages/backend/src/models/ValidationModel/ValidationModel.ts
+++ b/packages/backend/src/models/ValidationModel/ValidationModel.ts
@@ -82,6 +82,7 @@ export class ValidationModel {
         await this.database.transaction(async (trx) => {
             // Lock the project to avoid concurrent validation updates
             await trx(ProjectTableName)
+                .select('project_uuid')
                 .where('project_uuid', projectUuid)
                 .forUpdate();
 
@@ -152,6 +153,7 @@ export class ValidationModel {
         await this.database.transaction(async (trx) => {
             // Lock the project to avoid concurrent validation updates
             await trx(ProjectTableName)
+                .select('project_uuid')
                 .where('project_uuid', projectUuid)
                 .forUpdate();
 


joaoviana:
* **#23104** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23104?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23104?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23119, opened 2026-05-15
fix: replace user attributes in query selects

notgiorgi:
## Summary

Fixes query compilation for selected dimension and metric SQL that references Lightdash user context variables such as `${ld.user.email}`.

Before this change, those variables were replaced in filter/sqlWhere paths but not in SELECT expressions. Valid model SQL could leak through to the warehouse as raw `${...}` syntax and Postgres failed with `syntax error at or near "$"`.

## Repro steps

Using the local Jaffle shop seed project:

1. Start the local dev stack and use project `3675b69e-8324-4110-bdca-059031aa8da3` (`Jaffle shop`).
2. Use the `orders` explore, which defines `orders_user_attribute_email_test` in `examples/full-jaffle-shop-demo/dbt/models/orders.yml` with `WHEN ${ld.user.email} = 'demo@lightdash.com'`.
3. Run an underlying-data query that selects all order fields and filters `orders_status = shipped`.
4. Observe the async query failing in `query_history` with `syntax error at or near "$"`.
5. Inspect `query_history.compiled_sql` for a failed query such as `bbd37028-383b-4268-a836-ce149c36145d`; the SELECT contains raw `${ld.user.email}` in `orders_user_attribute_email_test`.

## Root cause

`MetricQueryBuilder` emitted selected dimension SQL from `getTimezoneAwareDimensionSql(...)` directly into the SELECT list, and emitted selected metric `compiledSql` directly as well. Neither path called `replaceUserAttributesAsStrings(...)`, so supported user context syntax was not resolved before the SQL reached the warehouse.

## Fix

- Replace user attributes in selected dimension SQL after timezone-aware dimension SQL resolution and before aliasing.
- Replace user attributes in selected metric SQL before aliasing.
- Use `noWrap` so SELECT expressions keep their existing expression shape.

## Validation

- `pnpm -F backend test -- src/utils/QueryBuilder/MetricQueryBuilder.test.ts --runInBand`

Added regression coverage for selected dimensions and selected metrics containing `${ld.user.email}`.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.test.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.test.ts
@@ -957,6 +957,68 @@ describe('Query builder', () => {
         ).toThrow(ForbiddenError);
     });
 
+    test('Should replace intrinsic user attributes in selected dimensions and metrics', () => {
+        const explore: Explore = {
+            ...EXPLORE,
+            tables: {
+                ...EXPLORE.tables,
+                table1: {
+                    ...EXPLORE.tables.table1,
+                    dimensions: {
+                        ...EXPLORE.tables.table1.dimensions,
+                        user_email_status: {
+                            type: DimensionType.STRING,
+                            name: 'user_email_status',
+                            label: 'user_email_status',
+                            table: 'table1',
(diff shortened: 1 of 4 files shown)


pimmey (approved):
👍

PR #23134, opened 2026-05-15
refactor(ai-agent): tighten discoverFields subagent prompt

joaoviana:
## Summary

Two prompt-only changes to the AI Analyst. No behaviour shift in the agent's tools, schema, or DB; only the system prompts that the parent and the `discoverFields` subagent read on each turn.

### 1. Move the joined-vs-base / events-vs-entities heuristic into the subagent

Previously lived in the parent template (`systemV2Template.ts`) as a runQuery-time rule, but field selection happens **inside the subagent's Step 4 (field shortlist)** — the rule should fire at the point the decision is being made, not after the handoff is already in the parent's hands. The parent now just tells the model to trust the discovery handoff's `isFromJoinedTable` marker.

### 2. Trim the subagent's "Status payloads" section from JSON examples to a 3-bullet summary

The `submitResult` tool's `inputSchema` (`discoverFieldsResultSchema`) is forwarded to the model verbatim by the AI SDK, so the field-level shape, value-type vocabulary (`number / string / date / timestamp / boolean`), and even the `suggestedQuestion` example are already visible to the model without the prompt restating them.

What we kept (rules that aren't on the Zod schema):

- **FILTERED 5–20 fields shortlist** — `z.array(...)` has no size constraint and "filtered" is judgment, not type.
- **`description` only when distinguishing similar fields** — schema makes it nullable; the conditional is behavioural.
- **No `findFields` call on ambiguous** — pure behaviour, no schema surface.
- **Single `submitResult` call as the terminal step** — invariant the schema can't enforce.
- **Brief `rationale`** — schema says `nullable string`, behavioural cue to keep it short.

## Diff

- `discoverFields/systemPrompt.ts`: **–66 lines** (84 deleted, 8 added). Status-payload section drops from ~65 lines of three JSON examples to 3 prose bullets.
- `systemV2Template.ts`: **–1 line**. One Field-usage bullet replaced with a tighter "trust the handoff" reminder.

## Regression analysis

- **Flag OFF** path (legacy parent template, `systemV2TemplateLegacy.ts`) is **untouched**. Users with the discoverFields subagent disabled see exactly the same prompt as today.
- **Flag ON** path: parent prompt loses one bullet (–67 words on parent context); subagent prompt loses ~3 KB of JSON examples but the same rules are still expressed. Schema is the source of truth — the AI SDK round-trip is identical.
- No tool wiring, no controllers, no DB shape changes. `pnpm generate-api` not required.
- The subagent's decision procedure (Steps 1–4) is unchanged; only the *result-format* prose around it shrank.

## Test plan

- [ ] `pnpm -F backend typecheck` and `pnpm -F backend lint`
- [ ] Flag ON: send a discovery prompt that resolves to a single explore. The handoff still includes the shortlist, descriptions only where they distinguish, and a brief rationale.
- [ ] Flag ON: send an ambiguous prompt (e.g. "show me revenue" with multiple revenue-bearing explores). Subagent returns `ambiguous` with ≥2 candidates and a `suggestedQuestion`, does NOT call `findFields`.
- [ ] Flag ON: send a prompt that mentions an entity name where both base and joined tables expose a similar field (e.g. customer name with `orders.customer_name` and `customers.name` both present). Subagent prefers the joined-table field.
- [ ] Flag OFF: confirm legacy prompt rendering is byte-identical (`systemV2TemplateLegacy.ts` not touched).

### DIFF
--- a/packages/backend/src/ee/services/ai/agents/discoverFields/systemPrompt.ts
+++ b/packages/backend/src/ee/services/ai/agents/discoverFields/systemPrompt.ts
@@ -14,81 +14,13 @@ Your sole job is to take the user's question and return a structured handoff des
 
 ## How you return your result
 
-Your run must finish with a single \`submitResult\` call. Don't write prose explaining the answer; the parent agent reads the \`submitResult\` arguments verbatim. The handoff payload has exactly this shape:
+Finish with a single \`submitResult\` call. Don't write prose explaining the answer; the parent reads the call's arguments verbatim. The argument schema is the source of truth — what follows is the *when* and *what to populate*, not the *shape*.
 
-\`\`\`
-{
-  "handoff": {
-    "status": "resolved" | "ambiguous" | "no_match",
-    ...status-specific fields described below
-  }
-}
-\`\`\`
+Pick exactly one status:
 
-## Status payloads
-
-You must pick exactly one of three statuses.
-
-### 1. "resolved" — exactly one explore is the right answer
-
-Call \`submitResult\` with:
-
-\`\`\`
-{
-  "handoff": {
-    "status": "resolved",
-    "explore": {
-      "name": "orders",
-      "label": "Orders",
-      "baseTable": "orders",
-      "joinedTables": ["customers", "payments"]
-    },
-    "fields": [
-      {
-        "fieldId": "orders_total_revenue",
-        "name": "total_revenue",
-        "label": "Total Revenue",
-        "table": "orders",
-        "fieldType": "metric",
-        "fieldValueType": "number",
-        "fieldFilterType": "number",
-        "isFromJoinedTable": false,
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#23117** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23134** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23134?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #21130, opened 2026-03-13
Deleted dbt model persists in dashboard filter dropdowns


torimwhaley:
## Summary

When a dbt model is deleted from the semantic layer, it correctly disappears from the explore/table list for building new charts, but it **still appears in existing dashboard filter dimension dropdowns**. Multiple project recompiles/redeploys do not clear the stale reference. The validation system (bell icon) does not flag this, so there is no way to fix it through the UI.

## Root Cause Analysis

Three issues contribute to this bug:

### 1. Validation gap — `TableDoesNotExist` errors never generated

In `packages/backend/src/services/ValidationService/ValidationService.ts` (line ~687-696), `validateDashboards()` only checks if `fieldId` exists in the flat `existingFieldIds` array. It does **not** validate that the `tableName` is a valid explore.

The infrastructure already exists but was never wired up:
- `DashboardFilterValidationErrorType.TableDoesNotExist` is defined in `packages/common/src/types/validation.ts`
- `ValidationModel.parseDashboardFilterError()` has a parser for "Table 'X' no longer exists" errors
- But `validateDashboards()` never generates these errors

### 2. Available Filters API returns stale fields

The filter dropdown is populated by `POST /dashboards/availableFilters` → `ProjectService.getAvailableFiltersForSavedQueries()`. This:
1. Gets all saved charts on the dashboard and their `tableName`
2. Calls `findExploresFromCache()` with those table names
3. Extracts filterable dimensions from each explore

If the explore is still in `cached_explores` for this project after recompile, it will keep appearing. `saveExploresToCache()` does a full delete-and-replace, so either:
- The dbt compile still produces this explore (e.g., it's a joined model referenced by another explore)
- The project refresh didn't fully succeed
- The in-memory `exploreCache` wasn't invalidated

### 3. No cleanup of saved filter configurations

Dashboard filters stored as JSONB in `dashboard_views.filters` are never cleaned up when an explore is deleted.

## Expected Behavior

1. Validation (bell icon) should flag dashboard filters that reference deleted tables/explores
2. The filter dimension dropdown should not show fields from deleted models
3. Ideally, stale filter references should surface a clear error or be automatically cleaned up

## Steps to Reproduce

1. Create a dashboard with charts from model X
2. Add a dashboard filter using a dimension from model X
3. Delete model X from the dbt project
4. Recompile/redeploy the project
5. Open the dashboard — the deleted model's dimensions still appear in the filter dropdown
[...]

PR #23137, opened 2026-05-15
feat(observability): re-land wide-event logs for dashboard loads, project compiles, and validation runs

charliedowler:
Related to: PROD-5931 (https://github.com/lightdash/lightdash/issues/21130)

Re-lands #23086 (which was reverted in #23118 while we investigated an OOM crashloop on the analytics instance). The OOM was subsequently traced to an unrelated 219× SSE-stream blowup in the `discoverFields` AI subagent (fixed in #23126, tripwire tracked in ZAP-398) — this PR is not implicated.

## What's in here

Three structured `Logger.info` wide events to help diagnose stale dashboard filter references in production without shipping a behavioural fix:

- **`dashboard.loaded`** (`DashboardService.getByIdOrSlug`) — fires on every dashboard render. Walks top-level filter targets and per-tile `tileTargets`, surfaces any field/table pointing at an explore no longer in `cached_explore`.
- **`compile.completed`** (`ProjectService.saveExploresToCacheAndIndexCatalog`) — fires on every project compile. Reports the diff of explore names vs the previously cached set; lets us correlate "table removed at T1" with "dashboard loaded with stale reference at T2".
- **`validation.dashboardScanned`** (`ValidationService.validateDashboards`) — fires once per dashboard validated. Breaks down filter rules, tile targets, error categories, and silent-suppression buckets. Helps diagnose why some dashboards have known-broken filter state but no validation errors recorded.

All three events are wrapped in try/catch so a logging failure can never affect the underlying request.

## Test plan

- [x] `pnpm -F backend lint`
- [x] `npx jest DashboardService` (22 pass)
- [x] `npx jest ProjectService` (59 pass)
- [x] `npx jest ValidationService` (21 pass)
- [x] Verified all three events fire on the local debug-local instance with the right shape and stale-reference detection (see prior PR #23086 description for sample payloads)
- [ ] **Perf testing under analytics-scale load before merge** — see internal notes for specific concerns

## Performance concerns to address before merge

A perf-testing follow-up will stress-test the validation path under analytics-scale data shape (250 explores × 25k field IDs × many dashboards). Specific hot-spots flagged for measurement: `existingFieldIds.includes(...)` is O(N) inside a per-tile-target loop; `dashboard.loaded` adds two DB queries per render. Mitigations on standby if needed: convert the includes() check to a `Set<string>`, gate by feature flag, and/or skip the event above a payload-size threshold.

### DIFF
--- a/packages/backend/src/models/ProjectModel/ProjectModel.ts
+++ b/packages/backend/src/models/ProjectModel/ProjectModel.ts
@@ -1263,6 +1263,13 @@ export class ProjectModel {
         );
     }
 
+    async getCachedExploreNames(projectUuid: string): Promise<string[]> {
+        const rows = await this.database(CachedExploreTableName)
+            .select<{ name: string }[]>('name')
+            .where('project_uuid', projectUuid);
+        return rows.map((row) => row.name);
+    }
+
     async findVirtualViewsFromCache(
         projectUuid: string,
     ): Promise<Record<string, Explore | ExploreError>> {
--- a/packages/backend/src/services/DashboardService/DashboardService.contentVerification.test.ts
+++ b/packages/backend/src/services/DashboardService/DashboardService.contentVerification.test.ts
@@ -95,9 +95,13 @@ describe('DashboardService - Content Verification', () => {
         pinnedListModel: {} as unknown as PinnedListModel,
         schedulerModel: {} as unknown as SchedulerModel,
         schedulerService: {} as unknown as SchedulerService,
-        savedChartModel: {} as unknown as SavedChartModel,
+        savedChartModel: {
+            getInfoForAvailableFilters: jest.fn(async () => []),
+        } as unknown as SavedChartModel,
         savedChartService: {} as unknown as SavedChartService,
-        projectModel: {} as unknown as ProjectModel,
+        projectModel: {
+            getCachedExploreNames: jest.fn(async () => []),
+        } as unknown as ProjectModel,
         slackClient: {} as unknown as SlackClient,
         schedulerClient: {} as unknown as SchedulerClient,
         catalogModel: {} as unknown as CatalogModel,
--- a/packages/backend/src/services/DashboardService/DashboardService.test.ts
+++ b/packages/backend/src/services/DashboardService/DashboardService.test.ts
@@ -78,6 +78,11 @@ const savedChartModel = {
         uuid: 'chart_uuid',
         projectUuid: 'project_uuid',
     })),
+    getInfoForAvailableFilters: jest.fn(async () => []),
+};
+
+const projectModel = {
+    getCachedExploreNames: jest.fn(async () => []),
 };
 
 const schedulerModel = {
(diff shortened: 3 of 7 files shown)


charliedowler:
## Perf-test follow-up: applied Set-conversion mitigation

Validated the three concerns from the perf brief against analytics-scale dimensions and pushed a follow-up commit (08032b6) that converts `existingFieldIds` to a `Set<string>` once per `validateDashboards` call.

### Concern A — validation O(N) lookup ❌ → ✅

Microbench at the production-stated dimensions (25,466 field IDs · 150 dashboards · 16/80 tile targets · 3 inner-loop passes per dashboard):

| Scenario | Before (Array.includes) | After (Set.has) | Speedup |
|---|---|---|---|
| 150 × p95 (16 tile targets) | p50 **1,513 ms** · p95 1,613 ms | p50 **1.3 ms** · p95 3.0 ms | **~1,150×** |
| 150 × max (80 tile targets) | p50 **5,089 ms** · p95 5,215 ms | p50 **2.4 ms** · p95 5.3 ms | **~2,000×** |

The single Set is threaded through both `containsFieldId` and the new wide-event diagnostic, so the existing filter/tile-target validation also gets the speedup — not just the new code. Memory cost: ~10 MB for a 25k Set, well within the 10% budget.

All 21 ValidationService unit tests pass; backend typechecks clean; live validation run against the jaffle demo emits 9 `validation.dashboardScanned` events with no warnings.

### Concern B — `dashboard.loaded` extra DB queries ✅

Two new queries per render, both well-indexed:
- `getCachedExploreNames`: index scan on `cached_explore_project_uuid_index` — **0.036 ms** on 62 rows; ~0.15 ms extrapolated to analytics's 262.
- `getInfoForAvailableFilters`: 5-table join with correlated subquery per chart UUID — **0.23 ms** for 5 chart UUIDs in jaffle.

End-to-end dashboard render comparison (19-tile dashboard, baseline = same branch with the three files reverted to main):

| Workload | Baseline | PR | Delta |
|---|---|---|---|
| Seq 50 | p50 78.5 ms · p95 92.7 ms · mean 79 ms · stdev 6.9 ms | p50 80.7 ms · p95 115 ms · mean 87 ms · stdev 33 ms | mean +10 %, p95 +24 %, **5× wider stdev** |
| 10× concurrent | p50 134 · p95 200 · p99 242 ms | p50 134 · p95 198 · p99 256 ms | within noise |
| 20× concurrent | p50 206 · p95 436 · p99 539 ms | p50 200 · p95 451 · p99 515 ms | within noise |

DB connection pool was not saturated at 20×. At analytics scale (262 explores, 40+ tile UUIDs/dashboard) expect ~5–20 ms p95 overhead — acceptable for the internal instance's 21 events/day. Worth watching if this ever ships to a busier customer.

### Concern C — log payload size ✅

JSON-serialized bytes (worst case = all caps hit):

| Event | Typical | Realistic | Worst case |
|---|---|---|---|
| `validation.dashboardScanned` | 471 B | 1.1 KB | **1.5 KB** |
| `dashboard.loaded` | 507 B | 1.6 KB | **4.5 KB** |
| `compile.completed` | 389 B | 0.6 KB | **3.6 KB** |

All under the 5 KB p99 target.

### Concern D — warn-on-failure visibility ⚠️

Confirmed via code-read + log inspection: `logger.warn` routes through Winston (level=1, above `info=2`) to console + file transports unconditionally. **Sentry does not alert on `warn`** — it only raises issues at `error` and above. The warning strings will show up in raw logs but won't page anyone.

Recommendation: monitor for the literal strings `validation.dashboardScanned log failed`, `dashboard.loaded log failed`, `compile.completed log failed` in production log search so a silent DB-blip → suppressed-event pattern is noticed.

### Verdict

All four acceptance criteria now met. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #23191, opened 2026-05-17
fix(bigquery): add drive oauth scopes for BigQuery external tables

### LINKED ISSUE #19636: SQL Runner cannot query BigQuery external tables backed by Google Sheets
### Description

Hello Lightdash Team! 👋
I’ve been enjoying using Lightdash and wanted to report a specific issue I encountered while working with BigQuery external tables.

**Issue summary:**
SQL Runner fails to fetch results from BigQuery external tables (Google Sheets) with a "Permission denied while getting Drive credentials" error. This occurs because the BigQuery client lacks the necessary OAuth scopes, even when the Service Account has direct file access to the Google Sheet.

Image

1. Create a BigQuery external table sourced from a Google Sheet.
2. Share the Google Sheet with the Service Account email.
3. Connect Lightdash to BigQuery using the Service Account JSON.
4. Run a query against the external table in SQL Runner and observe the access denied error.

**Expected Behavior:**
SQL Runner should successfully retrieve data from external tables if the Service Account has direct access to the source file.

**Suggested improvements:**
Update the BigQuery client initialization in the backend to include drive and spreadsheets scopes.

### version

v0.1999.0

### Cloud or self-hosting

self-hosting

gracia10:
Closes: #19636

### Description

- Querying BigQuery external tables backed by Google Sheets/Drive can fail with `Permission denied while getting Drive credentials`.
- [BigQuery docs](https://cloud.google.com/bigquery/docs/query-drive-data) confirm that `drive.readonly` OAuth scope is required for Drive-backed external tables.
- This PR adds `AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE` ENV variable to conditionally include `drive.readonly` across all BigQuery auth paths.
- Defaults to off; self-hosted operators using their own OAuth app can enable it during rollout.

### Changes

| Auth path | Effect when `AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE=true` |
|---|---|
| `/login/bigquery` | Adds `drive.readonly` to OAuth consent |
| `/login/google` | Adds `drive.readonly` only when `AUTH_GOOGLE_INCLUDE_BIGQUERY_SCOPE=true` |
| Service Account / ADC | Adds `drive.readonly` to BigQuery SDK `scopes` option |

#### Note
- drive.readonly only works if the Sheets/Drive file is shared with the Google identity used for the query.
- Existing OAuth users authenticated without Drive scope will need to re-authenticate after enabling the flag.

### Test plan

> `FLAG` = `AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE`

- [x] `pnpm -F @lightdash/warehouses test`
- [x] `pnpm -F @lightdash/warehouses typecheck`
- [x] `pnpm -F @lightdash/warehouses lint`
- [x] `pnpm -F backend test`
- [x] `pnpm -F backend lint`
- [x] FLAG unset: no Drive scope in OAuth consent (`/login/bigquery`, `/login/google`)
- [x] FLAG=true: SA-based external table query succeeds (no `Permission denied`)
- [x] FLAG=true + `AUTH_GOOGLE_INCLUDE_BIGQUERY_SCOPE=true`: both OAuth paths show BigQuery + Drive scopes
- [x] FLAG=true + `AUTH_GOOGLE_INCLUDE_BIGQUERY_SCOPE=false`: `/login/google` does NOT add Drive scope alone

#### Screenshots

- External table configuration

  External table configuration

- Google OAuth consent screen for BigQuery connection (with `AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE=true`)
  BigQuery OAuth Connection screen

- SQL Runner query result from Google Sheets external table

  SQL Runner query result

### DIFF
--- a/packages/backend/src/config/lightdashConfig.mock.ts
+++ b/packages/backend/src/config/lightdashConfig.mock.ts
@@ -26,6 +26,7 @@ export const lightdashConfigMock: LightdashConfig = {
             enableGCloudADC: false,
             enabled: false,
             includeBigqueryScope: false,
+            enableBigqueryDriveScope: false,
         },
         okta: {
             loginPath: '',
--- a/packages/backend/src/config/parseConfig.test.ts
+++ b/packages/backend/src/config/parseConfig.test.ts
@@ -799,6 +799,19 @@ test('should set groups.enabled only when the environment variable is set', () =
     expect(falseConfig.groups.enabled).toBe(false);
 });
 
+test('should enable BigQuery Drive scope only when AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE is true', () => {
+    const undefinedConfig = parseConfig();
+    expect(undefinedConfig.auth.google.enableBigqueryDriveScope).toBe(false);
+
+    process.env.AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE = 'true';
+    const trueConfig = parseConfig();
+    expect(trueConfig.auth.google.enableBigqueryDriveScope).toBe(true);
+
+    process.env.AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE = 'false';
+    const falseConfig = parseConfig();
+    expect(falseConfig.auth.google.enableBigqueryDriveScope).toBe(false);
+});
+
 describe('getMultiProjectSetupConfig', () => {
     beforeEach(() => {
         delete process.env.LD_SETUP_PROJECTS;
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -1434,6 +1434,7 @@ export type AuthGoogleConfig = {
      * See PROD-7783.
      */
     includeBigqueryScope: boolean;
+    enableBigqueryDriveScope: boolean;
 };
 
 type AuthOktaConfig = {
@@ -1863,6 +1864,8 @@ export const parseConfig = (): LightdashConfig => {
                 enableGCloudADC: process.env.AUTH_ENABLE_GCLOUD_ADC === 'true',
                 includeBigqueryScope:
                     process.env.AUTH_GOOGLE_INCLUDE_BIGQUERY_SCOPE === 'true',
+                enableBigqueryDriveScope:
+                    process.env.AUTH_ENABLE_BIGQUERY_DRIVE_SCOPE === 'true',
             },
             okta: {
                 oauth2Issuer: process.env.AUTH_OKTA_OAUTH_ISSUER,
--- a/packages/backend/src/routers/apiV1Router.ts
+++ b/packages/backend/src/routers/apiV1Router.ts
@@ -428,10 +428,14 @@ apiV1Router.get(
     lightdashConfig.auth.google.loginPath,
     storeOIDCRedirect,
     (req, res, next) => {
-        const { includeBigqueryScope } = lightdashConfig.auth.google;
+        const { includeBigqueryScope, enableBigqueryDriveScope } =
+            lightdashConfig.auth.google;
         const scope = ['profile', 'email'];
         if (includeBigqueryScope) {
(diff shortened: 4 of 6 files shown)


rephus:
Hi @gracia10 first of all, thanks for your contribution, and creating an issue. 

While your solution seem good to solve this particular usecase. Most Lightdash users will be concerned about giving `drive` (even read only) access when using Bigquery. If we still want to keep this approach, we should make it conditional (for example, behind an ENV variable). 

This morning I added a similar change, on login, based on an ENV variable. https://github.com/lightdash/lightdash/pull/23330

I believe there is a workaround for your use case, if you authenticate with "google" for using `upload go toogle sheets` it will let you authenticate with `google` to give you the drive access. And since we have `includeGrantedScopes` , your token will include both `bigquery` and `drive`. Could you test this workaround to see if it fixes your particular usecase ? (we only grant drive.file, so I'm not sure if it will work, see another comment below.) 

CleanShot 2026-05-20 at 10 45 57

rephus:
Also, I found another blocker for this. 

We use on gdrive the scope `drive.file` which gives access to files used by the app. This scope is non sensitive. 

However, `drive.readonly` is a `sensitive` scope, we will have to reauthorize our existing Google application (which takes weeks) just to include this scope, which is a big blocker for us. They require Google’s mandatory verification process before the app can be used publicly. So if we still want to take this approach, even behind an ENV variable in the cloud, it will take some time until we get this live. 

Adding this scope without being verified by Google it will show you `security` warnings when you try to authenticate.

gracia10:
Hi @rephus , thanks for the feedback!

I cleared my Google permissions and re-tested the incremental auth flow:

1. Re-authenticated BigQuery (BQ scope only)
2. Exported to Google Sheets (`drive.file` granted)
3. Re-authenticated BigQuery again (BQ + existing Drive export scope active)
4. Ran the SQL Runner query → still hits `Permission denied while getting Drive credentials`

It turns out the existing `/login/gdrive` + `includeGrantedScopes` path does not grant the specific Drive access BigQuery needs for Drive-backed external tables. Furthermore, this approach cannot support service account-based connections, as those queries bypass user OAuth tokens entirely.

You're right that `drive.readonly` being a sensitive scope is a valid concern. 
To avoid changing default Lightdash scopes or forcing Cloud through re-verification, we could introduce a narrow, default-off flag: `AUTH_GOOGLE_BIGQUERY_DRIVE_SCOPE_ENABLED=true`.

Happy to update the PR if this direction works for you.

rephus:
> we could introduce a narrow, default-off flag: AUTH_GOOGLE_BIGQUERY_DRIVE_SCOPE_ENABLED=true.

Yes @gracia10  that seems like a good approach, this way we can keep it off in the cloud but default, enabled it on some customers temporarily while we do Google verification, and selfhosters can enable it using their own verified tokens.

gracia10:
**Implementation note following the rollout-flag direction discussed above:**

I intentionally kept `drive.readonly` scope injection outside persisted `CreateBigqueryCredentials` state so the rollout flag can be removed cleanly later without DB cleanup or migration concerns.

I also considered threading this through warehouse client creation options instead, but the backend and CLI currently do not share the same parsed config path. Supporting both call stacks would require additional config propagation and parsing logic across multiple layers.

Longer term, once Google verification is complete and the rollout flag is no longer needed, promoting `drive.readonly` into `CreateBigqueryCredentials.scopes` as a first-class field would probably be the cleaner direction.

gracia10:
Hi @rephus — updated the PR following the rollout-flag direction discussed above and marked it ready for review again. Would appreciate another look when you have a chance.

rephus:
Hi @gracia10 , sorry for the long delay, I missed the update among all the other PRs that were created. Your final approach looks good to me, unfortunately, it's been a long time, so there were some file conflicts. So I'll close this PR for now, but feel free to create a new PR or reopen this one with a rebased version if you are interested on this change.

PR #23208, opened 2026-05-18
chore(ai): prefer admin-verified content in findContent and getDashboardCharts

joaoviana:
# AI agent — prefer admin-verified content (Stage A)

## Summary

When the in-product AI agent searches for charts or dashboards, results that admins have verified are now ranked first and visibly marked. The agent's system prompt teaches it to surface those verified items, name the verifier and date in its response, and not apologize when only unverified items match.

This is the first of three planned stages. Stage B (embedding-based RAG of verified chart configs into context) and Stage C (verified-content-aware ambiguity resolution in `discoverFields`) are deferred to follow-up PRs.

## What changes for the agent

- `findContent` and `getDashboardCharts` now sort verified items first within each search query's result set. Within each bucket (verified / unverified), the existing `search_rank` order is preserved — relevance still wins inside a bucket.
- Every verified chart and dashboard is rendered with a `<verified by="Sarah Khan" at="2 weeks ago" />` element. Inner-dashboard chart previews (the small chart list shown when `findContent` returns a dashboard) get the same treatment when an individual chart is independently verified.
- The system prompt has a new short section teaching the agent what `<verified>` means and how to use it.

## What does NOT change

- Human-facing search ranking (the UI's search box). The sort lives in the AI tool's XML rendering layer only.
- Verification semantics. Verification stays atomic — verifying a dashboard does NOT implicitly verify its constituent charts. Charts that are NOT independently verified will not be marked, even if they live on a verified dashboard.
- SQL charts (`saved_sql`). They aren't verifiable at the product level today; they simply won't get the badge.
- The `findContent` tool's input schema. No new parameters, no new modes.

## Why tool-time ranking, not embedding-based RAG

The original ticket framing (PROD-7128) proposed extending the `getVerifiedArtifactsForAgent` embedding pipeline to also pull from `content_verification`, embed those rows, and inject by similarity at conversation start. That works, but Stage A goes a different route because:

1. **The data is already plumbed.** `SearchModel` already attaches `verification: ContentVerificationInfo | null` to every chart and dashboard search result. The current `findContent` rendering drops that information on the floor. Surfacing it costs ~20 lines of code.
2. **Tool-time signals fire every search, not just once.** A user who pivots topic mid-conversation triggers a fresh `findContent`; the verified signal is there each time. RAG-injection only fires at prompt start.
3. **It's the cheapest way to validate the hypothesis.** If telemetry shows the agent ignores the signal even with the sort + annotation + prompt, that's diagnostic — we know the embedding pipeline alone wouldn't have helped either.
4. **No backfill problem.** Embedding admin-verified content would require deciding what to embed (name+description? plus field list?) and a one-shot backfill for existing verified items. Tool-time ranking reads the live data each time.

The deeper RAG path (Stage B) remains in scope for later if Stage A's telemetry shows the agent benefits from seeing verified chart **configurations** verbatim (e.g. to copy filter conventions when generating a new chart). The tickets stay open.

## System prompt addition

A new `## Verified content` section, positioned right after `## Tool workflow` so it lives near the `findContent` guidance. Roughly:

> Some content returned by `findContent` and `getDashboardCharts` is marked with a `<verified by="..." at="..." />` element. Verified items have been explicitly approved by an organization or project admin as canonical, trustworthy content. When you find a verified item that matches the user's question:
> - Prefer it over unverified alternatives and surface it first in your response.
> - Mention in your response that it's verified and who verified it (e.g. "this is a verified dashboard, approved by Sarah Khan 2 weeks ago").
> - If only unverified items match, use them normally — don't apologize for the lack of verification.

The line about "don't apologize" is what handles projects with zero verified content — no dynamic prompt injection is needed.

## Rollout

- **No feature flag and no config kill-switch.** This is an additive UX improvement; the worst case is the agent ignores the signal, not generates wrong answers.
- **Default-on, project-aware behavior.** Projects with zero verified items see the same agent behavior they see today (no `<verified>` elements appear in results, prompt guidance falls through to the "don't apologize" path).
- **Coverage telemetry only.** Per `findContent` call we record:
  - `verifiedResultCount` (how many of the returned items were verified)
  - `totalResultCount`
  - `topResultVerified` (boolean — was the top-ranked result verified?)
  Aggregated by project + agent. Tells us per-customer whether the signal has any chance to fire. We intentionally do NOT track whether the agent "picked up" the verified item in its response (substring matching for "verified by" is fragile and noisy).

## Regression analysis

- **Verified-content feature itself**: unaffected. Verification reads still go through the same `ContentVerificationModel` path. No new permissions checks added in the agent path because `SearchModel` already surfaces verification info to anyone with access to the underlying content.
- **Non-AI search**: unaffected. The sort lives inside the AI tool's rendering layer, not in `SearchModel`.
- **Service accounts / custom roles**: no permissions changes, just rendering. A service account that can already see a chart will see its verification status in `findContent` results, same as a human user would. No org-level / cross-project leakage — verification is project-scoped via the existing `content_verification` table.
- **Projects with no verified content**: behave identically to today. Empty-state is handled by the prompt instruction "don't apologize for the lack of verification."
- **MCP `find_content` (external Claude Code / IDE users)**: unchanged in this PR. PROD-7128's "Theme 1.1" follow-up to extend the same signal to the external MCP tool is filed separately.

## Tickets

- PROD-7128 stays open — solved-different-way for the in-product agent (tool-time ranking); embedding-based RAG path deferred to Stage B.
- PROD-7130 stays open — find/recommend portion done; "use verified content as templates when generating new charts" portion deferred to Stage B.

## Test plan

- [ ] Unit: `findContent.test.ts` — verified item sorts first; `<verified>` element renders with correct `by` / `at` attributes; no element when verification is null; stable sort preserves `search_rank` within each bucket.
- [ ] Manual: in the seeded dev DB, verify a chart and a dashboard via the UI, then ask the AI agent a question that matches them. Confirm: (a) agent surfaces them first, (b) agent's response cites verification status, (c) projects without verified content behave normally.
- [ ] Telemetry: confirm `verifiedResultCount` / `totalResultCount` / `topResultVerified` appear in the `findContent` tool-call analytics event.

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -1915,6 +1915,23 @@ export type AiAgentArtifactsRetrievedEvent = BaseTrack & {
     };
 };
 
+export type AiAgentFindContentCoverageEvent = BaseTrack & {
+    event: 'ai_agent.find_content_coverage';
+    userId: string;
+    properties: {
+        organizationId: string;
+        projectId: string;
+        aiAgentId: string;
+        agentName: string;
+        threadId: string;
+        promptId: string;
+        searchQuery: string;
+        totalResultCount: number;
+        verifiedResultCount: number;
+        topResultVerified: boolean;
+    };
+};
+
 export type SchedulerOwnershipReassignedEvent = BaseTrack & {
     event: 'scheduler.ownership_reassigned';
     properties: {
@@ -2052,6 +2069,7 @@ type TypedEvent =
     | AiAgentToolCallEvent
     | AiAgentArtifactVersionVerifiedEvent
     | AiAgentArtifactsRetrievedEvent
+    | AiAgentFindContentCoverageEvent
     | ContentVerificationEvent
     | SchedulerOwnershipReassignedEvent
     | ImpersonationEvent;
--- a/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
+++ b/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
@@ -90,6 +90,7 @@ import {
     AiAgentEvalAppendedEvent,
     AiAgentEvalCreatedEvent,
     AiAgentEvalRunEvent,
+    AiAgentFindContentCoverageEvent,
     AiAgentPromptCreatedEvent,
     AiAgentPromptFeedbackEvent,
     AiAgentResponseStreamed,
@@ -3975,7 +3976,10 @@ Use your existing tools to inspect them when relevant to the user's question. Wh
                 update: UpdateSlackResponse | UpdateWebAppResponse,
             ) => this.aiAgentModel.updateModelResponse(update),
             trackEvent: (
-                event: AiAgentResponseStreamed | AiAgentToolCallEvent,
+                event:
+                    | AiAgentResponseStreamed
+                    | AiAgentToolCallEvent
+                    | AiAgentFindContentCoverageEvent,
             ) => this.analytics.track(event),
 
             createOrUpdateArtifact: async (data) => {
--- a/packages/backend/src/ee/services/McpService/McpService.ts
+++ b/packages/backend/src/ee/services/McpService/McpService.ts
@@ -580,6 +580,7 @@ export class McpService extends BaseService {
                 const findContentTool = getFindContent({
                     findContent,
                     siteUrl: this.lightdashConfig.siteUrl,
+                    trackCoverage: () => {},
                 });
                 const result = await findContentTool.execute!(argsWithProject, {
                     toolCallId: '',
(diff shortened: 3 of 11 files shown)


joaoviana:
* **#23211** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23208** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23211, opened 2026-05-18
feat(ai): teach discoverFields subagent to prefer verified-chart fields

joaoviana:
# AI agent — verified-content tiebreaker in field discovery (Stage C)

## Summary

Teach the `discoverFields` subagent to prefer fields and explores that admins have endorsed via the existing content verification feature. When the subagent is breaking a tie between explores, or shortlisting fields for the parent agent, it now sees per-field `usageInVerifiedCharts` alongside the existing `usageInCharts` count — and the system prompt instructs it to use the verified count as a sharper tiebreaker (subordinate to agent-specific instructions).

This is the second of three planned stages. Stage A (verified-first ranking + `<verified>` marker in `findContent` / `getDashboardCharts`) shipped in #23208. Stage B (embedding-based RAG of verified chart configs into the generation context) is still deferred.

## What changes for the agent

- Every `<field>` rendered by `findExplores` and `findFields` now carries an additional `usageInVerifiedCharts={N}` attribute alongside `usageInCharts`. `N` is the number of admin-verified charts in the project that use this field.
- `discoverFields/systemPrompt.ts` Step 3 (explore ambiguity check) now uses `usageInVerifiedCharts` as a sharper tiebreaker before falling through to raw chart usage.
- `discoverFields/systemPrompt.ts` Step 4 (field shortlisting) instructs the subagent to prefer fields with non-zero `usageInVerifiedCharts` between equally-relevant alternatives — explicitly as a tiebreaker, not a hard filter, so a clearly more relevant field with `verifiedChartUsage=0` still wins over an irrelevant verified one.
- Agent-specific instructions (`{{agent_instruction}}` block at the prompt tail) keep their existing precedence by virtue of positional recency. No "OVERRIDE" prepend is added.

## What does NOT change

- Schema. **No migration, no new column on `catalog_search`, no hooks in verify/unverify/chart-save.** The signal is computed live at agent-call time via a single grouped query.
- The Spotlight / catalog UI. Verification metadata exists on `catalog_search` results no differently than before; only the AI agent's service-layer wrapper enriches with the new count.
- Verification semantics. Still atomic — a verified dashboard's constituent charts don't get counted unless they themselves are verified.

## Implementation

- **New query: `ContentVerificationModel.getVerifiedFieldUsage(projectUuid)`** — returns `Map<"${fieldId}::${fieldType}", number>` keyed by the chart-field-usage table's natural key (`saved_queries_version_fields.name` is already the fieldId format `${table}_${field}`). One GROUP BY on `content_verification → saved_queries → latest saved_queries_version → saved_queries_version_fields` filtered to `content_type=CHART` and soft-delete-aware.
- **Enrichment in the AI service-layer wrappers** (`AiAgentService.findExplores`, `AiAgentService.findFields`). Both fetch the map lazily via a closure-scoped memo so the same agent run pays for it at most once, and attach `verifiedChartUsage` to each catalog item before returning.
- **Type surface**: `CatalogField.verifiedChartUsage?: number` in `@lightdash/common` (optional, since non-AI callers of `CatalogModel.search` don't populate it). Mirror field added to Zod metadata schemas (`toolFindExploresArgs`, `toolFindFieldsArgs`).
- **XML rendering**: `findExplores.tsx` and `findFields.tsx` add `usageInVerifiedCharts={N}` next to the existing `usageInCharts={N}` attribute on every `<field>` element.

## Why not a denormalized `verified_chart_usage` column on `catalog_search`

We considered it. Trade-off table:

|                            | Denormalize column                                                | Enrich at query time (chosen)                          |
|----------------------------|-------------------------------------------------------------------|--------------------------------------------------------|
| Migration + backfill       | Required                                                          | None                                                   |
| Hooks on verify/unverify/chart-save | 3+ services touched, drift risk                          | Not needed                                             |
| Read cost per agent call   | Pure SELECT (no extra JOIN)                                       | One extra grouped query (memoized per agent run)       |
| Available to non-AI surfaces | Yes (Spotlight, etc.) — broader long-term utility               | No — strictly AI service-layer                         |

We went with enrichment because the read cost is bounded (verified-chart corpus is small in practice), there's zero drift risk, and the data path matches how `SearchModel` already enriches per-result verification info for `findContent` (Stage A). If broader UI demand materialises, promoting to a denormalized column becomes a focused follow-up PR.

## Regression analysis

- **Existing callers of `CatalogModel.search`**: unaffected. `verifiedChartUsage` is optional on `CatalogField` and only populated by the AI service-layer wrappers. Spotlight UI, Catalog API endpoints, and any other consumer see the same shape they got before.
- **Existing callers of `FindExploresFn` / `FindFieldFn`**: the new field is optional, so existing tests / mocks need no update. The XML rendering simply emits `usageInVerifiedCharts="0"` when verification data is absent.
- **Subagent in projects with zero verified content**: the prompt explicitly says "tiebreaker, not hard filter" — when `usageInVerifiedCharts` is `0` everywhere, the existing Step 3/4 logic (joined tables, raw `usageInCharts`, search rank) handles the decision as before.
- **Soft-deleted charts**: filtered out in `getVerifiedFieldUsage` via `whereNull(saved_queries.deleted_at)`.
- **Performance**: one additional grouped query per `AiAgent.findExplores` / `AiAgent.findFields` call (memoized per agent run). Verified-chart corpus is bounded (typically <100 items per project), so the JOIN and GROUP BY are cheap. Wrapped in a Sentry transaction for visibility.

## Tickets

Stage C is the final piece of:

- **PROD-7128** ("Include admin-verified content in in-product AI retrieval") — together with Stage A (#23208), admin-verified content is now visible to the agent at every retrieval surface: `findContent`, `getDashboardCharts`, `findExplores`, `findFields`. Recommend closing on merge of both PRs.
- **PROD-7130** ("Teach the in-product AI agent to reference verified content in its system prompt") — together with Stage A's parent-agent prompt update, the subagent's prompt now also treats verified content as canonical at the ambiguity-tiebreaker and field-shortlisting decision points. Recommend closing on merge of both PRs.

Stage B (embedding-based RAG of verified chart configs into the generation context, to let the agent copy verified patterns when building new charts) remains a potential future option, gated on whether telemetry from Stage A's `ai_agent.find_content_coverage` event suggests the tool-time signal isn't enough.

## Test plan

- [ ] Manual: in the seeded dev DB, verify a chart in two competing explores, ask a vague data question that could plausibly hit either explore, confirm the subagent picks the explore with verified usage.
- [ ] Manual: in a project with zero verified charts, confirm the subagent behaves identically to before (Step 3/4 falls through to existing tiebreakers).
- [ ] Manual: verify a chart with two metrics, ask for one of those metrics, confirm the resolved field shortlist preferentially includes the verified metric.
- [ ] Manual: confirm an agent with a custom instruction telling it to "always use `net_revenue`" still picks `net_revenue` even when `gross_revenue` has higher `usageInVerifiedCharts`.

### DIFF
--- a/packages/backend/src/ee/index.ts
+++ b/packages/backend/src/ee/index.ts
@@ -136,6 +136,8 @@ export async function getEnterpriseAppArguments(): Promise<EnterpriseAppArgument
                     aiAgentModel: models.getAiAgentModel(),
                     changesetModel: models.getChangesetModel(),
                     catalogModel: models.getCatalogModel(),
+                    contentVerificationModel:
+                        models.getContentVerificationModel(),
                     groupsModel: models.getGroupsModel(),
                     featureFlagService: repository.getFeatureFlagService(),
                     slackClient: clients.getSlackClient(),
--- a/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
+++ b/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
@@ -112,6 +112,7 @@ import {
     CatalogSearchContext,
 } from '../../../models/CatalogModel/CatalogModel';
 import { ChangesetModel } from '../../../models/ChangesetModel';
+import { ContentVerificationModel } from '../../../models/ContentVerificationModel';
 import { GroupsModel } from '../../../models/GroupsModel';
 import { OpenIdIdentityModel } from '../../../models/OpenIdIdentitiesModel';
 import { ProjectModel } from '../../../models/ProjectModel/ProjectModel';
@@ -213,6 +214,7 @@ type AiAgentServiceDependencies = {
     catalogService: CatalogService;
     catalogModel: CatalogModel;
     changesetModel: ChangesetModel;
+    contentVerificationModel: ContentVerificationModel;
     searchModel: SearchModel;
     featureFlagService: FeatureFlagService;
     groupsModel: GroupsModel;
@@ -270,6 +272,8 @@ export class AiAgentService extends BaseService {
     private readonly catalogModel: CatalogModel;
 
     private readonly changesetModel: ChangesetModel;
+
+    private readonly contentVerificationModel: ContentVerificationModel;
 
     private readonly featureFlagService: FeatureFlagService;
 
@@ -338,6 +342,7 @@ export class AiAgentService extends BaseService {
         this.catalogService = dependencies.catalogService;
         this.catalogModel = dependencies.catalogModel;
         this.changesetModel = dependencies.changesetModel;
+        this.contentVerificationModel = dependencies.contentVerificationModel;
         this.searchModel = dependencies.searchModel;
         this.featureFlagService = dependencies.featureFlagService;
         this.groupsModel = dependencies.groupsModel;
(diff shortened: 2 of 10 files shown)


joaoviana:
* **#23211** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23211?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23211?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23208** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #23213, opened 2026-05-18
fix(sum-distinct): update sum distinct to partition only on distinct_keys

jesshitchcock:
closes:  [SPK-450](https://linear.app/lightdash/issue/SPK-450/remove-extra-dimensions-from-sum-distinct-partitioning)

Fixes sum_distinct and average_distinct so they return the correct deduplicated total when a user selects a dimension that's unrelated to the metric's distinct_keys.                                                                                                                                                    
                                                                                                                                                                    
Before this PR, selecting a dimension that wasn't part of distinct_keys caused the metric to dedupe per dimension group instead of globally. So the "total" shown was actually a per-group sum, hiding the real deduplicated value.
                                                                                                                                                                                   
The metric now always deduplicates at the distinct_keys grain. Three cases at the join:                                                                            
  - None of the selected dims are distinct_keys → one global value, repeated on every row.                                                                               
  - Some of the selected dims are distinct_keys → group by those overlapping dims, join on them.
  - All selected dims are distinct_keys → one value per group, joined on every key.                                                                                 
                                                                                                                                                                    
Tests                                                                                                                                                             
  - New snapshot tests cover the multi-key cases (one distinct_key selected, both selected, neither selected).                                                      
  - Cypress e2e (sumDistinctWithFanout.cy.ts) updated for the new behaviour.
  - Removed duplicate inline assertions that the snapshots already cover.                                                                                           
                              
Note this PR reverses the SPK-334 behaviour of partitioning by the selected dimensions, which produced per-group sums and obscured the global deduplicated total when the selected dimension was orthogonal to the distinct_keys grain.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.mock.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.mock.ts
@@ -2390,6 +2390,18 @@ export const EXPLORE_WITH_SUM_DISTINCT: Explore = {
                     tablesReferences: ['orders'],
                     hidden: false,
                 },
+                line_item_id: {
+                    type: DimensionType.STRING,
+                    name: 'line_item_id',
+                    label: 'Line Item ID',
+                    table: 'orders',
+                    tableLabel: 'orders',
+                    fieldType: FieldType.DIMENSION,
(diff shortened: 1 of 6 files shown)


jesshitchcock:
confirmed test is passing locally (myself and with Lukas)

PR #23064, opened 2026-05-14
fix(ai-agent): defer artifact auto-open until thread streaming completes

joaoviana:
## Summary

The thread-level "open the latest artifact" effect in `useAiAgentThreadArtifact` was firing the moment a new artifact got appended to `message.artifacts` — which happens as soon as the `runQuery` / `generateDashboard` tool result lands, **before** the assistant has finished producing its closing text after the tool call. The artifact panel popped open mid-stream and yanked focus away from the text the user was reading.

**Repro:** thread `607d6f8d-def6-46b5-b828-7578226d1ebc` — the prompt ran tool calls, generated a visualization artifact, and the panel auto-opened before the final response text finished streaming.

## Fix

Gate the dispatch on `!isStreaming` for the thread. The hook now subscribes to `useAiAgentThreadStreaming(threadUuid)` and adds it to the effect's deps, so when streaming flips to false the effect re-runs and opens the artifact normally.

## Regression analysis

Unchanged behaviour for:
- **Page reload on an existing thread** — `isStreaming` is false, the latest artifact opens as before.
- **User explicitly opening an artifact mid-stream via a content link** — the existing `artifact?.messageUuid === latestAssistantMessage.uuid` early-return short-circuits before the new streaming guard.
- **"User manually closed" tracking** — the separate effect that captures `lastHandledMessageUuidRef.current` when the user clears the panel still fires on artifact-cleared transitions regardless of streaming state.

## Test plan

- [x] `pnpm -F frontend typecheck`
- [x] `pnpm -F frontend lint` (2 pre-existing warnings in unrelated files)
- [ ] Manual: send a prompt that triggers `runQuery` → verify the artifact panel **does not** open until the assistant's final markdown finishes streaming (reviewer)
- [ ] Manual: reload a thread that has a visualization → verify the latest artifact opens automatically as before (reviewer)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #23224, opened 2026-05-18
fix(ai): render artifact viz on tool result, defer only panel auto-open

joaoviana:
## Summary

Closes [ZAP-394](https://linear.app/lightdash/issue/ZAP-394/ai-agent-render-artifact-viz-on-tool-result-defer-only-the-panel-auto).

Restores the pre-#23064 "time to chart" while keeping streaming focus from being yanked when the side panel opens mid-stream.

- **Drop the `isStreaming` guard in `useAiAgentThreadArtifact`** — the panel now auto-opens (and the chart renders) the moment `runQuery` resolves and the artifact lands on `message.artifacts` via `refetchThread`. Pre-#23064 timing is back, recovering ~3s of perceived responsiveness per query.
- **Drop `!isPending` from `isArtifactAvailable`** — the artifact button is no longer gated on `responded_at` being set, so it appears in lockstep with the panel rather than waiting for stream end.
- **Show "Finishing up…" below the latest text segment** when the stream is still going, the message has at least one artifact, and the last `parts` segment isn't text. Bridges the gap between the artifact landing and the closing text streaming in — the user sees the chart immediately, with a clear cue that more text is coming, so the panel opening no longer reads as a focus-yank.
- **`TypingDots` now takes an optional `label` prop**, defaulting to `"Working on your request"` so the existing "no parts yet" call site is unchanged.

## Regression analysis

Behavior change is scoped to:

- **Streaming chart queries (`runQuery` tool)** — these get the time-to-chart benefit back, plus the new "Finishing up…" indicator. This is the only path that calls `refetchThread` on tool result, so it's the only path where artifacts land mid-stream.
- **Streaming dashboard queries (`generateDashboard`)** — no behavior change. `generateDashboard` doesn't currently trigger `refetchThread`, so artifacts only land at stream end; the panel still opens at stream end as it does today.
- **Post-load / non-streaming view** — no change. `isStreaming` is false, the hook runs once, panel opens normally. The "Finishing up…" indicator is gated on `isStreaming` so it never renders on a loaded thread.
- **Failed streams / no-artifact streams** — no change. Indicator requires `message.artifacts?.length > 0` so it only appears once at least one artifact has been produced.

The artifact button now appears slightly earlier (at tool-result time, not at stream end) for streaming `runQuery` results. The button just opens the panel — its earlier appearance has no side effects beyond the obvious one (visible earlier).

## Test plan

- [ ] Send a chart-producing prompt to an agent. Expected sequence:
    1. Preamble text streams in.
    2. Activity card shows `findExplores` → `findFields` → `runQuery`.
    3. On `runQuery` resolve: artifact button appears, panel opens with the chart, "Finishing up…" appears below the preamble in the bubble.
    4. Closing text streams in → "Finishing up…" disappears, text takes its place.
    5. Stream ends.
- [ ] Manually close the panel mid-stream → it should NOT auto-reopen for the same message (existing `lastHandledMessageUuidRef` behavior preserved).
- [ ] Reload a completed thread → panel opens normally, no "Finishing up…" indicator.
- [ ] Send a `generateDashboard` prompt → still opens at stream end, no regression there.

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AgentChatAssistantBubble.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AgentChatAssistantBubble.tsx
@@ -392,6 +392,11 @@ const AssistantBubbleContent: FC<{
                             s.kind === 'text',
                     );
                     const latestTextSeg = textSegments[textSegments.length - 1];
+                    // bridges the gap between artifact landing and closing text.
+                    const showFinishingUp =
+                        isStreaming &&
+                        !!message.artifacts?.length &&
+                        segments[segments.length - 1]?.kind !== 'text';
                     const finalAnswerMd = latestTextSeg ? (
                         <Box
                             className={`${styles.aiMarkdown} ${
@@ -491,6 +496,9 @@ const AssistantBubbleContent: FC<{
                                     {finalAnswerMd}
                                 </Box>
                             ) : null}
+                            {showFinishingUp && (
+                                <TypingDots label="Finishing up" />
+                            )}
                         </Stack>
                     );
                 }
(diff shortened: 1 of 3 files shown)


joaoviana:
* **#23224** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23224?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23224?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23232, opened 2026-05-18
feat(ai): data-aware agent suggestion chips (empty-state + post-response)

joaoviana:
## Summary

End-to-end LLM-backed agent suggestion chips for both **empty-state** (new thread) and **post-response** (after the agent has replied) modes. Chips render inside the input chrome, hide while scrolling up, and are grounded in real catalog/thread/verified-content data — no invented field names.

This rolls up three stacked WIP branches into one PR for local testing; we'll clean up the history before merging.

## What's in here

**Schema & types** (`@lightdash/common`)
- `AgentSuggestion` is a discriminated union: `kind: 'prompt'` (current data-question shape) | `kind: 'navigate'` (label + server-resolved URL).
- Model emits navigate chips with `recentConversationIndex`; the server resolves to a real thread URL post-LLM. UUIDs never reach the model.
- `validateAgentSuggestion` lives in `common/AiAgent/validators.ts` next to `getValidAiQueryLimit`. Drops chips that reference explores the user can't reach or navigate chips whose index didn't resolve.

**Backend** (`AiAgentService` + `suggestionGenerator`)
- `GET /projects/:p/aiAgents/:a/suggestions` now accepts optional `threadUuid` and `afterMessageUuid`. When present the prompt branches into post-response mode.
- Empty-state context: agent metadata, available explores, verified questions/tags, user-visible verified content, and the user's 5 most recent threads (with `threadUuid` server-only).
- Post-response context: last 6 thread messages, latest assistant text, heuristic flags (`askedClarifyingQuestion`, `refused`), and `latestQueryExplore` — the explore the agent actually queried this turn (dimensions + metrics resolved to labels). Tight scope, low hallucination surface.
- System prompt branches: empty-state allows at most one navigate chip + 3-5 prompt chips; post-response forbids navigate chips and prefers `latestQueryExplore` fields.
- Server-side validation runs every chip post-`generateObject`; drops invalid chips and falls back to `SUGGESTION_FALLBACK_CHIPS` if it kills everything.

**Frontend** (`AgentChatInput` + `AgentSuggestionChips`)
- Chips render in a soft strip directly above the input (squarish 6px corners, indigo accent on navigate chips with leading `IconArrowUpRight`).
- Prompt chip click: empty-state inserts a mention you can compose around; post-response auto-submits the label.
- Navigate chip click: react-router navigates to the resolved thread URL.
- Scroll-aware visibility: chips fade + collapse when the user scrolls more than ~40px from the bottom of the thread; reappear within 8px. Hysteresis prevents flicker from the layout shift.
- Tracking emits `chipKind` + `mode` so we can measure empty-state vs post-response engagement.

**Streamdown table breathing room**
- Small vertical margin on `.aiMarkdown table` so streamdown tables sit naturally in prose instead of butting against paragraphs.

## Regression analysis

- **System roles**: chips were previously empty-state-only behind `FeatureFlags.AiAgentSuggestions`. Behaviour for users without that flag enabled is unchanged.
- **Service accounts**: suggestions endpoint is authenticated; service accounts can hit it but won't typically (no UI surface). No new scope requirements.
- **Custom roles**: verified content fetch uses `getVerifiedContentForHomepage` which already filters by space access. Recent threads fetch uses `findThreads({userUuid: user.userUuid})` — only the user's own threads. RBAC-safe.
- **Cross-project**: `agentUuid` already scopes to a single agent + project.

## Test plan

- [ ] Empty-state (new thread for an agent with recent activity): expect 1 navigate chip + 3-5 prompt chips. Verify the navigate chip has the indigo treatment + arrow icon and that clicking it opens the named thread.
- [ ] Empty-state (agent with no recent threads): expect prompt chips only, referencing real explore fields and verified content tags.
- [ ] Post-response (after the agent has just replied): expect 2-5 prompt chips, no navigate chips. Verify field names match the agent's last query (the `latestQueryExplore`).
- [ ] Post-response after a refusal turn ("the explore doesn't have X"): chips should pivot, not repeat the refused line.
- [ ] Post-response after a clarifying-question turn: chips should be likely answers to the agent's question.
- [ ] Scroll up the thread: chips fade and collapse smoothly without flicker. Scroll back down: chips reappear.
- [ ] Click a prompt chip in empty-state: inserts as a mention in the editor.
- [ ] Click a prompt chip in post-response: auto-submits and triggers a new turn.
- [ ] Click a navigate chip: opens the linked thread via react-router (no full page reload).
- [ ] Streamdown tables (when the agent answers with a markdown table): have visible margin above + below the table.


joaoviana:
Closing in favor of the stacked PRs:

- #23267
- #23266
- #23265

This splits the original broad PR into base UX, validation/gating, and post-response streaming.

PR #23250, opened 2026-05-18
feat(backend): structured observability for scheduler & warehouse queries

charliedowler:
## Summary

Promotes scheduler/warehouse debugging signals from unstructured log text to filterable structured fields so future investigations land in one Cloud Logging query instead of cross-referencing message strings.

### What changes

1. **Tag warehouse errors with scheduler context.** `RunQueryTags` now carries `scheduler_uuid`, `scheduler_name`, `saved_sql_uuid`, and `job_id` alongside the existing `chart_uuid` / `dashboard_uuid` / `query_context`. `AsyncQueryService` reads scheduler context from a request-scoped `ExecutionContext` and spreads it onto every `queryTags` assembly, and the warehouse-error catch block now spreads the full `queryTags` so a single log line maps a generic warehouse error (e.g. ``\"Connection terminated unexpectedly\"``) back to the originating sync.

2. **Log the column-discovery phase as distinct events.** The synchronous LIMIT-1 column-discovery query in ``prepareSqlChartAsyncQueryArgs`` was silent before. It now emits ``column_discovery.started`` / ``.completed`` / ``.failed`` with ``durationMs`` / ``sqlBytes`` / ``columnCount`` plus the queryTags. The existing ``performance.now()`` timers (``warehouseMs``, ``userAttributesMs``, ``columnDiscoveryMs``, ``queryBuildingMs``, ``sqlGenerationMs``) are emitted as numeric structured fields instead of stringified-and-nested under ``sections``.

3. **Emit ``duration_ms`` as a top-level structured field on every HTTP access log.** A new winston format aliases ``express-winston``'s ``responseTime`` to ``duration_ms``, so latency dashboards and alerting can use ``jsonPayload.duration_ms > 60000`` instead of regexing the message text for ``\"… 600612 ms\"``.

4. **Correlation ID in failure emails.** ``sendGoogleSheetsErrorNotificationEmail`` and ``sendScheduledDeliveryFailureEmail`` now take a ``correlationId`` parameter and append ``?ref=<jobId>`` to the existing deep-link. Support can paste one ID into Cloud Logging and land on the failing job instead of triangulating from project UUID + approximate timestamp.

5. **Stamp every backend log with ``organization_uuid`` + ``organization_name``.** Extends the existing AsyncLocalStorage-backed ``ExecutionContext`` so a new express middleware (HTTP) and ``setSchedulerJobLogContext()`` (scheduler tasks) attach org context once, and every downstream ``Logger.info`` / ``Logger.error`` call automatically inherits it. Multi-tenant log filtering becomes ``jsonPayload.organization_name=\"acme-co\"`` instead of grepping UUIDs across pods.

**Note:** the spec mentioned ``organization_slug`` but no such field exists in this codebase — substituted ``organization_name`` (the closest human-readable equivalent).

### Why now

Motivated by a recent investigation where a generic warehouse error in the logs took significant time to map back to a specific scheduled sync, because the log line had project UUID + endpoint but no scheduler attribution. Each of these five changes shrinks that loop:

| Before | After |
|---|---|
| ``error: Connection terminated unexpectedly`` (with ``project_uuid`` only) | ``error: Connection terminated unexpectedly`` + ``scheduler_uuid``, ``scheduler_name``, ``saved_sql_uuid``, ``job_id``, ``organization_name`` |
| Column-discovery LIMIT-1 query: silent | ``column_discovery.completed durationMs:7.97 columnCount:2 …`` |
| ``POST … 200 - 600612 ms`` (regex the message) | ``{\"duration_ms\":600612, …}`` |
| Failure email links to ``/scheduler/settings`` | Failure email links to ``/scheduler/settings?ref=<jobId>`` |
| Filter pods by grepping ``172a2270-…`` in message text | Filter by ``organization_name=\"acme-co\"`` structured field |

### Sample log lines (verified live)

HTTP access log:
```json
{\"duration_ms\":79,\"level\":\"http\",\"message\":\"POST /api/v2/projects/…/query/metric-query 200 - 79 ms\",\"organization_name\":\"Acme Co\",\"organization_uuid\":\"…\"}
```

Deep inside ``AsyncQueryService`` (no HTTP middleware involvement — proves the ``ExecutionContext`` propagation works):
```json
{\"level\":\"info\",\"message\":\"Query … completed: source=warehouse …\",\"organization_name\":\"Acme Co\",\"organization_uuid\":\"…\",\"serviceName\":\"AsyncQueryService\"}
```

Column-discovery event:
```json
{\"event\":\"column_discovery.completed\",\"durationMs\":7.97,\"columnCount\":2,\"project_uuid\":\"…\",\"query_context\":\"sqlRunner\",\"organization_name\":\"Acme Co\",…}
```

## Test plan

- [x] ``pnpm -F backend typecheck`` passes
- [x] ``pnpm -F backend lint`` passes
- [x] ``pnpm -F common build`` passes
- [x] Smoke test: triggered metric query — confirmed ``duration_ms`` + ``organization_uuid`` + ``organization_name`` on HTTP and service logs
- [x] Smoke test: triggered SQL runner query — confirmed ``column_discovery.started`` / ``completed`` events with the new structured fields
- [x] No API/scheduler crash-loop after restart
- [ ] Trigger an actual scheduler failure to verify ``?ref=`` appears in the email link (code reviewed; requires a forced delivery failure to exercise live)

## Out of scope

- ``organization_slug`` (does not exist in this codebase)
- Scheduler-context tagging for non-scheduler entry points (embed/JWT queries explicitly exclude these tags via ``Omit<>`` in ``EmbedService``)

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -61,6 +61,7 @@ import Logger from './logging/logger';
 import {
     expressWinstonMiddleware,
     expressWinstonPreResponseMiddleware,
+    requestExecutionContextMiddleware,
 } from './logging/winston';
 import { sessionAccountMiddleware } from './middlewares/accountMiddleware';
 import { jwtAuthMiddleware } from './middlewares/jwtAuthMiddleware';
@@ -616,6 +617,9 @@ export default class App {
         // We'll also be able to add the user to Sentry for embedded users.
         expressApp.use(jwtAuthMiddleware);
         expressApp.use(sessionAccountMiddleware);
+        // Must run after auth so req.user is populated. Stamps every downstream
+        // log line with organization_uuid + organization_name via ExecutionContext.
+        expressApp.use(requestExecutionContextMiddleware);
 
         expressApp.use((req, res, next) => {
             if (req.user) {
--- a/packages/backend/src/clients/EmailClient/EmailClient.ts
+++ b/packages/backend/src/clients/EmailClient/EmailClient.ts
@@ -28,6 +28,17 @@ import Logger from '../../logging/logger';
 
 const RETRYABLE_ERROR_CODES = ['ECONNRESET', 'ETIMEDOUT', 'ENOTFOUND'];
 
+/**
+ * Appends `?ref=<id>` (or `&ref=`) to the URL so support can paste the
+ * correlation ID into Cloud Logging and land on the failing job directly,
+ * instead of triangulating from project UUID + approximate timestamp.
+ */
+function appendCorrelationRef(url: string, correlationId?: string): string {
+    if (!correlationId) return url;
+    const separator = url.includes('?') ? '&' : '?';
+    return `${url}${separator}ref=${encodeURIComponent(correlationId)}`;
+}
+
 function isNodemailerSmtpError(
     error: unknown,
 ): error is SMTPConnection.SMTPError {
@@ -369,6 +380,7 @@ export default class EmailClient {
         schedulerUrl: string,
         errorMessage?: string,
         disabledSync: boolean = true,
+        correlationId?: string,
     ) {
         // Sync failure but not disabled - will be retried
         if (!this.canSendEmail()) {
@@ -383,6 +395,8 @@ export default class EmailClient {
             throw new Error('Email transporter not configured');
         }
 
+        const urlWithRef = appendCorrelationRef(schedulerUrl, correlationId);
+
         // Sync has been disabled
         if (disabledSync) {
             return this.sendEmail({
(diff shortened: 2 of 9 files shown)


charliedowler:
Addressed in 0afebf1: split the helper into a pure `buildSchedulerLogContext` (returns the sub-context or null) plus a `setSchedulerJobLogContext` that takes an injected `SchedulerLogContextUpdater` (default writes through `ExecutionContext`). Tests now pass a jest spy instead of needing an AsyncLocalStorage setup — 6 unit tests added in `SchedulerTask.test.ts` covering empty-input, nullish-field, populated-field, and skip-when-empty behavior.

charliedowler:
Partially addressed in 557bd45: dropped the single `invokes the injected updater with the built sub-context` test — it's the closest fit to the "data transformation / implementation details" heuristic. Field-shape assertions stay on `buildSchedulerLogContext` (the pure function) where they belong.

Keeping the other three:

- **`skips the updater entirely when no attribution fields are set`** — verifies a behavior contract (don't pollute ExecutionContext with empty writes), not call-count.
- **`writes through the default ExecutionContext updater`** — integration test against a real `ExecutionContext`, no mock. Validates the production code path.
- **`is a no-op when called outside an ExecutionContext`** — safety-guard test for the `exists()` check.

Worth flagging that this rule reads as somewhat in tension with the earlier DI suggestion in the same review — if a dependency is injected for testability, verifying it's used correctly is the natural follow-on.

PR #23256, opened 2026-05-19
feat: add source field picker to custom metric modal with hover preview support

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7755/i-want-to-be-able-to-see-change-the-field-used-to-create-my-custom

CleanShot 2026-05-19 at 08 57 52

### Description:

Adds a **Source field** picker to the Custom Metric modal that lets users change which dimension a custom metric aggregates over when editing, without having to delete and recreate the metric.

- When editing a custom metric, a searchable "Source field" dropdown appears showing all dimensions on the same table that are compatible with the metric's aggregation type. Each option renders with its field icon for easy identification.
- Selecting a different field swaps the underlying SQL and table reference on save while preserving the metric's ID, name, filters, and format options — so saved charts and dashboards continue to work without any changes.
- An info callout is shown when a different source field is selected, explaining what will change.
- When creating a new custom metric from a dimension, the source field is shown as read-only to make it clear which field is being aggregated.
- The hover preview card for custom metrics now shows a "Source field" section displaying the dimension's icon, table label, and field label.
- Custom metric hover cards also now correctly display the aggregated SQL (e.g. `SUM(${table.column})`) and the metric's description.
- Fixed the HoverCard horizontal offset for custom metrics and custom dimensions, which don't render the filter action icon that regular fields do, so the popover now aligns consistently.

### DIFF
--- a/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
+++ b/packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
@@ -7,8 +7,10 @@ import {
     isField,
     isFilterableField,
     isTimeInterval,
+    MetricType,
     timeFrameConfigs,
     type AdditionalMetric,
+    type Dimension,
     type FilterableField,
     type Item,
 } from '@lightdash/common';
@@ -37,6 +39,7 @@ import {
     useExplorerSelector,
     type ExplorerStoreState,
 } from '../../../../../features/explorer/store';
+import { useExplore } from '../../../../../hooks/useExplore';
 import { useAddFilter } from '../../../../../hooks/useFilters';
 import useTracking from '../../../../../providers/Tracking/useTracking';
 import { EventName } from '../../../../../types/Events';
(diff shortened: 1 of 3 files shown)


rephus:
* **#23256** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/components/Explorer/ExploreTree/TableTree/Tree/TreeSingleNode.tsx
```diff
@@ -68,6 +71,30 @@ const NavItemIcon = ({
 
 NavItemIcon.displayName = 'NavItemIcon';
 
+/**
+ * Mirrors `getDefaultMetricSql` from the warehouse package so we can show the
```
rephus, line 75: we should probably move to common then

PR #23264, opened 2026-05-19
feat(ai): repair invalid tool-call inputs before they poison the conversation

joaoviana:
Related to: https://lightdash.sentry.io/issues/7490528088/?alert_rule_id=15833412&alert_type=issue&environment=cloud_beta&project=5959292&referrer=slack

## Summary

When the model emits syntactically broken JSON for a tool call (e.g. an omitted value like `{"query": ,`), the AI SDK throws `AI_InvalidToolInputError`. The assistant turn carrying the malformed `tool_use` block stays in the conversation, and the next round to Anthropic gets a hard 400:

```
AI_APICallError: messages.N.content.M.tool_use.input: Input should be an object
```

…which then cascades into further failure modes for the same thread.

This PR plugs `experimental_repairToolCall` into both the streaming and non-streaming agent paths so the AI SDK gets one chance to fix the arguments before the bad turn hits the wire.

## How it works

```ts
experimental_repairToolCall: buildRepairToolCall(args),
```

The repair helper:

1. Short-circuits on `NoSuchToolError` (not repairable — the model called a tool that doesn't exist).
2. On `InvalidToolInputError`, asks the configured model to re-emit valid arguments against the tool's input schema via `generateObject` (matching the pattern used by other generators in `packages/backend/src/ee/services/ai/agents/`).
3. Returns the corrected `LanguageModelV3ToolCall` with a re-stringified `input`, which the SDK then retries the tool call with.
4. If the repair itself errors, captures to Sentry under `errorType: 'AiAgentRepairToolCall'` and returns `null` so the original error propagates unchanged.

## Regression analysis

- Well-formed tool calls → repair function is never invoked; zero behaviour change.
- Calls to non-existent tools → still surface the original `NoSuchToolError` (we return `null` for that case).
- Calls with malformed input → one extra `generateObject` round against the same model + provider options the agent already uses; on success the tool runs as if the model had emitted valid JSON to begin with, on failure the original error stands.
- Cost/latency: a single extra model call per malformed tool call. Repair is rare in steady state, and the alternative (a poisoned conversation that 400s on every subsequent stream) is strictly worse.

## Test plan

- [ ] Backend typecheck passes (`pnpm -F backend typecheck`)
- [ ] Backend lint passes (`pnpm -F backend lint`)
- [ ] Manually trigger an invalid tool call (e.g. ask the agent something that historically produced malformed JSON for `searchFieldValues`) and confirm the repair round happens and the thread completes

### DIFF
--- a/packages/backend/src/ee/services/ai/agents/agentV2.ts
+++ b/packages/backend/src/ee/services/ai/agents/agentV2.ts
@@ -1,12 +1,16 @@
 import { AgentToolOutput, assertUnreachable, Explore } from '@lightdash/common';
 import * as Sentry from '@sentry/node';
 import {
+    generateObject,
     generateText,
+    jsonSchema,
+    NoSuchToolError,
     smoothStream,
     stepCountIs,
     streamText,
     StreamTextResult,
     type Output,
+    type ToolCallRepairFunction,
     type ToolSet,
 } from 'ai';
 import Logger from '../../../../logging/logger';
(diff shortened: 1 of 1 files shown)


joaoviana:
* **#23264** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23264?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23264?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joaoviana:
Reworked the repair function to follow the AI SDK's canonical [reask example](https://github.com/vercel/ai/blob/fa67f5fb93d088e433c189899ab21858f84bbb55/examples/ai-functions/src/generate-text/mock/tool-call-repair-reask.ts) instead of the isolated `generateObject` approach in the first revision.

**Why the reask pattern is better here:** the failures we see in production aren't the model misunderstanding the tool's schema — they're streamed JSON truncation in long tool args (`runQuery`, `searchFieldValues`). The reask flow replays the full conversation (system prompt, prior tool results, discovered fields) and surfaces the validation error as a tool-error result, so the model re-emits the call with everything it already knows. The original isolated approach only had the bad args + error string in scope, which is a thin context for repairing calls that span dozens of fields.

**Side benefits:**
- Drops `generateObject` (deprecated in AI SDK v6) and `jsonSchema`. The reask path uses `generateText` end-to-end, which is the forward-compatible API.
- The corrected call comes from `result.toolCalls.find(tc => tc.toolName === toolCall.toolName)` — if the model decides on a different tool entirely (e.g. legitimately wrong tool choice), we bail to `null` rather than blindly re-stringify.

`NoSuchToolError` short-circuit and Sentry capture on repair-internal failure are preserved.

PR #23281, opened 2026-05-19
fix: flag dashboard filters referencing deleted explores as TableDoesNotExist

rephus:
Closes: [https://linear.app/lightdash/issue/PROD-5931/deleted-dbt-model-persists-in-dashboard-filter-dropdowns](https://linear.app/lightdash/issue/PROD-5931/deleted-dbt-model-persists-in-dashboard-filter-dropdowns)

### Description:

Dashboard filters that reference a deleted dbt model (explore) were previously silently ignored during validation — the existing field-level check only caught renamed or removed fields within a table that still existed. This meant deleted models could linger in the filter dropdown without any warning.

This adds a table-existence check to dashboard filter validation. A new `buildExistingTableNames` helper collects all known table names from compiled explores (including joined tables and broken-but-present models). During dashboard validation, both top-level filter targets and per-tile `tileTargets` are checked against this set before the field-level check runs. If the table is missing, a `"Table 'X' no longer exists"` error is emitted — matching the exact wording that `ValidationModel.parseDashboardFilterError` uses to categorise the error as `TableDoesNotExist` in the UI, enabling the rename/delete action in the bell-icon notification.

When the table exists but the field has been renamed or removed, the existing `FieldDoesNotExist` error is still emitted as before.

### DIFF
--- a/packages/backend/src/services/ValidationService/ValidationService.test.ts
+++ b/packages/backend/src/services/ValidationService/ValidationService.test.ts
@@ -322,6 +322,180 @@ describe('validation', () => {
         );
     });
 
+    it('Should flag dashboard filter referencing a deleted explore as TableDoesNotExist', async () => {
+        (
+            projectModel.findExploresFromCache as jest.Mock
+        ).mockImplementationOnce(async () => [explore]);
+        (
+            dashboardModel.findDashboardsForValidation as jest.Mock
+        ).mockImplementationOnce(async () => [
+            {
+                ...dashboardForValidation,
+                filters: {
+                    dimensions: [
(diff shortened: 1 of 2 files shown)


rephus:
* **#23281** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23292, opened 2026-05-19
fix: preserve original URL across SSO redirects via query param

rephus:
Closes: https://linear.app/lightdash/issue/PROD-4795/when-im-logged-out-upon-viewing-a-certain-piece-of-content-i-am-not

### Description:

When unauthenticated users visit a protected route, the original URL is now preserved as a `redirect` query parameter on the `/login` redirect. This complements the existing `state.from` mechanism, which can be lost across hard redirects (`window.location.href`), server-issued redirects, and certain browser refresh paths — the query parameter survives all of these cases.

For SSO flows with `forceRedirect`, the `redirect` query parameter is forwarded to the SSO URL so the backend can use it as a `returnTo` value, bringing the user back to their originally requested page after authentication completes instead of dropping them on `/`.

### DIFF
--- a/packages/frontend/src/components/PrivateRoute.tsx
+++ b/packages/frontend/src/components/PrivateRoute.tsx
@@ -32,10 +32,20 @@ const PrivateRoute: FC<React.PropsWithChildren> = ({ children }) => {
     }
 
     if (!health.data?.isAuthenticated) {
+        // Also pass the original URL as a query param. State.from is the
+        // primary mechanism, but it can be lost across hard redirects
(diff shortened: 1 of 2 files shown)


rephus:
* **#23292** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23300, opened 2026-05-19
[codex] Unify SQL auto-approval mode

joaoviana:
## Summary

- Replaces the loose auto-approval boolean/user UUID pair with an explicit SqlApprovalMode type.

- Keeps eval SQL runs auto-approved with the eval runner recorded as the approver.

- Preserves Slack thread auto-approval as a separate null-audited path for now.

## Validation
- pnpm --filter backend test -- --runTestsByPath src/ee/services/ai/tools/runSql.test.ts --runInBand
- pre-commit hook: backend linter/formatter and frontend unused exports


joaoviana:
need to chat with @tatianainama first

PR #23312, opened 2026-05-20
fix: forward `redirect` param to SSO URL so post-login returns to correct page

rephus:
Closes: https://linear.app/lightdash/issue/PROD-4795/when-im-logged-out-upon-viewing-a-certain-piece-of-content-i-am-not

### Description:
When a user belonging to an SSO-only organisation attempts to access a specific page, they were always redirected to `/` after completing authentication instead of being returned to their intended destination.

This fix appends the intended `redirect` URL as a query parameter to the SSO redirect URI before navigating the user to the identity provider. This allows the backend's `storeOIDCRedirect` to persist the `returnTo` value, ensuring users land on the correct page after completing SSO login. The `redirectUrl` dependency has also been added to the `useEffect` to keep the behaviour reactive to changes in the redirect target.

### DIFF
--- a/packages/frontend/src/features/users/components/LoginLanding.tsx
+++ b/packages/frontend/src/features/users/components/LoginLanding.tsx
@@ -97,10 +97,17 @@ const Login: FC<{}> = () => {
     useEffect(() => {
         if (loginOptions && loginOptionsSuccess) {
             if (loginOptions.forceRedirect && loginOptions.redirectUri) {
-                window.location.href = loginOptions.redirectUri;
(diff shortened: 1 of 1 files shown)


rephus:
* **#23312** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23330, opened 2026-05-20
feat: bundle BigQuery scope into Google login flow to avoid double OAuth consent (PROD-7783)

### LINKED ISSUE #23206: Auto-redirect to BigQuery OAuth after Google SSO login
### Description

When using Google SSO login with BigQuery SSO warehouse auth, users must complete two separate Google OAuth flows. Chain them automatically so users authenticate once seamlessly.

## Feature Request

When using Google SSO for login and BigQuery SSO for warehouse authentication, users are required to complete two separate Google OAuth flows — one for login (`profile, email` scope) and another for BigQuery access (`bigquery` scope). This dual authentication causes confusion, particularly for non-technical users.

Since BigQuery SSO was made the default authentication method in v0.2370.0 (#19667), this UX friction now affects a wider range of deployments.

### Why this is needed

- Users who log in with Google SSO and use BigQuery SSO must complete two separate Google OAuth flows with different scopes (`profile, email` vs `bigquery`).
- The BigQuery OAuth is only accessible through the settings page ("My Warehouse connections"), which is not discoverable for non-technical users.
- Organizations with daily re-authentication policies (e.g., short session TTL or token rotation) are particularly impacted — users go through both flows every day.
- Now that BigQuery SSO is the default (#19667), this affects more deployments than before.

### Proposed Solution

After a successful Google login, automatically redirect the user to the BigQuery OAuth flow (`/api/v1/login/bigquery`) if:

1. The project's warehouse type is BigQuery with `requireUserCredentials: true`, **and**
2. The user does not have valid BigQuery warehouse credentials stored

This would chain the two OAuth flows so users see:

```
Google Login consent → BigQuery consent → App dashboard
```

instead of:

```
Google Login consent → App dashboard → (user navigates to Settings) → BigQuery consent
```

### Implementation Notes

*Based on AI-assisted code research — may not reflect the actual implementation.*

The existing architecture already supports this with minimal changes:

- The Google passport strategy callback (`googleStrategy.ts`) already detects the `bigquery` scope via `params.scope.includes('bigquery')` and creates warehouse credentials automatically.
- The `getOidcRedirectURL` middleware (`middlewares.ts`) supports a `returnTo` session field that controls post-login redirect.
- The `/api/v1/login/bigquery` route already exists and requests `bigquery` scope with `accessType: 'offline'`.

A possible approach: in `googleStrategy.ts`, when the login OAuth callback completes without `bigquery` scope, set `req.session.oauth.returnTo = '/api/v1/login/bigquery'` to chain the flows. This could be gated behind an environment variable (e.g., `AUTH_GOOGLE_AUTO_BIGQUERY_SSO=true`) so it's opt-in.

### Workaround

Currently, admins must instruct users to manually navigate to Settings → My Warehouse connections → Add new credentials → Sign in with Google after every login/session expiry.

### Related

- Slack community discussion: https://lightdash-community.slack.com/archives/C03MG2VTR08/p1778843892797129

rephus:
Closes:  https://linear.app/lightdash/issue/PROD-7783/auto-redirect-to-bigquery-oauth-after-google-sso-login
Closes: https://github.com/lightdash/lightdash/issues/23206

### Description:
When BigQuery SSO is enabled, users previously had to complete two separate OAuth consent screens — one for login and one for BigQuery warehouse access. This adds a new config option (`AUTH_GOOGLE_INCLUDE_BIGQUERY_SCOPE`) that, when enabled, bundles the BigQuery scope into the initial Google login flow so users only see a single consent screen.

When `includeBigqueryScope` is `true`, the login request includes the BigQuery scope alongside `profile` and `email`, and requests offline access with a forced consent prompt to ensure a refresh token is returned. The BigQuery warehouse credentials are then created after the user is resolved from the login flow, rather than before, ensuring the authenticated user object is available when storing the credentials.

### DIFF
--- a/packages/backend/src/controllers/authentication/strategies/googleStrategy.ts
+++ b/packages/backend/src/controllers/authentication/strategies/googleStrategy.ts
@@ -65,19 +65,6 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
                   };
                   const hasBigqueryScope = params.scope.includes('bigquery');
 
-                  if (hasBigqueryScope) {
-                      // we'll also be adding the token to the warehouse credentials
-                      // so they can use it to query bigquery
-                      Logger.info(
-                          `Creating user warehouse credentials for bigquery on Google OAuth`,
-                      );
-                      await req.services
-                          .getUserService()
-                          .createBigqueryWarehouseCredentials(
-                              req.user!,
-                              refreshToken,
-                          );
-                  }
                   const user = await req.services
                       .getUserService()
                       .loginWithOpenId(
@@ -92,6 +79,18 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
                               ),
                           },
                       );
+
+                  if (hasBigqueryScope && user) {
+                      Logger.info(
+                          `Creating user warehouse credentials for bigquery on Google OAuth`,
+                      );
+                      await req.services
+                          .getUserService()
+                          .createBigqueryWarehouseCredentials(
+                              user,
+                              refreshToken,
+                          );
+                  }
                   return done(null, user);
               } catch (e) {
                   if (e instanceof LightdashError) {
--- a/packages/backend/src/config/lightdashConfig.mock.ts
+++ b/packages/backend/src/config/lightdashConfig.mock.ts
@@ -25,6 +25,7 @@ export const lightdashConfigMock: LightdashConfig = {
             googleDriveApiKey: undefined,
             enableGCloudADC: false,
             enabled: false,
+            includeBigqueryScope: false,
         },
         okta: {
             loginPath: '',
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -1398,6 +1398,13 @@ export type AuthGoogleConfig = {
     googleDriveApiKey: string | undefined;
     enabled: boolean;
     enableGCloudADC: boolean;
+    /**
+     * When true, the Google login flow also requests the BigQuery scope so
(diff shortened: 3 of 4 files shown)


rephus:
* **#23330** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23330?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23330?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/controllers/authentication/strategies/googleStrategy.ts
```diff
@@ -92,6 +79,18 @@ export const googlePassportStrategy: GoogleStrategy | undefined = !(
                               ),
                           },
                       );
+
+                  if (hasBigqueryScope && user) {
```
rephus, line 83: The original code at googleStrategy.ts:66-80 was designed for an already-logged-in user adding BigQuery from Settings → it called
  createBigqueryWarehouseCredentials(req.user!, refreshToken). In that flow, req.user is already populated from the existing session.

  But when the BQ scope is bundled into the login flow (your new env-var path), the callback runs before the user has been authenticated — req.user is
   still undefined. The ! told TypeScript to ignore it, but at runtime it blew up at user.userUuid inside createBigqueryWarehouseCredentials.

  The fix was to swap the order: call loginWithOpenId first (which creates/logs in the user and returns a SessionUser), then pass that returned user
  into createBigqueryWarehouseCredentials. The "logged-in user adds BQ from Settings" path still works because loginWithOpenId is idempotent for
  already-authenticated users.

PR #23389, opened 2026-05-21
fix: prevent dashboard filter popover from dismissing on child popup interaction

rephus:
Closes: 

See thread  https://lightdash.slack.com/archives/C01788UVBUG/p1779365281928669
Area

### Description:
 Bug 1: Panel grew horizontally when adding values

  Symptom: Selecting multiple values in a multi-select pushed pills onto the same row, expanding the popover panel width from ~350px up to ~520px.

  Cause: Tabs.Panel value={FilterTabs.SETTINGS} had miw={350} maw={520} — a flexible range that grew with content.

  Fix: Pinned to fixed width.
  -<Tabs.Panel value={FilterTabs.SETTINGS} miw={350} maw={520}>
  +<Tabs.Panel value={FilterTabs.SETTINGS} w={400}>
  Pills now wrap to a new line instead of widening the panel.

  ---
  Bug 2: Apply needed two clicks when a dropdown was open

  Symptom: With an inline dropdown (operator / values / boolean) open, the first click on Apply only closed the dropdown — a second click was needed
  to actually apply and close the popover.

  Cause: Mantine v6's useClickOutside fires on mousedown to close the dropdown, and the subsequent click event on the Apply button never fires (likely
   because the re-render or focus shift swallows it).

  Fix: Switch Apply from onClick to onMouseDown, which fires before Mantine's outside-click detector runs.
  -onClick={() => {
  +onMouseDown={() => {
       setSelectedTabId(FilterTabs.SETTINGS);
       if (!!draftFilterRule) onSave(draftFilterRule);
   }}

### DIFF
--- a/packages/frontend/src/features/dashboardFilters/FilterConfiguration/index.tsx
+++ b/packages/frontend/src/features/dashboardFilters/FilterConfiguration/index.tsx
@@ -410,7 +410,7 @@ const FilterConfiguration: FC<Props> = ({
                     </Tabs.List>
                 ) : null}
 
-                <Tabs.Panel value={FilterTabs.SETTINGS} miw={350} maw={520}>
+                <Tabs.Panel value={FilterTabs.SETTINGS} w={400}>
                     <Stack spacing="sm">
                         {isCreatingNew ? (
                             !!fields && fields.length > 0 ? (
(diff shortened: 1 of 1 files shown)


rephus:
* **#23389** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23389?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23389?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23397, opened 2026-05-21
feat: add `showRowGrouping` pivot design spec (PROD-5901)

pimmey:
Closes: PROD-5901

### Description:

Adds a design spec for allowing users to visually group (deduplicate) repeated row header values in pivot tables without requiring subtotals to be enabled.

Currently, row value deduplication is only available as a side effect of enabling `showSubtotals`, which also renders aggregate subtotal rows. This design separates the two concerns by introducing a new `showRowGrouping` boolean in the `TableChart` config.

Key design decisions:
- `showRowGrouping` is independent of `showSubtotals`. When subtotals are enabled, row grouping is implicitly active and the new toggle is disabled in the UI with an explanatory tooltip.
- The existing TanStack `setGrouping` machinery is reused — the new mode simply suppresses rendering of aggregate rows when `showRowGrouping=true` and `showSubtotals=false`.
- Groups start fully expanded with no visible expand/collapse caret when subtotals are off.
- The existing `.slice(0, -1)` rule (which excludes the leaf dimension from grouping) is preserved.
- Defaults to `false` for new and existing charts with no migration required.
- CSV/XLSX exports are unaffected as deduplication is purely visual.

### DIFF
--- a/docs/superpowers/specs/2026-05-21-prod-5901-pivot-row-grouping-design.md
+++ b/docs/superpowers/specs/2026-05-21-prod-5901-pivot-row-grouping-design.md
@@ -1,0 +1,89 @@
+# PROD-5901 — Group repeated row values in pivot tables (without requiring subtotals)
+
+**Status:** Design — not implemented
+**Linear:** [PROD-5901](https://linear.app/lightdash/issue/PROD-5901/groupmerge-same-row-values-in-pivot-tables-without-requiring-subtotals)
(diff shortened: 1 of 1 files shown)


pimmey:
* **#23397** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23397?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23397?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23357, opened 2026-05-20
feat: add theme management UI

magnew:
Relates to: PROD-7799

Description:

Add FE page for org themes

PR #23406, opened 2026-05-21
fix: load images from theme correctly in apps

magnew:
Relates to: PROD-7840

### Description:

Fix image URLs for images loaded from theme.

The data-app preview iframe now carries its JWT auth token as a URL path segment (/api/apps/{appUuid}/versions/{version}/t/{token}/...) instead of a query parameter (?token=...). This was driven by a runtime auth failure — Vite content-hashes asset URLs into the JS bundle at build time, so React-rendered calls were hitting our auth-gated asset endpoint without the token and getting 401s. The previous attempt to fix this required injecting an inline bootstrap script that monkey-patched DOM prototype setters (HTMLImageElement.prototype.src, setAttribute, etc.) plus a MutationObserver to rewrite every src/href assignment, all gated by a per-request CSP nonce. With the token now in the path, Vite's relative URLs (./assets/foo.js) naturally resolve against the iframe URL and inherit the token segment automatically — no rewriting needed. Security-wise this is a clear win: CSP tightens from script-src 'self' 'nonce-XXX' (inline scripts permitted) to script-src 'self' (no inline scripts at all), ~110 lines of HTML-mutation and DOM-patching code go away, a new isPlausibleToken regex gates input before HMAC verify as defense in depth, and CDN cache keys now naturally include the token (closing a default-config footgun where authed responses could be served to anonymous requests). The one operational follow-up is that path-segment secrets are less commonly auto-redacted by log scrubbers than ?token= query params, so a small URL-sanitizing logger middleware is worth adding — but the pre-existing scheme had the same exposure, so this is a refinement rather than a regression.

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -313,7 +313,7 @@ export default class App {
         // route. The matching GET handler remains auth-gated by `requireToken`.
         if (this.lightdashConfig.appRuntime.s3) {
             expressApp.options(
-                '/api/apps/:appUuid/versions/:version/assets/:filename',
+                '/api/apps/:appUuid/versions/:version/t/:token/assets/:filename',
                 (_req, res) => {
                     res.setHeader('Access-Control-Allow-Origin', '*');
                     res.setHeader(
--- a/packages/backend/src/routers/appPreviewRouter.ts
+++ b/packages/backend/src/routers/appPreviewRouter.ts
@@ -62,18 +62,12 @@ const isSafeFilename = (filename: string): boolean =>
     /^[a-zA-Z0-9._-]+$/.test(filename) && !filename.includes('..');
 
 /**
- * Rewrites Vite-generated asset references in HTML to include a token
- * query parameter so asset requests authenticate themselves.
- *
- * Matches patterns like:
- *   src="./assets/index-BxK29f.js"
- *   href="./assets/style-DhJ93k.css"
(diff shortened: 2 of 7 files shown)


magnew:
* **#23415** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23415?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23407** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23406** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23406?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23406?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23393** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23393?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23357** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23356** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23356?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #23426, opened 2026-05-21
feat: enhancements to transient network retries

agha4to:
Closes: N/A

### Description:
Continuing work in https://github.com/lightdash/lightdash/pull/20400, when `LIGHTDASH_QUERY_RETRY_ON_TRANSIENT_ERRORS`, is enabled these changes turn on network retries by default unless explicitly opted out with `retry: false`.

Notes:
1. "Synthetic" network errors no longer retried so errors thrown by the frontend do not trigger unnecessary retries
2. Retry timing and "jitter" changed to allow more time for servers to recover. In case of network issues, UI components will take longer to finalize rendering than previous version, but there should be fewer outright failures now.
3. Retries occur closer to the network layer whereas some retries in the previous version would trigger errors in the UI, causing some retries to be unused
4. While ['health'] is still unloaded so the value of `LIGHTDASH_QUERY_RETRY_ON_TRANSIENT_ERRORS` is not yet known, retries are enabled by default. Once health has loaded, the value of `LIGHTDASH_QUERY_RETRY_ON_TRANSIENT_ERRORS` is used.

We've been testing this the past few weeks. Happy to discuss our results and open to any changes. Thank you.


ZeRego:
Closing this in favour of #23831, which shipped the core of this change — defaulting the React Query client to retry transient `NetworkError` transport failures (5 attempts, exp backoff capped at 8s) so a single blip recovers silently instead of surfacing the "unable to reach the Lightdash server" toast. Synthetic `queryFn` errors are excluded there by error name, and the ~40 `retry: false` opt-out hooks are untouched.

That makes the bulk of this PR redundant, and it now conflicts on `createQueryClient.ts`. A few ideas here were deliberately *not* included in #23831 and could come back as small, focused follow-ups if we still want them:
- retrying 5xx server errors (not just `NetworkError`)
- retrying transient stream/fetch failures from `getResultsFromStream`
- the `LIGHTDASH_QUERY_RETRY_ON_TRANSIENT_ERRORS` feature-flag kill switch
- longer full-jitter backoff to give servers more recovery time
- offline-awareness + retry analytics
- removing the legacy `useQueryRetry` hook

Thanks for the work and the testing — happy to review any of the above as a slim PR on top of the merged version.

PR #23427, opened 2026-05-22
fix: replace boolean `enabled` with `permission_mode` enum for MCP server tool settings

notgiorgi:
Closes:

### Description:

Replaces the boolean `enabled` field on `ai_agent_mcp_server_tool` with a `permission_mode` column that supports three explicit states: `always_allow`, `ask`, and `always_deny`.

A database migration adds the `permission_mode` column with a default of `always_deny` and ensures the existing `enabled` column has a safe default for rolling deploys. The `enabled` column is marked as deprecated in the type definition and retained only for backward compatibility during the transition.

Internally, all model methods now read and write `permission_mode` instead of `enabled`. The public API surface is kept stable — `listAgentMcpServerTools` and `updateAgentMcpServerTools` still accept and return the `enabled` boolean, with `AiAgentService` handling the translation between the boolean API shape and the new `permission_mode` values (`true` → `always_allow`, `false` → `always_deny`).

Integration tests are extended to assert that the correct `permission_mode` values are persisted to the database at each stage of the tool settings lifecycle.

### DIFF
--- a/packages/backend/src/ee/database/entities/aiAgent.ts
+++ b/packages/backend/src/ee/database/entities/aiAgent.ts
@@ -264,11 +264,25 @@ export type AiMcpServerToolTable = Knex.CompositeTableType<
 
 export const AiAgentMcpServerToolTableName = 'ai_agent_mcp_server_tool';
 
+export const AiAgentMcpServerToolPermissionModes = [
+    'always_allow',
+    'ask',
+    'always_deny',
+] as const;
+
+export type DbAiAgentMcpServerToolPermissionMode =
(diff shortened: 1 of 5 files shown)


notgiorgi:
* **#23427** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23427?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23427?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23432, opened 2026-05-22
feat: render multi-tab dashboard exports in per-tab grids with deterministic screenshot target

rephus:
Closes: [https://linear.app/lightdash/issue/PROD-2505/issues-with-orphan-tiles-in-dashboards-with-tabs](https://linear.app/lightdash/issue/PROD-2505/issues-with-orphan-tiles-in-dashboards-with-tabs)

Before

After

### Description:

Fixes multi-tab dashboard PDF exports where tiles from different tabs were being incorrectly reflowed into each other's empty cells by `react-grid-layout`'s vertical compaction.

Previously, all tabs' tiles were rendered in a single `<ResponsiveGridLayout>`, which caused the grid library to compact tiles across tab boundaries. Now, each tab gets its own `<ResponsiveGridLayout>` instance so every tab starts at its own `y=0` origin, preventing cross-tab tile reflow.

To give the backend a single deterministic screenshot target that encompasses all per-tab grids, a wrapper `<div>` with the class `lightdash-dashboard-screenshot-target` is introduced. The `DASHBOARD_GRID_CLASS` constant is updated from `react-grid-layout` (a class applied by the third-party library to every grid instance, causing multiple matches) to this new wrapper class.

Orphan tiles (tiles with no `tabUuid`) are merged into the first tab's grid group to preserve their coordinate space relative to first-tab tiles, matching how the regular dashboard view renders them.

A rolling-deploy fallback is added in `UnfurlService` so that if a new backend is briefly paired with an old frontend bundle that doesn't yet render the new wrapper, the screenshot selector gracefully falls back to `.react-grid-layout` rather than failing to find any element.

### DIFF
--- a/packages/frontend/src/pages/MinimalDashboard.tsx
+++ b/packages/frontend/src/pages/MinimalDashboard.tsx
@@ -6,6 +6,7 @@ import type {
 } from '@lightdash/common';
 import {
     assertUnreachable,
+    DASHBOARD_GRID_CLASS,
     DashboardTileTypes,
     isDashboardScheduler,
     isTileInSelectedTabs,
@@ -47,9 +48,16 @@ type TabWithUrls = DashboardTab & {
     selfUrl: string;
 };
 
+type TabGroup = {
+    key: string;
+    tiles: Dashboard['tiles'];
+    layouts: { lg: Layout[]; md: Layout[] };
+};
+
 type MinimalDashboardContentProps = {
     filteredAndSortedDashboardTiles: Dashboard['tiles'];
     layouts: { lg: Layout[]; md: Layout[] };
+    tabGroups: TabGroup[] | null;
     gridProps: ReturnType<typeof getResponsiveGridLayoutProps>;
     isTabEmpty: boolean;
     canNavigateBetweenTabs: boolean;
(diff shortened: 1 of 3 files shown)


rephus:
* **#23432** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23432?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23432?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/frontend/src/pages/MinimalDashboard.tsx
```diff
@@ -47,19 +48,98 @@ type TabWithUrls = DashboardTab & {
...
     layouts: { lg: Layout[]; md: Layout[] };
+    tabGroups: TabGroup[] | null;
     gridProps: ReturnType<typeof getResponsiveGridLayoutProps>;
     isTabEmpty: boolean;
     canNavigateBetweenTabs: boolean;
     tabsWithUrls: TabWithUrls[];
     activeTab: DashboardTab | null;
 };
 
+const renderDashboardTile = (tile: Dashboard['tiles'][number]) => {
```
rephus, line 68: refactored, extract previous inline switch with this function

PR #23452, opened 2026-05-22
fix(pivot): add `passthroughDimensions` to carry hidden non-sort dims through SQL for cross-field `richText`/`image` templates

pimmey:
Closes: PROD-7873

### Description:

Fixes a bug where hidden pivot-column dimensions were being dropped entirely from the SQL pipeline, causing `richText` and `image:` templates on visible fields that referenced those hidden fields via `row.<table>.<field>.raw` to silently resolve to `undefined` and break image rendering in pivot tables.

**Root cause:** When a pivot-column dimension was hidden (`visible: false`) and not used as a sort target, it was excluded from `group_by_query` SELECT and GROUP BY entirely. This meant its value never reached the result rows, so any cross-field template referencing it had nothing to resolve.

**Fix:** Introduces a `passthroughDimensions` bucket in `PivotConfiguration`. Hidden pivot-column and row-index dimensions that are not sort targets are now routed into `passthroughDimensions` instead of being dropped. These dimensions:
- Are included in `group_by_query` SELECT and GROUP BY so their values flow through the SQL pipeline to each result row.
- Do **not** appear as rendered pivot column headers.
- Do **not** participate in `column_ranking` ORDER BY (they don't drive sort).
- Are registered as hidden TanStack columns (via `columnVisibility`) so their values surface in `getAllCells()` row context for richText/image template resolution.
- Are explicitly excluded from CSV/XLSX exports — the "hide" semantic applies to exports as well.

A `status_image_url` additional dimension has been added to the jaffle shop demo `orders.yml` to reproduce and validate the richText + hide-pivot-column interaction, using a `CONCAT`-derived placeholder image URL and a `richText` template on the `status` field that references it via `row.orders.status_image_url.raw`.

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -11722,11 +11722,11 @@ const models: TsoaRoute.Models = {
                                                     subSchemas: [
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['rejected'],
+                                                            enums: ['accepted'],
                                                         },
                                                         {
                                                             dataType: 'enum',
-                                                            enums: ['accepted'],
+                                                            enums: ['rejected'],
                                                         },
                                                         {
                                                             dataType:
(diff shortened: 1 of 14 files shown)


pimmey:
* **#23452** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23452?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23454, opened 2026-05-22
feat: auto-trigger OAuth connection after MCP server creation

notgiorgi:
Closes:

### Description:

When creating an MCP server with OAuth authentication, the OAuth connection flow now starts automatically immediately after the server is created, rather than requiring users to manually trigger it from the MCP settings card afterward.

To support this, a popup window is pre-opened before the server creation request is made (to avoid popup blockers), and is passed directly into the OAuth connection flow on success. If server creation fails, the popup is closed cleanly. If the OAuth step fails after a successful server creation, the error is surfaced via the existing toast and the newly created MCP remains attached.

The button label for OAuth servers has been updated to **"Create and connect"** to reflect this combined action, and the informational alert explaining the two-step process has been removed since it is no longer needed.

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/AiAgentFormSetup.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/AiAgentFormSetup.tsx
@@ -139,6 +139,30 @@ export const AiAgentFormSetup = ({
         userGroupsFeatureFlagQuery.isSuccess &&
         userGroupsFeatureFlagQuery.data.enabled;
 
+    const handlePersistedMcpServerChange = useCallback(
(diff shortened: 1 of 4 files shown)


notgiorgi:
* **#23454** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23454?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23454?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23460, opened 2026-05-25
fix(sql-runner): preserve user ORDER BY → LIMIT adjacency

Spissable:
Closes: https://linear.app/lightdash/issue/PROD-7880/sql-runner-returns-wrong-aggregate-values-when-user-provided-limit-is

### Description:
Customer reported SQL Runner returning different aggregate values on Snowflake depending on the user-supplied LIMIT. Reproduced locally by running a CTE + COUNT(DISTINCT) + ORDER BY against JAFFLE.TRACKS_RAW via `lightdash sql`: rows came back unsorted — `report_exported,3352` appeared at row 16 below `ticket_reopened,345`, even though the user SQL ended with `ORDER BY unique_users DESC, EVENT_NAME ASC`. Snowflake QUERY_HISTORY confirmed the warehouse received:

    SELECT "EVENT_NAME" AS "EVENT_NAME", ...
    FROM (
      <user SQL ending with ORDER BY ...>
    ) AS "sql_query"
    LIMIT 50000

The user's ORDER BY is inside the subquery, LIMIT is on the outer SELECT. Per ANSI SQL — and explicitly per Snowflake/BigQuery/Redshift docs — ORDER BY inside an unconstrained subquery is a hint the optimizer is free to drop. So the outer LIMIT N returns N arbitrary rows. With small user-LIMITs (e.g. LIMIT 5) every run looks "wrong" in a different way; with the default 50k limit the bug was masked when all rows fit but ordering was still non-deterministic.

The wrap is built by SqlQueryBuilder, originally introduced to allow applying dashboard filters on top of user SQL. But SQL charts don't support filters yet (see AsyncQueryService.ts comment), so in 100% of today's production traffic the wrap is purely cosmetic — and it silently breaks ORDER BY.

Fix: when `from.sql` is set and no filters apply, bypass the wrap and return the user's SQL with LIMIT appended via the existing `applyLimitToSqlQuery` helper. The wrap branch is preserved for the unused filter path so a future SQL-chart-with-filters feature isn't broken by this change.

Why safe:
- The new shape is identical to the legacy `streamSqlQueryIntoFile` path that has shipped `applyLimitToSqlQuery` in production for years. We're aligning the v2 async path with the legacy path.
- Single production callsite (AsyncQueryService.prepareSqlChartAsync QueryArgs); the conditional gate is precise.
- Verified live on Snowflake (bug fixed — rows now sorted), Postgres (no observable change — Postgres preserved subquery ORDER BY de facto), and BigQuery (no observable change on the F1 dataset; fix eliminates latent optimizer risk on larger plans).
- 411 QueryBuilder tests pass, including a new regression test asserting ORDER BY stays adjacent to LIMIT for a customer-shaped query. Three existing subquery+limit snapshots updated to expect the new flat output.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.mocks.ts
+++ b/packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.mocks.ts
@@ -61,11 +61,13 @@ export const QUERY_WITH_FILTER_SQL =
 export const QUERY_WITH_TWO_FILTERS_SQL =
     'SELECT\n"test_table"."test_field" AS "test_field"\nFROM "test_table"\nWHERE ((\n("test_table"."test_field") IN (\'test_value\')\n) AND (\n("test_table"."test_field") IS NOT NULL\n))';
 
-export const QUERY_WITH_SUBQUERY_SQL =
-    'SELECT\n"test_field" AS "test_field"\nFROM (\nSELECT test_field FROM source_table WHERE test_field IS NOT NULL\n) AS "subquery"';
+// User-provided FROM SQL with no filters is no longer wrapped in an outer
+// SELECT — wrapping severs the user's ORDER BY from any LIMIT (PROD-7880).
+export const UNWRAPPED_FROM_SUBQUERY_SQL =
+    'SELECT test_field FROM source_table WHERE test_field IS NOT NULL';
 
-export const QUERY_WITH_SUBQUERY_SEMICOLON_COMMENTS_SQL =
-    'SELECT\n"test_field" AS "test_field"\nFROM (\nSELECT test_field FROM source_table WHERE test_field IS NOT NULL\n) AS "subquery"';
+export const UNWRAPPED_FROM_SUBQUERY_WITH_COMMENTS_SQL =
+    'SELECT test_field FROM source_table WHERE test_field IS NOT NULL';
 
 export const QUERY_WITH_NESTED_FILTERS_SQL =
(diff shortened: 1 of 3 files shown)


packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.mocks.ts
```diff
@@ -61,11 +61,13 @@ export const QUERY_WITH_FILTER_SQL =
 export const QUERY_WITH_TWO_FILTERS_SQL =
     'SELECT\n"test_table"."test_field" AS "test_field"\nFROM "test_table"\nWHERE ((\n("test_table"."test_field") IN (\'test_value\')\n) AND (\n("test_table"."test_field") IS NOT NULL\n))';
 
+// User-provided FROM SQL with no filters is no longer wrapped in an outer
+// SELECT — wrapping severs the user's ORDER BY from any LIMIT (PROD-7880).
 export const QUERY_WITH_SUBQUERY_SQL =
```
magnew: Minor, but this name is wrong now. Same with the next one.

magnew (approved):
I think this is good.

packages/backend/src/utils/QueryBuilder/SqlQueryBuilder.mocks.ts
Spissable: Thanks for spotting! 🕵️ Fixed

PR #23462, opened 2026-05-25
feat: add per-org Okta SSO configuration with DB-stored config and admin settings panel

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7549/admin-panel-okta-sso-configuration

### Description: 

Adds per-organization Okta SSO support, mirroring the existing Azure AD per-org SSO design. Organizations on shared multi-tenant instances can now configure their own Okta tenant via the Settings → Single Sign-On admin panel, independently of any instance-level Okta environment variables.

**How it works:**

- The `'okta'` passport strategy is now always registered at startup (no longer gated on instance-level env vars), since it resolves its config per-request rather than at registration time.
- On login, `resolveOktaConfig` checks for a matching per-org DB config via `findEnabledMethodForEmail(email, OKTA)`, falling back to the instance env config when no per-org match is found.
- The resolved `organizationUuid` is persisted on the session so the OAuth callback can re-resolve the same config without the login hint being present on the redirect back from Okta.
- A new `OktaSsoConfig` type and `OKTA` provider variant are added to `OrganizationSsoProvider`. The `organization_sso_configurations` table stores the encrypted config blob with fields for `oauth2Issuer`, `oktaDomain`, `oauth2ClientId`, `oauth2ClientSecret`, `authorizationServerId`, and `extraScopes`.
- New `/api/v1/org/sso/okta` GET/PUT/DELETE endpoints are exposed via `OrganizationSsoController`.
- The previously Azure-AD-specific `findEnabledAzureAdMethodForEmail` and `getAzureAdConfigForOrganization` service methods are replaced with generic `findEnabledMethodForEmail<P>` and `getConfigForOrganization<P>` variants used by both providers.
- A new `OktaSsoPanel` frontend component provides the credential form and discovery/password flags for org admins, mounted alongside the existing Azure AD panel on the `/sso` settings route.
- SSO provider logos (Google, Okta, OneLogin, Microsoft) are extracted into a shared `ssoProviderLogos.ts` file so both the login buttons and settings panels stay in sync.
- The Okta sign-in button now renders when `forceShow` is set, enabling it to appear for per-org users even when the instance-level Okta env config is absent.
- Unit tests cover `resolveOktaConfig` (env fallback, per-org precedence, missing config) and `getLoginOptions` discovery for per-org Okta (suppression of instance Google, `forceRedirect` behaviour, cross-org hijack defence).
- A new `docs/okta-sso.md` documents the dual-path design, storage schema, discovery flow, login/callback route, and differences from the Azure AD implementation.


rephus:
* **#23462** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23462?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23462?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23467, opened 2026-05-25
fix: include charts in loose dashbaords on upload

### LINKED ISSUE #23438: [CLI] `--include-charts` ignores loose dashboard files when extracting chart slugs
## Bug description

`lightdash upload --dashboards <slug> --include-charts` fails to extract chart slugs when the dashboard YAML file is stored outside the `dashboards/` subdirectory (a "loose" file classified by `contentType`). The chart upload step is skipped entirely with "No charts filters provided, skipping", while the dashboard upload succeeds because `upsertResources` handles loose files via `extraItems`.

The "Total charts updated: N" in the output is misleading - these come from the server-side dashboard promotion re-saving existing chart state, not from uploading local YAML changes.

## Steps to reproduce

1. Have a project with YAML files stored as loose files (outside `charts/` and `dashboards/` subdirectories), each with a `contentType` field
2. Run `lightdash upload --force --dashboards <slug> --include-charts --project <uuid>`
3. CLI output shows "No charts filters provided, skipping" followed by "Total charts updated: N"
4. Dashboard filters/layout update correctly but chart content is unchanged

## Expected behavior

`--include-charts` should discover the dashboard from loose files (just like `upsertResources` does) and extract chart slugs from its tiles, so that matching chart YAML files are also uploaded.

## Actual behavior

`getDashboardChartSlugs()` in `download.ts:1417` calls `readCodeFiles('dashboards', ...)` which only reads from the `dashboards/` subdirectory. Loose dashboard files (discovered by `readLooseCodeFiles`) are not considered. This results in an empty `chartSlugs` array, and the chart upload is skipped.

## Root cause

`getDashboardChartSlugs` uses `readCodeFiles` (folder-only) but `uploadHandler` already uses `readLooseCodeFiles` + `extraItems` to handle loose files for the actual upload. The fix is to also include loose dashboard files in `getDashboardChartSlugs`.

## Environment

- CLI version: latest
- File layout: flat/loose YAML with `contentType` fields (not nested into `charts/`/`dashboards/` subdirectories)

## Workaround

Explicitly pass chart slugs with `-c` instead of relying on `--include-charts`:
```
lightdash upload --force -d <dashboard-slug> -c <chart-slug-1> <chart-slug-2> ... --project <uuid>
```

Or move dashboard and chart YAML files into the standard `dashboards/` and `charts/` subdirectories.

magnew:
Closes: #23438, PROD-7869 

### Description:  

lightdash upload --dashboards <slug> --include-charts skips chart upload when dashboards are stored as "loose" YAML files (outside the dashboards/ subdirectory, classified by contentType). Output shows No charts filters provided, skipping, and local chart edits silently never reach the server. The misleading Total charts updated: N comes from the server re-saving existing chart state during dashboard promotion, not from the local YAML.

Root cause
getDashboardChartSlugs reads only via readCodeFiles('dashboards', ...) (folder-only), even though the rest of uploadHandler already discovers loose dashboards via readLooseCodeFiles. Loose dashboards never contribute slugs, so the chart upload step gets an empty filter and is skipped.

Fix
getDashboardChartSlugs now unions folder dashboards with loose dashboards passed in by the caller.
uploadHandler now calls readLooseCodeFiles before computing chartSlugs and passes looseFiles.dashboards in. Avoids a second directory walk.

### DIFF
--- a/packages/cli/src/handlers/download.test.ts
+++ b/packages/cli/src/handlers/download.test.ts
@@ -1,0 +1,113 @@
+import { DashboardAsCode } from '@lightdash/common';
+import { promises as fs } from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { testHelpers } from './download';
+
+const { getDashboardChartSlugs } = testHelpers;
+
+type LooseDashboard = DashboardAsCode & { needsUpdating: boolean };
+
+const makeLooseDashboard = (
+    slug: string,
+    chartSlugs: (string | undefined)[],
+): LooseDashboard =>
+    ({
+        slug,
+        name: slug,
+        spaceSlug: 'test-space',
+        version: 1,
+        tiles: chartSlugs.map((chartSlug) => ({
+            properties: chartSlug ? { chartSlug } : {},
+        })),
+        needsUpdating: false,
+    }) as unknown as LooseDashboard;
+
+const writeFolderDashboard = async (
+    baseDir: string,
+    slug: string,
+    chartSlugs: string[],
+) => {
+    const tilesYaml = chartSlugs
+        .map(
+            (s) =>
+                `  - properties:\n      chartSlug: ${s}\n    type: saved_chart`,
+        )
+        .join('\n');
+    const yaml = `contentType: dashboard\nname: ${slug}\nslug: ${slug}\nspaceSlug: test-space\ntiles:\n${tilesYaml}\nversion: 1\n`;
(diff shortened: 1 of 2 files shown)


magnew:
* **#23467** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23467?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23467?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23470, opened 2026-05-25
feat: redesign org SSO panels with collapsible config, provider logos, and status badges

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7894/unify-settings-on-single-sing-on

CleanShot 2026-05-25 at 15 59 21

### Description:

Redesigns the Azure AD and Okta SSO configuration panels in the organization settings to use a card-based layout with collapsible configuration sections.

Each provider card now displays the provider logo, name, and a short description at the top level. When a provider is already configured, an enabled/disabled toggle appears inline and persists immediately without requiring the form to be opened. Toggling enabled/disabled no longer requires re-entering the client secret. When no configuration exists, a "Not configured" badge is shown instead of the toggle.

The detailed configuration form is hidden behind an "Edit configuration" / "Set up [Provider]" button that expands a collapsible `FormSection`. This keeps the settings page clean while still allowing full access to all credential, discovery, and password sign-in options.

The SSO settings page itself has been simplified — the provider title, logo, and description have moved into each panel component, so the page now renders a minimal section header followed by the two provider cards.


rephus:
* **#23470** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23470?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23470?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23487, opened 2026-05-26
feat: add per-org Google SSO opt-out policy with admin settings panel and login enforcement

rephus:
Closes: [https://linear.app/lightdash/issue/PROD-7901/customers-can-disable-google-login-on-their-domains](https://linear.app/lightdash/issue/PROD-7901/customers-can-disable-google-login-on-their-domains)

### Description:

Adds per-organization Google SSO policy management, allowing organizations to explicitly disable Google sign-in for their domains while keeping the shared instance OAuth app as the default.

**Backend:**

- Adds `OrganizationSsoProvider.GOOGLE` to the provider enum and introduces `GoogleSsoConfig`, `GoogleSsoConfigSummary`, `UpsertGoogleSsoConfig`, and the corresponding API response types.
- Adds `GET /api/v1/org/sso/google`, `PUT /api/v1/org/sso/google`, and `DELETE /api/v1/org/sso/google` endpoints to `OrganizationSsoController`. A `null` result on GET means no explicit policy exists and Google follows the instance default.
- Adds `findGoogleMethodsForEmailDomain` to `OrganizationSsoModel`, which — unlike the existing enabled-only discovery query — also returns disabled rows, since a `google` row with `enabled: false` is the mechanism for opting out.
- Extracts the email-domain match clause into a reusable `emailDomainMatchClause` helper on the model.
- Adds `getGoogleConfig`, `upsertGoogleConfig`, and `deleteGoogleConfig` to `OrganizationSsoService`.
- Updates `UserService.isLoginMethodAllowed` to check per-org Google policy rows for the email's domain, honouring the opt-out only for users who belong to the disabling org (cross-org domain-hijack guard preserved).
- Updates `UserService.getLoginOptions` to suppress the Google button in all three login-option paths (per-org SSO discovery, returning-user linked-issuer list, and new-signup fallback) when the org has disabled Google for the domain.

**Frontend:**

- Adds `GoogleSsoPanel` to the organization SSO settings page. The panel shows a toggle to enable/disable Google (persisted immediately on change) and an expandable section for domain-override and password-sign-in flags. A "Reset to default" action deletes the policy row entirely.
- Adds `useGoogleSsoConfig`, `useUpsertGoogleSsoConfig`, and `useDeleteGoogleSsoConfig` hooks.
- The panel is only rendered when Google is enabled at the instance level (`health.auth.google.enabled`).


rephus:
* **#23487** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23487?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23487?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23504, opened 2026-05-26
fix(compiler): include custom granularity dims in joined tables when join specifies fields (PROD-7858)

jelstongreen:
## Summary

- Joined tables that restrict visible dimensions via `fields:` were silently dropping custom granularity child dimensions (defined via `custom_granularities` in `lightdash.config.yml`), while standard time interval children (`_day`, `_month`, ...) survived the filter
- Dashboard date zoom on a chart whose date dim comes from such a join returned `dateZoomApplied: false` for custom granularities — the bug from [PROD-7858](https://linear.app/lightdash/issue/PROD-7858)
- The visibility check at `exploreCompiler.ts:456` only accepted `dimension.timeInterval`; custom granularity dims store the granularity on `customTimeInterval` with `timeInterval: undefined`, so they failed the check
- Fix extends the condition to `(timeInterval || customTimeInterval)`, mirroring the existing behavior for standard intervals

## Test plan

- [x] New unit test added: `should include time interval and custom granularity child dimensions when their base is listed in join.fields` (asserts base, standard `_month`, and custom `_week_starting_friday` are all retained; unrelated dim is excluded)
- [x] `pnpm -F common test exploreCompiler` — 161/161 pass
- [x] `pnpm -F common typecheck` clean
- [x] `pnpm -F common lint` clean
- [x] End-to-end repro on local jaffle_shop: added `fields: [order_date, status]` to the `orders` join in `payments.yml` (order_date has `fiscal_quarter` and `biweekly` custom granularities). With fix: 12 dims under joined `orders`, including `order_date_fiscal_quarter` and `order_date_biweekly`. Without fix (reverted in built artifact): 10 dims, both custom granularities dropped. Confirms exact PROD-7858 behavior.

### DIFF
--- a/packages/common/src/compiler/exploreCompiler.test.ts
+++ b/packages/common/src/compiler/exploreCompiler.test.ts
@@ -2,6 +2,7 @@ import { SupportedDbtAdapter } from '../types/dbt';
 import { CompileError } from '../types/errors';
 import { DimensionType, FieldType, friendlyName } from '../types/field';
 import { FilterOperator } from '../types/filter';
+import { TimeFrames } from '../types/timeFrames';
 import {
     ExploreCompiler,
     parseAllReferences,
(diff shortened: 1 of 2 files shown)


Spissable (approved):
First contribution, tdd, _chefs kiss_ :shipit:

PR #23512, opened 2026-05-26
fix(pivot): correct passthrough merge index for `metricsAsRows` fanout

pimmey:
Closes: PROD-7933

### Description:

In `metricsAsRows: true` mode, each input row fans out to `baseMetricsArray.length` output rows (one per metric). The passthrough dimension merge was using the raw output row index to look up the originating input row, causing it to read from the wrong input row. For example, with 2 input rows and 2 metrics producing 4 output rows, output rows 1, 2, and 3 would read from incorrect or undefined input rows — resulting in mismatched passthrough values (e.g. product images) across metric rows for the same dimension value.

The fix divides the output row index by the fanout factor (`baseMetricsArray.length`) before indexing into the original `rows` array, ensuring all metric rows expanded from the same input row correctly inherit that row's passthrough values.

The invariant warning in dev mode has also been updated to account for the fanout, so it correctly validates `allCombinedData.length === rows.length * metricsFanout` rather than a strict 1-to-1 match.

A regression test covering the 2 input rows × 2 metrics = 4 output rows case has been added, pinning all four output positions to their expected passthrough values.

### DIFF
--- a/packages/common/src/pivot/pivotQueryResults.test.ts
+++ b/packages/common/src/pivot/pivotQueryResults.test.ts
@@ -1,4 +1,10 @@
 import { type ItemsMap } from '../types/field';
+import { type ResultRow } from '../types/results';
+import {
+    SortByDirection,
+    VizAggregationOptions,
+    VizIndexType,
+} from '../visualizations/types';
 import {
     convertSqlPivotedRowsToPivotData,
     pivotQueryResults,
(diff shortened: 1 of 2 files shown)


pimmey:
* **#23512** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23607, opened 2026-05-28
feat(ai-agents): navigate to dashboard on content tool output and internal link clicks

notgiorgi:
Closes: 

### Description:

When the AI agent creates or edits a dashboard, the UI now automatically navigates to that dashboard instead of requiring the user to manually click a link.

This is implemented through a new `onToolOutput` callback that fires during streaming whenever a tool call completes. The `createContent` and `editContent` tool outputs are inspected to extract the resulting dashboard slug, which is then used to construct a navigation URL.

Navigation is handled differently depending on context:
- In the **launcher panel** (floating chat), clicking a dashboard link or receiving a dashboard tool output minimizes the chat and navigates to the dashboard using React Router with a view transition.
- On the **full agent thread page**, the same navigation happens inline via `navigateFromAgentChat`, which is passed down through the outlet context from `AgentPage`.
- In both cases, navigation is skipped if the target URL resolves to the page the user is already on (`isSamePageNavigation`).

Dashboard and chart links rendered inside assistant message bubbles (`ContentLink`) now also support an `onInternalLinkClick` callback. When provided, clicking a link triggers in-app navigation instead of opening a new tab.

Two new utilities are introduced:
- `isSamePageNavigation` — compares a target URL against the current location, with special handling for dashboard/chart route variants (e.g. `/minimal/`, `/view`, `/tabs/...`).
- `getInternalNavigationUrl` — strips the origin from internal Lightdash URLs to produce a relative path suitable for React Router.

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AgentChatAssistantBubble.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/AgentChatAssistantBubble.tsx
@@ -333,7 +333,8 @@ const AssistantBubbleContent: FC<{
     projectUuid: string;
     agentUuid: string;
     mcpServers?: AiMcpServer[];
-}> = ({ message, projectUuid, agentUuid, mcpServers }) => {
+    onInternalLinkClick?: (href: string) => void;
+}> = ({ message, projectUuid, agentUuid, mcpServers, onInternalLinkClick }) => {
     const canManageAgents = useAiAgentPermission({
         action: 'manage',
         projectUuid,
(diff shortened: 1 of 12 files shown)


notgiorgi:
* **#23607** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23617, opened 2026-05-28
fix(ai-agents): wire implicit signals to finding promotion in review judge

joaoviana:
## Problem

When `AiAgentReviewClassifier` was enabled in prod, 100% of 24 live-event judgments came back identical:

```
signal=normal_refinement
confidence=high
promotedToFinding=false
primaryRootCause=not_a_failure
```

Local testing produced varied output (3/8 promoted across `implicit_correction` / `explicit_dispute` / `semantic_layer` root causes), so the regression only surfaced once real Slack/app traffic hit the classifier.

## Root cause

The judge LLM correctly extracted `implicitSignalSources` from the evidence packet — for example, on a prod thread where the agent told the user "this project doesn't have product/SKU-level data linked to support tickets," the judge tagged `["assistant_no_answer"]` and `["next_user_retry"]` on the two prompts. **It then classified both as `normal_refinement / not_a_failure / promotedToFinding=false`.**

The judge was doing fact-extraction correctly and then throwing the facts away during classification. The prompt asks the LLM to do both jobs in one call but provides no bridge between them — the only narrative pattern it promotes is loud explicit disputes (`next_user_dispute` / explicit pushback). Quiet failures — workarounds for missing data, polite rephrasing, retries, tool errors — fall through to the prompt's "default to false for normal_refinement / acceptance / new_question" escape hatch.

Real Slack traffic is overwhelmingly quiet; the local fixtures used to validate the v1 prompt were mostly loud.

Not the model — same `claude-haiku-4-5` runs locally and in prod via `useFastModel: true` with `AI_DEFAULT_PROVIDER=anthropic`. Not the evidence packet — the judge demonstrably had the right data in hand.

## Fix

- Add ordered decision rules to the judge system prompt: if `implicitSignalSources` is non-empty (excluding `output_shape_correction` alone), promote and map the dominant signal to a specific root cause (e.g. \`assistant_no_answer\` + named missing join → \`semantic_layer\`).
- Move the "default to false" rule into a guarded clause: only fires when \`implicitSignalSources\` is empty AND the assistant answered the question directly.
- Bump \`JUDGE_PROMPT_HASH\` from \`ai-agent-review-judge-v1\` to \`v2\` so existing v1 rows in \`ai_agent_review_run\` remain attributable and we can split before/after in the admin UI.

## Regression analysis

- **Orgs with \`AiAgentReviewClassifier\` flag enabled** will start seeing real findings instead of uniform \`normal_refinement\`. Expected and desired — that is the bug. The admin queue will populate. Step 3 of the new rules is explicit about what NOT to promote (empty signals + clean answer), so this should not cause false positives on healthy turns.
- **Orgs with the flag disabled** — no behavior change; entry point is unchanged.
- **Custom roles / service accounts / cross-project** — classifier runs as a background worker per turn, not user-triggered. No role-based or cross-project impact.
- **Cost** — same model (\`useFastModel: true\` → Haiku 4.5 / gpt-5-mini), same call shape via \`generateObject\`. Prompt is ~50% longer (≈350 input tokens added). Negligible per-turn cost; output token count unchanged because schema is unchanged.
- **DB / API** — no schema change; \`judge_prompt_hash\` is a text column already. No TSOA contract change. Same enums, same return shapes.
- **Backfill / old rows** — v1 rows in \`ai_agent_review_turn_signal\` keep their old hash. Admins can re-run any prompt through the backfill path with v2 if they want a new judgment.

## Test plan

- [x] \`AiAgentReviewClassifierService\` + \`AiAgentReviewClassifierModel\` unit tests pass (12/12).
- [ ] Re-run one of the prod prompts (e.g. \`8dfe8af4-967d-4a2f-89f4-619f30431f39\`) through the admin backfill path to confirm v2 promotes correctly before enabling broadly.
- [ ] Verify in GCP \`Logger.info('AI agent review agent judged turn', ...)\` shows mixed \`primaryRootCause\` values within ~24h of rollout.

Stacked under #23618 which adds telemetry to catch this kind of silent regression next time.

### DIFF
--- a/packages/backend/src/ee/services/AiAgentReviewClassifierService.test.ts
+++ b/packages/backend/src/ee/services/AiAgentReviewClassifierService.test.ts
@@ -68,7 +68,7 @@ const makeRun = (overrides: Record<string, unknown> = {}) => ({
     organizationUuid: ORGANIZATION_UUID,
     status: 'running' as const,
     reviewAgentVersion: 'llm-judge-v1',
-    judgePromptHash: 'ai-agent-review-judge-v1',
+    judgePromptHash: 'ai-agent-review-judge-v2',
     agentConfigSnapshotHash: null,
     agentConfigSnapshot: null,
     agentConfigSnapshotAgentUpdatedAt: null,
--- a/packages/backend/src/ee/services/AiAgentReviewClassifierService.ts
+++ b/packages/backend/src/ee/services/AiAgentReviewClassifierService.ts
@@ -35,7 +35,7 @@ import { defaultAgentOptions } from './ai/agents/agentV2';
 import { getModel } from './ai/models';
 
 const REVIEW_AGENT_VERSION = 'llm-judge-v1';
-const JUDGE_PROMPT_HASH = 'ai-agent-review-judge-v1';
+const JUDGE_PROMPT_HASH = 'ai-agent-review-judge-v2';
 
 type AiAgentReviewClassifierJudge = (
     candidate: AiAgentReviewClassifierTurnCandidate,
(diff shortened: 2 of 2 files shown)


joaoviana:
* **#23618** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23618?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23617** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23618, opened 2026-05-28
chore(ai-agents): log implicit signals on review judgments

joaoviana:
## Why

When v1 of the judge prompt was silently dropping every implicit signal (root cause in #23617), the always-on \`Logger.info('AI agent review agent judged turn', ...)\` line was the only signal available in GCP — and it did not include \`implicitSignalSources\`. The bug was invisible from logs; you had to query the DB to see that the judge had extracted failure signals and then ignored them.

This adds enough fields to that existing log line so the same regression can be spotted in a single GCP query next time, without flipping \`AI_COPILOT_DEBUG_LOGGING_ENABLED=true\` in prod.

## What

Extends the existing judged-turn info log with:

- \`implicitSignalSources\` — array of signals the judge extracted.
- \`hasImplicitSignal\` — boolean, easy GCP filter.
- \`droppedImplicitSignal\` — boolean. True iff the judge identified signals AND did not promote. This is the canary: should be ~0% in healthy operation; spikes to ~100% in the failure mode the parent PR fixes.
- \`judgePromptHash\` — so v1 vs v2 rollouts can be split in GCP without a DB join.

## Regression analysis

- **No new log volume** — extends an existing line that already fires once per classified turn.
- **No PII/PHI risk** — added fields are enum values (\`assistant_no_answer\`, \`next_user_correction\`, etc.) and the prompt-version string. No user prompt content, no project content.
- **No behavior change** — purely additive telemetry; classification logic is unchanged.

## Test plan

- [x] Existing \`AiAgentReviewClassifier\` unit tests still pass (12/12).
- [ ] After rollout, GCP query \`jsonPayload.droppedImplicitSignal=true\` should return ~0 results once #23617 is enabled.

Stacked on #23617.

### DIFF
--- a/packages/backend/src/ee/services/AiAgentReviewClassifierService.test.ts
+++ b/packages/backend/src/ee/services/AiAgentReviewClassifierService.test.ts
@@ -68,7 +68,7 @@ const makeRun = (overrides: Record<string, unknown> = {}) => ({
     organizationUuid: ORGANIZATION_UUID,
     status: 'running' as const,
     reviewAgentVersion: 'llm-judge-v1',
-    judgePromptHash: 'ai-agent-review-judge-v1',
+    judgePromptHash: 'ai-agent-review-judge-v2',
     agentConfigSnapshotHash: null,
     agentConfigSnapshot: null,
     agentConfigSnapshotAgentUpdatedAt: null,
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#23618** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23617** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23617?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #23620, opened 2026-05-28
style(ai-agents): compact admin review tables for first-time scannability

joaoviana:
## Why

The admin review tables (\`/ai-agents/admin/reviews\`) felt overwhelming for first-time users — 86–92px row heights, lots of paragraph text per cell, and several redundant subtitles competing for attention. With the v2 judge (#23617) starting to populate real findings, this page is going to get a lot more traffic, so it needs to be scannable at a glance.

## What changed

Density and information architecture pass on \`AiAgentAdminReviewItemsTable.tsx\` + its CSS module. Same data — just less of it competing for attention.

### Findings tab
- Drop fixed \`h: 86\` row height; let content size the row (~38px each).
- Cell padding \`md xl\` → \`xs md\`.
- "Suggested next step: <action>" sentence-prefix → inline arrow chip (\`→ <action>\`) sitting alongside the category tags.
- "What happened" subtext clamped to 1 line (was 2).
- "Why" text in the Review note column clamped to 1 line (was 2).
- Agent cell: drop the \`Project: <name>\` subtitle. Project name moves to a tooltip on the agent badge.
- Last seen cell: drop the \`N findings\` subtitle (each row is already a finding).
- Column widths nudged tighter: Review note 430→380, Agent 220→180, Last seen 140→110.

### Signals tab
- Drop fixed \`h: 92\` row height.
- Same cell padding tightening.
- Result column: drop the redundant "Promoted by judge" / "Not promoted" subtitle (the tag already encodes promotion state). Confidence icon inlines next to the tag instead of stacking below.
- Signal column: same arrow-chip pattern as Findings.
- Turn cell: response preview clamped to 1 line.
- Agent cell: same agent-only treatment (project in tooltip).
- Reviewed cell: drop the \`finding\` / \`signal\` subtitle.

### Token + chip styling
- \`.categoryToken\` height 22px→18px, padding 3px 8px→1px 6px, border radius 5→4. Adds \`white-space: nowrap\` so multi-word tags (\`Implicit correction\`, \`domain abbreviation misinterpretation\`) don't wrap inside the token at narrow column widths.
- New \`.suggestedStep\` style for the arrow chip with a tooltip showing the full action label when it truncates.
- \`.confidenceIcon\` shrunk to 16×16.

## Regression analysis

- **Purely visual change** — no data fetching, no API shape change, no behavior change. Same hooks, same query keys, same column accessors.
[...]

PR #23624, opened 2026-05-28
feat(ai-agents): project + root-cause filters and shared FilterFacet

joaoviana:
## Why

Following the density polish in #23620, the admin reviews page still didn't feel like a real observability tool: no way to scope to a project, no way to filter by failure type, and the row hover was missing the "open thread" affordance everywhere else in the product uses.

Three jobs an admin needs to do on this page:
1. "Show me only X project's queue" — needs a project filter.
2. "Where are failures concentrated this week?" — needs distribution visible at a glance.
3. "Open the underlying conversation" — needs to feel like a row action, not a separate column.

This PR delivers all three.

### New shared component: \`FilterFacet\`

\`packages/frontend/src/components/common/FilterFacet/\`. Button + popover + checkbox list with per-option counts, dashed border idle / indigo border selected (matching existing \`ProjectsFilter\` visual language), scrollable for long lists. Generic API — works with any table.

\`\`\`tsx
<FilterFacet
    label="Project"
    icon={IconBox}
    options={[{ value, label, count }, ...]}
    selected={selectedProjectUuids}
    onChange={setSelectedProjectUuids}
/>
\`\`\`

\`count\` is optional — when omitted the option just appears without a number. Counts can be cross-filtered or absolute depending on what the caller passes.

### Reviews page now has facets

Two filters in the toolbar above the table, both showing live counts:

- **Project** filter — multi-select across all projects that currently have review items / signals.
- **Root cause** filter (Findings tab) / **Signal** filter (Signals tab) — same UI, scoped to the relevant taxonomy.

Counts are **cross-filtered**: the Project facet shows counts after the root-cause / signal filter is applied, and vice versa. So you see "if I also pick Jaffle Shop, I'll have 6 items" before you click.

\`Clear filters\` button appears next to the filters when any selection is active. The header count badge updates to \`6 of 24 items\` when filtered.

### Date + thread icon merged

The thread-preview icon used to be its own narrow column with a permanent visible icon — wasting horizontal space and feeling like a separate concept from the row. Now it lives inside the \`Last seen\` / \`Reviewed\` cell, hidden by default and revealed on row hover (\`opacity 0 → 1\` over 120ms, also visible on keyboard focus). That bought back ~96px of column width per row.

Also tightened: Agent column 180 → 140, Last seen / Reviewed 110 → 120 (slightly wider to host the icon comfortably).

### Other typography nudges from review

- \`Suggested next step\` chip — bumped to \`fw=600 / ldGray.8\` so it reads as an action.
- Secondary category token (subcategory like \`domain abbreviation misinterpretation\`) — bumped from \`ldGray.6 / fw=450\` to \`ldGray.8 / fw=500\` so the supporting tag isn't visually inert.

## Regression analysis

- **No data fetching change.** Same hooks, same query keys. All filtering is client-side over already-loaded data. With ~hundreds-of-items pages this is fine; if review queues grow into thousands of rows we'd push filters server-side via the existing \`AiAgentAdminFilters\` shape (already supports \`projectUuids\`).
- **No URL persistence yet.** Filter state is local \`useState\`. Existing \`useAiAgentAdminFilters\` hook handles URL params for the Threads / Agents tabs; wiring the reviews filters into it (so the page is shareable) is a clean follow-up.
- **No behavior change for users without filters applied.** Default view is identical to pre-PR.
- **No props / API changes** to \`useAiAgentAdminReviewItems\` or \`useAiAgentAdminReviewSignals\`.
- **Accessibility:** the merged date/icon cell preserves keyboard access — the action icon stays focusable and becomes visible on \`:focus-visible\`. Tooltip remains for screen readers.
- **Visual consistency:** \`FilterFacet\` matches the dashed/indigo treatment of the existing \`ProjectsFilter\` / \`AgentsFilter\` components, so it doesn't introduce a new design language. Migrating those two to use the new shared component is a small follow-up and would surface counts there for free.

## Follow-ups (out of scope for this PR)

- Migrate \`ProjectsFilter\` / \`AgentsFilter\` in the Threads / Agents admin tabs to use the shared \`FilterFacet\` — they keep their public API, gain counts.
- Persist reviews filter state to URL via \`useAiAgentAdminFilters\` so admins can share filtered queue links.
- Server-side filtering when item counts grow past in-memory practicality.

## Test plan

- [x] Typecheck clean (\`pnpm -F frontend typecheck\`).
- [x] Visual verification at 1440×900: Findings and Signals tabs, filter open / closed states, indigo selected state, count badge, "Clear filters" button, hover-revealed thread icon.
- [ ] Verify dark mode for the new filter button and option hover state.
- [ ] Verify keyboard nav into a hidden thread icon focuses it and reveals it (focus-visible CSS).

Standalone PR — does not depend on #23618 (telemetry, still open).

### DIFF
--- a/packages/frontend/src/components/common/FilterFacet/FilterFacet.module.css
+++ b/packages/frontend/src/components/common/FilterFacet/FilterFacet.module.css
@@ -1,0 +1,38 @@
+.filterButton {
+    border: 1px dashed var(--mantine-color-ldGray-3);
+    background-color: transparent;
+    text-overflow: ellipsis;
+    box-shadow: var(--mantine-shadow-subtle);
+}
+
+.filterButton:hover {
+    background-color: var(--mantine-color-ldGray-0);
+    transition: background-color var(--mantine-other-transitionDuration)
+        var(--mantine-other-transitionTimingFunction);
+}
+
+.filterButtonSelected {
+    border: 1px solid var(--mantine-color-indigo-2);
+    background-color: var(--mantine-color-indigo-0);
+    text-overflow: ellipsis;
+    box-shadow: var(--mantine-shadow-subtle);
+}
+
+.filterButtonSelected:hover {
+    background-color: var(--mantine-color-ldGray-0);
+    transition: background-color var(--mantine-other-transitionDuration)
+        var(--mantine-other-transitionTimingFunction);
+}
+
+.option {
+    display: block;
+    width: 100%;
+    border-radius: 4px;
+}
+
+.option:hover {
+    background-color: light-dark(
+        var(--mantine-color-ldGray-0),
+        var(--mantine-color-dark-6)
+    );
+}
--- a/packages/frontend/src/components/common/FilterFacet/FilterFacet.tsx
+++ b/packages/frontend/src/components/common/FilterFacet/FilterFacet.tsx
@@ -1,0 +1,185 @@
+import {
+    Badge,
+    Button,
+    Checkbox,
+    Group,
+    Popover,
+    ScrollArea,
+    Stack,
+    Text,
+    Tooltip,
+    UnstyledButton,
+} from '@mantine-8/core';
+import { IconChevronDown } from '@tabler/icons-react';
+import { type Icon as TablerIcon } from '@tabler/icons-react';
+import MantineIcon from '../MantineIcon';
+import classes from './FilterFacet.module.css';
+
+export type FilterFacetOption = {
+    value: string;
+    label: string;
+    count?: number;
+};
+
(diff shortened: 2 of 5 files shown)


joaoviana:
* **#23624** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23624?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23624?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23634, opened 2026-05-29
feat: replace hand-maintained public email domain list with `free-email-domains` package

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7962/replace-hardcoded-public-domains

CleanShot 2026-05-29 at 09 57 19

Intentional regression: we also limit the domains you can add on allowed list of domains: 

CleanShot 2026-05-29 at 09 58 32

zoho.com is free webmail, so allowing it as an auto-join domain would let anyone with a @zoho.com address join your org. Before this change zoho.com
   wasn't in the 51-domain list, so it would have been accepted — that was the hole. Now it's correctly refused, both in the frontend (the inline
  error + disabled Update you see) and the backend.

  This is not the regression I flagged. The regression is a different scenario: an org that already had one of these domains saved before the change,
  then tries to save the panel for an unrelated reason (e.g. changing the Default role) and gets blocked because the full list is re-validated. What
  you're doing here — adding a brand-new public-provider domain and being stopped — is the correct, desired behavior.

### Description:

Replaces the hand-maintained `PUBLIC_EMAIL_PROVIDER_DOMAINS` list (~50 entries) with the Kickbox [`free-email-domains`](https://github.com/nicehash/free-email-domains) package (~10k entries). This significantly expands coverage of public/consumer email providers that cannot be legitimately owned by a single organization, catching providers like `proton.me`, `protonmail.com`, `gmx.com`, `zoho.com`, `qq.com`, `163.com`, `yandex.com`, and `fastmail.com` that the old list missed.

A small `ADDITIONAL_PUBLIC_PROVIDER_DOMAINS` list is kept alongside the package to cover:
- Corporate provider roots (`google.com`, `microsoft.com`, `onmicrosoft.com`) that aren't free webmail but still can't be owned by a single customer on a shared instance.
- Consumer providers the maintained list happens to omit (e.g. `sfr.fr`, carried over from the previous hand-maintained list to avoid regression).

The domain check remains a vendored static set — no runtime fetches. It is a best-effort heuristic; a brand-new provider may slip through until the `free-email-domains` package updates.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -43,6 +43,7 @@
     "dependency-graph": "0.11.0",
     "echarts": "6.0.0",
     "fast-json-patch": "3.1.1",
+    "free-email-domains": "1.9.16",
     "handlebars": "4.7.9",
     "js-yaml": "4.1.1",
     "liquidjs": "10.26.0",
--- a/packages/common/src/utils/email.test.ts
+++ b/packages/common/src/utils/email.test.ts
@@ -105,11 +105,26 @@ describe('isPublicEmailProviderDomain', () => {
         expect(isPublicEmailProviderDomain('google.com')).toBe(true);
         expect(isPublicEmailProviderDomain('microsoft.com')).toBe(true);
         expect(isPublicEmailProviderDomain('onmicrosoft.com')).toBe(true);
+        // sfr.fr is absent from free-email-domains; kept via the additional
+        // list so the swap doesn't regress the old hand-maintained coverage.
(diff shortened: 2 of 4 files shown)


rephus:
* **#23634** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23634?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23638, opened 2026-05-29
fix: use advisory lock to deduplicate concurrent forced-slug creates (PROD-7883)

rephus:
Closes: https://linear.app/lightdash/issue/PROD-7883/investigate-duplicate-slugs-on-prod

Complete:
  - ✅ Root cause confirmed (concurrent content-as-code upsert race + no DB constraint)
  - ✅ Fix implemented (advisory lock + dedupe in DashboardModel.create, createSavedChart, SlugUtils.ts)
  - ✅ Verified live: 2 and 6 concurrent same-slug upserts now produce 1 row (was 2 before); sequential upsert still updates correctly
  - ✅ Typecheck passes, lint clean, 30 existing tests pass, test data cleaned up
  - 
### Description:

When content-as-code or promotion flows create charts and dashboards, they force a specific slug from the YAML/source rather than generating a unique one. Because the create-vs-update decision is a non-atomic find-then-create, two concurrent uploads of the same slug could both decide to insert, resulting in duplicate slugs with no DB unique constraint to prevent it.

This fix introduces a transaction-scoped Postgres advisory lock (`acquireProjectSlugLock`) keyed on `(projectUuid, slug)`. When a forced slug is used during chart or dashboard creation, the lock is acquired before the existence check. A second concurrent caller with the same key will block until the first transaction commits, at which point the newly created row is visible and the second caller can return the existing record instead of inserting a duplicate. The lock is automatically released when the transaction ends.

### DIFF
--- a/packages/backend/src/models/DashboardModel/DashboardModel.ts
+++ b/packages/backend/src/models/DashboardModel/DashboardModel.ts
@@ -80,7 +80,10 @@ import { SpaceTableName } from '../../database/entities/spaces';
 import { UserTable, UserTableName } from '../../database/entities/users';
 import { DbValidationTable } from '../../database/entities/validation';
 import Logger from '../../logging/logger';
-import { generateUniqueSlug } from '../../utils/SlugUtils';
+import {
+    acquireProjectSlugLock,
+    generateUniqueSlug,
+} from '../../utils/SlugUtils';
 import { ContentVerificationModel } from '../ContentVerificationModel';
 import { SpaceModel } from '../SpaceModel';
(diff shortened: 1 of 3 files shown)


rephus:
* **#23638** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23638?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23638?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23641, opened 2026-05-29
fix(content-as-code): add view scope to split read from write (PROD-7379)

rephus:
## Summary

Closes [PROD-7379](https://linear.app/lightdash/issue/PROD-7379). Lets a custom role **download** production content-as-code without also being able to **upload** to it — closing the "a mis-targeted `lightdash upload` overwrites prod" gap.

Today `manage:ContentAsCode` bundles download (read) + upload (write) into one scope. Anyone who needs to pull production content for local dev also gets push rights to that project. This splits read from write so write access can be restricted via custom roles, while reads stay open.

## Approach

- **NEW `view:ContentAsCode`** — gates the three download endpoints (`getDashboards` / `getCharts` / `getSqlCharts`). CASL `manage` implies `view`, so developers / admins / service accounts keep working with **zero behavior change**.
- **`manage:ContentAsCode`** — continues to gate the upload (upsert) endpoints; unchanged power.
- **Editors** (project + org) gain `view:ContentAsCode` so non-developers can pull YAML for diffing/inspection without push rights.

A custom role can now grant `view:ContentAsCode` only on a production project: the holder can download/diff prod YAML but cannot push content-as-code back. Writes get gated to a controlled promotion path while reads stay open.

**No data migration required** — additive scope; `manage:ContentAsCode` is unchanged (not renamed/removed), so existing custom roles keep working.

> **Note on previews / `@self`:** the ticket also floated a `manage:ContentAsCode@self` scope so users could write content-as-code to their *own* previews. That was dropped from this PR: owning a preview requires `create:Project@preview` (Developer tier only), so `@self` would be a no-op for Editors and is redundant for Developers (who already hold broad `manage`). Left for a follow-up if we decide to extend preview ownership to editors.

## Changes
- `scopes.ts` — add `view:ContentAsCode`
- `roleToScopeMapping.ts` — editor gains `view:ContentAsCode` (also added to the enterprise set)
- `projectMemberAbility.ts` / `organizationMemberAbility.ts` — editor `view:ContentAsCode` rule
- `CoderService.ts` — the three download endpoints now check `view` instead of `manage`

## Tests
- Integration tests in `scopeAbilityBuilder.test.ts`: view-only (download yes / upload no), full manage (download + upload, `manage`⊇`view`), and enterprise gating.
- Verified: common authorization suites (503 pass incl. role↔scope parity + vocabulary coverage), CoderService (23 pass), common + backend typecheck, lint.
- **Verified live** on a local enterprise instance: a `member` user with a custom role holding only `view:ContentAsCode` could download charts/dashboards (200) but got **403** uploading; admin (`manage`) retained both.

## Regression risk
**LOW** — additive at the CASL layer (`manage` ⊇ `view`), no migration, no API shape change, no frontend gating (content-as-code is CLI-driven). The only behavior change is intended: editors can now download content-as-code.

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.ts
+++ b/packages/backend/src/services/CoderService/CoderService.ts
@@ -685,7 +685,7 @@ export class CoderService extends BaseService {
         const auditedAbility = this.createAuditedAbility(user);
         if (
             auditedAbility.cannot(
-                'manage',
+                'view',
                 subject('ContentAsCode', {
                     projectUuid: project.projectUuid,
                     organizationUuid: project.organizationUuid,
@@ -821,7 +821,7 @@ export class CoderService extends BaseService {
         const auditedAbility = this.createAuditedAbility(user);
         if (
             auditedAbility.cannot(
-                'manage',
+                'view',
                 subject('ContentAsCode', {
                     projectUuid: project.projectUuid,
                     organizationUuid: project.organizationUuid,
@@ -954,7 +954,7 @@ export class CoderService extends BaseService {
         const auditedAbility = this.createAuditedAbility(user);
         if (
             auditedAbility.cannot(
-                'manage',
+                'view',
                 subject('ContentAsCode', {
                     projectUuid: project.projectUuid,
                     organizationUuid: project.organizationUuid,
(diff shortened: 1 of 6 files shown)


rephus:
* **#23641** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23641?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23641?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23647, opened 2026-05-29
fix: prevent AI agent form crash with custom group roles

IrakliJani:
## Problem

Opening the add/edit **AI Agent** form white-screens when a project has a **custom role assigned to a group** that contains an org **Member**:

```
Project role <uuid> does not match Organization roles
```

Reported via customer support; reproduces on current `main`.

## Root cause

Custom roles are stored as `role_uuid` and **coalesced into the legacy `role` string field** (`ProjectModel.getProjectGroupAccesses`, `GroupService`). The system-role helpers in `packages/common/src/utils/projectMemberRole.ts` are exhaustive over the 5 `ProjectMemberRole` enum values, so a custom-role UUID hits `assertUnreachable` and throws.

- **Frontend** (the reported white-screen): `useProjectUsersWithRoles` resolves group roles client-side and cast the UUID string into the enum (`as ProjectMemberRole`), so the throw happens inside a `useMemo` during render → ErrorBoundary.
- **Backend**: the same helper also 500s `OrganizationService.getMembers` when listing members with `includeGroups`.

## Fix

- **`projectMemberRole.ts`** (single source of truth — fixes both surfaces): add an `isProjectMemberRole` type guard; `getHighestProjectRole` normalizes an unrankable custom-role UUID to `VIEWER`; `convertProjectRoleToSpaceRole` / `convertProjectRoleToOrganizationRole` fall back to `VIEWER` instead of throwing.
- **`useProjectUsersWithRoles.ts`**: drop the unsafe `as ProjectMemberRole` cast — custom-role UUIDs resolve to `undefined`, so the user keeps their real org/project role.
- Unit tests covering the UUID path.

Permission **enforcement is unaffected** — it runs through the scope-based path (`scopeAbilityBuilder.ts`), not these display helpers. This cannot over-grant. The only effect is cosmetic: in these legacy listings a custom role shows as the user's real role / "Viewer" rather than its custom name.

## Follow-up (not in this PR)

Type-honesty fix: model the role as a discriminated union (system vs custom) or add a `roleUuid` field, thread `roleUuid` + name through `getProjectGroupAccesses` / `getMembers`, and render the custom-role **name** in the access list. That also closes the related display bug (PROD-7538).

Closes PROD-7999

### DIFF
--- a/packages/common/src/utils/projectMemberRole.ts
+++ b/packages/common/src/utils/projectMemberRole.ts
@@ -1,4 +1,3 @@
-import { NotSupportedError } from '../types/errors';
 import { OrganizationMemberRole } from '../types/organizationMemberProfile';
 import {
     ProjectMemberRole,
@@ -12,16 +11,6 @@ import {
 } from '../types/projectMemberRole';
 import { SpaceMemberRole } from '../types/space';
 import assertUnreachable from './assertUnreachable';
-
-const projectMemberRoleValues = new Set<string>(
-    Object.values(ProjectMemberRole),
-);
-
-// `ProjectGroupAccess.role` (and other legacy role fields) can hold a custom-role
-// UUID instead of a system `ProjectMemberRole`. Use this guard before passing a
-// role into the system-role helpers below, which only understand the 5 enum values.
-export const isProjectMemberRole = (role: string): role is ProjectMemberRole =>
-    projectMemberRoleValues.has(role);
 
 export const convertOrganizationRoleToProjectRole = (
     organizationRole: OrganizationMemberRole,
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/projectMemberRole.ts
```diff
@@ -114,6 +133,9 @@ export const convertProjectRoleToSpaceRole = (
 export const convertProjectRoleToOrganizationRole = (
     projectRole: ProjectMemberRole,
 ): OrganizationMemberRole => {
+    if (!isProjectMemberRole(projectRole)) {
+        return OrganizationMemberRole.VIEWER;
```
rephus: this should probably be Member...

packages/common/src/utils/projectMemberRole.ts
```diff
@@ -92,6 +108,9 @@ export const getHighestSpaceRole = (
 export const convertProjectRoleToSpaceRole = (
     projectRole: ProjectMemberRole,
 ): SpaceMemberRole => {
+    if (!isProjectMemberRole(projectRole)) {
+        return SpaceMemberRole.VIEWER;
```
rephus: will this gran access to all private spaces ?

IrakliJani: good point, I'll check. I can also throw in these weird cases

packages/common/src/utils/projectMemberRole.ts
IrakliJani: good point. I can throw when this happens

IrakliJani: just to isolate frontend only fix

PR #23656, opened 2026-05-29
fix(ai-agents): replace violet color with ldGray in AgentSelector and AgentsRouterPage

tatianainama:
### Description

Tones down the purple accents on the "Ask AI" router page so the generic entry point reads calm, while keeping violet where it carries real meaning.

The page is the generic "Ask AI" landing state, so the violet accents were pulling too much attention and gave the "Auto" mode (a non-choice) the visual weight of a primary action. Neutralizing those accents fits the page better.

#### Decision: neutral by default, violet only as a selection signal

The violet was doing two different jobs that got conflated. As decoration on the landing state it was over-weighted, but inside the picker it is a genuine selection signal, the one place an accent color earns its keep. So the split is:

**Neutralized to `ldGray`:**
- Hero avatar + sparkles icon
- "Thinking" routing dots
- Auto mode chip, label, and sparkles in the shared `AgentSelector`
- Dropdown selection checkmarks

**Kept violet (intentionally):**
- The "Recommended" candidate in the agent picker: border, hover fill, description tint, chevron, settle glow, and the "Recommended" badge

Violet stays on the recommended candidate because it is a real selection signal, and because it reads better in dark mode: violet's light variant has enough chroma to lift off the dark surface where flat gray collapses into the background.

#### Other
- Fixed a typo in the Auto dropdown copy: "besy-fit" to "best-fit"

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/AgentSelector/AgentSelector.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/AgentSelector/AgentSelector.tsx
@@ -102,11 +102,11 @@ export const AgentSelector = ({
                 >
                     <Group gap={6} wrap="nowrap" align="center" w="100%">
                         {isAuto ? (
-                            <Avatar size={22} color="violet" radius="xl">
+                            <Avatar size={22} color="ldGray" radius="xl">
                                 <MantineIcon
                                     icon={IconSparkles}
                                     size="sm"
-                                    color="violet.6"
+                                    color="ldGray.6"
                                 />
                             </Avatar>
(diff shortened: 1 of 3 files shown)


tatianainama:
* **#23656** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23674, opened 2026-05-29
test: cover DB feature-flag resolution priority and defaults (GLITCH-312)

pimmey:
Closes:

### Description:

Expands the `FeatureFlagModel` test suite with a lightweight fake database layer that simulates the user-override → org-override → flag-default precedence chain without requiring a real database connection.

New test coverage includes:

- **Database resolution**: verifies that a user override takes priority over an org override, that the org override is used when no user override exists, and that the flag's `default_enabled` value is used as a final fallback.
- **`default_enabled` vs env fallback**: confirms that an explicit `default_enabled: false` in the database wins over a config-based fallback of `true`, while `default_enabled: null` correctly falls through to the config value.
- **Error resilience**: adds a `Logger.warn` spy to the existing DB-error tests so the expected warning is asserted on rather than silently swallowed, and keeps the test output clean.

### DIFF
--- a/packages/backend/src/models/FeatureFlagModel/FeatureFlagModel.test.ts
+++ b/packages/backend/src/models/FeatureFlagModel/FeatureFlagModel.test.ts
@@ -2,6 +2,13 @@ import { FeatureFlags } from '@lightdash/common';
 import { Knex } from 'knex';
 import { lightdashConfigMock } from '../../config/lightdashConfig.mock';
 import { LightdashConfig } from '../../config/parseConfig';
+import {
+    DbFeatureFlag,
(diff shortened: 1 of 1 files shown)


pimmey:
* **#23674** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23674?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23674?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23680, opened 2026-05-29
feat(pivot): replace group-header rows with `rowSpan`-merged dimension cells (PROD-7938)

pimmey:
Closes: PROD-7938

### Description:

Replaces the TanStack group-header row rendering in the pivot table's `showRowGrouping` mode (when subtotals are off) with HTML `rowSpan`-merged dimension cells, giving the table a real spreadsheet-pivot appearance.

**Before:** Each grouped dimension value rendered as a separate group-header row above its data rows, leaving dimension cells blank on data rows, showing `∅`/`-` in empty data cells, and producing non-sequential row numbers.

**After:** The grouped dimension value renders once per consecutive run, top-aligned and vertically spanning its block via `rowSpan`. Leaf dimension and metric cells render per-row as before. Empty pivot data cells are fully blank (no `∅`/`-`). Row numbers are sequential with no gaps.

Key changes:

- **New pure helper `getRowSpanMerges.ts`** computes per-row/per-column `{ isBlockStart, rowSpan }` using nested consecutive run-length encoding. A block starts when the prefix tuple of grouped dimension values changes from the previous row, ensuring outer dimensions span wider than inner ones and an inner value repeating under a different outer value is never incorrectly merged.
- **`getGroupedDimColumnIds`** extracts the mergeable dimension columns (all index dims except the leaf) and is shared between the span computation and the TanStack grouping setup to keep them in sync.
- **TanStack grouping is no longer engaged** in grouping-only mode (`showRowGrouping && !showSubtotals`). The flat row model is used directly, keeping `rowIndex` aligned with all per-row lookups (conditional formatting, cell menus, etc.). The subtotals path is untouched.
- **Virtualization is disabled** in merge mode so `rowSpan` blocks remain contiguous — a block-start row is never unmounted while its absorbed siblings are visible.
- Absorbed rows return `null` from the cell renderer (no `<td>` emitted); block-start cells receive `rowSpan` and the `mergedDimCell` top-align CSS class.

### DIFF
--- a/docs/superpowers/plans/2026-05-29-prod-7938-pivot-rowspan-merge.md
+++ b/docs/superpowers/plans/2026-05-29-prod-7938-pivot-rowspan-merge.md
@@ -1,0 +1,605 @@
+# Pivot `rowSpan`-Merged Dimension Cells Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Replace the grouping-only `showRowGrouping` rendering in the pivot table (separate TanStack group-header rows) with top-aligned `rowSpan`-merged dimension cells, blank empty cells, and no virtualization in that mode.
+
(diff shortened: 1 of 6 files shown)


pimmey:
* **#23680** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23680?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23702, opened 2026-06-01
feat: add per-org scheduled-delivery link expiry settings with per-channel overrides and transparent persistent-URL escalation (PROD-7217)

rephus:
Closes: [https://linear.app/lightdash/issue/PROD-7217/add-organization-level-scheduled-delivery-settings](https://linear.app/lightdash/issue/PROD-7217/add-organization-level-scheduled-delivery-settings)

Only show  enabled integrations, like google chat (but not ms teams)

### Description:

Adds organization-level **Exporting** settings under **Settings → Organization → Exporting**, giving admins control over how long scheduled-delivery download links (email, Slack, Microsoft Teams, Google Chat) remain valid — without requiring changes to instance-wide environment variables.

**Per-org expiry configuration**

Admins can set a base "Download link expiry" (in days) that all channels inherit, plus optional per-channel overrides for Email, Slack, and Microsoft Teams. These are stored as new nullable columns on `organization_settings`. A `null` value means "inherit", so orgs with no configuration continue to behave exactly as before, falling back to the existing env var defaults (`PERSISTENT_DOWNLOAD_URL_EXPIRATION_SECONDS` and its per-channel variants).

Expiry resolution per channel follows this precedence:

```
org channel → org base → env channel → env base
```

**Transparent raw S3 vs. persistent URL selection**

`PersistentDownloadFileService.createPersistentUrl()` now automatically switches to the persistent-download-URL system when the requested expiry exceeds AWS's 7-day SigV4 presigned URL hard limit — even if `PERSISTENT_DOWNLOAD_URLS_ENABLED` is not set. This prevents the previous footgun where a >7-day raw link would silently 403. The `PERSISTENT_DOWNLOAD_URLS_ENABLED` default remains `false`, so existing behavior for links ≤ 7 days is unchanged.

**Frontend**

A new Exporting panel renders a base expiry input (days) and an optional per-channel section (Email, Slack, Microsoft Teams). Each per-channel input shows a clear (✕) affordance to reset back to inheriting the base. The panel is accessible to org admins via a new "Exporting" nav item in the Organization settings sidebar.

**Validation**

The settings API rejects expiry values that are not positive integers. There is no maximum — values over 7 days are valid and transparently route through the persistent-download-URL system. The frontend applies a 1–365 day guardrail.

### DIFF
--- a/packages/backend/src/database/migrations/20260601072726_add_exporting_settings_to_organization_settings.ts
+++ b/packages/backend/src/database/migrations/20260601072726_add_exporting_settings_to_organization_settings.ts
@@ -5,6 +5,7 @@ const BASE_COLUMN = 'scheduled_delivery_expiration_seconds';
 const EMAIL_COLUMN = 'scheduled_delivery_expiration_seconds_email';
 const SLACK_COLUMN = 'scheduled_delivery_expiration_seconds_slack';
 const MSTEAMS_COLUMN = 'scheduled_delivery_expiration_seconds_msteams';
+const GOOGLECHAT_COLUMN = 'scheduled_delivery_expiration_seconds_googlechat';
 
 export async function up(knex: Knex): Promise<void> {
     await knex.schema.alterTable(ORGANIZATION_SETTINGS_TABLE, (table) => {
@@ -16,6 +17,7 @@ export async function up(knex: Knex): Promise<void> {
         table.integer(EMAIL_COLUMN).nullable();
         table.integer(SLACK_COLUMN).nullable();
         table.integer(MSTEAMS_COLUMN).nullable();
+        table.integer(GOOGLECHAT_COLUMN).nullable();
     });
 }
 
@@ -25,5 +27,6 @@ export async function down(knex: Knex): Promise<void> {
         table.dropColumn(EMAIL_COLUMN);
         table.dropColumn(SLACK_COLUMN);
         table.dropColumn(MSTEAMS_COLUMN);
+        table.dropColumn(GOOGLECHAT_COLUMN);
     });
 }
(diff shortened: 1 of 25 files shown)


rephus:
* **#23702** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23702?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23702?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/database/migrations/20260601072726_add_exporting_settings_to_organization_settings.ts
```diff
@@ -0,0 +1,32 @@
...
+const GOOGLECHAT_COLUMN = 'scheduled_delivery_expiration_seconds_googlechat';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(ORGANIZATION_SETTINGS_TABLE, (table) => {
+        // Per-org base override (seconds) for how long scheduled-delivery
+        // download links stay valid. Nullable with no stored default, so
+        // NULL/absent inherits PERSISTENT_DOWNLOAD_URL_EXPIRATION_SECONDS.
+        table.integer(BASE_COLUMN).nullable();
+        // Optional per-channel overrides; NULL/absent inherits the base above.
+        table.integer(EMAIL_COLUMN).nullable();
```
rephus, line 17: we could all make this a jsonblob instead of separate columns, but I think this is  going to scale better, even if we need to add more platforms in the future

PR #23712, opened 2026-06-01
feat(ai-router): respect user's default agent

tatianainama:
Closes: [PROD-8027](https://linear.app/lightdash/issue/PROD-8027/respect-users-default-agent-over-the-ai-router-on-the-agents-page)

### Description

The AI router auto-selects the best-fit agent based on the user's prompt, and it had become the default landing experience on `/ai-agents` — overriding a user's explicitly chosen default agent. This PR makes the default agent preference take precedence over the router's auto-selection, while keeping the router reachable on demand.

### What changed

The `/ai-agents` index (`AgentsWelcome`) now decides where to start in this priority order:

1. **No agents** — welcome / empty state.
2. **Exactly one agent** — open it.
3. **Valid default agent preference** — open it, _unless_ the user explicitly asked for Auto. A preference pointing at a deleted agent is ignored.
4. **Router enabled** — show the auto-router (`AgentsRouterPage`).
5. **Otherwise** — open the first agent.

Previously the router was checked before the preference, so an enabled router always won.

### Keeping "Auto" reachable

The agent selector's **Auto** option navigates to the bare `/ai-agents` index, so the new precedence would have made Auto unreachable for anyone with a default agent (it'd bounce straight back to their agent). To fix this:

- Picking **Auto** now sets a `?routing=auto` search param, which `AgentsWelcome` reads to force the router even when a default preference exists.
- Navigating to a specific agent strips the param so it doesn't leak into thread URLs.
- A search param (rather than transient router state) is used so the router survives a page refresh.

### Files

- `AgentsWelcome.tsx` — preference-before-router precedence; `forceRouter` override; ignores stale preferences.
- `AgentSelector.tsx` — sets `routing=auto` on explicit Auto; strips it for agent navigation.
- `AgentSelectorUtils.tsx` — shared `AI_ROUTING_SEARCH_PARAM` / `AI_ROUTING_AUTO_VALUE` constants.

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/AgentSelector/AgentSelector.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/AgentSelector/AgentSelector.tsx
@@ -20,7 +20,12 @@ import { LightdashUserAvatar } from '../../../../../components/Avatar';
 import MantineIcon from '../../../../../components/common/MantineIcon';
 import { useAiRouterConfig } from '../../hooks/useAiRouter';
 import styles from './AgentSelector.module.css';
-import { getAgentOptions, type Agent } from './AgentSelectorUtils';
+import {
+    AI_ROUTING_AUTO_VALUE,
+    AI_ROUTING_SEARCH_PARAM,
+    getAgentOptions,
+    type Agent,
+} from './AgentSelectorUtils';
 
 type Props = {
     agents: Agent[];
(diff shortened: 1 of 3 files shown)


tatianainama:
* **#23712** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23747, opened 2026-06-01
feat: add project-level default agent setting

nanic:
Closes: [#21858](https://github.com/lightdash/lightdash/issues/21858)

### Description:

Adds a project-level default AI agent that admins configure under **Settings → Ask AI → Agents**. When a user has no personal default, the app uses the project default for agent selection, landing routes, and the home AI search box.

Resolution order: **user default → project default → first accessible agent**. Deleted or inaccessible defaults fall back automatically.

Screenshot 2026-06-01 at 3 43 10 PM

### DIFF
--- a/packages/backend/src/controllers/projectController.ts
+++ b/packages/backend/src/controllers/projectController.ts
@@ -35,6 +35,7 @@ import {
     SqlChartAsCode,
     UpdateDefaultUserSpaces,
     UpdateMetadata,
+    UpdateProjectDefaultAgent,
     UpdateProjectMember,
     UserWarehouseCredentials,
     type ApiCalculateSubtotalsResponse,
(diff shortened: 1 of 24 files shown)


tatianainama:
Hey @nanic 👋 thanks for taking the time to put this together

Looking at the diff, I don't think we can take it in this shape. The change reaches into our OSS core (projects table, controller and types) to store an AI agent reference. AI agents are an enterprise feature, so coupling them into the core project model crosses a boundary we work hard to keep clean, and it would affect non-EE users who don't have agents at all

For what you're after here, I think [AI Router](https://docs.lightdash.com/guides/ai-agents/ai-router) already gets you most of the way. You can enable it for your project and add a custom routing instruction that points users at a specific agent. The AI Router was shipped in version `0.3061.0` if you want to try it out!

PR #23751, opened 2026-06-02
fix(ai-agents): when agent modifies a dashboard, properly refresh the tiles/tabs

notgiorgi:
Closes: 

### Description:

Fixes a bug where navigating to a dashboard via an AI Copilot content link was being incorrectly suppressed when the current URL contained the dashboard's UUID, even though the link used the dashboard's slug (or vice versa). The check now uses both UUID and slug to determine whether the user is already on the target dashboard, and also validates that the project matches.

The dashboard tab selection logic has been updated so that switching between tabs no longer resets the active tab to the first tab when the URL doesn't specify one — the current tab is now preserved in that case.

The AI agent change handling logic in `DashboardAiAgentContextBridge` has been extracted into a pure `planDashboardAiAgentChanges` function, making it independently testable. The `scrollToChartTile` helper no longer attempts to switch tabs itself, since tab switching is now handled separately via the URL-driven tab selection.

A shared `dashboardRoutes` utility has been introduced with `isSameLocation` and `isSameDashboardRoute` helpers, replacing ad-hoc pathname matching logic that existed in multiple places.

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/ContentLink.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/ChatElements/ContentLink.tsx
@@ -33,11 +33,6 @@ type ContentLinkProps = {
     onDashboardLinkClick?: (url: string) => void;
 };
 
-const getDashboardRouteKey = (pathname: string) => {
-    const match = pathname.match(/^\/projects\/([^/]+)\/dashboards\/([^/]+)/);
-    return match ? `${match[1]}/${match[2]}` : null;
-};
-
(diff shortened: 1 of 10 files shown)


notgiorgi:
* **#23751** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23751?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23751?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23768, opened 2026-06-02
feat: scope DuckDB S3 pre-aggregate materialization access by org and project

rephus:
Closes: https://linear.app/lightdash/issue/PROD-8068/fix-s3-leak-on-pre-aggregates

  - Materializations are stored under pre-aggregates/<orgUuid>/<projectUuid>/.
  - DuckDB S3 secrets use SCOPE.
  - Active materialization URIs outside the current org/project scope are rejected.
  - Worker-side scoped-storage references are checked before execution.
  - Tests cover backend routing, materialization prefixing, URI rejection, and DuckDB SCOPE.
  - I also validated it live: scoped DuckDB could read the valid object, could not read a copied fake cross-org object, while unscoped DuckDB could
    read that same fake object.

### Description:

Adds org/project-scoped S3 access controls for pre-aggregate materialization files. Pre-aggregate result files are now stored under a `pre-aggregates/{organizationUuid}/{projectUuid}/` path prefix, and DuckDB S3 secrets are configured with a matching `SCOPE` clause to restrict access to only that path.

When resolving a pre-aggregate query, the materialization URI is validated to ensure it falls within the expected org/project scope. If a scoped URI is detected that belongs to a different org or project, resolution is rejected with an `INVALID_MATERIALIZATION_URI` reason rather than allowing cross-tenant data access.

DuckDB warehouse clients are now cached per scope (rather than as a single global instance), so scoped and unscoped clients can coexist without interfering with each other.

### DIFF
--- a/packages/backend/src/ee/services/AsyncQueryService/PreAggregateStrategy.ts
+++ b/packages/backend/src/ee/services/AsyncQueryService/PreAggregateStrategy.ts
@@ -145,12 +145,14 @@ export class PreAggregateStrategy implements IPreAggregateStrategy {
     }
 
     async resolveExecution({
+        organizationUuid,
         projectUuid,
         queryUuid,
         warehouseQuery,
         preAggregationRoute,
         resolveArgs,
     }: {
+        organizationUuid: string;
         projectUuid: string;
         queryUuid: string;
         warehouseQuery: string;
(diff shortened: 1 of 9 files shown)


rephus:
* **#23778** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23768** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23768?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23788, opened 2026-06-02
fix(ai): add timeout to MCP client connect and tool-discovery calls

ZeRego:
Closes: PROD-8039

### Description:

MCP client calls had no timeout, so a hanging MCP server could stall admin create/refresh requests and end-user chat requests for up to ~5 minutes (the undici default) with no clear error. This adds an `MCP_CONNECTION_TIMEOUT_MS` config (default 20s) and wraps both the connection and the `tools()`/`listTools()` calls at all three call sites (`testConnection`, `listTools`, `resolveTools`) in a timeout helper, using a per-server timeout in `resolveTools` so one slow server doesn't block the rest. On timeout the server is marked `error` with a user-facing message, and a client that connects after the timeout fires is closed to avoid leaking a transport connection. Includes unit tests covering the timeout and late-close behaviour.

### DIFF
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -1259,6 +1259,7 @@ export type LightdashConfig = {
     mcp: {
         enabled: boolean;
         runSqlMaxLimit: number;
+        timeoutMs: number;
     };
     customRoles: {
         enabled: boolean;
@@ -2332,6 +2333,10 @@ export const parseConfig = (): LightdashConfig => {
             runSqlMaxLimit:
                 getIntegerFromEnvironmentVariable('MCP_RUN_SQL_MAX_LIMIT') ||
                 AI_DEFAULT_MAX_QUERY_LIMIT,
+            timeoutMs:
+                getIntegerFromEnvironmentVariable(
+                    'MCP_CONNECTION_TIMEOUT_MS',
+                ) || 20_000,
         },
         customRoles: {
             enabled: process.env.CUSTOM_ROLES_ENABLED === 'true',
--- a/packages/backend/src/config/parseConfig.ts
+++ b/packages/backend/src/config/parseConfig.ts
@@ -987,6 +987,9 @@ export const getAiConfig = () => ({
         getFloatFromEnvironmentVariable(
             'AI_VERIFIED_ANSWER_SIMILARITY_THRESHOLD',
         ) ?? 0.6,
+    mcpConnectionTimeoutMs:
+        getIntegerFromEnvironmentVariable('AI_COPILOT_MCP_TIMEOUT_MS') ||
+        20_000,
 });
 
 export type LoggingConfig = {
@@ -1259,7 +1262,6 @@ export type LightdashConfig = {
     mcp: {
         enabled: boolean;
         runSqlMaxLimit: number;
-        timeoutMs: number;
     };
     customRoles: {
         enabled: boolean;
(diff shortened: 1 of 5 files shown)


packages/backend/src/config/parseConfig.ts
```diff
@@ -2339,6 +2333,10 @@ export const parseConfig = (): LightdashConfig => {
             runSqlMaxLimit:
                 getIntegerFromEnvironmentVariable('MCP_RUN_SQL_MAX_LIMIT') ||
                 AI_DEFAULT_MAX_QUERY_LIMIT,
+            timeoutMs:
+                getIntegerFromEnvironmentVariable(
+                    'MCP_CONNECTION_TIMEOUT_MS',
+                ) || 20_000,
```
notgiorgi: this config should be under `ai.copilot`, `mcp` is for Lightdash MCP

packages/backend/src/ee/services/ai/AiAgentMcpRuntimeClient.ts
```diff
@@ -531,14 +542,73 @@ export const createHttpMcpClient = async (
...
+        );
+    });
+
+export const createHttpMcpClientWithTimeout = (
+    mcpServer: McpServerConnectionArgs,
+    timeoutMs: number,
+    onUncaughtError?: (error: unknown) => void,
+): Promise<MCPClient> =>
+    withMcpTimeout(
+        createHttpMcpClient(mcpServer, onUncaughtError),
```
notgiorgi: I'd also pass a custom fetch to `createMCPClient` `transport.fetch` param


```
const mergeWithTimeoutSignal = (
    signal: AbortSignal | null | undefined,
    timeoutMs: number,
): AbortSignal => {
    const timeoutSignal = AbortSignal.timeout(timeoutMs);

    if (!signal) {
        return timeoutSignal;
    }

    return AbortSignal.any([signal, timeoutSignal]);
};

const createMcpTimeoutFetch =
    (timeoutMs: number): typeof fetch =>
    async (input, init = {}) => {
        try {
            return await fetch(input, {
                ...init,
                signal: mergeWithTimeoutSignal(init.signal, timeoutMs),
            });
        } catch (error) {
            if (isMcpTimeoutError(error)) {
                throw new McpTimeoutError(timeoutMs, error);
            }

            throw error;
        }
    };
```

packages/backend/src/config/parseConfig.ts
ZeRego: Moved to `ai.copilot.mcpConnectionTimeoutMs` (env `AI_COPILOT_MCP_TIMEOUT_MS`, default 20s) in 83b5684b29. Top-level `mcp` is untouched.

packages/backend/src/ee/services/ai/AiAgentMcpRuntimeClient.ts
ZeRego: Done in 83b5684b29. `createMcpTimeoutFetch` wraps each request in `AbortSignal.timeout(ms)` (merged with any existing signal via `AbortSignal.any`) and is passed as `transport.fetch` for both the OAuth and non-OAuth transports, so a hung server's in-flight fetch is actually aborted; the abort maps to `McpTimeoutError` and flows through the existing error funnel.

One nuance worth flagging: the same transport/client is reused for the agent run, so this also bounds each tool *execution* call to `mcpConnectionTimeoutMs` (20s default), not just discovery. Fine for now IMO, but happy to split it into a separate/larger tool-call budget if you'd rather not cap long-running tools at the connection timeout.

PR #23789, opened 2026-06-02
fix: always skip cache in sql runner edit mode

### LINKED ISSUE #21371: I want a way to bypass or break the cache in SQL Runner queries
## Description
Currently, the SQL Runner in Lightdash caches query results. This can lead to users seeing stale data, especially when debugging or expecting fresh results from the database. The only current workaround is to modify the query (e.g., by adding a trivial `WHERE 1=1` clause) to force a cache invalidation.

## Expected Behavior
Clicking 'Run query' should always fetch fresh results from the database, or there should be an explicit option to bypass the cache and force a refresh.

## Suggested Solution

Add an option or button in SQL Runner to bypass the cache and fetch fresh results from the database.

## References
- [Lightdash caching documentation](https://docs.lightdash.com/guides/developer/caching#scope-of-the-cache) (updated just now to match logic in the code)

nanic:
Closes: #21371

### Description:

SQL Runner edit mode always invalidates the cache on run, matching explore edit mode. It wires `invalidateCache` through the existing query execution path.


joaoviana:
Initially re-read the ticket spec and thought it would have to include a button to invalidate the cache, but after reading the comments on the issue I understand this implementation is correct 👍 will be reviewed soon @nanic

packages/frontend/src/features/sqlRunner/components/ContentPanel.tsx
```diff
@@ -415,6 +415,8 @@ export const ContentPanel: FC = () => {
                     downloadLimit === null
                         ? MAX_SAFE_INTEGER
                         : (downloadLimit ?? limit),
+                    undefined,
```
joaoviana, line 419: 🤔 I believe this arg is for `parameters` could you check how to fetch them to pass them here instead of `undefined`?

tbf they were undefined before so can probably be skipped here for now, with a TODO 👍

packages/frontend/src/features/sqlRunner/store/thunks.test.ts
```diff
@@ -0,0 +1,44 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { executeSqlQuery } from '../../queryRunner/executeQuery';
+import { runSqlQuery } from './thunks';
+
+vi.mock('../../queryRunner/executeQuery', () => ({
```
joaoviana: I think we can remove `packages/frontend/src/features/sqlRunner/store/thunks.test.ts`.

It only tests thunk-to-helper wiring (`runSqlQuery` calls `executeSqlQuery(..., true)`), which feels like implementation detail.

joaoviana (commented):
looking great

joaoviana:
@nanic could you run `pnpm fix-format` and push?

joaoviana:
fix-format commit had no changes, pushed my commit here

nanic:
> fix-format commit had no changes, pushed my commit here

not sure what happened, but thanks

PR #23792, opened 2026-06-02
fix: always skip cache in sql runner edit mode

### LINKED ISSUE #21371: I want a way to bypass or break the cache in SQL Runner queries

joaoviana:
Duplicate of #23789 from @nanic, copied to a Lightdash-owned branch.

Closes: #21371

### Description:

SQL Runner edit mode always invalidates the cache on run, matching explore edit mode. It wires `invalidateCache` through the existing query execution path.


joaoviana:
main pr merged

PR #23806, opened 2026-06-02
fix: conditional formatting works with hidden metrics in pivot tables

### LINKED ISSUE #19288: Conditional formatting allows users to use hidden metrics but the output does not work.
### Description

When creating a table, if a user hides a given metric, they are still able to reference that metric within conditional formatting.

However, the conditional formatting doesn't work if the metric is hidden, if it isn't hidden then it will work just fine.

We should either fix whatever is causing the conditional formatting to not work, or remove the ability to choose hidden fields to compare against.

1. Create a pivot table with two dimensions and two metrics
2. Hide one of the metrics
3. Create conditional formatting for the unhidden metrics based on a field value from the hidden metric
4. Observe no formatting is applied
5. Unhiding the metric will then apply the formatting

### version

_No response_

### Cloud or self-hosting

None

jelstongreen:
## Summary

In pivot tables, conditional formatting that referenced a **hidden** metric rendered nothing. Hidden metrics are filtered out of the pivot output (`pivotQueryResults`), so their values never reached the conditional-formatting lookup. (Flat tables were unaffected — they build row fields from all cells, including hidden ones.)

This adds a context-keyed side-channel on `PivotData` carrying the raw values of hidden metrics, populated during the pivot data pass and consumed by the pivot `rowFields` builders. A shared `getPivotRowContextKey` helper is used on both the producer and consumer sides so the lookup keys cannot drift (keyed on displayed index + header dimensions).

The pivot output shape is unchanged — `dataValues`, column/row counts, and totals are untouched; the new `PivotData.hiddenContextValues` field is additive and optional.

## Test Plan

- [x] Unit test: `pivotQueryResults` stashes hidden metric values keyed by dimension context without changing the visible output shape
- [x] Unit test: `getPivotRowContextKey` is order-stable and distinguishes fields/values
- [ ] Manual: pivot with 2 dimensions + 2 metrics, hide one metric, add conditional formatting on the visible metric comparing to the hidden one — formatting now renders (both `metricsAsRows` orientations)

Closes: PROD-2372
Closes: #19288

### DIFF
--- a/packages/common/src/pivot/pivotQueryResults.test.ts
+++ b/packages/common/src/pivot/pivotQueryResults.test.ts
@@ -1,6 +1,7 @@
 import { type ItemsMap } from '../types/field';
 import { type PivotData } from '../types/pivot';
 import { type ResultRow } from '../types/results';
+import { getPivotRowContextKey } from '../utils/conditionalFormatting';
 import {
     SortByDirection,
     VizAggregationOptions,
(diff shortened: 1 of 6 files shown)


pimmey (changes requested):
@jelstongreen this fix is mostly good, but it targets the deprecated and now removed (https://github.com/lightdash/lightdash/pull/23731) front-end pivoting. The SQL backend pivoting now is the default codepath now since last week!

This would needs rebasing onto main first to get the latest and re-target the producer logic into `convertSqlPivotedRowsToPivotData`, sourcing hidden metrics from `pivotDetails.valuesColumns` and reading their pivoted cells.

The consumer-side `PivotTable` change and `getPivotRowContextKey` helper are mostly reusable!

jelstongreen:
@pimmey thanks — rebased onto main and re-targeted the producer.

- Confirmed: the legacy `pivotQueryResults` path was gone (#23731) with no remaining callers, so the original side-channel was sitting on dead code.
- Moved the producer logic into `convertSqlPivotedRowsToPivotData`: hidden metrics are now sourced from `pivotDetails.valuesColumns` (the columns dropped by the visibility filter), reading their pivoted cells via `row[col.pivotColumnName].value`.
- Keyed with the existing `getPivotRowContextKey` on the same displayed context the renderer rebuilds — visible `indexColumns` + visible `groupByColumns` pivot values — so producer/consumer keys can't drift. The field is only attached when a hidden metric exists.
- Consumer-side `PivotTable` change and the helper are reused as-is.
- Rewrote the producer unit test against the SQL path (keyed lookup across rows, visible output shape unchanged, and `hiddenContextValues` omitted when nothing is hidden). `common` tests/typecheck/lint green.

Ready for another look.

pimmey (approved):
Nice!

PR #23807, opened 2026-06-02
feat: full-row conditional formatting based on a single column's value

### LINKED ISSUE #23750: Full-row conditional formatting in pivot tables based on a single column's value
### Description

Customers want to apply conditional formatting to an **entire row** in a (pivot) table, triggered by the value of a single column, rather than only formatting the specific cell the rule is attached to.

* They want certain rows of a P&L to be highlighted (e.g. paint the whole row pink for subtotal rows, bold for total rows).
* Today, conditional formatting only paints the cell that matches the rule. They'd like the rule to be able to extend across the full row.

### Ask

* Add an option on a conditional formatting rule to apply the style to the entire row (and ideally the entire column in pivots), not just the matched cell.

### Related

* lightdash/lightdash#19288 — Conditional formatting using hidden metrics doesn't render (separate but adjacent bug)
* [#5079](<https://github.com/lightdash/lightdash/issues/5079>) (closed, stale) — Earlier request for entire row/column highlighting

jelstongreen:
## Summary

Adds a **Row** option to conditional formatting so a rule can paint an entire row's background when its single trigger column matches — e.g. highlight whole subtotal/total rows in a P&L by the value of a row-type column.

- New `ROW` value on `ConditionalFormattingColorApplyTo` + a pure `getRowConditionalFormattingColor` evaluator (single-color rules only).
- Applied in both the flat table and the pivot table; in pivots the fill spans **all** column groups, including the leftmost label/index column.
- **Precedence:** a cell that independently matches a Cell/Text rule keeps its own styling — cell rules paint over the row fill; under a Text rule the row background stays continuous.
- Config UI gains a "Row" segment in the apply-to control (offered only for single-color rules).

Out of scope (deferred): column-span, and row-level text colour.

> **Stacked on #23806** (the hidden-metric fix) — the headline use case drives the row highlight from a hidden helper column, which needs that fix to resolve the trigger value in pivots. Review/merge #23806 first.

## Test Plan

- [x] Unit tests: `getRowConditionalFormattingColor` matches only `applyTo: ROW` rules with a target, returns null for gradients, first-match wins, ignores Cell/Text
- [ ] Manual (flat): a Row rule fills the whole row; a Cell rule on another column still overrides its own cell
- [ ] Manual (pivot): hide a row-type helper metric, add a Row rule keyed on it — the whole visual row fills across all pivot column groups (both `metricsAsRows` orientations)

Closes: PROD-8058
Closes: #23750

### DIFF
--- a/packages/common/src/types/conditionalFormatting.ts
+++ b/packages/common/src/types/conditionalFormatting.ts
@@ -120,6 +120,7 @@ export enum ConditionalFormattingComparisonType {
 export enum ConditionalFormattingColorApplyTo {
     CELL = 'cell',
     TEXT = 'text',
+    ROW = 'row',
 }
 
 export type ConditionalRuleLabel = {
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -4409,7 +4409,7 @@ const models: TsoaRoute.Models = {
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
     ConditionalFormattingColorApplyTo: {
         dataType: 'refEnum',
-        enums: ['cell', 'text'],
+        enums: ['cell', 'text', 'row'],
     },
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
     ConditionalFormattingConfigWithSingleColor: {
--- a/packages/backend/src/generated/swagger.json
+++ b/packages/backend/src/generated/swagger.json
@@ -5051,7 +5051,7 @@
                 ]
             },
             "ConditionalFormattingColorApplyTo": {
-                "enum": ["cell", "text"],
+                "enum": ["cell", "text", "row"],
                 "type": "string"
             },
             "ConditionalFormattingConfigWithSingleColor": {
--- a/packages/common/src/schemas/json/chart-as-code-1.0.json
+++ b/packages/common/src/schemas/json/chart-as-code-1.0.json
@@ -529,7 +529,7 @@
             "type": "string"
         },
         "ConditionalFormattingColorApplyTo": {
-            "enum": ["cell", "text"],
+            "enum": ["cell", "text", "row"],
             "type": "string"
         },
         "ConditionalFormattingColorRange": {
--- a/packages/common/src/utils/conditionalFormatting.test.ts
+++ b/packages/common/src/utils/conditionalFormatting.test.ts
@@ -1,9 +1,11 @@
+import { ConditionalFormattingColorApplyTo } from '../types/conditionalFormatting';
 import { DimensionType, FieldType } from '../types/field';
 import { FilterOperator } from '../types/filter';
 import {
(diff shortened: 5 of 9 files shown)


jelstongreen:
## ✅ Live verification (local dev)

Ran the branch in a real browser against a seeded demo project and asserted the actual computed `background-color` of each cell (not a screenshot eyeball). A flat table and a pivot table were configured with a single **Row** rule: `status = "completed" → #ffc0cb`.

**Flat table** — the matched row is painted across every column; non-matching rows are untouched:

| row | status | computed bg of all cells |
|-----|--------|--------------------------|
| 1 | completed | `rgb(255,192,203)` (= #ffc0cb) ✅ |
| 2–5 | placed / returned / return_pending / shipped | `rgba(0,0,0,0)` ✅ |

**Pivot table** (second dimension pivoted to column headers, `metricsAsRows: false`) — the matched row is painted across the index + metric columns of the pivot group; non-matching rows untouched:

| row | status | bg (row# / Status / metric) |
|-----|--------|-----------------------------|
| 1 | completed | transparent / `rgb(255,192,203)` / `rgb(255,192,203)` ✅ |
| 2–5 | others | transparent ✅ |

Also confirmed: the `applyTo: "row"` value round-trips through chart-config serialization (visible in the explore URL state), and the pivot render path throws no runtime errors.

**Covered:** Row fill in flat + pivot with a visible string trigger; matched-vs-unmatched rows; config serialization.

**Not exercised in this pass (suggest a reviewer spot-check):**
- Cell/Text rule precedence painting over a row fill on the same cell.
- Hidden-metric trigger (the PROD-2372 integration, #23806) — tested here with a *visible* trigger only.
- `metricsAsRows: true` orientation.

**Minor (matches the review):** in the pivot, the leftmost row-number gutter cell is not painted (it is in the flat table). Cosmetic, not a correctness issue.

jelstongreen:
image

packages/common/src/types/conditionalFormatting.ts
```diff
@@ -120,6 +120,7 @@ export enum ConditionalFormattingComparisonType {
 export enum ConditionalFormattingColorApplyTo {
     CELL = 'cell',
     TEXT = 'text',
+    ROW = 'row',
```
pimmey, line 123: This would need `pnpm run generate-api` to re-gen types -- it will add some auto-generated files to the commit.

pimmey (changes requested):
~~Seems to work well with unpivoted tables, but stumbles with pivots? Perhaps still need need to rebase.~~
 
~~https://github.com/user-attachments/assets/8f435518-6190-44da-8aa4-74a46e59339a~~

Actually my bad! I think this was cache or something on my side. Works well! 

Just needs swagger and routes re-generated then!

packages/common/src/types/conditionalFormatting.ts
jelstongreen: Done — ran `pnpm run generate-api`. Committed the regenerated `routes.ts`, `swagger.json`, and the chart-as-code JSON schema with the new `row` enum value (375fbf0). Kept it scoped to the enum change; left out unrelated pre-existing generator drift (success/error ordering, searchRank type) so the diff stays reviewable.

pimmey (approved):

Issue #16961, opened 2025-09-17
Validator's "Rename model" corrupts custom dimensions and metrics by performing partial updates


IrakliJani:
### Description

**Description**

When a user renames a model in their dbt project and then uses the "Rename model" feature in the Lightdash validator to fix broken content, the update logic is incomplete. It fails to correctly update all references within custom dimensions and custom metrics that depend on the old model name.

This leads to a corrupted state where:
-   **Custom dimensions:** have their `sql` definition updated but retain the old `table` name.
-   **Custom metrics (based on custom dimensions):** have their `table` name updated but retain the old `sql` definition, which still references the old table.

This desynchronization breaks any charts using these custom fields and can lead to further corruption if the user attempts to run the rename operation again. A customer experienced this, which resulted in table names being repeatedly prefixed (e.g., `shopify_shopify_shopify_orders`). While this exact prefixing behavior was not reproduced, the underlying partial update bug was.

**Root Cause Hypothesis**

The "Rename model" service logic does not correctly traverse all dependencies for custom fields. It appears to miss updating:
1.  The `table` column in `saved_queries_version_custom_sql_dimensions`.
2.  The `sql` column in `saved_queries_version_additional_metrics` when the metric is based on a custom dimension.

### Steps to Reproduce the Bug or Issue

**Reproduction Steps**

1.  **Project Setup:**
    *   In your dbt project, create two identical models: `orders` and `shopify_orders`.
    *   Deploy the changes to Lightdash (seed / compile)

2.  **Create Dependent Content in Lightdash:**
    *   Navigate to the `orders` explore.
    *   Create a **custom dimension** with the following properties:
        *   Name: `All Caps Status`
        *   Table: `orders`
        *   SQL: `UPPER(${orders.status})`
    *   Create a **custom metric** based on the new custom dimension (e.g., a "Count" of `All Caps Status`).
    *   Create a chart that uses both the custom dimension and the custom metric. Save the chart.

3.  **Simulate dbt Changes:**
    *   In your dbt project, delete or rename the `orders` model, leaving only `shopify_orders`.
    *   Deploy changes to lightdash (seed / compile)

4.  **Use the Validator to Rename:**
    *   Open the validator in Lightdash. You will see errors related to the missing `orders` table.
    *   Use the "Rename model" feature to change all references from `orders` to `shopify_orders`.

5.  **Observe the Bug:**
    *   Navigate to the chart you created in Step 2. It will be broken.
[...]

PR #23845, opened 2026-06-03
fix(rename): make model rename idempotent when new name extends old

pimmey:
Closes: PROD-8098
Relates: #16961

### Description

Renaming a model to a name that **contains the old name as a prefix** (e.g. `orders` → `orders_restricted`) was not idempotent. The prefix replacers re-matched their own output, so re-running the same rename stacked the suffix (`orders_restricted_restricted_…`).

Field IDs doubled (via the `^from_` regex) while `tableName` stayed correct (it has an exact-match guard) — leaving charts internally inconsistent: the table resolves but every field under it breaks. Combined with a separate CLI bug that made a successful rename look like a failure, users retried and compounded the corruption.

Affects **both** the CLI `lightdash rename` and the validator "Fix" UI, which share the rename engine.

### Fix

Guard the three prefix replacers so they skip already-renamed values, but **only when the new name literally extends the old one** (i.e. `to` starts with `from_`):

- `replaceId` — no-op if the id already starts with the new prefix
- `replaceString` — negative lookahead skips occurrences already followed by the new suffix
- `replaceFull` — no-op if the value already equals the new name (this was doubling `exploreName` / `table`)

Normal renames (e.g. `payment` → `invoice`) take the exact original path — no behaviour change.

### Tradeoff

A field literally named `restricted_*` on the original model (id `orders_restricted_foo`) is treated as already-renamed and skipped on a first `orders → orders_restricted` rename. Rare, and far preferable to the compounding corruption.

### Tests

Adds regression tests to `rename.test.ts`: idempotency (`renameTwice === renameOnce` on a full chart), per-replacer no-op-on-already-renamed, and that a normal rename is unaffected. All 55 RenameService tests pass; backend typecheck + lint clean.

### DIFF
--- a/packages/backend/src/services/RenameService/rename.test.ts
+++ b/packages/backend/src/services/RenameService/rename.test.ts
@@ -1564,3 +1564,82 @@ describe('getNameChanges', () => {
         });
     });
 });
+
+describe('idempotent model rename (new name extends old, e.g. orders -> orders_restricted)', () => {
+    const nameChanges = {
+        from: 'orders',
+        to: 'orders_restricted',
+        fromReference: 'orders',
+        toReference: 'orders_restricted',
+        fromFieldName: undefined,
+        toFieldName: undefined,
+    };
+    const factory = createRenameFactory({ ...nameChanges, isPrefix: true });
+
+    test('replaceId renames bare ids but is a no-op on already-renamed ids', () => {
+        expect(factory.replaceId('orders_status')).toBe(
+            'orders_restricted_status',
+        );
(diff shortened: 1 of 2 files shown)


pimmey:
* **#23845** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23845?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23856, opened 2026-06-03
feat(ai-project-context): preview the writeback diff before opening the PR

joaoviana:
## What & why

When opening a `project_context` writeback PR, an admin had to trust it blind and check the diff on GitHub afterwards. This adds an **inline diff preview** so you review the exact file change *before* the PR is created.

## Flow

Clicking **Create PR** on a `project_context` finding opens a modal that renders the change to `lightdash.project_context.yml`, then **"Looks good, open PR"** fires the writeback. Other strategies (semantic_layer / sandbox) open the PR directly as before.

## Backend

- New read-only endpoint `GET /aiAgents/admin/review-items/:fingerprint/writeback-preview` → `{ available, before, after, fileName, op, entryId }`.
- Reuses `applyProjectContextWriteback` (deterministic, **no PR opened**); `semantic_layer` returns `{ available: false }` since it runs in a sandbox with no static diff.

## Frontend

- Renders the diff with **`@pierre/diffs`** (Apache-2.0): Shiki-highlighted YAML, split view, scrollable via its `<Virtualizer>`, theme pinned to Mantine's computed light/dark scheme.
- `useAiAgentReviewItemWritebackPreview` hook; modal gated to `project_context` rows.

## Also (folds in the clean-diff dependency)

`applyProjectContextWriteback` now serializes with `flowCollectionPadding: false`, so untouched flow lines (`terms: [a, b]`) aren't re-spaced to `[ a, b ]`. Without this the preview showed a noisy −78/+87 full-file rewrite; now it's just the touched entry.

## Notes

- New dep `@pierre/diffs` pulls in Shiki (lazy-loaded in the modal) + a Vite web worker.
- Built on the now-merged project_context writeback (#23841).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #23864, opened 2026-06-03
fix(ai-project-context): serialize entries before jsonb insert

joaoviana:
## Summary

Project compile failed with `invalid input syntax for type json` when caching the project_context document.

`ProjectContextModel.replaceEntriesForProject` passed the `entries` **array** straight to Knex for the `jsonb` column. The `pg` driver serializes a top-level JS array as a Postgres **array literal** (`{...}`), not JSON, so the `jsonb` column rejected it. (A plain object would have been auto-stringified by `pg` — arrays are the exception, which is why this slipped through.)

Fix: `JSON.stringify(entries)` so the column receives valid JSON.

```
insert into "project_context_document" ("entries", "project_uuid", "updated_at", "version")
  values ($1, $2, CURRENT_TIMESTAMP, $3)
  on conflict ("project_uuid") do update set ...
-- ERROR: invalid input syntax for type json
```

## Why the test didn't catch it

The existing unit test used `knex-mock-client`, which doesn't replicate real `pg` array-vs-JSON binding behavior — it actually **asserted the buggy unstringified array**. The test is flipped to assert the stringified binding, so it now matches what real Postgres requires.

## Regression analysis

- **Scope:** EE-only, behind the `ai-project-context` feature flag. The write path runs on project compile/refresh (and PR-merge reconcile) for GitHub-connected projects that have a `lightdash.project_context.yml`.
- **Affected:** any project whose ingest produced a non-empty `entries` array — the compile step that calls `replaceEntriesForProject` threw. Empty arrays happened to serialize to a valid-JSON literal, so unset projects were unaffected.
- **Not affected:** OSS installs, projects without the flag, projects with no project_context file, and the read path (`getDocument`) — `jsonb` is returned already-parsed by `pg`, so reads were never broken.
- No migration or data backfill needed; the column type is unchanged.

### DIFF
--- a/packages/backend/src/ee/models/ProjectContextModel.test.ts
+++ b/packages/backend/src/ee/models/ProjectContextModel.test.ts
@@ -24,7 +24,7 @@ describe('ProjectContextModel', () => {
         tracker.reset();
     });
 
-    test('stores entries as JSONB data instead of a JSON string', async () => {
+    test('serializes entries to a JSON string so pg writes valid jsonb', async () => {
         const entries: ProjectContextEntry[] = [
             {
                 id: 'hr',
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#23864** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23856** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #23890, opened 2026-06-04
fix(spaces): search nested spaces in Spaces list (#23887)

### LINKED ISSUE #23887: [Spaces] Search in Spaces list doesn't find child/nested spaces, inconsistent with global search
## Bug description

The search box in the Spaces list page (`/spaces`) only searches root-level spaces (due to a `nlevel(path) = 1` constraint in the backend query). Nested/child spaces are not returned when searching by name, even though the top global search bar correctly finds those same spaces.

This creates a confusing UX inconsistency: a user can find a space by name using the global search but get "No records to display" searching the exact same name in the Spaces list.

## Steps to reproduce

1. Create a child space nested under a root space
2. Navigate to Home > Spaces
3. Type the child space's name in the search box
4. Observe: "No records to display" even though the space exists
5. Use the top global search bar and type the same name
6. Observe: the space is found and shown in results

## Expected behavior

Searching by space name in the Spaces list should either:
- Return matching spaces at any nesting level (most intuitive), or
- Make it clear to the user that the search is scoped to root-level spaces only

## Actual behavior

"No records to display" is shown for a valid space name when the space is nested. The global search finds it without issue.

## Root cause

In `SpaceContentConfiguration.getSummaryQuery()`, when `filters.space.rootSpaces = true` (the default for the `/spaces` page), the query applies `AND nlevel(path) = 1` restricting results to root spaces only. The search term is applied in addition to this constraint, so nested spaces are never returned.

The global search uses a different code path that does not apply the `nlevel(path) = 1` filter.

## Environment

- Lightdash version: cloud
- Affected file: `packages/backend/src/models/ContentModel/ContentConfigurations/SpaceContentConfiguration.ts`

## Additional context

This inconsistency is particularly noticeable for users who are project admins and can see spaces in both views. The Spaces list search and the global search use different traversal strategies, leading to different results for the same query.

jelstongreen:
### Problem
The search box on the Spaces list page (`/spaces`) only returned root-level spaces. Nested/child spaces never appeared when searched by name, even though the global search bar finds those exact same spaces — a confusing UX inconsistency.

### Root cause
In `SpaceContentConfiguration.getSummaryQuery()`, the `rootSpaces` query path (the default for `/spaces`) always applies `AND nlevel(path) = 1`. The search term is applied on top of this, so nested spaces are filtered out before the name match can ever succeed.

### Fix
Only apply the `nlevel(path) = 1` root-level restriction when there is **no** search term. When the user is searching, results are matched at any nesting level — consistent with global search. Access control is unaffected: the existing `space_uuid IN (...)` filter already scopes results to spaces the user can view (it contains accessible spaces at every level).

### Testing
Added a unit regression test (`SpaceContentConfiguration.test.ts`) asserting the generated SQL:
- contains `nlevel(path) = 1` with no search term
- omits it when a search term is present

```
✓ restricts to root spaces when no search term is provided
✓ matches spaces at any nesting level when searching (issue #23887)
```

### DIFF
--- a/packages/backend/src/models/ContentModel/ContentConfigurations/SpaceContentConfiguration.test.ts
+++ b/packages/backend/src/models/ContentModel/ContentConfigurations/SpaceContentConfiguration.test.ts
@@ -1,0 +1,26 @@
+import knex from 'knex';
+import { MockClient } from 'knex-mock-client';
+import { ContentFilters } from '../ContentModelTypes';
+import { spaceContentConfiguration } from './SpaceContentConfiguration';
+
+const db = knex({ client: MockClient, dialect: 'pg' });
+
+const buildSql = (filters: ContentFilters): string =>
+    spaceContentConfiguration.getSummaryQuery(db, filters).toSQL().sql;
+
+const baseFilters: ContentFilters = {
+    spaceUuids: ['00000000-0000-0000-0000-000000000001'],
+    space: { rootSpaces: true },
+};
+
+describe('spaceContentConfiguration.getSummaryQuery rootSpaces search', () => {
+    it('restricts to root spaces when no search term is provided', () => {
+        const sql = buildSql(baseFilters);
+        expect(sql).toContain('nlevel(path) = 1');
+    });
+
+    it('matches spaces at any nesting level when searching (issue #23887)', () => {
+        const sql = buildSql({ ...baseFilters, search: 'nested space' });
+        expect(sql).not.toContain('nlevel(path) = 1');
+    });
+});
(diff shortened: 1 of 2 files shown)


almeidabbm (approved):
Code looks good and does what it's supposed to 🚀 

Main concern is the search UX, do we expect to see child spaces at the same level as parent spaces in this UI?

https://github.com/user-attachments/assets/15375e6d-1891-4980-9c6a-5f7e7c0ee0fd

jelstongreen:
> Code looks good and does what it's supposed to 🚀
> 
> Main concern is the search UX, do we expect to see child spaces at the same level as parent spaces in this UI?

This is a good point, what about if we showed them under their parents like a tree structure?

almeidabbm:
> > Code looks good and does what it's supposed to 🚀
> > Main concern is the search UX, do we expect to see child spaces at the same level as parent spaces in this UI?
> 
> This is a good point, what about if we showed them under their parents like a tree structure?

That would be a good one, but could possibly go against our current ux there? we don't have this notion of tree structure, could be a big rewrite of this view 😬

PR #23899, opened 2026-06-04
fix(ai-review): make turn-signal supersede atomic and concurrency-safe

joaoviana:
## Problem

The AI review classifier keeps **one signal per turn**: a re-review (a feedback change, or the next turn supplying a correction) *supersedes* the prior signal. `createTurnSignal` did this with a `DELETE … WHERE ai_prompt_uuid = ?` followed by an `INSERT`, and then wrote the promoted **review item** in a separate step *outside* the transaction.

There is no unique constraint on `ai_prompt_uuid` and no per-turn serialization, so two reviews of the same turn can run concurrently (e.g. a feedback event and a next-turn re-review, or two feedback events for the same message). When they interleave:

1. **Lost update** — both run `delete … then insert`; whichever commits last wins, so a *promoted* classification can be silently deleted by a concurrent *non-promoted* re-review.
2. **Orphaned review items** — because the review item was upserted outside the supersede transaction and never reconciled, the item survives even after its backing signal is deleted. The result is a review item with **no signal**, so its root cause and any context entry are unrecoverable, and the queue shows items that can't be traced back to a finding.

## Fix

`createTurnSignal` is now atomic and concurrency-safe:

- **Serialize per turn** with a transaction-scoped `pg_advisory_xact_lock` keyed on the prompt, so concurrent reviews of the same turn can't interleave their delete/insert. The lock releases on commit/rollback.
- **Write the review item in the same transaction** as the signal supersede, so a signal and the item it promotes are always created together (no out-of-band window).
- **Reconcile orphans**: when superseding, capture the prior fingerprints; after writing the new signal, delete any review item whose fingerprint no longer has *any* backing signal — but only while it is **pristine** (`open`, unassigned, no linked PR/issue, no writeback, no human status change). Items a human has triaged are never discarded.

- Only affects the EE AI-agent review classifier write path (`createTurnSignal`). No API, schema, or read-path changes.
- The advisory lock is keyed on a hash of the prompt id; hash collisions can only cause two unrelated turns to serialise briefly — correctness is unaffected.
- Orphan cleanup is intentionally conservative (pristine-only) so no human-actioned review item is ever deleted. A re-review that legitimately downgrades a finding to "not a failure" now correctly removes the stale auto-generated item instead of leaving it orphaned.
- Pre-existing orphaned rows are not touched by this change; they can be cleaned up separately if desired.

## Tests

`AiAgentReviewClassifierModel.test.ts` — extended `createTurnSignal` coverage:
- asserts the per-turn advisory lock is taken;
- a re-review that drops the finding deletes the now-orphaned item, guarded to pristine rows;
- a re-review with an unchanged fingerprint keeps (re-touches) the item rather than deleting it.

`pnpm -F backend typecheck:fast` ✅ · `jest AiAgentReviewClassifierModel` ✅ (17 passed)

---

### DIFF
--- a/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
+++ b/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
@@ -421,6 +421,8 @@ describe('AiAgentReviewClassifierModel', () => {
 
     describe('createTurnSignal', () => {
         it('persists a classified signal with inline finding fields', async () => {
+            tracker.on.any(/pg_advisory_xact_lock/).response([]);
+            tracker.on.select(AiAgentTurnSignalTableName).responseOnce([]);
             tracker.on.delete(AiAgentTurnSignalTableName).responseOnce(0);
             tracker.on.insert(AiAgentTurnSignalTableName).responseOnce([
                 {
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#23899** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23899?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23910, opened 2026-06-04
fix: speed up SavedChartModel.get with lateral version join

### LINKED ISSUE #23862: SavedChartModel.get() slow due to OR space join and unbounded versions join
SavedChartModel.get() (packages/backend/src/models/SavedChartModel.ts ~L917) has two structural query performance problems that compound at scale. This method is one of the most widely called in the backend — it runs on every chart view, query execution, dashboard load, scheduled delivery, embed view, and more.

### 1. OR condition in the space join

```
.innerJoin(SpaceTableName, function spaceJoin() {
    this.on(
        ${SpaceTableName}.space_id, '=', ${DashboardsTableName}.space_id,
    ).orOn(
        ${SpaceTableName}.space_id, '=', ${SavedChartsTableName}.space_id,
    );
})
```

PostgreSQL cannot use a B-tree index on an OR join condition. It typically falls back to a hash join or nested loop with a sequential scan on the spaces table. As the spaces/dashboards tables grow this gets progressively worse.

### 2. INNER JOIN on all saved_queries_versions

```ts
.innerJoin(
    'saved_queries_versions',
    ${SavedChartsTableName}.saved_query_id,
    'saved_queries_versions.saved_query_id',
)
// ...
.orderBy('saved_queries_versions.created_at', 'desc')
.limit(1)
```

Every historical version of the chart is joined into the intermediate result set, then ORDER BY created_at DESC LIMIT 1 picks the latest. For charts that have accumulated many versions this significantly inflates the intermediate row count before the LIMIT applies.

## Impact

Becomes visible at scale: large number of chart versions, large spaces/dashboards tables, or high-frequency usage (embeds, scheduled deliveries). Surfaces as elevated query times in DB logs.

## Suggested fix

*OR space join:* Replace with a CTE or derived table that resolves the correct space_id before the join, e.g. COALESCE(dashboards.space_id, saved_queries.space_id), then join spaces on that single value.

*Versions join:* Pre-filter to the latest version per saved_query_id in a subquery/CTE before joining, rather than joining all rows and sorting afterward. Something like:

```sql
WITH latest_versions AS (
    SELECT DISTINCT ON (saved_query_id) *
    FROM saved_queries_versions
    ORDER BY saved_query_id, created_at DESC
)
```

Then join latest_versions instead of the full saved_queries_versions table.

almeidabbm:
`SavedChartModel.get()` runs on every chart view, dashboard tile, embed and scheduled delivery. Two problems made it slow and CPU-heavy at scale:

1. The versions join pulled in **every** historical version of the chart, then sorted and took one.
2. The `uuid OR slug` filter let the planner fall back to a parallel seq scan under degraded stats

### Fix

- **Versions** → `CROSS JOIN LATERAL (… ORDER BY created_at DESC LIMIT 1)`: a single index probe instead of joining/sorting all versions (same pattern as #21145). `versionUuid` moves inside the lateral.
- **Space join** → `spaces.space_id = COALESCE(saved_queries.space_id, dashboards.space_id)`.
- **Lookup** → a `chart_lookup` CTE with two single-predicate index probes (uuid, slug) instead of `uuid OR slug`, so there's no seq-scan plan available under bad stats. Candidates are applied as `saved_query_id = ANY(ARRAY(SELECT … FROM chart_lookup))`, which folds to an InitPlan constant and keeps the scan on the index. The outer `ORDER BY versions.created_at DESC LIMIT 1` is unchanged, so uuid/slug collisions and duplicate slugs resolve exactly as before (latest-version wins).

No schema or behavior changes; slug-only lookups untouched.

### Results

1M charts / 2M versions, target chart with 50k versions, same DB:

| | before | after |
|---|---|---|
| query, healthy stats | 57–70 ms | **0.4–1.2 ms** |
| query, missing stats | 877 ms (parallel seq scan) | **43 ms** (on indexes) |
| throughput (8 clients) | 63 tps | **1,850 tps** |
| DB CPU at equal 40 qps load | ~190% | **~28%** |

### Verification

Lock-down tests (#23971, `packages/api-tests`) pass against both the legacy and new query, returning the identical row (executed SQL confirmed via pg log). Two cross-project tests are CI-only. Typecheck/lint clean.

Closes: PROD-8118
Closes: #23862

test-all

### DIFF
--- a/packages/backend/src/models/SavedChartModel.ts
+++ b/packages/backend/src/models/SavedChartModel.ts
@@ -966,17 +966,9 @@ export class SavedChartModel {
                         `${DashboardsTableName}.dashboard_uuid`,
                         `${SavedChartsTableName}.dashboard_uuid`,
                     )
-                    .innerJoin(SpaceTableName, function spaceJoin() {
-                        this.on(
-                            `${SpaceTableName}.space_id`,
-                            '=',
-                            `${DashboardsTableName}.space_id`,
-                        ).orOn(
-                            `${SpaceTableName}.space_id`,
-                            '=',
-                            `${SavedChartsTableName}.space_id`,
-                        );
-                    })
+                    .joinRaw(
+                        `INNER JOIN ${SpaceTableName} ON ${SpaceTableName}.space_id = COALESCE(${SavedChartsTableName}.space_id, ${DashboardsTableName}.space_id)`,
+                    )
                     .innerJoin(
                         ProjectTableName,
                         `${SpaceTableName}.project_id`,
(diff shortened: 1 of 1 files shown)


almeidabbm:
* **#23910** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#23971** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23971?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ZeRego (approved):

PR #23920, opened 2026-06-04
fix(bigquery): reject empty SSO keyfiles and hide Save for OAuth credential types

### LINKED ISSUE #23914: BigQuery per-user SSO: empty keyfile persisted → all queries fail with "does not contain a client_email field"
## Summary

On projects using BigQuery with **"User Account (Sign in with Google)"** authentication and **"Require users to provide their own credentials"** (`requireUserCredentials`) enabled, some users end up with their per-user warehouse credential stored with an **empty keyfile**. Every query they run (charts, dashboards, AI) then fails at the authentication step with:

> The incoming JSON object does not contain a client_email field

It is especially common on instances where users sign in to Lightdash via a **non-Google IdP (e.g. Azure AD / generic OIDC)**, because those teams manage per-user BigQuery credentials by hand and aren't auto-repaired by a Google login. It **recurs**: re-authenticating fixes it temporarily, then it breaks again.

## Affected configuration

- Warehouse: **BigQuery**
- Auth type: **SSO / "User Account (Sign in with Google)"**
- `requireUserCredentials` = **true**

## Root cause (verified)

The affected user's stored per-user warehouse credential decrypts to:

```json
{"type":"bigquery","keyfileContents":{}}
```

i.e. an **empty `keyfileContents`**. At query time this empty keyfile is passed to the BigQuery client (`new BigQuery({ credentials: keyfileContents })`) and `google-auth-library` throws the `client_email` error.

**Where the empty keyfile comes from (confirmed): the "Edit credentials" modal** in Settings → My warehouse connections — *not* the OAuth/token path.

- The API strips secrets before sending to the browser, so the client only ever has the public credential (`{ type: 'bigquery' }`).
- The edit modal seeds its form with a placeholder, which for BigQuery is `{ type: 'bigquery', keyfileContents: {} }`.
- For BigQuery the modal body shows only a "Sign in with Google" button (there is no keyfile field) **but still renders a primary "Save" button**.
- Clicking **Save** submits the placeholder, and the update does a **full replace** of the stored credential — overwriting the valid keyfile with `{}`.

## Reproduction (isolated)

Constructing `@google-cloud/bigquery@7.9.2` directly with each keyfile shape and running a query:

| `keyfileContents` | Error at query time |
| -- | -- |
| `{}` (empty) | **"…does not contain a `client_email` field"** ← matches reports |
| `{ type:'authorized_user', client_id, client_secret }` (no refresh_token) | "…does not contain a `refresh_token` field" |
| `{ type:'authorized_user', …, refresh_token:'' }` | "…does not contain a `refresh_token` field" |
| `{ type:'authorized_user', …, refresh_token:'<token>' }` | `invalid_client` (proceeds to auth) |

So **only an empty `{}` keyfile reproduces the reported error** — confirming the stored credential is genuinely empty.

## Why it recurs, and why users get stuck

- **Recurs:** re-authenticating via the "Sign in with Google" popup deletes + recreates a valid credential, so it fixes the error — until someone opens the Edit modal and clicks **Save** again, which overwrites it back to `{}`.
- **Stuck:** the in-app prompt to (re)connect only reopens when the user has **zero** credentials. An empty-but-present credential counts as one, so the prompt is suppressed — and the frontend can't detect the emptiness because the keyfile is stripped from API responses.

pimmey:
Closes: PROD-8137
Closes: #23914

### Description:

Fixes a bug where BigQuery SSO credentials could be persisted as an empty `keyfileContents: {}` placeholder, which would silently overwrite a working credential and later surface as the opaque error "does not contain a client_email field".

**Root cause:** the `EditCredentialsModal` seeds its form with a masked placeholder (the real keyfile is stripped from API responses) but still shows a generic "Save" button. For BigQuery there is no keyfile field to fill in, so clicking Save persisted `{"type":"bigquery","keyfileContents":{}}` — exactly the value observed on affected instances. Re-authenticating via the Google popup fixed it; the next Edit + Save wiped it again.

A new `bigquerySsoUserCredentialsSchema` Zod schema validates that BigQuery keyfiles contain a non-empty `refresh_token`, since per-user BigQuery credentials are always `authorized_user` (SSO) type. It is applied in two places:

- **On write**: both `create` and `update` reject invalid credentials before encryption/persistence, preventing an empty shell from overwriting a working credential (covers the Edit modal and any direct API call).
- **On read**: an invalid BigQuery keyfile throws a typed `BigqueryTokenError` (HTTP 401) prompting reauthentication, instead of the confusing downstream `client_email` error.

The "Save" button is also hidden in `EditCredentialsModal` for SSO-based warehouse types (BigQuery, Snowflake, Databricks), mirroring `CreateCredentialsModal` — their credentials can only be set through the OAuth popup, so a generic save would have no valid secret to persist. This also closes the identical exposure for Snowflake SSO and Databricks U2M edit flows.

### DIFF
--- a/packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
+++ b/packages/backend/src/models/UserWarehouseCredentials/UserWarehouseCredentialsModel.ts
@@ -1,10 +1,13 @@
 import {
     assertUnreachable,
+    bigquerySsoUserCredentialsSchema,
+    BigqueryTokenError,
     CreateWarehouseCredentials,
     DatabricksAuthenticationType,
     databricksOauthU2mUserCredentialsSchema,
     DatabricksTokenError,
     NotFoundError,
+    ParameterError,
     ProjectType,
     SnowflakeAuthenticationType,
     snowflakeSsoUserCredentialsSchema,
@@ -377,6 +380,25 @@ export class UserWarehouseCredentialsModel {
                 }
             }
 
+            // Reject empty BigQuery keyfiles before they reach the warehouse
+            // client (where they surface as "does not contain a client_email
+            // field"), so the user is prompted to reauthenticate instead.
+            // Per-user BigQuery credentials are always SSO, so the refresh_token
+            // requirement applies to every BigQuery user credential.
+            if (
+                credentialsWithSecrets.credentials.type ===
+                WarehouseTypes.BIGQUERY
+            ) {
+                const result = bigquerySsoUserCredentialsSchema.safeParse(
+                    credentialsWithSecrets.credentials,
+                );
+                if (!result.success) {
+                    throw new BigqueryTokenError(
+                        `Please reauthenticate to access BigQuery`,
+                    );
+                }
+            }
+
             return credentialsWithSecrets;
         }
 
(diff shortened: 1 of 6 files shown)


pimmey:
* **#23920** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23920?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23929, opened 2026-06-05
fix(data-apps): add `view:Explore` to `Viewer` to allow running queries

### LINKED ISSUE #23928: [Data Apps] Project Viewers cannot access Data Apps - 'You don't have access to this resource or action'
## Bug description

Users with the project **Viewer** role get "You don't have access to this resource or action" when opening a Data App via its preview URL, even when the app is in a space they have access to. Users with the project **Developer** (Editor) role in the same space can open the app without issue.

The `view:DataApp` permission scope appears to be missing from the project Viewer role.

## Steps to reproduce

1. Create a Data App and add it to a shared space.
2. Invite two users to the project: one as Developer, one as Viewer. Both should have access to the space.
3. Share the app's preview URL with both users.
4. Developer opens the URL successfully.
5. Viewer receives "You don't have access to this resource or action".

## Expected behavior

Project Viewers with space access should be able to open and interact with Data Apps. The `view:DataApp` scope should be included in the Viewer role.

## Actual behavior

Project Viewers are blocked with an access denied error. Project Developers (and above) can access the app fine.

## Workaround

Upgrade the affected user's project role from Viewer to Interactive Viewer or Developer.

Spissable:
Closes: https://github.com/lightdash/lightdash/issues/23928

### Description:
Data apps are using the metric-query endpoint which is gated behind `view:Explore`. However, this scope didn't exist yet so we strictly relied on `manage:Explore` (Interactive Viewer+) for being able to run the queries. A simple "Viewer" would experience 403 responses when running any queries via the data apps.

### DIFF
--- a/packages/common/src/authorization/organizationMemberAbility.ts
+++ b/packages/common/src/authorization/organizationMemberAbility.ts
@@ -86,6 +86,9 @@ export const applyOrganizationMemberStaticAbilities: Record<
         can('view', 'Project', {
             organizationUuid: member.organizationUuid,
         });
+        can('view', 'Explore', {
+            organizationUuid: member.organizationUuid,
+        });
         can('view', 'Organization', {
             organizationUuid: member.organizationUuid,
         });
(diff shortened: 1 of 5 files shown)


Spissable:
We decided to go the opposite route. It will be required to be at least Interactive Viewer to view data apps

PR #23941, opened 2026-06-05
fix(pivot-table): force pivot dimensions visible when dragged to columns

### LINKED ISSUE #23937: Pivot table renders blank when all pivot dimensions are hidden
## Description

A table chart pivoted on a dimension renders completely blank (no error, no "no data" state) when that pivoted dimension's column is hidden. The query runs fine and the raw results come back correctly, but the response has `pivotDetails: null`, so the pivot table has nothing to render. Reproduces in both the saved-chart view and the explore/edit view.

## Root cause

The pivot config drag handler (`packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx`, `onDragEnd`) only mutates `pivotDimensions` when a dimension is dragged from Rows into Columns. It never updates the column's `visible` property:

```ts
// Add pivot — row → column
const fieldId = rows[source.index];
setPivotDimensions([...]);
```

So a dimension that was hidden while it lived in Rows carries `visible: false` with it into the pivot. That is a state the UI is meant to forbid — the visibility toggle is disabled for pivot dimensions (`disableHidingDimensions` in `ColumnConfiguration.tsx`, "Cannot hide dimensions when pivoting"), so the user can't even un-hide it.

Downstream, `derivePivotConfigurationFromChart` excludes hidden dimensions from `groupByColumns`. With the only pivot dimension hidden, `groupByColumns` is empty, `isValid` rejects the config, derivation returns `undefined`, the backend runs the query unpivoted, and `pivotDetails` comes back `null`.

This became visible after the SQL pivot migration (#23672 forced SQL pivot on, #23785 removed the legacy in-memory pivot). The old in-memory pivot ignored `visible: false` on pivot dimensions and spread them anyway, so these charts rendered fine before.

## Steps to reproduce

1. Create a table chart with two dimensions and a metric.
2. Hide one dimension's column (while it's a row).
3. Drag that hidden dimension into Columns to pivot on it.
4. The pivot table renders blank.

## Expected behavior

The pivot table renders with the dimension's values spread as column headers. Promoting a dimension to a pivot column should make it visible.

## Fix

In `onDragEnd`, when a dimension is dragged into Columns, force `visible: true` on it. Gate this on the `hide-pivot-dimensions` feature flag (`useIsHidePivotDimsEnabled`) so the in-progress feature that intentionally allows hidden pivot dimensions is untouched.

## Workaround

Drag the dimension out of Columns and back in (with the fix) — or, today, recreate the pivot with the column visible.

almeidabbm:
Dragging a dimension from Rows into Columns to pivot on it only updated `pivotDimensions`, never the column's `visible` flag. A dimension hidden while it was a row carried `visible: false` into the pivot, which left `groupByColumns` empty, derivation returned `undefined`, and the pivot table rendered blank.

This forces the dimension visible when it's dragged into Columns, gated on the `hide-pivot-dimensions` flag so the feature that intentionally allows hidden pivot dimensions is unaffected.

**Before**

https://github.com/user-attachments/assets/8d198f6c-0038-491b-b957-790136adfaf2

**After**

https://github.com/user-attachments/assets/baa9de35-4e50-4e8f-aa31-53b69be595ea

Closes #23937

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/TableConfigPanel/GeneralSettings.tsx
@@ -109,6 +109,13 @@ const GeneralSettings: FC = () => {
                         fieldId,
                         ...columns.slice(destination.index),
                     ]);
+                    // Pivot columns are always visible — a dim hidden as a row
+                    // would otherwise carry visible:false in and blank the table.
+                    if (!isHidePivotDimsEnabled) {
+                        chartConfig.updateColumnProperty(fieldId, {
+                            visible: true,
+                        });
+                    }
                 } else {
                     // Remove pivot
                     const fieldId = columns[source.index];
@@ -144,6 +151,7 @@ const GeneralSettings: FC = () => {
             setPivotDimensions,
             showToastError,
             handleToggleMetricsAsRows,
+            isHidePivotDimsEnabled,
         ],
     );
 


almeidabbm:
* **#23941** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23941?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23941?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23958, opened 2026-06-05
fix(filters): read async field values from result column key

### LINKED ISSUE #18633: Dropdown not working in Lightdash dashboard filter
### Description

I've been noticing a recurring issue where the filter dropdown doesn't reliably open when users attempt to change the filter value. In many cases, the dropdown fails to appear at all, forcing users to manually type in their selection. Has this behaviour been flagged or submitted as a bug previously?

Image

1. Create charts in Lightdash dashbord.
2. Add filter to the top of the dashboard (no value is selected by default)
3. Click on the filter but no values appear. However, it works when the value is manually typed in.

### version

_No response_

### Cloud or self-hosting

None

almeidabbm:
### Description

Filter value dropdowns showed "No results found" for dimensions on aliased joins when the async query path is enabled (`ResultsCacheEnabled` flag). The backend resolves an aliased join to the original explore and rewrites the fieldId, so result rows come back keyed by the rewritten id (e.g. `generated_b_region`) while the frontend looked rows up by the requested id (`join_1_region`), silently dropping every value. Manually typing values still worked, which is how this was reported.

Fix: read the single result column key from the query results instead of recomputing it client-side, falling back to the requested fieldId. Field value search always selects exactly one dimension, and the columns map is built from the same warehouse fields that key the rows.

Repro: `generated_a` explore in the demo project joins `generated_b` as `join_1`. With the results cache flag enabled, a dashboard or explorer filter on `Join 1 Region` showed an empty dropdown; after the fix it populates and searching works.

**Before**
CleanShot 2026-06-05 at 15 09 42

**After**
CleanShot 2026-06-05 at 15 06 57

Closes: PROD-1148
Closes: #18633

### DIFF
--- a/packages/frontend/src/hooks/useFieldValues.test.ts
+++ b/packages/frontend/src/hooks/useFieldValues.test.ts
@@ -1,7 +1,11 @@
 import { QueryHistoryStatus } from '@lightdash/common';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { lightdashApi } from '../api';
-import { MAX_POLL_ATTEMPTS, pollForFieldValueResults } from './useFieldValues';
+import {
+    getFieldValuesAsync,
+    MAX_POLL_ATTEMPTS,
+    pollForFieldValueResults,
+} from './useFieldValues';
 
 vi.mock('../api', () => ({
     lightdashApi: vi.fn(),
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#23958** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23958?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23958?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23968, opened 2026-06-05
fix(pivot): clear stale re-run prompt when a sort-only pivot dim is hidden

pimmey:
Closes: PROD-8158

### Description:

Fixes a bug where the "re-run query" prompt would remain permanently stale when a pivot dimension was configured as sort-only (hidden). Hidden sort-only pivot dimensions are routed to `sortOnlyDimensions` and never appear in `resultsData.pivotDetails.groupByColumns`, so comparing the full `pivotDimensions` array against the results would always produce a mismatch.

The fix filters `pivotDimensions` down to only visible columns before comparing against the results, ensuring the stale query indicator accurately reflects only the dimensions that actually appear in the pivot output.

### DIFF
--- a/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
+++ b/packages/frontend/src/hooks/tableVisualization/useTableConfig.ts
@@ -215,8 +215,18 @@ const useTableConfig = (
             resultsData?.pivotDetails?.groupByColumns?.map(
                 (col) => col.reference,
             ) ?? [];
-        return !isEqual(pivotDimensions ?? [], resultsPivotDimensions);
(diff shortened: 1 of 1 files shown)


pimmey:
* **#23968** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23968?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23968?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23973, opened 2026-06-05
chore(dev): add generate-api fast script

IrakliJani:
## Summary
- Add `pnpm generate-api:fast` using the tsgo common build path
- Keep generated schema post-step aligned with `pnpm generate-api`

## Testing
- `pnpm generate-api:fast`


IrakliJani:
Note: `postgenerate-api:fast` is not faster by itself. It exists because pnpm runs the matching `post<script>` hook for `generate-api:fast`; keeping it mirrors the normal `postgenerate-api` behavior so the fast command still regenerates the related schemas.

PR #23992, opened 2026-06-08
fix: surface partial field compilation errors

### LINKED ISSUE #23673: Missing project-scoped parameter causes silent SQL compilation error instead of surfacing a visible error
## Summary

When a chart references a project-scoped parameter (defined in `lightdash.config.yaml`) that is not present in the target project, Lightdash silently generates an incorrect SQL query rather than surfacing a visible error. The chart renders as empty with no error message, making it very difficult to diagnose.

## Steps to reproduce

1. Define a project-scoped parameter (e.g. `financial_display_units`) in `lightdash.config.yaml` in one dbt project (Project B).
2. Build charts in a Lightdash project that reference this parameter.
3. Deploy to a second Lightdash project (e.g. PROD) from a different dbt project (Project A) that does not include the `lightdash.config.yaml` from Project B - so the parameter is absent.
4. Promote a dashboard containing those charts from the preview (where the parameter exists) to PROD (where it does not).
5. Observe: the chart appears empty in PROD with no error message. The generated SQL is silently incorrect.

## Expected behavior

Lightdash should detect that a referenced parameter is missing and surface a visible error on the chart tile (e.g. "Parameter 'financial_display_units' is not defined in this project"), rather than silently producing malformed SQL.

## Actual behavior

The chart renders as empty. No error is shown. The SQL is compiled but incorrect, making the root cause very hard to find without inspecting the raw query.

## Why this matters

Silent failures in SQL compilation are particularly hard to debug. Users can spend significant time ruling out data issues, deploy issues, and permissions before discovering a missing parameter is the cause. A clear error message would surface the problem immediately.

## Notes

- This is especially likely to surface in multi-project dbt setups where different projects may have different `lightdash.config.yaml` files.
- The workaround is to move the parameter definition to the individual models rather than the project-level config, so it travels with the models regardless of which project deploys them.

notgiorgi:
### Description
- Surface field-level partial compilation errors during metric query compilation.
- Prevent queries from silently generating SQL for selected, sorted, or filtered fields compiled as `NULL`.
- Add regression coverage for selected fields with partial compilation errors.

### Repro steps
Using the Jaffle shop project:

1. Add a model dimension that references a project-level parameter that does not exist in `lightdash.config.yml`.

   ```yaml
   - name: missing_parameter_cli_verify
     description: Missing parameter verification dimension
     meta:
       dimension:
         sql: ${lightdash.parameters.missing_parameter}
     type: string
   ```

2. Deploy the semantic layer with the Lightdash CLI:

   ```bash
   lightdash deploy --project <project_uuid> --select orders
   ```

3. Confirm the CLI reports a partial compile, not a hard deploy failure:

   ```text
   PARTIAL_SUCCESS> orders
   Dimension "missing_parameter_cli_verify" failed to compile: Missing parameters: missing_parameter.
   ```

4. In Lightdash, create or open a chart on the `orders` explore that selects:
   - Dimension: `orders_missing_parameter_cli_verify`
   - Metric: `orders_unique_order_count`

5. Add that chart to a dashboard and load the dashboard.

Before this fix, the dashboard tile query succeeds and generates SQL like:

```sql
SELECT
  NULL AS "orders_missing_parameter_cli_verify",
  COUNT(DISTINCT "orders".order_id) AS "orders_unique_order_count"
FROM ...
```

The tile can then render empty/wrong data without surfacing the field compilation error.

After this fix, the dashboard tile query fails visibly at query time with:

```text
Dimension "missing_parameter_cli_verify" failed to compile: Missing parameters: missing_parameter.
```

<details>
<summary>Screenshots</summary>
Before:

image

After:

image

</details>

### Testing
- `pnpm -F backend test -- MetricQueryBuilder.test.ts --runInBand`
- `pnpm -F backend typecheck:fast`
- `pnpm -F backend format`
- Verified through a Jaffle shop CLI partial deploy and dashboard tile query.

Closes: PROD-8010
Closes: #23673

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.test.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.test.ts
@@ -83,6 +83,41 @@ const buildQuery = (
         parameterDefinitions: {},
     }).compileQuery();
 
+describe('field compilation errors', () => {
+    const exploreWithErroredDimension: Explore = {
+        ...EXPLORE,
+        tables: {
+            ...EXPLORE.tables,
+            table1: {
+                ...EXPLORE.tables.table1,
+                dimensions: {
+                    ...EXPLORE.tables.table1.dimensions,
+                    dim1: {
+                        ...EXPLORE.tables.table1.dimensions.dim1,
+                        compiledSql: 'NULL',
+                        compilationError: {
+                            message:
+                                'Dimension "dim1" failed to compile: Missing parameters: missing_parameter',
+                        },
+                    },
+                },
+            },
+        },
+    };
+
+    it('throws when a selected field has a partial compilation error', () => {
+        expect(() =>
+            buildQuery({
+                explore: exploreWithErroredDimension,
+                compiledMetricQuery: METRIC_QUERY,
+                warehouseSqlBuilder: warehouseClientMock,
+                intrinsicUserAttributes: {},
+                timezone: QUERY_BUILDER_UTC_TIMEZONE,
+            }),
+        ).toThrow('Missing parameters: missing_parameter');
+    });
+});
+
 const POP_TEST_POP_METRIC_NAME = 'total_order_amount__pop__year_1__testpop';
 const POP_TEST_POP_METRIC_ID = `orders_${POP_TEST_POP_METRIC_NAME}`;
 const POP_TEST_FANOUT_POP_METRIC_NAME = 'metric_amount__pop__year_1__fanout';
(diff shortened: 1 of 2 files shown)


notgiorgi:
* **#23992** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23992?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/23992?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24003, opened 2026-06-08
fix(dashboards): reconcile stale filter ids from shared links

### LINKED ISSUE #23803: [Dashboard Filters] URL filter ID mismatch causes tileTargets to be ignored - filter applies to all tiles
## Bug description

When a shared dashboard URL contains a filter whose `id` no longer matches any saved native filter `id` (e.g. after the native filter was deleted and re-added, or its target field was changed), Lightdash treats the URL filter as a new ad-hoc dimension with no `tileTargets`. This causes the filter to apply to all tiles instead of honoring the per-tile configuration on the original native filter.

This is silent and dangerous: end users see data that appears filtered to a single segment even on tiles explicitly excluded from that filter, leading to incorrect data being displayed.

## Steps to reproduce

1. Create a dashboard with a native dimension filter configured with custom tileTargets (applied to some tiles, excluded from others)
2. Share the dashboard URL with that filter active
3. Delete and re-add the same native filter (or change its target field in edit mode), then save the dashboard
4. Open the previously shared URL

## Expected behavior

The URL filter should either:
- Match by target fieldId + tableName when no ID match is found and inherit tileTargets from the matching native filter, or
- Warn the user that the URL contains a stale/unrecognized filter ID

## Actual behavior

The URL filter (now unmatched by ID) is appended as a new dimension with no tileTargets via `applyDimensionOverrides` in `packages/common/src/utils/filters.ts`. This creates a second filter that applies to every tile, overriding the intended per-tile configuration from the native filter.

## Root cause

`applyDimensionOverrides` matches URL overrides to native filters strictly by `id`. When no ID match is found, the override is appended as a new dimension with no tileTargets.

Filter IDs are regenerated client-side via `uuidv4()` whenever a native filter is deleted and re-added, or when its target field is changed in edit mode. Normal value/operator edits preserve the ID. There is no UI warning that certain edit operations will invalidate previously shared URLs for that filter.

## Environment

- Lightdash version: cloud

## Additional context

Related closed issue: https://github.com/lightdash/lightdash/issues/8008

The lack of fallback matching by fieldId+tableName means that any dashboard edit touching filter identity silently breaks distributed shared links. This is particularly harmful when the filter controls tile-level visibility of segmented data.

tatianainama:
## What

Shared dashboard links carry saved-filter overrides by `id`. When a saved filter is deleted and re-added, or its target field changes, that `id` no longer matches any saved filter, and the override was appended as a duplicate dimension with no `tileTargets`. The result: the filter applied to **every** tile, ignoring the per-tile configuration on the original filter, so viewers saw segmented data even on tiles explicitly excluded from the filter.

## Fix

`applyDimensionOverrides` now matches a stale override by `target.fieldId` + `target.tableName` when its `id` doesn't match, merging it onto the saved filter and inheriting that filter's `tileTargets` (and `id`, lock state). The function's signature, return type, and append-the-unmatched behavior are unchanged, this is purely a smarter match.

That satisfies the issue's first proposed fix (match by fieldId + tableName) and resolves the reported case where a re-added/renamed filter rotates its id.

Closes #23803
Linear: PROD-8077

### DIFF
--- a/packages/common/src/types/applyDimensionOverrides.test.ts
+++ b/packages/common/src/types/applyDimensionOverrides.test.ts
@@ -506,4 +506,187 @@ describe('applyDimensionOverrides', () => {
             );
         });
     });
+
+    // A shared link carries a filter id that no longer matches the saved filter
+    // (the id is regenerated when a filter is deleted and re-added or its field
+    // is changed). The override must fall back to matching by fieldId +
+    // tableName so it merges into the saved filter and inherits its tileTargets,
+    // instead of being appended as a duplicate that applies to every tile.
+    describe('field-match fallback for stale ids', () => {
+        it('reconciles a stale id by field and inherits the saved tileTargets', () => {
+            const savedFilters: DashboardFilters = {
+                dimensions: [
+                    createBaseDashboardFilter(
+                        'new-id',
+                        'customers_customer_id',
+                        'customers',
+                        ['123'],
+                        {
+                            'tile-1': {
+                                fieldId: 'customers_customer_id',
+                                tableName: 'customers',
+                            },
+                            'tile-2': false,
+                        },
+                    ),
+                ],
+                metrics: [],
+                tableCalculations: [],
+            };
+            const overrides: DashboardFilterRule[] = [
+                createOverrideFilter(
+                    'old-id',
+                    'customers_customer_id',
+                    'customers',
+                    ['789'],
+                ),
+            ];
+
+            const result = applyDimensionOverrides(savedFilters, overrides);
+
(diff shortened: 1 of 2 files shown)


tatianainama:
* **#24003** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24003?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24003?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24016, opened 2026-06-08
fix(pivot): order hidden sort-only dims at their declared position (PROD-8159)

pimmey:
Closes: PROD-8159

### Description:

Introduces a `pivotColumnsOrder` field to `PivotConfiguration` that records the declared order of pivot-column dimensions (visible `groupByColumns` plus hidden `sortOnlyDimensions`) as they appear in `pivotConfig.columns`.

Previously, when a sort-only dimension was hidden, the `PivotQueryBuilder` would hoist it to the front of the `column_ranking` ORDER BY. This caused the hidden sort helper to group columns by its own values *across* outer visible dimensions, breaking the expected column grouping and scrambling rendered pivot headers.

With `pivotColumnsOrder` provided, the builder now sorts each dimension at its **declared position** rather than hoisting hidden sort-only dims to the front. Hiding a dimension therefore leaves the visible column order identical to when it was visible.

`derivePivotConfigurationFromChart` now computes and emits `pivotColumnsOrder` whenever `sortOnlyDimensions` are present, filtering out passthrough dims (which don't drive sort). When `pivotColumnsOrder` is absent, the builder falls back to the previous legacy hoisting behavior, preserving backwards compatibility.

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -22710,6 +22710,10 @@ const models: TsoaRoute.Models = {
                     dataType: 'array',
                     array: { dataType: 'refAlias', ref: 'GroupByColumn' },
                 },
+                pivotColumnsOrder: {
+                    dataType: 'array',
+                    array: { dataType: 'refAlias', ref: 'GroupByColumn' },
+                },
                 sortOnlyColumns: {
                     dataType: 'array',
                     array: { dataType: 'refAlias', ref: 'ValuesColumn' },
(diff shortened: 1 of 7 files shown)


pimmey:
* **#24016** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24016?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24016?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24023, opened 2026-06-08
feat(ai-agents): one-click "Connect GitLab" MCP setup

charliedowler:
## Summary

Mirrors the existing one-click **"Connect GitHub"** MCP setup for **GitLab**, so an AI agent can connect to the hosted GitLab MCP server using the organization's existing GitLab connection — no manual MCP URL/auth.

Unlike GitHub (which collects a fine-grained PAT and swaps in the App-installation token at runtime), GitLab is a **true one-click**: the org's GitLab OAuth token is reused, so there's no token to paste. GitLab can be self-hosted, so the MCP URL is derived from the connected instance (`{instanceUrl}/api/v4/mcp`) rather than a fixed host.

Off by default behind `gitlab-mcp-one-click`, so it can ship incrementally.

## Changes
- **common** — `FeatureFlags.GitlabMcpOneClick`; `GITLAB_MCP_SERVER_NAME` + `getGitlabMcpServerUrl`/`isGitlabMcpServerUrl` helpers; `AiMcpGitlabAvailability` / `ApiConnectGitlabMcpServerBody` / response types.
- **backend** — `getGitlabMcpAvailability` / `connectGitlabMcpServer` / `refreshGitlabMcpCredentials` on `AiAgentService` (the bearer token is refreshed per run from the org's GitLab OAuth install via `getOrRefreshToken`, same pattern as writeback); two TSOA routes (`GET/POST .../mcpServers/gitlab/availability|connect`); a `gitlab_mcp_connected` analytics event; `GitlabAppInstallationsModel` injected into the service.
- **frontend** — `useGitlabMcpAvailability` / `useConnectGitlabMcpServerMutation` hooks; a "GitLab" connect button in `AiAgentNewThreadMcpConnections`, gated on availability (one-click, no modal).

## Testing
- Typecheck clean across common / backend / frontend; `generate-api` regenerated routes + swagger; lint/format applied.
- Not yet exercised end-to-end against a live GitLab MCP (feature flag is off).

## Open question for review
The exact GitLab hosted-MCP endpoint (`{instanceUrl}/api/v4/mcp`) and that it accepts the org OAuth token as a bearer is my best-reasonable design from GitLab's MCP docs — it mirrors the GitHub flow and compiles, but should be confirmed against a live GitLab MCP before enabling the flag.

Linear: PROD-8162

### DIFF
--- a/packages/backend/src/analytics/LightdashAnalytics.ts
+++ b/packages/backend/src/analytics/LightdashAnalytics.ts
@@ -1835,6 +1835,17 @@ export type AiAgentGithubMcpConnectedEvent = BaseTrack & {
     };
 };
 
+export type AiAgentGitlabMcpConnectedEvent = BaseTrack & {
+    event: 'ai_agent.gitlab_mcp_connected';
+    userId: string;
+    properties: {
+        organizationId: string;
+        projectId: string;
+        mcpServerId: string;
+        method: 'one_click' | 'one_click_reconnect';
+    };
+};
+
 export type AiAgentDeletedEvent = BaseTrack & {
     event: 'ai_agent.deleted';
     userId: string;
@@ -2296,6 +2307,7 @@ type TypedEvent =
     | DeprecatedRouteCalled
     | AiAgentCreatedEvent
     | AiAgentGithubMcpConnectedEvent
+    | AiAgentGitlabMcpConnectedEvent
     | AiAgentDeletedEvent
     | AiAgentUpdatedEvent
     | AiAgentDocumentCreatedEvent
--- a/packages/backend/src/ee/controllers/aiAgentController.ts
+++ b/packages/backend/src/ee/controllers/aiAgentController.ts
@@ -30,6 +30,7 @@ import {
     ApiAiAgentVerifiedArtifactsResponse,
     ApiAiAgentVerifiedQuestionsResponse,
     ApiAiMcpGithubAvailabilityResponse,
+    ApiAiMcpGitlabAvailabilityResponse,
     ApiAiMcpOAuthCredentialRequest,
     ApiAiMcpServerListResponse,
     ApiAiMcpServerResponse,
@@ -39,6 +40,7 @@ import {
     ApiAppendInstructionResponse,
     ApiCloneThreadResponse,
     ApiConnectGithubMcpServerBody,
+    ApiConnectGitlabMcpServerBody,
     ApiCreateAiAgent,
     ApiCreateAiAgentResponse,
     ApiCreateAiMcpServer,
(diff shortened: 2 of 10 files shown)


charliedowler:
### Tested the connect button live (local)

Exercised the **GitLab** one-click button end-to-end with a seeded org GitLab install + the flag on:

- ✅ **Availability** path renders the button (`GET .../mcpServers/gitlab/availability` → `{available:true}`).
- ✅ **Click → connect** fires `POST .../mcpServers/gitlab/connect`, runs the new service method, and the error path round-trips back into a toast (no crash).

Pushing to the live OAuth refresh (against gitlab.com) surfaced — and this PR now fixes — a real rough edge:

- `getOrRefreshToken` **throws** when the org's stored OAuth token can't be refreshed (revoked/expired). Previously that raw error (`HTTP 400: {"error":"invalid_grant",...}`) leaked into the connect toast, **and** would have thrown during the per-run `refreshGitlabMcpCredentials` (i.e. could break agent runs).
- Fix: `resolveGitlabMcpCredentials` now swallows refresh failures → returns `null` (logged), so connect shows a friendly "make sure GitLab is connected / reconnect if expired" message and agent runs degrade gracefully instead of crashing.

### Note for reviewers
`getGitlabMcpAvailability` only checks that an install **row** exists, while connect additionally validates the OAuth token. So `available:true` can still fail at connect if the token is dead — acceptable (availability stays cheap, no network call per new-thread render), but flagging it.

A genuine *successful* connect still needs a real GitLab OAuth app + a live GitLab MCP endpoint that accepts the token — not verifiable locally.

charliedowler:
### Verified end-to-end against the **real** GitLab MCP (live OAuth token)

Connected a real GitLab OAuth app + org connection (real `glpat`-equivalent OAuth token, gitlab.com) and clicked the one-click **GitLab** button. Result is conclusive:

- ✅ **Auth works** — the org OAuth token validated against `gitlab.com/api/v4/user` and was accepted (got all the way past `resolveGitlabMcpCredentials`).
- ✅ **URL is correct** — `https://gitlab.com/api/v4/mcp` is the right endpoint.
- ❌ **Transport mismatch (the blocker)** — GitLab's MCP responded:
  > `MCP HTTP Transport Error: POSTing to endpoint (HTTP 404): {"message":"404 Not Found"}. This server does not support HTTP transport. Try using `sse` transport instead`

**Root cause:** `AiAgentMcpRuntimeClient` only implements **Streamable-HTTP** (`StreamableHTTPClientTransport`, ~L553 & L834) — there's no SSE transport. GitLab's hosted MCP is **SSE-only**. So this one-click feature is correct (auth/URL/flow all validated against the real service) but **cannot complete the handshake until the MCP runtime supports SSE** (ideally Streamable-HTTP with an SSE fallback, which would also help other SSE-only MCP servers).

The failure is handled gracefully (400 + a clear toast), so nothing crashes — but the connect can't succeed yet.

**Suggested follow-up (separate PR, shared-runtime change):** add `SSEClientTransport` support to `AiAgentMcpRuntimeClient` with HTTP→SSE fallback. This PR's GitLab wiring is otherwise complete and stays behind `gitlab-mcp-one-click` (off by default).

PR #24043, opened 2026-06-09
fix: reject nested filter values

rephus:
Closes: https://linear.app/lightdash/issue/SPK-478/veria-4-sql-injection-in-data-warehouse-queries-via-filter-array

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.ts
+++ b/packages/common/src/compiler/filtersCompiler.ts
@@ -687,17 +687,11 @@ const escapeStringValuesOnFilterRule = (
     escapeString: (string: string) => string,
 ): FilterRule<FilterOperator, unknown> => ({
     ...filterRule,
-    values: filterRule.values?.map((v) => {
-        if (typeof v === 'object' && v !== null) {
-            throw new CompileError(
-                'Complex objects or arrays are not permitted as filter values',
-            );
-        }
-
-        return typeof v === 'string'
+    values: filterRule.values?.map((v) =>
+        typeof v === 'string'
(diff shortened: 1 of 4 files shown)


rephus:
* **#24043** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24043?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
Failing relevant tests, investigating

packages/common/src/compiler/filtersCompiler.ts
```diff
@@ -687,11 +687,17 @@ const escapeStringValuesOnFilterRule = (
     escapeString: (string: string) => string,
 ): FilterRule<FilterOperator, unknown> => ({
     ...filterRule,
-    values: filterRule.values?.map((v) =>
-        typeof v === 'string'
+    values: filterRule.values?.map((v) => {
+        if (typeof v === 'object' && v !== null) {
+            throw new CompileError(
+                'Complex objects or arrays are not permitted as filter values',
```
rephus: we need to confirm customers are not using this filtering already

rephus: since we can't confirm if this is already used by some customers, we are going to preserve behavior, but instrument the malformed shape so we can see whether it exists in the wild before turning it into a hard rejection

PR #24052, opened 2026-06-09
fix(scheduler): only notify delivery failures on the final attempt (PROD-8164)

### LINKED ISSUE #23976: Sync/delivery failure notifications fire on every retry, not just the final failure
## Bug description

Google Sheets sync failure notifications are sent on **every failed retry attempt**, not just the final one. `UPLOAD_GSHEETS` jobs are queued with `maxAttempts: 6` and retried with natural backoff over \~10 minutes (intended to ride out transient warehouse/quota errors). However, the error handler sends a failure email **and** posts a Slack notification-channel message on each failed attempt before re-throwing so the job can be retried.

The result: a sync that fails on attempts 1–5 but **succeeds on attempt 6** still generates up to 5 failure notifications, and there is no corresponding "it actually succeeded" notification. Users cannot distinguish a sync that is genuinely broken from one that fails transiently and then recovers on retry. With a folder of syncs that intermittently hit slow/contended warehouse connections, this produces a steady stream of alarming failure emails for syncs that are, in fact, working.

The notification is fired unconditionally in the `catch` block — there is no check of the current attempt against the job's max attempts, so it cannot tell a non-final retry from the final one.

## Steps to reproduce

1. Set up a Google Sheets sync on a SQL Runner chart (or any gsheets sync).
2. Make the underlying query/connection occasionally slow or flaky, so the first attempt(s) fail but a later retry succeeds (e.g. transient warehouse connection/queue contention).
3. Let the scheduled sync fire.
4. Observe: a failure email + Slack message is sent for each failed attempt, and **no** notification is sent when a later retry succeeds.

## Expected behavior

Only notify on a genuine, terminal failure:

* If a later retry succeeds, send **no** failure notification at all.
* If all retries are exhausted, send a single failure notification on the final attempt.

(Non-retryable errors that disable the sync — e.g. `NotFoundError`, `ForbiddenError`, `MissingConfigError` — should still notify immediately, as they do today, since there will be no further retries.)

## Actual behavior

A failure email and Slack notification-channel message are sent on every failed attempt. No success notification is sent when a retry recovers, so each transient failure looks like a real failure.

## Code reference

* Gsheets jobs are queued with 6 max attempts: `packages/backend/src/scheduler/SchedulerClient.ts` (`UPLOAD_GSHEETS`, `maxAttempts: 6`).
* The notification + re-throw happens unconditionally in the gsheets `catch` block: `packages/backend/src/scheduler/SchedulerTask.ts` (`uploadGsheets` → `slackClient.postMessageToNotificationChannel(...)` and `emailClient.sendGoogleSheetsErrorNotificationEmail(...)`, then `throw e` to let graphile retry). There is no `job.attempts` vs `job.max_attempts` guard.

The fix likely needs the worker to pass the current attempt / max-attempts into the error handler (graphile exposes `helpers.job.attempts` and `helpers.job.max_attempts`) and only notify when `attempts >= max_attempts` (or on a non-retryable/disable error).

## Possibly also affects scheduled deliveries

The same "notify on a non-final attempt" pattern exists for scheduled deliveries, though with a smaller blast radius:

* Most `HANDLE_SCHEDULED_DELIVERY` jobs use `maxAttempts: 1` (no retry), so they aren't affected.
* Image-format **dashboard** deliveries are queued with `maxAttempts: 2` (`SchedulerClient.ts`), and the delivery error handler posts a Slack failure message before re-throwing (`SchedulerTask.ts`, IMAGE/PDF `catch` blocks). A transient first-attempt failure that succeeds on retry would still post one failure message.

Worth fixing both paths with the same "only notify on the final attempt" guard.

## Note

Distinct from lightdash/lightdash#23793 (the column-discovery hang that *causes* some of these transient gsheets failures). This issue is specifically about the notification noise — even after the underlying failures become rare/transient, the per-retry notifications remain misleading.

tatianainama:
### Description

Gsheets sync and image/PDF delivery failures were notified on every failed retry. A sync that failed transiently on attempts 1–5 but recovered on attempt 6 still sent up to 5 false-alarm emails and Slack messages, with no way to tell a genuinely broken sync from one that recovered.

Notification ownership now lives in the scheduler worker handler, which knows the attempt count:

- `uploadGsheets` is report-only: on failure it throws a `SchedulerDeliveryError` carrying the original error as `cause` plus the notification context, instead of notifying itself.
- The worker notifies only on a terminal outcome — a non-retryable error, or the final attempt of a retryable one — and disables the sync for non-retryable errors. The original error is re-thrown to graphile, so retry handling, logs and Sentry grouping are unchanged.
- The scheduled image/PDF delivery path gates its notification-channel posts on `isFinalAttempt` too. Only image-dashboard deliveries retry (`maxAttempts: 2`); single-attempt jobs always pass the final attempt, so their behaviour is unchanged.

Closes: PROD-8164
Closes: #23976

### DIFF
--- a/packages/backend/src/scheduler/SchedulerDeliveryError.test.ts
+++ b/packages/backend/src/scheduler/SchedulerDeliveryError.test.ts
@@ -1,0 +1,73 @@
+import { NotFoundError } from '@lightdash/common';
+import {
+    resolveSchedulerDeliveryFailureAction,
+    SchedulerDeliveryError,
+} from './SchedulerDeliveryError';
+
+const buildError = (isNonRetryable: boolean) =>
+    new SchedulerDeliveryError({
+        cause: new Error('warehouse timeout'),
+        isNonRetryable,
+        createdByUserUuid: 'user-1',
+        notification: {
+            organizationUuid: 'org-1',
+            recipientEmail: 'creator@example.com',
+            schedulerName: 'Weekly sync',
+            deliveryUrl: 'https://app/scheduler',
+        },
+    });
+
+describe('SchedulerDeliveryError', () => {
+    // The worker re-throws `err.cause`, so preserving the original error is the
+    // one constructor behaviour worth pinning down.
+    test('preserves the original error as its cause', () => {
+        const cause = new NotFoundError('sheet is gone');
+        const err = new SchedulerDeliveryError({
+            cause,
+            isNonRetryable: true,
+            createdByUserUuid: 'user-1',
+            notification: {
+                organizationUuid: 'org-1',
+                recipientEmail: undefined,
+                schedulerName: 'Weekly sync',
+                deliveryUrl: 'https://app/scheduler',
+            },
+        });
+
+        expect(err).toBeInstanceOf(SchedulerDeliveryError);
+        expect(err.cause).toBe(cause);
+    });
+});
+
+describe('resolveSchedulerDeliveryFailureAction', () => {
+    test('does nothing when the error is not a SchedulerDeliveryError', () => {
+        expect(resolveSchedulerDeliveryFailureAction(undefined, true)).toEqual({
+            notify: false,
+            disable: false,
+        });
+    });
+
+    test('stays silent on a retryable, non-final attempt', () => {
+        expect(
+            resolveSchedulerDeliveryFailureAction(buildError(false), false),
+        ).toEqual({ notify: false, disable: false });
+    });
+
+    test('notifies once on the final attempt of a retryable error', () => {
+        expect(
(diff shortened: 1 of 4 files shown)


tatianainama:
* **#24052** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24052?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24086, opened 2026-06-09
feat(common): consolidate the calendar-value predicate (GLITCH-450)

pimmey:
Closes: GLITCH-450

### Description:

Consolidates the duplicated "is this a calendar value?" check into a single predicate, `isCalendarValueDimension` (`packages/common/src/utils/formatting.ts`), and routes the two remaining inlined `timeIntervalBaseDimensionType === DimensionType.DATE` sites through it:

- the `formatItemValue` DATE timezone gate, and
- the ECharts category-date axis snap (`getCategoryDateAxisConfig`).

Closes design principle 18 (`gap-calendar-predicate`) and unblocks the Phase 2 cast-to-DATE work (GLITCH-452): with one predicate keyed on the **base** type, Phase 2 can change warehouse output types without re-deciding the rule at each site.

### Behavior change (principle 15 — acceptance criterion #5)

The predicate is intentionally **wider** than the old inlined checks: it treats **plain DATE columns** and **DATE metrics** as calendar values, not only DATE-base truncations.

- **`formatItemValue` (real, intended change):** a plain DATE value or DATE metric is no longer timezone-shifted under a non-UTC display timezone. This is the contract-correct fix — it removes a negative-offset off-by-one for wall-clock dates. ("DATE metric" here means an explicit `MetricType.DATE` metric; MIN/MAX date metrics render via a separate `formatTimestamp` path and are unaffected.)
- **ECharts axis (no behavior change):** behavior-preserving for every shape the compiler actually produces — the snap only runs for `WEEK/MONTH/QUARTER/YEAR`, those dims always carry a base type, and plain DATE columns have no coarse interval (they early-return). Migrated for single-source-of-truth only.

### Tests

- New unit matrix for `isCalendarValueDimension` (DATE/TIMESTAMP × grains, plain DATE base, DATE metric, `skipTimezoneConversion`).
- New `formatItemValue` test for the widened plain-DATE / DATE-metric no-shift behavior.
- New ECharts characterization test (TIMESTAMP-base dim shifts — the complement of the existing DATE-base UTC test).
- Corrected one pre-existing canary fixture that asserted a DATE-DAY shift but omitted `timeIntervalBaseDimensionType` (a shape the compiler never emits); given the TIMESTAMP base it represents, it still asserts the shift.
- Full common suite (2216 tests) + ECharts suite green; `typecheck:fast` and lint clean.

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -10,6 +10,7 @@ import {
     NumberSeparator,
     type CustomFormat,
     type Dimension,
+    type Metric,
 } from '../types/field';
 import { TimeFrames } from '../types/timeFrames';
 import {
@@ -25,6 +26,7 @@ import {
     formatValueWithExpression,
     getCustomFormatFromLegacy,
     getEffectiveSeparator,
+    isCalendarValueDimension,
     isMomentInput,
     shouldShiftItemTimezone,
     toIsoWithProjectOffset,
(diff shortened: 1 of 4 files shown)


pimmey:
* **#24086** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24086?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24086?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24100, opened 2026-06-10
feat(instance-config): declare user attributes & group mappings via LD_SETUP_USER_ATTRIBUTES

### LINKED ISSUE #24054: Declare user attributes & group mappings via LD_SETUP_* instance config (reconcile on deploy)
## Problem

Self-hosted instances can already declare and **reconcile** a lot of org/project config on every deploy via the `LD_SETUP_*` env vars consumed by `InstanceConfigurationService`:

* `initialSetup` (`getInitialSetupConfig`, `packages/backend/src/config/parseConfig.ts`) — creates org/admin/projects on first boot when no org exists.
* `updateSetup` (`getUpdateSetupConfig`, same file) — targets an existing instance via `LD_SETUP_ORGANIZATION_UUID` / `LD_SETUP_PROJECT_UUID` and re-applies on each deploy: org admin, email domains, default role, admin API key, project warehouse/dbt connection, service account tokens, embed config.

**What's missing:** neither flow can declare **user attributes** or **group → attribute-value mappings**. Those are API-only at runtime (`UserAttributesService` → `POST/PUT /api/v1/org/attributes`). So orgs managing Lightdash as code have to run a manual post-deploy CLI/API step to (re)apply group→attribute mappings, which drifts and isn't reproducible across environments.

This is the common pattern for gating sensitive/PII data: an IdP/SCIM-synced group is mapped to a user attribute, and that attribute gates sensitive dimensions/tables in dbt. Today the project connection updates on deploy but the attribute mapping does not.

## Proposed solution

Extend `initialSetup`/`updateSetup` to cover user attributes, consistent with how `LD_SETUP_PROJECTS` already works (a JSON env var):

```
LD_SETUP_USER_ATTRIBUTES='[
  {
    "name": "is_privileged",
    "description": "PII access",
    "attributeDefault": null,
    "groups": [
      { "group": "Privileged Data Analyst", "value": "true" }
    ]
  }
]'
```

* On deploy, `InstanceConfigurationService` upserts each attribute against `LD_SETUP_ORGANIZATION_UUID` and resolves `group` **by name** to its `groupUuid` (important: SCIM-synced groups are created by the IdP, so their UUIDs aren't known at deploy time — referencing by name is required).
* Reconciling semantics matching the existing `POST/PUT /api/v1/org/attributes` behaviour (full replace of the attribute's user/group set).

## Why this matters

Enterprises governing PII map IdP/SCIM groups to user attributes that gate sensitive data in dbt. The project connection already reconciles on deploy via `updateSetup`; the attribute mapping is the one piece still requiring a manual CLI step. Closing this makes governance fully declarative and removes drift between environments.

Related: lightdash/lightdash#23931 (custom roles as code) — same config-as-code family, different object.

jelstongreen:
## What

Self-hosted instances can already reconcile org/project config on every deploy via `LD_SETUP_*` (`InstanceConfigurationService`). User attributes and group→attribute mappings were the gap — API/UI-only, so orgs managing Lightdash as code had to run a manual post-deploy CLI/API step to (re)apply them.

This adds a new `LD_SETUP_USER_ATTRIBUTES` JSON env var so they can be declared as instance config and reconciled on every deploy.

```jsonc
LD_SETUP_USER_ATTRIBUTES='[
  {
    "name": "is_privileged",
    "description": "sensitive-data access",
    "attributeDefault": null,
    "groups": [ { "group": "Privileged Analyst", "value": "true" } ]
  }
]'
```

## How

- **`parseConfig`**: `getUserAttributesSetupConfig()` parses + validates the env var (Zod; `ParseError` on malformed JSON / duplicate names; defaults applied) and surfaces it on `updateSetup.userAttributes`.
- **`InstanceConfigurationService.updateUserAttributesConfiguration()`** (called from the every-deploy `updateInstanceConfiguration`):
  - Upserts each attribute **by name** (create vs update).
  - Resolves group mappings **by name** → `groupUuid`. Group UUIDs aren't known at deploy time (SCIM-synced groups are created by the IdP), so names are required. Unresolved groups are **warn-and-skipped** and reconciled on a later deploy — never fatal.
  - **Non-destructive**: attributes absent from config are left untouched (no pruning).

No schema change — an earlier draft added a `managed` provenance column; that was dropped per review (the reconcile runs every deploy, so when UI-lock enforcement is eventually built the column can be added then and re-marked losslessly).

## Ordering / SCIM

Group bindings only apply via the update path (an existing org); at first boot SCIM has synced nothing, so group names fail-soft and reconcile on a later deploy. SCIM provisioning is async/decoupled from deploys, so warn-and-skip is required regardless.

## Tests

- `parseConfig.test.ts`: valid parse, defaults, malformed JSON, missing name, duplicate names, `updateSetup` wiring.
- `InstanceConfigurationService.test.ts`: create-by-name, update-by-name, fail-soft skip on unresolved group, no-op when unconfigured.
- Manually verified end-to-end on a licensed instance (see comment below), including the full SCIM provisioning flow.

## Out of scope (follow-ups)

- `managed` UI/API lock enforcement (shared with the custom-roles-as-code work).
- Per-user (by-email) mappings — groups-only for v1.
- Docs: `LD_SETUP_USER_ATTRIBUTES` in the self-host env-var reference.

Closes: PROD-8194
Closes: #24054


jelstongreen:
Dropped the migration and the `managed` column in `aed6d0aead` — agreed it's not needed.

Reasoning for landing it now was provenance backfill, but that doesn't hold: the reconcile path runs on every deploy, so when UI-lock enforcement is eventually built, the column can be added then and re-marked losslessly on the next deploy. No schema change for v1.

Approach now matches what you outlined — `getUpdateSetupConfig` reads `LD_SETUP_USER_ATTRIBUTES`, `updateInstanceConfiguration` reads the config, and `UserAttributesModel` upserts the existing table (by name; group mappings resolved by name, fail-soft when not yet synced; non-destructive). No `UserAttributesModel`/entity changes remain.

jelstongreen:
## Verification evidence

### Unit tests
`parseConfig.test.ts` + `InstanceConfigurationService.test.ts` — **126 passing** (parse valid/defaults/malformed/duplicate, config wiring, upsert create-vs-update, group-by-name resolution, fail-soft skip, non-destructive). Typecheck + lint clean; `Build & Test` CI green.

### Live end-to-end (licensed instance)
Booted with `LD_SETUP_USER_ATTRIBUTES` set; observed the reconcile + the DB:

```
[InstanceConfigurationService] info: Update instance: Created user attribute "is_privileged"
[InstanceConfigurationService] warn: Update instance: group "Ghost Group" for user attribute "region_pii" not found (not yet synced?); skipping this mapping
[InstanceConfigurationService] info: Update instance: Created user attribute "region_pii"
-- restart (same config) --
[InstanceConfigurationService] info: Update instance: Updated user attribute "is_privileged"
```

- create → update across restarts (idempotent); `is_privileged` group-mapping count stays **1** (full-replace, no dupes).
- fail-soft: unresolved group skipped, attribute still created, no crash.
- non-destructive: pre-existing attributes untouched.

### SCIM provisioning flow (the real customer scenario)
Exercised the actual SCIM API to prove a group an IdP provisions is what the config resolves by name:

```
1. POST /api/v1/scim/v2/Groups {"displayName":"Privileged Analyst"}  → 201 (id 0f221450…)
2. groups table                                                      → name = "Privileged Analyst"
3. LD_SETUP_USER_ATTRIBUTES groups:[{group:"Privileged Analyst",value:"true"}] → reconcile on boot
4. group_user_attributes:
     attribute      | scim_group          | value
     ---------------+---------------------+-------
     is_privileged  | Privileged Analyst  | true
```

Confirms the full chain: **IdP → SCIM-provisioned group → attribute mapping applied on deploy**. `ScimService.createGroup` writes `groups.name = displayName`, the same field `updateUserAttributesConfiguration` resolves via `groupsModel.find({ name })`.

_Minor finding while testing (not blocking, unrelated to this PR): the SCIM org-access-token endpoint rejects PAT auth (401) and needs a session — worth confirming that's intentional._

rephus (approved):

PR #24124, opened 2026-06-10
refactor: convert date/timestamp filter renderers to named args

almeidabbm:
Converts the three date/timestamp filter SQL renderers in `filtersCompiler.ts` from positional arguments to named (object) arguments. Their signatures had grown to nine and ten parameters, which made call sites hard to read and easy to get wrong.

## Functions changed

- `renderDateFilterSql`
- `renderTimestampFilterSql`
- `renderDateOrTimestampFilterSql` (the shared private helper, the one with the longest list)

## Behavior

Pure refactor, no SQL output changes. The 270 existing tests in `filtersCompiler.test.ts` pass unchanged (84 call sites updated to the object form).

## Type note

Moving to object literals surfaced `no-unsafe-assignment` errors that the positional form never triggered (this config has `no-unsafe-argument` off but `no-unsafe-assignment` on). The `filter` value flowing in is an `any`-tainted `FilterRule`, so the wrappers and helper now type `filter` as `FilterRule<FilterOperator, unknown>` to match what callers actually pass.

The helper's body still needs the date settings, so it narrows once with the existing `isDateFilterRule` guard:

```ts
const settings = isDateFilterRule(filter) ? filter.settings : undefined;
```

This is behavior-identical because the settings reads only happen in relative-operator branches, where the filter always carries `unitOfTime`. It also let us drop a pre-existing `(filter as DateFilterRule)` cast.

## `isDateFilterRule` parity fix

The narrowing above exposed a latent gap: `isDateFilterRule` only matched on `unitOfTime`, so a filter carrying `completed` without `unitOfTime` would collapse `settings` to `undefined` and silently flip `completed` to `false`, diverging from the previous direct `filter.settings?.completed` reads. The guard now also matches on `completed`:

```ts
'unitOfTime' in (filter.settings || {}) ||
'completed' in (filter.settings || {})
```

This is a shared guard, so its other consumers were audited. The only filter whose classification changes is the (non-constructable in practice) `completed`-without-`unitOfTime` case, and in every consumer that shift is benign or strictly more correct:

- `filterGrammarConversion.ts` — such a filter would throw `completed not supported` instead of the generic `No function implemented`; both throw.
- `fields.ts` custom-metric matching — `settingsMatch` only shifts for that non-constructable case.
- `ItemDetailPreview.tsx` — would render the `completed` label, which is more correct.

`completed` only ever appears on date filters, so the broadened definition is also semantically sound.

## Checks

- oxfmt, eslint, typecheck: clean
- 270/270 filter compiler tests pass; 462/462 across filter/fields/grammar suites pass

Relates: GLITCH-453

### DIFF
--- a/packages/common/src/compiler/filtersCompiler.test.ts
+++ b/packages/common/src/compiler/filtersCompiler.test.ts
@@ -153,16 +153,16 @@ describe('Filter SQL', () => {
         (unitOfTime) => {
             jest.setSystemTime(new Date('04 Apr 2020 06:12:30 GMT').getTime());
             expect(
-                renderDateFilterSql(
-                    DimensionSqlMock,
-                    {
+                renderDateFilterSql({
+                    dimensionSql: DimensionSqlMock,
+                    filter: {
                         ...InTheCurrentFilterBase,
                         settings: { unitOfTime },
                     },
-                    adapterType.default,
-                    'UTC',
-                    formatTimestamp,
-                ),
+                    adapterType: adapterType.default,
+                    timezone: 'UTC',
+                    boundaryDateFormatter: formatTimestamp,
+                }),
             ).toStrictEqual(ExpectedInTheCurrentFilterSQL[unitOfTime]);
         },
     );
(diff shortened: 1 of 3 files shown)


almeidabbm:
* **#24141** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24141?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24126** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24126?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24124** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24124?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24124?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24129, opened 2026-06-10
feat(ai): support non-streaming LLM gateways via per-provider supportsStreaming capability

### LINKED ISSUE #23948: Add option to disable streaming on AI agent responses for LLM gateway compatibility
## Use case

Organizations routing Lightdash AI agent traffic through an LLM gateway or proxy may find that their gateway does not support streaming (SSE) responses. Since the AI agent exclusively uses `streamText` with `stream: true` for all user-facing interactions, there is no way to fall back to a non-streaming mode, making the AI agent unusable in these environments.

## Proposed solution

Add a configuration option (e.g. an environment variable like `AI_DISABLE_STREAMING=true`) that switches the AI agent's user-facing endpoint from the streaming path (`streamAgentThreadResponse`, `stream: true`) to the non-streaming path already present in `generateAgentThreadResponse` (`stream: false`). The non-streaming path already exists internally (used for evals / non-web callers), it just needs to be exposed as a runtime toggle.

## Workaround

None available on Lightdash's side today. The only short-term path is adding streaming support to the LLM gateway itself.

## Additional context

The non-streaming code path (`generateAgentThreadResponse`) already exists in `AiAgentService` and is used internally for evaluation. The `streamAgentResponse` function in `agentV2.ts` always calls `streamText` from the Vercel AI SDK. A toggle that bypasses SSE delivery and returns the full response at once would unblock deployments behind streaming-incompatible gateways.

tatianainama:
Adds a per-provider `supportsStreaming` capability (default `true`) so the AI agent works behind OpenAI-compatible LLM gateways that don't support streaming (SSE) completions.

When a provider is marked non-streaming (e.g. `OPENAI_SUPPORTS_STREAMING=false`), `getModel` wraps the resolved model with the AI SDK's `simulateStreamingMiddleware`: each model call becomes a single non-streaming request whose result is replayed as a simulated stream. `streamText`, multi-step tool calls, and the browser-facing SSE endpoint are unchanged — no call-site or frontend changes, existing deployments unaffected. Follows the per-provider config pattern from #23448 (`customHeaders`).

- New opt-out env vars: `OPENAI_` / `AZURE_AI_` / `ANTHROPIC_` / `OPENROUTER_` / `BEDROCK_SUPPORTS_STREAMING`
- Rationale + proxy-timeout caveat documented in `docs/ai-agent-non-streaming-transport.md`

**Testing**: config unit tests for the env vars; unit test proving a wrapped model with a throwing `doStream` completes a multi-step tool loop with zero streaming calls; manual end-to-end run against a strict local gateway that rejects `stream: true` — all requests went out non-streaming and the agent worked, while without the flag it reproduces the reported failure.

Closes: PROD-8155
Closes: #23948

### DIFF
--- a/docs/ai-agent-non-streaming-transport.md
+++ b/docs/ai-agent-non-streaming-transport.md
@@ -1,0 +1,87 @@
+# AI Agent: Non-Streaming LLM Gateways (`supportsStreaming`)
+
+This document describes the per-provider `supportsStreaming` capability (PROD-8155,
+[#23948](https://github.com/lightdash/lightdash/issues/23948)): how to run the AI agent against an
+OpenAI-compatible LLM gateway that does not support streaming (SSE) completions.
+
+---
+
+## The two streaming layers
+
+"Streaming" in the AI agent means two independent things:
+
+1. **Backend ↔ provider**: `streamText` issues a *streaming completion request* to the model
+   endpoint. This is the part a non-streaming gateway can't serve.
+2. **Frontend ↔ backend**: the `/stream` SSE endpoint that delivers tokens to the browser. This is
+   our own infrastructure and works regardless of how the provider call is made.
+
+`supportsStreaming` only changes layer 1. The browser-facing SSE endpoint, the chat UI, tool cards,
+and `smoothStream` word-chunking are unchanged.
+
+## Configuration
+
+Streaming support is a property of the gateway/endpoint a provider points at — the same place you
+configure `*_BASE_URL` and `*_CUSTOM_HEADERS` (see PR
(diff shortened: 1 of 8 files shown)


tatianainama:
* **#24129** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24129?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24137, opened 2026-06-10
fix(common): format MIN/MAX timestamp aggregations in project timezone (GLITCH-485)

pimmey:
Closes: GLITCH-485

### Description:

MIN/MAX aggregations over a TIMESTAMP column were never formatted in the resolved project timezone — they rendered in UTC (fresh queries) or as raw ISO 8601 (cached results), while the raw dimension beside them formatted correctly.

**Root cause (deeper than the ticket's "only when `value instanceof Date`"):** MIN/MAX metrics are *numeric*, so `getCustomFormat` returns an auto-derived `{ type: NUMBER }` format — never `undefined`. The existing guard `value instanceof Date && customFormat === undefined` was therefore **always false**, making the `formatTimestamp` branch unreachable. Both paths fell into `applyCustomFormat`:
- a `Date` → `formatTimestamp` with **no timezone** → UTC
- an ISO string (rehydrated from cached S3 results) → `applyDefaultFormat` → **raw value**

### Fix

In the MIN/MAX branch, format any temporal value — a `Date` (fresh query) or an ISO **datetime** string (cached) — through `formatTimestamp` in the resolved timezone, matching dimension formatting. Gated only on a user-chosen display format (`DATE`/`TIMESTAMP`/`CUSTOM`), which still wins; the auto-derived numeric format no longer suppresses temporal formatting. A `^\d{4}-\d{2}-\d{2}[T ]\d{2}:\d{2}` guard keeps numbers and date-only values out.

Naturally Phase-2-gated: without the timezone feature flag the resolved timezone is UTC, so behavior is unchanged; with it on, MIN/MAX now match dimensions.

### Tests (`formatting.test.ts`)

- MIN/MAX timestamp formats in the resolved project tz **identically from a `Date` and from an ISO string** (fresh vs cached parity).
- MIN/MAX does **not** coerce non-temporal values — numeric aggregates stay numbers (fresh number + cached string); a date-only aggregate is never tz-shifted (verified under a -11:00 zone).
- Full common suite green (2262 passed); typecheck + lint clean.

### Known edge (out of scope)

A MAX-**of-DATE** that arrives as a `Date` instance at midnight is indistinguishable from a midnight timestamp and would be tz-formatted; real DATE columns come back as date-only strings (handled correctly), so this only bites if a driver returns a `Date`-at-midnight.

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -28,6 +28,7 @@ import {
     getEffectiveSeparator,
     isCalendarValueDimension,
     isMomentInput,
+    isTimestampString,
     shouldShiftItemTimezone,
     toIsoWithProjectOffset,
 } from './formatting';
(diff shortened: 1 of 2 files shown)


pimmey:
* **#24137** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24137?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #23669, opened 2026-05-29
fix(ai-review): writeback hardening — progress-write race + resilience

joaoviana:
Part 8/8 (hardening). Two fixes to the worker writeback, squashed together:

**Progress-write race:** progress writes were fire-and-forget and could land after the terminal write, leaving a row stuck on `running`. Progress now uses a guarded UPDATE (`WHERE pr_writeback_status NOT IN ('completed','failed')`) so it can never resurrect a terminal row.

**Resilience:**
- Worker crash / 30-min timeout → row marked `failed` (timeout handler); on read, any queued/running older than 35 min is shown `failed` ("timed out").
- Stale in-flight runs are retryable (idempotency check).
- Missing `E2B_API_KEY` / Anthropic key → fail fast at the click.
- `getInstallationToken` failure no longer breaks the review-queue endpoint.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24143, opened 2026-06-10
fix(ai-review): stop false "Writeback timed out" on non-UTC servers

sdolidze:
## Summary

On any backend whose process timezone is ahead of UTC (e.g. Europe/London during BST), every review-item writeback was presented as failed with "Writeback timed out" the moment it started — while the writeback itself ran fine and opened its PR a couple of minutes later. Clicking **Create PR** again during that window persisted the running remediation as failed and launched a concurrent duplicate writeback.

UTC deployments (our cloud, CI) were unaffected, which is why this only surfaced on local dev machines.

## Root cause

The review tables store timestamps in `timestamp without time zone` columns holding UTC wall time, but node-postgres parses tz-less timestamps in the **process's local timezone**, so on a UTC+1 machine every read comes back one hour in the past. The staleness check compared that skewed value against the app clock:

```ts
// AiAgentAdminService.ts (before)
const isWritebackStale = (item, now) =>
    (item.prWritebackStatus === 'queued' || item.prWritebackStatus === 'running') &&
    now - new Date(item.updatedAt).getTime() > WRITEBACK_STALE_MS; // 35 min
```

```
DB stores (UTC wall):     updated_at = 14:44:06
node-pg parses (BST):     14:44:06 local → 13:44:06Z
Date.now():               14:47:11Z
perceived age:            63.1 min  >  35 min threshold → "Writeback timed out"
actual age:               ~3 min (writeback running)
```

The comparison shipped in #23669 and was extended to remediations in #24014; the columns have been tz-less since #23548.

## Fix

Staleness is now computed entirely in SQL — DB clock against DB clock — so it is immune to driver timezone parsing and to app/DB clock drift. (A `timestamptz` column migration was considered and skipped: `useTz: false` is the prevailing convention across the schema, and SQL-side ages make the column type irrelevant for this check.)

- `AiAgentReviewClassifierModel.ts` — item and remediation read queries select `(extract(epoch from (now() - updated_at)) * 1000)::float8 as updated_at_age_ms`; the stale presentation override ("Writeback timed out" for queued/running past 35 min, never for `pr_open`/`preview_ready`) moved here, next to the data it depends on. New `failStaleReviewRemediation` persists a stale remediation as failed with the same guard re-checked in SQL (`status IN ('queued','running') AND updated_at < now() - make_interval(...)`).
- `AiAgentAdminService.ts` — deletes the `Date.now()`-based helpers (`isWritebackStale` / `isRemediationStale` / `withStaleWritebackOverride`); the three call sites now consume the model's already-corrected summaries. The persist-before-retry path calls `failStaleReviewRemediation`, and the remediation insert now maps a unique-constraint race (concurrent retry) to a 409 `AlreadyExistsError` instead of an unhandled 500.
- Out of scope: `createdAt`/`updatedAt` fields serialized from these tables still parse in process-local time on non-UTC servers (display-only skew, schema-wide pattern).

## Before

`GET /api/v1/aiAgents/admin/review-items` ~3 minutes after starting a writeback, backend under BST:

```json
{ "prWritebackStatus": "failed", "prWritebackMessage": "Writeback timed out" }
```

## After

Same request, same conditions (verified live on a BST dev instance against this branch):

```json
{ "prWritebackStatus": "running", "prWritebackMessage": "Compiling project" }
```

Backdating the row 40 minutes still presents `failed` / "Writeback timed out", and the list read persists nothing — the override stays presentation-only.

## Test plan

- [x] `pnpm -F backend typecheck:fast` and `pnpm -F backend lint`
- [x] `jest AiAgentReviewClassifierModel AiAgentAdminService` — 38 tests pass; new regression tests stub `updated_at_age_ms` independently of `updated_at` so both a revert to Date-based math and removal of the override fail
- [x] Manual on a BST dev instance: fresh running writeback presents `running` (was `failed`), 40-min-old writeback presents `failed`/"Writeback timed out", raw DB rows untouched by list reads

Closes: ZAP-490

### DIFF
--- a/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
+++ b/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
@@ -7,7 +7,10 @@ import {
     AiAgentReviewRemediationTableName,
     AiAgentTurnSignalTableName,
 } from '../database/entities/aiAgentReviewClassifier';
-import { AiAgentReviewClassifierModel } from './AiAgentReviewClassifierModel';
+import {
+    AiAgentReviewClassifierModel,
+    WRITEBACK_STALE_MS,
+} from './AiAgentReviewClassifierModel';
 
 const ORGANIZATION_UUID = '00000000-0000-0000-0000-000000000001';
 const PROJECT_UUID = '00000000-0000-0000-0000-000000000002';
@@ -148,6 +151,7 @@ const makeRemediationRow = (
     resolved_at: null,
     created_at: SEEN_AT,
     updated_at: SEEN_AT,
+    updated_at_age_ms: 0,
     ...overrides,
 });
 
@@ -393,6 +397,7 @@ describe('AiAgentReviewClassifierModel', () => {
                     status_updated_by_user_uuid: null,
                     created_at: SEEN_AT,
                     updated_at: SEEN_AT,
+                    updated_at_age_ms: 0,
                 },
             ]);
             tracker.on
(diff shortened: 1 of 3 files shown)


sdolidze:
* **#24143** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24143?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24143?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24144, opened 2026-06-10
fix: emit terminal error event when SDK metric-query POST fails

magnew:
## Summary                                                                                                                                                                              
                                                                                                                                                                                          
  - When a data-app SDK metric-query POST fails (non-`ok` response or thrown                                                                                                              
    fetch error), the bridge now emits a synthetic terminal `error`                                                                                                                       
    `QueryEvent` re-keyed to the POST id. Previously only the success path
    (POST → queryUuid → GET poll → ready/error) emitted terminal events, so a                                                                                                             
    failed POST left a pending entry in the consumer's in-flight set forever.                                                                                                           
  - The visible impact is in `MinimalApp` (scheduled-delivery screenshots):                                                                                                               
    a stuck pending id keeps `activeQueryIds.size > 0`, which blocks the                                                                                                                  
    screenshot-ready indicator and times the delivery out at 60s.                                                                                                                         
  - Failure modes covered: backend returns `{ status: 'error', ... }`,                                                                                                                    
    non-2xx HTTP response, and the `fetch` itself throwing (network, abort).

PR #24145, opened 2026-06-10
fix: improve data apps screenshot ready indicator

magnew:
### Description:

- Scheduled-delivery screenshots of data apps were intermittently blank or  
mid-animation. Root cause: `MinimalApp` gated its screenshot-ready  
indicator on the iframe `load` event, which fires when the HTML document  
loads — not when the SDK bundle has bootstrapped. The  
`iframeLoaded && activeQueryIds.size === 0` debounce could resolve in  
the window between iframe HTML load and SDK bootstrap, before any  
queries had been registered, so the indicator mounted on a bare iframe  
shell and the headless browser captured an empty page.
- Now gates on `lightdash:sdk:screenshot-available` (the SDK's own  
bootstrap announce). The wiring already existed for the Inspect /  
Screenshot toolbar buttons — `MinimalApp` just wasn't consuming it.
- 8s fallback for apps generated before `lightdash:sdk:screenshot-available`  
shipped in the template (their bundle still loads and runs queries — it  
just never announces). Without the fallback, those scheduled deliveries  
would time out at 60s every run.
- Stacks on top of the SDK metric-query POST error fix.

### DIFF
--- a/packages/frontend/src/pages/MinimalApp.tsx
+++ b/packages/frontend/src/pages/MinimalApp.tsx
@@ -2,7 +2,7 @@ import { FeatureFlags } from '@lightdash/common';
 import { Box, Loader, Stack, Text } from '@mantine-8/core';
 import { useDebouncedValue } from '@mantine-8/hooks';
 import { IconAppsOff } from '@tabler/icons-react';
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { Navigate, useParams } from 'react-router';
 import ScreenshotReadyIndicator from '../components/common/ScreenshotReadyIndicator';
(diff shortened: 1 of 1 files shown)


magnew:
* **#24145** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24145?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24144** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24144?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Issue #23805, opened 2026-06-02
[Custom Roles] Organization scopes have no effect when assigned via a project-level custom role


jpetey75:
## Bug description

A user assigned a custom role that includes any manage Organization scopes e.g. `manage:OrganizationDesign` and `manage:Organization` is unable to create manage these settings. The permission check fails and they see a forbidden error, even though the scopes allow these users to view the related org settings. 

## Steps to reproduce

1. Create a custom role with `manage:Organization` scopes
2. Assign the custom role to a user at the project level
3. Log in as that user and navigate to Settings and note that the user can view the org settings
4. Attempt to edit an org setting e.g. modify a theme or create a new one

## Problem

The user is expecting to be able to be able to manage org settings when provided with the manage:Organization scopes for a particular project. The problem is that custom roles are not set at the org level, they are assigned at the project level which means that it doesn't make sense to assign a project level role and then get access to an org level scope.  

## Additional context

The `manage:Organization` scopes are selectable in the custom role builder UI, which makes this feel like it should work. The workaround is to assign the user the built-in org Admin role.

jesshitchcock:
Example here shows the permissions denied error when a user is set to interactive viewer at the org level but has a custom role to `manage:Organization` <img width="5524" height="3524" alt="Image" src="https://github.com/user-attachments/assets/ddbf3c2a-ff1c-4fb8-a3e5-83c5f8c5ad56" />

jesshitchcock:
One possible solution would be to hide org-level scopes from project-level custom role configuration in the UI

PR #24153, opened 2026-06-10
feat: add @self scopes for editing content in previews you created

### LINKED ISSUE #24130: [Custom Roles] Add @self-preview scopes for editing content in previews you created (manage:Dashboard@selfPreview, manage:SavedChart@selfPreview)
## Problem

Enterprise customers using custom roles want a governed self-serve workflow for non-analyst dashboard builders:

1. User **cannot** edit dashboards in shared/production spaces (view-only)
2. User **can** create a preview project and edit the copied content there
3. A reviewer with promotion permissions promotes the change back to production

Steps 1 and 3 work today with existing scopes. Step 2 does not: `manage:Dashboard@space` / `manage:SavedChart@space` require Editor/Admin access on the space, and when a preview is created the upstream project's space access is copied verbatim — so a user who is view-only in production is also view-only inside their **own** preview. The result is the familiar "shuttle loop": users can't iterate on governed content anywhere without being granted edit rights in production.

The only workarounds today are (a) someone with space-admin manually granting the user Editor on the preview's spaces after every preview creation, or (b) the content-as-code CLI path (`view:ContentAsCode` + `manage:ContentAsCode@self`) — which works but is CLI-only, the wrong interface for non-analyst builders.

## Proposal

Add preview-self scope variants for content editing:

- `manage:Dashboard@selfPreview` — edit dashboards in preview projects created by the user
- `manage:SavedChart@selfPreview` — edit saved charts in preview projects created by the user
- (possibly `manage:Space@selfPreview` so they can organise content in their own preview)

The condition machinery already exists and is used by `update:Project@self`, `delete:Project@self`, `manage:DeployProject@self`, and `manage:ContentAsCode@self` in `packages/common/src/authorization/scopes.ts`:

```ts
{
    projectUuid: context.projectUuid,
    createdByUserUuid: context.userUuid || false,
    type: ProjectType.PREVIEW,
}
```

This would just extend the same pattern to the Dashboard/SavedChart subjects, bypassing the space-access condition when (and only when) the content lives in a preview the user created.

## Verified behaviour today (local, custom roles enabled)

With a custom role containing `view:*` content scopes, `manage:Dashboard@space`, `promote:Dashboard@space`, `create:Project@preview`, `manage:CompileProject`, `create:Job`, `manage:Tags`, `update:Project@self`:

| Action | Result |
|---|---|
| Edit dashboard in production space | 403 ✅ (intended) |
| Create preview with content copy | ✅ |
| Edit the copied dashboard in own preview | **403** ❌ (space access copied from upstream) |
| After manual Editor grant on the preview space | edit works ✅ |
| Promote from preview (same user) | 403 ✅ (needs editor on destination — review gate holds) |
| Promote by a user with promote permissions | ✅ updates the production dashboard in place (slug-matched) |

## Related

- #23805 — org-level scopes have no effect in project-level custom roles (same theme: scope vocabulary vs. assignment level)
- Custom roles docs: "Restrict developers to preview uploads (agentic development workflow)" — this proposal is the UI-editing equivalent of that CLI workflow

jelstongreen:
Closes #24130 (PROD-8234)

## Summary

Adds four custom-role scopes that grant edit rights on content inside preview projects the user created, without granting anything in production:

- `manage:Dashboard@self`
- `manage:SavedChart@self`
- `manage:Space@self`
- `manage:Explore@self`

This unlocks the governed self-serve workflow for non-analyst builders: view-only in production → create a preview with content copy → edit freely in your own preview → a reviewer promotes back. Previously step 3 was impossible because preview creation copies upstream space access verbatim, so a view-only user was view-only inside their own preview.

## Design: build-time context instead of subject enrichment

These reuse the existing `self` scope modifier. "Is this membership's project a preview created by this user" is resolved **at ability-build time** rather than by enriching subjects — adding the project's `type`/`createdByUserUuid` to every Dashboard/SavedChart/Explore CASL subject would mean ~74 backend construction sites plus ~20 frontend ones (and the API types feeding them).

Instead, the membership queries already join `projects`, so they now carry `projectType` + `projectCreatedByUserUuid` into the scope context, and the new scopes emit a rule only when the context says self-created-preview — otherwise no rule at all (`null` return from `getConditions`; `[]` still means unconditional). This works because preview creation copies custom-role assignments to the preview (`copyUserAccessOnPreview` passes `roleUuid`), so the ability builder runs per-preview-membership with the preview's uuid. The rule serializes into `abilityRules`, so **every existing `user.can('manage', …)` check is satisfied without modification** and the frontend works with zero component changes.

Notes:
- `manage` implies `view` in CASL, so the content scopes (`Dashboard`/`SavedChart`/`Space`) are gated by the same public-or-member space conditions as their `view:*` counterparts — otherwise `manage` would re-expose copied private-space content the user couldn't see in production (and `manage:Space` would let them edit a private space's access list to add themselves). `manage:Space@self` adds a condition for creating a brand-new space (metadata-only ability subject, no access to gate on); creating a dashboard/chart already carries the destination space's access context. `manage:Explore@self` stays project-wide — Explore is a project-level data resource with no space access to gate on.
- The promotion review gate is untouched — promoting from the preview still requires promote/editor rights on the destination.
- The four scopes are wired into the DEVELOPER tier in `roleToScopeMapping.ts` (redundant for system developers, same pattern as `manage:ContentAsCode@self`) so the vocabulary-coverage parity test passes and cloned roles surface them.
- Org-level role assignments get no rule from these scopes (no project context) — consistent with #23805.

## Follow-up (pre-existing gap, not addressed here)

`copyUserAccessOnPreview` copies user-level `roleUuid` to the preview but the group-access copy drops it — group-assigned custom roles don't carry into previews.

## Test plan

- [x] Unit: space-aware grant/deny matrix in `scopeAbilityBuilder.test.ts` — in the own preview: public/shared ✓, private space the user is a member of ✓, private space they are **not** a member of ✗ (the leak being closed), brand-new space creation ✓ (Space); plus other project ✗, non-preview ✗, other creator ✗, missing context ✗, org-level ✗ — and end-to-end builder test in `getUserAbilityBuilder.test.ts`. Full `authorization` suite green (513 passed, 1 skipped); `common` typecheck + lint green.
- [x] Live (local EE): builder custom role (view-only + the four `@self` scopes + preview-create checklist) on a seeded viewer. In their own preview, a copied private space they're not a member of is hidden (space list excludes it; direct GET `403`), while a public/shared space stays editable (`200`); in production the private space is hidden. Flipping the grant back to unconditional reproduced the leak (private space listed; direct GET `200`), confirming the gate is what closes it.

### DIFF
--- a/packages/common/src/authorization/scopes.ts
+++ b/packages/common/src/authorization/scopes.ts
@@ -40,16 +40,49 @@ const addAccessCondition = (context: ScopeContext, role?: SpaceMemberRole) => ({
 const addDefaultUuidCondition = flow(addUuidCondition, Array.of);
 
 /**
- * Project-wide grant, but only when the role assignment's project is a
- * preview created by the current user. Returns null (no rule) otherwise —
- * including for org-level assignments, where there is no project context.
+ * True only inside a preview the current user created. Shared by the @self
+ * preview scopes; returns false for org-level assignments (no project context).
  */
-const addSelfPreviewCondition = (context: ScopeContext) =>
-    context.projectUuid &&
-    context.projectType === ProjectType.PREVIEW &&
-    context.userUuid &&
-    context.projectCreatedByUserUuid === context.userUuid
-        ? [{ projectUuid: context.projectUuid }]
+const isSelfPreview = (context: ScopeContext) =>
+    Boolean(
+        context.projectUuid &&
+        context.projectType === ProjectType.PREVIEW &&
+        context.userUuid &&
+        context.projectCreatedByUserUuid === context.userUuid,
+    );
+
+/**
+ * Project-wide grant inside the user's own preview. For subjects with no space
+ * access to gate on (Explore). Returns null (no rule) outside the own preview.
+ */
+const selfPreviewProjectCondition = (context: ScopeContext) =>
+    isSelfPreview(context) ? [{ projectUuid: context.projectUuid }] : null;
+
+/**
+ * Space-gated grant inside the user's own preview. Mirrors the view:* gate
+ * (public/shared OR member-of) so that manage — which implies view in CASL —
+ * can't reach copied private spaces the user isn't a member of. `allowCreate`
+ * adds the new-space case, whose ability subject is metadata-only and so
+ * carries no access field (Dashboard/SavedChart creation carries the
+ * destination space's access, so they don't need it). Null outside own preview.
+ */
+const selfPreviewSpaceCondition = (
+    context: ScopeContext,
+    allowCreate = false,
+) =>
+    isSelfPreview(context)
+        ? [
+              addUuidCondition(context, { inheritsFromOrgOrProject: true }),
+              addAccessCondition(context),
+              ...(allowCreate
+                  ? [
+                        {
+                            ...addUuidCondition(context),
+                            access: { $exists: false },
+                        },
+                    ]
+                  : []),
+          ]
         : null;
 
 const scopes: Scope[] = [
@@ -87,7 +120,7 @@ const scopes: Scope[] = [
             'Create, edit, and delete dashboards in preview projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.CONTENT,
-        getConditions: addSelfPreviewCondition,
+        getConditions: selfPreviewSpaceCondition,
     },
     {
         name: 'view:SavedChart',
@@ -123,7 +156,7 @@ const scopes: Scope[] = [
             'Create, edit, and delete saved charts in preview projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.CONTENT,
-        getConditions: addSelfPreviewCondition,
+        getConditions: selfPreviewSpaceCondition,
     },
     {
         name: 'view:Space',
@@ -173,7 +206,7 @@ const scopes: Scope[] = [
             'Create, edit, and delete spaces in preview projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.CONTENT,
-        getConditions: addSelfPreviewCondition,
+        getConditions: (context) => selfPreviewSpaceCondition(context, true),
     },
     {
         name: 'view:DashboardComments',
@@ -613,7 +646,7 @@ const scopes: Scope[] = [
             'Explore and query data in preview projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.DATA,
-        getConditions: addSelfPreviewCondition,
+        getConditions: selfPreviewProjectCondition,
     },
     {
         name: 'manage:SqlRunner',
--- a/packages/common/src/authorization/scopes.ts
+++ b/packages/common/src/authorization/scopes.ts
@@ -608,6 +608,14 @@ const scopes: Scope[] = [
         getConditions: addDefaultUuidCondition,
     },
     {
+        name: 'manage:Explore@selfPreview',
+        description:
+            'Explore and query data in preview projects created by the user',
+        isEnterprise: false,
+        group: ScopeGroup.DATA,
+        getConditions: addSelfPreviewCondition,
+    },
+    {
         name: 'manage:SqlRunner',
         description:
             'Run SQL queries, execute SQL charts, and browse warehouse schema',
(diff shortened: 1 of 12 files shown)


jelstongreen:
Two follow-up fixes from manual UI testing of the preview-builder persona:

**1. `manage:Explore@selfPreview` (new, fourth scope).** "Add tile → New chart" routes to the Explorer, which renders `ForbiddenPanel` without `manage:Explore` on the project — so a view-only builder couldn't create (or query-edit) charts even inside their own preview. Same build-time self-preview condition as the other three scopes: explore/query in your own preview, production Explorer stays forbidden.

**2. Promote actions hidden without destination access.** The promote menu items (DashboardHeader, SavedChartsHeader, ResourceActionMenu) only checked `can('promote', …source…)`, which any `manage` rule satisfies via the CASL wildcard — so preview builders saw a Promote button that always 403'd. The UI now mirrors the backend's second check: when an upstream project exists, the action also requires promote ability on it. Builders no longer see Promote; developers/admins (and custom reviewer roles holding broad `promote:*`) still do. The "enable an upstream project" disabled-tooltip UX is unchanged. Known approximation: a custom role holding *only* `promote:X@space` (space-conditioned, no broad promote) will have the button hidden even where the backend might allow — the UI can't see upstream space access at render time.

Verified live (local EE harness): demo3 (builder role) gets working tables/explorer in own preview, forbidden in production; Promote absent from demo3's menus, present for admin in the same preview.

packages/common/src/authorization/scopes.ts
```diff
@@ -39,6 +39,70 @@ const addAccessCondition = (context: ScopeContext, role?: SpaceMemberRole) => ({
...
+ * Project-wide self-preview grant. Returns null (no rule) outside the user's
+ * own preview — including for org-level assignments, where there is no project
+ * context.
+ *
+ * Only for subjects that carry no space access to gate on (e.g. Explore, a
+ * project-level data resource rather than space content). Space content uses
+ * addSelfPreviewContentCondition / addSelfPreviewSpaceCondition so that manage
+ * — which implies view in CASL — cannot re-expose copied private-space content.
+ */
+const addSelfPreviewCondition = (context: ScopeContext) =>
```
rephus: all these helpers are very verbose and redundant, we already do this preview check in many places, I don't think we need this . 

All you need is to add this to the scope 

> context.projectType === ProjectType.PREVIEW 
> context.projectCreatedByUserUuid === context.userUuid

packages/common/src/types/scopes.ts
```diff
@@ -35,19 +36,31 @@ type BaseScopeContext = {
...
 export type ScopeContext = OrganizationScopeContext | ProjectScopeContext;
 
-export type ScopeModifer = 'self' | 'public' | 'assigned' | 'space' | 'preview';
+export type ScopeModifer =
+    | 'self'
+    | 'public'
+    | 'assigned'
+    | 'space'
+    | 'preview'
+    | 'selfPreview';
```
rephus: we added a new scopeModifier , but we can reuse the `self`

jelstongreen: Done — dropped the `selfPreview` modifier and reused `self`. The four scopes are now `manage:{Dashboard,SavedChart,Space,Explore}@self`. No migration needed: the old names were introduced in this PR, so nothing in `scoped_roles` references them.

packages/common/src/authorization/scopes.ts
jelstongreen: Collapsed the helpers to one shared guard (`isSelfPreview`) plus two condition builders.

I kept the space-gate rather than emitting a bare `{ projectUuid }`, though — that's the one place I didn't take the literal suggestion. `manage` implies `view` in CASL, so the unconditional grant re-exposes copied private-space content the builder couldn't see in prod, and `manage:Space` would even let them edit a private space's access list to self-add. Reproduced locally: with the bare guard, a view-only builder lists and reads (GET 200) a private space they're not a member of inside their own preview; with the gate, same space is hidden (GET 403) while public/shared content stays editable (200). So the extra conditions are load-bearing.

On "we already do this preview check in many places" — the bare `type === PREVIEW` is common, but I couldn't find an existing helper that combines it with `createdBy === me`; `isSelfPreview` centralises that for the four scopes. Happy to move/rename it if there's a preferred home.

PR #24170, opened 2026-06-11
fix: calculate data app full height for screenshots

magnew:
### Description:

 - Scheduled-delivery screenshots of data apps were cropping off the
    bottom of the app — only the top region rendered, leaving content
    past the iframe's initial 100vh viewport invisible (charts in lower
    rows, footer, background art).
  - Root cause: `UnfurlService` stretched the parent's iframe element to
    the inner document's `scrollHeight`. That value is unreliable for
    apps whose container has `overflow: hidden` (or equivalent) — content
    past 100vh doesn't contribute, so the iframe never stretched and the
    screenshot captured only what fit in the initial viewport. The
    inverse failure (over-report → white slab at the bottom) is also
    possible when `min-height: 100vh` is set on body/html.
  - Now walks leaf elements inside the iframe (no children, positive
    dimensions) and takes `max(getBoundingClientRect().bottom)`.
    `getBoundingClientRect` isn't clipped by ancestor overflow, so the
    natural position of every rendered leaf is visible to the
    measurement. Falls back to `scrollHeight` only if no leaves are
    found (empty body — shouldn't happen in practice).
  - Logs the measured content height at info level so future regressions
    can be diagnosed from delivery logs.
  - Stacks on top of the SDK-readiness gating fix.

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -1501,13 +1501,63 @@ export class UnfurlService extends BaseService {
                                 (f) => f !== mainFrame,
                             );
                             if (appFrame) {
+                                // `scrollHeight` is unreliable: it
+                                // under-reports when ancestor overflow
+                                // hides content, and over-reports when
+                                // `min-height: 100vh` is set. Instead,
+                                // walk leaf elements and use
+                                // getBoundingClientRect, which reflects
+                                // actual rendered position regardless
+                                // of overflow.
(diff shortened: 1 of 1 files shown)


magnew:
* **#24170** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24170?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24170?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24178, opened 2026-06-11
fix: omnibar "Created by" filter loses selection when searching

### LINKED ISSUE #24174: Omnibar "Created by" filter: selection is lost when picking a user after typing in the search
### Description

In the global search (omnibar), the "Created by" filter doesn't apply when you use the search input inside the dropdown before clicking a user. The dropdown closes and the filter shows "Created by" again as if nothing was selected.

Additionally, if a user is already selected, re-selecting the same user silently clears the filter.

Expected: clicking a user in the dropdown (with or without typing first) applies the "Created by" filter, and re-selecting the current user keeps it.

1. Open the omnibar (Cmd+K) in a project
2. Open the "Created by" filter dropdown
3. Type part of a user's name (e.g. "edi") in the search input
4. Click the matching user in the dropdown
5. The menu closes but the filter still shows "Created by" — no filter applied

For the second issue:

1. Apply a "Created by" filter by clicking a user from the full list (works)
2. Reopen the dropdown and click the same user again
3. The filter is cleared

### version

v0.3135.0

### Cloud or self-hosting

cloud

tatianainama:
## Description

Fixes the omnibar "Created by" filter losing the selection when a user is picked after typing in the dropdown's search input, and the filter being silently cleared when re-selecting the currently selected user.

Two underlying problems in `OmnibarFilters.tsx`:

1. **Lost click after typing**: the Select's options dropdown rendered in a portal, so pressing the mouse on an option registered as an *outside click* for the wrapping `Menu`, which started closing and re-rendered the component. The `data` array was rebuilt inline on every render and `value` was `undefined` when no filter was set, so Mantine treated the Select as uncontrolled and reset its search text on the `data` identity change — between `mousedown` and `mouseup`. The filtered list snapped back to the full list, the option under the cursor was replaced, and the click never landed.
2. **Same-user re-select cleared the filter**: Mantine v8 `Select` defaults `allowDeselect` to `true`, so clicking the already-selected option fires `onChange(null)`.

## Changes

- Pass `value={filters?.createdByUuid ?? null}` so the Select stays controlled (no search reset on `data` changes)
- Memoize the user options so `data` keeps a stable identity
- Render the Select's combobox with `withinPortal: false` so clicking an option isn't an outside click that closes the Menu mid-selection
- Set `allowDeselect={false}` (clearing is still available via the clear button)
- Close the menu in `onOptionSubmit` instead of `onChange`, since `onChange` doesn't fire when re-selecting the current value

## How tested

Verified in the browser with real clicks: selecting a user after searching applies the filter, re-selecting the same user keeps it, switching users via search works, selecting from the unfiltered list works, and the dropdown renders without clipping inside the menu.

Closes: PROD-8256
Closes: #24174

### DIFF
--- a/packages/frontend/src/features/omnibar/components/OmnibarFilters.tsx
+++ b/packages/frontend/src/features/omnibar/components/OmnibarFilters.tsx
@@ -101,6 +101,15 @@ const OmnibarFilters: FC<Props> = ({ filters, onSearchFilterChange }) => {
         );
     }, [filters]);
 
+    const userOptions = useMemo(
+        () =>
+            organizationUsers?.map((user) => ({
+                value: user.userUuid,
+                label: `${user.firstName} ${user.lastName}`,
+            })) || [],
+        [organizationUsers],
+    );
+
     return (
         <Group px="md" py="sm">
             <Menu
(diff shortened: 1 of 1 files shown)


tatianainama:
* **#24178** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24178?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24178?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24181, opened 2026-06-11
feat(common): add sessionTimezone slot to resolveQueryTimezone

almeidabbm:
Part 1 of 3 of the embed per-session timezone stack.

Adds a new highest-priority `sessionTimezone` slot to the shared `resolveQueryTimezone` resolver so a per-session override can outrank the chart pin, user timezone, and project default. This is a pure foundation: every existing call site passes `sessionTimezone: null`, so there is no behaviour change.

Precedence is now `sessionTimezone ?? metricQuery.timezone (chart pin) ?? user timezone ?? project timezone`. Invalid timezones still throw `ParameterError` via the existing `isValidTimezone` check.

Unit tests cover the new precedence (session wins over pin, user, and project) and the invalid-throw.

Relates: GLITCH-488

PR #24182, opened 2026-06-11
feat(embed): thread session timezone into query resolution

almeidabbm:
Part 2 of 3 of the embed per-session timezone stack. Stacked on #24181.

Threads the embed session timezone from the request body into query timezone resolution:
* `resolveTimezoneContext` → `prepareMetricQueryAsyncQueryArgs` → `executeAsyncDashboardChartQuery` (dashboard tile render path)
* `EmbedService.searchFilterValues` (filter-value autocomplete direct resolver call)
* adds an optional `timezone` field to the `/query/dashboard-tile` and `/filter/:filterUuid/search` embed request bodies

Totals and subtotals need no change. They re-query from the persisted `metricQuery.timezone` on the originating tile query, so they inherit the session timezone automatically.

The session timezone is gated at the `EmbedService` boundary on `EnableTimezoneSupport`: when the flag is off the param is dropped to null before resolution, so it is fully inert (not resolved, not validated). The gate lives in `EmbedService`, the single source of the URL param, rather than in the shared resolver, so the non-embed call sites stay untouched. When the flag is on, the session timezone overrides the chart pin and an invalid value returns a 400.

API behaviour test (`packages/api-tests`) covers: the tile resolves to the session timezone, totals inherit it without re-passing, and an invalid timezone returns 400 when timezone support is on (and is inert when off).

Relates: GLITCH-488

test-all

### DIFF
--- a/packages/api-tests/tests/embedTimezone.test.ts
+++ b/packages/api-tests/tests/embedTimezone.test.ts
@@ -214,13 +214,30 @@ describe('Embed dashboard-tile session timezone (?timezone=)', () => {
         expect(totalsResp.body.results.resolvedTimezone).toBe(resolvedTimezone);
     });
 
-    it('rejects an invalid session timezone with a 400', async () => {
+    it('rejects an invalid session timezone with a 400 when timezone support is on', async () => {
         const token = await freshDashboardJwt();
+
+        // Probe the flag: resolvedTimezone is null when EnableTimezoneSupport
+        // is off, in which case the session timezone is gated out before
+        // resolution and never validated.
+        const probe = await executeTile(token, SESSION_TIMEZONE);
+        expect(probe.status).toBe(200);
+        const timezoneSupportEnabled =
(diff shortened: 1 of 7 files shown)


almeidabbm:
* **#24183** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24183?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24182** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24182?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24182?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24181** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24181?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #24188, opened 2026-06-11
fix(ai): make promptless AI agent threads fetchable

joaoviana:
## Problem

Clicking **Test fix** on an Ask AI review finding navigated to the remediation's preview thread but the page failed with `AI agent thread not found for uuid: <agent>/<thread>` — the pre-created preview work thread has no prompts yet, and the thread read path could not return promptless threads.

Two layered causes in `buildThreadSummaryQuery`:
1. An **inner join** on `ai_prompt` (with the first-prompt correlation in the `WHERE` clause) filtered out any thread with zero prompts.
2. The thread's `user` was derived from the first prompt's creator, so even with the join fixed, an empty thread would have no owner and the UI would mark it read-only for everyone.

## Fix

- Resolve the first prompt via `LEFT JOIN LATERAL ... LIMIT 1` so promptless threads are returned.
- Resolve the thread user via `COALESCE(first_prompt.created_by_user_uuid, ai_web_app_thread.user_uuid)` — empty web-app threads fall back to their web-app owner.
- `AiAgentThreadSummary.firstMessage` is now `{...} | null` (it was always nullable in practice: `POST /threads` accepts an empty body and previously 500'd on its own trailing `getThread`). Regenerated OpenAPI and updated the five frontend consumers with safe fallbacks ("New thread").
- **Migration**: adds `ai_prompt(ai_thread_uuid, created_at)` index — the FK column had no index at all, so both the old correlated `MIN(created_at)` subquery and the new lateral join were scanning `ai_prompt` per thread on every thread-list query.

## Who is affected

- Empty threads now appear in thread lists (previously silently hidden) and are fetchable. Threads with prompts behave exactly as before — same first-prompt selection, same owner resolution (prompt creator wins over web-app owner).
- API consumers of thread summaries see `firstMessage: null` for promptless threads — previously those threads were unreachable (404), so no consumer could have depended on the old shape.

## Testing

- Verified live against local dev: `POST /threads` with empty body returns the summary with `firstMessage: null`; `GET /threads/:uuid` (the failing call) returns 200; lists include empty threads; prompt-ful threads unchanged.
- Backend + frontend aiCopilot suites pass; migration applied and rolled back locally.

Relates: (Linear ticket TBD)

### DIFF
--- a/packages/backend/src/ee/database/migrations/20260611150000_add_ai_prompt_thread_created_at_index.ts
+++ b/packages/backend/src/ee/database/migrations/20260611150000_add_ai_prompt_thread_created_at_index.ts
@@ -1,0 +1,16 @@
+import { Knex } from 'knex';
+
+const AI_PROMPT_TABLE_NAME = 'ai_prompt';
+const INDEX_NAME = 'ai_prompt_thread_uuid_created_at_index';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(AI_PROMPT_TABLE_NAME, (table) => {
+        table.index(['ai_thread_uuid', 'created_at'], INDEX_NAME);
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(AI_PROMPT_TABLE_NAME, (table) => {
+        table.dropIndex(['ai_thread_uuid', 'created_at'], INDEX_NAME);
+    });
+}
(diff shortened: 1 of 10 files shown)


joaoviana:
* **#24188** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24188?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joaoviana:
Closing — rethought this. The only producer of promptless threads was the review-remediation seeding flow, which now creates thread + first prompt atomically in one transaction (#24190), so a promptless (unfetchable) thread can no longer exist. That removes the need to weaken `firstMessage` to nullable in the public API, the lateral-join rewrite of the thread summary query, and the new index migration.

PR #19759, opened 2026-01-28
fix: ignore timezone when formatting date values with expressions

notgiorgi:
Closes: [PROD-2727: Using date format shifts my x-axis labels](https://linear.app/lightdash/issue/PROD-2727/using-date-format-shifts-my-x-axis-labels)

Closes: [PROD-2726: Chart tooltips convert timestamps differently than x-axis](https://linear.app/lightdash/issue/PROD-2726/chart-tooltips-convert-timestamps-differently-than-x-axis)

### Description:

Add `ignoreTimezone: true` option to `formatWithExpression` when formatting date values to ensure consistent date formatting regardless of timezone settings.

This code explains the issue and the fix:

```ts
import { format } from "numfmt";
import dayjs from "dayjs";
import utc from "dayjs/plugin/utc";
dayjs.extend(utc);

const date = new Date("2024-11-01T00:00:00Z");

console.log({
  dateISO: date.toISOString(),
  dateLocal: `${date.toLocaleDateString()} ${date.toLocaleTimeString()}`,
  formattedISO: dayjs.utc(date).format("D MMM YYYY HH:mm:ss"),
  formattedLocal: dayjs(date).format("D MMM YYYY HH:mm:ss"),
  numfmt_current: format("d mmm yyyy HH:mm:ss", date), // current impl
  // https://github.com/borgar/numfmt/issues/6
  numfmt_fixed: format("d mmm yyyy HH:mm:ss", date, { ignoreTimezone: true }),
});

// {
//   dateISO: "2024-11-01T00:00:00.000Z",
//   dateLocal: "10/31/2024 7:00:00 PM",
//   formattedISO: "1 Nov 2024 00:00:00",
//   formattedLocal: "31 Oct 2024 19:00:00",
//   numfmt_current: "31 Oct 2024 19:00:00",
//   numfmt_fixed: "1 Nov 2024 00:00:00",
// }
```

PR #24200, opened 2026-06-11
fix(formatting): apply project tz to custom-format timestamps

### LINKED ISSUE #24160: Custom value formatting on a timestamp reverts the displayed time to UTC instead of the project timezone
### Describe the bug

When a dashboard tile (or chart) displays a single timestamp value, the value is stored in UTC and — with the project timezone set to a non-UTC zone such as `America/New_York` — it is correctly converted and shown in that timezone.

However, as soon as **any custom formatting is applied to the value**, the displayed timestamp reverts to UTC, even though the chart still indicates the time should be in the project timezone (`America/New_York`). The timezone conversion is silently dropped.

### Steps to reproduce

1. Set the project timezone to a non-UTC zone (e.g. `America/New_York`).
2. Build a chart/tile that surfaces a single timestamp dimension or metric whose underlying value is stored in UTC.
3. Confirm the timestamp displays correctly converted to the project timezone.
4. Add a custom format / format expression to that field (e.g. a date format string like `yyyy-mm-dd hh:mm:ss`).
5. Observe that the displayed value now shows the UTC time, not the project-timezone time — despite the chart still indicating `America/New_York`.

### Expected behavior

Applying custom formatting should change *how* the timestamp is rendered, not *which timezone* it is rendered in. The value should remain converted to the project timezone after formatting is applied.

### Actual behavior

Adding a custom format to a timestamp field bypasses timezone conversion entirely and the value renders in UTC.

almeidabbm:
### Problem

A timestamp field rendered with a **custom format expression** (e.g. `yyyy-mm-dd hh:mm:ss`) reverted the displayed time to **UTC**, even when the project timezone was non-UTC. A plain timestamp in the same table/tile rendered correctly in the project timezone, so the two diverged.

### Root cause

`formatValueWithExpression` had no timezone awareness. Its date branch fed the raw UTC instant to numfmt with `{ ignoreTimezone: true }` (added in #19759 to stop the *incidental runtime-local* zone from drifting the output). That made expression-formatted timestamps render their UTC wall-clock unconditionally, dropping the project-timezone shift that the default `formatTimestamp` path applies.

### Fix

- Thread the resolved timezone through `formatValueWithExpression`. In the date branch, shift the UTC instant into the project tz (`moment.utc(value).tz(timezone).utc(true).toDate()`) before handing it to numfmt, keeping `ignoreTimezone: true`. This refines #19759 rather than reverting it — we still ignore the incidental runtime-local zone, but now honor the *configured* one.
- Gate the shift at the call sites with the existing `shouldShiftItemTimezone` helper, so calendar DATEs and `skipTimezoneConversion` fields stay put.

### Gating

Inherits `EnableTimezoneSupport` — the `timezone` reaching the formatter is the API's flag-gated `displayTimezone` (null when the flag is off), so flag-off and explicit-UTC output is bit-identical. Chart (cartesian) axis/tooltip formatting is unaffected: it formats via a separate `seriesValueFormatter` that passes no timezone, and its timezone handling lives in `applyTimezoneShiftToEchartsOptions`.

### Tests

Two TDD cases in `formatting.test.ts`: project-tz shift for a custom-format timestamp, and no-shift when the flag is off / timezone is UTC.

**Before**

https://github.com/user-attachments/assets/76fe4a83-ca98-4cd0-8038-b6b9cdcddd56

**After**

https://github.com/user-attachments/assets/9747e7e4-22d5-4588-8ae4-ce2193416ce8

Closes: GLITCH-492
Closes: #24160
Relates: #19759

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -1190,56 +1190,6 @@ describe('Formatting', () => {
             ).toEqual('2024-01-14 20:00:00');
         });
 
-        test('formatItemValue shifts a custom-format MAX timestamp metric into the project tz from a Date OR an ISO string (GLITCH-492)', () => {
-            const maxMetric = {
-                ...metric,
-                type: MetricType.MAX,
-                format: 'yyyy-mm-dd hh:mm:ss',
-            };
-            const isoString = '2024-01-14T20:00:00.000Z';
-            const asDate = new Date(isoString); // 15:00 in NY (EST)
-
-            const fromDate = formatItemValue(
-                maxMetric,
-                asDate,
-                false,
-                undefined,
-                'America/New_York',
-            );
-            const fromString = formatItemValue(
-                maxMetric,
-                isoString,
-                false,
-                undefined,
-                'America/New_York',
-            );
-
-            expect(fromDate).toEqual('2024-01-14 15:00:00');
-            expect(fromString).toEqual(fromDate);
-        });
-
-        test('formatItemValue shifts a CUSTOM-format MAX timestamp metric into the project tz (GLITCH-492)', () => {
-            const maxMetric = {
-                ...metric,
-                type: MetricType.MAX,
-                formatOptions: {
-                    type: CustomFormatType.CUSTOM,
-                    custom: 'yyyy-mm-dd hh:mm:ss',
-                },
-            };
-            const value = new Date('2024-01-14T20:00:00.000Z'); // 15:00 in NY (EST)
-
-            expect(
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#24200** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24200?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24222, opened 2026-06-12
fix: detect Windows when $OSTYPE is cygwin in install.sh

### LINKED ISSUE #24211: scripts/install.sh fails on Windows Git Bash/MSYS2 — is_windows() only matches msys*, not cygwin*
## Description

Running `./scripts/install.sh` on Windows under Git Bash / MSYS2 fails because the OS-detection logic doesn't recognise the shell as Windows.

`is_windows()` only matches `OSTYPE == msys*`, but depending on the Git Bash / MSYS2 build, `$OSTYPE` can be `cygwin` (or `win32`).

When detection fails, `check_os()` falls through to the Linux branch (`cat /etc/*-release`), sets the OS to `Not Found`, and `start_docker()` then runs `sudo systemctl start docker.service` — neither `sudo` nor `systemctl` exists on Windows. The script aborts with a misleading **"The containers didn't seem to start correctly"** message, even though Docker Desktop is installed and running.

## Environment

| | |
|---|---|
| **OS** | Windows 11 |
| **Shell** | Git Bash / MSYS2 — `uname -a` reports `MINGW64_NT-10.0-26200 ... Msys`, but `echo $OSTYPE` returns `cygwin` |
| **Docker** | Docker Desktop 29.5.3 (running, `desktop-linux` context) |
| **Docker Compose** | v5.1.4 |
| **Lightdash** | 0.3143.1 (fast install / `lightdash/lightdash:latest`) |

## Steps to reproduce

1. On Windows, open Git Bash where `echo $OSTYPE` returns `cygwin`.
2. Clone the repo and run `./scripts/install.sh`.
3. Choose the fast install.

## Actual behaviour

```text
Detecting your OS ...
cat: '/etc/*-release': No such file or directory

✅ Detected OS: Not Found:
...
Starting Docker ...
Sudo is disabled on this machine. ...
Starting docker service
Sudo is disabled on this machine. ...
🔴 The containers didn't seem to start correctly.
```

## Expected behaviour

The script detects Windows, prints the "Make sure Docker Desktop is running" guidance, and skips the Linux `systemctl` path.

## Root cause

`scripts/install.sh`:

```bash
is_windows() {
    [[ $OSTYPE == msys* ]]
}
```

`$OSTYPE` is not guaranteed to be `msys` on Windows Git Bash; it can be `cygwin` or `win32`.

## Suggested fix

```bash
is_windows() {
    [[ $OSTYPE == msys* || $OSTYPE == cygwin* || $OSTYPE == win32 ]]
}
```

With this change the script detects Windows correctly and the fast install completes.

## Related rough edges noticed

While debugging the above, two other independent (platform-agnostic) issues in `scripts/install.sh` made this harder than it needed to be:

1. **The real `docker compose` error is swallowed, producing a misleading message.** The `up` command is wrapped in `|| true`, so when a container genuinely fails to start, the script continues and only surfaces the generic *"The containers didn't seem to start correctly"* via the health-check timeout — the actual `docker compose` stderr is never shown. The inline comment acknowledges the `|| true` is a workaround for a `--detach` quirk, but it also hides legitimate failures. Surfacing stderr (or checking the exit code separately) would make install failures self-diagnosing.

2. **The health check ignores the user-chosen port.** `set_vars` lets the user pick a port and the final success message prints `http://localhost:$port`, but `wait_for_containers_start` polls `http://localhost:8080` unconditionally (and the fast-install path never passes `PORT` to `docker compose` at all). A custom port therefore yields a false "containers didn't start" failure.

Minor: `check_ports_occupied` is commented out (line ~378), so the advertised port-availability pre-check never actually runs.

1. On Windows, open Git Bash where `echo $OSTYPE` returns `cygwin`.
2. Clone the repo and run `./scripts/install.sh`.
3. Choose the fast install.

### version

0.3143.1

### Cloud or self-hosting

self-hosting

back1ply:
Closes: #24211

### Description

`scripts/install.sh` fails on Windows under Git Bash / MSYS2 / Cygwin because `is_windows()` only matched `$OSTYPE == msys*`. Depending on the build, `$OSTYPE` can be `cygwin` instead. When detection fails, `check_os()` falls through to the Linux branch (`cat /etc/*-release` → OS "Not Found"), and `start_docker()` then runs `sudo systemctl start docker.service` — neither `sudo` nor `systemctl` exists on Windows. The script aborts with a misleading **"The containers didn't seem to start correctly"** even though Docker Desktop is installed and running.

```bash
is_windows() {
    [[ $OSTYPE == msys* || $OSTYPE == cygwin* ]]
}
```

### Testing

Verified end-to-end on Windows 11 (Git Bash, `$OSTYPE=cygwin`, Docker Desktop 29.5.3):

- Before: detection failed → OS "Not Found" → `sudo systemctl` path → misleading failure.
- After: `✅ Detected OS: Windows`, fast install completes, all containers `Up`, `/api/v1/health` returns `200`.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -52,7 +52,7 @@ is_mac() {
 }
 
 is_windows() {
-    [[ $OSTYPE == msys* ]]
+    [[ $OSTYPE == msys* || $OSTYPE == cygwin* ]]
 }
 
 check_os() {


notgiorgi (changes requested):
Please scope PR to a single fix

back1ply:
Thanks for the review! Scoped down to just the `is_windows()` cygwin detection fix. The other rough edges I'd noted (port-aware health check, surfacing the real `docker compose` error, teardown hint) are out of scope here — happy to open separate PRs for those if they're wanted.

notgiorgi (approved):

notgiorgi:
Hey @back1ply, Commits must have verified signatures

notgiorgi:
@all-contributors add @back1ply  for code

PR #24225, opened 2026-06-12
fix: support upstream self preview permissions

### LINKED ISSUE #24224: Service accounts with project-scoped @self preview permissions cannot deploy or upload to their own previews
## Problem

Customers want to run local development and AI/CI workflows with least-privilege service accounts. A common desired setup is:

* Service account has scoped access to a non-production source project, e.g. `DBT_DEV`
* Service account can create preview projects from that source project
* Service account can deploy/upload content-as-code only to previews it created from that source project
* Service account cannot deploy/upload to production projects

Today this does not work reliably for the `@self` preview scopes. A service account can be granted `manage:DeployProject@self` or `manage:ContentAsCode@self` on the source project, but the actual deploy/upload action targets the generated preview project. The permission rule is scoped to the source project UUID, while the permission check sees the preview project UUID, so the request is denied.

## Example

1. Admin creates a service account for local dev/agent workflows.
2. Admin grants it a project-scoped custom role on `DBT_DEV` with:
   * `create:Project@preview`
   * `manage:DeployProject@self`
   * `manage:ContentAsCode@self`
3. Service account creates preview `preview-123` from `DBT_DEV`.
4. Service account tries to run `lightdash deploy --project preview-123` or upload content-as-code to `preview-123`.
5. Expected: allowed, because `preview-123` was created by that service account from `DBT_DEV`.
6. Actual: denied, because the permission check matches against `preview-123` while the grant is scoped to `DBT_DEV`.

## Desired behavior

`@self` preview permissions should understand the relationship between a preview project and its upstream/source project.

A service account or user with an upstream project grant should be allowed to use `@self` preview permissions when all are true:

* The target project is a preview project
* The target preview was created by the same principal
* The target preview was created from the project where the grant applies

They should still be denied when:

* The target is a production/default project
* The preview was created by another user/service account
* The preview was created from a different upstream project

## Acceptance criteria

* A project-scoped grant on `DBT_DEV` with `manage:DeployProject@self` allows deploy only to previews created by the same principal from `DBT_DEV`.
* A project-scoped grant on `DBT_DEV` with `manage:ContentAsCode@self` allows CAC upload only to previews created by the same principal from `DBT_DEV`.
* The same grant does not allow deploy/CAC upload to production projects.
* The same grant does not allow deploy/CAC upload to previews created by another principal.
* The same grant does not allow deploy/CAC upload to previews created from a different upstream project.
* Behavior works for service accounts as well as human users.

notgiorgi:
### Description

Fixes `@self` preview permissions so project-scoped grants on an upstream/source project can apply to previews created from that project by the same principal.

This lets a user or service account with `manage:DeployProject@self`, `manage:ContentAsCode@self`, `update:Project@self`, or `delete:Project@self` on a source project act on their own previews from that source project, without granting access to production projects, other users' previews, or previews from another source project.

### Testing

- `pnpm -F backend test -- DeployService.test.ts`
- `pnpm -F common test -- scopeAbilityBuilder.test.ts`
- `pnpm -F common test -- roleToScopeParity.test.ts`
- `pnpm -F common test -- projectMemberAbility.test.ts`
- `pnpm -F common typecheck:fast`
- `git diff --check`

`pnpm -F backend typecheck:fast` currently fails on existing common/backend type drift unrelated to this change.

Closes: PROD-8281
Closes: #24224

### DIFF
--- a/packages/common/src/authorization/scopes.ts
+++ b/packages/common/src/authorization/scopes.ts
@@ -51,6 +51,36 @@ const isSelfPreview = (context: ScopeContext) =>
         context.projectCreatedByUserUuid === context.userUuid,
     );
 
+const ownPreviewProjectConditions = (context: ScopeContext) => {
+    if (context.organizationUuid) {
+        return [
+            {
+                // Org assignments can reach any preview created by this principal.
+                organizationUuid: context.organizationUuid,
+                createdByUserUuid: context.userUuid || false,
+                type: ProjectType.PREVIEW,
+            },
+        ];
+    }
+
+    if (!context.projectUuid || !context.userUuid) return null;
+
+    return [
+        {
+            // Direct preview assignment: the grant is on the preview itself.
+            projectUuid: context.projectUuid,
+            createdByUserUuid: context.userUuid,
+            type: ProjectType.PREVIEW,
+        },
+        {
+            // Upstream assignment: the grant is on the source project.
+            upstreamProjectUuid: context.projectUuid,
+            createdByUserUuid: context.userUuid,
+            type: ProjectType.PREVIEW,
+        },
+    ];
+};
+
 /**
  * Project-wide grant inside the user's own preview. For subjects with no space
  * access to gate on (Explore). Returns null (no rule) outside the own preview.
@@ -339,13 +369,7 @@ const scopes: Scope[] = [
         description: 'Update projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.PROJECT_MANAGEMENT,
-        getConditions: (context) => [
-            {
-                projectUuid: context.projectUuid,
-                createdByUserUuid: context.userUuid || false,
-                type: ProjectType.PREVIEW,
-            },
-        ],
+        getConditions: ownPreviewProjectConditions,
     },
     {
         name: 'delete:Project',
@@ -359,13 +383,7 @@ const scopes: Scope[] = [
         description: 'Delete projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.PROJECT_MANAGEMENT,
-        getConditions: (context) => [
-            {
-                projectUuid: context.projectUuid,
-                createdByUserUuid: context.userUuid || false,
-                type: ProjectType.PREVIEW,
-            },
-        ],
+        getConditions: ownPreviewProjectConditions,
     },
     {
         name: 'manage:Project',
@@ -393,13 +411,7 @@ const scopes: Scope[] = [
         description: 'Deploy to preview projects created by the user',
         isEnterprise: false,
         group: ScopeGroup.PROJECT_MANAGEMENT,
-        getConditions: (context) => [
-            {
-                projectUuid: context.projectUuid,
-                createdByUserUuid: context.userUuid || false,
-                type: ProjectType.PREVIEW,
-            },
-        ],
+        getConditions: ownPreviewProjectConditions,
     },
     {
         name: 'manage:Validation',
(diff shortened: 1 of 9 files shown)


notgiorgi:
* **#24225** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24225?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/common/src/authorization/scopes.ts
```diff
@@ -58,6 +58,36 @@ const isSelfPreview = (context: ScopeContext) =>
 const selfPreviewProjectCondition = (context: ScopeContext) =>
     isSelfPreview(context) ? [{ projectUuid: context.projectUuid }] : null;
 
+const ownPreviewProjectConditions = (context: ScopeContext) => {
+    if (context.organizationUuid) {
+        return [
+            {
+                // Org assignments can reach any preview created by this principal.
+                organizationUuid: context.organizationUuid,
+                createdByUserUuid: context.userUuid,
```
sdolidze: The org-level branch drops the `|| false` guard that the project/upstream branches below — and every other `@self` scope — keep on `createdByUserUuid`.

Not currently exploitable: `context.userUuid` is always a non-empty string in the runtime paths, and CASL won't match a `null`-creator subject (orphaned preview) against an `undefined` condition. But it removes a defense-in-depth guard and is inconsistent with the rest of the file, so it would silently re-arm if any future caller passed an empty/undefined `userUuid`.

Suggest `createdByUserUuid: context.userUuid || false` here too. Also worth an org-level test case — this branch is currently uncovered, and it's the one with the missing guard.

packages/common/src/authorization/scopes.ts
```diff
@@ -393,13 +411,7 @@ const scopes: Scope[] = [
...
         isEnterprise: false,
         group: ScopeGroup.PROJECT_MANAGEMENT,
-        getConditions: (context) => [
-            {
-                projectUuid: context.projectUuid,
-                createdByUserUuid: context.userUuid || false,
-                type: ProjectType.PREVIEW,
-            },
-        ],
+        getConditions: ownPreviewProjectConditions,
```
sdolidze, line 414: These four scopes (`update`/`delete:Project@self`, `manage:DeployProject@self`, and `manage:ContentAsCode@self`) now widen to previews created from a granted upstream project — but `manage:Explore@self` (line 649) still uses the old `selfPreviewProjectCondition` and is **not** widened.

Net effect: a user with a source-project grant can deploy/update/delete their preview but can't explore data in it. Is that asymmetry intentional? If not, `manage:Explore@self` should move to `ownPreviewProjectConditions` too; if so, a one-line comment noting why would help future readers.

sdolidze (commented):
Reviewed for correctness, security, and design — solid fix. The new `ownPreviewProjectConditions` upstream logic checks out: it correctly matches own previews created from a granted source project and rejects production projects, other users' previews, and previews from a different source (all gated on `type: PREVIEW`). The privilege-escalation path is closed since preview creation already requires `view:Project` + `create:Project` on the upstream.

Two author's-call items inline below; neither blocks merge. Two follow-up notes not worth inlining: (1) the direct-preview branch (`projectUuid: context.projectUuid`) lost its test coverage when the ContentAsCode test was switched to the upstream case — worth re-adding; (2) longer term, a shared `projectSubject(project)` builder would stop future `subject('Project'|'DeployProject'|'ContentAsCode', …)` check sites from silently omitting `upstreamProjectUuid` and under-granting, since CASL subjects are untyped.

PR #24226, opened 2026-06-12
feat: create embed dashboard

rephus:
## Summary

Adds embedded dashboard creation/editing through the SDK by reusing existing dashboard, content, SQL chart, and async query endpoints instead of introducing embed-specific API routes.

### What changed

- Exposes a new SDK dashboard builder flow that creates an empty dashboard in the JWT `writeActions.spaceUuid` and lets the host app control edit mode.
- Allows embedded users with write actions to rename dashboards, add saved charts from the allowed space, move/resize tiles, save changes, and then view the dashboard through the normal embedded dashboard renderer.
- Reuses existing dashboard tile components for saved charts, SQL charts, markdown, headings, and dashboard layout behavior where possible.
- Filters add-tile content to the JWT write space so users cannot choose charts outside the allowed space.
- Keeps dashboards created/edited from embed equivalent to normal dashboards: they can be viewed/edited from Lightdash and vice versa.

### JWT write access and security

- Existing endpoints now accept JWT accounts only where the SDK needs dashboard builder behavior.
- JWTs without `writeActions` fail closed for write-capable paths.
- JWT write actions execute as the resolved write actor and are constrained to `writeActions.spaceUuid`.
- Dashboard create/update/duplicate validates target spaces and tile chart ownership server-side, including SQL chart tiles.
- Async dashboard queries preserve existing non-write JWT behavior: a dashboard JWT can still render charts already on its embedded dashboard, while write-action previews may use dashboards from the allowed write space.

### Testing

- `pnpm generate-api`
- `pnpm -F common test -- TableDataModel.test.ts`
- `pnpm -F backend test -- account.test.ts DashboardService.test.ts AsyncQueryService.test.ts`
- `pnpm -F backend test -- AsyncQueryService.test.ts DashboardService.test.ts`
- `pnpm -F backend typecheck`
- `pnpm -F frontend typecheck`
- `pnpm -F backend lint`
- `pnpm -F frontend lint`
- `pnpm sdk-build:fast`

Closes: PROD-569

### DIFF
--- a/packages/backend/src/controllers/projectController.ts
+++ b/packages/backend/src/controllers/projectController.ts
@@ -922,7 +922,6 @@ Migrate to the v2 async query flow: [Execute SQL query](https://docs.lightdash.c
         @Request() req: express.Request,
         @Query() duplicateFrom?: string,
     ): Promise<ApiCreateDashboardResponse> {
-        assertRegisteredAccount(req.account);
         const dashboardService = this.services.getDashboardService();
         this.setStatus(201);
 
@@ -935,8 +934,8 @@ Migrate to the v2 async query flow: [Execute SQL query](https://docs.lightdash.c
                 );
             }
 
-            results = await dashboardService.duplicate(
-                toSessionUser(req.account),
+            results = await dashboardService.duplicateFromAccount(
+                req.account!,
                 projectUuid,
                 duplicateFrom.toString(),
                 body,
@@ -948,8 +947,8 @@ Migrate to the v2 async query flow: [Execute SQL query](https://docs.lightdash.c
                 );
             }
 
-            results = await dashboardService.create(
-                toSessionUser(req.account),
+            results = await dashboardService.createFromAccount(
+                req.account!,
                 projectUuid,
                 body,
             );
--- a/packages/backend/src/controllers/v2/ContentController.ts
+++ b/packages/backend/src/controllers/v2/ContentController.ts
@@ -29,7 +29,7 @@ import {
     Tags,
 } from '@tsoa/runtime';
 import express from 'express';
-import { toSessionUser } from '../../auth/account';
+import { getAccountWriteContext, toSessionUser } from '../../auth/account';
 import { ContentArgs } from '../../models/ContentModel/ContentModelTypes';
 import { allowApiKeyAuthentication, isAuthenticated } from '../authentication';
 import { BaseController } from '../baseController';
@@ -58,15 +58,19 @@ export class ContentController extends BaseController {
         @Query() sortBy?: ContentArgs['sortBy'],
         @Query() sortDirection?: ContentArgs['sortDirection'],
     ): Promise<ApiContentResponse> {
-        assertRegisteredAccount(req.account);
+        const { user, embedWriteActions } = getAccountWriteContext(
+            req.account!,
+        );
         this.setStatus(200);
         return {
             status: 'ok',
             results: await this.services.getContentService().find(
-                toSessionUser(req.account),
+                user,
                 {
                     projectUuids,
-                    spaceUuids,
+                    spaceUuids: embedWriteActions
+                        ? [embedWriteActions.spaceUuid]
+                        : spaceUuids,
                     contentTypes,
                     search,
                 },
--- a/packages/backend/src/controllers/v2/QueryController.ts
+++ b/packages/backend/src/controllers/v2/QueryController.ts
@@ -321,11 +321,6 @@ export class QueryController extends BaseController {
 
         const context = body.context ?? getContextFromHeader(req);
 
-        if (req.account!.isJwtUser()) {
-            // we need more granular CASTL abilities before enabling this
-            throw new ForbiddenError('Feature not available for JWT users');
-        }
-
         const results = await this.services
             .getAsyncQueryService()
             .executeAsyncDashboardChartQuery({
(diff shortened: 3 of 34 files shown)


rephus:
* **#24226** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/backend/src/controllers/v2/QueryController.ts
```diff
@@ -321,11 +321,6 @@ export class QueryController extends BaseController {
 
         const context = body.context ?? getContextFromHeader(req);
 
-        if (req.account!.isJwtUser()) {
```
rephus, line 324: why awas this removed ?

packages/backend/src/controllers/v2/ContentController.ts
```diff
@@ -58,15 +58,19 @@ export class ContentController extends BaseController {
...
         this.setStatus(200);
         return {
             status: 'ok',
             results: await this.services.getContentService().find(
-                toSessionUser(req.account),
+                user,
                 {
                     projectUuids,
-                    spaceUuids,
+                    spaceUuids: embedWriteActions
```
rephus, line 71: For embed write-action JWTs, the requested `spaceUuids` are intentionally ignored and replaced with the token write space. The add-tile picker uses this endpoint, so the JWT should not be able to enumerate charts/content from another space by changing query params.

packages/backend/src/controllers/projectController.ts
```diff
@@ -947,8 +946,8 @@ Migrate to the v2 async query flow: [Execute SQL query](https://docs.lightdash.c
                 );
             }
 
-            results = await dashboardService.create(
-                toSessionUser(req.account),
+            results = await dashboardService.createFromAccount(
```
rephus, line 950: This deliberately routes through the account-aware dashboard service instead of converting straight to `SessionUser`. The embed JWT write flow needs the original account so the service can see `writeActions.spaceUuid` and use the configured service-account user while still enforcing the write-space boundary.

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -745,6 +753,153 @@ export class DashboardService
...
+            dashboard.tiles,
+            embedWriteActions.spaceUuid,
+            projectUuid,
+        );
+
+        return {
+            user,
+            dashboard: {
+                ...dashboard,
+                spaceUuid: embedWriteActions.spaceUuid,
```
rephus, line 909: The token write space is the source of truth here, even if the request body includes a different `spaceUuid`. That keeps embedded dashboard creation equivalent to normal dashboard creation while preventing the embed caller from choosing a target space outside the JWT grant.

packages/backend/src/services/DashboardService/DashboardService.ts
```diff
@@ -978,11 +1178,61 @@ export class DashboardService
...
+                dashboard.spaceUuid &&
+                dashboard.spaceUuid !== embedWriteActions.spaceUuid
+            ) {
+                throw new ForbiddenError(
+                    'Embed token does not allow moving dashboards outside the write space',
+                );
+            }
+
+            if (isDashboardVersionedFields(dashboard)) {
+                await this.assertDashboardTilesBelongToWriteSpace(
```
rephus, line 1296: On update we re-check the full tile payload, not just newly added tiles. The client sends the dashboard version as a complete array, so without this guard an embed caller could save a dashboard that references a saved chart or SQL chart outside the JWT write space.

packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
```diff
@@ -5906,30 +5952,70 @@ export class AsyncQueryService extends ProjectService {
...
+        // For JWT users without write actions, the dashboard scope is bound to
+        // the token. Write-action embeds may use a dashboard from their write
+        // space so newly added draft tiles can preview with dashboard context.
+        if (isJwtUser(account)) {
+            const embedWriteActions = account.authentication.data.writeActions;
+            const writeSpaceUuid = account.embedWriteUser
+                ? embedWriteActions?.spaceUuid
+                : undefined;
+
+            if (writeSpaceUuid) {
```
rephus: This branch is what lets embedded dashboard editing preview newly added SQL chart tiles. It is intentionally narrower than normal embed chart access: both the SQL chart and the dashboard must be in the JWT write space before we bypass the per-chart embed permission check.

packages/frontend/sdk/index.tsx
```diff
@@ -145,68 +226,170 @@ const Dashboard: FC<DashboardProps> = ({
...
+                    containerStyles={getDashboardContainerStyles(styles, theme)}
+                    isEditMode={isEditMode}
+                    onEditModeChange={onEditModeChange}
+                />
+            </EmbedProvider>
+        </SdkProviders>
+    );
+};
+
+const dashboardBuilderCreatePromises = new Map<
```
rephus, line 258: This module-level map is an SDK-specific guard against duplicate dashboard creation from React remounts/StrictMode while the create request is in flight. It is keyed by project + write space (+ dashboard content id when present), and failures clear the entry so the visible error state can be retried by remounting.

PR #24229, opened 2026-06-12
feat(embedding): sticky filter bar and tabs on embedded dashboards

### LINKED ISSUE #24140: [Embedding] Sticky filter bar not applied to embedded dashboards
## Use case

When a dashboard is embedded, the filter bar (and tab switcher, when tabs are enabled) scrolls out of view as the user scrolls down through charts. This makes it hard for end users to see which filters are active or change them without scrolling all the way back up. This is especially problematic on long dashboards with multiple charts per tab.

The regular dashboard view already ships a sticky filter bar (via `StickyWithDetection` in `features/dashboardTabs/index.tsx`), but the embed view does not apply the same treatment.

## Proposed solution

Apply sticky positioning to `EmbedDashboardHeader` (currently `pos="relative"`) so that the filter bar - and tab switcher when tabs are enabled - stays pinned to the top of the embed's scroll container, mirroring the behaviour in the regular dashboard view.

The embed scroll container is a `div` with `height: 100vh; overflowY: auto`. Pinning the header to the top of this container (rather than the viewport) would keep it visible while the rest of the dashboard scrolls.

## Additional context

The sticky filter bar for regular dashboards was shipped in a recent release. The embed view was not updated at the same time, leaving a parity gap between the two surfaces.

tatianainama:
### Description:

Adds an opt-in sticky header for embedded dashboards, plus a reworked embed header layout.

**New `stickyHeader` embed option** (JWT content flag, default off — existing embeds keep the current scroll-away behavior):
- When enabled, dashboard tabs and the filter bar pin to the top of the embed scroll container while tiles scroll underneath.
- Plumbed through the embed JWT schema, `EmbedService`, the embed preview form (new "Appearance" toggle), the generated backend code snippets (TS/Python/Go), and the `embed_dashboard.viewed` analytics event.

**Header rework** (applies to all embeds):
- The header is now a stateless CSS grid: tabs (or filter bar) + a reserved actions column for the print button, with an optional full-width filter bar row below the tabs.
- Stickiness and the "stuck" bottom border are pure CSS: `data-sticky` attribute + `position: sticky`, with stuck detection via scroll-state container queries (`container-type: scroll-state`) — no IntersectionObserver or React state. Browsers without scroll-state support (Firefox/Safari) keep the sticky bar and just never show the border.
- The print button no longer floats over chart content; it has a dedicated slot in the header in every layout, is hidden from printed output via `@media print` (along with any open tooltip), and `EmbedDashboardExportPdf` is migrated from Mantine v6 to v8.
- The filter bar collapses to a summary ("N filters · Date Zoom"), only shows the Hide button when there are actually filters/parameters to hide, and renders nothing when interactivity is enabled but there's no visible content (no more empty padded row under the tabs).

Closes: #24140

#### Preview

CleanShot 2026-06-12 at 12 33 16

CleanShot 2026-06-12 at 12 37 58

### DIFF
--- a/packages/backend/src/ee/analytics/index.ts
+++ b/packages/backend/src/ee/analytics/index.ts
@@ -15,6 +15,7 @@ export type EmbedDashboardViewed = BaseTrack & {
         canExportImages?: boolean;
         canExportPagePdf: boolean;
         canDateZoom?: boolean;
+        stickyHeader?: boolean;
     };
 };
 
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -549,6 +549,7 @@ export class EmbedService extends BaseService {
             canExplore,
             canViewUnderlyingData,
             canViewDataApps,
+            stickyHeader,
         } = decodedToken.content;
         // Embed paletteUuid query param overrides everything; otherwise fall back
         // through chart → dashboard → space → project → org via the resolver.
@@ -595,6 +596,7 @@ export class EmbedService extends BaseService {
                 canExportImages,
                 canExportPagePdf: canExportPagePdf ?? true,
                 canDateZoom,
+                stickyHeader,
                 ...(account.access.filtering
                     ? {
                           dashboardFiltersInteractivity: {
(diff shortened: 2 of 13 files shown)


tatianainama:
* **#24229** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24229?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24239, opened 2026-06-12
fix(ai-review): wait for preview compile before verification; skip reviews on preview projects

joaoviana:
## Problem

When a review finding's writeback opens a PR, the remediation flow spins up a preview project and immediately seeds a verification prompt on the cloned agent. The preview's `COMPILE_PROJECT` job is still running at that point (the prompt fired ~4s after project creation in a traced run; the compile took ~14s), and the agent run snapshots `availableExplores` once at run start — so `generateVisualization` failed with `Explore not found` for the entire run even after the compile landed mid-run, while the live-querying catalog tools (`findExplores`/`findFields`) did see the new metric.

Faced with those contradictory signals, the agent fell back to raw `runSql` against the warehouse — bypassing the explore's `sql_filter` — and reported confident but ungoverned numbers plus a false "issue only partially resolved" verdict.

On top of that, the verification thread's turns were themselves picked up by the review classifier, feeding the reviewer's own verification output back into review findings.

## Changes

- New `aiAgentReviewRemediationCompile` scheduler task: a self-rescheduling poll (10s interval, 10 min timeout via `startedAt`) that checks the preview's compile job and only seeds the verification thread once it is `DONE`. Mirrors the existing `aiAgentReviewRemediationPreview` poll pattern — each poll is a short job, state lives in the job queue, so no worker slot is held during the wait and a deploy or pod restart mid-wait cannot strand the remediation (the writeback job itself has `maxAttempts: 1`, so an inline wait would have been killed permanently). Compile failure or timeout marks the remediation `failed` with a clear message; a poll firing after the remediation left `pr_open` is a no-op.
- The verification prompt now instructs the agent to answer through governed explores only and to report — not work around — explores that can't be queried.
- Review classification skips preview projects entirely:
  - `runLiveEvent` exits early when the event's project is a `PREVIEW` project.
  - `listTurnReviewCandidates` joins `projects` and excludes preview-project threads, so org-wide backfill/batch runs can't sweep them in either.

## Regression analysis

- **Writeback remediation flow**: verification now starts up to ~compile-duration later (typically seconds, bounded at 10 min). The writeback job itself finishes sooner — it ends after enqueueing the poll instead of blocking until verification is seeded, matching how the CI-comment preview path already behaves.
- **Review classification on real (non-preview) projects**: unchanged. The new `projects` join is an inner join on `thread.project_uuid`; threads on deleted projects were already unreviewable.
- **Preview projects**: lose review classification by design. No existing feature depends on reviewing preview turns.
- **`createPreviewForPullRequest` callers**: unchanged — the wait lives in the remediation flow only, so the in-thread writeback path (agent replying with a preview link mid-conversation) does not block on compile.

## Testing

- Unit tests: compile-poll enqueue from the writeback job; poll behavior for DONE / RUNNING (re-enqueue) / ERROR / timeout / stale-status guard; the governed-explores prompt instruction; the live-event preview guard; and the candidate-query preview exclusion.
- Root cause traced end-to-end on a live instance (thread/tool-call/job timeline from the production DB).

Relates: ZAP-486

### DIFF
--- a/packages/backend/src/ee/index.ts
+++ b/packages/backend/src/ee/index.ts
@@ -334,6 +334,7 @@ export async function getEnterpriseAppArguments(): Promise<EnterpriseAppArgument
                     lightdashConfig: context.lightdashConfig,
                     writebackPreviewService:
                         repository.getWritebackPreviewService<WritebackPreviewService>(),
+                    jobModel: models.getJobModel(),
                 }),
             aiRouterService: ({ models, repository, context }) =>
                 new AiRouterService({
@@ -362,6 +363,7 @@ export async function getEnterpriseAppArguments(): Promise<EnterpriseAppArgument
                     aiOrganizationSettingsModel:
                         models.getAiOrganizationSettingsModel(),
                     catalogModel: models.getCatalogModel(),
+                    projectModel: models.getProjectModel(),
                     featureFlagService: repository.getFeatureFlagService(),
                     lightdashConfig: context.lightdashConfig,
                 }),
--- a/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
+++ b/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
@@ -1,3 +1,4 @@
+import { ProjectType } from '@lightdash/common';
 import knex, { Knex } from 'knex';
 import { getTracker, MockClient, Tracker } from 'knex-mock-client';
 import { AiPromptTableName } from '../database/entities/ai';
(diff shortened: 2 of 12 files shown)


joaoviana:
* **#24239** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24239?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24239?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24236, opened 2026-06-12
feat: add underlying data download to query sdk

magnew:
Relates to: PROD-8065

### Description:

Adds an callback for data apps to download underlying data through the BE mechanism.

PR #24234, opened 2026-06-12
feat: export data from data apps

magnew:
Relates to: #23759 

### Description:

Adds backend-powered data downloads for data apps via the query SDK.

- Adds downloadResults() to useLightdash()
- Supports CSV/XLSX, formatted/raw values, and table/all/custom row limits
- Uses the existing Lightdash backend export job flow instead of serializing data in the iframe
- Allows the data app iframe to trigger backend-generated downloads
- Extends the app SDK bridge allowlist for download scheduling and job polling
- Updates data app docs, SDK README, and agent skill guidance so generated apps can add export controls with loading states

PR #24201, opened 2026-06-11
chore: update data apps internal docs

magnew:
### Description:

Add recent features to data apps docs

PR #24243, opened 2026-06-12
feat: add FE PDF exports for data apps

### LINKED ISSUE #23759: [Data Apps] Enable file downloads (allow-downloads) and client-side PDF export from the iframe sandbox
## Use case

Developers building data apps want to offer end users a "Download PDF" button that saves a formatted report of dashboard tables directly to disk. This is currently blocked at two layers.

## What's blocked

1. The data app iframe runs with `sandbox="allow-scripts allow-modals allow-popups"` (`packages/frontend/src/features/apps/AppIframePreview.tsx:183`). It does not include `allow-downloads`, so even a correctly generated Blob/data URI cannot be saved via a standard `<a download>` click.
2. The approved package set for data apps (`sandboxes/data-apps/template/package.json`) has no client-side PDF generation library, so apps cannot programmatically build a PDF in the browser.

## Proposed solution

1. Add `allow-downloads` to the iframe `sandbox` attribute. This is independent of `allow-same-origin` (deliberately omitted) and does not weaken origin isolation. It unblocks all blob downloads — CSV, JSON, images — not just PDF, so it is valuable on its own.
2. Add a client-side PDF library to the approved packages (e.g. `jsPDF`, plus `jspdf-autotable` for table formatting). Same category of addition as `recharts` / `d3` / `html-to-image` already present. Requires updating the locked `pnpm-lock.yaml` and the approved-package list in the data app template/skill (installs are disabled in the sandbox by design).

With both, an app can generate a PDF blob and trigger a direct download — no print dialog or popup required.

## Current workaround

Build a print-optimised HTML doc in memory, open via `window.open()`, and call `window.print()` on load (the sanctioned path today, enabled by `allow-modals`). Fails silently when popups are blocked and needs a manual "Save as PDF" step.

## Notes

`allow-downloads` is the high-value, near-zero-cost change and could ship independently of the PDF library.

magnew:
Closes: #23759 

### Description:

Adds PDF downloads of Data Apps. This PR makes it possible to download as *image-based* PDFs. 

Text PDFs are out of scope.

### DIFF
--- a/docs/data-apps.md
+++ b/docs/data-apps.md
@@ -435,7 +435,7 @@ The preview iframe uses `sandbox="allow-scripts allow-modals allow-downloads all
 - The iframe cannot access the parent page's cookies or storage
 - The iframe cannot make credentialed requests to the Lightdash API directly
 - All API communication goes through a `postMessage` bridge
-- The iframe can trigger user-initiated downloads for backend-generated CSV/XLSX exports
+- The iframe can trigger user-initiated downloads for backend-generated CSV/XLSX exports and client-generated PDF files
 
 `allow-modals` is enabled so that PDF Report templates (and any generated app that wants a Print button) can call
 `window.print()`. It also enables `alert`/`confirm`/`prompt`, which are an accepted trade-off: the iframe is still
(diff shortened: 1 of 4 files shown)


magnew:
* **#24243** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24243?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24236** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24236?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24201** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24234** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24234?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #24247, opened 2026-06-12
fix: widen slack profile photo url column to prevent save error

### LINKED ISSUE #24242: Saving Slack custom settings fails when the Profile photo URL is too long
### Description

Updating the Slack integration's custom settings fails with a server error when the **Profile photo URL** is longer than 255 characters.

The URL input only validates that the value is a well-formed URL, so a long but otherwise valid URL (for example, a CDN or signed URL with a long query string) is accepted in the UI. When the settings are saved, the request fails and the settings are not persisted.

**Expected:** longer URLs should be accepted, or the form should show a clear validation message before submitting instead of failing on save.

1. Go to **Settings → Integrations → Slack** custom settings.
2. In the **Profile photo URL** field, paste a valid URL that is longer than 255 characters.
3. Click **Save**.
4. The save fails with a server error and the change is not saved.

### version

_Not specified_

### Cloud or self-hosting

_Not specified_

tatianainama:
## What & why

The `app_profile_photo_url` column on `slack_auth_tokens` used Knex's default `varchar(255)`, so a long but valid profile-photo URL (e.g. a signed CDN/object-storage URL) overflowed the column and `PUT /slack/custom-settings` failed with a 500.

## Changes

- **Migration**: widen `app_profile_photo_url` to `varchar(2048)` (the de-facto safe URL length). This is a metadata-only change in Postgres — no table rewrite.
- **Frontend**: add a matching `.max(2048)` to the Slack settings Zod schema so an over-long URL surfaces an inline validation message instead of triggering the "Failed to update Slack app settings" server-error toast.

## Testing

- Migration applied locally; column confirmed at `varchar(2048)`.
- Backend + frontend typecheck and lint pass on the changed files.

Closes: #24242

### DIFF
--- a/packages/backend/src/database/migrations/20260612151235_widen_slack_app_profile_photo_url.ts
+++ b/packages/backend/src/database/migrations/20260612151235_widen_slack_app_profile_photo_url.ts
@@ -1,0 +1,18 @@
+import { Knex } from 'knex';
+
+const SlackAuthTokensTableName = 'slack_auth_tokens';
+
+// Profile photo URLs (e.g. signed CDN/object-storage URLs) can exceed the
+// varchar(255) Knex default, which caused saves to fail. 2048 is the de-facto
+// safe URL length. Widening is a metadata-only change in Postgres (no rewrite).
+export async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(SlackAuthTokensTableName, (table) => {
+        table.string('app_profile_photo_url', 2048).nullable().alter();
+    });
+}
+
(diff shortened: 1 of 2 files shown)


tatianainama:
* **#24247** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24247?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24247?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24250, opened 2026-06-12
fix: import moment-timezone in time.ts so CLI compile of inThePast timestamp filters works

### LINKED ISSUE #24176: [Filters] CLI compile fails with getMomentDateWithCustomStartOfWeek(...).tz is not a function when metric filter uses inThePast on a timestamp dimension
## Bug description

When a metric in a dbt YAML file defines an `inThePast` filter against a `timestamp` dimension, running `lightdash compile` via the CLI fails with the following error:

`(0 , time_1.getMomentDateWithCustomStartOfWeek)(...).tz is not a function`

## Steps to reproduce

1. Have a dimension backed by a Snowflake TIMESTAMP_TZ or TIMESTAMP_NTZ column
2. Define a metric in dbt YAML with an `inThePast` filter against that dimension
3. Run `lightdash compile` from the CLI
4. Compilation fails with: `getMomentDateWithCustomStartOfWeek(...).tz is not a function`

## Expected behavior

`lightdash compile` succeeds. `inThePast` filters on timestamp dimensions compile correctly in the CLI, consistent with server-side compile behavior.

## Actual behavior

Compilation crashes with a TypeError, tracing back to the call chain in `filtersCompiler.ts` where `.tz(timezone)` is called on the moment object returned by `getMomentDateWithCustomStartOfWeek`.

## Proposed fix

Change `packages/common/src/utils/time.ts` to import from `moment-timezone` instead of plain `moment`:

```diff
- import moment from 'moment';
+ import moment from 'moment-timezone';
```

Both packages are already declared as dependencies of `@lightdash/common`, so no new dependency is needed.

## Environment

- Lightdash CLI: 0.3098.x
- Warehouse: Snowflake (TIMESTAMP_TZ and TIMESTAMP_NTZ dimension types both affected)

tatianainama:
## Summary

`lightdash compile` in the CLI crashes with `getMomentDateWithCustomStartOfWeek(...).tz is not a function` when a dbt metric defines an `inThePast` filter against a `timestamp` dimension.

`getMomentDateWithCustomStartOfWeek` (in `packages/common/src/utils/time.ts`) imported plain `moment`, whose objects have no `.tz()` method. `filtersCompiler.ts` calls `.tz(timezone)` on that result in the relative-date operator paths (`IN_THE_PAST` etc.).

`.tz` only exists once `moment-timezone` has patched the shared `moment` singleton. On the server and frontend that always happens (moment-timezone is loaded elsewhere first), so the bug is invisible there. In the **ncc-bundled CLI**, `time.ts`'s `moment` reference isn't guaranteed to be the patched one at call time, so `.tz` is `undefined` and compile crashes.

## Fix

Import `moment-timezone` in `time.ts` (the package is already a dependency and is already imported by other files in `common`, including `filtersCompiler.ts`). This guarantees `.tz` is present regardless of bundling/load order.

## Notes

- **why** — the plain-`moment` import dates to 2023. The crash only became reachable when `.tz()` was applied to the `IN_THE_PAST` path (#21637, Apr 2026), which is why reports only surfaced recently and only via the CLI.

Closes #24176

### DIFF
--- a/packages/common/src/utils/time.ts
+++ b/packages/common/src/utils/time.ts
@@ -1,4 +1,4 @@
-import moment from 'moment';
+import moment from 'moment-timezone';
 import { isWeekDay, type WeekDay } from './timeFrames';
 
 // from 0 (Monday) to 6 (Sunday) to 0 (Sunday) to 6 (Saturday)


tatianainama:
* **#24250** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24250?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24250?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24255, opened 2026-06-12
docs: update CLI skill deploy guidance

jpetey75:
Closes:

### Description:

Updates the `developing-in-lightdash` skill that is installed by `lightdash install-skills` so agents do not treat `lightdash deploy` as the default way to update production projects.

What changed:
- Adds explicit guidance that production semantic-layer updates should almost always go through CI/CD, such as GitHub Actions after merge, or be refreshed from the Lightdash UI.
- Reframes direct `lightdash deploy` usage as appropriate for non-production projects or rare user-confirmed exceptions.
- Updates workflow examples so manual/preview flows do not end with an ad hoc production deploy, while preserving production `lightdash deploy` examples inside CI/CD workflows.

Why:
- Agents were sometimes running `lightdash deploy` directly, which can bypass review and the team's normal production release path.

Validation:
- `git diff --check`
- Commit hooks: git-secrets and configured staged-file checks


jpetey75:
@oli-rmsy can you review this update and let me know if you disagree with the changes at all? My AIs are reaching for `lightdash deploy` when they shouldn't and I think that could cause bad things.

PR #24268, opened 2026-06-15
fix: dashboard chart totals and subtotals use active parameter value instead of default

### LINKED ISSUE #24149: [Parameters] Dashboard chart totals and subtotals use default parameter value instead of active selection
## Bug description

On a dashboard with a parameter control, changing the parameter correctly re-queries and updates per-row values of a flat table. However, two rendering paths ignore the change and always display values at the parameter default:

1. Subtotal rows (showSubtotals: true) - frozen at the default parameter value.
2. Column total row (showColumnCalculation: true) - frozen at the default, even when the individual rows in a flat table have updated correctly.

## Steps to reproduce

1. Define a model-level parameter with options A and B and default A.
2. Define dimensions whose SQL switches on the parameter value (CASE WHEN mode = A THEN col_a ELSE col_b END).
3. Build two table charts with identical queries: Chart 1 with showSubtotals and showColumnCalculation enabled, Chart 2 with only showColumnCalculation enabled (flat).
4. Add both charts to a dashboard that exposes the parameter control.
5. Change the parameter from A to B and wait for tiles to re-run.

## Expected behavior

Changing the parameter should re-query and update the subtotal rows of Chart 1 and the column total row of both charts, exactly as the flat per-row values already update in Chart 2.

## Actual behavior

- Chart 1 (showSubtotals: true): row values do not change at all. Frozen at default (A). x
- Chart 2 (showSubtotals: false): per-row values change correctly. check
- Both charts: the bottom Total row stays at the default (A) value. x

## Root cause

In AsyncQueryService.executeAsyncDashboardChartQuery, the requestParameters object stored to query history omits the parameters field. When executeAsyncCalculateTotalFromQueryHistory re-runs totals or subtotals for the source query, it reads source.requestParameters?.parameters which is undefined, so the totals fall back to project-default parameter values. The Explore-view path (executeAsyncMetricQuery) correctly stores parameters: combinedParameters and is not affected.

## Fix

Add parameters: combinedParameters to the requestParameters object in executeAsyncDashboardChartQuery after combinedParameters is derived.

## Workarounds

- Subtotals bug: disable subtotals (showSubtotals: false) so rows respond to the parameter.
- Column total bug: hide the total row (showColumnCalculation: false) to avoid showing a misleading value.

## Environment

- Lightdash version: 0.2995.0
- Deployment: Cloud

tatianainama:
## What

When a dashboard has a parameter control, changing the parameter correctly updates per-row values in flat table charts, but totals and subtotals were always frozen at the default parameter value.

**Root cause:** In `executeAsyncDashboardChartQuery`, the `requestParameters` object was constructed before `combinedParameters` was resolved, so `parameters` was never persisted to query history. When `executeAsyncCalculateTotalFromQueryHistory` re-ran totals/subtotals, it read `source.requestParameters?.parameters` and got `undefined`, causing the totals query to fall back to project-default parameter values.

The explore-view path (`executeAsyncMetricQuery`) was not affected because it correctly stores `parameters: combinedParameters`.

## Fix

Move `requestParameters` construction to after `combinedParameters` is resolved and add `parameters: combinedParameters` to it.

Closes: #24149

### DIFF
--- a/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
+++ b/packages/backend/src/services/AsyncQueryService/AsyncQueryService.ts
@@ -5081,17 +5081,6 @@ export class AsyncQueryService extends ProjectService {
             };
         }
 
-        const requestParameters: ExecuteAsyncDashboardChartRequestParams = {
-            tileUuid,
-            chartUuid,
-            context,
-            dashboardUuid: resolvedDashboardUuid,
-            dashboardFilters,
-            dashboardSorts,
-            dateZoom,
-            limit,
-        };
-
         const queryTags: RunQueryTags = {
             ...this.getUserQueryTags(account),
             ...AsyncQueryService.getSchedulerQueryTags(),
@@ -5147,6 +5136,18 @@ export class AsyncQueryService extends ProjectService {
             dashboardParameters,
             projectParameters,
         );
+
+        const requestParameters: ExecuteAsyncDashboardChartRequestParams = {
+            tileUuid,
+            chartUuid,
+            context,
+            dashboardUuid: resolvedDashboardUuid,
+            dashboardFilters,
+            dashboardSorts,
+            dateZoom,
+            limit,
+            parameters: combinedParameters,
+        };
 
         const { fields, dateZoomApplied } = await this.getMetricQueryFields({
             metricQuery: metricQueryWithLimit,


tatianainama:
* **#24268** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24268?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24292, opened 2026-06-15
fix(echarts): restore native sub-day time-axis ticks (GLITCH-502)

almeidabbm:
## What & why

Closes: **GLITCH-502** (follow-up to GLITCH-449)

GLITCH-449 fixed sub-day (SECOND/MINUTE/HOUR) DST breakage by positioning bars at raw UTC instants and pinning the axis ticks/labels to **every distinct bucket instant** via `axisTick.customValues` + `axisLabel.customValues`. Side effect: ticks became static — one per bucket, no adaptive density, no re-densification on zoom.

The real GLITCH-449 acceptance criteria only required DST continuity (no jump / duplicated / missing hour at spring-forward and fall-back). The per-bucket tick pinning was self-imposed and is the thing this PR removes.

## Change

Subtractive, one file (`useEchartsCartesianConfig.ts`). In the `sub-day-format` branch of `getSubDayTimeAxisConfig`, drop both `customValues` and keep only the in-zone label formatter:

- ECharts runs its **native adaptive time-axis tick algorithm** again (axis is already raw-UTC + `useUTC: true`).
- `formatSubDayAxisLabel` relabels each native tick into the resolved project zone.
- Bar positions (raw UTC instants) and the `resolveAxisTimezone` routing are untouched, so **DST continuity is preserved**.
- Also removes the per-render O(rows) `customValues` scan.

This is the `useUTC` + label-formatter path the ticket itself suggested.

## Gating (unchanged)

The flag gate is upstream in `resolveAxisTimezone` → `detectTimezoneShiftedField`, which early-returns when `resolvedTimezone` is null/`UTC`. With `enable-timezone-support` **off**, `timeAxisField` is `undefined`, so `getSubDayTimeAxisConfig` returns `{}` and ECharts' default UTC ticks/labels render — byte-identical to pre-feature. Verified empirically: with the flag off, `formatSubDayAxisLabel` is invoked **0 times** (instrumented, then reverted).

## Test plan

- Unit: `useEchartsCartesianConfig.test.ts` + `timezoneShift.test.ts` — 121 pass. Retired the `customValues`-equal-bucket-instants tests, added a native-ticks assertion (`axisTick` undefined, formatter present) and a flag-off gating test.
- Live (BigQuery `timezone_test`, HOUR grain, America/New_York): native adaptive ticks with the flag on; identical tick density + format with the flag off (only the label timezone differs); DST fall-back (Nov 3) keeps the two folded 01:00 hours as distinct bars at raw-UTC positions.

## Screenshots

| | Flag OFF | Flag ON |
|---|---|---|
| **Before fix** | before-fix-flagOFF | before-fix-flagON |
| **After fix** | after-fix-flagOFF | after-fix-flagON |

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.test.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.test.ts
@@ -897,7 +897,7 @@ describe('getSubDayTimeAxisConfig', () => {
         expect(result).toEqual({});
     });
 
-    test('uses raw UTC instants as custom tick values across DST fall-back', () => {
+    test('leaves tick placement to ECharts (emits no customValues)', () => {
         const result = getSubDayTimeAxisConfig(
             axisId,
             createRows([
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#24296** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24296?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24292** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24295, opened 2026-06-15
feat: support one-character CJK omnibar searches

### LINKED ISSUE #24286: [Global Search] Support 1-character minimum for CJK language queries in omnibar
## Use case

In CJK (Chinese, Japanese, Korean) languages, meaningful words are frequently 1-2 characters long. For example, common Japanese business terms like 売上 ("sales") or 表 ("table") are only 2 characters. The current 3-character minimum means these queries never trigger a search, making the global search (Cmd/Ctrl+K omnibar) effectively unusable for CJK language users.

## Proposed solution

Detect whether the query contains any CJK Unicode characters and apply a lower minimum threshold (1 character) in that case, while keeping the existing 3-character minimum for non-CJK text.

The relevant constant is `OMNIBAR_MIN_QUERY_LENGTH = 3` in `packages/frontend/src/features/omnibar/hooks/useSearch.ts`. The `enabled` check and the `hasEnteredMinQueryLength` guard in `Omnibar.tsx` would both need to account for CJK input.

CJK Unicode blocks to detect:
- U+4E00–U+9FFF: CJK Unified Ideographs
- U+3040–U+309F: Hiragana
- U+30A0–U+30FF: Katakana
- U+AC00–U+D7AF: Hangul Syllables
- U+3400–U+4DBF: CJK Unified Ideographs Extension A

## Alternatives considered

- Lowering the global minimum to 1-2 characters: could increase noise and unnecessary API calls for non-CJK users.
- Using `Intl.Segmenter` for language-aware segmentation: more robust but more complex; per-character CJK detection is simpler and likely sufficient.

st-lab7:
Closes #24286

### Description:
This PR updates omnibar search minimum query length handling for CJK queries.

Previously, omnibar search required at least 3 characters before showing results. That works well for English queries, but it makes Japanese, Chinese, and Korean searches harder to use because a single CJK character can already be a meaningful search term.

This change keeps the existing 3-character minimum for non-CJK queries, while allowing queries that contain CJK characters to search from 1 character.

The shared minimum-length helper is used in:
- omnibar search query enabling
- omnibar empty/searching state handling
- settings search items
- AI search dropdown behavior

It also adds unit tests covering:
- empty and whitespace-only queries
- ASCII queries below and at the 3-character minimum
- Han, Hiragana, Katakana, and Hangul single-character queries
- mixed ASCII/CJK queries

image

### DIFF
--- a/packages/frontend/src/ee/components/Home/AiSearchBox/SearchDropdown.tsx
+++ b/packages/frontend/src/ee/components/Home/AiSearchBox/SearchDropdown.tsx
@@ -15,7 +15,7 @@ import MantineIcon from '../../../../components/common/MantineIcon';
 import OmnibarItem from '../../../../features/omnibar/components/OmnibarItem';
 import { getSearchResultsGroupsSorted } from '../../../../features/omnibar/components/utils';
 import useSearch, {
-    OMNIBAR_MIN_QUERY_LENGTH,
+    hasMinQueryLength,
 } from '../../../../features/omnibar/hooks/useSearch';
 import { type SearchItem } from '../../../../features/omnibar/types/searchItem';
 import { getSearchItemLabel } from '../../../../features/omnibar/utils/getSearchItemLabel';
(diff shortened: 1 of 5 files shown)


jpetey75:
I tested this locally and worked well for me! I'll loop in someone from the engineering team for final review and to get it merged.

tatianainama:
Thank you @st-lab7 for the contribution! this looks great. I'm running e2e tests as last check and we will be merging this as soon as everything is green!

tatianainama:
@all-contributors add @st-lab7 for code

tatianainama (approved):

PR #24304, opened 2026-06-15
perf(preview): stream-parse the S3 manifest, keeping only used keys

### LINKED ISSUE #24299: UI preview: auto-source the combined manifest.json from S3 (multi-repo dbt)
## Problem

Multi-repo dbt setups merge models from many repos into a single **combined** `manifest.json`. In production that combined manifest is large — the pilot customer's is **\~400MB**. UI preview creation only accepts a *pasted* manifest string, which is unusable at this size. The CLI handles it via `lightdash preview --combine-manifest=<path>`, but that's CLI-only and can't be delegated to non-technical users. Without the combined manifest a UI preview only sees one repo's models, so cross-repo joins are skipped and dashboards look broken.

This blocks the self-serve preview → promote flow (the new preview-edit custom-role scopes) for the customer's non-technical stakeholders — the exact users it was built for.

**Context (internal):** customer Slack thread, 2026-06-15. Both paste and file upload are rejected by the customer: *"It's not great us having end users downloading/reuploading 400MB of JSON … we'd really rather that we just shipped this file directly to S3 and Lightdash pick it up/use it from there."* The customer already publishes the combined manifest from their CI pipeline. (Account specifics live in the customer-success repo, not here.)

## Decision

Project is configured with an **S3 location** holding the prod combined manifest. UI previews fetch it from S3 at compile time and use it automatically. The customer's pipeline keeps the S3 object current. We do **not** ask the user to supply anything.

## Key constraints (these shape the design)

1. **\~400MB.** Cannot go in the API request body (`express.json` limit `LIGHTDASH_MAX_PAYLOAD`, default 5mb), cannot go through the browser, and **cannot be stored inline** in the `projects.dbt_connection` encrypted Buffer — that would bloat Postgres on every preview. The preview must store an **S3 reference**, fetched lazily.
2. **Already combined.** The customer's pipeline merges the repos. The backend does **not** need `combineManifests` — it just fetches and uses. (Server-side combine of multiple manifests stays out of scope; that's the CLI's job for single-repo local previews.)
3. **Backend already has S3 access** via `FileStorageClient` / `S3Client` — `getFileStream(key)` streams arbitrary keys (`packages/backend/src/clients/Aws/S3Client.ts:240`).

### 1. Project-level config — S3 manifest source

* New nullable column `projects.preview_manifest_s3_path` (text). Migration follows the `20260515094200_add_preview_expiration_project_settings.ts` pattern.
* Add to `DbProject` (`packages/backend/src/database/entities/projects.ts`) and read/write in `ProjectModel`.
* v1: settable by project admin (project settings), or seeded internally for the pilot project to unblock fast. Productize the settings UI as a follow-up.

### 2. Preview dbt connection references S3, not inline

Today `maybeOverrideDbtConnection` (`packages/common/src/types/projects.ts:855`) builds `DbtManifestProjectConfig = { type: MANIFEST; manifest: string; hideRefreshButton: true }` — inline only. Extend the manifest source to a discriminated shape (no duck typing, per repo style):

```
manifestSource:
  | { type: 'inline'; manifest: string }   // existing path (paste/CLI)
  | { type: 's3'; key: string }            // new
```

Previews are **ephemeral** (they expire), so no data migration of existing `dbt_connection` blobs is needed; keep reading the legacy inline `manifest` field during transition.

### 3. Wire-up at preview creation — `ProjectService.createPreview` (`ProjectService.ts:7898`)

* If the request supplies an explicit manifest override → inline path, as today.
* Else if the upstream project has `preview_manifest_s3_path` set → build the preview dbt connection as MANIFEST with `manifestSource { type: 's3', key }`. **No S3 fetch at creation time** — keep creation fast (preview creation is already slow at ~4–5 min; don't make it worse).

### 4. Fetch + use at compile time — `dbtManifestProjectAdapter` / `ManifestDbtClient` (`packages/backend/src/projectAdapters/dbtManifestProjectAdapter.ts:59`)

* When `manifestSource.type === 's3'`: `getFileStream(key)` → parse → use as `DbtManifest`. Wrap parse in try/catch (repo rule).
* **Memory is the real risk:** a 400MB `JSON.parse` needs multi-GB heap in the scheduler worker. Prefer a **streaming JSON parse** that extracts only the keys we use (`nodes`, `metadata`, `metrics`, `docs`) to cap memory, rather than bumping worker heap. Treat as a hard requirement, not a nice-to-have.
* **Reduce:** port the frontend `reduceManifest` (`CreatePreviewProjectModal.tsx:415` — keeps nodes/metadata/metrics/docs, drops exposures/macros/sources/etc.) into `packages/common` as a shared util and apply after fetch. Combine with the streaming parse above.
* **Errors:** object missing / no access / parse failure → fail the compile with a clear project error surfaced on the preview, **not** a silent skipped-join (the current broken-dashboard symptom).

### 5. Upload path (how the manifest lands in S3) — operational, not blocking the read path

The customer's pipeline writes the combined manifest to the agreed S3 location as part of their prod deploy. For the first customer this is a location Lightdash already has access to. Productization (later): a presigned-PUT endpoint or a documented bucket/prefix convention. The read-path wiring above is independent of how the object gets there.

## Out of scope

* Server-side combine of multiple manifests (the S3 object is already combined).
* Branch-aware UI previews that recompile the user's repo changes and merge the prod manifest — S3 previews reflect the prod manifest as-is. Fine for the dashboard-editing stakeholder use case; note it explicitly.
* **The file-upload modal change is superseded** (built earlier, but the customer rules it out at 400MB). Decision pending: revert, or keep as a small-manifest fallback.

## Affected files

* `packages/backend/src/database/migrations/<ts>_add_preview_manifest_s3_path_to_projects.ts` (new)
* `packages/backend/src/database/entities/projects.ts` — `DbProject`
* `packages/backend/src/models/ProjectModel/*` — read/write the column
* `packages/common/src/types/projects.ts` — `DbtManifestProjectConfig` + `maybeOverrideDbtConnection`
* `packages/backend/src/services/ProjectService/ProjectService.ts` — `createPreview`
* `packages/backend/src/projectAdapters/dbtManifestProjectAdapter.ts` (+ `ManifestDbtClient`) — S3 fetch + streaming parse
* `packages/common/src/**` — shared `reduceManifest` util
* `packages/backend/src/controllers/projectController.ts` — expose config (if surfacing in settings API)
* `packages/api-tests/tests/createPreviewWithManifest.test.ts` — extend for the S3 source path

## Open decisions

1. Config surface for v1: project-settings UI vs internal seeding for the pilot.
2. Streaming parse library (e.g. `stream-json`) vs heap bump — recommend streaming.
3. S3 key convention + who provisions the object.
4. Fate of the superseded file-upload modal change.

## Rollout

Only affects projects with `preview_manifest_s3_path` set, so zero impact on everyone else. Pilot with the first multi-repo customer (renewal-critical — unblocks their non-technical self-serve story).

---

*Note: this description syncs to the public GitHub issue* lightdash/lightdash#24299 *— keep it free of customer names. Account-specific context lives in the customer-success repo.*

jelstongreen:
**Stacked on #24323** (the S3 manifest source feature) — review/merge that first. This PR's diff is only the streaming change.

## What & why

#24323 reads the S3 manifest with a plain stream read, which is fine for typical manifests but has two problems at the hundreds-of-MB scale these multi-repo combined manifests reach:
- the full document is held as one string + fully-materialised object (heavy), and
- above ~512 MB, V8 can't even create the string (`Cannot create a string longer than 0x1fffffe8 characters`).

This swaps that for a **streaming reduce**: the S3 object is piped through a JSON token parser and only the keys Lightdash uses (`nodes`, `metadata`, `metrics`, `docs`) are assembled; everything else (macros, sources, parent/child maps, …) is dropped as it streams.

## How

- `ManifestDbtClient` takes a discriminated `ManifestSource` (`inline | s3`) and streams + reduces for the s3 case; inline paste keeps the simple parse path.
- Adds `stream-json@1.9.1` (pinned to the stable 1.x line; 2.x/3.x are weeks-old rewrites whose filter/assembler reduce path returns wrong output).
- Unit test streams a manifest from a mock S3 stream and asserts unused keys are dropped.

## Measured heap headroom (Node 20, default ~2 GB heap)

| Manifest | Streaming (this PR) | Naive `JSON.parse` |
|---|---|---|
| 410 MB | ✅ 554 MB RSS | ✅ 1535 MB RSS (~2.8× heavier) |
| 652 MB | ✅ 610 MB RSS | ❌ V8 max string length |

→ target ~400 MB manifest peaks ~550 MB RSS, comfortably within default worker heap; no `--max-old-space-size` bump needed.

## Note on the Socket alert
The "obfuscated code" flag on `stream-json` is a **false positive** — the source is clean, readable JS (the heuristic trips on the regex-dense tokenizer in `Parser.js`); it's a 7.5M-downloads/week package by uhop. Resolve via PR-scoped `@SocketSecurity ignore` or dashboard triage.


jelstongreen:
### Heap headroom — measured

Generated synthetic manifests and measured peak RSS (macOS `/usr/bin/time -l`, Node 20, **default** old-space heap ~2GB):

| Manifest size | Streaming reduce (this PR) | Naive `JSON.parse(readFileSync)` |
|---|---|---|
| **410 MB** | ✅ **554 MB** RSS / 488 MB heap | ✅ 1535 MB RSS / 835 MB heap (~2.8× heavier) |
| **652 MB** | ✅ **610 MB** RSS / ~766 MB heap | ❌ `Cannot create a string longer than 0x1fffffe8 characters` |

**Takeaways**
- For the target ~400 MB manifest, the streaming reduce peaks around **~550 MB RSS** — comfortably within the default worker heap, with large headroom. **No `--max-old-space-size` bump needed.**
- Above ~512 MB, naive parsing is **impossible** (V8 max string length), so streaming is required, not just leaner.

*Caveat: synthetic nodes are simpler than real dbt nodes (fewer fields), so real-world heap for the same on-disk size will be somewhat higher — but the ~2.8× advantage and the >512 MB hard wall for the naive path are structural.*

jelstongreen:
> Warning
> 
> **Review the following alerts detected in dependencies.**
> 
> According to your organization's Security Policy, it is recommended to resolve "Warn" alerts. Learn more about [Socket for GitHub](https://socket.dev?utm_medium=gh).
> 
> Action	Severity	Alert  (click "▶" to expand/collapse)
> Warn	[High](https://socket.dev/dashboard/org/lightdash/diff-scan/71727ade-81b1-412d-826c-7e93ac467e7a/alert/QAQEfVeSQJA5nBVcldHCc6JkT8ikBKildex91hLejbns)	
> **Obfuscated code**: npm `stream-json` is 90.0% likely obfuscated
> **Confidence:** 0.90
> 
> **Location:** [Package overview](https://socket.dev/npm/package/stream-json/overview/1.9.1)
> 
> **From:** [packages/backend/package.json](https://github.com/lightdash/lightdash/pull/24304/files#diff-e095a6dd10bb40447d8290e79f70c15334e874e757e25a569b9d3449ed4fe27d) → `npm/stream-json@1.9.1`
> 
> ℹ Read more on: [This package](https://socket.dev/npm/package/stream-json/overview/1.9.1) | [This alert](https://socket.dev/dashboard/org/lightdash/diff-scan/71727ade-81b1-412d-826c-7e93ac467e7a/alert/QAQEfVeSQJA5nBVcldHCc6JkT8ikBKildex91hLejbns) | [What is obfuscated code?](https://socket.dev/alerts/obfuscatedFile)
> 
> > _Next steps:_ Take a moment to review the security alert above. Review
> >   the linked package source code to understand the potential risk. Ensure the
> >   package is not malicious before proceeding. If you're unsure how to proceed,
> >   reach out to your security team or ask the Socket team for help at
> >   `support@socket.dev`.
> > 
> > _Suggestion:_ Packages should not obfuscate their code. Consider not using packages with obfuscated code.
> > 
> >     _Mark the package as acceptable risk_. To ignore this alert only
> >     in this pull request, reply with the comment
> >     `@SocketSecurity ignore npm/stream-json@1.9.1`. You can
> >     also ignore all packages with `@SocketSecurity ignore-all`.
> >     To ignore an alert for all future pull requests, use Socket's Dashboard to
> >     change the [triage state of this alert](https://socket.dev/dashboard/org/lightdash/diff-scan/71727ade-81b1-412d-826c-7e93ac467e7a/alert/QAQEfVeSQJA5nBVcldHCc6JkT8ikBKildex91hLejbns).
> 
> [View full report](https://socket.dev/dashboard/org/lightdash/diff-scan/71727ade-81b1-412d-826c-7e93ac467e7a?tab=alerts&action=error%2Cwarn)

@ZeRego is this something I can safely add an `ignore` for?

PR #24327, opened 2026-06-16
feat(ai-agents): enable content editing from chat for non-editing agents (ZAP-506 pt 1)

joaoviana:
## What & why

Follow-up to the ZAP-505 fix (which stopped non-editing agents from *suggesting* edit actions they can't perform). This PR gives the user a constructive next step instead of a dead end: when the current AI agent can't edit content but the user asked it to, an in-chat callout offers a one-click **"Enable content editing"** for admins/developers.

This is **PR 1 of 2** for ZAP-506. PR 2 (stacked) adds the second action — routing the conversation to an edit-capable agent.

## How it works

**Detection is deterministic — no extra LLM call.** The post-response suggestions endpoint already runs cheap regex turn-analysis (`detectRefusal`, `detectClarifyingQuestion`). This adds a sibling `detectContentEditBlock` that fires only when the assistant's own reply both expresses inability *and* targets editing existing content (so generic refusals — e.g. about data access — don't trigger it). The endpoint returns a single `contentEditBlocked` boolean.

**The callout decides what to show from real state only.** `AgentContentEditCallout` renders beside the suggestion chips (outside the chips' scroll-reveal wrapper, so it stays visible and clickable) when `contentEditBlocked` is true. The "Enable content editing" button shows only for users with `manage` permission on the agent. Clicking it opens a small confirm popover (enabling content editing also turns on data access), PATCHes the agent, invalidates the suggestions query, and auto-resends the user's last message so the now-capable agent performs the edit.

## Changes

- `detectContentEditBlock` regex detector + unit tests (backend).
- `contentEditBlocked: boolean` added to `ApiAgentSuggestionsResponse`, computed in `getAgentSuggestions` (returns `false` in empty-state and on early-exit paths).
- `getContentEditCalloutActions` / `hasContentEditCalloutActions` pure helper + unit tests (frontend).
- `AgentContentEditCallout` component (Enable action) + rendering in `AgentChatInput`.

## Regression analysis

- **Agents with content editing enabled:** `contentEditBlocked` is always `false` for them — no callout, no behavior change.
- **Empty-state / new threads:** flag is `false` (no assistant turn to analyze) — no callout.
- **Non-admin users with a no-edit agent:** the Enable action is gated by `manage` permission, so they see nothing in this PR (PR 2 adds the routing action for them). No dead ends.
- The detector is conservative (requires inability + content-edit-target cues in the same reply); false positives still only surface a valid, dismissible affordance.

## Testing

- Backend: `detectContentEditBlock.test.ts` (positive: overwrite tile / update saved chart / modify dashboard; negative: normal reply, data-access refusal, empty string).
- Frontend: `contentEditCallout.test.ts` (visibility matrix for the action helper).
- `pnpm -F common typecheck:fast` ✓; changed backend/frontend files type-clean; frontend lint clean.

Note: generated OpenAPI files include some unrelated cosmetic enum-reorder drift from `generate-api` (known nondeterministic codegen); the meaningful change is the new `contentEditBlocked` field.

Relates: ZAP-506

### DIFF
--- a/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
+++ b/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
@@ -409,6 +409,22 @@ function detectRefusal(text: string): boolean {
     return REFUSAL_RE.test(text);
 }
 
+// Matches the agent stating it cannot change EXISTING saved content. Requires
+// an inability cue AND an edit-verb-on-content cue so generic refusals (data
+// access, out of scope) don't fire. Deterministic sibling to detectRefusal.
+const CONTENT_EDIT_INABILITY_RE =
+    /\b(can'?t|cannot|can not|unable to|not able to|don'?t have (?:the )?(?:ability|capability)|isn'?t (?:enabled|available|supported)|not (?:enabled|available|supported))\b/i;
+const CONTENT_EDIT_TARGET_RE =
+    /\b(edit|editing|update|updating|overwrite|overwriting|modify|modifying|change|changing|replace|replacing)\b[^.?!]*\b(chart|dashboard|tile|saved|content|filter)\b/i;
+
+export function detectContentEditBlock(text: string): boolean {
+    if (!text) return false;
+    return (
+        CONTENT_EDIT_INABILITY_RE.test(text) &&
+        CONTENT_EDIT_TARGET_RE.test(text)
+    );
+}
+
 // Find the explore the agent's most recent query-producing tool call hit.
 // Returns a compact slice of its fields (labels) so the suggestion prompt can
 // stay grounded in fields the agent JUST used instead of the full catalogue.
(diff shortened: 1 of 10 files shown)


packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
```diff
@@ -409,6 +409,20 @@ function detectRefusal(text: string): boolean {
...
+const CONTENT_EDIT_TARGET_RE =
+    /\b(edit|editing|update|updating|overwrite|overwriting|modify|modifying|change|changing|replace|replacing)\b[^.?!]*\b(chart|dashboard|tile|saved|content|filter)\b/i;
+
+export function detectContentEditBlock(text: string): boolean {
+    if (!text) return false;
+    return (
+        CONTENT_EDIT_INABILITY_RE.test(text) &&
+        CONTENT_EDIT_TARGET_RE.test(text)
+    );
+}
```
joaoviana, line 424: this is just wrong

PR #24283, opened 2026-06-15
refactor(ai): carry AI custom metrics on the metric query

IrakliJani:
Stacked on #24263.

Follow-up cleanup split out of #24261 (kept off that PR deliberately). `parseVizConfig` exposed a top-level `customMetrics` sibling next to `metricQuery`, and `executeAsyncAiMetricQuery` took the full custom-metric set as a separate 5th argument — both because the AI metric query had no home for the full set (incl. `periodComparison`) alongside the aggregation-only `additionalMetrics`.

This threads it onto the metric query itself:

- `AiMetricQuery` gains `customMetrics: TransformedCustomMetric[] | null` — the **full** set used for SQL population, distinct from `additionalMetrics` (`Omit<AdditionalMetric,'sql'>[]`, the aggregation subset used for field-existence validation).
- All `AiMetricQueryWithFilters` constructors set it (the 3 viz builders, `parseVizConfig`, `runQuery`, `runSavedChart`, `runContentQuery`).
- `parseVizConfig` drops the `customMetrics` sibling; the two `AiAgentService` call sites read `parsedVizConfig.metricQuery.customMetrics`.
- `executeAsyncAiMetricQuery` drops its 5th arg and reads `metricQuery.customMetrics` (the redundant `?? additionalMetrics` fallback is gone — `populateCustomMetricsSQL` accepts `null`).

### Scope / follow-up

This unifies the **artifact re-render** path (`executeAsyncAiMetricQuery`). The **live** execution path (`runAsyncQuery` in `AiAgentToolsService`) still populates from `metricQuery.additionalMetrics` and ignores its dead 2nd `_additionalMetrics` arg. Fully collapsing custom/additional metrics into one representation means unifying those two populate paths — left as a separate effort.

PR #23229, opened 2026-05-18
feat(ai): add period-over-period to runQuery tool

joaoviana:
## Summary

Closes: ZAP-407

Adds period-over-period (PoP) comparison support to the AI `runQuery` tool. The user can now ask the agent for "revenue vs last month", "year-over-year users", "compare this quarter to the previous one", etc., and the agent has a dedicated, validated way to express it.

## Regression analysis

- **No behavior change for existing callers.**

PR #17991, opened 2025-11-10
refactor: remove unused AI visualization tool files

IrakliJani:
### Description:
Removed deprecated visualization generation tools from the AI service. This includes:
- `generateBarVizConfig.ts`
- `generateTableVizConfig.ts`
- `generateTimeSeriesVizConfig.ts`

These tools were likely replaced by a more unified approach to visualization generation in the AI service.

PR #24344, opened 2026-06-16
refactor(ai): give period-comparison metrics their own field

IrakliJani:
Follow-up to #24283. That PR added `AiMetricQuery.customMetrics` (the full custom-metric set) next to `additionalMetrics` — but the name made them look like duplicates and kept raising "are we losing period-over-period (PoP)?".

This splits them into two honest, non-overlapping fields:

```ts
additionalMetrics: Omit<AdditionalMetric, 'sql'>[];      // aggregation/real metrics — id-addressable (validation, FE, getItemId)
periodComparisonMetrics: PeriodComparisonCustomMetric[]; // PoP specs ONLY — expanded server-side by populateCustomMetricsSQL
```

Why this is the right shape:
- `additionalMetrics` is consumed by `getItemId` (FE chart renderer + backend validators), so it can only hold metrics with a `table`/`name`. A `periodComparison` spec has neither, so keeping it out of `additionalMetrics` is what prevents the `getItemId` garbage-id problem.
- PoP is **never stripped** — it lives in its own field and is concatenated back at population time: `populateCustomMetricsSQL([...additionalMetrics, ...periodComparisonMetrics])`.
- The viz builders (bar/timeseries/table) set `periodComparisonMetrics: []` — those tools were removed (#17991, Nov 2025) before PoP shipped (#23229, May 2026), so they can never carry it.

`periodComparisonMetrics` is non-empty only in `parseVizConfig`'s runQuery branch and is consumed only by the artifact re-render path (`executeAsyncAiMetricQuery`); the live `runAsyncQuery` path is untouched.

### DIFF
--- a/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
+++ b/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
@@ -1517,7 +1517,10 @@ export class AiAgentService extends BaseService {
         );
 
         const populatedCustomMetrics = populateCustomMetricsSQL(
-            metricQuery.customMetrics,
+            [
+                ...metricQuery.additionalMetrics,
+                ...metricQuery.periodComparisonMetrics,
+            ],
             explore,
         );
         const metricQueryWithCustomMetrics = {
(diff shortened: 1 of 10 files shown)


IrakliJani:
* **#24344** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24344?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24344?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24351, opened 2026-06-16
feat(formatting): render timestamp metrics in the project timezone

almeidabbm:
Closes: https://linear.app/lightdash/issue/GLITCH-498/applycustomformat-drops-the-project-timezone-for-formatted-temporal

## What

Timestamp metrics (a MIN or MAX over a timestamp column) were not being shifted into the project timezone once a display format was applied. They rendered as raw UTC like `2025-01-15T17:00:00.000Z`, while the same column shown as a dimension correctly displayed the project-timezone wall-clock like `2025-01-15, 08:00:00 (-09:00)`.

This affected the explore results table, Big Number tiles, and cartesian chart tooltips.

## Why it happened

A metric's `type` is its aggregation (`max`/`min`), not the underlying temporal type, so the existing shape check (`shouldShiftItemTimezone`) could not tell that a MAX was over a timestamp. The value also comes back from query results as an ISO string, which `applyCustomFormat` treated as non-numeric and returned untouched, dropping both the chosen format and the timezone.

## What changed

* `applyCustomFormat` now takes a `timezone` and shifts timestamp string/Date values through the temporal formatters.
* `formatItemValue` threads the timezone through the MIN/MAX path.
* The cartesian tooltip body formatter now passes the timezone (previously only the tooltip header did).
* New `getFormatterTimezone(item, value, timezone)` is the single source of truth for the "which timezone does this value shift into?" decision. It composes the existing shape predicates with a by-value fallback for type-opaque metrics, and the three formatter call sites now share it. The shape predicates stay as primitives for the value-less callers (exports, filters, pivots, sidebar).

## Flag

Gated behind `EnableTimezoneSupport`. With the flag off (no project timezone resolved) every path is byte-identical to before.

**Before**
Cartesian
before-metric-table-and-tooltip

Big number
before-bignumber

Table calcs
before-tablecalcs

**After**
Cartesian
after-metric-table-and-tooltip

Big number
after-bignumber

Table calcs
after-tablecalcs

### DIFF
--- a/packages/common/src/utils/formatting.test.ts
+++ b/packages/common/src/utils/formatting.test.ts
@@ -123,6 +123,25 @@ describe('Formatting', () => {
             ).toBeUndefined();
         });
 
+        test('STRING item carrying a timestamp value does NOT shift', () => {
+            // The by-value fallback is scoped to MIN/MAX; a known non-temporal
+            // type is trusted even when the value looks like an instant.
+            expect(
+                getFormatterTimezone(
+                    { ...dimension, type: DimensionType.STRING },
+                    tsString,
+                    tz,
+                ),
+            ).toBeUndefined();
+            expect(
+                getFormatterTimezone(
+                    { ...dimension, type: DimensionType.STRING },
+                    new Date(tsString),
+                    tz,
+                ),
+            ).toBeUndefined();
+        });
+
         test('DATE table calc does NOT shift even with a midnight-ISO value', () => {
             // The bug: a date-typed table calc whose value arrives as a full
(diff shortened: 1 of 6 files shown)


almeidabbm:
* **#24351** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24351?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24369, opened 2026-06-16
feat(ai): admin-only access control for AI agents

### LINKED ISSUE #24102: [AI Agents] Admin-only visibility mode for iterating on agents before publishing to all users
## Use case

When an admin creates a new AI agent and wants to test its accuracy and iterate on its configuration before making it available to the wider team, there is currently no way to keep it visible only to themselves. The "User Access" control only allows selecting lower-privilege users (Interactive Viewers, Viewers) to grant access to - admins and developers are always implicitly included. This means the two available states are:
- **Empty user list** = agent is accessible to all users in the project
- **Specific users listed** = agent is accessible to those users plus all admins/developers

There is no way to restrict an agent to admins/developers only while iterating, which blocks a natural development-to-production workflow.

## Proposed solution

Add an explicit visibility toggle or state to agents, such as:
- **Admin only** (default for new agents) - only admins and developers can see and use the agent
- **Everyone** - all users with project access can see and use the agent

This gives admins a safe iteration space to test and evaluate agent accuracy before rolling out to end users.

## Alternatives considered

- Adding admins/developers to the User Access picker (current design intentionally excludes them)
- Creating a separate "test" project to develop agents in (heavy workaround, not practical)

## Workaround

No workaround currently available. Once an agent is created with an empty user access list, all project users can see it.

## Additional context

This is a common request pattern from teams who want to iterate on agent quality (prompt, explores, space access) before a production rollout. The inability to do a private test-then-publish cycle is a meaningful friction point for new agent setup.

tatianainama:
Adds an admin-only access mode for AI agents, so an agent can be kept visible to admins & developers only while it's being set up — before opening it to the wider team.

Access control becomes an explicit choice: Everyone / Admins & developers only / Specific users & groups.

Closes: PROD-8221
Closes: #24102

### DIFF
--- a/packages/backend/src/ee/database/entities/aiAgent.ts
+++ b/packages/backend/src/ee/database/entities/aiAgent.ts
@@ -15,6 +15,7 @@ export type DbAiAgent = {
     enable_data_access: boolean;
     enable_self_improvement: boolean;
     enable_content_tools: boolean;
+    admin_only: boolean;
     /**
      * @deprecated Per-agent reasoning toggle was removed. The gating feature flag
      * `agent-reasoning` was never enabled so this column was effectively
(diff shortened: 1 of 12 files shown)


tatianainama:
📝 Docs: lightdash/mintlify-docs#864

PR #24370, opened 2026-06-16
feat(slack): stream agent cards without assistant scopes

joaoviana:
## Summary\n- removes assistant:write / im:history from required Slack scopes\n- keeps modern Slack streaming/cards behind the feature flag\n- makes assistant status/title/prompt APIs best-effort so missing assistant scope does not block streaming\n\n## Tests\n- pnpm -F backend typecheck --pretty false\n- pnpm -F backend test -- --runTestsByPath src/ee/services/ai/utils/getSlackBlocks.test.ts --runInBand

PR #24365, opened 2026-06-16
feat(slack): polish AI agent cards and image previews

joaoviana:
## Summary
Polishes the Slack AI-agent output cards and chart image delivery.

## Changes
- Captures chart image URLs from `runQuery`/visualization results.
- Serves Slack preview images directly from Lightdash so Slack can render them reliably with S3 enabled.
- Adds chart card actions for opening the image and exploring in Lightdash.
- Modernizes managed-agent digest Slack blocks without streaming scheduled digests.

## Validation
- Backend typecheck passed on the full stack.
- Focused Slack block tests passed on the full stack.

PR #24364, opened 2026-06-16
feat(slack): stream AI agent replies with task progress

joaoviana:
## Summary
Wires Slack AI-agent conversations to Slack streaming behind the feature flag.

## Changes
- Uses assistant status/loading messages and Slack stream chunks when enabled.
- Streams task progress from tool calls/results and writeback stages.
- Falls back to existing Slack posting behavior when streaming is unavailable or disabled.
- Adds assistant thread event handlers and feedback button action handling.
- Preserves SQL approval behavior.

## Validation
- Backend typecheck passed on the full stack.
- Focused Slack block tests passed on the full stack.

PR #24362, opened 2026-06-16
feat(slack): add modern AI agent block renderers

joaoviana:
## Summary
Adds the Slack block rendering layer for the modern AI-agent experience.

## Changes
- Adds task title mapping and task update builders.
- Adds final markdown/card/feedback renderers.
- Adds PR and artifact card builders.
- Keeps legacy project picker values within Slack limits.
- Adds focused unit coverage for task chunks, cards, feedback, markdown chunking, and picker limits.

## Validation
- `pnpm -F backend test -- --runTestsByPath src/ee/services/ai/utils/getSlackBlocks.test.ts --runInBand`
- Backend typecheck passed on the full stack.

PR #24374, opened 2026-06-16
feat(slack): modern AI agent streaming cards behind a flag

joaoviana:
Consolidates the modern Slack AI agent streaming experience into one PR, replacing the 5-PR stack (#24361, #24362, #24364, #24365, #24370). Gated behind the `AiAgentSlackModernBlocks` feature flag.

## No assistant scopes required
`assistant:write` / `im:history` are **not** requested, so existing Slack installs need no re-authorization to roll this out. `setAssistantStatus` / `setAssistantTitle` / `setSuggestedPrompts` degrade to best-effort no-ops without them; `chat.startStream` / `appendStream` / `stopStream` work on the scopes installs already have.

## What's included
- Streaming task cards (`chat.*Stream`) with a plan title + per-tool task rows
- **Early per-tool progress** — `in_progress` fires when a tool's `execute` starts, not after it finishes, so the card lights up in ~3s instead of ~28s
- Reasoning / plan-title narration between tool steps so the card isn't frozen while the model thinks
- `runSql` rendered **in-card** (the legacy bolt-gif placeholder is suppressed when the flag is on; approval buttons unchanged)
- Cleaner `runSql` result summary (`Returned N rows` instead of a raw CSV dump)
- Modern block renderers, image previews

## Regression analysis
- **Flag off** → the classic Slack path is structurally unchanged; modern code is gated at `replyToSlackPrompt`.
- Shared `generateAgentResponse` changes (early tool progress + dedup of the duplicate `in_progress` emit) also affect the **classic Slack path and web** — they surface progress *earlier* with no behavior loss.
- **Flag on** → full modern streaming card.

## Follow-ups
- Verify classic (flag-off) Slack progress timing in review.
- Reasoning task-card system is still being iterated (hence draft).

Linear: _(add ticket)_


packages/backend/src/ee/services/SlackService/SlackService.ts
```diff
@@ -28,13 +29,24 @@ export class CommercialSlackService extends SlackService {
...
+        // Bolt's current event typings lag Slack's assistant events, so keep
+        // these registrations typed loosely until the SDK exposes them.
+        (slackApp as AnyType).event('assistant_thread_started', (m: AnyType) =>
+            this.aiAgentService.handleAssistantThreadStarted(m),
+        );
+        (slackApp as AnyType).event(
+            'assistant_thread_context_changed',
+            (m: AnyType) =>
+                this.aiAgentService.handleAssistantThreadContextChanged(m),
+        );
```
joaoviana: do we need this?

packages/backend/src/ee/services/CommercialSlackIntegrationService.ts
```diff
@@ -14,15 +16,43 @@ import { CommercialSlackAuthenticationModel } from '../models/CommercialSlackAut
...
+    private async isModernSlackAiAgentEnabled(user: SessionUser) {
+        const flag = await this.featureFlagModel.get({
+            user,
+            featureFlagId: FeatureFlags.AiAgentSlackModernBlocks,
+        });
+
+        return flag.enabled;
+    }
+
+    async getSlackInstallOptions(user: SessionUser) {
```
joaoviana, line 28: is this necessary?

packages/backend/src/ee/services/ai/agents/agentV2.ts
```diff
@@ -64,6 +68,153 @@ const createAiAgentLogger =
 
 const STEP_CAP = 40;
 
+const oneLine = (value: string, maxLength = 180) => {
+    const text = value.replace(/\s+/g, ' ').trim();
+    return text.length > maxLength
+        ? `${text.slice(0, maxLength - 3)}...`
+        : text;
+};
```
joaoviana: move this and new utils in file to their own file in common?

packages/backend/src/ee/services/SlackService/SlackService.ts
joaoviana: Removed in 8cd63e38d8. These `assistant_thread_*` events only fire when the app has `assistant:write` (the assistant container), which we no longer request — so the registrations and their handlers were dead.

packages/backend/src/ee/services/CommercialSlackIntegrationService.ts
joaoviana: Removed in 8cd63e38d8 — it was never called. It used to gate the modern scopes, but scopes are now unconditional, so it (and its `featureFlagModel` dependency) are gone.

packages/backend/src/controllers/slackController.ts
```diff
@@ -262,6 +262,59 @@ export class SlackController extends BaseController {
...
+                    )}`,
+                );
+                Sentry.captureException(e, {
+                    tags: { feature: 'slack-card-image' },
+                    extra: { previewId: id },
+                });
+            }
+            return SlackController.sendPlaceholder(req.res!);
+        }
+    }
```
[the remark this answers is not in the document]
joaoviana: The safeguards in the suggestion are already in place (8cd63e38d8 context): `{id}` is validated against a 21-char nanoid regex and the response sets `Cache-Control: no-store` + `X-Robots-Tag: noindex, nofollow` (lines 277–280). Added a doc comment making the intentional-public design explicit.

packages/backend/src/ee/services/ai/agents/agentV2.ts
joaoviana: Extracted to `ai/utils/toolSummaries.ts` in 8cd63e38d8. Kept in backend rather than common — they're agent-only with no common/frontend consumer.

PR #24387, opened 2026-06-16
feat(charts): add min tick interval axis option

### LINKED ISSUE #24376: [Charts] Y-axis shows decimal tick marks when data values are small integers
## Bug description

When a bar/line chart's y-axis contains only small integer values (e.g. 0, 1, 2), ECharts auto-scales the axis with 0.5-unit tick intervals, causing duplicate integer labels (e.g. 0, 1, 1, 2, 2) to appear because the tick values are fractional but the displayed labels round to the same integer.

Users attempting to fix this via `minInterval: 1` in the chart YAML are unable to - the `Axis` type does not expose `minInterval`, so the setting is silently ignored.

## Steps to reproduce

1. Create a bar chart with a date dimension on the x-axis and a count metric on the y-axis
2. Ensure the metric values are small integers (0, 1, or 2) across most data points
3. Hover over or inspect the y-axis labels

## Expected behavior

Y-axis tick marks use integer intervals (0, 1, 2) when data values are small whole numbers.

## Actual behavior

Y-axis shows fractional tick marks (0, 0.5, 1, 1.5, 2) which display as duplicate integer labels due to label formatting.

## Additional context

ECharts supports `yAxis.minInterval` to enforce a minimum tick interval. Lightdash's `Axis` type (`packages/common/src/types/savedCharts.ts`) does not expose this option. A fix could either: (a) expose `minInterval` in the `Axis` schema, or (b) automatically detect integer-valued metrics and set `minInterval: 1` when the axis scale is small.

jpetey75:
## Problem

On bar/line charts whose value axis holds only small integers (e.g. 0, 1, 2), ECharts auto-scales the axis into 0.5 tick steps. With integer label formatting that renders as duplicate labels (0, 1, 1, 2, 2). Users couldn't fix this with `minInterval: 1` in chart YAML because the `Axis` type never exposed `minInterval`, so the setting was silently dropped.

## Changes

- **`Axis.minInterval`** — added to the `Axis` type (`packages/common/src/types/savedCharts.ts`) and applied to value-type axes (left/right Y, bottom/top X) in the ECharts config builder. Gated to value axes so it never collides with the existing time-axis auto `minInterval`.
- **UI control** — new "Min tick interval" input (`AxisMinInterval`) in the axis config panel for each numeric axis, plus `setYMinInterval`/`setXMinInterval` setters.
- **Truncate x-axis integration** — when a min tick interval is set, the Truncate x-axis (offset) bounds are snapped to multiples of the interval, so the edge ticks stay clean integers (previously e.g. `0.44999999999999996` / `3.55`).
- **Generated schemas** — regenerated OpenAPI (`swagger.json`, `routes.ts`) and the chart-as-code JSON schema so the `lightdash upload` YAML path validates `minInterval`. Scoped to the `minInterval` addition only.

## Testing

- Verified tick output with headless ECharts (SSR): `minInterval: 1` turns `0, 0.5, 1, 1.5, 2, 2.5, 3` into `0, 1, 2, 3`; truncate-snapping turns `0.4499…, 1, 2, 3, 3.55` into `0, 1, 2, 3, 4`.
- Added regression tests for the setters; full `useCartesianChartConfig` suite passes (47/47).
- `pnpm check:chart-as-code-schema` passes; frontend + common typecheck and lint clean.

## Notes

- Separate, pre-existing issue spotted while testing: end bars can clip on a numeric (value) axis because ECharts doesn't reserve band space on value axes. Verified independent of this change (the axis extent is identical with/without `minInterval`); tracked separately for its own focused fix.

Closes: PROD-8333
Closes: #24376

### DIFF
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/AxisMinInterval.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/AxisMinInterval.tsx
@@ -1,0 +1,25 @@
+import { Group, NumberInput } from '@mantine-8/core';
+import { type FC } from 'react';
+import { Config } from '../../common/Config';
+
+type Props = {
+    label: string;
+    value: number | undefined;
+    onChange: (value: number | undefined) => void;
+};
+
+export const AxisMinInterval: FC<Props> = ({ label, value, onChange }) => (
+    <Group wrap="nowrap" gap="xs">
+        <Config.Label>{label}</Config.Label>
+        <NumberInput
+            placeholder="Auto"
+            value={value ?? ''}
+            min={0}
+            step={1}
+            w={80}
+            onChange={(newValue) =>
+                onChange(typeof newValue === 'number' ? newValue : undefined)
+            }
+        />
+    </Group>
+);
--- a/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/AxisMinInterval.tsx
+++ b/packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/AxisMinInterval.tsx
@@ -18,7 +18,11 @@ export const AxisMinInterval: FC<Props> = ({ label, value, onChange }) => (
             step={1}
             w={80}
             onChange={(newValue) =>
-                onChange(typeof newValue === 'number' ? newValue : undefined)
+                onChange(
+                    typeof newValue === 'number' && newValue > 0
+                        ? newValue
+                        : undefined,
+                )
             }
         />
     </Group>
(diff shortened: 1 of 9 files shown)


packages/frontend/src/components/VisualizationConfigs/ChartConfigPanel/Axes/AxisMinInterval.tsx
```diff
@@ -0,0 +1,25 @@
...
+    onChange: (value: number | undefined) => void;
+};
+
+export const AxisMinInterval: FC<Props> = ({ label, value, onChange }) => (
+    <Group wrap="nowrap" gap="xs">
+        <Config.Label>{label}</Config.Label>
+        <NumberInput
+            placeholder="Auto"
+            value={value ?? ''}
+            min={0}
```
tatianainama, line 17: Minor / non-blocking: entering `0` persists as `minInterval: 0` and behaves like Auto, so the ticks go back to `0, 0.5, 1, 1.5, 2`. Looks like a no-op but it's still stored. Not important, just flagging in case we'd rather have `0` resolve to Auto. Tested on the preview deploy 👍

tatianainama (approved):
🚀

PR #24407, opened 2026-06-17
feat(slack): durable SQL approval (suspend/resume) for the AI agent

joaoviana:
## Problem

When the Slack AI agent proposes a SQL query that needs approval, the run **blocked a Graphile worker** polling for the decision for up to 5 minutes, then **timed out and lost the in-progress answer** if no one responded in time. The timeout existed only to stop abandoned approvals from starving the worker pool — not as a product choice.

## What this does

Replaces the blocking wait with **suspend / resume** on the modern Slack card path:

- `runSql` declares the AI SDK's native **`needsApproval`** → the agent loop halts on a `tool-approval-request` and **the worker job ends** (no worker held while waiting).
- The reply flow detects the suspended call, finalizes the task card, and posts the **approval card as its own message after it** (also a small UX win — approval used to render above the task list).
- On Approve/Reject, the button handler records the decision and **re-enqueues the reply job once**. Resume rebuilds history from the DB — now including the `tool-approval-response` — and the SDK **re-executes `runSql` itself** (or skips it on reject) and streams the answer in a fresh card below the approval.

Net effect: the approval window is now **effectively unbounded** (approve minutes or hours later — Slack buttons don't expire), and no worker is held while waiting. There is **no timeout** anymore, because a suspended run costs nothing.

## How resume works without new persistence

Everything resume needs is already stored: `sql`/`limit` in `ai_agent_tool_call.tool_args`, the decision in `ai_sql_approval`, prompt/thread via existing lookups. Because we reconstruct the message history ourselves, the SDK approval id is derived deterministically (`sql-approval:<toolCallId>`), so request/response always match — **no new table, no migration**.

## Scope / regression analysis

- **Affected:** only the **modern Slack card path** (`AiAgentSlackModernBlocks` on) for **non-auto-approved** SQL. That's the only path where `needsApproval` returns true.
- **Unchanged:**
  - **Classic Slack path** (flag off) → still the blocking `waitForSqlApproval` poll + emoji/legacy card.
  - **Web UI** → `needsApproval` is false for web prompts, so it keeps its existing blocking approval behaviour (web holds a cheap HTTP/SSE connection, not a worker — not the problem this solves).
  - **Auto-approve / "don't ask again" threads** → bypass approval entirely, as before.
- **No DB changes.** `ai_sql_approval` stays as the decision record (now read by resume instead of a poll); `waitForSqlApproval` is simply bypassed on the modern path.

## Testing

- Validated **live end-to-end** on the Jaffle dev Slack agent: suspend → worker released → approve after a delay → resume → SDK executes the SQL → answer posted. Reject and "don't ask again" also exercised.
- Unit tests for the history reconstruction (`buildToolCallTurnMessages` / `hasUnresolvedSqlApproval`) covering the resume case that must not regress (approved-but-unexecuted call → request + response, no `tool-result`), and for the suspend-detection helpers.

## Follow-ups (not in this PR)

- Web-UI parity (the reconstruction is already shared; needs a web suspend state + a resume endpoint).
- An end-to-end integration test now that the working message shape is known.

### DIFF
--- a/packages/backend/src/ee/models/AiAgentModel.ts
+++ b/packages/backend/src/ee/models/AiAgentModel.ts
@@ -5242,6 +5242,45 @@ export class AiAgentModel {
             });
     }
 
+    // A runSql call the agent suspended on awaiting approval: it has no result
+    // (never executed) and no recorded decision yet. Used to detect a suspended
+    // run and to post the approval card.
+    async getPendingSqlApprovalForPrompt(
+        promptUuid: string,
+    ): Promise<{ toolCallId: string; sql: string } | null> {
+        const row = await this.database(AiAgentToolCallTableName)
+            .leftJoin(
+                AiAgentToolResultTableName,
+                `${AiAgentToolCallTableName}.tool_call_id`,
+                `${AiAgentToolResultTableName}.tool_call_id`,
+            )
+            .leftJoin(
+                AiSqlApprovalTableName,
+                `${AiAgentToolCallTableName}.tool_call_id`,
+                `${AiSqlApprovalTableName}.tool_call_id`,
+            )
+            .where(`${AiAgentToolCallTableName}.ai_prompt_uuid`, promptUuid)
+            .where(`${AiAgentToolCallTableName}.tool_name`, 'runSql')
+            .whereNull(
+                `${AiAgentToolResultTableName}.ai_agent_tool_result_uuid`,
+            )
+            .whereNull(`${AiSqlApprovalTableName}.tool_call_id`)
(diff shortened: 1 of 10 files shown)


joaoviana:
* **#24407** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24410, opened 2026-06-17
feat: embed AI agent explore from here

rephus:
Closes: 

### Description:

Adds AI agent support for embed JWT users by introducing a `getAiAgentEmbedAccount` helper that validates whether an embedded JWT user has permission to use the AI agent (`canUseAiAgent`). If the permission is not granted, a `ForbiddenError` is thrown using the configured `aiAgentErrorMessage`. This account resolution is applied across explore, project, and share controller endpoints.

The `ShareController` is updated to use the write context user directly when creating share URLs for AI agent embed actors, replacing the previous session user approach.

The "Explore from here" flow in the AI chat is updated for embed contexts: instead of creating a share URL and opening a new tab, it now calls `onExplore` with the current chart and a `returnUrl` so the user can navigate back to the agent after exploring. The "Back to Dashboard" button label in the Explorer header is made configurable via a new `exploreBackLabel` prop, which shows "Back to agent" when navigating from the AI agent and "Back to Dashboard" otherwise.

Tests are added for the `ShareController` covering AI agent embed write actors, embed JWTs without AI agent permission, and embed JWTs without write actions.


rephus:
* **#24410** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24410?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24410?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24415, opened 2026-06-17
feat(timezone): merge DST fall-back hour on ClickHouse behind flag

almeidabbm:
## What

Makes the DST fall-back hour bucket consistently on ClickHouse by switching its timezone-wrapped truncation from the instant domain to the naive wall-clock domain, so the two folded "1 AM" hours merge into one `count = 2` bucket like the other merge adapters.

Also includes a small docs wording fix (`timezone-questions.md`) correcting the Postgres 3-arg `date_trunc` baseline to PG12 and reframing split bucketing as costlier rather than impossible.

## Why

At a DST fall-back (New York 2024-11-03), 1 AM happens twice. ClickHouse's `toTimeZone(...)` truncation kept the two instants distinct, so charts showed two `01:00` bars while Postgres, Snowflake and the rest showed one `01:00` bar with double the count. Under Lightdash's wall-clock grouping contract both folds are "1 AM" locally, so merge is the contract-consistent answer. This aligns ClickHouse with the other adapters.

## How

`dateTruncTimezoneConversions[CLICKHOUSE]` now renders the project-TZ wall-clock to a string and re-parses it as naive UTC before truncation, then re-attaches the zone in `toUTC`. It uses `formatDateTime` with `%f` and `DateTime64(3)` so millisecond-grain data keeps its precision (a second-precision format would round ms data to whole seconds).

Only the timezone-wrapped path changes, so this rides on the existing `EnableTimezoneSupport` gating. Flag off means no resolved timezone, the bare path runs, and the SQL stays byte-identical to today.

## Verification

- Unit tests in `timeFrames.test.ts` assert the naive-domain SQL shape.
- Live query check on ClickHouse and Postgres: both return one `01:00` bucket, count 2.
- Manual UI check (chart, drill, view underlying data) on ClickHouse: the merged `01:00` bar drills to exactly the two folded events.

Relates: GLITCH-509


almeidabbm:
* **#24419** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24419?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24416** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24416?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24415** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24415?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24415?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24416, opened 2026-06-17
feat(timezone): merge DST fall-back hour on BigQuery behind flag

almeidabbm:
## What

Same DST fall-back merge fix for BigQuery, stacked on the ClickHouse PR. Folds BigQuery into the generic naive truncation mechanism so the fall-back hour merges into one `count = 2` bucket.

## Why

BigQuery used 3-arg `TIMESTAMP_TRUNC(ts, part, tz)`, which truncates in the instant domain and split the fall-back into two `01:00` bars. This brings it in line with every other adapter.

## How

Three coordinated changes in `timeFrames.ts`:
1. `dateTruncTimezoneConversions[BIGQUERY]` shifts the UTC instant into a naive project-TZ `DATETIME` (`toProjectTz`) and re-attaches the zone (`toUTC`), instead of being no-ops.
2. `bigqueryConfig` truncates the wrapped path with `DATETIME_TRUNC` (the bare path keeps `TIMESTAMP_TRUNC`).
3. The BigQuery-specific `DATE(expr, tz)` day-grain branch is removed. The truncated value is now a naive `DATETIME` at tz-midnight, so the generic `CAST(... AS DATE)` reads the right calendar date.

Gated behind `EnableTimezoneSupport` via the same wrapped-path mechanism. Flag off is byte-identical.

## Verification

- Unit tests updated to the naive-domain shape (`DATETIME` / `DATETIME_TRUNC` / `TIMESTAMP`, and `CAST(... AS DATE)` for day grain).
- Live query check on BigQuery: one `01:00` bucket count 2; DAY grain returns correct calendar dates with no positive-offset off-by-one.
- Manual UI check across all six warehouses (Postgres, ClickHouse, BigQuery, Snowflake, Trino, Databricks): merged `01:00` bar with count 2 everywhere; drill and view-underlying-data return exactly the two folded events.

Relates: GLITCH-509

PR #24419, opened 2026-06-17
fix(echarts): shift sub-day time-axis to project timezone like DAY+

almeidabbm:
## What

Stacked on the BigQuery merge PR. Makes hour/minute/second time-axis charts render on the project-timezone wall-clock timeline, like DAY+ already does, so axis ticks land on real day boundaries and read as dates instead of off-midnight times.

## Why

Sub-day axes were positioned by raw UTC instants with a label-only formatter (the `sub-day-format` path). ECharts (`useUTC: true`) then placed daily ticks at UTC midnight, which in a far-from-UTC zone (e.g. New York, `00:00Z` = `19:00`/`20:00`) rendered as repeated off-midnight times with no dates, and a date label would sit ~the zone offset away from the real day boundary.

That raw-UTC carve-out existed only to avoid a DST fall-back collision: when a warehouse split the fold into two instants, shifting both to wall-clock landed two bars on the same position. The merge work earlier in this stack collapses the fall-back into one bucket per wall-clock hour on every warehouse, so the collision can no longer happen.

## How

Route sub-day grains through the same wall-clock `shift` path as DAY+ in `resolveAxisTimezone`. ECharts places ticks and renders labels natively on the wall-clock timeline:
- axis ticks fall on real project-local day boundaries and show dates
- the DST fall-back is one doubled bar
- spring-forward leaves the skipped hour empty

Deletes `formatSubDayAxisLabel`, `getSubDayTimeAxisConfig`, `SubDayTimeAxisConfig`, and the `sub-day-format` mode discriminant.

**Before**
subday-axis-BEFORE

**After**
subday-axis-AFTER

test-all

Closes: GLITCH-509


almeidabbm:
* **#24419** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24416** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24416?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24415** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24415?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #24421, opened 2026-06-17
fix(gsheets): stop re-running the query when an export fails, and surface a clear error

### LINKED ISSUE #24397: Google Sheets export: poll-timeout + job retry amplifies slow queries
## Summary

When a "Export Google Sheets" job's underlying query is slow, the job's result-polling times out at 5 minutes, the job then retries up to 3 times, and each retry re-executes the entire warehouse query from scratch. A single intermittently-slow query therefore turns into a 15–20 minute wait that frequently ends in failure. The export pipeline is healthy for fast/cached queries — the problem is the timeout-and-retry behaviour amplifying slow queries (and tripling warehouse load on queries that are already struggling).

## Symptoms

- A chart/tile GSheets export that previously completed in a few minutes can take 15–20 min.
- Intermittent — the same report sometimes exports in seconds, sometimes hangs for 15–20 min, sometimes never completes.
- Affects exports from both a standalone saved chart and a chart tile within a dashboard (same code path).
- Tends to recur on the same heavier reports.

## Directions to investigate (needs design input)

1. **Don't retry on poll-timeout** — distinguish "query still running" from genuinely transient failures (Google API 5xx / socket errors) and only retry the latter.
2. **Reattach instead of re-execute** — on retry, or when the same query is already in-flight / cached by `cache_key`, attach to the existing query/results rather than launching a fresh execution.
3. **Tune the export poll timeout** — the 5-min default may be too short for legitimately heavy exports; make it explicit/intentional rather than an inherited default.
4. **Fail fast** with an actionable message (and offer CSV fallback) once a query exceeds a threshold, instead of silently retrying for 15+ min.

## Acceptance criteria

- A slow/timed-out GSheets export no longer silently re-runs the full query 3×.
- Poll-timeout vs. transient-error retry behaviour is intentional and documented.
- The export-path poll timeout is explicit, not an inherited default.

tatianainama:
Closes: #24397

### Description

A Google Sheets export ran the warehouse query and the upload inside a single background job enqueued with `maxAttempts: 3`. Any failure re-ran the whole job — re-executing the (sometimes slow) warehouse query each time — so a slow query turned into a long sequence of repeated executions that usually still failed, while tripling warehouse load.

This PR:

- **Stops re-execution on failure** — the export job is now a single attempt (`maxAttempts: 1`). A slow/failed query is terminal rather than re-run.
- **Retries only the upload** — the Google Sheets write is retried in-process on transient Google errors, where the results are already in memory, so a retry never re-runs the query. The pivot-vs-flat upload is encapsulated in a private method so the main flow stays unaware of retries.
- **Clearer failure** — on failure the user is shown a short message ("the query took too long or failed" / "couldn't write the results to Google Sheets"). The raw error is preserved for engineers in the worker logs, Sentry, and `scheduler_log.details.error`.

### DIFF
--- a/packages/backend/src/services/SchedulerService/SchedulerService.ts
+++ b/packages/backend/src/services/SchedulerService/SchedulerService.ts
@@ -1354,7 +1354,8 @@ export class SchedulerService extends BaseService {
         }
         if (job.status === 'error') {
             throw new NotFoundError(
-                job.details?.error ?? 'Unable to download CSV',
+                job.details?.userFacingError ??
+                    'This export could not be completed.',
             );
         }
         return job;
--- a/packages/backend/src/scheduler/SchedulerClient.ts
+++ b/packages/backend/src/scheduler/SchedulerClient.ts
@@ -1139,9 +1139,7 @@ export class SchedulerClient {
             payload,
             now,
             JobPriority.HIGH,
-            // Single attempt: a job-level retry would re-run the (often slow) warehouse
-            // query. The query step is terminal on failure; only the upload is retried,
-            // in-process, by the task itself.
+            // Single attempt: a retry would re-run the slow warehouse query.
             1,
         );
 
--- a/packages/backend/src/scheduler/SchedulerClient.ts
+++ b/packages/backend/src/scheduler/SchedulerClient.ts
@@ -1139,7 +1139,10 @@ export class SchedulerClient {
             payload,
             now,
             JobPriority.HIGH,
-            3,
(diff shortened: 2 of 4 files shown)


packages/backend/src/services/SchedulerService/SchedulerService.ts
```diff
@@ -1354,7 +1354,8 @@ export class SchedulerService extends BaseService {
         }
         if (job.status === 'error') {
             throw new NotFoundError(
```
tatianainama, line 1356: why is this notFoundError 🥲

PR #24424, opened 2026-06-17
feat: gsheets sync from data apps

### LINKED ISSUE #23760: [Data Apps SDK] Expose an on-demand exportToSheets method in query-sdk
## Use case

Developers building data apps want an on-demand "Export to Google Sheets" button that creates a new spreadsheet in Google from data shown in the app. A fully client-side approach is not viable: OAuth popups are blocked by the iframe sandbox, and exposing Google API tokens in front-end app code is not acceptable.

## Proposed solution

Expose a platform-level export method in `@lightdash/query-sdk` that delegates to the Lightdash backend, e.g.:

    import { exportToSheets } from '@lightdash/query-sdk';

    await exportToSheets({
      title: 'My Export',
      sheets: [{ name: 'Sheet 1', rows: myData }],
    });

The SDK is currently a read-only query bridge (`packages/query-sdk` — `query` / `drillDown` / `useLightdash`, communicating via `createPostMessageTransport` to the parent frame). This is a new class of operation and would require:

- a new SDK method + a new postMessage message type
- a parent-frame handler (the OAuth redirect/popup must originate from the parent, not the sandboxed iframe)
- a new backend endpoint that creates and populates the spreadsheet

## Reuse + open questions

The backend Google Sheets integration already exists for scheduled deliveries: `GoogleDriveClient` (`createNewSheet`, `appendCsvToSheet`) authenticated with a per-user refresh token from `openid_identities`. `GoogleDriveClient` is reusable, but the auth model does not map cleanly to on-demand in-app export:

- Scheduled-delivery auth uses the *delivery creator's* connected Google account. An in-app export concerns the *end user viewing the app*, who may not have a Google account connected to Lightdash (or be a full Lightdash user) — the "not yet connected" path needs an on-demand OAuth flow handled by the parent frame.
- The proposed API passes client-supplied `rows`. Letting a sandboxed app POST arbitrary in-memory data to an endpoint that writes to a user's Drive is an authz/data-governance question; re-running the query server-side instead of trusting client rows is likely the safer design.

This is a scoped feature, not a passthrough of the existing method — a short design note resolving the viewer-identity/auth question should come first.

## Current workaround

Copy TSV to the clipboard via `navigator.clipboard.writeText()` with a `document.execCommand('copy')` fallback, and ask users to paste manually. The fallback is deprecated and being removed by browsers. A "Download CSV" button (unblocked once `allow-downloads` is added — see related issue) is a better near-term stopgap.

magnew:
Closes: #23760 

### Description:

Adds `exportToSheets({ title, columns, rows })` to `@lightdash/query-sdk` so data apps can offer a one-click "Open in Google Sheets" button that creates a new spreadsheet from in-memory rows.

  Flow: SDK in the sandboxed iframe → postMessage to the parent host (`useAppSdkBridge`) → parent runs the existing Google OAuth popup if needed → POSTs to a new endpoint `POST /api/v1/gdrive/upload-gsheet-from-rows` → reuses the
  existing `UPLOAD_GSHEET_FROM_QUERY` Graphile task, now dispatched on `payload.source: 'metricQuery' | 'rows'`. Returns `{ fileUrl }` for the app to open or render.

  Scope: first-party Lightdash users with a connected Google account. JWT/embed viewers are gated off via a new `capabilities.gsheetExport` flag on `useAppSdkBridge` (closed by default). Anonymous/embed export will need its own auth
  design and is deferred.

  ## Notable

  - **Sheets formula-injection sanitisation in `GoogleDriveClient.formatCell`.** Leading `=`/`@`/`\t`/`\r` get a single-quote prefix so cells stay literal. Cells were written with`valueInputOption: 'RAW'` already, so injection was a
  re-entry/re-import threat rather than immediate execution — but apps now supply arbitrary client-side row content, which materially widens the surface. The fix also closes the same hole on the existing chart-export path. Leading
  `-`/`+` are deliberately not sanitised so signed numeric strings (`-100`, `+0.5`) aren't mangled.
  - **Capability gate is closed-by-default.** `useAppSdkBridge` gained a `capabilities` option; `AppGenerate`, `AppPreviewTest`, and `DashboardDataAppTile` opt in. `EmbedApp`, `EmbedDataAppTile`, `MinimalApp` deliberately do not, and
  the SDK rejects on those surfaces with `"Google Sheets export is not available in this context"`.
  - **Body-parser ordering.** The 25 MB per-route `express.json()` is mounted *before* the global parser — Express body-parser is a no-op once `req.body` is set, so ordering matters. A row-count cap of **100,000** and a payload cap of
  **25 MB** are enforced server-side.
  - **`uploadGsheetFromQuery` was split into an outer wrapper + two inner methods** (`...FromMetricQueryInner`, `...FromRowsInner`). The metric-query branch is a verbatim move; chart-export behaviour should be unchanged, but the pivot
  path is the highest-risk regression area to spot-check.
  - **`sandboxes/data-apps/template/skill.md`** documents the new method so the in-sandbox agent stops recommending CSV copy-paste workarounds. Resumed E2B sandboxes will keep thestale skill until rebuilt — coordinate the sandbox image
   rebuild with this merge.

### DIFF
--- a/packages/backend/src/App.ts
+++ b/packages/backend/src/App.ts
@@ -9,6 +9,7 @@ import {
     MissingConfigError,
     OauthAuthenticationError,
     UnexpectedServerError,
+    UPLOAD_GSHEET_FROM_ROWS_MAX_BYTES,
 } from '@lightdash/common';
 import * as Sentry from '@sentry/node';
 import flash from 'connect-flash';
@@ -369,6 +370,14 @@ export default class App {
             );
         }
 
+        // Raise body-parser limit for the data-apps gsheet export route only.
+        // Must run BEFORE the global parser below — `express.json()` is a no-op
+        // when `req.body` is already populated, so whichever parser sees the
+        // request first wins.
+        expressApp.use(
+            '/api/v1/gdrive/upload-gsheet-from-rows',
+            express.json({ limit: UPLOAD_GSHEET_FROM_ROWS_MAX_BYTES }),
+        );
         expressApp.use(
             express.json({ limit: this.lightdashConfig.maxPayloadSize }),
         );
--- a/packages/backend/src/clients/Google/GoogleDriveClient.test.ts
+++ b/packages/backend/src/clients/Google/GoogleDriveClient.test.ts
@@ -3,6 +3,26 @@ import { GoogleDriveClient } from './GoogleDriveClient';
 
 describe('GoogleDriveClient', () => {
     describe('formatRow', () => {
+        test('should defuse spreadsheet formula injection', () => {
+            // Leading =, @, tab, and CR are unambiguous formula-injection
+            // vectors when a viewer edits and re-enters the cell. Prefix with
+            // a single quote to keep the cell literal.
+            expect(GoogleDriveClient.formatCell('=HYPERLINK("evil")')).toEqual(
+                `'=HYPERLINK("evil")`,
+            );
+            expect(GoogleDriveClient.formatCell('@SUM(A1:A10)')).toEqual(
+                "'@SUM(A1:A10)",
+            );
+            expect(GoogleDriveClient.formatCell('\t=A1')).toEqual("'\t=A1");
+            expect(GoogleDriveClient.formatCell('\r=A1')).toEqual("'\r=A1");
+            // Leading - / + are NOT sanitised — negative numeric strings like
+            // "-100" are extremely common and would otherwise be mangled.
+            expect(GoogleDriveClient.formatCell('-100')).toEqual('-100');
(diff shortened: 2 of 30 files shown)


magnew:
* **#24424** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24424?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24424?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #24253, opened 2026-06-12
Extend table row grouping ("Group repeated row values") to non-pivoted tables


pimmey:
### Summary

The **"Group repeated row values"** row-grouping for table charts (spreadsheet-style `rowSpan`-merged dimension cells, added in #23680) only renders on **pivoted** tables — i.e. when there's a field in **Columns**. On a non-pivoted (flat) table the toggle has no effect. We should extend it to non-pivoted tables.

### Current behaviour

- The row-grouping helper `getRowSpanMerges` is used only in `PivotTable` / `ExplorerPivotTable` (`packages/frontend/src/components/common/PivotTable/`).
- `SimpleTable/index.tsx` renders the pivot table (which receives `showRowGrouping`) only when `isPivotTableEnabled` — i.e. there is ≥1 pivot/Columns dimension. Otherwise it renders the flat `Table`, which is passed `showSubtotals` but **not** `showRowGrouping`.
- The config toggle is gated on `canUseSubtotals` (`≥2 un-pivoted dimensions` + subtotals off), **not** on having a pivot — so a user can enable it on a non-pivoted table, where it then does nothing. The tooltip text already anticipates the non-pivoted case (`isPivotTableEnabled ? 'un-pivoted' : ''`).

### Expected behaviour

"Group repeated row values" should also work on **non-pivoted** tables with ≥2 row dimensions: the outer dimension value(s) merge across consecutive rows via `rowSpan` (like the pivoted version) — without requiring subtotals and without requiring a Columns field.

> Note: subtotals-based grouping already works on flat tables (the flat `Table` receives `showSubtotals`). This issue is specifically about the **no-subtotals** `rowSpan`-merge view.

### Implementation sketch

- Pass `showRowGrouping` into the flat `Table` render path in `SimpleTable/index.tsx`.
- Reuse the existing pure helper `getRowSpanMerges` (`components/common/PivotTable/getRowSpanMerges.ts`) in the flat-table renderer.
- Keep the config-toggle gating in sync (this removes the current "toggle on, no effect" inconsistency on non-pivoted tables).

### Related

- #23680 — introduced the `rowSpan` row grouping for pivot tables.
- Feature flag: `pivot-row-grouping`.

PR #24435, opened 2026-06-17
feat(table): row grouping (merge repeated row values) on non-pivoted tables (PROD-8296)

pimmey:
Closes: #[24253](https://github.com/lightdash/lightdash/issues/24253) & PROD-8296

### Description:

Adds a `showRowGrouping` mode to the scrollable table that visually merges repeated dimension values across consecutive rows using `rowSpan`, matching the behaviour of the pivot table. When row grouping is enabled without subtotals, block-start cells span their full group while absorbed cells are omitted from the DOM entirely.

To support this:

- A new `getGroupedDimensionColumnIds` utility extracts dimension column IDs in display order, excluding the leaf dimension (grouping on the leaf would produce one group per row). This logic is now shared between the subtotals grouping setup in `TableHeader` and the new row-grouping body, keeping the two in sync.
- `BodyCell` accepts an optional `rowSpan` prop and applies a `mergedDimCell` CSS class to top-align and nudge the value away from the cell border when spanning multiple rows.
- Because `rowSpan` requires all spanned rows to be present in the DOM simultaneously, row-grouping mode bypasses virtualization and renders all rows eagerly.
- `getRowSpanMerges` (from the pivot table) is reused to compute per-column merge metadata, keyed by column ID and row index.

Unit tests for `getGroupedDimensionColumnIds` cover ordering by `columnOrder`, interleaved metric columns, single-dimension edge cases, and the no-dimension case.

### DIFF
--- a/packages/frontend/src/components/SimpleTable/index.tsx
+++ b/packages/frontend/src/components/SimpleTable/index.tsx
@@ -394,6 +394,7 @@ const SimpleTable: FC<SimpleTableProps> = ({
                 showColumnCalculation={showColumnCalculation}
                 showSubtotals={showSubtotals}
                 showSubtotalsExpanded={showSubtotalsExpanded}
+                showRowGrouping={showRowGrouping}
                 conditionalFormattings={conditionalFormattings}
                 minMaxMap={minMaxMap}
                 onColumnWidthChange={onColumnWidthChange}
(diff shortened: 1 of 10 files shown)


pimmey:
* **#24435** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24444, opened 2026-06-18
fix(promote): stop hiding promote button from developers when upstream exists

### LINKED ISSUE #24352: Developers are no longer able to promote dashboards
Developers are not seeing the 'promote dashboard' button after a change last week.

Initially, the below pattern was reported, but this now seems to be for orgs not using custom roles too! Check the customer conversations that are attached for further context.

When a custom role has promote:Dashboard@space (without promote:Dashboard), the Promote button is hidden in the UI.

Root cause: The UI runs two ability checks to show the button. The second check - against the upstream/destination project - passes no space access context. The @space CASL condition requires an access array with an elemMatch, so it always evaluates to false against a subject with no access field. The button stays hidden even when the user has editor access to the relevant spaces.

The backend checkPromoteDashboardPermissions correctly populates upstream space access context and would allow the promote. The bug is in the UI visibility check only.

Steps to reproduce:
1. Create a custom role with promote:Dashboard@space and promote:SavedChart@space, without promote:Dashboard and promote:SavedChart
2. Assign to a user on a project with an upstream project configured
3. Give the user editor or full access to the relevant space
4. Open a dashboard - Promote button is not visible

Expected: Promote button shown; backend enforces destination space restriction at promote time.
Actual: Promote button always hidden for @space-only roles.

tatianainama:
Closes: #24352

### Description:

The dashboard and chart headers gated the **Promote** button on a second ability check against the **upstream** (destination) project. That check passes a `subject` with no space `access` field, so the org-level promote rule for Developer/Editor — which requires `access: { $elemMatch: { role: EDITOR } }` — can never match. As a result the button is hidden whenever an upstream project exists (i.e. always, in previews). Admins were unaffected because their `manage` ability carries no `access` condition, which is why an admin could see the button on a dashboard where a developer could not.

The frontend cannot evaluate upstream promote access correctly: it has no access context for the destination space (it would need to resolve the destination space in the other project and fetch its membership). The backend already enforces the destination-space restriction at promote time, so this drops the upstream gate and shows the button based on the current-project promote check only.

The regression was introduced in #24153.

### DIFF
--- a/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
+++ b/packages/frontend/src/components/Explorer/SavedChartsHeader/index.tsx
@@ -320,24 +320,13 @@ const SavedChartsHeader: FC = () => {
             subject('SavedChart', { ...savedChart }),
         );
 
-    // Mirror the backend promote check: promoting also requires promote
-    // rights on the upstream (destination) project, so hide the action when
-    // an upstream exists but the user has no promote access there.
     const userCanPromoteChart =
         savedChart &&
         !savedChart?.dashboardUuid &&
         user.data?.ability?.can(
             'promote',
             subject('SavedChart', { ...savedChart }),
-        ) &&
-        (project?.upstreamProjectUuid === undefined ||
-            user.data?.ability?.can(
-                'promote',
-                subject('SavedChart', {
-                    organizationUuid: savedChart.organizationUuid,
-                    projectUuid: project.upstreamProjectUuid,
-                }),
-            ));
+        );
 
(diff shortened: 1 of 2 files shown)


tatianainama:
* **#24444** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24445, opened 2026-06-18
feat(common): carry temporal base type onto MIN/MAX metrics

almeidabbm:
A MIN/MAX over a DATE column or day-grain dimension is a calendar value, but its warehouse value is a midnight Date/ISO that is shape-indistinguishable from a TIMESTAMP instant. Record the underlying base type and time interval on the metric at compile time so the query builder (#24446) and formatter (#24447) can branch on it.

Only set for a MIN/MAX over a known temporal base: column metrics that aggregate their own column, and custom metrics resolved from `baseDimensionName`. Arbitrary-SQL and model-level metrics stay undefined. Inert until consumed, so no behaviour change.

### Custom `sql` on a YAML metric
- Column metric whose `sql` differs from its column → `baseDimensionType` stays undefined (a `sql` override can aggregate a different column than it is declared under), so it is left untouched. No fix, no regression.
- Column metric whose `sql` equals the column sql, or omits `sql` → inherits the type correctly.
- Model-level metric → never receives a base type → always undefined.

### Rollout
Origin of the recompile dependency for the whole stack: `baseDimensionType` for **YAML/dbt-defined** MIN/MAX metrics is baked into the compiled explore (`cached_explores`), so those metrics need a **project recompile / refresh** after deploy before the fix takes effect. **Custom (ad-hoc) metrics** compute it at query time and are picked up automatically. No flag gate here by design — nothing reads these fields unless #24446/#24447 are also active.

Relates: GLITCH-499

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -3470,6 +3470,8 @@ const models: TsoaRoute.Models = {
             distinctKeys: { dataType: 'array', array: { dataType: 'string' } },
             formatOptions: { ref: 'CustomFormat' },
             dimensionReference: { dataType: 'string' },
+            baseDimensionType: { ref: 'DimensionType' },
+            baseDimensionTimeInterval: { ref: 'TimeFrames' },
             requiredAttributes: {
                 ref: 'Record_string.string-or-string-Array_',
             },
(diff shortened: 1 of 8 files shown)


almeidabbm:
* **#24447** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24446** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24446?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24445** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24445?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24445?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24446, opened 2026-06-18
fix(timezone): tz-aware SQL for MIN/MAX over interval DATE dimensions

almeidabbm:
A custom MIN/MAX over a day-grain DATE dimension inherited the dimension's raw SQL and skipped timezone-aware compilation, so it truncated in UTC and returned a timestamp instead of the project-tz calendar date. The aggregated day could differ from the dimension column.

Re-point the aggregate at the dimension's timezone-aware `compiledSql` so the metric and its dimension agree on the day. Only fires for a MIN/MAX whose base is a truncatable interval `DATE` over a TIMESTAMP; plain DATE columns and TIMESTAMP bases are untouched, and the substring swap is a safe no-op if the base SQL does not match.

### Flag safety
Behind `useTimezoneAwareDateTrunc` (← `EnableTimezoneSupport`); first-line early return when off, so the emitted SQL is byte-identical with the flag off (test: `flag off leaves the MIN/MAX aggregate un-cast`).

### Scope / rollout
Resolves the base via the custom metric's `baseDimensionName`, so **only custom metrics** are affected — YAML/model metrics never hit this path (and don't need it: the drift is specific to a DATE-cast trunc of a TIMESTAMP). Metadata is query-time, so once deployed + flag on it is **picked up automatically, no recompile**.

Builds on #24445.

Relates: GLITCH-499


almeidabbm:
* **#24447** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24446** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24446?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24446?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24445** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24445?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Issue #24404, opened 2026-06-17
SCIM role updates fail when IdP sends PATCH Add for role changes (org role treated as additive instead of replacement)


notgiorgi:
Some SCIM identity providers (including Microsoft Entra) send role changes for existing users as a PATCH Add operation on the roles attribute. Lightdash applies this additively to the user's existing SCIM role array. When the user already has an organization role and the incoming role is a different organization role, validation fails because Lightdash enforces exactly one organization role per user.

This is distinct from the identical-duplicate case fixed in lightdash/lightdash#23037. That fix deduplicates [member, member] to [member]. It does not handle role changes like [member, viewer].

### Steps to reproduce:

1. Provision a user via SCIM with organization role 'member'.
2. Send a SCIM PATCH with op: Add, path: roles, value containing a different org role (e.g. 'viewer').
3. patchUser loads the user's existing roles from DB and converts them to a SCIM user object.
4. scim-patch applies the Add operation, appending the incoming role: [member, viewer].
5. updateUser calls validateRolesArray on the result.
6. Validation throws 400 invalidValue: 'Roles array must contain exactly one organization role, found 2'.

### Current behavior:

Existing user has org role 'member'. SCIM PATCH sends op: Add, path: roles, value 'viewer'. Lightdash builds [member, viewer]. Request fails with 'found 2'.

### Expected behavior:

For SCIM PATCH Add on roles, an incoming organization role should replace the existing org role rather than append to it. Lightdash should continue enforcing exactly one org role after normalization. Per-project role validation (one role per project UUID) should be unaffected.

#### Suggested fix:

Normalize the roles array in patchUser before passing to validateRolesArray. If the incoming patch contains an org role, remove any existing org roles before applying. If it contains a project role, replace the existing role for that project UUID. Keep identical-value dedupe as a fallback (covers the lightdash/lightdash#23037 case).

#### Impact:

Existing users cannot have their org role updated by IdPs that emit role changes as PATCH Add. Provisioning retries continue to fail until roles are manually aligned in Lightdash or the IdP is reconfigured. Affects role changes specifically - initial provisioning (POST) and identical re-assertions (PATCH Add same role) are unaffected.

PR #24455, opened 2026-06-18
fix: handle scim group patch errors

### LINKED ISSUE #24458: SCIM Group membership PATCH returns 500 "unknown error" when a member cannot be resolved
### Description

When an identity provider (e.g. Microsoft Entra / Azure AD) syncs group memberships via SCIM, a `PATCH /api/v1/scim/v2/Groups/{id}` "add member" operation can fail with a generic server error:

```
HTTP 500
{
  "schemas": ["urn:ietf:params:scim:api:messages:2.0:Error"],
  "detail": "Failed to patch SCIM group: unknown error",
  "status": "500"
}
```

This happens when the member being added cannot be resolved to an existing organization member (for example, the user is not provisioned in the org, or the IdP references a stale/unknown user id). The condition is a client-side data problem but is returned as a `500` rather than a meaningful `4xx`.

Two consequences:

1. **Wrong status code → infinite retry.** Because the IdP receives a `500`, it treats it as a transient server fault and **retries the same operation indefinitely** (observed roughly every ~20 minutes). The group membership never converges and the affected user is never added.

2. **No visibility.** The underlying error is swallowed on the `500` path — nothing is logged or reported to Sentry — so the failure is invisible on the Lightdash side, and the IdP's provisioning logs only show an opaque "unknown error" with no indication of which member is invalid.

**Impact:** the affected user is never added to the SCIM-managed group, so any access granted through that group (project/space access) is missing. Other members on the same group sync normally — only the unresolved member's `PATCH` fails, but it retries on every cycle. Sustained provisioning errors can also lead the IdP to quarantine the whole provisioning job.

The failing path is `ScimService.updateGroup`, handling `PATCH /api/v1/scim/v2/Groups/{id}`. The PUT/replace variant and other SCIM write handlers may exhibit the same behaviour and should be checked.

1. Configure SCIM provisioning from an IdP (e.g. Microsoft Entra) with group sync enabled.
2. Add a user to a synced group in the IdP where that user is **not** a resolvable organization member in Lightdash (e.g. not provisioned, or referenced by a stale id).
3. Let the IdP send the membership update — `PATCH /api/v1/scim/v2/Groups/{id}` with an "add member" operation.
4. Observe the response: `500` with `detail: "Failed to patch SCIM group: unknown error"`.
5. Observe in the IdP provisioning logs that the group `Update` operation fails and retries on every cycle, and the member is never added in Lightdash.

### version

Cloud (latest)

### Cloud or self-hosting

cloud

---

Fixed by #24455. Scope note: this covers only the group-PATCH 500 bug. The separate destructive-SCIM-deactivation behaviour (users stripped / left Pending on re-provision) is a different issue and not covered here.

notgiorgi:
### Description

Mitigates SCIM group PATCH failures by translating expected model errors into SCIM 4xx responses instead of a generic retryable 500.

This does not make invalid or out-of-order group membership updates succeed. It stops known validation failures from spiraling as unknown server errors in IdP provisioning retries.

### Testing

- `pnpm -F backend test -- ScimService.test.ts --runInBand`
- `pnpm -F backend typecheck:fast`

Relates: #24404

Closes: https://github.com/lightdash/lightdash/issues/24458

### DIFF
--- a/packages/backend/src/ee/services/ScimService/ScimService.test.ts
+++ b/packages/backend/src/ee/services/ScimService/ScimService.test.ts
@@ -1,6 +1,10 @@
 import {
+    GroupWithMembers,
     LightdashUser,
+    NotFoundError,
     OrganizationMemberRole,
+    ParameterError,
+    ScimError,
     ScimSchemaType,
     ScimUserRole,
 } from '@lightdash/common';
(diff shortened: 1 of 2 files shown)


notgiorgi:
* **#24455** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24455?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24501, opened 2026-06-19
feat(reviews): workspace entry points on the board & drawer

joaoviana:
Part of the AI-review remediation workspace stack — stacked on #24500.

Wires the board and drawer into the workspace:

- **Open workspace** entry on the kanban card and the review-item actions.
- Once a remediation exists, the actions collapse to a single **Open workspace** button — drops the redundant *View PR* / *Test fix* (the workspace owns both).
- Hides the "Preview project" card footer when a workspace is attached.
- Removes the duplicate *Test fix* row from the remediation activity timeline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24500, opened 2026-06-19
feat(reviews): paired remediation workspace UI

joaoviana:
Part of the AI-review remediation workspace stack — stacked on #24499.

Full-page workspace at `/generalSettings/ai/reviews/:fingerprint`:

- **Numbered ① Build → ② Test** flow with a connector arrow, so two chats read as one directional flow.
- **Build pane** (primary, wider): the live, interactive Build-fix thread — `editDbtProject` follow-ups adopt the same PR; SQL auto-approved; AI suggestions off.
- **Test pane** (verification rail, preview-blue accent): the read-only verification conversation + per-pane status (*Building preview / Testing / Verdict ready / Stale*) + **Retest**.
- Each pane is labelled with its **project + agent**.
- Header: root-cause badge, **View changes** (inline `@pierre/diffs` PR diff) and **View PR**.
- **project_context** collapses to a single verdict-focused column (no empty build chat).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24499, opened 2026-06-19
feat(reviews): build-fix thread + verification backend for remediation workspace

joaoviana:
Part of a 3-PR stack for the AI-review **remediation workspace** (paired Build → Test flow):

1. **this PR** — backend + types
2. #24500 — paired workspace UI
3. #24501 — board/drawer entry points

### What this PR does

Semantic-layer review remediation now opens its PR through a **real conversational agent thread** (the *Build-fix thread*) instead of the headless sandbox:

- `createReviewItemWriteback` seeds a `web_app` thread in the source project, pinned with the original failing conversation as context, and stores it on the remediation via a new nullable `work_thread_uuid` column.
- `runReviewItemWritebackJob` runs that thread via `generateAgentThreadResponse` with `editDbtProject` **force-pinned on the opening turn** (`prepareStep` toolChoice) and SQL **auto-approved**, then resolves the opened PR from `ai_writeback_thread` and continues the existing preview → compile → verification flow. It no longer creates a second preview (the tool's own preview is suppressed in this context).
- A shared `editDbtProject` **seam** emits a `pr_updated` event whenever a commit lands in a remediation's work thread (so user-driven Continue PR turns are reflected too).
- **Verdict staleness** is derived from event ordering — latest commit event newer than the latest `verification_completed` — with no PR head SHA stored (see ADR-0002).
- New **Retest** endpoint recompiles the existing preview in place and re-runs verification.
- `autoApproveSql` added to the web-app stream so workspace Continue-PR turns are frictionless.

Design decisions recorded in `docs/adr/0001` (conversational build thread) and `docs/adr/0002` (event-ordering staleness); domain glossary in `CONTEXT.md`.

### Regression analysis

- **semantic_layer remediation** now depends on the `AiWriteback` feature flag (the `editDbtProject` tool is gated on it) and a GitHub/GitLab-connected project. This is acceptable: the flag already gates all agent writeback, and `aiWritebackService.run()` itself already threw `ForbiddenError` without it — so no previously-working path regresses.
- **project_context remediation** is unchanged (still the deterministic writeback).
- **Normal agent chat is unaffected**: force-calling `editDbtProject` is gated behind a new `forceToolHints` flag set only by the review job, and `autoApproveSql` defaults to `false` everywhere else.
- Additive only: one nullable column + one new event-union member (`pr_updated`). No data migration; the activity-timeline switch has a `default` case so no exhaustiveness break.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24517, opened 2026-06-19
fix(reviews): exclude remediation build-fix threads from review classifier

joaoviana:
## Problem

The AI-review **remediation workspace** opens its PR through a real conversational **Build-fix thread** — a normal `web_app` agent thread created in the *source* project (see #24499). When the review classifier next ran, it picked up that build-fix thread as a reviewable conversation and created a **review item on the agent's own fix** — a feedback loop (review → fix → review of the fix → …).

## Root cause

`fetchBaseCandidateRows` in `AiAgentReviewClassifierModel` selects every `web_app`/`slack` agent turn with a response/error for classification. It already excludes **preview** projects — which is why the *verification* (Test, preview-mode) thread was never reviewed — but the **build-fix** thread lives in the source project, so nothing filtered it out.

## Fix

Add a correlated `NOT EXISTS` guard to the candidate query that drops any thread recorded as an `ai_agent_review_remediation.work_thread_uuid`, mirroring the existing preview-project exclusion. Build-fix threads are the reviewer's own output and must never be re-reviewed.

```sql
not exists (
  select 1 from ai_agent_review_remediation as remediation
  where remediation.work_thread_uuid = thread.ai_thread_uuid
)
```

`work_thread_uuid` is `NULL` for project_context remediations and for prompt remediations before their thread exists, so those rows never match and exclude nothing. The remediation row is written immediately after the work thread — and long before that thread produces a reviewable response — so there's no race window where an unlinked work thread could be classified.

## Regression analysis

- **Normal agent threads** (no remediation): unaffected — `NOT EXISTS` only matches threads that are a remediation's `work_thread_uuid`.
- **Verification / preview threads**: already excluded by the preview-project guard; unchanged.
- **The original failing conversation** that spawned a review: still reviewable — only the synthetic build-fix thread is excluded, not its source thread.
- Read-path query change only; no schema or data migration.

## Test

`AiAgentReviewClassifierModel.test.ts` → `listTurnReviewCandidates › excludes remediation build-fix work threads` asserts the candidate SQL references the remediation table and `work_thread_uuid`. Full model suite: 38/38 green.

---

> ⚠️ **Linear ticket**: none was provided — please add a `Closes: PROD-XXXX` line. This is a follow-up fix to the remediation workspace stack (#24499 / #24500 / #24501).

### DIFF
--- a/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
+++ b/packages/backend/src/ee/models/AiAgentReviewClassifierModel.test.ts
@@ -219,6 +219,18 @@ describe('AiAgentReviewClassifierModel', () => {
             expect(query.sql).toContain('project_type');
             expect(query.bindings).toContain(ProjectType.PREVIEW);
         });
+
+        it('excludes remediation build-fix work threads', async () => {
+            tracker.on.select(AiPromptTableName).responseOnce([]);
+
+            await model.listTurnReviewCandidates({
+                organizationUuid: ORGANIZATION_UUID,
+            });
+
+            const [query] = tracker.history.select;
+            expect(query.sql).toContain(AiAgentReviewRemediationTableName);
+            expect(query.sql).toContain('work_thread_uuid');
+        });
     });
 
     describe('remediation events', () => {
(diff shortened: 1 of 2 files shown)


joaoviana:
* **#24517** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24534, opened 2026-06-19
fix(ai-agents): deliver long Slack agent answers without `msg_too_long`

### LINKED ISSUE #24512: Slack AI agent: long responses fail with "Co-pilot failed" (Slack msg_too_long)
## Description

When the Slack AI agent produces a long response (typically from complex, multi-part questions that trigger many tool calls), the answer is generated successfully but cannot be posted back to Slack because it exceeds Slack's message size limits. Slack returns a `msg_too_long` API error, and instead of the answer the user sees the generic:

> ⚠️ Co-pilot failed to generate a response. Please try again.

The response is actually generated and saved — it's viewable in the web app for the same thread — so this is a Slack **delivery** failure, not a generation failure. The underlying error is only logged (not surfaced to the user or persisted on the message), which makes it look like the agent failed to answer.

**Expected:** long responses should be chunked/split (or gracefully truncated) so they fit within Slack's limits and the user receives the answer in Slack.

## Steps to Reproduce the Bug or Issue

1. In Slack, @mention an AI agent with a complex/long question that requires several tool calls (e.g. multiple sequential analytical sub-questions in a single message).
2. Wait for the agent to finish (can take several minutes).
3. Instead of the answer, the agent replies "Co-pilot failed to generate a response. Please try again."
4. Open the same thread in the web app — the full answer is there.

## version

cloud

## Cloud or self-hosting

cloud

tatianainama:
Closes: PROD-8382
Closes: #24512

### Description

When the Slack AI agent produced a long response, the modern streaming reply finalized the whole answer in a single `chat.stopStream` message. Long answers — and tool-heavy runs — exceeded Slack's per-message size/block limits, so Slack returned `msg_too_long`. The error was caught generically and shown as **"Co-pilot failed to generate a response"**, even though the answer was generated and persisted (visible in the web app). This is a delivery failure, not a generation failure.

There were two independent causes, both fixed:

**1. Oversized answer text**
The answer is now split into multiple thread messages (`splitMarkdownIntoMessages`), each within a per-message size/block budget. The first message finalizes the live stream, the rest post as thread replies, and the trailing artifact/feedback cards ride on the last message. If the answer is still too large for the message cap, we link to the full answer in Lightdash.

**2. Oversized task card**
Repeated same-type tool calls each appended a task block, overflowing Slack's 50-block-per-message cap. Task updates are now aggregated by tool type with a running count (e.g. **`Building chart ×11`**), and each detail/output line is streamed at most once (Slack appends deltas), keeping the card small.

A `msg_too_long` is now treated as a delivery failure — the user gets the answer (across messages, or a link to Lightdash) instead of a false "Co-pilot failed".

### DIFF
--- a/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
+++ b/packages/backend/src/ee/services/AiAgentService/AiAgentService.ts
@@ -7539,10 +7539,9 @@ Use your existing tools to inspect them when relevant to the user's question. Wh
         };
     }
 
-    // Delivers a finished answer across one or more Slack messages: the first
-    // finalizes the live stream, the rest post as replies. Falls back to a
-    // Lightdash link if the answer is too large or still rejected — it's already
-    // persisted, so a delivery failure must never read as a generation failure.
+    // Delivers a finished answer across one or more Slack messages (first
+    // finalizes the stream, rest post as replies), falling back to a Lightdash
+    // link if too large — the answer is already persisted, so this never fails.
     private async deliverModernSlackAnswer({
         slackPrompt,
         threadTs,
@@ -7586,7 +7585,7 @@ Use your existing tools to inspect them when relevant to the user's question. Wh
                 threadUrl,
             ),
         ];
-        // `text` is just the notification/fallback; keep it small.
+        // Notification fallback only; keep it small.
         const notificationText = slackifiedMarkdown.slice(0, 3000);
         const isLast = (index: number) => index === messages.length - 1;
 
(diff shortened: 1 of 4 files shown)


tatianainama:
* **#24534** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24536, opened 2026-06-19
fix(reviews): source review-item assignees from project users-with-roles (ZAP-520)

joaoviana:
## Problem

The "Assign to" picker on AI-review board cards/drawer (`ReviewAssigneeMenu`) sourced its candidate list from `useProjectAccess`, which returns only **direct** project members (`project_memberships`). Anyone whose project access is **inherited** — via a group with project access, or via an org-level role (e.g. org admins) — never appeared, so the picker was missing legitimate users.

## Fix

Switch `ReviewAssigneeMenu` to `useProjectUsersWithRoles` — the same hook the project **users & groups** table (`ProjectAccess`) uses. It returns every user who can access the project (direct + group- + org-inherited), so the picker now lists them all, with search unchanged.

## Behaviour change

The old picker also filtered to direct `ADMIN`/`DEVELOPER` members only. That filter is removed: the picker now mirrors the users & groups table and lists all project users (the v2 hook exposes only each user's *direct* role, so a meaningful "can-remediate" filter can't be derived from it without re-deriving inheritance). If we want to re-scope assignees to people who can act on a remediation, that's a follow-up — say the word and I'll add an effective-role filter.

## Regression analysis

- **Read-only change** on the frontend candidate list; no API, schema, or assignment-write behaviour changes.
- Assignment still writes `assignedToUserUuid` exactly as before; only the set of selectable users widened.
- Empty `projectUuid` path is unchanged (component returns `null`; the hook's project query is `enabled: !!projectUuid`).

## Test

`ReviewAssigneeMenu.test.tsx` opens the picker and asserts both a direct-role user **and** a user with inherited access (no direct `projectRole`) are listed — the latter is the case this fixes. Board + menu suites green.

Closes: ZAP-520

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/Admin/ReviewAssigneeMenu.test.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/Admin/ReviewAssigneeMenu.test.tsx
@@ -1,0 +1,55 @@
+import { fireEvent, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { renderWithProviders } from '../../../../../testing/testUtils';
+import { ReviewAssigneeMenu } from './ReviewAssigneeMenu';
+
+const updateMutate = vi.fn();
+
+vi.mock('../../hooks/useAiAgentAdmin', () => ({
+    useUpdateAiAgentReviewItemAssignee: () => ({
+        mutate: updateMutate,
+        isLoading: false,
+    }),
+}));
+
+// One user with a direct project role, one whose project access is inherited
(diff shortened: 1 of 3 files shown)


joaoviana:
* **#24536** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24536?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24538, opened 2026-06-19
fix(reviews): default SQL mode on in workspace build threads (ZAP-517)

joaoviana:
## Problem

In the remediation **workspace**, the Build-fix pane (`WorkspaceThreadPane`) never wired the SQL-mode toggle into `AgentChatInput`. So:

- the toggle didn't render in the build pane at all, and
- the per-thread SQL-mode state fell back to the global default (`false`),

even though the pane's submit handler always sent `enableSqlMode: true`. The control was both **hidden and misleading** — SQL mode was effectively on, but nothing showed it and the user couldn't see or change it.

## Fix

Wire the toggle into `WorkspaceThreadPane`, mirroring the normal chat (`LauncherPanel`):

- Gate visibility on `useAiAgentSqlModeAvailable` (the `manage:SqlRunner` ability), same as everywhere else.
- Default **on** for build threads via a new `selectThreadSqlModeRaw` selector (returns `undefined` when the thread has no stored toggle, so the pane applies `?? true` without changing the global default used by normal chat).
- `handleSubmit` now sends `enableSqlMode: sqlMode` (the toggle value) instead of a hard-coded `true`, so the toggle is meaningful; `autoApproveSql` and the `editDbtProject` tool pin are unchanged.

When the user lacks the SqlRunner ability the toggle stays hidden and `sqlMode` remains its `true` default — i.e. no behaviour change from today for those users.

## Verification

Verified live in the workspace: the Build pane now renders **"SQL mode on"** with the toggle pressed by default.

`aiAgentThreadModeSlice.test.ts`: `selectThreadSqlMode` still defaults `false`; `selectThreadSqlModeRaw` is `undefined` until toggled (so `raw ?? true` → on for build threads) and reflects the stored value once set.

Closes: ZAP-517

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/Admin/WorkspaceThreadPane.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/Admin/WorkspaceThreadPane.tsx
@@ -4,11 +4,20 @@ import {
 } from '@lightdash/common';
 import { Center, Loader } from '@mantine-8/core';
 import { useCallback, useMemo, type FC } from 'react';
+import { useAiAgentSqlModeAvailable } from '../../hooks/useAiAgentSqlModeAvailable';
 import { usePendingThreadRefetch } from '../../hooks/usePendingThreadRefetch';
 import {
     useAiAgentThread,
     useCreateAgentThreadMessageMutation,
 } from '../../hooks/useProjectAiAgents';
+import {
+    selectThreadSqlModeRaw,
+    setThreadSqlMode,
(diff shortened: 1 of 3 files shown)


joaoviana:
* **#24538** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24538?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24547, opened 2026-06-22
feat(reviews): filter review items by assignee (ZAP-524)

joaoviana:
## Problem

The AI-review board and findings table could be filtered by **Project** and **Cause**, but not by who a finding is assigned to. With assignees now first-class (board cards show them; you can assign from the card/drawer), there was no way to narrow the queue to "my findings" or to a specific teammate's.

## Fix

Add an **Assignee** `FilterFacet` to both the board (`ReviewKanbanBoard`) and the findings table (`AiAgentAdminReviewItemsTable`), next to the existing Project and Cause facets. Options are ordered **current user first**, then by count, then alphabetically, with a trailing **Unassigned** bucket. The shared facet logic (option building + filter predicate + the unassigned sentinel) lives in a new `reviewAssigneeFacet.ts` so the two surfaces stay in sync.

## Name resolution

Assignee display names resolve from `useOrganizationUsers` (the org-wide superset) rather than per-project members, because the board and table span multiple projects in a single view — a project-scoped hook can't label an assignee who belongs to a different project. Unresolved uuids fall back to "Unknown user".

## Regression analysis

- **Additive, read-only filter.** No API, schema, or assignment-write changes; `assignedToUserUuid` is read, never written.
- When no assignee is selected the facet is a no-op (`matchesAssigneeFilter` returns `true` for an empty selection), so default board/table contents are unchanged.
- The Unassigned bucket only renders when at least one unassigned item is in view; the current-user `(you)` suffix only when the current user has assigned items.
- Counts are computed off the same partially-filtered set the sibling facets use (post project/search), matching existing facet behaviour.

## Test

- New `reviewAssigneeFacet.test.ts` covers ordering (current user first; then count desc, name asc), the Unassigned bucket appearing only when present, the unknown-user fallback, and `matchesAssigneeFilter` (empty = pass-through, unassigned sentinel vs `null` assignee, specific uuid).
- Because the board and table now consume `useApp()` + `useOrganizationUsers()`, their existing test suites (`ReviewKanbanBoard.test.tsx`, `AiAgentAdminReviewItemsTable.test.tsx`) were updated to mock those hooks with stable values — the table's row-click test was otherwise re-rendering mid-interaction (the `AppProviderMock` user query settling) and detaching the clicked node.
- All three suites green (9/9); oxlint clean.

Closes: ZAP-524

### DIFF
--- a/packages/frontend/src/ee/features/aiCopilot/components/Admin/AiAgentAdminReviewItemsTable.test.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/Admin/AiAgentAdminReviewItemsTable.test.tsx
@@ -21,6 +21,7 @@ vi.mock('../../../../../hooks/useOnboardingMock', () => ({
 
 vi.mock('../../../../../hooks/useOrganizationUsers', () => ({
     useOrganizationUsers: () => ({ data: [] }),
+    useOrgUsersByUuid: () => new Map(),
 }));
 
 // Stable so the async health/user queries behind useApp don't re-render the
--- a/packages/frontend/src/ee/features/aiCopilot/components/Admin/AiAgentAdminReviewItemsTable.test.tsx
+++ b/packages/frontend/src/ee/features/aiCopilot/components/Admin/AiAgentAdminReviewItemsTable.test.tsx
@@ -19,6 +19,20 @@ vi.mock('../../../../../hooks/useOnboardingMock', () => ({
     useOnboardingMock: (_examples: unknown, _enabled: boolean) => undefined,
 }));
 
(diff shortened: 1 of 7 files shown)


joaoviana:
* **#24547** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24547?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24549, opened 2026-06-22
feat(reviews): sortable board with reorder + persisted order (ZAP-525)

joaoviana:
## Problem

The review board's drag-and-drop was weak: you couldn't reorder cards within a column, and the drop preview didn't show where a card would land. There was no persisted ordering at all.

## Fix

Refactor the board to **`@dnd-kit/sortable`** (the Multiple Lists pattern) and persist order end-to-end:

- **Reorder within a column** — cards are now sortable; dragging opens a live **insertion gap** showing exactly where the card will land.
- **Cross-lane moves** still change status (and kick off the deterministic writeback when a card enters *In progress*), with a live preview as the card crosses into the target lane.
- **Persisted order** — a nullable, fractional `board_position` column on `ai_agent_review_item`; `listReviewItems` sorts by it (nulls fall back to the existing last-seen order). A drop reindexes the target lane via a new `PATCH /review-items/reorder` endpoint (`reorderReviewItems` → `setReviewItemBoardPositions`, one transaction, scope resolved per fingerprint so it works on the cross-project board). The source lane keeps its relative order, so only the dropped lane is rewritten.
- **Optimistic UX, no drop flash** — on drop the dropped arrangement (lane order + cross-lane status) is written to the query cache **synchronously**, batched with clearing the transient drag order, so the board hands off to the server-derived lanes in a single commit. `onSettled` invalidates and the refetch reconciles to the authoritative order. (Previously the drag order was cleared synchronously while the optimistic cache write was deferred a microtask behind React Query v4's awaited cache-level `onMutate`, so the board painted the pre-drop order for a frame before snapping to the dropped slot.)

The previous in-progress auto-partition (active-first + divider) is removed — manual ordering replaces it.

## Backend

- Migration `20260622120000_add_board_position_to_review_items` (nullable `double precision`). `down()` is a no-op — dropping a column is a destructive, disallowed operation.
- `UpdateAiAgentReviewItemStatus` gains an optional `boardPosition`; `AiAgentReviewItem` carries `boardPosition`.
- New `reorderAiAgentReviewItems` route (declared before `/review-items/{fingerprint}` so "reorder" isn't matched as a fingerprint). Reorder writes the whole lane's `board_position` in **one transaction** so a mid-list failure can't half-reorder it. Generated spec regenerated.

## Regression analysis

- semantic_layer/project_context writeback flows are untouched; only the board's drag handling + ordering changed.
- The reorder endpoint sets **only** `board_position` (status/assignee/timestamps preserved on existing rows); it inserts a triage-status row for never-acted-on cards so they can still be ordered.
- `listReviewItems` ordering is additive — with no `board_position` set, behaviour is the previous last-seen order.

## Test

- Backend: typecheck clean; `AiAgentAdminService` + `buildReviewWritebackPrompt` suites green (32 tests); migration applied locally.
- Frontend: oxlint + tsc + oxfmt clean; `ReviewKanbanBoard.test.tsx` green.
- **DnD verified hands-on** in the preview app — reorder within a lane and across lanes lands the card in its dropped slot with no flash back to the original position.

Closes: ZAP-525

### DIFF
--- a/packages/backend/src/ee/controllers/AiAgentAdminController.ts
+++ b/packages/backend/src/ee/controllers/AiAgentAdminController.ts
@@ -11,9 +11,11 @@ import {
     ApiAiAgentSummaryResponse,
     ApiAiOrganizationSettingsResponse,
     ApiErrorPayload,
+    ApiSuccessEmpty,
     ApiUpdateAiOrganizationSettingsResponse,
     assertRegisteredAccount,
     KnexPaginateArgs,
+    ReorderAiAgentReviewItems,
     UpdateAiAgentReviewItemAssignee,
     UpdateAiAgentReviewItemStatus,
     UpdateAiOrganizationSettings,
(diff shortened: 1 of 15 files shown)


joaoviana:
* **#24549** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24549?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24549?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24552, opened 2026-06-22
fix(filters): save typed filter value on input blur

### LINKED ISSUE #3500: Filter input should "save" value when field focus changes
### Description

Is common for users to write something in the filter input and run the query without pressing enter. 
When user focus moves away from the input field it should "save" the filter value.

### To Reproduce

have to click enter

### App Version

0.287.1

ZeRego:
Closes: PROD-8401
Closes: #3500

### Description:

String filter value inputs (`FilterStringAutoComplete` and `FilterMultiStringInput`) previously only committed the typed value when the user pressed Enter or explicitly picked a dropdown item, so typing a value and clicking away silently discarded it. They now commit the in-progress search value on blur, matching the existing `TagInput` `addOnBlur` behaviour. Number and tag-based inputs already saved on focus change, so they are unchanged. Added unit tests covering blur-commit and the empty-input no-op case.

### DIFF
--- a/packages/frontend/src/components/common/Filters/FilterInputs/FilterMultiStringInput.tsx
+++ b/packages/frontend/src/components/common/Filters/FilterInputs/FilterMultiStringInput.tsx
@@ -16,6 +16,7 @@ const FilterMultiStringInput: FC<Props> = ({
     disabled,
     onChange,
     placeholder,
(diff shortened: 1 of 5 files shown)


ZeRego:
bad solution

PR #24561, opened 2026-06-22
fix(pivot): carry bin _order column through group_by_query for sort-only dims

### LINKED ISSUE #24559: Pivot query fails with "column does not exist" when a custom-bin dimension is a hidden sorted pivot column
### Description

When a custom-bin dimension is used as a pivot column that is **hidden** and **sorted**, the generated query fails at the warehouse with an error like:

```
column "amount_binned_amount_order" does not exist
```

### Root cause

A hidden, sorted pivot-column dimension is classified as a `sortOnlyDimension` (drives column `ORDER BY` but is not spread as a pivot header). Custom-bin dimensions sort by a hidden `<ref>_order` numeric companion column, so the column-ranking `ORDER BY` emits `g."<ref>_order"`.

That `_order` column is only carried through the intermediate `group_by_query` CTE for references collected by `getSortedBinDimensionReferences`, which was called with `groupByColumns` and `indexColumns` only — **not** `sortOnlyDimensions`. So for a sort-only bin dimension, `group_by_query` never selects `<ref>_order`, but the downstream `ORDER BY` still references it, producing a hard query failure.

### Steps to reproduce

1. Create a custom bin dimension (e.g. on an `amount` field).
2. Build a pivot table chart using that bin dimension as a pivot column.
3. Hide that bin-dimension column and sort by it.
4. Run the chart — the query fails at the warehouse with a `column "<ref>_order" does not exist` error.

### Fix

Include `sortOnlyDimensions` in the set passed to `getSortedBinDimensionReferences` in `PivotQueryBuilder.ts`, so the `_order` companion column is carried through `group_by_query`'s `SELECT`/`GROUP BY`. Adding `_order` to the `GROUP BY` is safe — it is functionally dependent on its bin label and does not change row cardinality.

almeidabbm:
Closes #24559

A custom-bin dimension used as a hidden, sorted pivot column lands in `sortOnlyDimensions`. Its `_order` companion column drives the column `ORDER BY`, but it was never carried through the `group_by_query` CTE, so the warehouse failed with `column "<ref>_order" does not exist`.

`getSortedBinDimensionReferences` was called with `groupByColumns` and `indexColumns` only. This includes `sortOnlyDimensions` too, so the `_order` companion survives `group_by_query`'s SELECT/GROUP BY.

Adding `_order` to the GROUP BY is safe: it is functionally dependent on its bin label and does not change row cardinality.

Added a regression test covering a sorted custom-bin dimension used as a `sortOnlyDimension`.

**Before**
CleanShot 2026-06-22 at 12 27 29

**After**
CleanShot 2026-06-22 at 12 26 08

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.test.ts
+++ b/packages/backend/src/utils/QueryBuilder/PivotQueryBuilder.test.ts
@@ -2818,6 +2818,55 @@ SELECT * FROM group_by_query LIMIT 50`);
                 'dense_rank() over (order by g."amount_binned_amount_order" asc, g."orders_category" desc)',
             );
         });
+
+        test('Should include _order column in group_by_query when sorted bin dimension is a sortOnlyDimension', () => {
+            // A custom bin dimension used as a hidden, sorted pivot column lands
+            // in sortOnlyDimensions. Its _order companion drives the column
+            // ORDER BY, so it must be carried through group_by_query — otherwise
+            // the warehouse errors with "column amount_binned_amount_order does not exist".
+            const itemsMap: ItemsMap = {
+                amount_binned_amount: binDimension,
+            };
+
+            const pivotConfiguration = {
+                indexColumn: [{ reference: 'date', type: VizIndexType.TIME }],
+                valuesColumns: [
+                    {
+                        reference: 'revenue',
+                        aggregation: VizAggregationOptions.SUM,
+                    },
+                ],
+                groupByColumns: [{ reference: 'event_type' }],
+                sortBy: [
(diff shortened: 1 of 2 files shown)


almeidabbm:
* **#24561** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24561?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24580, opened 2026-06-22
feat: change data app theme

### LINKED ISSUE #24498: [Data Apps] Allow applying or changing a theme on an existing app after creation
## Use case

When a user creates a Data App without selecting a theme (either by choice or because they lacked the permissions to view/select themes at creation time), there is currently no way to apply a theme afterwards. The app is then stuck with no branding, and the only recourse is to recreate the app from scratch - losing all prior iteration history.

## Proposed solution

Add a way to apply or change the org design (theme) on an existing Data App without requiring a full rebuild. This could be a settings panel on the app, or an option in the app editor that re-applies the selected theme to the current version.

## Alternatives considered

- Recreating the app from scratch with the correct theme selected - this works but loses all iteration history.
- The theme picker in the app creation wizard already handles initial selection well; this request extends that capability to existing apps.

## Workaround

No workaround available without losing iteration history. The only option is to delete and recreate the app.

## Additional context

Themes are currently locked at creation time in the backend (`designUuid` is only honored on initial creation; iterations always inherit from `apps.design_uuid`). A fix would need to expose a way to update `apps.design_uuid` for an existing app and re-run the theme injection into the app's source tree.

Related: the theme access permission issue is tracked at https://github.com/lightdash/lightdash/issues/23805.

magnew:
Closes: #24498 

### Description:

- Makes it possible to change the theme of a data app
- When you select a new theme, the agent gets all of the theme files and a prompt to change it
- The agent will make an attempt to change the theme without changing any of the app content
- If bigger changes need to be made, they can be prompted. The agent now has the new theme.

### DIFF
--- a/docs/data-apps.md
+++ b/docs/data-apps.md
@@ -159,6 +159,23 @@ When users send follow-up prompts, the system creates a new `DbAppVersion` and e
 
 This means Claude can see what it built previously and make targeted changes rather than starting from scratch.
 
+### Changing themes
+
+The org design (theme) can be applied, changed, or removed from an existing app from the theme chip above the prompt
+input. This creates a normal new app version, but the backend sends Claude a constrained style-only instruction instead
+of treating it like an open-ended user prompt: preserve content, queries, filters, chart semantics, layout intent, and
+interactions; only rework visual styling and theme asset usage.
+
+How the theme-change flow differs from a regular iteration:
+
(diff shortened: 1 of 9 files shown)


magnew:
* **#24580** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24580?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24629, opened 2026-06-23
fix: omit hardcoded pivot values from downloaded chart-as-code YAML (PROD-8299)

charliedowler:
## Problem

When a cartesian chart with a pivot/group-by dimension is downloaded via `lightdash download` (chart-as-code), `chartConfig.config.eChartsConfig.series` was serialized with **one entry per pivot dimension value**, each hardcoding the actual data value under `encode.yRef.pivotValues[].value` (plus data-derived `name`/`color`/computed hashes).

This made the YAML **non-portable**: uploading it to an environment with a different set of pivot values is wrong, and the series array grows with the cardinality of the pivot dimension.

## Root cause

Pivoted cartesian series are auto-generated at load time from `pivotConfig` + live query data, but the download path persisted the materialised, data-specific series verbatim.

## Fix

- Add `removePivotedSeriesValuesFromChartConfig` in `@lightdash/common`, which strips the data-level pivot values from pivoted series (using the existing `isPivotReferenceWithValues`) and dedupes the result by `getSeriesId`, so the series regenerate from `pivotConfig` on load.
- Apply it in `CoderService` when serializing `chartConfig` for download.
- Non-pivoted series and their per-series customizations are left untouched.

## Verification

- New unit tests in `savedCharts.test.ts` (5 passing): strips + dedupes pivoted series, keeps multiple metrics as separate portable series, leaves non-pivoted series/customizations untouched, passes through non-cartesian and empty/missing-series configs.
- `pnpm -F common typecheck:fast` and `pnpm -F backend typecheck:fast` clean; lint clean (no new warnings).

Fixes PROD-8299.

### DIFF
--- a/packages/backend/src/services/CoderService/CoderService.ts
+++ b/packages/backend/src/services/CoderService/CoderService.ts
@@ -31,6 +31,7 @@ import {
     ProjectType,
     PromotionAction,
     PromotionChanges,
+    removePivotedSeriesValuesFromChartConfig,
     SavedChartDAO,
     SessionUser,
     Space,
@@ -213,7 +214,9 @@ export class CoderService extends BaseService {
             tableName: chart.tableName,
             updatedAt: chart.updatedAt,
             metricQuery: chart.metricQuery,
-            chartConfig: chart.chartConfig,
+            chartConfig: removePivotedSeriesValuesFromChartConfig(
+                chart.chartConfig,
+            ),
             pivotConfig: chart.pivotConfig,
             dashboardSlug: chart.dashboardUuid
                 ? dashboardSlugs[chart.dashboardUuid]
--- a/packages/common/src/types/savedCharts.test.ts
+++ b/packages/common/src/types/savedCharts.test.ts
@@ -1,0 +1,126 @@
+import {
+    CartesianSeriesType,
+    ChartType,
+    removePivotedSeriesValuesFromChartConfig,
+    type CartesianChartConfig,
+    type ChartConfig,
+    type Series,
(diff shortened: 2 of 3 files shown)


charliedowler:
## Evidence — verified on a real chart in the running dev instance

Chart `orders-area-month` (Jaffle shop) has a pivot/group-by dimension (`orders_shipping_method`), so its stored `eChartsConfig.series` hardcodes one entry per pivot value. Comparing the **stored** config against the **chart-as-code download** after the fix:

**BEFORE — stored chart config (the bug): 3 series, each with a hardcoded pivot value**
```
series[0].encode.yRef = {"field": "orders_total_order_amount", "pivotValues": [{"field": "orders_shipping_method", "value": "standard"}]}
series[1].encode.yRef = {"field": "orders_total_order_amount", "pivotValues": [{"field": "orders_shipping_method", "value": "overnight"}]}
series[2].encode.yRef = {"field": "orders_total_order_amount", "pivotValues": [{"field": "orders_shipping_method", "value": "express"}]}
```

**AFTER — `GET /projects/{uuid}/charts/code` download (the fix): 1 portable series, no hardcoded values**
```
series count: 1
series[0].encode.yRef = {"field": "orders_total_order_amount"}
```

The downloaded YAML is now portable across environments — the per-pivot-value series (and their baked-in data values) are stripped and regenerate from `pivotConfig` on load. Unit tests covering the transform (strip + dedupe, multi-metric, non-pivoted untouched, non-cartesian/empty passthrough) also pass.

PR #24630, opened 2026-06-23
fix: keep reference line visible when hiding grouped series via legend (PROD-7119)

charliedowler:
## Problem

Reference lines (ECharts `markLine`) on grouped line/bar charts disappeared when a user hid a series via the **interactive legend** (single-click to hide one group, or double-click to isolate one). The markLine is attached to a single data series at config time, so when that series is toggled off at ECharts runtime, its markLine vanishes with it — even though the reference line relates to the whole dataset, not one group.

This is a **separate code path** from PROD-2619, which only handled series hidden via the chart Series settings panel (config-time `applyReferenceLines`). Legend toggling happens at ECharts runtime and bypasses that logic.

## Root cause

Legend selection is captured into React state (`useLegendDoubleClickSelection` → `legendselectchanged`) and fed back into `useEchartsCartesianConfig` as `validCartesianConfigLegend`, but the series→markLine attachment was never reconciled against which series are currently visible.

## Fix

Add `relocateMarkLinesToVisibleSeries` in `useEchartsCartesianConfig`: when a series carrying a markLine is hidden via the legend, re-attach its reference-line data to a still-visible series (preferring one without its own markLine, falling back to any visible series). It is applied via a `useMemo` keyed on the series and the legend selection, and **recomputes purely from the base series on every legend change**, so repeated toggling never duplicates or accumulates reference lines, and un-hiding restores the original attachment.

## Verification

- Confirmed the runtime wiring: `legendselectchanged → onLegendChange → selectedLegends` state is passed into `useEchartsCartesianConfig`, so the relocation triggers on every legend interaction.
- `pnpm -F frontend typecheck:fast` clean for the changed file (the only failures are pre-existing, unrelated `toHaveAttribute` jest-dom typings in EE aiCopilot test files). Lint + format clean via pre-commit.
- **Suggested manual check before merge:** grouped line chart (date x-axis, metric y-axis, ≥3 groups) + horizontal metric reference line → hide the first group via the legend and double-click to isolate another group; the reference line should remain visible in all states.

Fixes PROD-7119.

### DIFF
--- a/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.test.ts
+++ b/packages/frontend/src/hooks/echarts/useEchartsCartesianConfig.test.ts
@@ -23,6 +23,7 @@ import {
     getStackTotalSeries,
     mergeLegendSettings,
     padDatasetForContinuousAxis,
+    relocateMarkLinesToVisibleSeries,
     resolveCartesianGranularityLabels,
     selectContinuousDateRange,
 } from './useEchartsCartesianConfig';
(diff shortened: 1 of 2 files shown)


charliedowler:
## Evidence — deterministic unit tests for the exact bug scenarios

`relocateMarkLinesToVisibleSeries` is now exported and covered by tests that reproduce the bug at the data layer: a reference line (`markLine`) is attached to one series; the legend selection state hides that series; the test asserts the reference line **survives** by being re-attached to a still-visible series.

```
✓ relocateMarkLinesToVisibleSeries (PROD-7119) > keeps the reference line on a visible series when its host is hidden via the legend
✓ relocateMarkLinesToVisibleSeries (PROD-7119) > keeps the reference line when a single series is isolated via double-click
✓ relocateMarkLinesToVisibleSeries (PROD-7119) > leaves series untouched when the reference-line host is visible
✓ relocateMarkLinesToVisibleSeries (PROD-7119) > returns series unchanged when there is no legend selection
Tests  4 passed
```

What each asserts:
- **Single-click hide** of the markLine's host series → the host loses its markLine and the reference line data (`{ yAxis: 100 }`) is re-attached to exactly one still-visible series. (Before the fix, the reference line vanished with the hidden series.)
- **Double-click isolate** (every other series hidden) → the reference line moves onto the isolated visible series.
- **Host visible** → series returned unchanged (reference line stays put; no duplication).
- **No legend selection** → identity passthrough.

Because the transform recomputes purely from the base series on every legend change, repeated toggling never duplicates or accumulates reference lines. Full frontend suite also green (797 passed).

charliedowler:
## Reworked — the first attempt didn't actually fix the bug

Live browser testing on a real grouped chart (orders by shipping method, single axis) revealed the original fix was a **no-op for grouped charts**, so the reference line still vanished when its host group was hidden.

**Root cause:** `isSeriesVisibleInLegend` matched series against the legend selection via `serie.name`. But pivoted/grouped series have **no top-level `name`** — ECharts derives their legend name (the key in the selection map) from `encode.seriesName` → the matching dimension's `displayName`. So every grouped series was treated as "visible", no markLine was ever orphaned, and nothing relocated. The earlier unit tests passed only because they hand-built series with an explicit `name` matching the legend keys — a shape the runtime never produces.

**Fix:**
- Resolve each series' legend name from `encode.seriesName`/`dimensions` and match *that* against the legend selection.
- Only relocate **absolute-value** reference lines; **series-relative** lines (use series average / min / max / median) belong to one group and correctly hide with it (guarded, with a test).
- Reworked the unit tests to use realistic pivoted series (no top-level `name`) — these fail against the old code and pass against the fix — plus a series-average guard test.

**Verified live (Chrome DevTools, fix running via HMR):** on a grouped area chart with a fixed-value reference line —
- hide the reference line's host group via the legend → line **stays visible** ✅
- double-click-isolate a different group (hides the host) → line **stays visible** ✅

(Before the rework these both showed the line disappearing.)

Note: the dual-axis edge case mentioned earlier (relocating to a series on a differently-scaled axis) is out of scope here — this targets the reported grouped-series case.

PR #24639, opened 2026-06-23
fix: count explicitly-set falsy parameter values and defaults as supplied

almeidabbm:
## What

Make the "missing required parameter" check use **key presence** instead of value/default truthiness, and route the dashboard through the same shared helper the Explorer already uses.

- `getMissingRequiredParameters` (`@lightdash/common`): a reference is missing only when its key is absent from the values map **and** its definition has no `default` key. Previously both were truthiness checks.
- `DashboardProvider`: drops its hand-rolled copy of the predicate and calls `getMissingRequiredParameters`, so the Explorer, the dashboard, and the backend all agree on one rule.

## Why

A parameter that was explicitly set to a **falsy value** (`0`, `''`, `[]`), or declared with a **falsy default** (e.g. a `number` parameter with `default: 0`), was wrongly flagged as "missing required" and blocked the query from running — even though the backend applies a default whenever `default !== undefined` and would have happily filled it.

The logic was also duplicated: the Explorer (`useExplorerQueryManager`) and the dashboard (`DashboardProvider`) each had their own predicate, and they had drifted from each other and from the backend. This consolidates them onto one helper.

## Scope

This is a general parameter-correctness fix, independent of the date-zoom reserved-parameter feature it sits on top of. It is split into its own branch (on top of `reserved-parameters-ui`) so the feature stack stays strictly scoped to "add the date_zoom reserved parameter + shadow collisions" with no behavioral changes, and this fix is easy to review and revert on its own.

Dashboard parameter "is set" semantics are preserved exactly (a key present with an empty value still counts as set); the only behavioral change is that falsy values/defaults are no longer mistaken for "missing".

## Tests

`getMissingRequiredParameters` unit tests cover: falsy value supplied (`0`/`''`/`[]`), falsy default declared (`0`/`''`), a parameter declared with no `default` key still flagged, and reserved parameters never flagged.

### DIFF
--- a/packages/common/src/compiler/parameters.test.ts
+++ b/packages/common/src/compiler/parameters.test.ts
@@ -479,8 +479,7 @@ describe('getMissingRequiredParameters', () => {
     });
 
     it('does not flag a parameter whose default is a falsy value (0 / empty string)', () => {
-        // A default counts as supplied whenever it is declared, matching the backend's
-        // `default !== undefined`. Truthiness would wrongly require a `default: 0` param.
+        // A declared default counts as supplied (backend uses `default !== undefined`).
         const result = getMissingRequiredParameters(
             ['count', 'note'],
             {},
--- a/packages/common/src/compiler/parameters.test.ts
+++ b/packages/common/src/compiler/parameters.test.ts
@@ -455,6 +455,20 @@ describe('getMissingRequiredParameters', () => {
         expect(result).toEqual([]);
     });
 
+    it('does not flag a parameter explicitly set to a falsy value', () => {
+        // Key presence, not truthiness: a user can deliberately pick 0, '' or [].
+        const result = getMissingRequiredParameters(
+            ['amount', 'label', 'tags'],
+            { amount: 0, label: '', tags: [] },
+            {
+                amount: { label: 'Amount' },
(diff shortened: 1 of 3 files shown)


almeidabbm:
* **#24644** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24644?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24639** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24639?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24586** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24586?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24585** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24585?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24584** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24584?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #24651, opened 2026-06-24
fix(table): keep bars-in-cells proportional regardless of label width

### LINKED ISSUE #24647: [Table] Bars in cells appear shorter for values with wider numeric labels (more digits)
## Bug description

In the Table visualization with "Bars in cells" enabled, the bar width is visually incorrect when column values have different digit counts. A higher value with more digits (e.g. 1,004) can render a shorter bar than a lower value with fewer digits (e.g. 994), even though the higher value should produce a wider bar.

## Steps to reproduce

1. Create a Table visualization.
2. Enable "Bars in cells" for a numeric column.
3. Ensure the column contains values that cross a digit-count boundary (e.g. values in the 900s and 1,000s).
4. Observe that the bar for the four-digit value appears shorter than the bar for the three-digit value.

## Expected behavior

Bars should be proportionally sized based on their value relative to the column's min/max range. A higher value must always render a wider (or equal) bar than a lower value.

## Actual behavior

The bar is sized as a percentage of the remaining cell width after the formatted number label is laid out. Because the number label has `width: auto`, wider labels (more digits/characters) consume more horizontal space, leaving less room for the bar. This causes the bar for a larger value with a wider label to appear shorter than a bar for a smaller value with a narrower label.

## Root cause

`TableCellBar.tsx` uses CSS grid with `gridTemplateColumns: '1fr auto'`. The bar element occupies the `1fr` column, but `1fr` resolves to the remaining space after the `auto` (number label) column - which varies per row. The bar's percentage width is therefore relative to this variable space, not the full cell width.

```tsx
// TableCellBar.tsx
<Box style={{ display: 'grid', gridTemplateColumns: '1fr auto', gap: '4px' }}>
    <Box h="20px" w={`${percentage}%`} />  {/* percentage of 1fr, not full width */}
    <Text>{formatted}</Text>               {/* auto width - varies by digit count */}
</Box>
```

## Expected fix

Size the bar as a percentage of the full cell width, independent of the number label width. Absolute positioning within a relative container would achieve this.

## Environment

- Lightdash version: current main
- Warehouse: any
- Browser: any

## Additional context

Related to (but distinct from) closed issue #17802, which described same-length bars for differing values.

jelstongreen:
## Problem

In a Table viz with **Bars in cells** enabled, a higher value could render a *shorter* bar than a lower value when the two labels have different digit counts (e.g. `1,002` vs `998`).

`TableCellBar` used a `1fr auto` CSS grid: the bar sat in the `1fr` track, the number label in `auto`. `1fr` resolves to whatever space is left *after* the label, so a wider label (more digits) shrank the bar track for that row. The bar's percentage was therefore relative to a per-row track, not a constant one.

## Fix

Reserve a **constant label gutter** sized to the column's widest label, and size every bar against the same remaining track.

- The widest formatted label per column is computed once in `TableProvider` (single pass over the already-loaded rows, gated on `minMaxMap` so it's free when no column uses bars) and threaded through table `meta` as `barLabelMaxMap`.
- `TableCellBar` holds that label in an invisible, `aria-hidden` sizer so the gutter is identical on every row. The sizer is only rendered when the row isn't already the widest, so the DOM isn't duplicated for the common case (and there's no overlap of the label onto the bar).

Bar geometry is now purely `(value - min) / range` against a fixed track, so a higher value always renders a wider-or-equal bar.

### Before → After

With a column of `300, 998, 1,002, 1,700`:

- **Before:** `998` rendered a *longer* bar than `1,002` — the wider `1,002` label stole track width.
- **After:** monotonic bars (`300 < 998 ≈ 1,002 < 1,700`); close values correctly produce near-equal bars; labels right-aligned in the reserved gutter with no overlap.

## Testing

- `pnpm -F frontend exec vitest run src/hooks/useColumns.test.tsx` — 24 passed, incl. two new cases covering the reserved gutter (invisible sizer holds the widest label) and the no-duplication path when the row is the widest.
- Lint clean on all changed files.

Closes: PROD-8457
Closes: #24647

### DIFF
--- a/packages/frontend/src/components/common/Table/TableProvider.tsx
+++ b/packages/frontend/src/components/common/Table/TableProvider.tsx
@@ -177,6 +177,26 @@ export const TableProvider: FC<React.PropsWithChildren<ProviderProps>> = ({
         return data.slice(start, end);
     }, [data, paginationState]);
 
+    // Widest formatted label per column. "Bars in cells" reserves a label
+    // gutter this wide so every row's bar is sized against the same remaining
+    // track — otherwise wider labels (more digits) shrink the bar (PROD-8457).
+    const barLabelMaxMap = useMemo(() => {
+        if (!minMaxMap) return undefined;
+        const result: Record<string, string> = {};
+        for (const row of data) {
+            for (const columnId of Object.keys(row)) {
+                const formatted = row[columnId]?.value?.formatted;
+                if (
+                    typeof formatted === 'string' &&
+                    formatted.length > (result[columnId]?.length ?? 0)
+                ) {
+                    result[columnId] = formatted;
+                }
+            }
+        }
+        return result;
+    }, [data, minMaxMap]);
+
     const table = useReactTable({
         data: isInfiniteScrollEnabled ? data : pageRows,
         columns: visibleColumns,
@@ -197,6 +217,7 @@ export const TableProvider: FC<React.PropsWithChildren<ProviderProps>> = ({
         meta: {
             columnProperties,
             minMaxMap,
+            barLabelMaxMap,
         },
         enableColumnPinning: true,
         onColumnVisibilityChange: setColumnVisibility,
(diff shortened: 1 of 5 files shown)


jelstongreen:
image

PR #24668, opened 2026-06-24
fix(map): make choropleth tooltip follow the cursor

### LINKED ISSUE #24663: [Map Chart] Choropleth tooltip renders off-screen for countries with overseas territories (France, etc.)
## Bug description

On a choropleth (area) map chart, hovering over the French mainland shows the tooltip anchored to a point far from the hovered location - the tooltip appears near South America (where French Guiana is) rather than following the cursor. When the map is zoomed into a regional view (e.g. Europe), the tooltip is completely off-screen and invisible.

This affects any country whose GeoJSON feature is a multi-polygon spanning a large geographic area due to overseas territories (France, USA with Hawaii/Alaska, etc.).

## Steps to reproduce

1. Create a map chart using the choropleth/area type
2. Use a dataset with French territory data (ISO code 'fra' or similar)
3. Hover over the French mainland on the map
4. Observe the tooltip appears far from the cursor - anchored to a point near South America
5. Zoom into Europe — the tooltip is no longer visible at all

## Expected behavior

The tooltip should appear near the mouse cursor when hovering over any region on the map, regardless of the geographic extent of the underlying GeoJSON feature.

## Actual behavior

The tooltip is anchored to Leaflet's default tooltip position for the layer, which is computed from the bounding box centroid of the entire GeoJSON feature. For France (which includes French Guiana, Réunion, Guadeloupe etc. as a multi-polygon), this centroid is far from mainland Europe - causing the tooltip to render off-screen when zoomed into a European view.

## Root cause

In `packages/frontend/src/components/SimpleMap/index.tsx`, `layer.bindTooltip(tooltipHtml)` is called without the `sticky: true` option. Adding `{ sticky: true }` makes the tooltip follow the mouse cursor position instead of anchoring to a fixed point derived from the feature geometry.

## Proposed fix

Change line ~914 in `SimpleMap/index.tsx`:

```js
// Before
layer.bindTooltip(tooltipHtml);

// After
layer.bindTooltip(tooltipHtml, { sticky: true });
```

## Environment

- Chart type: Map (choropleth/area mode)
- Affected countries: Any with overseas territories that inflate the GeoJSON bounding box (e.g. France, USA)

## Additional context

Screenshots confirm: hovering mainland France shows the tooltip positioned near South America. When zoomed into Europe, the tooltip is fully off-screen.

jelstongreen:
## Problem

On a choropleth (area) map chart, hovering a country whose GeoJSON feature is a multi-polygon spanning a large area (overseas territories — France, USA, etc.) shows the tooltip anchored far from the cursor. Leaflet anchors the tooltip to the feature's centroid; for France (mainland + French Guiana, Réunion, Guadeloupe…) that centroid sits in the Atlantic, so the tooltip renders well below the mainland — and fully off-screen when zoomed into a regional view like Europe.

## Fix

Bind the tooltip with `{ sticky: true }` so it tracks the cursor instead of anchoring to the geometry-derived point.

```diff
-                    layer.bindTooltip(tooltipHtml);
+                    layer.bindTooltip(tooltipHtml, { sticky: true });
```

## Testing

Verified locally on a World choropleth (`world_country_metrics`, colored by revenue):

- The bound tooltip now reports `sticky: true`.
- France's feature bounds span lat −21°→51°, lng −62°→56° (a 21-polygon multi-polygon). The default centroid anchor is ~117px below mainland France even in world view, and leaves the viewport entirely when zoomed to Europe.
- With the fix, hovering mainland France shows the tooltip at the cursor.

Closes: PROD-8468
Closes: #24663

### DIFF
--- a/packages/frontend/src/components/SimpleMap/index.tsx
+++ b/packages/frontend/src/components/SimpleMap/index.tsx
@@ -911,7 +911,7 @@ const SimpleMap: FC<SimpleMapProps> = memo(
                 );
 
                 if (layer instanceof L.Path) {
-                    layer.bindTooltip(tooltipHtml);
+                    layer.bindTooltip(tooltipHtml, { sticky: true });
 
                     // Determine the correct base and hover opacity for this region
                     const baseOpacity = regionEntry


jelstongreen:
image

PR #24723, opened 2026-06-25
fix: share uploaded images to the target Slack channel

### LINKED ISSUE #24719: [Slack] Slack notifications for IMAGE format scheduled deliveries and alerts never arrive
### Description

When a scheduled delivery or threshold alert is configured with IMAGE format and a Slack target, no message arrives in the channel at all. Not just a missing image but zero message.

The root cause might be that `SlackClient.uploadFile()` calls `files.uploadV2` without `channel_id`, so the file is never shared to any channel. The `url_private` of that unshared file then gets used as `image_url` in a Slack Block Kit image block. Slack rejects it with `invalid_blocks`, `postMessage` re-throws, and the whole notification fails silently.

CSV, XLSX, and PDF formats work fine.

1. Set up a Slack integration
2. Create a scheduled delivery or threshold alert using IMAGE format with a Slack target
3. Trigger the delivery via Send Now or wait for the scheduled time
4. Check the Slack channel and observe that no message arrives

### version

v0.3240.0

### Cloud or self-hosting

self-hosting

Abdul-Moiz31:
## Summary

* Fixes lightdash/lightdash#24719 — Slack notifications for IMAGE format scheduled deliveries and alerts never arrive
* `SlackClient.uploadFile()` called `files.uploadV2` without `channel_id`, so the uploaded image was never shared to any channel. Its `url_private` was then used as `image_url` in a Block Kit image block, which Slack rejects with `invalid_blocks` — causing `postMessage` to throw and the whole notification to fail silently (not just the image).
* `uploadFile()` now resolves and shares the file to the same channel the notification will be posted to, reusing the channel/DM resolution logic already used by `postFileToThread` (extracted into a shared `resolveChannelIdForFileUpload` helper).
* Threaded the channel through `tryUploadingImageToSlack` and its two call sites in `SchedulerTask` (threshold alerts and IMAGE-format scheduled deliveries).

## Test plan

- [ ] Configure a Slack integration and create a scheduled delivery with IMAGE format
- [ ] Trigger via "Send Now" and confirm the message (with image) arrives in the channel
- [ ] Repeat for a threshold alert with IMAGE format
- [ ] Confirm CSV/XLSX/PDF deliveries still work unaffected
- [ ] Confirm DM-target deliveries (user channel) still work for IMAGE format


tatianainama:
https://github.com/lightdash/lightdash/pull/24844#issuecomment-5411690992

PR #24724, opened 2026-06-25
fix: model-level metrics inherit required_attributes from referenced dimensions

### LINKED ISSUE #24709: required_attributes on dimensions not enforced in derived metrics with native Lightdash YAML format
### Description

When using the native Lightdash YAML format, required_attributes is respected on the dimension but silently ignored on derived metrics. Users without the required attribute can still see and query protected metrics. The same configuration works correctly in the dbt meta YAML format.

1. Create a native Lightdash YAML file with a protected dimension and a metric with required_attributes:

```
type: model
name: my_model
sql_from: "my_project.my_dataset.my_table"

dimensions:
  - name: sensitive_field
    sql: ${TABLE}.sensitive_field
    type: number
    label: Sensitive Field
    required_attributes:
      access_groups: "data"

metrics:
  sensitive_metric:
    label: Sensitive Metric
    type: sum
    sql: ${sensitive_field}
```

2. Create a users without access_groups = data

3. Log in as the user without the attribute and open the explore — sensitive_metric is visible and queryable.

The same protection applied via dbt meta YAML does work as expected, both the dimension and metric are correctly hidden from users without the attribute.

```
columns:
  - name: sensitive_field
    meta:
      dimension:
        required_attributes:
          access_groups: "data"
      metrics:
        sensitive_metric:
          type: sum
          label: Sensitive Metric
```

### version

v0.2761.1

### Cloud or self-hosting

self-hosting

Abdul-Moiz31:
## Summary

* Fixes lightdash/lightdash#24709 — `required_attributes` on dimensions not enforced in derived metrics with native Lightdash YAML format
* Metrics defined in a model's top-level `metrics:` block (native Lightdash YAML, and equivalently dbt's model-level `meta.metrics`) never received `requiredAttributes` during conversion — only metrics nested under a column's `meta.dimension` did. So a metric like `sql: ${sensitive_field}` referencing a protected dimension was visible/queryable to users without the required attribute, even though the dimension itself was correctly hidden.
* `convertTable` now scans a model-level metric's `sql` for `${dimension}` references within the same table and merges in the `required_attributes`/`any_attributes` of any protected dimension it touches, mirroring the inheritance column-nested metrics already get.

## Test plan

- [X] Added a unit test in `translator.test.ts` covering a model with a protected dimension and a model-level metric referencing it via `sql: ${dimension}` — verifies the metric inherits `requiredAttributes`
- [ ] Manually reproduce the issue's repro steps (native YAML model + user without `access_groups: data`) and confirm `sensitive_metric` is now hidden/blocked
- [ ] Confirm dbt-meta column-nested metrics still work as before (no regression)


tatianainama:
https://github.com/lightdash/lightdash/pull/24844#issuecomment-5411690992

PR #24522, opened 2026-06-19
fix: make browserless launch Chrome with the right window size

magnew:
### Description:
Fix data app scheduled delivery screenshots by launching Browserless app captures with a 1400x4000 Chrome window.

This avoids the Browserless/CDP 800x600 default capture surface, which setViewportSize was not reliably overriding for app screenshots.

PR #24436, opened 2026-06-17
fix: screenshot size sync for data apps delivery

magnew:
Fix data app scheduled delivery screenshots by explicitly applying the app viewport after creating the Browserless/CDP page.

newPage({ viewport }) can be ignored over CDP, leaving app screenshots at the default 800x600. Calling setViewportSize before navigation restores the intended 1400x4000 viewport and prevents clipped captures.

PR #24284, opened 2026-06-15
fix: viewport animation race in data apps deliveries

magnew:
### Description:

Data apps scheduled deliveries were re-sizing the viewport, then immediately taking a snapshot, resulting in a re-render race condition. This give some time for the re-render.

PR #24733, opened 2026-06-25
fix: bypass playwright for app viewport height

magnew:
## Problem

Data app scheduled deliveries arrive cropped — only the top-left ~800×600 of a
much taller intended render shows in the delivered image.

Diagnostic logging on an APP delivery confirmed the root cause: Playwright
reports the configured viewport (1400×4000), but Chrome's real surface stays at
800×600 (`window.innerWidth/innerHeight = 800/600`), so content lays out at 800
wide and gets clipped.

The reason is `connectOverCDP`: Playwright connects with
`noDefaultViewport: true`, so for pages in the (Browserless-default) context it
deliberately **never sends** `Emulation.setDeviceMetricsOverride`. Every prior
attempt routed through this same gated path — `setViewportSize` (#24284,
#24436, #24459) and the `--window-size` launch arg (#24522) — which is why none
of them took effect.

## Fix

For `LightdashPage.APP` only, after `newPage()` and before navigation, send
`Emulation.setDeviceMetricsOverride` directly on a raw `CDPSession`. This
bypasses Playwright's viewport plumbing and reaches Chrome's renderer directly,
applying the override Playwright skips.

- **APP-scoped**: entire block guarded by `lightdashPage === LightdashPage.APP`;
  non-APP delivery paths are unchanged.
- **Fully isolated**: wrapped in try/catch that only logs a warning and never
  rethrows — on failure we fall through to existing behavior, no worse than
  today.
- The prior `--window-size` arg and the `[APP-DIAG]` logging are left in place
  so we can verify the fix in production (`parent.innerW` should now read 1400).

### DIFF
--- a/packages/backend/src/services/UnfurlService/UnfurlService.ts
+++ b/packages/backend/src/services/UnfurlService/UnfurlService.ts
@@ -1102,6 +1102,33 @@ export class UnfurlService extends BaseService {
                                 .internalLightdashHostIgnoreHttpsErrors,
                     });
 
+                    if (lightdashPage === LightdashPage.APP) {
+                        // Browserless drops Playwright's viewport over CDP (see [APP-DIAG]
+                        // logs stuck at 800×600); push the override straight to Chrome.
+                        try {
+                            const cdp = await page
+                                .context()
+                                .newCDPSession(page);
+                            await cdp.send(
+                                'Emulation.setDeviceMetricsOverride',
+                                {
+                                    width: initialViewport.width,
+                                    height: initialViewport.height,
+                                    deviceScaleFactor: 1,
+                                    mobile: false,
+                                },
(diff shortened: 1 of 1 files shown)


magnew:
* **#24736** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24736?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24733** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24740, opened 2026-06-25
feat: add app uuid to query tags for data app queries

### LINKED ISSUE #24693: Tag data app warehouse queries with an app id so that they can be attributed to the app
### Description

Tag data app warehouse queries with an app id so that they can be attributed to the app. 

Orgs could track which warehouse queries came from which app.

magnew:
Closes: #24693 

### Description:

Data-app queries are tagged in the warehouse with their originating
`app_uuid`. This moves that attribution off the `executeAsyncMetricQuery`
service args and onto an ambient request header (`Lightdash-App-Uuid`),
mirroring how scheduler tags already flow through the request-scoped
ExecutionContext.

- Frontend SDK bridge sends the header instead of a `source` body field
- Middleware stamps it into ExecutionContext; query tagging reads it there
- Reverts the `source` plumbing through QueryController + the args type

The id is self-reported and unvalidated — tracking metadata only, not a
security boundary. As a bonus it now also stamps app traffic log lines.

### DIFF
--- a/packages/backend/src/ee/services/EmbedService/EmbedService.ts
+++ b/packages/backend/src/ee/services/EmbedService/EmbedService.ts
@@ -998,6 +998,7 @@ export class EmbedService extends BaseService {
         queryTags: Omit<
             Required<RunQueryTags>,
             | 'user_uuid'
+            | 'app_uuid'
             | 'chart_uuid'
             | 'dashboard_uuid'
             // Scheduler-attribution tags only apply to scheduler-driven jobs,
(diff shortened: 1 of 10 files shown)


magnew:
* **#24740** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24741, opened 2026-06-25
fix(sdk): make embedded dashboards render under a strict CSP

### LINKED ISSUE #21276: CSP-compatible SDK build
### Description

Root cause: 3 of 4 problematic libraries (AJV, PEG.js, Vega) are not even needed by embedded dashboards —

   they're backend/CLI utilities bundled because @lightdash/common lacks sideEffects: false and re-exports 
  everything. The 4th (ECharts) has only a dead-code new Function fallback that never executes in modern   
  browsers.                                                                                                
                                                                                                           
  The plan (3 phases):                                                                                     
                                                                                                           
  1. Tree-shake AJV & PEG.js — Add sideEffects to common's package.json so Vite eliminates unused          
  backend/CLI code from the SDK bundle. Fallback: add explicit externals in SDK build config.
  2. Externalize Vega — Mark vega packages as external in the SDK build. Custom visualizations are a       
  power-user feature not relevant to standard embedded charts. Make the static vega-themes import dynamic  
  to avoid import failure.
  3. Patch ECharts — A Vite transform plugin replaces the dead new Function fallback in GeoJSONResource.js 
  with just JSON.parse(source) during SDK build only.                                                      
  
  Total files changed: 3 (common/package.json, vite.config.sdk.ts, CustomVisualization/index.tsx). No      
  library upgrades needed. Main frontend app build is unaffected.


---

CSP-Compatible SDK Build                                                                                  
                                                        
 Context

 A customer with strict CSP (no unsafe-eval) cannot use @lightdash/sdk because four bundled libraries use eval()/new Function() at runtime. Their CSP blocks these calls, preventing the  SDK from rendering dashboards. They're falling back to iframe embedding but prefer the React SDK for tighter integration.

 Key insight: 3 of 4 libraries (AJV, PEG.js, Vega) are not needed by embedded dashboards but are bundled
 because @lightdash/common lacks sideEffects: false and re-exports everything via export *.

 Plan

 Phase 1: Tree-shake AJV and PEG.js out of SDK bundle

 These libraries are only used by backend/CLI code paths but get pulled into the SDK because common
 re-exports them and doesn't declare sideEffects: false.

 1a. Add sideEffects to common's package.json

 File: packages/common/package.json

 "sideEffects": false

 Caveat: packages/common/src/index.ts has dayjs.extend(utc) at line 44, which IS a side effect. Options:
 - Use "sideEffects": ["./src/index.ts", "./dist/**/*"] to keep the barrel file as side-effectful - Or move the dayjs.extend call to each module that needs it (cleaner but more invasive)

 Recommend the array form for safety.

 1b. If tree-shaking is insufficient, add explicit externals

 File: packages/frontend/vite.config.sdk.ts

 Add to rollupOptions.external:
 'ajv', 'ajv-errors', 'ajv-formats', 'ajv/dist/2020',
 'better-ajv-errors', 'pegjs'

 1c. Verify by building and searching the output:
 pnpm -F frontend build-sdk
 grep -c "new Function" packages/frontend/sdk/dist/sdk.es.js
 grep -c "eval(" packages/frontend/sdk/dist/sdk.es.js

 Phase 2: Externalize Vega from SDK bundle

 Vega is only used for custom visualizations (a power-user feature), not standard embedded charts. It's lazy-loaded via React.lazy but Vite lib mode inlines dynamic imports into a single bundle.

 2a. Add vega packages to SDK externals

 File: packages/frontend/vite.config.sdk.ts

 Add to rollupOptions.external:
 'react-vega', 'vega', 'vega-lite', 'vega-embed', 'vega-themes'

 2b. Make vega-themes import lazy

 File: packages/frontend/src/components/CustomVisualization/index.tsx

 Currently line 5 imports vega-themes statically:
 import { dark as vegaDarkTheme } from 'vega-themes';

 Move this inside the lazy-loaded component or import dynamically, so it doesn't fail when externalized.

 2c. Add graceful error handling

 The existing React.lazy + Suspense + error boundary should catch the missing module. Optionally add a
 try-catch around the lazy import to show a "Custom visualizations not available in embedded mode" message
  instead of an error.

 Phase 3: Patch ECharts dead code path

 The only new Function in ECharts is in GeoJSONResource.js — a fallback for environments without JSON.parse. Never actually executed in browsers, but CSP scanners flag it.

 3a. Add a Vite transform plugin to the SDK build config

 File: packages/frontend/vite.config.sdk.ts

```
 {
     name: 'patch-echarts-csp',
     transform(code, id) {
         if (id.includes('GeoJSONResource')) {
             return code.replace(
                 /new Function\('return \(' \+ source \+ ';\)'\)\(\)/g,
                 'JSON.parse(source)'
             );
         }
     }
 }
```

 This is safe because the new Function is only reached when JSON.parse doesn't exist.

 Verification

 1. Build SDK: pnpm -F frontend build-sdk
 2. Search for violations: grep -c "new Function\|eval(" packages/frontend/sdk/dist/sdk.es.js
 3. Run common tests: pnpm -F common test
 4. Run frontend typecheck: pnpm -F frontend typecheck
 5. Test with SDK test app: load a dashboard in packages/sdk-test-app and verify charts render
 6. Test CSP: create an HTML page with <meta http-equiv="Content-Security-Policy" content="script-src
 'self';">, load the SDK, verify no console violations

 A customer with strict CSP (no unsafe-eval) cannot use @lightdash/sdk because four bundled libraries use eval()/new Function() at runtime. Their CSP blocks these calls, preventing the  SDK from rendering dashboards. They're falling back to iframe embedding but prefer the React SDK for tighter integration.

charliedowler:
## What & why

The `@lightdash/sdk` bundle ran runtime code-generation (`eval`/`new Function`/`Function`) that a strict Content-Security-Policy (`script-src` without `'unsafe-eval'`) blocks, so embedded dashboards failed to render for consumers with a hardened CSP — they had to fall back to iframe embedding.

Closes #21276 · PROD-6084

## Root cause (found by browser-testing the SDK under a strict CSP)

Two distinct runtime code-generators fired during a normal embedded render:

1. **ajv — breaks every dashboard at SDK load.** `packages/common/src/dbt/validation.ts` constructed `new Ajv2020()` + `addSchema([...])` **at module top level**, so merely importing the module compiled the draft-2020 meta-schema validator via `new Function` *at SDK load time*. Under a strict CSP this throws `EvalError` and the SDK crashes before any chart renders. `ManifestValidator` only validates dbt manifests on the backend/CLI — never at SDK render — so the ajv instance is now **lazily constructed** and never runs in the browser.

2. **Vega — breaks dashboards with a custom-viz tile at render.** A `ChartType.CUSTOM` (Vega-Lite) tile renders via react-vega's `VegaEmbed`, and `vega-runtime` compiles the spec's expressions with the bare `Function(...)` constructor while rendering. So a dashboard containing a custom visualization still threw `EvalError` under a strict CSP. Fixed by switching vega-embed to its **AST expression interpreter** (`ast: true`) — no codegen.

## Changes

- **ajv (the main fix):** lazy-init the `Ajv2020` instance in `dbt/validation.ts` so importing the module no longer compiles validators at load.
- **PEG.js filter grammar:** precompiled into a committed, eval-free recursive-descent parser (`filterGrammar.parser.ts`) instead of calling `peg.generate()` at runtime — the idiomatic PEG.js workflow. The grammar moves to `filterGrammar.grammar.ts`; `generate:`/`check:filter-grammar-parser` scripts keep it in sync (the check regenerates + formats + compares by source, so any grammar change is caught even without a matching test case).
- **Vega custom visualizations:** pass `ast: true` to `VegaEmbed` (`CustomVisualization`). vega-embed already bundles `expressionInterpreter` and selects it when `ast` is set, so this is a one-line switch with **no new dependency** and unchanged bundle text — the cost is runtime-only (AST interpretation vs. compiled accessors, negligible at dashboard scale).
- **Regression guards:**
  - A runtime test (`validation.csp.test.ts`) traps `new Function` during import and fails if any code is generated at module load — this guards the ajv fix, which a bundle text-grep cannot (lazy vs eager ajv produces identical bundle text).
  - A runtime test (`CustomVisualization/index.csp.test.tsx`) asserts the `ast: true` option is set — likewise invisible to a bundle grep.
  - A `check:sdk-csp` script + a parallel CI job grep the built bundle for the removed PEG.js signatures and enforce a total code-generation count. The check now counts bare `Function(...)` as well as `new Function`/`eval(` (Vega uses the bare form), so it is no longer blind to the most CSP-relevant code-generator in the bundle.

No library upgrades; the main frontend app build is unaffected.

## Test plan

- `pnpm -F common test` — passes (filter-grammar, dbt, translator, validation); typecheck/lint clean.
- `pnpm -F common check:filter-grammar-parser` — parser in sync with the grammar (source + behaviour).
- **ajv still validates correctly (e2e):** direct `ManifestValidator` run accepts valid models and rejects schema-invalid input (bad metric/dimension type) with correct error paths; a live backend project recompile completed (`jobStatus=DONE`) and produced 60 explores.
- **Renders under CSP (e2e):** built the SDK and loaded the embedded demo dashboard in `sdk-test-app` under an enforced `script-src 'self' 'unsafe-inline'` (no `'unsafe-eval'`). Before this change the dashboard was blank with an `EvalError`; after it, all charts, the big-value tile and the results table render with **zero CSP violations**.
- **Vega custom-viz renders under CSP:** reproduced the render path (real `vega-embed`, trapping the `Function` constructor to simulate CSP): default options make **54** `Function()` codegen calls and throw `EvalError`; with `ast: true` → **0** codegen calls and the chart renders. `pnpm -F frontend check:sdk-csp` passes with 0 forbidden signatures.

## Scope / known limitations

Verified for the embedded dashboard render path — ECharts charts, tables, big-value tiles, markdown, **and custom Vega-Lite tiles** — which is now eval-free *at runtime* under a strict CSP.

Some dependencies still ship dormant `new Function`/`eval`/`Function` **text** that **never executes on a dashboard render** but would be flagged by a *static* CSP scanner:

- **ajv** validator codegen (now never invoked in the browser, but the code is bundled).
- **Vega** expression compiler (now bypassed at render by the AST interpreter, but the compiled-codegen path remains in `vega-runtime`).
- **ECharts** `GeoJSONResource` fallback — a provably-unreachable dead branch (guarded by `typeof JSON.parse`, present in every CSP-capable browser). *Intentionally left as-is* rather than rewritten at build time.
- **html2canvas** (image/PDF export) and **source-map** — only execute on those specific feature actions, not on render.

Driving the bundled text to zero for static-scanner consumers (ajv standalone validators, lazy-loading html2canvas/source-map) is a separate follow-up; this PR fixes the **runtime** blockers that prevented dashboards from rendering.


charliedowler:
## Controlled before/after: reproduced the original bug, proved the fix

To verify this actually resolves the reported problem (rather than "should"), I ran a single-variable controlled experiment in the SDK test app: same app, same **enforced** CSP (`script-src 'self' 'unsafe-inline'` — no `'unsafe-eval'`), same embedded dashboard. The **only** thing toggled was `packages/common/src/dbt/validation.ts`.

### BEFORE — `main`'s code (module-level `new Ajv2020()`)

Console under the strict CSP:

```
Error compiling schema, function code: ...return function validate1(data...)...
  schemaPath:"#/definitions/nonNegativeInteger/type"...
Uncaught EvalError: Evaluating a string as JavaScript violates the following Content Security
Policy directive because 'unsafe-eval' is not an allowed source of script: script-src 'self' 'unsafe-inline'
```

→ The SDK throws while ajv compiles a validator via `new Function` **at module load**. **The dashboard renders blank.** This is the issue's exact symptom.

### AFTER — this PR (lazy `getAjv()`)

Same harness, same CSP, only `validation.ts` changed:

- No `EvalError`, no CSP violation (console clean apart from an unrelated locale 404).
- **Dashboard renders fully** — revenue-by-payment-method, orders-over-time and average-spend charts, the “Total revenue” tile, and the results table all populate with live data.

### Why this is conclusive

The *only* difference between the two runs is the lazy-init change, so the appearance/disappearance of the CSP `EvalError` is unambiguously attributable to it — not to any other build or environment difference.

**Caveat:** the A/B ran on the source-build test harness (the published `dist` tarball has a separate React peer-dep quirk in this monorepo test setup), but it builds the SDK from the same source and exercises the same `validation.ts` path, so the single-variable comparison holds. The `dist` bundle was separately confirmed free of the removed eval signatures via `check:sdk-csp`.

PR #24750, opened 2026-06-25
fix(mcp): surface queryUuid in metric query result text

### LINKED ISSUE #24671: MCP `render_chart` fails for Editor role when previewing query visualizations in Claude
A customer reported that when using the Lightdash MCP from Claude chats with a user on the Editor role, running a query and previewing the visualization fails with the error: `The render_chart call is failing to find the query in project history - this sometimes happens due to timing.`

tatianainama:
## What & why

`render_chart` loads a completed query from query history by the `queryUuid` returned by a prior `run_metric_query`. That tool's machine-readable `structuredContent.result` includes the `queryUuid`, but its `content` was CSV only — the id appeared nowhere in the content blocks.

MCP clients/agents that act on `content` (rather than `structuredContent`) therefore had no way to recover the `queryUuid`, and would call `render_chart` with an **invented id**. That id doesn't exist in query history, so the call fails with:

> Error rendering chart: Query <id> not found for project <id>

…which the agent typically rationalises as a transient/timing issue. It reproduces regardless of role (admin and editor alike) and regardless of project — it's purely a response-contract gap, not a permissions problem.

## Change

- Emit the `queryUuid` as a **separate `text` content block** in the metric-query result (`buildMetricQueryPollResult`, shared by `run_metric_query` and `get_query_result`). The existing CSV block is left **byte-identical**, so anything that renders/parses it is unaffected; agents that read the content blocks can now recover the id.
- Tighten the `render_chart` tool + param descriptions: copy the `queryUuid` from the prior response's `queryUuid: <id>` block; never invent, guess, or reuse one from another query/session.
- Update the metric-query response-shape description, the affected unit test, and the contract snapshot.

`structuredContent` is unchanged — the spec's typed contract (`outputSchema` → `structuredContent`) is untouched, and multiple `content` blocks are explicitly allowed by the MCP spec.

## Testing

- `pnpm -F backend typecheck:fast` and `pnpm -F common typecheck:fast` pass.
- `McpService.queryPolling` and `mcpToolContracts.snapshot` suites pass (snapshot updated).
- Verified end-to-end against a live MCP server: `run_metric_query` returns the CSV block plus a `queryUuid: <id>` block, and `render_chart` succeeds with that id.

Closes: #24671

### DIFF
--- a/packages/backend/src/ee/services/McpService/McpService.queryPolling.test.ts
+++ b/packages/backend/src/ee/services/McpService/McpService.queryPolling.test.ts
@@ -1487,6 +1487,10 @@ describe('MCP async query polling', () => {
                     type: 'text',
                     text: 'orders_count\n1\n',
                 },
+                {
+                    type: 'text',
+                    text: `queryUuid: ${queryUuid}`,
+                },
             ],
             structuredContent: {
                 result: {
--- a/packages/backend/src/ee/services/McpService/McpService.ts
+++ b/packages/backend/src/ee/services/McpService/McpService.ts
@@ -964,11 +964,21 @@ export class McpService extends BaseService {
             columns: csvHeaders,
         });
 
+        // render_chart needs the queryUuid, and clients that only surface
+        // `content` (not structuredContent) can't otherwise recover it and may
+        // invent an invalid id. Emit it as a separate text block so the result
+        // block stays untouched for clients that render it directly.
+        const body = rows.length === 0 ? 'Query returned 0 rows.' : csv;
+
         return {
             content: [
                 {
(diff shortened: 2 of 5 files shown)


tatianainama:
* **#24750** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24779, opened 2026-06-25
feat(ai): show Ask AI admin settings to project-scoped users

sdolidze:
## Summary

PR 2 of 2. Surfaces the org-wide AI admin pages (Threads, Agents, Reviews, prompt activity) to **project-scoped AI admins**, filtered to the projects they can manage. Previously these surfaces were org-admin-only and hidden from anyone who only administers AI in specific projects.

Stacks on #24778, which adds the `OrganizationAiAgent` scope this PR gates on.

## Changes

**Backend** — `AiAgentAdminService`
- New `resolveReadScope(user)` → `{ kind: 'all' }` for org AI admins (`manage:OrganizationAiAgent`), else `{ kind: 'projects', projectUuids }` filtered by `manage:AiAgent` over the org's projects (mirrors `OrganizationService.getProjects`). Throws when the principal has neither.
- `getAllThreads` (project allowlist via `filters.projectUuids`, empty-intersection short-circuits to `[]` so the model never falls back to all rows), `listAgents` (project allowlist), `listReviewItems` (in-memory filter), and `getPromptActivity` are now project-aware.
- The five per-item review getters call `assertProjectInScope(scope, item.projectUuid)` — guards fingerprint IDOR (an off-limits item returns 403, never its body).
- Writes (status / reorder / assignee / writeback / retest), review-signals, and notification settings stay org-only.

**Frontend**
- `useSettingsContext` computes `canManageOrgAiAgent` and `hasAnyAiAgentAccess` (org `manage:OrganizationAiAgent` OR project `manage:AiAgent` over the user's projects).
- The settings nav + routes show "Ask AI" for project-scoped users; **General** (router / org config) stays org-only; the `/ai` redirect lands project-scoped users on Threads.
- Org gating repointed to `OrganizationAiAgent` via a new `useAiAgentOrgPermission` hook (NavBar review badges, AiSearchBox).

## Read-scope resolution

```
manage:OrganizationAiAgent (org)   → { kind: 'all' }       → every project
  else manage:AiAgent per project   → { kind: 'projects' }  → only those projects
  else                              → ForbiddenError
```

Gating on `manage` (not `view`) is deliberate: org `view:AiAgent` / `view:OrganizationAiAgent` is granted org-wide down to interactive_viewer, so a `view` gate would hand every interactive_viewer full cross-project admin reads. `manage` preserves the prior admin/developer floor.

## Access matrix (verified live against a two-project org)

```
Principal                          Threads / Agents / Reviews / Activity
[...]

PR #24778, opened 2026-06-25
feat(ai): split org AI admin into OrganizationAiAgent scope

sdolidze:
## What this does

PR 1 of 2 (base of the stack; #24779 builds on it).

The `AiAgent` CASL subject was overloaded: it gated **both** project-level agent management and org-level AI administration (router config, org AI settings, review-notification recipients). Org grants condition only on `organizationUuid`, so an org grant satisfied project-scoped checks and vice-versa — the two capabilities could not be granted independently.

This keeps `AiAgent` as the **project-level** subject and adds a new **org-level** `OrganizationAiAgent` subject. The org-settings backend checks move to it. No behaviour change — see the per-principal table.

**Scope mapping**

```
Before:  view/manage:AiAgent  → project agents AND org AI settings (conflated)

After:   view/manage:AiAgent             → project agent management   (project-assignable only)
         view/manage:OrganizationAiAgent → org AI settings            (org-assignable, enterprise)
```

## Check sites moved to OrganizationAiAgent

| Check | Subject before | Subject after |
| --- | --- | --- |
| `AiRouterService.assertManage/ViewAiAgent` | `AiAgent {org}` | `OrganizationAiAgent {org}` |
| `AiOrganizationSettingsService.checkManageAiAgentAccess` | `AiAgent {org}` | `OrganizationAiAgent {org}` |
| `AiAgentReviewNotificationService` recipient filter | `AiAgent {org}` | `OrganizationAiAgent {org}` |

Project-level `subject('AiAgent', { projectUuid })` checks in `AiAgentService` are unchanged.

## Who's affected — nobody (reach-preserving)

| Principal | Before | After |
| --- | --- | --- |
| Org admin / developer (system) | org AI settings via `manage:AiAgent{org}` | same — `manage:OrganizationAiAgent{org}` added, `manage:AiAgent` retained |
| Org interactive_viewer (system) | view via `view:AiAgent{org}` | same — `view:OrganizationAiAgent{org}` added |
| Service accounts (ORG_READ / ORG_ADMIN) | view/manage via `AiAgent{org}` | same — `OrganizationAiAgent` added alongside |
| Project custom role (`manage:AiAgent`) | project agents only; no org settings | unchanged (`AiAgent` stays project-assignable) |
| Org custom role | could not hold `AiAgent` (not org-assignable) → no org AI | unchanged; may now opt into `OrganizationAiAgent` |

[...]

PR #24780, opened 2026-06-25
fix(custom-roles): show correct scopes after switching role level

sdolidze:
## Summary

When editing a custom role, opening a scope group (e.g. **AI**), then switching the role **level** (Organization ↔ Project), the right-hand panel showed **0 selected** even though that group had scopes selected — while the left sidebar count for the same group correctly showed e.g. **12**. Switching to another group and back fixed it.

Stacks on #24779. Only affects the scope selector after a level/search change while a group is open; the persisted role data was always correct.

## Root cause

`ScopeSelector` stored the entire selected-group **object** in state. The grouped-scope objects are rebuilt with fresh, per-level scope lists whenever `level` or the search term changes, but `effectiveSelectedGroup` returned the **stale** object as long as a group with the same key still existed — and `AI` exists at both levels:

```ts
// effectiveSelectedGroup — returned the stale object, not the rebuilt one
if (selectedGroup && filteredScopes.find((g) => g.group === selectedGroup.group)) {
    return selectedGroup; // old level's scope names
}
```

So `ScopePanel` rendered the previous level's scope names, which the level switch had just re-derived to unchecked → nothing selected. The sidebar count is computed from the fresh `filteredScopes`, so the two disagreed. Clicking another group called `setSelectedGroup(freshObject)`, which healed the display.

## Fix

Store the group **key** instead of the object and resolve the active group from the freshly-built `filteredScopes` during render, so a level/search change can never leave a stale group object on screen.

- `ScopeSelector.tsx` — `selectedGroup: GroupedScopes` state → `selectedGroupKey: GroupedScopes['group']`; `effectiveSelectedGroup` now always `filteredScopes.find(g => g.group === selectedGroupKey) ?? filteredScopes[0] ?? null`; the group click stores `group.group`.

## Disagreement that the fix removes

```
group "AI", after Org → Project switch
                 before fix          after fix
  sidebar count   12  ✓ (fresh data)   12  ✓
  right panel      0  ✗ (stale object) 12  ✓
```

## Test plan

- [x] `pnpm -F frontend typecheck` (no ScopeSelector errors) + lint clean
- [ ] Manual repro: duplicate the Developer system role at Organization level → open AI → switch to Project level → the 12 project AI scopes show as selected (previously 0); sidebar and panel agree
- [ ] Manual: search filtering still selects/restores the right group; switching groups still works

## Related (not fixed here)

The edit/save path silently self-heals mis-leveled scopes in the DB: mis-leveled scopes load as unselected, and the first save diffs against the original list so they land in `scopes.remove` (the remove path isn't level-validated, unlike the add path's `validateScopesLevel`). Net: bad data is stripped on the next save with no warning. Worth a separate ticket — this PR only fixes the display bug.

### DIFF
--- a/packages/frontend/src/ee/features/customRoles/components/ScopeSelector/ScopeSelector.tsx
+++ b/packages/frontend/src/ee/features/customRoles/components/ScopeSelector/ScopeSelector.tsx
@@ -159,9 +159,12 @@ const ScopePanel: FC<{
 export const ScopeSelector: FC<ScopeSelectorProps> = ({ form, level }) => {
     const [searchTerm, setSearchTerm] = useState('');
     const [debouncedSearchTerm] = useDebouncedValue(searchTerm, 300);
-    const [selectedGroup, setSelectedGroup] = useState<GroupedScopes | null>(
-        null,
-    );
+    // Store the selected group's key, not the object. The group objects are
+    // rebuilt (with fresh scope lists) whenever `level` or the search term
+    // changes, so holding the object would render a stale group's scopes.
+    const [selectedGroupKey, setSelectedGroupKey] = useState<
+        GroupedScopes['group'] | null
+    >(null);
 
     const allGroupedScopes = useMemo(
         () => getScopesByGroup(true, level),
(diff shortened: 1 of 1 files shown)


sdolidze:
* **#24780** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24780?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24779** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24779?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24778** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24778?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sdolidze:
### Merge activity

* **Jun 25, 4:48 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/24780).
* **Jun 25, 4:52 PM UTC**: [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/24780) rebased this pull request as part of a merge.
* **Jun 25, 4:52 PM UTC**: @sdolidze merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightdash/lightdash/24780).

PR #24790, opened 2026-06-26
feat: add dynamic SI compact formatting

### LINKED ISSUE #20651: Dynamic compact format showcase of the numbers
### Description

Hey I love your product,

I would love to see dynamic compact format of the number data type, if something is in billions -> B, millions -> M, etc. and not showcase 0.003B. So this would dynamically check the number size and apply compact formating.

I believe that this is a small feature that can improve UX for teams that have large span of numbers. I am aware that this can be applied in data exploration, but why have an extra step when dealing with something simple as this. For example what if we change the date granularity from month -> day where metric in daily granularity has millions, and on monthly it has billions, it would be nice to see this dynamically change, numbers would be cleaner instead of 0. something.

nazihkalo:
## Summary
- add native dynamic SI compact formatting via `Compact.AUTO` and legacy `Format.SI`
- dynamically scale numeric and currency values to K/M/B/T by magnitude while preserving fixed compact formats
- expose SI formatting in cartesian axis controls and Auto compact style in Big Number controls
- update chart/model/dbt schemas and add focused formatter/UI option tests

Fixes #20651

## Validation
- `git diff --cached --check` before commit: passed
- JSON schema parse check: passed before commit
- attempted `npm exec --yes pnpm@10.33.0 -- install --frozen-lockfile`; dependency bootstrap could not complete under local Node v25.9.0 because native deps failed to build (`cpu-features`/`nan` V8 API errors, then `duckdb` source compile was interrupted after the prior failure)
- targeted tests were therefore not runnable in this checkout because local test binaries depend on the failed install

### DIFF
--- a/packages/backend/src/generated/routes.ts
+++ b/packages/backend/src/generated/routes.ts
@@ -3724,6 +3724,7 @@ const models: TsoaRoute.Models = {
     Compact: {
         dataType: 'refEnum',
         enums: [
+            'auto',
             'thousands',
             'millions',
             'billions',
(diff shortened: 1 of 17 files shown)


sdolidze (changes requested):
Thanks for this @nazihkalo, the approach is spot on. Per-value compaction (rather than per-column) is exactly the direction we'd decided on internally, and `Compact.AUTO` slots in cleanly. I checked the branch out and ran the suite locally; here's what stands between this and merge.

**Blockers (CI is currently red on these):**

1. **Your new `CartesianChartDataModel` value-label test fails** (`formats SI value labels dynamically`, expects `1.2M`, gets `-`).
   - The production formatter is correct; it matches the existing `PERCENT`/stack-100 pattern, where ECharts passes `params.value` as an object keyed by dimension name.
   - The test mock passes a positional array instead, so the lookup misses.
   - Fix the mock to object form, e.g. `value: { category: 'Jan', value: 1200000 }`.

2. **Lint error** at `formatting.test.ts:701`.
   - `const AUTO = Compact.AUTO;` trips `prefer-destructuring`.

3. **The JSON schemas are generated, not hand-edited.**
   - `pnpm check:chart-as-code-schema` fails because the hand edits don't match generator output.
   - Run `pnpm generate-api` first (updates `swagger.json` from your new enum values), then `pnpm generate:chart-as-code-schema`, and commit the result.
   - Also add `"auto"` to the `Compact` enum in `packages/common/src/dbt/schemas/lightdashMetadata.json`, which is the compile-time dbt validation schema and was missed, so `compact: auto` in a dbt model would currently fail validation.

**Worth resolving before merge (correctness):**

4. **`Compact.AUTO` is a no-op in the format-expression (numfmt) path.**
   - `customFormatConversionFnMap.compact` returns the expression unchanged for AUTO.
   - That path drives Excel exports, metric/dimension format overrides in result tables, and the live format preview.
   - So a field set to "auto" would compact in charts/big-numbers but not in those surfaces, and the preview would show no compaction while the table does.
   - ECMA-376 can't express value-dependent scaling, so we need a deliberate decision here rather than a silent pass-through.

Nice catch switching BigNumber's `StyleOptions` off `Object.values(CompactConfigMap)`, that was leaking byte units into that dropdown, so that's a real fix beyond the feature itself.

Happy to help get these over the line if useful.

sdolidze:
Merged: https://github.com/lightdash/lightdash/pull/24882

PR #24837, opened 2026-06-26
fix(catalog): scope table-selection filter to non-allowlist searches

### LINKED ISSUE #24829: AI agent can't query fields that its Data access preview shows as available
## Description

An AI agent cannot query fields that its **Data access** settings show as available.

When a project's table selection is restricted to a specific tag (so some dimension-only models are excluded from the table selection), fields in those excluded models are filtered out of the agent's field search at query time — **even when the field is tagged for the agent and joined into an explore that is included in the table selection**.

The result is a mismatch between two views of the same agent:

- The agent's **Data access** preview lists the field as available (it filters by the agent's tags).
- The agent's **actual field search** at query time does not return the field, so the agent reports the field doesn't exist.

This is confusing because the field is correctly tagged, the preview confirms it, but the agent still can't use it. It systematically affects dimensions that live in join-only models (e.g. country, date) when those models are excluded from the table selection.

### Expected behaviour

An agent should be able to query any field shown as available in its Data access preview. Agent field visibility should depend only on the agent's tags and the user's access — not on the project's table-selection setting.

## Steps to Reproduce the Bug or Issue

1. Set the project's table selection to "models with tag A" (so at least one dimension-only model, joined into an included model, is **not** tagged A and is therefore excluded).
2. On a dimension in that excluded model, add the agent's tag (e.g. `ai`) at the column level. The model is joined into an included explore that also has a metric.
3. Open the agent's **Data access** settings and confirm the dimension shows as available.
4. Ask the agent to break that metric down by the tagged dimension.
5. The agent reports it can't find the dimension, despite step 3 showing it as available.

## version

cloud

tatianainama:
Closes: #24829

### Description:
Catalog search applied the project table-selection config even when the caller passed an explicit explore allow-list. That allow-list is already the authoritative visibility set, so applying the table-selection on top of it hid fields the caller was entitled to see.

This skips the table-selection filter when an explore allow-list is provided; visibility is then governed solely by the allow-list and the existing user-attribute checks. Searches without an allow-list (catalog browse/search, metrics explorer) are unchanged.

Since the allow-list filter only runs for an explicit `Table`/`Field` catalog type, the method now asserts that an allow-list is paired with a type — turning a would-be silent visibility gap into a loud failure rather than relying on the convention implicitly.

### DIFF
--- a/packages/backend/src/models/CatalogModel/CatalogModel.ts
+++ b/packages/backend/src/models/CatalogModel/CatalogModel.ts
@@ -783,6 +783,19 @@ export class CatalogModel {
         hasTimeDimension?: boolean;
         tags?: string[];
     }): Promise<KnexPaginatedData<CatalogItem[]>> {
+        // filteredExplores enforces visibility through the type-specific allow-list
+        // filter below, which only runs for Table/Field. Without an explicit type that
+        // filter is skipped and nothing constrains the results, so fail loud instead.
+        if (
+            filteredExplores &&
+            type !== CatalogType.Table &&
+            type !== CatalogType.Field
+        ) {
+            throw new UnexpectedServerError(
+                'filteredExplores requires an explicit Table or Field catalog type',
+            );
+        }
+
         // Use websearch_to_tsquery for AI Agent queries for better natural language support
         const useWebSearch =
             context === CatalogSearchContext.AI_AGENT ||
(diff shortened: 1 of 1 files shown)


tatianainama:
* **#24837** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24837?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #24844, opened 2026-06-28
fix: respect convert_timezone: false in filter WHERE clause generation

### LINKED ISSUE #24840: [Timezones] convert_timezone: false not respected by filter WHERE clause generation
## Bug description

When a timestamp dimension is annotated with `convert_timezone: false`, display and grouping correctly treat the column as a wall-clock value and skip project-timezone conversion. However, filter SQL still applies the project-timezone conversion to the column, causing a mismatch: a row displayed as one date can be matched by a filter targeting the prior date. If you've already done timezone conversion (the mostly likely case for using that config), filtering should be consistent with display and the config, and no conversion should be applied.

## Steps to reproduce

1. Enable the timezone support feature flag and set a non-UTC project timezone (e.g. America/Chicago).
2. Define a timestamp dimension with `convert_timezone: false` on a column that stores pre-converted local-time values.
3. Run a query - the column displays the stored value correctly (no shifting).
4. Identify a row with a time near midnight, say 11:45pm on 2026-06-01 in local Chicago time — then set a date filter for the next day, for this example "is 2026-06-02".
5. Observe that it captures the June 1st row, even though the filter is clearly June 2nd — `convert_timezone: false` was respected for the label but not the filter.

## Expected behavior

`convert_timezone: false` should disable timezone conversion for that column everywhere in the UX - display, grouping, and filters. The filter should compare against the raw stored value without applying project-timezone wrapping.

More fundamentally, the time displayed should be the time everything in the UX responds to.

## Actual behavior

Filters apply project-timezone conversion to the column regardless of the `convert_timezone: false` flag. This creates a mismatch between what the user sees in the table and what the filter matches.

## Potential root cause analysis

In `MetricQueryBuilder.getFilterRuleSQL`, `getTimezoneAwareDimensionSql` is called with `respectConvertTimezone: false`, explicitly bypassing the `skipTimezoneConversion` flag that display and grouping already respect. The code comment notes this is intentional: "Filters always wrap by project tz - even for dims with `convert_timezone: false`."

## Environment

- Lightdash version: v0.3238.0
- Warehouse: Databricks

## Additional context

The `skipTimezoneConversion` flag is already propagated to all dimension children and respected by both display formatting and the timezone-aware DATE_TRUNC path. The fix is to align the filter path with the same flag, so `convert_timezone: false` means no timezone conversion anywhere for that column, consistent with what the config name implies.

Abdul-Moiz31:
## Summary
- Dimensions annotated with `convert_timezone: false` already skip project-timezone conversion for display and grouping. Filters did not — `getFilterRuleSQL` forced `respectConvertTimezone: false`, so the WHERE clause kept wrapping the column in a project-timezone conversion regardless of the flag. This caused filter results to mismatch what's displayed near midnight boundaries (e.g. a row shown as day N could be matched by a filter for day N+1).
- `getFilterRuleSQL` now goes through the same `skipTimezoneConversion`-aware path as display/grouping, so `convert_timezone: false` means no timezone conversion anywhere for that column — consistent with the config name.
- The filter's target *value* (what the user types/selects) is unaffected and still gets converted by project timezone as before — only the column-side wrapping changes.
- Removed the now-unused `respectConvertTimezone` override parameter from `getTimezoneAwareDimensionSql` since no caller needs it anymore.

## Test plan
- [x] Updated `MetricQueryBuilder.test.ts` test (`convert_timezone: false on base dim`) to assert SELECT and WHERE both render the bare (unwrapped) expression
- [x] `pnpm -F backend exec jest MetricQueryBuilder.test.ts` — 142/142 passing
- [x] Lint clean on changed files
- [ ] Manual verification: timestamp dimension with `convert_timezone: false` in a non-UTC project, filter near a local-midnight boundary no longer matches the wrong day

Closes: PROD-8537
Closes: #24840


tatianainama:
Thanks for the PR. We've changed our [contribution policy](https://github.com/lightdash/lightdash/blob/main/.github/CONTRIBUTING.md), so I'm closing this. #24840 stays open and we'll pick it up on our side.

PR #24845, opened 2026-06-29
fix: deduplicate distinct metrics within selected dimension groups

rephus:
Closes: https://linear.app/lightdash/issue/PROD-8531/sum-distinct-wrong-values-when-grouping-by-dimensions-not-in-distinct

CleanShot 2026-06-29 at 11 24 00

Before the fix, the deduped column would have behaved like a scalar CROSS JOIN, so every row would show the same grand total:
bank_transfer  $299.00
coupon         $299.00
credit_card    $299.00
gift_card      $299.00
After the fix, it partitions the distinct calculation by the selected dimension too, so you get the correct per-group values:
bank_transfer  $67.50
coupon         $51.00
credit_card    $135.00
gift_card      $45.50

### Description:

Previously, when a `sum_distinct` or `average_distinct` metric was queried alongside dimensions that were not part of the metric's `distinct_keys`, the deduplication CTE would collapse to a scalar value and be `CROSS JOIN`ed onto the base CTE. This meant every output row received the same global deduplicated total, regardless of the selected dimension grouping.

This changes the deduplication semantics so that selected dimensions are always included in the `PARTITION BY` clause of the inner `ROW_NUMBER()` subquery, and the outer CTE groups by all selected dimensions. The CTE is then `INNER JOIN`ed to the base CTE on those dimensions rather than `CROSS JOIN`ed. As a result, each output row receives a correctly deduplicated value scoped to its own dimension group, rather than a single global total repeated across all rows.

The `CROSS JOIN` path is now only used when there are no selected dimensions at all.

New unit tests verify that `sum_distinct` and `average_distinct` both produce `INNER JOIN` (not `CROSS JOIN`) and include the selected dimensions in the `PARTITION BY` when dimensions are present. Snapshot tests and the E2E fanout deduplication test have been updated to reflect the corrected per-group behavior.

### DIFF
--- a/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.test.ts
+++ b/packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.test.ts
@@ -47,6 +47,7 @@ import {
     INTRINSIC_USER_ATTRIBUTES,
     METRIC_QUERY,
     METRIC_QUERY_AVERAGE_DISTINCT_NO_DIMS,
+    METRIC_QUERY_AVERAGE_DISTINCT_WITH_DIMS,
     METRIC_QUERY_CROSS_MODEL_SUM_DISTINCT,
     METRIC_QUERY_CROSS_MODEL_SUM_DISTINCT_NO_DIMS,
     METRIC_QUERY_CROSS_TABLE,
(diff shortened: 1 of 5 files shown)


rephus:
* **#24845** <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightdash/lightdash/24845?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rephus:
Unrelated error tests 
CleanShot 2026-06-29 at 11 19 45
image

PR #24850, opened 2026-06-29
fix(timezone): honour convert_timezone false in filter where clause

### LINKED ISSUE #24840: [Timezones] convert_timezone: false not respected by filter WHERE clause generation

almeidabbm:
Closes https://github.com/lightdash/lightdash/issues/24840

## Problem

A timestamp dimension with `convert_timezone: false` renders its raw warehouse value on the SELECT side (display + grouping skip the project-timezone conversion), but the filter WHERE clause still wrapped the column in the project timezone. On a time-interval (truncated/extracted) dimension this made the displayed bucket and the filter window use different frames of reference, so a row shown under one day could be matched by a filter for an adjacent day.

## Root cause

`MetricQueryBuilder.getFilterRuleSQL` called `getTimezoneAwareDimensionSql` with `respectConvertTimezone: false`, explicitly bypassing the `skipTimezoneConversion` flag that the SELECT/grouping path already honours.

## Fix

The filter path now reuses the same unwrapped column expression as the SELECT. Only the column expression changes; the filter value side is untouched, so relative date windows ("today", "this week", "in the last N days") still resolve in the project timezone. With no caller overriding it anymore, the now-constant `respectConvertTimezone` parameter is removed.

## Behaviour change (warehouse data, project tz `America/Chicago`)

Seed row stored `2024-01-15 00:00:00`, which displays/groups as `2024-01-15` under `convert_timezone: false`:

| filter | before | after |
| --- | --- | --- |
| `is 2024-01-15` (the displayed day) | no match | **matches** |
| `is 2024-01-14` (adjacent day) | **matches** (bug) | no match |

After the fix, the filter agrees with what the row displays as.

## Manual verification

Ran against the `timezone_test` seed model with `EnableTimezoneSupport` on, project timezone `America/Chicago`, isolated to the row above (`event_id = 12`).

**Absolute filter** on `event_timestamp_raw_utc_day` (the `convert_timezone: false` dimension):

```sql
-- is 2024-01-15  (the displayed day)
WHERE (DATE_TRUNC('DAY', "timezone_test".event_timestamp_raw_utc)) = ('2024-01-15')
-- -> 1 row

-- is 2024-01-14  (adjacent day)
WHERE (DATE_TRUNC('DAY', "timezone_test".event_timestamp_raw_utc)) = ('2024-01-14')
-- -> 0 rows
```

**Relative filter** (large window needed because the seed data is from 2024; "today" is 2026-06-29):

```sql
-- in the last 1000 days
WHERE (DATE_TRUNC('DAY', "timezone_test".event_timestamp_raw_utc)) >= ('2023-10-03')
  AND (DATE_TRUNC('DAY', "timezone_test".event_timestamp_raw_utc)) <= ('2026-06-29')
-- -> 1 row

-- in the last 100 days
WHERE (DATE_TRUNC('DAY', "timezone_test".event_timestamp_raw_utc)) >= ('2026-03-21')
  AND (DATE_TRUNC('DAY', "timezone_test".event_timestamp_raw_utc)) <= ('2026-06-29')
-- -> 0 rows
```

Note the column expression is bare (no `AT TIME ZONE`) on both SELECT and WHERE, and the relative window boundaries are real calendar dates computed in the project timezone, not UTC.

## Tests

- `MetricQueryBuilder.test.ts`: a `convert_timezone: false` dimension now renders the bare expression on both SELECT and WHERE; a dimension without the flag still wraps the WHERE in the project timezone.
- `api-tests/convertTimezoneFilter.test.ts`: end-to-end filter on the opted-out dimension matches the raw bucket day, and a normal dimension still shifts with the query timezone.

### DIFF
--- a/docs/timezones/draft-user-documentation.md
+++ b/docs/timezones/draft-user-documentation.md
@@ -104,7 +104,7 @@ columns:
         convert_timezone: false
 ```
 
-Use cases: audit logs, system timestamps, pre-converted values. The column will render exactly what the warehouse stores.
+Use cases: audit logs, system timestamps, pre-converted values. The column will render exactly what the warehouse stores. Filtering on the column uses the same raw value, so a filter matches the day the row displays under.
 
 ### `wall_clock_timezone:` — declare a non-UTC source zone for one column
 
--- a/docs/timezones/timezone-handling.md
+++ b/docs/timezones/timezone-handling.md
@@ -162,9 +162,9 @@ A "Day of week" or "Month number" dimension grouped next to a DATE_TRUNC sibling
 
 **Files:** `packages/common/src/utils/timeFrames.ts`, `packages/backend/src/utils/QueryBuilder/MetricQueryBuilder.ts`
 
-### Per-dimension display opt-out (`convert_timezone: false`)
+### Per-dimension timezone opt-out (`convert_timezone: false`)
 
-By default every TIMESTAMP dimension follows the project timezone for display. Some columns (system timestamps, audit logs, pre-converted values) need to render in their raw warehouse value instead. Set `convert_timezone: false` on the dimension's YAML meta to opt that single column out:
+By default every TIMESTAMP dimension follows the project timezone. Some columns (system timestamps, audit logs, pre-converted values) need to use their raw warehouse value instead. Set `convert_timezone: false` on the dimension's YAML meta to opt that single column out:
 
 ```yaml
 - name: created_at_utc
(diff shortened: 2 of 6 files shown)


almeidabbm:
Current behavior matches the docs, might need discussion if we need to tackle it

